PR #5293, opened 2025-07-03
Meta: Add formatter to PropertyID code generator

tcl3:
This means I don't have to type as much when debugging things involving PropertyIDs.


tcl3:
> 🎉 🎉 🎉 Yes!!! 🎉 🎉 🎉

Glad it's not just me that has a use for this :laughing:

PR #5298, opened 2025-07-04
LibWeb: Fix hit testing of anonymous containers' contents

gmta:
A `PaintableWithLines` will first try to see if there are any fragments that have a hit. If not, it falls back to hit testing against its border box rect.

However, inline content is often hoisted out of its parent into an anonymous container to maintain the invariant that all layout nodes either have inline or block level children. If that's the case, we should not check the border box rect of the anonymous container, because we might trigger a hit too early if the node has previous siblings in its original parent node that overlap with their bounds.

By ignoring anonymous nodes, we leave the border box hit testing to the nearest non-anonymous ancestor, which correctly applies the hit testing order to its children.

### DIFF
--- a/Libraries/LibWeb/DOM/Document.cpp
+++ b/Libraries/LibWeb/DOM/Document.cpp
@@ -5660,11 +5660,10 @@ GC::RootVector<GC::Ref<Element>> Document::elements_from_point(double x, double 
     // 3. For each box in the viewport, in paint order, starting with the topmost box, that would be a target for
     //    hit testing at coordinates x,y even if nothing would be overlapping it, when applying the transforms that
     //    apply to the descendants of the viewport, append the associated element to sequence.
-    if (auto const* paintable_box = this->paintable_box(); paintable_box) {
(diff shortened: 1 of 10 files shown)


tcl3:
This change causes a couple of `elementsFromPoint` WPT tests to fail. This one for example: http://wpt.live/css/cssom-view/elementsFromPoint-inline-htb-ltr.html. In this case the topmost `target` element is included in the returned Vector of elements twice.

gmta:
@tcl3 

> This change causes a couple of elementsFromPoint WPT tests to fail.

Good catch - the new code keeps those WPT results intact, no new failures. I've discovered that the points that we pass into PaintableBox / PaintableWithLines / StackingContext's `::hit_test()` can be both transformed or untransformed according to the current applicable transforms, so it's a bit of a mess. This PR doesn't improve on that but it also doesn't make the situation worse, so I aim to pick that up in a new PR.

tcl3 (approved):

Issue #2893, opened 2024-12-12
The fetch process hangs when a fetch completion promise invokes a second fetch via `document.write`


trflynn89:
Repro HTML:
```html
<script>
fetch("frame.html")
    .then(response => response.text())
    .then(text => {
        let iframe = document.createElement("iframe");
        document.body.append(iframe);

        iframe.contentDocument.write(text);
        iframe.contentDocument.close();
    });
</script>
```

Where frame.html is:
```html
<script src=frame.js></script>
```
And frame.js is an empty file.

--------------------

All promise callbacks are executed on the microtask queue by way of `vm.host_enqueue_promise_job`. Only one microtask is allowed to execute at a time (see `EventLoop::perform_a_microtask_checkpoint`).

So, when we write the contents of frame.html to the iframe, we will eventually fetch the script's source URL in `HTMLScriptElement::prepare_script`. The call to `document.write` will block until this fetch is complete here:
https://github.com/LadybirdBrowser/ladybird/blob/66519af43f8aa1c0acac134ed1345417f64ef7be/Libraries/LibWeb/HTML/Parser/HTMLParser.cpp#L3035-L3039

But because we are already in a microtask, the promises involved in this inner fetch are not able to complete, as those microtasks are effectively blocked.

This is reached in the following WPT tests:
https://wpt.fyi/results/upgrade-insecure-requests/gen/iframe-blank-inherit.meta/unset/fetch.https.html?diff=&filter=ADC&run_id=5161602225799168&run_id=5177547862310912

sideshowbarker:
Seems to have been fixed by b1ccab81ada from #9105.

PR #5301, opened 2025-07-04
LibWeb/DOM: Abort any ongoing navigation in Document.open

shannonbooth:
Unfortunately this does not achieve it's intended effect
in at least the WPT test of:

document-open-cancels-javascript-url-navigation.html

As we do not implement step 2 of Document::abort. Since this does
not seem to break anything and is also updating an out of date
spec comment, put this step in place anyhow.

Spurred from investigation of #2893 as well looking into navigation
issues in general.

### DIFF
--- a/Libraries/LibWeb/DOM/Document.cpp
+++ b/Libraries/LibWeb/DOM/Document.cpp
@@ -707,6 +707,10 @@ WebIDL::ExceptionOr<Document*> Document::open(Optional<String> const&, Optional<
     // If document belongs to a child navigable, we need to make sure its initial navigation is done,
     // because subsequent steps will modify "initial about:blank" to false, which would cause
     // initial navigation to fail in case it was "about:blank".
+    //
+    // FIXME: We should get rid of this, this is not correct and is likely causing problems other than what
+    //        we are working around above. Hopefully this can be removed once we properly implement Document::abort
+    //        to cancel ongoing navigations.
     if (auto navigable = this->navigable(); navigable && navigable->container() && !navigable->container()->content_navigable_has_session_history_entry_and_ready_for_navigation()) {
         HTML::main_thread_event_loop().spin_processing_tasks_with_source_until(HTML::Task::Source::NavigationAndTraversal, GC::create_function(heap(), [navigable_container = navigable->container()] {
             return navigable_container->content_navigable_has_session_history_entry_and_ready_for_navigation();
--- a/Libraries/LibWeb/DOM/Document.cpp
+++ b/Libraries/LibWeb/DOM/Document.cpp
@@ -740,7 +740,9 @@ WebIDL::ExceptionOr<Document*> Document::open(Optional<String> const&, Optional<
     if (m_active_parser_was_aborted)
         return this;
 
-    // FIXME: 8. If document's browsing context is non-null and there is an existing attempt to navigate document's browsing context, then stop document loading given document.
+    // 8. If document's node navigable is non-null and document's node navigable's ongoing navigation is a navigation ID, then stop loading document's node navigable.
+    if (auto navigable = this->navigable(); navigable && navigable->ongoing_navigation().has<String>())
+        navigable->stop_loading();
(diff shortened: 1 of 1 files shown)


Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -740,7 +740,9 @@ WebIDL::ExceptionOr<Document*> Document::open(Optional<String> const&, Optional<
     if (m_active_parser_was_aborted)
```
gmta, line 744: Commit message: it's -> its

Also, doesn't seem auto-wrapped at 72 characters?

Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -707,6 +707,10 @@ WebIDL::ExceptionOr<Document*> Document::open(Optional<String> const&, Optional<
     // If document belongs to a child navigable, we need to make sure its initial navigation is done,
     // because subsequent steps will modify "initial about:blank" to false, which would cause
     // initial navigation to fail in case it was "about:blank".
+    //
+    // FIXME: We should get rid of this, this is not correct and is likely causing problems other than what
+    //        we are working around above. Hopefully this can be removed once we properly implement Document::abort
+    //        to cancel ongoing navigations.
```
kalenikaliaksandr, line 713: could you expand the comment a bit to clarify more specifically why it's wrong? I mean, all spin untils we have in navigation code are hacks to some degree, and ofc we should find a way to remove them eventually, but then lets put some actionable information into FIXME comments.

shannonbooth: I think the information here is actionable, no? At least more actionable than the existing comment which isnt even marked AD-HOC right now.

Even though we have spins in other places in navigation, at least most of those are tied to where we need to syncronously wait for something and let other stuff run in parallel.

I suspect doing this causes issues like this:
https://github.com/LadybirdBrowser/ladybird/commit/30377e6e3500049c5442fac9d242d3b795861116

Not sure what else we should be adding here

shannonbooth: Maybe should just say "by allowing other code to run" or something along those lines?

kalenikaliaksandr: if you could create a page that reproduces problems caused by this ad hoc step, then open an issue and add it to the comments, that would be perfect. otherwise I would rather prefix existing comment with AD-HOC.

shannonbooth: why is that?

kalenikaliaksandr: my point is that if I had to work on this function again I would not be able to tell what specifically is wrong and how to verify that the problem is gone. it also doesn't suggest alternative way to work around the issue. IMO it's kinda frustrating to have FIXMEs that essentially say "This is probably/likely/maybe wrong". if you disagree then feel free to merge it as is, I might be wrong here.

shannonbooth: Ah right, thanks!

What my comment is trying to say is that step 8. is what I understand we're meant to be doing here (aborting the ongoing navigation) instead of waiting for the ongoing navigation. I believe the test I refer to in the previous commit is one such example which we're not currently doing correctly since the javascript URL navigation should be stopped, not waited for.

Unfortunately this is in an extremely hairy part of navigation that ties into "apply the history step", but I pointed to Document::abort as I _think_ that is the primary step which needs to be implemented before that can be removed.

I'll think of a better comment to write though since it's confusing, I was avoiding using AD-HOC since it makes it seem to me like this spin is an implementation detail

Interested if you agree with my interpretation though, navigation is a bit of a mind boggler to me

shannonbooth: Aha, https://github.com/whatwg/html/issues/3975 seems to align with above, that is probably a good thing to link for background info since it has some good details in it

PR #5303, opened 2025-07-04
LibWeb: Fix partially selecting non-text nodes

### LINKED ISSUE #5289: Inconsistent range behavior between ladybird and Chrome/Firefox
### Summary

See attached test case.

### Operating system

Linux

### Steps to reproduce

1. Open attached test case.
2. Observe that `Bar` is selected.
3. This is not the case in Firefox or Chrome.

### Expected behavior

`Bar` should not be selected.

### Actual behavior

`Bar` does get selected

N/A

```html
<!DOCTYPE html>
<div id="foo">Foo<span id="bar">Bar</span></div>
<script>
  let range = document.createRange();
  range.setStart(bar, 1);
  range.setEnd(foo, 2);
  window.getSelection().addRange(range);
</script>
```

```shell
None
```

Ladybird:

Image

Firefox / Chrome:

Image

Image

None

### Contribute a patch?

Psychpsyo:
Fixes #5289

Steps 4 and 5 were swapped since marking all the nodes between the start and end of the selection now also marks the end node as full, even if it should be marked as End.  
There could be extra logic to avoid marking it if it is a text node, but this seems easier.


kalenikaliaksandr (changes requested):
this change doesn't simply swap steps 4 and 5 but rewrites them to do something different. why this NOTE and corresponding code and gone?
```
NOTE: If the start node is a descendant of the end node, we cannot traverse from it to the end node since the end node is before it in tree order.
```
also, please, put explanation into commit description, instead of PR description.

Psychpsyo:
> this change doesn't simply swap steps 4 and 5 but rewrites them to do something different. why this NOTE and corresponding code are gone?
> 
> ```
> NOTE: If the start node is a descendant of the end node, we cannot traverse from it to the end node since the end node is before it in tree order.
> ```
> 
> also, please, put explanation into commit description instead of PR description.

Yes, sorry. I've now added more information and put it in the actual commit message.

gmta:
This breaks some selections, such as on our own homepage:

Screenshot 2025-07-10 at 10 32 02

My selection was from the left of the "i" of "independent" to the right of the "o" of "browser". With this PR, the line after it and the "Get Involved" button are also selected. This does not happen on `master`.

Psychpsyo:
> This breaks some selections, such as on our own homepage:

Good point, fixed that now.
The issue was that the loop didn't stop at end nodes without children. (since it used the `stay_within` of `next_in_pre_order()` to walk into the end node, but not out of it, which doesn't work for nodes without children.)

Psychpsyo:
Did not notice that the test case needed to be updated due to text color stuff, this should be ready now.
Also, the failing popover test failure seems like an unrelated flake. (I can also run it just fine locally)

EDIT: Nevermind, test is still failing on MacOS

AtkinsSJ:
The test is failing on Mac CI, it needs some fuzziness added.

```
mismatch: pixel error count 204, with maximum error 1. (No fuzzy config defined)
```

PR #5304, opened 2025-07-04
LibWeb: Set Document origin for DOMParser created documents

tcl3:
Previously, a crash would occur when accessing the origin of a document created with DOMParser.

Relevant spec issue: https://github.com/whatwg/html/issues/11429

### DIFF
--- a/Libraries/LibWeb/DOM/Document.cpp
+++ b/Libraries/LibWeb/DOM/Document.cpp
@@ -850,7 +850,7 @@ GC::Ptr<HTML::WindowProxy const> Document::default_view() const
 
 URL::Origin const& Document::origin() const
 {
-    return m_origin.value();
+    return m_origin;
 }
 
 void Document::set_origin(URL::Origin const& origin)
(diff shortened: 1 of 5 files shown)


tcl3:
I've added a change to set the Document origin to be an opaque origin by default, since that matches what the spec says. I have also made the same change I made for `DOMParser.parseFromString()` for `Document.parseHTMLUnsafe()`.

Setting the origin default value is the only thing that is required to fix the crash demonstrated by the imported test, but the AD-HOC fixes are required to match the behavior of existing browsers, which I have tested with the following (run from  a local server with the domain `localhost`):

```html
<!DOCTYPE html>
<script>
    let policy = window.trustedTypes?.createPolicy('default', {
        createHTML: (input) => input
    });

    let unsafeHTML = "<!DOCTYPE html>"
    let maybeSafeHTML = policy?.createHTML(unsafeHTML) ?? unsafeHTML;
    let newDocument = new DOMParser().parseFromString(maybeSafeHTML, "text/html");
    console.log(`Original document domain: ${document.domain}, new document domain ${newDocument.domain}`);
    let newDocument2 = Document.parseHTMLUnsafe(unsafeHTML);
    console.log(`Original document domain: ${document.domain}, new document domain ${newDocument2.domain}`);
</script>
```

Without the AD-HOC fixes in this PR the domain of the new document would be empty, where it should match the domain of the original document.

shannonbooth (approved):

PR #5305, opened 2025-07-05
LibWeb: Serialize colors with unresolved components correctly

Calme1709:
We now serialize RGB, HSL and HWB colors correctly when we don't have all the required information (e.g. colors such as `rgb(sign(1px - 1em) 0 0 / 0)` that rely on compute-time information in `getPropertyValue()`)

Gains us 44 WPT subtests passes.

As part of this I have marked the existing `CalculatedStyleValue::resolve_*` methods as deprecated and added new ones which are built on top of the (spec adherent) `simplify_a_calculation_tree` algorithm. There is still more work to be done here to port all callers over to these new methods.

### DIFF
--- a/Libraries/LibWeb/CSS/CSSStyleValue.h
+++ b/Libraries/LibWeb/CSS/CSSStyleValue.h
@@ -393,7 +393,7 @@ public:
 
     virtual ValueComparingNonnullRefPtr<CSSStyleValue const> absolutized(CSSPixelRect const& viewport_rect, Length::FontMetrics const& font_metrics, Length::FontMetrics const& root_font_metrics) const;
 
-    virtual Color to_color(Optional<Layout::NodeWithStyle const&>, CalculationResolutionContext const&) const { return {}; }
+    virtual Optional<Color> to_color(Optional<Layout::NodeWithStyle const&>, CalculationResolutionContext const&) const { return {}; }
     Keyword to_keyword() const;
 
     virtual String to_string(SerializationMode) const = 0;
--- a/Libraries/LibWeb/CSS/StyleValues/CalculatedStyleValue.cpp
+++ b/Libraries/LibWeb/CSS/StyleValues/CalculatedStyleValue.cpp
@@ -2669,7 +2669,7 @@ bool CalculatedStyleValue::equals(CSSStyleValue const& other) const
     return m_calculation->equals(*other.as_calculated().m_calculation);
 }
 
-Optional<Angle> CalculatedStyleValue::resolve_angle(CalculationResolutionContext const& context) const
+Optional<Angle> CalculatedStyleValue::resolve_angle_deprecated(CalculationResolutionContext const& context) const
 {
     auto result = m_calculation->resolve(context);
     if (result.type().has_value() && result.type()->matches_angle(m_context.percentages_resolve_as))
@@ -2677,7 +2677,7 @@ Optional<Angle> CalculatedStyleValue::resolve_angle(CalculationResolutionContext
     return {};
 }
 
-Optional<Flex> CalculatedStyleValue::resolve_flex(CalculationResolutionContext const& context) const
+Optional<Flex> CalculatedStyleValue::resolve_flex_deprecated(CalculationResolutionContext const& context) const
 {
     auto result = m_calculation->resolve(context);
     if (result.type().has_value() && result.type()->matches_flex(m_context.percentages_resolve_as))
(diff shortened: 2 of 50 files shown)


Libraries/LibWeb/CSS/StyleValues/CalculatedStyleValue.cpp
```diff
@@ -2740,6 +2826,22 @@ Optional<double> CalculatedStyleValue::resolve_number_deprecated(CalculationReso
...
 
+Optional<double> CalculatedStyleValue::resolve_number(CalculationResolutionContext const& context) const
+{
+    auto result = resolve_value(context);
+
+    if (result.has_value() && result.value().type().has_value() && result.value().type()->matches_number(m_context.percentages_resolve_as)) {
+        auto value = result.value().value();
+
+        if (isnan(value))
+            return 0.;
```
AtkinsSJ, line 2846: This should really be done for all types, not just `<number>`. (So, in `resolve_value()`.)

I'm fuzzy on whether we resolve `calc()` in specified values or later, but [this section of the spec](https://drafts.csswg.org/css-values/#calc-range) also wants us to clamp values to the valid range, which might also go in `resolve_value()`?

Libraries/LibWeb/CSS/CSSStyleValue.h
```diff
@@ -393,7 +393,7 @@ class CSSStyleValue : public RefCounted<CSSStyleValue> {
 
     virtual ValueComparingNonnullRefPtr<CSSStyleValue const> absolutized(CSSPixelRect const& viewport_rect, Length::FontMetrics const& font_metrics, Length::FontMetrics const& root_font_metrics) const;
 
-    virtual Color to_color(Optional<Layout::NodeWithStyle const&>, CalculationResolutionContext const&) const { return {}; }
+    virtual Optional<Color> to_color(Optional<Layout::NodeWithStyle const&>, CalculationResolutionContext const&) const { return {}; }
```
AtkinsSJ, line 396: From the commit message:
> Slight regression in that we now serialize (RGB, HSL and HWB) colors
with components that rely on compute-time information as an empty
string, but that will be fixed in the next commit.

Do any tests fail at this commit? That's something we really want to avoid because it means the commit isn't atomic, and it makes bisecting harder.

I guess if we were doing the wrong thing before then we're probably not affecting test results though.

AtkinsSJ (changes requested):

Libraries/LibWeb/CSS/CSSStyleValue.h
Calme1709: I have run comparisons for the `css/css-colors` and `css/css-values` suites and can't see any new failures as a result of this change.

The wrong thing we were doing before was always resolving relative lengths `0`, so the only scenario I can think of is if we had a false positive because two relative lengths cancelled eachother out (e.g. `rgb(calc(1 * sign(1em - 1em)) 255 255 / 1)` but it doesn't seem as though we have any of those.

Libraries/LibWeb/CSS/StyleValues/CalculatedStyleValue.cpp
Calme1709: >This should really be done for all types, not just <number>. (So, in resolve_value().)

Agreed this should definitely be the case _eventually_

Only doing this for `<number>` matches the behavior of the (now) deprecated resolve methods and my intention was to make minimal changes until we have more callers migrated over to these new methods so that we can properly test changes when making them.

> I'm fuzzy on whether we resolve calc() in specified values or later, but this section of the spec also wants us to clamp values to the valid range, which might also go in resolve_value()?

We resolve values at use-time (or compute time for some contexts which we can guarantee will resolve, e.g. colors as percentages are disallowed), we merely simplify at parse time (see step 6 [here](https://drafts.csswg.org/css-values-4/#parse-a-calculation)) and compute time (see [here](https://drafts.csswg.org/css-values/#calc-computed-value))

The computed-value portion of the linked spec is covered in step 1 [here](https://drafts.csswg.org/css-values/#calc-computed-value) and I agree that the resolved value portion should be done in `resolve_value` but see above for why I think this should be handled later.

I have added FIXMEs for each of range checking and NaN censorship in `resolve_value` for now.

Libraries/LibWeb/CSS/CSSStyleValue.h
AtkinsSJ: That's fine then, I misunderstood what you meant by a regression.

Libraries/LibWeb/CSS/StyleValues/CalculatedStyleValue.cpp
AtkinsSJ: > Only doing this for <number> matches the behavior of the (now) deprecated resolve methods and my intention was to make minimal changes until we have more callers migrated over to these new methods so that we can properly test changes when making them.

That's both very sensible and also more self control than I have sometimes. :laughing:

And thanks, that all makes sense.

AtkinsSJ (approved):

PR #2854, opened 2024-12-09
LibWeb: Implement Content Security Policy Level 3

Lubrsi:
This implements the core of Content Security Policy Level 3. This does not implement directives outside of the base specification yet, such as `trusted-types`, `require-trusted-types-for` and `upgrade-insecure-requests`. Additionally, the `report-to` directive does not have an effect yet, as it relies on implementing the Reporting specification.

There are several implementations and fixes done on the side to support this implementation, such as implementing a new GenericLexer method and a new TC39 Stage 3 proposal called "Dynamic Code Brand Checks".

There are several specification issues that I found whilst implementing this. They are marked with FIXMEs through this change set. I haven't filed them yet as there are many and could already be known about, as there are 178 open issues on their GitHub repository. However, I thought I'd get this open for review before I start filing spec issues :^).

Additionally, this does not import any WPT tests, as it relies on the WPT server to check CSP reports, and it does some cross-origin resource requesting that confuses the WPT importer, as it'll treat such requests as part of the path of the wpt.live origin, instead of being a different origin entirely. This PR is big enough as is, without going into fixing the WPT importer.

PR #5308, opened 2025-07-05
LibWeb/CSP: Implement manifest-src, media-src and object-src directives

Lubrsi:
Part 9 of splitting up https://github.com/LadybirdBrowser/ladybird/pull/2854

A few more simple directives before introducing the more complex script-src directives.

### DIFF
--- a/Libraries/LibWeb/ContentSecurityPolicy/Directives/MediaSourceDirective.h
+++ b/Libraries/LibWeb/ContentSecurityPolicy/Directives/MediaSourceDirective.h
@@ -1,0 +1,28 @@
+/*
+ * Copyright (c) 2024, Luke Wilde <luke@ladybird.org>
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#pragma once
+
+#include <LibWeb/ContentSecurityPolicy/Directives/Directive.h>
+
+namespace Web::ContentSecurityPolicy::Directives {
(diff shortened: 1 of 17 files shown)


Libraries/LibWeb/ContentSecurityPolicy/Directives/MediaSourceDirective.h
```diff
@@ -0,0 +1,28 @@
...
+
+// https://w3c.github.io/webappsec-csp/#directive-media-src
+class MediaSourceDirective final : public Directive {
+    GC_CELL(MediaSourceDirective, Directive)
+    GC_DECLARE_ALLOCATOR(MediaSourceDirective);
+
+public:
+    virtual ~MediaSourceDirective() = default;
+
+    [[nodiscard]] virtual Result pre_request_check(GC::Heap&, GC::Ref<Fetch::Infrastructure::Request const>, GC::Ref<Policy const>) const override;
```
shannonbooth: 🤔 After seeing this pattern a few more times, I'm now starting to wonder if we should be making this enum nodiscard instead of all of the individual overrides

Lubrsi: Oh, I forgot about that you could do that! I'll add that in now before it's changing a lot of files in one PR :^)

PR #5314, opened 2025-07-05
LibWeb: Respect presentation attributes that apply to not all elements

tcl3:
Some SVG presentation attributes are only supported on certain elements. We now support these special cases for attributes and elements that we currently have implemented.

Fixes 16 WPT subtests in `svg`.


Libraries/LibWeb/SVG/SVGElement.cpp
```diff
@@ -32,75 +33,92 @@ void SVGElement::initialize(JS::Realm& realm)
...
+        , supported_elements(move(supported_elements))
     {
     }
 
     CSS::PropertyID id;
     StringView name;
+    Vector<FlyString> supported_elements;
 };
 
-static Array const attribute_style_properties {
```
tcl3, line 45: I had to move this list into a method to be able to use `SVG::TagNames::*`. Leaving the list here caused the FlyStrings to appear to be uninitialized.

I'm assuming this is because the static variable is initialized before the FlyString itself, but I'm not sure.

Libraries/LibWeb/SVG/SVGElement.cpp
```diff
@@ -32,75 +33,92 @@ void SVGElement::initialize(JS::Realm& realm)
...
-    NamedPropertyID(CSS::PropertyID::TransformOrigin),
-    NamedPropertyID(CSS::PropertyID::UnicodeBidi),
-    NamedPropertyID(CSS::PropertyID::Visibility),
-    NamedPropertyID(CSS::PropertyID::WhiteSpace),
-    NamedPropertyID(CSS::PropertyID::WordSpacing),
-    NamedPropertyID(CSS::PropertyID::WritingMode),
-};
+static Vector<NamedPropertyID> attribute_style_properties()
+{
+    static Vector const properties = {
```
trflynn89: Does this need to be a Vector? Can we still store an Array and return a ReadonlySpan?
(if it does need to be a Vector for some reason, let's at least return it by const&)

Libraries/LibWeb/SVG/SVGElement.cpp
```diff
@@ -32,75 +33,92 @@ void SVGElement::initialize(JS::Realm& realm)
...
-    NamedPropertyID(CSS::PropertyID property_id)
+    NamedPropertyID(CSS::PropertyID property_id, Vector<FlyString> supported_elements = {})
         : id(property_id)
         , name(CSS::string_from_property_id(property_id))
+        , supported_elements(move(supported_elements))
     {
     }
 
     CSS::PropertyID id;
     StringView name;
```
trflynn89: Unrelated to your change, just noticed while we're here - this could be a FlyString right? That would make the name comparisons below O(1) in the happy case I think

tcl3: I don't think we'll get a constant time speedup at the moment because comparisons are done case-insensitively - although I'm not sure SVG attributes should be case-insensitive.

Libraries/LibWeb/SVG/SVGElement.cpp
tcl3: That's better. I've changed this to return a `ReadonlySpan`.

Libraries/LibWeb/SVG/SVGElement.cpp
trflynn89: Right, that's what I meant by the happy case - there's a fast-path for `FlyString::equals_ignoring_ascii_case` for when we are comparing identical FlyStrings.

tcl3: D'oh, of course! Changed to `FlyString`.

PR #5315, opened 2025-07-05
LibWeb: Change SessionHistoryTraversalQueue to use priority queue

prajjwal000:
If src of an iframe/frame changes and if the new document creates a new child navigable we get into a deadlock on SessionHistoryTraversalQueue.So use a priority queue to run the child navigable creation steps with higher priority.

Fixes at least one wpt test that was previously timing out:
- html/syntax/parsing/html5lib_template.html?run_type=uri


tcl3:
This change fixes these WPT tests :tada: :
* http://wpt.live/html/syntax/parsing/html5lib_template.html?run_type=uri
* http://wpt.live/html/syntax/parsing/html5lib_tests1.html?run_type=uri
* http://wpt.live/html/syntax/parsing/html5lib_tests5.html?run_type=uri
* http://wpt.live/html/syntax/parsing/html5lib_tests15.html?run_type=uri
* http://wpt.live/html/syntax/parsing/html5lib_tests16.html?run_type=uri
* http://wpt.live/html/syntax/parsing/html5lib_tests19.html?run_type=uri

These tests were imported a while ago, but were skipped due to timing out. You can unskip these tests by removing them from [TestConfig.ini](https://github.com/LadybirdBrowser/ladybird/blob/85001185a78a218466da118a66ca54ee61497e1a/Tests/LibWeb/TestConfig.ini#L36-L42). The expectations for these tests will then need to be regenerated, which you can do by running: `./Meta/ladybird.py run test-web --rebaseline`

prajjwal000:
I think I fixed the changes. Could someone review it again.

Libraries/LibWeb/HTML/SessionHistoryTraversalQueue.cpp
```diff
@@ -27,10 +27,16 @@ void SessionHistoryTraversalQueueEntry::visit_edges(JS::Cell::Visitor& visitor)
...
     m_timer = Core::Timer::create_single_shot(0, [this] {
+        while (m_higher_priority_queue.size() > 0) {
+            auto entry = m_higher_priority_queue.take_first();
+            entry->execute_steps();
+        }
+
         if (m_is_task_running && m_queue.size() > 0) {
             m_timer->start();
             return;
         }
```
gmta, line 33: Aren't you bypassing the `m_is_task_running` logic with the higher priority tasks now?

Libraries/LibWeb/HTML/SessionHistoryTraversalQueue.cpp
```diff
@@ -44,19 +50,28 @@ void SessionHistoryTraversalQueue::visit_edges(JS::Cell::Visitor& visitor)
...
     Base::visit_edges(visitor);
     visitor.visit(m_queue);
+    visitor.visit(m_higher_priority_queue);
 }
 
-void SessionHistoryTraversalQueue::append(GC::Ref<GC::Function<void()>> steps)
+void SessionHistoryTraversalQueue::append(GC::Ref<GC::Function<void()>> steps, int priority)
 {
-    m_queue.append(SessionHistoryTraversalQueueEntry::create(vm(), steps, nullptr));
+    if (priority == 1) {
```
gmta: You're not implementing actual _priority_ as in an int designating some relative weight to the individual tasks to execute, but rather a queue selection. I feel this would work better as an enum:

```
enum class QueuePriority : u8 {
    High,
    Normal,
};
```

Libraries/LibWeb/HTML/SessionHistoryTraversalQueue.cpp
prajjwal000: Thats the reason for the deadlock. The flow goes like this 
1. Iframe src attribute change
2. process iframe attributes -> navigate -> begin_navigation
3. populate_session_history_entry_document which defers document load and in its completion_steps appends finalize_a_cross_document_navigation to the queue.
4. finalize_a_cross_document_navigation runs and in the apply_the_history_step we call spin_until.

5. document loads and if that document has an iframe/frame it calls create_new_child_navigable which appends steps for set_content_navigable to the queue. But since the parent's step is running it just waits.

So, as a solution I thought of making the child_navigable creation flow check if traversable is in middle of finalize_a_cross_document and bypass the queue for those steps.

The solution is a bit hacky, is there be a better solution?

Libraries/LibWeb/HTML/SessionHistoryTraversalQueue.cpp
prajjwal000: yes, I will implement that.

Libraries/LibWeb/HTML/SessionHistoryTraversalQueue.cpp
gmta: OK, my point was that `m_is_task_running` guards against iterating over the queue again while we're still processing it, and we have no such protection for the new, higher priority queue.

gmta: I agree it feels a bit hacky though - have to think about a better approach 😅

Libraries/LibWeb/HTML/TraversableNavigable.cpp
```diff
@@ -1179,6 +1179,12 @@ TraversableNavigable::HistoryStepResult TraversableNavigable::update_for_navigab
     // 1. Let step be traversable's current session history step.
     auto step = current_session_history_step();
 
+    // AD-HOC: If the parent traversable is inside a cross-document navigation and hasn't updated the target
+    //         step yet, then add 1 to the step.
+    if (m_session_history_step_is_locked) {
+        step++;
+    }
```
kalenikaliaksandr: if reaching `update_for_navigable_creation_or_destruction()` during parent traversable navigation is a problem, then could we rearrange callers of this function to wait for navigation to finish?

Libraries/LibWeb/HTML/HTMLIFrameElement.cpp
```diff
@@ -97,9 +97,17 @@ void HTMLIFrameElement::post_connection()
 
         if (auto navigable = content_navigable()) {
             auto traversable = navigable->traversable_navigable();
+            // AD-HOC: Since the session history traversal queue is not actually parallel, there will be a deadlock if
+            //         we append a new child navigable creation steps to the queue while we're in execution of steps for the parent traversable.
+            //         To avoid this, we set the priority of the new child navigable steps greater than parent's ie. 1.
+            int priority = 0;
+            if (traversable->is_session_history_step_locked()) {
+                priority = 1;
+            }
```
kalenikaliaksandr: > Since the session history traversal queue is not actually parallel

"parallel" in specification terminology means that queue items have to be processed sequentially but (optionally) in parallel with other tasks (not initiated by SHTQ). here you break sequential processing by adding priority. I would prefer to avoid doing that since that's assumption made by many navigation spec algorithms.

prajjwal000:
After going on multiple wrong side tangents I think I got the issue. 

1. Load_document queues a global task which will run after document becomes active.
2. Navigable's SHTQ entry runs and queues global task for activating document.
3. If queue is jammed just right, load_document might run in another navigable's SHTQ entry's spin_until.
4. If the new document has its own navigable it appends steps to SHTQ and calls spin_until leading to nested spin_until deadlock.

We should wait for document load to finish before proceeding to next entry.

Libraries/LibWeb/HTML/SessionHistoryTraversalQueue.cpp
```diff
@@ -31,32 +33,44 @@ SessionHistoryTraversalQueue::SessionHistoryTraversalQueue()
             m_timer->start();
             return;
         }
+
         while (m_queue.size() > 0) {
+            // Note: Check if current entry's load_document finished before executing the next entry
+            if (m_current_document && m_current_document->ready_state() == "loading" && m_current_document->is_active()) {
+                m_timer->start();
+                return;
+            }
```
kalenikaliaksandr: instead of making SHTQ responsible for checking the ready state, can we pass a Core::Promise into the "steps" and make SHTQ wait until this promise is resolved before proceeding to the next item? I think this change addresses one instance of a more general problem that although queued steps are considered synchronous that's not always true in practice. we need an API that allows us to explicitly signal when a step has completed.

prajjwal000:
I am in a bit of pickle trying to implement promises.
 "steps" has to resolve promise when document is loaded but if I use spin_until to check for document load we are back at square one with nested spin_until deadlock if new document has navigable.
 Document checking in SHTQ works because it runs even in the the spin_until of new navigable creation.
 
 Do we have any alternative to spin_until which could be used?

kalenikaliaksandr:
> I am in a bit of pickle trying to implement promises. "steps" has to resolve promise when document is loaded but if I use spin_until to check for document load we are back at square one with nested spin_until deadlock if new document has navigable. Document checking in SHTQ works because it runs even in the the spin_until of new navigable creation.

can we pass promise through all navigation algorithms to the point when document is marked as loaded? it's likely lots of places that have to be updated, but would be a lot nicer than using spin_untils.

Libraries/LibWeb/HTML/SessionHistoryTraversalQueue.cpp
```diff
@@ -31,12 +32,21 @@ SessionHistoryTraversalQueue::SessionHistoryTraversalQueue()
             m_timer->start();
             return;
         }
+
         while (m_queue.size() > 0) {
+            if (m_current_promise && !m_current_promise->is_resolved() && !m_current_promise->is_rejected()) {
+                m_timer->start();
+                return;
+            }
```
kalenikaliaksandr, line 39: can we use `Promise::await()` instead of checking whether promise resolved in a loop?

prajjwal000: Using await would lead to deadlocks as it is similar to spin_until in regard to nested deadlocks.
It wouldn't be able to check its condition when load document resolves the promise , appends steps to SHTQ and uses spin_until to check for steps completion.

prajjwal000:
There still is a bug that if there are a lot of iframes and all of them are changing their src non-stop we may get a hangout .
I am looking into it but for practical sites it should be work fine.

prajjwal000:
Interestingly the bug doesn't have anything to do with SHTQ. Its in- 

`void ReadableStreamDefaultReader::read_all_bytes(GC::Ref<ReadLoopReadRequest::SuccessSteps> success_steps, GC::Ref<ReadLoopReadRequest::FailureSteps> failure_steps)
`
The code is very opaque so I dont really understand it well but load_html_document puts parsing steps and promise resolution in success steps here. Interestingly sometimes neither success_steps nor failure_steps are called.

```
void ReadableByteStreamController::pull_steps(GC::Ref<ReadRequest> read_request)
{
    auto& realm = this->realm();
    dbgln("ReadableByteStreamController: pull_steps");

    // 1. Let stream be this.[[stream]].

    // 2. Assert: ! ReadableStreamHasDefaultReader(stream) is true.
    VERIFY(readable_stream_has_default_reader(*m_stream));

    // 3. If this.[[queueTotalSize]] > 0,
    if (m_queue_total_size > 0) {
        // 1. Assert: ! ReadableStreamGetNumReadRequests(stream) is 0.
        VERIFY(readable_stream_get_num_read_requests(*m_stream) == 0);

        // 2. Perform ! ReadableByteStreamControllerFillReadRequestFromQueue(this, readRequest).
        readable_byte_stream_controller_fill_read_request_from_queue(*this, read_request);

        // 3. Return.
        return;
     ///// Other code
    }
```

In successful cases readable_byte_stream_controller_fill_read_request_from_queue(*this, read_request) is called
So it might be some race condition of not having chunks available.

prajjwal000:
It was a race between document population and stream popuation of Blob.

prajjwal000:
It works now but I am not sure about my fix for stream population. Is it fine?

Libraries/LibWeb/HTML/Navigable.cpp
```diff
@@ -1315,15 +1316,18 @@ WebIDL::ExceptionOr<void> Navigable::populate_session_history_entry_document(
...
         // NOTE: This check is not in the spec but we should not continue navigation if navigable has been destroyed.
         if (has_been_destroyed())
             return;
 
         // 1. If navigable's ongoing navigation no longer equals navigationId, then run completionSteps and abort these steps.
         if (navigation_id.has_value() && (!ongoing_navigation().has<String>() || ongoing_navigation().get<String>() != *navigation_id)) {
-            if (completion_steps)
+            if (completion_steps) {
+                // Note: Use Core::Promise to signal SessionHistoryTraversalQueue that it can continue to execute next entry.
+                promise->resolve({});
```
konradekk: it’d probably be good to name the `Promise` variables differently than plain `promise` (everywhere), e.g. `continue_session_history_processing_signal` or something along those lines?

Libraries/LibWeb/HTML/TraversableNavigable.h
```diff
@@ -86,14 +86,14 @@ class TraversableNavigable final : public Navigable {
     void definitely_close_top_level_traversable();
     void destroy_top_level_traversable();
 
-    void append_session_history_traversal_steps(GC::Ref<GC::Function<void()>> steps)
+    void append_session_history_traversal_steps(GC::Ref<GC::Function<void()>> steps, NonnullRefPtr<Core::Promise<Empty>> promise_to_signal_steps_completion)
     {
-        m_session_history_traversal_queue->append(steps);
+        m_session_history_traversal_queue->append(steps, promise_to_signal_steps_completion);
```
kalenikaliaksandr: sorry, maybe I should've given more detailed explanation in https://github.com/LadybirdBrowser/ladybird/pull/5315#discussion_r2214754993
my proposal was to make SHTQ responsible for allocating promise that is then passed into a task execution steps, like `entry->execute_steps(promise);`

Libraries/LibWeb/HTML/SessionHistoryTraversalQueue.cpp
kalenikaliaksandr: I think it could be addressed by making `SessionHistoryTraversalQueue::append()` return a promise that is resolved when task steps are done completed, so instead of using `spin_until()` we could use `then()`.

Libraries/LibWeb/HTML/Navigable.cpp
```diff
@@ -1770,22 +1775,35 @@ void Navigable::begin_navigation(NavigateParams params)
         // 9. Attempt to populate the history entry's document for historyEntry, given navigable, "navigate",
         //    sourceSnapshotParams, targetSnapshotParams, userInvolvement, navigationId, navigationParams,
         //    cspNavigationType, with allowPOST set to true and completionSteps set to the following step:
-        populate_session_history_entry_document(history_entry, source_snapshot_params, target_snapshot_params, user_involvement, navigation_id, navigation_params, csp_navigation_type, true, GC::create_function(heap(), [this, history_entry, history_handling, navigation_id, user_involvement] {
+
+        // Note: Use Core::Promise to signal SessionHistoryTraversalQueue that it can continue to execute next entry.
+        auto signal_to_continue_session_history_processing = Core::Promise<Empty>::construct();
+        populate_session_history_entry_document(history_entry, source_snapshot_params, target_snapshot_params, user_involvement, navigation_id, navigation_params, csp_navigation_type, true, GC::create_function(heap(), [this, signal_to_continue_session_history_processing, history_entry, history_handling, navigation_id, user_involvement] {
```
prajjwal000: My bad for misinterpreting your intention, that would be a better way of structuring code. 
Only blocker would be the case when promise is resolved outside of steps as here - populate_session_history_entry_document which calls load_document queues a promise resolve(on the condition that document becomes active in the corresponding SHTQ step processing) .

One solution could be to have steps become GC::Ref<GC::Function<NonnullRefPtr<Core::Promise<AK::Empty>>()>> and be the one to return their promise. Then if promise has to be initialised before the steps it could be passed to them as required here.

PS. Sorry for the trouble, I am a new dev and english isn't my native language.

prajjwal000: I pushed the changes so you can review that.

kalenikaliaksandr: thanks, the latest change is a step in the right direction, but could we make SHTQ steps accept promise, instead of returning it, which means instead of `GC::Ref<GC::Function<NonnullRefPtr<Core::Promise<Empty>>()>>` we would have `GC::Ref<GC::Function<void(NonnullRefPtr<Core::Promise<Empty>>)>>`. why duplicate promise construction into every steps instance when we could make SHTQ responsible for that. does that make sense?

Libraries/LibWeb/DOM/DocumentLoading.cpp
```diff
@@ -201,7 +215,7 @@ static WebIDL::ExceptionOr<GC::Ref<DOM::Document>> load_xml_document(HTML::Navig
 }
 
 // https://html.spec.whatwg.org/multipage/document-lifecycle.html#navigate-text
-static WebIDL::ExceptionOr<GC::Ref<DOM::Document>> load_text_document(HTML::NavigationParams const& navigation_params, MimeSniff::MimeType type)
+static WebIDL::ExceptionOr<GC::Ref<DOM::Document>> load_text_document(HTML::NavigationParams const& navigation_params, MimeSniff::MimeType type, RefPtr<Core::Promise<Empty>> signal_to_continue_session_history_processing)
```
kalenikaliaksandr: it should be possible to pass NonnullRefPtr here

Libraries/LibWeb/DOM/DocumentLoading.cpp
```diff
@@ -265,7 +282,7 @@ static WebIDL::ExceptionOr<GC::Ref<DOM::Document>> load_text_document(HTML::Navi
 }
 
 // https://html.spec.whatwg.org/multipage/document-lifecycle.html#navigate-media
-static WebIDL::ExceptionOr<GC::Ref<DOM::Document>> load_media_document(HTML::NavigationParams const& navigation_params, MimeSniff::MimeType type)
+static WebIDL::ExceptionOr<GC::Ref<DOM::Document>> load_media_document(HTML::NavigationParams const& navigation_params, MimeSniff::MimeType type, RefPtr<Core::Promise<Empty>> signal_to_continue_session_history_processing)
```
kalenikaliaksandr: same, it should be possible to pass NonnullRefPtr here.

Libraries/LibWeb/HTML/Navigable.cpp
prajjwal000: Its just that I don't know how to resolve the promise when they aren't in the steps. Promise could be resolved by a task that is queued even before the corresponding steps are appended to SHTQ.

Libraries/LibWeb/DOM/DocumentLoading.cpp
prajjwal000: sorry, it got left there when I was refactoring

Libraries/LibWeb/HTML/Navigable.cpp
prajjwal000: The flow goes like this-
1. If iframe has a change in src, we initialise a new history_entry and call populate_session_history_entry_document.

2. populate_session_history_entry_document calls load_document which returns a document and queues parsing steps as global task (this queued task won't run right now because document isn't active).

3. At the end it calls completion_steps which appends steps to SHTQ. It is in these appended steps that the document might become active. The whole reason we were getting hangouts was because those queued parsing steps could run in next SHTQ entry.

Promise has to be passed to parsing steps which are queued even before the the corresponding SHTQ steps are appended.

PR #5316, opened 2025-07-05
CMake: Add unity build support

R-Goc:
This commit adds unity build support behind ENABLE_UNITY_BUILD. Does not include all libs as some are too small or don't build on windows yet.


R-Goc:
Not sure if it should be enabled by default. It is for now so that CI can actually run on it.

Libraries/LibWeb/CMakeLists.txt
```diff
@@ -1021,3 +1021,1213 @@ if (ENABLE_SWIFT)
...
+    function(add_sources_to_unity_group)
+        set(CURRENT_GROUP "${_UNITY_GROUP_COUNTER}")
+        math(EXPR _UNITY_GROUP_COUNTER "${_UNITY_GROUP_COUNTER} + 1")
+
+        set_source_files_properties(${ARGN} PROPERTIES UNITY_GROUP ${CURRENT_GROUP})
+
+        set(_UNITY_GROUP_COUNTER ${_UNITY_GROUP_COUNTER} PARENT_SCOPE)
+    endfunction()
+
+    add_sources_to_unity_group(
```
gmta: This seems like a pretty maintenance-heavy feature, plus the commit message does not describe what the advantages / disadvantages are. How are we going to prevent bitrot, similar to the gn situation?

R-Goc: This solution was indeed bad.

R-Goc:
Some performance numbers for building LibWeb from a clean state: 
```
unity build with default 8 batch size:
[1675/1675] Linking CXX shared library bin\lagom-web.dll
real    7m 57.31s

normal build:
[2895/2895] Linking CXX shared library bin\lagom-web.dll
real    11m 33.94s
```
Memory usage increase is insignificant at this batch size. For a random source file I looked at it's rebuild was 2x more expensive (from 3 to 6 seconds) for the 8 batch size. Making the batches 32 large would reduce build time by 30 more seconds, but it increases memory use a lot and would make incremental builds very bad. If there was a way to get the bindings sources to be built in a unity build then it would make a huge difference. On my machine the build of the unity sources finishes before 2 minutes of the 8 minute build time, the rest being the generated bindings. ~400 of the bindings have impl_from functions that simply cannot be compiled in a unity build. The other ~500 causes a lot of issues due to being unable to resolve URL::URL from LibURL with Web::CSS::URL. Some singular files have issues with Web::Crypto and Crypto:: but those would be an easy exclude.

R-Goc:
Not much maintainer interest. This would slow down incremental builds slightly and require maintenance, mostly due to bindings including everything. As such making bindings include what they need would be a better step. At that point maybe a unity build would make sense for just the ~500 bindings that don't have a function that only differs in return type, as those are all rebuilt all the time either way. But in the current status this would need changes to the ~60 files and would require writing ::URL in large parts of LibWeb.

PR #5320, opened 2025-07-06
LibWeb: Handle null namespace in prefix map serializing XML

shannonbooth:
A "[namespace prefix map](https://w3c.github.io/DOM-Parsing/#the-namespace-prefix-map
)", is meant to also hold null namespaces:
    
 > where namespaceURI values are the map's unique keys
 > (which can include the null value representing no namespace)
    
Which we previously neglected. This resulted in a crash for the updated WPT test.


AK/Optional.h
```diff
@@ -656,6 +657,16 @@ ALWAYS_INLINE constexpr bool operator==(Optional<T> const& first, OptionalNone)
     return !first.has_value();
 }
 
+template<typename T>
+struct Traits<Optional<T>> : public DefaultTraits<Optional<T>> {
+    static unsigned hash(Optional<T> const& optional)
+    {
+        if (!optional.has_value())
+            return 134;
```
tcl3: It might be good to assign this magic number to a variable and/or leave a comment making it clear that the value is arbitrary.

I'm not sure how much it matters, but arbitrary values in hashing functions are often prime numbers.

shannonbooth: Yeah left a comment for it, I didn't think particularly hard about it to be honest, just that 0 would have been a bad choice 😅

PR #5278, opened 2025-07-03
CMake: Add a flatpak build for org.ladybird.Ladybird

ADKaster:
This build depends on the KDE Flatpak SDK, and builds any missing
dependencies manually as source modules.

The flatpak can be built with the following command:

```sh
flatpak-builder --user --force-clean --install-deps-from=flathub \
    --ccache --repo=Build/repo --install Build/flatpak \
    Meta/CMake/flatpak/org.ladybird.Ladybird.json
```

After building, the flatpak can be run with:

```sh
flatpak run --user --devel org.ladybird.Ladybird
```

If there are issues launching RequestServer, the .pid and .sock files
under $XDG_RUNTIME_DIR may need removed.

```sh
flatpak run --user --command=sh --devel org.ladybird.Ladybird
rm -f $XDG_RUNTIME_DIR/Ladybird.*
```

PR #5325, opened 2025-07-06
CMake: Prep work for flatpak build to make installing better

ADKaster:
- Make it easier to link against non-vcpkg angle
- Set visibility for imported skia skcms target
- Install missing files for settings gui, mostly
- do a find_package for the header-only fast_float library.

cc @fgaz . presumably the new hard dep on ANGLE causes/caused issues for the nixpkgs port.


fgaz:
oof yes angle is going to be very difficult to package

ADKaster:
re: angle, in https://github.com/LadybirdBrowser/ladybird/pull/5278 I opted to give up and just use `glcient sync` plus its native gn build.

It pulls like 20 GiB of dependencies including its own build of clang++ to use to compile itself.

Ali at least showed me a similar approach for the AUR PKGBUILD he maintains, though it doesn't look like that's been pushed to ladybird-git :thinking: .

nico:
> including its own build of clang++ to use to compile itself.

That's only around 50 MB. And you don't _have_ to use it as compiler with the right gn args (but then you're in unsupported land. Several others do that though, and it'll likely work).

fgaz:
> It pulls like 20 GiB of dependencies including its own build of clang++ to use to compile itself.

Unfortunately that isn't going to fly in nixpkgs, where the fetch and build phases are completely separate. We also need need to supply our own compiler and build tools.

We have a tool that converts gclient to a workable format (we do have to build chromium after all) but it's sparsely documented and when I ran it on angle it failed :(

alimpfard:
>  it doesn't look like that's been pushed to ladybird-git

Yeah I'm still debating whether I want to maintain an ANGLE (and skia I suppose) port, so that's on hold for now

PR #5326, opened 2025-07-06
Meta+RequestServer: Remove local download of ca-certificates

ADKaster:
We haven't required a local copy of the ca-certificates since switching
to OpenSSL as the backend for TLS. Remove the script to download the
PEM file, and update the tests to use the system's CA certificates.


Services/RequestServer/main.cpp
```diff
@@ -55,10 +55,8 @@ ErrorOr<int> serenity_main(Main::Arguments arguments)
     if (wait_for_debugger)
         Core::Process::wait_for_debugger_and_break();
 
-    // Ensure the certificates are read out here.
-    if (certificates.is_empty())
-        certificates.append(TRY(find_certificates(serenity_resource_root)));
-    else
+    // FIXME: Update RequestServer to support multiple custom root certificates.
+    if (!certificates.is_empty())
         RequestServer::g_default_certificate_path = certificates.first();
```
ADKaster, line 48: Free FIXME. It hasn't worked to pass multiple `--certificate=` arguments for a while now. Whether we want that or not, now the code better respects the fact that we only have the one.

Services/RequestServer/CMakeLists.txt
```diff
@@ -7,16 +7,20 @@ set(SOURCES
...
 )
 
+set(GENERATED_SOURCES
+    RequestClientEndpoint.h
+    RequestServerEndpoint.h
+)
+
 if (ANDROID)
-    add_library(requestserverservice SHARED
+    list(APPEND SOURCES
```
ADKaster, line 17: wait I broke this, this needs to be shared on el Android

Tests/LibDNS/TestDNSResolver.cpp
```diff
@@ -93,8 +79,7 @@ TEST_CASE(test_tls)
         [&] -> ErrorOr<DNS::Resolver::SocketResult> {
             Core::SocketAddress addr = { IPv4Address::from_string("1.1.1.1"sv).value(), static_cast<u16>(853) };
 
-            TLS::Options options;
-            options.root_certificates_path = locate_ca_certs_file();
+            TLS::Options options = {};
```
alimpfard, line 82: Since you're dropping this already, could you also remove the cert path from https://github.com/LadybirdBrowser/ladybird/blob/809e465e05352f3360f26ac3ab94e5bc19dbcfa6/Utilities/dns.cpp#L81 ?

ADKaster: you don't want the option to pass a different root cert to your dns tool? I could see `if (!cert_path.is_empty()) options.cert_path = cert_path;` making sense. iirc the options struct holds an Optional

alimpfard: I suppose that could come in handy, so sure, let's do that.

gmta (approved):

PR #5331, opened 2025-07-06
WIP: Remove Qt linkage from WebContent and WebWorker when not necessary

ADKaster:
For some reason, this causes WebContent to spin in poll() on Linux.

I see no good reason why though.


shannonbooth:
See also https://github.com/LadybirdBrowser/ladybird/pull/3145/files but I believe we had to revert in https://github.com/LadybirdBrowser/ladybird/commit/be38acfad99ae9862cf6acdb935749cf7ab24575

ADKaster:
Yeesh so this really has been an issue since forever.

Maybe solving it will fix random other flakiness too 🤔

ADKaster:
@Hendiadyoin1 @spholz  the first 3 commits are good cherry-pick candidates. May help with your current  async-RS debugging.

trflynn89:
Follow-up for posterity: #5352

PR #5334, opened 2025-07-06
Remove unnecessary own clip rect application

kalenikaliaksandr:
...in PaintableWithLines and ImagePaintable.


tcl3 (approved):
I'm struggling to figure out why I added these in the first place :sweat_smile:

PR #5337, opened 2025-07-07
Meta: Add `--wpt-base-url` option to the WPT importer

tcl3:
This allows files to be downloaded locally by first running: `./Meta/WPT.sh serve` then importing with
`--wpt-base-url http://web-platform.test:8000`.

I intend to make this the default behavior once I figure out how to start and stop the WPT server in a nice way.


Meta/WPT.sh
```diff
@@ -620,13 +620,15 @@ import_wpt()
     pushd "${LADYBIRD_SOURCE_DIR}" > /dev/null
         ./Meta/ladybird.py build test-web
         set +e
+        trap 'exit 1' EXIT INT TERM
```
gmta, line 622: Why do we need that `set +e` there? If it's test-web, couldn't we `|| true` that if it's always supposed to continue?

tcl3: `set +e` is used is for the `import-wpt-test.py`  call as well as the `test-web` call. IMO avoiding `set +e` here, while possible, makes things more difficult to read.

gmta: But `import-wpt-test.py` is inside an `if`, that shouldn't trigger `set -e`?

tcl3: You're right, just adding  `|| true` to the `test-web` call works.

gmta (approved):

PR #5339, opened 2025-07-07
Meta+RequestServer: Enable HTTP/3 for curl

Lubrsi:
This adds an overlay port for curl that adds the features required for
HTTP/3.

This is not quite compatible with the upstream vcpkg.json, because
enabling HTTP/2 makes it use the default SSL backend, which is
sectransp for macOS and schannel on Windows. These backends are not
compatible with ngtcp2. Additionally, we can not build curl with
multiple SSL backends when using ngtcp2.

I couldn't find a way to selectively disable/enable dependencies based
on what features are enabled, so I made HTTP/2 pick OpenSSL in our
overlay port. Upstream vcpkg will likely want to support wolfSSL and
GnuTLS backends for ngtcp2, so they'll be additional work to get this
into upstream.

Screenshot 2025-07-07 at 16 22 04

Screenshot 2025-07-07 at 16 21 58

### DIFF
--- a/Meta/CMake/vcpkg/overlay-ports/curl/portfile.cmake
+++ b/Meta/CMake/vcpkg/overlay-ports/curl/portfile.cmake
@@ -14,6 +14,7 @@ vcpkg_from_github(
 vcpkg_check_features(OUT_FEATURE_OPTIONS FEATURE_OPTIONS
     FEATURES
         http2       USE_NGHTTP2
+        http3       USE_NGTCP2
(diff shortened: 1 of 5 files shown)


ADKaster (approved):
Looks reasonable. iirc upstream vcpkg is in a real pickle as far as http/3 support goes. They really don't like too many mutually exclusive features in ports. We'll likely have to keep the overlay around.

PR #5340, opened 2025-07-07
CI+CMake: Enable vcpkg source assets cache

gmta:
I've set up a shared vcpkg source assets cache using Azurite. By default, it runs in read-only mode allowing anyone who builds Ladybird to use the cache to speed up downloads of source tarballs. This should alleviate some of the more slower vcpkgs downloads I've been seeing lately.

CI runs on master put vcpkg in readwrite mode, updating the cache for everyone.

### DIFF
--- a/.github/workflows/lagom-template.yml
+++ b/.github/workflows/lagom-template.yml
@@ -126,7 +126,7 @@ jobs:
         run: sqlite3 $HOME/Library/Application\ Support/com.apple.TCC/TCC.db "INSERT OR IGNORE INTO access VALUES ('kTCCServiceMicrophone','/usr/local/opt/runner/provisioner/provisioner',1,2,4,1,NULL,NULL,0,'UNUSED',NULL,0,1687786159,NULL,NULL,'UNUSED',1687786159);"
 
       - name: Create Build Environment
-        if: ${{ inputs.os_name != 'Windows' && inputs.build_preset != 'Fuzzers_CI' }}
+        if: ${{ inputs.os_name != 'Windows' && inputs.build_preset != 'Fuzzers' }}
         working-directory: ${{ github.workspace }}
         env:
           VCPKG_CACHE_SAS: ${{ github.ref == 'refs/heads/master' && secrets.VCPKG_CACHE_SAS || '' }}
@@ -148,10 +148,10 @@ jobs:
           cmake --preset ${{ inputs.build_preset }} -B Build
 
       - name: Create Build Environment (Fuzzers)
-        if: ${{ inputs.build_preset == 'Fuzzers_CI' }}
+        if: ${{ inputs.build_preset == 'Fuzzers' }}
         working-directory: ${{ github.workspace }}
         run: |
-          cmake --preset=Distribution_CI -S Meta/Lagom -B ${{ github.workspace }}/Build/tools-build \
+          cmake --preset=Distribution -S Meta/Lagom -B ${{ github.workspace }}/Build/tools-build \
             -DLAGOM_TOOLS_ONLY=ON \
             -DINSTALL_LAGOM_TOOLS=ON \
             -DCMAKE_INSTALL_PREFIX=${{ github.workspace }}/Build/tools-install \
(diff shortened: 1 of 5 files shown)


.github/workflows/lagom-template.yml
```diff
@@ -113,6 +113,12 @@ jobs:
           download_cache_path: ${{ github.workspace }}/Build/caches
           vcpkg_cache_path: ${{ github.workspace }}/Build/caches/vcpkg-binary-cache
 
+      - name: Enable read-write vcpkg source assets cache
+        if: github.ref == 'refs/heads/master'
+        run: |
+          echo "VCPKG_CACHE_SAS=${{ secrets.VCPKG_CACHE_SAS }}" >> "$GITHUB_ENV"
```
ADKaster: I think conventional wisdom is to export secrets as close to time of use as possible. ... This is pretty close though. 

This step will export the secret into the environment for all subsequent steps, but we only "need" it in the env for the configure steps. 

So perhaps putting it in the env: dictionary for the individual confugure steps (all half dozen of them for some reason) would be 'safer'?

gmta: Hm, I did this since I only want to set those env vars for `master`. Can we do that in `env:` as well? 🤔

ADKaster: Uh. That's a good question. I don't think env: works that way

gmta: Hm, we might be able to do something like:

```
env:
  VCPKG_SAS:  ${{ github.ref == 'refs/heads/master' && secrets.VCPKG_SAS || '' }}
```

But that'd require us repeating that for the three "Create Build Environment" steps 🤔

gmta: Updated - there were only two steps, that's still manageable I'd say. As a bonus, I also found a way to unify our different build steps into one.

.github/workflows/lagom-template.yml
```diff
@@ -165,29 +167,11 @@ jobs:
...
 
-      - name: Build (Windows)
-        if: ${{ inputs.os_name == 'Windows' }}
-        working-directory: ${{ github.workspace }}
-        run: |
-          cmake --build --preset ${{ inputs.build_preset }}
-
-      - name: Enable the Ladybird AppKit chrome
-        if: ${{ inputs.os_name == 'macOS' && inputs.build_preset == 'Sanitizer_CI' }}
-        working-directory: ${{ github.workspace }}
```
ADKaster, line 183: These extra macOS steps are actually here to make sure that the Qt code on macOS doesn't bitrot. First build with qt, then with appkit, run tests with appkit.

If we remove the CI, we should remove the option to build with qt on macOS

gmta: Oh right, missed that. I'll reverse the logic - build AppKit by default, and prevent Qt bitrot in a separate step.

ADKaster: Just be careful that we don't end up testing the Qt chrome in CI. There's different event loop implementations in play. Would be tedious to get crashes and weird behavior locally that doesn't repro in CI (and vice versa) due to Qt vs CFEventLoop.

ADKaster: Er actually. We always use EventLoopImplementationUnix in test-web right? So that comment of mine might be old thinking in play. I don't think we actually test the UI apps in CI :thinking: And thanks to Tim's ev investigation we now always use the poll-based event loop in WebContent and WebWorker as well, *unless* we choose to use QtMultimedia for audio.

trflynn89: Right - the order doesn't matter much here because the CF event loop gets zero coverage in CI :muscle:

gmta: I've now set up a separate build step clearly marked as "macOS with Qt" so it's a bit clearer why we have that. The main build step just uses whatever the default for `ENABLE_QT` is on the build platform.

gmta:
Since the build presets were renamed, GitHub's required checks will not work for this PR.

PR #5348, opened 2025-07-08
LibWeb: Do not crash when navigating to `mailto:` links

gmta:
We forgot to implement a couple of "otherwise," statements from the "populating a session history entry" spec. While we're here, let's update the spec copy where relevant.

### DIFF
--- a/Tests/LibWeb/Text/input/navigation/navigate-mailto-crash.html
+++ b/Tests/LibWeb/Text/input/navigation/navigate-mailto-crash.html
@@ -1,0 +1,14 @@
+<!DOCTYPE html>
+<script src="../include.js"></script>
+<a href="mailto:contact@ladybird.org">please don't crash</a>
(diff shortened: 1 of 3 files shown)


Tests/LibWeb/Text/input/navigation/navigate-mailto-crash.html
```diff
@@ -0,0 +1,14 @@
...
+<script>
+asyncTest(done => {
+    document.querySelector('a').click();
+
+    // The crash happened asynchronously, so we wait for a small amount of time before passing the test.
+    setTimeout(() => {
+        println('PASS (did not crash)');
+        done();
+    }, 100);
+});
```
gmta, line 13: @tcl3 I've switched to a text test and used this timeout to wait for the async code to crash, which works every time on my machine. Let's see if CI agrees :)

PR #5350, opened 2025-07-08
Meta: Disable clang-tidy's const correctness checks

gmta:
Using `const` should not be warned about everywhere if it does not have a clear advantages. Compilers are able to deduce constness in most cases and on top of that, it's generally accepted that using `const` communicates developer intent above all else.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -32,6 +32,7 @@ Checks: >
   -bugprone-macro-parentheses,
   -bugprone-reserved-identifier,-cert-dcl37-c,-cert-dcl51-cpp,
   -cert-dcl21-cpp,
+  -misc-const-correctness,
   -misc-include-cleaner,
   -misc-no-recursion,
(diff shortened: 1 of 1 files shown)


.clang-tidy
```diff
@@ -32,6 +32,7 @@ Checks: >
   -bugprone-macro-parentheses,
   -bugprone-reserved-identifier,-cert-dcl37-c,-cert-dcl51-cpp,
   -cert-dcl21-cpp,
+  -misc-const-correctness,
```
trflynn89, line 35: Are you seeing this a lot? I usually see it only on methods I write that could be const that I forgot to make const, which I've actually found nice

gmta: This is about variables, not method signatures - so that functionality remains in place. See here:
https://clang.llvm.org/extra/clang-tidy/checks/misc/const-correctness.html

PR #5353, opened 2025-07-08
LibWeb: A couple of weeks' worth of HTML spec changes

AtkinsSJ:
A mix of "barely distinguishable editorial changes" and "fixes a bug and comes with a shiny test".

### DIFF
--- a/Libraries/LibWeb/HTML/CustomElements/CustomElementName.cpp
+++ b/Libraries/LibWeb/HTML/CustomElements/CustomElementName.cpp
@@ -1,42 +1,53 @@
 /*
  * Copyright (c) 2023, Srikavin Ramkumar <me@srikavin.me>
+ * Copyright (c) 2025, Sam Atkins <sam@ladybird.org>
  *
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
-#include <AK/StringView.h>
 #include <AK/Utf8View.h>
+#include <LibWeb/DOM/Element.h>
 #include <LibWeb/HTML/CustomElements/CustomElementName.h>
 
 namespace Web::HTML {
 
-// https://html.spec.whatwg.org/multipage/custom-elements.html#custom-elements-core-concepts:prod-pcenchar
-static bool is_pcen_char(u32 code_point)
-{
-    return code_point == '-'
-        || code_point == '.'
-        || (code_point >= '0' && code_point <= '9')
-        || code_point == '_'
-        || (code_point >= 'a' && code_point <= 'z')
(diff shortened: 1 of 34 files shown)


Libraries/LibWeb/HTML/CustomElements/CustomElementName.cpp
```diff
@@ -1,42 +1,53 @@
...
-        || (code_point >= 0x2c00 && code_point <= 0x2fef)
-        || (code_point >= 0x3001 && code_point <= 0xD7ff)
-        || (code_point >= 0xf900 && code_point <= 0xfdcf)
-        || (code_point >= 0xfdf0 && code_point <= 0xfffd)
-        || (code_point >= 0x10000 && code_point <= 0xeffff);
-}
-
 // https://html.spec.whatwg.org/multipage/custom-elements.html#valid-custom-element-name
-bool is_valid_custom_element_name(StringView name)
+bool is_valid_custom_element_name(String const& name)
```
tcl3, line 15: It might be worth importing `custom-elements/registries/valid-custom-element-names.html`, which passes after this change.

AtkinsSJ: Thanks! I've done so.

The few tests I had imported are because I saw their wpt commits on the PR pages.

Libraries/LibWeb/HTML/ElementInternals.cpp
```diff
@@ -108,15 +108,17 @@ WebIDL::ExceptionOr<void> ElementInternals::set_validity(ValidityStateFlags cons
...
-    // FIXME: 7. Set element's validation anchor to null if anchor is not given. Otherwise, if anchor is not a shadow-including descendant of element, then throw a "NotFoundError" DOMException. Otherwise, set element's validation anchor to anchor.
+    // 7. If anchor is not given, then set it to element.
     if (!anchor.has_value() || !anchor.value().ptr()) {
-        // FIXME
-    } else if (!anchor.value()->is_shadow_including_descendant_of(element)) {
+        anchor = element;
+    }
+
+    // 8. Otherwise, if anchor is not a shadow-including inclusive descendant of element, then throw a "NotFoundError" DOMException.
+    else if (!anchor.value()->is_shadow_including_descendant_of(element)) {
```
tcl3: Should this be `is_shadow_including_inclusive_descendant_of()` instead?

AtkinsSJ: Yes! And that gets us an extra pass. Thanks!

PR #5361, opened 2025-07-09
WebContent+UI: Don't open view source multiple times

### LINKED ISSUE #5220: View source of view source page
### Summary

When you view the source of a page, you can view the source of the view source page, and the source of the view source page; you can do this until the browser crashes

(Note: The browser was run in an Ubuntu virtual machine on Windows 11)

### Operating system

Linux

### Steps to reproduce

1. Open a page
2. View the source of the page
3. View the source of the view source page
4. Repeat step 3 until the browser crashes

### Expected behavior

You cannot view the source of a view source page

### Actual behavior

You can view the source of a view source page and repeat until the browser crashes

N/A

```html
N/A
```

```shell
firecrafter28@Ubuntu:~/dev$ cd ladybird/
firecrafter28@Ubuntu:~/dev/ladybird$ ./Meta/ladybird.py run
ninja: Entering directory `/home/firecrafter28/dev/ladybird/Build/release'
ninja: no work to do.
100.587 WebContent(2756): FIXME: Unimplemented Navigator.maxTouchPoints
100.704 WebContent(2756): ResourceLoader: Filtered request to: "https://clients1.google.com/generate_204"
100.757 WebContent(2756): ResourceLoader: Filtered request to: "https://www.google.com/gen_204?atyp=i&ct=psnt&cad=&nt=null&ei=eXpcaKa8MojXptQP6I2L6AE&zx=1750891130576&opi=89978449"
112.525 WebContent(2809): FIXME: Handle null navigation_request in navigation response Content Security Policy check.
121.984 WebContent(2756): ResourceLoader: Failed load of: "https://ssl.gstatic.com/gb/js/sem_d24f657a240731546d221f34a4e04946.js", Error: Unable to connect (status: 0), Duration: 21216ms
121.984 WebContent(2756): HTMLScriptElement: Refusing to run script because the element's result is null.
151.443 WebContent(2844): FIXME: Handle null navigation_request in navigation response Content Security Policy check.
175.705 WebContent(2917): FIXME: Handle null navigation_request in navigation response Content Security Policy check.
382.122 Ladybird(2733): WebContent process crashed! Last page loaded: about:srcdoc
382.126 Ladybird(2733): Consider raising an issue at https://github.com/LadybirdBrowser/ladybird/issues/new/choose
383.143 WebContent(2943): FIXME: Handle null navigation_request in navigation response Content Security Policy check.
```

_No response_

_No response_

### Contribute a patch?

ChaseKnowlden:
Fixes the issue where you open view source multiple times in a loop, either the computer freezes, or the computer runs out of memory.

Fixes #5220


rcorsi:
Definitely I like disabling the menu option better based on a flag set when the view source is requested.  I actually tried it briefly yesterday and noted that we need to re-enable the menu if we have other tabs that are not showing source, and also if the current tab gets reloaded with some other content the flag needs to be cleared.

rcorsi:
Is there other ways, other than via menus, to trigger the view source?  From JS maybe?

AtkinsSJ:
I had a thought last night: If we used the `view-source:` protocol like other browsers do, then we could use that to determine if we should disable the context menu option, without having to set any other state. It'd also give us the nice feature of being able to type `view-source:https://example.com` in the URL bar directly. However, it'd require a more significant change to how we implement source viewing.

(I've noticed while testing that FF and Chrome don't let you link to a view-source: URL, they do nothing with it, which is interesting.)

rcorsi:
It seems possible to avoid checking the title of the tab.

Here is my attempt (it might have flaws and might need improvement), but I don't have Macos AppKit implementation only Qt.  

https://github.com/LadybirdBrowser/ladybird/compare/master...rcorsi:ladybird:block_view_source

Please steal the ideas and work them into your commit if you can.

rcorsi:
I just built Ladybird on MacOS for the first time.  Is it normal that when using view source, nothing appears in the view source tab?

I can see the memory usage going up a lot as had been reported when recursively doing view source.

Am I just not waiting long enough for the HTML to be rendered?

ChaseKnowlden:
> I just built Ladybird on MacOS for the first time. Is it normal that when using view source, nothing appears in the view source tab?
> 
> I can see the memory usage going up a lot as had been reported when recursively doing view source.
> 
> Am I just not waiting long enough for the HTML to be rendered?

Sometimes happens to me.

ChaseKnowlden:
Fixed the Qt Port.

rcorsi:
I tried the changes on my Mac (one of the last Intel MacBook Pro's) and the changes don't work for me. I can see the large memory use when repeatedly viewing source, but strangely I can't view the source.

rcorsi:
the "on_activate_tab" WebViewCallback doesn't seem to get triggered.

UI/AppKit/Interface/LadybirdWebView.mm
```diff
@@ -1324,9 +1333,10 @@ - (NSMenu*)page_context_menu
                                                         keyEquivalent:@""]];
         [_page_context_menu addItem:[NSMenuItem separatorItem]];
 
-        [_page_context_menu addItem:[[NSMenuItem alloc] initWithTitle:@"View Source"
-                                                               action:@selector(viewSource:)
-                                                        keyEquivalent:@""]];
+        auto* view_source_menu_item = [[NSMenuItem alloc] initWithTitle:@"View Source"
+                                                                 action:@selector(viewSource:)
+                                                          keyEquivalent:@""];
+        [view_source_menu_item setTag:CONTEXT_MENU_VIEW_SOURCE_TAG];
```
rcorsi, line 1339: after line 1339, add line with `[_page_context_menu addItem:view_source_menu_item];` which was removed.

rcorsi: Also in `UI/AppKit/Application/ApplicationDelegate.mm` there is a createInspectMenu function there that needs some handling for the "View Source" found in the Inspect sub-menu.

Sorry, maybe you are already working on this. I wasn't sure if you were done or not.

ChaseKnowlden: I'm not done.

ChaseKnowlden: I fixed the context menu option.

ChaseKnowlden: I'm done!

PR #5362, opened 2025-07-09
LibWeb: Cache compiled shaders for masks in DisplayListPlayerSkia

kalenikaliaksandr:
By avoiding recompilation every time `apply_mask_bitmap()` is called, we save ~5 ms (20ms -> 15ms) in rendering of browser channel on Discord on my machine.


Libraries/LibWeb/Painting/DisplayListPlayerSkia.cpp
```diff
@@ -1060,40 +1064,65 @@ void DisplayListPlayerSkia::apply_transform(ApplyTransform const& command)
...
+}
+
 void DisplayListPlayerSkia::apply_mask_bitmap(ApplyMaskBitmap const& command)
 {
     auto& canvas = surface().canvas();
 
     auto const* mask_image = command.bitmap->sk_image();
 
-    char const* sksl_shader = nullptr;
+    auto compile_effect = [&](char const* sksl_shader) {
```
gmta: You don't need `[&]` here, it all looks pretty static :)

PR #5366, opened 2025-07-09
LibWeb: Improve support for logical border shorthands

Calme1709:

Gains us 18 WPT tests.


Libraries/LibWeb/CSS/Properties.json
```diff
@@ -579,6 +587,15 @@
...
       "border-block-end-width"
+    ],
+    "max-values": 2,
+    "valid-types": [
+      "length [0,∞]"
+    ],
+    "valid-identifiers": [
+      "medium",
+      "thick",
+      "thin"
```
AtkinsSJ, line 607: You don't have to make the change here, (I see we already had this for the border-width properties,) but we really should define an enum for these instead of repeating them so many times.

PR #5370, opened 2025-07-09
LibWeb/CSS: Construct all CSS Tokens in a consistent way

AtkinsSJ:
This is prep for a PR where I'll need to construct various Tokens to test with. No observed (or intended) behaviour changes.


tcl3 (approved):
I assume the original at-rule case sensitivity change made a difference at the time :sweat_smile:

AtkinsSJ:
I'm sure it did!

PR #5373, opened 2025-07-09
LibWeb: Don't fail on non-fill keyword in <border-image-slice>

Calme1709:
Previously if we encountered a keyword other than `fill` when parsing `<border-image-slice` we would return a nullptr.

This could cause issues when we parse `<border-image-slice>` as part of parsing `border-image`, for example `border-image: 100% none` would fail as we would try parse `none` as part of the `<border-image-slice>` instead of `<border-image-source>`.

This change makes it so that we don't consume the token and leave it to be parsed as part of the next section of the grammar.

PR also includes an unrelated commit to remove some unused code in `StyleComputer::for_each_property_expanding_shorthands`


AtkinsSJ (approved):
:+1:

Calme1709:
Two approvals before CI has even finished! Very efficient today!

PR #5375, opened 2025-07-09
LibWeb: Don't crash when parsing ASF with non comma-separated arguments

tcl3:
Fixes a crash in: http://wpt.live/css/css-values/attr-all-types.html


AtkinsSJ (approved):
Thanks! And now you're fixing my bugs. 😆

PR #5379, opened 2025-07-09
LibWeb/SVG: Don't crash when a filter has no valid effects

tcl3:
Fixes 57 WPT crashes in the `css/filter-effects/` directory.


ananas-dev:
Nice catch 👍🏻

gmta (approved):

PR #5384, opened 2025-07-10
LibWeb: Implement the unreachable scrollable overflow

gmta:
Whenever we end up with a scrollable overflow rect that goes beyond either of its axes (i.e. the rect has a negative X or Y position), we need to clip that rect to prevent going into the "unreachable scrollable overflow".

This fixes the horizontal scrolling on https://ladybird.org (gets more pronounced if you make the window very narrow).


Tests/LibWeb/Text/expected/wpt-import/css/css-flexbox/negative-overflow.txt
```diff
@@ -2,16 +2,16 @@ Harness status: OK
...
-Fail	.flexbox 2
-Fail	.flexbox 3
+7 Pass
+4 Fail
+Pass	.flexbox 1
+Pass	.flexbox 2
+Pass	.flexbox 3
 Pass	.flexbox 4
-Pass	.flexbox 5
+Fail	.flexbox 5
```
gmta, line 11: Note: this is now failing because of the new clipping code not supporting `flex-wrap: wrap-reverse`, which moves the scroll origin to the bottom of the box. There's a `FIXME` for this.

PR #5386, opened 2025-07-10
LibWeb: Generically serialize "positional-value-list-shorthand"s

Calme1709:
Introduces the concept of "positional-value-list-shorthand"s and adds generic functionality for serializing them.

I am not sold on the "positional-value-list-shorthand" name and would be eager to hear alternatives, as far as I can tell there isn't a spec term for this concept, but MDN refers to the syntax it uses as the "1-to-x-value syntax" [here](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_cascade/Shorthand_properties#edges_of_a_box).

There is still more work to be done around this concept, specifically around parsing directly to `ShorthandStyleValue`s which I have added FIXMEs for.

Gains us 10 new WPT passes.


AtkinsSJ:
Drive-by comment: We called this a "sided shorthand" in some of the code you removed, (maybe elsewhere too?) and I kinda like that name.

Calme1709:
I did see that but it doesn't quite cover all the cases, e.g. it works for `margin` but not really for `gap` or `margin-inline`.

PR #5390, opened 2025-07-10
LibWeb: Add `internals` call to dump display list

kalenikaliaksandr:
It's useful to have tests that dump display list items, so we can more easily see how changes to the display list recording process affect the output. Even the small sample test added in this commit shows that we currently record an unnecessary AddClipRect item for empty paint phases.

For now, the dump doesn't include every single property of an item, but we can shape it to include more useful information as we iterate on it.


Libraries/LibWeb/Painting/DisplayList.cpp
```diff
@@ -13,6 +13,17 @@ void DisplayList::append(Command&& command, Optional<i32> scroll_frame_id)
     m_commands.append({ scroll_frame_id, move(command) });
 }
 
+String DisplayList::dump() const
+{
+    StringBuilder builder;
+    for (size_t command_index = 0; command_index < m_commands.size(); ++command_index) {
+        auto const& command = m_commands[command_index];
```
gmta: This can be a range-based for loop instead

Libraries/LibWeb/DOM/Document.h
```diff
@@ -112,7 +112,8 @@ enum class InvalidateLayoutTreeReason {
     X(SVGDecodedImageDataRender)           \
     X(SVGGraphicsElementGetBBox)           \
     X(SourceSetNormalizeSourceDensities)   \
-    X(WindowScroll)
+    X(WindowScroll)                        \
+    X(DumpDisplayList)
```
gmta: Let's keep this alphabetical

UI/AppKit/Application/ApplicationDelegate.mm
```diff
@@ -628,6 +628,9 @@ - (NSMenuItem*)createDebugMenu
     auto* menu = [[NSMenuItem alloc] init];
     auto* submenu = [[NSMenu alloc] initWithTitle:@"Debug"];
 
+    [submenu addItem:[[NSMenuItem alloc] initWithTitle:@"Dump Display List"
+                                                action:@selector(dumpDisplayList:)
+                                         keyEquivalent:@""]];
```
gmta: The order of the dump operations was from more abstract to more concrete, i.e. DOM > Layout > Paint etc., I think Dump Display List should go last instead of first here.

UI/Qt/BrowserWindow.cpp
```diff
@@ -395,6 +395,13 @@ BrowserWindow::BrowserWindow(Vector<URL::URL> const& initial_urls, IsPopupWindow
         debug_request("dump-paint-tree");
     });
 
+    auto* dump_display_list = new QAction("Dump Display List", this);
+    dump_display_list->setIcon(load_icon_from_uri("resource://icons/16x16/layout.png"sv));
+    debug_menu->addAction(dump_display_list);
+    QObject::connect(dump_display_list, &QAction::triggered, this, [this] {
+        debug_request("dump-display-list");
+    });
```
gmta: This is now one position higher than the AppKit version 😅

PR #5391, opened 2025-07-10
Utilites: Remove extra dashes from raw strings argument to js

ADKaster:
While hilarious, we don't need the long argument name as `----raw-strings`. cc @alimpfard


CanadaHonk:
great bit

image

PR #5398, opened 2025-07-11
LibWeb/Platform: Replace Serenity with Ladybird in class names

ayeteadoe:


circl-lastname:
`Serenity` -> `LibCore` would be more intuitive IMO, that's what provides the underlying implementation, like Qt in the other plugin

ayeteadoe:
Yes I agree "Ladybird" isn't ideal, even something like "Impl" probably makes more sense.

Will wait on feedback from maintainers though.

PR #5400, opened 2025-07-11
LibWeb/CSS: Parse the `<syntax>` type, and then use it to support `attr(foo type())`

AtkinsSJ:
`<syntax>` is a limited subset of the general "value definition syntax" used in CSS specs. Supporting it is necessary for `@property` and `attr(foo type())`, but I'm also hoping to use an extended version for parsing properties and descriptors in the future.

`attr(foo type())` was the most convenient test-case of the two. It lets the attribute's value be interpreted as an arbitrary CSS type or custom keyword, instead of just as a string.

The `<syntax>` is parsed into a tree of `SyntaxNode`s, which are then applied to a list of CVs that we want to parse a value from.

The jankiest part of all this is needing to take our newly-parsed StyleValue and turn it back into a list of ComponentValues. I've implemented this "properly" for a few of the simpler types, but left in a "turn it into a string and then parse that as CVs" method for everything else. It's not the most elegant or performant solution, but it's one that is guaranteed to work right as long as we serialize things properly. We can write proper implementations for these if and when they become a performance bottleneck.


Libraries/LibWeb/CSS/Parser/ArbitrarySubstitutionFunctions.cpp
```diff
@@ -145,29 +159,35 @@ static Vector<ComponentValue> replace_an_attr_function(DOM::AbstractElement& ele
...
         return failure();
 
-    if (component_value.is(Token::Type::Number) && is_dimension_unit)
-        return { Token::create_dimension(component_value.token().number_value(), syntax) };
+    if (unit_name.has_value()) {
+        // https://drafts.csswg.org/css-values-5/#ref-for-typedef-attr-type%E2%91%A0
+        // If given as an <attr-unit> value, the value is first parsed as if type(<number>) was specified, then the
+        // resulting numeric value is turned into a dimension with the corresponding unit, or a percentage if % was
+        // given. Values that fail to parse as a <number> trigger fallback.
+        VERIFY(parsed_value->is_number());
```
tcl3: `parsed_value` isn't guaranteed to be a number at this point. This causes a crash in https://wpt.live/css/css-values/crashtests/chrome-405422528-crash.html, where `parsed_value` is a calculated value.

AtkinsSJ: Thanks, I'll pull some more tests in and have a more thorough look.

AtkinsSJ:
Imported more tests, and then fixed the crash. (Thanks Tim!) Filed a spec bug about the ambiguity of non-literal `<number>`s should be treated with `<attr-unit>`: https://github.com/w3c/csswg-drafts/issues/12479

AtkinsSJ:
(Second push just to rebase.)

AtkinsSJ:
Changes: Rebased to fix conflicts. (Got an extra subtest pass out of it somehow!)

tcl3 (approved):

PR #5406, opened 2025-07-11
LibWeb/SVG: Prefer `href` to `xlink:href` attribute on images

tcl3:
This stops these WPT tests from crashing: 
* http://wpt.live/svg/embedded/image-remove-href-2.svg
 * http://wpt.live/svg/embedded/image-modify-href-2.svg
 * https://wpt.live/svg/linking/reftests/href-image-element.html  - this one was flaky on WPT but crashed every time in the browser.
 
Unfortunately, they can't be imported since they still fail for other reasons :sweat_smile:


Libraries/LibWeb/SVG/SVGImageElement.cpp
```diff
@@ -60,7 +61,19 @@ void SVGImageElement::attribute_changed(FlyString const& name, Optional<String>
...
+        // For backwards compatibility, elements with an ‘href’ attribute also recognize an ‘href’ attribute in the
+        // XLink namespace. If the element is in the XLink namespace, it does not recognize an ‘href’ attribute in the
+        // SVG namespace. When the ‘href’ attribute is present in both the XLink namespace and without a namespace, the
+        // value of the attribute without a namespace shall be used. The attribute in the XLink namespace shall be ignored.
+        if (namespace_ == Namespace::XLink && has_attribute_ns({}, name))
+            return;
+
+        auto href = value;
+        if (!namespace_.has_value() && !href.has_value() && has_attribute_ns(Namespace::XLink, SVG::AttributeNames::href))
+            href = get_attribute_value(SVG::AttributeNames::href, Namespace::XLink);
```
gmta: I think you can use `get_attribute_ns` here, which returns `Optional<String>`. Then you can remove `has_attribute_ns(..)` from the condition.

gmta (approved):

PR #5409, opened 2025-07-11
Meta: Clean up flatpak manifest and remove most linter warnings

ADKaster:
We now clean up installed helper tools, includes from dependencies, and pkgconfig/CMake files. This decreases the size of the flatpak from ~206MiB to ~150 MiB on my machine.

The manifest is also mostly clean of linter warnings from the flatpack-builder manifest linter, with the exception of the overly broad session bus policy.

The docs at https://docs.flathub.org/docs/for-app-authors/linter list a method for selecting the correct session bus policies, but it is unclear how to actually get the full set.

### DIFF
--- a/Meta/CMake/flatpak/org.ladybird.Ladybird.json
+++ b/Meta/CMake/flatpak/org.ladybird.Ladybird.json
@@ -13,6 +13,17 @@
     "--socket=pulseaudio",
     "--socket=session-bus"
   ],
+  "cleanup": [
+    "/sbin",
+    "*.a",
+    "/lib/cmake",
+    "/lib/pkgconfig",
+    "/include",
+    "/share/aclocal",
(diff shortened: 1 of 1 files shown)


Meta/CMake/flatpak/org.ladybird.Ladybird.json
```diff
@@ -476,6 +506,8 @@
         "-DCMAKE_BUILD_TYPE=Release",
         "-DENABLE_LTO_FOR_RELEASE=OFF",
         "-DENABLE_INSTALL_FREEDESKTOP_FILES=ON",
+        "-DENABLE_INSTALL_HEADERS=OFF",
+        "-DBUILD_TESTING=OFF  ",
```
gmta: Some stray spaces there

PR #5410, opened 2025-07-11
LibWeb: Implement the `SVGGeometryElement.pathLength` attribute

tcl3:
This gets us +11 WPT passes.


Libraries/LibWeb/SVG/SVGGeometryElement.cpp
```diff
@@ -38,4 +44,11 @@ GC::Ref<Geometry::DOMPoint> SVGGeometryElement::get_point_at_length(float distan
     return Geometry::DOMPoint::construct_impl(realm(), 0, 0, 0, 0);
 }
 
+GC::Ptr<SVGAnimatedNumber> SVGGeometryElement::path_length()
```
shannonbooth: nit but this can return a GC::Ref

PR #5412, opened 2025-07-12
LibWeb: Add `target` and `referrerPolicy` to `SVGAElement`

tcl3:
This PR adds the `target` and `referrerPolicy` attributes to `SVGAElement`. These properties reflect their respective content attributes.

Side note: As part of this PR, I had intended to implement the `HTMLHyperlinkElementUtils` interface on `SVGAElement`, but I ran into a problem with the `href` attribute because it is defined in both the `HTMLHyperlinkElementUtils` and `SVGURIReference` mixins. Other browsers seem to cleverly get around this by just not bothering to include  `HTMLHyperlinkElementUtils` on `SVGAElement`.

Gets us 8 WPT passes.


Libraries/LibWeb/SVG/SVGAElement.cpp
```diff
@@ -61,6 +61,14 @@ i32 SVGAElement::default_tab_index_value() const
     return 0;
```
shannonbooth, line 62: typo in commit message of "Implememnt"

gmta (approved):

PR #5414, opened 2025-07-12
LibWeb: Parse the ::slotted pseudo-element

shannonbooth:
Dipping my toes into some CSS, in draft while I try to understand a bit better some of the test failures for parsing.

### DIFF
--- a/Libraries/LibWeb/CSS/Parser/SelectorParsing.cpp
+++ b/Libraries/LibWeb/CSS/Parser/SelectorParsing.cpp
@@ -466,6 +466,20 @@ Parser::ParseErrorOr<Selector::SimpleSelector> Parser::parse_pseudo_simple_selec
                         return ParseError::SyntaxError;
                     }
                     break;
+                case PseudoElementMetadata::ParameterType::CompoundSelector: {
+                    auto compound_selector_or_error = parse_compound_selector(function_tokens);
+                    if (compound_selector_or_error.is_error() || !compound_selector_or_error.value().has_value()) {
+                        dbgln_if(CSS_PARSER_DEBUG, "Failed to parse :{}() parameter as a compound selector", pseudo_name);
+                        return ParseError::SyntaxError;
+                    }
+
+                    auto compound_selector = compound_selector_or_error.release_value().release_value();
+                    compound_selector.combinator = Selector::Combinator::None;
+
+                    Vector compound_selectors { move(compound_selector) };
+                    value = Selector::create(move(compound_selectors));
+                    break;
+                }
                 case PseudoElementMetadata::ParameterType::PTNameSelector: {
                     // <pt-name-selector> = '*' | <custom-ident>
                     // https://drafts.csswg.org/css-view-transitions-1/#typedef-pt-name-selector
--- a/Libraries/LibWeb/CSS/Selector.h
+++ b/Libraries/LibWeb/CSS/Selector.h
@@ -31,7 +31,7 @@ public:
             FlyString value {};
         };
 
-        using Value = Variant<Empty, PTNameSelector>;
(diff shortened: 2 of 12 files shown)


AtkinsSJ:
I wouldn't worry too much about those failures. They all seem related to what's allowed after a pseudo-element, which is something that's not limited to slotted. 😅

Libraries/LibWeb/CSS/Parser/SelectorParsing.cpp
```diff
@@ -466,6 +466,20 @@ Parser::ParseErrorOr<Selector::SimpleSelector> Parser::parse_pseudo_simple_selec
                         return ParseError::SyntaxError;
                     }
                     break;
+                case PseudoElementMetadata::ParameterType::CompoundSelector: {
+                    auto compound_selector_or_error = parse_compound_selector(function_tokens);
+                    if (compound_selector_or_error.is_error() || !compound_selector_or_error.value().has_value()) {
+                        dbgln_if(CSS_PARSER_DEBUG, "Failed to parse :{}() parameter as a compound selector", pseudo_name);
```
AtkinsSJ: Nit: Should be `::` for a pseudo-element, not `:`. I see that same mistake is in the PTNameSelector branch though.

Libraries/LibWeb/CSS/Selector.h
```diff
@@ -59,11 +59,14 @@ class Selector : public RefCounted<Selector> {
         PseudoElement type() const { return m_type; }
 
         PTNameSelector const& pt_name_selector() const { return m_value.get<PTNameSelector>(); }
+        // FIXME: Ideally this would be a CompoundSelector, but SimpleSelector requires PseudoElementSelector,
+        //        and CompoundSelector requires SimpleSelector.
```
AtkinsSJ: It's a bit unfortunate... but then, I think we'll probably need a full `Selector` to actually do the matching, so storing it as one is fine.

Libraries/LibWeb/Dump.cpp
```diff
@@ -613,6 +613,7 @@ void dump_selector(StringBuilder& builder, CSS::Selector const& selector, int in
 
                 switch (pseudo_element_metadata.parameter_type) {
                 case CSS::PseudoElementMetadata::ParameterType::None:
+                case CSS::PseudoElementMetadata::ParameterType::CompoundSelector:
```
AtkinsSJ, line 616: Dump the selector please!

shannonbooth: I think I've completely misunderstood, but don't we dump the serialization of the pseudo element above which includes it? Or is this just an extra level of detail on top of it?

shannonbooth:
> Generally looks good. Is there anything that confused you while working on this, that could be documented/implemented more clearly?

No it was all pretty straight forward, though the bit I am looking at now / trying to figure out now is where to be touching to actually implement this thing! (pointers appreciaited :P)

Libraries/LibWeb/Dump.cpp
AtkinsSJ: Oh you're absolutely correct! Makes me wonder why PTName does something special below, but this is fine.

AtkinsSJ:
Hmm, I think that's going to be a bit involved.

Currently, we use `CSS::PseudoElement` to identify which pseudo-element something is. But that's an enum and so doesn't support ones like `::slotted()` which have a parameter. You'd need to swap that for some kind of struct/class which holds the parameters too.  (And probably rename the enum to `PseudoElementType` or something.)

The match code in `SelectorEngine.cpp` would need some special logic for matching `::slotted()` because it needs to cross the shadow boundary, and because they're backed by real elements, but that's probably not too hard in comparison.

PR #5418, opened 2025-07-12
LibWeb: Implement `SVGViewElement`

tcl3:
This identifies a particular region of an SVG image, which can then be linked to by an SVG fragment identifier.

Currently, we only support the shorthand bare name form of fragment identifier.

Gets us 13 WPT subtest passes in the `svg` directory.


Libraries/LibWeb/SVG/SVGSVGElement.cpp
```diff
@@ -309,9 +314,14 @@ SVGSVGElement::NaturalMetrics SVGSVGElement::negotiate_natural_metrics(SVG::SVGS
...
-        // FIXME: 2. If an SVG View is active:
-        // FIXME:    1. let viewbox be the viewbox defined by the active SVG View
-        // FIXME:    2. return viewbox.width / viewbox.height
+        // 2. If an SVG View is active:
+        if (auto active_view_element = svg_root.active_view_element(); active_view_element && active_view_element->view_box().has_value()) {
+            // 1. let viewbox be the viewbox defined by the active SVG View
+            auto view_box = active_view_element->view_box().value();
+
+            // 2. return viewbox.width / viewbox.height
+            return view_box.width / view_box.height;
```
konradekk: nit: it'd probably advisable to take some of the logic down below to avoid possible division by zero.

tcl3: I've used the same logic as below for dealing with a division by zero here.

tcl3:
My latest push ensures that the view respected when visiting the view's fragment URL directly.

PR #5421, opened 2025-07-13
LibWeb: Avoid division by zero with small aspect ratios

tcl3:
This fixes a crash in http://wpt.live/css/css-sizing/aspect-ratio/small-aspect-ratio-crash.html


Libraries/LibWeb/PixelUnits.h
```diff
@@ -405,6 +405,8 @@ constexpr CSSPixels CSSPixels::operator/(CSSPixelFraction const& other) const
 {
     i64 wide_value = raw_value();
     wide_value *= other.denominator().raw_value();
+    if (other.numerator().raw_value() == 0)
```
kalenikaliaksandr: I would rather make caller perform this check. making `operator/` silence division by zero errors will make it harder to find other bugs which are currently discoverable by checking if value is saturated.

kalenikaliaksandr (changes requested):

Libraries/LibWeb/PixelUnits.h
tcl3: My latest push protects against division by zero at the offending call sites.

This is the change I intended to push originally, but I ended up pushing a branch I was using for testing :sweat_smile:

PR #5422, opened 2025-07-13
LibWeb: Avoid accessing opaque origin port during CSP checks

tcl3:
This fixes 2 CSP WPT tests, which previously crashed.


Libraries/LibWeb/ContentSecurityPolicy/Directives/DirectiveOperations.cpp
```diff
@@ -541,7 +541,7 @@ MatchResult does_url_match_expression_in_origin_with_redirect_count(URL::URL con
             origin_port = origin.port();
```
gmta, line 541: Typo in commit title: "acessing"

gmta (approved):

PR #5287, opened 2025-07-03
LibWeb: Generate logical property mappings

AtkinsSJ:
To support this, how we declare logical property aliases has changed. Instead of `logical-alias-for` being a list of properties, it's now an object with a `group` and `mapping`. The group is the name of a logical property group in LogicalPropertyGroups.json. The mapping is which side/dimension/corner this property is. Hopefully it's self-explanatory enough.

The generated code is very much a copy of what was previously in `StyleComputer::map_logical_alias_to_physical_property_id()`, so there should be no behaviour change.

PR #5424, opened 2025-07-13
LibWeb: Respect logical property groups

Calme1709:
Builds on the work @AtkinsSJ did in #5287 around defining logical property groups to tick off a couple of FIXMEs we have around that concept.

Gains us a grand total of 1 new WPT pass.

### DIFF
--- a/Libraries/LibWeb/CSS/CSSStyleProperties.cpp
+++ b/Libraries/LibWeb/CSS/CSSStyleProperties.cpp
@@ -1225,10 +1225,39 @@ String CSSStyleProperties::serialized() const
                 if (!all_declarations_have_same_important_flag)
                     continue;
 
-                // FIXME: 6. If there is any declaration in declaration block in between the first and the last longhand
-                //           in current longhands which belongs to the same logical property group, but has a different
-                //           mapping logic as any of the longhands in current longhands, and is not in current
-                //           longhands, continue with the steps labeled shorthand loop.
+                // 6. If there is any declaration in declaration block in between the first and the last longhand
+                //    in current longhands which belongs to the same logical property group, but has a different
+                //    mapping logic as any of the longhands in current longhands, and is not in current
+                //    longhands, continue with the steps labeled shorthand loop.
+                auto first_current_longhand_index = m_properties.find_first_index_if([&](StyleProperty const& current_declaration) { return current_declaration.property_id == current_longhands[0].property_id; });
+                auto last_current_longhand_index = m_properties.find_first_index_if([&](StyleProperty const& current_declaration) { return current_declaration.property_id == current_longhands[current_longhands.size() - 1].property_id; });
+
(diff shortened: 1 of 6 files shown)


Meta/Lagom/Tools/CodeGenerators/LibWeb/GenerateCSSPropertyID.cpp
```diff
@@ -1606,6 +1622,47 @@ PropertyID map_logical_alias_to_physical_property(PropertyID property_id, Logica
...
     generator.append(R"~~~(
+Optional<LogicalPropertyGroup> logical_property_group_for_property(PropertyID property_id)
+{
+    switch(property_id) {
+)~~~");
+
+    HashMap<String, Vector<String>> logical_property_group_members;
+
+    logical_property_groups.for_each_member([&](auto& logical_property_group_name, auto& mapping) {
+        logical_property_group_members.ensure(logical_property_group_name);
```
AtkinsSJ: `ensure()` returns the existing/added value, so you could use that instead of calling `get()` again repeatedly in the loop below.

Libraries/LibWeb/CSS/CSSStyleProperties.cpp
```diff
@@ -1225,10 +1225,39 @@ String CSSStyleProperties::serialized() const
...
+                auto first_current_longhand_index = m_properties.find_first_index_if([&](StyleProperty const& current_declaration) { return current_declaration.property_id == current_longhands[0].property_id; });
+                auto last_current_longhand_index = m_properties.find_first_index_if([&](StyleProperty const& current_declaration) { return current_declaration.property_id == current_longhands[current_longhands.size() - 1].property_id; });
+
+                VERIFY(first_current_longhand_index.has_value());
+                VERIFY(last_current_longhand_index.has_value());
+
+                bool should_continue = false;
+
+                for (auto current_declaration_index = first_current_longhand_index.value(); current_declaration_index <= last_current_longhand_index.value(); ++current_declaration_index) {
+                    // Declaration is in current longhands
```
AtkinsSJ: Nit: When we have spec comments, it's good to mark any non-spec ones.

AtkinsSJ (changes requested):
Thanks! It's only 1 pass but it's an important one to get right. :^)

Libraries/LibWeb/CSS/CSSStyleProperties.cpp
Calme1709: Fixed.

Meta/Lagom/Tools/CodeGenerators/LibWeb/GenerateCSSPropertyID.cpp
Calme1709: Done.

PR #5428, opened 2025-07-13
LibWeb/WebAudio: Add renderQuantumSize

Prospero23:
Adds this attribute to BaseAudioContext.idl and adds associated test for OfflineAudioContext. This gives a set value that can be used to start implementing OfflineAudioContext::start_rendering().


Tests/LibWeb/Text/input/WebAudio/offline-audiocontext-renderquantum.html
```diff
@@ -0,0 +1,8 @@
+<!DOCTYPE html>
+<script src="../include.js"></script>
+<script>
+    test(() => {
+        const audioContext = new OfflineAudioContext(1, 1, 44100)
+        println(`${audioContext.renderQuantumSize}`);
```
gmta: Instead of having a single test for a single attribute, maybe you could test the other OfflineAudioContext attributes here as well? Makes it a bit more useful going forward.

Libraries/LibWeb/WebAudio/BaseAudioContext.h
```diff
@@ -95,6 +96,7 @@ class BaseAudioContext : public DOM::EventTarget {
 
     float m_sample_rate { 0 };
     double m_current_time { 0 };
+    WebIDL::UnsignedLong m_render_quantum_size { 128 };
```
gmta: Reading the spec, I get the idea that this is a pretty fixed value that isn't going to change at runtime anytime soon. Shouldn't we make this a `static constexpr` instead?

Tests/LibWeb/Text/input/WebAudio/offline-audiocontext-renderquantum.html
Prospero23: The ctor-offlineaudiocontext WPT covers the other attributes and is included in the test suite. Since this part of the spec is only in the working draft, I assume submitting a PR for the WPT is off the table? 

There would be multiple tests in this file once support for renderSizeHint is added.

gmta: Right, that makes sense then. You could add a comment to the test explaining how it could probably be removed as soon as we can import the updated WPT ctor test.

Libraries/LibWeb/WebAudio/BaseAudioContext.h
Prospero23: That makes sense. Fixed.

Tests/LibWeb/Text/input/WebAudio/offline-audiocontext-renderquantum.html
Prospero23: Done.

Libraries/LibWeb/WebAudio/BaseAudioContext.h
```diff
@@ -95,6 +96,7 @@ class BaseAudioContext : public DOM::EventTarget {
 
     float m_sample_rate { 0 };
     double m_current_time { 0 };
+    static constexpr WebIDL::UnsignedLong m_render_quantum_size { 128 };
```
gmta: Please move this up to the top (near `private:`) and use either `s_render_quantum_size` or `RENDER_QUANTUM_SIZE` as a name, since we use `m_` for private members.

Prospero23: Done. I think I moved it to the correct position now.

Prospero23:
After looking at it again today, I think that naming the test file OfflineAudioContext is better. There are some other parts of this class that are not covered in the WPT that should be tested, and I can just use that file to add those as well.

gmta (approved):

PR #5429, opened 2025-07-13
LibWeb: Don't crash when appending to an XML document template element

tcl3:
When the XML parser appends child nodes to a template element, it must actually append the template element's contents. This special behavior caused us to return to the wrong parent element after adding child nodes to a template element, leading to a crash.

Fixes the imported WPT test, which previously crashed.


Libraries/LibWeb/XML/XMLDocumentBuilder.h
```diff
@@ -41,6 +41,7 @@ class XMLDocumentBuilder final : public XML::Listener {
     Optional<FlyString> namespace_for_name(XML::Name const&);
 
     GC::Ref<DOM::Document> m_document;
+    GC::Ptr<DOM::Node> m_parent_node;
```
shannonbooth: do we need to visit this?

tcl3: Because nothing else in this file is visited, my assumption was that this object can't live long enough for GC to be an issue. That probably isn't true though.

shannonbooth: Hmm, actually, maybe it was 'safe' before since XMLDocumentBuilder appears to be stored on the stack: https://github.com/LadybirdBrowser/ladybird/blob/a3cacbfaaa9a47a29e35f8101f498731def5a6d9/Libraries/LibWeb/DOM/DocumentLoading.cpp#L58

But now that we have GC members in a Vector it is maybe not safe?

tcl3: I've made this a `RootVector` instead.

gmta (approved):

PR #5430, opened 2025-07-13
Initial implementation of the Trusted types Proposal

tete17:


Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.idl
```diff
@@ -0,0 +1,19 @@
+[Exposed=(Window,Worker)] interface TrustedTypePolicyFactory {
```
tcl3: A spec link should be added here: https://w3c.github.io/trusted-types/dist/spec/#trustedtypepolicyfactory

Libraries/LibWeb/SVG/SVGAnimationElement.idl
```diff
@@ -0,0 +1,22 @@
+#import <SVG/SVGElement.idl>
+
+[Exposed=Window]
+interface SVGAnimationElement : SVGElement {
```
tcl3, line 5: Could we add a spec link to https://svgwg.org/specs/animations/#InterfaceSVGAnimationElement

Libraries/LibWeb/SVG/SVGAnimationElement.cpp
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright (c) 2022, Sam Atkins <atkinssj@serenityos.org>
```
tcl3: I don't think Sam wrote this :wink:

Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.cpp
```diff
@@ -18,6 +25,41 @@ GC::Ref<TrustedTypePolicyFactory> TrustedTypePolicyFactory::create(JS::Realm& re
     return realm.create<TrustedTypePolicyFactory>(realm);
 }
 
+Optional<String> TrustedTypePolicyFactory::get_attribute_type(String& tagName, String& attribute, Optional<String> elementNS, Optional<String> attrNS)
```
tcl3: A spec link should be added here: https://w3c.github.io/trusted-types/dist/spec/#dom-trustedtypepolicyfactory-getattributetype.

`snake_case` should be used for method argument names.

Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.cpp
```diff
@@ -18,6 +25,41 @@ GC::Ref<TrustedTypePolicyFactory> TrustedTypePolicyFactory::create(JS::Realm& re
...
+{
+    // 1. Set localName to tagName in ASCII lowercase.
+    auto const localName = tagName.to_ascii_lowercase();
```
tcl3: `snake_case` should be used for local variable declarations.

Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.h
```diff
@@ -42,4 +39,14 @@ class TrustedTypePolicyFactory final : public Bindings::PlatformObject {
     Vector<String> created_policy_names = {};
 };
 
+struct TrustedTypeData {
+    String element;
+    String attributeNs;
+    String attributeLocalName;
+    String trustedType;
+    String Sink;
```
tcl3: `snake_case` should be used for these.

Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.cpp
```diff
@@ -0,0 +1,109 @@
...
+    if (!elementNS.has_value() || elementNS.value().is_empty())
+        elementNS = String { Namespace::HTML };
+
+    // 4. If attrNs is an empty string, set attrNs to null.
+    if (attrNS.has_value() && attrNS.value().is_empty())
+        attrNS.clear();
+
+    // FIXME
+    // 5. Let interface be the element interface for localName and elementNs.
+    auto const& interface = localName;
```
tcl3: Ideally we'd have a method in `ElementFactrory` that could give us the element interface name for a given local name, but I don''t think we have anything currently that does that. 

For now, would something like this work? It accounts for every row in the `get_trusted_type_data_for_attribute()` table.

```cpp
String interface;
if (local_name == HTML::TagNames::iframe && element_ns == Namespace::HTML) {
    interface = "HTMLIFrameElement"_string;
} else if (local_name == HTML::TagNames::script && element_ns == Namespace::HTML) {
    interface = "HTMLScriptElement"_string;
} else if (local_name == SVG::TagNames::script && element_ns == Namespace::SVG) {
    interface = "SVGScriptElement"_string;
} else {
    interface = "Element"_string;
}
```

Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.cpp
```diff
@@ -0,0 +1,109 @@
...
+void TrustedTypePolicyFactory::initialize(JS::Realm& realm)
+{
+    WEB_SET_PROTOTYPE_FOR_INTERFACE(TrustedTypePolicyFactory);
+    Base::initialize(realm);
+}
+
+// https://w3c.github.io/trusted-types/dist/spec/#abstract-opdef-get-trusted-type-data-for-attribute
+Optional<TrustedTypeData> get_trusted_type_data_for_attribute(String const& element, String const& attribute, Optional<String> const& attributeNs)
+{
+    // 1.Let data be null.
```
tcl3: I know it makes no functional difference, but it would be closer to the spec to define `data` as an empty optional here and return it at step 4.

nit: missing space.

Libraries/LibWeb/SVG/SVGAnimationElement.h
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright (c) 2022, Sam Atkins <atkinssj@serenityos.org>
```
tcl3: Your name should go here.

Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.h
```diff
@@ -0,0 +1,52 @@
...
+
+public:
+    [[nodiscard]] static GC::Ref<TrustedTypePolicyFactory> create(JS::Realm&);
+
+    virtual ~TrustedTypePolicyFactory() override { }
+
+    // FIXME: Not implemented
+    bool is_html(JS::Value&) const { return false; }
+    bool is_script(JS::Value&) const { return false; }
+    bool is_script_url(JS::Value&) const { return false; }
```
tcl3: It's probably better to mark these methods as `[FIXME]` in the IDL file rather than stubbing them like this - unless something in particular relies on this method being implemented.

tcl3 (changes requested):
Thanks for working on this!

Most of the comments I have are minor style issues.

Libraries/LibWeb/SVG/SVGAnimationElement.idl
tete17: Indeed

Libraries/LibWeb/SVG/SVGAnimationElement.cpp
tete17: That is what I get when I copy paste :)

Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.cpp
tete17: Should have read the guides :)

Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.h
tete17: Done

Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.cpp
tete17: I added your suggestion and kept a FIXME for the time that we actually have such a function.

This enabled the pass of 20+ more WPT tests. Thanks :)

Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.cpp
tete17: The compiler will output the exact same thing so lets keep it close to the spec :)

Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.h
tete17: I removed all non implemented methods and replaced the for [FIXME] is the IDL.

It will make for a good objective on the next round of MRs

Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.cpp
```diff
@@ -0,0 +1,120 @@
...
+        interface = "SVGScriptElement"_string;
+    } else {
+        interface = "Element"_string;
+    }
+
+    // 6. Let expectedType be null.
+    Optional<String> expectedType {};
+
+    // 7. Set attributeData to the result of Get Trusted Type data for attribute algorithm, with the following arguments, interface as element, attribute, attrNs
+    auto const attributeData = get_trusted_type_data_for_attribute(interface, attribute, attr_ns);
```
tcl3: These should be `snake_case`.

Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.h
```diff
@@ -0,0 +1,43 @@
...
+
+    virtual ~TrustedTypePolicyFactory() override { }
+
+    Optional<String> get_attribute_type(String&, String&, Optional<String>, Optional<String>);
+
+private:
+    explicit TrustedTypePolicyFactory(JS::Realm&);
+    virtual void initialize(JS::Realm&) override;
+
+    Vector<String> created_policy_names = {};
```
tcl3: We use an `m_` prefix for member variables and we don't use `=` when initializing members. Typically for `Vector` members we would leave out the initializing braces entirely.

Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.h
```diff
@@ -0,0 +1,43 @@
...
+class TrustedTypePolicyFactory final : public Bindings::PlatformObject {
+    WEB_PLATFORM_OBJECT(TrustedTypePolicyFactory, Bindings::PlatformObject);
+    GC_DECLARE_ALLOCATOR(TrustedTypePolicyFactory);
+
+public:
+    [[nodiscard]] static GC::Ref<TrustedTypePolicyFactory> create(JS::Realm&);
+
+    virtual ~TrustedTypePolicyFactory() override { }
+
+    Optional<String> get_attribute_type(String&, String&, Optional<String>, Optional<String>);
```
tcl3: It would probably be good to name the arguments here, since it isn't obvious what each string is supposed to represent at a glance.

tcl3 (commented):
Just a few minor style comments

Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.h
tete17: Sure, even clang-tidy was complaining

Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.h
tcl3: To be clear, I think it would be better like this: 
```cpp
Optional<String> get_attribute_type(String& tag_name, String& attribute, Optional<String> element_ns, Optional<String> attr_ns);`
```

Also `tag_name` can be `const&` here.

Libraries/LibWeb/CMakeLists.txt
```diff
@@ -810,6 +810,7 @@ set(SOURCES
     Streams/WritableStreamDefaultController.cpp
     Streams/WritableStreamDefaultWriter.cpp
     Streams/WritableStreamOperations.cpp
+    TrustedTypes/TrustedTypePolicyFactory.cpp
```
AtkinsSJ: Nit: This is in the wrong place alphabetically.

Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.idl
```diff
@@ -0,0 +1,20 @@
+// https://w3c.github.io/trusted-types/dist/spec/#trustedtypepolicyfactory
+[Exposed=(Window,Worker)] interface TrustedTypePolicyFactory {
```
AtkinsSJ: Nit: We place the `[Exposed]` annotation on its own line.

Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.cpp
```diff
@@ -18,6 +25,51 @@ GC::Ref<TrustedTypePolicyFactory> TrustedTypePolicyFactory::create(JS::Realm& re
...
+    } else if (local_name == HTML::TagNames::script && element_ns == Namespace::HTML) {
+        interface = "HTMLScriptElement"_string;
+    } else if (local_name == SVG::TagNames::script && element_ns == Namespace::SVG) {
+        interface = "SVGScriptElement"_string;
+    } else {
+        interface = "Element"_string;
+    }
+
+    // 6. Let expectedType be null.
+    Optional<String> expectedType {};
```
AtkinsSJ: Nit: Capitalize like `expected_type` even when it's a spec variable name.

Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.cpp
```diff
@@ -29,4 +81,40 @@ void TrustedTypePolicyFactory::initialize(JS::Realm& realm)
...
 
+// https://w3c.github.io/trusted-types/dist/spec/#abstract-opdef-get-trusted-type-data-for-attribute
+Optional<TrustedTypeData> get_trusted_type_data_for_attribute(String const& element, String const& attribute, Optional<String> const& attribute_ns)
+{
+    // 1. Let data be null.
+    Optional<TrustedTypeData> data {};
+
+    // 2. If attributeNs is null, and attribute is the name of an event handler content attribute, then:
+    if (!attribute_ns.has_value()) {
+#undef __ENUMERATE
```
AtkinsSJ, line 94: This shouldn't be necessary. Is something else leaving an `__ENUMERATE()` definition around?

Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.cpp
```diff
@@ -29,4 +81,40 @@ void TrustedTypePolicyFactory::initialize(JS::Realm& realm)
...
+#define __ENUMERATE(attribute_name, event_name)                                                                \
+    if (attribute == HTML::AttributeNames::attribute_name) {                                                   \
+        /* 1. Return (Element, null, attribute, TrustedScript, "Element " + attribute). */                     \
+        return TrustedTypeData { "Element"_string, {}, attribute, "TrustedScript"_string, "Element "_string }; \
```
AtkinsSJ: You're not including `attribute` in the final parameter here.

Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.h
tete17: Silly me I did that for the other function name

Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.cpp
```diff
@@ -29,4 +81,40 @@ void TrustedTypePolicyFactory::initialize(JS::Realm& realm)
...
+    }
+
+    static Vector<TrustedTypeData> const table {
+        { "HTMLIFrameElement"_string, {}, "srcdoc"_string, "TrustedHTML"_string, "HTMLIFrameElement srcdoc"_string },
+        { "HTMLScriptElement"_string, {}, "src"_string, "TrustedScriptURL"_string, "HTMLScriptElement src"_string },
+        { "SVGScriptElement"_string, {}, "href"_string, "TrustedScriptURL"_string, "SVGScriptElement href"_string },
+        { "SVGScriptElement"_string, Namespace::XLink.to_string(), "href"_string, "TrustedScriptURL"_string, "SVGScriptElement href"_string },
+    };
+
+    // 3. Find the row in the following table, where element is in the first column, attributeNs is in the second column, and attribute is in the third column. If a matching row is found, set data to that row.
```
AtkinsSJ: Nit: Wrap long spec comments at around 120 characters.

Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.cpp
```diff
@@ -29,4 +81,40 @@ void TrustedTypePolicyFactory::initialize(JS::Realm& realm)
...
+    for (auto const& row : table) {
+        if (row.element == element && row.attribute_ns == attribute_ns && row.attribute_local_name == attribute)
+            data = row;
+    }
```
AtkinsSJ: Use `table.first_matching(*lambda*)` for this. (Or at least `break` when you find a match.)

Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.cpp
tete17: I was just following the pattern of usage of the other places where these expansion macro was used.

Do you mind if we leave it in as a safety of mind thing?

Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.cpp
tete17: I wasn't capable of using any + operator not append function with a string.

Since it is not used for now I though it would be fine to leave it as it.

Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.cpp
tete17: I also wrapped another comment

Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.cpp
tete17: Sounds better indeed

tete17:
Sorry @AtkinsSJ & @tcl3 for all the little dumb mistakes :)

And thanks for the time and review

Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.cpp
AtkinsSJ: So there's a couple of things you can do. One is `MUST(String::formatted("Element {}", attribute))`, but as this is a macro you can also use a C++ trick: Two or more quoted strings next to each other are treated as a single string.

So you can do something like:
```c++
"Element " #attribute_name ""_string
```

Tests/LibWeb/Text/expected/all-window-properties.txt
```diff
@@ -419,6 +420,7 @@ TransformStream
 TransformStreamDefaultController
 TransitionEvent
 TreeWalker
+TrustedTypePolicyFactory
```
ADKaster, line 427: This property is added to this text file in the first commit, but it should be added later, in the commit that defines this interface.

ADKaster:
Well I say that but your test seems to be crashing a sanitizer build

`Crashed: Text/input/wpt-import/trusted-types/TrustedTypePolicyFactory-getAttributeType-event-handler-content-attributes.tentative.html`

Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.cpp
```diff
@@ -0,0 +1,122 @@
...
+}
+
+void TrustedTypePolicyFactory::initialize(JS::Realm& realm)
+{
+    WEB_SET_PROTOTYPE_FOR_INTERFACE(TrustedTypePolicyFactory);
+    Base::initialize(realm);
+}
+
+// https://w3c.github.io/trusted-types/dist/spec/#abstract-opdef-get-trusted-type-data-for-attribute
+Optional<TrustedTypeData const&> get_trusted_type_data_for_attribute(String const& element, String const& attribute, Optional<String> const& attribute_ns)
```
tcl3: It looks like the sanitizer is tripping up because you're returning a reference to a stack allocated variable here.

tete17:
Sorry guys i am in the middle of moving and the only powerfull enough compyter to make changes is in a box. I havent forgotten about this. I hope to get back to it this week

Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.cpp
tete17: Good point thank you for the tip

Tests/LibWeb/Text/expected/all-window-properties.txt
tete17: Indeed

Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.cpp
tete17: Fixed it by making a copy at the last moment.

tcl3: I think it should be possible to avoid the copy by doing something like this:
```diff
diff --git a/Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.cpp b/Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.cpp
index bf639be35ca..3b66f8842a8 100644
--- a/Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.cpp
+++ b/Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.cpp
@@ -87,7 +87,7 @@ void TrustedTypePolicyFactory::initialize(JS::Realm& realm)
 Optional<TrustedTypeData> get_trusted_type_data_for_attribute(String const& element, String const& attribute, Optional<String> const& attribute_ns)
 {
     // 1. Let data be null.
-    Optional<TrustedTypeData const&> data {};
+    Optional<TrustedTypeData> data {};

     // 2. If attributeNs is null, and attribute is the name of an event handler content attribute, then:
     if (!attribute_ns.has_value()) {
@@ -111,12 +111,14 @@ Optional<TrustedTypeData> get_trusted_type_data_for_attribute(String const& elem

     // 3. Find the row in the following table, where element is in the first column, attributeNs is in the second column,
     // and attribute is in the third column. If a matching row is found, set data to that row.
-    data = table.first_matching([&element, &attribute, &attribute_ns](auto const& row) {
+    auto matching_row = table.first_matching([&element, &attribute, &attribute_ns](auto const& row) {
         return row.element == element && row.attribute_ns == attribute_ns && row.attribute_local_name == attribute;
     });
+    if (matching_row.has_value())
+        data = matching_row.release_value();

     // 4. Return data
-    return data.copy();
+    return data;
 }

 }
```

tete17: Doesn't it still incur in a hidden copy? when we release the value and gets passed to data?

tcl3: I think you're right. This just ends up doing a copy with extra steps. My apologies.

tete17: Running it with GDB and breaking in TrustedTypeData::TrustedTypeData show 2 copies in your scenario.

One in `data = matching_row.release_value();` & another in the `return` statement

I am going to be honest not sure exactlly why this one makes 2 copies you probally need a phd in c++ to be sure but my understanding is we need to make a copy of the data otherwise we will be handing over a reference to static variable of a function

tcl3 (approved):
Thanks for working on this!

PR #5431, opened 2025-07-14
LibWeb: Expand ClipFrame into clip rectangles during display list replay

kalenikaliaksandr:
Until now, every paint phase of every PaintableBox injected its own
clipping sequence into the display list:
```
before_paint: Save
              AddClipRect (1)
              ...clip rectangles for each containing block with clip...
              AddClipRect (N)

paint:        ...paint phase items...

after_paint:  Restore
```

Because we ran that sequence for every phase of every box, Skia had to
rebuild clip stack `paint_phases * paintable_boxes` times. Worse,
usually most paint phases contribute no visible drawing at all, yet we
still had to emit clipping items because `before_paint()` has no way to
know that in advance.

This change takes a different approach:
- Clip information is now attached as metadata `ClipFrame` to each
  DisplayList item.
- `DisplayListPlayer` groups consecutive commands that share a
  `ClipFrame`, applying the clip once at the start of the group and
  restoring it once at the end.

Going from 10 ms to 5 ms in rasterization on Discord might not sound
like much, but keep in mind that for 60fps we have 16 ms per frame and
there is a lot more work besides display list rasterization we do in
each frame.

* https://discord.com/channels/1247070541085671459/1247090064480014443
  - DisplayList items:  81844  -> 3671
  - rasterize time:     10 ms  -> 5 ms
  - record time:        5 ms   -> 3 ms

* https://github.com/LadybirdBrowser/ladybird
  - DisplayList items:  7902  -> 1176
  - rasterize time:     4 ms  -> 4 ms
  - record time:        3 ms  -> 2 ms


Libraries/LibWeb/Painting/DisplayList.cpp
```diff
@@ -68,11 +98,36 @@ void DisplayListPlayer::execute_impl(DisplayList& display_list, ScrollStateSnaps
     auto const& commands = display_list.commands();
     auto device_pixels_per_css_pixel = display_list.device_pixels_per_css_pixel();
 
+    DevicePixelConverter device_pixel_converter { device_pixels_per_css_pixel };
+
     VERIFY(!m_surfaces.is_empty());
 
+    Vector<RefPtr<ClipFrame const>> clip_frames_stack;
```
gmta, line 105: Can we move this to a member and call `.clear_with_capacity()` on it here? That way we'd only occasionally allocate in the render path.

gmta (commented):
Awesome!

Tests/LibWeb/Text/expected/display_list/simple-overflow-hidden.txt
```diff
@@ -1,27 +1,9 @@
...
-Save
-AddClipRect rect=[10,10 200x100]
-Restore
 FillPathUsingColor
-Save
-AddClipRect rect=[10,10 200x100]
-Restore
-Save
-AddClipRect rect=[10,10 200x100]
 FillRect rect=[10,10 300x150] color=rgb(240, 128, 128)
```
konradekk, line 5: not sure, but it seems that `AddClipRect` has been lost here? (same below; should these steps share it?)

Libraries/LibWeb/Painting/DisplayList.cpp
kalenikaliaksandr: `DisplayListPlayer::execute_impl()` could be called recursively during display list execution, so moving `clip_frames_stack` into a member would make it harder to pop only clip frames that belong to current display list.

gmta: Oh that's annoying. Let's deal with it when it pops up in profiles then :)

Tests/LibWeb/Text/expected/display_list/simple-overflow-hidden.txt
kalenikaliaksandr: no, all these AddClipRect should be gone, because now every display list item has its clip frame as a metadata.

gmta (approved):

PR #5439, opened 2025-07-14
test-web: Wait longer before completing crash tests

tcl3:
Previously, a lot of imported WPT crash tests wouldn't wait long enough for a crash to occur.

I've also converted a text test that  should have been a crash test initially. I verified to test behaves as expected by manually reintroducing the bug the crash test was written for.


Tests/LibWeb/test-web/main.cpp
```diff
@@ -285,11 +285,30 @@ static void run_dump_test(TestWebView& view, Test& test, URL::URL const& url, in
...
         };
     } else if (test.mode == TestMode::Crash) {
-        view.on_load_finish = [on_test_complete = move(on_test_complete), url](auto const& loaded_url) {
+        view.on_load_finish = [on_test_complete, url, &view, &test](auto const& loaded_url) {
             // We don't want subframe loads to trigger the test finish.
             if (!url.equals(loaded_url, URL::ExcludeFragment::Yes))
                 return;
-            on_test_complete();
+            test.did_finish_loading = true;
+            auto wait_for_crash_test_completion = R"(
```
trflynn89: nit: This string could be `static`

Tests/LibWeb/Text/input/SVG/empty-filter-crash.html
```diff
@@ -1,12 +0,0 @@
...
-<svg>
-    <filter id="empty"></filter>
-    <g filter="url(#empty)"/>
-</svg>
-<script src="../include.js"></script>
-<script>
-    // FIXME: This should be a crash test.
-    test(() => {
-        println("PASS (didn't crash)");
-    })
```
trflynn89, line 11: woops didn't notice before, but shouldn't we also delete the expectation file?

tcl3: I also didn't notice :sweat_smile:. Deleted.

PR #5441, opened 2025-07-14
LibWeb: Prioritize inheriting `text-align` for `<th>`

gmta:
Because we defined `th { text-align: center }` in our UA stylesheet, it received a higher precedence than inherited (inline) styles. Firefox deals with this by defining a custom `text-align` value that prioritizes any inherited value before defaulting to `text-align: center`.

We now do this as well :^)

Fixes the heading alignment on our organization page.

| Before | After|
|--|--|
| image | image |

### DIFF
--- a/Libraries/LibWeb/CSS/StyleComputer.cpp
+++ b/Libraries/LibWeb/CSS/StyleComputer.cpp
@@ -1708,6 +1708,24 @@ void StyleComputer::compute_defaulted_values(ComputedProperties& style, DOM::Ele
         auto const& inherited_value = get_inherit_value(CSS::PropertyID::Color, element, pseudo_element);
         style.set_property(CSS::PropertyID::Color, inherited_value);
     }
+
+    // AD-HOC: The -libweb-inherit-or-center style defaults to centering, unless a style value usually would have been
+    //         inherited. This is used to support the ad-hoc default <th> text-align behavior.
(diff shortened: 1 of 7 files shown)


Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1708,6 +1708,24 @@ void StyleComputer::compute_defaulted_values(ComputedProperties& style, DOM::Ele
...
+            auto parent_cascaded = parent_element->cascaded_properties(pseudo_element);
+            if (!parent_computed || !parent_cascaded)
+                break;
+            if (parent_cascaded->property(PropertyID::TextAlign)) {
+                auto const& style_value = parent_computed->property(PropertyID::TextAlign);
+                style.set_property(PropertyID::TextAlign, style_value, ComputedProperties::Inherited::Yes);
+                break;
+            }
+        }
+    }
```
gmta, line 1728: @AtkinsSJ I feel bad about doing this in every `compute_defaulted_values()` call; is there a way we can limit this somehow? _Technically_ we just need to support this for `<th>` elements.

gmta: @awesomekling I've limited it to `<th>` for now, still not the prettiest. But at least we don't do the property lookup for other elements anymore.

PR #5443, opened 2025-07-14
LibWeb: Use enum for serialization and reimplement interface exposure

Lubrsi:
Our currently implementation of structured serialization has a design
flaw, where if the serialized/transferred type was not used in the
destination realm, it would not be seen as exposed and thus we would
not re-create the type on the other side.

This is very common, for example, transferring a MessagePort to a just
inserted iframe, or the just inserted iframe transferring a MessagePort
to it's parent. This is what Google reCAPTCHA does.

This flaw occurred due to relying on lazily populated HashMaps of
constructors, namespaces and interfaces. This commit changes it so that
per-type "is exposed" implementations are generated.

Since it no longer relies on interface name strings, this commit
changes serializable types to indicate their type with an enum,
in line with how transferrable types indicate their type.

This makes Google reCAPTCHA work on https://www.google.com/recaptcha/api2/demo
It currently doesn't work on non-Google origins due to a separate
same-origin policy bug.

https://github.com/user-attachments/assets/ab9b35cc-a6af-4fb8-b5fd-83f0a9a821a8

This seems to make the cross-piping WPT crash test crash at the moment. Further investigation is needed.

### DIFF
--- a/Libraries/LibWeb/HTML/StructuredSerialize.cpp
+++ b/Libraries/LibWeb/HTML/StructuredSerialize.cpp
@@ -31,10 +31,25 @@
 #include <LibJS/Runtime/StringObject.h>
 #include <LibJS/Runtime/TypedArray.h>
 #include <LibJS/Runtime/VM.h>
+#include <LibWeb/Bindings/DOMExceptionPrototype.h>
+#include <LibWeb/Bindings/DOMMatrixPrototype.h>
+#include <LibWeb/Bindings/DOMMatrixReadOnlyPrototype.h>
+#include <LibWeb/Bindings/DOMPointPrototype.h>
+#include <LibWeb/Bindings/DOMPointReadOnlyPrototype.h>
+#include <LibWeb/Bindings/DOMQuadPrototype.h>
+#include <LibWeb/Bindings/DOMRectPrototype.h>
+#include <LibWeb/Bindings/DOMRectReadOnlyPrototype.h>
 #include <LibWeb/Bindings/ExceptionOrUtils.h>
+#include <LibWeb/Bindings/FileListPrototype.h>
+#include <LibWeb/Bindings/FilePrototype.h>
+#include <LibWeb/Bindings/ImageBitmapPrototype.h>
 #include <LibWeb/Bindings/Intrinsics.h>
+#include <LibWeb/Bindings/MessagePortPrototype.h>
+#include <LibWeb/Bindings/ReadableStreamPrototype.h>
 #include <LibWeb/Bindings/Serializable.h>
 #include <LibWeb/Bindings/Transferable.h>
+#include <LibWeb/Bindings/TransformStreamPrototype.h>
+#include <LibWeb/Bindings/WritableStreamPrototype.h>
 #include <LibWeb/Crypto/CryptoKey.h>
 #include <LibWeb/FileAPI/Blob.h>
 #include <LibWeb/FileAPI/File.h>
@@ -46,6 +61,7 @@
 #include <LibWeb/Geometry/DOMQuad.h>
 #include <LibWeb/Geometry/DOMRect.h>
 #include <LibWeb/Geometry/DOMRectReadOnly.h>
+#include <LibWeb/HTML/ImageBitmap.h>
 #include <LibWeb/HTML/ImageData.h>
 #include <LibWeb/HTML/MessagePort.h>
 #include <LibWeb/HTML/StructuredSerialize.h>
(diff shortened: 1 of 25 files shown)


tcl3:
WPT guy here - not a review, just chiming in to say that this fixes: http://wpt.live/html/infrastructure/safe-passing-of-structured-data/structuredclone_0.html, which was previously skipped.

EDIT: I missed the conversation on Discord where you mentioned this already :sweat_smile: 
I'm getting a failure locally in `Crash/wpt-import/streams/piping/crashtests/cross-piping.html`, which I assume is why CI isn't happy. It's not always easy to see exactly which test failed in the CI logs.

trflynn89:
On this PR, we're crashing in `set_up_cross_realm_transform_writable` called from `WritableStream::transfer_receiving_steps`. 

On master, we currently don't enter `WritableStream::transfer_receiving_steps` at all because we hit this `DataCloneError` - thus don't enter `create_transferred_value`, which would call `WritableStream::transfer_receiving_steps`
https://github.com/LadybirdBrowser/ladybird/blob/423cdd447d306f58eb081db06be00015b67cef07/Libraries/LibWeb/HTML/StructuredSerialize.cpp#L1400-L1407

Here's a slightly simpler repro:
```html
<script>
    let r = new ReadableStream();
    let w = new WritableStream();
    let s = AbortSignal.abort();

    r.pipeTo(w, { signal: s }).catch(() => {
        structuredClone(undefined, { transfer: [w] });
    });
</script>
```

I think the issue is that when we enter `set_up_cross_realm_transform_writable`, we invoke `MessagePort::enable`:
https://github.com/LadybirdBrowser/ladybird/blob/423cdd447d306f58eb081db06be00015b67cef07/Libraries/LibWeb/Streams/AbstractOperations.cpp#L337-L338

This then tries to read as much as possible from the underlying transport socket, which assumes the port is entangled. It's disentangled though, because the `AbortSignal` will have cancelled the streams before the transfer.

The following patch fixes the crash for me, and seems reasonable given that the read is a best-effort attempt anyways:
```diff
diff --git a/Libraries/LibWeb/HTML/MessagePort.cpp b/Libraries/LibWeb/HTML/MessagePort.cpp
index 9824a8dfcf7..cce0cc7e0ed 100644
--- a/Libraries/LibWeb/HTML/MessagePort.cpp
+++ b/Libraries/LibWeb/HTML/MessagePort.cpp
@@ -302,6 +302,9 @@ void MessagePort::read_from_transport()
 {
     VERIFY(m_enabled);

+    if (!is_entangled())
+        return;
+
     auto schedule_shutdown = m_transport->read_as_many_messages_as_possible_without_blocking([this](auto&& raw_message) {
         FixedMemoryStream stream { raw_message.bytes.span(), FixedMemoryStream::Mode::ReadOnly };
         IPC::Decoder decoder { stream, raw_message.fds };
```

Lubrsi:
Do you think it'll be best to open a separate PR for that, open an issue as a reminder and add a test skip in this PR? I think the test crashing is now the actual result of the test with this PR, given it wasn't entering `WritableStream::transfer_receiving_steps` before.

trflynn89:
I think it's fine to just add a commit to this PR with the fix, but if you'd rather make an issue + disable the test for now, that's fine as well!

Lubrsi:
You're right, it's not that much of a deal to do it in a separate PR, especially with it being a small fix. Added to this PR.

Lubrsi:
> WPT guy here - not a review, just chiming in to say that this fixes: http://wpt.live/html/infrastructure/safe-passing-of-structured-data/structuredclone_0.html, which was previously skipped.
> 
> EDIT: I missed the conversation on Discord where you mentioned this already 😅 I'm getting a failure locally in `Crash/wpt-import/streams/piping/crashtests/cross-piping.html`, which I assume is why CI isn't happy. It's not always easy to see exactly which test failed in the CI logs.

Strange, when I re-enable the test, it fails when using test-web, but doesn't fail when running it in on wpt.live 🤔

tcl3:
> Strange, when I re-enable the test, it fails when using test-web, but doesn't fail when running it in on wpt.live

Running `./Meta/WPT.sh import --force html/infrastructure/safe-passing-of-structured-data/structuredclone_0.html` will update the test to the latest version and rebaseline it.

Lubrsi:
Oops, thanks for pointing out the rebaseline!

Libraries/LibWeb/HTML/StructuredSerialize.cpp
```diff
@@ -1047,44 +1063,82 @@ class Deserializer {
     GC::RootVector<JS::Value> m_memory; // Index -> JS value
     size_t m_position { 0 };
 
-    static GC::Ref<Bindings::PlatformObject> create_serialized_type(StringView interface_name, JS::Realm& realm)
+    static bool is_serializable_interface_exposed_on_target_realm(SerializeType name, JS::Realm& realm)
```
trflynn89, line 1066: Just an idea for the future, but I wonder if we can generate `is_serializable_interface_exposed_on_target_realm` and `create_serialized_type`? Would be nice to not have to maintain these large switches

trflynn89 (approved):

PR #5447, opened 2025-07-14
AK: Use Deducing this for OptionalBase

Hendiadyoin1:
This is taken from and akin to
https://github.com/SerenityOS/serenity/pull/25894

----

This greatly reduces the code duplication and is a bit more strict about forwarding

### DIFF
--- a/AK/Optional.h
+++ b/AK/Optional.h
@@ -8,7 +8,11 @@
 #pragma once
 
 #include <AK/Assertions.h>
+#include <AK/Concepts.h>
+#include <AK/Forward.h>
 #include <AK/Noncopyable.h>
+#include <AK/Platform.h>
+#include <AK/StdLibExtraDetails.h>
 #include <AK/StdLibExtras.h>
 #include <AK/Traits.h>
 #include <AK/Try.h>
@@ -32,10 +36,17 @@ struct ConditionallyResultType<false, T> {
     using Type = T;
 };
 
+template<typename Self, typename T>
+struct AddConstIfNeeded {
+    using Type = Conditional<IsConst<RemoveReference<Self>> && !IsConst<T>, AddConst<T>, T>;
+};
+
 }
 
 template<auto condition, typename T>
 using ConditionallyResultType = typename Detail::ConditionallyResultType<condition, T>::Type;
+template<typename Self, typename T>
+using AddConstIfNeeded = typename Detail::AddConstIfNeeded<Self, T>::Type;
 
 // NOTE: If you're here because of an internal compiler error in GCC 10.3.0+,
 //       it's because of the following bug:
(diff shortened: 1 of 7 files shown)


Libraries/LibWeb/Fetch/Request.cpp
```diff
@@ -467,7 +467,7 @@ WebIDL::ExceptionOr<GC::Ref<Request>> Request::construct_impl(JS::Realm& realm,
     }
 
     // 38. Let inputOrInitBody be initBody if it is non-null; otherwise inputBody.
-    auto input_or_init_body = init_body.value_or<Optional<Infrastructure::Request::BodyType const&>>(input_body);
+    auto input_or_init_body = init_body.value_or<Optional<Infrastructure::Request::BodyType const&>, Optional<Infrastructure::Request::BodyType const&>>(input_body);
```
Hendiadyoin1: @alimpfard Any clue why this change is needed?

gmta:
@Hendiadyoin1 CI is failing :-(

Hendiadyoin1:
Yeah, probably need to bump the ci compiler version

ADKaster:
If it doesn't build with Xcode 16.3 clang, we're gonna have to hold off on this. @BertalanD , wanna try this out on the latest Xcode 26 preview? Looks like beta 3 came out a week ago

Hendiadyoin1:
Ah whoops mainly forgot to remove a now superfluous thing
This should work on Xcode16.3 as that did bump the clang version to "17", which is actually LLVM 19
<https://en.wikipedia.org/wiki/Xcode#cite_ref-208>

BertalanD:
Can confirm that Xcode 26 beta 3 (clang-1700.3.13.4) successfully builds this.

Libraries/LibWeb/Fetch/Request.cpp
alimpfard: Yes, you're passing `Optional<Infrastructure::Request::BodyType const&>` as Self (which is ok in this particular instance)
to make that work correctly-

AK/Optional.h
```diff
@@ -52,116 +56,91 @@ struct OptionalNone {
...
-        if (static_cast<Self const&>(*this).has_value())
-            return static_cast<Self const&>(*this).value();
-        return fallback;
+        return self.has_value() ? &self.value() : nullptr;
     }
 
-    template<typename O = T, typename Fallback = O>
-    requires(!IsLvalueReference<O> && !IsRvalueReference<O>)
-    [[nodiscard]] ALWAYS_INLINE constexpr O value_or(Fallback&& fallback) &&
+    template<typename Self, typename O = T, typename Fallback = O>
```
alimpfard: name this
```suggestion
    template<typename O = T, typename Fallback = O, typename Self = OptionalBase<T>>
```

(similarly for anything that went through `<Ts...>` -> `<Self, Ts...>`)

Hendiadyoin1: Thanks, moved the `Self`s to the ends

AK/Optional.h
```diff
@@ -32,10 +36,17 @@ struct ConditionallyResultType<false, T> {
     using Type = T;
 };
 
+template<typename Self, typename T>
+struct AddConstIfNeeded {
+    using Type = Conditional<IsConst<RemoveReference<Self>> && !IsConst<T>, AddConst<T>, T>;
```
alimpfard, line 41: ~~🤔 what's the point of this? `AddConst<T const>` is just `T const` anyway~~
This name is really confusing, I think the intent is `CopyConst<RemoveReference<Self>, T>` but to keep `T = U const` as const if `Self` isn't const?

regardless the `&& !IsConst<T>` is unnecessary, AddConst is idempotent for T const.

Hendiadyoin1: Ah, good looked at `AddConst` and did not see any special handling for `T const`, good to know that it is idempotent anyway
And yes this is supposed to be `CopyConstButDontRemoveItPls`

alimpfard: maybe a different name then, `AddConstFrom<F, T>` maybe?
```cpp
template<typename From, typename To>
struct AddConstFrom {
    using Type = Conditional<IsConst<RemoveReference<From>>, T, AddConst<T>>;
};
```

the RemoveReference bit seems a bit odd to me, but probably okay.

PR #5456, opened 2025-07-15
LibWeb/CSS: Add basic registered properties support with initial values

Norbiros:
This is my first contribution to Ladybird. As suggested by @AtkinsSJ, I started with a simple implementation that can be improved later. 

This PR adds a basic connection between `@property` definitions and `var()` usage. It's noticeable on sites using TailwindCSS classes like `translate-x-[-1]`.

**Before:** Before

**After:** After

### DIFF
--- a/Libraries/LibWeb/CSS/Parser/Parser.cpp
+++ b/Libraries/LibWeb/CSS/Parser/Parser.cpp
@@ -145,6 +145,13 @@ GC::RootVector<GC::Ref<CSSRule>> Parser::convert_rules(Vector<Rule> const& raw_r
 
             m_declared_namespaces.set(as<CSSNamespaceRule>(*rule).prefix());
             break;
+        case CSSRule::Type::Property: {
+            auto& property_rule = as<CSSPropertyRule>(*rule);
+            if (m_document) {
+                const_cast<DOM::Document*>(m_document.ptr())->registered_custom_properties().set(property_rule.name(), property_rule);
+            }
+            [[fallthrough]];
+        }
         default:
             import_rules_valid = false;
             namespace_rules_valid = false;
--- a/Libraries/LibWeb/CSS/StyleComputer.cpp
+++ b/Libraries/LibWeb/CSS/StyleComputer.cpp
@@ -3106,33 +3106,25 @@ void StyleComputer::unload_fonts_from_sheet(CSSStyleSheet& sheet)
     }
 }
 
-static NonnullRefPtr<CSSStyleValue const> custom_property_initial_value(FlyString const& name)
-{
-    // FIXME: Look-up initial value for registered properties. (@property)
-    (void)name;
-
-    // For non-registered properties, the initial value is the guaranteed-invalid value.
-    return GuaranteedInvalidStyleValue::create();
-}
-
 NonnullRefPtr<CSSStyleValue const> StyleComputer::compute_value_of_custom_property(DOM::AbstractElement abstract_element, FlyString const& name, Optional<Parser::GuardedSubstitutionContexts&> guarded_contexts)
 {
     // https://drafts.csswg.org/css-variables/#propdef-
     // The computed value of a custom property is its specified value with any arbitrary-substitution functions replaced.
     // FIXME: These should probably be part of ComputedProperties.
+    auto& document = abstract_element.document();
 
     auto value = abstract_element.get_custom_property(name);
     if (!value || value->is_initial())
-        return custom_property_initial_value(name);
+        return document.custom_property_initial_value(name);
 
     // Unset is the same as inherit for inherited properties, and by default all custom properties are inherited.
     // FIXME: Support non-inherited registered custom properties.
     if (value->is_inherit() || value->is_unset()) {
         if (!abstract_element.parent_element())
-            return custom_property_initial_value(name);
+            return document.custom_property_initial_value(name);
         auto inherited_value = DOM::AbstractElement { const_cast<DOM::Element&>(*abstract_element.parent_element()) }.get_custom_property(name);
         if (!inherited_value)
-            return custom_property_initial_value(name);
+            return document.custom_property_initial_value(name);
         return inherited_value.release_nonnull();
     }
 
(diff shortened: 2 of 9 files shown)


Libraries/LibWeb/CSS/Parser/Parser.cpp
```diff
@@ -145,6 +145,16 @@ GC::RootVector<GC::Ref<CSSRule>> Parser::convert_rules(Vector<Rule> const& raw_r
...
+        case CSSRule::Type::Property: {
+            auto& property_rule = as<CSSPropertyRule>(*rule);
+            if (m_document) {
+                const_cast<DOM::Document*>(m_document.ptr())->set_registered_custom_property(
+                    property_rule.name(),
+                    property_rule
+                );
+            }
+            [[fallthrough]];
+        }
```
AtkinsSJ: Unfortunately it's not correct to register the custom property in the Parser. JS can create, remove, or modify `@property` rules, and we want to update our list of registered custom properties whenever those things happen.

It's fine if you just want to leave a FIXME comment here and someone can sort that out later. The only similar case I can think of is `@font-face` and we don't seem to do that correctly so it's not a good example to copy. :sweat_smile:

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -3105,10 +3105,15 @@ void StyleComputer::unload_fonts_from_sheet(CSSStyleSheet& sheet)
     }
 }
 
-static NonnullRefPtr<CSSStyleValue const> custom_property_initial_value(FlyString const& name)
+static NonnullRefPtr<CSSStyleValue const> custom_property_initial_value(DOM::Document& document, FlyString const& name)
```
AtkinsSJ: This seems like it should be a Document method as that's needed. I just put it here because I didn't know any better.

Libraries/LibWeb/DOM/Document.h
```diff
@@ -23,6 +23,7 @@
 #include <LibUnicode/Forward.h>
 #include <LibWeb/CSS/CSSStyleSheet.h>
 #include <LibWeb/CSS/StyleSheetList.h>
+#include <LibWeb/CSS/CSSPropertyRule.h>
 #include <LibWeb/Cookie/Cookie.h>
```
AtkinsSJ: Make sure to run `clang-format` on your changes. That'll put these in alphabetical order, among other things. You can make that run automatically by installing the commit hook: https://github.com/LadybirdBrowser/ladybird/blob/master/CONTRIBUTING.md#commit-hooks

Libraries/LibWeb/DOM/Document.h
```diff
@@ -1265,6 +1276,10 @@ class Document
     HashTable<GC::Ref<Element>> m_render_blocking_elements;
 
     HashTable<WeakPtr<Node>> m_pending_nodes_for_style_invalidation_due_to_presence_of_has;
+
+    // https://www.w3.org/TR/css-properties-values-api-1/#dom-window-registeredpropertyset-slot
+    // FIXME: For now, it's the "raw" CSSPropertyRule. Later, we'll need to try to parse it into a "real" style value using the property's syntax.
+    HashMap<FlyString, Web::CSS::CSSPropertyRule*> m_registered_custom_properties;
```
AtkinsSJ: I think storing CSSPropertyRule pointers here is actually correct and you don't need the FIXME. (Correct me if I'm wrong though.)

However, just storing `CSSPropertyRule*` isn't safe. It's a garbage-collected type, which means these pointers could become invalid at some point. To make that safe, store it using a `GC::Ref<CSS::CSSPropertyRule>`, and then visit `m_registered_custom_properties` in `Document::visit_edges()` which will tell the garbage collector that someone's using it and it's not garbage.

AtkinsSJ (changes requested):
This is a good start. :^)

Please make any changes by amending the existing commit. [This video](https://www.youtube.com/watch?v=ElRzTuYln0M) can help with that if you're not familiar with how to do it. Or you can ask someone on Discord.

Libraries/LibWeb/DOM/Document.h
```diff
@@ -905,6 +906,18 @@ class Document
 
     StyleInvalidator& style_invalidator() { return m_style_invalidator; }
 
+    // https://www.w3.org/TR/css-properties-values-api-1/#dom-window-registeredpropertyset-slot
+    Optional<GC::Ref<Web::CSS::CSSPropertyRule>> registered_custom_property(FlyString const& property) const
+    {
+        return m_registered_custom_properties.get(property);
+    }
+    void set_registered_custom_property(FlyString const& property, GC::Ref<Web::CSS::CSSPropertyRule> const& rule) const
```
AtkinsSJ: This shouldn't be `const`, and the hash table shouldn't be `mutable`.

Libraries/LibWeb/CSS/Parser/Parser.cpp
```diff
@@ -145,6 +145,15 @@ GC::RootVector<GC::Ref<CSSRule>> Parser::convert_rules(Vector<Rule> const& raw_r
 
             m_declared_namespaces.set(as<CSSNamespaceRule>(*rule).prefix());
             break;
+        case CSSRule::Type::Property: {
+            auto& property_rule = as<CSSPropertyRule>(*rule);
+            if (m_document) {
+                m_document->set_registered_custom_property(
+                    property_rule.name(),
+                    property_rule);
```
AtkinsSJ: ...instead, use a `const_cast` here to make the document non-const.

Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -6575,6 +6577,19 @@ String Document::dump_display_list()
...
+{
+    auto maybe_custom_property = registered_custom_property(name);
+    if (maybe_custom_property.has_value()) {
+        auto parsed_value = maybe_custom_property.value()->initial_style_value();
+        if (!parsed_value)
+            return CSS::GuaranteedInvalidStyleValue::create();
+        return parsed_value.release_nonnull();
+    }
+
+    return CSS::GuaranteedInvalidStyleValue::create();
```
AtkinsSJ, line 6597: Previously we had a comment here about this being the default initial value, can we keep that? Maybe link to https://drafts.csswg.org/css-variables/#propdef- which defines that.

AtkinsSJ (approved):
Looks good to me! Thanks, and welcome to Ladybird. :^)

PR #5459, opened 2025-07-15
LibWeb/CSS: Remove `:target-within`, add `:unchecked`

AtkinsSJ:
A couple of small recent psuedo-class changes.


Libraries/LibWeb/DOM/Element.cpp
```diff
@@ -1443,6 +1443,26 @@ bool Element::matches_checked_pseudo_class() const
     return false;
 }
 
+bool Element::matches_unchecked_pseudo_class() const
+{
+    // AD-HOC: There is no spec for this yet, so it's based on the spec for :checked, assuming that :unchecked applies to the same cases but with a `false` value.
+    if (is<HTML::HTMLInputElement>(*this)) {
+        auto const& input_element = static_cast<HTML::HTMLInputElement const&>(*this);
```
gmta: This is a perfect candidate for `if (auto const* input_element = as_if<HTML::HTMLInputElement>(this)) {` (and the HTMLOptionElement below as well)

AtkinsSJ: :man_facepalming: I did use that at one point but I backed out some changes to `:indeterminate` I wasn't confident on and must have removed it.

AtkinsSJ:
Changes: Switched to using `as_if()` in the code for matching `:checked` and `:unchecked`.

PR #5462, opened 2025-07-15
LibWeb: Reset document't cursor blink timer after selection change

kalenikaliaksandr:
For contenteditable `Selection` in collapsed state represents cursor position, so whenever it's update, we should also reset the blink timer.

Fixes bug on Discord when cursor is blinking while it is being moved by pressing arrow keys.


Lubrsi:
typo: document't -> document's in commit title

kalenikaliaksandr:
oops, fixed!

PR #5466, opened 2025-07-15
Devcontainer: Enable vcpkg asset cache for vcpkg cache feature

ADKaster:
This enables our vcpkg asset cache in the devcontainer setup, protecting
CI jobs against flaky upstream tarball downloads.


ADKaster:
Food for thought: vcpkg build system integration sure is a drag. Would be interesting to see if just manually building a "staging root" using the install_vcpkg.py we have in Meta/gn/build would be less hassle.

PR #5468, opened 2025-07-16
Meta: Add `WPT.sh bisect` for finding WPT regressions automatically

tcl3:
This PR adds the a `bisect` command to `WPT.sh`. The command takes a bad commit ref, a good commit ref and a list of tests. For example, if we know the test `http://wpt.live/css/css-variables/variable-exponential-blowup.html` crashes on `master` but is OK on commit `43778e9493`, then we can run:
```sh
./Meta/WPT.sh bisect master 43778e9493 css/css-variables/variable-exponential-blowup.html
```

The command will first create a baseline log file and bisect based on whether it encounters results that differ from the baseline. It should finish when it gets to the first commit where the test results differ from the baseline.

If the build fails at any point, it will first be retried, then the bisect script will skip the commit where the build is broken.

The biggest limitation of the script currently is that it relies on `ladybird.py` from the repository. Therefore, it can't bisect commits before that file existed. In theory we should be able to get round this by copying the necessary build files elsewhere before starting to bisect, but I wasn't able to get this to work.

### DIFF
--- a/Meta/WPT.sh
+++ b/Meta/WPT.sh
@@ -16,6 +16,8 @@ BUILD_PRESET=${BUILD_PRESET:-default}
 
 BUILD_DIR=$(get_build_dir "$BUILD_PRESET")
 
+TMPDIR=${TMPDIR:-/tmp}
+
 : "${TRY_SHOW_LOGFILES_IN_TMUX:=false}"
 : "${SHOW_LOGFILES:=true}"
 : "${SHOW_PROGRESS:=true}"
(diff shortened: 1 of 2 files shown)


Meta/WPT.sh
```diff
@@ -680,7 +680,12 @@ if [[ "$CMD" =~ ^(update|clean|run|serve|compare|import|list-tests)$ ]]; then
             ;;
 
         compare)
-            INPUT_LOG_NAME="$(pwd -P)/$1"
+            log_file_directory="$(dirname "$1")"
+            if [[ "${log_file_directory}" == "." ]]; then
+                INPUT_LOG_NAME="$(pwd -P)/$1"
+            else
+                INPUT_LOG_NAME="${log_file_directory}/$(basename "$1")"
+            fi
```
gmta: I think you can use `realpath` to resolve any relative components in the provided path?

Meta/wpt-bisect-helper.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
+
```
gmta, line 2: I think we can add `set -euo pipefail` here to prevent some common issues.

gmta (approved):

PR #5470, opened 2025-07-16
Tests: Implement WPT's fuzzy matching, enable all screenshot tests for all platforms

gmta:
We now read WPT's `<meta name="fuzzy">` tags if present in tests and apply them while comparing screenshots. By adding these tags to our own screenshot tests, we can limit the amount of color/pixel errors present in them as a sort of baseline, and run these tests again on all platforms.


Tests/LibWeb/test-web/main.cpp
```diff
@@ -450,6 +449,14 @@ static void run_ref_test(TestWebView& view, Test& test, URL::URL const& url, int
         auto mismatch_references = metadata_object.get_array("mismatch_references"sv);
         VERIFY(!match_references->is_empty() || !mismatch_references->is_empty());
 
+        // Read fuzzy configuration.
+        test.fuzzy_matches.clear_with_capacity();
+        auto fuzzy_values = metadata_object.get_array("fuzzy"sv);
+        for (size_t i = 0; i < fuzzy_values->size(); ++i) {
+            auto fuzzy_match = parse_fuzzy_match(fuzzy_values->at(i).as_string());
```
tcl3: Should we output some kind of error if the fuzzy value is empty? It might be useful to indicate that the content attribute on the meta tag is missing or empty.

Tests/LibWeb/test-web/Fuzzy.cpp
```diff
@@ -0,0 +1,104 @@
...
+    // * Unnamed: "(#X-)#Y;(#X-)#Y" (maxDifference and totalPixels are assumed in this order)
+    auto config_parts = MUST(fuzzy_value.split(';'));
+    for (size_t i = 0; i < config_parts.size(); ++i) {
+        auto const& config_part = config_parts[i];
+        auto named_parts = MUST(config_part.split('='));
+        if (named_parts.size() > 1) {
+            if (named_parts[0] == "maxDifference"sv)
+                result.color_value_error = parse_fuzzy_range(named_parts[1]);
+            else if (named_parts[0] == "totalPixels"sv)
+                result.pixel_count_error = parse_fuzzy_range(named_parts[1]);
```
tcl3: It might be worth outputting a similar invalid configuration warning to the one below if `color_value_error` or `pixel_count_error` already have a value.

Tests/LibWeb/test-web/Fuzzy.cpp
```diff
@@ -0,0 +1,104 @@
...
+                return false;
+        } else if (diff.maximum_error > 0) {
+            return false;
+        }
+
+        if (fuzzy_match.pixel_count_error.has_value()) {
+            auto pixel_count_range = fuzzy_match.pixel_count_error.value();
+            if (diff.pixel_error_count < pixel_count_range.minimum_value || diff.pixel_error_count > pixel_count_range.maximum_value)
+                return false;
+        } else if (diff.pixel_error_count > 0) {
```
tcl3: It looks like if `maxDifference` is given but `totalPixels` is not the test will always fail here? I'm not sure what the intended behavior here is because there doesn't seem to be any examples in the WPT repo where only one value is set.

I'm not sure this is wrong, but I'm not sure how it's supposed to work.

Tests/LibWeb/test-web/Fuzzy.cpp
```diff
@@ -0,0 +1,104 @@
...
+    for (auto const& fuzzy_match : fuzzy_matches) {
+        if (fuzzy_match.reference.has_value()) {
+            dbgln("FIXME: Fuzzy match restricted to '{}', but this is not implemented.", fuzzy_match.reference.value());
+            continue;
+        }
+
+        if (fuzzy_match.color_value_error.has_value()) {
+            auto color_error_range = fuzzy_match.color_value_error.value();
+            if (diff.maximum_error < color_error_range.minimum_value || diff.maximum_error > color_error_range.maximum_value)
+                return false;
```
tcl3: It would be nice to output some kind of error at these returns. At the moment it's quite difficult to figure out why a reftest fails.

Tests/LibWeb/test-web/Fuzzy.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+// https://web-platform-tests.org/writing-tests/reftests.html#fuzzy-matching
+bool fuzzy_screenshot_match(Gfx::Bitmap const& bitmap_a, Gfx::Bitmap const& bitmap_b, Vector<FuzzyMatch> const& fuzzy_matches)
+{
+    // No fuzzy matching: exact match.
+    auto diff = bitmap_a.diff(bitmap_b);
+    if (fuzzy_matches.is_empty())
+        return diff.identical;
+
+    // Apply each fuzzy match to the diff result.
```
tcl3: My understanding of the WPT runner is that it allows one fuzzy match per "match" or "mismatch" attribute. Given that we currently only allow 1 match or mismatch attribute, I think we should only be applying a single fuzzy match here.

gmta:
@tcl3 

> It might be worth adding some reftests to test various "fuzzy", "match" and "mismatch" combinations. I've been using this to test, for example.

I've imported a couple of WPT infrastructure tests for that!

Tests/LibWeb/test-web/Fuzzy.cpp
gmta: You're right - changed this to only match a single fuzzy config. I've also implemented support for the reference argument, so we now only select the fuzzy config that would apply to the reference URL.

Tests/LibWeb/test-web/Fuzzy.cpp
gmta: We should always test for both `maxDifference` and `totalPixels`, so this logic is now significantly simplified.

Tests/LibWeb/test-web/main.cpp
gmta: Added lots of `ErrorOr<..>` to the fuzzy parsing, thanks :)

Tests/LibWeb/test-web/Fuzzy.cpp
gmta: I've added a formatter for `FuzzyRange` and made it perform a `warnln(..)` when something mismatches, for example:

```
Fuzzy mismatch: maximum error 5 is outside FuzzyRange [0-4]
```

Tests/LibWeb/test-web/Fuzzy.cpp
gmta: Yeah, both named and positional parsing could do with some extra sanity checks. Added them!

PR #5472, opened 2025-07-16
Meta: Update vcpkg, including simdutf for a `Uint8Array.fromBase64` fix

trflynn89:
Updating simdutf for a fix for handling invalid trailing padding characters.

Updated other packages while I'm here since this is a cache-busting change.

test262 diff:
```
Diff Tests:
    test/built-ins/Uint8Array/fromBase64/last-chunk-invalid.js ❌ -> ✅
```

### DIFF
--- a/Libraries/LibCrypto/Hash/HKDF.cpp
+++ b/Libraries/LibCrypto/Hash/HKDF.cpp
@@ -32,8 +32,16 @@ ErrorOr<ByteBuffer> HKDF::derive_key(Optional<ReadonlyBytes> maybe_salt, Readonl
         OSSL_PARAM_END,
         OSSL_PARAM_END,
     };
+
     if (maybe_salt.has_value()) {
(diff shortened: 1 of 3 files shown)


Libraries/LibCrypto/Hash/HKDF.cpp
```diff
@@ -32,8 +32,14 @@ ErrorOr<ByteBuffer> HKDF::derive_key(Optional<ReadonlyBytes> maybe_salt, Readonl
         OSSL_PARAM_END,
         OSSL_PARAM_END,
     };
+
     if (maybe_salt.has_value()) {
-        params[3] = OSSL_PARAM_octet_string(OSSL_KDF_PARAM_SALT, const_cast<u8*>(maybe_salt->data()), maybe_salt->size());
+        static constexpr u8 empty_salt[0] {};
```
gmta, line 37: We should probably have a `FIXME` here to remind us to check if this is still necessary after the next OpenSSL update.

trflynn89: Done. I also left the folks at openssl a comment earlier asking if this was indeed an intended change.

gmta (approved):

PR #5475, opened 2025-07-16
LibWeb: Improve graphical list item marker positioning

gmta:
While 788d5368a7b4e898d51c280ba17f3ee5916bf50e took care of better text marker positioning, this improves graphical marker positioning instead.

By looking at how Firefox and Chrome render markers, it's clear that there are three parts to positioning a graphical marker:

  * The containing space that the marker resides in;
  * The marker dimensions;
  * The distance between the marker and the start of the list item.

The space that the marker can be contained in, is the area to the left of the list item with a height of the marker's line-height. The marker dimensions are relative to the marker's font's pixel size: most of them are a square at 35% of the font size, but the disclosure markers are sized at 50% instead. Finally, the marker distance is always gauged at 50% of the font size.

So for example, a list item with `list-style-type: disc` and `font-size: 20px`, has 10px between its start and the right side of the marker, and the marker's dimensions are 7x7.

The percentages I've chosen closely resemble how Firefox lays out its list item markers.

| Before | After | Source |
|--|--|--|
| image | image | [website](https://www.w3schools.com/cssref/tryit.php?filename=trycss_list-style-type_all) |
| image | image | [website](https://nos.nl/collectie/13959/liveblog/2574990-81-doden-gemeld-door-israelische-aanvallen-gaza-ambassadeur-vs-bij-proces-netanyahu) |


AtkinsSJ (approved):
Looks good!

PR #5484, opened 2025-07-17
LibWeb/CSS: Support legacy value aliases for the properties that have them

AtkinsSJ:
[Legacy value aliases](https://drafts.csswg.org/css-cascade-5/#css-legacy-value-alias) are, for now, keywords that get swapped for other keywords at parse-time. Specifically there are two users of this now, at least according to [webdex](https://dontcallmedom.github.io/webdex/l.html#legacy%20value%20alias%40%40CSS%40dfn):

- `overlay` in `overflow` is an alias for `auto`, also applying to `overflow-x` and `overflow-y`.
- `distribute` in `text-justify` is an alias for `inter-character`.

`overflow: overlay` specifically showed up on bbc.co.uk.

Until there are more uses of this, it doesn't seem worth implementing a proper automatic system to handle them, but I've left some FIXMEs about that.

### DIFF
--- a/Libraries/LibWeb/CSS/Enums.json
+++ b/Libraries/LibWeb/CSS/Enums.json
@@ -638,8 +638,7 @@
     "auto",
     "none",
     "inter-word",
-    "inter-character",
-    "distribute=inter-character"
+    "inter-character"
   ],
   "text-overflow": [
     "clip",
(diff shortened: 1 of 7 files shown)


AtkinsSJ:
Changes: I noticed we apparently had a system for this, but turns out it's different. Some keywords are kept around but behave the same as others. Legacy value aliases instead have to *replace* the original value with the new one. So I've removed the wrong fix that we had for `text-justify: distribute`.

PR #5488, opened 2025-07-17
LibWeb: Fix up same-origin domain check by implementing document.domain

Lubrsi:
The same-origin domain check always returned true if only the scheme matched. This was because of missing steps to check for the origin's domain, which didn't exist.

This implements it!

I don't think our testing infrastructure supports cross-origin setups at the time of writing, so this doesn't add any tests.


Libraries/LibURL/Origin.h
```diff
@@ -41,6 +42,9 @@ class Origin {
     Host const& host() const { return m_state.get<Tuple>().host; }
     Optional<u16> port() const { return m_state.get<Tuple>().port; }
 
+    void set_domain(Optional<Host> value) { m_state.get<Tuple>().domain = value; }
```
shannonbooth: nit but can be moved

Libraries/LibURL/Origin.h
```diff
@@ -112,6 +122,7 @@ class Origin {
         Optional<String> scheme;
         Host host;
         Optional<u16> port;
+        Optional<Host> domain;
```
shannonbooth, line 123: This also needs to be serialized over IPC in decoder/encoder

shannonbooth:
Is it not possible to use an iframe for the cross origin setup? 🤔

Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -3841,9 +3841,9 @@ WebIDL::ExceptionOr<void> Document::set_domain(String const& domain)
 
     // FIXME: 6. If the surrounding agent's agent cluster's is origin-keyed is true, then return.
```
shannonbooth, line 3842: I wasnt sure about enabling the document domain setter without having this, but I suppose the previous same origin check was much more broken

Lubrsi:
Adding to the IPC encoder/decoder and hash algorithm for Origin made me realise that Origin should really be immutable, so I've removed the set_domain method and made it construct a new Origin instead.

Lubrsi:
> Is it not possible to use an iframe for the cross origin setup? 🤔

I think that can only really test about, file and blob URLs, which makes it difficult to test the port and such changing, since we don't have a HTTP server to host different resources on different origins at the moment, as far as I can tell.

shannonbooth:
> Adding to the IPC encoder/decoder and hash algorithm fors Origin made me realise that Origin should really be immutable, so I've removed the set_domain method and made it construct a new Origin instead.

Right, ideally! Though the whole concept of domain in origins competely subverts an origin being immutable, so that setter also kind of reflects that ugliness

> I think that can only really test about, file and blob URLs, which makes it difficult to test the port and such changing, since we don't have a HTTP server to host different resources on different origins at the moment, as far as I can tell.

I think it is still testable using document.domain setter by creating two documents under the same origin and then using the document.domain setter on one or both of those documents. But it is quite a bit of a pain, it took me a bit to write a test for even two opaque origin behaviour when implementing that 😅 Fair enough to skip, I think best solution longer term is our test harness being able to handle the WPT web server stuff. 

Though I do wonder if we should skip implementing the document.domain setter for now and only be fixing the same origin domain implementation until we implement that origin-keyed concept. Its already a not very good API security wise even with that check, and I havent seen any site call it so far. But maybe that isnt right and we should just add it.

PR #5492, opened 2025-07-17
LibWeb+WebWorker: Use IPC mechanics for structured serialization

trflynn89:
Our structured serialization implementation had its own bespoke encoder and decoder to serialize JS values. It also used a `u32` buffer under the hood, which made using its structures a bit awkward. We had previously worked around its data structures in transferable streams, which nested transfers of `MessagePort` instances. We basically had to add hooks into the `MessagePort` to route to the correct transfer receiving steps, and we could not invoke the correct AOs directly as the spec dictates.

We now use IPC mechanics to encode and decode data. This works because, although we are encoding JS values, we are only ultimately encoding primitive and basic AK types. The resulting data structures actually enforce that we implement transferable streams exactly as the spec is worded.


Libraries/LibWeb/Geometry/DOMQuad.cpp
```diff
@@ -103,48 +103,51 @@ GC::Ref<DOMRect> DOMQuad::get_bounds() const
...
-WebIDL::ExceptionOr<void> DOMQuad::deserialization_steps(ReadonlySpan<u32> const& serialized, size_t& position, HTML::DeserializationMemory& memory)
+WebIDL::ExceptionOr<void> DOMQuad::deserialization_steps(HTML::TransferDataDecoder& serialized, HTML::DeserializationMemory& memory)
 {
+    auto& vm = this->vm();
     auto& realm = this->realm();
+
+    auto deserialize_dom_point = [&](GC::Ref<DOMPoint>& storage) -> WebIDL::ExceptionOr<void> {
+        auto deserialized = TRY(HTML::structured_deserialize_internal(vm, serialized, realm, memory));
+
+        if (auto* dom_point = as_if<DOMPoint>(deserialized.as_object()))
```
ADKaster: this looks pre-existing but uh. what if we deserialized something that's not a DOMPoint? surely that should be an error instead of glossed over? or perhaps runtime asserted?

Libraries/LibWeb/HTML/StructuredSerialize.cpp
```diff
@@ -452,147 +420,34 @@ class Serializer {
...
+                        serialized.append(move(output_value));
 
-                        property_count++;
+                        ++property_count;
                     }
                 }
-                memcpy(m_serialized.data() + count_offset, &property_count, sizeof(property_count));
+
+                auto* data = const_cast<u8*>(serialized.buffer().data().data());
+                memcpy(data + count_offset, &property_count, sizeof(property_count));
```
ADKaster:
```suggestion
                serialized.buffer().data().span().overwrite(count_offset, &property_count, sizeof(property_count);
```
?

but this makes me question how we know we reserved/allocated enough space in the message buffer.

Libraries/LibWeb/HTML/StructuredSerialize.cpp
```diff
@@ -683,7 +536,7 @@ WebIDL::ExceptionOr<void> serialize_viewed_array_buffer(JS::VM& vm, Vector<u32>&
 
     // 4. Assert: bufferSerialized.[[Type]] is "ArrayBuffer", "ResizableArrayBuffer", "SharedArrayBuffer", or "GrowableSharedArrayBuffer".
     // NOTE: Object reference + memory check is required when ArrayBuffer is transferred.
-    auto tag = buffer_serialized[0];
+    auto tag = static_cast<ValueTag>(buffer_serialized.data()[0]);
```
ADKaster: this is fine because ValueTag is `: u8`, but this kind of interpretation of raw bytes from the network as enum values always makes me nervous.

Libraries/LibWeb/HTML/StructuredSerialize.cpp
```diff
@@ -1144,96 +993,11 @@ class Deserializer {
...
-    auto string = TRY(deserialize_string_primitive(vm, vector, position));
-    auto string_view = TRY(Bindings::throw_dom_exception_if_needed(vm, [&string]() {
-        return string->utf8_string_view();
-    }));
-    auto bigint = MUST(::Crypto::SignedBigInteger::from_base(10, string_view.substring_view(0, string_view.length() - 1)));
-    return JS::BigInt::create(vm, bigint);
-}
+    JS::VM& m_vm;
+    TransferDataDecoder& m_serialized;
+    GC::RootVector<JS::Value> m_memory; // Index -> JS value
```
ADKaster: comment here seems out of date. from what I remember this used to be a HashMap

ADKaster:
Windows build seems to complain about a missing AK header 
```
C:\a\ladybird\ladybird\Libraries\LibWeb/HTML/StructuredSerialize.h(61,5): error: unknown type name 'FixedMemoryStream'
   61 |     FixedMemoryStream m_stream;
      |     ^
1 error generated.
```

Libraries/LibWeb/HTML/StructuredSerialize.cpp
trflynn89: `span()` is read-only, we can't call `overwrite`

> how we know we reserved/allocated enough space in the message buffer

We encoded a zero placeholder above the loop. If that failed, we'd have crashed due to small OOM.

Libraries/LibWeb/HTML/StructuredSerialize.cpp
trflynn89: Ah, I wrote this before I wrote `TransferDataDecoder` and neglected to go back and make use of it. We now decode this the normal way.

Libraries/LibWeb/Geometry/DOMQuad.cpp
trflynn89: I was wondering why these don't just assert. That might be better than throwing an error? Not sure. If an attacker got a hold of the transport socket, crashing WebContent / WebWorker might be the better route?

trflynn89: Since almost everyone is already assuming the correct type is encoded, so I've added a commit to make DOMQuad/FileList/ImageData make the same assumptions. So we'll consistently crash WebContent if we have a bad actor instead of silently moving on with bad data.

Libraries/LibWeb/Geometry/DOMMatrixReadOnly.cpp
```diff
@@ -702,92 +702,95 @@ WebIDL::ExceptionOr<String> DOMMatrixReadOnly::to_string() const
...
-        HTML::serialize_primitive_type(serialized, this->m22());
+        serialized.encode(m22());
         // 5. Set serialized.[[M41]] to value’s m41 element.
-        HTML::serialize_primitive_type(serialized, this->m41());
+        serialized.encode(m41());
         // 6. Set serialized.[[M42]] to value’s m42 element.
-        HTML::serialize_primitive_type(serialized, this->m42());
+        serialized.encode(m42());
         // 7. Set serialized.[[Is2D]] to true.
-        // NOTE: This is set in the beginning of the function.
```
shannonbooth, line 723: I don't believe this is intentional

trflynn89: ha! I think what happened was I was multi-line editing all of the `this` instances, accidentally picked this up, then deleted `this` + 2 chars

PR #5497, opened 2025-07-18
LibWeb: Generate some more code for serializable objects

shannonbooth:
Sitting on top of #5492.

Makes use of the IDL generator to reduce some of the boilerplate of serializable objects and makes it a bit easier to check if a given platform object is exposed for a given realm.


Libraries/LibWeb/HTML/StructuredSerialize.cpp
```diff
@@ -1105,16 +1064,15 @@ WebIDL::ExceptionOr<SerializedTransferRecord> structured_serialize_with_transfer
 
 static bool is_transferable_interface_exposed_on_target_realm(TransferType name, JS::Realm& realm)
```
shannonbooth, line 1065: We could probably also replace the `TransferType` enum with `InterfaceName`, which will also allow us to get rid of this function as well, but we would need to add `ArrayBuffer` and `ResizableArrayBuffer` to `InterfaceName`. Maybe that is fine though?

Meta/Lagom/Tools/CodeGenerators/LibWeb/GenerateWindowOrWorkerInterfaces.cpp
```diff
@@ -178,77 +180,70 @@ void Intrinsics::create_web_namespace<@namespace_class@>(JS::Realm& realm)
...
+// FIXME: Make this compatible with non-interface types.
+bool is_exposed(InterfaceName name, JS::Realm& realm)
+{
+    auto const& global_object = realm.global_object();
 
-        auto shadow_realm_exposed_iterator = interface_sets.shadow_realm_exposed.find_if([&name](IDL::Interface const& interface) {
-            return interface.name == name;
-        });
+    // 1. If construct’s exposure set is not *, and realm.[[GlobalObject]] does not implement an interface that is in construct’s exposure set, then return false.
+    if (is<HTML::Window>(global_object) && is_window_exposed(name))
```
Lubrsi: nit: For this and the following if checks, we could flip the `is_*_exposed` check and the `is<>()` to avoid the cost of the `is<>()` if it's not exposed for that global object.

shannonbooth: Ah yup good point

Libraries/LibWeb/HTML/StructuredSerialize.cpp
ADKaster: that seems fine to me. we could hardcode it in the generator. 

though, it seems like we've got a couple uses for all this type metadata. one is "is this object exposed on a  global of type X". Another is "can we serialize or transfer this type". Is it correct to use the same generic-sounding "interface name" enum for all those?

PR #5498, opened 2025-07-18
LibGfx: Add PNG preprocess validator

mikiubo:
This validator skips malformed chunks,
particularly those with chunk type headers containing non-ASCII characters.

This allows passing the WPT test:
png/errors/unknown-ancillary-error-recovery


Libraries/LibGfx/ImageFormats/PNGLoader.cpp
```diff
@@ -135,8 +135,83 @@ static void log_png_warning(png_structp, char const* warning_message)
...
+    }
+
+    return output_buffer;
+}
+
 ErrorOr<void> PNGImageDecoderPlugin::initialize()
 {
+    auto sanitized_data = TRY(preprocess_png_data(m_context->data));
+    ReadonlyBytes clean_data = sanitized_data.bytes();
+    m_context->data = clean_data;
```
gmta, line 213: If I understand correctly, "preprocessing the PNG data" here means copying the entire input buffer to a new byte buffer and skipping chunks with invalid tags? That seems very wasteful, especially with large PNGs - isn't there a libpng way of getting the test to pass?

LucasChollet: I've looked into this issue in the past and the "state of the art" way of solving this issue is to parse the chunk headers yourself and baby spoon feed `libpng` with the chunk you want using the progressive API. This is what both WebKit and Blink do.
I submitted a patch to `libpng` a few months ago to solve this issue. However there is arguably a spec issue and while the image is in WPT, it might be invalid. The patch was then reverted because of that and the `libpng` maintainer wanted to fix the spec issue first. This was a few months ago though, the third edition was published in the meantime and I don't know what happened to that specific issue.
My opinion was that we could wait until this is solved in `libpng` or do it the Blink/WebKit way if we find a faulty image like this in the wild.

mikiubo: You are completely right. In the beginning, I also tried some approaches with libpng, but I failed.
The best solution, as you suggested, will be to wait for the patch in libpng.

Maybe I’ll also try to check WebKit’s approach in the meantime.
If the patch in libpng is released, I’ll close this PR — or if you prefer, I can close it now?

It wasn’t useless for me; I learned a lot about how PNG works.
Thanks!

PR #5500, opened 2025-07-18
Meta: Have an offline Angle build in Flatpak manifest

JanKoudijs:
This PR picks up the first bullet point of the [open points](https://github.com/LadybirdBrowser/ladybird/pull/5278#pullrequestreview-2998471480) in the Flatpak manifest:

> - bad angle build (not offline, too many deps downloaded)

The taken approach is to parse the [angle/DEPS](https://github.com/google/angle/blob/chromium/7258/DEPS) file and generate Flatpak manifest sources from it, using the `condition` to determine what needs to be included.

A `deps` entry like:

```python
  'build': {
    'url': Var('chromium_git') + '/chromium/src/build.git@169fcf699b64d2d5e75a391beaec8a7ad6e41a7f',
    'condition': 'not build_with_chromium',
  },
```

Can be converted into:

```json
    {
        "type": "git",
        "url": "https://chromium.googlesource.com/chromium/src/build.git",
        "commit": "169fcf699b64d2d5e75a391beaec8a7ad6e41a7f",
        "dest": "angle/build",
        "disable-submodules": true
    },
```

The tricky part is in the `cipd` entries. I haven't found a way to get a static url to a package, it seems that only temporary download links are provided (who are of no use for the Flatpak manifest). Instead, the cipd entries are converted into a [cipd ensure file](https://pkg.go.dev/go.chromium.org/luci/cipd/client/cipd/ensure), this file can be passed to the `cipd` tool, but it requires online access:

    cipd ensure -root angle -ensure-file angle.ensure

The PR (currently) has separate commits to show the steps I have taken. Especially the [last commit](https://github.com/LadybirdBrowser/ladybird/commit/d14db339776f425f882b2b392dec7cb458cce44e) needs more pairs of eyes to see if this causes any unwanted dropping of Angle features.

The thing is that I dropped the use of the [ensure file](https://github.com/JanKoudijs/ladybird/blob/meta-flatpak-angle/Meta/CMake/flatpak/angle.ensure) to come to an offline build, and although the build passes it still has to be figured out if no build features get dropped by this action. If it does then these dependencies have to be added in another way.

It is for this reason that I made this a draft PR, but please provide feedback as I'm new to Ladybird and its dependencies. Also the Python code of the generator has to land somewhere, suggestions for where to put this are appreciated.


Meta/CMake/flatpak/org.ladybird.Ladybird.json
```diff
@@ -82,18 +82,141 @@
...
-          "--share=network"
-        ]
-      },
+      "build-commands": [
+        "./angle-checkout.sh",
+        "./angle-configure.sh",
+        "./angle-build.sh",
+        "./angle-install.sh"
+      ],
       "sources": [
```
ADKaster, line 110: the final list of sources you have here looks... close? I don't think we want to download all these sysroots though.

It would be nice to avoid the gclient post-hooks as well. Or at least, avoid the rather exciting list of files that are added to our repo to support calling those python scripts.

I'd rather patch the gn build to skip things that we aren't using than download a bunch of extra stuff to fit into the 'standard chromium build flow' or w/e.

JanKoudijs: As a general strategy, I tried to stay as close as possible to what is fetched by the `gclient sync` command.

Is there a way I can figure out what dependencies are actually needed by the Ladybird build?

> I don't think we want to download all these sysroots though.

The current `angle-configure.sh` script depends on the amd64 sysroot.

> I'd rather patch the gn build to skip things that we aren't using than download a bunch of extra stuff to fit into the 'standard chromium build flow' or w/e.

Good to know. Do you consider gn the best candidate for a build that has a clear separation between a fetch and actual build phase, as required by Flatpak and probably other package formats?

ADKaster: Well, the vcpkg port we're using is kinda cursed. It downloads the CMake build that WebKit created, and jams it into the vcpkg port system. I'd rather avoid doing that. 

My ideal build for this flatpak environment would use the flatpak /usr prefix as its sysroot -- that is, no sysroot at all. Well. With /app as a secondary I guess, for any dependencies we build ourselves before Angle. But that "should" be handled by environment variables that the flatpak-builder environment is setting. If not, we'd have to add them via cflags/ldflags gn args. I'd also like to not use the "bundled" chromium clang, but instead use the system compiler -- in the case of the KDE 6.9 SDK, that's gcc 14. If that's not possible, we can add the llvm SDK extension and use that clang.

As far as what dependencies are actually needed, I suspect the list is actually tiny. We want to use system mesa and system Vulkan headers, we want to skip everything needed for SwiftShader and Dawn support. Afaik we're not using any components that require rust yet. The vcpkg build ends up only adding in chromium's zlib fork as a dependency. But that might be down to some quirks of WebKit's CMake files for Angle. I could see us needing one of those json libs as well.

JanKoudijs: Thanks for this helpful background info!

Do you see value in moving on with this PR?
The list of dependencies might make it easier to strip down the build, compared to the current state.
There is still some cleanup that has to be done before the actual merge.

ADKaster: I'm poking at this myself now, and it looks like ... some of this is reasonable for pulling into master. I'd rather not whip up a CMake build for this, so trying to integrate with the existing gn build seems prudent. I mean we could try to Bazel it up as well, but I know next to nothing about Bazel (besides whatever concepts it inherits from being a modern build system in a world where CMake/meson/gn etc exist).

I think that a closer look at the `//build_overrides` folder will be key to keeping the sources list from exploding. At the very least, the sysroots are not needed, as there's a `use_sysroot` flag we can turn off.

When angle is included by skia, they use a flag called `angle_has_build`. Setting this to false disables the dependency on chromium's //build and //testing folders. However, actually ditching those requires coming up with a custom `//.gn` and `BUILDCONFIG.gn` for the project, which while appealing, is a lot of work/boilerplate to replace.

JanKoudijs: > so trying to integrate with the existing gn build seems prudent.

As a side step, I'm looking into the gn build on Linux (see `#build-problems`). Have to see what remains after the directory structure is lined up again. If I succeed, it will be a separate PR (and maybe a prerequisite for this one?).

JanKoudijs: > I think that a closer look at the //build_overrides folder will be key to keeping the sources list from exploding.

I did a cleanup on the sources list in https://github.com/LadybirdBrowser/ladybird/pull/5500/commits/7f9b0163b6b1ea1f93eb696642226f0372d1e8d5. Entries like dawn, llvm, rust, SwiftShader and more are gone now.

ADKaster: Your cleanup looks like a step in the right direction. Here's my take on further cleanup, including yeeting the sysroots, and yeeting the custom clang. <https://github.com/ADKaster/ladybird-browser/commit/220f1fc0d8fb04269e809451b13cd4e7f7ec07f5>

I suspect we can probably get rid of more dependencies, but something close to what's in this branch, fixup-squashed with all the angle-specific files in an `angle` subfolder of Meta/CMake/flatpak would probably be enough to get us a workable offline build

JanKoudijs: > fixup-squashed with all the angle-specific files in an angle subfolder of Meta/CMake/flatpak would probably be enough to get us a workable offline build

Done. I will close this thread and get the PR out of draft, to have the finishing off in the review.

ADKaster: oh, hold up, I did that work already :sweat_smile:

ADKaster: ah, you did a squash. hm. I have a branch with several commits chunked up

JanKoudijs: I included your commit, as referenced above, and indeed squashed. Did I miss some of your work?

ADKaster: nah, i've been working in tandem and forgot to comment in the thread here

Meta/CMake/flatpak/angle/angle-configure.sh
```diff
@@ -2,10 +2,12 @@
 
 set -e
 
-export PATH=$PWD/depot_tools:$PATH
 cd angle
 
 gn gen out --args='
+    cc="gcc"
+    cxx="g++"
```
JanKoudijs: These 2 have no effect according to `gn gen` output:

```
sh-5.2$ ./angle-configure.sh 
WARNING at the command-line "--args":2:8: Build argument has no effect.
    cc="gcc"
       ^----
The variable "cc" was set as a build argument
but never appeared in a declare_args() block in any buildfile.

To view all possible args, run "gn args --list <out_dir>"

The build continued as if that argument was unspecified.
```

ADKaster: yes, that's true. the toolchain()s defined in //build/toolchain/linux don't  have that as a user-configurable toolchain_arg.  I've got another set of diffs to make things work for aarch64, since the default `//build/toolchain/linux:arm64` toolchain sets a 'toolchain prefix of `aarch64-linux-gnu-` to all the tools. The flatpak SDK we're using doesn't prefix its tools like that in the aarch64 case. iirc the chromium build bots like to cross-compile every target from x64 linux -- even windows and macOS builds.

JanKoudijs: > I've got another set of diffs to make things work for aarch64, since the default //build/toolchain/linux:arm64 toolchain sets a 'toolchain prefix of aarch64-linux-gnu- to all the tools.

Just wanted to ask you about that. As I wondered what would happen if I build the flatpak on my Raspberry Pi 4  2GB. As a temporary workaround I created some symbolic links:

```
      "build-commands": [
        "ln -s /usr/bin/ar aarch64-linux-gnu-ar",
        "ln -s /usr/bin/g++ aarch64-linux-gnu-g++",
        "ln -s /usr/bin/gcc aarch64-linux-gnu-gcc",
        "ln -s /usr/bin/nm aarch64-linux-gnu-nm",
        "ln -s /usr/bin/readelf aarch64-linux-gnu-readelf",
        "./angle-configure.sh",
        "PATH=$PATH:/run/build/angle ./angle-build.sh",
        "./angle-install.sh"
      ],
```

Just a for fun attempt, and a training in patience :)
The Uptime is the time this exercise already took:

image

ADKaster: I've been trying to get a github actions build going with my branch... current problem is it keeps running out of disk space on the default runners :sweat_smile:  "14GiB minimum" is not that much in the grand scheme of things.

But the gist is to drop a BUILD.gn with 
```
import("//build/toolchain/gcc_toolchain.gni")

# This is a copy of the arm64 toolchain from //build/toolchain/linux/BUILD.gn,
# but without a toolprefix for non-cross builds for arm64 linux.
gcc_toolchain("arm64") {
   ...
```

somewhere and pass it via gn args `custom_toolchain` and `host_toolchain`.

JanKoudijs: If this is specific for the flatpak build, then we can also have a patch entry in the manifest:

**angle/build-arm64.patch**
```patch
diff --git a/toolchain/linux/BUILD.gn b/toolchain/linux/BUILD.gn
index a6e20a05c..ecb40361b 100644
--- a/toolchain/linux/BUILD.gn
+++ b/toolchain/linux/BUILD.gn
@@ -31,8 +31,6 @@ clang_toolchain("clang_arm64") {
 }
 
 gcc_toolchain("arm64") {
-  toolprefix = "aarch64-linux-gnu-"
-
   cc = "${toolprefix}gcc"
   cxx = "${toolprefix}g++"

```

**org.ladybird.Ladybird.json**
```json
        {
          "type": "patch",
          "path": "angle/build-arm64.patch",
          "dest": "angle/build"
        },
```

ADKaster: That makes a lot more sense. I've pushed a reorg of the changes here from my tree to this branch. 

I've got a CI job running here with the same changes (minus some git-fu) that seems promising

https://github.com/ADKaster/ladybird-browser/actions/runs/16479144268

ADKaster (approved):

PR #5501, opened 2025-07-18
LibWeb: Stub CSSCounterStyleRule

gmta:
Just the interface. Adds 6 new WPT subtest passes.


AtkinsSJ:
Looks good, but you need to rebaseline the all-properties test.

PR #5505, opened 2025-07-18
LibWeb: Implement `InputEvent.getTargetRanges()`

tcl3:
This returns a list of ranges that would be affected by a change to the DOM if the input event is not cancelled.

This gets us +930 WPT subtest passes in the `input-events` directory. Unfortunately, these tests are not importable because they rely on testdriver. I've made a test that uses `internals` instead.

### DIFF
--- a/Libraries/LibWeb/UIEvents/InputEvent.cpp
+++ b/Libraries/LibWeb/UIEvents/InputEvent.cpp
@@ -12,9 +12,9 @@ namespace Web::UIEvents {
 
 GC_DEFINE_ALLOCATOR(InputEvent);
 
-GC::Ref<InputEvent> InputEvent::create_from_platform_event(JS::Realm& realm, FlyString const& event_name, InputEventInit const& event_init)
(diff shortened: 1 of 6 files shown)


Libraries/LibWeb/UIEvents/InputEvent.cpp
```diff
@@ -43,10 +44,19 @@ void InputEvent::initialize(JS::Realm& realm)
...
+void InputEvent::visit_edges(Visitor& visitor)
 {
-    dbgln("FIXME: Implement InputEvent::get_target_ranges()");
-    return {};
+    Base::visit_edges(visitor);
+    visitor.visit(m_target_ranges);
+}
+
+// https://w3c.github.io/input-events/#dom-inputevent-gettargetranges
+Vector<GC::Ref<DOM::StaticRange>> InputEvent::get_target_ranges() const
```
shannonbooth: This vector of GC ref usages look like they could be misused since a copy would not take any ref. Would be a bit safer if this was either a RootVector if we did need copies or maybe a span (I do wonder what the best pattern is)

tcl3: I've updated `get_target_ranges()` to return a `ReadonlySpan` instead. There are probably other places in the codebase where we could do this.

PR #5507, opened 2025-07-18
 LibWeb/CSS: Add `CSS.registerProperty` JS method 

Norbiros:
This adds an *almost* complete implementation of `CSS.registerProperty`
method enabling further progress on the `@property` feature.

I also imported some WPT tests related to that.

In the next PR(s), I'll add `<syntax>` support to `@property,` handle the `inherit` flag, and work on a more complete implementation.

### DIFF
--- a/Tests/LibWeb/Text/expected/wpt-import/css/css-properties-values-api/register-property.txt
+++ b/Tests/LibWeb/Text/expected/wpt-import/css/css-properties-values-api/register-property.txt
@@ -1,0 +1,12 @@
+Harness status: OK
+
+Found 6 tests
+
+5 Pass
+1 Fail
+Pass	registerProperty requires a Dictionary type
+Pass	registerProperty requires a name matching <custom-property-name>
+Pass	registerProperty only allows omitting initialValue if syntax is '*'
+Pass	registerProperty fails for an already registered property
+Pass	registerProperty requires inherits
+Fail	Registering a property should not cause a transition
--- a/Tests/LibWeb/Text/expected/wpt-import/css/css-properties-values-api/registered-properties-inheritance.txt
+++ b/Tests/LibWeb/Text/expected/wpt-import/css/css-properties-values-api/registered-properties-inheritance.txt
@@ -1,0 +1,14 @@
+Harness status: OK
+
+Found 8 tests
+
+2 Pass
+6 Fail
+Fail	Registered properties are correctly inherited (or not) depending on the inherits flag.
+Fail	Explicitly inheriting from a parent with an invalid value results in initial value.
+Pass	Explicitly inheriting from a parent with no value results in initial value.
+Pass	Explicitly inheriting from a parent with a value results in that value.
+Fail	Reference to undefined variable results in inherited value
+Fail	Reference to syntax-incompatible variable results in inherited value
+Fail	Font-relative units are absolutized before before inheritance
+Fail	Calc expressions are resolved before inheritance
--- a/Tests/LibWeb/Text/expected/wpt-import/css/css-properties-values-api/var-reference-registered-properties-cycles.txt
+++ b/Tests/LibWeb/Text/expected/wpt-import/css/css-properties-values-api/var-reference-registered-properties-cycles.txt
@@ -1,0 +1,11 @@
+Harness status: OK
+
+Found 5 tests
+
+2 Pass
+3 Fail
+Fail	A var() cycle between two registered properties is handled correctly.
+Fail	A var() cycle between a registered properties and an unregistered property is handled correctly.
+Fail	A var() cycle between a two unregistered properties is handled correctly.
+Pass	A var() cycle between a syntax:'*' property and an unregistered property is handled correctly.
+Pass	Custom properties with universal syntax become guaranteed-invalid when invalid at computed-value time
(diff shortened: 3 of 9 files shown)


Libraries/LibWeb/CSS/CSS.cpp
```diff
@@ -8,9 +8,14 @@
...
 #include <LibWeb/CSS/CSS.h>
 #include <LibWeb/CSS/Parser/Parser.h>
+#include <LibWeb/CSS/Parser/Syntax.h>
+#include <LibWeb/CSS/Parser/SyntaxParsing.h>
 #include <LibWeb/CSS/PropertyID.h>
 #include <LibWeb/CSS/PropertyName.h>
 #include <LibWeb/CSS/Serialize.h>
+#include <LibWeb/HTML/Window.h>
+
+using Web::CSS::Parser::SyntaxNode;
```
AtkinsSJ: We avoid `using`, and I don't think you're actually using this.

Libraries/LibWeb/CSS/CSS.cpp
```diff
@@ -59,4 +64,82 @@ WebIDL::ExceptionOr<bool> supports(JS::VM& vm, StringView condition_text)
...
+    auto& window = static_cast<Web::HTML::Window&>(realm.global_object());
+    auto& document = window.associated_document();
+
+    // 2. If name is not a custom property name string, throw a SyntaxError and exit this algorithm.
+    if (!is_a_custom_property_name_string(definition.name))
+        return WebIDL::SyntaxError::create(realm, "Invalid property name"_string);
+
+    // 3. If property set already contains an entry with name as its property name (compared codepoint-wise),
+    //    throw an InvalidModificationError and exit this algorithm.
+    if (document.registered_custom_properties().get(definition.name).has_value())
```
AtkinsSJ: `contains(definition.name)` is a little clearer than `get(definition.name).has_value()`.

Libraries/LibWeb/CSS/CSS.cpp
```diff
@@ -59,4 +64,82 @@ WebIDL::ExceptionOr<bool> supports(JS::VM& vm, StringView condition_text)
...
+        return WebIDL::InvalidModificationError::create(realm, "Property already registered"_string);
+
+    // 4. Attempt to consume a syntax definition from syntax. If it returns failure, throw a SyntaxError.
+    //    Otherwise, let syntax definition be the returned syntax definition.
+    auto syntax_component_values = parse_component_values_list(CSS::Parser::ParsingParams {}, definition.syntax);
```
AtkinsSJ: Pass `document` to the `ParsingParams` constructor, here and below. (Maybe create the `ParsingParams` separately too and reuse it.) I can't remember exactly what the parser uses it for, but we have a document so it's better to let it use it.

Libraries/LibWeb/CSS/CSS.cpp
```diff
@@ -59,4 +64,82 @@ WebIDL::ExceptionOr<bool> supports(JS::VM& vm, StringView condition_text)
...
+    // 1. Let property set be the value of the current global object’s associated Document’s [[registeredPropertySet]] slot.
+    auto& realm = *vm.current_realm();
+    auto& window = static_cast<Web::HTML::Window&>(realm.global_object());
+    auto& document = window.associated_document();
+
+    // 2. If name is not a custom property name string, throw a SyntaxError and exit this algorithm.
+    if (!is_a_custom_property_name_string(definition.name))
+        return WebIDL::SyntaxError::create(realm, "Invalid property name"_string);
+
+    // 3. If property set already contains an entry with name as its property name (compared codepoint-wise),
```
AtkinsSJ: This is part of step 2

Libraries/LibWeb/CSS/CSS.cpp
```diff
@@ -59,4 +64,82 @@ WebIDL::ExceptionOr<bool> supports(JS::VM& vm, StringView condition_text)
...
+    //    throw an InvalidModificationError and exit this algorithm.
+    if (document.registered_custom_properties().get(definition.name).has_value())
+        return WebIDL::InvalidModificationError::create(realm, "Property already registered"_string);
+
+    // 4. Attempt to consume a syntax definition from syntax. If it returns failure, throw a SyntaxError.
```
AtkinsSJ: This is step 3

Libraries/LibWeb/CSS/CSS.cpp
```diff
@@ -59,4 +64,82 @@ WebIDL::ExceptionOr<bool> supports(JS::VM& vm, StringView condition_text)
...
+    //    Otherwise, let syntax definition be the returned syntax definition.
+    auto syntax_component_values = parse_component_values_list(CSS::Parser::ParsingParams {}, definition.syntax);
+    auto maybe_syntax = parse_as_syntax(syntax_component_values);
+    if (!maybe_syntax) {
+        return WebIDL::SyntaxError::create(realm, "Invalid syntax definition"_string);
+    }
+
+    RefPtr<CSSStyleValue const> initial_value_maybe;
+
+    // 4. Parse initialValue according to the following rules:
```
AtkinsSJ: "Parse initialValue according to the following rules:" doesn't seem to appear in the spec text, and generally these comments in step 4 seem to be modified a lot. Please include the spec text exactly as it's written. It makes it much easier to tell when our code is out of date compared to the current spec.

Issue #5494, opened 2025-07-17
`clip-path` doesn't work on SVG elements (except `<svg>`)


AtkinsSJ:
Reproduction:

```html
<!DOCTYPE html>
<style>
  .clipped {
    clip-path: circle(100px at top left);
  }
</style>
<svg class="clipped">
  <g>
    <rect width="100" height="100" fill="green"/>
  </g>
</svg>

<br>

<svg>
  <g class="clipped">
    <rect width="100" height="100" fill="green"/>
  </g>
</svg>

<br>

<svg>
  <g>
    <rect class="clipped" width="100" height="100" fill="green"/>
  </g>
</svg>
```

### Firefox

<img width="350" height="523" alt="Image" src="https://github.com/user-attachments/assets/dfb14b51-152d-4087-b2a0-2f334b3329f1" />

### Ladybird

<img width="965" height="742" alt="Image" src="https://github.com/user-attachments/assets/b4ea4b80-84ac-4f74-8d67-5ce5f71f16dc" />

shannonbooth:
Works fine on master:

<img width="1395" height="1220" alt="Image" src="https://github.com/user-attachments/assets/21704aa4-ce90-48fd-bf7f-5f76e5deff35" />

PR #5508, opened 2025-07-18
LibWeb: Implement stacking context rules for SVG elements

AtkinsSJ:
This is an attempt at solving #5494 which didn't work, but seems useful on its own.

`Screenshot/input/svg-foreign-object-mask.html` failed for me locally on macOS because of some minor edge-pixel differences. I've left it as-is to see how it goes for other OSes as maybe it needs fuzzy-matching.

### DIFF
--- a/Libraries/LibWeb/DOM/Node.h
+++ b/Libraries/LibWeb/DOM/Node.h
@@ -164,7 +164,6 @@ public:
     virtual bool is_shadow_root() const { return false; }
 
     virtual bool requires_svg_container() const { return false; }
-    virtual bool is_svg_container() const { return false; }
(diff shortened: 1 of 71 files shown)


kalenikaliaksandr:
previously we allowed svg boxes to establish a stacking context but the problem is that doesn't match with behavior of all other engiens. see 83b6bc4 for details (I made a mistake by not allowing <svg> to establish a stacking context and fixed that recently).

kalenikaliaksandr:
to solve https://github.com/LadybirdBrowser/ladybird/issues/5494 we would need to update `SVGSVGPaintable::paint_svg_box()` to take care of clip-path property.

PR #3495, opened 2025-02-07
LibWeb: Implement most `CredentialsContainer` methods 

devgianlu:
This PR mplements `get`, `store`, `create` on `CredentialsContainer`.

Notably, it also introduces the `CredentialInterface` abstract class to deal with credentials internal methods and metadata. It has been implemented as a singleton for each credential type. This is the design that I managed to come up with, any suggestion is appreciated.

The spec isn't the best and contains some bugs:
- The `options.signal && options.signal->aborted()` check in `get` and `create` is not deferred therefore fails two WPT tests
- The `interfaces.size() < 1` check in `create` is custom and avoids an out of bound access in the next step -> [Spec issue](https://github.com/w3c/webappsec-credential-management/issues/267)
- The `CredentialRequestOptions` IDL dictionary has a default value for `password` which renders the credential type always active. I think we are supposed to reject it when computing the [relevant credential interface objects](https://w3c.github.io/webappsec-credential-management/#credentialrequestoptions-relevant-credential-interface-objects) if `password` is `false`, but that's not specified

I'll create some spec issues and link them here even if the repository has been inactive for the last 6 months.

Lastly, I am unsure how to check `If settings’s relevant global object has no associated Document` so I've left two TODOs.

PR #5509, opened 2025-07-18
LibWeb: Start implementing Credential Management APIs

kennethmyhra:
This implements the following AOs:
- [Create a PasswordCredential from PasswordCredentialData.](https://www.w3.org/TR/credential-management-1/#abstract-opdef-create-a-passwordcredential-from-passwordcredentialdata)
- [Create a PasswordCredential from an HTMLFormElement.](https://www.w3.org/TR/credential-management-1/#abstract-opdef-create-a-passwordcredential-from-an-htmlformelement)

Which corresponds to these PasswordCredential ctors:
- constructor(PasswordCredentialData)
- constructor(HTMLFormElement)

The commits contains more information about the individual changes.

### DIFF
--- a/Libraries/LibWeb/CredentialManagement/FederatedCredential.h
+++ b/Libraries/LibWeb/CredentialManagement/FederatedCredential.h
@@ -10,10 +10,13 @@
 #include <LibWeb/Bindings/FederatedCredentialPrototype.h>
 #include <LibWeb/Bindings/PlatformObject.h>
 #include <LibWeb/CredentialManagement/Credential.h>
+#include <LibWeb/CredentialManagement/CredentialUserData.h>
 
 namespace Web::CredentialManagement {
 
-class FederatedCredential final : public Credential {
+class FederatedCredential final
+    : public Credential
+    , public CredentialUserData {
     WEB_PLATFORM_OBJECT(FederatedCredential, Credential);
     GC_DECLARE_ALLOCATOR(FederatedCredential);
 
--- a/Libraries/LibWeb/CredentialManagement/PasswordCredential.cpp
+++ b/Libraries/LibWeb/CredentialManagement/PasswordCredential.cpp
@@ -5,35 +5,43 @@
  */
 
 #include <LibWeb/CredentialManagement/PasswordCredential.h>
+#include <LibWeb/CredentialManagement/PasswordCredentialOperations.h>
 
 namespace Web::CredentialManagement {
 
 GC_DEFINE_ALLOCATOR(PasswordCredential);
 
-GC::Ref<PasswordCredential> PasswordCredential::create(JS::Realm& realm)
-{
-    return realm.create<PasswordCredential>(realm);
-}
-
 // https://www.w3.org/TR/credential-management-1/#dom-passwordcredential-passwordcredential
-WebIDL::ExceptionOr<GC::Ref<PasswordCredential>> PasswordCredential::construct_impl(JS::Realm& realm, HTML::HTMLFormElement const&)
+WebIDL::ExceptionOr<GC::Ref<PasswordCredential>> PasswordCredential::construct_impl(JS::Realm& realm, GC::Ptr<HTML::HTMLFormElement> form)
 {
-    return realm.vm().throw_completion<JS::InternalError>(JS::ErrorType::NotImplemented, "construct"sv);
+    // 1. Let origin be the current settings object's origin.
+    auto origin = HTML::current_principal_settings_object().origin();
+
+    // 2. Let r be the result of executing Create a PasswordCredential from an HTMLFormElement given form and origin.
+    // 3. If r is an exception, throw r. Otherwise, return r.
+    return create_password_credential(realm, form, origin);
 }
 
 // https://www.w3.org/TR/credential-management-1/#dom-passwordcredential-passwordcredential-data
-WebIDL::ExceptionOr<GC::Ref<PasswordCredential>> PasswordCredential::construct_impl(JS::Realm& realm, PasswordCredentialData const&)
+WebIDL::ExceptionOr<GC::Ref<PasswordCredential>> PasswordCredential::construct_impl(JS::Realm& realm, PasswordCredentialData const& options)
 {
-    return realm.vm().throw_completion<JS::InternalError>(JS::ErrorType::NotImplemented, "construct"sv);
+    // 1. Let r be the result of executing Create a PasswordCredential from PasswordCredentialData on data.
+    // 2. If r is an exception, throw r.
+    return create_password_credential(realm, options);
 }
 
 PasswordCredential::~PasswordCredential()
 {
 }
 
-PasswordCredential::PasswordCredential(JS::Realm& realm)
+PasswordCredential::PasswordCredential(JS::Realm& realm, PasswordCredentialData const& data)
     : Credential(realm)
+    , m_password(data.password)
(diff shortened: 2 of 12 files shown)


devgianlu:
There have been multiple attempts to get this implemented, but it doesn't seem that there's much interest in it (understandbly). See #3495 and https://github.com/LadybirdBrowser/ladybird/compare/master...devgianlu:ladybird:webauthn.

kennethmyhra:
@devgianlu posted a follow-up here: https://discord.com/channels/1247070541085671459/1256703242679746560/1401202508935991388

I noticed #3495 has unresolved conversations which usually is the case for PRs going stale. There is also a conversation marked as resolved, but the requested change is not implemented https://github.com/LadybirdBrowser/ladybird/pull/3495#discussion_r2038651562.

devgianlu:
All the unresolved stuff was there mainly because of open discussion between the maintainers. The other stuff I left there waiting for feedback on the other points because they required major refactoring.

Admittedly, I didn't follow-up on those issues, but I interpreted the long review times as low-interest and therefore I let it go stale.

Just wanted to point out that there's already some code written for this so that we don't waste too much time reimplementing the same stuff over and over 😅.

Libraries/LibWeb/CredentialManagement/PasswordCredentialOperations.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+    // 4. Let elements be a list of all the submittable elements whose form owner is form, in tree order.
+    auto elements = form->get_submittable_elements();
+
+    // 5. Let newPasswordObserved be false.
+    bool new_password_observed = false;
+
+    // 6. For each field in elements, run the following steps:
+    for (auto const& field : elements) {
+        dbgln("field->name(): {}", field->name());
```
Lubrsi: issue: Leftover dbgln

Libraries/LibWeb/CredentialManagement/PasswordCredentialOperations.cpp
```diff
@@ -0,0 +1,136 @@
...
+                        auto id = username.get<String>();
+                        dbgln("username: {}", id);
+                        data.id = id;
+                    }
+                }
+            }
+        }
+    }
+
+    dbgln("Done iterating over tokens");
```
Lubrsi: issue: Leftover dbgln

Libraries/LibWeb/CredentialManagement/PasswordCredentialOperations.cpp
```diff
@@ -0,0 +1,136 @@
...
+                if (token.equals_ignoring_ascii_case("nickname"sv)) {
+                    if (auto nickname = form_data->get(name.value()); nickname.has<String>())
+                        data.name = nickname.get<String>();
+                }
+                //    - "username"
+                //      Set data’s id member’s value to the result of executing formData’s get() method on name.
+                if (token.equals_ignoring_ascii_case("username"sv)) {
+                    if (auto username = form_data->get(name.value()); username.has<String>()) {
+                        auto id = username.get<String>();
+                        dbgln("username: {}", id);
```
Lubrsi: issue: Leftover dbgln

Libraries/LibWeb/CredentialManagement/PasswordCredentialOperations.cpp
```diff
@@ -0,0 +1,136 @@
...
+            continue;
+
+        // 2. Let name be the value of field’s name attribute.
+        // 3. If formData’s has() method returns false when executed on name, then skip to the next field.
+        auto name = field->attribute(HTML::AttributeNames::name);
+        if (!name.has_value() || !form_data->has(name.value()))
+            continue;
+
+        // 4. If field’s autocomplete attribute’s value contains one or more autofill detail tokens (tokens), then:
+        if (auto tokens = field->attribute(HTML::AttributeNames::autocomplete); tokens.has_value()) {
```
Lubrsi: suggestion: This is always true because the field is skipped by step 1 if there is no autocomplete attribute. It would be nice to have the check for `one or more autofill detail tokens`, but it's non-blocking since the following code works on a subset of known good tokens.

Libraries/LibWeb/CredentialManagement/PasswordCredential.cpp
```diff
@@ -5,35 +5,43 @@
...
 
-PasswordCredential::PasswordCredential(JS::Realm& realm)
+PasswordCredential::PasswordCredential(JS::Realm& realm, PasswordCredentialData const& data)
     : Credential(realm)
+    , m_password(data.password)
+    , m_origin(data.origin)
 {
+    m_id = data.id;
+    m_icon_url = data.icon_url.value_or(String {});
+    m_name = data.name.value_or(String {});
```
Lubrsi: nit: These could be initialised in the same manner as `m_password` and `m_origin` above.

Lubrsi (changes requested):
Looks good! Just a few comments.

Libraries/LibWeb/CredentialManagement/PasswordCredentialOperations.cpp
kennethmyhra: Removed the if-statement and left step 4 as a FIXME for now.

Libraries/LibWeb/CredentialManagement/PasswordCredential.cpp
kennethmyhra: They can't be initialized in the same manner as they are not a member of the initialized class. I can move them to the constructors of the class they are member of though if you want me to, but that would have to be in a separate commit to avoid going through unnecessary hoops before the implementation of the constructor of `FederatedCredential` commit.

lukewarlow:
Personally I recommend against implementing FederatedCredential. It's effectively deprecated. FedCM is the API that's actually worth implementing. That uses IdentityCredential iirc.

kennethmyhra:
Open to removing `FederatedCredential` completely. The note on FedCM @lukewarlow refers to can be found on this page: https://developer.mozilla.org/en-US/docs/Web/API/FederatedCredential

lukewarlow:
https://chromestatus.com/metrics/feature/timeline/popularity/3992 - I believe this is the relevant usage of it too and it's dropped off a cliff. I'm trying to get it removed from the spec even.

kennethmyhra:
That graph, if correct, probably says it all for `FederatedCredential`. Thanks for chiming in on this, Luke

Libraries/LibWeb/CredentialManagement/FederatedCredential.h
```diff
@@ -37,6 +38,9 @@ class FederatedCredential final
 
     String m_provider;
     Optional<String> m_protocol;
+
+    // https://www.w3.org/TR/credential-management-1/#dom-credential-origin-slot
+    URL::Origin m_origin;
```
shannonbooth, line 42: Origin does not have a default constructor so I think this should probably be squashed into the next commit to compile? Otherwise rest LGTM

kennethmyhra: Thanks! Squashed `Origin` into the two next relevant commits. Also made sure that each commit compiles + tests running green.

kennethmyhra:
CI failure on Linux seems unrelated:
```
Fail: Text/input/GamepadAPI/gamepad-rumble.html
```

PR #5511, opened 2025-07-18
LibWeb/CSS: Parse full `mask` property and its longhands

InvalidUsernameException:
Only parsing, no actual masking functionality is implemented in here.

This gives >400 new WPT subtest passes. Remaining test failures in the imported tests are due to:
* Missing support for `mask-border-*` properties.
* Incorrect behavior of some of the existing data types used by the `mask` properties.


Libraries/LibWeb/CSS/Properties.json
```diff
@@ -2390,6 +2418,65 @@
...
+    ],
+    "initial": "border-box"
+  },
+  "mask-position": {
+    "animation-type": "repeatable-list",
+    "affects-layout": false,
+    "inherited": false,
+    "initial": "0% 0%",
+    "valid-types": [
+      "background-position"
```
Calme1709: It would be nice if we could rename `background-position` to `position` as it is in the spec - akin to what you did for `background-repeat` -> `repeat`. Not necessary for this PR so a FIXME would be enough IMO

InvalidUsernameException: Yeah, I was seeing what looked like some other naming inconsistencies. However I only glanced at the spec and then decided not to look into it further since things seems to be inconsistent there as well.

Taking a closer look at this case now, it turns out this was just wrong. `background-position` and `position` are two separate types, the latter's valid values being a subset of the former's. And there even is support for this in the codebase, so I just changed the `valid-types` to `position` here.

This change doesn't have any behavioral effect though. The parsing code is overwritten in the large `switch` statement in `Parser::parse_css_value()` to support the comma separated list. And the call there was using the right variant of the `position`-parsing function already.

Calme1709: Looks like I also missed that `background-position` was a distinct type as well.

Just as an FYI - there is a behavioral change here in that we now disallow those `background-position` valid values within the `mask` shorthand, this is because the calls to `parse_css_value_for_properties` works on `valid-types`, bypassing the `parse_a_css_value` switch statement.

Looks good to me now.

Libraries/LibWeb/CSS/Properties.json
```diff
@@ -2432,6 +2432,23 @@
...
+    "valid-types": [
+      "length [0,∞]",
+      "percentage [0,∞]"
+    ],
+    "valid-identifiers": [
+      "auto",
+      "cover",
+      "contain"
+    ],
+    "percentages-resolve-to": "length"
```
AtkinsSJ, line 2484: Not a change to make in this PR, but we could really do with `<bg-size>` being a properly supported type instead of having to specify all its requirements inside `Properties.json`.

Libraries/LibWeb/CSS/Properties.json
```diff
@@ -2370,10 +2370,17 @@
   },
   "mask": {
     "inherited": false,
-    "__comment": "FIXME: add longhands mask-clip, mask-composite, mask-mode, mask-origin, mask-position, mask-repeat, mask-size; also reset mask-border",
```
AtkinsSJ, line 2378: Can you keep the note about `mask-border`?

Libraries/LibWeb/CSS/StyleValues/ShorthandStyleValue.cpp
```diff
@@ -489,6 +489,109 @@ String ShorthandStyleValue::to_string(SerializationMode mode) const
...
         return MUST(String::formatted("{} / {}", start->to_string(mode), end->to_string(mode)));
     }
+    case PropertyID::Mask: {
+        StringBuilder builder;
+
+        auto serialize_layer = [mode, &builder](String image_value_string, String position_value_string, String size_value_string, String repeat_value_string, String origin_value_string, String clip_value_string, String composite_value_string, String mode_value_string) {
+            PropertyID canonical_property_order[] = {
+                PropertyID::MaskImage,
+                PropertyID::MaskPosition,
+                // Intentioanlly skipping MaskSize here, it is handled together with MaskPosition.
```
AtkinsSJ: Nit: "Intentionally"

Libraries/LibWeb/CSS/StyleValues/ShorthandStyleValue.cpp
```diff
@@ -489,6 +489,109 @@ String ShorthandStyleValue::to_string(SerializationMode mode) const
...
+            };
+
+            bool layer_is_empty = true;
+            for (size_t i = 0; i < array_size(canonical_property_order); i++) {
+                auto property = canonical_property_order[i];
+                auto const& value = property_value(property);
+
+                if (is_initial_value(property)) {
+                    if (property != PropertyID::MaskOrigin)
+                        continue;
```
AtkinsSJ: I think this will skip serializing `mask-position` if it's the initial value - but that's only safe to do if `mask-size` is also the initial value.

AtkinsSJ (changes requested):
Thanks for tackling this! Looks good except for one small bug and a couple of nits.

Libraries/LibWeb/CSS/Properties.json
InvalidUsernameException: That's what I get for not reading the whole line. Didn't even see there was more than just the list of longhands.

Libraries/LibWeb/CSS/StyleValues/ShorthandStyleValue.cpp
InvalidUsernameException: Good catch. One more edge case. With the two I stumbled across during development and the one that came out of Calme1709's comment that makes four overall now. I guess it is time for a follow-up PR to the WPT repo since none of that is covered very well by the tests.

Note that this change caused some animation tests to fail now. But I don't think that is a regression. It looks like the incorrect serialization of the mask property was masking (no pun intended) a different serialization bug. 

While fixing this I noticed that `auto` values for background size were serialized as `auto auto`, which, based on my understanding, is incorrect because it is not minimal. Since that was a straightforward fix that affected the mask-tests I've included an extra commit for it here.

Libraries/LibWeb/CSS/Properties.json
InvalidUsernameException: Resolving this with no action then if I understood correctly.

InvalidUsernameException:
Relevant changes from addressing the review comments are in the second force-push, which unfortunately isn't visible anymore in the PR history, but corresponds to [this diff](https://github.com/LadybirdBrowser/ladybird/compare/cfc41c47078243004e1b058fd195bbbb1bb88f37..efc066fb795235fa9336953a5f58ca70b5299090).

The others are a rebase and several flaky CI runs.

InvalidUsernameException:
I've dropped the aforementioned change to background-size serialization in the last rebase again because the same change was already done in #5677.

Rebasing this on `master` also regressed 5 subtests from the tests imported for this PR:

https://github.com/LadybirdBrowser/ladybird/blob/ff529d11964dd0681a6ca6e0f4bc74f1d00f3fdb/Tests/LibWeb/Text/expected/wpt-import/css/css-masking/parsing/mask-valid.sub.txt#L53-L61

Before the rebase all 9 of these tests passed successfully. This is caused by 4c91b42dd27c14409dfb37f68e038a34d01a5eeb, reverting that commit locally fixes those tests again. @tcl3 Do you know what is going on here? Your change seems sensible to me, but the tests expect the repeated values _not_ to be collapsed into a single one for the longhands. And the three major engines [all pass these tests](https://wpt.fyi/results/css/css-masking/parsing/mask-valid.sub.html?label=experimental&label=master&product=ladybird&product=chrome&product=firefox&product=safari&product=flow&product=servo).

tcl3:
The shortest serialization principle I tried to follow in my PR comes from this part of the CSS specification: https://drafts.csswg.org/cssom/#serialize-a-css-value

It says:
> If component values can be omitted or replaced with a shorter representation without changing the meaning of the value, omit/replace them.

It looks like shortening repeated values in a shorthand if they occur across multiple longhands isn't valid because it would change the meaning of the value.

In addition to the tests you've highlighted, it looks like I also introduced a few regressions in: http://wpt.live/css/css-backgrounds/parsing/background-valid.html, which fails for the same reason.

One thing we could do is only shorten repeated list values if they are separated by a space (not a comma), this fixes 5 of the regressed mask subtests and all of the regressed background subtests. It doesn't seem to introduce any regressions in other parsing tests -  it does seem a bit hacky though.

Just for completeness, the change I made was (I'm not sure whether doing this is most correct fix though):
```diff
diff --git a/Libraries/LibWeb/CSS/StyleValues/StyleValueList.cpp b/Libraries/LibWeb/CSS/StyleValues/StyleValueList.cpp
index 94351b656c7..c8d3c5aba70 100644
--- a/Libraries/LibWeb/CSS/StyleValues/StyleValueList.cpp
+++ b/Libraries/LibWeb/CSS/StyleValues/StyleValueList.cpp
@@ -35,7 +35,7 @@ String StyleValueList::to_string(SerializationMode mode) const
     }

     auto first_value = m_properties.values.first();
-    if (all_of(m_properties.values, [&](auto const& property) { return property == first_value; }))
+    if (m_properties.separator == Separator::Space && all_of(m_properties.values, [&](auto const& property) { return property == first_value; }))
         return first_value->to_string(mode);

     StringBuilder builder;
```

InvalidUsernameException:
> It looks like shortening repeated values in a shorthand if they occur across multiple longhands isn't valid because it would change the meaning of the value.

In what scenario exactly? Both background- and mask-spec say that say that missing values are filled by repeating the existing ones. Changing `foo, foo` to `foo` just means that the single value needs to be repeated more often when using the properties for rendering. Is there an edge case I am not seeing?

> The background of a box can have multiple background image layers. **The number of layers is determined by the number of comma-separated values in the background-image property.** Note that a value of none still creates a layer.
>
> Each of the background images is sized, positioned, and tiled according to the corresponding value in the other background properties. The lists are matched up from the first value: excess values at the end are not used. **If a property doesn’t have enough comma-separated values to match the number of layers, the UA must calculate its used value by repeating the list of values until there are enough.**

https://drafts.csswg.org/css-backgrounds-3/#layering

----

> this fixes 5 of the regressed mask subtests

That is all of the regressed mask subtests ;)

AtkinsSJ:
> Changing `foo, foo` to `foo` just means that the single value needs to be repeated more often when using the properties for rendering. Is there an edge case I am not seeing?

It's part of what you quoted: For `background`, the number of layers is determined by how many layers are given for `background-image`, so if you deduplicated `background-image: foo, foo, foo;`, it would change the effect.

For the other properties that don't determine the layer count, they wouldn't have an effect, but I guess they're keep as-is for consistency.

InvalidUsernameException:
> so if you deduplicated `background-image: foo, foo, foo;`, it would change the effect.

Oh yeah, right. For that one specifically it matters.

Anyway, since this is a regression unrelated to the changes in this PR, I think this is ready for a second review.

Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
```diff
@@ -3497,6 +3501,244 @@ RefPtr<CSSStyleValue const> Parser::parse_list_style_value(TokenStream<Component
...
+        // If one <geometry-box> value and the no-clip keyword are present then <geometry-box> sets mask-origin and
+        // no-clip sets mask-clip to that value.
+        // If one <geometry-box> value and no no-clip keyword are present then <geometry-box> sets both mask-origin and
+        // mask-clip to that value.
+        // If two <geometry-box> values are present, then the first sets mask-origin and the second mask-clip.
+        case PropertyID::MaskOrigin:
+        case PropertyID::MaskClip: {
+            VERIFY(value->is_keyword());
+            if (value->as_keyword().keyword() == Keyword::NoClip) {
+                VERIFY(!mask_clip);
```
AtkinsSJ, line 3666: This VERIFY() isn't safe. If we have `no-clip` occur twice, we'll crash, when we should just return nullptr.

InvalidUsernameException: Hmm, I think that case should already be filtered above. On the second occurrence of `mask-clip` we will call `parse_css_value_for_properties()` with a property list that won't contain `PropertyID::MaskClip` anymore. So that function should just return a none-value because `no-clip` is not a valid keyword for the parameters it is called with. And so we should have returned `nullptr` already because the `has_value()`-check on the optional fails.

A couple test cases I just wrote around this case refuses to parse for me as expected. What is the test case that crashes for you?

AtkinsSJ: Ah no you're absolutely right! It's fine as-is. (I didn't test and find that, I just concluded it could happen, apparently without thinking hard enough. :sweat_smile: )

AtkinsSJ (approved):

Issue #840, opened 2024-07-26
LibWeb: Certain nodes not treated correctly when parsing an XHTML document


aescarias:
It seems like the XHTML document parser does not include comments and processing instructions which are valid nodes and hence should be in `Node.childNodes`. It also seems to treat CDATA sections as Text nodes when that shouldn't be the case.

Considering the following document (making sure the file extension is `.xhtml`):

```xhtml
<html lang="en" xmlns="http://www.w3.org/1999/xhtml">
    <head>
        <title>Document</title>
    </head>
    <body>
        <div id="elem"><![CDATA[beep]]><?test test?><!-- I am a comment --><p>Hello</p><p>World</p></div>

        <script>
            console.log(elem.childNodes.length)
            console.log(`${elem.childNodes[0].toString()}, ${elem.childNodes[0].nodeType}`)
            console.log(`${elem.childNodes[1].toString()}, ${elem.childNodes[2].toString()}`)
        </script>
    </body>
</html>
```

On Firefox 128 (Windows), this outputs:
```
5
[object CDATASection], 4
[object ProcessingInstruction], [object Comment]
```

On Ladybird commit [`cdb9081`](https://github.com/LadybirdBrowser/ladybird/tree/cdb9081a8690f021c438f296cde5c57ae56ad606) (WSL2/Ubuntu 24.04), however, this outputs:
```
3
[object Text], 3
[object HTMLParagraphElement], [object HTMLParagraphElement]
```

PR #5512, opened 2025-07-18
LibXML: Add parser hooks for CDATASection and ProcessingInstructions

ADKaster:
This allows listeners to be notified when a CDATASection or
ProcessingInstruction is encountered during parsing. The non-listener
path still has the incorrect behavior of silently treating CDATASection
as Text nodes, but this allows listeners to handle them correctly.

Fixes: https://github.com/LadybirdBrowser/ladybird/issues/840

### DIFF
--- a/Libraries/LibWeb/XML/XMLDocumentBuilder.cpp
+++ b/Libraries/LibWeb/XML/XMLDocumentBuilder.cpp
@@ -4,8 +4,10 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
+#include <LibWeb/DOM/CDATASection.h>
 #include <LibWeb/DOM/DocumentType.h>
 #include <LibWeb/DOM/Event.h>
+#include <LibWeb/DOM/ProcessingInstruction.h>
 #include <LibWeb/HTML/HTMLTemplateElement.h>
 #include <LibWeb/HTML/Window.h>
 #include <LibWeb/HighResolutionTime/TimeOrigin.h>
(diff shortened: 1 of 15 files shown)


tcl3:
It might be worth importing http://wpt.live/dom/nodes/Node-lookupPrefix.xhtml, which is fixed by this change.

http://wpt.live/dom/nodes/Comment-in-doctype.xhtml now fails. I'm not sure how easy it is to fix, or whether it's really in the scope of this PR, since the reason it worked before is that comments just didn't show up at all.

ADKaster:
Given the explanation in <https://github.com/web-platform-tests/wpt/pull/38136> it sounds like this is a separate behavior...ish? It will require extra code in LibXML to ensure that while inside the doctype, we suppress comments and add PIs as its children.

ADKaster:
yeah I think properly handling Comment-in-doctype.xhtml will require updating `parse_markup_declaration` 

```c++
    if (auto result = parse_processing_instruction(); !result.is_error()) {
        rollback.disarm();
        return Optional<MarkupDeclaration> {};
    }
    if (auto result = parse_comment(); !result.is_error()) {
        rollback.disarm();
        return Optional<MarkupDeclaration> {};
    }
 ```
 
 Such that it actually passes PIs and Comments as decls for the calling code to determine what to do with. That change sounds more relevant to the FIXME I added to the effect of "handle CDATASections in non-listener mode". That is, treating Comments, CDATA and PIs as less 'ignore this legacy junk' and more 'real nodes in the parse tree'.

File: Libraries/LibWeb/XML/XMLDocumentBuilder.cpp [hunk 1]
```diff
@@ -262,7 +264,26 @@ void XMLDocumentBuilder::comment(StringView data)
 {
     if (m_has_error)
         return;
-    MUST(m_document->append_child(m_document->create_comment(MUST(String::from_utf8(data)))));
+
+    MUST(m_current_node->append_child(m_document->create_comment(MUST(String::from_utf8(data)))));
```
alimpfard, line 268: This should be something like `if (!m_current_node) return;` to drop comments (also for PIs) that aren't attached to at least the root node (in reference to the previous comments also)

ADKaster: aha. yeah. with this fix, all the tests added by the mentioned WPT PR pass. I've imported them.

Issue #3512, opened 2025-02-08
Border collapses inside an element when border-radius is set to a very big value


sekpd:
### Summary

When an element has `border` and `border-radius` that is set to a very large value, the border looks collapsed inside the element.
It starts to become visible around `18000000px` and higher in case of a 256x256 px element.

### Operating system

Linux

### Steps to reproduce

1. Open the reduced test case in Ladybird
2. Compare how it looks in Chrome or Firefox

### Expected behavior

Border doesn't collapse inside the element and aligns with the edge of the element.

### Actual behavior

Border collapses inside the element.

### URL for a reduced test case

N/A

### HTML/SVG/etc. source for a reduced test case

```html
<!DOCTYPE html>
<style>
    div {
        width: 256px;
        height: 256px;
        background-color: red;
        border: 2px solid black;
        border-radius: 99999999px;
    }
</style>
<div></div>
```

### Log output and (if possible) backtrace

```shell
None
```

### Screenshots or screen recordings

Ladybird:
![Ladybird](https://github.com/user-attachments/assets/f09f4c19-f5dc-4a2a-ad02-e7ad41209fa2)

Firefox:
![Firefox](https://github.com/user-attachments/assets/cb35552a-3845-445d-bc92-176833c8b853)

### Build flags or config settings

None

### Contribute a patch?

- [ ] I’ll contribute a patch for this myself.

PR #5513, opened 2025-07-18
LibWeb/Painting: Prevent border implosion when border-radius is set to large values 

FadedBronze:
Fixes: #3512

In the [spec](https://www.w3.org/TR/css-backgrounds-3/#corner-overlap) it says that large border radii that sum to be greater than their respective side lengths should scale down proportionally based on which radii had larger values. To retain that scaling in the implementation there is two iterations of normalization on the radii. 

This avoids the precision errors encountered before that lead to instability (starts at around 17000000px). 

2025-07-18-155323_hyprshot


kalenikaliaksandr (changes requested):
could you add screenshot test for this change please

Libraries/LibWeb/Painting/Paintable.cpp
```diff
@@ -229,24 +229,36 @@ Painting::BorderRadiiData normalize_border_radii_data(Layout::Node const& node,
...
-    auto s_bottom = (bottom_left_radius_px.horizontal_radius + bottom_right_radius_px.horizontal_radius);
-    auto s_left = (top_left_radius_px.vertical_radius + bottom_left_radius_px.vertical_radius);
-    CSSPixelFraction f = 1;
-    f = (s_top != 0) ? min(f, l_top / s_top) : f;
-    f = (s_right != 0) ? min(f, l_right / s_right) : f;
-    f = (s_bottom != 0) ? min(f, l_bottom / s_bottom) : f;
-    f = (s_left != 0) ? min(f, l_left / s_left) : f;
-
-    // If f < 1, then all corner radii are reduced by multiplying them by f.
-    if (f < 1) {
```
InvalidUsernameException, line 243: This comment was a quote from the spec text, so it should be kept as per [LibWebPatterns.md#comments](https://github.com/LadybirdBrowser/ladybird/blob/master/Documentation/LibWebPatterns.md#comments).

Similarly since the `f < 1` condition is based on spec-text, I'd argue it should not be inverted here to keep the code more verbatim wrt the spec description.

FadedBronze:
I put in the screenshot test, put back the comment. and inverted that if statement

InvalidUsernameException:
This fixes one of the button-related problems on chatgpt.com


| Before | After  |
|--------|--------|
| Before | After |

dan-v:
this can be closed now that https://github.com/LadybirdBrowser/ladybird/pull/5850 has merged

PR #5514, opened 2025-07-19
AK+LibWeb/CSS: Add `lower-greek` counter style

dmaivel:
This PR adds support for `lower-greek` as a counter style.

I implemented a member function `greek_letter_from` in AK's `String` type as it appears that `bijective_base_from` doesn't handle multibyte characters correctly. I wasn't sure, but it may be better to implement a variant of `bijective_base_from` which accepts multibyte characters instead of having a seperate function for greek letters, as it'll allow us to add support for other styles like `georgian`. My only concern with this approach is where we would place the respective alphabets.

The test case for `lower-greek` in `css/cssom/serialize-values.html` passes. 

The test cases in `css/css-counter-styles/lower-greek/` appear to fail, but when testing cases like those in `css/css-counter-styles/upper-roman`, those fail too, so I'm not sure if its important to get these cases passing.

### DIFF
--- a/AK/String.cpp
+++ b/AK/String.cpp
@@ -492,6 +492,21 @@ String String::bijective_base_from(size_t value, Case target_case, unsigned base
     return MUST(from_utf8(ReadonlyBytes(buffer.data(), i)));
 }
 
+String String::greek_letter_from(size_t value)
+{
+    static StringView const map = "αβγδεζηθικλμνξοπρστυφχψω"sv;
+    static unsigned const base = 24;
+
+    StringBuilder builder;
+    while (value > 0) {
+        value--;
+        builder.append(map.substring_view((value % base) * 2, 2));
+        value /= base;
+    }
+
+    return MUST(builder.to_string_without_validation().reverse());
+}
+
 String String::roman_number_from(size_t value, Case target_case)
 {
     if (value > 3999)
--- a/Tests/LibWeb/Text/expected/wpt-import/css/cssom/serialize-values.txt
+++ b/Tests/LibWeb/Text/expected/wpt-import/css/cssom/serialize-values.txt
@@ -2,8 +2,8 @@ Harness status: OK
 
 Found 687 tests
 
-658 Pass
-29 Fail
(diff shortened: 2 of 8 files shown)


Tests/LibWeb/Text/expected/wpt-import/css/cssom/serialize-values.txt
```diff
@@ -2,8 +2,8 @@ Harness status: OK
 
 Found 687 tests
 
-658 Pass
-29 Fail
+659 Pass
```
trflynn89, line 5: This commit should be squashed into the previous commit. We are big fans of atomic commits in this project - every commit should build and pass tests on its own. If we were to bisect between this commit and the previous one, this test would fail.

(More practically, if we had to revert the previous commit, we would then also have to know to revert this one).

AK/String.cpp
```diff
@@ -492,6 +492,21 @@ String String::bijective_base_from(size_t value, Case target_case, unsigned base
...
+    static unsigned const base = 24;
+
+    StringBuilder builder;
+    while (value > 0) {
+        value--;
+        builder.append(map.substring_view((value % base) * 2, 2));
+        value /= base;
+    }
+
+    return builder.to_string_without_validation().reverse().value();
```
trflynn89:
```c++
return MUST(builder.to_string_without_validation().reverse());
```

trflynn89 (commented):
Adding a special method for Greek is fine for now, but I wonder if in the future we could make use of ICU to handle such conversions.

dmaivel:
Thanks for comments, I squashed the commits into one and added `MUST` to the return.

AtkinsSJ:
> Adding a special method for Greek is fine for now, but I wonder if in the future we could make use of ICU to handle such conversions.

In the future we can have the standard set of `@counter-style`s defined in the UA stylesheet, which covers this and a bunch of other cases. But I got lost on a yak shaving expedition while trying to work on that. 😅

AtkinsSJ (approved):

PR #5515, opened 2025-07-19
LibWeb: Remove animations from timeline in `Element::removed_from()`

kalenikaliaksandr:
`AnimationTimeline` visits pointers of all registered animations, so if element is removed from DOM tree but its animations remain registered in timeline, then `Animation` and owner `Element` will be kept alive until `AnimationTimeline` is destroyed.


Libraries/LibWeb/Animations/Animatable.cpp
```diff
@@ -218,6 +218,15 @@ void Animatable::clear_transitions(Optional<CSS::PseudoElement> pseudo_element)
     transition.transition_attributes.clear();
 }
 
+void Animatable::remove_animations_from_timeline()
+{
+    // This is needed to avoid leaking Animation objects
+    auto& impl = ensure_impl();
+    for (auto animation : impl.associated_animations) {
+        animation->set_timeline({});
```
shannonbooth, line 226: Would we run into the same issue if the node is moved into a disconnected DOM tree? 🤔

kalenikaliaksandr: Yeah, probably we need similar fix for "move" operation.

PR #5517, opened 2025-07-19
LibWeb: Correctly resolve colors in layout-node-less contexts

Calme1709:
This PR means that we now:
	- Support relative lengths within `calc`s in `stop-color`
	- Render <feFlood> using the specified `flood-color`

### DIFF
--- a/Libraries/LibWeb/CSS/ComputedProperties.cpp
+++ b/Libraries/LibWeb/CSS/ComputedProperties.cpp
@@ -443,16 +443,6 @@ ClipRule ComputedProperties::clip_rule() const
     return keyword_to_fill_rule(value.to_keyword()).release_value();
 }
 
-Color ComputedProperties::flood_color(Layout::NodeWithStyle const& node) const
-{
-    auto const& value = property(PropertyID::FloodColor);
-    if (value.has_color()) {
-        return value.to_color(ColorResolutionContext::for_layout_node_with_style(node)).value();
-    }
-
-    return InitialValues::flood_color();
-}
-
 float ComputedProperties::flood_opacity() const
 {
     auto const& value = property(PropertyID::FloodOpacity);
--- a/Libraries/LibWeb/CSS/ComputedProperties.h
+++ b/Libraries/LibWeb/CSS/ComputedProperties.h
@@ -193,7 +193,6 @@ public:
     float stroke_opacity() const;
     FillRule fill_rule() const;
     ClipRule clip_rule() const;
-    Color flood_color(Layout::NodeWithStyle const&) const;
     float flood_opacity() const;
 
     Gfx::FontCascadeList const& computed_font_list() const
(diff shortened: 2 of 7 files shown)


Libraries/LibWeb/CSS/CalculationResolutionContext.h
```diff
@@ -0,0 +1,21 @@
+/*
```
gmta, line 1: Commit title: "it's" -> "its"

Libraries/LibWeb/CSS/CSSStyleValue.cpp
```diff
@@ -73,6 +73,26 @@
 
 namespace Web::CSS {
 
+ColorResolutionContext ColorResolutionContext::for_element(DOM::AbstractElement const& element)
+{
+    auto preferred_color_scheme = element.document().page().preferred_color_scheme();
+
+    // FIXME: We can't just check for Auto because page().preferred_color_scheme() returns garbage data after startup.
```
gmta: Seems like it would be better to properly initialize that, instead of dealing with a random value? 🤔

Calme1709: Haven't looked much into this issue - I copied both the FIXME and the workaround from https://github.com/LadybirdBrowser/ladybird/blob/5a1de8a187e89a6fc7a4e3bfe8355dede7751e4b/Libraries/LibWeb/Layout/Node.cpp#L379

I have added a commit to upstream handling of `Auto` resolution to `Page::preferred_color_scheme` so at least we only have one place with this FIXME.

Libraries/LibWeb/CSS/StyleValues/CalculatedStyleValue.h
```diff
@@ -31,11 +31,6 @@ struct CalculationContext {
     Optional<ValueType> percentages_resolve_as {};
     bool resolve_numbers_as_integers = false;
 };
-// Contains the context for resolving the calculation.
-struct CalculationResolutionContext {
-    Variant<Empty, Angle, Frequency, Length, Time> percentage_basis {};
-    Optional<Length::ResolutionContext> length_resolution_context;
-};
```
AtkinsSJ, line 38: This commit looks like it won't build - nothing includes the new `CalculationResolutionContext.h` header.

AtkinsSJ (changes requested):
Generally looks good, just one point I noticed.

Libraries/LibWeb/CSS/StyleValues/CalculatedStyleValue.h
Calme1709: Have moved the addition of `#include <LibWeb/CSS/CalculationResolutionContext.h>` in CSSStyleValue.h into this commit. Unfortunately not on a device where I can try compiling but I think that should fix it.

Also rebased to make sure CI still passes.

Calme1709: Have now confirmed that commit builds. Have rebased again and I am pretty sure the previous CI fails were unrelated flakes.

AtkinsSJ (approved):

PR #5519, opened 2025-07-19
LibWeb: A few Wasm IDL fixes

trflynn89:
Fixes ~1000 WPT tests in `/wasm`. Full comparison log here:
[wasm.txt](https://github.com/user-attachments/files/21327822/wasm.txt)

Also causes ~30 "regressions" in `/wasm`. These were actually incorrectly passing previously:
```
  ▶ Unexpected subtest result in /wasm/core/data.wast.js.html:
  │ FAIL [expected PASS] #53 Test that a WebAssembly module is uninstantiable
  │   → assert_true: expected link error, observed LinkError: Data segment attempted to write to out-of-bounds memory (0) in memory of size 0
    at Error
    at assert_uninstantiable (http://web-platform.test:8000/wasm/core/js/harness/async_index.js:372:15)
    at data_wast_js (http://web-platform.test:8000/wasm/core/js/data.wast.js:85:22)
    at http://web-platform.test:8000/wasm/core/js/data.wast.js:186:3
```

(Despite the "expected link error" text, the test is expecting a `RuntimeError`):
https://github.com/web-platform-tests/wpt/blob/6b2239ddd085f1945d5e41406baff9df57f7a42a/wasm/core/js/harness/async_index.js#L378-L379

Before this change, `LinkError` and `RuntimeError` were both just `Error`. Thus, `LinkError() instanceof RuntimeError` would be true and the test "passed". It now correctly fails.

### DIFF
--- a/Libraries/LibWeb/WebAssembly/WebAssembly.cpp
+++ b/Libraries/LibWeb/WebAssembly/WebAssembly.cpp
@@ -232,37 +232,38 @@ JS::ThrowCompletionOr<NonnullOwnPtr<Wasm::ModuleInstance>> instantiate_module(JS
     if (!module.import_section().imports().is_empty() && !import_object) {
         return vm.throw_completion<JS::TypeError>("ImportObject must be provided when module has imports"sv);
     }
+
     // 2. Let imports be « ».
     HashMap<Wasm::Linker::Name, Wasm::ExternValue> resolved_imports;
     if (import_object) {
         dbgln_if(LIBWEB_WASM_DEBUG, "Trying to resolve stuff because import object was specified");
+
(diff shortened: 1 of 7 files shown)


trflynn89:
@alimpfard I wasn't able to run a full WPT to see if I got the same regressions as you due to the IDL `object` change, my computer OOM'd. So I spot checked folders like `/html` and `/dom` and saw no regressions :man_shrugging:

alimpfard:
I'll start a run, my runner is idle for now

alimpfard:
Seems to be fine, I compared with a fresh run of master (+484 -146 passes)
Here's the diff: [diff.txt](https://github.com/user-attachments/files/21330770/diff.txt)

alimpfard (approved):

PR #5527, opened 2025-07-20
LibWeb: Don't when creating an empty bitmap from a HTMLCanvasElement

tcl3:


Gingeh:
> Don't when creating a ...

You may have forgotten a verb there

tcl3:
> You may have forgotten a verb there

Gotta get under that 72 character limit :sweat_smile:

Issue #3579, opened 2025-02-14
Buttons aren't y-aligned right


Psychpsyo:
### Summary

When I put a button in ladybird on the same line as something else, it is too low.

Reduced from [the yugoslavia.best scrunklo gacha](https://yugoslavia.best/gacha/)

### Operating system

Linux

### Steps to reproduce

1. Open the reduced test case.
2. Notice that the button doesn't line up with the input.

### Expected behavior

The button should line up with the input.

### Actual behavior

The button does not line up with the input.

### URL for a reduced test case

`data:text/html,<input><button>Hi!</button>`

### HTML/SVG/etc. source for a reduced test case

```html
<!DOCTYPE html>
<input><button>Hi!</button>
```

### Log output and (if possible) backtrace

```shell
none
```

### Screenshots or screen recordings

Top: ladybird
Bottom: firefox
![Image](https://github.com/user-attachments/assets/c1b3bd99-2a7c-4e2b-b646-7ef7b8fd1814)

### Build flags or config settings

none

### Contribute a patch?

- [ ] I’ll contribute a patch for this myself.

Kuba314:
This is a more reduced example of https://github.com/LadybirdBrowser/ladybird/issues/3579.

Psychpsyo:
> This is a more reduced example of [#3579](https://github.com/LadybirdBrowser/ladybird/issues/3579).

Uh.. this is #3579?

Kuba314:
Oops, I meant https://github.com/LadybirdBrowser/ladybird/issues/759

tcl3:
This now appears to render correctly.

Issue #759, opened 2024-07-21
Alignment issue in form input elements when navigating to https://lore.kernel.org/lists.html


kemzeb:
When navigating to https://lore.kernel.org/lists.html, the search inbox form on the top-left appears to have some misalignment between the text input and button elements:

Ladybird:
![Screenshot 2024-07-21 155854](https://github.com/user-attachments/assets/d84bdd73-1f66-4761-b143-01f6cc48a24b)

Firefox:
![Screenshot 2024-07-21 160058](https://github.com/user-attachments/assets/41928f2e-e656-48c6-b8c1-1ca4ee4f9ebf)

The minimal reproducible where I was able to see the issue:
```html
<html>

<head>
</head>

<body>
    <pre><input name="q" type="text/"><input type="submit" value="locate inbox"><input type="submit" name="a" value="search all inboxes"></pre>
</body>

</html>
```

simonkrauter:
Looks like it has been fixed:
<img width="350" alt="Image" src="https://github.com/user-attachments/assets/960cbe7b-f956-4f10-acdd-651e36ecf2ac" />

tcl3:
I can no longer reproduce this issue. The minimal reproduction currently creates a very tall input, but this issue is unrelated (see #7721)

PR #5530, opened 2025-07-20
Fix cursor not showing in empty input boxes

### LINKED ISSUE #1769: No cursor blinking, with blank text
I've noticed on `https://google.com` (spoofed Chrome under Linux user agent) that there's no cursor blinking, if you click on the search text field, unless you type a character.

This might be a minor thing, but it looks and feels weird, if you click somewhere you should be able to type, but don't get a cursor.

Chromium 129.0.6668.89:

Screenshot_20241013_135941

Ladybird (34261e54901):

Screenshot_20241013_135955

zacoons:
This PR primarily fixes #1769, but it also:
- Makes the cursor less blurry looking, and
- Removes a method I think was unnecessary

What I have currently doesn't seem like an ideal solution. I had the idea that the paintable box could be responsible for rendering the cursor, instead of the fragment. That way it doesn't matter if the box has content or not. This idea seems like a much better solution, but I don't know how the cursor would be positioned/scaled then.

Any thoughts?


Libraries/LibWeb/Painting/Paintable.cpp
```diff
@@ -165,10 +165,9 @@ void Paintable::set_needs_display(InvalidateDisplayList should_invalidate_displa
 
     if (!is<Painting::PaintableWithLines>(*containing_block))
         return;
-    static_cast<Painting::PaintableWithLines const&>(*containing_block).for_each_fragment([&](auto& fragment) {
+    for (const auto& fragment : static_cast<Painting::PaintableWithLines const&>(*containing_block).fragments()) {
```
gmta: Please use `auto const&` (east const)

gmta:
You've based your work on a commit that's 3 weeks old; if you want CI to run you need to rebase your branch :)

zacoons:
Seems like other browsers may use this "fake empty text" technique to position the caret in an empty input field as well, so maybe it's not as gross as I thought. I've done some tidy up and it should be good to go now.

Libraries/LibWeb/Layout/InlineFormattingContext.cpp
```diff
@@ -263,15 +263,15 @@ void InlineFormattingContext::generate_line_boxes()
...
-        }
+        // if (item.is_collapsible_whitespace && (line_boxes.is_empty() || line_boxes.last().is_empty_or_ends_in_whitespace())) {
+        //     if (item.node->computed_values().text_wrap_mode() == CSS::TextWrapMode::Wrap) {
+        //         auto next_width = iterator.next_non_whitespace_sequence_width();
+        //         if (next_width > 0)
+        //             line_builder.break_if_needed(next_width);
+        //     }
+        //     leading_margin_from_collapsible_whitespace += item.margin_start;
+        //     continue;
+        // }
```
kalenikaliaksandr: could you rearrange commit history instead of adding commented out code?

Libraries/LibWeb/Layout/InlineLevelIterator.cpp
```diff
@@ -498,9 +498,14 @@ Optional<InlineLevelIterator::Item> InlineLevelIterator::next_without_lookahead(
...
-            skip_to_next();
-            return next_without_lookahead();
+            if (!m_text_node_context->is_last_chunk && !m_text_node_context->chunk_iterator.peek(0).has_value() && text_node.text_for_rendering().byte_count() == 0) {
+                chunk_opt = m_text_node_context->chunk_iterator.create_empty_chunk();
+            }
+            else {
+                m_text_node_context = {};
+                skip_to_next();
+                return next_without_lookahead();
+            }
```
kalenikaliaksandr: adding new layout test is strongly encouraged for any change that affects layout code.

Libraries/LibWeb/Painting/PaintableBox.cpp
```diff
@@ -713,26 +713,30 @@ void paint_cursor_if_needed(PaintContext& context, TextPaintable const& paintabl
...
         fragment_rect.top(),
         1,
         fragment_rect.height()
     };
-
     auto cursor_device_rect = context.rounded_device_rect(cursor_rect).to_type<int>();
 
-    context.display_list_recorder().draw_rect(cursor_device_rect, caret_color);
+    // Add it to be rendered
+    context.display_list_recorder().fill_rect(cursor_device_rect, caret_color);
```
kalenikaliaksandr: IMO these newly added comments don't bring any value. please update commit description with explanation why it improves cursor appearance.

kalenikaliaksandr (changes requested):

Libraries/LibWeb/Painting/PaintableBox.cpp
zacoons: Ah, I can see how this was unclear. I'll remove those comments if you don't think they're useful. The meat of this commit was changing `draw_rect` to `fill_rect` on line 739.

Libraries/LibWeb/Layout/InlineLevelIterator.cpp
zacoons: Doing this has led me to discover a number of broken tests from my change. It may take me some time to wrap my head around all of this.

I suspect it's because it's generating an empty text chunk for any type of empty text, not just inputs/text areas/editables. So I need a way to check if a given `Layout::TextNode` is part of an input/text area/editable. Do you know how I could do that?

EDIT:
I solved the first problem, but now I've found another two. The cursor appears at the end of the placeholder text, and it doesn't appear at all for empty text areas.

EDIT:
I have it fully working on the `<input>` element, there's just still some odd behaviour on the `<textarea>` element that I need to iron out. It's because the text area element isn't being repainted when the blink state changes, because its fragment is being added to the wrong parent (see `LayoutState.cpp:try_to_relocate_fragment_in_inline_node`) for some reason. I will address this separately.

gmta:
@zacoons Mind separating out that first commit (improve caret appearance) into a new PR while you get the rest of this PR sorted out?

zacoons:
@gmta sure thing: #5664

AtkinsSJ:
I haven't reviewed this, but I just came across a layout bug and thought your changes might fix it... and they do!

Specifically, empty `<input>` elements get vertically aligned wrong. For example, this HTML:

```html
<!DOCTYPE html>
<span style="border: 1px solid black">Span</span>
<span style="border: 1px solid black; display: inline-block">Inline-block</span>
<input/>
<input type="button" value="Button input">
<button>Button</button>
```
Without this PR:
Screenshot_20250806_164507

With this PR:
Screenshot_20250806_164544

(I'm not 100% sure we're doing the right thing still, but at least it's now consistent between `<input>`s with and without a value or placeholder.)

zacoons:
@AtkinsSJ looks like it's still roughly 1 pixel below the other `<input>`s on the right though :(
BTW `placesholder` is a typo in your HTML there, which is why the placeholder isn't showing.

AtkinsSJ:
> BTW `placesholder` is a typo in your HTML there, which is why the placeholder isn't showing.

Oh yeah that was intentional to disable the placeholder. I shouldn't have left that in.

Libraries/LibWeb/Layout/InlineLevelIterator.cpp
```diff
@@ -491,20 +492,33 @@ Optional<InlineLevelIterator::Item> InlineLevelIterator::next_without_lookahead(
         return {};
 
     if (is<Layout::TextNode>(*m_current_node)) {
-        auto& text_node = static_cast<Layout::TextNode const&>(*m_current_node);
+        auto const& text_node = static_cast<Layout::TextNode const&>(*m_current_node);
```
AtkinsSJ: The changes to InlineLevelIterator are separate enough that they belong in their own commit. (Generally, if you end up putting "and" or "also" in a commit message, it's a clue that it might want to be multiple commits. Sometimes that's not possible though.)

Tests/LibWeb/Layout/expected/block-and-inline/inline-block-leading-and-trailing-metrics.txt
```diff
@@ -18,13 +18,13 @@ Viewport <#document> at (0,0) content-size 800x600 children: not-inline
       BlockContainer <(anonymous)> at (8,28) content-size 784x0 children: inline
         TextNode <#text>
       BlockContainer <div#b> at (9,29) content-size 782x18 children: inline
-        frag 0 from BlockContainer start: 0, length: 0, rect: [70.296875,29 88.453125x18] baseline: 13.796875
+        frag 0 from BlockContainer start: 0, length: 0, rect: [50.296875,29 88.453125x18] baseline: 13.796875
```
AtkinsSJ, line 21: Please update test results in the same commit that modifies them. (So, I'm assuming this and most of these others should go in the previous commit.)

Adding *new* tests in a separate commit is fine.

AtkinsSJ (changes requested):
I think the changes themselves are fine, just a couple of issues with the commits themselves. Also: It looks like you've wrapped a couple of commit messages at 50 columns instead of 72.

Libraries/LibWeb/Layout/InlineLevelIterator.cpp
zacoons: How do you like the commit history now?
(Discussed on Discord)

Libraries/LibWeb/Painting/Paintable.cpp
```diff
@@ -186,10 +185,10 @@ CSSPixelPoint Paintable::box_type_agnostic_position() const
 
     CSSPixelPoint position;
     if (auto const* block = containing_block(); block && is<Painting::PaintableWithLines>(*block)) {
-        static_cast<Painting::PaintableWithLines const&>(*block).for_each_fragment([&](auto& fragment) {
-            position = fragment.absolute_rect().location();
-            return IterationDecision::Break;
-        });
+        auto const fragments = static_cast<Painting::PaintableWithLines const&>(*block).fragments();
```
gmta: This should be `auto const&` to prevent a vector copy

Libraries/LibWeb/Layout/InlineLevelIterator.cpp
```diff
@@ -495,6 +495,8 @@ Optional<InlineLevelIterator::Item> InlineLevelIterator::next_without_lookahead(
 
         if (!m_text_node_context.has_value())
             enter_text_node(text_node);
+        else
+            m_text_node_context->is_first_chunk = false;
```
gmta, line 498: Good find! The only place we use this in, is `InlineLevelIterator::add_extra_box_model_metrics_to_item` as the second argument (`bool add_leading_metrics`). Apparently that bool was always true, but we didn't see this pop up as a bug? So either this bool is useless and the behavior was correct through other metrics, or we are missing a test exposing this bug.

Could you check if something's fixed as a result of your change and if so, add a test preventing regressions?

Libraries/LibWeb/Layout/InlineLevelIterator.cpp
```diff
@@ -498,15 +499,26 @@ Optional<InlineLevelIterator::Item> InlineLevelIterator::next_without_lookahead(
...
+        auto is_empty_editable = false;
         if (!chunk_opt.has_value()) {
-            m_text_node_context = {};
-            skip_to_next();
-            return next_without_lookahead();
-        }
+            auto const is_only_chunk = m_text_node_context->is_first_chunk && m_text_node_context->is_last_chunk;
+            if (is_only_chunk && text_node.text_for_rendering().is_empty()) {
+                if (auto const* shadow_root = as_if<DOM::ShadowRoot>(text_node.dom_node().root()))
+                    is_empty_editable = shadow_root->host() && (is<HTML::FormAssociatedTextControlElement>(shadow_root->host()) || shadow_root->host()->is_editable_or_editing_host());
```
gmta: I'm not sure if you need to take a look at the shadow root here, that seems awfully specific for our input fields. For example, this testcase does not work:

```html
<!DOCTYPE html>
<div contenteditable>foobar</div>
```

If you remove "foobar", the cursor disappears.

Maybe it's enough to take a look at the node's parent and check if `::is_editing_host()` is true?

zacoons: Ack, looks like the div doesn't have a TextNode when it's empty. So the cursor won't get rendered even if I was checking the right things here. I'll have to get it to add an empty TextNode when the div is contenteditable.

zacoons:
The layout issue mentioned above by @AtkinsSJ is related to #3579 and #759. However, this PR doesn't completely fix that alignment issue since it's still a tiny bit offset vertically.

Libraries/LibWeb/Layout/InlineLevelIterator.cpp
zacoons: I've made a rather ugly fix to this in the latest commit to this branch, but I can't think of any other way to fix it. Let me know what you think.

gmta: I cannot get the test case mentioned earlier in this discussion to work properly. An empty `<div contenteditable></div>` also doesn't show the cursor?

Libraries/LibWeb/Painting/Paintable.cpp
```diff
@@ -165,13 +165,10 @@ void Paintable::set_needs_display(InvalidateDisplayList should_invalidate_displa
     if (should_invalidate_display_list == InvalidateDisplayList::Yes)
         document.invalidate_display_list();
 
-    auto* containing_block = this->containing_block();
-    if (!containing_block)
+    auto* parent = as_if<Painting::PaintableWithLines>(this->parent());
+    if (!parent)
```
gmta: Commit message: please don't [use weasel-words such as "fix"](https://github.com/LadybirdBrowser/ladybird/blob/master/CONTRIBUTING.md#code-submission-policy), but rather describe what you changed and why.

Libraries/LibWeb/Layout/InlineLevelIterator.cpp
```diff
@@ -587,7 +599,7 @@ Optional<InlineLevelIterator::Item> InlineLevelIterator::next_without_lookahead(
         CSSPixels chunk_width = CSSPixels::nearest_value_for(glyph_run->width() + x);
 
         // NOTE: We never consider `content: ""` to be collapsible whitespace.
-        bool is_generated_empty_string = text_node.is_generated_for_pseudo_element() && chunk.length == 0;
+        bool is_generated_empty_string = (is_empty_editable || text_node.is_generated_for_pseudo_element()) && chunk.length == 0;
```
gmta: Isn't `chunk.length == 0` implied when `is_empty_editable` is `true?

Libraries/LibWeb/Layout/InlineLevelIterator.cpp
gmta: If you have a `FormAssociatedTextControlElement`, shouldn't we also look at `::is_mutable()`? If the control element is disabled or readonly, we probably shouldn't entertain adding empty fragments.

zacoons: In the former case you have to click on it again to see the cursor. I don't know why. I can investigate further if you think this could be a valid solution.

The latter case - of a div that's empty in the first place - is a separate layout issue which I have opened here: #5872

Try testing with a border around the div so you can see where it is:
```html
<!DOCTYPE html>
<div style="border:solid 1px black;" contenteditable>foobar</div>
```

zacoons: Good thinking

Libraries/LibWeb/Layout/InlineLevelIterator.cpp
zacoons: I added this line and tried a whole bunch of different websites and ran test-web, but it didn't get called that I saw.
```cpp
void InlineLevelIterator::add_extra_box_model_metrics_to_item(Item& item, bool add_leading_metrics, bool add_trailing_metrics)
{
    if (!add_leading_metrics && m_extra_leading_metrics.has_value())
        dbgln("\n\n\n\n\n{} {} {} {} {}\n\n\n\n\n", m_extra_leading_metrics->margin, m_extra_leading_metrics->border, m_extra_leading_metrics->padding, item.node->debug_description(), item.node->parent()->debug_description());
```

AHA, never mind that, I found a test case. It's a weird one though.

AtkinsSJ (approved):
I think we can merge this. Thanks!

PR #5531, opened 2025-07-20
LibWeb: Check if transferred ImageBitmap is exposed in target realm

Lubrsi:
This was forgotten in ed3d0d7.


shannonbooth:
Woopsy as well, I missed even lookjng a mt it recently whth https://github.com/LadybirdBrowser/ladybird/pull/5497#discussion_r2214878498

Maybe I should just replace it

PR #5534, opened 2025-07-20
LibWeb: Serialize empty ImageBitmap data as null

tcl3:


Libraries/LibWeb/HTML/ImageBitmap.cpp
```diff
@@ -20,19 +20,26 @@ GC_DEFINE_ALLOCATOR(ImageBitmap);
     return Gfx::Bitmap::create_wrapper(format, alpha_type, Gfx::IntSize(width, height), pitch, data.data());
 }
 
-static void serialize_bitmap(HTML::TransferDataEncoder& encoder, Gfx::Bitmap& bitmap)
+static void serialize_bitmap(HTML::TransferDataEncoder& encoder, RefPtr<Gfx::Bitmap> bitmap)
```
trflynn89: This can be `const&`

PR #5535, opened 2025-07-20
LibWeb: Implement WebAudio control message queue

treyshaffer:
## Summary
- Implements the WebAudio control message queue as specified in the Web Audio API

## Changes
- Added `ControlMessageQueue` class with thread-safe operations
- Integrated queue into `BaseAudioContext` 
- Updated `AudioContext` methods (suspend/resume/close) to use the queue
- Updated `AudioScheduledSourceNode` to queue start/stop messages

## Testing
- Added custom integration test for the control message queue

### DIFF
--- a/Libraries/LibWeb/WebAudio/ControlMessageQueue.cpp
+++ b/Libraries/LibWeb/WebAudio/ControlMessageQueue.cpp
@@ -1,0 +1,91 @@
+/*
+ * Copyright (c) 2025, Trey Shaffer <trey@trsh.dev>
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include <LibJS/Runtime/Realm.h>
+#include <LibWeb/WebAudio/ControlMessageQueue.h>
+
+namespace Web::WebAudio {
+
+GC_DEFINE_ALLOCATOR(ControlMessageQueue);
+
+GC::Ref<ControlMessageQueue> ControlMessageQueue::create(JS::Realm& realm)
+{
+    return realm.create<ControlMessageQueue>(realm);
+}
+
+ControlMessageQueue::ControlMessageQueue(JS::Realm& realm)
+    : Bindings::PlatformObject(realm)
+{
+}
+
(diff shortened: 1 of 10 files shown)


Libraries/LibWeb/WebAudio/ControlMessageQueue.cpp
```diff
@@ -0,0 +1,91 @@
...
+                return;
+
+            if (m_messages.is_empty())
+                return;
+
+            m_is_processing = true;
+
+            // Snapshot prevents deadlock from reentrant enqueue
+            messages_to_process = move(m_messages);
+            m_messages.clear();
```
gmta, line 63: You don't need to `.clear()` if you've just moved from it. See `Vector(Vector&& other)`.

Libraries/LibWeb/WebAudio/ControlMessageQueue.cpp
```diff
@@ -0,0 +1,91 @@
...
+    }
+}
+
+void ControlMessageQueue::visit_edges(Cell::Visitor& visitor)
+{
+    Base::visit_edges(visitor);
+
+    Threading::MutexLocker locker(m_mutex);
+    for (auto& message : m_messages)
+        visitor.visit(message);
```
gmta, line 88: You can just do `visitor.visit(m_messages)` here.

Libraries/LibWeb/WebAudio/ControlMessageQueue.cpp
```diff
@@ -0,0 +1,91 @@
...
+        // FIXME: Should be called from the rendering thread
+
+        for (auto& message : messages_to_process) {
+            message->function()();
+        }
+
+        {
+            Threading::MutexLocker locker(m_mutex);
+            if (m_messages.is_empty()) {
+                m_is_processing = false;
```
gmta, line 75: What is supposed to happen if a message is queued during message execution? Shouldn't `m_is_processing` always be set to false when we're done iterating over our copy of the messages?

PR #5545, opened 2025-07-21
LibWeb: Only cascade relevant CSS properties for logical alias context

awesomekling:
We only need the cascaded values for writing-mode and direction when resolving the logical alias context, so we can skip all other properties.

### DIFF
--- a/Libraries/LibWeb/CSS/StyleComputer.cpp
+++ b/Libraries/LibWeb/CSS/StyleComputer.cpp
@@ -876,12 +876,20 @@ void StyleComputer::cascade_declarations(
     CascadeOrigin cascade_origin,
     Important important,
     Optional<FlyString> layer_name,
-    Optional<LogicalAliasMappingContext> logical_alias_mapping_context) const
+    Optional<LogicalAliasMappingContext> logical_alias_mapping_context,
+    ReadonlySpan<PropertyID> properties_to_cascade) const
 {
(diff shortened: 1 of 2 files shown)


Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -882,6 +882,15 @@ void StyleComputer::cascade_declarations(
     auto cascade_style_declaration = [&](CSSStyleProperties const& declaration) {
         seen_properties.fill(false);
         for (auto const& property : declaration.properties()) {
+
+            // OPTIMIZATION: If we're computing the logical alias mapping context, we only care about the properties
+            //               that affect the logical mappings.
+            if (!logical_alias_mapping_context.has_value()) {
+                if (property.property_id != PropertyID::WritingMode && property.property_id != PropertyID::Direction) {
```
AtkinsSJ: `text-direction` is also needed, once we implement that. Maybe add a FIXME about it?

awesomekling: I made it so that the set of whitelisted properties is an input to the cascade. That way the function that creates the logical alias context has the list locally :)

AtkinsSJ: Even better!

AtkinsSJ (approved):

PR #5548, opened 2025-07-21
LibRegex+LibJS: Fix non-ASCII unicode regex matching

gmta:
There apparently is a bit of a disconnect between the spec asking us to construct the pattern using code points and LibRegex not being able to swallow those. Whenever we had multi-byte code points in the pattern and tried to match that in unicode mode, we would fail.

Change the parser to encode all non-ASCII code units. Fixes 2 test262 cases in `language/literals/regexp`.

### DIFF
--- a/Libraries/LibJS/Runtime/RegExpObject.cpp
+++ b/Libraries/LibJS/Runtime/RegExpObject.cpp
@@ -96,19 +96,10 @@ ErrorOr<String, ParseRegexPatternError> parse_regex_pattern(StringView pattern, 
     auto utf16_pattern = Utf16String::from_utf8(pattern);
     StringBuilder builder;
 
-    // If the Unicode flag is set, append each code point to the pattern. Otherwise, append each
-    // code unit. But unlike the spec, multi-byte code units must be escaped for LibRegex to parse.
+    // FIXME: We need to escape multi-byte code units for LibRegex to parse since the lexer there doesn't handle unicode.
     auto previous_code_unit_was_backslash = false;
-    for (size_t i = 0; i < utf16_pattern.length_in_code_units();) {
-        if (unicode || unicode_sets) {
-            auto code_point = code_point_at(utf16_pattern, i);
-            builder.append_code_point(code_point.code_point);
-            i += code_point.code_unit_count;
-            continue;
-        }
-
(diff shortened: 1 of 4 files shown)


Libraries/LibJS/Runtime/RegExpObject.cpp
```diff
@@ -96,19 +96,10 @@ ErrorOr<String, ParseRegexPatternError> parse_regex_pattern(StringView pattern,
...
     auto previous_code_unit_was_backslash = false;
-    for (size_t i = 0; i < utf16_pattern.length_in_code_units();) {
-        if (unicode || unicode_sets) {
-            auto code_point = code_point_at(utf16_pattern, i);
-            builder.append_code_point(code_point.code_point);
-            i += code_point.code_unit_count;
-            continue;
-        }
-
+    for (size_t i = 0; i < utf16_pattern.length_in_code_units(); ++i) {
```
shannonbooth, line 101: Just a question - but I had a thought from the commkt message - if the issue is LibRegex not being able to handle it, I wonder if this should instead be in LibRegex. We do have callers which do need to directly invoke LibRegex and expect the same result as what the ECMA implementation should

gmta: I agree - I could change the `// AD-HOC` to `// FIXME` for now to better indicate there's still an issue here? I haven't really been able to pinpoint the issue in LibRegex though.

alimpfard: iirc the parser in LibRegex uses GenericLexer, which can't handle unicode (or couldn't?)
Not too sure, but that's what I remember being the main cause.

alimpfard (approved):

Libraries/LibJS/Runtime/RegExpObject.cpp
gmta: That makes sense; I've updated the FIXME to reflect that.

AtkinsSJ: There was a PR working on a unicode GenericLexer. Can't remember what state that's in.

nico:
Follow-up fix in #5593

PR #5550, opened 2025-07-21
LibRegex: Compare code units (not code points) in non-Unicode char range

### LINKED ISSUE #5549: LibRegex: Surrogate pair matching with character classes does not match
For example:
```
> /[\ud800-\udbff][\udc00-\udfff]/.exec("\uD83D\uDE42")
null
```
Whereas Node produces:
```
> /[\ud800-\udbff][\udc00-\udfff]/.exec("\uD83D\uDE42")
[ '🙂', index: 0, input: '🙂', groups: undefined ]
```

Found on Discord, where it appears to be using Regex to match emoji, to then replace native emoji with Twemoji.

trflynn89:
Fixes #5549

No test262 diff.

This lets Discord render emoji as it sees fit.

Before:
before

After:
after

### DIFF
--- a/Libraries/LibRegex/RegexByteCode.cpp
+++ b/Libraries/LibRegex/RegexByteCode.cpp
@@ -116,7 +116,7 @@ static void advance_string_position(MatchState& state, RegexStringView view, Opt
 
     if (view.unicode()) {
         if (!code_point.has_value() && (state.string_position_in_code_units < view.length_in_code_units()))
-            code_point = view[state.string_position_in_code_units];
+            code_point = view.code_point_at(state.string_position_in_code_units);
         if (code_point.has_value())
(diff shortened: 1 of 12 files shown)


alimpfard (approved):
looks good, I assume you will be wanting to rename those at()/[] methods?

trflynn89:
> looks good, I assume you will be wanting to rename those at()/[] methods?

I think that will conflict will Jelle's PR so I think we can either handle that there or in a follow-up

gmta:
> > looks good, I assume you will be wanting to rename those at()/[] methods?
> 
> I think that will conflict will Jelle's PR so I think we can either handle that there or in a follow-up

Feel free to change them, I'll rebase mine later

alimpfard:
Much nicer, thanks!

PR #5552, opened 2025-07-21
LibWeb: Run tasks queued while parsing fragments

skyz1:
This fixes a bug where `toggle` events are not fired when details elements are inserted using `innerHtml` and immediately closed.

This issue still persists when using `DOMParser().parseFromString`. So maybe an alternative fix would be to set a `tasks_are_runnable` (or similar) flag for temporary documents and run all or at least DOMMutation tasks associated with these documents whether they are fully active or not. However I am not sure if this would break something.

### DIFF
--- a/Libraries/LibWeb/HTML/EventLoop/Task.cpp
+++ b/Libraries/LibWeb/HTML/EventLoop/Task.cpp
@@ -67,6 +67,11 @@ DOM::Document const* Task::document() const
     return m_document.ptr();
 }
 
+void Task::set_document(GC::Ptr<DOM::Document const> document)
+{
+    m_document = document;
+}
+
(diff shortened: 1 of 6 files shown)


AtkinsSJ:
I don't feel confident enough to say if this is a good solution or not, but I've marked it as not stale. It might be a spec bug, I've encountered bugs before involving fragment parsing.

skyz1:
> I don't feel confident enough to say if this is a good solution or not, but I've marked it as not stale. It might be a spec bug, I've encountered bugs before involving fragment parsing.

I think another solution would be to destroy the temporary document after parsing, but this has other side effect, so I was not sure if this would be the right thing to do.

PR #5555, opened 2025-07-21
LibWeb/HTML: Sanitize email input with multiple attribute

skyz1:
This implements the missing part of the value sanitization algorithm for email inputs with the multiple attribute.

### DIFF
--- a/Libraries/LibWeb/HTML/FormAssociatedElement.h
+++ b/Libraries/LibWeb/HTML/FormAssociatedElement.h
@@ -56,7 +56,7 @@ private:                                                                        
     {                                                                                                                                                                       \
         ElementBaseClass::attribute_changed(name, old_value, value, namespace_);                                                                                            \
         form_node_attribute_changed(name, value);                                                                                                                           \
-        form_associated_element_attribute_changed(name, value, namespace_);                                                                                                 \
+        form_associated_element_attribute_changed(name, old_value, value, namespace_);                                                                                      \
     }
 
 // https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#selection-direction
@@ -152,7 +152,7 @@ protected:
     virtual void form_associated_element_was_inserted() { }
     virtual void form_associated_element_was_removed(DOM::Node*) { }
     virtual void form_associated_element_was_moved(GC::Ptr<DOM::Node>) { }
-    virtual void form_associated_element_attribute_changed(FlyString const&, Optional<String> const&, Optional<FlyString> const&) { }
+    virtual void form_associated_element_attribute_changed(FlyString const&, Optional<String> const&, Optional<String> const&, Optional<FlyString> const&) { }
 
     void form_node_was_inserted();
     void form_node_was_removed();
--- a/Libraries/LibWeb/HTML/HTMLInputElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLInputElement.cpp
@@ -1380,7 +1380,7 @@ void HTMLInputElement::did_lose_focus()
     commit_pending_changes();
 }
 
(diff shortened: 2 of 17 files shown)


Libraries/LibWeb/HTML/HTMLInputElement.cpp
```diff
@@ -1657,18 +1660,34 @@ String HTMLInputElement::value_sanitization_algorithm(String const& value) const
...
+            if (value.bytes_as_string_view().contains('\r') || value.bytes_as_string_view().contains('\n')) {
+                StringBuilder builder;
+                for (auto c : value.bytes_as_string_view()) {
+                    if (c != '\r' && c != '\n')
+                        builder.append(c);
+                }
+                return MUST(String::from_utf8(builder.string_view().trim(Infra::ASCII_WHITESPACE)));
             }
-            return MUST(String::from_utf8(builder.string_view().trim(Infra::ASCII_WHITESPACE)));
+            return MUST(value.trim(Infra::ASCII_WHITESPACE));
```
tcl3: You can use `.to_ascii_whitespace()` here.

Libraries/LibWeb/HTML/HTMLInputElement.cpp
```diff
@@ -1657,18 +1660,34 @@ String HTMLInputElement::value_sanitization_algorithm(String const& value) const
...
         }
-        return MUST(value.trim(Infra::ASCII_WHITESPACE));
+
+        // https://html.spec.whatwg.org/multipage/input.html#email-state-(type=email):value-sanitization-algorithm-2
+        // 1. Split on commas the element's value, strip leading and trailing ASCII whitespace from each resulting token, if any,
+        //    and let the element's values be the (possibly empty) resulting list of (possibly empty) tokens, maintaining the original order.
+        Vector<String> values {};
+        for (auto const& token : MUST(value.split(','))) {
```
tcl3: You can use `value.split(',', SplitBehavior::KeepEmpty)` to preserve empty values.

Libraries/LibWeb/HTML/HTMLInputElement.cpp
```diff
@@ -1657,18 +1660,34 @@ String HTMLInputElement::value_sanitization_algorithm(String const& value) const
...
+            values.append(MUST(token.trim(Infra::ASCII_WHITESPACE)));
```
tcl3: You can use `.to_ascii_whitespace()` here.

Libraries/LibWeb/HTML/FormAssociatedElement.h
```diff
@@ -152,7 +152,7 @@ class FormAssociatedElement {
     virtual void form_associated_element_was_inserted() { }
     virtual void form_associated_element_was_removed(DOM::Node*) { }
     virtual void form_associated_element_was_moved(GC::Ptr<DOM::Node>) { }
-    virtual void form_associated_element_attribute_changed(FlyString const&, Optional<String> const&, Optional<FlyString> const&) { }
+    virtual void form_associated_element_attribute_changed(FlyString const&, Optional<String> const&, Optional<String> const&, Optional<FlyString> const&) { }
```
tcl3, line 155: It's fine to name parameters in header files when it isn't immediately clear what each argument is supposed to represent. I think adding parameter names here would make things more readable.

skyz1: Not sure if I understood that correctly. At this point, I can't add the names because the function body is empty. I have now added them to the other header files.

tcl3: Ah, of course - sorry. This can be left as is.

PR #5556, opened 2025-07-21
LibWeb: Support interpolating rotate values

tcl3:
This PR adds support for interpolating values of the `rotate` property, which are interpolated using spherical linear interpolation treating `none` as `0deg`

This improves the imported test, but there are a number of failures which are off by fractions of a degree, which leads me to believe there is a rounding issue that I've missed.

The spec would also leads me to think that we should probably be reusing some of the code in `interpolate_transform` to do this.

Currently gets us +185/-14 WPT subtests in the `css/css-transforms` directory.


AtkinsSJ (approved):
I am not a maths person, but this looks right to me and the test results speak for themselves. :sweat_smile: 

> This improves the imported test, but there are a number of failures which are off by fractions of a degree, which leads me to believe there is a rounding issue that I've missed.

Maybe using doubles everywhere would help? We do store doubles in the CSS types but in most calculations we reduce to floats. I don't know if that would make a large enough difference.

tcl3:
FWIW, I did try doing this and it didn't solve this particular issue. Converting numbers from double to float definitely introduces some precision loss, so it may not be a bad idea either way.

AtkinsSJ:
Alas, that would have been too simple. :pensive:

PR #5557, opened 2025-07-22
LibWeb: Return base Document for non-HTML parses

mikiubo:
The HTML specification does not explicitly require a specific return type for parseFromString(),
but according to Web Platform TestsDOMParser-parseFromString.html, the expected return value for
XML MIME types is a Document—not an XMLDocument.

### DIFF
--- a/Libraries/LibWeb/HTML/DOMParser.cpp
+++ b/Libraries/LibWeb/HTML/DOMParser.cpp
@@ -56,7 +56,7 @@ GC::Ref<DOM::Document> DOMParser::parse_from_string(StringView string, Bindings:
         document->parse_html_from_a_string(string);
     } else {
         // -> Otherwise
-        document = DOM::XMLDocument::create(realm(), associated_document.url());
+        document = DOM::Document::create(realm(), associated_document.url());
         document->set_content_type(Bindings::idl_enum_to_string(type));
         document->set_document_type(DOM::Document::Type::XML);
 
--- a/Tests/LibWeb/Text/expected/DOM/DOMParser-xml-document.txt
+++ b/Tests/LibWeb/Text/expected/DOM/DOMParser-xml-document.txt
@@ -1,1 +1,1 @@
-XMLDocument
+Document
--- a/Tests/LibWeb/Text/expected/wpt-import/domparsing/DOMParser-parseFromString-xml.txt
+++ b/Tests/LibWeb/Text/expected/wpt-import/domparsing/DOMParser-parseFromString-xml.txt
@@ -1,0 +1,25 @@
+Harness status: OK
+
+Found 20 tests
+
+20 Pass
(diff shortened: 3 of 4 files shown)


ADKaster:
This change causes at least `Tests/LibWeb/Text/input/DOM/DOMParser-xml-document.html` to fail. Could you run your branch against all of the dom/ wpt tests to see what's going on here? It seems like either our local test is wrong, or there's some inconsistency we're papering over somewhere else.

`./Meta/WPT.sh run dom` should do the trick (modulo extra args)

mikiubo:
> This change causes at least `Tests/LibWeb/Text/input/DOM/DOMParser-xml-document.html` to fail. Could you run your branch against all of the dom/ wpt tests to see what's going on here? It seems like either our local test is wrong, or there's some inconsistency we're papering over somewhere else.
> 
> `./Meta/WPT.sh run dom` should do the trick (modulo extra args)

this is the run result:
WithDocument


i also run without the commit:
WithXmlDocument

I add one observation:
all major browser fail this wpt test https://wpt.fyi/results/domparsing/DOMParser-parseFromString-xml.html
but for example flow 7.0 pass it. So i think there is something strange about the test.
Also there is a discussion on chrome about it:
https://issues.chromium.org/issues/41305457

let me know 
thanks

tcl3:
I had a quick look and there's no regressions in either `dom` or `domparsing`. 
Looking at the [WPT results](<https://wpt.fyi/results/domparsing/DOMParser-parseFromString-xml.html?label=experimental&label=master&aligned>) for the imported test, it looks like all major browsers return an `XMLDocument` rather than `Document` from `parseFromString` in this case, which disagrees with the test expectations.

I'm not sure whether the test is wrong or not, but if the imported test is correct then the failing test simply needs to be rebaselined, as it's testing exactly what you've changed in this PR.

mikiubo:
Ok i change the local test.
Let me know if you need anything else.
Thanks

PS: when i run ./Meta/WPT.sh run dom against the main branch i got the result above.
It is a problem of my local config or is normal?

tcl3:
> PS: when i run ./Meta/WPT.sh run dom against the main branch i got the result above.
> It is a problem of my local config or is normal?

The results you got are what I would expect. To compare WPT results this is what I did:

```sh
> git checkout master
> ./Meta/WPT.sh run --log results.log dom domparsing
> git checkout <this-pr-branch>
> ./Meta/WPT.sh compare results.log dom domparsing
```

mikiubo:
> The results you got are what I would expect. To compare WPT results this is what I did:
> 
> ```shell
> > git checkout master
> > ./Meta/WPT.sh run --log results.log dom domparsing
> > git checkout <this-pr-branch>
> > ./Meta/WPT.sh compare results.log dom domparsing
> ```

Thank you.

ADKaster (approved):

Issue #5547, opened 2025-07-21
AK/Utf8View: code_point_offset < length() verification failure when pasting emoji into a textbox or the js repl


Lubrsi:
For example, pasting 🙂into the Google search box or the JS repl crashes with:
```
VERIFICATION FAILED: code_point_offset < length() at /Users/lukewilde/Repositories/ladybird/AK/Utf8View.cpp:52
0   liblagom-ak.0.0.0.dylib             0x0000000102d716a8 dump_backtrace + 72
1   liblagom-ak.0.0.0.dylib             0x0000000102d71960 ak_trap + 12
2   liblagom-ak.0.0.0.dylib             0x0000000102d71a0c ak_assertion_failed + 0
3   liblagom-ak.0.0.0.dylib             0x0000000102db3cf4 AK::Utf8CodePointIterator::peek(unsigned long) const (.cold.1) + 0
4   liblagom-ak.0.0.0.dylib             0x0000000102dafdcc AK::Utf8View::byte_offset_of(unsigned long) const + 552
5   liblagom-line.0.0.0.dylib           0x00000001028c7108 Line::Editor::handle_read_event() + 5344
6   liblagom-line.0.0.0.dylib           0x00000001028c5b74 Line::Editor::try_update_once() + 80
7   liblagom-line.0.0.0.dylib           0x00000001028d3374 AK::Function<void ()>::CallableWrapper<Line::Editor::get_line(AK::ByteString const&)::$_0>::call() + 48
8   liblagom-core.0.0.0.dylib           0x00000001029c6120 Core::Notifier::event(Core::Event&) + 136
9   liblagom-core.0.0.0.dylib           0x00000001029bcf90 Core::EventReceiver::dispatch_event(Core::Event&, Core::EventReceiver*) + 84
10  liblagom-core.0.0.0.dylib           0x00000001029ca170 Core::ThreadEventQueue::process() + 428
11  liblagom-core.0.0.0.dylib           0x00000001029cf794 Core::EventLoopImplementationUnix::exec() + 44
12  liblagom-core.0.0.0.dylib           0x00000001029bb424 Core::EventLoop::exec() + 68
13  liblagom-line.0.0.0.dylib           0x00000001028c3f04 Line::Editor::get_line(AK::ByteString const&) + 1268
14  js                                  0x0000000102814c44 ladybird_main(Main::Arguments) + 3852
15  js                                  0x000000010281b9c0 main + 196
[...]

PR #5558, opened 2025-07-22
AK: Allow indexing at length in Utf8View::byte_offset_of() and ::code_point_offset_of()

gmta:
Some of these `VERIFY`s of the view's length were introduced recently, but they caused the parsing of named capture groups in RegexParser to crash in some situations.

Instead, allow indexing at the view's length: the byte offset of code point `length()` is known, even though that code point does not exist in the view. Similarly, we know the code point offset at byte offset `byte_length()`. Beyond those offsets, we still crash.

Fixes 13 failures in test262's `language/literals/regexp/named-groups`.

### DIFF
--- a/AK/Utf8View.cpp
+++ b/AK/Utf8View.cpp
@@ -32,24 +32,25 @@ Utf8CodePointIterator Utf8View::iterator_at_byte_offset_without_validation(size_
 
 size_t Utf8View::code_point_offset_of(size_t byte_offset) const
 {
-    VERIFY(byte_offset < byte_length());
+    VERIFY(byte_offset <= byte_length());
 
     // Fast path: each code point is represented by a single byte.
     if (length() == byte_length())
         return byte_offset;
 
     size_t code_point_offset = 0;
-    for (auto it = begin(); !it.done(); ++it) {
-        if (it.m_ptr > begin_ptr() + byte_offset)
+    auto it = begin();
+    while (!it.done()) {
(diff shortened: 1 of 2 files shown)


AK/Utf8View.cpp
```diff
@@ -32,24 +32,25 @@ Utf8CodePointIterator Utf8View::iterator_at_byte_offset_without_validation(size_
...
 
     // Fast path: each code point is represented by a single byte.
     if (length() == byte_length())
         return byte_offset;
 
     size_t code_point_offset = 0;
-    for (auto it = begin(); !it.done(); ++it) {
-        if (it.m_ptr > begin_ptr() + byte_offset)
+    auto it = begin();
+    while (!it.done()) {
```
trflynn89, line 43: welp this made me realize how wrong `Utf16View::code_point_offset_of` is :sweat_smile: 

FYI we currently don't assert on the length in `Utf16Vew` because the JS spec purposefully throws an index larger than the length at it by way of `AdvanceStringIndex`:

https://tc39.es/ecma262/#sec-advancestringindex
```c++
// 4. If index + 1 ≥ length, return index + 1.
if (index + 1 >= string.length_in_code_units())
    return index + 1;
```

But maybe we should revisit that and add some protection at the caller.

trflynn89 (approved):

trflynn89:
Oh also does this fix or partially address #5547?
It looks like LibLine throws length+1 at the view :thinking:

Issue #3840, opened 2025-03-06
Assertion failed  is<HTML::Window>(global_object) on NavigationTiming/PerformanceTiming.cpp:30 on web.whatsapp.com


teaalltr:
### Summary

This shows up when trying to load web.whatsapp.com, it's an assertion failed `is<HTML::Window>(global_object)`.

I'm able to reproduce it multiple times, doing the procedure below

### Operating system

Linux

### Steps to reproduce

1. Go to web.whatsapp.com (from ladybird command line,  `./Meta/ladybird.sh run ladybird https://web.whatsapp.com/`
2. It starts loading the QR code page
3. But it can't

### Expected behavior

Page should load :-P

### Actual behavior

Page not load

### URL for a reduced test case

None

### HTML/SVG/etc. source for a reduced test case

```html
None
```

### Log output and (if possible) backtrace

```shell
2404.381 WebContent(28867): FIXME: Unimplemented IDL interface: 'IDBFactoryPrototype.databases'
2404.384 WebContent(28867): FIXME: Unimplemented IDL interface: 'IDBFactoryPrototype.databases'
2404.528 WebContent(28867): FIXME: Unimplemented IDL interface: 'IDBTransactionPrototype.db'
2404.821 WebContent(28867): FIXME: Unimplemented IDL interface: 'ServiceWorkerContainerPrototype.controller'
2404.821 WebContent(28867): FIXME: Unimplemented IDL interface: 'ServiceWorkerContainerPrototype.controller'
2404.878 WebContent(28867): FIXME: Unimplemented IDL interface: 'ServiceWorkerContainerPrototype.controller'
2405.959 WebContent(28867): FIXME: InlineFormattingContext::dimension_box_on_line got unexpected box in inline context:
2405.959 WebContent(28867): Label <label.x17fgdl5.x1f6kntn.xt0psk2> (not painted) children: inline
  TextNode <#text>

2405.959 WebContent(28867): FIXME: InlineFormattingContext::dimension_box_on_line got unexpected box in inline context:
2405.959 WebContent(28867): Label <label.x17fgdl5.x1f6kntn.xt0psk2> (not painted) children: inline
  TextNode <#text>

2405.960 WebContent(28867): FIXME: InlineFormattingContext::dimension_box_on_line got unexpected box in inline context:
2405.960 WebContent(28867): Label <label.x17fgdl5.x1f6kntn.xt0psk2> (not painted) children: inline
  TextNode <#text>

[...]

PR #5565, opened 2025-07-22
LibWeb: Only expose performance.{timing,navigation} on Window

Lubrsi:
They are only valid in a Window context, and the spec only exposes them in Window contexts.

Fixes workers crashing on:
- https://web.whatsapp.com/
- https://pro.kraken.com/app/trade/btc-usd


Libraries/LibWeb/HighResolutionTime/Performance.idl
```diff
@@ -37,8 +37,8 @@ interface Performance : EventTarget {
 
     // https://w3c.github.io/navigation-timing/#extensions-to-the-performance-interface
     // Obsolete "Navigation Timing" extensions to the Performance interface
-    [SameObject]
+    [SameObject, Exposed=Window]
```
trflynn89: Theoretically - we now support `partial interface` after https://github.com/LadybirdBrowser/ladybird/pull/5559, so we could yoink these APIs to their own file and use the literal text from the spec?
```
[Exposed=Window]
partial interface Performance {
  [SameObject]
  readonly attribute PerformanceTiming timing;
  [SameObject]
  readonly attribute PerformanceNavigation navigation;
};
```

ADKaster:
iirc this affects #https://github.com/LadybirdBrowser/ladybird/issues/3840

Libraries/LibWeb/HighResolutionTime/Performance.idl
gmta: Note that #5559 is pretty simplistic, you'd need to test whether or not `[Exposed=..]` actually gets transferred to attributes if it's defined on the partial interface level.

Lubrsi: I tried it out and it seems extended attributes don't get transferred to the attributes. I tried doing this, but it didn't seem to help:
```patch
diff --git a/Libraries/LibIDL/Types.cpp b/Libraries/LibIDL/Types.cpp
index aaaf21e86d..56b168d1fe 100644
--- a/Libraries/LibIDL/Types.cpp
+++ b/Libraries/LibIDL/Types.cpp
@@ -307,11 +307,39 @@ void EffectiveOverloadSet::remove_all_other_entries()
 
 void Interface::extend_with_partial_interface(Interface const& partial)
 {
-    attributes.extend(partial.attributes);
-    static_attributes.extend(partial.static_attributes);
+    for (auto const& attribute : partial.attributes) {
+        auto attribute_copy = attribute;
+        for (auto const& [key, value] : partial.extended_attributes) {
+            attribute_copy.extended_attributes.set(key, value);
+        }
+        attributes.append(move(attribute_copy));
+    }
+
+    for (auto const& static_attribute : partial.static_attributes) {
+        auto static_attribute_copy = static_attribute;
+        for (auto const& [key, value] : partial.extended_attributes) {
+            static_attribute_copy.extended_attributes.set(key, value);
+        }
+        static_attributes.append(move(static_attribute_copy));
+    }
+
     constants.extend(partial.constants);
-    functions.extend(partial.functions);
-    static_functions.extend(partial.static_functions);
+
+    for (auto const& function : partial.functions) {
+        auto function_copy = function;
+        for (auto const& [key, value] : partial.extended_attributes) {
+            function_copy.extended_attributes.set(key, value);
+        }
+        functions.append(move(function_copy));
+    }
+
+    for (auto const& static_function : partial.static_functions) {
+        auto static_function_copy = static_function;
+        for (auto const& [key, value] : partial.extended_attributes) {
+            static_function_copy.extended_attributes.set(key, value);
+        }
+        static_functions.append(move(static_function_copy));
+    }
 }
 
 }
 ```

PR #5571, opened 2025-07-22
LibRegex: Fix capture groups in quantified alternations

aplefull:
This prevents empty matches from overwriting non-empty captures in quantified alternations. Fixes patterns like `(a|a?)+` where the optional branch would incorrectly overwrite meaningful captures with empty strings.

No test262 diff.

### DIFF
--- a/Tests/LibRegex/TestRegex.cpp
+++ b/Tests/LibRegex/TestRegex.cpp
@@ -1330,6 +1330,29 @@ TEST_CASE(optimizer_repeat_offset)
     }
 }
 
+TEST_CASE(quantified_alternation_capture_groups)
+{
+    {
+        // Ensure that (a|a?)+ captures the last meaningful match, not empty string
+        Regex<ECMA262> re("^(a|a?)+$");
+        auto result = re.match("a"sv);
+
+        EXPECT_EQ(result.success, true);
+        EXPECT_EQ(result.matches.size(), 1u);
+        EXPECT_EQ(result.matches.first().view.to_byte_string(), "a"sv);
+        EXPECT_EQ(result.capture_group_matches.first()[0].view.to_byte_string(), "a"sv);
+    }
+    {
+        Regex<ECMA262> re("^(a|a?)+$");
+        auto result = re.match("aa"sv);
+
+        EXPECT_EQ(result.success, true);
+        EXPECT_EQ(result.matches.size(), 1u);
+        EXPECT_EQ(result.matches.first().view.to_byte_string(), "aa"sv);
+        EXPECT_EQ(result.capture_group_matches.first()[0].view.to_byte_string(), "a"sv);
+    }
+}
+
 TEST_CASE(zero_width_backreference)
 {
     {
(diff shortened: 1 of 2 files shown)


Tests/LibRegex/TestRegex.cpp
```diff
@@ -1330,6 +1330,29 @@ TEST_CASE(optimizer_repeat_offset)
     }
 }
 
+TEST_CASE(quantified_alternation_capture_groups)
+{
+    {
+        // Ensure that (a|a?)+ captures the last meaningful match, not empty string
+        Regex<ECMA262> re("^(a|a?)+$");
```
alimpfard, line 1337: 🤔 this seems very weird to me, if `a?` matched last then that should be the contents of the group, is this specified somewhere?
it does seem like other engines do this as well, so I'm okay with merging this anyway.

aplefull: I'm also confused about this tbh. I tried to search for this behaviour in spec and test262, but couldn't find anything. From my limited understanding this is what should happen:

1. Left `"a"` matches `"a"` and we store it.
2. `+` attempts a second iteration -> `"a"` doesn't match against `""` -> `"a?`" matches against `""` -> we save it to a capture group.
3. Final state is ["a", ""]

But in practice, both V8 and SpiderMonkey store the last meaningful match instead of the empty string for any input I've tried to construct, so I decided to do the same for compatibility. Getting a comment from someone who works on the spec or test262 would be nice, but I have no idea where to ask 😅

alimpfard: I went looking in the spec, and the behaviour is actually very weird
https://tc39.es/ecma262/#sec-runtime-semantics-repeatmatcher-abstract-operation actually has `a?` fail to match (via https://tc39.es/ecma262/#step-repeatmatcher-done), but take the continuation anyway.

so in the spec's version of reality what happens is:
- (a) RepeatMatcher for `(a|a?)` with min=1 max=inf greedy=true
    - (b) MatchTwoAlternatives with m1=`a` m2=`a?`
        - (c) `a`
        - (d) RepeatMatcher for `a` with min=0 max=1 greedy=true

So the continuation chain is:
```
a(x, c0) = match b(clear_captures x, c1)
        FAIL = c0(x)
        z = z
    c1(y) = RepeatMatcher for `(a|a?)` with min=0 max=inf greedy=true
    b(x, c0) = match c(x, c0)
            FAIL = d(x, c0)
            z = z
    d(x, c0) = match c(x, c2)
            FAIL = c0(x)
            z = z
        where c2(y) = if x.index = y.index then FAIL else c0(x)
```

so...given that this only affects the captures in `x`, we're good to go and what you have _is_ the correct effective behaviour.

alimpfard: Maybe paste a link to that as a note in the code lol

aplefull: Ooh, I would never notice that myself in a million years. I'll add a note now

PR #5574, opened 2025-07-23
LibWeb: Convert Editing API internals to UTF-16

gmta:
Both sides of the Editing internals now have to deal with some awkward converting between UTF-8 and UTF-16, but the upside is that it immediately exposed an issue with the `insertText` command: instead of dealing with code units, it was iterating over code points causing the selection to be updated only once instead of twice. This resulted in the final selection potentially ending up in between a surrogate pair.

Fixes https://github.com/LadybirdBrowser/ladybird/issues/5547 (pasting/typing surrogate pairs).

### DIFF
--- a/AK/Utf16View.h
+++ b/AK/Utf16View.h
@@ -381,7 +381,7 @@ public:
         return substring_view(substring_start, substring_length);
     }
 
-    [[nodiscard]] constexpr Utf16View trim_whitespace(TrimMode mode = TrimMode::Both) const
+    [[nodiscard]] constexpr Utf16View trim_ascii_whitespace(TrimMode mode = TrimMode::Both) const
     {
         static constexpr Utf16View white_space { u" \n\t\v\f\r", 6uz };
         return trim(white_space, mode);
--- a/AK/Utf16View.h
+++ b/AK/Utf16View.h
@@ -463,17 +463,10 @@ public:
         if (needle.length_in_code_units() > length_in_code_units())
             return false;
 
-        if (has_ascii_storage() && needle.has_ascii_storage()) {
-            if (m_string.ascii == needle.m_string.ascii)
-                return true;
+        if (has_ascii_storage() && needle.has_ascii_storage())
             return ascii_span().starts_with(needle.ascii_span());
-        }
-
-        if (!has_ascii_storage() && !needle.has_ascii_storage()) {
-            if (m_string.utf16 == needle.m_string.utf16)
-                return true;
+        if (!has_ascii_storage() && !needle.has_ascii_storage())
             return utf16_span().starts_with(needle.utf16_span());
-        }
 
         for (auto this_it = begin(), needle_it = needle.begin(); needle_it != needle.end(); ++needle_it, ++this_it) {
             if (*this_it != *needle_it)
--- a/AK/Utf16View.h
+++ b/AK/Utf16View.h
@@ -456,24 +456,24 @@ public:
 
     [[nodiscard]] constexpr bool starts_with(Utf16View const& needle) const
     {
-        if (needle.is_empty())
+        auto needle_length = needle.length_in_code_units();
+        if (needle_length == 0)
             return true;
-        if (is_empty())
-            return false;
-        if (needle.length_in_code_units() > length_in_code_units())
+        if (is_empty() || needle_length > length_in_code_units())
             return false;
 
-        if (has_ascii_storage() && needle.has_ascii_storage())
-            return ascii_span().starts_with(needle.ascii_span());
-        if (!has_ascii_storage() && !needle.has_ascii_storage())
-            return utf16_span().starts_with(needle.utf16_span());
+        return substring_view(0, needle_length) == needle;
+    }
 
-        for (auto this_it = begin(), needle_it = needle.begin(); needle_it != needle.end(); ++needle_it, ++this_it) {
-            if (*this_it != *needle_it)
-                return false;
-        }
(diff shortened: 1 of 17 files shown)


Libraries/LibGfx/Color.h
```diff
@@ -491,6 +491,7 @@ class Color {
     ByteString to_byte_string() const;
     ByteString to_byte_string_without_alpha() const;
     static Optional<Color> from_string(StringView);
+    static Optional<Color> from_utf16_string(Utf16View);
```
trflynn89: `Utf16View` (like `Utf8View`) should be passed by `const&`, it's a bit big to pass by value.

AK/Utf16View.h
```diff
@@ -456,31 +456,24 @@ class Utf16View {
...
-                return true;
-            return ascii_span().starts_with(needle.ascii_span());
-        }
-
-        if (!has_ascii_storage() && !needle.has_ascii_storage()) {
-            if (m_string.utf16 == needle.m_string.utf16)
-                return true;
-            return utf16_span().starts_with(needle.utf16_span());
-        }
+        return substring_view(0, needle_length) == needle;
```
trflynn89, line 465: This does lose an optimization where we checked if the pointers were the same after checking the length. That was based on:
https://github.com/LadybirdBrowser/ladybird/blob/265e278275f699859092bcb2743c083591678ff2/AK/Utf8View.cpp#L117-L118

I don't know if that was particularly significant though, because it appears other string APIs don't do this either, and it was just part of the initial commit to add `Utf8View::starts_with` (edecf8f6a375b04e60e7c4ddcc10b40f350a5d77).

Libraries/LibWeb/Editing/Commands.h
```diff
@@ -32,62 +32,62 @@ struct CommandDefinition {
...
-bool command_create_link_action(DOM::Document&, String const&);
-bool command_default_paragraph_separator_action(DOM::Document&, String const&);
-String command_default_paragraph_separator_value(DOM::Document const&);
-bool command_delete_action(DOM::Document&, String const&);
-bool command_font_name_action(DOM::Document&, String const&);
-bool command_font_size_action(DOM::Document&, String const&);
-String command_font_size_value(DOM::Document const&);
-bool command_fore_color_action(DOM::Document&, String const&);
-bool command_format_block_action(DOM::Document&, String const&);
+bool command_back_color_action(DOM::Document&, Utf16View);
```
trflynn89: These and elsewhere should be `const&` as well.

Libraries/LibWeb/Editing/Commands.cpp
```diff
@@ -509,37 +509,37 @@ enum class FontSizeMode : u8 {
 };
 
 // https://w3c.github.io/editing/docs/execCommand/#the-fontsize-command
-bool command_font_size_action(DOM::Document& document, String const& value)
+bool command_font_size_action(DOM::Document& document, Utf16View value)
 {
     // 1. Strip leading and trailing whitespace from value.
-    auto resulting_value = MUST(value.trim_ascii_whitespace());
+    auto resulting_value = value.trim_whitespace();
```
trflynn89: oh maybe we should rename `Utf16View::trim_whitespace` to `Utf16View::trim_ascii_whitespace` to make it clear it's ASCII since, like `String`, it's supposed to be Unicode-aware unless otherwise mentioned.

Libraries/LibWeb/Editing/Internal/Algorithms.cpp
```diff
@@ -4388,7 +4388,7 @@ bool values_are_equivalent(FlyString const& command, Optional<String> a, Optiona
 }
 
 // https://w3c.github.io/editing/docs/execCommand/#loosely-equivalent-values
-bool values_are_loosely_equivalent(FlyString const& command, Optional<String> a, Optional<String> b)
+bool values_are_loosely_equivalent(FlyString const& command, Optional<Utf16View> a, Optional<Utf16View> b)
```
trflynn89: It seems every caller to this function is casting from `Optional<Utf16String>` to `Optional<Utf16View>` at the call site. Maybe this should just take the former?

BTW I think all the `Optional<String>` / `Optional<Utf16String>` could be `Optional<String const&>` and `Optional<Utf16String const&>` to avoid ref-count churn.

trflynn89 (commented):
Very cool!

AK/Utf16View.h
gmta: Oh, there is a nice way to get that back: we can reintroduce the pointer check in `TypedTransfer<T>::compare()`, which does not need to perform any comparisons if it was passed two identical pointers.

Libraries/LibWeb/Editing/Internal/Algorithms.cpp
gmta: > It seems every caller to this function is casting from Optional<Utf16String> to Optional<Utf16View> at the call site. Maybe this should just take the former?

The function is designed to take two arbitrary string values, but in practice the first argument was always an `Optional<Utf16String>` and the second one an `Optional<Utf16View>`... I've updated it to always take two `Optional<Utf16String const&>`.

gmta: > BTW I think all the Optional<String> / Optional<Utf16String> could be Optional<String const&> and Optional<Utf16String const&> to avoid ref-count churn.

Agreed, done!

trflynn89 (approved):

Issue #3572, opened 2025-02-14
Minimal repro for flex button layout bug on twitch.tv


Psychpsyo:
### Summary

When I open twitch.tv in Ladybird, the buttons in the cookie banner at the top are too small for their text, making it impossible to read.

### Operating system

Linux

### Steps to reproduce

1. Open the reduced test case.
2. Notice that the text overflows the button.
3. Verify that this does not happen in Firefox or Chrome.

### Expected behavior

The button should grow to contain the text.

### Actual behavior

The button does not grow to contain the text.

### URL for a reduced test case

N/A

### HTML/SVG/etc. source for a reduced test case

```html
<!DOCTYPE html>
<style>
body {
  display: flex;
  max-width: 50px;
}
button {
  display: inline-flex;
  justify-content: center;
  width: 100%;
}
</style>
<div>
  <button>
    LOOOOOOOOOOOOOOOOOOOOONG
  </button>
</div>
```

### Log output and (if possible) backtrace

```shell
none
```

### Screenshots or screen recordings

Top: Ladybird
Bottom: Firefox
![Image](https://github.com/user-attachments/assets/8569c633-5af6-4b71-a6ef-258c6f2fe5da)

### Build flags or config settings

none

### Contribute a patch?

- [ ] I’ll contribute a patch for this myself.

Kuba314:
This is a smaller minimal reproducible example, which still produces the same difference between Firefox and Ladybird for me:
```html
<!DOCTYPE html>
<style>
body {
  display: flex;
  max-width: 50px;
}
button { 
  width: 100%;
}
</style>
<div>
  <button>
    LOOOOOOOOOOOOOOOOOOOOONG
  </button>
</div>
```

It seems that Ladybird honors the `body`'s max-width, while Firefox doesn't and enables the button to grow to fit the text.

PR #5579, opened 2025-07-23
LibWeb: Don't treat non-replaced sizes as 0 for min-content contrib

awesomekling:
This behavior is part of the cyclic percentage contribution logic from
CSS-SIZING-3 which explicitly only applies to non-replaced boxes.

This fixes an issue on Discord where buttons in the settings UI were
cropped to a narrower width than intended.

Before:
1before

After:
1after


Psychpsyo:
This looks like it might fix https://github.com/LadybirdBrowser/ladybird/issues/3572

awesomekling:
> This looks like it might fix #3572

It totally does! Noted this in the commit message, and added the test case from that bug as a layout test as well, thank you! :)

PR #5582, opened 2025-07-23
LibWeb: Don't import raw X25519 key of incorrect length

hopafoot:
Gets 4 more WPT subtests related to rejecting imports of bad X25519 keys in raw format.

[wpt tests](https://wpt.live/WebCryptoAPI/import_export/okp_importKey_failures_X25519.https.any.worker.html)

Screenshot 2025-07-23 at 3 33 21 PM


Libraries/LibWeb/Crypto/CryptoAlgorithms.cpp
```diff
@@ -7258,19 +7258,23 @@ WebIDL::ExceptionOr<GC::Ref<CryptoKey>> X25519::import_key([[maybe_unused]] Web:
         if (!usages.is_empty())
             return WebIDL::SyntaxError::create(m_realm, "Usages must be empty"_string);
 
-        // 2. Let algorithm be a new KeyAlgorithm object.
+        // 2. If the the key length is not 32 bytes then throw an error
+        if (32 != key_data.get<ByteBuffer>().size())
+            return WebIDL::DataError::create(m_realm, "X25519 key must be 32 bytes"_string);
```
trflynn89: The comments here weren't made up by us, they are the literal text from the spec:
https://w3c.github.io/webcrypto/#x25519-operations-import-key

So we should not invent new spec text here. Is there somewhere else the spec mentions an explicit check on the data size? If not, this sounds like it could be a bug with the spec itself.

Libraries/LibWeb/Crypto/CryptoAlgorithms.cpp
```diff
@@ -7258,19 +7258,23 @@ WebIDL::ExceptionOr<GC::Ref<CryptoKey>> X25519::import_key([[maybe_unused]] Web:
...
-        // 4. Let key be a new CryptoKey associated with the relevant global object of this [HTML], and representing the key data provided in keyData.
+        // 5. Let key be a new CryptoKey associated with the relevant global object of this [HTML], and representing the key data provided in keyData.
         key = CryptoKey::create(m_realm, key_data);
 
-        // 5. Set the [[type]] internal slot of key to "public"
+        // 6. Set the [[type]] internal slot of key to "public"
         key->set_type(Bindings::KeyType::Public);
 
-        // 6. Set the [[algorithm]] internal slot of key to algorithm.
+        // 7. Set the [[algorithm]] internal slot of key to algorithm.
```
awesomekling: The step numbers here are taken directly from the spec: <https://w3c.github.io/webcrypto/#x25519-operations-import-key>

We should not change them. Instead, if we need to perform some work that isn't in the spec, we can mark it with an "AD-HOC" comment, i.e

```cpp
		// AD-HOC: If the the key length is not 32 bytes then throw an error
        if (32 != key_data.get<ByteBuffer>().size())
            return WebIDL::DataError::create(m_realm, "X25519 key must be 32 bytes"_string);
```

Also, if this is an error in the spec, please report a bug against it over in https://github.com/w3c/webcrypto - then we can reference the bug in a comment here as well, until it gets resolved :)

Libraries/LibWeb/Crypto/CryptoAlgorithms.cpp
hopafoot: Sorry for the lack of awareness and thanks for the feedback. I've opened an issue with w3c to get clarification.

Libraries/LibWeb/Crypto/CryptoAlgorithms.cpp
hopafoot: Thanks for pointing me in the right direction, I've opened an [issue with w3c](https://github.com/w3c/webcrypto/issues/409) to get some clarification.

Considering the dearth of Ad-HOC comments elsewhere in this file I will avoid setting a new trend.

ADKaster: fwiw, force-pushing an update that uses the structure 

```
// AD-HOC: If the key length is not xyz throw an error, see https://github.com/w3c/specname/issueLink
if (32 != .... )
   return FooError::create(...);
```  

would be reasonable for the purposes of this PR. We don't need to wait for a reply for obvious spec issues like this, esp when there's WPTs to back it up.

hopafoot: @ADKaster @awesomekling, I've done as asked. I can update the comment once the issue is resolved.

ADKaster:
Well I say that. but can you rebaseline/update the failing test?

 `Fail: Text/input/wpt-import/WebCryptoAPI/import_export/okp_importKey_failures_X25519.https.any.html`

hopafoot:
@ADKaster, done. I think I need approval to rerun CI, though. Sorry for missing this before, this is my first contribution to this project.

PR #5583, opened 2025-07-23
LibWeb/HTML: Only include direction if dirname applies

skyz1:
This commit changes form data to only include the direction of auto directionality form associated elements.

This fixes [this test](http://wpt.live/html/semantics/forms/attributes-common-to-form-controls/dirname-only-if-applies.html), but I struggled to import the test. It submits the form in an iframe to check the search parameters, but when running the test from a file, this fails because the iframe is a cross-origin object. Changing the action attribute of the form from `resources/dirname-iframe.html` to `.` allows the test to run, but I am not sure if it is fine to change imported WPT tests.

### DIFF
--- a/Libraries/LibWeb/HTML/FormControlInfrastructure.cpp
+++ b/Libraries/LibWeb/HTML/FormControlInfrastructure.cpp
@@ -189,10 +189,10 @@ WebIDL::ExceptionOr<Optional<Vector<XHR::FormDataEntry>>> construct_entry_list(J
             entry_list.append(TRY(create_entry(realm, name.to_string(), control_as_form_associated_element->value())));
         }
 
-        // 11. If the element has a dirname attribute, and that attribute's value is not the empty string, then:
-        if (auto attribute = control->get_attribute(HTML::AttributeNames::dirname); attribute.has_value() && !attribute.value().is_empty()) {
+        // 11. If the element has a dirname attribute, that attribute's value is not the empty string, and the element is an auto-directionality form-associated element:
+        if (auto attribute = control->get_attribute(HTML::AttributeNames::dirname); attribute.has_value() && !attribute.value().is_empty() && control->is_auto_directionality_form_associated_element()) {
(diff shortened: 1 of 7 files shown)


tcl3 (commented):
> This fixes [this test](http://wpt.live/html/semantics/forms/attributes-common-to-form-controls/dirname-only-if-applies.html), but I struggled to import the test. It submits the form in an iframe to check the search parameters, but when running the test from a file, this fails because the iframe is a cross-origin object. Changing the action attribute of the form from `resources/dirname-iframe.html` to `.` allows the test to run, but I am not sure if it is fine to change imported WPT tests.

It looks like the test fails because the importer doesn't download the file pointed to by the `action` attribute. You can download it manually by running: `curl http://wpt.live/html/semantics/forms/attributes-common-to-form-controls/resources/dirname-iframe.html > Tests/LibWeb/Text/input/wpt-import/html/semantics/forms/attributes-common-to-form-controls/resources/dirname-iframe.html`. With this file present, the test should be able to run unmodified.

Libraries/LibWeb/HTML/FormControlInfrastructure.cpp
```diff
@@ -189,10 +189,10 @@ WebIDL::ExceptionOr<Optional<Vector<XHR::FormDataEntry>>> construct_entry_list(J
             entry_list.append(TRY(create_entry(realm, name.to_string(), control_as_form_associated_element->value())));
         }
 
-        // 11. If the element has a dirname attribute, and that attribute's value is not the empty string, then:
-        if (auto attribute = control->get_attribute(HTML::AttributeNames::dirname); attribute.has_value() && !attribute.value().is_empty()) {
+        // 11. If the element has a dirname attribute, and that attribute's value is not the empty string, and the element is an auto-directionality form-associated element, then:
```
gmta: Please copy the spec text as linked, i.e. without the ", then" part

tcl3 (approved):
Thanks for taking the time to fix the importer.

PR #5585, opened 2025-07-23
Documentation: Add instructions on how to build the flatpak

ADKaster:
cc @trflynn89  @JanKoudijs

### DIFF
--- a/Documentation/AdvancedBuildInstructions.md
+++ b/Documentation/AdvancedBuildInstructions.md
@@ -100,6 +100,37 @@ By default, ccache will include the plugins themselves in file hashes. So if a p
 will change, and you will be stuck with an uncached build. This setting will prevent ccache from using plugins in the
 file hashes.
 
+## Building the Flatpak
+
+Ladybird has an in-tree Flatpak manifest that can be used to build a Flatpak package for Ladybird. The recommended way
+to build the Flatpak is to use the `flatpak-builder` tool, which is available in most distributions. See the
+[Flatpak setup documentation](https://flatpak.org/setup/) on how to configure your environment for user Flatpak
+builds, and to configure the Flathub repo. The Ladybird Flatpak manifest at
+`Meta/CMake/flatpak/org.ladybird.Ladybird.json` can be built with the following command:
+
+```bash
+flatpak-builder --user --force-clean --install-deps-from=flathub \
+  --ccache --repo=Build/repo --install Build/flatpak \
+  Meta/CMake/flatpak/org.ladybird.Ladybird.json 
+```
+
+This command will build the Flatpak bundle and install it into the local Flatpak repository at `Build/repo`. Expect this
+to take a long time, as it will download and build all the dependencies of Ladybird as well. `flatpak-builder` will drop
+caches and build files in `.flatpak-builder`, as well as in the two subfolders `Build/repo` and `Build/flatpak`.
+
+The Flatpak can be run with the following command:
+
+```bash
+flatpak run --user org.ladybird.Ladybird
+```
+
+In order to debug the Flatpak, you can pass the `--devel` and `--command=sh` flags to `flatpak run`. This will drop you
+into a shell in the Flatpak sandbox:
+
+```bash
+flatpak run --user --command=sh --devel org.ladybird.Ladybird
+```
+
 ## Debugging without any optimizations
 
 It’s possible that when trying to inspect certain frame variables in your debugger, you’ll get an error similar to the following:


AtkinsSJ:
Commit message typo: "Documetation". Otherwise seems fine to me

Documentation/AdvancedBuildInstructions.md
```diff
@@ -100,6 +100,38 @@ By default, ccache will include the plugins themselves in file hashes. So if a p
 will change, and you will be stuck with an uncached build. This setting will prevent ccache from using plugins in the
 file hashes.
 
+## Building the flatpak
```
JanKoudijs: Should Flatpak be used instead, for the occurrences in text? To bring it in line with how Ladybird and CMake are capitalized according to the project naming.

Documentation/AdvancedBuildInstructions.md
```diff
@@ -100,6 +100,38 @@ By default, ccache will include the plugins themselves in file hashes. So if a p
...
+
+Ladybird has an in-tree flatpak manifest that can be used to build a flatpak package for Ladybird. The recommended way
+to build the flatpak is to use the `flatpak-builder` tool, which is available in most distributions. See the
+[flatpak setup documentation](https://flatpak.org/setup/) on how to configure your environment for user flatpak
+builds, and to configure the flathub repo. The Ladybird flatpak manifest at
```
JanKoudijs: Should Flathub be used instead? To bring it in line with how Ladybird and CMake are capitalized according to the project naming.

JanKoudijs:
One related question:

I'm fan of the [Flatpak artifacts](https://github.com/LadybirdBrowser/ladybird/actions/runs/16484738954#artifacts) we now have, but wonder if this is intentional, as there were no artifacts before. If it is intentional, then maybe they should also be documented.

Documentation/AdvancedBuildInstructions.md
```diff
@@ -100,6 +100,38 @@ By default, ccache will include the plugins themselves in file hashes. So if a p
 will change, and you will be stuck with an uncached build. This setting will prevent ccache from using plugins in the
```
gmta, line 100: Commit title typo: "Documetation"

Documentation/AdvancedBuildInstructions.md
```diff
@@ -100,6 +100,38 @@ By default, ccache will include the plugins themselves in file hashes. So if a p
...
+
+```bash
+flatpak-builder --user --force-clean --install-deps-from=flathub \
+  --ccache --repo=Build/repo --install Build/flatpak \
+  Meta/CMake/flatpak/org.ladybird.Ladybird.json 
+```
+
+This command will build the flatpak bundle and install it into the local flatpak repository at `Build/repo`. Expect this
+to take a long time, as it will download and build all the dependencies of Ladybird as well. `flatpak-builder` will drop
+caches and  build files in `.flatpak-builder`, as well as in the two subfolders `Build/repo` and `Build/flatpak`.
```
gmta: Double space between "and" and "build"

Documentation/AdvancedBuildInstructions.md
```diff
@@ -100,6 +100,38 @@ By default, ccache will include the plugins themselves in file hashes. So if a p
...
+
+The flatpak can be run with the following command:
+
+```bash
+flatpak run --user org.ladybird.Ladybird
+```
+
+In order to debug the flatpak, you can pass the `--devel` and `--command=sh` flags to `flatpak run`. This will drop you
+into a  shell in the flatpak sandbox:
```
gmta: Another double space here

Documentation/AdvancedBuildInstructions.md
```diff
@@ -100,6 +100,38 @@ By default, ccache will include the plugins themselves in file hashes. So if a p
...
+
+```bash
+flatpak run --user --command=sh --devel org.ladybird.Ladybird
+```
+
+
```
gmta: Some stray newlines?

Documentation/AdvancedBuildInstructions.md
ADKaster: yes, looking at the official docs it should be "Flatpak"

Documentation/AdvancedBuildInstructions.md
ADKaster: Yeup, Flathub's guidelines say capital F for this as well.

ADKaster:
> I'm fan of the [Flatpak artifacts](https://github.com/LadybirdBrowser/ladybird/actions/runs/16484738954#artifacts) we now have, but wonder if this is intentional, as there were no artifacts before. If it is intentional, then maybe they should also be documented.

It is intentional, but I'll document them later. Still feeling out how to make them into more official nightlys.

PR #5588, opened 2025-07-24
LibWeb: Don't clip descendants outside stacking context root rect

awesomekling:
Skia allows you to pass a bounding rect to its saveLayer() function as an optimization when you know that you won't paint outside those bounds. Unfortunately, we were passing a too-small rectangle that didn't take into account transformed descendants, etc.

For now, simply pass null instead of a bounding rect. This way, Skia figures it out internally. It may allocate larger temporary bitmaps than needed this way, but at least we get more correct results. I've left re-enabling the optimization as a FIXME in the code.

This fixes unwanted clipping in various parts of the Discord UI.

Before:
Screenshot 2025-07-24 at 12 41 39

After:
Screenshot 2025-07-24 at 12 41 14

### DIFF
--- a/Libraries/LibWeb/Painting/DisplayListPlayerSkia.cpp
+++ b/Libraries/LibWeb/Painting/DisplayListPlayerSkia.cpp
@@ -211,14 +211,13 @@ void DisplayListPlayerSkia::push_stacking_context(PushStackingContext const& com
     auto matrix = to_skia_matrix(new_transform);
 
     if (command.opacity < 1 || command.compositing_and_blending_operator != Gfx::CompositingAndBlendingOperator::Normal || command.isolate) {
-        auto source_paintable_rect = to_skia_rect(command.source_paintable_rect);
-        SkRect dest;
-        matrix.mapRect(&dest, source_paintable_rect);
-
         SkPaint paint;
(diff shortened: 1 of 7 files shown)


Libraries/LibWeb/Painting/DisplayListPlayerSkia.cpp
```diff
@@ -211,14 +211,13 @@ void DisplayListPlayerSkia::push_stacking_context(PushStackingContext const& com
     auto matrix = to_skia_matrix(new_transform);
 
     if (command.opacity < 1 || command.compositing_and_blending_operator != Gfx::CompositingAndBlendingOperator::Normal || command.isolate) {
-        auto source_paintable_rect = to_skia_rect(command.source_paintable_rect);
-        SkRect dest;
-        matrix.mapRect(&dest, source_paintable_rect);
```
kalenikaliaksandr, line 216: can we remove `source_paintable_rect` from PushStackingContext. seems like it's no longer used.

awesomekling: Sure!

kalenikaliaksandr (approved):

PR #5590, opened 2025-07-24
LibRegex: Improve capture group and backreference support

aplefull:
This fixes several issues with capture groups and backreferences. Previously, backreferences to undefined or non-participating capture groups weren't handled correctly, and named capture groups weren't following source order in results. Also, added support for forward references to named groups, so things like `\k<name>(?<name>x)` now work.

I was confused about how to properly deal with the source order of capture groups, since, from my understanding, the spec assumes that they are already in the correct order. But here's the note I've found in our implementation:
https://github.com/LadybirdBrowser/ladybird/blob/b1c3ce807b94d2d38e8c52296c9382373ec91bd8/Libraries/LibJS/Runtime/RegExpPrototype.cpp#L103-L117
So I had to add an extra step where we put them in the correct order. I left a note there as well. Please let me know if I'm missing something, and it can be done differently.

test262 diff:
    Summary:
        +14 ✅    -14 ❌   

Diff Tests:
    test/built-ins/RegExp/match-indices/indices-groups-properties.js                       ❌ -> ✅
    test/built-ins/RegExp/named-groups/duplicate-names-exec.js                             ❌ -> ✅
    test/built-ins/RegExp/named-groups/duplicate-names-group-property-enumeration-order.js ❌ -> ✅
    test/built-ins/RegExp/named-groups/duplicate-names-match.js                            ❌ -> ✅
    test/built-ins/RegExp/named-groups/groups-properties.js                                ❌ -> ✅
    test/built-ins/RegExp/named-groups/non-unicode-references.js                           ❌ -> ✅
    test/built-ins/RegExp/named-groups/unicode-references.js                               ❌ -> ✅
    test/built-ins/RegExp/prototype/exec/duplicate-named-groups-properties.js              ❌ -> ✅
    test/built-ins/RegExp/prototype/exec/duplicate-named-indices-groups-properties.js      ❌ -> ✅
    test/built-ins/String/prototype/match/duplicate-named-groups-properties.js             ❌ -> ✅
    test/built-ins/String/prototype/match/duplicate-named-indices-groups-properties.js     ❌ -> ✅
    test/language/literals/regexp/named-groups/forward-reference.js                        ❌ -> ✅
    test/staging/built-ins/RegExp/named-groups/duplicate-named-groups-replace.js           ❌ -> ✅
    test/staging/built-ins/RegExp/named-groups/duplicate-named-groups.js                   ❌ -> ✅

### DIFF
--- a/Libraries/LibRegex/RegexParser.cpp
+++ b/Libraries/LibRegex/RegexParser.cpp
@@ -173,6 +173,7 @@ ALWAYS_INLINE void Parser::reset()
     m_parser_state.capture_groups_count = 0;
     m_parser_state.named_capture_groups_count = 0;
     m_parser_state.named_capture_groups.clear();
+    m_parser_state.unresolved_named_references.clear();
 }
 
 Parser::Result Parser::parse(Optional<AllOptions> regex_options)
@@ -182,10 +183,13 @@ Parser::Result Parser::parse(Optional<AllOptions> regex_options)
     reset();
     if (regex_options.has_value())
         m_parser_state.regex_options = regex_options.value();
-    if (parse_internal(m_parser_state.bytecode, m_parser_state.match_length_minimum))
+    if (parse_internal(m_parser_state.bytecode, m_parser_state.match_length_minimum)) {
         consume(TokenType::Eof, Error::InvalidPattern);
-    else
+        if (!resolve_forward_named_references())
+            set_error(Error::InvalidNameForCaptureGroup);
+    } else {
         set_error(Error::InvalidPattern);
+    }
 
     auto capture_groups = m_parser_state.named_capture_groups.keys();
 
@@ -1641,23 +1645,32 @@ bool ECMA262Parser::parse_atom_escape(ByteCode& stack, size_t& match_length_mini
         }
 
         auto it = m_parser_state.named_capture_groups.find(name);
-        if (it == m_parser_state.named_capture_groups.end()) {
-            set_error(Error::InvalidNameForCaptureGroup);
-            return false;
+        if (it != m_parser_state.named_capture_groups.end()) {
+
+            // Use the first occurrence of the named group for the backreference
+            // This follows ECMAScript behavior where \k<name> refers to the first
+            // group with that name in left-to-right order, regardless of alternative
+            auto group_index = it->value.first().group_index;
+            auto maybe_length = m_parser_state.capture_group_minimum_lengths.get(group_index);
+            if (maybe_length.has_value()) {
+                // Backward reference
+                stack.insert_bytecode_compare_values({ { CharacterCompareType::NamedReference, static_cast<ByteCodeValueType>(group_index) } });
+            } else {
+                // Self-reference or forward reference
+                auto placeholder_index = 0;
+                auto bytecode_offset = stack.size();
+                stack.insert_bytecode_compare_values({ { CharacterCompareType::NamedReference, static_cast<ByteCodeValueType>(placeholder_index) } });
+
+                m_parser_state.unresolved_named_references.append({ name, bytecode_offset + 1 });
+            }
+        } else {
+            // Forward reference
+            auto placeholder_index = 0;
+            auto bytecode_offset = stack.size();
+            stack.insert_bytecode_compare_values({ { CharacterCompareType::NamedReference, static_cast<ByteCodeValueType>(placeholder_index) } });
+
+            m_parser_state.unresolved_named_references.append({ name, bytecode_offset + 1 });
         }
-
-        // Use the first occurrence of the named group for the backreference
-        // This follows ECMAScript behavior where \k<name> refers to the first
-        // group with that name in left-to-right order, regardless of alternative
-        auto group_index = it->value.first().group_index;
-
-        auto maybe_length = m_parser_state.capture_group_minimum_lengths.get(group_index);
-        if (!maybe_length.has_value()) {
-            set_error(Error::InvalidNameForCaptureGroup);
-            return false;
-        }
-
-        stack.insert_bytecode_compare_values({ { CharacterCompareType::NamedReference, (ByteCodeValueType)group_index } });
         return true;
     }
 
@@ -2815,6 +2828,22 @@ size_t ECMA262Parser::ensure_total_number_of_capturing_parenthesis()
 
     m_total_number_of_capturing_parenthesis = count;
     return count;
+}
+
+bool Parser::resolve_forward_named_references()
+{
+    for (auto const& unresolved_ref : m_parser_state.unresolved_named_references) {
+        auto it = m_parser_state.named_capture_groups.find(unresolved_ref.name);
+        if (it == m_parser_state.named_capture_groups.end()) {
+            return false;
+        }
+
+        auto group_index = it->value.first().group_index;
+
+        m_parser_state.bytecode.at(unresolved_ref.bytecode_offset) = (ByteCodeValueType)group_index;
+    }
+
+    return true;
 }
 
 }
--- a/Libraries/LibRegex/RegexParser.h
+++ b/Libraries/LibRegex/RegexParser.h
@@ -90,6 +90,7 @@ public:
 
 protected:
     virtual bool parse_internal(ByteCode&, size_t& match_length_minimum) = 0;
+    bool resolve_forward_named_references();
 
     ALWAYS_INLINE bool match(TokenType type) const;
     ALWAYS_INLINE bool match(char ch) const;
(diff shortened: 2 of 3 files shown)


Libraries/LibRegex/RegexParser.h
```diff
@@ -121,6 +121,7 @@ class REGEX_API Parser {
         AllOptions regex_options;
         HashMap<size_t, size_t> capture_group_minimum_lengths;
         HashMap<FlyString, Vector<NamedCaptureGroup>> named_capture_groups;
```
gmta: Couldn't this be an OrderedHashMap instead?

aplefull: Didn't know it exists, updated!

Libraries/LibRegex/RegexByteCode.cpp
```diff
@@ -628,6 +637,46 @@ ALWAYS_INLINE ExecutionResult OpCode_Compare::execute(MatchInput const& input, M
             }
             break;
         }
+        case CharacterCompareType::NamedReference: {
+            auto reference_number = ((size_t)m_bytecode->at(offset++)) - 1;
```
gmta, line 641: Why are you casting a `u64` to `size_t` here before subtracting one?

Libraries/LibRegex/RegexParser.cpp
```diff
@@ -187,6 +187,10 @@ Parser::Result Parser::parse(Optional<AllOptions> regex_options)
     else
         set_error(Error::InvalidPattern);
 
+    Vector<FlyString> capture_groups;
+    for (auto const& [name, _] : m_parser_state.named_capture_groups)
+        capture_groups.append(name);
```
gmta: This can be replaced with `auto capture_groups = m_parser_state.named_capture_groups.keys();`

Libraries/LibRegex/RegexParser.cpp
```diff
@@ -1643,23 +1647,32 @@ bool ECMA262Parser::parse_atom_escape(ByteCode& stack, size_t& match_length_mini
...
-        // group with that name in left-to-right order, regardless of alternative
-        auto group_index = it->value.first().group_index;
+            // Use the first occurrence of the named group for the backreference
+            // This follows ECMAScript behavior where \k<name> refers to the first
+            // group with that name in left-to-right order, regardless of alternative
+            auto group_index = it->value.first().group_index;
+            auto maybe_length = m_parser_state.capture_group_minimum_lengths.get(group_index);
+            if (maybe_length.has_value()) {
+                // Backward reference
+                stack.insert_bytecode_compare_values({ { CharacterCompareType::NamedReference, (ByteCodeValueType)group_index } });
```
gmta: Please use `static_cast<ByteCodeValueType>` instead of C-type casts

Libraries/LibJS/Runtime/RegExpPrototype.cpp
```diff
@@ -390,6 +390,21 @@ static ThrowCompletionOr<Value> regexp_builtin_exec(VM& vm, RegExpObject& regexp
     if (has_indices) {
         // a. Let indicesArray be MakeMatchIndicesIndexPairArray(S, indices, groupNames, hasGroups).
         auto indices_array = make_match_indices_index_pair_array(vm, string->utf16_string_view(), indices, group_names, has_groups);
+
+        // NOTE: We do this to add all groups to the indices array, even if they didn't participate in the match.
```
gmta: So is this a spec issue and should we raise one upstream?

Libraries/LibRegex/RegexByteCode.cpp
aplefull: That's how it was written in the case above:
https://github.com/LadybirdBrowser/ladybird/blob/cf211cf99ac5b27c733f22cc3a9ec463a718975f/Libraries/LibRegex/RegexByteCode.cpp#L611
I'm inexperienced in C++, so I don't know what is better/worse and usually I just try to follow examples of what I see in the codebase

Libraries/LibJS/Runtime/RegExpPrototype.cpp
aplefull: It's not a spec issue - what we have in RegExpPrototype diverges from the spec. We only add groups (and indices) that matched to the final list. As I understood it when reading notes in code, we are skipping them intentionally due to how LibRegex is structured. So I added this workaround. That said, LibRegex is still hard for me to comprehend, so maybe I misunderstood something and we coud do it in a better way

aplefull:
I see this PR has gone stale. I could remove the commits related to ordering if that solution is not up to standards. Or perhaps @alimpfard could take a look at how this could be done properly? Commenting to unstale for now as I don't have access to my laptop until sep 13

Libraries/LibRegex/RegexByteCode.cpp
```diff
@@ -609,12 +609,21 @@ ALWAYS_INLINE ExecutionResult OpCode_Compare::execute(MatchInput const& input, M
...
+                if (current_inversion_state())
+                    inverse_matched = true;
+                break;
+            }
 
             auto groups = state.capture_group_matches(input.match_index);
-            if (groups.size() <= reference_number)
-                return ExecutionResult::Failed_ExecuteLowPrioForks;
+
+            if (groups.size() <= reference_number) {
```
alimpfard, line 621: I was quite surprised that this _is_ the correct behaviour (as per <https://tc39.es/ecma262/#sec-backreference-matcher> step g).

Libraries/LibRegex/RegexByteCode.cpp
```diff
@@ -628,6 +637,46 @@ ALWAYS_INLINE ExecutionResult OpCode_Compare::execute(MatchInput const& input, M
...
+
+            auto str = groups.at(reference_number).view;
+
+            if (str.is_null() && groups.size() > 1) {
+                for (size_t i = 0; i < groups.size(); ++i) {
+                    if (i != reference_number && !groups[i].view.is_null()) {
+                        str = groups[i].view;
+                        break;
+                    }
+                }
```
alimpfard: I’m not sure what this means exactly? Why are we trying to compare to the first non-empty capture group that is not the one we’ve been asked to compare against?

Libraries/LibJS/Runtime/RegExpPrototype.cpp
```diff
@@ -368,6 +368,19 @@ static ThrowCompletionOr<Value> regexp_builtin_exec(VM& vm, RegExpObject& regexp
         }
     }
 
+    // NOTE: Groups are required to be in the order of their appearance in the source.
```
alimpfard: This is essentially a part of step 34 in <https://tc39.es/ecma262/#sec-regexpbuiltinexec>, not exactly worded this way but effectively similar.
Since this is in LibJS, let’s follow the spec directly :)

Libraries/LibRegex/RegexByteCode.cpp
aplefull: It's been quite a while, so I may not remember all the problems I encountered. But I think the main issue I was trying to fix is cases like this: `"bb".match(/(?:(?<x>a)|(?<x>b))\k<x>/`. 
1. First alternative `(?<x>a)` fails -> group 0 is null
2. Second alternative `(?<x>b)` matches -> group 1 captures "b"
3. `\k<x>` should reference group that actually participated in the match, not just the first one.

So without this, we would get `["b",null,"b"]` instead of `["bb",null,"b"]`

I see now that my impl is definitely not correct since it searches any non-null group, not specifically groups with the same name, but if I understand correctly, we should find first group with the same duplicate name that participated in the match. Not sure how to go about that though, since I don't think we have access to group names in bytecode rn?

alimpfard: My impression of the whole problem space was that groups with the same name should all get the same capture index, why is that not possible/sufficient?

aplefull: I didn't think about it, that would be much cleaner, but I think each `(?<name>...)` group is supposed to get its own slot in the results array? Like in the example above, we should get `["bb",null,"b"]`. Without separate indices we'd lose this distinction and get ["bb", "b"] instead. Similar thing for `indices` field when we have `d` flag enabled

alimpfard: Hmmmm, that's annoying yeah
Having aliases would certainly work but maybe what you're doing is better (sans the "any nonnull capture" bit :)

Libraries/LibJS/Runtime/RegExpPrototype.cpp
aplefull: Oh no, forgot to actually submit the comment. I changed RegExpBuiltinExec to follow spec steps, but I still had to add a part that puts groups into the source order. I guess spec just expects them to be in the right order, without mentioning it as an explicit step?

Libraries/LibRegex/RegexByteCode.cpp
aplefull: Changed it to search only groups with the same name now

alimpfard:
Sorry it took me a while to get back to this, I totally forgot about it...

PR #5595, opened 2025-07-24
LibGfx: VulkanContext coverity reports integer_overflow on index

rcorsi:
Coverity static analysis reports that the code that scans the queue families for one that has the graphics bit, can be -1 if none are found, which could cause a problem when the -1 (signed) value is used later as an index in a uint32_t (unsigned) variable.

Its not immediately clear how often this could occur, not finding a queue family with the graphics bit, but adding some protecting just in case.

Screenshot From 2025-07-22 17-08-12

### DIFF
--- a/Libraries/LibGfx/VulkanContext.cpp
+++ b/Libraries/LibGfx/VulkanContext.cpp
@@ -75,12 +75,17 @@ static ErrorOr<VkDevice> create_logical_device(VkPhysicalDevice physical_device)
     queue_families.resize(queue_family_count);
     vkGetPhysicalDeviceQueueFamilyProperties(physical_device, &queue_family_count, queue_families.data());
 
-    int graphics_queue_family_index = -1;
-    for (int i = 0; i < static_cast<int>(queue_families.size()); i++) {
-        if (queue_families[i].queueFlags & VK_QUEUE_GRAPHICS_BIT) {
-            graphics_queue_family_index = i;
+    bool graphics_queue_family_found = false;
+    uint32_t graphics_queue_family_index = 0;
+    for (; graphics_queue_family_index < queue_families.size(); ++graphics_queue_family_index) {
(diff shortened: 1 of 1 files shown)


Libraries/LibGfx/VulkanContext.cpp
```diff
@@ -75,14 +75,19 @@ static ErrorOr<VkDevice> create_logical_device(VkPhysicalDevice physical_device)
     queue_families.resize(queue_family_count);
     vkGetPhysicalDeviceQueueFamilyProperties(physical_device, &queue_family_count, queue_families.data());
 
-    int graphics_queue_family_index = -1;
-    for (int i = 0; i < static_cast<int>(queue_families.size()); i++) {
+    bool graphics_queue_family_found = false;
+    uint32_t graphics_queue_family_index = 0;
+    for (auto& i = graphics_queue_family_index; i < queue_families.size(); ++i) {
```
kalenikaliaksandr: it doesn't seem like `auto&` is intentional here. did you mean to use just `auto`?

kalenikaliaksandr (changes requested):

Libraries/LibGfx/VulkanContext.cpp
rcorsi: Definitely the auto& is intentional.  Variables 'i' and 'graphics_queue_family_index' need to be the same value, so they represent the same stack variable.  I want to use the name 'i' because it is shorter for the loop, but I want the value available later on line 93 with the longer named 'graphics_queue_family_index'.

If this is not the clearest, and simplest, well I would agree. I could resubmit using only one variable name if this is a problem.

kalenikaliaksandr: we could make this loop increment `graphics_queue_family_index` and remove `i` completely.

rcorsi: Sounds good. I'll do that.

Issue #5486, opened 2025-07-17
Scrolling updates broken/delayed on w3schools.com


gmta:
Bisected to 6507d23e2908156fc53104c1fbeb4609cd38db71 (CC @kalenikaliaksandr)

The right scrollable frame on this page does not update properly when using the mousewheel:

https://www.w3schools.com/cssref/tryit.php?filename=trycss_list-style-type_all

If you click the frame, it is updated immediately.

PR #5596, opened 2025-07-25
LibWeb: Fix scroll state refresh in cached display list for iframes

kalenikaliaksandr:
6507d23 introduced a bug when snapshot for iframe is saved in `PaintNestedDisplayList` and, since display lists are immutable, it's not possible to update before the next repaint.

This change fixes the issue by moving `ScrollStateSnapshot` for nested display lists from `PaintNestedDisplayList` to `HashMap<NonnullRefPtr<DisplayList>, ScrollStateSnapshot>` that is placed into pending rendering task, making it possible to update snapshots for all display lists before the next repaint.

This change doesn't have a test because it's really hard to make a ref test that will specifically check scenario when scroll offset of an iframe is advanced after display list is cached. We already have `Tests/LibWeb/Ref/input/scroll-iframe.html` but unfortunately it did not catch this bug.

Fixes https://github.com/LadybirdBrowser/ladybird/issues/5486

### DIFF
--- a/Libraries/LibWeb/CSS/StyleValues/CursorStyleValue.cpp
+++ b/Libraries/LibWeb/CSS/StyleValues/CursorStyleValue.cpp
@@ -94,8 +94,7 @@ Optional<Gfx::ImageCursor> CursorStyleValue::make_image_cursor(Layout::NodeWithS
         case DisplayListPlayerType::SkiaCPU: {
             auto painting_surface = Gfx::PaintingSurface::wrap_bitmap(bitmap);
             Painting::DisplayListPlayerSkia display_list_player;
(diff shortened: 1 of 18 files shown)


dotnetCarpenter:
I wonder if [`requestIdleCallback` with a low timeout](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback#options) or an old school `setTimeout` would be feasible for a test scenario, that you talk about?

Edit: it probably make for a racy/finicky test… _Feel free to ignore_

kalenikaliaksandr:
I am going to merge this fix as is. maybe in the future we could redo it in a simpler way like @gmta described here https://discord.com/channels/1247070541085671459/1247090064480014443/1398227828755009566

PR #5600, opened 2025-07-25
AK: Add and use HashMap::update()

gmta:


Lubrsi:
I was looking for such an API when working on https://github.com/LadybirdBrowser/ladybird/pull/5565, thanks for adding it :^)

alimpfard:
🚲shed: "extend" implies (to me) that only new keys are added, maybe "update"?

gmta:
> 🚲shed: "extend" implies (to me) that only new keys are added, maybe "update"?

~~merge? Because "update" implies to me that values might be overwritten.~~
wait, they are 🤔

alimpfard:
"merge" is okay too.
Yeah the overwriting thing is the main reason i thought of update.

Hendiadyoin1:
~~I second Alis comment, this~~ `update` is also the wording python uses
(STL has `merge` but that preserves old values afaict, not sure if you can jinx their `insert` to do an `update`)

PR #5606, opened 2025-07-25
LibWeb+LibWebView+WebContent+UI: Set the paint interval to the refresh rate of the current screen

Lubrsi:
This allows us to paint web content at, for example, 120 FPS. This affects all painting, whether it's scrolling or driving a WebGL application with requestAnimationFrame.

https://github.com/user-attachments/assets/7a44cada-17ad-4974-a223-4ee1fc72078d

### DIFF
--- a/Services/WebContent/PageClient.h
+++ b/Services/WebContent/PageClient.h
@@ -55,6 +55,7 @@ public:
     void set_viewport_size(Web::DevicePixelSize const&);
     void set_screen_rects(Vector<Web::DevicePixelRect, 4> const& rects, size_t main_screen_index) { m_screen_rect = rects[main_screen_index]; }
     void set_device_pixels_per_css_pixel(float device_pixels_per_css_pixel) { m_device_pixels_per_css_pixel = device_pixels_per_css_pixel; }
+    void set_maximum_frames_per_second(u64 maximum_frames_per_second);
     void set_preferred_color_scheme(Web::CSS::PreferredColorScheme);
     void set_preferred_contrast(Web::CSS::PreferredContrast);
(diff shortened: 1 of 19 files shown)


Services/WebContent/PageClient.h
```diff
@@ -186,6 +187,7 @@ class PageClient final : public Web::PageClient {
     RefPtr<Gfx::PaletteImpl> m_palette_impl;
     Web::DevicePixelRect m_screen_rect;
     float m_device_pixels_per_css_pixel { 1.0f };
+    double m_maximum_frames_per_second { 60.0 };
```
kalenikaliaksandr, line 190: shouldn't we use uint for fps?

kalenikaliaksandr (changes requested):

Services/WebContent/PageClient.h
Lubrsi: I originally had it as a uint, however Qt returns a double and it really can be decimal, e.g. my 144Hz monitor reports 143.85Hz, 60Hz as 59.95Hz, etc. Though we likely don't need the precision, though I wonder what rounding we should use (e.g. if it was to report 60.01Hz, a round up would go to 61Hz, when we just wanted 60, so on so forth)

Psychpsyo: I think we would always want to round up.
If we round down, we're drawing less frames in total than what should appear and this would lead to us skipping a frame every now and then.
(rounding up leads to us rendering twice for a frame every now and then, but that's probably not an issue)

kalenikaliaksandr: got it, I think in that case using double is ok for now.

PR #5612, opened 2025-07-26
LibWasm: Return canonical NaN for min/max binops

gmta:
Instead of returning whichever argument was NaN, return the canonical NaN instead.

Fixes 448 WPT subtests in `wasm/core`.


nico:
Is there a bug for WPT and spec disagreeing?

tcl3:
Which WPT tests does this improve for you? I'm not seeing any difference when comparing all the `wasm` tests on both Mac and Linux.

alimpfard:
The wasm/core tests are imported from the wasm testsuite anyway, this sort of behaviour issue would show up in the base test suite as well.
I'd only look to the LibWeb/WebAssembly layer for any discrepancies in this particular case.

gmta:
@nico 
> Is there a bug for WPT and spec disagreeing?

I don't think so, and I've waited with raising an issue myself since I kinda expected _some_ discussion on this PR.

@tcl3 
> Which WPT tests does this improve for you? I'm not seeing any difference when comparing all the `wasm` tests on both Mac and Linux.

Hm, they don't show up using `Meta/WPT.sh compare ...` 🤔 they do show up in the subtest failures though:

`wasm/core/simd` run on master:
```
Ran 51532 checks (51474 subtests, 58 tests)
Expected results: 51372
Unexpected results: 160
  subtest: 160 (160 fail)
```

`wasm/core/simd` run on this branch:
```
Ran 51532 checks (51474 subtests, 58 tests)
Expected results: 51532
Unexpected results: 0
```

This comes from the `--log-mach` output by the way.

@alimpfard 
> The wasm/core tests are imported from the wasm testsuite anyway, this sort of behaviour issue would show up in the base test suite as well. I'd only look to the LibWeb/WebAssembly layer for any discrepancies in this particular case.

~~Do we run that testsuite as part of ctest / CI?~~ The failing tests look at the result of executing the wasm function, which lead to an unreachable instruction. There's no arguments to the function being run.

gmta:
@tcl3 @alimpfard for easier comparison, I've added one of the tests that was previously failing. If you revert my changes to `Operators.h`, you'll see which ones start to complain.

Tested on a Macbook M4 Pro, if you don't see the same failures pop up we might have something of an architecture discrepancy going on.

tcl3:
Just to expand on why `./Meta/WPT.sh compare` doesn't pick these differences up - it relies on the name of each subtest name being consistent across runs. With these tests the correctness of the test can affect the subtest names. The effect of this can be seen by looking at one of the tests on wpt.fyi: https://wpt.fyi/results/wasm/core/simd/simd_f32x4.wast.js.html?label=master&product=ladybird%5Bexperimental%5D&product=chrome

alimpfard:
Here's one of the failing tests in readable form (it seems to be generated from the spectest, it's not exactly the same):
```wast
(function
  (locals
  )
  (body
    (block (structured
      (type block empty)
      (else (none)) (end 12)))
    (v128.const ffa00000ffa00000'ffa00000ffa00000)
    (v128.const ffa00000ffa00000'ffa00000ffa00000)
    (call (function index 0)) ;; f32x4.max
    (v128.const 7fc000007fc00000'7fc000007fc00000)
    (v128.and)
    (v128.const 7fc000007fc00000'7fc000007fc00000)
    (i8x16.eq)
    (i8x16.all_true)
    (i32.eqz)
    (br.if (label index 0))
    (return)
    (synthetic:end)
    (unreachable)
  )
)
```
Equivalent to:
```wast
(assert_return (invoke "f32x4.max" (v128.const f32x4 -nan:0x200000 -nan:0x200000 -nan:0x200000 -nan:0x200000)
                                   (v128.const f32x4 -nan:0x200000 -nan:0x200000 -nan:0x200000 -nan:0x200000))
                                   (v128.const f32x4 nan:arithmetic nan:arithmetic nan:arithmetic nan:arithmetic))
```

Now `nan:arithmetic` is defined as "NaN with 1 in MSB of payload" (https://github.com/WebAssembly/spec/blob/main/interpreter/README.md#s-expression-syntax), so WPT is right.

I'm happy with your impl, but please include this:
```diff
diff --git a/Tests/LibWasm/test-wasm.cpp b/Tests/LibWasm/test-wasm.cpp
index 9fec682b05f..52e5637e523 100644
--- a/Tests/LibWasm/test-wasm.cpp
+++ b/Tests/LibWasm/test-wasm.cpp
@@ -224,14 +224,16 @@ TESTJS_GLOBAL_FUNCTION(is_canonical_nan64, isCanonicalNaN64)

 TESTJS_GLOBAL_FUNCTION(is_arithmetic_nan32, isArithmeticNaN32)
 {
-    auto value = bit_cast<float>(TRY(vm.argument(0).to_u32(vm)));
-    return isnan(value);
+    auto const bits = TRY(vm.argument(0).to_u32(vm));
+    auto const payload = bits & 0x007FFFFF;
+    return (bits & 0x7F800000) == 0x7F800000 && payload >= 0x00400001;
 }

 TESTJS_GLOBAL_FUNCTION(is_arithmetic_nan64, isArithmeticNaN64)
 {
-    auto value = bit_cast<double>(TRY(vm.argument(0).to_bigint_uint64(vm)));
-    return isnan(value);
+    auto const bits = TRY(vm.argument(0).to_bigint_uint64(vm));
+    auto const payload = bits & 0x000FFFFFFFFFFFFFULL;
+    return (bits & 0x7FF0000000000000ull) == 0x7FF0000000000000ull && payload >= 0x0008000000000001ull;
 }

 TESTJS_GLOBAL_FUNCTION(test_simd_vector, testSIMDVector)
```

gmta:
@alimpfard done!

gmta:
@alimpfard Sanitizer build tests disagree with your new definitions :(

alimpfard:
Wait what? How does changing test-wasm break bigint tests?

gmta:
> Wait what? How does changing test-wasm break bigint tests?

That's a fluke, take a look at the macOS results

alimpfard:
Re the other CI failure, I may have written the check wrong? Will double-check in a bit.

alimpfard:
ah sorry, `0x400000`/`0x8000000000000` should also be valid:
```
diff --git a/Tests/LibWasm/test-wasm.cpp b/Tests/LibWasm/test-wasm.cpp
index 9fec682b05f..52e5637e523 100644
--- a/Tests/LibWasm/test-wasm.cpp
+++ b/Tests/LibWasm/test-wasm.cpp
@@ -224,14 +224,16 @@ TESTJS_GLOBAL_FUNCTION(is_canonical_nan64, isCanonicalNaN64)

 TESTJS_GLOBAL_FUNCTION(is_arithmetic_nan32, isArithmeticNaN32)
 {
-    auto value = bit_cast<float>(TRY(vm.argument(0).to_u32(vm)));
-    return isnan(value);
+    auto const bits = TRY(vm.argument(0).to_u32(vm));
+    auto const payload = bits & 0x007FFFFF;
+    return (bits & 0x7F800000) == 0x7F800000 && payload >= 0x400000;
 }

 TESTJS_GLOBAL_FUNCTION(is_arithmetic_nan64, isArithmeticNaN64)
 {
-    auto value = bit_cast<double>(TRY(vm.argument(0).to_bigint_uint64(vm)));
-    return isnan(value);
+    auto const bits = TRY(vm.argument(0).to_bigint_uint64(vm));
+    auto const payload = bits & 0x000FFFFFFFFFFFFFull;
+    return (bits & 0x7FF0000000000000ull) == 0x7FF0000000000000ull && payload >= 0x0008000000000000ull;
 }

 TESTJS_GLOBAL_FUNCTION(test_simd_vector, testSIMDVector)
```

gmta:
Thanks, updated. I've also removed the WPT test case since it was huge & relatively slow.

PR #5618, opened 2025-07-28
Invalidate less elements affected by CSS custom properties

kalenikaliaksandr:
Before this change, whenever element's attributes changed, we would add
a flag to "pending invalidation", indicating that all descendants whose
style uses CSS custom properties needed to be recomputed. This resulted
in severe overinvalidation, because we would run invalidation regardless
of whether any custom property on affected element actually changed.

This change takes another approach, and now we decide whether
descendant's style needs to be recomputed based on whether ancestor's
style recomputation results in a change of custom properties, though
this approach adds a little overhead to style computation as now we have
to compare old vs new hashmap of custom properties.

This brings substantial improvement on discord and x.com where, before
this change, advantage of using invalidation sets was lost and we had
to recompute all descendants, because they almost all use some custom
properties.


AK/HashMap.h
```diff
@@ -322,6 +322,27 @@ class HashMap {
         return hash_map_clone;
     }
 
+    bool operator==(HashMap const& other) const
```
tcl3, line 325: Can we add a basic test for this.

Libraries/LibWeb/DOM/Element.h
```diff
@@ -201,7 +201,7 @@ class Element
 
     void run_attribute_change_steps(FlyString const& local_name, Optional<String> const& old_value, Optional<String> const& value, Optional<FlyString> const& namespace_);
 
-    CSS::RequiredInvalidationAfterStyleChange recompute_style();
+    CSS::RequiredInvalidationAfterStyleChange recompute_style(bool& did_change_custom_properties);
```
awesomekling, line 204: Would it make sense for `did_change_custom_properties` to be a member of `RequiredInvalidationAfterStyleChange` instead?

kalenikaliaksandr: `recompute_style()` is only used from `update_style_recursively()`, which is the only place where it's known how to process `did_change_custom_properties`. Other places that act based on flags in `RequiredInvalidationAfterStyleChange` won't be able to process `did_change_custom_properties`, so I don't see value in putting it there.

PR #5623, opened 2025-07-28
LibRegex: Account for opcode size when calculating incoming jump edges

Calme1709:
Not accounting for opcode size when calculating incoming jump edges meant that we were merging nodes where we otherwise shouldn't have been, for example `/.*a|.*b/`


Calme1709:
Whoops - looks like repeat is different from the others in that the offset is relative to instruction itself rather than the next instruction. Have rolled the change for this as the original implementation was correct.

alimpfard (approved):
Thanks!

PR #5628, opened 2025-07-28
LibWeb+UI: Add an explicit IPC to handle mouse leave events

trflynn89:
The faux position we created here is adjusted by the device pixel ratio later on, which would invoke integer overflow on screens with a DPR greater than 1.

Instead of creating special data for a mouse move event, let's just add an explicit leave event handler.


trflynn89:
In practice, this actually behaves a bit better than before as well. If a button was partially in view / partially clipped by the widget window, it would previously remain hovered on leave events.

PR #5629, opened 2025-07-28
LibWeb/CSS: Support interpolating `filter` and `backdrop-filter` values

tcl3:
This PR brings interpolation of `filter` and `backdrop-filter` values closer to spec.

`drop-shadow()` values currently fall back to using discrete interpolation.

Gets us +687/-116 WPT subtest passes in the `css/filter-effects` directory.


tcl3:
My latest push ensures each filter type uses the correct default value for interpolation when replacing the value `none`.

This gets us to +687/-116 WPT subtest passes.

gmta (approved):

PR #5633, opened 2025-07-28
LibWeb: Port the document title to UTF-16

trflynn89:
Mostly just to get the IPC hooks out of the way.

### DIFF
--- a/AK/Utf16StringData.h
+++ b/AK/Utf16StringData.h
@@ -34,6 +34,7 @@ public:
     static NonnullRefPtr<Utf16StringData> from_utf16(Utf16View const&);
     static NonnullRefPtr<Utf16StringData> from_utf32(Utf32View const&);
     static NonnullRefPtr<Utf16StringData> from_string_builder(StringBuilder&);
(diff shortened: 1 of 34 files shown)


AK/Utf16StringData.h
```diff
@@ -33,6 +33,7 @@ class Utf16StringData final : public RefCounted<Utf16StringData> {
     static NonnullRefPtr<Utf16StringData> from_utf8(StringView, AllowASCIIStorage);
     static NonnullRefPtr<Utf16StringData> from_utf16(Utf16View const&);
     static NonnullRefPtr<Utf16StringData> from_utf32(Utf32View const&);
+    static ErrorOr<NonnullRefPtr<Utf16StringData>> from_stream(Stream&, size_t length_in_code_units, bool is_ascii);
```
gmta: This API's `bool is_ascii` leaks an implementation detail; maybe introduce a `Badge<>` to limit its usage to IPC decoding?

trflynn89: Yeah I didn't love this either - we can't badge it though, because the `decode` overloads are free functions. What I've done for now is name the function to indicate it's specifically an IPC stream reader.

PR #5634, opened 2025-07-29
LibCrypto: Implement Ed25519 Small Order Points

ChaseKnowlden:
Fixes All Tests in the WebCryptoAPI/sign_verify directory.


Libraries/LibCrypto/Curves/EdwardsCurve.cpp
```diff
@@ -65,8 +66,67 @@ ErrorOr<ByteBuffer> SignatureEdwardsCurve::sign(ReadonlyBytes private_key, Reado
...
+        { 0xc7, 0x17, 0x6a, 0x70, 0x3d, 0x4d, 0xd8, 0x4f, 0xba, 0x3c, 0x0b, 0x76, 0x0d, 0x10, 0x67, 0x0f,
+            0x2a, 0x20, 0x53, 0xfa, 0x2c, 0x39, 0xcc, 0xc6, 0x4e, 0xc7, 0xfd, 0x77, 0x92, 0xac, 0x03, 0xfa }, // pubKeys[0]
+        { 0xf7, 0xba, 0xde, 0xc5, 0xb8, 0xab, 0xea, 0xf6, 0x99, 0x58, 0x39, 0x92, 0x21, 0x9b, 0x7b, 0x22,
+            0x3f, 0x1d, 0xf3, 0xfb, 0xbe, 0xa9, 0x19, 0x84, 0x4e, 0x3f, 0x7c, 0x55, 0x4a, 0x43, 0xdd, 0x43 }, // pubKeys[1] (actually part of a signature, but used as small-order)
+        // pubKeys[5] - same as pubKeys[4] but with different case, still the same small-order point
+        { 0xEC, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
+            0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0x7F } } };
+
+    for (auto const& small_order_point : small_order_points) {
+        if (memcmp(public_key.data(), small_order_point.data(), 32) == 0)
```
gmta: This should be `if (public_key == small_order_point)`

Libraries/LibCrypto/Curves/EdwardsCurve.cpp
```diff
@@ -65,8 +66,67 @@ ErrorOr<ByteBuffer> SignatureEdwardsCurve::sign(ReadonlyBytes private_key, Reado
...
+    }
+
+    return false;
+}
+
 ErrorOr<bool> SignatureEdwardsCurve::verify(ReadonlyBytes public_key, ReadonlyBytes signature, ReadonlyBytes message, ReadonlyBytes context)
 {
+    // For Ed25519, reject small-order points for security
+    // This is required by RFC 8032 and the Web Crypto API specification
+    if (strcmp(m_curve_name, "ED25519") == 0) {
```
gmta: Why is `m_curve_name` a `char const*` instead of an enum?

ChaseKnowlden: It's used for several good reasons.

gmta: It only seems to be relevant in EdwardsCurve.cpp, I'd opt for an enum and a static method returning a `char const*` whenever we need to have a value for OpenSSL. That'd make the above code a bit easier on the eyes.

Libraries/LibCrypto/Curves/EdwardsCurve.cpp
ChaseKnowlden: Fixed.

Libraries/LibCrypto/Curves/EdwardsCurve.cpp
ChaseKnowlden: Fixed now.

PR #5641, opened 2025-07-29
[[WIP]] LibJS: Use call-threading instead of direct jump threading

Hendiadyoin1:
This allows us to have different instruction impls hidden in their own files, and feels a bit cleaner.

---

Mainly a POC,
currently this somehow causes a segfault somewhere, will investigate that later (or when someone finds it before me even better)

Will also need to move some stuff around to get the same inlining behavior as before, the ` FLATTEN_ON_CLANG` might be superfluous with this as the functions are way smaller now

### DIFF
--- a/Libraries/LibJS/Bytecode/Instructions/Comparison.cpp
+++ b/Libraries/LibJS/Bytecode/Instructions/Comparison.cpp
@@ -76,7 +76,7 @@ ALWAYS_INLINE FLATTEN void Interpreter::handle_comparison(u8 const* bytecode, si
 #define HANDLE_COMPARISON_OP(op_TitleCase, op_snake_case, numeric_operator)                  \
     void Interpreter::handle_Jump##op_TitleCase(u8 const* bytecode, size_t& program_counter) \
     {                                                                                        \
-        return handle_comparison<Op::Jump##op_TitleCase, op_snake_case,                      \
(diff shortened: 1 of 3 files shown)


Libraries/LibJS/Bytecode/Instructions.h
```diff
@@ -0,0 +1,73 @@
...
+            state.program_counter += sizeof(Op::name);     \
+    } while (0)
+
+#if defined(AK_COMPILER_CLANG)
+#    define DISPATCH_NEXT(state)                                                                                        \
+        do {                                                                                                            \
+            auto& next_instruction = *reinterpret_cast<Instruction const*>(&state.bytecode[state.program_counter]);     \
+            [[clang::musttail]] return dispatch_instruction_table[static_cast<size_t>(next_instruction.type())](state); \
+        } while (0)
+#else // GCC does not support musttail, so let's hack it
```
Hendiadyoin1: Not sure if you can have `#if` in defines, so we get this for now

awesomekling:
What's the performance impact?

Hendiadyoin1:
I will tell you once I find whats causing the segfault...

Libraries/LibJS/Bytecode/Instructions.h
circl-lastname: This is the correct way of making conditional macros

Issue #4782, opened 2025-05-16
Use the special `list-item` counter to generate markers on list items


manuel-za:
Before: List item markers are implemented using [HTML ordinal value](https://html.spec.whatwg.org/multipage/grouping-content.html#ordinal-value).

[CSS implicit list-item counter](https://drafts.csswg.org/css-lists-3/#list-item-counter) supersedes it.
- Based on [CSS counters](https://drafts.csswg.org/css-lists-3/#auto-numbering) [mostly] implemented in https://github.com/LadybirdBrowser/ladybird/pull/720.
   - Does not support [reverse counters](https://github.com/LadybirdBrowser/ladybird/pull/720#issuecomment-2248243241) yet.
- Simplifies calculation through a style cascade (vs. intercept of `ordinal_value` calls).
- Supports [tiered numbering](https://drafts.csswg.org/css-lists-3/#example-50e7c443) of list elements using the [CSS content property](https://drafts.csswg.org/css-content-3/#content-property). Nice!
- Firefox and Chrome already implement it.
- _"we shouldn't need a separate system for calculating list-item ordinals - we should be able to just ask for the value of the list-item counter"_ (from [here](https://github.com/LadybirdBrowser/ladybird/pull/4442#issuecomment-2844592662))

PR #5016, opened 2025-06-05
LibWeb/CSS: Add support for reversed counters

manuel-za:
Fixes #4867 :
- [Reversed counters](https://drafts.csswg.org/css-lists-3/#css-counter-reversed) are in the spec.
- Firefox supports them, but not Chrome.
- Implementing this :point_down:  would be a lot cleaner with reversed counters: 
        - https://github.com/LadybirdBrowser/ladybird/issues/4782

<details><summary>Details</summary>

<br/>

Already, counters are read as a CSS style, then a special `CountersSet` structure is created and maintained while building the layout tree. Very nice! :sunglasses: 

<br/>

What's tricky:

- The initial value of a `reversed` counter depends on subsequent occurrences of `counter-increment`  and cannot be calculated when building the layout tree in one pass (all occurrences must be seen).
- Counting the initial value of a `reversed` counter stops when encountering the first `counter-set` (only count down to the first set value). See example.
- Each counter points to an `originator_element`, where the counter was encountered for the first time. In the case of `reversed` counters, that must be an element with a `counter-reset` style.
- Normally, `reversed` counters don't have initial values + the spec is ambiguous about [how to treat initial values for reversed counters](https://github.com/w3c/csswg-drafts/issues/6231).
     - Just like Firefox, we ignore `reversed` if an initial value is specified. 

```html
<style>
    div > div::before { content: counter(item) ": "; }
    .Rev { counter-reset: reversed(item); }
    .Dec { counter-increment: item -1; }
    .Set8 { counter-set: item 8; }
    .Set4 { counter-set: item 4; }
</style>

<div id="a" class='Rev'>
    <div id="b" class='Dec'>Ten</div>
    <div id="c" class='Dec'>Nine</div>
    <div id="d" class='Set8' title='First `counter-set` gives initial value. `reversed` is meaningless beyond this point.'>Eight</div>
    <div id="e" class='Dec'>Seven</div>
    <div id="f" class='Set4' title='Second `counter-set` totally behaves as if the counter was not reversed.'>Four</div>
    <div id="f" class='Dec'>Three</div>
</div>
```

Solution:
- Progressively calculate the initial value of reversed counters as `counter-increment` and `counter-set` styles are found during the first pass of building the layout tree
     - [Algorithm for instantiating reversed counters without a value](https://drafts.csswg.org/css-lists-3/#instantiating-counters).
- Add to the value in each of the counted elements the "initial" value of the counter (from the `originator_element`).
[...]

PR #5645, opened 2025-07-29
LibWeb/CSS: Add support for content to the ::marker pseudo-element

### LINKED ISSUE #5019: Add support for `content` to the `::marker` pseudo-element
The [spec](https://drafts.csswg.org/css-lists-3/#content-property) says:

>The contents of the [marker box](https://drafts.csswg.org/css-lists-3/#marker) are determined as defined by the [content](https://drafts.csswg.org/css-content-3/#propdef-content) property

...but we don't support that yet.

```html
<style>
div { display: list-item; list-style-position: inside;}
div::marker { content: counter(list-item) ': ';}
</style>
<div>One</div>
```

Firefox:
Image

Ladybird:
Image

manuel-za:
Fixes #5019 

FIXME: Markers for `list-item` elements with `content: normal` are still maintained using `ordinal` and `ListMarkerBox`.
This should disappear after #4782.


Libraries/LibWeb/Layout/BlockFormattingContext.cpp
```diff
@@ -769,7 +771,10 @@ void BlockFormattingContext::layout_block_level_box(Box const& box, BlockContain
     // If we do not do this then left-floating elements inside the list item will push the marker to the right,
     // in some cases even causing it to overlap with the non-floating content of the list.
     CSSPixels left_space_before_children_formatted;
-    if (is<ListItemBox>(box)) {
+    if (is<ListItemBox>(box) && (!(box.dom_node() && box.dom_node()->is_element() && static_cast<DOM::Element const*>(box.dom_node())->computed_properties(CSS::PseudoElement::Marker)->property(CSS::PropertyID::Content).is_content()))) {
```
konradekk: wouldn't it be better to use a name instead? (a variable or inlined method; it seems it is used in two, almost three, places already?)

Libraries/LibWeb/Layout/TreeBuilder.cpp
```diff
@@ -277,14 +283,41 @@ void TreeBuilder::create_pseudo_element_if_needed(DOM::Element& element, CSS::Ps
...
+
+    // For ::marker with content 'normal', create the marker pseudo-element from a ListItemMarkerBox
+    // FIXME: This + ListItemBox + ListItemMarkerBox will disappear once ::marker pseudo-elements with 'normal' content
+    //        are rendered using the special list-item counter.
+    //        See: https://github.com/LadybirdBrowser/ladybird/issues/4782
+    if (element.layout_node() && is<ListItemBox>(*element.layout_node()) && pseudo_element == CSS::PseudoElement::Marker && pseudo_element_content.type == CSS::ContentData::Type::Normal) {
+        auto& style_computer = document.style_computer();
+        auto layout_node = element.layout_node();
+
+        auto marker_style = style_computer.compute_style(element, CSS::PseudoElement::Marker);
```
AtkinsSJ: Do we need to compute this here (and call `element.set_computed_properties()` below), or can we just use `pseudo_element_style`? I think it should be equivalent, but I'm not 100% sure.

Tests/LibWeb/Layout/input/list-item-marker-pseudo-placement.html
```diff
@@ -4,14 +4,28 @@
...
+    "A block container either [...] or contains only block-level boxes [...] possibly generating anonymous block boxes to ensure this constraint [...]"
+   In our case, the contents of the ListItemBox are wrapped inside an anonymous block container (ABC). Nice.
+
+ - The CSS Lists spec (https://drafts.csswg.org/css-lists-3/#marker-pseudo) says:
+    "The marker box is generated [...] as the list item’s first child, before the ::before pseudo-element."
+    In our case, the ::before pseudo-element is a child of ABC.
+
+ - The most reasonable interpretation seems to be that the marker box is a "sibling of the ::before pseudo-element"
+   and NOT "a sibling of ABC".
+-->
```
AtkinsSJ: I'm not sure how this relates to this particular test. Is it observable? If so, have you checked how other browsers behave?

manuel-za: Indeed, other browsers interpret as "a sibling of ABC". I should have thought of that :thinking: . There's probably a lesson here somewhere...
Moving the issue I was trying to deal with (embedded lists) in a FIXME.

Libraries/LibWeb/Layout/BlockFormattingContext.cpp
```diff
@@ -816,23 +818,28 @@ void BlockFormattingContext::layout_block_level_box(Box const& box, BlockContain
         resolve_used_height_if_treated_as_auto(box, available_space_for_height_resolution);
     }
 
+    // This monster basically means: "a ListItemBox that does not have specified content in the ::marker pseudo-element".
+    // This happens for ::marker with content 'normal'.
+    bool is_list_item_box_without_css_content = is<ListItemBox>(box) && (!(box.dom_node() && box.dom_node()->is_element() && static_cast<DOM::Element const*>(box.dom_node())->computed_properties(CSS::PseudoElement::Marker)->property(CSS::PropertyID::Content).is_content()));
+
```
konradekk, line 814: it seems you’re using `is<ListItemBox>(box)` below in form of `auto const* li_box = as_if<ListItemBox>(box)`. maybe it’d be better to start with that here and use below, and define `is_list_item_box_without_css_content` in terms of `li_box`? (i guess that you can also assume, that `dom_node()` return value exists, all these would probably significantly simplify this “monster” as you call it.)

Libraries/LibWeb/Layout/TreeBuilder.cpp
```diff
@@ -277,14 +283,39 @@ void TreeBuilder::create_pseudo_element_if_needed(DOM::Element& element, CSS::Ps
...
+    if (pseudo_element == CSS::PseudoElement::Marker
+        && (pseudo_element_display.is_none() || pseudo_element_content.type == CSS::ContentData::Type::None))
+        return;
+
+    // For ::marker with content 'normal', create the marker pseudo-element from a ListItemMarkerBox
+    // FIXME: This + ListItemBox + ListItemMarkerBox will disappear once ::marker pseudo-elements with 'normal' content
+    //        are rendered using the special list-item counter.
+    //        See: https://github.com/LadybirdBrowser/ladybird/issues/4782
+    if (element.layout_node() && is<ListItemBox>(*element.layout_node()) && pseudo_element == CSS::PseudoElement::Marker && pseudo_element_content.type == CSS::ContentData::Type::Normal) {
+        auto layout_node = element.layout_node();
```
konradekk: it seems you could also benefit from using `as_if` on `element.layout_node()` directly to get relevant `layout_node` used inside (esp. that you later also cast this into `ListItemBox`).

konradekk (commented):
spotted few places that could be expressed more clearly or succinctly using existing infra.

Libraries/LibWeb/Layout/BlockFormattingContext.cpp
manuel-za: Great suggestions. Thanks!

I don't know that I can assume that the `dom_node()` value exists... Are list item boxes guaranteed not be anonymous? :thinking: ... Could not find anything in the spec... I left the test in. Could you point me to specs + I'll change it.

Libraries/LibWeb/Layout/BlockFormattingContext.cpp
```diff
@@ -805,11 +807,16 @@ void BlockFormattingContext::layout_block_level_box(Box const& box, BlockContain
         resolve_used_height_if_treated_as_auto(box, available_space_for_height_resolution);
     }
 
+    // This monster basically means: "a ListItemBox that does not have specified content in the ::marker pseudo-element".
+    // This happens for ::marker with content 'normal'.
+    ListItemBox const* li_box = as_if<ListItemBox>(box);
+    bool is_list_item_box_without_css_content = li_box && (!(box.dom_node() && box.dom_node()->is_element() && as_if<DOM::Element const>(box.dom_node())->computed_properties(CSS::PseudoElement::Marker)->property(CSS::PropertyID::Content).is_content()));
```
AtkinsSJ, line 813: Pseudo-elements can also produce ListItemBoxes. (Specifically `::before` and `::after` can, but maybe others.)

For example:
```html
<!doctype html>
<style>
    div {
        margin: 100px;
    }
    div::before {
        display: list-item;
        content: "before";
        color: blue;
    }
    div::before::marker {
        content: "marker";
        color: orange;
    }
</style>
<div>Hello</div>
```

Looks like this in Firefox:
image

We don't currently allow `::before::marker`, but please add a FIXME here.

yisibl: And more `::before::column`, `::before::column::scroll-marker`...
https://github.com/browserscore/features/pull/269/files

manuel-za:
Would it be worth it to also take a look at #5016? (closed stale by bot)
I'm working on #4782 (expect PR next week). `list-item` counters will need to handle `reversed`...
Thanks!!

Issue #4188, opened 2025-04-02
macOS autocomplete popover is focus-stealing on some systems


trflynn89:
When autocomplete is enable via `about:settings`, we display a `NSPopover` (https://github.com/LadybirdBrowser/ladybird/blob/master/UI/AppKit/Interface/Autocomplete.mm) to show autocomplete results. On some macs, this popover removes focus from the location text field when results change.

https://github.com/user-attachments/assets/7c45dd73-8fdf-438a-94a9-529cd4518248

Ref https://github.com/LadybirdBrowser/ladybird/pull/4156#issuecomment-2769130100

nico:
Maybe one of you has full keyboard access enabled and the other doesn't?

trflynn89:
Ah! Yeah If I enable that, then I get the same behavior as Jelle.
@gmta do you have that enabled?

(I had to disable it again to write this comment, this comment box kept loosing focus too 🙃 )

gmta:
@nico @trflynn89 Nope, do not have that enabled.

gmta:
@cheif I tried your branch, didn't fix the issue unfortunately.

I'm not sure if I mentioned this before, but if I try to click an entry in the autocomplete popover, nothing happens apart from the popover disappearing. So not only does it steal the focus, the autocomplete feature seems to not be working even if I click an item.

cheif:
@gmta Interesting!

> So not only does it steal the focus, the autocomplete feature seems to not be working even if I click an item.

When I started testing this I realized that the autocomplete feature doesn't work if you don't also enable search, since when you click an item in the autocomplete list it triggers the `SearchEngine` for doing the search, and if that's not enabled, nothing happens. That sounds very much like what you're describing.

I was going to file another issue about it, but haven't gotten to it yet, in my opinion there's (at least) two ways forward to solve that particular case:
1. Simply disable autocomplete if search isn't enabled, making it clear that there's a dependency on search for autocomplete to work.
2. Make autocomplete a sub-option of search, so if you've enabled search and selected one of the engines that support autocomplete, we give you the option to also enable autocomplete. This would make the settings screen a tad bit easier, but maybe there's legitimate cases where you'd be interested in having different engines for search/autocomplete?

hugmouse:
@trflynn89 can you try again with the latest version/commit? Can't replicate the issue :o

gmta:
[...]

PR #5648, opened 2025-07-30
UI/AppKit: auto-focus location on new tab

### LINKED ISSUE #1512: Search bar not auto-selected when opening a new tab in window mode on macOS
When using the browser on macOS, opening a new tab in windowed mode (which is the default mode the browser starts in) does not automatically focus on the search bar. This requires the user to manually click on the search bar in order to type. However, when the browser is switched to fullscreen mode, opening a new tab does auto-select the search bar, allowing immediate typing.

**Steps to Reproduce:**

1. Open Ladybird Browser on macOS.
2. Ensure the browser is in windowed mode (not fullscreen).
3. Open a new tab.
4. Attempt to type in the search bar.

**Expected Behavior:**
The search bar should automatically be selected when a new tab is opened, allowing immediate typing without needing to manually click the search bar.

**Actual Behavior:**
In windowed mode, the search bar is not selected automatically, and the user must manually click into the search bar to begin typing. This behavior does not occur in fullscreen mode, where the search bar is auto-selected as expected.

**Operating System:**
macOS 14.6.1 (23G93)

**Browser Chrome:**
AppKit

I would like to try and fix this bug myself, however I am not sure of my skills so any support in pointing me in the right direction is appreciated.

rhelmer:
Fixes #1512

Auto-focus on location/address bar when opening a new tab on macOS. I tested with QT and it seemed to behave this way already.


trflynn89:
Hmm our AppKit app actually already behaves this way for me on the `master` branch. We invoke the same method in the tab controller's `showWindow` method already:
https://github.com/LadybirdBrowser/ladybird/blob/8537f0fa829ab225c348d1de4ecceaa2ba61ff92/UI/AppKit/Interface/TabController.mm#L614

https://github.com/user-attachments/assets/e93038c3-5813-4923-93bf-92517ce161b6

Is that not actually behaving this way for you?

If not, we probably don't need the invocation to `focusLocationToolbarItem` in both locations, i.e. it could be removed from `showWindow` (perhaps there's a timing issue here).

(I would also suggest we don't need the code comment. The code is self-explanatory here.)

rhelmer:
> Hmm our AppKit app actually already behaves this way for me on the `master` branch. We invoke the same method in the tab controller's `showWindow` method already:
> 
> https://github.com/LadybirdBrowser/ladybird/blob/8537f0fa829ab225c348d1de4ecceaa2ba61ff92/UI/AppKit/Interface/TabController.mm#L614
> 
> Screen.Recording.2025-08-02.at.8.43.17.AM.mp4
> 
> Is that not actually behaving this way for you?

I just pulled and built `master` (8a4e9f571f546d6f7e5d5c4d6396fc677a5277f2) as a "Release" build, here's what it looks like for me:

https://github.com/user-attachments/assets/486d8df7-08c9-4831-a0dc-75b04cf3de94

> If not, we probably don't need the invocation to `focusLocationToolbarItem` in both locations, i.e. it could be removed from `showWindow` (perhaps there's a timing issue here).

I noticed testing just now that the new window doesn't always get focus, removing from `showWindow` doesn't seem to make it any worse. I managed to catch it happening over several restarts:

https://github.com/user-attachments/assets/37f8b9ff-e927-4d3f-bcab-3372177bffde

rhelmer:
It looks like the focus issue is caused by `onURLChange`, which was overriding the location bar focus.

In commit 8dad9e841e73bebfe1fecee6bb20733db05d29d8, I added a guard that checks `self.window.firstResponder` before calling `makeFirstResponder`. I've tried several restarts and cannot reproduce the original behavior.

Let me know if you're prefer a different approach.

trflynn89:
Ah, that line in `onURLChange` was just added a couple days ago in 8537f0fa829ab225c348d1de4ecceaa2ba61ff92. CC @gmta .

gmta:
> Ah, that line in `onURLChange` was just added a couple days ago in [8537f0f](https://github.com/LadybirdBrowser/ladybird/commit/8537f0fa829ab225c348d1de4ecceaa2ba61ff92). CC @gmta .

I cannot reproduce this issue on current `master`... I thought this might relate to #4188 somehow, but that issue does not reproduce for me on my M4 Pro 🤔 what machine are you running this on @rhelmer ?

rhelmer:
> > Ah, that line in `onURLChange` was just added a couple days ago in [8537f0f](https://github.com/LadybirdBrowser/ladybird/commit/8537f0fa829ab225c348d1de4ecceaa2ba61ff92). CC @gmta .
> 
> I cannot reproduce this issue on current `master`... I thought this might relate to #4188 somehow, but that issue does not reproduce for me on my M4 Pro 🤔 what machine are you running this on @rhelmer ?

This is an Apple M1 8GB.

PR #5650, opened 2025-07-30
LibWeb: Implement validationMessage for elements that used it

abnormalhare:
HTML Elements with the field validationMessage are now handled to contain the correct message when retrieved.

### DIFF
--- a/Libraries/LibWeb/HTML/FormAssociatedElement.cpp
+++ b/Libraries/LibWeb/HTML/FormAssociatedElement.cpp
@@ -58,6 +58,28 @@ GC::Ref<ValidityState const> FormAssociatedElement::validity() const
     return realm.create<ValidityState>(realm, *this);
 }
 
+// https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#dom-cva-validationmessage
+String FormAssociatedElement::validation_message()
+{
+    // 1. If this element is not a candidate for constraint validation
+    // or if this element satisfies its constraints, then return the empty string.
+    if (!is_candidate_for_constraint_validation() || satisfies_its_constraints()) {
+        return {};
+    }
+
+    // 2. Return a suitably localized message that the user agent would show the user
+    // if this were the only form control with a validity constraint problem. If the user agent would not
+    // actually show a textual message in such a situation (e.g., it would show a graphical cue instead),
+    // then return a suitably localized message that expresses (one or more of) the validity constraint(s) that the control
+    // does not satisfy. If the element is a candidate for constraint validation and is suffering from a custom error,
+    // then the custom validity error message should be present in the return value.
+    if (suffering_from_a_custom_error()) {
+        return m_custom_validity_error_message;
+    }
+
+    return format_validation_message();
+}
+
 // https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#dom-cva-setcustomvalidity
 void FormAssociatedElement::set_custom_validity(String& error)
 {
--- a/Libraries/LibWeb/HTML/HTMLInputElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLInputElement.cpp
@@ -2855,6 +2855,59 @@ WebIDL::ExceptionOr<bool> HTMLInputElement::report_validity()
     return report_validity_steps();
 }
 
+// https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#dom-cva-validationmessage
+String HTMLInputElement::format_validation_message()
+{
+    if (suffering_from_being_missing()) {
+        return "Please input a value."_string;
+    }
+    if (suffering_from_a_type_mismatch()) {
+        switch (type_state()) {
+        case TypeAttributeState::Color:
+            return "Please input a valid color. (#rrggbb)"_string;
+        case TypeAttributeState::Date:
+            return "Please input a valid date. (yy-mm-dd)"_string;
+        case TypeAttributeState::Email:
+            return "Please input a valid email. (username@example.com)"_string;
+        case TypeAttributeState::Number:
+        case TypeAttributeState::Range:
+            return "Please input a valid number."_string;
+        case TypeAttributeState::Password:
+            return "Please input a valid password."_string;
+        case TypeAttributeState::Month:
+            return "Please input a valid month."_string;
+        case TypeAttributeState::Week:
+            return "Please input a valid week."_string;
+        case TypeAttributeState::Telephone:
+            return "Please input a valid phone number."_string;
+        case TypeAttributeState::URL:
+            return "Please input a valid URL. (https://example.com)"_string;
+        default:
+            return "Please input a valid value."_string;
+        }
+    }
+    if (suffering_from_a_pattern_mismatch() || suffering_from_bad_input()) {
+        return "Please input a valid value."_string;
+    }
+    if (suffering_from_being_too_long()) {
+        return MUST(String::formatted("Please limit your character count to {} characters.", max_length()));
+    }
+    if (suffering_from_being_too_short()) {
+        return MUST(String::formatted("Please increase your character count to {} characters.", min_length()));
+    }
+    if (suffering_from_an_underflow()) {
+        return MUST(String::formatted("Please select a number at or above {}.", min()));
+    }
+    if (suffering_from_an_overflow()) {
+        return MUST(String::formatted("Please select a number at or below {}.", max()));
+    }
+    if (suffering_from_a_step_mismatch()) {
+        return "Please input a valid number."_string;
+    }
+
+    return {};
+}
+
 Optional<ARIA::Role> HTMLInputElement::default_role() const
 {
     // http://wpt.live/html-aam/roles-dynamic-switch.tentative.window.html "Disconnected <input type=checkbox switch>"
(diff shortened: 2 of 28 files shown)


Libraries/LibWeb/HTML/FormAssociatedElement.cpp
```diff
@@ -58,6 +58,26 @@ GC::Ref<ValidityState const> FormAssociatedElement::validity() const
...
+
+    // 2b. If the element is a candidate for constraint validation
+    //   and is suffering from a custom error, then the custom validity error message
+    //   should be present in the return value.
+    if (suffering_from_a_custom_error()) {
+        return m_custom_validity_error_message;
+    }
+
+    // (the rest of the errors should be handled in derived classes)
+    return {};
```
tcl3: Rather than having the derived classes call this base class method, you could have a have a virtual method that returns the empty string by default that you override in each derived class and call here. This means that you don't end up repeating steps like calling `suffering_from_a_custom_error()` multiple times.

That would make this function look like this:

```c++
String FormAssociatedElement::validation_message() const
{
    // 1. If this element is not a candidate for constraint validation
    //   or if this element satisfies its constraints, then return the empty string.
    if (!is_candidate_for_constraint_validation() || satisfies_its_constraints()) {
        return {};
    }

    // 2b. If the element is a candidate for constraint validation
    //   and is suffering from a custom error, then the custom validity error message
    //   should be present in the return value.
    if (suffering_from_a_custom_error()) {
        return m_custom_validity_error_message;
    }

    return format_validation_message();
}
```
And in each derived class you'd do something like this:
```c++
String HTMLButtonElement::format_validation_message() const
{
    if (suffering_from_being_missing()) {
        return "Please input a value."_string;
    }
    if (suffering_from_bad_input()) {
        return "Please input a valid value."_string;
    }

    return {};
}
```

EDIT: I made a mistake with the above example initially, sorry about that!

Would something like this work?

Libraries/LibWeb/HTML/HTMLFieldSetElement.cpp
```diff
@@ -99,4 +100,11 @@ bool HTMLFieldSetElement::will_validate()
     return false;
 }
 
+// https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#dom-cva-validationmessage
+String HTMLFieldSetElement::validation_message() const
+{
+    // Returns the empty string, as fieldset elements are not subject to constraint validation.
+    return {};
+}
```
tcl3: I don't think this method is necessary. If this method wasn't here `FormAssociatedElement::validation_message()` should be called, which will return the empty string

Libraries/LibWeb/HTML/HTMLInputElement.cpp
```diff
@@ -11,6 +11,7 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
+#include "LibWeb/HTML/FormAssociatedElement.h"
```
tcl3: Includes should use angle brackets and be sorted alphabetically.

Libraries/LibWeb/HTML/HTMLObjectElement.cpp
```diff
@@ -92,6 +92,13 @@ bool HTMLObjectElement::will_validate()
     return false;
 }
 
+// https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#dom-cva-validationmessage
+String HTMLObjectElement::validation_message() const
+{
+    // Returns the empty string, as object elements are not subject to constraint validation.
+    return {};
+}
```
tcl3: I don't think this method is necessary because `FormAssociatedElement::validation_message()` would also return the empty string.

Libraries/LibWeb/HTML/HTMLOutputElement.cpp
```diff
@@ -121,4 +121,11 @@ bool HTMLOutputElement::will_validate()
     return false;
 }
 
+// https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#dom-cva-validationmessage
+String HTMLOutputElement::validation_message() const
+{
+    // Returns the empty string, as output elements are not subject to constraint validation.
+    return {};
+}
```
tcl3: I don't think this method is necessary because `FormAssociatedElement::validation_message()` would also return the empty string.

tcl3 (changes requested):
This change fixes quite a few Web Platform Tests that were previously failing:

* http://wpt.live/html/semantics/forms/constraints/form-validation-validity-customError.html
* http://wpt.live/html/semantics/forms/the-button-element/button-validationmessage.html
* http://wpt.live/html/semantics/forms/the-input-element/input-validationmessage.html
* http://wpt.live/html/semantics/forms/the-fieldset-element/fieldset-validationmessage.html
* http://wpt.live/html/semantics/forms/the-output-element/output-validity.html

It would be great if you could import these tests, to prevent future regressions. You can do that by running `./Meta/WPT.sh import html/semantics/forms/constraints/form-validation-validity-customError.html html/semantics/forms/the-button-element/button-validationmessage.html html/semantics/forms/the-input-element/input-validationmessage.html html/semantics/forms/the-fieldset-element/fieldset-validationmessage.html html/semantics/forms/the-output-element/output-validity.html`

You should use `git commit --amend` to update your initial commit when making changes, rather than adding new commits to fix previous ones. [This video](<https://www.youtube.com/watch?v=ElRzTuYln0M>) gives a good overview of how to rewrite git history. Feel free to ask on Discord if you run into any trouble.

Libraries/LibWeb/HTML/FormAssociatedElement.cpp
```diff
@@ -58,6 +58,26 @@ GC::Ref<ValidityState const> FormAssociatedElement::validity() const
     return realm.create<ValidityState>(realm, *this);
 }
 
+// https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#dom-cva-validationmessage
+String FormAssociatedElement::validation_message() const
+{
+    // 1. If this element is not a candidate for constraint validation
+    //   or if this element satisfies its constraints, then return the empty string.
+    if (is_candidate_for_constraint_validation() || satisfies_its_constraints()) {
```
stelar7: Missing ! here?

abnormalhare: no, its not. if the constraints are satisfied, it should just return an empty string.

gmta: I'm guessing @stelar7 meant the first condition here, `is_candidate_for_constraint_validation()`, which indeed seems to be missing `!` in front of it.

Libraries/LibWeb/HTML/FormAssociatedElement.cpp
abnormalhare: yes

abnormalhare:
The tests that ran from your command all passed!

Libraries/LibWeb/HTML/FormAssociatedElement.cpp
tcl3: A few subtests on the tests you've imported are failing, negating `is_candidate_for_constraint_validation()`, as suggested should fix these. After doing this you'll need to rebaseline the expectation files for the affected tests. You can do this by running `./Meta/ladybird.py run test-web --rebaseline`

abnormalhare: dang, i really thought i got DeMorgan on that one. next commit should fix it

abnormalhare:
Latest force push updated test results

Libraries/LibWeb/HTML/FormAssociatedElement.cpp
```diff
@@ -58,6 +58,26 @@ GC::Ref<ValidityState const> FormAssociatedElement::validity() const
     return realm.create<ValidityState>(realm, *this);
 }
 
+// https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#dom-cva-validationmessage
+String FormAssociatedElement::validation_message()
+{
+    // 1. If this element is not a candidate for constraint validation
+    //   or if this element satisfies its constraints, then return the empty string.
```
tcl3: Please make sure each new line is aligned with the text above.

Libraries/LibWeb/HTML/FormAssociatedElement.cpp
```diff
@@ -58,6 +58,26 @@ GC::Ref<ValidityState const> FormAssociatedElement::validity() const
...
+    if (!is_candidate_for_constraint_validation() || satisfies_its_constraints()) {
+        return {};
+    }
+
+    // 2b. If the element is a candidate for constraint validation
+    //   and is suffering from a custom error, then the custom validity error message
+    //   should be present in the return value.
```
tcl3: This comment does not match the spec text from the link above.

Libraries/LibWeb/HTML/FormAssociatedElement.cpp
```diff
@@ -58,6 +58,26 @@ GC::Ref<ValidityState const> FormAssociatedElement::validity() const
...
+    if (suffering_from_a_custom_error()) {
+        return m_custom_validity_error_message;
+    }
+
+    // (the rest of the errors should be handled in derived classes)
```
tcl3: I think this comment can be removed, since what is happening here should be explained by the spec text above.

PR #5654, opened 2025-07-30
Base: Stop relying on const-ness not being enforced by LibJS

emmanuel-ferdman:
## PR Summary
This PR fixes a runtime issue in `addCustomSearchEngine` by replacing `const` with `let` in a for loop to allow correct iteration and engine name comparison.

### DIFF
--- a/Base/res/ladybird/about-pages/settings/search.js
+++ b/Base/res/ladybird/about-pages/settings/search.js
@@ -164,7 +164,7 @@ function addCustomSearchEngine() {
     searchCustomName.classList.remove("error");
     searchCustomURL.classList.remove("error");
 
-    for (const i = 0; i < searchEngine.length; ++i) {
+    for (let i = 0; i < searchEngine.length; ++i) {
         if (searchCustomName.value === searchEngine.item(i).value) {
             searchCustomName.classList.add("error");
             return;


Base/res/ladybird/about-pages/settings/search.js
```diff
@@ -164,7 +164,7 @@ function addCustomSearchEngine() {
     searchCustomName.classList.remove("error");
     searchCustomURL.classList.remove("error");
 
-    for (const i = 0; i < searchEngine.length; ++i) {
+    for (let i = 0; i < searchEngine.length; ++i) {
```
trflynn89, line 167: Can you explain what issue this fixes? We are able to iterate the engine list just fine with `const` here

awesomekling: Seems more like there is an issue in LibJS if we allow `++i` where `const i`, no? :thonk:

trflynn89: ah yup seems that way!

In that case, @emmanuel-ferdman could you update the commit message to be more descriptive? See this note in our [contributing guide](https://github.com/LadybirdBrowser/ladybird/blob/master/CONTRIBUTING.md):
> Don't:
> * Use weasel-words like "refactor" or "fix" to avoid explaining what's being changed.

awesomekling: Filed #5683 for the LibJS issue.

InvalidUsernameException:
You only changed the PR description, but not the commit message. And the commits are the important part, because in this repo PRs are merged by cherry-picking the individual commits on the `master` branch.

Also I'd argue that the new message could still be better. Right now it basically just repeats what the commit diff already shows. Something like this would explain the rationale for the change:
> Base: Stop relying on const-ness not being enforced by LibJS

PR #5655, opened 2025-07-30
LibWeb: Implement `anchor-size(..)` function parsing

gmta:
This parses `anchor-size(..)` functions in CSS, but does not yet result in a useful `Size`: we need style & layout interleaving similar to container queries for this, since the resulting value depends on layout results.

Not supported yet: `anchor-size()` appearing inside a `calc()` node.

Adds 4280 WPT subtest passes in `css/css-anchor-position`.


Libraries/LibWeb/CSS/Parser/ValueParsing.cpp
```diff
@@ -791,6 +795,107 @@ RefPtr<CSSStyleValue const> Parser::parse_percentage_value(TokenStream<Component
...
+    // Parse optional anchor name and anchor size in arbitrary order.
+    auto parse_name_or_size = [&]() {
+        argument_tokens.discard_whitespace();
+        auto peek_token = argument_tokens.next_token();
+        if (!peek_token.is(Token::Type::Ident))
+            return;
+
+        // <anchor-name> = <dashed-ident>
+        if (auto dashed_ident = parse_dashed_ident(argument_tokens); dashed_ident.has_value()) {
+            anchor_name = dashed_ident.value();
```
tcl3, line 861: Parsing should fail if anchor name already has a value.

Libraries/LibWeb/CSS/Parser/ValueParsing.cpp
```diff
@@ -791,6 +795,107 @@ RefPtr<CSSStyleValue const> Parser::parse_percentage_value(TokenStream<Component
...
+            return;
+        }
+
+        // <anchor-size> = width | height | block | inline | self-block | self-inline
+        auto keyword = keyword_from_string(peek_token.token().ident());
+        if (keyword.has_value() && first_is_one_of(*keyword, Keyword::Width, Keyword::Height, Keyword::Block, Keyword::Inline, Keyword::SelfBlock, Keyword::SelfInline)) {
+            argument_tokens.discard_a_token();
+            anchor_size = keyword;
```
tcl3: Parsing should fail if anchor size already has a value.

Libraries/LibWeb/CSS/Parser/ValueParsing.cpp
```diff
@@ -791,6 +795,107 @@ RefPtr<CSSStyleValue const> Parser::parse_percentage_value(TokenStream<Component
...
+    auto argument_tokens = TokenStream { function_token.function().value };
+
+    Optional<FlyString> anchor_name;
+    Optional<Keyword> anchor_size;
+    Variant<Empty, LengthPercentage, NonnullRefPtr<AnchorSizeStyleValue const>> fallback_value;
+
+    // Parse optional anchor name and anchor size in arbitrary order.
+    auto parse_name_or_size = [&]() {
+        argument_tokens.discard_whitespace();
+        auto peek_token = argument_tokens.next_token();
```
tcl3: This can be `auto&`

Libraries/LibWeb/CSS/Parser/ValueParsing.cpp
```diff
@@ -791,6 +795,107 @@ RefPtr<CSSStyleValue const> Parser::parse_percentage_value(TokenStream<Component
...
+
+        // <anchor-name> = <dashed-ident>
+        if (auto dashed_ident = parse_dashed_ident(argument_tokens); dashed_ident.has_value()) {
+            anchor_name = dashed_ident.value();
+            return;
+        }
+
+        // <anchor-size> = width | height | block | inline | self-block | self-inline
+        auto keyword = keyword_from_string(peek_token.token().ident());
+        if (keyword.has_value() && first_is_one_of(*keyword, Keyword::Width, Keyword::Height, Keyword::Block, Keyword::Inline, Keyword::SelfBlock, Keyword::SelfInline)) {
```
tcl3: Parsing should fail if the parsed keyword isn't valid.

Rather than checking individual keyword values it may be better to add an `anchor-size` enum to `Enums.json` then you could use `keyword_to_anchor_size()` to check whether the parsed keyword is valid. I'm assuming that enum would come in useful later too.

Libraries/LibWeb/CSS/Parser/ValueParsing.cpp
```diff
@@ -791,6 +795,107 @@ RefPtr<CSSStyleValue const> Parser::parse_percentage_value(TokenStream<Component
...
+
+    auto context_guard = push_temporary_value_parsing_context(FunctionContext { function_token.function().name });
+    auto argument_tokens = TokenStream { function_token.function().value };
+
+    Optional<FlyString> anchor_name;
+    Optional<Keyword> anchor_size;
+    Variant<Empty, LengthPercentage, NonnullRefPtr<AnchorSizeStyleValue const>> fallback_value;
+
+    // Parse optional anchor name and anchor size in arbitrary order.
+    auto parse_name_or_size = [&]() {
```
tcl3: This needs some way to indicate failure if we attempt to do something invalid. Elsewhere in the parser, this sort of thing is done with a while loop with a `maximum_allowed_values` counter.

Libraries/LibWeb/CSS/Parser/ValueParsing.cpp
```diff
@@ -791,6 +795,107 @@ RefPtr<CSSStyleValue const> Parser::parse_percentage_value(TokenStream<Component
...
+    // FIXME: Nested anchor sizes should actually be handled by parse_length_percentage()
+    if (auto nested_anchor_size = parse_anchor_size(argument_tokens))
+        fallback_value = nested_anchor_size.release_nonnull();
+    else if (auto length_percentage = parse_length_percentage(argument_tokens); length_percentage.has_value())
+        fallback_value = length_percentage.value();
+
+    if (fallback_value.has<Empty>() && fallback_argument_required)
+        return {};
+
+    if (!tokens.next_token().is(Token::Type::EndOfFile))
```
tcl3: This is equivalent to `tokens.has_next_value()`.

Libraries/LibWeb/CSS/Parser/ValueParsing.cpp
```diff
@@ -791,6 +795,107 @@ RefPtr<CSSStyleValue const> Parser::parse_percentage_value(TokenStream<Component
...
+    }
+    if (!valid_property_context)
+        return {};
+
+    auto context_guard = push_temporary_value_parsing_context(FunctionContext { function_token.function().name });
+    auto argument_tokens = TokenStream { function_token.function().value };
+
+    Optional<FlyString> anchor_name;
+    Optional<Keyword> anchor_size;
+    Variant<Empty, LengthPercentage, NonnullRefPtr<AnchorSizeStyleValue const>> fallback_value;
```
tcl3: I'm not 100% sure, but I think it may be better to parse the `LengthPercentage` as a `LengthPercentageStyleValue`, then `fallback_value` could be a `RefPtr<CSSStyleValue const>`. I think you might see an advantage when serializing calculated values in `AnchorSizeStyleValue`.

tcl3 (commented):
Can we import the improved parsing tests. http://wpt.live/css/css-anchor-position/anchor-size-parse-valid.html runs reasonably quickly on my machine when imported.

If it's too slow can we at least import `css/css-anchor-position/anchor-size-parse-invalid.html` to prevent future regressions.

Libraries/LibWeb/CSS/Parser/ValueParsing.cpp
gmta: Couldn't find any `maximum_allowed_values`, but this is well-suited for a simple `for` loop :)

Libraries/LibWeb/CSS/Parser/ValueParsing.cpp
gmta: Done!

Libraries/LibWeb/CSS/Parser/ValueParsing.cpp
```diff
@@ -791,6 +795,112 @@ RefPtr<CSSStyleValue const> Parser::parse_percentage_value(TokenStream<Component
...
+    if (auto nested_anchor_size = parse_anchor_size(argument_tokens))
+        fallback_value = nested_anchor_size.release_nonnull();
+    else if (auto length_percentage = parse_length_percentage_value(argument_tokens))
+        fallback_value = length_percentage.release_nonnull();
+
+    if (fallback_value.has<Empty>() && comma_present)
+        return {};
+    if (!fallback_value.has<Empty>() && !comma_present && has_name_or_size)
+        return {};
+    if (tokens.has_next_token())
```
tcl3: Changing this to `argument_tokens.has_next_token()` fixes the failing subtests in `anchor-size-parse-invalid.html`

Libraries/LibWeb/CSS/Parser/ValueParsing.cpp
```diff
@@ -791,6 +795,112 @@ RefPtr<CSSStyleValue const> Parser::parse_percentage_value(TokenStream<Component
...
+    }
+    if (!valid_property_context)
+        return {};
+
+    auto context_guard = push_temporary_value_parsing_context(FunctionContext { function_token.function().name });
+    auto argument_tokens = TokenStream { function_token.function().value };
+
+    Optional<FlyString> anchor_name;
+    Optional<AnchorSize> anchor_size;
+    Variant<Empty, NonnullRefPtr<CSSStyleValue const>> fallback_value;
```
tcl3: I don't feel too strongly about it and I understand why this might be preferable, but personally I would change this to `{ValueComparing}RefPtr<CSSStyleValue const>` for consistency - since that is the pattern we use everywhere else.

gmta: Excellent suggestion; done!

tcl3 (approved):

PR #5657, opened 2025-07-30
LibWeb: More thoroughly detach layout/paint trees from each other

awesomekling:
Before committing a new layout (and thus building a new paint tree) we now go through both the old paint tree and the layout tree and detach them from each other.

This is a little extra work, but it ensures that there are no lingering references across the trees, which we were apparently accumulating in some cases on Discord, causing GC leaks.


Libraries/LibWeb/Layout/LayoutState.cpp
```diff
@@ -214,14 +214,23 @@ static void build_paint_tree(Node& node, Painting::Paintable* parent_paintable =
...
+    // Go through the old paintable tree and detach everything from the layout tree.
+    // The layout tree should only point to the new paintable tree which we're about to build.
+    if (auto* old_paintable_root = root.first_paintable()) {
+        old_paintable_root->for_each_in_inclusive_subtree([&](Painting::Paintable& paintable) {
+            paintable.detach_from_layout_node();
+            return TraversalDecision::Continue;
+        });
+    }
+
+    // For completeness, also go thorough the layout tree and detach all paintables.
```
kalenikaliaksandr: typo: thorough -> through

PR #5664, opened 2025-07-31
LibWeb: Improve cursor appearance

zacoons:
Makes the cursor less blurry looking.

`draw_rect` renders the outline of a rectangle. `fill_rect` seems like a better option for rendering the cursor.


gmta:
Thanks!

PR #5665, opened 2025-07-31
LibWeb: Start implementing Encrypted Media Extensions

stelar7:
Also fixes 2 IDL bugs :)


Meta/Lagom/Tools/CodeGenerators/LibWeb/BindingsGenerator/IDLGenerators.cpp
```diff
@@ -1831,6 +1831,7 @@ static void generate_wrap_statement(SourceGenerator& generator, ByteString const
 {
     auto scoped_generator = generator.fork();
     scoped_generator.set("value", value);
+    scoped_generator.set("value_no_dot", value.replace("."sv, "_"sv));
```
gmta: We've used `cpp_name` (or `value_cpp_name` in this case) in other places; seems more fitting.

Libraries/LibWeb/EncryptedMediaExtensions/KeySystem.h
```diff
@@ -0,0 +1,71 @@
...
+    ClearKeySystem() = default;
+    virtual ~ClearKeySystem() override = default;
+
+    // https://w3c.github.io/encrypted-media/#clear-key-behavior
+    virtual bool supports_init_data_type(String const& init_data_type) const override
+    {
+        // Implementations SHOULD support the "keyids" type
+        // TODO: Implementations MAY support any combination of registered Initialization Data Types
+        // https://www.w3.org/TR/eme-initdata-registry/
+        auto registered_init_data_types = Vector {
```
gmta: We normally use `constexpr Array`s for these types of lists.

Libraries/LibWeb/EncryptedMediaExtensions/MediaKeySystemAccess.idl
```diff
@@ -0,0 +1,32 @@
...
+    MediaKeysRequirement                    distinctiveIdentifier = "optional";
+    MediaKeysRequirement                    persistentState = "optional";
+    sequence<DOMString>                     sessionTypes;
+};
+
+// https://w3c.github.io/encrypted-media/#dom-mediakeysystemmediacapability
+dictionary MediaKeySystemMediaCapability {
+    DOMString contentType = "";
+    DOMString? encryptionScheme = null;
+    DOMString robustness = "";
```
gmta: Could we use the UTF-16 variants for all of these DOMStrings?

Libraries/LibWeb/DOM/Document.h
```diff
@@ -158,6 +158,7 @@ struct ElementCreationOptions {
 enum class PolicyControlledFeature : u8 {
     Autoplay,
     FocusWithoutUserActivation,
+    EncryptedMedia,
```
gmta, line 161: Let's keep this enum in alphabetical order

Libraries/LibWeb/EncryptedMediaExtensions/Algorithms.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+        // 13. If mimeType is not strictly an audio/video type, continue to the next iteration.
+        if (!mime_type->is_audio_or_video())
+            continue;
+
+        // 14. If encryption scheme is non-null and is not recognized or not supported by implementation, continue to the next iteration.
+        if (encryption_scheme.has_value() && !implementation->supports_encryption_scheme(*encryption_scheme))
+            continue;
+
+        // 15. If robustness is not the empty string and contains an unrecognized value or a value not supported by implementation, continue to the next iteration. String comparison is case-sensitive.
```
gmta: Could you limit these lines [to 120 characters](https://github.com/LadybirdBrowser/ladybird/blob/master/Documentation/CodingStyle.md#comments)?

Libraries/LibWeb/EncryptedMediaExtensions/Algorithms.cpp
```diff
@@ -361,4 +361,24 @@ Optional<ConsentConfiguration> get_supported_configuration(RefPtr<KeySystem> imp
     return supported_configuration;
 }
 
+// https://w3c.github.io/encrypted-media/#dfn-common-key-systems
+bool is_supported_key_system(String const& key_system)
+{
+    auto const& supported_key_systems = Vector {
```
gmta: This can be an `Array` as well.

Libraries/LibWeb/EncryptedMediaExtensions/KeySystem.h
```diff
@@ -0,0 +1,71 @@
...
+        };
+
+        return registered_init_data_types.contains_slow(init_data_type);
+    }
+
+    // https://w3c.github.io/encrypted-media/#clear-key-capabilities
+    virtual bool supports_encryption_scheme(String const& encryption_scheme) const override
+    {
+        // encryptionScheme: Implementations MUST support the "cenc" scheme, and MAY support other schemes.
+        return encryption_scheme == "cenc"_string;
```
trflynn89: This PR contains a lot of `== "something"_string`. Those are unnecessary allocations - they should just be `== "something"sv`.

Libraries/LibWeb/EncryptedMediaExtensions/EncryptedMediaExtensions.h
```diff
@@ -1,20 +1,42 @@
...
+};
+
+// https://w3c.github.io/encrypted-media/#dom-mediakeysystemconfiguration
+struct MediaKeySystemConfiguration {
+    String label;
+    Vector<String> init_data_types;
+    Vector<MediaKeySystemMediaCapability> audio_capabilities;
+    Vector<MediaKeySystemMediaCapability> video_capabilities;
+    Bindings::MediaKeysRequirement distinctive_identifier = Bindings::MediaKeysRequirement::Optional;
+    Bindings::MediaKeysRequirement persistent_state = Bindings::MediaKeysRequirement::Optional;
```
trflynn89: We typically use `{ Bindings::MediaKeysRequirement::Optional }` for struct/class member init

Libraries/LibWeb/EncryptedMediaExtensions/Algorithms.cpp
```diff
@@ -0,0 +1,122 @@
...
+namespace Web::EncryptedMediaExtensions {
+
+bool supports_container([[maybe_unused]] String const& container)
+{
+    // FIXME: Check FFmpeg?
+    return true;
+}
+
+// https://w3c.github.io/encrypted-media/#get-supported-capabilities-for-audio-video-type
+Vector<Bindings::MediaKeySystemMediaCapability> get_supported_capabilities_for_audio_video_type(RefPtr<KeySystem> implementation, CapabilitiesType type, Vector<Bindings::MediaKeySystemMediaCapability> requested_capabilities, Bindings::MediaKeySystemConfiguration config, MediaKeyRestrictions restrictions)
```
trflynn89: There's a lot of unnecessary copying going on in this function declaration:
* `RefPtr<KeySystem> implementation` -> `RefPtr<KeySystem> const& implementation`
* `Vector<Bindings::MediaKeySystemMediaCapability> requested_capabilities` -> `Span<Bindings::MediaKeySystemMediaCapability> requested_capabilities`
* `Bindings::MediaKeySystemConfiguration config` -> `Bindings::MediaKeySystemConfiguration const& config`

trflynn89: Though, this function is assuming `implementation` is non-null. Therefore, it should be `NonnullRefPtr<KeySystem> const& implementation` or, better yet, `KeySystem const& implementation`.

Libraries/LibWeb/EncryptedMediaExtensions/Algorithms.cpp
```diff
@@ -0,0 +1,122 @@
...
+            continue;
+
+        // 9. Let parameters be the "codecs" and "profiles" RFC 6381 [RFC6381] parameters, if any, of mimeType.
+        auto parameters = mime_type->parameters();
+
+        // FIXME: 10. If the user agent does not recognize one or more parameters, or
+        // if any parameters are not valid per the relevant specification, continue to the next iteration.
+
+        // 11. Let media types be the set of codecs and codec constraints specified by parameters.
+        // The case-sensitivity of string comparisons is determined by the appropriate RFC or other specification.
```
trflynn89: Please hang the indentation on new lines:
```c++
        // 11. Let media types be the set of codecs and codec constraints specified by parameters. The case-sensitivity 
        //     of string comparisons is determined by the appropriate RFC or other specification.
```

(here and in other places)

Libraries/LibWeb/EncryptedMediaExtensions/Algorithms.cpp
```diff
@@ -0,0 +1,122 @@
...
+    // 2. Let supported media capabilities be an empty sequence of MediaKeySystemMediaCapability dictionaries.
+    Vector<Bindings::MediaKeySystemMediaCapability> supported_media_capabilities;
+
+    // 3. For each requested media capability in requested media capabilities:
+    for (auto& capability : requested_capabilities) {
+        // 1. Let content type be requested media capability's contentType member.
+        auto content_type = capability.content_type;
+
+        // 2. Let encryption scheme be requested media capability’s encryptionScheme member.
+        auto encryption_scheme = capability.encryption_scheme;
```
trflynn89: `auto const&`

Libraries/LibWeb/EncryptedMediaExtensions/Algorithms.cpp
```diff
@@ -0,0 +1,122 @@
...
+    // 1. Let local accumulated configuration be a local copy of accumulated configuration.
+    Bindings::MediaKeySystemConfiguration accumulated_configuration = config;
+
+    // 2. Let supported media capabilities be an empty sequence of MediaKeySystemMediaCapability dictionaries.
+    Vector<Bindings::MediaKeySystemMediaCapability> supported_media_capabilities;
+
+    // 3. For each requested media capability in requested media capabilities:
+    for (auto& capability : requested_capabilities) {
+        // 1. Let content type be requested media capability's contentType member.
+        auto content_type = capability.content_type;
```
trflynn89: `auto const&`

Libraries/LibWeb/EncryptedMediaExtensions/Algorithms.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+        // 2. Let encryption scheme be requested media capability’s encryptionScheme member.
+        auto encryption_scheme = capability.encryption_scheme;
+
+        // 3. Let robustness be requested media capability's robustness member.
+        auto robustness = capability.robustness;
```
trflynn89: `auto const&`

Libraries/LibWeb/EncryptedMediaExtensions/Algorithms.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+        // 5. Let mimeType be the result of running parse a MIME type with content type.
+        auto mime_type = MimeSniff::MimeType::parse(content_type);
+
+        // 6. If mimeType is failure or is unrecognized, continue to the next iteration.
+        if (!mime_type.has_value())
+            continue;
+
+        // 7. Let container be the container type specified by mimeType.
+        auto container = mime_type->essence();
```
trflynn89: `auto const&`

Libraries/LibWeb/EncryptedMediaExtensions/Algorithms.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+        // 8. If the user agent does not support container, continue to the next iteration.
+        // The case-sensitivity of string comparisons is determined by the appropriate RFC.
+        if (!supports_container(container))
+            continue;
+
+        // 9. Let parameters be the "codecs" and "profiles" RFC 6381 [RFC6381] parameters, if any, of mimeType.
+        auto parameters = mime_type->parameters();
```
trflynn89, line 58: `auto const&`

Libraries/LibWeb/EncryptedMediaExtensions/Algorithms.cpp
```diff
@@ -0,0 +1,122 @@
...
+        // 12. If media types is empty:
+        if (media_types.is_empty()) {
+            // FIXME: If container normatively implies a specific set of codecs and codec constraints:
+            if (false) {
+                // Let parameters be that set.
+            } else {
+                // Otherwise:
+                // Continue to the next iteration.
+                continue;
+            }
```
trflynn89: We use `->` when the spec text has that hanging arrow thing. And the `Otherwise` belongs above the `else`:
```c++
// 12. If media types is empty:
if (media_types.is_empty()) {
    // FIXME: -> If container normatively implies a specific set of codecs and codec constraints:
    if (false) {
        // Let parameters be that set.
    }
    // -> Otherwise:
    else {
        // Continue to the next iteration.
        continue;
    }
}
```

Libraries/LibWeb/EncryptedMediaExtensions/Algorithms.cpp
```diff
@@ -0,0 +1,122 @@
...
+            continue;
+
+        // 14. If encryption scheme is non-null and is not recognized or not supported by implementation,
+        // continue to the next iteration.
+        if (encryption_scheme.has_value() && !implementation->supports_encryption_scheme(*encryption_scheme))
+            continue;
+
+        // 15. If robustness is not the empty string and contains an unrecognized value or
+        // a value not supported by implementation, continue to the next iteration.
+        // String comparison is case-sensitive.
```
trflynn89: On top of the hanging indent, the wrapping here is very inconsistent. Please wrap at about 120 characters.

Libraries/LibWeb/EncryptedMediaExtensions/Algorithms.cpp
```diff
@@ -0,0 +1,122 @@
...
+            if (type == CapabilitiesType::Audio) {
+                // Add requested media capability to the audioCapabilities member of local accumulated configuration.
+                accumulated_configuration.audio_capabilities.append(capability);
+            }
+        }
+    }
+
+    // 4. If supported media capabilities is empty, return null.
+    if (supported_media_capabilities.is_empty())
+        return {};
```
trflynn89, line 117: Instead of an empty vector representing null, we should be returning an `Optional<Vector<...>>`, where an empty optional represents null.

Libraries/LibWeb/EncryptedMediaExtensions/Algorithms.cpp
```diff
@@ -119,4 +119,248 @@ Vector<Bindings::MediaKeySystemMediaCapability> get_supported_capabilities_for_a
...
+    // 2. Follow the steps for the value of session type from the following list:
+
+    // * "temporary"
+    if (session_type == "temporary"_string) {
+        // Return false.
+        return false;
+    }
+
+    // * "persistent-license"
+    if (session_type == "persistent-license"_string) {
```
trflynn89: (these should also not be allocating strings for a comparison)

Libraries/LibWeb/EncryptedMediaExtensions/Algorithms.cpp
```diff
@@ -119,4 +119,248 @@ Vector<Bindings::MediaKeySystemMediaCapability> get_supported_capabilities_for_a
...
+    (void)restrictions;
+    (void)origin;
+
+    dbgln("get_consent_status: Not implemented, returning Allowed by default");
+
+    return ConsentStatus::Allowed;
+}
+
+// https://w3c.github.io/encrypted-media/#get-supported-configuration-and-consent
+Optional<ConsentConfiguration> get_supported_configuration_and_consent(RefPtr<KeySystem> implementation, Bindings::MediaKeySystemConfiguration const& candidate_configuration, MediaKeyRestrictions& restrictions, URL::Origin const& origin)
```
trflynn89: Given that this function is also assuming `implementation` is non-null, it at least be taking a `NonnullRefPtr` by const-ref, but can also just be `KeySystem const&`.

Libraries/LibWeb/EncryptedMediaExtensions/Algorithms.cpp
```diff
@@ -119,4 +119,248 @@ Vector<Bindings::MediaKeySystemMediaCapability> get_supported_capabilities_for_a
...
+        auto video_capabilities = get_supported_capabilities_for_audio_video_type(implementation, CapabilitiesType::Video, candidate_configuration.video_capabilities, accumulated_configuration, restrictions);
+
+        // 2. If video capabilities is null, return NotSupported.
+        if (video_capabilities.is_empty())
+            return {};
+
+        // 3. Set the videoCapabilities member of accumulated configuration to video capabilities.
+        accumulated_configuration.video_capabilities = move(video_capabilities);
+    } else {
+        // Otherwise:
```
trflynn89: `Otherwise` goes above the `else` (here and elsewhere)

Libraries/LibWeb/EncryptedMediaExtensions/Algorithms.cpp
```diff
@@ -383,4 +383,24 @@ Optional<ConsentConfiguration> get_supported_configuration(RefPtr<KeySystem> imp
...
+{
+    constexpr Array<StringView, 1> supported_key_systems = {
+        // https://w3c.github.io/encrypted-media/#clear-key
+        "org.w3.clearkey"sv,
+    };
+
+    return supported_key_systems.contains_slow(key_system);
+}
+
+RefPtr<KeySystem> key_system_from_string(String const& key_system)
```
trflynn89: If this only returns a valid `RefPtr`, it should return `NonnullRefPtr`.

Can I ask why this type is ref-counted to begin with? Is it ever actually shared? If not, this could be `NonnullOwnPtr` instead.

Libraries/LibWeb/EncryptedMediaExtensions/NavigatorEncryptedMediaExtensionsPartial.cpp
```diff
@@ -0,0 +1,84 @@
...
+#include <LibWeb/EncryptedMediaExtensions/MediaKeySystemAccess.h>
+#include <LibWeb/EncryptedMediaExtensions/NavigatorEncryptedMediaExtensionsPartial.h>
+#include <LibWeb/HTML/Navigator.h>
+#include <LibWeb/HTML/Scripting/TemporaryExecutionContext.h>
+#include <LibWeb/HTML/Window.h>
+#include <LibWeb/Platform/EventLoopPlugin.h>
+
+namespace Web::EncryptedMediaExtensions {
+
+WebIDL::ExceptionOr<GC::Ref<WebIDL::Promise>> NavigatorEncryptedMediaExtensionsPartial::request_media_key_system_access(String key_system, Vector<Bindings::MediaKeySystemConfiguration> supported_configurations)
```
trflynn89: Same comment here about unnecessary copies

trflynn89: Please add a spec link

Libraries/LibWeb/WebIDL/Types.h
```diff
@@ -51,4 +51,8 @@ using UnsignedLongLong = Infra::Unsigned64BitInteger;
 // double-precision 64-bit IEEE 754 floating point numbers. [IEEE-754]
 using Double = f64;
 
+// https://webidl.spec.whatwg.org/#idl-DOMString
+// The DOMString type corresponds to strings.
+using DOMString = Infra::String;
```
trflynn89: I don't know how much I love this yet. In IDL, we are currently treating `DOMString` as UTF-8. Seems like it would only be confusing to then have it be UTF-16 in C++.

Libraries/LibWeb/EncryptedMediaExtensions/Algorithms.cpp
```diff
@@ -40,7 +40,7 @@ Vector<Bindings::MediaKeySystemMediaCapability> get_supported_capabilities_for_a
             return {};
 
         // 5. Let mimeType be the result of running parse a MIME type with content type.
-        auto mime_type = MimeSniff::MimeType::parse(content_type);
+        auto mime_type = MimeSniff::MimeType::parse(content_type.ascii_view());
```
trflynn89: This is not safe, unless you 100% know for sure that the UTF-16 string is only ever ASCII. Given that this comes from a user script, that is not going to be the case.
```c++
auto mime_type = MimeSniff::MimeType::parse(content_type.to_utf8_but_should_be_ported_to_utf16());
```

Libraries/LibWeb/EncryptedMediaExtensions/Algorithms.cpp
```diff
@@ -120,20 +120,20 @@ Vector<Bindings::MediaKeySystemMediaCapability> get_supported_capabilities_for_a
...
-bool is_persistent_session_type(String const& session_type)
+bool is_persistent_session_type(WebIDL::DOMString const& session_type)
 {
     // 1. Let the session type be the specified MediaKeySessionType value.
 
     // 2. Follow the steps for the value of session type from the following list:
 
     // * "temporary"
-    if (session_type == "temporary"_string) {
+    if (session_type == "temporary"_utf16) {
```
trflynn89: `== "temporary"sv`

Libraries/LibWeb/EncryptedMediaExtensions/Algorithms.cpp
```diff
@@ -384,19 +384,19 @@ Optional<ConsentConfiguration> get_supported_configuration(RefPtr<KeySystem> imp
...
-bool is_supported_key_system(String const& key_system)
+bool is_supported_key_system(WebIDL::DOMString const& key_system)
 {
     constexpr Array<StringView, 1> supported_key_systems = {
         // https://w3c.github.io/encrypted-media/#clear-key
         "org.w3.clearkey"sv,
     };
 
-    return supported_key_systems.contains_slow(key_system);
+    return supported_key_systems.contains_slow(key_system.ascii_view());
```
trflynn89: Also not safe, but can be avoided if the array above is made to hold `Utf16View`.
```c++
constexpr Array<Utf16View, 1> supported_key_systems = {
    // https://w3c.github.io/encrypted-media/#clear-key
    "org.w3.clearkey"sv,
};
```

Libraries/LibWeb/EncryptedMediaExtensions/EncryptedMediaExtensions.idl
```diff
@@ -1,10 +0,0 @@
-// https://w3c.github.io/encrypted-media/#dom-mediakeysystemconfiguration
-dictionary MediaKeySystemConfiguration {
```
trflynn89: This feels like it was squashed into the wrong commit.

Libraries/LibWeb/EncryptedMediaExtensions/MediaKeySystemAccess.idl
```diff
@@ -15,20 +15,20 @@ enum MediaKeysRequirement {
 
 // https://w3c.github.io/encrypted-media/#dom-mediakeysystemconfiguration
 dictionary MediaKeySystemConfiguration {
-    DOMString                               label = "";
-    sequence<DOMString>                     initDataTypes = [];
+    Utf16DOMString                          label = "";
```
trflynn89, line 18: The changes to port to UTF-16 should be squashed into the appropriate commits earlier in the PR.

Libraries/LibWeb/EncryptedMediaExtensions/Algorithms.cpp
```diff
@@ -372,4 +372,24 @@ Optional<ConsentConfiguration> get_supported_configuration_and_consent(KeySystem
...
+
+    // 2. Initialize restrictions to indicate that no configurations have had user consent denied.
+    MediaKeyRestrictions restrictions;
+
+    // 3. Repeat the following step while supported configuration is ConsentDenied:
+    while (supported_configuration.has_value() && supported_configuration->status == ConsentStatus::ConsentDenied) {
+        // 1. Let supported configuration and, if provided, restrictions be the result of executing
+        //    the Get Supported Configuration and Consent algorithm with implementation, candidate configuration, restrictions and origin.
+        supported_configuration = get_supported_configuration_and_consent(implementation, candidate_configuration, restrictions, origin);
+    }
```
gmta, line 396: This loop looks weird to me. If what `get_supported_configuration_and_consent(..)` returns is a configuration with status ConsentDenied, we'll loop here forever.

The only thing that the spec hints at that could change from one iteration to the next, is `restrictions` - but we're not updating that here.

Could you explain how this will not loop forever, or add a safeguard to prevent us from looping here indefinitely?

stelar7: Its not possible in its current implementation to return `ConsentDenied`, as `get_consent_status` always returns `Allowed` for now.

In the future, as more of these FIXMEs get resolved, the restictions would also get updated (its a reference), and then the next iteration would return `NotSupported`

gmta: I don't like that we're depending on the "current implementation" for this loop to halt at some point; that would cause some spooky action at a distance when the implementation is changed in the future.

Why not add a simple safeguard with a maximum number of iterations for example?

gmta (approved):

PR #5667, opened 2025-07-31
LibWeb: Update `border-spacing` to follow the `css-tables` specification

tcl3:
Previously, our definition of `border-spacing` followed an older definition in the CSS2 specification.

I've included a commit, which makes `StyleValueList` only serialize a single value if all values are repeated. This follows the "shortest serialization" principle.

Gets us at least 124 WPT subtest passes that I was able to find.


Libraries/LibWeb/CSS/StyleValues/StyleValueList.cpp
```diff
@@ -31,6 +31,20 @@ String StyleValueList::to_string(SerializationMode mode) const
...
+    auto first_value = m_properties.values.first();
+    auto all_values_equal = true;
+    for (size_t i = 1; i < m_properties.values.size(); ++i) {
+        if (m_properties.values[i] != first_value) {
+            all_values_equal = false;
+            break;
+        }
+    }
+    if (all_values_equal)
+        return first_value->to_string(mode);
```
gmta, line 39: I think this can be simplified to:

```cpp
    auto first_value = m_properties.values.first();
    if (all_of(m_properties.values, [&](auto const& property) { return property == first_value; }))
        return first_value->to_string(mode);
```

Libraries/LibWeb/CSS/StyleValues/StyleValueList.cpp
```diff
@@ -31,6 +31,20 @@ String StyleValueList::to_string(SerializationMode mode) const
         VERIFY_NOT_REACHED();
     }
 
+    if (m_properties.values.is_empty())
+        return {};
```
gmta: This can be moved higher up, to the top of this method.

Libraries/LibWeb/CSS/StyleValues/StyleValueList.cpp
gmta: Or if you'd rather not iterate over the first item twice (which I don't think _really_ matters here?) we could invent something like `AK::all_equal`. We apply this pattern a lot.

gmta (approved):

PR #5669, opened 2025-07-31
LibWeb: Rename PaintContext to DisplayListRecordingContext

kalenikaliaksandr:
PaintContext dates back to a time when display lists didn't exist and it truly represented "paint context". Renaming it to better align with its current role.


Libraries/LibWeb/Painting/DisplayList.cpp
```diff
@@ -10,22 +10,22 @@
 
 namespace Web::Painting {
 
-void DisplayList::append(Command&& command, Optional<i32> scroll_frame_id, RefPtr<ClipFrame const> clip_frame)
+void DisplayList::append(DisplayListItem&& command, Optional<i32> scroll_frame_id, RefPtr<ClipFrame const> clip_frame)
```
gmta: Not sure "item" is an improvement on "command", I think these still actually represent commands to execute? `DisplayListCommand`?

kalenikaliaksandr: if you see DisplayListItem somewhere in painting code, it's a lot easier to tell what it's for rather than trying to figure purpose of Command.

gmta: Sure, but "item" is similar to "data" in that it does not convey any useful information. `DisplayListCommand` would, however.

kalenikaliaksandr: sure, renamed it to DisplayListCommand.

PR #5674, opened 2025-08-01
LibGfx+LibWeb: Delete unused `PaintStyle::paint()` implementations

kalenikaliaksandr:
These are unused since we moved to Skia and it's misleading to keep them around.


gmta:
Fails to build!

PR #5675, opened 2025-08-01
LibWeb: Add indentation to display list dumps

gmta:
Output display list dumps with an indentation level to show balanced commands. It makes it much easier to see what is happening between e.g. PushStackingContext and PopStackingContext, or SaveLayer and Restore.

### DIFF
--- a/Libraries/LibWeb/Painting/DisplayList.cpp
+++ b/Libraries/LibWeb/Painting/DisplayList.cpp
@@ -18,9 +18,28 @@ void DisplayList::append(DisplayListCommand&& command, Optional<i32> scroll_fram
 String DisplayList::dump() const
 {
     StringBuilder builder;
-    for (auto const& command : m_commands) {
-        command.command.visit([&builder](auto const& cmd) { cmd.dump(builder); });
-        builder.appendff("\n");
+    int indentation = 0;
+    for (auto const& command_list_item : m_commands) {
+        auto const& command = command_list_item.command;
+
+        command.visit([&indentation](auto const& command) {
+            if constexpr (requires { command.nesting_level_change; }) {
+                if (command.nesting_level_change < 0 && indentation >= -command.nesting_level_change)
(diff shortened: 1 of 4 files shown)


kalenikaliaksandr (approved):

gmta:
~~Made this a draft because I've missed a couple of commands that call Skia's saveLayer internally.~~

Libraries/LibWeb/Painting/DisplayList.cpp
```diff
@@ -18,9 +18,20 @@ void DisplayList::append(Command&& command, Optional<i32> scroll_frame_id, RefPt
...
+        auto const& command = command_list_item.command;
+        if (indentation > 0 && (command.has<PopStackingContext>() || command.has<Restore>()))
+            --indentation;
+
+        if (indentation > 0)
+            builder.append(MUST(String::repeated("  "_string, indentation)));
+        command.visit([&builder](auto const& cmd) { cmd.dump(builder); });
+        builder.append('\n');
+
+        if (command.has<ApplyCompositeAndBlendingOperator>() || command.has<ApplyFilter>() || command.has<ApplyOpacity>()
```
kalenikaliaksandr: maybe we could add some helper that returns bool indicating whether command increases nesting level and reuse it in DisplayListRecorder?

gmta: Sure!

gmta: I've added a field to all relevant commands. Ideally we could use that same field for `m_save_nesting_level` but that required more messing with `APPEND(..)` than I was comfortable with.

gmta: @kalenikaliaksandr I've also found a way to support this in `APPEND(..)`, added it as a new commit

kalenikaliaksandr (approved):

PR #5678, opened 2025-08-01
LibWeb: Set filename of module scripts to full URL instead of basename

Lubrsi:
Atlassian login gets the base URL for its module scripts by throwing an error and pulling out the current script's URL from error.stack with regex.

Since we only returned a basename for module scripts, it would fail to match and try and use `/` as a base URL (because it does [matched_string] + "/"), which is not a valid base URL.


jamesb3ll:
This code is from Parcel 2 bundler, added in 2019, possibly lots of frontend apps using this:
https://github.com/parcel-bundler/parcel/pull/2534/files#diff-39112519f42223a8e331d0dac27d01f8a70db6fe3d515ad8016856931a6ccbda

Issue #2373, opened 2024-11-16
Mouse events do not work if element is transformed and scrolled off the top of the screen.


Psychpsyo:
### Summary

If an element has a transform that makes it so that it is visible on-screen, despite its original position being **scrolled** completely off the page, it does not receive mouse events.

Manually abspos'ing it into a position like that has it still receive events.

### Operating system

Linux

### Steps to reproduce

1. Open this document:
```html
<!DOCTYPE html>
<style>
  body {
    height: 150vh;
  }
  div {
    background-color: gray;
    height: 10px;
    transform: scaleY(100);
  }
  div:hover {
    background-color: green;
  }
</style>
<div onClick="console.log('it worked!')"></div>
```
3. Scroll down a bit
4. Try to hover or click the div.

### Expected behavior

The div should still receive mouse events.

### Actual behavior

The div does not receive mouse events.

### URL for a reduced test case

See above.

### HTML/SVG/etc. source for a reduced test case

```html
<!DOCTYPE html>
<style>
  body {
    height: 150vh;
  }
  div {
    background-color: gray;
    height: 10px;
    transform: scaleY(100);
  }
  div:hover {
    background-color: green;
  }
</style>
<div onClick="console.log('it worked!')"></div>
```

### Log output and (if possible) backtrace

```shell
None, I think.
```

### Contribute a patch?

- [ ] I’ll contribute a patch for this myself.

PR #5681, opened 2025-08-02
LibWeb: Fix hit testing on CSS transformed elements

### LINKED ISSUE #3727: translation is incorrectly applied to pointer position vs layout
### Summary

When you go to my home website, `https://vaxry.net`, on the top, you have a navbar. The position of the buttons is correct, but you can't hover over them, because the "hover area" is way off to the left.

If you remove `translateX(-50%)` from the css property, the hover area matches the layout.

### Operating system

Linux

### Steps to reproduce

1. Go to `https://vaxry.net`
2. Hover over the buttons on the top (blog, hyprland, github)
3. oh no, nothing happens, gotta go to the left.

### Expected behavior

Layout should match the hover / click areas.

### Actual behavior

Hover / click areas are offset to the left.

https://vaxry.net

```html
<head>
    <title>vaxry.net</title>
</head>

<body>
    <style>
        body {
            background-color: #141111;
            min-height: 100vh;
            min-width: 100vw;
            overflow: hidden;
            margin: 0;
        }

        .navbar {
            display: block;
            position: fixed;
            top: 0;
            left: 0;
            width: 100vw;
            height: auto;
            z-index: 100;
        }

        .navbarLeft {
            display: block;
            position: relative;
            text-align: left;
            height: 4rem;
        }

        .navbarIcon {
            margin-left: 2rem;
            height: 3rem;
            border: 0.15rem solid #A6A6A6;
            border-radius: 50%;
        }

        .navbarMiddle {
            position: absolute;
            top: 0.5rem;
            left: 50%;
            transform: translateX(-50%);
        }

        .navbarButton {
            font-family: "Hack";
            font-size: 0.9rem;
            color: #A6A6A6;
            text-decoration: none;
            margin-left: 5rem;
            margin-right: 5rem;
            transition: 0.2s ease-in-out;
        }

        .navbarButton:hover {
            color: #da9385;
        }
    </style>
    <div class="navbar">
        <div class="navbarMiddle">
            <a href="https://blog.vaxry.net" class="navbarButton">blog</a>
            <a href="https://github.com/vaxerski" class="navbarButton">github</a>
            <a href="https://github.com/hyprwm/hyprland" class="navbarButton">hyprland</a>
        </div>
    </div>
</body>
```

```shell
No logs
```

_No response_

_No response_

### Contribute a patch?

zacoons:
Fixes #3727

### DIFF
--- a/Libraries/LibWeb/Painting/PaintableBox.cpp
+++ b/Libraries/LibWeb/Painting/PaintableBox.cpp
@@ -1093,17 +1093,20 @@ CSSPixelPoint PaintableBox::adjust_position_for_cumulative_scroll_offset(CSSPixe
     return position.translated(-cumulative_offset_of_enclosing_scroll_frame());
 }
 
-TraversalDecision PaintableBox::hit_test(CSSPixelPoint position, HitTestType type, Function<TraversalDecision(HitTestResult)> const& callback) const
+TraversalDecision PaintableBox::hit_test(CSSPixelPoint raw_position, HitTestType type, Function<TraversalDecision(HitTestResult)> const& callback) const
 {
-    if (clip_rect_for_hit_testing().has_value() && !clip_rect_for_hit_testing()->contains(position))
+    if (clip_rect_for_hit_testing().has_value() && !clip_rect_for_hit_testing()->contains(raw_position))
         return TraversalDecision::Continue;
-
-    auto position_adjusted_by_scroll_offset = adjust_position_for_cumulative_scroll_offset(position);
 
     if (computed_values().visibility() != CSS::Visibility::Visible)
         return TraversalDecision::Continue;
 
-    if (hit_test_scrollbars(position_adjusted_by_scroll_offset, callback) == TraversalDecision::Break)
+    auto position_adjusted_by_scroll_offset = adjust_position_for_cumulative_scroll_offset(raw_position);
+    // NOTE: This CSSPixels -> Float -> CSSPixels conversion is because we can't AffineTransform::map() a CSSPixelPoint.
+    auto offset_position = position_adjusted_by_scroll_offset.translated(-transform_origin()).to_type<float>();
+    auto position = Gfx::extract_2d_affine_transform(transform()).inverse().value_or({}).map(offset_position).to_type<CSSPixels>() + transform_origin();
+
+    if (hit_test_scrollbars(position, callback) == TraversalDecision::Break)
         return TraversalDecision::Break;
 
     if (is_viewport()) {
@@ -1120,7 +1123,7 @@ TraversalDecision PaintableBox::hit_test(CSSPixelPoint position, HitTestType typ
     if (!visible_for_hit_testing())
         return TraversalDecision::Continue;
 
-    if (!absolute_border_box_rect().contains(position_adjusted_by_scroll_offset))
+    if (!absolute_border_box_rect().contains(position))
         return TraversalDecision::Continue;
 
     if (hit_test_continuation(callback) == TraversalDecision::Break)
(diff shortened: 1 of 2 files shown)


Psychpsyo:
Question: Does this also fix https://github.com/LadybirdBrowser/ladybird/issues/2373?

tcl3:
Not a review, just dropping in to say I believe the relevant WPT test for this is: http://wpt.live/css/css-transforms/transform-hit-testing.html. You should be able to import it by running `./Meta/WPT.sh import css/css-transforms/transform-hit-testing.html`.

Libraries/LibWeb/Painting/StackingContext.cpp
```diff
@@ -384,7 +384,7 @@ void StackingContext::paint(DisplayListRecordingContext& context) const
 
 TraversalDecision StackingContext::hit_test(CSSPixelPoint raw_position, HitTestType type, Function<TraversalDecision(HitTestResult)> const& callback) const
 {
-    if (!paintable_box().is_visible())
+    if (!paintable_box().visible_for_hit_testing())
```
gmta, line 387: Could we have a test for this? I.e. hit testing a z-indexed opacity: 0 \<div\> for example.

Libraries/LibWeb/Painting/PaintableBox.cpp
gmta, line 1: Commit title: please avoid using weasel-words such as "fix", you could say something like 'Prevent hit testing from transforming position more than once' or something similar.

Libraries/LibWeb/Painting/PaintableBox.cpp
```diff
@@ -1162,17 +1162,20 @@ CSSPixelPoint PaintableBox::adjust_position_for_cumulative_scroll_offset(CSSPixe
     return position.translated(-cumulative_offset_of_enclosing_scroll_frame());
 }
 
-TraversalDecision PaintableBox::hit_test(CSSPixelPoint position, HitTestType type, Function<TraversalDecision(HitTestResult)> const& callback) const
+TraversalDecision PaintableBox::hit_test(CSSPixelPoint raw_position, HitTestType type, Function<TraversalDecision(HitTestResult)> const& callback) const
```
gmta: The original `position` and `transformed_position` terminology was clearer to me than using `raw_position` and `position`. Because you renamed these variables in the same commit as the change to StackingContext where you prevent passing in the transformed position, it's pretty difficult to discern what exactly is the fix and what has to do with the renames.

Could you either
1) drop the renames and only include the fix to the doubly transformed position, or
2) put the renames in their own separate commit?

Libraries/LibWeb/Painting/PaintableBox.cpp
zacoons: Sure, sorry about that. I thought it would be fine since I describe the changes in the commit description.

Libraries/LibWeb/Painting/StackingContext.cpp
zacoons: There is already in a test: `Text/input/wpt-import/css/filter-effects/css-filters-opacity-hit-testing.html`.
This change fixes that test, which was failing after the first commit of this branch.

Libraries/LibWeb/Painting/PaintableBox.cpp
zacoons: I changed it because `raw_position` shouldn't really be used anywhere. Since `position` is used in most places it's a lot shorter than writing `transformed_position` and results in shorter (and more readable) lines. Would you still prefer it if I reverted the rename?

Libraries/LibWeb/Painting/StackingContext.cpp
gmta: Ah right, could this change be a part of that commit then, to make things easily revertible in the future?

Libraries/LibWeb/Painting/PaintableBox.cpp
gmta: Using `position` is fine!

Libraries/LibWeb/Painting/StackingContext.cpp
zacoons: I didn't do this in order to avoid doing multiple things in one commit. FYI, the second commit also fixes a test that was broken in the first commit. Do you just want me to squish all three commits together and put all the details in the commit message?

gmta: If you break tests in a commit, you usually should include the fixes for them in the same commit, yeah.

zacoons: I've put everything in one commit with a detailed description. How do you like it?

Libraries/LibWeb/Painting/PaintableBox.cpp
zacoons: I decided not to do a rename because it seems like a waste of a commit just to do this rename, which doesn't really have that much benefit. Thanks for your patience :)

Libraries/LibWeb/Painting/StackingContext.cpp
gmta: Yeah that seems fine! Thanks

gmta (approved):

PR #5682, opened 2025-08-02
LibWeb/Crypto: Spec updates

devgianlu:
There have been two spec updates that affect our WPT tests:
- https://github.com/WICG/webcrypto-secure-curves/pull/34
- https://github.com/WICG/webcrypto-secure-curves/pull/35

There are still some WPT tests that we fail because, in my opinion, the spec doesn't match WPT: https://github.com/WICG/webcrypto-secure-curves/issues/33#issuecomment-3146435568


Libraries/LibWeb/Crypto/CryptoAlgorithms.cpp
```diff
@@ -6050,19 +6050,26 @@ WebIDL::ExceptionOr<GC::Ref<CryptoKey>> ED25519::import_key(
             }
         }
 
-        // 2. Let algorithm be a new KeyAlgorithm object.
+        // 2. Let data be keyData.
+        auto data = key_data.get<ByteBuffer>();
```
trflynn89: `auto data = ` in each of of these is creating a copy of the `ByteBuffer`:

```c++
auto const& data = key_data.get<ByteBuffer>();
```

Libraries/LibWeb/Crypto/CryptoAlgorithms.cpp
```diff
@@ -6050,19 +6050,26 @@ WebIDL::ExceptionOr<GC::Ref<CryptoKey>> ED25519::import_key(
...
         auto algorithm = KeyAlgorithm::create(m_realm);
 
-        // 3. Set the name attribute of algorithm to "Ed25519".
+        // 5. Set the name attribute of algorithm to "Ed25519".
         algorithm->set_name("Ed25519"_string);
 
-        // 4. Let key be a new CryptoKey associated with the relevant global object of this [HTML], and representing the key data provided in keyData.
-        key = CryptoKey::create(m_realm, key_data);
+        // 6. Let key be a new CryptoKey associated with the relevant global object of this [HTML], and that represents data.
+        key = CryptoKey::create(m_realm, data);
```
trflynn89: Hmm ok, if the goal is to pass the byte buffer to `CryptoKey::create` then we should be doing this above:
```c++
auto data = move(key_data.get<ByteBuffer>());
```

And this here:
```c++
key = CryptoKey::create(m_realm, move(data));
```

devgianlu: There's ~10 occurrences of this pattern just in this file, should I fix all of them or leave it for further refactoring?

trflynn89: Up to you, I would not mind another commit to fix them here. But at least the ones that were changed in this commit should be fixed here, because otherwise we're now going from making 1 copy of the buffer to 2 copies.

PR #5685, opened 2025-08-02
LibWeb: Replace DrawTriangleWave with StrokePathUsingColor

kalenikaliaksandr:
There's no need to have separate display list item for drawing triangle wave when we could simply use StrokePathUsingColor. By switching to StrokePathUsingColor we could also reduce painting because it supports filtering out by bounding box.


gmta:
Builds are failing

gmta (approved):

PR #5691, opened 2025-08-03
LibWeb: Parse `anchor()` function for inset properties

tcl3:
This adds `anchor(...)` parsing for the `top`, `bottom`, `left` and `right` properties. Currently, we don't do anything useful with the value after parsing.

Adds 2353 subtest passes in `css/css-anchor-position`.


Libraries/LibWeb/CSS/Properties.json
```diff
@@ -1123,7 +1123,8 @@
     "initial": "auto",
     "valid-types": [
       "length [-∞,∞]",
-      "percentage [-∞,∞]"
+      "percentage [-∞,∞]",
+      "anchor"
```
gmta, line 1127: The spec says that `anchor()` is invalid outside of the "inset properties", which is not only top/right/bottom/left but also `inset-block(-start/end)`, `inset-inline(-start/end)` and `inset`.

tcl3: My understanding is that these logical aliases end up getting mapped to a one of the top/bottom/left/right physical properties, so we should be able to use `anchor()` with them as-is.

Tested with the following, which shows that the property does get set:
```html
<!DOCTYPE html>
<body></body>
<script>
    const element = document.createElement("div");
    const property =  "inset-inline-start";
    const value = "anchor(inside)";
    element.style[property] = value;
    console.log(`Specified value for ${property}: "${element.style[property]}"`);
</script>
```

gmta: Oh that's sweet! I could've used that with `anchor-size()` as well, instead of manually checking the property context 🤔

PR #5694, opened 2025-08-03
LibWeb/HTML: Avoid Storage unexpected broadcast events

adamantike:
* Fixes `Storage::set_item` to not broadcast events when setting a key to the same value.
* Fixes `Storage::clear` to not broadcast events when the storage is already empty.


Libraries/LibWeb/HTML/Storage.cpp
```diff
@@ -141,6 +158,13 @@ void Storage::remove_item(String const& key)
 // https://html.spec.whatwg.org/multipage/webstorage.html#dom-storage-clear
 void Storage::clear()
 {
+    // NOTE: Spec is not explicit about what to do if the storage is already empty, but
+    //       `StorageEvent` description implies it must only be fired when storage areas change.
+    //       This matches the behavior of `set_item` and `remove_item` methods.
+    //       See: https://html.spec.whatwg.org/multipage/indices.html#event-storage
```
konradekk, line 164: wpt tests support that – probably this should be noted in the comment instead.

adamantike: Which comment are you referring to?

adamantike:
The CI run seems to intermittently fail on this new `event_no_duplicates.html` test.

I haven't been able to reproduce it locally, but I wonder if this is because of the CI workers running tests in parallel, and the test validating that a `storage.clear()` call does not broadcast when the storage is empty actually gets a non-empty local storage.

adamantike:
This is still relevant, waiting for feedback.

gmta (approved):

Libraries/LibWeb/HTML/Storage.cpp
gmta: I don't agree with @konradekk here, WPT tests are subject to changes all the time. The best source is "what do other browsers do", after that, the spec. In this case, both align and you've already mentioned that the spec says "only fire if storage areas change", so that seems good enough to me.

gmta:
On macOS, I can reproduce the flake often by running:

```
% Meta/ladybird.sh run test-web -f Text/input/wpt-import/webstorage
...
Test failed: file:///path/to/ladybird/Tests/LibWeb/Text/input/wpt-import/webstorage/event_no_duplicates.html
--- /path/to/ladybird/Tests/LibWeb/Text/expected/./wpt-import/webstorage/event_no_duplicates.txt
+++ /path/to/ladybird/Tests/LibWeb/Text/expected/./wpt-import/webstorage/event_no_duplicates.txt
@@ -2,7 +2,8 @@

 Found 8 tests

-8 Pass
+7 Pass
+1 Fail
 Pass	sessionStorage exists
 Pass	Setting to same value does not trigger event for sessionStorage
 Pass	Deleting non-existent key does not trigger event for sessionStorage
@@ -10,4 +11,4 @@
 Pass	localStorage exists
 Pass	Setting to same value does not trigger event for localStorage
 Pass	Deleting non-existent key does not trigger event for localStorage
-Pass	Clearing empty storage does not trigger event for localStorage
+Fail	Clearing empty storage does not trigger event for localStorage
```

Libraries/LibWeb/HTML/Storage.cpp
konradekk: what you’ve noted about wpt tests could be also said about other implementations and the spec itself – all of these are really closely related and changing all the time. (not sure which of them should be considered most conclusive, but i’d probably vote for the tests, then the spec, finishing with other browsers.)

yet it seems there’s a good enough™ invoked from the maintainer and i have nothing else to add here, so please proceed! 🙂‍↕️

adamantike:
Thanks for validating the test run. I was able to easily reproduce the test fail with your command, but then adding `-j 1` makes it always pass. As I was suspecting, it seems `Storage` is not isolated per job, so calls to `clear()` are flaky when running tests concurrently.

If there's someone else with context on how the test runner is set, and where we could instantiate a separate `Storage` per worker, that would be useful to avoid marking these tests as flaky for the time being.

adamantike:
Gentle ping to @trflynn89, who implemented test concurrency, in case this is a known limitation. Ideally, each worker should have an isolated `Storage` or storage namespace, so they don't affect each other.

I also found open issue https://github.com/whatwg/html/issues/3099, indicating that behavior for `file://` document origin is not explicitly defined. If localStorage were scoped per file, this would be a non-issue, as each file would act as its own namespace and storage wouldn't collide between WPT test files. It's not clear, however, if there's going to be any alignment on how to resolve that in the near future.

adamantike:
I have rebased, and marked the new test as flaky because of the concurrency issue affecting the `Storage` singleton used within `test-web`.

Tests/LibWeb/TestConfig.ini
```diff
@@ -351,3 +351,6 @@ Text/input/wpt-import/html/semantics/popovers/popover-toggle-source.tentative.ht
 
 ; Flaky: https://github.com/LadybirdBrowser/ladybird/issues/5257
 Text/input/selection-over-multiple-code-units.html
+
+; Flaky on concurrent runs, see #5694
+Text/input/wpt-import/webstorage/event_no_duplicates.html
```
gmta, line 356: Commit title: the category for this should probably be "Tests: [..]"

PR #5696, opened 2025-08-04
CI: Enable compiler cache for Windows

ayeteadoe:
Addresses [this message](https://discord.com/channels/1247070541085671459/1306918361732616212/1392426076512190565) in ui-windows

**NOTE:** I will be adding a test/throwaway commit to confirm if Windows CI uses the cache. Once that is confirmed I will rebase without that commit


.github/actions/cache-restore/action.yml
```diff
@@ -78,6 +78,22 @@ runs:
           CCACHE_DIR=${{ inputs.ccache_path }} ccache -s
           CCACHE_DIR=${{ inputs.ccache_path }} ccache -z
 
+    - name: 'Configure Compiler Cache (Windows)'
+      if: ${{ inputs.os == 'Windows' && inputs.ccache_path != '' }}
+      shell: pwsh
+      run: |
+        $env:CCACHE_DIR = "${{ inputs.ccache_path }}"
```
gmta: Couldn't we use the [`env:` key](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idenv) for this step to do so? And the previous step could do with a similar approach.

.github/actions/cache-save/action.yml
```diff
@@ -45,10 +56,17 @@ runs:
...
+      if: ${{ inputs.os != 'Windows' && inputs.ccache_path != '' }}
       run: |
         CCACHE_DIR=${{ inputs.ccache_path }} ccache -s
 
+    - name: 'Cache Stats (Windows)'
+      shell: pwsh
+      if: ${{ inputs.os == 'Windows' && inputs.ccache_path != '' }}
+      run: |
+        $env:CCACHE_DIR = "${{ inputs.ccache_path }}"
+        ccache -s
```
gmta, line 54: I think you can unify these steps if we use the `env:` key here as well.

.github/actions/cache-save/action.yml
```diff
@@ -6,6 +6,10 @@ inputs:
   runner_labels:
     description: 'Runner selection labels'
     required: true
+  os:
+    description: 'Operating System to restore caches for'
+    required: true
+    default: 'Linux'
```
gmta: You no longer need this input variable

PR #5700, opened 2025-08-04
LibWeb/Layout: Implement table-row collapsing

AtkinsSJ:
A table-row or table-row-group with `visibility: collapse` should take up no vertical space.

Progress on https://www.w3.org/Style/CSS/all-properties.en.html :

## Before

image

## After

image

One of the imported tests still fails. If a table has a fixed height, and an auto-sized row is then collapsed, the table's height should be reduced by that much. I'm a bit lost though so I'm keeping this PR limited to the simpler cases. :sweat_smile:


Loirooriol:
Does Ladybird support table column collapsing?

AtkinsSJ:
> Does Ladybird support table column collapsing?

No. I don't think we do much for columns at all, but I might be wrong on that.

PR #5702, opened 2025-08-04
LibWeb: Add initial support for bitmap cropping to createImageBitmap

IdanHo:
Fixes 10 WPT tests.


Libraries/LibGfx/Bitmap.cpp
```diff
@@ -160,6 +160,10 @@ void Bitmap::apply_mask(Gfx::Bitmap const& mask, MaskKind mask_kind)
 
 ErrorOr<NonnullRefPtr<Gfx::Bitmap>> Bitmap::cropped(Gfx::IntRect crop, Gfx::Color outside_color) const
 {
+    // OPTIMIZATION: Skip slow manual copying for NO-OP crops
+    if (crop == rect())
+        return clone();
```
gmta, line 165: While this makes sense, I suspect this method is mainly slow because of the format limitation you removed in a previous commit: we needed to have pixel-to-pixel conversion happening.

Now you have an output bitmap that always has the same format as the input bitmap, you should be able to fill a bitmap with the outside color (if larger than `rect()`) and do `TypedTransfer<T>::copy` operations on partial or whole scanlines instead.

IdanHo: I agree, I added the clone optimization because the new caller calls it with the default rect in the common case, and I felt bad having to run through this code each time.
Do you feel like rewriting the rest should be part of this PR as well?

gmta: No, doesn't need to be part of this PR :) let's pick that up elsewhere.

Libraries/LibWeb/HTML/WindowOrWorkerGlobalScope.cpp
```diff
@@ -126,7 +126,86 @@ GC::Ref<WebIDL::Promise> WindowOrWorkerGlobalScopeMixin::create_image_bitmap(Ima
...
+        // the height of sourceRectangle, times the value of the resizeWidth member of options, divided by the width of sourceRectangle, rounded up to the nearest integer
+        output_height = ceil_div(source_rectangle.height() * options->resize_width.value(), source_rectangle.width());
+    }
+    // -> If neither resizeWidth nor resizeHeight are specified
+    else {
+        // the height of sourceRectangle
+        output_height = source_rectangle.height();
+    }
+
+    // 5. Place input on an infinite transparent black grid plane, positioned so that its top left corner is at the origin of the plane, with the x-coordinate increasing to the right, and the y-coordinate increasing down, and with each pixel in the input image data occupying a cell on the plane's grid.
```
gmta: Could you word wrap some of these steps [to 120 characters](https://github.com/LadybirdBrowser/ladybird/blob/master/Documentation/CodingStyle.md#comments)?

IdanHo: definitely, fixed!

gmta (approved):

PR #5706, opened 2025-08-04
LibWeb/Painting: Implement missing border-styles

AtkinsSJ:
We had 3 unimplemented styles: `double`, `groove`, and `ridge`. (We had a FIXME on `inset` and `outset` too, but they're actually implemented already.)

I've implemented these by splitting them into 2 borders, an inner and outer one, and recursively calling `paint_border()` to paint them. Generally that works pretty well, but in some cases the border-radii can cause issues, which I've not been smart enough today to figure out. :sweat_smile:

There's also the awkwardness that we do this once we've already converted everything to DevicePixels, and then have to return to floats to get decent values. That could definitely be improved.

For screenshots, see the test ref images. :^)


AtkinsSJ:
Changes: Rebaselined `Tests/LibWeb/Screenshot/input/css-background-clip-text.html` because it turns out that uses `border-style: double`.

AtkinsSJ:
Changes: Rebaselined, and added/updated fuzzy matching values to the tests so they pass on macOS.

PR #5707, opened 2025-08-04
LibWeb: Remove extraneous trailing '/' from service worker script scope

IdanHo:
The spec language specifies 'Set maxScopeString to "/", followed by the strings in XXXX’s path (including empty strings), separated from each other by "/"': That is, adjacent components are separated by a '/', but the last component does not get a trailing '/'.

This resulted in the generated scope string ending with '//' in some cases, incorrectly tripping the 'Service-Worker-Allowed' security check.

Found this while investigating some unrelated WPT failures.

### DIFF
--- a/Libraries/LibWeb/ServiceWorker/Job.cpp
+++ b/Libraries/LibWeb/ServiceWorker/Job.cpp
@@ -289,10 +289,7 @@ static void update(JS::VM& vm, GC::Ref<Job> job)
             auto join_paths_with_slash = [](URL::URL const& url) -> ByteString {
                 StringBuilder builder;
                 builder.append('/');
-                for (auto const& component : url.paths()) {
-                    builder.append(component);
-                    builder.append('/');
-                }
+                builder.join('/', url.paths());
                 return builder.to_byte_string();
             };
 


Libraries/LibWeb/ServiceWorker/Job.cpp
```diff
@@ -288,10 +288,11 @@ static void update(JS::VM& vm, GC::Ref<Job> job)
...
             auto join_paths_with_slash = [](URL::URL const& url) -> ByteString {
                 StringBuilder builder;
-                builder.append('/');
+                if (url.paths().size() == 0)
+                    builder.append('/');
                 for (auto const& component : url.paths()) {
-                    builder.append(component);
                     builder.append('/');
+                    builder.append(component);
                 }
```
gmta: Couldn't this be simplified to:

```cpp
StringBuilder builder;
builder.append('/');
builder.join('/', url.paths());
return builder.to_byte_string();
```

IdanHo: Yep, that's much nicer, updated

PR #5708, opened 2025-08-04
LibCrypto: Add missing test buffer initialization in roundtrip testcase

IdanHo:
target_buffer was being initialized using UnsignedBigInteger::export_data, which does not necessarily overwrite the whole given buffer, leading to flaky test failures.


Tests/LibCrypto/TestBigInteger.cpp
```diff
@@ -212,7 +212,7 @@ TEST_CASE(test_unsigned_bigint_base10_to_string)
 TEST_CASE(test_bigint_import_big_endian_decode_encode_roundtrip)
 {
     u8 random_bytes[128];
-    u8 target_buffer[128];
+    u8 target_buffer[128] {};
     fill_with_random(random_bytes);
     auto encoded = Crypto::UnsignedBigInteger::import_data(random_bytes, 128);
     (void)encoded.export_data({ target_buffer, 128 });
```
gmta: This seems like a pretty big footgun. Instead of exposing libttommath `mp_to_ubin`'s  implementation detail like this, maybe `UnsignedBigInteger::export_data` could inspect the number of bytes written and set the remaining bytes to zeroes?

gmta: This can then be complemented by a test that specifically checks the output buffer for null bytes.

Tests/LibCrypto/TestBigInteger.cpp
```diff
@@ -212,7 +212,7 @@ TEST_CASE(test_unsigned_bigint_base10_to_string)
 TEST_CASE(test_bigint_import_big_endian_decode_encode_roundtrip)
 {
     u8 random_bytes[128];
-    u8 target_buffer[128];
+    u8 target_buffer[128] {};
```
trflynn89: Hmm are we sure this fixes the flake? One way I was able to repro the flake locally was when `random_bytes[0]` was randomly 0. So if I force that: 

```diff
diff --git a/Tests/LibCrypto/TestBigInteger.cpp b/Tests/LibCrypto/TestBigInteger.cpp
index aeab0d8c474..73c2e2f98d3 100644
--- a/Tests/LibCrypto/TestBigInteger.cpp
+++ b/Tests/LibCrypto/TestBigInteger.cpp
@@ -212,10 +212,16 @@ TEST_CASE(test_unsigned_bigint_base10_to_string)
 TEST_CASE(test_bigint_import_big_endian_decode_encode_roundtrip)
 {
     u8 random_bytes[128];
-    u8 target_buffer[128];
+    u8 target_buffer[128] {};
     fill_with_random(random_bytes);
+    random_bytes[0] = 0;
+
     auto encoded = Crypto::UnsignedBigInteger::import_data(random_bytes, 128);
     (void)encoded.export_data({ target_buffer, 128 });
+
+    dbgln(" random_bytes={}", Bytes { random_bytes });
+    dbgln("target_buffer={}", Bytes { target_buffer });
+
     EXPECT(memcmp(target_buffer, random_bytes, 128) == 0);
 }
```

We still fail the test, at least on my machine. This logs:
```
 random_bytes=00b99565ca4d6c173bf869712226d5cf59234c9f6a34d388d9ba54d7b0c71df814ce769a3ecf152d356905ad4dc13126636ee2ec3b8253e2646e26da305f935a973a429ebb924162cba8b2614602e931ebdd542bfef60e544b0b3ff079f64d928e98b78baa6dc8d22ad002eb1171b8e05fbaa827b3108c5af83123e0322155cc
target_buffer=b99565ca4d6c173bf869712226d5cf59234c9f6a34d388d9ba54d7b0c71df814ce769a3ecf152d356905ad4dc13126636ee2ec3b8253e2646e26da305f935a973a429ebb924162cba8b2614602e931ebdd542bfef60e544b0b3ff079f64d928e98b78baa6dc8d22ad002eb1171b8e05fbaa827b3108c5af83123e0322155cc00
FAIL: /home/flynn/workspace/ladybird/Tests/LibCrypto/TestBigInteger.cpp:224: EXPECT(memcmp(target_buffer, random_bytes, 128) == 0) failed
```

So in the target buffer, the 0s are on the wrong end (so to speak)

IdanHo: Yep, this is a second issue - my fix only resolved some of cases (in hindsight just checking for an all zeros random_bytes was not good enough)
Looks like the issue here is that the original big integer implementation preserved leading zeros, and libtommath doesn't.

Tests/LibCrypto/TestBigInteger.cpp
IdanHo: sure.

Tests/LibCrypto/TestBigInteger.cpp
IdanHo: I can't think of a very nice way of fixing this aside from manually skipping any leading zeros in the input data when comparing.

gmta: You could make `::export_data()` just move the bytes to the right `max_bytes - written` positions with `TypedTransfer<T>::move(..)`?

IdanHo: I mean I could, but that wouldn't make export_data "more correct", it would just be overfitting to this test. That is, I'm not sure the right adjusting the encoded number needs to be part of the export_data API.

gmta: I agree with you there, it does seem `::export_data()` requires its callers to handle the amount of bytes written.

So maybe the test should take the return value of `::export_data()` and do a memcmp between`target_buffer` and `random_bytes + (max_bytes - bytes_written)` for `bytes_written` bytes?

trflynn89: I'm thinking the interface here is a bit C-like and could be made more like `AK::Stream` - i.e. it should just return the bytes that were filled in:
```c++
Bytes export_data(Bytes) const;
```

Whether the method filled in the most- or least-significant bytes shouldn't really matter to any caller (the implementation would have to know though, of course). In this case, the caller would just receive a 127 byte span.

As far as dealing with `random_bytes` goes, that becomes a test-only concern IMO. If we randomly generated the bytes `00 00 00 12`, we don't need to compare those 3 leading 00s to any output bytes - we should just take the span after any leading zeros. We could even use `Bytes::ends_with` here I think?

Libraries/LibCrypto/ASN1/DER.cpp
```diff
@@ -305,32 +305,21 @@ ErrorOr<void> Encoder::write_arbitrary_sized_integer(UnsignedBigInteger const& v
     auto max_byte_size = max(1ull, value.byte_length()); // At minimum, we need one byte to encode 0.
     ByteBuffer buffer;
     auto output = TRY(buffer.get_bytes_for_writing(max_byte_size));
-    auto size = value.export_data(output);
+    auto result = value.export_data(output);
     // DER does not allow empty integers, encode a zero if the exported size is zero.
-    if (size == 0) {
+    if (result.size() == 0) {
```
alimpfard: Nit: now that we have spans, could this be `result.is_empty()`?

Libraries/LibWeb/Crypto/CryptoAlgorithms.cpp
```diff
@@ -88,20 +88,8 @@ static ::Crypto::UnsignedBigInteger big_integer_from_api_big_integer(GC::Ptr<JS:
...
-                reversed_buffer[buffer.size() - i - 1] = buffer[i];
-            }
-
-            return ::Crypto::UnsignedBigInteger::import_data(reversed_buffer.data(), reversed_buffer.size());
-        } else {
-            return ::Crypto::UnsignedBigInteger::import_data(buffer.data(), buffer.size());
-        }
-    }
+    if (buffer.size() > 0)
+        return ::Crypto::UnsignedBigInteger::import_data(buffer);
```
alimpfard, line 92: 🤔 dropping the LE case?

Libraries/LibCrypto/ASN1/DER.cpp
IdanHo: fixed

Libraries/LibWeb/Crypto/CryptoAlgorithms.cpp
IdanHo: Yes, similarly to export_data, libtommath assumes the input is always big-endian, so it's incorrect to byte-swap it before importing on LE

PR #5715, opened 2025-08-05
LibWeb: Support percentages in `word-spacing`

Calme1709:
Fixes crash in the created test as well as https://wpt.live/css/css-text/word-spacing/reference/word-spacing-percent-001-ref.html. The WPT test hasn't been imported as it passing is currently a false-positive due to the fact that we don't yet respect `word-spacing` in most cases.


Libraries/LibWeb/CSS/ComputedValues.h
```diff
@@ -451,7 +451,7 @@ class ComputedValues {
     CSS::Positioning position() const { return m_noninherited.position; }
     CSS::WhiteSpaceCollapse white_space_collapse() const { return m_inherited.white_space_collapse; }
     WhiteSpaceTrimData white_space_trim() const { return m_noninherited.white_space_trim; }
-    CSS::LengthOrCalculated word_spacing() const { return m_inherited.word_spacing; }
+    LengthPercentage word_spacing() const { return m_inherited.word_spacing; }
```
tcl3: Could we return this by const reference to avoid the copy?

tcl3 (approved):

PR #5727, opened 2025-08-05
LibIDL: Parse exteded attributes before `required` keyword

AronHetLam:
According to spec Extended attributes are allowed both before and after the `required` keyword in WebIDL.
The linked spec below can be expanded to `ExtendedAttributeList required TypeWithExtendedAttributes identifier ;`
https://webidl.spec.whatwg.org/#prod-DictionaryMember


Libraries/LibIDL/IDLParser.cpp
```diff
@@ -955,13 +955,16 @@ void Parser::parse_dictionary(Interface& interface)
...
+            extended_attributes = parse_extended_attributes();
+
         if (lexer.consume_specific("required"sv)) {
             required = true;
             consume_whitespace();
         }
 
         if (lexer.consume_specific('['))
-            extended_attributes = parse_extended_attributes();
+            extended_attributes.update(parse_extended_attributes());
```
gmta, line 967: Typo in commit title: "exteded"

gmta (commented):
Thank you for looking at this!

AtkinsSJ (approved):
Looks good!

PR #5728, opened 2025-08-05
CodeGenerators: Use is_optional for integer dictionary member wrapper

AronHetLam:
Optional integral members in WebIDL dictionaries where generated with accesser that did not take into account that it was optional.
This PR solves this by using the already calculated `is_optional` flag together with `type.is_nullable`


AronHetLam:
This solves the same issue as #5724

trflynn89:
I've merged #5724. There are several other conditions/branches involved before `generate_from_integral` is called which affect whether the generated value is actually optional, which that PR is handling.

PR #5732, opened 2025-08-06
Base: Wrap error URL

### LINKED ISSUE #5730: URL on crash page extends beyond the edge of the screen
### Summary

Title

### Operating system

Linux

### Steps to reproduce

1. Find a situation that crashes the page with an abnormally long URL (in my case Steam Authentication attempt)

### Expected behavior

Text field should have some mechanism to wrap around and prevent the text from extending offscreen

### Actual behavior

Text extends beyond the edges of the screen

Steam authentication process (or https://steamcommunity.com/openid/login)

```html
N/A
```

```shell
N/A
```

Image

_No response_

### Contribute a patch?

ChaseKnowlden:
Fixes text overlap on error page.

Fix #5730

### DIFF
--- a/Base/res/ladybird/templates/error.html
+++ b/Base/res/ladybird/templates/error.html
@@ -37,6 +37,7 @@
         h1 {
             margin: 0;
             font-size: 1.5rem;
+            text-wrap: wrap;
         }
         p {
             font-size: 1rem;


Base/res/ladybird/templates/error.html
```diff
@@ -1,57 +1,62 @@
...
+                    d="M11.75.75h-9c-1.1 0-2 .9-2 2v16c0 1.1.9 2 2 2h12c1.1 0 2-.9 2-2v-13l-5-5z"
+                />
+                <path
+                    d="M10.75.75v4c0 1.1.9 2 2 2h4M5.75 9.75v2M11.75 9.75v2M5.75 16.75c1-2.67 5-2.67 6 0"
+                />
+            </svg>
+            <h1>Failed to load %failed_url%</h1>
+        </header>
+        <p>%error_message%</p>
+    </body>
```
gmta: You've reformatted this page which makes the diff hard to read. In my opinion, not indenting `<head>` and `<body>` is fine, and it's already commonplace in many of the `.html` files in the tree. Either don't reformat this page, or split it out to two separate commits please.

ChaseKnowlden: Fixed.

Base/res/ladybird/templates/error.html
```diff
@@ -37,6 +37,7 @@
         h1 {
             margin: 0;
             font-size: 1.5rem;
+            text-wrap: wrap;
```
gmta, line 40: Isn't `text-wrap: wrap` already the default? Maybe `overflow-wrap: break-word;` could help but I don't think Ladybird supports that property yet.

Issue #5683, opened 2025-08-02
LibJS: Assignment to const local is allowed in loop update expression


awesomekling:
We should throw on `++i` here but we don't:

```js
for (const i = 0; i < 3; ++i) {
    console.log(i);
}
```

trflynn89:
Just a couple notes:

We do throw on:
```js
const i = 0;
++i;
```

But not if we're in a non-global scope, i.e. this does not throw:
```js
(() => {
    const i = 0;
    ++i;
})();
```

The bytecode for the global `++i` is:
```
[   0]    0: InitializeLexicalBinding i, src:Int32(0)
[  18]       GetGlobal dst:reg5, i
[  30]       Increment dst:reg5
[  40]       SetGlobal i, src:reg5
[  58]       End value:reg5
```

The bytecode for the local `++i` is:
```
JS bytecode executable ""
[   0]    0: Mov dst:i~0, src:Int32(0)
[  18]       Increment dst:i~0
[  28]       End value:Undefined
```

Lubrsi:
Note that side effects complicate the implementation of this. For example, Increment still needs to run arbitrary valueOf functions, but only throw when it tries to set the local at the end.
```js
(() => {
    const i = {
        valueOf() {
            console.log("hello!");
            return 1;
        }
    };
    i++;
    return i;
})();
```

pwespi:
I have a proposal for a fix for this: https://github.com/LadybirdBrowser/ladybird/pull/7706

pwespi:
test262 diff:

```
Duration:
     +0.29s

Summary:
    Diff Tests:
        +3 ✅    -3 ❌   

Diff Tests:
    test/language/statements/const/syntax/const-invalid-assignment-next-expression-for.js   ❌ -> ✅
    test/language/statements/const/syntax/const-invalid-assignment-statement-body-for-in.js ❌ -> ✅
    test/language/statements/const/syntax/const-invalid-assignment-statement-body-for-of.js ❌ -> ✅
```

PR #5733, opened 2025-08-06
LibJS: Throw error when value of local const variable is reassigned

kalenikaliaksandr:
Fixes https://github.com/LadybirdBrowser/ladybird/issues/5683

test262 diff (tests are located in `language/statements/const/syntax/`):
const-invalid-assignment-next-expression-for.js   ❌ -> ✅
const-invalid-assignment-statement-body-for-in.js ❌ -> ✅
const-invalid-assignment-statement-body-for-of.js ❌ -> ✅

### DIFF
--- a/Libraries/LibJS/Bytecode/Generator.cpp
+++ b/Libraries/LibJS/Bytecode/Generator.cpp
@@ -882,6 +882,10 @@ void Generator::emit_set_variable(JS::Identifier const& identifier, ScopedOperan
 {
     if (identifier.is_local()) {
         auto local_index = identifier.local_index();
+        if (initialization_mode == Bytecode::Op::BindingInitializationMode::Set && identifier.declaration_kind() == DeclarationKind::Const) {
+            emit<Bytecode::Op::ThrowInvalidAssignToConst>();
+            return;
+        }
         if (value.operand().is_local() && local_index.is_variable() && value.operand().index() == local_index.index) {
             // Moving a local to itself is a no-op.
             return;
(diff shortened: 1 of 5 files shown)


Lubrsi:
While this throws an error, the issue is that the const has still been changed, because the instructions such as Increment write directly to the local register.

Libraries/LibJS/Bytecode/Op.h
```diff
@@ -2373,6 +2373,17 @@ class ThrowIfTDZ final : public Instruction {
     Operand m_src;
 };
 
+class ThrowInvalidAssignToConst final : public Instruction {
```
Lubrsi, line 2376: issue: Since this is an unconditional throw, IsTerminator needs setting to true
suggestion: Copy the line from Throw

Libraries/LibJS/Bytecode/Generator.cpp
```diff
@@ -882,6 +882,10 @@ void Generator::emit_set_variable(JS::Identifier const& identifier, ScopedOperan
 {
     if (identifier.is_local()) {
         auto local_index = identifier.local_index();
+        if (initialization_mode == Bytecode::Op::BindingInitializationMode::Set && identifier.declaration_kind() == DeclarationKind::Const) {
+            emit<Bytecode::Op::ThrowInvalidAssignToConst>();
```
Lubrsi, line 886: issue: Since this is an unconditional throw, perform_needed_unwind should be called.
suggestion:
```suggestion
            perform_needed_unwinds<Bytecode::Op::Throw>();
            emit<Bytecode::Op::ThrowInvalidAssignToConst>();
```

Lubrsi (changes requested):

PR #5179, opened 2025-06-23
LibWeb: Fix text-shadow position being incorrectly scaled with non 100% zoom or high DPI monitors

twhlynch:
~text-shadow was being offset to the left by a margin derived from both sides of blur radius instead of just the left side.~
Was not an issue with the margin being wrong, but rather appeared that way on my computer, having a high DPI screen. Now, this should fix the bug which appears when zoomed.

Ive added a FIXME as while the behaviour is improved and seems perfect for integer value scales, there are still sometimes small incorrect offsets.

Correct:
Screenshot 2025-06-23 at 10 31 55 pm
Screenshot 2025-06-23 at 10 58 54 pm

Before:
Screenshot 2025-06-23 at 10 32 11 pm
Screenshot 2025-06-23 at 10 59 46 pm

After:
Screenshot 2025-06-23 at 10 32 16 pm
Screenshot 2025-06-23 at 10 52 52 pm
Slight error mentioned:
Screenshot 2025-06-24 at 7 36 21 pm

PR #5734, opened 2025-08-06
AK/LibWeb: Clamp and censor top-level calc results

Calme1709:
This PR adds functionality to clamp and censor values returned from top level CSS math functions. As part of this we needed to support formatting floats in scientific notation when we have a specified precision, as otherwise we would incorrectly format `AK::NumericLimits<float>::max()` and `AK::NumericLimits<float>::lowest()`. This in turn revealed inaccuracies when converting `transform` rotate functions to matrices which I have added in an ad-hoc workaround for.

Gains us 65 WPT tests.


Meta/Lagom/Tools/CodeGenerators/LibWeb/GenerateCSSPropertyID.cpp
```diff
@@ -808,6 +808,61 @@ bool property_accepts_type(PropertyID property_id, ValueType value_type)
...
+            return "Resolution"sv;
+        if (type_name == "string")
+            return "String"sv;
+        if (type_name == "time")
+            return "Time"sv;
+        if (type_name == "url")
+            return "Url"sv;
+
+        VERIFY_NOT_REACHED();
+    };
```
gmta: Feels like a case for `camel_casify`, you'd have to update `OpenTypeTag` to be `OpentypeTag` though.

Libraries/LibWeb/CSS/StyleValues/CalculatedStyleValue.cpp
```diff
@@ -141,8 +183,17 @@ static String serialize_a_math_function(CalculationNode const& fn, CalculationCo
...
+
+        return clamped_value.visit(
+            [&](Number const& value) { return value.to_string(); },
+            [&](Angle const& value) { return value.to_string(serialization_mode); },
+            [&](Flex const& value) { return value.to_string(serialization_mode); },
+            [&](Frequency const& value) { return value.to_string(serialization_mode); },
+            [&](Length const& value) { return value.to_string(serialization_mode); },
+            [&](Percentage const& value) { return value.to_string(); },
+            [&](Resolution const& value) { return value.to_string(serialization_mode); },
+            [&](Time const& value) { return value.to_string(serialization_mode); });
```
gmta: Maybe we should introduce `SerializationMode` in the two missing methods so we can use `auto` for the three blocks in this commit:

```
             [&](auto const& value) { return value.to_string(serialization_mode); });
```

Libraries/LibWeb/CSS/CSSStyleProperties.cpp
```diff
@@ -737,40 +737,46 @@ RefPtr<CSSStyleValue const> CSSStyleProperties::style_value_for_computed_propert
             return true;
         };
 
+        // FIXME: Converting rotation functions to matrices is not perfectly accurate, for example rotate(270deg) is
+        //        converted to (1.192488e-8, -1, 1, 1.192488-8, 0, 0) rather than the expected (0, -1, 1, 0, 0, 0), to
+        //        avoid including these inaccuracies in the serialized matrices we round the matrix components to 6dps
+        //        below. Ideally we would find a way to compute the matrices without these inaccuracies.
+        static float matrix_component_precision = 1000000;
```
gmta: What's the rationale for using 6 here? And how come other browsers do not have this inaccuracy?

Libraries/LibWeb/CSS/StyleValues/NumberStyleValue.cpp
```diff
@@ -13,7 +13,7 @@ namespace Web::CSS {
 
 String NumberStyleValue::to_string(SerializationMode) const
 {
-    return String::number(m_value);
+    return MUST(String::formatted("{:.6}", m_value));
```
gmta, line 22: Why do we need to do this? Is there a spec somewhere that asks us to show 6 decimal places?

tcl3: Not the PR author, but I believe the relevant spec text is here: https://drafts.csswg.org/cssom/#serialize-a-css-value.

> &lt;number&gt;
> A base-ten number using digits 0-9 (U+0030 to U+0039) in the shortest form possible, using "." to separate decimals (if any), rounding the value if necessary to not produce more than 6 decimals, preceded by "-" (U+002D) if it is negative.

AtkinsSJ: That should probably be pulled out into Serialise.h/cpp with a spec link, so it's in one central location. Not necessarily in this PR though.

gmta: Just to be sure; I'd like to have a spec link here regardless of whether it is going to be moved as part of this PR 😅

Calme1709: I have added a spec link and a FIXME to move it to Serialize.h

Libraries/LibWeb/CSS/CSSStyleProperties.cpp
Calme1709: I have removed this now and actually addressed the source of the inaccuracies in line with what other browsers do.

Libraries/LibWeb/CSS/StyleValues/CalculatedStyleValue.cpp
Calme1709: Done - unfortunately only compresses one block (the other two aren't related to serialization). I also added commits for two small housekeeping things in the process of doing this (removing unused imports in Percentage.h and moving the Percentage formatter from PercentageOr.h to Percentage.h)

Meta/Lagom/Tools/CodeGenerators/LibWeb/GenerateCSSPropertyID.cpp
Calme1709: Done, `OpentypeTag` is more in line with spec anyway.

Libraries/LibWeb/CSS/CSSStyleProperties.cpp
Calme1709: See https://github.com/LadybirdBrowser/ladybird/pull/5734/commits/b094d53bd758635ccf78349d0963d3dd26e415b4

Calme1709:
Rebased to resolve merge conflicts - this PR now regresses 2 tests in `css/css-values/calc-infinity-nan-computed.html` which I believe can be avoided by merging #5179 first

EDIT: It turns out these aren't actually regressions as the expectation file on master is incorrect (as it has no associated test file).

Libraries/LibGfx/Matrix4x4.h
```diff
@@ -63,6 +63,13 @@ constexpr static Matrix4x4<T> rotation_matrix(Vector3<T> const& axis, T angle)
 {
     T c, s;
     AK::sincos(angle, s, c);
+
+    if (-AK::NumericLimits<T>::epsilon() < c && c < AK::NumericLimits<T>::epsilon())
+        c = 0.0;
+
+    if (-AK::NumericLimits<T>::epsilon() < s && s < AK::NumericLimits<T>::epsilon())
+        s = 0.0;
```
Hendiadyoin1: Consider comparing against `abs(c)`/`abs(s)`

AK/Format.cpp
```diff
@@ -584,34 +568,47 @@ ErrorOr<void> FormatBuilder::put_f32_or_f64(
...
+            TRY(special_case_builder.try_append(upper_case ? "INF"sv : "inf"sv));
 
-        return put_string(builder.string_view(), align, min_width, NumericLimits<size_t>::max(), fill);
+        return put_string(special_case_builder.string_view(), align, min_width, NumericLimits<size_t>::max(), fill);
     }
 
+    if (precision.has_value() || base != 10)
+        return put_f64_with_precision(value, base, upper_case, zero_pad, use_separator, align, min_width, precision.value_or(6), fill, sign_mode, display_mode);
+
+    if (value == static_cast<T>(0.0) || value == static_cast<T>(-0.0)) {
```
Hendiadyoin1: `0` and `-0` equal each other, so a plain `0` check suffices

Meta/Lagom/Tools/CodeGenerators/LibWeb/GenerateCSSPropertyID.cpp
```diff
@@ -830,59 +830,8 @@ bool property_accepts_type(PropertyID property_id, ValueType value_type)
...
-                    property_generator.appendln("        case ValueType::String:");
-                } else if (type_name == "time") {
-                    property_generator.appendln("        case ValueType::Time:");
-                } else if (type_name == "url") {
-                    property_generator.appendln("        case ValueType::Url:");
-                } else {
-                    VERIFY_NOT_REACHED();
-                }
+                property_generator.set("type_name", title_casify(type_name));
+                property_generator.appendln("        case ValueType::@type_name@:");
```
Hendiadyoin1, line 840: This change seems to be in the wrong commit?

Libraries/LibWeb/CSS/StyleValues/CalculatedStyleValue.cpp
```diff
@@ -130,6 +130,48 @@ static NonnullRefPtr<CalculationNode const> simplify_2_children(T const& origina
...
+        [&](Flex const& value) -> CalculationNode::NumericValue { return Flex { clamp_and_censor(value.raw_value(), accepted_range->min, accepted_range->max), value.type() }; },
+        [&](Frequency const& value) -> CalculationNode::NumericValue { return Frequency { clamp_and_censor(value.raw_value(), accepted_range->min, accepted_range->max), value.type() }; },
+        [&](Length const& value) -> CalculationNode::NumericValue { return Length { clamp_and_censor(value.raw_value(), accepted_range->min, accepted_range->max), value.type() }; },
+        [&](Percentage const& value) -> CalculationNode::NumericValue { return Percentage { clamp_and_censor(value.value(), accepted_range->min, accepted_range->max) }; },
+        [&](Resolution const& value) -> CalculationNode::NumericValue { return Resolution { clamp_and_censor(value.raw_value(), accepted_range->min, accepted_range->max), value.type() }; },
+        [&](Time const& value) -> CalculationNode::NumericValue { return Time { clamp_and_censor(value.raw_value(), accepted_range->min, accepted_range->max), value.type() }; });
```
AtkinsSJ: Nit: Can we break these up onto multiple lines? It's a bit hard to read.

Libraries/LibWeb/CSS/StyleValues/CalculatedStyleValue.cpp
```diff
@@ -2685,12 +2728,40 @@ Optional<CalculatedStyleValue::ResolvedValue> CalculatedStyleValue::resolve_valu
...
+        accepted_range = m_context.accepted_type_ranges.get(ValueType::Length);
+    else if (value->type()->matches_percentage())
+        accepted_range = m_context.accepted_type_ranges.get(ValueType::Percentage);
+    else if (value->type()->matches_resolution(m_context.percentages_resolve_as))
+        accepted_range = m_context.accepted_type_ranges.get(ValueType::Resolution);
+    else if (value->type()->matches_time_percentage(m_context.percentages_resolve_as))
+        accepted_range = m_context.accepted_type_ranges.get(ValueType::Time);
+
+    if (!accepted_range.has_value()) {
+        dbgln_if(LIBWEB_CSS_DEBUG, "FIXME: Calculation context missing accepted type range");
```
AtkinsSJ: Please include the type's serialization in the logging, so it's easier to debug later.

Libraries/LibGfx/Matrix4x4.h
Calme1709: I think it is fine as is - a function call to abs would be more expensive aswell IMO

Libraries/LibWeb/CSS/StyleValues/CalculatedStyleValue.cpp
Calme1709: Done.

Libraries/LibWeb/CSS/StyleValues/CalculatedStyleValue.cpp
Calme1709: Not 100% on which type you are referring to here - I have updated it to print the `NumericCalculationNode`'s numeric type. I have also updated in `resolve_value`

Meta/Lagom/Tools/CodeGenerators/LibWeb/GenerateCSSPropertyID.cpp
Calme1709: Nope, this being part of `LibWeb: Rename ValueType::OpenTypeTag to ValueType::OpentypeTag` is as intended, the reason for renaming OpenTypeTag to OpentypeTag was so we could use `title_casify` here rather than the collection of if statements we had prior, see https://github.com/LadybirdBrowser/ladybird/pull/5734#discussion_r2256174475

AK/Format.cpp
Calme1709: Done.

Calme1709:
Reverted the change to `AK/Format.cpp` made [here](https://github.com/LadybirdBrowser/ladybird/compare/8b6cefea6bf81f84f2958255318b64f81696520c..8561a55d34a8a5618b0e944f28f02416d45f77c2#diff-ed46a10c48f9704561d6d3668b75fc3bec72ea9a36adc72c00793a6c82496dcb) in an attempt to resolve [this crash](https://github.com/LadybirdBrowser/ladybird/actions/runs/16793737936/job/47560143852?pr=5734)

Libraries/LibGfx/Matrix4x4.h
Hendiadyoin1: `abs` is always inlined* and as cheap as an `&`

*: It's an intrinsic to be precise

Calme1709: Good point - done.

Calme1709:
Looks like the crash was on master (introduced in 15cee5cc15d453387db2e3403ce922da22e2befd) but wasn't triggered until the `Tests/LibWeb/Text/input/wpt-import/css/css-values/typed_arithmetic.html` test was imported on this branch, I have added a new commit to this branch to fix it: 93437963d457a1a2fd90253586f4cec098aad312.

This is ready to merge from my end unless there is any further feedback.

Calme1709:
Rebased to resolve conflicts

Calme1709:
Rebased to resolve conflicts

PR #5736, opened 2025-08-06
LibWeb/CSS: Implement `env()`

### LINKED ISSUE #5705: `env()` CSS function support is missing
Reduced from https://web.telegram.org/a
```html
<!DOCTYPE html>
<html>
    <head>
        <style>
            .MiddleHeader {
                width: 100px;
                height: 100px;
                padding-right: max(200px, env(safe-area-inset-right));
                background: red;
            }
        </style>
    </head>
    <body>
        <div class="MiddleHeader"></div>
    </body>
</html>
```

AtkinsSJ:
Technically, `env()` should not be an ASF. (😱) This is why some tests
still fail - `env()` as specced is expected to have its syntax checked
fully at parse-time, whereas ASFs are not properly syntax-checked until
later. However, I think this approach was worth doing for a few reasons:

- `env()` behaves like an ASF otherwise. (It is replaced with a set of
  arbitrary component-values that are not known until computed-value
  time.)
- `env()` was defined before the ASF concept existed, so I strongly
  suspect it will be updated in the future to match that definition,
  with a couple of adjustments. (eg, `env()` is allowed in some extra
  places compared to `var()` and `attr()`.)
- This was much quicker and easier to implement (under 3 hours in total)
  compared to the greater amount of work to implement a whole separate
  system just for `env()`.
- Most of these tests are marked tentative, and the spec definition of
  `env()` is still somewhat in flux, so failing some is not a huge deal.

If in the future I turn out to be wrong on this, we can convert it to
its own special thing.

Fixes #5705


Libraries/LibWeb/CSS/Parser/ArbitrarySubstitutionFunctions.cpp
```diff
@@ -198,6 +201,64 @@ static Vector<ComponentValue> replace_an_attr_function(DOM::AbstractElement& ele
     // NB: Step 6 is a lambda defined at the top of the function.
 }
 
+// https://drafts.csswg.org/css-env/#substitute-an-env
+static Vector<ComponentValue> replace_an_env_function(DOM::AbstractElement& element, GuardedSubstitutionContexts& guarded_contexts, ArbitrarySubstitutionFunctionArguments const& arguments)
+{
+    // AD-HOC: env() is not defined as an ASF, (and was defined before the ASF concept was,) but behaves a lot like one.
```
kalenikaliaksandr: typo: "was,) but" -> "was), but"

AtkinsSJ: Actually not a typo - I've been doing this wrong my entire life. :sweat_smile: Fixed!

PR #5737, opened 2025-08-06
LibWeb/SVG: Apply SVGFeBlendElement blend mode

tcl3:
Previously, the blend mode was always assumed to be `normal`.

The failing tests on http://wpt.livecss/filter-effects/svgfeblendelement-mode-001.html are because the `SVGFEBlendElement.mode` setter doesn't currently prevent the value from being changed.


Libraries/LibGfx/CompositingAndBlendingOperator.h
```diff
@@ -8,13 +8,14 @@
 
 namespace Gfx {
 
+// The values of this emun should match the constants in SVGFEBlendELement.idl
```
AtkinsSJ: IMO, if these need to match those exactly, they should all have an explicit value.

gmta: Also typo: "emun"

Tests/LibWeb/Ref/expected/svg/blend-filter-ref.html
```diff
@@ -0,0 +1,5 @@
+<!DOCTYPE html>
+<svg width="100" height="100" xmlns="http://www.w3.org/2000/svg">
+    <rect x="0" y="0" width="100" height="100" fill="green"/>
+    <circle cx="50" cy="50" r="50" fill="black" filter="url(#blendFilter)"/>
```
gmta: This `filter` attribute should be removed

tcl3: Oops :sweat_smile:

Libraries/LibGfx/CompositingAndBlendingOperator.h
tcl3: > IMO, if these need to match those exactly, they should all have an explicit value.

The issue is that only the values up to `CompositingAndBlendingOperator::Luminosity` correspond to IDL values in `SVGFEBlendElement` and setting explicit values for only some of the elements makes the linter complain.

gmta (approved):

PR #5738, opened 2025-08-06
Tests: Improve fuzzy test reports

AtkinsSJ:
Main thing here is to report the fuzzy difference even if we're not doing a fuzzy test, so that we can more easily know what the configuration should be.

### DIFF
--- a/Tests/LibWeb/test-web/Fuzzy.cpp
+++ b/Tests/LibWeb/test-web/Fuzzy.cpp
@@ -13,7 +13,7 @@
 namespace TestWeb {
 
 // https://web-platform-tests.org/writing-tests/reftests.html#fuzzy-matching
-bool fuzzy_screenshot_match(URL::URL const& reference, Gfx::Bitmap const& bitmap_a, Gfx::Bitmap const& bitmap_b,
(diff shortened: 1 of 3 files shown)


Tests/LibWeb/test-web/Fuzzy.cpp
```diff
@@ -27,8 +27,10 @@ bool fuzzy_screenshot_match(URL::URL const& reference, Gfx::Bitmap const& bitmap
             return fuzzy_match.reference.value().equals(reference);
         return true;
     });
-    if (!fuzzy_match.has_value())
+    if (!fuzzy_match.has_value()) {
+        warnln("Screenshot mismatch: pixel error count {}, with maximum error {}. (No fuzzy config defined)", diff.pixel_error_count, diff.maximum_error);
         return diff.identical;
```
gmta: Ooh we can just do `return false;` here.

gmta (commented):
Nice!

Tests/LibWeb/test-web/Fuzzy.cpp
AtkinsSJ: Good point! Done

gmta (approved):

AtkinsSJ:
Changes: Realised while using this that we need the test URL logged too, otherwise it's hard to identify which report comes from which test when there are multiple.

gmta (approved):

PR #5743, opened 2025-08-06
LibWeb: Use the soon-to-be hovered element for tooltip notifications

awesomekling:
Instead of the currently hovered element.

cc @AtkinsSJ

### DIFF
--- a/Libraries/LibWeb/Page/EventHandler.cpp
+++ b/Libraries/LibWeb/Page/EventHandler.cpp
@@ -848,7 +848,7 @@ EventResult EventHandler::handle_mousemove(CSSPixelPoint viewport_position, CSSP
     }
 
     if (hovered_node_changed) {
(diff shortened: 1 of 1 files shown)


gmta (approved):

AtkinsSJ:
Looks good. There's still one oddity: On that test page, once the tooltip appears, the cursor switches from a caret to a standard pointer. I don't remember if that happened before, but it makes me suspect there's still something slightly wrong in this event code.

awesomekling:
> Looks good. There's still one oddity: On that test page, once the tooltip appears, the cursor switches from a caret to a standard pointer. I don't remember if that happened before, but it makes me suspect there's still something slightly wrong in this event code.

:thonk: Can't reproduce on macOS. Or maybe I need to build the Qt port..

AtkinsSJ:
Ah yeah, that was the Qt chrome on Linux. Makes me think it's probably unrelated to this then.

PR #5744, opened 2025-08-06
LibWeb: Implement `<feComposite>` SVG filter

### LINKED ISSUE #5793: BootstrapVue Icons in Phoronix forums render with grey background
### Summary

In the [Phoronix Forums](https://www.phoronix.com/forums/forum), the icons aren't displayed correctly. They are rendered with a grey background.

I included a reduced repro test case.

The CSS naming convention looks like it is from BootstrapVue. The framework is based on Bootstrap 4 but still pretty popular with vue.js users.

### Operating system

Linux

### Steps to reproduce

1. Visit [the Phoronix forums](https://www.phoronix.com/forums/forum)
2. See the icons with grey background

### Expected behavior

The icons are rendered transparently.

### Actual behavior

The icons are rendered with grey background.

N/A

```html
<!doctype html>                                                                      
<html>
  <head>
    <style>
.b-icon {
    display: inline-block;
    background: transparent url("./sprite_icons_general.svg") no-repeat scroll 0 0;
    overflow: hidden;
    text-indent: -99999px
}

.b-icon__bell {
    background-position: -646px -16px;
    width: 16px;
    height: 14px
}
    </style>
  </head>

  <body>
    <span class="b-icon b-icon__bell"></span>
  </body>
</html>
```
Icons file: 

Image

```shell
N/A
```

|Firefox|Ladybird|
|-|-|
|Image|Image|

Image

N/A

### Contribute a patch?

tcl3:
I have added a FIXME for the arithmetic operator, which I couldn't get working :sweat_smile: 

Fixes #5793


konradekk:
you could start with `SkiaUtils.cpp`:

```cpp
sk_sp<SkBlender> to_skia_blender(float k1, float k2, float k3, float k4)
{
    sk_sp<SkData> data = SkData::MakeWithCopy((float[4]) { k1, k2, k3, k4 }, sizeof(float) * 4);

    return SkRuntimeEffect::MakeForBlender(SkString(R"(
        uniform vec4 k;

        vec4 main(vec4 source, vec4 destination) {
            return k[0] * source * destination + k[1] * source + k[2] * destination + k[3];
        }
    )"))
        .effect->makeBlender(data);
}
```

if i got that right; then, instead of your `FIXME` in `SVGFECompositeElement.cpp`, put

```cpp
        case CompositingOperator::Arithmetic:
            return Gfx::CompositingAndBlendingOperator::Arithmetic;
```

with appended `Arithmetic` to the `Gfx::CompositingAndBlendingOperator`, so you could add the following in `Filter.cpp`

```cpp
Filter Filter::arithmetic_blend(Optional<Filter const&> background, Optional<Filter const&> foreground, float k1, float k2, float k3, float k4)
{
    sk_sp<SkImageFilter> background_skia = background.has_value() ? background->m_impl->filter : nullptr;
    sk_sp<SkImageFilter> foreground_skia = foreground.has_value() ? foreground->m_impl->filter : nullptr;

    auto filter = SkImageFilters::Blend(to_skia_blender(k1, k2, k3, k4), background_skia, foreground_skia);
    return Filter(Impl::create(filter));
}
```

and then `SVGFilterElement.cpp:L128-129` could be 

```cpp
            auto operator_ = composite_primitive->operator_();

            if (operator_ == Gfx::CompositingAndBlendingOperator::Arithmetic) {
                auto k1 = composite_primitive->k1()->base_val();
                auto k2 = composite_primitive->k2()->base_val();
                auto k3 = composite_primitive->k3()->base_val();
                auto k4 = composite_primitive->k4()->base_val();

                root_filter = Gfx::Filter::arithmetic_blend(background, foreground, k1, k2, k3, k4);
            } else {
                root_filter = Gfx::Filter::blend(background, foreground, operator_);
            }
```

i'm not sure, though, if it is working as intended. feel free to use any/all of the above! 🫡

tcl3:
My latest push adds the arithmetic filter, but I'm still getting a some color differences compared to other browsers.

I've tested the arithmetic filter with the following SVG:
```html
<svg width="100" height="100" fill="white" xmlns="http://www.w3.org/2000/svg">
<defs>
  <filter id="compositeFilter">
    <feFlood flood-color="white" result="whiteFill" />
    <feComposite in2="whiteFill" in="SourceGraphic" operator="arithmetic" k1=".1" k2=".2" k3=".3" k4=".4" />
  </filter>
</defs>

<rect x="0" y="0" width="100" height="100" fill="#000000" filter="url(#compositeFilter)" />
</svg>
```

If we treat the image as monochrome for this example the values for the relevant variables are:
* in1 = 0 (black)
* in2 = 1 (white)
* k1 = 0.1
* k2 = 0.2
* k3 = 0.3
* k4 = 0.4

Given the equation for the arithmetic filter:
 `(k1 * in1 * in2) + (k2 * in1) + (k3 * in2) + k4`
 
Python tells me the value I should get out is 0.7.

According to the GIMP eyedropper tool the value I get for each color channel is 0.698 - I assume the difference is due to quantization. However, the value that all other browsers get according to the eyedropper tool is 0.855.

I'm not sure why other browsers produce this brighter value.

i-love-coffee-i-love-tea:
> Given the equation for the arithmetic filter:
 `(k1 * in1 * in2) + (k2 * in1) + (k3 * in2) + k4`
> 
> Python tells me the value I should get out is 0.7.

The first two bracketed terms are 0 because of multiplication with 0. Then there is only (0.3 * 1) + 0.4 left. It's obviously mathematically correct. The equation is correct too (https://www.w3.org/TR/filter-effects-1/#elementdef-fecomposite). There must be something else happening.

gmta:
Results from Chrome on macOS:

<details><summary>Testcase SVG</summary>

NOTE: I've swapped in/in2 compared to the testcase mentioned earlier.

```
<svg width="100" height="100" fill="white" xmlns="http://www.w3.org/2000/svg">
<defs>
  <filter id="compositeFilter">
    <feFlood flood-color="white" />
    <feComposite in2="SourceGraphic" operator="arithmetic" k1="0" k2="0" k3="0.3" k4="0.4" />
  </filter>
</defs>

<rect x="0" y="0" width="100" height="100" fill="#000000" filter="url(#compositeFilter)" />
</svg>
```

</details>

| k4=0.4, k3: | 0.0 | 0.1 | 0.2 | 0.3 | 0.4 | 0.5 | 0.6 | 0.7 | 0.8 | 0.9 | 1.0 |
|--|--|--|--|--|--|--|--|--|--|--|--|
| Value | 1.0 | 0.953 | 0.902 | 0.847 | 0.788 | 0.729 | 0.667 | 0.667 | 0.667 | 0.667 | 0.667 |

| k3=0.3, k4: | 0.0 | 0.1 | 0.2 | 0.3 | 0.4 | 0.5 | 0.6 | 0.7 | 0.8 | 0.9 | 1.0 |
|--|--|--|--|--|--|--|--|--|--|--|--|
| Value | 0.698 | 0.816 | 0.831 | 0.843 | 0.847 | 0.851 | 0.851 | 0.855 | 0.906 | 0.957 | 1.0 |

Looking at these values, sure looks like some curve is being applied. Gamma application or sRGB -> linear conversion?

Also, the only value corresponding to our result in Chrome is at k3=0.3, k4=0.0.

gmta:
Based on the partial calculation `(k3 * in2) + k4`, if I apply the sRGB-linear -> sRGB transformation I get values that seem to correspond a bit better:

`sRGB = 1.055 * pow(sRGB_linear, 1 / 2.4) - 0.055;`

This gives me `0.853` for `k3=0.3,k4=0.4` and `0.665` for `k3=1.0,k4=0.4`.

tcl3:
I think you're right, the difference appears to be a linear RGB -> sRGB conversion from the application of the [`color-interpolation-filters` property](https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Attribute/color-interpolation-filters), which has a default value of `linearRGB`. 

If I set `<filter color-interpolation-filters="sRGB" ...>`, then we get the same value as Chrome/Firefox.

Libraries/LibWeb/SVG/SVGFECompositeElement.idl
```diff
@@ -0,0 +1,26 @@
...
+    const unsigned short SVG_FECOMPOSITE_OPERATOR_OVER = 1;
+    const unsigned short SVG_FECOMPOSITE_OPERATOR_IN = 2;
+    const unsigned short SVG_FECOMPOSITE_OPERATOR_OUT = 3;
+    const unsigned short SVG_FECOMPOSITE_OPERATOR_ATOP = 4;
+    const unsigned short SVG_FECOMPOSITE_OPERATOR_XOR = 5;
+    const unsigned short SVG_FECOMPOSITE_OPERATOR_ARITHMETIC = 6;
+
+    readonly attribute SVGAnimatedString in1;
+    readonly attribute SVGAnimatedString in2;
+    [ImplementedAs=operator_for_bindings] readonly attribute SVGAnimatedEnumeration operator_for_bindings;
```
konradekk: wasn't that
```suggestion
    [ImplementedAs=operator_for_bindings] readonly attribute SVGAnimatedEnumeration operator;
```

konradekk (commented):
the filter itself looks legit; the color space issues could be taken care of separately? (a todo somewhere would be enough?)

tcl3:
> the color space issues could be taken care of separately? (a todo somewhere would be enough?)

I have an implementation of `color-interpolation-filters`, which corrects the color difference, but it relies on changes in #5869 and #5856, so I'd prefer to add that as a separate PR.

Tests/LibWeb/Ref/expected/svg/composite-filter-ref.html
```diff
@@ -0,0 +1,4 @@
+<!DOCTYPE html>
+<svg width="110" height="110" xmlns="http://www.w3.org/2000/svg">
+    <rect x="10" y="10" width="100" height="100" fill="green" filter="url(#compositeFilter)" />
```
gmta: You should drop the `filter=".."` here

tcl3: Fixed.

gmta (approved):

PR #5751, opened 2025-08-06
Tests: Fix Text/input/XHR/XMLHttpRequest-override-mimetype-blob.html …

bn3t:
…is very slow on macOS


Tests/LibWeb/TestConfig.ini
gmta, line 1: Commit title: please describe the change instead of "Fix slow", something like:

> Tests: Use 127.0.0.1 instead of localhost to prevent slow DNS resolution

bn3t: Done

AtkinsSJ (approved):

Issue #5551, opened 2025-07-21
Build and Runtime Report on Ubuntu 25.04 (Successful with Workarounds; Runtime Warnings on Complex Sites)


DavidRexGlenn:
### Summary

Successful build of Ladybird on Ubuntu 25.04 after multiple workarounds for dependency resolution and slow linking due to LTO; runtime functions but shows warnings, unimplemented features, and JS exceptions on complex, JS-heavy sites like Newegg.

Operating system
*Linux (Ubuntu 25.04)

### Operating system

Linux

### Steps to reproduce

1. Update packages: sudo apt update && sudo apt install build-essential cmake ninja-build pkg-config libssl-dev libmpfr-dev libsqlite3-dev qt6-base-dev qt6-tools-dev-tools qt6-wayland qt6-svg-dev qt6-multimedia-dev libglib2.0-dev libcairo2-dev libjpeg-dev libgif-dev libwebp-dev libopenjp2-7-dev libpulse-dev ccache nasm fonts-liberation2 curl unzip zip autoconf autoconf-archive automake libavcodec-dev libgl1-mesa-dev tar git python3-venv mesa-common-dev libicu-dev.
2. Clone repo: git clone https://github.com/LadybirdBrowser/ladybird.git && cd ladybird && git submodule update --init --recursive.
3. Build missing deps from source (simdutf, FastFloat, libtommath): Clone each from GitHub, cmake -B build -DCMAKE_BUILD_TYPE=Release, cmake --build build, sudo cmake --install build.
4. Install vcpkg: git clone https://github.com/microsoft/vcpkg.git && cd vcpkg && ./bootstrap-vcpkg.sh.
5. Set env to disable LTO: export CFLAGS="-fno-lto -fno-fat-lto-objects" && export CXXFLAGS="-fno-lto -fno-fat-lto-objects" && export LDFLAGS="-fno-lto".
[...]

PR #5757, opened 2025-08-07
Documentation: Fix gpg key file format for llvm repo

high3eam:
Fix the file format for the imported gpg key of the llvm repository.
This way, the installation of the required build tools succeeds using the imported signing key.

### DIFF
--- a/Documentation/BuildInstructionsLadybird.md
+++ b/Documentation/BuildInstructionsLadybird.md
@@ -47,12 +47,12 @@ sudo apt update -y && sudo apt install cmake -y
 
 ```bash
 # Add LLVM GPG signing key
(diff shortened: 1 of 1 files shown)


trflynn89:
CC @ADKaster. There were some format adjustments around this made in commit 53f4e988182bd0f5ec0fc8c14c5af72cb93acfeb, so let's be sure they're aligned.

ADKaster:
That's strange. I guess for context this is related to #5551 ? 

We have a devconatiner feature under `.devcontainer/` that does this same thing to add the key, and it works fine there. Is this an Ubuntu/apt change? Or something up with the llvm apt repos recently? Why do we need to "dearmor" the key anyway?

ghost:
@ADKaster I am on Ubuntu 25.04 and when I add the key to the system using the current command in the docs, it says that the key filetype is incorrect and that no publick key could be found (paths are correctly configured).

PR #5766, opened 2025-08-07
LibWeb: Add initial CookieStore support

IdanHo:
This PR adds support for the CookieStore API (https://cookiestore.spec.whatwg.org/) to ladybird.
It should be mostly feature-complete, aside from features that are reliant on more general concepts we do not yet fully support. (mostly service workers)
Correspondingly, it makes us pass all relevant WPT tests that do not rely on service workers and such.

While writing the implementation I found and reported 2 small spec issues:
https://github.com/whatwg/cookiestore/issues/277 (already fixed :slightly_smiling_face:)
https://github.com/whatwg/cookiestore/issues/282

The vast majority of WPT tests are unfortunately not currently importable as the API only supports adding secure cookies, and our test runner always runs in a non-secure (non-https) origin.


Libraries/LibWeb/ServiceWorker/ServiceWorkerGlobalScope.cpp
```diff
@@ -78,4 +86,15 @@ GC::Ptr<WebIDL::CallbackType> ServiceWorkerGlobalScope::onmessageerror()
...
 
+// https://cookiestore.spec.whatwg.org/#ServiceWorkerGlobalScope
+GC::Ref<CookieStore::CookieStore> ServiceWorkerGlobalScope::cookie_store()
+{
+    auto& realm = this->realm();
+
+    // The cookieStore getter steps are to return this’s associated CookieStore.
+    if (!m_cookie_store)
+        m_cookie_store = realm.create<CookieStore::CookieStore>(realm);
+    return GC::Ref { *m_cookie_store };
```
trflynn89: No need to write out `GC::Ref` manually:
```c++
return *m_cookie_store;
```

Meta/gn/secondary/Userland/Libraries/LibWeb/CookieStore/BUILD.gn
```diff
@@ -0,0 +1,5 @@
+source_set("CookieStore") {
```
trflynn89, line 1: FYI you don't need to worry about the GN build. It's very, very broken at this point.
(You also don't need to undo this change, just a heads up for the future).

Libraries/LibWeb/CookieStore/CookieStore.cpp
```diff
@@ -23,4 +30,102 @@ void CookieStore::initialize(JS::Realm& realm)
...
+    return CookieListItem {
+        .name = cookie.name,
+        .value = cookie.value,
+    };
+}
+
+// https://cookiestore.spec.whatwg.org/#query-cookies
+static Vector<CookieListItem> query_cookies(PageClient& client, URL::URL const& url, Optional<String> const& name)
+{
+    // 1. Perform the steps defined in Cookies § Retrieval Model to compute the "cookie-string from a given cookie store" with url as request-uri. The cookie-string itself is ignored, but the intermediate cookie-list is used in subsequent steps.
```
trflynn89: Let's wrap at ~120 chars (here and subsequently)

Libraries/LibWeb/CookieStore/CookieStore.cpp
```diff
@@ -23,4 +30,102 @@ void CookieStore::initialize(JS::Realm& realm)
...
+
+    // 3. If origin is an opaque origin, then return a promise rejected with a "SecurityError" DOMException.
+    if (origin.is_opaque())
+        return WebIDL::create_rejected_promise(realm, WebIDL::SecurityError::create(realm, "Document origin is opaque"_string));
+
+    // 4. Let url be settings’s creation URL.
+    auto url = settings.creation_url;
+
+    // 5. Let p be a new promise.
+    auto p = WebIDL::create_promise(realm);
```
trflynn89: Generally prefer to give one-letter spec variables a full name, e.g. `auto promise`

Libraries/LibWeb/CookieStore/CookieStore.h
```diff
@@ -6,19 +6,41 @@
...
+    virtual void visit_edges(Cell::Visitor&) override;
+
+    GC::Ref<PageClient> m_client;
 };
 
 }
+
+namespace Web::Bindings {
+
+JS::Value cookie_list_item_to_value(JS::Realm&, CookieStore::CookieListItem);
```
trflynn89: Hmm can the declaration be generated in the prototype header?

Libraries/LibWeb/CookieStore/CookieStore.h
```diff
@@ -6,19 +6,41 @@
 
 #pragma once
 
+#include <AK/Optional.h>
+#include <AK/String.h>
+#include <LibWeb/Bindings/CookieStorePrototype.h>
 #include <LibWeb/DOM/EventTarget.h>
+#include <LibWeb/HighResolutionTime/DOMHighResTimeStamp.h>
```
trflynn89, line 13: Is this used?

Libraries/LibWeb/CookieStore/CookieStore.cpp
```diff
@@ -128,4 +129,53 @@ GC::Ref<WebIDL::Promise> CookieStore::get(String name)
     return p;
 }
 
+static JS::Value cookie_list_to_value(JS::Realm& realm, Vector<CookieListItem> const& cookie_list)
+{
+    auto array = MUST(JS::Array::create(realm, 0));
+    for (auto i = 0u; i < cookie_list.size(); ++i)
+        MUST(array->create_data_property_or_throw(i, Bindings::cookie_list_item_to_value(realm, cookie_list[i])));
+    return array;
```
trflynn89: Could potentially use `JS::Array::create_from` here

Libraries/LibWeb/CookieStore/CookieStore.cpp
```diff
@@ -129,6 +131,77 @@ GC::Ref<WebIDL::Promise> CookieStore::get(String name)
...
+        if (!parsed.has_value())
+            return WebIDL::create_rejected_promise(realm, JS::TypeError::create(realm, "url is invalid"sv));
+
+        // 2. If this’s relevant global object is a Window object and parsed does not equal url with exclude fragments set to true, then return a promise rejected with a TypeError.
+        if (is<HTML::Window>(HTML::relevant_global_object(*this)) && !parsed->equals(url, URL::ExcludeFragment::Yes))
+            return WebIDL::create_rejected_promise(realm, JS::TypeError::create(realm, "url does not match creation URL"sv));
+
+        // 3. If parsed’s origin and url’s origin are not the same origin, then return a promise rejected with a TypeError.
+        if (parsed->origin() != url.origin())
+            return WebIDL::create_rejected_promise(realm, JS::TypeError::create(realm, "url's origin does not match creation URL;s origin"sv));
```
trflynn89: Typo - `URL;s`

Libraries/LibWeb/CookieStore/CookieStore.cpp
```diff
@@ -251,4 +251,66 @@ GC::Ref<WebIDL::Promise> CookieStore::get_all(String name)
...
```
trflynn89: Here too

Libraries/LibWeb/CookieStore/CookieStore.cpp
```diff
@@ -313,4 +314,208 @@ GC::Ref<WebIDL::Promise> CookieStore::get_all(CookieStoreGetOptions const& optio
...
+
+        // 6. Let encodedDomain be the result of UTF-8 encoding parsedDomain.
+        auto encoded_domain = parsed_domain->serialize();
+
+        // 7. If the byte sequence length of encodedDomain is greater than the maximum attribute value size, then return failure.
+        if (encoded_domain.byte_count() > maximum_attribute_value_size)
+            return false;
+
+        // 8. Append `Domain`/encodedDomain to attributes.
+        parsed_cookie.domain = encoded_domain;
```
trflynn89: Could `move` this string

Libraries/LibWeb/CookieStore/CookieStore.cpp
```diff
@@ -313,4 +314,208 @@ GC::Ref<WebIDL::Promise> CookieStore::get_all(CookieStoreGetOptions const& optio
...
+
+    // 7. Let encodedName be the result of UTF-8 encoding name.
+    // 8. Let encodedValue be the result of UTF-8 encoding value.
+
+    // 9. If the byte sequence length of encodedName plus the byte sequence length of encodedValue is greater than the maximum name/value pair size, then return failure.
+    if (name.byte_count() + value.byte_count() > maximum_name_value_pair_size)
+        return false;
+
+    // 10. Let host be url’s host
+    auto host = url.host();
```
trflynn89: `auto const&`

Libraries/LibWeb/CookieStore/CookieStore.cpp
```diff
@@ -518,4 +518,47 @@ GC::Ref<WebIDL::Promise> CookieStore::set(String name, String value)
...
+    // 4. Let url be settings’s creation URL.
+    auto url = settings.creation_url;
+
+    // 5. Let p be a new promise.
+    auto p = WebIDL::create_promise(realm);
+
+    // 6. Run the following steps in parallel:
+    Platform::EventLoopPlugin::the().deferred_invoke(GC::create_function(realm.heap(), [&realm, client = m_client, p, url = move(url), options = options]() {
+        // 1. Let r be the result of running set a cookie with url, options["name"], options["value"], options["expires"], options["domain"], options["path"], options["sameSite"], and options["partitioned"].
+        auto r = set_a_cookie(client, url, options.name, options.value, options.expires, options.domain, options.path, options.same_site, options.partitioned);
```
trflynn89: Same here re: one-letter var names. `result` would make sense

Libraries/LibWeb/CookieStore/CookieStore.h
```diff
@@ -26,6 +26,17 @@ struct CookieStoreGetOptions {
...
 
+// https://cookiestore.spec.whatwg.org/#dictdef-cookieinit
+struct CookieInit {
+    String name;
+    String value;
+    Optional<HighResolutionTime::DOMHighResTimeStamp> expires;
+    Optional<String> domain;
+    String path;
+    Bindings::CookieSameSite same_site;
+    bool partitioned;
```
trflynn89: We should init plain types in structs:
```c++
bool partitioned { false };
```

Libraries/LibWeb/CookieStore/CookieStore.h
```diff
@@ -37,6 +37,14 @@ struct CookieInit {
     bool partitioned;
 };
 
+// https://cookiestore.spec.whatwg.org/#dictdef-cookiestoredeleteoptions
+struct CookieStoreDeleteOptions {
+    String name;
+    Optional<String> domain;
+    String path;
+    bool partitioned;
```
trflynn89: Same here

AK/HashTable.h
```diff
@@ -478,6 +478,25 @@ class HashTable {
         return has_removed_anything;
     }
 
+    template<typename TUnaryPredicate>
+    Vector<T> take_all_matching(TUnaryPredicate const& predicate)
```
trflynn89, line 482: Let's add a unit test

Libraries/LibWebView/CookieJar.cpp
```diff
@@ -607,7 +624,14 @@ UnixDateTime CookieJar::TransientStorage::purge_expired_cookies(Optional<AK::Dur
...
 
     auto is_expired = [&](auto const&, auto const& cookie) { return cookie.expiry_time < now; };
-    m_cookies.remove_all_matching(is_expired);
+    auto removed_entries = m_cookies.take_all_matching(is_expired);
+    if (!removed_entries.is_empty()) {
+        Vector<Web::Cookie::Cookie> removed_cookies;
+        removed_cookies.ensure_capacity(removed_entries.size());
+        for (auto const& entry : removed_entries)
+            removed_cookies.unchecked_append(move(entry.value));
+        notify_cookies_changed(move(removed_cookies));
```
trflynn89, line 633: I might be squinting at this wrong, but do we need to make a copy of the vector?

trflynn89: Oh, `take_all_matching` is returning a list of key-value pairs. A bit unfortunate, but probably fine for now

Libraries/LibWeb/CookieStore/CookieStore.cpp
```diff
@@ -665,4 +666,170 @@ GC::Ref<WebIDL::Promise> CookieStore::delete_(CookieStoreDeleteOptions const& op
...
+    // The observable changes for url are the set of cookie changes to cookies in a cookie store which meet the
+    // requirements in step 1 of Cookies § Retrieval Algorithm’s steps to compute the "cookie-string from a given
+    // cookie store" with url as request-uri, for a "non-HTTP" API.
+    // https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-14#name-retrieval-algorithm
+    auto canonicalized_domain = Cookie::canonicalize_domain(url);
+    if (!canonicalized_domain.has_value())
+        return {};
+
+    // FIXME: The retrieval's same-site status is "same-site" if the Document's "site for cookies" is same-site with the
+    //  top-level origin as defined in Section 5.2.1 (otherwise it is "cross-site"), and the retrieval's type is "non-HTTP".
```
trflynn89: "top-level" should be indented to match "The" above

Libraries/LibWeb/ServiceWorker/ServiceWorkerGlobalScope.cpp
IdanHo: cool, updated

Libraries/LibWeb/CookieStore/CookieStore.cpp
IdanHo: done

Libraries/LibWeb/CookieStore/CookieStore.cpp
IdanHo: sure, updated

Libraries/LibWeb/CookieStore/CookieStore.h
IdanHo: no, I tried doing that initially, but since the dictionary struct is defined in the non-generated header, we would have to include that header in the prototype header, and that would create an include loop (since the non-generated header also needs to include the prototype header)

Libraries/LibWeb/CookieStore/CookieStore.h
IdanHo: nope, good catch

Libraries/LibWeb/CookieStore/CookieStore.cpp
IdanHo: It's not shorter, but it is a bit less involved, so switched to that

Libraries/LibWeb/CookieStore/CookieStore.cpp
IdanHo: fixed

Libraries/LibWeb/CookieStore/CookieStore.cpp
IdanHo: fixed

Libraries/LibWeb/CookieStore/CookieStore.cpp
IdanHo: yep, fixed

Libraries/LibWeb/CookieStore/CookieStore.cpp
IdanHo: fixed

Libraries/LibWeb/CookieStore/CookieStore.cpp
IdanHo: yep, updated

Libraries/LibWeb/CookieStore/CookieStore.h
IdanHo: done

Libraries/LibWeb/CookieStore/CookieStore.h
IdanHo: done

AK/HashTable.h
IdanHo: good point, added

Libraries/LibWeb/CookieStore/CookieStore.cpp
IdanHo: sure

Libraries/LibWeb/CookieStore/CookieStore.h
trflynn89: Oh wait, just noticed - the generated method should be taking the value to be converted as a `const&` anyways. Then a forward declaration of `CookieListItem` can be added to `LibWeb/Forward.h`, or generated - then this method declaration can also be generated.

IdanHo: I checked, this is still non-trivial, since we don't have a solution at the moment in the IDL generator to resolve the namespace of things - i.e. in the cpp files we just add `using namespace` for every single namespace,  which we wouldn't want to do in header files.
If we still want to do this, we would have to declare the dictionary in the global Web namespace

trflynn89: Got it, let's not for now then!
We should still take the param by `const&` though, no need to copy

IdanHo: Unfortunately we do have to copy atm, since the current implementation of the IDL C++ => javascript code generations assumes it gets everything by-value (e.g. it does `release_value` on optionals and so on), since that is always correct for return values where it's mostly used.
I can change it to take it by const& and do the copy inside the generated code, as well as adding a nearby FIXME to remove that if/once we update the IDL generator to support that.

trflynn89: Ah, that's unfortunate. We can stick with this for now and make the generator behave better for this case later.

Libraries/LibWeb/CookieStore/CookieStore.cpp
```diff
@@ -129,4 +130,54 @@ GC::Ref<WebIDL::Promise> CookieStore::get(String name)
     return promise;
 }
 
+static JS::Value cookie_list_to_value(JS::Realm& realm, Vector<CookieListItem> const& cookie_list)
+{
+    Vector<JS::Value> cookie_values;
+    cookie_values.ensure_capacity(cookie_list.size());
+    for (auto const& cookie : cookie_list)
+        cookie_values.unchecked_append(Bindings::cookie_list_item_to_value(realm, cookie));
+    return JS::Array::create_from(realm, cookie_values);
```
trflynn89: Oh, what I meant here was using the `create_from` overload that accepts a mapping function. Something like:
```c++
return JS::Array::create_from<CookieListItem>(realm, cookie_list, [&](auto const& cookie) {
    return Bindings::cookie_list_item_to_value(realm, cookie);
});
```

That overload is generally safer too as the `Vector<JS::Value>` here is not safe from GC.

IdanHo: Oh, that is much better, not sure how I missed that overload. Updated

trflynn89 (approved):

PR #5769, opened 2025-08-07
Tests/LibJS+Utilities: Enable abench+dns+image+wasm+xml and test-js on Windows

ayeteadoe:
If desirable I can file an issue for this for eventual follow-up but I think having test-js running in Windows CI now, despite not having ASAN enabled, is still valuable to catch general LibJS Windows bitrot at CI-time

### DIFF
--- a/Libraries/LibDNS/CMakeLists.txt
+++ b/Libraries/LibDNS/CMakeLists.txt
@@ -2,5 +2,5 @@ set(SOURCES
     Message.cpp
 )
 
-ladybird_lib(LibDNS dns)
+ladybird_lib(LibDNS dns EXPLICIT_SYMBOL_EXPORT)
 target_link_libraries(LibDNS PRIVATE LibCore PUBLIC LibCrypto)
--- a/Libraries/LibDNS/Message.h
+++ b/Libraries/LibDNS/Message.h
@@ -11,6 +11,7 @@
 #include <AK/IPv6Address.h>
 #include <AK/RedBlackTree.h>
 #include <AK/Time.h>
+#include <LibDNS/Export.h>
 
 namespace DNS {
 namespace Messages {
@@ -84,7 +85,7 @@ struct Header {
     NetworkOrdered<u16> additional_count;
 };
 
-struct DomainName {
+struct DNS_API DomainName {
     Vector<ByteString> labels;
 
     static DomainName from_string(StringView);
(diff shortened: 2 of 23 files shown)


alimpfard:
Sorry for the conflicts 😅 
The annotations should be relatively simple to port over though, I didn't change the API surface

ayeteadoe:
No problem! It was good timing actually as you removed a lot of code in `wasm.cpp` that I initially had to workaround via ifdef's.  The key things that still need addressing for Windows in `wasm.cpp` is Wasi support and `Line::Editor` support, but I plan to tackle that in a future PR as I don't want this one to get too large.

Libraries/LibWasm/AbstractMachine/BytecodeInterpreter.cpp
```diff
@@ -2539,7 +2539,13 @@ CompiledInstructions try_compile_instructions(Expression const& expression, Span
...
     CompiledInstructions result;
     result.dispatches.ensure_capacity(expression.instructions().size());
-    result.extra_instruction_storage.ensure_capacity(ceil_div(expression.instructions().size(), 2ul)); // At most half of the instructions can be replaced with synthetic instructions, as the detected sequences are 3 long.
+#if !defined(AK_OS_WINDOWS)
+#    define INSTRUCTIONS_SIZE_DENOM 2ul
+#else
+    // NOTE: ceil_div static_Assert fails on Windows when using an unsigned long: "error: static assertion failed due to requirement 'sizeof(unsigned long long) == sizeof(unsigned long)'"
+#    define INSTRUCTIONS_SIZE_DENOM 2ull
+#endif
+    result.extra_instruction_storage.ensure_capacity(ceil_div(expression.instructions().size(), INSTRUCTIONS_SIZE_DENOM)); // At most half of the instructions can be replaced with synthetic instructions, as the detected sequences are 3 long.
```
R-Goc: just uz works

ayeteadoe:
Windows is passing for me locally but seems a few other symbols (and the odd run time error on Linux Clang?) need to be addressed. While waiting for my local Linux build to repro, going to let this CI run go to completion so Windows caches get saved and see if macOS also has issues that I'll need to address locally.

ADKaster:
The clang error is not actually a runtime error. It's a sanitizer error. 

```
 =================================================================
==20645==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 256 byte(s) in 1 object(s) allocated from:
    #0 0x561bbed55268 in malloc build-llvm/tools/clang/stage2-bins/runtimes/runtimes-bins/compiler-rt/lib/asan/asan_malloc_linux.cpp:67:3
    #1 0x7ffa9559bd08 in FcPatternObjectInsertElt fcpat.c

Indirect leak of 64 byte(s) in 2 object(s) allocated from:
    #0 0x561bbed5543d in calloc build-llvm/tools/clang/stage2-bins/runtimes/runtimes-bins/compiler-rt/lib/asan/asan_malloc_linux.cpp:74:3
    #1 0x7ffa9559c890 in FcPatternObjectAddWithBinding fcpat.c

SUMMARY: AddressSanitizer: 320 byte(s) leaked in 3 allocation(s).
Sanitizer errors happened while running tests; see the Test step above.
```

Both of those fontconfig errors are supposed to be suppressed. Last time I ran into this it was an issue with the default `llvm-symbolizer` binary not being found, and not matching the backtraces properly.

ayeteadoe:
Yeah I forgot to look at the sanitizer output section in CI, but I saw it was sanitizer issue when I repro'd locally.

Given symbol export for `LibGfx` was just an extra and not the core purpose of this PR, I'm just confirming locally that removing that commit makes Linux happy. If so, I could remove that commit and start a new draft PR where this Linux LSan issue can be properly resolved

PR #5772, opened 2025-08-07
LibWeb: Enforce Integrity Policy on Fetch requests

kennethmyhra:


Libraries/LibWeb/ContentSecurityPolicy/BlockingAlgorithms.cpp
```diff
@@ -131,6 +135,66 @@ Directives::Directive::Result should_request_be_blocked_by_content_security_poli
...
+    if (!is<HTML::Window>(global) || !is<HTML::WorkerGlobalScope>(global))
+        return Directives::Directive::Result::Allowed;
+
+    // 10. Let block be a boolean, initially false.
+    bool block = false;
+
+    // FIXME: 11. Let reportBlock be a boolean, initially false.
+    [[maybe_unused]] auto report_block = false;
+
+    // 12 If policy’s sources contains "inline" and policy’s blocked destinations contains request’s destination, set block to true.
```
AtkinsSJ: Nit: Missing the `.` after 12.

Otherwise this commit looks good to me, though maybe Luke will want to take a look once it's undrafted.

Libraries/LibWeb/ContentSecurityPolicy/BlockingAlgorithms.cpp
```diff
@@ -131,6 +135,66 @@ Directives::Directive::Result should_request_be_blocked_by_content_security_poli
...
+        return Directives::Directive::Result::Allowed;
+
+    // 10. Let block be a boolean, initially false.
+    bool block = false;
+
+    // FIXME: 11. Let reportBlock be a boolean, initially false.
+    [[maybe_unused]] auto report_block = false;
+
+    // 12. If policy’s sources contains "inline" and policy’s blocked destinations contains request’s destination, set block to true.
+    if (policy.sources.contains_slow("inline"sv)
```
Lubrsi, line 181: question: Should this be a case insensitive match? Ditto for report_policy below.

Libraries/LibWeb/ContentSecurityPolicy/BlockingAlgorithms.cpp
```diff
@@ -131,6 +135,66 @@ Directives::Directive::Result should_request_be_blocked_by_content_security_poli
...
+
+    // 7. If both policy and reportPolicy are empty integrity policys, return "Allowed".
+    if (policy.is_empty() && report_policy.is_empty())
+        return Directives::Directive::Result::Allowed;
+
+    // 8. Let global be request’s client’s global object.
+    auto& global = request->client()->global_object();
+
+    // 9. If global is not a Window nor a WorkerGlobalScope, return "Allowed".
+    if (!is<HTML::Window>(global) || !is<HTML::WorkerGlobalScope>(global))
```
Lubrsi: issue: This will always return true, since a global can't be both types.
suggestion: Replace `||` with `&&`
```suggestion
    if (!is<HTML::Window>(global) && !is<HTML::WorkerGlobalScope>(global))
```

Lubrsi (changes requested):

Libraries/LibWeb/ContentSecurityPolicy/BlockingAlgorithms.cpp
kennethmyhra: This should probably be a case-sensitive match. From what I can see spec steps usually specifically say `case-insensitive match` if that is the case.

~~I have changed this part and the one below to use the `Vector::first_matching(predicate)` instead.~~

Libraries/LibWeb/ContentSecurityPolicy/BlockingAlgorithms.cpp
kennethmyhra: Not sure what I was thinking here, amended per your suggestion.

Libraries/LibWeb/ContentSecurityPolicy/BlockingAlgorithms.cpp
kennethmyhra: Looked further up if this should be `case-insensitive` or `case-sensitive`.

The infra spec states:
> A string a is or is identical to a string b if it consists of the same sequence of code units.
> 
> Except where otherwise stated, all string comparisons use is.

https://infra.spec.whatwg.org/#string-is

Lubrsi (approved):

PR #5773, opened 2025-08-08
LibWeb: Improve support for CSVs in transform functions

Calme1709:
This PR improves support for CalculatedStyleValues in `scale` and `translate` CSS transform functions - see individual commits for details.

I also added a couple of formatters which I have found useful for debugging.

Gains us 75 WPT tests.


Libraries/LibWeb/CSS/CSSStyleValue.h
```diff
@@ -465,3 +465,189 @@ struct AK::Formatter<Web::CSS::CSSStyleValue> : Formatter<StringView> {
         return Formatter<StringView>::format(builder, style_value.to_string(Web::CSS::SerializationMode::Normal));
     }
 };
+
+template<>
+struct AK::Formatter<Web::CSS::CSSStyleValue::Type> : Formatter<StringView> {
```
tcl3, line 251: Elsewhere, this sort of thing has been done with an X macro to avoid having to change code in multiple places when adding a value.

Something like:
```c++
#define ENUMERATE_CSS_STYLE_VALUE_TYPES                             \
    __ENUMERATE_STYLE_VALUE_TYPE(Anchor)                              \
    __ENUMERATE_STYLE_VALUE_TYPE(AnchorSize)                          \
    __ENUMERATE_STYLE_VALUE_TYPE(Angle)                               \
    (etc...)
```
```c++
enum class Type {
#define __ENUMERATE_STYLE_VALUE_TYPE(x) x,
    ENUMERATE_CSS_STYLE_VALUE_TYPES
#undef __ENUMERATE_STYLE_VALUE_TYPE
};
```
```c++
ErrorOr<void> format(FormatBuilder& builder, Web::CSS::CSSStyleValue::Type type)
{
    StringView type_name;
    switch (type) {
#define __ENUMERATE_STYLE_VALUE_TYPE(x) case Web::CSS::CSSStyleValue::Type::x: type_name = #x##sv; break;
        ENUMERATE_CSS_STYLE_VALUE_TYPES
#undef __ENUMERATE_STYLE_VALUE_TYPE
    }

    return Formatter<StringView>::format(builder, type_name);
}
```
(I've not tested this at all, so it may be wrong :sweat_smile:)

Also, it might be a good idea to have an "unknown" state, or a `VERIFY_NOT_REACHED()` if you see a value you don't expect.

Calme1709: Done - I also converted the `is_{type}` and `as_{type}` methods to use this new macro as well, see https://github.com/LadybirdBrowser/ladybird/pull/5773/commits/bb6364033e52182c3728b00cdc2822d9450bb060

AtkinsSJ: We used to have this, and I removed it [here](https://github.com/LadybirdBrowser/ladybird/commit/2e1f62681cf8c11b00eccf041aabe2d4bd5c2941). BUT, I was wrong to do so because I incorrectly thought we could make our StyleValue types match typed-om directly.

AtkinsSJ: Oh and I'm *very* happy to see this formatter, I've wanted this so often!

AtkinsSJ (approved):

PR #5781, opened 2025-08-08
LibWeb: Ignore all `CharacterData` nodes when evaluating filters

tcl3:
This prevents an unknown filter debug message would be printed for comments.


awesomekling:
Could also swap out `for_each_child` for `for_each_child_of_type<DOM::Element>`

tcl3:
> Could also swap out `for_each_child` for `for_each_child_of_type<DOM::Element>`

Changed to do this instead :)

PR #4958, opened 2025-06-02
LibWGSL+LibWebGPUNative+LibWeb/WebGPU: Implement Vulkan backend for rendering a triangle 

ayeteadoe:
Some WIP infrastructure that implements the [Sokol "clear" example](https://floooh.github.io/sokol-html5/clear-sapp.html) in WebGPU with the initial goal of building up to rendering the standard "Hello World Triangle".

There is still a decent amount of things to cleanup/improve here and FIXMEs that need to be addressed; however, I wanted to put it up before I started on that to get initial feedback/thoughts in case there isn't interest in my approach being upstreamed.

Main things left before this is no longer a draft:

- [x] Create a new library, LibWGSL, that compiles the simple triangle shader WGSL source to SPIR-V, or MSL for use in the Vulkan, or Metal backends respectively
- [ ] Create a new utility, `wgslc` (`Utilities/wgslc.cpp`), inspired by [glslc](https://github.com/google/shaderc/tree/main/glslc) that wraps `WGSL::Compiler`. This will allow for a more robust testsuite of actual `.wgsl` shaders and the expected SPIR-V/MSL outputs checked into `Tests/LibWGSL` instead of embedding it all into the `TestWGSLCompiler` binary, given over time we'll want to test hundreds/thousands of scenarios so the current system would end up with massive binary size bloat.
- [ ] Implement enough of the spec such that it can do everything that [this sample](https://mdn.github.io/dom-examples/webgpu-render-demo/) can
- [ ] Use more Vk-Bootstrap for all relevant initializations
- [ ] Use [vulkan-memory-allocator](https://vcpkg.link/ports/vulkan-memory-allocator) helper 
- [ ] Make queue submission asynchronous
- [ ] Use double buffering in `GPUCanvasContext`
- [ ] Add support for [OffscreenCanvas](https://html.spec.whatwg.org/multipage/canvas.html#offscreencanvas) in `GPUCanvasContext`
- [ ] Add proper spec comments
- [ ] Ensure any other low hanging fruit wrt optimizations is handled
- [ ] Get Vulkan SDK working in macOS CI

### Ladybird WebGPU Clear Demo

[webgpu-clear-demo.webm](https://github.com/user-attachments/assets/86c210dc-92f9-4f47-933f-246b116058b9)

PR #5783, opened 2025-08-09
Meta+LibWeb/WebGPU: Implement initialization via Dawn

ayeteadoe:
Second attempt (see initial work [here](https://github.com/LadybirdBrowser/ladybird/pull/4958)) to start implementing WebGPU in Ladybird. This time, instead of implementing our own homegrown WGSL compiler and backend-agnostic Native GPU API layer, I opted for using Google's dawn library, which is a WebGPU native implementation. Just like with our usage of angle and skia, this allows for focusing on the actual `LibWeb/WebGPU` layer of the work and lowers the maintenance burden.

What I have so far seems like a good initial checkpoint for feedback but in future work, if this is accepted, I plan on resurrecting the remaining  `LibWeb/WebGPU` portions of my initial attempt and getting to the point where we can start to render various 2D and 3D geometries. I won't add any tests to LibWeb until CI has GPU-enabled machines or if we want to use Swiftshader in CI instead (dawn has support for that), but there is a [WebGPU Conformance Testsuite](https://github.com/gpuweb/cts) that can be used to track and gradually build up support.

## Demo

The following web page was used to produce these demos:

<details>

<summary>webgpu-init.html</summary>

```html
<!DOCTYPE html>
<html>

<head>
    <title>Ladybird WebGPU: Initialization</title>
</head>

<body>
<div id="output">
    <h2>WebGPU Instance Information:</h2>
    <div id="instance-info"></div>

    <h2>WebGPU Adapter Information:</h2>
    <div id="adapter-info"></div>

    <h2>WebGPU Device Information:</h2>
    <div id="device-info"></div>
</div>

<script>
    async function initWebGPU() {
        let gpu = null;

        const instanceInfoText = document.getElementById("instance-info");
        try {
            gpu = navigator.gpu;
            if (!gpu) {
                throw Error("WebGPU not supported");
            }

            instanceInfoText.innerHTML += `<p><strong>Preferred Canvas Format:</strong> ${gpu.getPreferredCanvasFormat()}</p>`;
            instanceInfoText.innerHTML += `<p><strong>WGSL Language Features:</strong> ${[...gpu.wgslLanguageFeatures].join(", ")}</p>`;
        } catch (error) {
            instanceInfoText.innerHTML += `<p style="color: red;"><strong>Error:</strong> ${error.message}</p>`;
            return;
        }

        let adapter = null;
        const adapterInfoText = document.getElementById("adapter-info");
        try {
            adapter = await navigator.gpu.requestAdapter({
                powerPreference: "high-performance",
            });
            if (!adapter) {
                throw Error("Unable to retrieve WebGPU adapter");
            }

            const info = adapter.info;
            adapterInfoText.innerHTML += `<p><strong>Vendor:</strong> ${info.vendor}</p>`;
            adapterInfoText.innerHTML += `<p><strong>Architecture:</strong> ${info.architecture}</p>`;
            adapterInfoText.innerHTML += `<p><strong>Device:</strong> ${info.device}</p>`;
            adapterInfoText.innerHTML += `<p><strong>Description:</strong> ${info.description}</p>`;
            adapterInfoText.innerHTML += `<p><strong>Subgroup (Min Size):</strong> ${info.subgroupMinSize}</p>`;
            adapterInfoText.innerHTML += `<p><strong>Subgroup (Max Size):</strong> ${info.subgroupMaxSize}</p>`;
            adapterInfoText.innerHTML += `<p><strong>Supported Features:</strong> ${[...adapter.features].join(", ")}</p>`;

            const limits = adapter.limits;
            const limitsData = [
                ["maxTextureDimension1D", limits.maxTextureDimension1D],
                ["maxTextureDimension2D", limits.maxTextureDimension2D],
                ["maxTextureDimension3D", limits.maxTextureDimension3D],
                ["maxTextureArrayLayers", limits.maxTextureArrayLayers],
                ["maxBindGroups", limits.maxBindGroups],
                ["maxBindGroupsPlusVertexBuffers", limits.maxBindGroupsPlusVertexBuffers],
                ["maxBindingsPerBindGroup", limits.maxBindingsPerBindGroup],
                ["maxDynamicUniformBuffersPerPipelineLayout", limits.maxDynamicUniformBuffersPerPipelineLayout],
                ["maxDynamicStorageBuffersPerPipelineLayout", limits.maxDynamicStorageBuffersPerPipelineLayout],
                ["maxSampledTexturesPerShaderStage", limits.maxSampledTexturesPerShaderStage],
                ["maxSamplersPerShaderStage", limits.maxSamplersPerShaderStage],
                ["maxStorageBuffersPerShaderStage", limits.maxStorageBuffersPerShaderStage],
                ["maxStorageTexturesPerShaderStage", limits.maxStorageTexturesPerShaderStage],
                ["maxUniformBuffersPerShaderStage", limits.maxUniformBuffersPerShaderStage],
                ["maxUniformBufferBindingSize", limits.maxUniformBufferBindingSize],
                ["maxStorageBufferBindingSize", limits.maxStorageBufferBindingSize],
                ["minUniformBufferOffsetAlignment", limits.minUniformBufferOffsetAlignment],
                ["minStorageBufferOffsetAlignment", limits.minStorageBufferOffsetAlignment],
                ["maxVertexBuffers", limits.maxVertexBuffers],
                ["maxBufferSize", limits.maxBufferSize],
                ["maxVertexAttributes", limits.maxVertexAttributes],
                ["maxVertexBufferArrayStride", limits.maxVertexBufferArrayStride],
                ["maxInterStageShaderVariables", limits.maxInterStageShaderVariables],
                ["maxColorAttachments", limits.maxColorAttachments],
                ["maxColorAttachmentBytesPerSample", limits.maxColorAttachmentBytesPerSample],
                ["maxComputeWorkgroupStorageSize", limits.maxComputeWorkgroupStorageSize],
                ["maxComputeInvocationsPerWorkgroup", limits.maxComputeInvocationsPerWorkgroup],
                ["maxComputeWorkgroupSizeX", limits.maxComputeWorkgroupSizeX],
                ["maxComputeWorkgroupSizeY", limits.maxComputeWorkgroupSizeY],
                ["maxComputeWorkgroupSizeZ", limits.maxComputeWorkgroupSizeZ],
                ["maxComputeWorkgroupsPerDimension", limits.maxComputeWorkgroupsPerDimension],
            ];

            adapterInfoText.innerHTML += `<p><strong>Supported Limits:</strong></p>`;
            let limitsTable = "<div style=\"margin-left: 20px;\"><table><tbody>";

            for (const [name, value] of limitsData) {
                limitsTable += `<tr><td>${name}</td><td>${value !== undefined ? value.toLocaleString() : "undefined"}</td></tr>`;
            }

            limitsTable += "</tbody></table></div>";
            adapterInfoText.innerHTML += limitsTable;
        } catch (error) {
            adapterInfoText.innerHTML += `<p style="color: red;"><strong>Error:</strong> ${error.message}</p>`;
            return;
        }

        let device = null;
        const deviceInfoText = document.getElementById("device-info");
        device = await adapter.requestDevice();
        if (!device) {
            throw Error("Unable to retrieve WebGPU device");
        }

        deviceInfoText.innerHTML += `<p><strong>Queue:</strong> ${device.queue}</p>`;
    }

    initWebGPU();
</script>
</body>

</html>
```

</details>

### Windows
https://github.com/user-attachments/assets/29bb2de6-1b46-4601-a758-efbbfc4e696b

### macOS
https://github.com/user-attachments/assets/aa8291f1-bf87-4306-ab1a-952a3815a4f0

### Linux
[webgpu-init-linux.webm](https://github.com/user-attachments/assets/a06852e9-25d3-4ed0-a15c-5d97a20ed449)


vcpkg.json
```diff
@@ -1,5 +1,5 @@
 {
-  "builtin-baseline": "b509a07261b982f35c663bf638aae5f77877d207",
+  "builtin-baseline": "64d355db6c46fc61013039164a3534d70b5de8de",
```
ayeteadoe: Dawns `BUILD.gn` files use a built-in function that didn't exist on the latest `vcpkg-tool-gn` vcpkg port, so I updated it to the latest `gn` and got that upstreamed: https://github.com/microsoft/vcpkg/pull/46818

.github/actions/setup/action.yml
```diff
@@ -66,6 +67,11 @@ runs:
       with:
         xcode-version: 16.3
 
+    - name: Setup Xcode active developer directory
+      if: ${{ inputs.os_name == 'macOS' }}
+      shell: bash
+      run: xcode-select -s /Applications/Xcode_16.3.0.app/Contents/Developer
+
```
ayeteadoe: Added this to attempt to fix the error in macOS CI (I don't have this issue locally) during the dawn ports `vcpkg_gn_configure()` call:

<details>
<summary>generate-arm64-osx-dynamic-rel-out.log</summary>

```
ERROR at //build/config/mac/mac_sdk.gni:133:19: Script returned non-zero exit code.
  _mac_sdk_result = exec_script(script_name, sdk_info_args, "scope")
                    ^----------
  Current dir: /Users/actions-runner/actions-runner/_work/ladybird/ladybird/Build/vcpkg/buildtrees/dawn/arm64-osx-dynamic-rel/
  Command: python3 /Users/actions-runner/actions-runner/_work/ladybird/ladybird/Build/vcpkg/buildtrees/dawn/src/e1d6059abf-00be5ebb5d.clean/build/config/apple/sdk_info.py macosx
  Returned 1.
  stderr:
  
  xcode-select: error: tool 'xcodebuild' requires Xcode, but active developer directory '/Library/Developer/CommandLineTools' is a command line tools instance
  Traceback (most recent call last):
    File "/Users/actions-runner/actions-runner/_work/ladybird/ladybird/Build/vcpkg/buildtrees/dawn/src/e1d6059abf-00be5ebb5d.clean/build/config/apple/sdk_info.py", line 167, in <module>
      sys.exit(main())
               ^^^^^^
    File "/Users/actions-runner/actions-runner/_work/ladybird/ladybird/Build/vcpkg/buildtrees/dawn/src/e1d6059abf-00be5ebb5d.clean/build/config/apple/sdk_info.py", line 152, in main
      FillXcodeVersion(settings, args.developer_dir)
    File "/Users/actions-runner/actions-runner/_work/ladybird/ladybird/Build/vcpkg/buildtrees/dawn/src/e1d6059abf-00be5ebb5d.clean/build/config/apple/sdk_info.py", line 61, in FillXcodeVersion
      lines = subprocess.check_output(['xcodebuild',
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/subprocess.py", line 466, in check_output
      return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/subprocess.py", line 571, in run
      raise CalledProcessError(retcode, process.args,
  subprocess.CalledProcessError: Command '['xcodebuild', '-version']' returned non-zero exit status 1.
  
  See //build/gn_logs.gni:32:3: whence it was imported.
    import("//build/config/mac/mac_sdk.gni")
    ^--------------------------------------
  See //BUILD.gn:130:3: whence it was imported.
    import("//build/gn_logs.gni")
    ^---------------------------
```

</details>

But the fix for this error recommend from https://apple.stackexchange.com/questions/397221/error-xcodebuild-requires-xcode didn't seem to work. Note that we can't do `sudo xcode-select ...` in this context. From the [macos-latest runner image](https://github.com/actions/runner-images/blob/main/images/macos/macos-15-arm64-Readme.md#xcode) it looks like I have the right Xcode path, but I'm sure the [Setup Xcode action](https://github.com/marketplace/actions/setup-xcode-version) we use above is doing this too.

I was thinking of just patching the dawn gn configure step to not use the system xcode and use the bundled/hermetic version, but it looks like they explicitly advise against that due to [this build override](https://dawn.googlesource.com/dawn/+/refs/heads/chromium/7258/build_overrides/build.gni#32). I tried anyways just to see, but it looks like it really is internal only, as attempting to run `mac_toolchain.py` to download the binaries is a no-op due to [these checks](https://chromium.googlesource.com/chromium/src/build/+/db0d31d702840881c049c523a2226e8e391929bf/mac/should_use_hermetic_xcode.py), and attempting to bypass the checks ends up needing to invoke `cipd`, so this option would probably fair even worse in CI.


After a bit more digging, I think setting the `DEVELOPER_DIR` environment variable before vcpkg install might be the solution. Update: That didn't work either 😞, going to revert that for now.


I see the chromium-base port had the same error on osx when it was first added: https://github.com/microsoft/vcpkg/pull/12396#issuecomment-657375152. I think the next thing I will try is to adapt the mac-specific portions of [this chromium-base patch](https://github.com/microsoft/vcpkg/blame/master/ports/chromium-base/res/0002-build.patch) into a patch for the chromium/build equivalent. Update: That was the fix 🎉 !

Meta/CMake/vcpkg/overlay-ports/dawn/portfile.cmake
```diff
@@ -0,0 +1,285 @@
...
+
+libxrandr-dev libxinerama-dev libxcursor-dev mesa-common-dev libx11-xcb-dev pkg-config
+
+They can be installed on Debian based systems via
+
+apt-get install libxrandr-dev libxinerama-dev libxcursor-dev mesa-common-dev libx11-xcb-dev pkg-config
+]]
+    )
+endif()
+
```
ayeteadoe: This portfile is mostly inspired by the [v8 portfile](https://github.com/microsoft/vcpkg/blob/master/ports/v8/portfile.cmake)

cqundefine:
Fwiw it's gonna break Android and FreeBSD builds, but I'm not sure how much that matters

ayeteadoe:
@Olekoop do you think you could try to see if Android can build this dawn port as-is? My assumption is yes given the [chromium build infra used to build dawn does support android](https://chromium.googlesource.com/chromium/src/build/+/db0d31d702840881c049c523a2226e8e391929bf/android/). Also @cqundefine you could try FreeBSD too, although I'm more confident in the former being relatively trivial to get working than the latter

cqundefine:
@ayeteadoe Yeah, it might take me a while because I don't have access to my usual setup, but I can take a look at what would it take to get this working on FreeBSD

Libraries/LibWeb/WebGPU/GPUAdapter.cpp
```diff
@@ -0,0 +1,56 @@
...
+#include <webgpu/webgpu_cpp.h>
+
+namespace Web::WebGPU {
+
+GC_DEFINE_ALLOCATOR(GPUAdapter);
+
+struct GPUAdapter::Impl {
+    wgpu::Adapter adapter { nullptr };
+
+    void dump_info() const
```
ayeteadoe: Given this isn't standard, I will adjust this commit to also expose [GPUAdapterInfo](https://www.w3.org/TR/webgpu/#gpuadapterinfo), that way the initial demo/test page can just be displaying the WGSL Language Features and Adapter Info in the body of the document instead of debug logs.

Libraries/LibWeb/WebGPU/GPU.cpp
```diff
@@ -0,0 +1,162 @@
...
+        request_adapter_options.powerPreference = wgpu::PowerPreference::LowPower;
+        break;
+    }
+    default:
+        break;
+    }
+
+    request_adapter_options.forceFallbackAdapter = options.force_fallback_adapter;
+
+    // FIXME: 1. Let contentTimeline be the current Content timeline.
```
ayeteadoe: After a brief read of this [timeline terminology](https://www.w3.org/TR/webgpu/#programming-model-timelines), essentially it seems there are 4 execution contexts:
1. Content is general/standard execution of the script via our LibWeb implementation of the IDL
2. Device is executing GPU setup/config operations (e.g. creating/querying device objects), spec seems to say we could optionally delegate these to a type of rerserved/dedicated "WebGPUWorker" process (maybe a future design consideration?)
3. Queue is executing actual tangible/compute utilizing GPU operations (e.g. drawing) where we have no control, we just wait until it is done
4. Any of the above (Unspecified/agnostic)

I will extend this commit to address these FIXMEs as it should be straight forward given Device timeline can just be done inline here as normal.

cqundefine:
@ayeteadoe I tried getting the build of dawn working on my FreeBSD VM with no luck, the GN build would require way too much patching for me to do right now. 

What I also noticed is that dawn has an official CMake build which we should probably make use of in the overlay port instead of the GN one. It would allow to drop a bunch of patches because everything can be configured with arguments and would eliminate the need to create the custom targets as it provides its own. Using the CMake build would also allow me to get it building on FreeBSD with minimal patches.

ayeteadoe:
@cqundefine based on what I've seen for google/chromium based vcpkg ports, most use the gn build. Also this port would eventually be used for the skia dawn feature (not relevant to Ladybird, but is relevant for eventually getting this upstreamed) and skia uses gn.

Maybe @ADKaster has some thoughts, but in my opinion it would be ideal to just leave FreeBSD support out for now until it is patched to be buildable via gn. Maybe I could add some stub infrastructure so platforms without dawn support can still build LibWeb? But not sure of the value for that if its only FreeBSD that needs the extra work.

.github/actions/setup/action.yml
```diff
@@ -41,7 +41,8 @@ runs:
         sudo apt-get update -y
         sudo apt-get install -y autoconf autoconf-archive automake build-essential ccache cmake curl fonts-liberation2 \
             gcc-14 g++-14 libcurl4-openssl-dev libegl1-mesa-dev libgl1-mesa-dev libpulse-dev libssl-dev \
-            libstdc++-14-dev lld-20 llvm-20 nasm ninja-build qt6-base-dev qt6-tools-dev-tools tar unzip zip
+            libstdc++-14-dev lld-20 llvm-20 libxrandr-dev libxinerama-dev libxcursor-dev mesa-common-dev \
+            libx11-xcb-dev nasm ninja-build qt6-base-dev qt6-tools-dev-tools tar unzip zip
```
ADKaster: I'd really rather avoid x11 dependencies... Can't we configure dawn to use wayland only instead?

Libraries/LibWeb/WebGPU/GPU.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+// https://www.w3.org/TR/webgpu/#dom-gpu-requestadapter
+GC::Ref<WebIDL::Promise> GPU::request_adapter(GPURequestAdapterOptions const& options)
+{
+    wgpu::RequestAdapterOptions request_adapter_options = {};
+
+    // https://www.w3.org/TR/webgpu/#dom-gpurequestadapteroptions-featurelevel
+    if (options.feature_level == "core"_string) {
+        request_adapter_options.featureLevel = wgpu::FeatureLevel::Core;
+    } else if (options.feature_level == "compatibility") {
```
ADKaster: generally preferred to use _sv literals for constant strings like this.

Libraries/LibWeb/WebGPU/GPU.cpp
ADKaster: presumably that waitAny call below is related to this? At the very least, we should do the hand-wavy "lmao run it on the event loop later" thing we've done for other actually-async items and return the promise to JS right away.

Libraries/LibWeb/WebGPU/GPU.cpp
```diff
@@ -0,0 +1,162 @@
...
+    request_adapter_options.forceFallbackAdapter = options.force_fallback_adapter;
+
+    // FIXME: 1. Let contentTimeline be the current Content timeline.
+
+    // 2. Let promise be a new promise.
+    auto& realm = this->realm();
+    GC::Ref promise = WebIDL::create_promise(realm);
+
+    // FIXME: 3. Issue the initialization steps on the Device timeline of this.
+    auto fut = m_impl->instance.RequestAdapter(&request_adapter_options, wgpu::CallbackMode::AllowProcessEvents, [&realm, promise](wgpu::RequestAdapterStatus status, wgpu::Adapter adapter, char const* message) {
```
ADKaster: Given the capitalization of RequestAdapter, this is calling into native code from the dawn library impl? In that case, the closure here is gc-unsafe. You must wrap both the realm and the promise in handles.

Libraries/LibWeb/WebGPU/GPUAdapter.cpp
ADKaster: Hooking this up to the WebContent process's `debug_request` infra might be useful in the future.

Libraries/LibWeb/WebGPU/GPU.cpp
ayeteadoe: Ah yeah I wasn't sure if we needed the deferred invoke here or not, I'll add that. And yeah, for `requestAdapater` the Device timeline can start at the beginning of  `Platform::EventLoopPlugin::the().deferred_invoke()`, then we switch back to the Content timeline for resolving/rejecting the promise

Libraries/LibWeb/WebGPU/GPU.cpp
ayeteadoe: Do you know of any examples in LibWeb you could point me to that do this realm and/or promise handle wrapping?

ADKaster: ah in that case I just meant
`[realm = GC::Root(realm), promise = GC::Root(promise)]( ...) { ... };`

ayeteadoe: Ah I see, so to ensure a reference to some object stays alive during the next GC sweep, you `GC::Root` things which will keep the object alive for at least as long as the lifetime of the new `GC::Root`

.github/actions/setup/action.yml
ayeteadoe: So far I've had to patch dawns  `gn` config to allow for X11 and Wayland to be mutually exclusive (right now they always enable X11 even if you set `dawn_use_wayland=true`). Then I started hitting some build issues, `wgpu::Status` being and unqualified-id in `dawn/native/Instance.cpp` and `dawn/native/sources/X11Functions.cpp` still being included in the build. So it will probably take a bit more patching/investigation on my side to get it working, but it should be possible :crossed_fingers: 

Update: Yup, just needed a small patch to fully remove x11 dependency

ayeteadoe:
One more rebase incoming as there are a few things I missed in the dawn vcpkg port (stray debug comments, incomplete usage/description, no features exposed) and I am cleaning that up now

Olekoop:
@ayeteadoe I have tried building the dawn branch for Android but it looks like you are using PkgConfig in order to find dawn's paths. Dawn compiles but Ladybird cannot find it. PkgConfig doesn't work on Android and it will try to use host's libraries (I don't have Dawn installed on my PC and it fails). It would be nice if there would be a cmake find file.

Olekoop:
@ayeteadoe I haven't compiled your PR yet for Android but I believe that finding Dawn via PkgConfig is a no go when compiling for Android. Usually PkgConfig tries to find host's libraries and that will obviously fail. It would be nice if there was a cmake find file. I woudn't mind breaking Android support for now.

cqundefine:
Just wanted to note that I've gotten the dawn port to build using GN on FreeBSD, tho I'm gonna submit the patches after this PR and the vcpkg PR get merged to not complicate stuff.

ayeteadoe:
Going to slightly extend this and implement [requesting the GPUDevice](https://www.w3.org/TR/webgpu/#dom-gpuadapter-requestdevice) as well. That means that this initial PR will cover the full [Initialization](https://www.w3.org/TR/webgpu/#initialization) section of the spec, this will be nice because then a bunch of subsequent work that only depends on the GPUDevice can be done in parallel, split into multiple PRs, as They'd just need to implement their respective IDL methods on GPUDevice and then their own IDL.

The final version of the test/demo page will be 3 sections:
1. WebGPU Instance Information: Will show the preferred canvas format and the WGSL language features
2. WebGPU Adapter Information: Will show all fields from GPUAdapterInfo, GPUSupportedLimits, and GPUSupportedFeatures
3. WebGPU Device Information: Will just show that the GPUQueue was created, all the other attributes should be equivalent to Adapter above (unless there is a difference between limits and features of the physical device (GPUAdapter) vs logical device (GPUDevice), I'll find out soon

Libraries/LibWeb/WebGPU/GPUAdapter.idl
```diff
@@ -0,0 +1,28 @@
...
+    "low-power",
+    "high-performance",
+};
+
+// https://www.w3.org/TR/webgpu/#dictdef-gpurequestadapteroptions
+dictionary GPURequestAdapterOptions {
+    DOMString featureLevel = "core";
+    GPUPowerPreference powerPreference;
+    boolean forceFallbackAdapter = false;
+    // FIXME: Support xrCompatible
```
cqundefine: Shouldn't unsupported fields be written as:
```
[FIXME] boolean xrCompatible;
```

and methods in their respective FIXME way?

Olekoop:
I wanted to note that I have managed to get Dawn compiled and get WebGPU running on Android. I will submit the patches once this PR gets merged. 
Screenshot_20250815-143741_Ladybird

Libraries/LibWeb/Painting/DisplayListPlayerSkia.cpp
```diff
@@ -806,7 +808,9 @@ void DisplayListPlayerSkia::paint_radial_gradient(PaintRadialGradient const& com
     SkMatrix matrix;
     // Skia does not support specifying of horizontal and vertical radius's separately,
     // so instead we apply scale matrix
-    matrix.setScale(size.width() / size.height(), 1.0f, center.x(), center.y());
+    auto const aspect_ratio = size.width() / size.height();
+    auto const sx = isinf(aspect_ratio) ? 1.0f : aspect_ratio;
+    matrix.setScale(sx, 1.0f, center.x(), center.y());
```
ayeteadoe, line 821: Initially after upgrading the test began to look like this:
beforefix-empty-radial-gradient-crash

Then after this fix it went back to the expected value:

afterfix-empty-radial-gradient-crash

ayeteadoe:
@cqundefine @Olekoop See first commit for context into the dawn port changes, but both Android and FreeBSD should be working with the current state of the PR now. If you have time to try it out let me know if there are any issues and I can adjust to get both of them working

Olekoop:
@ayeteadoe I have managed to compile it for Android and it works without needing to modify your code.

ayeteadoe:
The CMake-based dawn port has been upstreamed into vcpkg, so I'll adjust to remove the overlay port given it should be working on all the platforms, but it can always be overlayed again if we run into issues.

cqundefine:
@ayeteadoe I have submitted a vcpkg PR to fix Dawn on FreeBSD, once this gets merged it will most probably break it temporarily but that's fine by me.

vcpkg.json
```diff
@@ -25,6 +25,47 @@
...
+      "name": "dawn",
+      "platform": "android",
+      "default-features": false,
+      "features": [
+        "vulkan"
+      ]
+    },
+    {
+      "name": "dawn",
+      "platform": "freebsd | openbsd",
```
cqundefine: Wayland is also support on FreeBSD and OpenBSD, so maybe this should just get merged with the Linux case, if not this is still a duplicate with Android.

Meta/CMake/vcpkg/overlay-ports/skia/vcpkg.json
```diff
@@ -72,17 +71,25 @@
...
+          "name": "dawn",
+          "default-features": false,
+          "features": [
+            "wayland"
+          ],
+          "platform": "linux"
+        },
+        {
+          "name": "dawn",
+          "platform": "android | freebsd | openbsd | osx | windows"
```
cqundefine: Similar comment here regarding Wayland on *BSD.

cqundefine (commented):
Just a nitpick regarding BSD support. While on that note I've upstreamed the necessary Dawn patches so this should™ just work on FreeBSD.

ayeteadoe:
Going to close this for now as I have some other PRs in the queue I want to focus on first and this one keeps getting merge conflicts. Thanks to all who helped test on Android/BSDs.

I've also filed https://github.com/LadybirdBrowser/ladybird/issues/6514 and will probably wait to re-open this until we get some more explicit discussion involving maintainers on the desired approach/direction to take WebGPU.

PR #5787, opened 2025-08-09
RequestServer: Sanitizing curl GET headers

Ricardev:
cURL is re-using open connections and some GET requests was including  a Content-Length header, which broke subsequent GET/HEAD requests that reused the same connection.

This fixes one WPT test.

Before the changes in this PR the results of ./Meta/WPT.sh run cors was:

> Ran 27 tests finished in 6.3 seconds.
> 	. 18 ran as expected. 0 tests skipped.
> 	. 9 tests had unexpected subtest results

After the changes is:

> Ran 27 tests finished in 6.3 seconds.
> 	.19 ran as expected. 0 tests skipped.
> 	. 8 tests had unexpected subtest results.

### DIFF
--- a/Libraries/LibHTTP/HeaderMap.h
+++ b/Libraries/LibHTTP/HeaderMap.h
@@ -49,6 +49,22 @@ public:
         return m_headers;
     }
 
+    void sanitize_request_headers_for_method(const ByteString& method)
+    {
+        auto method_allows_body = method.is_one_of("POST"sv, "PUT"sv, "PATCH"sv, "DELETE"sv);
+
+        if (!method_allows_body) {
+            m_map.remove_all_matching([](auto const& header_name, auto const&) {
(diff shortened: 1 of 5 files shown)


alimpfard (commented):
I don't see how this is an RS problem, the loader should not be including these headers to begin with...?

Ricardev:
I'm new to the project and I'm not familiar with its components yet...or each one responsability.
So you're suggesting the header Sanitizing should be in Resource Loader instead of RS?

alimpfard:
I'm not sure why this should be necessary at all.
if the bad request comes from Fetch, that's where we should be looking first; if we have some ad-hoc code producing the request...then that's where we should look.

either way we should not be getting any of those headers in any valid request that trickles down to RS.
Any Fetch knowers with more info on this?

PR #5788, opened 2025-08-09
LibWeb: Assume CSP list is empty if document has no associated window

tcl3:
Previously, a crash would occur when checking whether a `<base>` element was allowed by CSP, if the relevant document didn't have an associated window.

I first encountered this in the wild on a BBC website but wasn't easily able to reproduce the problem. I found that this WPT test has the same issue: https://wpt.live/xhr/responsexml-document-properties.htm. 

This test is fixed by this PR, but isn't importable because it relies on a Python file provided by the WPT runner.


tcl3:
This is the same fix as #5871. Closing.

PR #5797, opened 2025-08-09
RequestServer: small improvements

rmg-x:
See commits.

### DIFF
--- a/Services/RequestServer/ConnectionFromClient.cpp
+++ b/Services/RequestServer/ConnectionFromClient.cpp
@@ -511,7 +511,7 @@ void ConnectionFromClient::start_request(i32 request_id, ByteString method, URL:
             struct curl_slist* curl_headers = nullptr;
 
             // NOTE: CURLOPT_POSTFIELDS automatically sets the Content-Type header.
-            //       Set it to empty if the headers passed in don't contain a content type.
+            //       Tell curl to remove it by setting a blank value if the headers passed in don't contain a content type.
             if (did_set_body && !request_headers.contains("Content-Type"))
                 curl_headers = curl_slist_append(curl_headers, "Content-Type:");
 
--- a/Services/RequestServer/ConnectionFromClient.cpp
+++ b/Services/RequestServer/ConnectionFromClient.cpp
@@ -6,7 +6,6 @@
 
 #include "WebSocketImplCurl.h"
 
-#include <AK/Badge.h>
 #include <AK/IDAllocator.h>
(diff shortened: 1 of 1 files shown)


i-love-coffee-i-love-tea:
Could you please explain why you think the comment is more accurate than before?
That change doesn't look right to me. It does what the comment said before?

rgret-dev:
@i-love-coffee-i-love-tea the final result isn't that the header is _empty_ (e.g. `Content-Type: `), it will be completely removed from the request. This way it's more clear especially given the block of code below this which _does_ set empty headers.

i-love-coffee-i-love-tea:
Thank you. It really is true. Curl doesn't send it. I had to make small test case and verify it. :smile:

InvalidUsernameException:
I think neither formulation is really clear. Maybe something like this could work:

```c++
// NOTE: CURLOPT_POSTFIELDS automatically sets the Content-Type header.
//       Setting it to empty if the headers passed in don't contain a content type
//       prevents curl from doing so and causes the header to be absent entirely.
```

rgret-dev:
@InvalidUsernameException it doesn't prevent curl from doing anything, the header is already there. This line just removes it.

i-love-coffee-i-love-tea:
I did some more testing.

```
headers = curl_slist_append(headers, "Content-Type:");
```
without a Content-Type header present, will not send an empty Content-Type header, but
if a Content-Type header is already present, it will not be removed by this.



Example: if you do
```
headers = curl_slist_append(headers, "Content-Type: bla");
headers = curl_slist_append(headers, "Content-Type:");
```

The bla header will still be there. It is not removed. If that is what is supposed to happen, it's fine.

rgret-dev:
It seems this only applies to internally added headers from within libcurl. So it's there but will prioritize whatever you override it with.

https://curl.se/libcurl/c/CURLOPT_HTTPHEADER.html
> If you provide a header that is otherwise generated and used by libcurl internally, your header alternative is used instead. If you provide a header without content (no data on the right side of the colon) as in Accept:, the internally used header is removed. To forcibly add a header without content (nothing after the colon), use the form name; (using a trailing semicolon).

InvalidUsernameException:
I guess then I misunderstood _when_ the actual logic in curl executes. Maybe something like this then?

```c++
// NOTE: CURLOPT_POSTFIELDS has automatically set an undesired Content-Type header already.
//       Instruct curl to remove it by setting it to empty if the headers passed in don't contain a content type.
```

PR #5803, opened 2025-08-09
Meta: Add lint-clang-format.py

ayeteadoe:


Meta/lint-clang-format.py
```diff
@@ -0,0 +1,134 @@
...
+        sys.exit(1)
+
+    print(f"Using {clang_format}")
+
+    platform = Platform()
+    if platform.host_system == HostSystem.Windows:
+        # Windows doesn't seem to be able to handle executing a command this long, even when
+        # HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\FileSystem\LongPathsEnabled is set to 1. So we'll execute
+        # the files in chunks of 10 at a time to ensure regardless of how much the codebase scales this command will
+        # succeed
```
InvalidUsernameException: I don't think this has anything to do with long path support and you are just looking at the general command line length limitations, [relevant Raymond Chen post](https://devblogs.microsoft.com/oldnewthing/?p=41553).

Technically there should also be a limit on Linux (given by `getconf ARG_MAX`), but it is so large that you usually don't hit it.

Meta/lint-clang-format.py
```diff
@@ -0,0 +1,134 @@
...
+        batch_size = 10
```
InvalidUsernameException: Could just define `batch_size = len(files)` for non-Windows and then run the same code on all platforms.

Meta/lint-ci.sh
```diff
@@ -44,8 +44,8 @@ else
     echo -e "[${GREEN}SKIP${NC}]: IPCMagicLinter (in Meta/lint-ci.sh)"
 fi
 
-if Meta/lint-clang-format.sh --overwrite-inplace "$@" && git diff --exit-code -- ':*.cpp' ':*.h' ':*.mm'; then
-    echo -e "[${GREEN}OK${NC}]: Meta/lint-clang-format.sh"
+if Meta/lint-clang-format.py --overwrite-inplace "$@" && git diff --exit-code -- ':*.cpp' ':*.h' ':*.mm'; then
+    echo -e "[${GREEN}OK${NC}]: Meta/lint-clang-format.py"
 else
     echo -e "[${BOLD_RED}FAIL${NC}]: Meta/lint-clang-format.sh"
```
InvalidUsernameException: You missed one rename here.

Meta/lint-clang-format.py
InvalidUsernameException, line 1: It there a reason to keep the `.sh` version around? Afaik, this script isn't really supposed to be called directly like `Meta/ladybird.sh` was, so a "transition period" shouldn't be necessary.

InvalidUsernameException, line 1: There is a reference to the `.sh` version in `Documentation/CodingStyle.md`. Should that be updated as well?

Meta/lint-clang-format.py
ayeteadoe: Good point, I should be more specific in the comment: I originally mentioned long paths in the comment as the specific error that occurred was:

> FileNotFoundError: [WinError 206] The filename or extension is too long

So because of the inaccurate/poor error reporting, I assumed one of the typical fixes for long path errors would resolve things, but then I discovered it was just a command line length problem. I will update the comment

Meta/lint-clang-format.py
ayeteadoe: I agree it can probably go, but I'll wait for feedback from the maintainers before deleting it

AtkinsSJ: +! to removing the bash script, we don't need both.

PR #5806, opened 2025-08-10
Meta: allow package manager disable vulkan

uis246:


uis246:
Better(and correct) solution probably would be to somehow detect Vulkan support in skia instead

uis246:
Any plans to merge simple pr?

PR #5807, opened 2025-08-10
Implement Serial.requestPort method & starting infrastructure for next WebSerial steps

EdwinHoksberg:
WebSerial API can now be enabled with the `--webserial-device-path <path>` command line argument. I've chosen for this option since it would be the simplest way of enabling the WebSerial policy(point 2. of requestPort) and letting the user chose a port without creating a whole new UI component for this(point 4. in requestPort).

I've opted for [ami-iit/serial_cpp](https://github.com/ami-iit/serial_cpp) as a library to handle all serial communication since it looks to have all features we need to further implement the WebSerial API.

(cc @gmta)


Libraries/LibWeb/Serial/SerialPort.h
```diff
@@ -92,10 +96,12 @@ class SerialPort : public DOM::EventTarget {
     virtual void visit_edges(Cell::Visitor&) override;
 
 private:
-    explicit SerialPort(JS::Realm&);
+    explicit SerialPort(JS::Realm&, serial_cpp::PortInfo m_device);
```
gmta: This no longer needs to be `explicit`

Libraries/LibWeb/Serial/SerialPort.cpp
```diff
@@ -11,8 +11,11 @@
 
 namespace Web::Serial {
 
-SerialPort::SerialPort(JS::Realm& realm)
+GC_DEFINE_ALLOCATOR(SerialPort);
+
+SerialPort::SerialPort(JS::Realm& realm, serial_cpp::PortInfo device)
     : DOM::EventTarget(realm)
+    , m_device(device)
```
gmta: `, m_device(move(device))`

Libraries/LibWebView/Application.cpp
```diff
@@ -172,6 +173,7 @@ ErrorOr<void> Application::initialize(Main::Arguments const& arguments)
     args_parser.add_option(dns_server_port, "Set the DNS server port", "dns-port", 0, "port (default: 53 or 853 if --dot)");
     args_parser.add_option(use_dns_over_tls, "Use DNS over TLS", "dot");
     args_parser.add_option(validate_dnssec_locally, "Validate DNSSEC locally", "dnssec");
+    args_parser.add_option(webserial_device_path, "Enables the WebSerial API and sets device to use", "webserial-device-path", 0, "path");
```
gmta, line 177: I imagine that it's not the goal to use this command line argument, but rather let serial_cpp enumerate available ports and let the user select/confirm the one to use. Could you add a FIXME somewhere to that effect?

EdwinHoksberg: Indeed, a device picker like Chrome has would be really neat to have. I added a NOTE for this at https://github.com/LadybirdBrowser/ladybird/pull/5807/files#diff-2bea0dcc11008cdc88cc922f08d9383be7131891361a869cf186bd3a8782a54cR101, would you like me to reword that a bit and change it into a FIXME?

gmta: That existing comment is fine, in the long term I suspect we want this command line argument to be gone - so a FIXME near here is an additional reminder to refactor this.

Libraries/LibWeb/Serial/SerialPort.h
EdwinHoksberg: Fixed

Libraries/LibWeb/Serial/SerialPort.cpp
EdwinHoksberg: Fixed

Libraries/LibWebView/Application.cpp
EdwinHoksberg: Makes sense, I've added a FIXME.

gmta:
@EdwinHoksberg This unfortunately ran into merge conflicts, could you update it so I can give it a test drive?

PR #5811, opened 2025-08-10
LibWeb: Implement `StorageManager::estimate()`

tcl3:
This asks the StorageManager to return how much storage is currently being used and how much space is available.

I've only exposed the `estimate()` method to `Window` currently, because I couldn't find an easy way to make `obtain_a_storage_shelf()` work for Workers.

This makes https://windows96.net report the available storage correctly:

Before:
image

After:
image


Libraries/LibWeb/StorageAPI/StorageManager.cpp
```diff
@@ -29,4 +33,75 @@ void StorageManager::initialize(JS::Realm& realm)
...
+    // 2. Let global be this’s relevant global object.
+    auto& global = HTML::relevant_global_object(*this);
+
+    // 3. Let shelf be the result of running obtain a local storage shelf with this’s relevant settings object.
+    auto& settings = HTML::relevant_settings_object(*this);
+    auto shelf = obtain_a_storage_local_shelf(settings);
+
+    // 4. If shelf is failure, then reject promise with a TypeError.
+    if (!shelf) {
+        WebIDL::reject_promise(realm, promise, JS::TypeError::create(realm, "no u"_string));
```
shannonbooth: no u

seems to be left over :P

Libraries/LibWeb/StorageAPI/StorageManager.cpp
```diff
@@ -29,4 +33,75 @@ void StorageManager::initialize(JS::Realm& realm)
...
+    }
+    // 5. Otherwise, run these steps in parallel:
+    else {
+        Platform::EventLoopPlugin::the().deferred_invoke(GC::create_function(realm.heap(), [&realm, &global, shelf = GC::Root { *shelf }, promise = GC::Root<WebIDL::Promise> { promise }]() {
```
shannonbooth: we should not need to root these since it is captured by GC function

Libraries/LibWeb/StorageAPI/StorageManager.cpp
```diff
@@ -29,4 +33,75 @@ void StorageManager::initialize(JS::Realm& realm)
...
+            });
+        }));
+    }
+
+    // 6. Return promise.
+    return promise;
+}
+
+// https://storage.spec.whatwg.org/#obtain-a-local-storage-shelf
+GC::Ptr<StorageShelf> StorageManager::obtain_a_storage_local_shelf(HTML::EnvironmentSettingsObject& settings)
```
shannonbooth: This seems to have local in a weird spot in the function name, and I am not sure this is the best file for this to live

Libraries/LibWeb/StorageAPI/StorageManager.cpp
tcl3: Was debugging and forgot to change this back to something sensible. Oops :sweat_smile:

Libraries/LibWeb/StorageAPI/StorageManager.cpp
tcl3: Not sure how I missed that. Updated.

Libraries/LibWebView/StorageJar.cpp
```diff
@@ -32,7 +32,7 @@ ErrorOr<NonnullOwnPtr<StorageJar>> StorageJar::create(Database& database)
     statements.get_item = TRY(database.prepare_statement("SELECT bottle_value FROM WebStorage WHERE storage_endpoint = ? AND storage_key = ? AND bottle_key = ?;"sv));
     statements.clear = TRY(database.prepare_statement("DELETE FROM WebStorage WHERE storage_endpoint = ? AND storage_key = ?;"sv));
     statements.get_keys = TRY(database.prepare_statement("SELECT bottle_key FROM WebStorage WHERE storage_endpoint = ? AND storage_key = ?;"sv));
-    statements.calculate_size_excluding_key = TRY(database.prepare_statement("SELECT SUM(LENGTH(bottle_key) + LENGTH(bottle_value)) FROM WebStorage WHERE storage_endpoint = ? AND storage_key = ? AND bottle_key != ?;"sv));
+    statements.calculate_size_including_key = TRY(database.prepare_statement("SELECT SUM(LENGTH(bottle_key) + LENGTH(bottle_value)) FROM WebStorage WHERE storage_endpoint = ? AND storage_key = ? AND bottle_key != ?;"sv));
```
trflynn89: I don't think the previous name meant it was excluding the key from the size, but rather that it was filtering out a specific bottle key. Maybe `calculate_storage_size_excluding_bottle` or something would make more sense.

trflynn89: Which then probably means the subsequent commit should not be using this statement, but a new statement that does not exclude a specific bottle.

Libraries/LibWebView/StorageJar.cpp
```diff
@@ -165,6 +172,19 @@ Vector<String> StorageJar::PersistedStorage::get_keys(StorageEndpointType storag
...
+u64 StorageJar::PersistedStorage::usage(StorageEndpointType storage_endpoint, String const& storage_key)
+{
+    u64 current_size_in_bytes = 0;
+    database.execute_statement(
+        statements.calculate_size_including_key,
+        [&](auto statement_id) {
+            current_size_in_bytes = database.result_column<int>(statement_id, 0);
+        },
+        static_cast<int>(to_underlying(storage_endpoint)),
+        storage_key);
```
trflynn89, line 184: Yeah, I think we need a new statement. The statement used here has 3 placeholder values, but we're only filling in 2.

PR #5812, opened 2025-08-11
LibWeb/HTML: Some WPT improvements for the `input` element

skyz1:
This PR fixes a few bugs in ladybirds implementation of the `input` element:
- Consider if the element is enabled when checking if it is mutable
- Trim seconds/milliseconds components of local datetime if they are zero
- Prevent month values from having the year zero
- Make sure range inputs cannot underflow, overflow or suffer a step mismatch
- Return the default value of range inputs if no value has been set
- Handle step up/down in the case of a step mismatch correctly
- Include non-zero second/millisecond components in time string when converting a number
- Consider the first week that overlaps with a year as the first week of the year (instead of the first fully contained week)


AK/Time.cpp
```diff
@@ -283,10 +283,13 @@ MonotonicTime MonotonicTime::now_coarse()
...
 {
     auto january_1_weekday = day_of_week(week_year, 1, 1);
-    i32 offset_to_monday = (january_1_weekday <= 3) ? -january_1_weekday : 7 - january_1_weekday;
-    i32 first_monday_of_year = 1 + offset_to_monday;
+    i32 first_monday_of_year = 1 - january_1_weekday;
     i32 day_of_year = (first_monday_of_year + (week - 1) * 7) + 1;
 
+    if (day_of_year < 0) {
+        return UnixDateTime::from_iso8601_week(week_year - 1, weeks_in_year(week_year - 1) + 1);
+    }
```
gmta: In your commit message you wrote:

> in the HTML spec, the first week is the earliest week that overlaps with the year.

That's not what ISO-8601 weeks represent; that first week was defined as the one containing the first Thursday of January (or the first week with at least 4 days of January in it).

gmta: We could do with some tests for this method in TestTime.cpp, especially with inputs around weeks 52, 53 and 1.

skyz1:
@gmta I wasn't familiar with ISO8601 so I skipped over this by accident😅 I did some research and reimplemented the algorithms - they should work now.

AK/Time.cpp
```diff
@@ -282,23 +283,34 @@ MonotonicTime MonotonicTime::now_coarse()
...
+        return UnixDateTime::from_ordinal_date(week_year + 1, ordinal_day - days_in_week_year);
+    return UnixDateTime::from_ordinal_date(week_year, ordinal_day);
+}
+
+UnixDateTime UnixDateTime::from_ordinal_date(u32 year, u32 day)
+{
+    static constexpr Array<u32, 12> month_starts_normal = { 1, 32, 60, 91, 121, 152, 182, 213, 244, 274, 305, 335 };
+    static constexpr Array<u32, 12> month_starts_leap = { 1, 32, 61, 92, 122, 153, 183, 214, 245, 275, 306, 336 };
+
+    auto month_starts = is_leap_year(year) ? month_starts_leap : month_starts_normal;
```
gmta: Can you make this `auto const&` so we don't copy the arrays?

AK/Time.cpp
```diff
@@ -282,23 +283,34 @@ MonotonicTime MonotonicTime::now_coarse()
...
+
+    // Estimate month using integer division (approx 30.6 days per month)
+    int m_est = (day * 12 + 6) / 367; // Gives 0-based month index
```
gmta: The `m_` prefix is used for member variables and this seems unsigned, so doesn't need to be an `int`; this should just be `u32 estimated_index` or so.

Libraries/LibWeb/HTML/HTMLInputElement.cpp
```diff
@@ -2383,26 +2383,18 @@ static Utf16String convert_number_to_week_string(double input)
...
 
-    if (week < 0) {
-        --year;
-        week = weeks_in_year(year) + week;
-    }
+    auto ordinal_day = (is_leap_year(year) ? leapYearDays[month - 1] : normalYearDays[month - 1]) + day;
 
+    unsigned week = (10 + ordinal_day - 1) / 7; // - 1, because the day of the week is always monday
+    if (week > iso8061_weeks_in_year(year))
+        return Utf16String::formatted("{:04d}-W{:02d}", year + 1, 1);
```
gmta: You could just do `++year; week = 1;` here so the string format takes place in a single line.

AK/Time.cpp
```diff
@@ -7,6 +7,7 @@
 
 #include <AK/Checked.h>
 #include <AK/DateConstants.h>
+#include <AK/Math.h>
```
gmta: What do you need this include for?

Libraries/LibWeb/HTML/Dates.cpp
```diff
@@ -90,10 +90,17 @@ bool is_valid_month_string(Utf16View const& value)
...
+    auto year_is_zero = true;
+    for (auto digit : parts[0]) {
         if (!is_ascii_digit(digit))
             return false;
 
+        if (digit != '0')
+            year_is_zero = false;
+    }
+    if (year_is_zero)
+        return false;
```
gmta, line 89: I think this code would be a bit more readable if it just said:

```cpp
if (parts[0] == "0000"sv)
    return false;
```

Libraries/LibWeb/HTML/HTMLInputElement.cpp
```diff
@@ -1716,19 +1721,60 @@ Utf16String HTMLInputElement::value_sanitization_algorithm(Utf16String const& va
...
+        // to the best representation, as a floating-point number, of the minimum.
+        if (is_number_underflowing(number_value)) {
+            number_value = min().value();
+        }
+
+        // When the element is suffering from a step mismatch, the user agent must round the element's value
+        // to the nearest number for which the element would not suffer from a step mismatch,
+        // and which is greater than or equal to the minimum, and, if the maximum is not less than the minimum,
+        // which is less than or equal to the maximum, if there is a number that matches these constraints.
+        // If two numbers match these constraints, then user agents must use the one nearest to positive infinity.
```
gmta: Please wrap all long comments at column 120.

Libraries/LibWeb/HTML/Dates.cpp
skyz1: The spec says:

> 1. Four or more ASCII digits, representing year, where year > 0

That means we must be able to handle any number of zeros if I am not mistaken.

trflynn89: Given that we parse the year below, could we just do the following?
```diff
diff --git a/Libraries/LibWeb/HTML/Dates.cpp b/Libraries/LibWeb/HTML/Dates.cpp
index a9dcbdc9db4..e3bfb345f5d 100644
--- a/Libraries/LibWeb/HTML/Dates.cpp
+++ b/Libraries/LibWeb/HTML/Dates.cpp
@@ -51,9 +51,9 @@ bool is_valid_week_string(Utf16View const& value)
     if (parts[1].length_in_code_units() != 3)
         return false;

-    for (auto digit : parts[0])
-        if (!is_ascii_digit(digit))
-            return false;
+    auto year = parts[0].to_number<u64>();
+    if (!year.has_value() || *year == 0)
+        return false;

     if (!parts[1].starts_with('W'))
         return false;
@@ -62,14 +62,8 @@ bool is_valid_week_string(Utf16View const& value)
     if (!is_ascii_digit(parts[1].code_unit_at(2)))
         return false;

-    u64 year = 0;
-    for (auto d : parts[0]) {
-        year *= 10;
-        year += parse_ascii_digit(d);
-    }
-
     auto week = (parse_ascii_digit(parts[1].code_unit_at(1)) * 10) + parse_ascii_digit(parts[1].code_unit_at(2));
-    return week >= 1 && week <= week_number_of_the_last_day(year);
+    return week >= 1 && week <= week_number_of_the_last_day(*year);
 }
```

skyz1: That works. Do you want me to do the same in the `is_valid_month_string` method? There we do not need to parse the year.

trflynn89: Ah woops, didn't realize I edited `is_valid_week_string` there - but yeah, I think it's clearer to just do it this way in both methods, since they both need to check that the year is non-zero.

trflynn89:
FYI about the build errors:
```
 /home/runner/_work/ladybird/ladybird/Libraries/LibWeb/HTML/HTMLSelectElement.cpp:498:59: error: no viable conversion from 'AK::String' to 'const Utf16String'
  498 |         return WebIDL::InvalidStateError::create(realm(), "Element is not mutable"_string);
```

Exception messages are now UTF-16, so that `_string` becomes `_utf16`.

skyz1:
> FYI about the build errors:
> 
> ```
>  /home/runner/_work/ladybird/ladybird/Libraries/LibWeb/HTML/HTMLSelectElement.cpp:498:59: error: no viable conversion from 'AK::String' to 'const Utf16String'
>   498 |         return WebIDL::InvalidStateError::create(realm(), "Element is not mutable"_string);
> ```
> 
> Exception messages are now UTF-16, so that `_string` becomes `_utf16`.

Thank you for the tip, I didn't notice that I reverted this change while resolving the conflict.

PR #5814, opened 2025-08-11
LibWeb: Fix CSS getComputedStyle() properties when using compound selectors

luizgfranca:
When using a compound selector like ".a .b" in the CSS rules the properties assigned would work correctly when rendering the page, for instance, if you set the background color and there was a div with class "a b" it would be correctly applied, but when querying its CSS properties, they would not be taken into account.

A simple example would be if you load ladybird on the following page:

```html
<!DOCTYPE html>

<html>
  <head>
    <style>
      .select .select-options {
          display: none;
      }
    </style>
  </head>
  <body>
    <div class="select">
      <div class="select-options">
      </div>
    </div>
  </body>
</html>
```

and in the devtools run the following lines:
```js
const x = document.querySelector('.select-options')
getComputedStyle(x)['display']
```

you would get that the display is "block", even though it was set to "none" in the style. Now if you remove the `.select` from the css rule it would return "none" correctly.

This kind of thing is frequently used by JQuery, and is breaking some features in sites that use it.

This was happening because when doing the style computation to return the property, the `selector` class wasn't in the ancestor filter. 

From what I could understand, this seems to happen because of a difference in the computation behavior in the first rendering and on the property computation ones: 

In the first rendering we traverse the entire tree, and this means that the ancestor elements will be pushed to the ancestor filter, and only when all their children are explored they are removed from it. 

But on the property computation behavior we look directly into the element, without having to traverse the entire tree.

The ancestor filter behavior tries to optimize away the computation of rules with ancestors not relevant to the current element, and it does that verifying if all the ancestor selectors in the list may be in the filter, but as said, if we didn't pass by all the ancestors to reach here, they will never have been pushed to the filter, and will not exist. 

This makes our valid rule not be computed, because it was rejected by the filter, assuming it was not relevant.

This solution creates a flag `enable_ancestor_filter`, to give control to the callers from the different callers of the style computation, and uses this flag in the `CSSStyleProperty` evaluations and on the `Dump All Resolved Styles` feature. This way we can avoid the ancestor filter optimization when it is not relevant.

I'm still new to the project, so I don't have the entire architecture in my head. I don't know if this is a good solution, and I'm open to feedback on a better way to solve this.

### DIFF
--- a/Tests/LibWeb/Text/expected/wpt-import/css/cssom/getComputedStyle-display-none-002.txt
+++ b/Tests/LibWeb/Text/expected/wpt-import/css/cssom/getComputedStyle-display-none-002.txt
@@ -1,0 +1,6 @@
+Harness status: OK
+
+Found 1 tests
+
+1 Pass
+Pass	getComputedStyle gets invalidated in display: none subtrees due to rule matching changes
--- a/Tests/LibWeb/Text/input/wpt-import/css/cssom/getComputedStyle-display-none-002.html
+++ b/Tests/LibWeb/Text/input/wpt-import/css/cssom/getComputedStyle-display-none-002.html
@@ -1,0 +1,34 @@
+<!doctype html>
+<meta charset="utf-8">
+<title>CSSOM: getComputedStyle gets invalidated for display: none elements (rules)</title>
+<link rel="help" href="https://drafts.csswg.org/cssom/#dom-window-getcomputedstyle">
+<link rel="help" href="https://bugs.webkit.org/show_bug.cgi?id=186882">
+<link rel="author" title="Emilio Cobos Álvarez" href="mailto:emilio@crisal.io">
+<script src=../../resources/testharness.js></script>
+<script src=../../resources/testharnessreport.js></script>
+<style>
+#undisplayed {
+  display: none;
+  color: red;
+}
+.sibling + #undisplayed {
+  color: green;
+}
+
+.sibling + #undisplayed > div {
+  color: yellow;
+}
+</style>
+<div id="helper"></div>
+<div id="undisplayed"><div id="child"></div></div>
+<script>
+  test(function() {
+    let undisplayed_style = getComputedStyle(undisplayed);
+    let undisplayed_child_style = getComputedStyle(child);
+    assert_equals(undisplayed_style.color, "rgb(255, 0, 0)");
+    assert_equals(undisplayed_child_style.color, "rgb(255, 0, 0)");
+    helper.classList.add("sibling");
+    assert_equals(undisplayed_style.color, "rgb(0, 128, 0)");
+    assert_equals(undisplayed_child_style.color, "rgb(255, 255, 0)");
+  }, "getComputedStyle gets invalidated in display: none subtrees due to rule matching changes");
+</script>


kalenikaliaksandr (changes requested):
let's use `enum class EnableAncestorFilter { Yes, No }` instead of `bool enable_ancestor_filter`, but overall change looks good!

Libraries/LibWeb/CSS/CSSStyleProperties.cpp
```diff
@@ -185,7 +185,7 @@ Optional<StyleProperty> CSSStyleProperties::property(PropertyID property_id) con
         }
 
         if (!layout_node) {
-            auto style = element.document().style_computer().compute_style(element, pseudo_element);
+            auto style = element.document().style_computer().compute_style(element, pseudo_element, {}, false);
```
gmta: After a short discussion in the #css channel; adding an argument to disable an internal functionality/optimization like this is not really a well thought-out API.

It would be better to support `compute_style()` for single elements without tree traversal. For example, you could do a short ancestor traversal on the spot so you don't need to disable the filter at all.

luizgfranca:
Here's a general description of what I did based on the suggestions.

Moved `update_style_recursively()` to `StyleComputer`. To do this I also had to migrate the functions used primarily by it:
- `Element::recompute_style()` renamed to `StyleComputer::update_element_computed_style()` encapsulating the reset of the element invalidation flags (since it is inside the element context)
- `Element::compute_required_invalidation()` , in this case, for some behavior that was local to element, also moved it to `Element::computed_properties_changed()`, since it made sense
- `Element::recompute_inherited_style()`,  renamed to `StyleComputer::update_element_inherited_computed_style()`, but also leaving a `Element::update_inherited_computed_style()` because of some internal behavior that uses it
I also created `StyleComputer::update_document_computed_style()` to encapsulate the ancestor filter reset and the call to `update_style_recursively()` to the whole document.

Moved `invalidate_style_for_elements_affected_by_pseudo_class_change()` to `StyleComputer` , because it used the `ancestor_filter` 

Remove ancestor filter references from `TreeBuilder` as it was not using it in any meaningful way. Its style computation functions were changed to single element computations. I didn't see any performance impact with this change.

With this I had moved all external references to the `ancestor_filter` to `StyleComputer`, so I privated it.

Created a new function `StyleComputer::compute_element_style()` to compute the style of a single element as suggested, and to do this behavior I added some functions to enable or disable the ancestor filter (to avoid having to keep passing parameters around). When the ancestor filter is disabled it always returns a positive result, because in this case the caller assumes the element may be on the filter, and takes it into account on the style computation.

I also added a WPT test we were failing that I've found after submitting the PR and that was fixed by this. :smile:

Libraries/LibWeb/DOM/Element.cpp
```diff
@@ -3856,6 +3686,22 @@ void Element::attribute_changed(FlyString const& local_name, Optional<String> co
 #undef __ENUMERATE_ARIA_ATTRIBUTE
 }
 
+void Element::computed_properties_changed(GC::Ptr<CSS::ComputedProperties> old_properties, GC::Ptr<CSS::ComputedProperties> new_properties)
+{
+    auto old_display_is_none = old_properties ? old_properties->display().is_none() : true;
+    auto new_display_is_none = new_properties->display().is_none();
```
AtkinsSJ, line 3692: Does `new_properties` always have a value? If so, making the parameter a `GC::Ref` would enforce that.

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -3977,23 +3977,42 @@ static void for_each_element_hash(DOM::Element const& element, auto callback)
...
 {
+    m_ancestor_filter_enabled = true;
     m_ancestor_filter->clear();
 }
 
 void StyleComputer::push_ancestor(DOM::Element const& element)
 {
+    if (!ancestor_filter_enabled()) {
+        return;
+    }
```
AtkinsSJ, line 3997: Nit: We avoid using `{}` for single-line if blocks. Same for the other places in this commit.

Libraries/LibWeb/CSS/CSSStyleProperties.cpp
```diff
@@ -182,7 +182,7 @@ Optional<StyleProperty> CSSStyleProperties::property(PropertyID property_id) con
         }
 
         if (!layout_node) {
-            auto style = abstract_element.document().style_computer().compute_style(abstract_element);
+            auto style = abstract_element.document().style_computer().compute_element_style(abstract_element);
```
AtkinsSJ, line 185: Commit message nit: "appropriately" is misspelled.

AtkinsSJ (changes requested):
This looks good, and it really does make more sense for style computation code to go in `StyleComputer`.

I only have very minor comments. Another that didn't apply to a particular place: In a few commit titles you've capitalised classes like `styleComputer` instead of `StyleComputer`. Please correct those.

PR #5821, opened 2025-08-11
CI: Create wasm artifact and use it in the js-benchmarks workflow

Lubrsi:


.github/workflows/js-benchmarks.yml
```diff
@@ -58,14 +58,28 @@ jobs:
...
           tar -xvzf ladybird-js-${{ matrix.os_name }}-${{ matrix.arch }}.tar.gz
 
-      - name: 'Run the JS benchmarks'
+      - name: 'Download Wasm repl artifact'
+        id: download-artifact
+        uses: dawidd6/action-download-artifact@v11
+        with:
+          run_id: ${{ github.event.workflow_run.id }}
+          name: ladybird-wasm-${{ matrix.package_type }}
+          path: wasm-repl
```
gmta, line 67: You need to check `workflow_run` at the top of this file and see how the wasm build workflow ties into this; you'd need to start this workflow only when _both_ workflows have successfully completed.

If that's not possible, I wouldn't mind combining the two repl builds into a single workflow for this.

Lubrsi: I had a look at the documentation and it doesn't have an example of having multiple strings in `workflow_runs` and looking at workflows from other projects (found by grep.app), it seems like it triggers when only one of them in the array finishes:
https://github.com/vercel/next.js/blob/2c3e9d83b2e1e8824c32788c3eed434e5a636029/.github/workflows/retry_test.yml#L5
https://github.com/hashicorp/terraform-provider-azurerm/blob/53081beb2231841024cf4d0e650b08d13f163eb2/.github/workflows/add-waiting-response-on-fail.yaml#L6

With that, I've combined them into a single workflow.

.github/workflows/js-and-wasm-artifacts.yml
```diff
@@ -94,18 +100,33 @@ jobs:
           vcpkg_cache_path: ${{ github.workspace }}/Build/caches/vcpkg-binary-cache
           vcpkg_cache_primary_key: ${{ steps.cache-restore.outputs.vcpkg_cache_primary_key }}
 
+      - name: Create artifacts package
```
gmta: This is extracting the package, not creating it

.github/workflows/js-and-wasm-artifacts.yml
```diff
@@ -94,18 +100,33 @@ jobs:
...
+        shell: bash
+        run: |
+          set -e
+          tar -xvzf Build/distribution/ladybird-js-and-wasm${{ matrix.package_type }}.tar.gz
```
gmta: Where did this new .tar.gz filename come from? It's not in `Utilities/CMakeLists.txt`

.github/workflows/js-and-wasm-artifacts.yml
```diff
@@ -1,4 +1,4 @@
-name: Package the js repl as a binary artifact
+name: Package the js and wasm repls as a binary artifact
```
gmta: Heads up: this might break external automations since the name of the workflow is part of the API path. (Things have broken before by changes to this)

.github/workflows/js-and-wasm-artifacts.yml
gmta: Maybe it would be better to maintain and upload two separate artifacts; one for `js` and one for `wasm`.

PR #5822, opened 2025-08-11
LibWeb: Fix color interpolation by premultiplying alpha

Norbiros:
Part 1 of fixing https://hack4krak.pl

Before:
image
After:
image

Another image (left before, right after):
image

The current Color::interpolate_color method does not follow the specs properly. Started improving it by handling premultiplied alpha in color interpolation.

Only one WPT test covers this (color-transition-premultiplied), which we currently pass due to a different approach in Color.mixed_with.


gmta (approved):
Color types and interpolation definitely need some cleanup 😅 thank you for looking into this!

PR #5823, opened 2025-08-11
LibWeb/CSS: Stub out the CSS Typed OM

AtkinsSJ:
We get a few test passes just from these types existing. This PR doesn't allow actually reading or modifying style this way, but that will come in subsequent PRs. This seemed like a good place to pause and submit the changes.

Over 7000 lines of this is the tests imported in the first commit. I promise this isn't actually scary! :laughing:


Libraries/LibWeb/CSS/CSSStyleValue.cpp
```diff
@@ -0,0 +1,74 @@
...
+void CSSStyleValue::initialize(JS::Realm& realm)
+{
+    WEB_SET_PROTOTYPE_FOR_INTERFACE(CSSStyleValue);
+    Base::initialize(realm);
+}
+
+void CSSStyleValue::visit_edges(Cell::Visitor& visitor)
+{
+    Base::visit_edges(visitor);
+}
```
trflynn89: Don't really need this if we're not visiting anything

Libraries/LibWeb/CSS/CSSStyleValue.cpp
```diff
@@ -0,0 +1,74 @@
...
+    return WebIDL::SimpleException { WebIDL::SimpleExceptionType::TypeError, "FIXME: Implement CSSStyleValue::parse()"_string };
+}
+
+// https://drafts.css-houdini.org/css-typed-om-1/#dom-cssstylevalue-parseall
+WebIDL::ExceptionOr<GC::RootVector<GC::Ref<CSSStyleValue>>> CSSStyleValue::parse_all(JS::VM&, String const& property, String const& css_text)
+{
+    (void)property;
+    (void)css_text;
+    return WebIDL::SimpleException { WebIDL::SimpleExceptionType::TypeError, "FIXME: Implement CSSStyleValue::parse_all()"_string };
+}
```
trflynn89: Is there value to this over just using the FIXME attribute in IDL?

Libraries/LibWeb/CSS/StylePropertyMapReadOnly.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+    visitor.visit(m_source_declaration);
+    if (m_source_element.has_value())
+        m_source_element->visit(visitor);
+}
+
+// https://drafts.css-houdini.org/css-typed-om-1/#dom-stylepropertymapreadonly-get
+WebIDL::ExceptionOr<Variant<GC::Ref<CSSStyleValue>, Empty>> StylePropertyMapReadOnly::get(String property)
+{
+    // The get(property) method, when called on a StylePropertyMapReadOnly this, must perform the following steps:
```
trflynn89, line 55: I think we typically don't include these header comments from the spec, but not a big deal

Libraries/LibWeb/CSS/StylePropertyMapReadOnly.idl
```diff
@@ -0,0 +1,12 @@
+#import <CSS/CSSStyleValue.idl>
+
+// https://drafts.css-houdini.org/css-typed-om-1/#stylepropertymapreadonly
+// FIXME: [Exposed=(Window, Worker, PaintWorklet, LayoutWorklet)]
+[Exposed=(Window, Worker)]
+interface StylePropertyMapReadOnly {
+    // FIXME: iterable<USVString, sequence<CSSStyleValue>>;
```
trflynn89, line 7: Could this be `[FIXME]` instead?

Libraries/LibWeb/CSS/CSSStyleRule.idl
```diff
@@ -1,9 +1,13 @@
...
+#import <CSS/StylePropertyMap.idl>
 
 // https://drafts.csswg.org/cssom/#the-cssstylerule-interface
 [Exposed=Window]
 interface CSSStyleRule : CSSGroupingRule {
     attribute CSSOMString selectorText;
     [SameObject, PutForwards=cssText] readonly attribute CSSStyleProperties style;
+    
+    // https://drafts.css-houdini.org/css-typed-om-1/#declared-stylepropertymap-objects
+    [SameObject] readonly attribute StylePropertyMap styleMap;
```
trflynn89: We have some support for partial interfaces now. Does that work here?

Libraries/LibWeb/CSS/StylePropertyMapReadOnly.cpp
AtkinsSJ: I tend to include them, but it's not something that's consistent either way.

Libraries/LibWeb/CSS/CSSStyleRule.idl
AtkinsSJ: Oh, neat! I'll do that for these, thanks.

AtkinsSJ: `partial interface mixin` evidently doesn't work yet, but the others are happy.

Libraries/LibWeb/CSS/StylePropertyMapReadOnly.idl
AtkinsSJ: No, because it tries to include a `StylePropertyMapReadOnlyIterator` class that doesn't exist yet.

Libraries/LibWeb/CSS/CSSStyleValue.cpp
AtkinsSJ: Actually no, I just thought I was going to flesh these out more than I did. I've replaced them with the attribute.

AtkinsSJ:
Changes:
- Rebased
- Implemented Tim's feedback
- Stopped assuming that `Element::m_inline_style` has a value when creating the attributeStyleMap, but instead make sure to create it first. This showed up in SAN builds on CI.

PR #5825, opened 2025-08-11
LibWeb+LibWebView+Base: Add basic parental controls

Psychpsyo:
Adds very basic parental controls to the settings to block websites with adult content on them. If you need an adult website to test this with, pornhub.com includes the relevant meta tags on their site. (I've also tested quite a few other adult sites to see how well the coverage is, it seems like most adult websites use at least one of these)

About the checked tags:
`"adult"` - commonly used on porn sites, seems to be given as a recommendation online
`"RTA-5042-1996-1400-1577-RTA"` - The [Restricted to Adults label](https://www.rtalabel.org/), probably the closest thing to a standard that exists here.* Also commonly used.
`"mature"` - used on some NSFW image boards, such as rule34.xxx

I did consider just checking for the presence of a rating meta tag, but some image boards (like safebooru.org) set it to "general" to indicate that the content is NOT adult content.

Note: Of course this setting is awful as actual parental controls, since any kid could just disable it. For that, we might want to consider password-locking this somehow or tying into the OS's parental control options in the future.

\* Except for the two actual standards that exist for this:
1. [age.xml](https://www.jugendschutzprogramm.de/en/provider/age-xml/), which is a lot more complex and therefore used a lot less on actual pages out in the wild. (example: pornhub does not have an age.xml)
2. The `prefer: safe` HTTP header. [Firefox sends this](https://blog.mozilla.org/netpolicy/2014/07/22/prefersafe-making-online-safety-simpler-in-firefox/) when the OS has parental controls enabled, though I'm not sure if websites actually respect that. It also comes with similar problems as the Do Not Track header, giving the server an extra bit of a user's tracking footprint + potentially informing bad actors of the fact that a user has parental controls enabled. (= is probably underage)

### DIFF
--- a/Base/res/ladybird/about-pages/about.html
+++ b/Base/res/ladybird/about-pages/about.html
@@ -9,6 +9,7 @@
     <h1>List of About URLs</h1>
     <ul>
         <li><a href="about:about">about:about</a></li>
+        <li><a href="about:blocked">about:blocked</a></li>
         <li><a href="about:newtab">about:newtab</a></li>
         <li><a href="about:processes">about:processes</a></li>
         <li><a href="about:settings">about:settings</a></li>
--- a/Base/res/ladybird/about-pages/blocked.html
+++ b/Base/res/ladybird/about-pages/blocked.html
@@ -1,0 +1,14 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="color-scheme" content="light dark">
+    <title>Blocked by Parental Controls</title>
+</head>
+<body>
+    <h1>Blocked by Parental Controls</h1>
+    <p>
+        The page you were trying to access has been blocked by parental controls.
+    </p>
+</body>
+</html>
--- a/Base/res/ladybird/about-pages/settings.html
+++ b/Base/res/ladybird/about-pages/settings.html
@@ -378,6 +378,19 @@
         </div>
 
         <div class="card">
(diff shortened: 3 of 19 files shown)


gmta:
This feels very off-brand for Ladybird to me, especially if there's no good web spec involved. I'd leave this type of functionality to future browser extensions. But other maintainers should probably chime in as well, just my 2 cents.

AtkinsSJ:
My reaction was also that it feels like extension functionality. I don't *object* to it being in, but I don't think it would be a useful feature until we have wider adoption, and by that point we'll probably have extension support.

Psychpsyo:
I do somewhat agree with the sentiment here and wouldn't have strong objections to this being closed.

Though I feel like an extension might not be the best way to go about this in particular, because
1. I doubt it'd be able to hook into OS-wide settings for this sort of thing in the future and
2. it should probably always be trivially easy to remove an extension for any user of the browser.

AtkinsSJ:
True... though, disabling the blocker is also trivial.

gmta:
@Psychpsyo

> 1. I doubt it'd be able to hook into OS-wide settings for this sort of thing in the future and

We could eventually decide on exposing pieces of the OS' configuration only to browser extensions, for example.

> 2. it should probably always be trivially easy to remove an extension for any user of the browser.

Continuing on the freedom of actions within the browser itself, it should then also be trivially easy to disable any parental controls, overriding whatever the OS tells the browser to do (if at all).

PR #5827, opened 2025-08-11
LibWeb/IDB: Recent spec changes

stelar7:
Corresponds to:
* https://github.com/w3c/IndexedDB/pull/467
* https://github.com/w3c/IndexedDB/pull/458
* https://github.com/w3c/IndexedDB/pull/460


Libraries/LibWeb/IndexedDB/IDBRequest.h
```diff
@@ -63,7 +65,7 @@ class IDBRequest : public DOM::EventTarget {
 
     // A request has a result and an error
     JS::Value m_result;
-    GC::Ptr<WebIDL::DOMException> m_error;
+    Optional<GC::Ptr<WebIDL::DOMException>> m_error;
```
AtkinsSJ, line 68: `Optional<Ptr>` is an odd pattern, is there a reason this can't stay a `Ptr` or be an `Optional<Ref>`?

stelar7: There is 3 values for this, `unset`, `null`, and `DOMException`, so this was the best fitting type in my mind. Im open to suggestions 😅

AtkinsSJ: Oh, that's super weird. Yeah this works then!

PR #5828, opened 2025-08-11
LibWeb Trusted types csp integration

tete17:
This changes reflect the Section [4.3 Integration with Content Security Policy](https://w3c.github.io/trusted-types/dist/spec/#integration-with-content-security-policy)

It sets the basis of the other integrations like HTML or DOM.

These 2 new directives control when the browser should enforce the use of TrustedTypes in the InjectionSinks apis

### DIFF
--- a/Tests/LibWeb/Text/expected/wpt-import/trusted-types/TrustedTypePolicy-CSP-no-name.txt
+++ b/Tests/LibWeb/Text/expected/wpt-import/trusted-types/TrustedTypePolicy-CSP-no-name.txt
@@ -1,0 +1,6 @@
+Harness status: OK
+
+Found 1 tests
+
+1 Pass
+Pass	No name list given - policy creation fails.
--- a/Tests/LibWeb/Text/expected/wpt-import/trusted-types/TrustedTypePolicy-CSP-wildcard.txt
+++ b/Tests/LibWeb/Text/expected/wpt-import/trusted-types/TrustedTypePolicy-CSP-wildcard.txt
@@ -1,0 +1,6 @@
+Harness status: OK
+
+Found 1 tests
+
+1 Pass
+Pass	CSP supports wildcards.
--- a/Tests/LibWeb/Text/expected/wpt-import/trusted-types/TrustedTypePolicyFactory-constants.txt
+++ b/Tests/LibWeb/Text/expected/wpt-import/trusted-types/TrustedTypePolicyFactory-constants.txt
@@ -1,0 +1,11 @@
+Harness status: OK
+
+Found 6 tests
+
+6 Pass
+Pass	trustedTypes.emptyHTML returns the intended value.
+Pass	trustedTypes.emptyHTML cannot be redefined.
+Pass	trustedTypes.emptyHTML cannot be redefined via defineProperty.
+Pass	trustedTypes.emptyScript returns the intended value.
+Pass	trustedTypes.emptyScript cannot be redefined.
+Pass	trustedTypes.emptyScript cannot be redefined via defineProperty.
--- a/Tests/LibWeb/Text/expected/wpt-import/trusted-types/idlharness.window.txt
+++ b/Tests/LibWeb/Text/expected/wpt-import/trusted-types/idlharness.window.txt
@@ -1,0 +1,105 @@
+Harness status: OK
+
+Found 100 tests
+
+100 Pass
+Pass	idl_test setup
+Pass	idl_test validation
+Pass	Partial interface mixin WindowOrWorkerGlobalScope: original interface mixin defined
+Pass	Partial interface mixin WindowOrWorkerGlobalScope: member names are unique
+Pass	Partial interface Window: member names are unique
+Pass	Window includes GlobalEventHandlers: member names are unique
+Pass	Window includes WindowEventHandlers: member names are unique
+Pass	Window includes WindowOrWorkerGlobalScope: member names are unique
+Pass	WorkerGlobalScope includes WindowOrWorkerGlobalScope: member names are unique
+Pass	Window includes AnimationFrameProvider: member names are unique
+Pass	Window includes WindowSessionStorage: member names are unique
+Pass	Window includes WindowLocalStorage: member names are unique
+Pass	TrustedHTML interface: existence and properties of interface object
+Pass	TrustedHTML interface object length
+Pass	TrustedHTML interface object name
+Pass	TrustedHTML interface: existence and properties of interface prototype object
+Pass	TrustedHTML interface: existence and properties of interface prototype object's "constructor" property
+Pass	TrustedHTML interface: existence and properties of interface prototype object's @@unscopables property
+Pass	TrustedHTML interface: stringifier
+Pass	TrustedHTML interface: operation toJSON()
+Pass	TrustedHTML must be primary interface of window.trustedTypes.createPolicy("SomeName1", { createHTML: s => s }).createHTML("A string")
+Pass	Stringification of window.trustedTypes.createPolicy("SomeName1", { createHTML: s => s }).createHTML("A string")
+Pass	TrustedHTML interface: window.trustedTypes.createPolicy("SomeName1", { createHTML: s => s }).createHTML("A string") must inherit property "toJSON()" with the proper type
+Pass	TrustedHTML interface: toJSON operation on window.trustedTypes.createPolicy("SomeName1", { createHTML: s => s }).createHTML("A string")
+Pass	TrustedScript interface: existence and properties of interface object
+Pass	TrustedScript interface object length
+Pass	TrustedScript interface object name
+Pass	TrustedScript interface: existence and properties of interface prototype object
+Pass	TrustedScript interface: existence and properties of interface prototype object's "constructor" property
+Pass	TrustedScript interface: existence and properties of interface prototype object's @@unscopables property
+Pass	TrustedScript interface: stringifier
+Pass	TrustedScript interface: operation toJSON()
+Pass	TrustedScript must be primary interface of window.trustedTypes.createPolicy("SomeName2", { createScript: s => s }).createScript("A string")
+Pass	Stringification of window.trustedTypes.createPolicy("SomeName2", { createScript: s => s }).createScript("A string")
+Pass	TrustedScript interface: window.trustedTypes.createPolicy("SomeName2", { createScript: s => s }).createScript("A string") must inherit property "toJSON()" with the proper type
+Pass	TrustedScript interface: toJSON operation on window.trustedTypes.createPolicy("SomeName2", { createScript: s => s }).createScript("A string")
+Pass	TrustedScriptURL interface: existence and properties of interface object
+Pass	TrustedScriptURL interface object length
+Pass	TrustedScriptURL interface object name
+Pass	TrustedScriptURL interface: existence and properties of interface prototype object
+Pass	TrustedScriptURL interface: existence and properties of interface prototype object's "constructor" property
+Pass	TrustedScriptURL interface: existence and properties of interface prototype object's @@unscopables property
+Pass	TrustedScriptURL interface: stringifier
+Pass	TrustedScriptURL interface: operation toJSON()
+Pass	TrustedScriptURL must be primary interface of window.trustedTypes.createPolicy("SomeName3", { createScriptURL: s => s }).createScriptURL("A string")
+Pass	Stringification of window.trustedTypes.createPolicy("SomeName3", { createScriptURL: s => s }).createScriptURL("A string")
+Pass	TrustedScriptURL interface: window.trustedTypes.createPolicy("SomeName3", { createScriptURL: s => s }).createScriptURL("A string") must inherit property "toJSON()" with the proper type
+Pass	TrustedScriptURL interface: toJSON operation on window.trustedTypes.createPolicy("SomeName3", { createScriptURL: s => s }).createScriptURL("A string")
+Pass	TrustedTypePolicyFactory interface: existence and properties of interface object
+Pass	TrustedTypePolicyFactory interface object length
+Pass	TrustedTypePolicyFactory interface object name
+Pass	TrustedTypePolicyFactory interface: existence and properties of interface prototype object
+Pass	TrustedTypePolicyFactory interface: existence and properties of interface prototype object's "constructor" property
+Pass	TrustedTypePolicyFactory interface: existence and properties of interface prototype object's @@unscopables property
+Pass	TrustedTypePolicyFactory interface: operation createPolicy(DOMString, optional TrustedTypePolicyOptions)
+Pass	TrustedTypePolicyFactory interface: operation isHTML(any)
+Pass	TrustedTypePolicyFactory interface: operation isScript(any)
+Pass	TrustedTypePolicyFactory interface: operation isScriptURL(any)
+Pass	TrustedTypePolicyFactory interface: attribute emptyHTML
+Pass	TrustedTypePolicyFactory interface: attribute emptyScript
+Pass	TrustedTypePolicyFactory interface: operation getAttributeType(DOMString, DOMString, optional DOMString?, optional DOMString?)
+Pass	TrustedTypePolicyFactory interface: operation getPropertyType(DOMString, DOMString, optional DOMString?)
+Pass	TrustedTypePolicyFactory interface: attribute defaultPolicy
+Pass	TrustedTypePolicyFactory must be primary interface of window.trustedTypes
+Pass	Stringification of window.trustedTypes
+Pass	TrustedTypePolicyFactory interface: window.trustedTypes must inherit property "createPolicy(DOMString, optional TrustedTypePolicyOptions)" with the proper type
+Pass	TrustedTypePolicyFactory interface: calling createPolicy(DOMString, optional TrustedTypePolicyOptions) on window.trustedTypes with too few arguments must throw TypeError
+Pass	TrustedTypePolicyFactory interface: window.trustedTypes must inherit property "isHTML(any)" with the proper type
+Pass	TrustedTypePolicyFactory interface: calling isHTML(any) on window.trustedTypes with too few arguments must throw TypeError
+Pass	TrustedTypePolicyFactory interface: window.trustedTypes must inherit property "isScript(any)" with the proper type
+Pass	TrustedTypePolicyFactory interface: calling isScript(any) on window.trustedTypes with too few arguments must throw TypeError
+Pass	TrustedTypePolicyFactory interface: window.trustedTypes must inherit property "isScriptURL(any)" with the proper type
+Pass	TrustedTypePolicyFactory interface: calling isScriptURL(any) on window.trustedTypes with too few arguments must throw TypeError
+Pass	TrustedTypePolicyFactory interface: window.trustedTypes must inherit property "emptyHTML" with the proper type
+Pass	TrustedTypePolicyFactory interface: window.trustedTypes must inherit property "emptyScript" with the proper type
+Pass	TrustedTypePolicyFactory interface: window.trustedTypes must inherit property "getAttributeType(DOMString, DOMString, optional DOMString?, optional DOMString?)" with the proper type
+Pass	TrustedTypePolicyFactory interface: calling getAttributeType(DOMString, DOMString, optional DOMString?, optional DOMString?) on window.trustedTypes with too few arguments must throw TypeError
+Pass	TrustedTypePolicyFactory interface: window.trustedTypes must inherit property "getPropertyType(DOMString, DOMString, optional DOMString?)" with the proper type
+Pass	TrustedTypePolicyFactory interface: calling getPropertyType(DOMString, DOMString, optional DOMString?) on window.trustedTypes with too few arguments must throw TypeError
+Pass	TrustedTypePolicyFactory interface: window.trustedTypes must inherit property "defaultPolicy" with the proper type
+Pass	TrustedTypePolicy interface: existence and properties of interface object
+Pass	TrustedTypePolicy interface object length
+Pass	TrustedTypePolicy interface object name
+Pass	TrustedTypePolicy interface: existence and properties of interface prototype object
+Pass	TrustedTypePolicy interface: existence and properties of interface prototype object's "constructor" property
+Pass	TrustedTypePolicy interface: existence and properties of interface prototype object's @@unscopables property
+Pass	TrustedTypePolicy interface: attribute name
+Pass	TrustedTypePolicy interface: operation createHTML(DOMString, any...)
+Pass	TrustedTypePolicy interface: operation createScript(DOMString, any...)
+Pass	TrustedTypePolicy interface: operation createScriptURL(DOMString, any...)
+Pass	TrustedTypePolicy must be primary interface of window.trustedTypes.createPolicy("SomeName", { createHTML: s => s })
+Pass	Stringification of window.trustedTypes.createPolicy("SomeName", { createHTML: s => s })
+Pass	TrustedTypePolicy interface: window.trustedTypes.createPolicy("SomeName", { createHTML: s => s }) must inherit property "name" with the proper type
+Pass	TrustedTypePolicy interface: window.trustedTypes.createPolicy("SomeName", { createHTML: s => s }) must inherit property "createHTML(DOMString, any...)" with the proper type
+Pass	TrustedTypePolicy interface: calling createHTML(DOMString, any...) on window.trustedTypes.createPolicy("SomeName", { createHTML: s => s }) with too few arguments must throw TypeError
+Pass	TrustedTypePolicy interface: window.trustedTypes.createPolicy("SomeName", { createHTML: s => s }) must inherit property "createScript(DOMString, any...)" with the proper type
+Pass	TrustedTypePolicy interface: calling createScript(DOMString, any...) on window.trustedTypes.createPolicy("SomeName", { createHTML: s => s }) with too few arguments must throw TypeError
+Pass	TrustedTypePolicy interface: window.trustedTypes.createPolicy("SomeName", { createHTML: s => s }) must inherit property "createScriptURL(DOMString, any...)" with the proper type
+Pass	TrustedTypePolicy interface: calling createScriptURL(DOMString, any...) on window.trustedTypes.createPolicy("SomeName", { createHTML: s => s }) with too few arguments must throw TypeError
+Pass	Window interface: attribute trustedTypes
--- a/Tests/LibWeb/Text/input/wpt-import/interfaces/trusted-types.idl
+++ b/Tests/LibWeb/Text/input/wpt-import/interfaces/trusted-types.idl
@@ -1,0 +1,65 @@
+// GENERATED CONTENT - DO NOT EDIT
+// Content was automatically extracted by Reffy into webref
+// (https://github.com/w3c/webref)
+// Source: Trusted Types (https://w3c.github.io/trusted-types/dist/spec/)
+
+[Exposed=(Window,Worker)]
+interface TrustedHTML {
+  stringifier;
+  DOMString toJSON();
+};
+
+[Exposed=(Window,Worker)]
+interface TrustedScript {
+  stringifier;
+  DOMString toJSON();
+};
+
+[Exposed=(Window,Worker)]
+interface TrustedScriptURL {
+  stringifier;
+  USVString toJSON();
+};
+
+[Exposed=(Window,Worker)] interface TrustedTypePolicyFactory {
+    TrustedTypePolicy createPolicy(
+        DOMString policyName, optional TrustedTypePolicyOptions policyOptions = {});
+    boolean isHTML(any value);
+    boolean isScript(any value);
+    boolean isScriptURL(any value);
+    readonly attribute TrustedHTML emptyHTML;
+    readonly attribute TrustedScript emptyScript;
+    DOMString? getAttributeType(
+        DOMString tagName,
+        DOMString attribute,
+        optional DOMString? elementNs = "",
+        optional DOMString? attrNs = "");
+    DOMString? getPropertyType(
+        DOMString tagName,
+        DOMString property,
+        optional DOMString? elementNs = "");
+    readonly attribute TrustedTypePolicy? defaultPolicy;
+};
+
+[Exposed=(Window,Worker)]
+interface TrustedTypePolicy {
+  readonly attribute DOMString name;
+  TrustedHTML createHTML(DOMString input, any... arguments);
+  TrustedScript createScript(DOMString input, any... arguments);
+  TrustedScriptURL createScriptURL(DOMString input, any... arguments);
+};
+
+dictionary TrustedTypePolicyOptions {
+   CreateHTMLCallback createHTML;
+   CreateScriptCallback createScript;
+   CreateScriptURLCallback createScriptURL;
+};
+callback CreateHTMLCallback = DOMString? (DOMString input, any... arguments);
+callback CreateScriptCallback = DOMString? (DOMString input, any... arguments);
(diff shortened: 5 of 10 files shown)


Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.h
```diff
@@ -51,6 +52,12 @@ class TrustedTypePolicyFactory final : public Bindings::PlatformObject {
 
     // https://w3c.github.io/trusted-types/dist/spec/#trustedtypepolicyfactory-default-policy
     GC::Ptr<TrustedTypePolicy> m_default_policy;
+
+    // https://www.w3.org/TR/trusted-types/#dom-trustedtypepolicyfactory-emptyhtml
+    const GC::Ref<TrustedHTML const> m_empty_html;
+
+    // https://www.w3.org/TR/trusted-types/#dom-trustedtypepolicyfactory-emptyscript
+    const GC::Ref<TrustedScript const> m_empty_script;
```
AtkinsSJ: Nit: east const

Libraries/LibWeb/TrustedTypes/InjectionSink.h
```diff
@@ -6,13 +6,32 @@
...
+
 namespace Web::TrustedTypes {
 
+#define EVENT_HANDLERS_INJECTION_SINKS(attribute_name, event_name) \
+    __ENUMERATE_INJECTION_SINKS(Element##attribute_name, "Element " #attribute_name)
+
 // https://w3c.github.io/trusted-types/dist/spec/#injection-sink
+#define ENUMERATE_INJECTION_SINKS                                                    \
+    __ENUMERATE_INJECTION_SINKS(Documentwrite, "Document write")                     \
+    __ENUMERATE_INJECTION_SINKS(Documentwriteln, "Document writeln")                 \
```
AtkinsSJ, line 21: Capitalising the names like this is a bit unfortunate, is that necessary?

Libraries/LibWeb/ContentSecurityPolicy/Directives/RequireTrustedTypesFor.cpp
```diff
@@ -0,0 +1,76 @@
...
+{
+    // 1. If request’s url’s scheme is not "javascript", return "Allowed" and abort further steps.
+    if (request->url().scheme() != "javascript"sv)
+        return Result::Allowed;
+
+    // 2. Let urlString be the result of running the URL serializer on request’s url.
+    auto url_string = request->url().serialize();
+
+    // 3. Let encodedScriptSource be the result of removing the leading "javascript:" from urlString.
+    auto const encoded_script_source = MUST(url_string.substring_from_byte_offset("javascript:"sv.length()));
```
AtkinsSJ: Using `substring_from_byte_offset_with_shared_superstring()` would avoid making a copy for this temporary string.

Libraries/LibWeb/ContentSecurityPolicy/Directives/RequireTrustedTypesFor.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+    // 4. Let convertedScriptSource be the result of executing Process value with a default policy algorithm, with the following arguments:
+    //    TrustedScript:
+    //      expectedType
+    //    request’s clients’s global object:
+    //      global
+    //    encodedScriptSource:
+    //      input
+    //    "Location href":
+    //      sink
```
AtkinsSJ: The parameter names and values are flipped here (the spec is weird). Please quote it verbatim, or at least flip them around so it's `parameter: value`.

Libraries/LibWeb/TrustedTypes/TrustedTypePolicy.cpp
```diff
@@ -182,4 +185,56 @@ WebIDL::ExceptionOr<JS::Value> TrustedTypePolicy::get_trusted_type_policy_value(
...
+WebIDL::ExceptionOr<JS::Value> process_value_with_a_default_policy(TrustedTypeName trusted_type_name, JS::Object& global, Variant<GC::Root<TrustedHTML>, GC::Root<TrustedScript>, GC::Root<TrustedScriptURL>, Utf16String> input, InjectionSink sink)
+{
+    auto& vm = global.vm();
+    auto& realm = *vm.current_realm();
+
+    // 1. Let defaultPolicy be the value of global’s trusted type policy factory’s default policy.
+    auto const& default_policy = as<HTML::WindowOrWorkerGlobalScopeMixin>(global).trusted_types()->default_policy();
+
+    // This algorithm routes a value to be assigned to an injection sink through a default policy, should one exist.
+    // FIXME: Open an issue upstream. It is not immediately clear what to do if the default policy does not exist.
```
AtkinsSJ, line 199: Please do open an issue! It looks like https://github.com/w3c/trusted-types/issues/ is the place for that.

Libraries/LibWeb/TrustedTypes/TrustedTypePolicy.cpp
```diff
@@ -182,4 +185,56 @@ WebIDL::ExceptionOr<JS::Value> TrustedTypePolicy::get_trusted_type_policy_value(
...
+    if (!default_policy)
+        return JS::js_null();
+
+    // 2. Let policyValue be the result of executing Get Trusted Type policy value, with the following arguments:
+    //  defaultPolicy as policy
+    //  stringified input as value
+    //  expectedType’s type name as trustedTypeName
+    //  « trustedTypeName, sink » as arguments
```
AtkinsSJ: Nit: Indent these with bullet points.

Libraries/LibWeb/TrustedTypes/TrustedTypePolicy.cpp
```diff
@@ -182,4 +185,56 @@ WebIDL::ExceptionOr<JS::Value> TrustedTypePolicy::get_trusted_type_policy_value(
...
+    auto arguments = GC::RootVector<JS::Value>(vm.heap());
+    arguments.append(JS::PrimitiveString::create(vm, to_string(trusted_type_name)));
+    arguments.append(JS::PrimitiveString::create(vm, to_string(sink)));
+    //  false as throwIfMissing
```
AtkinsSJ: Also move this with the rest of the comment, so it's clear it's connected.

Libraries/LibWeb/TrustedTypes/TrustedTypePolicy.cpp
```diff
@@ -182,4 +185,56 @@ WebIDL::ExceptionOr<JS::Value> TrustedTypePolicy::get_trusted_type_policy_value(
...
+    //  6. Return a new instance of an interface with a type name trustedTypeName, with its associated data value set to dataString.
+    switch (trusted_type_name) {
+    case TrustedTypeName::TrustedHTML:
+        return realm.create<TrustedHTML>(realm, move(data_string));
+    case TrustedTypeName::TrustedScript:
+        return realm.create<TrustedScript>(realm, move(data_string));
+    case TrustedTypeName::TrustedScriptURL:
+        return realm.create<TrustedScriptURL>(realm, move(data_string));
+    default:
+        VERIFY_NOT_REACHED();
```
AtkinsSJ: If you move the `VERIFY_NOT_REACHED()` outside of the switch, it behaves the same, but the compiler will complain if any new values get added to the enum and they're not handled here.

Libraries/LibWeb/ContentSecurityPolicy/Directives/RequireTrustedTypesFor.cpp
```diff
@@ -105,4 +106,76 @@ bool does_sink_require_trusted_types(JS::Object& global, String sink_group, Incl
...
+    auto result = Directive::Result::Allowed;
+
+    // 2. Let sample be source.
+    auto sample = source.substring_view(0);
+
+    // 3. If sink is "Function", then:
+    if (sink == TrustedTypes::InjectionSink::Function) {
+        // 1. If sample starts with "function anonymous", strip that from sample.
+        if (sample.starts_with("function anonymous"sv)) {
+            sample = sample.substring_view("function anonymous"sv.length());
```
AtkinsSJ: This pattern is a little spooky. `sample` is utf16, but I believe the `"function anonymous"sv` will be a plain `StringView`, which just counts bytes. (We seem to have definitions of `operator""sv` for both `StringView` and `Utf16StringView` somehow?) So you're turning a byte length into a code-point offset, which aren't the same thing.

cc @trflynn89, I haven't been following the Utf16 changes so maybe I'm wrong on this?

Libraries/LibWeb/TrustedTypes/TrustedTypePolicy.cpp
```diff
@@ -238,4 +238,78 @@ WebIDL::ExceptionOr<JS::Value> process_value_with_a_default_policy(TrustedTypeNa
...
+// https://www.w3.org/TR/trusted-types/#get-trusted-type-compliant-string-algorithm
+WebIDL::ExceptionOr<Utf16String> get_trusted_type_compliant_string(TrustedTypeName expected_type, JS::Object& global, Variant<GC::Root<TrustedHTML>, GC::Root<TrustedScript>, GC::Root<TrustedScriptURL>, Utf16String> input, InjectionSink sink, String sink_group)
+{
+    auto& vm = global.vm();
+
+    // 1. If input is an instance of expectedType, return stringified input and abort these steps.
+    switch (expected_type) {
+    case TrustedTypeName::TrustedHTML:
+        if (input.has<GC::Root<TrustedHTML>>())
+            return input.get<GC::Root<TrustedHTML>>()->to_string();
```
AtkinsSJ: `Variant::get_pointer<T>()` can be used to combine the `has()` and `get()` calls: it'll return null if it doesn't hold a `T`. Variant operations can be slow so it's nice to reduce the number of times we access its value.

Libraries/LibWeb/TrustedTypes/TrustedTypePolicy.cpp
```diff
@@ -238,4 +238,78 @@ WebIDL::ExceptionOr<JS::Value> process_value_with_a_default_policy(TrustedTypeNa
...
+    case TrustedTypeName::TrustedScript:
+        if (input.has<GC::Root<TrustedScript>>())
+            return input.get<GC::Root<TrustedScript>>()->to_string();
+        break;
+    case TrustedTypeName::TrustedScriptURL:
+        if (input.has<GC::Root<TrustedScriptURL>>())
+            return input.get<GC::Root<TrustedScriptURL>>()->to_string();
+        break;
+    default:
+        VERIFY_NOT_REACHED();
```
AtkinsSJ: Same point as before about not having a `default` case so that the compiler will warn us if/when a new enum value is added. (Also applies below.)

AtkinsSJ (changes requested):
Generally looks good to me! I'm not familiar with CSP, but I've had a look through and identified a few little things.

Libraries/LibWeb/TrustedTypes/InjectionSink.h
tete17: I am afraid this was the only way I knew.

It has to do with `ENUMERATE_GLOBAL_EVENT_HANDLERS` && `ENUMERATE_WINDOW_EVENT_HANDLERS`.

They both define the events in lowercase and I think there is no way to capitalize a string in c pre-processor.

Libraries/LibWeb/ContentSecurityPolicy/Directives/RequireTrustedTypesFor.cpp
tete17: I will be honest this also confused me. I didn't knew it was legal for the same operator to be defined twice. I feel like we are doing something illegal here.

Libraries/LibWeb/TrustedTypes/TrustedTypePolicy.cpp
tete17: Indeed https://github.com/w3c/trusted-types/issues/595

Libraries/LibWeb/ContentSecurityPolicy/Directives/RequireTrustedTypesFor.cpp
```diff
@@ -0,0 +1,181 @@
+/*
```
Lubrsi: suggestion: This file and its corresponding header could be moved into the TrustedTypes folder and namespace. Additionally, `Directive` should be added onto the end of the name to match the other directives.

Libraries/LibWeb/ContentSecurityPolicy/Directives/RequireTrustedTypesFor.cpp
```diff
@@ -0,0 +1,181 @@
...
+{
+    // 1. If request’s url’s scheme is not "javascript", return "Allowed" and abort further steps.
+    if (request->url().scheme() != "javascript"sv)
+        return Result::Allowed;
+
+    // 2. Let urlString be the result of running the URL serializer on request’s url.
+    auto url_string = request->url().serialize();
+
+    // 3. Let encodedScriptSource be the result of removing the leading "javascript:" from urlString.
+    auto const encoded_script_source = MUST(url_string.substring_from_byte_offset_with_shared_superstring("javascript:"sv.length()));
```
Lubrsi: question: I've not seen this String API before 😅 , could this be `substring_from_byte_offset` instead?

Libraries/LibWeb/ContentSecurityPolicy/Directives/RequireTrustedTypesFor.cpp
```diff
@@ -0,0 +1,181 @@
...
+    for (auto const policy : PolicyList::from_object(global)->policies()) {
+        // 1. If policy’s directive set does not contain a directive whose name is "require-trusted-types-for", skip to the next policy.
+        if (!policy->contains_directive_with_name(Names::RequireTrustedTypesFor))
+            continue;
+
+        // 2. Let directive be the policy’s directive set’s directive whose name is "require-trusted-types-for"
+        auto const directive = policy->get_directive_by_name(Names::RequireTrustedTypesFor);
+
+        // 3. If directive’s value does not contain a trusted-types-sink-group which is a match for sinkGroup, skip to the next policy.
+        if (directive->value().find_if([&sink_group](auto const& directive_value) { return directive_value.equals_ignoring_ascii_case(sink_group); }).is_end())
```
Lubrsi: suggestion: There's a bit too much going on in one line here, this could be:
```suggestion
        auto maybe_sink_group = directive->value().find_if([&sink_group](auto const& directive_value) {
            return directive_value.equals_ignoring_ascii_case(sink_group);
        });
        if (maybe_sink_group.is_end())
```

Libraries/LibWeb/ContentSecurityPolicy/Directives/RequireTrustedTypesFor.cpp
```diff
@@ -0,0 +1,181 @@
...
+        TrustedTypes::TrustedTypeName::TrustedScript,
+        request->client()->global_object(),
+        Utf16String::from_utf8(encoded_script_source),
+        TrustedTypes::InjectionSink::Locationhref);
+
+    // If that algorithm threw an error or convertedScriptSource is not a TrustedScript object, return "Blocked" and abort further steps.
+    if (converted_script_source.is_error() || !converted_script_source.value().is_object())
+        return Result::Blocked;
+
+    auto const converted_script_source_value = as_if<TrustedTypes::TrustedScript>(converted_script_source.value().as_object());
```
Lubrsi: suggestion: To make it clear that this is a pointer, add `*` after `const`
```suggestion
    auto const* converted_script_source_value = as_if<TrustedTypes::TrustedScript>(converted_script_source.value().as_object());
```

Libraries/LibWeb/ContentSecurityPolicy/Directives/RequireTrustedTypesFor.cpp
```diff
@@ -0,0 +1,181 @@
...
+    for (auto const policy : PolicyList::from_object(global)->policies()) {
+        // 1. If policy’s directive set does not contain a directive whose name is "require-trusted-types-for", skip to the next policy.
+        if (!policy->contains_directive_with_name(Names::RequireTrustedTypesFor))
+            continue;
+
+        // 2. Let directive be the policy’s directive set’s directive whose name is "require-trusted-types-for"
+        auto const directive = policy->get_directive_by_name(Names::RequireTrustedTypesFor);
+
+        // 3. If directive’s value does not contain a trusted-types-sink-group which is a match for sinkGroup, skip to the next policy.
+        if (directive->value().find_if([&sink_group](auto const& directive_value) { return directive_value.equals_ignoring_ascii_case(sink_group); }).is_end())
```
Lubrsi: suggestion: There's a bit too much going on in one line, this could be:
```suggestion
        auto maybe_sink_group = directive->value().find_if([&sink_group](auto const& directive_value) {
            return directive_value.equals_ignoring_ascii_case(sink_group);
        });
        if (maybe_sink_group.is_end())
```

Libraries/LibWeb/TrustedTypes/TrustedTypePolicy.cpp
```diff
@@ -169,4 +186,131 @@ WebIDL::ExceptionOr<JS::Value> TrustedTypePolicy::get_trusted_type_policy_value(
...
+            [](Utf16String& value) { return value; }),
+        arguments,
+        ThrowIfCallbackMissing::No));
+
+    //  4. If policyValue is null or undefined, return policyValue.
+    if (policy_value.is_nullish())
+        return policy_value;
+
+    //  5. Let dataString be the result of stringifying policyValue.
+    auto data_string = TRY(policy_value.to_utf16_string(vm));
```
Lubrsi: issue: TrustedScriptURL's callback returns USVString:
```suggestion
    Utf16String data_string;
    switch (trusted_type_name) {
    case TrustedTypeName::TrustedHTML:
    case TrustedTypeName::TrustedScript:
        data_string = TRY(WebIDL::to_utf16_string(vm, policy_value));
        break;
    case TrustedTypeName::TrustedScriptURL:
        data_string = TRY(WebIDL::to_utf16_usv_string(vm, policy_value));
        break;
    default:
        VERIFY_NOT_REACHED();
    }
```

Libraries/LibWeb/ContentSecurityPolicy/Directives/RequireTrustedTypesFor.cpp
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    // 2. Return false.
+    return false;
+}
+
+// https://w3c.github.io/trusted-types/dist/spec/#should-block-sink-type-mismatch
+Directive::Result should_sink_type_mismatch_violation_be_blocked_by_content_security_policy(JS::Object& global, TrustedTypes::InjectionSink sink, String sink_group, Utf16String source)
+{
+    auto& realm = *global.vm().current_realm();
```
Lubrsi: suggestion: This relies on the JavaScript executing when this runs, use relevant_realm to not rely on it:
```suggestion
    auto& realm = HTML::relevant_realm(global);
```

Libraries/LibWeb/TrustedTypes/TrustedTypePolicy.cpp
```diff
@@ -169,4 +186,131 @@ WebIDL::ExceptionOr<JS::Value> TrustedTypePolicy::get_trusted_type_policy_value(
...
+    case TrustedTypeName::TrustedScript:
+        VERIFY(is<TrustedScript>(converted_input.as_object()));
+        break;
+    case TrustedTypeName::TrustedScriptURL:
+        VERIFY(is<TrustedScriptURL>(converted_input.as_object()));
+        break;
+    }
+
+    // 8. Return stringified convertedInput.
+    return TRY(converted_input.to_utf16_string(vm));
```
Lubrsi: suggestion: It's not entirely clear by how the specification words this, but I believe this is intended to stringify the trusted type like in step 1.
```suggestion
    // 7. Assert: convertedInput is an instance of expectedType.
    // 8. Return stringified convertedInput.
    VERIFY(converted_input.is_object());
    switch (expected_type) {
    case TrustedTypeName::TrustedHTML:
        return as<TrustedHTML>(converted_input.as_object()).to_string();
    case TrustedTypeName::TrustedScript:
        return as<TrustedScript>(converted_input.as_object()).to_string();
    case TrustedTypeName::TrustedScriptURL:
        return as<TrustedScriptURL>(converted_input.as_object()).to_string();
    }
    
    VERIFY_NOT_REACHED();
```
I tested this in the other engines with this code, which would be great to add as one of our own tests:
```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <title>Document</title>
    <meta http-equiv="Content-Security-Policy" content="require-trusted-types-for 'script';">
</head>
<body>
    <script>
        trustedTypes.createPolicy("default", {
            createScript: () => {
                return {
                    toString() {
                        console.log("createScript getter called");
                        return "hello from getter";
                    }
                }
            },
        });

        TrustedScript.prototype.toString = () => {
            console.log("overriden toString called");
            return "overriden";
        };

        const scr = document.createElement("script");
        scr.innerText = "test";
        console.log("scr.innerText:", scr.innerText);
    </script>
</body>
</html>
```

Libraries/LibWeb/TrustedTypes/TrustedTypePolicy.cpp
```diff
@@ -169,4 +186,131 @@ WebIDL::ExceptionOr<JS::Value> TrustedTypePolicy::get_trusted_type_policy_value(
...
+    // 6. If convertedInput is null or undefined, execute the following steps:
+    if (converted_input.is_nullish()) {
+        // 1. Let disposition be the result of executing Should sink type mismatch violation be blocked by Content Security Policy?
+        //    algorithm, passing global, stringified input as source, sinkGroup and sink.
+        auto const disposition = ContentSecurityPolicy::Directives::should_sink_type_mismatch_violation_be_blocked_by_content_security_policy(
+            global,
+            sink,
+            sink_group,
+            input.visit(
+                [](auto&& value) { return value->to_string(); },
```
Lubrsi: issue: `&&` implies that the value is being moved into the callback, but it is passed in by reference. Ditto for all the other places that use this.
```suggestion
                [](auto& value) { return value->to_string(); },
```

Libraries/LibWeb/TrustedTypes/TrustedTypePolicy.cpp
```diff
@@ -169,4 +186,131 @@ WebIDL::ExceptionOr<JS::Value> TrustedTypePolicy::get_trusted_type_policy_value(
...
+            return (*value)->to_string();
+        break;
+    }
+
+    // 2. Let requireTrustedTypes be the result of executing Does sink type require trusted types? algorithm, passing global, sinkGroup, and true.
+    auto const require_trusted_types = ContentSecurityPolicy::Directives::does_sink_require_trusted_types(global, sink_group, ContentSecurityPolicy::Directives::IncludeReportOnlyPolicies::Yes);
+
+    // 3. If requireTrustedTypes is false, return stringified input and abort these steps.
+    if (!require_trusted_types)
+        return input.visit([](auto&& value) { return value->to_string(); }, [](Utf16String& value) { return value; });
```
Lubrsi: suggestion: This can put across multiple lines:
```suggestion
    if (!require_trusted_types) {
        return input.visit(
            [](auto& value) {
                return value->to_string();
            },
            [](Utf16String& value) {
                return value;
            });
    }
```

Libraries/LibWeb/TrustedTypes/TrustedTypePolicy.cpp
```diff
@@ -169,4 +186,131 @@ WebIDL::ExceptionOr<JS::Value> TrustedTypePolicy::get_trusted_type_policy_value(
...
+        auto const disposition = ContentSecurityPolicy::Directives::should_sink_type_mismatch_violation_be_blocked_by_content_security_policy(
+            global,
+            sink,
+            sink_group,
+            input.visit(
+                [](auto&& value) { return value->to_string(); },
+                [](Utf16String& value) { return value; }));
+
+        // 2. If disposition is “Allowed”, return stringified input and abort further steps.
+        if (disposition == ContentSecurityPolicy::Directives::Directive::Result::Allowed)
```
Lubrsi: ditto about using multiple lines

Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.cpp
```diff
@@ -134,6 +135,8 @@ Optional<Utf16String> TrustedTypePolicyFactory::get_property_type(Utf16String co
 
 TrustedTypePolicyFactory::TrustedTypePolicyFactory(JS::Realm& realm)
     : PlatformObject(realm)
+    , m_empty_html(realm.create<TrustedHTML>(realm, ""_utf16))
+    , m_empty_script(realm.create<TrustedScript>(realm, ""_utf16))
```
Lubrsi: suggestion: We typically allocate constant objects like this on demand when they're first requested. For example: https://github.com/LadybirdBrowser/ladybird/blob/ce03b8b3b1d9998b3936fa2f816a34270d287724/Libraries/LibWeb/HTML/Window.cpp#L1106-L1114

Libraries/LibWeb/ContentSecurityPolicy/Directives/RequireTrustedTypesFor.cpp
tete17: Sure thing :)

Libraries/LibWeb/TrustedTypes/TrustedTypePolicy.cpp
tete17: Really nice catch :)

Libraries/LibWeb/TrustedTypes/TrustedTypePolicy.cpp
tete17: I think I cover them all let me know if I missed any :)

Libraries/LibWeb/ContentSecurityPolicy/Directives/RequireTrustedTypesFor.cpp
tete17: We relay on js being executed because we go through the `.vm()` function?

I will be honest I tend to prompt the editor autocomplete until a function that gets me a realm and don't think too much about it. Also what is the difference between a realm and the global object?

Libraries/LibWeb/TrustedTypes/TrustedTypePolicy.cpp
tete17: I am afraid I won't be able to add this test until after another MR since I don't have the plumbing in place to affect src.innerText in this MR.

I also don't think this is the part where this test would be affected right?

`get_trusted_type_complaint_string` allready drop us a guarenteed TrustedXXX type. This means calling .to_utf16_string should yeild the same result as casting it to the right type and calling the toString function directly.

I think the interesting part happend in `process_value_with_a_default_policy`. Chrome seems to agree with this implementation as in your example the toString function is being called and so should we right? Mainly because we use WebIDL::to_utf16_(usv_)string and that should call the underlayng toString object right?

Lubrsi: For prosperity sake since this was discussed on Discord:
The issue is that even though get_trusted_type_complaint_string is guaranteed to give a Trusted Type, ToString treats the passed in object as generic and will observably lookup Symbol.toPrimitive and toString and call them if they exist, which can run arbitrary JS.

The other implementations don't do this and thus don't print `"overriden toString called"`.

There's a NOTE in the spec (which would be nice to convert to an abstract operation) that stringifying a Trusted Type will return the internal data attribute, instead of trying to call toString:
> All Trusted Types stringifiers return the inner string value. This makes it easy to incrementally migrate the application code into using Trusted Types in place of DOM strings (it’s possible to start producing types in parts of the application, while still using and accepting strings in other parts of the codebase). In that sense, Trusted Types are backwards-compatible with the regular DOM APIs.

https://www.w3.org/TR/trusted-types/#trusted-types

tete17: I will add an explicit test for this case on the next MR when trusted type enforcement happens in document.innerText

tete17:
Thank @AtkinsSJ & @Lubrsi both for the review and the good discussions :)

This should cover all the review items.

Libraries/LibWeb/TrustedTypes/TrustedTypePolicy.cpp
```diff
@@ -169,4 +186,154 @@ WebIDL::ExceptionOr<JS::Value> TrustedTypePolicy::get_trusted_type_policy_value(
     return policy_value;
 }
 
+// https://www.w3.org/TR/trusted-types/#process-value-with-a-default-policy-algorithm
+WebIDL::ExceptionOr<JS::Value> process_value_with_a_default_policy(TrustedTypeName trusted_type_name, JS::Object& global, Variant<GC::Root<TrustedHTML>, GC::Root<TrustedScript>, GC::Root<TrustedScriptURL>, Utf16String> input, InjectionSink sink)
```
Lubrsi: suggestion: This could return TrustedType instead of the generic JS::Value:
```suggestion
WebIDL::ExceptionOr<Optional<TrustedType>> process_value_with_a_default_policy(TrustedTypeName trusted_type_name, JS::Object& global, Variant<GC::Root<TrustedHTML>, GC::Root<TrustedScript>, GC::Root<TrustedScriptURL>, Utf16String> input, InjectionSink sink)
```

tete17: I am not 100% in love with the change as it deviates slight more from the spec than I would love.

It is not a big deal so I made the change anyway and I will let you be the judge of what do you like more.
It is true than with this approach it is harder to make bugs like the ones we discuss about prototype pollution and "stringification" so that is that :)

Libraries/LibWeb/TrustedTypes/TrustedTypePolicy.cpp
```diff
@@ -169,4 +186,158 @@ WebIDL::ExceptionOr<JS::Value> TrustedTypePolicy::get_trusted_type_policy_value(
...
+        break;
+    case TrustedTypeName::TrustedScript:
+        if (auto* const value = input.get_pointer<GC::Root<TrustedScript>>(); value)
+            return (*value)->to_string();
+        break;
+    case TrustedTypeName::TrustedScriptURL:
+        if (auto* const value = input.get_pointer<GC::Root<TrustedScriptURL>>(); value)
+            return (*value)->to_string();
+        break;
+    }
```
Lubrsi, line 274: nit: We typically put the `*` after `const`:
```suggestion
    switch (expected_type) {
    case TrustedTypeName::TrustedHTML:
        if (auto const* value = input.get_pointer<GC::Root<TrustedHTML>>(); value)
            return (*value)->to_string();
        break;
    case TrustedTypeName::TrustedScript:
        if (auto const* value = input.get_pointer<GC::Root<TrustedScript>>(); value)
            return (*value)->to_string();
        break;
    case TrustedTypeName::TrustedScriptURL:
        if (auto const* value = input.get_pointer<GC::Root<TrustedScriptURL>>(); value)
            return (*value)->to_string();
        break;
    }
```

Lubrsi (changes requested):
LGTM, just one final nitpick

Libraries/LibWeb/TrustedTypes/TrustedTypePolicy.cpp
tete17: I think they even mean different things.

It used to be a constant pointer to a type while now it is a pointer to a constant type :)

Lubrsi:
LGTM, thanks again for your patience :^)

tete17:
Thanks to you and all the team for the nice feedback :)

PR #5832, opened 2025-08-12
CI: Remove Microphone access step for macOS

gmta:
This no longer seems to be necessary.


gmta:
A CI run on `ladybird-mac-runner` has shown this to work 🥳

PR #5833, opened 2025-08-12
LibWeb: Discard inaccuracies when interpolating `rotate`

Calme1709:


tcl3 (approved):
Thanks for tracking this down

PR #5836, opened 2025-08-12
LibWeb: Implement HTMLElement.draggable()

lukasxcs:
This passes some (not all) of the draggable WPT tests.


Libraries/LibWeb/HTML/HTMLElement.cpp
```diff
@@ -2093,4 +2094,44 @@ String HTMLElement::access_key_label() const
...
 }
 
+// https://html.spec.whatwg.org/multipage/dnd.html#attr-draggable
+bool HTMLElement::draggable() const
+{
+    auto attribute = get_attribute(HTML::AttributeNames::draggable);
+
+    if (!attribute.has_value()) {
+        return false;
+    }
```
tcl3: I don't think this is correct. If the `draggable` content attribute is not set then it has the state `Auto` and is handled by the code below. Instead, you could remove this and change the if statements below to make sure `attribute_has_value()` is true.

Libraries/LibWeb/HTML/HTMLElement.h
```diff
@@ -157,6 +157,10 @@ class HTMLElement
 
     bool is_inert() const { return m_inert; }
 
+    bool draggable() const;
+    bool is_draggable() const { return m_draggable; }
```
tcl3: AFAICT `is_draggable()` isn't used and can be removed.

Libraries/LibWeb/HTML/HTMLElement.h
```diff
@@ -157,6 +157,10 @@ class HTMLElement
...
+    void set_draggable(bool draggable) { m_draggable = draggable; }
```
tcl3: Rather than having `set_draggable()` set a member variable, it should set the content attribute to the appropriate value. So this function should do something like: `MUST(set_attribute(HTML::AttributeNames::draggable, draggable ? "true"_string : "false"_string ));`

Libraries/LibWeb/HTML/HTMLElement.cpp
```diff
@@ -2093,4 +2094,44 @@ String HTMLElement::access_key_label() const
     return String {};
 }
 
+// https://html.spec.whatwg.org/multipage/dnd.html#attr-draggable
```
tcl3: This is the spec link for the `draggable` content attribute. What you want is the link for the `draggable` IDL attribute, which I believe is: https://html.spec.whatwg.org/multipage/dnd.html#dom-draggable.

Libraries/LibWeb/HTML/HTMLElement.cpp
```diff
@@ -2093,4 +2094,44 @@ String HTMLElement::access_key_label() const
...
+
+    // If the element is an img element, the draggable IDL attribute must return true.
+    if (is<HTML::HTMLImageElement>(*this)) {
+        return true;
+    }
+
+    // If the element is an object element that represents an image, the draggable IDL attribute must return true.
+    if (is<HTML::HTMLObjectElement>(*this) && has_attribute(HTML::AttributeNames::type) && get_attribute(HTML::AttributeNames::type)->equals_ignoring_ascii_case("image"sv)) {
+        return true;
+    }
```
tcl3: I think you can rearrange this slightly to avoid the call to `has_attribute()`. Something like:

```c++
if (is<HTML::HTMLObjectElement>(*this)) {
    if (auto type_attribute = get_attribute(HTML::AttributeNames::type); type_attribute->equals_ignoring_ascii_case("image"sv))
        return true;
}
```

Libraries/LibWeb/HTML/HTMLElement.cpp
tcl3: It looks like you beat me to this. Just a nitpick but personally I would rewrite it as:
```c++
if (attribute.has_value() && attribute->equals_ignoring_ascii_case("true"sv)) {
    return true;
}

// If an element's draggable content attribute has the state False, the draggable IDL attribute must return false.
if (attribute.has_value() && attribute->equals_ignoring_ascii_case("false"sv)) {
    return false;
}
```
to avoid the extra level of indentation.

lukasxcs:
Thanks for the feedback! Should i force-push everything into one commit?

tcl3:
> Thanks for the feedback! Should i force-push everything into one commit?

Yes, please.

lukasxcs:
All tests do indeed pass now :) Thanks alot

tcl3:
> All tests do indeed pass now :) 

:tada: 

Feel free to close comments after you've dealt with review feedback.

Libraries/LibWeb/HTML/HTMLElement.cpp
```diff
@@ -2093,4 +2094,41 @@ String HTMLElement::access_key_label() const
...
+    // Otherwise, the element's draggable content attribute has the state Auto.
+
+    // If the element is an img element, the draggable IDL attribute must return true.
+    if (is<HTML::HTMLImageElement>(*this)) {
+        return true;
+    }
+
+    // If the element is an object element that represents an image, the draggable IDL attribute must return true.
+    if (is<HTML::HTMLObjectElement>(*this)) {
+        if (auto type_attribute = get_attribute(HTML::AttributeNames::type); type_attribute->equals_ignoring_ascii_case("image"sv))
```
tcl3: Sorry, this should actually be `if (auto type_attribute = get_attribute(HTML::AttributeNames::type); type_attribute.has_value() && type_attribute->equals_ignoring_ascii_case("image"sv))`.

We need to check whether type_attribute has a value before trying to access it.

tcl3 (approved):
This fixes a total of 167 WPT subtests in the `html/editing/dnd` directory :tada:

lukasxcs:
Nice! Thanks for your help :) Looking forward to contributing some more.

PR #5839, opened 2025-08-12
Everywhere: Cleanup some platform code and remove unused code

cqundefine:
This includes a few things I've noticed while looking at portability and working on FreeBSD support.

This PR removes ifdefs for:
- Emscripten / WASM - it doesn't really make much sense to compile Ladybird to WASM
- SerenityOS - this was mostly a remnant of before the fork and getting this tree building on Serenity in it's current state isn't really feasible, this had the side effect of showing a couple of functions / files that did nothing on all other platforms

This also brings up a question if it would be okay to remove some more ifdefs:
- There are the supported platforms like Linux, MacOS and to some extent Android and Windows
- I'm keeping eye on FreeBSD building and working
- OpenBSD shouldn't be that difficult to get working as there are up to date patches to Chromium that support both it and FreeBSD which means the tricky dependencies like skia or angle could be gotten to work
- iOS is not really supported right now in any practical way but it's worth leaving for the future
- And then there are platforms like NetBSD, DragonFly, Solaris and Haiku that are unsupported by vcpkg and it might be not be as easy to get some of the dependencies working and they're most probably broken in the current state of this tree. So maybe we should get rid of those? Tt would clean some of the ifdef soup quite nicely.

### DIFF
--- a/AK/Types.h
+++ b/AK/Types.h
@@ -111,10 +111,6 @@ using Conditional = typename __Conditional<condition, TrueType, FalseType>::Type
 #include <stdint.h>
 #include <sys/types.h>
 
-#ifdef __ptrdiff_t
-using __ptrdiff_t = __PTRDIFF_TYPE__;
-#endif
-
 #if defined(AK_OS_WINDOWS)
 using ssize_t = AK::Detail::MakeSigned<size_t>;
 using mode_t = unsigned short;


Libraries/LibCore/Process.h
```diff
@@ -63,15 +62,10 @@ class Process {
...
-    static ErrorOr<Process> spawn(StringView path, ReadonlySpan<StringView> arguments, ByteString working_directory = {}, KeepAsChild keep_as_child = KeepAsChild::No);
+    static ErrorOr<Process> spawn(StringView path, ReadonlySpan<ByteString> arguments, KeepAsChild keep_as_child = KeepAsChild::No);
+    static ErrorOr<Process> spawn(StringView path, ReadonlySpan<StringView> arguments, KeepAsChild keep_as_child = KeepAsChild::No);
 
     static ErrorOr<String> get_name();
-    enum class SetThreadName {
-        No,
-        Yes,
-    };
-    static ErrorOr<void> set_name(StringView, SetThreadName = SetThreadName::No);
```
R-Goc, line 74: While this is indeed not implementable on Windows, it can be done on linux with prctl or a glibc specific function. So that depends on whether we want to be naming threads or not.

R-Goc:
This will need a windows label. It touches a lot of ifdefs around platforms.

Libraries/LibCore/Process.cpp
```diff
@@ -327,11 +242,7 @@ pid_t Process::pid() const
 ErrorOr<void> Process::disown()
```
rgret-dev: Couldn't `disown()` be removed completely if it does nothing?
(along with `m_should_disown`)

ADKaster:
> GNU/Hurd - build for this was broken for quite a long while, the platform itself is quite broken and to my knowledge no other browser runs on it at all

fwiw, @bugaevc was moderately supporting this platform before we went wild with extra dependencies. If the ifdef code isn't hurting anything, I wouldn't mind keeping it around for when someone Hurd-adjacent wrangles our deps and makes it build again.

ADKaster:
> Emscripten / WASM - it doesn't really make much sense to compile Ladybird to WASM

This was in place for some fun times getting and hosting webpage with a js repl compiled to wasm. we already removed most of that code and infra, though it lives on in the SerenityOS repo. So no qualms with removing that.

Libraries/LibCore/Process.cpp
```diff
@@ -94,15 +94,6 @@ ErrorOr<Process> Process::spawn(ProcessSpawnOptions const& options)
         posix_spawn_file_actions_destroy(&spawn_actions);
     };
 
-    if (options.working_directory.has_value()) {
-#ifdef AK_OS_SERENITY
-        CHECK(posix_spawn_file_actions_addchdir(&spawn_actions, options.working_directory->characters()));
```
ADKaster, line 103: true. we used this for the test-runner on Serenity. it might :tm: be useful at some point, but if we're not using it, we're not using it.

AK/Assertions.cpp
```diff
@@ -31,11 +31,7 @@
 #    include <cxxabi.h>
 #endif
 
-#if defined(AK_OS_SERENITY)
-#    define ERRORLN dbgln
-#else
-#    define ERRORLN warnln
-#endif
+#define ERRORLN warnln
```
ADKaster: should just replace all uses of this macro with warnln

AK/Math.h
```diff
@@ -631,13 +611,7 @@ constexpr T tan(T angle)
 
     return ret;
 #else
-#    if defined(AK_OS_SERENITY)
```
ADKaster, line 634: I'm skeptical whether we *really* need Math.h. There's been murmurs over in SerenityOS for years that we should only be using these builtins and such in the implementation of `libm`. But that's a giant yak in of itself.

AK/Types.h
```diff
@@ -107,42 +107,18 @@ using Conditional = typename __Conditional<condition, TrueType, FalseType>::Type
...
-using pid_t = int;
-
-#else
-#    include <stddef.h>
-#    include <stdint.h>
-#    include <sys/types.h>
-
-#    ifdef __ptrdiff_t
+#ifdef __ptrdiff_t
 using __ptrdiff_t = __PTRDIFF_TYPE__;
```
ADKaster: huh. do we actually use this type? feels like a relic

Libraries/LibCore/Process.cpp
ADKaster: yes, that concept is unused. even in Serenity I think it's unused these days.

Libraries/LibCore/Process.cpp
ADKaster: also there's a hunk from a subsequent patch that removes code that was relying on this feature, so perhaps it should be moved back.

ADKaster:
> platforms like NetBSD, DragonFly, Solaris and Haiku 

If they can build the AK and LibCore targets, I don't see a reason to remove the code. Now, getting a build with just those targets might be a bit exciting to test. But I doubt we've broken them *too* badly. Haiku and Solaris used to be somewhat upkept by a Haiku contributor, but after the split from SerenityOS and adding of dependencies they ... lost interest let's say. From what I remember scanning the relevant Haiku discourse threads, there is still *interest* in a port. Just not from the original contributor who was submitting patches.

Libraries/LibCore/Process.h
cqundefine: While I don't doubt this could be implemented, this function was never called anywhere so I think it's safe to get rid of it and to add it again with an actual implementation if needed.

AK/Math.h
cqundefine: I think tackling Math.h might be a bit harder, not sure what should be done instead, using the builtins directly in other places, using the system math library, if so how much would losing constexpr hurt?

Libraries/LibCore/Process.cpp
cqundefine: Are you sure there's code later that relies on this? I couldn't find any.

ADKaster: Maybe it was later in the same commit... The everywhere: remove serenity stuff commit towards the end of the PR is quite chonky

cqundefine: You think I should split that commit up?

AK/Math.h
R-Goc: From what I [see](https://en.cppreference.com/w/cpp/compiler_support/23) no STL implementation implements [p0533r9](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p0533r9.pdf)(C++23) constexpr for cmath and cstdlib, let alone [p1383r2](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p1383r2.pdf)(C++26) more constexpr for cmath. Not sure about the state in the other libraries, but the issue for it in MSVC/STL doesn't really have a timeline and is blocked on the compiler/UCRT team. see https://github.com/microsoft/STL/issues/2530

cqundefine:
> > GNU/Hurd - build for this was broken for quite a long while, the platform itself is quite broken and to my knowledge no other browser runs on it at all
> 
> fwiw, @bugaevc was moderately supporting this platform before we went wild with extra dependencies. If the ifdef code isn't hurting anything, I wouldn't mind keeping it around for when someone Hurd-adjacent wrangles our deps and makes it build again.

I actually reverted the GNU/Hurd commit, it doesn't change that much

bugaevc:
> GNU/Hurd - build for this was broken for quite a long while, the platform itself is quite broken

quite broken how?

> and to my knowledge no other browser runs on it at all

that, if true, makes it so much more important to keep Ladybird working

AK/Math.h
Hendiadyoin1: So from my side of things:
Having a templated version of math things is way nicer than plain libm,
It is less footgun-y
And in the past allowed us to force constexpr in a few more cases
(As well as inline things more eagerly before we had proper sse support)

But going forward it might be better to strip this header of it's implementations a bit more fully relying on builtins where possible,
Note that in the past some trivial functions where not constexpr, which is why we have some workarounds for these only in the constexpr case (IIRC)

R-Goc: Do we have an easy to access list of which functions are constexpr? Looking at llvm at least there are still builtins that are not constexpr, mainly things related to fp. AFAIK on windows nothing really is constexpr as LLVM libc is not supported on windows.

cqundefine:
> quite broken how?

Just looking at the FOSS ecosystem I felt like getting all the dependencies wrangled up would be the most difficult, also looking at the fact that Ladybird seems to target mostly Wayland + Vulkan, I'm not sure that is that mature on Hurd, but I don't have anything particular against keeping code for it.

> that, if true, makes it so much more important to keep Ladybird working

That comment was mostly about the fact that if no other browser supports it, it might be the most difficult to get working.

PR #5840, opened 2025-08-12
LibWeb: Implement VTTCue.{line,position}

jamierocks:


Libraries/LibWeb/WebVTT/VTTCue.h
```diff
@@ -42,6 +44,9 @@ class VTTCue final : public HTML::TextTrackCue {
     bool snap_to_lines() const { return m_snap_to_lines; }
     void set_snap_to_lines(bool snap_to_lines) { m_snap_to_lines = snap_to_lines; }
 
+    LineAndPositionSetting line() const { return m_line; }
+    void set_line(LineAndPositionSetting line) { m_line = line; }
```
gmta, line 48: Similarly to `.position`, shouldn't we have a `computed_line()` as well?

jamierocks: There [is](https://w3c.github.io/webvtt/#cue-computed-line), but so much of it would currently need to be FIXME'd.

Happy to add what could be done though :^) if you'd like

gmta: Please! I don't mind it turning out to be a minefield of `FIXME`s, it will give some pointers on actually doing something with the line value.

jamierocks: Done :) To be fair, it's not as FIXME heavy as I seemed to think.

Libraries/LibWeb/WebVTT/VTTCue.cpp
```diff
@@ -156,6 +157,26 @@ double VTTCue::computed_line()
     return n;
 }
 
+// https://w3c.github.io/webvtt/#cue-computed-position
+double VTTCue::computed_position()
+{
+    // 1. If the position is numeric between 0 and 100, then return the value of the position and abort these steps.
+    //    (Otherwise, the position is the special value auto.)
+    if (m_position.has<double>())
```
gmta: This is missing the "between 0 and 100" check

jamierocks: Good catch! Fixed :)

jamierocks:
Hmm, 2 different tests have failed on each build. I'm not convinced this PR is causing the errors, especially as it's only failing on one platform, but happy to be corrected :)

PR #5841, opened 2025-08-13
RequestServer: Assortment of small changes

gmta:


Services/RequestServer/ConnectionFromClient.cpp
```diff
@@ -96,6 +96,8 @@ ByteString build_curl_resolve_list(DNS::LookupResult const& dns_result, StringVi
         resolve_opt_builder.append(formatted_address);
     }
 
+    dbgln_if(REQUESTSERVER_DEBUG, "RequestServer: Resolve list: {}", MUST(resolve_opt_builder.to_string()));
+
     return resolve_opt_builder.to_byte_string();
```
trflynn89: We can't call `to_byte_string()` after `to_string()` - if its byte buffer was outlined, it will have been handed off to the `String` for keeping. You can log `resolve_opt_builder.string_view()` instead.

Services/RequestServer/ConnectionFromClient.cpp
```diff
@@ -331,8 +331,7 @@ ConnectionFromClient::ConnectionFromClient(NonnullOwnPtr<IPC::Transport> transpo
     set_option(CURLMOPT_TIMERDATA, this);
 
     m_timer = Core::Timer::create_single_shot(0, [this] {
-        int still_running = 0;
-        auto result = curl_multi_socket_action(m_curl_multi, CURL_SOCKET_TIMEOUT, 0, &still_running);
+        auto result = curl_multi_socket_action(m_curl_multi, CURL_SOCKET_TIMEOUT, 0, nullptr);
```
trflynn89, line 332: There's 2 other `curl_multi_socket_action` calls that also do nothing with this param if we want to clean those up too

Services/RequestServer/ConnectionFromClient.cpp
gmta: Ooh that's a footgunny API if I ever saw one. Thanks!

Services/RequestServer/ConnectionFromClient.cpp
gmta: Nice, cleaned up those remaining two as well.

PR #5846, opened 2025-08-13
LibWeb: Implement DataTransfer.set_data()

lukasxcs:
The tests fail because there is a bug, which makes Safari fail them aswell (https://bugs.webkit.org/show_bug.cgi?id=266890)

### DIFF
--- a/Libraries/LibWeb/HTML/DataTransfer.cpp
+++ b/Libraries/LibWeb/HTML/DataTransfer.cpp
@@ -182,6 +182,47 @@ String DataTransfer::get_data(String const& format_argument) const
     return MUST(String::from_utf8(result));
 }
 
+// https://html.spec.whatwg.org/multipage/dnd.html#dom-datatransfer-setdata
+void DataTransfer::set_data(String const& format_argument, String const& value)
+{
+    // 1. If the DataTransfer object is no longer associated with a drag data store, then return.
+    if (!m_associated_drag_data_store)
+        return;
+
+    // 2. If the drag data store's mode is not the read/write mode, return. Nothing happens.
+    if (m_associated_drag_data_store->mode() != DragDataStore::Mode::ReadWrite)
+        return;
+
+    // 3. Let format be the first argument, converted to ASCII lowercase.
+    auto format = format_argument.to_ascii_lowercase();
+
+    // 4. If format equals "text", change it to "text/plain".
+    //    If format equals "url", change it to "text/uri-list".
+    if (format == "text"sv) {
+        format = "text/plain"_string;
+    } else if (format == "url"sv) {
+        format = "text/uri-list"_string;
+    }
+
+    // 5. Remove the item in the drag data store item list whose kind is text and whose type string is equal to format, if there is one.
+    for (auto const& [i, item] : enumerate(m_associated_drag_data_store->item_list())) {
+        if (item.kind == DragDataStoreItem::Kind::Text && item.type_string == format) {
+            remove_item(i);
+            break;
+        }
+    }
+
+    // 6. Add an item to the drag data store item list
+    // whose kind is text, whose type string is equal to format, and whose data is the string given by the method's second argument.
+    DragDataStoreItem item {};
+    item.kind = DragDataStoreItem::Kind::Text;
+    item.type_string = format;
+    item.data = MUST(ByteBuffer::copy(value.bytes()));
+    add_item(move(item));
+
+    update_data_transfer_types_list();
+}
+
 // https://html.spec.whatwg.org/multipage/dnd.html#dom-datatransfer-files
 GC::Ref<FileAPI::FileList> DataTransfer::files() const
 {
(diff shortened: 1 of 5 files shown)


Libraries/LibWeb/HTML/DataTransfer.cpp
```diff
@@ -180,6 +180,45 @@ String DataTransfer::get_data(String const& format_argument) const
     return MUST(String::from_utf8(result));
 }
 
+// https://html.spec.whatwg.org/multipage/dnd.html#dom-datatransfer-setdata
+void DataTransfer::set_data(String const& format_argument, String value)
```
tcl3: Changing `String value` to `String const& value` avoids a copy here.

Libraries/LibWeb/HTML/DataTransfer.cpp
```diff
@@ -180,6 +180,45 @@ String DataTransfer::get_data(String const& format_argument) const
...
+        if (item.kind == DragDataStoreItem::Kind::Text && item.type_string == format)
+            m_associated_drag_data_store->remove_item_at(i);
+    }
+
+    // 6. Add an item to the drag data store item list
+    // whose kind is text, whose type string is equal to format, and whose data is the string given by the method's second argument.
+    DragDataStoreItem item {};
+    item.kind = DragDataStoreItem::Kind::Text;
+    item.type_string = format;
+    item.data = value.to_byte_string().to_byte_buffer();
```
tcl3: This creates a new `ByteString` object before copying it into a `ByteBuffer`. You should be able to do `MUST(ByteBuffer::copy(value.bytes()));` to avoid creating the intermediate `ByteString` object

Libraries/LibWeb/HTML/DataTransfer.cpp
```diff
@@ -180,6 +180,45 @@ String DataTransfer::get_data(String const& format_argument) const
...
+    if (format == "text"sv) {
+        format = "text/plain"_string;
+    } else if (format == "url"sv) {
+        format = "text/uri-list"_string;
+    }
+
+    // 5. Remove the item in the drag data store item list whose kind is text and whose type string is equal to format, if there is one.
+    for (auto const& [i, item] : enumerate(m_associated_drag_data_store->item_list())) {
+        if (item.kind == DragDataStoreItem::Kind::Text && item.type_string == format)
+            m_associated_drag_data_store->remove_item_at(i);
```
tcl3: Is it true that this condition can only be true for a single item in the list? If so, then it should be safe to `break` after removing the item. 

If it is possible that more than one item could be removed then you need to make sure that `i` points to the correct index after each removal.

tcl3 (commented):
It looks like the tests you've imported fail because our `FrozenArray` implementation isn't exactly to spec and getting the same unmodified `FrozenArray` multiple times doesn't return the same object.

I've created an issue for this (#5848), since it seems to be outside the scope of this PR.

Libraries/LibWeb/HTML/DataTransfer.cpp
lukasxcs: It can only be true for one, you're right.

Libraries/LibWeb/HTML/DataTransfer.cpp
```diff
@@ -180,6 +180,47 @@ String DataTransfer::get_data(String const& format_argument) const
...
+        }
+    }
+
+    // 6. Add an item to the drag data store item list
+    // whose kind is text, whose type string is equal to format, and whose data is the string given by the method's second argument.
+    DragDataStoreItem item {};
+    item.kind = DragDataStoreItem::Kind::Text;
+    item.type_string = format;
+    item.data = MUST(ByteBuffer::copy(value.bytes()));
+    m_associated_drag_data_store->add_item(move(item));
```
tcl3: I think this needs to be just `add_item(move(item))` instead. This ensures that `DataTransfer::m_item_list` and `DragDataStore::m_item_list` remain in sync.

Libraries/LibWeb/HTML/DataTransfer.cpp
```diff
@@ -180,6 +180,47 @@ String DataTransfer::get_data(String const& format_argument) const
...
+    if (format == "text"sv) {
+        format = "text/plain"_string;
+    } else if (format == "url"sv) {
+        format = "text/uri-list"_string;
+    }
+
+    // 5. Remove the item in the drag data store item list whose kind is text and whose type string is equal to format, if there is one.
+    for (auto const& [i, item] : enumerate(m_associated_drag_data_store->item_list())) {
+        if (item.kind == DragDataStoreItem::Kind::Text && item.type_string == format) {
+            m_associated_drag_data_store->remove_item_at(i);
```
tcl3: I've added a `DataTransfer::remove_item(size_t index)` method in #6088, I think that method should be used here instead - to ensure the `DataTransfer` and `DragDataStore` item lists remain in sync.

lukasxcs:
Rebased your changes and the tests do pass now :)

lukasxcs:
well, ive screwed that up :D

PR #5847, opened 2025-08-13
WebDriver: Make WebDriver unix socket fixed length

AronHetLam:
Unix socket max length of 104 characters on mac and 109 on Linux makes runing WPT locally unable to connect to the WebDriver for users with long usernames.
This places the socket in `/tmp/...` excluding the username from the socket path.

See [discusstion in test channel](https://discord.com/channels/1247070541085671459/1247090166909374474/1402663665722720286)

### DIFF
--- a/Services/WebDriver/Session.cpp
+++ b/Services/WebDriver/Session.cpp
@@ -254,7 +254,7 @@ ErrorOr<void> Session::start(LaunchBrowserCallback const& launch_browser_callbac
 {
     auto promise = ServerPromise::construct();
 
-    m_web_content_socket_path = ByteString::formatted("{}/webdriver/session_{}_{}", TRY(Core::StandardPaths::runtime_directory()), getpid(), m_session_id);
+    m_web_content_socket_path = ByteString::formatted("{}/webdriver/session_{}_{}", Core::StandardPaths::tempfile_directory(), getpid(), m_session_id);
     m_web_content_server = TRY(create_server(promise));
 
     m_browser_process = TRY(launch_browser_callback(*m_web_content_socket_path, m_options.headless));
--- a/Services/WebDriver/main.cpp
+++ b/Services/WebDriver/main.cpp
@@ -100,7 +100,7 @@ ErrorOr<int> ladybird_main(Main::Arguments arguments)
 
     Web::WebDriver::set_default_interface_mode(headless ? Web::WebDriver::InterfaceMode::Headless : Web::WebDriver::InterfaceMode::Graphical);
 
-    auto webdriver_socket_path = ByteString::formatted("{}/webdriver", TRY(Core::StandardPaths::runtime_directory()));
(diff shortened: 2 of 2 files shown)


Services/WebDriver/main.cpp
```diff
@@ -100,7 +100,7 @@ ErrorOr<int> ladybird_main(Main::Arguments arguments)
 
     Web::WebDriver::set_default_interface_mode(headless ? Web::WebDriver::InterfaceMode::Headless : Web::WebDriver::InterfaceMode::Graphical);
 
-    auto webdriver_socket_path = ByteString::formatted("{}/webdriver", TRY(Core::StandardPaths::runtime_directory()));
+    auto webdriver_socket_path = ByteString::formatted("{}/webdriver", Core::StandardPaths::tempfile_directory());
```
gmta, line 103: It looks like `Core::StandardPaths::runtime_directory()` is doing the wrong thing for macOS, looking at how we implemented it for other platforms. So instead of changing this to `tempfile_directory()`, could you modify `runtime_directory()` instead? You could add a comment there explaining why we wouldn't want something like the username in the path.

AronHetLam: I had some time again and took a look at it, but changing the `runtime_directory()` to eg `/Library/Application Support` has issues. It seems that macOS restricts the creation of unix sockets in `/Library/Application Support`, which currently just makes my build crash after launch, as it cannot create ipc sockets in `LibWebView/Process.cpp`. The correct solution seems to use `mach` ipc ports on macOS for this. I don't have any experience with those, but I can see they're used other places in the codebase.

AronHetLam: Using Mach ports within Ladybirds own processes seems like the right solution, but for the WebDriver I'm uncertain if it would go against spec. My guess is it would, so maybe using 'tempfile_directory()' is the right solution here to avoid restrictions.

Services/WebDriver/Session.cpp
```diff
@@ -254,7 +254,7 @@ ErrorOr<void> Session::start(LaunchBrowserCallback const& launch_browser_callbac
 {
     auto promise = ServerPromise::construct();
 
-    m_web_content_socket_path = ByteString::formatted("{}/webdriver/session_{}_{}", TRY(Core::StandardPaths::runtime_directory()), getpid(), m_session_id);
+    m_web_content_socket_path = ByteString::formatted("{}/webdriver/session_{}_{}", Core::StandardPaths::tempfile_directory(), getpid(), m_session_id);
```
gmta, line 257: Looking at usage of `Core::StandardsPath::runtime_directory()`, we seem to have the same issue in `Process::paths_for_process()`. Could you update that as well?

AronHetLam: If I can find some time, yes. I'm back from vacation since last I contributed 😊
But for those we should use mach ports on macOS if I remember correctly.
I'm not sure what to do for Linux yet.

PR #5850, opened 2025-08-14
LibWeb: Implement CSS border radius overlapping algorithm

dan-v:
Fixes rendering of elements with large border-radius values by scaling radii proportionally when they exceed element dimensions per CSS spec.

Noticed this on a number of elements on beta.digg.com

> before
Screenshot 2025-08-13 at 8 38 50 PM
Screenshot 2025-08-14 at 12 20 51 PM

> after
Screenshot 2025-08-13 at 9 39 03 PM
Screenshot 2025-08-14 at 12 20 59 PM

### DIFF
--- a/Libraries/LibWeb/Painting/Paintable.cpp
+++ b/Libraries/LibWeb/Painting/Paintable.cpp
@@ -212,52 +212,54 @@ Gfx::AffineTransform Paintable::compute_combined_css_transform() const
 
 Painting::BorderRadiiData normalize_border_radii_data(Layout::Node const& node, CSSPixelRect const& rect, CSS::BorderRadiusData const& top_left_radius, CSS::BorderRadiusData const& top_right_radius, CSS::BorderRadiusData const& bottom_right_radius, CSS::BorderRadiusData const& bottom_left_radius)
 {
-    Painting::BorderRadiusData bottom_left_radius_px {};
-    Painting::BorderRadiusData bottom_right_radius_px {};
-    Painting::BorderRadiusData top_left_radius_px {};
-    Painting::BorderRadiusData top_right_radius_px {};
-
-    bottom_left_radius_px.horizontal_radius = bottom_left_radius.horizontal_radius.to_px(node, rect.width());
-    bottom_right_radius_px.horizontal_radius = bottom_right_radius.horizontal_radius.to_px(node, rect.width());
-    top_left_radius_px.horizontal_radius = top_left_radius.horizontal_radius.to_px(node, rect.width());
-    top_right_radius_px.horizontal_radius = top_right_radius.horizontal_radius.to_px(node, rect.width());
-
-    bottom_left_radius_px.vertical_radius = bottom_left_radius.vertical_radius.to_px(node, rect.height());
-    bottom_right_radius_px.vertical_radius = bottom_right_radius.vertical_radius.to_px(node, rect.height());
-    top_left_radius_px.vertical_radius = top_left_radius.vertical_radius.to_px(node, rect.height());
-    top_right_radius_px.vertical_radius = top_right_radius.vertical_radius.to_px(node, rect.height());
+    Painting::BorderRadiiData radii_px {
+        .top_left = {
+            top_left_radius.horizontal_radius.to_px(node, rect.width()),
+            top_left_radius.vertical_radius.to_px(node, rect.height()) },
+        .top_right = { top_right_radius.horizontal_radius.to_px(node, rect.width()), top_right_radius.vertical_radius.to_px(node, rect.height()) },
+        .bottom_right = { bottom_right_radius.horizontal_radius.to_px(node, rect.width()), bottom_right_radius.vertical_radius.to_px(node, rect.height()) },
+        .bottom_left = { bottom_left_radius.horizontal_radius.to_px(node, rect.width()), bottom_left_radius.vertical_radius.to_px(node, rect.height()) }
+    };
 
     // Scale overlapping curves according to https://www.w3.org/TR/css-backgrounds-3/#corner-overlap
     // Let f = min(Li/Si), where i ∈ {top, right, bottom, left},
     // Si is the sum of the two corresponding radii of the corners on side i,
     // and Ltop = Lbottom = the width of the box, and Lleft = Lright = the height of the box.
-    auto l_top = rect.width();
-    auto l_bottom = l_top;
-    auto l_left = rect.height();
-    auto l_right = l_left;
-    auto s_top = (top_left_radius_px.horizontal_radius + top_right_radius_px.horizontal_radius);
-    auto s_right = (top_right_radius_px.vertical_radius + bottom_right_radius_px.vertical_radius);
-    auto s_bottom = (bottom_left_radius_px.horizontal_radius + bottom_right_radius_px.horizontal_radius);
-    auto s_left = (top_left_radius_px.vertical_radius + bottom_left_radius_px.vertical_radius);
-    CSSPixelFraction f = 1;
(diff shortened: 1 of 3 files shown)


Tests/LibWeb/TestBorderRadiusNormalization.cpp
```diff
@@ -0,0 +1,94 @@
...
+        f = min(f, height / right_sum);
+    return f;
+}
+
+TEST_CASE(border_radius_normalization_no_overlap)
+{
+    Painting::BorderRadiiData radii;
+    radii.top_left = radii.top_right = radii.bottom_right = radii.bottom_left = { CSSPixels(10), CSSPixels(10) };
+
+    auto f = calculate_scaling_factor(radii, CSSPixels(100), CSSPixels(100));
```
konradekk: you are testing the test; it’d be better to import the extracted snippet actually used in the application instead.

dan-v: moved this logic to `normalize_overlapping_radii` so it could be used directly in the tests

InvalidUsernameException:
Does this fix the same issue as #5513?

The new `BorderRadiiData::normalize_overlapping_radii()` looks a lot like the existing `Paintable::normalize_border_radii_data()`. The other PR changed the latter to execute twice. I guess, in this PR each of them is executing once, having the same effect overall?

dan-v:
> Does this fix the same issue as #5513?
> 
> The new `BorderRadiiData::normalize_overlapping_radii()` looks a lot like the existing `Paintable::normalize_border_radii_data()`. The other PR changed the latter to execute twice. I guess, in this PR each of them is executing once, having the same effect overall?

This does look like the same issue as #5513. I think some form of dual pass is actually necessary here to deal with precision issues, so I updated with that approach instead with some minor optimization. Also made a bit more comprehensive screenshot test. Feel free to take original PR though since this just adds slight tweaks to it now.

konradekk:
consider merging the effort and solutions marking the co-author in the commit and the code? (just an idea and suggestion, ignore it freely!)

dan-v:
@FadedBronze i added you as co-author

Tests/LibWeb/Screenshot/expected/border-radius-normalization-ref.html
```diff
@@ -0,0 +1,54 @@
...
+</head>
+<body>
+  <h2>Border Radius Normalization Test</h2>
+
+  <div class="test-case">
+    <div class="label">Float max precision test:</div>
+    <div class="float-max-precision"></div>
+  </div>
+</body>
+</html>
```
gmta: You've made a ref test, not a screenshot test. Which is better! So you can delete the PNG file and move this to `Tests/LibWeb/Ref` instead :-)

dan-v: Oops - thanks for the help! Moved this to `Tests/LibWeb/Ref` now and ran LibWeb tests successfully - should be good to go hopefully.

gmta (approved):

gmta:
@dan-v Code linter is failing in CI

dan-v:
> @dan-v Code linter is failing in CI

Sorry about that 🤦. Just pushed a fix for lint issue.

Tests/LibWeb/Ref/input/border-radius-normalization.html
```diff
@@ -0,0 +1,56 @@
+<!DOCTYPE html>
+<html>
+<link rel="match" href="../expected/border-radius-normalization-ref.html" />
+<meta name="fuzzy" content="maxDifference=0-200;totalPixels=0-900">
```
gmta: Sorry, didn't notice this before: why is this fuzzy error so large? Is that the actual error between Linux and macOS?

dan-v: I believe the large fuzzy range is needed here (or at least doesn't pass without it) because this test exposes a fixed-point arithmetic precision issue in border radius rendering. From what I can tell, when border radius values saturate to `CSSPixels::max` (33,554,431), the scaling factor calculation produces extremely small values. For example, with a 200×100px box:

* Sum of horizontal radii: 33,554,431 + 33,554,431 = 67,108,862
* Scaling factor: 200 / 67,108,862 ≈ 3×10⁻⁶

This scaling factor would be roughly 5,000 times smaller than our fixed-point resolution of 0.015625 (CSSPixels uses 6 fractional bits), which causes precision loss that affects the entire element's rendering.

I tested with smaller extreme values instead of float max and still needed similar fuzzy ranges, so it seems like any value triggering this precision issue requires substantial fuzziness. There likely could be platform differences as well here, but I didn't verify this.

gmta: The denominator never gets to 67,108,862, since it is saturated to 33,554,431. In your example, the corners effectively get set to 100 (shortest side) in the first iteration. Only in the second iteration do they get divided in half.

But since they're fixed precision, I'd not expect any pixel differences between platforms because of that.

dan-v: Ah yes you're correct! Dug a little deeper, of course it ended up being something way simpler. In `border-radius-normalization-ref.html` I had incorrectly set border-radius to 20px as i thought it should be half the height (40px), but really should be (height + borders)/2 = (40+2)/2 = 21px. With that change I'm able to set fuzziness to 0.

Tests/LibWeb/Ref/input/border-radius-normalization.html
```diff
@@ -0,0 +1,56 @@
+<!DOCTYPE html>
+<html>
+<link rel="match" href="../expected/border-radius-normalization-ref.html" />
+<meta name="fuzzy" content="maxDifference=0;totalPixels=0">
```
gmta: You can leave out this tag if there's no fuzzy matching required anymore :-)

Libraries/LibWeb/Painting/Paintable.cpp
```diff
@@ -212,52 +212,54 @@ Gfx::AffineTransform Paintable::compute_combined_css_transform() const
...
-        top_right_radius_px.vertical_radius *= f;
-        bottom_right_radius_px.horizontal_radius *= f;
-        bottom_right_radius_px.vertical_radius *= f;
-        bottom_left_radius_px.horizontal_radius *= f;
-        bottom_left_radius_px.vertical_radius *= f;
+    //
+    // NOTE: We iterate twice as a form of iterative refinement. A single scaling pass using
+    // fixed-point arithmetic can result in small rounding errors, causing the scaled radii to
+    // still slightly overflow the box dimensions. A second pass corrects this remaining error.
+    for (int iteration = 0; iteration < 2; ++iteration) {
```
konradekk, line 232: from the theoretical perspective it would be probably better to loop on `f` with some invariants (as you already hint that `break`ing on `f >= 1` is a good stopping condition); possibly with a `do-while`. but this might get cumbersome, so treat that only as an observation from my side (or just continue with maintainer guidance!).

Tests/LibWeb/Ref/input/border-radius-normalization.html
dan-v: Removed!

Libraries/LibWeb/Painting/Paintable.cpp
gmta: That would take a good deal of code feng shuiing without a significant tangible benefit, imo. You need to calculate `f` before entering the loop, or add the condition add the end but skip part of the application conditionally anyway. This approach with 2 iterations seems fine for now!

PR #5851, opened 2025-08-14
LibWeb: Correct typos for user validity of form elements on submission

CountBleck:
There were some `set_user_validity(false)`s when they should've been `set_user_validity(true)`s.


gmta:
Thank you for fixing this!

PR #5852, opened 2025-08-14
LibWeb: Improvements to transitions

Calme1709:
This PR fixes a couple issues with CSS transitions:
 - Properly initialize the KeyframeEffect contained within a CSSTransition
 - Only associate newly constructed CSSTransitions with the relevant element once
 - Inherit animated CSS property values separately from non-animated values

The last one does come with a bunch of (~250) new WPT test failures from the various 'events' tests in
[css/css-transitions/properties-value-inherit-001.html](https://wpt.live/css/css-transitions/properties-value-inherit-001.html), however these also fail in other browsers (Chrome, Edge and Firefox) and the behavior that causes these failures is specifically mentioned in the spec (see note [here](drafts.csswg.org/css-transitions-1/#after-change-style)).


Poldraunic:
Link to CSS transitions spec is borked.

Calme1709:
Thanks - fixed

Calme1709:
Rebased to resolve conflicts

PR #5853, opened 2025-08-14
LibIDL: Add parsing support for `[Exposed=(PaintWorklet,LayoutWorklet)]`

AtkinsSJ:
Having to FIXME these bothered me, and because just about every CSS typed-om type is exposed to these worklets, let's make them parse to avoid adding a whole load more FIXMEs.

Of course, we don't have these worklets yet so nothing is actually exposed on them, but once we do add them, we won't have to update a bunch of IDL files.

### DIFF
--- a/Libraries/LibWeb/CSS/CSSStyleValue.idl
+++ b/Libraries/LibWeb/CSS/CSSStyleValue.idl
@@ -1,6 +1,5 @@
 // https://drafts.css-houdini.org/css-typed-om-1/#cssstylevalue
-// FIXME: [Exposed=(Window, Worker, PaintWorklet, LayoutWorklet)]
-[Exposed=(Window, Worker)]
+[Exposed=(Window, Worker, PaintWorklet, LayoutWorklet)]
 interface CSSStyleValue {
     stringifier;
     [FIXME, Exposed=Window] static CSSStyleValue parse(USVString property, USVString cssText);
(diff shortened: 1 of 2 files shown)


Libraries/LibIDL/ExposedTo.h
```diff
@@ -19,7 +19,10 @@ enum class ExposedTo {
     Window = 0x10,
     ShadowRealm = 0x20,
     Worklet = 0x40,
-    AllWorkers = DedicatedWorker | SharedWorker | ServiceWorker | AudioWorklet, // FIXME: Is "AudioWorklet" a Worker? We'll assume it is for now (here, and line below)
+    PaintWorklet = 0x80,
+    LayoutWorklet = 0x100,
+    // FIXME: Are the FooWorklets Workers? We'll assume they are for now.
+    AllWorkers = DedicatedWorker | SharedWorker | ServiceWorker | AudioWorklet | PaintWorklet | LayoutWorklet,
```
gmta: I believe `Exposed=Worklet` is a thing that exists, which might give rise to `AllWorklets` instead?

AtkinsSJ: Yeah it doesn't seem right the way it is here, but I didn't want to get into a rabbit hole so I just left it as-is. :sweat_smile:

I can try and figure it out if you want though.

gmta: Do things break if you leave `AllWorkers` at its original definition? 🤔

AtkinsSJ: Nah, I can change it back. (And I've just done so!) I just added the new worklets in because AudioWorklet was already there.

gmta (approved):

PR #5856, opened 2025-08-14
LibWeb: Implement `<feImage>` SVG filter

tcl3:
Because the size of the image depends on the size of the box of the referenced element, I had to move the place where the `Gfx::Filter` is constructed from a `CSS::Filter`, this now occurs when layout metrics are available.


Libraries/LibWeb/SVG/SVGFilterElement.cpp
```diff
@@ -129,6 +133,39 @@ Optional<Gfx::Filter> SVGFilterElement::gfx_filter()
...
+            auto height = image_primitive->height()->base_val()->value();
+            if (height == 0)
+                height = bitmap->height();
+
+            Gfx::IntRect src_rect {
+                image_primitive->x()->base_val()->value(),
+                image_primitive->y()->base_val()->value(),
+                width,
+                height
+            };
```
gmta: Maybe all this `src_rect` logic should go into `SVGFEImageElement`? That would keep the filter logic here a bit simpler.

tcl3: That's a more natural place for this logic to be. Moved as suggested.

AtkinsSJ (approved):

PR #5860, opened 2025-08-14
LibWeb: Don't process mouseleave events while primary button is pressed

### LINKED ISSUE #5844: Crash when scrollbar is dragged outside window
### Summary

When I drag the scrollbar outside of browser window, the tab crashes.

### Operating system

macOS

### Steps to reproduce

1. Open any page with enough length for scrollbar to appear
2. Drag the scrollbar thumb outside of browser window
3. Tab crashes.
4. Sometimes whole browser crashes.

### Expected behavior

No crashing, scroll follows pointer.

### Actual behavior

Browser crashes in both AppKit and Qt build.

[Crash Report (download)](https://github.com/user-attachments/files/21757728/Crash.Report.txt)

N/A

```html
N/A
```

```shell
% ./Meta/ladybird.py run
ninja: Entering directory `/Users/seohy/ladybird/Build/release'
ninja: no work to do.
Ladybird PID file '/Users/seohy/Library/Application Support/Ladybird.pid' exists with PID 98808, but process cannot be found
VERIFICATION FAILED: max >= min at /Users/seohy/ladybird/AK/StdLibExtras.h:73
0   liblagom-ak.0.0.0.dylib             0x00000001055e5698 dump_backtrace + 72
1   liblagom-ak.0.0.0.dylib             0x00000001055e5950 ak_trap + 12
2   liblagom-ak.0.0.0.dylib             0x00000001055e59fc ak_assertion_failed + 0
3   liblagom-web.0.0.0.dylib            0x000000010987ac78 Web::CSS::CalculatedStyleValue::resolve_value(Web::CSS::CalculationResolutionContext const&) const (.cold.1) + 0
4   liblagom-web.0.0.0.dylib            0x0000000108d079f0 Web::Painting::PaintableBox::scroll_to_mouse_position(Gfx::Point<Web::CSSPixels>) + 924
5   liblagom-web.0.0.0.dylib            0x0000000108d074ac Web::Painting::PaintableBox::handle_mousemove(AK::Badge<Web::EventHandler>, Gfx::Point<Web::CSSPixels>, unsigned int, unsigned int) + 132
6   liblagom-web.0.0.0.dylib            0x0000000108cd5c0c Web::EventHandler::handle_mousemove(Gfx::Point<Web::CSSPixels>, Gfx::Point<Web::CSSPixels>, unsigned int, unsigned int) + 324
7   liblagom-web.0.0.0.dylib            0x0000000108ac3654 Web::HTML::EventLoop::process_input_events() const + 404
8   liblagom-web.0.0.0.dylib            0x0000000108ac3bfc Web::HTML::EventLoop::update_the_rendering() + 80
9   liblagom-web.0.0.0.dylib            0x0000000108ac58d4 Web::HTML::Task::execute() + 92
10  liblagom-web.0.0.0.dylib            0x0000000108ac2f2c Web::HTML::EventLoop::process() + 84
11  liblagom-web.0.0.0.dylib            0x0000000108d20370 AK::Function<void ()>::CallableWrapper<Web::Platform::TimerSerenity::TimerSerenity()::$_0>::call() + 100
12  liblagom-core.0.0.0.dylib           0x0000000104ec59d8 Core::Timer::timer_event(Core::TimerEvent&) + 208
13  liblagom-core.0.0.0.dylib           0x0000000104ec4998 Core::ThreadEventQueue::process() + 424
14  liblagom-core.0.0.0.dylib           0x0000000104eca07c Core::EventLoopImplementationUnix::exec() + 44
15  liblagom-core.0.0.0.dylib           0x0000000104eb6eac Core::EventLoop::exec() + 68
16  WebContent                          0x000000010497c670 ladybird_main(Main::Arguments) + 4224
17  WebContent                          0x0000000104a1ff0c main + 184
18  dyld                                0x000000018ef72b98 start + 6076
UNEXPECTED ERROR: Trying to post_message during IPC shutdown at /Users/seohy/ladybird/Build/release/Lagom/Services/WebContent/WebContentServerEndpoint.h:4878
0   liblagom-ak.0.0.0.dylib             0x0000000101511698 dump_backtrace + 72
1   liblagom-ak.0.0.0.dylib             0x0000000101511950 ak_trap + 12
2   liblagom-webview.0.0.0.dylib        0x0000000101183d20 WebContentServerProxy<WebContentClientEndpoint, WebContentServerEndpoint>::async_mouse_event(unsigned long long, Web::MouseEvent const&) + 788
3   liblagom-webview.0.0.0.dylib        0x0000000101173b30 WebView::ViewImplementation::enqueue_input_event(AK::Variant<Web::KeyEvent, Web::MouseEvent, Web::DragEvent>) + 224
4   Ladybird                            0x0000000100c69454 Ladybird::WebViewBridge::enqueue_input_event(Web::MouseEvent) + 168
5   Ladybird                            0x0000000100c5d6ec -[LadybirdWebView mouseDragged:] + 128
6   AppKit                              0x0000000193e59abc -[NSWindow _handleMouseDraggedEvent:] + 52
7   AppKit                              0x0000000193447bc4 -[NSWindow(NSEventRouting) _reallySendEvent:isDelayedEvent:] + 348
8   AppKit                              0x00000001934478a0 -[NSWindow(NSEventRouting) sendEvent:] + 288
9   AppKit                              0x0000000193cbf6c0 -[NSApplication(NSEventRouting) sendEvent:] + 1504
10  Ladybird                            0x0000000100c4df2c -[Application sendEvent:] + 104
11  AppKit                              0x00000001938be42c -[NSApplication _handleEvent:] + 60
12  AppKit                              0x0000000193314c0c -[NSApplication run] + 520
13  Ladybird                            0x0000000100c76f2c WebView::EventLoopImplementationMacOS::exec() + 32
14  liblagom-core.0.0.0.dylib           0x000000010134eeac Core::EventLoop::exec() + 68
15  liblagom-webview.0.0.0.dylib        0x00000001011324f0 WebView::Application::execute() + 384
16  Ladybird                            0x0000000100c4c3f8 ladybird_main(Main::Arguments) + 664
17  Ladybird                            0x0000000100c74d34 main + 184
18  dyld                                0x000000018ef72b98 start + 6076
```

_No response_

_No response_

### Contribute a patch?

hopafoot:
Prevents scrolling from being attempted if the thumb size exceeds the gutter size. This should never happen, however, if you drag the thumb while moving the mouse outside the window the gutter is no longer rendered. As such, the gutter has a size of 0 while the, still rendered, thumb has a positive size. 

Fixes #5844

### DIFF
--- a/Libraries/LibWeb/Page/EventHandler.cpp
+++ b/Libraries/LibWeb/Page/EventHandler.cpp
@@ -586,6 +586,7 @@ after_node_use:
     if (button == UIEvents::MouseButton::Primary) {
         m_in_mouse_selection = false;
         m_mouse_selection_target = nullptr;
+        m_primary_button_held = false;
     }
     return handled_event;
 }
@@ -612,6 +613,8 @@ EventResult EventHandler::handle_mousedown(CSSPixelPoint viewport_position, CSSP
         document->set_hovered_node(node);
     };
 
+    auto primary_button_event = UIEvents::MouseButton::Primary == button;
+
     {
         GC::Ptr<Painting::Paintable> paintable;
         if (auto result = target_for_mouse_position(viewport_position); result.has_value())
@@ -634,8 +637,12 @@ EventResult EventHandler::handle_mousedown(CSSPixelPoint viewport_position, CSSP
             return EventResult::Dropped;
 
         if (is<HTML::HTMLIFrameElement>(*node)) {
-            if (auto content_navigable = static_cast<HTML::HTMLIFrameElement&>(*node).content_navigable())
-                return content_navigable->event_handler().handle_mousedown(viewport_position.translated(compute_mouse_event_offset({}, *paintable)), screen_position, button, buttons, modifiers);
+            if (auto content_navigable = static_cast<HTML::HTMLIFrameElement&>(*node).content_navigable()) {
+                auto result = content_navigable->event_handler().handle_mousedown(viewport_position.translated(compute_mouse_event_offset({}, *paintable)), screen_position, button, buttons, modifiers);
+                if ((EventResult::Accepted == result || EventResult::Handled == result) && primary_button_event)
+                    m_primary_button_held = true;
+                return result;
+            }
             return EventResult::Dropped;
         }
 
@@ -658,8 +665,11 @@ EventResult EventHandler::handle_mousedown(CSSPixelPoint viewport_position, CSSP
     }
 
     // NOTE: Dispatching an event may have disturbed the world.
-    if (!paint_root() || paint_root() != node->document().paintable_box())
+    if (!paint_root() || paint_root() != node->document().paintable_box()) {
+        if (primary_button_event)
+            m_primary_button_held = true;
         return EventResult::Accepted;
+    }
 
     if (button == UIEvents::MouseButton::Primary) {
         if (auto result = paint_root()->hit_test(viewport_position, Painting::HitTestType::TextCursor); result.has_value()) {
@@ -715,6 +725,9 @@ EventResult EventHandler::handle_mousedown(CSSPixelPoint viewport_position, CSSP
             }
         }
     }
+
+    if (primary_button_event)
+        m_primary_button_held = true;
 
     return EventResult::Handled;
 }
@@ -878,6 +891,8 @@ EventResult EventHandler::handle_mouseleave()
     if (!m_navigable->active_document())
         return EventResult::Dropped;
     if (!m_navigable->active_document()->is_fully_active())
+        return EventResult::Dropped;
+    if (m_primary_button_held)
         return EventResult::Dropped;
 
     m_navigable->active_document()->update_layout(DOM::UpdateLayoutReason::EventHandlerHandleMouseMove);
--- a/Libraries/LibWeb/Page/EventHandler.h
+++ b/Libraries/LibWeb/Page/EventHandler.h
@@ -70,6 +70,7 @@ private:
     GC::Ref<HTML::Navigable> m_navigable;
 
     bool m_in_mouse_selection { false };
+    bool m_primary_button_held { false };
     InputEventsTarget* m_mouse_selection_target { nullptr };
 
     GC::Ptr<Painting::Paintable> m_mouse_event_tracking_paintable;


trflynn89:
Hmm this feels a bit like a band-aid. I *think* the actual issue is that we should not be handling `mouseleave` events in Libraries/LibWeb/Page/EventHandler.cpp if the mouse was pressed down on some element when we left the browser window.

I'm not sure what's changed, but that's actually how this used to behave when I first wrote the mouseleave handler a couple weeks ago, as seen in the screen recording here: https://github.com/LadybirdBrowser/ladybird/pull/5615

In any case, I think the solution will be to track whether we have the mouse pressed down on something in the mousedown/mouseup handlers (if we don't already), and to ignore mouseleave events while that condition is true.

hopafoot:
@trflynn89, you're right. It felt wrong to do it that way. Thanks for pointing me in the right direction.
Could you please check the force-pushed changes, particularly that the result should be "Dropped" in this case?

Libraries/LibWeb/Page/EventHandler.cpp
```diff
@@ -879,6 +892,8 @@ EventResult EventHandler::handle_mouseleave()
         return EventResult::Dropped;
     if (!m_navigable->active_document()->is_fully_active())
         return EventResult::Dropped;
+    if (m_primary_button_held)
+        return EventResult::Dropped;
```
gmta, line 896: Does this mean we no longer fire `mouseleave` events when holding the primary button, even if you let go of the primary button afterwards?

hopafoot: In mouseup, ~line 589, we set the value to false. m_primary_button_held should only be true while the button is actively being pressed down.

gmta: I understand, but shouldn't we fire `mouseleave` as soon as the button is released?

konradekk: it feels kind of wrong not to trigger `mouseleave` when pointer does leave an element: not sure what browsers do and spec says, but my gut feeling is that `mouseenter` and `mouseleave` should just happen with or without button pressed (hovering in and out, e.g. a link). excuses for the noise!

rcorsi:
I've created a different solution to this issue #5882

rcorsi:
Issue was introduced at commit 67cc02ab59375ae7520098056dd58ba20b4d44f8

rcorsi:
@trflynn89 could it be that MacOSX is working correctly by sending mouseleave event when leaving the window (and holding button down) and another mouseleave event when releasing the button?  And Linux is not working correctly... by not sending the mouseleave while holding the button.

When I use the `xev` utility in Linux (Ubuntu 25.04) I see a `LeaveNotify` event when leaving the window (and hold button down) and when button is released I get `ButtonRelease` event immediately followed by `LeaveNotify` event.

rcorsi:
I was curious if using an X desktop would result in a different outcome (vs Wayland) for the mouseleave event, so I installed Xubuntu to play around in.  So far same results on Xubuntu, the mouseleave is not called when button pressed.

I also did some more checking with the `xev` and `xwininfo`  commands using Xubuntu, and `LeaveNotify` is acting same as mouseleave.  So seems that this is a consistent Qt+Linux behaviour.

It would seem that MacOSX with Cocoa+AppKit and Cocoa+Qt have their own behaviour.

Libraries/LibWeb/Page/EventHandler.cpp
rcorsi: > I understand, but shouldn't we fire mouseleave as soon as the button is released?

On MacOSX, when pressing mouse button while scrolling there is a mouseleave at the moment the mouse leaves the main window and another mouseleave is sent when the mouse button is released.  This patch is attempting to drop the first type.

Issue #226, opened 2024-06-20
LibWeb: Cloudflare Turnstile differences between us and other engines


Lubrsi:
NOTE: This is a living document, it does not contain every difference currently and will be added to.

Neat site for checking the connection side of things (not sure of trustworthiness, take it with a grain of salt): https://cloudflare.manfredi.io/en/tools/connection/

- [ ] navigator.plugins is empty (presumably because we report no PDF support, see `navigator.pdfViewerEnabled`)
- [ ] navigator.mimeTypes is empty (presumably because we report no PDF support, see `navigator.pdfViewerEnabled`)
- [ ] Missing `first-input` performance entry type
- [ ] Missing `paint` performance entry type (including `first-paint` and `first-contentful-paint`)
- [ ] Missing `navigation` performance entry type
- [ ] Missing `resource` performance entry type (definitely needs implementing, disabling `dom.enable_resource_timing` in Firefox makes it never pass)
- [ ] Missing `visibility-state` performance entry type
- [ ] Missing `long-animation-frame` performance entry type
- [ ] Missing WEBGL_debug_renderer_info extension for unmasked vendor/renderer info
- [x] Blob URLs for workers don't work, reporting `DedicatedWorkerHost: Unable to fetch script blob:https://localhost:8080/c838e9e6-b238-4210-a349-9c909821d0a5 because script was null`
- [x] navigator.hardwareConcurrency always reports 1
- [ ] window.outerWidth and window.outerHeight report 0 (used in zoom calculation if it thinks we're chromium: Math.round((window.outerWidth / window.innerWidth) * 100) / 100)
- [ ] Thinks we're Firefox due to missing window.chrome and window.webkitRequestAnimationFrame, which makes it resort to probing `min--moz-device-pixel-ratio` via `window.matchMedia` for zoom calculation, and reporting "ff" in the same test
- [x] typeof `document.all` is not `"undefined"`
- [ ] Missing screen.availTop
- [ ] Missing screen.availLeft
- [x] Missing screen.orientation (angle, onchange, lock, unlock, type, inherits EventTarget)
- [x] Missing screen.onchange
- [x] Missing screen.isExtended
- [x] Screen does not inherit EventTarget
- [ ] Missing OfflineAudioContext (random fingerprint challenge, see https://github.com/LadybirdBrowser/ladybird/issues/224)
- [ ] This spec issue for WWW-Authenticate needs resolving:
[...]

PR #5861, opened 2025-08-14
LibJS: Implement console.dirxml

tete17:
This function actually gets tested in WPT.

https://wpt.fyi/results/console/idlharness.any.html


Libraries/LibJS/Console.cpp
```diff
@@ -398,6 +398,32 @@ ThrowCompletionOr<Value> Console::dir()
...
+{
+    auto& vm = realm().vm();
+
+    // 1. Let finalList be a new list, initially empty.
+    GC::RootVector<Value> final_list(vm.heap());
+
+    // 2. For each item of data:
+    for (auto const& item : vm_arguments()) {
+        // 1. Let converted be a DOM tree representation of item if possible;
+        //    otherwise let converted be item with optimally useful formatting applied.
```
trflynn89: Please wrap comments at ~120 characters instead

Libraries/LibJS/Console.cpp
```diff
@@ -398,6 +398,32 @@ ThrowCompletionOr<Value> Console::dir()
...
+        // TODO: "Optimally-useful formatting"
```
trflynn89: We prefer to use `FIXME` over `TODO`

Libraries/LibJS/Console.cpp
```diff
@@ -398,6 +398,32 @@ ThrowCompletionOr<Value> Console::dir()
...
+// 1.1.11 dirxml(...data) https://console.spec.whatwg.org/#dirxml
+ThrowCompletionOr<Value> Console::dirxml()
+{
+    auto& vm = realm().vm();
+
+    // 1. Let finalList be a new list, initially empty.
+    GC::RootVector<Value> final_list(vm.heap());
+
+    // 2. For each item of data:
+    for (auto const& item : vm_arguments()) {
```
trflynn89: This is creating another `RootVector` holding all of the arguments. Seems pretty wasteful. We can just do:
```c++
for (size_t i = 0; i < vm.argument_count(); ++i) {
    auto item = vm.arguments(i);
```

Libraries/LibJS/Console.cpp
```diff
@@ -398,6 +398,32 @@ ThrowCompletionOr<Value> Console::dir()
...
+        //    otherwise let converted be item with optimally useful formatting applied.
+        // TODO: "Optimally-useful formatting"
+
+        // 2. Append converted to finalList.
+        final_list.append(item);
+    }
+
+    // 3. Perform Logger("dirxml", finalList).
+    if (m_client) {
+        return m_client->logger(LogLevel::Warn, final_list);
```
trflynn89: The spec step states we should use "dirxml", not "warn".

Lubrsi:
This is also tested by Cloudflare Turnstile, as per https://github.com/LadybirdBrowser/ladybird/issues/226 :^)

PR #5862, opened 2025-08-15
LibWeb/CSS: Start reifying some CSSStyleValues!

AtkinsSJ:
This lets pages actually read properties from a typed OM StylePropertyMap.

For now this is just CSSStyleValue itself, so all that JS can do is serialize them as strings. I've implemented a couple of the types on a separate branch but I'm trying to PR this a bit more gradually.


Libraries/LibWeb/CSS/CSSDescriptors.cpp
```diff
@@ -281,6 +281,14 @@ RefPtr<StyleValue const> CSSDescriptors::descriptor_or_initial_value(DescriptorI
     return descriptor_initial_value(m_at_rule_id, descriptor_id);
```
Calme1709, line 281: Should commit refer to `has_property()` rather than `has()`?

AtkinsSJ: Yes it should! Done.

Issue #5319, opened 2025-07-06
play.pvzge.com shows a blank screen


jrapf32:
### Summary

When I load play.pvzge.com in Ladybird, it shows a blank screen.

### Operating system

macOS

### Steps to reproduce

1. Load play.pvzge.com
2. Profit!

### Expected behavior

It gets to the loading flower.

### Actual behavior

It shows a black screen

### URL for a reduced test case

play.pvzge.com

### HTML/SVG/etc. source for a reduced test case

```html
N/A
```

### Log output and (if possible) backtrace

```shell
2025-07-05 18:33:54.074 Ladybird[89577:6917961] +[IMKClient subclass]: chose IMKClient_Modern
2025-07-05 18:33:54.137 Ladybird[89577:6917961] +[IMKInputSession subclass]: chose IMKInputSession_Modern
626062.738 WebContent(89581): ResourceLoader: Filtered request to: "https://static.cloudflareinsights.com/beacon.min.js/vcd15cbe7772f49c399c6a5babf22c1241717689176015"
626062.778 WebContent(89581): ResourceLoader: Filtered request to: "https://fundingchoicesmessages.google.com/i/pub-7637695321442015?ers=1"
626062.778 WebContent(89581): ResourceLoader: Filtered request to: "https://pagead2.googlesyndication.com/pagead/js/adsbygoogle.js?client=ca-pub-7637695321442015"
626062.778 WebContent(89581): ResourceLoader: Filtered request to: "https://www.googletagmanager.com/gtag/js?id=G-FNC28KV0ZR"
626062.778 WebContent(89581): HTMLScriptElement: Refusing to run script because the element's result is null.
626062.791 WebContent(89581): ResourceLoader: Filtered request to: "https://www.googletagmanager.com/gtag/js?id=G-FNC28KV0ZR"
626062.792 WebContent(89581): ResourceLoader: Filtered request to: "https://pagead2.googlesyndication.com/pagead/js/adsbygoogle.js?client=ca-pub-7637695321442015"
626062.792 WebContent(89581): ResourceLoader: Filtered request to: "https://fundingchoicesmessages.google.com/i/pub-7637695321442015?ers=1"
626062.792 WebContent(89581): ResourceLoader: Filtered request to: "https://fundingchoicesmessages.google.com/b/pub-7637695321442015"
626062.795 WebContent(89581): HTMLScriptElement: Refusing to run script because the element's result is null.
[...]

Issue #5378, opened 2025-07-09
Turbowarp fails to load


jrapf32:
### Summary

When I load a Turbowarp project i,t fails to compile a shader

### Operating system

Linux

### Steps to reproduce

1. Load https://turbowarp.org/9738004483
2. Profit
3.

### Expected behavior

It shows

### Actual behavior

Shows black screen

### URL for a reduced test case

https://turbowarp.org/9738004483

### HTML/SVG/etc. source for a reduced test case

```html
N/A
```

### Log output and (if possible) backtrace

```shell
950476.813 WebContent(66587): (js error) "gui" "error" "sw error" [InternalError] TODO(Run Service Worker is not implemented in LibJS)
950483.080 WebContent(66587): FIXME: Unimplemented IDL interface: 'WebGL2RenderingContextPrototype.drawingBufferFormat'
950483.080 WebContent(66587): FIXME: Unimplemented IDL interface: 'WebGL2RenderingContextPrototype.drawingBufferColorSpace'
950483.080 WebContent(66587): FIXME: Unimplemented IDL interface: 'WebGL2RenderingContextPrototype.unpackColorSpace'
950483.080 WebContent(66587): FIXME: Unimplemented IDL interface: 'WebGL2RenderingContextPrototype.getBufferSubData'
950483.080 WebContent(66587): FIXME: Unimplemented IDL interface: 'WebGL2RenderingContextPrototype.framebufferTextureLayer'
950483.080 WebContent(66587): FIXME: Unimplemented IDL interface: 'WebGL2RenderingContextPrototype.invalidateSubFramebuffer'
950483.080 WebContent(66587): FIXME: Unimplemented IDL interface: 'WebGL2RenderingContextPrototype.copyTexSubImage3D'
950483.080 WebContent(66587): FIXME: Unimplemented IDL interface: 'WebGL2RenderingContextPrototype.compressedTexImage3D'
950483.080 WebContent(66587): FIXME: Unimplemented IDL interface: 'WebGL2RenderingContextPrototype.compressedTexSubImage3D'
950483.080 WebContent(66587): FIXME: Unimplemented IDL interface: 'WebGL2RenderingContextPrototype.getFragDataLocation'
950483.080 WebContent(66587): FIXME: Unimplemented IDL interface: 'WebGL2RenderingContextPrototype.uniform1uiv'
[...]

Issue #3366, opened 2025-01-25
Scratch Failed to load big projects.


jrapf32:
### Summary

When I load a big project like [https://scratch.mit.edu/projects/60917032/](This), it either loads infinitely or it says an error Occured. 

### Operating system

macOS

### Steps to reproduce

1. Go to scratch.mit.edu
2. Load a big project
3. It will load infinitely or will say "An Error Occurred".

### Expected behavior

It works

### Actual behavior

It doesn't work.

### URL for a reduced test case

https://scratch.mit.edu/projects/60917032/

### HTML/SVG/etc. source for a reduced test case

```html
N/A
```

### Log output and (if possible) backtrace

```shell
/Volumes/Disk/ladybird/Build/release/bin/Ladybird.app/Contents/MacOS/Ladybird
Ladybird PID file '/Users/john/Library/Application Support/Ladybird.pid' exists with PID 60140, but process cannot be found
2025-01-25 12:29:16.013 Ladybird[65594:2571979] +[IMKClient subclass]: chose IMKClient_Modern
2025-01-25 12:29:16.159 Ladybird[65594:2571979] +[IMKInputSession subclass]: chose IMKInputSession_Modern
227251.623 WebContent(65599): ResourceLoader: Filtered request to: "https://www.googletagmanager.com/gtm.js?id=GTM-NBMRDKQ"
227251.635 WebContent(65599): HTMLScriptElement: Refusing to run script because the element's result is null.
227255.139 WebContent(65599): FIXME: Unimplemented Navigator.maxTouchPoints
227255.312 WebContent(65599): FIXME: Unimplemented IDL interface: 'CanvasRenderingContext2DPrototype.globalCompositeOperation'
227256.321 WebContent(65599): (js log) "Stop!"
227256.321 WebContent(65599): (js log) "This is part of your browser intended for developers. If someone told you to copy-and-paste something here, don't do it! It could allow them to take over your Scratch account, delete all of your projects, or do many other harmful things. If you don't understand what exactly you are doing here, you should close this window without doing anything."
227256.591 WebContent(65599): FIXME: Unimplemented IDL interface: 'WebGLRenderingContextPrototype.drawingBufferFormat'
227256.591 WebContent(65599): FIXME: Unimplemented IDL interface: 'WebGLRenderingContextPrototype.drawingBufferColorSpace'
[...]

Issue #2679, opened 2024-12-01
Telegram Web does not load


shlyakpavel:
### Summary

When I try messaging my friends on Telegram, Ladybird does not load the web version

### Operating system

macOS

### Steps to reproduce

1. Open Ladybird
2. Navigate to https://web.telegram.org
3. Observe

### Expected behavior

Telegram web at least displays login page 

### Actual behavior

There's no login page or anything

### URL for a reduced test case

https://web.telegram.org/

### HTML/SVG/etc. source for a reduced test case

```html
N/A
```

### Log output and (if possible) backtrace

```shell
ninja: Entering directory `/Users/pavel/Develop/ladybird/Build/release'
[0/2] Re-checking globbed directories...
[1016/1017] Linking CXX executable bin/Ladybird.app/Contents/MacOS/Ladybird
INFO | fixing executable '/Users/pavel/Develop/ladybird/Build/release/bin/Ladybird.app/Contents/MacOS/Ladybird'
INFO | fixing plugins
INFO | macdeployqtfix terminated with success
Ladybird PID file '/Users/pavel/Library/Application Support/Ladybird.pid' exists with PID 29975, but process cannot be found
2024-12-01 16:36:08.247 Ladybird[34407:10464466] +[IMKClient subclass]: chose IMKClient_Modern
2024-12-01 16:36:08.361 Ladybird[34407:10464466] +[IMKInputSession subclass]: chose IMKInputSession_Modern
745755.775 WebContent(34410): FIXME: Implement 'Content-Security-Policy' http-equiv state
745755.871 WebContent(34410): ImageDecoderClient: Failed to decode image with ID 1: Decoding failed: Could not find suitable image decoder plugin for data
745756.077 WebContent(34410): No property (from 5 properties) matched Token: Comma
745756.077 WebContent(34410):     animation-iteration-count
745756.077 WebContent(34410):     animation-direction
745756.077 WebContent(34410):     animation-play-state
745756.077 WebContent(34410):     animation-delay
745756.077 WebContent(34410):     animation-fill-mode
745756.077 WebContent(34410): No property (from 5 properties) matched Token: Comma
745756.077 WebContent(34410):     animation-iteration-count
745756.077 WebContent(34410):     animation-direction
745756.077 WebContent(34410):     animation-play-state
745756.077 WebContent(34410):     animation-delay
745756.077 WebContent(34410):     animation-fill-mode
[...]

PR #5864, opened 2025-08-15
LibGfx: enable WebGL for Linux

erik-kz:
This series enables WebGL for Linux. For now it's using ANGLE's OpenGL backend as Vulkan doesn't seem to be enabled in the vcpkg package, but that could be changed easily in the future.

The painting surface will be allocated through Vulkan and imported by both Skia and EGL. The Vulkan image allocation code should also be useful for other things in the future like allocating the main compositing surface so it can be shared directly with the UI process.

To synchronize rendering with presentation, we're just using a glFinish, which is admittedly suboptimal. Eventually we should get a fence from EGL and plumb that through to Skia so that it will wait for it before reading from the surface. However that can be implemented in a future change.

I've tested this with the proprietary NVIDIA driver as well as Mesa on an Intel GPU using the demos on webglsamples.org.

### DIFF
--- a/Libraries/LibWeb/WebGL/OpenGLContext.cpp
+++ b/Libraries/LibWeb/WebGL/OpenGLContext.cpp
@@ -10,25 +10,22 @@
 #include <LibGfx/PaintingSurface.h>
 #include <LibWeb/WebGL/OpenGLContext.h>
 
-#ifdef AK_OS_MACOS
-#    include <EGL/egl.h>
-#    include <EGL/eglext.h>
-#    define EGL_EGLEXT_PROTOTYPES 1
+#include <EGL/egl.h>
+#include <EGL/eglext.h>
+#define EGL_EGLEXT_PROTOTYPES 1
 extern "C" {
-#    include <EGL/eglext_angle.h>
+#include <EGL/eglext_angle.h>
 }
-#    define GL_GLEXT_PROTOTYPES 1
-#    include <GLES2/gl2.h>
-#    include <GLES2/gl2ext.h>
+#define GL_GLEXT_PROTOTYPES 1
+#include <GLES2/gl2.h>
+#include <GLES2/gl2ext.h>
 extern "C" {
-#    include <GLES2/gl2ext_angle.h>
+#include <GLES2/gl2ext_angle.h>
 }
-#endif
 
 namespace Web::WebGL {
 
 struct OpenGLContext::Impl {
-#ifdef AK_OS_MACOS
     EGLDisplay display { nullptr };
     EGLConfig config { nullptr };
     EGLContext context { nullptr };
@@ -37,6 +34,13 @@ struct OpenGLContext::Impl {
     GLuint framebuffer { 0 };
     GLuint color_buffer { 0 };
     GLuint depth_buffer { 0 };
+
+#ifdef USE_VULKAN
+    EGLImage egl_image { EGL_NO_IMAGE };
+    struct {
+        PFNEGLQUERYDMABUFFORMATSEXTPROC query_dma_buf_formats { nullptr };
+        PFNEGLQUERYDMABUFMODIFIERSEXTPROC query_dma_buf_modifiers { nullptr };
+    } ext_procs;
 #endif
 };
 
@@ -49,14 +53,11 @@ OpenGLContext::OpenGLContext(NonnullRefPtr<Gfx::SkiaBackendContext> skia_backend
 
 OpenGLContext::~OpenGLContext()
 {
-#ifdef AK_OS_MACOS
     free_surface_resources();
     eglMakeCurrent(m_impl->display, EGL_NO_SURFACE, EGL_NO_SURFACE, EGL_NO_CONTEXT);
     eglDestroyContext(m_impl->display, m_impl->context);
-#endif
 }
 
-#ifdef AK_OS_MACOS
 void OpenGLContext::free_surface_resources()
 {
     if (m_impl->framebuffer) {
@@ -74,15 +75,22 @@ void OpenGLContext::free_surface_resources()
         m_impl->depth_buffer = 0;
     }
 
+#ifdef USE_VULKAN
+    if (m_impl->egl_image != EGL_NO_IMAGE) {
+        eglDestroyImage(m_impl->display, m_impl->egl_image);
+        m_impl->egl_image = EGL_NO_IMAGE;
+    }
+#endif
+
     if (m_impl->surface != EGL_NO_SURFACE) {
+#ifdef AK_OS_MACOS
         eglReleaseTexImage(m_impl->display, m_impl->surface, EGL_BACK_BUFFER);
+#endif
         eglDestroySurface(m_impl->display, m_impl->surface);
         m_impl->surface = EGL_NO_SURFACE;
     }
 }
-#endif
 
-#ifdef AK_OS_MACOS
 static EGLConfig get_egl_config(EGLDisplay display)
 {
     EGLint const config_attribs[] = {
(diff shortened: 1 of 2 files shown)


zacoons:
Does this fix #5319, #5378, #3366
And make progress on #2679
?

erik-kz:
https://play.pvzge.com seems to be working.

https://turbowarp.org/ does not work. It no longer presents a blank screen, but it now displays an "Oops!" error message probably due to one of the following errors...

> 860082.234 WebContent(263089): (js warn) "vm" "warn" "Error loading vector image: TypeError: null cannot be converted to an object"

or

> 860082.285 WebContent(263089): Unhandled JavaScript exception (in promise): [TypeError] Cannot access property "getBBox" on undefined object "i.children[0]"

https://scratch.mit.edu/projects/60917032/ does not work. It shows the loading screen but not the actual game. I get a bunch of errors like the one below, maybe those are the reason? Like it's not able to load any of the graphics?

> 860378.460 WebContent(263286): (js log) "vm" "warn" "Error loading vector image: TypeError: null cannot be converted to an object"

I think https://web.telegram.org is working? If I click "I'm feeling lucky" on the unsupported browser page it proceeds to display the login prompt.

kalenikaliaksandr (changes requested):
I've tried to compile commits in the branch one by one and `3132b23` (LibGfx: Free WebGL painting surface resources on resize) currently fails to build on mac:
```
/Users/kalenik/code/ladybird/Libraries/LibWeb/WebGL/OpenGLContext.cpp:107:2: error: #endif without #if
  107 | #endif
      |  ^
1 error generated.
ninja: build stopped: subcommand failed.
```

erik-kz:
> I've tried to compile commits in the branch one by one and `3132b23` (LibGfx: Free WebGL painting surface resources on resize) currently fails to build on mac:
> 
> ```
> /Users/kalenik/code/ladybird/Libraries/LibWeb/WebGL/OpenGLContext.cpp:107:2: error: #endif without #if
>   107 | #endif
>       |  ^
> 1 error generated.
> ninja: build stopped: subcommand failed.
> ```

Should be fixed in updated patches.

Also, the "css-background-blob-url.png" test seems to be failing for several other open PRs besides this one so I don't think this change is really the cause. It doesn't touch any relevant code.

Libraries/LibWeb/WebGL/OpenGLContext.cpp
```diff
@@ -285,16 +389,18 @@ void OpenGLContext::present(bool preserve_drawing_buffer)
...
+#elif defined(USE_VULKAN)
+    // FIXME CPU sync for now, but it would be better to export a fence and have Skia wait for it before reading from the surface
+    glFinish();
 #endif
 
     // "By default, after compositing the contents of the drawing buffer shall be cleared to their default values, as shown in the table above.
     // This default behavior can be changed by setting the preserveDrawingBuffer attribute of the WebGLContextAttributes object.
     // If this flag is true, the contents of the drawing buffer shall be preserved until the author either clears or overwrites them."
     if (!preserve_drawing_buffer) {
+        // XXX we're assuming the clear operation won't actually be submitted to the GPU
```
kalenikaliaksandr: `XXX` comment prefix is not commonly used in the codebase. maybe let's use `FIXME:` here?

Libraries/LibWeb/WebGL/OpenGLContext.cpp
```diff
@@ -285,16 +389,18 @@ void OpenGLContext::present(bool preserve_drawing_buffer)
     // With Metal, glFlush flushes the command buffer, but without waiting for it to be scheduled or completed.
     // eglWaitUntilWorkScheduledANGLE flushes the command buffer, and waits until it has been scheduled, hence the name.
     // eglWaitUntilWorkScheduledANGLE only has an effect on CGL and Metal backends, so we only use it on macOS.
-#ifdef AK_OS_MACOS
+#if defined(AK_OS_MACOS)
     eglWaitUntilWorkScheduledANGLE(m_impl->display);
-#else
-    // FIXME: When enabling WebGL for Linux, we need to use glFlush() here.
+#elif defined(USE_VULKAN)
+    // FIXME CPU sync for now, but it would be better to export a fence and have Skia wait for it before reading from the surface
```
kalenikaliaksandr: `FIXME` -> `FIXME:`. little thing but let's keep the comments style consistent 😅

Libraries/LibWeb/WebGL/OpenGLContext.cpp
```diff
@@ -237,14 +282,77 @@ void OpenGLContext::allocate_painting_surface_if_needed()
...
     glGenTextures(1, &m_impl->color_buffer);
     glBindTexture(texture_target_name == EGL_TEXTURE_RECTANGLE_ANGLE ? GL_TEXTURE_RECTANGLE_ANGLE : GL_TEXTURE_2D, m_impl->color_buffer);
     auto result = eglBindTexImage(display, m_impl->surface, EGL_BACK_BUFFER);
     VERIFY(result == EGL_TRUE);
+#elif defined(USE_VULKAN)
+    VkFormat vulkan_format = VK_FORMAT_B8G8R8A8_UNORM;
+    uint32_t drm_format = Gfx::vk_format_to_drm_format(vulkan_format);
+
+    // Ensure that our format is supported by the implementation.
+    // FIXME try other formats if not?
```
kalenikaliaksandr: same here

Libraries/LibWeb/WebGL/OpenGLContext.cpp
```diff
@@ -37,6 +34,13 @@ struct OpenGLContext::Impl {
     GLuint framebuffer { 0 };
     GLuint color_buffer { 0 };
     GLuint depth_buffer { 0 };
+
+#ifdef USE_VULKAN
+    EGLImage egl_image { EGL_NO_IMAGE };
+    struct {
+        PFNEGLQUERYDMABUFFORMATSEXTPROC QueryDmaBufFormatsEXT { nullptr };
+        PFNEGLQUERYDMABUFMODIFIERSEXTPROC QueryDmaBufModifiersEXT { nullptr };
```
kalenikaliaksandr: is diverging from snake case intentional here?

Libraries/LibGfx/VulkanContext.cpp
```diff
@@ -171,7 +186,217 @@ ErrorOr<VulkanContext> create_vulkan_context()
         .graphics_queue_family = graphics_queue_family,
         .command_pool = command_pool,
         .command_buffer = command_buffer,
+        .ext_procs = {
+            .GetMemoryFdKHR = pfnVkGetMemoryFdKHR,
+            .GetImageDrmFormatModifierPropertiesEXT = pfnVkGetImageDrmFormatModifierPropertiesEXT,
```
kalenikaliaksandr: is diverging from snake case intentional here?

Libraries/LibGfx/VulkanContext.cpp
```diff
@@ -171,7 +186,217 @@ ErrorOr<VulkanContext> create_vulkan_context()
...
+{
+    if (image != VK_NULL_HANDLE) {
+        vkDestroyImage(context.logical_device, image, nullptr);
+    }
+    if (memory != VK_NULL_HANDLE) {
+        vkFreeMemory(context.logical_device, memory, nullptr);
+    }
+}
+
+void VulkanImage::transition_layout(VkImageLayout oldLayout, VkImageLayout newLayout)
```
kalenikaliaksandr: `oldLayout` -> `old_layout` and `newLayout` -> `new_layout`.

Libraries/LibGfx/VulkanContext.cpp
erik-kz: not intentional

Libraries/LibWeb/WebGL/OpenGLContext.cpp
erik-kz: not intentional

PR #4975, opened 2025-06-02
LibWeb: Implement up/down arrow navigation in textarea

mikiubo:

PR #5868, opened 2025-08-15
LibWeb: Ensure up/down arrow navigation is grapheme-aware

trflynn89:
I went out for a run after merging #4975, and got all of 5 steps in before my brain said "but what about graphemes"

This ended up being more complicated than I thought it would be on the surface, but the end result is pretty natural feeling arrow navigation across lines:

https://github.com/user-attachments/assets/e41e8e0d-3f9f-4e23-a738-012423c9ae11


Libraries/LibWeb/HTML/FormAssociatedElement.cpp
```diff
@@ -1034,70 +1035,136 @@ static constexpr size_t find_line_end(Utf16View const& view, size_t offset)
...
+        width_to_right_edge = width_to_left_edge + current_width;
+
+        if (width_to_right_edge >= source_line_width)
+            return IterationDecision::Break;
+
+        left_edge = index;
+        width_to_left_edge += current_width;
+
+        return IterationDecision::Continue;
+    });
```
gmta, line 1102: This feels overly complex for what you'd like to achieve, which is finding the closest boundary position for a given X position from the source line (or the source line width). Couldn't you do something similar to `PaintableFragment::index_in_node_for_point()`, which calculates the distance and as soon as that distance grows, it returns the previous position?

trflynn89: I actually tried mimicking `PaintableFragment` at one point, but that algorithm is insufficient - see what happens here when you click in the middle of a grapheme:

https://github.com/user-attachments/assets/56672135-c639-4e7b-bed1-a6c0a0d3d33f

The cursor is in the wrong spot and you can add text in the middle of the grapheme.

What I'm doing in `FormAssociatedElement` here is trying to be spiritually the same while also being aware of graphemes. I was also trying to avoid shaping the entire line if we don't have to, i.e. only shape up until the point of interest.

gmta: Yeah that method is insufficient since it doesn't consider graphemes at all; I meant replicating the base logic of it instead of keeping track of left _and_ right edges, which I don't think you need to do?

trflynn89: Ah we track the edges because once we find the glyph at the desired width, we want the cursor to end up at whichever edge of that glyph it was closer to. For example see here:

https://github.com/user-attachments/assets/392058ba-481c-4249-a4d6-569e25875072

The starting point for the cursor after the "o" is somewhere in between the left-and-right edges of the space below it, but visually much closer to the left edge. If we don't track that, when we press the down arrow, we end up on the right edge.

gmta: Gotcha - compared to PaintableFragment we don't visit all the right edges of the graphemes, which you account for.

PR #5869, opened 2025-08-15
LibWeb: Implement the `color-interpolation` property for SVG gradients

tcl3:
This changes the operating color space for gradient `<linearGradient>` and `<radialGradient>` elements.

Gets us +50 WPT subtests in the `svg/painting` directory

Examples of the use of this property can be seen on  https://developer.mozilla.org/en-US/docs/Web/CSS/color-interpolation  and https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Attribute/color-interpolation.

### DIFF
--- a/Tests/LibWeb/Text/expected/wpt-import/css/cssom/shorthand-values.txt
+++ b/Tests/LibWeb/Text/expected/wpt-import/css/cssom/shorthand-values.txt
@@ -2,15 +2,15 @@ Harness status: OK
 
 Found 20 tests
 
-15 Pass
-5 Fail
+14 Pass
+6 Fail
 Pass	The serialization of border: 1px; border-top: 1px; should be canonical.
 Pass	The serialization of border: 1px solid red; should be canonical.
 Pass	The serialization of border: 1px red; should be canonical.
 Pass	The serialization of border: red; should be canonical.
 Fail	The serialization of border-top: 1px; border-right: 1px; border-bottom: 1px; border-left: 1px; border-image: none; should be canonical.
 Fail	The serialization of border-top: 1px; border-right: 1px; border-bottom: 1px; border-left: 1px; should be canonical.
-Pass	The serialization of border-top: 1px; border-right: 2px; border-bottom: 3px; border-left: 4px; should be canonical.
+Fail	The serialization of border-top: 1px; border-right: 2px; border-bottom: 3px; border-left: 4px; should be canonical.
 Fail	The serialization of border: 1px; border-top: 2px; should be canonical.
 Fail	The serialization of border: 1px; border-top: 1px !important; should be canonical.
 Fail	The serialization of border: 1px; border-top-color: red; should be canonical.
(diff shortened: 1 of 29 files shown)


Tests/LibWeb/Text/expected/wpt-import/css/cssom/shorthand-values.txt
```diff
@@ -2,15 +2,15 @@ Harness status: OK
...
+14 Pass
+6 Fail
 Pass	The serialization of border: 1px; border-top: 1px; should be canonical.
 Pass	The serialization of border: 1px solid red; should be canonical.
 Pass	The serialization of border: 1px red; should be canonical.
 Pass	The serialization of border: red; should be canonical.
 Fail	The serialization of border-top: 1px; border-right: 1px; border-bottom: 1px; border-left: 1px; border-image: none; should be canonical.
 Fail	The serialization of border-top: 1px; border-right: 1px; border-bottom: 1px; border-left: 1px; should be canonical.
-Pass	The serialization of border-top: 1px; border-right: 2px; border-bottom: 3px; border-left: 4px; should be canonical.
+Fail	The serialization of border-top: 1px; border-right: 2px; border-bottom: 3px; border-left: 4px; should be canonical.
```
tcl3, line 13: It looks like this was passing by coincidence previously. This failure can be replicated by just adding the `color-interpolation` property to `Properties.json`.

AtkinsSJ: That's really odd. What's the actual result for it now?

tcl3: The test output is:
`assert_equals: expected "border-width: 1px 2px 3px 4px; border-style: none; border-color: currentcolor;" but got "border-style: none; border-width: 1px 2px 3px 4px; border-color: currentcolor;"`

tcl3: I think these should be being sorted into [preferred order](https://drafts.csswg.org/cssom/#concept-shorthands-preferred-order), but it doesn't look like we're doing that?

AtkinsSJ: That would explain it. That's unfortunate. But yeah acceptable for now.

Libraries/LibGfx/ColorInterpolation.h
```diff
@@ -0,0 +1,16 @@
...
+ * Copyright (c) 2025, Tim Ledbetter <tim.ledbetter@ladybird.org>.
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#pragma once
+
+namespace Gfx {
+
+enum class ColorInterpolation {
```
gmta: Maybe this is a bit pedantic, but we're defining the color space the interpolation takes place in, not the type of interpolation itself. It makes sense to use the nomenclature "color interpolation" in context of CSS, but here I'd call it ColorSpace.

tcl3: I agree, that's not the best name outside of CSS. I've changed the name of the enum to `InterpolationColorSpace` to avoid potential confusion with the set of color spaces used by color functions.

gmta (approved):

PR #5870, opened 2025-08-16
LibWeb: Eagerly create stacking context if `will-change` set for property

tcl3:
This PR adds parsing and for the `will-change` property and eagerly creates stacking contexts for properties which may create a stacking context if `will-change` is set for them.

Using `will-change` to apply actual performance optimizations is left as a future exercise. 

Gives us +36 subtest passes in the `css/css-will-change` directory.


tcl3:
My latest push skips the test imported from: http://wpt.live/css/css-transforms/individual-transform/stacking-context-001.html. The reference page for this test now renders correctly, while the test itself does not.

Libraries/LibWeb/CSS/Properties.json
```diff
@@ -3556,6 +3556,20 @@
       "unitless-length"
     ]
   },
+  "will-change": {
```
AtkinsSJ, line 3568: I think this should have `"affects-stacking-context": true`. The second commit means it can create stacking contexts.

PR #5871, opened 2025-08-16
LibWeb: Use document's global object in `is_base_allowed_for_document`

Calme1709:
Previously we were using the document's window - this was both contrary to spec and causing crashes when the document did not have a window (for instance the `temp_document` in `HTMLParser::parse_html_fragment`.

This means we no longer crash when navigating between pages on https://rocketlabcorp.com


awesomekling:
Sounds like we should audit every instance of `document->window()` in the codebase!

PR #5874, opened 2025-08-16
LibWeb: Make text cursor same height as text

Calme1709:
Previously we would paint the cursor the entire height of the text fragment - this didn't look great with large line-heights. Now we only paint it the height of the actual text, with the top of the cursor aligning with the font "ascent" and the bottom the "descent".

Before:
cursor-before

After:
cursor-after


awesomekling:
That certainly looks more pleasant! Nice :)

PR #5882, opened 2025-08-17
LibWeb: Don't stop drawing scrollbar on mouseleave

rcorsi:
PaintableBox::handle_mouseleave is turning off scrollbar updating, but the user might still have the primary button down. Don't turn it off here.

Resolves crashing on MacOSX AppKit and Qt where gutter_size is 0 when mouse is moved outside window.

fixes: #5844

### DIFF
--- a/Libraries/LibWeb/Painting/PaintableBox.cpp
+++ b/Libraries/LibWeb/Painting/PaintableBox.cpp
@@ -1049,6 +1049,11 @@ Paintable::DispatchEventOfSameName PaintableBox::handle_mousemove(Badge<EventHan
 
 void PaintableBox::handle_mouseleave(Badge<EventHandler>)
 {
+    // FIXME: early return needed as MacOSX calls this even when user is pressing mouse button
+    // https://github.com/LadybirdBrowser/ladybird/issues/5844
+    if (m_scroll_thumb_dragging_direction.has_value())
+        return;
+
     auto previous_draw_enlarged_horizontal_scrollbar = m_draw_enlarged_horizontal_scrollbar;
     m_draw_enlarged_horizontal_scrollbar = false;
     if (previous_draw_enlarged_horizontal_scrollbar != m_draw_enlarged_horizontal_scrollbar)


rcorsi:
This issue in #5844 is strange for only occurring in MacOSX, with the gutter_size going to 0, but this doesn't happen on Linux.  Seems to be a smoking gun of some other undiscovered issue.

trflynn89:
Hiding the scroll bar on mouseleave was more or less the entire point of adding that event. See: #5615

rcorsi:
> Hiding the scroll bar on mouseleave was more or less the entire point of adding that event. See: #5615

I see your point.  Pushing a new version soon that addresses that.

rcorsi:
@trflynn89 are you concerned about gutter_size being 0 in MacOSX, but not on Linux?  I do find it strange that the two environments are not behaving the same.

I'll see if I can find the commit where this issue started on MacOSX , so that we might get some idea about what could be the root cause.

rcorsi:
Issue was introduced at commit 67cc02ab59375ae7520098056dd58ba20b4d44f8

@trflynn89 would you have an idea why gutter_size would be 0 in MacOSX but not in Linux.  Seems strange.

rcorsi:
@trflynn89 and just in case you didn't realize it, I did change the commit, restoring all the code that I removed, but I put an if block around it so that it only runs at the appropriate time.  Looks really good now.

Libraries/LibWeb/Painting/PaintableBox.cpp
```diff
@@ -1049,15 +1049,17 @@ Paintable::DispatchEventOfSameName PaintableBox::handle_mousemove(Badge<EventHan
...
-    auto previous_draw_enlarged_horizontal_scrollbar = m_draw_enlarged_horizontal_scrollbar;
-    m_draw_enlarged_horizontal_scrollbar = false;
-    if (previous_draw_enlarged_horizontal_scrollbar != m_draw_enlarged_horizontal_scrollbar)
-        set_needs_display();
-
-    auto previous_draw_enlarged_vertical_scrollbar = m_draw_enlarged_vertical_scrollbar;
-    m_draw_enlarged_vertical_scrollbar = false;
-    if (previous_draw_enlarged_vertical_scrollbar != m_draw_enlarged_vertical_scrollbar)
-        set_needs_display();
+    if (!m_scroll_thumb_dragging_direction.has_value()) {
```
trflynn89: Instead of indenting the entire function under an if statement, let's use an early return:
```c++
if (m_scroll_thumb_grab_position.has_value())
    return;
```

trflynn89 (commented):
The difference is just that Qt on Linux does not trigger its `leaveEvent` interface while the mouse is pressed down, whereas macOS does seem to trigger `mouseExited` regardless. Qt on macOS just carries that behavior forward.

I think ideally, we would make each UI behave consistently here - only fire the leave event once the mouse is released. I'm not sure the best way to do that, so for now I guess we can move forward with this to at least prevent a very reproducible crash.

But let's add a `// FIXME:` comment above the early return that explains the situation, and that we should make our UIs behave consistently.

Libraries/LibWeb/Painting/PaintableBox.cpp
rcorsi: No problem. I just pushed the change.  See if I did it correctly.

rcorsi:
I'll have a further look with the debugger, I got the impression that on Linux the handle_mouseleave() was indeed being called.   I was starting to suspect PaintableBox::compute_scrollbar_data() had some issue and setting the scrollbar_data.gutter_rect to zeros on MacOSX but not on Linux.

trflynn89:
All I did was add a print statement to the leave event:
```diff
» git --no-pager diff
diff --git a/UI/Qt/WebContentView.cpp b/UI/Qt/WebContentView.cpp
index 259062151b1..29272168f32 100644
--- a/UI/Qt/WebContentView.cpp
+++ b/UI/Qt/WebContentView.cpp
@@ -404,6 +404,7 @@ QVariant WebContentView::inputMethodQuery(Qt::InputMethodQuery) const

 void WebContentView::leaveEvent(QEvent* event)
 {
+    dbgln("WebContentView::leaveEvent");
     static QMouseEvent mouse_event { QEvent::Type::Leave, {}, {}, Qt::MouseButton::NoButton, Qt::MouseButton::NoButton, Qt::KeyboardModifier::NoModifier };
     enqueue_native_event(Web::MouseEvent::Type::MouseLeave, mouse_event);
```

It then doesn't fire (on Linux) when I leave the area while the mouse is pressed. It fires once I let go.


https://github.com/user-attachments/assets/683cb94e-1ca3-416c-9395-6ebe057a0201

rcorsi:
@trflynn89 Your right about PaintableBox::handle_mouseleave not being called on Linux.

So what about #5860, is that a dead end or something useful there?  Should we continue the discussion there?

PR #5883, opened 2025-08-17
LibWeb/HTML: Check all radio buttons in group for required attribute

skyz1:
This fixes a bug in the algorithm for determining if radio buttons are missing their value. Previously it was only checked if the button itself is required. Now the algorithm checks if the radio button group contains a required radio button in order to determine if the value is required.


Libraries/LibWeb/HTML/HTMLInputElement.cpp
```diff
@@ -3323,7 +3324,7 @@ bool HTMLInputElement::suffering_from_being_missing() const
             if (is_in_same_radio_button_group(*this, element)) {
                 if (element.checked())
                     has_checkedness_false_for_all_elements_in_group = false;
-                if (has_attribute(HTML::AttributeNames::required))
+                if (element.has_attribute(HTML::AttributeNames::required))
```
gmta, line 3327: Commit message: could you try to avoid "fix" in the title and instead describe what you changed? E.g.

LibWeb: Check all radio buttons in group for required attribute

PR #5884, opened 2025-08-17
LibWeb/HTML: Account for floating point error when calculating step mismatch

skyz1:
This introduces a new `is_integral_multiple` method that replaces the `fmod(...) != 0` check. This fixes some WPT-Tests, where the step cannot be accurately represented as a double.

In the long term we might want to replace doubles with a type that has a higher precision (e.g. a custom decimal implementation).

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLInputElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLInputElement.cpp
@@ -60,6 +60,25 @@
 #include <LibWeb/WebIDL/DOMException.h>
 #include <LibWeb/WebIDL/ExceptionOr.h>
 
+namespace {
+
+bool is_integral_multiple(double value, double step)
+{
+    VERIFY(step > 0);
+
+    // The step is so small, that the integral multiple closest to the value cannot be accurately expressed as a double
+    if (fabs(value) / ldexp(1.0, NumericLimits<double>::digits()) > step)
+        return true;
+
+    auto division_remainder = fabs(remainder(value, step));
+    // NOTE: There is no spec for the integral multiple. However, chromium uses tolerance so we comply here.
+    //       https://github.com/chromium/chromium/blob/18aeefb6d0fe94e267c08e1aaeaf2632937f4ce2/third_party/blink/renderer/core/html/forms/step_range.cc#L67
+    double acceptable_error = round(step) == step ? 0 : step / static_cast<double>(1 << NumericLimits<float>::digits());
+    return division_remainder <= acceptable_error || division_remainder >= (step - acceptable_error);
+}
+
+}
+
 namespace Web::HTML {
 
 GC_DEFINE_ALLOCATOR(HTMLInputElement);
@@ -2913,7 +2932,7 @@ WebIDL::ExceptionOr<void> HTMLInputElement::step_up_or_down(bool is_down, WebIDL
 
     // 7. If value subtracted from the step base is not an integral multiple of the allowed value step, then set value to the nearest value that,
     // when subtracted from the step base, is an integral multiple of the allowed value step, and that is less than value if the method invoked was the stepDown() method, and more than value otherwise.
-    if (fmod(step_base() - value, allowed_value_step) != 0) {
+    if (!is_integral_multiple(step_base() - value, allowed_value_step)) {
         if (is_down) {
             value = step_base() + floor((value - step_base()) / allowed_value_step) * allowed_value_step;
         } else {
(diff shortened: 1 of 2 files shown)


gmta:
❗doesn't build

skyz1:
Sorry that I didn't fix that right away, but it worked now.

Libraries/LibWeb/HTML/HTMLInputElement.cpp
```diff
@@ -59,6 +59,24 @@
 #include <LibWeb/WebIDL/DOMException.h>
 #include <LibWeb/WebIDL/ExceptionOr.h>
 
+namespace {
+
+// In the longterm we might want to replace doubles with a type that has a higher precision (e.g. a manual decimal implementation).
```
gmta: Do we know of cases where this actually is needed to support the web? If so, consider making this a `// FIXME` or create a GitHub issue with a test case. Otherwise, I'd leave this comment out.

skyz1: I'm not entirely sure, I just had this thought because Firefox does it this way and can therefore handle extremely small/large numbers. I'm not sure, but the spec only requires 32-bit float, right? In that case, it wouldn't be necessary, and I think inaccuracies in the calculation would be in line with the specification.

gmta: In that case I'd remove this comment from the code, and if you find some cases where it's actually necessary to have more precision we can always raise an issue / create a new PR. No need to keep track of things that seem to work just fine imo :)

Libraries/LibWeb/HTML/HTMLInputElement.cpp
```diff
@@ -60,6 +60,23 @@
...
+bool is_integral_multiple(double value, double step)
+{
+    VERIFY(step > 0);
+
+    // The step is so small, that the integral multiple closest to the value cannot be accurately expressed as a double
+    if (fabs(value) / ldexp(1.0, NumericLimits<double>::digits()) > step)
+        return true;
+
+    auto division_remainder = fabs(remainder(value, step));
+    double acceptable_error = round(step) == step ? 0 : step * NumericLimits<float>::epsilon();
```
gmta: Why do we take the float's epsilon instead of a double's here?

skyz1: I adopted this from chromium/blink and they use the float epsilon there. Their code also contains [this comment](https://chromium.googlesource.com/chromium/src/+/HEAD/third_party/blink/renderer/core/html/forms/step_range.cc#65) saying that the double epsilon should be used according to the spec. I tried the double epsilon, but this does not satisfy [this test](https://wpt.fyi/results/html/semantics/forms/constraints/form-validation-validity-stepMismatch.html?label=master&label=experimental&aligned). Should I still use the double epsilon?

gmta: If you can find the relevant spec link, could you add it as a comment?

Also, since this is a bit surprising, we would at least need to add a comment here explaining why you suddenly switch to float when all the other types are doubles.

skyz1: The spec does not basically just talks about the "integral multiple" and does not specify further what this exactly means. It only specifies the type of the involved numbers as IEEE 754 double-precision floating-point values. 

I looked at the chromium history and it seems like it got carried over from webkit, and goes back to [this commit](https://github.com/chromium/chromium/commit/4b14ce44def041237c07da8a643af6ac2608408d) as far as I can tell.

I also just noticed that I made a mistake and chromium and webkit actually use `NumericLimits<float>::epsilon()/2`...

I see two options for moving forward.
1. We scrap this change as the spec does not actually require any tolerance when performing the integral multiple. However that would mean we deviate from the other browsers.
2. We use Chromiums/WebKits tolerance and comply with the other browsers, but apart from that, we don't really have an explanation for why we do it.

Which do you prefer?

gmta: I'd say comply + add a note comment!

skyz1: Sure, I adjusted the calculation to match [chromium](https://chromium.googlesource.com/chromium/src/+/HEAD/third_party/blink/renderer/core/html/forms/step_range.cc#67) more closely and added a note.

Libraries/LibWeb/HTML/HTMLInputElement.cpp
```diff
@@ -60,6 +60,24 @@
...
+bool is_integral_multiple(double value, double step)
+{
+    VERIFY(step > 0);
+
+    // The step is so small, that the integral multiple closest to the value cannot be accurately expressed as a double
+    if (fabs(value) / ldexp(1.0, NumericLimits<double>::digits()) > step)
+        return true;
+
+    auto division_remainder = fabs(remainder(value, step));
+    // AD-HOC: The spec does not specify any tolerance. However, chromium uses tolerance so we comply here.
```
konradekk: your comment refers to the spec, but it’d be good to have it cited & linked above.

skyz1: There is no spec for integral multiple. However I now linked the specific line in the chromium source as well.

PR #5888, opened 2025-08-17
LibWeb: Remove unnecessary as_if<SVGGradientPaintStyle>() on paint_style

kennethmyhra:
This causes a compiler error on GCC 15.0.1 and should be unnecessary since paint_style already is of type SVGGradientPaintStyle.


tcl3 (approved):
Oops! Thanks for catching this.

PR #5889, opened 2025-08-17
LibWeb: Implement resizable ArrayBuffers for Wasm memories

CountBleck:
Work-in-progress.

Stuff that's missing:
* `HostResizeArrayBuffer` for Wasm memories (requires "the surrounding agent’s associated Memory object cache")
* `HostGrowSharedArrayBuffer` for Wasm memories (requires the aforementioned cache) (also requires implementing, not just that hook, but growable `SharedArrayBuffer`s as a whole in LibJS)
* `toFixedLengthBuffer()` for shared Wasm memories (a very different case from non-shared Wasm memories, and requires the aforementioned cache)

The goal is to pass [these Interop 2025 WPT tests for WebAssembly memories](https://wpt.fyi/results/wasm/jsapi/memory?label=master&product=chrome%5Bexperimental%5D&product=firefox%5Bexperimental%5D&product=safari%5Bexperimental%5D&product=ladybird&aligned&view=interop&q=label%3Ainterop-2025-webassembly).

Bonus fix: the last WPT subtest in [`wasm/jsapi/memory/grow.any.html`](https://wpt.fyi/results/wasm/jsapi/memory/grow.any.html?label=master&product=chrome[experimental]&product=firefox[experimental]&product=safari[experimental]&product=ladybird&aligned).


CountBleck:
Some extra attention might be warranted to the change I made to `ArrayBuffer` in LibJS...I believe that code fully functions but it feels a bit questionable to me.

Libraries/LibJS/Runtime/ArrayBuffer.h
```diff
@@ -35,10 +35,24 @@ struct DataBlock {
...
+
     ByteBuffer& buffer()
     {
         ByteBuffer* ptr { nullptr };
-        byte_buffer.visit([&](Empty) { VERIFY_NOT_REACHED(); }, [&](auto* pointer) { ptr = pointer; }, [&](auto& value) { ptr = &value; });
+        byte_buffer.visit(
+            [&](Empty) { VERIFY_NOT_REACHED(); },
+            [&](ByteBuffer& value) { ptr = &value; },
+            [&](UnownedFixedLengthByteBuffer& value) { ptr = value.buffer; });
         return *ptr;
```
alimpfard: Now that we're here,
```suggestion
        return byte_buffer.visit(
            [](Empty) -> ByteBuffer* { VERIFY_NOT_REACHED(); },
            [](ByteBuffer& value) { return &value; },
            [](UnownedFixedLengthByteBuffer& value) { return value.buffer; });
```

Libraries/LibJS/Runtime/ArrayBuffer.h
```diff
@@ -35,10 +35,24 @@ struct DataBlock {
         Yes,
     };
 
+    struct UnownedFixedLengthByteBuffer {
+        UnownedFixedLengthByteBuffer(ByteBuffer* buffer)
+            : buffer(buffer)
+            , size(buffer ? buffer->size() : 0)
```
alimpfard, line 41: Can this ever be passed a null?

Libraries/LibWeb/WebAssembly/Memory.cpp
```diff
@@ -80,32 +95,42 @@ WebIDL::ExceptionOr<u32> Memory::grow(u32 delta)
...
-            MUST(JS::detach_array_buffer(vm, *m_buffer, JS::PrimitiveString::create(vm, "WebAssembly.Memory"_string)));
+            MUST(JS::detach_array_buffer(vm, *buffer, JS::PrimitiveString::create(vm, "WebAssembly.Memory"_string)));
         }
+
+        // 2. Let newBuffer be the result of creating a fixed length memory buffer from memaddr.
+        // 3. Set memory.[[BufferObject]] to newBuffer.
     }
 
-    m_buffer = TRY(create_a_fixed_length_memory_buffer(vm, realm, m_address, m_shared));
+    buffer = TRY(create_a_fixed_length_memory_buffer(vm, realm, address, memory->m_shared));
```
alimpfard: should the assignment & co be moved into the if? the spec steps you have seem to suggest so at least.

Libraries/LibWeb/WebAssembly/Instance.cpp
```diff
@@ -71,10 +71,11 @@ void Instance::initialize(JS::Realm& realm)
                 m_exports->define_direct_property(name, *object, JS::default_attributes);
```
alimpfard, line 71: commit message: s/distint/distinct

Libraries/LibWeb/WebAssembly/Memory.cpp
```diff
@@ -100,8 +100,92 @@ WebIDL::ExceptionOr<u32> Memory::grow(u32 delta)
...
+        // 1. If IsFixedLengthArrayBuffer(buffer) is true, return buffer.
+        if (m_buffer && m_buffer->is_fixed_length())
+            return GC::Ref(*m_buffer);
+
+        // 2. Otherwise,
+        // 1. Let fixedBuffer be the result of creating a fixed length memory buffer from memaddr.
+        auto fixed_buffer = TRY(create_a_fixed_length_memory_buffer(vm, realm(), m_address, m_shared));
+
+        // 2. Perform ! DetachArrayBuffer(buffer, "WebAssembly.Memory").
+        if (m_buffer)
```
alimpfard: :thinking: the spec seems to imply that `m_buffer` is never null?

Libraries/LibWeb/WebAssembly/Memory.cpp
```diff
@@ -128,10 +212,12 @@ WebIDL::ExceptionOr<void> Memory::refresh_the_memory_buffer(JS::VM& vm, JS::Real
 
         // 2. Let newBuffer be the result of creating a fixed length memory buffer from memaddr.
         // 3. Set memory.[[BufferObject]] to newBuffer.
+        buffer = TRY(create_a_fixed_length_memory_buffer(vm, realm, address, memory->m_shared));
+    } else {
+        auto& bytes = cache.abstract_machine().store().get(address)->data();
```
alimpfard, line 221: No spec for this bit?

Libraries/LibJS/Runtime/ArrayBuffer.h
CountBleck: Yes. For whatever reason, `allocate_array_buffer` in `ArrayBuffer.cpp` calls `ordinary_create_from_constructor` with a `nullptr` (I'd assume it should use an empty data block, but idk). In that case, the data block is set immediately after the `ArrayBuffer` is constructed. Not handling that null led test-js to segfault (due to `buffer->size()`).

Libraries/LibWeb/WebAssembly/Memory.cpp
CountBleck: The existing code initialized `m_buffer` lazily upon first access to the JS `buffer` getter (`Memory::buffer()`). I don't know the reason, and it's not exactly according to the spec. I think I'll have to make it not-lazy anyway, because the `buffer` getter in the threads JS API spec is actually a bit more complicated, and `m_buffer` being lazy would get in the way.

Libraries/LibWeb/WebAssembly/Memory.cpp
CountBleck: i forgor 😭

Libraries/LibWeb/WebAssembly/Memory.cpp
CountBleck: I did that in a later commit, since I was going to touch that function later anyway. There weren't any functional changes to `refresh_the_memory_buffer` in this commit besides making it static, using the "Memory object cache", and the renames associated with the two.

Libraries/LibWeb/WebAssembly/Memory.cpp
CountBleck: Okay, I removed all `m_buffer` checks in my latest changes (which makes `m_buffer` non-lazy).

Libraries/LibWeb/WebAssembly/Memory.cpp
CountBleck: Fixed.

Libraries/LibJS/Runtime/ArrayBuffer.h
```diff
@@ -35,11 +35,23 @@ struct DataBlock {
...
+        {
+        }
+
+        ByteBuffer* buffer;
+        size_t size;
```
alimpfard: style nit: please give these a null/zero default value.

Libraries/LibJS/Runtime/ArrayBuffer.h
```diff
@@ -35,11 +35,23 @@ struct DataBlock {
         Yes,
     };
 
+    struct UnownedFixedLengthByteBuffer {
+        UnownedFixedLengthByteBuffer(ByteBuffer* buffer)
```
alimpfard: and mark this as `explicit` (you're already calling it explicitly)

PR #5890, opened 2025-08-17
LibJS: Revert common error types to only hold a single string

trflynn89:
This brings Octane back down to previous numbers on my machine.


gmta (approved):

gmta:
Thanks for looking into this!

PR #5897, opened 2025-08-18
LibWeb/CSS: Use response to get stylesheet encoding in at-rule import

criskell:
A change to fix [this WPT test](https://wpt.live/css/CSS2/syntax/at-charset-002.xht) (EDIT: Also:  [4º](https://wpt.live/css/CSS2/syntax/at-charset-004.xht), [5º](https://wpt.live/css/CSS2/syntax/at-charset-005.xht) and [6º](https://wpt.live/css/CSS2/syntax/at-charset-006.xht)).

According to the [CSS Syntax spec](https://drafts.csswg.org/css-syntax/#input-byte-stream):

> If HTTP or equivalent protocol provides an encoding label (e.g. via the charset parameter of the Content-Type header) for the stylesheet, [get an encoding](https://encoding.spec.whatwg.org/#concept-encoding-get) from encoding label. If that does not return failure, return it.

### DIFF
--- a/Libraries/LibWeb/CSS/CSSImportRule.cpp
+++ b/Libraries/LibWeb/CSS/CSSImportRule.cpp
@@ -140,8 +140,29 @@ void CSSImportRule::fetch()
 
             // 4. Let importedStylesheet be the result of parsing byteStream given parsedUrl.
             // FIXME: Tidy up our parsing API. For now, do the decoding here.
-            // FIXME: Get the encoding from the response somehow.
+
+            // https://drafts.csswg.org/css-syntax/#determine-the-fallback-encoding
             auto encoding = "utf-8"sv;
+
+            // 1. If HTTP or equivalent protocol provides an encoding label (e.g. via the charset parameter of the Content-Type header)
+            //    for the stylesheet, get an encoding from encoding label. If that does not return failure, return it.
+            auto extracted_mime_type = response->header_list()->extract_mime_type();
+
+            if (extracted_mime_type.has_value()) {
+                auto maybe_encoding = extracted_mime_type->parameters().get("charset"sv);
+
+                if (maybe_encoding.has_value())
+                    encoding = maybe_encoding.value();
+            }
+
+            // FIXME: 2. Otherwise, check stylesheet’s byte stream. If the first 1024 bytes of the stream begin with the hex sequence
+            //           40 63 68 61 72 73 65 74 20 22 XX* 22 3B
+            //           where each XX byte is a value between 0₁₆ and 21₁₆ inclusive or a value between 23₁₆ and 7F₁₆ inclusive,
+            //           then get an encoding from a string formed out of the sequence of XX bytes, interpreted as ASCII.
+            //           If the return value was utf-16be or utf-16le, return utf-8; if it was anything else except failure, return it.
+            // FIXME: 3. Otherwise, if an environment encoding is provided by the referring document, return it.
+            // FIXME: 4. Otherwise, return utf-8.
+
             auto maybe_decoder = TextCodec::decoder_for(encoding);
             if (!maybe_decoder.has_value()) {
                 dbgln_if(CSS_LOADER_DEBUG, "CSSImportRule: Failed to decode CSS file: {} Unsupported encoding: {}", parsed_url, encoding);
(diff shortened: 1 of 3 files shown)


Libraries/LibWeb/CSS/CSSImportRule.cpp
```diff
@@ -140,8 +140,17 @@ void CSSImportRule::fetch()
...
+
             auto encoding = "utf-8"sv;
+            auto extracted_mime_type = response->header_list()->extract_mime_type();
+
+            if (extracted_mime_type.has_value()) {
+                auto maybe_encoding = extracted_mime_type->parameters().get("charset"sv);
+
+                if (maybe_encoding.has_value())
+                    encoding = maybe_encoding.value();
+            }
```
gmta, line 156: Instead of adding this code to the existing steps, we should have a separate method implementing the ["determine the fallback encoding" algorithm](https://drafts.csswg.org/css-syntax/#determine-the-fallback-encoding). As far as I can see, this behaves the same as the current code but having it as a separate allows it to be reused in the future and makes it easier to compare its implementation to the spec text.

Libraries/LibWeb/CSS/CSSImportRule.cpp
gmta, line 1: Could you add the relevant bits of your PR description to the commit message?

Libraries/LibWeb/CSS/CSSImportRule.cpp
criskell: What this does is check the `charset` of the `Content-Type`, and if it doesn't match, we'll default to UTF-8.

This actually also fixes tests [4](https://wpt.live/css/CSS2/syntax/at-charset-004.xht), [5](https://wpt.live/css/CSS2/syntax/at-charset-005.xht) and [6](https://wpt.live/css/CSS2/syntax/at-charset-006.xht), I forgot to mention in the PR.

I can try to complete the other steps of this algorithm in a separate PR, okay? I can also create an issue reporting the problem.

gmta: Why in a separate PR? The algorithm is 4 steps. You don't need to implement all steps immediately, you can leave them as `// FIXME:`s.

Libraries/LibWeb/CSS/CSSImportRule.cpp
criskell: For sure!

Libraries/LibWeb/CSS/CSSImportRule.cpp
criskell: Got it, thanks!
I added // FIXMEs and comments explaining the first step of the algorithm, inspired by the other files.

> Why in a separate PR? 

Because this PR would only implement the first step of the algorithm, I thought it was necessary to implement it all at once.

Tests/LibWeb/Text/input/css/import-rule-shift-jis.html
```diff
@@ -1,10 +1,13 @@
 <!DOCTYPE html>
 <style>
-    @import "data:text/css;charset=shift_jis;base64,77u/I2RpdjIKewogICAgY29sb3I6IGdyZWVuOwp9Cgoulb2YYQp7CiAgICBjb2xvcjogcmVkOwp9Cg==";
+    @import "data:text/css;charset=shift_jis;base64,QGNoYXJzZXQgIndpbmRvd3MtMTI1MiI7Ci6VvZhhLCAjZGl2Mgp7CiAgICBjb2xvcjogZ3JlZW47Cn0=";
```
criskell, line 3: Unfortunately, it's not possible to import WPT tests because the test runner currently doesn't support [".headers"](https://github.com/web-platform-tests/wpt/blob/247ab0f5c7b9c83b49f4a1fbe30884a468ac3e94/css/CSS2/syntax/support/at-charset-002.css.headers) files, which would be necessary to mock the Content-type header.
Alternatively, I used data URLs.

Libraries/LibWeb/CSS/CSSImportRule.cpp
```diff
@@ -140,8 +140,29 @@ void CSSImportRule::fetch()
...
+
+            // FIXME: 2. Otherwise, check stylesheet’s byte stream. If the first 1024 bytes of the stream begin with the hex sequence
+            //           40 63 68 61 72 73 65 74 20 22 XX* 22 3B
+            //           where each XX byte is a value between 0₁₆ and 21₁₆ inclusive or a value between 23₁₆ and 7F₁₆ inclusive,
+            //           then get an encoding from a string formed out of the sequence of XX bytes, interpreted as ASCII.
+            //           If the return value was utf-16be or utf-16le, return utf-8; if it was anything else except failure, return it.
+            // FIXME: 3. Otherwise, if an environment encoding is provided by the referring document, return it.
+            // FIXME: 4. Otherwise, return utf-8.
```
AtkinsSJ, line 164: I think you missed it in Jelle's comment before, but please put this section in a separate function. Otherwise the step numbering gets confusing.

PR #5899, opened 2025-08-18
LibWeb: Use `shape-rendering` to control anti aliasing for SVG paths

tcl3:
Anti-aliasing is disabled if `shape-rendering` is set to `optimizeSpeed` or `crispEdges`.

Here's an example from https://developer.mozilla.org/en-US/docs/Web/CSS/shape-rendering - previously all of these ellipses looked the same:
image


AtkinsSJ (approved):
Looks good to me!

PR #5902, opened 2025-08-18
LibWeb: Implement the Gamepad API with SDL3

Lubrsi:
https://github.com/user-attachments/assets/d555b72e-07d6-462e-b68d-29e384d8030c


Libraries/LibWeb/CMakeLists.txt
```diff
@@ -366,8 +366,8 @@ set(SOURCES
     FileAPI/FileList.cpp
     FileAPI/FileReader.cpp
     FileAPI/FileReaderSync.cpp
-    Gamepad/Gamepad.cpp
-    Gamepad/NavigatorGamepad.cpp
```
trflynn89, line 385: :thinking: a bit confused by this because we have other namespace/class repeats like this, e.g. `Clipboard::Clipboard` and `DOMURL::DOMURL`. Is something in the bindings not using the interface's fully qualified name?

Libraries/LibWeb/GamepadAPI/Gamepad.cpp
```diff
@@ -21,4 +149,478 @@ void Gamepad::initialize(JS::Realm& realm)
...
+void Gamepad::visit_edges(Cell::Visitor& visitor)
+{
+    Base::visit_edges(visitor);
+    visitor.visit(m_buttons);
+    visitor.visit(m_vibration_actuator);
+}
+
+void Gamepad::finalize()
+{
+    SDL_CloseGamepad(m_sdl_gamepad);
```
gmta: We're only closing the SDL resource once the cell gets finalized by GC - is that OK? Or should we be more proactive about this? (e.g. on gamepad disconnect)

Lubrsi: As far as I can tell it's okay, it'll just stop received updates, plus joystick IDs will never be reused in the applications lifetime. Additionally, SDL_CloseGamepad frees all the memory associated with the gamepad (https://github.com/libsdl-org/SDL/blob/main/src/joystick/SDL_gamepad.c#L3918-L3922), so you could cause a UAF by holding onto the haptic actuator and sending a rumble effect to the dangling `SDL_Gamepad*` when a gamepad is disconnected. Haptic actuator holds onto a reference to the Gamepad object and visits it, so that shouldn't happen.

Libraries/LibWeb/CMakeLists.txt
Lubrsi: Turns out generate_to_cpp did not account for type names conflicting with namespaces. `Clipboard::Clipboard` was working because it was only used in impl_from, which takes account of this.

Issue #5893, opened 2025-08-18
GFC::layout_absolutely_positioned_element() causes a crash on Tweakers.net


gmta:
Since #5885 was merged, trying to load https://tweakers.net immediately results in a crash:

```
VERIFICATION FAILED: i < m_size at /Users/jelle/Projects/ladybird/AK/Vector.h:165
0   liblagom-ak.0.0.0.dylib             0x0000000104c016a0 dump_backtrace + 72
1   liblagom-ak.0.0.0.dylib             0x0000000104c01958 ak_trap + 12
2   liblagom-ak.0.0.0.dylib             0x0000000104c01a04 ak_assertion_failed + 0
3   liblagom-web.0.0.0.dylib            0x0000000108ec6c10 AK::ErrorOr<AK::Empty, AK::Error>::release_error() (.cold.1) + 0
4   liblagom-web.0.0.0.dylib            0x0000000108373504 Web::Layout::GridFormattingContext::layout_absolutely_positioned_element(Web::Layout::Box const&) + 196
5   liblagom-web.0.0.0.dylib            0x0000000108373c64 Web::Layout::GridFormattingContext::parent_context_did_dimension_child_root_box() + 208
...
```

gmta:
Fixed by #5903

PR #5903, opened 2025-08-18
LibWeb: Handle correctly case when abspos and relpos GFC is interleaved

kalenikaliaksandr:
...by another GFC. When searching for grid item that contains an abspos box positioned relative to GFC, it's incorrect to assume that the closest ancestor box whose parent establishes GFC is always the one we are looking for, because there may be non-positioned GFC in between.

Fixes regression introduced in 80c8e78.


gmta:
This PR fixes #5893

PR #5906, opened 2025-08-19
LibWeb: Propagate border-box dimensions when getting max content width

Calme1709:
This means that we now calculate the inner width correctly for `display: inline-block` nodes when we have `box-sizing: border-box` and `min-width`, as we would previously assume these dimensions were all `0`


gmta (approved):

gmta:
@Calme1709 the new test in this PR is failing CI.

Calme1709:
Looks like that is because the expected file was generated before f4b04beccdf6295089c713282819ac1e4afd9500 - rebaselined and rebased.

gmta (approved):

Issue #245, opened 2024-06-22
LibWeb: Float issue with full width divs


jamierocks:
## Test Program

```html
<div style="width: 100%; float: left; background-color: red;">
    <h2>Well</h2>
    <p>well</p>
</div>
<div style="width: 100%; float: left; background-color: red;">
    <h2>Hello</h2>
    <p>hello</p>
</div>
<div style="width: 100%; float: left; background-color: red;">
    <h2>Friends</h2>
    <p>friends</p>
</div>
```

## Expected Result (Firefox)

![image](https://github.com/LadybirdBrowser/ladybird/assets/5103549/4bb1b3e4-9f0c-4a03-a945-34e22678121c)

## Actual Result (Ladybird)

![image](https://github.com/LadybirdBrowser/ladybird/assets/5103549/c6fbf5e2-0ed3-4905-856f-3abd5706f8fc)

Note the white line between the second and third divs.

## Thoughts

Might be related to GH-107 and GH-231. Might be a similar issue to https://github.com/SerenityOS/serenity/pull/24241.

jamierocks:
I think this might be a rounding issue, the issue seems to occur between all divs if you zoom in.

![image](https://github.com/LadybirdBrowser/ladybird/assets/5103549/10d94e16-a2dc-4e9e-84b8-3e61c0332d80)

shannonbooth:
This appears to have been fixed now:

![Image](https://github.com/user-attachments/assets/4ee0904f-bc2d-4485-9d11-a7eae6bb4901)

PR #5911, opened 2025-08-19
LibWeb: Maintain rect positioning when rounding to device pixel rects

gmta:
When rounding a CSSPixelRect to a DevicePixelRect, we simply pulled its width and height through round() and called it a day. Unfortunately this could negatively affect the rect's perceived positioning.

A rect at { 0.5, 0.0 } with size { 19.5 x 20.0 } should have its right edge at position 20, but after rounding it would end up at { 1, 0 } with size { 20 x 20 }, causing its right edge to be at position 21 instead.

Fix this by first rounding the right and bottom edges of the input rect, and then determining the dimensions by subtracting its rounded position.

Gets rid of the grid gap artifacts on https://duotrigordle.com:

| Before | After |
|--|--|
| Screenshot 2025-08-19 at 13 39 57 | Screenshot 2025-08-19 at 13 41 03 |


jamierocks:
I wonder if this might resolve GH-245.

gmta:
> I wonder if this might resolve [GH-245](https://github.com/LadybirdBrowser/ladybird/issues/245).

It seems like it does!

| Before | After |
|--|--|
| image | image |

PR #5912, opened 2025-08-19
Meta: Use `pathlib.Path(file).parent.parent` instead of...

h3xOo:
... `os.path.dirname(file) + "/.."` in Python scripts

This makes them a little more system agnostic.

Since Python 3.6 `chdir` accepts [path-like object](https://docs.python.org/3/library/os.html#os.chdir).

### DIFF
--- a/Meta/check-html-doctype.py
+++ b/Meta/check-html-doctype.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 import os
+import pathlib
 import re
 import subprocess
 import sys
@@ -52,5 +53,5 @@ def run():
 
 
 if __name__ == "__main__":
-    os.chdir(os.path.dirname(__file__) + "/..")
+    os.chdir(pathlib.Path(__file__).parent.parent)
     run()
--- a/Meta/check-idl-files.py
+++ b/Meta/check-idl-files.py
@@ -86,5 +86,5 @@ def run():
 
 
 if __name__ == "__main__":
-    os.chdir(os.path.dirname(__file__) + "/..")
+    os.chdir(pathlib.Path(__file__).parent.parent)
     run()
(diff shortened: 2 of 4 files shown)


h3xOo:
Wasn't an ellipsis used for long commit titles?

If not, how could I rename it?

Hendiadyoin1:
Ellipses are fine I think the commit linter does not like triple dots though as it sees that as a dot at the end,

You can use an ellipsis character instead `…`

(Single commit titles are best changed with `git commit --amend`)

h3xOo:
> Ellipses are fine I think the commit linter does not like triple dots though as it sees that as a dot at the end,
> 
> You can use an ellipsis character instead `…`
> 
> (Single commit titles are best changed with `git commit --amend`)

Thanks!

alimpfard:
Do note that these are not semantically equal (whether that's a problem is probably worth debating), `.parent` is a purely lexical operation whereas `/..` will go one level up from the resolved path (e.g. `foo/bar -> /a/b/c`, then `foo/bar/..` is `/a/b` but `"foo/bar".parent` is `foo`)

h3xOo:
> Do note that these are not semantically equal (whether that's a problem is probably worth debating), `.parent` is a purely lexical operation whereas `/..` will go one level up from the resolved path (e.g. `foo/bar -> /a/b/c`, then `foo/bar/..` is `/a/b` but `"foo/bar".parent` is `foo`)

I can add some `.resolve(strict=True)` call(s)

alimpfard:
That's fine, just giving a heads-up to make sure we don't break anything intended.
Not sure if we even want to care about this.

h3xOo:
Well, every change breaks someone's workflow: https://xkcd.com/1172/

But I think it should work, as we're not leaving Ladybird's git root directory when `chdir`'ing.

h3xOo:
Go away, stalebot

PR #5914, opened 2025-08-19
LibGfx: Only compile dma-buf-related code on Linux

erik-kz:
Shareable Vulkan image allocation on Linux relies on the dma-buf interface, which is a Linux-specific thing. Therefore, we should only be compiling it (and any code that uses it) on Linux. This change adds preprocessor guards to do that. Enabling similar functionality on other operating systems will need to leverage analogous interfaces on those platforms, e.g. win32 handles on Windows.


Libraries/LibWeb/WebGL/OpenGLContext.cpp
```diff
@@ -252,7 +268,7 @@ void OpenGLContext::allocate_painting_surface_if_needed()
 
     EGLint texture_target_name = 0;
 
-#if defined(AK_OS_MACOS)
+#    if defined(AK_OS_MACOS)
```
kalenikaliaksandr: this indentation change looks suspicious. I think it was fine and we don't need these extra spaces?

Libraries/LibWeb/WebGL/OpenGLContext.cpp
```diff
@@ -289,7 +305,7 @@ void OpenGLContext::allocate_painting_surface_if_needed()
     glBindTexture(texture_target_name == EGL_TEXTURE_RECTANGLE_ANGLE ? GL_TEXTURE_RECTANGLE_ANGLE : GL_TEXTURE_2D, m_impl->color_buffer);
     auto result = eglBindTexImage(display, m_impl->surface, EGL_BACK_BUFFER);
     VERIFY(result == EGL_TRUE);
-#elif defined(USE_VULKAN)
+#    elif defined(USE_VULKAN_IMAGES)
```
kalenikaliaksandr: same here

kalenikaliaksandr (changes requested):
commit has changes in both LibGfx and LibWeb. in such cases we prefix commit name with both "LibGfx+LibWeb:".

Libraries/LibWeb/WebGL/OpenGLContext.cpp
erik-kz: I think it's because I've guarded the entire function body with "#ifdef ENABLE_WEBGL". I agree it looks a bit weird. I suppose I could move the nested-conditional blocks into helper functions, that might make it more readable.

PR #5918, opened 2025-08-19
Meta+UI: Add scaffolding for a GTK 4 UI 

ADKaster:
The example app uses GTK 4 and libadwaita.

The GTK vcpkg dependencies are separated via feature flags. While
mutually exclusive flags are banned in upstream vcpkg, we are an end
application and thus can do whatever we want. It just so happens that
the features feature of vcpkg manifests is a convenient way to control
what vcpkg dependencies are used from CMake.

The new config can be selected by setting the CMake cache variable `-DLADYBIRD_GUI_FRAMEWORK=GTK`

Draft: Several of these overlay port updates could/should be upstreamed.

Notable exceptions: `gtk`, as I both updated *and* set `-Dwayland-backend=true -Dx11-backend=false -Dvulkan=enabled`


jdahlin:
As far as I know, none of the other browsers using gtk (web/epiphany, firefox to some extent, webkitgtk) depends on gtkmm/glibmm, they call the C API directly.

ADKaster:
That's fine, but I don't want to write C. Looking around it seems like there's at least one active maintainer for gtkmm, and at least one large app (Inkscape) is making heavy use of them.

Meta/CMake/vcpkg/overlay-ports/gtk/portfile.cmake
```diff
@@ -19,22 +16,25 @@ vcpkg_add_to_path("${PKGCONFIG_DIR}") # Post install script runs pkg-config so i
...
+set(wayland false)
+set(vulkan disabled)
 set(win32 false)
 set(osx false)
 if(VCPKG_TARGET_IS_LINUX)
-    set(OPTIONS -Dwayland-backend=false) # CI missing at least wayland-protocols
-    set(x11 true)
-    # Enable the wayland gdk backend (only when building on Unix except for macOS)
+    set(wayland true)
+    set(vulkan enabled)
```
cqundefine, line 25: If vulkan is now enabled shouldn't the port depend on `vulkan-headers`?

ADKaster: Hmm. The gtk port? Likely it should. But in our vcpkg overlay mess I don't think it would help.

We already stub out vulkan-loader because I didn't want to bother with wrangling the way it wants to build both x11 and Wayland support.

I don't intend to upstream the enabling of Vulkan and Wayland. They'd likely reject it. The version update I will though.

PR #5926, opened 2025-08-20
LibWeb: Propagate animated values in `recompute_inherited_style`

Calme1709:
As `recompute_inherited_style` works in-place rather than building ComputedProperties from scratch we need to keep track of which animated properties are inherited to know whether we should remove them when we have no more inherited value.


Libraries/LibWeb/CSS/ComputedProperties.h
```diff
@@ -270,6 +273,7 @@ class ComputedProperties final : public JS::Cell {
     Array<RefPtr<StyleValue const>, number_of_properties> m_property_values;
     Array<u8, ceil_div(number_of_properties, 8uz)> m_property_important {};
     Array<u8, ceil_div(number_of_properties, 8uz)> m_property_inherited {};
+    Array<u8, ceil_div(number_of_properties, 8uz)> m_animated_property_inherited {};
```
AtkinsSJ, line 276: Looks like we could do with an inline-storage version of `AK::Bitmap`.

AtkinsSJ (approved):

PR #5927, opened 2025-08-20
LibWeb: Remove compute_defaulted_{values|property_value} 

Calme1709:
The majority of the logic in these functions was redundant with the only actual functionality being:
  - Resolving `initial` and `inherit` when we originally had `unset` when called from `compute_properties`.
  - Resolving inheritance for `text-align: -libweb-inherit-or-center` when called from `compute_properties`
  - Resolving inheritance for `color: currentcolor` when called from `compute_properties`
  - Setting all values to `property_initial_value` when called from `create_document_style`

There are more natural places to handle each of these so let's do it there and remove these functions.

Only functionality change is that we now correctly handle inheritance of animated color values for `color: currentcolor` (atleast once #5926 is merged)


Calme1709:
Rebased to resolve conflict

PR #5929, opened 2025-08-20
CI: Fixate js-benchmarks to current master

gmta:
The output format for js-benchmarks is going to change, and while the webhook parsing the results has not yet been updated fixate the js-benchmark checkout to a specific commit.


alimpfard (approved):

rgret-dev:
Nit: Fixate -> Pin

PR #5931, opened 2025-08-20
LibWeb: Emit XMLHttpRequest timeout event when the request times out

joanvilarrasa:
When an XMLHttpRequest times out, we now call handle_errors() to fire the timeout event. This is to prevent abort from firing if it gets called after the request has timed out. This change fixes 2 WPT tests that are also now imported into the tests folder.

### DIFF
--- a/Libraries/LibWeb/XHR/XMLHttpRequest.cpp
+++ b/Libraries/LibWeb/XHR/XMLHttpRequest.cpp
@@ -885,6 +885,8 @@ WebIDL::ExceptionOr<void> XMLHttpRequest::send(Optional<DocumentOrXMLHttpRequest
                 if (!request->done()) {
                     m_timed_out = true;
                     m_fetch_controller->terminate();
+
+                    handle_errors().release_value_but_fixme_should_propagate_errors();
                 }
             });
 
--- a/Tests/LibWeb/Text/expected/wpt-import/xhr/abort-after-stop.window.txt
+++ b/Tests/LibWeb/Text/expected/wpt-import/xhr/abort-after-stop.window.txt
@@ -1,0 +1,6 @@
+Harness status: OK
+
+Found 1 tests
+
+1 Fail
+Fail	XMLHttpRequest: abort event should fire when stop() method is used
--- a/Tests/LibWeb/Text/input/wpt-import/xhr/abort-after-timeout.any.js
+++ b/Tests/LibWeb/Text/input/wpt-import/xhr/abort-after-timeout.any.js
@@ -1,0 +1,43 @@
+// META: title=XMLHttpRequest: abort() after a timeout should not fire "abort" event
+
+    var test = async_test();
+
+    test.step(function() {
+        // timeout is 100ms
+        // the download would otherwise take 1000ms
+        // we check after 300ms to make sure abort does not fire an "abort" event
+
+        var timeoutFired = false;
+
+        var client = new XMLHttpRequest();
+
+        assert_true('timeout' in client, 'xhr.timeout is not supported in this user agent');
+
+        client.timeout = 100;
+
+        test.step_timeout(() => {
+            assert_true(timeoutFired);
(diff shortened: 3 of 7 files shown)


Tests/LibWeb/Text/expected/wpt-import/xhr/abort-after-stop.window.txt
```diff
@@ -0,0 +1,6 @@
+Harness status: OK
+
+Found 1 tests
+
+1 Fail
+Fail	XMLHttpRequest: abort event should fire when stop() method is used
```
trflynn89: The commit description states:
> This change fixes 2 WPT tests that are also now imported into the tests folder.

But both of the imported tests are failing?

Tests/LibWeb/Text/input/wpt-import/xhr/abort-after-timeout.any.js
```diff
@@ -0,0 +1,44 @@
...
+        };
+
+        client.onabort = test.step_func(function () {
+            // this should not fire!
+
+            assert_unreached("abort() should not cause the abort event to fire");
+        });
+
+        // NOTE: I put the wpt live url here so that we do not have to run a local server.
+        client.open("GET", "https://wpt.live/common/blank.html?pipe=trickle(d1)", true);
```
joanvilarrasa: I changed this line of the test, just added the url of the wpt site so that we don't have to run a local server for the tests. Is this the right way to approach this?

joanvilarrasa:
Also the second test that passes is the the same but for the worker (https://wpt.live/xhr/abort-after-timeout.any.worker.html)
I could not get it to run locally for the same reason as the other one, it requires to call this /common/blank.html?pipe=trickle(d1) in the worker script.

Here is an image of the url above with ladybird passing the test:
2025-08-21-200158_hyprshot

Tests/LibWeb/Text/input/wpt-import/xhr/abort-after-timeout.any.js
kennethmyhra: I don't think we want to call out to external servers from our test harness. In any case we shouldn't modify the tests we pull in that will cause issues later when we update those from the WPT source.

Libraries/LibWeb/XHR/XMLHttpRequest.cpp
```diff
@@ -885,6 +885,8 @@ WebIDL::ExceptionOr<void> XMLHttpRequest::send(Optional<DocumentOrXMLHttpRequest
                 if (!request->done()) {
                     m_timed_out = true;
                     m_fetch_controller->terminate();
+
+                    handle_errors().release_value_but_fixme_should_propagate_errors();
```
kennethmyhra: I believe this is better to `MUST` we are in an asynchronous context here and our implementation will only throw in a synchronous context. Can also drop the extra line break.

```suggestion
                    MUST(handle_errors())
```

kennethmyhra (commented):
Added a few comments. Also, squash the fixup commits into the original commit.

joanvilarrasa:
Removed the tests so that they do not cause issues when updating them from the WPT.

kennethmyhra (approved):
Looks good to me. Please resolve the conversations that have been resolved

Libraries/LibWeb/XHR/XMLHttpRequest.cpp
```diff
@@ -876,6 +876,7 @@ WebIDL::ExceptionOr<void> XMLHttpRequest::send(Optional<DocumentOrXMLHttpRequest
                 if (!request->done()) {
                     m_timed_out = true;
                     m_fetch_controller->terminate();
+                    MUST(handle_errors())
```
tcl3: Missing semicolon.

joanvilarrasa:
Added the semicolon

PR #5880, opened 2025-08-17
LibWeb/CSS: Fix basic absolute positioning inside grid containers

Norbiros:
Now elements with position `absolute` properly resolve their position inside parent elements with `grid`. I also imported some WPT tests related to that topic.

Part 2 of resolving issues on https://hack4krak.pl

Before:
Screenshot From 2025-08-17 08-32-36

After:
image

PR #5936, opened 2025-08-21
LibWeb: Fix border radius rendering artifacts with extreme values

dan-v:
Clamp border radius coordinates to element bounds when extreme values cause rendering artifacts. Previously, very large border-radius values could cause vertical lines to extend beyond element boundaries.

> Before
Screenshot 2025-08-21 at 2 17 54 AM

> After
Screenshot 2025-08-21 at 2 18 22 AM

### DIFF
--- a/Libraries/LibWeb/Painting/BorderPainting.cpp
+++ b/Libraries/LibWeb/Painting/BorderPainting.cpp
@@ -426,7 +426,10 @@ void paint_border(DisplayListRecorder& painter, BorderEdge edge, DevicePixelRect
             points.append(Gfx::FloatPoint(rect.top_left().to_type<int>()));
         } else {
             Gfx::FloatPoint inner_right_angle_offset = Gfx::FloatPoint(0, joined_border_width.value() - radius.horizontal_radius);
-            points.append(Gfx::FloatPoint(rect.top_left().to_type<int>()) + inner_right_angle_offset);
+            auto top_left_point = Gfx::FloatPoint(rect.top_left().to_type<int>()) + inner_right_angle_offset;
+            // Clamp the final point to stay within the rect bounds to prevent rendering artifacts
+            top_left_point.set_y(AK::clamp(top_left_point.y(), rect.top().value(), rect.bottom().value()));
+            points.append(top_left_point);
         }
 
         if (opposite_joined_corner_has_inner_corner) {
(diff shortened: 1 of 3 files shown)


Libraries/LibWeb/Painting/BorderPainting.cpp
```diff
@@ -441,7 +444,10 @@ void paint_border(DisplayListRecorder& painter, BorderEdge edge, DevicePixelRect
             points.append(Gfx::FloatPoint(rect.bottom_left().to_type<int>()) + inner_corner);
         } else {
             Gfx::FloatPoint inner_right_angle_offset = Gfx::FloatPoint(0, opposite_joined_border_width.value() - opposite_radius.horizontal_radius);
-            points.append(Gfx::FloatPoint(rect.bottom_left().to_type<int>()) - inner_right_angle_offset);
+            auto bottom_left_point = Gfx::FloatPoint(rect.bottom_left().to_type<int>()) - inner_right_angle_offset;
+            // Clamp the final point to stay within the rect bounds to prevent rendering artifacts
+            bottom_left_point.set_y(AK::clamp(bottom_left_point.y(), rect.top().value(), rect.bottom().value()));
+            points.append(bottom_left_point);
```
gmta, line 450: Instead of setting up a separate variable, you could just do `.constrained(rect)` to get the same result.

Libraries/LibWeb/Painting/BorderPainting.cpp
```diff
@@ -426,7 +426,10 @@ void paint_border(DisplayListRecorder& painter, BorderEdge edge, DevicePixelRect
             points.append(Gfx::FloatPoint(rect.top_left().to_type<int>()));
         } else {
             Gfx::FloatPoint inner_right_angle_offset = Gfx::FloatPoint(0, joined_border_width.value() - radius.horizontal_radius);
-            points.append(Gfx::FloatPoint(rect.top_left().to_type<int>()) + inner_right_angle_offset);
+            auto top_left_point = Gfx::FloatPoint(rect.top_left().to_type<int>()) + inner_right_angle_offset;
+            // Clamp the final point to stay within the rect bounds to prevent rendering artifacts
+            top_left_point.set_y(AK::clamp(top_left_point.y(), rect.top().value(), rect.bottom().value()));
+            points.append(top_left_point);
```
gmta, line 432: Why is all this just happening for `BorderEdge::Right`? Does the issue not exist for other borders?

Libraries/LibWeb/Painting/BorderPainting.cpp
gmta, line 1: Question about this general approach: it seems a bit late to be fixing up excessive border radius values during painting. Is this not something that should have been caught by #5880, for example?

PR #5937, opened 2025-08-21
LibWeb/CSS: Implement CSSUnitValue

AtkinsSJ:
This covers dimensions such as `<length>` and `<angle>`, as well as `<percentage>` and, for some reason, `<number>`.

This includes 63 different `CSS.unit(number)` helper methods for creating them in JS, and also means a bunch of the tests finally run, so there's a big test diff even though a lot of the "new" tests are failing.

The `CSSNumericValue` base type is mostly just stubbed out for now. Most of its methods require `CSSMathValue` to also be implemented, so I'm going to wait until after that's done to flesh these out.

As before, the typed OM API is still read-only; you can't use these to set any style yet.


konradekk:
(begs for generators for first few files on the list?)

AtkinsSJ:
Good point... And I missed this in the spec before:
> If an implementation supports additional CSS units that do not have a corresponding method in the above list, but that do correspond to one of the existing [CSSNumericType](https://drafts.css-houdini.org/css-typed-om-1/#dictdef-cssnumerictype) values, it must additionally support such a method, named after the unit in its defined canonical casing, using the generic behavior defined above.

So really, it should be a 1-1 match for units we support. I'd like to leave that as a future change though.

PR #5938, opened 2025-08-21
LibCore: Add AVIF to registered types in MimeData

zacoons:
This allows you to open local avif files using the `file://` schema.

### DIFF
--- a/Libraries/LibCore/MimeData.cpp
+++ b/Libraries/LibCore/MimeData.cpp
@@ -114,6 +114,7 @@ static Array const s_registered_mime_type = {
     MimeType { .name = "font/woff"sv, .common_extensions = { "woff"sv }, .description = "WOFF font"sv, .magic_bytes = Vector<u8> { 'W', 'O', 'F', 'F' } },
     MimeType { .name = "font/woff2"sv, .common_extensions = { "woff2"sv }, .description = "WOFF2 font"sv, .magic_bytes = Vector<u8> { 'W', 'O', 'F', '2' } },
 
+    MimeType { .name = "image/avif"sv, .common_extensions = { ".avif"sv }, .description = "AVIF image data"sv },
     MimeType { .name = "image/bmp"sv, .common_extensions = { ".bmp"sv }, .description = "BMP image data"sv, .magic_bytes = Vector<u8> { 'B', 'M' } },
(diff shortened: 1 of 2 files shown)


Libraries/LibCore/MimeData.cpp
```diff
@@ -114,6 +114,7 @@ static Array const s_registered_mime_type = {
     MimeType { .name = "font/woff"sv, .common_extensions = { "woff"sv }, .description = "WOFF font"sv, .magic_bytes = Vector<u8> { 'W', 'O', 'F', 'F' } },
     MimeType { .name = "font/woff2"sv, .common_extensions = { "woff2"sv }, .description = "WOFF2 font"sv, .magic_bytes = Vector<u8> { 'W', 'O', 'F', '2' } },
 
+    MimeType { .name = "image/avif"sv, .common_extensions = { ".avif"sv }, .description = "AVIF image data"sv },
```
trflynn89, line 117: Should we have magic bytes here? CC @Zaggy1024 

Please also add a test case to Tests/LibCore/TestLibCoreMimeType.cpp

zacoons: We discussed this on Discord and there are no consistent magic bytes for AVIF. This is also indicated by the MIME type assignment for AVIF, which says `Magic number(s): none` https://www.iana.org/assignments/media-types/image/avif

I've added tests for AVIF and some other common image types to `TestLibCoreMimeType.cpp`.

Zaggy1024: Yeah, unfortunately AVIF can have any number of variations in its first box, and it's not required to announce AVIF nor AVIS as its major brand, so there's not really any one or even two magic bytes we can use.

I've never really looked at the mime sniffing code, so not sure the best way to make an equivalent to what is done with these magic bytes, but the algorithm for detecting AVIF would be something like:

1. Read `u32 size` of the ftyp box
2. Skip `ftyp`
3. If `u32 major_brand` is `avif` or `avis`, return true
4. Skip `u32 minor_version`
5. If `u32 minor_brand` is `avif` or `avis`, return true
6. Repeat step 5 until the end of the next read is larger than `size`

Zaggy1024 (approved):

PR #5941, opened 2025-08-21
LibWeb: Clamp negative values for `padding-*` and `font-size`

Calme1709:
Interpolation can leave these values negative which was causing us to end up with negative paintable sizes which in turn caused crashes in `normalize_border_radii_data`.

Fixes 6 crashes in imported tests and gains us 45 WPT passes.


Libraries/LibWeb/CSS/ComputedProperties.cpp
```diff
@@ -211,37 +211,53 @@ Size ComputedProperties::size_value(PropertyID id) const
...
 
-    if (value.is_length())
-        return value.as_length().length();
+        // FIXME: This value can be negative as interpolation does not yet clamp values to allowed ranges - remove this
+        //        once we do that.
+        if (disallow_negative_lengths == DisallowNegativeLengths::Yes && percentage.as_fraction() < 0)
+            return {};
+
+        return percentage;
+    }
```
AtkinsSJ: Hmm... I'm not sure about clamping percentages. See the note on step 4 of https://drafts.csswg.org/css-values-4/#simplify-a-calculation-tree :

> If a percentage is left at this point, it will usually block simplification of the node, since it needs to be resolved against another value using information not currently available. (Otherwise, it would have been converted to a different value in an earlier step.) This includes operations such as "min", since percentages might resolve against a negative basis, and thus end up with an opposite comparative relationship than the raw percentage value would seem to indicate.

Now, I don't know if there are any real situations where a `<length-percentage>` resolves percentages against a possibly-negative length... plus this is a temporary hack. So I'm not objecting to merging this, just something to be aware of.

Libraries/LibWeb/CSS/ComputedProperties.h
```diff
@@ -77,9 +77,13 @@ class ComputedProperties final : public JS::Cell {
 
     Size size_value(PropertyID) const;
     [[nodiscard]] Variant<LengthPercentage, NormalGap> gap_value(PropertyID) const;
-    LengthPercentage length_percentage_or_fallback(PropertyID, LengthPercentage const& fallback) const;
-    Optional<LengthPercentage> length_percentage(PropertyID) const;
-    LengthBox length_box(PropertyID left_id, PropertyID top_id, PropertyID right_id, PropertyID bottom_id, Length const& default_value) const;
+    enum class DisallowNegativeLengths {
+        No,
+        Yes,
+    };
```
AtkinsSJ: Nit: `DisallowNLs::No` being a double-negative requires a bit more mental gymnastics to understand than `AllowNLs::Yes`. (Or `ClampNLs::No`)

Libraries/LibWeb/CSS/ComputedProperties.cpp
Calme1709: I'm pretty sure this is fine as:
1. [The valid value for the padding-* properties is defined as <length-percentage [0, ∞]>](https://drafts.csswg.org/css-box-4/#propdef-padding-top). (Although now that I think about it I am not 100% sure on whether that means percentages of the literal value [0, ∞] or that resolve to a length in that range)
2. I don't think there are any situations where inline-size will be negative, although perhaps someone with more knowledge around layout could provide insight there.

I am actively working towards changes which will include clamping of interpolated values which will mean we can remove this (although that has been in the works for the past month or so with no ETA yet).

Libraries/LibWeb/CSS/ComputedProperties.h
Calme1709: Agreed - changed to `ClampNegativeLengths`

Calme1709:
Rebased to resolve conflicts and added a commit https://github.com/LadybirdBrowser/ladybird/pull/5941/commits/4ecacdf05248107ba3cb38642bce0cc59b02692e to resolve crash in one of the imported tests which was introduced in https://github.com/LadybirdBrowser/ladybird/commit/1a3da3d825d5644196c55e0edb2e28770064d0dc

Calme1709:
Rebased to resolve conflicts

AtkinsSJ (approved):

Calme1709:
I believe the CI failure here was a flake - are you waiting on anything else from me @AtkinsSJ?

AtkinsSJ:
Thanks, I've re-run that job. Hopefully it'll pass this time.

PR #5943, opened 2025-08-21
LibWeb/XHR: Minor cleanup in XMLHttpRequest

kennethmyhra:
Mostly removing outdated FIXME comments, and adding one missing spec comment.

Also, MUSTing `handle_{errors,response_end_of_body}` since these are not supposed to throw in an async context, and we no longer want to handle minor OOMs.

### DIFF
--- a/Libraries/LibWeb/XHR/XMLHttpRequest.cpp
+++ b/Libraries/LibWeb/XHR/XMLHttpRequest.cpp
@@ -808,8 +808,6 @@ WebIDL::ExceptionOr<void> XMLHttpRequest::send(Optional<DocumentOrXMLHttpRequest
             if (!length.has<u64>())
                 length = 0;
 
-            // FIXME: We can't implement these steps yet, as we don't fully implement the Streams standard.
-
             // 10. Let processBodyChunk given bytes be these steps:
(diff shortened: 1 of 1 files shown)


Libraries/LibWeb/XHR/XMLHttpRequest.cpp
```diff
@@ -779,8 +779,7 @@ WebIDL::ExceptionOr<void> XMLHttpRequest::send(Optional<DocumentOrXMLHttpRequest
 
             // 2. Handle errors for this.
             // NOTE: This cannot throw, as `handle_errors` only throws in a synchronous context.
-            // FIXME: However, we can receive allocation failures, but we can't propagate them anywhere currently.
-            handle_errors().release_value_but_fixme_should_propagate_errors();
+            MUST(handle_errors());
```
kennethmyhra, line 781: I believe it's correct to `MUST` these, the only place this can throw is in `request_errors_steps()`, and then only when we are in a synchronous context, same goes for `handle_response_end_of_body()`

https://github.com/LadybirdBrowser/ladybird/blob/cb1a1a5cb5838489472c0135c17b57c1ef3c2b2c/Libraries/LibWeb/XHR/XMLHttpRequest.cpp#L1251-L1254

PR #5946, opened 2025-08-21
LibWeb+WebContent: A couple of cookie-store related fixes

IdanHo:
This fixes 4 WPT tests.


Libraries/LibWeb/CookieStore/CookieChangeEvent.cpp
```diff
@@ -37,4 +37,12 @@ void CookieChangeEvent::initialize(JS::Realm& realm)
     Base::initialize(realm);
 }
 
+void CookieChangeEvent::visit_edges(Cell::Visitor& visitor)
+{
+    Base::visit_edges(visitor);
+
+    visitor.visit(m_cached_changed);
+    visitor.visit(m_cached_deleted);
```
trflynn89: Should we also add a `VISIT_CACHED_ATTRIBUTE` macro? Just seems a little off to have to know the exact name of the cached members here, but nowhere else. No strong feelings.

Libraries/LibWeb/Cookie/ParsedCookie.h
```diff
@@ -32,6 +32,8 @@ bool cookie_contains_invalid_control_character(StringView);
 bool domain_matches(StringView string, StringView domain_string);
 String default_path(URL::URL const&);
 
+static constexpr AK::Duration maximum_cookie_age = AK::Duration::from_seconds(400LL * 24 * 60 * 60);
```
trflynn89: `static` in a header doesn't do anything, generally should be `constexpr inline`

Libraries/LibWeb/CookieStore/CookieStore.cpp
```diff
@@ -455,8 +455,21 @@ static bool set_a_cookie(PageClient& client, URL::URL const& url, String name, S
...
+        // 3. Let cookie-age-limit be the maximum age of the cookie (which SHOULD be 400 days in the future or sooner, see
+        //    Section 5.5).
+        auto cookie_age_limit = UnixDateTime::now() + Cookie::maximum_cookie_age;
+
+        // 4. If the expiry-time is more than cookie-age-limit, the user agent MUST set the expiry time to cookie-age-limit
+        //    in seconds.
+        if (expiry_time.seconds_since_epoch() > cookie_age_limit.seconds_since_epoch())
+            expiry_time = cookie_age_limit;
+
+        parsed_cookie.expiry_time_from_expires_attribute = expiry_time;
```
trflynn89, line 471: Should there be an issue in the cookie store repo about this, or does that spec generally just defer to the RFC?

Libraries/LibWeb/CookieStore/CookieChangeEvent.cpp
IdanHo: Sure, added, it is a bit nicer

Libraries/LibWeb/Cookie/ParsedCookie.h
IdanHo: changed

Libraries/LibWeb/CookieStore/CookieStore.cpp
IdanHo: No, this is just technically part of the second-to-last step (step 23), since it says to run the algorithm from the RFC as if what we constructed so far was just received as a cookie.

trflynn89 (approved):

PR #5949, opened 2025-08-22
LibWeb: Fix pathological border-radius normalization artifacts

dan-v:
When extreme border-radius values are normalized, they can result in asymmetric values causing rendering artifacts where triangular slices appear instead of the expected border.

Fix by detecting these cases and forcing square corners as a fallback.

> before (current master)
Screenshot 2025-08-21 at 10 30 25 PM

> before with fix from #5936 
Screenshot 2025-08-21 at 10 32 24 PM

> after
Screenshot 2025-08-21 at 10 29 20 PM

### DIFF
--- a/Libraries/LibWeb/Painting/Paintable.cpp
+++ b/Libraries/LibWeb/Painting/Paintable.cpp
@@ -264,6 +264,16 @@ Painting::BorderRadiiData normalize_border_radii_data(Layout::Node const& node, 
         }
     }
 
+    // Sanity check: If normalization resulted in pathological asymmetric values,
+    // zero out the border radius to force square corners
+    for (auto* corner : { &radii_px.top_left, &radii_px.top_right, &radii_px.bottom_right, &radii_px.bottom_left }) {
+        bool has_extreme_asymmetry = (corner->horizontal_radius > rect.width() / 4 && corner->vertical_radius < 1) || (corner->vertical_radius > rect.height() / 4 && corner->horizontal_radius < 1);
+        if (has_extreme_asymmetry) {
+            corner->horizontal_radius = 0;
+            corner->vertical_radius = 0;
+        }
+    }
+
     return radii_px;
 }
 
--- a/Tests/LibWeb/Ref/expected/border-radius-asymmetric-normalization-ref.html
+++ b/Tests/LibWeb/Ref/expected/border-radius-asymmetric-normalization-ref.html
@@ -1,0 +1,23 @@
+<!DOCTYPE html>
+<html>
+<head>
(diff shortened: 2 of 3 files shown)


Libraries/LibWeb/Painting/Paintable.cpp
```diff
@@ -264,6 +264,16 @@ Painting::BorderRadiiData normalize_border_radii_data(Layout::Node const& node,
...
 
+    // Sanity check: If normalization resulted in pathological asymmetric values,
+    // zero out the border radius to force square corners
+    for (auto* corner : { &radii_px.top_left, &radii_px.top_right, &radii_px.bottom_right, &radii_px.bottom_left }) {
+        bool has_extreme_asymmetry = (corner->horizontal_radius > rect.width() / 4 && corner->vertical_radius < 1) || (corner->vertical_radius > rect.height() / 4 && corner->horizontal_radius < 1);
+        if (has_extreme_asymmetry) {
+            corner->horizontal_radius = 0;
+            corner->vertical_radius = 0;
+        }
+    }
```
gmta, line 275: This seems like a very subjective patch job instead of a sanity check. If I check the values of both iterations of correction happening above on `master`, I get:

```
f.numerator(): 164
f.denominator(): 33554431.984375
corner->horizontal_radius: 33554431.984375 -> 164
corner->vertical_radius: 342423 -> 1.671875
corner->horizontal_radius: 33554431.984375 -> 164
corner->vertical_radius: 342423 -> 1.671875
corner->horizontal_radius: 33554431.984375 -> 164
corner->vertical_radius: 342423 -> 1.671875
corner->horizontal_radius: 33554431.984375 -> 164
corner->vertical_radius: 342423 -> 1.671875
---
f.numerator(): 164
f.denominator(): 328
corner->horizontal_radius: 164 -> 82
corner->vertical_radius: 1.671875 -> 0.828125
corner->horizontal_radius: 164 -> 82
corner->vertical_radius: 1.671875 -> 0.828125
corner->horizontal_radius: 164 -> 82
corner->vertical_radius: 1.671875 -> 0.828125
corner->horizontal_radius: 164 -> 82
corner->vertical_radius: 1.671875 -> 0.828125
```

These values don't seem to be that esoteric, aren't we just displaying / rendering them wrong?

Also, looking at your test case in Firefox, the border radii are definitely not 0.

dan-v: That's fair! I couldn't think of a nice way to resolve this, just was a test case I saw that was rendering incorrectly. I'll close this one out.

PR #5953, opened 2025-08-22
LibWeb: Replace usages of `dynamic_cast` with `as` and `as_if`

tcl3:
This PR removes all usages of `dynamic_cast` from LibWeb, apart from those used to implement `as` and as_if`.


Libraries/LibWeb/ContentSecurityPolicy/Violation.cpp
```diff
@@ -47,7 +47,7 @@ GC::Ref<Violation> Violation::create_a_violation_object_for_global_policy_and_di
 
     // 3. If global is a Window object, set violation’s referrer to global’s document's referrer.
     if (global_object) {
-        if (auto* window = dynamic_cast<HTML::Window*>(global_object.ptr())) {
+        if (auto* window = as_if<HTML::Window>(*global_object)) {
```
gmta: You can remove the outer `if (global_object) {` if you keep using `global_object.ptr()` here

Libraries/LibWeb/DOM/EventTarget.cpp
```diff
@@ -423,12 +422,11 @@ WebIDL::CallbackType* EventTarget::get_current_value_of_event_handler(FlyString
...
         // 5. If element is not null and element has a form owner, let form owner be that form owner. Otherwise, let form owner be null.
         GC::Ptr<HTML::HTMLFormElement> form_owner;
-        if (is<HTML::FormAssociatedElement>(element.ptr())) {
-            auto* form_associated_element = dynamic_cast<HTML::FormAssociatedElement*>(element.ptr());
-            VERIFY(form_associated_element);
-
-            if (form_associated_element->form())
-                form_owner = form_associated_element->form();
+        if (element) {
+            if (auto* form_associated_element = as_if<HTML::FormAssociatedElement>(*element)) {
```
gmta: `element.ptr()` allows you to drop the outer `if (element) {`

Libraries/LibWeb/DOM/EventTarget.cpp
```diff
@@ -423,12 +422,11 @@ WebIDL::CallbackType* EventTarget::get_current_value_of_event_handler(FlyString
 
         // 5. If element is not null and element has a form owner, let form owner be that form owner. Otherwise, let form owner be null.
         GC::Ptr<HTML::HTMLFormElement> form_owner;
-        if (is<HTML::FormAssociatedElement>(element.ptr())) {
-            auto* form_associated_element = dynamic_cast<HTML::FormAssociatedElement*>(element.ptr());
-            VERIFY(form_associated_element);
-
-            if (form_associated_element->form())
-                form_owner = form_associated_element->form();
+        if (element) {
```
trflynn89: Do we need the outer `if`?
```c++
if (auto* form_associated_element = as_if<HTML::FormAssociatedElement>(element.ptr())) {
    if (auto* form = form_associated_element->form())
        form_owner = form;
}
```

trflynn89: (on that note, I've also been wanting to add `as` overloads for `GC::Ptr` and friends so we don't have to add `.ptr()` in places like this)

Libraries/LibWeb/Layout/ImageBox.cpp
```diff
@@ -67,8 +67,10 @@ void ImageBox::dom_node_did_update_alt_text(Badge<ImageProvider>)
 
 bool ImageBox::renders_as_alt_text() const
 {
-    if (auto const* image_provider = dynamic_cast<ImageProvider const*>(dom_node().ptr()))
-        return !image_provider->is_image_available();
+    if (auto dom_node = this->dom_node()) {
+        if (auto const* image_provider = as_if<ImageProvider>(*dom_node))
```
gmta: Also here, `dom_node().ptr()` so you can drop the outer `if`

Libraries/LibWeb/Layout/ImageBox.cpp
```diff
@@ -67,8 +67,10 @@ void ImageBox::dom_node_did_update_alt_text(Badge<ImageProvider>)
...
+            return !image_provider->is_image_available();
+    }
```
trflynn89: Same here I think:
```c++
if (auto const* image_provider = as_if<ImageProvider>(dom_node().ptr()))
    return !image_provider->is_image_available();
```

tcl3:
My latest push uses `as<Bar>(foo.ptr())` where possible. I've also updated the `as` and `is` implementation to ignore CV references when comparing types. This allows us to avoid dynamic casts more often and allows us to detect more possible bogus casts.

Libraries/LibWeb/DOM/EventTarget.cpp
tcl3: Versions of `as`, `as_if` and `is` that take `GC::Ref<T>` and `GC::Ptr<T>` and return GC wrapped types would be nice. I'd rather do this as a separate PR my initial regex grepping shows a fair number of places would want changing to use these new functions.

AK/TypeCasts.h
```diff
@@ -40,7 +40,7 @@ ALWAYS_INLINE CopyConst<InputType, OutputType>* as_if(InputType& input)
 {
     if (!is<OutputType>(input))
         return nullptr;
-    if constexpr (IsBaseOf<InputType, OutputType>) {
+    if constexpr (IsBaseOf<RemoveCVReference<InputType>, RemoveCVReference<OutputType>>) {
```
gmta: I believe `__is_base_of` [ignores](<https://gcc.gnu.org/onlinedocs/gcc/Type-Traits.html#index-_005f_005fis_005fbase_005fof>) top-level CV qualifications?

Libraries/LibWeb/DOM/EventTarget.cpp
trflynn89: Oh for sure, I wasn't trying to suggest we do that here :)

AK/TypeCasts.h
tcl3: So it is. Looks like this change is unnecessary then!

gmta: `RemoveReference<T>` might still be useful though?

tcl3: After going to the trouble of checking how often we were doing a `static_cast` vs `dynamic_cast`, it turns out that we were missing a lot of opportunities to do static casts.

My latest push replaces this with a `requires` clause that guarantees a static cast is used when possible.

gmta (approved):

PR #5955, opened 2025-08-22
LibWeb/CSS: Implement typed OM math functions

AtkinsSJ:
We're starting to get somewhere now!

Still a lot of CSSNumericValue stuff left to do, but this lets us reify calculations - at least, the ones that the typed-om spec allows. It's missing quite a few of the math functions that have been added later.

A decent chunk of the diff is tests that now actually run instead of crashing, but don't pass at all. Also, boy is there a lot of boilerplate with all these IDL types. `*_*`

Again, more is in the works.

### DIFF
--- a/Libraries/LibWeb/CSS/CSSMathClamp.cpp
+++ b/Libraries/LibWeb/CSS/CSSMathClamp.cpp
@@ -102,4 +102,18 @@ GC::Ref<CSSNumericValue> CSSMathClamp::upper() const
     return m_upper;
 }
 
+// https://drafts.css-houdini.org/css-typed-om-1/#equal-numeric-value
+bool CSSMathClamp::is_equal_numeric_value(GC::Ref<CSSNumericValue> other) const
+{
+    // AD-HOC: Spec doesn't handle clamp()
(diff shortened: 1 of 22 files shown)


Libraries/LibGC/Ptr.h
```diff
@@ -183,6 +183,14 @@ class Ptr {
 
     operator T*() const { return m_ptr; }
 
+    Ref<T> release_nonnull()
```
trflynn89: I don't know that a "release" API really makes sense here. It's used in `OwnPtr` and `RefPtr` due to ownership mechanics, but that doesn't apply here - `GC::Ptr` / `GC::Ref` aren't owning objects.

So maybe just:
```c++
Ref<T> as_nonnull()
{
    ASSERT(m_ptr);
    return *m_ptr;
}
```

AtkinsSJ: Agreed, it didn't make much sense. Fixed!

AtkinsSJ:
First push: Change `GC::Ptr::release_nonnull()` to `::as_nonnull()`. Second push: rebase on master.

PR #5952, opened 2025-08-22
LibWeb: Editing host, selection & focus improvements

gmta:
* Nested editing hosts now propagate focus to the nearest uninterrupted ancestor editing host
* Selection updates on uneditable contents using keyboard navigation are now only allowed if the Shift key is held
* Focusing spec steps were updated; unfocusable elements can now no longer become the focused area

PR #5956, opened 2025-08-22
LibWeb: Implement up/down arrow navigation for EditingHostManager

trflynn89:
Works just the same as input element navigation


Libraries/LibWeb/Selection/Selection.cpp
```diff
@@ -653,4 +654,40 @@ void Selection::move_offset_to_previous_word(bool collapse_selection)
     }
 }
 
+void Selection::move_offset_to_next_line(bool collapse_selection)
+{
+    auto* text_node = as_if<DOM::Text>(anchor_node().ptr());
```
gmta, line 656: I think this should be `focus_node()` in both methods; I've got some relevant changes for that in #5952

gmta: Oh whoops, I don't have that change in there anymore - regardless, I think the relevant part is that both the anchor and focus nodes are supposed to be in text nodes here, but we're moving from the focus node instead of the anchor node.

trflynn89: To clarify, is there a change to be made here then? :sweat_smile:  Looking at your PR, the adjacent cursor APIs are using the anchor node.

gmta: Well logically, I should update those to use `focus_node()` as well, just remarking that it's something you might want to use here. If a selection over multiple DOM::Text nodes works as you expect it to, then you don't necessarily need to update it :P

trflynn89: Forgot to get back to this - indeed, both the existing and new methods added here don't handle moving across DOM nodes within a contenteditable node. I've added a FIXME for now.

PR #5969, opened 2025-08-23
Add "Default Zoom Level" setting

rgret-dev:
I was slightly annoyed having to re-zoom each time I opened the browser :^)

image

### DIFF
--- a/Tests/LibWeb/test-web/main.cpp
+++ b/Tests/LibWeb/test-web/main.cpp
@@ -644,6 +644,8 @@ static ErrorOr<int> run_tests(Core::AnonymousBuffer const& theme, Web::DevicePix
     for (size_t i = 0; i < concurrency; ++i) {
         auto view = TestWebView::create(theme, window_size);
         view->on_load_finish = [&](auto const&) { ++loaded_web_views; };
+        // FIXME: Figure out a better way to ensure that tests use default browser settings.
+        view->reset_zoom();
 
         views.unchecked_append(move(view));
     }


Libraries/LibWebView/Settings.h
```diff
@@ -95,6 +99,7 @@ class Settings {
     ByteString m_settings_path;
 
     URL::URL m_new_tab_page_url;
+    double m_default_zoom_level_factor;
```
trflynn89: We prefer to always initialize plain in the member list:
```c++
double m_default_zoom_level_factor { 0 };
```
even if the value is set in the constructor.

Base/res/ladybird/about-pages/settings/default-zoom-level.js
```diff
@@ -0,0 +1,71 @@
+const defaultZoomLevelDropdown = document.querySelector("#default-zoom-level");
+
+const zoomLevelFactors = [
+    0.3,
+    1 / 3.0,
```
trflynn89: Having 30% and 33% feels a bit fine-grained
zoom

Tests/LibWeb/test-web/main.cpp
```diff
@@ -644,6 +644,7 @@ static ErrorOr<int> run_tests(Core::AnonymousBuffer const& theme, Web::DevicePix
     for (size_t i = 0; i < concurrency; ++i) {
         auto view = TestWebView::create(theme, window_size);
         view->on_load_finish = [&](auto const&) { ++loaded_web_views; };
+        view->reset_zoom();
```
trflynn89, line 648: Let's add a FIXME here. This is fine for now, but I think we want to more generally have a way to ensure test-web runs with default settings. We don't want to special-case every setting that affects tests here.

Base/res/ladybird/about-pages/settings/default-zoom-level.js
rgret-dev: I've kept 33% and removed the 30% option

Issue #5959, opened 2025-08-22
GC clang plugin: Check for missing visit on Variant<...>


awesomekling:
If a cell-derived object has a member `Variant<T1, T2...>` where any of `T1`, `T2`, etc are cell-derived, we should warn if the member is not visited.

PR #5970, opened 2025-08-23
ClangPlugins: Fix a couple of edge cases

### LINKED ISSUE #5958: GC clang plugin: Check for missing visit on Optional<T>
If a cell-derived object has a member `Optional<T>` where `T` is cell-derived, we should warn if the member is not visited.

IdanHo:
This resolves #5958 and #5959.
We were already checking for missing vists on templated member fields (like Optional<T> and Variant<...>), but there were 2 separate issues that were preventing the plugin from detecting the missing visit in each of these issues.

### DIFF
--- a/Tests/ClangPlugins/LibJSGCTests/missing_member_in_visit_edges.cpp
+++ b/Tests/ClangPlugins/LibJSGCTests/missing_member_in_visit_edges.cpp
@@ -6,6 +6,8 @@
 
 // RUN: %clang++ -Xclang -verify %plugin_opts% -c %s -o %t 2>&1
 
+#include <AK/Optional.h>
+#include <AK/Variant.h>
 #include <LibJS/Runtime/Object.h>
 
 class ForwardDeclaredObject;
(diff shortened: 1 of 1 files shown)


ADKaster:
Given this error from the plugin job
```
/home/runner/_work/ladybird/ladybird/Libraries/LibJS/Runtime/Realm.h:90:5: error: GC-allocated member is not visited in Realm::visit_edges
   90 |     AK::Array<GC::Ptr<NativeFunction>, to_underlying(Bytecode::Builtin::__Count)> m_builtins;
      |     ^
1 error generated.
```

A commit to fix transgressions of these newly caught rules before adding the rules seems necessary.

Separately, if these changes to handle forward declared types and such do fix the referenced issues, a test or two to specifically validate Optional and Variant with GC-allocated-T inside would be nice to have.

IdanHo:
> A commit to fix transgressions of these newly caught rules before adding the rules seems necessary.

Of course, I did a full `./Meta/ladybird.py rebuild` and assumed it found no issues, but looks like the ccache just prevented the rebuild of most files :disappointed: 

> Separately, if these changes to handle forward declared types and such do fix the referenced issues, a test or two to specifically validate Optional and Variant with GC-allocated-T inside would be nice to have.

Sure, I'll add them separately though, since these commits do not in fact add support for validating Optional and Variant, as they were supported already.

PR #5979, opened 2025-08-24
Devcontainer+Documentation: Add Devcontainer documentation and simplify use

JanKoudijs:
After building Ladybird using the Devcontainer I had to head over to Discord history to figure out how to do the actual running and debugging.

This pull request adds Devcontainer documentation and makes some container modifications to reduce manual setup, e.g. by setting up environment variables.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -33,6 +33,11 @@
             "label": "Web VNC"
         }
     },
+    "remoteEnv": {
+        "CC": "clang-20",
+        "CXX": "clang++-20",
+        "XDG_RUNTIME_DIR": "/tmp"
+    },
     // Use 'postCreateCommand' to run commands after the container is created.
     "postCreateCommand": "pre-commit install; pre-commit install --hook-type commit-msg",
     // Configure tool-specific properties.
@@ -41,7 +46,8 @@
             "extensions": [
                 "ms-vscode.cmake-tools",
                 "llvm-vs-code-extensions.vscode-clangd",
-                "eamodio.gitlens"
+                "eamodio.gitlens",
+                "vadimcn.vscode-lldb"
             ],
             "settings": {
                 // Excluding the generated directories keeps your file view clean and speeds up search.
@@ -70,7 +76,7 @@
                 "git.inputValidationSubjectLength": 72,
                 // If clangd was obtained from a package manager, its path can be set here.
                 // Note: This has to be adjusted manually, to the "llvm_version" from above
-                "clangd.path": "clangd-19",
+                "clangd.path": "clangd-20",
                 "clangd.arguments": [
                     "--header-insertion=never" // See https://github.com/clangd/clangd/issues/1247
                 ]
--- a/Documentation/EditorConfiguration/VSCodeDevContainer.md
+++ b/Documentation/EditorConfiguration/VSCodeDevContainer.md
@@ -1,0 +1,17 @@
+# Visual Studio Code Dev Container
+
+Visual Studio Code has support for containerized development environments, the Ladybird code base has such a Dev Container included to provided a consistent and easily reproducible development environment.
+
+Using the Dev Container requires the installation of the [Dev Containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers) extension.
+
+After installation the extension will detect the Ladybird Dev Container and provide the option to build it. Start the container build and wait for it to complete, the building process takes quite some time, view the log to monitor progress.
+
+The Dev Container already includes the recommended extensions and settings, as laid out in the [Visual Studio Code Project Configuration](VSCodeConfiguration.md), but it is still a good read for further details.
+
+## Usage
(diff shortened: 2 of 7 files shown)


.devcontainer/devcontainer.json
```diff
@@ -70,7 +76,7 @@
                 "git.inputValidationSubjectLength": 72,
                 // If clangd was obtained from a package manager, its path can be set here.
                 // Note: This has to be adjusted manually, to the "llvm_version" from above
-                "clangd.path": "clangd-19",
+                "clangd.path": "clangd-20",
                 "clangd.arguments": [
                     "--header-insertion=never" // See https://github.com/clangd/clangd/issues/1247
```
ADKaster, line 76: oh hey. with clangd-20, we should be able to add config for system header insertion now that this issue is closed. I'll make an issue. (i.e. no change requested here).

.devcontainer/devcontainer.json
```diff
@@ -33,6 +33,11 @@
             "label": "Web VNC"
         }
     },
+    "remoteEnv": {
+        "CC": "clang-20",
+        "CXX": "clang++-20",
```
ADKaster: should this be added in the ladybird feature that determines the llvm_version installed instead of here at the top level?

Documentation/README.md
```diff
@@ -20,6 +20,7 @@ you are welcome to ask on [Discord](../README.md#get-in-touch-and-participate).
 * [Qt Creator](EditorConfiguration/QtCreatorConfiguration.md)
 * [Vim](EditorConfiguration/VimConfiguration.md)
 * [VS Code](EditorConfiguration/VSCodeConfiguration.md)
+* [VS Code Dev Container](EditorConfiguration/VSCodeDevContainer.md)
```
ADKaster, line 23: hm. Technically, the devcontainer is not vscode specific. several IDEs have support. Though we do have vscode-specific config in there for it to use.

Documentation/EditorConfiguration/VSCodeDevContainer.md
```diff
@@ -0,0 +1,17 @@
...
+
+After installation the extension will detect the Ladybird Dev Container and provide the option to build it. Start the container build and wait for it to complete, the building process takes quite some time, view the log to monitor progress.
+
+The Dev Container already includes the recommended extensions and settings, as laid out in the [Visual Studio Code Project Configuration](VSCodeConfiguration.md), but it is still a good read for further details.
+
+## Usage
+Run `./Meta/ladybird.py run ladybird` at least once to kick off downloading and building vcpkg dependencies.
+
+On execution, the Ladybird processes run in the container without a browser window. It can be interacted with by connecting to the container using VNC. This can be done from a browser by accessing http://localhost:6080, or from a VNC application by accessing `localhost:5901`. Apps to use are [noVNC](https://novnc.com) in the former case, and [TigerVNC](https://tigervnc.org) in the latter case, or any alternative already available on the system.
```
ADKaster: I've been wondering about this bit. On my ubuntu system, VsCode automagically hooks up a wayland socket, and running Ladybird directly from a terminal inside the devcontainer (ubuntu or fedora) "just works" to open windows on my host system.

Does it truly not work on yours?

I also tried running the linux containers from macOS and ran into problem after problem, figuring it wasn't worth the hassle.

ADKaster:
Also I haven't checked in a while, but does the "vcpkg cache" feature properly use clang-20? or does launching the container and running ./Meta/ladybird.py run do a full vcpkg rebuild? I remember looking at that in the past but don't remember if I fixed it.

.devcontainer/devcontainer.json
ADKaster: https://github.com/LadybirdBrowser/ladybird/issues/5980

.devcontainer/devcontainer.json
JanKoudijs: Environment variables can indeed be moved to `devcontainer-feature.json`.

Documentation/README.md
JanKoudijs: Yeah, both settings and extensions are vscode-specific. I guess only vscode currently gives the ready-to-use container experience. Should we leave this one unchanged till the developer that uses the container in another editor shows up?

Documentation/EditorConfiguration/VSCodeDevContainer.md
JanKoudijs: I'm using the devcontainer (or Flatpak) to overcome the "too old for Ladybird" tools on my dev setup, it also runs X11 (yes, I should upgrade). I can  change the wording that if the Wayland pass-through doesn't work that there is the VNC fallback.

JanKoudijs:
> Also I haven't checked in a while, but does the "vcpkg cache" feature properly use clang-20? or does launching the container and running ./Meta/ladybird.py run do a full vcpkg rebuild? I remember looking at that in the past but don't remember if I fixed it.

For me it always has been the case that ./Meta/ladybird.py triggers a full vcpkg rebuild (only recently started using the container). I wasn't aware of the cache in the container. Btw, the move to clang-20 is not part of this PR, that has been done previously.

Issue #38, opened 2024-06-04
Windows support


elcuco:
This is a meta issue to track all the issues needed for a Windows support.

Goals:

1.  We aim to support clang-cl. This is the first priority.
2. Later on, cl, from 2022. This is a very low priority.
3. If we manage to get some mingw toolchain to compile the program - I am happy. But this is not a goal

Problems:

- [ ] Some compiler flags are not compatible with clang-cl. 
- [ ] We need to pull code from previous attempts of this kind of work. 

Please do not comment on this issue unless you have something to contribute. This issue is one of the boss-issues - it will take time. Problems and goals will change, nothing is written in stone, provide code if you can.

asutoshvariar:
Could std::chrono be used as a replacement for the POSIX-only sys/time.h, as it's present in the C++ 20 standard?

ADKaster:
We have a patch hiding in one of those mentioned branches to make AK::Time and AK::Duration work well on windows, so that's not an issue at the moment.

seanmiddleditch:
> [[no_unique_address]] not supported (am I using a wrong compiler? I use MSVC 2022 and its clang-cl)

A separate `msvc::no_unique_address` attribute must currently be used instead for ABI reasons: https://devblogs.microsoft.com/cppblog/msvc-cpp20-and-the-std-cpp20-switch/#msvc-extensions-and-abi

This is true in clang-cl too, using a recent-enough version: https://github.com/llvm/llvm-project/pull/65675

R-Goc:
Why find an alternative to fontconfig? It works on windows and even compiles with clang-cl. It needs to have config files shipped for it as windows doesn't provide any. It is used in GIMP for example.

niutech:
Have anyone tried to build Ladybird in Cygwin? It provides some POSIX compliance for Windows.

ADKaster:
As stated in the first post of the issue, Cygwin and Mingw are not a goal. Native windows with clang-cl and wsl2 with wslg are the only two configurations we will support in this repository.

R-Goc:
Maybe a github discussions tab, or other way of collaboration could be opened to avoid spamming this tracking issue. This will need more work. (Also how to get cmake to find ICU data?)

bojantorbica:
[...]

PR #5981, opened 2025-08-24
CMake+LibWebView+LibRequests+RequestServer+WebContent: Add ability to launch Ladybird and load web pages on Windows

ayeteadoe:
This is the work that allows us to launch Ladybird with the Qt UI framework on Windows and load web pages.

Main things that still need to be addressed:

- [ ] Fix WebContent process from crashing on IPC shutdown requests. When in `Web::HTML::Navigable::navigate()`, before requesting and loading the web page, the query to `PageClient::is_url_suitable_for_same_process_navigation()` instructs us to shutdown the current WebContent process and load the URL on a new process. The IPC call eventually invokes `ViewImplementation::create_new_process_for_cross_site_navigation()`, but we crash in `PosixSocketHelper::close()` for which means we never gracefully shutdown. This causes us to never actually send the request (still need to investigate thje root cause, but my breakpoints don't get hit in any RequestServer code when the behaviour is active).
- [ ] Fix HTTP(S) Request instability. For some reason periodically (~70% of the time), HTTP(S) requests seem to fail in such a way that crashes the Browser UI process. I haven't investigation in much depth yet, but seems like it is due to an unknown network error that isn't handled and ends up calling VERIFY_NOT_REACHED() when converting error types in the `on_buffered_request_finished` callback for `Web::ResourceLoader::load()`. So might be more Windows-specific curl request configuration/response handling needed.

#### Loading a web page from a `file://` request
ladybird_windows_file_requests

#### Loading a web page from a `http://` request

ladybird_windows_http_requests

#### Loading a web page from a `https://` request

ladybird_windows_load_https


Services/WebContent/PageClient.cpp
```diff
@@ -120,9 +120,15 @@ bool PageClient::is_connection_open() const
     return client().is_open();
 }
 
-bool PageClient::is_url_suitable_for_same_process_navigation(URL::URL const& current_url, URL::URL const& target_url) const
+bool PageClient::is_url_suitable_for_same_process_navigation([[maybe_unused]]URL::URL const& current_url, [[maybe_unused]] URL::URL const& target_url) const
 {
+#if defined(AK_OS_WINDOWS)
+    // FIXME: Attempting to request a new process for navigation causes the existing WebContent process to crash, which results in the actual request to not be initiated in the first place.
+    //  Note that this crash occurs for any type of request to shutdown (e.g. manually closing a tab). The Winsock2 error code seems to be WSANOTINITIALISED, so seems to be related to our init/cleanup of WinSock implementation
```
ayeteadoe: FYI @stasoid in case you have any idea off the top of your head, I haven't started to look into this deeply yet.

Libraries/LibWebView/Utilities.cpp
```diff
@@ -112,7 +112,7 @@ ErrorOr<Vector<ByteString>> get_paths_for_helper_process(StringView process_name
     auto application_path = TRY(application_directory());
     Vector<ByteString> paths;
 
-#if !defined(AK_OS_MACOS) && !defined(AK_OS_WINDOWS)
+#if !defined(AK_OS_MACOS)
     auto prefix = find_prefix(LexicalPath(application_path));
     TRY(paths.try_append(LexicalPath::join(prefix.string(), libexec_path, process_name).string()));
```
ADKaster, line 117: surely libexec isn't a thing on windows?

Shouldn't the helpers be in the same directory as the Ladybird.exe file? Is application_directory() returning one up from where it should?

Meta/CMake/sanitizers.cmake
```diff
@@ -13,6 +13,7 @@ if (ENABLE_ADDRESS_SANITIZER)
         link_libraries(clang_rt.asan_dynamic-x86_64.lib -wholearchive:clang_rt.asan_dynamic_runtime_thunk-x86_64.lib)
         get_clang_resource_dir(CLANG_RESOURCE_DIR)
         file(COPY "${CLANG_RESOURCE_DIR}/lib/windows/clang_rt.asan_dynamic-x86_64.dll" DESTINATION "${CMAKE_BINARY_DIR}/bin")
+        file(COPY "${CLANG_RESOURCE_DIR}/lib/windows/clang_rt.asan_dynamic-x86_64.dll" DESTINATION "${CMAKE_BINARY_DIR}/libexec")
```
ADKaster: 🤔 or I guess we are using libexec? Is that... normal?

Services/ImageDecoder/CMakeLists.txt
```diff
@@ -24,3 +24,13 @@ target_include_directories(imagedecoderservice PRIVATE ${LADYBIRD_SOURCE_DIR}/Se
 
 target_link_libraries(ImageDecoder PRIVATE imagedecoderservice LibCore LibMain LibThreading)
 target_link_libraries(imagedecoderservice PRIVATE LibCore LibGfx LibIPC LibImageDecoderClient LibMain LibThreading)
+
+if (WIN32)
+    add_custom_command(TARGET ImageDecoder POST_BUILD
```
ADKaster: there's a helper for this that should be used instead

Services/RequestServer/ConnectionFromClient.cpp
```diff
@@ -488,7 +491,20 @@ void ConnectionFromClient::start_request(i32 request_id, ByteString method, URL:
                 return;
             }
 
+        // FIXME: Should we create an abstraction here (e.g. RequestStream?) for the RequestServer read/write fds creation and response data writes?
```
ADKaster: >should we create an abstraction here

yes

Libraries/LibRequests/Request.cpp
```diff
@@ -37,8 +37,13 @@ void Request::set_request_fd(Badge<Requests::RequestClient>, int fd)
     VERIFY(m_fd == -1);
     m_fd = fd;
 
-    auto notifier = Core::Notifier::construct(fd, Core::Notifier::Type::Read);
+#if defined(AK_OS_WINDOWS)
```
ADKaster: answered by the comment on the fixme in the previous commit, but... 

would be nice to hide this ifdef soup in a helper class outside of this method.

Meta/CMake/sanitizers.cmake
R-Goc: no definitely not normal convention

Libraries/LibWebView/Process.cpp
```diff
@@ -89,7 +81,16 @@ ErrorOr<Optional<pid_t>> Process::get_process_pid(StringView process_name, Strin
         TRY(Core::System::unlink(pid_path));
         return OptionalNone {};
     }
-    if (kill(*pid, 0) < 0) {
+    bool const process_not_found = [&pid]() {
```
R-Goc, line 84: How is the code above here meant to work? Windows doesn't have a pid_path concept.

ADKaster: The pid path is just a normal file with a process ID in it that we write ourselves as a janky single-instance measure. Presumably on Windows we should use some app-specific uuid or something but that's way out of scope for initial bringup

Libraries/LibWebView/Utilities.cpp
ayeteadoe: Yeah everything should be in `bin`, I didn't bother fixing that up front in case we didn't care, but I'll adjust the CMake infra to remove `libexec` on Windows which will replace that commit that copies the ASAN DLL to libexec

Services/ImageDecoder/CMakeLists.txt
ayeteadoe: Ah yeah thanks I forgot I added that, this was all done before that change was made in that initial test-web PR

Services/WebContent/PageClient.cpp
stasoid: I don't think I saw this error, this is something new.

R-Goc: Not sure if relevant to this specifically but I just got this from application verifier:
```
VERIFIER STOP 000000000000E100: pid 0x10D0: Illegal networking API called from DllMain 

	00007FFDB4800410 : Networking function being called from DllMain
	000001A886064FD4 : Name of Dll making invalid call if not NULL
```
relevant stack frame:
```
03 00007ffd`c563e716     : 00000000`00000007 00000000`00000000 00000027`c9dff590 00000027`c9dffa30 : lagom_coreminimal!Core::System::init_crt_and_wsa+0x28 [D:\lib\ladybird\Libraries\LibCore\SystemWindows.cpp @ 35] 
04 00007ffd`9b7d0fb6     : 00000000`00000000 00000027`c9dff590 00000027`c9dffa30 00000000`00000001 : ucrtbase!initterm+0x36
05 00007ffd`9b7d119b     : 00000000`00000001 00000027`c9dffa30 000001a8`81676f00 00000000`00000001 : lagom_coreminimal!dllmain_crt_process_attach+0x9a [D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\dll_dllmain.cpp @ 66] 
06 00007ffd`ad390ec4     : 00007ffd`9b7c0000 00000000`00000001 00000027`c9dffa30 000001a8`81678fb0 : lagom_coreminimal!dllmain_dispatch+0x6f [D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\dll_dllmain.cpp @ 276] 
```

R-Goc: Right, more info on discord, but since we have a dynamic build as each library gets loaded with LoadLibrary the constructors for static variables are called from within the automatically created DllMain. Since we do WSAStartup as part of a static variable construction we are asking for issues, same with the call to that crt function. WSAStartup even comes with a warning about it, but the list of things we cannot do in these is significantly longer. As we don't know if CRT is initialized or not.

ayeteadoe:
So for the HTTP(S) request instability, it looks like this is actually an issue at the IPC layer. I've confirmed every attempt I've tried (even when we crash) has `CURLE_OK` and we invoke `async_request_finished()` with an empty `Optional<Requests::NetworkError>`. I also confirmed that `RequestFinished::static_encode()` encodes the network error as an empty Optional, but immediately after in `RequestFinished::decode()` the Optional network error ends up being set with a garbage enum value.  I don't see any obvious issues with how `Optional<T>` is implemented in `IPC::Encoder/Decoder` and switching using a non-optional enum with an extra entry for Success still resulted in a garbage value being decoded, so maybe something during transport is corrupting the stream?

Services/WebContent/PageClient.cpp
ayeteadoe: Yeah the static init calling `WSAStartup` (and lackof `WSACleanup`?) was the initial thing I was assuming was causing the `WebContent` crash on shutdown issue. Moving the `WSAStartup/Cleanup()` calls to `LibMain` fixed the crash due to `WSANOTINITIALISED` when calling `closesocket()`; however, that was just the first layer of the onion and there is more to fix to stop crashing on shutdown:

1. `System::is_socket()` is returning false for some socket FDs and we end up trying to call `CloseHandle()` on them
2. A rough hack I tried to get past that issue now has us crashing in skia DLL which I haven't investigated at all yet

ayeteadoe:
So to try and confirm if data was getting corrupted during transport and not an issue with encoding/decoding, I switched the `network_error` field to not be the last thing encoded in the IPC message stream as so far I had never noticed any corruption in the other fields:

<details>

<summary>requestfinished_ipc_field_swap.patch</summary>

```patch
diff --git a/Libraries/LibRequests/RequestClient.cpp b/Libraries/LibRequests/RequestClient.cpp
index e07427c75f..54741a5450 100644
--- a/Libraries/LibRequests/RequestClient.cpp
+++ b/Libraries/LibRequests/RequestClient.cpp
@@ -73,7 +73,7 @@ bool RequestClient::set_certificate(Badge<Request>, Request& request, ByteString
     return IPCProxy::set_certificate(request.id(), move(certificate), move(key));
 }
 
-void RequestClient::request_finished(i32 request_id, u64 total_size, RequestTimingInfo timing_info, Optional<NetworkError> network_error)
+void RequestClient::request_finished(i32 request_id, Optional<NetworkError> network_error, RequestTimingInfo timing_info, u64 total_size)
 {
     RefPtr<Request> request;
     if ((request = m_requests.get(request_id).value_or(nullptr))) {
diff --git a/Libraries/LibRequests/RequestClient.h b/Libraries/LibRequests/RequestClient.h
index 5c74e3edd7..0e4aad1a1c 100644
--- a/Libraries/LibRequests/RequestClient.h
+++ b/Libraries/LibRequests/RequestClient.h
@@ -45,7 +45,7 @@ private:
     virtual void die() override;
 
     virtual void request_started(i32, IPC::File) override;
-    virtual void request_finished(i32, u64, RequestTimingInfo, Optional<NetworkError>) override;
+    virtual void request_finished(i32, Optional<NetworkError>, RequestTimingInfo, u64) override;
     virtual void certificate_requested(i32) override;
     virtual void headers_became_available(i32, HTTP::HeaderMap, Optional<u32>, Optional<String>) override;
 
diff --git a/Services/RequestServer/ConnectionFromClient.cpp b/Services/RequestServer/ConnectionFromClient.cpp
index 2d947e44e7..dcb02e2877 100644
--- a/Services/RequestServer/ConnectionFromClient.cpp
+++ b/Services/RequestServer/ConnectionFromClient.cpp
@@ -473,13 +473,13 @@ void ConnectionFromClient::start_request(i32 request_id, ByteString method, URL:
         ->when_rejected([this, request_id](auto const& error) {
             dbgln("StartRequest: DNS lookup failed: {}", error);
             // FIXME: Implement timing info for DNS lookup failure.
-            async_request_finished(request_id, 0, {}, Requests::NetworkError::UnableToResolveHost);
+            async_request_finished(request_id, Requests::NetworkError::UnableToResolveHost, {}, 0);
         })
         .when_resolved([this, request_id, host = move(host), url = move(url), method = move(method), request_body = move(request_body), request_headers = move(request_headers), proxy_data](auto const& dns_result) mutable {
             if (dns_result->is_empty() || !dns_result->has_cached_addresses()) {
                 dbgln("StartRequest: DNS lookup failed for '{}'", host);
                 // FIXME: Implement timing info for DNS lookup failure.
-                async_request_finished(request_id, 0, {}, Requests::NetworkError::UnableToResolveHost);
+                async_request_finished(request_id, Requests::NetworkError::UnableToResolveHost, {}, 0);
                 return;
             }
 
@@ -752,7 +752,7 @@ void ConnectionFromClient::check_active_requests()
                 }
             }
 
-            async_request_finished(request->request_id, request->downloaded_so_far, timing_info, network_error);
+            async_request_finished(request->request_id, network_error, timing_info, request->downloaded_so_far);
         }
 
         request->notify_about_fetching_completion();
diff --git a/Services/RequestServer/RequestClient.ipc b/Services/RequestServer/RequestClient.ipc
index 12d39bd055..d6f397f332 100644
--- a/Services/RequestServer/RequestClient.ipc
+++ b/Services/RequestServer/RequestClient.ipc
@@ -6,7 +6,7 @@
 endpoint RequestClient
 {
     request_started(i32 request_id, IPC::File fd) =|
-    request_finished(i32 request_id, u64 total_size, Requests::RequestTimingInfo timing_info, Optional<Requests::NetworkError> network_error) =|
+    request_finished(i32 request_id, Optional<Requests::NetworkError> network_error, Requests::RequestTimingInfo timing_info, u64 total_size) =|
     headers_became_available(i32 request_id, HTTP::HeaderMap response_headers, Optional<u32> status_code, Optional<String> reason_phrase) =|
 
     // Websocket API

```


The above is the non-generated code changes, which makes the generated encode/decode now look like this:
```cpp
    static ErrorOr<NonnullOwnPtr<RequestFinished>> decode(Stream& stream, Queue<IPC::File>& files)
    {
        IPC::Decoder decoder { stream, files };
        auto request_id = TRY((decoder.decode<i32>()));
        auto network_error = TRY((decoder.decode<Optional<Requests::NetworkError>>()));
        auto timing_info = TRY((decoder.decode<Requests::RequestTimingInfo>()));
        auto total_size = TRY((decoder.decode<u64>()));
        return make<RequestFinished>(move(request_id), move(total_size), move(timing_info), move(network_error));
    }

    static ErrorOr<IPC::MessageBuffer> static_encode(i32 request_id, u64 total_size, Requests::RequestTimingInfo const& timing_info, Optional<Requests::NetworkError> const& network_error)
    {
        IPC::MessageBuffer buffer;
        IPC::Encoder stream(buffer);
        TRY(stream.encode(ENDPOINT_MAGIC));
        TRY(stream.encode((int)MessageID::RequestFinished));
        TRY(stream.encode(request_id));
        TRY(stream.encode(network_error));
        TRY(stream.encode(timing_info));
        TRY(stream.encode(total_size));
        return buffer;
    }

```

</details>


 And when I did that, there were a few observations leading me to believe something in `TransportWindows` is probably not right:
 1. For the field that is now encoded and decoded last (`u64 total_size`), I have not seen a single attempt when debugging where the decoded value was the same as what we encoded. It always comes out as some garbage number.
 2. After over 20 tries in a row of loading the demo web pages from HTTP(S) requests above, I have not crashed or observed the `Optional<NetworkError>` issue at all. Before this, I would've been lucky to get 20-30% page load requests without crashing.
 
So this explains why seemingly at random the `Optional<NetworkError>` would get decoded as being initialized with a garbage enum value, as the data from the last field seemingly always was corrupted so how the bytes actually got interpreted when decoding the `Optional<NetworkError>` never matched it's original state.

@stasoid So looks like there may be something fishy with how the [write, read, or both is happening in TransportSocketWindows](https://github.com/LadybirdBrowser/ladybird/blob/master/Libraries/LibIPC/TransportSocketWindows.cpp#L53-L205). What I might try next is adding a CRC32C-based checksum (temporarily for debugging) to the MessageHeader to see if the problem is in fact at this layer and data is corrupted, but do you see anything that sticks out in the current impl? No worries if not as I can try to dig in and investigate further as I haven't looked at this code in depth yet.

stasoid:
The only idea I have is that maybe the recent changes to catch up with the Linux version introduced the error, but that's just a shot in the dark. That code has some boundary checks: `index + sizeof(MessageHeader) <= m_unprocessed_bytes.size()`.

Services/WebContent/PageClient.cpp
ayeteadoe: Okay, turns out the LibMain fix and then avoiding the usage of `System::is_socket()` in `PosixSocketHelper::close()` was all that was needed. Pushed up the 2 commits that fix things and removed the  temporary `PageClient` commit as HTTP(S) page loading in a newly requested WebContent process is now working.

Libraries/LibCore/SystemWindows.cpp
```diff
@@ -27,19 +27,14 @@ static void invalid_parameter_handler(wchar_t const*, wchar_t const*, wchar_t co
...
-    int rc = WSAStartup(version, &wsa);
-    VERIFY(!rc && wsa.wVersion == version);
-
     // Make _get_osfhandle return -1 instead of crashing on invalid fd in release (debug still __debugbreak's)
     _set_invalid_parameter_handler(invalid_parameter_handler);
     return 0;
 }
 
-static auto dummy = init_crt_and_wsa();
+static auto dummy = init_crt();
```
R-Goc: Are you sure this is fine? Calling CRT functions during during CRT init and DllMain is generally considered unsafe. To be sure can we move it to where WSA init is now? Testing on my machine I just moved the whole function to libmain and called it there.

ayeteadoe: I was just focused on the WinSock2 problem, but I'll create a separate commit to split this out too

ayeteadoe: Ah actually, I'll need to re-work this a bit given LibTest does not use LibMain. Windows CI should fail for TestLibCoreStream, but going to let it run to save the build caches before pushing up the fix

ayeteadoe:
@stasoid I found the issue! I was able to confirm that the bytes being sent over were fine, plus if there was a bug in transport then pretty much every IPC call should've been failing, not just this one. After more playing around with the content that was encoded/decoded, I noticied the reason my initial encoding/decoding order switch worked was because I moved the `Optional<NetworkError>` to be decoded _**before**_ the `RequestTimingInfo` struct was encoded/decoded.

And sure enough, if you look at how we [IPC encode/decode the RequestTimingInfo struct](https://github.com/LadybirdBrowser/ladybird/blob/master/Libraries/LibRequests/RequestTimingInfo.h#L16-L75), we encode with `i64`s (8 bytes), but decode with `long`s (8 bytes on 64-bit Unix, but 4 bytes on Windows). So no wonder requests succeeding felt like RNG 😅 

Once I fixed that, the HTTP(S) requests became stable and we no longer crash 70-80% of the time

Libraries/LibTest/TestMain.cpp
```diff
@@ -18,13 +21,20 @@ int TEST_MAIN(int argc, char** argv)
         return 1;
     }
 
+#if defined(AK_OS_WINDOWS)
+    windows_init();
```
R-Goc, line 25: I was actually thinking about what to do with tests and the invalid parameter handler. We might still want it to break even in release modes for tests.

stasoid:
Nice, shooting yourself in the foot with a long shotgun. Classic C(++).

diegoiast:
wow... impressive! once this gets merged, I am so closing https://github.com/LadybirdBrowser/ladybird/issues/38

At that stage, Win64 support at the same level as Linux/macOS. Respect! nicely done!

R-Goc:
> At that stage, Win64 support at the same level as Linux/macOS. Respect! nicely done!

While the optimism is nice to see, we are still quite far away from the same level of support. Both at the technical level and the dev time dedicated to windows.

UI/Qt/CMakeLists.txt
```diff
@@ -16,10 +16,10 @@ target_sources(ladybird PRIVATE
     ladybird.qrc
 )
 target_link_libraries(ladybird PRIVATE Qt::Core Qt::Gui Qt::Widgets)
-lagom_copy_runtime_dlls(ladybird)
 create_ladybird_bundle(ladybird)
 
 if (WIN32)
+    lagom_windows_bin(ladybird)
```
sakgoyal, line 22: I don't understand CMake well, but would changing this to `_bin(Ladybird)` (capital L) be better? 
that way you dont need to patch the `ladybird.py` for windows.

ayeteadoe: The actual target is `ladybird` (see [target declaration](https://github.com/LadybirdBrowser/ladybird/blob/master/UI/Qt/CMakeLists.txt#L1)), so in CMake-land we use `ladybird`. If you look at [how ladybird.py builds](https://github.com/LadybirdBrowser/ladybird/blob/master/Meta/ladybird.py#L336), it is just invoking `ninja`, not `cmake --build`.

The `Ladybird` alias target is possible to begin with due to [us setting the output name to Ladybird](https://github.com/LadybirdBrowser/ladybird/blob/master/UI/CMakeLists.txt#L5). On Unix, ninja will properly recognize `Ladybird` and map it to the actual `ladybird` target; however, it won't do that on Windows.

<details>
<summary>Here are some details showing what is happening</summary>

Here is a minimal repro of the scenario:
```
cmake_minimum_required(VERSION 3.25)

project(cmake-ninja-test)

add_executable(foobar main.cpp)

set_target_properties(foobar PROPERTIES
    OUTPUT_NAME "Foobar"
)

```

On Unix, the following will work:
```
cmake -G Ninja -S . -B build
ninja -C build Foobar
```

On Windows, we get:
```
ninja: error: unknown target 'Foobar', did you mean 'foobar'?
ninja: Entering directory `build'
```

Probably due to the difference that the actual output binary on Windows is `bin\Ladybird.exe` not `bin/Ladybird` so ninja can't map it? Not 100% sure there.

</details>

ayeteadoe:
While the work here is complete, I'm drafting this for now as I'm going to split out a couple of the simple commits here that don't require/benefit from the general "Support Ladybird on Windows" context into standalone PRs. Once those are merged I'll mark this as ready for review again.

AK/Windows.h
```diff
@@ -18,4 +20,40 @@
...
+    VERIFY(rc == 0 && wsa.wVersion == version);
+}
+
+inline void terminate_wsa()
+{
+    int rc = WSACleanup();
+    VERIFY(rc == 0);
+}
+
+static void invalid_parameter_handler(wchar_t const*, wchar_t const*, wchar_t const*, unsigned int, uintptr_t)
```
ADKaster, line 39: static function in a header is no bueno -- even if it's empty.

It makes sense for these methods to be explicitly included into their respective main()s, but having them inline in each TU that has AK/Windows.h feels meh.

Non-blocking comment for sure though.

ADKaster (approved):
Verified this works on google.com and ladybird.org with

VS 2022 Version 17.14.19
clang version 19.1.5

The scrolling perf on ladybird.org is atrocious, and the default font makes no sense, but this PR has been open forever and works so let's go

PR #5982, opened 2025-08-25
Meta: Fix variable name `switfc` to `swiftc`

sayhany:
**File:** `Meta/ladybird.py`

- In `configure_main()`, `switfc: Optional[str] = None` is initialized. For Swift
  presets, `(cc, cxx, swiftc)` is assigned via `pick_swift_compilers()`.
- The conditional checks `switfc` instead of `swiftc`, so `-DCMAKE_Swift_COMPILER`
  is never appended to the CMake arguments even when a Swift preset is used.
- Rename `switfc` to `swiftc` and update the conditional to correctly pass the
  Swift compiler to CMake.

This change affects only Swift presets; no behavior change otherwise.


gmta:
Thank you @sayhany!

PR #5983, opened 2025-08-25
LibWeb: Respect IncludeCredentials for Set-Cookie during fetch

jfdoming:
This is my first contribution to this project, so any feedback is quite welcome :) Especially wondering about the approach for testing. I had to modify the `http-test-server.py` in order to be able to inspect the cookies that were set in the test environment; but open to other approaches.

Fixes one Web Platform subtest: [Don't obey Set-Cookie when withCredentials=false](https://wpt.live/cors/credentials-flag.htm)


Tests/LibWeb/Fixtures/http-test-server.py
```diff
@@ -138,7 +147,13 @@ def handle_echo(self):
                     self.send_header(header, value)
                 self.end_headers()
 
-            response_body = echo.body or ""
+            if echo.reflect:
```
rgret-dev: I think the headers are already reflected just above, no?

jfdoming: They are, but that sends them as headers rather than in the body. Some headers aren't accessible as response headers, including Cookie headers

rgret-dev: That makes sense. I would suggest a more descriptive name, e.g. `reflect_headers_in_body`

Libraries/LibWeb/Loader/ResourceLoader.cpp
```diff
@@ -575,9 +575,14 @@ void ResourceLoader::handle_network_response_headers(LoadRequest const& request,
     if (!request.page())
         return;
 
-    for (auto const& [header, value] : response_headers.headers()) {
-        if (header.equals_ignoring_ascii_case("Set-Cookie"sv)) {
-            store_response_cookies(*request.page(), request.url().value(), value);
+    if (request.respects_set_cookie()) {
+        // From https://fetch.spec.whatwg.org/#http-network-fetch:
```
kennethmyhra: nit: This should point to this url: `https://fetch.spec.whatwg.org/#concept-http-network-fetch`. Then it is easier to search through the codebase and find the corresponding method in `Fetch/Fetching/Fetching.cpp`.

Libraries/LibWeb/Loader/LoadRequest.h
```diff
@@ -41,6 +41,9 @@ class WEB_API LoadRequest {
     ByteBuffer const& body() const { return m_body; }
     void set_body(ByteBuffer body) { m_body = move(body); }
 
+    bool respects_set_cookie() const { return m_respects_set_cookie; }
+    void set_respects_set_cookie(bool respects_set_cookie) { m_respects_set_cookie = respects_set_cookie; }
```
kennethmyhra: I am not a big fan of the name on these method I would rather see them named like below. But let one of the maintainers voice their opinion on this before you change them. (same goes of course for the member variable)
```cpp
bool store_set_cookie_headers() const { return m_store_set_cookie_headers; }
void set_store_set_cookie_headers(bool store_set_cookie_headers) { m_store_set_cookie_headers = store_set_cookie_headers; }
```

kennethmyhra (commented):
Given the non-standard approach we already have in place for [HTTP-network fetch](https://fetch.spec.whatwg.org/#concept-http-network-fetch) I think this is a reasonable way to do this.

Libraries/LibWeb/Loader/ResourceLoader.cpp
```diff
@@ -575,9 +575,14 @@ void ResourceLoader::handle_network_response_headers(LoadRequest const& request,
     if (!request.page())
         return;
 
-    for (auto const& [header, value] : response_headers.headers()) {
-        if (header.equals_ignoring_ascii_case("Set-Cookie"sv)) {
-            store_response_cookies(*request.page(), request.url().value(), value);
+    if (request.respects_set_cookie()) {
```
konradekk: just wondering: why not to pass `include_credentials` enumeration down here, if the spec you cite explicitly mentions it?

Libraries/LibWeb/Loader/ResourceLoader.cpp
jfdoming: Wasn't aware of this, thanks! Updated.

Libraries/LibWeb/Loader/LoadRequest.h
jfdoming: I like your naming much better, updated.

Libraries/LibWeb/Loader/ResourceLoader.cpp
jfdoming: It's a good question. My understanding was that ResourceLoader is supposed to be more generic, so I was trying to limit the amount of fetch-specific stuff I add to it. But don't feel super strongly here

konradekk: maybe it just belongs to `fetch` then? the spec citation you gave, seems to be torn out from `Fetching.cpp:2407`: there’s already unused `request`, included `include_credentials` (both prefixed with `(void)` – to be removed then), so this seem to be a perfect place to update `response` headers.

tcl3 (approved):
Thanks for this!

PR #5984, opened 2025-08-25
LibWeb: Dont allocate space for shorthands in ComputedProperties

Calme1709:
We only ever compute values for longhands so having space allocated for shorthands is just wasted memory.

Most of the commits are around making sure we don't ever try to get shorthands from ComputedProperties but there are a couple miscellaneous cleanups that don't warrant their own PR as well - see individual commits for details.

### DIFF
--- a/Libraries/LibWeb/CSS/ComputedProperties.h
+++ b/Libraries/LibWeb/CSS/ComputedProperties.h
@@ -258,8 +258,6 @@ public:
     }
 
 private:
-    friend class StyleComputer;
-
     ComputedProperties();
 
     virtual void visit_edges(Visitor&) override;
--- a/Libraries/LibWeb/CSS/ComputedProperties.h
+++ b/Libraries/LibWeb/CSS/ComputedProperties.h
@@ -227,8 +227,7 @@ public:
 
     [[nodiscard]] CSSPixels line_height() const { return *m_line_height; }
     void set_line_height(Badge<StyleComputer> const&, CSSPixels line_height) { m_line_height = line_height; }
-    [[nodiscard]] CSSPixels font_size() const { return *m_font_size; }
-    void set_font_size(Badge<StyleComputer> const&, CSSPixels font_size) { m_font_size = font_size; }
+    [[nodiscard]] CSSPixels font_size() const;
 
     bool operator==(ComputedProperties const&) const;
 
@@ -283,7 +282,6 @@ private:
     RefPtr<Gfx::Font const> m_first_available_computed_font;
 
     Optional<CSSPixels> m_line_height;
-    Optional<CSSPixels> m_font_size;
 
     PseudoClassBitmap m_attempted_pseudo_class_matches;
 };
--- a/Libraries/LibWeb/CSS/ComputedProperties.h
+++ b/Libraries/LibWeb/CSS/ComputedProperties.h
@@ -66,7 +66,6 @@ public:
         Yes,
     };
     StyleValue const& property(PropertyID, WithAnimationsApplied = WithAnimationsApplied::Yes) const;
-    StyleValue const* maybe_null_property(PropertyID) const;
     void revert_property(PropertyID, ComputedProperties const& style_for_revert);
 
     GC::Ptr<CSSStyleDeclaration const> animation_name_source() const { return m_animation_name_source; }
--- a/Libraries/LibWeb/CSS/ComputedProperties.h
+++ b/Libraries/LibWeb/CSS/ComputedProperties.h
@@ -31,7 +31,6 @@ class WEB_API ComputedProperties final : public JS::Cell {
 
 public:
     static constexpr double normal_line_height_scale = 1.15;
-    static constexpr size_t number_of_properties = to_underlying(last_property_id) + 1;
 
     virtual ~ComputedProperties() override;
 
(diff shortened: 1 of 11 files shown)


Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -2143,21 +2143,18 @@ void StyleComputer::absolutize_values(ComputedProperties& style) const
...
+        style.set_property(PropertyID::LineHeight, LengthStyleValue::create(Length::make_px(line_height)), style.is_property_inherited(PropertyID::LineHeight) ? ComputedProperties::Inherited::Yes : ComputedProperties::Inherited::No);
 
-    for (size_t i = 0; i < style.m_property_values.size(); ++i) {
-        auto& value_slot = style.m_property_values[i];
-        if (!value_slot)
-            continue;
-        value_slot = value_slot->absolutized(viewport_rect(), font_metrics, m_root_element_font_metrics);
-    }
+    style.for_each_property([&](PropertyID property_id, auto& value) {
+        style.set_property(property_id, value.absolutized(viewport_rect(), font_metrics, m_root_element_font_metrics), style.is_property_inherited(property_id) ? ComputedProperties::Inherited::Yes : ComputedProperties::Inherited::No);
```
gmta: Maybe extract `value.absolutized(..)` or `style.is_property_inherited(..) ? :` into a separate variable here and a couple of lines above, to reduce line length a bit.

gmta (commented):
Nice!

Libraries/LibWeb/CSS/ComputedProperties.h
```diff
@@ -283,7 +282,6 @@ class WEB_API ComputedProperties final : public JS::Cell {
     RefPtr<Gfx::Font const> m_first_available_computed_font;
 
     Optional<CSSPixels> m_line_height;
-    Optional<CSSPixels> m_font_size;
```
awesomekling, line 286: We originally added this because `font_size()` was getting called all over the place in layout (due to units like `em` resolving against the font size). It seems likely that changing this to a `property()` lookup every time would cause a performance regression :)

Calme1709: I'm not sure that's correct - this appears to me to have been added in error in b85a8a23a704c61326381afbb93a0903bf505dff which is related to handling of percentage values for font-size rather than anything to do with performance.

`ComputedValues::font_size()` is rather hot but `ComputedProperties::font_size()` is only used in a couple places, none of which are for layout:
 - `Length::ResolutionContext::for_element` (which is only used on the very rare occasion when we don't have a layout node, e.g. `SVGFEFloodElement::flood_color`).
 - `StyleComputer::collect_animation_into`
 - `StyleComputer::absolutize_values`
 - `NodeWithStyle::apply_style` (as with all the other computed properties)

A profile of loading https://github.com puts this function at 0.00% on this branch.

Libraries/LibWeb/CSS/StyleComputer.cpp
Calme1709: Done

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -2136,48 +2136,28 @@ Gfx::Font const& StyleComputer::initial_font() const
...
 
-    for (size_t i = 0; i < style.m_property_values.size(); ++i) {
-        auto& value_slot = style.m_property_values[i];
-        if (!value_slot)
-            continue;
-        value_slot = value_slot->absolutized(viewport_rect(), font_metrics, m_root_element_font_metrics);
-    }
+    style.for_each_property([&](PropertyID property_id, auto& value) {
+        auto const& absolutized_value = value.absolutized(viewport_rect(), font_metrics, m_root_element_font_metrics);
+        auto const& is_inherited = style.is_property_inherited(property_id) ? ComputedProperties::Inherited::Yes : ComputedProperties::Inherited::No;
```
gmta: You don't need to take a reference to an enum value here, just `auto` suffices

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -2136,48 +2136,28 @@ Gfx::Font const& StyleComputer::initial_font() const
...
+    auto const& line_height_specified_value = style.property(CSS::PropertyID::LineHeight, ComputedProperties::WithAnimationsApplied::No);
 
     auto line_height = style.compute_line_height(viewport_rect(), font_metrics, m_root_element_font_metrics);
     font_metrics.line_height = line_height;
 
     // NOTE: line-height might be using lh which should be resolved against the parent line height (like we did here already)
-    if (line_height_value_slot && line_height_value_slot->is_length())
-        line_height_value_slot = LengthStyleValue::create(Length::make_px(line_height));
+    if (line_height_specified_value.is_length() || line_height_specified_value.is_percentage())
+        style.set_property(PropertyID::LineHeight, LengthStyleValue::create(Length::make_px(line_height)), style.is_property_inherited(PropertyID::LineHeight) ? ComputedProperties::Inherited::Yes : ComputedProperties::Inherited::No);
```
gmta: This line could do with a similar length reduction by reformatting / extracting variables. We try to aim for a width of 120 characters.

Libraries/LibWeb/CSS/ComputedProperties.h
awesomekling: Ah you are right! I misremembered :)

Libraries/LibWeb/CSS/StyleComputer.cpp
Calme1709: Whoops - fixed.

Libraries/LibWeb/CSS/StyleComputer.cpp
Calme1709: Done

gmta (approved):

PR #5988, opened 2025-08-25
LibWeb: Support decoding SVG favicons

kalenikaliaksandr:
Adds a path that checks if blob contains SVG image before reaching for image decoder.

Fixes logged image decoding errors on https://chatgpt.com/

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLLinkElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLLinkElement.cpp
@@ -9,6 +9,7 @@
 
 #include <AK/ByteBuffer.h>
 #include <AK/Debug.h>
+#include <LibGfx/ImmutableBitmap.h>
 #include <LibTextCodec/Decoder.h>
 #include <LibURL/URL.h>
 #include <LibWeb/Bindings/HTMLLinkElementPrototype.h>
@@ -34,6 +35,7 @@
 #include <LibWeb/Loader/ResourceLoader.h>
 #include <LibWeb/Page/Page.h>
 #include <LibWeb/Platform/ImageCodecPlugin.h>
+#include <LibWeb/SVG/SVGDecodedImageData.h>
 
 namespace Web::HTML {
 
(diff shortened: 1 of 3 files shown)


Libraries/LibWeb/HTML/HTMLLinkElement.cpp
```diff
@@ -596,8 +598,31 @@ void HTMLLinkElement::resource_did_load_favicon()
     document().check_favicon_after_loading_link_resource();
 }
 
-static NonnullRefPtr<Core::Promise<Web::Platform::DecodedImage>> decode_favicon(ReadonlyBytes favicon_data, URL::URL const& favicon_url, GC::Ref<DOM::Document> document)
+static NonnullRefPtr<Core::Promise<bool>> decode_favicon(ReadonlyBytes favicon_data, URL::URL const& favicon_url, GC::Ref<DOM::Document> document)
```
konradekk, line 601: nit: you should possibly get away with `Core::Promise<void>` here, as you seem not to use this `bool` anywhere?

Libraries/LibWeb/HTML/HTMLLinkElement.cpp
```diff
@@ -596,8 +598,31 @@ void HTMLLinkElement::resource_did_load_favicon()
...
 {
+    if (favicon_url.basename().ends_with(".svg"sv)) {
```
dotnetCarpenter, line 603: It might be infeasible to do but shouldn’t you check for `image/svg+xml` mime type instead of file extension?
https://www.w3.org/TR/SVG11/mimereg.html

Libraries/LibWeb/HTML/HTMLLinkElement.cpp
kalenikaliaksandr: right, but our Promise class has assumption that Result type could be referenced. not sure if I want to change only for this use case.

konradekk: might be usable in the future as a generic idea, but probably you’re right to postpone such changes until more use cases present themselves. 👍

PR #5989, opened 2025-08-25
LibWeb: Fix WPT CSSOM variable-names test

luizgfranca:
This PR fixes at least the CSSOM variable-names WPT test.

This was done by:
 - Correctly dealing with Custom Properties on `CSSStyleProperties` accessors and serialization.
 - Fixing serialization of identifiers by escaping them


gmta:
@luizgfranca quick tip: make sure your commits are atomic, i.e. all related functional changes are packed into a single commit. So for example, if you update `.length` to include custom properties, you should also be able to enumerate them in that same commit.

luizgfranca:
> @luizgfranca quick tip: make sure your commits are atomic, i.e. all related functional changes are packed into a single commit. So for example, if you update `.length` to include custom properties, you should also be able to enumerate them in that same commit.

Makes sense, I went committing as I was solving the sub-tests, but thinking back, you suggestion makes things clearer for someone looking back at the history

Libraries/LibWeb/CSS/CSSStyleProperties.cpp
```diff
@@ -148,7 +146,15 @@ String CSSStyleProperties::item(size_t index) const
         return string_from_property_id(property_id).to_string();
     }
 
-    return CSS::string_from_property_id(m_properties[index].property_id).to_string();
+    if (index < custom_properties_count) {
+        auto keys = m_custom_properties.keys();
+        auto custom_property = m_custom_properties.get(keys[index]);
+        if (custom_property.has_value()) {
+            return CSS::string_from_property_id(custom_property.ptr()->property_id).to_string();
+        }
```
AtkinsSJ: This `if` shouldn't be needed: If the HashMap can't find a value for its own key, something has gone very wrong. Also in that case, we'd end up falling through to the return below this which would do a lookup with a wrong index.

Libraries/LibWeb/CSS/Serialize.cpp
```diff
@@ -238,7 +238,8 @@ String serialize_a_css_declaration(StringView property, StringView value, Import
     StringBuilder builder;
 
     // 2. Append property to s.
-    builder.append(property);
+    // AD-HOC: we need to seralize the property identifier here, otherwise we woudn't be able to assign the declaration again, and the seralization would be incorrect
```
AtkinsSJ: Can you report a spec issue for this please? And then include the URL here.

Tests/LibWeb/Text/expected/wpt-import/css/css-nesting/nested-declarations-cssom.txt
```diff
@@ -2,14 +2,14 @@ Harness status: OK
...
+11 Pass
+1 Fail
+Pass	Trailing declarations
 Fail	Mixed declarations
-Fail	CSSNestedDeclarations.style
+Pass	CSSNestedDeclarations.style
 Pass	Nested group rule
 Pass	Nested @scope rule
-Fail	Inner rule starting with an ident
+Pass	Inner rule starting with an ident
```
AtkinsSJ, line 12: This test expectation change should go in the commit that affects the results.

PR #5997, opened 2025-08-26
LibWeb/HTML: Implement date conversions for month and week inputs

skyz1:
This implements the conversion algorithms between strings and dates for inputs of the types month and week.


Libraries/LibWeb/HTML/HTMLInputElement.cpp
```diff
@@ -2376,6 +2376,15 @@ static Utf16String convert_number_to_month_string(double input)
     return Utf16String::formatted("{:04d}-{:02d}", static_cast<int>(year), static_cast<int>(months) + 1);
 }
 
+// https://html.spec.whatwg.org/multipage/input.html#month-state-(type=month):concept-input-value-date-string
+static Utf16String convert_date_to_month_string(double input)
+{
+    // The algorithm to convert a number to a string, given a number input, is as follows: Return a valid month
+    // string that represents the month that has input months between it and January 1970.
+    auto date = AK::UnixDateTime::from_seconds_since_epoch(input / 1000.);
```
gmta: Why not `::from_milliseconds_since_epoch()`?

skyz1: The `convert_number_to_date_string` method also used `from_seconds_since_epoch`. That's why I thought there was a reason to do it that way. I can also fix it right away in the other method.

gmta: Right - please change it in the other method as well! Then this seems ready to be merged.

tcl3:
Running WPT against all `css/*/animation[s]/*` tests I'm seeing +653/-1068. A lot of the failures are for properties using `calc-size` which we don't implement.

Calme1709:
> Running WPT against all `css/*/animation[s]/*` tests I'm seeing +653/-1068. A lot of the failures are for properties using `calc-size` which we don't implement.

@tcl3 Is this comment on the correct PR? Perhaps you meant #6002?

tcl3:
> > Running WPT against all `css/*/animation[s]/*` tests I'm seeing +653/-1068. A lot of the failures are for properties using `calc-size` which we don't implement.
> 
> @tcl3 Is this comment on the correct PR? Perhaps you meant #6002?

Yes! Not sure how I managed to put this in completely the wrong place :sweat_smile:

PR #5999, opened 2025-08-27
LibWeb: Layout tree dump improvements

gmta:
* Inline nodes now dump their position
* Box model info is now always dumped
* The "content-size" dump was removed in favor of the box model dump, which shows the same info
* Misc. cleanup

A nice side effect of this is that it's painfully clear that all those `TextNode`s don't translate into paintables.


AtkinsSJ (approved):
Let's go!

PR #6001, opened 2025-08-27
AK: Use simdutf to validate UTF-16 strings as ASCII

trflynn89:
Updated other packages while we're here since this busts the vcpkg cache. The ffmpeg update includes support for VAAPI, which seems of interest to us looking at discord history.

### DIFF
--- a/AK/Utf16View.cpp
+++ b/AK/Utf16View.cpp
@@ -151,9 +151,7 @@ bool Utf16View::is_ascii() const
 {
     if (has_ascii_storage())
         return true;
-
-    // FIXME: Petition simdutf to implement an ASCII validator for UTF-16.
-    return all_of(utf16_span(), AK::is_ascii);
+    return simdutf::validate_utf16_as_ascii(m_string.utf16, length_in_code_units());
 }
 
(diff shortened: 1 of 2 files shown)


AK/Utf16View.cpp
```diff
@@ -151,9 +151,7 @@ bool Utf16View::is_ascii() const
 {
     if (has_ascii_storage())
         return true;
-
-    // FIXME: Petition simdutf to implement an ASCII validator for UTF-16.
-    return all_of(utf16_span(), AK::is_ascii);
+    return simdutf::validate_utf16_as_ascii(m_string.utf16, length_in_code_units());
```
gmta, line 154: Just out of curiosity, did you see a difference in performance?

trflynn89: Very rough benchmark, but:
```c++
BENCHMARK_CASE(ascii)
{
    auto string = u"FooBar😂FooBar😂FooBar😂"sv;

    for (size_t i = 0; i < 1'000'000'000; ++i) {
        (void)string.is_ascii();
    }
}
```

Goes from 4331ms before to 3492ms on my M2.

PR #6004, opened 2025-08-27
LibWeb: Start removing `auto` from Length

AtkinsSJ:
This is part 1 because we have a lot of places where we put `auto` inside a Length. :sweat_smile:

I'm not 100% sold on the LengthOrAutoOrCalculated name in particular!

### DIFF
--- a/Libraries/LibWeb/CSS/Parser/Parser.cpp
+++ b/Libraries/LibWeb/CSS/Parser/Parser.cpp
@@ -1648,14 +1648,19 @@ Optional<StyleProperty> Parser::convert_to_style_property(Declaration const& dec
     return StyleProperty { declaration.important, property_id.value(), value.release_value(), {} };
 }
 
-Optional<LengthOrCalculated> Parser::parse_source_size_value(TokenStream<ComponentValue>& tokens)
+Optional<LengthOrAutoOrCalculated> Parser::parse_source_size_value(TokenStream<ComponentValue>& tokens)
 {
     if (tokens.next_token().is_ident("auto"sv)) {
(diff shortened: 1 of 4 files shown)


Libraries/LibWeb/Painting/PaintableBox.cpp
```diff
@@ -1569,10 +1569,23 @@ void PaintableWithLines::resolve_paint_properties()
...
+            auto const& thickness = text_node.computed_values().text_decoration_thickness();
+            return thickness.value.visit(
+                [glyph_height](CSS::TextDecorationThickness::Auto) {
+                    // The UA chooses an appropriate thickness for text decoration lines; see below.
+                    // https://drafts.csswg.org/css-text-decor-4/#valdef-text-decoration-thickness-auto
+                    return max(glyph_height.scaled(0.1), 1);
+                },
+                [glyph_height](CSS::TextDecorationThickness::FromFont) {
+                    // If the first available font has metrics indicating a preferred underline width, use that width,
+                    // otherwise behaves as auto.
```
AtkinsSJ, line 1581: I tried looking into where this is defined in OpenType or other font formats, but I completely failed to find any mention of it, except articles about the `text-decoration-thickness` property.

AtkinsSJ:
Changes: Made `LengthOrAuto` implicitly constructable from `Length`, instead of it being `LengthOrAuto::make_length()`. That was just unnecessarily verbose, here and in my follow-up work.

AtkinsSJ:
Doing this in two parts is feeling messy, and I keep wanting to modify what I did here. So I'm closing this and will re-open as a single PR.

PR #6007, opened 2025-08-28
LibWeb: Begin storing ComputedProperties in computed form

Calme1709:
`StyleValue`s stored within `ComputedProperties` should be in their computed forms, this is for various reasons including:
 - Inheritance should be of computed values
 - Animations should work on computed values
 - Triggering transitions should work on computed values

Currently we store `StyleValue`s in an absolutized version of the specified value - this is equivalent to the computed form in many cases which is why this hasn't been causing significant issues but there are some cases such as `border-*-width` keywords where this is not the case.

Gains us 158 new WPT passes and 8 new fails.


AtkinsSJ (approved):
Nice work! I'm very happy to see this being tackled. I do hope that we won't need specific code for computing every single longhand though. :sweat_smile:

Calme1709:
Thankfully many properties' computed value is the same as it's specified value, and most of the remainder already have logic elsewhere in the codebase

PR #6008, opened 2025-08-28
LibWeb: Invalidate layout when view box properties are changed

tcl3:
This ensures the page updates immediately when the `viewBox` or `preserveAspectRatio` properties are updated

The majority of the code changed in this PR is a non-functional change that moves the implementation of `SVGFitToViewBox` into its own class. This is closer to the spec and ensures that we only implement the functionality once.

Gains us 5 WPT subtests in the `svg` directory.


Libraries/LibWeb/SVG/SVGFitToViewBox.cpp
```diff
@@ -38,6 +38,7 @@ void SVGFitToViewBox::attribute_changed(DOM::Element&, FlyString const& name, Op
     } else if (name.equals_ignoring_ascii_case(SVG::AttributeNames::preserveAspectRatio)) {
         m_preserve_aspect_ratio = AttributeParser::parse_preserve_aspect_ratio(value.value_or(String {}));
     }
+    element.set_needs_layout_tree_update(true, DOM::SetNeedsLayoutTreeUpdateReason::SVGViewBoxChange);
```
gmta: Shouldn't we only do this if we matched one of the attribute names?

tcl3: Good point. Fixed.

gmta (approved):

Issue #5408, opened 2025-07-11
LibWeb: Transferring a previously transferred MessagePort crashes


Lubrsi:
For example:
```html
<!DOCTYPE html>
<head>
</head>
<body>
    <script>
        window.onmessage = ({ data, ports }) => {
            if (data === "done")
                return;

            window.postMessage("done", "*", ports);
        };

        const channel = new MessageChannel();
        window.postMessage("port", "*", [ channel.port2 ]);
    </script>
</body>
</html>
```

Found on https://vscode.dev/ where it transfers a port to a "worker host iframe", which then transfers the port to a Worker.

trflynn89:
The issue is this bit here:
https://github.com/LadybirdBrowser/ladybird/blob/dbba6c0df948b616ee313baaa2d5fe57169fe6cc/Libraries/LibWeb/HTML/MessagePort.cpp#L99-L102

We are entangled but `m_remote_port` is null. The wording on the spec steps here ("entangled with another remote port remotePort") makes me think we should be returning false from `MessagePort::is_entangled()` if the remote port is null. This fixes the crash:
```diff
diff --git a/Libraries/LibWeb/HTML/MessagePort.cpp b/Libraries/LibWeb/HTML/MessagePort.cpp
index 9824a8dfcf7..dcc0de044ec 100644
--- a/Libraries/LibWeb/HTML/MessagePort.cpp
+++ b/Libraries/LibWeb/HTML/MessagePort.cpp
@@ -79,7 +79,7 @@ void MessagePort::visit_edges(Cell::Visitor& visitor)

 bool MessagePort::is_entangled() const
 {
-    return m_transport;
+    return m_transport && m_remote_port;
 }

 void MessagePort::set_worker_event_target(GC::Ref<DOM::EventTarget> target)
```

I think we've had similar issues in the past where we think we're entangled but the remote port is null.
CC @ADKaster

PR #6010, opened 2025-08-28
LibWeb: Don't die when transferring the same MessagePort more than once

awesomekling:
One MessagePort can be entangled with another MessagePort, either in the same agent, or in another agent.

In the same-agent case, the MessagePort objects point to each other via the MessagePort::m_remote_port field.

In the separate-agent case, they live in separate processes entirely and thus can't point at each other.

In both cases, the MessagePorts have an underlying transport channel, which means they are "entangled". However, we can't assume that being entangled means having a non-null m_remote_port.

This patch simply adds a missing null check for m_remote_port and thus makes https://vscode.dev/ stop crashing with a null dereference.

### DIFF
--- a/Libraries/LibWeb/HTML/MessagePort.cpp
+++ b/Libraries/LibWeb/HTML/MessagePort.cpp
@@ -99,7 +99,11 @@ WebIDL::ExceptionOr<void> MessagePort::transfer_steps(HTML::TransferDataEncoder&
     // 3. If value is entangled with another port remotePort, then:
     if (is_entangled()) {
         // 1. Set remotePort's has been shipped flag to true.
-        m_remote_port->m_has_been_shipped = true;
+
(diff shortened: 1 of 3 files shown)


Lubrsi:
Fixes #5408

trflynn89:
Related issue: #5408

Not clear to me whether we should be checking `m_remote_port` specifically here, or not entering the `is_entangled()` branch at all. The spec wording "If value is entangled with another port remotePort" had me leaning towards the latter.

awesomekling:
> Related issue: #5408
> 
> Not clear to me whether we should be checking `m_remote_port` specifically here, or not entering the `is_entangled()` branch at all. The spec wording "If value is entangled with another port remotePort" had me leaning towards the latter.

Duh, first time I'm seeing that issue. Should have searched before spending an hour debugging it 😅 

If we don't enter the branch, many of our tests begin to time out.

PR #6018, opened 2025-08-28
LibWeb: Fixes for scratch.mit.edu

erik-kz:
Three bug fixes that get the Scratch programming language working... mostly. There are a few games whose graphics don't render correctly, e.g. https://scratch.mit.edu/projects/1185431027/, but that seems to be an general issue with ladybird's SVG code. If I open the relevant SVGs on their own they still look wrong. Also audio does not work, and playing around with the graphics editor will occasionally crash the WebContent process during garbage collection.

But ignoring those caveats things seem to be working reasonably well. And these fixes might benefit other web apps as well, particularly if they rely on drawing images to a canvas which is currently pretty broken.


erik-kz:
Whoops, noticed that this breaks the naturalWidth-naturalHeight-width-height.tentative.html test. Marking as a draft for now.

erik-kz:
It didn't actually break the naturalWidth-naturalHeight-width-height.tentative.html test. A number of cases in that test are currently expected to fail and with this change 8 of those now pass. Unmarking as draft.

Libraries/LibWeb/SVG/ViewBox.cpp
```diff
@@ -28,8 +28,8 @@ Optional<ViewBox> try_parse_view_box(StringView string)
     ViewBox view_box;
 
     while (!lexer.is_eof()) {
-        lexer.consume_while([](auto ch) { return is_ascii_space(ch); });
-        auto token = lexer.consume_until([](auto ch) { return is_ascii_space(ch) && ch != ','; });
+        lexer.consume_while([](auto ch) { return is_ascii_space(ch) || ch == ','; });
```
AtkinsSJ: This seems subtly wrong... It'll treat `1,,,,,,2` the same as `1,2`.

`SVG::AttributeParser` has a `parse_comma_whitespace()` which does that more correctly. Maybe this whole method should go in `AttributeParser` and make use of the methods there.

(Also, a test for this would be very nice.)

Libraries/LibWeb/WebGL/OpenGLContext.cpp
```diff
@@ -60,8 +60,8 @@ OpenGLContext::OpenGLContext(NonnullRefPtr<Gfx::SkiaBackendContext> skia_backend
 OpenGLContext::~OpenGLContext()
 {
 #ifdef ENABLE_WEBGL
-    free_surface_resources();
     eglMakeCurrent(m_impl->display, EGL_NO_SURFACE, EGL_NO_SURFACE, EGL_NO_CONTEXT);
+    // eglDestroyContext implicitly destroys all objects associated with the context
```
AtkinsSJ: Can we get a test for this? eg, one that crashes without the change, and now doesn't. I'm not sure how deterministic this is, if it's not then don't worry.

Libraries/LibWeb/HTML/HTMLImageElement.cpp
```diff
@@ -206,7 +206,14 @@ Optional<CSSPixelFraction> HTMLImageElement::intrinsic_aspect_ratio() const
     return {};
 }
 
-RefPtr<Gfx::ImmutableBitmap> HTMLImageElement::current_image_bitmap(Gfx::IntSize size) const
+RefPtr<Gfx::ImmutableBitmap> HTMLImageElement::current_image_bitmap() const
+{
+    int w = intrinsic_width().value_or({}).to_int();
+    int h = intrinsic_height().value_or({}).to_int();
+    return current_image_bitmap_sized({ w, h });
+}
```
AtkinsSJ: As this is the same code for every class, it could go in `Web::Layout::ImageProvider` instead to avoid repeating it.

AtkinsSJ (changes requested):

Libraries/LibWeb/SVG/ViewBox.cpp
erik-kz: Ah true. I've written what should be a more robust implementation in AttributeParser.cpp along with a bunch of tests. This lets us do away with ViewBox.cpp entirely.

Libraries/LibWeb/WebGL/OpenGLContext.cpp
erik-kz: Well, this issue didn't actually cause a crash. It just left us with an incomplete framebuffer so nothing would get rendered.

But beyond that, currently all of the tests are run with GPU rendering disabled, which also disables WebGL. It probably would be a good idea to have a few basic tests for WebGL eventually, e.g. maybe a screenshot test that just draws a triangle or something. I'm not sure if there would be any difficulties getting that to work on GitHub's CI infrastructure.

kalenikaliaksandr (approved):

PR #6021, opened 2025-08-28
LibWeb: Hit test StackingContext's children before testing visibility

### LINKED ISSUE #6017: Regression: Hit testing no longer works on X.com
To reproduce, go to https://x.com/ and notice how none of the normally clickable elements are clickable, or even react to hover.

Regressed in #5681 (cc @zacoons)

gmta:
If a node that establishes a StackingContext has `pointer-events: none`, hit testing should first proceed with hit testing the SC's children before deciding to bail. We were checking for `pointer-events` too early, causing large parts of certain websites to be noninteractive.

Fixes #6017.

CC @awesomekling @zacoons


awesomekling:
Awesome, thanks for tracking it down :)

PR #6025, opened 2025-08-29
README.md updated

npc-gnu:
Hello. Normally, i was trying to read code but my brain gives kernel panic while trying to read it so i only added **Contributors** to README.md. I am accepting that i don't reserve a contributor status, but like i said; i did not understood anything from code.
Sorry again.


trflynn89:
Thank you for your interest in the project, but there's really no value in having a contributer list displayed in the README. There's already a contributer list on the repo home page.

PR #6027, opened 2025-08-29
LibWeb: Misc. cleanup of LayoutState

gmta:
A couple of small changes I made while trying to debug an issue.

### DIFF
--- a/Libraries/LibWeb/Layout/LayoutState.cpp
+++ b/Libraries/LibWeb/Layout/LayoutState.cpp
@@ -214,17 +214,9 @@ static void build_paint_tree(Node& node, Painting::Paintable* parent_paintable =
 
 void LayoutState::commit(Box& root)
 {
-    // Go through the old paintable tree and detach everything from the layout tree.
-    // The layout tree should only point to the new paintable tree which we're about to build.
-    if (auto* old_paintable_root = root.first_paintable()) {
-        old_paintable_root->for_each_in_inclusive_subtree([&](Painting::Paintable& paintable) {
-            paintable.detach_from_layout_node();
-            return TraversalDecision::Continue;
-        });
-    }
-
-    // For completeness, also go through the layout tree and detach all paintables.
-    root.for_each_in_inclusive_subtree([&](Layout::Node& node) {
+    // Go through the layout tree and detach all paintables. The layout tree should only point to the new paintable tree
+    // which we're about to build.
+    root.for_each_in_inclusive_subtree([](Node& node) {
         node.clear_paintables();
         return TraversalDecision::Continue;
     });
--- a/Libraries/LibWeb/Layout/LayoutState.cpp
+++ b/Libraries/LibWeb/Layout/LayoutState.cpp
@@ -56,7 +56,7 @@ static CSSPixelRect measure_scrollable_overflow(Box const& box)
     if (!box.paintable_box())
         return {};
 
-    auto& paintable_box = const_cast<Painting::PaintableBox&>(*box.paintable_box());
+    auto const& paintable_box = *box.paintable_box();
 
     if (paintable_box.scrollable_overflow_rect().has_value())
         return paintable_box.scrollable_overflow_rect().value();
@@ -151,7 +151,7 @@ static CSSPixelRect measure_scrollable_overflow(Box const& box)
         has_scrollable_overflow = !paintable_absolute_padding_box.contains(scrollable_overflow_rect);
     }
 
-    paintable_box.set_overflow_data(Painting::PaintableBox::OverflowData {
+    const_cast<Painting::PaintableBox&>(paintable_box).set_overflow_data({
         .scrollable_overflow_rect = scrollable_overflow_rect,
         .has_scrollable_overflow = has_scrollable_overflow,
     });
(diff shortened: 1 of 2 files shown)


Libraries/LibWeb/Layout/LayoutState.cpp
```diff
@@ -278,63 +270,57 @@ void LayoutState::commit(Box& root)
...
-                auto& svg_graphics_paintable = static_cast<Painting::SVGGraphicsPaintable&>(paintable_box);
+                auto& svg_graphics_paintable = static_cast<Painting::SVGGraphicsPaintable&>(*paintable_box);
                 svg_graphics_paintable.set_computed_transforms(*used_values.computed_svg_transforms());
             }
 
             if (used_values.computed_svg_path().has_value() && is<Painting::SVGPathPaintable>(paintable_box)) {
-                auto& svg_geometry_paintable = static_cast<Painting::SVGPathPaintable&>(paintable_box);
+                auto& svg_geometry_paintable = static_cast<Painting::SVGPathPaintable&>(*paintable_box);
                 svg_geometry_paintable.set_computed_path(move(*used_values.computed_svg_path()));
             }
```
tcl3: You could simplify this slightly like this:
```c++
if (auto* svg_geometry_paintable = as_if<Painting::SVGGraphicsPaintable>(paintable_box)) {
    if (used_values.computed_svg_transforms().has_value()) {
        svg_geometry_paintable->set_computed_transforms(*used_values.computed_svg_transforms());
    }

    if (used_values.computed_svg_path().has_value()) {
        svg_geometry_paintable->set_computed_path(move(*used_values.computed_svg_path()));
    }
}
```

gmta: I don't think I can, `::set_computed_path()` doesn't exist on `SVGGraphicsPaintable`.

tcl3: Of course :facepalm: You'd need 2 separate `as_if`s to do this:
```c++
if (auto* svg_graphics_paintable = as_if<Painting::SVGGraphicsPaintable>(paintable_box)) {
    if (used_values.computed_svg_transforms().has_value())
        svg_graphics_paintable->set_computed_transforms(*used_values.computed_svg_transforms());
}

if (auto* svg_geometry_paintable = as_if<Painting::SVGPathPaintable>(paintable_box)) {
    if (used_values.computed_svg_path().has_value())
         svg_geometry_paintable->set_computed_path(move(*used_values.computed_svg_path()));
}
```
(Actually tested this time)

gmta: Right, so I left those alone since it didn't result in fewer lines of code and changed the order of operations (`is<..>` check before `.has_value()`. Don't know if that's actually better?

tcl3: Personally, I'd like to see all static_casts of GC-allocated objects like this be replaced with `as`/`as_if` calls that ensure the cast is valid. Maybe a bit pedantic here, but I'd rather be over-cautious with this sort of thing than miss a potential type confusion bug.

An alternative here would be to replace the static_cast with `as`.

gmta: Done!

Libraries/LibWeb/Layout/LayoutState.cpp
```diff
@@ -278,63 +270,55 @@ void LayoutState::commit(Box& root)
...
                 }
             }
 
-            if (used_values.computed_svg_transforms().has_value() && is<Painting::SVGGraphicsPaintable>(paintable_box)) {
-                auto& svg_graphics_paintable = static_cast<Painting::SVGGraphicsPaintable&>(paintable_box);
-                svg_graphics_paintable.set_computed_transforms(*used_values.computed_svg_transforms());
-            }
+            if (auto* svg_graphics_paintable = as_if<Painting::SVGGraphicsPaintable>(paintable.ptr());
+                svg_graphics_paintable && used_values.computed_svg_transforms().has_value())
+                svg_graphics_paintable->set_computed_transforms(*used_values.computed_svg_transforms());
```
tcl3: This is a bit of a nitpick, but the indentation makes this hard to read. Can you put braces around this to make it more obvious where the block starts and ends? Alternatively, it could be split it into 2 if statements.

Libraries/LibWeb/Layout/LayoutState.cpp
```diff
@@ -278,63 +270,55 @@ void LayoutState::commit(Box& root)
...
 
-            if (used_values.computed_svg_path().has_value() && is<Painting::SVGPathPaintable>(paintable_box)) {
-                auto& svg_geometry_paintable = static_cast<Painting::SVGPathPaintable&>(paintable_box);
-                svg_geometry_paintable.set_computed_path(move(*used_values.computed_svg_path()));
-            }
+            if (auto* svg_path_paintable = as_if<Painting::SVGPathPaintable>(paintable.ptr());
+                svg_path_paintable && used_values.computed_svg_path().has_value())
+                svg_path_paintable->set_computed_path(move(*used_values.computed_svg_path()));
```
tcl3: Same here.

Libraries/LibWeb/Layout/LayoutState.cpp
gmta: Sure!

tcl3 (approved):

PR #6028, opened 2025-08-29
Everywhere: Slap some [[clang::lifetimebound]] where appropriate

alimpfard:
This first pass only applies to the following two cases:
- Public functions returning a view type into an object they own
- Public ctors storing a view type

This catches a grand total of one (1) issue, which is fixed in this very commit.


Libraries/LibMedia/DecoderError.h
```diff
@@ -66,14 +66,14 @@ class DecoderError {
     StringView string_literal() const { return m_description; }
 
 private:
-    DecoderError(DecoderErrorCategory category, ByteString description)
+    DecoderError(DecoderErrorCategory category, StringView description)
```
alimpfard: P.S. error fixed here.

nico:
Put fix in its own commit?

alimpfard:
Sure. Looks like there are somehow test failures anyway

gmta (approved):

PR #6029, opened 2025-08-30
LibWeb: Support CSS and JSON modules

skyz1:
This adds support for CSS and JSON modules. WASM module support is planned as well.

This PR is still in its early stages, but I wanted to open it to let others know that I'm working on the feature and will probably need a few more days to finish it. There is currently also a linker error :^(

### DIFF
--- a/Libraries/LibWeb/HTML/Scripting/Fetching.cpp
+++ b/Libraries/LibWeb/HTML/Scripting/Fetching.cpp
@@ -64,7 +64,7 @@ ScriptFetchOptions default_script_fetch_options()
 String module_type_from_module_request(JS::ModuleRequest const& module_request)
 {
     // 1. Let moduleType be "javascript".
-    String module_type = "javascript"_string;
+    String module_type = "javascript-or-wasm"_string;
 
     // 2. If moduleRequest.[[Attributes]] has a Record entry such that entry.[[Key]] is "type", then:
     for (auto const& entry : module_request.attributes) {
@@ -72,7 +72,7 @@ String module_type_from_module_request(JS::ModuleRequest const& module_request)
             continue;
 
         // 1. If entry.[[Value]] is "javascript", then set moduleType to null.
-        if (entry.value == "javascript"_string)
+        if (entry.value == "javascript-or-wasm"_string)
             module_type = ""_string; // FIXME: This should be null!
         // 2. Otherwise, set moduleType to entry.[[Value]].
         else
@@ -583,7 +583,7 @@ WebIDL::ExceptionOr<void> fetch_worklet_module_worker_script_graph(URL::URL cons
         }
 
         // 2. Fetch the descendants of and link result given fetchClient, destination, and onComplete. If performFetch was given, pass it along as well.
-        fetch_descendants_of_and_link_a_module_script(realm, as<JavaScriptModuleScript>(*result), fetch_client, destination, move(perform_fetch), on_complete);
+        fetch_descendants_of_and_link_a_module_script(realm, as<ModuleScript>(*result), fetch_client, destination, move(perform_fetch), on_complete);
     });
 
     // 2. Fetch a single module script given url, fetchClient, destination, options, settingsObject's realm, "client", true,
@@ -622,8 +622,8 @@ void fetch_single_module_script(JS::Realm& realm,
     PerformTheFetchHook perform_fetch,
     OnFetchScriptComplete on_complete)
 {
-    // 1. Let moduleType be "javascript".
-    String module_type = "javascript"_string;
+    // 1. Let moduleType be "javascript-or-wasm".
+    String module_type = "javascript-or-wasm"_string;
 
     // 2. If moduleRequest was given, then set moduleType to the result of running the module type from module request steps given moduleRequest.
     if (module_request.has_value())
@@ -686,7 +686,7 @@ void fetch_single_module_script(JS::Realm& realm,
     //     Otherwise, fetch request with processResponseConsumeBody set to processResponseConsumeBody as defined below.
     //     In both cases, let processResponseConsumeBody given response response and null, failure, or a byte sequence bodyBytes be the following algorithm:
     auto process_response_consume_body = [&module_map, url, module_type, &module_map_realm, on_complete](GC::Ref<Fetch::Infrastructure::Response> response, Fetch::Infrastructure::FetchAlgorithms::BodyBytes body_bytes) {
-        // 1. If either of the following conditions are met:
+        // 1. If any of the following are true:
         //    - bodyBytes is null or failure; or
         //    - response's status is not an ok status,
         if (body_bytes.has<Empty>() || body_bytes.has<Fetch::Infrastructure::FetchAlgorithms::ConsumeBodyFailureTag>() || !Fetch::Infrastructure::is_ok_status(response->status())) {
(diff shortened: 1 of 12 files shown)


Libraries/LibWeb/HTML/Scripting/ModuleScript.cpp
```diff
@@ -133,18 +135,64 @@ WebIDL::ExceptionOr<GC::Ptr<ModuleScript>> ModuleScript::create_a_json_module_sc
...
 }
 
+// https://html.spec.whatwg.org/multipage/webappapis.html#creating-a-webassembly-module-script
+// https://whatpr.org/html/9893/webappapis.html#creating-a-webassembly-module-script
+WebIDL::ExceptionOr<GC::Ptr<ModuleScript>> ModuleScript::create_a_webassembly_module_script(ByteString const& filename, ByteBuffer body_bytes, JS::Realm& realm, URL::URL base_url)
+{
+    // 1. If scripting is disabled for realm, then set bodyBytes to the byte sequence 0x00 0x61 0x73 0x6d 0x01 0x00 0x00 0x00.
+    // NOTE: This byte sequence corresponds to an empty WebAssembly module with only the magic bytes and version number provided.
+    if (HTML::is_scripting_disabled(realm))
+        ReadonlyBytes { { 0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x00 } }.copy_to(body_bytes);
```
alimpfard: This will not magically delete the rest of the body_bytes data, and the parser will most likely complain about it being invalid.
`body_bytes = body_bytes.slice(...)` probably?

Libraries/LibWeb/WebAssembly/WebAssemblyModule.cpp
```diff
@@ -0,0 +1,505 @@
...
+
+    // 9. Set module.[[ModuleRecord]] to moduleRecord.
+    module_record->m_module_record = module_record;
+
+    // 10. Return moduleRecord.
+    return module_record;
+}
+
+// https://webassembly.github.io/esm-integration/js-api/index.html#export-name-list
+Vector<Utf16FlyString> WebAssemblyModule::export_name_list()
```
alimpfard, line 116: Should we be caching this? resolve_export seems unnecessarily expensive as-is.

Libraries/LibWeb/HTML/Scripting/Fetching.cpp
```diff
@@ -721,10 +721,13 @@ void fetch_single_module_script(JS::Realm& realm,
             //    and options.
             // FIXME: Pass options.
             if (mime_type.has_value() && mime_type->is_javascript() && module_type == "javascript-or-wasm")
-                module_script = JavaScriptModuleScript::create(url.to_byte_string(), source_text, module_map_realm, response->url().value_or({})).release_value_but_fixme_should_propagate_errors();
+                module_script = ModuleScript::create_a_javascript_module_script(url.to_byte_string(), source_text, module_map_realm, response->url().value_or({})).release_value_but_fixme_should_propagate_errors();
+
+            // 3. If the MIME type essence of mimeType is "text/css" and moduleType is "css", then set moduleScript to
+            //    the result of creating a CSS module script given sourceText and moduleMapRealm.
+            if (mime_type->essence() == "text/css"sv && module_type == "css")
```
AtkinsSJ: The branch above (step 2) checks if `mime_type` has a value, is it safe to dereference it like this?

Libraries/LibWeb/HTML/Scripting/ModuleScript.cpp
```diff
@@ -70,9 +63,40 @@ WebIDL::ExceptionOr<GC::Ptr<JavaScriptModuleScript>> JavaScriptModuleScript::cre
...
+    // 3. Set script's base URL and fetch options to null.
+    auto script = realm.create<ModuleScript>(Optional<URL::URL> {}, filename, realm);
+
+    // 4. Set script's parse error and error to rethrow to null.
+    script->set_parse_error(JS::js_null());
+    script->set_error_to_rethrow(JS::js_null());
+
+    // 5. Let sheet be the result of running the steps to create a constructed CSSStyleSheet with an empty dictionary as
+    //    the argument.
+    auto sheet = CSS::CSSStyleSheet::construct_impl(realm).release_value_but_fixme_should_propagate_errors();
```
AtkinsSJ: `TRY()` ?

Tests/LibWeb/Text/expected/wpt-import/html/semantics/scripting-1/the-script-element/json-module/module.txt
AtkinsSJ, line 1: Is this file supposed to be empty?

Libraries/LibWeb/HTML/Scripting/ModuleScript.cpp
```diff
@@ -94,6 +94,45 @@ WebIDL::ExceptionOr<GC::Ptr<ModuleScript>> ModuleScript::create_a_css_module_scr
...
+    if (result.is_error()) {
+        script->set_parse_error(result.error().value());
+        return script;
+    }
+
+    // 6. Set script's record to result.
+    script->m_record = result.value();
+
+    // NON-STANDARD: Clean up after parsing json with realm.
+    realm.vm().pop_execution_context();
```
AtkinsSJ: If we return in step 5, we won't pop this execution context. Maybe use a `ScopeGuard` instead.

... Actually, it should probably be a direct `TemporaryExecutionContext` instead of a manually constructed one? I'm not 100% sure though.

Libraries/LibWeb/HTML/Scripting/Fetching.cpp
```diff
@@ -728,8 +728,10 @@ void fetch_single_module_script(JS::Realm& realm,
             if (mime_type->essence() == "text/css"sv && module_type == "css")
                 module_script = ModuleScript::create_a_css_module_script(url.to_byte_string(), source_text, module_map_realm).release_value_but_fixme_should_propagate_errors();
 
-            // FIXME: 4. If mimeType is a JSON MIME type and moduleType is "json", then set moduleScript to the result of
-            //           creating a JSON module script given sourceText and moduleMapRealm.
+            // 4. If mimeType is a JSON MIME type and moduleType is "json", then set moduleScript to the result of
+            //    creating a JSON module script given sourceText and moduleMapRealm.
+            if (mime_type->is_json() && module_type == "json")
```
AtkinsSJ: Same point about dereferencing `mime_type`.

Libraries/LibWeb/HTML/Scripting/Fetching.cpp
```diff
@@ -705,12 +705,16 @@ void fetch_single_module_script(JS::Realm& realm,
...
         // FIXME: 5. If referrerPolicy is not the empty string, set options's referrer policy to referrerPolicy.
 
-        //  FIXME: 6. If mimeType's essence is "application/wasm" and moduleType is "javascript-or-wasm", then set moduleScript
-        //            to the result of creating a WebAssembly module script given bodyBytes, settingsObject, response's URL, and
-        //            options.
+        //  6. If mimeType's essence is "application/wasm" and moduleType is "javascript-or-wasm", then set moduleScript
+        //     to the result of creating a WebAssembly module script given bodyBytes, moduleMapRealm, response's URL, and
+        //     options.
+        // FIXME: Pass options.
+        if (mime_type->essence() == "application/wasm"sv && module_type == "javascript-or-wasm") {
```
AtkinsSJ: `mime_type` dereference again.

Libraries/LibWeb/HTML/Scripting/ModuleScript.cpp
skyz1: Ah I see... I updated this step and now it should work, but not sure if this is the best way to do it.

Libraries/LibJS/CyclicModule.cpp
```diff
@@ -417,7 +417,8 @@ ThrowCompletionOr<GC::Ref<Promise>> CyclicModule::evaluate(VM& vm)
     }
 
     // 11. Return capability.[[Promise]].
-    return as<Promise>(*m_top_level_capability->promise());
+    // AD-HOC: Return the promise capability and let the caller unwrap the promise
+    return GC::Ref<PromiseCapability>(*m_top_level_capability);
```
konradekk, line 421: why exactly this deviates? could you shed some light on this decision – it seems you go slightly against the spec in length, what is the benefit specifically? 🤔

(might be beneficial to provide rationale in one of the strategic places of such approach. at least that’s what missing for me at the moment, as it is not apparently clear from skimming through the code.)

skyz1: The problem is Step 8 of [running a module script](https://whatpr.org/html/9893/webappapis.html#run-a-module-script). This spec uses promise capabilities, while all other module related specs use promises. However, it seems like there is no way to go from a promise to a promise capability, so I have to return that and unwrap at the call site everywhere except in this place. Please correct me if I'm wrong, as I agree that returning a promise would be much cleaner.

I could maybe use `perform_then` to do something `upon_rejection`, but I am not sure.

konradekk: hmm… i guess everything’s fine if it does not switch from sync to async processing. 🤔 (which is not the case here, right?)

skyz1: What exactly do you mean by sync/async processing?

PR #6037, opened 2025-08-31
RequestServer: Allow \v and \f characters in response headers

veeti:
Adds two WPT passes in xhr/getallresponseheaders.htm and xhr/getresponseheader.any.html. However they don't work in the ladybird.py test runner, so I did not import them in this commit.

I am not sure if this is the right thing to do but I humbly submit it to be looked at by someone who knows.


rgret-dev:
You can write a simple test case using the HTTP echo server (search for `httpTestServer();` in the codebase for examples)
The test can be rolled up into the same commit.

Services/RequestServer/ConnectionFromClient.cpp
```diff
@@ -8,6 +8,7 @@
```
alimpfard: commit message nit:
This seems to be _removing_ \v\f instead of allowing it, we were already "allowing" them to pass through.
Something saying "trim \v\f as well" would probably be nicer.

Services/RequestServer/ConnectionFromClient.cpp
```diff
@@ -232,7 +237,7 @@ size_t ConnectionFromClient::on_header_received(void* buffer, size_t size, size_
 
     if (auto colon_index = header_line.find(':'); colon_index.has_value()) {
         auto name = header_line.substring_view(0, colon_index.value()).trim_whitespace();
-        auto value = header_line.substring_view(colon_index.value() + 1, header_line.length() - colon_index.value() - 1).trim_whitespace();
+        auto value = trim_header_whitespace(header_line.substring_view(colon_index.value() + 1, header_line.length() - colon_index.value() - 1));
```
alimpfard:
```suggestion
        auto value = header_line.substring_view(colon_index.value() + 1, header_line.length() - colon_index.value() - 1).trim(" \n\t\r"sv);
```

alimpfard (commented):
I don't so much mind the lack of tests, but please do add some if you can.

Services/RequestServer/ConnectionFromClient.cpp
alimpfard: (and remove the lambda)

Services/RequestServer/ConnectionFromClient.cpp
alimpfard: Ohhh, sorry, nevermind this. we're _now_ letting \v\f to pass through.

alimpfard:
It's irritatingly unspecified, but all mentions of whitespace in https://httpwg.org/specs/rfc9110.html are only to SP|HTAB (i.e. `" \t"sv`), so let's go with that (and please leave a comment explaining it)

veeti:
Okay, I managed to come up with a test. The CI failure seems like a freak accident...

PR #6039, opened 2025-08-31
LibWeb: Implement Blob constructor type normalization

mikiubo:
Normalize the type option in the Blob constructor according to the File API specification:

This makes Blob.type spec-compliant and fixes the WPT: xhr/setrequestheader-content-type.htm

### DIFF
--- a/Libraries/LibWeb/FileAPI/Blob.cpp
+++ b/Libraries/LibWeb/FileAPI/Blob.cpp
@@ -37,6 +37,16 @@ GC::Ref<Blob> Blob::create(JS::Realm& realm, ByteBuffer byte_buffer, String type
     return create(realm, move(byte_buffer), move(options));
 }
 
+GC::Ref<Blob> Blob::create(JS::Realm& realm, ByteBuffer byte_buffer, String type, TypeNormalization typeNormalization)
+{
+    BlobPropertyBag options = {
+        .type = move(type),
+        .endings = Bindings::EndingType::Transparent,
+        .typeNormalization = typeNormalization
+    };
+    return create(realm, move(byte_buffer), move(options));
+}
+
 // https://w3c.github.io/FileAPI/#convert-line-endings-to-native
 ErrorOr<String> convert_line_endings_to_native(StringView string)
 {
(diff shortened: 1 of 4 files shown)


Libraries/LibWeb/FileAPI/Blob.cpp
```diff
@@ -130,11 +130,27 @@ Blob::Blob(JS::Realm& realm)
...
+    // Step 2: Check for non-ASCII-printable characters
+    for (auto c : type.bytes_as_string_view()) {
+        if (c < 0x20 || c > 0x7E) {
+            m_type = String {};
+            return;
+        }
+    }
+
+    // Step 3: Normalize to ASCII lowercase
+    m_type = type.to_lowercase().release_value();
```
kennethmyhra: We have factory methods for constructor steps. You can find the steps you are mentioning here as FIXMEs starting roughly at line 200 https://github.com/LadybirdBrowser/ladybird/blob/master/Libraries/LibWeb/FileAPI/Blob.cpp#L198-L211

We have a method called `is_basic_latin()` which returns `false` if there are any characters falling outside the range U+0020 to U+007E, otherwise `true`.

mikiubo: > We have factory methods for constructor steps. You can find the steps you are mentioning here as FIXMEs starting roughly at line 200 https://github.com/LadybirdBrowser/ladybird/blob/master/Libraries/LibWeb/FileAPI/Blob.cpp#L198-L211

Sorry, but I don't understand one thing. The documentation specifies that those normalization steps must be done in the constructor. If I only put the normalization steps in that create method, won't all the other calls to create (for example, in XMLHttpRequest) skip the type normalization?

kennethmyhra: > If I only put the normalization steps in that create method, won't all the other calls to create (for example, in XMLHttpRequest) skip the type normalization?

This is solved by the refactoring done in #6064. Meaning you can apply those steps in the factory method as mentioned above

mikiubo: > This is solved by the refactoring done in #6064. Meaning you can apply those steps in the factory method as mentioned above


Thank you very much!
Now i think i rebase my pr, CI works fine 
Let me know if you notice anything wrong

kennethmyhra:
This might cause other regressions elsewhere, see this PR for some history why those steps where left as FIXMEs: https://github.com/LadybirdBrowser/ladybird/pull/1619

With just the diff below I am able to pass the test you have as passed in your patch. (I am not saying this is correct, it might just be patchwork and the actual fix should be elsewhere)

```diff
diff --git a/Libraries/LibWeb/FileAPI/Blob.cpp b/Libraries/LibWeb/FileAPI/Blob.cpp
index 48a332bfd43..9aa98c5cdae 100644
--- a/Libraries/LibWeb/FileAPI/Blob.cpp
+++ b/Libraries/LibWeb/FileAPI/Blob.cpp
@@ -215,6 +215,8 @@ GC::Ref<Blob> Blob::create(JS::Realm& realm, Optional<BlobPartsOrByteBuffer> con
         auto maybe_parsed_type = MimeSniff::MimeType::parse(options->type);
         if (maybe_parsed_type.has_value())
             type = maybe_parsed_type->serialized();
+
+        type = type.to_ascii_lowercase();
     }
```

mikiubo:
> ```diff
> type = type.to_ascii_lowercase();
> ```

You are completely right—I totally missed the regression on that WPT.
My idea was that if an override is specified in the XHR request, the lowercase normalization should not be applied, but I think I was wrong.

It seems the WPT is not yet aligned with the  spec—for example, both Firefox and Chrome fail the wpt you mentioned because of the lowercase issue.
In my opinion, maybe we should wait until the WPT is updated and only then implement the fix?

If that’s the case, I’ll close the PR.
Let me know what you think.

PR #6043, opened 2025-09-01
LibWeb: Remove Length's "auto" state

AtkinsSJ:
This has always been a bit of a hack. Initially it made sense as a lot of properties that accept a length also accept `auto`, but while convenient, that leads to problems: It's easy to forget to check if a length is `auto`, and places that don't accept it end up with an invalid state lurking in the type system, which makes things unclear.

So instead, we now have some types that explicitly hold `auto`:
- `LengthOrAuto`
- `LengthOrAutoOrCalculated`
- `LengthPercentageOrAuto`

Hopefully usage of them is clear enough.

The `LengthBox` commit ends up touching a lot of layout code, mostly because of the API name changes. Instead of `resolved()` it's `resolved_or_auto()` because it can return `auto`, and instead of `to_px()` we have `to_px_or_zero()` to reflect the fact that `auto` becomes `0px`. Both of these are the same behaviour as before, but these quirks were hidden away and not obvious. Hopefully the name changes make the layout code slightly easier to understand.

A nice bonus of all this is that we now interpolate `auto` background sizes more correctly, which passes something like 175 WPT subtests.


AtkinsSJ:
Changes:
- Rebaselined to fix conflicts.
- Redid the SourceSet changes, to only use `auto` where it's actually necessary.
- A couple of tweaks to the GridSize code.

PR #6046, opened 2025-09-01
LibWeb/Gamepad: Forward declare SDL components to fix Windows build

ayeteadoe:
The recent Gamepad PR added the SDL3 dependency, which caused the following build error on Windows:
```
In file included from C:\Users\ayeteadoe\Development\ladybird\Libraries\LibWeb\Gamepad\Gamepad.cpp:14:
In file included from C:\Users\ayeteadoe\Development\ladybird\Libraries\LibWeb\HTML\Navigator.h:14:
In file included from C:\Users\ayeteadoe\Development\ladybird\Libraries\LibWeb\HTML\NavigatorConcurrentHardware.h:10:
In file included from C:\Users\ayeteadoe\Development\ladybird\Libraries\LibCore\System.h:35:
C:\Users\ayeteadoe\Development\ladybird\Libraries\LibCore\SocketAddressWindows.h(21,9): error: 'WINAPI_FAMILY_PARTITION' macro redefined [-Werror,-Wmacro-redefined]
   21 | #define WINAPI_FAMILY_PARTITION(x) 1
      |         ^
C:\Program Files (x86)\Windows Kits\10\include\10.0.26100.0\shared\winapifamily.h(226,9): note: previous definition is here
  226 | #define WINAPI_FAMILY_PARTITION(Partitions)     (Partitions)
      |         ^
1 error generated.
```

We have to prevent from including any SDL headers in LibWeb headers. Otherwise on there will be transitive Windows.h includes that will re-declare some of our existing forward decls/defines in LibCore/SocketAddressWindows.h

### DIFF
--- a/Libraries/LibWeb/Gamepad/Gamepad.h
+++ b/Libraries/LibWeb/Gamepad/Gamepad.h
@@ -9,8 +9,8 @@
 
 #include <LibWeb/Bindings/GamepadPrototype.h>
 #include <LibWeb/Bindings/PlatformObject.h>
+#include <LibWeb/Gamepad/SDLGamepadForward.h>
 #include <LibWeb/HighResolutionTime/DOMHighResTimeStamp.h>
-#include <SDL3/SDL_gamepad.h>
 
 namespace Web::Gamepad {
 
--- a/Libraries/LibWeb/Internals/InternalGamepad.h
+++ b/Libraries/LibWeb/Internals/InternalGamepad.h
@@ -7,8 +7,7 @@
 #pragma once
 
 #include <LibWeb/Bindings/PlatformObject.h>
-#include <SDL3/SDL_gamepad.h>
-#include <SDL3/SDL_joystick.h>
+#include <LibWeb/Gamepad/SDLGamepadForward.h>
 
 namespace Web::Internals {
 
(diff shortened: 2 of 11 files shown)


ayeteadoe:
FYI @Lubrsi

Libraries/LibWeb/Gamepad/Gamepad.h
```diff
@@ -10,7 +10,11 @@
 #include <LibWeb/Bindings/GamepadPrototype.h>
 #include <LibWeb/Bindings/PlatformObject.h>
 #include <LibWeb/HighResolutionTime/DOMHighResTimeStamp.h>
-#include <SDL3/SDL_gamepad.h>
+
+typedef uint32_t Uint32;
```
Lubrsi: suggestion: Could these be added to a common header? e.g. SDLGamepadForward.h

Libraries/LibWeb/Internals/InternalGamepad.h
```diff
@@ -7,8 +7,55 @@
 #pragma once
 
 #include <LibWeb/Bindings/PlatformObject.h>
-#include <SDL3/SDL_gamepad.h>
-#include <SDL3/SDL_joystick.h>
+
+// NOTE: Unfortunately we have to just manually define these enums given forward declaring is not supported, so updating SDL versions that adjust the content of these enums may break things
+
+typedef enum SDL_GamepadButton {
```
Lubrsi: suggestion: It may be better to make buttons, axes and triggers return `Array<i32, [length]>` instead. It doesn't have to be super pretty with being an internal testing class not exposed to the web.

Libraries/LibWeb/SDLGamepadForward.h
```diff
@@ -0,0 +1,16 @@
+/*
```
Lubrsi: suggestion: This header should be in the Gamepad folder since it's all about Gamepad :^)

Lubrsi (approved):

PR #6047, opened 2025-09-01
LibWeb+LibWebView: Remove support for JS+CSS syntax highlighting

rgret-dev:
When viewing the page source, other browsers only highlight HTML related
tokens. This commit makes Ladybird behave the same to avoid unnecessary
memory overhead and additional work.

### DIFF
--- a/Libraries/LibWeb/HTML/SyntaxHighlighter/SyntaxHighlighter.cpp
+++ b/Libraries/LibWeb/HTML/SyntaxHighlighter/SyntaxHighlighter.cpp
@@ -6,9 +6,6 @@
  */
 
 #include <AK/Debug.h>
-#include <LibJS/SyntaxHighlighter.h>
-#include <LibJS/Token.h>
-#include <LibWeb/CSS/SyntaxHighlighter/SyntaxHighlighter.h>
 #include <LibWeb/HTML/Parser/HTMLTokenizer.h>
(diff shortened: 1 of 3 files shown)


gmta:
> unnecessary memory overhead and additional work.

But we're talking about viewing the source of a page, which is not a feature that many people use. The people that _do_ use it probably don't mind if slightly more resources were used to improve the source's readability?

rgret-dev:
A lot of pages have minified JS so readability is already poor. On google.com for example, the View Source tab uses over 1GB of memory without this patch and about ~400MB with it applied.


I personally don't see a point to do extra highlighting work when someone will reach for a proper editor if they need it.

gmta:
> On google.com for example, the View Source tab uses over 1GB of memory without this patch and about ~400MB with it applied.

Nice data point! But that's a lot of memory either way. I agree with you that highlighting minified JS/CSS seems relatively pointless, but then again, viewing a page's source has not seen much optimization and we might be able to do highlighting more lazily in the future. At the moment it's not broken or unused, so that's why I'm hesitant about removing this.

AtkinsSJ:
I often end up using our view-source instead of other browsers' just because we have highlighting and they don't. I'm very against removing this.

If it's problematic, maybe we could add a setting for more minimal highlighting. But making Ladybird more efficient is probably the way to go.

rgret-dev:
Fair enough. Closing.

PR #6049, opened 2025-09-02
LibWeb: Ensure Noah's Ark clause is called in the Parser

lpas:
This fixes some WPT tests for html Parsing

### DIFF
--- a/Libraries/LibWeb/HTML/Parser/ListOfActiveFormattingElements.cpp
+++ b/Libraries/LibWeb/HTML/Parser/ListOfActiveFormattingElements.cpp
@@ -9,6 +9,7 @@
 
 namespace Web::HTML {
 
+static constexpr size_t NoahsArkCapacity = 3;
 ListOfActiveFormattingElements::~ListOfActiveFormattingElements() = default;
 
 void ListOfActiveFormattingElements::visit_edges(JS::Cell::Visitor& visitor)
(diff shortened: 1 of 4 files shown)


Libraries/LibWeb/HTML/Parser/ListOfActiveFormattingElements.cpp
```diff
@@ -17,9 +18,46 @@ void ListOfActiveFormattingElements::visit_edges(JS::Cell::Visitor& visitor)
...
+        if (entry.element->local_name() == element.local_name()
+            && entry.element->namespace_uri() == element.namespace_uri()
+            && entry.element->attribute_list_size() == element.attribute_list_size())
+            possible_matches.append(entry);
+    }
+
+    if (possible_matches.size() < NoahsArkCapacity)
+        return;
+
+    element.for_each_attribute([&](auto& name, auto& value) {
```
Psychpsyo, line 39: ~~This only ensures that the possible matches have all the attributes that element has.~~
~~Should they also be rejected if they have any extra attributes that element doesn't have?~~

EDIT: Nevermind, that gets checked earlier.

Psychpsyo:
This doesn't compare the attributes "as they were when the elements were created by the parser", which breaks on HTML like this:
```html
<!DOCTYPE html>
<style>
  i:not(span i)::before {
    content: "<i>";
  }
</style>
<span>
  <i>
    <script>
      document.querySelector("i").dataset.foo = "bar";
    </script>
    <i data-foo="bar">
      <i data-foo="bar">
        <i data-foo="bar">
</span>
This sentence should be wrapped in 4 i-elements.
```
Maybe there should be a FIXME about that.

lpas:
@Psychpsyo you are right we currently don't keep the original elements. I Added a FIXME

Psychpsyo:
Github is telling me a review was requested from me...
In case it didn't just auto-do that: I'm not a maintainer, so I'm not sure I should actually be reviewing this. The FIXME looks good though.

Guess I'll put my stamp of approval, whatever it may be worth, lol

Psychpsyo (approved):

Libraries/LibWeb/HTML/Parser/ListOfActiveFormattingElements.cpp
```diff
@@ -17,9 +18,47 @@ void ListOfActiveFormattingElements::visit_edges(JS::Cell::Visitor& visitor)
         visitor.visit(entry.element);
 }
 
+// If there are already three elements in the list of active formatting elements after the last marker,
+// if any, or anywhere in the list if there are no markers, that have the same tag name, namespace, and attributes as element,
+// then remove the earliest such element from the list of active formatting elements.
+// For these purposes, the attributes must be compared as they were when the elements were created by the parser;
+// two elements have the same attributes if all their parsed attributes can be paired such that the two attributes
+// in each pair have identical names, namespaces, and values (the order of the attributes does not matter).
```
tcl3: Rather than adding this comment here, I would add the 2 spec steps from https://html.spec.whatwg.org/multipage/parsing.html#push-onto-the-list-of-active-formatting-elements to `ListOfActiveFormattingElements::add` below.

tcl3 (commented):
LGTM, just one nitpick. Sorry, this got overlooked!

PR #6051, opened 2025-09-02
Documentation: Better compatible build instructions for ubuntu distros

syr1ne:
i was facing issue with the ubuntu build instructions provided, on my linuxmint.
using the official documentation scripts of apt.kitware.com and apt.llvm.org will be a much easier and better choice, even in case of any updates made by kitware and llvm

### DIFF
--- a/Documentation/BuildInstructionsLadybird.md
+++ b/Documentation/BuildInstructionsLadybird.md
@@ -25,17 +25,9 @@ sudo apt install autoconf autoconf-archive automake build-essential ccache cmake
 
 - Recommendation: Install `CMake 3.25` or newer from [Kitware's apt repository](https://apt.kitware.com/):
 
-> [!NOTE]
-> This repository is Ubuntu-only
-
 ```bash
-# Add Kitware GPG signing key
-wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | gpg --dearmor - | sudo tee /usr/share/keyrings/kitware-archive-keyring.gpg >/dev/null
-
-# Optional: Verify the GPG key manually
-
-# Use the key to authorize an entry for apt.kitware.com in apt sources list
-echo "deb [signed-by=/usr/share/keyrings/kitware-archive-keyring.gpg] https://apt.kitware.com/ubuntu/ $(lsb_release -sc) main" | sudo tee /etc/apt/sources.list.d/kitware.list
+# Set up apt.kitware.com repository into APT package manager
+sudo bash -c "$(wget -O - https://apt.kitware.com/kitware-archive.sh)"
 
 # Update apt package list and install cmake
 sudo apt update -y && sudo apt install cmake -y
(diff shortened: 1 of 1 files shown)


Documentation/BuildInstructionsLadybird.md
```diff
@@ -25,17 +25,9 @@ sudo apt install autoconf autoconf-archive automake build-essential ccache cmake
...
 ```bash
-# Add Kitware GPG signing key
-wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | gpg --dearmor - | sudo tee /usr/share/keyrings/kitware-archive-keyring.gpg >/dev/null
-
-# Optional: Verify the GPG key manually
-
-# Use the key to authorize an entry for apt.kitware.com in apt sources list
-echo "deb [signed-by=/usr/share/keyrings/kitware-archive-keyring.gpg] https://apt.kitware.com/ubuntu/ $(lsb_release -sc) main" | sudo tee /etc/apt/sources.list.d/kitware.list
+# Set up apt.kitware.com repository into APT package manager
+sudo bash -c "$(wget -O - https://apt.kitware.com/kitware-archive.sh)"
```
gmta, line 30: 1. Could you use curl rather than wget?
2. If we're going to ask users to `sudo ..` anything from a potentially untrusted source, we better make sure it's exactly as we've tested it ourselves. Can we add a sha256 sum check before executing anything? And if the script changes too often, is there a versioned URL available?

syr1ne: 1. yea, we can use curl instead to wget, something like this: `sudo bash -c "$(curl -sL https://apt.kitware.com/kitware-archive.sh)"`
2. you are right, it is unsafe to use sudo on untrusted source's bash script, but this untrusted source is also providing packages that are being used, which eventually make those packages unsafe too. the point i wanna make is if any harm this untrusted source wanna make will eventually make it anyway through their packages and there is no straightforward  solution to it. and yes, for the integrity of bash script, we can use sha256 sum. i am not sure if the script changes too often, but there is no issue with that. the only thing script doing is just adding kitware repo to the APT package manager, thats it.

gmta: With packages it's harder to inject something malicious. Usually adding a repo involves adding the (known, validated) public GPG key, so that should be part of the script we checksum. Then, packages are signed using their private key and validated by the end user's apt install process.

The difference here is that a malicious actor could obtain write access to the repository, but is still missing the private key to properly sign the packages - which will then be rejected.

syr1ne: interesting. i didn't knew that. so we can use checksum for both of them:
```
curl -sL https://apt.kitware.com/kitware-archive.sh -o /tmp/kitware.sh && echo "4c16054d0a4808c9871e347dd1b10c1e4bbd3880b31235c06d6be91f86f4bf8f  /tmp/kitware.sh" | sha256sum -c - && sudo bash /tmp/kitware.sh

curl -sL https://apt.llvm.org/llvm.sh -o /tmp/llvm.sh && echo "70f8080fd855cb10cd79b140b8589a83f8a1f83d25b0bbf4436f55525425ac05  /tmp/llvm.sh" | sha256sum -c - && sudo bash /tmp/llvm.sh
```

alimpfard: 🤔 why are the spooky .sh's preferred or "better compatible"?
doing this makes it our problem to track upstream changes to those, and subsequently makes verifying them our responsibility (when e.g. the sha256 changes).

AtkinsSJ: Both apt.kitware.com and apt.llvm.org have installation instructions. Can we just say "follow the instructions at LINK" instead, and avoid having to maintain/take responsibility at all?

syr1ne: these bash scripts are better because i was facing an issue on my linuxmint with buildinstructions provided which may only work fine with ubuntu. in my idea, this seems simplest way and i dont think these scripts will gonna update frequently. rest you can suggest any other better way.

alimpfard: what sort of problem exactly?
Either way I think what Sam suggested is the best - ask the user to follow the instructions given by kitware/llvm.

syr1ne: image
it was giving me 404

alimpfard: ah I guess lsb-release gives you mint's release name, not the ubuntu it's based on.

syr1ne: right!

PR #6052, opened 2025-09-02
LibIDL: Allow `iterable<T>` to be defined before property getter 

### LINKED ISSUE #5940: IDLParser requires that `iterable<foo>` comes after the indexed property getter's definition
I've come across multiple IDL definitions that put the `iterable` line near the top of the interface. [For example:](https://drafts.css-houdini.org/css-typed-om-1/#cssunparsedvalue)

```webidl
interface CSSUnparsedValue : CSSStyleValue {
    constructor(sequence<CSSUnparsedSegment> members);
    iterable<CSSUnparsedSegment>;
    readonly attribute unsigned long length;
    getter CSSUnparsedSegment (unsigned long index);
    setter undefined (unsigned long index, CSSUnparsedSegment val);
};
```

IDLParser throws an error when it sees this, thinking there is no indexed property getter because it hasn't seen it yet.

The temporary solution is to move the `iterable` line further down, but ideally IDLParser would be smart enough to understand it in either order.

tcl3:
Previously, an error would occur if an `iterable<T>` was defined before the required value iterator. An is now also reported if an`iterable<T>` is defined before a pair iterator, previously this error would only be reported if the iterable was defined after the pair iterator.

Fixes #5940

### DIFF
--- a/Libraries/LibIDL/IDLParser.cpp
+++ b/Libraries/LibIDL/IDLParser.cpp
@@ -518,18 +518,14 @@ void Parser::parse_iterable(Interface& interface)
     assert_specific('<');
     auto first_type = parse_type();
     if (lexer.next_is(',')) {
-        if (interface.supports_indexed_properties())
-            report_parsing_error("Interfaces with a pair iterator must not support indexed properties."sv, filename, input, lexer.tell());
-
         assert_specific(',');
         consume_whitespace();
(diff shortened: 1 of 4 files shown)


AtkinsSJ (approved):
Nice! That's a lot simpler than I expected.

PR #6053, opened 2025-09-02
LibCore: Use IOCPs for the event loop on windows

R-Goc:
This PR introduces an IOCP based implementation of the event loop. Currently only one thread is used to wait. The number of threads to use and the number of entries to remove at once will probably need to be tuned. For a more exhuastive description see the commit message. 
As a drive by use a stack buffer to format windows errors as allocation fails during program destruction.

### DIFF
--- a/Libraries/LibCore/EventLoopImplementationWindows.cpp
+++ b/Libraries/LibCore/EventLoopImplementationWindows.cpp
@@ -2,57 +2,104 @@
  * Copyright (c) 2023, Andreas Kling <andreas@ladybird.org>
  * Copyright (c) 2024-2025, stasoid <stasoid@yahoo.com>
  * Copyright (c) 2025, ayeteadoe <ayeteadoe@gmail.com>
+ * Copyright (c) 2025, Ryszard Goc <ryszardgoc@gmail.com>
  *
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
+#include <AK/Assertions.h>
+#include <AK/Diagnostics.h>
 #include <AK/HashMap.h>
+#include <AK/NonnullOwnPtr.h>
+#include <AK/Windows.h>
 #include <LibCore/EventLoopImplementationWindows.h>
 #include <LibCore/Notifier.h>
 #include <LibCore/ThreadEventQueue.h>
 #include <LibCore/Timer.h>
+#include <LibThreading/Mutex.h>
 
-#include <AK/Windows.h>
-
-struct Handle {
+struct OwnHandle {
     HANDLE handle = NULL;
 
-    explicit Handle(HANDLE h = NULL)
+    explicit OwnHandle(HANDLE h = NULL)
         : handle(h)
     {
     }
-    Handle(Handle&& h)
+
+    OwnHandle(OwnHandle&& h)
     {
         handle = h.handle;
         h.handle = NULL;
     }
-    void operator=(Handle&& h)
+
+    // This operation can only be done when handle is NULL
+    OwnHandle& operator=(OwnHandle&& other)
     {
         VERIFY(!handle);
-        handle = h.handle;
-        h.handle = NULL;
+        if (this == &other)
+            return *this;
+        handle = other.handle;
+        other.handle = NULL;
+        return *this;
     }
-    ~Handle()
+
+    ~OwnHandle()
     {
         if (handle)
             CloseHandle(handle);
     }
 
-    bool operator==(Handle const& h) const { return handle == h.handle; }
+    bool operator==(OwnHandle const& h) const { return handle == h.handle; }
     bool operator==(HANDLE h) const { return handle == h; }
 };
 
 template<>
-struct Traits<Handle> : DefaultTraits<Handle> {
-    static unsigned hash(Handle const& h) { return Traits<HANDLE>::hash(h.handle); }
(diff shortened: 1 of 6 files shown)


R-Goc:
There is some additional complexity here as we are doing something very unintended. Io completion ports as the name suggest notify on completion. Using the nt kernel apis we are able to wait on an event as well. As such we are able to use WSAEventSelect to associate an event with a socket and get a select like api that notifies us when a socket is readable. However, doing this is not possible for files, so this model cannot be used for file IO. To implement async file IO we would need to switch to a completion notification model but that seems impossible for all platforms and even if it was there would be a very large maintenance burden associated with maintaining multiple implementations. Some more details both about the implementation and how bad moving to a completion model would be are in #ui-windows on discord.

Libraries/LibCore/EventLoopImplementationWindows.cpp
```diff
@@ -65,22 +156,50 @@ struct ThreadData {
...
-    HashMap<Handle, EventLoopTimer> timers;
-    HashMap<Handle, Notifier*> notifiers;
+    Handle iocp;
+
+    // These are only used to register and unregister. The event loop doesn't access these.
+    // Each thread has its own timers, and a wake event.
+    HashMap<intptr_t, NonnullOwnPtr<EventLoopTimer>> timers;
+    HashMap<Notifier*, NonnullOwnPtr<EventLoopNotifier>> notifiers;
 
     // The wake event is used to notify another event loop that someone has called wake().
```
R-Goc: Maybe this whole wake event thing should just be replaced by the wake function posting a completion status to the queue.

Libraries/LibCore/EventLoopImplementationWindows.cpp
```diff
@@ -176,47 +298,95 @@ static int notifier_type_to_network_event(NotificationType type)
...
-        ThreadData::the()->timers.remove(reinterpret_cast<HANDLE>(timer_id));
+    if (auto* thread_data = ThreadData::the()) {
+        Threading::MutexLocker lock(thread_data->mutex);
+        auto maybe_timer = thread_data->timers.get(timer_id);
+        if (!maybe_timer.has_value())
+            return;
+        auto* timer = maybe_timer.value();
+        // NOTE: Should we be removing a signalled packet that has yet to be dequed or not? Currently we do, but it can be changed.
+        ntdll.NtCancelWaitCompletionPacket(timer->wait_packet.handle, TRUE);
+        thread_data->timers.remove(timer_id);
```
Zaggy1024: Can we use `HashMap::find()` here to avoid a double lookup? (Actually, perhaps `HashMap::take()` is even more ideal.)

Also, is there a reason you suspect that removing a signaled timer would be a problem? I don't think it's reasonable for a timer to fire after it's been unregistered, I suspect the only way that may be _somewhat_ reasonable is if it's being unregistered from a different thread.

Libraries/LibCore/EventLoopImplementationWindows.cpp
Zaggy1024: (This comment appears to have been addressed?)

R-Goc: Yes, I didn't look at the PR on GitHub for a while

Libraries/LibCore/EventLoopImplementationWindows.cpp
R-Goc: Well the case I was considering is that an event was signalled(either for a timer or notifier), the event loop thread is blocked on some work, and another thread removes the notifier. We are blocked, so the signaled event didn't get dequeued and processed. For example a read was completed on a socket. So not sure what I should be doing there. For timers I remove them as it is simpler to handle that way since periodic timer packets get associated with the even loop again when dequeued. For notifiers it currently doesn't dequeue them if signalled.

Zaggy1024: Hmm, if my understanding is correct, then maybe we should have an atomic flag on `CompletionPacket` to signify that it should be removed upon the next `pump()` (in a loop before waiting, or upon being handled)? That way, we can (I believe) be sure that we won't have a UAF when handling `EventLoopTimer` or `EventLoopNotifier`.

R-Goc: Well if we remove the signaled packet that is not an issue since it won't even be dequeued by GetQueuedCompletionStatus. Thinking about the logic here, I think removing signaled packets is the correct thing to do, since the caller of unregister_notifer/timer expects it to be removed since we don't return an ErrorOr. So the caller getting a notfication when it isn't expecting one would be wrong no?

R-Goc:
Before merging this it still needs implementing accessing another thread's event loop. Edit: not applicable anymore

Libraries/LibCore/EventLoopImplementationWindows.cpp
Zaggy1024: Agreed. The thing I'm concerned about is whether, in the time between calling `GetQueuedCompletionStatusEx()` and using the raw `CompletionPacket*`, a thread could remove the `OwnPtr<CompletionPacket>` from the map, causing it to be destroyed before it is used. This could be solved by atomically ref-counting it, but I think that's way overkill considering that only the `pump()` method needs to hold onto the `CompletionPacket*` after it's been unregistered.

PR #6056, opened 2025-09-02
LibWeb: Improve inspector overlays

InvalidUsernameException:
Makes the inspector overlays work for SVG elements and improves their behavior in a couple other edge cases. See individual commits for more details. Some examples:

Masking before/after:

https://github.com/user-attachments/assets/6353daf8-aa8d-411e-b867-08ab781e24a2

https://github.com/user-attachments/assets/52d302bc-aeac-440a-8b4a-b82a08a947c0

SVG:

https://github.com/user-attachments/assets/e1ed87d5-1b9b-485c-9061-093acd16f4e2

### DIFF
--- a/Libraries/LibWeb/Painting/Paintable.cpp
+++ b/Libraries/LibWeb/Painting/Paintable.cpp
@@ -6,6 +6,8 @@
  */
 
 #include <LibWeb/DOM/Document.h>
+#include <LibWeb/Painting/DisplayListRecorder.h>
+#include <LibWeb/Painting/DisplayListRecordingContext.h>
 #include <LibWeb/Painting/Paintable.h>
 #include <LibWeb/Painting/PaintableBox.h>
 #include <LibWeb/Painting/StackingContext.h>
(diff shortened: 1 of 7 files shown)


Libraries/LibWeb/Painting/PaintableBox.cpp
```diff
@@ -947,7 +947,7 @@ void paint_text_fragment(DisplayListRecordingContext& context, TextPaintable con
         auto fragment_enclosing_device_rect = context.enclosing_device_rect(fragment_absolute_rect).to_type<int>();
 
         if (paintable.document().highlighted_layout_node() == &paintable.layout_node())
```
gmta: Couldn't you just add `|| context.should_show_line_box_borders()` here at this point and remove the separate check in `PaintableWithLines::paint()`?

Libraries/LibWeb/Painting/StackingContext.cpp
gmta, line 1: Commit message typo: "insted"

Libraries/LibWeb/Painting/Paintable.cpp
```diff
@@ -158,6 +160,44 @@ StackingContext* Paintable::enclosing_stacking_context()
...
+                };
+                context.display_list_recorder().push_stacking_context(push_stacking_context_params);
+            }
+        }
+    }
+
+    paint_inspector_overlay_internal(context);
+
+    for (auto const* paintable : self_and_ancestors) {
+        if (auto const* box = as_if<PaintableBox>(paintable); box && box->stacking_context()) {
```
gmta: You should remove `; box && box->stacking_context()` here since you check it again on the next line.

Libraries/LibWeb/Painting/PaintableBox.cpp
InvalidUsernameException: Yeah, I guess I can. `paint_text_fragment_debug_highlight()` gets two different call-sites again in the last commit, but having the line-box code here seems a bit nicer anyway.

Libraries/LibWeb/Painting/Paintable.cpp
InvalidUsernameException: Thx, that must have been a leftover from an earlier version of the code.

PR #6057, opened 2025-09-02
Trusted types integration dom

tete17:
This should cover most of the integration with TrustedTypes and the DOM apis

In particular this is the work required to implement Section 4.2 of the TrustedTypes Spec.

It follows the MR of https://github.com/whatwg/dom/pull/1268 as close as we can given it is not merged yet.

There is 2 special points we need to be aware of:
 * The integration with the `execCommand` is ad-hoc since that standard was abandoned. There are some WPT test to cover them and my goal was to make the test pass and what made sense in my head.
 * A lot of WPT test relay on the WebSocket implementation to be properly hooked with the Fetch api. In particular it is used to force a connect-src violation at the end of the test. I was able to make the test work by replacing it with a simple fetch call in an auxiliary file `csp-violations.js` but it means we won't see them in our official WPT score until we solve https://github.com/LadybirdBrowser/ladybird/blob/master/Libraries/LibWeb/WebSockets/WebSocket.cpp#L191. I gave it quick try with not much hope but it is clearly outside the scope of this MR.


Libraries/LibWeb/DOM/Attr.cpp
```diff
@@ -69,17 +71,43 @@ void Attr::set_owner_element(Element* owner_element)
...
     // 1. If attribute’s element is null, then set attribute’s value to value.
     if (!owner_element()) {
         m_value = move(value);
     }
-    // 2. Otherwise, change attribute to value.
+    // 2. Otherwise:
     else {
-        change_attribute(move(value));
+        // 1. Let originalElement be attribute’s element.
+        auto const original_element = owner_element();
```
Lubrsi: suggestion: To make it clear this is a pointer type:
```suggestion
        auto const* original_element = owner_element();
```

Libraries/LibWeb/DOM/Attr.cpp
```diff
@@ -69,17 +71,43 @@ void Attr::set_owner_element(Element* owner_element)
...
+
+        // 2. Let verifiedValue be the result of calling get Trusted Types-compliant attribute value with
+        //    attribute’s local name, attribute’s namespace, this, and value.
+        auto const verified_value = TRY(TrustedTypes::get_trusted_types_compliant_attribute_value(
+            owner_element()->local_name(),
```
Lubrsi: issue: This should look at Attr's local_name instead of Element's local_name
```suggestion
            local_name(),
```

Libraries/LibWeb/DOM/Attr.cpp
```diff
@@ -69,17 +71,43 @@ void Attr::set_owner_element(Element* owner_element)
...
+            namespace_uri().has_value() ? Utf16String::from_utf8(namespace_uri().value()) : Optional<Utf16String>(),
+            *owner_element(),
```
Lubrsi: todo:
```suggestion
            *element,
```

Libraries/LibWeb/DOM/Attr.cpp
```diff
@@ -69,17 +71,43 @@ void Attr::set_owner_element(Element* owner_element)
...
     }
-    // 2. Otherwise, change attribute to value.
+    // 2. Otherwise:
     else {
-        change_attribute(move(value));
+        // 1. Let originalElement be attribute’s element.
+        auto const original_element = owner_element();
+
+        // 2. Let verifiedValue be the result of calling get Trusted Types-compliant attribute value with
+        //    attribute’s local name, attribute’s namespace, this, and value.
```
Lubrsi: issue: `this` should be `element` instead.

Libraries/LibWeb/DOM/Attr.cpp
Lubrsi: todo: It seems the name of this variable was renamed to `element` in the PR.

Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -641,41 +643,56 @@ GC::Ptr<Selection::Selection> Document::get_selection() const
...
 
     // 3. For each value of text:
     for (auto const& value : text) {
-        // FIXME: 1. If value is a TrustedHTML object, then append value's associated data to string.
-
-        // 2. Otherwise, append value to string.
-        string.append(value);
+        string.append(value.visit(
+                               // 1. If value is a TrustedHTML object, then append value's associated data to string.
+                               [](const GC::Root<TrustedTypes::TrustedHTML>& value) { return value->to_string(); },
```
Lubrsi: style: west const to east const

Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -6342,10 +6359,18 @@ void Document::parse_html_from_a_string(StringView html)
...
+WebIDL::ExceptionOr<GC::Root<DOM::Document>> Document::parse_html_unsafe(JS::VM& vm, TrustedTypes::TrustedHTMLOrString const& html)
 {
     auto& realm = *vm.current_realm();
-    // FIXME: 1. Let compliantHTML to the result of invoking the Get Trusted Type compliant string algorithm with TrustedHTML, this's relevant global object, html, "Document parseHTMLUnsafe", and "script".
+
+    // 1. Let compliantHTML to the result of invoking the Get Trusted Type compliant string algorithm with
+    //    TrustedHTML, this's relevant global object, html, "Document parseHTMLUnsafe", and "script".
+    auto const compliant_html = TRY(TrustedTypes::get_trusted_type_compliant_string(
+        TrustedTypes::TrustedTypeName::TrustedHTML,
+        realm.global_object(),
```
Lubrsi: suggestion: I'm not sure if this has any functional difference:
```suggestion
        HTML::relevant_global_object(*this),
```

Libraries/LibWeb/DOM/Document.idl
```diff
@@ -29,6 +29,7 @@
 #import <ViewTransition/ViewTransition.idl>
 #import <XPath/XPathResult.idl>
 #import <XPath/XPathExpression.idl>
+#import <TrustedTypes/TrustedHTML.idl>
```
Lubrsi: style: This should come after `#import <Selection/Selection.idl>` to maintain alphabetical order

Libraries/LibWeb/DOM/Element.idl
```diff
@@ -52,8 +54,8 @@ interface Element : Node {
     sequence<DOMString> getAttributeNames();
     DOMString? getAttribute(DOMString qualifiedName);
     DOMString? getAttributeNS([FlyString] DOMString? namespace, [FlyString] DOMString localName);
-    [CEReactions] undefined setAttribute(DOMString qualifiedName, DOMString value);
-    [CEReactions] undefined setAttributeNS([FlyString] DOMString? namespace , [FlyString] DOMString qualifiedName , DOMString value);
+    [CEReactions] undefined setAttribute(DOMString qualifiedName, (TrustedType or DOMString) value);
+    [CEReactions] undefined setAttributeNS([FlyString] DOMString? namespace , [FlyString] DOMString qualifiedName , (TrustedType or DOMString) value);
```
Lubrsi: question: Should `DOMString` become `Utf16DOMString` now for these APIs?

Libraries/LibWeb/DOM/Element.cpp
```diff
@@ -349,14 +362,25 @@ WebIDL::ExceptionOr<QualifiedName> validate_and_extract(JS::Realm& realm, Option
...
+// https://whatpr.org/dom/1268.html#dom-element-setattributens
+WebIDL::ExceptionOr<void> Element::set_attribute_ns(Optional<FlyString> const& namespace_, FlyString const& qualified_name, Variant<GC::Root<TrustedTypes::TrustedHTML>, GC::Root<TrustedTypes::TrustedScript>, GC::Root<TrustedTypes::TrustedScriptURL>, String> const& value)
 {
     // 1. Let (namespace, prefix, localName) be the result of validating and extracting namespace and qualifiedName given "element".
     auto extracted_qualified_name = TRY(validate_and_extract(realm(), namespace_, qualified_name, ValidationContext::Element));
 
-    // 2. Set an attribute value for this using localName, value, and also prefix and namespace.
-    set_attribute_value(extracted_qualified_name.local_name(), value, extracted_qualified_name.prefix(), extracted_qualified_name.namespace_());
+    // 2. Let verifiedValue be the result of calling get Trusted Types-compliant attribute value
+    //    with localName, namespace, element, and value.
```
Lubrsi: todo: A comment should be left on the PR stating `element` here should instead be `this`

Libraries/LibWeb/DOM/Element.cpp
```diff
@@ -2062,15 +2093,22 @@ WebIDL::ExceptionOr<GC::Ref<DOM::DocumentFragment>> Element::parse_fragment(Stri
...
 }
 
 // https://html.spec.whatwg.org/multipage/dynamic-markup-insertion.html#dom-element-outerhtml
-WebIDL::ExceptionOr<void> Element::set_outer_html(String const& value)
+WebIDL::ExceptionOr<void> Element::set_outer_html(TrustedTypes::TrustedHTMLOrString const& value)
 {
-    // 1. FIXME: Let compliantString be the result of invoking the Get Trusted Type compliant string algorithm with TrustedHTML, this's relevant global object, the given value, "Element outerHTML", and "script".
+    // 1. Let compliantString be the result of invoking the Get Trusted Type compliant string algorithm with
+    //    TrustedHTML, this's relevant global object, the given value, "Element outerHTML", and "script".
+    auto const compliant_html = TRY(TrustedTypes::get_trusted_type_compliant_string(
```
Lubrsi: suggestion: This should be `compliant_string` to match the spec text.

Libraries/LibWeb/TrustedTypes/TrustedTypePolicy.cpp
```diff
@@ -340,4 +340,51 @@ WebIDL::ExceptionOr<Utf16String> get_trusted_type_compliant_string(TrustedTypeNa
...
+{
+    // 1. If attributeNs is the empty string, set attributeNs to null.
+    if (attribute_ns.has_value() && attribute_ns.value().is_empty())
+        attribute_ns.clear();
+
+    // 2. Set attributeData to the result of Get Trusted Type data for attribute algorithm, with the following arguments:
+    //    element
+    //    attributeName
+    //    attributeNs
+    auto const attribute_data = get_trusted_type_data_for_attribute(Utf16String::from_utf8(element.node_name()), Utf16String::from_utf8(attribute_name), attribute_ns);
```
Lubrsi: issue: I'm not sure if this will ever match a `TrustedTypeData`, because `get_trusted_type_data_for_attribute` expects the interface name of the element instead of the node name.

Libraries/LibWeb/TrustedTypes/TrustedTypePolicy.cpp
```diff
@@ -340,4 +340,51 @@ WebIDL::ExceptionOr<Utf16String> get_trusted_type_compliant_string(TrustedTypeNa
...
+        // 2. Assert: newValue is TrustedHTML or TrustedScript or TrustedScriptURL.
+        VERIFY(new_value.has<GC::Root<TrustedHTML>>() || new_value.has<GC::Root<TrustedScript>>() || new_value.has<GC::Root<TrustedScriptURL>>());
+
+        // 3. Return value’s associated data.
+        // FIXME: This is badly worded in the spec it should say "Return stringified newvalues's"
+        return new_value.downcast<TrustedType>().visit([](auto& value) { return value->to_string(); });
+    }
+
+    // 4. Let expectedType be the value of the fourth member of attributeData.
+    auto const expectedType = attribute_data->trusted_type;
```
Lubrsi: style:
```suggestion
    auto const expected_type = attribute_data->trusted_type;
```

Libraries/LibWeb/TrustedTypes/TrustedTypePolicy.cpp
```diff
@@ -340,4 +340,51 @@ WebIDL::ExceptionOr<Utf16String> get_trusted_type_compliant_string(TrustedTypeNa
...
+
+    // 6. Return the result of executing Get Trusted Type compliant string with the following arguments:
+    //      expectedType
+    //      newValue as input
+    //      element’s node document’s relevant global object as global
+    //      sink
+    //      'script' as sinkGroup
+    return get_trusted_type_compliant_string(
+        expectedType,
+        element.realm().global_object(),
```
Lubrsi: suggestion: To match the spec text:
```suggestion
        HTML::relevant_global_object(element.document()),
```

Libraries/LibWeb/DOM/Document.cpp
tete17: I am afraid this is a static method so no `this`

Libraries/LibWeb/DOM/Element.cpp
tete17: https://github.com/whatwg/dom/pull/1268/files#r2425834862

Libraries/LibWeb/DOM/Document.cpp
AtkinsSJ: Could you open a spec issue for that? That way we know what should be used instead of `this`.

Libraries/LibWeb/DOM/Element.idl
tete17: I thought It would be harder but turns out to be simpler than expected :)

Libraries/LibWeb/TrustedTypes/TrustedTypePolicy.cpp
tete17: I will be honest I am not 100% convinced neither but the `node_name` implementation returns `m_qualified_name` which seems to fit the description right?

Libraries/LibWeb/DOM/Document.cpp
tete17: https://github.com/whatwg/html/issues/11778

Should we merge it as is or wait until they come back?

Libraries/LibWeb/TrustedTypes/TrustedTypePolicy.cpp
Lubrsi: Not quite, QualifiedName gives you node name, prefix and namespace. For example, for `<iframe>`, it will give you `"iframe"` and not '"HTMLIFrameElement"`, which `get_trusted_type_data_for_attribute` expects to have instead. The same translation `get_attribute_type` uses can be used here as well.
https://github.com/LadybirdBrowser/ladybird/blob/d06517179142f3a901a93ea93ac5a5cd09eb13e3/Libraries/LibWeb/TrustedTypes/TrustedTypePolicyFactory.cpp#L51-L63

Libraries/LibWeb/DOM/Document.cpp
tete17: It seems they want to replace it for [current global object](https://html.spec.whatwg.org/multipage/webappapis.html#current-global-object). I updated the code and left a FIXME referencing the issue upstream

Libraries/LibWeb/TrustedTypes/TrustedTypePolicy.cpp
tete17: You are absolutely right.

I extracted the little hack I build a few commits into a function so it can be reused.

This change actually gave us 7 WTP's test more :)

PR #6058, opened 2025-09-02
Trusted types integration with Scripts

tete17:
This should cover most of the integration with TrustedTypes and the esecution of JS

In particular this is the work required to implement Section [4.1.2](https://w3c.github.io/trusted-types/dist/spec/#enforcement-in-scripts) of the TrustedTypes Spec.

There is a special point we need to be aware of, a lot of WPT test relay on the WebSocket implementation to be properly hooked with the Fetch api. In particular it is used to force a connect-src violation at the end of the test. I was able to make the test work by replacing it with a simple fetch call in an auxiliary file csp-violations.js but it means we won't see them in our official WPT score until we solve https://github.com/LadybirdBrowser/ladybird/blob/master/Libraries/LibWeb/WebSockets/WebSocket.cpp#L191. I gave it quick try with not much hope but it is clearly outside the scope of this MR.

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLScriptElement.h
+++ b/Libraries/LibWeb/HTML/HTMLScriptElement.h
@@ -76,6 +76,8 @@ public:
 
     virtual WebIDL::ExceptionOr<void> cloned(Node&, bool) const override;
 
+    void set_string_text(Utf16String const& value) { m_script_text = value; }
+
 protected:
     // https://html.spec.whatwg.org/multipage/urls-and-fetching.html#implicitly-potentially-render-blocking
     virtual bool is_implicitly_potentially_render_blocking() const override;
--- a/Libraries/LibWeb/HTML/Parser/HTMLParser.cpp
+++ b/Libraries/LibWeb/HTML/Parser/HTMLParser.cpp
@@ -3268,6 +3268,10 @@ void HTMLParser::handle_text(HTMLToken& token)
         // Increment the parser's script nesting level by one.
         increment_script_nesting_level();
 
+        // https://w3c.github.io/trusted-types/dist/spec/#setting-slot-values-from-parser
+        // Set script’s script text value to its child text content.
+        script->set_string_text(script->child_text_content());
+
         // If the active speculative HTML parser is null, then prepare the script element script.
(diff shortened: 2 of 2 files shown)


Libraries/LibWeb/HTML/HTMLScriptElement.cpp
```diff
@@ -639,6 +641,26 @@ void HTMLScriptElement::unmark_as_parser_inserted(Badge<DOM::Range>)
...
+    //    TrustedScript, this’s relevant global object, the given value, HTMLScriptElement text, and script.
+    auto const value = TRY(get_trusted_type_compliant_string(
+        TrustedTypes::TrustedTypeName::TrustedScript,
+        HTML::relevant_global_object(*this),
+        text,
+        TrustedTypes::InjectionSink::HTMLScriptElementtext,
+        TrustedTypes::Script.to_string()));
+
+    // 2. Set this’s script text value to the given value.
+    m_string_text = value;
```
AtkinsSJ: I think calling the field `m_script_text` to match the spec term is a little less confusing.

Tests/LibWeb/Text/input/usvstring-url-reflection.html
```diff
@@ -16,7 +16,6 @@
             { "link": "href" },
             { "object": "codeBase" },
             { "object": "data" },
-            { "script": "src" },
```
AtkinsSJ, line 19: Can you write in the commit description why this needed to be removed?

AtkinsSJ (changes requested):

gmta (approved):

shlyakpavel:
This PR seems to have broken Telegram Web (version A) in Ladybird, check https://github.com/LadybirdBrowser/ladybird/issues/2679#issuecomment-3315247847

tete17:
I am so sorry @shlyakpavel 

Given how security concious telegram is maybe this MR triggered them to start using trusted types and that broke some other unimplemented part. 

Given the logs you posted i cant be for sure what or where.

I am on long holidays for 2 weeks or so and i wont have access to a proper machine until then, feel free to revere the MR if you feel is appropiate and i can take a look when i come back 😁

shlyakpavel:
@tete17 check #6253 and #6255

PR #6059, opened 2025-09-03
LibWeb: Do not wrap nested editing hosts for insertParagraph

gmta:
If a node is an editing host, but it is itself contained in an editable or another editing host, it should not be subject to having itself wrapped in a new container when the `insertParagraph` command is executed. Both Chrome and Firefox duplicate the existing editing host.

No change in the WPT `editing` category.


gmta:
Closing this in favor of #6060.

PR #6061, opened 2025-09-03
LibWeb: Add `::slotted()` pseudo-element support

kalenikaliaksandr:
Implements `::slotted()` to enough extent we could pass the imported WPT test and make substantial layout correctness improvement on https://www.rottentomatoes.com/

Before:
CleanShot 2025-09-03 at 17 05 14@2x

After:
CleanShot 2025-09-03 at 17 05 45@2x


AtkinsSJ (approved):
`::slotted()` is a bit mind-bending, but everything here looks correct to me. Sorry about the completely unhelpful advice before! :sweat_smile:

PR #6065, opened 2025-09-04
LibWeb: Check for Svg & MathML tags in stack of open elements scope

lpas:
list & button scope need to check SVG & MathML elements besides the list from s_base_list https://html.spec.whatwg.org/multipage/parsing.html#has-an-element-in-the-specific-scope


Libraries/LibWeb/HTML/Parser/StackOfOpenElements.h
```diff
@@ -63,7 +63,7 @@ class StackOfOpenElements {
     void visit_edges(JS::Cell::Visitor&);
 
 private:
-    bool has_in_scope_impl(FlyString const& tag_name, Vector<FlyString> const&) const;
+    bool has_in_scope_impl(FlyString const& tag_name, Vector<FlyString> const&, bool checkMathAndSvg = false) const;
```
AtkinsSJ: A couple of code-style things:
- We use `snake_case` for variable names.
- We like to use named enum classes instead of bools for function parameters, as it makes it easier to see what they do.

So something like:

```c++
enum class CheckMathAndSVG : u8 {
    No,
    Yes,
};
bool has_in_scope_impl(FlyString const& tag_name, Vector<FlyString> const&, CheckMathAndSVG = CheckMathAndSVG::No) const;
```

lpas:
Hey, thanks for the review. I addressed your comments and updated the code. 

> Another option would be to pass the namespace of each element as part of the `list` parameter, but this works too.

I found this easier to read but happy to change this if needed.

Libraries/LibWeb/HTML/Parser/StackOfOpenElements.cpp
```diff
@@ -20,20 +22,24 @@ void StackOfOpenElements::visit_edges(JS::Cell::Visitor& visitor)
...
+bool StackOfOpenElements::has_in_scope_impl(FlyString const& tag_name, Vector<FlyString> const& list, CheckMathAndSVG check_math_and_svg) const
 {
     for (auto const& element : m_elements.in_reverse()) {
         if (element->local_name() == tag_name)
             return true;
         if (list.contains_slow(element->local_name()))
             return false;
+        if (check_math_and_svg == CheckMathAndSVG::Yes && element->namespace_uri() == Namespace::SVG && element->local_name().is_one_of(SVG::TagNames::foreignObject, SVG::TagNames::desc, SVG::TagNames::title))
+            return false;
+        if (check_math_and_svg == CheckMathAndSVG::Yes && element->namespace_uri() == Namespace::MathML && element->local_name().is_one_of(MathML::TagNames::mi, MathML::TagNames::mo, MathML::TagNames::mn, MathML::TagNames::ms, MathML::TagNames::mtext))
```
gmta: Isn't this missing MathML's `annotation-xml`?

Libraries/LibWeb/HTML/Parser/StackOfOpenElements.h
```diff
@@ -63,7 +63,12 @@ class StackOfOpenElements {
     void visit_edges(JS::Cell::Visitor&);
 
 private:
-    bool has_in_scope_impl(FlyString const& tag_name, Vector<FlyString> const&) const;
+    enum class CheckMathAndSVG : u8 {
+        No,
+        Yes,
+    };
+
+    bool has_in_scope_impl(FlyString const& tag_name, Vector<FlyString> const&, CheckMathAndSVG = CheckMathAndSVG::No) const;
```
gmta: I'd rather not have the default `CheckMathAndSVG::No` value - it makes it clearer in the code what we're checking against.

Libraries/LibWeb/HTML/Parser/StackOfOpenElements.cpp
lpas: oh, wow good catch! thanks

PR #6066, opened 2025-09-04
LibWeb: Simplify `SourceSet::normalize_source_densities`

Calme1709:
Relative lengths should be resolved against the initial font rather than the element's font so all the work around updating layout is unnecessary


AtkinsSJ (approved):
Always nice when the fix makes things simpler!

PR #6068, opened 2025-09-04
LibWeb: Throw parsing error if `::slotted()` argument is not simple

kalenikaliaksandr:
...selector. Grammar per spec: `::slotted( <compound-selector> )`, so we should reject selector as invalid if first compound selector is followed by something else.

This change makes layout more correct on https://www.rottentomatoes.com/

Before:
CleanShot 2025-09-04 at 13 21 45@2x

After:
CleanShot 2025-09-04 at 13 22 16@2x

### DIFF
--- a/Libraries/LibWeb/CSS/Parser/SelectorParsing.cpp
+++ b/Libraries/LibWeb/CSS/Parser/SelectorParsing.cpp
@@ -517,6 +517,15 @@ Parser::ParseErrorOr<Selector::SimpleSelector> Parser::parse_pseudo_simple_selec
                         });
                         return ParseError::SyntaxError;
                     }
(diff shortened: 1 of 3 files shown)


AtkinsSJ (commented):
Code is good, terminology is slightly wrong. A "simple selector" is a single component of a "compound selector". For example, `a.cool:visited` is a compound selector, made of 3 simple selectors: `a`, `.cool`, and `:visited`. (https://drafts.csswg.org/selectors-4/#structure)

So the commit message and the comment in the test should talk about compound selectors, not simple ones.

kalenikaliaksandr:
thanks, for clarifying this! I was feeling uncertain about correctness of terminology I used. commit description is updated.

PR #6071, opened 2025-09-04
LibWeb/CSS: Generate code for CSS dimension units

AtkinsSJ:
Generate ALL the code!

Generate the dimension-unit enums, along with code to parse/serialize them, and convert between units. This also includes a couple of small code tweaks and (of course) a bindings generator fix.

A few things motivated this:
- The `CSS.unit()` OM methods were very repetitive, and could easily go out of sync with what we support (and in fact already had units we didn't support, oops)
- Typed OM requires a way to check if two units are compatible, and directly convert between compatible units
- It's always nice to not have manual to/from string functions

The one thing here that I'm not 100% sold on is putting the unit ratios in the JSON. These tend to be quite nasty numbers with recurring digits, when written out literally. Alternatives to that are to make the JSON field a string and directly dump that in the code; or to remove the JSON field and adapt the code in `to_degrees()` etc to return the ratio instead. The numbers I've included come from WolframAlpha, so they have a ridiculous, guaranteed-to-be-more-than-enough number of digits, just to make sure I'm not introducing precision issues. (Before that I used a calculator app which gave me something like 12 decimal places, which *wasn't* enough to prevent regressions on some resolution media-query tests.)

### DIFF
--- a/Libraries/LibWeb/CSS/Angle.cpp
+++ b/Libraries/LibWeb/CSS/Angle.cpp
@@ -48,22 +48,12 @@ String Angle::to_string(SerializationMode serialization_mode) const
 
 double Angle::to_degrees() const
 {
-    switch (m_unit) {
-    case AngleUnit::Deg:
-        return m_value;
-    case AngleUnit::Grad:
-        return m_value * (360.0 / 400.0);
-    case AngleUnit::Rad:
-        return AK::to_degrees(m_value);
-    case AngleUnit::Turn:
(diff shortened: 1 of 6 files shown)


AtkinsSJ:
Rebased to fix conflicts.

AtkinsSJ:
Same. :sweat_smile:

AtkinsSJ:
Another rebase to fix conflicts.

PR #6072, opened 2025-09-04
LibWeb: Store some font-* properties in computed form

Calme1709:
This PR updates font-size, font-weight, font-width, and font-style to be stored in their computed form within ComputedProperties.

Also includes some other smaller fixes/improvements to accommodate this as well as improved parsing for the associated `@font-face` descriptors.


Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1950,10 +1961,14 @@ CSSPixels StyleComputer::relative_size_mapping(RelativeSize relative_size, CSSPi
...
 
-RefPtr<Gfx::FontCascadeList const> StyleComputer::compute_font_for_style_values(StyleValue const& font_family, CSSPixels const& font_size, StyleValue const& font_style, StyleValue const& font_weight, StyleValue const& font_stretch) const
+RefPtr<Gfx::FontCascadeList const> StyleComputer::compute_font_for_style_values(StyleValue const& font_family, CSSPixels const& font_size, StyleValue const& font_style, double font_weight, StyleValue const& font_stretch) const
 {
     auto width = font_stretch.to_font_width();
-    auto weight = font_weight.to_font_weight();
+
+    // FIXME: We round to int here as that is what is expected by our font infrastructure below expects
+    // FIXME: This can end up negative due to interpolation - we can remove the max() once we correctly handle
+    //        interpolation clamping
```
AtkinsSJ: #6113 is merged which should solve this.

Libraries/LibWeb/DOM/Element.cpp
```diff
@@ -813,6 +813,9 @@ CSS::RequiredInvalidationAfterStyleChange Element::recompute_inherited_style()
             computed_properties->set_property(property_id, *preabsolutized_value, is_inherited ? CSS::ComputedProperties::Inherited::Yes : CSS::ComputedProperties::Inherited::No);
             old_values_with_relative_units.set(i, old_value);
         }
+
+        // FIXME: We should also consider properties which depend on their inherited for their computed forms (e.g.
```
AtkinsSJ: Nit: Inherited what?

Libraries/LibWeb/CSS/StyleValues/ShorthandStyleValue.cpp
```diff
@@ -348,23 +349,68 @@ String ShorthandStyleValue::to_string(SerializationMode mode) const
...
+                //  - We are working with computed values in which case we would have already converted any
+                //    CalculatedStyleValues values to normal PercentageStyleValues
+                font_width_as_percentage = font_width->as_calculated().resolve_percentage({}).map([](auto percentage) { return percentage.value(); });
+
+            if (!font_width_as_percentage.has_value())
                 return {};
-        }
+
+            // ultra-condensed 50%
+            if (*font_width_as_percentage == 50)
```
AtkinsSJ: Nit: You don't need the `*` here and below: `Optional<T>` is directly comparable with `T`.

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1961,9 +1964,10 @@ CSSPixels StyleComputer::relative_size_mapping(RelativeSize relative_size, CSSPi
     VERIFY_NOT_REACHED();
 }
 
-RefPtr<Gfx::FontCascadeList const> StyleComputer::compute_font_for_style_values(StyleValue const& font_family, CSSPixels const& font_size, StyleValue const& font_style, double font_weight, StyleValue const& font_stretch) const
+RefPtr<Gfx::FontCascadeList const> StyleComputer::compute_font_for_style_values(StyleValue const& font_family, CSSPixels const& font_size, StyleValue const& font_style, double font_weight, Percentage const& font_width) const
 {
-    auto width = font_stretch.to_font_width();
+    // FIXME: We round to int here as that is what is expected by our font infrastructure below expects
```
AtkinsSJ: Nit: "expected" and "expects" in one sentence. :^)

Looks like that mistake crept in below too.

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -3429,6 +3439,58 @@ NonnullRefPtr<StyleValue const> StyleComputer::compute_font_weight(NonnullRefPtr
...
+    // <percentage [0,∞]>
+    if (specified_value->is_percentage())
+        return specified_value;
+
+    // AD-HOC: We support calculated percentages as well
+    if (specified_value->is_calculated())
+        return PercentageStyleValue::create(specified_value->as_calculated().resolve_percentage({ .length_resolution_context = parent_length_resolution_context }).value());
+
+    // ultra-condensed 50%
+    if (specified_value->to_keyword() == Keyword::UltraCondensed)
```
AtkinsSJ: Nit: A switch on `specified_value->to_keyword()` seems less verbose to me.

AtkinsSJ (changes requested):
I'm very happy to see those `StyleValue::to_font_foo()` methods gone!

Libraries/LibWeb/CSS/StyleComputer.cpp
Calme1709: Correct - removed both here and for font-size below.

tcl3:
FWIW, I had a quick look into the cause of the new failures in `font-weight-computed.html`. The problem seems to be an under-invalidation when the element's style is updated.

I was able to "fix" the issue by doing this:
```diff
diff --git a/Libraries/LibWeb/DOM/Element.cpp b/Libraries/LibWeb/DOM/Element.cpp
index 8a97662872b..c10b6ffca85 100644
--- a/Libraries/LibWeb/DOM/Element.cpp
+++ b/Libraries/LibWeb/DOM/Element.cpp
@@ -2523,7 +2523,7 @@ void Element::invalidate_style_after_attribute_change(FlyString const& attribute
     }

     changed_properties.append({ .type = CSS::InvalidationSet::Property::Type::Attribute, .value = attribute_name });
-    invalidate_style(StyleInvalidationReason::ElementAttributeChange, changed_properties, style_invalidation_options);
+    invalidate_style(StyleInvalidationReason::ElementAttributeChange);
 }

 bool Element::is_hidden() const
```

I think this issue is probably outside the scope of this PR, but I thought it was worth mentioning. I'm happy to create a minimal reduction and dig a bit deeper into this once this PR is merged.

Calme1709:
Agreed - I did mention this in the associated commit message but probably should have included in PR description as well:
> This does expose a few false positives in the font-weight-computed.html WPT test. This is because we don't recompute non-inherited font-weight within `recompute_inherited_style` which means that relative keyword values can fall out of sync with their parent's value. These previously passed as we treated `bolder` and `lighter` as aliases for `bold` and `normal` respectively.

FIXME [here](https://github.com/LadybirdBrowser/ladybird/blob/c4739020c0891ac0993c0eeef9d9c13dc59100a3/Libraries/LibWeb/DOM/Element.cpp#L818)

tcl3:
> Agreed - I did mention this in the associated commit message but probably should have included in PR description as well:

Nice! It probably would have been helpful if I'd taken the time read the relevant commit messages in full :sweat_smile:

Calme1709:
Rebased to resolve conflicts

Calme1709:
Rebased to resolve conflicts

Calme1709:
Rebased to resolve conflicts - are you waiting on anything else @AtkinsSJ?

PR #6073, opened 2025-09-04
LibWeb: Rewrite "destroy a document and descendants" without spin until

kalenikaliaksandr:
Replaces spin until with GC-allocated counting object that invokes destruction callback once all child navigable documents are destroyed.

The change doesn't have a test but not using spin until is strictly better than using it. Also improves https://www.rottentomatoes.com/ where previously we would hang or crash after loading.


Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -4168,6 +4168,40 @@ void Document::make_unsalvageable([[maybe_unused]] String reason)
...
+        Base::visit_edges(visitor);
+        visitor.visit(document);
+        visitor.visit(after_all);
+    }
+
+    void increment_destroyed()
+    {
+        --remaining_children;
+        if (remaining_children > 0)
+            return;
```
gmta, line 4200: In this new case, instead of hanging we never destroy the document, right? Maybe we could add a timeout as well to detect this happening?

kalenikaliaksandr: yeah, having timeout is a good idea. I added one that logs FIXME in case 15s was not enough to destroy all documents. but just to clarify: `increment_destroyed()` is supposed to be invoked multiply times until `remaining_children` is 0 and that's what happens now on https://www.rottentomatoes.com/ that was previously stuck in spin until.

PR #6075, opened 2025-09-04
LibWeb: Implement CSS-Multicol's 'Pseudo-Algorithm'

Psychpsyo:
This implements [the pseudo-algorithm](https://drafts.csswg.org/css-multicol-2/#pseudo-algorithm) from the CSS Multicol spec.
I know, `determine_used_value_for_column_width` gets added but currently goes unused, it'd get used right where the FIXME got added. I hope that's fine. It's kinda part of the same algorithm.

I'm not sure if any of this is properly testable, since it only causes a BFC to be established on multicol elements for now. Any tests for that would likely have to be rewritten once actual column behavior gets implemented.

### DIFF
--- a/Libraries/LibWeb/Layout/BlockFormattingContext.cpp
+++ b/Libraries/LibWeb/Layout/BlockFormattingContext.cpp
@@ -71,6 +71,13 @@ static bool margins_collapse_through(Box const& box, LayoutState& state)
 
 void BlockFormattingContext::run(AvailableSpace const& available_space)
 {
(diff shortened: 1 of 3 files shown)


AtkinsSJ:
Looks reasonable. But please rebase as it's been a while.

Psychpsyo:
> Looks reasonable. But please rebase as it's been a while.

Done. CI issue looks like a flake.

AtkinsSJ (approved):

PR #6080, opened 2025-09-04
Update BuildInstructionsLadybird.md

morhook:
I had a problem building ladybird on FFMPEG building. Documenting what I did to fix it on my XUbuntu 25.04 on i3wm.


morhook:
Probably there’s a better way to fix this. On Ubuntu 24.04 this problem didn’t happen. This vcpkg is a little bit unbeknownst to me.

Issue #6063, opened 2025-09-03
WebContent process crash on visiting developer.mozilla.org


pranair:
### Summary

When visiting [developer.mozilla.org](developer.mozilla.org), WebContent process crashes with "Ladybird  flew off-course" error. Issue seems to be related to `@property`+`replaceSync` scenario.

### Operating system

Linux

### Steps to reproduce

1. Go to [developer.mozilla.org](developer.mozilla.org) or minimal html file
2. Crash

### Expected behavior

No crash

### Actual behavior

Crash

### URL for a reduced test case

N/A

### HTML/SVG/etc. source for a reduced test case

```html
<!DOCTYPE html>
<html>
<head>
  <title>@property + replaceSync Crash Test</title>
</head>
<body>
  <h1>Test for @property crash with replaceSync</h1>
  <p>This test should crash if the bug is present.</p>
  <script>
    if (window.CSSStyleSheet && CSSStyleSheet.prototype.replaceSync) {
      const sheet = new CSSStyleSheet();
      const css = `
        @property --test {
          syntax: '<color>';
          inherits: false;
          initial-value: black;
        }
      `;
      sheet.replaceSync(css);
      document.adoptedStyleSheets = [sheet];
      console.log("Test completed without crash.");
    } else {
      console.log("Constructible stylesheets or replaceSync not supported.");
    }
  </script>
</body>
</html>
```

### Log output and (if possible) backtrace

```shell
➜  bin git:(master) ✗ ./Ladybird ../../../minimal_crash.html 
17088.113 Ladybird(144978): WebContent process crashed! Last page loaded: file:///home/user/code/ladybird/minimal_crash.html
17088.113 Ladybird(144978): Consider raising an issue at https://github.com/LadybirdBrowser/ladybird/issues/new/choose
17088.120 WebContent(145013): FIXME: Handle null navigation_request in navigation response Content Security Policy check.
```

### Contribute a patch?

- [x] I’ll contribute a patch for this myself.

vpzomtrrfrt:
Backtrace:

```
#0  Web::Bindings::PlatformObject::realm () at /home/colin/ladybird/Libraries/LibWeb/Bindings/PlatformObject.cpp:30
#1  0x00007fc77b0962b9 in Web::CSS::Parser::ParsingParams::ParsingParams ()
    at /home/colin/ladybird/Libraries/LibWeb/CSS/Parser/Parser.cpp:51
[...]

PR #6081, opened 2025-09-05
LibWeb: Avoid dereferencing a null pointer to document

pranair:
This is the fix for [https://github.com/LadybirdBrowser/ladybird/issues/6063](https://github.com/LadybirdBrowser/ladybird/issues/6063)

This commit fixes a crash in the LibWeb when converting an at-rule to a property rule. It handles a case where a document might not be available by falling back to using the realm to get the ParsingParams.

### DIFF
--- a/Libraries/LibWeb/CSS/Parser/RuleParsing.cpp
+++ b/Libraries/LibWeb/CSS/Parser/RuleParsing.cpp
@@ -754,7 +754,12 @@ GC::Ptr<CSSPropertyRule> Parser::convert_to_property_rule(AtRule const& rule)
         return {};
     }
 
-    auto parsing_params = CSS::Parser::ParsingParams { *document() };
+    CSS::Parser::ParsingParams parsing_params;
+    if (document())
+        parsing_params = CSS::Parser::ParsingParams { *document() };
+    else
+        parsing_params = CSS::Parser::ParsingParams { realm() };
+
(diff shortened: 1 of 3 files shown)


Libraries/LibWeb/CSS/Parser/RuleParsing.cpp
```diff
@@ -754,7 +754,12 @@ GC::Ptr<CSSPropertyRule> Parser::convert_to_property_rule(AtRule const& rule)
         return {};
     }
 
-    auto parsing_params = CSS::Parser::ParsingParams { *document() };
+    CSS::Parser::ParsingParams parsing_params;
+    if (auto const* document_ptr = document())
```
AtkinsSJ: `document_ptr` isn't necessary, you can just do `if (document())`

AtkinsSJ (changes requested):
Welcome! This looks like a nice fix.

Please add a test for it. Remove some of the unnecessary parts from your test in #6063 and make it use our test functions. See https://github.com/LadybirdBrowser/ladybird/blob/master/Tests/LibWeb/Text/input/DOM/createElement-with-emoji-in-tag-name.html for a simple example.

Also, the commit message should have a capital letter after the `:`, and we like to avoid words like `fix`. Instead say what you did, which is avoid dereferencing a null pointer.

pranair:
Hello!

Thank you for the suggestions, I've implemented them.

PR #6083, opened 2025-09-05
LibWeb/CSS: Ensure properties are clamped to valid ranges

InvalidUsernameException:
This ensures that we clamp values for properties like `padding-*` to valid ranges (non-negative in this case) if they are specified with `calc()`.

This PR fixes the primary layout issue on https://lwn.net/ (yes, not the first place I would have expected problems either).

| Before | After |
|--------|--------|
| before | after |

### DIFF
--- a/Libraries/LibWeb/CSS/Angle.cpp
+++ b/Libraries/LibWeb/CSS/Angle.cpp
@@ -100,12 +100,11 @@ Optional<Angle::Type> Angle::unit_from_name(StringView name)
 
 Angle Angle::resolve_calculated(NonnullRefPtr<CalculatedStyleValue const> const& calculated, Layout::Node const& layout_node, Angle const& reference_value)
 {
-    return calculated->resolve_angle_deprecated(
-                         {
-                             .percentage_basis = reference_value,
-                             .length_resolution_context = Length::ResolutionContext::for_layout_node(layout_node),
-                         })
-        .value();
+    CalculationResolutionContext context {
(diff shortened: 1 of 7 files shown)


Libraries/LibWeb/CSS/StyleValues/CalculatedStyleValue.cpp
```diff
@@ -3217,19 +3217,19 @@ NonnullRefPtr<CalculationNode const> simplify_a_calculation_tree(CalculationNode
                 [&](Angle const& angle) -> RefPtr<NumericCalculationNode const> {
                     VERIFY(context.percentages_resolve_as == ValueType::Angle);
                     if (angle.type() == Angle::Type::Deg)
-                        return nullptr;
+                        return NumericCalculationNode::create(angle.percentage_of(*percentage), context);
```
AtkinsSJ, line 3220: I have no idea why I had `nullptr` here or why nobody caught that for so long. :laughing:

AtkinsSJ (approved):
Thanks!

Libraries/LibWeb/CSS/StyleValues/CalculatedStyleValue.cpp
InvalidUsernameException: Yeah, I was a bit surprised when I found this. Maybe it is because of the else branch (step 1.2), which is a similar-looking `visit()` where `nullptr` is the correct thing for canonical units.

PR #6088, opened 2025-09-05
LibWeb: Add missing `DataTransfer` and `DataTransferIttemList` methods

### LINKED ISSUE #5848: Calling  a `FrozenArray` getter multiple times doesn't return the same object
Multiple calls to a getter that returns a WebIDL `FrozenArray<T>` type should return the same object. Our implementation of `FrozenArray` doesn't currently do this.

Found while reviewing #5846.

Minimal reproduction 
```html
<!DOCTYPE html>
<script>
    let dt = new DataTransfer();
    // This should return true, but we currently return false
    console.log(dt.types === dt.types);
</script>
```

tcl3:
This PR adds `DataTransferItemList.remove()` , `DataTransferItemList.clear()` and `DataTransfer.clearData()`. 

I've also added `[CachedAttribute]` to `DataTransfer.types` - clearing the cache when `m_types` is modified ensures that the same object is returned if the array is unmodifed. There was discussion on #5848 that using the bindings generator may not be the best way to implement this (https://github.com/LadybirdBrowser/ladybird/issues/5848#issuecomment-3213693538), but it seemed like the easiest way to implement the required behavior to me.

The other relevant WPT test for these changes is: http://wpt.live/html/editing/dnd/datastore/datatransfer-types.html, this has been added in #5846 and all tests should pass if that PR is rebased on top of this. 

Fixes #5848

### DIFF
--- a/Libraries/LibWeb/HTML/DataTransfer.cpp
+++ b/Libraries/LibWeb/HTML/DataTransfer.cpp
@@ -74,6 +74,8 @@ void DataTransfer::visit_edges(JS::Cell::Visitor& visitor)
     Base::visit_edges(visitor);
     visitor.visit(m_items);
     visitor.visit(m_item_list);
+
+    VISIT_CACHED_ATTRIBUTE(types);
 }
 
 void DataTransfer::set_drop_effect(String const& drop_effect)
@@ -374,6 +376,7 @@ void DataTransfer::update_data_transfer_types_list()
     }
 
     // 3. Set the DataTransfer object's types array to the result of creating a frozen array from L.
+    set_cached_types(nullptr);
     m_types = move(types);
 }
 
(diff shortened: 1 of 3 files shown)


Libraries/LibWeb/HTML/DataTransfer.cpp
```diff
@@ -248,6 +248,22 @@ GC::Ref<DataTransferItem> DataTransfer::add_item(DragDataStoreItem item)
...
+void DataTransfer::remove_item(size_t index)
+{
+    VERIFY(m_associated_drag_data_store);
+    VERIFY(index < m_item_list.size());
+
+    m_associated_drag_data_store->remove_item_at(index);
+    auto& item = m_item_list.at(index);
+    item->set_item_index({}, OptionalNone {});
+    m_item_list.remove(index);
+    for (size_t i = 0; i < m_item_list.size(); ++i) {
```
gmta: Couldn't `size_t i = 0` be `size_t i = index` here? 🤔

tcl3: Yes, there's absolutely no need to update every index. Updated as suggested.

PR #6089, opened 2025-09-05
LibWasm: Fix memory.fill ignoring memory index and unsafe bounds check

shlyakpavel:
Previously, the memory.fill instruction always wrote to memory 0, ignoring the selected memory index. This caused incorrect behavior in multi-memory modules (e.g. filling mem0 instead of mem1). Additionally, the bounds check used `destination_offset + count` without overflow checking, which could wrap and bypass validation.

This patch:
- Passes `args.memory_index` into store_to_memory, so the correct memory is filled.
- Uses Checked<u32> for destination_offset + count, consistent with memory.copy and memory.init, to prevent overflow.

Minimal repro:

    (module
      (memory $m0 1)
      (memory $m1 1)

      (func (export "go") (result i32)
        ;; Fill mem1[0] with 0xAA
        i32.const 0
        i32.const 170
        i32.const 1
        memory.fill (memory 1)

        ;; Return (mem1[0] << 8) | mem0[0]
        i32.const 0
        i32.load8_u (memory 1)
        i32.const 8
        i32.shl
        i32.const 0
        i32.load8_u (memory 0)
        i32.or
      )
    )

Before fix: returns 170 (0x00AA).
After fix:  returns 43520 (0xAA00).


konradekk:
consider adding the repro as a test (or at least the relevant part of it).

shlyakpavel:
@konradekk I don't quite understand how to import a WebAssembly app as test

alimpfard:
You can put a wasm file [here](https://github.com/LadybirdBrowser/ladybird/tree/master/Libraries/LibWasm/Tests/Fixtures/Modules), make a new subdirectory [here](https://github.com/LadybirdBrowser/ladybird/tree/master/Libraries/LibWasm/Tests) with some .js in it that describes the test.

Check [these tests](https://github.com/LadybirdBrowser/ladybird/blob/master/Libraries/LibWasm/Tests/Parser/test-basic-load.js) and test-wasm.cpp for a reference.

shlyakpavel:
Well, I have added the test

PR #6093, opened 2025-09-05
LibWeb: Ensure parser cannot change the mode is handled

Euro20179:
This fixes at least 1 wpt bug where text/plain documents are rendered in quirks mode. The test in question: https://wpt.live/html/browsers/browsing-the-web/read-text/load-text-plain.html

### DIFF
--- a/Libraries/LibWeb/DOM/Document.h
+++ b/Libraries/LibWeb/DOM/Document.h
@@ -414,6 +414,9 @@ public:
     bool in_limited_quirks_mode() const { return m_quirks_mode == QuirksMode::Limited; }
     void set_quirks_mode(QuirksMode mode) { m_quirks_mode = mode; }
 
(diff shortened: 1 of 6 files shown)


tcl3:
> This fixes at least 1 wpt bug where text/plain documents are rendered in quirks mode. The test in question: https://wpt.live/html/browsers/browsing-the-web/read-text/load-text-plain.html

It would be good if you could import this test into our test suite. Our importer isn't currently smart enough to download the `text-plain.txt` file used by this test, so you'll need to download it manually.

Running these commands should work:
```sh
curl http://wpt.live/common/text-plain.txt > Tests/LibWeb/Text/input/wpt-import/common/text-plain.txt  
./Meta/WPT.sh update
./Meta/WPT.sh import html/browsers/browsing-the-web/read-text/load-text-plain.html 
```

Euro20179:
Thanks for the help!
I have added the tests.

PR #6095, opened 2025-09-06
LibWasm: Use 0x40 flag for SIMD memory memidx like scalar ops

shlyakpavel:
SIMD loads/stores checked bit 0x20 of the align immediate to detect a following memory index, unlike scalar mem ops which use 0x40 per the multi-memory encoding. This caused the memidx byte to be misparsed as the next immediate (e.g. offset).

Update both SIMD sites (v128 load/store and lane variants) to check and clear 0x40, then read LEB128<u32> memidx.

Repro:
```
  (module (memory $m0 1) (memory $m1 1)
    (func (export "go")
      i32.const 0
      v128.load (memory 1)
      drop))
```
Before: printed memidx 0 with offset 1.
After:  prints memidx 1 with offset 0.


alimpfard:
0x20 is the new 1<<6.
~~(Why is it 0x40 again? Isn't that checking bit 7?)~~ I bet I wrote that at 6am too

PR #6096, opened 2025-09-06
LibWasm: Fix pushes for i16x8.replace_lane in Opcode table

shlyakpavel:
The opcode entry declared i16x8_replace_lane with pushes = -1, but replace_lane pops 2 (vector, lane value) and pushes 1 result vector. Set pushes to 1 to match the other replace_lane opcodes.


alimpfard:
nice catch!

PR #6097, opened 2025-09-06
LibGfx+LibWeb: Support DirectX-based WebGL on Windows

ayeteadoe:
I initially tried to build off of the recent Vulkan work for Linux WebG by using the `VK_KHR_EXTERNAL_MEMORY_WIN32_EXTENSION_NAME` device extension and `VK_EXTERNAL_MEMORY_HANDLE_TYPE_D3D11_TEXTURE_BIT_KHR` handle type when creating the `VkImage`.  That was all working; however, I was unable to create the acutal EGL surface (or image) with `EGL_D3D_TEXTURE_2D_SHARE_HANDLE_ANGLE` from the vulkan-created shared handle.

It seems to only work with getting the shared handle from a D3D11 texture (https://github.com/Microsoft/angle/wiki/Interop-with-other-DirectX-code). So I switched to Direct 3D on Windows for skia to get a painting surface compatible with the texture, and it seems to just work™

### Demo

https://github.com/user-attachments/assets/ef8d101a-5f7d-4efe-b1b3-662781f48d13


Libraries/LibWeb/WebGL/OpenGLContext.cpp
```diff
@@ -363,6 +372,55 @@ void OpenGLContext::allocate_vkimage_painting_surface()
 }
 #endif
 
+#if defined(AK_OS_WINDOWS)
+void OpenGLContext::allocate_d3d11texture_painting_surface()
+{
+    char const* extensions = eglQueryString(m_impl->display, EGL_EXTENSIONS);
+    if (!strstr(extensions, "EGL_ANGLE_d3d_share_handle_client_buffer")) {
```
ADKaster: strstr? I mean it works but surely we're parsing these egl strings using AK methods elsewhere.

Libraries/LibGfx/Direct3DContext.cpp
```diff
@@ -124,4 +124,54 @@ ErrorOr<NonnullOwnPtr<Direct3DContext>> Direct3DContext::try_create()
...
+    if (m_impl->shared_handle != INVALID_HANDLE_VALUE)
+        return m_impl->shared_handle;
+
+    if (HRESULT const hr = m_impl->dxgi_resource->GetSharedHandle(&m_impl->shared_handle); FAILED(hr)) {
+        return Error::from_windows_error(hr);
+    }
+    return m_impl->shared_handle;
+}
+
+ErrorOr<NonnullRefPtr<Direct3D11Texture>> Direct3D11Texture::try_create_shared(Direct3DContext const& context, u32 const width, u32 const height, DXGI_FORMAT const format)
```
ADKaster: generally we don't use const on params.

Libraries/LibGfx/Direct3DContext.cpp
```diff
@@ -124,4 +124,54 @@ ErrorOr<NonnullOwnPtr<Direct3DContext>> Direct3DContext::try_create()
...
+{
+    D3D11_TEXTURE2D_DESC tex_desc = {};
```
ADKaster, line 153: any difference b/w the D3D11 struct and a DirectX 12+ struct? i.e. was this taken from an example, or is the v11 struct the latest?

Given that one of either angle or skia seems to require v12, I don't see a reason to try to support earlier.

Libraries/LibGfx/Direct3DContext.h
```diff
@@ -9,11 +9,16 @@
 #include <AK/Error.h>
 #include <AK/NonnullOwnPtr.h>
 #include <AK/Platform.h>
+#include <AK/RefCounted.h>
+
+#include <dxgiformat.h>
```
ADKaster, line 14: this header is `windows.h`-safe?

Libraries/LibGfx/Direct3DContext.cpp
```diff
@@ -0,0 +1,127 @@
...
+ */
+
+#include <AK/Format.h>
+#include <AK/NonnullRefPtr.h>
+#include <LibGfx/Direct3DContext.h>
+
+#include <d3d11.h>
+#include <d3d12.h>
+#include <dxgi.h>
+#include <wrl.h>
```
ADKaster: here and in a later commit. should we be pulling in Windows.h explicitly? either from the windows sdk or from AK.

vcpkg.json
```diff
@@ -178,7 +178,19 @@
...
       "name": "skia",
-      "platform": "linux | freebsd | openbsd | windows",
+      "platform": "windows",
+      "features": [
+        "direct3d",
+        "freetype",
+        "fontconfig",
+        "harfbuzz",
+        "icu",
+        "vulkan"
```
ADKaster: if the way to make skia, angle, and WebGL/PaintingSurface play nice with textures on Windows requires the DirectX backend, do we still need vulkan?

Libraries/LibWeb/WebGL/OpenGLContext.cpp
```diff
@@ -23,11 +23,18 @@ extern "C" {
...
-// Enable WebGL if we're on MacOS and can use Metal or if we can use shareable Vulkan images
-#if defined(AK_OS_MACOS) || defined(USE_VULKAN_IMAGES)
+// Enable WebGL if we're on MacOS and can use Metal, if we're on Windows and can use DirectX, or if we can use shareable Vulkan images
+#if defined(AK_OS_MACOS) || defined(AK_OS_WINDOWS) || defined(USE_VULKAN_IMAGES)
 #    define ENABLE_WEBGL 1
 #endif
 
+#if defined(AK_OS_WINDOWS)
+#    include <LibGfx/Direct3DContext.h>
+#    include <d3d11.h>
```
ADKaster: same question. to Windows.h or not to windows.h

Libraries/LibGfx/Direct3DContext.cpp
ayeteadoe: Yeah they are different: [D3D11 struct](https://learn.microsoft.com/en-us/windows/win32/api/d3d11/ns-d3d11-d3d11_texture2d_desc) vs [D3D12 struct](https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_resource_desc) 

Skia is the one that requires D3D12, Angle does not support D3D12, only [D3D9 and D3D11](https://github.com/google/angle?tab=readme-ov-file#platform-support-via-backing-renderers)

So given Angle requires D3D11 and OpenGLContext is the place that needs to explicitly create a texture, we create a shareable D3D11Texture that D3D12 Skia can also use.

vcpkg.json
ayeteadoe: Yeah we shouldn't need Vulkan anymore for skia on Windows. For future contexts that benefit from GPU accelerated rendering (e.g. Video) we still have the potential to interop with Vulkan and a D3D12-based skia painting surface. @Zaggy1024 and I discussed that a bit in `#ui-windows` [here](https://discord.com/channels/1247070541085671459/1306918361732616212/1413783166891589754).

I'll remove the vulkan feature for windows skia vcpkg port

Libraries/LibGfx/Direct3DContext.h
ayeteadoe: Yeah, this header is simply the [DXGI_FORMAT enum](https://learn.microsoft.com/en-us/windows/win32/api/dxgiformat/ne-dxgiformat-dxgi_format#syntax) with no transitive includes

Libraries/LibGfx/Direct3DContext.cpp
ayeteadoe: d3d11.h/d3d12.h/dxgi.h transitively include `windows.h`. Given we don't need any WinSock2 stuff here we could just explicitly do `#include <windows.h>`

Libraries/LibGfx/CMakeLists.txt
```diff
@@ -149,6 +151,7 @@ if (NOT ANDROID)
             #  alone causes transitive issues in other TUs that depend on it
             target_compile_options(LibGfx PRIVATE -fno-sanitize=address)
         endif()
+        target_link_libraries(LibGfx PUBLIC d3d11 d3d12 dxgi)
```
ADKaster, line 156: what does this require? I feel like at the very least, something like `$WINSDK_ROOT` or some such is being implicitly searched by the linker here? A FIXME or issue to create a proper find module for DirectX libs/headers is probably in order. Just assuming the user's env isn't messed up is fine for now though.

Libraries/LibGfx/Direct3DContext.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+using namespace Microsoft::WRL;
+
+namespace Gfx {
+
+struct Direct3DContext::Impl {
+    ComPtr<IDXGIFactory1> factory;
```
ADKaster: The docs for the  Windows Runtime C++ Template Library say it's deprecated, and that new apps should use C++/WinRT projections instead.

https://learn.microsoft.com/en-us/cpp/cppcx/wrl/windows-runtime-cpp-template-library-wrl?view=msvc-180

Another issue for the issue list I s'pose, this code doesn't look 'old'.

Libraries/LibGfx/Direct3DContext.cpp
```diff
@@ -0,0 +1,129 @@
...
+        adapter = fallback_adapter;
+    }
+
+    if (adapter == nullptr) {
+        return Error::from_string_literal("Unable to retrieve adapter");
+    }
+
+    ComPtr<ID3D12Device> d12_device;
+    if (HRESULT const hr = D3D12CreateDevice(adapter.Get(), D3D_FEATURE_LEVEL_12_0, IID_PPV_ARGS(&d12_device));
+        FAILED(hr)) {
```
ADKaster: formatting put this FAILED on a newline whereas the ones below are in-line.

Libraries/LibGfx/CMakeLists.txt
ayeteadoe: https://learn.microsoft.com/en-us/cpp/build/reference/libpath-additional-libpath?view=msvc-170 states:

> Use the /LIBPATH option to override the environment library path. The linker first searches in the path specified by this option, and then searches in the path specified in the LIB environment variable.

According to https://lld.llvm.org/windows_support.html, the `lld-link.exe` driver is a pretty faithful `link.exe` implementation, it also supports the `/libpath` flag, so it should have the same behaviour described in the docs above.

This is the content of `LIB` on my system:
```
C:\Users\ayeteadoe\Development\ladybird>echo %LIB%
C:\Program Files\Microsoft Visual Studio\18\Insiders\VC\Tools\MSVC\14.50.35717\ATLMFC\lib\x64;C:\Program Files\Microsoft Visual Studio\18\Insiders\VC\Tools\MSVC\14.50.35717\lib\x64;C:\Program Files (x86)\Windows Kits\NETFXSDK\4.8\lib\um\x64;C:\Program Files (x86)\Windows Kits\10\lib\10.0.26100.0\ucrt\x64;C:\Program Files (x86)\Windows Kits\10\\lib\10.0.26100.0\\um\x64
```

And this is the Windows SDK directory path on my system:
```
C:\Users\ayeteadoe\Development\ladybird>echo %WindowsSdkDir%
C:\Program Files (x86)\Windows Kits\10\
```

The `C:\Program Files (x86)\Windows Kits\10\Lib\10.0.26100.0\um\x64` path contains these DirectX import libs, and given these are all system DLLs (i.e. they reside in `C:\Windows\System32`). Because the [DLL search order does fallback to the system directory](https://learn.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order#standard-search-order-for-unpackaged-apps), we can load the DLLs successfully at run time.

---

So to summarize, yes, we are depending on implicit environment (Windows SDK+System) behaviour here for this to all work. Given the Windows SDK environment setup is handled implicitly by Visual Studio, and the DLLs are in the system dir and will always be found at runtime, I'm not sure we'd need to do anything more here, given we don't have similar FIXME's for our usage of `Bcrypt` which is in an identical scenario. I can still leave a NOTE comment giving some details so others don't need to re-confirm the information I gave above.

ADKaster: Sure, that all makes sense, for a normal install and normal developer workstation.

I don't doubt that some people will have multiple versions of the Windows SDK installed for multiple different projects, or multiple VS installs, or other shenanigans in the future though. Explicit better than implicit, etc. I know that the gn builds for skia and angle require things like winsdkdir passed and such in order to properly build (as I patched the skia vcpkg port to do just that last year).

In the future :tm: it would be good to have Find modules for these windows components

ayeteadoe: Yeah good points, I'll add a FIXME comment about setting up find cmake modules

ayeteadoe:
Oops, I just realized when I switched to using the D3D backend context for navigables, I missed updating `BackingStoreManager` to handle Windows now, as `USE_VULKAN` in [reallocate_backing_stores](https://github.com/LadybirdBrowser/ladybird/blob/master/Libraries/LibWeb/Painting/BackingStoreManager.cpp#L124) is no longer used. This means we were falling back to bitmap wrapped backing stores.

So one more rebase incoming that will change that check to `#if defined(USE_VULKAN) || defined(AK_OS_WINDOWS)` in the `LibWeb: Use D3D12 SkiaBackendContext for Navigables on Windows` commit.

I initially missed this because I was only testing https://webglsamples.org on this branch. But the performance slowdown with the bitmap wrapped backing stores was very noticable when I finally tested https://ladybird.org compared to current master.

Libraries/LibGfx/Direct3DContext.cpp
```diff
@@ -0,0 +1,180 @@
...
+#include <d3d11.h>
+#include <d3d12.h>
+#include <dxgi.h>
+#include <wrl.h>
+
+using namespace Microsoft::WRL;
+
+namespace Gfx {
+
+// FIXME: Use the modern ComPtr equivalent from the Windows Runtime C++ Template Library (WRL)
```
R-Goc: For ComPtr you only need Wrl/client.h, see: https://github.com/LadybirdBrowser/ladybird/pull/6246/files.
I don't think this is a good fixme as the newer lib (WRT) uses exceptions for everything. So WRL is the newest library that doesn't rely on exceptions.

ayeteadoe: Ah yeah my comment is incorrect, as the review request was to eventually use `winrt::com_ptr`, I am just going to switch to that now anyways given it is pretty low effort.

Libraries/LibGfx/Direct3DContext.cpp
ayeteadoe: It was pretty low effort to move to `winrt::com_ptr` from `Microsoft::WRL::ComPtr` so I've done that instead of an issue/FIXME.

R-Goc: The thing is wrt uses exceptions. Com_ptr throws in many of its functions.

R-Goc: The thing is wrt uses exceptions. Com_ptr throws in many of its functions.

R-Goc: The thing is wrt uses exceptions. Com_ptr throws in many of its functions.

ayeteadoe: > Com_ptr throws in many of its functions.

I checked https://learn.microsoft.com/en-us/uwp/cpp-ref-for-winrt/com-ptr and the definition in `base.h` itself, and I don't think there is a concern here. The methods on the public API that throw (`capture()` and `as()`) have `try_*` non-throw equivalents if they are ever needed (I wouldn't use either for any Direct3D APIs).

PR #6109, opened 2025-09-07
LibWeb: Backspace+modifiers respects punctuation in text inputs

kaiwildberger:
Implements Ctrl+Backspace and Super+Backspace to delete words individually or entire text entries respectively.

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLMediaElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLMediaElement.cpp
@@ -1227,7 +1227,7 @@ WebIDL::ExceptionOr<void> HTMLMediaElement::process_media_data(Function<void(Str
         //    named resize at the media element.
         if (video_track && is<HTMLVideoElement>(*this)) {
             auto duration = video_track ? video_track->duration() : audio_track->duration();
-            set_duration(static_cast<double>(duration.to_milliseconds()) / 1000.0);
+            set_duration(duration.to_seconds_f64());
 
             auto& video_element = as<HTMLVideoElement>(*this);
             video_element.set_video_width(video_track->pixel_width());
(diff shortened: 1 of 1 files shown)


Libraries/LibWeb/Page/EventHandler.cpp
```diff
@@ -1283,12 +1283,32 @@ EventResult EventHandler::handle_keydown(UIEvents::KeyCode key, u32 modifiers, u
     auto* target = document->active_input_events_target();
     if (target) {
         if (key == UIEvents::KeyCode::Key_Backspace) {
+            if ((modifiers & UIEvents::KeyCode::Key_LeftControl) != 0) {
```
konradekk: nit: virtually just noting that probably depends on the taste and might be os-dependent, but you probably should check whether `modifiers` exactly match given combination, e.g. left control, not whether it contains it?

(it might be just more key combinations should be introduced here, in proper order, as each action is marked as handled, so current resolution is just okay.)

kaiwildberger:
Reopening this; I took out the local include and it should pass all the checks.

Libraries/LibWeb/Page/EventHandler.cpp
```diff
@@ -1284,12 +1284,12 @@ EventResult EventHandler::handle_keydown(UIEvents::KeyCode key, u32 modifiers, u
     auto* target = document->active_input_events_target();
     if (target) {
         if (key == UIEvents::KeyCode::Key_Backspace) {
-            if ((modifiers & UIEvents::KeyCode::Key_LeftControl) != 0) {
+            if ((modifiers & UIEvents::KeyModifier::Mod_PlatformWordJump) == UIEvents::KeyModifier::Mod_PlatformWordJump) {
```
gmta, line 1287: Please put this change in the commit that originally introduced it (i.e., the previous one) so you end up with a single commit.

Libraries/LibWeb/Page/EventHandler.cpp
gmta, line 1: Could you add a test for this?

PR #6111, opened 2025-09-07
LibWeb/IndexedDB: Throw an InvalidStateException when handling a removed index / object store

lukasxcs:
This fixes a bunch of FIXME's and passes a previously crashing subtest on WPT https://wpt.fyi/results/IndexedDB/idbindex-rename-errors.any.html.

### DIFF
--- a/Libraries/LibWeb/IndexedDB/IDBDatabase.cpp
+++ b/Libraries/LibWeb/IndexedDB/IDBDatabase.cpp
@@ -186,6 +186,8 @@ WebIDL::ExceptionOr<void> IDBDatabase::delete_object_store(String const& name)
     // 7. Destroy store.
     database->remove_object_store(*store);
 
+    store->mark_deleted();
+
     return {};
 }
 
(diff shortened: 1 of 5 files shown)


lukasxcs:
The tests run into a timeout after correctly throwing an InvalidStateException for readonly transactions. Not sure why that happens.

Libraries/LibWeb/IndexedDB/IDBDatabase.cpp
```diff
@@ -186,6 +186,8 @@ WebIDL::ExceptionOr<void> IDBDatabase::delete_object_store(String const& name)
     // 7. Destroy store.
     database->remove_object_store(*store);
 
+    store->mark_deleted();
```
konradekk, line 189: this suggestion might be plainly wrong, but: did you consider calling this from `remove_object_store` above?

lukasxcs: That won't compile because of forward decleration.. `error: member access into incomplete type`.

No idea how to fix that, though i like your idea as it adds validation to the removal per `if (remove) mark_deleted()`

PR #6112, opened 2025-09-07
LibWeb: Store opacity properties in computed form

Calme1709:


Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -3314,6 +3316,30 @@ NonnullRefPtr<StyleValue const> StyleComputer::compute_border_or_outline_width(N
...
+NonnullRefPtr<StyleValue const> StyleComputer::compute_opacity(NonnullRefPtr<StyleValue const> const& specified_value, PropertyValueComputationContext const& computation_context)
+{
+    // https://drafts.csswg.org/css-color-4/#transparency
+    // specified number, clamped to the range [0,1]
+
+    // <number>
+    if (specified_value->is_number())
+        return NumberStyleValue::create(clamp(specified_value->as_number().number(), 0, 1));
+
+    // AD-HOC: We also support calc()'d numbers
```
AtkinsSJ: This is a nit, but I wouldn't call these "ad-hoc". It's just the nature of `<number>` and `<percentage>` that they can be a calculation.

Calme1709: Updated to "NOTE" rather than "AD-HOC"

tcl3 (approved):

PR #6115, opened 2025-09-07
LibJS: Track current shape dictionary generation in PropertyLookupCache

Lubrsi:
When an object becomes too big (currently 64 properties or more), we
change its shape to a dictionary and don't do any further transitions.

However, this means the Shape of the object no longer changes, so the
cache invalidation check of `current_shape != cache.shape` is no longer
a valid check.

This fixes that by keeping track of a generation number for the Shape
both on the Shape object and in the cache, allowing that to be checked
instead of the Shape identity. The generation is incremented whenever
the dictionary is mutated.

Fixes stale cache lookups on Gmail preventing emails from being
displayed.

I was not able to produce a reproduction for this, plus the generation
count was over the 20k mark on Gmail.

### DIFF
--- a/Libraries/LibJS/Bytecode/Executable.h
+++ b/Libraries/LibJS/Bytecode/Executable.h
@@ -43,6 +43,7 @@ struct PropertyLookupCache {
         Optional<u32> property_offset;
         GC::Weak<Object> prototype;
         GC::Weak<PrototypeChainValidity> prototype_chain_validity;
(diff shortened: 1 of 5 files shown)


awesomekling:
The idea is that dictionary shapes remain valid cache-wise as long as you’re only adding properties, since the property IDs are then just serially incrementing. Is the issue that we have some kinda invalidation bug when using the delete operator perhaps?

PR #6117, opened 2025-09-07
LibWeb: Use ResizeObserver instance as this in callback function

kalenikaliaksandr:
ResizeObserver spec doesn't seem to mention this behavior explicitly,
but https://wpt.live/resize-observer/observe.html test expects it (we
can't import the WPT test because it hangs for some reason).


kalenikaliaksandr:
> > we can't import the WPT test because it hangs for some reason
> 
> The test freezes because http://wpt.live/resize-observer/resources/iframe.html doesn't get downloaded automatically. If you download this file manually with `curl` the test should work.

thanks, I managed to import the test by manually downloading iframe.html

PR #6118, opened 2025-09-07
LibWeb: Use AbstractElement more

AtkinsSJ:
Not everywhere, but this is what I got done when I had a spare bit of time.


AtkinsSJ:
Changes: Rebased, and then filled in more of the places that used separate Element and PseudoElement parameters. The main place this is awkward is when working with animations - I think AbstractElement should have helpers for those, but that would currently mean pulling in a lot of headers for every user of AbstractElement. I'd like to avoid that, but it'll take a bit more figuring out.

AtkinsSJ:
Change: Fixed a sily bug caught by sanitizers, where I was storing a reference to the contents of a temporary Optional. :sweat_smile:

PR #6122, opened 2025-09-08
LibWeb: Throw range error when initial is greater than maximum

me-it-is:
constructing WebAssembly.Memory if initial is greater than maximum a range error will be thrown.

fixes
https://wpt.fyi/results/wasm/jsapi/memory/constructor.any.worker.html?product=ladybird Initial value exceeds maximum


me-it-is:
It seems like the sanitizers failing is unrelated to this pr.

Libraries/LibWeb/WebAssembly/Memory.cpp
```diff
@@ -23,6 +23,13 @@ WebIDL::ExceptionOr<GC::Ref<Memory>> Memory::construct_impl(JS::Realm& realm, Me
     auto& vm = realm.vm();
 
     // https://webassembly.github.io/threads/js-api/index.html#dom-memory-memory
+    // 3. If maximum is not empty and maximum < initial, throw a RangeError exception.
+    if (descriptor.maximum.has_value()) {
+        if (descriptor.maximum.value() < descriptor.initial) {
+            return vm.throw_completion<JS::RangeError>("Initial is larger than maximum."sv);
+        }
+    }
```
gmta: You don't need to nest these conditions:

```cpp
if (descriptor.maximum.has_value() && descriptor.maximum.value() < descriptor.initial)
    return vm.throw_completion<JS::RangeError>("Initial is larger than maximum."sv);
```

Libraries/LibWeb/WebAssembly/Memory.cpp
```diff
@@ -23,6 +23,13 @@ WebIDL::ExceptionOr<GC::Ref<Memory>> Memory::construct_impl(JS::Realm& realm, Me
     auto& vm = realm.vm();
```
alimpfard, line 23: Please rewrap the commit message and use proper capitalisation.
(maybe something like "Fixes the 'Initial value exceeds maximum' test in\n\<that url\>"?)

alimpfard:
Thank you!

PR #6131, opened 2025-09-08
LibWeb: Add activeElement attribute in ShadowRoot

F3n67u:


Libraries/LibWeb/DOM/DocumentOrShadowRoot.h
```diff
@@ -0,0 +1,50 @@
...
+
+    // 3. If candidate's root is not this, then return null.
+    if (&candidate->root() != &self)
+        return nullptr;
+
+    // 4. If candidate is not a Document object, then return candidate.
+    if (!is<Document>(candidate))
+        return as<Element>(candidate);
+
+    auto* candidate_document = static_cast<Document*>(candidate);
```
tcl3: `as<T>()`/`as_if<T>()` should be used where possible instead of `static_cast`

```suggestion
    auto* candidate_document = as_if<Document>(candidate);
    if (!candidate_document)
        return as<Element>(candidate);
```

Libraries/LibWeb/DOM/DocumentOrShadowRoot.h
```diff
@@ -0,0 +1,50 @@
...
+
+    // 5. If candidate has a body element, then return that body element.
+    if (candidate_document->body())
+        return candidate_document->body();
+
+    // 6. If candidate's document element is non-null, then return that document element.
+    if (candidate_document->document_element())
+        return candidate_document->document_element();
```
tcl3: If-init statements can be used here to avoid repetition:
```suggestion
    if (auto* body = candidate_document->body())
        return body;

    // 6. If candidate's document element is non-null, then return that document element.
    if (auto* document_element = candidate_document->document_element())
        return document_element;
```

Libraries/LibWeb/DOM/DocumentOrShadowRoot.h
```diff
@@ -0,0 +1,50 @@
...
+#include <LibWeb/DOM/Utils.h>
+
+namespace Web::DOM {
+
+template<typename T>
+concept DocumentOrShadowRoot = OneOf<T, Document, ShadowRoot>;
+
+// https://html.spec.whatwg.org/multipage/interaction.html#dom-documentorshadowroot-activeelement
+template<DocumentOrShadowRoot T>
+Element* calculate_active_element(T& self)
```
tcl3: `GC::Ptr<T>`  shouuld be preferred to `T*` where possible, so the return type of this function should be `GC::Ptr<DOM::Element>`

Libraries/LibWeb/DOM/ShadowRoot.h
```diff
@@ -52,6 +52,8 @@ class WEB_API ShadowRoot final : public DocumentFragment {
 
     WebIDL::ExceptionOr<String> get_html(GetHTMLOptions const&) const;
 
+    Element* active_element();
```
tcl3: `GC::Ptr<DOM::Element>`should be used as the return type here.

Libraries/LibWeb/DOM/DocumentOrShadowRoot.h
F3n67u: Thanks a lot for reviewing this! I resolved all those comments. please take another look.

tcl3 (approved):

PR #6133, opened 2025-09-08
LibWeb: Position absolute block-level boxes below previous siblings in inline formatting contexts

### LINKED ISSUE #5798: LibWeb: Incorrect absolute position inside inline container
### Summary

Reduction below. Reduced from [https://osu.ppy.sh/beatmapsets](https://osu.ppy.sh/beatmapsets), line is supposed to be under "beatmaps":

Image 

I successfully failed to fix it myself, so here's a litle bit of what I've figured out:

Layout tree is incorrect - `div.b` is on the same level as `span.a`, while it should be inside.
```
16466.493 WebContent(190440): Viewport <#document> at (0,0) content-size 286x29 [0+0+0 286 0+0+0] [0+0+0 29 0+0+0] children: not-inline
  BlockContainer <html> at (0,0) content-size 286x34.015625 [0+0+0 286 0+0+0] [0+0+0 34.015625 0+0+0] [BFC] children: not-inline
    BlockContainer <body> at (8,8) content-size 270x18.015625 [8+0+0 270 0+0+8] [8+0+0 18.015625 0+0+8] children: inline
      InlineNode <span.a> [0+0+0 74.375 0+0+0] [0+0+0 18 0+0+0]
        frag 0 from TextNode start: 1, length: 8, rect: [8,8 74.375x18] baseline: 15.21875
            "beatmaps"
        TextNode <#text>
        TextNode <#text>
      BlockContainer <div.b> at (8,29) content-size 286x3 positioned [0+0+0 286 0+0+0] [0+0+0 3 0+0+0] [BFC] children: not-inline
      TextNode <#text>

```
But even after fixing this part, containing block for `div.b` was `#document` instead of span. Aaand, even after hardcoding it to be a span for testing purposes, `div.b` is incorrectly positioned  - I suspect due to this piece of code:
https://github.com/LadybirdBrowser/ladybird/blob/b15c4cccbf92d4eb80096868da2931e72e1b2fa1/Libraries/LibWeb/Layout/InlineFormattingContext.cpp#L476-L503
Here we set our x for `div.b` to be the width of the last fragment:
`x = last_fragment->offset().x() + last_fragment->width();`, but it's not always the case - x should be set to 0 when we are positioning **block** element. 

This monstrosity fixed this part, but I'm fairly sure that's not how it should be done, so I'm leaving this for smart layout people 😅
```cpp
if (last_fragment) {
            bool originally_block_outside = true;

            if (is<DOM::Element>(box.dom_node())) {
                auto const& element = static_cast<DOM::Element const&>(*box.dom_node());
                if (auto cascaded_props = element.cascaded_properties({})) {
                    if (auto original_display_value = cascaded_props->property(CSS::PropertyID::Display)) {
                        if (original_display_value->is_display()) {
                            originally_block_outside = original_display_value->as_display().display().is_block_outside();
                        }
                    }
                }
            }

            if (originally_block_outside) {
                // block-level absolutely positioned element: start of new line
                x = 0;
                y = last_fragment->offset().y() + last_fragment->height();
            } else {
                // inline-level absolutely positioned element: positioned inline after previous content
                x = last_fragment->offset().x() + last_fragment->width();
                y = last_fragment->offset().y() + last_fragment->height();
            }
        }
```

### Operating system

Linux

### Steps to reproduce

N/A

### Expected behavior

N/A

### Actual behavior

N/A

data:text/html;base64,PCFET0NUWVBFIGh0bWw+DQo8c3R5bGU+DQogIC5hIHsNCiAgICBwb3NpdGlvbjogcmVsYXRpdmU7DQogIH0NCg0KICAuYiB7DQogICAgaGVpZ2h0OiAzcHg7DQogICAgcG9zaXRpb246IGFic29sdXRlOw0KICAgIHRvcDogMTAwJTsNCiAgICB3aWR0aDogMTAwJTsNCiAgICBiYWNrZ3JvdW5kLWNvbG9yOiBjcmltc29uOw0KICB9DQo8L3N0eWxlPg0KPHNwYW4gY2xhc3M9ImEiPg0KICBiZWF0bWFwcw0KICA8ZGl2IGNsYXNzPSJiIj48L2Rpdj4NCjwvc3Bhbj4=

```html
<!DOCTYPE html>
<style>
  .a {
    position: relative;
  }

  .b {
    height: 3px;
    position: absolute;
    top: 100%;
    width: 100%;
    background-color: crimson;
  }
</style>
<span class="a">
  beatmaps
  <div class="b"></div>
</span>
```

```shell
N/A
```

_No response_

_No response_

### Contribute a patch?

Pasi4K5:
Fixes #5798

Whether an absbox is positioned below or to the right of its previous sibling in an `InlineFormattingContext` is determined by the display-outside value before blockification, so we store the pre-blockification `display` value in `ComputedValues` to access it in `InlineFormattingContext` and position the box accordingly.

Before:
image

After:
image

### DIFF
--- a/Libraries/LibWeb/CSS/ComputedProperties.h
+++ b/Libraries/LibWeb/CSS/ComputedProperties.h
@@ -186,6 +186,9 @@ public:
     MixBlendMode mix_blend_mode() const;
     Optional<FlyString> view_transition_name() const;
 
+    Display display_before_box_type_transformation() const;
+    void set_display_before_box_type_transformation(Display value);
+
     static Vector<Transformation> transformations_for_style_value(StyleValue const& value);
     Vector<Transformation> transformations() const;
     TransformBox transform_box() const;
@@ -278,6 +281,8 @@ private:
     Array<u8, ceil_div(number_of_longhand_properties, 8uz)> m_animated_property_inherited {};
 
     HashMap<PropertyID, NonnullRefPtr<StyleValue const>> m_animated_property_values;
+
+    Display m_display_before_box_type_transformation { InitialValues::display() };
 
     int m_math_depth { InitialValues::math_depth() };
     RefPtr<Gfx::FontCascadeList const> m_font_list;
--- a/Libraries/LibWeb/CSS/StyleComputer.cpp
+++ b/Libraries/LibWeb/CSS/StyleComputer.cpp
@@ -2154,6 +2154,8 @@ void StyleComputer::compute_property_values(ComputedProperties& style) const
 
         style.set_property(property_id, absolutized_value, is_inherited);
     });
+
+    style.set_display_before_box_type_transformation(style.display());
 }
 
 void StyleComputer::resolve_effective_overflow_values(ComputedProperties& style) const
--- a/Libraries/LibWeb/Layout/InlineFormattingContext.cpp
+++ b/Libraries/LibWeb/Layout/InlineFormattingContext.cpp
@@ -503,8 +503,10 @@ StaticPositionRect InlineFormattingContext::calculate_static_position_rect(Box c
 
     CSSPixelPoint position;
     if (auto const* sibling = box.previous_sibling()) {
-        // We're calculating the position for an absolutely positioned box with a previous sibling in an IFC. We need to
-        // position the box at the top right corner of the last fragment of this sibling.
+        // We're calculating the position for an absolutely positioned box with a previous sibling in an IFC.
+        // We need to position the box...
+        // ...below the last fragment of this sibling, if the display-outside value (before box type transformation) is block.
+        // ...at the top right corner of the last fragment of this sibling otherwise.
         LineBoxFragment const* last_fragment = nullptr;
         auto const& cb_state = m_state.get(*sibling->containing_block());
         for (auto const& line_box : cb_state.line_boxes) {
@@ -514,8 +516,15 @@ StaticPositionRect InlineFormattingContext::calculate_static_position_rect(Box c
             }
         }
         if (last_fragment) {
-            position.set_x(last_fragment->offset().x() + last_fragment->width());
-            position.set_y(last_fragment->offset().y());
+            if (box.display_before_box_type_transformation().is_block_outside()) {
+                // Display-outside value is block => position below
+                position.set_x(0);
+                position.set_y(last_fragment->offset().y() + last_fragment->height());
+            } else {
+                // Display-outside value is not block => position to the right
+                position.set_x(last_fragment->offset().x() + last_fragment->width());
+                position.set_y(last_fragment->offset().y());
+            }
         }
     }
 
--- a/Libraries/LibWeb/CSS/ComputedProperties.cpp
+++ b/Libraries/LibWeb/CSS/ComputedProperties.cpp
@@ -132,6 +132,16 @@ void ComputedProperties::revert_property(PropertyID id, ComputedProperties const
     set_property_inherited(id, style_for_revert.is_property_inherited(id) ? Inherited::Yes : Inherited::No);
 }
 
+Display ComputedProperties::display_before_box_type_transformation() const
+{
+    return m_display_before_box_type_transformation;
+}
+
+void ComputedProperties::set_display_before_box_type_transformation(Display value)
+{
+    m_display_before_box_type_transformation = value;
+}
+
 void ComputedProperties::set_animated_property(PropertyID id, NonnullRefPtr<StyleValue const> value, Inherited inherited)
 {
     m_animated_property_values.set(id, move(value));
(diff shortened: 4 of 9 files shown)


Libraries/LibWeb/Layout/InlineFormattingContext.cpp
```diff
@@ -515,8 +518,25 @@ StaticPositionRect InlineFormattingContext::calculate_static_position_rect(Box c
...
+
+            if (is<DOM::Element>(box.dom_node())) {
+                auto const& element = static_cast<DOM::Element const&>(*box.dom_node());
+                if (auto const& specified_display_prop = element.cascaded_properties({})->property(CSS::PropertyID::Display)) {
+                    if (specified_display_prop->is_display())
+                        block_outside_specified = specified_display_prop->as_display().display().is_block_outside();
+                }
+            }
+
+            if (block_outside_specified) {
```
awesomekling: No need to go through the DOM here; you can access the display directly via `box.display()` :)

Pasi4K5: `box.display()` returns the *computed* display value, but because it's an absbox, that is always `block`, even if, for example, we have an element `<span style="position:absolute">foo</span>` or `<div style="position:absolute;display:inline">foo</div>`.

To be honest, I don't know why that's the way it works, but Chromium does the same thing, so I was assuming this is intentional behaviour.

Going over the DOM allows us to determine the *specified* value instead. Of course there might be an easier way that I am unaware of.

awesomekling: Layout should always act on the computed value, so it sounds like there's something here we need to understand better.

Note that `display: block` means `display: block flow`, so `is_block_outside()` will still return true, btw!

awesomekling: > To be honest, I don't know why that's the way it works, but Chromium does the same thing, so I was assuming this is intentional behaviour.

Indeed, this is a process called blockification. Very strange if you don't know what's going on 😅 

See CSS-DISPLAY-3 here: https://drafts.csswg.org/css-display/#transformations:

> Absolute positioning or floating an element [blockifies](https://drafts.csswg.org/css-display/#blockify) the box’s display type.

awesomekling: As far as I can tell, both the testcase and the original reported bug are fixed if you simply use `box.display()` :)

Pasi4K5: > As far as I can tell, both the testcase and the original reported bug are fixed if you simply use `box.display()` :)

Correct, but it causes the test `Layout/input/block-and-inline/abspos-inline.html` to fail, because now it *always* positions the box below, which it shouldn't do if the specified value isn't `box`.

Here's a reduced case that doesn't work as expected with `box.display()`:
```html
<!DOCTYPE html>
foo<span style="position:absolute">bar</span>
```
`foo` and `bar` *should* be next to each other (or at least they are in Chromium and Firefox), but with `box.display()`, `bar` is below `foo`.

awesomekling: Ah of course, because we want to be looking at the display value as it was before blockification saw `position:absolute` and turned `display:inline` into `display:block`.

I think what we really want here is to save the pre-blockification display value and keep it alongside the final display value in the `ComputedValues` (i.e what `box.display()` ends up getting the data from). That's a couple extra steps, but it will allow us to more accurately model the "static position" concept, which is supposed to be peeking into an alternate reality where the box had `position:static` instead of `position:absolute`.

Libraries/LibWeb/CSS/ComputedProperties.h
```diff
@@ -184,6 +184,9 @@ class WEB_API ComputedProperties final : public JS::Cell {
     MixBlendMode mix_blend_mode() const;
     Optional<FlyString> view_transition_name() const;
 
+    Display specified_display() const;
+    void set_specified_display(Display value);
```
awesomekling: Technically this is not the *specified* `display` since it cannot be things like `inherit`, `initial`, etc. Can we find a more precise/accurate name, something like `display_before_box_type_transformation` perhaps?

Pasi4K5: sure

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -2251,6 +2251,10 @@ void StyleComputer::compute_property_values(ComputedProperties& style) const
     });
 
     style.set_line_height({}, line_height);
+
+    if (auto const& display = style.property(PropertyID::Display); display.is_display()) {
```
Calme1709: As far as I can tell display will always be a `DisplayStyleValue` since that is all that the parser can produce and we have already handled CSS-wide keywords (`initial`, `inherit`, etc) so this `if` is unnecessary.

PR #6135, opened 2025-09-09
LibWeb: Suppress rendering due to view transitions

Psychpsyo:
This also fixes a bug in the view transitions code that was required to get the imported test to pass. The code for setting the initial containing block size just did not set the right thing, since doing so would trigger an error later on. (I incorrectly believed that the line setting the size was the problem for a bit...)

That later error resulted from walking up the tree, without considering that the document element has a parent that is not itself an element. (and then doing element things to it)


gmta (approved):

Psychpsyo:
Test failure is an unrelated flake

PR #6136, opened 2025-09-09
LibWeb: Parse "positional-value-list-shorthands` directly

Calme1709:
Continues the work started in https://github.com/LadybirdBrowser/ladybird/pull/5386 to simplify handling of positional value list shorthand properties.

Previously we would parse these as `StyleValueList`s and then rely on `StyleComputer::for_each_property_expanding_shorthands` to expand them into longhands.

This required a bit of work to handle `ShorthandStyleValue`s for the `@page` `margin` descriptor.


Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
```diff
@@ -1552,6 +1552,15 @@ RefPtr<StyleValue const> Parser::parse_border_value(PropertyID property_id, Toke
         VERIFY_NOT_REACHED();
     }
 
+    auto const make_single_value_shorthand = [&](PropertyID property_id, Vector<PropertyID> const& longhands, ValueComparingNonnullRefPtr<StyleValue const> const& value) {
+        Vector<ValueComparingNonnullRefPtr<StyleValue const>> longhand_values;
+
+        for (size_t i = 0; i < longhands.size(); i++)
+            longhand_values.append(value);
```
gmta: I think you should be able to do `longhand_values.resize_with_default_value(longhands.size(), value);` here

Calme1709: Done - thanks for the tip.

gmta (approved):

PR #6137, opened 2025-09-09
LibWeb: Make word-spacing and letter-spacing default value compute to 0

tcl3:
This PR ensures that a specified value of `normal` maps to a computed value of 0 for the `word-spacing` and `letter-spacing` properties.

`letter-spacing` also has an unusual legacy behavior where a computed value of 0 maps to a resolved value of `normal` (i.e: the value returned by `getComputedStyle()`). We now implement this behavior too.

Fixes 81 WPT subtests in the `css/css-text` directory.


Libraries/LibWeb/CSS/ComputedProperties.cpp
```diff
@@ -946,20 +946,22 @@ WhiteSpaceTrimData ComputedProperties::white_space_trim() const
     VERIFY_NOT_REACHED();
 }
 
-Optional<LengthOrCalculated> ComputedProperties::letter_spacing() const
+Optional<LengthPercentage> ComputedProperties::letter_spacing() const
```
Calme1709: We should be resolving percentages against the computed font size rather than the glyph width of the ASCII space character as we do below in `InlineLevelIterator::next_without_lookahead()`.

If we do that here we can produce a CSSPixels rather than a LengthPercentage which will save us the work of resolving that during layout. There's also no reason for this to be Optional since we always return a value (or crash).

```c++
CSSPixels ComputedProperties::letter_spacing() const
{
    auto const& value = property(PropertyID::LetterSpacing);
    if (value.is_keyword() && value.to_keyword() == Keyword::Normal)
        return 0;

    if (value.is_length())
        // NOTE: This length was absolutized in `StyleComputer::compute_property_values`
        return value.as_length().length().absolute_length_to_px();

    if (value.is_percentage())
        return font_size().scale_by(value.as_percentage().percentage().as_fraction());

    if (value.is_calculated())
        // NOTE: We don't need to pass a length resolution context as we have already absolutized any relative lengths
        //       in `StyleComputer::compute_property_values`
        return value.as_calculated().resolve_length({ .percentage_basis = Length::make_px(font_size()), .length_resolution_context = {} })->absolute_length_to_px();

    VERIFY_NOT_REACHED();
}
```

tcl3: I've made this change and a similar change for the `word-spacing` property.

Tests/LibWeb/Text/expected/css/CSSStyleProperties-all-supported-properties-and-default-values.txt
```diff
@@ -735,8 +735,8 @@ All supported properties and their default values exposed from CSSStylePropertie
 'will-change': 'auto'
```
Calme1709, line 735: This test doesn't pass at this commit, perhaps the following commit should be moved before this one?
```diff
-'letterSpacing': 'normal'
-'letter-spacing': 'normal'
+'letterSpacing': '0px'
+'letter-spacing': '0px'
```

Tests/LibWeb/Text/expected/css/getComputedStyle-print-all.txt
```diff
@@ -65,7 +65,7 @@ text-wrap-style: auto
 visibility: visible
```
Calme1709, line 65: Doesn't pass, same as above

```diff
-letter-spacing: normal
+letter-spacing: 0px
```

Libraries/LibWeb/CSS/CSSStyleProperties.cpp
```diff
@@ -815,6 +815,13 @@ RefPtr<StyleValue const> CSSStyleProperties::style_value_for_computed_property(L
     }
     case PropertyID::WebkitTextFillColor:
         return resolve_color_style_value(get_computed_value(property_id), layout_node.computed_values().webkit_text_fill_color());
+    case PropertyID::LetterSpacing: {
+        // https://drafts.csswg.org/css-text-4/#letter-spacing-property
+        // For legacy reasons, a computed letter-spacing of zero yields a resolved value (getComputedStyle() return value) of normal.
+        if (auto value = layout_node.computed_values().letter_spacing(); value == 0)
```
Calme1709: No need to assign to `value`

Calme1709:
Looks good to me!

PR #6138, opened 2025-09-09
LibWeb: Properly clamp interpolated opacity values

Calme1709:
Opacity values are unique in that the range which calculated and interpolated values should be clamped to [0,1] is different from the range of allowed values [-∞,∞].

This fixes 28 WPT tests that were regressed in #6112


Calme1709:
Marked as draft until #6072 is merged as this is reliant on c94615d40e589d86555edae74d12d705af817c35

Calme1709:
Rebased and marked as ready to review after preceding PR was merged.

Meta/Lagom/Tools/CodeGenerators/LibWeb/GenerateCSSPropertyID.cpp
```diff
@@ -891,35 +891,42 @@ AcceptedTypeRangeMap property_accepted_type_ranges(PropertyID property_id)
 
             StringBuilder ranges_builder;
 
-            for (auto& type : valid_types.values()) {
-                VERIFY(type.is_string());
+            // Opacity values are unique in that the range which calculated and interpolated values should be clamped
+            // to [0,1] is different from the range of allowed values [-∞,∞]. To handle this we set the allowed range
+            // in Properties.json to [-∞,∞] but overwrite it to [0,1] here.
```
AtkinsSJ, line 896: Can you mark this as a FIXME? With this, we'd now have different functions `property_accepts_number/percentage()` and `property_accepted_type_ranges()` which sound very similar, but produce different results. I don't like that much, even though it does solve the problem and those functions are used at different stages.

These properties all take [`<opacity-value>`](https://drafts.csswg.org/css-color-4/#typedef-opacity-opacity-value), so it might be possible to instead parse/compute that in a special way instead. Though I'm struggling to think how that would work.

Calme1709: I have added a FIXME:

I think it is fine for these functions to return different values as they represent different things - `property_accepts_{number,percentage}` relating to specified values and `property_accepted_type_ranges` relating to computed values - perhaps this should be resolved by using clearer names for these functions?

AtkinsSJ: Thanks. And yeah changing the names would be enough to make it clear. (Names are hard :sweat_smile:)

AtkinsSJ:
Seems like a suspicious coincidence, because I can't see why your changes would cause this, but Linux/Clang/Sanitizers just had a timeout on this test:

`Screenshot/input/opacity-stacking.html (duration: 27 minutes 53.656 seconds)`

I'll restart it, and maybe keep an eye out to see if that flakes again.

PR #6140, opened 2025-09-09
LibWasm: Validate faster darn it

alimpfard:


Hendiadyoin1 (commented):
The last commit seems really dirty, and I don't think that the addition call calls to `value` instead of going through the `operator==` really makes a difference

alimpfard:
~~Unfortunately it does, I forgot to mention that but operator== was actually ~3% of runtime.~~
~~it might not be important anymore though, since we're not doing O(n^2) calls to it :P~~

it's not, it was just showing up as hot because we were spamming it so hard, there's no difference now with or without.

alimpfard:
Looks like I've broken more stuff than I expected, will draft for now.

Libraries/LibWasm/AbstractMachine/BytecodeInterpreter.cpp
```diff
@@ -25,6 +25,15 @@
...
 using namespace AK::SIMD;
 
+// ASAN allocates frames for all the not-explicitly-tail-called functions,
+// which blows out the stack. So disable direct threading when ASAN is enabled.
+// We use the explicit annotation on clang, so allow it there.
+#if defined(HAS_ADDRESS_SANITIZER) && !defined(AK_COMPILER_CLANG)
+constexpr static auto should_try_to_use_direct_threading = false;
+#else
+constexpr static auto should_try_to_use_direct_threading = true;
+#endif
```
Hendiadyoin1, line 45: Didn't you add the missing tail call annotations, or are there still some missing?

alimpfard: None missing, but gcc<15 doesn't have the annotation.

Hendiadyoin1: asking because of the ASAN part still there, not the GCC portion

alimpfard: `HAS_TAILCALL` is only undefined on GCC<=14.

Hendiadyoin1: Ah sorry, misread/interpreted the `#if` condition, mb

AK/Vector.h
```diff
@@ -503,6 +503,44 @@ requires(!IsRvalueReference<T>) class Vector {
         update_metadata();
     }
 
+    /// The iterator pair identify a set of indices *in ascending order* to be removed.
+    template<typename It, IteratorPairWith<It> EndIt, typename ToIndex = decltype (*declval<It>()) (*)(It const&)>
+    void remove_all(It&& it, EndIt const& end, ToIndex const& to_index = [](It const& it) -> decltype(auto) { return *it; })
+    {
+        if (!(it != end))
```
R-Goc, line 514:
```cpp
if (it == end)
```
seems simply cleaner here, in the verify at the bottom as well.

alimpfard: Iterators aren't guaranteed to have operator==, just operator!=

Hendiadyoin1: Doesn't https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2468r2.html fix this?
(Not 100% sure but I remember it changing some overload resolution and transforms)

alimpfard: AFAIK all current proposals derive from ==, no explicit == means no ==

alimpfard:
I'm gonna merge this, haven't hit any issues with it browsing around anyway.

PR #6142, opened 2025-09-09
LibWeb: Improve interpolation of mixed percentage-dimension values

tcl3:
If we are interpolating between a dimension and a percentage value and the dimension component is 0, we now return a percentage value rather than a `calc()` value.

Fixes at least 94 WPT interpolation subtests.


tcl3:
Rebased to include newly passing tests after #6146 was merged.

Libraries/LibWeb/CSS/Interpolation.cpp
```diff
@@ -1100,9 +1100,47 @@ static RefPtr<StyleValue const> interpolate_value_impl(DOM::Element& element, Ca
...
+                case StyleValue::Type::Length:
+                    return value.as_length().raw_value();
+                case StyleValue::Type::Number:
+                    return value.as_number().number();
+                case StyleValue::Type::Percentage:
+                    return value.as_percentage().raw_value();
+                case StyleValue::Type::Time:
+                    return value.as_time().raw_value();
+                default:
+                    VERIFY_NOT_REACHED();
```
AtkinsSJ, line 1120: Observation: Looking at this, maybe it *would* be useful to have a base `DimensionStyleValue` type. :thinking:

Libraries/LibWeb/CSS/Interpolation.cpp
```diff
@@ -1100,9 +1100,47 @@ static RefPtr<StyleValue const> interpolate_value_impl(DOM::Element& element, Ca
...
+                }
+            };
+
             auto from_node = to_calculation_node(from);
             auto to_node = to_calculation_node(to);
```
AtkinsSJ: These aren't used until after your new code, so they should probably come after it to save some work.

Libraries/LibWeb/CSS/Interpolation.cpp
tcl3: Having the equivalent of `raw_value()` on a base class would certainly be easier than what I'm doing here :smile:

PR #6149, opened 2025-09-10
LibWeb: Update animation type of compositing properties

Calme1709:
This updates `background-blend-mode`, `mix-blend-mode`, and `isolation` to be animation type `discrete` rather than `none`. This reflects https://github.com/w3c/fxtf-drafts/commit/169b0e1 but isn't reflected on the published spec as the published spec is out of date - see https://github.com/w3c/fxtf-drafts/issues/547

Gains us 26 WPT tests.


tcl3 (approved):
Thanks for tracking this down! I made the same fix but couldn't find anything in the specs to justify it.

PR #6154, opened 2025-09-10
LibWeb: Simplify interpolation of mixed percentage-dimension values

tcl3:
This is a non-functional change that simplifies mixed percentage-dimension interpolation by casting the relevant StyleValues to `DimensionStyleValue`.

After reading [Sam's comment](https://github.com/LadybirdBrowser/ladybird/pull/6142#discussion_r2337048596) on #6142, I didn't realize that this simplification was already possible :sweat_smile:


AtkinsSJ:
> After reading [Sam's comment](https://github.com/LadybirdBrowser/ladybird/pull/6142#discussion_r2337048596) on #6142, I didn't realize that this simplification was already possible 😅

:rofl: This is my daily reminder that my memory is completely awful.

AtkinsSJ:
Oh! Actually, `DimensionStyleValue` doesn't cover the `number` case, so this is a behaviour change. Though, I don't think number should be included here anyway as it's about `<dimension-percentage>` types and `<number-percentage>` isn't a thing. I guess we can see if WPT cares or not.

tcl3:
I agree, Number and Integer shouldn't have been included previously. I did run all of the WPT interpolation tests and there was no change.

PR #6155, opened 2025-09-10
LibWeb: Add support for HTMLHRElement size presentational hint

veeti:


Libraries/LibWeb/HTML/HTMLHRElement.cpp
```diff
@@ -71,6 +72,40 @@ void HTMLHRElement::apply_presentational_hints(GC::Ref<CSS::CascadedProperties>
...
             }
         }
+
+        // If an hr element has either a color attribute or a noshade attribute, and furthermore also has a size attribute,
+        // and parsing that attribute's value using the rules for parsing non-negative integers doesn't generate an error,
+        // then the user agent is expected to use the parsed value divided by two as a pixel length for presentational hints
+        // for the properties 'border-top-width', 'border-right-width', 'border-bottom-width', and 'border-left-width' on the element.
+        bool has_color_or_noshade = has_attribute(HTML::AttributeNames::color) || has_attribute(HTML::AttributeNames::noshade);
+        if (name == HTML::AttributeNames::size && has_color_or_noshade) {
+            auto size = parse_non_negative_integer(value);
```
tcl3: This doesn't appear necessary. You should be able to use `parsed_value` instead of `size` below.

Libraries/LibWeb/HTML/HTMLHRElement.cpp
```diff
@@ -71,6 +72,40 @@ void HTMLHRElement::apply_presentational_hints(GC::Ref<CSS::CascadedProperties>
...
+            if (auto parsed_value = parse_non_negative_integer(value); parsed_value.has_value()) {
+                auto size_value = CSS::LengthStyleValue::create(CSS::Length::make_px(size.value() / 2));
```
tcl3: This performs an integer division. Should we be dividing by `2.0` instead?

veeti:
Resolved comments.

PR #6156, opened 2025-09-10
LibJS: Remove incorrect VERIFY

caztanj:
This VERIFY is both incorrect and redundant. The VERIFY at step 2 verifies the status when evaluate is called on m_cycle_root.

### DIFF
--- a/Libraries/LibJS/CyclicModule.cpp
+++ b/Libraries/LibJS/CyclicModule.cpp
@@ -344,7 +344,6 @@ ThrowCompletionOr<GC::Ref<Promise>> CyclicModule::evaluate(VM& vm)
     if ((m_status == ModuleStatus::EvaluatingAsync || m_status == ModuleStatus::Evaluated) && m_cycle_root != this) {
         // Note: This will continue this function with module.[[CycleRoot]]
         VERIFY(m_cycle_root);
-        VERIFY(m_cycle_root->m_status == ModuleStatus::Linked);
         dbgln_if(JS_MODULE_DEBUG, "[JS MODULE] evaluate[{}](vm) deferring to cycle root at {}", this, m_cycle_root.ptr());
         return m_cycle_root->evaluate(vm);
     }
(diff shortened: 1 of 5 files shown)


Libraries/LibJS/CyclicModule.cpp
```diff
@@ -342,7 +342,6 @@ ThrowCompletionOr<GC::Ref<Promise>> CyclicModule::evaluate(VM& vm)
     if ((m_status == ModuleStatus::EvaluatingAsync || m_status == ModuleStatus::Evaluated) && m_cycle_root != this) {
         // Note: This will continue this function with module.[[CycleRoot]]
         VERIFY(m_cycle_root);
-        VERIFY(m_cycle_root->m_status == ModuleStatus::Linked);
```
gmta, line 347: Step 2 verifies `m_status`, not `m_cycle_root->m_status` - but even then, completely different conditions are checked. So I don't really get why this should be removed?

caztanj: `ModuleStatus::Linked`, `ModuleStatus::EvaluatingAsync`, `ModuleStatus::Evaluated` are all valid states for `m_cycle_root->m_status`.  `evaluate` is being called on `m_cycle_root` at the end of the if block. In this recursive call to `evaluate`, `m_status` is the correct `m_status`.

index.html
```
<!doctype html>
<html>
	<head>
		<script src="/index.mjs"></script>
	</head>
</html>
```

index.mjs
```
(async function() {
	const first = import("./a.mjs");
	const second = import("./b.mjs");

	await first;
	await second;

	console.log("Both imports done.");
})();
```

a.mjs
```
import { foo } from "./b.mjs";

console.log("A: Start evaluation.");

await new Promise(resolve => setTimeout(resolve, 200));

console.log("A: After await.");
```

b.mjs
```
import "./a.mjs";

console.log("B: Start evaluation.");

export const foo = "value from B";
```

gmta: Right - in your example, continuing with `m_cycle_root` in `::evaluate()` will have its status set to `ModuleStatus::EvaluatingAsync` instead of `::Linked`, causing a crash.

Could you please add that test case as a regression crash test?

PR #6158, opened 2025-09-10
LibWeb: Require layout update for less properties in getComputedStyle()

kalenikaliaksandr:
Some properties like `justify-items`, `grid`, or `display` do affect layout, but their used values can be obtained without performing a layout calculation.

This change introduces a new helper,
`property_needs_layout_for_getcomputedstyle()`, specifically for use by `CSSStyleProperties::property()`. It returns true only for properties such as `width`, `height`, `margin`, `padding`, `top`, and `left`, where an up-to-date layout is required to return the correct used value.

### DIFF
--- a/Libraries/LibWeb/CSS/Properties.json
+++ b/Libraries/LibWeb/CSS/Properties.json
@@ -489,7 +489,8 @@
       "mapping": "block-size"
     },
     "initial": "auto",
-    "max-values": 1
+    "max-values": 1,
+    "needs-layout-for-getcomputedstyle": true
   },
   "border": {
     "inherited": false,
@@ -498,7 +499,8 @@
       "border-width",
       "border-style",
       "border-color"
-    ]
+    ],
+    "needs-layout-for-getcomputedstyle": true
   },
   "border-block": {
     "inherited": false,
@@ -507,7 +509,8 @@
       "border-block-width",
       "border-block-style",
       "border-block-color"
-    ]
+    ],
+    "needs-layout-for-getcomputedstyle": true
   },
   "border-block-color": {
     "inherited": false,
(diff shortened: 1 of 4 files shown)


awesomekling:
Great idea and execution! I have a worse version of this in my stash somewhere, but yours is much cleaner :)

Libraries/LibWeb/CSS/Properties.json
```diff
@@ -2078,25 +2099,28 @@
...
       "mapping": "block-start"
     },
-    "max-values": 1
+    "max-values": 1,
+    "needs-layout-for-getcomputedstyle": true
   },
   "inset-inline": {
     "initial": "auto",
-    "positional-value-list-shorthand": true,
+    "positional-value-list-shorthand": false,
```
Zaggy1024: I noticed this, was this intended/should it be in a separate commit?

kalenikaliaksandr: Oops, nope, it was added by accident. Thanks for catching this.

Calme1709:
I believe this new property/function should be documented in [CSSGeneratedFiles.md](https://github.com/LadybirdBrowser/ladybird/blob/master/Documentation/CSSGeneratedFiles.md)

Meta/Lagom/Tools/CodeGenerators/LibWeb/GenerateCSSPropertyID.cpp
```diff
@@ -714,6 +715,36 @@ bool property_affects_stacking_context(PropertyID property_id)
...
+)~~~");
+
+    properties.for_each_member([&](auto& name, auto& value) {
+        VERIFY(value.is_object());
+        if (is_legacy_alias(value.as_object()))
+            return;
+
+        bool needs_used_value_for_cssom_read = false;
+        if (value.as_object().has("needs-layout-for-getcomputedstyle"sv))
+            needs_used_value_for_cssom_read = value.as_object().get_bool("needs-layout-for-getcomputedstyle"sv).value_or(false);
```
AtkinsSJ: The `has()` seems redundant. `get_bool()` will return an empty Optional if there's nothing there.

PR #6159, opened 2025-09-11
LibMedia: Use FFmpeg's info to tell us if YUVJ is full-range

Zaggy1024:
It reports this correctly for the video linked in #4993, and likely any other videos using the YUVJ format, so we don't need this extra check.

### DIFF
--- a/Libraries/LibMedia/FFmpeg/FFmpegVideoDecoder.cpp
+++ b/Libraries/LibMedia/FFmpeg/FFmpegVideoDecoder.cpp
@@ -168,8 +168,9 @@ DecoderErrorOr<NonnullOwnPtr<VideoFrame>> FFmpegVideoDecoder::get_decoded_frame(
             case AV_PIX_FMT_YUV422P12:
             case AV_PIX_FMT_YUV444P12:
                 return 12;
+            default:
+                VERIFY_NOT_REACHED();
             }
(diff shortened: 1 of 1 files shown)


Libraries/LibMedia/FFmpeg/FFmpegVideoDecoder.cpp
```diff
@@ -168,8 +168,9 @@ DecoderErrorOr<NonnullOwnPtr<VideoFrame>> FFmpegVideoDecoder::get_decoded_frame(
             case AV_PIX_FMT_YUV422P12:
             case AV_PIX_FMT_YUV444P12:
                 return 12;
+            default:
+                VERIFY_NOT_REACHED();
             }
-            VERIFY_NOT_REACHED();
```
gmta, line 181: We usually leave out the `default:` case in switches if possible, since if you put in an enum you'll get a compilation error if you add a new enum value. This is not a switch on an enum though, but that's the reason why they look like this :)

Zaggy1024: Right. In this case, with it not being an enum class, it warns about possible values not being covered by the switch statement, which is why I put in the default case. Definitely something to keep in mind though!

PR #6167, opened 2025-09-12
LibWeb: Handle interpolation of BorderRadius style value type

MFMF-EGY:


Libraries/LibWeb/CSS/Interpolation.cpp
```diff
@@ -1320,6 +1321,20 @@ static RefPtr<StyleValue const> interpolate_value_impl(DOM::Element& element, Ca
 
         return StyleValueList::create(move(interpolated_values), from_list.separator());
     }
+    case StyleValue::Type::BorderRadius: {
```
tcl3: Please keep the cases in this switch statement in alphabetical order.

tcl3 (commented):
Per the [contributing guidelines](https://github.com/LadybirdBrowser/ladybird/blob/master/CONTRIBUTING.md), commit messages should be written in the imperative mood. So in this case, something like this would be better: `LibWeb: Handle interpolation of BorderRadius style value type`.

Libraries/LibWeb/CSS/Interpolation.cpp
MFMF-EGY: Done.

Libraries/LibWeb/CSS/Interpolation.cpp
```diff
@@ -1193,6 +1194,20 @@ static RefPtr<StyleValue const> interpolate_value_impl(DOM::Element& element, Ca
...
+        auto const& from_vertical_radius = from.as_border_radius().vertical_radius();
+        auto const& to_vertical_radius = to.as_border_radius().vertical_radius();
+        auto const interpolated_horizontal_radius = interpolate_length_percentage(calculation_context, from_horizontal_radius, to_horizontal_radius, delta);
+        auto const interpolated_vertical_radius = interpolate_length_percentage(calculation_context, from_vertical_radius, to_vertical_radius, delta);
+        if (!interpolated_horizontal_radius.has_value() || !interpolated_vertical_radius.has_value())
+            return {};
+        return BorderRadiusStyleValue::create(
+            interpolated_horizontal_radius.value(),
+            interpolated_vertical_radius.value()
+            );
```
tcl3: The linter on CI is complaining about this.
```suggestion
            interpolated_vertical_radius.value());
```

tcl3 (approved):

Issue #566, opened 2024-07-11
Text selection breaks with floated elements


vpzomtrrfrt:
```html
<!DOCTYPE html>
<html>
	<head>
		<style>
			.left { float: left }
		</style>
	</head>
	<body>
		<nav class="left">
			All Local
		</nav>
		<p>what</p>
	</body>
</html>
```

Text selection seems generally wrong when viewing this page.
- dragging over "what" does nothing
- double-clicking "what" selects "Local" instead
- selecting "Local" and then copying actually copies "ll Lo"

vpzomtrrfrt:
Confirmed. The third issue is still present as of 8986e1f

AtkinsSJ:
Pinging @trflynn89 and @gmta as you've both been working on related things recently.

gmta:
Reduced case:

```html
<!DOCTYPE html>
<div>    All Local</div>
```

Looks like the leading whitespace gets removed, and we're indexing into the DOM string instead of the paintable string for the copy data.

This also results in broken selections:

```html
<!DOCTYPE html>
<div>    All Local</div>
<script>
    const divElm = document.querySelector('div');
    const range = document.createRange();
    range.setStart(divElm.firstChild, 4);
    range.setEnd(divElm.firstChild, 7);
    document.getSelection().addRange(range);
    console.log(range.toString());
</script>
```

PR #2789, opened 2024-12-05
LibWeb: Bring whitespace handling closer to spec

AtkinsSJ:
This replaces our ad-hoc code for collapsing whitespace, with the spec algorithm.

It doesn't quite do what I wanted, which is to fix innerText. To quote the [spec](https://html.spec.whatwg.org/multipage/dom.html#rendered-text-collection-steps):

> If node is a [Text](https://dom.spec.whatwg.org/#interface-text) node, then for each CSS text box produced by node, in content order, compute the text of the box after application of the CSS ['white-space'](https://drafts.csswg.org/css-text/#white-space-property) processing rules and ['text-transform'](https://drafts.csswg.org/css-text/#text-transform-property) rules, set items to the [list](https://infra.spec.whatwg.org/#list) of the resulting strings, and return items. The CSS ['white-space'](https://drafts.csswg.org/css-text/#white-space-property) processing rules are slightly modified: collapsible spaces at the end of lines are always collapsed, but they are only removed if the line is the last line of the block, or it ends with a [br](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-br-element) element. Soft hyphens should be preserved.

It's not clear to me how this could be implemented. I think we need to actually do all of this text processing at a different level - elsewhere the spec describes processing an entire inline formatting context at once, which would make it more possible to know if a `<br>` element is there, or to collapse whitespace across element boundaries. For now I've had enough. :sweat_smile:

PR #6169, opened 2025-09-12
LibWeb: Store correct text offsets in PaintableFragment

gmta:
Previously, we were collapsing whitespace in Layout::TextNode and then passed the resulting string for further processing through ChunkIterator -> InlineLevelIterator -> InlineFormattingContext -> LineBuilder -> LineBoxFragment -> PaintableFragment. Our painting tree is where we deal with things like range offsets into the underlying text nodes, but since we modified the original string, the offsets were wrong.

This changes the way we generate fragments:

  * Layout::TextNode no longer collapses whitespace as part of its stored "text for rendering", but moves this logic to ChunkIterator which splits up this text into separate views whenever whitespace needs to be collapsed.

  * Layout::LineBox now only extends the last fragment if its end offset is equal to the new fragment's start offset. Otherwise, there's a gap caused by collapsing whitespace and we need to generate a separate fragment for that in order to have a correct start offset.

Some tests need new baselines because of the fixed start offsets.

Fixes https://github.com/LadybirdBrowser/ladybird/issues/566.

### DIFF
--- a/Libraries/LibWeb/Layout/TextNode.cpp
+++ b/Libraries/LibWeb/Layout/TextNode.cpp
@@ -282,10 +282,9 @@ static Utf16String apply_text_transform(Utf16String const& string, CSS::TextTran
         return apply_math_auto_text_transform(string);
     case CSS::TextTransform::Capitalize:
         return string.to_titlecase(locale, TrailingCodePointTransformation::PreserveExisting);
-    case CSS::TextTransform::FullSizeKana: {
+    case CSS::TextTransform::FullSizeKana:
         dbgln("FIXME: Implement text-transform full-size-kana");
         return string;
-    }
     case CSS::TextTransform::FullWidth:
         return string.to_fullwidth();
     }
@@ -306,7 +305,6 @@ Utf16String const& TextNode::text_for_rendering() const
     return *m_text_for_rendering;
 }
 
-// NOTE: This collapses whitespace into a single ASCII space if the CSS white-space property tells us to.
 void TextNode::compute_text_for_rendering()
 {
     if (dom_node().is_password_input()) {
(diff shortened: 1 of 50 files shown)


Libraries/LibWeb/Layout/TextNode.cpp
```diff
@@ -280,17 +280,15 @@ static Utf16String apply_text_transform(Utf16String const& string, CSS::TextTran
         return string;
     case CSS::TextTransform::MathAuto:
         return apply_math_auto_text_transform(string);
-    case CSS::TextTransform::Capitalize: {
+    case CSS::TextTransform::Capitalize:
         return string.to_titlecase(locale, TrailingCodePointTransformation::PreserveExisting);
-    }
     case CSS::TextTransform::FullSizeKana: {
-        // FIXME: Implement this!
+        dbgln("FIXME: Implement text-transform full-size-kana");
```
trflynn89, line 286: nit: You could remove the braces around this case as well

Libraries/LibWeb/Layout/TextNode.h
```diff
@@ -45,6 +45,10 @@ class TextNode final : public Node {
         ChunkIterator(TextNode const&, bool wrap_lines, bool respect_linebreaks);
         ChunkIterator(TextNode const&, Utf16View const&, Unicode::Segmenter&, bool wrap_lines, bool respect_linebreaks);
 
+        bool wrap_lines() const { return m_wrap_lines; }
+        bool respect_linebreaks() const { return m_respect_linebreaks; }
+        bool collapse_whitespace() const { return m_collapse_whitespace; }
```
AtkinsSJ: These functions all sound more like actions than flag getters. Looks like the previous code put `do_` on the front for `respect_linebreaks` at least.

Libraries/LibWeb/Layout/TextNode.cpp
```diff
@@ -306,61 +306,91 @@ Utf16String const& TextNode::text_for_rendering() const
...
-    }
+    // https://drafts.csswg.org/css-text-4/#white-space-phase-1
+    bool convert_newlines = false;
+    bool convert_tabs = false;
+
+    // If white-space-collapse is set to collapse or preserve-breaks, white space characters are considered collapsible
+    // and are processed by performing the following steps:
+    auto white_space_collapse = computed_values().white_space_collapse();
+    if (first_is_one_of(white_space_collapse, CSS::WhiteSpaceCollapse::Collapse, CSS::WhiteSpaceCollapse::PreserveBreaks)) {
+        // 1. FIXME: Any sequence of collapsible spaces and tabs immediately preceding or following a segment break is removed.
```
AtkinsSJ, line 338: I have an old draft PR where I tried implementing a lot of this: https://github.com/LadybirdBrowser/ladybird/pull/2789

I had issues with it, where it was failing a lot of tests and I wasn't sure how to resolve that. But maybe it's useful to you in some way?

Libraries/LibWeb/Layout/TextNode.cpp
```diff
@@ -306,61 +306,91 @@ Utf16String const& TextNode::text_for_rendering() const
...
-            builder.append_code_unit(data.code_unit_at(index));
-            ++index;
+    // AD-HOC: It's important to not change the amount of code units in the resulting transformed text, so ChunkIterator
+    //         can pass views to this string with associated code unit offsets that still match the original text.
+    if (convert_newlines || convert_tabs) {
+        StringBuilder text_builder { StringBuilder::Mode::UTF16, text.length_in_code_units() };
+        for (auto code_point : text) {
+            if (code_point == '\n' && convert_newlines)
+                code_point = ' ';
+            if (code_point == '\t' && convert_tabs)
```
AtkinsSJ: `else if`?

Libraries/LibWeb/Layout/TextNode.cpp
```diff
@@ -306,61 +306,91 @@ Utf16String const& TextNode::text_for_rendering() const
...
-    while (index < data.length_in_code_units()) {
-        if (is_ascii_space(data.code_unit_at(index))) {
-            builder.append(' ');
-            ++index;
-            skip_over_whitespace();
-        } else {
-            builder.append_code_unit(data.code_unit_at(index));
-            ++index;
+    // AD-HOC: It's important to not change the amount of code units in the resulting transformed text, so ChunkIterator
+    //         can pass views to this string with associated code unit offsets that still match the original text.
```
trflynn89, line 382: This comment caught my eye - does this account for the text-transform applied above changing the code point length? For example:
```html
<p style="text-transform: uppercase;">well     ß   hello   friends</p>
```

Renders as:


If I select "WELL SS HELLO FR":


Then we are off-by-one in what we consider to be selected:
```js
> window.getSelection().toString()
"well     ß   hello   fri" 
```

trflynn89: (of course this is still more accurate than how we currently behave on master:)
```js
window.getSelection().toString()
"well     ß   hel" 
```

Libraries/LibWeb/Layout/TextNode.h
gmta: Those `do_` prefixes came from another place, these methods reflected the names of the members so I kept that theme going. Added `should_` as a prefix to all three of them!

Libraries/LibWeb/Layout/TextNode.cpp
gmta: Even better, just a single `if`!

Libraries/LibWeb/Layout/TextNode.cpp
gmta: I could and I did!

Libraries/LibWeb/Layout/TextNode.cpp
gmta: Ooh nice find, the fact that we add or remove code units depending on `text-transform` definitely means this isn't the best solution. In your example, we should render "SS" but treat them as a single code unit as far as selections go.

We would need a different structure for that, e.g. keep track of code unit offsets for fragments that do not map 1 to 1 to the original string. Let's do that in a new PR 😅

gmta:
Some WPT diffs I could find:

* `css/css-text/white-space`: +14 subtest passes
* `html/dom/elements`: +86 subtest passes

Libraries/LibWeb/Layout/TextNode.cpp
gmta: I'm not seeing anything very relevant immediately, but then again, this PR is mostly tightening the nuts on some behaviors that were very clearly wrong - we still need a bigger refactor to support the rest of the white-space collapsing algo.

Libraries/LibWeb/Layout/TextNode.cpp
trflynn89: > Let's do that in a new PR 😅

Agreed! Can we just add a FIXME here so we don't lose track of it?

gmta: I've added a FIXME a bit higher up, pointing to #6177.

PR #6172, opened 2025-09-12
Meta+Everywhere: Add a few cleanups for FreeBSD and enable WebGL

cqundefine:


Libraries/LibCore/Environment.cpp
gmta, line 1: Commit message typo: "[..] like we already for [..]"

gmta (approved):

PR #2820, opened 2024-12-07
Serialize getComputedStyle() values in a separate mode so we can simplify StyleComputer

awesomekling:
`CSSStyleValue::to_string()` now takes a `SerializationMode` which can be either `Normal` or `ResolvedValue`.

The only difference in `ResolvedValue` mode is that we serialize RGB/RGBA color values as `rgb()` or `rgba()`, but never as the original named color string, even if they originated from one.

This allows us to remove an ad-hoc pass in `StyleComputer` that would rewrite all colors in an element's computed style.

PR #6174, opened 2025-09-12
LibWeb/CSS: Fix compilation with CSS_TRANSITIONS_DEBUG

socram8888:
A new parameter was added to Web::CSS::StyleValue::to_string() in PR #2820 but this debug message was never updated. If CSS_TRANSITIONS_DEBUG was enabled, compilation would fail.


socram8888:
Thank you! It seems there are some checks that fail though seem to be unrelated to my changes (it's on JS code I haven't changed).

PR #6182, opened 2025-09-13
LibWeb/IndexedDB: Replace spin_until with async observers

Lubrsi:
This removes all spin_until uses in IndexedDB, where it was common for infinite spins to happen, especially in close_a_database_connection waiting for all transactions to finish.

This allows the Google Docs signed in homepage to load, as it was stuck in such an infinite spin.

### DIFF
--- a/Libraries/LibWeb/IndexedDB/Internal/Algorithms.cpp
+++ b/Libraries/LibWeb/IndexedDB/Internal/Algorithms.cpp
@@ -181,10 +181,16 @@ void open_a_database_connection(JS::Realm& realm, StorageAPI::StorageKey storage
                 }
 
                 db->wait_for_connections_to_close(open_connections, GC::create_function(realm.heap(), [&realm, connection, version, request, on_complete] {
+                    dbgln_if(IDB_DEBUG, "open_a_database_connection: finished waiting for step 10.5");
+
                     // 6. Run upgrade a database using connection, version and request.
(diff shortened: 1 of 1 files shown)


gmta:
CC @stelar7

stelar7:
Looks good, i think this should cover the main part that is going wrong ([lifetime handling](https://w3c.github.io/IndexedDB/#transaction-lifetime) step 5)

I think being async here might go against the spec, as a spin_until is ment to freeze any further execution of the event loop (?)

Lubrsi:
As far as I can tell, the majority of these algorithms are running "in parallel", which means they should run at the same time alongside the event loop, which to me says we shouldn't freeze the event loop when waiting on a parallel condition.

The main oddity is IDBDatabase::close, which only calls `close_a_database_connection`, implies a synchronous close by not returning a promise, but MDN (note that it's non-authoritative) implies that IDBDatabase::close is an asynchronous operation as implemented in engines. https://developer.mozilla.org/en-US/docs/Web/API/IDBDatabase/close
> The close() method of the [IDBDatabase](https://developer.mozilla.org/en-US/docs/Web/API/IDBDatabase) interface returns immediately and closes the connection in a separate thread.

The only other user of `close_a_database_connection` is `open_a_database_connection` when closing the connection when the upgrade errors. `open_a_database_connection` runs this in parallel, which makes me think `close_a_database_connection` should be an asynchronous operation.

stelar7:
Ive left an issue on their github https://github.com/w3c/IndexedDB/issues/474, maybe its an oversight

PR #6183, opened 2025-09-13
LibWeb: Interpolate some basic shape values

tcl3:
Fixes 260 WPT subtests.


Libraries/LibWeb/CSS/Interpolation.cpp
```diff
@@ -1238,6 +1238,26 @@ static RefPtr<StyleValue const> interpolate_value_impl(DOM::Element& element, Ca
                     return;
                 interpolated_shape = Rect { *interpolated_rect_box };
             },
+            [&](Polygon const& from_polygon) {
```
AtkinsSJ: We could maybe use some spec quotes for these. `polygon()` at least has:

> If both shapes are of type [polygon()](https://drafts.csswg.org/css-shapes-1/#funcdef-basic-shape-polygon), both polygons have the same number of vertices, and use the same [<'fill-rule'>](https://svgwg.org/svg2-draft/painting.html#FillRuleProperty), interpolate between each value in the shape functions.

https://drafts.csswg.org/css-shapes-1/#basic-shape-interpolation

Though, a lot of these functions seem to be missing from there.

AtkinsSJ (commented):
Nice!

Libraries/LibWeb/CSS/Interpolation.cpp
tcl3: Relevant spec text added :^)

PR #6184, opened 2025-09-13
LibJS: Add PutByNumericId and change PutById to be string key only

kalenikaliaksandr:
Previously, PutById constructed a PropertyKey from the identifier,
which coerced numeric-like strings to numbers. This moves that decision
to bytecode generation: the bytecode generator now emits PutByNumericId
for numeric keys and PutById for string keys. This removes per-execution
parsing from the interpreter.

1.4x speedup on the following microbenchmark:
```js
const o = {};
for (let i = 0; i < 10_000_000; i++) {
    o.a = 1;
    o.b = 2;
    o.c = 3;
}
```


awesomekling:
Please add some tests to test-js that actually use the new opcodes. Or do we already have that?

kalenikaliaksandr:
> Please add some tests to test-js that actually use the new opcodes. Or do we already have that?

sure, writing a test made me realize property key cannot be numeric for PutByIdWithThis 😅

PR #6185, opened 2025-09-13
LibJS+LibWeb: Inline fast path for `Value::to_object()`

kalenikaliaksandr:
Adds inline implementation for the most common case when `Value` is already an object.

1.47x improvement on the following benchmark:
```js
const o = {};
for (let i = 0; i < 10_000_000; i++) {
    o.a = 1;
    o.b = 2;
    o.c = 3;
}
```

### DIFF
--- a/Libraries/LibJS/Runtime/ValueInlines.h
+++ b/Libraries/LibJS/Runtime/ValueInlines.h
@@ -48,4 +48,11 @@ inline ThrowCompletionOr<Value> Value::to_primitive(VM& vm, PreferredType prefer
     return to_primitive_slow_case(vm, preferred_type);
 }
 
+inline ThrowCompletionOr<GC::Ref<Object>> Value::to_object(VM& vm) const
+{
+    if (is_object())
+        return const_cast<Object&>(as_object());
+    return to_object_slow(vm);
+}
+
 }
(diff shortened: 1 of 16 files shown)


Libraries/LibJS/Runtime/ValueInlines.h
```diff
@@ -48,4 +48,11 @@ inline ThrowCompletionOr<Value> Value::to_primitive(VM& vm, PreferredType prefer
     return to_primitive_slow_case(vm, preferred_type);
 }
 
+inline ThrowCompletionOr<GC::Ref<Object>> Value::to_object(VM& vm) const
+{
+    if (is_object())
+        return const_cast<Object&>(as_object());
+    return to_object_slow(vm);
```
konradekk, line 55: for the record: you have `*_slow_case` in other places. shall other places also use this fast path? (slow case seem to be internal only for other types).

there's also a number check in the slow case; maybe it is enough to comibne these two instead and request `realm` in branches it is needed to achieve the same? 🤔

kalenikaliaksandr: > for the record: you have *_slow_case in other places. shall other places also use this fast path? (slow case seem to be internal only for other types).

`to_object_slow()` is used only by `Value::to_object()` with this change.

> there's also a number check in the slow case; maybe it is enough to comibne these two instead and request realm in branches it is needed to achieve the same? 🤔

sure, I think we could do that later. right now I don't have any use cases in mind that would benefit from doing that.

PR #6186, opened 2025-09-13
Tests: Enable previously disabled tests

lpas:
This PR enables some tests that previously crashed


gmta (approved):

AtkinsSJ:
These tests are timing out on 2 of the CI runs. Just enabling them again probably isn't a good idea until they can run in good time. (Or maybe they need configuring to be given >5 minutes, but IDK.)

PR #6189, opened 2025-09-13
AK: Fix is_within_range when converting from float

me-it-is:
Within range now uses the max capacity of a type rather than its size. 
This fixes some subtests in
https://wpt.fyi/results/wasm/core/conversions.wast.js.html?product=ladybird

when just using the size of the type it was assumed that if the destination was larger or the same size and both were signed that it would always fit witch is only true with integers


AtkinsSJ:
Looks good to me, but could you include a test for this? We have https://github.com/LadybirdBrowser/ladybird/blob/master/Tests/AK/TestChecked.cpp which you could add to.

me-it-is:
Test fails seem to be unrelated

PR #6190, opened 2025-09-14
LibWeb: Add initial support for `text-underline-position`

Calme1709:
This PR adds support initial support for `text-underline-position`, currently we only respect the `auto` and `under` values, leaving `from-font`, `left`, and `right` to implement.

This introduces the `TextUnderlinePositionStyleValue` class, it is possible to represent `text-underline-position` as a `StyleValueList` but would have required ugly workarounds for either serialization or in `ComputedProperties::text_underline_position`

Also includes a small fix around parsing shorthands which is included here as otherwise we begin passing the relevant test by coincidence thanks to adding a new `PropertyID` changing HashMap ordering.


Libraries/LibWeb/Painting/PaintableBox.cpp
```diff
@@ -850,10 +851,41 @@ void paint_text_decoration(DisplayListRecordingContext& context, TextPaintable c
...
+                case Web::CSS::TextUnderlinePositionHorizontal::Under:
+                    // The underline is positioned under the element’s text content. In this case the underline usually
+                    // does not cross the descenders. (This is sometimes called “accounting” underline.)
+                    return fragment.baseline() + CSSPixels { font.pixel_metrics().descent };
+                }
+
+                VERIFY_NOT_REACHED();
+            }();
+
+            dbgln_dump(text_underline_offset);
```
AtkinsSJ: This probably shouldn't still be here. :^)

AtkinsSJ (commented):
Nice!

Libraries/LibWeb/Painting/PaintableBox.cpp
Calme1709: Whoops, good catch.

AtkinsSJ:
Looks like you're passing more tests than expected, so this needs rebaselining. :tada:

Issue #6188, opened 2025-09-13
crash loading www.misen.com; assert in Web::Fetch::Infrastructure::Request::add_origin_header


danielbprice:
### Summary

misen.com is a shopping website.  Ladybird does a lot of network fetches trying to load this page, then crashes repeatably.

### Operating system

macOS

### Steps to reproduce

1. Install ladybird@0076ab74b26b4294736442d75f2bc77a97f3a854 
2. Load https://www.misen.com
3. Wait for crash

n.b. I tried hard to work out how to set a breakpoint in ak_assertion_failed in order to look at what is happening, but I couldn't work out how to get it to fire-- maybe because there are multiple processes, I dunno.  Do you have a guide to debugging ladybird?

### Expected behavior

Load page and display properly.

### Actual behavior

Crash every time.

### URL for a reduced test case

https://www.misen.com

### HTML/SVG/etc. source for a reduced test case

```html
N/A -- because I could not figure out how to inspect what resource was being loaded by http_fetch here, I couldn't really make an educated guess about how to start on reducing a test case, sorry!
```

### Log output and (if possible) backtrace

```shell
VERIFICATION FAILED: has<T>() at /Users/dp/ws/ladybird/AK/Variant.h:387
0   liblagom-ak.0.0.0.dylib             0x0000000103a81698 dump_backtrace + 72
1   liblagom-ak.0.0.0.dylib             0x0000000103ac60ac ak_trap + 12
2   liblagom-ak.0.0.0.dylib             0x0000000103a8199c ak_assertion_failed + 0
3   liblagom-web.0.0.0.dylib            0x0000000105cb75a0 Web::Fetch::Infrastructure::Request::add_origin_header() + 1784
4   liblagom-web.0.0.0.dylib            0x0000000105c94128 Web::Fetch::Fetching::http_network_or_cache_fetch(JS::Realm&, Web::Fetch::Infrastructure::FetchParams const&, Web::Fetch::Fetching::IsAuthenticationFetch, Web::Fetch::Fetching::IsNewConnectionFetch) + 2616
5   liblagom-web.0.0.0.dylib            0x0000000105c918c4 Web::Fetch::Fetching::http_fetch(JS::Realm&, Web::Fetch::Infrastructure::FetchParams const&, Web::Fetch::Fetching::MakeCORSPreflight) + 860
[...]

PR #6192, opened 2025-09-14
LibWeb/Fetch: Add missing spec step for checking for tuple origin

Lubrsi:
Fixes https://github.com/LadybirdBrowser/ladybird/issues/6188

### DIFF
--- a/Libraries/LibWeb/Fetch/Infrastructure/HTTP/Requests.cpp
+++ b/Libraries/LibWeb/Fetch/Infrastructure/HTTP/Requests.cpp
@@ -346,7 +346,7 @@ void Request::add_origin_header()
             case ReferrerPolicy::ReferrerPolicy::StrictOriginWhenCrossOrigin:
                 // If request’s origin is a tuple origin, its scheme is "https", and request’s current URL’s scheme is
                 // not "https", then set serializedOrigin to `null`.
-                if (m_origin.has<URL::Origin>() && m_origin.get<URL::Origin>().scheme() == "https"sv && current_url().scheme() != "https"sv)
+                if (m_origin.has<URL::Origin>() && !m_origin.get<URL::Origin>().is_opaque() && m_origin.get<URL::Origin>().scheme() == "https"sv && current_url().scheme() != "https"sv)
                     serialized_origin = MUST(ByteBuffer::copy("null"sv.bytes()));
                 break;
             // -> "same-origin"


Libraries/LibWeb/Fetch/Infrastructure/HTTP/Requests.cpp
```diff
@@ -346,7 +346,7 @@ void Request::add_origin_header()
             case ReferrerPolicy::ReferrerPolicy::StrictOriginWhenCrossOrigin:
                 // If request’s origin is a tuple origin, its scheme is "https", and request’s current URL’s scheme is
                 // not "https", then set serializedOrigin to `null`.
-                if (m_origin.has<URL::Origin>() && m_origin.get<URL::Origin>().scheme() == "https"sv && current_url().scheme() != "https"sv)
+                if (m_origin.has<URL::Origin>() && !m_origin.get<URL::Origin>().is_opaque() && m_origin.get<URL::Origin>().scheme() == "https"sv && current_url().scheme() != "https"sv)
```
kennethmyhra, line 349: Maybe we should extract the concept [tuple origin](https://html.spec.whatwg.org/multipage/browsers.html#concept-origin-tuple) into a helper like `is_opaque()`? Also I am a bit confused by us checking for only `https` and `http` schemes. The tuple origin concept says (among other things) that scheme should be an ASCII string and the definition I find for that is: _An ASCII string is a string whose code points are all ASCII code points_.

Psychpsyo: The scheme here is a string and we're only checking for HTTPS because this part of the spec is only interested in whether or not the origin's scheme is "https". The check for whether or not this is a tuple origin is only the `!m_origin.get<URL::Origin>().is_opaque()` and nothing else in this line.

kennethmyhra: Yeah, I do not know why I didn't see that when commenting, I got to focused on the tuple origin concept. Disregard my comment

danielbprice:
(I filed the bug).

I applied this fix, and retested.  It seems to help, although the website now crashes ("Ladybird flew off course") in a new and different way.  There is no backtrace on my console that I can find; I only see:

```
3399281.599 Ladybird(69585): WebContent process crashed! Last page loaded: https://misen.com/
```

I can see that this site is a horrible mess of trackers and other shopping analytics-ware.  Yuck.

gmta:
@danielbprice You can start Ladybird with `--debug-process WebContent` and attach a debugger to the reported PID to generate crash backtraces manually.

Issue #6173, opened 2025-09-12
Compiling with ENABLE_ALL_THE_DEBUG_MACROS is broken


socram8888:
### Summary

When compiling with `-DENABLE_ALL_THE_DEBUG_MACROS=ON`, the `WASI_DEBUG` flag is set. This causes a compilation failure in the WASI library.

### Operating system

Linux

### Steps to reproduce

1. Enable ENABLE_ALL_THE_DEBUG_MACROS by any means, such as patching `Meta/ladybug.py` to enable all debug macros. This will in turn enable WASI_DEBUG:
```diff
diff --git a/Meta/ladybird.py b/Meta/ladybird.py
index d58cbd3304..9c238b987c 100755
--- a/Meta/ladybird.py
+++ b/Meta/ladybird.py
@@ -215,6 +215,7 @@ def configure_main(platform: Platform, preset: str, cc: str, cxx: str) -> Path:
         build_preset_dir,
         f"-DCMAKE_C_COMPILER={cc}",
         f"-DCMAKE_CXX_COMPILER={cxx}",
+        f"-DENABLE_ALL_THE_DEBUG_MACROS=ON",
     ]
 
     if swiftc:
```
2. Attempt to compile with `BUILD_PRESET=Debug ./Meta/ladybird.py run -j6`

### Expected behavior

Code compiles

### Actual behavior

Compilation fails due to a bug in the WASI library.

### URL for a reduced test case

N/A

### HTML/SVG/etc. source for a reduced test case

```html
N/A
```

### Log output and (if possible) backtrace

```shell
user@Ubuntu:~/Desktop/ladybird$ BUILD_PRESET=Debug ./Meta/ladybird.py run -j6
ninja: Entering directory `/home/user/Desktop/ladybird/Build/debug'
[9/2772] Building CXX object Lagom/Libraries/LibWasm/CMakeFiles/LibWasm.dir/WASI/Wasi.cpp.o
FAILED: Lagom/Libraries/LibWasm/CMakeFiles/LibWasm.dir/WASI/Wasi.cpp.o 
[...]

PR #960, opened 2024-08-04
LibWasm: Bytecode interpreter performance pack

dzfrias:
Makes fib(30) 2 times faster.

PR #6197, opened 2025-09-14
LibWasm: Missing argument formatter (Wasm::Value) for WASI_DEBUG

rcorsi:
When building with WASI_DEBUG or ENABLE_ALL_THE_DEBUG_MACROS, build will fail (see #6173).

Required formatting code was remove in PR #960 for performance improvements.

This PR is removing the display of arguments in dbgln_if to allow WASI_DEBUG MACRO compilation to work.

Initial support for showing the Wasm function arguments when WASI_DEBUG was ON, is from a608e87


rcorsi:
@alimpfard Was wondering if you have any comments about this PR?  A while back you added the code to display the Wasi function arguments when compiled with WASI_DEBUG (a608e87ff9daa417b70546d2f766937227573b8b).  Over time the required code was removed for performance improvement (PR #960)

alimpfard:
eh, sure, we could probably print out the raw values instead still...but I've been removing that log every time I want to debug anyway :P

rcorsi:
Sorry if it's a dumb question, for raw values do you mean `arguments.data()` or something else?

rcorsi:
@alimpfard is there an easy test to run to see these WASI_DEBUG dbgln_if logs?  I can't see anything in the output with `WASI:`

alimpfard:
You'd need to run a wasi program with it, "easy" would be to grab e.g. zig1.wasm and try to run it.

Raw data as in printing [u128], each from Value::value().

rcorsi:
ok thanks for the suggestions. I ended up using emscripten and a really simple webassembly hello world program.  I was able to see the dbgln_if WASI_DEBUG logs using the wasm binary.

I'll see if I can get the raw arguments shown next.

alimpfard:
theoretically you just need to add an AK::Format formatter for Wasm::Value, Vector<T>'s formatter should handle the rest.

rcorsi:
I added the Wasm::Value formatter, this is what the output looks like for a very simple hello world that uses a printf statement, compiled via emscripten.

```
$ BUILD_PRESET=Debug Meta/ladybird.py run wasm -w -e main --export-noop ../wasm_hello/hello.wasm 
ninja: Entering directory `/home/bobo/projects/myladybird/Build/debug'
ninja: no work to do.
say_hello() to WebAssembly world
WASI: fd_write([ 1, fec0, 2, febc ]) = {Ok(33)}
Returned:
  -> 0
```

Does this look right?

alimpfard:
sure, it's definitely more useful than just printing the name of the call.
Thanks!

rcorsi:
@alimpfard do we need anything else to get the PR merged?  Does someone else need to review it?

alimpfard:
Nope, just wanted to wait for CI.

rcorsi:
Thank You!

PR #6200, opened 2025-09-15
LibWeb/Geometry: Correct typos in DOMMatrix "not 0 or -0" checks

AtkinsSJ:
Assuming 0 and -0 are distinct values, `a != 0 || a != -0` is always true.

I don't know enough about how -0 is represented to know if is actually distinct from 0, but let's write it correctly anyway. :^)


Gingeh:
Wikipedia says:
> According to the IEEE 754 standard, negative zero and positive zero should compare as equal with the usual (numerical) comparison operators, like the == operators of C and Java.

but it's still good to match the spec text.

PR #6207, opened 2025-09-15
LibJS: Rebind Window this to WindowProxy in OrdinaryCallBindThis

mikiubo:
Without this, functions would incorrectly see the raw Window object, which implements EventTarget but is not the object JavaScript code should interact with.
The HTML spec requires WindowProxy as the global this.


gmta:
@mikiubo doesn't compile in CI!

mikiubo:
> @mikiubo doesn't compile in CI!

I suspected this. On my local environment, I have to delete the build folder twice before it builds correctly, so I think some folder might be messed up.

The CI results confirm that something is wrong on my local setup.

I’ll take the chance to ask your opinion on the approach:
The WPT I’m trying to fix fails because the "_this_" used inside the event is a Window object instead of a WindowProxy.

Two approaches come to mind:

Recast the object before the function is called.

Pass a WindowProxy in the event chain.

At first, I tried approach 2, but it seems WindowProxy does not implement the EventTarget interface.
Now I’m trying approach 1, but I get that error—and I don’t think it’s correct to add an ad-hoc reference to an HTML object inside LibJS.

So now I’m leaning towards approach 2.
Let me know what you think—thanks a lot!
@gmta

mikiubo:
@gmta As I mentioned, I changed the approach.
It seems cleaner now. The CI passes, but I’m not sure if all the other WPT tests are fine.
Do you know of any tests that could be impacted by this code change?

Let me know.
Thanks.

Libraries/LibWeb/DOM/EventDispatcher.cpp
```diff
@@ -89,7 +90,14 @@ bool EventDispatcher::inner_invoke(Event& event, Vector<GC::Root<DOM::DOMEventLi
...
         // 11. Call a user object’s operation with listener’s callback, "handleEvent", « event », and event’s currentTarget attribute value.
         // FIXME: These should be wrapped for us in call_user_object_operation, but it currently doesn't do that.
-        auto* this_value = event.current_target().ptr();
+        JS::Object* this_value = event.current_target().ptr();
+
+        // AD-HOC: Rebind `this_value` to the WindowProxy object if it is a Window.
+        if (is<HTML::Window>(this_value)) {
+            auto& window = as<HTML::Window>(global);
+            this_value = window.self().ptr();
+        }
```
gmta: Looking at the `FIXME` above this code, maybe this needs to be handled in `WebIDL::call_user_object_operation()` instead? There's another FIXME in there related to the wrapping of arguments, which I guess covers Window/WindowProxy as well?

mikiubo: Do you mean the fixme at step 10?

gmta: I mean the FIXME at step 11 here, and then the FIXME at step 10 in the `call_user_object_operation` algo. That seems to reference [converting arguments](<https://whatpr.org/webidl/1437.html#dfn-convert-idl-to-javascript-value>) such as platform objects to JS objects, which for Window might mean using WindowProxy instead.

mikiubo: Ok ok  I thought that step 10 in call_user_ refers to args not the this_value my error.

Another concern That I have: 
Is there any problem if I import the libweb class inside a libjs?

gmta: No, you're right in that it only mentions the arguments, not thisValue. Maybe this is a spec issue or intentionally left as an implementation detail, but we need to figure out _where_ in the spec that is.

mikiubo: Ok I understand.
Maybe this evening I will try to research deeper . Thanks for support!

mikiubo: I looked into it a bit more, and from the documentation it seems that call_user_object_operation already expects a wrapped object.
It also looks like the wrapping shouldn’t be done where I placed it.
The two options that appear to be specified in the docs are:

- in the get_parent chain of the event_target, or

- in the functions that convert objects into JS::Value.

I’m not sure yet, I’ll look into it further.

mikiubo: I wasn’t able to find a point where it’s explicitly stated that the windowproxy object should be wrapped. I tried also checking the Firefox code but without success.

As I had already noted in a previous comment, I would avoid doing the wrapping at the get_parent level since the windowproxy object does not inherit from EventTarget.

The last remaining option is that the wrapping happens when a window object is cast to JS::Object and JS::Value, but I think an action like this could have major impacts in other parts of the code.

Before implementing this solution, I’d like to hear your opinion, as I don’t want to have missed something.

@gmta  Thanks.

gmta: Apologies it took so long to get back to this - I've found a very relevant test case that isn't fixed by this PR:

```html
<!DOCTYPE html>
<script>
window.addEventListener('load', e => {
    console.log(e.currentTarget === window);
});
</script>
```

Outputs `false` in Ladybird, but `true` in Firefox and Chrome.

Apparently we're failing to initialize event's `.currentTarget` with WindowProxy instead of Window somewhere. This is also the "this value" that's being used by `EventDispatcher::inner_invoke()`. Seems like we should fix that instead!

gmta: Just to add a data point: Firefox has [EventTarget::GetTargetForDOMEvent()](https://searchfox.org/firefox-main/source/dom/events/EventTarget.h#156-162), which according to the doc returns `this` for everything except `Window`, which returns `WindowProxy` instead.

mikiubo: ok thanks for the suggestion! 
so i think the way is to make the windowproxy inherits from EventTarget ?
Sorry if this is a silly question as you notice i am not so  familiar (but i enjoy learning new things!)

anyway i will try this way
thaks a lot!

gmta: No, I think you should find the place where an event's `currentTarget` is set, and implement an override there.

Libraries/LibWeb/DOM/Event.h
```diff
@@ -98,8 +98,9 @@ class WEB_API Event : public Bindings::PlatformObject {
     u16 event_phase() const { return m_phase; }
     void set_phase(Phase phase) { m_phase = phase; }
 
-    GC::Ptr<EventTarget> current_target() const { return m_current_target; }
-    void set_current_target(EventTarget* current_target) { m_current_target = current_target; }
+    GC::Ptr<JS::Object> current_target() const;
```
gmta: We usually prioritize the C++ API by marking the method for binding with "*_for_bindings()", e.g.: `current_target_for_bindings()`. That way, we can keep the existing `current_target()` and `set_current_target()` methods like they were.

Libraries/LibWeb/DOM/Event.cpp
```diff
@@ -263,4 +265,17 @@ Vector<GC::Root<EventTarget>> Event::composed_path() const
...
+{
+
+    if (!m_current_target) {
+        return m_current_target;
+    }
+    if (is<HTML::Window>(*m_current_target)) {
+        auto& window = as<HTML::Window>(*m_current_target);
+        return window.window();
+    }
+    return current_target_internal();
```
gmta: This method can be simplified to:

```cpp
    if (auto* window = as_if<HTML::Window>(m_current_target.ptr()))
        return window->window();
    return m_current_target;
```

Libraries/LibWeb/DOM/Event.idl
```diff
@@ -9,7 +9,7 @@ interface Event {
     readonly attribute DOMString type;
     readonly attribute EventTarget? target;
     readonly attribute EventTarget? srcElement; // legacy
-    readonly attribute EventTarget? currentTarget;
+    readonly attribute Object? currentTarget;
```
gmta: This IDL definition came from the spec, so we usually don't modify this! Can you confirm that WindowProxy in other browsers does indeed not implement EventTarget?

mikiubo: I think i misinterpreted your response 

Screenshot From 2025-09-29 14-24-12

i think you already checked the firefox you mentioned and you want me to not implement the eventtarget in windowproxy.

sorry for my bad comunication skill.
i will check the code of other browser and fix.

gmta: Well I'm not sure either, I haven't checked Firefox' WindowProxy's interfaces. Maybe WindowProxy _should_ be an EventTarget? That's what you need to find out in order for the IDL spec to make sense.

gmta: @mikiubo Did you check what other browsers do with WindowProxy? Seems like you have decided on making WindowProxy an EventTarget.

Could you add an AD-HOC comment explaining to current_target_for_bindings() for future reference?

mikiubo: It took me some time, but I think I found the point where WindowProxy extends EventTarget.
Screenshot From 2025-10-05 11-45-36

I also think I found this comment in Chrome’s code:
Screenshot From 2025-10-05 11-44-37



These convinced me to make that modification.
I didn’t notify you earlier because I’m currently having an issue with the sanitizer build (I only noticed it when running the CI, not locally).

I think I have to change this:
`class WEB_API WindowProxy final : public DOM::EventTarget {
    JS_OBJECT(WindowProxy, JS::Object);`
    
 because is not a js object anymore but when i change it i get a lot of error related to the lifecycle of the windowproxy reference as platform object.
 
because it’s no longer a JS object.
However, when I change it, I get a lot of errors related to the lifecycle of the WindowProxy reference as a platform object.

Anyway,
@gmta — thank you for your interest!
If I find the correct way to fix it, I’ll definitely add the ad-hoc solution as you mentioned.

mikiubo: @gmta
 I think I find a solution

Let me know if I have to change anything else.

Libraries/LibWeb/DOM/Event.cpp
gmta, line 1: Commit message typo: "fixe" instead of "fixes"

Libraries/LibWeb/DOM/Event.h
```diff
@@ -101,6 +101,10 @@ class WEB_API Event : public Bindings::PlatformObject {
     GC::Ptr<EventTarget> current_target() const { return m_current_target; }
     void set_current_target(EventTarget* current_target) { m_current_target = current_target; }
 
+    // AD-HOC: Needed to return a WindowProxy when the current target is a Window,
+    // ensuring that JavaScript sees the correct global object instead of the internal Window.
```
gmta: Please move this comment to the implementation (.cpp) so it's immediately clear what the behavior is. I've tried to find some spec links supporting this behavior that we could add, but didn't find anything useful unfortunately.

Libraries/LibWeb/HTML/WindowProxy.cpp
```diff
@@ -24,10 +25,15 @@ GC_DEFINE_ALLOCATOR(WindowProxy);
...
 WindowProxy::WindowProxy(JS::Realm& realm)
-    : JS::Object(realm, nullptr, MayInterfereWithIndexedPropertyAccess::Yes)
+    : DOM::EventTarget(realm, MayInterfereWithIndexedPropertyAccess::Yes)
 {
 }
 
+void WindowProxy::initialize(JS::Realm& realm)
+{
+    Base::initialize(realm);
+}
```
gmta: You can leave out `WindowProxy::initialize()` since you simply call the base method.

Libraries/LibWeb/DOM/Event.cpp
gmta: I think you introduced new typos in the commit message? "that contains" -> "that contain"

Also the wrapping seems off. Your message should look like this:

```
LibWeb: Make Event.currentTarget return WindowProxy instead of Window

Make WindowProxy implement the EventTarget interface. Add a new method
current_target_for_bindings() that returns a WindowProxy object instead
of directly exposing the Window object.

These changes fixes several WPT tests that contain `window ===
currentTarget`.
```

PR #6210, opened 2025-09-15
CI+Meta: Add option to enable baseline CPU target for CI

gmta:
We currently have the issue that our macOS builds are being hosted by both Apple M1 and M4 runners. By default, these target `-mcpu=apple-m1` and `-mcpu=apple-m3` respectively, causing their ccache artifacts to be incompatible and effectively thrashing each other's ccache cache on each run.

Add a new CMake option `ENABLE_CI_BASELINE_CPU` that toggles a baseline `-mcpu/-march` compile option for each of the supported build platforms. When disabled (the default), we now default to `-march=native` which on Linux could theoretically lead to better performance.

The new CPU baseline is used by the `lagom-template.yml` and `js-and-wasm-artifacts.yml` workflows, since they both produce artifacts or caches that might be picked up by other runners.


.github/workflows/lagom-template.yml
```diff
@@ -140,7 +140,8 @@ jobs:
           VCPKG_CACHE_SAS: ${{ github.ref == 'refs/heads/master' && secrets.VCPKG_CACHE_SAS || '' }}
           VCPKG_CACHE_MODE: ${{ github.ref == 'refs/heads/master' && 'write' || '' }}
         run: |
-          cmake --preset ${{ inputs.build_preset }} -B Build
+          cmake --preset ${{ inputs.build_preset }} -B Build `
+            -DENABLE_CI_BASELINE_CPU=ON
```
trflynn89, line 144: Looks like there's an extraneous backtick after `Build ` - was that supposed to be be `\`?
(Unless this is some fun alternate yaml syntax)

Meta/CMake/common_compile_options.cmake
```diff
@@ -49,6 +49,23 @@ macro(add_swift_link_options)
...
+    if (APPLE AND CMAKE_SYSTEM_PROCESSOR STREQUAL "arm64")
+        add_cxx_compile_options(-mcpu=apple-m1)
+    elseif (CMAKE_SYSTEM_PROCESSOR MATCHES "^(aarch64|arm64)$")
+        add_cxx_compile_options(-march=armv8.2-a)
+    elseif (CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64")
+        add_cxx_compile_options(-march=x86-64-v3)
+    endif()
+else()
+    # In all other cases, compile for the native architecture of the host system.
+    add_cxx_compile_options(-march=native)
```
trflynn89, line 66: Will this impact the flatpak image we produce? Not sure what the native arch is on that runner vs. what users will be likely to have.

.github/workflows/lagom-template.yml
gmta: This is passed verbatim to Powershell, which uses backticks as its escape character.

Meta/CMake/common_compile_options.cmake
gmta: I think it will, good point! We could simply pass `ENABLE_CI_BASELINE_CPU` in our flatpak's `config-opts` for Ladybird as well, not sure if that would be a good default to have? For Linux that would mean targeting x86_64-v3.

.github/workflows/lagom-template.yml
trflynn89: Aha I missed that this is the Windows section - does setting this option here actually do anything then? Don't we hit the `if (MSVC)` branch before checking this option?

Meta/CMake/common_compile_options.cmake
trflynn89: Yeah honestly I don't know what the right option here is :upside_down_face: Maybe we would then want to produce separate arm and x64 binaries? (meaning eventually, not now). Assuming x64 for now is probably fine for our limited audience, it does feel a bit off though

.github/workflows/lagom-template.yml
gmta: It doesn't do anything - I've added it to show intent, even though the CMake implementation effectively ignores it because MSVC doesn't support it.

Want me to remove it?

Meta/CMake/common_compile_options.cmake
gmta: Flatpak already builds x86_64 and aarch64 artifacts; so it would mean x86_64-v3 and armv8.2-a respectively.

.github/workflows/lagom-template.yml
trflynn89: Gotcha, that's fine then. I find it a tiny bit confusing that we set an option that isn't used, but don't feel strongly about it at all

Meta/CMake/common_compile_options.cmake
trflynn89: Ah neat! Yeah I'm not sure what's the right call here, native or x86-64-v3 / armv8.2-a. The consistency side of my brain leans towards the latter just so our flatpak is doing the same thing as our `js` release.

gmta: Added it to the flatpak config!

trflynn89 (approved):

PR #6213, opened 2025-09-16
LibWeb: Apply composite operator to keyframe effects

tcl3:
This PR applies composite operation given by `ComputedKeyframe.composite` or `KeyframeEffect.composite` to the given effect. I've added code to handle composition of the most common style types, although some of the more complex cases like `TransformStyleValue` are not currently handled.

This gains us 603 WPT subtests.

While these tests pass, it's worth bearing in mind that they only use animations with 2 keyframes. I think ideally we should be following the spec for calculating [the effect value of a keyframe effect](https://drafts.csswg.org/web-animations-1/#the-effect-value-of-a-keyframe-animation-effect), which would ensure we produce the correct result for more complex animations.


Libraries/LibWeb/Animations/KeyframeEffect.cpp
```diff
@@ -877,6 +877,7 @@ WebIDL::ExceptionOr<void> KeyframeEffect::set_keyframes(Optional<GC::Root<JS::Ob
 
         for (auto [property_id, property_value] : keyframe.parsed_properties()) {
             resolved_keyframe.properties.set(property_id, property_value);
+            resolved_keyframe.composite = keyframe.composite;
```
AtkinsSJ: This doesn't need to be inside the loop does it?

Libraries/LibWeb/CSS/Interpolation.cpp
```diff
@@ -1559,4 +1559,120 @@ RefPtr<StyleValue const> interpolate_value(DOM::Element& element, CalculationCon
...
+        auto& animated_list = a_animated_value.as_value_list();
+        if (underlying_list.size() != animated_list.size() || underlying_list.separator() != animated_list.separator())
+            return {};
+        StyleValueVector values;
+        values.ensure_capacity(underlying_list.size());
+        for (size_t i = 0; i < underlying_list.size(); ++i) {
+            auto composited_value = composite_value(underlying_list.values()[i], animated_list.values()[i], composite_operation);
+            if (!composited_value)
+                return {};
+            values.append(*composited_value);
```
AtkinsSJ: You can `append_unchecked()`

Libraries/LibWeb/CSS/Interpolation.cpp
```diff
@@ -1559,4 +1559,120 @@ RefPtr<StyleValue const> interpolate_value(DOM::Element& element, CalculationCon
     return interpolate_discrete(from, to, delta, allow_discrete);
 }
 
+template<typename T>
+static T composite_raw_values(T underlying_raw_value, T animated_raw_value)
+{
+    return underlying_raw_value + animated_raw_value;
+}
+
+RefPtr<StyleValue const> composite_value(StyleValue const& a_underlying_value, StyleValue const& a_animated_value, Bindings::CompositeOperation composite_operation)
```
AtkinsSJ: Nit: I'm not a fan of the `a_` prefix of these, and I don't think `underlying_value` and `animated_value` would collide with anything.

AtkinsSJ (changes requested):
Just nits

Libraries/LibWeb/CSS/Interpolation.cpp
tcl3: I've renamed the variables that were causing collisions.

AtkinsSJ (approved):

Issue #6129, opened 2025-09-08
css: can't apply color in :hover pseudo selector


warpdesign:
### Summary

color css property doesn't work in `:hover` pseudo class.

### Operating system

macOS

### Steps to reproduce

1. Open the reduced test case
2. Move the mouse over the button

### Expected behavior

The text should turn red.

### Actual behavior

The text stays black.

### URL for a reduced test case

N/A

### HTML/SVG/etc. source for a reduced test case

```html
<html>
<body>
<style type="text/css">
button:hover{
  color:red
}
</style>
<body>
<button>Hover Me!</button>
```

### Log output and (if possible) backtrace

```shell
N/A
```

### Screenshots or screen recordings

Chrome:
<img width="180" height="117" alt="Image" src="https://github.com/user-attachments/assets/c46f9e15-e180-4cff-bf94-c2f6f4e28b61" />

Ladybird:
<img width="153" height="109" alt="Image" src="https://github.com/user-attachments/assets/7b280358-076d-4d87-ac67-4b2dd6b224ab" />

### Contribute a patch?

- [ ] I’ll contribute a patch for this myself.

shouvik-maity:
I think you have forgotten to give semicolon(;) after red.So I guess that's the issue.

AtkinsSJ:
> I think you have forgotten to give semicolon(;) after red.So I guess that's the issue.

That doesn't affect the outcome. CSS doesn't require a semicolon if there isn't another property after it.

tcl3:
Fixed by 010b0b00ff9

PR #6214, opened 2025-09-16
LibWeb: Propagate style to nested anonymous wrappers

pranair:
If we construct button under the condition specified in TreeBuilder::wrap_in_button_layout_tree_if_needed, we end up creating <button> with nested anonymous wrappers and style changes were not being propagated recursively.

This should resolve [https://github.com/LadybirdBrowser/ladybird/issues/6129](https://github.com/LadybirdBrowser/ladybird/issues/6129).


gmta:
Could you please add a regression test that shows the issue has been fixed?

pranair:
Hello @gmta,

 I have not added any tests for the time being. 

- Since we're dealing with anonymous wrappers, I've ruled out `Text` tests as we can't inspect the anonymous wrapper from JS (as far as I could tell)
- `Layout` tests don't seem to print out any information on `color` property, so it's not useful
- `Ref` tests and `Screenshot` don't seem to work either. For some reason, `--layout-test-mode` needs to be removed while taking screenhots for the bug to be replicated. But since the tests run with `--layout-test-mode`, it's a no-go.

Please let me know if I'm missing something!

gmta:
@pranair A screenshot test should definitely work, but you can also add a display list text which is a `Text` test that runs `internals.dumpDisplayList()`. That includes the color of the text. Check out `Tests/LibWeb/Text/input/display_list/button-with-text-decoration.html` for an example on how to do that.

pranair:
@gmta, Thanks for the suggestion, the example mentioned seem to work perfectly in this case.

PR #6216, opened 2025-09-16
LibWeb: Implement css decode bytes algo

lpas:


Libraries/LibWeb/CSS/Parser/Helpers.cpp
```diff
@@ -138,4 +140,76 @@ Vector<CSS::Parser::ComponentValue> parse_component_values_list(CSS::Parser::Par
...
+{
+    // https://drafts.csswg.org/css-syntax/#determine-the-fallback-encoding
+    auto determine_the_fallback_encoding = [&mime_type_charset, &environment_encoding, &encoded_string]() -> StringView {
+        // 1. If HTTP or equivalent protocol provides an encoding label (e.g. via the charset parameter of the Content-Type header) for the stylesheet,
+        //    get an encoding from encoding label. If that does not return failure, return it.
+        if (mime_type_charset.has_value()) {
+            if (auto encoding = TextCodec::get_standardized_encoding(mime_type_charset.value()); encoding.has_value())
+                return encoding.value();
+        }
+        // 2. Otherwise, check stylesheet’s byte stream. If the first 1024 bytes of the stream begin with the hex sequence
```
AtkinsSJ, line 154: This apostrophe has ended up encoded wrong (which is ironic considering this is an encoding algorithm!). Please correct that, and also wrap the spec comments so no lines are longer than 120 columns. ( https://github.com/LadybirdBrowser/ladybird/blob/master/Documentation/CodingStyle.md#comments )

Libraries/LibWeb/CSS/Parser/Helpers.cpp
```diff
@@ -138,4 +140,76 @@ Vector<CSS::Parser::ComponentValue> parse_component_values_list(CSS::Parser::Par
...
+            size_t scan_length = min(encoded_string.size(), 1024);
+            auto pattern_start = "@charset \""sv;
+            auto pattern_end = "\";"sv;
+
+            if (scan_length < pattern_start.length())
+                return {};
+
+            StringView start_view(encoded_string.data(), pattern_start.length());
+            if (start_view != pattern_start)
+                return {};
```
AtkinsSJ, line 168: If you create `buffer_view` before this, then you can use `buffer_view.starts_with(pattern_start)`.

Libraries/LibWeb/CSS/Parser/Helpers.cpp
```diff
@@ -138,4 +140,76 @@ Vector<CSS::Parser::ComponentValue> parse_component_values_list(CSS::Parser::Par
...
+
+            auto encoding_start = pattern_start.length();
+            StringView buffer_view(encoded_string.data(), scan_length);
```
AtkinsSJ: This feels a bit dangerous, and we have more type-safe ways of doing it:
- `encoded_string.bytes()` returns a Bytes, which is a view onto the data (no copies)
- ...which you can get the first 1024 chars of with `.slice(start, length)`
- ...and then you can construct a `StringView` directly from that.

Libraries/LibWeb/CSS/Parser/Helpers.cpp
```diff
@@ -138,4 +140,76 @@ Vector<CSS::Parser::ComponentValue> parse_component_values_list(CSS::Parser::Par
...
+
+    // 1. Determine the fallback encoding of stylesheet, and let fallback be the result.
+    auto fallback = determine_the_fallback_encoding();
+    auto decoder = TextCodec::decoder_for(fallback);
+    if (!decoder.has_value()) {
+        // If we don't support the encoding yet, let's error out instead of trying to decode it as something it's most likely not.
+        dbgln("FIXME: Style sheet encoding '{}' is not supported yet", fallback);
+        return Error::from_string_literal("No Decoder found");
+    }
+    // 2. Decode stylesheet’s stream of bytes with fallback encoding fallback, and return the result.
```
AtkinsSJ, line 210: Another wrongly-encoded apostrophe.

Libraries/LibWeb/HTML/HTMLLinkElement.cpp
```diff
@@ -469,43 +472,33 @@ void HTMLLinkElement::process_stylesheet_resource(bool success, Fetch::Infrastru
...
-        if (!encoding.has_value())
-            encoding = document().encoding_or_default();
-
-        auto decoder = TextCodec::decoder_for(*encoding);
-
-        if (!decoder.has_value()) {
-            // If we don't support the encoding yet, let's error out instead of trying to decode it as something it's most likely not.
-            dbgln("FIXME: Style sheet encoding '{}' is not supported yet", encoding);
+            if (auto environment_encoding = TextCodec::get_standardized_encoding(charset.release_value()); environment_encoding.has_value())
+                environment_encoding = environment_encoding.value();
```
AtkinsSJ, line 478: Please use `{}` for blocks that span multiple lines.

AtkinsSJ (changes requested):

Libraries/LibWeb/CSS/Parser/Helpers.cpp
lpas: Hey, how would I see that it is encoded wrong? On my end this looks correctly utf-8 encoded in vscode and here on github. If I can't see what's wrong it is hard to fix :>

AtkinsSJ: Oh weirdly, it looks fine on GitHub [here](https://github.com/LadybirdBrowser/ladybird/pull/6216#discussion_r2359177600) but not when viewed [here](https://github.com/LadybirdBrowser/ladybird/pull/6216/files/38ae76ada47590eb57822cbbb0f7bb45b44e3b64#r2358053008)

Everything should use utf-8 encoding, so maybe you have your IDE or browser set to something else? I know very little about encodings sorry. :sweat_smile:

lpas: oh wow I'm not sure this is a me problem, 

this looks wrong here (the old review view)
Image
but fine here (the new review mode https://github.com/orgs/community/discussions/163932)
Image

AtkinsSJ: It does look fine in the IDEs I've tried, actually. So feel free to ignore this.

lpas: the file seemed to be ASCII before my edit maybe that is a problem for github :/ either way I would have no Idea how to fix this
```bash
git show @~2:Libraries/LibWeb/CSS/Parser/Helpers.cpp | file -
/dev/stdin: C++ source, ASCII text
```

```bash
git show :Libraries/LibWeb/CSS/Parser/Helpers.cpp | file -
/dev/stdin: C++ source, Unicode text, UTF-8 text
```

AtkinsSJ: ASCII text is also UTF-8. So yeah just don't worry about this at all. :sweat_smile:

lpas:
Hey thanks for the review, and the tips. I have addressed everything but the encoding not sure what this is never seen this before.

Libraries/LibWeb/CSS/Parser/Helpers.cpp
lpas: yeah but the apostrophe is not. what made it from ascii -> utf-8 I would assume. But fun never seen this,  every day something new

Libraries/LibWeb/CSS/CSSImportRule.cpp
gmta, line 1: Commit title: let's call it "CSS" instead of "css".

Libraries/LibWeb/CSS/Parser/Helpers.cpp
```diff
@@ -138,4 +140,76 @@ Vector<CSS::Parser::ComponentValue> parse_component_values_list(CSS::Parser::Par
...
+    auto maybe_decoded_string = TextCodec::convert_input_to_utf8_using_given_decoder_unless_there_is_a_byte_order_mark(*decoder, encoded_string);
+    return maybe_decoded_string;
```
gmta: No need to put this in a variable, just `return` it immediately.

Libraries/LibWeb/CSS/Parser/Helpers.cpp
```diff
@@ -138,4 +140,76 @@ Vector<CSS::Parser::ComponentValue> parse_component_values_list(CSS::Parser::Par
...
+            if (scan_length < pattern_start.length())
+                return {};
+
+            StringView buffer_view = encoded_string.bytes().slice(0, scan_length);
+            if (!buffer_view.starts_with(pattern_start))
+                return {};
+
+            auto encoding_start = pattern_start.length();
+            auto end_index = buffer_view.find(pattern_end, encoding_start);
+            if (!end_index.has_value())
```
gmta, line 172: Maybe check if end_index is at least larger than encoding_start here as well, so we don't end up underflowing encoding_length?

Libraries/LibWeb/CSS/Parser/Helpers.cpp
```diff
@@ -138,4 +140,76 @@ Vector<CSS::Parser::ComponentValue> parse_component_values_list(CSS::Parser::Par
...
+                if ((byte < 0x01 || byte > 0x21) && (byte < 0x23 || byte > 0x7F)) {
+                    return {};
+                }
+            }
+
+            return encoding_view;
+        };
+        // If the return value was utf-16be or utf-16le, return utf-8; if it was anything else except failure, return it.
+        auto byte_stream_value = check_stylesheets_byte_stream();
+        if (byte_stream_value.has_value() && (byte_stream_value->equals_ignoring_ascii_case("utf-16be"sv) || byte_stream_value->equals_ignoring_ascii_case("utf-16le"sv)))
```
gmta: Why do we do case insensitive matches here?

Libraries/LibWeb/CSS/Parser/Helpers.cpp
lpas: Doesn't take the `find` take care of that for me? I would have assumed that the second parameter in `find` is the place where we start searching and thus if we find something it must be greater that `encoding_start`?

Libraries/LibWeb/CSS/Parser/Helpers.cpp
lpas: You are totally right the spec text doesn't spell this out.

Not sure what to do here. In general charsets are ASCII case-insensitive (https://infra.spec.whatwg.org/#ascii-case-insensitive) and I would assume that this is the case here as well. But the spec also doesn't spell this out here.

Libraries/LibWeb/CSS/Parser/Helpers.cpp
gmta: Ah that's silly - you're right!

Libraries/LibWeb/CSS/Parser/Helpers.cpp
gmta: If other browsers perform case-insensitive matching as well, it only seems right to copy their behavior.

[This is the spec](<https://encoding.spec.whatwg.org/#concept-encoding-get>) that standardizes getting the effective encoding, it also performs whitespace trimming on top of what you're already doing. We have an implementation of this in `TextCodec::get_standardized_encoding()`.

gmta: I just now noticed that you're already using `TextCodec::get_standardized_encoding()`, but you forgot to use it in the subsequent spec step.

lpas: You are totally right, not sure why I didn't do this in the first place as this is what the spec says :/ sorry for that

gmta (approved):

PR #5972, opened 2025-08-24
Services/WebDriver: Enable on Windows

ayeteadoe:
This is the final relatively trivial target to get building on Windows. I believe all the remaining after this is:
1. LibCore/FileWatcher
2. LibCore/TimeZoneWatcher
3. LibCore/ProcessStatistics
4. test262-runner
5. LibLine
6. LibWasm/WASI

For 1-2 I have some in progress impls of them that I may look to complete soon-ish as well as 3 given it should be pretty straight forward.

For 4-6 see recent discussion in `#ui-windows` about them, so I'm not planning on looking into any of them myself for now.

PR #6218, opened 2025-09-17
LibCore+LibUnicode+LibWebView+UI/Qt: Fix stale timezone retrieval and support TimeZoneWatcher on Windows

ayeteadoe:
See comments in https://github.com/LadybirdBrowser/ladybird/pull/5972 regarding a few of the TimeZoneWatcher refactoring commits.

This work was initially only focused on adding Windows support for TImeZoneWatcher; however, when I went to actually test it, I found a bug that seemed to also be an issue on other platforms due to how we were retrieving the timezone string from `icu::TimeZone` when the cache was stale.

### Demo

<details>

<summary>Page used in the following demos</summary>

```html
<!DOCTYPE html>
<html>
<head>
    <title>TimeZoneWatcher</title>
</head>
<body>
<div id="output">
    <h2>Current Date:</h2>
    <div id="current-date"></div>
</div>
<script>
    const currentDateText = document.getElementById("current-date");
    currentDateText.innerHTML += `<p style="color: blue;">${new Date()}</p>`;
</script>
</body>
</html>
```

</details>

#### Before `icu::TimeZone` fix

https://github.com/user-attachments/assets/a824c2d1-a8ab-4317-b486-dbb842d176d1

#### After `icu::TimeZone` fix

We now match the behaviour of Firefox and Chromium when refreshing the page.

https://github.com/user-attachments/assets/451f8190-140d-44e7-a989-4daeb118b7ae

#### Windows support

https://github.com/user-attachments/assets/0e41b0c9-947e-4d2c-b0ed-75cfcb762e97


Libraries/LibCore/CMakeLists.txt
gmta, line 1: Commit message typo: Winodws

Also, it looks like you're manually wrapping your commit messages. Please set up your editor to auto-wrap at 72 characters.

Libraries/LibCore/Forward.h
```diff
@@ -44,7 +44,6 @@ class TCPServer;
 class TCPSocket;
 class Timer;
 class TimerEvent;
-class TimeZoneWatcher;
```
gmta, line 47: TimeZoneWatcher feels like something that belongs in LibCore though. The fact that one of its implementations require a WebView context feels like we're doing something in the wrong order or are depending on the wrong things?

I'd rather not have the requirement of setting up a web view just to watch for platform timezone changes.

Couldn't we expose a public API for the Windows TimeZoneWatcher implementation that allows the GUI part to inform it of transpired events such as WM_TIMECHANGE?

Libraries/LibWebView/CMakeLists.txt
```diff
@@ -104,6 +106,8 @@ if (ENABLE_QT)
     if (WIN32)
         find_package(pthread REQUIRED)
         target_include_directories(LibWebViewQt PRIVATE $<BUILD_INTERFACE:${PTHREAD_INCLUDE_DIR}>)
+        # NOTE: Required for the Qt-only Windows TimeZoneWatcher implementation
+        target_link_libraries(LibWebView PUBLIC Qt::Core)
```
trflynn89: We 100% want to be removing Qt dependencies from LibWebView, not adding them.

Can't we do something like:
1. Have `EventLoopImplementationWindows` listen for `WM_*` messages of interest
2. Expose an API that allows interested parties to register observers for these messages
3. `TimeZoneWatcherWindows` registers for `WM_TIMECHANGE` and handles it accordingly

Libraries/LibWebView/TimeZoneWatcher/TimeZoneWatcherWindowsQt.cpp
```diff
@@ -0,0 +1,72 @@
...
+    QAbstractNativeEventFilter& qt_native_event_filter() override
+    {
+        return m_time_change_event_filter;
+    }
+
+protected:
+    void set_on_time_zone_changed(Function<void()> on_time_zone_changed) override
+    {
+        m_time_change_event_filter.on_time_zone_changed = move(on_time_zone_changed);
+    }
```
trflynn89: I don't think we need to make this virtual. Could just do something like:
```c++
m_time_change_event_filter.on_time_zone_changed = [&]() {
	if (on_time_zone_changed)
        on_time_zone_changed();
};
```

Or use multiple inheritance

Libraries/LibWebView/CMakeLists.txt
R-Goc: Well our event loop in LibCore is not a message loop at all. All of that is handled by qt.

Libraries/LibCore/Forward.h
R-Goc: > Couldn't we expose a public API for the Windows TimeZoneWatcher implementation that allows the GUI part to inform it of transpired events such as WM_TIMECHANGE?

This seems like the most reasonable approach.

Libraries/LibWebView/CMakeLists.txt
ayeteadoe: `WM_TIMECHANGE` events are only sent to top level windows (`HWND`s), Which means the message pump setup by the QGuiApplication is where they are received.

> Expose an API that allows interested parties to register observers for these messages

That is done via creating a custom [WndProc](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nc-winuser-wndproc), which requires registering the top level `HWND`s to have a callback that actually emits `WM_TIMECHANGE`. We can get top level windows from `QGuiApplication::topLevelWindows()`, but now we need to handle when new windows are created and/or existing registered windows disappear. The main window rendering also broke in my initial naiive attempt to go this route, so I stuck with a `QAbstractNativeEventFilter` approach while keeping everything in `LibCore` like suggested by gmta above.

Libraries/LibUnicode/TimeZone.cpp
```diff
@@ -38,7 +34,25 @@ String current_time_zone()
...
+    return current_time_zone_impl(adopt_own_if_nonnull(icu::TimeZone::createDefault()));
+}
+
+String current_time_zone()
+{
+    if (cached_system_time_zone.has_value())
+        return *cached_system_time_zone;
+
+    cached_system_time_zone = current_host_time_zone();
     return *cached_system_time_zone;
```
gmta:
```cpp
return cached_system_time_zone.ensure([] { return current_host_time_zone(); });
```

gmta (approved):

PR #6219, opened 2025-09-17
LibWeb: Add a simplified Notification constructor

NiccoloAntonelliDziri:
This allows the Notification object to be created in javascript without any additional functionalities.

It passes two wpt tests which require a call to the notification constructor with no arguments.

https://wpt.live/notifications/constructor-basic.https.html

https://wpt.live/notifications/constructor-invalid.https.html


Libraries/LibWeb/NotificationsAPI/Notification.idl
```diff
@@ -0,0 +1,76 @@
+#import <HighResolutionTime/EpochTimeStamp.idl>
+#import <DOM/EventTarget.idl>
+
+// https://notifications.spec.whatwg.org/#api
+[Exposed=(Window,Worker)]
+interface Notification : EventTarget {
+    constructor(DOMString title, optional NotificationOptions options = {});
+
+//  static readonly attribute NotificationPermission permission;
+//  [Exposed=Window] static Promise<NotificationPermission> requestPermission(optional NotificationPermissionCallback deprecatedCallback);
```
AtkinsSJ: Instead of commenting these out, you can usually add a `FIXME` annotation and we'll generate a stub. eg:

```suggestion
    [FIXME] static readonly attribute NotificationPermission permission;
    [FIXME, Exposed=Window] static Promise<NotificationPermission> requestPermission(optional NotificationPermissionCallback deprecatedCallback);
```

It doesn't work for everything, but it does a lot of the time. For any that won't compile that way, add a `// FIXME` comment so we can find them easily.

Libraries/LibWeb/NotificationsAPI/Notification.idl
```diff
@@ -0,0 +1,76 @@
+#import <HighResolutionTime/EpochTimeStamp.idl>
+#import <DOM/EventTarget.idl>
+
+// https://notifications.spec.whatwg.org/#api
```
AtkinsSJ: In the spec, `interface Foo` will be a link, which gives you a more direct URL. Here, that's `https://notifications.spec.whatwg.org/#notification`

Libraries/LibWeb/NotificationsAPI/Notification.idl
```diff
@@ -0,0 +1,76 @@
+#import <HighResolutionTime/EpochTimeStamp.idl>
+#import <DOM/EventTarget.idl>
```
AtkinsSJ: Alphabetical order please. :^)

Libraries/LibWeb/NotificationsAPI/Notification.cpp
```diff
@@ -0,0 +1,46 @@
...
+    JS::Realm& realm,
+    String title,
+    Optional<NotificationOptions> options)
+{
+
+    // FIXME: all of the steps specified in the spec
+
+    // This is temporary to avoid the error: unused parameter
+    dbgln(title);
+    dbgln(options->body);
```
AtkinsSJ: You can do this more simply with `(void)thing_that_is_unused`. That's how we do it in most places.

Libraries/LibWeb/NotificationsAPI/Notification.cpp
```diff
@@ -0,0 +1,46 @@
...
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include <LibJS/Runtime/Realm.h>
+#include <LibWeb/Bindings/Intrinsics.h>
+#include <LibWeb/Bindings/NotificationPrototype.h>
+#include <LibWeb/NotificationsAPI/Notification.h>
+
+// FIXME: Remove this include when removing `dbgln` from `construct_impl`
+#include <AK/Format.h>
```
AtkinsSJ: ...and then you don't need this.

Libraries/LibWeb/NotificationsAPI/Notification.idl
```diff
@@ -0,0 +1,76 @@
...
+dictionary NotificationOptions {
+    NotificationDirection dir = "auto";
+    DOMString lang = "";
+    DOMString body = "";
+    USVString navigate;
+    DOMString tag = "";
+    USVString image;
+    USVString icon;
+    USVString badge;
+// VibratePattern vibrate;
```
AtkinsSJ: This is an example of where `[FIXME]` won't work, because `VibratePatten` isn't defined anywhere yet.

Libraries/LibWeb/NotificationsAPI/Notification.cpp
```diff
@@ -0,0 +1,46 @@
+/*
+ * Copyright (c) 2025,  Niccolo Antonelli-Dziri <niccolo.antonelli-dziri@protonmail.com>
```
AtkinsSJ: Nit: You have an extra space before your name in both files' copyright.

Libraries/LibWeb/NotificationsAPI/Notification.h
```diff
@@ -0,0 +1,73 @@
...
+
+constexpr std::nullptr_t null = nullptr;
+
+}
+
+using NotificationsAPI::Notification;
+using NotificationsAPI::NotificationAction;
+using NotificationsAPI::NotificationOptions;
+
+}
```
AtkinsSJ: I was very confused why you needed this... and you actually don't, but you need to add a fix elsewhere! This function sticks a `using namespace` into the generated code for every `Web::Foo` namespace:

https://github.com/LadybirdBrowser/ladybird/blob/a54215c07d161a98bc8e07d68e2e32803eb3f5da/Meta/Lagom/Tools/CodeGenerators/LibWeb/BindingsGenerator/IDLGenerators.cpp#L4909

You added a new namespace, so you also need to add it to the list here (and correct it in `Forward.h`). Then the generated code will be able to see your types and you can remove this code.

It's a hack because the code generator isn't smart enough to put namespaces before type names. But it works so nobody's fixed it. :^)

Also, needs the fix in #6225 so we stop trying to assign `null` to `NotificationOptions::silent`.

Libraries/LibWeb/Forward.h
```diff
@@ -910,6 +910,15 @@ class PerformanceTiming;
 
 }
 
+namespace NotificationsAPI {
```
AtkinsSJ: This also needs to be `Web::NotificationsAPI`.

AtkinsSJ (changes requested):
Generally looks good! Most of the comments are pretty small changes.

You managed to run into a bindings generator bug which is always fun, but I've PRed a fix for that separately: #6225.

Tests/LibWeb/Text/expected/all-window-properties.txt
```diff
@@ -304,6 +304,7 @@ Node
 NodeFilter
 NodeIterator
 NodeList
+Notification
```
AtkinsSJ, line 307: Oh I forgot to say as well, this change should just be part of the previous commit. That way each commit is "atomic", and tests always pass.

AtkinsSJ:
A couple of quick things:
1. Click the `[Resolve conversation]` button on feedback that you've made changes for, so reviewers can see that it's dealt with.
2. Please amend changes into your existing commits instead of creating a new "fixes" commit. We don't squash PRs, we just rebase the commits directly into the history, so we like those commits to be neat and atomic.

AtkinsSJ (approved):
I think this is fine as a starting point. Thanks!

PR #6220, opened 2025-09-17
AK: Add Optional<T>::ensure() 

gmta:
We often use Optional<T> in a cache pattern such as:

```cpp
  if (!m_cache.has_value())
      m_cache = slow_thing();
  return m_cache.value();
```

The new ::ensure() makes it a bit simpler:

```cpp
  return m_cache.ensure([&] { return slow_thing(); });
```

### DIFF
--- a/AK/Optional.h
+++ b/AK/Optional.h
@@ -1,6 +1,7 @@
 /*
  * Copyright (c) 2018-2021, Andreas Kling <andreas@ladybird.org>
  * Copyright (c) 2021, Daniel Bertalan <dani@danielbertalan.dev>
+ * Copyright (c) 2025, Jelle Raaijmakers <jelle@ladybird.org>
  *
  * SPDX-License-Identifier: BSD-2-Clause
  */
@@ -121,6 +122,14 @@ public:
         if (static_cast<Self const&>(*this).has_value())
             return static_cast<Self const&>(*this).value();
         return TRY(callback());
+    }
+
+    template<typename Callable>
+    [[nodiscard]] ALWAYS_INLINE constexpr T& ensure(Callable callable) &
+    {
+        if (!static_cast<Self&>(*this).has_value())
+            static_cast<Self&>(*this) = callable();
+        return static_cast<Self&>(*this).value();
     }
 
     [[nodiscard]] ALWAYS_INLINE constexpr T const& operator*() const { return static_cast<Self const&>(*this).value(); }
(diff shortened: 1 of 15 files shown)


Libraries/LibWeb/CSS/CSSRule.h
```diff
@@ -70,9 +70,7 @@ class WEB_API CSSRule : public Bindings::PlatformObject {
 
     [[nodiscard]] FlyString const& parent_layer_internal_qualified_name() const
     {
-        if (!m_cached_layer_name.has_value())
-            return parent_layer_internal_qualified_name_slow_case();
-        return m_cached_layer_name.value();
+        return m_cached_layer_name.ensure([&] { return parent_layer_internal_qualified_name_slow_case(); });
```
trflynn89, line 73: Hmm this one's a little weird because `parent_layer_internal_qualified_name_slow_case` internally sets `m_cached_layer_name`. So we should probably remove that and have it return the generated `FlyString` directly.

AK/Optional.h
```diff
@@ -604,6 +613,14 @@ requires(IsLvalueReference<T>) class [[nodiscard]] Optional<T> {
         return TRY(callback());
     }
 
+    template<typename Callback>
+    [[nodiscard]] ALWAYS_INLINE constexpr T ensure(Callback callback) &
+    {
+        if (m_pointer == nullptr)
+            m_pointer = callback();
```
trflynn89: I don't think I saw any `Optional<T&>::ensure` users in the next commit, but I'm wondering how this will semantically work. It looks like the callback would be expected to return a pointer - maybe we want to expect it to return a reference? Just to match other APIs here. So then this would be:
```c++
m_pointer = &callback();
```

Libraries/LibWeb/CSS/CSSRule.h
gmta: Good catch!

AK/Optional.h
gmta: Changed!

AK/Optional.h
```diff
@@ -123,6 +124,14 @@ requires(!IsLvalueReference<Self>) class [[nodiscard]] OptionalBase {
         return TRY(callback());
     }
 
+    template<typename Callable>
+    [[nodiscard]] ALWAYS_INLINE constexpr T& ensure(Callable callable) &
```
rgret-dev, line 128: Maybe `value_or_else()` would be a better name to indicate it's a computed fallback?

gmta: This reflects the existing `HashMap::ensure()` API.

rgret-dev: Hmm okay, I just think it's more intuitive since there is an existing `value_or()` API on Optional for fallback values.

gmta: The `value_or_*()` APIs do not mutate the optional however, so I think `ensure` is still the way to go.

PR #6221, opened 2025-09-17
LibWebView+UI: A couple menu fixes and improvements, and generate the Inspect menu

trflynn89:
* A couple minor fixes and simplifications after #6062
* Add a couple more menu icons for macOS Tahoe
* Generate the actions to open `about:` pages
* Generate the remaining actions for the Inspect menu and then the menu itself

### DIFF
--- a/Libraries/LibWebView/Application.cpp
+++ b/Libraries/LibWebView/Application.cpp
@@ -278,6 +278,12 @@ ErrorOr<void> Application::initialize(Main::Arguments const& arguments)
     return {};
 }
 
+void Application::open_url_in_new_tab(URL::URL const& url, Web::HTML::ActivateTab activate_tab) const
+{
+    if (auto view = open_blank_new_tab(activate_tab); view.has_value())
+        view->load(url);
+}
+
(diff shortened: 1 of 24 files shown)


Libraries/LibWebView/Application.cpp
```diff
@@ -818,15 +845,26 @@ void Application::apply_view_options(Badge<ViewImplementation>, ViewImplementati
...
+    } else {
+        TRY(launch_devtools_server());
     }
 
-    TRY(launch_devtools_server());
-    return DevtoolsState::Enabled;
+    return {};
+}
+
+void Application::devtools_enabled() const
```
konradekk: not the best name according to a common rule, where each function name contains a verb in active voice describing body of that function (unless it's an accessor/mutator). there's a different convention for listeners, often having shape of `on_{event_name}`. maybe something like `update_*`, `notify_*`, or `on_*`?

AtkinsSJ: It looks like these names existed before, but I still much prefer `on_devtools_enabled/disabled()`.

PR #6222, opened 2025-09-17
Add missing X_API macros to classes needing them for DEBUG macro builds

rcorsi:
When building with ENABLE_ALL_THE_DEBUG_MACROS there are a number of issues encountered.  See #6173 for the details.

This PR fixes the remaining linking issues (undefined references) by adding X_API macros (DNS_API and REGEX_API) to the specific classes that encountered the issue.

1- DNS_DEBUG issue with `Libraries/LibDNS/Resolver.h:1003` with undefined reference `DNS::Messages::Records::SIG::to_string() const`

which impacts

 - bin/dns
 - bin/TestDNSResolver
 - libexec/RequestServer

2- REGEX_DEBUG issue with `Libraries/LibRegex/RegexDebug.h:76` with undefined reference `regex::OpCode_Compare::variable_arguments_to_byte_string(AK::Optional<regex::MatchInput const&>) const`

which impacts:

 - bin/TestRegex

3- REGEX_DEBUG issue with `Libraries/LibRegex/RegexByteCode.h:672` with undefined reference `regex::OpCode::name(regex::OpCodeId)`

which impacts:

 - bin/TestRegex

fixes: #6173

### DIFF
--- a/Libraries/LibDNS/Message.h
+++ b/Libraries/LibDNS/Message.h
@@ -537,7 +537,7 @@ struct CDS : public DS {
     static constexpr ResourceType type = ResourceType::CDS;
     static ErrorOr<CDS> from_raw(ParseContext& raw) { return DS::from_raw(raw); }
 };
-struct SIG {
+struct DNS_API SIG {
     ResourceType type_covered;
     DNSSEC::Algorithm algorithm;
     u8 label_count;
--- a/Libraries/LibRegex/RegexByteCode.h
+++ b/Libraries/LibRegex/RegexByteCode.h
@@ -640,7 +640,7 @@ StringView boundary_check_type_name(BoundaryCheckType);
 StringView character_compare_type_name(CharacterCompareType result);
 StringView character_class_name(CharClass ch_class);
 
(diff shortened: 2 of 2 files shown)


rcorsi:
@alimpfard Thanks for the merge.

About the X_API macros, I was thinking maybe there should also be a X_DEBUG_API version, so that symbols are clearly marked for export when they are only for debug purposes.  Maybe that is overkill.

I was also wondering if I should have some `#ifdef DNS_DEBUG` (or somehow use constexpr) around the class to make it dependant on if the debug macro is defined.  Again maybe too messy, or overkill.

alimpfard:
My problem is not annotating these, but rather the fact that the annotation being necessary is purely dependent on some random TU's usage of it.
It's entirely self-inflicted pain (_everything_ in a public header is public API, otherwise it would go in a private header or inside a cpp file), I understand the technical reason but no likely.

Either way this is fine, maybe a good extra would be to figure out a way to error when an X_API class is actually entirely unused even with all debug macros on.

PR #6223, opened 2025-09-17
LibThreading: Forward-declare Thread and remove its Weakable functionality

Zaggy1024:
`Threading::Thread` is ref-counted, so using forward declarations in a lot of headers currently including Thread.h is an easy change.

Also, `Weakable` is not thread-safe, so taking a strong reference from a `WeakPtr<Thread>` may result in a use-after-free. We don't use this functionality anywhere anyway, so remove it.

No functional changes, hopefully.

### DIFF
--- a/Libraries/LibThreading/Forward.h
+++ b/Libraries/LibThreading/Forward.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023, Gregory Bertilson <gregory@ladybird.org>
+ * Copyright (c) 2023-2025, Gregory Bertilson <gregory@ladybird.org>
  *
  * SPDX-License-Identifier: BSD-2-Clause
  */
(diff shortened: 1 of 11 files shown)


Libraries/LibThreading/Forward.h
```diff
@@ -1,6 +1,5 @@
 /*
- * Copyright (c) 2019-2020, Sergey Bugaev <bugaevc@serenityos.org>
- * Copyright (c) 2021, Spencer Dixon <spencercdixon@gmail.com>
+ * Copyright (c) 2023-2025, Gregory Bertilson <gregory@ladybird.org>
```
konradekk, line 2: 🤔

Zaggy1024: Did you check the commit message? :^)

Zaggy1024: Just to make it absolutely clear the reason for the change, I've split it into a separate commit.

PR #6225, opened 2025-09-18
IDLGenerators: Support `null` as a default value for dictionary types

AtkinsSJ:
Specifically, NotificationOptions has this:

```webidl
dictionary NotificationOptions {
    // ...
    boolean? silent = null;
    // ...
};
```
https://notifications.spec.whatwg.org/#dictdef-notificationoptions

Without this patch, we generate this code, which isn't valid:

```c++
silent_value_11 = static_cast<bool>(null);
```

Treating `null` the same as no default value seems like the best option, as they're equivalent in our C++ types.


AtkinsSJ:
Updated to only affect `generate_to_integral()`, as I was failing tests by applying this to all `= null` defaults.

PR #6227, opened 2025-09-18
LibWeb/CSS: Implement CSSTransformValue and friends

AtkinsSJ:
Implement the different typed-OM versions of transform functions, and then `CSSTransformValue` which contains a list of them. There's a lot of boilerplate here.

I'm not super confident working with matrices, so a lot of the reification code uses `DOMMatrix` methods. This means there's an unnecessary multiplication with the identity matrix, but it seems a lot less error-prone to reuse existing code for that. But in simpler cases I've manually assigned the matrix values to avoid that.

As `CSSTransformValue` is the last `CSSStyleValue` type tested by the per-property harness, this means those finally run! However, they don't pass much because we don't yet implement setting styles with `StylePropertyMap`. But at least once we do, we'll actually see if they function properly!


AtkinsSJ:
Drafting this to prioritise fixing the IDL harness test. (#6230)

AtkinsSJ:
OK! So, changes are just that we are more aware of how many test passes we get for each commit. (It's a bigger number, which is the best kind of number.)

AtkinsSJ:
Changes: A couple of minor fixes (and a missing visit). Then a rebase. I'm calling it good. :^)

PR #6235, opened 2025-09-18
Add the `overflow-inline` and `overflow-block` properties

tcl3:
Gains us 17 WPT subtests in the `css/css-overflow` directory.

### DIFF
--- a/Meta/Lagom/Tools/CodeGenerators/LibWeb/GenerateCSSPropertyID.cpp
+++ b/Meta/Lagom/Tools/CodeGenerators/LibWeb/GenerateCSSPropertyID.cpp
@@ -1695,6 +1695,14 @@ PropertyID map_logical_alias_to_physical_property(PropertyID property_id, Logica
             return PropertyID::@bottom-left:titlecase@;
         return PropertyID::@top-left:titlecase@;
 )~~~");
+            } else if (mapping == "block-xy"sv) {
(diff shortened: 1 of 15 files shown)


Meta/Lagom/Tools/CodeGenerators/LibWeb/GenerateCSSPropertyID.cpp
```diff
@@ -1702,6 +1702,14 @@ PropertyID map_logical_alias_to_physical_property(PropertyID property_id, Logica
         if (mapping_context.writing_mode == WritingMode::HorizontalTb)
             return PropertyID::@y:titlecase@;
         return PropertyID::@x:titlecase@;
+)~~~");
+            } else if (mapping == "overflow-inline"sv) {
```
AtkinsSJ: If this could have a non-property-specific name, I'd prefer that. eg, `inline-xy` or something. Then it's clearer how other properties would use the same thing. (eg, I can see there are `overflow-behavior-x` and `overflow-behaviour-block` etc. which would use the same mapping.)

It's confusing because a few of these mapping names *are* property names, just because that happened to be the most descriptive to what the mapping is. :sweat_smile:

tcl3: That's a much better name! Updated.

AtkinsSJ (approved):

PR #6241, opened 2025-09-19
LibWeb/HTML: Correct color parsing for CanvasGradient color stops

ljamar:
This resolves all WPT timeouts in
wpt/html/canvas/offscreen/fill-and-stroke-styles


Tests/LibWeb/Text/input/canvas/add-color-stop-color-function-should-not-crash.html
```diff
@@ -0,0 +1,17 @@
...
+    const ctx = canvas.getContext("2d");
+    var g = ctx.createLinearGradient(0, 0, 100, 0);
+    g.addColorStop(0, "color(srgb 1 0 0)");
+    g.addColorStop(1, "color(srgb 0 1 0)");
+    ctx.fillStyle = g;
+    ctx.fillRect(0, 0, 100, 50);
+</script>
+<script>
+    test(() => {
+        println("PASS (didn't crash)");
```
gmta: You can put this in `Tests/LibWeb/Crash`, which is a separate category for crash tests. You can leave out this `println`.

gmta (commented):
Welcome to the project!

Tests/LibWeb/Crash/JS/canvas-gradient-add-color-stop-color-function.html
```diff
@@ -0,0 +1,9 @@
+<!DOCTYPE html>
+<canvas id="canvas" width="100" height="50"> </canvas>
+<script>
+    const canvas = document.getElementById("canvas");
+    const ctx = canvas.getContext("2d");
+    var g = ctx.createLinearGradient(0, 0, 100, 0);
+    g.addColorStop(0, "color(srgb 1 0 0)");
+    g.addColorStop(1, "color(srgb 0 1 0)");
+</script>
```
gmta: I've tried running your test case on `master`, but it doesn't crash. It does throw an exception however. In that case, you should make it a `Text` test instead, and try/catch the exception.

ljamar: Could you clarify what crash means? I assumed it was a failing test, but I'm guessing you mean a browser crash?

For the test output, is a simple `println("PASS")` and `println("FAIL")` enough?

gmta: A crash is when the process is abruptly killed as a result of a trap or segfault; using PASS/FAIL is the way to go!

PR #6245, opened 2025-09-19
Implement interpolation and compositing of `font-variation-settings` values

tcl3:
We also now only de-duplicate and sort the computed value rather than doing so at parse time.

The spec is written in a way that it doesn't necessarily expect the `font-variation-settings` values to be de-duplicated and sorted at the point we are interpolating them. In our implementation de-duplication and sorting has already happened by this time, so while the animation type is specified as "custom", there would actually be no difference if the animation type was set to "by-computed-value".

Gains us 129 WPT subtests in the `css/css-fonts` directory.


Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -901,7 +902,10 @@ void StyleComputer::collect_animation_into(DOM::AbstractElement abstract_element
         return;
     }
 
-    auto key = static_cast<i64>(round(output_progress.value() * 100.0 * Animations::KeyframeEffect::AnimationKeyFrameKeyScaleFactor));
+    double progress = output_progress.value() >= NumericLimits<double>::max() / (100.0 * Animations::KeyframeEffect::AnimationKeyFrameKeyScaleFactor)
+        ? NumericLimits<double>::max()
+        : round(output_progress.value() * 100.0 * Animations::KeyframeEffect::AnimationKeyFrameKeyScaleFactor);
+    auto key = progress >= NumericLimits<i64>::max() ? NumericLimits<i64>::max() : static_cast<i64>(progress);
```
tcl3: Before this change, UBSAN would complain when `key` was outside the range of `i64`. This change ensures the value of `key` is clamped to the correct range, but the subtests that triggered this failure still fail.

AtkinsSJ: I think it's possible for `progress` to be negative too, that should similarly be clamped to `NumericLimits<i64>::min()`

Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
```diff
@@ -3014,7 +3014,7 @@ RefPtr<StyleValue const> Parser::parse_font_variation_settings_value(TokenStream
     // previous value for that axis. This deduplication is observable by accessing the computed value of this property."
     // So, we deduplicate them here using a HashSet.
 
-    OrderedHashMap<FlyString, NonnullRefPtr<OpenTypeTaggedStyleValue const>> axis_tags_map;
+    StyleValueVector axis_tags;
```
AtkinsSJ, line 3101: The comment above needs removing/moving to the new location.

Libraries/LibWeb/CSS/Interpolation.cpp
```diff
@@ -1663,6 +1663,16 @@ RefPtr<StyleValue const> composite_value(StyleValue const& underlying_value, Sty
         auto result = composite_raw_values(underlying_value.as_number().number(), animated_value.as_number().number());
         return NumberStyleValue::create(result);
     }
+    case StyleValue::Type::OpenTypeTagged: {
```
AtkinsSJ, line 1666: Can we get a test for this? :^)

Libraries/LibWeb/CSS/StyleComputer.cpp
tcl3: Altered to clamp to the range of i64 in both directions. I've also added a FIXME, since there are WPT tests that rely on a larger value than we now clamp to (although I couldn't find where/if the relevant range is specced).

Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
tcl3: Comment moved and relevant spec link added.

Libraries/LibWeb/CSS/Interpolation.cpp
tcl3: I totally forgot to add the relevant test for this. Now added.

PR #6246, opened 2025-09-19
WIP

R-Goc:


konradekk:
(not really important: `result` instead of `hr`?)

R-Goc:
Calling an HRESULT hr is the typical convention. But if a maintainer prefers result I can change it.

Zaggy1024:
I'm also very accustomed to hr, but others may feel differently. I think following "convention" seems to be pretty normal if file descriptors are any indication, though.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.h
```diff
@@ -0,0 +1,39 @@
...
+
+#pragma once
+
+#include "PlaybackStream.h"
+#include <AK/Error.h>
+#include <AK/NonnullRefPtr.h>
+
+namespace Audio {
+
+struct AudioState;
```
Zaggy1024: Can AudioState be a nested private class so we don't pollute the namespace?

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
```diff
@@ -0,0 +1,359 @@
...
+
+// GUID for the playback session. That way all render streams have a single volume slider in the OS interface
+constexpr GUID PlaybackSessionGUID = { /* 22f2ca89-210a-492c-a0aa-f25b1d2f33a1 */
+    0x22f2ca89,
+    0x210a,
+    0x492c,
+    { 0xa0, 0xaa, 0xf2, 0x5b, 0x1d, 0x2f, 0x33, 0xa1 }
+};
+
+struct AudioState : public RefCounted<AudioState> {
```
Zaggy1024: This should be AtomicRefCounted, since it's a shared reference between two threads.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
```diff
@@ -0,0 +1,359 @@
...
+
+    state->buffer_event = CreateEvent(NULL, FALSE, FALSE, NULL);
+    if (!state->buffer_event)
+        return Error::from_windows_error(hr);
+
+    hr = state->audio_client->SetEventHandle(state->buffer_event);
+    if (FAILED(hr))
+        return Error::from_windows_error(hr);
+
+    state->audio_thread = Threading::Thread::construct([&state] { return AudioState::render_thread_loop(state.ptr()); });
```
Zaggy1024: We need to add a refcount to state here to ensure it survives for the duration of the thread. The refcount can be added by taking `state` by value in the captures, and then passed by reference to `render_thread_loop`.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
```diff
@@ -0,0 +1,359 @@
...
+        if (FAILED(hr))
+            return Error::from_windows_error(hr);
+    }
+
+    return TRY(adopt_nonnull_ref_or_enomem(new (nothrow) PlaybackStreamWasapi(move(state))));
+}
+
+int AudioState::render_thread_loop(void* context)
+{
+    auto* state = static_cast<AudioState*>(context);
```
Zaggy1024: Why does it need to take the audio state as a void pointer?

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
```diff
@@ -0,0 +1,359 @@
...
+        if (bytes_written.is_empty()) {
+            if (state->underrun_callback)
+                state->underrun_callback();
+            // TODO: Is making sure the audio is silent all that is needed if we have no data?
+            buffer_flags |= AUDCLNT_BUFFERFLAGS_SILENT;
+        }
+        // TODO: Have some heuristic on how many bytes need to be in the buffer before silence needs to be added to the end.
+        // If there was just one byte written we would quickly run out.
+        frames_available = bytes_written.size() / block_align;
+        hr = state->render_client->ReleaseBuffer(frames_available, buffer_flags);
```
Zaggy1024: Not sure I understand this part, does it play more frames from the buffer than the number of frames passed to ReleaseBuffer?

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
```diff
@@ -0,0 +1,359 @@
...
+    UINT64 frequency;
+    UINT64 position;
+    hr = m_state->clock->GetFrequency(&frequency);
+    if (FAILED(hr))
+        return Error::from_windows_error(hr);
+    hr = m_state->clock->GetPosition(&position, NULL);
+    if (FAILED(hr))
+        return Error::from_windows_error(hr);
+    // This is a narrowing conversion. Not much can be done though.
+    return AK::Duration::from_seconds(position / frequency);
```
Zaggy1024: This should definitely at the very least return `from_milliseconds(position * 1000 / frequency)`, if not `from_nanoseconds(position * 1'000'000'000 / frequency)`. I have some helper functions I've been unfortunately duplicating in a couple places for even more lossless conversion of sample index to duration that I should probably generalize, but either of the above should be okay for this.

If this warns about a narrowing conversion, use a static_cast to make it explicit.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
```diff
@@ -0,0 +1,359 @@
...
+    if (FAILED(hr)) {
+        promise->reject(Error::from_windows_error(hr));
+        return promise;
+    }
+    ScopeGuard uninitialize_com = [] { CoUninitialize(); };
+
+    float clamped_volume = static_cast<float>(clamp(volume, 0.0, 1.0));
+    auto channels = m_state->wave_format->nChannels;
+    Vector<float> volumes;
+    volumes.resize(channels);
```
Zaggy1024: I'm mildly concerned about malloc churn when dragging the volume slider, perhaps this could be allocated once ahead of time?

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
```diff
@@ -0,0 +1,359 @@
...
+
+// NOTE: Not using the newer winrt that supersedes wrl as that uses exceptions for error handling
+#include <wrl/client.h>
+
+namespace Audio {
+
+using namespace Microsoft::WRL;
+
+// GUID for the playback session. That way all render streams have a single volume slider in the OS interface
+constexpr GUID PlaybackSessionGUID = { /* 22f2ca89-210a-492c-a0aa-f25b1d2f33a1 */
```
Zaggy1024: Style nitpick: We avoid `/**/` comments wherever possible. Also, it might be okay to just omit a comment here, given that it reads in the same order in its initialization.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
Zaggy1024: Why does this need to take the audio state as a void pointer?

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
R-Goc: It doesn't. Not sure how that callback works. Can it write just a single frame or will it always write the whole thing? If it writes a single frame(extreme case) that will play and then the buffer will be empty before the we could get a chance at the next iteration. Not actually sure what happens if we didn't provide data on time. Hope it's just silence but I will check.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
R-Goc: That was just autogenerated by uuidgen. Should I remove it?

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
R-Goc: Forgot we don't need to do that with our thread impl.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
Zaggy1024: The callback is allowed to write any amount of data to the buffer up to its size, and it's required to write a multiple of a frame's size in bytes.

My hope/expectation would be that the audio client would play the frames that have been written, then stop increasing the result of `GetPosition()` and wait for more data. If it continues increasing when no data is written, then we may have a problem with audio timing/synchronization if we don't work around it. Definitely give that a test however you can to ensure that it handles that case without desyncing the time.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
Zaggy1024: You could remove it, or change it to a `//` comment. I don't really have a preference either way :^)

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
R-Goc: Reading the docs again the order here is wrong. ReleaseBuffer needs to return exactly as much as was requested by GetBuffer.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
```diff
@@ -0,0 +1,359 @@
...
+    PlaybackStreamWasapi::AudioDataRequestCallback data_request_callback;
+    Function<void()> underrun_callback;
+
+    Threading::Mutex mutex;
+    Atomic<bool> playing = false;
+    Atomic<bool> drain_and_suspend = false;
+    Atomic<bool> exit_requested = false;
+
+    static int render_thread_loop(void* context);
+    RefPtr<Threading::Thread> audio_thread;
```
Zaggy1024: Instead of holding onto the thread here, you can detach the thread and let your references to it go out of scope. The thread class guarantees that its action and all its captures stay alive for the duration of its runtime while detached.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
Zaggy1024: My understanding is that's only the case if we're using event-driven buffering _and_ exclusive mode, which I think we should only be doing the former?

R-Goc: It isn't:
> The client must release the same number of frames that it requested in the preceding call to the [IAudioRenderClient::GetBuffer](https://learn.microsoft.com/en-us/windows/desktop/api/audioclient/nf-audioclient-iaudiorenderclient-getbuffer) method. The single exception to this rule is that the client can always call ReleaseBuffer to release 0 frames (unless the stream is exclusive mode and uses event-driven buffering).

Zaggy1024: Huh, the phrasing of other parts of the documentation of that method seems to imply that it's capable of accepting a value less than that:

> [in] NumFramesWritten
> The number of audio frames written by the client to the data packet. The value of this parameter must be less than or equal to the size of the data packet, as specified in the NumFramesRequested parameter passed to the [IAudioRenderClient::GetBuffer](https://learn.microsoft.com/en-us/windows/desktop/api/audioclient/nf-audioclient-iaudiorenderclient-getbuffer) method.

> AUDCLNT_E_BUFFER_SIZE_ERROR
> The stream is exclusive mode and uses event-driven buffering, but the client attempted to release a packet that was not the size of the buffer.

If testing indicates that it doesn't support any value other than 0 or the buffer size, though, that's that I suppose, and we'll have to implement our own timing based off GetPosition() while accounting for smaller buffer sizes.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
R-Goc: Does this still apply

Zaggy1024: I would imagine so, the usual model is to have the thread's function hold a strong reference to an AtomicRefCounted with the data that is shared between it and the public class (which also holds a strong reference).

Zaggy1024: To expand on this, the way it should be written is:

```cpp
auto thread = Threading::Thread::construct([state /* : NonnullRefPtr<AudioState> copy constructor ref()s */] {
    return AudioState::render_thread_loop(*state);  // Dereference the pointer and pass it as a reference, it will live for the duration of the call
});
thread->start();
thread->detach();
```

Thus, as long as the thread's function is running (even after the `PlaybackStream` is destroyed), the `AudioState` will stay alive. This setup prevents a reference loop (`Thread` -> `AudioState` -> `Thread`) while ensuring that the don't UAF the `AudioState` in the thread.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
```diff
@@ -0,0 +1,502 @@
...
+    0x492c,
+    { 0xa0, 0xaa, 0xf2, 0x5b, 0x1d, 0x2f, 0x33, 0xa1 }
+};
+
+struct Task {
+    enum class Kind : u8 {
+        Play,
+        DrainAndSuspend,
+        DiscardAndSuspend,
+    };
```
Zaggy1024: A bit nitpicky, but we could represent the tasks through a `Variant` of POD structs, so that we can differentiate the promise parameter type between each task. Then handling can be a simple `visit()`.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
```diff
@@ -0,0 +1,502 @@
...
+            [](Empty) { VERIFY_NOT_REACHED(); },
+            [&error](auto& promise) {
+                promise->reject(move(error));
+            });
+    }
+    Variant<Empty, NonnullRefPtr<Core::ThreadedPromise<void>>, NonnullRefPtr<Core::ThreadedPromise<AK::Duration>>> promise;
+    Kind kind;
+};
+
+class ComHelper {
```
Zaggy1024: Nit: it feels like this could use a more specific name, like COMUninitializer.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
```diff
@@ -0,0 +1,502 @@
...
+public:
+    ~ComHelper()
+    {
+        if (initialized)
+            CoUninitialize();
+    }
+    bool initialized = false;
+};
+
+static thread_local ComHelper com_helper {};
```
Zaggy1024:
```suggestion
static thread_local ComHelper s_com_helper {};
```

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
```diff
@@ -0,0 +1,502 @@
...
+    VERIFY(task_event);
+}
+
+PlaybackStreamWasapi::AudioState::~AudioState()
+{
+    CloseHandle(buffer_event);
+}
+
+// The caller has the responsibility to initialize com
+ALWAYS_INLINE AK::Duration PlaybackStreamWasapi::total_time_played_impl(PlaybackStreamWasapi::AudioState* state)
```
Zaggy1024: We could integrate the comment into the function name: `total_time_played_with_com_initialized`

Libraries/LibMedia/Audio/PlaybackStreamWasapi.h
```diff
@@ -0,0 +1,39 @@
...
+
+#pragma once
+
+#include "PlaybackStream.h"
+#include <AK/Error.h>
+#include <AK/NonnullRefPtr.h>
+
+namespace Audio {
+
+class PlaybackStreamWasapi final : public PlaybackStream {
```
Zaggy1024: nit: Our usual naming scheme for classes capitalizes all characters in such abbreviations, i.e. AVIFLoader, so this should be named PlaybackStreamWASAPI.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
```diff
@@ -0,0 +1,502 @@
...
+PlaybackStreamWasapi::PlaybackStreamWasapi(NonnullRefPtr<AudioState> state)
+    : m_state(move(state))
+{
+}
+
+PlaybackStreamWasapi::~PlaybackStreamWasapi()
+{
+    m_state->exit_requested.store(true, MemoryOrder::memory_order_release);
+    // Poke the event to wake the thread up from wait
+    if (m_state->buffer_event)
```
Zaggy1024: It looks to me like it's not possible for `buffer_event` to be null, so we should instead `VERIFY(m_state->buffer_event != nullptr)`.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
Zaggy1024: Additionally, we don't need to call `Thread::join()` if we detach it, so we can avoid blocking the main thread upon shutdown of the stream.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
```diff
@@ -0,0 +1,502 @@
...
+            dbgln("GetBuffer failed with: {}", Error::from_windows_error(hr));
+            VERIFY_NOT_REACHED();
+        }
+
+        u32 buffer_size = frames_available * block_align;
+        ReadonlyBytes bytes_written = state->data_request_callback({ buffer, buffer_size }, PcmSampleFormat::Float32, frames_available);
+        if (bytes_written.is_empty()) {
+            if (state->underrun_callback)
+                state->underrun_callback();
+            buffer_flags |= AUDCLNT_BUFFERFLAGS_SILENT;
```
Zaggy1024, line 426: With `buffer_flags` at the function scope, won't this cause the audio to remain silent permanently after the first underrun?

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
```diff
@@ -0,0 +1,502 @@
...
+        // If there was just one byte written we would quickly run out.
+        frames_available = bytes_written.size() / block_align;
+        hr = state->render_client->ReleaseBuffer(frames_available, buffer_flags);
+        if (FAILED(hr)) {
+            dbgln("ReleaseBuffer failed with: {}", Error::from_windows_error(hr));
+            VERIFY_NOT_REACHED();
+        }
+    }
+
+    AvRevertMmThreadCharacteristics(task_handle);
```
Zaggy1024: If it's necessary to call this before the thread exits, it would be nice to put this in a scope guard so that it can exist next to where it's being set.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.h
```diff
@@ -0,0 +1,39 @@
...
+
+    virtual AK::Duration total_time_played() const override;
+
+    virtual NonnullRefPtr<Core::ThreadedPromise<void>> set_volume(double) override;
+
+private:
+    struct AudioState;
+    explicit PlaybackStreamWasapi(NonnullRefPtr<AudioState>);
+    static ALWAYS_INLINE AK::Duration total_time_played_impl(AudioState* state);
+    ~PlaybackStreamWasapi();
```
Zaggy1024: Very nitpicky style critique, it might be nice to have some spacing to group things here:

```suggestion
    struct AudioState;

    explicit PlaybackStreamWasapi(NonnullRefPtr<AudioState>);
    ~PlaybackStreamWasapi();

    static ALWAYS_INLINE AK::Duration total_time_played_impl(AudioState* state);
```

Services/WebContent/main.cpp
```diff
@@ -55,6 +59,12 @@ ErrorOr<int> ladybird_main(Main::Arguments arguments)
 {
     AK::set_rich_debug_enabled(true);
 
+#if defined(AK_OS_WINDOWS)
+    // Otherwise SDL inits COM in the APARTMENTTHREADED model which we don't want.
```
Zaggy1024: This seems fine, but it might be nice to note the reason we need the multithreaded model here as well. Also, imperative mood probably sounds better here:

> Initialize COM with the MULTITHREADED model, or SDL will initialize it later with the APARTMENTTHREADED model instead, which is incompatible with WASAPI.

Services/WebContent/main.cpp
```diff
@@ -55,6 +59,12 @@ ErrorOr<int> ladybird_main(Main::Arguments arguments)
...
+    HRESULT hr = CoInitializeEx(0, COINIT_MULTITHREADED);
+    VERIFY(SUCCEEDED(hr));
+    ScopeGuard uninitizalize_com = []() { CoUninitialize(); };
```
Zaggy1024: Typo in the variable name, it has a stray z 😅

Zaggy1024 (changes requested):

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
```diff
@@ -0,0 +1,502 @@
...
+    WORD block_align = state->wave_format->Format.nBlockAlign;
+
+    ScopeGuard uninitialize_com = [] { CoUninitialize(); };
+
+    DWORD task_index = 0;
+    HANDLE task_handle = AvSetMmThreadCharacteristicsW(L"Pro Audio", &task_index);
+
+    while (!state->exit_requested.load(MemoryOrder::memory_order_acquire)) {
+        Array<HANDLE, 2> handles = { state->task_event, state->buffer_event };
+        DWORD result = WaitForMultipleObjects(2, handles.data(), FALSE, INFINITE);
```
Zaggy1024: I feel like this should either use a stack-allocated array:

```suggestion
        auto handles = { state->task_event, state->buffer_event };
        DWORD result = WaitForMultipleObjects(2, handles.data(), FALSE, INFINITE);
```

or infer the template parameters for `Array` and use its `size()`:

```suggestion
        Array handles = { state->task_event, state->buffer_event };
        DWORD result = WaitForMultipleObjects(handles.size(), handles.data(), FALSE, INFINITE);
```

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
```diff
@@ -0,0 +1,502 @@
...
+    ~AudioState();
+
+    ComPtr<IMMDeviceEnumerator> enumerator;
+    ComPtr<IMMDevice> device;
+    ComPtr<IAudioClient> audio_client;
+    ComPtr<IAudioRenderClient> render_client;
+    ComPtr<IAudioStreamVolume> audio_stream_volume;
+    ComPtr<IAudioClock> clock;
+
+    NonnullOwnPtr<WAVEFORMATEXTENSIBLE> wave_format;
```
Zaggy1024: This shouldn't need to be a pointer. Perhaps we could even get away with stack-allocating it in the `create()` function, then just store the specific information that is needed by the thread loop?

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
```diff
@@ -0,0 +1,502 @@
...
+    state->data_request_callback = move(data_request_callback);
+
+    WAVEFORMATEXTENSIBLE* ideal_format;
+    state->audio_client->GetMixFormat(reinterpret_cast<WAVEFORMATEX**>(&ideal_format));
+    dbgln("Mixing engine audio format:\n");
+    print_audio_format(ideal_format);
+    CoTaskMemFree(ideal_format);
+    // Set up 32bit float PCM
+    u16 bits_per_sample = 32;
+    u16 block_align = (channels * bits_per_sample) / 8;
```
Zaggy1024: Would be nice to make this a bit clearer:

```suggestion
    u16 bits_per_sample = sizeof(float) * __CHAR_BIT__;
    u16 block_align = channels * sizeof(float);
```

Zaggy1024 (commented):
A couple more things I've noticed.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
```diff
@@ -0,0 +1,508 @@
...
+    state->data_request_callback = move(data_request_callback);
+
+    WAVEFORMATEXTENSIBLE* device_format;
+    state->audio_client->GetMixFormat(reinterpret_cast<WAVEFORMATEX**>(&device_format));
+    ScopeGuard free_mix_format = [&device_format] { CoTaskMemFree(device_format); };
+
+    dbgln("Mixing engine audio format:\n");
+    print_audio_format(device_format);
+
+    Vector<Channel, 32> channel_vec;
```
Zaggy1024: This should use `ChannelMap::capacity()` to determine the vector inline size.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
```diff
@@ -0,0 +1,508 @@
...
+    auto channels = device_format->Format.nChannels;
+
+    // Conveniently the enum is in the same order. The difference is that KSMedia uses a bitmask instaead
+    for (u64 i = 0; i < channels; i++) {
+        u32 mask = 1 << i;
+        if (device_format->dwChannelMask & mask)
+            // i + 1 as the first bit is at i == 0
+            channel_vec.append(static_cast<Channel>(i + 1));
+    }
+    ChannelMap channel_map { channel_vec };
```
Zaggy1024: Would be nice to factor out the channel map conversion (or if you prefer, the entire sample specification) into a function that takes `WAVEFORMATEXTENDED&`.

Shouldn't we also handle cases where `wFormatTag != WAVE_FORMAT_EXTENSIBLE`? (When I implemented a conversion function myself, I just checked nChannels for mono/stereo when it isn't extensible.)

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
```diff
@@ -0,0 +1,508 @@
...
+    state->wave_format->Format.nBlockAlign = device_format->Format.nBlockAlign;
+    state->wave_format->Format.wBitsPerSample = device_format->Format.wBitsPerSample;
+    state->wave_format->Format.cbSize = sizeof(WAVEFORMATEXTENSIBLE) - sizeof(WAVEFORMATEX);
+    state->wave_format->Samples.wValidBitsPerSample = device_format->Samples.wValidBitsPerSample;
+    state->wave_format->dwChannelMask = device_format->dwChannelMask;
+    state->wave_format->SubFormat = device_format->SubFormat;
+
+    // FIXME: From what I understand the audio mixer uses 32bit floats internally. If that is not always the case we need to handle it.
+    VERIFY(device_format->SubFormat == KSDATAFORMAT_SUBTYPE_IEEE_FLOAT);
+    VERIFY(device_format->Format.wBitsPerSample == 32);
```
Zaggy1024: Instead of verifying this, could we just create a format that reuses the sample rate and channel count/mask, but with the float subformat? Then we don't need a verify here, and we can just return an error if the API rejects the format.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
```diff
@@ -0,0 +1,508 @@
...
+    bool drain_and_suspend = false;
+    Atomic<bool> exit_requested = false;
+
+    static int render_thread_loop(AudioState& state);
+    RefPtr<Threading::Thread> audio_thread;
+    RefPtr<Core::ThreadedPromise<AK::Duration>> resume_promise;
+    RefPtr<Core::ThreadedPromise<void>> suspend_promise;
+
+    // I doubt we will have anyone with more than 7.1
+    Vector<float, 8> channel_volumes;
```
Zaggy1024: We could just use `ChannelMap::capacity()` here, if you want to have a named magic number to use. :^)

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
```diff
@@ -0,0 +1,508 @@
...
+        auto output_buffer = Bytes(buffer, buffer_size).reinterpret<float>();
+        auto bytes_written = state.data_request_callback(output_buffer);
+        if (bytes_written.is_empty()) [[unlikely]] {
+            if (state.underrun_callback)
+                state.underrun_callback();
+            buffer_flags |= AUDCLNT_BUFFERFLAGS_SILENT;
+        }
+        // TODO: Have some heuristic on how many bytes need to be in the buffer before silence needs to be added to the end.
+        // If there was just one byte written we would quickly run out.
+        frames_available = bytes_written.size() / block_align;
```
Zaggy1024: This should divide by the channel count, since the size of the span is already in samples.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
R-Goc: It can just sit as a member in AudioState. It isn't huge.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
R-Goc: This is never the case. GetMixFormat always returns a WAVEFORMATEXTENSIBLE,  I could add the docs that say that as a comment

Zaggy1024: Then we should definitely assert that that's the case before we try to convert it, otherwise we're going to be accessing data that hasn't been initialized.

R-Goc: No real point in my opinion, but sure. It's about the same as an api that returns bool and is documented to only return true. Would you prefer an assert on the format type or size of the extension?

Zaggy1024: Both would be preferable.

The documentation states that it can return WAVEFORMATEX or WAVEFORMATEXTENSIBLE, which makes me think that the remarks indicating that it always returns WAVEFORMATEXTENSIBLE was actually a change they made sometime after the API was initially created, so I'd rather not just assume that it is always or will remain the case.

R-Goc: In general remarks is the part that matters. Waveformatextensible can represent all waveformatex values. Exactly this part:
> For this reason, the GetMixFormat method retrieves a format descriptor that is in the form of a WAVEFORMATEXTENSIBLE structure instead of a standalone WAVEFORMATEX structure.

Zaggy1024: Indeed, I know that the remarks section said that it returns the extensible struct, but when it comes to things that would result in out-of-bounds reads, I'd rather write our assumptions into the code than make them implicit.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
```diff
@@ -0,0 +1,511 @@
...
+    if (!state->buffer_event)
+        return Error::from_windows_error(hr);
+
+    TRY_HR(state->audio_client->SetEventHandle(state->buffer_event));
+    TRY_HR(state->clock->GetFrequency(&state->audio_client_clock_frequency));
+
+    if (initial_output_state == OutputState::Playing)
+        state->playing = true;
+
+    state->audio_thread = Threading::Thread::construct([state] {
```
Zaggy1024: We can't hold a strong reference to the audio thread here, or we'll end up with a reference loop and never destroy the state.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
```diff
@@ -0,0 +1,512 @@
...
+
+PlaybackStreamWASAPI::AudioState::AudioState()
+{
+    task_event = CreateEvent(NULL, FALSE, FALSE, NULL);
+    VERIFY(task_event);
+}
+
+PlaybackStreamWASAPI::AudioState::~AudioState()
+{
+    CloseHandle(buffer_event);
```
Zaggy1024: Do we need to close `task_event` here too?

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
```diff
@@ -0,0 +1,512 @@
...
+
+    WAVEFORMATEXTENSIBLE* device_format;
+    state->audio_client->GetMixFormat(reinterpret_cast<WAVEFORMATEX**>(&device_format));
+    ScopeGuard free_mix_format = [&device_format] { CoTaskMemFree(device_format); };
+
+    dbgln("Mixing engine audio format:\n");
+    print_audio_format(*device_format);
+
+    VERIFY(device_format->Format.nChannels <= 32);
+    VERIFY(device_format->dwChannelMask != 0);
```
Zaggy1024: To make it a little clearer what's going on here, perhaps:

```cpp
VERIFY(device_format->Format.nChannels != 0);
VERIFY(popcount(device_format->dwChannelMask) == device_format->Format.nChannels);
```

Also, this should go after verifying `WAVE_FORMAT_EXTENSIBLE` so we're not accessing uninitialized memory.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
```diff
@@ -0,0 +1,512 @@
...
+        Span<u8> { reinterpret_cast<char*>(&format.SubFormat), 16 });
+}
+
+static ChannelMap convert_bitmask_to_channel_map(u32 channel_mask, u32 channels)
+{
+    Vector<Channel, ChannelMap::capacity()> channel_vec;
+    for (u64 i = 0; i < channels; i++) {
+        if (channel_mask & (1 << i))
+            // i + 1 as the first bit is at i == 0
+            channel_vec.append(static_cast<Channel>(i + 1));
```
Zaggy1024: I don't like to assume that the enum order is the same (or will stay the same) here, since it's not based on any spec at the moment. All other conversions use an enum to enum mapping to avoid this assumption.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
```diff
@@ -0,0 +1,512 @@
...
+        format.Samples.wValidBitsPerSample,
+        format.dwChannelMask,
+        Span<u8> { reinterpret_cast<char*>(&format.SubFormat), 16 });
+}
+
+static ChannelMap convert_bitmask_to_channel_map(u32 channel_mask, u32 channels)
+{
+    Vector<Channel, ChannelMap::capacity()> channel_vec;
+    for (u64 i = 0; i < channels; i++) {
+        if (channel_mask & (1 << i))
```
Zaggy1024: Doesn't this assume that the channel mask has bits 0 to `channels - 1` set, rather than checking all bits?

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
```diff
@@ -0,0 +1,512 @@
...
+    SetEvent(m_state->buffer_event);
+}
+
+ErrorOr<NonnullRefPtr<PlaybackStream>> PlaybackStream::create(OutputState initial_output_state, u32 target_latency_ms, SampleSpecificationCallback&& specification_callback, AudioDataRequestCallback&& data_callback)
+{
+    return PlaybackStreamWASAPI::create(initial_output_state, target_latency_ms, move(specification_callback), move(data_callback));
+}
+
+static void print_audio_format(WAVEFORMATEXTENSIBLE& format)
+{
```
Zaggy1024, line 167: If we keep using this function, we should also assert that the format is extensible, or build a string based on which fields are present.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
```diff
@@ -0,0 +1,512 @@
...
+    bool playing = false;
+    bool drain_and_suspend = false;
+    Atomic<bool> exit_requested = false;
+
+    static int render_thread_loop(AudioState& state);
+    RefPtr<Threading::Thread> audio_thread;
+    RefPtr<Core::ThreadedPromise<AK::Duration>> resume_promise;
+    RefPtr<Core::ThreadedPromise<void>> suspend_promise;
+
+    // I doubt we will have anyone with more than 7.1
```
Zaggy1024: This comment is no longer relevant.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
```diff
@@ -0,0 +1,512 @@
...
+PlaybackStreamWASAPI::AudioState::~AudioState()
+{
+    CloseHandle(buffer_event);
+}
+
+ALWAYS_INLINE AK::Duration PlaybackStreamWASAPI::total_time_played_with_com_initialized(PlaybackStreamWASAPI::AudioState& state)
+{
+    UINT64 position;
+    MUST_HR(state.clock->GetPosition(&position, nullptr));
+    // FIXME: This is a narrowing conversion. We might want a u64 overload of from_nanoseconds.
```
Zaggy1024: This comment is outdated, but probably fine to do away with anyway, since we're clamping to i64.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
Zaggy1024: CloseHandle documentation states that the handle should be open, but buffer_event can be null here if we fail somewhere in `PlaybackStreamWASAPI::create` before it is assigned. We probably need a null check here?

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
```diff
@@ -0,0 +1,512 @@
...
+        auto floats_written = state.data_request_callback(output_buffer);
+        if (floats_written.is_empty()) [[unlikely]] {
+            if (state.underrun_callback)
+                state.underrun_callback();
+            buffer_flags |= AUDCLNT_BUFFERFLAGS_SILENT;
+        }
+        // TODO: Have some heuristic on how many bytes need to be in the buffer before silence needs to be added to the end.
+        // If there was just one byte written we would quickly run out.
+        frames_available = floats_written.size() / state.wave_format.Format.nChannels;
+        MUST_HR(state.render_client->ReleaseBuffer(frames_available, buffer_flags));
```
Zaggy1024: Not sure I understand the comment here, I forget if we discussed this before. However, based on my testing, we should never release the buffer with more frames than we were given by the callback unless running out causes corruption, since it counts samples accurately in the audio clock.

If this is about zeroing the buffers, currently AudioMixingSink already does that to the whole buffer in order to mix audio properly. We could potentially defensively zero the remaining bytes in the buffer after the callback, but I don't think it's likely to be necessary.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
R-Goc: Not sure what you mean here. Which reference should I make weak?

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
R-Goc: The real solution would be an owning handle type but yes.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
R-Goc: The issue is this: 
> The client must release the same number of frames that it requested in the preceding call to the [IAudioRenderClient::GetBuffer](https://learn.microsoft.com/en-us/windows/desktop/api/audioclient/nf-audioclient-iaudiorenderclient-getbuffer) method.

https://learn.microsoft.com/en-us/windows/win32/api/audioclient/nf-audioclient-iaudiorenderclient-releasebuffer#remarks

I think we need to pad it if the callback ever returned too little. I haven't seen it happen, or haven't noticed it but there may be some edge case where it happens
And yes that comment doesn't make sense

R-Goc: So I think it should actually look like:
```cpp
        BYTE* buffer;
        MUST_HR(state.render_client->GetBuffer(frames_available, &buffer));

        DWORD buffer_flags = 0;
        u32 buffer_size = frames_available * block_align;
        auto output_buffer = Bytes(buffer, buffer_size).reinterpret<float>();
        auto floats_written = state.data_request_callback(output_buffer);
        if (floats_written.is_empty()) [[unlikely]] {
            if (state.underrun_callback)
                state.underrun_callback();
            buffer_flags |= AUDCLNT_BUFFERFLAGS_SILENT;
        }
        u64 bytes_filled = floats_written.size() * sizeof(float);
        if (bytes_filled != buffer_size) {
            BYTE* end_of_fill = buffer + bytes_filled;
            memset(end_of_fill, 0, buffer_size - bytes_filled);
        }
        MUST_HR(state.render_client->ReleaseBuffer(frames_available, buffer_flags));
        ```

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
R-Goc: I did want to avoid macros, but seems like there is no way to do so.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
Zaggy1024: You don't need to store the `Thread` instance at all, since it's kept alive by the thread itself. You can simply construct the thread, start it, detach it, then let the reference go immediately. A good example of this is either of the DataProvider.cpp files.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
Zaggy1024: Yeah, it's not necessarily the prettiest thing, but I think it's a small price to pay here.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
Zaggy1024: I wanna say that I tested stopping writing partway through a buffer in AudioMixingSink to test the audio clock stopping somewhere other than a buffer boundary, and didn't I run into issues with that. It would definitely be worth confirming that this is necessary before we start padding the buffers.

If we do have to pad it, then we'll have to handle this in the time played calculation in order to make sure that we don't count those padding frames, which seems like it would be a bit of a pain.

Zaggy1024:
Not sure if the latest changes were supposed to resolve this, but it looks like I can still get a seek to make it stop processing the latest task and desync the playback state.

Also, it would be good to make sure that all of AudioState's primitive fields are default-initialized, I forgot to keep an eye out for that before. I noticed you set some of them in the latest force push, but not all. 😅

R-Goc:
No it wasn't meant to do so yet. I'll probably add processing multiple tasks next.

Zaggy1024 (approved):

Issue #5962, opened 2025-08-22
Crash when passing Float64Array to `WebGLRenderingContext.uniform4fv()`


tcl3:
I initially came across this issue when trying to load: https://renewables-map.robinhawkes.com/

I was able to reduce this to the following:
```html
<!DOCTYPE html>
<canvas id="glcanvas"></canvas>
<script>
    const canvas = document.getElementById("glcanvas");
    const gl = canvas.getContext("webgl");
    const program = gl.createProgram();
    const uColor = gl.getUniformLocation(program, "uColor");
    gl.uniform4fv(uColor, new Float64Array([0.1, 0.2, 0.3, 0.4]));
</script>
```

Our code assumes that if a TypedArray type is passed as the second argument that it must be a Float32Array, but other browsers seem perfectly happy to take a Float64Array.

I came up with a [more complete example](https://gist.github.com/tcl3/f2cc9b75143a2e6f29ca1bcd407614ca), to prove that other browsers are happy to take a Float64Array instead of a Float32Array here.

PR #6248, opened 2025-09-19
Webgl handle float64array uniform variable update

Davenporten:
Created to fix this issue: https://github.com/LadybirdBrowser/ladybird/issues/5962

These changes include the following for both WebGL 1 and 2:

- Abstracting the extraction of buffer data for all unifor*fv methods to be in one place.
- Adds support for passing in `Float64Array` buffers (previously only straight vectors or `Float32Array` were supported; this includes transforming the Float64 data to Float32 for the opengl api.
- Adds relevant tests.

Notes:

- The helper methods for both WebGL 1 and 2 are identical. I considered moving them out of the separate classes, but wanted to avoid any unnecessary coupling.
- This logic could likely be generalized to more than just the uniform variable update methods, but as this was my first attempted change to the repo I wanted to keep the scope as narrow as possible.


gmta:
Closing as superseded by #6346 - thank you for looking into this, regardless!

PR #6249, opened 2025-09-19
LibGfx: Remove unnecessary return statements and preprocessor directive

okayscylla:
The switch statement for font_format_is_supported() returned false for every case which is not necessary so was reorganised. There was also a preprocessor directive referencing a definition that seemingly did not exist that was removed.

(I am sleepy right now so if this is a stupid pr then I am sorry for the inconvenience :smile:)


Libraries/LibGfx/Font/FontSupport.cpp
```diff
@@ -14,20 +14,15 @@ bool font_format_is_supported(FontFormat const format)
...
     case FontFormat::WOFF:
-        return true;
     case FontFormat::WOFF2:
         return true;
+    case FontFormat::EmbeddedOpenType:
+    case FontFormat::SVG:
+        return false;
     }
 
     return false;
```
R-Goc: There's no point of this return at the bottom existing if a default case is added. Does doing:
```cpp
bool font_format_is_supported(FontFormat const format)
{
    // FIXME: Determine these automatically.
    switch (format) {
    case FontFormat::OpenType:
    case FontFormat::TrueType:
    case FontFormat::TrueTypeCollection:
    case FontFormat::WOFF:
    case FontFormat::WOFF2:
        return true;
    default:
        return false;
    }
}
```
work without warnings?
And as for the fixme I believe that it refers to querying some kind of API to figure out supported formats at runtime instead of hardcoding them.

okayscylla: I'm stupid lol. I just pushed a commit that should fix this so if you have time let me know if this works. Thank you!

okayscylla:
Oops I didn't mean to close it sorry :)

shlyakpavel:
Non-maintainer opinion:
I would prefer these two commits to be squished

okayscylla:
im silly i did it again lol

PR #6251, opened 2025-09-20
LibWeb/HTML: Two small crash fixes for OffscreenCanvas

shannonbooth:
Haven't contributed in a little bit, figured I'd patch two low hanging crash fixes from WPT


shannonbooth:
oops, I didn't switch on my brain properly, need to not set the fields when throwing

PR #6252, opened 2025-09-20
LibWeb: Add MathML Element presentational hints

lpas:
This PR Adds 2 MathML Elements mspace & mi and the presentational hints for MathML Elements

### DIFF
--- a/Libraries/LibWeb/Forward.h
+++ b/Libraries/LibWeb/Forward.h
@@ -865,6 +865,8 @@ struct LayoutState;
 namespace Web::MathML {
 
 class MathMLElement;
+class MathMLMspaceElement;
+class MathMLMiElement;
 
 }
 
--- a/Libraries/LibWeb/MathML/MathMLElement.cpp
+++ b/Libraries/LibWeb/MathML/MathMLElement.cpp
@@ -6,6 +6,12 @@
 
 #include <LibWeb/Bindings/ExceptionOrUtils.h>
 #include <LibWeb/Bindings/MathMLElementPrototype.h>
+#include <LibWeb/CSS/Parser/Parser.h>
+#include <LibWeb/CSS/StyleValues/IntegerStyleValue.h>
+#include <LibWeb/CSS/StyleValues/KeywordStyleValue.h>
+#include <LibWeb/CSS/StyleValues/MathDepthStyleValue.h>
+#include <LibWeb/HTML/Numbers.h>
+#include <LibWeb/HTML/Parser/HTMLParser.h>
 #include <LibWeb/MathML/MathMLElement.h>
 #include <LibWeb/MathML/TagNames.h>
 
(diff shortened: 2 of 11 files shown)


Libraries/LibWeb/MathML/MathMLMiElement.h
```diff
@@ -0,0 +1,27 @@
...
+#include <LibWeb/MathML/MathMLElement.h>
+
+namespace Web::MathML {
+
+class MathMLMiElement final : public MathMLElement {
+    WEB_PLATFORM_OBJECT(MathMLMiElement, MathMLElement);
+    GC_DECLARE_ALLOCATOR(MathMLMiElement);
+
+public:
+    virtual ~MathMLMiElement() override = default;
```
gmta, line 18: Isn't this the default? I believe you can omit this here and in MathMLspaceElement.

Libraries/LibWeb/MathML/MathMLElement.cpp
```diff
@@ -59,4 +65,61 @@ void MathMLElement::visit_edges(JS::Cell::Visitor& visitor)
...
+}
+
+void MathMLElement::apply_presentational_hints(GC::Ref<CSS::CascadedProperties> cascaded_properties) const
+{
+    for_each_attribute([&](auto& name, auto& value) {
+        if (name == "dir") {
+            // https://w3c.github.io/mathml-core/#attributes-common-to-html-and-mathml-elements
+            // The dir attribute, if present, must be an ASCII case-insensitive match to ltr or rtl. In that case, the user agent is expected to treat the attribute as a presentational hint setting the
+            // element's direction property to the corresponding value. More precisely, an ASCII case-insensitive match to rtl is mapped to rtl while an ASCII case-insensitive match to ltr is mapped to ltr.
+            // if ("ltr".equals_ignoring_ascii_case){}
```
gmta: Leftover commented out code?

Tests/LibWeb/Text/expected/MathML/presentational_hints.txt
```diff
@@ -0,0 +1,24 @@
+direction: ltr
+direction: ltr
+direction: rtl
+color: red
+color: rgb(255, 0, 0)
+background-color: blue
```
gmta, line 6: In Chrome, these values are `rgb(255, 0, 0)` and `rgb(0, 0, 255)` instead of `red` and `blue`. Maybe you should resolve these values?

lpas: I don't think this is a ML "problem", if this is a problem. This also happens if I would have used 
```html
    <div data-prop="background-color">
        <span style="background-color:red;"></span>
    </div>
```
I assume the named-colors don't get resolved in `computedStyleMap`

Libraries/LibWeb/MathML/MathMLMiElement.h
lpas: Yes this is the default. Seems like the code base is not particular consistent about this. I'm ok with leaving or removing.

lpas:
> Oh wow, we're really missing that `MathML::AtributeNames::*` huh 😅

Hehe, yes I contemplated If I should add these. In the end I decided against it as they are only used in one file each.

Libraries/LibWeb/MathML/MathMLElement.cpp
```diff
@@ -59,4 +65,60 @@ void MathMLElement::visit_edges(JS::Cell::Visitor& visitor)
     HTMLOrSVGElement::visit_edges(visitor);
 }
 
+bool MathMLElement::is_presentational_hint(FlyString const& name) const
+{
+    return first_is_one_of(name, "dir", "mathcolor", "mathbackground", "mathsize", "displaystyle", "scriptlevel");
```
AtkinsSJ: Regarding `MathML::AtributeNames`, the benefit is that those would be `FlyString`s, meaning that comparing them is much faster.

Libraries/LibWeb/MathML/MathMLElement.cpp
```diff
@@ -59,4 +65,60 @@ void MathMLElement::visit_edges(JS::Cell::Visitor& visitor)
...
+}
+
+void MathMLElement::apply_presentational_hints(GC::Ref<CSS::CascadedProperties> cascaded_properties) const
+{
+    for_each_attribute([&](auto& name, auto& value) {
+        if (name == "dir") {
+            // https://w3c.github.io/mathml-core/#attributes-common-to-html-and-mathml-elements
+            // The dir attribute, if present, must be an ASCII case-insensitive match to ltr or rtl. In that case, the user agent is expected to treat the attribute as a presentational hint setting the
```
AtkinsSJ: Nit: We wrap long comments at 120 columns.

Libraries/LibWeb/MathML/MathMLElement.cpp
```diff
@@ -59,4 +65,60 @@ void MathMLElement::visit_edges(JS::Cell::Visitor& visitor)
...
+            if (auto parsed_value = parse_css_value(CSS::Parser::ParsingParams { document() }, value, CSS::PropertyID::Color))
+                cascaded_properties->set_property_from_presentational_hint(CSS::PropertyID::Color, parsed_value.release_nonnull());
+        } else if (name == "mathbackground") {
+            if (auto parsed_value = parse_css_value(CSS::Parser::ParsingParams { document() }, value, CSS::PropertyID::BackgroundColor))
+                cascaded_properties->set_property_from_presentational_hint(CSS::PropertyID::BackgroundColor, parsed_value.release_nonnull());
+        } else if (name == "mathsize") {
+            // https://w3c.github.io/mathml-core/#dfn-mathsize
+            // The mathsize attribute, if present, must have a value that is a valid <length-percentage>.
+            // In that case, the user agent is expected to treat the attribute as a presentational hint setting the element's font-size property to the corresponding value.
+            if (auto parsed_value = HTML::parse_dimension_value(value))
```
AtkinsSJ, line 103: I don't think `HTML::parse_dimension_value()` is quite right, but we don't have anything more suitable. I've made a note to expose a "parse a specific CSS value type" function.

Libraries/LibWeb/MathML/MathMLElement.cpp
```diff
@@ -59,4 +65,60 @@ void MathMLElement::visit_edges(JS::Cell::Visitor& visitor)
...
+                cascaded_properties->set_property_from_presentational_hint(CSS::PropertyID::FontSize, parsed_value.release_nonnull());
+        } else if (name == "displaystyle") {
+            // https://w3c.github.io/mathml-core/#dfn-displaystyle
+            // The displaystyle attribute, if present, must have a value that is a boolean. In that case, the user agent is expected to treat the attribute as a presentational hint setting the
+            // element's math-style property to the corresponding value.
```
AtkinsSJ: The spec continues:
> More precisely, an [ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive) match to true is mapped to normal while an [ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive) match to false is mapped to compact.

I think that'd be useful to include here, to explain why we map these values.

Tests/LibWeb/Text/expected/MathML/presentational_hints.txt
AtkinsSJ: Yeah this is a separate problem where we're not computing the value of colours properly. It's not related to this PR.

AtkinsSJ (changes requested):
Nice to see some MathML stuff happening.

Libraries/LibWeb/MathML/AttributeNames.h
```diff
@@ -0,0 +1,31 @@
...
+    __ENUMERATE_MATHML_ATTRIBUTE(dir, "dir")                       \
+    __ENUMERATE_MATHML_ATTRIBUTE(displaystyle, "displaystyle")     \
+    __ENUMERATE_MATHML_ATTRIBUTE(mathbackground, "mathbackground") \
+    __ENUMERATE_MATHML_ATTRIBUTE(mathcolor, "mathcolor")           \
+    __ENUMERATE_MATHML_ATTRIBUTE(mathsize, "mathsize")             \
+    __ENUMERATE_MATHML_ATTRIBUTE(scriptlevel, "scriptlevel")       \
+    __ENUMERATE_MATHML_ATTRIBUTE(mathvariant, "mathvariant")       \
+    __ENUMERATE_MATHML_ATTRIBUTE(width, "width")                   \
+    __ENUMERATE_MATHML_ATTRIBUTE(height, "height")                 \
+    __ENUMERATE_MATHML_ATTRIBUTE(depth, "depth")
```
AtkinsSJ: Nit: Please put these in alphabetical order.

Libraries/LibWeb/Forward.h
```diff
@@ -876,6 +876,8 @@ struct LayoutState;
 namespace Web::MathML {
 
 class MathMLElement;
+class MathMLMspaceElement;
+class MathMLMiElement;
```
gmta, line 879: Please keep these in alphabetical order.

PR #6254, opened 2025-09-20
LibWeb: Cache results of harfbuzz text shaping

awesomekling:
This patch introduces a per-`Gfx::Font` cache for harfbuzz text shaping
results. As long as the same OpenType features are used, we now cache
the results of harfbuzz shaping, saving massive amounts of time during
text layout.
    
As an example, it saves multiple seconds when loading the ECMAScript
specification at <https://tc39.es/ecma262>. Before this change, harfbuzz
shaping was taking up roughly 11% of a profile of loading that page.
The cache brings it down to 1.8%.
    
Note that the cache currently grows unbounded. I've left a FIXME about
adding some limits.

### DIFF
--- a/Libraries/LibGfx/TextLayout.cpp
+++ b/Libraries/LibGfx/TextLayout.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018-2020, Andreas Kling <andreas@ladybird.org>
+ * Copyright (c) 2018-2025, Andreas Kling <andreas@ladybird.org>
  * Copyright (c) 2021, sin-ack <sin-ack@protonmail.com>
  * Copyright (c) 2024-2025, Aliaksandr Kalenik <kalenik.aliaksandr@gmail.com>
  * Copyright (c) 2025, Jelle Raaijmakers <jelle@ladybird.org>
@@ -7,8 +7,8 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
+#include <AK/Utf16String.h>
 #include <AK/Utf16View.h>
-#include <AK/Utf8View.h>
 #include <LibGfx/Point.h>
 #include <LibGfx/TextLayout.h>
 #include <harfbuzz/hb.h>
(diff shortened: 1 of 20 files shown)


Libraries/LibGfx/TextLayout.cpp
```diff
@@ -98,11 +86,40 @@ static hb_buffer_t* setup_text_shaping(UnicodeView const& string, Font const& fo
     return buffer;
 }
 
-template<typename UnicodeView>
-NonnullRefPtr<GlyphRun> shape_text(FloatPoint baseline_start, float letter_spacing, UnicodeView const& string, Font const& font, GlyphRun::TextType text_type, ShapeFeatures const& features)
+NonnullRefPtr<GlyphRun> shape_text(FloatPoint baseline_start, float letter_spacing, Utf16View const& string, Font const& font, GlyphRun::TextType text_type, ShapeFeatures const& features)
 {
-    auto* buffer = setup_text_shaping(string, font, features);
+    auto& shaping_cache = font.shaping_cache();
+    // dbgln("SHAPE '{}' (size {}, cap {})", string, shaping_cache.map.size(), shaping_cache.map.capacity());
```
tcl3: Left over from debugging?

awesomekling: Clearly! 😅

Libraries/LibGfx/TextLayout.cpp
```diff
@@ -55,25 +54,15 @@ Vector<NonnullRefPtr<GlyphRun>> shape_text(FloatPoint baseline_start, UnicodeVie
...
-    } else if constexpr (IsSame<UnicodeView, Utf16View>) {
-        if (string.has_ascii_storage())
-            hb_buffer_add_utf8(buffer, string.ascii_span().data(), string.length_in_code_units(), 0, -1);
-        else
-            hb_buffer_add_utf16(buffer, reinterpret_cast<u16 const*>(string.utf16_span().data()), string.length_in_code_units(), 0, -1);
-    } else {
-        static_assert(DependentFalse<UnicodeView>);
-    }
+    if (string.has_ascii_storage())
+        hb_buffer_add_utf8(buffer, string.ascii_span().data(), string.length_in_code_units(), 0, -1);
```
khaledhosny, line 62: You might want to use [`hb_buffer_add_latin1()`](https://harfbuzz.github.io/harfbuzz-hb-buffer.html#hb-buffer-add-latin1) here.

PR #6255, opened 2025-09-20
LibWeb: Make HTMLScriptElement.src getter resolve to absolute URL

### LINKED ISSUE #6253: HTMLScriptElement.src returns relative path instead of absolute URL, breaking Webpack publicPath detection
### Summary

After commit [f65dca1b5334](https://github.com/LadybirdBrowser/ladybird/commit/f65dca1b5334bde716dc45a0c90659ebe36a7f22) (LibWeb: Extend src attribute of HTMLScriptElement with TrustedTypes), the behavior of HTMLScriptElement.src regressed.

When loading an external script, document.currentScript.src now returns a relative path (e.g. "./app.js") instead of the fully resolved absolute URL. This breaks applications such as [Telegram Web](https://github.com/LadybirdBrowser/ladybird/issues/2679#issuecomment-3315247847) (and any Webpack build), which rely on currentScript.src to compute the runtime publicPath for dynamic chunk loading. The result is malformed chunk URLs and 404 errors.

### Operating system

macOS

### Steps to reproduce

1. Save index.html and app.js side by side
2. Open index.html in Ladybird.
3. Compare with any other browser

### Expected behavior

currentScript.src = "file:///.../app.js"
computed chunkUrl = "file:///.../chunk.1234.js"
Not allowed to load local resource: file:///.../chunk.1234.js

### Actual behavior

currentScript.src = "./app.js"
computed chunkUrl = "./chunk.1234.js"
ResourceLoader: Failed load ... "file:///.../chunk.1234.js"
HTMLScriptElement: Refusing to run script because the element's result is null.

N/A

```html
index.html

<!doctype html>
<meta charset="utf-8">
<script src="./app.js"></script>

app.js

console.log("currentScript.src =", document.currentScript && document.currentScript.src);
const s = document.currentScript;
const publicPath = s && s.src ? s.src.replace(/[^/]+$/, "") : "";
const chunkUrl = publicPath + "chunk.1234.js";
console.log("computed chunkUrl =", chunkUrl);

const el = document.createElement("script");
el.src = chunkUrl;
el.onerror = () => console.log("Chunk load failed:", el.src);
document.head.append(el);
```

```shell
356847.446 WebContent(60680): (js log) "currentScript.src =" "./app.js"
356847.447 WebContent(60680): (js log) "computed chunkUrl =" "./chunk.1234.js"
356847.448 WebContent(60680): ResourceLoader: Failed load of: "file:///Users/pavel/Develop/ladybird/chunk.1234.js", Error: No such file or directory (errno=2), Duration: 1ms
356847.448 WebContent(60680): HTMLScriptElement: Refusing to run script because the element's result is null.
356847.448 WebContent(60680): (js log) "Chunk load failed:" "./chunk.1234.js"
```

_No response_

_No response_

### Contribute a patch?

shlyakpavel:
LibWeb: Make HTMLScriptElement.src getter resolve to absolute URL

The src IDL attribute was previously implemented as an inline getter that returned the raw attribute value. This broke spec semantics and sites like Telegram Web that rely on document.currentScript.src to compute Webpack’s publicPath.

According to the HTML Standard:
https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#reflecting-content-attributes-in-idl-attributes

For URL-reflecting attributes:
  1. If contentAttributeValue is null, then return the empty string.
  2. Let urlString be the result of encoding-parsing-and-serializing a URL given contentAttributeValue, relative to element’s node document.
  3. If urlString is not failure, then return urlString.
  
  This PR implements these steps

  Fixes #6253

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLScriptElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLScriptElement.cpp
@@ -641,6 +641,31 @@ void HTMLScriptElement::unmark_as_parser_inserted(Badge<DOM::Range>)
     m_parser_document = nullptr;
 }
 
+// https://github.com/w3c/trusted-types/blob/0f01aee35c3684e8cb20e7d83445ad65938f545d/spec/index.bs#L1341
+TrustedTypes::TrustedScriptURLOrString HTMLScriptElement::src() const
+{
+    // 1. Let element be the result of running this's get the element.
+
+    // 2. Let contentAttributeValue be the result of running this's get the content attribute.
+    auto maybe_src = get_attribute(AttributeNames::src);
+
+    // 3. If contentAttributeValue is null, then return the empty string.
+    if (!maybe_src.has_value())
+        return Utf16String {};
+
+    auto const& raw = *maybe_src;
+
+    // 4. Let urlString be the result of encoding-parsing-and-serializing a URL given contentAttributeValue, relative to element's node document.
+    auto url_string = document().encoding_parse_and_serialize_url(raw);
+
+    // 5. If urlString is not failure, then return urlString.
+    if (url_string.has_value())
+        return Utf16String::from_utf8_without_validation(*url_string);
+
+    // 6. Return contentAttributeValue, converted to a scalar value string.
+    return Utf16String {};
+}
+
 // https://www.w3.org/TR/trusted-types/#the-text-idl-attribute
 WebIDL::ExceptionOr<void> HTMLScriptElement::set_text(TrustedTypes::TrustedScriptOrString text)
 {
--- a/Libraries/LibWeb/HTML/HTMLScriptElement.h
+++ b/Libraries/LibWeb/HTML/HTMLScriptElement.h
@@ -62,7 +62,7 @@ public:
     TrustedTypes::TrustedScriptOrString text() const { return child_text_content(); }
     WebIDL::ExceptionOr<void> set_text(TrustedTypes::TrustedScriptOrString);
 
-    TrustedTypes::TrustedScriptURLOrString src() const { return Utf16String::from_utf8(get_attribute_value(AttributeNames::src)); }
+    TrustedTypes::TrustedScriptURLOrString src() const;
     WebIDL::ExceptionOr<void> set_src(TrustedTypes::TrustedScriptURLOrString);
 
     Variant<GC::Root<TrustedTypes::TrustedScript>, Utf16String, Empty> text_content() const;
--- a/Tests/LibWeb/Text/expected/htmlscript-src-reflection.txt
+++ b/Tests/LibWeb/Text/expected/htmlscript-src-reflection.txt
@@ -1,0 +1,3 @@
+is_string = true
+has_protocol = true
+ends_with_dummy = true
--- a/Tests/LibWeb/Text/expected/usvstring-url-reflection.txt
+++ b/Tests/LibWeb/Text/expected/usvstring-url-reflection.txt
@@ -11,6 +11,7 @@ input.src final URL path segment: %EF%BF%BDa%EF%BF%BDb%EF%BF%BD
 link.href final URL path segment: %EF%BF%BDa%EF%BF%BDb%EF%BF%BD
 object.codeBase final URL path segment: %EF%BF%BDa%EF%BF%BDb%EF%BF%BD
 object.data final URL path segment: %EF%BF%BDa%EF%BF%BDb%EF%BF%BD
+script.src final URL path segment: %EF%BF%BDa%EF%BF%BDb%EF%BF%BD
 source.src final URL path segment: %EF%BF%BDa%EF%BF%BDb%EF%BF%BD
 track.src final URL path segment: %EF%BF%BDa%EF%BF%BDb%EF%BF%BD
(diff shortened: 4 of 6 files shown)


Libraries/LibWeb/HTML/HTMLScriptElement.cpp
```diff
@@ -641,6 +641,27 @@ void HTMLScriptElement::unmark_as_parser_inserted(Badge<DOM::Range>)
     m_parser_document = nullptr;
 }
 
+// https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#reflecting-content-attributes-in-idl-attributes
+TrustedTypes::TrustedScriptURLOrString HTMLScriptElement::src() const
```
shannonbooth, line 645: The regression was due to losing the URL extended attribute from the IDL in f65dca1b5334bde716dc45a0c90659ebe36a7f22. Since that was originally added as a non standard attribute, this is now specified as ReflectURL in the HTML spec here: https://html.spec.whatwg.org/multipage/scripting.html#the-script-element:xattr-reflecturl

Ideally, the fix here would be just adding back that extended attribute. However, the IDL from the trusted types spec isn't updated there, they probably need to do an update. I am also not sure if our IDL generator supports that reflection since the trusted types type is no longer just a USVString

shlyakpavel: I don't think our IDL generator supports that, at least I get some errors locally trying to make it Reflect URL. What should I do to make this PR mergable? Fix IDL generator and move it there?

shannonbooth: If its not a easy fix in IDL generator since this is a bit of a special case I think it should be fine to leave a bit of an explanation comment about this manual implementation of the IDL attribute

tete17:
Should you also revert the unit tests changes i made in https://github.com/LadybirdBrowser/ladybird/pull/6058/commits/56ded2b9cc05cc820ba4a870aeb0df952eff5f5c ?

Libraries/LibWeb/HTML/HTMLScriptElement.cpp
shlyakpavel: I added the comment

tete17: To add to this I think we should also open an issue in https://github.com/w3c/trusted-types reflecting that their partial interface in https://github.com/w3c/trusted-types/blob/fbd1144e4c1978b7927d93b763b16fe8a569a657/spec/index.bs#L1283 should include the `ReflectURL` attribute

tete17:
It all looks good from my side apart from the mention we probably want an issue upstream in the trusted types spec.

I am no maintainer so I can't approve nor merge. Maybe @gmta or @AtkinsSJ can help out

shlyakpavel:
This actually also fixes vk.com crash

Libraries/LibWeb/HTML/HTMLScriptElement.cpp
```diff
@@ -641,6 +641,28 @@ void HTMLScriptElement::unmark_as_parser_inserted(Badge<DOM::Range>)
...
+
+    auto raw = get_attribute_value(AttributeNames::src);
+
+    // 2. Let urlString be the result of encoding-parsing-and-serializing a URL given
+    //    contentAttributeValue, relative to element’s node document.
+    auto url_string = document().encoding_parse_and_serialize_url(raw);
+
+    // 3. If urlString is not failure, then return urlString.
+    if (url_string.has_value())
+        return Utf16String::from_utf8(*url_string);
```
gmta: You can use `::from_utf8_without_validation` here since the UTF-8 source (`encoding_parse_and_serialize_url`) should be trusted / validated already.

Libraries/LibWeb/HTML/HTMLScriptElement.cpp
```diff
@@ -641,6 +641,28 @@ void HTMLScriptElement::unmark_as_parser_inserted(Badge<DOM::Range>)
...
 
+// NOTE: Manual [ReflectURL] until IDL generator supports TrustedScriptURLOrString.
+// https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#reflecting-content-attributes-in-idl-attributes
+TrustedTypes::TrustedScriptURLOrString HTMLScriptElement::src() const
+{
+    // 1. If contentAttributeValue is null, then return the empty string.
+    if (!has_attribute(AttributeNames::src))
+        return Utf16String {};
+
+    auto raw = get_attribute_value(AttributeNames::src);
```
gmta: You can prevent doing two lookups by calling into `get_attribute` or `get_attribute_ns`, which both return `Optional<String>`.

Libraries/LibWeb/HTML/HTMLScriptElement.cpp
gmta: @shlyakpavel Can you add the link to the raised upstream issue to a comment here or in the IDL?

shlyakpavel:
The upstream issue was resolved, I need to update this PR. Do not merge.

shlyakpavel:
I updated the comments to follow the fixed spec, and updated the spec link. CI is failing for the spec, so I could only add a GitHub permalink, I hope it's fine for now. Ready for review.

PR #6256, opened 2025-09-21
LibWeb: Derive implicit aria-level for h1–h6 from tag name

Dowsley:
LibWeb: HTMLHeadingElement: Derive implicit aria-level for h1–h6 from tag name

Implement the HTML-ARIA mapping for headings: if aria-level is present, use it; otherwise derive the level from the tag name (h1 -> "1" … h6 -> "6"). See https://www.w3.org/TR/html-aria/#el-h1-h6

Implementation details:
- Prefer the `aria-level` attribute when present.
- If not present, read `local_name()`, take the second char (the number), build a string and return it.

This assumes `local_name()`, at that point in time, will always return a valid 2-character string between h1 or h6.

Verification:
```
bin/test-web --test-path /Users/dowsley/Code/ladybird/Tests/LibWeb --filter "*wpt-import/css/selectors/heading.html*"
bin/test-web --test-path /Users/dowsley/Code/ladybird/Tests/LibWeb --filter "*wpt-import/html/dom/aria-attribute-reflection.html*"
```

Both passed.


shlyakpavel:
Looks good to me

Dowsley:
Wondering how to make those last 3 sanitizer workflows run. Does it need manual approval from the maintainers?

AtkinsSJ:
> Wondering how to make those last 3 sanitizer workflows run. Does it need manual approval from the maintainers?

Yes, and it will if you update it too. At least for your first PR. Sorry this has been overlooked - please post in #code-review in Discord if your PR isn't getting attention.

Dowsley:
No problem, thanks. I will join the discord meanwhile.

I see the Lint Code check failed on code I did not change. Looks flaky. Should I rebase?

Libraries/LibWeb/HTML/HTMLHeadingElement.h
```diff
@@ -26,8 +26,12 @@ class HTMLHeadingElement final : public HTMLElement {
 
     virtual Optional<String> aria_level() const override
     {
-        // TODO: aria-level = the number in the element's tag name
-        return get_attribute("aria-level"_fly_string);
+        if (auto const attr = get_attribute("aria-level"_fly_string); attr.has_value())
```
AtkinsSJ: We create FlyStrings for standard attribute names already. For this, it'd be `ARIA::AttributeNames::aria_level`, in `LibWeb/ARIA/AttributeNames.h`.

Libraries/LibWeb/HTML/HTMLHeadingElement.h
```diff
@@ -26,8 +26,12 @@ class HTMLHeadingElement final : public HTMLElement {
...
+            return attr;
+
+        // Implicit defaults to the number in the element's tag name.
+        auto const digit_view = local_name().bytes_as_string_view().substring_view(1, 1);
+        return String::from_utf8_without_validation(digit_view.bytes());
```
AtkinsSJ: You don't need to limit the substring to being 1 byte long - just skipping the first byte is enough, and would make it a bit more future proof in case HTML decides to add further h levels.

Also, you can write this a bit more directly. Something like:
```c++
MUST(local_name().to_string().substring_from_byte_offset(1));
```
(Working with byte counts is a bit sketchy in general, but should be fine here as we know the first character is `h` so it's a single byte.)

(You might be able to use `substring_from_byte_offset_with_shared_superstring()` but I'm not 100% confident on the FlyString->String conversion still pointing to the same data.)

AtkinsSJ (changes requested):
Could you add a test for this? I've tried looking at WPT and the only tests there are manual ones, so I think you'll need to write your own.

> I see the Lint Code check failed on code I did not change. Looks flaky. Should I rebase?

Yes please, that was fixed a little while ago, but apparently after your submitted this PR.

Libraries/LibWeb/HTML/HTMLHeadingElement.h
Dowsley: I followed through with `MUST(local_name().to_string().substring_from_byte_offset(1))`, thanks.

Libraries/LibWeb/HTML/HTMLHeadingElement.h
Dowsley: Using it now, thanks for the feedback

Dowsley:
Addressed the requested changes and added a new test for validation.

Some other changes had to be made to make the aria label level accessible to the test. Updated the description with details.

Libraries/LibWeb/Internals/Internals.cpp
```diff
@@ -5,6 +5,9 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
+#include <LibWeb/ARIA/AriaData.h>
+#include <LibWeb/ARIA/StateAndProperties.h>
+
 #include <AK/JsonObject.h>
```
AtkinsSJ: Can you make this as one sorted block? I'm guessing your IDE does this, I know mine likes to mess with includes. :sweat_smile:

Dowsley: Done! Yup it was CLion. Slipped through the cracks!

PR #2806, opened 2024-12-06
LibWeb: Layout standalone SVG document with specified dimensions

manuel-za:
Fixes #2668 

Before:  standalone SVG documents were stretched to fit the agent
viewport, ignoring specified width / height attributes.

After this: SVG documents with specified width/height attributes are
laid out respecting these attributes.

Regarding layout of standalone SVG documents, this brings Ladybird to:
- parity with Chromium (131.0.6778.85)
- almost parity with Firefox (129.0.2) -- difference not material

PR #6258, opened 2025-09-21
LibWeb: Do not stomp over content size of SVGs

InvalidUsernameException:
When displaying SVGs inside a navigable directly, we want to display them in the size specified by their width and height attributes instead of stretching them to the viewport as layout normally would.

However, when doing so, we need to actually check that the SVG we are laying out is indeed directly inside the navigable. Otherwise we just blindly overwrite whatever content sizes have been calculated by layout code for e.g. SVG elements inlined somewhere in an HTML document.

Due to the way this was written originally, the bug was not very noticable. The code overwrote the content width/height with the computed width/height, which was often still correct in the sense that those two had the same value. However, content size may also be the result of {min,max}-{width,height} constraints, which can make it differ from the computed values.

This bug was likely a regression introduced in #2806.

| Before | After |
|--------|--------|
| min-max-size-constraints before | min-max-size-constraints after |

### DIFF
--- a/Libraries/LibWeb/Layout/SVGFormattingContext.cpp
+++ b/Libraries/LibWeb/Layout/SVGFormattingContext.cpp
@@ -179,7 +179,8 @@ void SVGFormattingContext::run(AvailableSpace const& available_space)
 
     auto& svg_box_state = m_state.get_mutable(context_box());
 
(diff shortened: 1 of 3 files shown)


AtkinsSJ (approved):
Great! Thanks for the thorough test, too.

PR #6262, opened 2025-09-21
LibWeb: Implement SVGFEColorMatrixElement

shlyakpavel:
It works well enough to run the demo https://kazzkiq.github.io/svg-color-filter/
End goal: fix some issues with Telegram Web ui #2679


shlyakpavel:
Alternatively, we could probably implement proper [SVGNumberList](https://svgwg.org/svg2-draft/types.html#InterfaceSVGNumberList) and write less code inside SVGFilterElement::gfx_filter
I'm not very happy about the idea of getting it into the IDL generator, but that should be done eventually.
What do you think?

Libraries/LibWeb/SVG/SVGFilterElement.cpp
```diff
@@ -170,6 +173,124 @@ Optional<Gfx::Filter> SVGFilterElement::gfx_filter(Layout::NodeWithStyle const&
...
+                        }
+                        // If there is no input (SourceGraphic), skip producing a filter,
+                        // so the pipeline defaults to painting the source as-is.
+                    } else {
+                        // Skia's color matrix uses 0..255 range for the bias (5th column).
+                        matrix[4] *= 255.0f;
+                        matrix[9] *= 255.0f;
+                        matrix[14] *= 255.0f;
+                        matrix[19] *= 255.0f;
+                        root_filter = Gfx::Filter::color_matrix(matrix, input);
```
konradekk: this should probably go deeper – no need for `LibWeb` to know about Skia being the backend.

probably the identity check and replacement could also be done deeper, as this is a general rule, not strictly related to SVG filtering?

Libraries/LibWeb/SVG/SVGFilterElement.cpp
```diff
@@ -170,6 +173,124 @@ Optional<Gfx::Filter> SVGFilterElement::gfx_filter(Layout::NodeWithStyle const&
...
+                update_result_map(*colormatrix_primitive);
+            } else if (equals_ci(type_value, "luminanceToAlpha"sv)) {
+                // values ignored; convert luminance to alpha and zero RGB.
+                float matrix[20] = {
+                    0, 0, 0, 0, 0,
+                    0, 0, 0, 0, 0,
+                    0, 0, 0, 0, 0,
+                    0.2126f, 0.7152f, 0.0722f, 0, 0
+                };
+                root_filter = Gfx::Filter::color_matrix(matrix, input);
```
konradekk, line 216: (consider creating separate `Gfx::Filter` for that? but probably not strictly necessary; still could be `static constexpr`?)

Libraries/LibWeb/SVG/SVGFilterElement.cpp
```diff
@@ -170,6 +173,124 @@ Optional<Gfx::Filter> SVGFilterElement::gfx_filter(Layout::NodeWithStyle const&
...
+                update_result_map(*colormatrix_primitive);
+            } else {
+                // Unknown type, treat as identity pass-through.
+                float identity[20] = {
+                    1, 0, 0, 0, 0,
+                    0, 1, 0, 0, 0,
+                    0, 0, 1, 0, 0,
+                    0, 0, 0, 1, 0
+                };
```
konradekk: isn’t it defined somewhere already? you could extract it (and the usage above, which is declared as `static constexpr`) and use it instead? same for the `matrix` just above?

Libraries/LibWeb/SVG/SVGFilterElement.cpp
shlyakpavel: Sure, there was no need to do it here, it's already done elsewhere

Libraries/LibWeb/SVG/SVGFilterElement.cpp
shlyakpavel: The point of it being here is it being part of the spec
https://www.w3.org/TR/2000/CR-SVG-20000802/filters.html
> For type="luminanceToAlpha", values is not applicable. A luminanceToAlpha operation is equivalent to the following matrix operation:

If we put it elsewhere, it might get eventually modified to reflect some color space stuff etc. I would personally prefer to keep it in-place

Libraries/LibWeb/SVG/SVGFilterElement.cpp
shlyakpavel: It turns out that the whole thing is safe to remove, I just dropped it

shlyakpavel:
I updated the code, cleaned out unnecessary transformations and inline matrixes, and added some tests

shlyakpavel:
Very fun fact about SVG filters in Ladybird: we do not limit they bounds
Libraries/LibWeb/Painting/SVGSVGPaintable.cpp:61
This way, with some filters (I succeeded setting alpha bias (matrix value 20th), one can change colors outside the filtered content
There's not straighforward solution, as some filters (gaussian blur) seem to modify things outside of SVG as part of their normal behavior 
This is a separate issue, outside this PR scope

Libraries/LibWeb/CMakeLists.txt
```diff
@@ -906,6 +906,7 @@ set(SOURCES
     SVG/SVGDescElement.cpp
     SVG/SVGElement.cpp
     SVG/SVGEllipseElement.cpp
+    SVG/SVGFEColorMatrixElement.cpp
```
gmta: Please keep this list in alphabetical order

Libraries/LibWeb/SVG/SVGFEColorMatrixElement.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+    u16 enum_value = 0; // SVG_FECOLORMATRIX_TYPE_UNKNOWN
+    if (type_attribute.is_empty() || type_attribute.equals_ignoring_ascii_case("matrix"sv))
+        enum_value = 1; // SVG_FECOLORMATRIX_TYPE_MATRIX
+    else if (type_attribute.equals_ignoring_ascii_case("saturate"sv))
+        enum_value = 2; // SVG_FECOLORMATRIX_TYPE_SATURATE
+    else if (type_attribute.equals_ignoring_ascii_case("hueRotate"sv))
+        enum_value = 3; // SVG_FECOLORMATRIX_TYPE_HUEROTATE
+    else if (type_attribute.equals_ignoring_ascii_case("luminanceToAlpha"sv))
+        enum_value = 4; // SVG_FECOLORMATRIX_TYPE_LUMINANCETOALPHA
```
gmta: We usually take these constants from the spec and add them to the header file, so you don't have to use literals here.

Libraries/LibWeb/SVG/SVGFilterElement.cpp
```diff
@@ -170,6 +173,87 @@ Optional<Gfx::Filter> SVGFilterElement::gfx_filter(Layout::NodeWithStyle const&
...
+        } else if (auto* colormatrix_primitive = as_if<SVGFEColorMatrixElement>(node)) {
+            auto in_attr = colormatrix_primitive->in1()->base_val();
+            auto input = resolve_input_filter(in_attr);
+
+            auto type_value = colormatrix_primitive->attribute(AttributeNames::type).value_or(String {});
+            auto values_value = colormatrix_primitive->attribute(AttributeNames::values).value_or(String {});
+
+            auto equals_ci = [](StringView a, StringView b) {
+                return AK::StringUtils::equals_ignoring_ascii_case(a, b);
+            };
```
gmta: This helper seems a bit useless, you can just call `type_value.equals_ignoring_ascii_case(StringView)`?

Tests/LibWeb/Ref/expected/svg-fecolormatrix-huerotate-ref.html
gmta, line 1: These are reference tests, not screenshot tests. Screenshot tests have a prerendered PNG assigned to them. They're easy to confuse since they are technically identical.

shlyakpavel: Yeah sure
I wanted to implement them as PNGs first, but decided to make them reference tests later and I forgot to move them. I'll fix that

PR #6267, opened 2025-09-22
LibWeb: Improved parsing of animation properties

Calme1709:
This PR improves our parsing and serialization of the `animation` property and it's longhands - see individual commits.

Gains us ~50 WPT tests.

There is still work to be done to implement actually triggering multiple animations from a single `animation` property.

### DIFF
--- a/Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
+++ b/Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
@@ -450,6 +450,18 @@ Parser::ParseErrorOr<NonnullRefPtr<StyleValue const>> Parser::parse_css_value(Pr
         if (auto parsed_value = parse_aspect_ratio_value(tokens); parsed_value && !tokens.has_next_token())
             return parsed_value.release_nonnull();
         return ParseError::SyntaxError;
+    case PropertyID::AnimationComposition:
+    case PropertyID::AnimationDelay:
+    case PropertyID::AnimationDirection:
+    case PropertyID::AnimationDuration:
+    case PropertyID::AnimationFillMode:
+    case PropertyID::AnimationIterationCount:
+    case PropertyID::AnimationName:
+    case PropertyID::AnimationPlayState:
+    case PropertyID::AnimationTimingFunction:
+        if (auto parsed_value = parse_simple_comma_separated_value_list(property_id, tokens); parsed_value && !tokens.has_next_token())
+            return parsed_value.release_nonnull();
+        return ParseError::SyntaxError;
     case PropertyID::BackdropFilter:
     case PropertyID::Filter:
         if (auto parsed_value = parse_filter_value_list_value(tokens); parsed_value && !tokens.has_next_token())
--- a/Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
+++ b/Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
@@ -450,6 +450,10 @@ Parser::ParseErrorOr<NonnullRefPtr<StyleValue const>> Parser::parse_css_value(Pr
         if (auto parsed_value = parse_aspect_ratio_value(tokens); parsed_value && !tokens.has_next_token())
             return parsed_value.release_nonnull();
         return ParseError::SyntaxError;
+    case PropertyID::Animation:
+        if (auto parsed_value = parse_animation_value(tokens); parsed_value && !tokens.has_next_token())
+            return parsed_value.release_nonnull();
+        return ParseError::SyntaxError;
     case PropertyID::AnimationComposition:
     case PropertyID::AnimationDelay:
     case PropertyID::AnimationDirection:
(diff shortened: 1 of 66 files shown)


Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -3156,9 +3156,31 @@ static CSSPixels snap_a_length_as_a_border_width(double device_pixels_per_css_pi
...
 {
     switch (property_id) {
+    case PropertyID::AnimationDelay:
+        return compute_animation_delay(specified_value, computation_context);
+    case PropertyID::AnimationDuration:
+        return compute_animation_duration(specified_value, computation_context);
+    case PropertyID::AnimationIterationCount:
+        return compute_animation_iteration_count(specified_value, computation_context);
+    case PropertyID::AnimationName:
+        return compute_animation_name(specified_value, computation_context);
```
AtkinsSJ: I'm a little bit nervous about this: We have 267 longhands and counting. For most of them, it seems like some general-purpose processing would achieve the same result with a lot less code. Things like converting calculations to literal dimensions, and converting other dimensions into their canonical unit. That's kind of what `absolutized()` did before.

Of the animation properties handled here, only `animation-name` has special behaviour, converting strings into custom-idents.

I'd propose putting as much logic as possible into the `default` case here, and saving the bespoke handling for properties that really need it.

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -3190,6 +3212,84 @@ NonnullRefPtr<StyleValue const> StyleComputer::compute_value_of_property(Propert
...
+                return entry;
+
+            return CustomIdentStyleValue::create(entry->as_string().string_value());
+        }
+
+        VERIFY_NOT_REACHED();
+    });
+}
+
+NonnullRefPtr<StyleValue const> StyleComputer::compute_animation_iteration_count(NonnullRefPtr<StyleValue const> const& specified_value, PropertyValueComputationContext const& computation_context)
```
AtkinsSJ: If we do keep this, it's missing the spec comments the others have.

Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
```diff
@@ -1229,6 +1233,79 @@ RefPtr<StyleValue const> Parser::parse_aspect_ratio_value(TokenStream<ComponentV
...
+                tokens.discard_a_token();
+            else
+                return {};
+        }
+    } while (tokens.has_next_token());
+
+    transaction.commit();
+
+    // FIXME: This is for compatibility with parse_comma_separated_value_list(), which returns a single value directly
+    //        instead of a list if there's only one, it would be nicer if we always returned a list.
```
AtkinsSJ, line 1291: At the time it made sense to me to avoid keeping things in a Vector if there was only one. But it does make the logic more complicated both on the parsing side and the processing side. :sweat_smile: So yeah if you want to make list-valued properties always use `StyleValueList`, that's fine by me.

Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
```diff
@@ -1229,6 +1233,79 @@ RefPtr<StyleValue const> Parser::parse_aspect_ratio_value(TokenStream<ComponentV
     return nullptr;
 }
 
+// https://drafts.csswg.org/css-animations-1/#animation
+RefPtr<StyleValue const> Parser::parse_animation_value(TokenStream<ComponentValue>& tokens)
```
AtkinsSJ, line 1237: Just want to say I like the way you've written this parsing code. I've been meaning to introduce a generic way of parsing ["coordinated list" shorthands](https://drafts.csswg.org/css-values-4/#linked-properties) like this, and this method could easily be adapted for that - just pass in the shorthand's `PropertyID` and look up its `longhand_ids`. That would affect `animation`, `background`, `mask` and `transition`, to my knowledge.

Though chances are there are unique quirks to most of them that would make it harder. (`background` definitely has them.) This is why we can't have nice things. :cry:

Libraries/LibWeb/CSS/StyleValues/ShorthandStyleValue.cpp
```diff
@@ -141,6 +141,74 @@ String ShorthandStyleValue::to_string(SerializationMode mode) const
...
+                    if (longhand_id == PropertyID::AnimationDuration && !get_longhand_as_vector(PropertyID::AnimationDelay)[i]->equals(property_initial_value(PropertyID::AnimationDelay)))
+                        return true;
+
+                    // https://drafts.csswg.org/css-animations-1/#animation
+                    // Furthermore, when serializing, default values of other properties must be output in at least the
+                    // cases necessary to distinguish an animation-name that could be a value of another property
+                    if (longhand_id == PropertyID::AnimationTimingFunction && first_is_one_of(animation_name, "linear"sv, "ease"sv, "ease-in"sv, "ease-out"sv, "ease-in-out"sv, "step-start"sv, "step-end"sv))
+                        return true;
+
+                    if (longhand_id == PropertyID::AnimationDirection && first_is_one_of(animation_name, "normal"sv, "reverse"sv, "alternate"sv, "alternate-reverse"sv))
```
AtkinsSJ: Doesn't need doing in this PR, but: if we generated `string -> enum` functions for each CSS enum, this could check `string_to_animation_direction(animation_name).has_value()` and avoid duplicating the list of possible values.

(I originally commented on the AnimationTimingFunction line, but detecting easing function names is a lot less trivial. :sweat_smile:)

Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
Calme1709: I did actually start off with this as a generic function but stopped since `animation` is one of those properties which has a unique quirk: once we support them we should reset but not parse values for `animation-range-start` and `animation-range-end` in a similar way to how `background` resets `background-image`

Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
Calme1709: Will add it to the list - I haven't looked closely but there are atleast a couple places that don't yet handle `StyleValueList`s when they should (animations being an obvious one)

Libraries/LibWeb/CSS/StyleComputer.cpp
Calme1709: My thinking behind having individual functions is that computation logic is defined in the spec by property rather than by type (StyleValue class) but I think it is definitely reasonable to have a default by type and only override that where necessary.

I have removed the logic for `animation-{delay,duration,iteration-count}` since these values aren't actually used yet and will open a new PR to implement default logic.

Libraries/LibWeb/CSS/StyleComputer.cpp
Calme1709: Removed.

Libraries/LibWeb/CSS/StyleValues/ShorthandStyleValue.cpp
Calme1709: I have done this by stringing together `keyword_from_string` and `keyword_to_*`, I also realized that my initial approach was case-sensitive which this fixes.

AnimationTimingFunction has also been updated to be case-insensitive.

Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
AtkinsSJ: They officially made this a thing at some point: [reset-only subproperties](https://drafts.csswg.org/css-cascade-5/#reset-only-sub-property). I'm thinking that could go in the JSON and then StyleComputer could handle that automatically, instead of needing to put it in the ShorthandStyleValue during parsing. Actually I'll get on that now.

Calme1709: You may have already thought this through but it's actually quite useful to have them in the ShorthandStyleValue as they affect serialization of that shorthand (i.e. by making serialization invalid if they are set to anything other than their initial value).

AtkinsSJ: Oh, thanks! I wasn't aware of that, and it also seems to not be mentioned in the specs - though WPT does expect it. Spec issue: https://github.com/w3c/csswg-drafts/issues/12846

tcl3 (approved):
LGTM. Turns out this isn't affected by #6287.

PR #6278, opened 2025-09-23
LibWasm: Moar proposals

alimpfard:
We've been a little behind on these, so let's catch up to the finished proposals:
- [Tail call](https://github.com/WebAssembly/tail-call)
- [Extended Constant Expressions](https://github.com/WebAssembly/extended-const)
- [Relaxed SIMD](https://github.com/WebAssembly/relaxed-simd)
- (Partial) [Memory64](https://github.com/WebAssembly/memory64)

Proposals that will definitely not be done in this PR:
- [Garbage collection](https://github.com/WebAssembly/gc) (ever so slightly stubbed a few types needed for WPT)
- [JS String Builtins](https://github.com/WebAssembly/js-string-builtins)

### DIFF
--- a/Libraries/LibWasm/Types.h
+++ b/Libraries/LibWasm/Types.h
@@ -169,6 +169,7 @@ public:
         FunctionReference,
         ExternReference,
         ExceptionReference,
+        UnsupportedHeapReference, // Stub for wasm-gc proposal's reference types.
     };
 
     explicit ValueType(Kind kind)
@@ -178,7 +179,7 @@ public:
 
     bool operator==(ValueType const&) const = default;
 
-    auto is_reference() const { return m_kind == ExternReference || m_kind == FunctionReference; }
+    auto is_reference() const { return m_kind == ExternReference || m_kind == FunctionReference || m_kind == UnsupportedHeapReference; }
     auto is_vector() const { return m_kind == V128; }
     auto is_numeric() const { return !is_reference() && !is_vector(); }
     auto kind() const { return m_kind; }
@@ -204,6 +205,8 @@ public:
             return "externref";
         case ExceptionReference:
             return "exnref";
+        case UnsupportedHeapReference:
+            return "todo.heapref";
         }
         VERIFY_NOT_REACHED();
     }
--- a/Libraries/LibWasm/AbstractMachine/AbstractMachine.h
+++ b/Libraries/LibWasm/AbstractMachine/AbstractMachine.h
@@ -99,6 +99,10 @@ public:
             // ref.null exnref
             m_value = u128(0, 4);
             break;
+        case ValueType::UnsupportedHeapReference:
+            // ref.null (todo)
+            m_value = u128(0, 5);
+            break;
         }
     }
 
--- a/Libraries/LibWasm/Constants.h
+++ b/Libraries/LibWasm/Constants.h
@@ -19,6 +19,20 @@ static constexpr auto v128_tag = 0x7b;
 static constexpr auto function_reference_tag = 0x70;
 static constexpr auto extern_reference_tag = 0x6f;
 
+// wasm-gc references
+static constexpr auto array_reference_tag = 0x6a;
+static constexpr auto struct_reference_tag = 0x6b;
+static constexpr auto i31_reference_tag = 0x6c;
+static constexpr auto eq_reference_tag = 0x6d;
+static constexpr auto any_reference_tag = 0x6e;
+static constexpr auto none_reference_tag = 0x71;
+static constexpr auto noextern_reference_tag = 0x72;
+static constexpr auto nofunc_reference_tag = 0x73;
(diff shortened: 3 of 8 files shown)


shlyakpavel:
Doesn't build here for me on Mac
```
/Users/pavel/Develop/ladybird/Libraries/LibWeb/WebAssembly/Memory.cpp:37:18: error: no matching constructor for initialization of 'Wasm::Limits'
   37 |     Wasm::Limits limits { descriptor.initial, move(descriptor.maximum) };
      |                  ^      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/pavel/Develop/ladybird/Libraries/LibWasm/Types.h:255:14: note: candidate constructor not viable: no known conversion from 'u32' (aka 'unsigned int') to 'AddressType' for 1st argument
  255 |     explicit Limits(AddressType address_type, u64 min, Optional<u64> max = {})
      |              ^      ~~~~~~~~~~~~~~~~~~~~~~~~
/Users/pavel/Develop/ladybird/Libraries/LibWasm/Types.h:253:7: note: candidate constructor (the implicit copy constructor) not viable: requires 1 argument, but 2 were provided
  253 | class Limits {
      |       ^~~~~~
/Users/pavel/Develop/ladybird/Libraries/LibWasm/Types.h:253:7: note: candidate constructor (the implicit move constructor) not viable: requires 1 argument, but 2 were provided
  253 | class Limits {
      |       ^~~~~~
1 error generated.
```

alimpfard:
it won't, for now all the changes are limited to LibWasm itself, integration stuff is TODO since I have no idea what the actual API will be in the end.

Libraries/LibWasm/AbstractMachine/AbstractMachine.cpp
```diff
@@ -76,6 +76,20 @@ Optional<ElementAddress> Store::allocate(ValueType const& type, Vector<Reference
...
+    return address;
+}
+
+Optional<ExceptionAddress> Store::allocate(TagInstance const& tag_instance, Vector<Value> params)
+{
+    ExceptionAddress address { m_exceptions.size() };
+    m_exceptions.append(ExceptionInstance { tag_instance, move(params) });
+    return address;
+}
+
```
Hendiadyoin1, line 92: Any reason why these return optionals?

Libraries/LibWasm/Types.h
```diff
@@ -568,6 +632,17 @@ struct CompiledInstructions {
     bool direct = false; // true if all dispatches contain handler_ptr, otherwise false and all contain instruction_opcode.
 };
 
+template<Enum auto... Vs>
+consteval auto as_ordered()
+{
+    using Type = CommonType<decltype(to_underlying(Vs))...>;
+    Array<Type, sizeof...(Vs)> result;
+    [&]<Type... Is>(IntegerSequence<Type, Is...>) {
+        ([&] { result[to_underlying(Vs)] = Is; }(), ...);
```
Hendiadyoin1: I think the inner part does not have to be an IIFE as assignment is an expression

Hendiadyoin1: (yup, <https://godbolt.org/z/e9h96G6rv>)

Libraries/LibWasm/AbstractMachine/AbstractMachine.cpp
alimpfard: "future handling of oom" basically

Libraries/LibWasm/Types.h
alimpfard: iirc I did that inner IIFE because something was ICE'ing clang, will take a second look.

alimpfard: nah, seems fine, I must've been doing something weird.

Libraries/LibWasm/AbstractMachine/Validator.cpp
```diff
@@ -2030,6 +2049,85 @@ VALIDATE_INSTRUCTION(if_)
...
+    auto tag_index = instruction.arguments().get<TagIndex>();
+    TRY(validate(tag_index));
+
+    auto tag_type = m_context.tags[tag_index.value()];
+    auto& type = m_context.types[tag_type.type().value()];
+    if (!type.results().is_empty())
+        return Errors::invalid("throw type"sv, "empty"sv, type.results());
+
+    for (size_t i = 0; i < type.parameters().size(); ++i)
+        TRY(stack.take(type.parameters()[type.parameters().size() - i - 1]));
```
gmta: `for (auto& parameter : type.parameters().in_reverse()) ...` ?

alimpfard: ah nice, I bet I was in "spec words to text" mode writing that :P

PR #6282, opened 2025-09-23
LibWeb: Handle interpolation of grid track size lists

tcl3:
This allows the `grid-template-rows` and `grid-template-columns` properties to be correctly interpolated.

Gains us 344 WPT subtest passes in `css/css-grid`.


Libraries/LibWeb/CSS/Interpolation.cpp
```diff
@@ -461,6 +461,141 @@ static RefPtr<StyleValue const> interpolate_rotate(DOM::Element& element, Calcul
...
+        default:
+            break;
+        }
+
+        return delta < 0.5f ? from_size : to_size;
+    };
+
+    auto interpolate_grid_size = [&](GridSize const& from_grid_size, GridSize const& to_grid_size) -> GridSize {
+        if (from_grid_size.is_flexible_length() || to_grid_size.is_flexible_length()) {
+            if (!from_grid_size.is_flexible_length() && from_grid_size.css_size().is_auto() && !to_grid_size.is_flexible_length() && to_grid_size.css_size().is_auto())
```
AtkinsSJ: This condition seems a bit wrong - either `from_grid_size` or `to_grid_size` must have `is_flexible_length() == true` because of the line above.

Libraries/LibWeb/CSS/Interpolation.cpp
```diff
@@ -461,6 +461,141 @@ static RefPtr<StyleValue const> interpolate_rotate(DOM::Element& element, Calcul
...
+                return delta < 0.5f ? from_grid_size : to_grid_size;
+
+            if (from_grid_size.is_flexible_length() && to_grid_size.is_flexible_length()) {
+                auto interpolated_flex = interpolate_raw(from_grid_size.flex_factor(), to_grid_size.flex_factor(), delta);
+                return GridSize { Flex { interpolated_flex, FlexUnit::Fr } };
```
AtkinsSJ: Nit: `Flex::make_fr(interpolated_flex)`

Libraries/LibWeb/CSS/Interpolation.cpp
```diff
@@ -461,6 +461,141 @@ static RefPtr<StyleValue const> interpolate_rotate(DOM::Element& element, Calcul
...
+        return result;
+    };
+
+    auto expanded_from = expand_tracks_and_lines(from);
+    auto expanded_to = expand_tracks_and_lines(to);
+
+    if (expanded_from.tracks.size() != expanded_to.tracks.size())
+        return {};
+
+    GridTrackSizeList interpolated_grid_tack_size_list;
```
AtkinsSJ: Typo: `tack`

AtkinsSJ (changes requested):
Review of the actual grid interpolation is a bit superficial because I do *not* understand grid. :sweat_smile:

Libraries/LibWeb/CSS/Interpolation.cpp
tcl3: This condition is clearly never true :sweat_smile: I think this is just the result of me rushing through a refactor and getting it a bit wrong.

My latest push removes this check.

PR #6284, opened 2025-09-23
LibWeb: Clamp interpolated `font-style` angle

tcl3:
This change adds the allowed angle range to the `font-style` property definition. This allows these angles to be clamped after interpolation.

Ideally, the generator should be updated so that we can specify the angle is in degrees. This would allow us to make use of this information during parsing, which we can't do currently because we don't know what the unit is. Using this value for interpolation purposes is fine because the angle has been converted to its canonical unit by this point.

### DIFF
--- a/Libraries/LibWeb/CSS/Parser/Parser.h
+++ b/Libraries/LibWeb/CSS/Parser/Parser.h
@@ -551,7 +551,6 @@ private:
         DescriptorID descriptor;
     };
     enum SpecialContext : u8 {
-        FontStyleObliqueAngle,
         ShadowBlurRadius,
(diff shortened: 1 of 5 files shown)


Calme1709:
I believe this change means we not longer need the functionality introduced in e17d91780d27bcc516717c7765ae508ccb44a95d

tcl3:
> I believe this change means we not longer need the functionality introduced in [e17d917](https://github.com/LadybirdBrowser/ladybird/commit/e17d91780d27bcc516717c7765ae508ccb44a95d)

I've removed this special-casing :^)

PR #6286, opened 2025-09-23
Meta+Tests: Update vcpkg dependencies

### LINKED ISSUE #6205: Some digits in Mastodon posts rendered as unrelated emoji.
In Mastodon posts some digits are displayed as an inline image of an unrelated emoji instead.

Example: On https://mstdn.social/@ElleGray/115192495037437176 the downward arrow emoji is supposed to be the number 5.

image

Reduced test case:
```html
<!DOCTYPE html>
<script>
console.log({ "⤵️": 42 });
</script>
```

In Ladybird this logs `{ "5": 42 }`, in other browsers it logs the object with the emoji unmodified.

This is because string keys inside a JS object are coerced to an integer if they represent one.

https://github.com/LadybirdBrowser/ladybird/blob/7b3465ab55142e7e8b5a8261531ce87960bbceac/Libraries/LibJS/Runtime/PropertyKey.h#L76-L89

Here `string` holds ⤵️. Calling `to_number<u32>()`  on it will eventually call into `fast_float` for the actual number-parsing. `fast_float` however has a bug when converting a wide string (UTF-16 or UTF-32) to an integer, that causes it to parse certain emojis and probably other Unicode characters into a single-digit number when it shouldn't. For the downward arrow, it returns 5, which we then store in the dictionary.

Mastodon uses a JS object with all emojis they know about to perform some replacement of Unicode emojis in the text with an image of the emoji instead. Since the dictionary keys for some of the emojis get corrupted in this way, Mastodon's code now mistakenly thinks that 5 is an emoji that needs to be replaced with an image.

This likely regressed as a side-effect of #5698, 0efa98a57a5a7bee63ed284fc38feccfc5ca9e62 (though I haven't checked). That commit changed the above mentioned code from UTF-8 to UTF-16. Since the bug in `fast_float` is specific to UTF-16/32 and does not appear with UTF-8, it seems logical that this was introduced then.

I have a bugfix prepared that I will upstream shortly. Opening this issue for tracking purposes and so that I have something to reference from the upstream PR.

InvalidUsernameException:
The primary motivation for this is version 8.1.0 of `fast_float`, which contains a change that fixes #6205.

Since the baseline update already caused everything to rebuild, I've also bumped all other dependencies that were a straightforward version change. There is nothing super relevant in the changelogs for these versions as far as I can tell. The most interesting thing might be some security fixes in the curl update, which are probably good to have.

### DIFF
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -216,7 +216,7 @@
     },
     {
       "name": "dirent",
-      "version": "1.25#0"
+      "version": "1.26#0"
     },
     {
       "name": "fast-float",
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -264,7 +264,7 @@
     },
     {
       "name": "libwebp",
-      "version": "1.6.0#0"
+      "version": "1.6.0#1"
     },
     {
       "name": "mman",
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -1,5 +1,5 @@
 {
-  "builtin-baseline": "120deac3062162151622ca4860575a33844ba10b",
+  "builtin-baseline": "b0b3de1b1a0aa4b8f2822460aa7f42f991629b3f",
   "dependencies": [
     {
       "name": "angle",
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -220,7 +220,7 @@
     },
     {
       "name": "fast-float",
-      "version": "8.0.2#0"
+      "version": "8.1.0#0"
     },
     {
       "name": "ffmpeg",
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -224,7 +224,7 @@
     },
     {
       "name": "ffmpeg",
-      "version": "7.1.1#4"
+      "version": "7.1.1#5"
     },
     {
       "name": "fontconfig",
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -212,7 +212,7 @@
     },
     {
       "name": "dbus",
-      "version": "1.16.2#1"
+      "version": "1.16.2#2"
     },
     {
       "name": "dirent",
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -208,7 +208,7 @@
     },
     {
       "name": "curl",
-      "version": "8.15.0#1"
+      "version": "8.16.0#0"
     },
     {
       "name": "dbus",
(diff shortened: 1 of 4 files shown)


InvalidUsernameException:
The GCC job [already passed](https://github.com/LadybirdBrowser/ladybird/actions/runs/18008166610) on an [earlier push that only differs in documentation changes](https://github.com/LadybirdBrowser/ladybird/commit/a75e96efa8c58814d4411fbeb2fc587cc1837823). I'd rather not force-push empty changes until everything is green for this PR because Windows CI takes almost 3 hours here.

Libraries/LibJS/Tests/object-with-unicode-emoji-as-key.js
```diff
@@ -0,0 +1,25 @@
+test("unicode emojis with the low byte corresponding to the value of an ascii digit should not coerce to integers", () => {
```
trflynn89: Let's instead put this test in Tests/AK/TestStringConversions.cpp, so that both UTF-8 and UTF-16 strings can be exercised.

vcpkg.json
```diff
@@ -232,7 +232,7 @@
     },
     {
       "name": "harfbuzz",
-      "version": "10.2.0#0"
+      "version": "11.3.3#0"
```
trflynn89: harfbuzz is currently pinned to version 10 on purpose, let's not update it here. See:
504be366a295e12bb300d2a6bad69f79585a34cf

Libraries/LibJS/Tests/object-with-unicode-emoji-as-key.js
InvalidUsernameException: I can move it to the cpp test, but I am not sure what inputs I would test for UTF-8. The bug is not related to specific code points but to the way these code points are encoded. It happens for no UTF-8-encoded code points, some UTF-16-encoded code points and a lot more UTF-32-encoded code points (see my [tests on the upstream PR](https://github.com/fastfloat/fast_float/pull/325/files#diff-586623fde0745f9ebf21853646442d18c2d589635d575d7d731294e4f2757c59))

Do you want me to just test the same code points for UTF-8 as well anyway?

trflynn89: I was thinking the same inputs, yes. The test already has examples that run the parser over UTF-8 and UTF-16 for the same inputs.

(I realize this was not a UTF-8 issue in fast-float, but if we're gonna add a regression test here, might as well test both encodings that we support)

PR #6291, opened 2025-09-24
LibWeb: Calculate and use bounds for "simple" stacking contexts

kalenikaliaksandr:
Teach the display list executor to derive a bounding rectangle for stacking contexts whose inner commands can all report bounds, that is, most contexts without nested stacking contexts.

This yields a large performance improvement on https://tc39.es/ecma262/ where the display list contains thousands of groups like:
```
PushStackingContext blending=Multiply
    DrawGlyphRun
PopStackingContext
```
Previously, `PushStackingContext` triggered an unbounded `saveLayer()` even when the glyph run lies wholly outside the viewport. With this change, we (1) cull stacking contexts that fall outside the viewport and (2) provide bounds to `saveLayer()` when they are visible.

With this change rendering thread goes from 70% to 1% in profiles of https://tc39.es/ecma262/. Also makes a huge performance difference on Discord.

### DIFF
--- a/Libraries/LibWeb/Painting/DisplayList.cpp
+++ b/Libraries/LibWeb/Painting/DisplayList.cpp
@@ -125,6 +125,28 @@ void DisplayListPlayer::execute_impl(DisplayList& display_list, ScrollStateSnaps
 
     VERIFY(!m_surfaces.is_empty());
 
+    auto translate_command_by_scroll = [&](auto& command, int scroll_frame_id) {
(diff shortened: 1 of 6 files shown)


Psychpsyo:
If the element for the stacking context has paint containment, the bounds could be further restricted to the box's `overflow_clip_edge_rect()` since anything inside it will be clipped to that anyways.
(If I'm understanding this change correctly)

Or it could be restricted to just that, to avoid walking the children. I think it's somewhat safe to assume that this would yield close to the optimal bounds anyways since elements tend to not have huge amounts of blank space inside them.

EDIT: Guess I'm making a PR for this.

PR #6295, opened 2025-09-24
LibWeb: Compute bounds for paint-contained stacking contexts

Psychpsyo:
This makes it so that the bounds for any paint-contained stacking context are not derived from its children, but rather just set to the rectangle that they will be clipped to anyways due to the paint containment. Should make rendering faster on pages that use paint containment.

This is what I meant when I wrote my comment on  #6291.

### DIFF
--- a/Libraries/LibWeb/Painting/DisplayListRecorder.h
+++ b/Libraries/LibWeb/Painting/DisplayListRecorder.h
@@ -111,6 +111,7 @@ public:
         bool isolate;
         StackingContextTransform transform;
         Optional<Gfx::Path> clip_path = {};
+        Optional<Gfx::IntRect> bounding_rect {};
 
         bool has_effect() const { return opacity != 1.0f || compositing_and_blending_operator != Gfx::CompositingAndBlendingOperator::Normal || isolate || clip_path.has_value() || !transform.is_identity(); }
     };
--- a/Libraries/LibWeb/Painting/DisplayList.cpp
+++ b/Libraries/LibWeb/Painting/DisplayList.cpp
@@ -193,7 +193,7 @@ void DisplayListPlayer::execute_impl(DisplayList& display_list, ScrollStateSnaps
 
         if (command.has<PushStackingContext>()) {
             auto& push_stacking_context = command.get<PushStackingContext>();
-            if (push_stacking_context.can_aggregate_children_bounds) {
+            if (push_stacking_context.can_aggregate_children_bounds && !push_stacking_context.bounding_rect.has_value()) {
                 bounding_rect = compute_stacking_context_bounds(push_stacking_context, command_index);
                 push_stacking_context.bounding_rect = bounding_rect;
             }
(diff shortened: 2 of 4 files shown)


kalenikaliaksandr (changes requested):
could you add a test when stacking context with transform is nested into stacking context with paint containment, because I don't think it's going to work correctly with this change.

Psychpsyo:
> could you add a test when stacking context with transform is nested into stacking context with paint containment, because I don't think it's going to work correctly with this change.

What exactly should it be testing for?
That the transformed element shows up at all? That it can't render beyond the paint-contained element's overflow clip edge?
I got this written up, which works with my changes and is expected to render as a 100x100 green square:
```html
<!DOCTYPE html>
<style>
  #outer {
    contain: paint;
    width: fit-content;
  }
  #inner {
    scale: 2;
    width: 100px;
    height: 100px;
    background-color: green;
  }
</style>
<div id="outer">
  <div id="inner"></div>
</div>
```
(both `contain` and `scale` establish a stacking context for `#inner` and `#outer`)

kalenikaliaksandr:
> > could you add a test when stacking context with transform is nested into stacking context with paint containment, because I don't think it's going to work correctly with this change.
> 
> What exactly should it be testing for? That the transformed element shows up at all? That it can't render beyond the paint-contained element's overflow clip edge? I got this written up, which works with my changes and is expected to render as a 100x100 green square:
> 
> ```
> <!DOCTYPE html>
> <style>
>   #outer {
>     contain: paint;
>     width: fit-content;
>   }
>   #inner {
>     scale: 2;
>     width: 100px;
>     height: 100px;
>     background-color: green;
>   }
> </style>
> <div id="outer">
>   <div id="inner"></div>
> </div>
> ```
> 
> (both `contain` and `scale` establish a stacking context for `#inner` and `#outer`)

what's computed bounding box of #outer in this html with your change?

Psychpsyo:
> what's computed bounding box of #outer in this html with your change?

Ah, I was missing a check for `bounding_rect.has_value()` in `PushStackingContextParams::has_effect()`.
Now that that's there, the bounding box of `#outer` is 100x100. (instead of none)
I'm not sure how I would test for that difference though.

kalenikaliaksandr:
thinking about this change made me realize I introduced a regression in the original optimization 😅 https://github.com/LadybirdBrowser/ladybird/pull/6296

Libraries/LibWeb/Painting/DisplayListRecorder.h
```diff
@@ -111,8 +111,9 @@ class WEB_API DisplayListRecorder {
         bool isolate;
         StackingContextTransform transform;
         Optional<Gfx::Path> clip_path = {};
+        Optional<Gfx::IntRect> bounding_rect {};
 
-        bool has_effect() const { return opacity != 1.0f || compositing_and_blending_operator != Gfx::CompositingAndBlendingOperator::Normal || isolate || clip_path.has_value() || !transform.is_identity(); }
+        bool has_effect() const { return opacity != 1.0f || compositing_and_blending_operator != Gfx::CompositingAndBlendingOperator::Normal || isolate || clip_path.has_value() || !transform.is_identity() || bounding_rect.has_value(); }
```
kalenikaliaksandr: we don't need `bounding_rect.has_value()` here, because paint containment doesn't produce any visual effect. otherwise LGTM.

Psychpsyo: Done.

Psychpsyo:
CI failure is in the gamepad API tests, those are unrelated and highly flaky.

kalenikaliaksandr (approved):

PR #6302, opened 2025-09-25
Add build support for RISC-V

evelance:
Enables Ladybird to be built on RISC-V (riscv64) Linux hosts.

Joint work of @Opvolger and me. There were a bunch of small issues in vcpkg that we already fixed upstream ([libvpx](https://github.com/microsoft/vcpkg/pull/47245), [vcpkg-tool-gn](https://github.com/microsoft/vcpkg/pull/47244), [vcpkg-tool-meson](https://github.com/microsoft/vcpkg/pull/47420), [pkgconf](https://github.com/microsoft/vcpkg/pull/47424)) so the vcpkg.json baseline needs to be bumped.

Built and tested with Debian Trixie on Starfive VisionFive 2 and Ubuntu 24.04 on OrangePi RV2.

### DIFF
--- a/Meta/CMake/common_compile_options.cmake
+++ b/Meta/CMake/common_compile_options.cmake
@@ -61,6 +61,11 @@ elseif (ENABLE_CI_BASELINE_CPU)
     elseif (CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64")
         add_cxx_compile_options(-march=x86-64-v3)
     endif()
+elseif (CMAKE_SYSTEM_PROCESSOR STREQUAL "riscv64")
+    # On RISC-V the generic -march=native is not yet supported and both gcc and clang require an explicit
+    # ISA or target string. Unfortunately hardware probing is also neither easy nor reliable at the moment.
+    # For the time being use the defaults for the best compatibility with existing hardware and toolchains.
+    # FIXME: Remove this branch once -march=native is supported.
 else()
     # In all other cases, compile for the native architecture of the host system.
     add_cxx_compile_options(-march=native)


Meta/ladybird.py
```diff
@@ -289,6 +289,8 @@ def configure_build_env(preset: str) -> tuple[Path, Path]:
     os.environ["PATH"] += os.pathsep + str(ladybird_source_dir.joinpath("Toolchain", "Local", "cmake", "bin"))
     os.environ["PATH"] += os.pathsep + vcpkg_root
     os.environ["VCPKG_ROOT"] = vcpkg_root
+    if platform.host_architecture == HostArchitecture.riscv64:
+        os.environ["VCPKG_FORCE_SYSTEM_BINARIES"] = "1"
```
gmta, line 296: Why do you need this? If it's needed, a comment explaining why should at least accompany it for future reference.

Meta/CMake/common_compile_options.cmake
```diff
@@ -61,7 +61,8 @@ elseif (ENABLE_CI_BASELINE_CPU)
     elseif (CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64")
         add_cxx_compile_options(-march=x86-64-v3)
     endif()
-else()
+elseif (NOT CMAKE_SYSTEM_PROCESSOR STREQUAL "riscv64")
+    # RISC-V does not yet support -march=native.
```
gmta: Instead of hijacking the default fallback (`else()`) for this, can you make a separate `elseif` to match riscv64 and put this comment there?

Also, given that `-march=native` is not supported, maybe it _is_ possible to find a common architecture and pass an `-march=..` that would work on modern machines?

Meta/ladybird.py
evelance: This is required for "weird" targets [as vcpkg puts it](https://github.com/microsoft/vcpkg-tool/blob/71538f2694db93da4668782d094768ba74c45991/include/vcpkg/base/message-data.inc.h#L1422). But it doesn't really work and only forces the use of some system binaries. That's why I had to fix vpkg's vendored gn, meson and pkgconf as well as linked in the initial comment of this PR because it doesn't use the system provided ones even with VCPKG_FORCE_SYSTEM_BINARIES set. I haven't really found a satisfying explanation on why those targets are treated differently in the first place either. I'll add a comment to the script.

Meta/CMake/common_compile_options.cmake
evelance: This is a big pain point for RISC-V currently as the release of RVA23 hardware is imminent but right now absolutely no RVA23 compatible hardware exists. So we have a weird zoo or RV64G and RVA22 some with vector some without etc. and the compiler flags are a complete mess.
I don't think putting complex probing logic here right now would be sensible and just using the default options is better until the hardware support and probing situation improves upstream. Likely next year.
Good point, I'll add another comment there.

Meta/CMake/vcpkg/generate_vcpkg_toolchain_variables.cmake
```diff
@@ -50,6 +50,8 @@ if (NOT DEFINED CACHE{VCPKG_TARGET_TRIPLET} AND NOT DEFINED CACHE{VCPKG_HOST_TRI
 
     if(os_platform MATCHES "^(x86_64|AMD64|amd64)$")
       set(arch x64)
+    elseif(os_platform STREQUAL "riscv64")
+      set(arch riscv64)
```
gmta: Just noticed that this line (and the `set(arch x64)` above it) have the wrong indentation; should be 4 spaces instead of 2.

Meta/CMake/common_compile_options.cmake
spholz: For SerenityOS, we currently set `-march=` to a common baseline of RV64GC:
https://github.com/SerenityOS/serenity/blob/bfb4e508d615d80d19a95a1d4c32b2c50c6ab47a/Toolchain/CMake/GNUToolchain.txt.in#L49

Though that should already be the default for GCC/Clang built with standard settings I think. Not sure if some linux distros already use different defaults (except maybe ubuntu, which wants to switch to RVA23 in 25.10). But no distro likely sets the default `-march=` to a value with even less extensions, so not setting it is likely fine.

For SerenityOS we just set it to RV64GC so we can assume that certain extensions are always present, but that shouldn't really matter for ladybird.

LLVM at least seems to support `-mcpu=native` using the linux hwprobe interface:
https://github.com/llvm/llvm-project/blob/ae11c5c2c4d7ae4cba4a8e05f0c7d85b316a2cf0/clang/lib/Driver/ToolChains/Arch/RISCV.cpp#L286. But GCC doesn't, so that doesn't help either.

Meta/CMake/vcpkg/generate_vcpkg_toolchain_variables.cmake
evelance: Should I fix this here? Or should this be done later in a broader cleanup?
Meta/CMake/skia.cmake, GenerateSwiftHeader.cmake, FindSwiftTesting.cmake and InitializeSwift.cmake also use the wrong indentation.
Edit: I'll update this commit to fix it here at least.

evelance: Done.

Meta/CMake/common_compile_options.cmake
evelance: With clang 18 on my OrangePi RV2 `-mcpu=native` results in this error: `unsupported argument 'generic' to option '-mcpu='`

Setting `-march=rv64gc` seems to produce identical code as not setting the flag.

I think not setting any flags would be the safest option for now. Once available, RVA23 should be perfect for Ladybird with mandatory extensions for vector, crypto, compression and bit manipulation.

Meta/ladybird.py
ADKaster: VCPKG_FORCE_SYSTEM_BINARIES is also required for linux arm64 hosts. It's set in a few different places already. namely, the devcontainer feature and in the arm64 CI. https://github.com/LadybirdBrowser/ladybird/blob/2fd424ccb6fc4b826a1feb6ddc307edbaca8cca8/.github/actions/setup/action.yml#L127. I don't remember if it's documented in the build docs for that arch yet. A follow-up to consistently explain how to build on arm64 and risc-v hosts would be useful.

Note that forcing this env variable in the script here means that a rebuild that needs vcpkg to do work will fail if run manually via `ninja`, which might be surprising if it's not documented anywhere in ladybird's docs.

PR #6319, opened 2025-09-26
Browser: Implement newtab page for Ladybird

EmpezoElJuego:
A esthetic tap page with search bar.
Captura de pantalla_20250926_170327

### DIFF
--- a/Base/res/ladybird/about-pages/newtab.html
+++ b/Base/res/ladybird/about-pages/newtab.html
@@ -1,12 +1,105 @@
 <!DOCTYPE html>
-<html lang="en">
-<head>
+<html>
+  <head>
     <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>New Tab</title>
+  </head>
+  <html>
     <style>
-        html {
-            color-scheme: light dark;
+    html,
+    body {
+      height: 100%;
+      padding: 0;
+      margin: 0;
+    }
+
+    @media (prefers-color-scheme: light) {
+        input {
+          background-color: rgb(255, 255, 255);
         }
+        button {
+          background-color: rgb(255, 255, 255);
+        }
+    }
+
+    @media (prefers-color-scheme: dark) {
+        input {
+          background-color: rgb(62, 62, 62);
+        }
+        button {
+          background-color: rgb(62, 62, 62);
+        }
+    }
+
+    body {
+      background-image: url("https://plus.unsplash.com/premium_photo-1673264933048-3bd3f5b86f9d?q=80&w=870&auto=format&fit=crop&ixlib=rb-4.1.0&ixid=M3wxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8fA%3D%3D");
+      background-repeat: no-repeat;
+      background-size: cover;
+      font-family: sans-serif;
+      color: hsl(0, 0%, 96%);
+      font-weight: 400;
+      line-height: 1.5;
+      display: flex;
+      flex-direction: column;
+      align-items: center;
+      justify-content: center;
+    }
+    .switch_button{
+      border-radius: 15px;
+      background-color: rgb(255, 255, 255);
+      color: rgb(116, 116, 116);;
+      border-color: rgba(0, 0, 0, 0);
+    }
+
+    form {
+      margin: 1em;
+    }
+
+
+    input {
+      width: 50vw;
+      height: 3vw;
+      border-radius: 15px;
+      border-color: rgba(0, 0, 0, 0);
+    }
+
+
+    a {
+      display: flex;
+      color: #ffffff;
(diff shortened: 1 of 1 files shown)


rcorsi:
Looks nice! :-)

To save you some time, and because I keep forgetting to do this myself, the commit message needs to be in a specific format. Per the [contribution guide](https://github.com/LadybirdBrowser/ladybird/blob/master/CONTRIBUTING.md#code-submission-policy), please write the commit message in the imperative mood:

- Write the commit message subject line in the imperative mood ("Foo: Change the way dates work", not "Foo: Changed the way dates work").

My two cents would be that you change it to: `Base: Add mountainous newtab page with search to Ladybird`
But please feel free to set it to something better. :-)

rcorsi:
That the "Lint Commit Messages ..."  is repeatedly failing, it is the git commit message that you should amend and update, not the name of this pull request.

trflynn89:
I am quite opposed to the new tab page loading a remote URL. IMO the page should be kept extremely simple at this point in time.

rcorsi:
Yes, and as trflynn89 is saying, lets not open other pull requests.  Do everything with this one.

Maybe you are new to github and/or git.  You should invest a little time to learn more about them. Lots of videos on youtube about the two.  Take your time and double check things.  No hurry.

rcorsi:
@trflynn89, what about if the image would be stored on ladybird.org or somewhere the project controls?  Also what about going to duckduckgo as a search engine?

I'm guessing it is too soon for this type of change, maybe closer to when we go alpha or beta might be more appropriate?

AtkinsSJ:
I'm pretty sure using an Unsplash+ image without paying is against their terms of service. Let's not do that.

> what about if the image would be stored on ladybird.org or somewhere the project controls?

That's still remote. Let's not ping ladybird.org every time you open a new tab.

> Also what about going to duckduckgo as a search engine?

We should honour the user's selected search engine. Personally I don't like having a search box on the new tab when there's already a location bar just above it that does the same thing.

rcorsi:
@EmpezoElJuego have you had a look in the "Ladybird Settings"? Maybe leave the blank newtab alone as developers probably want the least distraction and nothing that could slow them down (for example, due to slow or not working network). Create a different page that fetches an image locally (Ladybird logo) from Base/res and gets the user's search engine preference from the settings.

Base/res/ladybird/about-pages/newtab.html
```diff
@@ -1,12 +1,105 @@
...
+
     </style>
-</head>
+    <body class="night_mode">
+      <a href="https://ladybird.org">
+        <img src="resource://icons/128x128/app-browser-dark.png" height="100px" width="100px"/><h1>Ladybird</h1>
+      </a>
+      <form action="https://duckduckgo.com/">
+        <input id="InputBar" name="q" placeholder="Search the web…" autofocus />
+        <button id="ButtonClick" type="submit"><svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 24 24" fill="none" stroke="#8f8f8f" stroke-width="3" stroke-linecap="round" stroke-linejoin="round" class="lucide lucide-search-icon lucide-search"><path d="m21 21-4.34-4.34"/><circle cx="11" cy="11" r="8"/></svg></button>
```
Psychpsyo, line 100: This SVG seems to come from https://lucide.dev/icons/search and is either MIT or ISC-licensed. (according to https://lucide.dev/license) Either of those require us to include a copy of their license.

You're also not using the classes on it or the IDs on the `<button>` and `<input>` so those can probably be removed.

Euro20179:
> Personally I don't like having a search box on the new tab when there's already a location bar just above it that does the same thing.

I agree, however many users do (and I have personally watched many people) use the central search box. And my understanding is that this browser is supposed to be a "just werks" browser, so I think including a central search box should be the default.

konradekk:
> I agree, however many users do (and I have personally watched many people) use the central search box. And my understanding is that this browser is supposed to be a "just werks" browser, so I think including a central search box should be the default.

just thinking about creating really custom start page, pretty much agreeing with mr. sam (understanding though, that people do not care about things being duplicated): what about sliding the address/search bar to the center of the screen for a new tab? having it very plain (just as html version would be), so users would see it in the middle of the window, but sliding it up through animation as soon as the tab would navigate somewhere? a single bar with clear connection of both presentations.

in a way that’s something possible with custom approach ladybird is taking. of course that’s pretty non-standard and involved (would need to mix native vs. in-house text input? the address bar could be only in-house only? but still would need to be displayed, probably not so easily, in the window chrome?) so just a food for thought!

Euro20179:
> what about sliding the address/search bar to the center of the screen for a new tab? 

I agree with you that the implementation of this would probably be clunky or very involved, but I still think it's a cool idea.

AtkinsSJ:
Hey folks, let's not use a PR for bikeshedding on what a new tab page should look like. :^)

PR #6320, opened 2025-09-27
LibWeb: Disallow ASFs in DOMMatrix constructor

Calme1709:
This PR implements `sibling-count()` and `sibling-index()` tree counting functions as ASFs. The spec doesn't explicitly state these functions are ASFs as it does with others (i.e. `var()`, `attr()`, `if()`) but they seem to match ASFs in functionality.

Still some work to do around resolving these functions on the correct element for element-backed pseudo-elements, as well as handling tree-scoping.

This also includes a small fix to the `DOMMatrix` constructor to avoid regressions in WPT tests.


AtkinsSJ:
Very cool to see these! Both of them are always one `<integer>`, so they shouldn't be ASFs I think, because eg `color: sibling-count()` will get rejected at parse time instead of being invalid at computed value time. I'm saying this without looking at your code or the spec though. 😅

Calme1709:
You are correct, I'm surprised that WPT doesn't test that. I have reduced this PR to just the commit for DOMMatrix and will take another stab at tree counting functions with that in mind.

PR #6328, opened 2025-09-27
Fix brave.com crash when entering the website

luizgfranca:
Currently if you enter brave.com using ladybird and wait a little bit the WebContent process crashes.

Looking into it it seems to be an error in the active element calculation algorithm for the shadow DOM. 

When there is no candidate for active element on ShadowDOM we cannot obtain the root element from it causing a segmentation fault. This does not happen in normal trees because we will have the Document in the worst case.
    
I think that handling the candidate being null does not invalidate the step 3 here because technically if there's no candidate the candidate root is also null, and because of this it will be different from Self (that should always be defined).

The added test reproduces the crash found on the brave website.


Libraries/LibWeb/DOM/DocumentOrShadowRoot.h
```diff
@@ -26,7 +26,7 @@ GC::Ptr<Element> calculate_active_element(T& self)
     candidate = as<Node>(retarget(candidate, &self));
 
     // 3. If candidate's root is not this, then return null.
-    if (&candidate->root() != &self)
+    if (candidate == nullptr || &candidate->root() != &self)
```
gmta, line 29: > Handling the candidate being null does not invalidate the step 3 here because tecnically if there's no candidate the candidate root is also null

Historically, this has not been true for a lot of specs. I'm actually currently working on a fix for this, where `document.focused_area()` will return an actual candidate.

luizgfranca: makes sense, I will close this PR then and let yours take care of this.

Issue #6232, opened 2025-09-18
LibJS: TypedArray#copyWithin with an end index does not work correctly


Lubrsi:
For example:
```js
> new Int32Array([0,54,999,1000,0,0,0,0,0,0,0,0,0,0,0,0]).copyWithin(3,2,4)
[Int32Array]
  buffer: [ArrayBuffer] @ 0x0000000104fc00e8
  length: 16
  byteLength: 64
[ 0, 54, 999, 1000, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 ]
```
However, it should be:
```js
new Int32Array([0,54,999,1000,0,0,0,0,0,0,0,0,0,0,0,0]).copyWithin(3,2,4)
final: 4
[Int32Array]
  buffer: [ArrayBuffer] @ 0x000000010327c0e8
  length: 16
  byteLength: 64
[ 0, 54, 999, 999, 1000, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 ]
```

Reduced from Google Sheets. Regressed by https://github.com/LadybirdBrowser/ladybird/commit/ada198bee0f88c2c03c9c380a4373d80714fd524
cc @awesomekling

vedant-pandey:
I would like to work on this issue.

PR #6331, opened 2025-09-28
LibJS: Fix Typed Array copyWithin with end index

vedant-pandey:
Optimization for non-shared Typed ArrayBuffers receives incorrect value for `from_byte_index` and `to_byte_index` since they get modified for byte-by-byte copy in the while loop that follows, this messes up the actual range and causes wrong range to be copied

Fixes: #6232


Libraries/LibJS/Runtime/TypedArrayPrototype.cpp
```diff
@@ -358,6 +358,25 @@ JS_DEFINE_NATIVE_FUNCTION(TypedArrayPrototype::copy_within)
...
+                && !to_end.has_overflow()
+                && from_end.value() <= buffer_byte_limit
+                && to_end.value() <= buffer_byte_limit) {
+                auto* base = buffer->buffer().data();
+                void const* src = base + from_byte_index;
+                void* dst = base + to_byte_index;
+                memmove(dst, src, count_bytes);
+                return typed_array;
+            }
+        }
```
gmta, line 378: Commit title and message: please describe what you're changing and why that helps fix the test case you added here. Please [avoid weasel words such as "fix"](https://github.com/LadybirdBrowser/ladybird/blob/master/CONTRIBUTING.md#code-submission-policy).

gmta (commented):
Hi @vedant-pandey, welcome to the project!

Libraries/LibJS/Runtime/TypedArrayPrototype.cpp
vedant-pandey: Hi @gmta I have updated the commit message and added cause block to better understand the cause and why this commit fixes it. Please let me know if I should add any other details.

gmta: You have updated the PR description instead of the commit message. Also, there are some typos in your commit title:

>  LibJS: Move Typed Array copyWithin optimization before index updation 

Should probably be:

>  LibJS: Move TypedArray copyWithin optimization before index update

trflynn89: @vedant-pandey The changes look good to me (and this bug actually becomes exposed after a recent spec change: https://github.com/tc39/ecma262/commit/d228070d9f9800e6320d89413540133aeaedec6c).

However, please include your PR description in the commit message itself (copied below with some grammatical errors fixed), as that is what shows up with when anyone looks at the `git log`:

> The optimization for non-shared ArrayBuffers operates on incorrect values of from_byte_index and to_byte_index because they will have been modified in the preceding steps. This causes the incorrect range to be copied within the buffer.

Issue #6259, opened 2025-09-21
flipkart.com crashes due to potentially bad serialized CSSRule


gotlougit:
### Summary

Ladybird currently crashes when trying to load the Flipkart home page (one of the biggest Indian e-commerce sites)

### Operating system

Linux

### Steps to reproduce

1. Go to flipkart.com

### Expected behavior

*Something* should load resembling the Flipkart home page.

### Actual behavior

The browser crashes and doesn't paint anything at all.

### URL for a reduced test case

https://flipkart.com

### HTML/SVG/etc. source for a reduced test case

```html
N/A
```

### Log output and (if possible) backtrace

```shell
VERIFICATION FAILED: m_ptr at /app/AK/RefPtr.h:280
/app/Build/release/lib/liblagom-ak.so.0(dump_backtrace+0x46) [0x7fc8f7afc296]
/app/Build/release/lib/liblagom-ak.so.0(ak_trap+0x6) [0x7fc8f7afc506]
/app/Build/release/lib/liblagom-ak.so.0(+0x1659d) [0x7fc8f7afc59d]
/app/Build/release/lib/liblagom-web.so.0(+0x58d2f6) [0x7fc8f90092f6]
/app/Build/release/lib/liblagom-web.so.0 Web::CSS::CSSRule::css_text() const 0xd) [0x7fc8f901d84d]
/app/Build/release/lib/liblagom-web.so.0(+0xe43299) [0x7fc8f98bf299]
/app/Build/release/lib/liblagom-js.so.0 JS::NativeFunction::call() 0xa1) [0x7fc8f8873d41]
/app/Build/release/lib/liblagom-js.so.0 JS::NativeFunction::internal_call(JS::ExecutionContext&, JS::Value) 0xb1) [0x7fc8f8873b41]
/app/Build/release/lib/liblagom-js.so.0 JS::call_impl(JS::VM&, JS::FunctionObject&, JS::Value, AK::Span<JS::Value const>) 0x167) [0x7fc8f877a857]
/app/Build/release/lib/liblagom-js.so.0 JS::Object::internal_get(JS::PropertyKey const&, JS::Value, JS::CacheableGetPropertyMetadata*, JS::Object::PropertyLookupPhase) const 0x293) [0x7fc8f887cbe3]
[...]

PR #6334, opened 2025-09-28
LibWeb: Check font family is present to avoid NPD when serializing CSS

vedant-pandey:
## Issue 
When serializing CSS with missing `Font-Family` the current implementation would crash due to a `Null Pointer Dereference` on missing attribute

## Fix
Ensure presence of `Font-Family` is checked to avoid the occurrence of the `NPD` 

## Additional
Refactor comments to stay coherent with the current spec

Fixes: #6259

### DIFF
--- a/Libraries/LibWeb/CSS/CSSFontFaceRule.cpp
+++ b/Libraries/LibWeb/CSS/CSSFontFaceRule.cpp
@@ -58,83 +58,102 @@ String CSSFontFaceRule::serialized() const
     // The result of concatenating the following:
 
     // 1. The string "@font-face {", followed by a single SPACE (U+0020).
+    auto is_empty = true;
     builder.append("@font-face { "sv);
 
-    // 2. The string "font-family:", followed by a single SPACE (U+0020).
-    builder.append("font-family: "sv);
-
-    // 3. The result of performing serialize a string on the rule’s font family name.
-    builder.append(descriptors.descriptor(DescriptorID::FontFamily)->to_string(SerializationMode::Normal));
-
-    // 4. The string ";", i.e., SEMICOLON (U+003B).
-    builder.append(';');
-
-    // 5. If the rule’s associated source list is not empty, follow these substeps:
-    if (auto sources = descriptors.descriptor(DescriptorID::Src)) {
-        // 1. A single SPACE (U+0020), followed by the string "src:", followed by a single SPACE (U+0020).
-        builder.append(" src: "sv);
-
-        // 2. The result of invoking serialize a comma-separated list on performing serialize a URL or serialize a LOCAL for each source on the source list.
-        builder.append(sources->to_string(SerializationMode::Normal));
-
-        // 3. The string ";", i.e., SEMICOLON (U+003B).
+    // 2. If the rule’s associated font-family is present, follow these substeps:
+    //    The string "font-family:", followed by a single SPACE (U+0020).
+    //    followed by the result of performing serialize a string on the rule’s font family name.
+    //    followed by the string ";", i.e., SEMICOLON (U+003B).
+    if (auto font_family = descriptors.descriptor(DescriptorID::FontFamily)) {
+        builder.append("font-family: "sv);
+        builder.append(descriptors.descriptor(DescriptorID::FontFamily)->to_string(SerializationMode::Normal));
         builder.append(';');
+        is_empty = false;
     }
 
-    // 6. If rule’s associated unicode-range descriptor is present, a single SPACE (U+0020), followed by the string "unicode-range:", followed by a single SPACE (U+0020), followed by the result of performing serialize a <'unicode-range'>, followed by the string ";", i.e., SEMICOLON (U+003B).
+    // 3. If the rule’s associated source list is not empty, a single SPACE (U+0020),
+    //    followed by the string "src:", followed by a single SPACE (U+0020).
+    //    followed by the result of invoking serialize a comma-separated list on performing serialize a URL or serialize a LOCAL for each source on the source list.
+    //    followed by the string ";", i.e., SEMICOLON (U+003B).
+    if (auto sources = descriptors.descriptor(DescriptorID::Src)) {
+        if (!is_empty)
+            builder.append(' ');
+        builder.append("src: "sv);
+        builder.append(sources->to_string(SerializationMode::Normal));
+        builder.append(';');
+        is_empty = false;
+    }
+
+    // 4. If rule’s associated unicode-range descriptor is present, a single SPACE (U+0020), followed by the string "unicode-range:", followed by a single SPACE (U+0020), followed by the result of performing serialize a <'unicode-range'>, followed by the string ";", i.e., SEMICOLON (U+003B).
     if (auto unicode_range = descriptors.descriptor(DescriptorID::UnicodeRange)) {
+        if (!is_empty)
+            builder.append(' ');
         builder.append(" unicode-range: "sv);
         builder.append(unicode_range->to_string(SerializationMode::Normal));
         builder.append(';');
+        is_empty = false;
     }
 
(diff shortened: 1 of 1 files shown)


Libraries/LibWeb/CSS/CSSFontFaceRule.cpp
```diff
@@ -58,83 +58,102 @@ String CSSFontFaceRule::serialized() const
...
         builder.append(';');
+        is_empty = false;
     }
 
-    // 6. If rule’s associated unicode-range descriptor is present, a single SPACE (U+0020), followed by the string "unicode-range:", followed by a single SPACE (U+0020), followed by the result of performing serialize a <'unicode-range'>, followed by the string ";", i.e., SEMICOLON (U+003B).
+    // 4. If rule’s associated unicode-range descriptor is present, a single SPACE (U+0020), followed by the string "unicode-range:", followed by a single SPACE (U+0020), followed by the result of performing serialize a <'unicode-range'>, followed by the string ";", i.e., SEMICOLON (U+003B).
     if (auto unicode_range = descriptors.descriptor(DescriptorID::UnicodeRange)) {
+        if (!is_empty)
+            builder.append(' ');
         builder.append(" unicode-range: "sv);
```
tcl3, line 86: The leading space is no longer necessary here. The same is true for each descriptor below that follows this pattern.

tcl3 (changes requested):
The comments you've modified come directly from [the spec](<https://drafts.csswg.org/cssom/#ref-for-cssfontfacerule>). Rather than modifying the spec comments could you add `// AD-HOC: ` comments that explain how what you've done diverges from the spec.

Could you add a test for this? In this case a crash test would be sufficient - this could look something like the [minimal reproduction](<https://github.com/LadybirdBrowser/ladybird/issues/6259#issuecomment-3328617910>) on the linked issue.

Libraries/LibWeb/CSS/CSSFontFaceRule.cpp
vedant-pandey: Hi @tcl3, 
> The comments you've modified come directly from [the spec](https://drafts.csswg.org/cssom/#ref-for-cssfontfacerule). Rather than modifying the spec comments could you add // AD-HOC:  comments that explain how what you've done diverges from the spec.

I have modified the comments to follow the same pattern, there is no divergence from the spec but instead intial few followed different pattern from the latter so made them all inline with the latter ones
For eg
Previous -
```
// 2. The string "font-family:", followed by a single SPACE (U+0020).
    builder.append("font-family: "sv);

    // 3. The result of performing serialize a string on the rule’s font family name.
    builder.append(descriptors.descriptor(DescriptorID::FontFamily)->to_string(SerializationMode::Normal));

    // 4. The string ";", i.e., SEMICOLON (U+003B).
    builder.append(';');

    // 5. If the rule’s associated source list is not empty, follow these substeps:
```
New - 
```
// 3. If the rule’s associated source list is not empty, a single SPACE (U+0020),
    //    followed by the string "src:", followed by a single SPACE (U+0020).
    //    followed by the result of invoking serialize a comma-separated list on performing serialize a URL or serialize a LOCAL for each source on the source list.
    //    followed by the string ";", i.e., SEMICOLON (U+003B).
```

Additionally fixed the numbering of the comments

> Could you add a test for this? In this case a crash test would be sufficient - this could look something like the https://github.com/LadybirdBrowser/ladybird/issues/6259#issuecomment-3328617910 on the linked issue.

Could you please point out any documentation for adding crash tests? I did see a `Tests/LibWeb/Crash/` folder but couldn't find a documentation on how to add a relevant test.

Psychpsyo: > > The comments you've modified come directly from [the spec](https://drafts.csswg.org/cssom/#ref-for-cssfontfacerule). Rather than modifying the spec comments could you add // AD-HOC:  comments that explain how what you've done diverges from the spec.
> 
> I have modified the comments to follow the same pattern, there is no divergence from the spec but instead intial few followed different pattern from the latter so made them all inline with the latter ones

Yes, the initial few follow a different pattern from the latter ones in the spec: https://drafts.csswg.org/cssom/#ref-for-cssfontfacerule
The fact that you changed them to be the same is a divergence from the spec.

vedant-pandey: Hi @Psychpsyo thanks for clarification, I've reverted the comments to be aligned with the spec.

Psychpsyo:
I think this also probably warrants a spec bug be filed over at https://github.com/w3c/csswg-drafts/labels/cssom-1, since the spec is telling us to do things that seem impossible.

Libraries/LibWeb/CSS/CSSFontFaceRule.cpp
```diff
@@ -60,23 +60,23 @@ String CSSFontFaceRule::serialized() const
...
 
-    // 2. The string "font-family:", followed by a single SPACE (U+0020).
-    builder.append("font-family: "sv);
-
-    // 3. The result of performing serialize a string on the rule’s font family name.
-    builder.append(descriptors.descriptor(DescriptorID::FontFamily)->to_string(SerializationMode::Normal));
-
-    // 4. The string ";", i.e., SEMICOLON (U+003B).
-    builder.append(';');
+    if (auto font_family = descriptors.descriptor(DescriptorID::FontFamily)) {
```
Psychpsyo, line 63: There should probably be a comment here, explaining why we are deviating from the spec on this. Something like
> // AD-HOC: Since the font family might be missing, we need to check if it exists first.

Maybe also with a link to the spec bug.

Libraries/LibWeb/CSS/CSSFontFaceRule.cpp
```diff
@@ -60,23 +60,23 @@ String CSSFontFaceRule::serialized() const
...
+        builder.append("font-family: "sv);
+        // 3. The result of performing serialize a string on the rule’s font family name.
+        builder.append(descriptors.descriptor(DescriptorID::FontFamily)->to_string(SerializationMode::Normal));
+        // 4. The string ";", i.e., SEMICOLON (U+003B).
+        builder.append(';');
+    }
 
     // 5. If the rule’s associated source list is not empty, follow these substeps:
+    //    followed by the result of invoking serialize a comma-separated list on performing serialize a URL or serialize a LOCAL for each source on the source list.
+    //    followed by the string ";", i.e., SEMICOLON (U+003B).
```
Psychpsyo, line 74: These additions don't belong here?

Libraries/LibWeb/CSS/CSSFontFaceRule.cpp
```diff
@@ -60,23 +60,23 @@ String CSSFontFaceRule::serialized() const
...
     if (auto sources = descriptors.descriptor(DescriptorID::Src)) {
         // 1. A single SPACE (U+0020), followed by the string "src:", followed by a single SPACE (U+0020).
         builder.append(" src: "sv);
-
         // 2. The result of invoking serialize a comma-separated list on performing serialize a URL or serialize a LOCAL for each source on the source list.
         builder.append(sources->to_string(SerializationMode::Normal));
-
```
Psychpsyo, line 79: The commit shouldn't remove lines in code that it isn't touching.

AtkinsSJ:
It looks like this got a little bit mangled. Can you please combine this into a single commit as it's a single change, and also correct the comments?

The way to handle the spec being wrong (as it is here) is:
- Keep the spec comments themselves as they are.
- Report a spec issue if there isn't one already.
- Add some `// AD-HOC: ...` comments of your own to the parts where you had to ignore what the spec currently says.

(I see Cameron has commented already while I was writing this, but I'll keep it anyway for emphasis.)

vedant-pandey:
Hi, sorry I was on vacation, will be completing this soon

PR #6336, opened 2025-09-28
Android build is broken (again)

Stewarpt:
Android build doesn't build


trflynn89:
IMO it is silly to keep adding/removing this note as the state of the Android build is regularly in flux. Instead, let's at most just add a note that the Android port is currently experimental and may or may not work at any given time.

Please also see https://github.com/LadybirdBrowser/ladybird/blob/master/CONTRIBUTING.md regarding how we format our commit messages.

PR #6340, opened 2025-09-29
LibWeb: Correctly resolve `position-area` computed value

tcl3:
This PR ensures that the correct short equivalents are used for the computed value of `position-area` when 2 keywords specified.

Fixes 112 WPT subtests.


Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -3567,6 +3569,106 @@ NonnullRefPtr<StyleValue const> StyleComputer::compute_opacity(NonnullRefPtr<Sty
...
+        case Keyword::SpanSelfBlockEnd:
+        case Keyword::SpanSelfInlineEnd:
+            return KeywordStyleValue::create(Keyword::SpanSelfEnd);
+        default:
+            break;
+        }
+        return keyword_value;
+    };
+
+    VERIFY(specified_value->is_value_list());
```
gmta: This `VERIFY` happens in `->as_value_list()` as well; imo we should just use the `as_...()` methods directly and depend on them to check their type.

PR #6342, opened 2025-09-29
Initial support for XPath evaluations

### LINKED ISSUE #5451: XPathEvaluator is not defined, breaking htmx
### Summary

When I open a website in Ladybird that makes use of htmx, functionality that makes use of it is broken due to missing XPathEvaluator support.

### Operating system

Linux

### Steps to reproduce

1. Go to any website with htmx (such as https://htmx.org/ or my website https://kilonova.ro)
2. Observe in the JS console how XPathEvaluator is not defined

### Expected behavior

The HTMX library loads, especially given the fact that until recently, it had support for IE11 (and now the minimum requirement is ES6)

### Actual behavior

The htmx code does not load because of this error, leaving `window.htmx` undefined and breaking all htmx integration. 

Sidenote: funnily enough, I reported this [exact same issue to Servo](https://github.com/servo/servo/issues/34060) when I tried it out in October.

N/A

```html
<script src="https://cdn.jsdelivr.net/npm/htmx.org@2.0.6/dist/htmx.min.js"></script>
  <!-- have a button POST a click via AJAX -->
<!-- the expected behaviour (in Firefox for example) is for either a CORS block (if run from file://) or 404. In Ladybird, nothing happens because the script above fails to load -->
  <button hx-post="/clicked" hx-swap="outerHTML">
    Click Me
  </button>
```

```shell
103453.949 WebContent(530270): Unhandled JavaScript exception: [ReferenceError] 'XPathEvaluator' is not defined
103453.949 WebContent(530270):     at https://cdn.jsdelivr.net/npm/htmx.org@2.0.6/dist/htmx.min.js:1:51005
    at https://cdn.jsdelivr.net/npm/htmx.org@2.0.6/dist/htmx.min.js:1:51005
```

_No response_

_No response_

### Contribute a patch?

johannesg:
This PR implements IDL types XPathEvaluator, XPathExpression, XPathResult and XPathNSResolver. Also `Document.evaluate` is included in this.

Lots of WPT tests pass now, although there are quite a few left to fix. 

Among things not implemented are support for namespaces. XPathNSResolver is currently just a stub and returns null.

Under the hood it uses libxml2, first by creating a mirror xml tree and then evaluating that (thanks to alimpfard for an initial POC). I'm not convinced this is sufficient long term, I suspect some of the WPT tests will never work with libxml2 as there seems to be some differences between DOM XPath and XML XPath. Eventually, a complete solution will probably need it's own parser and xpath processor. However, this PR should get some sites working, for example, https://htmx.org seems to loads fine.

This is my first PR to ladybird so probably lots of things I have done wrong or missed, so please let me know any issues and I will try to fix them.

For example:
1. Should I squash all commits into one or leave the as is? I have squashed some of them to try to make the commits meaningful bug can probably be improved further.
2. I created a new namespace, DOM::XPath for the implementation. Should they be part of the DOM namespace instead?

Fixes #5451


gmta:
Hi, welcome to the project! You should start off by applying a bit of commit hygiene, especially: don't have "initial work", "more implementation", "got tests working", "various fixes", etc. commits. You should have concise commits that describe what they implement instead. You should not have separate fix commits, but rather rework the fixes in the original commits that introduced the issue originally.

johannesg:
I have reworked the commits now. Hopefully it looks a lot cleaner.

gmta:
I tried this PR on my Gitea instance, the dashboard looks a lot better with XPath support!

| Before | After |
|--|--|
| before | after |

Libraries/LibWeb/CMakeLists.txt
```diff
@@ -1064,6 +1064,10 @@ set(SOURCES
     XHR/XMLHttpRequestUpload.cpp
     XLink/AttributeNames.cpp
     XML/XMLDocumentBuilder.cpp
+    XPath/XPathResult.cpp
+    XPath/XPathExpression.cpp
+    XPath/XPathEvaluator.cpp
+    XPath/XPathNSResolver.cpp
```
gmta, line 1072: Please keep these entries in alphabetical order

Libraries/LibWeb/Forward.h
```diff
@@ -1290,3 +1290,12 @@ class TrustedTypePolicyFactory;
 struct TrustedTypePolicyOptions;
 
 }
+
+namespace Web::XPath {
+
+class XPathResult;
+class XPathEvaluator;
+class XPathExpression;
+class XPathNSResolver;
```
gmta, line 1300: Same here - alphabetical order please :)

Libraries/LibWeb/XPath/XPathEvaluator.idl
```diff
@@ -0,0 +1,22 @@
+#import <DOM/Node.idl>
+
+// FIXME
+// callback interface XPathNSResolver {
+//   DOMString? lookupNamespaceURI(DOMString? prefix);
+// };
+
+interface mixin XPathEvaluatorBase {
```
gmta, line 9: Please add spec links to all the IDL files you add in this commit.

Libraries/LibWeb/XPath/XPathResult.cpp
```diff
@@ -0,0 +1,57 @@
...
+{
+    WEB_SET_PROTOTYPE_FOR_INTERFACE(XPathResult);
+    Base::initialize(realm);
+}
+
+void XPathResult::visit_edges(Cell::Visitor& visitor)
+{
+    Base::visit_edges(visitor);
+    for (auto n : m_node_set)
+        visitor.visit(n);
```
gmta: You should be able to visit `m_node_set` directly.

Libraries/LibWeb/XPath/XPathResult.cpp
```diff
@@ -0,0 +1,57 @@
...
+{
+    if (m_node_set_iter == m_node_set.end())
+        return nullptr;
+
+    return *m_node_set_iter++;
+}
+
+GC::Ptr<DOM::Node> XPathResult::snapshot_item(int index)
+{
+    size_t i = index;
```
gmta: Why do you need this variable?

Libraries/LibWeb/XPath/XPath.cpp
```diff
@@ -0,0 +1,237 @@
...
+}
+
+WebIDL::ExceptionOr<GC::Ref<XPathExpression>> create_expression(JS::Realm& realm, String const& expression, GC::Ptr<XPathNSResolver> resolver)
+{
+    return realm.create<XPathExpression>(realm, expression, resolver);
+}
+
+// TODO:
+// Namespaces
+// Namespace resolver callbacks
```
gmta: If you intend to keep track of longer living TODOs, please mark them as `// FIXME: [..]` comments and/or create a GitHub issue for them to link to.

Libraries/LibWeb/XPath/XPathNSResolver.h
```diff
@@ -0,0 +1,36 @@
...
+
+    virtual ~XPathNSResolver() = default;
+
+    // There is no Prototype for this. Still needed?
+    // virtual void initialize(JS::Realm&) override;
+
+    String* lookupNamespaceURI(String* prefix);
+
+private:
+    // TODO: GC::Ref field. Implement visit_edges() and visit() for the GC to find it.
```
AtkinsSJ: CI is complaining because visit_edges() isn't implemented.

Libraries/LibWeb/XPath/XPath.cpp
```diff
@@ -0,0 +1,237 @@
...
+            GC::Ptr<DOM::Node> ref(*dom_node);
+
+            node_list.append(ref);
+
+            dbgln("XPath: Result Node: {}", dom_node->node_name());
+        }
+        result->set_node_set(move(node_list));
+    }
+
+    dbgln("XPath: Result Type: {}", static_cast<int>(xpath_result->type));
```
gmta: Are these `dbgln`s supposed to be here?

Libraries/LibWeb/XPath/XPath.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+    switch (xpath_result->type) {
+    case XPATH_UNDEFINED:
+        break;
+    case XPATH_NODESET: {
+        // The node set is stored the same regardless of the result type, so just set it to whatever the user wants
+        if (type >= 4 && type <= 9)
```
gmta: Can we compare against constants/an enum instead of literals here?

Libraries/LibWeb/XPath/XPathNSResolver.h
johannesg: Added visit_edges() to the class

Libraries/LibWeb/CMakeLists.txt
johannesg: Should be fixed now

gmta:
@johannesg Please incorporate the changes you make into the original commits instead of adding new commits to your PR branch.

Libraries/LibWeb/Forward.h
johannesg: Should be fixed now

Libraries/LibWeb/XPath/XPathEvaluator.idl
johannesg: Done

Libraries/LibWeb/XPath/XPathResult.cpp
johannesg: Done

Libraries/LibWeb/XPath/XPathResult.cpp
johannesg: I had to cast it to size_t. Changed it to an inline static_cast instead

Libraries/LibWeb/XPath/XPath.cpp
johannesg: I removed the comments as they are easily discovered anyway in the WPT tests

Libraries/LibWeb/XPath/XPath.cpp
johannesg: Probably not. Removed

Libraries/LibWeb/XPath/XPath.cpp
johannesg: Yes. Also, moved the code into XPathResult instead

johannesg:
> @johannesg Please incorporate the changes you make into the original commits instead of adding new commits to your PR branch.

Absolutely! I just thought it would be easer for you to see the changes during the review as individual commits. But will squash it together as soon as everything looks ok.

gmta:
> Absolutely! I just thought it would be easer for you to see the changes during the review as individual commits. But will squash it together as soon as everything looks ok.

As long as you don't rebase your PR branch at the same time, there's a neat little button "View changes" in GitHub's UI that shows only the relevant changes you pushed!

johannesg:
> > Absolutely! I just thought it would be easer for you to see the changes during the review as individual commits. But will squash it together as soon as everything looks ok.
> 
> As long as you don't rebase your PR branch at the same time, there's a neat little button "View changes" in GitHub's UI that shows only the relevant changes you pushed!

Alright then! I have done this now.

johannesg:
Pushed a fix for the two failing tests

Libraries/LibWeb/XPath/XPathNSResolver.h
```diff
@@ -0,0 +1,36 @@
...
+
+public:
+    [[nodiscard]] static GC::Ref<XPathNSResolver> create(JS::Realm&, GC::Ref<WebIDL::CallbackType>);
+    XPathNSResolver(JS::Realm&, GC::Ref<WebIDL::CallbackType>);
+
+    virtual ~XPathNSResolver() = default;
+    virtual void visit_edges(Cell::Visitor&) override;
+
+    // There is no Prototype for this. Still needed?
+    // virtual void initialize(JS::Realm&) override;
```
gmta: You can leave these lines out if you don't need to initialize a prototype.

Libraries/LibWeb/XPath/XPathNSResolver.h
```diff
@@ -0,0 +1,36 @@
...
+
+    String* lookupNamespaceURI(String* prefix);
```
gmta: What is this? If it's needed,
* please use snake_case, i.e. `lookup_namespace_uri`
* don't return a pointer to a String but a String (copy), StringView or Utf8View

Libraries/LibWeb/XPath/XPathResult.h
```diff
@@ -0,0 +1,61 @@
...
+    unsigned short const ANY_TYPE = 0;
+    unsigned short const NUMBER_TYPE = 1;
+    unsigned short const STRING_TYPE = 2;
+    unsigned short const BOOLEAN_TYPE = 3;
+    unsigned short const UNORDERED_NODE_ITERATOR_TYPE = 4;
+    unsigned short const ORDERED_NODE_ITERATOR_TYPE = 5;
+    unsigned short const UNORDERED_NODE_SNAPSHOT_TYPE = 6;
+    unsigned short const ORDERED_NODE_SNAPSHOT_TYPE = 7;
+    unsigned short const ANY_UNORDERED_NODE_TYPE = 8;
+    unsigned short const FIRST_ORDERED_NODE_TYPE = 9;
```
gmta: Types related to web specs often use the WebIDL types, i.e. `WebIDL::UnsignedShort` instead of `unsigned short`. For all other code, we usually use the types as defined in `AK/Types.h`, i.e. `u16`.

Same for the fields `m_result_type` and `m_snapshot_length`.

Libraries/LibWeb/XPath/XPath.cpp
```diff
@@ -0,0 +1,223 @@
...
+#include <libxml/valid.h>
+#include <libxml/xmlstring.h>
+#include <libxml/xpath.h>
+
+#include "XPath.h"
+#include "XPathExpression.h"
+
+namespace Web::XPath {
+
+namespace {
```
gmta: Why a nested empty namespace?

Libraries/LibWeb/XPath/XPath.cpp
```diff
@@ -0,0 +1,223 @@
...
+WebIDL::ExceptionOr<GC::Ref<XPathExpression>> create_expression(JS::Realm& realm, String const& expression, GC::Ptr<XPathNSResolver> resolver)
+{
+    return realm.create<XPathExpression>(realm, expression, resolver);
+}
+
+WebIDL::ExceptionOr<GC::Ref<XPathResult>> evaluate(JS::Realm& realm, String const& expression, DOM::Node const& context_node, GC::Ptr<XPathNSResolver> /*resolver*/, unsigned short type, GC::Ptr<XPathResult> result)
+{
+    // Parse the expression as xpath
+    ByteString bytes = expression.bytes_as_string_view();
+    auto* xpath_compiled = xmlXPathCompile(reinterpret_cast<xmlChar const*>(bytes.characters()));
```
gmta: Can you use `bit_cast` here and in `mirror_node`?

Libraries/LibWeb/XPath/XPath.cpp
```diff
@@ -0,0 +1,223 @@
...
+        auto* xml_fragment = xmlNewDocFragment(doc);
+        xml_fragment->_private = bit_cast<void*>(&node);
+        auto children = fragment.children_as_vector();
+        for (auto& child : children) {
+            xmlAddChild(xml_fragment, mirror_node(doc, *child));
+        }
+        return xml_fragment;
+    }
+    case DOM::NodeType::NOTATION_NODE: {
+        TODO(); // Unused in libxml2
```
gmta: If these are not to be expected from libxml2, maybe `VERIFY_NOT_REACHED()` is better for these cases? There's nothing for us "to do" after all.

Libraries/LibWeb/XPath/XPath.cpp
```diff
@@ -0,0 +1,223 @@
...
+        Vector<GC::Ptr<DOM::Node>> node_list;
+
+        if (xpath_result->nodesetval && xpath_result->nodesetval->nodeNr > 0) {
+            for (int i = 0; i < xpath_result->nodesetval->nodeNr; i++) {
+                auto* node = xpath_result->nodesetval->nodeTab[i];
+                auto* dom_node = static_cast<DOM::Node*>(node->_private);
+
+                GC::Ptr<DOM::Node> ref(*dom_node);
+
+                node_list.append(ref);
```
gmta: You should be able to simply append `dom_node` here as a pointer. Make it `unchecked_append` if you added `ensure_capacity` above.

Libraries/LibWeb/XPath/XPath.cpp
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    switch (xpath_result->type) {
+    case XPATH_UNDEFINED:
+        break;
+    case XPATH_NODESET: {
+        Vector<GC::Ptr<DOM::Node>> node_list;
+
+        if (xpath_result->nodesetval && xpath_result->nodesetval->nodeNr > 0) {
+            for (int i = 0; i < xpath_result->nodesetval->nodeNr; i++) {
```
gmta, line 132: Since you know the number of nodes you're going to append, you can add `node_list.ensure_capacity(xpath_result->nodesetval->nodeNr)` above this line and use `.unchecked_append(..)` below.

Libraries/LibWeb/XPath/XPathResult.h
```diff
@@ -19,16 +19,16 @@ class XPathResult : public Bindings::PlatformObject {
...
+    static unsigned short const ANY_TYPE = 0;
+    static unsigned short const NUMBER_TYPE = 1;
+    static unsigned short const STRING_TYPE = 2;
+    static unsigned short const BOOLEAN_TYPE = 3;
+    static unsigned short const UNORDERED_NODE_ITERATOR_TYPE = 4;
+    static unsigned short const ORDERED_NODE_ITERATOR_TYPE = 5;
+    static unsigned short const UNORDERED_NODE_SNAPSHOT_TYPE = 6;
+    static unsigned short const ORDERED_NODE_SNAPSHOT_TYPE = 7;
+    static unsigned short const ANY_UNORDERED_NODE_TYPE = 8;
+    static unsigned short const FIRST_ORDERED_NODE_TYPE = 9;
```
gmta: Please add these `static`s in the previous commit, where you introduced these lines.

Libraries/LibWeb/XPath/XPath.cpp
johannesg: This was actually suggested by some linter I had. Functions inside unnamed namespaces have internal linkage and are invisible outside the cpp file. So just a way of isolating the implementation somewhat, which might not be necessary as it is already in the XPath namespace . I can remove it if you want.

gmta: We usually just add static functions to the .cpp instead; I don't think it's necessary to keep it here. You can just keep the implementations in the same namespace :)

johannesg: That's what I had initally :-D

Libraries/LibWeb/XPath/XPath.cpp
johannesg: Better to return nullptr instead IMO. If the DOM actually contains any of these, then VERIFY_NOT_REACHED() would crash right?

gmta: That's right, but TODO() also crashes - so I assumed that's what you wanted.

johannesg:
I have pushed fixes for (hopefully) everything above. Please review again at your convenience.

johannesg:
Fixed the latest CI errors, plus rebased on latest master

gmta:
@johannesg If you're sure you've resolved a comment, feel free to mark it as resolved it yourself!

Libraries/LibWeb/XPath/XPathNSResolver.h
gmta: I think you haven't addressed the "if it's needed" part of my comment; it doesn't seem like this method is used anywhere? Either remove this method altogether, or if it's used in an algorithm, use it there and leave actionable `// FIXME: ..` comments. Generally we don't add unused code.

Libraries/LibWeb/XPath/XPath.cpp
```diff
@@ -0,0 +1,214 @@
...
+#include <libxml/valid.h>
+#include <libxml/xmlstring.h>
+#include <libxml/xpath.h>
+
+#include "XPath.h"
+
+namespace Web::XPath {
+
+static xmlNodePtr mirror_node(xmlDocPtr doc, Web::DOM::Node const& node);
+static void convert_xpath_result(xmlXPathObjectPtr xpath_result, XPath::XPathResult* result, unsigned short type);
```
gmta: You can simply put the implementation of these methods here instead of forward-declaring them - that makes it a bit easier to see that the methods are static.

Libraries/LibWeb/XPath/XPathNSResolver.h
johannesg: Fair enough. I suppose I could remove the class altogether until I have an implementation ready for namespace lookups. On the other hand evaluate does take a resolver as arguments, so it is there more like a placeholder. Maybe it can be replaced with a GC::Ref<CallbackType>?

gmta: I'd leave XPathNSResolver as-is since it's actually being referenced by other code; but this method serves no purpose.

Not sure what you mean by `GC::Ref`, the difference with `GC::Ptr` (which you use now) is that you _need_ to provide an argument, instead of it being optional.

Libraries/LibWeb/XPath/XPath.cpp
johannesg: Personal preference, but I prefer to have "top level" methods at the top of the file, hence the forward declaration. Anyways, I moved them up now 👍 .

Libraries/LibWeb/XPath/XPathNSResolver.h
johannesg: Nevermind, I just mixed them up (GC::Ref)

PR #6343, opened 2025-09-29
Meta: Add .env to .gitignore

birkoffe:
It is convenient to have a local .env file to store variables to avoid a long startup command like `CC=$(brew --prefix llvm)/bin/clang CXX=$(brew --prefix llvm)/bin/clang++ ./Meta/ladybird.py run`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,7 @@ output/
 .venv/
 .venv*/
 venv*/
+.env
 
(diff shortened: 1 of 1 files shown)


nico:
You can add `.env` to .git/info/exclude, then your workflow doesn't have to leak into upstream.

PR #6346, opened 2025-09-30
LibWeb/WebGL: Define Float32List exactly like in the spec

kalenikaliaksandr:
Use `Float32Array or sequence<GLfloat>` instead of
`BufferSource or sequence<GLfloat>`. This meaningfully changes behavior
for `Float16Array` and `Float64Array`: they are now converted to
`sequence<GLfloat>` by iterating the typed array, rather than being
treated as a `BufferSource`. As a result, many WebGL calls now work
correctly where we previously crashed in `VERIFY_NOT_REACHED()` due to
the assumption that a `BufferSource` was always a `Float32Array`.

Fixes https://github.com/LadybirdBrowser/ladybird/issues/5962


tcl3:
Can we add a test that exposes the crash this PR fixes?

Lubrsi:
I don't think we currently have the infrastructure to create a non-GPU backed WebGL context, which would make `getContext` return `null`. However, it seems ANGLE does have a way to have a test driver, we need to request a NULL device type by setting `EGL_PLATFORM_ANGLE_DEVICE_TYPE_ANGLE` to `EGL_PLATFORM_ANGLE_DEVICE_TYPE_NULL_ANGLE` in display_attributes:
https://github.com/LadybirdBrowser/ladybird/blob/820fee434e53655ae4670a09e3e6b085768ec9c7/Libraries/LibWeb/WebGL/OpenGLContext.cpp#L134-L144
https://github.com/google/angle/blob/fe939e6ad00626ea57a74b8ef4ce9ec3bdeb76b0/extensions/EGL_ANGLE_platform_angle.txt
https://github.com/google/angle/blob/fe939e6ad00626ea57a74b8ef4ce9ec3bdeb76b0/extensions/EGL_ANGLE_platform_angle_null.txt

Meta/Lagom/Tools/CodeGenerators/LibWeb/BindingsGenerator/IDLGenerators.cpp
```diff
@@ -1449,11 +1449,19 @@ static void generate_to_cpp(SourceGenerator& generator, ParameterType& parameter
...
         });
 
-        if (has_typed_array_name || includes_object) {
+        if (typed_array_name != types.end()) {
+            union_generator.set("typed_array_type", (*typed_array_name)->name());
+            union_generator.append(R"~~~(
+            if (is<JS::@typed_array_type@>(@js_name@@js_suffix@_object)) {
+                auto& typed_array = static_cast<JS::@typed_array_type@&>(@js_name@@js_suffix@_object);
+                return GC::make_root(typed_array);
+            }
```
gmta:
```cpp
            if (auto* typed_array = as_if<JS::@typed_array_type@>(@js_name@@js_suffix@_object))
                return GC::make_root(*typed_array);
```

gmta:
Nice code dedupication! 🚀  just need a test as well, like @tcl3 mentioned

kalenikaliaksandr:
> Nice code dedupication! 🚀 just need a test as well, like @tcl3 mentioned

like Luke mentioned we currently don't have a way to create non-GPU WebGL context. I am not gonna bring up WebGL testing infrastructure in this PR, it's definitely will be a separate change.

gmta:
> > Nice code dedupication! 🚀 just need a test as well, like @tcl3 mentioned
> 
> like Luke mentioned we currently don't have a way to create non-GPU WebGL context. I am not gonna bring up WebGL testing infrastructure in this PR, it's definitely will be a separate change.

Oh that's unfortunate - should we have an issue for that at least, referencing the issue you fixed with this PR?

kalenikaliaksandr:
> > > Nice code dedupication! 🚀 just need a test as well, like @tcl3 mentioned
> > 
> > 
> > like Luke mentioned we currently don't have a way to create non-GPU WebGL context. I am not gonna bring up WebGL testing infrastructure in this PR, it's definitely will be a separate change.
> 
> Oh that's unfortunate - should we have an issue for that at least, referencing the issue you fixed with this PR?

sure, commit description mentions issue fixed by this change https://github.com/LadybirdBrowser/ladybird/issues/5962 it's now possible to open https://renewables-map.robinhawkes.com and it won't crash 😅

gmta:
> sure, commit description mentions issue fixed by this change #5962 it's now possible to open https://renewables-map.robinhawkes.com and it won't crash 😅

Not quite what I meant, but opened #6349 so you don't have to :)

kalenikaliaksandr:
> > sure, commit description mentions issue fixed by this change #5962 it's now possible to open [renewables-map.robinhawkes.com](https://renewables-map.robinhawkes.com) and it won't crash 😅
> 
> Not quite what I meant, but opened #6349 so you don't have to :)

ah, yes, I was in the process of creating the same issue once I realized I misunderstood you 😅

Issue #6279, opened 2025-09-23
site serves PNG favicon named 'favicon.svg', leads to XMLDocumentBuilder assert


danielbprice:
### Summary

When I try to load https://www.letsbuildui.dev/articles/fun-with-custom-cursors/ in ladybird, it always crashes.  I am current with fceb73e65aa7a3bd6f3a4a30f8bc1fea8c4249e2.

### Operating system

macOS

### Steps to reproduce

1. Load https://www.letsbuildui.dev/articles/fun-with-custom-cursors/
2. Observe crash

### Expected behavior

No crash.

### Actual behavior

Crash.  Seems to occur because the website has this declaration:

```
$ curl --silent 'https://www.letsbuildui.dev/articles/fun-with-custom-cursors/'|grep favicon
    <link rel="icon" href="https://www.letsbuildui.dev/content/images/size/w256h256/format/png/2023/08/favicon.svg" type="image/png">
```
So it's called favicon.svg, but it's actually `image/png`...

```
$ curl -LO 'https://www.letsbuildui.dev/content/images/size/w256h256/format/png/2023/08/favicon.svg'
$ file favicon.svg
favicon.svg: PNG image data, 256 x 256, 8-bit/color RGBA, non-interlaced
```
Yes, a png.  But from the stack trace, we're clearly using SVG code to parse the favicon.  That seems to correspond with the code in Libraries/LibWeb/HTML/HTMLLinkElement.cpp:

```
  static NonnullRefPtr<Core::Promise<bool>> decode_favicon(ReadonlyBytes favicon_data, URL::URL const& favicon_url, GC::Ref<DOM::Document> document)
  {
      if (favicon_url.basename().ends_with(".svg"sv)) {
          auto result = SVG::SVGDecodedImageData::create(document->realm(), document->page(), favicon_url, favicon_data);
          auto promise = Core::Promise<bool>::construct();
          if (result.is_error()) {
              promise->reject(Error::from_string_view("Failed to decode SVG favicon"sv));
              return promise;
          }
...
      }
```

While this portion seems to be coded defensively, it seems like the issue might be in the XML parser underneath.

### URL for a reduced test case

https://www.letsbuildui.dev/articles/fun-with-custom-cursors/

### HTML/SVG/etc. source for a reduced test case

```html
N/A
```

### Log output and (if possible) backtrace

```shell
[...]

PR #6353, opened 2025-09-30
LibWeb+LibXML: Make `Listener::set_source(ByteString)` fallible

rmg-x:
`set_source` takes a ByteString but the implementation might require a
specific encoding. Make it fallible so that we don't need to crash in
the case of invalid UTF-8 or similar.

The test includes a sequence of invalid UTF-8 bytes that crash the
browser without this change.

Related issue: #6279, related PR #6348

### DIFF
--- a/Libraries/LibXML/Parser/Parser.cpp
+++ b/Libraries/LibXML/Parser/Parser.cpp
@@ -199,7 +199,12 @@ ErrorOr<void, ParseError> Parser::parse_with_listener(Listener& listener)
 {
     m_listener = &listener;
     ScopeGuard unset_listener { [this] { m_listener = nullptr; } };
-    m_listener->set_source(m_source);
+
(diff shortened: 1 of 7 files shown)


Libraries/LibXML/Parser/Parser.cpp
```diff
@@ -199,7 +199,13 @@ ErrorOr<void, ParseError> Parser::parse_with_listener(Listener& listener)
 {
     m_listener = &listener;
     ScopeGuard unset_listener { [this] { m_listener = nullptr; } };
-    m_listener->set_source(m_source);
+
+    if (auto const maybe_source_error = m_listener->set_source(m_source); maybe_source_error.is_error()) {
+        auto const error_message = maybe_source_error.error().string_literal();
+        dbgln("Failed to set XML parser source: {}", error_message);
```
trflynn89: Let's not keep this debug statement

PR #6355, opened 2025-09-30
LibWeb: Remove spin_until from create_navigation_params_by_fetching and exception throwing from Fetch

Lubrsi:
This is in similar fashion and reason to e095bf3

### DIFF
--- a/Libraries/LibWeb/HTML/Navigable.cpp
+++ b/Libraries/LibWeb/HTML/Navigable.cpp
@@ -59,30 +59,57 @@ namespace Web::HTML {
 
 GC_DEFINE_ALLOCATOR(Navigable);
 
-class ResponseHolder : public JS::Cell {
-    GC_CELL(ResponseHolder, JS::Cell);
-    GC_DECLARE_ALLOCATOR(ResponseHolder);
+struct NavigationParamsFetchStateHolder : public JS::Cell {
+    GC_CELL(NavigationParamsFetchStateHolder, JS::Cell);
+    GC_DECLARE_ALLOCATOR(NavigationParamsFetchStateHolder);
 
-public:
-    [[nodiscard]] static GC::Ref<ResponseHolder> create(JS::VM& vm)
+    NavigationParamsFetchStateHolder(OpenerPolicyEnforcementResult&& coop_enforcement_result, URL::URL current_url, GC::Ref<Fetch::Infrastructure::Request> request)
+        : coop_enforcement_result(move(coop_enforcement_result))
+        , current_url(move(current_url))
+        , request(request)
     {
-        return vm.heap().allocate<ResponseHolder>();
     }
 
-    [[nodiscard]] GC::Ptr<Fetch::Infrastructure::Response> response() const { return m_response; }
-    void set_response(GC::Ptr<Fetch::Infrastructure::Response> response) { m_response = response; }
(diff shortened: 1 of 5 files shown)


Libraries/LibWeb/Fetch/Fetching/Fetching.cpp
```diff
@@ -2561,14 +2555,14 @@ WebIDL::ExceptionOr<GC::Ref<PendingResponse>> cors_preflight_fetch(JS::Realm& re
         bool first = true;
         for (auto const& header : headers) {
             if (!first)
-                TRY_OR_THROW_OOM(vm, value.try_append(','));
-            TRY_OR_THROW_OOM(vm, value.try_append(header));
+                MUST(value.try_append(','));
+            MUST(value.try_append(header));
```
trflynn89: Stuff like this should just be `value.append`

Libraries/LibWeb/HTML/TraversableNavigable.cpp
```diff
@@ -652,11 +652,11 @@ TraversableNavigable::HistoryStepResult TraversableNavigable::apply_the_history_
...
-                                 VERIFY(active_window());
-                                 queue_global_task(Task::Source::NavigationAndTraversal, *active_window(), GC::create_function(this->heap(), [after_document_populated, populated_target_entry]() mutable {
-                                     after_document_populated->function()(true, populated_target_entry);
-                                 }));
-                             }));
+                        VERIFY(active_window());
+                        queue_global_task(Task::Source::NavigationAndTraversal, *active_window(), GC::create_function(this->heap(), [after_document_populated, populated_target_entry]() mutable {
+                            after_document_populated->function()(true, populated_target_entry);
+                        }));
+                    }));
```
trflynn89, line 659: This probably should have been in the previous commit?

Libraries/LibWeb/HTML/Navigable.cpp
```diff
@@ -57,30 +57,58 @@ namespace Web::HTML {
...
 GC_DEFINE_ALLOCATOR(Navigable);
 
-class ResponseHolder : public JS::Cell {
-    GC_CELL(ResponseHolder, JS::Cell);
-    GC_DECLARE_ALLOCATOR(ResponseHolder);
+struct NavigationParamsFetchStateHolder : public JS::Cell {
+    GC_CELL(NavigationParamsFetchStateHolder, JS::Cell);
+    GC_DECLARE_ALLOCATOR(NavigationParamsFetchStateHolder);
 
 public:
```
trflynn89: Don't need `public` if it's a `struct`

trflynn89 (commented):
Regarding the first commit, this is an area where dealing with large amounts of memory is easily possible, e.g. downloading files in the range of gigabytes. Maybe crashing the WebContent process is fine here, but also maybe something we also want to explicitly handle with a failed network request?

kalenikaliaksandr (approved):
I found that the change times out on CI because it introduces NavigationObserver usage along the page navigation path. Since we were leaking all NavigationObserver instances, this caused [dump-gc-graph.html](https://github.com/LadybirdBrowser/ladybird/blob/master/Tests/LibWeb/Crash/Internals/dump-gc-graph.html)
 to take an enormous amount of time to complete. I've added a commit that fixes the NavigationObserver leak.

PR #6360, opened 2025-10-01
LibIPC: Delete unused code in Connection

kalenikaliaksandr:


Libraries/LibIPC/Connection.cpp
```diff
@@ -20,8 +20,6 @@ ConnectionBase::ConnectionBase(IPC::Stub& local_stub, NonnullOwnPtr<Transport> t
     , m_transport(move(transport))
     , m_local_endpoint_magic(local_endpoint_magic)
 {
-    m_responsiveness_timer = Core::Timer::create_single_shot(3000, [this] { may_have_become_unresponsive(); });
```
trflynn89, line 23: Shouldn't `may_have_become_unresponsive` (and `did_become_responsive`) also be removed?

trflynn89: The LibCore/Timer.h header should also be removed as it is now unused

PR #6363, opened 2025-10-01
LibWeb: Parse some css-anchor-position properties

tcl3:
This PR adds parsing logic for the `anchor-name`, `position-anchor`, `anchor-try-fallbacks`, `anchor-try-order`, `position-visibility` and `anchor-scope` properties.

Currently nothing is done with any of these properties after parsing them.

Gives us +682 WPT subtests in the `css/css-anchor-position` directory.


Libraries/LibWeb/CSS/Properties.json
```diff
@@ -194,6 +194,18 @@
     "longhands": [],
     "_comment": "The 'longhands' array is populated in the code generator to avoid having to maintain it manually"
   },
+  "anchor-name": {
+    "affects-layout": false,
```
AtkinsSJ: Should this be `true` because changing it could affect which element descendants are positioned relative to? Or am I misunderstanding something?

Libraries/LibWeb/CSS/Properties.json
```diff
@@ -194,6 +194,18 @@
...
+    "animation-type": "discrete",
+    "inherited": false,
+    "initial": "none",
+    "valid-identifiers": [
+      "none"
+    ],
+    "valid-types": [
+      "custom-ident ![none]"
```
AtkinsSJ, line 205: :thonk: It'd probably be good to promote `<dashed-ident>` to its own parsable type at some point, but not right now.

Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
```diff
@@ -4183,6 +4187,20 @@ RefPtr<StyleValue const> Parser::parse_place_self_value(TokenStream<ComponentVal
...
 }
 
+// https://drafts.csswg.org/css-anchor-position/#position-anchor
+RefPtr<StyleValue const> Parser::parse_position_anchor_value(TokenStream<ComponentValue>& tokens)
+{
+    // auto | <anchor-name>
+    if (auto auto_keyword = parse_all_as_single_keyword_value(tokens, Keyword::Auto))
+        return auto_keyword;
+
+    // <anchor-name> = <dashed-ident>
```
AtkinsSJ, line 4226: :thonk: It's weird that the spec defines this type and then sometimes uses `<dashed-ident>` instead. I've created a [spec issue](https://github.com/w3c/csswg-drafts/issues/12895).

Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
```diff
@@ -4201,17 +4204,108 @@ RefPtr<StyleValue const> Parser::parse_position_anchor_value(TokenStream<Compone
...
+RefPtr<StyleValue const> Parser::parse_single_position_try_fallbacks_value(TokenStream<ComponentValue>& tokens)
+{
+    // [<dashed-ident> || <try-tactic>] | <'position-area'>
+    auto transaction = tokens.begin_transaction();
+    tokens.discard_whitespace();
+
+    if (auto position_area = parse_position_area(tokens)) {
+        transaction.commit();
+        return position_area;
+    }
```
AtkinsSJ, line 4255: This isn't quite what the spec says, but I'm also not sure if the spec is mistaken:

`<'position-area'>` means the property's value, which is `none | <position-area>`. But a comma-separated list with `none` in it seems weird to me. Maybe that makes more sense to you as I'm not familiar with the spec. :sweat_smile:

In any case, to match the spec we'd also need to parse `none` here. If the tests don't expect that, maybe it is just a spec bug, and should say `[<dashed-ident> || <try-tactic>] | <position-area>`.

EDIT: Yeah WPT expects no `none` in a comma-separated list, so this is a bug. Please report it. :^)

Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
```diff
@@ -4201,17 +4204,108 @@ RefPtr<StyleValue const> Parser::parse_position_anchor_value(TokenStream<Compone
...
+            break;
+        }
+        tokens.discard_whitespace();
+    }
+
+    StyleValueVector values;
+    if (dashed_ident.has_value())
+        values.append(CustomIdentStyleValue::create(dashed_ident.release_value()));
+    if (try_tactic)
+        values.append(try_tactic.release_nonnull());
```
AtkinsSJ, line 4278: Are we supposed to modify the order at parse-time? That sounds like a "do it at computed value time" thing but I'm not sure.

Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
```diff
@@ -4413,16 +4507,14 @@ RefPtr<StyleValue const> Parser::parse_position_area_value(TokenStream<Component
         return nullptr;
 
     tokens.discard_whitespace();
-    if (!tokens.has_next_token()) {
+    auto second_keyword_value = parse_keyword_value(tokens);
+    if (!second_keyword_value) {
```
AtkinsSJ, line 4537: It's conceivably possible for a `<position-area>` to be followed by an unrelated keyword, in which case we shouldn't consume it, and should just parse the first keyword value. Could you check for that with a transaction here? Maybe it doesn't happen in practice, but this feels like a future footgun.

Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
```diff
@@ -4555,6 +4559,60 @@ RefPtr<StyleValue const> Parser::parse_position_area_value(TokenStream<Component
...
+        }
+        tokens.discard_whitespace();
+    }
+
+    if (anchors_valid_value)
+        values.append(anchors_valid_value.release_nonnull());
+    if (anchors_visible_value)
+        values.append(anchors_visible_value.release_nonnull());
+    if (no_overflow_value)
+        values.append(no_overflow_value.release_nonnull());
```
AtkinsSJ, line 4629: Again, not sure if we should correct the order of these during parsing.

Libraries/LibWeb/CSS/Properties.json
tcl3: I think you're probably right, but because I couldn't verify that it was necessary I've removed it entirely for now.

Libraries/LibWeb/CSS/Properties.json
tcl3: Agreed.

Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
tcl3: There is already an issue for this (https://github.com/w3c/csswg-drafts/issues/12838). I have referenced it in a comment.

Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
tcl3: If the second keyword is not a valid keyword all of the branches below will fail and the `return nullptr` at the bottom of the method will be hit. A transaction has already been started above which is committed by `create_keyword_list()`. 

Not sure if I'm missing something, but I don't think any change is needed here?

Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
tcl3: Existing implementations and the WPT tests expect the keywords to be in this order. FWIW, the `contain` property also behaves like this.

Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
tcl3: Existing implementations reorder the properties so that the dashed ident comes first. The WPT tests also expect this.

Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
AtkinsSJ: I'm imagining some future property that could take a grammar like `<position-area> || wiggle`. We wouldn't want `<position-area>` to consume `wiggle`, just because it happens to be a keyword.

But yeah it's not based on a real situation I think.

Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
```diff
@@ -4201,17 +4204,110 @@ RefPtr<StyleValue const> Parser::parse_position_anchor_value(TokenStream<Compone
...
+    return parse_comma_separated_value_list(tokens, [&](auto& inner_tokens) {
+        return parse_single_position_try_fallbacks_value(inner_tokens);
+    });
+}
+
+RefPtr<StyleValue const> Parser::parse_single_position_try_fallbacks_value(TokenStream<ComponentValue>& tokens)
+{
+    // The grammar here refers to the position-area property, but existing implementations parse the <position-area> value.
+    // See: https://github.com/w3c/csswg-drafts/issues/12838
+    // [<dashed-ident> || <try-tactic>] | <'position-area'>
```
AtkinsSJ: Hehe, good news! The spec just got fixed: https://github.com/w3c/csswg-drafts/commit/dcad1fe0bf640ea308400838c62a113707e2fb68

tcl3: I've updated the relevant comments :tada:

AtkinsSJ (approved):

PR #6366, opened 2025-10-02
LibWeb/CSS: Stop rejecting declarations with vendor-prefixed keywords

AtkinsSJ:
property_accepts_type() only looks at the property itself, not any longhands it might have, so we thought that `font` didn't accept `<custom-ident>`s, and seeing "-apple-..." makes us throw out the declaration even though it's valid as a font name.

We'll reject these like any other unrecognized value if it's somewhere that's not supported, so this was really just an optimization for a rare edge case, and removing the check doesn't have any observable effect except fixing this bug and any similar cases.

Changing property_accepts_type() to look at longhands is not straightforward, as not all longhand values are valid in the shorthand.

----

I am *really* hoping there isn't a font in the wild named `-apple-flibbertigibbet`.


gmta (approved):
Flibbertigibbet approves!

PR #6370, opened 2025-10-02
LibWeb/CSS: Allow setting values on a StylePropertyMap

AtkinsSJ:
With this PR, we can set property values to `CSSKeywordValue`s, `CSSImageValue`s, and any direct `CSSStyleValue`s created from our internal types.

There are >200 imported WPT passes with this, but we barely import any of the per-property tests, so I'm expecting probably several thousand to show up once this is merged, but we'll see!


tcl3:
I ran WPT against these changes and I'm seeing +6394 subtest passes :tada: 

I'm seeing a crash in `css/css-typed-om/the-stylepropertymap/properties/font-style.html`, but I don't think it's directly related to these changes.

AtkinsSJ:
> I ran WPT against these changes and I'm seeing +6394 subtest passes 🎉

*happy dance*

> I'm seeing a crash in `css/css-typed-om/the-stylepropertymap/properties/font-style.html`, but I don't think it's directly related to these changes.

Hmm, looking at it now. That does seem related, I'm probably missing a check to prevent a wrong-typed value being put somewhere.

(Typed-OM is used rarely enough that this might not be a PR blocker, but I'll see if I can fix it.)

AtkinsSJ:
Changes: Insert a commit at the beginning that makes computing `font-style` and `math-depth` safe when they are set from `StylePropertyMap`.

PR #6373, opened 2025-10-02
LibWeb: Fixed typo in comment: fullfilled --> fulfilled

tuxikus:
This is my first contribution to an open source project... 🤓

### DIFF
--- a/Libraries/LibWeb/WebIDL/Promise.cpp
+++ b/Libraries/LibWeb/WebIDL/Promise.cpp
@@ -218,7 +218,7 @@ GC_DEFINE_ALLOCATOR(WaitForAllResults);
 void wait_for_all(JS::Realm& realm, Vector<GC::Ref<Promise>> const& promises, Function<void(Vector<JS::Value> const&)> success_steps, Function<void(JS::Value)> failure_steps)
 {
     // FIXME: Fix spec typo, fullfilled --> fulfilled
-    // 1. Let fullfilledCount be 0.
+    // 1. Let fulfilledCount be 0.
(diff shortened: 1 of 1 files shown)


Libraries/LibWeb/WebIDL/Promise.cpp
```diff
@@ -218,7 +218,7 @@ GC_DEFINE_ALLOCATOR(WaitForAllResults);
 void wait_for_all(JS::Realm& realm, Vector<GC::Ref<Promise>> const& promises, Function<void(Vector<JS::Value> const&)> success_steps, Function<void(JS::Value)> failure_steps)
 {
     // FIXME: Fix spec typo, fullfilled --> fulfilled
-    // 1. Let fullfilledCount be 0.
+    // 1. Let fulfilledCount be 0.
```
trflynn89, line 221: These comments are directly from the spec. We prefer to copy the spec text verbatim, typos and all. As the FIXME comment above this indicates, this should be fixed in the spec itself. After that, we can update our copied spec text.

tuxikus: Thank you for the fast reply and the good explanation. I will look for other ways to contribute to the project.

InvalidUsernameException: The spec is also just a GitHub repository: https://github.com/whatwg/webidl/blob/a79e69ef7398ece546e526a1eebec977363c7304/index.bs#L8714-L8735

So you can open a PR to fix the typo there. And once that is merged, this PR can be merged as well.

PR #6375, opened 2025-10-03
LibWeb: Absolutization simplifications

Calme1709:
This PR includes a couple of simplifications to the `StyleValue` absolutization process:
 - Absolutize all values at the start of `compute_value_of_property`
 - Update `absolutized` to take a single struct as a param rather than multiple individual params.
 - Update some `StyleValue` classes to store their sub-values directly as `StyleValue`s rather than {Percentage,Calculated}Or wrappers

There are also a couple of drive-by improvements and I added `absolutized` methods to a couple of `StyleValue` classes for which it was missing.

### DIFF
--- a/Libraries/LibWeb/CSS/CalculationResolutionContext.h
+++ b/Libraries/LibWeb/CSS/CalculationResolutionContext.h
@@ -9,13 +9,24 @@
 #include <LibWeb/CSS/Angle.h>
 #include <LibWeb/CSS/Frequency.h>
 #include <LibWeb/CSS/Length.h>
+#include <LibWeb/CSS/StyleValues/ComputationContext.h>
 #include <LibWeb/CSS/Time.h>
 
 namespace Web::CSS {
 
 struct CalculationResolutionContext {
-    Variant<Empty, Angle, Frequency, Length, Time> percentage_basis {};
+    using PercentageBasis = Variant<Empty, Angle, Frequency, Length, Time>;
+
+    PercentageBasis percentage_basis {};
     Optional<Length::ResolutionContext> length_resolution_context;
+
+    static CalculationResolutionContext from_computation_context(ComputationContext const& computation_context, PercentageBasis percentage_basis = {})
+    {
+        return {
+            .percentage_basis = percentage_basis,
+            .length_resolution_context = computation_context.length_resolution_context,
+        };
+    }
 };
 
 }
--- a/Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
+++ b/Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
@@ -1934,20 +1934,20 @@ RefPtr<StyleValue const> Parser::parse_border_radius_value(TokenStream<Component
 {
     if (tokens.remaining_token_count() == 2) {
         auto transaction = tokens.begin_transaction();
-        auto horizontal = parse_length_percentage(tokens);
-        auto vertical = parse_length_percentage(tokens);
-        if (horizontal.has_value() && vertical.has_value()) {
+        auto horizontal = parse_length_percentage_value(tokens);
+        auto vertical = parse_length_percentage_value(tokens);
+        if (horizontal && vertical) {
             transaction.commit();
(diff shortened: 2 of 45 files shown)


Libraries/LibWeb/CSS/Parser/ValueParsing.cpp
```diff
@@ -778,9 +778,8 @@ RefPtr<StyleValue const> Parser::parse_integer_value(TokenStream<ComponentValue>
         tokens.discard_a_token(); // integer
         return IntegerStyleValue::create(peek_token.token().number().integer_value());
     }
-    if (auto calc = parse_calculated_value(peek_token); calc
-        && (calc->is_integer() || (calc->is_calculated() && calc->as_calculated().resolves_to_number()))) {
 
+    if (auto calc = parse_calculated_value(peek_token); calc && calc->as_calculated().resolves_to_number()) {
```
AtkinsSJ, line 782: Can we make this contractual by making `RefPtr<CalculatedStyleValue const>` the return type of `parse_calculated_value()` too?

Libraries/LibWeb/CSS/CalculationResolutionContext.h
```diff
@@ -9,13 +9,29 @@
...
 struct CalculationResolutionContext {
-    Variant<Empty, Angle, Frequency, Length, Time> percentage_basis {};
+    using PercentageBasis = Variant<Empty, Angle, Frequency, Length, Time>;
+
+    PercentageBasis percentage_basis {};
     Optional<Length::ResolutionContext> length_resolution_context;
+
+    static CalculationResolutionContext from_computation_context(ComputationContext const& computation_context)
+    {
+        return from_computation_context(computation_context, {});
```
AtkinsSJ: Maybe have a single overload for this with a default value for the percentage_basis?

Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
```diff
@@ -1623,48 +1623,23 @@ RefPtr<StyleValue const> Parser::parse_single_background_size_value(PropertyID p
...
-    auto y_value = maybe_y_value.release_nonnull();
-    auto x_size = get_length_percentage_or_auto(*x_value);
-    auto y_size = get_length_percentage_or_auto(*y_value);
-
-    if (!x_size.has_value() || !y_size.has_value())
-        return nullptr;
-
     transaction.commit();
-    return BackgroundSizeStyleValue::create(x_size.release_value(), y_size.release_value());
+    return BackgroundSizeStyleValue::create(*maybe_x_value, *maybe_y_value);
```
AtkinsSJ: If you want to convert a RefPtr to a NonnullRefPtr, `.release_nonnull()` is best as it avoids some ref-counted bookkeeping.

Libraries/LibWeb/CSS/StyleValues/BackgroundSizeStyleValue.h
```diff
@@ -18,26 +18,26 @@ namespace Web::CSS {
...
 
 private:
-    BackgroundSizeStyleValue(LengthPercentageOrAuto size_x, LengthPercentageOrAuto size_y);
+    BackgroundSizeStyleValue(ValueComparingNonnullRefPtr<StyleValue const> size_x, ValueComparingNonnullRefPtr<StyleValue const> size_y);
 
     struct Properties {
-        LengthPercentageOrAuto size_x;
-        LengthPercentageOrAuto size_y;
+        ValueComparingNonnullRefPtr<StyleValue const> size_x;
+        ValueComparingNonnullRefPtr<StyleValue const> size_y;
```
AtkinsSJ, line 40: IIRC, the idea was to try and make StyleValues held by StyleValues a bit more type-safe, instead of them just being "some kind of StyleValue", but that never really worked and we didn't get much value from it. Seeing how much code you removed here is much nicer. :^)

AtkinsSJ (changes requested):
Nice stuff!

Calme1709:
Implemented feedback and (hopefully) fixed a GCC compile error

AtkinsSJ (approved):

PR #6376, opened 2025-10-03
LibGfx: Free the harfbuzz buffer when measuring text width

Zaggy1024:
I spotted this leak when WebContent was exiting with ASan enabled on a page with a media element. MediaPaintable calls Gfx::Font::width(), which calls through to measure_text_width(), which then drops an hb_buffer_t* without freeing it.

### DIFF
--- a/Libraries/LibGfx/TextLayout.cpp
+++ b/Libraries/LibGfx/TextLayout.cpp
@@ -191,6 +191,7 @@ float measure_text_width(Utf16View const& string, Font const& font, ShapeFeature
     for (size_t i = 0; i < glyph_count; ++i)
         point_x += positions[i].x_advance;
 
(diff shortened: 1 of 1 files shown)


gmta (approved):

trflynn89:
Thinking out loud - given that we have forgotten to free the HB buffer here, and we are also manually freeing it in LibGfx/Font/Font.cpp when the font cache is cleared, it feels like we should be creating an RAII wrapper around that buffer to free it automatically.

awesomekling:
> Thinking out loud - given that we have forgotten to free the HB buffer here, and we are also manually freeing it in LibGfx/Font/Font.cpp when the font cache is cleared, it feels like we should be creating an RAII wrapper around that buffer to free it automatically.

Or even better - make the measurement code use the cache :)

PR #6378, opened 2025-10-03
LibWeb/CSS: More `StylePropertyMap.set()` conversions and some fixes

AtkinsSJ:
Adds/fixes support for setting:
- CSSUnitValues, so all our numbers and percentages and dimensions
- CSS-wide keywords like `initial` or `inherit`
- CSSUnparsedValues, which contain things like `var()`
- Shorthands

Should get us another nice chunk of WPT passes. As noted, CSSUnparsedValue doesn't yet behave as the tests expect, because we strip whitespace when parsing properties. At least now this gives me some proper motivation to fix that. :sweat_smile:

### DIFF
--- a/Libraries/LibWeb/CSS/CSSStyleProperties.cpp
+++ b/Libraries/LibWeb/CSS/CSSStyleProperties.cpp
@@ -419,13 +419,25 @@ WebIDL::ExceptionOr<void> CSSStyleProperties::set_property_style_value(PropertyN
         return {};
     }
 
-    m_properties.remove_first_matching([&property](StyleProperty const& style_property) {
-        return style_property.property_id == property.id();
-    });
-    m_properties.append(StyleProperty {
-        .important = Important::No,
-        .property_id = property.id(),
-        .value = move(style_value),
+    // FIXME: This should have been rejected earlier, but property_accepts_type() is too basic for what we need.
+    if (property_is_positional_value_list_shorthand(property.id())
+        && !style_value->is_shorthand()
+        && !style_value->is_unresolved()
+        && !style_value->is_pending_substitution()
+        && !style_value->is_guaranteed_invalid()
+        && !style_value->is_css_wide_keyword()) {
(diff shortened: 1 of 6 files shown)


AtkinsSJ:
Should be good to go now! I'd estimate around another 1000 WPT passes from that CSS-wide keywords change, and a few hundred from `CSSUnitValue` and shorthand support.

tcl3 (commented):
Running WPT on these changes I'm seeing +1155/-21 subtests. There are new crashes in these
* http://wpt.live/css/css-typed-om/the-stylepropertymap/properties/line-height.html - because `line-height` doesn't support calculated percentages.
* http://wpt.live/css/css-typed-om/the-stylepropertymap/properties/margin.html and http://wpt.live/css/css-typed-om/the-stylepropertymap/properties/logical.html - this crash happens when a value gets assigned to a shorthand property.

The PASS -> FAIL regressions (for example in: http://wpt.live/css/css-typed-om/the-stylepropertymap/properties/font-feature-settings.html) happen because we're not throwing a TypeError when the property gets set to an out of range value.

AtkinsSJ:
Thanks! I'll take a look at those tomorrow.

AtkinsSJ:
> The PASS -> FAIL regressions (for example in: http://wpt.live/css/css-typed-om/the-stylepropertymap/properties/font-feature-settings.html) happen because we're not throwing a TypeError when the property gets set to an out of range value.

Ohhh, this is the same issue as I mentioned for `cursor`: Our data says "this accepts `<number>`", and the typed-om code interprets this as "the entire value can be one `<number>`" when actually it takes `<ident> <number>`. That one's going to need some significant work.

AtkinsSJ:
Changes:
- Rebased on master
- New commit at the beginning which stops us crashing when setting `line-height` to a calculated percentage with typed-om
- Throw a TypeError instead of crashing when setting a single value on a `margin`-like property. This is a bit of a hack, but like with the `cursor`/`font-feature-settings` issues, it requires us to be smarter in a way that's going to be significant work.

AtkinsSJ:
Changes: Rebased to fix a small conflict in "LibWeb/CSS: Support calculated percentages in line-height".

PR #6383, opened 2025-10-03
LibJS: Remove files from .prettierignore

adamrt:
This is housekeeping to remove files from `.prettierignore` and then format those files. The files affected are strictly tests. I took care to not let prettier change the code in ways that affect tests by using `// prettier-ignore`, with comments, selectively.

I used `prettier == 3.6.2` for this work. Previous versions might have issues, but I believe `using` has been supported since `3.0.3` (August 2023) based on the [release notes](https://github.com/prettier/prettier/blob/main/CHANGELOG.md#303). The original comment about `using` was added to `.prettierignore` in January 2023, so this seems logical version-wise.

There are 3 commits to simplify the PR review processes.

- First commit handled low hanging fruit that didn't require any changes other than running prettier against the Tests. this removes 3 out of 4 of the files.
- Second commit is for `using-declaration.js` but only fixes simple changes (single->double quotes and basic formatting changes from prettier)
- **Third commit needs the most review**. This contains some scoping changes for a switch statement to the tests that are required for `prettier+using declarations`. **Best to view this one individually.** The commit messages shows the error that required this change.

Thank you. This is my first commit (not counting SerenityOS) so please let me know of anyways I can improve the PR.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -2,9 +2,6 @@ Libraries/LibJS/Tests/invalid-lhs-in-assignment.js
 Libraries/LibJS/Tests/unicode-identifier-escape.js
 Libraries/LibJS/Tests/modules/failing.mjs
 
-# FIXME: Remove once prettier is updated to support using declarations.
-Libraries/LibJS/Tests/using-declaration.js
-
 Tests/LibWeb/Crash/wpt-import/
 Tests/LibWeb/Ref/input/wpt-import
 Tests/LibWeb/Text/input/wpt-import
--- a/Libraries/LibJS/Tests/using-declaration.js
+++ b/Libraries/LibJS/Tests/using-declaration.js
@@ -307,9 +307,14 @@ describe("works in a bunch of scopes", () => {
         }
 
         switch (2) {
-            case 3:
-                using notDisposed = { [Symbol.dispose]() { expect().fail("not-disposed 1"); } };
-            case 2:
+            case 3: {
+                using notDisposed = {
+                    [Symbol.dispose]() {
+                        expect().fail("not-disposed 1");
+                    },
+                };
+            }
+            case 2: {
                 expect(disposeFull).toEqual([]);
                 using a = pusher("a");
                 expect(disposeFull).toEqual([]);
@@ -319,21 +324,34 @@ describe("works in a bunch of scopes", () => {
                 expect(b.val).toBe("b");
 
                 expect(disposeInner).toBeFalse();
-                // fallthrough
-            case 1: {
-                expect(disposeFull).toEqual([]);
-                expect(disposeInner).toBeFalse();
-
-                using inner = { [Symbol.dispose]() { disposeInner = true; } }
-
-                expect(disposeInner).toBeFalse();
             }
+            // fallthrough
+            case 1:
+                {
+                    expect(disposeFull).toEqual([]);
+                    expect(disposeInner).toBeFalse();
+
+                    using inner = {
+                        [Symbol.dispose]() {
+                            disposeInner = true;
+                        },
+                    };
+
+                    expect(disposeInner).toBeFalse();
+                }
                 expect(disposeInner).toBeTrue();
-                using c = pusher("c");
-                expect(c.val).toBe("c");
+                {
+                    using c = pusher("c");
+                    expect(c.val).toBe("c");
+                }
                 break;
-            case 0:
-                using notDisposed2 = { [Symbol.dispose]() { expect().fail("not-disposed 2"); } };
+            case 0: {
+                using notDisposed2 = {
+                    [Symbol.dispose]() {
+                        expect().fail("not-disposed 2");
+                    },
+                };
+            }
         }
 
         expect(disposeInner).toBeTrue();


Libraries/LibJS/Tests/using-for-loops.js
```diff
@@ -82,11 +82,10 @@ describe("basic usage", () => {
         const obj = {
             [Symbol.dispose]() {
                 disposableCalls++;
-            }
+            },
         };
 
-        for (using a of [obj, obj, obj])
-            expect(disposableCalls).toBe(i++);
+        for (using a of [obj, obj, obj]) expect(disposableCalls).toBe(i++);
```
gmta: When we leave out the braces for a single-line block, we still want the block to be [on its own line](https://github.com/LadybirdBrowser/ladybird/blob/master/Documentation/CodingStyle.md#omission-of-curly-braces-from-statement-blocks) (C++ style, but should also apply to JS imo).

trflynn89: Yeah, this is something `prettier` just does. We have to add braces to keep the lines separated.

Libraries/LibJS/Tests/functions/arrow-functions.js
```diff
@@ -90,6 +90,8 @@ test("currying", () => {
 
 test("with comma operator", () => {
     let foo, bar;
+    // prettier-ignore
+    // Ignored so the expressions arent wrapped in parens
     (foo = bar), baz => {};
```
trflynn89: We prefer our commits to be atomic. So if the purpose of this commit is to remove files from .prettierignore, then the changes to handle differences from prettier 3.6 should be in another commit. Personally, I don't mind if prettier adds parentheses here so long as older prettier versions (namely, the version used in CI) don't complain. We also have the option to update the version of prettier used in CI.

Libraries/LibJS/Tests/object-basic.js
```diff
@@ -27,6 +27,8 @@ describe("correct behavior", () => {
 
         expect(o.foo).toBe("bar");
         expect(o["foo"]).toBe("bar");
+        // prettier-ignore
+        // Ignored so the expressions arent wrapped in parens
         expect(o.qux).toBe(10), expect(o.hello).toBe("friends");
```
trflynn89, line 30: Alternatively, I don't see a reason for these to be on one line - we can just make them separate lines instead of adding a prettier directive

Libraries/LibJS/Tests/using-for-loops.js
adamrt: Tim is right. I didn't make any manual changes to the formatting. These files were just out of sync with prettier formatting since they were being ignored.

Libraries/LibJS/Tests/functions/arrow-functions.js
adamrt: Thank you for the feedback. I assumed the current CI version couldn't handle the 'using' declaration, because of the comment in `.prettierignore`. 

Luckily I was incorrect and when using the CI version `3.3.3`, it doesn't add these extra parentheses. So I will remove these `// prettier-ignore` spots and with `3.3.3` the code remains the same.

If we want to upgrade later we can do the upgrade in CI and update the changes to be more atomic like you suggest.

Libraries/LibJS/Tests/object-basic.js
adamrt: This one was about parentheses, not new lines. But this is moot now that I'm running with prettier `3.3.3`.

Thanks again for the feedback.

Libraries/LibJS/Tests/using-for-loops.js
gmta: I'm not saying this is something prettier doesn't do. I disagree with prettier's formatting for those lines :)

adamrt:
This has been updated to be in sync with the same version of prettier as CI (`3.3.3`).

Libraries/LibJS/Tests/using-for-loops.js
gmta: @adamrt Can you restore the multi-line nature of those collapsed `if` blocks?

adamrt: @gmta Definitely. If the for-clause is long enough, prettier will split them - but not at this length. Since there is no configuration option to split these lines via prettier, I can do either 

- Add brackets around the body of the loop
- Manually change them and add `// prettier-ignore` to each case

I'm assuming you prefer the first option. But wanted to make sure we were in agreement that prettier won't do it without some changes to the code.

Do you have a preference?

gmta: Braces!

Libraries/LibJS/Tests/object-basic.js
trflynn89: Yes, what I meant was that instead of adding parentheses or a prettier-ignore directive, we should instead just split the lines when we update to 3.6 in the future:
```js
expect(o.qux).toBe(10);
expect(o.hello).toBe("friends");
```

trflynn89 (commented):
Per [CONTRIBUTING.md](https://github.com/LadybirdBrowser/ladybird/blob/master/CONTRIBUTING.md#code-submission-policy), please squash the post-review commits into the commits they are addressing.

> Amend your existing commits when adding changes after a review, where relevant.

adamrt:
> Amend your existing commits when adding changes after a review, where relevant.

Thanks for the feedback @trflynn89. I believe it should be all set now. I did a soft reset, amended and then force-pushed. If I can tweak anything else please let me know.

PR #6389, opened 2025-10-04
LibWasm: Fix dumb bug

alimpfard:
Will GCC complain? let's seeeeee


Libraries/LibWasm/AbstractMachine/BytecodeInterpreter.cpp
```diff
@@ -1581,6 +1581,7 @@ HANDLE_INSTRUCTION(memory_fill)
         auto instance = configuration.store().get(address);
         // bounds checked by verifier.
         auto count = configuration.take_source(0, addresses.sources).to<u32>();
+        u8 value = static_cast<u8>(configuration.take_source(1, addresses.sources).to<u32>());
```
gmta, line 1584: Can we have regression test pls? :)

alimpfard: Yep, looking into it :)

alimpfard:
Note for future, test is:
```
Function #0 (stack usage = 3):
  [000] { () -> stack                      }  (i32.const (const 10))
  [001] { () -> stack                      }  (i32.const (const 42))
  [002] { () -> stack                      }  (i32.const (const 5))
  [003] { () -x                            }  (synthetic:expression.end)

Function #1 'test_fill' (stack usage = 3):
  [000]  (call (function index 0))
  [001] { (stack, stack, stack) -x         }  (memory.fill (memory index 0))
  [002] { () -x                            }  (synthetic:expression.end)

Function #2 'get_value' (stack usage = 1):
  [000] { () -> stack                      }  (local.get (local index 0))
  [001] { (stack) -> stack                 }  (i32.load8_u (memory index 0 (align 0) (offset 0)))
  [002] { () -x                            }  (synthetic:expression.end)
```

gmta (approved):

PR #6410, opened 2025-10-07
LibWeb+LibMedia: Rewrite the playback system :^)

Zaggy1024:
This PR replaces the old separate video and audio playback systems with a new system that plays both in perfect sync through PlaybackManager.

Video tracks are decoded off the main thread through VideoDataProvider, which also implements accurate both fast and accurate seeking for its track. The current frame to display is determined by a DisplayingVideoSink, which will check if the frame needs to change when its update function is called.

Like video tracks, audio tracks are also decoded off the main thread through AudioDataProvider, but in this case the only implementing accurate seeking. Fast seeking is only really necessary for video, since audio decodes orders of magnitude faster than video. In order to play back audio, an AudioMixingSink is created, which manages a PlaybackStream which retrieves its data from multiple AudioDataProviders and mixes them into the output.

Video frame timing and the playback position are driven by a MediaTimeProvider class, which has two implementations:
- GenericTimeProvider: Derives playback time from the MonotonicTime class. This is used to drive video when no audio is available.
- WrapperTimeProvider<AudioMixingSink>: Used to wrap the AudioMixingSink in a separate atomically-ref-counted class. This is used when audio _is_ available.

It's worth testing this across as many platforms as possible, especially to make sure that the playback with both audio and video together works well. Some PlaybackStream implementations determine the current device time from the number of samples that have been requested by the underlying API. This means that the audio time may increase in chunks, instead of increasing gradually with real time. This may result in videos appearing to play back at a low frame rate. However, this doesn't appear to be a problem on macOS, so it may work well in other cases as well.

AudioCodecPlugin and its derived classes have been removed, along with all references to Qt Multimedia, as both are unused. The last commit contains these changes, and is worth a check to make sure the build instructions make sense and the build succeeds on all supported platforms.


Lubrsi:
Great work, I tried this branch out and found a few bugs:

1. Soundcloud doesn't update the playback position during play, only on seek
2. The old player used by TCRF crashes with:
```
VERIFICATION FAILED: m_ptr at /Users/lukewilde/Repositories/ladybird/AK/RefPtr.h:280
0   liblagom-ak.0.0.0.dylib             0x0000000103a596a8 dump_backtrace + 72
1   liblagom-ak.0.0.0.dylib             0x0000000103a59960 ak_trap + 12
2   liblagom-ak.0.0.0.dylib             0x0000000103a59a0c ak_assertion_failed + 0
3   liblagom-media.0.0.0.dylib          0x00000001030f2e54 Media::PlaybackManager::get_or_create_the_displaying_video_sink_for_track(Media::Track const&) (.cold.2) + 0
4   liblagom-media.0.0.0.dylib          0x00000001030ce674 Media::PlaybackManager::remove_the_displaying_video_sink_for_track(Media::Track const&) + 276
5   liblagom-web.0.0.0.dylib            0x0000000106ea78b4 Web::HTML::HTMLMediaElement::set_selected_video_track(AK::Badge<Web::HTML::VideoTrack>, GC::Ptr<Web::HTML::VideoTrack>) + 100
6   liblagom-web.0.0.0.dylib            0x0000000106f7d284 Web::HTML::VideoTrack::set_selected(bool) + 440
7   liblagom-web.0.0.0.dylib            0x0000000106ea8114 Web::HTML::HTMLMediaElement::process_media_data(AK::Function<void (AK::String)>) + 1600
8   liblagom-web.0.0.0.dylib            0x0000000106eac1b0 _ZN2AK8FunctionIFvvEE15CallableWrapperIZZZN3Web4HTML16HTMLMediaElement14fetch_resourceERKN3URL3URLENS0_IFvNS_6StringEEEEEN3$_0clIN2GC3RefINS4_5Fetch14Infrastructure8ResponseEEEEEDaT_ENUlNS_6Detail10ByteBufferILm32EEEE_clESQ_EUlvE_E4callEv + 68
9   liblagom-web.0.0.0.dylib            0x0000000106e39bf0 Web::HTML::Task::execute() + 92
10  liblagom-web.0.0.0.dylib            0x0000000106e37258 Web::HTML::EventLoop::process() + 84
11  liblagom-web.0.0.0.dylib            0x00000001070a08b0 AK::Function<void ()>::CallableWrapper<Web::Platform::TimerSerenity::TimerSerenity()::$_0>::call() + 100
12  liblagom-core.0.0.0.dylib           0x000000010358c688 Core::Timer::timer_event(Core::TimerEvent&) + 208
13  liblagom-core.0.0.0.dylib           0x000000010358b648 Core::ThreadEventQueue::process() + 424
14  liblagom-core.0.0.0.dylib           0x00000001035900e8 Core::EventLoopImplementationUnix::exec() + 44
15  liblagom-core.0.0.0.dylib           0x000000010357dcc0 Core::EventLoop::exec() + 68
16  WebContent                          0x0000000102d28f20 ladybird_main(Main::Arguments) + 4208
17  WebContent                          0x0000000102dcd770 main + 184
18  dyld                                0x0000000181f8dd54 start + 7184
```
3. Seeking can sometimes (I struggled in the clip) report an old time value and increment from there
4. JS doesn't seem to realise that media has finished (I don't remember which event it is) so it doesn't go to the next track and the playback position goes beyond the end (which has a funny visual glitch on Bandcamp of the pill going past the end)

Here's the clip of me demonstrating these bugs:

https://github.com/user-attachments/assets/958cfa1d-6df4-4a1d-a05c-b6ead4481e4a

Zaggy1024:
> 2\. The old player used by TCRF crashes with:

This should be fixed now!

Zaggy1024:
> Great work, I tried this branch out and found a few bugs:
>
> 1. Soundcloud doesn't update the playback position during play, only on seek
>
> ...
>
> 3. Seeking can sometimes (I struggled in the clip) report an old time value and increment from there
>
> 4. JS doesn't seem to realise that media has finished (I don't remember which event it is) so it doesn't go to the next track and the playback position goes beyond the end (which has a funny visual glitch on Bandcamp of the pill going past the end)

These should be fixed now as well!

Lubrsi:
Found a bug on Shopify's landing page where their WebMs don't play, even if you show controls and manually seek:
Screenshot 2025-10-09 at 16 14 14

Documentation/BuildInstructionsLadybird.md
```diff
@@ -67,34 +64,34 @@ sudo apt update && sudo apt install g++-14 libstdc++-14-dev
 
 #### Audio support:
 
-- Recommendation: Install PulseAudio development package:
+- Install PulseAudio development package:
 
 ```bash
 sudo apt install libpulse-dev
 ```
```
mo271, line 71: perhaps this could just be moved to the `apt` command at the top, now?

Zaggy1024: I was considering that, but it is still optional (in theory, on some level), so I'm unsure if there might be situations in which people may want to be aware that it isn't absolutely necessary. Also, I'm not sure if there may be other packages providing the headers.

Definitely, if the consensus is that it's never a problem to install these headers, absolutely, it could go in the main list.

Libraries/LibMedia/TimedImage.h
```diff
@@ -0,0 +1,51 @@
...
+    }
+    TimedImage() = default;
+
+    bool is_valid() const
+    {
+        return m_image != nullptr;
+    }
+    AK::Duration const& timestamp() const
+    {
+        VERIFY(is_valid());
```
gmta, line 31: Sort of an interesting API choice: we lose the ability to get the timestamp when `release_image()` was invoked? Maybe this doesn't need to be that strict?

Zaggy1024: The idea is that TimedImage has a default state, and returns to that when the frame is retrieved. DisplayingVideoSink uses release_image() after it's reached the timestamp and no longer needs it, so to ensure that we never try to check the timestamp after that has happened and before a new frame is emplaced, I added that assertion. Perhaps there's a better way to set this up while still also allowing us to avoid the cost of an atomic ref increment?

Libraries/LibMedia/Providers/VideoDataProvider.h
```diff
@@ -0,0 +1,92 @@
...
+    ~VideoDataProvider();
+
+    void set_error_handler(ErrorHandler&&);
+
+    TimedImage retrieve_frame();
+
+    void seek(AK::Duration timestamp);
+
+private:
+    class ThreadData final : public AtomicRefCounted<ThreadData> {
```
R-Goc, line 52: Not fully sure about the naming here. I see ffmpeg also uses the ThreadData name for this kind of class, but in ladybird we already have a ThreadData class name where we use it for thread local data. That code is in LibCore/LibWebView/UI for the event loop implementations. This might be fine being in LibMedia though.

Libraries/LibMedia/Providers/VideoDataProvider.cpp
```diff
@@ -0,0 +1,220 @@
...
+}
+
+void VideoDataProvider::ThreadData::push_data_and_decode_some_frames()
+{
+    // FIXME: Check if the PlaybackManager's current time is ahead of the next keyframe, and seek to it if so.
+    //        Demuxers currently can't report the next keyframe in a convenient way, so that will need implementing
+    //        before this functionality can exist.
+
+#if PLAYBACK_MANAGER_DEBUG
+    auto start_time = MonotonicTime::now();
```
R-Goc: This is unused.

Libraries/LibMedia/Sinks/DisplayingVideoSink.cpp
```diff
@@ -0,0 +1,77 @@
...
+}
+
+DisplayingVideoSink::DisplayingVideoSink(NonnullRefPtr<MediaTimeProvider> const& time_provider)
+    : m_time_provider(time_provider)
+{
+}
+
+DisplayingVideoSink::~DisplayingVideoSink() = default;
+
+void DisplayingVideoSink::verify_track(Track const& track) const
```
R-Goc, line 27: Maybe a verify_track fucntion that returns an ErrorOr<void> and doesn't VERIFY internally would be better instead. That way the caller can decide if they want to crash by either using a MUST or otherwise handling the error.

Libraries/LibMedia/Providers/VideoDataProvider.h
Zaggy1024: It's already a pattern in PlaybackStream, and it's an unexposed implementation detail, so hopefully it won't matter much.

Libraries/LibMedia/Providers/VideoDataProvider.cpp
Zaggy1024: Good catch, that's leftover from the old PlaybackManager code I think. :sweat_smile:

Libraries/LibMedia/Sinks/DisplayingVideoSink.cpp
Zaggy1024: This is just a private function to ensure that any using code is only ever expecting the DisplayingVideoSink to work with one data provider at a time, there shouldn't be any case where we need to choose to return an error rather than crashing (currently, anyway).

AK/Time.h
```diff
@@ -213,6 +213,16 @@ class Duration {
...
 public:
     [[nodiscard]] constexpr static Duration from_seconds(i64 seconds) { return Duration(seconds, 0); }
+    [[nodiscard]] constexpr static Duration from_seconds_f64(f64 seconds)
+    {
+        seconds = AK::max(seconds, static_cast<f64>(NumericLimits<i64>::min()));
+        seconds = AK::min(seconds, static_cast<f64>(NumericLimits<i64>::max()));
+        i64 integer_seconds = static_cast<i64>(seconds);
+        if (static_cast<f64>(integer_seconds) > seconds)
+            integer_seconds--;
+        u32 integer_nanoseconds = static_cast<u32>((seconds - static_cast<f64>(integer_seconds)) * 1'000'000'000);
```
gmta, line 227: There's some potential issues with the conversions happening here, e.g. NaNs get through or when `seconds` is large enough, it might not be enough to just do `integer_seconds--` since the difference with `seconds` might be larger than 1. This also could cause a negative value for `seconds - static_cast<f64>(integer_seconds)` resulting in an underflow when casting to `u32`. You might want to differentiate between the scenarios where the integral part of `seconds` is fully represented and where floating point imprecisions might turn up.

At any rate: please add a test case for this method so it's easier to expose and fix bugs in the future :)

Libraries/LibMedia/Containers/Matroska/MatroskaDemuxer.cpp
```diff
@@ -66,7 +68,15 @@ static TrackType track_type_from_matroska_track_type(TrackEntry::TrackType type)
 
 static Track track_from_track_entry(TrackEntry const& track_entry)
 {
-    Track track(track_type_from_matroska_track_type(track_entry.track_type()), track_entry.track_number());
+    auto name = Utf16String::from_utf8(track_entry.name());
```
gmta, line 70: At this point `name()` should return valid UTF-8, so you can use `::from_utf8_without_validation()`.

Libraries/LibMedia/Sinks/AudioMixingSink.cpp
```diff
@@ -0,0 +1,333 @@
...
+
+        VERIFY(format == Audio::PcmSampleFormat::Float32);
+        VERIFY(!Checked<i64>::multiplication_would_overflow(sample_count, channel_count));
+        auto float_buffer_count = static_cast<i64>(sample_count) * channel_count;
+        auto float_buffer_size = float_buffer_count * sizeof(float);
+        VERIFY(buffer.size() >= float_buffer_size);
+        auto* float_buffer = reinterpret_cast<float*>(buffer.data());
+
+        for (i64 i = 0; i < float_buffer_count; i++)
+            float_buffer[i] = 0;
```
gmta: Consider making a `Span<float>` from the buffer so you can do things like `.fill(0.f)`.

Libraries/LibMedia/AudioBlock.h
```diff
@@ -0,0 +1,63 @@
...
+        return m_sample_rate == 0;
+    }
+    i64 data_count() const
+    {
+        return static_cast<i64>(data().size());
+    }
+    i64 sample_count() const
+    {
+        return data_count() / m_channel_count;
+    }
```
gmta, line 53: Should `data_count()` and `sample_count()` ever return negative values? Seems like we should make these `size_t` and let the callers deal with any unfortunate conversions that need to happen.

Libraries/LibWeb/HTML/MediaTrackBase.cpp
```diff
@@ -0,0 +1,57 @@
...
+        //     not available or cannot be determined.  Omitting the language tag (where permitted) is preferred.  The 'und'
+        //     subtag might be useful for protocols that require a language tag to be provided or where a primary language
+        //     subtag is required (such as in "und-Latn").  The 'und' subtag MAY also be useful when matching language tags
+        //     in certain situations.
+        //
+        // Matroska's TrackEntry->Language element is required, and will use "und" as a placeholder as mentioned above. We
+        // don't want to return anything when that placeholder is found:
+        if (language == "und")
+            return Utf16String();
+        return Utf16String::from_utf8(language);
```
gmta: No need to validate UTF-8 here

Libraries/LibWeb/Page/Page.cpp
```diff
@@ -502,6 +502,16 @@ void Page::unregister_media_element(Badge<HTML::HTMLMediaElement>, UniqueNodeID
     });
 }
 
+void Page::update_all_media_element_video_sinks()
+{
+    for (auto media_id : m_media_elements) {
+        if (auto* node = DOM::Node::from_unique_id(media_id)) {
+            auto& media_element = as<HTML::HTMLMediaElement>(*node);
```
gmta: This seems to be a recurring pattern for Page; maybe create a helper `for_each_media_element`?

AK/Time.h
Zaggy1024: The `integer_seconds--` is here to account for truncation towards 0 when `seconds < 0`, I think there shouldn't be any cases where the difference is greater than 1.

Good point on the NaNs, though, I've added a `VERIFY(!isnan(seconds))`, along with some tests for NaN, normal values, as well as some values approaching the numeric limits involved.

Zaggy1024:
Rebased so that TestVorbisDecode can hopefully pass on Windows.

gmta:
Fun fact: this fixes the firing sound effects on https://playbiolab.com/ which makes this one of the best PRs ever!

Libraries/LibWeb/HTML/HTMLMediaElement.cpp
```diff
@@ -1619,6 +1626,12 @@ WebIDL::ExceptionOr<void> HTMLMediaElement::play_element()
         //    notify about playing for the element.
         else {
             notify_about_playing();
+
+            // AD-HOC: If the official playback position is at the end of the media data here, that means that we haven't
+            //         run the reached the end of media playback steps. This can happen if we seeked to the end while paused
+            //         and looping.
```
gmta: Issue / PR link

Libraries/LibWeb/HTML/HTMLMediaElement.cpp
```diff
@@ -267,17 +267,22 @@ double HTMLMediaElement::current_time() const
...
     // position to the new value; otherwise, it must set the official playback position to the new value and then seek to the new
     // value. The new value must be interpreted as being in seconds.
     if (m_ready_state == ReadyState::HaveNothing) {
         m_default_playback_start_position = current_time;
     } else {
-        m_official_playback_position = current_time;
+        // AD-HOC: Don't set the official playback position here, as seek_element() will set it according to the seekable
+        //         ranges. We return the value passed to the setter to ensure that chained assignments like
+        //             videoA.currentTime = videoB.currentTime = Number.MAX_VALUE;
+        //         will seek both videoA and videoB to their corresponding ending positions.
```
gmta: Issue / PR link

PR #6412, opened 2025-10-07
LibWeb: Media query improvements

Calme1709:
A collection of improvements related to media queries, see individual commits for details.


AtkinsSJ (approved):
Very nice! Good catch with the window/document thing, I think you're right that we should always resolve lengths against the document not the window.

PR #6415, opened 2025-10-07
LibWeb: Exclude namespaced attributes from getElementsByClassName()

benjamintory:
Fixes at lease one wpt test https://wpt.live/dom/nodes/getElementsByClassName-11.xml

Previously, XML documents incorrectly treated attributes like `t:class` as regular `class` attributes. This happened because the `localName` was compared without considering the attribute's namespace.

In HTML documents, the local name includes the namespace implicitly, but in XML, the namespace is stored separately. This change ensures that only non-namespaced `class` attributes affect `m_classes`.

Based on the DOM specification behavior (search for "if localName is id") https://dom.spec.whatwg.org/


Libraries/LibWeb/DOM/Element.cpp
```diff
@@ -3780,7 +3780,7 @@ void Element::attribute_changed(FlyString const& local_name, Optional<String> co
 
     auto value_or_empty = value.value_or(String {});
 
-    if (local_name == HTML::AttributeNames::id) {
+    if (local_name == HTML::AttributeNames::id && !namespace_.has_value()) {
```
gmta, line 3783: Wouldn't this need to consider Web::Namespace::HTML as well?

benjamintory: I don't think it does, as I understand it for html, local_name has a prefix and then attribute type like "t:class" so checking "t:class" == "class" is already enough.

For xml is doesn't seem to have a prefix here so local_name is always just attribute type like "class" and then namespace is the namespace.

XML, logging localname and namespace
```
33175.636 WebContent(145954): localname: xmlns, namespace: http://www.w3.org/2000/xmlns/
33175.636 WebContent(145954): localname: g, namespace: http://www.w3.org/2000/xmlns/
33175.636 WebContent(145954): localname: h, namespace: http://www.w3.org/2000/xmlns/
33175.636 WebContent(145954): localname: t, namespace: http://www.w3.org/2000/xmlns/
33175.636 WebContent(145954): localname: src, namespace: None
33177.151 WebContent(145954): localname: src, namespace: None
33177.796 WebContent(145954): localname: id, namespace: None
33177.796 WebContent(145954): localname: id, namespace: None
33177.796 WebContent(145954): localname: class, namespace: None
33177.797 WebContent(145954): localname: class, namespace: None
33177.797 WebContent(145954): localname: class, namespace: http://tc.labs.opera.com/#test
33177.797 WebContent(145954): localname: class, namespace: http://www.w3.org/1999/xhtml
33177.797 WebContent(145954): localname: class, namespace: http://www.w3.org/2000/svg
33177.797 WebContent(145954): localname: class, namespace: http://tc.labs.opera.com/#test
33177.797 WebContent(145954): localname: class, namespace: http://www.w3.org/1999/xhtml
33177.797 WebContent(145954): localname: class, namespace: http://www.w3.org/2000/svg
33177.797 WebContent(145954): localname: class, namespace: None
33177.797 WebContent(145954): localname: class, namespace: http://tc.labs.opera.com/#test
33177.797 WebContent(145954): localname: class, namespace: http://www.w3.org/1999/xhtml
33177.797 WebContent(145954): localname: class, namespace: http://www.w3.org/2000/svg
```
local_name is always just class,id, etc. so we need to check namespace


HTML,  logging localname and namespace
```
33287.005 WebContent(145996): localname: xmlns:h, namespace: None
33287.005 WebContent(145996): localname: xmlns:t, namespace: None
33287.005 WebContent(145996): localname: id, namespace: None
33287.005 WebContent(145996): localname: class, namespace: None
33287.005 WebContent(145996): localname: class, namespace: None
33287.005 WebContent(145996): localname: t:class, namespace: None
33287.005 WebContent(145996): localname: h:class, namespace: None
33287.005 WebContent(145996): localname: g:class, namespace: None
```
local_name contains the prefix for each namespace so when it is compared against class,id, etc. it doesn't match

This is my first commit and this is my basic understanding of the project so I could be wrong. I guess checking Web::Namespace::HTML wouldn't hurt but I'm not sure that it is necessary
Thankyou

PR #6418, opened 2025-10-07
LibWeb: Implement scoped custom element registries

AtkinsSJ:
This has been hanging around in a branch since April, and I don't like the idea of it going to waste. There are some problems:
- Some test regressions (but significantly less than there are new passes, I believe)
- Some tests are skipped because they're flaky.
- It's a bit disorganised. This is one big change that I did try and split up a bit, but later gave up on because it's too interconnected. So I'm sorry about that. :sweat_smile: 

It's also debatable whether "import all the tests" is a good idea. The first commit is 15731 lines of just tests. I can trim that down if we want.

I have a pretty poor memory of what exactly I was doing. I've updated to include some more recent spec changes - mostly in that final commit. This being new code, maybe I should try to move those changes into the original commits adding those algorithms, but I didn't want to mess with them too much. :sweat_smile:

So yeah, it's a big chunky PR that doesn't work entirely correctly. But I figure it's better in-tree where it can be iterated on, than bit-rotting in a branch on my machine. Scoped custom element registries are only implemented in WebKit so far, so there's a decent chance that the spec will still change before it's finalised. I did come across some [issues](https://github.com/whatwg/dom/issues/1379) in the past, some still outstanding. So I'm not entirely concerned about us not passing all the tests yet.

### DIFF
--- a/Libraries/LibWeb/DOM/Document.cpp
+++ b/Libraries/LibWeb/DOM/Document.cpp
@@ -2340,20 +2340,27 @@ WebIDL::ExceptionOr<GC::Ref<Node>> Document::import_node(GC::Ref<Node> node, Var
     // 3. Let registry be null.
     GC::Ptr<HTML::CustomElementRegistry> registry;
 
-    options.visit(
+    TRY(options.visit(
         // 4. If options is a boolean, then set subtree to options.
-        [&subtree](bool const& value) {
+        [&subtree](bool const& value) -> WebIDL::ExceptionOr<void> {
             subtree = value;
+            return {};
         },
         // 5. Otherwise:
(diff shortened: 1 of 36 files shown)


AtkinsSJ:
Looks like there's still some flakiness and crashing in the tests I added. I'll draft this for now and take a look later.

AtkinsSJ:
After some discussion, I'm closing this for now. SCER is still only implemented in WebKit, and until Blink or Gecko do so, the spec is likely to continue to change. (And no websites are going to use it.) I'll try not to let the branch bitrot too much but it can wait until I can write a more positive PR description than "this thing is big and doesn't really work." :laughing:

Issue #6362, opened 2025-10-01
Failing LibWeb text test abortsignal-timeout.html


rcorsi:
### Summary

Randomly the Text/input/abortsignal-timeout.html test in LibWeb fails.

Running LibWeb test 100 times locally on a Virtualbox VM with Ubuntu 25.04, I had it fail 8 times.

I have similar results on real hardware, fail between 5-8% of the time.

When it fails, all we have is the following
```
Time passed before abort event fired is at least 10 milliseconds: false
Reason type: TimeoutError
onabort event isTrusted: true
```

I modified the test Text/input/abortsignal-timeout.html to expose more information and when it fails we have the following.

```
Time passed before abort event fired is at least 10 milliseconds: false
Reason type: TimeoutError
onabort event isTrusted: true
start time: 14.399999998509884
abort time: 24
difference: 9.600000001490116
Math.ceil difference: 10
```

```
Time passed before abort event fired is at least 10 milliseconds: false
Reason type: TimeoutError
onabort event isTrusted: true
start time: 7.199999999254942
abort time: 15.899999998509884
difference: 8.699999999254942
Math.ceil difference: 9
```

```
Time passed before abort event fired is at least 10 milliseconds: false
Reason type: TimeoutError
onabort event isTrusted: true
start time: 11.5
abort time: 20.099999997764826
difference: 8.599999997764826
Math.ceil difference: 9
```

```
Time passed before abort event fired is at least 10 milliseconds: false
Reason type: TimeoutError
onabort event isTrusted: true
start time: 17.900000002235174
abort time: 25.100000001490116
difference: 7.199999999254942
Math.ceil difference: 8
```

```
Time passed before abort event fired is at least 10 milliseconds: false
Reason type: TimeoutError
onabort event isTrusted: true
start time: 9.399999998509884
abort time: 19.199999999254942
difference: 9.800000000745058
Math.ceil difference: 10
```

### Operating system

Linux

### Steps to reproduce

1. Run LibWeb test repeatedly, like 100 times
2. Check output for "^Fail: " output
3. Note the number of failures of Text/input/abortsignal-timeout.html

or use the reduced test case as a .html file in Ladybird browser, to repeatedly run the script code and see the issue interactively.

### Expected behavior

It shouldn't fail at all.

### Actual behavior

Fails randomly. 5-8% of the time on my systems.

### URL for a reduced test case

none

### HTML/SVG/etc. source for a reduced test case

[...]

PR #6422, opened 2025-10-07
Tests: Adjust timing in abortsignal-timeout test to avoid failure

rcorsi:
Many clock related functions in Linux (and Ladybird) use the clock type called CLOCK_MONOTONIC_COARSE to obtain timestamps. This is a less precise clock and can be skewed by 1 or more milliseconds. Often coarse clocks are used to defend against timing attacks or for fingerprinting a system.

This less precise clock timing is causing the abortsignal-timeout.html test to fail randomly as it expects the abort signal to arrive after 10ms, but in some cases it arrives at 9ms causing the test to fail. In very rare cases it could even arrive in 7ms or 8ms.

Test is changed to accept 9ms as a good result and if that is still not enough the test also will display the timings so that further adjustments to the test can be made in the future.

Issue discussed further in #6362

fixes: #6362

### DIFF
--- a/Tests/LibWeb/Text/input/abortsignal-timeout.html
+++ b/Tests/LibWeb/Text/input/abortsignal-timeout.html
@@ -7,10 +7,16 @@
         const signal = AbortSignal.timeout(timeout_milliseconds);
         signal.onabort = (event) => {
             const abort_event_time = performance.now();
-            const time_taken_milliseconds = abort_event_time - test_start_time;
+            const time_taken_milliseconds = (abort_event_time - test_start_time) + 1; // Add 1ms for possible coarse clock skew
             println(`Time passed before abort event fired is at least ${timeout_milliseconds} milliseconds: ${time_taken_milliseconds >= timeout_milliseconds}`);
             println(`Reason type: ${signal.reason.name}`);
             println(`onabort event isTrusted: ${event.isTrusted}`);
+            if (time_taken_milliseconds < timeout_milliseconds) {
+               println(`start time: ${test_start_time}`);
+               println(`abort time: ${abort_event_time}`);
+               println(`difference (coarse adjusted): ${time_taken_milliseconds}`);
+               println(`difference (actual): ${time_taken_milliseconds - 1}`);
+            }
             done();
         };
     });


Tests/LibWeb/Text/input/abortsignal-timeout.html
```diff
@@ -7,10 +7,16 @@
         const signal = AbortSignal.timeout(timeout_milliseconds);
         signal.onabort = (event) => {
             const abort_event_time = performance.now();
```
gmta, line 9: Reading through #6362, I think it's important to consider that `performance.now()` uses a high resolution time source (not coarse) and coarsens it by 5 or 100 _microseconds_ depending on whether cross-origin isolation comes into play.

It then seems like `AbortSignal.timeout()` is the offender here. I'd expect a timeout signal to always take _at least_ the specified amount of time, but never less than that. Do other browsers trigger the timeout earlier than 10ms?

rcorsi: On real hardware with Ubuntu 25.04 results, loading the test html file in the browser and repeatedly hitting reload button.

Chromium, solid 10ms
Firefox, solid, 9, 10, or 11ms, so it would fail sometimes.
Ladybird, variable, 7,8,9,10,11...all the way to 35 ms.

gmta: Firefox' 9ms is something I could understand given that `performance.now()` returns milliseconds, and unfortunate rounding could come into play together with the 'coarsened' time. For example, 995.5 -> 1005.4 (9.9ms) would round to 996 -> 1005 (9ms).

gmta: Adjusting for unfortunate rounding (e.g. time + 1) should be fine I guess? But seeing 8ms or 7ms would really surprise me and hints at an issue with the way we trigger that timeout imo.

rcorsi: Strangely on Firefox, all the numbers appear to be integers, no decimal points visible in any printouts. I found it curious.

rcorsi: I'll have a look at the timeout trigger, to see why 7ms might be happening.

gmta: > Strangely on Firefox, all the numbers appear to be integers, no decimal points visible in any printouts. I found it curious.

Isn't that supposed to happen? `performance.now()` returns milliseconds, with millisecond precision?

rcorsi: I just found curious that Firefox is automatically rounding to the millisecond in the printouts, but Chromium and Ladybird don't.  All numbers in javascript are floating point double precision.

A few samples of each from my virtualbox vm Ubuntu 25.04:

Firefox

```
Time passed before abort event fired is at least 10 milliseconds: true
Reason type: TimeoutError
onabort event isTrusted: true
start time: 12
abort time: 21
difference (coarse adjusted): 10
difference (actual): 9

Time passed before abort event fired is at least 10 milliseconds: true
Reason type: TimeoutError
onabort event isTrusted: true
start time: 9
abort time: 19
difference (coarse adjusted): 11
difference (actual): 10

Time passed before abort event fired is at least 10 milliseconds: true
Reason type: TimeoutError
onabort event isTrusted: true
start time: 7
abort time: 18
difference (coarse adjusted): 12
difference (actual): 11
```

Chromium

```
Time passed before abort event fired is at least 10 milliseconds: true
Reason type: TimeoutError
onabort event isTrusted: true
start time: 12
abort time: 22.299999997019768
difference (coarse adjusted): 11.299999997019768
difference (actual): 10.299999997019768

Time passed before abort event fired is at least 10 milliseconds: true
Reason type: TimeoutError
onabort event isTrusted: true
start time: 15.200000002980232
abort time: 26.5
difference (coarse adjusted): 12.299999997019768
difference (actual): 11.299999997019768

Time passed before abort event fired is at least 10 milliseconds: true
Reason type: TimeoutError
onabort event isTrusted: true
start time: 17.899999998509884
abort time: 28.799999997019768
difference (coarse adjusted): 11.899999998509884
difference (actual): 10.899999998509884
```

Ladybird

```
Time passed before abort event fired is at least 10 milliseconds: true
Reason type: TimeoutError
onabort event isTrusted: true
start time: 16
abort time: 25.700000002980232
difference (coarse adjusted): 10.700000002980232
difference (actual): 9.700000002980232

Time passed before abort event fired is at least 10 milliseconds: true
Reason type: TimeoutError
onabort event isTrusted: true
start time: 5.199999995529652
abort time: 15.599999994039536
difference (coarse adjusted): 11.399999998509884
difference (actual): 10.399999998509884

Time passed before abort event fired is at least 10 milliseconds: true
Reason type: TimeoutError
onabort event isTrusted: true
start time: 7.800000004470348
abort time: 19.400000005960464
difference (coarse adjusted): 12.600000001490116
difference (actual): 11.600000001490116
```

gmta: FYI, looks like `WindowOrWorkerGlobalScopeMixin::run_steps_after_a_timeout_impl()` is barely implemented to spec. It looks like it should wait at least the given amount of time, which we clearly do not do at this point. That's probably our current timer's implementation fault?

At any rate, the test seems correct, we seem to be missing a correct implementation.

rcorsi: Yeah, I also noticed that several steps are missing in the implementation with 7 FIXMEs!

I did do some testing with a timeout of 100ms and 1000ms, just to see what the difference would be and seem like we are always very close to the expected value, just sometimes we undershoot by a 1ms.

rcorsi: Should I look into anything else? It seems safe to merge the test. Should improve the reliability of the CI a bit.

gmta: Could you add a comment to the test explaining why you are doing the `+ 1` there? It might also be worthwhile to open a new GitHub issue referencing the timeout implementation (and lack thereof) referencing this flaky test.

rcorsi: I've update the test by adding a comment at the end of the line where the 1ms second change is implemented? Hope that is where you expected it.

PR #6425, opened 2025-10-09
LibWeb: Parse and interpolate `corner-*-shape` properties

Calme1709:

Gains us ~730 WPT passes.


Calme1709:
Changes: Fixed crash when interpolating superellipse with calculated parameter.

PR #6426, opened 2025-10-09
LibWeb: Add support for tree-counting functions

Calme1709:
This PR adds support for the two CSS tree-counting functions `sibling-count()` and `sibling-index()`

There are two places we support these functions:
1. Top-level value contexts: we allow these functions anywhere that we parse `<number>` or `<integer>`, this works by parsing as a `TreeCountingFunctionStyleValue` and then converting that to the appropriate `NumberStyleValue` or `IntegerStyleValue` within `absolutized`
2. Within `calc()`: We parse this to a new `NonMathFunctionCalculationNode` which we resolve in the appropriate place in `simplify_a_calculation_tree`.

The tricky part of this is invalidating elements which rely on these tree counting functions when their sibling-count or index changes. The current system I have is:
1. Mark an element as relying on tree counting functions if we encounter one during `StyleValue` absolutization.
2. When an element has it's children (i.e. it's children's siblings) changed we iterate over each child and check whether it relies on tree counting functions, if it does we mark it as needing a style update.

This works but isn't ideal as:
1. We pass an out parameter (`PropertyComputationDependencies`) through the entire property value computation callstack (including `simplify_a_calculation_tree`) just to keep track of whether we encountered a tree counting function
2. We have to iterate over all of an element's children whenever one is inserted/removed.

Suggestions on a better approach for invalidation would be very welcome

### DIFF
--- a/Libraries/LibWeb/CSS/Parser/ValueParsing.cpp
+++ b/Libraries/LibWeb/CSS/Parser/ValueParsing.cpp
@@ -4188,7 +4188,7 @@ RefPtr<CalculationNode const> Parser::parse_a_calc_function_node(Function const&
     if (auto maybe_function = parse_math_function(function, context)) {
         // NOTE: We have to simplify manually here, since parse_math_function() is a helper for calc() parsing
         //       that doesn't do it directly by itself.
-        return simplify_a_calculation_tree(*maybe_function, context, CalculationResolutionContext {});
+        return simplify_a_calculation_tree(*maybe_function, context, CalculationResolutionContext {}, nullptr);
     }
 
     return nullptr;
@@ -4468,7 +4468,7 @@ RefPtr<CalculationNode const> Parser::parse_a_calculation(Vector<ComponentValue>
         return nullptr;
 
     // 6. Return the result of simplifying a calculation tree from values.
-    return simplify_a_calculation_tree(*calculation_tree, context, CalculationResolutionContext {});
+    return simplify_a_calculation_tree(*calculation_tree, context, CalculationResolutionContext {}, nullptr);
 }
 
 // https://drafts.csswg.org/css-values-5/#tree-counting
--- a/Libraries/LibWeb/CSS/StyleComputer.cpp
+++ b/Libraries/LibWeb/CSS/StyleComputer.cpp
@@ -895,6 +895,16 @@ static void cascade_custom_properties(DOM::AbstractElement abstract_element, Vec
     custom_properties.update(important_custom_properties);
 }
 
+// TODO: This could be useful for marking which properties need recomputation when inherited style changes
+ALWAYS_INLINE static void mark_property_computation_dependencies(Optional<DOM::AbstractElement> abstract_element, PropertyComputationDependencies const& property_computation_dependencies)
+{
+    if (!abstract_element.has_value())
+        return;
+
+    if (property_computation_dependencies.tree_counting_function)
+        abstract_element->element().set_style_uses_tree_counting_function();
+}
+
 void StyleComputer::collect_animation_into(DOM::AbstractElement abstract_element, GC::Ref<Animations::KeyframeEffect> effect, ComputedProperties& computed_properties) const
 {
     auto animation = effect->associated_animation();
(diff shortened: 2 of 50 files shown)


Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -3411,22 +3411,13 @@ NonnullRefPtr<StyleValue const> StyleComputer::compute_font_style(NonnullRefPtr<
...
-        return specified_value;
-
-    if (angle_value->is_angle())
-        return FontStyleStyleValue::create(specified_value->as_font_style().font_style(), AngleStyleValue::create(Angle::make_degrees(angle_value->as_angle().angle().to_degrees())));
-
-    if (angle_value->is_calculated())
-        return FontStyleStyleValue::create(specified_value->as_font_style().font_style(), AngleStyleValue::create(angle_value->as_calculated().resolve_angle(CalculationResolutionContext::from_computation_context(computation_context)).value()));
-
-    VERIFY_NOT_REACHED();
+    return absolutized_value;
```
AtkinsSJ: `absolutized_value` is only used if `specified_value->is_keyword()` is false, so these could be in a different order and save some redundant work.

Libraries/LibWeb/CSS/Parser/ValueParsing.cpp
```diff
@@ -4403,6 +4419,25 @@ RefPtr<CalculationNode const> Parser::parse_a_calculation(Vector<ComponentValue>
     return simplify_a_calculation_tree(*calculation_tree, context, CalculationResolutionContext {});
 }
 
+// https://drafts.csswg.org/css-values-5/#tree-counting
+RefPtr<TreeCountingFunctionStyleValue const> Parser::parse_tree_counting_function(ComponentValue const& token, TreeCountingFunctionStyleValue::ComputedType computed_type)
```
AtkinsSJ: I'd rather not introduce new parsing functions that take a ComponentValue instead of a TokenStream. Can you switch that?

Ah, I see later that it's also called in `convert_to_calculation_node()`... You could use `TokenStream::of_single_token()` for that, but it's a little awkward.

Libraries/LibWeb/CSS/StyleValues/TreeCountingFunctionStyleValue.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+namespace Web::CSS {
+
+String TreeCountingFunctionStyleValue::to_string(SerializationMode) const
+{
+    switch (m_function) {
+    case TreeCountingFunction::SiblingCount:
+        return MUST(String::from_utf8("sibling-count()"sv));
+    case TreeCountingFunction::SiblingIndex:
+        return MUST(String::from_utf8("sibling-index()"sv));
```
AtkinsSJ: These can just be literals: `"sibling-count()"_string` and `"sibling-index()"_string`.

Libraries/LibWeb/CSS/StyleValues/TreeCountingFunctionStyleValue.h
```diff
@@ -0,0 +1,50 @@
...
+    };
+
+    static ValueComparingNonnullRefPtr<TreeCountingFunctionStyleValue const> create(TreeCountingFunction function, ComputedType computed_type)
+    {
+        return adopt_ref(*new (nothrow) TreeCountingFunctionStyleValue(function, computed_type));
+    }
+    virtual ~TreeCountingFunctionStyleValue() override = default;
+
+    virtual String to_string(SerializationMode) const override;
+    // FIXME: Do we need to reify this?
```
AtkinsSJ: Currently no, the spec doesn't define a CSSStyleValue equivalent for it (that I'm aware of). So it'll just fall back to the StyleValue -> CSSStyleValue reification, like with any other types that aren't defined.

Libraries/LibWeb/CSS/StyleValues/CalculatedStyleValue.cpp
```diff
@@ -400,10 +400,12 @@ static String serialize_a_calculation_tree(CalculationNode const& root, Calculat
     // NOTE: Already the case.
 
     // 2. If root is a numeric value, or a non-math function, serialize root per the normal rules for it and return the result.
-    // FIXME: Support non-math functions in calculation trees.
     if (root.type() == CalculationNode::Type::Numeric)
         return static_cast<NumericCalculationNode const&>(root).value_to_string();
 
+    if (root.type() == CalculationNode::Type::NonMathFunction)
+        return static_cast<NonMathFunctionCalculationNode const&>(root).function().to_string();
```
AtkinsSJ: I know this is just imitating the line above, but we try and use `as<Foo>(bar)` now instead of `static_cast`.

Libraries/LibWeb/CSS/StyleValues/CalculatedStyleValue.cpp
```diff
@@ -2952,7 +2952,17 @@ NonnullRefPtr<CalculationNode const> simplify_a_calculation_tree(CalculationNode
...
-    // FIXME: We don't yet allow any of these inside a calculation tree. Revisit once we do.
+    if (root->type() == CalculationNode::Type::NonMathFunction) {
+        auto const& non_math_function = as<NonMathFunctionCalculationNode>(*root);
+
+        if (resolution_context.tree_counting_function_resolution_context.has_value()) {
+            if (non_math_function.function().name == "sibling-count"sv)
+                return NumericCalculationNode::create(Number { Number::Type::Integer, static_cast<double>(resolution_context.tree_counting_function_resolution_context->sibling_count) }, context);
+
+            if (non_math_function.function().name == "sibling-index"sv)
+                return NumericCalculationNode::create(Number { Number::Type::Integer, static_cast<double>(resolution_context.tree_counting_function_resolution_context->sibling_index) }, context);
```
AtkinsSJ: So currently, this would:
- Parse a StyleValue
- Discard it
- Store the Function itself
- Manually evaluate the function here

I'm wondering if `NonMathFunctionCalculationNode` could hold a `NonnullRefPtr<StyleValue>` instead. Maybe even define a base class for this that `TreeCountingFunctionStyleValue` would extend, so they all implement a `resolve_to_numeric_calculation_node()` method or similar. Then we wouldn't need to duplicate the logic for resolving these to a numeric value, and it seems like new functions could more easily be added.

There are probably aspects to this I'm missing, or I'm over-engineering this. Just wanted to raise the idea and see what you think.

Libraries/LibWeb/CSS/Parser/ValueParsing.cpp
Calme1709: Done.

Libraries/LibWeb/CSS/StyleValues/TreeCountingFunctionStyleValue.cpp
Calme1709: Done.

Libraries/LibWeb/CSS/StyleValues/TreeCountingFunctionStyleValue.h
Calme1709: Comment removed.

Libraries/LibWeb/CSS/StyleValues/CalculatedStyleValue.cpp
Calme1709: Done.

Libraries/LibWeb/CSS/StyleComputer.cpp
Calme1709: Done.

Libraries/LibWeb/CSS/StyleValues/CalculatedStyleValue.cpp
Calme1709: That seems reasonable. Perhaps slightly over-engineered for this case but it is nicer to have all the resolution code in one place and I can definitely see it coming in handy when we implement more non-math functions.

I have updated `NonMathFunctionCalculationNode` to hold a `NonnullRefPtr<AbstractNonMathCalcFunctionStyleValue>` which has a `resolve_to_calculation_node()` method.

We gain a couple of WPT tests with this as we now correctly serialize `sibling-{count,index}( )` as `sibling-{count,index}()` omitting the whitespace argument

Calme1709:
Rebased to resolve conflicts

AtkinsSJ:
...Oh, GitHub isn't letting me merge because of "conflicts", but doesn't say what they are.

Calme1709:
Rebased to resolve conflicts after #6459 was merged - hopefully that fixed the original issue as well. Also added a FIXME around clamping computed values for tree counting functions (same as we do for calculated values).

PR #4574, opened 2025-05-03
Add missing Ubuntu dependency

efeme4:
I have recently tried to build Ladybird and I noticed it fails to build without the qt6-multimedia-dev library

PR #6428, opened 2025-10-09
Meta: Update BuildInstructionsLadybird.md

mo271:
Add `qt5-multimedia-dev` to the apt command in the build instructions.

### DIFF
--- a/Documentation/BuildInstructionsLadybird.md
+++ b/Documentation/BuildInstructionsLadybird.md
@@ -18,7 +18,7 @@ CMake 3.25 or newer must be available in $PATH.
 
 <!-- Note: If you change something here, please also change it in the `devcontainer/devcontainer.json` file. -->
 ```bash
-sudo apt install autoconf autoconf-archive automake build-essential ccache cmake curl fonts-liberation2 git libdrm-dev libgl1-mesa-dev libtool nasm ninja-build pkg-config python3-venv qt6-base-dev qt6-tools-dev-tools qt6-wayland tar unzip zip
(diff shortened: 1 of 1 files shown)


Documentation/BuildInstructionsLadybird.md
tcl3, line 79: `qt6-multimedia-dev` is already mentioned here. My understanding is that it is only required if you are not using pulseaudio, so it isn't included in the list above. Are you able to build without `qt6-multimedia-dev` if you install `libpulse-dev` instead?

mo271: Ah, that explains it! I didnd't have interest in "Audio support", hence I didn't even read this part, since I only wanted to test something with images. 

My expectation was that running the `apt` command at the top and then `./Meta/ladybird.py run` would work. It didn't and told me it was missing qt-multimedia, hence I installed that and then it worked, which is great!

I tested it now and it also builds when installing `libpulse-dev` instead - nice. 

Perhaps one could add `libpulse-dev` to the packages in the top command? Or explain somewhere on the top that the list of packages provided in the first `apt` command is not sufficient to compile with the default command, one also has to chose something for Audio support. 

Searching more on this in the repo, I don't seem to be the only one who has not read the build instructions thoroughly enough: https://github.com/LadybirdBrowser/ladybird/pull/4574

If you think the instructions are good enough as they are, feel free to close this pr: no hard feelings...

Zaggy1024: If you happen to be able to remove libpulse-dev without too much trouble, could I interest you in checking the build instructions on https://github.com/LadybirdBrowser/ladybird/pull/6410? I'm removing all usage of qt-multimedia from the codebase, but I'm not easily able to test the build without the pulseaudio headers installed.

mo271: I can build it at df909738 without `ibpulse-dev` installed. 

But from this I see that this pr will be made obsolete by #6410 anyhow, so closing it now..

PR #6430, opened 2025-10-09
Add pinch-to-zoom support

kalenikaliaksandr:
For now pinch event handling is implemented only for AppKit UI.

https://github.com/user-attachments/assets/b6bd7dd0-c380-4534-9e27-16bf056d220e

### DIFF
--- a/Libraries/LibWeb/Internals/Internals.cpp
+++ b/Libraries/LibWeb/Internals/Internals.cpp
@@ -224,6 +224,13 @@ void Internals::wheel(double x, double y, double delta_x, double delta_y)
     page.handle_mousewheel(position, position, 0, 0, 0, delta_x, delta_y);
 }
 
+void Internals::pinch(double x, double y, double scale_delta)
+{
+    auto& page = this->page();
+    auto position = page.css_to_device_point({ x, y });
+    page.handle_pinch_event(position, scale_delta);
+}
+
 WebIDL::ExceptionOr<bool> Internals::dispatch_user_activated_event(DOM::EventTarget& target, DOM::Event& event)
 {
     event.set_is_trusted(true);
--- a/Libraries/LibWeb/Internals/Internals.h
+++ b/Libraries/LibWeb/Internals/Internals.h
@@ -41,6 +41,7 @@ public:
     void mouse_down(double x, double y);
     void move_pointer_to(double x, double y);
     void wheel(double x, double y, double delta_x, double delta_y);
+    void pinch(double x, double y, double scale_delta);
 
     WebIDL::ExceptionOr<bool> dispatch_user_activated_event(DOM::EventTarget&, DOM::Event& event);
 
--- a/Libraries/LibWeb/Internals/Internals.idl
+++ b/Libraries/LibWeb/Internals/Internals.idl
@@ -33,6 +33,7 @@ interface Internals {
     undefined mouseDown(double x, double y);
     undefined movePointerTo(double x, double y);
     undefined wheel(double x, double y, double deltaX, double deltaY);
+    undefined pinch(double x, double y, double scaleDelta);
 
     boolean dispatchUserActivatedEvent(EventTarget target, Event event);
     undefined spoofCurrentURL(USVString url);
--- a/Tests/LibWeb/Text/expected/vv-resize-on-pinch-to-zoom.txt
+++ b/Tests/LibWeb/Text/expected/vv-resize-on-pinch-to-zoom.txt
@@ -1,0 +1,1 @@
+vv resize: width=533.3333333333334, height=400, scale=1.5, offsetLeft=33.328125, offsetTop=33.328125
--- a/Tests/LibWeb/Text/input/vv-resize-on-pinch-to-zoom.html
+++ b/Tests/LibWeb/Text/input/vv-resize-on-pinch-to-zoom.html
@@ -1,0 +1,19 @@
+<!DOCTYPE html>
+<html lang="en">
+    <script src="include.js"></script>
+    <body></body>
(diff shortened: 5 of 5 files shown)


Libraries/LibWeb/CSS/VisualViewport.cpp
```diff
@@ -73,7 +70,7 @@ double VisualViewport::page_left() const
 
     // FIXME: 2. Otherwise, return the offset of the left edge of the visual viewport
```
konradekk: is FIXME still relevant? 🤔

Libraries/LibWeb/CSS/VisualViewport.cpp
```diff
@@ -151,4 +145,48 @@ WebIDL::CallbackType* VisualViewport::onscrollend()
...
+Gfx::AffineTransform VisualViewport::transform() const
+{
+    Gfx::AffineTransform transform;
+    auto offset = m_offset.to_type<double>() * m_scale;
+    transform.translate(-offset.x(), -offset.y());
+    transform.scale({ m_scale, m_scale });
+    return transform;
+}
+
+void VisualViewport::handle_pinch_event(CSSPixelPoint focal_point, double scale_delta)
```
konradekk: `handle_zoom_event` sounds slightly better to me.

in fact, that should not be an event handler but a comand (you already handled the event by marking it as handled; this probably should be just `zoom` or `apply_zoom` (with `focal`?).

in theory `focal_point` is just `focus`… but i think you're well aware of that! you could also use a mathematical naming here, `fixed_point`, if that's the case.

Tests/LibWeb/Text/expected/css/text-underline-position.txt
```diff
@@ -1,8 +1,11 @@
-SaveLayer
-  DrawGlyphRun rect=[35,8 8x18] translation=[35.15625,21.796875] color=rgb(0, 0, 0) scale=1
-  DrawGlyphRun rect=[8,8 28x18] translation=[8,21.796875] color=rgb(0, 0, 0) scale=1
-  DrawLine from=[8,24] to=[35,24] color=rgb(0, 0, 0) thickness=2
-  DrawGlyphRun rect=[43,8 28x18] translation=[43.15625,21.796875] color=rgb(0, 0, 0) scale=1
-  DrawLine from=[43,27] to=[71,27] color=rgb(0, 0, 0) thickness=2
+Save
```
gmta, line 1: Seems silly to do this Save/Restore for the root display list. We don't _need_ to restore anything right? Can't we skip those?

Tests/LibWeb/Text/expected/css/text-underline-position.txt
```diff
@@ -1,8 +1,11 @@
...
+  ApplyTransform matrix=[1 0 0 1 0 0]
```
gmta, line 2: ApplyTransform for identity matrices should not be necessary?

Tests/LibWeb/Text/expected/css/text-underline-position.txt
```diff
@@ -1,8 +1,11 @@
-SaveLayer
```
gmta, line 1: You've updated these tests in the wrong commit - should go in the previous one.

gmta:
I noticed that if you pinch and then dump the display list, the `ApplyTransform` still shows up with an identity matrix. Would be useful if we could see the effective values there!

gmta:
Another finding (for a future PR possibly): an `` with an SVG source does not update its rendering for the effective scale. It calculates the device pixel rect, which does not take the new scale into account. Would this be a third concept? (CSS rect, device pixel rect, visual rect?)

Tests/LibWeb/Text/expected/css/text-underline-position.txt
kalenikaliaksandr: we reuse DisplayListPlayerSkia across repaints, so if we skip those transform would leak to subsequent repaint.

Tests/LibWeb/Text/expected/css/text-underline-position.txt
kalenikaliaksandr: yes, ApplyTransform for identity matrix is not necessary, but we can't drop this one, because it's reserved to be patched up with visual viewport transform. yes, it's 3 additional items in every display list, but it's more efficient than resizing display list backing vector, in case we need to prepend a transform.

Tests/LibWeb/Text/expected/css/text-underline-position.txt
kalenikaliaksandr: ooops, moved test updated to the right commit.

kalenikaliaksandr:
> I noticed that if you pinch and then dump the display list, the `ApplyTransform` still shows up with an identity matrix. Would be useful if we could see the effective values there!

sure, I moved patching of VV transform into `Document::record_display_list()` and now it's observable in dump.

kalenikaliaksandr:
> Another finding (for a future PR possibly): an `` with an SVG source does not update its rendering for the effective scale. It calculates the device pixel rect, which does not take the new scale into account. Would this be a third concept? (CSS rect, device pixel rect, visual rect?)

yeah, we would have to take in account VV transform and rerasterize SVG images. and we likely want to figure how to throttle this process to avoid rerasterization of all SVG images after each repaint during zoom in/out. as you may've guessed that's not something I plan to solve within this PR 😅

Tests/LibWeb/Text/expected/css/text-underline-position.txt
gmta: Ah makes sense then.

PR #6432, opened 2025-10-09
NotificationAPI::Notification constructor and getter methods + some fixes needed for Notification in IDLGenerators.

NiccoloAntonelliDziri:
This PR is for adding the full constructor and main getter methods to the Notification class.

There are two commits. One fixes `boolean? silent = null` in the IDL thanks to Zaggy1024.

The other one is for Notification. The modification of the IDL for required members in dictionary is in the same commit in order to correctly build ladybird.

This passes quite a few more WPT tests essentially because (almost) all of the attributes of Notification are created now.

### DIFF
--- a/Libraries/LibWeb/DOM/Document.cpp
+++ b/Libraries/LibWeb/DOM/Document.cpp
@@ -360,7 +360,7 @@ WebIDL::ExceptionOr<GC::Ref<Document>> Document::create_and_initialize(Type type
     // AD-HOC: The response object no longer has an associated timing info object. For now, we use response's non-standard response time property,
     //         which represents the time that the time that the response object was created.
     auto response_creation_time = navigation_params.response->monotonic_response_time().nanoseconds() / 1e6;
-    load_timing_info.navigation_start_time = HighResolutionTime::coarsen_time(response_creation_time, HTML::relevant_settings_object(*window).cross_origin_isolated_capability() == HTML::CanUseCrossOriginIsolatedAPIs::Yes);
+    load_timing_info.navigation_start_time = HighResolutionTime::coarsen_time(response_creation_time, HTML::relevant_settings_object(*window).cross_origin_isolated_capability());
 
     // 9. Let document be a new Document, with
     //    type: type
--- a/Libraries/LibWeb/Fetch/Infrastructure/FetchTimingInfo.cpp
+++ b/Libraries/LibWeb/Fetch/Infrastructure/FetchTimingInfo.cpp
@@ -39,22 +39,20 @@ GC::Ref<FetchTimingInfo> create_opaque_timing_info(JS::VM& vm, FetchTimingInfo c
 
 void FetchTimingInfo::update_final_timings(Requests::RequestTimingInfo const& final_timings, HTML::CanUseCrossOriginIsolatedAPIs cross_origin_isolated_capability)
 {
-    bool has_cross_origin_isolated_capability = cross_origin_isolated_capability == HTML::CanUseCrossOriginIsolatedAPIs::Yes;
-
     auto domain_lookup_start_time_milliseconds = m_start_time + (static_cast<HighResolutionTime::DOMHighResTimeStamp>(final_timings.domain_lookup_start_microseconds) / 1000.0);
-    auto coarsened_domain_lookup_start_time = HighResolutionTime::coarsen_time(domain_lookup_start_time_milliseconds, has_cross_origin_isolated_capability);
+    auto coarsened_domain_lookup_start_time = HighResolutionTime::coarsen_time(domain_lookup_start_time_milliseconds, cross_origin_isolated_capability);
 
     auto domain_lookup_end_time_milliseconds = m_start_time + (static_cast<HighResolutionTime::DOMHighResTimeStamp>(final_timings.domain_lookup_end_microseconds) / 1000.0);
-    auto coarsened_domain_lookup_end_time = HighResolutionTime::coarsen_time(domain_lookup_end_time_milliseconds, has_cross_origin_isolated_capability);
+    auto coarsened_domain_lookup_end_time = HighResolutionTime::coarsen_time(domain_lookup_end_time_milliseconds, cross_origin_isolated_capability);
 
     auto connect_start_time_milliseconds = m_start_time + (static_cast<HighResolutionTime::DOMHighResTimeStamp>(final_timings.connect_start_microseconds) / 1000.0);
-    auto coarsened_connection_start_time = HighResolutionTime::coarsen_time(connect_start_time_milliseconds, has_cross_origin_isolated_capability);
+    auto coarsened_connection_start_time = HighResolutionTime::coarsen_time(connect_start_time_milliseconds, cross_origin_isolated_capability);
 
     auto connect_end_time_milliseconds = m_start_time + (static_cast<HighResolutionTime::DOMHighResTimeStamp>(final_timings.connect_end_microseconds) / 1000.0);
-    auto coarsened_connection_end_time = HighResolutionTime::coarsen_time(connect_end_time_milliseconds, has_cross_origin_isolated_capability);
+    auto coarsened_connection_end_time = HighResolutionTime::coarsen_time(connect_end_time_milliseconds, cross_origin_isolated_capability);
 
     auto secure_connect_start_time_milliseconds = m_start_time + (static_cast<HighResolutionTime::DOMHighResTimeStamp>(final_timings.secure_connect_start_microseconds) / 1000.0);
-    auto coarsened_secure_connection_start_time = HighResolutionTime::coarsen_time(secure_connect_start_time_milliseconds, has_cross_origin_isolated_capability);
+    auto coarsened_secure_connection_start_time = HighResolutionTime::coarsen_time(secure_connect_start_time_milliseconds, cross_origin_isolated_capability);
 
     m_final_connection_timing_info = ConnectionTimingInfo {
         .domain_lookup_start_time = coarsened_domain_lookup_start_time,
@@ -66,11 +64,11 @@ void FetchTimingInfo::update_final_timings(Requests::RequestTimingInfo const& fi
     };
 
     auto request_start_time_milliseconds = m_start_time + (static_cast<HighResolutionTime::DOMHighResTimeStamp>(final_timings.request_start_microseconds) / 1000.0);
-    auto coarsened_request_start_time = HighResolutionTime::coarsen_time(request_start_time_milliseconds, has_cross_origin_isolated_capability);
+    auto coarsened_request_start_time = HighResolutionTime::coarsen_time(request_start_time_milliseconds, cross_origin_isolated_capability);
     m_final_network_request_start_time = coarsened_request_start_time;
 
     auto response_start_time_milliseconds = m_start_time + (static_cast<HighResolutionTime::DOMHighResTimeStamp>(final_timings.response_start_microseconds) / 1000.0);
-    auto coarsened_response_start_time = HighResolutionTime::coarsen_time(response_start_time_milliseconds, has_cross_origin_isolated_capability);
+    auto coarsened_response_start_time = HighResolutionTime::coarsen_time(response_start_time_milliseconds, cross_origin_isolated_capability);
     m_final_network_response_start_time = coarsened_response_start_time;
 }
 
--- a/Libraries/LibWeb/HTML/Scripting/Environments.h
+++ b/Libraries/LibWeb/HTML/Scripting/Environments.h
@@ -154,7 +154,7 @@ public:
         auto unsafe_walltime = HighResolutionTime::wall_clock_unsafe_current_time();
 
         // 2. Return the result of calling coarsen time with unsafeWallTime and settingsObject's cross-origin isolated capability.
-        return HighResolutionTime::coarsen_time(unsafe_walltime, cross_origin_isolated_capability() == HTML::CanUseCrossOriginIsolatedAPIs::Yes);
+        return HighResolutionTime::coarsen_time(unsafe_walltime, cross_origin_isolated_capability());
     }
 
 protected:
--- a/Libraries/LibWeb/HighResolutionTime/TimeOrigin.h
+++ b/Libraries/LibWeb/HighResolutionTime/TimeOrigin.h
@@ -9,17 +9,18 @@
 
 #include <LibJS/Forward.h>
 #include <LibWeb/Export.h>
+#include <LibWeb/HTML/Scripting/SerializedEnvironmentSettingsObject.h>
 #include <LibWeb/HighResolutionTime/DOMHighResTimeStamp.h>
 
 namespace Web::HighResolutionTime {
 
 DOMHighResTimeStamp estimated_monotonic_time_of_the_unix_epoch();
 DOMHighResTimeStamp get_time_origin_timestamp(JS::Object const&);
-DOMHighResTimeStamp coarsen_time(DOMHighResTimeStamp timestamp, bool cross_origin_isolated_capability = false);
+DOMHighResTimeStamp coarsen_time(DOMHighResTimeStamp timestamp, HTML::CanUseCrossOriginIsolatedAPIs cross_origin_isolated_capability = HTML::CanUseCrossOriginIsolatedAPIs::No);
 DOMHighResTimeStamp current_high_resolution_time(JS::Object const&);
 DOMHighResTimeStamp relative_high_resolution_time(DOMHighResTimeStamp, JS::Object const&);
 DOMHighResTimeStamp relative_high_resolution_coarsen_time(DOMHighResTimeStamp, JS::Object const&);
-WEB_API DOMHighResTimeStamp coarsened_shared_current_time(bool cross_origin_isolated_capability = false);
+WEB_API DOMHighResTimeStamp coarsened_shared_current_time(HTML::CanUseCrossOriginIsolatedAPIs cross_origin_isolated_capability = HTML::CanUseCrossOriginIsolatedAPIs::No);
 DOMHighResTimeStamp wall_clock_unsafe_current_time();
 WEB_API DOMHighResTimeStamp unsafe_shared_current_time();
 
--- a/Libraries/LibWeb/Fetch/Fetching/Fetching.cpp
+++ b/Libraries/LibWeb/Fetch/Fetching/Fetching.cpp
@@ -191,7 +191,7 @@ GC::Ref<Infrastructure::FetchController> fetch(JS::Realm& realm, Infrastructure:
     //    shared current time given crossOriginIsolatedCapability, and render-blocking is set to request’s
     //    render-blocking.
     auto timing_info = Infrastructure::FetchTimingInfo::create(vm);
-    auto now = HighResolutionTime::coarsened_shared_current_time(cross_origin_isolated_capability == HTML::CanUseCrossOriginIsolatedAPIs::Yes);
+    auto now = HighResolutionTime::coarsened_shared_current_time(cross_origin_isolated_capability);
     timing_info->set_start_time(now);
     timing_info->set_post_redirect_start_time(now);
     timing_info->set_render_blocking(request.render_blocking());
@@ -1167,7 +1167,7 @@ GC::Ref<PendingResponse> http_fetch(JS::Realm& realm, Infrastructure::FetchParam
 
         // 3. Let serviceWorkerStartTime be the coarsened shared current time given fetchParams’s cross-origin isolated
         //    capability.
-        auto service_worker_start_time = HighResolutionTime::coarsened_shared_current_time(fetch_params.cross_origin_isolated_capability() == HTML::CanUseCrossOriginIsolatedAPIs::Yes);
+        auto service_worker_start_time = HighResolutionTime::coarsened_shared_current_time(fetch_params.cross_origin_isolated_capability());
 
         // FIXME: 4. Set response to the result of invoking handle fetch for requestForServiceWorker, with fetchParams’s
         //           controller and fetchParams’s cross-origin isolated capability.
@@ -1469,7 +1469,7 @@ GC::Ptr<PendingResponse> http_redirect_fetch(JS::Realm& realm, Infrastructure::F
 
     // 16. Set timingInfo’s redirect end time and post-redirect start time to the coarsened shared current time given
     //     fetchParams’s cross-origin isolated capability.
-    auto now = HighResolutionTime::coarsened_shared_current_time(fetch_params.cross_origin_isolated_capability() == HTML::CanUseCrossOriginIsolatedAPIs::Yes);
+    auto now = HighResolutionTime::coarsened_shared_current_time(fetch_params.cross_origin_isolated_capability());
     timing_info->set_redirect_end_time(now);
     timing_info->set_post_redirect_start_time(now);
 
--- a/Libraries/LibWeb/HTML/BrowsingContext.cpp
+++ b/Libraries/LibWeb/HTML/BrowsingContext.cpp
@@ -204,7 +204,7 @@ WebIDL::ExceptionOr<BrowsingContext::BrowsingContextAndDocument> BrowsingContext
     auto load_timing_info = DOM::DocumentLoadTimingInfo();
     load_timing_info.navigation_start_time = HighResolutionTime::coarsen_time(
         unsafe_context_creation_time,
-        as<WindowEnvironmentSettingsObject>(Bindings::principal_host_defined_environment_settings_object(window->realm())).cross_origin_isolated_capability() == CanUseCrossOriginIsolatedAPIs::Yes);
+        as<WindowEnvironmentSettingsObject>(Bindings::principal_host_defined_environment_settings_object(window->realm())).cross_origin_isolated_capability());
 
     // 15. Let document be a new Document, with:
     auto document = HTML::HTMLDocument::create(window->realm());
--- a/Libraries/LibWeb/HTML/Scripting/WorkerEnvironmentSettingsObject.cpp
+++ b/Libraries/LibWeb/HTML/Scripting/WorkerEnvironmentSettingsObject.cpp
@@ -114,7 +114,7 @@ CanUseCrossOriginIsolatedAPIs WorkerEnvironmentSettingsObject::cross_origin_isol
 double WorkerEnvironmentSettingsObject::time_origin() const
 {
     // Return the result of coarsening unsafeWorkerCreationTime with worker global scope's cross-origin isolated capability.
-    return HighResolutionTime::coarsen_time(m_unsafe_worker_creation_time, cross_origin_isolated_capability() == CanUseCrossOriginIsolatedAPIs::Yes);
+    return HighResolutionTime::coarsen_time(m_unsafe_worker_creation_time, cross_origin_isolated_capability());
 }
 
 void WorkerEnvironmentSettingsObject::visit_edges(JS::Cell::Visitor& visitor)
--- a/Libraries/LibWeb/HighResolutionTime/TimeOrigin.cpp
+++ b/Libraries/LibWeb/HighResolutionTime/TimeOrigin.cpp
@@ -45,13 +45,13 @@ DOMHighResTimeStamp get_time_origin_timestamp(JS::Object const& global)
 }
 
 // https://w3c.github.io/hr-time/#dfn-coarsen-time
-DOMHighResTimeStamp coarsen_time(DOMHighResTimeStamp timestamp, bool cross_origin_isolated_capability)
+DOMHighResTimeStamp coarsen_time(DOMHighResTimeStamp timestamp, HTML::CanUseCrossOriginIsolatedAPIs cross_origin_isolated_capability)
 {
     // 1. Let time resolution be 100 microseconds, or a higher implementation-defined value.
     auto time_resolution_milliseconds = 0.1;
 
     // 2. If crossOriginIsolatedCapability is true, set time resolution to be 5 microseconds, or a higher implementation-defined value.
-    if (cross_origin_isolated_capability)
+    if (cross_origin_isolated_capability == HTML::CanUseCrossOriginIsolatedAPIs::Yes)
         time_resolution_milliseconds = 0.005;
 
     // 3. In an implementation-defined manner, coarsen and potentially jitter timestamp such that its resolution will not exceed time resolution
@@ -79,7 +79,7 @@ DOMHighResTimeStamp relative_high_resolution_time(DOMHighResTimeStamp time, JS::
     auto& settings = HTML::relevant_principal_settings_object(global);
 
     // 2. Let coarse time be the result of calling coarsen time with time and settings's cross-origin isolated capability.
-    auto coarse_time = coarsen_time(time, settings.cross_origin_isolated_capability() == HTML::CanUseCrossOriginIsolatedAPIs::Yes);
+    auto coarse_time = coarsen_time(time, settings.cross_origin_isolated_capability());
 
     // 2. Return the relative high resolution coarse time for coarse time and global.
     return relative_high_resolution_coarsen_time(coarse_time, global);
@@ -94,7 +94,7 @@ DOMHighResTimeStamp relative_high_resolution_coarsen_time(DOMHighResTimeStamp co
 }
 
 // https://w3c.github.io/hr-time/#dfn-coarsened-shared-current-time
-DOMHighResTimeStamp coarsened_shared_current_time(bool cross_origin_isolated_capability)
+DOMHighResTimeStamp coarsened_shared_current_time(HTML::CanUseCrossOriginIsolatedAPIs cross_origin_isolated_capability)
 {
     // The coarsened shared current time given an optional boolean crossOriginIsolatedCapability (default false), must return the result of calling coarsen time with the unsafe shared current time and crossOriginIsolatedCapability.
     return coarsen_time(unsafe_shared_current_time(), cross_origin_isolated_capability);


Meta/Lagom/Tools/CodeGenerators/LibWeb/BindingsGenerator/IDLGenerators.cpp
```diff
@@ -621,7 +624,7 @@ static void generate_to_integral(SourceGenerator& scoped_generator, ParameterTyp
 
     // FIXME: The Optional<foo> defaults to empty, and can't be explicitly set to `null`.
     //        So, just skip the assignment.
```
gmta: This FIXME now no longer relates to the code underneath it; maybe remove it? I don't think it's relevant anymore after this change.

Libraries/LibWeb/NotificationsAPI/Notification.cpp
gmta, line 1: Commit message typos: "ladybird" -> "Ladybird", "Notificaion.idl" -> "Notification.idl".

Also this sentence suddenly ends:
"This passes a least a few more tests because"

Libraries/LibWeb/NotificationsAPI/Notification.cpp
```diff
@@ -16,18 +19,179 @@ GC_DEFINE_ALLOCATOR(Notification);
...
+    Optional<NotificationOptions> options,
+    URL::Origin origin,
+    URL::URL base_URL,
+    HighResolutionTime::EpochTimeStamp fallbackTimestamp)
+{
+    // 1. Let notification be a new notification.
+    ConceptNotification notification;
+
+    // 2. If options["silent"] is true and options["vibrate"] exists, then throw a TypeError.
+    // FIXME: add the condition when VibratePattern is implemented
```
gmta: Please mark step 2 as FIXME as well, i.e.

```cpp
// FIXME: 2. ...
```

Do this for all spec steps that have a FIXME in some form.

Libraries/LibWeb/NotificationsAPI/Notification.cpp
```diff
@@ -16,18 +19,179 @@ GC_DEFINE_ALLOCATOR(Notification);
 Notification::Notification(JS::Realm& realm)
     : DOM::EventTarget(realm)
 {
+    this->body();
+}
+
+WebIDL::ExceptionOr<ConceptNotification> Notification::create_a_notification(
```
gmta, line 26: Please add a spec link here

Libraries/LibWeb/NotificationsAPI/Notification.cpp
```diff
@@ -16,18 +19,179 @@ GC_DEFINE_ALLOCATOR(Notification);
...
+
+        // 6. Append action to notification’s actions.
+        notification.actions.append(action);
+    }
+
+    // 22. Return notification.
+    return notification;
+}
+
+WebIDL::ExceptionOr<ConceptNotification> Notification::create_a_notification_with_a_settings_object(
```
gmta, line 142: Please add a spec link here as well

Libraries/LibWeb/NotificationsAPI/Notification.cpp
```diff
@@ -16,18 +19,179 @@ GC_DEFINE_ALLOCATOR(Notification);
...
+    GC::Ptr<HTML::EnvironmentSettingsObject> settings)
+{
+    // 1. Let origin be settings’s origin.
+    URL::Origin origin = settings->origin();
+
+    // 2. Let baseURL be settings’s API base URL.
+    URL::URL base_URL = settings->api_base_url();
+
+    // 3. Let fallbackTimestamp be the number of milliseconds from the Unix epoch to settings’s current wall time, rounded to the nearest integer.
+    HighResolutionTime::EpochTimeStamp fallbacktimestamp = static_cast<HighResolutionTime::EpochTimeStamp>(UnixDateTime::now().milliseconds_since_epoch());
```
gmta: Seems like we're ignoring the "settings’s current wall time" part, is that supposed to happen?

Libraries/LibWeb/NotificationsAPI/Notification.cpp
```diff
@@ -16,18 +19,179 @@ GC_DEFINE_ALLOCATOR(Notification);
...
+        notification.icon_URL = try_set_URL(options->icon.value());
+    }
+
+    // 14. If options["badge"] exists, then parse it using baseURL, and if that does not return failure, set notification’s badge URL to the return value. (Otherwise notification’s badge URL is not set.)
+    if (options->badge.has_value()) {
+        notification.badge_URL = try_set_URL(options->badge.value());
+    }
+
+    // 15. If options["vibrate"] exists, then validate and normalize it and set notification’s vibration pattern to the return value.
+    // FIXME
```
Psychpsyo: `// FIXME: Implement this`. As it stands right now, it looks kinda unfinished.
(though I'm also not a maintainer. I just tend to search the code for "FIXME:" with the colon)

Libraries/LibWeb/NotificationsAPI/Notification.cpp
```diff
@@ -16,18 +19,179 @@ GC_DEFINE_ALLOCATOR(Notification);
...
+
+    if (exception_or_notification.is_exception())
+        return exception_or_notification.release_error();
+    ConceptNotification notification = exception_or_notification.release_value();
+
+    // 4. Associate this with notification.
+    this_notification->m_notification = notification;
+
+    // 5. Run these steps in parallel:
+
```
Psychpsyo, line 187: There is some FIXMEs missing here.

Libraries/LibWeb/NotificationsAPI/Notification.cpp
NiccoloAntonelliDziri: At first I didn't find anything apart from  `time_origin()` virtual method defined here: 

https://html.spec.whatwg.org/multipage/webappapis.html#concept-settings-object-time-origin

It should return a number used as the baseline for performance-related timestamps, which does not seem useful here.

But by taking a closer look, the settings current wall time is defined like this in https://w3c.github.io/hr-time/#dfn-eso-current-wall-time

```
An [environment settings object](https://html.spec.whatwg.org/multipage/webappapis.html#environment-settings-object) settingsObject's current wall time is the result of the following steps:

    1. Let unsafeWallTime be the [wall clock](https://w3c.github.io/hr-time/#dfn-wall-clock)'s [unsafe current time](https://w3c.github.io/hr-time/#wall-clock-unsafe-current-time).
    2. Return the result of calling [coarsen time](https://w3c.github.io/hr-time/#dfn-coarsen-time) with unsafeWallTime and settingsObject's [cross-origin isolated capability](https://html.spec.whatwg.org/multipage/webappapis.html#concept-settings-object-cross-origin-isolated-capability).
```
But it does not appear to be exist yet in the codebase.

I can either add the method in `EnvironmentSettingsObject` or directly implement those two steps in Notification in order to just get the setting's current wall time for this specific use case.

gmta: Either seems fine, as long as you leave a `// FIXME:` somewhere to show if you don't yet follow the complete spec :)

Libraries/LibWeb/HTML/Scripting/Environments.h
```diff
@@ -138,6 +139,21 @@ struct WEB_API EnvironmentSettingsObject : public Environment {
 
     virtual void discard_environment() override;
 
+    // FIXME: This method below is from HighResolutionTime spec in section 3. Section for Specification Authors.
+    // There are other methods that are not yet implemented in ladybird.
```
konradekk: not sure how informative this is.

Meta/Lagom/Tools/CodeGenerators/LibWeb/BindingsGenerator/IDLGenerators.cpp
```diff
@@ -622,8 +622,6 @@ static void generate_to_integral(SourceGenerator& scoped_generator, ParameterTyp
 )~~~");
     }
 
-    // FIXME: The Optional<foo> defaults to empty, and can't be explicitly set to `null`.
-    //        So, just skip the assignment.
```
Zaggy1024, line 641: This removal should be in the previous commit, it seems.

Meta/Lagom/Tools/CodeGenerators/LibWeb/BindingsGenerator/IDLGenerators.cpp
```diff
@@ -2154,7 +2152,7 @@ static void generate_wrap_statement(SourceGenerator& generator, ByteString const
                 //      the generated code to do some metaprogramming to inspect the type of the member in the C++ struct to
                 //      determine whether the type is present or not (e.g through a has_value() on an Optional<T>, or a null
                 //      check on a GC::Ptr<T>). So to save some complexity in the generator, give ourselves a hint of what to do.
-                bool is_optional = !member.extended_attributes.contains("GenerateAsRequired") && !member.default_value.has_value();
+                bool is_optional = !member.required && !member.extended_attributes.contains("GenerateAsRequired") && !member.default_value.has_value();
```
Zaggy1024, line 2219: Is this the change mentioned in the second commit's message, which is required for Ladybird to build? Does the previous commit not build with this change?

If it can be part of the previous commit changing IDLGenerators, then please make sure to reword the commit not to mention it, as well.

Libraries/LibWeb/NotificationsAPI/Notification.h
```diff
@@ -34,12 +36,48 @@ struct NotificationOptions {
     // VibratePattern vibrate;  // FIXME: properly implement vibrate pattern
     Optional<HighResolutionTime::EpochTimeStamp> timestamp;
     bool renotify = false;
-    Optional<bool> silent;
+    Optional<bool> silent = OptionalNone();
```
Zaggy1024: Does this do anything? Optional should default initialize to none.

Libraries/LibWeb/HighResolutionTime/TimeOrigin.h
```diff
@@ -9,17 +9,19 @@
...
 #include <LibWeb/Export.h>
+#include <LibWeb/HTML/Scripting/SerializedEnvironmentSettingsObject.h>
 #include <LibWeb/HighResolutionTime/DOMHighResTimeStamp.h>
 
 namespace Web::HighResolutionTime {
 
 DOMHighResTimeStamp estimated_monotonic_time_of_the_unix_epoch();
 DOMHighResTimeStamp get_time_origin_timestamp(JS::Object const&);
-DOMHighResTimeStamp coarsen_time(DOMHighResTimeStamp timestamp, bool cross_origin_isolated_capability = false);
+// DOMHighResTimeStamp coarsen_time(DOMHighResTimeStamp timestamp, bool cross_origin_isolated_capability = false);
```
Zaggy1024: Please remove commented code!

Zaggy1024 (changes requested):

Meta/Lagom/Tools/CodeGenerators/LibWeb/BindingsGenerator/IDLGenerators.cpp
NiccoloAntonelliDziri: For the second commit to build, it needs this change, that's why it was initially in this commit. But you're right in the fact that it can be moved to the previous one, it will build just fine with the benefit of tidying commits up, so I moved it.

Libraries/LibWeb/HTML/Scripting/Environments.h
```diff
@@ -138,6 +139,21 @@ struct WEB_API EnvironmentSettingsObject : public Environment {
...
+
+    // https://w3c.github.io/hr-time/#dfn-eso-current-wall-time
+    HighResolutionTime::DOMHighResTimeStamp current_wall_time() const
+    {
+        // An environment settings object settingsObject's current wall time is the result of the following steps:
+
+        // 1. Let unsafeWallTime be the wall clock's unsafe current time.
+        HighResolutionTime::DOMHighResTimeStamp unsafeWallTime = HighResolutionTime::wall_clock_unsafe_current_time();
```
AtkinsSJ: Please use `snake_case` for variable names, even ones from the spec. eg, `unsafe_wall_time`.

Also nit: We tend to use `auto` for things where the type is obvious.

Libraries/LibWeb/HTML/Scripting/Environments.h
```diff
@@ -138,6 +139,21 @@ struct WEB_API EnvironmentSettingsObject : public Environment {
...
+
+        // 2. Return the result of calling coarsen time with unsafeWallTime and settingsObject's cross-origin isolated capability.
+        return HighResolutionTime::coarsen_time(unsafeWallTime, this->cross_origin_isolated_capability() == CanUseCrossOriginIsolatedAPIs::Yes);
```
AtkinsSJ: Nit: You don't need `this->` and we generally avoid it unless it's necessary.

Libraries/LibWeb/NotificationsAPI/Notification.cpp
```diff
@@ -16,18 +20,179 @@ GC_DEFINE_ALLOCATOR(Notification);
...
+}
+
+// https://notifications.spec.whatwg.org/#create-a-notification
+WebIDL::ExceptionOr<ConceptNotification> Notification::create_a_notification(
+    JS::Realm& realm,
+    String title,
+    Optional<NotificationOptions> options,
+    URL::Origin origin,
+    URL::URL base_URL,
+    HighResolutionTime::EpochTimeStamp fallbackTimestamp)
```
AtkinsSJ: `fallback_timestamp`

Libraries/LibWeb/NotificationsAPI/Notification.cpp
```diff
@@ -16,18 +20,179 @@ GC_DEFINE_ALLOCATOR(Notification);
...
+    this->body();
+}
+
+// https://notifications.spec.whatwg.org/#create-a-notification
+WebIDL::ExceptionOr<ConceptNotification> Notification::create_a_notification(
+    JS::Realm& realm,
+    String title,
+    Optional<NotificationOptions> options,
+    URL::Origin origin,
+    URL::URL base_URL,
```
AtkinsSJ: `base_url`

Libraries/LibWeb/NotificationsAPI/Notification.cpp
```diff
@@ -16,18 +20,179 @@ GC_DEFINE_ALLOCATOR(Notification);
...
+    // 7. Set notification’s language to options["lang"].
+    notification.language = options->lang;
+
+    // 8. Set notification’s origin to origin.
+    notification.origin = move(origin);
+
+    // 9. Set notification’s body to options["body"].
+    notification.body = options->body;
+
+    auto try_set_URL = [&base_URL](String const& url) -> Optional<URL::URL> {
```
AtkinsSJ: `try_set_url`

Libraries/LibWeb/NotificationsAPI/Notification.cpp
```diff
@@ -16,18 +20,179 @@ GC_DEFINE_ALLOCATOR(Notification);
...
+        auto maybe_URL = base_URL.complete_url(url);
+        if (maybe_URL.has_value())
+            return maybe_URL.release_value();
+        return {};
+    };
+
+    // 10. If options["navigate"] exists, then parse it using baseURL, and if that does not return failure, set notification’s navigation URL to the return value. (Otherwise notification’s navigation URL remains null.)
```
AtkinsSJ: Please wrap long comments at 120 columns.

Libraries/LibWeb/NotificationsAPI/Notification.cpp
```diff
@@ -16,18 +20,179 @@ GC_DEFINE_ALLOCATOR(Notification);
...
+    notification.silent_preference = options->silent;
+
+    // 19. Set notification’s require interaction preference to options["requireInteraction"].
+    notification.require_interaction_preference = options->require_interaction;
+
+    // 20. Set notification’s actions to « ».
+    notification.actions = {};
+
+    // 21. For each entry in options["actions"], up to the maximum number of actions supported (skip any excess entries):
+    for (auto entry : options->actions) {
```
AtkinsSJ: `auto const&`, to avoid making copies of the entries.

Libraries/LibWeb/NotificationsAPI/Notification.cpp
```diff
@@ -16,18 +20,179 @@ GC_DEFINE_ALLOCATOR(Notification);
...
+    JS::Realm& realm,
+    String title,
+    Optional<NotificationOptions> options,
+    GC::Ptr<HTML::EnvironmentSettingsObject> settings)
+{
+    // 1. Let origin be settings’s origin.
+    URL::Origin origin = settings->origin();
+
+    // 2. Let baseURL be settings’s API base URL.
+    URL::URL base_URL = settings->api_base_url();
```
AtkinsSJ: `base_url`

Libraries/LibWeb/NotificationsAPI/Notification.cpp
```diff
@@ -16,18 +20,179 @@ GC_DEFINE_ALLOCATOR(Notification);
...
+
+    // 3. Let fallbackTimestamp be the number of milliseconds from the Unix epoch to settings’s current wall time, rounded to the nearest integer.
+    HighResolutionTime::EpochTimeStamp fallbacktimestamp = round_to<HighResolutionTime::EpochTimeStamp>(settings->current_wall_time());
```
AtkinsSJ: `auto fallback_timestamp`

Libraries/LibWeb/NotificationsAPI/Notification.cpp
```diff
@@ -16,18 +20,179 @@ GC_DEFINE_ALLOCATOR(Notification);
...
+    // 2. If options["actions"] is not empty, then throw a TypeError.
+    if (!options->actions.is_empty())
+        return WebIDL::SimpleException { WebIDL::SimpleExceptionType::TypeError, "Options `action` is not empty"sv };
+
+    // 3. Let notification be the result of creating a notification with a settings object given title, options, and this’s relevant settings object.
+    auto exception_or_notification = create_a_notification_with_a_settings_object(realm, move(title), options, relevant_settings_object);
+
+    if (exception_or_notification.is_exception())
+        return exception_or_notification.release_error();
+    ConceptNotification notification = exception_or_notification.release_value();
```
AtkinsSJ: Use the `TRY()` macro for things like this:
```c++
auto notification = TRY(create_a_notification_with_a_settings_object(realm, move(title), options, relevant_settings_object));
```
If it's an error/exception, that's returned from the function. Otherwise `notification` is set to the value.

Libraries/LibWeb/NotificationsAPI/Notification.h
```diff
@@ -40,6 +42,42 @@ struct NotificationOptions {
...
+// https://notifications.spec.whatwg.org/#concept-notification
+// This is the notification described as "notification" in the spec. Do not confuse it with "notification" as in the IDL which is just the JS wrapper.
+// "A notification is an abstract representation of something that happened, such as the delivery of a message."
+struct ConceptNotification {
+    // FIXME: A notification has an associated service worker registration (null or a service worker registration). It is initially null.
+    String title;
+    Bindings::NotificationDirection direction;
+    String language;
+    String body;
+    Optional<URL::URL> navigation_URL;
```
AtkinsSJ: Lowercase url

Libraries/LibWeb/NotificationsAPI/Notification.h
```diff
@@ -40,6 +42,42 @@ struct NotificationOptions {
...
+    String tag;
+    HTML::SerializationRecord data;
+    HighResolutionTime::EpochTimeStamp timestamp;
+    URL::Origin origin = URL::Origin({}); // FIXME: Is this a hack ? There is no default constructor to URL::Origin and the value for `origin` is set in `create-a-notification-with-a-settings-object`
+    bool renotify_preference;
+    Optional<bool> silent_preference;
+    bool require_interaction_preference;
+    Optional<URL::URL> image_URL;
+    Optional<URL::URL> icon_URL;
+    Optional<URL::URL> badge_URL;
```
AtkinsSJ: Lowercase url

Libraries/LibWeb/NotificationsAPI/Notification.h
```diff
@@ -40,6 +42,42 @@ struct NotificationOptions {
...
+
+    // FIXME: add the ressources from m_image_URL, m_icon_URL and m_badge_URL
```
AtkinsSJ: Typo: resources

Libraries/LibWeb/NotificationsAPI/Notification.h
```diff
@@ -40,6 +42,42 @@ struct NotificationOptions {
...
+
+    // FIXME: A notification has an associated vibration pattern (a list). It is initially « ».
+
+    // https://notifications.spec.whatwg.org/#action
+    struct Action {
+        String name;
+        String title;
+        Optional<URL::URL> navigation_URL;
+        Optional<URL::URL> icon_URL;
```
AtkinsSJ: Lowercase url

Libraries/LibWeb/NotificationsAPI/Notification.h
```diff
@@ -40,6 +42,42 @@ struct NotificationOptions {
...
+        // FIXME: icon ressource
```
AtkinsSJ: Same typo. I can only apologise for the disaster that is English spelling.

Libraries/LibWeb/NotificationsAPI/Notification.h
```diff
@@ -51,10 +89,89 @@ class WEB_API Notification final : public DOM::EventTarget {
...
+        GC::Ptr<HTML::EnvironmentSettingsObject> settings);
+
+    // https://notifications.spec.whatwg.org/#create-a-notification
+    static WebIDL::ExceptionOr<ConceptNotification> create_a_notification(
+        JS::Realm& realm,
+        String title,
+        Optional<NotificationOptions> options,
+        URL::Origin origin,
+        URL::URL base_URL,
+        HighResolutionTime::EpochTimeStamp fallbackTimestamp);
```
AtkinsSJ: More variable casing

Libraries/LibWeb/NotificationsAPI/Notification.h
```diff
@@ -51,10 +89,89 @@ class WEB_API Notification final : public DOM::EventTarget {
...
+    String body() const { return this->m_notification.body; }
+    String navigate() const { return this->m_notification.navigation_URL.has_value() ? this->m_notification.navigation_URL->serialize() : ""_string; }
+    String tag() const { return this->m_notification.tag; }
+    String image() const { return this->m_notification.image_URL.has_value() ? this->m_notification.image_URL->serialize() : ""_string; }
+    String icon() const { return this->m_notification.icon_URL.has_value() ? this->m_notification.icon_URL->serialize() : ""_string; }
+    String badge() const { return this->m_notification.badge_URL.has_value() ? this->m_notification.badge_URL->serialize() : ""_string; }
+    HighResolutionTime::EpochTimeStamp timestamp() const { return this->m_notification.timestamp; }
+    bool renotify() const { return this->m_notification.renotify_preference; }
+    Optional<bool> silent() const { return this->m_notification.silent_preference; }
+    bool require_interaction() const { return this->m_notification.require_interaction_preference; }
```
AtkinsSJ: None of these need `this->`

Libraries/LibWeb/NotificationsAPI/Notification.h
```diff
@@ -51,10 +89,89 @@ class WEB_API Notification final : public DOM::EventTarget {
...
+    Vector<NotificationAction> actions() const
+    {
+        // 1. Let frozenActions be an empty list of type NotificationAction.
+        Vector<NotificationAction> frozen_actions;
+
+        // 2. For each entry of this’s notification’s actions:
+        for (auto entry : this->m_notification.actions) {
```
AtkinsSJ: `for (auto const& entry : m_notification.actions) {`

Libraries/LibWeb/NotificationsAPI/Notification.h
```diff
@@ -51,10 +89,89 @@ class WEB_API Notification final : public DOM::EventTarget {
...
+            // 7. Append action to frozenActions.
+            frozen_actions.append(action);
+        }
+
+        // FIXME: 3. Return the result of create a frozen array from frozenActions.
+        return frozen_actions;
+    }
+    JS::Value data() const
+    {
+        auto deserialized_data = HTML::structured_deserialize(this->vm(), this->m_notification.data, this->realm());
```
AtkinsSJ: Don't need `this->`

Libraries/LibWeb/Fetch/Infrastructure/FetchTimingInfo.cpp
```diff
@@ -39,7 +39,7 @@ GC::Ref<FetchTimingInfo> create_opaque_timing_info(JS::VM& vm, FetchTimingInfo c
 
 void FetchTimingInfo::update_final_timings(Requests::RequestTimingInfo const& final_timings, HTML::CanUseCrossOriginIsolatedAPIs cross_origin_isolated_capability)
 {
-    bool has_cross_origin_isolated_capability = cross_origin_isolated_capability == HTML::CanUseCrossOriginIsolatedAPIs::Yes;
+    HTML::CanUseCrossOriginIsolatedAPIs has_cross_origin_isolated_capability = cross_origin_isolated_capability;
```
AtkinsSJ: I think you can use `cross_origin_isolated_capability` directly, and don't need this variable.

AtkinsSJ (changes requested):
Mostly just code-style nits - this looks good to me in general.

On a similar note, please start your commit titles with a verb. eg, "LibWeb: Add constructor for Notification and getter methods" instead of "LibWeb: Constructor for Notification and getter methods".

Also I saw you were worried about it going stale. As long as there's activity on a PR, StaleBot will stay away. It's only if nothing happens for a month that it'll get closed.

Libraries/LibWeb/Fetch/Infrastructure/FetchTimingInfo.cpp
```diff
@@ -39,6 +39,7 @@ GC::Ref<FetchTimingInfo> create_opaque_timing_info(JS::VM& vm, FetchTimingInfo c
 
 void FetchTimingInfo::update_final_timings(Requests::RequestTimingInfo const& final_timings, HTML::CanUseCrossOriginIsolatedAPIs cross_origin_isolated_capability)
 {
+
```
AtkinsSJ: Nit: Random whitespace change.

Meta/Lagom/Tools/CodeGenerators/LibWeb/BindingsGenerator/IDLGenerators.cpp
```diff
@@ -569,7 +569,7 @@ static void generate_from_integral(SourceGenerator& scoped_generator, IDL::Type
 }
```
gmta, line 580: Commit title:

> IDLGenerators: IDL `boolean? silent = null` to Optional<bool>`

This isn't a fully formed sentence and is missing a backtick. It's also a bit superfluous to repeat "IDL" which is easily derived from "IDLGenerators" already. Maybe something like this?

> IDLGenerators: Translate `boolean? silent = null` to `Optional<bool>`

Libraries/LibWeb/Fetch/Infrastructure/FetchTimingInfo.cpp
gmta: Stray newline, please remove

Libraries/LibWeb/HTML/Scripting/Environments.h
gmta: The spec link already shows that the method is from section 3, and instead of mentioning "other methods" I'd either leave out this `FIXME` entirely since it's not immediately actionable, or mention the exact methods that still need implementing.

Libraries/LibWeb/NotificationsAPI/Notification.cpp
```diff
@@ -16,18 +20,185 @@ GC_DEFINE_ALLOCATOR(Notification);
 Notification::Notification(JS::Realm& realm)
     : DOM::EventTarget(realm)
 {
+    this->body();
```
gmta: Why is this here?

Libraries/LibWeb/NotificationsAPI/Notification.cpp
```diff
@@ -16,18 +20,185 @@ GC_DEFINE_ALLOCATOR(Notification);
...
+    URL::URL base_url,
+    HighResolutionTime::EpochTimeStamp fallback_timestamp)
+{
+    // 1. Let notification be a new notification.
+    ConceptNotification notification;
+
+    // FIXME: 2. If options["silent"] is true and options["vibrate"] exists, then throw a TypeError.
+
+    // 3. If options["renotify"] is true and options["tag"] is the empty string, then throw a TypeError.
+    if (options->renotify && options->tag.is_empty())
```
gmta: You're dereferencing `options` here, which is an `Optional`, without checking if it has a value. Either we should assert that a value is always present, or even better: don't make it an `Optional`. The spec seems to indicate that the dictionary is always there.

Libraries/LibWeb/NotificationsAPI/Notification.cpp
```diff
@@ -16,18 +20,185 @@ GC_DEFINE_ALLOCATOR(Notification);
...
+
+// https://notifications.spec.whatwg.org/#create-a-notification-with-a-settings-object
+WebIDL::ExceptionOr<ConceptNotification> Notification::create_a_notification_with_a_settings_object(
+    JS::Realm& realm,
+    String title,
+    Optional<NotificationOptions> options,
+    GC::Ptr<HTML::EnvironmentSettingsObject> settings)
+{
+    // 1. Let origin be settings’s origin.
+    URL::Origin origin = settings->origin();
```
gmta, line 149: Same here: `settings` gets dereferenced, but it's a `GC::Ptr` instead of a `GC::Ref` and as such it should either be checked for not being a nullptr, or changed to a `GC::Ref`.

Libraries/LibWeb/NotificationsAPI/Notification.h
```diff
@@ -51,10 +89,89 @@ class WEB_API Notification final : public DOM::EventTarget {
...
+        // It seems like Chrome is 2, Firefox is undefined, Safari is undefined
+        return 0;
+    }
+
+    String title() const { return m_notification.title; }
+    Bindings::NotificationDirection dir() const { return m_notification.direction; }
+    String lang() const { return m_notification.language; }
+    String body() const { return m_notification.body; }
+    String navigate() const { return m_notification.navigation_url.has_value() ? m_notification.navigation_url->serialize() : ""_string; }
+    String tag() const { return m_notification.tag; }
```
gmta: Can some of these return `String const&` instead to prevent copying strings?

Libraries/LibWeb/NotificationsAPI/Notification.h
```diff
@@ -51,10 +89,89 @@ class WEB_API Notification final : public DOM::EventTarget {
...
+    String image() const { return m_notification.image_url.has_value() ? m_notification.image_url->serialize() : ""_string; }
+    String icon() const { return m_notification.icon_url.has_value() ? m_notification.icon_url->serialize() : ""_string; }
+    String badge() const { return m_notification.badge_url.has_value() ? m_notification.badge_url->serialize() : ""_string; }
+    HighResolutionTime::EpochTimeStamp timestamp() const { return m_notification.timestamp; }
+    bool renotify() const { return m_notification.renotify_preference; }
+    Optional<bool> silent() const { return m_notification.silent_preference; }
+    bool require_interaction() const { return m_notification.require_interaction_preference; }
+    Vector<NotificationAction> actions() const
```
gmta: Let's put this relatively large method out of line in the `.cpp`, same for `data()`

Libraries/LibWeb/NotificationsAPI/Notification.h
```diff
@@ -51,10 +89,89 @@ class WEB_API Notification final : public DOM::EventTarget {
...
+    {
+        // 1. Let frozenActions be an empty list of type NotificationAction.
+        Vector<NotificationAction> frozen_actions;
```
gmta: You can call `.ensure_capacity()` on this vector since you know the size of `m_notifications.actions`.

Libraries/LibWeb/NotificationsAPI/Notification.idl
```diff
@@ -9,29 +9,29 @@ interface Notification : EventTarget {
...
-    [FIXME] readonly attribute DOMString tag;
-    [FIXME] readonly attribute USVString image;
-    [FIXME] readonly attribute USVString icon;
-    [FIXME] readonly attribute USVString badge;
+    readonly attribute DOMString title;
+    readonly attribute NotificationDirection dir;
+    readonly attribute DOMString lang;
+    readonly attribute DOMString body;
+    readonly attribute USVString navigate;
+    readonly attribute DOMString tag;
```
gmta: Can these be `Utf16DOMString` instead? That's the actual (and temporary) type that we should use for `DOMString` until we have fully migrated away from `String` for these APIs.

Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -351,7 +351,7 @@ WebIDL::ExceptionOr<GC::Ref<Document>> Document::create_and_initialize(Type type
     // AD-HOC: The response object no longer has an associated timing info object. For now, we use response's non-standard response time property,
```
gmta, line 360: Commit title:

> LibWeb: Bool to `CanUseCrossOriginIsolatedAPIs`

This is not an imperative sentence. Please write it as an actual change you're performing, e.g.:

> LibWeb: Use enum instead of bool for `CanUseCrossOriginIsolatedAPIs`

Libraries/LibWeb/NotificationsAPI/Notification.idl
NiccoloAntonelliDziri: Do you mean only inside of the .idl file ? I don't why it cannot be Utf16DOMString. I have just copied the API from the spec: https://notifications.spec.whatwg.org/#api

gmta: Yes, but our DOMString essentially results in the wrong type. Check out #5592 for how that came to be.

NiccoloAntonelliDziri: I have changed the `DOMString` s in the readonly attributes of the Notification interface. Do you want me to change also all of the other occurrences in the file ?

NiccoloAntonelliDziri:
Adding this comment to remove the stale label.

Libraries/LibWeb/NotificationsAPI/Notification.cpp
```diff
@@ -18,16 +22,182 @@ Notification::Notification(JS::Realm& realm)
...
+
+    // 9. Set notification’s body to options["body"].
+    notification.body = options.body;
+
+    auto try_set_url = [&base_url](String const& url) -> Optional<URL::URL> {
+        auto maybe_url = base_url.complete_url(url);
+        if (maybe_url.has_value())
+            return maybe_url.release_value();
+        return {};
+    };
```
AtkinsSJ: This is the same as just doing `base_url.complete_url(url)` isn't it? It's already an Optional unless I'm missing something. I think you can get rid of this lambda and just call that directly.

Libraries/LibWeb/NotificationsAPI/Notification.cpp
```diff
@@ -36,4 +206,47 @@ void Notification::initialize(JS::Realm& realm)
     Base::initialize(realm);
 }
 
+Vector<NotificationAction> Notification::actions() const
```
AtkinsSJ, line 203: Add a spec link here please.

Libraries/LibWeb/NotificationsAPI/Notification.cpp
```diff
@@ -36,4 +206,47 @@ void Notification::initialize(JS::Realm& realm)
...
+    return frozen_actions;
+}
+
+JS::Value Notification::data() const
+{
+    auto deserialized_data = HTML::structured_deserialize(vm(), m_notification.data, realm());
+    if (!deserialized_data.is_exception())
+        return deserialized_data.release_value();
+    return JS::js_null();
+}
```
AtkinsSJ, line 247: Please add a spec link and comment to this too. https://notifications.spec.whatwg.org/#dom-notification-data (We really like spec comments.)

Tests/LibWeb/Text/input/wpt-import/notifications/idlharness.https.any.js
```diff
@@ -0,0 +1,32 @@
...
+// META: script=/resources/WebIDLParser.js
+// META: script=/resources/idlharness.js
+
+'use strict';
+
+// https://notifications.spec.whatwg.org/
+
+idl_test(
+  ['notifications'],
+  ['service-workers', 'hr-time', 'html', 'dom'],
```
AtkinsSJ, line 11: For this test to function, you need to make sure these files are imported. They're all located at `https://wpt.live/interfaces/NAME.idl`, eg `https://wpt.live/interfaces/notifications.idl`. You can just save them to `Tests/LibWeb/Text/input/wpt-import/interfaces` and that should work.

Libraries/LibWeb/NotificationsAPI/Notification.cpp
AtkinsSJ: I meant that it should also have the spec text.
> The data getter steps are to return StructuredDeserialize(this’s notification’s data, this’s relevant Realm). If this throws an exception, then return null. 

It's not a numbered-step algorithm but the spec is still telling us what should happen, so it's useful to have that next to the code.

NiccoloAntonelliDziri: Oh, sorry about this I misunderstood.

PR #6436, opened 2025-10-10
LibWeb: Delete non-const `layout_node_with_style_and_box_metrics()`

kalenikaliaksandr:
...from PaintableBox. It was used exclusively to go to corresponding DOM node which could be done via direct DOM node pointer owned by paintable.


gmta:
CI is super mad

Libraries/LibWeb/Painting/PaintableBox.cpp
```diff
@@ -105,9 +105,10 @@ CSSPixelPoint PaintableBox::scroll_offset() const
     if (auto pseudo_element = node.generated_for_pseudo_element(); pseudo_element.has_value())
         return node.pseudo_element_generator()->scroll_offset(*pseudo_element);
 
+    if (!dom_node())
+        return {};
     if (auto const* element = as_if<DOM::Element>(*dom_node()))
```
gmta:
```
if (auto const* element = as_if<DOM::Element>(dom_node()))
```

If you use this check, you don't need the `if (!dom_node())` check.

Issue #351, opened 2024-07-02
Bold text is no longer displaying as bold


alec3660:
![not-so-bold](https://github.com/LadybirdBrowser/ladybird/assets/153211869/ddefa0b8-ac9f-4917-afff-ac615b52fe7d)

OS: Fedora 40 x86_64
Top: Firefox 127.0.2 (64-bit)
Bottom: Ladybird @ commit 09f7609

skyrising:
What is the source code for that HTML file?

alec3660:
> What is the source code for that HTML file?

`If I may be so <b>bold</b>`

skyrising:
Interestingly it works as expected for me
![image](https://github.com/LadybirdBrowser/ladybird/assets/4679350/4ee8697a-da58-475a-9304-69003c4027ef)

Do you know when this regressed for you?

One possibility I can think of is that font-config doesn't pick up the bold variant of your serif font.
Does `If I may be so <b style="font-family: sans-serif">bold</b>` look bold for you?

alec3660:
![Screenshot_2024-07-02_03-12-39](https://github.com/LadybirdBrowser/ladybird/assets/153211869/9fc93078-c966-48ff-9f75-e2e1b5509e79)

Same result with sans-serif. I'm unsure when the issue first occurred, I just happened to notice it while browsing.

AtkinsSJ:
My guess is somehow that font doesn't have a bold variant, or Ladybird isn't finding it... We need better font debugging tools.

Equinox4:
Note that the "b" tag has long ceased to mean "bold text", but rather " bring attention to". The browser is not required to apply any particular style to this tag by default.

skyrising:
While it's true that `<b>` nowadays has a semantic meaning rather than a stylistic one, the HTML spec still expects it to have `font-weight: bolder`: <https://html.spec.whatwg.org/multipage/rendering.html#phrasing-content-3>.
Which Ladybird has in [Default.css](https://github.com/LadybirdBrowser/ladybird/blob/af909784547c4f0ea8218ce464f45b3a1b0f4cb7/Userland/Libraries/LibWeb/CSS/Default.css#L276), so that's not the cause of this issue.

Equinox4:
[...]

PR #6439, opened 2025-10-10
LibWeb/LibGfx: Variable fonts support

Norbiros:
Resolves https://github.com/LadybirdBrowser/ladybird/issues/351

## Description
This PR adds support for [variable fonts](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_fonts/Variable_fonts_guide). It significantly improves font rendering accuracy, making sites like [nuxt.com](https://nuxt.com/) appear much closer to how they render in other engines.

Before:
Screenshot From 2025-10-09 15-55-59

After:
Screenshot From 2025-10-09 15-56-44

## ToDo
- [x] support `font-variation-settings`
- [x] properly handle `font-stretch`, `font-style` and `font-weight`
- [x] try detecting if font has different axies and then trying to load them

_Part 3 of resolving issues on https://hack4krak.pl/_


Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1728,8 +1730,10 @@ RefPtr<Gfx::FontCascadeList const> StyleComputer::find_matching_font_weight_asce
     auto pred = inclusive ? Fn([&](auto const& matching_font_candidate) { return matching_font_candidate.key.weight >= target_weight; })
                           : Fn([&](auto const& matching_font_candidate) { return matching_font_candidate.key.weight > target_weight; });
     auto it = find_if(candidates.begin(), candidates.end(), pred);
+    Vector<std::pair<Gfx::FourByteTag, float>> axes;
+    axes.append({ Gfx::MakeFourByteTag('w', 'g', 'h', 't'), static_cast<float>(target_weight) });
```
konradekk: you seem to bleed these details needlessly here: maybe a small structure holding some font details would do better? (constructing the tag even near the skia call-site? i guess that holding that in cache should be also okay).

Norbiros:
This implementation isn’t final. It will evolve over time, especially as font handling improves — including a proper font selection algorithm and full support for italic and oblique styles. I do not have enough time or knowledge to do it on my own, so I want to contribute this "scaffolding" which also makes a lot of websites look better in ladybird.

I think the fail is unrelated to my PR

Libraries/LibGfx/Font/FontVariationSettings.h
```diff
@@ -0,0 +1,72 @@
...
+#pragma once
+
+namespace Gfx {
+
+using FourByteTag = uint32_t;
+
+constexpr FourByteTag MakeFourByteTag(char a, char b, char c, char d)
+{
+    return (static_cast<uint32_t>(a) << 24) | (static_cast<uint32_t>(b) << 16) | (static_cast<uint32_t>(c) << 8) | static_cast<uint32_t>(d);
+}
```
AtkinsSJ: LibGfx actually has a type for this already: `FourCC`.

Libraries/LibGfx/Font/FontVariationSettings.h
```diff
@@ -0,0 +1,72 @@
...
+    }
+};
+
+// FIXME: Support other named axes like 'slnt', 'ital', 'opsz', 'GRAD', etc.
+struct FontVariationSettings {
+    HashMap<FourByteTag, float> axes;
+
+    FontVariationSettings() = default;
+
+    // https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_fonts/Variable_fonts_guide#weight
```
AtkinsSJ: Link to a spec where possible. (MDN is a guide but not a spec.) From that page I found [this](https://learn.microsoft.com/en-us/typography/opentype/spec/dvaraxisreg#registered-axis-tags) which has individual pages on some registered axes, so you could link to those.

Libraries/LibGfx/Font/FontVariationSettings.h
```diff
@@ -0,0 +1,72 @@
...
+    void weight(float value)
```
AtkinsSJ: `set_weight()` instead of `weight()`.

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1926,10 +1901,32 @@ RefPtr<Gfx::FontCascadeList const> StyleComputer::compute_font_for_style_values(
...
+                if (string_view.length() != 4)
+                    continue;
+
+                auto tag = Gfx::MakeFourByteTag(
+                    string_view[0],
+                    string_view[1],
+                    string_view[2],
+                    string_view[3]);
+
+                variation.axes.set(tag, value.value());
```
AtkinsSJ: Calling `value.value()` isn't safe - it's a `NumberOrCalculated` which might be a calculation. You should be able to call `value.resolved()` instead.

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -91,6 +91,7 @@
 #include <LibWeb/Namespace.h>
 #include <LibWeb/Painting/PaintableBox.h>
 #include <LibWeb/Platform/FontPlugin.h>
+#include <include/core/SkTypes.h>
```
kalenikaliaksandr: could we avoid including Skia header from StyleComputer?

Libraries/LibGfx/Font/TypefaceSkia.h
```diff
@@ -7,6 +7,8 @@
 #pragma once
 
 #include <LibGfx/Font/Typeface.h>
+#include <core/SkStream.h>
+#include <include/core/SkTypes.h>
```
kalenikaliaksandr: could we rearrange things so that Skia headers are included only from .cpp files? otherwise, we end up leaking Skia definitions, which is something we want to avoid.

Libraries/LibGfx/Font/Typeface.cpp
```diff
@@ -59,8 +61,15 @@ NonnullRefPtr<Font> Typeface::font(float point_size) const
     if (m_fonts.size() > max_cached_font_size_count)
         m_fonts.remove(m_fonts.begin());
 
-    auto font = adopt_ref(*new Font(*this, point_size, point_size));
-    m_fonts.set(point_size, font);
+    RefPtr<Typeface> used_typeface = const_cast<Typeface*>(this);
+    if (!variations.is_empty()) {
+        if (auto const* skia_typeface = dynamic_cast<TypefaceSkia const*>(this))
```
kalenikaliaksandr: it should be possible to use `as_if` instead of `dynamic_cast`.

Norbiros:
Changes:
- rebased on top of master (around 200 commits)
- resolved conflicts

Libraries/LibGfx/FourCC.h
```diff
@@ -42,7 +42,26 @@ struct [[gnu::packed]] FourCC {
...
 
+    bool operator<(FourCC const& other) const
+    {
+        for (int i = 0; i < 4; ++i) {
+            unsigned char a = static_cast<unsigned char>(cc[i]);
+            unsigned char b = static_cast<unsigned char>(other.cc[i]);
+            if (cc[i] != other.cc[i])
+                return a < b;
+        }
+        return false;
```
konradekk: might be a silly question: would it work if you compared it like so?

```suggestion
		return this.to_u32() < other.to_u32();
```

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1926,10 +1930,33 @@ RefPtr<Gfx::FontCascadeList const> StyleComputer::compute_font_for_style_values(
         auto result = Gfx::FontCascadeList::create();
         if (auto it = m_loaded_fonts.find(key); it != m_loaded_fonts.end()) {
             auto const& loaders = it->value;
+
+            Gfx::FontVariationSettings variation;
+            variation.set_weight(font_weight);
+            variation.set_width(font_width.value());
```
konradekk: there’s `int` version of this, `width`, above (the `FIXME` comment above it might be somewhat misleading – ignore if not relevant).

Norbiros: AFAIK variable font axis actually **do** support fractional weights, so there is no point to round it

Libraries/LibGfx/FourCC.h
Norbiros: When I was implementing it I wanted to be sure it will work on le, be systems etc, but now when I think about it, it just has to be ordered, not alphabetically

konradekk:
i don't have any further suggestions for improvements here. you'll need to wait for feedback from the maintainers, but it seems you've already earned their approval.

Libraries/LibGfx/Font/FontVariationSettings.h
```diff
@@ -0,0 +1,66 @@
...
+
+    bool operator==(FontVariationAxis const& other) const
+    {
+        return tag == other.tag && value == other.value;
+    }
+};
+
+// FIXME: Support other named axes like 'slnt', 'ital', 'opsz', 'GRAD', etc.
+struct FontVariationSettings {
+    HashMap<FourCC, float> axes;
```
kalenikaliaksandr, line 34: if we know all possible axes in advance, let's just preallocate a vector large enough to hold all of them instead of creating a HashMap, which would only ever hold at most two values.

Libraries/LibGfx/Font/Typeface.h
```diff
@@ -51,7 +76,7 @@ class Typeface : public RefCounted<Typeface> {
     virtual u16 width() const = 0;
     virtual u8 slope() const = 0;
 
-    [[nodiscard]] NonnullRefPtr<Font> font(float point_size) const;
+    [[nodiscard]] NonnullRefPtr<Font> font(float point_size, Gfx::FontVariationSettings const& variations = {}) const;
```
kalenikaliaksandr: you could drop Gfx prefix here, since we already in Gfx namespace.

Libraries/LibGfx/Font/TypefaceSkia.cpp
```diff
@@ -56,6 +56,36 @@ ErrorOr<NonnullRefPtr<TypefaceSkia>> TypefaceSkia::load_from_buffer(AK::Readonly
     return adopt_ref(*new TypefaceSkia { make<TypefaceSkia::Impl>(skia_typeface), buffer, ttc_index });
 }
 
+RefPtr<TypefaceSkia> TypefaceSkia::clone_with_variations(Vector<FontVariationAxis> const& axes) const
+{
+    if (axes.is_empty())
+        return const_cast<TypefaceSkia*>(this);
```
kalenikaliaksandr: could we return `RefPtr<TypefaceSkia const>` from this function instead of using `const_cast` here?

Libraries/LibGfx/Font/TypefaceSkia.h
```diff
@@ -7,6 +7,7 @@
 #pragma once
 
 #include <LibGfx/Font/Typeface.h>
+#include <core/SkStream.h>
```
kalenikaliaksandr: let's not leak Skia headers.

Libraries/LibGfx/Font/FontVariationSettings.h
Norbiros: As mentioned [here](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variation-settings#registered_and_custom_axes) we **do not** know all possible axes in advance. Some of them are standardized, but:

> Any axis can be used as long as it is given a unique 4-character axis

Libraries/LibGfx/Font/PathFontProvider.cpp
```diff
@@ -58,8 +58,12 @@ RefPtr<Gfx::Font> PathFontProvider::get_font(FlyString const& family, float poin
     if (it == m_typeface_by_family.end())
         return nullptr;
     for (auto const& typeface : it->value) {
-        if (typeface->weight() == weight && typeface->width() == width && typeface->slope() == slope)
-            return typeface->font(point_size);
+        if (typeface->weight() == weight && typeface->width() == width && typeface->slope() == slope) {
+            FontVariationSettings font_variation_settings;
+            font_variation_settings.set_weight(static_cast<float>(weight));
```
konradekk, line 63: i’m (slightly) concerned with missing width update here.

Norbiros: For now, only the `weight` property is supported in all places. Adding width is the next planned step, but it requires some refactoring of already existing font-related code, so I'll try to do it in the future. To avoid confusion I removed all uses of `set_width` method

Libraries/LibGfx/Font/FontVariationSettings.h
```diff
@@ -0,0 +1,66 @@
...
+    void set_width(float value)
+    {
+        axes.set(FourCC("wdth"), value);
+    }
+
+    bool is_empty() const { return axes.is_empty(); }
+
+    Vector<FontVariationAxis> to_sorted_list() const
+    {
+        Vector<FontVariationAxis> list;
```
gmta, line 54:
```cpp
list.ensure_capacity(axes.size());
//...
list.unchecked_append(..);
```

Libraries/LibGfx/Font/Typeface.h
```diff
@@ -34,6 +36,29 @@ struct ScaledFontMetrics {
...
+    float point_size;
+    Vector<FontVariationAxis> axes;
+
+    bool operator==(FontCacheKey const& other) const
+    {
+        if (point_size != other.point_size)
+            return false;
+        for (size_t i = 0; i < axes.size(); ++i)
+            if (axes[i] != other.axes[i])
+                return false;
```
gmta: I think you should be able to simply compare `axes == other.axes`?

Lubrsi:
Great! I think the main improvement that can be done for variable fonts is telling HarfBuzz about the variations, so it can shape text with them.

See: https://harfbuzz.github.io/fonts-and-faces-variable.html

Libraries/LibWeb/CSS/StyleComputer.h
```diff
@@ -207,7 +207,7 @@ class WEB_API StyleComputer final : public GC::Cell {
     static NonnullRefPtr<StyleValue const> compute_font_style(NonnullRefPtr<StyleValue const> const& specified_value, ComputationContext const&);
     static NonnullRefPtr<StyleValue const> compute_font_weight(NonnullRefPtr<StyleValue const> const& specified_value, double inherited_font_weight, ComputationContext const&);
     static NonnullRefPtr<StyleValue const> compute_font_width(NonnullRefPtr<StyleValue const> const& specified_value, ComputationContext const&);
-    static NonnullRefPtr<StyleValue const> compute_font_variation_settings(NonnullRefPtr<StyleValue const> const& absolutized_value);
+    static NonnullRefPtr<StyleValue const> compute_font_variation_settings(NonnullRefPtr<StyleValue const> const& absolutized_value, ComputationContext const&);
```
AtkinsSJ: Nit: Change `absolutized_value` to `specified_value`.

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -3201,7 +3207,7 @@ NonnullRefPtr<StyleValue const> StyleComputer::compute_animation_name(NonnullRef
     });
 }
 
-NonnullRefPtr<StyleValue const> StyleComputer::compute_font_variation_settings(NonnullRefPtr<StyleValue const> const& absolutized_value)
+NonnullRefPtr<StyleValue const> StyleComputer::compute_font_variation_settings(NonnullRefPtr<StyleValue const> const& absolutized_value, ComputationContext const& computation_context)
```
AtkinsSJ: Same thing about the `absolutized_value` name.

I think if you put
```c++
    auto const& absolutized_value = specified_value->absolutized(computation_context);
```
at the start of the method, it'll match the others a bit better, you won't have to call `absolutized()` on the individual axis_tags below, and it might catch other things that needs absolutizing.

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1825,7 +1829,7 @@ CSSPixels StyleComputer::relative_size_mapping(RelativeSize relative_size, CSSPi
     VERIFY_NOT_REACHED();
 }
 
-RefPtr<Gfx::FontCascadeList const> StyleComputer::compute_font_for_style_values(StyleValue const& font_family, CSSPixels const& font_size, int slope, double font_weight, Percentage const& font_width) const
+RefPtr<Gfx::FontCascadeList const> StyleComputer::compute_font_for_style_values(StyleValue const& font_family, CSSPixels const& font_size, int slope, double font_weight, Percentage const& font_width, HashMap<FlyString, NumberOrCalculated> font_variation_settings, Length::ResolutionContext const& length_resolution_context) const
```
AtkinsSJ: I think you want to pass `font_variation_settings` as `const&` here.

Libraries/LibGfx/Font/TypefaceSkia.cpp
```diff
@@ -56,6 +57,36 @@ ErrorOr<NonnullRefPtr<TypefaceSkia>> TypefaceSkia::load_from_buffer(AK::Readonly
...
+        return this;
+
+    SkFontArguments font_args;
+
+    Vector<SkFontArguments::VariationPosition::Coordinate> coords;
+    coords.resize(axes.size());
+    for (size_t i = 0; i < axes.size(); ++i) {
+        coords[i].axis = axes[i].tag.to_u32();
+        coords[i].value = axes[i].value;
+    }
```
AtkinsSJ: Nit: `ensure_capacity()` and `unchecked_append()` would be more idiomatic.

PR #6442, opened 2025-10-10
LibWeb: Implement WebAudio FFT using fftw from vcpkg

tete17:
We still lack a good complex type but an Array<f32,2> will do for now.

We have a special case for purely real FFT, but we still probably spend more in copying in and out the data :)

Ref: 
 * https://www.fftw.org/fftw3_doc/Complex-One_002dDimensional-DFTs.html#Complex-One_002dDimensional-DFTs
 * https://www.fftw.org/fftw3_doc/One_002dDimensional-DFTs-of-Real-Data.html#One_002dDimensional-DFTs-of-Real-Data

### DIFF
--- a/Libraries/LibWeb/CMakeLists.txt
+++ b/Libraries/LibWeb/CMakeLists.txt
@@ -1122,8 +1122,9 @@ set(GENERATED_SOURCES
 ladybird_lib(LibWeb web EXPLICIT_SYMBOL_EXPORT)
 
 find_package(LibXml2 REQUIRED)
+find_package(FFTW3f CONFIG REQUIRED)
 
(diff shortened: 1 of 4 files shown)


gmta:
We could revive `AK/Complex.h`?
https://github.com/SerenityOS/serenity/blob/master/AK/Complex.h

tete17:
> We could revive `AK/Complex.h`? https://github.com/SerenityOS/serenity/blob/master/AK/Complex.h

Happy to do so but we would then still incur in the copy cost when moving data to fftw right?

Zaggy1024:
I think it should be possible to statically assert that the data layout of Complex matches what the API expects, so you can just reinterpret-cast it?

Also, I'm wondering if it would be possible to implement this in LibMedia via FFmpeg rather than bringing in a new library: https://www.ffmpeg.org/doxygen/0.7/avfft_8h.html

cqundefine:
Just pointing out that fftw3 is GPL-2.0 licensed so I'm not sure that it's viable for use in here as the Distribution preset uses static libraries which might cause license issues.

PR #6446, opened 2025-10-10
LibWeb: Add missing SVG presentation attributes

tcl3:
This allows `font-variant` and `text-decoration` to be used as presentation attributes in SVGs.


tcl3:
My latest push adds support for the `font-stretch` presentation attribute, which maps to the `font-width` property.

PR #6449, opened 2025-10-11
LibWeb/CSS: Check overflow value before determining box baseline

ArmindoFlores:
The CSS spec says the baseline of an inline-block should be the bottom margin when either the overflow property is not 'visible' or there are no in-flow line boxes. Previously, only the latter case was checked.

This fixes 1 WPT test:
https://wpt.live/css/css-align/baseline-of-scrollable-1a.html

### DIFF
--- a/Libraries/LibWeb/Layout/FormattingContext.cpp
+++ b/Libraries/LibWeb/Layout/FormattingContext.cpp
@@ -1827,7 +1827,10 @@ CSSPixels FormattingContext::box_baseline(Box const& box) const
         }
     }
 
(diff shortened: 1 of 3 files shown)


tcl3:
Welcome to the project! It would be great if you could import the newly-passing WPT test into our test suite. You should be able to do that with this command: `./Meta/WPT.sh update && ./Meta/WPT.sh import css/css-align/baseline-of-scrollable-1a.html`.

Once the test is imported you should amend your existing commit to add the test.

ArmindoFlores:
Thank you, glad to be starting out. I added the test as asked.

PR #6452, opened 2025-10-11
LibWeb: Avoid crash evaluating media query in document lacking window

Calme1709:
In some cases a document may lack an associated window - to fix this for now we just return false but perhaps there are some media queries we should still attempt to resolve.

Fixes a crash introduced in 05c336e

### DIFF
--- a/Libraries/LibWeb/CSS/MediaQuery.cpp
+++ b/Libraries/LibWeb/CSS/MediaQuery.cpp
@@ -99,7 +99,12 @@ String MediaFeature::to_string() const
 MatchResult MediaFeature::evaluate(DOM::Document const* document) const
 {
     VERIFY(document);
-    VERIFY(document->window());
+
+    // FIXME: In some cases (e.g. when parsing HTML using DOMParser::parse_from_string()) a document may not be associated with a window -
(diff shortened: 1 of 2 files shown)


Libraries/LibWeb/CSS/MediaQuery.cpp
```diff
@@ -99,7 +99,12 @@ String MediaFeature::to_string() const
 MatchResult MediaFeature::evaluate(DOM::Document const* document) const
 {
     VERIFY(document);
-    VERIFY(document->window());
+
+    // FIXME: In some cases (e.g. when parsing HTML using DOMParser::parse_from_string()) a document may not be associated with a window -
+    //        for now we just return false but perhaps there are some media queries we should still attempt to resolve.
```
AtkinsSJ, line 104: I think this is fine as-is and doesn't need a FIXME... IIUC, any `<style>` elements in that state have no effect on the document, so we don't need to evaluate media queries for them. If and when they do get attached to a live document, they'll get evaluated then.

Calme1709: Does not having effect also apply to other places we evaluate media queries (e.g.  sizes attribute, @import conditions, etc)?

PR #6453, opened 2025-10-12
LibGfx: Add robust collinear line intersection

Dowsley:
# LibGfx: Add robust collinear line intersection + tests for Line

## Summary

Previously, the method `Gfx::Line<T>::intersected` only covered collinear lines (segments lying on the same infinite line) if they had shared endpoints. Leaving completely uncovered the case that two collinear lines can overlap without having two perfectly coexisting endpoints.

This change aims to strengthen this method by handling all collinear-line intersection configurations. Now after the changes, when the two lines are collinear, the will return a single representative point (or no point) out of these 3 paths:
1. The midpoint of the overlap when the overlap has non-zero length
2. The shared endpoint when they only touch endpoints
3. None, if the collinear lines are disjointed

This preserves previous functionality. To guarantee this, a new LibGfx test suite for Gfx::Line was added.

## Changes
- Implemented collinear intersection via parameter-space overlap (see references: [one](https://en.wikipedia.org/wiki/Line%E2%80%93line_intersection#Given_two_points_on_each_line_segment) and [two](https://en.wikipedia.org/wiki/Intersection_(geometry)#Two_line_segments)):
  - Added `collinear_intersection_point` helper using dot-product parameterization `t = dot(p - A0, r) / dot(r, r)`.
  - For two collinear segments:
    - If they overlap with non-zero length, return the midpoint of the overlap (deterministic representative)
    - If they only touch at a single endpoint, return that endpoint
    - If they are disjoint, return empty
  - Effectively replaces previous ad-hoc checks and completes the prior TODO

- Clarified collinearity handling in `intersected`:
  - `denom == 0 && num == 0` -> collinear -> delegated to the new helper.
  - Non-collinear endpoint-only intersections remain handled by the general `t,u ∈ [0,1]` branch

- Tests: Added `Tests/LibGfx/TestLine.cpp` and registered it in `Tests/LibGfx/CMakeLists.txt`
  - Non-collinear:
    - Endpoint intersection (returns shared endpoint)
    - Proper interior intersection
    - Parallel with no intersection
  - Collinear:
    - Total containment (horizontal)
    - Total containment (diagonal)
    - Partial overlaps (left/right extrude)
    - No overlap (collinear)
    - Shared endpoint (horizontal && diagonal)
    - Zero-length (point) on segment (hit) and off segment (miss)
    - Deterministic perfect overlap (identical segments reversed) -> returns midpoint

## Verification
```
/Users/dowsley/Code/ladybird/Build/release/bin/TestLine
[...]
Finished 13 tests and 0 benchmarks in 0ms (0ms tests, 0ms benchmarks, 0ms other).
All 13 tests passed.
```

## Notes
- Existing integer rounding TODO is preserved: integer variants still use truncation when mapping back from `t`. A follow-up could define a rounding policy if needed (?)
- No behavior change for non-collinear intersections
- Usually I'm not a fan of leaving lots of comments, but since this involves non-intuitive math (for non-intuitive purposes), I preferred leaving them justifying the choices at every few lines for the helper function.

### DIFF
--- a/Libraries/LibGfx/Line.h
+++ b/Libraries/LibGfx/Line.h
@@ -8,6 +8,7 @@
 
 #include <AK/ByteString.h>
 #include <AK/Format.h>
+#include <AK/Math.h>
 #include <AK/Optional.h>
 #include <LibGfx/Forward.h>
 #include <LibGfx/Point.h>
@@ -54,40 +55,33 @@ public:
         auto delta_a = other.m_a - m_a;
         auto num = cross_product(delta_a, r);
         auto denom = cross_product(r, s);
-        if (denom == 0) {
-            if (num == 0) {
-                // Lines are collinear, check if line ends are touching
-                if (m_a == other.m_a || m_a == other.m_b)
-                    return m_a;
-                if (m_b == other.m_a || m_b == other.m_b)
-                    return m_b;
-                // Check if they're overlapping
-                if (!(m_b.x() - m_a.x() < 0 && m_b.x() - other.m_a.x() < 0 && other.m_b.x() - m_a.x() && other.m_b.x() - other.m_a.x())) {
-                    // Overlapping
-                    // TODO find center point?
-                }
-                if (!(m_b.y() - m_a.y() < 0 && m_b.y() - other.m_a.y() < 0 && other.m_b.y() - m_a.y() && other.m_b.y() - other.m_a.y())) {
-                    // Overlapping
-                    // TODO find center point?
-                }
-                return {};
-            } else {
-                // Lines are parallel and not intersecting
-                return {};
-            }
+
+        bool parallel = denom == 0;
+        bool collinear = num == 0;
+        if (parallel) {
+            if (collinear)
+                return collinear_intersection_point(other);
+            return {};
         }
-        auto u = static_cast<float>(num) / static_cast<float>(denom);
-        if (u < 0.0f || u > 1.0f) {
+
+        using Calc = Conditional<IsFloatingPoint<T>, T, double>;
+
+        auto u = static_cast<Calc>(num) / static_cast<Calc>(denom);
+        if (u < static_cast<Calc>(0) || u > static_cast<Calc>(1)) {
             // Lines are not parallel and don't intersect
             return {};
         }
-        auto t = static_cast<float>(cross_product(delta_a, s)) / static_cast<float>(denom);
-        if (t < 0.0f || t > 1.0f) {
+        auto t = static_cast<Calc>(cross_product(delta_a, s)) / static_cast<Calc>(denom);
+        if (t < static_cast<Calc>(0) || t > static_cast<Calc>(1)) {
             // Lines are not parallel and don't intersect
             return {};
         }
-        // TODO: round if we're dealing with int
-        return Point<T> { m_a.x() + static_cast<T>(t * r.x()), m_a.y() + static_cast<T>(t * r.y()) };
+        auto x = static_cast<Calc>(m_a.x()) + t * static_cast<Calc>(r.x());
+        auto y = static_cast<Calc>(m_a.y()) + t * static_cast<Calc>(r.y());
+        if constexpr (IsFloatingPoint<T>)
+            return Point<T> { static_cast<T>(x), static_cast<T>(y) };
+        else
+            return Point<T> { round_to<T>(x), round_to<T>(y) };
     }
 
     float length() const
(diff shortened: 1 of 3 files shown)


Libraries/LibGfx/Line.h
```diff
@@ -168,6 +155,56 @@ class Line {
...
+    // Compute the 1D overlap via the segment parameter t (dot-product projection onto r),
+    // then return the midpoint in t mapped back to 2D. If overlap length is zero, this is
+    // the shared endpoint; if there is no overlap, return empty.
+    Optional<Point<T>> collinear_intersection_point(Line const& other) const
+    {
+        auto r = m_b - m_a;
+        auto s = other.m_b - other.m_a;
+
+        auto dot = [](auto const& p, auto const& q) {
+            return static_cast<double>(p.x()) * static_cast<double>(q.x())
```
Hendiadyoin1: forcing `double` seems a bit overkill, especially if it is already a float

Hendiadyoin1: Also none of the algorithms described in the old fixmes this is replacing inherently need floating point algorithms as far as I can tell

Dowsley: I will double check if we really need those to be floats

Dowsley: I am reverting the double part to use float instead.

The floating point is being used for convenience to compute parameters t0, t1 in [0,1] and find their midpoint, so I personally think it's justified and clear from first glance how it works.

Hendiadyoin1: Whoops forgot to answer,
So as far as I can tell this PR is to resolve the 2 FIXMEs in the old code,
but as far as I can see resolving those does not inherently require floating point precision, and there should be a general solution using integer math
If I am not mistaken taking the midpoint of the x-intersection range and y-intersection range separately should work,
which can be tested for easily

Hendiadyoin1: In any case the codepath should be agnostic to the type that was handed in, so `Line<double>` should not demote to `floats` in the calculations

Dowsley: In that case, I have refactored the code. Took the opportunity to solve the other TODOs related to types - hopefully not too much scope creep. They were related anyway.

Libraries/LibGfx/Line.h
```diff
@@ -54,40 +55,33 @@ class Line {
...
+        if (t < static_cast<Calc>(0) || t > static_cast<Calc>(1)) {
             // Lines are not parallel and don't intersect
             return {};
         }
-        // TODO: round if we're dealing with int
-        return Point<T> { m_a.x() + static_cast<T>(t * r.x()), m_a.y() + static_cast<T>(t * r.y()) };
+        auto x = static_cast<Calc>(m_a.x()) + t * static_cast<Calc>(r.x());
+        auto y = static_cast<Calc>(m_a.y()) + t * static_cast<Calc>(r.y());
+        if constexpr (IsFloatingPoint<T>)
+            return Point<T> { static_cast<T>(x), static_cast<T>(y) };
```
Hendiadyoin1, line 82: I think these casts are superfluous here

as Calc is T in the FloatingPoint case

Libraries/LibGfx/Line.h
```diff
@@ -104,15 +98,20 @@ class Line {
...
+        if (param > static_cast<Calc>(1))
             return m_b;
-        // TODO: round if we're dealing with int
-        return { static_cast<T>(m_a.x() + param * delta_c), static_cast<T>(m_a.y() + param * delta_d) };
+        auto rx = static_cast<Calc>(m_a.x()) + param * static_cast<Calc>(delta_c);
+        auto ry = static_cast<Calc>(m_a.y()) + param * static_cast<Calc>(delta_d);
+        if constexpr (IsFloatingPoint<T>)
+            return { static_cast<T>(rx), static_cast<T>(ry) };
+        else
+            return { round_to<T>(rx), round_to<T>(ry) };
```
Hendiadyoin1, line 114: Maybe do this  in a separate commit

Libraries/LibGfx/Line.h
```diff
@@ -168,6 +167,57 @@ class Line {
     ByteString to_byte_string() const;
 
 private:
+    /*
+     * @brief Return a single point representing the intersection of two collinear segments.
+     * Compute the midpoint of the overlap in X and Y independently, which maps to the
+     * midpoint of the overlapping segment for collinear lines. If there is no overlap,
+     * return empty.
+     */
```
Hendiadyoin1, line 175: We dont do annotations in this style, usually

Libraries/LibGfx/Line.h
```diff
@@ -168,6 +167,57 @@ class Line {
...
+
+        auto bx0 = min(other.m_a.x(), other.m_b.x());
+        auto bx1 = max(other.m_a.x(), other.m_b.x());
+        auto by0 = min(other.m_a.y(), other.m_b.y());
+        auto by1 = max(other.m_a.y(), other.m_b.y());
+
+        auto ox0 = max(ax0, bx0);
+        auto ox1 = min(ax1, bx1);
+        auto oy0 = max(ay0, by0);
+        auto oy1 = min(ay1, by1);
```
Hendiadyoin1, line 204: getting really nitpicky now, but maybe add underscores like `a_x0,...`, but thats probably opinionated

Libraries/LibGfx/Line.h
```diff
@@ -168,6 +167,57 @@ class Line {
...
+
+        if (ox1 < ox0 || oy1 < oy0)
+            return {};
+
+        // midpoint helper avoiding overflow for integers: start + (end - start) / 2
+        auto midpoint = [](auto const& start, auto const& end) {
+            if constexpr (IsFloatingPoint<T>)
+                return static_cast<T>((start + end) / static_cast<T>(2));
+            else
+                return static_cast<T>(start + (end - start) / 2);
```
Hendiadyoin1, line 214: To get really really nit picky, we could do a fair-er rounding here, emulating a break tie, but thats likely overkill

Libraries/LibGfx/Line.h
```diff
@@ -54,40 +55,33 @@ class Line {
...
+        bool collinear = num == 0;
+        if (parallel) {
+            if (collinear)
+                return collinear_intersection_point(other);
+            return {};
         }
-        auto u = static_cast<float>(num) / static_cast<float>(denom);
-        if (u < 0.0f || u > 1.0f) {
+
+        using Calc = Conditional<IsFloatingPoint<T>, T, double>;
```
Hendiadyoin1, line 67: the default seemed to be `float` previously, any particular reason for the promotion?

Libraries/LibGfx/Line.h
konradekk: (a test case for that would be helpful here! 🫣)

PR #6454, opened 2025-10-12
LibWeb: Ignore non-finite args in CanvasRenderingContext2d::clear_rect()

ljamar:


Libraries/LibWeb/HTML/CanvasRenderingContext2D.cpp
```diff
@@ -100,8 +100,13 @@ void CanvasRenderingContext2D::fill_rect(float x, float y, float width, float he
     fill_internal(rect_path(x, y, width, height), Gfx::WindingRule::EvenOdd);
 }
 
+// https://html.spec.whatwg.org/multipage/canvas.html#dom-context-2d-clearrect
 void CanvasRenderingContext2D::clear_rect(float x, float y, float width, float height)
 {
+    // 1. If any of the arguments are infinite or NaN, then return.
+    if (isinf(x) || isinf(y) || isinf(width) || isinf(height) || isnan(x) || isnan(y) || isnan(width) || isnan(height))
```
Lubrsi: suggestion: `isfinite` covers both Infinity and NaN, so these can be consolidated into one isfinite per argument.

PR #6462, opened 2025-10-13
LibDevTools: Show the accessibility tree in DevTools

AtkinsSJ:
image

Being able to inspect the accessibility tree is something we've been missing since switching to DevTools, so here it is! A lot of things need populating before they can be properly useful, but having the tree show up at all and be inspectable is a good starting point.

### DIFF
--- a/Libraries/LibDevTools/Actors/WalkerActor.cpp
+++ b/Libraries/LibDevTools/Actors/WalkerActor.cpp
@@ -6,6 +6,7 @@
 
 #include <AK/JsonArray.h>
 #include <AK/StringUtils.h>
+#include <LibDevTools/Actors/AccessibilityNodeActor.h>
 #include <LibDevTools/Actors/LayoutInspectorActor.h>
 #include <LibDevTools/Actors/TabActor.h>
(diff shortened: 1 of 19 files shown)


Libraries/LibDevTools/Actors/WalkerActor.cpp
```diff
@@ -127,6 +128,55 @@ void WalkerActor::handle_message(Message const& message)
...
+                    node.set("newParents"sv, JsonArray {});
+
+                    response.set("node"sv, move(node));
+                    send_response(message, move(response));
+                    return;
+                }
+            }
+        }
+
+        // We don't know what's being requested.
```
trflynn89: I don't think code comments like these add much value, we already have that context from the error message below

AtkinsSJ: That's completely fair, I added most of these while I was trying to figure out what was going on, and should have removed a lot of them.

AtkinsSJ:
Changes: Rebased, and removed that one redundant comment. (I reviewed the other comments and they seem useful.)

PR #6468, opened 2025-10-14
LibRegex: Add recursion depth limit to prevent stack overflow …

MrBrain295:
…in regex parser

I wanted to see if copilot could fix a real issue. I ran the tests for regexp and they all passed.

### DIFF
--- a/Libraries/LibRegex/RegexParser.cpp
+++ b/Libraries/LibRegex/RegexParser.cpp
@@ -993,6 +993,12 @@ bool ECMA262Parser::has_duplicate_in_current_alternative(FlyString const& name)
 
 bool ECMA262Parser::parse_disjunction(ByteCode& stack, size_t& match_length_minimum, ParseFlags flags)
 {
+    // Check recursion depth to prevent stack overflow
+    if (m_recursion_depth >= MAX_RECURSION_DEPTH)
+        return set_error(Error::ReachedMaxRecursion);
+
+    ++m_recursion_depth;
+
     size_t total_match_length_minimum = NumericLimits<size_t>::max();
     Vector<ByteCode> alternatives;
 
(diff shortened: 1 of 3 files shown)


Libraries/LibRegex/RegexParser.cpp
```diff
@@ -993,6 +993,12 @@ bool ECMA262Parser::has_duplicate_in_current_alternative(FlyString const& name)
 
 bool ECMA262Parser::parse_disjunction(ByteCode& stack, size_t& match_length_minimum, ParseFlags flags)
 {
+    // Check recursion depth to prevent stack overflow
+    if (m_recursion_depth >= MAX_RECURSION_DEPTH)
+        return set_error(Error::ReachedMaxRecursion);
+
+    ++m_recursion_depth;
```
alimpfard, line 1000: This + the `--`s somewhere down there:
```suggestion
    TemporaryChange recursion_depth_change { m_recursion_depth, m_recursion_depth + 1 };
```

Libraries/LibRegex/RegexParser.h
```diff
@@ -319,6 +319,10 @@ class REGEX_API ECMA262Parser final : public Parser {
     // by requiring that (...)+ only contain the matches for the last iteration.
     // To do that, we have to keep track of which capture groups are "in scope", so we can clear them as needed.
     Vector<Vector<size_t>> m_capture_groups_in_scope;
+
+    // Recursion depth tracking to prevent stack overflow
+    size_t m_recursion_depth { 0 };
```
alimpfard, line 324: Why only on that specific production? is there no other recursive rule?

Libraries/LibRegex/RegexParser.h
```diff
@@ -319,6 +319,10 @@ class REGEX_API ECMA262Parser final : public Parser {
...
+    static constexpr size_t MAX_RECURSION_DEPTH = 512;
```
alimpfard, line 325: this should just be in the C++ file, it's an impl detail no one needs to know about.

alimpfard (changes requested):
as far as I can tell, this is a _very_ topical fix, and I'm not actually sold on limiting to 512 anyhow.
I think I'd rather see StackInfo used to get an actual "real" limit instead of whatever arbitrary choice this 512 is.

alimpfard:
oh also, the category should not be LibJS, this is a LibRegex fix.

PR #6469, opened 2025-10-14
LibWeb: Count the width of inline-blocks in InlineNodes only once

### LINKED ISSUE #567: Wrong layout of minimal page with `<span style="display: inline-block;">`
I have extracted this minimal example from Wikipedia:
```html
<!DOCTYPE html>
<html>
<head>
<style>
* {
  border: 1px solid black;
  padding: 15px;
}
</style>
</head>
<body>
<span style="background-color: yellow; border: 2px solid red;">
  <span style="display: inline-block;">aaa</span>
  <span>bbb</span>
</span>
```

Current layout in Ladybird:
ladybird1
or
ladybird2
(Page reload sometimes toggles between the two layouts.)

Expected layout (from Firefox):
firefox

lpas:
An BlockContainer inside an InlineNode is called from the `for each in inclusive_subtree_of_type`  but is also a fragment
 of that InlineNode. Don't count the the Node twice.

[‎Tests/LibWeb/Layout/input/block-and-inline/inline-block-contained-by-inline.html](https://github.com/LadybirdBrowser/ladybird/pull/6469/files#diff-86e89ca82774d7ee06527a70c637cbab4945a56fff6791e3d58c33feb55a8790)

Before: 
image
After: 
image

fixes #567

### DIFF
--- a/Tests/LibWeb/Layout/expected/abspos-static-position-containing-block-inline-situation.txt
+++ b/Tests/LibWeb/Layout/expected/abspos-static-position-containing-block-inline-situation.txt
@@ -1,6 +1,6 @@
 Viewport <#document> at [0,0] [0+0+0 800 0+0+0] [0+0+0 600 0+0+0] children: not-inline
   BlockContainer <html> at [0,0] [0+0+0 800 0+0+0] [0+0+0 18 0+0+0] [BFC] children: inline
-    InlineNode <body> at [0,13] [8+0+0 36.84375 0+0+8] [8+0+0 0 0+0+8]
+    InlineNode <body> at [8,13] [8+0+0 0 0+0+8] [8+0+0 0 0+0+8]
       frag 0 from BlockContainer start: 0, length: 0, rect: [8,13 0x0] baseline: 0
       BlockContainer <main> at [8,13] inline-block [0+0+0 0 0+0+0] [0+0+0 0 0+0+0] [BFC] children: not-inline
         BlockContainer <div> at [8,0] positioned [0+0+0 36.84375 0+0+0] [0+0+0 18 0+0+0] [BFC] children: inline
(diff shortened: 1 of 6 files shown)


Tests/LibWeb/Layout/expected/abspos-static-position-containing-block-inline-situation.txt
```diff
@@ -1,6 +1,6 @@
 Viewport <#document> at [0,0] [0+0+0 800 0+0+0] [0+0+0 600 0+0+0] children: not-inline
   BlockContainer <html> at [0,0] [0+0+0 800 0+0+0] [0+0+0 18 0+0+0] [BFC] children: inline
-    InlineNode <body> at [0,13] [8+0+0 36.84375 0+0+8] [8+0+0 0 0+0+8]
+    InlineNode <body> at [8,13] [8+0+0 0 0+0+8] [8+0+0 0 0+0+8]
```
gmta, line 3: This doesn't seem to be correct. `document.body.offsetHeight` returns `0` now. In Firefox, I get `16`.

lpas: yes, you are right but I don't think this comes from this change, seems to be a bug before and still after the fix. 

But the behavior didn't seem to change there. If we have a html like 
```html
<!doctype html><style>body {display: inline;position: relative;}main {display: inline-block;}</style><body><main>
```

wen get 0 in LB before and after my change and in Firefox a 17

gmta: Oh apologies! Misread that width as height, we now actually match Firefox' width of 0.

gmta (approved):

PR #6471, opened 2025-10-14
LibWeb: Do not execute unclosed SVG script tags

lpas:


Libraries/LibWeb/HTML/Parser/HTMLParser.cpp
```diff
@@ -4777,7 +4783,7 @@ void HTMLParser::process_using_the_rules_for_foreign_content(HTMLToken& token)
     }
 
     // -> An end tag whose tag name is "script", if the current node is an SVG script element
-    if (token.is_end_tag() && current_node()->namespace_uri() == Namespace::SVG && current_node()->local_name() == SVG::TagNames::script) {
+    if (token.is_end_tag() && token.tag_name() == SVG::TagNames::script && current_node()->namespace_uri() == Namespace::SVG && current_node()->local_name() == SVG::TagNames::script) {
```
tcl3, line 4787: I don't think the `current_node()->local_name() == SVG::TagNames::script` check is necessary here.

lpas: That is what the spec line above says. And we would crash for an malformed svg for instance: 

```html 
<html><svg></script></svg></html>
```

gmta (dismissed):

Libraries/LibWeb/HTML/Parser/HTMLParser.cpp
gmta, line 1: Commit message: please avoid [using weasel words like "fix"](https://github.com/LadybirdBrowser/ladybird/blob/master/CONTRIBUTING.md#code-submission-policy). It's better to describe what you change + why you do so.

gmta: Suggestion:

> LibWeb: Do not execute unclosed SVG script tags

PR #6472, opened 2025-10-14
LibIPC: Simplify IPC read hook

kalenikaliaksandr:
- Return `PeerEOF` enum instead of `Error` containing string from `drain_messages_from_peer()`. There are no other error types to return from this function, so boolean-like enum is sufficient.
- Don't override read hook in `ConnectionFromClient` constructor. It was previously redefined only to suppress EOF error returned by `drain_messages_from_peer()`.
- Remove call to `handle_messages()` from read hook, because it's already called by `drain_messages_from_peer()`.

### DIFF
--- a/Libraries/LibIPC/Connection.cpp
+++ b/Libraries/LibIPC/Connection.cpp
@@ -21,10 +21,7 @@ ConnectionBase::ConnectionBase(IPC::Stub& local_stub, NonnullOwnPtr<Transport> t
 {
     m_transport->set_up_read_hook([this] {
         NonnullRefPtr protect = *this;
-
-        if (auto result = drain_messages_from_peer(); result.is_error())
-            dbgln("Read hook error while draining messages: {}", result.error());
-
+        drain_messages_from_peer();
         handle_messages();
     });
 }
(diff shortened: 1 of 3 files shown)


Libraries/LibIPC/Connection.cpp
```diff
@@ -21,11 +21,7 @@ ConnectionBase::ConnectionBase(IPC::Stub& local_stub, NonnullOwnPtr<Transport> t
 {
     m_transport->set_up_read_hook([this] {
         NonnullRefPtr protect = *this;
-
-        if (auto result = drain_messages_from_peer(); result.is_error())
-            dbgln("Read hook error while draining messages: {}", result.error());
-
-        handle_messages();
```
gmta, line 28: Looks like the default behavior for non-ConnectionFromClient classes was to immediately handle messages instead of through deferred_invoke, is that change intended?

kalenikaliaksandr: it was intentional, but I didn't realize going processing messages from the hook vs using deferred invoke would make much difference. I just found out that actually it does and resizing becomes unresponsive if I remove message processing from the hook. I am really a bit puzzled why this happens. would be really interesting to figure why because that's the problem I am dealing with in my ongoing work on making TransportSocket do reading & writing on the separate thread.

kalenikaliaksandr: hmmm, somehow EventLoopManagerMacOS delays all `deferred_invoke()` callbacks until resizing is done.

gmta (approved):

PR #6477, opened 2025-10-14
LibWeb: Implement the `font-size-adjust` property

tcl3:
This PR implements the `font-size-adjust` property, which can be used to control how fonts are scaled based on various metrics. It can be used to ensure fallback fonts remain readable when their metrics differ significantly from the preferred font.

I believe the reason that some of the relevant reftests don't pass is that we are using font metrics that have been converted to CSSPixels, which results in noticable precision loss.

I'm also still trying to track down why I'm not getting the correct values when using `from-font`.

Results in a diff of +743/-7 WPT subtests in the `css/css-fonts` directory.

### DIFF
--- a/Libraries/LibGfx/Font/Font.cpp
+++ b/Libraries/LibGfx/Font/Font.cpp
@@ -45,6 +45,7 @@ Font::Font(NonnullRefPtr<Typeface const> typeface, float point_width, float poin
     metrics.x_height = skMetrics.fXHeight;
     metrics.advance_of_ascii_zero = font.measureText("0", 1, SkTextEncoding::kUTF8);
     metrics.ascent = -skMetrics.fAscent;
+    metrics.cap_height = skMetrics.fCapHeight;
     metrics.descent = skMetrics.fDescent;
     metrics.line_gap = skMetrics.fLeading;
 
--- a/Libraries/LibGfx/Font/Font.h
+++ b/Libraries/LibGfx/Font/Font.h
@@ -29,6 +29,9 @@ struct FontPixelMetrics {
     // Number of pixels the font extends above the baseline.
     float ascent { 0 };
 
+    // Height of an uppercase letter.
+    float cap_height { 0 };
+
     // Number of pixels the font descends below the baseline.
     float descent { 0 };
 
--- a/Libraries/LibWeb/CSS/Length.cpp
+++ b/Libraries/LibWeb/CSS/Length.cpp
@@ -25,9 +25,7 @@ namespace Web::CSS {
 Length::FontMetrics::FontMetrics(CSSPixels font_size, Gfx::FontPixelMetrics const& pixel_metrics, CSSPixels line_height)
     : font_size(font_size)
     , x_height(pixel_metrics.x_height)
-    // FIXME: This is only approximately the cap height. The spec suggests measuring the "O" glyph:
-    //        https://www.w3.org/TR/css-values-4/#cap
-    , cap_height(pixel_metrics.ascent)
+    , cap_height(pixel_metrics.cap_height)
     , zero_advance(pixel_metrics.advance_of_ascii_zero)
     , line_height(line_height)
 {


Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
```diff
@@ -3087,6 +3091,48 @@ RefPtr<StyleValue const> Parser::parse_font_feature_settings_value(TokenStream<C
...
+    tokens.discard_whitespace();
+    if (auto none = parse_all_as_single_keyword_value(tokens, Keyword::None)) {
+        transaction.commit();
+        return none;
+    }
+
+    auto first_value = parse_css_value_for_property(PropertyID::FontSizeAdjust, tokens);
+    if (!first_value)
+        return nullptr;
+    if (first_value->is_number() || first_value->is_calculated() || first_value->to_keyword() == Keyword::FromFont) {
```
Calme1709, line 2960: If/when #6426 is merged this won't be enough since `<number>` may produce a `TreeCountingFunctionStyleValue` - perhaps it should instead be written more like:

```c++
if (auto maybe_number_value = parse_number_value(tokens)) {
	if (tokens.has_next_token())
		return nullptr;
	transaction.commit();
	return maybe_number_value ;
}

auto keyword_value = parse_keyword_value(tokens);

if (!keyword_value)
	return nullptr;

if (keyword_value->to_keyword() == Keyword::FromFont) {
	if (tokens.has_next_token())
		return nullptr;
	transaction.commit();
	return keyword_value;
}

auto font_metric_keyword = keyword_to_font_metric(keyword_value->to_keyword());

if (!font_metric_keyword.has_value())
	return nullptr;

...
```

Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
```diff
@@ -3087,6 +3091,48 @@ RefPtr<StyleValue const> Parser::parse_font_feature_settings_value(TokenStream<C
...
+    if (!second_value)
+        return nullptr;
+    if (!second_value->is_number() && !second_value->is_calculated() && second_value->to_keyword() != Keyword::FromFont)
+        return nullptr;
+
+    transaction.commit();
+    if (first_value->is_keyword() && first_value->as_keyword().keyword() == Keyword::ExHeight)
+        return second_value;
+
+    if (second_value->is_keyword() && second_value->as_keyword().keyword() == Keyword::ExHeight)
```
Calme1709: I don't think this is ever true since we check above that `second_value` is `NumberStyleValue`, `CalculatedStyleValue` or `from-font`.

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -3444,6 +3446,69 @@ NonnullRefPtr<StyleValue const> StyleComputer::compute_font_size(NonnullRefPtr<S
     VERIFY_NOT_REACHED();
 }
 
+NonnullRefPtr<StyleValue const> StyleComputer::compute_font_size_adjust(NonnullRefPtr<StyleValue const> const& specified_value, ComputationContext const& computation_context)
```
Calme1709, line 2525: We have already absolutized the `StyleValue` at the start of `compute_value_of_property` so we can:
 - Rename the parameter to `absolutized_value` in line with the other non-font computation functions.
 - Not pass a length resolution context to `resolve_number` in a couple places below
 - Update this function to just take a `FontMetrics` rather than a full `ComputationContext`

Edit: I see now that we also call this from `compute_font` where we don't absolutize the value.

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -3444,6 +3446,69 @@ NonnullRefPtr<StyleValue const> StyleComputer::compute_font_size(NonnullRefPtr<S
...
+{
+    if (specified_value->is_keyword() && specified_value->to_keyword() == Keyword::None)
```
Calme1709, line 2530: Missing a spec link

Libraries/LibWeb/CSS/ComputedProperties.cpp
```diff
@@ -2085,6 +2085,30 @@ WillChange ComputedProperties::will_change() const
...
+        return FontSizeAdjust { .font_metric = FontMetric::ExHeight, .number = value.as_number().number() };
+
+    VERIFY(value.is_value_list());
+    auto const& value_list = value.as_value_list().values();
+    VERIFY(value_list.size() == 2);
+    auto font_metric = keyword_to_font_metric(value_list[0]->to_keyword());
+    auto number = 0.0;
+    if (value_list[1]->is_number()) {
+        number = value_list[1]->as_number().number();
+    } else if (value.is_calculated() && value.as_calculated().resolves_to_number()) {
```
Calme1709: Have we not already converted `CalculatedStyleValue` to `NumberStyleValue` in `compute_font_size_adjust`?

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -3478,15 +3517,13 @@ NonnullRefPtr<StyleValue const> StyleComputer::compute_font_size_adjust(NonnullR
                 return KeywordStyleValue::create(Keyword::None);
             number = *resolved_metric;
         } else {
-            font_metric_keyword = keyword;
+            VERIFY_NOT_REACHED();
```
Calme1709, line 2560: Should these changes be ammended into the "LibWeb: Resolve font-size-adjust computed value" commit?

Libraries/LibWeb/CSS/ComputedProperties.h
```diff
@@ -265,6 +266,9 @@ class WEB_API ComputedProperties final : public JS::Cell {
         m_attempted_pseudo_class_matches = results;
     }
 
+    float adjusted_font_size() const { return m_adjusted_font_size; }
```
Calme1709: This doesn't respect animated values of `font-size-adjust`

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -2067,7 +2067,16 @@ void StyleComputer::compute_font(ComputedProperties& style, Optional<DOM::Abstra
 
     auto const& font_family = style.property(CSS::PropertyID::FontFamily);
 
-    auto font_list = compute_font_for_style_values(font_family, style.font_size(), style.font_slope(), style.font_weight(), style.font_width());
+    auto const& font_size_adjust_specified_value = style.property(PropertyID::FontSizeAdjust, ComputedProperties::WithAnimationsApplied::No);
+    style.set_property(
+        PropertyID::FontSizeAdjust,
+        compute_font_size_adjust(font_size_adjust_specified_value, font_computation_context),
```
Calme1709: It's not great that we are calling `compute_font_size_adjust` twice for non-animated values (once here and then again in `compute_property_values`), ideally we would only compute the values required for creating a length resolution context here and then do `compute_font_for_style_values` at `NodeWithStyle::apply_style()` time, avoiding this call - it would have the added benefit of respecting animated values.

`font_computation_context`'s length resolution context is based on the parent element where I presume we should be using one based on this element.

PR #6479, opened 2025-10-15
LibWeb: Add loading event to style element

### LINKED ISSUE #1290: Implement HTMLStyleElement `load` and `error` events
`<style>` elements need to fire a `load` or `error` event once they have finished loading any "critical subresources", according to [this algorithm](https://html.spec.whatwg.org/multipage/semantics.html#the-style-element:critical-subresources).

Discovered this when trying to run http://wpt.live/css/css-cascade/layer-import.html on my `@layer` branch - the test times out waiting for the `load` event, and I expect there are other CSS WPT tests that also time out for this reason.

lpas:
This PR implements only the same part as the link element implements as far as when we fire the load event. The waiting for critical resources is outside of the scope. 

fixes  #1290

### DIFF
--- a/Libraries/LibWeb/DOM/Element.cpp
+++ b/Libraries/LibWeb/DOM/Element.cpp
@@ -4333,4 +4333,35 @@ GC::Ptr<Element const> Element::element_to_inherit_style_from(Optional<CSS::Pseu
     return parent_or_shadow_host_element();
 }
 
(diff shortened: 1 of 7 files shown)


awesomekling:
Please add a test so we know this works as intended :)

lpas:
> Please add a test so we know this works as intended :)

done

PR #6484, opened 2025-10-15
LibWeb/IDB: Apply default cursor direction

stelar7:
We are supposed to use a default value for direction in the case we have a `IDBGetAllOptions` object, so lets actually do that.

Fixes remaining tests on https://wpt.live/IndexedDB/idbobjectstore_getAllKeys-options.any.html

Related spec bug: https://github.com/w3c/IndexedDB/pull/478

### DIFF
--- a/Libraries/LibWeb/IndexedDB/Internal/Algorithms.cpp
+++ b/Libraries/LibWeb/IndexedDB/Internal/Algorithms.cpp
@@ -2405,17 +2405,15 @@ WebIDL::ExceptionOr<GC::Ref<IDBRequest>> create_a_request_to_retrieve_multiple_i
     // 6. Let range be a key range.
     GC::Ptr<IDBKeyRange> range;
 
-    // 7. Let direction be a cursor direction.
-    Bindings::IDBCursorDirection direction;
+    // 7. Let direction be "next".
+    // FIXME: Spec bug: https://github.com/w3c/IndexedDB/pull/478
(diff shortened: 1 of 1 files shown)


Libraries/LibWeb/IndexedDB/Internal/Algorithms.cpp
```diff
@@ -2436,8 +2433,6 @@ WebIDL::ExceptionOr<GC::Ref<IDBRequest>> create_a_request_to_retrieve_multiple_i
             direction = Bindings::IDBCursorDirection::Prev;
         else if (direction_value == "prevunique")
             direction = Bindings::IDBCursorDirection::Prevunique;
-        else
-            return WebIDL::SimpleException { WebIDL::SimpleExceptionType::TypeError, "Invalid direction value"_string };
```
konradekk, line 2440: shall that be `VERIFY_NOT_REACHED` instead to avoid surprises? 🤔

stelar7: It cant be a verify, because this field is optional. An unset/invalid value is not a documented error state. Waiting for a response on the spec bug

gmta: Could you please include a bit of the PR description in your commit message, including the upstream issue? Could also link to it in a `// FIXME: ..` comment.

stelar7: Added 👍

PR #6487, opened 2025-10-15
RequestServer: Issue a network request for failed cached responses

trflynn89:
If transferring a cached response body fails for any reason, we will now issue a network request instead of failing the request outright.

The catch here is that we will have already transferred the response code and headers to the client, and potentially some of the body. So we attempt to only request the remaining data over the network using a range request. This feels a bit sketchy, but this is also how Chromium behaves.

However, the server may or may not support range requests. If they do, we can expect an HTTP 206 response with the bytes we need. If not, we will receive an HTTP 200 (assuming the request succeeded), along with the entire object's body. In this case, we also behave like Chromium, and internally drop number of bytes we had already transferred.

### DIFF
--- a/Services/RequestServer/ConnectionFromClient.cpp
+++ b/Services/RequestServer/ConnectionFromClient.cpp
@@ -117,17 +117,23 @@ struct ConnectionFromClient::ActiveRequest : public Weakable<ActiveRequest> {
     i32 request_id { 0 };
     WeakPtr<ConnectionFromClient> client;
     int writer_fd { 0 };
-    HTTP::HeaderMap headers;
-    bool got_all_headers { false };
     bool is_connect_only { false };
     size_t downloaded_so_far { 0 };
     URL::URL url;
     ByteString method;
     Optional<String> reason_phrase;
     ByteBuffer body;
+
     AllocatingMemoryStream send_buffer;
     NonnullRefPtr<Core::Notifier> write_notifier;
     bool done_fetching { false };
+
+    Optional<long> http_status_code;
+    HTTP::HeaderMap headers;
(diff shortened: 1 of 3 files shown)


alimpfard:
I haven't looked at the impl yet, so maybe you're doing this already, but should we check that the hash of the response we've stored matches the one we reconstructed from the new request? not sure how this _should_ work but I'd expect atomic behaviour.

trflynn89:
Right, there's currently a fixme to validate the hash of the data. I've extended that fixme to include this reconstructed request.

Services/RequestServer/ConnectionFromClient.cpp
```diff
@@ -614,6 +689,14 @@ void ConnectionFromClient::start_request(i32 request_id, ByteString method, URL:
                 request->curl_string_lists.append(curl_headers);
             }
 
+            if (resume_request.has_value()) {
+                auto range = ByteString::formatted("{}-", resume_request->start_offset);
+                set_option(CURLOPT_RANGE, range.characters());
+
+                request->got_all_headers = true; // Don't re-send the headers for resumed requests.
```
gmta, line 696: Hm, I guess this means we don't update the cache entry's headers based on the (potentially) partial response? But is that the correct behavior, i.e. what should we do if other headers actually change between the initial response and the partial response? An alternative would be to take the headers into account, but leave out `Content-Length` and `Content-Range` if present.

trflynn89: At this point the cache entry will be removed, so there's no cache to be updated. We could potentially optimize this to update the cached headers and the partial response, but for now we delete the cache if we can't read it.

PR #6493, opened 2025-10-16
LibWeb: Parse the `css-shapes` properties

Calme1709:
This PR implements parsing the `shape-outside`, `shape-image-threshold` and `shape-margin` properties.  We don't actually do anything with these properties yet.

Gains us ~1400 WPT tests.


Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
```diff
@@ -2282,6 +2286,60 @@ RefPtr<StyleValue const> Parser::parse_single_shadow_value(TokenStream<Component
...
+        if (auto maybe_keyword_value = parse_keyword_value(tokens)) {
+            if (shape_box_value)
+                return nullptr;
+            if (!keyword_to_shape_box(maybe_keyword_value->to_keyword()).has_value())
+                return nullptr;
+            shape_box_value = maybe_keyword_value;
+            break;
+        }
+
+        return nullptr;
```
tcl3, line 2320: This doesn't handle the case where both a `basic-shape` and a `shape-box` value are given (e.g: `shape-outside: circle() padding-box`).

WPT doesn't seem to test this case, but it is definitely supported in other engines.

Maybe something like this would work:
```c++
if (auto maybe_basic_shape_value = parse_basic_shape_value(tokens)) {
    if (basic_shape_value)
        return nullptr;
    basic_shape_value = maybe_basic_shape_value;
    continue;
}

if (auto maybe_keyword_value = parse_keyword_value(tokens)) {
    if (shape_box_value)
        return nullptr;
    if (!keyword_to_shape_box(maybe_keyword_value->to_keyword()).has_value())
        return nullptr;
    shape_box_value = maybe_keyword_value;
    continue;
}

break;
```

Libraries/LibWeb/CSS/Parser/ValueParsing.cpp
```diff
@@ -4455,6 +4455,21 @@ RefPtr<CalculationNode const> Parser::parse_a_calculation(Vector<ComponentValue>
     return simplify_a_calculation_tree(*calculation_tree, context, CalculationResolutionContext {});
 }
 
+// https://www.w3.org/TR/css-color-4/#typedef-opacity-opacity-value
```
tcl3: Editor's draft spec links are preferred: https://drafts.csswg.org/css-color-4/#typedef-opacity-opacity-value.

Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
Calme1709: Whoops those `break`s were definitely meant to be `continue`s - fixed.

I think it should still be `return nullptr` at the end rather than `break` since `break` would mean that `shape-outside: circle() foobar` would be valid and parsed as equivalent to just `circle()`

tcl3 (approved):

PR #6494, opened 2025-10-16
LibIPC: Change TransportSocket to read and write on I/O thread

kalenikaliaksandr:
Previously, TransportSocket sent queued messages from a separate thread but performed all reading on the main thread. With this change, both reading and writing are handled on the same I/O thread. This would allow us to read IPC messages even while the main thread is blocked and
process them on a different thread (e.g., a rendering thread).

This is a preparation work before allowing rendering thread to try process mouse events while the main thread is busy.

### DIFF
--- a/Libraries/LibIPC/TransportSocket.cpp
+++ b/Libraries/LibIPC/TransportSocket.cpp
@@ -18,15 +18,6 @@ void SendQueue::enqueue_message(Vector<u8>&& bytes, Vector<int>&& fds)
     Threading::MutexLocker locker(m_mutex);
     VERIFY(MUST(m_stream.write_some(bytes.span())) == bytes.size());
     m_fds.append(fds.data(), fds.size());
-    m_condition.signal();
-}
-
-SendQueue::Running SendQueue::block_until_message_enqueued()
-{
-    Threading::MutexLocker locker(m_mutex);
-    while (m_stream.is_eof() && m_fds.is_empty() && m_running)
-        m_condition.wait();
-    return m_running ? Running::Yes : Running::No;
 }
 
 SendQueue::BytesAndFds SendQueue::peek(size_t max_bytes)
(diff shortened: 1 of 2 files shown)


Libraries/LibWebView/EventLoop/EventLoopImplementationQt.cpp
```diff
@@ -13,30 +13,64 @@
...
 
 namespace WebView {
 
 struct ThreadData;
-static thread_local ThreadData* s_thread_data;
+static thread_local OwnPtr<ThreadData> s_this_thread_data;
+static HashMap<pthread_t, ThreadData*> s_thread_data;
+static thread_local pthread_t s_thread_id;
+static pthread_key_t s_thread_key;
+static pthread_rwlock_t s_thread_data_lock_impl;
```
trflynn89: Can we use `Threading::RWLock` and `Threading::RWLockLocker` here instead?

Libraries/LibWebView/EventLoop/EventLoopImplementationQt.cpp
```diff
@@ -13,30 +13,64 @@
...
+            pthread_key_create(&s_thread_key, [](void*) {
+                s_this_thread_data.clear();
+            });
         }
-        return *s_thread_data;
+        if (s_thread_id == 0)
+            s_thread_id = pthread_self();
+        if (!s_this_thread_data) {
+            s_this_thread_data = make<ThreadData>();
+            pthread_rwlock_wrlock(&*s_thread_data_lock);
```
trflynn89: (If we can't use RWLock) - de-referencing and then taking the address seems weird. Is this any different than just:
```c++
pthread_rwlock_wrlock(s_thread_data_lock);
```

Libraries/LibWebView/EventLoop/EventLoopImplementationQt.cpp
```diff
@@ -293,16 +327,31 @@ void EventLoopManagerQt::register_notifier(Core::Notifier& notifier)
...
 }
 
 void EventLoopManagerQt::unregister_notifier(Core::Notifier& notifier)
 {
-    ThreadData::the().notifiers.remove(&notifier);
+    auto* thread_data = ThreadData::for_thread(notifier.owner_thread());
+    if (!thread_data)
+        return;
+    Threading::MutexLocker locker(thread_data->mutex);
+    auto deleted_notifier = thread_data->notifiers.take(&notifier).release_value();
```
trflynn89, line 338: One thing that catches my eye is that this now assumes the notifier map had this notifier in it, and we'll crash if that wasn't the case. I don't recall the details but I (very) vaguely remember seeing that happen.

Libraries/LibWebView/EventLoop/EventLoopImplementationMacOS.mm
```diff
@@ -23,22 +23,41 @@
...
+static pthread_key_t s_thread_key;
+static pthread_rwlock_t s_thread_data_lock_impl;
+static pthread_rwlock_t* s_thread_data_lock = nullptr;
+
 struct ThreadData {
     static ThreadData& the()
     {
-        static thread_local ThreadData s_thread_data;
-        return s_thread_data;
+        if (!s_thread_data_lock) {
```
trflynn89: Same general comments here as from the Qt commit.

Libraries/LibWebView/EventLoop/EventLoopImplementationMacOS.mm
```diff
@@ -246,9 +265,30 @@ static void post_application_event()
...
+}
+
+static void socket_notifier(CFSocketRef socket, CFSocketCallBackType notification_type, CFDataRef, void const*, void* info)
 {
-    auto& notifier = ThreadData::the().notifier_by_fd(CFSocketGetNative(socket));
+    if (!info)
+        return;
+    auto& callback_context = *reinterpret_cast<SocketNotifierCallbackContext*>(info);
+    if (!callback_context.notifier)
+        return;
```
trflynn89: This and the ones above can *probably* be a static cast? We do the same in `handle_signal` below.

Libraries/LibWebView/EventLoop/EventLoopImplementationMacOS.mm
```diff
@@ -280,7 +320,10 @@ static void socket_notifier(CFSocketRef socket, CFSocketCallBackType notificatio
         break;
     }
 
-    CFSocketContext context { .version = 0, .info = nullptr, .retain = nullptr, .release = nullptr, .copyDescription = nullptr };
+    auto callback_context = adopt_ref(*new SocketNotifierCallbackContext);
+    callback_context->notifier = notifier.make_weak_ptr();
+    callback_context->ref();
```
trflynn89: Are we sure the calls to `ref` and `unref` are balanced? It *looks* like we'll `ref` once more than `unref`, but I haven't run this locally - does AppKit invoke `release_socket_notifier_callback_context` one extra time?

Libraries/LibIPC/TransportSocket.cpp
```diff
@@ -55,93 +46,165 @@ void SendQueue::discard(size_t bytes_count, size_t fds_count)
 void SendQueue::stop()
 {
     Threading::MutexLocker locker(m_mutex);
```
trflynn89: :confused:  so does this function need to exist? All it's doing now is taking a lock and immediately releasing it

Libraries/LibIPC/TransportSocket.cpp
```diff
@@ -55,93 +46,165 @@ void SendQueue::discard(size_t bytes_count, size_t fds_count)
...
+    m_incoming_cv.broadcast();
+    return 0;
 }
 
-TransportSocket::~TransportSocket()
+void TransportSocket::wake_io_thread()
 {
-    stop_send_thread();
+    Array<u8, 1> bytes = { 0 };
+    (void)Core::System::write(m_wakeup_io_thread_write_fd->value(), bytes);
```
trflynn89, line 139: I'm not quite follow the reason for the change from using condition variables to wake the IO thread to now using a pipe

trflynn89 (commented):
Given that this is a rather big change to a fundamental part of the system - what testing was done to really stress the new IPC transport?

Libraries/LibWebView/EventLoop/EventLoopImplementationQt.cpp
kalenikaliaksandr: It might have happened before in case unregistering happens on a different thread than registering, but now we should expect notifier to always be present in a HashMap.

Libraries/LibIPC/TransportSocket.cpp
kalenikaliaksandr: with this change IO thread is now blocked by poll() until either IPC socket has bytes to read or something is queued to write. waking up when new message is queued to send seems straightforward to implement by creating pipe and passing another fd to poll().

kalenikaliaksandr:
> Given that this is a rather big change to a fundamental part of the system - what testing was done to really stress the new IPC transport?

besides our internal tests suit I mostly checked that all websites I usually try work as before this change. successful run of WPT would be good sign that change is fine, but I am not gonna do this locally.

Libraries/LibIPC/TransportSocket.cpp
kalenikaliaksandr: Ooops, didn't notice this one doesn't do anything anymore.

Libraries/LibIPC/TransportSocket.cpp
R-Goc: Isn't using an eventfd cleaner for this?

kalenikaliaksandr: I didn't know about eventfd, but it does look better for this purpose. unfortunately is not supported on macOS, so I am going to stick to using pipes.

R-Goc: I missed that. Indeed a linux invention. FreeBSD having it fooled me.

Libraries/LibWebView/EventLoop/EventLoopImplementationQt.cpp
kalenikaliaksandr: sure, it was direct usage of pthread apis only because I borrowed it from `EventLoopImplementationUnix`. I updated all Qt, MacOS and Unix implementations to use `RWLock`.

Libraries/LibWebView/EventLoop/EventLoopImplementationMacOS.mm
kalenikaliaksandr: indeed, `static_cast` is fine here.

Libraries/LibIPC/TransportSocket.cpp
kalenikaliaksandr: remove this unused method.

Libraries/LibWebView/EventLoop/EventLoopImplementationMacOS.mm
kalenikaliaksandr: nope, this extra ref was a mistake. I got rid of it and now `SocketNotifierCallbackContext` destructor is actually invoked after unregistering.

Libraries/LibIPC/TransportSocket.cpp
```diff
@@ -52,96 +43,163 @@ void SendQueue::discard(size_t bytes_count, size_t fds_count)
...
 
-            if (transfer_data(remaining_bytes_to_send, fds) == TransferState::SocketClosed)
-                break;
+    {
+        auto fds = MUST(Core::System::pipe2(O_CLOEXEC | O_NONBLOCK));
+        m_notify_hook_read_fd = adopt_ref(*new AutoCloseFileDescriptor(fds[0]));
+        m_notify_hook_write_fd = adopt_ref(*new AutoCloseFileDescriptor(fds[1]));
+    }
+
+    m_io_thread_state = IOThreadState::Running;
```
R-Goc: the thread state is initialized to running already.

Libraries/LibIPC/TransportSocket.cpp
```diff
@@ -52,96 +43,163 @@ void SendQueue::discard(size_t bytes_count, size_t fds_count)
...
+        if (pollfds[1].revents & POLLIN) {
+            char buf[64];
+            MUST(Core::System::read(m_wakeup_io_thread_read_fd->value(), { buf, sizeof(buf) }));
+        }
+
+        if (pollfds[0].revents & POLLIN)
+            read_incoming_messages();
+
+        if (pollfds[0].revents & POLLHUP) {
+            m_io_thread_state = IOThreadState::Stopped;
```
R-Goc, line 111: nit: This can be a store with a release memory ordering. By default it is sequentially consistent. Same for the store to the thread state below.

Libraries/LibIPC/TransportSocket.cpp
```diff
@@ -52,96 +43,163 @@ void SendQueue::discard(size_t bytes_count, size_t fds_count)
...
 void TransportSocket::wait_until_readable()
 {
-    Threading::RWLockLocker<Threading::LockMode::Read> lock(m_socket_rw_lock);
-    auto maybe_did_become_readable = m_socket->can_read_without_blocking(-1);
-    if (maybe_did_become_readable.is_error()) {
-        dbgln("TransportSocket::wait_until_readable: {}", maybe_did_become_readable.error());
-        warnln("TransportSocket::wait_until_readable: {}", maybe_did_become_readable.error());
-        VERIFY_NOT_REACHED();
+    Threading::MutexLocker lock(m_incoming_mutex);
+    while (m_incoming_messages.is_empty() && m_io_thread_state == IOThreadState::Running) {
```
R-Goc, line 198: This could be a relaxed load no?

Libraries/LibIPC/TransportSocket.cpp
```diff
@@ -52,96 +43,163 @@ void SendQueue::discard(size_t bytes_count, size_t fds_count)
...
+{
+    Array<struct pollfd, 2> pollfds;
+    for (;;) {
+        auto want_to_write = [&] {
+            auto [bytes, fds] = m_send_queue->peek(1);
+            return !bytes.is_empty() || !fds.is_empty();
+        }();
+
+        auto state = m_io_thread_state.load();
+        if (state == IOThreadState::Stopped)
```
R-Goc, line 80: none of this is atomic but it should work since you can never set the thread back to running currently. The only condition this causes is if something stops the io_thread from thread A, then thread B changes the state to SendPendingMessagesAndStop and this falls within the two instruction windows the thread will stop even though the state is SendPendingMessagesAndStop. If we could set the state back to running this would be more serious.

Libraries/LibIPC/TransportSocket.cpp
```diff
@@ -52,96 +43,163 @@ void SendQueue::discard(size_t bytes_count, size_t fds_count)
...
+            break;
+        if (state == IOThreadState::SendPendingMessagesAndStop && !want_to_write) {
+            m_io_thread_state = IOThreadState::Stopped;
+            break;
         }
+        VERIFY(state == IOThreadState::Running || state == IOThreadState::SendPendingMessagesAndStop);
```
R-Goc: Not really sure what is the point of this. In the next instruction a thread might as well set the state to stopped. Also these aren't atomic comparisons either.

Zaggy1024:
For functions in `EventLoopImplementationUnix` that call `ThreadData::for_thread()`, I suspect that we should probably hold a read lock for as long as we're using the `ThreadData*`, so that it cannot be destroyed between getting it and using it. Alternatively, ThreadData could be atomically ref-counted, but I want to say that the read lock should be enough.

Zaggy1024:
Also, you may be able to delete the `pthread_key_t`, as that appears to only be intended to delete the `thread_local OwnPtr<ThreadData> s_this_thread_data`, yet the key has no associated data to cause it to call the destructor callback. With it being a thread-local, it should be cleaned up automatically regardless.

kalenikaliaksandr:
> For functions in `EventLoopImplementationUnix` that call `ThreadData::for_thread()`, I suspect that we should probably hold a read lock for as long as we're using the `ThreadData*`, so that it cannot be destroyed between getting it and using it. Alternatively, ThreadData could be atomically ref-counted, but I want to say that the read lock should be enough.

right, I made callers of `ThreadData::for_thread()` hold a read lock for as long as returned pointer is alive.

> Also, you may be able to delete the `pthread_key_t`, as that appears to only be intended to delete the `thread_local OwnPtr<ThreadData> s_this_thread_data`, yet the key has no associated data to cause it to call the destructor callback. With it being a thread-local, it should be cleaned up automatically regardless.

indeed, I removed it.

PR #6495, opened 2025-10-16
LibGfx: Check for Vulkan device extension availability on Linux

rcorsi:
Currently on many Linux distributions the sharable Vulkan image device extension is not available. Due to this, the Vulkan context fails to be created and ladybird falls back to CPU painter.

Add physical device extension availablity check to exclude the usage of sharable Vulkan image if extension is not available.


rcorsi:
PR was mistakenly created. Will be closing it.

rcorsi:
PR opened by mistake. Will be making a new PR when this is completely ready.

Issue #6470, opened 2025-10-14
Layout issue on MacRumors Forums Post


andrewrothman:
### Summary

My first issue report on Ladybird - please let me know if I should report issues differently.

When I open a post on the MacRumors Forum, the pager elements (CSS `.pageNav` and children) on the top and bottom of the page are shown vertically instead of horizontally.

### Operating system

macOS

### Steps to reproduce

1. Visit https://forums.macrumors.com/threads/ios-26-1-beta-3-bug-fixes-changes-and-improvements.2469035/

### Expected behavior

The pager elements (CSS `.pageNav` and children) on the top and bottom of the page should be shown horizontally.

### Actual behavior

The pager elements (CSS `.pageNav` and children) on the top and bottom of the page are shown vertically.

### URL for a reduced test case

N/A

### HTML/SVG/etc. source for a reduced test case

```html
<!DOCTYPE html>
<html>
    <head>
        <style>
            div { display: flex; }
            ul { display: inline-table; }
            li { display: inline-block; }
        </style>
    </head>
    <body>
        <div>
            HELLO
            WORLD
        </div>
        <div>
            <ul>
                FOO
                BAR
            </ul>
        </div>
        <div>
            <ul>
                <li>BAZ</li>
                <li>BUQ</li>
            </ul>
        </div>
    </body>
</html>
```

### Log output and (if possible) backtrace

```shell
1104030.945 WebContent(81864): ResourceLoader: Failed load of: "http://localhost:8000/favicon.ico", Error: Load failed (status: 404 Not Found), Duration: 3ms
```

### Screenshots or screen recordings

# Forum Post in Brave
<img width="1225" height="648" alt="Image" src="https://github.com/user-attachments/assets/788fc209-939e-47f8-a7f4-73a0a6c678d9" />

# Forum Post in Ladybird
<img width="1380" height="767" alt="Image" src="https://github.com/user-attachments/assets/355dd8d4-96fc-41f8-96f8-eeea02ec94e2" />

# Reduced Test Case in Brave
<img width="200" height="200" alt="Image" src="https://github.com/user-attachments/assets/5adffd3c-424e-486c-98d5-572e23ab8e2b" />

# Reduced Test Case in Ladybird
<img width="200" height="200" alt="Image" src="https://github.com/user-attachments/assets/683c66dd-39a5-4819-a57a-bfd9a2dc7c0b" />

### Contribute a patch?

- [ ] I’ll contribute a patch for this myself.

[...]

PR #6499, opened 2025-10-17
LibWeb: Allow in-flow flex/grid items to avoid anonymous wrappers

allanhanan:
This extends the existing fix for absolutely positioned elements (commit 7b4c76788b) to also apply to in-flow elements in flex/grid containers.

Previously, blockified inline-table elements inside flex containers were wrapped in anonymous blocks, preventing them from becoming proper flex items and breaking horizontal layout.

This change allows inline-table and similar elements to become direct flex items as per CSS Flexbox specifications.

Tested locally with inline-table in flex containers; layout now works horizontally as expected.

Relates to #6470 (partial fix).

### DIFF
--- a/Libraries/LibWeb/Layout/TreeBuilder.cpp
+++ b/Libraries/LibWeb/Layout/TreeBuilder.cpp
@@ -25,6 +25,7 @@
 #include <LibWeb/Dump.h>
 #include <LibWeb/HTML/HTMLInputElement.h>
 #include <LibWeb/HTML/HTMLSlotElement.h>
+#include <LibWeb/Layout/BreakNode.h>
 #include <LibWeb/Layout/FieldSetBox.h>
 #include <LibWeb/Layout/ImageBox.h>
 #include <LibWeb/Layout/InlineNode.h>
@@ -68,7 +69,7 @@ static bool has_in_flow_block_children(Layout::Node const& layout_node)
 // The insertion_parent_for_*() functions maintain the invariant that the in-flow children of
 // block-level boxes must be either all block-level or all inline-level.
 
-static Layout::Node& insertion_parent_for_inline_node(Layout::NodeWithStyle& layout_parent)
+static Layout::Node& insertion_parent_for_inline_node(Layout::NodeWithStyle& layout_parent, Layout::Node& layout_node)
 {
     auto last_child_creating_anonymous_wrapper_if_needed = [](auto& layout_parent) -> Layout::Node& {
         if (!layout_parent.last_child()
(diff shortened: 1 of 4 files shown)


allanhanan:
Ok, I've amended the commit message to wrap lines per policy and rebased onto latest master for a linear history (with no merges)

allanhanan:
> please add a new layout test for this change

I've added a layout test at `Tests/LibWeb/Layout/input/flex/inline-table-no-anonymous-wrapper.html` verifying that `display: inline-table` elements in flex containers are now correctly treated as direct flex items (TableWrapper marked as flex-item) without anonymous block wrappers. Test passes locally. Ready for re-review!

Libraries/LibWeb/Layout/TreeBuilder.cpp
```diff
@@ -125,6 +125,11 @@ static Layout::Node& insertion_parent_for_block_node(Layout::NodeWithStyle& layo
         return layout_parent;
     }
 
+    if (layout_parent.display().is_flex_inside() || layout_parent.display().is_grid_inside()) {
+        // Flex and grid containers can have mixed inline and block children.
+        return layout_parent;
+    }
```
gmta, line 144: If you move this above the `if (layout_node.is_out_of_flow() && ...` block above, you can remove the `!layout_parent.display().is_flex_inside() && !layout_parent.display().is_grid_inside()` part of that condition.

allanhanan: Good catch! I will move the flex/grid check earlier to simplify the out-of-flow condition and amend the changes

allanhanan: Done! I moved the flex/grid check before the out-of-flow condition and removed the redundant checks. Thanks for catching that! Its much cleaner now.

gmta:
@allanhanan Tests are failing - you need to run the tests locally (`Meta/ladybird.py run test-web`) so you can see what changes.

I've looked at some of them, and it's mostly layout tests where the anonymous containers disappear which is expected. Those tests you should be able to rebaseline with the `-f` and `--rebaseline` arguments to `test-web`. [More on that](https://github.com/LadybirdBrowser/ladybird/blob/master/Documentation/Testing.md#writing-tests).

allanhanan:
> @allanhanan Tests are failing - you need to run the tests locally (`Meta/ladybird.py run test-web`) so you can see what changes.
> 
> I've looked at some of them, and it's mostly layout tests where the anonymous containers disappear which is expected. Those tests you should be able to rebaseline with the `-f` and `--rebaseline` arguments to `test-web`. [More on that](https://github.com/LadybirdBrowser/ladybird/blob/master/Documentation/Testing.md#writing-tests).

Done! Ran the full test suite locally, had to do a few fixes and rebaselined the affected tests. All are tests passing now:
```
Done!
Pass: 5011, Fail: 0, Skipped: 278, Timeout: 0, Crashed: 0
```

allanhanan:
Ran clang-format locally and pushed the changes. It should pass all CI runs now.

allanhanan:
Resolved the merge conflict, verified tests pass and pushed changes, awaiting review

allanhanan:
it seems that the macOS still fails due to CI env issues, other platforms pass tho. Let me know is there's anything to adjust from my side.

gmta:
@allanhanan 

> it seems that the macOS still fails due to CI env issues

That's not the case, seems like a flaky test is preventing it from running to completion successfully.

gmta:
@allanhanan please rebase your branch on the latest `master` so we can run all required CI checks.

allanhanan:
I've Rebased the commit to the latest master branch and force pushed it with no merge conflicts. It must be ready for CI now

allanhanan:
I think the `Layout/input/flex/inline-table-no-anonymous-wrapper.html` test just needs a rebaseline. CI shows it as the only LibWeb failure and everything else passes. I tried to rebaseline the test, but on my system (Fedora 43) test-web segfaults in `libskia.so` (from vcpkg) before tests can run. Is there any other way to get it rebaselined? I tried looking at the CI artifacts but it doesn’t upload test-dumps for this case.

Issue #6392, opened 2025-10-04
Weird drawing artifacts on move/resize when using CPU painting


manuel-za:
- a098b5fcc4; `gcc`; Ubuntu-24.04 guest (KVM; `llvmpipe`) on Ubuntu 22.04 host; beefy laptop; Nvidia GEForce RTX 4060 + nouveau
- Start Ladybird with `--force-cpu-painting`
- Resizing + moving a Ladybird window creates weird artifacts.
- Introduced in c18314b942. The changes are pretty involved cc: @kalenikaliaksandr suggestions?

<img width="418" height="359" alt="Image" src="https://github.com/user-attachments/assets/760b1b30-50e8-4880-90d5-fa27c8cc708f" />
<br>
<br>

Solving this would hide most of the effects in #6031, since in all examples Ladybird falls back on CPU painting after Vulkan logical device creation failed.

The error `Vulkan context creation failed: Logical device creation failed` (e.g. because of unsupported Vukan extensions) should be dealt with separately.

ElyasSindi2020:
im just gonna put this out there, (i also have a 4060 laptop) when i use the nouveau driver or llvmpipe something graphics/performance related always goes wrong. wether its weird artifacts or bad performance. but with the official nvidia driver it fixes everything

manuel-za:
> 4060 [...] nouveau [...] llvmpipe [...] graphics/performance [...]  goes wrong [...] official nvidia driver it fixes everything

I think unrelated. Background: new Dell G15 + Ubuntu 22.04 + official nvidia drivers (for pytorch). Worked fine for 6 mo or so then intermittent boot issues (guess: new kernel), worsening. Got to a point of system don't boot. Upgraded everything no joy. Changed to nouveau. All works well. (Using nvidia server drivers for ML; no games). KVM and VirtualBox VMs work well. Firefox + Chromium in KVM guest no prob.

👈 The issue here is CPU painting (_not_ using `Vulkan`) in Ladybird (`--force-cpu-painting`). Works fine before c18314b. Paints wonky after.
- Exposed by adding support for `VK_EXT_IMAGE_DRM_FORMAT_MODIFIER_EXTENSION_NAME` + fallback to CPU painting on unsupported architectures (see discussion in #6031).
- I believe it's a problem everywhere. Please disagree if you think different.

rcorsi:
[...]

Issue #6322, opened 2025-09-27
Resizing Ladybird window creates weird artifacts


manuel-za:
- `gcc`; Ubuntu-24.04 guest (KVM; `llvmpipe`) on Ubuntu 22.04 host; beefy laptop; Nvidia GEForce RTX 4060 + nouveau
- `master` at 11a6b6f38f 
- Resizing + moving a Ladybird window creates weird artifacts.

<img width="418" height="359" alt="Image" src="https://github.com/user-attachments/assets/760b1b30-50e8-4880-90d5-fa27c8cc708f" />

<br>
<br>

<details><summary>Click for detailed walk-through</summary>

1. Open Ladybird. Go to https://ladybird.org (or any other web site). Works.
 
    - Window can be minimized, moved + continues to paint correctly. Nice! 😎 

<img width="418" height="359" alt="Image" src="https://github.com/user-attachments/assets/7f0157fd-0a3d-46af-b1b6-7f8d085df6e0" />

<br>
<br>

2. Resize  window (increase size). Notice unpainted bottom + right edges. 

<img width="418" height="359" alt="Image" src="https://github.com/user-attachments/assets/36d93335-d50c-4eaf-8a81-ccc056b371b7" />

<br>
<br>

3.  Move window. Notice weird toolbar artifacts.

<img width="418" height="359" alt="Image" src="https://github.com/user-attachments/assets/4ca9962a-1b42-4943-a0e2-71920dfb4d72" />

<br>
<br>

4. Refresh or reload. Everything is back to normal. 

   - Window can be minimized, moved + continues to paint correctly.

<img width="418" height="359" alt="Image" src="https://github.com/user-attachments/assets/94f2edab-4539-43df-a30a-f99610379543" />

<br>
<br>

5. Resize window (reduce size). Result looks ok.

<img width="418" height="359" alt="Image" src="https://github.com/user-attachments/assets/fbfd7109-a8f7-42a0-9b70-f1b8f331b51e" />

<br>
<br>

6. ...but if you move the window, the background bleeds through.

<img width="418" height="359" alt="Image" src="https://github.com/user-attachments/assets/760b1b30-50e8-4880-90d5-fa27c8cc708f" />

7. Refresh or reload. Everything is back to normal. 

     - Window can be minimized, moved + continues to paint correctly. 

<img width="418" height="359" alt="Image" src="https://github.com/user-attachments/assets/40110e8c-a487-40de-bd49-ffe4d03c55f5" />

[...]

Issue #6031, opened 2025-08-30
Ladybird has trouble drawing window content at all


Psychpsyo:
### Summary

This happens when I start ladybird on my machine:

https://github.com/user-attachments/assets/87ee5043-1280-407f-b724-986e4bf2b3e2

It happens consistently and I can fix it by clicking around in the window some. Also, resizing the window tends to break it again.

### Operating system

Linux

### Steps to reproduce

1. Start Ladybird

### Expected behavior

It should not be doing that.

### Actual behavior

It is doing that.

### URL for a reduced test case

---

### HTML/SVG/etc. source for a reduced test case

```html
---
```

### Log output and (if possible) backtrace

```shell
69409.199 WebContent(86557): Vulkan context creation failed: Logical device creation failed
69409.511 WebContent(86563): Vulkan context creation failed: Logical device creation failed
```

### Screenshots or screen recordings

See above.

### Build flags or config settings

None

### Contribute a patch?

- [ ] I’ll contribute a patch for this myself.

rcorsi:
I have seen the same.

It is most likely related to what we see on WSL2 and being discussed in #5963

Are you running Ubuntu 24.04 in a VM?  Can you give us additional details about your environment.

Errors about `Vulkan context creation failed: Logical device creation failed` are due to VulkanContext failing to be created and the system is falling back to software based GPU (same as --force-cpu-painting).  If you are not using WSL2 you should be able to get the VulkanContext working.

Psychpsyo:
> Are you running Ubuntu 24.04 in a VM? Can you give us additional details about your environment.

This is Ubuntu 24.04.1 LTS x86_64 running inside VirtualBox, on Windows 10. (no WSL involved)

rcorsi:
Ok, I have Ubuntu 25.04, VirtualBox, Windows 11 host, 12 cores assigned to the vm and I see exactly the same thing as you.

If you want to run some additional tests, you could always vary the number of cores (1, 2, 4, 8, 12, 16, 20, 24, 28, 32 if you have some to spare?) assigned to the VM and see if the problem goes away or gets worse or different.  Or build in Debug and see if anything changes.  Just to gather more information, if you like. 

It does appear to be some race condition in the code when the software based GPU is used.

[...]

Issue #5963, opened 2025-08-22
Initial tab has transparent background


HelloWorld-n:
<img width="1892" height="907" alt="Image" src="https://github.com/user-attachments/assets/1c561a5a-08e7-4138-a0f4-385231569f71" />

Does not occur on any subsequent tags.
Not even in with URL `about:newtab`.

rcorsi:
Could you tell us a little more about your environment? Desktop?  Anything special about it?  Version of Linux or Ubuntu? Has Ladybird worked well for you for some time and now today this happened.

Message about "..Wayland does not support..." is something I've never seen before.

rcorsi:
Is this WSL 2 with WSLg in Windows?

rcorsi:
I just built Ladybird on WSL2 with Ubuntu 24.04.1 LTS, works pretty good. I didn't encounter the issue you reported.

I get these messages when starting
```
bobo@DESKTOP-7615CN0:~/projects/ladybird$ ./Build/release/bin/Ladybird
Failed to create wl_display (No such file or directory)
qt.qpa.plugin: Could not load the Qt platform plugin "wayland" in "" even though it was found.
4170.505 WebContent(114148): Vulkan context creation failed: Logical device creation failed
4170.516 WebContent(114152): Vulkan context creation failed: Logical device creation failed
```

and if I avoid the GPU using `--force-cpu-painting` 

```
bobo@DESKTOP-7615CN0:~/projects/ladybird$ ./Build/release/bin/Ladybird --force-cpu-painting
Failed to create wl_display (No such file or directory)
qt.qpa.plugin: Could not load the Qt platform plugin "wayland" in "" even though it was found.
```

and if I avoid using wayland using the QT_QPA_PLATFORM environment variable to use xcb instead.

```
bobo@DESKTOP-7615CN0:~/projects/ladybird$ QT_QPA_PLATFORM=xcb ./Build/release/bin/Ladybird --force-cpu-painting
```

Are you using some older Ubuntu version on WSL2?

I put the the error message: `qt.qpa.wayland: Wayland does not support QWindow::requestActivate()` into google and there are some discussions about some older Qt versions causing this error message. Might be related to your issue.

wpenistone:
> Is this WSL 2 with WSLg in Windows?

Most likely, I am on WSL 2 and I have the same issue.
started happening after I set the mesa d3d12 device to my dedicated GPU

HelloWorld-n:
[...]

PR #6503, opened 2025-10-17
UI: Fix transparent or artifacting tab painting

rcorsi:
When ladybird falls back to CPU backend painter, same as using --force-cpu-painting, resizing or moving a window causes all types of artifacts.  On WSL2 (with Ubuntu 24.04 for example) the window will be transparent on startup.

I can't completely say why I tried to change from Format_RGB32 to Format_ARGB32, but I noticed it was used in other places and gave it a try.  It seems to work, but is it the appropriate fix, I don't know.

fixes: #5963 
fixes: #6031
fixes: #6322
fixes: #6392

### DIFF
--- a/UI/Qt/WebContentView.cpp
+++ b/UI/Qt/WebContentView.cpp
@@ -504,7 +504,7 @@ void WebContentView::paintEvent(QPaintEvent*)
     }
 
     if (bitmap) {
-        QImage q_image(bitmap->scanline_u8(0), bitmap->width(), bitmap->height(), bitmap->pitch(), QImage::Format_RGB32);
+        QImage q_image(bitmap->scanline_u8(0), bitmap->width(), bitmap->height(), bitmap->pitch(), QImage::Format_ARGB32);
         painter.drawImage(QPoint(0, 0), q_image, QRect(0, 0, bitmap_size.width(), bitmap_size.height()));
 
         if (bitmap_size.width() < width()) {


rcorsi:
This fix is an improvement, as the strange window artifacts on resize and the transparent window (on WSL2) on startup will be gone, but there still remains an additional issue when the window is resized the output to the active tab will not be working for some time, probably until an event is triggered. As an example, with `about:processes`, if the window is resized the output will be blanked out and after 3 seconds the about process page will reappear.

rcorsi:
On macOS we have `UI/AppKit/Interface/LadybirdWebView.mm:894` which has:

```
VERIFY(bitmap.format() == Gfx::BitmapFormat::BGRA8888);
```

Not sure if it needs an adjustment too?

rcorsi:
I suspect `Gfx::BitmapFormat::BGRA8888` and `QImage::Format_ARGB32` are the same format.

`ARGB` is the reverse of `BGRA`, I mean the bits are little endian vs big endian.  And 8888 bits is same as 32 bits.

rcorsi:
Just checked this issue on my Intel based macOS system and it is not affected by the artifacts or transparent window.

It is affected by the other problem when the window is resized it takes some time to update and it has the same issue with `about:processes` that disappears for about 3 seconds when window is resized.  I need to create a followup about this.

rcorsi:
Followup #6505 created

rcorsi:
As a double check I added these lines below the `if (bitmap) {` at the location `UI/Qt/WebContentView.cpp:506`

```
        dbgln("Bitmap has alpha channel: {}", bitmap->has_alpha_channel());
        VERIFY(bitmap->format() == Gfx::BitmapFormat::BGRA8888);
```

Outputs a `true` for the alpha channel check and the VERIFY is happy.

rcorsi:
So the bitmaps are Gfx::BitmapFormat::BGRA8888 with an alpha channel and premultiplied, when they are created in `Libraries/LibWeb/Painting/BackingStoreManager.cpp:117`

```
    auto front_bitmap = Gfx::Bitmap::create_shareable(Gfx::BitmapFormat::BGRA8888, Gfx::AlphaType::Premultiplied, size).release_value();
    auto back_bitmap = Gfx::Bitmap::create_shareable(Gfx::BitmapFormat::BGRA8888, Gfx::AlphaType::Premultiplied, size).release_value();
```

rcorsi:
@kalenikaliaksandr I think you're familiar with this bitmap code causing problems when resizing or moving the ladybird window, but only if in CPU backend painting.  Are you familiar with the Qt side?  Hope you can review and merge if appropriate. Thank you.

rcorsi:
Ok, the fix in #6527 also fixes this issue.  So maybe #6503 is not needed and #6527 is the correct solution.

rcorsi:
Proper fix is to use is #6527.  In the worse case #6503 is not needed, at best is an optional fix.

gmta:
Closing this after merging #6527.

PR #6504, opened 2025-10-17
LibWeb/IDB: Dont run wait callbacks as a database task

stelar7:
The spec doesnt say to `queue a database task` for these.
Not sure if it was done for a reason, but if so, we should create spec issues for them

cc: @Lubrsi


tcl3:
I ran WPT on this change and it apparently fixes 27 crashes and 51 timeouts. I'm not hugely knowledgeable in this area though, so I'll wait for Luke or someone else to OK it.

Lubrsi:
LGTM, I was just doing `queue_a_database_task` as a blanket for asynchronous operations without fully considering if it was valid or not. If this fixes issues, then go for it :^)

PR #6506, opened 2025-10-18
LibWeb: peek after extracting next chunk in InlineLevelIterator

### LINKED ISSUE #6384: Text next to <a> or <span> element not take padding/background image size into account for positioning
### Summary

This is what happens in Chrome for the code of the reduction below:

Image

and this is on Ladybird:
Image

it looks like the padding is not taken into account for some reason (or maybe the image width), when positioning the side text (at least that's how I interpret it). Originally taken from https://wiki.openstreetmap.org/wiki/Lane_assist/Examples/Motorway_exit

This does not work for `a` and `span` elements, so maybe something related to the inline layout (dunno really)

### Operating system

Linux

### Steps to reproduce

1. Try the minimal repro
2. See the difference

### Expected behavior

As in Chrome/Firefox

### Actual behavior

Overlap of image on side text

N/A

```html
<style>
  .external {
    background-image: url("https://wiki.openstreetmap.org/w/skins/Vector/resources/skins.vector.styles.legacy/images/link-external-small-ltr-progressive.svg?fb64d");
    background-position: center right;
    background-repeat: no-repeat;
    padding-right: 1em;
  }
</style>

<span class="external">Location</span> on the map.
```

```shell
None
```

See above

LB b1a7782c593d97d02a960ac5f054e5387a58a77d

Linux, Fedora

### Contribute a patch?

amhndu:
Peek is used to determine if the current chunk is the last and if so, add trailing box metrics.
For this to work correctly, it must be done only after calling next(), otherwise peek gives us the current chunk.

minimal repro:
```html
<!DOCTYPE html>
<style>
    #inline-margin {
        background-color: red;
        margin-right: 25px;
    }
</style><span id="inline-margin">foo</span>bu
```
| Before patch | After patch | Firefox |
| ------ | ----- | ------- |
|  repro | repro | repro |

Affected test cases:
| Test case | Before patch | After patch | Firefox |
| -------   | ------ | ----- | ------- |
| inline-block-leading-and-trailing-metrics |  |  |  |
| leading-margin-on-inline-content-that-starts-with-collapsible-whitespace |  |  |  |
| inline-nested-with-padding |  |  |  |

Fixes #6384

### DIFF
--- a/Libraries/LibWeb/Layout/InlineLevelIterator.cpp
+++ b/Libraries/LibWeb/Layout/InlineLevelIterator.cpp
@@ -497,10 +497,11 @@ Optional<InlineLevelIterator::Item> InlineLevelIterator::next_without_lookahead(
         else
             m_text_node_context->is_first_chunk = false;
 
+        auto chunk_opt = m_text_node_context->chunk_iterator.next();
+
         if (!m_text_node_context->chunk_iterator.peek(0).has_value())
             m_text_node_context->is_last_chunk = true;
 
-        auto chunk_opt = m_text_node_context->chunk_iterator.next();
         auto is_empty_editable = false;
         if (!chunk_opt.has_value()) {
             auto const is_only_chunk = m_text_node_context->is_first_chunk && m_text_node_context->is_last_chunk;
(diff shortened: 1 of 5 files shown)


paaspaas00:
Could it also help with this https://github.com/LadybirdBrowser/ladybird/issues/6384? That or a similar fix

PR #6510, opened 2025-10-19
LibJS: Use call-threading for the interpreter loop

Hendiadyoin1:
This has a few benefits:
* Better backtraces, as now we have a function per opcode
* Flatten on GCC should work now

### DIFF
--- a/Meta/find_compiler.py
+++ b/Meta/find_compiler.py
@@ -19,7 +19,7 @@ from Meta.host_platform import Platform
 from Meta.utils import run_command
 
 CLANG_MINIMUM_VERSION = 19
-GCC_MINIMUM_VERSION = 14
+GCC_MINIMUM_VERSION = 15
 XCODE_MINIMUM_VERSION = ("16.3", 17000013)
 
 COMPILER_VERSION_REGEX = re.compile(r"(\d+)(\.\d+)*")
@@ -116,7 +116,7 @@ def pick_host_compiler(platform: Platform, cc: str, cxx: str) -> tuple[str, str]
 
         gcc_candidates = [
             "gcc",
-            "gcc-14",
+            "gcc-15",
         ]
 
         if platform.host_system == HostSystem.BSD:
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -40,8 +40,8 @@ runs:
 
         sudo apt-get update -y
         sudo apt-get install -y autoconf autoconf-archive automake build-essential ccache cmake curl fonts-liberation2 \
-            gcc-14 g++-14 libcurl4-openssl-dev libdrm-dev libegl1-mesa-dev libgl1-mesa-dev libpulse-dev libssl-dev \
-            libstdc++-14-dev lld-20 llvm-20 nasm ninja-build qt6-base-dev qt6-tools-dev-tools tar unzip zip
+            gcc-15 g++-15 libcurl4-openssl-dev libdrm-dev libegl1-mesa-dev libgl1-mesa-dev libpulse-dev libssl-dev \
+            libstdc++-15-dev lld-20 llvm-20 nasm ninja-build qt6-base-dev qt6-tools-dev-tools tar unzip zip
 
         if ${{ inputs.toolchain == 'Clang' }} ; then
           sudo apt-get install -y clang-20 clang++-20 clang-tools-20
@@ -50,8 +50,8 @@ runs:
         fi
 
         sudo update-alternatives --install /usr/bin/llvm-symbolizer llvm-symbolizer /usr/bin/llvm-symbolizer-20 100
-        sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-14 100
-        sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-14 100
+        sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-15 100
+        sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-15 100
 
     - name: 'Select latest Xcode'
       if: ${{ inputs.os == 'macOS' || inputs.os == 'Android' }}
--- a/.github/workflows/lagom-template.yml
+++ b/.github/workflows/lagom-template.yml
@@ -75,8 +75,8 @@ jobs:
               echo "host_cc=clang-20" >> "$GITHUB_OUTPUT"
               echo "host_cxx=clang++-20" >> "$GITHUB_OUTPUT"
             elif ${{ inputs.toolchain == 'GNU' }} ; then
-              echo "host_cc=gcc-14" >> "$GITHUB_OUTPUT"
-              echo "host_cxx=g++-14" >> "$GITHUB_OUTPUT"
+              echo "host_cc=gcc-15" >> "$GITHUB_OUTPUT"
+              echo "host_cxx=g++-15" >> "$GITHUB_OUTPUT"
             fi
           elif ${{ inputs.os_name == 'macOS' }} ; then
             echo "host_cc=$(xcrun --find clang)" >> "$GITHUB_OUTPUT"
--- a/Documentation/BuildInstructionsLadybird.md
+++ b/Documentation/BuildInstructionsLadybird.md
@@ -4,7 +4,7 @@
 
 Qt6 development packages, nasm, additional build tools, and a C++23 capable compiler are required.
 
-We currently use gcc-14 and clang-20 in our CI pipeline. If these versions are not available on your system, see
+We currently use gcc-15 and clang-20 in our CI pipeline. If these versions are not available on your system, see
 [`Meta/find_compiler.py`](../Meta/find_compiler.py) for the minimum compatible version.
 
 CMake 3.25 or newer must be available in $PATH.
@@ -59,7 +59,7 @@ sudo apt update -y && sudo apt install clang-20 clangd-20 clang-tools-20 clang-f
 
 ```bash
 sudo add-apt-repository ppa:ubuntu-toolchain-r/test
-sudo apt update && sudo apt install g++-14 libstdc++-14-dev
+sudo apt update && sudo apt install g++-15 libstdc++-15-dev
 ```
 
 #### Audio support:
--- a/Documentation/Troubleshooting.md
+++ b/Documentation/Troubleshooting.md
@@ -13,7 +13,7 @@ version of CMake, you can download a binary release from the [CMake website](htt
 
 Ensure your gcc version is supported by the [build system](BuildInstructionsLadybird.md#build-prerequisites) with
 `gcc --version`. Otherwise, install it. If your gcc binary is not called `gcc` you have to specify the names of your
-C and C++ compiler when you run cmake, e.g. `cmake ../.. -GNinja -DCMAKE_C_COMPILER=gcc-13 -DCMAKE_CXX_COMPILER=g++-13`.
+C and C++ compiler when you run cmake, e.g. `cmake ../.. -GNinja -DCMAKE_C_COMPILER=15 -DCMAKE_CXX_COMPILER=g++-15`.
 
 ### Legacy renegotiation is disabled
 


Hendiadyoin1:
X86/Linux results seem to be mixed on my end, but that is known to act odd in some cases
@alimpfard tested on a mac, where it seemed to gain some points on octane

Libraries/LibJS/Bytecode/Instructions.h
```diff
@@ -0,0 +1,218 @@
...
+    } while (0)
+
+#define HANDLE_EXCEPTION(result, interpreter)                                                                                                    \
+    do {                                                                                                                                         \
+        if (result.is_error()) [[unlikely]] {                                                                                                    \
+            if (interpreter.handle_exception(program_counter, result.error_value()) == Interpreter::HandleExceptionResponse::ExitFromExecutable) \
+                return;                                                                                                                          \
+            DISPATCH_NEXT(interpreter);                                                                                                          \
+        }                                                                                                                                        \
+    } while (0)
```
alimpfard: No one else needs these macros, please put them in the .cpp file.

Libraries/LibJS/Bytecode/Instructions.h
```diff
@@ -0,0 +1,218 @@
...
+template<>
+void handle<Op::ContinuePendingUnwind>(Interpreter& interpreter, u8 const* bytecode, size_t& program_counter);
+template<>
+void handle<Op::ScheduleJump>(Interpreter& interpreter, u8 const* bytecode, size_t& program_counter);
+template<>
+void handle<Op::Await>(Interpreter& interpreter, u8 const* bytecode, size_t& program_counter);
+template<>
+void handle<Op::Return>(Interpreter& interpreter, u8 const* bytecode, size_t& program_counter);
+template<>
+void handle<Op::Yield>(Interpreter& interpreter, u8 const* bytecode, size_t& program_counter);
```
alimpfard: Same for all of these, straight into the .cpp file please.

Libraries/LibJS/Bytecode/Interpreter.cpp
```diff
@@ -355,353 +355,237 @@ FLATTEN_ON_CLANG void Interpreter::run_bytecode(size_t entry_point)
     size_t& program_counter = running_execution_context.program_counter;
     program_counter = entry_point;
 
-    // Declare a lookup table for computed goto with each of the `handle_*` labels
-    // to avoid the overhead of a switch statement.
-    // This is a GCC extension, but it's also supported by Clang.
+    auto& next_instruction = *reinterpret_cast<Instruction const*>(&bytecode[program_counter]);
+    auto fn = Interpreter::dispatch_instruction_table[static_cast<size_t>(next_instruction.type())];
```
alimpfard: Maybe experiment with putting these pointers into the bytecode stream? this lookup is probably the single most mispredicted thing in your code.

Libraries/LibJS/Bytecode/Interpreter.cpp
```diff
@@ -355,353 +355,237 @@ FLATTEN_ON_CLANG void Interpreter::run_bytecode(size_t entry_point)
...
-
-        handle_Mov: {
-            auto& instruction = *reinterpret_cast<Op::Mov const*>(&bytecode[program_counter]);
-            set(instruction.dst(), get(instruction.src()));
-            DISPATCH_NEXT(Mov);
-        }
+template<>
+FLATTEN void handle<Op::Mov>(Interpreter& interpreter, u8 const* bytecode, size_t& program_counter)
+{
+    auto& instruction = *reinterpret_cast<Op::Mov const*>(&bytecode[program_counter]);
```
alimpfard: why not pass this in? doing the arithmetic for `&bytecode[program_counter]` again seems useless.

Libraries/LibJS/Bytecode/Interpreter.cpp
```diff
@@ -355,353 +355,237 @@ FLATTEN_ON_CLANG void Interpreter::run_bytecode(size_t entry_point)
...
-        for (;;) {
-            goto* bytecode_dispatch_table[static_cast<size_t>((*reinterpret_cast<Instruction const*>(&bytecode[program_counter])).type())];
-
-        handle_Mov: {
-            auto& instruction = *reinterpret_cast<Op::Mov const*>(&bytecode[program_counter]);
-            set(instruction.dst(), get(instruction.src()));
-            DISPATCH_NEXT(Mov);
-        }
+template<>
+FLATTEN void handle<Op::Mov>(Interpreter& interpreter, u8 const* bytecode, size_t& program_counter)
```
alimpfard: this program_counter ref is _probably_ also forcing a load/store, might be worth having this return the next PC instead similar to wasm if the compiler isn't optimising these mem lookups away.

Libraries/LibJS/Bytecode/Interpreter.h
```diff
@@ -106,6 +107,63 @@ class JS_API Interpreter {
...
+#undef ComparisonOpImpl
+
+    // clang-format off
+    // Clang keeps increasing the indent, as the macros dont have trailing commas
+    #pragma GCC diagnostic push
+    #pragma GCC diagnostic ignored "-Wc99-designator"
+    // The [index] = value, array declaration style is a C99 feature, but both GCC and Clang support it in C++,
+    // but warn about it.
+    // This is mainly used to allow for allowing us to declare the values out of order,
+    // with macros generating most of the entries.
```
alimpfard: You can do this without the silly feature, just generate it in a consteval iife like-

Libraries/LibJS/Bytecode/Interpreter.h
```diff
@@ -106,6 +107,63 @@ class JS_API Interpreter {
...
+        JS_ENUMERATE_SIMPLE_JUMP_OPS(JumpOp)
+        [to_underlying(Instruction::Type::Mov)] = handle<Op::Mov>,
+        [to_underlying(Instruction::Type::End)] = handle<Op::End>,
+        [to_underlying(Instruction::Type::EnterUnwindContext)] = handle<Op::EnterUnwindContext>,
+        [to_underlying(Instruction::Type::ContinuePendingUnwind)] = handle<Op::ContinuePendingUnwind>,
+        [to_underlying(Instruction::Type::ScheduleJump)] = handle<Op::ScheduleJump>,
+        [to_underlying(Instruction::Type::Await)] = handle<Op::Await>,
+        [to_underlying(Instruction::Type::Return)] = handle<Op::Return>,
+        [to_underlying(Instruction::Type::Yield)] = handle<Op::Yield>,
+    };
```
alimpfard:
```suggestion
    typedef void (*dispatch_instruction_table_t)(Interpreter&, u8 const*, size_t&);
    static constexpr auto dispatch_instruction_table = [] {
        Array<dispatch_instruction_table_t, to_underlying(Instruction::Type::__Last)> result; // add __Last to the enum.
        JS_ENUMERATE_WHATEVER(...);
        result[to_underlying(Instruction::Type::Mov)] = handle<Op::Mov>;
        return result;
    }();
```

Libraries/LibJS/Bytecode/Instructions.h
```diff
@@ -0,0 +1,218 @@
+/*
+ * Copyright (c) 2025, Leon Albrecht <leon.a@serenityos.org>
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#pragma once
+
+#include "Instruction.h"
```
alimpfard: first, `#include <LibJS/Bytecode/Instruction.h>`
second, really not a nice name - Instructions.h vs Instruction.h is just confusing.

Libraries/LibJS/Bytecode/Instructions.h
```diff
@@ -0,0 +1,218 @@
...
+    X(UnaryPlus)                             \
+    X(UnsignedRightShift)
+
+#define JS_ENUMERATE_SIMPLE_JUMP_OPS(X) \
+    X(Jump)                             \
+    X(JumpIf)                           \
+    X(JumpTrue)                         \
+    X(JumpFalse)                        \
+    X(JumpNullish)                      \
+    X(JumpUndefined)
```
alimpfard: Didn't we have a different copy of this somewhere? please define that in terms of this instead.

Libraries/LibJS/Bytecode/Instructions.h
```diff
@@ -0,0 +1,218 @@
...
+
+template<typename OP>
+void handle(Interpreter& interpreter, u8 const* bytecode, size_t& program_counter);
+template<typename OP>
+void handle_generic(Interpreter& interpreter, u8 const* bytecode, size_t& program_counter);
+template<typename OP>
+void handle_jump(Interpreter& interpreter, u8 const* bytecode, size_t& program_counter);
+
+// FIXME: For some reason we need to pre-define the specializations some instructions?
```
alimpfard: > pre-define

forward-declare is the correct term :)

Libraries/LibJS/Bytecode/Instructions.h
Hendiadyoin1: These need to be visible by the `Interpreter.h` otherwise that one instantiates them before they are specialized,
at least that's what clang tells me

Libraries/LibJS/Bytecode/Interpreter.cpp
Hendiadyoin1: yeah that helped a bit with code-gen, will run benchmarks again once I am home

Libraries/LibJS/Bytecode/Interpreter.cpp
Hendiadyoin1: that might indeed be interesting, especially when we start pre-fetching these
but that might be something for a followup PR

Libraries/LibJS/Bytecode/Interpreter.cpp
Hendiadyoin1: you mean incrementing and decrementing the `bytecode` ptr as needed?
that might work will look into it once I have the results from passing in the pc by value

Libraries/LibJS/Bytecode/Interpreter.h
Hendiadyoin1: Sure

Libraries/LibJS/Bytecode/Interpreter.cpp
alimpfard: No, I mean passing in the &bytecode[pc] you already calculated to get the address to this handler.

Libraries/LibJS/Bytecode/Instructions.h
alimpfard: Hmm. okay.

Libraries/LibJS/Bytecode/Interpreter.h
alimpfard: (not really suggesting this over what you have btw, I just saw the comment and gave an alternative)

Libraries/LibJS/Bytecode/Instructions.h
Hendiadyoin1: Fair point, will think of a better name

Libraries/LibJS/Bytecode/Instructions.h
Hendiadyoin1: I think we only have one with all instructions, not some that differentiate between how they are implemented

alimpfard: Right, so move these there and define those in terms of these :)

Libraries/LibJS/Bytecode/Interpreter.cpp
Hendiadyoin1: so what would be the signature you propose?
I am not quite following
what I understood is
```c++
u8 const* handle<Op>(Interpreter&, u8 const* current_insn)
```
where we recover the PC as the delta from the return value and the original `bytecode` pointer,
jumps would need to calculate a delta, but that could be moved into codegen at some point

alimpfard: what I meant is that you pass all three and return the next pc for when you need to go back to the interpreter loop (void is fine if you don't need that):

```cpp
size_t handle<op>(Interpreter&, u8 const* bytecode, u8 const* insn_ptr, size_t pc)
{
   // do the thing
   insn = &bytecode[++pc];
   // do stuff to find handler
   TAILCALL return handler(interp, bytecode, insn, pc);
}
```

Libraries/LibJS/Bytecode/Interpreter.cpp
Hendiadyoin1: Ok I found out why this wont work
The Error constructor looks at the current execution context, which is the owner of the program counter,
so when we only update it when exiting the interpreter loop we have the wrong value in there, when we
construct an error

Libraries/LibJS/Bytecode/Instructions.h
Hendiadyoin1: Did so in a separate commit, which cleaned some things up a bit, but not sure how nice it really is

Hendiadyoin1: Did not 100% like that commit, dropped it for now

Meta/find_compiler.py
```diff
@@ -18,8 +18,8 @@
 from Meta.host_platform import Platform
 from Meta.utils import run_command
 
-CLANG_MINIMUM_VERSION = 19
-GCC_MINIMUM_VERSION = 14
+CLANG_MINIMUM_VERSION = 20
```
trflynn89: ? The documentation does not state we require clang 20, it just states that is what is used in CI.

Hendiadyoin1: Then I cant read, will fix

Hendiadyoin1:
Draft until
A) Ubuntu adds gcc-15 to the toolchain ppa on noble
or
B) I add a tail call less workaround, which likely causes a performance regression on GCC

Issue #6163, opened 2025-09-11
WSL2 VulkanContext creation failure lacking device extension


rcorsi:
### Summary

Related and followup to #5963 where on WSL2 the VulkanContext fails to be created because the Microsoft Linux Kernel 6.6.x used with all WSL2 Linux distributions is lacking support for shareable vulkan image.

Because the VulkanContext fails to be created the system falls back to Software GPU (`--force-cpu-painting`) and that appears to be another bug as described in #5963 for WSL2 and also #6031 for other environments, leading to the browser painting being transparent or very broken.

When starting Ladybird, the following output shows that the VulkanContext failed to be created

```
1319.390 WebContent(15777): Vulkan context creation failed: Logical device creation failed
1319.407 WebContent(15784): Vulkan context creation failed: Logical device creation failed
```

Specifically the device extension support for `VK_EXT_image_drm_format_modifier` is missing on WSL2 due to the Kernel 6.6.x lacking the needed device drivers.

Missing device extension might also be known as:

```
VK_EXT_IMAGE_DRM_FORMAT_MODIFIER_EXTENSION_NAME
```

To workaround the issue, comment out the line on Libraries/LibGfx/VulkanContext.h:18 with the text

```
#        define USE_VULKAN_IMAGES 1
```

And if the VulkanContext works as expected after building ladybird with the workaround, meaning that the following lines DO NOT appear,
```
1319.390 WebContent(15777): Vulkan context creation failed: Logical device creation failed
1319.407 WebContent(15784): Vulkan context creation failed: Logical device creation failed
```

then the system is correctly configured.

It might be possible to build a custom kernel to resolve the lacking device driver support or lacking that request that Microsoft fix the issue by requesting that the feature be supported by creating an issue at the link below.  It will probably take some time to get resolved...

https://github.com/microsoft/WSL2-Linux-Kernel

### Operating system

Linux

### Steps to reproduce

1. On Windows 10/11 install the latest Ubuntu in WSL2 virtual environment
2. build ladybird in Ubuntu WSL2 environment
3. VulkanContext will fail to be created and system will fall back to software CPU (`--force-cpu-painting`)
4. It is important to make sure the all required Ubuntu packages to properly support llvmpipe be installed (or whatever physical GPU is being used)

### Expected behavior

VulkanContext should work.

### Actual behavior

VulkanContext fails due to device extension VK_EXT_image_drm_format_modifier is not available on WSL2

[...]

PR #6512, opened 2025-10-19
LibGfx: Check for Vulkan device extension drm availability on Linux

rcorsi:
Add physical device extension availablity check to exclude the usage of sharable Vulkan image if extension is not available.

Currently on many Linux distributions the sharable Vulkan image device extension is not available. Due to this, the VulkanContext fails to be created and ladybird falls back to the CPU backend painter.

Adding a configure tool to check if the required extensions are available and set appropriate variables.

Issue was first introduce when the shareable Vulkan images where incorporated in #5864 specifically in this commit 60a7359f0f6827799621db8320b43e5d60d08ef2

Even on Ubuntu 24.04 it fails and the following error will be seen. 
```
1738.682 WebContent(100803): Vulkan context creation failed: Logical device creation failed
1738.769 WebContent(100808): Vulkan context creation failed: Logical device creation failed
```
The above error message does not reveal the real reason for the issue, but in many instances the cause is lack of support for the device extension: `VK_EXT_IMAGE_DRM_FORMAT_MODIFIER_EXTENSION_NAME` 

fixes: #6163

Related to: #5963, #6031, #6322, #6392, #6503

Tool is based on content in `Libraries/LibGfx/VulkanContext.cpp`, but has been reworked to remove dependency on AK library code. It is using just standard c++23 code.

An attempt was made to use the AK library code as part of the solution, as can be seen in this branch https://github.com/LadybirdBrowser/ladybird/compare/master...rcorsi:ladybird:vulkan_extension_check , but encountered issues with TRY macro initially (which is removed) and then undefined references to ak_verification_failed, so AK/Assertions.cpp was added to the list of SOURCES, and then the problem was AK/Backtrace.h that had to be generated. I did try to remove reliance on AK/Vector by switching to std::vector, helped a bit, but not completely. It was getting out of hand, so I didn't continue on this avenue.  Maybe someone can help out if that would be the better way.

### DIFF
--- a/Libraries/LibGfx/VulkanContext.cpp
+++ b/Libraries/LibGfx/VulkanContext.cpp
@@ -100,13 +100,15 @@ static ErrorOr<VkDevice> create_logical_device(VkPhysicalDevice physical_device,
 
     VkPhysicalDeviceFeatures deviceFeatures {};
 #ifdef USE_VULKAN_IMAGES
+    dbgln("Using shareable Vulcan images");
     char const* device_extensions[] = {
         VK_KHR_EXTERNAL_MEMORY_FD_EXTENSION_NAME,
         VK_EXT_IMAGE_DRM_FORMAT_MODIFIER_EXTENSION_NAME
     };
     uint32_t device_extension_count = array_size(device_extensions);
 #else
-    const char** device_extensions = nullptr;
+    dbgln("Not using shareable Vulcan images");
+    char const** device_extensions = nullptr;
     uint32_t device_extension_count = 0;
 #endif
     VkDeviceCreateInfo create_device_info {};
--- a/Libraries/LibGfx/VulkanContext.h
+++ b/Libraries/LibGfx/VulkanContext.h
@@ -7,15 +7,13 @@
 #pragma once
 
 #ifdef USE_VULKAN
-
-#    include <AK/Assertions.h>
-#    include <AK/NonnullRefPtr.h>
-#    include <AK/RefCounted.h>
 #    include <vulkan/vulkan.h>
-#    ifdef AK_OS_LINUX
+
+#    ifdef USE_VULKAN_IMAGES
+#        include <AK/Assertions.h>
+#        include <AK/NonnullRefPtr.h>
+#        include <AK/RefCounted.h>
 #        include <libdrm/drm_fourcc.h>
-// Sharable Vulkan images are currently only implemented on Linux
-#        define USE_VULKAN_IMAGES 1
 #    endif
 
 namespace Gfx {
--- a/Meta/CMake/vulkan-drm-checker.cpp
+++ b/Meta/CMake/vulkan-drm-checker.cpp
@@ -1,0 +1,133 @@
+/*
+ * Copyright (c) 2024, Aliaksandr Kalenik <kalenik.aliaksandr@gmail.com>
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include <expected>
+#include <iostream>
+#include <string>
+#include <vector>
+#include <vulkan/vulkan.h>
+
+using std::cout;
+using std::expected;
+using std::string;
+using std::unexpected;
+using std::vector;
+
+static expected<VkInstance, string> create_instance(uint32_t api_version)
+{
+    VkInstance instance;
+
+    VkApplicationInfo app_info {};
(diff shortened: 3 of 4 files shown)


rcorsi:
@manuel-za this is the fix for the DRM or the shareable vulkan images issue.

Currently having a problem in CI where the fuzzer wants to include my tool as part of the fuzzing. Need to figure out how to remove that.

rcorsi:
Haven't been able to remove the fuzzer from the linking of the tool. I tried a few things.

rcorsi:
For disabling fuzzing during build I did find the following in `Meta/Lagom/CMakeLists.txt` on lines 72-79

```
# We need to make sure not to build code generators for Fuzzer builds, as they already have their own main.cpp
# Instead, we import them from a previous install of Lagom. This mandates a two-stage build for fuzzers.
# The same concern goes for cross-compile builds, where we need the tools built for the host
set(BUILD_LAGOM_TOOLS ON)
if (ENABLE_FUZZERS OR CMAKE_CROSSCOMPILING)
    find_package(LagomTools REQUIRED)
    set(BUILD_LAGOM_TOOLS OFF)
endif()
```

In the case of this tool that I am attempting to build it has its own main(), so we need to not have the fuzzer linked in.  Not sure how to resolve this.

InvalidUsernameException:
As a general thought: Would it make sense to do this check at runtime rather than at compile time?

With binary distributed versions in the future, I presume a single binary might run both on systems where the extensions are available, and on some where they are not (real hardware vs. WSL vs. regular VM). Given that the compile-time version currently is having some difficulties with the build system, that might save some headaches now and could be more future proof anyway.

rcorsi:
@InvalidUsernameException  Doing it at runtime is a good idea, for sure.  Actually that is what I initially did, but maybe not in the right way, so I came back to this build time detection.

If I remember correctly when I used the vulkan validation layers some time ago, it was saying something that as a best practice you should check the availability of the extensions before using them.

It would certainly make the feature flexible to fit any situation, where it may or may not be supported in a particular Linux distribution, due to Vulkan version, due to kernel version or kernel issues, or driver issues, or setup issues, or hardware issues, etc...

Let me see what I can come up with.

Thanks for the feedback.

manuel-za:
Suggestions:

1. Two checks may be optimal:
   - At build time: avoid the runtime "failed Vulkan context creation" overhead every time you open a tab.
   - At runtime: protect against [some] changes to the machine configuration.

2. It may be more efficient + future-proof to simply test for success (vs. going through the enumeration). I propose the following algorithms:

```
// Build environment detection
if ( create_logical_device( ...VK_EXT_IMAGE_DRM_FORMAT... ))
   USE_VULKAN and USE_VULKAN_IMAGES
else if ( create_logical_device( no VULKAN_IMAGES )
   USE_VULKAN and do not USE_VULKAN_IMAGES
else
   do not USE_VULKAN
```

```
// Runtime
if( create_logical_device( USE_VULKAN_IMAGES ) fails): fallback on USE_VULKAN
if( create_logical_device( USE_VULKAN ) fails: fallback on CPU painting
```

manuel-za:
Happy to take this on. Still unsolved + I still think it's useful. cc: @rcorsi

rcorsi:
> Happy to take this on. Still unsolved + I still think it's useful. cc: @rcorsi

Yeah, I agree it would be very nice to solve. I think I will put my other commits aside and work on this one. Seems to be the most important thing I've looked at recently.

rcorsi:
@manuel-za I will be closing this PR and opening a new one when I have the updated version.  It will be based on your and @InvalidUsernameException comments in this PR.

This new PR will:
1- check for the image drm extensions at runtime only once at start and save the state if they are available or not.
2- also once at start the picked GPU or IPU or falling back to backend CPU painting will be reported, avoiding spamming the log
3- every time a new vulkan context is needed, the extensions will be included or not, based on the known state discovered at the start. (item #1)
4- WebGL supported correctly by this change. I need to look into what WebGL is doing as they may need some special handling, from what I remember WebGL was expecting the image drm extension to be available if the vulkan context is available.  Now we may have a vulkan context, without an image drm extension. Not sure the impacts of that on WebGL.
5- Support macOS Metal context correctly by this change.
6- maybe some other stuff I haven't though of yet.

rcorsi:
Closing for now, but expecting to create a new PR soon.

PR #6519, opened 2025-10-20
Better CSS inheritance for nodes that represent a pseudo-element

awesomekling:
When we compute style for elements inside a UA-internal shadow tree that represent a pseudo-element (e.g `::placeholder`), we actually run the `StyleComputer` machinery for (host element :: pseudo-element).
    
While that lets us match the correct selectors, it was incorrectly applying CSS inheritance, since we'd also then inherit from whatever was above the host element in the tree.
    
This patch fixes the issue by introducing an inheritance override in `AbstractElement` and then using that to force inheritance from whatever is actually directly above in the DOM for these elements instead of jumping all the way up past the host.
    
This fixes an issue where `text-align: center` on `input type=text` elements would render the main text centered but placeholder text was still left-aligned.

Visual progression on https://gymgrossisten.com

Before:
Screenshot 2025-10-20 at 14 20 41

After:
Screenshot 2025-10-20 at 12 31 46

### DIFF
--- a/Libraries/LibWeb/DOM/AbstractElement.h
+++ b/Libraries/LibWeb/DOM/AbstractElement.h
@@ -35,6 +35,8 @@ public:
     Optional<AbstractElement> previous_sibling_in_tree_order() { return walk_layout_tree(WalkMethod::PreviousSibling); }
     bool is_before(AbstractElement const&) const;
 
+    void set_inheritance_override(GC::Ref<Element> element) { m_inheritance_override = element; }
+
     GC::Ptr<CSS::ComputedProperties const> computed_properties() const;
 
     void set_custom_properties(OrderedHashMap<FlyString, CSS::StyleProperty>&& custom_properties);
@@ -63,6 +65,8 @@ private:
 
     GC::Ref<Element> m_element;
     Optional<CSS::PseudoElement> m_pseudo_element;
+
+    GC::Ptr<Element> m_inheritance_override;
 };
 
 }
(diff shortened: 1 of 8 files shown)


Libraries/LibWeb/DOM/AbstractElement.h
```diff
@@ -61,6 +63,8 @@ class WEB_API AbstractElement {
 
     GC::Ref<Element> m_element;
     Optional<CSS::PseudoElement> m_pseudo_element;
+
+    GC::Ptr<Element> m_inheritance_override;
```
AtkinsSJ, line 69: This needs visiting.

awesomekling: There is no visit function as this is never heap—allocated :)

It gets marked by conservative stack scan.

AtkinsSJ: GC still confuses me sometimes. 😅 

We do have `AbstractElement::visit()`, because for a lot of places `AbstractElement` is used as a long-term handle to an element. But yeah this mechanism is only used on a temporary `AbstractElement`.

It's kind of weird to me to put state on this thing, to be honest. But it does solve this one issue very simply.

awesomekling: Ah didn't see that `visit()`, good catch!

Hmm, agree that adding state here is awkward. Let me try a different approach...

awesomekling: I spent some time to redo this by passing the inheritance override to everyone who needs it in StyleComputer instead. It got incredibly invasive and messy :(
IMO let's roll with the slightly hackish override in AbstractElement for now and maybe we'll think of a better approach later.

AtkinsSJ: > It got incredibly invasive and messy :(

Yeah I can imagine. :sweat_smile:

I'm fine with it for now.

PR #6520, opened 2025-10-20
LibWeb/WebGL: Replace manual pointer math with Span

Lubrsi:
This deduplicates a lot of sensitive pointer math by using Span, which
performs this math for us with more safety checks and with type
information.

This also allows us to use the correctly typed Span for typed arrays,
which automatically fixes srcOffset to now offset by the number of
elements instead of bytes. This goes for srcLengthOverride too.

Fixes the Rive animations on Shopify's homepage not appearing.
Fixes some Unity applications such as ArcViewer having missing
graphics.

### DIFF
--- a/Libraries/LibWeb/WebGL/WebGL2RenderingContextImpl.cpp
+++ b/Libraries/LibWeb/WebGL/WebGL2RenderingContextImpl.cpp
@@ -31,6 +31,16 @@ extern "C" {
 #include <LibWeb/WebGL/WebGLVertexArrayObject.h>
 #include <LibWeb/WebIDL/Buffers.h>
 
+#define SET_ERROR_VALUE_IF_ERROR(expression, error_value) \
+    ({                                                    \
+        auto maybe_error = expression;                    \
+        if (maybe_error.is_error()) [[unlikely]] {        \
+            set_error(error_value);                       \
+            return;                                       \
+        }                                                 \
+        maybe_error.release_value();                      \
+    })
+
 namespace Web::WebGL {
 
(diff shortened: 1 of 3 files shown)


AK/Span.h
```diff
@@ -354,6 +354,20 @@ class Span : public Detail::Span<T> {
         }
         return {};
     }
+
+    template<typename TargetType>
+    ALWAYS_INLINE constexpr Span<TargetType> reinterpret()
+    requires(sizeof(T) % sizeof(TargetType) == 0)
```
gmta: So this would allow for a span of `u16`s to be reinterpreted as `u8`s, but not the other way around? Maybe instead of a `requires` you could introduce a constexpr/runtime check on the size() if `sizeof(T) % sizeof(TargetType) > 0`?

Lubrsi: Maybe it could be a separate reinterpret implementation that can return an `ErrorOr<Span<T>>` and returns an error if there's not enough bytes, otherwise the reinterpreted Span? That way we can keep this implementation where it's guaranteed we won't drop bytes and thus don't have to return an error.

gmta: That would work!

Lubrsi: Thinking about it, would it be preferable to return an Error or just VERIFY instead? None of the other Span APIs return an Error and just VERIFY.

gmta: Feels natural to trap if we pass an incompatible span size to a reinterpret action imo?

Lubrsi: Sure, changed it to do that :^)

gmta (approved):

PR #6525, opened 2025-10-20
CSS: StepsEasingFunction use i64 instead of long

R-Goc:
Unbreaks windows build. 32bit long on Windows strikes back. resolve_integer returns an i64 making creation of a StepsEasingFunction with its result a narrowing conversion before this change.

### DIFF
--- a/Libraries/LibWeb/CSS/EasingFunction.h
+++ b/Libraries/LibWeb/CSS/EasingFunction.h
@@ -42,7 +42,7 @@ struct CubicBezierEasingFunction {
 };
 
 struct StepsEasingFunction {
-    long interval_count;
+    i64 interval_count;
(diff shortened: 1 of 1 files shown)


Libraries/LibWeb/CSS/EasingFunction.h
gmta, line 1: Commit title: move some words around, i.e. "CSS: Use i64 instead of long in StepsEasingFunction"

Commit message: use "Windows" instead of "windows". Also, "32bit long on Windows strikes back." does not add anything of value.

R-Goc: Doesn't fit by one character, so ommited the instead of long.

PR #6526, opened 2025-10-21
LibWeb: Improve performance around CSS tree counting functions

Calme1709:
We can skip a bunch of unnecessary work in cases where tree counting functions aren't actually used.

Reduces runtime of `DOM::Element::children_changed()` from 37.80% to 0.12% and `DOM::AbstractElement::tree_counting_function_resolution_context()` from 37.77% to 0% in a profile of http://wpt.live/encoding/legacy-mb-korean/euc-kr/euckr-decode-cseuckr.html?11001-12000

I am not sure if storing the callback as a `Function*` in `ComputationContext` is the right approach but it needs to be optional and copyable (for use in `CalculationResolutionContext::from_computation_context`), would appreciate feedback.

### DIFF
--- a/Libraries/LibWeb/CSS/StyleComputer.cpp
+++ b/Libraries/LibWeb/CSS/StyleComputer.cpp
@@ -895,16 +895,6 @@ static void cascade_custom_properties(DOM::AbstractElement abstract_element, Vec
     custom_properties.update(important_custom_properties);
 }
 
-// TODO: This could be useful for marking which properties need recomputation when inherited style changes
-ALWAYS_INLINE static void mark_property_computation_dependencies(Optional<DOM::AbstractElement> abstract_element, PropertyComputationDependencies const& property_computation_dependencies)
-{
-    if (!abstract_element.has_value())
-        return;
-
-    if (property_computation_dependencies.tree_counting_function)
-        abstract_element->element().set_style_uses_tree_counting_function();
-}
-
 void StyleComputer::collect_animation_into(DOM::AbstractElement abstract_element, GC::Ref<Animations::KeyframeEffect> effect, ComputedProperties& computed_properties) const
 {
     auto animation = effect->associated_animation();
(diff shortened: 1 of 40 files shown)


Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -2062,11 +2062,11 @@ void StyleComputer::compute_font(ComputedProperties& style, Optional<DOM::Abstra
     auto inherited_font_size = inheritance_parent_has_computed_properties ? inheritance_parent->computed_properties()->font_size() : InitialValues::font_size();
     auto inherited_math_depth = inheritance_parent_has_computed_properties ? inheritance_parent->computed_properties()->math_depth() : InitialValues::math_depth();
 
-    auto tree_counting_function_resolution_context = abstract_element.map([](auto abstract_element) { return abstract_element.tree_counting_function_resolution_context(); }).value_or({ .sibling_count = 1, .sibling_index = 1 });
+    Function<TreeCountingFunctionResolutionContext()> tree_counting_function_resolution_context = [&abstract_element] { return abstract_element.map([](auto abstract_element) { return abstract_element.tree_counting_function_resolution_context(); }).value_or({ .sibling_count = 1, .sibling_index = 1 }); };
```
gmta: Could you spread some of these longer lines over multiple lines? We wrap at column 120.

Libraries/LibWeb/DOM/Element.cpp
```diff
@@ -1395,14 +1395,17 @@ void Element::children_changed(ChildrenChangedMetadata const* metadata)
...
-    for_each_child([&](DOM::Node& child) {
-        if (auto* element = as_if<DOM::Element>(child); element && element->style_uses_tree_counting_function()) {
-            element->set_needs_style_update(true);
-            set_child_needs_style_update(true);
-        }
 
-        return IterationDecision::Continue;
-    });
+    if (child_style_uses_tree_counting_function()) {
+        for_each_child([&](DOM::Node& child) {
```
gmta: You can use `for_each_child_of_type<DOM::Element>` instead

Libraries/LibWeb/DOM/Element.h
```diff
@@ -637,6 +646,7 @@ class WEB_API Element
     bool m_style_uses_attr_css_function : 1 { false };
     bool m_style_uses_var_css_function : 1 { false };
     bool m_style_uses_tree_counting_function : 1 { false };
+    bool m_child_style_uses_tree_counting_function : 1 { false };
```
gmta, line 651: If styles are changed and tree counting functions are no longer used, how are these reset to false?

Libraries/LibWeb/CSS/CalculationResolutionContext.h
```diff
@@ -19,7 +19,7 @@ struct CalculationResolutionContext {
 
     PercentageBasis percentage_basis {};
     Optional<Length::ResolutionContext> length_resolution_context {};
-    Optional<TreeCountingFunctionResolutionContext> tree_counting_function_resolution_context {};
+    Function<TreeCountingFunctionResolutionContext()>* tree_counting_function_resolution_context = nullptr;
```
gmta: Not sure if this is safe to use like this, but someone else will have to step in to confirm 😅

AtkinsSJ:
Something I noticed when reviewing the previous PR yesterday, is that those `Optional<AbstractElement>` params don't need to be `Optional`. The only place they don't have a value is `StyleComputer::create_document_style()`, where we could create an `AbstractElement` from the Document's `document_element()`.

Without that, every caller seems to use `AbstractElement::tree_counting_function_resolution_context()` directly, so I think we could just store the `AbstractElement` on the context instead. I don't know how that interacts with GC, but I don't think it's *less* safe than this `Function*` approach, as that already captures the `AbstractElement`.

Calme1709:
I think passing the `AbstractElement` as part of the `{Computation,CalculationResolution}Context` is probably a good approach - it will also save us having to pass around the ugly `PropertyComputationDependencies` struct as well since we can just set the flags directly. Will look at doing that.

Libraries/LibWeb/DOM/Element.h
Calme1709: Currently, as with the other `m_style_uses_{attr,var}_css_function` flags they are not - I have added a FIXME that doing so could be a potential performance improvement.

Libraries/LibWeb/DOM/Element.cpp
Calme1709: Done.

Libraries/LibWeb/CSS/CalculationResolutionContext.h
Calme1709: No longer relevant.

Libraries/LibWeb/CSS/StyleComputer.cpp
Calme1709: No longer relevant.

gmta (dismissed):

Libraries/LibWeb/DOM/AbstractElement.h
```diff
@@ -27,7 +27,11 @@ class WEB_API AbstractElement {
     GC::Ptr<Layout::NodeWithStyle> layout_node();
     GC::Ptr<Layout::NodeWithStyle const> layout_node() const { return const_cast<AbstractElement*>(this)->layout_node(); }
 
-    CSS::TreeCountingFunctionResolutionContext tree_counting_function_resolution_context() const;
+    struct TreeCountingFunctionResolutionContext {
+        size_t sibling_count;
+        size_t sibling_index;
+    };
```
gmta, line 33: Moving the struct causes this commit not to build for me:

```
/ladybird/Libraries/LibWeb/DOM/AbstractElement.cpp:30:1: error: unknown type name 'TreeCountingFunctionResolutionContext'
   30 | TreeCountingFunctionResolutionContext AbstractElement::tree_counting_function_resolution_context() const
      | ^
/ladybird/Libraries/LibWeb/DOM/AbstractElement.cpp:49:16: error: excess elements in scalar initializer
   49 |         return { .sibling_count = 1, .sibling_index = 1 };
      |                ^                   ~~~~~~~~~~~~~~~~~~~~~~
/ladybird/Libraries/LibWeb/DOM/AbstractElement.cpp:60:12: error: excess elements in scalar initializer
   60 |     return {
      |            ^
   61 |         .sibling_count = count,
      |                               ~
   62 |         .sibling_index = index
      |         ~~~~~~~~~~~~~~~~~~~~~~
   63 |     };
      |     ~
3 errors generated.
```

Calme1709: Whoops - accidentally split that over two commits, fixed.

Issue #6505, opened 2025-10-18
On window resize, tab content disappears for some time or wrong content presented


rcorsi:
Especially when in CPU backend painter 'mode', same as --force-cpu-painting, when the main window is resized the tab content will be completely removed and will reappear after some time.

If multiple tabs are open with different content, and after the window is resized, when moving from one tab to the next, content from other tab may appear in the wrong tab.

As a concrete example, using `about:processes` (and --force-cpu-painting), with the processes listed and being updated every second, when the window is resized the content will disappear and reappear about 3 seconds later.

The issue is not usually present when the VulkanContext is successfully created.

With MetalContext on MacOS Intel this problem may be present. Unknown what occurs on MacOS Apple Silicon (Mx series).

This is a followup to #6503, as this issue is still present after that fix is included.

Attaching some older screen capture where the issue is visible.

- On Linux VM with Ubuntu 25.04, both the #6503 can be seen and also this issue.

https://github.com/user-attachments/assets/c8929b78-49dc-4415-a7f6-59bbd57abf89

- On a MacOS Intel based system, where the --force-cpu-painting was NOT used, and based on deeper checks the MetalContext was successfully created.  This video shows the `about:processes` and google web site having screen disappearing/no updating for some time.

https://github.com/user-attachments/assets/731116f3-dccd-4027-9bad-4fc4caf0e401

I could make new videos with #6503 in place if needed, but the problem with the `about:processes` would be nearly identical.

rcorsi:
@manuel-za you're also seeing this issue, right?

If I look in your #6322 I think I get my answer, but I wouldn't mind an update with #6503 being used. It should change the result a bit.

rcorsi:
Best demo of the issue is to open three tabs:

1- `about:processes`
2- some web site, google for example
3- `about:version`

After resizing and content disappearing, 1 and 2 will eventually get updated.

Item 3 is a bit interesting, it doesn't get updated until manual reload is done with button or CTRL-R. Also if you resize the window and click on another tab and come back to the `about:version` tab, and just repeatedly clicking inside the tab, content will reappear after about 3 seconds, similar to `about:processes` where the content reappears after 3 seconds.

rcorsi:
[...]

PR #6527, opened 2025-10-21
LibWeb: Update painting backing stores in all cases

rcorsi:
To consistently update the backing stores in reallocate_backing_stores() the variables storing the backing stores need to be set to nullptr at the start so that they are properly updated later on near the end of the function.

fixes: #6505

also corrects

fixes: #5963
fixes: #6031
fixes: #6322
fixes: #6392

### DIFF
--- a/Libraries/LibWeb/Painting/BackingStoreManager.cpp
+++ b/Libraries/LibWeb/Painting/BackingStoreManager.cpp
@@ -59,6 +59,10 @@ BackingStoreManager::BackingStore BackingStoreManager::acquire_store_for_next_fr
 void BackingStoreManager::reallocate_backing_stores(Gfx::IntSize size)
 {
     auto skia_backend_context = m_navigable->skia_backend_context();
+
+    m_front_store = nullptr;
+    m_back_store = nullptr;
+
 #ifdef AK_OS_MACOS
     if (skia_backend_context && s_browser_mach_port.has_value()) {
         auto back_iosurface = Core::IOSurfaceHandle::create(size.width(), size.height());


rcorsi:
@manuel-za could you check out this fix.  I think it will fix all the issues with the cpu backend painter and seems that #6503 is not needed.

rcorsi:
I can confirm, this fixes the WSL2 transparent window on startup, fixes the artifacts when resizing or moving the window, fixes the 3 second blank screen when resizing or moving window, fixes some pages not appearing or delayed in appearing, fixes similar issues on macOS Intel based.

Its the good fix.

Right now I think #6503 is not needed or at best is an optional fix.

rcorsi:
> Oh, that's a silly mistake - thanks for figuring this out!

Well it was a really weird issue, so I like those.  Thankfully I had a lot of clues to help me out.

manuel-za:
Well done @rcorsi!

rcorsi:
You're welcome and everyone else that contributed reproductions and different clues and also to confirm that I was not the only one seeing problems.  Sometimes I think no one else is experiencing the same issues.  This one was a great exception as several people were reporting issues on difference platforms.  So thank you to all of you! :-)

rcorsi:
@HelloWorld-n and @Psychpsyo does this fix solve most of the graphical problems you were seeing?

Psychpsyo:
> @HelloWorld-n and @Psychpsyo does this fix solve most of the graphical problems you were seeing?

I sadly cannot tell you since the Linux VM in which I had those problems sank with my Windows install.

PR #6529, opened 2025-10-21
LibWeb: Implement error propagation in AudioCodecPluginAgnostic

boris-chu:
This PR implements proper error propagation for all asynchronous audio operations in `AudioCodecPluginAgnostic`. Previously, all promise rejections were silently ignored with `FIXME: Propagate errors` comments, leading to silent failures when audio operations failed.

## Changes

Added `.when_rejected()` error handlers for all 5 async operations:
- `resume_playback()` - handles device unavailability, resource conflicts, permission issues
- `pause_playback()` - handles buffer drain failures, stream suspension errors  
- `set_volume()` - handles audio device configuration issues
- `seek()` - handles both seek operation failures and post-seek resume failures

All error handlers:
- Use `make_weak_ptr<AudioCodecPluginAgnostic>()` to prevent use-after-free bugs
- Format errors with context using `String::formatted()`
- Propagate to `on_decoder_error()` callback for HTMLMediaElement notification

Added comprehensive architectural documentation (27 lines) explaining the error handling pattern, propagation strategy, common scenarios, and error flow.

## Testing

Created `Tests/LibWeb/Platform/TestAudioCodecErrorHandling.cpp` (267 lines) with 3 test cases:
- Error callback mechanism verification
- Error message formatting validation  
- Documentation completeness check

Tests gracefully handle CI environments without audio hardware. Manual testing scenarios documented for real hardware testing.

## Impact

This ensures:
- HTMLMediaElement receives notifications when audio operations fail
- JavaScript can handle errors via `onerror` event handlers
- Users get meaningful error messages instead of silent failures
- Developers can debug audio issues with proper context

Properly handles common failures like device disconnection, resource exhaustion, permission denial, and driver errors.

### DIFF
--- a/Libraries/LibWeb/Platform/AudioCodecPluginAgnostic.cpp
+++ b/Libraries/LibWeb/Platform/AudioCodecPluginAgnostic.cpp
@@ -4,6 +4,34 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
+// NOTE: Error Handling Pattern in AudioCodecPluginAgnostic
+// ========================================================
+//
+// This class uses an asynchronous promise-based pattern for audio operations. All async operations
+// can fail, and it's critical to propagate these errors to the media element for proper user feedback.
+//
+// Error Propagation Strategy:
+// ---------------------------
+// - All promise rejections are handled via `.when_rejected()` callbacks
+// - Errors are formatted with context and passed to `on_decoder_error()` callback
+// - The `on_decoder_error` function (inherited from AudioCodecPlugin) notifies the HTML media element
+// - Each error handler captures a weak pointer to `this` to safely handle object destruction
+// - Error messages are formatted using String::formatted() to provide actionable diagnostics
+//
+// Common Error Scenarios:
+// ----------------------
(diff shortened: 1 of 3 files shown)


Zaggy1024:
A few things here:
- AudioCodecPluginAgnostic is going to be deleted in favor of directly using AudioDataProvider, AudioMixingSink and PlaybackStream when #6410 is merged.
- PlaybackStream promises are resolved off the main thread, and Weakable/WeakPtr are not thread-safe, so this implementation may UAF if the plugin is destroyed as an error is occurring. The new playback system solves this by dispatching events in the promise callbacks to the main thread to run higher-level callbacks.
- The big comments explaining error propagation strategy shouldn't be necessary in the code, only explain things that aren't obvious from reading the code itself.

boris-chu:
Thanks for the detailed feedback, Gregory!

You're right about all three points:
1. Since AudioCodecPluginAgnostic is being removed in #6410, this work won't provide long-term value
2. The WeakPtr thread safety issue is real - I see now that promises resolve off the main thread
3. The comments were indeed too verbose

I'll close this PR. Thanks for explaining both the architectural direction and the thread safety issue - valuable learning experience!

PR #6535, opened 2025-10-21
LibWebView: Add common source loop source for deferred_invoke() on macOS

kalenikaliaksandr:
Previously, `EventLoopImplementationMacOS` didn't process `deferred_invoke()` callbacks during window resizing because they were scheduled in the "default" run-loop mode, which isn't serviced while the run loop is in "Event Tracking" mode.

Commit 3bbe1b0c changed socket notifiers to schedule in `kCFRunLoopCommonModes`, which kept IPC reads responsive during resizing (we process IPC messages in the notifier's read hook). This change does the same for `deferred_invoke()`: it installs a `CFRunLoopSource` in `kCFRunLoopCommonModes` and signals it whenever a deferred callback is enqueued.

### DIFF
--- a/Libraries/LibWebView/EventLoop/EventLoopImplementationMacOS.mm
+++ b/Libraries/LibWebView/EventLoop/EventLoopImplementationMacOS.mm
@@ -202,6 +202,40 @@ static void post_application_event()
     [NSApp postEvent:event atStart:NO];
 }
 
+struct EventLoopImplementationMacOS::Impl {
+    Impl(EventLoopImplementationMacOS& event_loop_implementation)
+        : run_loop(CFRunLoopGetCurrent())
+    {
+        CFRunLoopSourceContext context {};
+        context.info = &event_loop_implementation;
+        context.perform = [](void* info) {
+            auto& self = *static_cast<EventLoopImplementationMacOS*>(info);
+            self.m_impl->deferred_source_pending = false;
+            self.m_thread_event_queue.process();
+        };
+
+        deferred_source = CFRunLoopSourceCreate(kCFAllocatorDefault, 0, &context);
+        CFRunLoopAddSource(run_loop, deferred_source, kCFRunLoopCommonModes);
+    }
+
+    ~Impl()
+    {
+        CFRunLoopRemoveSource(run_loop, deferred_source, kCFRunLoopCommonModes);
+        CFRelease(deferred_source);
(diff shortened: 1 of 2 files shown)


Libraries/LibWebView/EventLoop/EventLoopImplementationMacOS.mm
```diff
@@ -202,6 +202,35 @@ static void post_application_event()
...
+    CFRunLoopSourceContext ctx {};
+    ctx.info = this;
+    ctx.perform = [](void* info) {
+        auto& self = *static_cast<EventLoopImplementationMacOS*>(info);
+        self.m_impl->deferred_source_pending = false;
+        self.m_thread_event_queue.process();
+    };
+    m_impl->deferred_source = CFRunLoopSourceCreate(kCFAllocatorDefault, 0, &ctx);
+    CFRunLoopAddSource(m_impl->run_loop, m_impl->deferred_source, kCFRunLoopCommonModes);
+}
```
trflynn89, line 235: This feels more appropriate to be in `Impl`'s constructor/destructor instead:
(let's also avoid shorthand names like `ctx`, we call related types `context` everywhere else in this file):

```c++
struct EventLoopImplementationMacOS::Impl {
    Impl()
        : run_loop(CFRunLoopGetCurrent())
    {
        CFRunLoopSourceContext context {};
        context.info = this;
        context.perform = [](void* info) {
            auto& self = *static_cast<EventLoopImplementationMacOS*>(info);
            self.m_impl->deferred_source_pending = false;
            self.m_thread_event_queue.process();
        };

        deferred_source = CFRunLoopSourceCreate(kCFAllocatorDefault, 0, &context);
        CFRunLoopAddSource(run_loop, deferred_source, kCFRunLoopCommonModes);
    }

    ~Impl()
    {
        CFRunLoopRemoveSource(run_loop, deferred_source, kCFRunLoopCommonModes);
        CFRelease(deferred_source);
    }

    CFRunLoopRef run_loop { nullptr };
    CFRunLoopSourceRef deferred_source { nullptr };
    Atomic<bool> deferred_source_pending { false };
};

EventLoopImplementationMacOS::EventLoopImplementationMacOS()
    : m_impl(make<Impl>())
{
}

EventLoopImplementationMacOS::~EventLoopImplementationMacOS() = default;
```

PR #6536, opened 2025-10-21
LibWeb: Reset Painter when resetting canvas to its initial state

tcl3:
This ensures that the canvas clipping area and transforms are reset when the canvas is reset.


Libraries/LibGfx/PainterSkia.cpp
```diff
@@ -115,6 +115,10 @@ static SkPaint to_skia_paint(Gfx::PaintStyle const& style, Optional<Gfx::Filter
 PainterSkia::PainterSkia(NonnullRefPtr<Gfx::PaintingSurface> painting_surface)
     : m_impl(adopt_own(*new Impl { move(painting_surface) }))
 {
+    m_impl->with_canvas([this](auto& canvas) {
+        m_initial_save_count = canvas.getSaveCount();
+        canvas.save();
```
gmta: I think you can just do

```
m_initial_save_count = canvas.save();
```

here. Returns the same value.

gmta: You changed it in the second commit, but you introduced this code in the first.

tcl3: Oops :sweat_smile:

PR #6537, opened 2025-10-21
Everywhere: Implement synchronization primitives for win32

R-Goc:
The goal of this PR is to remove the dependency on pthread outside of LibThreading. Pthreads are an implementation detail and we shouldn't be relying on them everywhere. This PR is the first part towards this. Synchronization APIs are abstracted away and added where needed. This will solve the biggest issue with pthreads-win32 which is lack of interprocess synchronization primitives. Future PRs will move further with this, but as this is very conflict prone I decided to split it up.


Libraries/LibThreading/Once.h
```diff
@@ -0,0 +1,36 @@
...
+    if (!flag.has_been_called.load(MemoryOrder::memory_order_acquire)) {
+        MutexLocker lock(flag.mutex);
+
+        // Another thread may have called the function while we were waiting on the mutex
+        // The mutex guarantees exclusivity so we can use relaxed ordering
+        if (flag.has_been_called.load(MemoryOrder::memory_order_relaxed))
+            return;
+
+        callable();
+        flag.has_been_called.store(true, MemoryOrder::memory_order_relaxed);
```
gmta: Shouldn't this use release instead, to match the earlier acquire? I think that's needed for proper ordering in threads that skip this whole block (including its mutex).

R-Goc: I was thinking about this. The second load can be relaxed for sure. I think I'll make the store release ordering, even if it is just for the compiler fence. Not entirely sure if it necessary with the mutex in place.

gmta: It's needed for threads that don't take the mutex at all I believe

R-Goc:
Opened this up for review.  I think I will drop the commit for ThreadID as it doesn't seem to make much sense standalone. The large number of files touched is due to the rename, which was needed due to LibThreading depending on LibCore while LibCore uses parts of LibThreading that were previously in headers. I decided to cut this here as it is rather conflict happy. What is left is implementing a thread local storage wrapper and implementing Thread natively on windows(+the threadID class if I drop it here). I am not happy with the API of ConditionVariable. I think it's wait function should take the lock as an argument, that way you are able to wait on more than one lock per ConditionVariable class and would allow having a single ConditionVariable class as on windows there are separate wait methods per underlying lock. I am not implementing an interprocess RWLock in this PR as that is something that would either need a library or care in writing one from scratch on windows. Same thing with interprocess condition variables. These would need to be written from scratch in userspace which would very likely end up with horrible performance.

Libraries/LibThreading/Once.h
Zaggy1024, line 1: Can't calling a function once be implemented simply through a compare/exchange, e.g.:

```cpp
static Atomic<bool> is_set_up = false;
auto expected = false;
if (is_set_up.compare_exchange_strong(expected, true))
    set_up();
```

Libraries/LibCore/ThreadEventQueue.cpp
```diff
@@ -46,11 +47,11 @@ struct ThreadEventQueue::Private {
 };
 
 static pthread_key_t s_current_thread_event_queue_key;
-static pthread_once_t s_current_thread_event_queue_key_once = PTHREAD_ONCE_INIT;
+static Threading::OnceFlag s_current_thread_event_queue_key_once {};
 
 ThreadEventQueue& ThreadEventQueue::current()
 {
-    pthread_once(&s_current_thread_event_queue_key_once, [] {
+    call_once(s_current_thread_event_queue_key_once, [] {
```
Zaggy1024: Instead of turning this into a `call_once()`, why not use `static thread_local ThreadEventQueue s_current_thread_event_queue`?

(And if we can do that safely, then we should probably also drop the commit adding `call_once()`, especially if it's possible to do this simply through compare/exchange, since that's not platform-dependent.)

Libraries/LibThreading/Thread.h
```diff
@@ -12,12 +12,34 @@
...
 #include <AK/DistinctNumeric.h>
 #include <AK/Function.h>
+#include <AK/Noncopyable.h>
 #include <AK/Result.h>
 #include <LibCore/EventReceiver.h>
 #include <pthread.h>
 
 namespace Threading {
 
+class ThreadId {
```
Zaggy1024: s/ThreadId/ThreadID/ please

Also, this should probably go in its own header, right? We may have uses for getting the thread ID outside of our self-spawned threads.

Libraries/LibSync/MutexPOSIX.cpp
Zaggy1024, line 1: Could we name this implementation file to indicate that it's specific to POSIX?

(And `ConditionVariable.cpp`)

Zaggy1024, line 1: Could we change this to a header with the functions defined as inline, conditionally-included in `Mutex.h`?

Same for `ConditionVariable.cpp`, and if one or both can't, then out-of-line is fine with me.

Libraries/LibThreading/BackgroundAction.cpp
```diff
@@ -6,8 +6,8 @@
  */
 
 #include <AK/Queue.h>
-#include <LibThreading/BackgroundAction.h>
 #include <LibSync/Mutex.h>
+#include <LibThreading/BackgroundAction.h>
```
Zaggy1024, line 10: Looks like this should go in the previous commit. This happened in `RenderingThread.h` as well, so it would be good to amend that commit with a full reformat to catch everything.

Meta/Lagom/CMakeLists.txt
```diff
@@ -66,7 +66,7 @@ if (ENABLE_LAGOM_CCACHE)
 endif()
 
 if (ENABLE_FUZZERS_LIBFUZZER OR ENABLE_FUZZERS_OSSFUZZ)
-	set(ENABLE_FUZZERS ON)
+    set(ENABLE_FUZZERS ON)
```
Zaggy1024: Looks like something reformatted this file? The change is fine, but it should probably be in a separate commit.

Libraries/LibSync/MutexPosix.cpp
```diff
@@ -6,25 +6,86 @@
...
 #include <pthread.h>
 
 namespace Sync {
 
-Mutex::Mutex()
-    : m_lock_count(0)
+template<typename R, typename I>
+MutexBase<R, I>::~MutexBase()
 {
+    int result = pthread_mutex_destroy(reinterpret_cast<pthread_mutex_t*>(m_storage));
```
Zaggy1024: Could we factor out these repeated reinterpret casts to a static function?

Zaggy1024 (changes requested):

Libraries/LibThreading/BackgroundAction.cpp
R-Goc: Not sure what you mean here? This is in split out sync primitives.

Meta/Lagom/CMakeLists.txt
R-Goc: Yeah didn't notice it at all. I wonder what I was doing here. I'll just remove this change

Libraries/LibThreading/BackgroundAction.cpp
Zaggy1024: The reorder of the includes is in the commit for out-of-line implementations, maybe GitHub is displaying it wrong on your end?

R-Goc: Wait what is happening here?

Zaggy1024: I'm seeing:

```diff
  #include <AK/Queue.h>
- #include <LibThreading/BackgroundAction.h>
  #include <LibSync/Mutex.h>
+ #include <LibThreading/BackgroundAction.h>
```

R-Goc: Ahh okay. The format did not run as I did the mass edit in VS, where I guess clang-format isn't set up correctly. Then github shows discussion for a line of source in every commit where a thing is changed so I thought you were talking about the rename

Zaggy1024: Ah, yeah, if you're not in the Conversations section, it may not show the change from the specific commit that I commented on.

Libraries/LibCore/ThreadEventQueue.cpp
R-Goc: Don't we want the cross thread access coming from TLS? I'd need someone's opinion on this. The event queue and eventloop could use more documentation as to what their assumptions are.

Libraries/LibThreading/Once.h
R-Goc: Actually no, this is wrong. As set_up() can take time. So the cas will make is_set_up true before set_up() finished running which could be very bad.

R-Goc:
Unstale. I'll get to this soon-ish.

Libraries/LibSync/MutexPOSIX.cpp
R-Goc: As per #ui-windows we will leave them as cpp files to keep it cleaner. If we see it in traces we can move it to headers.

Libraries/LibThreading/Thread.cpp
```diff
@@ -5,9 +5,15 @@
  */
 
 #include <LibThreading/Thread.h>
+#include <pthread.h>
 
 namespace Threading {
 
+bool ThreadID::operator==(ThreadID const& other)
```
ADKaster: this snippet feels like it should be moved to this commit in some way: UI/Qt/EventLoopImplementationQt.cpp:

```c++
#if defined(AK_OS_WINDOWS)
namespace AK {

template<>
struct Traits<pthread_t> : public DefaultTraits<pthread_t> {
    static unsigned hash(pthread_t thread_id)
    {
        return Traits<intptr_t>::hash(reinterpret_cast<intptr_t>(thread_id.p));
    }

    static constexpr bool equals(pthread_t a, pthread_t b)
    {
        return Traits<intptr_t>::equals(reinterpret_cast<intptr_t>(a.p), reinterpret_cast<intptr_t>(b.p));
    }
};

}

#endif
```

Libraries/LibSync/ConditionVariable.h
```diff
@@ -1,63 +1,73 @@
...
+        return sizeof(void*);
+#else
+        return sizeof(pthread_cond_t);
+#endif
     }
 
-private:
-    pthread_cond_t m_condition;
-    Mutex& m_to_wait_on;
+    alignas(void*) unsigned char m_storage[get_storage_size()];
```
ADKaster:
```suggestion
#ifdef AK_OS_WINDOWS
    using StorageType = void*;
#else
    using StorageType = pthread_cond_t;
    
    alignas(StorageType) unsigned char m_storage[sizeof(StorageType)];
    MutexType& m_to_wait_on;
```

Libraries/LibThreading/Thread.cpp
R-Goc: Just dropped it for now. It need a bit more thinking and probably ifdefs to work in Core. As we would want it in the event loop implementations, so we would have a circular dependency.

Libraries/LibSync/MutexPosix.cpp
R-Goc: Changed these to an internal function where it made sense. In the mutex implementation on windows we have 3 different types so I didn't do it there.

Libraries/LibCore/ThreadEventQueue.cpp
R-Goc: After the changes made here it becomes more complicated to do this and keep the current_or_null. I'll leave this for a future PR.

Libraries/LibCore/EventLoop.cpp
```diff
@@ -186,7 +186,7 @@ StrongEventLoopReference::StrongEventLoopReference(WeakEventLoopReference& event
 
 StrongEventLoopReference::~StrongEventLoopReference()
 {
-    m_event_loop_weak->m_lock.unlock();
+    m_event_loop_weak->m_lock.unlock_read();
```
ADKaster, line 193: Is this the same operation? In the commit `LibThreading/LibSync: Split out sync primitives` I would expect t only find `s/Threading/Sync/g` changes, rather than a possible bugfix.

Libraries/LibIPC/CMakeLists.txt
```diff
@@ -18,4 +18,4 @@ else()
 endif()
 
 ladybird_lib(LibIPC ipc)
-target_link_libraries(LibIPC PRIVATE LibCore LibURL LibThreading)
+target_link_libraries(LibIPC PRIVATE LibCore LibURL LibThreading LibSync)
```
ADKaster: does this belong in the previous commit, the one that added LibSync?

Libraries/LibSync/Policy.h
```diff
@@ -0,0 +1,16 @@
...
+ */
+
+#pragma once
+
+namespace Sync {
+
+struct PolicyNonRecursive { };
+struct PolicyRecursive { };
+struct PolicyIntraprocess { };
+struct PolicyInterprocess { };
```
ADKaster, line 16: Do we actually have users for inter-process mutexes?

Libraries/LibCore/EventLoop.cpp
R-Goc: This is not a bugfix, and breaks this commit. I guess git absorb put it in the wrong commit or I otherwise messed up a fixup. This method doesn't exist yet, it was only added in implement RWLock

Libraries/LibIPC/CMakeLists.txt
R-Goc: Looks like it, I will move it.

Libraries/LibSync/Policy.h
R-Goc: Not yet. However, I did need to figure out if they would work with the design I was going for. My main thought is that if someone that is primarily a linux dev wanted to add in an interprocess mutex they would have a hard time implementing it on windows. I added the interprocess version where it exists as an OS primitive. For the ones that don't exist and would need to be written with other primitives I left them unimplemented.

Zaggy1024 (approved):

PR #6540, opened 2025-10-21
LibWeb: Ensure drawImage() always uses the first image frame

tcl3:
Previously, `drawImage()` used the current frame of animated images.


Libraries/LibWeb/HTML/HTMLImageElement.cpp
```diff
@@ -206,10 +206,10 @@ Optional<CSSPixelFraction> HTMLImageElement::intrinsic_aspect_ratio() const
     return {};
 }
 
-RefPtr<Gfx::ImmutableBitmap> HTMLImageElement::current_image_bitmap_sized(Gfx::IntSize size) const
+RefPtr<Gfx::ImmutableBitmap> HTMLImageElement::current_image_bitmap_sized(Gfx::IntSize size, Optional<size_t> frame_index) const
```
gmta: It seems to me these `current_*` methods use `m_current_frame_index` for a reason. Maybe rename this to `::image_bitmap_sized()` with a non-optional `frame_index` and make a new `::current_image_bitmap_sized()` that calls into this with `m_current_frame_index`?

tcl3: I completely missed that `default_image_bitmap_sized()` already exists :sweat_smile: . I've updated the PR to use that instead.

gmta (approved):

PR #6545, opened 2025-10-22
LibRegex: Add unicode string properties 

aplefull:
This PR adds unicode string properties from https://tc39.es/ecma262/#table-binary-unicode-properties-of-strings, and a small fix for escaping line terminators - we were producing incorrect regex.source when expression had terminators like `\n`, `\r`.

Test 262 diff:   

Summary:
        +14 ✅    -14 ❌   

Diff Tests:
    test/built-ins/RegExp/unicodeSets/generated/character-class-difference-property-of-strings-escape.js             ❌ -> ✅
    test/built-ins/RegExp/unicodeSets/generated/character-class-escape-difference-property-of-strings-escape.js      ❌ -> ✅
    test/built-ins/RegExp/unicodeSets/generated/character-class-escape-intersection-property-of-strings-escape.js    ❌ -> ✅
    test/built-ins/RegExp/unicodeSets/generated/character-class-intersection-property-of-strings-escape.js           ❌ -> ✅
    test/built-ins/RegExp/unicodeSets/generated/character-difference-property-of-strings-escape.js                   ❌ -> ✅
    test/built-ins/RegExp/unicodeSets/generated/character-intersection-property-of-strings-escape.js                 ❌ -> ✅
    test/built-ins/RegExp/unicodeSets/generated/character-property-escape-difference-property-of-strings-escape.js   ❌ -> ✅
    test/built-ins/RegExp/unicodeSets/generated/character-property-escape-intersection-property-of-strings-escape.js ❌ -> ✅
    test/built-ins/RegExp/unicodeSets/generated/property-of-strings-escape-difference-property-of-strings-escape.js  ❌ -> ✅
    test/built-ins/RegExp/unicodeSets/generated/property-of-strings-escape-intersection-character-class-escape.js    ❌ -> ✅
    test/built-ins/RegExp/unicodeSets/generated/property-of-strings-escape-intersection-character-class.js           ❌ -> ✅
    test/built-ins/RegExp/unicodeSets/generated/property-of-strings-escape-intersection-character-property-escape.js ❌ -> ✅
    test/built-ins/RegExp/unicodeSets/generated/property-of-strings-escape-intersection-character.js                 ❌ -> ✅
    test/staging/sm/RegExp/escape.js                                                                                 ❌ -> ✅

### DIFF
--- a/Libraries/LibJS/Runtime/RegExpObject.cpp
+++ b/Libraries/LibJS/Runtime/RegExpObject.cpp
@@ -275,12 +275,30 @@ String RegExpObject::escape_regexp_pattern() const
     // FIXME: Check the 'u' and 'v' flags and escape accordingly
     StringBuilder builder;
     auto escaped = false;
+    auto in_character_class = false;
 
     for (auto code_point : m_pattern) {
         if (escaped) {
             escaped = false;
             builder.append_code_point('\\');
-            builder.append_code_point(code_point);
+
+            switch (code_point) {
+            case '\n':
+                builder.append_code_point('n');
+                break;
+            case '\r':
+                builder.append_code_point('r');
+                break;
+            case LINE_SEPARATOR:
+                builder.append("u2028"sv);
+                break;
+            case PARAGRAPH_SEPARATOR:
+                builder.append("u2029"sv);
+                break;
+            default:
+                builder.append_code_point(code_point);
+                break;
+            }
             continue;
         }
 
@@ -289,9 +307,18 @@ String RegExpObject::escape_regexp_pattern() const
             continue;
         }
 
+        if (code_point == '[') {
+            in_character_class = true;
+        } else if (code_point == ']') {
+            in_character_class = false;
+        }
+
         switch (code_point) {
         case '/':
-            builder.append("\\/"sv);
+            if (in_character_class)
+                builder.append_code_point('/');
+            else
+                builder.append("\\/"sv);
             break;
         case '\n':
             builder.append("\\n"sv);
--- a/Libraries/LibJS/Tests/builtins/RegExp/RegExp.prototype.source.js
+++ b/Libraries/LibJS/Tests/builtins/RegExp/RegExp.prototype.source.js
@@ -5,3 +5,33 @@ test("basic functionality", () => {
     expect(/\n/.source).toBe("\\n");
     expect(/foo\/bar/.source).toBe("foo\\/bar");
 });
+
+test("escaped characters", () => {
+    const tests = [
+        { regex: /\\n/, source: "\\\\n" },
+        { regex: /\\\n/, source: "\\\\\\n" },
+        { regex: /\\\\n/, source: "\\\\\\\\n" },
+        { regex: /\\r/, source: "\\\\r" },
+        { regex: /\\\r/, source: "\\\\\\r" },
+        { regex: /\\\\r/, source: "\\\\\\\\r" },
+        { regex: /\\u2028/, source: "\\\\u2028" },
+        { regex: /\\\u2028/, source: "\\\\\\u2028" },
+        { regex: /\\\\u2028/, source: "\\\\\\\\u2028" },
+        { regex: /\\u2029/, source: "\\\\u2029" },
+        { regex: /\\\u2029/, source: "\\\\\\u2029" },
+        { regex: /\\\\u2029/, source: "\\\\\\\\u2029" },
+        { regex: /\//, source: "\\/" },
+        { regex: /\\\//, source: "\\\\\\/" },
+        { regex: /[/]/, source: "[/]" },
+        { regex: /[\/]/, source: "[\\/]" },
+        { regex: /[\\/]/, source: "[\\\\/]" },
+        { regex: /[\\\/]/, source: "[\\\\\\/]" },
+        { regex: /\[\/\]/, source: "\\[\\/\\]" },
+        { regex: /\[\/\]/, source: "\\[\\/\\]" },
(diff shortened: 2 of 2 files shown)


Libraries/LibRegex/RegexParser.cpp
```diff
@@ -2506,11 +2519,22 @@ bool ECMA262Parser::parse_unicode_property_escape(PropertyEscape& property, bool
...
 
     return property.visit(
-        [this](Unicode::Property property) {
+        [this, negated](Unicode::Property property) {
             if (!Unicode::is_ecma262_property(property)) {
                 set_error(Error::InvalidNameForProperty);
                 return false;
             }
+            if (Unicode::is_ecma262_string_property(property)) {
+                if (!m_parser_state.regex_options.has_flag_set(AllFlags::UnicodeSets)) {
```
gmta: You can add `|| negated` to this condition so you can drop the if-block below.

Libraries/LibJS/Runtime/RegExpObject.cpp
```diff
@@ -275,12 +275,33 @@ String RegExpObject::escape_regexp_pattern() const
...
 
     for (auto code_point : m_pattern) {
         if (escaped) {
             escaped = false;
             builder.append_code_point('\\');
-            builder.append_code_point(code_point);
+
+            switch (code_point) {
+            case '/':
+                builder.append_code_point('/');
```
gmta: This is identical to the default case right?

Libraries/LibRegex/RegexParser.cpp
aplefull: Done

Libraries/LibJS/Runtime/RegExpObject.cpp
aplefull: Ah, right, removed it

gmta (approved):

Libraries/LibRegex/RegexParser.cpp
```diff
@@ -2463,6 +2463,19 @@ bool ECMA262Parser::parse_nested_class(Vector<regex::CompareTypeAndValuePair>& c
                 compares.empend(CompareTypeAndValuePair { CharacterCompareType::Inverse, 0 });
             property.visit(
                 [&](Unicode::Property property) {
+                    if (Unicode::is_ecma262_string_property(property)) {
```
trflynn89, line 2466: This commit should contain in-tree tests in Tests/LibRegex/TestRegex.cpp and test-js

Libraries/LibUnicode/CharacterTypes.cpp
```diff
@@ -289,6 +291,11 @@ bool is_ecma262_property(Property property)
     case UCHAR_QUOTATION_MARK:
     case UCHAR_RADICAL:
     case UCHAR_REGIONAL_INDICATOR:
+    case UCHAR_RGI_EMOJI:
+    case UCHAR_RGI_EMOJI_FLAG_SEQUENCE:
+    case UCHAR_RGI_EMOJI_MODIFIER_SEQUENCE:
+    case UCHAR_RGI_EMOJI_TAG_SEQUENCE:
+    case UCHAR_RGI_EMOJI_ZWJ_SEQUENCE:
```
trflynn89: Instead of repeating all these string properties here (which looks confusing when we compare this list to the table in the spec), maybe just return `is_ecma262_string_property(property)` in the default case?

Alternatively, it looks like we could also just rearrange the caller in RegexParser.cpp a bit to avoid touching this function at all, which then lets this function continue to match the spec table exactly:
```c++
[this, negated](Unicode::Property property) {
    if (Unicode::is_ecma262_string_property(property)) {
        if (!m_parser_state.regex_options.has_flag_set(AllFlags::UnicodeSets) || negated) {
            set_error(Error::InvalidNameForProperty);
            return false;
        }
    } else if (!Unicode::is_ecma262_property(property)) {
        set_error(Error::InvalidNameForProperty);
        return false;
    }
    return true;
},
```

Libraries/LibJS/Runtime/RegExpObject.cpp
```diff
@@ -275,12 +275,30 @@ String RegExpObject::escape_regexp_pattern() const
     // FIXME: Check the 'u' and 'v' flags and escape accordingly
```
alimpfard, line 275: commit: please mention that it's about escaping _regex_ source, e.g. `LibJS: Escape line terminators in regex source` (looks like we weren't doing that before anyway)

Libraries/LibRegex/RegexParser.cpp
```diff
@@ -2463,6 +2463,19 @@ bool ECMA262Parser::parse_nested_class(Vector<regex::CompareTypeAndValuePair>& c
...
+                        if (negated) {
+                            set_error(Error::InvalidCharacterClass);
+                            return;
+                        }
+
+                        for (auto const& compare : compares) {
+                            if (compare.type == CharacterCompareType::Inverse) {
+                                set_error(Error::InvalidCharacterClass);
+                                return;
+                            }
```
alimpfard, line 2476: hmm...should we have a new error for this? it sounds like a separate class of errors to me.

alimpfard (commented):
I'm surprised this needed no change in the ops, nice!

alimpfard:
Regex bits LGTM, thanks!

aplefull:
Rebased on master to pass newly updated IndexedDB tests :)

Libraries/LibUnicode/CharacterTypes.cpp
aplefull: Rearranged it like you suggested and added test cases for the change :saluting_face:

PR #6547, opened 2025-10-22
LibWeb/WebAudio: Implement basic startRendering

Prospero23:
Creates basic `offlineAudioContext.startRendering()` and adds a passing WPT.

This PR doesn't handle actually rendering audio or creating a new thread. 

Also, `begin_offline_rendering()` uses normal `DOM::Event` for now. I am planning on submitting a second PR to implement `OfflineAudioCompletionEvent`.

### DIFF
--- a/Libraries/LibWeb/WebAudio/OfflineAudioContext.cpp
+++ b/Libraries/LibWeb/WebAudio/OfflineAudioContext.cpp
@@ -1,12 +1,17 @@
 /*
  * Copyright (c) 2024, Shannon Booth <shannon@serenityos.org>
+ * Copyright (c) 2025, Ben Eidson <b.e.eidson@gmail.com>
  *
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
 #include <LibWeb/Bindings/Intrinsics.h>
+#include <LibWeb/DOM/Event.h>
 #include <LibWeb/HTML/EventNames.h>
+#include <LibWeb/HTML/Navigable.h>
+#include <LibWeb/HTML/Scripting/TemporaryExecutionContext.h>
 #include <LibWeb/HTML/Window.h>
+#include <LibWeb/WebAudio/AudioBuffer.h>
 #include <LibWeb/WebAudio/AudioDestinationNode.h>
 #include <LibWeb/WebAudio/OfflineAudioContext.h>
 
@@ -23,7 +28,7 @@ WebIDL::ExceptionOr<GC::Ref<OfflineAudioContext>> OfflineAudioContext::construct
     TRY(verify_audio_options_inside_nominal_range(realm, context_options.number_of_channels, context_options.length, context_options.sample_rate));
 
     // Let c be a new OfflineAudioContext object. Initialize c as follows:
-    auto c = realm.create<OfflineAudioContext>(realm, context_options.length, context_options.sample_rate);
+    auto c = realm.create<OfflineAudioContext>(realm, context_options.number_of_channels, context_options.length, context_options.sample_rate);
 
     // 1. Set the [[control thread state]] for c to "suspended".
     c->set_control_state(Bindings::AudioContextState::Suspended);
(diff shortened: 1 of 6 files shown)


SephReed:
@Prospero23 would you like some help working on the web audio support?

Libraries/LibWeb/WebAudio/OfflineAudioContext.cpp
```diff
@@ -58,7 +63,82 @@ OfflineAudioContext::~OfflineAudioContext() = default;
 // https://webaudio.github.io/web-audio-api/#dom-offlineaudiocontext-startrendering
 WebIDL::ExceptionOr<GC::Ref<WebIDL::Promise>> OfflineAudioContext::start_rendering()
 {
-    return WebIDL::NotSupportedError::create(realm(), "FIXME: Implement OfflineAudioContext::start_rendering"_utf16);
+    auto& realm = this->realm();
+    auto& window = as<HTML::Window>(HTML::relevant_global_object(*this));
```
gmta: It's a bit clearer if you move this line to after the spec step 1 comment, where the relevant global object is actually mentioned.

Libraries/LibWeb/WebAudio/OfflineAudioContext.cpp
```diff
@@ -58,7 +63,82 @@ OfflineAudioContext::~OfflineAudioContext() = default;
...
+
+    // 1. If this’s relevant global object’s associated Document is not fully active then return a promise rejected with "InvalidStateError" DOMException.
+    auto const& associated_document = window.associated_document();
+    if (!associated_document.is_fully_active()) {
+        auto error = WebIDL::InvalidStateError::create(realm, "Document is not fully active"_utf16);
+        return WebIDL::create_rejected_promise_from_exception(realm, error);
+    }
+
+    // NOTE: Not in spec explicitly, but this should account for detached iframes too. See /the-offlineaudiocontext-interface/startrendering-after-discard.html WPT.
```
gmta: We use `// AD-HOC: ..` comments when diverging from the spec.

Libraries/LibWeb/WebAudio/OfflineAudioContext.cpp
```diff
@@ -58,7 +63,82 @@ OfflineAudioContext::~OfflineAudioContext() = default;
...
+
+    // 3. Set the [[rendering started]] slot of the OfflineAudioContext to true.
+    m_rendering_started = true;
+
+    // 4. Let promise be a new promise.
+    auto promise = WebIDL::create_promise(realm);
+
+    // 5. Create a new AudioBuffer, with a number of channels, length and sample rate equal respectively to the
+    // numberOfChannels, length and sampleRate values passed to this instance’s constructor in the contextOptions parameter.
+    //  Assign this buffer to an internal slot [[rendered buffer]] in the OfflineAudioContext.
```
gmta: Please indent these wrapped lines, i.e.:

```cpp

    // 5. Create a new AudioBuffer, with a number of channels, length and sample rate equal respectively to the
    //    numberOfChannels, length and sampleRate values passed to this instance’s constructor in the contextOptions
    //    parameter.
```

Libraries/LibWeb/WebAudio/OfflineAudioContext.cpp
```diff
@@ -58,7 +63,82 @@ OfflineAudioContext::~OfflineAudioContext() = default;
...
+    auto buffer_result = this->create_buffer(this->m_number_of_channels, this->length(), this->sample_rate());
```
gmta: You don't need to do `this->` here; that's only necessary to prevent name conflicts with local variables.

Libraries/LibWeb/WebAudio/OfflineAudioContext.cpp
gmta: Please move the "// Assign this buffer ..." comment down to where you actually assign something to `m_rendered_buffer`. Optionally you can mention that you've shuffled the spec step around a bit.

Libraries/LibWeb/WebAudio/OfflineAudioContext.cpp
```diff
@@ -58,7 +63,82 @@ OfflineAudioContext::~OfflineAudioContext() = default;
...
+    // 6. If an exception was thrown during the preceding AudioBuffer constructor call, reject promise with this exception.
+    if (buffer_result.is_exception()) {
+        return WebIDL::create_rejected_promise_from_exception(realm, buffer_result.exception());
+    }
+
+    m_rendered_buffer = buffer_result.release_value();
+
+    // 8. Append promise to [[pending promises]].
+    // NOTE: Placed before step seven to make sure promise is still pending.
+    m_pending_promises.append(promise);
```
gmta: Is this the right thing to do? Nowhere does it say you need to add this promise to [[pending promises]] before "begin offline rendering". And in that algorithm, it mentions "the promise created by startRendering()", and not "[[pending promises]]".

Prospero23:
@SephReed most definitely. 

Right now, I am really trying to just get the audio rendering happening enough to make it easy to work on all of the different nodes. A next step for that is maybe reviving some version of  https://github.com/LadybirdBrowser/ladybird/pull/5535. 

The problem I ran into when I was trying to work on nodes before this step is finished is a good number of the WPT that are helpful for testing rely on offline audio rendering stuff to be implemented to not time out.

Libraries/LibWeb/WebAudio/OfflineAudioContext.cpp
Prospero23: I am not sure. That was the one part of the spec I was/am struggling with.

 If we append after, do we run the risk of appending a non-pending promise? Not sure that even matters, but it seems like it would. 

In my head, "the promise created by startRendering()" and "[[pending promises]]" are one in the same. At this point, the `startRendering()` promise is the only one that could have been added. 

The spec mentions resolving the promise but nothing about removing it from "[[pending promises]]". It seems like maybe this is an error? I don't see why resolved promises should still be in that array.

Prospero23: I would be happy to make an issue on the spec, but it may just be a lack of understanding on my part.

Libraries/LibWeb/WebAudio/OfflineAudioContext.cpp
```diff
@@ -58,7 +63,83 @@ OfflineAudioContext::~OfflineAudioContext() = default;
...
+    //    numberOfChannels, length and sampleRate values passed to this instance’s constructor in the contextOptions
+    //    parameter.
+    auto buffer_result = create_buffer(m_number_of_channels, length(), sample_rate());
+
+    // 6. If an exception was thrown during the preceding AudioBuffer constructor call, reject promise with this exception.
+    if (buffer_result.is_exception()) {
+        return WebIDL::create_rejected_promise_from_exception(realm, buffer_result.exception());
+    }
+
+    //  Assign this buffer to an internal slot [[rendered buffer]] in the OfflineAudioContext.
```
gmta: One space too many at the beginning

Libraries/LibWeb/WebAudio/OfflineAudioContext.cpp
gmta: I've been reading through the spec and I agree with you, they're very explicit on removing promises from `[[pending promises]]` when they get rejected, but removing them when they get resolved seems to be implied.

I think you should raise an upstream issue for this so you can link to it in an `// AD-HOC: ..` comment. I'd try to keep as close to the original spec text and order as possible, but with the addition of removing the promise from the `[[pending promises]]` when you resolve it.

Note that the spec passes the created `promise` along to the next algorithm, so they're definitely not iterating over all `[[pending promises]]`!

SephReed:
@Prospero23 I'm `seph_yo` on Discord.  Please message me so we can synchronize.

In case it matters, I once built a VST using JUCE, and did a couple DSP projects in college.  Also, I produce (Bitwig main).

Prospero23:
I think everything is fixed. Happy to revise anything else :)

PR #6552, opened 2025-10-23
LibWeb: CSS selector read-write honor `is_allowed_to_be_readonly`

lpas:


Libraries/LibWeb/CSS/SelectorEngine.cpp
```diff
@@ -462,12 +462,9 @@ static bool matches_read_write_pseudo_class(DOM::Element const& element)
...
     if (is<HTML::HTMLInputElement>(element)) {
-        auto& input_element = static_cast<HTML::HTMLInputElement const&>(element);
-        if (input_element.has_attribute(HTML::AttributeNames::readonly))
-            return false;
-        if (!input_element.enabled())
-            return false;
-        return true;
+        auto const& input_element = static_cast<HTML::HTMLInputElement const&>(element);
+        return input_element.is_allowed_to_be_readonly()
+            && (!input_element.has_attribute(HTML::AttributeNames::readonly) && input_element.enabled());
```
AtkinsSJ: A couple of nits:
- While you're here, could you change this to use `as_if<HTML::HTMLInputElement>` instead of `is` and `static_cast`? eg, `if (auto* input_element = as_if<HTML::HTMLInputElement>(element)) { ... }`
- You don't need the extra `()` in the return statement, and I think it'd be clearer without them.

AtkinsSJ (approved):
Looks good. Thanks!

AtkinsSJ:
...Ah, you need to rebase on master because those IndexedDB tests got updated.

PR #6557, opened 2025-10-23
Everywhere: Avoid passing parameters by value where unnecessary

tcl3:
This PR applies `performance-unnecessary-value-param` fixes found by `clang-tidy` (with some manual intervention)

The command I ran to find these was: ` run-clang-tidy -p Build/release -checks='-*,performance-unnecessary-value-param' -fix`


Libraries/LibCrypto/PK/RSA.h
```diff
@@ -208,7 +208,7 @@ class RSA_EME : public RSA {
 public:
     template<typename... Args>
     RSA_EME(Hash::HashKind hash_kind, Args... args)
-        : RSA(args...)
+        : RSA(move(args)...)
```
trflynn89: We should be perfect-forwarding variadic template args:
```c++
template<typename... Args>
RSA_EME(Hash::HashKind hash_kind, Args&&... args)
    : RSA(forward<Args>(args)...)
```
(note the added `&&` as well)

Libraries/LibJS/Runtime/Error.cpp
```diff
@@ -39,7 +39,7 @@ GC::Ref<Error> Error::create(Realm& realm)
     return realm.create<Error>(realm.intrinsics().error_prototype());
 }
 
-GC::Ref<Error> Error::create(Realm& realm, Utf16String message)
+GC::Ref<Error> Error::create(Realm& realm, Utf16String const& message)
```
trflynn89, line 42: False positive? We've `move`ing both strings changed here

trflynn89: Oh huh `PrimitiveString` takes strings by `const&` now I guess. So just a nit: we'll now get clangd warnings above moving a constant variable:
https://clang.llvm.org/extra/clang-tidy/checks/performance/move-const-arg.html

tcl3: I got rid of the unnecessary `move()` in this case. I tried applying all the `clang-tidy`  `move-const-arg` suggestions but `clang-tidy` is a bit overzealous with these.

Libraries/LibWasm/AbstractMachine/AbstractMachine.h
```diff
@@ -672,7 +672,7 @@ class Label {
 
 class Frame {
 public:
-    explicit Frame(ModuleInstance const& module, Vector<Value> locals, Expression const& expression, size_t arity)
+    explicit Frame(ModuleInstance const& module, Vector<Value> const& locals, Expression const& expression, size_t arity)
```
alimpfard: This makes the move useless, and adds an unnecessary copy.

Libraries/LibWeb/Crypto/CryptoAlgorithms.cpp
```diff
@@ -7860,7 +7860,7 @@ static WebIDL::ExceptionOr<ByteBuffer> hmac_calculate_message_digest(JS::Realm&
     return hmac.process(message);
 }
 
-static WebIDL::ExceptionOr<WebIDL::UnsignedLong> hmac_hash_block_size(JS::Realm& realm, HashAlgorithmIdentifier hash)
+static WebIDL::ExceptionOr<WebIDL::UnsignedLong> hmac_hash_block_size(JS::Realm& realm, HashAlgorithmIdentifier const& hash)
```
alimpfard, line 7863: This is passing an enum by reference, not a fan.

Libraries/LibWeb/WebAssembly/WebAssembly.cpp
```diff
@@ -134,7 +134,7 @@ WebIDL::ExceptionOr<GC::Ref<WebIDL::Promise>> compile(JS::VM& vm, GC::Root<WebID
 }
 
 // https://webassembly.github.io/spec/web-api/index.html#dom-webassembly-compilestreaming
-WebIDL::ExceptionOr<GC::Ref<WebIDL::Promise>> compile_streaming(JS::VM& vm, GC::Root<WebIDL::Promise> source)
+WebIDL::ExceptionOr<GC::Ref<WebIDL::Promise>> compile_streaming(JS::VM& vm, GC::Root<WebIDL::Promise> const& source)
```
alimpfard: GC::Root is essentially a pointer, not sure why you'd want to pass it as const&

trflynn89: Root is a bit more than a pointer, as opposed to GC::Ref/GC::Ptr that are free to copy.

Libraries/LibWeb/Crypto/CryptoAlgorithms.cpp
trflynn89: `HashAlgorithmIdentifier` is a `Variant<GC::Root<JS::Object>, String>`

Libraries/LibWeb/WebAssembly/WebAssembly.cpp
alimpfard: Ah, my bad then.

Libraries/LibWeb/Crypto/CryptoAlgorithms.cpp
alimpfard: Oh...that needs a better name 😅

Libraries/LibJS/AST.h
```diff
@@ -49,7 +49,7 @@ class SharedFunctionInstanceData;
 
 template<class T, class... Args>
 static inline NonnullRefPtr<T>
-create_ast_node(SourceRange range, Args&&... args)
+create_ast_node(SourceRange const& range, Args&&... args)
 {
     return adopt_ref(*new T(move(range), forward<Args>(args)...));
```
Hendiadyoin1: To this file in genral
it seem to take the source range by `const&` and then tries to move it?

Libraries/LibWeb/WebAssembly/WebAssembly.cpp
tcl3: I undid the changes to all `GC::Root`s before I saw the follow-up. I'm not sure if there's much of a performance gain passing these by const reference anyway.

trflynn89: It'd be ref-count churn, same as String and friends

PR #6562, opened 2025-10-23
Documentation: Update EventLoop.md

R-Goc:
This updates EventLoop.md with the results of the discussion had on discord regarding the expected behavior when registering or unregistering objects with the event loop from another thread.

For context #6494 originally added mutexes to the register/unregister functions as it was attempting to have the main thread register notifiers for the io thread. The mutex was inside the ThreadData Struct. However, acquiring the mutex inside the ThreadData wouldn't prevent the thread that owns the structure terminating and ThreadData destructing, between acquiring the ThreadData's pointer and taking the mutex. This would lead to a use after free.

This could be solved by acquiring a global read lock for that stretch of time but this has performance implications. More importantly the ownership of resources associated with an event loop becomes difficult to reason about.

As such it was decided not to allow other threads to register/unregister objects with the event loop.

Pull request #6494: https://github.com/LadybirdBrowser/ladybird/pull/6494
Discord message link: https://discord.com/channels/1247070541085671459/1306918361732616212/1430978833447915641


Zaggy1024:
Could we make the commit summary a bit more specific?

R-Goc:
> Could we make the commit summary a bit more specific?

specific as in shorter or more detailed?

kalenikaliaksandr (changes requested):
currently we do unregister notifiers, timers and signals from threads other than the one where object was constructed. wouldn't it be more reasonable to add this doc line alongs with the fixes that implement this suggestion?

R-Goc:
Where do we do that? I was under the impression your PR added that behavior.

kalenikaliaksandr:
> Where do we do that? I was under the impression your PR added that behavior.

our current Notifier implementation allows that. my IO thread PR only highlighted the problem by doing it in a hot code path.

Issue #302, opened 2024-06-28
Site isolation


awesomekling:
We should implement site isolation, where each site (scheme + domain + port) gets its own WebContent process, and code from each origin only runs in that process.

When one site embeds an iframe from another site, we'll have two WebContent processes who will have to coordinate communication, layout, rendering, input, etc.

Here's how they've done it in Chromium:
https://www.chromium.org/Home/chromium-security/site-isolation/

DemiMarie:
A related issue is site sandboxing, which is ensuring that a compromised renderer process cannot access or tamper with data for origins it isn’t responsible for.

That’s arguably a separate issue (for which this one is a prerequisite), but I’m mentioning it here because site isolation without site sandboxing only protects against information leaks, not code execution.

manuel-za:
#3882 implements "simple" site isolation for "sites" (TLD + one level, as in the chromium description). Nice!
Since this is done in `Navigable::navigate`, it should also apply to iframes?

Would be good to know whether this issue is still considered open, incl. remaining scope. 
- #3882 only mentions one limitation: redirects.
- #3958 seems like a regression issue

kalenikaliaksandr:
> Since this is done in Navigable::navigate, it should also apply to iframes?

nope, because currently all iframes live within one webcontent process.

DemiMarie:
> Would be good to know whether this issue is still considered open, incl. remaining scope.

From a security perspective, yes.  Full site isolation requires that all cross-origin iframes be in separate processes.

Issue #292, opened 2024-06-26
Meta: Investigate Seccomp profiles on Linux


ADKaster:
We should look into how other browsers implement seccomp and other syscall-sandboxing measures for their different process classes.

We should be able to abstract this in a way that's similar to how SerenityOS/OpenBSD's pledge works, to not go crazy with massive syscall list files.

Ideally we would be able to sandbox the WebContent/WebWorker, RequestServer, ImageDecoder, and any future GPU process in a way that they have privileges limited. Locking down at least the ability to create new IPC sockets and requiring those to come from the UI process would be a good first step.

We should also look into similar features for other operating systems.

colleirose:
I spent some time looking into this for Windows and made some very bad and untested code for Windows that I hope to fix eventually. ~~The overall idea is roughly based on the open-source BSD-3 Chromium sandboxing code; I don't know if including Chromium sandboxing code would violate the principals of being an independent browser or if BSD-3 licensed code can be used, although I did make sure to rework a lot of it and it isn't directly adding anything from Chromium as a dependency.~~ (I removed most of the Chromium code by now. I don't have as much free time lately so I haven't gone back to working on this but I will soon.) More importantly, working on that led me to do more research about sandboxing and mitigation policies.

Basically, the following sandboxing & hardening features are available on Windows:
- [Process mitigation policies](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-setprocessmitigationpolicy) to restrict access to some potentially dangerous features and configure things like ASLR, DEP, and control flow guard.
- [AppContainer](https://learn.microsoft.com/en-us/windows/win32/secauthz/appcontainer-isolation) as a direct sandboxing API
- [Creating separate desktop and window station objects](https://chromium.googlesource.com/chromium/src/+/HEAD/docs/design/sandbox.md#the-alternate-desktop) to prevent the [Windows "shatter attacks"](https://en.wikipedia.org/wiki/Shatter_attack).
- Creating processes with [a low-privileged token](https://chromium.googlesource.com/chromium/src/+/HEAD/docs/design/sandbox.md#the-token) that has [more restrictive access control and integrity policies](https://chromium.googlesource.com/chromium/src/+/HEAD/docs/design/sandbox.md#the-integrity-levels) to reduce process capabilities. (This would require stronger process isolation as some things in Ladybird require more privileges than others.)
[...]

PR #6564, opened 2025-10-23
LibCrypto+AK: Merge LibCrypto/SecureRandom into AK/Random

colleirose:
`AK/Random` does the same thing as SecureRandom but with worse error handling.  It can't be removed because it's needed in `AK/`, and OpenSSL's `RAND_bytes` ideally shouldn't be included in `AK/`. So I've kept `AK/Random`'s approach of using the operating system CSPRNG directly, but improved the error handling, and removed `SecureRandom`.

This also guarantees that `BlockAllocator::allocate_block` in BlockAllocator.cpp will always either use a CSPRNG for allocation or throw an error, rather than falling back to the insecure RNG in the event of an error.


colleirose:
Fixed the lint issues

AK/Random.cpp
gmta, line 1: Commit message: please auto-wrap your commit messages to 72 characters. An exception to this is when you put a URL on its own line. Not sure why our commit linter did not pick up on this 🤔

AK/Random.cpp
```diff
@@ -1,68 +1,98 @@
...
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
 #include <AK/Platform.h>
 #include <AK/Random.h>
+#include <AK/Assertions.h>
 #include <AK/UFixedBigInt.h>
 #include <AK/UFixedBigIntDivision.h>
 
+#include <openssl/rand.h>
```
gmta: Let's not have OpenSSL as a dependency to AK, and drop RAND_bytes() in favor of a static_assert. I'd be interested to see if there are platforms out there where there's no CSPRNG API available as a sensible default.

AK/Random.cpp
```diff
@@ -1,68 +1,98 @@
...
+#    include <windows.h>
 #    include <AK/NumericLimits.h>
 #    include <AK/Windows.h>
-#    include <bcrypt.h>
-#    include <ntstatus.h>
 #endif
 
-namespace AK {
+#if defined(AK_OS_LINUX)
+static inline int csprng_chunk(void * const buf, const size_t size)
```
gmta: We use east-const everywhere.

gmta: This method is only used in one spot; consider just putting the code there instead of having a separate method.

If you want to keep it as a separate method: using `ErrorOr<void>` would be the idiomatic way of returning a potential error. Let's try to not expose any OS-specific error codes if we've dealt with them and work with our own error system as soon as possible.

AK/Random.cpp
```diff
@@ -1,68 +1,98 @@
...
-    static constexpr size_t getentropy_length_limit = 256;
-    auto iterations = bytes.size() / getentropy_length_limit;
-
-    for (size_t i = 0; i < iterations; ++i) {
-        if (getentropy(bytes.data(), getentropy_length_limit) != 0) {
-            fill_with_random_fallback();
-            return;
+    arc4random_buf(buf, size);
+#elif defined(AK_OS_LINUX)
+    size_t chunk_size = 256U;
```
gmta: Please put a `static constexpr size_t chunk_size = 256u;` at the top and reuse that wherever applicable.

AK/Random.cpp
```diff
@@ -1,68 +1,98 @@
...
+
+    do {
+        if (size < chunk_size) {
+            chunk_size = size;
+            VERIFY(chunk_size > (size_t) 0U);
```
gmta: If you change the `do {` into a `while (size > 0u) {`, you can drop this superfluous `VERIFY`.

AK/Random.cpp
```diff
@@ -1,68 +1,98 @@
...
 
-// NOTE: This function is supposed to always give a random number. If possible it is of good quality, but it can fall
-//       back to rand() if it fails on some systems. For high speed you should probably use a different generator.
-//       See MathObject::random() from LibJS. Where cryptographic security is needed use LibCrypto/SecureRandom.h.
-void fill_with_random([[maybe_unused]] Bytes bytes)
+    return ret;
+}
+#endif
+
+static int csprng(void * const buf, size_t size)
```
gmta: East-constness please

gmta: Use `ErrorOr<void>` instead of exposing OS-specific error codes as an `int`.

AK/Random.cpp
```diff
@@ -1,68 +1,98 @@
...
-    if (result == STATUS_SUCCESS)
-        return;
-#    endif
+namespace AK {
 
-    fill_with_random_fallback();
-#endif
+void fill_with_random(Bytes bytes)
+{
+    auto const size = static_cast<size_t>(bytes.size());
```
gmta: `bytes.size()` is already a `size_t`?

AK/Random.cpp
```diff
@@ -1,68 +1,98 @@
...
-namespace AK {
+#if defined(AK_OS_LINUX)
+static inline int csprng_chunk(void * const buf, const size_t size)
+{
+    int ret;
+
+    VERIFY(size <= 256U);
+    do {
+        ret = getrandom(buf, size, 0);
+    } while (ret == -1 && (errno == EINTR || errno == EAGAIN));
```
gmta: Can we even get EAGAIN if we don't pass GRND_NONBLOCK as a flag?

gmta:
@colleirose CI is failing; please make sure everything compiles locally.

AK/Random.cpp
R-Goc: Code format is screaming at it. Format the whole thing with the correct clang-format version

AK/Random.cpp
```diff
@@ -1,68 +1,98 @@
...
+    //
+    // Although the documentation states that it "may be altered or unavailable in future versions",
+    // this has never happened and probably won't happen, and it's still used in Firefox, libsodium, and other applications.
+    //
+    // The main reason that some cryptography libraries (RustCrypto, BoringSSL, etc)
+    // have stopped using this function is that it's not available in UWP apps,
+    // but UWP isn't a build target for us, and even if it is some day, this would be very simple to change.
+    #define RtlGenRandom SystemFunction036
+    extern "C" BOOLEAN NTAPI RtlGenRandom(PVOID RandomBuffer, ULONG RandomBufferLength);
+    #pragma comment(lib, "advapi32.lib")
```
R-Goc: We avoid linking in cpp files with pragmas. This should be in AK CMakeLists

R-Goc: As this is an unstable api we might want to use GetProcAddress and pull it in from a dynamic lib at runtime. We currently do that for Nt functions inside of my event loop PR, but I am not opposed to moving the mechanism for it to a header.

AK/Random.cpp
```diff
@@ -1,68 +1,98 @@
...
+    // This is the simplest CSPRNG API in Windows, and all the other ones are just a wrapper around it.
+    //
+    // Although the documentation states that it "may be altered or unavailable in future versions",
+    // this has never happened and probably won't happen, and it's still used in Firefox, libsodium, and other applications.
+    //
+    // The main reason that some cryptography libraries (RustCrypto, BoringSSL, etc)
+    // have stopped using this function is that it's not available in UWP apps,
+    // but UWP isn't a build target for us, and even if it is some day, this would be very simple to change.
+    #define RtlGenRandom SystemFunction036
+    extern "C" BOOLEAN NTAPI RtlGenRandom(PVOID RandomBuffer, ULONG RandomBufferLength);
```
R-Goc: Are you sure this will work? None of these expand to dllimport. Also use WINAPI instead of NTAPI since this is not an NT function.

R-Goc:
Why is the windows workflow not running?

gmta:
> Why is the windows workflow not running?

Because CI does not run automatically for new contributors. I'm waiting with approval until the compilation errors on Linux have been fixed.

AK/Random.cpp
```diff
@@ -1,68 +1,98 @@
...
 #include <AK/UFixedBigIntDivision.h>
 
+#include <openssl/rand.h>
+
+#if defined(AK_OS_LINUX)
+#   include <sys/random.h>
+#endif
+
 #if defined(AK_OS_WINDOWS)
+#    include <windows.h>
```
R-Goc: This is an unnecessary include. AK/Windows.h already includes windows.h

AK/Random.cpp
R-Goc: Also extern declarations with language specification can only appear at namespace scope. This code doesn't compile because of this.

colleirose:
> Because CI does not run automatically for new contributors. I'm waiting with approval until the compilation errors on Linux have been fixed.

I didn't experience compilation errors on Linux but I will apply the requested changes once I'm free and check again that it all compiles correctly. I will also test in a Windows VM to make sure that the code works correctly there too, I probably should have thought of that earlier.

AK/Random.cpp
colleirose: Do you mean the commit description? `LibCrypto+AK: Merge LibCrypto/SecureRandom into AK/Random` is 57 characters. I would assume that the commit description wouldn't have a firm length limit.

R-Goc: It does, at 72 characters. Your commit editor should've picked it up, not sure why it didn't.

colleirose: OK, got it. I'll push the fixes in a bit once I have enough time to finish writing and testing them.

colleirose: This has been fixed

AK/Random.cpp
colleirose: Fixed

AK/Random.cpp
colleirose: Fixed, I had installed the wrong clang-format version by mistake.

colleirose: Fixed

AK/Random.cpp
colleirose: Done

AK/Random.cpp
colleirose: Fixed

AK/Random.cpp
colleirose: Fixed

colleirose: Done

AK/Random.cpp
colleirose: I don't think so, I adjusted it for that.

AK/Random.cpp
colleirose: Fixed

AK/Random.cpp
colleirose: Fixed by using the BCryptGenRandom again. I'll look into using RtlGenRandom again in the future but I don't have a Windows computer to test with (and my computer is slow so using a VM is difficult) and I was told on Discord at https://discord.com/channels/1247070541085671459/1247090705013280769/1431384527699185726 "If you are somewhat confident that things should work and you are just looking for compilation failures as opposed to runtime failures,  you could also ask for a Windows label on the PR and let CI verify it."

AK/Random.cpp
colleirose: Fixed, see above

AK/Random.cpp
colleirose: Fixed

rgret-dev:
I'm not a huge fan of this change for the following reasons:

1. We lose the explicitness in code (e.g. `fill_with_random` vs `fill_with_secure_random`)
2. CSPRNGs are fast, but still more complex and generally slower than PRNGs

I'm not a maintainer, but I think we should keep a dedicated CSPRNG in LibCrypto and potentially remove AK/Random until a high-performance PRNG is actually needed. Unless I missed something, the conversation with Jelle on discord also says something similar:

> You'd have two different classes for two different purposes, regardless of their current implementation

AK/Random.cpp
```diff
@@ -1,68 +1,92 @@
...
 
-    if (bytes.is_empty() || getentropy(bytes.data(), bytes.size()) == 0)
-        return;
-#    elif defined(AK_OS_WINDOWS)
-
-    if (bytes.size() > NumericLimits<u32>::max()) [[unlikely]] {
-        fill_with_random_fallback();
-        return;
-    }
+    NTSTATUS result = ::BCryptGenRandom(NULL, buf, size, BCRYPT_USE_SYSTEM_PREFERRED_RNG);
```
R-Goc: By convention an NTSTATUS value is called status. Instead of using BRCYPT_USE_SYSTEM_PREFERRED_RNG use the algorithm pseudo handle instead.

AK/Random.cpp
```diff
@@ -1,68 +1,92 @@
...
+    if (result != STATUS_SUCCESS) [[unlikely]]
+        return Error::from_windows_error();
```
R-Goc: This under the hood calls GetLastError which will return 0 since there is no winapi error. That will cause a verify in Error. You should pass it a dos error, which you can obtain by calling RtlNtStatusToDosError on the status returned.

AK/Random.cpp
```diff
@@ -1,68 +1,92 @@
...
-    if (bytes.is_empty() || getentropy(bytes.data(), bytes.size()) == 0)
-        return;
-#    elif defined(AK_OS_WINDOWS)
-
-    if (bytes.size() > NumericLimits<u32>::max()) [[unlikely]] {
-        fill_with_random_fallback();
-        return;
-    }
+    NTSTATUS result = ::BCryptGenRandom(NULL, buf, size, BCRYPT_USE_SYSTEM_PREFERRED_RNG);
+    if (result != STATUS_SUCCESS) [[unlikely]]
```
R-Goc: This should be using the NT_SUCCESS macro. There is more than 1 success values for NT_STATUS. See: https://learn.microsoft.com/en-us/windows-hardware/drivers/kernel/using-ntstatus-values

R-Goc (changes requested):

R-Goc:
> I'm not a huge fan of this change for the following reasons:
> 
> 1. We lose the explicitness in code (e.g. `fill_with_random` vs `fill_with_secure_random`)
> 2. CSPRNGs are fast, but still more complex and generally slower than PRNGs
> 
> I'm not a maintainer, but I think we should keep a dedicated CSPRNG in LibCrypto and potentially remove AK/Random until a high-performance PRNG is actually needed.

Indeed we do lose the more explicit naming, however the quality shouldn't go down? I can read into whatever openssl is doing and see. 

We don't need this to be the fastest thing on the planet. I was already able to get gigabytes per second of throughput with the windows implementation, I believe Linux shouldn't be far behind if at all. Writing a fast PRNG is also not trivial and usually needs intrinsics. 

We cannot remove AK/Random as there are things that cannot depend on LibCrypto and need a random function.

colleirose:
I explained in the PR and on Discord why I think this is a good change, but as to the point that we lose explicitness, yes, but that doesn't really matter. In fact, part of the point here is to remove the need to make the distinction, to essentially completely eliminate the possibility of using insecure randomness. I've seen this done in other projects that need cryptographic randomness, to just have one RNG function that always uses the CSPRNG, therefore completely eliminating the possibility of accidentally using insecure randomness when secure randomness is needed.

We also are not really losing explicitness given that, as I explained, AK/Random just uses a CSPRNG but with worse error handling. It just creates ambiguity when ultimately the two functions will usually do the same thing, although AK/Random lacks the same correctness checks as SecureRandom. There isn't any practical performance benefit to removing error checking, as checking for an error is probably just a compare and jump-if instruction. If two instructions are going to make the RNG unacceptably slow then it's already not fast enough and probably can't be fast enough.

I spent a while looking into ways to make a faster non-CSPRNG but the only thing I could really think of was getting the time and/or using rand() and possibly feeding it through XorShift, but I am not convinced these are faster given that all modern operating systems have very fast sources of randomness (TPM-provided TRNG, hard disk seek times, input from USB devices and similar peripherals, CPU timing jitter, RDRAND on x86, RNDR or whatever it's called on ARM, etc.) The developer of WireGuard also made many contributions to the Linux CSPRNG system in the past few years further improving both the performance and security of it.

Also, I checked the source code for Firefox and Chromium to see if they have a faster non-CSPRNG, but they both only use the system CSPRNG as a randomness source. (Firefox appears to also have functions to collect entropy from user input and other sources, but I didn't look into it further. At the very least it is unnecessary because the kernel/OS already does that, but also, not nearly as useful given that it might not be able to collect input from all connected peripherals to all applications, given the existence of sandboxing measures on some systems, and could possibly require running as root/administrator which would create a lot of unnecessary attack surface.)

If there is a system that can't quickly produce entropy with any of these sources and/or is slow at applying the mixing algorithm (ChaCha on Linux, AES on Windows) then it would almost certainly be too slow to run a web browser. In fact, such a system would likely be too slow to run any modern operating system to begin with.

And, as I explained in the PR, OpenSSL just uses the system CSPRNG and feeds it into a cipher, and while they used to have other entropy sources, they were also either removed or disabled due to being insecure and/or breaking other things. So, this gets the same amount of entropy as OpenSSL, but with less overhead, and therefore it would improve performance.

> I think we should keep a dedicated CSPRNG in LibCrypto and potentially remove AK/Random until a high-performance PRNG is actually needed.

I was going to do this until I realized that the RNG function is used in one of the AK/ files, and I don't believe that the libraries can be used in AK, as I found doing so required some changes to the build system and also isn't done anywhere else. I didn't see any point in arguing about or investigating the norms/convention so I just kept AK/Random.

Of course, we could rename AK/Random to AK/SecureRandom and so on, but I just feel that it would be unnecessary.

> You'd have two different classes for two different purposes, regardless of their current implementation

I don't really understand how this could apply to what I've submitted unless there's other RNG classes besides AK/Random and LibCrypto/SecureRandom. Removing SecureRandom and making AK/Random always use CSPRNG values specifically solves the problem of having two classes that do the same thing.

The discussion on Discord was much longer and received more opposition than I expected before it seemed to reach the conclusion that there's nothing wrong with the change but I'd probably have to submit a PR for it myself as it's just a change to slightly tidy up the code, and so I did submit a PR. Much of the original argument seemed to be from a misunderstanding of either what I wanted to change and/or what AK/Random does, and I apologize for that as I am not always great at communicating my intentions. 

Ultimately this is just a small change to keep the code tidy and you seem to misunderstand the purpose and practical effects of it.

AK/Random.cpp
colleirose: Thanks, I'll fix these issues once I have some time.

rgret-dev:
I don't find the loss of explicitness very tidy, that's all. If someone is skimming the code, it's much easier to see that it's using a dedicated CSPRNG rather than having to dig into the implementation. I also find the ifdefs hard to read, but I understand there will be some of that if we want to use OS APIs directly.

> We cannot remove AK/Random as there are things that cannot depend on LibCrypto

What things use random but cannot depend on LibCrypto? Just `IDAllocator` in AK?

colleirose:
> I don't find the loss of explicitness very tidy, that's all. 

It returns a random number that's suitable for all use cases. I just don't see there being an issue. This completely eliminates the need for a developer to decide which to use and therefore prevents someone from using insecure randomness when security is needed. [Usually it's a good idea to provide high-level abstractions for cryptographic features so that developers don't need to worry about making mistakes.](https://soatok.blog/2021/02/24/cryptography-interface-design-is-a-security-concern/) Someone appears to have already made this type of mistake, as the insecure RNG is being used for allocation in `BlockAllocator::allocate_block` to "reduce predictability".

The Go programming language [also did something similar](https://github.com/C2SP/C2SP/blob/main/chacha8rand.md), where they made their standard randomness function cryptographically secure.

> If someone is skimming the code, it's much easier to see that it's using a dedicated CSPRNG rather than having to dig into the implementation.

And yet the current implementation is confusing because we see that there are two random functions, and when we check the functions, they do the same thing. If you seriously think this is a problem, I can make a SecureRandom and InsecureRandom file, but I don't know what you hope to change in the InsecureRandom file. Or I can just name AK/Random to AK/SecureRandom and so on, but this would be confusing when tests and the like need to use the RNG because secure randomness wouldn't be needed there. Your original request was to remove AK/Random but that cannot be done.

The best I can think of would be to name the randomness function something like `crypto_get_random` or whatever, but I still don't know if that's necessary. If someone actually needs to understand what the RNG does for security or whatever other reasons, they should just read the file. I don't know why anyone would be writing something that needs security and not bother to investigate what the function they're using actually does.

I can, of course, look into making a fast non-cryptographic PRNG for something like Math.random(), but this is just another thing that can break and has to be maintained and I really don't think it could improve performance. The best I could think of is creating a randomness pool at application launch using the SecureRandom function and then feeding it through XorShift and reseeding as needed, but IIRC something like this has led to browser fingerprinting in the past. This would also require maintaining a fairly large pool for it to get better throughput and latency than the CSPRNG function, which would massively increase memory usage. (Unless we feed the pool with repetitive data like the current time and then compress it, therefore increasing CPU usage, etc.)

> I also find the ifdefs hard to read, but I understand there will be some of that if we want to use OS APIs directly.

Those ifdef statements **are already there**. I strongly encourage you to read the current AK/Random file. For reasons I've already explained we need an RNG in AK/. I just slightly modified the existing file to add more clear comments, improve error handling, and prevent signal interrupts from causing issues with the RNG on Linux.

If you want to avoid this by using a library or something (there is no other way to obtain secure randomness without the ifdef statements), I already explained that RAND_bytes has unnecessary complexity that makes it slower, and I was told that we don't want to have an OpenSSL dependency in AK.

Alternatively we can just use libsodium's RNG function, which does essentially the same thing as this PR does, although this would again require adding a dependency in AK. I just don't think it's necessary.

> What things use random but cannot depend on LibCrypto? Just IDAllocator in AK?

Yes. There may be something else I'm not aware of but that was the one I ran into.

rgret-dev:
I'm not contesting that AK/Random currently has ifdef-soup or that its implementation uses CSPRNGs already. 

My final thoughts are:

1. We keep `SecureRandom` in LibCrypto (and implement any OS-specific things in there)
2. Work towards removing `AK/Random` (I'm experimenting with this locally at the moment)
3. Anything that needs a high performance PRNG can introduce a dedicated class with the name of that generator (e.g. `XorshiftGenerator` or something) in the future

Again, non-maintainer take so feel free to ignore and let maintainers decide but that is my opinion.

rgret-dev:
Also wanted to respond to what Jelle said:

> drop RAND_bytes() in favor of a static_assert. I'd be interested to see if there are platforms out there where there's no CSPRNG API available

OpenSSL will return an error value that we currently `VERIFY()` on if entropy source fails or is not available, see https://docs.openssl.org/master/man3/RAND_bytes/#notes

With that, I think we should keep OpenSSL as a fallback (which would be impossible if we moved this to AK) so that other, less used platforms can generate secure random values where we lack OS-specific calls.

R-Goc:
Is anyone actually confident in what method openssl even uses for the rand_bytes? I see they seed it with system randomness or other things if that fails but I can't actually tell what is being used. The whole thing is a mess of engine ifdefs, dynamically getting a method, and the methods being on a ctx stuct which is hard to read without an LSP. The docs say the default is a drbg as described in NIST SP 800-90A Rev. 1. That means it is an AES-CTR prng. Reading into the code for the aes rng the code there is basically a reference implementation and is completely unoptimized. So it is very possible the OS is doing something far faster than that. The thing is, with randomness you should have a secure default. If the OS csprng is fast enough, I see no reason at all to use it.

AK/Random.cpp
R-Goc: Also like the build failure says, buf needs to be a PUCHAR so an unsigned char*

colleirose:
> Is anyone actually confident in what method openssl even uses for the rand_bytes?

Well, I'm not. I just gave my general understanding of it, but it is very unclear to me.

> I see they seed it with system randomness or other things if that fails but I can't actually tell what is being used. 

Just to quote https://googleprojectzero.blogspot.com/2016/11/breaking-chain.html?m=1: "There was some OpenSSL code in Flash which uses the desktop window as a source of entropy, but as this could never work in the Chrome sandbox it’s clear that this was vestigial (or Flash’s SSL random number generator is broken, chose one or the other)."

That RNG got removed (it was from the OpenSSL upstream source) but these things are why almost everything besides system entropy was removed from the OpenSSL as an entropy source. 

> The whole thing is a mess of engine ifdefs, dynamically getting a method, and the methods being on a ctx stuct which is hard to read without an LSP. The docs say the default is a drbg as described in NIST SP 800-90A Rev. 1. That means it is an AES-CTR prng. 

I've been looking into this, and, while I know very little about this, I suspect this may be because of FIPS compliance, as aws-lc and BoringSSL both seem to have mostly gotten rid of the other sources but still use the same cipher, and [Google says this is needed for FIPS compliance](https://github.com/google/boringssl/blob/945f4a5d8fd6abf0c7471242151c727e02e889a2/crypto/fipsmodule/FIPS.md#rng-design).

In aws-lc, it is just AES-CTR(system RNG), while in BoringSSL, they do AES-CTR(RDRAND) on Intel CPUs and AES-CTR(system RNG) on other CPUs, because [apparently they say RDRAND is fast on Intel CPUs](https://github.com/google/boringssl/blob/815857bdc3ba18bc666851ac3adf69a668624a1c/crypto/fipsmodule/rand/rand.cc.inc#L346-L359).

We do not need to worry about FIPS compliance. However, if there is a desire for further optimization, we could do something like putting RDRAND and maybe some other entropy source through ChaCha8 or ChaCha12. Of course, RDRAND instructions have been defective before and paranoid users might be suspicious that the NSA has a secret RDRAND exploit.

> Reading into the code for the aes rng the code there is basically a reference implementation and is completely unoptimized. So it is very possible the OS is doing something far faster than that. 

I agree and that's why I believe this PR is also a performance optimization.

> The thing is, with randomness you should have a secure default. If the OS csprng is fast enough, I see no reason at all to use it.

Yes, that's part of the point I tried to make above. A secure default is better IMO. 

----

Also, I will hopefully fix the issues you mentioned later today. I have some code that compiles and works on my Linux machine fine but VirtualBox and VMWare are both being buggy and as I have other things to do it might take a while before I can finish debugging them.

colleirose:
> We keep SecureRandom in LibCrypto (and implement any OS-specific things in there)

That's fine with me as long as the RNG function won't be needed in areas that can't use LibCrypto. That concern is why I kept it in AK/ instead.

> Work towards removing AK/Random (I'm experimenting with this locally at the moment)

Well, let me know if you figure it out. Doing so would require either removing IDAllocator's dependency on Random.cpp, which I don't believe is practical (though perhaps I'm wrong, I only very briefly looked at the file), or allow using LibCrypto in AK/ which I don't believe is going to happen given it contradicts how the existing architecture has been designed. Regardless, imposing a constraint that we can't use the RNG function in AK/ sounds pointless.

> Anything that needs a high performance PRNG can introduce a dedicated class with the name of that generator (e.g. XorshiftGenerator or something) in the future

That's fine but I think you're missing the point. Either we keep it named something like get_random and lose explicitness/clarity or name it something like get_secure_random which could lead to some confusion/uncertainty when someone is working on a file that actually doesn't need secure randomness.

I believe that the best way to address your concern would be naming the RNG function as something like `crypto_get_random`,  `crypto_random`, `crypto_rng`, etc. so that it's less ambiguous that it's a secure RNG but wouldn't seem as strange when used somewhere that doesn't need secure randomness, although of course this is subjective and you may think it still would sound the same. 

Like R-Goc said: "The thing is, with randomness you should have a secure default. If the OS csprng is fast enough, I see no reason at all to use it." And while I know you're not disputing that the operating system CSPRNG is fast enough, I think it seems to be acceptable to use here as a general-purpose secure default. 

> With that, I think we should keep OpenSSL as a fallback (which would be impossible if we moved this to AK) so that other, less used platforms can generate secure random values where we lack OS-specific calls.

The problem with this is that 1) RAND_bytes is slow and inefficient (see above discussion), and 2) I don't believe there are any targets where that fallback would be needed. If someone wants to add a new target they would need to update AK/Random to use its native CSPRNG, which IMO is fairly reasonable.

colleirose:
> With that, I think we should keep OpenSSL as a fallback (which would be impossible if we moved this to AK) so that other, less used platforms can generate secure random values where we lack OS-specific calls.

The problem with this is that 1) RAND_bytes is slow and inefficient (see above discussion), and 2) I don't believe there are any targets where that fallback would be needed. If someone wants to add a new target they would need to update AK/Random to use its native CSPRNG, which IMO is fairly reasonable.

colleirose:
Also, I should be able to test the functionality on Windows tomorrow.

R-Goc:
I ran a benchmark on windows of the current fill_with_random vs fill_with_secure random implementation. The test was 20 runs of filling a total of 1GB of data, at different chunk sizes. For each chunk size there was a warmup of 10 runs of both interleaved. (I can share the code, but it is quite ugly  and non-portable, due to an issue with AK::MonotonicTime on windows I had to use the performace counter directly). https://gist.github.com/R-Goc/cfc91fb86aa21332576844118866174e

colleirose:
My assumption is that OpenSSL maintaining its own entropy pool is faster for larger accesses as getting data from the RAM and CPU cache in that case is faster than the system call, but the system call is faster otherwise. 

I guess the most efficient way to do this would be to maintain an entropy pool for large sizes that's mixed with a faster CSPRNG like ChaCha8 or ChaCha12, using it for large allocations and using the syscall (or maybe RDRAND?) otherwise.

I'll spend some time looking into this I suppose

R-Goc:
Still we are multiple GB/s. So more than enough.

colleirose:
I think what's most relevant would be latency for things like Math.random(), where we don't need very high throughput but a fast response could be useful. I'll need to look into it more later. 

I am pretty sure I could do it but it doesn't seem to matter too much. Maybe in the future tho, I just don't have enough time right now.

colleirose:
I believe the merge conflict should be fixed, I'll need to test on my machine.

colleirose:
Fixed the merge conflicts and the lint issue. Can confirm it's working on Linux, but my VM is still very laggy and so I can't test Windows myself yet though I'll keep trying what I can.

colleirose:
I had attempted to rename the branch... whatever.

AK/Random.h
```diff
@@ -18,34 +18,34 @@
...
 
 }
 
 #if USING_AK_GLOBALLY
-using AK::fill_with_random;
-using AK::get_random;
-using AK::get_random_uniform;
+using AK::crypto_random;
+using AK::crypto_random_uniform;
+using AK::crypto_randombytes_buf;
```
gmta: Imo, the old names were much better. And if our default random is secure (which it should be - no need to use a less secure RNG for anything), we don't necessarily need to reflect that in the name. A higher performance RNG could be named as such to make its intention clear.

AK/Random.cpp
```diff
@@ -1,98 +1,126 @@
...
+#    include <winternl.h>
 #endif
 
-namespace AK {
-
-// NOTE: This function is supposed to always give a random number. If possible it is of good quality, but it can fall
-//       back to rand() if it fails on some systems. For high speed you should probably use a different generator.
-//       See MathObject::random() from LibJS. Where cryptographic security is needed use LibCrypto/SecureRandom.h.
-void fill_with_random([[maybe_unused]] Bytes bytes)
+static inline ErrorOr<void> csprng(void* const buf, size_t size)
```
gmta, line 23: Why is this `void* const buf`? You're immediately `const_cast`ing it in the Windows case.

AK/Random.cpp
```diff
@@ -1,98 +1,126 @@
...
+        if (ret > 0) [[likely]] {
+            size -= ret;
+            buf += ret;
         }
-
-        bytes = bytes.slice(getentropy_length_limit);
     }
+#elif defined(AK_OS_WINDOWS)
+    if (size > NumericLimits<u32>::max()) [[unlikely]]
+        return Error::from_string_literal("The input is too large");
```
gmta: This looks like a footgun: either we should support `size_t` on all plaforms (e.g. by handling this Windows case in chunks) or we should accept a `u32` so we don't end up with cross-platform issues if we're ever requesting bizarre amounts of random data.

AK/Random.cpp
```diff
@@ -1,98 +1,126 @@
...
+    NTSTATUS status = ::BCryptGenRandom(BCRYPT_RNG_ALG_HANDLE, ptr, size, 0);
+    if (!NT_SUCCESS(status)) [[unlikely]] {
+        ULONG error = RtlNtStatusToDosError(status);
+        return Error::from_windows_error(error);
     }
+#else
+    // There shouldn't be a build target where this can happen.
+    //
+    // As for MacOS and iOS, they appear to both support arc4random_buf, and AK/Platform.h defines AK_OS_BSD_GENERIC when building for them,
+    // so they will be covered by the compile target that uses arc4random_buf.
```
gmta: This note feels a bit out of place here. If you're worried that people might misunderstand the first `#if defined(..)`, add a comment there to explain it includes macOS and iOS.

AK/Random.cpp
```diff
@@ -1,98 +1,126 @@
...
-        return;
-#    endif
+namespace AK {
 
-    fill_with_random_fallback();
-#endif
+void crypto_randombytes_buf(Bytes bytes)
+{
+    if (csprng(bytes.data(), bytes.size()).is_error())
+        VERIFY_NOT_REACHED();
```
gmta: The idiomatic way to do this is

```cpp
MUST(csprng(bytes.data(), bytes.size()));
```

AK/Random.h
colleirose: Someone else had said that the generic names like `get_random` lose explicitness, hence why I changed them. I will change them back.

AK/Random.cpp
colleirose: That's a good point, there was this condition earlier but it defaulted back to rand(), and I didn't think to add the chunking aspect.

I got a different VM program working so I should be able to fix the Windows issues soon

AK/Random.cpp
colleirose: Got it, I'll change that

AK/Random.cpp
colleirose: Got it. I'll work on fixing these and hopefully push the changes in a few days, I need to fix my Windows VM to be able to test.

AK/Random.cpp
colleirose: Fixed, using the simpler RtlGenRandom for Windows now. It takes `PVOID` so this isn't a problem anymore.

Libraries/LibCrypto/PK/RSA.cpp
colleirose, line 1: Note that `AK/Random.h` is just removed from the imports here because it isn't actually being used in the file. Noticed this while working on the changes and thought it would be best to remove it given it's a trivial fix.

AK/Random.cpp
```diff
@@ -1,68 +1,116 @@
...
-    NTSTATUS result = ::BCryptGenRandom(NULL, bytes.data(), bytes.size(), BCRYPT_USE_SYSTEM_PREFERRED_RNG);
-    if (result == STATUS_SUCCESS)
-        return;
-#    endif
+    if (!ProcessPrngFunction) [[unlikely]]
+        return Error::from_windows_error();
 
-    fill_with_random_fallback();
+    // ProcessPrng is documented to always return true, so this is just a sanity check.
+    VERIFY(ProcessPrngFunction((PBYTE)buf, size) == TRUE);
```
R-Goc: If it always returns true then don't check it.

AK/Random.cpp
```diff
@@ -1,68 +1,116 @@
...
+    }();
 
-    if (bytes.size() > NumericLimits<u32>::max()) [[unlikely]] {
-        fill_with_random_fallback();
-        return;
-    }
+    if (!hmod) [[unlikely]]
+        return Error::from_windows_error();
+
+    static decltype(&ProcessPrng) ProcessPrngFunction = reinterpret_cast<decltype(&ProcessPrng)>(
```
R-Goc: We now have the SystemApi struct in Windows.h to load various windows functions. Then the function is used as g_system.ApiName().

AK/Random.cpp
colleirose: Got it, I'll change this

AK/Random.cpp
colleirose: Got it, thanks. I'll update the PR with the needed changes soon.

AK/Random.cpp
```diff
@@ -1,68 +1,95 @@
...
+    //
+    // See:
+    // https://learn.microsoft.com/en-us/windows/win32/seccng/processprng
+    // https://github.com/rust-random/getrandom/blob/59d52049773a4f974da4818cf58f96edfc9159f1/src/backends/windows.rs#L1-L22
+    // https://docs.google.com/document/d/13n1t5ak0yofzcadQCF7Ew5TewSUkNfQ3n-IYodjeRYc/edit
+    // https://issues.chromium.org/issues/40088338
+    g_system.ProcessPrng((PBYTE)buf, size);
+#else
+    // There shouldn't be a build target where this can happen.
+    static_assert(false, "This build target doesn't have a valid CSPRNG interface specified in AK/Random.cpp. This needs to be fixed before you can build for this target.");
```
gmta: > This needs to be fixed before you can build for this target.

This is already implied by the `static_assert(false, ..)`

AK/Random.cpp
```diff
@@ -1,68 +1,95 @@
...
-
-    if (bytes.size() > NumericLimits<u32>::max()) [[unlikely]] {
-        fill_with_random_fallback();
-        return;
     }
+#elif defined(AK_OS_WINDOWS)
+    // ProcessPRNG is currently the best option:
+    // - It includes very few imports and doesn't create any handles, therefore being the best for sandboxing and reducing unnecessary imports.
+    // - The BCryptPrimitives DLL is already being imported by libtommath and probably other dependencies so importing it here doesn't matter.
+    // - It's very simple and is documented to always return TRUE
```
gmta: This is all very verbose - please put your reasoning for API choices in the commit messages, and use code comments to explain what or why you're doing something if that's not immediately clear from the code itself.

AK/Random.cpp
```diff
@@ -1,68 +1,95 @@
...
+        // The possible errors that can be returned here are:
+        //
+        // EINTR, interrupted by a signal handler.
+        // EAGAIN, the requested entropy wasn't available. Only possible if the GRND_NONBLOCK flag is set, which it isn't.
+        // EINVAL, invalid flags provided, which there aren't.
+        // ENOSYS, the kernel doesn't implement the syscall. The syscall has been available since Linux 3.17, which was released in 2019.
+        // EFAULT, the address is outside the accessible space; the caller is responsible for providing a valid address.
+        //
+        // Therefore, the only somewhat plausible errors are EFAULT and EINTR.
+        // Nonetheless, we will check for the possibility of all the error codes out of an abundance of caution.
```
gmta: That's a lot of text just to say "we're checking all error cases" - let's put this in the commit message if it's really relevant enough.

AK/Random.cpp
colleirose: Fixed

AK/Random.cpp
colleirose: Fixed

AK/Random.cpp
colleirose: The reason I added a comment here is mainly to note that the ProcessPRNG interface is specifically ideal to use to provide compatibility with future sandboxing measures (see #292, #302, etc.), hence the comment is there to explain why this particular function is being used given that there are many other CSPRNG APIs on Windows. Apparently commit messages (not just commit titles) have to be 72 characters or less as well, hence why I didn't provide an explanation in the commit message. I shortened the explanation now although I can still remove it if needed.

gmta: Commit messages need to be 72 characters per line maximum, not in total.

AK/Windows.h
```diff
@@ -49,23 +51,29 @@ NTAPI NTSTATUS NtCreateWaitCompletionPacket(
...
 inline struct SystemApi {
     PFN_NtAssociateWaitCompletionPacket NtAssociateWaitCompletionPacket = NULL;
     PFN_NtCancelWaitCompletionPacket NtCancelWaitCompletionPacket = NULL;
     PFN_NtCreateWaitCompletionPacket NtCreateWaitCompletionPacket = NULL;
+    PFN_ProcessPrng ProcessPrng = NULL;
 
     SystemApi()
     {
         HMODULE hNtdll = GetModuleHandleW(L"ntdll.dll");
+        HMODULE hBcrypt = LoadLibraryW(L"bcryptprimitives.dll");
```
R-Goc: Nit: could we keep these alphabetic? Doesn't matter now but might be nice in the future. Same with the declaration on top. The link to it( https://learn.microsoft.com/en-us/windows/win32/seccng/processprng) might also be useful since there is no header/import lib, but that's your call.

colleirose: Sure, that sounds fine. I changed the lines there to be in alphabetical order.

I added the documentation link in the top of the header like this:
```
extern "C" {
// NOTE: These are documented here: https://learn.microsoft.com/en-us/windows/win32/devnotes/-win32-misclowlevelclientsupport
// If the function signature changes, we should catch it with GetProcAddress failing in most cases.
// None of these are marked deprecated and they seem to be used internally in the kernel.

NTAPI NTSTATUS NtAssociateWaitCompletionPacket(

. . .

// https://learn.microsoft.com/en-us/windows/win32/seccng/processprng
BOOL WINAPI ProcessPrng(PBYTE pbData, SIZE_T cbData);
}
```

To keep the declarations at the top in alphabetical order while also separating it from the previous statements so it doesn't seem like it's saying ProcessPrng is documented at https://learn.microsoft.com/en-us/windows/win32/devnotes/-win32-misclowlevelclientsupport

But I can move the documentation link further down in the file or something like that if you want.

AK/Random.cpp
colleirose: I see, thank you. I've fixed it.

colleirose:
Merge conflict is fixed and I've verified that everything works correctly

AK/Windows.h
```diff
@@ -44,28 +44,37 @@ NTAPI NTSTATUS NtCreateWaitCompletionPacket(
...
     PFN_NtAssociateWaitCompletionPacket NtAssociateWaitCompletionPacket = NULL;
     PFN_NtCancelWaitCompletionPacket NtCancelWaitCompletionPacket = NULL;
     PFN_NtCreateWaitCompletionPacket NtCreateWaitCompletionPacket = NULL;
+    PFN_ProcessPrng ProcessPrng = NULL;
 
     SystemApi()
     {
+        HMODULE hBcrypt = LoadLibraryW(L"bcryptprimitives.dll");
         HMODULE hNtdll = GetModuleHandleW(L"ntdll.dll");
+        VERIFY(hBcrypt);
```
R-Goc: Missed this before but a small nit is could we name it hBcryptprimitives? Those are two different dlls.

AK/Random.cpp
```diff
@@ -1,68 +1,77 @@
...
+        // > bytes requested via buflen if either GRND_RANDOM was specified in
+        // > flags and insufficient entropy was present in the random source or
+        // > the system call was interrupted by a signal.
+        //
+        // > The user of getrandom() *must* always check the return value, to
+        // > determine whether either an error occurred or fewer bytes than
+        // > requested were returned.
+        if (ret > 0) [[likely]] {
+            size -= ret;
+            buf += ret;
```
R-Goc: Please rebase to have CI rerun. Also check for compile errors locally as this doesn't appear to compile, and even if it did after fixing the const issue and we allowed the gnu extension, it is a code smell. You are doing void* arithmetic here. Also you are assigning to a const variable. https://godbolt.org/z/z54bW34z1
This whole thing is very verbose too, like it was said previously.

R-Goc (changes requested):

AK/Random.cpp
colleirose: Thank you, I will fix this again. I'm not having compile errors locally and it runs for me which is why I assumed that anything that seems odd about it was just because I didn't understand the language well enough. I'll look at this and fix it.

Edit: 

I have glibc on my system, so `AK_LIBC_GLIBC_PREREQ` led to it not compiling the `getrandom()` code. I did local testing by removing the `AK_LIBC_GLIBC_PREREQ` block and adding a dbgln statement to verify that the `getrandom()` code was being reached, then made changes and made sure it would run fine, as well as adding logging statements for each time the block was reached and what the value of `ret` was. I removed the testing code and pushed the new changes. The code seems to work now.

I don't have a working Windows VM, and rebasing the branch still doesn't seem to have triggered CI. However, I will attempt to get a working Windows VM again. If I can make Windows work, I should be able to test this locally.

AK/Windows.h
colleirose: Yes, I can change that

colleirose:
I've fixed the issues on Linux. I tried `git rebase` a second time and it seems to still have not run CI. I still can't make VirtualBox or QEMU work, so I don't have the ability yet to test this on Windows (made some progress in VirtualBox but it randomly crashes after about 30 minutes and the performance has only marginally improved, I just don't have the time to troubleshoot it). Let me know if anything else needs to be changed.

Tests/AK/TestRandom.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+    u32 first_u32 = get_random_uniform(100);
+    u32 second_u32 = get_random_uniform(100);
+
+    u64 first_u64 = AK::get_random_uniform_64(100);
+    u64 second_u64 = AK::get_random_uniform_64(100);
+
+    EXPECT_NE(first_buffer, second_buffer);
+    EXPECT_NE(first_u32, second_u32);
+    EXPECT_NE(first_u64, second_u64);
```
gmta: This seems very flaky - isn't there an approx. 2% chance that either of these will fail during an average run? You're expecting two random numbers from 0 to 99 to not be equal to each other.

You could increase the range by a lot _and_ allow for accidental failures up to a certain threshold before definitively failing the test.

konradekk: maybe use a proper tool for testing instead? a public domain [PractRand](https://pracrand.sourceforge.net) might be just right (cf. `Tests_usage.txt`). this of course can be postponed for a different change.

colleirose: > This seems very flaky - isn't there an approx. 2% chance that either of these will fail during an average run? You're expecting two random numbers from 0 to 99 to not be equal to each other.
> 
> You could increase the range by a lot _and_ allow for accidental failures up to a certain threshold before definitively failing the test.

That's a fair point. I will change this.

> maybe use a proper tool for testing instead? a public domain [PractRand](https://pracrand.sourceforge.net) might be just right (cf. `Tests_usage.txt`). this of course can be postponed for a different change.

This could be more useful if we ever expanded `AK/Random.cpp` to collect entropy from additional sources or something like that. However, it is currently just a wrapper around the system CSPRNG, so that would just be testing the quality of the system's CSPRNG.

colleirose:
Fixed the test case

AK/Windows.h
```diff
@@ -44,28 +44,37 @@ NTAPI NTSTATUS NtCreateWaitCompletionPacket(
...
 
 inline struct SystemApi {
     PFN_NtAssociateWaitCompletionPacket NtAssociateWaitCompletionPacket = NULL;
     PFN_NtCancelWaitCompletionPacket NtCancelWaitCompletionPacket = NULL;
     PFN_NtCreateWaitCompletionPacket NtCreateWaitCompletionPacket = NULL;
+    PFN_ProcessPrng ProcessPrng = NULL;
 
     SystemApi()
     {
+        HMODULE hBcryptprimitives = LoadLibraryW(L"bcryptprimitives.dll");
```
R-Goc, line 65: This should actually be 
```cpp
LoadLibraryExW(L"bcryptprimitives.dll", NULL, LOAD_LIBRARY_SEARCH_SYSTEM32);
```
I'll be doing a cleanup PR for windows stuff later anyway so don't view this as something to block merging this PR if not changed.

Tests/AK/TestRandom.cpp
```diff
@@ -0,0 +1,42 @@
...
+            failures++;
+
+        if (first_u32 == second_u32)
+            failures++;
+
+        if (first_u64 == second_u64)
+            failures++;
+    }
+
+    EXPECT(failures < max_failures);
```
gmta, line 41: This is not the right condition for this test. E.g. if `AK::get_random_uniform_64` fails every time by always returning `42`, but the other two methods work correctly, you almost always still trigger this expectation.

I'd test these in isolation, and _then_ apply a maximum of 3 consecutive failures. Also, 800 bytes seems like it's enough, but for the random number tests I'd use something much larger.

shannonbooth: One note: we're already effectively testing random number equality in CI of 128 bits since an opaque origin is a 128 bit nonce which we compare for equality in various places

AK/Windows.h
colleirose: Thanks, I'll change this

Tests/AK/TestRandom.cpp
colleirose: > This is not the right condition for this test. E.g. if `AK::get_random_uniform_64` fails every time by always returning `42`, but the other two methods work correctly, you almost always still trigger this expectation.

I may be misunderstanding something, so my apologies if this is just wrong, but it seems that each function runs 3 times:
```cpp
TEST_CASE(csprng_generates_unique_values)
{
    constexpr size_t size = 10000;
    constexpr size_t max_failures = 3;
    size_t failures = 0;

    for (size_t i = 0; i < 3; i++) {
```

And in each run, each of these is triggered:
```cpp
        ByteBuffer first_buffer = MUST(ByteBuffer::create_zeroed(size));
        ByteBuffer second_buffer = MUST(ByteBuffer::create_zeroed(size));

        AK::fill_with_random(first_buffer);
        AK::fill_with_random(second_buffer);

        u32 first_u32 = AK::get_random_uniform(size);
        u32 second_u32 = AK::get_random_uniform(size);

        u64 first_u64 = AK::get_random_uniform_64(size);
        u64 second_u64 = AK::get_random_uniform_64(size);
```

And any of them failing all will increment the fail count by 1:
```cpp
        if (first_buffer == second_buffer)
            failures++;

        if (first_u32 == second_u32)
            failures++;

        if (first_u64 == second_u64)
            failures++;
    }

    EXPECT(failures < max_failures);
}
```

Even if only `get_random_uniform_64()` was failing each time, `failures` would increase by one each time it failed and reach 3, and `EXPECT(failures < max_failures)` would fail.

As I explained below I mostly made this test to ensure that the CSPRNG function was not just doing nothing or entirely incorrect. If it was changed to mix in other entropy or something then it'd make sense to perform (better) checks of each function but in this case it's supposed to be testing that each function is still capable of running in a way that seems correct. Nonetheless I will change this as requested to check for each return value separately.

> One note: we're already effectively testing random number equality in CI of 128 bits since an opaque origin is a 128 bit nonce which we compare for equality in various places

The particular reason I made a CSPRNG test was because I couldn't get a working Windows VM and was told that I should create a test so that it could be run automatically by CI: https://discord.com/channels/1247070541085671459/1247090705013280769/1441158902891937883. I don't know if this is really a good reason to make a test case but it seemed like the best suggestion given that I couldn't test it on Windows.

gmta: Completely agree - I misread your test! Thanks :)

PR #6567, opened 2025-10-24
LibRegex: Properly track code units in unicode mode

aplefull:
We were incorrectly tracking code units before this change, for example for string `"𠮷a𠮷b𠮷"`, we were getting:
```
state.string_position=0, (in code units: 0)
state.string_position=1, (in code units: 1)
state.string_position=2, (in code units: 2)
```
which resulted in incorrect matches, now code units update properly:
```
state.string_position=0, (in code units: 0)
state.string_position=1, (in code units: 2)
```
Test 262 diff:
        +3 ✅    -3 ❌   

Diff Tests:
    test/built-ins/RegExp/prototype/exec/regexp-builtin-exec-v-u-flag.js           ❌ -> ✅
    test/built-ins/String/prototype/matchAll/regexp-prototype-matchAll-v-u-flag.js ❌ -> ✅
    test/built-ins/String/prototype/replace/regexp-prototype-replace-v-u-flag.js   ❌ -> ✅

### DIFF
--- a/Libraries/LibJS/Tests/builtins/RegExp/RegExp.js
+++ b/Libraries/LibJS/Tests/builtins/RegExp/RegExp.js
@@ -156,3 +156,37 @@ test("Unicode properties of strings", () => {
         expect(re.test(str)).toBeFalse();
     }
 });
+
+test("Unicode matching with u and v flags", () => {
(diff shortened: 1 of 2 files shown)


alimpfard:
LGTM, thanks!

PR #6572, opened 2025-10-24
TESTING 1-2-3 [skip ci]

rcorsi:
Doing a test of the skip ci to see if I could upload something just to "show it" without running the CI jobs.


rcorsi:
Seems to work.  Though I guess this is not really needed. I could just push to my fork of ladybird and point there instead.

PR #6576, opened 2025-10-25
LibWeb: Reduce rebuild time for common CSS changes

Calme1709:
Significantly reduces the rebuild required when editing a couple of often edited files in CSS-land (PropertyID.h, StyleComputer.h and ComputedProperties.h)

### DIFF
--- a/Libraries/LibWeb/Animations/AnimationEffect.cpp
+++ b/Libraries/LibWeb/Animations/AnimationEffect.cpp
@@ -13,6 +13,7 @@
 #include <LibWeb/Bindings/Intrinsics.h>
 #include <LibWeb/CSS/ComputedProperties.h>
 #include <LibWeb/CSS/Parser/Parser.h>
+#include <LibWeb/CSS/PropertyID.h>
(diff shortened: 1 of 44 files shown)


Libraries/LibWeb/DOM/Element.h
```diff
@@ -502,33 +501,7 @@ class WEB_API Element
...
-                continue; // Skip nodes that do not participate in the layout.
-
-            if (!element->computed_properties()->display().is_list_item())
-                continue; // Skip nodes that are not list items.
-
-            if (callback(element) == IterationDecision::Break)
-                return;
-        }
-    }
+    void for_each_numbered_item_owned_by_list_owner(Callback callback);
```
AtkinsSJ: This should probably be private?

Calme1709: Looks like all callers are within `Element` so may as well

AtkinsSJ: Yeah, and it wouldn't be valid to call it from elsewhere because its definition is out of line.

PR #5435, opened 2025-07-14
LibCore: Implement TCPServer+UDPServer+LocalServer on Windows

ayeteadoe:
### Demo

To test my impls I made simple Marco-Polo servers where a Python or C++ client connects and starts a 5 entry long chain of "macro!" (Client) and "polo!" (Server) back and forth messages. After the 5th "polo!" is sent, the Server shuts down.

#### TCPServer

https://github.com/user-attachments/assets/7e0857e2-a0a3-4c8e-91a4-4b5fb85d4124

<details>

<summary>Here is the code for both the Macro-Polo server and the python Client.</summary>

### TestTCPServer.cpp
```cpp
#include <AK/StringView.h>
#include <AK/Vector.h>
#include <LibCore/EventLoop.h>
#include <LibCore/Socket.h>
#include <LibCore/TCPServer.h>
#include <LibMain/Main.h>
#include <LibTest/TestCase.h>

using namespace Core;
class MarcoPoloServer : public EventReceiver {
    C_OBJECT(MarcoPoloServer)
public:
    static ErrorOr<NonnullRefPtr<MarcoPoloServer>> try_create()
    {
        auto server = TRY(TCPServer::try_create());
        return adopt_nonnull_ref_or_enomem(new (nothrow) MarcoPoloServer(move(server)));
    }

    ErrorOr<void> start_listening(u16 port = 8080)
    {
        TRY(m_server->listen(IPv4Address::from_string("127.0.0.1"sv).value(), port));

        m_server->on_ready_to_accept = [this]() {
            handle_new_connection();
        };

        outln("Server listening on localhost:{}", port);
        return {};
    }

    void shutdown()
    {
        outln("Shutting down server gracefully...");
        m_clients.clear();
        Core::EventLoop::current().quit(0);
    }

private:
    explicit MarcoPoloServer(NonnullRefPtr<TCPServer> server)
        : m_server(move(server))
    {
    }

    bool handle_new_connection()
    {
        auto socket_result = m_server->accept();
        if (socket_result.is_error()) {
            outln("Failed to accept connection: {}", socket_result.error());
            return false;
        }

        auto socket = socket_result.release_value();
        outln("New client connected!");

        auto client = make<Client>(move(socket), *this);
        m_clients.append(move(client));
        return true;
    }

    struct Client {
        NonnullOwnPtr<TCPSocket> socket;
        MarcoPoloServer& server;
        int exchange_count = 0;
        Vector<u8> buffer;

[...]

PR #6579, opened 2025-10-25
RequestServer: De-duplicate some disk cache requests

trflynn89:
We previously had no protection against the same URL being requested multiple times at the same time. For example, if a URL did not have any cache entry and became requested twice, we would open two cache writers concurrently. This would result in both writers piping the response to disk, and we'd have a corrupt cache file.

This PR changes RequestServer to keep requests on hold until the existing cache entry has been complete (note we still allow multiple read entries at a time).

Doing so got a bit hairy with the existing `ActiveRequest` infra, so first this PR changes RS to manage request lifetimes as a simple state machine. This made it trivial to add a waiting-for-disk-cache state.

An easy example of duplicate requests is on xkcd.com, where the comic image is requested twice:
Screenshot_20251025_123449

### DIFF
--- a/Services/RequestServer/Cache/CacheEntry.h
+++ b/Services/RequestServer/Cache/CacheEntry.h
@@ -55,6 +55,8 @@ class CacheEntry {
 public:
     virtual ~CacheEntry() = default;
 
+    u64 cache_key() const { return m_cache_key; }
+
     void remove();
 
     void mark_for_deletion(Badge<DiskCache>) { m_marked_for_deletion = true; }
--- a/Services/RequestServer/Cache/DiskCache.cpp
+++ b/Services/RequestServer/Cache/DiskCache.cpp
@@ -33,47 +33,54 @@ DiskCache::DiskCache(NonnullRefPtr<Database::Database> database, LexicalPath cac
 {
 }
 
-Optional<CacheEntryWriter&> DiskCache::create_entry(Request& request)
+Variant<Optional<CacheEntryWriter&>, DiskCache::CacheHasOpenEntry> DiskCache::create_entry(Request& request)
 {
     if (!is_cacheable(request.method()))
-        return {};
+        return Optional<CacheEntryWriter&> {};
 
     auto serialized_url = serialize_url_for_cache_storage(request.url());
     auto cache_key = create_cache_key(serialized_url, request.method());
+
+    if (check_if_cache_has_open_entry(request, cache_key, CheckReaderEntries::Yes))
+        return CacheHasOpenEntry {};
 
     auto cache_entry = CacheEntryWriter::create(*this, m_index, cache_key, move(serialized_url), request.request_start_time());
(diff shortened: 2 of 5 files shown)


Services/RequestServer/CURL.cpp
```diff
@@ -0,0 +1,58 @@
...
+    case CURLE_COULDNT_CONNECT:
+        return Requests::NetworkError::UnableToConnect;
+    case CURLE_OPERATION_TIMEDOUT:
+        return Requests::NetworkError::TimeoutReached;
+    case CURLE_TOO_MANY_REDIRECTS:
+        return Requests::NetworkError::TooManyRedirects;
+    case CURLE_SSL_CONNECT_ERROR:
+        return Requests::NetworkError::SSLHandshakeFailed;
+    case CURLE_PEER_FAILED_VERIFICATION:
+        return Requests::NetworkError::SSLVerificationFailed;
```
konradekk, line 48: not sure if `SSL` (here and above) is something should be retained; `TLS` or even `Secure` would be better instead?

konradekk (commented):
naming nit.

Services/RequestServer/CURL.cpp
trflynn89: This is pre-existing code that I only moved, I won't be renaming these here - there's enough going on in this PR without such refactors.

rgret-dev:
Not directly related, but the method `close_and_destory_cache_entry` has a typo:
"destory" -> "destroy"

Services/RequestServer/Resolver.cpp
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+#if !defined(AK_OS_WINDOWS)
+        return DNS::Resolver::SocketResult {
+            MaybeOwned<Core::Socket>(TRY(Core::BufferedUDPSocket::create(TRY(Core::UDPSocket::connect(*dns_info.server_address))))),
+            DNS::Resolver::ConnectionMode::UDP,
+        };
+#else
+        return Error::from_string_literal("Core::UDPSocket::connect() and Core::BufferedUDPSocket::create() are not implemented on Windows");
+#endif
```
ayeteadoe: This is actually supported on Windows since https://github.com/LadybirdBrowser/ladybird/pull/5435/commits/334da2072e8a34e348b8a6961d69dc7c97072446.

I have removed the if-def in the pending "Launch Ladybird on Windows" PR: https://github.com/LadybirdBrowser/ladybird/pull/5981/commits/8e5605ed18261d60821a9a43e53acfb970a9abc9, but it'd make resolving my merge conflicts a bit easier if it is addressed in this PR so I can just drop the commit in my PR

trflynn89: Sure, I pulled 8e5605ed18261d60821a9a43e53acfb970a9abc9 into this PR.

PR #6584, opened 2025-10-26
LibWeb: Sort Table rows HTMLCollection

lpas:
Add optional sorting function to HTMLCollections.
We are using insertion_sort to have an stable sorting algorithm if we need to keep the current tree order.

### DIFF
--- a/Libraries/LibWeb/DOM/HTMLCollection.cpp
+++ b/Libraries/LibWeb/DOM/HTMLCollection.cpp
@@ -5,6 +5,7 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
+#include <AK/InsertionSort.h>
 #include <LibWeb/Bindings/HTMLCollectionPrototype.h>
 #include <LibWeb/Bindings/Intrinsics.h>
 #include <LibWeb/DOM/Document.h>
@@ -17,15 +18,16 @@ namespace Web::DOM {
 
 GC_DEFINE_ALLOCATOR(HTMLCollection);
 
-GC::Ref<HTMLCollection> HTMLCollection::create(ParentNode& root, Scope scope, Function<bool(Element const&)> filter)
+GC::Ref<HTMLCollection> HTMLCollection::create(ParentNode& root, Scope scope, Function<bool(Element const&)> filter, Function<bool(Element const&, Element const&)> sort)
 {
-    return root.realm().create<HTMLCollection>(root, scope, move(filter));
+    return root.realm().create<HTMLCollection>(root, scope, move(filter), move(sort));
 }
 
-HTMLCollection::HTMLCollection(ParentNode& root, Scope scope, Function<bool(Element const&)> filter)
+HTMLCollection::HTMLCollection(ParentNode& root, Scope scope, Function<bool(Element const&)> filter, Function<bool(Element const&, Element const&)> sort)
     : PlatformObject(root.realm())
     , m_root(root)
     , m_filter(move(filter))
+    , m_sort(move(sort))
     , m_scope(scope)
 {
     m_legacy_platform_object_flags = LegacyPlatformObjectFlags {
@@ -95,6 +97,13 @@ void HTMLCollection::update_cache_if_needed() const
             return IterationDecision::Continue;
         });
     }
+
+    if (m_sort) {
+        insertion_sort(m_cached_elements, [this](auto const& a, auto const& b) {
+            return this->m_sort(a, b);
+        });
+    }
+
     m_cached_dom_tree_version = root()->document().dom_tree_version();
 }
 
--- a/Libraries/LibWeb/DOM/HTMLCollection.h
+++ b/Libraries/LibWeb/DOM/HTMLCollection.h
@@ -30,7 +30,7 @@ public:
         Children,
         Descendants,
     };
-    [[nodiscard]] static GC::Ref<HTMLCollection> create(ParentNode& root, Scope, ESCAPING Function<bool(Element const&)> filter);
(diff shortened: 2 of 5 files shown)


Libraries/LibWeb/DOM/HTMLCollection.cpp
```diff
@@ -5,6 +5,7 @@
  * SPDX-License-Identifier: BSD-2-Clause
```
gmta, line 5: Commit title: currently it's not really a sentence. Please use the imperative to describe what you changed.

Commit message: please auto-wrap at 72 characters. "HTMLCollections" does not exist. Typo: "an stable" -> "a stable". "Add optional" -> "Add an optional"

Libraries/LibWeb/HTML/HTMLTableElement.cpp
```diff
@@ -398,11 +398,10 @@ GC::Ref<HTMLTableSectionElement> HTMLTableElement::create_t_body()
...
     HTMLTableElement* table_node = this;
-    // FIXME:  The elements in the collection must be ordered such that those elements whose parent is a thead are
-    //         included first, in tree order, followed by those elements whose parent is either a table or tbody
-    //         element, again in tree order, followed finally by those elements whose parent is a tfoot element,
-    //         still in tree order.
-    // How do you sort HTMLCollection?
+    // The elements in the collection must be ordered such that those elements whose parent is a thead are
+    // included first, in tree order, followed by those elements whose parent is either a table or tbody
+    // element, again in tree order, followed finally by those elements whose parent is a tfoot element,
+    // still in tree order.
```
gmta: Please rewrap this comment at column 120.

Libraries/LibWeb/DOM/HTMLCollection.cpp
gmta: Commit message: "We are using insertion_sort to have a stable sorting algorithm if we need to keep
the current tree order."

It's not "if", you always need to maintain tree order as far as I can see. You could change this sentence to say something like:
"We use insertion_sort as a stable sorting algorithm so tree order can be maintained."

lpas: I guess I thought about this in a more abstract way if someone uses the sort and needs to keep tree order the option is there. In my case I need to preserve tree order always yes.

Libraries/LibWeb/DOM/HTMLCollection.h
```diff
@@ -30,7 +30,7 @@ class HTMLCollection : public Bindings::PlatformObject {
         Children,
         Descendants,
     };
-    [[nodiscard]] static GC::Ref<HTMLCollection> create(ParentNode& root, Scope, ESCAPING Function<bool(Element const&)> filter);
+    [[nodiscard]] static GC::Ref<HTMLCollection> create(ParentNode& root, Scope, ESCAPING Function<bool(Element const&)> filter, Function<bool(Element const&, Element const&)> sort = nullptr);
```
AtkinsSJ: I think `sort` needs to be marked `ESCAPING` as well. (Also in the constructor.)

Libraries/LibWeb/HTML/HTMLTableElement.cpp
```diff
@@ -415,13 +412,28 @@ GC::Ref<DOM::HTMLCollection> HTMLTableElement::rows()
...
 
-            if (element.parent_element() && (element.parent_element()->local_name() == TagNames::thead || element.parent_element()->local_name() == TagNames::tbody || element.parent_element()->local_name() == TagNames::tfoot)
-                && element.parent()->parent() == table_node) {
+            if (element.parent_element() && element.parent_element()->local_name().is_one_of(TagNames::thead, TagNames::tbody, TagNames::tfoot) && element.parent()->parent() == table_node)
                 return true;
-            }
 
-            return false;
-        });
+            return false; }, [](Element const& a, Element const& b) -> bool {
```
AtkinsSJ: I don't know what clang-format was thinking with this line, but can you make it not do that? :sweat_smile:

Libraries/LibWeb/HTML/HTMLTableElement.cpp
```diff
@@ -415,13 +412,28 @@ GC::Ref<DOM::HTMLCollection> HTMLTableElement::rows()
...
+                auto static sort_prio = [](Element const& element) {
```
AtkinsSJ: `sort_priority` rather than `sort_prio` please, we try to avoid abbreviations generally.

Libraries/LibWeb/HTML/HTMLTableElement.cpp
```diff
@@ -415,13 +412,28 @@ GC::Ref<DOM::HTMLCollection> HTMLTableElement::rows()
...
+                    auto const& parentTag = element.parent_element()->local_name();
```
AtkinsSJ: `parent_tag`

Libraries/LibWeb/DOM/HTMLCollection.cpp
```diff
@@ -95,6 +97,13 @@ void HTMLCollection::update_cache_if_needed() const
             return IterationDecision::Continue;
         });
     }
+
+    if (m_sort) {
+        insertion_sort(m_cached_elements, [this](auto const& a, auto const& b) {
+            return this->m_sort(a, b);
+        });
```
AtkinsSJ, line 104: Do you need to wrap `sort` in a lambda, or would `insertion_sort(m_cached_elements, sort);` work? (I'm genuinely not sure, I never remember the nuances.)

lpas: I tried that before, but I couldn't get it to work.

PR #6585, opened 2025-10-26
LibWeb: Support multiple animations per animation property

Calme1709:
Builds upon the work done in #6267 and actually triggers multiple animations if required by an element's `animation-*` properties.


Calme1709:
AFAICT this isn't actually tested in WPT further than parsing which we already had (atleast in /css/css-animations or /web-animations) but here's hoping

PR #6588, opened 2025-10-26
LibWeb: implement the MediaSession API

dd-dreams:
https://w3c.github.io/mediasession

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLMediaElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLMediaElement.cpp
@@ -12,7 +12,6 @@
 #include <LibMedia/Track.h>
 #include <LibWeb/Bindings/HTMLMediaElementPrototype.h>
 #include <LibWeb/Bindings/Intrinsics.h>
-#include <LibWeb/Bindings/MediaSessionPrototype.h>
 #include <LibWeb/DOM/Document.h>
 #include <LibWeb/DOM/DocumentObserver.h>
 #include <LibWeb/DOM/Event.h>
@@ -30,7 +29,6 @@
 #include <LibWeb/HTML/HTMLSourceElement.h>
 #include <LibWeb/HTML/HTMLVideoElement.h>
 #include <LibWeb/HTML/MediaError.h>
-#include <LibWeb/HTML/Navigator.h>
 #include <LibWeb/HTML/PotentialCORSRequest.h>
 #include <LibWeb/HTML/Scripting/Environments.h>
 #include <LibWeb/HTML/Scripting/TemporaryExecutionContext.h>
@@ -40,7 +38,6 @@
 #include <LibWeb/HTML/TrackEvent.h>
 #include <LibWeb/HTML/VideoTrack.h>
 #include <LibWeb/HTML/VideoTrackList.h>
-#include <LibWeb/HTML/Window.h>
 #include <LibWeb/Layout/Node.h>
 #include <LibWeb/MimeSniff/MimeType.h>
 #include <LibWeb/Page/Page.h>
@@ -2215,20 +2212,10 @@ bool HTMLMediaElement::handle_keydown(Badge<Web::EventHandler>, UIEvents::KeyCod
     if (modifiers != UIEvents::KeyModifier::Mod_None)
         return false;
 
-    auto const& mediasession = document().window()->navigator()->media_session();
-
     switch (key) {
-    case UIEvents::KeyCode::Key_PlayPause:
-    case UIEvents::KeyCode::Key_Space: {
-        if (!mediasession->has_action_handler(Bindings::MediaSessionAction::Play) &&
-                mediasession->has_action_handler(Bindings::MediaSessionAction::Pause))
-            toggle_playback();
-        else if (potentially_playing())
-            mediasession->handle_action(Bindings::MediaSessionAction::Play);
-        else
-            mediasession->handle_action(Bindings::MediaSessionAction::Pause);
+    case UIEvents::KeyCode::Key_Space:
+        toggle_playback();
         break;
-    }
 
     case UIEvents::KeyCode::Key_Home:
         set_current_time(0);
--- a/Libraries/LibWeb/MediaSession/MediaSession.cpp
+++ b/Libraries/LibWeb/MediaSession/MediaSession.cpp
@@ -1,44 +1,71 @@
+#include <LibWeb/WebIDL/AbstractOperations.h>
 #include <LibWeb/MediaSession/MediaSession.h>
-#include <LibWeb/HTML/Window.h>
 #include <LibWeb/Platform/EventLoopPlugin.h>
 #include <LibWeb/HTML/Navigator.h>
 #include <LibWeb/Bindings/MediaSessionPrototype.h>
+#include <LibWeb/Bindings/Intrinsics.h>
 
 namespace Web::MediaSession {
 
 GC_DEFINE_ALLOCATOR(MediaSession);
 
-// void MediaSession::update_metadata(GC::Ref<MediaMetadata> metadata) const {
-//     // const auto& navigator = as<HTML::Navigator>(*this);
-// }
-
-GC::Ref<MediaSession> MediaSession::create(HTML::Window& window) {
-    return window.realm().create<MediaSession>(window);
+GC::Ref<MediaSession> MediaSession::create(JS::Realm& realm) {
+    return realm.create<MediaSession>(realm);
 }
 
-MediaSession::MediaSession(HTML::Window& window)
-    : DOM::EventTarget(window.realm())
-    , m_window(window)
+MediaSession::MediaSession(JS::Realm& realm)
+    : Bindings::PlatformObject(realm)
 {
 }
 
 MediaSession::~MediaSession() = default;
 
+// https://w3c.github.io/mediasession/#dom-mediasession-setactionhandler
 WebIDL::ExceptionOr<void> MediaSession::set_action_handler(Bindings::MediaSessionAction action, MediaSessionActionHandler handler) {
+    // https://w3c.github.io/mediasession/#update-action-handler-algorithm
     if (!handler)
-        m_action_handlers.remove(action);
+        m_supported_media_session_actions.remove(action);
     else
-        m_action_handlers.set(action, handler);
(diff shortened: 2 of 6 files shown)


dd-dreams:
There might be some code linting that should be done, but first I would like to hear some suggestions/improvements if there are any.

Libraries/LibWeb/HTML/HTMLMediaElement.cpp
```diff
@@ -259,6 +263,12 @@ double HTMLMediaElement::current_time() const
 // https://html.spec.whatwg.org/multipage/media.html#dom-media-currenttime
 void HTMLMediaElement::set_current_time(double current_time)
 {
+    auto const& mediasession = document().window()->navigator()->media_session();
+
+    if (mediasession->has_action_handler(Bindings::MediaSessionAction::Seekto))
```
Psychpsyo: The spec says nothing about doing this here. If that's a bug with the spec, you should probably file an issue there. If it isn't, then this is probably in the wrong place. Either way, it should probably have some comment.
Also, `// FIXME:`, not `// TODO:`.

Libraries/LibWeb/HTML/HTMLMediaElement.cpp
```diff
@@ -2012,29 +2022,72 @@ WebIDL::ExceptionOr<bool> HTMLMediaElement::handle_keydown(Badge<Web::EventHandl
     if (modifiers != UIEvents::KeyModifier::Mod_None)
         return false;
 
+    auto const& mediasession = document().window()->navigator()->media_session();
+
+    auto details = MediaSession::MediaSessionActionDetails {
+        Bindings::MediaSessionAction::Pause,
+        0,
+        0,
+        true, // because we triggered an hardware key event
```
Psychpsyo: Typo: "a hardware key event", not "an hardware key event"

Libraries/LibWeb/MediaSession/MediaSession.cpp
```diff
@@ -0,0 +1,175 @@
...
+#include <LibWeb/Platform/EventLoopPlugin.h>
+#include <LibWeb/HTML/Navigator.h>
+#include <LibWeb/Bindings/MediaSessionPrototype.h>
+#include <LibWeb/HTML/HTMLHtmlElement.h>
+
+namespace Web::MediaSession {
+
+GC_DEFINE_ALLOCATOR(MediaSession);
+
+// void MediaSession::update_metadata(GC::Ref<MediaMetadata> metadata) const {
```
Psychpsyo: There shouldn't be any code that's just commented out.

Libraries/LibWeb/MediaSession/MediaSession.h
```diff
@@ -0,0 +1,76 @@
...
+    void update_metadata(GC::Ref<MediaMetadata>) const;
+
+    GC::Ptr<MediaMetadata> m_metadata;
+
+    Bindings::MediaSessionPlaybackState m_playback_state;
+
+    MediaPositionState m_position_state;
+
+    // https://w3c.github.io/mediasession/#supported-media-session-actions
+    HashMap<Bindings::MediaSessionAction, MediaSessionActionHandler> m_action_handlers;
```
Psychpsyo: Any reason for deviating from the spec and not calling this m_supported_media_session_actions?

Psychpsyo:
It would be nice if all the new functions and stuff had spec links to make it easy to see what they're actually implementing.

Libraries/LibWeb/HTML/HTMLMediaElement.cpp
dd-dreams: I moved this to `seek_element`, although this is not specified. Basically `seekto` handler should be fired when we seek, so this is probably the correct place. Should I file an issue to W3C?

Psychpsyo: > Basically `seekto` handler should be fired when we seek

Where in the spec does it say that? Cause I can't find it anywhere.
[Handle a media session action](https://w3c.github.io/mediasession/#handle-media-session-action) only gets called when a [media session action source](https://w3c.github.io/mediasession/#media-session-action-source) produces a media session action.
HTMLMediaElement, or any of the spec algorithms on it (like [seek](https://html.spec.whatwg.org/multipage/media.html#dom-media-seek)) are not a media session action source and shouldn't be calling these.

As far as I can see in the spec, a media session action source is either a physical button on something like a keyboard or headset or something equivalent like the media buttons in the UI on a phone's lock screen. (or any other platform UI that gets created for the media session)
`video` or `audio` elements in any open website do not qualify.

TL;DR: You have nowhere to be calling this from yet.

dd-dreams: I figured that's what this action should do. This my own perception. Looked at some examples and it looked like this what should be done. It might be that the spec didn't say how to implement this yet. It seems although that these "actions" should also be fired from the UI if some UI button is pressed: https://w3c.github.io/mediasession/#media-session-action-source:~:text=UI%20surfaces%20created%20by%20the%20user%20agent.

dd-dreams: Right, will remove this. I should also probably remove the space key as a "source".

dd-dreams: > HTMLMediaElement, or any of the spec algorithms on it (like [seek](https://html.spec.whatwg.org/multipage/media.html#dom-media-seek)) are not a media session action source and shouldn't be calling these.

I used HTMLMediaElement because I wanted to get the keys that the media element receives. Should this actually happen somewhere else? Basically, is `setActionHandler` currently should be left unused?

Psychpsyo: > > HTMLMediaElement, or any of the spec algorithms on it (like [seek](https://html.spec.whatwg.org/multipage/media.html#dom-media-seek)) are not a media session action source and shouldn't be calling these.
> 
> I used HTMLMediaElement because I wanted to get the keys that the media element receives. Should this actually happen somewhere else? Basically, is `setActionHandler` currently should be left unused?

They keys on the media element are a bad choice because a page might have 10 videos on it that are all completely unrelated to a MediaSession that it opened.

`setActionHandler` is a thing for JavaScript in the website to call. This is available to JS because of the bindings. It shouldn't be called from anywhere if the spec doesn't explicitly say to call it.

dd-dreams: > They keys on the media element are a bad choice because a page might have 10 videos on it that are all completely unrelated to a MediaSession that it opened.

As far as I understood, this is the responsibility of the site developer to make sure `navigator.mediaSession` is actually related to the video that is currently playing on the same site.

> `setActionHandler` is a thing for JavaScript in the website to call. This is available to JS because of the bindings. It shouldn't be called from anywhere if the spec doesn't explicitly say to call it.

Sorry, I meant like the actual implementation. For example, if we get the `play` action, we need to run the action callback when we hit the `play/pause` button. Currently, the media on the Ladybird isn't shared with the OS.

Psychpsyo: > As far as I understood, this is the responsibility of the site developer to make sure `navigator.mediaSession` is actually related to the video that is currently playing on the same site.

The [abstract of the spec](https://w3c.github.io/mediasession/#abstract) reads:
> This specification enables web developers to show customized media metadata on platform UI, customize available platform media controls, and access platform media keys such as hardware keys found on keyboards, headsets, remote controls, and software keys found in notification areas and on lock screens of mobile devices.

The entirety of the media session spec is to interface with OS/Hardware media buttons and UIs. It is not meant to extend parts of the webpage, beyond what JS does in its callbacks.
JS can register callbacks, specifically so that it can play/pause/seek/whatever the piece of media when such a hardware button, that would not usually do this, is pressed.
A `play` action handler is there to inform JS that the user wants the media to start playing, so that JS can go ahead and make that happen. It's not there to inform JS of the fact that the media just started playing, if JS wanted to know about that, it should be using the [events that get dispatched on a media element](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement#events) when these things happen.

> Sorry, I meant like the actual implementation. For example, if we get the `play` action, we need to run the action callback when we hit the `play/pause` button. Currently, the media on the Ladybird isn't shared with the OS.

If a media session gets a play action (the user pressed a media button on their keyboard/headset/OS...) we need to run the action handler that JS gave us so that that action handler can go and hit `play()` on the media element. The MediaSession API should not be interacting with media elements itself at all, outside of monitoring them to determine the [guessed playback state](https://w3c.github.io/mediasession/#guessed-playback-state) and the [position state](https://w3c.github.io/mediasession/#position-state).

dd-dreams: Alright it makes sense. So the MediaSession API has no interaction whatsoever with the media elements? That makes things much simpler.

dd-dreams: We would also need to monitor these kind of media keys, where it should be done if not in the HTMLMediaElement?

Psychpsyo: > We would also need to monitor these kind of media keys, where it should be done if not in the HTMLMediaElement?

Media keys... maybe. Spacebar, arrow keys and so on, no.
For the media keys it'd probably be relevant to know how operating systems handle media sessions, and if they somehow do the bit of associating hardware keys to them.

dd-dreams: We already have a few keys available such as `UIEvents::KeyCode::Key_PlayPause`, but they're not responsive. There needs to be some registration with the OS to say that Ladybird is currently the player and receive commands from the OS.

dd-dreams: For example for macos: https://developer.apple.com/documentation/mediaplayer/mpnowplayingsession

dd-dreams: Removed what I added in HTMLMediaElement. Going now to implement Now Playing for macOS.

dd-dreams:
I'll continue working on this soon. What is left is just registering the browser as a NowPlayable app on macOS. Still haven't worked on Linux and Windows. On Windows I probably won't.

PR #6589, opened 2025-10-26
Meta+Everywhere: Fix OpenBSD support and add minor FreeBSD improvements

cqundefine:


Meta/CMake/vcpkg/overlay-ports/angle/portfile.cmake
```diff
@@ -131,6 +131,9 @@ vcpkg_download_distfile(WK_ANGLE_CMAKE_WEBKITCOMPILERFLAGS
     FILENAME "WebKitCompilerFlags.cmake"
     SHA512 8b281ffcf9209c845a5fdae48a4e05f08ca677c37a7fb00d9270de81bd103160d26e091724ce8df8d428ad604900b5202b221fed5bafffd4bf00025718ef9d8e
 )
+file(READ "${WK_ANGLE_CMAKE_WEBKITCOMPILERFLAGS}" _wk_compilerflags_content)
+string(REPLACE "-Wl,--no-undefined" "" _wk_compilerflags_content "${_wk_compilerflags_content}")
+file(WRITE "${WK_ANGLE_CMAKE_WEBKITCOMPILERFLAGS}" "${_wk_compilerflags_content}")
```
trflynn89: Can you explain why we're removing `-Wl,--no-undefined`? Just calling it a fix doesn't explain much, and it's seemingly been fine on our supported systems thus far.

An aside, one concern I have with our overlay ports is keeping track of changes we made to upstream files. When we go to update the angle port, it's not going to be obvious that these 3 lines were added by us and should be retained. I'm thinking it will be much clearer if we mark blocks we added with a `# Ladybird: <reason for this change>` comment, surrounded by newlines:

```cmake
vcpkg_download_distfile(WK_ANGLE_CMAKE_WEBKITCOMPILERFLAGS
    URLS "https://github.com/WebKit/WebKit/raw/${ANGLE_WEBKIT_BUILDSYSTEM_COMMIT}/Source/cmake/WebKitCompilerFlags.cmake"
    FILENAME "WebKitCompilerFlags.cmake"
    SHA512 8b281ffcf9209c845a5fdae48a4e05f08ca677c37a7fb00d9270de81bd103160d26e091724ce8df8d428ad604900b5202b221fed5bafffd4bf00025718ef9d8e
)

# Ladybird: abcdefghi...
file(READ "${WK_ANGLE_CMAKE_WEBKITCOMPILERFLAGS}" _wk_compilerflags_content)
string(REPLACE "-Wl,--no-undefined" "" _wk_compilerflags_content "${_wk_compilerflags_content}")
file(WRITE "${WK_ANGLE_CMAKE_WEBKITCOMPILERFLAGS}" "${_wk_compilerflags_content}")

file(COPY "${WK_ANGLE_CMAKE_WEBKITCOMPILERFLAGS}" DESTINATION "${SOURCE_PATH}/cmake")
```

Libraries/LibCore/System.cpp
```diff
@@ -811,7 +853,6 @@ ErrorOr<ByteString> current_executable_path()
     return Error::from_string_literal("current_executable_path() unknown on this platform");
 #else
 #    warning "Not sure how to get current_executable_path on this platform!"
-    // GetModuleFileName on Windows, unsure about OpenBSD.
```
trflynn89, line 804: The comment about Windows is still relevant.

Libraries/LibCore/System.cpp
```diff
@@ -793,6 +793,48 @@ ErrorOr<ByteString> current_executable_path()
...
+    if (sysctl(mib, 4, nullptr, &argv_length, nullptr, 0) < 0)
+        return Error::from_syscall("sysctl"sv, errno);
+
+    auto argv_buffer = TRY(ByteBuffer::create_uninitialized(argv_length));
+
+    if (sysctl(mib, 4, argv_buffer.data(), &argv_length, nullptr, 0) < 0)
+        return Error::from_syscall("sysctl"sv, errno);
+
+    char const* c_argv0 = bit_cast<char const**>(argv_buffer.data())[0];
+    StringView argv0 = { c_argv0, strlen(c_argv0) };
```
trflynn89: Do we actually need the byte buffer / bit cast? Could we do something like:
```c++
Vector<char*> argv;
argv.resize(argv_length);

if (sysctl(mib, 4, argv.data(), &argv_length, nullptr, 0) < 0)
    return Error::from_syscall("sysctl"sv, errno);

StringView argv0 { argv[0], strlen(argv[0]) };
```

Libraries/LibCore/System.cpp
```diff
@@ -793,6 +793,48 @@ ErrorOr<ByteString> current_executable_path()
     size_t len = sizeof(path);
     if (sysctl(mib, 4, path, &len, nullptr, 0) < 0)
```
trflynn89, line 786: In the commit message: "I also that this is what many other projects" is not worded well

Meta/ladybird.py
```diff
@@ -197,7 +197,10 @@ def configure_main(platform: Platform, preset: str, cc: str, cxx: str) -> Path:
         return build_preset_dir
```
trflynn89, line 198: On the commit message: "to primitive" -> "too primitive"

Documentation/BuildInstructionsLadybird.md
```diff
@@ -212,6 +212,20 @@ Or, download a version of Gradle >= 8.0.0, and run the ``gradlew`` program in ``
 pkg install autoconf-archive automake autoconf bash cmake curl gmake gn libdrm libtool libxcb libxkbcommon libX11 libXrender libXi nasm ninja patchelf pkgconf python3 qt6-base qt6-multimedia unzip zip
 ```
 
+### OpenBSD
+
+#### CMake
+
+At the time of writing the OpenBSD ports version of CMake have patches applied that make it unsuitable for use with vcpkg (the ports version doesn't create namelinks, which are shared library symlinks without versions in their name). To build Ladybird you will need to build your own version of cmake from the official source without the OpenBSD patches.
```
trflynn89: Please wrap text lines at ~120 characters.

Meta/Lagom/Tools/CodeGenerators/LibWeb/GenerateWindowOrWorkerInterfaces.cpp
```diff
@@ -437,6 +438,12 @@ ErrorOr<int> ladybird_main(Main::Arguments arguments)
     VERIFY(!paths.is_empty());
     VERIFY(!base_paths.is_empty());
 
+#if !defined(AK_OS_WINDOWS)
+    // Increase the memory limit, because on OpenBSD the default limit is to low
+    if (auto result = Core::System::set_resource_limits(RLIMIT_DATA, 4 * GiB); result.is_error())
```
trflynn89: This seems to be causing our linux CI runners to fail.

Libraries/LibCore/System.cpp
cqundefine: On Windows this is actually implemented in `SystemWindows.cpp` and this file doesn't get compiled at all.

Libraries/LibCore/System.cpp
cqundefine: Yeah this works and is much cleaner.

Meta/CMake/vcpkg/overlay-ports/angle/portfile.cmake
cqundefine: I added a note about that flag breaking build on OpenBSD.

Meta/CMake/vcpkg/overlay-ports/skia/allow-disabling-lib-dl.patch
```diff
@@ -0,0 +1,28 @@
+diff --git a/BUILD.gn b/BUILD.gn
```
trflynn89: Question - how does this compare with #6599? Is this commit meant to be a full version update to match upstream's skia port, or would we be partway in between versions?

Meta/CMake/vcpkg/overlay-ports/angle/portfile.cmake
trflynn89: Ok, if this only affects BSD, can we limit it to executing on BSD systems? `if (BSD)` or some such.

Meta/Lagom/Tools/CodeGenerators/LibWeb/GenerateWindowOrWorkerInterfaces.cpp
```diff
@@ -437,6 +438,13 @@ ErrorOr<int> ladybird_main(Main::Arguments arguments)
     VERIFY(!paths.is_empty());
     VERIFY(!base_paths.is_empty());
 
+#if !defined(AK_OS_WINDOWS)
+    // Increase the memory limit, because on OpenBSD the default limit is to low
+    // FIXME: Figure out why when building with sanitizers this requires ridiculous limits, as it fails even with 64GiB
```
trflynn89, line 443: Also since this has only affected BSD systems so far, maybe let's limit this to FreeBSD+OpenBSD in scope?

cqundefine: FWIW, I limited it to only OpenBSD as this doesn't seem to affect FreeBSD.

Meta/CMake/vcpkg/overlay-ports/skia/allow-disabling-lib-dl.patch
cqundefine: I ended up dropping that patch, as it was only some changes from the upstream update, that PR is a cleaner solution to this and does the same thing when it comes to BSD support. I will mark the PR as draft until that gets merged.

cqundefine:
I found so more issues with the build on OpenBSD which led me to disabling shared libraries on that platform all together. It makes everything much simpler and work much better.

I'm also gonna undraft this PR as the changes here are no dependent on the skia update and while that PR still is needed for proper OpenBSD support, this can get merged independently of that PR.

PR #6591, opened 2025-10-26
LibWeb: Don't paint canvas objects with non-visible fill styles

tcl3:
Previously, calling `fillRect()` with a non-visible gradient fill style would paint a black rectangle.


Libraries/LibWeb/HTML/CanvasRenderingContext2D.cpp
```diff
@@ -373,6 +373,8 @@ void CanvasRenderingContext2D::stroke_internal(Gfx::Path const& path)
         return;
 
     auto& state = drawing_state();
+    if (!state.stroke_style.to_gfx_paint_style()->is_visible())
```
gmta: Can you extract these `.to_gfx_paint_style()` values into a separate variable so you can reuse them a couple lines down? No need to do the work twice.

PR #2375, opened 2024-11-16
LibWebView+QT: Support bangs in the URL bar

sakgoyal:
Allow "Bangs" in the URL bar to select a search engine directly through the browser. 
Additional Reference: https://duckduckgo.com/bangs

TODO: use an external list instead of putting it here. I just did it like this for now to simplify the PR/keep the change small for review. if people like this change, I can add a proper list of bangs in a config file

PR #6594, opened 2025-10-26
LibWebView+UI/QT: Support bangs in the URL bar

sakgoyal:
Allow "Bangs" in the URL bar to select a search engine directly through the browser. 

* Redo https://github.com/LadybirdBrowser/ladybird/pull/2375
* prioritize first bang over last bang https://github.com/LadybirdBrowser/ladybird/pull/2375#discussion_r1845698317

Notes: 
* existing database from ddg: https://duckduckgo.com/bang.js
* Additional Reference: https://duckduckgo.com/bangs

## TODO
* use an external list instead of hard coding it on the array. 
  * Currently just a PoC

### DIFF
--- a/Libraries/LibWebView/SearchEngine.cpp
+++ b/Libraries/LibWebView/SearchEngine.cpp
@@ -4,27 +4,46 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
+#include <AK/Find.h>
 #include <LibURL/URL.h>
 #include <LibWebView/SearchEngine.h>
 
 namespace WebView {
 
 static auto s_builtin_search_engines = to_array<SearchEngine>({
-    { "Bing"_string, "https://www.bing.com/search?q=%s"_string },
-    { "Brave"_string, "https://search.brave.com/search?q=%s"_string },
-    { "DuckDuckGo"_string, "https://duckduckgo.com/?q=%s"_string },
-    { "Ecosia"_string, "https://ecosia.org/search?q=%s"_string },
-    { "Google"_string, "https://www.google.com/search?q=%s"_string },
-    { "Kagi"_string, "https://kagi.com/search?q=%s"_string },
-    { "Mojeek"_string, "https://www.mojeek.com/search?q=%s"_string },
-    { "Startpage"_string, "https://startpage.com/search?q=%s"_string },
-    { "Yahoo"_string, "https://search.yahoo.com/search?p=%s"_string },
-    { "Yandex"_string, "https://yandex.com/search/?text=%s"_string },
+    { "Bing"_string, "https://www.bing.com/search?q=%s"_string, "b"_string },
+    { "Brave"_string, "https://search.brave.com/search?q=%s"_string, "brave"_string },
+    { "DuckDuckGo"_string, "https://duckduckgo.com/?q=%s"_string, "ddg"_string },
+    { "Ecosia"_string, "https://ecosia.org/search?q=%s"_string, "ec"_string },
+    { "Google"_string, "https://www.google.com/search?q=%s"_string, "g"_string },
+    { "Kagi"_string, "https://kagi.com/search?q=%s"_string, "kagi"_string },
+    { "Mojeek"_string, "https://www.mojeek.com/search?q=%s"_string, "mojeek"_string },
+    { "Startpage"_string, "https://startpage.com/search?q=%s"_string, "startpage"_string },
+    { "Yahoo"_string, "https://search.yahoo.com/search?p=%s"_string, "y"_string },
+    { "Yandex"_string, "https://yandex.com/search/?text=%s"_string, "ya"_string },
+    { "YouTube"_string, "https://www.youtube.com/results?search_query=%s"_string, "yt"_string },
 });
 
 ReadonlySpan<SearchEngine> builtin_search_engines()
 {
     return s_builtin_search_engines;
+}
+
+Optional<SearchEngine const&> find_search_engine_by_bang(String bang)
+{
+    bang = MUST(bang.substring_from_byte_offset(1));
+    if (bang.length_in_code_units() == 0)
+        return {};
+
+    auto it = AK::find_if(s_builtin_search_engines.begin(), s_builtin_search_engines.end(),
+        [&](auto const& engine) {
+            return engine.bang == bang;
+        });
+
+    if (it == s_builtin_search_engines.end())
+        return {};
+
+    return *it;
 }
 
 String SearchEngine::format_search_query_for_display(StringView query) const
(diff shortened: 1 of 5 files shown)


trflynn89:
Please don't open untested, proof-of-concept PRs.

sakgoyal:
I currently have the WSL compile running. it will take some time. You are right, I should have waited. but I did leave it as a draft PR for this reason

sakgoyal:
ok I ran some tests and it seemed to work as expected. 

I need to ensure it's a single commit

Should I expand the list in this PR or should I make a separate PR for implementing the bigger list from ddg?

and is there anything else I should do before marking this ready for review

Libraries/LibWebView/Settings.cpp
```diff
@@ -336,14 +337,18 @@ Optional<SearchEngine> Settings::parse_custom_search_engine(JsonValue const& sea
...
     if (!name.has_value() || !url.has_value())
         return {};
 
     auto parsed_url = URL::Parser::basic_parse(*url);
     if (!parsed_url.has_value())
         return {};
 
-    return SearchEngine { .name = name.release_value(), .query_url = url.release_value() };
+    if (!bang.has_value())
+        bang = "!unknownbang"_string;
```
AtkinsSJ: Seems like `bang` should be an `Optional<String>` instead. Then you also wouldn't need to modify `TestWebViewURL.cpp`.

UI/Qt/LocationEdit.cpp
```diff
@@ -41,7 +41,31 @@ LocationEdit::LocationEdit(QWidget* parent)
         auto ctrl_held = QApplication::keyboardModifiers() & Qt::ControlModifier;
         auto append_tld = ctrl_held ? WebView::AppendTLD::Yes : WebView::AppendTLD::No;
 
-        if (auto url = WebView::sanitize_url(query, WebView::Application::settings().search_engine(), append_tld); url.has_value())
+        auto const bang_char = '!'; // should this be a setting?
+
+        auto search_engine = WebView::Application::settings().search_engine();
+        if (search_engine.has_value()) {
```
AtkinsSJ: Why do we only allow bang searching when a default search engine is selected?

Libraries/LibWebView/SearchEngine.cpp
```diff
@@ -4,29 +4,44 @@
...
-    { "Brave"_string, "https://search.brave.com/search?q=%s"_string },
-    { "DuckDuckGo"_string, "https://duckduckgo.com/?q=%s"_string },
-    { "Ecosia"_string, "https://ecosia.org/search?q=%s"_string },
-    { "Google"_string, "https://www.google.com/search?q=%s"_string },
-    { "Kagi"_string, "https://kagi.com/search?q=%s"_string },
-    { "Mojeek"_string, "https://www.mojeek.com/search?q=%s"_string },
-    { "Startpage"_string, "https://startpage.com/search?q=%s"_string },
-    { "Yahoo"_string, "https://search.yahoo.com/search?p=%s"_string },
-    { "Yandex"_string, "https://yandex.com/search/?text=%s"_string },
+    { "Bing"_string, "https://www.bing.com/search?q=%s"_string, "!b"_string },
```
AtkinsSJ: Do we need to store the `!` in each of these? It'd be easier to make the bang character a setting if it's not hard-coded in the data here.

Libraries/LibWebView/Settings.cpp
sakgoyal: I used `Optional` and the code did not compile because the designatedd init was missing in the test file. so I would have to change it unfortunately.

Libraries/LibWebView/Settings.cpp
```diff
@@ -336,14 +337,17 @@ Optional<SearchEngine> Settings::parse_custom_search_engine(JsonValue const& sea
...
         return {};
 
     auto parsed_url = URL::Parser::basic_parse(*url);
     if (!parsed_url.has_value())
         return {};
 
-    return SearchEngine { .name = name.release_value(), .query_url = url.release_value() };
+    auto bangVal = bang.has_value() ? bang.release_value() : Optional<String> {}; // this doesnt feel like the best way
+
+    return SearchEngine { .name = name.release_value(), .query_url = url.release_value(), .bang = bangVal };
```
AtkinsSJ, line 350: I think you can just do `SearchEngine { /* ... */ .bang = move(bang) }` and don't need `bangVal` at all, unless I'm missing something.

Libraries/LibWebView/SearchEngine.cpp
```diff
@@ -4,29 +4,48 @@
...
 });
 
 ReadonlySpan<SearchEngine> builtin_search_engines()
 {
     return s_builtin_search_engines;
 }
 
+Optional<SearchEngine const&> find_search_engine_by_bang(String bang)
+{
+    bang = MUST(bang.substring_from_byte_offset(1));
```
AtkinsSJ, line 34: I'd feel better about this if it read the byte length of the bang prefix from somewhere instead of being hard-coded to 1.

Also if you don't reassign to `bang`, you can use `substring_from_byte_offset_with_shared_superstring()` which avoids copying the string data.

UI/Qt/LocationEdit.cpp
```diff
@@ -41,7 +41,31 @@ LocationEdit::LocationEdit(QWidget* parent)
...
+            }
+            if (auto const first = splits.value().first(); first.starts_with(bang_char)) {
+                auto exist = WebView::find_search_engine_by_bang(first);
+                if (exist.has_value()) {
+                    search_engine->name = exist->name;
+                    search_engine->query_url = exist->query_url;
+                    query = MUST(query.substring_from_byte_offset(first.bytes().size() + 1));
+                }
+            }
+        }
```
AtkinsSJ, line 66: This feels like it belongs somewhere else - all UIs should use the same bang logic. We handle parsing as either a URL or a search term in [sanitize_url()](https://github.com/LadybirdBrowser/ladybird/blob/0dd8e1e6f23ce15112e8ef5f4d540584d6baecbc/Libraries/LibWebView/URL.cpp#L17) so that might be better?

AtkinsSJ (changes requested):

Libraries/LibWebView/Settings.cpp
sakgoyal: That won't work unfortunately

image

AtkinsSJ: OK. Then please just rename `bangVal` to match our conventions.

You can probably use `bang.map()` to convert from `String const&` to `String`, that might be cleaner. But what you have is fine too. No need for the comment.

PR #6605, opened 2025-10-27
Everywhere: WIP - Coroutines instead of spin_until

alimpfard:
Let's break literally everything...in moderation.


R-Goc:
And the breaking starts with the github ui. On a more serious note this will need the windows label when it is no longer a draft. Next edition of breaking literally everything would be to replace notifiers with a completion based model using coroutines (maybe like what asio does).

alimpfard:
I don't want to switch every callback to coroutines, that is largely what AK::GenericAwaiter is for.
Also re draft status, I'm stuck trying to make sense of sync fetch, so this will probably take a while anyway.

PR #6608, opened 2025-10-27
LibWeb/CSS: Cascade and compute custom properties more like regular properties

AtkinsSJ:
We previously did a few weird things with custom properties that made them harder to work with:
- We computed them in the middle of cascading them.
- They were never inherited, instead you had to walk up the ancestor chain when trying to get one.
- We awkwardly stored them on Element and PseudoElement instead of alongside other cascaded/computed properties.
- We really didn't distinguish between cascaded or computed values, but used the same HashMap for both.

This PR moves those values onto CascadedProperties and ComputedProperties, and generally makes them more correct. There's still work to do, but hopefully follow-up work with be easier.

One thing I hoped to remove but didn't manage to, is that if a custom property is referred to by other custom properties, it gets computed repeatedly, one for each user. When I attempted to avoid that, I broke cycle-detection in some cases, and so I'm leaving it for now. There quite probably is a smart way of solving that.

### DIFF
--- a/Libraries/LibWeb/CSS/CSSStyleProperties.cpp
+++ b/Libraries/LibWeb/CSS/CSSStyleProperties.cpp
@@ -580,7 +580,7 @@ RefPtr<StyleValue const> CSSStyleProperties::style_value_for_computed_property(L
     }
 
     if (property.is_custom_property())
-        return owner_node()->get_custom_property(property.name());
+        return owner_node()->computed_properties()->property(property);
 
     auto property_id = property.id();
 
--- a/Libraries/LibWeb/CSS/StyleComputer.cpp
+++ b/Libraries/LibWeb/CSS/StyleComputer.cpp
@@ -3026,6 +3026,11 @@ NonnullRefPtr<StyleValue const> StyleComputer::resolve_custom_property_value(Pro
             return initial_value();
 
         // FIXME: revert, revert-layer.
+        dbgln("Here: {} (type {})", style_value->to_string(SerializationMode::Normal), style_value->type());
+        if (style_value->is_revert())
+            dbgln("!!! REVERT !!!");
+        if (style_value->is_revert_layer())
+            dbgln("!!! REVERT-LAYER !!!");
         return style_value;
     };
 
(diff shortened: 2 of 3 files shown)


Libraries/LibWeb/CSS/Parser/ValueParsing.cpp
```diff
@@ -4718,11 +4719,40 @@ RefPtr<FontSourceStyleValue const> Parser::parse_font_source_value(TokenStream<C
...
+        // https://drafts.csswg.org/css-values-5/#invalid-at-computed-value-time
+        // When substitution results in a property’s value containing the guaranteed-invalid value, this makes the
+        // declaration invalid at computed-value time. When this happens, the computed value is one of the
+        // following depending on the property’s type:
+
+        // -> The property is a non-registered custom property
+        // -> The property is a registered custom property with universal syntax
+        //    The computed value is the guaranteed-invalid value.
+        if (property.is_custom_property()) {
+            // The computed value is the guaranteed-invalid value.
```
Calme1709: Same comment here as two lines above

AtkinsSJ:
I ran WPT on the css dir for this:


Ran 30419 tests finished in 1730.9 seconds.
  • 30368 ran as expected. 0 tests skipped.
  • 6 tests crashed unexpectedly
  • 7 tests failed unexpectedly
  • 7 tests passed unexpectedly
  • 1 tests unexpectedly okay
  • 31 tests had unexpected subtest results

Some of these crashes and regressions seem legitimate so I'll draft this while I look into them.

AtkinsSJ:
Changes:
- Remove custom property storage from DOM::PseudoElement, which I missed before. (I'd committed this already but forgot to push it...)
- Resolve UnresolvedStyleValues in compute_logical_alias_mapping_context(), instead of crashing.

Remaining issues seem to be from not invalidating enough style when a custom property changes. I'll look into that on Monday when I'm more awake. :sweat_smile:

AtkinsSJ:
Imported some tests to track the regressions more easily:
- `--foo: var(--nonexistent, revert)` and the same with `revert-layer`, are broken, after the "LibWeb/CSS: Resolve UnresolvedStyleValues as part of computation" commit. I don't know how to make those work without returning to the original "jankily compute custom properties while cascading them" system.
- Modifying custom properties isn't updating style correctly, after the "LibWeb/CSS: Store custom properties on ComputedProperties" commit.

PR #6609, opened 2025-10-27
LibWeb: Produce containing absolute boxes with will-change properties

lpas:
Fixes 5 WPTs

### DIFF
--- a/Libraries/LibWeb/Layout/Node.cpp
+++ b/Libraries/LibWeb/Layout/Node.cpp
@@ -81,27 +81,37 @@ bool Node::can_contain_boxes_with_position_absolute() const
     if (!is<Box>(*this))
         return false;
 
-    if (computed_values().position() != CSS::Positioning::Static)
+    auto const& computed_values = this->computed_values();
+
+    if (computed_values.position() != CSS::Positioning::Static)
         return true;
 
     if (is<Viewport>(*this))
         return true;
 
+    // https://drafts.csswg.org/css-will-change/#will-change
+    // If any non-initial value of a property would cause the element to generate a containing block for absolutely
+    // positioned elements, specifying that property in will-change must cause the element to generate a containing
+    // block for absolutely positioned elements.
+    auto will_change_property = [&](CSS::PropertyID property_id) {
+        return computed_values.will_change().has_property(property_id);
+    };
+
     // https://w3c.github.io/csswg-drafts/css-transforms-1/#propdef-transform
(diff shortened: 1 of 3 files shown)


Libraries/LibWeb/Layout/Node.cpp
```diff
@@ -81,35 +81,45 @@ bool Node::can_contain_boxes_with_position_absolute() const
...
 
     if (is<Viewport>(*this))
         return true;
 
+    // https://drafts.csswg.org/css-will-change/#will-change
+    // If any non-initial value of a property would cause the element to generate a containing block for absolutely
+    // positioned elements, specifying that property in will-change must cause the element to generate a containing
+    // block for absolutely positioned elements.
+    auto will_change_property = [&](CSS::PropertyID property_id) {
+        return computed_values.will_change().has_property(property_id);
```
AtkinsSJ, line 97: This repeatedly calls `will_change()` which is quite expensive, and then does a linear search through its properties list.

One way to improve that would be to add some kind of `has_any_property(...)` method that takes a list of all the properties we care about. But that would mean checking the property and will-change in different places (plus `will-change` shouldn't have many entries usually). So I think the best thing to do would be to store `computed_values.will_change()` in a variable and use that.

lpas: I mirrored this after https://github.com/LadybirdBrowser/ladybird/blob/fb9286eccb3b8fd81086c22972c629548866d93c/Libraries/LibWeb/Layout/Node.cpp#L199-L204

If the `will_change()` call is expensive should I adjust this place as well?

tcl3: The call to `ComputedValues::will_change()` just returns a `WillChange` struct, so it's actually pretty quick (actually we could return it by `const&` to avoid unnecessary copies, but that's a separate issue.)

The reason I made `has_property()` do a linear search is that the underlying Vector will be empty or have only one or two values the vast majority of the time. If performance is a concern here we could look at using a HashTable or bitfields instead.

lpas: So should I do something here or do we keep this as is?

AtkinsSJ: Ohhh, I was looking at `ComputedProperties::will_change()` instead by mistake. 🤦‍♂️ Yeah, completely ignore what I said. 😅

PR #6612, opened 2025-10-28
LibWeb: Properly simplify sum nodes containing negated sum nodes

Calme1709:
This is ad-hoc, see https://github.com/w3c/csswg-drafts/issues/13020

Gains us 5 WPT tests

### DIFF
--- a/Libraries/LibWeb/CSS/StyleValues/CalculatedStyleValue.cpp
+++ b/Libraries/LibWeb/CSS/StyleValues/CalculatedStyleValue.cpp
@@ -3070,6 +3070,22 @@ NonnullRefPtr<CalculationNode const> simplify_a_calculation_tree(CalculationNode
         if (child.type() == CalculationNode::Type::Negate)
             return as<NegateCalculationNode>(child).child();
 
+        // AD-HOC: Convert negated sums into sums of negated nodes - see https://github.com/w3c/csswg-drafts/issues/13020
+        if (child.type() == CalculationNode::Type::Sum) {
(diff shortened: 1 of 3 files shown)


Libraries/LibWeb/CSS/StyleValues/CalculatedStyleValue.cpp
```diff
@@ -3070,6 +3070,20 @@ NonnullRefPtr<CalculationNode const> simplify_a_calculation_tree(CalculationNode
...
 
+        // AD-HOC: Convert negated sums into sums of negated nodes - see https://github.com/w3c/csswg-drafts/issues/13020
+        if (child.type() == CalculationNode::Type::Sum) {
+            Vector<NonnullRefPtr<CalculationNode const>> negated_sum_components;
+
+            for (auto const& sum_child : child.children()) {
+                if (sum_child->type() == CalculationNode::Type::Numeric)
+                    negated_sum_components.append(as<NumericCalculationNode>(*sum_child).negated(context));
+                else
+                    negated_sum_components.append(NegateCalculationNode::create(sum_child));
```
AtkinsSJ, line 3083: Hmm... If `sum_child` is already a negate node that couldn't be simplified earlier, we'll wrap a negate with a negate here I think? I'm not sure if that can happen or not.

Calme1709: Good catch, it definitely can happen i.e. `calc(100 - (50 - sign(1em - 1px)))` was being simplified to `calc(50 - (-1 * sign(1em - 1px)))`

Have pushed a fix.

PR #6615, opened 2025-10-28
IDLGenerators: Allow set_attribute's to throw exceptions

tete17:
This should fix the crashing errors recently introduced in the trusted-types folder.

I have also added a subset of the trusted-types tests that were previously crashing. The rest now hang again due to our Webhooks implementation lacking integration with our fetch implementation, but that is a problem for a different time.

Some of the test fail but I have a strong suspicion they are not 100% up to spec. That again if for a different time. For now they serve as a way not to make sure they don't crash on us on future changes.


AtkinsSJ (approved):
:+1:

PR #6619, opened 2025-10-28
LibWeb/IDB: Correctly implement IDBRecord getters

stelar7:
Gives ~75 WPT passes when combined with the worker tests

Thanks @trflynn89 @Lubrsi for the pointers 🙂

### DIFF
--- a/Tests/LibWeb/Text/expected/wpt-import/IndexedDB/idbindex_getAllRecords.any.txt
+++ b/Tests/LibWeb/Text/expected/wpt-import/IndexedDB/idbindex_getAllRecords.any.txt
@@ -1,0 +1,31 @@
+Harness status: OK
+
+Found 25 tests
+
+24 Pass
+1 Fail
+Pass	Single item
+Pass	Empty index
+Pass	Get all records
+Pass	Get all records with empty options
+Pass	Get all records with large value
+Pass	Count
+Pass	Query with bound range
+Pass	Query with bound range and count
+Pass	Query with upper excluded bound range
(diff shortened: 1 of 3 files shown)


Lubrsi:
Would it be possible to import those WPT tests? :^)

Libraries/LibWeb/IndexedDB/IDBRecord.h
```diff
@@ -10,6 +10,7 @@
 #include <LibWeb/Bindings/Intrinsics.h>
 #include <LibWeb/Bindings/PlatformObject.h>
 #include <LibWeb/HTML/StructuredSerializeTypes.h>
+#include <LibWeb/IndexedDB/Internal/Algorithms.h>
```
trflynn89: Seems like this should be in the .cpp file, no need to pollute the header

stelar7: Seems like it was forgotten in my moving it.. Fixed 👍

Tests/LibWeb/Text/expected/wpt-import/IndexedDB/idbindex_getAllRecords.any.txt
```diff
@@ -0,0 +1,31 @@
+Harness status: OK
```
trflynn89, line 1: This is fine, but in the future, it makes more sense for the tests and the code being tested to be in the same commit IMO.

trflynn89 (approved):

PR #6623, opened 2025-10-28
AK: Increase MonotonicTime precision on Windows

R-Goc:
MonotonicTime was using 32 bit floats for operations on the values converted from an i64 returned by the performance counter. This caused to either precision losses or complete losses of the data for timing even things as long as hundreds of miliseconds, which should never be a problem with the resolution of the performance counter. This change fixes that behavior.

### DIFF
--- a/AK/Time.cpp
+++ b/AK/Time.cpp
@@ -248,17 +248,16 @@ Duration now_time_from_filetime()
 Duration now_time_from_query_performance_counter()
 {
     static LARGE_INTEGER ticks_per_second;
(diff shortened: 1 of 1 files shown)


AK/Time.cpp
```diff
@@ -249,16 +249,16 @@ Duration now_time_from_query_performance_counter()
 {
     static LARGE_INTEGER ticks_per_second;
     // FIXME: Limit to microseconds for now, but could probably use nanos?
```
gmta: Didn't you resolve this `FIXME` now?

R-Goc: right

PR #6625, opened 2025-10-29
RequestServer+LibRequests: Create abstractions for request data transfer

ayeteadoe:
Split out from https://github.com/LadybirdBrowser/ladybird/pull/5981 as there has been some activity recently in RequestServer-land causing consistent merge conflicts, so having this upstreamed now will make it easier to maintain that branch while it is waiting on review.

These abstractions were requested in https://github.com/LadybirdBrowser/ladybird/pull/5981#discussion_r2296852733 and https://github.com/LadybirdBrowser/ladybird/pull/5981#discussion_r2296852925

### DIFF
--- a/Libraries/LibRequests/Request.cpp
+++ b/Libraries/LibRequests/Request.cpp
@@ -9,6 +9,20 @@
 
 namespace Requests {
 
+ErrorOr<NonnullOwnPtr<ReadStream>> ReadStream::create(int reader_fd)
+{
+#if defined(AK_OS_WINDOWS)
+    auto local_socket = TRY(Core::LocalSocket::adopt_fd(reader_fd));
+    auto notifier = local_socket->notifier();
+    VERIFY(notifier);
+    return adopt_own(*new ReadStream(move(local_socket), notifier.release_nonnull()));
+#else
+    auto file = TRY(Core::File::adopt_fd(reader_fd, Core::File::OpenMode::Read));
+    auto notifier = Core::Notifier::construct(reader_fd, Core::Notifier::Type::Read);
+    return adopt_own(*new ReadStream(move(file), move(notifier)));
+#endif
+}
+
 Request::Request(RequestClient& client, i32 request_id)
     : m_client(client)
     , m_request_id(request_id)
@@ -37,11 +51,11 @@ void Request::set_request_fd(Badge<Requests::RequestClient>, int fd)
     VERIFY(m_fd == -1);
     m_fd = fd;
 
-    auto notifier = Core::Notifier::construct(fd, Core::Notifier::Type::Read);
-    auto stream = MUST(Core::File::adopt_fd(fd, Core::File::OpenMode::Read));
+    auto read_stream = MUST(ReadStream::create(fd));
+    auto notifier = read_stream->notifier();
     notifier->on_activation = move(m_internal_stream_data->read_notifier->on_activation);
-    m_internal_stream_data->read_notifier = move(notifier);
-    m_internal_stream_data->read_stream = move(stream);
+    m_internal_stream_data->read_notifier = notifier;
+    m_internal_stream_data->read_stream = move(read_stream);
 }
 
 void Request::set_buffered_request_finished_callback(BufferedRequestFinished on_buffered_request_finished)
@@ -117,7 +131,7 @@ void Request::set_up_internal_stream_data(DataReceived on_data_available)
     m_internal_stream_data = make<InternalStreamData>();
     m_internal_stream_data->read_notifier = Core::Notifier::construct(fd(), Core::Notifier::Type::Read);
     if (fd() != -1)
-        m_internal_stream_data->read_stream = MUST(Core::File::adopt_fd(fd(), Core::File::OpenMode::Read));
+        m_internal_stream_data->read_stream = MUST(ReadStream::create(fd()));
 
     auto user_on_finish = move(on_finish);
     on_finish = [this](auto total_size, auto const& timing_info, auto network_error) {
--- a/Libraries/LibRequests/Request.h
+++ b/Libraries/LibRequests/Request.h
@@ -21,6 +21,27 @@ namespace Requests {
 
 class RequestClient;
 
+class ReadStream {
+public:
+    static ErrorOr<NonnullOwnPtr<ReadStream>> create(int reader_fd);
+
+    NonnullRefPtr<Core::Notifier> const& notifier() const { return m_notifier; }
+
+    bool is_eof() const { return m_stream->is_eof(); }
+
+    ErrorOr<Bytes> read_some(Bytes bytes) { return m_stream->read_some(bytes); }
+
+private:
+    ReadStream(NonnullOwnPtr<Stream> stream, NonnullRefPtr<Core::Notifier> notifier)
+        : m_stream(move(stream))
+        , m_notifier(move(notifier))
+    {
+    }
+
+    NonnullOwnPtr<Stream> m_stream;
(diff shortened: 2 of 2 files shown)


Services/RequestServer/ConnectionFromClient.h
```diff
@@ -12,6 +12,7 @@
 #include <LibWebSocket/WebSocket.h>
 #include <RequestServer/Forward.h>
 #include <RequestServer/RequestClientEndpoint.h>
+#include <RequestServer/RequestPipe.h>
```
trflynn89: Is this needed here?

Services/RequestServer/RequestPipe.h
```diff
@@ -0,0 +1,37 @@
...
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#pragma once
+
+#include <AK/RefCounted.h>
+
+namespace RequestServer {
+
+class RequestPipe : public RefCounted<RequestPipe> {
```
trflynn89: Does this need to be ref-counted? Seems single owner to me, and seems like it can just live on the stack (instead of being heap-allocated)

Services/RequestServer/RequestPipe.h
```diff
@@ -0,0 +1,37 @@
...
+public:
+    static ErrorOr<NonnullRefPtr<RequestPipe>> try_create();
```
trflynn89: No need for the `try_` prefix on this factory name

Services/RequestServer/RequestPipe.h
```diff
@@ -0,0 +1,37 @@
...
+
+    ~RequestPipe();
+
+    int reader_fd() const { return m_reader_fd; }
+    int writer_fd() const { return m_writer_fd; }
+
+    ErrorOr<ssize_t> write(Vector<u8> const& bytes);
```
trflynn89: Let's take this parameter as:
```c++
ErrorOr<ssize_t> write(ReadonlyBytes bytes);
```

Services/RequestServer/Request.cpp
```diff
@@ -510,6 +507,18 @@ size_t Request::on_data_received(void* buffer, size_t size, size_t nmemb, void*
     return total_size;
 }
 
+int Request::client_writer_fd() const
+{
+    VERIFY(m_client_request_pipe);
+    return m_client_request_pipe->writer_fd();
+}
+
+ErrorOr<ssize_t> Request::write_bytes_to_client(Vector<u8> const& bytes) const
```
trflynn89: Same here re: `ReadonlyBytes`.

Though I reckon we don't need either of these methods if we just keep `RequestPipe` on the stack. We can just have an `Optional<RequestPipe>` I think.

Libraries/LibRequests/Request.h
```diff
@@ -21,6 +21,27 @@ namespace Requests {
 
 class RequestClient;
 
+class ReadStream {
+public:
+    static ErrorOr<NonnullOwnPtr<ReadStream>> try_create(int reader_fd);
```
trflynn89: Don't need the `try_` prefix

Libraries/LibRequests/Request.h
```diff
@@ -21,6 +21,27 @@ namespace Requests {
...
+
+    RefPtr<Core::Notifier> notifier() const { return m_notifier; }
```
trflynn89: Let's return by `const&` - no need to force a ref count increase unless the caller needs it

Libraries/LibRequests/Request.cpp
```diff
@@ -37,11 +52,11 @@ void Request::set_request_fd(Badge<Requests::RequestClient>, int fd)
     VERIFY(m_fd == -1);
     m_fd = fd;
 
-    auto notifier = Core::Notifier::construct(fd, Core::Notifier::Type::Read);
-    auto stream = MUST(Core::File::adopt_fd(fd, Core::File::OpenMode::Read));
+    auto read_stream = MUST(ReadStream::try_create(fd));
+    auto notifier = read_stream->notifier();
```
trflynn89, line 55: We previously had a `NonnullRefPtr<Notifier>` here by way of `Notifier::construct`. We now have a `RefPtr` that we're assuming is nonnull. 

If we're going to do that, let's have `ReadStream` hold a `NonnullRefPtr` to begin with. The Windows factory would then look like:

```c++
ErrorOr<NonnullOwnPtr<ReadStream>> ReadStream::try_create(int const reader_fd)
{
    auto local_socket = TRY(Core::LocalSocket::adopt_fd(reader_fd));

    auto notifier = local_socket->notifier();
    VERIFY(notifier);

    return adopt_own(*new ReadStream(move(local_socket), notifier.release_nonnull()));
}
```

Libraries/LibRequests/Request.cpp
```diff
@@ -9,6 +9,21 @@
 
 namespace Requests {
 
+#if defined(AK_OS_WINDOWS)
+ErrorOr<NonnullOwnPtr<ReadStream>> ReadStream::try_create(int const reader_fd)
```
trflynn89: style nit: Let's put the Windows guards inside the function body, so we don't need to repeat the function declarations.

style nit 2: Making the `reader_fd` parameter const just adds noise, IMO.

Libraries/LibRequests/Request.h
```diff
@@ -21,6 +21,27 @@ namespace Requests {
...
+
+    bool is_eof() const { return m_stream->is_eof(); }
+
+    ErrorOr<Bytes> read_some(Bytes bytes) { return m_stream->read_some(bytes); }
+
+private:
+    explicit ReadStream(NonnullOwnPtr<Stream> stream, RefPtr<Core::Notifier> notifier)
```
trflynn89: `explicit` keyword does nothing here

Services/RequestServer/Request.cpp
ayeteadoe: Yeah I can remove these as `Optional::value()` already does the `VERIFY` check

Libraries/LibRequests/Request.h
ayeteadoe: In `Request::set_request_fd` we need to set the notifiers `on_activation`, so returning by `&` seems necessary here

Services/RequestServer/RequestPipe.h
ayeteadoe: One thing I needed to do now that RequestPipe has value semantics is remove the dtor and switch to an explicit close() method called in the `RequestServer::Request` dtor like was done previously. Otherwise due to destructive moves we will close the writer fd right before we invoke `async_request_started` and several test-web tests start to fail

Libraries/LibRequests/Request.h
trflynn89: That is confusing `RefPtr<T> const&` with `RefPtr<T const>`. You will still be able to set the activation callback on the former, which is what I'm suggesting:
```c++
RefPtr<Core::Notifier> const& notifier() const { return m_notifier; }
```

You can do this without affecting your current call site at all.

Services/RequestServer/RequestPipe.h
trflynn89: The standard practice here is to write a move constructor, and delete the copy constructor (we have a macro for the latter in AK/Noncopyable.h). The move constructor sets the moved-from object's fd's to -1, and the destructor then checks for -1 before closing.

Libraries/LibRequests/Request.h
ayeteadoe: Ah yeah my bad, thought you meant to return `Core::Notifier const&`

PR #6627, opened 2025-10-29
UI/AppKit: Support dead keys on MacOS

### LINKED ISSUE #6624: Dead keys support
When the keyboard layout is set to one supporting dead keys (such as "U.S. International - PC") on macOS, we seem to receive key events with the key code set, but no code point (since `event.characters` is empty) for dead keys such as the apostrophe or tilde. Other browsers seem to have explicit handling of dead keys where they show a temporary underlined character in input fields before committing it (e.g. after pressing space) or replacing it with an accented character (like "é").

**Reproduction**
1. Have keyboard layout set to U.S. International - PC
2. Go to any input field inside Ladybird
3. Press apostrophe, quote, tilde or backtick on your keyboard
4. -> Nothing happens

tcl3:
We now pass key press events to `interpretKeyEvents()`, which handles dead key composition for us.

This change brings gives us equivalent dead key functionality on MacOS and Linux. We don't currently support displaying the appropriate pre-edit text in either UI.

Please bear in mind that I have very little idea what I'm doing with MacOS stuff, so there may be a better way to do this.

Tested with a Spanish keyboard layout, as that's what I had set up on my Mac.

Fixes #6624

### DIFF
--- a/UI/AppKit/Interface/LadybirdWebView.mm
+++ b/UI/AppKit/Interface/LadybirdWebView.mm
@@ -321,7 +321,7 @@ struct HideCursor {
         if (self == nil) {
             return;
         }
-        NSEvent* event = Ladybird::key_event_to_ns_event(key_event);
+        auto* event = Ladybird::key_event_to_ns_event(key_event);
 
         self.event_being_redispatched = event;
         [NSApp sendEvent:event];
@@ -1072,8 +1072,7 @@ struct HideCursor {
         return;
     }
 
-    auto key_event = Ladybird::ns_event_to_key_event(Web::KeyEvent::Type::KeyDown, event);
-    m_web_view_bridge->enqueue_input_event(move(key_event));
+    [self interpretKeyEvents:@[ event ]];
 }
 
 - (void)keyUp:(NSEvent*)event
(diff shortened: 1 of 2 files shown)


UI/AppKit/Interface/LadybirdWebView.mm
```diff
@@ -1072,8 +1072,78 @@ - (void)keyDown:(NSEvent*)event
         return;
     }
 
-    auto key_event = Ladybird::ns_event_to_key_event(Web::KeyEvent::Type::KeyDown, event);
-    m_web_view_bridge->enqueue_input_event(move(key_event));
+    [self interpretKeyEvents:@[ event ]];
+}
+
+- (void)insertText:(id)string replacementRange:(NSRange)replacementRange
```
trflynn89: Organizationally, instead of putting these methods in between the `keyDown` and `keyUp` overrides, let's move them to their own section below prefixed with a mark:

```c++
#pragma mark - NSTextInputClient
```

The mark just makes it clearer which methods belong to which interface, and marks show up in code navigation in some IDEs

UI/AppKit/Interface/LadybirdWebView.mm
```diff
@@ -1072,8 +1072,78 @@ - (void)keyDown:(NSEvent*)event
...
+- (void)doCommandBySelector:(SEL)selector
+{
+    NSEvent* event = [NSApp currentEvent];
+    if (event && event.type == NSEventTypeKeyDown) {
+        auto key_event = Ladybird::ns_event_to_key_event(Web::KeyEvent::Type::KeyDown, event);
+        m_web_view_bridge->enqueue_input_event(move(key_event));
+    }
+}
+
+- (BOOL)acceptsFirstResponder
```
trflynn89: (Though some of these don't seem to be `NSTextInputClient` thus should be in their own mark - looks like `NSResponder` here)

UI/AppKit/Interface/LadybirdWebView.mm
```diff
@@ -1072,8 +1072,78 @@ - (void)keyDown:(NSEvent*)event
...
+{
+    return NSNotFound;
+}
+
+- (NSRect)firstRectForCharacterRange:(NSRange)range actualRange:(NSRangePointer)actualRange
+{
+    return NSZeroRect;
 }
 
 - (void)keyUp:(NSEvent*)event
```
trflynn89: Do we need to hook `keyUp` into this infra somehow? What do the JS events look like compared to Chrome or Firefox?

tcl3: Given this example page:
```html
<!DOCTYPE html>
<textarea rows=20 cols=30></textarea>
<script>
    const textarea = document.querySelector("textarea");
    textarea.addEventListener("keyup", (e) => {
        console.log(`KeyDown event key: ${e.key}`);
    });
    textarea.addEventListener("keydown", (e) => {
        console.log(`KeyUp event key: ${e.key}`);
    });
</script>
```

We get the following output when typing an ` + e dead key combination:

Qt (on Linux):
```
KeyDown event key: Unidentified"
KeyUp event key: è"
KeyDown event key: e"
```

AppKit:
```
KeyDown event key: `"
KeyUp event key: è"
KeyDown event key: e"
```

Compared to other engines:

Firefox (on Linux)
```
KeyUp event key: Dead
KeyUp event key: e
KeyDown event key: Dead
KeyDown event key: e
```

Chromium (on Linux)
```
KeyUp event key: Process
KeyUp event key: Process
KeyDown event key: Dead
KeyDown event key: e
```

Firefox and Chrome (on Mac)
```
KeyUp event key: Dead
KeyUp event key: è
KeyDown event key: Dead
KeyDown event key: e
```

It looks like we should be reporting the dead key on both keyUp and keyDown, although exact behavior is obviously not consistent across platforms / browsers.

I tried using `interpretKeyEvents()` for the `keyUp` event, but I ended up with strangely inconsistent results. Given that our Qt and Appkit implementations are currently pretty similar, I'm tempted to leave it as-is for now and improve both Qt and AppKit behavior in a future PR.

trflynn89: Wow that varies way more than I would have expected. But handling dead key reporting in a future PR sounds good to me.

gmta:
@tcl3 I didn't get to test this properly before it got merged, but there's this issue when trying to combine keys that can't be combined:

"~" + "n" = "ñ" (works)

"~" + "e" = "~~" (should be "~e")

PR #6630, opened 2025-10-29
Trusted Types spec update

tete17:
These changes should cover the recent changes on the spec I didn't pick up on my initial PR's

Don't be scared for the number of lines changes 99% of them are from importing passing WPT's

All that is left is to check each of the failing WPT tests understand the failures and fix any of the bugs, either:
 * in our code
 * in the spec
 * in the WPT

:)

### DIFF
--- a/Libraries/LibWeb/DOM/Attr.cpp
+++ b/Libraries/LibWeb/DOM/Attr.cpp
@@ -70,41 +70,34 @@ void Attr::set_owner_element(Element* owner_element)
     m_owner_element = owner_element;
 }
 
-// FIXME: Trusted Types integration with DOM is still under review https://github.com/whatwg/dom/pull/1268
-// https://whatpr.org/dom/1268.html#set-an-existing-attribute-value
+// https://dom.spec.whatwg.org/#set-an-existing-attribute-value
 WebIDL::ExceptionOr<void> Attr::set_value(String value)
 {
-    // 1. If attribute’s element is null, then set attribute’s value to value.
+    // 1. If attribute’s element is null, then set attribute’s value to value and return.
     if (!owner_element()) {
         m_value = move(value);
+        return {};
     }
-    // 2. Otherwise:
-    else {
-        // 1. Let element be attribute’s element.
-        auto const* element = owner_element();
 
-        // 2. Let verifiedValue be the result of calling get Trusted Types-compliant attribute value with
-        //    attribute’s local name, attribute’s namespace, element, and value.
-        auto const verified_value = TRY(TrustedTypes::get_trusted_types_compliant_attribute_value(
-            local_name(),
-            namespace_uri().has_value() ? Utf16String::from_utf8(namespace_uri().value()) : Optional<Utf16String>(),
-            *element,
-            Utf16String::from_utf8(value)));
+    // 2. Let element be attribute’s element.
+    auto const& element = *owner_element();
 
-        // 3. If attribute’s element is null, then set attribute’s value to verifiedValue, and return.
-        if (!owner_element()) {
-            m_value = verified_value.to_utf8_but_should_be_ported_to_utf16();
-            return {};
-        }
+    // 3. Let verifiedValue be the result of calling get Trusted Types-compliant attribute value with
+    //    attribute’s local name, attribute’s namespace, element, and value.
+    auto const verified_value = TRY(TrustedTypes::get_trusted_types_compliant_attribute_value(
+        local_name(),
+        namespace_uri().has_value() ? Utf16String::from_utf8(namespace_uri().value()) : Optional<Utf16String>(),
+        element,
+        Utf16String::from_utf8(value)));
 
-        // 4. If attribute’s element is not element, then return.
-        if (owner_element() != element) {
-            return {};
-        }
-
-        // 5. Change attribute to verifiedValue.
-        change_attribute(verified_value.to_utf8_but_should_be_ported_to_utf16());
+    // 4. If attribute’s element is null, then set attribute’s value to verifiedValue, and return.
+    if (!owner_element()) {
+        m_value = verified_value.to_utf8_but_should_be_ported_to_utf16();
+        return {};
     }
+
+    // 5. Change attribute to verifiedValue.
+    change_attribute(verified_value.to_utf8_but_should_be_ported_to_utf16());
 
     return {};
 }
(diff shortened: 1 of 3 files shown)


Libraries/LibWeb/TrustedTypes/TrustedTypePolicy.h
```diff
@@ -71,6 +71,12 @@ WebIDL::ExceptionOr<Utf16String> get_trusted_type_compliant_string(TrustedTypeNa
 
 WebIDL::ExceptionOr<Utf16String> get_trusted_types_compliant_attribute_value(FlyString const& attribute_name, Optional<Utf16String> attribute_ns, DOM::Element const& element, Variant<GC::Root<TrustedHTML>, GC::Root<TrustedScript>, GC::Root<TrustedScriptURL>, Utf16String> const& new_value);
 
-Utf16String element_interface_name(Utf16String const& local_name, Utf16String const& element_ns);
+// FIXME: Add-hoc definition of an element interface
+struct ElementInterface {
+    Utf16String element_name;
+    Utf16String element_ns;
```
gmta: We could at least make these `FlyString`s, since that is what `Web::Namespace`, `Web::HTML::TagNames` etc. use as well. Then you also need to convert a bit less.

Libraries/LibWeb/TrustedTypes/TrustedTypePolicy.cpp
gmta, line 1: Commit message: WPT's -> WPT subtests

Libraries/LibWeb/HTML/HTMLScriptElement.cpp
```diff
@@ -82,6 +82,31 @@ void HTMLScriptElement::attribute_changed(FlyString const& name, Optional<String
     }
 }
 
+// https://www.w3.org/TR/trusted-types/#prepare-script-text
+WebIDL::ExceptionOr<void> HTMLScriptElement::prepare_script_text()
+{
+    // 1. Let sink be "HTMLScriptElement text" if script is an HTMLScriptElement; otherwise "SVGScriptElement text".
+    constexpr auto sink = TrustedTypes::InjectionSink::HTMLScriptElement_text;
```
gmta, line 89: I'm aware that these were introduced in a previous PR, but this is an interesting mix of our variable naming styles 😅 I'd expect either `html_script_element_text` or `HTMLScriptElementText`.

Libraries/LibWeb/DOM/Attr.cpp
```diff
@@ -70,42 +70,35 @@ void Attr::set_owner_element(Element* owner_element)
...
-        // 4. If attribute’s element is not element, then return.
-        if (owner_element() != element) {
-            return {};
-        }
-
-        // 5. Change attribute to verifiedValue.
-        change_attribute(verified_value.to_utf8_but_should_be_ported_to_utf16());
+
+    // 2. Let element be attribute’s element.
+    auto const* element = owner_element();
```
gmta: You dereference this variable a couple of lines down, so you should be able to make this an `auto const&`

Libraries/LibWeb/DOM/Attr.cpp
```diff
@@ -70,42 +70,35 @@ void Attr::set_owner_element(Element* owner_element)
...
+    // 3. Let verifiedValue be the result of calling get Trusted Types-compliant attribute value with
+    //    attribute’s local name, attribute’s namespace, element, and value.
+    auto const verified_value = TRY(TrustedTypes::get_trusted_types_compliant_attribute_value(
+        local_name(),
+        namespace_uri().map<Utf16String (*)(FlyString const&)>(Utf16String::from_utf8),
+        *element,
+        Utf16String::from_utf8(value)));
+
+    // 4. If attribute’s element is null, then set attribute’s value to verifiedValue, and return.
+    if (!owner_element()) {
```
gmta, line 94: How can this ever be true after passing steps 1 through 3? 🤔

Libraries/LibWeb/DOM/Attr.cpp
```diff
@@ -70,42 +70,35 @@ void Attr::set_owner_element(Element* owner_element)
...
-        change_attribute(verified_value.to_utf8_but_should_be_ported_to_utf16());
+
+    // 2. Let element be attribute’s element.
+    auto const* element = owner_element();
+
+    // 3. Let verifiedValue be the result of calling get Trusted Types-compliant attribute value with
+    //    attribute’s local name, attribute’s namespace, element, and value.
+    auto const verified_value = TRY(TrustedTypes::get_trusted_types_compliant_attribute_value(
+        local_name(),
+        namespace_uri().map<Utf16String (*)(FlyString const&)>(Utf16String::from_utf8),
```
gmta: I much prefer your old syntax of converting the namespace URI to UTF-16.

Libraries/LibWeb/HTML/HTMLScriptElement.cpp
tete17: Yeah I was never convinced ever since I introduced them. We changed them recently in another MR https://github.com/LadybirdBrowser/ladybird/pull/6632

Happy to change them but I wonder if we are running in circles since non of the options are pretty

gmta: Not really, Sam mentioned being annoyed by the "srcdoc" part which I agree with since that should probably have been "Srcdoc" 😛 unless there's a very good reason to diverge, let's just use one of the two styles. Personally I'd go with the CamelCase version.

Libraries/LibWeb/DOM/Attr.cpp
tete17: This deserves a issue in the spec

Libraries/LibWeb/DOM/Attr.cpp
tete17: I thought I was being clever :)

Libraries/LibWeb/DOM/Attr.cpp
tete17: I made https://github.com/whatwg/dom/issues/1421 and added a comment in our code

tete17: I forgot that `get_trusted_types_compliant_attribute_value` can indeed run arbitrary code if the default policy is invoked. So I think this is indeed needed.

Libraries/LibWeb/HTML/HTMLScriptElement.cpp
gmta: Coming back to this, I think I've seen other places where we use `::CamelCase_somethingElse`, so let's drop this for now.

PR #6631, opened 2025-10-29
Everywhere: Cleanup some platform code and remove unused code

cqundefine:
This is basically a reopened #5839 without the big commit that removes SerenityOS code, as that should probably be a PR of it's own. Everything removed here is basically entirely unused / relics from SerenityOS days that are now no-ops.

I have not tried removing support for any platforms (well, except Emscripten but that's a different case), as in my own testing I found that fixing support for those other platforms won't be as bad as I thought - I already have Ladybird working on FreeBSD, OpenBSD, I also have it building for NetBSD and Haiku, but those builds are not fully functional.

### DIFF
--- a/AK/String.h
+++ b/AK/String.h
@@ -26,15 +26,6 @@
 
 namespace AK {
 
-// FIXME: Remove this when OpenBSD Clang fully supports consteval.
-//        And once oss-fuzz updates to clang >15.
-//        And once Android ships an NDK with clang >14
-#if defined(AK_OS_OPENBSD) || defined(OSS_FUZZ) || defined(AK_OS_ANDROID)
-#    define AK_SHORT_STRING_CONSTEVAL constexpr
-#else
-#    define AK_SHORT_STRING_CONSTEVAL consteval
-#endif
-
 // String is a strongly owned sequence of Unicode code points encoded as UTF-8.
 // The data may or may not be heap-allocated, and may or may not be reference counted.
 // There is no guarantee that the underlying bytes are null-terminated.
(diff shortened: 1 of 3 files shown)


diversys:
> I also have it building for NetBSD and Haiku, but those builds are not fully functional.

Hi, did you manage to port vcpkg to Haiku? If yes, do you happen to have patches?

cqundefine:
> > I also have it building for NetBSD and Haiku, but those builds are not fully functional.
> 
> Hi, did you manage to port vcpkg to Haiku? If yes, do you happen to have patches?

Yeah I kinda did a while ago, I never got around to upstreaming it because for it to build all the Ladybird dependencies I had to hold it together with glue and tape. It had a bunch of issues like the fact that vcpkg expects all CMake configs to be in `share` while Haiku patches it to be `data`. Also a bunch of dependencies needed patches. If you want I can share my diff of both `vcpkg-tool` and `vcpkg`, though don't expect it to just work :).

diversys:
@cqundefine Yes, please share the diff.

diversys:
@cqundefine ping :)

cqundefine:
@diversys Sorry I didn't reply earlier, got busy with other stuff and then simply forgot, I pushed my work in all of the 3 respective repos here:
vcpkg: https://github.com/cqundefine/vcpkg/tree/haiku_dump
vcpkg-tool: https://github.com/cqundefine/vcpkg-tool/tree/haiku_dump
ladybird: https://github.com/cqundefine/ladybird/tree/haiku_dump

A bunch of this work is from quite a while ago so rebasing this might be quite tricky, but overall this is like 90% of the changes that I needed for that to build at that time. Some manual changes to the vcpkg install directory were still required unfortunately.

PR #6632, opened 2025-10-29
LibWeb: Avoid invoking Trusted Types where avoidable

Lubrsi:
Prevents observably calling Trusted Types, which can run arbitrary JS, cause crashes due to use of MUST and allow arbitrary JS to modify internal elements.

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLIFrameElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLIFrameElement.cpp
@@ -19,8 +19,8 @@
 #include <LibWeb/HTML/Parser/HTMLParser.h>
 #include <LibWeb/HTML/TraversableNavigable.h>
 #include <LibWeb/Layout/NavigableContainerViewport.h>
-#include <LibWeb/TrustedTypes/TrustedTypePolicy.h>
 #include <LibWeb/TrustedTypes/RequireTrustedTypesForDirective.h>
+#include <LibWeb/TrustedTypes/TrustedTypePolicy.h>
 
 namespace Web::HTML {
 
--- a/Libraries/LibWeb/ARIA/ARIAMixin.h
+++ b/Libraries/LibWeb/ARIA/ARIAMixin.h
@@ -34,7 +34,7 @@ public:
 
 #define __ENUMERATE_ARIA_ATTRIBUTE(name, attribute) \
     virtual Optional<String> name() const = 0;      \
-    virtual WebIDL::ExceptionOr<void> set_##name(Optional<String> const&) = 0;
+    virtual void set_##name(Optional<String> const&) = 0;
     ENUMERATE_ARIA_ATTRIBUTES
 #undef __ENUMERATE_ARIA_ATTRIBUTE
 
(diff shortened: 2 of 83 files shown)


tete17:
I wonder if the test brakes around iframes have anything to do here?

Libraries/LibWeb/HTML/HTMLIFrameElement.cpp
```diff
@@ -19,6 +19,8 @@
 #include <LibWeb/HTML/Parser/HTMLParser.h>
 #include <LibWeb/HTML/TraversableNavigable.h>
 #include <LibWeb/Layout/NavigableContainerViewport.h>
+#include <LibWeb/TrustedTypes/TrustedTypePolicy.h>
+#include <LibWeb/TrustedTypes/RequireTrustedTypesForDirective.h>
```
AtkinsSJ: CI is complaining that these are in the wrong order.

Libraries/LibWeb/HTML/HTMLIFrameElement.cpp
```diff
@@ -305,4 +307,31 @@ void HTMLIFrameElement::set_current_navigation_was_lazy_loaded(bool value)
...
+// https://html.spec.whatwg.org/multipage/iframe-embed-object.html#dom-iframe-srcdoc
+WebIDL::ExceptionOr<void> HTMLIFrameElement::set_srcdoc(TrustedTypes::TrustedHTMLOrString const& value)
+{
+    // 1. Let compliantString be the result of invoking the Get Trusted Type compliant string algorithm with
+    //      TrustedHTML, this's relevant global object, the given value, "HTMLIFrameElement srcdoc", and "script".
+    auto const compliant_string = TRY(TrustedTypes::get_trusted_type_compliant_string(
+        TrustedTypes::TrustedTypeName::TrustedHTML,
+        HTML::relevant_global_object(*this),
+        value,
+        TrustedTypes::InjectionSink::HTMLIFrameElementsrcdoc,
```
AtkinsSJ: A tangent, but can we rename these enum values to be clearer? eg, `HTMLIFrameElement_srcdoc`? It's bugging me that there's no break between the two parts. :sweat_smile:

Libraries/LibWeb/SVG/SVGURIReference.h
```diff
@@ -31,7 +32,7 @@ class SVGURIReferenceMixin {
         //      deprecated attribute.
         if (!m_href_animated_string) {
             auto& this_svg_element = as<SVGElement>(*this);
-            m_href_animated_string = SVGAnimatedString::create(this_svg_element.realm(), this_svg_element, AttributeNames::href, supports_xlink_href == SupportsXLinkHref::Yes ? Optional<FlyString> { AttributeNames::xlink_href } : OptionalNone {});
+            m_href_animated_string = SVGAnimatedString::create(this_svg_element.realm(), this_svg_element, DOM::QualifiedName { AttributeNames::href, OptionalNone {}, OptionalNone {} }, supports_xlink_href == SupportsXLinkHref::Yes ? Optional<DOM::QualifiedName> { DOM::QualifiedName { AttributeNames::href, "xlink"_fly_string, Namespace::XLink } } : OptionalNone {});
```
AtkinsSJ: This is a bit much for a single line. :sweat_smile:

PR #6634, opened 2025-10-29
LibWeb: Add stub implementation for Navigator.getBattery()

estefysc:
Adds a stub that returns a cached rejected promise with a not yet implemented error. This converts
battery-status/battery-promise-window.https.html from timeout to pass. Full implementation still needed.

### DIFF
--- a/Libraries/LibWeb/HTML/Navigator.cpp
+++ b/Libraries/LibWeb/HTML/Navigator.cpp
@@ -76,6 +76,7 @@ void Navigator::visit_edges(Cell::Visitor& visitor)
     visitor.visit(m_service_worker_container);
     visitor.visit(m_media_capabilities);
     visitor.visit(m_credentials);
+    visitor.visit(m_battery_promise);
 }
 
 GC::Ref<MimeTypeArray> Navigator::mime_types()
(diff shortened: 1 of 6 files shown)


Libraries/LibWeb/HTML/Navigator.cpp
```diff
@@ -148,4 +149,29 @@ GC::Ref<MediaCapabilitiesAPI::MediaCapabilities> Navigator::media_capabilities()
...
 
+// https://w3c.github.io/battery/
+GC::Ref<WebIDL::Promise> Navigator::get_battery()
+{
+    auto& realm = this->realm();
+
+    // FIXME: This is a stub implementation. getBattery() should:
+    //        1. Query actual system battery status
+    //        2. Resolve with a BatteryManager object
+    //        3. Update when battery status changes
```
gmta: This method is specced here:
https://w3c.github.io/battery/#the-getbattery-method

Could you please use this link instead, and copy over the spec steps? Any spec step you don't intend to implement should be marked as a `// FIXME: ...` comment.

estefysc: Will do

Libraries/LibWeb/HTML/Navigator.cpp
```diff
@@ -148,4 +149,35 @@ GC::Ref<MediaCapabilitiesAPI::MediaCapabilities> Navigator::media_capabilities()
...
+    //    policy-controlled feature, then reject this.[[BatteryPromise]] with a "NotAllowedError" DOMException.
+    // FIXME: Check if Document is allowed to use "battery" policy-controlled feature
+
+    // 3. Otherwise:
+    //    1. If this.[[BatteryManager]] is null, then set it to the result of creating a new BatteryManager
+    //       in this's relevant realm.
+    // FIXME: Create new BatteryManager in this's relevant realm
+
+    //    2. Resolve this.[[BatteryPromise]] with this.[[BatteryManager]].
+    // FIXME
```
gmta: Instead of adding new `FIXME` lines, you can just `FIXME` the spec steps themselves. E.g.:

```cpp
    // FIXME: 1. If this.[[BatteryManager]] is null, then set it to the result of creating a new BatteryManager
    //    in this's relevant realm.
```

gmta:
Thanks @estefysc !

PR #6635, opened 2025-10-29
LibWeb: Bind the default frame/render buffers when resetting context

Zaggy1024:
This fixes the masking and warping effects on pixijs.com.

image

Credit to Luke for debugging this!


Zaggy1024:
Marking this as a draft since in the process of creating a test, it turned out that we differ from other browsers in my test even after the patch.

PR #6641, opened 2025-10-30
LibWeb: Prevent crash in content calculations for ListItemMarkerBox

ayeteadoe:
I was looking into some web platform test results for parsing pseudo elements, and I noticed that Ladybird is currently missing all the tests in `css/css-pseudo/parsing/marker-supported-properties-in-animation.html` due to a crash. See https://wpt.fyi/results/css/css-pseudo/parsing/marker-supported-properties-in-animation.html?label=master&product=chrome%5Bexperimental%5D&product=ladybird&aligned.

The crash is in code introduced in https://github.com/LadybirdBrowser/ladybird/commit/299775345d24633fc4b7bcea27ce6751df5f80ed, as we assume if we can't create a FormattingContext the box is a BlockContainer, but in the test linked above the box is a ListItemMarkerBox.

As I'm not yet familiar with the CSS/Layout infra, this is just a hack for now to get these tests actually running so I can investigate the pseudo element parsing done in this test.


Libraries/LibWeb/Layout/FormattingContext.cpp
```diff
@@ -226,6 +226,20 @@ OwnPtr<FormattingContext> FormattingContext::create_independent_formatting_conte
...
+    auto context = create_independent_formatting_context_if_needed(state, layout_mode, child_box);
+    if (context)
+        return context.release_nonnull();
+
+    auto child_block_container = as_if<BlockContainer>(child_box);
+    if (child_block_container)
+        return make<BlockFormattingContext>(state, layout_mode, *child_block_container, nullptr);
+
+    // HACK: Instead of crashing in scenarios that assume the formatting context can be created, create a dummy formatting context that does nothing.
+    return make<DummyFormattingContext>(state, layout_mode, child_box);
```
AtkinsSJ, line 239: Can we log something here so we know it's happening?

Libraries/LibWeb/Layout/FormattingContext.cpp
```diff
@@ -226,6 +226,20 @@ OwnPtr<FormattingContext> FormattingContext::create_independent_formatting_conte
...
 
+NonnullOwnPtr<FormattingContext> FormattingContext::create_independent_formatting_context(LayoutState& state, LayoutMode layout_mode, Box const& child_box)
+{
+    auto context = create_independent_formatting_context_if_needed(state, layout_mode, child_box);
+    if (context)
+        return context.release_nonnull();
+
+    auto child_block_container = as_if<BlockContainer>(child_box);
+    if (child_block_container)
+        return make<BlockFormattingContext>(state, layout_mode, *child_block_container, nullptr);
```
AtkinsSJ: Nit: We tend to put declarations into the `if` condition where it's readable. eg:

```c++
    if (auto child_block_container = as_if<BlockContainer>(child_box))
        return make<BlockFormattingContext>(state, layout_mode, *child_block_container, nullptr);
```

PR #6642, opened 2025-10-30
LibWeb: A couple of animation fixes/cleanups

Calme1709:

### DIFF
--- a/Libraries/LibWeb/CSS/StyleComputer.cpp
+++ b/Libraries/LibWeb/CSS/StyleComputer.cpp
@@ -2568,17 +2568,6 @@ GC::Ref<ComputedProperties> StyleComputer::compute_properties(DOM::AbstractEleme
     // 5. Add or modify CSS-defined animations
     process_animation_definitions(computed_style, abstract_element);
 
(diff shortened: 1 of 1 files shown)


Libraries/LibWeb/Animations/Animation.h
```diff
@@ -49,6 +49,7 @@ class Animation : public DOM::EventTarget {
     WebIDL::ExceptionOr<void> set_playback_rate(double value);
 
     Bindings::AnimationPlayState play_state() const;
+    Bindings::AnimationPlayState play_state_impl() const;
```
AtkinsSJ: With both of these being public, it's not clear when one or the other should be used, which feels like a footgun. Maybe the second could have a clearer name to say it doesn't update style first?

Calme1709: I have renamed the original function back to `play_state` and renamed the new one to `play_state_for_bindings` to make it clear it shouldn't be used in most cases.

AtkinsSJ (approved):
Looks good! Just has the CI wasm issue we need to wait for a fix for.

Calme1709:
Rebased to latest master

PR #6646, opened 2025-10-30
LibWeb/CSS: Stop stripping out whitespace tokens before parsing property values

AtkinsSJ:
Removing whitespace before parsing seemed like a good idea to me multiple years ago, but caused issues: For one, we didn't remove it from nested functions and blocks, so lots of parsing code still needed to care about whitespace. Another issue is that we then had to re-insert the whitespace when serializing, which produced incorrect results. Further, it meant we always copied the input `ComponentValues`, which we can now avoid. :^)

To avoid breaking a large number of tests at once and then having a nightmarish debugging session, this is structured like so:
1. Introduce a `strip-whitespace` field to every property in `Properties.json`, which preserves the old behaviour.
2. Gradually remove that from groups of properties, and fix the issues that arise.
3. Remove the `strip-whitespace` field again.
4. Clean up some parts of the code that previously had to deal with missing whitespace.

It's resulted in a lot of commits, but a lot of them are just a small JSON diff.

It's possible that this could cause some regressions in places we don't have good test coverage, so maybe wait until after the newsletter and monthly update video before merging. 😉

### DIFF
--- a/Libraries/LibWeb/CSS/Serialize.cpp
+++ b/Libraries/LibWeb/CSS/Serialize.cpp
@@ -199,11 +199,9 @@ String serialize_a_css_declaration(StringView property, StringView value, Import
 }
 
 // https://drafts.csswg.org/css-syntax/#serialization
-String serialize_a_series_of_component_values(ReadonlySpan<Parser::ComponentValue> component_values, InsertWhitespace insert_whitespace)
+String serialize_a_series_of_component_values(ReadonlySpan<Parser::ComponentValue> component_values)
 {
     // FIXME: There are special rules here where we should insert a comment between certain tokens. Do that!
-    if (insert_whitespace == InsertWhitespace::Yes)
-        return MUST(Infra::strip_and_collapse_whitespace(MUST(String::join(' ', component_values))));
     return MUST(String::join(""sv, component_values));
 }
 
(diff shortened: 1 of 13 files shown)


Libraries/LibWeb/CSS/Properties.json
```diff
@@ -3445,22 +3445,19 @@
...
   },
   "scrollbar-color": {
-    "strip-whitespace": true,
     "affects-layout": false,
     "animation-type": "by-computed-value",
     "inherited": "yes",
     "initial": "auto"
   },
   "scrollbar-gutter": {
-    "strip-whitespace": true,
```
tcl3: I ran all the WPT CSS parsing  tests and I'm seeing a failure in http://wpt.live/css/css-overflow/parsing/scrollbar-gutter-valid.html. It looks like `Parser::parse_scrollbar_gutter_value()` needs to explicitly `discard_whitespace()` in a few places.

AtkinsSJ:
Changes: Fixed the `scrollbar-gutter` parsing regression and imported the test for it. Thanks Tim!

AtkinsSJ:
Changes: Just a rebase, to fix the wasm issue.

AtkinsSJ:
Looks like this did cause a regression here: https://wpt.fyi/results/css/css-syntax/serialize-consecutive-tokens.html?diff&filter=ADC&run_id=5176746556260352&run_id=5090655627116544

I'm confused why we were doing it right before. I'll take a look tomorrow.

PR #6649, opened 2025-10-30
Integrate `eval` with TrustedTypes API 

tete17:
This was the missing puzzle of the spec that doesn't actually live in the spec.

We add the ability for CSP to block execution of `eval` if the polices require TrustedTypes and allow `eval` to execute TrustedScripts.

We get around 100 new WPT's with this change :)

### DIFF
--- a/Tests/LibWeb/Text/expected/wpt-import/trusted-types/Window-block-eval-function-constructor.txt
+++ b/Tests/LibWeb/Text/expected/wpt-import/trusted-types/Window-block-eval-function-constructor.txt
@@ -1,0 +1,11 @@
+Harness status: OK
+
+Found 6 tests
+
+6 Pass
+Pass	Blocked eval in Window.
+Pass	Blocked indirect eval in Window.
+Pass	Blocked Function constructor in Window.
+Pass	Blocked AsyncFunction constructor in Window.
+Pass	Blocked GeneratorFunction constructor in Window.
+Pass	Blocked AsyncGeneratorFunction constructor in Window.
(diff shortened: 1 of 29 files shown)


Libraries/LibWeb/Bindings/MainThreadVM.cpp
```diff
@@ -125,8 +125,10 @@ void initialize_main_thread_vm(AgentType type)
     };
 
     // 8.1.6.3 HostGetCodeForEval(argument), https://html.spec.whatwg.org/multipage/webappapis.html#hostgetcodeforeval(argument)
-    s_main_thread_vm->host_get_code_for_eval = [](JS::Object const&) -> GC::Ptr<JS::PrimitiveString> {
-        // FIXME: 1. If argument is a TrustedScript object, then return argument's data.
+    s_main_thread_vm->host_get_code_for_eval = [](JS::Object const& argument) -> GC::Ptr<JS::PrimitiveString> {
+        // 1. If argument is a TrustedScript object, then return argument's data.
+        if (auto const* trusted_script = as_if<TrustedTypes::TrustedScript>(argument); trusted_script)
+            return JS::PrimitiveString::create(argument.vm(), trusted_script->to_json());
```
Lubrsi: suggestion: Even though `to_json` is the same as `to_string` for Trusted Types, it is strange seeing it used and might confuse someone unfamiliar with Trusted Types.
```suggestion
            return JS::PrimitiveString::create(argument.vm(), trusted_script->to_string());
```

tete17: Sure thing :)

Issue #6650, opened 2025-10-31
Crash with "content-security-policy: require-trusted-types-for 'script'" and a `<select>` element


JeremiasRy:
### Summary

Ladybird crashes when trying to render `<select>` tag with response header containing `content-security-policy: require-trusted-types-for 'script'`.

### Operating system

Linux

### Steps to reproduce

1. Visit [consent.google](https://consent.google.com/ml?continue=https://www.google.com/&gl=FI&m=0&pc=shp&uxe=none&cm=2&hl=fi&src=1)
2. Ladybird flew off-course!

### Expected behavior

Successfully load the web page.

### Actual behavior

Ladybird crashes.

### URL for a reduced test case

N/A

### HTML/SVG/etc. source for a reduced test case

```html
// HTML itself doesn't produce the error, we need a response header in conjunction with the content to produce the crash

func main() {
	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
		html, _ := os.ReadFile("index.html")
		w.Header().Set("content-security-policy", "require-trusted-types-for 'script'")
		fmt.Fprint(w, string(html))
	})

	log.Println("Listening...")
	log.Fatal(http.ListenAndServe(":8080", nil))
}


<!DOCTYPE html>
<html lang="en">

<head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>Document</title>
</head>

<body>
    <select>
        <option>1</option>
        <option>2</option>
        <option>3</option>
    </select>
</body>

</html>
```

### Log output and (if possible) backtrace

```shell
se it does not appear in the 'require-trusted-types-for' directive.
UNEXPECTED ERROR: Sink Element innerHTML of type 'script' requires a TrustedType to be used at /home/rytta/repos/ladybird/Libraries/LibWeb/HTML/HTMLSelectElement.cpp:628
Stack trace (most recent call first):
#0  (inlined)          in Web::DOM::Node::insert_before(GC::Ref<Web::DOM::Node>, GC::Ptr<Web::DOM::Node>, bool)::$_0::operator()(Web::DOM::Node&) const at /home/rytta/repos/ladybird/Libraries/LibWeb/DOM/Node.cpp:734:34
[...]

PR #6651, opened 2025-10-31
LibWeb/HTML: Use DOM::create_element to create chevron SVG for select element

JeremiasRy:
Fixes [6650](https://github.com/LadybirdBrowser/ladybird/issues/6650),

The crash outlined in above issue happens when we reach [get_trusted_type_compliant_string()](https://github.com/LadybirdBrowser/ladybird/blob/2f7797f8543cd938ef836d85baaab15aa65f586e/Libraries/LibWeb/TrustedTypes/TrustedTypePolicy.cpp#L261) with a raw string (that is our chevron icon). The function attempts to parse it as `TrustedTypeName::TrustedHTML`. But it's not expecting a string, but a `TrustedHTML` and thus fails to get the string value.

Simpler fix for this would be to just then return the string value inside the function, but this didn't feel right.

For attributes `fill` and `d`, I didn't find `AttributeNames::` definitions, should these attributes be added to [SVG/AttributeNames.h](https://github.com/LadybirdBrowser/ladybird/blob/2f7797f8543cd938ef836d85baaab15aa65f586e/Libraries/LibWeb/SVG/AttributeNames.h)?

I'm not sure if this is a _better way to add a chevron icon_ if there is an alternative way, I would appreciate any help in guiding me to that direction.

Before patch:

image

After:

image

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLSelectElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLSelectElement.cpp
@@ -603,8 +603,14 @@ void HTMLSelectElement::create_shadow_tree_if_needed()
     )~~~"_string));
     MUST(border->append_child(*m_inner_text_element));
 
-    // FIXME: Find better way to add chevron icon
-    static constexpr auto chevron_svg = "<svg xmlns=\"http://www.w3.org/2000/svg\" viewBox=\"0 0 24 24\"><path fill=\"currentcolor\" d=\"M7.41,8.58L12,13.17L16.59,8.58L18,10L12,16L6,10L7.41,8.58Z\"/></svg>"sv;
+    auto chevron_svg = DOM::create_element(document(), HTML::TagNames::svg, Namespace::HTML).release_value_but_fixme_should_propagate_errors();
+    MUST(chevron_svg->set_attribute(SVG::AttributeNames::viewBox, "0 0 24 24"_string));
+    MUST(chevron_svg->set_attribute("xmlns"_fly_string, "http://www.w3.org/2000/svg"_string));
+
+    auto chevron_path = DOM::create_element(document(), SVG::TagNames::path, Namespace::SVG).release_value_but_fixme_should_propagate_errors();
+    MUST(chevron_path->set_attribute("fill"_fly_string, "currentcolor"_string));
+    MUST(chevron_path->set_attribute("d"_fly_string, "M7.41,8.58L12,13.17L16.59,8.58L18,10L12,16L6,10L7.41,8.58Z"_string));
+    MUST(chevron_svg->append_child(*chevron_path));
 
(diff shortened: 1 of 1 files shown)


Libraries/LibWeb/HTML/HTMLSelectElement.cpp
```diff
@@ -603,16 +603,23 @@ void HTMLSelectElement::create_shadow_tree_if_needed()
...
 
-    // FIXME: Find better way to add chevron icon
-    static constexpr auto chevron_svg = "<svg xmlns=\"http://www.w3.org/2000/svg\" viewBox=\"0 0 24 24\"><path fill=\"currentcolor\" d=\"M7.41,8.58L12,13.17L16.59,8.58L18,10L12,16L6,10L7.41,8.58Z\"/></svg>"sv;
+    auto chevron_svg = DOM::create_element(document(), HTML::TagNames::svg, Namespace::HTML).release_value_but_fixme_should_propagate_errors();
+    MUST(chevron_svg->set_attribute(SVG::AttributeNames::viewBox, "0 0 24 24"_string));
+    MUST(chevron_svg->set_attribute("xmlns"_fly_string, "http://www.w3.org/2000/svg"_string));
+
+    auto chevron_path = DOM::create_element(document(), SVG::TagNames::path, Namespace::SVG).release_value_but_fixme_should_propagate_errors();
+    MUST(chevron_path->set_attribute("fill"_fly_string, "currentcolor"_string));
+    MUST(chevron_path->set_attribute("d"_fly_string, "M7.41,8.58L12,13.17L16.59,8.58L18,10L12,16L6,10L7.41,8.58Z"_string));
```
AtkinsSJ, line 612: > For attributes fill and d, I didn't find AttributeNames:: definitions, should these attributes be added to [SVG/AttributeNames.h](https://github.com/LadybirdBrowser/ladybird/blob/2f7797f8543cd938ef836d85baaab15aa65f586e/Libraries/LibWeb/SVG/AttributeNames.h)?

Yes please!

JeremiasRy:
Thanks for the welcome @AtkinsSJ !

I just noticed that while this fixes the crash, the select element now misses the chevron. I'll have to dig a little bit to figure out why. (I guess there was a reason for the svg being created with the raw string?)

tete17:
It feels like this changes ara a subste of https://github.com/LadybirdBrowser/ladybird/pull/6632/files#diff-165889c8934d23700a269b376bb7f2f6c025aed087641bb46879d12596877863

Cc @Lubrsi

JeremiasRy:
> It feels like this changes ara a subste of https://github.com/LadybirdBrowser/ladybird/pull/6632/files#diff-165889c8934d23700a269b376bb7f2f6c025aed087641bb46879d12596877863

It definitely is! I tried to do some scanning if this was being handled already, but missed your draft. 

Thanks everyone for participating, I'll close this.

Issue #6647, opened 2025-10-30
LibRegex: Crash in the optimizer and parser


aplefull:
### Summary

Found a candidate for a monthly regex bug :)   

These two crash optimizer (logs below):
`"".match(/([[[]]])*0/v);`
`"".match(/(([[[]]]{2,})\s)*/gv);`

Also found a crash in the parser while trying to create these reductions 😅
`"".match(/[[a][]]/gv);`

### Operating system

Linux

### Steps to reproduce

N/A

### Expected behavior

N/A

### Actual behavior

N/A

### URL for a reduced test case

data:text/html;base64,PHNjcmlwdD4NCiAgICAiIi5tYXRjaCgvW1thXVtdXS9ndik7DQogICAgIiIubWF0Y2goLyhbW1tdXV0pKjAvdik7DQogICAgIiIubWF0Y2goLygoW1tbXV1dezIsfSlccykqL2d2KTsNCjwvc2NyaXB0Pg==

### HTML/SVG/etc. source for a reduced test case

```html
<script>
    "".match(/[[a][]]/gv);
    "".match(/([[[]]])*0/v);
    "".match(/(([[[]]]{2,})\s)*/gv);
</script>
```

### Log output and (if possible) backtrace

```shell
VERIFICATION FAILED: in_or at /home/aplefull/Repos/ladybird/Libraries/LibRegex/RegexOptimizer.cpp:594
Stack trace (most recent call first):
#0  (inlined)          in has_overlap at /home/aplefull/Repos/ladybird/Libraries/LibRegex/RegexOptimizer.cpp:594:13
#1  (inlined)          in operator()<AK::Vector<regex::CompareTypeAndValuePair, 0UL, (AK::FastLastAccess)0> > at /home/aplefull/Repos/ladybird/Libraries/LibRegex/RegexOptimizer.cpp:764:76
[...]

PR #6660, opened 2025-11-01
LibRegex: Track local compares in nested classes

aplefull:
This prevents verification failures when parsing nested empty character classes. We now only count compares added by the current parse_nested_class invocation.

Fixes one crash in #6647,  but I don’t dare touch the optimizer, so hopefully @alimpfard can take a look at the second crash there. :sweat_smile:

### DIFF
--- a/Tests/LibRegex/TestRegex.cpp
+++ b/Tests/LibRegex/TestRegex.cpp
@@ -891,6 +891,7 @@ TEST_CASE(ECMA262_unicode_sets_match)
         { "[\\w&&\\d]"sv, "a"sv, false },
         { "[\\w&&\\d]"sv, "4"sv, true },
         { "([^\\:]+?)"sv, "a"sv, true },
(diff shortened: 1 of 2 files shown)


Tests/LibRegex/TestRegex.cpp
```diff
@@ -891,6 +891,7 @@ TEST_CASE(ECMA262_unicode_sets_match)
         { "[\\w&&\\d]"sv, "a"sv, false },
         { "[\\w&&\\d]"sv, "4"sv, true },
         { "([^\\:]+?)"sv, "a"sv, true },
+        { "[[a][]]"sv, "a"sv, true },
```
alimpfard: Can you add a ref like `// ladybird#6647` to this?

aplefull: Sure, done!

PR #3549, opened 2025-02-12
LibCore: Do not delete ThreadEventQueue

stasoid:
Partially reverts 920f4707358.

On Linux the deleted code doesn't work anyway -- ThreadEventQueue destructor is never called. To test this, add dbgln to ~ThreadEventQueue and run either headless-browser or ladybird.

On Windows it works, but results in use-after-free: ~ThreadEventQueue is called after WebView::WebContentClient::s_clients destructor, but ~ThreadEventQueue indirectly calls
~WebContentClient, which calls `s_clients.remove(this);`

I tried to use OwnPtr\<ThreadEventQueue> (see #3240), but LibWeb tests started failing.

Deleting ThreadEventQueue is pointless. Raymond Chen, in [this article](https://devblogs.microsoft.com/oldnewthing/20120105-00/?p=8683):
> The process is exiting. All that memory will be freed when the address space is destroyed. *Stop wasting time and just exit already.*

PR #6661, opened 2025-11-01
LibCore+Tests/LibWeb: Implement HttpEchoServerFixture on Windows

ayeteadoe:
The first task towards the goal of getting full `test-web` runs working on Windows so it can be enabled in CI.

### Demo

We can ran run and pass a single test that utilizes the python-based HTTP test server via:

`python Meta\ladybird.py run --preset Windows_CI test-web -f Text\\input\\test-http-test-server.html`

https://github.com/user-attachments/assets/9b070845-8535-4452-8f6d-7055389e4592

### DIFF
--- a/Tests/LibWeb/test-web/Application.cpp
+++ b/Tests/LibWeb/test-web/Application.cpp
@@ -16,7 +16,11 @@ namespace TestWeb {
 Application::Application(Optional<ByteString> ladybird_binary_path)
     : WebView::Application(move(ladybird_binary_path))
     , test_concurrency(Core::System::hardware_concurrency())
+#if defined(AK_OS_WINDOWS)
+    , python_executable_path("python")
+#else
     , python_executable_path("python3")
+#endif
 {
     if (auto ladybird_source_dir = Core::Environment::get("LADYBIRD_SOURCE_DIR"sv); ladybird_source_dir.has_value())
         test_root_path = LexicalPath::join(*ladybird_source_dir, "Tests"sv, "LibWeb"sv).string();
--- a/Tests/LibWeb/test-web/Fixture.cpp
+++ b/Tests/LibWeb/test-web/Fixture.cpp
@@ -13,6 +13,10 @@
 #include <LibCore/StandardPaths.h>
 #include <LibCore/System.h>
 
+#if defined(AK_OS_WINDOWS)
+#    include <AK/Windows.h>
+#endif
+
 namespace TestWeb {
 
 static ByteString s_fixtures_path;
@@ -45,21 +49,10 @@ public:
 private:
     ByteString m_script_path { "http-test-server.py" };
     Optional<Core::Process> m_process;
-};
-
 #if defined(AK_OS_WINDOWS)
-
-ErrorOr<void> HttpEchoServerFixture::setup(WebView::WebContentOptions&)
-{
-    VERIFY(0 && "HttpEchoServerFixture::setup is not implemented");
-}
-
-void HttpEchoServerFixture::teardown_impl()
-{
-    VERIFY(0 && "HttpEchoServerFixture::teardown_impl is not implemented");
-}
-
-#else
+    HANDLE m_stderr_log_handle { nullptr };
+#endif
+};
 
 ErrorOr<void> HttpEchoServerFixture::setup(WebView::WebContentOptions& web_content_options)
 {
@@ -68,23 +61,54 @@ ErrorOr<void> HttpEchoServerFixture::setup(WebView::WebContentOptions& web_conte
 
     // FIXME: Pick a more reasonable log path that is more observable
     auto const log_path = LexicalPath::join(Core::StandardPaths::tempfile_directory(), "http-test-server.log"sv).string();
+    auto const stderr_log_path = ByteString::formatted("{}.stderr", log_path);
 
-    auto stdout_fds = TRY(Core::System::pipe2(0));
-
-    auto const process_options = Core::ProcessSpawnOptions {
+    auto process_options = Core::ProcessSpawnOptions {
         .executable = Application::the().python_executable_path,
         .search_for_executable_in_path = true,
         .arguments = arguments,
-        .file_actions = {
-            Core::FileAction::OpenFile { ByteString::formatted("{}.stderr", log_path), Core::File::OpenMode::Write, STDERR_FILENO },
-            Core::FileAction::DupFd { stdout_fds[1], STDOUT_FILENO } }
     };
+
+    int stdout_read_fd = -1;
+
+#if defined(AK_OS_WINDOWS)
+    HANDLE stdout_read_handle { nullptr };
+    HANDLE stdout_write_handle { nullptr };
+    if (!CreatePipe(&stdout_read_handle, &stdout_write_handle, nullptr, 0))
+        return Error::from_windows_error();
+    if (!SetHandleInformation(stdout_write_handle, HANDLE_FLAG_INHERIT, HANDLE_FLAG_INHERIT))
+        return Error::from_windows_error();
+
+    m_stderr_log_handle = CreateFile(stderr_log_path.characters(), GENERIC_WRITE, FILE_SHARE_WRITE, nullptr, CREATE_ALWAYS, FILE_ATTRIBUTE_NORMAL, nullptr);
+    if (m_stderr_log_handle == INVALID_HANDLE_VALUE)
+        return Error::from_windows_error();
+    if (!SetHandleInformation(m_stderr_log_handle, HANDLE_FLAG_INHERIT, HANDLE_FLAG_INHERIT))
+        return Error::from_windows_error();
+
+    process_options.use_std_handles_startup_info = Core::StartupInfo::UseStdHandles { .stderr_handle = m_stderr_log_handle, .stdout_handle = stdout_write_handle, .stdin_handle = nullptr };
+    process_options.create_new_process_group = true;
+
+    stdout_read_fd = to_fd(stdout_read_handle);
+#else
+    auto stdout_fds = TRY(Core::System::pipe2(0));
+
+    process_options.file_actions = {
+        Core::FileAction::OpenFile { stderr_log_path, Core::File::OpenMode::Write, STDERR_FILENO },
+        Core::FileAction::DupFd { stdout_fds[1], STDOUT_FILENO }
+    };
+
+    stdout_read_fd = stdout_fds[0];
+#endif
 
     m_process = TRY(Core::Process::spawn(process_options));
 
+#if defined(AK_OS_WINDOWS)
+    CloseHandle(stdout_write_handle);
+#else
     TRY(Core::System::close(stdout_fds[1]));
+#endif
 
-    auto const stdout_file = MUST(Core::File::adopt_fd(stdout_fds[0], Core::File::OpenMode::Read));
+    auto const stdout_file = MUST(Core::File::adopt_fd(stdout_read_fd, Core::File::OpenMode::Read));
 
     auto buffer = MUST(ByteBuffer::create_uninitialized(5));
     TRY(stdout_file->read_some(buffer));
@@ -96,6 +120,13 @@ ErrorOr<void> HttpEchoServerFixture::setup(WebView::WebContentOptions& web_conte
     else
         warnln("Failed to read echo server port from buffer: '{}'", raw_output);
 
+#if defined(AK_OS_WINDOWS)
+    // Currently our File/IPC/Event Loop infrastructure on Windows assumes we don't use pipes, only regular file handles
+    // and WinSock2-based socket fds. So currently if we let Core::File try to close the read FD we hit fail an assertion as
+    // System::close() thinks we're a socket. We will just manually close the read FD here as the rest of the codebase is not
+    // planning on using Windows pipes
+    CloseHandle(to_handle(stdout_file->leak_fd()));
+#endif
     return {};
 }
 
@@ -105,6 +136,9 @@ void HttpEchoServerFixture::teardown_impl()
 
     auto script_path = LexicalPath::join(s_fixtures_path, m_script_path);
 
+#if defined(AK_OS_WINDOWS)
+    CloseHandle(m_stderr_log_handle);
+#endif
     if (auto kill_or_error = Core::System::kill(m_process->pid(), SIGINT); kill_or_error.is_error()) {
         if (kill_or_error.error().code() != ESRCH) {
             warnln("Failed to kill HTTP echo server, error: {}", kill_or_error.error());
@@ -115,8 +149,6 @@ void HttpEchoServerFixture::teardown_impl()
 
     m_process = {};
 }
-
-#endif
 
 void Fixture::initialize_fixtures()
 {


Tests/LibWeb/test-web/Fixture.cpp
```diff
@@ -67,23 +62,67 @@ ErrorOr<void> HttpEchoServerFixture::setup(WebView::WebContentOptions& web_conte
...
+
+    SECURITY_ATTRIBUTES security_attrs = {};
+    security_attrs.nLength = sizeof(SECURITY_ATTRIBUTES);
+    security_attrs.bInheritHandle = TRUE;
+    security_attrs.lpSecurityDescriptor = nullptr;
+
+    if (!CreatePipe(&stdout_read_handle, &stdout_write_handle, &security_attrs, 0))
+        return Error::from_windows_error();
+
+    if (!SetHandleInformation(stdout_read_handle, HANDLE_FLAG_INHERIT, 0))
```
R-Goc: You already set bInheritHandle to TRUE on the security attributes so the calls to SetHandleInformation are unnecesary.

Tests/LibWeb/test-web/Fixture.cpp
```diff
@@ -67,23 +62,67 @@ ErrorOr<void> HttpEchoServerFixture::setup(WebView::WebContentOptions& web_conte
...
+    if (!CreatePipe(&stderr_read_handle, &stderr_write_handle, &security_attrs, 0))
+        return Error::from_windows_error();
+
+    if (!SetHandleInformation(stderr_read_handle, HANDLE_FLAG_INHERIT, 0))
+        return Error::from_windows_error();
+
+    process_options.startup_info = Core::StartupInfo::UseStdHandles { .stderr_handle = stderr_write_handle, .stdout_handle = stdout_write_handle, .stdin_handle = nullptr };
+
+    stdout_read_fd = to_fd(stdout_read_handle);
+    m_stderr_read_handle = stderr_read_handle;
```
R-Goc: Any reasons not to save directly to the member variables?

Tests/LibWeb/test-web/Fixture.cpp
ayeteadoe: This is disabling handle inheritance for the read handles on the pipes. The child process only needs to inherit the write side. I could also not set inherit in sec attrs and use these 2 calls to just enable handle inheritance for the pipes write handles. I'll go that route to make things less confusing

Tests/LibWeb/test-web/Fixture.cpp
ayeteadoe: Update: After a refactor that simplified things, this comment has implicitly been addressed

No functional reason, just wanted to be more explicitly for readability. We're declaring several handles but only a subset need to to be used during teardown, so this way keeps that setup code consistent instead of a mix of local/member variables. It's just a style thing though, so I'll leave as-is for now to see if a maintainer also prefers the more direct approach.

Tests/LibWeb/test-web/Application.cpp
```diff
@@ -16,7 +16,11 @@ namespace TestWeb {
 Application::Application(Optional<ByteString> ladybird_binary_path)
     : WebView::Application(move(ladybird_binary_path))
     , test_concurrency(Core::System::hardware_concurrency())
+#if defined(AK_OS_WINDOWS)
+    , python_executable_path("python")
+#else
     , python_executable_path("python3")
+#endif
```
ADKaster, line 23: is `python3` not a thing on your system?

```powershell
PS C:\Users\Andrew\source\repos> (Get-Command python3).Path
C:\Users\Andrew\AppData\Local\Microsoft\WindowsApps\python3.exe
PS C:\Users\Andrew\source\repos> (Get-Command python).Path
C:\Users\Andrew\AppData\Local\Microsoft\WindowsApps\python.exe
```

Libraries/LibCore/Process.h
```diff
@@ -44,6 +58,11 @@ struct ProcessSpawnOptions {
 
     using FileActionType = Variant<FileAction::OpenFile, FileAction::CloseFile, FileAction::DupFd>;
     Vector<FileActionType> file_actions {};
+
+#if defined(AK_OS_WINDOWS)
+    using StartupInfoType = Variant<StartupInfo::UseStdHandles>;
```
ADKaster: variant with one type? do you have WIP work to add extra startup info types? It looks like you always name `UseStdHandles` in this PR instead of using StartupInfoType

Tests/LibWeb/test-web/Application.cpp
nico: I think if you get python from python.org, it gives you python.exe on win

ayeteadoe: On Windows `py` or `python` is the typical approach. I don't think I've ever had `python3` on any Windows system exist/work (as nico states, due to python.org-based installs).

And from https://docs.python.org/3/using/windows.html#basic-uses it seems that it is not standard/expected:

> A python3 command is also included that mimics the python command. It is intended to catch accidental uses of the typical POSIX command on Windows, but is not meant to be widely used or recommended.

You would do `py -<version>` to distinguish between multiple python installs

Libraries/LibCore/Process.h
ayeteadoe: There are other startup flags, where some have mutually exclusive or require certain struct fields set, so this was just modelling how `STARTUPINFO` can be configured. That being said, looking at the other options I'd be surprised if we needed them, so I can simplify to an `Optional` struct if preferred.

ayeteadoe:
**Update:** My investigation was already discovered in https://github.com/LadybirdBrowser/ladybird/pull/3549 and discussed in discord. So the following is nothing new, but just leaving it here for future reference. 

<details> 

<summary>UAF issue when `test-web` shuts down on Windows</summary>

---

ASAN found a Windows-only use after free when running the demo in the first comment. It seems to be in `test-web` (i.e. the headless implementation) only, as I have not seen this specific error in running `Ladybird` with ASAN.

To repro:
1. Checkout this branch
2. Run `python Meta\ladybird.py run --preset Windows_Sanitizer_CI test-web -f Text\\input\\test-http-test-server.html`
3. ASAN reports the following:
```
SUMMARY: AddressSanitizer: heap-use-after-free C:\Users\ayeteadoe\Development\ladybird\AK\HashTable.h:639 in AK::HashTable<class WebView::WebContentClient *, struct AK::Traits<class WebView::WebContentClient *>, 0>::lookup_with_hash<class `public: class AK::HashTableIterator<class AK::HashTable<class WebView::WebContentClient *, struct AK::Traits<class WebView::WebContentClient *>, 0>, class WebView::WebContentClient *, struct AK::HashTable<class WebView::WebContentClient *, struct AK::Traits<class WebView::WebContentClient *>, 0>::Bucket> __cdecl AK::HashTable<class WebView::WebContentClient *, struct AK::Traits<class WebView::WebContentClient *>, 0>::find(class WebView::WebContentClient *const &)'::`1'::<lambda_1>>(unsigned int, class `public: class AK::HashTableIterator<class AK::HashTable<class WebView::WebContentClient *, struct AK::Traits<class WebView::WebContentClient *>, 0>, class WebView::WebContentClient *, struct AK::HashTable<class WebView::WebContentClient *, struct AK::Traits<class WebView::WebContentClient *>, 0>::Bucket> __cdecl AK::HashTable<class WebView::WebContentClient *, struct AK::Traits<class WebView::WebContentClient *>, 0>::find(class WebView::WebContentClient *const &)'::`1'::<lambda_1>) const
```

In short, https://github.com/LadybirdBrowser/ladybird/blob/master/AK/HashTable.h#L639 is being invoked on a `m_buckets` that has already been deleted.

Expand below for the full details/context, but to summarize the root cause, when the main thread for `test-web` is shutting down:
- On all platforms, the main thread `ThreadEventQueue` has a single pending deferred invoke event, where the receiver is a `WebContentClient`
- For POSIX pthread, the destructor we setup in `ThreadEventQueue::current()` does not get invoked implicitly, only when `pthread_exit()` is called, and since we do not call that for the main thread, the destructor does not get invoked on macOS/Linux
- For Windows pthreads4w implementation, we **do** invoke the destructor when the main thread is exiting. This means that we end up executing that deferred invoke event when the main thread is being destroyed, which after it completes the `WebContentClient` has its reference count drop to 0 and it is destroyed. But the static `HashTable` that was keeping track of active clients was destroyed already (I'm assuming all static storage cleanup has already occurred)


<details>

<summary>Full Investigation and Root Cause Identification</summary>

### The Problem

This issue seems to stem from a destruction ordering issue. First, apply the following patch to this branch,

```patch
diff --git a/AK/HashTable.h b/AK/HashTable.h
index 44aa24f756..4b8d14da6d 100644
--- a/AK/HashTable.h
+++ b/AK/HashTable.h
@@ -190,6 +190,12 @@ public:
         }
 
         kfree_sized(m_buckets, size_in_bytes(m_capacity));
+        m_buckets = reinterpret_cast<BucketType*>(0xDEADBEEF);
+    }
+
+    void ensure_valid_bucket() const
+    {
+        VERIFY(m_buckets != reinterpret_cast<BucketType*>(0xDEADBEEF));
     }
 
     HashTable(HashTable const& other)
diff --git a/Libraries/LibWebView/WebContentClient.cpp b/Libraries/LibWebView/WebContentClient.cpp
index 4178e43353..de98736653 100644
--- a/Libraries/LibWebView/WebContentClient.cpp
+++ b/Libraries/LibWebView/WebContentClient.cpp
@@ -41,6 +41,7 @@ WebContentClient::WebContentClient(NonnullOwnPtr<IPC::Transport> transport)
 
 WebContentClient::~WebContentClient()
 {
+    s_clients.ensure_valid_bucket();
     s_clients.remove(this);
 }
```

Re-running the test command from the repro shows that we are calling a `WebContentClient` destructor after the `s_clients` hash table has been deleted. Here is what we get on Windows
```
Running 1 tests...
Done!
Pass: 1, Fail: 0, Skipped: 0, Timeout: 0, Crashed: 0
VERIFICATION FAILED: m_buckets != reinterpret_cast<BucketType*>(0xDEADBEEF) at C:\Users\ayeteadoe\Development\ladybird\AK/HashTable.h:198
#0  0x00007ffae9042f86 in WebView::WebContentClient::~WebContentClient() at C:\Users\ayeteadoe\Development\ladybird\Libraries\LibWebView\WebContentClient.cpp:44
#1  0x00007ffae90610d5 in WebView::WebContentClient::~WebContentClient() at C:\Users\ayeteadoe\Development\ladybird\Libraries\LibWebView\WebContentClient.cpp:43
#2  0x00007ffaf18c2fd8 in Core::EventReceiver::deferred_invoke::<lambda_0>::~`lambda at C:\Users\ayeteadoe\Development\ladybird\Libraries\LibCore\EventReceiver.cpp:63:27'() at C:\Users\ayeteadoe\Development\ladybird\Libraries\LibCore\EventReceiver.cpp:63
#3  0x00007ffaf18c3589 in AK::Function<void ()>::CallableWrapper<`lambda at C:\Users\ayeteadoe\Development\ladybird\Libraries\LibCore\EventReceiver.cpp:63:27'>::~CallableWrapper() at C:\Users\ayeteadoe\Development\ladybird\AK\Function.h:229
#4  0x00007ffaf18c00b8 in AK::Function<void ()>::clear(bool) at C:\Users\ayeteadoe\Development\ladybird\AK\Function.h:301
#5  0x00007ffaf18c1cdc in Core::DeferredInvocationEvent::~DeferredInvocationEvent() at C:\Users\ayeteadoe\Development\ladybird\Libraries\LibCore\Event.h:47
#6  0x00007ffaf18c1c45 in Core::DeferredInvocationEvent::~DeferredInvocationEvent() at C:\Users\ayeteadoe\Development\ladybird\Libraries\LibCore\Event.h:47
#7  0x00007ffaf18f82fa in AK::Vector<Core::ThreadEventQueue::Private::QueuedEvent, 0, 0>::clear_with_capacity() at C:\Users\ayeteadoe\Development\ladybird\AK\Vector.h:464
#8  0x00007ffaf18f4ddc in AK::DefaultDelete<Core::ThreadEventQueue::Private>::operator()(Core::ThreadEventQueue::Private*) at C:\Users\ayeteadoe\Development\ladybird\AK\DefaultDelete.h:17
#9  0x00007ffaf18f18ae in Core::ThreadEventQueue::~ThreadEventQueue() at C:\Users\ayeteadoe\Development\ladybird\Libraries\LibCore\ThreadEventQueue.cpp:68
#10 0x00007ffaf18f4b24 in Core::ThreadEventQueue::current::<lambda_0>::operator()::<lambda_1>::__invoke() at C:\Users\ayeteadoe\Development\ladybird\Libraries\LibCore\ThreadEventQueue.cpp:49
#11 0x00007ffaf8a813e9 in  ??
#12 0x00007ffaf8a88a8d in pthread_win32_thread_detach_np
#13 0x00007ffaf8a81037 in  ??
#14 0x00007ffaf8a89c65 in sem_wait
#15 0x00007ffb13eff96d in RtlEncodeRemotePointer
#16 0x00007ffb13dabcad in RtlRaiseException
#17 0x00007ffb13e2d4ae in LdrShutdownProcess
#18 0x00007ffb13e2c67d in RtlExitUserProcess
#19 0x00007ffb127718aa in FatalExit
#20 0x00007ffb11a90092 in logbf
#21 0x00007ff63cafa00a in __scrt_common_main_seh() at D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl:295
#22 0x00007ffb1275e8d6 in BaseThreadInitThunk
#23 0x00007ffb13e2c53b in RtlUserThreadStart
```

While on both macOS/Linux we are fine:
```
Running 1 tests...
Done!
Pass: 1, Fail: 0, Skipped: 0, Timeout: 0, Crashed: 0
```

When debugging the `test-web` process only on macOS/Linux, if I put a breakpoint in the `WebContentClient` destructor, it is only hit once, with the following call stack:
```
WebView::WebContentClient::~WebContentClient() WebContentClient.cpp:45
WebView::WebContentClient::~WebContentClient() WebContentClient.cpp:43
AK::AtomicRefCounted::unref() AtomicRefCounted.h:76
[Inlined] AK::unref_if_not_null(WebView::WebContentClient *) NonnullRefPtr.h:32
[Inlined] AK::RefPtr::clear() RefPtr.h:223
[Inlined] AK::RefPtr::~RefPtr() RefPtr.h:103
WebView::Application::~Application() Application.cpp:70
TestWeb::Application::~Application() Application.cpp:33
TestWeb::Application::~Application() Application.cpp:30
AK::NonnullOwnPtr::clear() NonnullOwnPtr.h:133
AK::NonnullOwnPtr::~NonnullOwnPtr() NonnullOwnPtr.h:50
ladybird_main(Arguments) main.cpp:851
main(int, char **) Main.cpp:46
```

On Windows, it is hit twice. The first time it has an identical call stack to macOS/Linux:
```
WebView::WebContentClient::~WebContentClient() WebContentClient.cpp:45
WebView::WebContentClient::~WebContentClient() WebContentClient.cpp:43
AK::AtomicRefCounted::unref() AtomicRefCounted.h:76
[Inlined] AK::unref_if_not_null(WebView::WebContentClient *) NonnullRefPtr.h:32
[Inlined] AK::RefPtr::clear() RefPtr.h:223
[Inlined] AK::RefPtr::~RefPtr() RefPtr.h:103
WebView::Application::~Application() Application.cpp:70
TestWeb::Application::~Application() Application.cpp:33
TestWeb::Application::~Application() Application.cpp:30
AK::NonnullOwnPtr::clear() NonnullOwnPtr.h:133
AK::NonnullOwnPtr::~NonnullOwnPtr() NonnullOwnPtr.h:50
ladybird_main(Arguments) main.cpp:851
main(int, char **) Main.cpp:46
```

But the second time the call stack shows somehow this is triggered by `Core::ThreadEventQueue` cleanup:
```
ak_trap() Assertions.cpp:134
ak_verification_failed(const char *) Assertions.cpp:172
AK::HashTable::ensure_valid_bucket() HashTable.h:198
WebView::WebContentClient::~WebContentClient() WebContentClient.cpp:44
WebView::WebContentClient::~WebContentClient() WebContentClient.cpp:43
AK::AtomicRefCounted::unref() AtomicRefCounted.h:76
[Inlined] AK::unref_if_not_null(Core::EventReceiver *) NonnullRefPtr.h:32
[Inlined] AK::NonnullRefPtr::~NonnullRefPtr() NonnullRefPtr.h:97
Core::EventReceiver::deferred_invoke::<lambda_0>::~`lambda at C:\Users\ayeteadoe\Development\ladybird\Libraries\LibCore\EventReceiver.cpp:63:27'() EventReceiver.cpp:63
AK::Function<void ()>::CallableWrapper<`lambda at C:\Users\ayeteadoe\Development\ladybird\Libraries\LibCore\EventReceiver.cpp:63:27'>::~CallableWrapper() Function.h:236
AK::Function<void ()>::CallableWrapper<`lambda at C:\Users\ayeteadoe\Development\ladybird\Libraries\LibCore\EventReceiver.cpp:63:27'>::~CallableWrapper() Function.h:229
AK::Function::clear(bool) Function.h:297
AK::Function::~Function() Function.h:100
Core::DeferredInvocationEvent::~DeferredInvocationEvent() Event.h:47
Core::DeferredInvocationEvent::~DeferredInvocationEvent() Event.h:47
AK::NonnullOwnPtr::clear() NonnullOwnPtr.h:133
AK::NonnullOwnPtr::~NonnullOwnPtr() NonnullOwnPtr.h:50
Core::ThreadEventQueue::Private::QueuedEvent::~QueuedEvent() ThreadEventQueue.cpp:31
AK::Vector::clear_with_capacity() Vector.h:464
AK::Vector::clear() Vector.h:453
AK::Vector::~Vector() Vector.h:134
Core::ThreadEventQueue::Private::~Private() ThreadEventQueue.cpp:19
AK::DefaultDelete::operator()(Core::ThreadEventQueue::Private *) DefaultDelete.h:17
AK::OwnPtr::clear() OwnPtr.h:110
AK::OwnPtr::~OwnPtr() OwnPtr.h:45
Core::ThreadEventQueue::~ThreadEventQueue() ThreadEventQueue.cpp:68
Core::ThreadEventQueue::current::<lambda_0>::operator()::<lambda_1>::operator()(void *) ThreadEventQueue.cpp:51
Core::ThreadEventQueue::current::<lambda_0>::operator()::<lambda_1>::__invoke(void *) ThreadEventQueue.cpp:49
```

So it seems to be event loop related. I'll note that I was able to reproduce this same issue when reverting back to the original `WaitForMultipleObjects` Windows event loop impl, so the new IOCP impl doesn't seem to be the root cause. As you can see from the stack trace, at the time of event queue destruction there is still a `DeferredInvocationEvent` queued, where the receiver of said event is a `WebContentClient` instance. The completion of that event occurs after the hash table has been destroyed.

### Root Cause

The above call stacks are all executed on the main thread, and I noticed the reason macOS/Linux doesn't get that second call stack is because the `ThreadEventQueue` destructor function is not called at all when the main thread cleanup is happening.

https://stackoverflow.com/a/6357265 states:
> The destructor function is not called on application exit. It is only called when a thread exits. 

And https://man7.org/linux/man-pages/man3/pthread_key_create.3.html states:
> The destr_function argument, if not NULL, specifies a destructor
       function associated with the key.  When a thread terminates via
       pthread_exit() or by cancelation, destr_function is called with
       arguments the value associated with the key in that thread.  The
       destr_function is not called if that value is NULL.  The order in
       which destructor functions are called at thread termination time
       is unspecified.

Which seems to align with that answer, given we're on the main thread in this context and we are not calling `pthread_exit()`, we don't actually explicitly clean up. So this seems to be expected/valid POSIX behaviour.

The Windows implementation at https://sourceforge.net/p/pthreads4w/code/ci/master/tree/pthread_key_create.c states:

> This function creates a thread-specific data key visible
      to all threads. All existing and new threads have a value
      NULL for key until set using pthread_setspecific. When any
     thread with a non-NULL value for key terminates, 'destructor'
     is called with key's current value for that thread.

So for pthreads4w it seems we invoke the destructor callback regardless of the thread. I also confirmed on macOS that when the `EventLoopImplementation` it shutting down (which has a non-owning reference to the `ThreadEventQueue`), that same single deferred invoke event is pending. So this seems like a `test-web` quirk only, which is why we don't see the issue on Windows when running `Ladybird`.


</details>


---

While addressing this doesn't need to block this PR (the issue still occurs if I fully disable the Fixture setup/teardown and make it a no-op); however, if we do fix it then I could add one more commit (done here or in a subsequent PR) that gets `test-http-test-server.html` actually running in Windows CI which would be pretty nice:

<details>

<summary>windows_ctest_enable.patch</summary>

```patch
diff --git a/Tests/LibWeb/test-web/CMakeLists.txt b/Tests/LibWeb/test-web/CMakeLists.txt
index 77ab79fbee..301af378d6 100644
--- a/Tests/LibWeb/test-web/CMakeLists.txt
+++ b/Tests/LibWeb/test-web/CMakeLists.txt
@@ -19,20 +19,27 @@ elseif (WIN32)
     lagom_windows_bin(test-web CONSOLE)
 endif()
 
-# FIXME: Increase support for building targets on Windows
-if (WIN32 AND ENABLE_WINDOWS_CI)
-    return()
-endif()
-
 if (BUILD_TESTING)
     find_package(Python3 REQUIRED)
 
-    add_test(
-        NAME LibWeb
-        COMMAND $<TARGET_FILE:test-web> --python-executable ${Python3_EXECUTABLE} --dump-failed-ref-tests --per-test-timeout 120 --verbose
-    )
-
-    set_tests_properties(LibWeb PROPERTIES
-        ENVIRONMENT LADYBIRD_SOURCE_DIR=${LADYBIRD_PROJECT_ROOT}
-        TIMEOUT_SIGNAL_NAME SIGTERM)
+    if (WIN32)
+        add_test(
+            NAME LibWeb
+            COMMAND $<TARGET_FILE:test-web>
+                    --python-executable ${Python3_EXECUTABLE}
+                    --dump-failed-ref-tests
+                    --per-test-timeout 120
+                    --verbose
+                    -f "Text\\\\input\\\\test-http-test-server.html"
+        )
+        set_tests_properties(LibWeb PROPERTIES ENVIRONMENT LADYBIRD_SOURCE_DIR=${LADYBIRD_PROJECT_ROOT})
+    else()
+        add_test(
+            NAME LibWeb
+            COMMAND $<TARGET_FILE:test-web> --python-executable ${Python3_EXECUTABLE} --dump-failed-ref-tests --per-test-timeout 120 --verbose
+        )
+        set_tests_properties(LibWeb PROPERTIES
+            ENVIRONMENT LADYBIRD_SOURCE_DIR=${LADYBIRD_PROJECT_ROOT}
+            TIMEOUT_SIGNAL_NAME SIGTERM)
+    endif()
 endif()

```

</details>

If I apply this patch with a hack (not my proposed fix, as this is a symptom only) to address the issue by making the `pthread_key_create` destructor parameter `nullptr`, then the following works:
```
C:\Users\ayeteadoe\Development\ladybird>ctest --preset Windows_Sanitizer_CI -R LibWeb
Test project C:/Users/ayeteadoe/Development/ladybird/Build/sanitizers
    Start 164: LibWeb
1/1 Test #164: LibWeb ...........................   Passed    2.25 sec

100% tests passed, 0 tests failed out of 1

Total Test time (real) =   2.26 sec
```

</details>

Libraries/LibCore/ProcessWindows.cpp
```diff
@@ -55,17 +55,26 @@ ErrorOr<Process> Process::spawn(ProcessSpawnOptions const& options)
...
     BOOL result = CreateProcess(
         NULL,
         (char*)command_line.data(),
-        NULL, // process security attributes
-        NULL, // primary thread security attributes
-        TRUE, // handles are inherited
-        0,    // creation flags
-        NULL, // use parent's environment
-        NULL, // working directory
+        NULL,                                                            // process security attributes
```
R-Goc, line 72: Are the comments needed here? With an LSP you get the argument names inlayed anyways.

ayeteadoe: I found it useful given auto complete is not always active. Going to leave as-is for now

ADKaster:
I'm getting this rebased on master

```
PS C:\Users\Andrew\source\repos\ladybird> python Meta\ladybird.py run test-web -f Test\\input\\test-http-server.html
ninja: Entering directory `C:\Users\Andrew\source\repos\ladybird\Build\release'
ninja: no work to do.
VERIFICATION FAILED: options.file_actions.is_empty() at C:\Users\Andrew\source\repos\ladybird\Libraries\LibCore\ProcessWindows.cpp:43
#0  0x00007ffaffaae9fa in Core::Process::spawn() at C:\Users\Andrew\source\repos\ladybird\Libraries\LibCore\ProcessWindows.cpp:43
#1  0x00007ffaee67f02c in WebView::Process::spawn_and_connect_to_process(bool) at C:\Users\Andrew\source\repos\ladybird\Libraries\LibWebView\Process.cpp:75
#2  0x00007ffaee676965 in WebView::Process::spawn<Requests::RequestClient>(Core::ProcessSpawnOptions&, bool) at C:\Users\Andrew\source\repos\ladybird\Libraries\LibWebView\Process.h:89
#3  0x00007ffaee671bc7 in WebView::launch_server_process<Requests::RequestClient>(AK::StringView, AK::Vector<AK::ByteString, 0, 0>) at C:\Users\Andrew\source\repos\ladybird\Libraries\LibWebView\HelperProcess.cpp:50
```

Is this user error, or code moving around?

ADKaster (changes requested):
This needs updated to behave alongside https://github.com/LadybirdBrowser/ladybird/commit/b5e01df79d8633a06a424a13e5d7fe66fb521174

there seems to be room for a generic "stdout/stderr stream of child" abstraction.

PR #6663, opened 2025-11-01
LibWeb/WebGL: Implement a bunch of stuff for SpectorJS

cqundefine:
This patch set makes the WebGL debugger SpectorJS work on Ladybird which makes for quite a cool demo: https://spector.babylonjs.com/demos/pbr/

### DIFF
--- a/Libraries/LibWeb/WebGL/WebGL2RenderingContextImpl.cpp
+++ b/Libraries/LibWeb/WebGL/WebGL2RenderingContextImpl.cpp
@@ -2436,6 +2436,11 @@ JS::Value WebGL2RenderingContextImpl::get_parameter(WebIDL::UnsignedLong pname)
             return JS::js_null();
         return JS::Value(m_framebuffer_binding);
     }
+    case GL_FRAGMENT_SHADER_DERIVATIVE_HINT: {
+        GLint result { 0 };
+        glGetIntegervRobustANGLE(GL_FRAGMENT_SHADER_DERIVATIVE_HINT, 1, nullptr, &result);
+        return JS::Value(result);
+    }
     case GL_FRONT_FACE: {
         GLint result { 0 };
         glGetIntegervRobustANGLE(GL_FRONT_FACE, 1, nullptr, &result);
@@ -2530,6 +2535,21 @@ JS::Value WebGL2RenderingContextImpl::get_parameter(WebIDL::UnsignedLong pname)
         glGetIntegervRobustANGLE(GL_PACK_ALIGNMENT, 1, nullptr, &result);
         return JS::Value(result);
     }
+    case GL_PACK_ROW_LENGTH: {
+        GLint result { 0 };
+        glGetIntegervRobustANGLE(GL_PACK_ROW_LENGTH, 1, nullptr, &result);
+        return JS::Value(result);
+    }
+    case GL_PACK_SKIP_ROWS: {
+        GLint result { 0 };
+        glGetIntegervRobustANGLE(GL_PACK_SKIP_ROWS, 1, nullptr, &result);
+        return JS::Value(result);
+    }
+    case GL_PACK_SKIP_PIXELS: {
+        GLint result { 0 };
+        glGetIntegervRobustANGLE(GL_PACK_SKIP_PIXELS, 1, nullptr, &result);
+        return JS::Value(result);
+    }
     case GL_POLYGON_OFFSET_FACTOR: {
         GLfloat result { 0.0f };
         glGetFloatvRobustANGLE(GL_POLYGON_OFFSET_FACTOR, 1, nullptr, &result);
@@ -2544,6 +2564,11 @@ JS::Value WebGL2RenderingContextImpl::get_parameter(WebIDL::UnsignedLong pname)
         GLfloat result { 0.0f };
         glGetFloatvRobustANGLE(GL_POLYGON_OFFSET_UNITS, 1, nullptr, &result);
         return JS::Value(result);
+    }
+    case GL_RASTERIZER_DISCARD: {
+        GLboolean result { GL_FALSE };
+        glGetBooleanvRobustANGLE(GL_RASTERIZER_DISCARD, 1, nullptr, &result);
+        return JS::Value(result == GL_TRUE);
     }
     case GL_RED_BITS: {
         GLint result { 0 };
@@ -2583,6 +2608,11 @@ JS::Value WebGL2RenderingContextImpl::get_parameter(WebIDL::UnsignedLong pname)
         GLfloat result { 0.0f };
         glGetFloatvRobustANGLE(GL_SAMPLE_COVERAGE_VALUE, 1, nullptr, &result);
         return JS::Value(result);
+    }
+    case GL_SAMPLER_BINDING: {
+        GLint handle { 0 };
+        glGetIntegervRobustANGLE(GL_SAMPLER_BINDING, 1, nullptr, &handle);
+        return WebGLSampler::create(m_realm, *this, handle);
     }
     case GL_SAMPLES: {
         GLint result { 0 };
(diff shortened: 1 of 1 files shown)


Libraries/LibWeb/WebGL/Types.h
```diff
@@ -15,7 +15,7 @@ using GLenum = unsigned int;
 using GLuint = unsigned int;
 using GLint = int;
 using GLsizei = int;
-using GLintptr = int;
+using GLintptr = long long;
```
gmta, line 18: Typo in commit message: "at" -> "as"

Libraries/LibWeb/WebGL/WebGL2RenderingContext.cpp
```diff
@@ -246,4 +256,9 @@ bool WebGL2RenderingContext::ext_texture_filter_anisotropic_extension_enabled()
     return !!m_ext_texture_filter_anisotropic;
 }
 
+Span<WebIDL::UnsignedLong const> WebGL2RenderingContext::enabled_comprssed_texture_formats() const
```
gmta: Typo: "comprssed" -> "compressed"

Libraries/LibWeb/WebGL/WebGLRenderingContextBase.h
```diff
@@ -32,6 +33,7 @@ class WebGLRenderingContextBase {
     virtual void visit_edges(JS::Cell::Visitor&) = 0;
     virtual OpenGLContext& context() = 0;
     virtual bool ext_texture_filter_anisotropic_extension_enabled() const = 0;
+    virtual Span<WebIDL::UnsignedLong const> enabled_comprssed_texture_formats() const = 0;
```
gmta: This could be `ReadonlySpan` if you want to return a span instead of a vector?

Libraries/LibWeb/WebGL/WebGLRenderingContextImpl.cpp
```diff
@@ -1501,6 +1501,12 @@ JS::Value WebGLRenderingContextImpl::get_parameter(WebIDL::UnsignedLong pname)
         set_error(GL_INVALID_ENUM);
         return JS::js_null();
     }
+    case COMPRESSED_TEXTURE_FORMATS: {
+        auto formats = enabled_comprssed_texture_formats();
+        auto byte_buffer = MUST(ByteBuffer::copy(formats.data(), formats.size() * sizeof(GLenum)));
```
gmta: Alternatively you could use `Span::reinterpret<U>()` so you don't need the `sizeof` here.

Libraries/LibWeb/WebGL/WebGL2RenderingContextImpl.cpp
```diff
@@ -3078,6 +3078,31 @@ Optional<String> WebGL2RenderingContextImpl::get_shader_info_log(GC::Root<WebGLS
...
+    }
+
+    GLint shader_source_length = 0;
+    glGetShaderiv(shader_handle, GL_SHADER_SOURCE_LENGTH, &shader_source_length);
+    if (!shader_source_length)
+        return String {};
+
+    Vector<GLchar> shader_source;
+    shader_source.resize(shader_source_length);
+    glGetShaderSource(shader_handle, shader_source_length, nullptr, shader_source.data());
```
gmta: Using `ByteBuffer` instead of `Vector` might be the more idiomatic way to have a place to put bytes in.

Libraries/LibWeb/WebGL/WebGL2RenderingContextImpl.cpp
```diff
@@ -3129,6 +3129,65 @@ GC::Root<WebGLUniformLocation> WebGL2RenderingContextImpl::get_uniform_location(
     return WebGLUniformLocation::create(m_realm, location);
 }
 
+JS::Value WebGL2RenderingContextImpl::get_vertex_attrib(WebIDL::UnsignedLong index, WebIDL::UnsignedLong pname)
+{
+    switch (pname) {
+    case GL_CURRENT_VERTEX_ATTRIB: {
+        Array<GLfloat, 4> result;
+        result.fill(0);
+        constexpr size_t buffer_size = 4 * sizeof(GLfloat);
```
gmta: You can take the `.span() const` here and use `.reinterpret()` to prevent having to use `sizeof` here.

Libraries/LibWeb/WebGL/WebGL2RenderingContextImpl.cpp
```diff
@@ -3129,6 +3129,65 @@ GC::Root<WebGLUniformLocation> WebGL2RenderingContextImpl::get_uniform_location(
...
+        glGetVertexAttribfvRobustANGLE(index, GL_CURRENT_VERTEX_ATTRIB, 4, nullptr, result.data());
+        auto byte_buffer = MUST(ByteBuffer::copy(result.data(), buffer_size));
+        auto array_buffer = JS::ArrayBuffer::create(m_realm, move(byte_buffer));
+        return JS::Float32Array::create(m_realm, 4, array_buffer);
```
gmta: Instead of using the hardcoded `4`s, you could pass `result.size()` instead.

Libraries/LibWeb/WebGL/WebGL2RenderingContextImpl.cpp
```diff
@@ -3103,6 +3103,12 @@ Optional<String> WebGL2RenderingContextImpl::get_shader_source(GC::Root<WebGLSha
     return String::from_utf8_without_validation(ReadonlyBytes { shader_source.data(), static_cast<size_t>(shader_source_length - 1) });
 }
 
+JS::Value WebGL2RenderingContextImpl::get_uniform(GC::Root<WebGLProgram>, GC::Root<WebGLUniformLocation>)
+{
+    dbgln("FIXME: Impelement get_uniform");
```
gmta: Typo: "Impelement" -> "Implement"

Libraries/LibWeb/WebGL/WebGLRenderingContextImpl.cpp
```diff
@@ -1667,6 +1667,12 @@ Optional<String> WebGLRenderingContextImpl::get_shader_source(GC::Root<WebGLShad
     return String::from_utf8_without_validation(ReadonlyBytes { shader_source.data(), static_cast<size_t>(shader_source_length - 1) });
 }
 
+JS::Value WebGLRenderingContextImpl::get_uniform(GC::Root<WebGLProgram>, GC::Root<WebGLUniformLocation>)
+{
+    dbgln("FIXME: Impelement get_uniform");
```
gmta: Another typo

Libraries/LibWeb/WebGL/WebGL2RenderingContextImpl.cpp
```diff
@@ -3194,6 +3194,18 @@ JS::Value WebGL2RenderingContextImpl::get_vertex_attrib(WebIDL::UnsignedLong ind
     }
 }
 
+WebIDL::LongLong WebGL2RenderingContextImpl::get_vertex_attrib_offset(WebIDL::UnsignedLong index, WebIDL::UnsignedLong pname)
+{
+    if (pname != GL_VERTEX_ATTRIB_ARRAY_POINTER) {
+        set_error(GL_INVALID_ENUM);
+        return -1;
```
gmta: Should we return -1 here instead of 0? Where does it say that? Couldn't find it in the specs

Zaggy1024:
Oooh cool, I was actually thinking about making some fixes for this when I found out about it while debugging some WebGL stuff!

Libraries/LibWeb/WebGL/WebGL2RenderingContextImpl.cpp
cqundefine: In that case we would lose constexpr for dubious benefit, is it really worth it?

Libraries/LibWeb/WebGL/WebGL2RenderingContextImpl.cpp
cqundefine: I changed it to 0, but AFAICT the spec doesn't specify what should be returned and looking at the official WebGL tests I wasn't able to find anything checking the return value of this so I don't think it matters that much. I originally made it -1 because that was done in another place.

Libraries/LibWeb/WebGL/WebGL2RenderingContextImpl.cpp
gmta: Both `.span()` and `.reinterpret()` are constexpr, apart from that I don't think anyone is going to notice if they weren't. It's a bit neater to specify hardcoded values just once imho.

Libraries/LibWeb/WebGL/WebGL2RenderingContextImpl.cpp
gmta: I believe we return 0 in most places, if we find a test case where it requires it to be -1 then let's note that in a comment. Otherwise 0 seems fine :)

Libraries/LibWeb/WebGL/WebGL2RenderingContextImpl.cpp
cqundefine: Done, `.reinterpret()` also shouldn't really be marked as constexpr as it does `reinterpret_cast` unconditionally which is not a thing you can do in constexpr.

Libraries/LibWeb/WebGL/WebGL2RenderingContextImpl.cpp
```diff
@@ -3128,6 +3128,65 @@ GC::Root<WebGLUniformLocation> WebGL2RenderingContextImpl::get_uniform_location(
...
 
+JS::Value WebGL2RenderingContextImpl::get_vertex_attrib(WebIDL::UnsignedLong index, WebIDL::UnsignedLong pname)
+{
+    switch (pname) {
+    case GL_CURRENT_VERTEX_ATTRIB: {
+        Array<GLfloat, 4> result;
+        result.fill(0);
+        glGetVertexAttribfvRobustANGLE(index, GL_CURRENT_VERTEX_ATTRIB, result.size(), nullptr, result.data());
+
+        auto byte_buffer = MUST(ByteBuffer::copy(result.data(), result.span().reinterpret<u8>().size()));
```
gmta: `::copy()` takes a `ReadonlyBytes` which is an alias for `ReadonlySpan<u8>`:

```cpp
auto byte_buffer = MUST(ByteBuffer::copy(result.span().reinterpret<u8>()));
```

gmta (approved):

PR #6664, opened 2025-11-01
LibUnicode: Update ICU4C to 78.1

socram8888:
The UTC has released a new version of ICU4C which also contains a new Unicode 17 version, which should be merged. 

However, it seems there's been many changes to the CLDR that cause tests to break:
```
 FAIL  builtins/Intl/DateTimeFormat/DateTimeFormat.prototype.format.js
    ❌ Suite:  dayPeriod
         Test:   all (failed):
                 ExpectationError: toBe: expected _pm ৫_, got _অপৰাহ্ন ৫_
         Test:   flexible day period rolls over midnight (failed):
                 ExpectationError: toBe: expected _5 at night_, got _5 in the morning_
         Test:   dayPeriod without time (failed):
                 ExpectationError: toBe: expected _pm_, got _অপৰাহ্ন_
    ❌ Suite:  timeZoneName
         Test:   all (failed):
                 ExpectationError: toBe: expected _12/7/2021, GMT_, got _12/7/2021, GMT+0_
 FAIL  builtins/Intl/DisplayNames/DisplayNames.prototype.of.js
    ❌ Suite:  correct behavior
         Test:   option type calendar (failed):
                 ExpectationError: toBe: expected _ISO-8601 Calendar_, got _Gregorian Calendar (ISO 8601 Weeks)_
 FAIL  builtins/Intl/DurationFormat/DurationFormat.prototype.format.js
    ❌ Suite:  correct behavior
         Test:   formats duration correctly (failed):
                 ExpectationError: toBe: expected _1 J, 2 M, 3 W, 3 T, 4 Std., 5 Min., 6 Sek., 7 ms, 8 μs und 9 ns_, got _1 J, 2 M, 3 W, 3 T, 4h, 5 Min., 6 Sek., 7 ms, 8 μs und 9 ns_
 FAIL  builtins/Intl/DurationFormat/DurationFormat.prototype.formatToParts.js
    ❌ Suite:  correct behavior
         Test:   formats duration correctly (failed):
                 ExpectationError: Expected _[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object]_, but got _[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object]_
 FAIL  builtins/Intl/NumberFormat/NumberFormat.prototype.format.js
    ❌ Suite:  style=currency
         Test:   notation=compact, compactDisplay=long (failed):
                 ExpectationError: toBe: expected _١٫٢ ألف US$_, got _‏١٫٢ ألف US$_
         Test:   notation=compact, compactDisplay=short (failed):
                 ExpectationError: toBe: expected _١٫٢ ألف US$_, got _‏١٫٢ ألف US$_
```

I'm not sure if the tests should be adapted to match the new output, or if the Unicode data should be patched to result the same data as before, as some of the changes to the CLDR are kinda weird.

For example, the calendar "iso8601" is now "Gregorian Calendar (ISO 8601 Weeks)" but only for en-US! For all other languages ([including other en variants like en-CA](https://github.com/unicode-org/icu/blob/673038beae65eaad87ed921bc2123ca8d1b0c2a2/icu4c/source/data/lang/en_CA.txt#L35)) I've seen it's still simply "ISO 8601 Calendar".

Others like changing "pm" for "অপৰাহ্ন ৫" is (according to ChatGPT, as I am not familiar with the language) simply changing a less formal with a more formal version, which I guess are okay?


trflynn89:
UCD/CLDR updates breaking tests is pretty normal, and we typically just update our tests to match the new expectations (assuming they don't uncover a bug on our end).

But we wait until [test262](https://github.com/tc39/test262) has been updated to match the new versions before we update, as that's a signal that other major engines have updated too.

socram8888:
Alright, thanks! I'll be having a look at test262. I see there are many PRs pending regarding these ICU changes.

socram8888:
test262 still not updated: https://github.com/tc39/test262/pulls?q=is%3Apr+is%3Aopen+CLDR

trflynn89:
It looks like some PRs are starting to land in test262 for Unicode 17, so I think we can move forward with this change. See:
`https://github.com/tc39/test262/pull/4572`
`https://github.com/tc39/test262/pull/4573`

Coincidentally, vcpkg itself now supports ICU 78.1: https://vcpkg.io/en/package/icu

The reason we have ICU as an overlay port is that upstream vcpkg was stuck on an old ICU version for ages. But it seems they resolved that issue. So we can actually remove our overlay port and just update the version in vcpkg.json instead.

(If you're busy or if updating the tests becomes a pain, let me know and I'm happy to do this - I wrote them and have gone through this process dozens of times)

socram8888:
Working on it! I'll let you know if I find myself unable :)

socram8888:
I've added an additional commit to rework a bit the Intl-related tests. In general:
```js
test("all", () => {
    data.forEach(d => {
```
Is now:
```js
data.forEach(d => {
    test(`prop1=${d.prop1}, prop2=${d.prop2}`, () => {
```
It makes much easier to know which exact "expect" failed, and allows failing multiple cases instead of halting on the first failed "expect":
screenshot
Before you'd see simply "Test: all" without any exact reference on what exactly was failing, which was kinda painful.

There seems to be some fails in the IdnaTestV2.window.html. Working on it.

shannonbooth:
The failure in IdnaTestV2.window.html can be solved by rebaslining the test - that was also updated to unicode 17. You should be able to fix this by updating IdnaTestV2.json in our tree (https://github.com/LadybirdBrowser/ladybird/blob/master/Tests/LibWeb/Text/input/wpt-import/url/resources/IdnaTestV2.json) with the latest version of https://github.com/web-platform-tests/wpt/blob/master/url/resources/IdnaTestV2.json

After that, ./Meta/ladybird.py run test-web --rebaseline can be run to update the test expectations to align with WPT

socram8888:
@shannonbooth thanks for the tip! I've updated IdnaTestV2 and now all test are passing, save for one on Mac which I'm not sure if it's my fault.

shannonbooth:
Cool! And yeah, I think the CI issue is unrelated. Could the commit for updating IDNA be squashed into the ICU update? Currently the commits are not atomic since just applying the ICU update commit on its own breaks test expectations

socram8888:
@shannonbooth you're right! I've squashed, and also rebased on master. Should be good now!

EDIT: One comment regarding the removed ""flexible day period rolls over midnight": it seems that this is no longer true, and why I removed the test. On Chrome for instance:
image

Seems like anything past 23:59:59 is now "morning", which doesn't make any sense to me, but I'm not the Unicode committee.

Libraries/LibJS/Tests/builtins/Intl/DateTimeFormat/DateTimeFormat.prototype.format.js
```diff
@@ -354,23 +354,25 @@ describe("dayPeriod", () => {
...
         expect(en.format(date2)).toBe("11 at night");
     });
+});
 
-    test("noon", () => {
-        const date0 = Date.UTC(2017, 11, 12, 12, 0, 0, 0);
-        const date1 = Date.UTC(2017, 11, 12, 12, 1, 0, 0);
-        const date2 = Date.UTC(2017, 11, 12, 12, 0, 1, 0);
-        const date3 = Date.UTC(2017, 11, 12, 12, 0, 0, 500);
+describe("dayPeriod at noon", () => {
```
trflynn89, line 359: Let's keep this a `test`

Libraries/LibJS/Tests/builtins/Intl/DateTimeFormat/DateTimeFormat.prototype.format.js
```diff
@@ -398,16 +400,18 @@ describe("dayPeriod", () => {
...
+});
 
-    test("dayPeriod without time", () => {
-        // prettier-ignore
-        const data = [
-            { dayPeriod: "narrow", en0: "in the afternoon", en1: "in the morning", ar0: "بعد الظهر", ar1: "صباحًا", as0: "pm", as1: "am"},
-            { dayPeriod: "short", en0: "in the afternoon", en1: "in the morning", ar0: "بعد الظهر", ar1: "ص", as0: "PM", as1: "AM"},
-            { dayPeriod: "long", en0: "in the afternoon", en1: "in the morning", ar0: "بعد الظهر", ar1: "صباحًا", as0: "PM", as1: "AM"},
-        ];
+describe("dayPeriod without time", () => {
```
trflynn89, line 405: Same here

Libraries/LibJS/Tests/builtins/Intl/DurationFormat/DurationFormat.prototype.formatToParts.js
```diff
@@ -57,6 +57,8 @@ describe("correct behavior", () => {
             { type: "literal", value: " ", unit: "nanosecond" },
             { type: "unit", value: "ns", unit: "nanosecond" },
         ]);
+    });
+    test("en locale", () => {
```
trflynn89, line 63: Let's put newlines between `test`s (here and below)

Libraries/LibJS/Tests/builtins/Intl/DateTimeFormat/DateTimeFormat.prototype.format.js
```diff
@@ -338,22 +338,6 @@ describe("dayPeriod", () => {
             expect(as.format(d1)).toBe(d.as1);
         });
     });
-
-    test("flexible day period rolls over midnight", () => {
```
trflynn89, line 342: Instead of removing this test, let's just update the expectation:
```js
expect(en.format(date1)).toBe("5 in the morning");
```

trflynn89: Ah actually, now I see - it looks like the day period in `en` no longer stretches across midnight. In CLDR 46, we had a `night1` period of [21:00, 06:00):

https://github.com/unicode-org/cldr-json/blob/46.0.0/cldr-json/cldr-core/supplemental/dayPeriods.json#L446-L449

Whereas in CLDR 48, we have [21:00, 00:00):

https://github.com/unicode-org/cldr-json/blob/48.0.0/cldr-json/cldr-core/supplemental/dayPeriods.json#L446-L449

So let's actually change the test to pick a locale for which the day period crosses midnight still. It looks like `hi-Latn` is a good candidate, with [20:00, 04:00):

https://github.com/unicode-org/cldr-json/blob/48.0.0/cldr-json/cldr-core/supplemental/dayPeriods.json#L788-L791

So we can do:
```js
test("flexible day period rolls over midnight", () => {
    const en = new Intl.DateTimeFormat("hi-Latn", {
        hour: "numeric",
        dayPeriod: "short",
        timeZone: "UTC",
    });

    // For the hi-Latn locale, these times (03:00 and 23:00) fall in the flexible day period range of
    // [20:00, 04:00), on either side of midnight.
    const date1 = Date.UTC(2017, 11, 12, 3, 0, 0, 0);
    const date2 = Date.UTC(2017, 11, 12, 23, 0, 0, 0);

    expect(en.format(date1)).toBe("3 raat");
    expect(en.format(date2)).toBe("11 raat");
});
```

(raat is defined as the night period name here:)
https://github.com/unicode-org/cldr-json/blob/48.0.0/cldr-json/cldr-cal-roc-full/main/hi-Latn/ca-roc.json#L233

socram8888: Ack, will do!

Libraries/LibJS/Tests/builtins/Intl/DateTimeFormat/DateTimeFormat.prototype.format.js
socram8888: I split these because they have multiple test cases. With a single `test` if the case in `data[0]` fails the other are never tested, which was kinda painful to fix those three sub-tests.

I can move it back inside `dayPeriod` but I'd really like to leave the order `forEach(test)` rather than `test(forEach)`.

trflynn89: I just meant to literally call the section itself a `test`, not a `describe`. i.e.:
```js
test("dayPeriod at noon", () => {
```

trflynn89: Oh wait! nevermind me, I missed that the tests are inside the section. Sorry!

PR #6670, opened 2025-11-03
LibWeb: Add synthetic height for contenteditable elements

mikiubo:
Contenteditable elements that would otherwise have zero height now get a minimum height equal to their line-height. This ensures they remain clickable and usable for text editing.

This fixes the WPT test:
contenteditable/synthetic-height.html

### DIFF
--- a/Libraries/LibWeb/Layout/BlockFormattingContext.cpp
+++ b/Libraries/LibWeb/Layout/BlockFormattingContext.cpp
@@ -710,6 +710,12 @@ CSSPixels BlockFormattingContext::compute_auto_height_for_block_level_element(Bo
             return marker_line_height;
     }
 
+    // AD-HOC: Contenteditable elements must have a minimum height (line-height) when empty, to remain clickable and
+    //         usable for text input, even though this is not specified.
+    //         See: https://github.com/w3c/editing/issues/70.
+    if (auto const* element = as_if<DOM::Element>(box.dom_node()); element && element->is_editing_host())
+        return box.computed_values().line_height();
+
     // 4. zero, otherwise
     return 0;
 }
--- a/Tests/LibWeb/Layout/expected/empty-editable-shows-cursor.txt
+++ b/Tests/LibWeb/Layout/expected/empty-editable-shows-cursor.txt
@@ -1,86 +1,86 @@
 Viewport <#document> at [0,0] [0+0+0 800 0+0+0] [0+0+0 600 0+0+0] [BFC] children: not-inline
-  BlockContainer <html> at [0,0] [0+0+0 800 0+0+0] [0+0+0 56 0+0+0] [BFC] children: not-inline
+  BlockContainer <html> at [0,0] [0+0+0 800 0+0+0] [0+0+0 74 0+0+0] [BFC] children: not-inline
     BlockContainer <(anonymous)> at [0,0] [0+0+0 800 0+0+0] [0+0+0 0 0+0+0] children: inline
       TextNode <#text> (not painted)
-    BlockContainer <body> at [8,8] [8+0+0 784 0+0+8] [8+0+0 40 0+0+8] children: not-inline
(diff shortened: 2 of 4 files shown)


Libraries/LibWeb/Layout/BlockFormattingContext.cpp
```diff
@@ -672,6 +672,15 @@ CSSPixels BlockFormattingContext::compute_auto_height_for_block_level_element(Bo
         }
     }
 
+    // AD-HOC: Contenteditable elements must have a minimum height (line-height) when empty
+    //         to remain clickable and usable for text input, even though this is not specified.
+    //         See: https://github.com/w3c/editing/issues/70
+    if (box.dom_node() && !box.is_anonymous() && !box.is_generated_for_pseudo_element()) {
+        if (auto const* element = dynamic_cast<DOM::Element const*>(box.dom_node()); element && element->is_editing_host()) {
```
gmta: We use `as_if` for this:

```cpp
if (auto const* element = as_if<DOM::Element>(box.dom_node()); element && element->is_editing_host()) {
```

Libraries/LibWeb/Layout/BlockFormattingContext.cpp
gmta, line 1: Commit message: please auto-wrap at 72 characters.

Libraries/LibWeb/Layout/BlockFormattingContext.cpp
```diff
@@ -672,6 +672,15 @@ CSSPixels BlockFormattingContext::compute_auto_height_for_block_level_element(Bo
         }
     }
 
+    // AD-HOC: Contenteditable elements must have a minimum height (line-height) when empty
+    //         to remain clickable and usable for text input, even though this is not specified.
```
gmta: Please auto-wrap comments at column 120.

mikiubo: do you use an automatic tool to do it?
i try the 'clang-format -i' as shown in the doc but it seems that does not work?

gmta: Basically all text editors and IDEs have a setting to autowrap at any given column!

mikiubo: Ok, I’m using CLion, but apparently I have some misconfiguration, so it’s not working properly -.-
I know that each IDE has an auto-wrap setting. I’m asking because it seems like you want a different behavior specifically for comments.

For now, I’ve changed it manually — let me know if it’s not okay.

Also, I don’t know why the CI is failing (it seems to be failing on something unrelated to my changes). I’ll try to rebase and relaunch the CI, and I’ll let you know if it still fails.

Thanks.

Libraries/LibWeb/Layout/BlockFormattingContext.cpp
```diff
@@ -717,6 +717,15 @@ CSSPixels BlockFormattingContext::compute_auto_height_for_block_level_element(Bo
...
     }
 
+    // AD-HOC: Contenteditable elements must have a minimum height (line-height) when empty, to remain clickable and
+    //         usable for text input, even though this is not specified.
+    //         See: https://github.com/w3c/editing/issues/70.
+    if (box.dom_node() && !box.is_anonymous() && !box.is_generated_for_pseudo_element()) {
+        if (auto const* element = as_if<DOM::Element>(box.dom_node()); element && element->is_editing_host()) {
+            return box.computed_values().line_height();
+        }
+    }
```
gmta: I think this block can be simplified to:

```cpp
if (auto const* element = as_if<DOM::Element>(box.dom_node()); element && element->is_editing_host())
    return box.computed_values().line_height();
```

`as_if<T>(pointer)` already checks `nullptr` so `box.dom_node()` is not necessary. Also I don't think anonymous boxes or boxes generated for pseudo elements can ever be editing hosts?

mikiubo: right! i think i fixed

gmta (approved):

Issue #5679, opened 2025-08-01
RequestServer: Support downgrade from HTTP/3 if a server is misconfigured to report HTTP/3 support


Lubrsi:
If a server reports that it supports HTTP/3 in Alt-Svc, curl will only attempt a connection by HTTP/3 (as if `CURLOPT_HTTP_VERSION` was set to `CURL_HTTP_VERSION_3ONLY`). However, if the server was misconfigured to support it, we will forever fail to connect to the server until the cache entry is removed (e.g. manually or expires). Two examples I've encountered so far is `grafana.app.ladybird.org` and `xp.atlassian.com`.

We can either detect a connection failed with HTTP/3 and not use Alt-Svc on retry (which can take a while to figure out if there's a long timeout, or the server is just slow), or try and patch curl to use it's built-in downgrade mechanism (I believe it's called Happy Eyeballs?) when Alt-Svc is in use.

Lubrsi:
Relevant curl PR: https://github.com/curl/curl/pull/17152

Lubrsi:
From @gmta, another website that erroneously reports HTTP/3 support: https://duotrigordle.com/

AtkinsSJ:
commafeed.com has the same issue: #5485

nevakrien:
Hopefully will be fixed by the next release (or the one after that). right now waiting on the last code review approval (famous last words). @Lubrsi, I would really like it if you put a specific reproducible example and ideally a test so the PR can address it directly. 
if you just give me a script, i could move that to a test.

Lubrsi:
I wasn't quite sure how to write a curl test for this, so I wrote a shell script that does this with node:
```bash
#!/bin/bash
# https://stackoverflow.com/questions/21397809/create-a-trusted-self-signed-ssl-certificate-for-localhost-for-use-with-expre
# https://stackoverflow.com/questions/34817633/i-need-to-create-a-hundred-self-signed-certificates-with-no-pem-pass-phrase

rm altsvc.txt

openssl req -x509 -newkey rsa:2048 -keyout keytmp.pem -out cert.pem -days 365 -subj "/C=AU/ST=Some-State/L=City/O=Company/OU=Unit/CN=localhost" -passout pass:
openssl rsa -in keytmp.pem -out key.pem -passin pass:

npm install express

node -e "const express = require('express')
const app = express()
const https = require('https')
const fs = require('fs')
const port = 3000
const { exec } = require('child_process')
[...]

PR #6673, opened 2025-11-03
RequestServer: Immediately attempt HTTP/3 connection 

tcl3:
We now attempt a HTTP/3 connection immediately and fall back to earlier HTTP versions if it isn't available, we also ignore `h3` if present in the `alt-svc` header, as this breaks sites that advertise HTTP/3 support with this header, but don't actually support it.

Related to #5679, in that this fixes http://grafana.app.ladybird.org and similarly configured sites. The proper fix for this issue is to support graceful downgrading from HTTP/3 to HTTP/2.

### DIFF
--- a/Services/RequestServer/Request.cpp
+++ b/Services/RequestServer/Request.cpp
@@ -348,6 +348,7 @@ void Request::handle_fetch_state()
     set_option(CURLOPT_ALTSVC, m_alt_svc_cache_path.characters());
     // FIXME: This ignores h3 in the alt-svc header, as downgrading to HTTP/2 on failure is not yet supported.
     set_option(CURLOPT_ALTSVC_CTRL, CURLALTSVC_H1 | CURLALTSVC_H2);
+    set_option(CURLOPT_HTTP_VERSION, CURL_HTTP_VERSION_NONE);
 
     set_option(CURLOPT_CUSTOMREQUEST, m_method.characters());
     set_option(CURLOPT_FOLLOWLOCATION, 0);


Lubrsi:
I remember trying this and some sites would take a very long time to load because curl would take a long time attempting HTTP/3 with exponential backoff before attempting a downgrade. I can't remember the sites off the top of my head though. Plus it will attempt HTTP/3 again for any future requests even when it does the downgrade for one of them, since we ask for HTTP/3 for all of them.

gmta:
Note that http://grafana.app.ladybird.org/ is no longer a test case for this PR since it has since been updated with HTTP/2 + HTTP/3 support.

tcl3:
Closing, as this clearly isn't a great solution. Ideally curl would handle alt-svc fallback internally - there is an open curl PR for this (https://github.com/curl/curl/pull/17152), but I'm not sure what its current status is.

nevakrien:
@tc https://github.com/curl/curl/pull/17152 is about trivial fails that are always safe to retry. because downgrading HTTP versions does not fall onto that category (and might actually not be allowed by the standard? like its murky) its out of scope for that curl PR. 

if an HTTP downgrade over an alt-svc retry is desirable please raise this as a separate issue for curl so that it can be addressed.
also if we do end up going with it a reproducible example using the curl CLI would go a long way to add this to the tests of curl to guarantee support.

implementing this on the lady bird side might actually be easier to you guys. curl can be configured to keep the connection opened. if it is then a retry on your end is identical to a retry within curl. and having the logic on your end would allow you to be more liberal on security as you seem to want to do. 

I am fine helping out if thats needed

PR #6674, opened 2025-11-03
LibWeb/WebGL: Don't set uniform if it's null or not from current context

Lubrsi:
Fixes regression on noclip.website where most of the textures were blank, since we overrided anything in uniform location 0 if the location was null, which is a valid location.

### DIFF
--- a/Libraries/LibWeb/WebGL/WebGL2RenderingContextImpl.cpp
+++ b/Libraries/LibWeb/WebGL/WebGL2RenderingContextImpl.cpp
@@ -248,89 +248,195 @@ void WebGL2RenderingContextImpl::tex_sub_image3d(WebIDL::UnsignedLong target, We
 void WebGL2RenderingContextImpl::uniform1ui(GC::Root<WebGLUniformLocation> location, WebIDL::UnsignedLong v0)
 {
     m_context->make_current();
-    glUniform1ui(location ? location->handle() : 0, v0);
+
+    // "If the passed location is not null and was not obtained from the currently used program via an earlier call to
+    //  getUniformLocation, an INVALID_OPERATION error will be generated. If the passed location is null, the data
+    //  passed in will be silently ignored and no uniform variables will be changed."
+    if (!location) [[unlikely]]
+        return;
+
+    auto handle_or_error = location->handle(this);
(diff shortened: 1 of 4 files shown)


Libraries/LibWeb/WebGL/WebGL2RenderingContextImpl.cpp
```diff
@@ -248,177 +248,333 @@ void WebGL2RenderingContextImpl::tex_sub_image3d(WebIDL::UnsignedLong target, We
...
+    //  passed in will be silently ignored and no uniform variables will be changed."
+    if (!location) [[unlikely]]
+        return;
+
+    auto handle_or_error = location->handle(this);
+    if (handle_or_error.is_error()) [[unlikely]] {
+        set_error(GL_INVALID_OPERATION);
+        return;
+    }
+    auto location_handle = handle_or_error.release_value();
```
gmta, line 263: Seems like we could/should use `SET_ERROR_VALUE_IF_ERROR` for all of these? And maybe it should actually use the returned error code instead of accepting the error as a second argument (or introduce a variant that does so).

Libraries/LibWeb/WebGL/WebGL2RenderingContextImpl.cpp
```diff
@@ -248,177 +248,333 @@ void WebGL2RenderingContextImpl::tex_sub_image3d(WebIDL::UnsignedLong target, We
 void WebGL2RenderingContextImpl::uniform1ui(GC::Root<WebGLUniformLocation> location, WebIDL::UnsignedLong v0)
 {
     m_context->make_current();
-    glUniform1ui(location ? location->handle() : 0, v0);
+
+    // "If the passed location is not null and was not obtained from the currently used program via an earlier call to
+    //  getUniformLocation, an INVALID_OPERATION error will be generated. If the passed location is null, the data
+    //  passed in will be silently ignored and no uniform variables will be changed."
```
gmta, line 254: Maybe we should make this note just once for all `uniform*` methods? Feels like a lot of copy to update if the spec ever changes, and this behavior does not feel that unintuitive that it warrants so many reminders imo.

PR #6675, opened 2025-11-03
LibWebView+RequestServer: Add UI to manage browsing data caches 

trflynn89:
This adds a settings section to allow users to clear the HTTP disk cache and cookies / local storage / session storage. There are a few options to limit this action to a specific time range (e.g. "last hour"). The user is informed how much disk space is being used currently, and how much will be removed given the selected time range.

The idea is that in the future, we can add more settings here to auto-delete data on exit, disable caching altogether, etc.

Some screenshots:


trflynn89:
EDIT: Those check boxes were a little off-center vertically, which is now fixed:
center

Base/res/ladybird/about-pages/settings.html
```diff
@@ -495,6 +518,45 @@ <h3 id="site-settings-title" class="dialog-title"></h3>
...
+                    <label for="clear-browsing-data-cached-files">
+                        Cached files<span id="clear-browsing-data-cached-files-size"></span>
+                        <p class="description">Remove items that help pages load faster</p>
+                    </label>
+                </div>
+                <div class="input-field-container">
+                    <input id="clear-browsing-data-site-data" type="checkbox" value="" checked />
+                    <label for="clear-browsing-data-site-data">
+                        Cookies and site data<span id="clear-browsing-data-site-data-size"></span>
+                        <p class="description">Remove items that may sign you out of most sites</p>
```
konradekk, line 551: (nit: nice explanation, but site data can also contain some locally stored state of the app, e.g. something that hasn’t been synchronized with or stored on the server yet. not sure how this could be put into words, but would be nice to have that mentioned along with the current explanation somehow…)

trflynn89: The language here matches how both chrome and firefox describe this data

Base/res/ladybird/about-pages/settings/privacy.js
```diff
@@ -1,15 +1,125 @@
...
+    const unit = BYTE_UNITS[index];
+
+    if (!BYTE_FORMATTERS[unit]) {
+        BYTE_FORMATTERS[unit] = new Intl.NumberFormat([], {
+            style: "unit",
+            unit: unit,
+            unitDisplay: unit === "byte" ? "long" : "short",
+            maximumFractionDigits: 1,
+        });
+    }
```
konradekk: this is suspicious: it seems it should be global, but then you would not be able to make it depend on `unit`/`index`;

maybe something like:

```ts
    const createFormat = (unit) => new Intl.NumberFormat([], {
        style: "unit",
        unit: unit,
        unitDisplay: unit === "byte" ? "long" : "short",
        maximumFractionDigits: 1,
    });

    const BYTE_FORMATTERS = Object.fromEntries(
        ["byte", "kilobyte", "megabyte", "gigabyte", "terabyte"]
            .map(unit => [unit, createFormat(unit)]
    );
```

Services/RequestServer/Cache/CacheIndex.cpp
```diff
@@ -101,6 +101,7 @@ ErrorOr<CacheIndex> CacheIndex::create(Database::Database& database)
     statements.select_entry = TRY(database.prepare_statement("SELECT * FROM CacheIndex WHERE cache_key = ?;"sv));
     statements.update_response_headers = TRY(database.prepare_statement("UPDATE CacheIndex SET response_headers = ? WHERE cache_key = ?;"sv));
     statements.update_last_access_time = TRY(database.prepare_statement("UPDATE CacheIndex SET last_access_time = ? WHERE cache_key = ?;"sv));
+    statements.estimate_cache_size_accessed_since = TRY(database.prepare_statement("SELECT SUM(data_size) + SUM(LENGTH(response_headers)) FROM CacheIndex WHERE last_access_time >= ?;"sv));
```
gmta: Instead of `length()`, you should use `octet_length()` on `response_headers` to get the number of bytes.

Libraries/LibWebView/StorageJar.cpp
```diff
@@ -197,6 +205,21 @@ Vector<String> StorageJar::TransientStorage::get_keys(StorageEndpointType storag
     return keys;
 }
 
+Requests::CacheSizes StorageJar::TransientStorage::estimate_storage_size_accessed_since(UnixDateTime since) const
+{
+    Requests::CacheSizes sizes;
+
+    for (auto const& [key, entry] : m_storage_items) {
+        auto size = key.storage_key.byte_count() + key.bottle_key.byte_count() + entry.value.byte_count();
```
gmta, line 229: Maybe we could have `HashTable` and `HashMap` expose the number of bytes used for a single slot instead?

Base/res/ladybird/about-pages/settings.html
```diff
@@ -372,6 +388,13 @@ <h1>Ladybird Settings</h1>
         <div class="card">
             <div class="card-header">Privacy</div>
             <div class="card-body">
+                <div class="card-group inline-container">
+                    <span>Browsing Data</span>
+                    <button id="clear-browsing-data" class="secondary-button">Remove...</button>
```
gmta: The verb you use in the modal dialog is "Clear" in the title but "Remove" in the primary action. I'm slightly leaning towards using "Clear" everywhere, but I don't mind as long as it's consistent :)

Base/res/ladybird/about-pages/settings.html
```diff
@@ -495,6 +518,45 @@ <h3 id="site-settings-title" class="dialog-title"></h3>
             </div>
         </dialog>
 
+        <dialog id="clear-browsing-data-dialog">
+            <div class="dialog-header">
+                <h3 id="clear-browsing-data-title" class="dialog-title">Clear Browsing Data</h3>
+                <button id="clear-browsing-data-close" class="close-button dialog-button">&times;</button>
+            </div>
+            <div class="dialog-body">
+                <p id="clear-browsing-data-total-size" class="description"></p>
```
gmta, line 527: The text in this paragraph gets slightly cut off at the bottom for me:

image

Services/RequestServer/Cache/CacheIndex.cpp
trflynn89: Ah TIL `length` decodes code points.

🤔  should we be using BLOB to store `ByteString`s in the DB in that case as well?

gmta: That makes sense, yeah!

Libraries/LibWebView/StorageJar.cpp
trflynn89: I started doing this but realized that doesn't work well for structs with heap-allocated data. i.e.:
```c++
HashTable<String> table;
table.set("Long string is more than just a few bytes"_string);
```

From within `HashTable`, we would only see used bucket taking up `sizeof(String)` (8) bytes.

gmta: Right - nevermind this then! It's an estimate anyway ~~, since the actual space of disk depends on the actual row size and page size used by sqlite.~~ :)

Base/res/ladybird/about-pages/settings.html
trflynn89: Huh, can you show me how the following renders for you?
```html
<head>
    <style>
        * {
            color-scheme: light dark;
            font-family: sans-serif;
        }
        .description {
            margin-bottom: 10px;
            font-size: 14px;
            opacity: 0.7;
        }
    </style>
</head>
<body>
    <p class="description">Your browsing data is currently using 0 of disk space</p>
    <p>Your browsing data is currently using 0 of disk space</p>
</body>
```

And then again with the `opacity: 0.7;` removed?

Oddly enough, I get this rendering, where the bottom of the `g` looks cut off on the first line:
Screenshot 2025-11-07 at 7 43 23 AM

But it looks fine with full opacity:
Screenshot 2025-11-07 at 7 44 47 AM

gmta: With `opacity: 0.7`:

| Scale 1x | Scale 2x |
|--|--|
| image | image |

Without `opacity: 0.7`:

| Scale 1x | Scale 2x |
|--|--|
| image | image |

Maybe we're messing up a clip in the display list when opacity is set?

trflynn89: Could be! I opened #6738 for this

Base/res/ladybird/about-pages/settings.html
trflynn89: Went with "Clear"

PR #6676, opened 2025-11-03
LibWeb/WebGL: Rearrange and deduplicate WebGL1 implementations from WebGL2

cqundefine:
This PR while it looks massive is mostly just moving code around. This allows us to not have to maintain two copies of the same functions with in 99% of cases the same implementations.


kalenikaliaksandr (approved):

Libraries/LibWeb/WebGL/WebGLRenderingContextImpl.cpp
```diff
@@ -1,14 +1,16 @@
...
 
 #define GL_GLEXT_PROTOTYPES 1
-#include <GLES2/gl2.h>
-#include <GLES2/gl2ext.h>
+
+#include <GLES3/gl3.h>
 extern "C" {
+#include <GLES2/gl2ext.h>
 #include <GLES2/gl2ext_angle.h>
 }
```
kalenikaliaksandr, line 15: why do we switch from gl2 to gl3 with this change? could you add some explanation in the commit description or even move into separate commit?

kalenikaliaksandr (changes requested):

Libraries/LibWeb/WebGL/WebGLRenderingContextImpl.cpp
cqundefine: The reason behind this is that some defines used in functions like getParameter that are WebGL2 specific are only defined in the ES3 headers not in ES2. In practice anything that is ES3 exclusive is locked behind a WebGL2 check anyways.

Libraries/LibWeb/WebGL/WebGLRenderingContextBase.h
```diff
@@ -117,6 +128,15 @@ class WebGLRenderingContextBase {
...
 
 protected:
+    static Vector<GLchar> null_terminated_string(StringView string)
+    {
+        Vector<GLchar> result;
+        for (auto c : string.bytes())
+            result.append(c);
+        result.append('\0');
+        return result;
+    }
```
konradekk, line 139: consider growing the vector to the target size and then copying the contents instead. (i believe you could re-use some existing copying mechanisms from the internals of `StringView`, `ByteString`, and `Vector`)

cqundefine: I added an `ensure_capacity`. I'm not sure if there's a point in changing this function too much as it's just getting moved and I feel like it's kind of out of scope for this PR.

kalenikaliaksandr: yeah, let's keep this PR focused on deduplication rather then rereviewing all code that has been moved around 😅

PR #6679, opened 2025-11-03
LibMedia: Begin audio seeking immediately if using an accurate seek

Zaggy1024:
When accurate seeking, we can assume that the chosen timestamp will be the target timestamp for the seek. Therefore, we can run the audio seeks simultaneously with the video seeks to allow them to finish (probably) before the video seeks themselves complete, making seeking slightly more responsive.

### DIFF
--- a/Libraries/LibMedia/PlaybackStates/SeekingStateHandler.h
+++ b/Libraries/LibMedia/PlaybackStates/SeekingStateHandler.h
@@ -103,6 +103,25 @@ private:
         return count;
     }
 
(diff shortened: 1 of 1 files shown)


Libraries/LibMedia/PlaybackStates/SeekingStateHandler.h
gmta, line 1: Commit message: I believe there's a missing in this sentence:

> [..] to allow them to finish (probably) the video seeks themselves complete, [..]

gmta (approved):

PR #6683, opened 2025-11-03
Base: Align text properly next to icon in version page

Pinguin2001:
Before:
Screenshot 2025-11-03 215744

After:
Screenshot 2025-11-03 215721

### DIFF
--- a/Base/res/ladybird/templates/version.html
+++ b/Base/res/ladybird/templates/version.html
@@ -7,6 +7,10 @@
             html {
                 color-scheme: light dark;
             }
+            header {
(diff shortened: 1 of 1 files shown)


Pinguin2001:
Pass: 5170, Fail: 1, Skipped: 278, Timeout: 0, Crashed: 0
Fail: Text/input/wpt-import/css/css-transitions/transitioncancel-003.html

The change should not affect wpt tests at all, it also passes locally:

image

I cannot see why it wouldn't pass on macOS, its the same engine after all

PR #6685, opened 2025-11-03
LibMedia: Correct FFmpeg's badly-muxed WebM files' timestamps

Zaggy1024:
## 7ms video frame offset
FFmpeg's libavcodec in version 59.30.100 and older had an issue that affected Matroska files containing Opus audio (pretty much every WebM file out there), which caused video frames' timestamps to be offset forward by 7ms. These commits implement a fix-up step to correct the timestamps in such files.

### Reasoning
Opus has a segment of audio before the start of the actual decoded audio called pre-skip, which is normally 6.5ms that the decoder drops before it actually begins outputting audio. Matroska has the CodecDelay element, which is intended to offset the timestamps of audio blocks backwards so that this pre-skip audio falls before the beginning of the file. The element should contain the Opus pre-skip converted to Matroska time units. However, the affected FFmpeg versions assumed that Matroska was not allowed to contain negative timestamps, so it would see that the audio timestamps fell before the zero timestamp, and offset every block in every track in the file forward to compensate.

In addition, the affected versions of FFmpeg measured the Opus pre-skip value by the actual sample rate of the track, while Opus specifies that it should always be measured according to a fixed sample rate of 48kHz. This resulted in some files having a smaller value for CodecDelay than the actual pre-skip that the Opus headers specified.

In order to fix up affected files, we must:
- Ignore the existing CodecDelay elements, since FFmpeg was already applying the timestamp offsets to the blocks before muxing them into the output.
- Scale the CodecDelay of Opus tracks according to the ratio of the 48kHz sample rate to the audio's actual sample rate.
- Find the largest (corrected) CodecDelay in all tracks, round that to the Matroska time units, and offset all tracks backward by that time.

The actual delay is 6.5ms, but in all files affected by this, the time units are a full millisecond, so the offset rounds up to 7ms.

## Seeking pre-roll
Opus tracks have a SeekPreRoll element to specify the amount of audio that needs to be decoded before the output converges, which is usually 80ms. Subtracting this pre-roll offset from all seeks ensures that the audio doesn't sound muffled briefly after a seek. This is especially noticeable when seeking to a second interval in [WPT's movie_300.webm](https://wpt.live/media/movie_300.webm) file.

### DIFF
--- a/Libraries/LibMedia/Containers/Matroska/Reader.cpp
+++ b/Libraries/LibMedia/Containers/Matroska/Reader.cpp
@@ -57,6 +57,7 @@ constexpr u32 TRACK_LANGUAGE_BCP_47_ID = 0x22B59D;
 constexpr u32 TRACK_CODEC_ID = 0x86;
 constexpr u32 TRACK_CODEC_PRIVATE_ID = 0x63A2;
 constexpr u32 TRACK_CODEC_DELAY_ID = 0x56AA;
+constexpr u32 TRACK_SEEK_PRE_ROLL_ID = 0x56BB;
 constexpr u32 TRACK_TIMESTAMP_SCALE_ID = 0x23314F;
 constexpr u32 TRACK_OFFSET_ID = 0x537F;
 constexpr u32 TRACK_VIDEO_ID = 0xE0;
@@ -509,6 +510,10 @@ static DecoderErrorOr<NonnullRefPtr<TrackEntry>> parse_track_entry(Streamer& str
         case TRACK_CODEC_DELAY_ID:
             track_entry->set_codec_delay(TRY_READ(streamer.read_u64()));
             dbgln_if(MATROSKA_TRACE_DEBUG, "Read Track's CodecDelay attribute: {}", track_entry->codec_delay());
+            break;
+        case TRACK_SEEK_PRE_ROLL_ID:
+            track_entry->set_seek_pre_roll(TRY_READ(streamer.read_u64()));
+            dbgln_if(MATROSKA_TRACE_DEBUG, "Read Track's SeekPreRoll attribute: {}", track_entry->seek_pre_roll());
             break;
         case TRACK_TIMESTAMP_SCALE_ID:
             track_entry->set_timestamp_scale(TRY_READ(streamer.read_float()));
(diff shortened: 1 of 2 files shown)


Libraries/LibMedia/Containers/Matroska/Reader.cpp
```diff
@@ -1012,13 +1012,13 @@ DecoderErrorOr<void> SampleIterator::seek_to_cue_point(CuePoint const& cue_point
...
-ErrorOr<ByteString> Streamer::read_string()
+ErrorOr<String> Streamer::read_string()
 {
     auto string_length = TRY(read_variable_size_integer());
     if (remaining() < string_length)
         return Error::from_string_literal("String length extends past the end of the stream");
-    auto string_data = data_as_chars();
-    auto string_value = ByteString(string_data, strnlen(string_data, string_length));
+    auto const* string_data = data_as_chars();
+    auto string_value = String::from_utf8_without_validation(ReadonlyBytes(string_data, strnlen(string_data, string_length)));
```
gmta: Shouldn't we do some validation here though? This seems like untrusted input.

Libraries/LibMedia/Containers/Matroska/Reader.cpp
```diff
@@ -546,9 +547,92 @@ DecoderErrorOr<void> Reader::parse_tracks(Streamer& streamer)
...
+        }();
+        if (!is_affected_version)
+            return;
+
+        u64 max_codec_delay = 0;
+        for (auto& [id, track] : m_tracks) {
+            auto delay = track->codec_delay();
+
+            if (codec_id_from_matroska_id_string(track->codec_id()) == CodecID::Opus && track->audio_track().has_value())
+                delay = delay * 48'000 / AK::clamp_to<u64>(track->audio_track()->sampling_frequency);
```
gmta: Unless I'm reading it incorrectly, `sampling_frequency` can still be a negative double at this point and we'll end up with a division by zero. Maybe we should detect this when reading the frequency from the data?

Zaggy1024: For now I've just made it bail out if the sampling frequency is zero here. Not sure what validation we may actually want for the sampling frequency, but this is the only place where it's used currently.

PR #6687, opened 2025-11-03
LibWeb/WebAudio: Implement control message queue

Prospero23:
Revives #5535 and splits it into two commits with fixes for changes requested by @gmta.

Message of original PR:
## Summary
Implements the WebAudio control message queue as specified in the Web Audio API
## Changes
-Added ControlMessageQueue class with thread-safe operations
-Integrated queue into BaseAudioContext
-Updated AudioContext methods (suspend/resume/close) to use the queue
-Updated AudioScheduledSourceNode to queue start/stop messages
## Testing
- Added custom integration test for the control message queue
## UPDATE
simplified queue from the original PR and switched testing to a c++ test.

### DIFF
--- a/Libraries/LibWeb/WebAudio/BaseAudioContext.h
+++ b/Libraries/LibWeb/WebAudio/BaseAudioContext.h
@@ -8,6 +8,7 @@
 
 #pragma once
 
+#include <AK/Function.h>
 #include <LibWeb/Bindings/BaseAudioContextPrototype.h>
 #include <LibWeb/DOM/EventTarget.h>
 #include <LibWeb/WebAudio/AnalyserNode.h>
@@ -16,6 +17,7 @@
 #include <LibWeb/WebAudio/ChannelMergerNode.h>
 #include <LibWeb/WebAudio/ChannelSplitterNode.h>
 #include <LibWeb/WebAudio/ConstantSourceNode.h>
+#include <LibWeb/WebAudio/ControlMessage.h>
 #include <LibWeb/WebAudio/DelayNode.h>
 #include <LibWeb/WebAudio/PeriodicWave.h>
 #include <LibWeb/WebAudio/ScriptProcessorNode.h>
@@ -25,6 +27,7 @@
 namespace Web::WebAudio {
 
 class AudioDestinationNode;
+class ControlMessageQueue;
 
 // https://webaudio.github.io/web-audio-api/#BaseAudioContext
 class BaseAudioContext : public DOM::EventTarget {
@@ -84,6 +87,8 @@ public:
 
     GC::Ref<WebIDL::Promise> decode_audio_data(GC::Root<WebIDL::BufferSource>, GC::Ptr<WebIDL::CallbackType>, GC::Ptr<WebIDL::CallbackType>);
 
+    void queue_control_message(ControlMessage);
+
 protected:
     explicit BaseAudioContext(JS::Realm&, float m_sample_rate = 0);
 
@@ -110,6 +115,8 @@ private:
     Bindings::AudioContextState m_rendering_thread_state = Bindings::AudioContextState::Suspended;
 
     HTML::UniqueTaskSource m_media_element_event_task_source {};
+
+    NonnullOwnPtr<ControlMessageQueue> m_control_message_queue;
 };
 
 }
--- a/Libraries/LibWeb/WebAudio/ControlMessageQueue.cpp
+++ b/Libraries/LibWeb/WebAudio/ControlMessageQueue.cpp
@@ -1,0 +1,22 @@
+/*
+ * Copyright (c) 2025, Ben Eidson <b.e.eidson@gmail.com>
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include <LibWeb/WebAudio/ControlMessageQueue.h>
+namespace Web::WebAudio {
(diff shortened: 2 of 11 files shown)


Prospero23:
I am not sure if `WEB_PLATFORM_OBJECT` and `GC_DECLARE_ALLOCATOR` should be on the queue but they were on the original PR so I left them. 

Should something that is not declared to JS still be a `WEB_PLATFORM_OBJECT`? Almost certainly just confusion caused by lack of understanding on my part.

Libraries/LibWeb/WebAudio/BaseAudioContext.h
```diff
@@ -84,6 +86,8 @@ class BaseAudioContext : public DOM::EventTarget {
 
     GC::Ref<WebIDL::Promise> decode_audio_data(GC::Root<WebIDL::BufferSource>, GC::Ptr<WebIDL::CallbackType>, GC::Ptr<WebIDL::CallbackType>);
 
+    void queue_control_message(GC::Ref<GC::Function<void()>>) const;
```
gmta: Why is this method `const`? Removing it also allows you to get rid of the `mutable` on `m_control_message_queue`.

Libraries/LibWeb/WebAudio/ControlMessageQueue.cpp
```diff
@@ -0,0 +1,85 @@
...
+    // FIXME: Signal the rendering thread when implemented
+}
+
+bool ControlMessageQueue::has_messages() const
+{
+    Threading::MutexLocker locker(m_mutex);
+    return !m_messages.is_empty();
+}
+
+void ControlMessageQueue::process_messages()
```
gmta: We're not calling this anywhere, are we?

Libraries/LibWeb/WebAudio/ControlMessageQueue.cpp
```diff
@@ -0,0 +1,85 @@
...
+            messages_to_process = move(m_messages);
+        }
+
+        // FIXME: Should be called from the rendering thread
+
+        for (auto& message : messages_to_process) {
+            message->function()();
+        }
+    }
+}
```
gmta: It's stated in the spec that the control thread employs a "traditional event loop, as described in [HTML]" - we should probably implement that?

Since this isn't called yet, what's your plan on implementing the control thread? Seems like the spec calls for two different concepts; the control thread whose job is to execute the messages on the control message queue, and the control message queue whose job it is to maintain the list of messages.

Libraries/LibWeb/WebAudio/ControlMessageQueue.h
```diff
@@ -0,0 +1,44 @@
...
+#include <LibThreading/Mutex.h>
+#include <LibWeb/Bindings/PlatformObject.h>
+#include <LibWeb/Forward.h>
+
+namespace Web::WebAudio {
+
+// https://webaudio.github.io/web-audio-api/#control-message-queue
+class ControlMessageQueue : public Bindings::PlatformObject {
+    WEB_PLATFORM_OBJECT(ControlMessageQueue, Bindings::PlatformObject);
+    GC_DECLARE_ALLOCATOR(ControlMessageQueue);
```
gmta: `ControlMessageQueue` is not a platform object but rather a helper object that will be used by the control thread to perform calls eventually picked up by the rendering thread. You can still make it GC-allocated though.

Libraries/LibWeb/WebAudio/BaseAudioContext.h
Prospero23: Fixed!

Libraries/LibWeb/WebAudio/ControlMessageQueue.h
Prospero23: Fixed by removing the need for GC-allocation :)

Libraries/LibWeb/WebAudio/ControlMessageQueue.cpp
Prospero23: My understanding is the control thread described in the spec is (or at least can be) the existing main JS/DOM event loop, so we don't need to implement a new thread for that.

What we need to implement is the rendering thread.  That thread has both real-time and non-real-time versions. My plan is to implement non-real-time rendering first as it is what is used in the WPT tests. 

`ControlMessageQueue` is simply used to queue control messages from the main thread so they can be applied by the rendering thread during the render loop in a non-blocking way.

Libraries/LibWeb/WebAudio/ControlMessageQueue.cpp
Prospero23: The API should be clearer now. The control thread enqueues and the rendering thread will drain it.

Prospero23:
Sorry for the delay in fixing everything. Everything should look better now. Happy to fix anything as always :)

failing tests are unrelated

Libraries/LibWeb/WebAudio/ControlMessageQueue.cpp
Prospero23: Open to being wrong in my understanding of the control thread

gmta: No I think you're right! Control thread is our regular event loop; rendering should be a separate thread.

PR #6690, opened 2025-11-04
LibWeb: Propagate style values in deep anonymous wrappers

### LINKED ISSUE #4676: CSS label color change on hover is not applied on button element
### Summary

See min repro below, the color of the text in the button is not changed when hovering. Possibly CSS other properties  have this issue too on hover (originated from looking at the auth buttons at the mottom of https://www.openstreetmap.org/user/new)

**Min repro screenshots**
Hovering in LB:
Image

Hovering in Chrome:
Image

### Operating system

Linux

### Steps to reproduce

See min repro and screenshots

### Expected behavior

Color of button text change

### Actual behavior

No change

NN

```html
<style>
.btn:hover {
	color:  #fff;
}
</style>

<button type="button" class="btn">BUTTON</button>
```

```shell
NN
```

_No response_

LB 123abe0e79354d4cfc005adc973688f6e378365a, Linux

### Contribute a patch?

lpas:
`wrap_in_button_layout_tree_if_needed` creates two nested anonymous wrapper. With this fix styles get propagated two the children.

fixes #4676

### DIFF
--- a/Libraries/LibWeb/Layout/Node.cpp
+++ b/Libraries/LibWeb/Layout/Node.cpp
@@ -922,6 +922,15 @@ void NodeWithStyle::apply_style(CSS::ComputedProperties const& computed_style)
         box_node->propagate_style_along_continuation(computed_style);
 }
 
+void NodeWithStyle::propagate_non_inherit_values(NodeWithStyle& target_node) const
+{
+    // NOTE: These properties are not inherited, but we still have to propagate them to anonymous wrappers.
+    target_node.mutable_computed_values().set_text_decoration_line(computed_values().text_decoration_line());
+    target_node.mutable_computed_values().set_text_decoration_thickness(computed_values().text_decoration_thickness());
+    target_node.mutable_computed_values().set_text_decoration_color(computed_values().text_decoration_color());
+    target_node.mutable_computed_values().set_text_decoration_style(computed_values().text_decoration_style());
+}
+
 void NodeWithStyle::propagate_style_to_anonymous_wrappers()
 {
     // Update the style of any anonymous wrappers that inherit from this node.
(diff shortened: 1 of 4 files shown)


awesomekling (changes requested):
Please wrap commit messages at 72 characters, not 50.

And more importantly, can we add a test? :)

lpas:
> And more importantly, can we add a test? :)

I'm not sure how to Test this. This can only be seen in layout on hover over a button.

awesomekling:
> > And more importantly, can we add a test? :)
> 
> I'm not sure how to Test this. This can only be seen in layout on hover over a button.

We can use the `internals.movePointerTo(double x, double y)` API available when running in test mode :) (See `Internals.idl` for a list of all the extra APIs you can use for testing)

lpas:
Oh I didn't think I could use JS in screenshot/ref tests. That is a very nice test setup!
 
I added a test.

Libraries/LibWeb/Layout/Node.cpp
```diff
@@ -913,6 +913,15 @@ void NodeWithStyle::apply_style(CSS::ComputedProperties const& computed_style)
         box_node->propagate_style_along_continuation(computed_style);
```
gmta, line 922: Commit message typos: "wrapper" -> "wrappers", "get propagated two the children" -> "get propagated to the children"

Your commit message does not really describe what's going on though: non-inherited styles did not get propagated as part of NodeWithStyle::apply_style(), _and_ we did not propagate to nested anonymous wrappers (i.e. descendants, not just children).

lpas: I updated the commit message

gmta (approved):

Issue #6549, opened 2025-10-22
ease-in-out animations don't look right


awesomekling:
This animation looks way too "snappy" in Ladybird compared to other browsers.

```html
<!DOCTYPE html><style>
  .dot {
    width: 100px;
    height: 100px;
    border-radius: 100%;
    background: orange;
    animation: bounce 2s ease-in-out infinite;
  }

  @keyframes bounce {
    0%, 100%   { transform: translate(0, 0);      }   
    50%        { transform: translate(50px, 50px);  }
  }

</style><body><div class="dot"></div>
```

CreatorsHiring:
can i take this issue? show a video and where exactly it is happening

gmta:
> can i take this issue? show a video and where exactly it is happening

You should take the reproduction HTML and run in it Ladybird and other browsers yourself, to see the difference.

Oreo-Stars:
Nice

CreatorsHiring:
im new to open source and i don't understand where to make the code changes like there are too many files.

calaveraDeluxe:
I have looked into this, and my current understanding is that Ladybird does not apply easing correctly when there are more than two keyframes. In those cases, the timing function should be applied for each keyframe (you can even specify different timing functions for each keyframe, see https://www.w3.org/TR/web-animations-1/#keyframes-section, §4). Ladybird seems to instead apply a single timing function over the whole animation. This leads to the behaviour above, where the circle, after starting "eased", keeps the same speed, only slowing sown at the very end. The correct behaviour would be 2 easings, one for each keyframe.

The source for this is around https://github.com/LadybirdBrowser/ladybird/blob/master/Libraries/LibWeb/CSS/StyleComputer.cpp#L867 in StyleComputer::collect_animation_into. Here, `effect->transformed_progress()` returns the overall and eased progress of the animation, which is used further on the get the current keyframe.

I haven't yet found the spec for how this should be implemented, but my guess would be:
- find out how far along the animation we are without any easing
- find the corresponding keyframe
- find out how far along this keyframe we are
- ease this value with the easing function of the keyframe
- use that value to interpolate the CSS values

I have a (maybe?) slightly simpler reproduction. This sets the animation to paused and uses an easing that always returns 1. In Firefox and Chromium, this shows a green dot, because the animation is on 0%, which gets interpolated to 1 by the first keyframe. In Ladybird, this shows a red dot, because the interpolation happens over the entire animation range.

```html
[...]

PR #6691, opened 2025-11-04
LibWeb/CSS: Apply easing function to segments of keyframes

JeremiasRy:
Fixes [6549](https://github.com/LadybirdBrowser/ladybird/issues/6549).

Ladybird is applying easing functions for the whole duration of an animation, while other browser apply it to segments of it.

### DIFF
--- a/Libraries/LibWeb/Animations/KeyframeEffect.cpp
+++ b/Libraries/LibWeb/Animations/KeyframeEffect.cpp
@@ -942,4 +942,62 @@ Bindings::CompositeOperation css_animation_composition_to_bindings_composite_ope
     }
 }
 
+Optional<double> KeyframeEffect::transformed_progress() const
+{
+    // 1. If the directed progress is unresolved, return unresolved.
+    auto directed_progress = this->directed_progress();
+    if (!directed_progress.has_value())
+        return {};
+
+    // 2. Calculate the value of the before flag as follows:
+
+    //    1. Determine the current direction using the procedure defined in §4.9.1 Calculating the directed progress.
+    auto current_direction = this->current_direction();
+
+    //    2. If the current direction is forwards, let going forwards be true, otherwise it is false.
+    auto going_forwards = current_direction == AnimationDirection::Forwards;
+
+    //    3. The before flag is set if the animation effect is in the before phase and going forwards is true; or if the animation effect
+    //       is in the after phase and going forwards is false.
+    auto before_flag = (is_in_the_before_phase() && going_forwards) || (is_in_the_after_phase() && !going_forwards);
+
+    if (key_frame_set()) {
+        auto& keyframes = key_frame_set()->keyframes_by_key;
+
+        if (keyframes.size() >= 3) {
+            double scale_factor = 100 * Animations::KeyframeEffect::AnimationKeyFrameKeyScaleFactor;
+            double progress = directed_progress.value() * scale_factor;
+            double segment_bottom = 0;
+            double segment_top = 0;
+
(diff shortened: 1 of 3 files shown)


Libraries/LibWeb/Animations/KeyframeEffect.cpp
```diff
@@ -942,4 +942,62 @@ Bindings::CompositeOperation css_animation_composition_to_bindings_composite_ope
     }
 }
 
+Optional<double> KeyframeEffect::transformed_progress() const
```
Psychpsyo, line 945: This is missing its spec link: https://www.w3.org/TR/web-animations-1/#transformed-progress

Libraries/LibWeb/Animations/KeyframeEffect.cpp
```diff
@@ -942,4 +942,62 @@ Bindings::CompositeOperation css_animation_composition_to_bindings_composite_ope
...
+    auto current_direction = this->current_direction();
+
+    //    2. If the current direction is forwards, let going forwards be true, otherwise it is false.
+    auto going_forwards = current_direction == AnimationDirection::Forwards;
+
+    //    3. The before flag is set if the animation effect is in the before phase and going forwards is true; or if the animation effect
+    //       is in the after phase and going forwards is false.
+    auto before_flag = (is_in_the_before_phase() && going_forwards) || (is_in_the_after_phase() && !going_forwards);
+
+    if (key_frame_set()) {
```
Psychpsyo, line 964: The spec doesn't say to do any of this here. Is that an issue with the spec or is this an ad-hoc fix for a problem somewhere else?
If it is a spec bug: File an issue on the spec itself and put an `// AD-HOC: ` comment here that links to that.
If it isn't, then this fix should probably be somewhere else.

JeremiasRy: This is just a fix to the issue mentioned in the description. I can see now how this is a wrong approach to solve this.

I wonder if [this](https://www.w3.org/TR/web-animations-1/#keyframes-section) is the spec we are looking for?

> Each keyframe also has a [timing function](https://www.w3.org/TR/css-easing-1/#easing-function) associated with it that is applied to the period of time between the keyframe on which it is specified and the next keyframe in the list. The timing function specified on the last keyframe in the list is never applied.

calaveraDeluxe: I think this is the right part of the spec: https://www.w3.org/TR/web-animations-1/#the-effect-value-of-a-keyframe-animation-effect.

JeremiasRy: > I think this is the right part of the spec: https://www.w3.org/TR/web-animations-1/#the-effect-value-of-a-keyframe-animation-effect.

Thanks!

Reading this and trying to understand it, is that the `transform_progress` only applies to the start and end i.e. the "full animation", after this we apply the interval easing for key frames (if they are present).

So looks to me like this [FIXME](https://github.com/LadybirdBrowser/ladybird/blob/9ff75f442b52f55a55d16231c2657a4d822babe5/Libraries/LibWeb/CSS/StyleComputer.cpp#L921) is somewhat relevant for this occasion and the fix should land somewhere after that computation.

I'll start looking into it, feel free to reach out to me on discord to discuss this. 

Thanks for the guidance!

JeremiasRy:
There's not much value in fixing this problem like this, a proper implementation of [5.3.4. The effect value of a keyframe effect](https://www.w3.org/TR/web-animations-1/#the-effect-value-of-a-keyframe-animation-effect) is required.

PR #6695, opened 2025-11-04
LibWeb: Support percentage values in min/max flex item size constraints

gmta:
Originally, https://github.com/LadybirdBrowser/ladybird/commit/7200b3a16c55221eda5da5997a747b4e5536fa6f introduced a two-pass system to determine hypothetical cross sizes. Later, this was partially reverted in https://github.com/LadybirdBrowser/ladybird/commit/0084d992d45695f485cb984d77f3b8a7f1ab8797, but some code was left behind that caused resolution of percentages in `{min/max}-{width/height}` size constraints not to work.

Through intrinsic sizing, we can potentially end up with a definite available space for the items in the last FC run. At that point we should be able to resolve percentages against the available space, but we were never doing that.

Fixes the size of the login box of Grafana:

| Before | After |
|--|--|
| image | image |

### DIFF
--- a/Libraries/LibWeb/Layout/FlexFormattingContext.cpp
+++ b/Libraries/LibWeb/Layout/FlexFormattingContext.cpp
@@ -155,9 +155,8 @@ void FlexFormattingContext::run(AvailableSpace const& available_space)
 
     // Cross Size Determination
     // 7. Determine the hypothetical cross size of each item
-    for (auto& item : m_flex_items) {
-        determine_hypothetical_cross_size_of_item(item, false);
-    }
+    for (auto& item : m_flex_items)
+        determine_hypothetical_cross_size_of_item(item);
 
     // 8. Calculate the cross size of each flex line.
     calculate_cross_size_of_each_flex_line();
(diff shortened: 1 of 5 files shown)


awesomekling:
Typo in commit message, definitive -> definite
Otherwise looks fine :)

awesomekling (approved):

Libraries/LibWeb/Layout/FlexFormattingContext.cpp
```diff
@@ -22,9 +23,9 @@ CSSPixels FlexFormattingContext::get_pixel_width(FlexItem const& item, CSS::Size
 
 CSSPixels FlexFormattingContext::get_pixel_height(FlexItem const& item, CSS::Size const& size) const
 {
-    if (is_row_layout()) {
-        // NOTE: In a row layout, after we've determined the main size, we use that as the available width
-        //       for any intrinsic sizing layout needed to resolve the height.
+    if (is_row_layout() && (size.is_min_content() || size.is_max_content())) {
```
awesomekling: Hmm, should this check `is_fit_content()` as well?

gmta: Definitely, added a new ::is_intrinsic_sizing_constraint() helper for that.

PR #6699, opened 2025-11-04
Render SVG images on the GPU if possible

awesomekling:
See individual commits.

Note that this only implements the macOS parts needed. Vulkan is left as a FIXME.

cc @kalenikaliaksandr 

Pretty decent performance improvement:

https://github.com/user-attachments/assets/fd6a85e8-6c99-4e63-b1af-25db6b702aef

### DIFF
--- a/Libraries/LibWeb/Painting/ImagePaintable.cpp
+++ b/Libraries/LibWeb/Painting/ImagePaintable.cpp
@@ -73,10 +73,10 @@ void ImagePaintable::paint(DisplayListRecordingContext& context, PaintPhase phas
                 context.display_list_recorder().draw_rect(enclosing_rect, Gfx::Color::Black);
                 context.display_list_recorder().draw_text(enclosing_rect, Utf16String::from_utf8(m_alt_text), *Platform::FontPlugin::the().default_font(12), Gfx::TextAlignment::Center, computed_values().color());
             }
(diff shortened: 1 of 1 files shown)


Libraries/LibWeb/SVG/SVGDecodedImageData.cpp
```diff
@@ -120,6 +119,48 @@ RefPtr<Gfx::Bitmap> SVGDecodedImageData::render(Gfx::IntSize size) const
...
+        surface = Gfx::PaintingSurface::create_from_iosurface(move(iosurface), *m_document->navigable()->skia_backend_context(), Gfx::PaintingSurface::Origin::TopLeft);
+    }
+#endif
+
+    // FIXME: Support Vulkan
+
+    if (!surface) {
+        auto bitmap = Gfx::Bitmap::create(Gfx::BitmapFormat::BGRA8888, Gfx::AlphaType::Premultiplied, size).release_value();
+        surface = Gfx::PaintingSurface::wrap_bitmap(*bitmap);
+    }
```
kalenikaliaksandr: since IOSurface is not shared either between processes or GPU contexts, we could simply use `PaintingSurface::create_with_size()` which will allow GPU acceleration for both Vulkan and Metal.

awesomekling: Oh hell yeah, I didn't realize it was that easy :)

PR #6702, opened 2025-11-04
LibWeb: Cache font matching algorithm results

tcl3:
The majority of time in `compute_font()` was spent in `font_matching_algorithm()` repeatedly computing the same values. We now cache these values to avoid unnecessary work.

This showed up when profiling: https://en.wikipedia.org/wiki/2023_in_American_television. 

Total time from `time ./Meta/ladybird.py run ladybird en.wikipedia.org/wiki/2023_in_American_television --headless=text`

Before: 4.068
After: 3.306

### DIFF
--- a/Libraries/LibWeb/CSS/StyleComputer.cpp
+++ b/Libraries/LibWeb/CSS/StyleComputer.cpp
@@ -131,6 +131,18 @@ struct Traits<Web::CSS::OwnFontFaceKey> : public DefaultTraits<Web::CSS::OwnFont
     static unsigned hash(Web::CSS::OwnFontFaceKey const& key) { return pair_int_hash(key.family_name.hash(), pair_int_hash(key.weight, key.slope)); }
 };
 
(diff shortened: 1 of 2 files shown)


Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1694,9 +1706,17 @@ RefPtr<Gfx::FontCascadeList const> StyleComputer::find_matching_font_weight_desc
     return {};
 }
 
+RefPtr<Gfx::FontCascadeList const> StyleComputer::font_matching_algorithm(FlyString const& family_name, int weight, int slope, float font_size_in_pt) const
+{
+    FontMatchingAlgorithmCacheKey key { move(family_name), weight, slope, font_size_in_pt };
```
kalenikaliaksandr: `move()` could be removed because it doesn't do anything in this case.

PR #6705, opened 2025-11-05
LibGfx: Use Skia to speed up alpha conversions on Linux

tcl3:
When profiling bbc.co.uk, this reduces the time spent in `Bitmap::set_alpha_destructive()` from 4.13% to 3.65%. This is not as much of a speed-up as I hoped for, but I thought I'd PR it anyway, since it is definitely faster. 

The SkPixmap methods still use the CPU, ideally we should find a way to do this conversion on the GPU if possible.

EDIT: I was able to improve performance significantly by not attempting to alter opaque images.

### DIFF
--- a/Libraries/LibGfx/Color.h
+++ b/Libraries/LibGfx/Color.h
@@ -316,28 +316,6 @@ public:
         return color_with_alpha;
     }
 
-    constexpr Color to_premultiplied() const
-    {
-        u32 a = alpha();
-        u8 r = static_cast<u8>((red() * a + 127) / 255);
-        u8 g = static_cast<u8>((green() * a + 127) / 255);
-        u8 b = static_cast<u8>((blue() * a + 127) / 255);
-        return Color(r, g, b, a);
-    }
-
-    constexpr Color to_unpremultiplied() const
-    {
-        u32 a = alpha();
-        if (a == 0)
-            return Color(0, 0, 0, 0);
-        if (a == 255)
-            return *this;
-        u8 r = static_cast<u8>(min(255u, (red() * 255u + a / 2) / a));
-        u8 g = static_cast<u8>(min(255u, (green() * 255u + a / 2) / a));
-        u8 b = static_cast<u8>(min(255u, (blue() * 255u + a / 2) / a));
-        return Color(r, g, b, a);
-    }
-
     constexpr Color blend(Color source) const
     {
         if (alpha() == 0 || source.alpha() == 255)


Libraries/LibGfx/Bitmap.cpp
```diff
@@ -353,20 +355,20 @@ void Bitmap::set_alpha_type_destructive(AlphaType alpha_type)
...
-        for (auto y = 0; y < height(); ++y) {
-            for (auto x = 0; x < width(); ++x)
-                set_pixel(x, y, get_pixel(x, y).to_unpremultiplied());
-        }
-    } else {
-        VERIFY_NOT_REACHED();
-    }
+    auto color_type = to_skia_color_type(m_format);
+    auto source_alpha = m_alpha_type == AlphaType::Premultiplied ? kPremul_SkAlphaType : kUnpremul_SkAlphaType;
+    auto destination_alpha = alpha_type == AlphaType::Premultiplied ? kPremul_SkAlphaType : kUnpremul_SkAlphaType;
```
gmta: Looking at `to_skia_alpha_type()`, wouldn't this also necessitate supporting `kOpaque_SkAlphaType`? Or do we never end up here if we're dealing with opaque surfaces?

tcl3: Changing this to use to_skia_alpha_type() instead makes Bitmap::set_alpha_destructive() use 0.03% of profile time.

Libraries/LibGfx/Bitmap.cpp
```diff
@@ -353,20 +355,33 @@ void Bitmap::set_alpha_type_destructive(AlphaType alpha_type)
...
-            for (auto x = 0; x < width(); ++x)
-                set_pixel(x, y, get_pixel(x, y).to_unpremultiplied());
-        }
-    } else {
+    auto color_type = to_skia_color_type(m_format);
+    auto source_alpha = to_skia_alpha_type(m_format, m_alpha_type);
+    BitmapFormat destination_format;
+    switch (m_format) {
+    case BitmapFormat::BGRA8888:
+    case BitmapFormat::BGRx8888:
```
gmta: If the source format does not have an alpha channel, I don't think we can do any alpha type conversion? Shouldn't we return early in that case?

Libraries/LibGfx/Bitmap.cpp
```diff
@@ -353,20 +355,33 @@ void Bitmap::set_alpha_type_destructive(AlphaType alpha_type)
...
+
+    auto color_space = SkColorSpace::MakeSRGB();
+
+    auto source_info = SkImageInfo::Make(width(), height(), color_type, source_alpha, color_space);
+    auto destination_info = SkImageInfo::Make(width(), height(), color_type, destination_alpha, color_space);
+
+    SkPixmap src_pixmap(source_info, m_data, pitch());
+    SkPixmap dst_pixmap(destination_info, m_data, pitch());
+
+    bool ok = src_pixmap.readPixels(dst_pixmap);
```
gmta, line 371: Did you check how this measures up to our custom macOS code for (un)premultiplying alpha? If this is just as fast, we might not need the macOS version.

Libraries/LibGfx/Bitmap.cpp
tcl3: I initially thought that, but then I looked at what the MacOS code was doing and copied how that worked.

I think you're right though, I don't see why we wouldn't just return early.

EDIT: I have now made it so we return early for opaque bitmap formats.

Libraries/LibGfx/Bitmap.cpp
tcl3: I can't see `Bitmap::set_alpha_destructive()` at all in a profile when using the Skia code on MacOS, so we're probably safe to remove it. I'd like to try and get a proper measurement first though.

tcl3: Testing on cloudflare.com with this code:
```c++
static u64 cumulative_microseconds = 0;
auto timer = Core::ElapsedTimer::start_new(Core::TimerType::Precise);
ScopeGuard guard = [&] {
    auto microseconds = timer.elapsed_time().to_microseconds();
    cumulative_microseconds += microseconds;
    dbgln("Time spend in Bitmap::set_alpha_type_destructive: {} microseconds (cumilative: {} microseconds)", microseconds, cumulative_microseconds);
};
```
I'm seeing a cumulative time of ~58000μs with the Skia code vs ~35000μs with the MacOS specific code, so it's probably worth keeping.

Libraries/LibGfx/Bitmap.cpp
```diff
@@ -321,17 +321,18 @@ void Bitmap::set_alpha_type_destructive(AlphaType alpha_type)
     if (alpha_type == m_alpha_type)
         return;
 
+    if (m_format == BitmapFormat::BGRx8888 || m_format == BitmapFormat::RGBx8888)
+        return;
```
gmta, line 328: Maybe this should still set `m_alpha_type` to the requested value, technically we should always pass along the right Skia alpha type based on the format but better to be safe than sorry?

Tangent: maybe "opaque" would be a nice third alpha type to have here, Skia probably has the right idea.

tcl3: > Maybe this should still set m_alpha_type to the requested value, technically we should always pass along the right Skia alpha type based on the format but better to be safe than sorry?

My latest push now sets updates the alpha type before returning

> Tangent: maybe "opaque" would be a nice third alpha type to have here, Skia probably has the right idea.

I agree, if we're going to have `AlphaType::Opaque`, we might want to stop specifying  whether the bitmap has an alpha channel in `BitmapFormat`.

PR #6707, opened 2025-11-05
LibWeb: Implement CSS `random()` function

Calme1709:

I'm not sure if LibCrypto is the correct place for XorShift128PlusRNG so open to suggestions on where that should live.

The imported tests which are not yet passing are either elsewhere in the codebase (e.g. we don't support the max-lines property) or are because the test doesn't align with the spec, see https://github.com/web-platform-tests/wpt/pull/55867

### DIFF
--- a/Libraries/LibWeb/CSS/Keywords.json
+++ b/Libraries/LibWeb/CSS/Keywords.json
@@ -192,6 +192,7 @@
   "ease-in",
   "ease-in-out",
   "ease-out",
+  "element-shared",
   "ellipsis",
   "embed",
   "emoji",
--- a/Libraries/LibWeb/CSS/Parser/Parser.h
+++ b/Libraries/LibWeb/CSS/Parser/Parser.h
@@ -603,6 +603,7 @@ private:
     bool context_allows_quirky_length() const;
     bool context_allows_tree_counting_functions() const;
     bool context_allows_random_functions() const;
+    FlyString random_value_sharing_auto_name() const;
 
     Vector<RuleContext> m_rule_context;
     HashTable<FlyString> m_declared_namespaces;
--- a/Libraries/LibWeb/CSS/StyleValues/CalculatedStyleValue.cpp
+++ b/Libraries/LibWeb/CSS/StyleValues/CalculatedStyleValue.cpp
@@ -2595,7 +2595,9 @@ String RandomCalculationNode::to_string(CalculationContext const& context, Seria
     StringBuilder builder;
 
     builder.append("random("sv);
-    builder.appendff("{}, ", m_random_value_sharing->to_string(serialization_mode));
+    auto random_value_sharing_stringified = m_random_value_sharing->to_string(serialization_mode);
+    if (!random_value_sharing_stringified.is_empty())
+        builder.appendff("{}, ", random_value_sharing_stringified);
     builder.appendff("{}, ", serialize_a_calculation_tree(m_minimum, context, serialization_mode));
     builder.append(serialize_a_calculation_tree(m_maximum, context, serialization_mode));
     if (m_step)
--- a/Libraries/LibWeb/CSS/StyleValues/RandomValueSharingStyleValue.cpp
+++ b/Libraries/LibWeb/CSS/StyleValues/RandomValueSharingStyleValue.cpp
@@ -7,11 +7,15 @@
 #include "RandomValueSharingStyleValue.h"
 #include <LibWeb/CSS/StyleValues/CalculatedStyleValue.h>
 #include <LibWeb/CSS/StyleValues/NumberStyleValue.h>
+#include <LibWeb/DOM/Document.h>
 
 namespace Web::CSS {
 
 ValueComparingNonnullRefPtr<StyleValue const> RandomValueSharingStyleValue::absolutized(ComputationContext const& computation_context) const
 {
+    // https://drafts.csswg.org/css-values-5/#random-caching
+    // Each instance of a random function in styles has an associated random base value.
+    // If the random function’s <random-value-sharing> is fixed <number>, the random base value is that number.
     if (m_fixed_value) {
         auto const& absolutized_fixed_value = m_fixed_value->absolutized(computation_context);
 
@@ -21,7 +25,29 @@ ValueComparingNonnullRefPtr<StyleValue const> RandomValueSharingStyleValue::abso
         return RandomValueSharingStyleValue::create_fixed(absolutized_fixed_value);
     }
 
-    TODO();
+    // Otherwise, the random base value is a pseudo-random real number in the range `[0, 1)` (greater than or equal to 0
+    // and less than 1), generated from a uniform distribution, and influenced by the function’s random caching key.
+
+    // A random caching key is a tuple of:
+    RandomCachingKey random_caching_key {
+        // 1. A string name: the value of the <dashed-ident>, if specified in <random-value-sharing>; or else a string
+        //    of the form "PROPERTY N", where PROPERTY is the name of the property the random function is used in
+        //    (before shorthand expansion, if relevant), and N is the index of the random function among other random
+        //    functions in the same property value.
+        .name = m_name.value(),
+
+        // 2. An element ID identifying the element the style is being applied to, or null if element-shared is
+        //    specified in <random-value-sharing>.
+        // FIXME: Use the pseudo element's unique_id() when that's accessible
+        .element_id = m_element_shared ? Optional<UniqueNodeID> { OptionalNone {} } : Optional<UniqueNodeID> { computation_context.abstract_element->element().unique_id() },
+
+        // 3. A document ID identifying the Document the styles are from.
+        // NB: This is implicit since the cache is stored on the document or the element (which is a child of the document).
+    };
+
+    auto random_base_value = const_cast<DOM::Element&>(computation_context.abstract_element->element()).ensure_css_random_base_value(random_caching_key);
+
+    return RandomValueSharingStyleValue::create_fixed(NumberStyleValue::create(random_base_value));
 }
 
 double RandomValueSharingStyleValue::random_base_value() const
@@ -43,7 +69,18 @@ String RandomValueSharingStyleValue::to_string(SerializationMode serialization_m
     if (m_fixed_value)
         return MUST(String::formatted("fixed {}", m_fixed_value->to_string(serialization_mode)));
 
-    TODO();
+    StringBuilder builder;
+
+    if (!m_is_auto)
+        builder.appendff("{}", m_name.value());
+
+    if (m_element_shared) {
+        if (!builder.is_empty())
+            builder.append(' ');
+        builder.append("element-shared"sv);
+    }
+
+    return builder.to_string_without_validation();
 }
 
 }
--- a/Libraries/LibWeb/DOM/Document.h
+++ b/Libraries/LibWeb/DOM/Document.h
@@ -267,6 +267,8 @@ public:
 
     CSS::StyleSheetList* style_sheets_for_bindings() { return &style_sheets(); }
 
+    double ensure_element_shared_css_random_base_value(CSS::RandomCachingKey const&);
+
     Optional<String> get_style_sheet_source(CSS::StyleSheetIdentifier const&) const;
 
     virtual FlyString node_name() const override { return "#document"_fly_string; }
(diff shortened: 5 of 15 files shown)


Calme1709:
Changes:
 - Rebased to resolve conflicts
 - Updated imported WPT tests to latest version after web-platform-tests/wpt#55867 was merged

Libraries/LibCrypto/SecureRandom.cpp
```diff
@@ -18,4 +18,41 @@ void fill_with_secure_random(Bytes bytes)
...
+    u64 s1 = m_low;
+    u64 const s0 = m_high;
+    u64 const result = s0 + s1;
+    m_low = s0;
+    s1 ^= s1 << 23;
+    s1 ^= s1 >> 18;
+    s1 ^= s0 ^ (s0 >> 5);
+    m_high = s1;
+    return result + s1;
+}
```
gmta: Pending #6564, xorshift128+ is a PRNG that could also live in AK - its usage of `Crypto::get_secure_random()` prohibits that at the moment, but I don't see why we couldn't switch to AK's `get_random()` since it'll be a secure source of random bytes anyway.

Libraries/LibWeb/CSS/StyleValues/CalculatedStyleValue.cpp
```diff
@@ -2434,6 +2442,133 @@ bool ModCalculationNode::equals(CalculationNode const& other) const
...
+    auto minimum = try_get_value_with_canonical_unit(m_minimum, context, resolution_context);
+    auto maximum = try_get_value_with_canonical_unit(m_maximum, context, resolution_context);
+
+    if (!minimum.has_value() || !maximum.has_value())
+        return {};
+
+    auto minimum_value = minimum->value();
+    auto maximum_value = maximum->value();
+
+    // If the maximum value is less than the minimum value, it behaves as if it’s equal to the minimum value.
```
gmta, line 2531: Please add the right spec link for these, i.e. https://drafts.csswg.org/css-values-5/#random-infinities

Libraries/LibWeb/CSS/StyleValues/CalculatedStyleValue.cpp
```diff
@@ -2434,6 +2442,133 @@ bool ModCalculationNode::equals(CalculationNode const& other) const
...
+    // In random(A, B), if A is infinite, the result is infinite.
+    if (isinf(minimum_value))
+        return CalculatedStyleValue::CalculationResult { AK::Infinity<double>, numeric_type() };
+
+    // If A is finite, but the difference between A and B is either infinite or large enough to be treated as infinite
+    // in the user agent, the result is NaN.
+    if (isinf(maximum_value))
+        return CalculatedStyleValue::CalculationResult { AK::NaN<double>, numeric_type() };
+
+    // Spec Note: As usual for math functions, if any argument calculation is NaN, the result is NaN.
```
gmta: Please copy the text verbatim, i.e. `// NOTE: ..`

Libraries/LibWeb/CSS/StyleValues/CalculatedStyleValue.cpp
```diff
@@ -2434,6 +2442,133 @@ bool ModCalculationNode::equals(CalculationNode const& other) const
...
+    if (simplified_random_value_sharing == m_random_value_sharing && simplified_minimum == m_minimum && simplified_maximum == m_maximum)
+        return *this;
+
+    return RandomCalculationNode::create(simplified_random_value_sharing.release_nonnull(), simplified_minimum, simplified_maximum);
+}
+
+// https://drafts.csswg.org/css-values-5/#random-evaluation
+Optional<CalculatedStyleValue::CalculationResult> RandomCalculationNode::run_operation_if_possible(CalculationContext const& context, CalculationResolutionContext const& resolution_context) const
+{
+    // NOTE: We don't want to resolve this before computation time even if it's possible
```
gmta: For our own notes inside spec implementations, we use `// NB: ..` now

Libraries/LibWeb/DOM/Document.h
```diff
@@ -1346,6 +1352,9 @@ class WEB_API Document
     HashMap<FlyString, GC::Ref<Web::CSS::CSSPropertyRule>> m_registered_custom_properties;
 
     CSS::StyleScope m_style_scope;
+
+    // https://drafts.csswg.org/css-values-5/#random-caching
+    HashMap<RandomCachingKey, double> m_css_random_base_value_cache;
```
gmta: How do we ensure this doesn't grow unbounded, e.g. with many appearing and disappearing elements in the document? Can we clear a key when its associated element is finalized?

Libraries/LibCrypto/SecureRandom.cpp
colleirose: XorShift isn't intended to be a secure RNG either way, so it would be fine to just use `AK::get_random()` now even while #6564 is pending because `AK::get_random()` not currently being guaranteed to use a CSPRNG doesn't really matter anyway. The only problem here would be the possibility of a merge conflict depending on where it's placed, but that should be simple enough to resolve for whichever PR gets merged last.

If there is a desire to specifically make an insecure but faster RNG (this could be relevant to things like `Math.random()` where throughput isn't as important but very low latency could be beneficial for some uses), it'd make sense to have a function for this but I'm unsure as to how seeding would work to make it faster than using the operating system CSPRNG.

Calme1709: I have moved `XorShift128PlusRNG` to AK/Random.{h,cpp} and updated it to use `AK::get_random()`

Libraries/LibWeb/DOM/Document.h
Calme1709: I have split this into two `HashMap`s, one for the `element-shared` values which lives on the document and one for the non-`element-shared` values which lives on the specific element (and will be removed when the element is)

Libraries/LibWeb/CSS/Parser/Parser.h
```diff
@@ -533,7 +533,7 @@ class Parser {
     RefPtr<StyleValue const> parse_list_of_time_values(PropertyID, TokenStream<ComponentValue>&);
 
     RefPtr<CalculationNode const> convert_to_calculation_node(CalcParsing::Node const&, CalculationContext const&);
-    RefPtr<CalculationNode const> parse_a_calculation(Vector<ComponentValue> const&, CalculationContext const&);
+    RefPtr<CalculationNode const> parse_a_calculation(TokenStream<ComponentValue>&, CalculationContext const&);
```
AtkinsSJ, line 543: Thanks! I've been meaning to do this for ages but somehow never got around to it.

Libraries/LibWeb/CSS/StyleValues/CalculatedStyleValue.cpp
```diff
@@ -284,6 +285,11 @@ static String serialize_a_math_function(CalculationNode const& fn, CalculationCo
         }
     }
 
+    // AD-HOC: We serialize random() directly since it has abnormal children (e.g. m_random_value_sharing which is not a
+    //         calculation node).
+    if (fn.type() == CalculationNode::Type::Random)
+        return as<RandomCalculationNode>(fn).to_string(context, serialization_mode);
+
```
AtkinsSJ, line 292: :thonk: This is a little unfortunate. I think maybe this sort of thing is supposed to be handled by step 1 of "serialize a calculation tree", which would be called on the random-value-sharing parameter by step 4 below. There's some ambiguity though, and it's absolutely not possible with the way this is currently all written.

Specifically, it depends if step 2 there:
> If root is a numeric value, or a non-[math function](https://drafts.csswg.org/css-values-4/#math-function), serialize root per the normal rules for it and return the result.

... means a *function* that isn't a math function, or just a *value* that isn't a math function. I assumed the former when I wrote that, but now I'm not sure. If it's the latter, maybe we could have a CalculationNode that's just a wrapper for an arbitrary StyleValue and use that.

(None of this is a criticism of the code you've written here btw! It's fine, and doesn't need changing right now. I was just surprised it's necessary.)

Libraries/LibWeb/CSS/StyleValues/CalculatedStyleValue.cpp
```diff
@@ -2434,6 +2442,134 @@ bool ModCalculationNode::equals(CalculationNode const& other) const
...
+        simplified_random_value_sharing = m_random_value_sharing;
+    }
+
+    ValueComparingNonnullRefPtr<CalculationNode const> simplified_minimum = simplify_a_calculation_tree(m_minimum, context, resolution_context);
+    ValueComparingNonnullRefPtr<CalculationNode const> simplified_maximum = simplify_a_calculation_tree(m_maximum, context, resolution_context);
+
+    if (simplified_random_value_sharing == m_random_value_sharing && simplified_minimum == m_minimum && simplified_maximum == m_maximum)
+        return *this;
+
+    return RandomCalculationNode::create(simplified_random_value_sharing.release_nonnull(), simplified_minimum, simplified_maximum);
```
AtkinsSJ: Nit: `move()` nonnullrefptrs when you're done with them, like here. It saves a little bit of ref-counted bookkeeping.

Libraries/LibWeb/CSS/StyleValues/RandomValueSharingStyleValue.cpp
```diff
@@ -0,0 +1,49 @@
...
+    if (m_fixed_value) {
+        auto const& absolutized_fixed_value = m_fixed_value->absolutized(computation_context);
+
+        if (m_fixed_value == absolutized_fixed_value)
+            return *this;
+
+        return RandomValueSharingStyleValue::create_fixed(absolutized_fixed_value);
+    }
+
+    VERIFY_NOT_REACHED();
```
AtkinsSJ: `TODO()` might be more appropriate for these "this isn't implemented yet" verifies.

Libraries/LibWeb/CSS/Keywords.json
```diff
@@ -191,6 +191,7 @@
   "ease-in",
   "ease-in-out",
   "ease-out",
+  "element-shared",
```
AtkinsSJ, line 195: Commit message typo: "the them"

Libraries/LibWeb/DOM/Element.cpp
```diff
@@ -4311,6 +4312,19 @@ GC::Ref<CSS::StylePropertyMapReadOnly> Element::computed_style_map()
     return *m_computed_style_map_cache;
 }
 
+double Element::ensure_css_random_base_value(CSS::RandomCachingKey const& random_caching_key)
+{
+    // NB: We cache element-shared random base values on the Document and non-element-shared ones on the Element itself
+    //     so that when an element is removed it takes it's non-shared cache with it.
```
AtkinsSJ: Super-nit: Should be "its", because English is silly.

Libraries/LibWeb/CSS/StyleValues/CalculatedStyleValue.cpp
Calme1709: Given that numeric value is listed explicitly when it would be included in the latter I assume the intent here was the latter, although that's largely irrelevant here as random is [explicitly stated to be a math function](https://drafts.csswg.org/css-values-5/#random):
> The random() function is a math function that represents a random value...

> Maybe we could have a CalculationNode that's just a wrapper for an arbitrary StyleValue and use that

We already do in the form of `NonMathFunctionCalculationNode`, although again it's not applicable here as `random()` is a math function, so we should presumably have that implemented as its own `CalculationNode` class.

AtkinsSJ: I explained that poorly I think.

I think the spec is saying, roughly:
- "Serialize a math function" step 3: Print out "random(".
- ...and step 4: Iterate the random()'s parameters and call "serialize a calculation tree" for each.
- Step 2 of "serialize a calculation tree" when root is the random-value-sharing parameter: it's not a math function or numeric value so serialize it directly.

Calme1709: Ah okay, I'm with you now. I still think that "non-math function" refers to functions other than math functions and this is just an oversight since (AFAIK) this is the only exception.

I agree that it's fine to leave as is for now, especially since the step function being nullable likely means we would need to do something AD-HOC anyway.

Calme1709:
Rebased to resolve conflicts

Libraries/LibWeb/CSS/StyleValues/CalculatedStyleValue.cpp
AtkinsSJ: I just went to raise a spec issue for clarification, and realised I already had one open, but at the time I hadn't even considered that "non-math function" might mean non-functions. I've added a response to it: https://github.com/w3c/csswg-drafts/issues/11783#issuecomment-3595781091

But yeah, marking this resolved again. I'll see what happens there.

PR #6710, opened 2025-11-05
Add lsb-release to Debian/Ubuntu build instructions

dieseltravis:
Category:Documentation
lsb_release isn't always included in Ubuntu, and not installing it will cause the kitware.list line to break apt.

### DIFF
--- a/Documentation/BuildInstructionsLadybird.md
+++ b/Documentation/BuildInstructionsLadybird.md
@@ -15,7 +15,7 @@ CMake 3.25 or newer must be available in $PATH.
 
 <!-- Note: If you change something here, please also change it in the `devcontainer/devcontainer.json` file. -->
 ```bash
-sudo apt install autoconf autoconf-archive automake build-essential ccache cmake curl fonts-liberation2 git libdrm-dev libgl1-mesa-dev libtool nasm ninja-build pkg-config python3-venv qt6-base-dev qt6-tools-dev-tools qt6-wayland tar unzip zip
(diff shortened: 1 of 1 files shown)


Documentation/BuildInstructionsLadybird.md
gmta, line 1: Commit message: linter is failing.

Documentation/BuildInstructionsLadybird.md
```diff
@@ -15,7 +15,7 @@ CMake 3.25 or newer must be available in $PATH.
 
 <!-- Note: If you change something here, please also change it in the `devcontainer/devcontainer.json` file. -->
 ```bash
-sudo apt install autoconf autoconf-archive automake build-essential ccache cmake curl fonts-liberation2 git libdrm-dev libgl1-mesa-dev libtool nasm ninja-build pkg-config python3-venv qt6-base-dev qt6-tools-dev-tools qt6-wayland tar unzip zip
+sudo apt install autoconf autoconf-archive automake build-essential ccache cmake curl fonts-liberation2 git libdrm-dev libgl1-mesa-dev libtool lsb-release nasm ninja-build pkg-config python3-venv qt6-base-dev qt6-tools-dev-tools qt6-wayland tar unzip zip
```
gmta, line 18: Instead of adding lsb-release, we could update the instructions for CMake to not use lsb-release at all. For example, we could use `/etc/os-release` to determine the codename.

PR #6713, opened 2025-11-06
LibWeb: Implement OffscreenCanvasRenderingContext2D and Integrate OffscreenCanvas

me-it-is:


Libraries/LibWeb/HTML/OffscreenCanvasRenderingContext2D.cpp
```diff
@@ -85,81 +122,417 @@ OffscreenCanvas const& OffscreenCanvasRenderingContext2D::canvas_element() const
...
+    auto bottom_left = Gfx::FloatPoint(x, y + height);
+    auto bottom_right = Gfx::FloatPoint(x + width, y + height);
+
+    Gfx::Path path;
+    path.move_to(top_left);
+    path.line_to(top_right);
+    path.line_to(bottom_right);
+    path.line_to(bottom_left);
+    path.line_to(top_left);
+    return path;
```
AtkinsSJ: This is identical to the `CanvasRenderingContext2D` code, as are a lot of other methods in this commit. I think it would make sense to pull those out into some kind of base class (or several), or some other way of avoiding the duplication. For example, these rect-related methods are defined on [`CanvasRect` in the spec](https://html.spec.whatwg.org/multipage/canvas.html#canvasrect). Maybe we should imitate that?

Libraries/LibWeb/HTML/OffscreenCanvasRenderingContext2D.cpp
```diff
@@ -85,81 +122,417 @@ OffscreenCanvas const& OffscreenCanvasRenderingContext2D::canvas_element() const
...
+    case Bindings::CanvasLineJoin::Round:
+        return Gfx::Path::JoinStyle::Round;
+    case Bindings::CanvasLineJoin::Bevel:
+        return Gfx::Path::JoinStyle::Bevel;
+    case Bindings::CanvasLineJoin::Miter:
+        return Gfx::Path::JoinStyle::Miter;
+    }
+
+    VERIFY_NOT_REACHED();
+}
```
AtkinsSJ: These definitely should be moved somewhere instead of copied.

Libraries/LibWeb/HTML/HTMLCanvasElement.cpp
```diff
@@ -367,6 +388,35 @@ WebIDL::ExceptionOr<void> HTMLCanvasElement::to_blob(GC::Ref<WebIDL::CallbackTyp
...
 }
 
+// https://html.spec.whatwg.org/multipage/canvas.html#dom-canvas-transfercontroltooffscreen
+JS::ThrowCompletionOr<GC::Ref<OffscreenCanvas>> HTMLCanvasElement::transfer_control_to_offscreen()
+{
+    // 1. If this canvas element's context mode is not set to none, throw an "InvalidStateError" DOMException.
+    auto is_valid = m_context.visit([&](GC::Ref<CanvasRenderingContext2D>&) { return false; }, [&](GC::Ref<WebGL::WebGLRenderingContext>&) { return false; }, [&](GC::Ref<WebGL::WebGL2RenderingContext>&) { return false; }, [&](GC::Ref<OffscreenCanvas>&) { return false; }, [](Empty) { return true; });
+    if (!is_valid) {
+        return throw_completion(WebIDL::InvalidStateError::create(realm(), Utf16String::formatted("Context mode is none")));
+    }
```
AtkinsSJ: You can simplify this to `if (!m_context.has<Empty>()) { ... }`

Libraries/LibWeb/HTML/HTMLCanvasElement.cpp
```diff
@@ -367,6 +388,35 @@ WebIDL::ExceptionOr<void> HTMLCanvasElement::to_blob(GC::Ref<WebIDL::CallbackTyp
...
+    auto offscreen_canvas = OffscreenCanvas::create(this->realm(), width(), height());
+
+    // 3. Set the offscreenCanvas's placeholder canvas element to a weak reference to this canvas element.
+    offscreen_canvas->set_placeholder_canvas(this->make_weak_ptr());
+
+    // 4. Set this canvas element's context mode to placeholder.
+    m_context = offscreen_canvas;
+
+    // 5. Set the offscreenCanvas's inherited language to the language of this canvas element.
+    // offscreen_canvas.
```
AtkinsSJ: Looks like this line shouldn't be here.

Libraries/LibWeb/HTML/OffscreenCanvas.cpp
```diff
@@ -82,25 +99,57 @@ OffscreenCanvas::OffscreenCanvas(JS::Realm& realm, RefPtr<Gfx::Bitmap> bitmap)
...
 OffscreenCanvas::~OffscreenCanvas() = default;
 
-WebIDL::ExceptionOr<void> OffscreenCanvas::transfer_steps(HTML::TransferDataEncoder&)
+// https://html.spec.whatwg.org/multipage/canvas.html#the-offscreencanvas-interface:transfer-steps
+WebIDL::ExceptionOr<void> OffscreenCanvas::transfer_steps(HTML::TransferDataEncoder& data_holder)
 {
-    // FIXME: Implement this
-    dbgln("(STUBBED) OffscreenCanvas::transfer_steps(HTML::TransferDataEncoder&)");
+    // 1. If value's context mode is not equal to none, then throw an "InvalidStateError" DOMException.
+    auto is_valid = m_context.visit([&](GC::Ref<OffscreenCanvasRenderingContext2D>&) { return false; }, [&](GC::Ref<WebGL::WebGLRenderingContext>&) { return false; }, [&](GC::Ref<WebGL::WebGL2RenderingContext>&) { return false; }, [](Detached) { return false; }, [](Empty) { return true; });
```
AtkinsSJ: See earlier point about `has<Empty>()`

Libraries/LibWeb/HTML/OffscreenCanvas.h
```diff
@@ -81,10 +97,16 @@ class OffscreenCanvas : public DOM::EventTarget
 
     void reset_context_to_default_state();
     WebIDL::ExceptionOr<void> set_new_bitmap_size(Gfx::IntSize new_size);
-
-    Variant<GC::Ref<HTML::OffscreenCanvasRenderingContext2D>, GC::Ref<WebGL::WebGLRenderingContext>, GC::Ref<WebGL::WebGL2RenderingContext>, Empty> m_context;
+    enum class Detached {
+        yes
+    };
```
AtkinsSJ: If you're just using this as a tag, make it an empty struct. `struct Detached {};`

Libraries/LibWeb/HTML/HTMLCanvasElement.h
```diff
@@ -65,7 +71,7 @@ class HTMLCanvasElement final : public HTMLElement {
     void reset_context_to_default_state();
     void notify_context_about_canvas_size_change();
 
-    Variant<GC::Ref<HTML::CanvasRenderingContext2D>, GC::Ref<WebGL::WebGLRenderingContext>, GC::Ref<WebGL::WebGL2RenderingContext>, Empty> m_context;
+    Variant<GC::Ref<HTML::CanvasRenderingContext2D>, GC::Ref<WebGL::WebGLRenderingContext>, GC::Ref<WebGL::WebGL2RenderingContext>, GC::Ref<OffscreenCanvas>, Empty> m_context;
```
AtkinsSJ, line 83: Do you need to add OffscreenCanvas to the `using RenderingContext` at the top too?

AtkinsSJ (changes requested):
I'm not especially familiar with canvas, but I've given this a look through.

Libraries/LibWeb/HTML/OffscreenCanvasRenderingContext2D.cpp
me-it-is: I just pushed a refactor, does the final state look good, before i go back to fixup the history?

Libraries/LibWeb/HTML/HTMLCanvasElement.cpp
me-it-is: I'm using the variant of off screen canvas as the placeholder should I instead use a struct with an off screen canvas or is there some better solution?

AtkinsSJ (changes requested):
I do see a couple of issues with this:
1. You've moved a lot of code into header files that doesn't need to be there. https://github.com/LadybirdBrowser/ladybird/commit/64f438857b28aecc6d0f8b88d53a5f1dda2f022c recently did the opposite to try and reduce build times.
2. I think moving functionality into `CanvasDrawPath`, `CanvasRect`, etc instead keeps things closer to the spec.

me-it-is:
I think the refactor is ready for review, all of the test failures seem unrealated but im not entirely sure.

Libraries/LibWeb/HTML/HTMLCanvasElement.h
me-it-is: I dont think so because RenderingContext is only returned from get rendering context which cant return a placeholder canvas

Libraries/LibWeb/HTML/HTMLCanvasElement.cpp
gmta: Sam is referring to this line:

```cpp
// offscreen_canvas.
```

This is not in the spec and it doesn't add anything, so it should be removed.

gmta (changes requested):

Libraries/LibWeb/HTML/HTMLCanvasElement.cpp
me-it-is: I see. its fixed now.

kalenikaliaksandr (changes requested):
I only glanced through the changes. The refactoring to share a common base between the two contexts looks like a reasonable direction, but if you are interested in OffscreenCanvas support, then we should start by figuring out how to share the underlying canvas surface between the WebContent and WebWorker processes, since, in my understanding, that's mainly how it's used on real websites.

me-it-is:
For offscreen canvas support you can use it in the same thread, though as you said its more common to move it to a seperate one. For now should i move it to a seprerate pr?

PR #6715, opened 2025-11-06
LibWeb: Fire click event when hitting Enter/space on focused element

### LINKED ISSUE #6703: `Enter` or space key should trigger `onclick` event for focused button
### Summary

When a button has focus, hitting the enter or space key should trigger it's `onclick` event.

This was fixed in https://github.com/LadybirdBrowser/ladybird/pull/3536#issue-2843936851 (minus the space key support), but was never merged.

For reference, here is the corresponding [WTP test](https://github.com/web-platform-tests/wpt/blob/2835fc21702841d140bad6701fd5aec0df38d312/html/editing/focus/sequential-focus-navigation-and-the-tabindex-attribute/focus-tabindex-event.html#L1), although that test only checks the enter key.

### Operating system

Linux

### Steps to reproduce

1. Load reproduction below
2. Hit tab key to focus button
3. Hit enter (or space)

### Expected behavior

Alert should pop up.

### Actual behavior

No alert pops up.

N/A

```html
<button onclick="alert(1)">Button</button>

<style>
button { border: none; }
button:focus { background: blue; }
</style>
```

```shell
N/A (no error from Ladybird).
```

Image

_No response_

### Contribute a patch?

dosisod:
Now when an element is focused, hitting the Enter/space key will fire a click event on that element. This behaviour is consistent with both Firefox and Chromium.

Closes #6703.

### Notes

One potential caveat is what to do with keyboard modifiers: On Linux, Firefox doesn't care if any modifier is used, but Chromium does not fire events for the Enter key when used with certain modifiers. I haven't tested any other browsers, or tested these browsers on MacOS/Windows.

Here is a list of the modifiers I tested, and how each browser handled it:

**Legend**

| Status | Meaning |
|--------|---------|
| (empty)| No click event fired |
| X      | Single click event fired |
| Each   | A click event is fired for Each key press/repeat |
| Up     | A click event is only fired on keyup |

| Browser  | Repeat? | Key + | None | NumPad | LeftCtrl | RightCtrl | LeftShift | RightShift | LeftAlt | RightAlt | Super |
|----------|---------|-------|------|--------|----------|-----------|-----------|------------|---------|----------|-------|
| Ladybird | No      | Enter | X    | X      | X        | X         | X         | X          | X       | X        | X     |
| Ladybird | No      | Space | X    | N/A    | X        | X         | X         | X          | X       | X        | X     |
| Ladybird | Yes     | Enter | Each | Each   | Each     | Each      | Each      | Each       | Each    | Each     | Each  |
| Ladybird | Yes     | Space | Up   | N/A    | Up       | Up        | Up        | Up         | Up      | Up       | Up    |
|          |         |       |      |        |          |           |           |            |         |          |       |
| Firefox  | No      | Enter | X    | X      | X        | X         | X         | X          | X       | X        | X     |
| Firefox  | No      | Space | X    | N/A    | X        | X         | X         | X          | X       | X        | X     |
| Firefox  | Yes     | Enter | Each | Each   | Each     | Each      | Each      | Each       | Each    | Each     | Each  |
| Firefox  | Yes     | Space | Up   | N/A    | Up       | Up        | Up        | Up         | Up      | Up       | Up    |
|          |         |       |      |        |          |           |           |            |         |          |       |
| Chromium | No      | Enter | X    | X      |          |           | X         | X          |         |          | X     |
| Chromium | No      | Space | X    | N/A    |          |           | X         | X          |         |          | X     | 
| Chromium | Yes     | Enter | Each | Each   |       |        | Each      | Each       |     |      | Each  |
| Chromium | Yes     | Space | Up   | N/A    | Up       | Up        | Up        | Up         | Up      | Up       | Up    |

~~I don't know what the procedure is for handling discrepancies between browsers, but for me, ignoring the modifier seems easiest.~~

I've opted to use Firefox's behavior since it is the most straightforward to implement.

### DIFF
--- a/Libraries/LibWeb/Page/EventHandler.cpp
+++ b/Libraries/LibWeb/Page/EventHandler.cpp
@@ -1519,6 +1519,12 @@ EventResult EventHandler::handle_keydown(UIEvents::KeyCode key, u32 modifiers, u
     case UIEvents::KeyCode::Key_End:
         document->window()->scroll_by(0, INT64_MAX);
         return EventResult::Handled;
+    case UIEvents::KeyCode::Key_Return:
+        if (auto* element = as_if<HTML::HTMLElement>(focused_area.ptr()); !element->has_attribute(HTML::AttributeNames::disabled)) {
+            element->click();
+            return EventResult::Handled;
+        }
+        break;
     default:
         break;
     }
@@ -1533,7 +1539,20 @@ EventResult EventHandler::handle_keyup(UIEvents::KeyCode key, u32 modifiers, u32
     if (repeat)
         return EventResult::Dropped;
 
-    return fire_keyboard_event(UIEvents::EventNames::keyup, m_navigable, key, modifiers, code_point, false);
+    auto dispatch_result = fire_keyboard_event(UIEvents::EventNames::keyup, m_navigable, key, modifiers, code_point, false);
+
+    if (dispatch_result != EventResult::Accepted)
+        return dispatch_result;
+
+    if (key == UIEvents::KeyCode::Key_Space) {
+        auto focused_area = m_navigable->active_document()->focused_area();
+        if (auto* element = as_if<HTML::HTMLElement>(focused_area.ptr()); !is<HTML::HTMLAnchorElement>(element) && !element->has_attribute(HTML::AttributeNames::disabled)) {
+            element->click();
+            return EventResult::Handled;
+        }
+    }
+
+    return dispatch_result;
 }
 
 EventResult EventHandler::handle_paste(Utf16String const& text)
--- a/Tests/LibWeb/Text/expected/keypress-on-focused-button-triggers-onclick.txt
+++ b/Tests/LibWeb/Text/expected/keypress-on-focused-button-triggers-onclick.txt
@@ -1,0 +1,8 @@
+<BUTTON id="button1">
+Clicked button 1
+Clicked button 1
+Clicked button 1
+Clicked button 1
+Clicked button 1
+Clicked button 1
+<BUTTON id="button2">
--- a/Tests/LibWeb/Text/input/keypress-on-focused-button-triggers-onclick.html
+++ b/Tests/LibWeb/Text/input/keypress-on-focused-button-triggers-onclick.html
@@ -1,0 +1,32 @@
+<!DOCTYPE html>
+<button id="button1">Button 1</button>
+<button id="button2">Button 2</button>
+<script src="include.js"></script>
+<script>
+    test(() => {
+        let button1 = document.getElementById("button1");
+        let button2 = document.getElementById("button2");
+
+        button1.addEventListener("click", e => {
+            println("Clicked button 1");
+        });
+        button2.addEventListener("click", e => {
+            println("Clicked button 2");
+        });
+
+        internals.sendKey(document.body, "Tab");
+        printElement(document.activeElement);
+        internals.sendKey(document.body, "Return");
+        internals.sendKey(document.body, "Return", internals.MOD_ALT);
+        internals.sendKey(document.body, "Return", internals.MOD_CTRL);
+        internals.sendKey(document.body, "Return", internals.MOD_SHIFT);
+        internals.sendKey(document.body, "Return", internals.MOD_SUPER);
+        internals.sendKey(document.body, "Return", internals.MOD_KEYPAD);
+
+        internals.sendKey(document.body, "Tab");
+        printElement(document.activeElement);
+        // FIXME: figure out a way to test keyup event
+        // document.activeElement.dispatchEvent(new KeyboardEvent("keyup", {key: "Space"}));
+        // internals.sendKey(document.body, "Space");
+    });
+</script>


Libraries/LibWeb/Page/EventHandler.cpp
```diff
@@ -1456,6 +1456,13 @@ EventResult EventHandler::handle_keydown(UIEvents::KeyCode key, u32 modifiers, u
     case UIEvents::KeyCode::Key_End:
         document->window()->scroll_by(0, INT64_MAX);
         return EventResult::Handled;
+    case UIEvents::KeyCode::Key_Return:
+    case UIEvents::KeyCode::Key_Space:
+        if (auto focused_area = document->focused_area(); is<HTML::HTMLElement>(*focused_area)) {
+            as<HTML::HTMLElement>(*focused_area).click();
+            return EventResult::Handled;
+        }
```
AtkinsSJ: Nit: You can combine the check and cast with `as_if`, which avoids some redundant type-checking:
```c++
if (auto* focused_area = as_if<HTML::HTMLElement>()) {
	focused_area->click();
	return EventResult::Handled;
}
```

trflynn89: And `focused_area` is actually already stored in a local variable above, so this would be something like:
```c++
if (auto* element = as_if<HTML::HTMLElement>(focused_area.ptr())) {
    element->click();
    return EventResult::Handled;
}
```

konradekk:
fyi: you seem to be adding some handling to `keydown`, but seemingly `Space` and `Enter`/`Return` handling differ slightly, as one is triggered on `keydown`, while other – on `keyup`.[1]

[1] https://adrianroselli.com/2022/04/brief-note-on-buttons-enter-and-space.html

AtkinsSJ:
> you seem to be adding some handling to `keydown`, but seemingly `Space` and `Enter`/`Return` handling differ slightly, as one is triggered on `keydown`, while other – on `keyup`.[1]
> 
> [1] [adrianroselli.com/2022/04/brief-note-on-buttons-enter-and-space.html](https://adrianroselli.com/2022/04/brief-note-on-buttons-enter-and-space.html)

Is that behaviour required? It sounds like an implementation quirk.

konradekk:
> Is that behaviour required? It sounds like an implementation quirk.

just noting! as for me, i was also quite unsettled when i first learned about those differences. i think we should stick to the specs (as mentioned); the rest is just the usual tension between that and established practice. up to you, as the maintainer, what goes in and what stays out! 😜

dosisod:
One extra wrinkle is that the spec explicitly specifies that `keydown` should trigger the click event, not `keyup`. From https://w3c.github.io/uievents/#event-type-keydown (emphasis added):

> The default action of the `keydown` event depends upon the key:
> 
> * If the key is the `Enter` or ` ` (Space) key and the current focus is on a state-changing element, ***the default action MUST be to dispatch a click event, and a DOMActivate event*** if that event type is supported by the user agent.

The [keyup](https://w3c.github.io/uievents/#keyup) event spec makes no such mention of how the Enter/space key should be handled.

I created an MVP of the [adrianroselli.com/2022/04/brief-note-on-buttons-enter-and-space.html](https://adrianroselli.com/2022/04/brief-note-on-buttons-enter-and-space.html) code for testing below:

<details>
<summary>Expand for code</summary>

```html
<button onclick="updateCount();" onkeydown="enterCount(event);" onkeyup="spaceCount(event);">
Trigger me
</button>


<label for="CountTotal">Total</label>
<input id="CountTotal" value="0" />
<label for="CountEnter">Enter Count</label>
<input id="CountEnter" value="0" />
<label for="SpaceEnter">Space Count</label>
<input id="SpaceEnter" value="0" />


<script>
function updateCount() {
  var CountTotal = document.getElementById('CountTotal');
  CountTotal.value = Number(CountTotal.value) + 1;
}

function enterCount() {
  if (event.keyCode == 13) {
    console.log('ENTER');
    var CountEnter = document.getElementById('CountEnter');
    CountEnter.value = Number(CountEnter.value) + 1;
  }
}

function spaceCount() {
  if (event.keyCode == 32) {
    console.log('SPACE');
    var SpaceEnter = document.getElementById('SpaceEnter');
    SpaceEnter.value = Number(SpaceEnter.value) + 1;
  }
}
</script>
```
</details>

Currently, this PR handles Enter/space the same, meaning holding Enter or space will both spam click events.

I believe Ladybird should follow the lead of Firefox and Chromium when it comes to key repeats to avoid breaking accessibility norms. With that being said, as it pertains to the `keydown` spec, either:

* The spec is wrong/outdated, and browsers have decided to follow this `keydown`/`keyup` repeat logic
* The spec is correct, and browsers have chosen to do something that is wrong
* There is a different spec which covers and explains all this behavior, and we need to find it

Thoughts on how to proceed?

AtkinsSJ:
Sorry for not looking at this!

> I believe Ladybird should follow the lead of Firefox and Chromium when it comes to key repeats to avoid breaking accessibility norms.

Agreed.

> Thoughts on how to proceed?

I think the best thing to do would be to raise an issue on the UI Events spec about it and see what they think. That'd be https://github.com/w3c/uievents/issues . It'll also give you something to link to in comments so it's easy to see what isn't to spec and why.

dosisod:
Seems like someone already asked the same question a while back: https://github.com/w3c/uievents/issues/174

I'll give some more context on that issue and see if they have any input.

In the mean time though I can mimic the same behavior as Chrome/Firefox.

dosisod:
Adding the finishing touches to this today. For now I will just mimic the key press/repeat behavior of Firefox, as their logic is the most sensible and straightforward.

Libraries/LibWeb/Page/EventHandler.cpp
```diff
@@ -1474,6 +1480,14 @@ EventResult EventHandler::handle_keyup(UIEvents::KeyCode key, u32 modifiers, u32
     if (repeat)
         return EventResult::Dropped;
 
+    if (key == UIEvents::KeyCode::Key_Space) {
+        auto focused_area = m_navigable->active_document()->focused_area();
+        if (auto* element = as_if<HTML::HTMLElement>(focused_area.ptr())) {
+            element->click();
+            return EventResult::Handled;
+        }
+    }
```
AtkinsSJ, line 1553: Should this happen after `fire_keyboard_event()`? That's the case for pressing the Return key. The difference is whether JS can intercept and cancel the event.

konradekk: note that if link is focused, space key does not activate it in generally all the cases (but scrolls the page; the same as if the document was focused).

note also that `disabled` elements do are not activated (the same for ones with `preventDefault` called). not sure what’s under `fire_keyboard_event` – if these things exactly, then definitely before, but… other events shall be dispatched after default activation (`click`), so definitely after. 😅

probably it is sensible to check for specific ui controls: `<button>`, `<input type="button">` and other button-like inputs (`submit` seems to be kind of special), `<input type="checkbox">`, `<input type="radio">`…?

i guess that adhering closely to spec might be crucial here (few more tests would be helpful here? for disabled and prevented states, dispatch order of events etc.).

dosisod: > note that if link is focused, space key does not activate it in generally all the cases

Done. AFAICT there is currently no scroll page support for the space key, that can come separately though.

> note also that `disabled` elements do are not activated

Disabled elements can't be focused, at least not from my testing (via tab or through `.focus()`). I *could* double check the disabled status before activating, but unless there is a different way of focusing elements, I don't see the point.

> the same for ones with `preventDefault` called

Good call. I did some testing and it looks like calling `preventDefault()` inside `keydown` doesn't prevent `keyup` from being called later. Might just be my branch, need to validate.

> probably it is sensible to check for specific ui controls

Agreed. This "small" feature is turning out to be more involved than I thought :smiling_face_with_tear: I also noticed some keyboard inconsistencies when using radio buttons, will open separate issue.

> i guess that adhering closely to spec might be crucial here (few more tests would be helpful here?

Agreed. I will make sure to test as many edge cases as I can think of. I will need a way to manually trigger `keyup` events to fully test this though

konradekk: your work is a good step forward, though! optionally just mark a cut-off point and try to integrate something with a clear way forward! 😅 (i’m the last one to stop the progress here! 😝)

dosisod:
I created a map of Firefox/Chromium's `keydown`/`keyup` event handing logic:

image

Ladybird's behavior mostly maps to this now, but one issue is the gray dashed line.

In short, if the `keydown` is cancelled when pressing the space key, the `click` event *should not* be fired after `keyup`. To support this, we need a way to convey to `keyup` that `keydown` was cancelled, and that it should not fire `click`.

Of course this is just my interpretation of Firefox/Chromium's behavior. I combed through the `keydown`/`keyup`/`click` UI event specs, and don't see any references to this behavior (see https://github.com/w3c/uievents/issues/174).

Should we include this logic here, or would it be better to leave it out for now and implement it at a later date?

Libraries/LibWeb/Page/EventHandler.cpp
dosisod: Marking this particular comment as resolved. I added explicit `disabled` checks just in case, and tested the default behavior with `input` radio/checkbox/buttons, all is good there (need to add actual tests though).

If there are specific issues later on we can make a separate review comment to discuss.

dosisod: And yes, I shuffled around the `fire_keyboard_event()` call to match the enter key logic.

konradekk:
> Of course this is just my interpretation of Firefox/Chromium's behavior. I combed through the `keydown`/`keyup`/`click` UI event specs, and don't see any references to this behavior (see [w3c/uievents#174](https://github.com/w3c/uievents/issues/174)).

you can check the wpt tests (or others) for expected behaviour – not sure if there’re any, but these drive the implementation in other areas too.

> Should we include this logic here, or would it be better to leave it out for now and implement it at a later date?

it’s mostly up to you, if you want to pursue it: i guess that you should declare some clear cut-off points with `TODO` and `FIXME` comments, or just comment here (or in a commit messages) what was done, what was skipped, what is the way ahead. that’s at least an idea for avoiding analysis paralysis or scope creep. just try to imagine what you want to achieve with this change exactly and go for it! 😁

note: it need not to be perfect, it needs to compile and be at least slightly better from what it was before. 🙂

AtkinsSJ:
> And yes, I shuffled around the fire_keyboard_event() call to match the enter key logic.

Looks like you didn't push those changes. :sweat_smile:

dosisod:
Ok, *now* they are pushed :smile:

Libraries/LibWeb/Page/EventHandler.cpp
```diff
@@ -1519,6 +1519,12 @@ EventResult EventHandler::handle_keydown(UIEvents::KeyCode key, u32 modifiers, u
     case UIEvents::KeyCode::Key_End:
         document->window()->scroll_by(0, INT64_MAX);
         return EventResult::Handled;
+    case UIEvents::KeyCode::Key_Return:
+        if (auto* element = as_if<HTML::HTMLElement>(focused_area.ptr()); !element->has_attribute(HTML::AttributeNames::disabled)) {
```
AtkinsSJ, line 1523: Condition should check `element` too, ie `element && !element->has_attribute(...)`

Libraries/LibWeb/Page/EventHandler.cpp
```diff
@@ -1533,7 +1539,20 @@ EventResult EventHandler::handle_keyup(UIEvents::KeyCode key, u32 modifiers, u32
...
 
-    return fire_keyboard_event(UIEvents::EventNames::keyup, m_navigable, key, modifiers, code_point, false);
+    auto dispatch_result = fire_keyboard_event(UIEvents::EventNames::keyup, m_navigable, key, modifiers, code_point, false);
+
+    if (dispatch_result != EventResult::Accepted)
+        return dispatch_result;
+
+    if (key == UIEvents::KeyCode::Key_Space) {
+        auto focused_area = m_navigable->active_document()->focused_area();
+        if (auto* element = as_if<HTML::HTMLElement>(focused_area.ptr()); !is<HTML::HTMLAnchorElement>(element) && !element->has_attribute(HTML::AttributeNames::disabled)) {
```
AtkinsSJ, line 1549: Missing an `element` check here too.

dosisod:
Sorry for letting this go stale, I haven't had any free time to contribute recently. If someone wants to resurrect this and take over, feel free!

Issue #2012, opened 2024-10-27
LibWeb: Crash on 3bmeteo.com


teaalltr:
The crash is related to formatting, CSS positioning, a `VERIFY_NOT_REACHED` fail saying `// If we get here, ancestor_box was not an ancestor of "box"!`. The crash happens at the very beginning of page load. Reduction is below

```
VERIFICATION FAILED: false at /home/marco/ladybird/Userland/Libraries/LibWeb/Layout/FormattingContext.cpp:1179
/home/marco/ladybird/Build/ladybird/libexec/../lib/liblagom-ak.so.0(ak_verification_failed+0x87) [0x7f745bcd49b7]
/home/marco/ladybird/Build/ladybird/libexec/../lib/liblagom-web.so.0(+0xb2a5c2) [0x7f74602705c2]
/home/marco/ladybird/Build/ladybird/libexec/../lib/liblagom-web.so.0 Web::Layout::BlockFormattingContext::layout_block_level_box(Web::Layout::Box const&, Web::Layout::BlockContainer const&, Web::CSSPixels&, Web::Layout::AvailableSpace const&) 0x166) [0x7f746025e426]
/home/marco/ladybird/Build/ladybird/libexec/../lib/liblagom-web.so.0 Web::Layout::BlockFormattingContext::layout_block_level_children(Web::Layout::BlockContainer const&, Web::Layout::AvailableSpace const&) 0x94) [0x7f746025f6e4]
/home/marco/ladybird/Build/ladybird/libexec/../lib/liblagom-web.so.0 Web::Layout::BlockFormattingContext::run(Web::Layout::AvailableSpace const&) 0x39) [0x7f746025faa9]
/home/marco/ladybird/Build/ladybird/libexec/../lib/liblagom-web.so.0(+0xb2fa73) [0x7f7460275a73]
/home/marco/ladybird/Build/ladybird/libexec/../lib/liblagom-web.so.0 Web::Layout::TableFormattingContext::compute_cell_measures() 0x155) [0x7f74602bd305]
/home/marco/ladybird/Build/ladybird/libexec/../lib/liblagom-web.so.0 Web::Layout::TableFormattingContext::run_until_width_calculation(Web::Layout::AvailableSpace const&) 0x72) [0x7f74602c13a2]
[...]

PR #6719, opened 2025-11-06
LibWeb: Fix crash when abspos element is relative to table row

Psychpsyo:
This occured because the abspos containing block is in-between the containing blocks of two regular ancestors in this case, so it was being skipped. (The containing block of a table cell is the table itself, not the table row.)

This allows us to open and read the blogpost about #3011 without crashing: https://www.ryanliptak.com/blog/better-named-character-reference-tokenization/

### DIFF
--- a/Libraries/LibWeb/Layout/FormattingContext.cpp
+++ b/Libraries/LibWeb/Layout/FormattingContext.cpp
@@ -1187,11 +1187,16 @@ CSSPixelRect FormattingContext::content_box_rect_in_static_position_ancestor_coo
     CSSPixelRect rect = { { 0, 0 }, box_used_values.content_size() };
     // FIXME: ListItemMarkerBox's should also run this assertion once it has a supported FormattingContext type
     VERIFY(box_used_values.offset.is_zero() || box.is_list_item_marker_box()); // Set as result of this calculation
-    for (auto const* current = box.static_position_containing_block(); current; current = current->containing_block()) {
+    auto next_containing_block = box.static_position_containing_block();
+    for (NodeWithStyle const* current = box.static_position_containing_block(); current; current = current->parent()) {
         if (current == box.containing_block())
             return rect;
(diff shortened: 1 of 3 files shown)


Psychpsyo:
Timeout seems unrelated (and did not happen locally for me)

AtkinsSJ:
Fix seems good but please rebase. Also, could we get a layout test for this and not just a crash test?

AtkinsSJ:
The new test is failing.

Libraries/LibWeb/Layout/FormattingContext.cpp
```diff
@@ -1187,11 +1187,16 @@ CSSPixelRect FormattingContext::content_box_rect_in_static_position_ancestor_coo
...
-        rect.translate_by(current_state.offset);
+        // Whenever we walk past other containing blocks, we need to offset the rect by those.
+        if (current == next_containing_block) {
+            auto const& current_state = m_state.get(*current);
+            rect.translate_by(current_state.offset);
+            next_containing_block = next_containing_block->containing_block();
+        }
     }
     // If we get here, `ancestor_box` was not in the containing block chain of the static position containing block of `box`!
     // Something about the containing block chain is set up incorrectly then.
```
InvalidUsernameException, line 1202: When I wrote this comment I was under the assumption that the case fixed in this PR is not possible. We are not really even walking the containing block chain anymore, so the comment is no longer accurate. Maybe we can go with a less specific comment again since it looks like I went too specific the last time.

InvalidUsernameException (commented):
I was kind of surprised to see this case happening, but this change appears to match the spec as far as I can tell.

Either way, this fixes #2012 and also a crash on [the Matomo dashboard](https://demo.matomo.cloud).

AtkinsSJ:
This fixes the issues I was having in #7158! :tada: I agree about changing the inaccurate comment, but otherwise I think this is good to go.

PR #6724, opened 2025-11-06
LibWeb: Improve scrollingElement handling

Psychpsyo:
This change is currently entirely undetectable because of what the added FIXME talks about.
Currently, the HTML element's overflow is always set to visible in both axes, so it getting set to "clip" in the imported test ends up not mattering at all.

I imported the test anyways cause it's very related and will make it easy for whoever decides to tackle that FIXME.


AtkinsSJ:
This seems likely to conflict with #6727 so maybe wait for that before rebasing.

AtkinsSJ:
Please rebase this. :^)

Psychpsyo:
> Please rebase this. :^)

Done now.

PR #6728, opened 2025-11-06
LibJS: Replace Array.fromAsync with a native JavaScript implementation

Lubrsi:
This allows us to use the bytecode implementation of await, which
correctly suspends execution contexts and handles completion
injections.

This gains us 4 test262 tests around mutating Array.fromAsync's
iterable whilst it's suspended as well.

This is also one step towards removing spin_until, which the
non-bytecode implementation of await uses.

```
Duration:
     -5.98s

Summary:
    Diff Tests:
        +4 ✅    -4 ❌

Diff Tests:
    [...]/Array/fromAsync/asyncitems-array-add-to-singleton.js ❌ -> ✅
    [...]/Array/fromAsync/asyncitems-array-add.js              ❌ -> ✅
    [...]/Array/fromAsync/asyncitems-array-mutate.js           ❌ -> ✅
    [...]/Array/fromAsync/asyncitems-array-remove.js           ❌ -> ✅
```


Libraries/LibJS/Bytecode/Instruction.h
```diff
@@ -172,6 +179,9 @@
     O(ThrowIfNotObject)                \
     O(ThrowIfNullish)                  \
     O(ThrowIfTDZ)                      \
+    O(ToBoolean)                       \
```
Hendiadyoin1: Couldn't this technically be done with `!!`
(alternatively we could now peephole `!!` to this)

Libraries/LibJS/Runtime/Intrinsics.cpp
```diff
@@ -490,6 +521,50 @@ GC::Ref<Intl::Collator> Intrinsics::default_collator()
     return *m_default_collator;
 }
 
+static constexpr char const ABSTRACT_OPERATIONS[] = {
+#embed "JavaScriptImplementations/AbstractOperations.js"
```
Hendiadyoin1: technically you could add the trailing comma with a `suffix`, but that only really matters if the source is empty

Libraries/LibJS/Runtime/JavaScriptImplementations/AbstractOperations.js
Hendiadyoin1, line 1: Any particular reason these are native now?
or were they inlined previously in the fromAsync AO

Lubrsi: Besides AsyncIteratorClose, there's not a major reason other than minimising the amount of macros needed. AsyncIteratorClose needs to be native because it uses await _and_ it's in an unwind context.

Libraries/LibJS/Bytecode/Generator.cpp
```diff
@@ -1471,4 +1472,302 @@ ScopedOperand Generator::add_constant(Value value)
...
+    if (constant_name == "undefined"sv) {
+        return add_constant(js_undefined());
+    }
+
+    if (!m_builtin_abstract_operations_enabled)
+        return OptionalNone {};
+
+    if (constant_name == "SYMBOL_ITERATOR"sv) {
+        auto constant_register = allocate_register();
+        emit<Op::GetWellKnownSymbol>(constant_register, Op::GetWellKnownSymbol::WellKnownSymbol::Iterator);
```
awesomekling: Can we just emit the well-known symbol as a constant here, instead of an instruction?

PR #6732, opened 2025-11-07
LibWeb: Check that import map elements are not arrays

danielbprice:
Previously
```
<html>
<head>
	<script type="importmap"> { "imports": [] } </script>
</head>
</html>
```

was accepted with no warning on console.  Now this will print

```
4530079.264 WebContent(16164): Unhandled JavaScript exception: [TypeError] The 'imports' top-level value of an importmap needs to be a JSON object.
4530079.264 WebContent(16164):
```

### DIFF
--- a/Libraries/LibWeb/HTML/Scripting/ImportMap.cpp
+++ b/Libraries/LibWeb/HTML/Scripting/ImportMap.cpp
@@ -17,6 +17,14 @@
 
 namespace Web::HTML {
 
+// Check that the JS value was produced from a JSON object ({}).
+// https://infra.spec.whatwg.org/#ordered-map
+static bool
+value_is_ordered_map(JS::Value const& value)
+{
+    return value.is_object() && !is<JS::Array>(value.as_object());
+}
+
 // https://html.spec.whatwg.org/multipage/webappapis.html#parse-an-import-map-string
 WebIDL::ExceptionOr<ImportMap> parse_import_map_string(JS::Realm& realm, ByteString const& input, URL::URL base_url)
 {
@@ -26,7 +34,7 @@ WebIDL::ExceptionOr<ImportMap> parse_import_map_string(JS::Realm& realm, ByteStr
     auto parsed = TRY(Infra::parse_json_string_to_javascript_value(realm, input));
 
     // 2. If parsed is not an ordered map, then throw a TypeError indicating that the top-level value needs to be a JSON object.
-    if (!parsed.is_object())
+    if (!value_is_ordered_map(parsed))
         return WebIDL::SimpleException { WebIDL::SimpleExceptionType::TypeError, "The top-level value of an importmap needs to be a JSON object."_string };
     auto& parsed_object = parsed.as_object();
 
@@ -38,7 +46,7 @@ WebIDL::ExceptionOr<ImportMap> parse_import_map_string(JS::Realm& realm, ByteStr
         auto imports = TRY(parsed_object.get("imports"_utf16_fly_string));
 
         // If parsed["imports"] is not an ordered map, then throw a TypeError indicating that the value for the "imports" top-level key needs to be a JSON object.
-        if (!imports.is_object())
+        if (!value_is_ordered_map(imports))
             return WebIDL::SimpleException { WebIDL::SimpleExceptionType::TypeError, "The 'imports' top-level value of an importmap needs to be a JSON object."_string };
 
         // Set sortedAndNormalizedImports to the result of sorting and normalizing a module specifier map given parsed["imports"] and baseURL.
@@ -53,7 +61,7 @@ WebIDL::ExceptionOr<ImportMap> parse_import_map_string(JS::Realm& realm, ByteStr
         auto scopes = TRY(parsed_object.get("scopes"_utf16_fly_string));
 
         // If parsed["scopes"] is not an ordered map, then throw a TypeError indicating that the value for the "scopes" top-level key needs to be a JSON object.
-        if (!scopes.is_object())
+        if (!value_is_ordered_map(scopes))
             return WebIDL::SimpleException { WebIDL::SimpleExceptionType::TypeError, "The 'scopes' top-level value of an importmap needs to be a JSON object."_string };
 
         // Set sortedAndNormalizedScopes to the result of sorting and normalizing scopes given parsed["scopes"] and baseURL.
@@ -68,7 +76,7 @@ WebIDL::ExceptionOr<ImportMap> parse_import_map_string(JS::Realm& realm, ByteStr
         auto integrity = TRY(parsed_object.get("integrity"_utf16_fly_string));
 
         // 1. If parsed["integrity"] is not an ordered map, then throw a TypeError indicating that the value for the "integrity" top-level key needs to be a JSON object.
-        if (!integrity.is_object())
+        if (!value_is_ordered_map(integrity))
             return WebIDL::SimpleException { WebIDL::SimpleExceptionType::TypeError, "The 'integrity' top-level value of an importmap needs to be a JSON object."_string };
 
         // 2. Set normalizedIntegrity to the result of normalizing a module integrity map given parsed["integrity"] and baseURL.


Libraries/LibWeb/HTML/Scripting/ImportMap.cpp
```diff
@@ -38,7 +39,7 @@ WebIDL::ExceptionOr<ImportMap> parse_import_map_string(JS::Realm& realm, ByteStr
         auto imports = TRY(parsed_object.get("imports"_utf16_fly_string));
 
         // If parsed["imports"] is not an ordered map, then throw a TypeError indicating that the value for the "imports" top-level key needs to be a JSON object.
-        if (!imports.is_object())
+        if (!imports.is_object() || is<JS::Array>(imports.as_object()))
             return WebIDL::SimpleException { WebIDL::SimpleExceptionType::TypeError, "The 'imports' top-level value of an importmap needs to be a JSON object."_string };
 
         // Set sortedAndNormalizedImports to the result of sorting and normalizing a module specifier map given parsed["imports"] and baseURL.
```
konradekk, line 52: not sure, but `null` is also an object… is this valid or we could add it here as an error too?

(i feel that it is the shape that ultimately should be checked, not type… 🤔)

danielbprice: > not sure, but null is also an object… is this valid or we could add it here as an error too?

The tests in `Tests/LibWeb/Text/input/HTML/import-maps-invalid.html` do test the `null` case, and that case seems to work correctly-- triggering the various guardrails in this code.  My reading of `JS::Value::is_object` is that it will only return True for things which are "actually" objects, and not for the `null` value.  The handling of null seems to be in the implementation of typeof, in `Value::typeof_`.   And otherwise handled by the `is_null()` test in `Value`.  *handwaving*.  I am not an expert!

> (i feel that it is the shape that ultimately should be checked, not type… 🤔)

I don't know enough browser internals to have an opinion.  Sorry!

I experimented with `imports.as_object().shape().is_dictionary()`, but that definitely doesn't work.

I am happy to make changes but I need a little more direction.  I agree that this implementation looks awkward.

konradekk: it seems we check (so far) whether `parsed` is a JS object, but it is necessary to validate against JSON object. it seems that your fix should suffice, as there’s only a handful of objects that can be a result of JSON conversion. imho it is not the cleanest approach and thus deserves a comment. 🤔 soon a maintainer should have a look on it to guide you appropriately! 👍

(your approach with dictionary shape is misled: it is a kind of optimisation mechanism and not every object will be using it.)

Zaggy1024: Definitely not my area of expertise, but it appears that `JS::Map` is ordered, so perhaps this check should be `!imports.is_object() || !imports.as_object().is_map_object()`. If that's the case, then all the methods that those `Object`s are passed to can probably instead accept `Map const&`, which you should be able to safely `static_cast` the `Object&` to.

Libraries/LibWeb/HTML/Scripting/ImportMap.cpp
```diff
@@ -7,6 +7,7 @@
 
 #include <LibJS/Console.h>
 #include <LibJS/Runtime/ConsoleObject.h>
+#include <LibJS/Runtime/Map.h>
```
Zaggy1024: Is this include used?

Libraries/LibWeb/HTML/Scripting/ImportMap.cpp
Zaggy1024: Also, changing those parameters to `Map const&` allows you to directly iterate over the object's key/value pairs, instead of going through the object shape, from what I can tell, which I would figure might be more efficient.

danielbprice: @Zaggy1024 this is also *definitely* not *my* area of expertise.  But I am learning!   What I think I am seeing here is that this thing is not of type `JS::Map` -- which is the only thing which cause `is_map_object()` to return `true`.  That seems to be for https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map.

The import-map is produced by calling `JSON.parse` inside of the JS runtime.  So the result is going to be the same kind of thing that you would get in JS if you parsed the json.  Therefore:

```
$ node
> x = JSON.parse("{}")
{}
> typeof(x)
'object'
> x instanceof Map
false
> x instanceof Array
false
> x instanceof Object
true
```

Libraries/LibWeb/HTML/Scripting/ImportMap.cpp
danielbprice: no!  removed.

konradekk:
btw. you seem to be amending the code against the spec; you should put some comments there to indicate that (`AD-HOC`, i guess?). raising an issue at the spec repo is also something people practice here (and people there might fix or explain the specs properly).

danielbprice:
> btw. you seem to be amending the code against the spec; you should put some comments there to indicate that (`AD-HOC`, i guess?). raising an issue at the spec repo is also something people practice here (and people there might fix or explain the specs properly).

Could you help me to understand specifically where these changes contradict a specification?  I am assuming you are referencing the WhatWG spec.  What I read there is the following:

https://html.spec.whatwg.org/multipage/webappapis.html#import-map
> 
> 8.1.5.3 Import map processing model
> 
> Formally, an import map is a [struct](https://infra.spec.whatwg.org/#struct) with three [items](https://infra.spec.whatwg.org/#struct-item):
> 
> imports, a [module specifier map](https://html.spec.whatwg.org/multipage/webappapis.html#module-specifier-map);
> 
> scopes, an [ordered map](https://infra.spec.whatwg.org/#ordered-map) of [URLs](https://url.spec.whatwg.org/#concept-url) to [module specifier maps](https://html.spec.whatwg.org/multipage/webappapis.html#module-specifier-map); and
> 
> integrity, a [module integrity map](https://html.spec.whatwg.org/multipage/webappapis.html#module-integrity-map).
> 
> 
> 
>
> A module specifier map is an [ordered map](https://infra.spec.whatwg.org/#ordered-map) whose [keys](https://infra.spec.whatwg.org/#map-key) are [strings](https://infra.spec.whatwg.org/#string) and whose [values](https://infra.spec.whatwg.org/#map-value) are either [URLs](https://url.spec.whatwg.org/#concept-url) or nulls.
> 
> A module integrity map is an [ordered map](https://infra.spec.whatwg.org/#ordered-map) whose [keys](https://infra.spec.whatwg.org/#map-key) are [URLs](https://url.spec.whatwg.org/#concept-url) and whose [values](https://infra.spec.whatwg.org/#map-value) are [strings](https://infra.spec.whatwg.org/#string) that will be used as [integrity metadata](https://fetch.spec.whatwg.org/#concept-request-integrity-metadata).
> 
> An empty import map is an [import map](https://html.spec.whatwg.org/multipage/webappapis.html#import-map) with its [imports](https://html.spec.whatwg.org/multipage/webappapis.html#concept-import-map-imports) and [scopes](https://html.spec.whatwg.org/multipage/webappapis.html#concept-import-map-scopes) both being empty maps.

[Something I don't fully understand about the spec is-- why do they specify that the map is ordered?  The JSON object that you use to declare an import map is itself unordered (as best I understand) so this is curious spec choice to me].

I do not believe that I am make the code more divergent from this spec.  It seems that by ensuring that the elements are not arrays-- and therefore are conceptually "maps"-- the code becomes more aligned with the spec.  

The new behavior also follows what happens in other browsers when given this test case:

```
<html>
<head>
	<script type="importmap"> { "imports": [] } </script>
</head>
</html>
```

Chrome:
```
Uncaught TypeError: Failed to parse import map: "imports" top-level key must be a JSON object.
```

Firefox:
```
Uncaught TypeError: the imports top-level key needs to be a JSON object 
```

Ladybird (without this change):
```
<no output>
```

Ladybird (with this change):
```
5359562.893 WebContent(72835): Unhandled JavaScript exception: [TypeError] The 'imports' top-level value of an importmap needs to be a JSON object.
```

------

Update: I guess maybe you are saying the spec is imprecise about the formal schema of the JSON import-map?  I will keep reading.

------

Update 2: Maybe the confusion is here?  The spec says:

> 8.1.5.2 Import maps
>
> The [child text content](https://dom.spec.whatwg.org/#concept-child-text-content) of a [script](https://html.spec.whatwg.org/multipage/scripting.html#the-script-element) element representing an [import map](https://html.spec.whatwg.org/multipage/webappapis.html#import-map) must match the following import map authoring requirements:
>
> It must be valid JSON. [[JSON]](https://html.spec.whatwg.org/multipage/references.html#refsJSON)
>
> The JSON must represent a JSON object, with at most the three keys "imports", "scopes", and "integrity".
> ...


A `JSON object` is not the same as a JS Object.  "JSON Object" here specifically refers to a key/value collection.  See https://datatracker.ietf.org/doc/html/rfc8259#section-4.  So the specification is sufficiently precise.  But because in the ladybird code we don't have the underlying JSON representation at this point in the code, we wind up needing to check the Javascript representation of the underlying JSON.

lpas:
> [Something I don't fully understand about the spec is-- why do they specify that the map is ordered? The JSON object that you use to declare an import map is itself unordered (as best I understand) so this is curious spec choice to me].

Map means the infra map https://infra.spec.whatwg.org/#ordered-map they are ordered for interoperability, see the set note https://infra.spec.whatwg.org/#ordered-set. That just means that if you have a map like `{"foo": "bar" , "foobar": "baz"}` and a user would loop over it with eg. a for loop like for(const x in map) that you will get the foo always as first item and the foobar as second item. That is just the normal behavior you have in JS as well.

From my point of view the code here looks right. The check for array seems right these are not JS maps

Zaggy1024:
Apologies for leaving this for so long, but it does appear to be correct from what I can tell (thanks lpas for finding that definition), and [Firefox agrees](https://searchfox.org/firefox-main/source/js/loader/ImportMap.cpp#213-232).

I wonder if there are any WPT tests for this case that we can import? If not, we should have a text test to cover this.

It would also be nice to factor these duplicated checks out into a function.

danielbprice:
> Apologies for leaving this for so long, but it does appear to be correct from what I can tell (thanks lpas for finding that definition), and [Firefox agrees](https://searchfox.org/firefox-main/source/js/loader/ImportMap.cpp#213-232).
> 
> I wonder if there are any WPT tests for this case that we can import? If not, we should have a text test to cover this.

Thank you @Zaggy1024 .   There is an existing test which I had previously enhanced, [Tests/LibWeb/Text/input/HTML/import-maps-invalid.html](https://github.com/LadybirdBrowser/ladybird/blob/master/Tests/LibWeb/Text/input/HTML/import-maps-invalid.html).  I'm not very satisfied with it, as it really only tests that the browser tolerates bad input without a crash.  It doesn't test that the correct errors are triggered or that the checks are accurate.  If I could assert that the console produced specific outputs, or that the import map parse resulted in some specific exceptions being raised, then the test could be more complete.  I do not know how to do that but if there's an example I can follow, I'll happily give it a try.

I'm not knowledgeable about WPT, but I did go through everything in https://wpt.fyi/results/import-maps?label=master&product=ladybird and couldn't find any tests which seemed to exercise these behaviors.

> It would also be nice to factor these duplicated checks out into a function.

Done.

Zaggy1024:
It should be okay to just use our own text test for this then, although it might be nice to upstream a wpt looking at the exception it throws for this as well, if that interests you.

Zaggy1024:
Actually, I had forgotten that text tests don't capture any console output, so as I understand it, we have no way to capture the result of the parse here for a test. This should be fine to merge then.

danielbprice:
@Zaggy1024 thanks so much for your help!  Much appreciated.

PR #6734, opened 2025-11-07
Base: Use the correct memory unit suffixes in `about::processes`

tcl3:
Previously the GB suffix was displayed as BB.

This seems to be a limitation of Intl.NumberFormat, as it also behaves this way in other browsers

Before:
image

After:
image

### DIFF
--- a/Base/res/ladybird/about-pages/processes.html
+++ b/Base/res/ladybird/about-pages/processes.html
@@ -76,18 +76,16 @@
             </thead>
             <tbody id="process-table"></tbody>
         </table>
-
-        <script type="text/javascript">
+        <script type="module">
+            import { getByteFormatter }  from "resource://ladybird/utils.js";
+            const memoryFormatter = getByteFormatter(() => {
+                return {
+                    unitDisplay: "narrow",
+                    minimumFractionDigits: 2,
+                    maximumFractionDigits: 2,
+                }
+            })
             const cpuFormatter = new Intl.NumberFormat([], {
-                minimumFractionDigits: 2,
-                maximumFractionDigits: 2,
-            });
-
-            const memoryFormatter = new Intl.NumberFormat([], {
-                style: "unit",
-                unit: "byte",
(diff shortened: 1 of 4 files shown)


Base/res/ladybird/about-pages/processes.html
```diff
@@ -83,14 +83,41 @@
...
 
-            const memoryFormatter = new Intl.NumberFormat([], {
-                style: "unit",
-                unit: "byte",
-                notation: "compact",
-                unitDisplay: "narrow",
-                minimumFractionDigits: 2,
-                maximumFractionDigits: 2,
-            });
+            const memoryFormatters = [];
```
trflynn89: Ohh I ran into this recently in #6675:
https://github.com/LadybirdBrowser/ladybird/pull/6675/commits/6130d33cf8fb0cadfb608e9ca985a43536ca57da#diff-e32009b90cf09eca69c20325482fdd357ef21351fc917e41b60cf722f6e31af5

I also found this `Math.log` based solution but ended up going with a simpler `/ 1024` loop, since it would be hit a max of like 5 times. I don't have a strong preference for which we use, but maybe we want to put this in a separate .js helper file, so that whatever we do, we do it consistently. (My usage of this is is still in a PR, so I'm happy to do that myself there).

Base/res/ladybird/about-pages/processes.html
```diff
@@ -83,14 +83,41 @@
...
+            const formatMemory = (bytes) => {
+                const units = [
+                    'byte',
+                    'kibibyte',
```
trflynn89: Is this right? `kibibyte` is not a sanctioned NF unit:
https://tc39.es/ecma402/#table-sanctioned-single-unit-identifiers

Base/res/ladybird/about-pages/processes.html
tcl3: I've moved the memory formatting code into a separate helper file.

Issue #5042, opened 2025-06-09
Span element with ::before pseudo-element is not painted


AmusedNetwork:
### Summary

When I have a span with a ::before pseudo-element applied to its class with `content: ""` and `display: block`, the text content of the span is not painted.

This works with other elements, such as `<p>`, as well as with the "::after" pseudo-element. 

I have seen this in the wild on https://www.gov.uk/government/get-involved. The numbers under "Respond to consultations" should have labels of "Open consultations", and "Closed consultations in the past 12 months". 

### Operating system

Linux

### Steps to reproduce

1. Open the reduced testcase
2. Observe the text displayed on the screen

### Expected behavior

The text "Test" should be displayed.

### Actual behavior

The page is blank; there is no text displayed.

### URL for a reduced test case

N/A

### HTML/SVG/etc. source for a reduced test case

```html
<!DOCTYPE html>
<html>

<style>
  .label::before{
    content: "";
    display: block
  }
</style>

<span class="label">
  Test
</span>

</html>
```

### Log output and (if possible) backtrace

```shell
Layout Tree:

29359.062 WebContent(217990): Viewport <#document> at (0,0) content-size 1636x942 [0+0+0 1636 0+0+0] [0+0+0 942 0+0+0] children: not-inline
  BlockContainer <html> at (0,0) content-size 1636x16 [0+0+0 1636 0+0+0] [0+0+0 16 0+0+0] [BFC] children: not-inline
    BlockContainer <body> at (8,8) content-size 1620x0 [8+0+0 1620 0+0+8] [8+0+0 0 0+0+8] children: inline
      InlineNode <span.label> [0+0+0 0 0+0+0] [0+0+0 18 0+0+0]
        BlockContainer <(anonymous)> (not painted) children: inline
          TextNode <#text>
        TextNode <#text>
      TextNode <#text>
```

### Screenshots or screen recordings

In Ladybird: 

![Image](https://github.com/user-attachments/assets/4465f257-5e48-4a64-abf6-73b1833ba8ee)

In Firefox:

![Image](https://github.com/user-attachments/assets/3f41e5cc-130b-456d-9af4-7cf7be419c98)

### Contribute a patch?

- [ ] I’ll contribute a patch for this myself.

Issue #3827, opened 2025-03-05
Incorrect text layout of text of "content" CSS prop set in a :before of a <i> tag


teaalltr:
### Summary

When opening the snippet below with Ladybird, here's what's shown:

![Image](https://github.com/user-attachments/assets/587e24a4-fe01-45c5-96ff-41eda42bc568)

The snippet is a minimal reporoduction, and the <i> tag is apparently needed since replacing it with <div> does not trigger the text overlap.
The position: absolute is also needed

My guess is there's something wrong with that :before behavior when position absolute is set, or something like that.
Similar stuff here: 
https://www.mozilla.org/it/firefox/download/thanks/
but for the :after pseudoelement of the <strong> related to the "download" text (which should show the squiggly line under the text and not at the bottom of the page, cfr with Firefox or Chrome).

### Operating system

Linux

### Steps to reproduce

1. Run the snippet below
2. Note the text overlap in LB buit not in Firefox or Chrome

### Expected behavior

The text should be shown inline, here Firefox for reference:

![Image](https://github.com/user-attachments/assets/8096e3a2-4388-4f28-8b92-0d67fc5cf539)

### Actual behavior

:before's text overlaps with the div's text

### URL for a reduced test case

None

### HTML/SVG/etc. source for a reduced test case

```html
<html>
    <meta>
    <style>
        .fas:before {
            position: absolute;
            content: "bar";
        }
    </style>
    </meta>
    <div>foo <i class="fas"></i>
    </div>
</html>
```

### Log output and (if possible) backtrace

```shell
None
```

### Build flags or config settings

Ladybird dd78d5d50e4b6e640909840d496a4c1e6265e8c6
Linux Kubuntu latest

### Contribute a patch?

- [ ] I’ll contribute a patch for this myself.

gmta:
Shorter reproduction:

```html
<style>
i::before {
    content: "bar";
    position: absolute;
}
</style>
<div>foo<i>
```

Took a short stab at it-

Because of `position: absolute`, blockification of the pseudo element's box takes place which causes the box to be hoisted out of the inline context and into the parent's block element. But because of `::before`, we're still _prepending_ that layout node causing it to appear before "foo" in the layout tree.

Fixing that still isn't enough, since you then end up with two block nodes that each get their own line.

PR #6739, opened 2025-11-07
LibWeb: Fixes for pseudo elements that are absolutely positioned or `display: block`

gmta:
Fixes the discount price strikethrough on https://tweakers.net/pricewatch:

| Before | After |
|--|--|
| image | image |

### DIFF
--- a/Libraries/LibWeb/Layout/TreeBuilder.cpp
+++ b/Libraries/LibWeb/Layout/TreeBuilder.cpp
@@ -113,8 +113,8 @@ static Layout::Node& insertion_parent_for_block_node(Layout::NodeWithStyle& layo
     if (!has_inline_or_in_flow_block_children(*new_parent))
         return *new_parent;
 
-    // If the block is out-of-flow,
-    if (layout_node.is_out_of_flow()) {
+    // If the block is out-of-flow and is not a pseudo element,
+    if (layout_node.is_out_of_flow() && !layout_node.is_generated_for_pseudo_element()) {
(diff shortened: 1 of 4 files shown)


Libraries/LibWeb/Layout/FormattingContext.cpp
```diff
@@ -1201,12 +1201,6 @@ void FormattingContext::layout_absolutely_positioned_element(Box const& box, Ava
     }
 
     auto& containing_block_state = m_state.get_mutable(*box.containing_block());
-
```
lpas, line 1204: I think this was an optimization in 221650f3cdd3b9254ed55daab693f1fa813873a7

gmta: You're right! Easily proven by verifying that both bools are `true` at that point in all cases, which they are not. Dropped that commit.

gmta:
@lpas 

> this fixes https://github.com/LadybirdBrowser/ladybird/issues/5042

Thanks, added it to the commit message.

> but would not fix https://github.com/LadybirdBrowser/ladybird/issues/3827

Interesting, however that is an _inline_ absolutely positioned box, and although this PR's description hints at all absolutely positioned pseudo elements to be fixed, I only dealt with block elements. I'll take a look and see if this is easily fixed as well.

gmta:
Looked at #3827 but determined its cause is a bit more involved than what was going on here. That's for a future PR.

PR #6741, opened 2025-11-07
Prettify about:about and about:processes

LeastSaneDeveloper:
Extended CSS and HTML in about.html and processes.html.

### DIFF
--- a/Base/res/ladybird/about-pages/processes.html
+++ b/Base/res/ladybird/about-pages/processes.html
@@ -1,181 +1,180 @@
-<!doctype html>
-<html>
-    <head>
-        <title>Task Manager</title>
-        <style>
-            @media (prefers-color-scheme: dark) {
-                :root {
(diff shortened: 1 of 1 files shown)


InvalidUsernameException:
Hi, welcome to the project.

The bot message is asking you to change the commit messages, not the PR title. In particular:
* You need a prefix on each commit. This is generally the directory or library you are changing in that commit. In this case `Base: ...` would be appropriate.
* There should be no merge commits in a PR, so you need to drop that one.
* Each commit should be atomic and fully correct. It look like the second and fourth commit don't fit that description and should be combined.

For those you will need to use `git rebase`. If you are not familiar with reabsing, [this video](https://youtu.be/ElRzTuYln0M) might help you.

LeastSaneDeveloper:
Ah! Silly me! I'll fix that later, thanks.

PR #6744, opened 2025-11-07
LibWeb/WebGL: A bunch more testsuite fixes

cqundefine:
Just fixing a few more failures I found in the WebGL testsuite.

### DIFF
--- a/Libraries/LibWeb/WebGL/WebGLRenderingContextImpl.cpp
+++ b/Libraries/LibWeb/WebGL/WebGLRenderingContextImpl.cpp
@@ -479,6 +479,8 @@ void WebGLRenderingContextImpl::delete_program(GC::Root<WebGLProgram> program)
         program_handle = handle_or_error.release_value();
     }
     glDeleteProgram(program_handle);
+    if (m_current_program == program)
+        m_current_program = nullptr;
 }
 
 void WebGLRenderingContextImpl::delete_renderbuffer(GC::Root<WebGLRenderbuffer> renderbuffer)


Libraries/LibWeb/WebGL/WebGL2RenderingContextImpl.cpp
```diff
@@ -964,6 +965,18 @@ void WebGL2RenderingContextImpl::bind_buffer_base(WebIDL::UnsignedLong target, W
             return;
         }
         buffer_handle = handle_or_error.release_value();
+
+        // WebGL 2.0 Spec (Section 5.1)
```
gmta: Maybe just link to the online spec? E.g. https://registry.khronos.org/webgl/specs/latest/2.0/#5.1

Libraries/LibWeb/WebGL/WebGL2RenderingContextImpl.cpp
gmta, line 1: Commit message typo: "it's" -> "its"

Libraries/LibWeb/WebGL/WebGLBuffer.h
```diff
@@ -22,10 +23,16 @@ class WebGLBuffer final : public WebGLObject {
...
+    void mark_as_index_buffer(bool value) { m_is_index_buffer = value ? TriState::True : TriState::False; }
+    TriState is_index_buffer() const { return m_is_index_buffer; }
+
 protected:
     explicit WebGLBuffer(JS::Realm&, WebGLRenderingContextBase&, GLuint handle);
 
     virtual void initialize(JS::Realm&) override;
+
+private:
+    TriState m_is_index_buffer { TriState::Unknown };
```
gmta: Instead of using this tristate, maybe we could add `Optional<GLenum> m_target` and compare that? It's a useful value to have for future debugging exercises as well.

Libraries/LibWeb/WebGL/WebGLUniformLocation.h
```diff
@@ -17,18 +19,19 @@ class WebGLUniformLocation final : public Bindings::PlatformObject {
...
+    ErrorOr<GLuint> handle(GC::Root<WebGLProgram> current_shader) const;
 
 protected:
-    explicit WebGLUniformLocation(JS::Realm&, GLuint handle);
+    explicit WebGLUniformLocation(JS::Realm&, GLuint handle, GC::Root<WebGLProgram> parent_shader);
 
     virtual void initialize(JS::Realm&) override;
 
     GLuint m_handle { 0 };
+    GC::Root<WebGLProgram> m_parent_shader;
```
gmta: Why are you using `GC::Root` everywhere instead of just `GC::Ref` and visit that in a `::visit_edges()`?

cqundefine:
> `location_handle` computation is very repetitive, could it be extracted to some (inline?) function instead?

Looking at how it's already pretty minimal, different between some functions and the fact that a similar pattern is already present in a lot of places I'm not sure there's any easy way to factor it out.

Libraries/LibWeb/WebGL/WebGLBuffer.cpp
```diff
@@ -33,4 +35,15 @@ void WebGLBuffer::initialize(JS::Realm& realm)
...
 
+bool WebGLBuffer::is_compatible_with(GLenum target)
+{
+    // https://registry.khronos.org/webgl/specs/latest/2.0/#5.1
+    if (!m_target.has_value()) {
+        m_target = target;
+        return true;
+    }
+
+    return (target == GL_ELEMENT_ARRAY_BUFFER && m_target.value() == GL_ELEMENT_ARRAY_BUFFER) || (target != GL_ELEMENT_ARRAY_BUFFER && m_target.value() != GL_ELEMENT_ARRAY_BUFFER);
```
gmta: I feel this would be easier to read as:

```cpp
if (target === GL_ELEMENT_ARRAY_BUFFER)
    return m_target.value() == GL_ELEMENT_ARRAY_BUFFER;

return m_target.value() != GL_ELEMENT_ARRAY_BUFFER;
```

PR #6748, opened 2025-11-07
Base: Prettify about:about and about:processes

LeastSaneDeveloper:
# about:about

**Before:**

Screenshot 2025-11-07 at 14 22 46

**After:**

Screenshot 2025-11-07 at 14 23 46

# about:processes

**Before:**

Screenshot 2025-11-07 at 14 23 07

**After:**

Screenshot 2025-11-07 at 14 23 58

### DIFF
--- a/Libraries/LibRegex/RegexOptimizer.cpp
+++ b/Libraries/LibRegex/RegexOptimizer.cpp
@@ -1436,6 +1436,8 @@ void Optimizer::append_alternation(ByteCode& target, Span<ByteCode> alternatives
     struct QualifiedIP {
         size_t alternative_index;
         size_t instruction_position;
+
+        bool operator==(QualifiedIP const& other) const = default;
     };
     struct NodeMetadataEntry {
         QualifiedIP ip;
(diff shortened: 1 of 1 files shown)


Hendiadyoin1:
BTW you do not need to open a new PR to fix the commit names
the bot gives a hint at how to fix it, and otherwise this video <https://youtu.be/ElRzTuYln0M>
is recommended as an introduction to git

Hendiadyoin1:
Also in cases where changes are purely aesthetical, providing a before and after in the PR description is usually appreciated

LeastSaneDeveloper:
Hello,
Yeah, I kinda screwed up the branches before (like...a lot), not just with the names, so I thought might as well start from scratch. I'll put screenshots, thanks.

Zaggy1024:
If it's intended that the indentation was changed in the modified files, it would be nice to separate that into another commit, as GitHub seems to have a lot of trouble telling where new code was inserted. :sweat_smile: 

My usual way of fixing things like that up is to interactive rebase, break before the commit, make the change, then when (probably) finds a conflict, just completely overwrite the file with the incoming changes.

LeastSaneDeveloper:
Crap, that was my extension that did that. Will do that after dinner.

Gingeh:
ooh I love the new designs

LeastSaneDeveloper:
Thank you :) Any suggestions would be helpful!

Base/res/ladybird/about-pages/about.html
```diff
@@ -1,18 +1,103 @@
 <!DOCTYPE html>
-<html lang="en">
+<html>
```
Zaggy1024: It seems like this tag shouldn't be removed, the default value is an empty string according to MDN: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/lang

LeastSaneDeveloper:
Got it, thanks.

PR #6749, opened 2025-11-07
A couple select fixes

lpas:
should fix ~10 WPTs

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLOptionsCollection.cpp
+++ b/Libraries/LibWeb/HTML/HTMLOptionsCollection.cpp
@@ -98,22 +98,22 @@ WebIDL::ExceptionOr<void> HTMLOptionsCollection::set_value_of_indexed_property(u
 
     auto root_element = root();
 
-    if (index >= length) {
-        // 3. Let delta be index minus length
-        auto delta = index - length;
+    // 3. Let delta be index minus length
+    auto delta = static_cast<i64>(index) - static_cast<i64>(length);
 
-        // 4. If delta is greater than zero, then append a DocumentFragment consisting of delta new option elements with
-        //    no attributes and no child nodes to the select element on which the HTMLOptionsCollection is rooted.
-        if (delta > 0) {
-            for (WebIDL::UnsignedLong i = 0; i < delta; i++) {
-                TRY(root_element->append_child(TRY(DOM::create_element(root_element->document(), HTML::TagNames::option, Namespace::HTML))));
-            }
+    // 4. If delta is greater than zero, then append a DocumentFragment consisting of delta new option elements with
+    //    no attributes and no child nodes to the select element on which the HTMLOptionsCollection is rooted.
+    if (delta > 0) {
+        for (auto i = 0; i < delta; i++) {
+            TRY(root_element->append_child(TRY(DOM::create_element(root_element->document(), HTML::TagNames::option, Namespace::HTML))));
         }
+    }
 
-        // 5. If delta is greater than or equal to zero, append value to the select element.
+    // 5. If delta is greater than or equal to zero, append value to the select element. Otherwise, replace the indexth
+    //    element in the collection by value.
+    if (delta >= 0) {
         TRY(root_element->append_child(option));
     } else {
-        // 5 (cont). Otherwise, replace the indexth element in the collection by value.
         TRY(root_element->replace_child(option, *item(index)));
     }
 


Libraries/LibWeb/HTML/HTMLSelectElement.cpp
```diff
@@ -170,6 +196,17 @@ WebIDL::ExceptionOr<void> HTMLSelectElement::add(HTMLOptionOrOptGroupElement ele
     return {};
 }
 
+WebIDL::ExceptionOr<void> HTMLSelectElement::set_value_of_indexed_property(u32 n, JS::Value new_value)
+{
+    // When the user agent is to set the value of a new indexed property or set the value of an existing indexed property
+    // for a select element, it must instead run the corresponding algorithm on the select element's options collection.
+    TRY(const_cast<HTMLOptionsCollection&>(*options()).set_value_of_indexed_property(n, new_value));
+
+    update_selectedness(); // Not in spec
```
Psychpsyo: Should probably have a `// AD-HOC: Reason for why it's needed`, and maybe a spec bug filed.

Libraries/LibWeb/HTML/HTMLSelectElement.cpp
```diff
@@ -152,6 +171,13 @@ HTMLOptionElement* HTMLSelectElement::item(WebIDL::UnsignedLong index)
     return as<HTMLOptionElement>(const_cast<HTMLOptionsCollection&>(*options()).item(index));
 }
 
+Optional<JS::Value> HTMLSelectElement::item_value(size_t index) const
```
Psychpsyo, line 162: Spec links on the new functions would be nice.

Libraries/LibWeb/HTML/HTMLSelectElement.cpp
lpas: When I was writing this part I took this part from the method above. I checked and don't seem to need this thus removed it from my code.

Libraries/LibWeb/HTML/HTMLOptionsCollection.cpp
```diff
@@ -104,7 +104,7 @@ WebIDL::ExceptionOr<void> HTMLOptionsCollection::set_value_of_indexed_property(u
 
         // 4. If n is greater than zero, then append a DocumentFragment consisting of n-1 new option elements with no attributes and no child nodes to the select element on which the HTMLOptionsCollection is rooted.
         if (n > 0) {
-            for (WebIDL::UnsignedLong i = 0; i < n - 1; i++) {
+            for (WebIDL::UnsignedLong i = 0; i <= n - 1; i++) {
```
gmta: This does not match the spec. For example, if `n == 1`, we should append `n - 1 -> 0` new option elements. You're now appending a single option element instead.

Libraries/LibWeb/HTML/HTMLSelectElement.h
```diff
@@ -142,7 +144,7 @@ class WEB_API HTMLSelectElement final
     mutable Vector<GC::Ref<HTMLOptionElement>> m_cached_list_of_options;
     mutable size_t m_cached_number_of_selected_options { 0 };
 
-    GC::Ptr<HTMLOptionsCollection> m_options;
+    GC::Ptr<HTMLOptionsCollection> mutable m_options;
```
gmta: We put `mutable` in front of the type, not after it. (See the lines above this one)

Libraries/LibWeb/HTML/HTMLOptionElement.cpp
```diff
@@ -86,6 +86,9 @@ void HTMLOptionElement::set_selected_internal(bool selected)
     m_selected = selected;
     if (selected)
         m_selectedness_update_index = m_next_selectedness_update_index++;
+
+    // AD-HOC this is here to break the cache on the HTMLCollection in HTMLSelectElement::selected_options
```
gmta: You don't need to tag your comments with `AD-HOC` if the implementation isn't following any particular specification.

Libraries/LibWeb/HTML/HTMLOptionsCollection.cpp
lpas: You are right I guess this is a spec bug I created an issue https://github.com/whatwg/html/issues/11905

Libraries/LibWeb/HTML/HTMLOptionElement.cpp
```diff
@@ -86,6 +86,9 @@ void HTMLOptionElement::set_selected_internal(bool selected)
     m_selected = selected;
     if (selected)
         m_selectedness_update_index = m_next_selectedness_update_index++;
+
+    // this is here to break the cache on the HTMLCollection in HTMLSelectElement::selected_options
```
konradekk: `purge` or something similar would sound better than `break`, i believe.

Libraries/LibWeb/HTML/HTMLOptionsCollection.cpp
lpas: The spec change should resolve this.

Libraries/LibWeb/HTML/HTMLOptionElement.cpp
lpas: Mh, not sure this doesn't really purge the cache it just breaks the cache condition

gmta: The usual terminology for that is "invalidate" though!

lpas: done

Libraries/LibWeb/HTML/HTMLOptionsCollection.cpp
konradekk: literally this would be something like the following, right?

```suggestion
    // 3. Let delta be index minus length
    auto delta = index - length;

    // 4. If delta is greater than zero, then append a DocumentFragment consisting of delta new option elements with
    //    no attributes and no child nodes to the select element on which the HTMLOptionsCollection is rooted.
    if (delta > 0) {
        for (WebIDL::UnsignedLong i = 0; i < delta; i++) {
            TRY(root_element->append_child(TRY(DOM::create_element(root_element->document(), HTML::TagNames::option, Namespace::HTML))));
        }
    }

    // 5. If delta is greater than or equal to zero, append value to the select element.
    //    Otherwise, replace the indexth element in the collection by value.
    if (delta >= 0) {
        TRY(root_element->append_child(option));
    } else {
        TRY(root_element->replace_child(option, *item(index)));
    }
```

lpas: From a spec point yes but index & length are unsinged.

gmta: @lpas there is this condition surrounding this code block:

```cpp
if (index >= length) {
```

And that's not in the spec - it does guarantee that what you do is correct, but we're not following the spec as close as we could be. If the spec changes in the future, this _might_ cause issues (ymmv).

lpas: Mh, maybe I miss understood the question/comment 

My understanding of this question was: Why isn't this code more literal to the spec. And my understanding was that the difference is here because index & length are unsinged. 

I just wanted to do a bug fix here to remove a `-1`. This triggered a spec change and thus I updated the text here, Am I missing something I should have done differently?

gmta: Even though we've chosen to make `index` and `length` unsigned, `delta` does not necessarily need to be unsigned. We don't get an underflow because of the guard around the whole code block. Note that the spec does not ask us to make anything unsigned, so we can't assume that the spec text is written with that type in mind.

Imo, the best thing is to keep to the spec as closely as possible (within reason) so the code becomes very easily verifiable by other people.

Concretely:
* Drop the `if (index >= length) {` since that's not in the spec
* Rework the steps so they more closely match the spec text

lpas: I feel like I'm running in circles here. And I'm not sure that anything needs to be changed here (feels like a nice solution for me with the index >=length part) for this PR to provide value. This part of the algorithm is already in the codebase and was not written by me.

gmta: You're not running in circles? We're asking you to implement things closer to the spec, which will also accommodate future changes to this implementation. That there is code that already exists, and that it wasn't written by you doesn't mean you can't change it?

lpas: Maybe there is a nicer solution for this. I don't know. 

I think the solution posted up top would not work, is because of the unsinged int variables. That was my answer and then this got into a kinda meta discussion here.  I'm really not sure what this discussion is about.

konradekk: > This part of the algorithm is already in the codebase and was not written by me.

missed this, no issues here, excuses on my side! 🙇🏻‍♂️

> I think the solution posted up top would not work, is because of the unsinged int variables.

a cast would probably solve the issue, but given the above, i guess that’s a thing for a different change (up to the maintainers whether they want to pursue this in this here or on different occassion).

lpas: @gmta coming back to this PR

I could change this to something like this would that be ok? 

```suggestion
    // 3. Let delta be index minus length
    auto delta = index - (i64)length;

    // 4. If delta is greater than zero, then append a DocumentFragment consisting of delta new option elements with
    //    no attributes and no child nodes to the select element on which the HTMLOptionsCollection is rooted.
    if (delta > 0) {
        for (auto i = 0; i < delta; i++) {
            TRY(root_element->append_child(TRY(DOM::create_element(root_element->document(), HTML::TagNames::option, Namespace::HTML))));
        }
    }
    if (delta >= 0) {
        // 5. If delta is greater than or equal to zero, append value to the select element.
        //    Otherwise, replace the indexth element in the collection by value.
        TRY(root_element->append_child(option));
    } else {
        TRY(root_element->replace_child(option, *item(index)));
    }

```

shannonbooth: Just giving my opinion, but yes, something along those lines would be an improvement. Though with some nits of:

- Normally we would place step 5 comment above the `if (delta >= 0) {` if statement instead of inside it since it encompasses both branches
- I think it would be more clear if we `static_cast<i64>` both index and length as to not need to think about promotion rules

PR #6750, opened 2025-11-08
AK: Add a formatter for Duration

Zaggy1024:
This will format Duration as seconds, with as much decimal precision as necessary to fully represent its value. The alternate format specifier can be used to make it print the units on the end, i.e. "1.23s".


gmta:
Test is failing though!

PR #6760, opened 2025-11-08
LibWeb: Calculate y offset properly for all cases

### LINKED ISSUE #6333: YouTube Music has overlapping menu icon and logo
### Summary

When going to Youtube Music's home page, one can notice in the top left corner that the Youtube Music logo and hamburger menu icon are overlapping with the Home button that is at the top of  the left sidebar.

### Operating system

Linux

### Steps to reproduce

1. Go to music.youtube.com

### Expected behavior

The logo and hamburger menu don't overlap.

### Actual behavior

They do overlap

https://music.youtube.com/

```html
<!DOCTYPE html>
<html>

<head>
    <style>
        #mini-guide {
            position: fixed;
        }

        #mini-guide-spacer {
            margin-top: 64px;
        }
    </style>
</head>

<body>
    <div>Guide</div>
    <div id="mini-guide">
        <div id="mini-guide-spacer"></div>
        <div>Home</div>
    </div>
</body>

</html>
```

```shell
N/A
```

<p>Expected output (from Firefox)</p>
Image

<p>Ladybird Output</p>

Image

<p>Firefox reduced test case output</p>

Image

<p>Ladybird reduced test case output</p>

Image

_No response_

### Contribute a patch?

gotlougit:
Fixes #6333

### DIFF
--- a/Libraries/LibWeb/Layout/BlockFormattingContext.cpp
+++ b/Libraries/LibWeb/Layout/BlockFormattingContext.cpp
@@ -874,7 +874,7 @@ void BlockFormattingContext::layout_block_level_box(Box const& box, BlockContain
         resolve_used_height_if_treated_as_auto(box, available_space_for_height_resolution, independent_formatting_context);
     }
 
-    if (independent_formatting_context || !margins_collapse_through(box, m_state)) {
+    if (independent_formatting_context || !margins_collapse_through(box, m_state) || block_container.is_out_of_flow()) {
         if (!m_margin_state.box_last_in_flow_child_margin_bottom_collapsed()) {
             m_margin_state.reset();
         }
--- a/Tests/LibWeb/Layout/input/position-fixed-empty-element-with-margin.html
+++ b/Tests/LibWeb/Layout/input/position-fixed-empty-element-with-margin.html
@@ -1,0 +1,18 @@
+<!DOCTYPE html>
+<head>
+<style>
+    #container {
+        position: fixed;
+    }
+    #spacer {
+       margin-top: 64px;
+    }
+</style>
+</head>
+<body>
+    <div>Heading</div>
+    <div id="container">
+        <div id="spacer"></div>
+        <div>Home</div>
+    </div>
+</body>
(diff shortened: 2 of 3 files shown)


Libraries/LibWeb/Layout/BlockFormattingContext.cpp
```diff
@@ -858,6 +858,8 @@ void BlockFormattingContext::layout_block_level_box(Box const& box, BlockContain
             m_margin_state.reset();
         }
         m_y_offset_of_current_block_container = box_state.offset.y() + box_state.content_height() + box_state.border_box_bottom();
+    } else if (creates_block_formatting_context(block_container) && block_container.is_absolutely_positioned()) {
+        m_y_offset_of_current_block_container = box_state.offset.y() + box_state.content_height() + box_state.border_box_bottom();
```
gmta: All tests pass for me as well when I add ` || block_container.is_out_of_flow()` to the existing `if` condition above. Are you sure the margin state should not be reset?

Either I missed something that your test does not include, or this could be a bit simpler :)

gotlougit: For me with this code somehow I am getting one failed test `Text/input/GamepadAPI/gamepad-iframe.html`. I had earlier tried to just add the check for absolute positioning in the existing if, but because of this failure I went with the solution in this PR. Since the gamepad api should really not be affected by this change, I am inclined to believe it is just flakiness on local setup.

gotlougit: Huh, there's a different test that fails with this code somehow, and it seemingly also has no relation with this PR: https://github.com/LadybirdBrowser/ladybird/actions/runs/19489255885/job/55778129669?pr=6760#step:15:17192

gmta: We have a couple of flaky tests - make sure to rebase on `master`, because `gamepad-iframe.html` should no longer give you issues there!

gotlougit: Yep this worked, thanks!

Tests/LibWeb/Layout/input/position-fixed-empty-element-with-margin.html
```diff
@@ -0,0 +1,19 @@
...
+        }
+</style>
+</head>
+<body>
+    <div>Heading</div>
+    <div id="container">
+        <div id="spacer"></div>
+        <div>Home</div>
+    </div>
+</body>
```
kalenikaliaksandr, line 18: please, use consistent 4 space indentation in this file.

gotlougit:
Seems like there is some new CI flakiness that is not resolved by rebasing on master, I will check again in few days.

PR #6768, opened 2025-11-09
LibCore: Introduce `REPORT_TIME()` for per call timing information

tcl3:
This macro outputs the time taken to reach the end of the current scope to the debug console. It also shows the average time, total cumulative time and the total number of calls.

It is also possible to limit the amount of debug output using `REPORT_TIME_EVERY(name, n)` to limit output to every `n` calls.

For example adding a measure time macro like this:

```c++
#include <LibCore/ReportTime.h>

void Foo::bar()
{
   REPORT_TIME(my_test_function)
   // code I want to measure
}
```
Will give debug output in this format for every call of the function `Foo::bar()`:
```
 [Timing info for: my_test_function] current: 23.235μs | average: 23.235μs | total: 23.235μs  | calls: 1
...
 [Timing info for: my_test_function] current: 38.454μs | average: 59.303μs | total: 209.931ms | calls: 3540
```

### DIFF
--- a/Libraries/LibCore/ReportTime.cpp
+++ b/Libraries/LibCore/ReportTime.cpp
@@ -1,0 +1,34 @@
+/*
+ * Copyright (c) 2025, Tim Ledbetter <tim.ledbetter@ladybird.org>
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include "ReportTime.h"
+
+#include <AK/Format.h>
+#include <AK/HashMap.h>
+#include <AK/NumberFormat.h>
+
+namespace Core {
+
+static HashMap<ByteString, TimingInfo> g_timing_info_table;
+
+void log_timing_info(ByteString const& name, AK::Duration const& elapsed_time, u64 print_every_n_calls)
+{
+    auto& timing_info = g_timing_info_table.ensure(name);
+    timing_info.call_count++;
+    timing_info.cumulative_time_nanoseconds += elapsed_time.to_nanoseconds();
+
+    u64 average_nanoseconds = timing_info.cumulative_time_nanoseconds / timing_info.call_count;
+    if (timing_info.call_count % print_every_n_calls != 0)
+        return;
+    dbgln("[Timing info for: {}] current: {:10} | average: {:10} | total: {:10} | calls: {}",
+        name,
+        human_readable_short_time(elapsed_time),
+        human_readable_short_time(AK::Duration::from_nanoseconds(average_nanoseconds)),
+        human_readable_short_time(AK::Duration::from_nanoseconds(timing_info.cumulative_time_nanoseconds)),
+        timing_info.call_count);
+}
+
+}
--- a/Libraries/LibCore/CMakeLists.txt
+++ b/Libraries/LibCore/CMakeLists.txt
@@ -47,6 +47,7 @@ set(SOURCES
     MappedFile.cpp
     MimeData.cpp
     Notifier.cpp
+    ReportTime.cpp
     Resource.cpp
     ResourceImplementation.cpp
     ResourceImplementationFile.cpp
--- a/Libraries/LibCore/ReportTime.h
+++ b/Libraries/LibCore/ReportTime.h
@@ -1,0 +1,35 @@
+/*
+ * Copyright (c) 2025, Tim Ledbetter <tim.ledbetter@ladybird.org>
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#pragma once
+
+#include <AK/ScopeGuard.h>
+#include <AK/Time.h>
+#include <LibCore/ElapsedTimer.h>
+
+namespace Core {
+
+struct TimingInfo {
+    u64 cumulative_time_nanoseconds { 0 };
+    u64 call_count { 0 };
+};
+
+void log_timing_info(ByteString const& name, AK::Duration const& elapsed_time, u64 print_every_n_calls);
+
+#define CONCAT(a, b) a##b
+#define UNIQUE_NAME(base) CONCAT(base, __LINE__)
+
+#define REPORT_TIME_EVERY(name, print_every_n_calls)                                                \
+    auto UNIQUE_NAME(report_time_timer_) = Core::ElapsedTimer::start_new(Core::TimerType::Precise); \
+    ScopeGuard UNIQUE_NAME(report_time_guard_) = [&] {                                              \
+        auto elapsed_time = UNIQUE_NAME(report_time_timer_).elapsed_time();                         \
+        Core::log_timing_info(#name, elapsed_time, print_every_n_calls);                            \
+    };
+
+#define REPORT_TIME(name) \
+    REPORT_TIME_EVERY(name, 1)
+
+}


Libraries/LibCore/MeasureTime.h
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright (c) 2025, Tim Ledbetter <tim.ledbetter@ladybird.org>
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#pragma once
+
+#include <AK/Format.h>
+#include <AK/HashMap.h>
```
gmta: Looks like these two includes can go in the .cpp instead

Libraries/LibCore/MeasureTime.cpp
```diff
@@ -0,0 +1,54 @@
...
+}
+
+void log_timing_info(ByteString const& name, u64 elapsed_nanoseconds, u64 print_every_n_calls)
+{
+    auto timing_info = g_timing_info_table.ensure(name);
+    auto previous_call_count = timing_info.call_count;
+    timing_info.call_count++;
+    timing_info.cumulative_time_nanoseconds += elapsed_nanoseconds;
+    u64 average_nanoseconds = timing_info.cumulative_time_nanoseconds / timing_info.call_count;
+    g_timing_info_table.set(name, timing_info);
```
gmta: Hm, can't we take a reference instead? (`auto& timing_info =` above)

Libraries/LibCore/MeasureTime.cpp
```diff
@@ -0,0 +1,54 @@
...
+    if (previous_call_count % print_every_n_calls != 0)
```
gmta: By taking the previous call count, doesn't this mean that for `print_every_n_calls == 2`, we print for the first time when the third call has finished, not the second one?

Libraries/LibCore/MeasureTime.cpp
```diff
@@ -0,0 +1,54 @@
...
+    } else if (value >= nanosecond_in_millisecond) {
+        value = nanoseconds / nanosecond_in_millisecond;
+        unit = "ms"sv;
+    } else if (value >= nanoseconds_in_microsecond) {
+        value = nanoseconds / nanoseconds_in_microsecond;
+        unit = "μs"sv;
+    }
+
+    return ByteString::formatted("{:.3}{}", value, unit);
+}
```
gmta: This might be a good expansion for `AK::human_readable_time`, then you can use the `Duration` directly as well.

Libraries/LibCore/MeasureTime.cpp
```diff
@@ -0,0 +1,54 @@
...
+ * Copyright (c) 2025, Tim Ledbetter <tim.ledbetter@ladybird.org>
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include "MeasureTime.h"
+
+namespace Core {
+
+HashMap<ByteString, TimingInfo> g_timing_info_table;
```
Hendiadyoin1: couldn't this be replaced by function static instances?
init would be slightly longer line-wise though*

or is there a plan to expose this for a togglable at-exit print of the stats?

*:
```cpp
static auto __timing_stats_##name = TimingInfo(name, print_every_n);
auto measure_time_timer_##name = Core::ElapsedTimer::start_new(Core::TimerType::Precise);
ScopeGuard measure_time_guard_##name = [&] { 
    auto elapsed_nanoseconds = measure_time_timer_##name.elapsed_time().to_nanoseconds();
    __timing_stats_##name.add_run(elapsed_nanoseconds);
};
```

Hendiadyoin1: also the `##name` could be replaced by `##__COUNTER__` but its likely not thaat useful to have timers of the same name in the same scope

tcl3: > couldn't this be replaced by function static instances?

The reason I used a global like this is so that I could use `MEASURE_TIME()` with the same name in multiple places and it would update the same stats.

> also the ##name could be replaced by ##__COUNTER__

I wasn't aware this existed. I've changed it to use this, as this allows a `name` that may not be a valid variable name.

Libraries/LibCore/MeasureTime.cpp
tcl3: Not sure what I was thinking there :sweat_smile:

Libraries/LibCore/MeasureTime.cpp
```diff
@@ -0,0 +1,57 @@
...
+}
+
+void log_timing_info(ByteString const& name, u64 elapsed_nanoseconds, u64 print_every_n_calls)
+{
+    auto& timing_info = g_timing_info_table.ensure(name);
+    timing_info.call_count++;
+    timing_info.cumulative_time_nanoseconds += elapsed_nanoseconds;
+    u64 average_nanoseconds = timing_info.cumulative_time_nanoseconds / timing_info.call_count;
+    g_timing_info_table.set(name, timing_info);
+    if (timing_info.call_count % print_every_n_calls != 0 && timing_info.call_count != 1)
```
gmta: I don't think this is right either - for `print_every_n_calls == 3`, you print when `call_count == 1 `?

Why isn't this just `if (timing_info.call_count % print_every_n_calls != 0) return;`?

tcl3: I was trying to avoid losing the first call. This makes more sense though. My latest update does this instead.

Libraries/LibCore/MeasureTime.cpp
tcl3: I've added `AK::human_readable_short_time()` and now use that to do time formatting. I still store nanoseconds in the HashMap otherwise we end up doing lots of unnecessary conversions.

tcl3:
My latest push also changes the name of the macro to `REPORT_TIME()`, as this name makes it clearer what the intention of `REPORT_TIME_EVERY()` is.

Libraries/LibCore/MeasureTime.cpp
tcl3: > ##name could be replaced by ##__COUNTER__

Actually, I couldn't get `__COUNTER__` to work correctly. It doesn't seem like there's an easy way to refer to the same variable in multiple places. My latest change uses `__LINE__` instead.

Libraries/LibCore/ReportTime.cpp
```diff
@@ -0,0 +1,35 @@
...
+static HashMap<ByteString, TimingInfo> g_timing_info_table;
+
+void log_timing_info(ByteString const& name, AK::Duration const& elapsed_time, u64 print_every_n_calls)
+{
+    auto& timing_info = g_timing_info_table.ensure(name);
+    timing_info.call_count++;
+    timing_info.cumulative_time_nanoseconds += elapsed_time.to_nanoseconds();
+
+    u64 average_nanoseconds = timing_info.cumulative_time_nanoseconds / timing_info.call_count;
+    g_timing_info_table.set(name, timing_info);
```
gmta: You're modifying the reference (`auto& timing_info`), no need to `.set()` it again.

gmta:
Fun fact: our padding (`"{:10}"`) breaks when you use multibyte codepoints such as "μ". 😅

tcl3:
> Fun fact: our padding (`"{:10}"`) breaks when you use multibyte codepoints such as "μ". 😅

Yeah, I noticed that. Not sure if there's an easy workaround that doesn't involve rewriting a bunch of foramtting code.

Issue #6767, opened 2025-11-09
Support SVG fragments in HTML


dotnetCarpenter:
### Summary

Support SVG fragments: https://caniuse.com/svg-fragment

Here is an article discussing the technique: https://simurai.com/blog/2012/04/02/svg-stacks

Test-case: https://dahlström.net/svg/structure/svgstack-usage-object.html

Ladybird does support SVG fragments, only not in HTML: https://xn--dahlstrm-t4a.net/svg/structure/resources/iconic-stack.svg#star and https://xn--dahlstrm-t4a.net/svg/structure/resources/iconic-stack.svg#book

### Operating system

Linux

### Steps to reproduce

1. Open Ladybird and go to https://dahlström.net/svg/structure/svgstack-usage-object.html
2. Observe that no SVG icons are rendered.
3. Go to https://xn--dahlstrm-t4a.net/svg/structure/resources/iconic-stack.svg#book and observe that Ladybird is able to show the "book" SVG fragment.

### Expected behavior

<img width="1233" height="287" alt="Image" src="https://github.com/user-attachments/assets/3571d5fa-31e5-4747-9fe5-67ccddb2bf01" />

_All 5 SVG fragments are rendered._

### Actual behavior

<img width="1473" height="270" alt="Image" src="https://github.com/user-attachments/assets/e1d18081-e2c2-43ff-aa3e-bf4dba472878" />

_None of the 5 SVG fragments are rendered._

But if you point Ladybird directly to the SVG with the URI fragment, it works:

<img width="950" height="420" alt="Image" src="https://github.com/user-attachments/assets/c2332395-32d2-4cda-a0de-80f3de17fbb2" />

### URL for a reduced test case

https://dahlström.net/svg/structure/svgstack-usage-object.html

### HTML/SVG/etc. source for a reduced test case

```html
[...]

PR #6770, opened 2025-11-09
LibWeb: Add failed test-case for URI fragment support in HTML/SVG

dotnetCarpenter:
This test-case is for issue #6767. Feel free to work on it!

I have no experience with C++, nor much time to dedicate to the project.

But I hope my humble contribution in the form of a test-case, will help whoever decide to donate their time.

Happy coding :heart:

### DIFF
--- a/Tests/LibWeb/Ref/data/svg-stack.svg
+++ b/Tests/LibWeb/Ref/data/svg-stack.svg
@@ -1,0 +1,71 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<svg id="icon" class="icon" version="1.1" xmlns="http://www.w3.org/2000/svg"
+	xmlns:xlink="http://www.w3.org/1999/xlink">
+	<desc> This svgstack uses icons from the Iconic iconset, http://somerandomdude.com/work/iconic/. This work is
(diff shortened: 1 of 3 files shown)


trflynn89:
It'd be better to just attach this test case to the issue rather than having it sit in a PR

dotnetCarpenter:
@trflynn89 do you mean in the form of a git patch, as fenced snippets of code in a comment or as files in a zip file?

dotnetCarpenter:
@trflynn89 It this patch what you meant? https://github.com/LadybirdBrowser/ladybird/issues/6767#issuecomment-3508820682

trflynn89:
I would just paste the HTML in separate code blocks, but any of the above is fine.

trflynn89:
@dotnetCarpenter You don't need to keep rebasing this PR (it doesn't even need to be open IMO), now that the repro test case is in the issue.

dotnetCarpenter:
I actually rebase it as I self-reminder to work on it when I get time. You can close it if it’s noisy Den 30. dec. 2025 kl. 20.45 skrev Tim Flynn ***@***.***>:﻿trflynn89 left a comment (LadybirdBrowser/ladybird#6770)
@dotnetCarpenter You don't need to keep rebasing this PR (it doesn't even need to be open IMO), now that the repro test case is in the issue.

Issue #4696, opened 2025-05-11
`<math>` not rendering


Randl:
### Summary

The symbols inside `<math>` do not appear to render.

### Operating system

Linux

### Steps to reproduce

1. Open HTML with math tags

### Expected behavior

Math renders

### Actual behavior

Math doesn't render. https://github.com/LadybirdBrowser/ladybird/issues/2332 seems related, but even usual characters like numbers do not render.

### URL for a reduced test case

https://codepen.io/evgeniizh/pen/MYYzwVV

### HTML/SVG/etc. source for a reduced test case

```html
N/A
```

### Log output and (if possible) backtrace

```shell
none
```

### Contribute a patch?

- [ ] I’ll contribute a patch for this myself.

AtkinsSJ:
Hi! This is expected at the moment. Very little work has gone into supporting MathML. If anyone is interested in working on it, that would be welcome.

Randl:
Should I close it as math tags appear to render now and open separate issues for specific rendering bugs, or keep it open as an umbrella issue?

AtkinsSJ:
I think closing it and then opening specific issues about bugs that are found makes sense.

PR #6772, opened 2025-11-10
LibWeb: Implement `MathML` rendering

bernaferrari:
Fix https://github.com/LadybirdBrowser/ladybird/issues/4696

1. it is my first PR here, so review carefully. I might be unfamiliar with Ladybird codebase/practices.
2. If you prefer, I can split into smaller PRs (seems like most of the logic is self contained, so I could split MathML elements).
3. I didn't include `maction` because it is a bit odd, but I have its implementation (maybe separate PR?). Most browsers don't fully support it, and it even shows as deprecated [here](https://developer.mozilla.org/en-US/docs/Web/MathML/Reference/Element/maction).


lpas:
Hey welcome to Ladybird. :) 

> 1. it is my first PR here, so review carefully. I might be unfamiliar with Ladybird codebase/practices.

We do in general have a lot of links to the spec in the code. I just had a quick look and could not directly see why some things are done the way they are done. 

> 2. If you prefer, I can split into smaller PRs (seems like most of the logic is self contained, so I could split MathML elements).

As a first PR this is a really big one, it maybe easier to reduce the size for your first one.

You can also pull in WPTs see https://github.com/LadybirdBrowser/ladybird/blob/master/Documentation/Testing.md#importing-web-platform-tests

there are a lot MathML tests https://wpt.fyi/results/mathml?label=experimental&label=master&aligned

PR #6773, opened 2025-11-10
LibWeb: A couple CSS performance improvements

Calme1709:
Fixes for a couple of performance bottlenecks I came across when profiling https://en.wikipedia.org/wiki/2023_in_American_television, totaling a ~15-25% reduction in load times for the aforementioned page.


awesomekling:
Very cool :)

PR #6777, opened 2025-11-10
LibWeb: Correctly sort animations returned by getAnimations()

Psychpsyo:
This test sometimes printed the properties in different orders in CI, so this sorts them before printing their values. Order does not matter in this test, because it just wants to make sure that all the durations are correct.

### DIFF
--- a/Libraries/LibWeb/Animations/KeyframeEffect.cpp
+++ b/Libraries/LibWeb/Animations/KeyframeEffect.cpp
@@ -627,11 +627,12 @@ void KeyframeEffect::generate_initial_and_final_frames(RefPtr<KeyFrameSet> keyfr
 // https://www.w3.org/TR/web-animations-1/#animation-composite-order
 int KeyframeEffect::composite_order(GC::Ref<KeyframeEffect> a, GC::Ref<KeyframeEffect> b)
 {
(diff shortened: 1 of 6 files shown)


Tests/LibWeb/Text/input/css/transition-coordination.html
gmta, line 1: Commit title: instead of "fix bla", can you state what you change? E.g. "LibWeb: Sort animations in test by transition property"

Commit message: please auto-wrap at 72 characters.

gmta:
Interestingly enough, it seems like our `Document::get_animations()` does not sort the animations [like the spec asks us to](https://drafts.csswg.org/web-animations-1/#dom-documentorshadowroot-getanimations), but simply keeps extending the vector with all of its child element's animations.

Psychpsyo:
> Interestingly enough, it seems like our `Document::get_animations()` does not sort the animations [like the spec asks us to](https://drafts.csswg.org/web-animations-1/#dom-documentorshadowroot-getanimations), but simply keeps extending the vector with all of its child element's animations.

I've now went ahead and changed this PR to instead implement the correct sorting here. It gives us at least 1 extra WPT subtest pass and I would hope that it also fixes the flake.

PR #6779, opened 2025-11-10
LibWeb: Avoid storing lower case name for `Attr`

tcl3:
This change gets rid of the `lowercase_name` property stored on both `SimpleSelector` and `Attr` by using case-insensitive hashing and equality checks.

### DIFF
--- a/Libraries/LibWeb/CSS/Selector.h
+++ b/Libraries/LibWeb/CSS/Selector.h
@@ -117,7 +117,7 @@ public:
         struct Name {
             Name(FlyString n)
                 : name(move(n))
-                , lowercase_name(name.to_string().to_lowercase().release_value_but_fixme_should_propagate_errors())
+                , lowercase_name(name.to_ascii_lowercase())
             {
             }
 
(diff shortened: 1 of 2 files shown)


awesomekling:
Is this actually faster? FlyString compare is normally O(1) but case insensitive compare is O(n)

tcl3:
> Is this actually faster? FlyString compare is normally O(1) but case insensitive compare is O(n)

The profiler shows a speed-up in selector parsing.

Output from `REPORT_TIME_EVERY(parse_complex_selector, 10000)` when loading https://amazon.com:

Before:
`[Timing info for: parse_complex_selector] current: 420ns  | average: 1.108µs | total: 11.085ms | calls: 10000`

After:
`[Timing info for: parse_complex_selector] current: 160ns  | average: 786ns   | total: 7.863ms  | calls: 10000`

I'm also not seeing a measurable change in selector matching performance.

awesomekling:
> I'm also not seeing a measurable change in selector matching performance.

How do you measure that? 🤔

tcl3:
Both the profiler and `REPORT_TIME()` showed no difference on Linux, however there was a significant performance regression on MacOS, so I've reverted the changes that caused the regression.

PR #6780, opened 2025-11-10
A couple tabindex / focusable changes

lpas:


Libraries/LibWeb/HTML/HTMLElement.cpp
```diff
@@ -147,9 +148,10 @@ void HTMLElement::set_dir(String const& dir)
     set_attribute_value(HTML::AttributeNames::dir, dir);
 }
 
+// https://html.spec.whatwg.org/multipage/interaction.html#tabindex-value
 bool HTMLElement::is_focusable() const
 {
-    return is_editing_host() || get_attribute(HTML::AttributeNames::tabindex).has_value();
+    return is_editing_host() || Web::HTML::parse_integer(get_attribute_value(HTML::AttributeNames::tabindex)).has_value();
```
gmta: We're already in the `Web::HTML` namespace, don't need that explicit prefix here.

Libraries/LibWeb/SVG/SVGElement.cpp
```diff
@@ -117,6 +118,12 @@ static ReadonlySpan<NamedPropertyID> attribute_style_properties()
     return properties;
 }
 
+// https://html.spec.whatwg.org/multipage/interaction.html#tabindex-value
+bool SVGElement::is_focusable() const
+{
+    return Web::HTML::parse_integer(get_attribute_value(HTML::AttributeNames::tabindex)).has_value();
```
gmta: Hm, maybe this could go in `DOM::Element::is_focusable()` instead? Then you'd no longer need this condition in both SVGElement and HTMLElement.

lpas: good point I looked before to add it to HTMLOrSVGElement but DOM::Element is even better as it also includes MathML

Libraries/LibWeb/DOM/Element.cpp
```diff
@@ -4257,6 +4257,12 @@ bool Element::should_indicate_focus() const
     return false;
 }
 
+// https://html.spec.whatwg.org/multipage/interaction.html#tabindex-value
+bool Element::is_focusable() const
+{
+    return HTML::parse_integer(get_attribute_value(HTML::AttributeNames::tabindex)).has_value();
+}
```
gmta, line 4264: Commit title: "LibWeb: LibWeb [..]"

Maybe something like this?

"LibWeb: Allow all elements with tabindex attribute to be focusable"

lpas: Oh, sorry :/ changed the message

PR #6786, opened 2025-11-10
LibMedia: Move overlapping audio block correction to the data provider

Zaggy1024:
This prevents PlaybackManager's seek while enabling an audio track from causing the AudioMixingSink to push audio blocks forward unnecessarily. Previously, the seek would cause the initial block or blocks to repeat from the perspective of AudioMixingSink, so it would think that it needs to shift the first block after the seek forward by a few samples. By moving this to the AudioDataProvider, we can clear the last sample index every time the decoder is flushed, ensuring that the block shifting always makes sense.

By doing this in AudioMixingSink instead of the Decoder implementations, we avoid having to duplicate this shifting logic across multiple implementations.

This also fixes an issue where multiple audio blocks occupying the same timestamp would be skipped while seeking, causing a significant break in audio.

After #6785 is merged, a check for audio block times can be added to the `decode_audio()` testing function.

### DIFF
--- a/Tests/LibMedia/TestMediaCommon.h
+++ b/Tests/LibMedia/TestMediaCommon.h
@@ -90,6 +90,7 @@ static inline void decode_audio(StringView path, u32 sample_rate, u8 channel_cou
     auto time_limit = AK::Duration::from_seconds(1);
     auto start_time = MonotonicTime::now_coarse();
 
+    i64 last_sample = 0;
     size_t sample_count = 0;
 
     while (true) {
@@ -100,6 +101,9 @@ static inline void decode_audio(StringView path, u32 sample_rate, u8 channel_cou
         } else {
             EXPECT_EQ(block.sample_rate(), sample_rate);
             EXPECT_EQ(block.channel_count(), channel_count);
+
+            VERIFY(sample_count == 0 || last_sample <= block.timestamp_in_samples());
+            last_sample = block.timestamp_in_samples() + static_cast<i64>(block.sample_count());
 
             sample_count += block.sample_count();
         }


Zaggy1024:
Added Windows tag due to messing with possibly platform-specific math. It works on my machine at least!

Zaggy1024:
(Again with less __int128!)

AK/Time.cpp
```diff
@@ -67,6 +70,29 @@ Duration Duration::from_timeval(const struct timeval& tv)
...
+    VERIFY(numerator != 0);
+    VERIFY(denominator != 0);
+    auto seconds_checked = [&] {
+        auto result = Checked<i64>(time_units);
+        result.mul(numerator);
+        result.div(denominator);
+        if (time_units < 0)
+            result.sub(1);
+        return result;
+    }();
```
gmta: Not sure why this is an IIFE, why not `auto seconds_checked = Checked<i64>(time_units);` and go from there?

AK/Time.cpp
```diff
@@ -8,9 +8,12 @@
 #include <AK/Checked.h>
 #include <AK/DateConstants.h>
 #include <AK/GenericLexer.h>
+#include <AK/Math.h>
 #include <AK/String.h>
 #include <AK/StringBuilder.h>
 #include <AK/Time.h>
+#include <AK/UFixedBigInt.h>
+#include <AK/UFixedBigIntDivision.h>
```
gmta: Doesn't seem like you use any of these new includes?

Tests/AK/TestTime.cpp
```diff
@@ -939,3 +939,44 @@ TEST_CASE(from_f64_seconds)
 
     EXPECT_DEATH("Converting float NaN seconds", (void)Duration::from_seconds_f64(NAN));
 }
+
+TEST_CASE(time_units)
```
gmta, line 943: Maybe add these tests in the same commits as the implementation, so if we ever need to revert anything all relevant things are grouped together. Also a bit easier to review if you're going commit by commit :)

AK/Time.cpp
Zaggy1024: The main reason was that it felt strange for the intermediate values to be in different units, but otherwise that form is fine.

It does feel weird to me that Checked doesn't declare operators or implicit casts to allow `Checked<i64>(a) * b` to work though. That would remove the need for an intermediate state that isn't representing seconds.

AK/Time.cpp
Zaggy1024: Ah, forgot to push after I removed these.

PR #6788, opened 2025-11-11
LibWeb: Extract `device_pixels_per_css_pixel` getter from loop

tcl3:
This reduces time spent in `StyleComputer::compute_property_values()` by ~10% when loading https://html.spec.whatwg.org/

### DIFF
--- a/Libraries/LibWeb/CSS/StyleComputer.cpp
+++ b/Libraries/LibWeb/CSS/StyleComputer.cpp
@@ -2148,8 +2148,9 @@ void StyleComputer::compute_property_values(ComputedProperties& style, Optional<
         return style.property(property_id);
     };
 
+    auto device_pixels_per_css_pixel = m_document->page().client().device_pixels_per_css_pixel();
     style.for_each_property([&](PropertyID property_id, auto& specified_value) {
-        auto const& computed_value = compute_value_of_property(property_id, specified_value, get_property_specified_value, computation_context, m_document->page().client().device_pixels_per_css_pixel());
+        auto const& computed_value = compute_value_of_property(property_id, specified_value, get_property_specified_value, computation_context, device_pixels_per_css_pixel);
 
         style.set_property_without_modifying_flags(property_id, computed_value);
     });


Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -2148,8 +2148,9 @@ void StyleComputer::compute_property_values(ComputedProperties& style, Optional<
         return style.property(property_id);
     };
 
+    auto device_pixels_per_css_pixel = m_document->page().client().device_pixels_per_css_pixel();
     style.for_each_property([&](PropertyID property_id, auto& specified_value) {
-        auto const& computed_value = compute_value_of_property(property_id, specified_value, get_property_specified_value, computation_context, m_document->page().client().device_pixels_per_css_pixel());
+        auto const& computed_value = compute_value_of_property(property_id, specified_value, get_property_specified_value, computation_context, device_pixels_per_css_pixel);
```
Calme1709, line 2153: Since we now pass the `AbstractElement` (from which we can get the `Document`) as part of the `ComputationContext` we don't actually need to pass this at all and can instead just compute it within `compute_border_or_outline_width` when required.

I haven't profiled this but I can't imagine it would have much impact or performance one way or the other and it would give us cleaner function signatures.

tcl3: > Since we now pass the AbstractElement (from which we can get the Document) as part of the ComputationContext we don't actually need to pass this at all and can instead just compute it within compute_border_or_outline_width when required.

AFAICT, you can't always get the document from AbstractElement since it doesn't always have a value. Unless it's OK to make assumptions about what the value is if AbstractElement has no value?

> I haven't profiled this but I can't imagine it would have much impact or performance one way or the other and it would give us cleaner function signatures.

I was also surprised the compiler couldn't figure out how to optimize this. Using `REPORT_TIME()`, I got these results for the first 300000 calls
Before:
[Timing info for: StyleComputer::compute_property_values] current: 3.186µs   | average: 3.506µs   | total: 1.052s     | calls: 300000

After:
[Timing info for: StyleComputer::compute_property_values] current: 2.765µs   | average: 3.074µs   | total: 922.429ms  | calls: 300000

I also saw similar numbers when running the profiler.

Calme1709: The only cases where `ComputationContext::abstract_element` doesn't have a value is where we are computing CSS values in a non-property context (e.g. creating a `ParsedFontFace`, the `OffscreenCanvasRenderingContext2d::font` setter, etc), none of which are currently relevant for `border-*-width` or `outline-width`. I think therefore it is a safe assumption that if we reach `compute_border_or_outline_width` the passed `ComputationContext` will have an `abstract_element` which has a value. More than happy to leave how it is in case I am missing something, just mentioning this as a suggestion.

I could have phrased the second half of this better, I meant that my suggestion above wouldn't have an impact on performance, not your original change which is very impressive for such a small diff.

gmta: Let's merge this immediate perf win and optionally continue this discussion in another PR :)

tcl3: I tried what @Calme1709  suggested and saw quite a few crashes. I haven't had the opportunity to track down exactly what the cause was though. Seems worth looking at.

PR #6776, opened 2025-11-10
LibWeb: Implement CSS `perspective` property

Psychpsyo:
Implements the CSS `perspective` property so that we can have 3D!

image

<sup>diagetic image, I am not adding 3D glasses to ladybird</sup>

PR #6789, opened 2025-11-11
LibWeb: Make transition order consider property name

Psychpsyo:
This should fix the flake that #6776 has been suffering from. (I was able to recreate it locally and this does seem to fix it.)

### DIFF
--- a/Libraries/LibWeb/Animations/Animation.h
+++ b/Libraries/LibWeb/Animations/Animation.h
@@ -108,7 +108,7 @@ public:
     void set_owning_element(GC::Ptr<DOM::Element> value) { m_owning_element = value; }
 
     virtual AnimationClass animation_class() const { return AnimationClass::None; }
-    virtual Optional<int> class_specific_composite_order(GC::Ref<Animation>) const { return {}; }
+    virtual int class_specific_composite_order(GC::Ref<Animation>) const { return 0; }
 
     unsigned int global_animation_list_order() const { return m_global_animation_list_order; }
 
--- a/Libraries/LibWeb/CSS/CSSTransition.cpp
+++ b/Libraries/LibWeb/CSS/CSSTransition.cpp
@@ -38,7 +38,7 @@ Animations::AnimationClass CSSTransition::animation_class() const
     return Animations::AnimationClass::CSSTransition;
 }
 
-Optional<int> CSSTransition::class_specific_composite_order(GC::Ref<Animations::Animation> other_animation) const
+int CSSTransition::class_specific_composite_order(GC::Ref<Animations::Animation> other_animation) const
 {
     auto other = GC::Ref { as<CSSTransition>(*other_animation) };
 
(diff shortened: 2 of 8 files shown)


Libraries/LibWeb/CSS/CSSTransition.cpp
```diff
@@ -74,10 +74,14 @@ Optional<int> CSSTransition::class_specific_composite_order(GC::Ref<Animations::
     if (m_transition_generation != other->m_transition_generation)
```
gmta, line 74: Commit title: "consier" -> "consider"

Libraries/LibWeb/CSS/CSSTransition.cpp
```diff
@@ -74,10 +74,14 @@ Optional<int> CSSTransition::class_specific_composite_order(GC::Ref<Animations::
...
-    // FIXME:
     // 5. Otherwise, sort A and B in ascending order by the Unicode codepoints that make up the expanded transition
     //    property name of each transition (i.e. without attempting case conversion and such that ‘-moz-column-width’
     //    sorts before ‘column-width’).
+    if (transition_property() > other->transition_property())
+        return 1;
+    else
+        return -1;
+
     return {};
```
gmta: This `return {};` is dead code. But why not just `return transition_property().compare(other->transition_property())`?

gmta: Note that a small `FIXME` should be left since we're operating on StringViews, not Unicode strings like the spec step asks us to.

Psychpsyo: Done. I've also now made the function return `int` instead of `Optional<int>` since it can (does) just return 0 now to mean unorderable.

Libraries/LibWeb/Animations/Animation.h
```diff
@@ -108,7 +108,7 @@ class Animation : public DOM::EventTarget {
     void set_owning_element(GC::Ptr<DOM::Element> value) { m_owning_element = value; }
 
     virtual AnimationClass animation_class() const { return AnimationClass::None; }
-    virtual Optional<int> class_specific_composite_order(GC::Ref<Animation>) const { return {}; }
+    virtual int class_specific_composite_order(GC::Ref<Animation>) const { return {}; }
```
gmta: You should now probably `return 0;` here and in step 3 of the algorithm.

Psychpsyo: Right, sorry. Done now.

Libraries/LibWeb/CSS/CSSTransition.cpp
Psychpsyo: > This `return {};` is dead code. But why not just `return transition_property().compare(other->transition_property())`?

I figured out why not. `>` was not made for sorting comparison, so it'll result in 0 or 1, not -1 or 1.
I've made a new PR to fix this: #6795

gmta: I never mentioned `>` though?

PR #6791, opened 2025-11-11
 LibWeb: Use a faster string search algorithm for content filtering

tcl3:
This PR improves the performance of `ContentFilter` by using the [Aho-Corasick algorithm](<https://en.wikipedia.org/wiki/Aho%E2%80%93Corasick_algorithm>), which allows multiple strings to be checked simultaneously. Previously, we were checking each string one at a time.

This does use more memory that the naive implementation - about 200kb per WebContent process for the default URLs in BrowserContentFilters.txt. It should be possible to reduce memory usage if we want to filter more URLs in future.

Some performance numbers from https://amazon.com, which loads more URLs than most other sites: 

Before:
 [Timing info for: is_filtered] current: 108.937µs | average: 61.161µs  | total: 19.878ms   | calls: 325

After:
[Timing info for: is_filtered] current: 2.615µs   | average: 2.525µs   | total: 820.784µs | calls: 325

---
And some before and after times for building the filter:

Before:
[Timing info for: set_patterns] current: 11.071µs  | average: 11.071µs  | total: 11.071µs  | calls: 1

After:
[Timing info for: set_patterns] current: 856.23µs  | average: 856.23µs  | total: 856.23µs  | calls: 1

### DIFF
--- a/Libraries/LibWeb/Loader/ContentFilter.cpp
+++ b/Libraries/LibWeb/Loader/ContentFilter.cpp
@@ -1,10 +1,14 @@
 /*
  * Copyright (c) 2021, Andreas Kling <andreas@ladybird.org>
+ * Copyright (c) 2025, Tim Ledbetter <tim.ledbetter@ladybird.org>
  *
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
-#include <AK/StringBuilder.h>
+#include <AK/BinarySearch.h>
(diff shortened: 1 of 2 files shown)


tcl3:
My latest push tidies this up slightly by moving the string matching logic into its own class. I also decreased memory usage slightly by enforcing that strings must be ASCII, which is always true for URLs.

PR #6793, opened 2025-11-11
LibWeb: Add transform-style and have it make a stacking context

Psychpsyo:
This doesn't have any of the actual effects of transform-style yet, but it's a start.
And iirc, I already got some rendering improvements on https://vaie.art/pudle/ when I combined just this with #6776.


gmta:
The `all-prop-revert-layer.html` test seems to have another pass that you should rebaseline in!

Psychpsyo:
> The `all-prop-revert-layer.html` test seems to have another pass that you should rebaseline in!

Done now.

gmta (approved):

PR #6794, opened 2025-11-11
LibWeb: Cache the value of `Element::lang()`

tcl3:
This reduces the time spent in `SelectorEngine::matches_lang_pseudo_class()` from 1.90% to 0.41% on https://cloudflare.com

### DIFF
--- a/Libraries/LibWeb/DOM/Element.cpp
+++ b/Libraries/LibWeb/DOM/Element.cpp
@@ -3907,6 +3907,11 @@ void Element::attribute_changed(FlyString const& local_name, Optional<String> co
             m_dir = Dir::Auto;
         else
             m_dir = {};
+    } else if (local_name == HTML::AttributeNames::lang) {
+        for_each_in_inclusive_subtree_of_type<Element>([](auto& element) {
+            element.invalidate_lang_value();
+            return TraversalDecision::Continue;
+        });
     }
 
     // https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#reflecting-content-attributes-in-idl-attributes:concept-element-attributes-change-ext
@@ -4020,7 +4025,7 @@ void Element::set_counters_set(OwnPtr<CSS::CountersSet>&& counters_set)
 // https://html.spec.whatwg.org/multipage/dom.html#the-lang-and-xml:lang-attributes
 Optional<String> Element::lang() const
 {
-    auto attempt_to_determine_lang_attribute = [&]() -> Optional<String> {
+    auto determine_lang_attribute = [&]() -> String {
         // 1. If the node is an element that has a lang attribute in the XML namespace set
         //      Use the value of that attribute.
         auto maybe_xml_lang = get_attribute_ns(Namespace::XML, HTML::AttributeNames::lang);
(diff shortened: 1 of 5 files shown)


Libraries/LibWeb/DOM/Element.cpp
```diff
@@ -4065,11 +4065,13 @@ Optional<String> Element::lang() const
...
-    auto maybe_lang = attempt_to_determine_lang_attribute();
-    if (!maybe_lang.has_value() || maybe_lang->is_empty())
+    if (!m_lang_value.has_value()) {
+        // If the resulting value is the empty string, then it must be interpreted as meaning that the language of the node is explicitly unknown.
+        m_lang_value = attempt_to_determine_lang_attribute().value_or(String {});
+    }
+    if (m_lang_value->is_empty())
         return {};
-    return maybe_lang.release_value();
+    return m_lang_value;
```
gmta, line 4080: Hm, shouldn't we invalidate this cache if any of the conditions/values in `attempt_to_determine_lang_attribute` have changed?

tcl3: We should, and I wrote some code to do that, which I then forgot to include in the commit :sweat_smile:. Looks like I need to add a test because we don't have anything that exercises this.

tcl3: My latest change includes the invalidation and adds a test. It turns out this test also fails on `master` because we weren't previously invalidating style on lang attribute changes.

Libraries/LibWeb/DOM/Element.cpp
```diff
@@ -4065,11 +4071,21 @@ Optional<String> Element::lang() const
         return {};
     };
 
-    // If the resulting value is the empty string, then it must be interpreted as meaning that the language of the node is explicitly unknown.
-    auto maybe_lang = attempt_to_determine_lang_attribute();
-    if (!maybe_lang.has_value() || maybe_lang->is_empty())
+    if (!m_lang_value.has_value()) {
+        // If the resulting value is the empty string, then it must be interpreted as meaning that the language of the node is explicitly unknown.
+        m_lang_value = attempt_to_determine_lang_attribute().value_or(String {});
```
gmta: If `attempt_to_determine_lang_attribute()` returns a `String` instead of an `Optional<String>`, you no longer need this `.value_or(String {})`. Then the "attempt_" part of the name can be dropped as well, since we already work through all the steps and always return a value :)

Libraries/LibWeb/DOM/Element.cpp
```diff
@@ -3907,6 +3907,12 @@ void Element::attribute_changed(FlyString const& local_name, Optional<String> co
             m_dir = Dir::Auto;
         else
             m_dir = {};
+    } else if (local_name == HTML::AttributeNames::lang) {
+        invalidate_lang_value();
```
gmta: Since you use `for_each_in_inclusive_subtree_of_type()` below, you can drop this line.

Libraries/LibWeb/DOM/Element.h
```diff
@@ -626,6 +626,9 @@ class WEB_API Element
     // https://html.spec.whatwg.org/multipage/grouping-content.html#ordinal-value
     Optional<i32> m_ordinal_value;
 
+    mutable Optional<String> m_lang_value;
+    void invalidate_lang_value();
```
gmta: We have some logic that sets the document's pragma-set default language when a `<meta http-equiv="content-language">` element is inserted, which should also invalidate this new cache. Probably call `invalidate_lang_value()` on the document element?

PR #6795, opened 2025-11-11
LibWeb: Make transition order return -1 and 1 instead of 0 and 1

Psychpsyo:
The old behavior was plain incorrect for a sorting function.


Libraries/LibWeb/CSS/CSSTransition.cpp
```diff
@@ -78,7 +78,7 @@ int CSSTransition::class_specific_composite_order(GC::Ref<Animations::Animation>
     //    property name of each transition (i.e. without attempting case conversion and such that ‘-moz-column-width’
     //    sorts before ‘column-width’).
     // FIXME: This should operate on Unicode strings, not StringViews.
-    return transition_property() > other->transition_property();
+    return transition_property() > other->transition_property() ? 1 : -1;
```
gmta: Why not use `.compare()` here? See discussion in previous PR.

Psychpsyo: Somehow I completely misread that in the last PR, sorry.
Done that now.

PR #6798, opened 2025-11-12
LibWeb: Calculate stacking context bounds while recording display list

gmta:
We were calculating the bounding rects for all commands between a PushStackingContext and PopStackingContext on every display list execution. To get useful rects, we needed to apply the relevant scroll frames to each individual command as well.

Instead, let's just calculate the initial bounding rect for the stacking context while _recording_ the display list. Then, just like with all other commands, the stacking context's bounding rect gets translated by relevant scroll frames. This should work because there is no way a stacking context's bounding rect can grow or shrink in size because of nested scroll frames.

CC @kalenikaliaksandr

### DIFF
--- a/Libraries/LibWeb/Painting/DisplayListCommand.cpp
+++ b/Libraries/LibWeb/Painting/DisplayListCommand.cpp
@@ -95,7 +95,7 @@ void AddClipRect::dump(StringBuilder& builder) const
 void PushStackingContext::dump(StringBuilder& builder) const
 {
     auto affine_transform = extract_2d_affine_transform(transform.matrix);
-    builder.appendff("PushStackingContext opacity={} isolate={} has_clip_path={} transform={}", opacity, isolate, clip_path.has_value(), affine_transform);
+    builder.appendff("PushStackingContext opacity={} isolate={} has_clip_path={} transform={} bounding_rect={}", opacity, isolate, clip_path.has_value(), affine_transform, bounding_rect);
 }
 
 void PopStackingContext::dump(StringBuilder& builder) const


kalenikaliaksandr (changes requested):
this assumption is wrong: "This should work because there is no way a stacking context's bounding rect can grow or shrink in size because of nested scroll frames.".

the followin page, which has a fixed-position box inside a stacking context with opacity, is rendered correctly on the master branch, but on yours the fixed-position box disappears after scrolling.:
```html
<!DOCTYPE html>
<style>
    .sc {
        opacity: 0.5;
        width: 300px;
        height: 300px;
        background-color: mediumseagreen;
        overflow: scroll;
    }
    .box {
        width: 100px;
        height: 100px;
        border: 5px solid black;
        background-color: magenta;
    }
    .box.filler {
        margin-top: 1000px;
    }
    .fixed {
        position: fixed;
        width: 100px;
        height: 100px;
        border: 5px solid yellow;
        left: 500px;
        top: 500px;
        background-color: red;
    }
    .outerscroller {
        width: 500px;
        height: 500px;
        position: relative;
        background-color: blue;
    }
</style>
<div class="outerscroller">
    <div class="sc">
        <div class="fixed"></div>
    </div>
    <div class="box filler"></div>
</div>
```

gmta:
Closing this PR for now; our display list setup might need some changes before this becomes feasible. Fixed position elements should create a new stacking context, but those should be pulled out of the regular SC transform + effects tree (and of course, any scroll translation). At that point we could refactor the entire display list to be immutable and use implicit scroll translations during display list execution.

PR #6803, opened 2025-11-12
Meta: Update curl to version 8.17.0

loskutov:


vcpkg.json
```diff
@@ -210,7 +210,7 @@
     },
     {
       "name": "curl",
-      "version": "8.16.0#0"
+      "version": "8.17.0#0"
```
gmta, line 223: You need to update the vcpkg baseline as well.

PR #6805, opened 2025-11-12
LibWeb: Implement CSS perspective-origin

Psychpsyo:
What it says on the tin.
Reuses the `Position` type from `object-position`, since those two are really the same thing. (Technically `background-position` could also use this, but it's caught up in `BackgroundLayerData`, so that's a refactor for another day.)


Libraries/LibWeb/Painting/DisplayListRecorder.cpp
```diff
@@ -12,13 +12,42 @@
...
+
+    // The perspective matrix is computed as follows:
+
+    // 1. Start with the identity matrix.
+    auto perspective_matrix = Gfx::FloatMatrix4x4::identity();
+
+    // 2. Translate by the computed X and Y values of 'perspective-origin'
+    auto computed_x = parent_perspective_origin.x();
+    auto computed_y = parent_perspective_origin.y();
+    perspective_matrix = perspective_matrix * Gfx::translation_matrix(Vector3<float>(computed_x, computed_y, 0));
```
konradekk: note: in theory you should be good with the following:

```suggestion
    // 1. Start with the identity matrix.
    // 2. Translate by the computed X and Y values of 'perspective-origin'
    auto computed_x = parent_perspective_origin.x();
    auto computed_y = parent_perspective_origin.y();
    auto perspective_matrix = Gfx::translation_matrix(Vector3<float>(computed_x, computed_y, 0));
```

Psychpsyo: Fair point, and I've made that change now.
Though I am wondering if this isn't a thing that the compiler would optimize away anyways. Afterall, it's just a static computation.

Psychpsyo:
This one's actually an unrelated flake.

Libraries/LibWeb/CSS/ComputedProperties.h
```diff
@@ -83,6 +83,7 @@ class WEB_API ComputedProperties final : public JS::Cell {
     Optional<LengthPercentage> length_percentage(PropertyID, Layout::NodeWithStyle const&, ClampNegativeLengths) const;
     LengthBox length_box(PropertyID left_id, PropertyID top_id, PropertyID right_id, PropertyID bottom_id, Layout::NodeWithStyle const&, ClampNegativeLengths, LengthPercentageOrAuto const& default_value) const;
     Color color_or_fallback(PropertyID, ColorResolutionContext, Color fallback) const;
+    Position position_value(PropertyID) const;
```
gmta: This can be private instead of public.

Libraries/LibWeb/Painting/DisplayListRecorder.cpp
```diff
@@ -12,13 +12,40 @@
...
+
+    // 3. Multiply by the matrix that would be obtained from the 'perspective()' transform function, where the
+    //    length is provided by the value of the perspective property
+    // NB: Length values less than 1px being clamped to 1px is handled by the perspective() function already.
+    perspective_matrix = perspective_matrix * CSS::Transformation(CSS::TransformFunction::Perspective, Vector<CSS::TransformValue>({ CSS::TransformValue(CSS::Length::make_px(parent_perspective.value())) })).to_matrix({}).release_value();
+
+    // 4. Translate by the negated computed X and Y values of 'perspective-origin'
+    perspective_matrix = perspective_matrix * Gfx::translation_matrix(Vector3<float>(-computed_x, -computed_y, 0));
+
+    return perspective_matrix;
```
gmta: This whole computation should happen earlier in my opinion. It feels wrong to reach into CSS::TransformValue logic from the display list recorder. Looks like we only need the perspective matrix, so let's calculate that and put just the result of that in the command.

Libraries/LibWeb/Painting/DisplayListRecorder.cpp
konradekk: i believe that matrix multiplication is not really static as compiler probably knows nothing about it (that it is a no-op; you’d probably establish some formal structure for compiler to do such optimisations). so the real cost would be: the matrix itself and multiplication – not much imho.

Libraries/LibWeb/Painting/DisplayListRecorder.cpp
konradekk: note: it’s seems that this applies to `scale` too (before the changes) and seemingly only the `origin` and `matrix` are required. 🤔

Libraries/LibWeb/Painting/DisplayListRecorder.cpp
gmta: > you’d probably establish some formal structure for compiler

Like `constexpr` you mean? Which `::identity()` is. You could use Compiler Explorer to see the difference.

Libraries/LibWeb/CSS/ComputedProperties.h
Psychpsyo: Done.

Libraries/LibWeb/Painting/DisplayListRecorder.cpp
Psychpsyo: > This whole computation should happen earlier in my opinion. It feels wrong to reach into CSS::TransformValue logic from the display list recorder. Looks like we only need the perspective matrix, so let's calculate that and put just the result of that in the command.

That's what I had at first. Then I noticed that, the further you scroll down on https://poke-holo.simey.me/, the more wrong the 3D transforms become with this change, so I thought I probably need to calculate out the scroll offset from the perspective origin in the display list. (`PushStackingContext::translate_by()`)
That fixes the cards on that page, and the WPT tests still pass, so I assumed that was the way to go.
It was not.
So I reverted that now, wrote a new [WPT test](https://github.com/web-platform-tests/wpt/pull/56030) for it and 'imported' it. Not entirely sure what is at fault there, but maybe the cards need `transform-style` implemented proper since they are using it.

TL;DR:
Done, but sadly the Pokemon cards look worse for now.

Libraries/LibWeb/Painting/DisplayListRecorder.cpp
konradekk: doesn't one need a `constexpr` function to be called with compile-time known values to be optimized during compilation? i'm pretty sure that this would not work here, as the matrix on the right is known only in the runtime. (i'm glad to learn more, though: thanks for all your explanations and patience!)

Psychpsyo: > doesn't one need a `constexpr` function to be called with compile-time known values to be optimized during compilation? i'm pretty sure that this would not work here, as the matrix on the right is known only in the runtime. (i'm glad to learn more, though: thanks for all your explanations and patience!)

I'm also no expert, but an identity matrix is entirely 0s and 1s, so most of the operations in the matrix multiplication will end up being `0 * something`, `1 * something` and `0 + something`, all of which are things I'd assume the compiler to optimize away.

Libraries/LibWeb/CSS/ComputedProperties.h
```diff
@@ -288,6 +288,7 @@ class WEB_API ComputedProperties final : public JS::Cell {
 
     Overflow overflow(PropertyID) const;
     Vector<ShadowData> shadow(PropertyID, Layout::Node const&) const;
+    Position position_value(PropertyID) const;
```
AtkinsSJ, line 291: This move belongs in the previous commit.

Psychpsyo: Good point, done!

gmta:
Oof I guess I can't convince you to get the cards on https://poke-holo.simey.me/ in better shape, can I? 😇

image

Psychpsyo:
> Oof I guess I can't convince you to get the cards on https://poke-holo.simey.me/ in better shape, can I? 😇

I kinda really want to.
Will need to look at how much of a pain `transform-style` is to get to a servicable state.
(I have the beginnings of it on a branch somewhere already)

Psychpsyo:
> Oof I guess I can't convince you to get the cards on https://poke-holo.simey.me/ in better shape, can I? 😇

Done now: #7813

PR #6806, opened 2025-11-12
LibWeb: Make CSSPerspective correctly clamp its input for toMatrix()

Psychpsyo:
The WPT test change is currently being reviewed here: https://github.com/web-platform-tests/wpt/pull/56008
The behavior is to spec though, so I'd expect it to be accepted.


Libraries/LibWeb/CSS/CSSPerspective.cpp
```diff
@@ -121,7 +121,7 @@ WebIDL::ExceptionOr<GC::Ref<Geometry::DOMMatrix>> CSSPerspective::to_matrix() co
         [&matrix](GC::Ref<CSSNumericValue> const& numeric_value) -> WebIDL::ExceptionOr<void> {
             // NB: to() throws a TypeError if the conversion can't be done.
             auto distance = TRY(numeric_value->to("px"_fly_string))->value();
-            matrix->set_m34(-1 / (distance <= 0 ? 1 : distance));
+            matrix->set_m34(-1 / (distance < 1 ? 1 : distance));
```
gmta: Let's use `max(distance, 1)` for this :)

Psychpsyo: Done.

PR #6807, opened 2025-11-12
Meta: Set default preset in ladybird.py to Windows_CI on Windows

ADKaster:
The Windows_Experimental preset isn't tested in CI, so having the
out of the box script runs using it is a bit weird.

### DIFF
--- a/Meta/ladybird.py
+++ b/Meta/ladybird.py
@@ -253,14 +253,15 @@ def configure_skia_jemalloc() -> list[str]:
     user_vars_cmake_module = Path("Meta", "CMake", "vcpkg", "user-variables.cmake")
     user_vars_cmake_module.parent.mkdir(parents=True, exist_ok=True)
 
+    file_contents = []
+    if pkg_config:
+        file_contents.append(f"set(PKGCONFIG {pkg_config})\n")
+    if gn:
+        file_contents.append(f"set(GN {gn})\n")
+    file_contents.append("\n")
+
     with open(user_vars_cmake_module, "w") as f:
-        f.writelines(
-            [
-                f"set(PKGCONFIG {pkg_config})\n",
-                f"set(GN {gn})\n",
-                "\n",
-            ]
-        )
+        f.writelines(file_contents)
 
     return cmake_args
 


Meta/ladybird.py
```diff
@@ -38,7 +38,7 @@ def main():
         "--preset",
         required=False,
         default=os.environ.get(
-            "BUILD_PRESET", "Windows_Experimental" if platform.host_system == HostSystem.Windows else "Release"
+            "BUILD_PRESET", "Windows_CI" if platform.host_system == HostSystem.Windows else "Release"
```
gmta, line 41: The `*_CI` presets are meant to run in CI though; shouldn't we let this reflect the setup of the Linux/macOS presets and have a sane default expanded on by _CI and _Experimental?

ayeteadoe: Is the `ENABLE_WINDOWS_CI` option that is set in the `*_CI` presets still providing value?

4 locations using it could be replaced with `ENABLE_ADDRESS_SANITIZER` instead:
- https://github.com/LadybirdBrowser/ladybird/blob/master/Libraries/LibGfx/CMakeLists.txt#L147
- https://github.com/LadybirdBrowser/ladybird/blob/master/AK/CMakeLists.txt#L110
- https://github.com/LadybirdBrowser/ladybird/blob/master/Libraries/LibJS/CMakeLists.txt#L289
- https://github.com/LadybirdBrowser/ladybird/blob/master/Libraries/LibUnicode/CMakeLists.txt#L34

Then the remaining locations are:
- [LibLine](https://github.com/LadybirdBrowser/ladybird/blob/master/Libraries/CMakeLists.txt#L33)
- [test-262-runner](https://github.com/LadybirdBrowser/ladybird/blob/master/Utilities/CMakeLists.txt#L19)
- [Including `test-web` in CTest](https://github.com/LadybirdBrowser/ladybird/blob/master/Tests/LibWeb/test-web/CMakeLists.txt#L23)

Given these 3 locations above are not required to actually run/test the browser on Windows in its current experimental state, we could just fully disable them on Windows with FIXME's and consolidate the presets to lineup with Unix:
- `Windows_Experimental_Release` (the default in `ladybird.py`)
- `Windows_Experimental_Debug`
- `Windows_Sanitizer` (this is what is used in CI now)

PR #6810, opened 2025-11-13
Everywhere: Clean up AD-HOC and FIXME comments without colons

Psychpsyo:
Personal pet peeve: "FIXME" without ":"
I wanted to add this to `check-style.py`, but it turns out that you can't write a proper regex for this without it rejecting some valid stuff. (like using FIXME as a word in the middle of a sentence)

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLMediaElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLMediaElement.cpp
@@ -1259,8 +1259,8 @@ WebIDL::ExceptionOr<void> HTMLMediaElement::process_media_data(Function<void(Str
         if (enable == TriState::True)
             audio_track->set_enabled(true);
 
-        // AD-HOC(ish): According to https://dev.w3.org/html5/html-sourcing-inband-tracks/, kind should be set according to format, and the following criteria within
-        //              the specified formats.
+        // NB: According to https://dev.w3.org/html5/html-sourcing-inband-tracks/, kind should be set according to format, and the following criteria within
+        //     the specified formats.
         // WebM:
         //     - "main": the FlagDefault element is set on the track
         //     - "translation": not first audio (video) track
@@ -1318,7 +1318,7 @@ WebIDL::ExceptionOr<void> HTMLMediaElement::process_media_data(Function<void(Str
         if (enable == TriState::True)
             video_track->set_selected(true);
 
-        // AD-HOC(ish): See the comment regarding AudioTrack.kind above with regard to https://dev.w3.org/html5/html-sourcing-inband-tracks/.
+        // NB: See the comment regarding AudioTrack.kind above with regard to https://dev.w3.org/html5/html-sourcing-inband-tracks/.
         video_track->set_kind(enable == TriState::True ? "main"_utf16 : "translation"_utf16);
 
         // 7. Fire an event named addtrack at this VideoTrackList object, using TrackEvent, with the track attribute initialized to the new VideoTrack object.
--- a/Meta/check-style.py
+++ b/Meta/check-style.py
@@ -57,6 +57,9 @@ LOCAL_INCLUDE_SUFFIX_EXCLUDES = [
 # We check for and disallow any comments linking to the single-page HTML spec because it takes a long time to load.
 SINGLE_PAGE_HTML_SPEC_LINK = re.compile("//.*https://html\\.spec\\.whatwg\\.org/#")
 
+# We similarily check and disallow AD-HOCs and FIXMEs that aren't followed by a colon.
+INVALID_AD_HOC_OR_FIXME = re.compile(r'^(?:[\s\d./\-(*]+(?:AD-HOC|FIXME)[^:]|.*"FIXME[^:"]).*$', re.MULTILINE)
+
 
 def should_check_file(filename):
     if not filename.endswith(".cpp") and not filename.endswith(".h"):
@@ -97,6 +100,7 @@ def run():
     errors_include_missing_local = []
     errors_include_bad_complex = []
     errors_single_page_html_spec = []
+    errors_invalid_ad_hoc_or_fixme = {}
 
     for filename in find_files_here_or_argv():
         with open(filename, mode="r", encoding="utf-8") as f:
@@ -146,6 +150,9 @@ def run():
                         errors_include_missing_local.append(filename)
         if SINGLE_PAGE_HTML_SPEC_LINK.search(file_content):
             errors_single_page_html_spec.append(filename)
+        invalid_ad_hocs_or_fixmes = INVALID_AD_HOC_OR_FIXME.findall(file_content)
+        if invalid_ad_hocs_or_fixmes:
+            errors_invalid_ad_hoc_or_fixme[filename] = invalid_ad_hocs_or_fixmes
 
     have_errors = False
     if errors_license:
@@ -183,6 +190,10 @@ def run():
         have_errors = True
     if errors_single_page_html_spec:
         print("Files with links to the single-page HTML spec:", " ".join(errors_single_page_html_spec))
+        have_errors = True
+    if errors_invalid_ad_hoc_or_fixme:
+        for file in errors_invalid_ad_hoc_or_fixme:
+            print(f"{file} contains invalid AD-HOC or FIXME usages:", "".join(errors_invalid_ad_hoc_or_fixme[file]))
         have_errors = True
 
     if have_errors:


alimpfard:
Seems fine to me?
```
❯ git diff Meta
diff --git a/Meta/check-style.py b/Meta/check-style.py
index f6bb76acd66..09bed21d17e 100755
--- a/Meta/check-style.py
+++ b/Meta/check-style.py
@@ -44,6 +44,8 @@ ANY_INCLUDE_PATTERN = re.compile('^ *# *include\\b.*[>"](?!\\)).*$', re.M)
 SYSTEM_INCLUDE_PATTERN = re.compile("^ *# *include *<([^>]+)>(?: /[*/].*)?$")
 LOCAL_INCLUDE_PATTERN = re.compile('^ *# *include *"([^>]+)"(?: /[*/].*)?$')

+BAD_FIXME_ADHOC_PATTERN = re.compile(r'//\s*(FIXME|AD-?HOC)[^:].*')
+
 INCLUDE_CHECK_EXCLUDES = {}

 LOCAL_INCLUDE_ROOT_OVERRIDES = {}
@@ -97,6 +99,7 @@ def run():
     errors_include_missing_local = []
     errors_include_bad_complex = []
     errors_single_page_html_spec = []
+    errors_adhoc_fixme_bad_format = []

     for filename in find_files_here_or_argv():
         with open(filename, mode="r", encoding="utf-8") as f:
@@ -118,6 +121,9 @@ def run():
             errors_include_libc.append(filename)
         if BAD_INCLUDE_COMPLEX.search(file_content):
             errors_include_bad_complex.append(filename)
+        if m:=BAD_FIXME_ADHOC_PATTERN.search(file_content):
+            errors_adhoc_fixme_bad_format.append(filename)
+            print(m)
         if not is_in_prefix_list(filename, INCLUDE_CHECK_EXCLUDES):
             if include_root := find_matching_prefix(filename, LOCAL_INCLUDE_ROOT_OVERRIDES):
                 local_include_root = pathlib.Path(include_root)
@@ -184,6 +190,9 @@ def run():
     if errors_single_page_html_spec:
         print("Files with links to the single-page HTML spec:", " ".join(errors_single_page_html_spec))
         have_errors = True
+    if errors_adhoc_fixme_bad_format:
+        print("Files with bad FIXME or AD-HOC comments:", " ".join(errors_adhoc_fixme_bad_format))
+        have_errors = True

     if have_errors:
         sys.exit(1)

[test@turingcompleteBook] [0 jobs] ~/source/ladybird
❯ python3 Meta/check-style.py
<re.Match object; span=(246, 309), match='// FIXME Make the whole API more error-friendly. >
<re.Match object; span=(94062, 94191), match='// AD-HOC\n// FIXME: We should add a generic floo>
<re.Match object; span=(18947, 19008), match='// FIXME 8. If IsSharedArrayBuffer(arrayBuffer) i>
<re.Match object; span=(32823, 32913), match='// AD-HOC\nCrypto::UnsignedBigInteger const& temp>
In Libraries/LibWasm/Parser/Parser.cpp: Cannot find Libraries/LibWasm/Parser/AK/Enumerate.h
<re.Match object; span=(36721, 36895), match="// FIXME 10. If document is the associated Docume>
<re.Match object; span=(1497, 1532), match='// AD-HOC\n        .origin = origin,'>
<re.Match object; span=(920, 980), match='// AD-HOC (https://github.com/w3c/selection-api/i>
<re.Match object; span=(8849, 8878), match='// FIXME is the type correct?'>
<re.Match object; span=(12376, 12497), match='// FIXME\n    // 2. Return a suitably localized m>
<re.Match object; span=(1548, 1601), match='// AD-HOC https://github.com/whatwg/html/issues/1>
<re.Match object; span=(21499, 21579), match='// FIXME 3. If subject has the autofocus attribut>
<re.Match object; span=(51249, 51332), match='// AD-HOC\n    dispatch_event(DOM::Event::create(>
<re.Match object; span=(62464, 62621), match='// AD-HOC(ish): According to https://dev.w3.org/h>
<re.Match object; span=(3703, 3836), match='// FIXME 2. Move all the tasks that are to fire m>
<re.Match object; span=(26821, 26910), match="// FIXME 2. Assert: this is running within traver>
<re.Match object; span=(10062, 10272), match='// FIXME 2. If this OffscreenCanvas object\'s con>
<re.Match object; span=(18449, 18511), match='// FIXME 1. Let firePageSwapBeforeUnload be the f>
<re.Match object; span=(34201, 34238), match='// FIXME 8. Let uniqueHandle be null.'>
<re.Match object; span=(1604, 1648), match='// FIXME actually determine values for these'>
<re.Match object; span=(19531, 19624), match='// FIXME 1. Font features enabled by default, inc>
<re.Match object; span=(2311, 2339), match='// FIXME set the line-ascent'>
<re.Match object; span=(18203, 18338), match='// FIXME 2. Let importRequest be a new request wh>
<re.Match object; span=(1332, 1431), match='// FIXME\n    //  declaring all the namespace pre>
<re.Match object; span=(10112, 10204), match='// FIXME : This could be a redirect or a 401 auth>
<re.Match object; span=(19406, 19499), match="// FIXME shouldn't crash: EXPECT_EQ(days_since_ep>
In Utilities/wasm.cpp: Cannot find Utilities/LibWasm/Types.h
Files that #include a missing local file: Libraries/LibWasm/Parser/Parser.cpp Utilities/wasm.cpp
Files with bad FIXME or AD-HOC comments: Libraries/LibCrypto/BigFraction/BigFraction.h Libraries/LibJS/Runtime/AbstractOperations.cpp Libraries/LibJS/Runtime/ArrayBuffer.h Libraries/LibJS/Runtime/Temporal/AbstractOperations.cpp Libraries/LibWeb/DOM/Document.cpp Libraries/LibWeb/DOM/DocumentLoading.h Libraries/LibWeb/DOM/SelectionchangeEventDispatching.h Libraries/LibWeb/Fetch/Infrastructure/HTTP/Responses.h Libraries/LibWeb/HTML/FormAssociatedElement.cpp Libraries/LibWeb/HTML/HTMLButtonElement.cpp Libraries/LibWeb/HTML/HTMLDialogElement.cpp Libraries/LibWeb/HTML/HTMLImageElement.cpp Libraries/LibWeb/HTML/HTMLMediaElement.cpp Libraries/LibWeb/HTML/MessagePort.cpp Libraries/LibWeb/HTML/Navigable.cpp Libraries/LibWeb/HTML/OffscreenCanvas.cpp Libraries/LibWeb/HTML/TraversableNavigable.cpp Libraries/LibWeb/HTML/WindowOrWorkerGlobalScope.cpp Libraries/LibWeb/HighResolutionTime/Performance.cpp Libraries/LibWeb/Layout/InlineLevelIterator.cpp Libraries/LibWeb/MathML/MathMLMspaceElement.cpp Libraries/LibWeb/ServiceWorker/Job.cpp Libraries/LibWeb/XML/XMLFragmentParser.cpp Libraries/LibWebSocket/WebSocket.cpp Tests/AK/TestTime.cpp
```

Psychpsyo:
> Seems fine to me?

Fails for these types of comments:
```
// AD-HOC: See AD-HOC comment on Element.m_had_duplicate_attribute_during_tokenization about why this is done.
// Pop off the dummy execution context. See the above FIXME block about why this is done.
// Step 4 handled separately by the FIXME above.
```
and also for all `attribute.extended_attributes.contains("FIXME")` in the IDL generator.

And it fails for the two `// AD-HOC(ish):` comments in `HTMLMediaElement.cpp`

alimpfard:
It only tries to match // FIXME or // AD-HOC so it can't fail for the first one
Second one is annoying to me anyway, ad-hoc is ad-hoc.

Psychpsyo:
> It only tries to match // FIXME or // AD-HOC so it can't fail for the first one Second one is annoying to me anyway, ad-hoc is ad-hoc.

Oh, right, that fails to do validation on all of these styles of FIXME then:
```cpp
// - FIXME: If realm is not null, each time job is invoked the implementation must perform implementation-defined steps such that execution is prepared to evaluate ECMAScript code at the time of job's invocation.
// 3. FIXME: The user agent may wait for more bytes of the resource to be available, either in this step or at any later step in this algorithm.
//     FIXME: navigation timing type: navTimingType
// (FIXME: Implement our own converter!)
dbgln("FIXME: Implement UTCTime with offset!");
```
Which I guess is fine, but it'd be weird to have a check that only complains sometimes, depending on how you put your punctuation.

Psychpsyo:
Actually, seeing that list now...
That might be a limited-enough subset to get with a regex. (I was trying to do a bit more earlier and gave up after a while)

PR #6811, opened 2025-11-13
LibWeb: Fix a few presentational-hint things

AtkinsSJ:
See commits for details. Main goal here was to introduce and use `parse_css_type()`, as a few places were using other functions that do a similar-but-wrong job, but there were a few easy fixes I stumbled across while looking into that.

### DIFF
--- a/Libraries/LibWeb/MathML/MathMLElement.cpp
+++ b/Libraries/LibWeb/MathML/MathMLElement.cpp
@@ -121,9 +121,9 @@ void MathMLElement::apply_presentational_hints(GC::Ref<CSS::CascadedProperties> 
             // are respectively mapped to add(<U>) add(<-U>) and <U>.
             if (Optional<StringView> parsed_value = HTML::parse_integer_digits(value); parsed_value.has_value()) {
                 auto string_value = parsed_value.value();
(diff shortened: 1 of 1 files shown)


Libraries/LibWeb/MathML/MathMLMspaceElement.cpp
```diff
@@ -60,7 +60,8 @@ void MathMLMspaceElement::apply_presentational_hints(GC::Ref<CSS::CascadedProper
     auto depth_value = parse_non_percentage_value(AttributeNames::depth);
 
     if (height_value && depth_value) {
-        // FIXME: set the presentational hint to calculate height + depth
+        auto height_string = MUST(String::formatted("calc({} + {})", attribute(AttributeNames::height).value(), attribute(AttributeNames::depth).value()));
```
lpas, line 63: this reads a little funny. I assume we can not use `height_value` & `depth_value` here directly ?

AtkinsSJ: It is a bit weird, but reading the attribute strings seems better to me than doing the work to serialize the StyleValues again.

It would be an option to try and construct a CalculatedStyleValue directly. I just didn't think the extra complexity would be worth it, for code that's going to run so rarely.

PR #6812, opened 2025-11-13
LibJS: Handle negativity in Temporal's ApplyUnsignedRoundingMode

trflynn89:
I don't fully understand the BigInt math here, as the computation for d1 and d2 don't align with the spec due to BigInt logic. This was discussed a bit in SerenityOS's Discord some years ago:

https://discord.com/channels/830522505605283862/851522357734408232/978786665306918932

But some new tests in test262 indicate that we need to handle negative values here, instead of just throwing away the sign.

test262 diff (back to 100% passing):
```
test/built-ins/Temporal/Instant/prototype/round/negative-instant.js                         ❌ -> ✅
test/built-ins/Temporal/Instant/prototype/toString/negative-instant-rounding.js             ❌ -> ✅
test/built-ins/Temporal/ZonedDateTime/prototype/toString/negative-zoneddatetime-rounding.js ❌ -> ✅
```

### DIFF
--- a/Libraries/LibJS/Runtime/Temporal/AbstractOperations.cpp
+++ b/Libraries/LibJS/Runtime/Temporal/AbstractOperations.cpp
@@ -947,10 +947,11 @@ Crypto::SignedBigInteger apply_unsigned_rounding_mode(Crypto::SignedDivisionResu
         return r2;
 
     // 6. Let d1 be x – r1.
-    auto d1 = x.remainder.unsigned_value();
+    auto const& d1 = x.remainder;
 
     // 7. Let d2 be r2 – x.
-    auto d2 = MUST(increment.minus(x.remainder.unsigned_value()));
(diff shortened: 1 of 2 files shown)


trflynn89:
The Temporal polyfill (written by the spec authors) is also confusing to me in that they do this computation before invoking `ApplyUnsignedRoundingMode` to avoid computing d1 and d2 at all:
https://github.com/tc39/proposal-temporal/blob/c1f7bd564466ebbdcfc16efbdb9d3d8d9dc379d6/polyfill/lib/ecmascript.mjs#L4124-L4129
```js
const cmp = remainder.times(2).abs().compare(increment) * (quantity.lt(0) ? -1 : 1);
const rounded = remainder.isZero()
    ? quotient
    : ApplyUnsignedRoundingMode(r1, r2, cmp, r1.isEven(), unsignedRoundingMode);
```

PR #6813, opened 2025-11-13
LibWeb: Add `--headless=manual` option to run until explicit exit

tcl3:
This headless mode will stay open until it is closed by manually or by calling `window.close()`.

I'm using this to run Speedometer headlessly.

### DIFF
--- a/Libraries/LibWebView/Application.cpp
+++ b/Libraries/LibWebView/Application.cpp
@@ -133,7 +133,7 @@ ErrorOr<void> Application::initialize(Main::Arguments const& arguments)
 
     args_parser.add_option(Core::ArgsParser::Option {
         .argument_mode = Core::ArgsParser::OptionArgumentMode::Optional,
-        .help_string = "Run Ladybird without a browser window. Mode may be 'screenshot' (default), 'layout-tree', or 'text'.",
+        .help_string = "Run Ladybird without a browser window. Mode may be 'screenshot' (default), 'layout-tree', 'text', or 'manual'.",
         .long_name = "headless",
         .value_name = "mode",
         .accept_value = [&](StringView value) {
@@ -146,6 +146,8 @@ ErrorOr<void> Application::initialize(Main::Arguments const& arguments)
                 headless_mode = HeadlessMode::LayoutTree;
             else if (value.equals_ignoring_ascii_case("text"sv))
                 headless_mode = HeadlessMode::Text;
+            else if (value.equals_ignoring_ascii_case("manual"sv))
+                headless_mode = HeadlessMode::Manual;
 
             return headless_mode.has_value();
         },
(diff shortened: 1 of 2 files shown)


Libraries/LibWebView/Application.cpp
```diff
@@ -146,6 +146,8 @@ ErrorOr<void> Application::initialize(Main::Arguments const& arguments)
                 headless_mode = HeadlessMode::LayoutTree;
             else if (value.equals_ignoring_ascii_case("text"sv))
                 headless_mode = HeadlessMode::Text;
+            else if (value.equals_ignoring_ascii_case("test"sv))
+                headless_mode = HeadlessMode::Test;
```
tcl3: Would it be worth renaming this mode to something more generic like `HeadlessMode::Manual` and `internals.signalTestIsDone()` to `internals.signalExit()`?

Other naming suggestions welcome

trflynn89: `Test` is used by test-web to keep the browser running after a page load, so it's sort of an internal thing. A separate flag would make more sense IMO, just so we aren't overloading it.

trflynn89: How does `internals.signalTestIsDone()` get invoked on real page loads?

tcl3: > Test is used by test-web to keep the browser running after a page load, so it's sort of an internal thing

Do you think it would be OK to rename `internals.signalTestIsDone()` and `on_test_finish` to something more generic, or would it be better to leave them as they are?

> How does `internals.signalTestIsDone()` get invoked on real page loads?

It would only be invoked when called explicitly. The way that I'm using it with Speedometer is that I've made a simpler version of the [`InteractiveBenchmarkRunner`](<https://github.com/WebKit/Speedometer/blob/release/3.1/resources/interactive.mjs>), which I use to run the tests. I call `internals.signalTestIsDone()` [here](https://github.com/WebKit/Speedometer/blob/1386415be8fef2f6b6bbdbe1828872471c5d802a/resources/interactive.mjs#L90-L92), when all tests have finished running and I have the results.

trflynn89: hmm I'm mostly just a little iffy about hooking `internals` into this. Seems a bit specific to this one use case. Maybe the general mode here can be to run until ctrl+c or `window.close()` is called? That would be something like:
```c++
static void load_page_and_exit_on_close(Core::EventLoop& event_loop, HeadlessWebView& view, URL::URL const& url)
{
    view.on_close = [&event_loop]() {
        event_loop.quit(0);
    };

    view.load(url);
}
```

And I was able to get that to work with this speedometer diff (using `http://127.0.0.1:8080/InteractiveRunner.html?startAutomatically&iterationCount=1` for a test):
```diff
diff --git a/resources/interactive.mjs b/resources/interactive.mjs
index bdb2bc4..8a3c604 100644
--- a/resources/interactive.mjs
+++ b/resources/interactive.mjs
@@ -85,10 +85,12 @@ class InteractiveBenchmarkRunner extends BenchmarkRunner {
         const pre = document.createElement("pre");
         document.body.appendChild(pre);
         pre.textContent = results;
+        console.log(results);
     }

     _done() {
         this.isRunning = false;
+        window.close();
     }

     runStep() {
```

tcl3: My latest push adds a `--headless=manual` option, which does as you suggested.

Since my script starts a HTTP server to run the Speedometer benchmark anyway, I added a POST endpoint that I send the results to - after the results are received the script send a signal to stop the Ladybird process. This avoids using `internals` entirely.

trflynn89 (approved):

PR #6815, opened 2025-11-13
LibWeb: Handle empty string namespaces in selector matching

mikiubo:
Per the CSS Namespaces spec, an empty string declared in an @namespace rule represents no namespace.

Fixes WPT:
- css/css-namespaces/prefix-002.xml
- css/css-namespaces/prefix-003.xml


AtkinsSJ (approved):
👍

PR #6816, opened 2025-11-13
LibWeb: Add WebXR's XRSystem

Psychpsyo:
No WPT passes yet, since that needs some webXR specific testing stuff implemented. (Though I'm working on that)

Continued in #6833.


Psychpsyo:
Not stale, just unreviewed

PR #6818, opened 2025-11-13
LibWeb: Fix dialog cleanup when disconnected from DOM

MrBrain295:
When a dialog element is removed from the DOM via element.remove(), properly clean up its state to prevent interference with subsequent close requests.
+1 WPT

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLDialogElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLDialogElement.cpp
@@ -50,18 +50,27 @@ void HTMLDialogElement::removed_from(Node* old_parent, Node& old_root)
 {
     HTMLElement::removed_from(old_parent, old_root);
 
-    // 1. If removedNode's close watcher is not null, then:
-    if (m_close_watcher) {
-        // 1.1. Destroy removedNode's close watcher.
-        m_close_watcher->destroy();
-        // 1.2. Set removedNode's close watcher to null.
-        m_close_watcher = nullptr;
-    }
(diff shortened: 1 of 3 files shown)


Libraries/LibWeb/HTML/HTMLDialogElement.cpp
gmta, line 1: Commit title: please do not [use weasel-words like "fix"](https://github.com/LadybirdBrowser/ladybird/blob/master/CONTRIBUTING.md#code-submission-policy), but instead describe what you're changing and why.

Commit message: maybe mention in which WPT subcategory you've achieved the +1 subtest passes.

Libraries/LibWeb/HTML/HTMLDialogElement.cpp
```diff
@@ -50,18 +50,27 @@ void HTMLDialogElement::removed_from(Node* old_parent, Node& old_root)
...
-        // 1.1. Destroy removedNode's close watcher.
-        m_close_watcher->destroy();
-        // 1.2. Set removedNode's close watcher to null.
-        m_close_watcher = nullptr;
-    }
+    // AD-HOC: When a dialog is removed from the document, we additionally set is_modal to false (below)
+    // and remove it from the open dialogs list (see line 65). This ensures it doesn't interfere with close
+    // requests and matches Chromium's behavior. The other steps below (removing from the top layer and destroying
+    // the close watcher) are mandated by the HTML spec. See https://wpt.live/close-watcher/user-activation/y-dialog-disconnected.html
+    set_is_modal(false);
```
gmta, line 57: If this is not in the spec, should it be? Is there an upstream spec issue that you can reference here?

Issue #6729, opened 2025-11-06
`float`ed inline content does not play nice with `width: max-content`


Psychpsyo:
### Summary

When floating an inline element right, it still counts towards the `max-content` of its parent as if it were part of its line, even if that line has plenty of free space to fit the element. This results in it looking offset.

### Operating system

Linux

### Steps to reproduce

1. Look at the reduced test case.

### Expected behavior

The floated element should not contribute to `max-content` if there is free space that it could float in.

### Actual behavior

The floated element does contribute to `max-content`.

### URL for a reduced test case

N/A

### HTML/SVG/etc. source for a reduced test case

```html
<!DOCTYPE html>
<div style="width: max-content">
	--------|<br>
	<span style="float: right">|</span>
</div>
```

### Log output and (if possible) backtrace

```shell
none
```

### Screenshots or screen recordings

In order: Firefox, Chromium, Webkit, Ladybird
<img width="272" height="689" alt="Image" src="https://github.com/user-attachments/assets/b4cbfde2-118f-4755-9753-de9bceec62bb" />

### Build flags or config settings

none

### Contribute a patch?

- [ ] I’ll contribute a patch for this myself.

PR #6819, opened 2025-11-14
fix: Break lines on non-whitespace items

### LINKED ISSUE #6266: layout bug with floating, full screen elements
### Summary

In my website there is a text that say "You May Also Enjoy". I've noticed the "You" disappear. after some research it seem that if element has  "float: left;width: 100%;" ladybird try to append the next element after that 100% width, which make it start out of the page bonds.

### Operating system

Linux

### Steps to reproduce

1. open a website with element that has `float: <any>;width: 100%;`
2. notice there are random text after the page width that shouldnt be there
3.

### Expected behavior

The layout is display correctly: `First Second Third` in the same line 

### Actual behavior

The first word (`First`) disappear to be after the page width.

N/A

```html
<div style="float: left;width: 100%;">
    Text
</div>


<h2>First Second Third</h2>
```

```shell
N/A
```

Image

_No response_

### Contribute a patch?

MiniX16:
## Description

This PR fixes the layout bug where text following a `float: left; width: 100%` element is pushed outside the visible page area.

Example from the issue:
```html
<div style="float: left; width: 100%;">Text</div>
<h2>First Second Third</h2>
```
Ladybird incorrectly placed “First” beyond the page boundary because the line breaker only checked `whitespace` when deciding whether text fits, and an extra float-intrusion check incorrectly treated the 100% float as non-blocking.

This patch:
- Applies break_if_needed() to all inline fragments, not just whitespace, ensuring text never gets placed when it doesn’t fit.
- Removes an overly permissive float-intrusion check so floats correctly reduce available line width.

As a result, text after full-width floats now renders in the correct position.

## Result

- Before changes:
image

- After changes:
image

## Related Issue

Closes #6266

### DIFF
--- a/Libraries/LibWeb/Layout/InlineFormattingContext.cpp
+++ b/Libraries/LibWeb/Layout/InlineFormattingContext.cpp
@@ -344,6 +344,8 @@ void InlineFormattingContext::generate_line_boxes()
                 // If whitespace caused us to break, we swallow the whitespace instead of putting it on the next line.
                 if (is_whitespace && next_width > 0 && line_builder.break_if_needed(item.border_box_width() + next_width))
                     break;
+                if (!is_whitespace)
+                    line_builder.break_if_needed(item.border_box_width());
             } else if (text_node.computed_values().text_overflow() == CSS::TextOverflow::Ellipsis
                 && text_node.computed_values().overflow_x() != CSS::Overflow::Visible) {
                 // We may need to do an ellipsis if the text is too long for the container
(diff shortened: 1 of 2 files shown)


Psychpsyo:
Curious if this might have an effect on https://github.com/LadybirdBrowser/ladybird/issues/6729

MiniX16:
Nahh, I am getting the same result. Even so, it seems to be related so I will try to fix that later in a new pr.

image

PR #6821, opened 2025-11-14
LibWeb: Hide stepper buttons with `appearance: textfield`

AtkinsSJ:
Progress on [Lego.com](https://www.lego.com/en-gb/product/dinosaur-fossils-tyrannosaurus-rex-76968):

Before:

image

After:

image

`appearance` is a bit of a weird one. The [spec](https://drafts.csswg.org/css-ui/#typedef-appearance-compat-special) allows for `appearance: textfield` and `appearance: menulist-button` to modify how inputs appear in a hand-wavy way. Firefox hides the stepper buttons on number inputs, which is what I've imitated. (Chrome doesn't, but provides a pseudo-element for the buttons that can be used to hide them instead.)

Though on Lego.com they don't set `appearance: textfield`, they set `-moz-appearance: textfield`. I've added `-moz-appearance` as a legacy alias, assuming that other sites probably also do this because it's a property that's been in a semi-specced state for a very long time.

### DIFF
--- a/Libraries/LibWeb/CSS/Properties.json
+++ b/Libraries/LibWeb/CSS/Properties.json
@@ -1,4 +1,7 @@
 {
+  "-moz-appearance": {
+    "legacy-alias-for": "appearance"
+  },
   "-webkit-align-content": {
     "legacy-alias-for": "align-content"
   },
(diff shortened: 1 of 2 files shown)


Libraries/LibWeb/CSS/Properties.json
```diff
@@ -1,4 +1,7 @@
 {
+  "-moz-appearance": {
```
tcl3, line 2: Are `-moz` prefixes supported outside of Firefox? Should we be supporting them?

AtkinsSJ: Are they? No as far as I can tell. Should we? Debatable.

I did so here because otherwise it didn't fix lego.com. They use the moz property for Firefox, and a WebKit-specific pseudo element for Safari/Chrome. Neither is spec, and this way was simple to implement. I don't think it's harmful and we've previously implemented WebKit-specific things we technically shouldn't, so I think it's ok?

Issue #3448, opened 2025-02-03
headless-browser: Ref-test screenshots are taken before CSS-linked resources are loaded


InvalidUsernameException:
> The page screenshot is taken when the `load` event fires, but this event does not wait for resources linked to by CSS (such as `background-image: url("foo.png")`, or `@font-face`, or the like). This makes testing certain CSS features very difficult. It would be helpful to wait until all such resources have loaded before taking the screenshot. (Note that this will require a separate mechanism, as the load event is specced to not wait for these, at least until https://github.com/w3c/csswg-drafts/issues/1088 gets resolved.)
>
> A hack that can be used for images for now, is to add an `<img src="foo.png">` tag to the document, since the load event will wait for those images to load.

Imported from https://github.com/SerenityOS/serenity/issues/21957

InvalidUsernameException:
Imported here so I have something to link to for a FIXME comment.

PR #6823, opened 2025-11-14
LibWeb/CSS: Implement 5.3.4: The effect value of a keyframe effect

JeremiasRy:
Implements [5.3.4. The effect value of a keyframe effect](https://www.w3.org/TR/web-animations-1/#the-effect-value-of-a-keyframe-animation-effect)

Fixes [6549](https://github.com/LadybirdBrowser/ladybird/issues/6549)

### DIFF
--- a/Libraries/LibWeb/Animations/KeyframeEffect.h
+++ b/Libraries/LibWeb/Animations/KeyframeEffect.h
@@ -70,6 +70,13 @@ public:
             HashMap<CSS::PropertyID, Variant<UseInitial, NonnullRefPtr<CSS::StyleValue const>>> properties {};
             Bindings::CompositeOperationOrAuto composite { Bindings::CompositeOperationOrAuto::Auto };
         };
+
+        struct ComputedKeyframe {
+            HashMap<CSS::PropertyID, NonnullRefPtr<CSS::StyleValue const>> properties {};
+            Bindings::CompositeOperationOrAuto composite { Bindings::CompositeOperationOrAuto::Auto };
+            double keyframe_offset_value;
+        };
+
         RedBlackTree<u64, ResolvedKeyFrame> keyframes_by_key;
     };
     static void generate_initial_and_final_frames(RefPtr<KeyFrameSet>, HashTable<CSS::PropertyID> const& animated_properties);
@@ -112,6 +119,9 @@ public:
 
     virtual void update_computed_properties(AnimationUpdateContext&) override;
 
+    CSS::EasingFunction const& default_keyframe_timing_function() { return m_keyframe_default_timing_function; }
+    void set_default_keyframe_timing_function(CSS::EasingFunction value) { m_keyframe_default_timing_function = move(value); }
+
 private:
     KeyframeEffect(JS::Realm&);
     virtual ~KeyframeEffect() override = default;
@@ -135,6 +145,8 @@ private:
     Vector<GC::Ref<JS::Object>> m_keyframe_objects {};
 
     RefPtr<KeyFrameSet const> m_key_frame_set {};
+
+    CSS::EasingFunction m_keyframe_default_timing_function { CSS::EasingFunction::linear() };
 };
 
 }
--- a/Libraries/LibWeb/CSS/StyleComputer.cpp
+++ b/Libraries/LibWeb/CSS/StyleComputer.cpp
@@ -569,49 +569,11 @@ void StyleComputer::collect_animation_into(DOM::AbstractElement abstract_element
         return;
     }
 
-    double progress = round(output_progress.value() * 100.0 * Animations::KeyframeEffect::AnimationKeyFrameKeyScaleFactor);
-    // FIXME: Support progress values outside the range of i64.
-    i64 key = 0;
-    if (progress > NumericLimits<i64>::max()) {
-        key = NumericLimits<i64>::max();
-    } else if (progress < NumericLimits<i64>::min()) {
-        key = NumericLimits<i64>::min();
-    } else {
-        key = static_cast<i64>(progress);
-    }
-    auto keyframe_start_it = [&] {
-        if (output_progress.value() <= 0) {
-            return keyframes.begin();
-        }
-        auto potential_match = keyframes.find_largest_not_above_iterator(key);
-        auto next = potential_match;
-        ++next;
-        if (next.is_end()) {
-            --potential_match;
-        }
-        return potential_match;
-    }();
-    auto keyframe_start = static_cast<i64>(keyframe_start_it.key());
-    auto keyframe_values = *keyframe_start_it;
-
-    auto keyframe_end_it = ++keyframe_start_it;
-    VERIFY(!keyframe_end_it.is_end());
-    auto keyframe_end = static_cast<i64>(keyframe_end_it.key());
-    auto keyframe_end_values = *keyframe_end_it;
-
-    auto progress_in_keyframe = (progress - keyframe_start) / static_cast<double>(keyframe_end - keyframe_start);
-
-    if constexpr (LIBWEB_CSS_ANIMATION_DEBUG) {
-        auto valid_properties = keyframe_values.properties.size();
-        dbgln("Animation {} contains {} properties to interpolate, progress = {}%", animation->id(), valid_properties, progress_in_keyframe * 100);
-    }
-
-    // FIXME: Follow https://drafts.csswg.org/web-animations-1/#ref-for-computed-keyframes in whatever the right place is.
-    auto compute_keyframe_values = [&computed_properties, &abstract_element, this](auto const& keyframe_values) {
-        HashMap<PropertyID, RefPtr<StyleValue const>> result;
-        HashMap<PropertyID, PropertyID> longhands_set_by_property_id;
-        auto property_is_set_by_use_initial = MUST(Bitmap::create(number_of_longhand_properties, false));
-
+    // FIXME: Figure out how to use effect->target_properties(), using it causes few dozen tests to fail
+    HashTable<CSS::PropertyID> target_properties;
+    // 5.3.3. Calculating computed keyframes
+    // https://drafts.csswg.org/web-animations-1/#ref-for-computed-keyframes
+    auto computed_keyframe_values = [&keyframes, &abstract_element, &computed_properties, &target_properties, this] {
         auto property_is_logical_alias_including_shorthands = [&](PropertyID property_id) {
             if (property_is_shorthand(property_id))
                 // NOTE: All expanded longhands for a logical alias shorthand are logical aliases so we only need to check the first one.
@@ -620,14 +582,15 @@ void StyleComputer::collect_animation_into(DOM::AbstractElement abstract_element
             return property_is_logical_alias(property_id);
         };
 
-        // https://drafts.csswg.org/web-animations-1/#ref-for-computed-keyframes
         auto is_property_preferred = [&](PropertyID a, PropertyID b) {
-            // If conflicts arise when expanding shorthand properties or replacing logical properties with physical properties, apply the following rules in order until the conflict is resolved:
+            // If conflicts arise when expanding shorthand properties or replacing logical properties with physical properties,
+            // apply the following rules in order until the conflict is resolved:
             // 1. Longhand properties override shorthand properties (e.g. border-top-color overrides border-top).
             if (property_is_shorthand(a) != property_is_shorthand(b))
                 return !property_is_shorthand(a);
 
-            // 2. Shorthand properties with fewer longhand components override those with more longhand components (e.g. border-top overrides border-color).
+            // 2. Shorthand properties with fewer longhand components override those with more longhand components
+            //    (e.g. border-top overrides border-color).
             if (property_is_shorthand(a)) {
                 auto number_of_expanded_shorthands_a = expanded_longhands_for_shorthand(a).size();
                 auto number_of_expanded_shorthands_b = expanded_longhands_for_shorthand(b).size();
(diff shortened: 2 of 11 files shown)


Libraries/LibWeb/CSS/StyleValues/AngleStyleValue.h
```diff
@@ -20,6 +20,11 @@ class AngleStyleValue : public DimensionStyleValue {
     {
         return adopt_ref(*new (nothrow) AngleStyleValue(move(angle)));
     }
+    static ValueComparingNonnullRefPtr<StyleValue const> neutral_value()
+    {
+        Angle angle(0, AngleUnit::Deg);
```
Calme1709: We tend to not use this constructor syntax and instead prefer the curly-brace one instead - although the `Angle::make_degrees` method is probably more useful here.

Libraries/LibWeb/CSS/StyleValues/BorderRadiusStyleValue.h
```diff
@@ -31,6 +32,14 @@ class BorderRadiusStyleValue final : public StyleValueWithDefaultOperators<Borde
 
     bool properties_equal(BorderRadiusStyleValue const& other) const { return m_properties == other.m_properties; }
 
+    static ValueComparingNonnullRefPtr<StyleValue const> neutral_value()
+    {
+        auto horizontal_radius = IntegerStyleValue::create(0);
```
Calme1709: These should be `LengthStyleValue` rather than `IntegerStyleValue`

Libraries/LibWeb/CSS/StyleValues/FlexStyleValue.h
```diff
@@ -33,6 +33,12 @@ class FlexStyleValue final : public DimensionStyleValue {
         return m_flex == other_flex.m_flex;
     }
 
+    static ValueComparingNonnullRefPtr<StyleValue const> neutral_value()
+    {
+        Flex flex = Flex(0, FlexUnit::Fr);
```
Calme1709: Same as above with `Angle` initialization syntax - the `Flex::make_fr()` method is useful here.

Libraries/LibWeb/CSS/StyleValues/FrequencyStyleValue.h
```diff
@@ -36,6 +36,12 @@ class FrequencyStyleValue final : public DimensionStyleValue {
         return m_frequency == other_frequency.m_frequency;
     }
 
+    static ValueComparingNonnullRefPtr<StyleValue const> neutral_value()
+    {
+        Frequency freq = Frequency(0, FrequencyUnit::Hz);
```
Calme1709: Same as above: `Frequency::make_hz()`

Libraries/LibWeb/CSS/StyleValues/RatioStyleValue.h
```diff
@@ -36,6 +36,11 @@ class RatioStyleValue final : public StyleValueWithDefaultOperators<RatioStyleVa
 
     bool properties_equal(RatioStyleValue const& other) const { return m_ratio == other.m_ratio; }
 
+    static ValueComparingNonnullRefPtr<StyleValue const> neutral_value()
+    {
+        return RatioStyleValue::create(Ratio(1, 0));
```
Calme1709: We tend to prefer using the uniform intialization syntax (i.e. `Ratio { 1, 0 }` instead of `Ratio(1, 0)`.

Also I haven't tested it but I'm skeptical of 1/0 being the "neutral" value for a ratio - 1/1 seems more likely.

Libraries/LibWeb/CSS/StyleValues/ScrollbarColorStyleValue.h
```diff
@@ -23,6 +23,12 @@ class ScrollbarColorStyleValue final : public StyleValueWithDefaultOperators<Scr
     NonnullRefPtr<StyleValue const> thumb_color() const { return m_thumb_color; }
     NonnullRefPtr<StyleValue const> track_color() const { return m_track_color; }
 
+    static ValueComparingNonnullRefPtr<StyleValue const> neutral_value()
+    {
+        auto transparent_color = ColorStyleValue::create_from_color(Color::Transparent, ColorSyntax::Modern, {});
```
Calme1709: Perhaps we could reuse the `ColorStyleValue::neutral_value()` method here?

Libraries/LibWeb/CSS/StyleValues/StyleValue.cpp
```diff
@@ -7,6 +7,7 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
+#include "StyleValue.h"
```
Calme1709: We already include this on line 69 - this is the correct syntax with the quotes though so please remove the other one.

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -906,156 +867,189 @@ void StyleComputer::collect_animation_into(DOM::AbstractElement abstract_element
...
+
+                if (!style_value) {
+                    specified_values.set(property_id, nullptr);
+                    continue;
+                }
 
-            for_each_property_expanding_shorthands(property_id, *style_value, [&](PropertyID longhand_id, StyleValue const& longhand_value) {
-                auto physical_longhand_id = map_logical_alias_to_physical_property(longhand_id, LogicalAliasMappingContext { computed_properties.writing_mode(), computed_properties.direction() });
-                auto physical_longhand_id_bitmap_index = to_underlying(physical_longhand_id) - to_underlying(first_longhand_property_id);
+                // If the style value is a PendingSubstitutionStyleValue we should skip it to avoid overwriting any value
```
Calme1709: Here and below: We should mark any comments in here that aren't part of the spec as `AD-HOC:` or `NOTE:`

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -906,156 +867,189 @@ void StyleComputer::collect_animation_into(DOM::AbstractElement abstract_element
...
-        if (auto const& font_size_specified_value = specified_values.get(PropertyID::FontSize); font_size_specified_value.has_value()) {
-            // FIXME: We need to respect the math-depth of this computed keyframe if it is present
-            auto computed_math_depth = computed_properties.math_depth();
-            auto inherited_font_size = inheritance_parent_has_computed_properties ? inheritance_parent->computed_properties()->font_size() : InitialValues::font_size();
-            auto inherited_math_depth = inheritance_parent_has_computed_properties ? inheritance_parent->computed_properties()->math_depth() : InitialValues::math_depth();
+                    longhands_set_by_property_id.set(physical_longhand_id, property_id);
+                    property_is_set_by_use_initial.set(physical_longhand_id_bitmap_index, is_use_initial);
+                    specified_values.set(physical_longhand_id, specified_value_with_css_wide_keywords_applied);
+                });
+                // FIXME: For logical properties [CSS-LOGICAL-1], add the equivalent physical properties [CSS-WRITING-MODES-4] based on the computed value of writing-mode and/or direction for the effect target.
```
Calme1709: I think this is handled within the `for_each_property_expanding_shorthand` lambda when we use `map_logical_alias_to_physical_property()`, so this doesn't need to be a FIXME

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -906,156 +867,189 @@ void StyleComputer::collect_animation_into(DOM::AbstractElement abstract_element
...
+            computed_keyframe.composite = keyframe->composite;
+            property_is_set_by_use_initial.fill(false);
+
+            // 2. For each property specified in keyframe:
+            for (auto [property_id, value] : keyframe->properties) {
+                // FIXME: we need this to pass test: Tests/LibWeb/Text/input/WebAnimations/misc/animating-all.html
+                if (property_id == PropertyID::AnimationTimingFunction) {
+                    continue;
+                }
+                // Compute a property value using the value specified on keyframe as the value, and the target element as the element; then add the property and resulting value to computed keyframe.
```
Calme1709: Here and below:  We try to wrap all comments as 120 characters

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1070,49 +1064,172 @@ void StyleComputer::collect_animation_into(DOM::AbstractElement abstract_element
...
-
-        if (!resolved_end_property) {
-            if (resolved_start_property) {
-                computed_properties.set_animated_property(it.key, *resolved_start_property);
-                dbgln_if(LIBWEB_CSS_ANIMATION_DEBUG, "No end property for property {}, using {}", string_from_property_id(it.key), resolved_start_property->to_string(SerializationMode::Normal));
+    // 5.3.4. The effect value of a keyframe effect
+    // https://www.w3.org/TR/web-animations-1/#the-effect-value-of-a-keyframe-animation-effect
+    auto keyframe_effect = [&](CSS::PropertyID target_property) {
+        AK::Variant<ValueComparingRefPtr<StyleValue const>, NoKeyframeEffect, KeyframeEffectInterpolateFailed> next_value = NoKeyframeEffect {};
+        // 1. If iteration progress is unresolved abort this procedure. (handled at the start)
```
Calme1709: Here and below:  Spec comments should be quoted verbatim and notes (i.e. "(handled at start)") should be added as `NOTE` or `AD-HOC` comments on the following line (i.e. `// NOTE: Handled at the start`)

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1070,49 +1064,172 @@ void StyleComputer::collect_animation_into(DOM::AbstractElement abstract_element
...
-        RefPtr resolved_end_property = computed_end_values.get(it.key).value_or(nullptr);
-
-        if (!resolved_end_property) {
-            if (resolved_start_property) {
-                computed_properties.set_animated_property(it.key, *resolved_start_property);
-                dbgln_if(LIBWEB_CSS_ANIMATION_DEBUG, "No end property for property {}, using {}", string_from_property_id(it.key), resolved_start_property->to_string(SerializationMode::Normal));
+    // 5.3.4. The effect value of a keyframe effect
+    // https://www.w3.org/TR/web-animations-1/#the-effect-value-of-a-keyframe-animation-effect
+    auto keyframe_effect = [&](CSS::PropertyID target_property) {
+        AK::Variant<ValueComparingRefPtr<StyleValue const>, NoKeyframeEffect, KeyframeEffectInterpolateFailed> next_value = NoKeyframeEffect {};
```
Calme1709: There doesn't seem to be much use for this `next_value` variable since we only ever assign to it immediately before returning, we can just instead return the values directly.

This may be done to explicitly set the return type but that would be better done using the `-> Type` syntax i.e. changing the line above to

```c++
auto keyframe_effect = [&](CSS::PropertyID target_property) -> Variant<ValueComparingRefPtr<StyleValue const>, NoKeyframeEffect, KeyframeEffectInterpolateFailed> {
```

It would also be nicer to return a `NonnullRefPtr` instead of a `RefPtr` since we know it's always non-null

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1070,49 +1064,172 @@ void StyleComputer::collect_animation_into(DOM::AbstractElement abstract_element
...
-    for (auto const& it : computed_start_values) {
-        auto resolved_start_property = it.value;
-        RefPtr resolved_end_property = computed_end_values.get(it.key).value_or(nullptr);
-
-        if (!resolved_end_property) {
-            if (resolved_start_property) {
-                computed_properties.set_animated_property(it.key, *resolved_start_property);
-                dbgln_if(LIBWEB_CSS_ANIMATION_DEBUG, "No end property for property {}, using {}", string_from_property_id(it.key), resolved_start_property->to_string(SerializationMode::Normal));
+    // 5.3.4. The effect value of a keyframe effect
+    // https://www.w3.org/TR/web-animations-1/#the-effect-value-of-a-keyframe-animation-effect
```
Calme1709: We prefer draft links: https://drafts.csswg.org/TR/web-animations-1/#the-effect-value-of-a-keyframe-animation-effect

JeremiasRy:
With counting the regressions in this change passes +7 tests, I'll see if I can reduce the regressions.

EDIT: +13 wpt tests

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1070,49 +1064,172 @@ void StyleComputer::collect_animation_into(DOM::AbstractElement abstract_element
...
+                }
+            }
 
-        auto const& underlying_value = computed_properties.property(it.key);
-        if (auto composited_start_value = composite_value(underlying_value, start, start_composite_operation))
-            start = *composited_start_value;
-
-        if (auto composited_end_value = composite_value(underlying_value, end, end_composite_operation))
-            end = *composited_end_value;
+            if (interval_endpoints.is_empty()) {
```
Calme1709: Assuming that

a) `property specified keyframes` are always in ascending order and;
b) a keyframe's `computed keyframe offset` is always equal to or greater than zero

We don't need this to be a separate loop: we know that if we come across a keyframe with a computed offset of 0 which is still greater than our iteration_progress in the above loop, none of the remaining keyframes will be less than or equal to it either, and we can just use this first keyframe we come across i.e. update the condition above to instead be

```c++
if ((keyframe.keyframe_offset_value < 1 && keyframe.keyframe_offset_value <= iteration_progress) || keyframe.keyframe_offset_value == 0)
```

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1070,49 +1064,172 @@ void StyleComputer::collect_animation_into(DOM::AbstractElement abstract_element
...
+            return next_value;
+        }
+        // 4. If the keyframe effect does not have an effect target, or if the effect target cannot have computed property values calculated, abort this procedure.
+        if (effect->target() == nullptr) {
+            return next_value;
+        }
+
+        // 5. Define the neutral value for composition as a value which, when combined with an underlying value using the add composite operation, produces the underlying value.
+        auto const& underlying_value = computed_properties.property(target_property);
+        auto neutral_value = StyleValue::get_neutral_value_for_type(underlying_value.type());
```
Calme1709: It would be ad-hoc but I think that we can avoid the entire concept of a `neutral value for composition` by instead using a composite operation of `replace` and `underlying value` instead of `add` and `neutral value for composition` respectively in the generated implicit keyframes (step 9 and 10) below - see w3c/csswg-drafts#2914

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1070,49 +1064,172 @@ void StyleComputer::collect_animation_into(DOM::AbstractElement abstract_element
...
+        if (iteration_progress < 0 && has_offset_zero > 1) {
+            // If iteration progress < 0 and there is more than one keyframe in property-specific keyframes with a computed keyframe offset of 0,
+            // Add the first keyframe in property-specific keyframes to interval endpoints.
+            interval_endpoints.append(property_specific_keyframes.first());
+        } else if (iteration_progress >= 1 && has_offset_one > 1) {
+            // If iteration progress ≥ 1 and there is more than one keyframe in property-specific keyframes with a computed keyframe offset of 1,
+            // Add the last keyframe in property-specific keyframes to interval endpoints.
+            interval_endpoints.append(property_specific_keyframes.last());
+        } else {
+            for (int current = property_specific_keyframes.size() - 2; current >= 0; --current) {
```
Calme1709, line 906: We look to be missing an "Otherwise" comment here

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1070,49 +1064,172 @@ void StyleComputer::collect_animation_into(DOM::AbstractElement abstract_element
...
-            computed_properties.set_animated_property(PropertyID::Visibility, KeywordStyleValue::create(Keyword::Hidden));
+                interval_endpoints.append(property_specific_keyframes[start_keyframe_index]);
+                interval_endpoints.append(property_specific_keyframes[start_keyframe_index + 1]);
+            }
         }
+        // 13. For each keyframe in interval endpoints:
+        for (auto& keyframe : interval_endpoints) {
+            // 1. If keyframe has a composite operation that is not replace, or keyframe has no composite operation and the composite operation of this keyframe effect is not replace, then perform the following steps:
+            if (keyframe.composite != Bindings::CompositeOperationOrAuto::Replace || (keyframe.composite == Bindings::CompositeOperationOrAuto::Auto && effect->composite() != Bindings::CompositeOperation::Replace)) {
+                Bindings::CompositeOperation composite_operation = to_composite_operation(keyframe.composite);
```
Calme1709, line 928: Missing comment for step 13.1.1

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1070,49 +1064,172 @@ void StyleComputer::collect_animation_into(DOM::AbstractElement abstract_element
...
+                // 2. Let value to combine be the property value of target property specified on keyframe.
+                auto value_to_combine = keyframe.properties.get(target_property).value().get_pointer<NonnullRefPtr<CSS::StyleValue const>>();
+                // 3. Replace the property value of target property on keyframe with the result of combining underlying value (Va) and value to combine (Vb) using the procedure for the composite operation to use corresponding to the target property’s animation type.
+                auto composed_style_value = composite_value(underlying_value, *value_to_combine, composite_operation);
+                Variant<Animations::KeyframeEffect::KeyFrameSet::UseInitial, NonnullRefPtr<CSS::StyleValue const>> replace_value = Animations::KeyframeEffect::KeyFrameSet::UseInitial {};
```
Calme1709: No need for this variable - we can just pass `composed_style_value.release_nonnull()` to `keyframe.properties.set()` directly.

Libraries/LibWeb/CSS/StyleComputer.h
```diff
@@ -76,6 +76,9 @@ class CountingBloomFilter {
     CounterType m_buckets[bucket_count];
 };
 
+struct NoKeyframeEffect { };
```
Calme1709: We could use the standard AK::Empty struct rather than introducing a new one.

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1070,49 +1064,172 @@ void StyleComputer::collect_animation_into(DOM::AbstractElement abstract_element
...
+
+                if (composed_style_value) {
+                    replace_value = composed_style_value.release_nonnull();
+                    keyframe.properties.set(target_property, move(replace_value));
+                }
+            }
+        }
+        // 14. If there is only one keyframe in interval endpoints return the property value of target property on that keyframe.
+        if (interval_endpoints.size() == 1) {
+            auto refpointer = *interval_endpoints.first().properties.get(target_property).value().get_pointer<NonnullRefPtr<CSS::StyleValue const>>();
```
Calme1709: If we update the return type of the function to be `NonnullRefPtr` rather than `RefPtr` we can avoid this dereferencing + creating a RefPtr by using the `get` method rather than the `get_pointer` method

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1070,49 +1064,172 @@ void StyleComputer::collect_animation_into(DOM::AbstractElement abstract_element
...
+        }
+        // 15. Let start offset be the computed keyframe offset of the first keyframe in interval endpoints.
+        double start_offset = interval_endpoints.first().keyframe_offset_value * 100 * Animations::KeyframeEffect::AnimationKeyFrameKeyScaleFactor;
+        // 16. Let end offset be the computed keyframe offset of last keyframe in interval endpoints.
+        double end_offset = interval_endpoints.last().keyframe_offset_value * 100 * Animations::KeyframeEffect::AnimationKeyFrameKeyScaleFactor;
+        // 17. Let interval distance be the result of evaluating (iteration progress - start offset) / (end offset - start offset).
+        double interval_distance = (round(iteration_progress * 100 * Animations::KeyframeEffect::AnimationKeyFrameKeyScaleFactor) - start_offset) / (end_offset - start_offset);
+        // 18. Let transformed distance be the result of evaluating the timing function associated with the first keyframe in interval endpoints passing interval distance as the input progress.
+
+        CSS::EasingFunction timing_function = effect->default_keyframe_timing_function();
```
Calme1709, line 951: The process you have here for resolving the timing function associated with a keyframe (i.e. looking at the AnimationTimingFunction property) only applies to CSS animations - web animations have a different mechanism (the easing property).

We should determine that before this point so I think that's out of scope for this PR and a FIXME will suffice

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1070,49 +1064,172 @@ void StyleComputer::collect_animation_into(DOM::AbstractElement abstract_element
...
+            dbgln_if(LIBWEB_CSS_ANIMATION_DEBUG, "Interpolated value for property {} at {}: {} -> {} = {}", string_from_property_id(target_property), transformed_distance, **v_start, **v_end, interpolated_value->to_string(SerializationMode::Normal));
+            return next_value;
+        }
+        dbgln_if(LIBWEB_CSS_ANIMATION_DEBUG, "Interpolated value for property {} at {}: {} -> {} is invalid", string_from_property_id(target_property), transformed_distance, v_start, v_end);
+        next_value = KeyframeEffectInterpolateFailed {};
+        return next_value;
+    };
+
+    for (auto target_property : target_properties) {
+        AK::Variant<ValueComparingRefPtr<StyleValue const>, NoKeyframeEffect, KeyframeEffectInterpolateFailed> next_value = keyframe_effect(target_property);
```
Calme1709: This could be `auto`

Libraries/LibWeb/Animations/KeyframeEffect.h
```diff
@@ -70,6 +70,13 @@ class KeyframeEffect final : public AnimationEffect {
             HashMap<CSS::PropertyID, Variant<UseInitial, NonnullRefPtr<CSS::StyleValue const>>> properties {};
             Bindings::CompositeOperationOrAuto composite { Bindings::CompositeOperationOrAuto::Auto };
         };
+
+        struct ComputedKeyframe {
+            HashMap<CSS::PropertyID, Variant<UseInitial, NonnullRefPtr<CSS::StyleValue const>>> properties {};
```
Calme1709: We will only ever have `NonnullRefPtr<StyleValue const>` here so this doesn't need to be a Variant including `UseInitial`

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1121,11 +1238,10 @@ static void apply_animation_properties(DOM::Document const& document, ComputedPr
     VERIFY(animation.effect());
 
     auto& effect = as<Animations::KeyframeEffect>(*animation.effect());
-
     effect.set_iteration_duration(animation_properties.duration);
     effect.set_start_delay(animation_properties.delay);
     effect.set_iteration_count(animation_properties.iteration_count);
-    effect.set_timing_function(animation_properties.timing_function);
+    effect.set_default_keyframe_timing_function(animation_properties.timing_function);
```
Calme1709, line 996: We should still set `timing_function` since it is observable e.g. through client side JS calling `AnimationEffect::get_{computed_?}timing`

What is the distinction between `timing_function` and `default_keyframe_timing_function`?

Libraries/LibWeb/CSS/StyleComputer.cpp
Psychpsyo: Alternatively to `NOTE:`, `NB:` also works and doesn't conflict with spec `NOTE:`s.

Libraries/LibWeb/CSS/StyleComputer.cpp
Psychpsyo: > We should still set `timing_function` since it is observable e.g. through client side JS calling `AnimationEffect::get_{computed_?}timing`
> 
> What is the distinction between `timing_function` and `default_keyframe_timing_function`?

The timing function given here needs to not apply to the animation itself, but to all keyframes instead.
The animation's timing function never gets set from CSS, only ever the one on all the keyframes.

Calme1709: Ah that makes sense, as I noted in a comment further up I think that keyframe specific timing functions should be determined earlier in the process and stored on the keyframes themselves so we will eventually remove this - maybe add a FIXME to remind us to do that?

Libraries/LibWeb/CSS/StyleComputer.cpp
JeremiasRy: Thanks, I'll add them in.

Libraries/LibWeb/CSS/StyleComputer.cpp
JeremiasRy: You are correct, it is much cleaner to just define the return type. `NonnullRefPtr` also makes sense.

Libraries/LibWeb/CSS/StyleComputer.cpp
JeremiasRy: Nice! Thanks for the simplification. 🙂

Libraries/LibWeb/CSS/StyleComputer.cpp
JeremiasRy: Good to know! Thanks.

Libraries/LibWeb/CSS/StyleComputer.h
JeremiasRy: Let's use it, no reason to create more empty structs.

Libraries/LibWeb/CSS/StyleComputer.cpp
JeremiasRy: This was actually my initial approach; resolve the timing function prior to `collect_animation_into`, but I ran into `const` problems, and resolving those made me modify a lot of function signatures that I felt I shouldn't be touching.

I'll add in the FIXME 👍

Libraries/LibWeb/Animations/KeyframeEffect.h
JeremiasRy: Oh nice catch! Thanks.

Libraries/LibWeb/CSS/StyleComputer.cpp
JeremiasRy: > I think that keyframe specific timing functions should be determined earlier in the process

As I also stated above in the reply to your comment above, this was my initial approach, but it resulted in bigger changes than I was hoping for, so I defaulted to this much simpler approach.

I added in the FIXME so that it can be cleaned up later.

Libraries/LibWeb/CSS/StyleComputer.cpp
JeremiasRy: I think this is a great idea, I added it in. It also brought the newly passing wpt count to 13 🙂

Libraries/LibWeb/CSS/StyleComputer.cpp
JeremiasRy: This is good to know, Thanks!

JeremiasRy:
Thanks @Calme1709 for the thorough review, really helps me to understand more of the coding style of the project. 

I've addressed all your comments.

(Also I hope I didn't step any toes implementing this.. It's not fun to work on a feature and see a PR appear suddenly implementing the exact same thing 😀)

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -906,156 +867,190 @@ void StyleComputer::collect_animation_into(DOM::AbstractElement abstract_element
...
+                // For shorthand properties, add the equivalent longhand properties.
+                for_each_property_expanding_shorthands(property_id, *style_value, [&](PropertyID longhand_id, StyleValue const& longhand_value) {
+                    auto physical_longhand_id = map_logical_alias_to_physical_property(longhand_id, LogicalAliasMappingContext { computed_properties.writing_mode(), computed_properties.direction() });
+                    auto physical_longhand_id_bitmap_index = to_underlying(physical_longhand_id) - to_underlying(first_longhand_property_id);
 
-                auto const& specified_value_with_css_wide_keywords_applied = [&]() -> StyleValue const& {
-                    if (longhand_value.is_inherit() || (longhand_value.is_unset() && is_inherited_property(longhand_id))) {
-                        if (auto inherited_animated_value = get_animated_inherit_value(longhand_id, abstract_element); inherited_animated_value.has_value())
-                            return inherited_animated_value.value();
+                    // Don't overwrite values if this is the result of a UseInitial
```
Calme1709: Should be NB

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -906,156 +867,190 @@ void StyleComputer::collect_animation_into(DOM::AbstractElement abstract_element
...
+                    if (specified_values.contains(physical_longhand_id) && specified_values.get(physical_longhand_id) != nullptr && is_use_initial)
+                        return;
 
-                        return get_non_animated_inherit_value(longhand_id, abstract_element);
-                    }
+                    // Don't overwrite unless the value was originally set by a UseInitial or
```
Calme1709: Should be NB

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1070,49 +1065,148 @@ void StyleComputer::collect_animation_into(DOM::AbstractElement abstract_element
...
-        RefPtr resolved_end_property = computed_end_values.get(it.key).value_or(nullptr);
-
-        if (!resolved_end_property) {
-            if (resolved_start_property) {
-                computed_properties.set_animated_property(it.key, *resolved_start_property);
-                dbgln_if(LIBWEB_CSS_ANIMATION_DEBUG, "No end property for property {}, using {}", string_from_property_id(it.key), resolved_start_property->to_string(SerializationMode::Normal));
+    // 5.3.4. The effect value of a keyframe effect
+    // https://drafts.csswg.org/web-animations-1/#the-effect-value-of-a-keyframe-animation-effect
+    auto keyframe_effect = [&](CSS::PropertyID target_property) -> AK::Variant<NonnullRefPtr<StyleValue const>, Empty, KeyframeEffectInterpolateFailed> {
+        // 1. If iteration progress is unresolved abort this procedure. "(handled at the start)"
```
Calme1709: Can we move "(handled at the start)" onto a newline and marked as NB?

```c++
// 1. If iteration progress is unresolved abort this procedure.
// NB: handled at the start
```

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1070,49 +1065,148 @@ void StyleComputer::collect_animation_into(DOM::AbstractElement abstract_element
...
+        auto iteration_progress = output_progress.value();
+        // 2. Let target property be the longhand property for which the effect value is to be calculated "(function parameter)".
```
Calme1709: Same as above - moved down to following line

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1070,49 +1065,148 @@ void StyleComputer::collect_animation_into(DOM::AbstractElement abstract_element
...
+        }
+        // 4. If the keyframe effect does not have an effect target, or if the effect target cannot have computed property values calculated, abort this procedure.
+        if (effect->target() == nullptr) {
+            return Empty {};
+        }
+
+        // 5. Define the neutral value for composition as a value which, when combined with an underlying value using the add composite operation, produces the underlying value.
+        // AD-HOC: To reduce complexity we'll use composite operation of REPLACE in conjunction with the underlying value
+        //         https://github.com/w3c/csswg-drafts/issues/2914
+        auto const& underlying_value = NonnullRefPtr<StyleValue const>(computed_properties.property(target_property));
```
Calme1709: Instead of explicitly constructing a NonnullRefPtr we can do it implicitly:
```c++
NonnullRefPtr<StyleValue const> const& underlying_value = computed_properties.property(target_property)
```

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1070,49 +1065,148 @@ void StyleComputer::collect_animation_into(DOM::AbstractElement abstract_element
...
+            if (keyframe.composite != Bindings::CompositeOperationOrAuto::Replace || (keyframe.composite == Bindings::CompositeOperationOrAuto::Auto && effect->composite() != Bindings::CompositeOperation::Replace)) {
+                // 1. Let composite operation to use be the composite operation of keyframe, or if it has none, the composite operation of this keyframe effect.
+                Bindings::CompositeOperation composite_operation = to_composite_operation(keyframe.composite);
+                // 2. Let value to combine be the property value of target property specified on keyframe.
+                auto value_to_combine = keyframe.properties.get(target_property).value();
+                // 3. Replace the property value of target property on keyframe with the result of combining underlying value (Va) and value to combine (Vb) using the procedure for the composite operation to use corresponding to the target property’s animation type.
+                auto composed_style_value = composite_value(underlying_value, *value_to_combine, composite_operation);
+                if (composed_style_value) {
```
Calme1709: We don't need to do the assignment on it's own line and can instead do

```c++
if (auto composed_style_value = composite_value(underlying_value, *value_to_combine, composite_operation))
	keyframe.properties.set(target_property, composed_style_value.release_nonnull());
```

Libraries/LibWeb/CSS/StyleComputer.h
```diff
@@ -76,6 +76,8 @@ class CountingBloomFilter {
     CounterType m_buckets[bucket_count];
 };
 
+struct KeyframeEffectInterpolateFailed { };
```
Calme1709: Since we only use this in one place we can just declare this inline, no need to have it in the header file.

Libraries/LibWeb/CSS/StyleComputer.cpp
Calme1709: There are still a couple comments that are over 120 chars (i.e. most of those within step 5.3.4)

Libraries/LibWeb/CSS/StyleComputer.cpp
Calme1709: We still need this comment since it's part of the spec, it just doesn't need to be FIXME

Libraries/LibWeb/CSS/StyleComputer.cpp
Calme1709: Could we bump these into their own comments rather than just adding quotes? We really want it to be clear at a glance what is and isn't part of the spec.

Libraries/LibWeb/CSS/StyleComputer.cpp
Calme1709: It would be nice to have some of these newly passing tests imported :)

Libraries/LibWeb/CSS/StyleComputer.cpp
Calme1709: Whoops look like I misguided you here, as you found you still need to create a `NonnullRefPtr` from the pointer returned by the `get` method, instead of constructing this directly we tend to use the `adopt_ref` method, i.e.

```c++
if (interval_endpoints.size() == 1)
	return adopt_ref(*interval_endpoints.first().properties.get(target_property).value());
```

Calme1709 (changes requested):
Thanks for looking at that so promptly. Absolutely no toes stepped on on my end - more than happy to have the help :)

A couple more comments, mainly minor nitpicks.

I also note that we have two commits in this PR, could you please squash them into one?

Libraries/LibWeb/CSS/StyleComputer.cpp
JeremiasRy: Oh of course.. I added it back in at the spot where you assumed it's handled.

JeremiasRy:
Thanks for the review, sorry for being a bit hasty in resolving your comments. I didn't fully follow your instructions previously, especially for the comment formatting, apologies for that and thanks for your patience. If I still missed something, just let me know :)

> I also note that we have two commits in this PR, could you please squash them into one?

Done!

I imported some of the passing WPT tests. (The import script seems to crash sometimes? Is this known or maybe it's something on my end?)

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1070,49 +1070,164 @@ void StyleComputer::collect_animation_into(DOM::AbstractElement abstract_element
...
-        RefPtr resolved_end_property = computed_end_values.get(it.key).value_or(nullptr);
-
-        if (!resolved_end_property) {
-            if (resolved_start_property) {
-                computed_properties.set_animated_property(it.key, *resolved_start_property);
-                dbgln_if(LIBWEB_CSS_ANIMATION_DEBUG, "No end property for property {}, using {}", string_from_property_id(it.key), resolved_start_property->to_string(SerializationMode::Normal));
+    // 5.3.4. The effect value of a keyframe effect
+    // https://drafts.csswg.org/web-animations-1/#the-effect-value-of-a-keyframe-animation-effect
+    struct KeyframeEffectInterpolateFailed { };
+    auto keyframe_effect = [&](CSS::PropertyID target_property) -> AK::Variant<NonnullRefPtr<StyleValue const>, Empty, KeyframeEffectInterpolateFailed> {
```
Calme1709: We can remove the `AK::` and `CSS::` namespaces here

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1070,49 +1070,164 @@ void StyleComputer::collect_animation_into(DOM::AbstractElement abstract_element
...
+        }
+        // 14. If there is only one keyframe in interval endpoints return the property value of target property on that keyframe.
+        if (interval_endpoints.size() == 1)
+            return adopt_ref(*interval_endpoints.first().properties.get(target_property).value());
+        // 15. Let start offset be the computed keyframe offset of the first keyframe in interval endpoints.
+        double start_offset = interval_endpoints.first().keyframe_offset_value * 100 * Animations::KeyframeEffect::AnimationKeyFrameKeyScaleFactor;
+        // 16. Let end offset be the computed keyframe offset of last keyframe in interval endpoints.
+        double end_offset = interval_endpoints.last().keyframe_offset_value * 100 * Animations::KeyframeEffect::AnimationKeyFrameKeyScaleFactor;
+        // 17. Let interval distance be the result of evaluating (iteration progress - start offset) / (end offset - start offset).
+        double interval_distance = (round(iteration_progress * 100 * Animations::KeyframeEffect::AnimationKeyFrameKeyScaleFactor) - start_offset) / (end_offset - start_offset);
```
Calme1709: Could we add an `AD-HOC:` comment explaining why this is rounded (as I can't see anything in your spec comments that suggests it should)

Calme1709 (changes requested):
Looks like that covers off all my original comments - I did notice two small things (apologies for not mentioning sooner) but apart from those this all looks good to me, it will of course require a review from a maintainer before being merged.

Regarding your issue with the WPT import script crashing, I personally am not aware of anything like that. Feel free to drop a message with the error you are facing in the #build-problems chat in Discord and someone should be able to help.

Libraries/LibWeb/CSS/StyleComputer.cpp
JeremiasRy: Good catch, I actually brought it over from the previous implementation without really examining it's purpose, seems that it's unnecessary.

JeremiasRy:
Thanks @Calme1709, for your time and the guidance here. This really helps for my future PRs to have a smoother review experience. 🙏 

Should I advertise the PR for maintainers or will they pick it up once they have time?

Calme1709:
Of course! Glad to have you working on the project!

Usually a maintainer will pick it up when they get a change but if you don't have any activity in a few days feel free to drop a message in the #code-review channel in Discord.

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -906,156 +869,193 @@ void StyleComputer::collect_animation_into(DOM::AbstractElement abstract_element
...
-            // already set by resolving the relevant shorthand's value.
-            if (style_value->is_pending_substitution())
-                continue;
-
-            if (style_value->is_unresolved())
-                style_value = Parser::Parser::resolve_unresolved_style_value(Parser::ParsingParams { abstract_element.document() }, abstract_element, PropertyNameAndID::from_id(property_id), style_value->as_unresolved());
+        // 2. For each keyframe in the list of keyframes specified on this keyframe effect, perform the following steps:
+        for (auto keyframe = keyframes.begin(); keyframe != keyframes.end(); ++keyframe) {
+            specified_values.clear();
+            longhands_set_by_property_id.clear();
```
AtkinsSJ: I think you're trying to avoid repeatedly allocating HashMaps by reusing them here, but `clear()` is basically the same as creating a new HashMap. I think `clear_with_capacity()` is what you want.

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1070,49 +1070,164 @@ void StyleComputer::collect_animation_into(DOM::AbstractElement abstract_element
...
+        NonnullRefPtr<StyleValue const> const& underlying_value = computed_properties.property(target_property);
+        // 6. Let property-specific keyframes be the result of getting the set of computed keyframes for this keyframe effect.
+        Vector<Animations::KeyframeEffect::KeyFrameSet::ComputedKeyframe> property_specific_keyframes;
+        // 7. Remove any keyframes from property-specific keyframes that do not have a property value for target property.
+        for (auto& keyframe : computed_keyframes) {
+            if (keyframe.properties.contains(target_property)) {
+                property_specific_keyframes.append(keyframe);
             }
-            continue;
         }
```
AtkinsSJ, line 854: Maybe put these spec steps together, with a note saying we're adding the ones we want instead of removing the ones we don't.

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1070,49 +1070,164 @@ void StyleComputer::collect_animation_into(DOM::AbstractElement abstract_element
...
-        if (!resolved_start_property || !resolved_end_property)
-            continue;
-
-        auto start = resolved_start_property.release_nonnull();
-        auto end = resolved_end_property.release_nonnull();
-
-        if (computed_properties.is_property_important(it.key)) {
-            continue;
+        // 8. If property-specific keyframes is empty, return underlying value.
+        if (property_specific_keyframes.size() == 0) {
```
AtkinsSJ: Nit: You can use `property_specific_keyframes.is_empty()`.

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1070,49 +1070,164 @@ void StyleComputer::collect_animation_into(DOM::AbstractElement abstract_element
...
+        // NB: function parameter
+        // 3. If animation type of the target property is not animatable abort this procedure since the effect cannot be applied.
+        if (!CSS::is_animatable_property(target_property)) {
+            return Empty {};
+        }
+        // 4. If the keyframe effect does not have an effect target, or if the effect target cannot have computed property
+        //    values calculated, abort this procedure.
+        if (effect->target() == nullptr) {
+            return Empty {};
+        }
```
AtkinsSJ: Super tiny nit: We often omit the `{}` for single-line statements like this, but it's up to you really.

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1070,49 +1070,164 @@ void StyleComputer::collect_animation_into(DOM::AbstractElement abstract_element
...
+        for (auto keyframe : property_specific_keyframes) {
+            if (keyframe.keyframe_offset_value == 0) {
+                ++has_offset_zero;
+            } else if (keyframe.keyframe_offset_value == 1) {
+                ++has_offset_one;
+            }
+        }
+        // 9. If there is no keyframe in property-specific keyframes with a computed keyframe offset of 0, create a new keyframe with
+        //    a computed keyframe offset of 0, a property value set to the neutral value for composition,
+        //    and a composite operation of add, and prepend it to the beginning of property-specific keyframes.
```
AtkinsSJ, line 871: I think this and step 10 could do with another note mentioning why we're replacing instead. Maybe just "see step 5" to avoid repeating yourself too much.

AtkinsSJ (changes requested):
Only tiny things, this looks good to me otherwise. :+1:

Libraries/LibWeb/CSS/StyleComputer.cpp
JeremiasRy: Yes this was the intent, didn't check the actual implementation of `clear()` and seems that it does exactly that. Thanks for pointing this out.

Libraries/LibWeb/CSS/StyleComputer.cpp
JeremiasRy: Not sure I'm following this correctly? Omitting the `{}` causes a compiler error, do you mean to omit the `Empty` and just use `{}`? Does it cause any problems with that we have 2 different empty structs (`Empty` and `KeyframeEffectInterpolateFailed`) as possible return values?

Also as you pointed super tiny nit, so maybe we just let it be?

JeremiasRy:
> Only tiny things, this looks good to me otherwise. 👍

Thanks for the prompt review! I addressed your comments.

Libraries/LibWeb/CSS/StyleComputer.cpp
AtkinsSJ: Oh sorry, I picked a poor set of lines for this. I meant for the if statement's block, like:
```c++
if (foo)
    bar();
```
It's not something you need to change, just letting you know really.

JeremiasRy: Ah yes! Ok I'll keep it mind for the future, also seems that a wpt test I imported doesn't pass, so I'll fix it and change these at the same time :)

JeremiasRy:
@AtkinsSJ, the test that was imported was a flaky pass that I happened to import, it had nothing to do with these changes. Could I request another CI run? Thanks!

(It was about `background-image` overwriting `background-color` and in ladybird sometimes the background color shows right before the image)

Tests/LibWeb/Ref/input/wpt-import/css/CSS2/backgrounds/background-image-applies-to-015.xht
```diff
@@ -0,0 +1,44 @@
...
+        <meta name="flags" content="image" />
+        <meta name="assert" content="The 'background-image' property applies to elements with 'display' set to 'table-caption'." />
+        <style type="text/css">
+            #table
+            {
+                display: table;
+            }
+            #caption
+            {
+                background-image: url('support/black15x15.png');
```
InvalidUsernameException: As you noticed, tests with the background-image property pointing to a URL are flaky in CI, which is because of #3448.

If I understand this correctly this is another test that is unrelated to your changes and could be removed.

If that is not the case and it should be kept as part of the PR, please add the following workaround to the end of the file to make sure this test is stable:

```html
<!-- FIXME: Workaround to ensure CSS background-image is loaded before taking screenshot: https://github.com/LadybirdBrowser/ladybird/issues/3448 -->
<img style="display: none" src="support/black15x15.png"/>
```

JeremiasRy: Thanks for jumping in, seems that I had a run that happened to pass few of these tests. I'll skip the import for these tests, since they are not related and we don't want to introduce flakiness to the CI.

PR #6825, opened 2025-11-14
LibWeb + View + UI: Add debug dump to file

michuhu:
While testing ladybird with some of the webpages I've noticed problems which required looking into DOM tree and other debug dumps. Using them dropped to the terminal was not very efficient, especially while working with big web apps and portals. I've added a feature to save those debug dumps to file(s) for easier human (and/or scripts) analysis . 

You can set the debug path in Settings. It defaults to users temporary folder using `Core::StandardPaths::tempfile_directory()`.
Changing the path is propagated to the backend. Files are successfully saved. Filenames reflect debug actions.

image

I hope the community finds it useful.

### DIFF
--- a/Libraries/LibWeb/Dump.cpp
+++ b/Libraries/LibWeb/Dump.cpp
@@ -6,9 +6,14 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
+#include <AK/LexicalPath.h>
 #include <AK/QuickSort.h>
 #include <AK/StringBuilder.h>
+#include <AK/Time.h>
 #include <AK/Utf8View.h>
+#include <LibCore/Directory.h>
+#include <LibCore/File.h>
+#include <LibCore/StandardPaths.h>
 #include <LibWeb/CSS/CSSDescriptors.h>
 #include <LibWeb/CSS/CSSFontFaceRule.h>
 #include <LibWeb/CSS/CSSImportRule.h>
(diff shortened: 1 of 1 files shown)


Base/res/ladybird/about-pages/settings.html
```diff
@@ -449,6 +449,16 @@ <h1>Ladybird Settings</h1>
...
 
+        <div class="card">
+            <div class="card-header">Debug settings</div>
+            <div class="card-body">
+                <div class="card-group">
+                    <label for="debug-dump-path">Path to debug dumps</label>
+                    <input id="debug-dump-path" type="text" />
+                </div>
+            </div>
+        </div>
```
gmta: Personally, I don't think we should add debug settings to the settings page at all - we would need to factor it out in the future and the benefit is arguably small. Defaulting to `Core::StandardPaths::tempfile_directory()` would be absolutely fine for me.

Libraries/LibWeb/Dump.cpp
```diff
@@ -61,6 +65,46 @@ static void indent(StringBuilder& builder, int levels)
...
+static String s_debug_dump_path;
+
+void set_debug_dump_path(StringView path)
+{
+    s_debug_dump_path = String::from_utf8(path).release_value_but_fixme_should_propagate_errors();
+}
+
+static void dump_to_file(StringBuilder& builder, StringView filename)
+{
+    ByteString dump_path = s_debug_dump_path.to_byte_string();
```
gmta: To reiterate, just using `Core::StandardPaths::tempfile_directory()` here would have been fine with me.

Libraries/LibWeb/Dump.cpp
```diff
@@ -61,6 +65,46 @@ static void indent(StringBuilder& builder, int levels)
...
+    }
+
+    // Create directory if it doesn't exist
+    auto create_result = Core::Directory::create(dump_path, Core::Directory::CreateDirectories::Yes);
+    if (create_result.is_error()) {
+        dbgln("Failed to create dump directory {}: {}", dump_path, create_result.error());
+        return;
+    }
+
+    auto full_path = LexicalPath(dump_path).append(filename).string();
```
gmta: Looks like we're overwriting files on repeated invocations - maybe we could make the filename more unique by appending a date and time, for example?

Base/res/ladybird/about-pages/settings.html
michuhu: Thanks @gmta for picking up the PR.
Ok, how about I add a toggle button in settings that would enable debug settings and debug menu in the top menu? In the future, this could work alike "Developer tools" in other browsers. Waiting for your opinion.

Libraries/LibWeb/Dump.cpp
michuhu: ok, waiting for the feedback on top comment.

Libraries/LibWeb/Dump.cpp
michuhu: You're right, I'll add date and time to filename.

lpas: If the file name changes you should maybe debug print the filename or path on success.

Base/res/ladybird/about-pages/settings.html
awesomekling: For internal engine debugging stuff, I agree with gmta that we don't want a bunch of UI. Better to just default to dumping in the standard temp directory.

And for now, I think we'd rather just always have the debug menu present. It's still a long time until we're in good-enough shape to ship something without debugging features 😅

michuhu:
I've added the changes @gmta and @awesomekling decided on. Moreover, I've added a timestamp to the filename and printing path on a successful dump.

PR #6829, opened 2025-11-15
LibWeb: Prevent race condition when setting up parent window listeners

lpas:
This fixes the sporadic timeouts of the gamepad-iframe test. 

I ran this test 300 times locally and got 9 timeouts before this fix and 0 after.

### DIFF
--- a/Tests/LibWeb/Text/input/GamepadAPI/gamepad-iframe.html
+++ b/Tests/LibWeb/Text/input/GamepadAPI/gamepad-iframe.html
@@ -30,13 +30,21 @@
 
         const sendMessageAndWait = (message) => {
             return new Promise((resolve) => {
-                window.onmessage = ({ data }) => {
+                const listener = ({ data }) => {
+                    window.removeEventListener('message', listener);
                     resolve(data);
                 };
-
+                window.addEventListener('message', listener);
                 testIframe.contentWindow.postMessage(message, "*");
             });
         };
+
+        // wait for the iframe to be ready
+        await new Promise((resolve) => {
+            testIframe.addEventListener('load', resolve);
+            if (testIframe.contentDocument.readyState === 'complete')
+                resolve();
+        });
 
         const gamepad = internals.connectVirtualGamepad();
         await handleSDLInputEvents();


Tests/LibWeb/Text/input/GamepadAPI/gamepad-iframe.html
```diff
@@ -28,16 +34,35 @@
...
+                    window.removeEventListener('message', listener);
                     resolve(data);
                 };
-
+                window.addEventListener('message', listener);
                 testIframe.contentWindow.postMessage(message, "*");
             });
         };
 
+        await waitForIframeReady();
```
gmta: If I understand correctly, you're effectively trying to guarantee that the `<script>` inside the iframe has run at this point. Wouldn't the following also work?

```js
        await new Promise((resolve) => {
            testIframe.onload = () => resolve();
        });
```

lpas: I'm not super sure. But in theory the load event could already be fired before we attach out listener that's the reason I did the "double tap" handshake. 

The code does iframe loads -> sends message -> event handler loads -> sends message -> iframe sends message again -> start

or event handler loads -> sends message -> iframe loads -> sends message -> start

Thus either way the iframe should be ready here. and we should not wait indefinitely

gmta: Eh sure, this should also work then?

```js
        await new Promise((resolve) => {
            testIframe.addEventListener('load', () => resolve());
            if (testIframe.contentDocument.readyState === 'complete')
                resolve();
        });
```

lpas: from my gut feeling this should work as well yes. But I would not be totally confident that we trigger the events at the right time and we really waited for the script in the iframe to be loaded parsed and be ready. 

You want easier to understand code for the change? 

I guess I was just annoyed by flakyness I saw from this test in CI and wanted a bullet proof fix :)

gmta: > You want easier to understand code for the change?

It took some time to understand what you were doing, yeah.

> But I would not be totally confident that we trigger the events at the right time and we really waited for the script in the iframe to be loaded parsed and be ready.

What in my example would not give you confidence that things are executed in the right order?

lpas: > What in my example would not give you confidence that things are executed in the right order?

Nothing. Before I wrote this I just had a quick loot at https://github.com/LadybirdBrowser/ladybird/blob/master/Libraries/LibWeb/HTML/HTMLIFrameElement.cpp#L228 
and this looked a little bare bone I didn't check further up if we do all the right things already. 

I'm happy to change this to your code. And if it is still flaky we can revisit this.

lpas: I adjusted the code and ran it locally, looks good.

Tests/LibWeb/Text/input/GamepadAPI/gamepad-iframe.html
```diff
@@ -30,14 +30,22 @@
...
                 };
-
+                window.addEventListener('message', listener);
                 testIframe.contentWindow.postMessage(message, "*");
             });
         };
 
+        // wait for the iframe to be ready
+        await new Promise((resolve) => {
+            testIframe.addEventListener('load', () => resolve());
```
gmta: Not sure why I wrote it like this, but should probable be this 😅 

```js
testIframe.addEventListener('load', resolve);
```

lpas: done

Tests/LibWeb/Text/input/GamepadAPI/gamepad-iframe.html
```diff
@@ -30,14 +30,22 @@
```
gmta, line 30: Commit title: nothing in it suggests a test change; maybe mark it as such?

E.g.
```
Tests/LibWeb: Prevent flake with iframe loading in gamepad-iframe.html
```

+ optionally an explanation of what was going on in the commit message.

gmta (approved):

PR #6830, opened 2025-11-15
Two HTTP cache improvements

awesomekling:
Two things here:

- Stop incorrectly overwriting Content-Length when revalidating.
- Implement a (pretty conservative) heuristic for freshness if there's no explicit expiration information present.

Net result: we cache more resources on many, many websites! Examples: github.com, archive.org

### DIFF
--- a/Services/RequestServer/Cache/Utilities.cpp
+++ b/Services/RequestServer/Cache/Utilities.cpp
@@ -84,6 +84,31 @@ bool is_cacheable(StringView method)
     return method.is_one_of("GET"sv, "HEAD"sv);
 }
 
+// https://datatracker.ietf.org/doc/html/rfc9110#name-overview-of-status-codes
+static bool is_heuristically_cacheable_status(u32 status_code)
+{
+    // Responses with status codes that are defined as heuristically cacheable
+    // (e.g., 200, 203, 204, 206, 300, 301, 308, 404, 405, 410, 414, and 501)
+    // can be reused by a cache with heuristic expiration [...]
+    switch (status_code) {
+    case 200:
+    case 203:
+    case 204:
+    case 206:
+    case 300:
+    case 301:
+    case 308:
+    case 404:
+    case 405:
+    case 410:
+    case 414:
+    case 501:
+        return true;
+    default:
+        return false;
+    }
+}
+
 // https://httpwg.org/specs/rfc9111.html#response.cacheability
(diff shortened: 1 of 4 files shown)


Services/RequestServer/Cache/Utilities.cpp
```diff
@@ -278,7 +278,7 @@ void update_header_fields(HTTP::HeaderMap& stored_headers, HTTP::HeaderMap const
         // * Header fields that are automatically processed and removed by the recipient, as described below, and
 
         // * The Content-Length header field.
-        if (name.equals_ignoring_ascii_case("Content-Type"sv))
+        if (name.equals_ignoring_ascii_case("Content-Length"sv))
```
trflynn89, line 389: whoops!

Services/RequestServer/Cache/Utilities.cpp
```diff
@@ -94,21 +119,28 @@ bool is_cacheable(u32 status_code, HTTP::HeaderMap const& headers)
...
 
     // * the no-store cache directive is not present in the response (see Section 5.2.2.5);
-    if (cache_control.has_value() && cache_control->contains("no-store"sv, CaseSensitivity::CaseInsensitive))
+    if (cache_control.has_value()
+        && cache_control->contains("no-store"sv, CaseSensitivity::CaseInsensitive))
         return false;
 
-    // * if the cache is shared: the private response directive is either not present or allows a shared cache to store
-    //   a modified response; see Section 5.2.2.7);
+    // * if the cache is shared: the private response directive is either not present
```
trflynn89: :thinking: why the comment wrapping changes?

Services/RequestServer/Cache/Utilities.cpp
```diff
@@ -179,9 +273,24 @@ AK::Duration calculate_freshness_lifetime(HTTP::HeaderMap const& headers)
...
 
+    bool heuristics_allowed = false;
+
+    // Because of the requirements in Section 3, heuristics can only be used on responses without explicit freshness
+    // whose status codes are defined as heuristically cacheable and on responses without explicit freshness that have
+    // been marked as explicitly cacheable (e.g., with a public response directive).
+    if (is_heuristically_cacheable_status(status_code))
+        heuristics_allowed = true;
+
+    if (cache_control.has_value()
```
trflynn89: This could be an `else if`

trflynn89 (approved):

PR #6826, opened 2025-11-14
LibWeb: Reimplement transform interpolation according to spec

gmta:
We had a partial implementation of transformation function interpolation that did not support numerical interpolation of simple functions (e.g. `scale(0)` -> `scale(1)`). This refactors the interpolation to follow the spec more closely.

Gains us 267 WPT subtest passes in `css/css-transforms`.

Fixes #6774.

CC: @fdellwing

PR #6836, opened 2025-11-16
LibWeb: Handle calc percentages when interpolating scale function

Calme1709:
Also clean up the corresponding handling in `TransformationStyleValue::to_transformation`


gmta:
@tcl3 thanks for the heads up, I could simply apply this patch to #6826 as well to make it work.

PR #6839, opened 2025-11-17
LibWeb: Use unbuffered network requests for all Fetch requests

kalenikaliaksandr:
Previously, unbuffered requests were only available as a special mode
for EventSource. With this change, they are enabled by default, which
means chunks can be read from the stream as soon as they arrive.

This unlocks some interesting possibilities, such as starting to parse
HTML documents before the entire response has been received (that, in
turn, allows us to initiate subresource fetches earlier or begin
executing scripts sooner), or start rendering videos before they are
fully downloaded.

Progress on https://chatgpt.com/:

Before:

https://github.com/user-attachments/assets/2afe1626-5b24-41ba-b477-0d8d4eff791b

After:

https://github.com/user-attachments/assets/2b9848e1-02ad-4980-8050-4c12eae25d2c

### DIFF
--- a/Libraries/LibWeb/Fetch/Infrastructure/HTTP/Requests.cpp
+++ b/Libraries/LibWeb/Fetch/Infrastructure/HTTP/Requests.cpp
@@ -264,7 +264,6 @@ GC::Ref<Request> Request::clone(JS::Realm& realm) const
     new_request->set_prevent_no_cache_cache_control_header_modification(m_prevent_no_cache_cache_control_header_modification);
     new_request->set_done(m_done);
     new_request->set_timing_allow_failed(m_timing_allow_failed);
-    new_request->set_buffer_policy(m_buffer_policy);
 
     // 2. If request’s body is non-null, set newRequest’s body to the result of cloning request’s body.
     if (auto const* body = m_body.get_pointer<GC::Ref<Body>>())
--- a/Libraries/LibWeb/Fetch/Infrastructure/HTTP/Requests.h
+++ b/Libraries/LibWeb/Fetch/Infrastructure/HTTP/Requests.h
@@ -159,13 +159,6 @@ public:
         Auto
     };
 
-    // AD-HOC: Some web features need to receive data as it arrives, rather than when the response is fully complete
-    //         or when enough data has been buffered. Use this buffer policy to inform fetch of that requirement.
-    enum class BufferPolicy {
-        BufferResponse,
-        DoNotBufferResponse,
-    };
-
     // Members are implementation-defined
     struct InternalPriority { };
 
@@ -334,9 +327,6 @@ public:
     {
         m_pending_responses.remove_first_matching([&](auto gc_ptr) { return gc_ptr == pending_response; });
     }
-
-    [[nodiscard]] BufferPolicy buffer_policy() const { return m_buffer_policy; }
-    void set_buffer_policy(BufferPolicy buffer_policy) { m_buffer_policy = buffer_policy; }
 
 private:
     explicit Request(GC::Ref<HeaderList>);
@@ -532,8 +522,6 @@ private:
 
     // Non-standard
     Vector<GC::Ref<Fetching::PendingResponse>> m_pending_responses;
-
-    BufferPolicy m_buffer_policy { BufferPolicy::BufferResponse };
 };
 
 WEB_API StringView request_destination_to_string(Request::Destination);
--- a/Libraries/LibWeb/HTML/EventSource.cpp
+++ b/Libraries/LibWeb/HTML/EventSource.cpp
@@ -78,10 +78,6 @@ WebIDL::ExceptionOr<GC::Ref<EventSource>> EventSource::construct_impl(JS::Realm&
     // 12. Set request's initiator type to "other".
     request->set_initiator_type(Fetch::Infrastructure::Request::InitiatorType::Other);
 
-    // AD-HOC: We must not buffer the response as the connection generally never ends, thus we can't wait for the end
-    //         of the response body.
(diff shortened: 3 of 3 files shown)


Libraries/LibWeb/Loader/ResourceLoader.cpp
```diff
@@ -184,6 +184,85 @@ static bool should_block_request(LoadRequest const& request)
...
 }
 
+struct FileLoadResult {
+    ReadonlyBytes data;
+    HTTP::HeaderMap response_headers;
+    Requests::RequestTimingInfo timing_info;
+};
+
+template<typename LoadCounterChangeCallback, typename FileHandler, typename ErrorHandler>
+static void request_file_for_load(LoadRequest& request, int& pending_loads, LoadCounterChangeCallback& on_load_counter_change, FileHandler on_file, ErrorHandler on_error)
```
trflynn89: It feels a little icky the way we're passing references to `ResourceLoader`'s internals here and in subsequent commits. Could we just make this a private class function? The compiler can handle the template params fine as long as this is defined above its usage (which is the case already here).

(nit: I also find the phrasing of "request file for load", "request about for load" awkward to read)

This would look something like:
```c++
template<typename FileHandler, typename ErrorHandler>
void ResourceLoader::handle_file_load_request(LoadRequest const& request, FileHandler on_file, ErrorHandler on_error)
```

Libraries/LibWeb/HTML/Scripting/Environments.cpp
```diff
@@ -125,8 +125,15 @@ EventLoop& EnvironmentSettingsObject::responsible_event_loop()
 RunScriptDecision can_run_script(JS::Realm const& realm)
 {
     // 1. If the global object specified by realm is a Window object whose Document object is not fully active, then return "do not run".
-    if (is<HTML::Window>(realm.global_object()) && !as<HTML::Window>(realm.global_object()).associated_document().is_fully_active())
-        return RunScriptDecision::DoNotRun;
+    if (auto const* window = as_if<HTML::Window>(realm.global_object())) {
+        auto const& document = window->associated_document();
+        // AD-HOC: We allow tasks for destroyed documents to run so that microtasks
+        //         queued during the fetch of a new document in a navigation can still
```
trflynn89: Let's wrap at 120 chars

Libraries/LibWeb/Fetch/Fetching/FetchedDataReceiver.cpp
```diff
@@ -36,35 +37,89 @@ void FetchedDataReceiver::visit_edges(Visitor& visitor)
...
-    if (!had_pending_promise && !m_buffer.is_empty()) {
-        on_data_received(m_buffer);
-        m_buffer.clear();
+    if (!m_buffer.is_empty()) {
+        pull_bytes_into_stream(move(m_buffer));
+    } else if (m_lifecycle_state == LifecycleState::ReadyToClose) {
+        WebIDL::resolve_promise(m_stream->realm(), *m_pending_promise, JS::js_undefined());
+        m_pending_promise = {};
+        m_lifecycle_state = LifecycleState::Closed;
+        m_stream->close();
```
trflynn89: Given that these 4 steps are repeated verbatim below, and that we're tracking quite a bit here - might be nicer to put this in a `close_stream` helper function

Libraries/LibWeb/Fetch/Fetching/FetchedDataReceiver.cpp
```diff
@@ -36,35 +37,89 @@ void FetchedDataReceiver::visit_edges(Visitor& visitor)
...
+    VERIFY(m_lifecycle_state == LifecycleState::Receiving || m_lifecycle_state == LifecycleState::CompletePending);
+
     // 3. Queue a fetch task to run the following steps, with fetchParams’s task destination.
+    VERIFY(!m_has_unfulfilled_promise);
+    m_has_unfulfilled_promise = true;
     Infrastructure::queue_fetch_task(
         m_fetch_params->controller(),
         m_fetch_params->task_destination(),
-        GC::create_function(heap(), [this, bytes = MUST(ByteBuffer::copy(bytes))]() mutable {
+        GC::create_function(heap(), [this, bytes = MUST(ByteBuffer::copy(bytes)), pending_promise = m_pending_promise]() mutable {
```
trflynn89: Let's make this: `bytes = move(bytes)` now that we already have a ByteBuffer in this function

Libraries/LibWeb/Loader/ResourceLoader.h
```diff
@@ -29,12 +29,6 @@ class WEB_API ResourceLoader : public Core::EventReceiver {
 
     void set_client(NonnullRefPtr<Requests::RequestClient>);
 
-    using SuccessCallback = GC::Function<void(ReadonlyBytes, Requests::RequestTimingInfo const&, HTTP::HeaderMap const& response_headers, Optional<u32> status_code, Optional<String> const& reason_phrase)>;
-    using ErrorCallback = GC::Function<void(ByteString const&, Requests::RequestTimingInfo const&, Optional<u32> status_code, Optional<String> const& reason_phrase, ReadonlyBytes payload, HTTP::HeaderMap const& response_headers)>;
-    using TimeoutCallback = GC::Function<void()>;
-
-    void load(LoadRequest&, GC::Root<SuccessCallback> success_callback, GC::Root<ErrorCallback> error_callback = nullptr, Optional<u32> timeout = {}, GC::Root<TimeoutCallback> timeout_callback = nullptr);
```
trflynn89, line 36: nit: I think we should also rename `load_unbuffered` to `load` now that there's no alternative

PR #6841, opened 2025-11-17
Documentation+LibWeb: Standardize our spec note comment prefixes

gmta:
We have a couple of ways to designate spec notes and (our) developer notes in comments, but we never really settled on a single approach. As a result, we have a bit of a mixed bag of note comments on our hands.

To the extent that I could find them, I changed developer notes to `// NB: ...` and changed spec notes to `// NOTE: ...`. The rationale for this is that in most web specs, notes are prefixed by `NOTE: ...` so this makes it easier to copy paste verbatim. The choice for `NB: ...` is pretty arbitrary, but it makes it stand out from the regular spec notes and it was already in wide use in our codebase.

### DIFF
--- a/Libraries/LibWeb/Crypto/CryptoAlgorithms.h
+++ b/Libraries/LibWeb/Crypto/CryptoAlgorithms.h
@@ -146,7 +146,7 @@ struct RsaKeyGenParams : public AlgorithmParams {
     }
 
     u32 modulus_length;
-    // NOTE that the raw data is going to be in Big Endian u8[] format
+    // NOTE: The raw data is going to be in Big Endian u8[] format
     ::Crypto::UnsignedBigInteger public_exponent;
 
     static JS::ThrowCompletionOr<NonnullOwnPtr<AlgorithmParams>> from_value(JS::VM&, JS::Value);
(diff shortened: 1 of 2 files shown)


Libraries/LibWeb/DOM/Node.cpp
```diff
@@ -786,10 +786,10 @@ void Node::insert_before(GC::Ref<Node> node, GC::Ptr<Node> child, bool suppress_
     children_changed(&metadata);
 
     // 10. Let staticNodeList be a list of nodes, initially « ».
-    // Spec-Note: We collect all nodes before calling the post-connection steps on any one of them, instead of calling
-    //            the post-connection steps while we’re traversing the node tree. This is because the post-connection
-    //            steps can modify the tree’s structure, making live traversal unsafe, possibly leading to the
-    //            post-connection steps being called multiple times on the same node.
+    // NB: We collect all nodes before calling the post-connection steps on any one of them, instead of calling the
```
trflynn89: This one is a spec note: https://dom.spec.whatwg.org/#concept-node-insert

gmta: Fixed!

Documentation/CodingStyle.md
```diff
@@ -589,6 +589,31 @@ catdog_widget.on_click = [&] {
 };
 ```
 
+#### Spec notes
+
+Many web specs include notes that are prefixed with `NOTE: ...`. To allow for verbatim copying of these notes into our
+code, we retain the `NOTE:` prefix and use `NB:` for our own notes.
```
trflynn89: Whoops thought I left this comment yesterday:

We should probably write here that NB stands for the Latin phrase "nota bene". I'm not sure how common it is globally.

gmta: I've added a link to Wikipedia :)

PR #6844, opened 2025-11-17
LibWeb/CSS: Stop inserting whitespace when serializing SimpleBlocks

AtkinsSJ:
I missed this in #6646. It mostly affects debug logging, but it's nice when logs reflect reality.


AtkinsSJ:
Changes:
- Turns out this fixes a test! :tada: 
- Also found another place we were inserting whitespace. I *think* that's everywhere.

gmta (approved):

PR #6853, opened 2025-11-18
LibWeb: Don't mangle inherited properties when updating animated style

Calme1709:


Tests/LibWeb/Text/input/css/animations-dont-mangle-inherited-animated-properties.html
gmta, line 1: Most of the tests we have are prefixed by a (very loosely chosen) category; maybe use `animation-dont-mangle-inherited-animated-properties.html` for this?

Calme1709: That's a much better name - done.

PR #6854, opened 2025-11-18
LibWeb: Improvements to transition cancelling

Calme1709:
A collection of improvements to the cancelling of transitions when their associated `transition-*` properties change, see individual commits for details.

Unfortunately daab35cbeaa49277e516438f306cd57fdee9de9c comes with a decent performance penalty as it effectively renders dd9d6d2 ineffective as we now register transitions for all properties by default (since the initial value for `transition-property` is `all`) but that's something we can look at improving in the future.

CC @gmta this makes progress on your [repro from Discord](https://discord.com/channels/1247070541085671459/1247090082297548832/1440000882187173900) but still has a couple bugs, namely:
 - The style gets "stuck" when we cancel the transition (this is because we don't update "idle" animations in `update_animated_style_if_needed`)
 - The next transition for that property on that element isn't triggered, I'm leaning towards this being a spec bug but I will need to do more investigation.

### DIFF
--- a/Libraries/LibWeb/CSS/StyleComputer.cpp
+++ b/Libraries/LibWeb/CSS/StyleComputer.cpp
@@ -1231,6 +1231,25 @@ static void compute_transitioned_properties(ComputedProperties const& style, DOM
 
     element.clear_registered_transitions(pseudo_element);
 
+    auto const& delay = style.property(PropertyID::TransitionDelay);
+    auto const& duration = style.property(PropertyID::TransitionDuration);
+
+    // FIXME: Change this to support the associated StyleValueList values when we update
+    //        parse_simple_comma_separated_value_list to always return a StyleValueList.
+    // OPTIMIZATION: Registered transitions with a "combined duration" of less than or equal to 0s are equivalent to not
+    //               having a transition registered at all, except in the case that we already have an associated
+    //               transition for that property, so we can skip registering them. This implementation intentionally
+    //               ignores some of those cases (e.g. transitions being registered but for other properties, multiple
+    //               transitions, negative delays, etc) since it covers the common (initial property values) case and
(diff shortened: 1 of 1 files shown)


Calme1709:
Changes:
 - Rebased to master
 - Added a commit to mitigate the aforementioned performance hit

Libraries/LibWeb/Animations/Animatable.cpp
gmta, line 1: Commit message: "Unfortuantely" -> "Unfortunately"

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1231,6 +1231,25 @@ static void compute_transitioned_properties(ComputedProperties const& style, DOM
...
+    //        parse_simple_comma_separated_value_list to always return a StyleValueList.
+    // OPTIMIZATION: Registered transitions with a "combined duration" of less than or equal to 0s are equivalent to not
+    //               having a transition registered at all, except in the case that we already have an associated
+    //               transition for that property, so we can skip registering them. This implementation intentionally
+    //               ignores some of those cases (e.g. transitions being registered but for other properties, multiple
+    //               transitions, negative delays, etc) since it covers the common (initial property values) case and
+    //               the other cases are rare enough that the cost of identifying them would likely more than offset any
+    //               gains.
+    if (
+        element.property_ids_with_existing_transitions(pseudo_element).size() == 0
```
gmta: `.is_empty()`

gmta (commented):
Only nits!

gmta:
Maybe mention that #6870 is fixed in one of the commit messages or the PR description so it gets auto-closed.

Calme1709:
> Maybe mention that #6870 is fixed in one of the commit messages or the PR description so it gets auto-closed.

Have added to the commit message for the relevant commit

PR #6858, opened 2025-11-18
LibWeb: Prevent view transition crashes due to lack of execution context

Psychpsyo:
Fixes the crash on https://angular.dev/

### DIFF
--- a/Libraries/LibWeb/ViewTransition/ViewTransition.cpp
+++ b/Libraries/LibWeb/ViewTransition/ViewTransition.cpp
@@ -621,19 +621,18 @@ void ViewTransition::call_the_update_callback()
         // We need to do it here manually.
         // https://webidl.spec.whatwg.org/#js-promise
 
+        HTML::TemporaryExecutionContext context(realm, HTML::TemporaryExecutionContext::CallbacksEnabled::Yes);
         // 1. Let promiseCapability be ? NewPromiseCapability(%Promise%).
         auto promise_capability = WebIDL::create_promise(realm);
         // 2. Perform ? Call(promiseCapability.[[Resolve]], undefined, « V »).
-        // FIXME: We should not need to push an incumbent realm here, but http://wpt.live/css/css-view-transitions/update-callback-timeout.html crashes without it.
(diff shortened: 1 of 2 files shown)


Libraries/LibWeb/ViewTransition/ViewTransition.cpp
```diff
@@ -621,6 +621,7 @@ void ViewTransition::call_the_update_callback()
         // We need to do it here manually.
         // https://webidl.spec.whatwg.org/#js-promise
 
+        HTML::TemporaryExecutionContext context(realm);
```
trflynn89: If you pass in `CallbacksEnabled::Yes`, you should then not need to manually push/pop a realm below.

It looks like that FIXME was added in 56739b4b162f99ad2e1addbec179878175e2908d a couple months ago - in general, if you ever find yourself with an empty execution context or empty realm stack, that means you probably just need to bring in `TemporaryExecutionContext`.

Libraries/LibWeb/ViewTransition/ViewTransition.cpp
```diff
@@ -668,6 +671,7 @@ void ViewTransition::call_the_update_callback()
     // AD-HOC: This can cause an assertion failure when the reaction algorithm ends up accessing the incumbent realm, which may not exist here.
     //         For now, lets just manually push something onto the incumbent realm stack here as a hack.
     //         A spec bug for this has been filed at https://github.com/w3c/csswg-drafts/issues/11990
+    HTML::TemporaryExecutionContext context(realm);
```
trflynn89: Same here (`CallbacksEnabled::Yes`).

Libraries/LibWeb/ViewTransition/ViewTransition.cpp
Psychpsyo: Done that now and removed the FIXMEs since TemporaryExecutionContext is basically a FIXME itself. (at least judging by the comment in its header file)

Psychpsyo:
image
Guess I need to push those changes tomorrow.

PR #6861, opened 2025-11-18
LibWeb+RequestServer: Add some tests that exercise the HTTP disk cache

trflynn89:
Our HTTP disk cache is currently manually tested against various sites. This adds some tests to cover various scenarios, including non-cacheable responses, expired responses, and revalidation. This PR also fixes some issues uncovered by the tests.

The main goal here is to have the infra such that any features added to the disk cache going forward can be tested immediately.


Services/RequestServer/Cache/CacheIndex.cpp
```diff
@@ -18,9 +18,6 @@ static ByteString serialize_headers(HTTP::HeaderMap const& headers)
     StringBuilder builder;
```
gmta, line 18: Commit message: "expempted" -> "exempted"

Services/RequestServer/Cache/CacheIndex.cpp
```diff
@@ -116,6 +113,15 @@ void CacheIndex::create_entry(u64 cache_key, String url, HTTP::HeaderMap respons
...
     auto now = UnixDateTime::now();
 
+    for (size_t i = 0; i < response_headers.headers().size();) {
+        auto const& header = response_headers.headers()[i];
+
+        if (is_header_exempted_from_storage(header.name))
+            response_headers.remove(header.name);
+        else
+            ++i;
+    }
```
gmta, line 123: Using `response_headers.headers().remove_all_matching([](auto& header) { .. })` might be more idiomatic here.

trflynn89: `HeaderMap` stores the headers in a couple of different structures, so we have to go through `HeaderMap::remove` here.
(`HeaderMap::headers` also returns an immutable reference, probably to prevent the structures from getting out-of-sync)

PR #6862, opened 2025-11-18
Meta: Compile for the native architecture of the host system on Windows

stasoid:
PR #6210 broke my local Windows build: IPCCompiler.exe crashes with error 0xc0000142 because of illegal instruction in lagom-ak.dll. This is the fix. Note that clang-cl accepts many clang options, including -march=native.

This PR needs the windows label.

### DIFF
--- a/Meta/CMake/common_compile_options.cmake
+++ b/Meta/CMake/common_compile_options.cmake
@@ -49,10 +49,7 @@ macro(add_swift_link_options)
     add_link_options($<$<LINK_LANGUAGE:Swift>:${args}>)
 endmacro()
 
-if (MSVC)
-    # On Windows, we always target the same architecture since there is no -march=native equivalent.
-    add_cxx_compile_options(/arch:AVX2)
-elseif (ENABLE_CI_BASELINE_CPU)
+if (ENABLE_CI_BASELINE_CPU)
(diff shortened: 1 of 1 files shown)


Meta/CMake/common_compile_options.cmake
```diff
@@ -50,8 +50,8 @@ macro(add_swift_link_options)
 endmacro()
 
 if (MSVC)
-    # On Windows, we always target the same architecture since there is no -march=native equivalent.
-    add_cxx_compile_options(/arch:AVX2)
+    # On Windows, compile for the native architecture of the host system.
+    add_cxx_compile_options(-march=native)
```
gmta: This was meant as a default matching `ENABLE_CI_BASELINE_CPU`, which is now less the case by targeting the host system.

You could remove the `if (MSVC)` condition so it defaults to `-march=native` in the `else()`, and add a specific architecture in the `ENABLE_CI_BASELINE_CPU` case instead.

stasoid: >  and add a specific architecture in the ENABLE_CI_BASELINE_CPU case instead.

I think when ENABLE_CI_BASELINE_CPU is ON there is no need for special Windows treatment either, -march=x86-64-v3 does the job.

PR #6863, opened 2025-11-18
RequestServer: Support HTTP response write retries on Windows

ayeteadoe:
See [discussion](https://discord.com/channels/1247070541085671459/1306918361732616212/1440408549602889885) in `#ui-windows` and commit for details. 

### Before

https://github.com/user-attachments/assets/02ff0a4d-0043-41e4-aa9f-8892427ac4e7

### After

https://github.com/user-attachments/assets/22dedb72-7cf7-4af8-a793-06125e2cf317

### DIFF
--- a/Services/RequestServer/Request.cpp
+++ b/Services/RequestServer/Request.cpp
@@ -6,6 +6,7 @@
  */
 
 #include <AK/Enumerate.h>
+#include <AK/GenericShorthands.h>
 #include <LibCore/Notifier.h>
 #include <LibTextCodec/Decoder.h>
 #include <RequestServer/CURL.h>
@@ -614,7 +615,7 @@ ErrorOr<void> Request::write_queued_bytes_without_blocking()
 
     auto result = m_client_request_pipe->write(bytes_to_send);
     if (result.is_error()) {
-        if (result.error().code() != EAGAIN)
+        if (!first_is_one_of(result.error().code(), EAGAIN, EWOULDBLOCK))
             return result.release_error();
 
         m_client_writer_notifier->set_enabled(true);
--- a/Services/RequestServer/RequestPipe.cpp
+++ b/Services/RequestServer/RequestPipe.cpp
@@ -59,8 +59,11 @@ ErrorOr<ssize_t> RequestPipe::write(ReadonlyBytes bytes)
 {
 #if defined(AK_OS_WINDOWS)
     auto sent = ::send(m_writer_fd, reinterpret_cast<char const*>(bytes.data()), bytes.size(), 0);
-    if (sent == SOCKET_ERROR)
+    if (sent == SOCKET_ERROR) {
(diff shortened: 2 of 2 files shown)


Services/RequestServer/Request.cpp
```diff
@@ -614,7 +614,10 @@ ErrorOr<void> Request::write_queued_bytes_without_blocking()
 
     auto result = m_client_request_pipe->write(bytes_to_send);
     if (result.is_error()) {
-        if (result.error().code() != EAGAIN)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wlogical-op"
+        if (auto code = result.error().code(); code != EAGAIN && code != EWOULDBLOCK)
+#pragma GCC diagnostic pop
```
ayeteadoe: Unfortunately the GCC Linux build was not happy with `code != EAGAIN && code != EWOULDBLOCK` when `EAGAIN == EWOULDBLOCK`:
```
/home/runner/_work/ladybird/ladybird/Services/RequestServer/Request.cpp:617:63: error: logical ‘and’ of equal expressions [-Werror=logical-op]
  617 |         if (auto code = result.error().code(); code != EAGAIN && code != EWOULDBLOCK)
      | 
```

macOS, Linux Clang, and Windows were fine with it, and this lets us keep the single branch fix

R-Goc: We have AK/Diagnostics.h AK_IGNORE_DIAGNOSTIC for this.

Services/RequestServer/RequestPipe.cpp
```diff
@@ -59,8 +59,11 @@ ErrorOr<ssize_t> RequestPipe::write(ReadonlyBytes bytes)
 {
 #if defined(AK_OS_WINDOWS)
     auto sent = ::send(m_writer_fd, reinterpret_cast<char const*>(bytes.data()), bytes.size(), 0);
-    if (sent == SOCKET_ERROR)
+    if (sent == SOCKET_ERROR) {
+        if (WSAGetLastError() == WSAEWOULDBLOCK)
+            return Error::from_errno(EWOULDBLOCK);
         return Error::from_windows_error();
```
R-Goc, line 65: In the path where the error isn't WSAEWOULDBLOCK won't this lose the error as the last call would be a success from WSAGetLastError?

Services/RequestServer/Request.cpp
trflynn89: Shot in the dark, could you do:
```c++
if (!first_is_one_of(result.error().code(), EAGAIN, EWOULDBLOCK))
```

Services/RequestServer/RequestPipe.cpp
ayeteadoe: No this is fine (as mentioned in the commit, we do it in `SocketWindows` already). Nothing resets the error code from the time it gets called again `from_windows_error()`, so subsequent calls return the same value.

PR #6865, opened 2025-11-19
Apply own clip rect for background phase only when clip used

kalenikaliaksandr:
Fixes a bug where we would clip `box-shadow` when `overflow: hidden` was set, which is not supposed to happen since `overflow` only affects clipping of an element's content.

<table>
<tr>
 <td>Before
 <td>After
<tr>
 <td>CleanShot 2025-11-19 at 04 26 51@2x
 <td>CleanShot 2025-11-19 at 04 26 23@2x
</table>


Libraries/LibWeb/Painting/ViewportPaintable.cpp
```diff
@@ -167,6 +167,7 @@ void ViewportPaintable::assign_clip_frames()
 
             auto clip_rect = block_paintable_box.overflow_clip_edge_rect();
             if (block_paintable_box.get_clip_rect().has_value()) {
+                clip_frame.includes_rect_from_clip_property = true;
```
gmta, line 170: This was added in the wrong commit

PR #6869, opened 2025-11-19
WebDriver: Remove almost all dependency on spin_until and Promise await

Lubrsi:
The remaining spin_until remains in taking screenshots, but attempting
to remove it caused errors in Skia which I couldn't figure out.

I don't think there's a neat way to split this into separate commits,
since it changes the fundamental infrastructure of how requests are
made and responded to.


Libraries/LibCore/Promise.h
```diff
@@ -47,32 +47,36 @@ class Promise : public PromiseBase {
...
-                    weak_promise->reject(resolved->error.release_value());
-            });
-
-            promise->add_child(*p);
+                });
+
+                promise->add_child(*p);
+            }
+        } else {
+            promise->resolve({});
```
gmta: Instead of indenting everything even further, maybe just

```cpp
if (promises.is_empty()) {
    promise->resolve({});
    return promise;
}
```

?

PR #6873, opened 2025-11-20
LibWeb/DOM: Invalidate relative font-weight/size when parent's changes

AtkinsSJ:
`font-weight` and `font-size` both can have keywords that are relative to their inherited value, and so need recomputing when that changes.

Fixes all but one subtest in font-weight-computed.html, because that remaining one uses container-query units. No font-size tests seem to be affected: font-size-computed.html doesn't update the parent element's `font-size` so this invalidation bug didn't apply.

### DIFF
--- a/Libraries/LibWeb/DOM/Element.cpp
+++ b/Libraries/LibWeb/DOM/Element.cpp
@@ -830,23 +830,32 @@ CSS::RequiredInvalidationAfterStyleChange Element::recompute_inherited_style()
 
     CSS::RequiredInvalidationAfterStyleChange invalidation;
 
-    HashMap<size_t, RefPtr<CSS::StyleValue const>> old_values_with_relative_units;
+    HashMap<size_t, RefPtr<CSS::StyleValue const>> property_values_affected_by_inherited_style;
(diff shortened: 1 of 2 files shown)


Libraries/LibWeb/DOM/Element.cpp
gmta, line 1: Commit title typo: "parent's changes"

AtkinsSJ: Can you clarify? :sweat_smile: I had trouble fitting it in the limit but a fuller version would be "Invalidate relative font-weight/font-size values when their parent's value changes". Maybe I worded it poorly.

gmta: Oh you're referencing font-weight/font-size with "parent's", I missed that. Maybe this?

`LibWeb: Invalidate changed relative font-weight/size for children`

That implies the parent -> child relationship and reads a bit easier.

AtkinsSJ: I've changed it, hopefully that's clearer!

PR #6874, opened 2025-11-20
LibWeb/CSS: Support calculated angles in conic-gradient()

AtkinsSJ:
More functionality with less code. :sunglasses:


AtkinsSJ:
Swapped to storing the angle as a StyleValue, not AngleOrCalculated.

PR #6879, opened 2025-11-20
LibWeb: Avoid unnecessary sorting work when getting animations

Psychpsyo:
This way, the list is not re-sorted on every recursive call.

No new test since this isn't about correctness, but rather just a performance thing.

### DIFF
--- a/Libraries/LibWeb/Animations/Animatable.cpp
+++ b/Libraries/LibWeb/Animations/Animatable.cpp
@@ -73,10 +73,10 @@ WebIDL::ExceptionOr<GC::Ref<Animation>> Animatable::animate(Optional<GC::Root<JS
 WebIDL::ExceptionOr<Vector<GC::Ref<Animation>>> Animatable::get_animations(Optional<GetAnimationsOptions> options)
 {
     as<DOM::Element>(*this).document().update_style();
-    return get_animations_internal(options);
+    return get_animations_internal(GetAnimationsSorted::Yes, options);
 }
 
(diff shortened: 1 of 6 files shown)


Libraries/LibWeb/Animations/Animatable.cpp
```diff
@@ -108,18 +110,20 @@ WebIDL::ExceptionOr<Vector<GC::Ref<Animation>>> Animatable::get_animations_inter
     if (options.has_value() && options->subtree) {
         Optional<WebIDL::Exception> exception;
         TRY(target->for_each_child_of_type_fallible<DOM::Element>([&](auto& child) -> WebIDL::ExceptionOr<IterationDecision> {
-            relevant_animations.extend(TRY(child.get_animations(options)));
+            relevant_animations.extend(TRY(child.get_animations_internal(GetAnimationsSorted::No, update_style, options)));
```
gmta: 🤔 aren't we now recursively updating the style for the same document if it's set to `::Yes`?

Psychpsyo: Which is the same as before.
I'm not 100% sure that's necessary, but this is updating the style for the children, not the original element, so... maybe?

gmta: You're correct in that it's the same as before, I'm just now noticing it! But how exactly is it updating the style for just the children if we're calling `document().update_style()`?

Psychpsyo: Good point, it isn't.
I should read the code more carefully before making assumptions about it.

Psychpsyo:
Test failure is a flake that'll hopefully be resolved by #6880

PR #6882, opened 2025-11-20
LibMedia: Keep data providers' ThreadDatas alive in deferred_invoke

Zaggy1024:
We need strong references to the thread data in order to prevent a UAF when, for example, a seek starts as GC is destroying a media element.

This fixes an AddressSanitizer error detected in `video-gc.html` which was causing some CI flakes.

### DIFF
--- a/Libraries/LibMedia/Providers/AudioDataProvider.cpp
+++ b/Libraries/LibMedia/Providers/AudioDataProvider.cpp
@@ -124,8 +124,8 @@ template<typename T>
 void AudioDataProvider::ThreadData::process_seek_on_main_thread(u32 seek_id, T&& function)
 {
     m_last_processed_seek_id = seek_id;
-    m_main_thread_event_loop.deferred_invoke([this, seek_id, function] mutable {
-        if (m_seek_id != seek_id)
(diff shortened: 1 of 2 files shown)


Lubrsi:
If this happens as a media element is going away, couldn't these be weak pointers instead? Since the media element is being GC'd and thus it's no longer observable if data stored with the element does or doesn't change.

Zaggy1024:
Right, but we don't have a thread-safe weak pointer implementation, so I would have to do that fairly manually in each class. I figured this was simpler, though we may have to see whether this becomes a bottleneck. (Theoretically, thread-safe weak pointers would be heavier since it may involve a mutex?)

Lubrsi:
That makes sense. I'm not sure of the exact implementation details of making a thread-safe WeakPtr, but at first glance, it uses the `RefCounted` `WeakLink` class, which could maybe be made into `AtomicRefCounted`, though it'll likely be more involved than that.

Zaggy1024:
I don't think we would want to make all Weakables thread-safe. I took a stab at making an atomically ref-counted version of it, but decided against it just because it involved a lot more invasive changes than I was willing to make, at least without duplicating a bunch of code. :sweat_smile:

PR #6887, opened 2025-11-20
LibWeb: Implement Clear-Site-Data HTTP response header

estefysc:
Add support for parsing and processing the Clear-Site-Data HTTP response header per W3C specification. This causes the navigation.https.html WPT test to go from timeout to pass.

Implemented:
- Storage clearing (origin-scoped): LocalStorage, SessionStorage, IndexedDB, ServiceWorkers, Caches
- Cookie clearing (global - CookieJar lacks domain-scoped API)
- Header parsing for `"cache"`, `"cookies"`, `"storage"`, `"*"` directives

Not implemented:
- Cache clearing (no API available)
- ExecutionContexts/reload

Testing:
WPT test `clear-site-data/navigation.https.html` passes via browser runner (./Meta/ladybird.py).

### DIFF
--- a/Libraries/LibWeb/Loader/ResourceLoader.cpp
+++ b/Libraries/LibWeb/Loader/ResourceLoader.cpp
@@ -27,6 +27,9 @@
 #include <LibWeb/Page/Page.h>
 #include <LibWeb/Platform/EventLoopPlugin.h>
 #include <LibWeb/Platform/Timer.h>
+#include <LibWeb/SecureContexts/AbstractOperations.h>
+#include <LibWeb/StorageAPI/StorageEndpoint.h>
+#include <LibWeb/StorageAPI/StorageKey.h>
 
 namespace Web {
 
@@ -120,6 +123,121 @@ static void store_response_cookies(Page& page, URL::URL const& url, StringView s
         return;
 
     page.client().page_did_set_cookie(url, cookie.value(), HTTP::Cookie::Source::Http);
+}
+
+enum class ClearSiteDataType : u8 {
+    None = 0,
+    Cache = 1 << 0,
+    ClientHints = 1 << 1,
+    Cookies = 1 << 2,
+    ExecutionContexts = 1 << 3,
+    Storage = 1 << 4,
+};
+
+AK_ENUM_BITWISE_OPERATORS(ClearSiteDataType);
+
+// https://w3c.github.io/webappsec-clear-site-data/#parsing
+static ClearSiteDataType parse_clear_site_data_header(HTTP::HeaderList const& response_headers)
+{
+    // 1. Let types be an empty list.
+    auto types = ClearSiteDataType::None;
+
+    // 2. Let header be the result of extracting header list values given Clear-Site-Data and response's header list.
+    auto header = response_headers.extract_header_list_values("Clear-Site-Data"sv);
+
+    // 3. If header is null or failure, return an empty list.
+    if (header.has<Empty>() || header.has<HTTP::HeaderList::ExtractHeaderParseFailure>())
+        return ClearSiteDataType::None;
+
+    // 4. For each type in header, execute the first matching statement, if any, switching on type:
+    for (auto const& type : header.get<Vector<ByteString>>()) {
+        // `"cache"`
+        if (type == "\"cache\"") {
+            // Append "cache" and "clientHints" to types.
+            types |= ClearSiteDataType::Cache;
+            types |= ClearSiteDataType::ClientHints;
+        }
+        // `"cookies"`
+        else if (type == "\"cookies\"") {
+            // Append "cookies" and "clientHints" to types.
+            types |= ClearSiteDataType::Cookies;
+            types |= ClearSiteDataType::ClientHints;
+        }
+        // `"storage"`
+        else if (type == "\"storage\"") {
+            // Append "storage" to types.
+            types |= ClearSiteDataType::Storage;
+        }
+        // `"executionContexts"`
+        else if (type == "\"executionContexts\"") {
+            // Append "executionContexts" to types.
+            types |= ClearSiteDataType::ExecutionContexts;
+        }
+        // `"clientHints"`
+        else if (type == "\"clientHints\"") {
+            // Append "clientHints" to types.
+            types |= ClearSiteDataType::ClientHints;
+        }
+        // `"*"`
+        else if (type == "\"*\"") {
+            // Append "cache", "cookies", "storage", "clientHints", and "executionContexts" to types.
+            types |= ClearSiteDataType::Cache;
+            types |= ClearSiteDataType::Cookies;
+            types |= ClearSiteDataType::Storage;
+            types |= ClearSiteDataType::ClientHints;
+            types |= ClearSiteDataType::ExecutionContexts;
+        }
+    }
+
+    // 5. Return types.
+    return types;
+}
+
+// https://w3c.github.io/webappsec-clear-site-data/#clear-response
+static void handle_clear_site_data(Page& page, URL::URL const& url, HTTP::HeaderList const& response_headers)
+{
+    // 1. If response's url is not an a priori authenticated URL, then break.
+    if (SecureContexts::is_url_potentially_trustworthy(url) == SecureContexts::Trustworthiness::NotTrustworthy)
+        return;
+
+    // 2. Let types be the result of parsing response's Clear-Site-Data header.
+    auto types = parse_clear_site_data_header(response_headers);
+    if (types == ClearSiteDataType::None)
+        return;
+
+    // 3. Let origin be response's url's origin.
+    auto origin = url.origin();
+
+    // FIXME: 4. Let browsing contexts be the result of preparing to clear data for origin and types.
+
+    // 5. For each type in types:
+    //     1. Execute the first matching statement, if any, switching on type:
+
+    // "cache"
+    if (has_flag(types, ClearSiteDataType::Cache)) {
+        // FIXME: Clear cache for origin.
+    }
+    // "cookies"
+    if (has_flag(types, ClearSiteDataType::Cookies)) {
+        // Clear cookies for origin.
+        page.client().page_did_clear_cookies_for_origin(origin);
+    }
+    // "storage"
+    if (has_flag(types, ClearSiteDataType::Storage)) {
+        // Clear DOM-accessible storage for origin.
+        auto storage_key = StorageAPI::StorageKey { url.origin() }.to_string();
+        page.client().page_did_clear_storage(StorageAPI::StorageEndpointType::LocalStorage, storage_key);
+        page.client().page_did_clear_storage(StorageAPI::StorageEndpointType::SessionStorage, storage_key);
+        page.client().page_did_clear_storage(StorageAPI::StorageEndpointType::IndexedDB, storage_key);
+        page.client().page_did_clear_storage(StorageAPI::StorageEndpointType::ServiceWorkerRegistrations, storage_key);
+        page.client().page_did_clear_storage(StorageAPI::StorageEndpointType::Caches, storage_key);
+    }
+    // "clientHints"
+    if (has_flag(types, ClearSiteDataType::ClientHints)) {
+        // FIXME: Empty Accept-CH cache[origin].
+    }
+
+    // FIXME: 6. If types contains "executionContexts", then reload browsing contexts.
 }
 
 static NonnullRefPtr<HTTP::HeaderList> response_headers_for_file(StringView path, Optional<time_t> const& modified_time)
@@ -510,11 +628,13 @@ void ResourceLoader::handle_network_response_headers(LoadRequest const& request,
         // From https://fetch.spec.whatwg.org/#concept-http-network-fetch:
         // 15. If includeCredentials is true, then the user agent should parse and store response
         //     `Set-Cookie` headers given request and response.
-        for (auto const& [header, value] : response_headers) {
+        for (auto const& [header, value] : response_headers.headers()) {
             if (header.equals_ignoring_ascii_case("Set-Cookie"sv)) {
                 store_response_cookies(*request.page(), request.url().value(), value);
             }
         }
+
+        handle_clear_site_data(*request.page(), request.url().value(), response_headers);
     }
 }
 
--- a/Libraries/LibWebView/CookieJar.cpp
+++ b/Libraries/LibWebView/CookieJar.cpp
@@ -487,6 +487,36 @@ void CookieJar::expire_cookies_accessed_since(UnixDateTime since)
     m_transient_storage.expire_and_purge_cookies_accessed_since(since);
 }
 
+// https://w3c.github.io/webappsec-clear-site-data/#clear-cookies
+void CookieJar::clear_cookies_for_origin(URL::Origin const& origin)
+{
+    // 1. Let registered be the registered domain of origin's host.
+    auto registered = URL::get_registrable_domain(origin.host().serialize());
+    if (!registered.has_value())
+        return;
+
+    // 2. Let cookie list be the set of cookies from the cookie store whose domain attribute is a domain-match with registered.
+    m_transient_storage.for_each_cookie([&](HTTP::Cookie::Cookie& cookie) {
+        if (!HTTP::Cookie::domain_matches(cookie.domain, registered.value()))
+            return;
+
+        // 3. For each cookie in cookie list:
+        //     1. Remove cookie from the cookie store.
+        cookie.expiry_time = UnixDateTime::earliest();
+
+        CookieStorageKey key { cookie.name, cookie.domain, cookie.path };
+        m_transient_storage.set_cookie(key, cookie);
+    });
+
+    m_transient_storage.purge_expired_cookies();
+
+    // FIXME: 4. If the user agent supports other forms of cookie-like storage, these MUST also be cleared for origins whose host's registered domain is registered.
+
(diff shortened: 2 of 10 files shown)


Libraries/LibWeb/Loader/ResourceLoader.cpp
```diff
@@ -110,6 +111,111 @@ static void store_response_cookies(Page& page, URL::URL const& url, ByteString c
     page.client().page_did_set_cookie(url, cookie.value(), Cookie::Source::Http); // FIXME: Determine cookie source correctly
 }
 
+// https://www.w3.org/TR/clear-site-data/#parsing
```
gmta: Could you first make sure your using the editor's draft of this spec?
I.e. https://w3c.github.io/webappsec-clear-site-data/

Libraries/LibWebView/WebContentClient.cpp
```diff
@@ -510,6 +510,14 @@ void WebContentClient::did_expire_cookies_with_time_offset(AK::Duration offset)
     Application::cookie_jar().expire_cookies_with_time_offset(offset);
 }
 
+// FIXME: https://www.w3.org/TR/clear-site-data/#clear-cookies
+// CookieJar lacks a domain-scoped API. We currently clear ALL cookies globally.
+void WebContentClient::did_clear_all_cookies(URL::URL url)
+{
+    (void)url;
+    Application::cookie_jar().expire_cookies_accessed_since(UnixDateTime::from_seconds_since_epoch(0));
```
trflynn89: I haven't thoroughly looked at this patch, but this caught my eye - I really do not want a single site sending a `Clear-Site-Data` header to delete all of my cookies.

Imagine github.com sending this header and suddenly you're logged out of youtube.com.

estefysc: yep, I understand that. I was trying to stay within scope for this PR (added the fixme noting per origin clearing needed to be implemented). If it's no problem, then I can implement the per origin clearing before resubmitting.

Libraries/LibWeb/Loader/ResourceLoader.cpp
estefysc: Will do.

Libraries/LibWebView/WebContentClient.cpp
gmta: This should definitely not clear _everything_; please update the PR to do this per origin instead.

estefysc: Will do

estefysc:
I am resubmitting with code that clears cookies per origin and with the editor's draft url for the spec.

AtkinsSJ:
This unfortunately needs a rebase on master because of an issue with a test that's since been fixed. Also, click the "resolve conversation" button when you've addressed feedback and submitted changes. It makes it clear to maintainers that it's ready for another review.

estefysc:
@AtkinsSJ Great, I will work on that. Thanks for the feedback!

Libraries/LibWeb/Loader/ResourceLoader.cpp
```diff
@@ -117,6 +118,117 @@ static void store_response_cookies(Page& page, URL::URL const& url, StringView s
     page.client().page_did_set_cookie(url, cookie.value(), HTTP::Cookie::Source::Http);
 }
 
+// https://w3c.github.io/webappsec-clear-site-data/#parsing
+// 4.1 Parsing
```
trflynn89: This "4.1 Parsing" line isn't needed. Similar comment on `handle_clear_site_data` and other methods you've added below.

Libraries/LibWeb/Loader/ResourceLoader.cpp
```diff
@@ -117,6 +118,117 @@ static void store_response_cookies(Page& page, URL::URL const& url, StringView s
...
+static Vector<String> parse_clear_site_data_header(HTTP::HeaderList const& response_headers)
+{
+    // 1. "Let types be an empty list."
```
trflynn89: Please don't surround spec steps with quotes like these. See other files in LibWeb for how we format spec comments. Basically just:
```c++
// 1. Let types be an empty list.
```
(wrapping long lines at ~120 characters)

Libraries/LibWeb/Loader/ResourceLoader.cpp
```diff
@@ -117,6 +118,117 @@ static void store_response_cookies(Page& page, URL::URL const& url, StringView s
...
+    Vector<String> types;
+
+    // FIXME: 2. "Let header be the result of extracting header list values given Clear-Site-Data and response's header list."
+    // https://fetch.spec.whatwg.org/#extract-header-list-values
+    //
+    // The Fetch spec's "extract header list values" algorithm is already implemented in
+    // LibWeb/Fetch/Infrastructure/HTTP/Headers.cpp:extract_header_list_values(), but it operates on
+    // Fetch::Infrastructure::HeaderList while we have HTTP::HeaderMap here in the network layer.
+    // Implementing a conversion is beyond the scope of this PR (Clear-Site-Data WPT test implementation).
```
trflynn89: I don't think there's any incompatibility here? This should just work:
```c++
// 2. Let header be the result of extracting header list values given Clear-Site-Data and response's header list.
auto header = response_headers.extract_header_list_values("Clear-Site-Data"sv);

// 3. If header is null or failure, return an empty list.
if (header.has<Empty>() || header.has<HTTP::HeaderList::ExtractHeaderParseFailure>())
    return ClearSiteDataType::None;
```

Libraries/LibWeb/Loader/ResourceLoader.cpp
trflynn89: Given that we are returning a set of known values, I think an enum is more appropriate here than a string. Something like:
```c++
enum class ClearSiteDataType : u8 {
    Cache,
    ClientHints,
    Cookies,
    ExecutionContexts,
    Storage,
};
```

This makes subsequent comparisons a simple integer comparison rather than a string comparison, and any typos would become a compile error.

Libraries/LibWeb/Loader/ResourceLoader.cpp
```diff
@@ -117,6 +118,117 @@ static void store_response_cookies(Page& page, URL::URL const& url, StringView s
...
+        }
+    }
+
+    // 3. "If header is null or failure, return an empty list."
+    if (extracted_values.is_empty())
+        return types;
+
+    // 4. "For each type in header, execute the first matching statement, if any, switching on type:"
+    for (auto const& value : extracted_values) {
+        if (value == "cache"sv) {
```
trflynn89: Please include the spec text for the switched statements. And by invoking `extract_header_list_values` directly, we will need to include the quotes like the spec states. So something like:
```c++
// 4. For each type in header, execute the first matching statement, if any, switching on type:
for (auto const& type : header.get<Vector<ByteString>>()) {
    // `"cache"`
    if (type == "\"cache\"") {
        // Append "cache" and "clientHints" to types.
        types |= ClearSiteDataType::Cache;
        types |= ClearSiteDataType::ClientHints;
    }
    // `"cookies"`
    else if (type == "\"cookies\"") {
        // Append "cookies" and "clientHints" to types.
        types |= ClearSiteDataType::Cookies;
        types |= ClearSiteDataType::ClientHints;
    }
    // `"storage"`
    else if (type == "\"storage\"") {
        // Append "storage" to types.
        types |= ClearSiteDataType::Storage;
    }
    // `"executionContexts"`
    else if (type == "\"executionContexts\"") {
        // Append "executionContexts" to types.
        types |= ClearSiteDataType::ExecutionContexts;
    }
    // `"clientHints"`
    else if (type == "\"clientHints\"") {
        // Append "clientHints" to types.
        types |= ClearSiteDataType::ClientHints;
    }
    // `"*"`
    else if (type == "\"*\"") {
        // Append "cache", "cookies", "storage", "clientHints", and "executionContexts" to types.
        types |= ClearSiteDataType::Cache;
        types |= ClearSiteDataType::Cookies;
        types |= ClearSiteDataType::Storage;
        types |= ClearSiteDataType::ClientHints;
        types |= ClearSiteDataType::ExecutionContexts;
    }
}
```

(The above assumes the `ClearSiteDataType` enum is used).

Libraries/LibWeb/Loader/ResourceLoader.cpp
```diff
@@ -117,6 +118,117 @@ static void store_response_cookies(Page& page, URL::URL const& url, StringView s
     page.client().page_did_set_cookie(url, cookie.value(), HTTP::Cookie::Source::Http);
 }
 
+// https://w3c.github.io/webappsec-clear-site-data/#parsing
+// 4.1 Parsing
+static Vector<String> parse_clear_site_data_header(HTTP::HeaderList const& response_headers)
+{
+    // 1. "Let types be an empty list."
+    Vector<String> types;
```
trflynn89: A couple of things here:

1. Rather than returning a list of strings, we have quite a limited set of known values here. IMO that makes an `enum` more appropriate. This makes subsequent comparisons a simple integer comparison, rather than a string comparison. It also ensures any typos would become a compile error rather than a runtime bug.

2. Although the spec states to create a list, I think making the enum a bitwise flag is better here. As a list, we can easily end up appending the same entry type more than once, causing IPC churn an unnecessary work. A bitwise flag would ensure we only handle each type of cache clear just once.

So that would be something like:
```c++
enum class ClearSiteDataType : u8 {
    None = 0,
    Cache = 1 << 0,
    ClientHints = 1 << 1,
    Cookies = 1 << 2,
    ExecutionContexts = 1 << 3,
    Storage = 1 << 4,
};

AK_ENUM_BITWISE_OPERATORS(ClearSiteDataType);
```

And here, step 1 would simply be:
```c++
// 1. Let types be an empty list.
auto types = ClearSiteDataType::None;
```

Libraries/LibWeb/Loader/ResourceLoader.cpp
```diff
@@ -117,6 +118,117 @@ static void store_response_cookies(Page& page, URL::URL const& url, StringView s
...
+    if (types.is_empty())
+        return;
+
+    // 3. "Let origin be response's url's origin."
+    auto origin = url.origin();
+
+    // FIXME: 4. "Let browsing contexts be the result of preparing to clear data for origin and types."
+
+    // 5. "For each type in types: Execute the first matching statement, if any, switching on type:
+    for (auto const& type : types) {
```
trflynn89: With the aforementioned enum, this would become:
```c++
// 5. For each type in types:
//     1. Execute the first matching statement, if any, switching on type:

// "cache"
if (has_flag(types, ClearSiteDataType::Cache)) {
    // FIXME: Clear cache for origin.
}
// "cookies"
if (has_flag(types, ClearSiteDataType::Cookies)) {
    // Clear cookies for origin.
    page.client().page_did_clear_cookies_for_registered_domain(origin);
}
// "storage"
if (has_flag(types, ClearSiteDataType::Storage)) {
    // Clear DOM-accessible storage for origin.
    auto storage_key = StorageAPI::StorageKey { url.origin() }.to_string();
    page.client().page_did_clear_storage(StorageAPI::StorageEndpointType::LocalStorage, storage_key);
    page.client().page_did_clear_storage(StorageAPI::StorageEndpointType::SessionStorage, storage_key);
    page.client().page_did_clear_storage(StorageAPI::StorageEndpointType::IndexedDB, storage_key);
    page.client().page_did_clear_storage(StorageAPI::StorageEndpointType::ServiceWorkerRegistrations, storage_key);
    page.client().page_did_clear_storage(StorageAPI::StorageEndpointType::Caches, storage_key);
}
// "clientHints"
if (has_flag(types, ClearSiteDataType::ClientHints)) {
    // FIXME: Empty Accept-CH cache[origin].
}
```

trflynn89: Also please be sure to copy spec text verbatim. I'm not sure where numbers like "4.2.4" and "4.2.5" came from here.

Libraries/LibWeb/Loader/ResourceLoader.cpp
```diff
@@ -117,6 +118,117 @@ static void store_response_cookies(Page& page, URL::URL const& url, StringView s
...
+    // 5. "Return types."
+    return types;
+}
+
+// https://w3c.github.io/webappsec-clear-site-data/#clear-response
+// 4.2 Clear data for response
+static void handle_clear_site_data(Page& page, URL::URL const& url, HTTP::HeaderList const& response_headers)
+{
+    // 1. "If response's url is not an a priori authenticated URL, then break."
+    // An a priori authenticated URL is one that is considered secure (e.g., HTTPS).
```
trflynn89: This isn't correct. The spec states:
> An a priori authenticated URL is equivalent to a potentially trustworthy URL

Which then has an algorithm to determine if it is trustworthy:
https://w3c.github.io/webappsec-secure-contexts/#potentially-trustworthy-url

Libraries/LibWeb/Loader/ResourceLoader.cpp
```diff
@@ -505,10 +617,27 @@ void ResourceLoader::handle_network_response_headers(LoadRequest const& request,
         // From https://fetch.spec.whatwg.org/#concept-http-network-fetch:
         // 15. If includeCredentials is true, then the user agent should parse and store response
         //     `Set-Cookie` headers given request and response.
-        for (auto const& [header, value] : response_headers) {
+        bool found_clear_site_data = false;
+        for (auto const& [header, value] : response_headers.headers()) {
             if (header.equals_ignoring_ascii_case("Set-Cookie"sv)) {
                 store_response_cookies(*request.page(), request.url().value(), value);
             }
+            if (header.equals_ignoring_ascii_case("Clear-Site-Data"sv)) {
```
trflynn89: I don't think we actually need this check. We can just invoke `handle_clear_site_data` below without tracking this boolean - it will handle the case that the header isn't present already.

Libraries/LibWebView/CookieJar.cpp
```diff
@@ -487,6 +487,46 @@ void CookieJar::expire_cookies_accessed_since(UnixDateTime since)
     m_transient_storage.expire_and_purge_cookies_accessed_since(since);
 }
 
+// https://w3c.github.io/webappsec-clear-site-data/#clear-cookies
+// 4.2.4 Clear cookies for origin
+void CookieJar::clear_cookies_for_registered_domain(URL::Origin const& origin)
```
trflynn89: Given that the spec function is called "clear cookies for origin", let's call our function `clear_cookies_for_origin`

Libraries/LibWebView/CookieJar.cpp
```diff
@@ -487,6 +487,46 @@ void CookieJar::expire_cookies_accessed_since(UnixDateTime since)
...
+{
+    // 1. Let registered be the registered domain of origin's host.
+    auto registered = URL::get_registrable_domain(origin.host().serialize());
+    if (!registered.has_value())
+        return;
+
+    // 2. Let cookie list be the set of cookies from the cookie store
+    //    whose domain attribute is a domain-match with registered.
```
trflynn89: Please only wrap comments at ~120 characters.

Libraries/LibWebView/CookieJar.cpp
```diff
@@ -487,6 +487,46 @@ void CookieJar::expire_cookies_accessed_since(UnixDateTime since)
...
+    Vector<HTTP::Cookie::Cookie> cookie_list;
+    m_transient_storage.for_each_cookie([&](HTTP::Cookie::Cookie const& cookie) {
+        if (HTTP::Cookie::domain_matches(registered.value(), cookie.domain)) {
```
trflynn89: I think these arguments are backwards? I believe "whose domain attribute is a domain-match with registered" means that the domain attribute goes first.
```c++
HTTP::Cookie::domain_matches(cookie.domain, registered.value())
```

Libraries/LibWebView/CookieJar.cpp
```diff
@@ -487,6 +487,46 @@ void CookieJar::expire_cookies_accessed_since(UnixDateTime since)
...
+    for (auto const& cookie : cookie_list) {
+        CookieStorageKey key { cookie.name, cookie.domain, cookie.path };
+        // Remove by marking as expired and purging
+        if (auto existing_cookie = m_transient_storage.get_cookie(key); existing_cookie.has_value()) {
+            auto expired_cookie = existing_cookie.value();
+            expired_cookie.expiry_time = UnixDateTime::earliest();
+            m_transient_storage.set_cookie(key, move(expired_cookie));
+        }
+    }
+    m_transient_storage.purge_expired_cookies();
```
trflynn89, line 511: I think this whole block can be simplified:
```c++
// 2. Let cookie list be the set of cookies from the cookie store whose domain attribute is a domain-match with registered.
m_transient_storage.for_each_cookie([&](HTTP::Cookie::Cookie& cookie) {
    if (!HTTP::Cookie::domain_matches(cookie.domain, registered.value()))
        return;

    // 3. For each cookie in cookie list:
    //     1. Remove cookie from the cookie store.
    cookie.expiry_time = UnixDateTime::earliest();

    CookieStorageKey key { cookie.name, cookie.domain, cookie.path };
    m_transient_storage.set_cookie(key, cookie);
});

m_transient_storage.purge_expired_cookies();
```

trflynn89: (Again, please be sure to copy spec text verbatim. Don't collapse step 3.1 into the same line as step 3.)

estefysc:
@trflynn89 thanks for the detailed feedback and sorry there had to be so many changes done.

PR #6894, opened 2025-11-21
LibWeb: Absolutize gradient StyleValues

AtkinsSJ:
...and also switch them to storing parameters as StyleValues, as discussed on Discord.

Gets us a casual +110 passes on this newly-imported WPT test. :^)

The remaining 110 failures are from not (yet) absolutizing PositionStyleValue and EdgeStyleValue. Those are a bit more involved.


Tests/LibWeb/Screenshot/images/css-gradients-ref.png
gmta, line 1: Please pull this through `pngcrush` or a similar tool :)

AtkinsSJ: Oops, I thought I already did but I guess I missed one of the times!

Libraries/LibWeb/Painting/GradientPainting.cpp
```diff
@@ -127,24 +140,41 @@ LinearGradientData resolve_linear_gradient_data(Layout::NodeWithStyle const& nod
...
         .length_resolution_context = CSS::Length::ResolutionContext::for_layout_node(node),
     };
-    return { conic_gradient.angle_degrees(context), resolved_color_stops, conic_gradient.interpolation_method() };
+    auto from_angle = conic_gradient.angle_degrees(context);
+
+    CSS::Angle const one_turn { 360.0f, CSS::AngleUnit::Deg };
+    context.percentage_basis = one_turn;
+    auto resolved_color_stops = resolve_color_stop_positions(
+        node, conic_gradient.color_stop_list(), [&](auto const& position) -> float {
+            if (position.is_angle())
```
Calme1709: Which ever is merged second of this or #6854 can update this to use the `Angle::from_style_value` method introduced there.

Libraries/LibWeb/CSS/Parser/GradientParsing.cpp
```diff
@@ -563,23 +563,22 @@ RefPtr<RadialGradientStyleValue const> Parser::parse_radial_gradient_function(To
...
+                return commit_value(EllipseSize { first_radius.release_nonnull(), second_radius.release_nonnull() },
                     transaction_size, transaction_second_dimension);
             }
         }
-        // FIXME: Support calculated lengths
-        if (first_radius->is_length())
-            return commit_value(CircleSize { first_radius->length() }, transaction_size);
+        // We parsed the first value as a <length-percentage>, but here we only want <length>s.
+        if (first_radius->is_length() || (first_radius->is_calculated() && first_radius->as_calculated().resolves_to_length()))
+            return commit_value(CircleSize { first_radius->as_length() }, transaction_size);
```
Calme1709: Calling `as_length` here crashes for calculated lengths

Libraries/LibWeb/CSS/Parser/GradientParsing.cpp
```diff
@@ -563,23 +563,22 @@ RefPtr<RadialGradientStyleValue const> Parser::parse_radial_gradient_function(To
...
-                return commit_value(EllipseSize { first_radius.release_value(), second_radius.release_value() },
+                return commit_value(EllipseSize { first_radius.release_nonnull(), second_radius.release_nonnull() },
                     transaction_size, transaction_second_dimension);
             }
         }
-        // FIXME: Support calculated lengths
-        if (first_radius->is_length())
-            return commit_value(CircleSize { first_radius->length() }, transaction_size);
+        // We parsed the first value as a <length-percentage>, but here we only want <length>s.
+        if (first_radius->is_length() || (first_radius->is_calculated() && first_radius->as_calculated().resolves_to_length()))
```
Calme1709: `resolves_to_length` returns true for calculated values that includes percentages which resolve as lengths, we probably want to use `contains_percentage` here

AtkinsSJ:
Changes:
- Rebase on master
- Make use of `Angle::from_style_value()`
- Remove dodgy use of `as_length()` that isn't needed
- Replace dubious `resolves_to_length()` usage
- optipng the image each time
- Change to using absolute-length-only methods in a couple of places after absolutization has happened
- Fiddle about with the number conversions to try and eliminate the rendering differences. I gave up in the end and just upped the fuzzy-matching values so that they pass for me on Linux and macOS. Not satisfying, but there's enough minor funkiness with our gradient rendering that it doesn't seem worth worrying too much. (See #6918)

Libraries/LibWeb/Painting/GradientPainting.cpp
```diff
@@ -127,10 +127,10 @@ LinearGradientData resolve_linear_gradient_data(Layout::NodeWithStyle const& nod
     auto resolved_color_stops = resolve_color_stop_positions(
         node, linear_gradient.color_stop_list(), [&](auto const& position) -> float {
             if (position.is_length())
-                return position.as_length().length().to_px_without_rounding(*context.length_resolution_context) / gradient_length_px;
+                return position.as_length().length().absolute_length_to_px_without_rounding() / gradient_length_px;
             if (position.is_percentage())
                 return position.as_percentage().percentage().as_fraction();
-            return position.as_calculated().resolve_length(context)->to_px_without_rounding(*context.length_resolution_context) / gradient_length_px;
+            return position.as_calculated().resolve_length(context)->absolute_length_to_px_without_rounding() / gradient_length_px;
```
Calme1709, line 132: We no longer need to set `length_resolution_context` on `context`. Same below in `radial_gradient_data`

Libraries/LibWeb/Painting/GradientPainting.cpp
```diff
@@ -127,10 +139,10 @@ LinearGradientData resolve_linear_gradient_data(Layout::NodeWithStyle const& nod
 
 ConicGradientData resolve_conic_gradient_data(Layout::NodeWithStyle const& node, CSS::ConicGradientStyleValue const& conic_gradient)
 {
-    CSS::Angle one_turn(360.0f, CSS::AngleUnit::Deg);
+    CSS::Angle const one_turn { 360.0f, CSS::AngleUnit::Deg };
     auto resolved_color_stops = resolve_color_stop_positions(
-        node, conic_gradient.color_stop_list(), [&](auto const& angle_percentage) {
-            return angle_percentage.resolved(node, one_turn).to_degrees() / one_turn.to_degrees();
+        node, conic_gradient.color_stop_list(), [&](auto const& position) -> float {
+            return CSS::Angle::from_style_value(position, one_turn).to_degrees() / one_turn.to_degrees();
```
Calme1709, line 144: Just as an FYI since it's fixed in a later commit but `Angle::from_style_value` expects the passed value to be absolutized which this isn't. Perhaps we should rename `from_style_value` to `from_absolutized_style_value` (also applies to the equivalent methods on `Time`, `LengthPercentage{OrAuto}` and `EasingFunction`).

Libraries/LibWeb/CSS/StyleValues/ConicGradientStyleValue.cpp
```diff
@@ -73,18 +73,11 @@ bool ConicGradientStyleValue::equals(StyleValue const& other) const
...
         return 0;
-    if (m_properties.from_angle->is_angle())
-        return m_properties.from_angle->as_angle().angle().to_degrees();
-    if (m_properties.from_angle->is_calculated()) {
-        if (auto maybe_angle = m_properties.from_angle->as_calculated().resolve_angle(context); maybe_angle.has_value())
-            return maybe_angle->to_degrees();
-        return 0;
-    }
-    VERIFY_NOT_REACHED();
+    return Angle::from_style_value(*m_properties.from_angle, {}).to_degrees();
```
Calme1709, line 94: Same issue here with `from_style_value` expecting an absolutized value.

Libraries/LibWeb/Painting/GradientPainting.cpp
AtkinsSJ: Right! I didn't quite understand before because I thought somehow a calc() could still contain relative lengths after absolutizing, but that shouldn't be the case. My brain overheats a bit trying to keep everything straight. Thanks for your patience. :sweat_smile:

Libraries/LibWeb/Painting/GradientPainting.cpp
AtkinsSJ: Ah, that's unfortunate. The LengthPercentage one is fine, it doesn't assume that the value is absolutized, so having the same name is a bit of a footgun.

I'll be a good boy and avoid using it until it's safe. Means things will be verbose briefly but the end result is the same.

AtkinsSJ:
Changes:
- Rebased on master.
- Avoided using `Angle::from_style_value()` before the absolutization commit. It now happens in that commit instead.
- The absolutization commit also now avoids passing length resolution contexts in as we don't need them.

Calme1709:
Looks good to me!

PR #6897, opened 2025-11-21
LibWeb: Fix some issues with dispatching slotchange events

AtkinsSJ:
See commits.

### DIFF
--- a/Libraries/LibWeb/DOM/Slottable.cpp
+++ b/Libraries/LibWeb/DOM/Slottable.cpp
@@ -256,7 +256,10 @@ void assign_a_slot(Slottable const& slottable)
 void signal_a_slot_change(GC::Ref<HTML::HTMLSlotElement> slottable)
 {
     // 1. Append slot to slot’s relevant agent’s signal slots.
-    HTML::relevant_similar_origin_window_agent(slottable).signal_slots.append(slottable);
+    // NB: signal_slots is supposed to be a set, so ensure the slottable is not already there.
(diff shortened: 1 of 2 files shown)


Libraries/LibWeb/DOM/Slottable.cpp
```diff
@@ -256,7 +256,10 @@ void assign_a_slot(Slottable const& slottable)
 void signal_a_slot_change(GC::Ref<HTML::HTMLSlotElement> slottable)
 {
     // 1. Append slot to slot’s relevant agent’s signal slots.
-    HTML::relevant_similar_origin_window_agent(slottable).signal_slots.append(slottable);
+    // NB: signal_slots is supposed to be a set, so ensure the slottable is not already there.
+    auto& signal_slots = HTML::relevant_similar_origin_window_agent(slottable).signal_slots;
+    if (!signal_slots.contains_slow(slottable))
+        signal_slots.append(slottable);
```
awesomekling, line 262: Can we solve this by using an `OrderedHashTable` for `signal_slots` instead of `Vector`?

AtkinsSJ: I tried that but wasn't sure how hashing should work, because it wants hash traits for HTMLSlotElement. (I hoped it would just hash the pointer address as it's a GC pointer type, but apparently not.)

awesomekling: That seems like a bug in `Traits<GC::Root<T>>`!

It should definitely hash the pointer if `T` is a pointer type. However, if `T` is `JS::Value`, we should use the `Traits<JS::Value>` instead.

PR #6898, opened 2025-11-21
LibMedia: Use u32 instead of uint to parse FFmpeg version numbers

Zaggy1024:
This was breaking the build on Windows, apparently we don't have uint there.

### DIFF
--- a/Libraries/LibMedia/Containers/Matroska/Reader.cpp
+++ b/Libraries/LibMedia/Containers/Matroska/Reader.cpp
@@ -590,23 +590,23 @@ void Reader::fix_ffmpeg_webm_quirk()
             return;
 
         auto is_affected_version = [&] {
-            constexpr uint final_major_version = 59;
-            constexpr uint final_minor_version = 30;
-            constexpr uint final_micro_version = 100;
+            constexpr u32 final_major_version = 59;
+            constexpr u32 final_minor_version = 30;
+            constexpr u32 final_micro_version = 100;
 
-            auto major_version = split[0].to_number<uint>();
+            auto major_version = split[0].to_number<u32>();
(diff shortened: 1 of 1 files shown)


Libraries/LibMedia/Containers/Matroska/Reader.cpp
```diff
@@ -590,23 +590,23 @@ void Reader::fix_ffmpeg_webm_quirk()
             return;
 
         auto is_affected_version = [&] {
-            constexpr uint final_major_version = 59;
-            constexpr uint final_minor_version = 30;
-            constexpr uint final_micro_version = 100;
+            constexpr u32 final_major_version = 59;
```
trflynn89, line 593: Right, `uint` is a legacy type provided by `sys/types.h` - we should use `AK/Types.h` types everywhere we can.

nico: add `-Duint="use u32 you buffoon"` to cflags?

Zaggy1024: I'm not even sure why I used uint here, honestly! I never do normally :sweat_smile:

stasoid: It should be `-Duint=use_u32_you_buffoon` then, otherwise [the error is confusing](https://tio.run/##Sy4o0E3OScxL//8/M69EITcxM09DU6GaSwEISkEiFQq2CgbWEH5Ofl66QiVEoPb//3/JaTmJ6cX/dV1AKm2VSotTFUqNjRQq80sVkkrT0vLz85RAkiBttkDJ@FIzk3igZDxUEgA).

PR #6899, opened 2025-11-21
LibWeb: Relax HTML parser to allow more tags inside <select>

F3n67u:
This implements parsing part of customizable <select> spec update. 
See whatwg/html#10548.


Tests/LibWeb/Text/expected/wpt-import/html/syntax/parsing/html5lib_innerHTML_tests_innerHTML_1.txt
```diff
@@ -78,7 +79,7 @@ Pass	html5lib_innerHTML_tests_innerHTML_1.html a23b70f1f246ba08d13b570319391b4a5
 Pass	html5lib_innerHTML_tests_innerHTML_1.html 9d5e0c25bfe921df9ea2897c027f42bc88950e69
 Pass	html5lib_innerHTML_tests_innerHTML_1.html 9210d577d6deecf5ab3505af86c501c5befa0b50
 Pass	html5lib_innerHTML_tests_innerHTML_1.html c34af491c0a339db6ba63fcc478108533347319b
-Pass	html5lib_innerHTML_tests_innerHTML_1.html 2c4284e6b2bb480daa50bca43bcbe29cfcdeeab4
+Fail	html5lib_innerHTML_tests_innerHTML_1.html 2c4284e6b2bb480daa50bca43bcbe29cfcdeeab4
```
gmta, line 82: Why are we now failing this test though?

F3n67u: Currently the spec doesn't allow `<input>` in `<select>` (See https://github.com/whatwg/html/issues/11288).

This test was [updated](https://github.com/html5lib/html5lib-tests/pull/178) from:

```txt
#data
<input><option>
#errors
(1,7): unexpected-input-in-select
#document-fragment
select
#document
| <option>
```

to

```txt
#data
<input><option>
#errors
#document-fragment
select
#document
| <input>
| <option>
```

With this change, the test now expects no parse errors when an `<input>` is placed inside `<select>`. That's why the test is currently failing.

The only remaining failure in `html/semantics/forms/the-select-element/customizable-select/select-parsing.html` comes from the same issue. 

These appears to be Chrome-specific expectations, since [Chrome passes all of these tests](https://wpt.fyi/results/html/semantics/forms/the-select-element/customizable-select/select-parsing.html) - meaning chrome has already implemented support for allowing `<input>` in `<select>`.

gmta: Makes sense - could you leave a little note about this in the commit message?

F3n67u: Updated. please take another look. Thanks!

F3n67u:
@gmta I resolved your comments. Could you please take a look when you get a chance? Thank you!

Libraries/LibWeb/HTML/Parser/HTMLParser.cpp
```diff
@@ -2875,36 +2900,75 @@ void HTMLParser::handle_in_body(HTMLToken& token)
 
     // -> A start tag whose tag name is "select"
     if (token.is_start_tag() && token.tag_name() == HTML::TagNames::select) {
+        // If the parser was created as part of the HTML fragment parsing algorithm (fragment case) and the context element passed to that algorithm is a select element:
```
AtkinsSJ: Nit: I know a lot of this file doesn't do this, but please wrap long comments at 120 columns, at least in new code.

F3n67u: Fixed, please take another look. Thanks.

PR #6902, opened 2025-11-21
test-web: Do not clear the WebContent crash handler between tests

trflynn89:
Clearing the callback opens a window for the WebContent process to crash while we do not have a callback set. In practice, we see this with ASAN crashes, where the crash occurs after we've already signaled that the test has completed.

We now set the crash handler once during init. This required moving the clearing of other callbacks to the test completion handler (we were clearing these callbacks in several different ways anyways, so now we will at least be consistent).

==================

There's some more work we can do here in the future: if we do get a crash from a previous test, it'd be nice to indicate that somehow (as opposed to falsely indicating that the next test is the one that crashed).

There's also the chance that ASAN kills the process during WebContent exit. I've coded a feature to wait for all WC processes to exit before exiting test-web, but then we get crash signals for the large amount of on-exit LSAN reports we currently have, so test-web would fail 100% of the time in CI.

So for now at least, this should make us not block for half an hour after ASAN crashes.

### DIFF
--- a/Tests/LibWeb/test-web/TestWeb.h
+++ b/Tests/LibWeb/test-web/TestWeb.h
@@ -14,6 +14,7 @@
 #include <AK/String.h>
 #include <AK/Time.h>
 #include <AK/Vector.h>
+#include <LibCore/Forward.h>
 #include <LibCore/Promise.h>
 #include <LibGfx/Forward.h>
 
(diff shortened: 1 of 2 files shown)


Tests/LibWeb/test-web/TestWeb.h
```diff
@@ -59,6 +60,8 @@ struct Test {
 
     RefPtr<Gfx::Bitmap const> actual_screenshot {};
     RefPtr<Gfx::Bitmap const> expectation_screenshot {};
+
+    RefPtr<Core::Timer> timer {};
```
gmta: Can we call this `timeout_timer`? That'd make the rest of the code a bit easier to understand imo.

trflynn89: Done!

PR #6906, opened 2025-11-22
AppKit: Prevent autocomplete popover from stealing focus

s21v1d9p:
The autocomplete popover on macOS was stealing focus from the location bar when suggestions were displayed. This change saves the current first responder before showing the popover and restores it immediately after, ensuring the user can continue typing without interruption.


gmta:
#4188 seems very related to this!

gmta (approved):

PR #6907, opened 2025-11-22
Implement TextDecoderStream API

### LINKED ISSUE #6769: Sourcegraph Search returns 0 results every time
### Summary

Searching for anything in [Sourcegraph](https://sourcegraph.com/search) yields zero results whereas there should be results and other browsers are able to handle the website properly.

### Operating system

Linux

### Steps to reproduce

1. Go to sourcegraph.com/search
2. Search for anything (eg. ladybird)
3. Hit enter to search

### Expected behavior

Search results with ladybird in them. (maybe repo name, text match in code etc)

### Actual behavior

Zero results are returned.

https://sourcegraph.com/search?q=ladybird&patternType=keyword&sm=0

```html
N/A
```

```shell
63399.687 WebContent(53): ResourceLoader: Filtered request to: "https://www.googletagmanager.com/gtm.js?id=GTM-TB4NLS7"
63399.688 WebContent(53): HTMLScriptElement: Refusing to run script because the element's result is null.
63399.807 WebContent(53): (js warn) "The Sentry loader you are trying to use isn't working anymore, check your configuration."
63399.812 WebContent(53): ResourceLoader: Filtered request to: "https://static.cloudflareinsights.com/beacon.min.js/vcd15cbe7772f49c399c6a5babf22c1241717689176015"
63399.812 WebContent(53): HTMLScriptElement: Refusing to run script because the element's result is null.
63400.621 WebContent(53): ImageDecoderClient: Failed to decode image with ID 0: Decoding failed: Could not find suitable image decoder plugin for data
63402.304 WebContent(53): FIXME: Unimplemented IDL interface: 'SVGSVGElementPrototype.preserveAspectRatio'
63402.304 WebContent(53): FIXME: Unimplemented IDL interface: 'SVGSVGElementPrototype.createSVGAngle'
63402.304 WebContent(53): FIXME: Unimplemented IDL interface: 'SVGSVGElementPrototype.createSVGTransformFromMatrix'
63402.305 WebContent(53): FIXME: Unimplemented IDL interface: 'HTMLImageElementPrototype.x'
63402.305 WebContent(53): FIXME: Unimplemented IDL interface: 'HTMLImageElementPrototype.y'
```

_No response_

_No response_

### Contribute a patch?

gotlougit:
Fixes #6769


Libraries/LibWeb/Encoding/TextDecoderCommon.h
```diff
@@ -0,0 +1,49 @@
...
+// https://encoding.spec.whatwg.org/#textdecodeoptions
+struct TextDecodeOptions {
+    bool stream = false;
+};
+
+class TextDecoderCommonMixin {
+public:
+    virtual ~TextDecoderCommonMixin();
+
+    // https://encoding.spec.whatwg.org/#dom-textdecoder-encoding
```
shannonbooth, line 29: I think these should be linking to the mixin definitions instead now e.g: https://encoding.spec.whatwg.org/#textdecoder-encoding

Libraries/LibWeb/Encoding/TextDecoderCommon.h
```diff
@@ -0,0 +1,49 @@
...
+    FlyString const& encoding() const { return m_encoding; }
+    // https://encoding.spec.whatwg.org/#dom-textdecoder-fatal
+    bool const& fatal() const { return m_fatal; }
+    // https://encoding.spec.whatwg.org/#dom-textdecoder-ignorebom
+    bool const& ignore_bom() const { return m_ignore_bom; }
```
shannonbooth: The boolean can just be passed by value without a const ref

Libraries/LibWeb/CMakeLists.txt
```diff
@@ -359,6 +359,7 @@ set(SOURCES
     Editing/Internal/Algorithms.cpp
     Encoding/TextDecoder.cpp
     Encoding/TextEncoder.cpp
+    Encoding/TextDecoderCommon.cpp
```
shannonbooth, line 368: nit, but let's keep these in alphabetical order

Libraries/LibWeb/Encoding/TextDecoderCommon.h
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright (c) 2024, Tim Flynn <trflynn89@ladybird.org>
```
shannonbooth: This copyright header doesn't seem right :P

edit: Oh - fixed in the next commit. Could that please be applied in this commit instead?

Libraries/LibWeb/Encoding/TextDecoderCommon.h
```diff
@@ -40,7 +40,7 @@ class TextDecoderCommonMixin {
         , m_ignore_bom(ignore_bom)
     {
     }
-    
+
```
shannonbooth, line 46: Could the formatting fixes please be squashed into the commit that introduced the trailing whitespace / include order etc?

Libraries/LibWeb/Encoding/TextDecoderCommon.idl
```diff
@@ -1,3 +1,14 @@
+// https://encoding.spec.whatwg.org/#textdecoderoptions
+dictionary TextDecoderOptions {
+    boolean fatal = false;
+    boolean ignoreBOM = false;
+};
+
+// https://encoding.spec.whatwg.org/#textdecodeoptions
```
shannonbooth, line 7: I think it makes sense to move these across in the previous commit, since that is where the cpp code was moved around as well

Libraries/LibWeb/Encoding/TextDecoderStream.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+GC_DEFINE_ALLOCATOR(TextDecoderStream);
+
+// https://encoding.spec.whatwg.org/#dom-textdecoderstream
+WebIDL::ExceptionOr<GC::Ref<TextDecoderStream>> TextDecoderStream::construct_impl(JS::Realm& realm, Optional<FlyString> label, Optional<TextDecoderOptions> const& options)
+{
+    auto& vm = realm.vm();
+
+    // 1. Let encoding be the result of getting an encoding from label.
+    // If label is not given, let encoding be UTF-8.
```
shannonbooth, line 32: I don't believe this step is required - it appears like label should always be provided over the IDL interface, and the type should instead just be a FlyString const& instead of an Optional<FlyString>. I believe that should allow removing this ad-hoc step.

Libraries/LibWeb/Encoding/TextDecoderStream.cpp
```diff
@@ -0,0 +1,274 @@
...
+    // 3. Set this's encoding to encoding.
+    // https://encoding.spec.whatwg.org/#dom-textdecoder-encoding
+    // The encoding getter steps are to return this's encoding's name, ASCII lowercased.
+    auto lowercase_encoding_name = encoding.value().to_ascii_lowercase_string();
+
+    // 4. If options["fatal"] is true, then set this's error mode to "fatal".
+    auto fatal = options.value_or({}).fatal;
+
+    // 5. Set this's ignore BOM to options["ignoreBOM"].
+    auto ignore_bom = options.value_or({}).ignore_bom;
```
shannonbooth: Similarly here, I believe the type of TextDecoderOptions should not be Optional, which would allow for removing this value_or call.

Libraries/LibWeb/Encoding/TextDecoderStream.cpp
```diff
@@ -0,0 +1,274 @@
...
+    m_io_queue.append(move(data_buffer));
+
+    // 3. Let output be the I/O queue of scalar values « end-of-queue ».
+    Vector<u32> output;
+
+    // 4. While true:
+    while (true) {
+        // 1. Let item be the result of reading from decoder's I/O queue.
+        if (m_io_queue.is_empty()) {
+            // 2. If item is end-of-queue, then:
```
shannonbooth: This spec comment should be moved above the if statement to align with the right nesting level of the steps

Libraries/LibWeb/Encoding/TextDecoderStream.cpp
```diff
@@ -0,0 +1,274 @@
...
+                TRY(Streams::transform_stream_default_controller_enqueue(*m_transform->controller(), JS::PrimitiveString::create(vm, output_chunk)));
+            }
+
+            // 3. Return.
+            return {};
+        }
+
+        auto bytes_to_process = m_io_queue.take_first();
+
+        // 3. Otherwise:
```
shannonbooth, line 163: The following section does not seem to align with the spec, e.g I can't see any step 3. Otherwise here. Maybe spec has changed since?

Libraries/LibWeb/Encoding/TextDecoderStream.cpp
```diff
@@ -0,0 +1,274 @@
...
+        // 2. Let result be the result of processing an item with item, decoder's decoder, decoder's I/O queue,
+        //    output, and decoder's error mode.
+        auto result = TRY_OR_THROW_OOM(vm, m_decoder.to_utf8({ bytes_to_process.data(), bytes_to_process.size() }));
+
+        // 3. If result is finished, then:
+        // Note: In our implementation, we check for errors
+        if (m_fatal && result.contains(0xfffd)) {
+            // If decoder's error mode is "fatal", then throw a TypeError.
+            return WebIDL::SimpleException { WebIDL::SimpleExceptionType::TypeError, "Decoding failed"sv };
+        }
```
shannonbooth, line 213: We seem to be deviating from the specification in this section of the code as well, it does not seem clear why.

Libraries/LibWeb/Encoding/TextDecoderStream.cpp
```diff
@@ -0,0 +1,274 @@
...
+    // 2. If outputChunk is not the empty string, then enqueue outputChunk in decoder's transform.
+    if (!output_chunk.is_empty()) {
+        TRY(Streams::transform_stream_default_controller_enqueue(*m_transform->controller(), JS::PrimitiveString::create(vm, output_chunk)));
+    }
+
+    // 3. Return.
+    return {};
+}
+
+// https://encoding.spec.whatwg.org/#serialize-io-queue
```
shannonbooth: This seems to be a dead link, and should instead be: https://encoding.spec.whatwg.org/#concept-td-serialize

Libraries/LibWeb/Encoding/TextDecoderStream.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+    // 1. Let output be the empty string.
+    StringBuilder output;
+
+    // 2. While true:
+    for (size_t i = 0; i < queue.size(); ++i) {
+        auto item = queue[i];
+
+        // Skip BOM handling for UTF-8 and UTF-16
+        // Note: do not flush is BOM seen flag in spec
```
shannonbooth: I think it would be best for us to align the names here with the spec naming, and ideally also align this with the spec ordering of this function

Libraries/LibWeb/Encoding/TextDecoderStream.cpp
gotlougit: https://encoding.spec.whatwg.org/#decode-and-enqueue-a-chunk

Step 3 in the while loop states 
> Let result be the result of [processing an item](https://encoding.spec.whatwg.org/#concept-encoding-process) with item, decoder’s [decoder](https://encoding.spec.whatwg.org/#textdecodercommon-decoder), decoder’s [I/O queue](https://encoding.spec.whatwg.org/#textdecodercommon-i-o-queue), output, and decoder’s [error mode](https://encoding.spec.whatwg.org/#textdecoder-error-mode).

Libraries/LibWeb/Encoding/TextDecoderStream.cpp
gotlougit: It is a bit unclear from this, but as per the spec at https://encoding.spec.whatwg.org/#flush-and-enqueue the queue is stopped being read from only when we get a special "finished" element from it and is triggering the exit condition only then. In this code we have the exit condition after the queue is empty, so some of the steps need to be a bit out of order here. This error handling is number 4 in the while loop in the spec:

> Otherwise, if result is [error](https://encoding.spec.whatwg.org/#error), [throw](https://webidl.spec.whatwg.org/#dfn-throw) a [TypeError](https://webidl.spec.whatwg.org/#exceptiondef-typeerror).

Libraries/LibWeb/Encoding/TextDecoderStream.cpp
gotlougit: I just noticed that we don't have a `bom_seen` anywhere in our encoding/decoding code, even though the spec says:
https://encoding.spec.whatwg.org/#textdecoder-bom-seen-flag
> The [TextDecoderCommon](https://encoding.spec.whatwg.org/#textdecodercommon) interface mixin defines common getters that are shared between [TextDecoder](https://encoding.spec.whatwg.org/#textdecoder) and [TextDecoderStream](https://encoding.spec.whatwg.org/#textdecoderstream) objects. These objects have an associated:
> [...]
> BOM seen
>    A boolean, initially false.

gotlougit:
Thanks for the review! I will try fixing up the commit history a bit, but wanted to know your thoughts on the two bits that have my notes on them.

Libraries/LibWeb/Encoding/TextDecoderStream.cpp
shannonbooth: We can add in such a flag now to that object

Libraries/LibWeb/Encoding/TextDecoderStream.cpp
shannonbooth: Right, that is my confusion - isn't that meant to be step 3 here instead?

gotlougit: Ah okay, I will just reword it then according to the spec

gotlougit:
I am not sure about the doc changes requested, but I think I've addressed the rest of the notes.

PR #6911, opened 2025-11-23
LibGfx+LibWeb: Don't crash on Apple Maps satellite view (+some LibGfx refactoring)

InvalidUsernameException:
This makes it so that we can visit [satellite](https://maps.apple.com/frame?map=satellite) and [hybrid](https://maps.apple.com/frame?map=hybrid) view on maps.apple.com. Previously this would crash from a bounds check due to us asking Skia to write 32bpp RGBx8888 data into a buffer sized for 24bpp RGB888. This change is the second to last commit.

The code in question was not testable in CI due to missing ability to use WebGL on CI runners. Therefore, the prior commits contain various refactorings, with the eventual goal to make the code independent from WebGL and therefore testable. The test is then introduced in the last commit together with tests for all other possible formats that we support right now.

### DIFF
--- a/Libraries/LibGfx/Bitmap.cpp
+++ b/Libraries/LibGfx/Bitmap.cpp
@@ -31,6 +31,18 @@ struct BackingStore {
     size_t size_in_bytes { 0 };
 };
 
+StringView bitmap_format_name(BitmapFormat format)
+{
+    switch (format) {
+#define ENUMERATE_BITMAP_FORMAT(format) \
+    case BitmapFormat::format:          \
+        return #format##sv;
+        ENUMERATE_BITMAP_FORMATS(ENUMERATE_BITMAP_FORMAT)
+#undef ENUMERATE_BITMAP_FORMAT
+    }
+    VERIFY_NOT_REACHED();
+}
+
 size_t Bitmap::minimum_pitch(size_t width, BitmapFormat format)
 {
     size_t element_size;
--- a/Libraries/LibGfx/Bitmap.h
+++ b/Libraries/LibGfx/Bitmap.h
@@ -20,13 +20,20 @@ namespace Gfx {
 // A pixel value that does not express any information about its component order
 using RawPixel = u32;
 
+#define ENUMERATE_BITMAP_FORMATS(X) \
+    X(Invalid)                      \
+    X(BGRx8888)                     \
+    X(BGRA8888)                     \
+    X(RGBx8888)                     \
+    X(RGBA8888)
+
 enum class BitmapFormat {
-    Invalid,
-    BGRx8888,
-    BGRA8888,
-    RGBx8888,
-    RGBA8888,
+#define ENUMERATE_BITMAP_FORMAT(format) format,
+    ENUMERATE_BITMAP_FORMATS(ENUMERATE_BITMAP_FORMAT)
+#undef ENUMERATE_BITMAP_FORMAT
 };
+
+[[nodiscard]] StringView bitmap_format_name(BitmapFormat);
 
 inline bool is_valid_bitmap_format(u32 const format)
 {
--- a/Libraries/LibGfx/ImmutableBitmap.cpp
+++ b/Libraries/LibGfx/ImmutableBitmap.cpp
@@ -16,6 +16,18 @@
 
 namespace Gfx {
 
+StringView export_format_name(ExportFormat format)
+{
+    switch (format) {
+#define ENUMERATE_EXPORT_FORMAT(format) \
+    case Gfx::ExportFormat::format:     \
+        return #format##sv;
+        ENUMERATE_EXPORT_FORMATS(ENUMERATE_EXPORT_FORMAT)
+#undef ENUMERATE_EXPORT_FORMAT
+    }
+    VERIFY_NOT_REACHED();
+}
+
 struct ImmutableBitmapImpl {
     sk_sp<SkImage> sk_image;
     SkBitmap sk_bitmap;
--- a/Libraries/LibGfx/ImmutableBitmap.h
+++ b/Libraries/LibGfx/ImmutableBitmap.h
@@ -21,19 +21,22 @@ namespace Gfx {
 
 struct ImmutableBitmapImpl;
 
+#define ENUMERATE_EXPORT_FORMATS(X) \
+    X(Gray8)                        \
+    X(Alpha8)                       \
+    X(RGB565)                       \
+    X(RGBA5551)                     \
+    X(RGBA4444)                     \
+    X(RGB888)                       \
+    X(RGBA8888)
+
 enum class ExportFormat : u8 {
-    // 8 bit
-    Gray8,
-    Alpha8,
-    // 16 bit
-    RGB565,
-    RGBA5551,
-    RGBA4444,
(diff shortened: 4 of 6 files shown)


Libraries/LibGfx/Color.h
```diff
@@ -17,7 +17,7 @@
 
 namespace Gfx {
 
-typedef u32 ARGB32;
+typedef u32 BGRA8888;
```
Hendiadyoin1: On this commit, maybe get a bit more into reasoning, as intuitively `ARGB32` and `BGRA8888` sound like the endian inverses of each other
(also which endiannessis it now?)

Hendiadyoin1: also maybe prefer `using`

Libraries/LibGfx/Bitmap.h
```diff
@@ -17,6 +17,9 @@
 
 namespace Gfx {
 
+// A pixel value that does not express any information about its component order
+typedef u32 RawPixel;
```
Hendiadyoin1: probably prefer `using`

Libraries/LibGfx/ImmutableBitmap.cpp
```diff
@@ -112,21 +119,35 @@ ErrorOr<BitmapExportResult> ImmutableBitmap::export_to_byte_buffer(ExportFormat
...
+        if (format == ExportFormat::RGB888) {
+            // 24 bit RGB is not supported by Skia, so we need to handle this format ourselves.
+            for (auto y = 0; y < height; y++) {
+                for (auto x = 0; x < width; x++) {
+                    auto pixel = get_pixel(x, y);
+                    auto target_y = flags & ExportFlags::FlipY ? height - y - 1 : y;
+                    auto buffer_offset = (target_y * buffer_pitch.value()) + (x * 3ull);
+                    buffer[buffer_offset + 0] = pixel.red();
+                    buffer[buffer_offset + 1] = pixel.green();
+                    buffer[buffer_offset + 2] = pixel.blue();
```
Hendiadyoin1: Note that compilers are really bad at hoisting the internal bounds check
so maybe get a pointer or another unsafe-er view

Libraries/LibGfx/Bitmap.h
gmta, line 1: Commit title typo: "unued"

Libraries/LibGfx/Bitmap.h
```diff
@@ -267,19 +244,20 @@ ALWAYS_INLINE Color Bitmap::get_pixel(int x, int y) const
...
-    case StorageFormat::RGBA8888:
+    case BitmapFormat::RGBA8888:
         scanline(y)[x] = (color.alpha() << 24) | (color.blue() << 16) | (color.green() << 8) | color.red();
         return;
-    case StorageFormat::RGBx8888:
+    case BitmapFormat::RGBx8888:
         scanline(y)[x] = (color.blue() << 16) | (color.green() << 8) | color.red();
         return;
+    default:
+        VERIFY_NOT_REACHED();
```
gmta, line 272: Why did you change this? The old version would result in a compile error if we forgot to include a format.

Libraries/LibGfx/Bitmap.cpp
```diff
@@ -241,7 +241,7 @@ Bitmap::~Bitmap()
 void Bitmap::strip_alpha_channel()
 {
     VERIFY(m_format == BitmapFormat::BGRA8888 || m_format == BitmapFormat::BGRx8888);
-    for (ARGB32& pixel : *this)
+    for (BGRA8888& pixel : *this)
         pixel = 0xff000000 | (pixel & 0xffffff);
```
gmta, line 257: Wouldn't BGRA imply the B channel in the highest (MS) 8 bits?

Libraries/LibGfx/ImmutableBitmap.cpp
InvalidUsernameException: Writing into the buffer data directly now.

Libraries/LibGfx/Color.h
InvalidUsernameException: Hmm, I meant to write something there, but must have forgotten it. Anyway, I've added some rationale for the change. I hope it is understandable.

Libraries/LibGfx/Bitmap.h
InvalidUsernameException: The old version did not compile anymore due to the additional `BitmapFormat::Invalid`. So I just made it structurally the same as `get_pixel()` already was without thinking about it too much. I've now changed both functions to give compile errors should a new enum value be added.

Libraries/LibGfx/Bitmap.cpp
InvalidUsernameException: This is due to the way the type name is now are based on the bytes in memory, but we are accessing it as u32 on a little-endian system, thereby inverting the component order. I went into some more details about this in the commit description I added. Does that make sense?

Libraries/LibGfx/Color.h
Hendiadyoin1: As with the confusion yelle also mentioned, maybe leave a comment next to the type, like
```c++
// Named after in memory-order (little-endian)
// e.g. 0xAARRGGBB
```

if I understood the commit and latter comment correctly

InvalidUsernameException: Added.

Libraries/LibGfx/ImmutableBitmap.h
```diff
@@ -20,6 +21,27 @@ namespace Gfx {
...
+    RGBA4444,
+    // 24 bit
+    RGB888,
+    // 32 bit
+    RGBA8888,
+};
+
+struct ExportFlags {
+    enum : u8 {
+        PremultiplyAlpha = 1 << 0,
```
gmta, line 43: This should be called "PremulipliedAlpha" imo, since it's not an action (like "FlipY") but rather a statement about the data.

Psychpsyo: Same could be argued for FlipY. `FlippedY` would be a description of the data, whereas `FlipY` is the action you perform to get the data into that state. Imo, the struct should pick one and stick to it. (though I'm also not a maintainer)

InvalidUsernameException: The names are based on what the [corresponding flags in the WebGL](https://registry.khronos.org/webgl/specs/latest/1.0/#6.10) are called (`UNPACK_FLIP_Y_WEBGL` and `UNPACK_PREMULTIPLY_ALPHA_WEBGL`). I'm fine using either, just wanted to give the rationale for the current name.

gmta: > Same could be argued for FlipY. `FlippedY` would be a description of the data, whereas `FlipY` is the action you perform to get the data into that state. Imo, the struct should pick one and stick to it. (though I'm also not a maintainer)

That's not what's going on though - `FlipY` actually translates and scales (flips) the surface. `PremultiplyAlpha` merely describes how the color data should be interpreted.

But having the flags named similarly to the WebGL flags makes some sense here, so let's keep them the same.

PR #6912, opened 2025-11-23
LibWeb: Parse the scroll driven animation CSS properties

Calme1709:
The first step towards implementing CSS scroll driven animations.

Some of the imported tests continue to fail due to the fact that we serialize coordinating value list longhands (e.g. scroll-timeline-axis, animation-duration, etc) incorrectly in that if all values are the same we serialize them as a single value.

Gains us ~320 WPT tests.

### DIFF
--- a/Libraries/LibWeb/CSS/CSSStyleProperties.cpp
+++ b/Libraries/LibWeb/CSS/CSSStyleProperties.cpp
@@ -21,6 +21,7 @@
 #include <LibWeb/CSS/StyleValues/PercentageStyleValue.h>
 #include <LibWeb/CSS/StyleValues/ShorthandStyleValue.h>
 #include <LibWeb/CSS/StyleValues/StyleValueList.h>
+#include <LibWeb/CSS/StyleValues/TimeStyleValue.h>
 #include <LibWeb/CSS/StyleValues/TransformationStyleValue.h>
 #include <LibWeb/DOM/Document.h>
 #include <LibWeb/DOM/Element.h>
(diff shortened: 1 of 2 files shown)


Calme1709:
Changes:
Added FIXMEs to propagate animation-duration `auto` value as distinct from `0s`, this is required as it is how the [normalize specified timing algorithm](https://drafts.csswg.org/web-animations-2/#normalize-specified-timing) distinguishes between time and progress-based animations.

Libraries/LibWeb/CSS/StyleValues/ShorthandStyleValue.cpp
```diff
@@ -849,6 +849,36 @@ String ShorthandStyleValue::to_string(SerializationMode mode) const
...
+
+            builder.append(name_values[i]->to_string(mode));
+
+            if (axis_values[i]->to_keyword() != Keyword::Block)
+                builder.appendff(" {}", axis_values[i]->to_string(mode));
+
+            auto stringified_inset = inset_values[i]->to_string(mode);
+
+            if (stringified_inset != "auto"sv)
+                builder.appendff(" {}", inset_values[i]->to_string(mode));
```
AtkinsSJ: You can use `stringified_inset` here.

Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
```diff
@@ -6347,6 +6349,113 @@ RefPtr<StyleValue const> Parser::parse_white_space_shorthand(TokenStream<Compone
...
+
+            if (axis)
+                axes.append(axis.release_nonnull());
+            else
+                axes.append(KeywordStyleValue::create(Keyword::Block));
+
+            if (inset)
+                insets.append(inset.release_nonnull());
+            else
+                insets.append(StyleValueList::create({ KeywordStyleValue::create(Keyword::Auto), KeywordStyleValue::create(Keyword::Auto) }, StyleValueList::Separator::Space));
```
AtkinsSJ, line 6390: Would it be worth using `property_initial_value()` for these instead of manually constructing them?

AtkinsSJ (changes requested):
Very cool! I wasn't expecting this any time soon. Just a couple of very minor things.

(It also bothers me every time reset-only longhands still have to be done manually, or  coordinating value lists, or parsing a list not always returning a list. I'll have to see if I can revive my branch for the former. 😅)

Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
AtkinsSJ: ...no it wouldn't, because those will eventually always parse as lists and we only want one item. :sweat_smile:

Calme1709: We do as such in the generic `parse_coordinating_value_list_shorthand`, we could do here if not for the same reason that we don't use the generic `coordinating_value_list_shorthand_to_string` in `ShorthandStyleValue::to_string()` i.e. that `view-timeline-name` doesn't work with our normal approach of
```c++
auto initial_value = property_initial_value(longhand_id);
if (initial_value->is_value_list())
  return initial_value->as_value_list().values()[0];
return initial_value;
```
since it is already a `StyleValueList` (containing the start and end insets).

I have added a FIXME to do this once we always parse as lists.

PR #6913, opened 2025-11-23
LibWeb/HTML: Invalidate HTMLCollection caches when selectedness of options changes

skyz1:
This adds an invalidation step to the method for updating the selectedness of a select element. This is necessary for updating the html collections for the selected options.


skyz1:
This fix is already included in #6749

PR #6914, opened 2025-11-23
LibWeb/HTML: Ignore form owner if an earlier non-form element shares the ID

skyz1:
Only set form owner when there is no non-form element with the same ID as the control's form attribute earlier in tree order.


Libraries/LibWeb/HTML/FormAssociatedElement.cpp
```diff
@@ -187,9 +187,10 @@ void FormAssociatedElement::reset_form_owner()
     if (is_listed() && html_element.has_attribute(HTML::AttributeNames::form) && html_element.is_connected()) {
         // 1. If the first element in element's tree, in tree order, to have an ID that is identical to element's form content attribute's value, is a form element, then associate the element with that form element.
         auto form_value = html_element.attribute(HTML::AttributeNames::form);
-        html_element.root().for_each_in_inclusive_subtree_of_type<HTMLFormElement>([this, &form_value](HTMLFormElement& form_element) {
+        html_element.root().for_each_in_inclusive_subtree_of_type<HTMLElement>([this, &form_value](auto& form_element) {
```
gmta: Let's rename `form_element` to `element` or `html_element` to better reflect the kind of nodes coming in.

gmta (approved):

PR #6917, opened 2025-11-24
LibWeb: Move mutation observers from IntrusiveList to GC::RootVector

gmta:
We need to prevent these mutation observers from being garbage collected, and since they are only part of SimilarOriginWindowAgent and themselves as part of the intrusive list, nobody is visiting them.

Make the list of pending mutation observers a GC::RootVector so we keep them alive until they have been processed in the microtask.

Restores 1400+ WPT subtest passes in `dom/nodes/Element-classlist.html`.

CC @AtkinsSJ

### DIFF
--- a/Libraries/LibWeb/HTML/Scripting/SimilarOriginWindowAgent.cpp
+++ b/Libraries/LibWeb/HTML/Scripting/SimilarOriginWindowAgent.cpp
@@ -1,10 +1,13 @@
 /*
  * Copyright (c) 2025, Shannon Booth <shannon@serenityos.org>
+ * Copyright (c) 2025, Jelle Raaijmakers <jelle@ladybird.org>
  *
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
 #include <LibWeb/Bindings/MainThreadVM.h>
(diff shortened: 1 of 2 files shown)


Libraries/LibWeb/Bindings/MainThreadVM.cpp
```diff
@@ -731,9 +731,7 @@ void queue_mutation_observer_microtask(DOM::Document const& document)
         surrounding_agent.mutation_observer_microtask_queued = false;
 
         // 2. Let notifySet be a clone of the surrounding agent’s pending mutation observers.
-        GC::RootVector<DOM::MutationObserver*> notify_set(heap);
-        for (auto& observer : surrounding_agent.pending_mutation_observers)
-            notify_set.append(&observer);
+        auto notify_set = move(surrounding_agent.pending_mutation_observers);
```
AtkinsSJ: This now does both steps 2 and 3. You can remove the NB below, rearrange things a little, and also remove the `// NB: Step 3, done later.` step at the end of the function.

gmta: 🤦 I actually saw those comments before and forgot about them. Thanks!

PR #6923, opened 2025-11-24
LibWebView: Fix serif font fallback order for macOS bold text

### LINKED ISSUE #6766: Inline styles for `<i>` and `<b>` are broken
Reproduction:

```html
<i>foo</i>
<b>bar</b>
```

| Ladybird | Firefox |
|--|--|
| Image | Image |

s21v1d9p:
On macOS, the "Times" font only has regular weight (400), preventing bold text from rendering correctly. Prioritizing "Times New Roman" before "Times" in the serif fallback list fixes this issue, as "Times New Roman" includes both regular and bold weight variants.

Fixes #6766


gmta:
This is not entirely the correct fix: we need to take font style availability into account when choosing an entry from this list. This is due to us not implementing the specced font selection algorithm.

s21v1d9p:
Thank you for the feedback. I understand this fix is a workaround rather than implementing the proper font selection algorithm. Since implementing the spec'd font matching is a larger effort beyond my current scope, I'll close this PR. The issue remains open for someone to implement the proper solution.

PR #6935, opened 2025-11-25
Integrate ML-DSA postquantum algorithm into WebCryptoAPI

tete17:
This MR achieves 3 things: 
 * Re-imports the imported WebCryptoAPI WPT tests
 * Add ML-DSA as an Public Key algorithm to LibCrypto
 * Exposes ML-DSA through WebCryptoAPI

 The re-import was needed as a lot of helper scripts had changed to accommodate the new key format these algorithms introduce.

The main code additions are in LibCrypto as the changes in LibWeb as mostly the boilerplate code needed by the spec and even then all the additions are exposing Openssl functionality.

This changes represent an increase 1600+ WPT subtests and making Ladybird the first major browser to support Quantum Secure algorithms for the web in WebCryptoAPI.


Libraries/LibCrypto/PK/MLDSA.h
```diff
@@ -0,0 +1,65 @@
...
+
+namespace Crypto::PK {
+
+enum MLDSASize {
+    MLDSA44,
+    MLDSA65,
+    MLDSA87,
+};
+
+char const* mldsa_size_to_openssl_name(MLDSASize);
```
R-Goc: This looks like it can just be in an anonymous namespace in the cpp file. There is no reason for it to be a public api from what I see.

R-Goc:
Could this get the Windows label just because of the sheer size? Also all the other crypto algorithms have tests under Tests/LibCrypto. While this does have tests from WPT, I think it would be good for it to have some tests of its own too, but that is a maintainer decision.

Libraries/LibCrypto/PK/MLDSA.h
tete17: Sure thing :)

tete17:
@R-Goc I added some test for LibCrypto but they are brutal, public keys are massive.

There is some randomness to the signing which means I can't verify with a fixed test data.

Libraries/LibWeb/Crypto/SubtleCrypto.cpp
```diff
@@ -1212,7 +1212,7 @@ SupportedAlgorithmsMap& supported_algorithms_internal()
     return s_supported_algorithms;
 }
 
-// https://w3c.github.io/webcrypto/#algorithm-normalization-internalS
+// https://w3c.github.io/webcrypto/#algorithm-normalization-internal
```
gmta, line 1215: The commit message does not match the change

Libraries/LibCrypto/CMakeLists.txt
```diff
@@ -20,8 +20,9 @@ set(SOURCES
     Hash/SHA1.cpp
     Hash/SHA2.cpp
     Hash/SHA3.cpp
-    PK/RSA.cpp
     PK/EC.cpp
+    PK/MLDSA.cpp
+    PK/RSA.cpp
```
gmta, line 25: Commit message typo: "algorith"

Libraries/LibWeb/Crypto/CryptoAlgorithms.cpp
```diff
@@ -8432,4 +8432,91 @@ WebIDL::ExceptionOr<Variant<GC::Ref<CryptoKey>, GC::Ref<CryptoKeyPair>>> MLDSA::
...
+        // 13. Set the [[extractable]] internal slot of key to extractable.
+        key->set_extractable(extractable);
+
+        // 14. Set the [[usages]] internal slot of key to usages.
+        key->set_usages(usages);
+    }
+    // FIXME -> If format is "pkcs8":
+    // FIXME -> If format is "raw-public":
+    // FIXME -> If format is "raw-seed":
+    // FIXME -> If format is "jwk":
```
gmta: We use `// FIXME: ...` as the prefix

Libraries/LibWeb/Crypto/CryptoAlgorithms.cpp
```diff
@@ -8519,4 +8519,65 @@ WebIDL::ExceptionOr<GC::Ref<CryptoKey>> MLDSA::import_key(AlgorithmParams const&
...
+        VERIFY(handle.has<::Crypto::PK::MLDSAPublicKey>());
+        auto data = TRY_OR_THROW_OOM(vm, ::Crypto::PK::wrap_in_subject_public_key_info(handle.get<::Crypto::PK::MLDSAPublicKey>().public_key(), algorithm_oid));
+
+        // 3. Let result be the result of DER-encoding data.
+        result = JS::ArrayBuffer::create(m_realm, data);
+    }
+    // FIXME  -> If format is "pkcs8":
+    // FIXME  -> If format is "raw-public":
+    // FIXME  -> If format is "raw-seed":
+    // FIXME  -> If format is "jwk":
```
gmta: Same here: `// FIXME: ...`

Libraries/LibWeb/Crypto/SubtleCrypto.cpp
tete17: Fixed :)

Libraries/LibCrypto/CMakeLists.txt
tete17: Silly me :)

tete17:
If the only review comments are some minor commit typos, I am going to be happy :)

PR #6940, opened 2025-11-26
LibWeb: Apply clip frames even when transformation is identity

InvalidUsernameException:
Clip frames for overflow were applied based on whether the box in question had a non-identity matrix transformation associated with it. That however is not correct, since specifying a no-op transform like `scale(1)` still needs to apply clip overflow rectangles. So instead we need to check whether the element associated with the box in question has any CSS transforms.

This appears to have been a regression from 9bbc1cd61889e8cf07d39b41b6a2938eac6d743d (#6285) and effectively reverts that commit, but keeps its effect by unifying on the check for CSS transforms instead.

This fixes some background boxes being rendered for the invisible items of the carousels on https://computerbase.de/.

Before:
Before

After:
After

/cc @kalenikaliaksandr

### DIFF
--- a/Libraries/LibWeb/Painting/StackingContext.cpp
+++ b/Libraries/LibWeb/Painting/StackingContext.cpp
@@ -338,8 +338,10 @@ void StackingContext::paint(DisplayListRecordingContext& context) const
         push_stacking_context_params.bounding_rect = context.enclosing_device_rect(paintable_box().overflow_clip_edge_rect());
     }
 
-    if (!transform_matrix.is_identity())
+    auto has_css_transform = paintable_box().has_css_transform();
+    if (has_css_transform) {
         paintable_box().apply_clip_overflow_rect(context, PaintPhase::Foreground);
+    }
     paintable_box().apply_scroll_offset(context);
 
     auto mask_image = computed_values.mask_image();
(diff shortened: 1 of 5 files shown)


Libraries/LibWeb/Painting/StackingContext.cpp
```diff
@@ -349,7 +351,13 @@ void StackingContext::paint(DisplayListRecordingContext& context) const
 
     bool needs_to_save_state = mask_image || paintable_box().get_masking_area().has_value();
 
-    if (push_stacking_context_params.has_effect()) {
+    bool push_stacking_context_has_effect = push_stacking_context_params.opacity != 1.0f
```
konradekk: naming nit: `push_stacking_context_has_effect` is somewhat a misnomer: `isolate` is not an effect and there are possibly other effects that do not need separate stacking context.

maybe something like `needs_separate_stacking_context` (new/isolated?), `create_stacking_context` (possibly with similar quantifiers), or `needs_stacking_context_push`, because this is what it is about?

InvalidUsernameException: In my head it read as "This `push_stacking_context` command has an observable effect(, so we need to issue it)". But point taken, I went with `needs_to_push_stacking_context`, which is consistent with the `needs_to_save_state` right below it.

konradekk: (kind of agree with you: both `need_to_•••` do not give the rationale, which is also fine to have, if its complicated or out of scope of current logic, etc. in fact you don't "need to save the state" or "need to push the context" in any way! you _want_ to some consistent effect[s], and these flags merely mark means to achieve that. you need to also restore the state and pop the context under flags named so. this means that the proper naming would be: `need_to_save_and_restore_state`, same for context. this makes me think that you want to have a isolated/independent/fresh/new/pristine state or context, so you store current one and restore previous one using a stack… well captain obvious here, but i hope you'll forgive me being so swollen-headed! 🙇🏻‍♂️)

PR #6945, opened 2025-11-26
LibMedia+LibWeb: Update the duration of an element based on media data

Zaggy1024:
This enables us to fully play files that have a duration smaller than the total duration of all frames in the file.

It's currently not possible to seek past the duration, I'll have to have a look if that's something the spec talks about.

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLMediaElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLMediaElement.cpp
@@ -1302,6 +1302,13 @@ void HTMLMediaElement::on_metadata_parsed()
     // FIXME: Handle unbounded media resources.
     set_duration(m_playback_manager->duration().to_seconds_f64());
 
+    // NB: Register the duration change handler here so that we don't set the duration when the
+    //     playback manager updates the duration after parsing.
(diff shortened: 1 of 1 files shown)


Libraries/LibWeb/Painting/MediaPaintable.cpp
```diff
@@ -417,6 +417,9 @@ void MediaPaintable::set_current_time(HTML::HTMLMediaElement& media_element, CSS
     auto x_percentage = static_cast<double>(x_offset) / static_cast<double>(timeline_rect.width());
     auto position = x_percentage * media_element.duration();
 
+    if (isnan(position))
```
gmta: If duration is zero, you'd get zero for position as well, not NaN? (Maybe the commit message needs to address this?)

Zaggy1024: Ah, good catch, I had misremembered the cause of the crash here. The duration itself could be NaN here if seeking was done before the duration was established from media data.

I've updated that commit to be clearer, both in the patch and the commit message.

kalenikaliaksandr (approved):

PR #6956, opened 2025-11-27
LibWeb/HTML: Update worker construction spec steps

AtkinsSJ:
This is largely editorial. One behaviour change is that events are now sent from a global task on the DOMManipiulation task source.

Somewhat awkwardly, the spec refers to `this` before the Worker exists. As it's for getting the relevent global object / settings object, I've had to work around that.

Corresponds to:
https://github.com/whatwg/html/commit/917c2f6a7367aa36d347f11e65dd4503ff87ddce


awesomekling (commented):
Commit message typos:
- DOMManipiulation
- relevent

AtkinsSJ:
Thanks. :sweat_smile: I dun spelled beter.

PR #6962, opened 2025-11-27
Implement Argon hash for subtleCrypto

tete17:
Unfortunately this only represents a 4 WPT subtest increase.

Most of the code is OpenSSL wrappers with some input validation sprinkled here and there

### DIFF
--- a/Services/WebWorker/main.cpp
+++ b/Services/WebWorker/main.cpp
@@ -8,6 +8,7 @@
 #include <LibCore/EventLoop.h>
 #include <LibCore/Process.h>
 #include <LibCore/System.h>
+#include <LibCrypto/OpenSSLForward.h>
 #include <LibFileSystem/FileSystem.h>
 #include <LibIPC/SingleServer.h>
 #include <LibMain/Main.h>
@@ -21,6 +22,8 @@
 #include <LibWebView/Plugins/ImageCodecPlugin.h>
 #include <LibWebView/Utilities.h>
 #include <WebWorker/ConnectionFromClient.h>
+
+#include <openssl/thread.h>
 
 static ErrorOr<void> initialize_image_decoder(int image_decoder_socket);
 static ErrorOr<void> initialize_resource_loader(GC::Heap&, int request_server_socket);
@@ -66,6 +69,8 @@ ErrorOr<int> ladybird_main(Main::Arguments arguments)
     Core::EventLoop event_loop;
 
     WebView::platform_init();
+
+    OPENSSL_TRY(OSSL_set_max_threads(nullptr, Core::System::hardware_concurrency()));
 
     TRY(initialize_image_decoder(image_decoder_socket));
 
--- a/Services/WebContent/main.cpp
+++ b/Services/WebContent/main.cpp
@@ -12,6 +12,7 @@
 #include <LibCore/Resource.h>
 #include <LibCore/System.h>
 #include <LibCore/SystemServerTakeover.h>
+#include <LibCrypto/OpenSSLForward.h>
 #include <LibGfx/Font/FontDatabase.h>
 #include <LibGfx/Font/PathFontProvider.h>
 #include <LibIPC/ConnectionFromClient.h>
@@ -37,6 +38,8 @@
 #include <WebContent/ConnectionFromClient.h>
 #include <WebContent/PageClient.h>
 #include <WebContent/WebDriverConnection.h>
+
+#include <openssl/thread.h>
 
 #if defined(AK_OS_MACOS)
 #    include <LibCore/Platform/ProcessStatisticsMach.h>
@@ -181,6 +184,8 @@ ErrorOr<int> ladybird_main(Main::Arguments arguments)
 #    endif
     }
 #endif
+
+    OPENSSL_TRY(OSSL_set_max_threads(nullptr, Core::System::hardware_concurrency()));
 
     TRY(initialize_image_decoder(image_decoder_socket));
 
--- a/Services/WebWorker/CMakeLists.txt
+++ b/Services/WebWorker/CMakeLists.txt
@@ -19,7 +19,7 @@ target_link_libraries(webworkerservice PRIVATE OpenSSL::Crypto OpenSSL::SSL)
 add_executable(WebWorker main.cpp)
 
 target_include_directories(WebWorker PRIVATE ${LADYBIRD_SOURCE_DIR})
-target_link_libraries(WebWorker PRIVATE webworkerservice)
+target_link_libraries(WebWorker PRIVATE webworkerservice OpenSSL::Crypto OpenSSL::SSL)
 
 if(WIN32)
     find_package(pthread REQUIRED)


Libraries/LibWeb/Crypto/CryptoAlgorithms.cpp
```diff
@@ -9021,6 +9022,79 @@ WebIDL::ExceptionOr<GC::Ref<CryptoKey>> Argon2::import_key(AlgorithmParams const
...
+        return WebIDL::OperationError::create(m_realm, "Memory is too low for the parallelism level"_utf16);
+
+    // 5. If the passes member of normalizedAlgorithm is zero, then throw an OperationError.
+    if (normalized_algorithm.passes == 0)
+        return WebIDL::OperationError::create(m_realm, "Invalid passes"_utf16);
+
+    auto const algorithm = [&]() {
+        // 6 => If the name member of normalizedAlgorithm is a case-sensitive string match for "Argon2d":
+        //      Let type be 0.
+        if (normalized_algorithm.name.equals_ignoring_ascii_case("Argon2d"_string))
```
shannonbooth: not a extensive review, I was just glancing through, but this should be a case **sensitive** match

tete17: Indeed, thank you for the glancing :)

Libraries/LibWeb/Crypto/CryptoAlgorithms.cpp
```diff
@@ -9146,4 +9146,10 @@ WebIDL::ExceptionOr<EncapsulatedBits> MLKEM::encapsulate(AlgorithmParams const&
     return result;
 }
 
+// https://wicg.github.io/webcrypto-modern-algos/#argon2-operations-get-key-length
+WebIDL::ExceptionOr<JS::Value> Argon2::get_key_length(AlgorithmParams const&)
+{
+    return JS::js_null();
```
shannonbooth, line 9318: Nit, but let's leave the:

> // 1. Return null.

Comment

Libraries/LibCrypto/Hash/Argon2.cpp
```diff
@@ -0,0 +1,70 @@
...
+    u32 passes,
+    u32 version,
+    Optional<ByteBuffer> secret_value,
+    Optional<ByteBuffer> associated_data,
+    u32 tag_length) const
+{
+
+    auto ctx = TRY(OpenSSL_KDF_CTX::wrap(EVP_KDF_CTX_new(m_kdf)));
+
+    OPENSSL_TRY(OSSL_set_max_threads(nullptr, parallelism));
```
shannonbooth: This doesn't seem right - from the shape of this API it seems like we are touching a global OpenSSL state. Should this not _somehow_ be associated with the hash algorithm itself (or put in a different place altogether if needed)

Libraries/LibCrypto/Hash/Argon2.cpp
```diff
@@ -0,0 +1,70 @@
...
+}
+
+ErrorOr<ByteBuffer> Argon2::derive_key(
+    ReadonlyBytes message,
+    ReadonlyBytes nonce,
+    u32 parallelism,
+    u32 memory,
+    u32 passes,
+    u32 version,
+    Optional<ByteBuffer> secret_value,
```
shannonbooth: These do not need to be copies, probably could even be converted over to `Optional<ReadonlyBytes>`

Libraries/LibCrypto/Hash/Argon2.h
```diff
@@ -0,0 +1,40 @@
...
+#include <AK/Error.h>
+#include <LibCrypto/OpenSSLForward.h>
+
+namespace Crypto::Hash {
+
+class Argon2 {
+    AK_MAKE_NONCOPYABLE(Argon2);
+
+public:
+    Argon2(char const* openssl_name);
```
shannonbooth: Instead of exposing the OpenSSL name in the public API, maybe this could instead be an enum class which is then mapped to the correct name?

Libraries/LibCrypto/Hash/Argon2.h
```diff
@@ -0,0 +1,40 @@
...
+
+    virtual ~Argon2()
```
shannonbooth: This shouldn't need to be virtual (and I think we may as well put this out of line)

shannonbooth (changes requested):

Libraries/LibCrypto/Hash/Argon2.cpp
tete17: Happy to take it out but I think in general it should be fine.

This api just sets a limit not a target. https://docs.openssl.org/3.6/man3/CRYPTO_THREAD_run_once/#description:~:text=a%20uint64_t%20value.-,OSSL_set_max_threads,-()%20sets%20the%20maximum

The parallelism is anyway the specified in the openssl alg anyway. My understanding is that it will try to use as many threads as specified in the algorithm with an upper bound of the max threads. Removing this api call will just make the process sequential, which will still work but may slow down the thing substantially.

tete17: Brain storming here but we can also just set it if the current value is lower than needed?

shannonbooth: Even though this is only invoked by javascript in a single thread, it still seems wrong to be putting this API call here, since it would fall apart if this function is invoked in another thread (and we should probably not let JS increase the max size of threads to something crazy). If we want to be increasing the number of max threads that OpenSSL can use, probably this should live in `main()`, maybe based on the hardware concurrency.

tete17: I come back with some interesting findings.

If we specify the parallelism, as the standard requires, we need to set a high enough value of max threads or else we get this error.

```
400F0C589C730000:error:1C8000EA:Provider routines:kdf_argon2_derive:invalid thread pool size:../src/nssl-3.5.3-78bf900265.clean/providers/implementations/kdfs/argon2.c:1073:requested 4 threads, available: 1
```

In my head there is many ways we could skin this cat:
 1. We could check the number of threads before running the algorithm and if it is lower than desired increase it.
 2. We could continue with your idea to match it with something like `std::threads::hardware_concurrency()` in main but then we need to deviate from the standard and make sure that the thread field is lower or equal than the value set at main
 3. We could kick this can down the road and set the thread value always to 0

But I wanted to pick your brain first

shannonbooth: > If we specify the parallelism, as the standard requires, we need to set a high enough value of max threads or else we get this error.
> We could kick this can down the road and set the thread value always to 0


So OpenSSL will allow any parallelism to be set for Argon2 if OSSL_set_max_threads is 0 and won't allow parallelism above max threads if OSSL_set_max_threads is above zero?

shannonbooth: But I think:

> We could check the number of threads before running the algorithm and if it is lower than desired increase it.

Does seem best, can set `OSSL_KDF_PARAM_THREADS` to `min(parallelism, Core::System::hardware_concurrency())`, doesn't make sense for it to be any higher anyhow. edit: oh wait, that is against standard? how so? might be missing something

tete17: Never mind as soon as I wrote this I realized I was wrong.

The spec specifies the lanes it obviouslly doesn't require to run them all at the same time since many people won't have that many threads.

We can simply query openssl for the current max number of threads and optionally set that in the algorithm.
The good thing is this is very flexible and works wether openssl threads are enable or not.

I added an extra commit to set the number of threads in the main of the processes that contain a JS intepreter since we still want the speed if we can have it.

Services/WebWorker/main.cpp
```diff
@@ -55,6 +59,7 @@ ErrorOr<int> ladybird_main(Main::Arguments arguments)
     args_parser.add_option(certificates, "Path to a certificate file", "certificate", 'C', "certificate");
     args_parser.add_option(wait_for_debugger, "Wait for debugger", "wait-for-debugger");
     args_parser.add_option(worker_type_string, "Type of WebWorker to start (dedicated, shared, or service)", "type", 't', "type");
+    args_parser.add_option(max_number_of_openssl_threads, "Maximum number of OpenSSL threads", "max-openssl-threads", 0, "max_openssl_threads");
```
shannonbooth: I don't think we will need a configurable number of threads like this (and if we did, this would require a bit more wiring up anyhow to pass through to the WebContent/WebWorker processes)

Libraries/LibCrypto/Hash/Argon2.cpp
```diff
@@ -0,0 +1,95 @@
...
+    u32 passes,
+    u32 version,
+    Optional<ReadonlyBytes> secret_value,
+    Optional<ReadonlyBytes> associated_data,
+    u32 tag_length) const
+{
+
+    auto ctx = TRY(OpenSSL_KDF_CTX::wrap(EVP_KDF_CTX_new(m_kdf)));
+
+    auto const hardware_concurrency = OSSL_get_max_threads(nullptr);
```
shannonbooth: The name here is a little misleading since max threads from this context is not the hardware concurrency - only the maximum number of threads

Libraries/LibCrypto/Hash/Argon2.cpp
```diff
@@ -0,0 +1,95 @@
...
+#include <openssl/core_names.h>
+#include <openssl/kdf.h>
+#include <openssl/params.h>
+#include <openssl/thread.h>
+
+namespace Crypto::Hash {
+
+char const* argon2_type_to_openssl_name(Argon2Type type)
+{
+
```
shannonbooth: nit, but stray newline, and can be file local/static

Services/WebWorker/main.cpp
tete17: Cool with hard-coding, less things to care about

Libraries/LibCrypto/Hash/Argon2.cpp
tete17: I inlined the variable, that way we don't need to name it

Libraries/LibCrypto/Hash/Argon2.cpp
tete17: indeed

PR #6966, opened 2025-11-28
AK+LibURL: Use AK::IPv4/6 in Host

ion232:
- This resolves two FIXME comments and improves code reuse.
- Changed the term `group` to `piece` in IPv6 to better match the spec.
- The underlying data could also be changed to Array<u16, 8> to better match the spec. I can make the change if you think this is worth doing.

### DIFF
--- a/AK/IPv6Address.h
+++ b/AK/IPv6Address.h
@@ -6,6 +6,7 @@
 
 #pragma once
 
+#include <AK/BitCast.h>
 #include <AK/Endian.h>
 #include <AK/Format.h>
 #include <AK/IPv4Address.h>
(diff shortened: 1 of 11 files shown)


AK/IPv6Address.h
```diff
@@ -36,6 +37,13 @@ class [[gnu::packed]] IPv6Address {
             m_data[i] = data[i];
     }
 
+    constexpr IPv6Address(Array<u16, 8> const& data)
+    {
+        u16* pieces = reinterpret_cast<u16*>(m_data);
```
shannonbooth: Sorry for the delay! Only comment is that this reinterpret cast is UB and we should do the shift approach instead.

On your question of:
> The underlying data could also be changed to Array<u16, 8> to better match the spec. I can make the change if you think this is worth doing.

I think it's okay as is. Eventually this will end up diverging away from the exact spec text anyway, as I expect our very unoptimized URL parsing will be optimized at some stage going forward

shannonbooth (changes requested):

AK/IPv6Address.h
ion232: No problem, thanks for the review. I've fixed the UB now

PR #6970, opened 2025-11-28
Integrate partially ML-KEM postquantum algorithm into WebCryptoAPI

tete17:
This PR contains 3 things: 
 * A few bug fixes from the previous PR https://github.com/LadybirdBrowser/ladybird/pull/6961
 * The Boilerplate code to connect OpenSSL and WebCryptoAPI with ML-KEM
 * An AD-HOC change to spec in the hopes to fix a potential issue https://github.com/WICG/webcrypto-modern-algos/issues/36

The implement is not complete but I though I had a good amount of changes/commits to make the first PR to avoid making them massive.


Libraries/LibCrypto/PK/MLKEM.h
```diff
@@ -0,0 +1,66 @@
...
+#include <AK/ByteBuffer.h>
+#include <AK/Error.h>
+
+namespace Crypto::PK {
+
+enum MLKEMSize {
+    MLKEM512,
+    MLKEM768,
+    MLKEM1024,
+};
```
shannonbooth: It looks like the same thing was done in MLDSA, but `enum class` for this

Libraries/LibCrypto/PK/MLKEM.h
```diff
@@ -0,0 +1,66 @@
...
+
+class MLKEMPublicKey {
+public:
+    MLKEMPublicKey(ByteBuffer m_public_key)
```
shannonbooth: Explicit here for single argument constructor

Libraries/LibWeb/Crypto/CryptoAlgorithms.cpp
```diff
@@ -9009,6 +9009,96 @@ WebIDL::ExceptionOr<Variant<GC::Ref<CryptoKey>, GC::Ref<CryptoKeyPair>>> MLKEM::
...
 
+// https://wicg.github.io/webcrypto-modern-algos/#ml-kem-operations-import-key
+WebIDL::ExceptionOr<GC::Ref<CryptoKey>> MLKEM::import_key(AlgorithmParams const& params, Bindings::KeyFormat key_format, CryptoKey::InternalKeyData key_data, bool extractable, Vector<Bindings::KeyUsage> const& usages)
+{
+    GC::Ptr<CryptoKey> key = nullptr;
+
+    // 1. Let keyData be the key data to be imported.
+    // 2. -> If format is "spki":
+    if (key_format == Bindings::KeyFormat::Spki) {
+        // FIXME: Spec starts at 2
```
shannonbooth: Has the spec been updated or am I missing something here about this FIXME?

Libraries/LibCrypto/PK/MLKEM.cpp
```diff
@@ -0,0 +1,69 @@
...
+    auto* key_ptr = key.ptr();
+    OPENSSL_TRY(EVP_PKEY_generate(ctx.ptr(), &key_ptr));
+
+    // We reserve enough memory for the key size to be able to fit them all
+    auto pub = TRY(ByteBuffer::create_uninitialized(1568));
+    auto priv = TRY(ByteBuffer::create_uninitialized(3168));
+    seed = TRY(ByteBuffer::create_uninitialized(64));
+
+    size_t priv_len, pub_len, seed_len;
+    OPENSSL_TRY(EVP_PKEY_get_octet_string_param(key.ptr(), OSSL_PKEY_PARAM_ML_KEM_SEED, seed.data(), seed.size(), &seed_len));
```
shannonbooth: Considering this pattern is also repeated in MLDSA, maybe it makes sense we should make a ByteBuffer wrapper of this function. And instead of hardcoding the maximum size of 3168 as an example, that wrapper can invokeEVP_PKEY_get_octet_string_param to first get the required buffer size, and then again with the properly sized ByteBuffer.

Libraries/LibWeb/Crypto/CryptoAlgorithms.cpp
tete17: Seems to have been fixed https://github.com/WICG/webcrypto-modern-algos/commit/4e3e644602243f401edd52ebdbb504c1489546a1

I will remove this comment from the commit

Libraries/LibCrypto/PK/MLKEM.cpp
tete17: How does `get_byte_buffer_param_from_key` sound?

shannonbooth (approved):

PR #6973, opened 2025-11-29
LibGfx+LibWeb: Fix compile errors in clang-cl from recent header cleanup

InvalidUsernameException:
The recent commits 28ba610f325aaed5d20daebdd3cf732a04e62eaf and 70c4ed261fb446fe25396164400d09e0c0df8bc6 adjusted some include directives to avoid excessive recompilation when changing some header files. This has broken compilation with clang-cl on Windows without getting noticed before the PRs were merged.

### DIFF
--- a/Libraries/LibWeb/CSS/CSSImportRule.h
+++ b/Libraries/LibWeb/CSS/CSSImportRule.h
@@ -13,6 +13,7 @@
 #include <LibWeb/CSS/URL.h>
 #include <LibWeb/DOM/DocumentLoadEventDelayer.h>
 #include <LibWeb/Export.h>
+#include <LibWeb/Forward.h>
 
 namespace Web::CSS {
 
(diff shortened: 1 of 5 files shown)


Libraries/LibWeb/CSS/CSSImportRule.h
```diff
@@ -10,6 +10,7 @@
 
 #include <LibWeb/CSS/CSSRule.h>
 #include <LibWeb/CSS/CSSStyleSheet.h>
+#include <LibWeb/CSS/Supports.h>
```
trflynn89: I think `Supports` should be fine to use as a forward declaration in this header if we outline the destructor? Like you did with `ImageBitmap`

InvalidUsernameException: Yeah, I guess I can do that. `CSSImportRule.h` only causing 10 files to be rebuilt is why I didn't initially.

nico:
I'm guessing that's because WEB_API forces emission of inlines. If you aren't already, you could consider using https://blog.llvm.org/2018/11/30-faster-windows-builds-with-clang-cl_14.html to make the win build more like the other platforms (makes the build faster, too).

InvalidUsernameException:
Looks like that flag is already used:

https://github.com/LadybirdBrowser/ladybird/blob/c1c24e8fd6fd3302d164e4198458d837523b2695/Meta/CMake/common_compile_options.cmake#L97

PR #6975, opened 2025-11-29
LibJS: Handful of optimizations for fromBase64/toBase64 performance

awesomekling:
I ran https://simdutf.github.io/browserbase64/ and got some pretty disappointing results in Ladybird.

With the following changes, the 16.0 KiB buffer size performance improves thusly:
- 3.8x faster fromBase64()
- 38.4x faster toBase64()

Note that none of them actually deal with actual encoder/decoder throughput, that stuff is already fast thanks to simdutf :smile:

We were just doing poorly on the test because of silly/inefficient things in our runtime.

### DIFF
--- a/Libraries/LibJS/Runtime/Uint8Array.cpp
+++ b/Libraries/LibJS/Runtime/Uint8Array.cpp
@@ -333,22 +333,33 @@ JS_DEFINE_NATIVE_FUNCTION(Uint8ArrayPrototypeHelpers::to_base64)
     }
 
     // 8. Let toEncode be ? GetUint8ArrayBytes(O).
-    auto to_encode = TRY(get_uint8_array_bytes(vm, typed_array));
-
     String out_ascii;
 
-    // 9. If alphabet is "base64", then
-    if (alphabet == Alphabet::Base64) {
-        // a. Let outAscii be the sequence of code points which results from encoding toEncode according to the base64
-        //    encoding specified in section 4 of RFC 4648. Padding is included if and only if omitPadding is false.
-        out_ascii = MUST(encode_base64(to_encode, omit_padding));
-    }
-    // 10. Else,
-    else {
-        // a. Assert: alphabet is "base64url".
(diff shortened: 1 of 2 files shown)


Libraries/LibJS/Runtime/Uint8Array.cpp
```diff
@@ -333,22 +333,33 @@ JS_DEFINE_NATIVE_FUNCTION(Uint8ArrayPrototypeHelpers::to_base64)
...
-    }
-    // 10. Else,
-    else {
-        // a. Assert: alphabet is "base64url".
-        // b. Let outAscii be the sequence of code points which results from encoding toEncode according to the base64url
-        //    encoding specified in section 5 of RFC 4648. Padding is included if and only if omitPadding is false.
-        out_ascii = MUST(encode_base64url(to_encode, omit_padding));
+    // OPTIMIZATION: If the ArrayBuffer is not shared, we can avoid copying the bytes.
+    if (!typed_array->viewed_array_buffer()->is_shared_array_buffer()
+        && typed_array->viewed_array_buffer()->has_buffer()) {
```
trflynn89: I think you could use `!is_detached`?

Libraries/LibJS/Runtime/PrimitiveString.cpp
```diff
@@ -21,6 +21,10 @@
 
 namespace JS {
 
+// Strings shorter than or equal to this length are cached in the VM and deduplicated.
```
trflynn89, line 24: Typo in commit: `hasing`

Libraries/LibJS/Runtime/Uint8Array.cpp
trflynn89: I think this is observable as well, we're skipping a check for out-of-bounds TAs here.

awesomekling: Yeah duh, it's even tested by test-js, duh. Will fix

PR #6976, opened 2025-11-29
LibGfx: Refactor Vulkan code and fix some bugs

cqundefine:
This PR makes VulkanContext and VulkanImage closer to the Ladybird C++ code style which also prevents exposed member variables. I've tried enabling validation layers which actually turned out to be quite helpful and exposed a few bugs. I've left that code there locked behind a debug macro.

### DIFF
--- a/Meta/check-debug-flags.sh
+++ b/Meta/check-debug-flags.sh
@@ -42,6 +42,7 @@ done < <(
         find_all_source_files "$@"
     fi \
     | xargs grep -E '(_DEBUG|DEBUG_)' \
+    | grep -v 'VK_' \
     | sed -re 's,^.*[^a-zA-Z0-9_]([a-zA-Z0-9_]*DEBUG[a-zA-Z0-9_]*).*$,\1,' \
     | sort -u
 )


Libraries/LibGfx/VulkanContext.cpp
```diff
@@ -1,79 +1,152 @@
...
+VulkanContext::~VulkanContext()
+{
+    if (m_logical_device != VK_NULL_HANDLE) {
+        vkDestroyDevice(m_logical_device, nullptr);
+    }
+
+    if (m_debug_messenger != VK_NULL_HANDLE) {
+        auto pfn_destroy_debug_messenger = reinterpret_cast<PFN_vkDestroyDebugUtilsMessengerEXT>(vkGetInstanceProcAddr(m_instance, "vkDestroyDebugUtilsMessengerEXT"));
+        pfn_destroy_debug_messenger(m_instance, m_debug_messenger, nullptr);
+    }
```
rcorsi, line 49: Shouldn't this be behind VULKAN_DEBUG because of m_debug_messenger?  Why didn't the CI highlight this?

Libraries/LibGfx/VulkanContext.cpp
```diff
@@ -1,79 +1,152 @@
...
     VkApplicationInfo app_info {};
     app_info.sType = VK_STRUCTURE_TYPE_APPLICATION_INFO;
     app_info.pApplicationName = "Ladybird";
     app_info.applicationVersion = VK_MAKE_VERSION(1, 0, 0);
     app_info.pEngineName = nullptr;
     app_info.engineVersion = VK_MAKE_VERSION(1, 0, 0);
-    app_info.apiVersion = api_version;
+    app_info.apiVersion = m_api_version;
+
+#ifdef VULKAN_DEBUG
```
rcorsi: all the "ifdef VULKAN_DEBUG" are in other parts of Ladybird as "if constexpr (VULKAN_DEBUG) {" blocks. In the cpp it should be easy enough, but in the header might be a problem. Not sure what to suggest for this. Just wondering if you thought of this?

Libraries/LibGfx/VulkanContext.cpp
cqundefine: The issue was that VULKAN_DEBUG was always defined, it was just meant to be used as 0/1. And yes this should be behind VULKAN_DEBUG.

Libraries/LibGfx/VulkanContext.cpp
cqundefine: In some places it is actually `#if`. I don't it matters that much in the end as both achieve the same and I think consistently using `#if` everywhere is better than using `if constexpr` in some places only as in some places it's not possible to use it.

rcorsi:
When `vulkan-validationlayers` package (on Ubuntu) is not present you get the error below that prevents VulkanContext creation.  Is that what you want or might it be better to just print a warning, and allow the VulkanContext creation, for example if in CI running test of All_Debug (currently not merged) that might impact some test that might need the VulkanContext? I don't know what the right thing is, just highlighting it to have you consider it. 

```
13055.044 WebContent(141245): Vulkan context creation failed: Request Vulkan validation layer not supported
13055.063 WebContent(141249): Vulkan context creation failed: Request Vulkan validation layer not supported
```

Also the message should be `Requested Vulkan validation layer not supported`, missing the `ed` at end of Request. 

Also what about `Requested Vulkan validation layer not available` or `not installed`.  Would that be better?

or  `Requested Vulkan validation layer not available, please install Vulkan SDK`.

Should we be mentioning any distribution packages needing to be installed or Vulkan SDK installation requirements somewhere in the ladybird documentation.

rcorsi:
Strictly speak, is it required to modify any code to get the validation layer messages? Can't that specific part of it be achieved with the Vulkan Configurator from the Vulkan SDK? Is there any additional benefits to putting it in the code.

I think that would allow you to remove all the VULKAN_DEBUG items.

cqundefine:
I changed the message and ended up making the lack of validation layer a non-critical failure. I wonder if it should remain obligatory as `VULKAN_DEBUG` would have to be explicitly enabled by someone, but if it's extended to do other stuff too it probably shouldn't prevent Vulkan from being used altogether.

> Strictly speak, is it required to modify any code to get the validation layer messages? Can't that specific part of it be achieved with the Vulkan Configurator from the Vulkan SDK? Is there any additional benefits to putting it in the code.

I believe the main benefit is convenience and not requiring the use of external tools. I do frequent Vulkan development and I actually don't have the Vulkan SDK installed at all. I think looking at how it's a relatively small amount of code I would be more inclined to keep it. I think it would be great to get an opinion on a maintainer on this.

rcorsi:
I would go to discord to try to get some attention on this one. It is a PR covering multiple Vulkan issues and with many commits so requires more dedicated attention. Not always possible, but simpler PRs are usually easier to digest during a review.

One thing I would have found useful is to only turn on the validation layers through code (without the other fixes) so as to check the messages myself. I know I could cherry-pick, and I quickly tried and got a merge conflict that I didn't try to solve at all (probably not that hard to resolve), but yeah that commit first in the stack would be ideal or maybe in its own PR separate from the actual Vulkan fixes. You don't have to do it, just kind of an observation, or you could have attached a log with the validation layer messages you got.

I realized soon after I left my post that having the validation layers in code was very convenient. No SDK needed, like you said.  I remember trying to use for the first time the Vulkan Configurator and it took some time (not that much though) to get it working right.

Regarding the VULKAN_DEBUG, I had the same thought about what if someone wants to use it for something not validation layer related.  Should there be a VULKAN_DEBUG macro (for general Vulkan debug messages) and a VULKAN_VALIDATION_LAYER_DEBUG macro to separate the functions, you might not want the two together?

Some general comments about the commit messages:

`Meta: Don't mistake Vulkan defines us our debug macros`, should the part `us our` really be `as`, so rewritten to `Meta: Don't mistake Vulkan defines as debug macros`.

Regarding the above commit, did you try to follow the instructions in Meta/check-debug-flags.sh to add the Vulkan defines in the false positives in `Meta/CMake/all_the_debug_macros.cmake` instead of the grep change to exclude VK_? No idea if this is easy or possible, just wondering if you gave it a try.

`LibGfx: Remove incorrect Vulakn queue family index from image creation` typo in the word Vulkan.

`AK+LibGfx: Add a VULKAN_DEBUG option that enables validations layers`, I don't think AK needs to be mentioned on this one. Not really an important change to AK.

rcorsi:
Hi @cqundefine, I just used the validation layer feature you added to troubleshoot #7207 and it gave me some output.  Any chance of you just breaking out only the validation layer feature code in a very specific PR so that at least that change can get merged? I would say it is a very useful debug feature to have and would be a good start to merging things.

A few things I would add / change if I could
1- Name the debug macro VULKAN_VALIDATION_LAYER_DEBUG
2- When a validation layer issue is triggered, dump a backtrace or at least the file and line origin.
3- Maybe this is too much, but item 2 from this list could be controlled with another debug macro VULKAN_VALIDATION_LAYER_BACKTRACE_DEBUG. Just an idea right now, don't know how important that would be.

If you don't have the time or inclination, would you be okay with me doing it, with you being the author of course and I will be in the `Co-authored-by:`?

cqundefine:
> Hi @cqundefine, I just used the validation layer feature you added to troubleshoot #7207 and it gave me some output. Any chance of you just breaking out only the validation layer feature code in a very specific PR so that at least that change can get merged? I would say it is a very useful debug feature to have and would be a good start to merging things.
> 
> A few things I would add / change if I could 1- Name the debug macro VULKAN_VALIDATION_LAYER_DEBUG 2- When a validation layer issue is triggered, dump a backtrace or at least the file and line origin. 3- Maybe this is too much, but item 2 from this list could be controlled with another debug macro VULKAN_VALIDATION_LAYER_BACKTRACE_DEBUG. Just an idea right now, don't know how important that would be.
> 
> If you don't have the time or inclination, would you be okay with me doing it, with you being the author of course and I will be in the `Co-authored-by:`?

I've split of the validation layer part (and only that) in #7293. I did change the macro name to `VULKAN_VALIDATION_LAYER_DEBUG` as it is indeed more descriptive and I've added a call to `dump_backtrace`. I'm not sure a separate macro would make sense for enabling backtraces as if you're using the validation layer the backtrace will most definitely be also useful. If you have any other feedback regarding that feel free to comment on that PR!

cqundefine:
(sorry didn't mean to request the review again)

PR #6978, opened 2025-11-30
LibWeb: Random WebGL improvements

cqundefine:


Libraries/LibWeb/WebGL/WebGLRenderingContextImpl.h
```diff
@@ -165,6 +165,9 @@ class WebGLRenderingContextImpl : public WebGLRenderingContextBase {
     GC::Ptr<WebGLTexture> m_texture_binding_3d;
     GC::Ptr<WebGLTransformFeedback> m_transform_feedback_binding;
     GC::Ptr<WebGLVertexArrayObject> m_current_vertex_array;
+    GC::Ptr<WebGLQuery> m_any_samples_passed;
+    GC::Ptr<WebGLQuery> m_any_samples_passed_conservative;
+    GC::Ptr<WebGLQuery> m_transform_feedback_primitives_written;
```
gmta, line 172: Shouldn't we visit these in `visit_edges`?

Libraries/LibWeb/WebGL/Extensions/OESElementIndexUint.idl
```diff
@@ -0,0 +1,9 @@
+// https://registry.khronos.org/webgl/extensions/OES_element_index_uint/
+// NOTE: Original OES_element_index_uint name is changed to title case,
+//       so it matches corresponding C++ class name, and does not require
+//       IDL generator to handle snake_case to TitleCase conversion.
+//       Having a different name is totally fine, because LegacyNoInterfaceObject
+//       prevents the name from being exposed to JavaScript.
```
gmta: Please reformat this comment to wrap at column 120. Also, should this be a FIXME for IDL generator?

cqundefine: We have the exact same comment in every single other extension IDL file, that's why it was wrapped the way that it was. Not sure what you mean by FIXME for IDL generator as this is all working as intended. The only thing is that in this example in the original IDL from the spec the actual interface name is `OES_element_index_uint`, while we change it to `OESElementIndexUint` to better match our naming scheme. Those comments in those files are there to note that changing the name is fine as it's invisible to JavaScript code.

Libraries/LibWeb/WebGL/WebGLRenderingContextImpl.cpp
```diff
@@ -885,6 +905,18 @@ JS::Value WebGLRenderingContextImpl::get_parameter(WebIDL::UnsignedLong pname)
         auto array_buffer = JS::ArrayBuffer::create(m_realm, move(byte_buffer));
         return JS::Float32Array::create(m_realm, 4, array_buffer);
     }
+    case GL_COLOR_WRITEMASK: {
+        Array<GLboolean, 4> result;
+        result.fill(0);
+        glGetBooleanvRobustANGLE(GL_COLOR_CLEAR_VALUE, 4, nullptr, result.data());
+
+        auto sequence = MUST(JS::Array::create(m_realm, 0));
```
konradekk:
```suggestion
        auto sequence = MUST(JS::Array::create(m_realm, 4));
```

Libraries/LibWeb/WebGL/WebGLRenderingContextImpl.cpp
```diff
@@ -886,6 +887,18 @@ JS::Value WebGLRenderingContextImpl::get_parameter(WebIDL::UnsignedLong pname)
         auto array_buffer = JS::ArrayBuffer::create(realm(), move(byte_buffer));
         return JS::Float32Array::create(realm(), 4, array_buffer);
     }
+    case GL_COLOR_WRITEMASK: {
+        Array<GLboolean, 4> result;
+        result.fill(0);
+        glGetBooleanvRobustANGLE(GL_COLOR_CLEAR_VALUE, 4, nullptr, result.data());
+
+        auto sequence = MUST(JS::Array::create(realm(), 4));
```
ADKaster: Is MUST correct here? is there a way for this method to set_error to ENOMEM or something? I see many other paths that MUST a byte buffer copy, but doing so on a js object feels awkward.

MUST for create_data_property as well feels uncomfortable.

This can probably be fixed in a follow-up or pushed aside with a FIXME though, the method returns JS::Value rather than an ExceptionOr type.

ADKaster (commented):
These changes all seem reasonable to me, though I have one comment about `getParameter(COLOR_WRITEMASK)` that I don't know the right answer for.

If @Lubrsi is fine with the current impl then we can merge this as-is, otherwise that commit might need some extra eyes.

cqundefine:
The Linux CI failure seems unrelated, it looks like `add-apt-repository` is failing for some reason.
(rebased on master to trigger a rerun, but it's still broken)

PR #6985, opened 2025-12-01
LibWeb: Make setInterval() reuse the timer to reduce drift

awesomekling:
Instead of creating a new single-shot timer every time a setInterval timer reschedules itself, we now use a repeating Core::Timer internally.

This dramatically reduces timer drift, since the next timeout is now based on when the timer fired, rather than when the timer callback completed (which could take arbitrarily long since we run JS).

It's not perfect, but it's a huge improvement and allows us to play DiabloWeb at full framerate (20 fps).

BEFORE (~13 fps):

https://github.com/user-attachments/assets/6697973a-b00d-40c7-a901-5b65b69e3f66

AFTER (20 fps):

https://github.com/user-attachments/assets/11ad8072-cb05-45b9-a439-53bacd4096f7

### DIFF
--- a/Libraries/LibWeb/HTML/Timer.h
+++ b/Libraries/LibWeb/HTML/Timer.h
@@ -22,20 +22,26 @@ class Timer final : public JS::Cell {
     GC_DECLARE_ALLOCATOR(Timer);
 
 public:
-    static GC::Ref<Timer> create(JS::Object&, i32 milliseconds, Function<void()> callback, i32 id);
+    enum class Repeating {
+        No,
+        Yes,
+    };
+
+    static GC::Ref<Timer> create(JS::Object&, i32 milliseconds, Function<void()> callback, i32 id, Repeating);
     virtual ~Timer() override;
 
     void start();
     void stop();
 
+    void set_callback(Function<void()>);
+
 private:
-    Timer(JS::Object& window, i32 milliseconds, GC::Ref<GC::Function<void()>> callback, i32 id);
+    Timer(JS::Object& window, i32 milliseconds, Function<void()> callback, i32 id, Repeating);
 
     virtual void visit_edges(Cell::Visitor&) override;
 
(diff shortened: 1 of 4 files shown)


Hendiadyoin1:
Phone review,
But what happens when the js takes long enough to miss an interval,
Or to be more precise what should happen then,
If I understood the change correctly we hit the due time immediately and run again, while the previous version somewhat used the next interval,
Should it instead skip that epoch?

Libraries/LibWeb/HTML/Timer.h
```diff
@@ -22,20 +22,26 @@ class Timer final : public JS::Cell {
     GC_DECLARE_ALLOCATOR(Timer);
 
 public:
-    static GC::Ref<Timer> create(JS::Object&, i32 milliseconds, Function<void()> callback, i32 id);
+    enum class Repeating {
```
R-Goc, line 25: On windows the default enum is an int. 
```suggestion
    enum class Repeating : u8 {
```

trflynn89: `int` is the default underlying type everywhere. But we typically only specify a specific storage if it's actually helpful, i.e. to reduce data sent over IPC. What would be the benefit here?

R-Goc: Lower argument size and not seeing the clang-tidy warning. Nothing major though.

awesomekling:
> Phone review, But what happens when the js takes long enough to miss an interval, Or to be more precise what should happen then, If I understood the change correctly we hit the due time immediately and run again, while the previous version somewhat used the next interval, Should it instead skip that epoch?

If it takes long enough to miss one interval, we simply schedule the next interval ASAP.

We never batch multiple missed callbacks.

This is consistent with other engines, as far as I understand.

Chromium does additional stuff to maintain symmetry with the original start time, but I'm leaving that as a FIXME for our future selves :)

PR #6987, opened 2025-12-01
LibWeb/CSS: Allow non-StyleValueLists in animation properties

AtkinsSJ:
Reverts 51f694c6af033336a8608598c96e4c343cdec64c and b52beb5105f149a7100a2b35a9ec6a268b3c1fe4.

The animation-* properties are in an awkward place currently, where they *should* be lists, and a lot of our code acts as if they are, but actually we parse them as single values. The above commits caused a few WPT tests to crash - see link below. I've imported one of them to prevent future regressions.
https://wpt.fyi/results/css/css-typed-om/the-stylepropertymap/properties?diff&filter=ADC&run_id=6293362870321152&run_id=5123272984494080


AtkinsSJ:
cc @Calme1709: Apparently there's a bit more to do before we can get rid of this entirely. :sweat_smile:

PR #6989, opened 2025-12-01
LibWeb: Resolve cursor to i-beam over contenteditable container

jonbgamble:
This PR aims to show the I-beam (indicating a click will result in caret insertion) when hovering within an editing host but outside of a textNode.

### DIFF
--- a/Libraries/LibWeb/Page/EventHandler.cpp
+++ b/Libraries/LibWeb/Page/EventHandler.cpp
@@ -809,7 +809,7 @@ EventResult EventHandler::handle_mousemove(CSSPixelPoint visual_viewport_positio
             if (hovered_link_element)
                 is_hovering_link = true;
 
-            if (paintable->layout_node().is_text_node()) {
+            if (paintable->layout_node().is_text_node() || node->is_editing_host()) {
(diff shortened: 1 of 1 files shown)


Libraries/LibWeb/Page/EventHandler.cpp
```diff
@@ -809,7 +809,7 @@ EventResult EventHandler::handle_mousemove(CSSPixelPoint visual_viewport_positio
             if (hovered_link_element)
                 is_hovering_link = true;
 
-            if (paintable->layout_node().is_text_node()) {
+            if (paintable->layout_node().is_text_node() || node->is_editing_host()) {
```
tcl3, line 812: I think you could use `is_editable_or_editing_host()` to account for `contenteditable` elements.

I tested with this code, which shows an i-beam over the image in other browsers:

```html
<!DOCTYPE html>
<p contenteditable>This text is editable<img src="test.png"></img>
</p>
```

jonbgamble: My first PR! Yeah, it was nice to effect a visible change in a new codebase but I suspect a lot of the editing code will be heavily refactored (if not rewritten) before alpha and I'm not sure it's a good idea to burden maintainers with cosmetics like this when they could be doing more important stuff.

PR #6993, opened 2025-12-02
LibCore: Remove the dummy EventReceiver from deferred_invoke()

Zaggy1024:
The DeferredInvocationContext only existed to satisfy the requirement in ThreadEventQueue that each event has an EventReceiver. However, deferred_invoke() was not even using the EventReceiver to call its callback. Therefore, we don't need to allocate one for every deferred invocation.

This also prevents WeakPtr::strong_ref() from racing and leaking the context object when invoking a function across threads.


R-Goc:
Could this get tested against the Windows CI as well?

Zaggy1024:
Ah, good point, I'll add the windows label.

PR #7011, opened 2025-12-03
LibCore: Explicitly export symbols in LibCore

R-Goc:
This patch removes LibCoreMinimal and enables explicit symbol export for LibCore. This simplifies the cmake and reduces the total number of symbols loaded.


Libraries/LibCore/CMakeLists.txt
```diff
@@ -25,6 +25,18 @@ else()
...
+    target_compile_options(LibCoreMinimal
+            PRIVATE
+                "$<$<COMPILE_LANGUAGE:CXX>:-fvisibility=hidden>"
+                "$<$<COMPILE_LANGUAGE:C>:-fvisibility=hidden>"
+        )
+else()
+    set_target_properties(LibCoreMinimal PROPERTIES WINDOWS_EXPORT_ALL_SYMBOLS OFF)
+endif()
+include(GenerateExportHeader)
+generate_export_header(LibCoreMinimal EXPORT_MACRO_NAME COREMINIMAL_API EXPORT_FILE_NAME "ExportCoreMinimal.h")
```
ayeteadoe: We can instead update `lagom_generate_export_header()` from `targets.cmake`  to take an optional parameter that lets us override the output header in the `generate_export_header()` call:

<details>

<summary>generate-export-header-expose-file-param.patch</summary>

```patch
diff --git a/Meta/CMake/targets.cmake b/Meta/CMake/targets.cmake
index 0f7e26d5fd..837710b6c0 100644
--- a/Meta/CMake/targets.cmake
+++ b/Meta/CMake/targets.cmake
@@ -4,6 +4,10 @@ function(lagom_generate_export_header name fs_name)
     # Temporary helper to allow libraries to opt-in to using X_API macros
     # to export symbols required by external consumers. This allows the codebase
     # to gradually slowly migrate instead of an all-or-nothing approach.
+    cmake_parse_arguments(LAGOM_GENERATE_EXPORT_HEADER "" "EXPORT_FILE_NAME" "" ${ARGN})
+    if (NOT LAGOM_GENERATE_EXPORT_HEADER_EXPORT_FILE_NAME)
+        set(LAGOM_GENERATE_EXPORT_HEADER_EXPORT_FILE_NAME "Export.h")
+    endif()
     if (NOT WIN32)
         target_compile_options(${name}
             PRIVATE
@@ -15,7 +19,7 @@ function(lagom_generate_export_header name fs_name)
     endif()
     include(GenerateExportHeader)
     string(TOUPPER ${fs_name} fs_name_upper)
-    generate_export_header(${name} EXPORT_MACRO_NAME ${fs_name_upper}_API EXPORT_FILE_NAME "Export.h")
+    generate_export_header(${name} EXPORT_MACRO_NAME ${fs_name_upper}_API EXPORT_FILE_NAME ${LAGOM_GENERATE_EXPORT_HEADER_EXPORT_FILE_NAME})
 endfunction()
 
 function(lagom_copy_runtime_dlls target_name)
```

</details>

Which allows us to do `lagom_generate_export_header(LibCoreMinimal coreminimal EXPORT_FILE_NAME "ExportCoreMinimal.h")`

R-Goc: Not sure if that makes sense, since this is the only place in the codebase where this happens that I know of. If a mainatainer prefers that I will change it.

ADKaster: CoreMinimal always felt like a CMake workaround rather than an actual library to me. We wanted a way to avoid circular dependencies for LibCore components that were generated by host tools.

These days most of those components have been replaced with third-party libs (icu, etc).

Are there any targets today that actually require the minimal LibCore rather than linking the whole shebang? If we really want these to be separate libs (which I doubt we do), they should be actually separate libs, rather than two similar libs living in the same directory.

Mostly, `COREMINIMAL_API` feels gross to put on classes, when the distinction between what is Core and what is CoreMinimal is arbitrary-ish.

cc @trflynn89

R-Goc: I'm looking if it would be possible to just make everything LibCore. The only real user of CoreMinimal is LibIDL, but even if a tool only build is used I think building core is fine.

R-Goc: Looks like it will work?

R-Goc:
Hmm. This causes the fuzzer build to fail as core depends on LibTextCodec having generated its export headers. It's easily solvable by moving LibTextCoded to Lagom\CMakeLists.txt but I'd rather move things away from there than to there. Does anyone have an idea on another way to do it?

R-Goc:
Moving things to Lagom\CMakeLists will work, but it requires moving TextCodec URL and Unicode, And LibURL depends on Regex and Crypto too. All of this will work, it's just adding more and more things to Lagom\CMakeLists.

PR #7013, opened 2025-12-03
Meta: Update libavif to version 1.3.0#1

socram8888:
To update to ICU 78.1 (https://github.com/LadybirdBrowser/ladybird/pull/6664), a bump to the Vcpkg baseline is required. However, for some reason 1.3.0#0 fails to compile on the newer baseline. Updating to 1.3.0#1 fixes the issue.


socram8888:
If this library is kept on 1.3.0#0 on the latest baseline, I get a "fatal: failed to unpack tree object b7a86219797184383b76c0efa3eeb54aaee060bd" error, the same this other user has on https://github.com/microsoft/vcpkg/issues/48209

I don't know the reason, but updating fixed it, and I guess it's better to run the latest version anyway.

PR #7014, opened 2025-12-04
LibWasm: Switch to wasm-tools for test gen and update to latest spectest

alimpfard:


AK/DoublyLinkedList.h
```diff
@@ -41,12 +41,11 @@ template<typename T, size_t node_cache_size>
...
     struct Node {
-        template<typename U>
-        explicit Node(U&& v)
+        template<typename... Args>
+        explicit Node(Args&&... args)
         {
-            new (m_value) T(forward<U>(v));
-            static_assert(
-                requires { T(v); }, "Conversion operator is missing.");
+            static_assert(requires { T(forward<Args>(args)...); });
```
Hendiadyoin1: Shouldnt that be a `requires`?

AK/Vector.h
```diff
@@ -860,8 +860,14 @@ requires(!IsRvalueReference<T>) class Vector {
...
-        for (size_t i = size(); i < new_size; ++i)
-            new (slot(i)) StorageType {};
+        if constexpr (Traits<StorageType>::is_trivial()) {
+            // For trivial types, we can just zero the new memory.
+            size_t old_size = size();
+            memset(slot(old_size), 0, (new_size - old_size) * sizeof(StorageType));
+        } else {
+            for (size_t i = size(); i < new_size; ++i)
+                new (slot(i)) StorageType {};
+        }
```
Hendiadyoin1: I thought we'd even elide the element destructors, at least that's what we do in some other containers
so this should also apply to `clear{,_with_capacity}`

Libraries/LibWasm/Parser/Parser.cpp
```diff
@@ -4,6 +4,8 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
+#include "AK/Enumerate.h"
```
Hendiadyoin1: clangd auto-include strikes again
(Films fix for that should be already merged, which decreases the chance of this happening)

Libraries/LibWasm/Parser/Parser.cpp
```diff
@@ -1474,9 +1476,32 @@ ParseResult<NonnullRefPtr<Module>> Module::parse(Stream& stream)
...
     }
 
+    module_ptr->preprocess();
+
     return module_ptr;
 }
 
+void Module::preprocess()
+{
+    for (auto const& [i, type_id] : enumerate(function_section().types())) {
```
Hendiadyoin1: I think you can elide the `&` here

Libraries/LibWasm/Printer/Printer.cpp
```diff
@@ -479,8 +479,15 @@ void Printer::print(Wasm::Instruction const& instruction)
     print_indent();
     print("({}", instruction_name(instruction.opcode()));
     if (instruction.arguments().has<u8>()) {
-        if (instruction.opcode() == Instructions::local_get || instruction.opcode() == Instructions::local_set || instruction.opcode() == Instructions::local_tee)
-            print(" (local index {})", instruction.local_index());
+        if (instruction.opcode() == Instructions::local_get
+            || instruction.opcode() == Instructions::local_set
+            || instruction.opcode() == Instructions::local_tee
+            || instruction.opcode() == Instructions::synthetic_argument_get) {
```
Hendiadyoin1: didnt we have a `first_is_any_of` helper

Libraries/LibWasm/Printer/Printer.cpp
```diff
@@ -536,8 +548,17 @@ void Printer::print(Wasm::Instruction const& instruction)
...
             [&](auto const& value) { print("(const {})", value); });
 
-        if (instruction.local_index().value())
-            print(" (local index {})", instruction.local_index().value());
+        if (instruction.opcode() == Instructions::local_get
+            || instruction.opcode() == Instructions::local_set
+            || instruction.opcode() == Instructions::local_tee
+            || instruction.opcode() == Instructions::synthetic_argument_get
+            || instruction.opcode() == Instructions::synthetic_local_seti32_const
+            || instruction.opcode() == Instructions::synthetic_i32_storelocal) {
```
Hendiadyoin1: as above, or at this point, maybe have a specific name for this group of opcodes, lest one forgets one somewhere

Libraries/LibWasm/AbstractMachine/AbstractMachine.cpp
```diff
@@ -565,7 +561,9 @@ Result AbstractMachine::invoke(Interpreter& interpreter, FunctionAddress address
     Configuration configuration { m_store };
     if (m_should_limit_instruction_count)
         configuration.enable_instruction_count_limit();
-    return configuration.call(interpreter, address, move(arguments));
+
+    Vector<Value, ArgumentsStaticSize> args = move(arguments);
+    return configuration.call(interpreter, address, args);
```
Hendiadyoin1: so this is now explicitly not reusing the allocation, and previously was, or am I missing something

Libraries/LibWasm/AbstractMachine/BytecodeInterpreter.cpp
```diff
@@ -403,660 +403,664 @@ HANDLE_INSTRUCTION(v128_store)
...
     TAILCALL return continue_(HANDLER_PARAMS(DECOMPOSE_PARAMS_NAME_ONLY));
 }
 
 HANDLE_INSTRUCTION(synthetic_i32_add2local)
 {
-    configuration.push_to_destination(Value(static_cast<i32>(Operators::Add {}(configuration.local_or_argument(instruction->local_index()).to<u32>(), configuration.local_or_argument(instruction->arguments().get<LocalIndex>()).to<u32>()))), addresses.destination);
+    configuration.push_to_destination<source_address_mix>(
+        Value(static_cast<i32>(Operators::Add {}(
+            configuration.local_or_argument(instruction->local_index()).to<u32>(),
+            configuration.local_or_argument(instruction->arguments().get<LocalIndex>()).template to<u32>()))),
```
Hendiadyoin1: Why does this call require an explicit `template` now?

Libraries/LibWasm/AbstractMachine/BytecodeInterpreter.cpp
```diff
@@ -105,12 +105,18 @@ constexpr static u32 default_sources_and_destination = (to_underlying(Dispatch::
 template<u64 opcode>
 struct InstructionHandler { };
 
+struct __attribute__((packed)) ShortenedIPAndAddresses {
```
Hendiadyoin1: you sure the `packed` is needed?

as it would be quite odd for a simple type to be sized like u32 and still have u64 like alignment

Hendiadyoin1: or does that affect the calling convention
(mainly making sure as packing can be a footgun when accidentally handing out references)

alimpfard: GCC was splitting the struct up and undoing the fix without it, I got annoyed and just slapped packed there to make it stop

Libraries/LibWasm/AbstractMachine/BytecodeInterpreter.cpp
alimpfard: It probably doesn't, that's gone through like 4 iterations and gathered unnecessary keywords. I'll clean them up.

Libraries/LibWasm/AbstractMachine/AbstractMachine.cpp
alimpfard: Can't pass a reference with different inline capacities, I think the move ctor just yoinks the allocation if its big enough anyway.

Libraries/LibWasm/Parser/Parser.cpp
alimpfard: Damn it I thought I'd gone through them all

AK/Vector.h
alimpfard: That'd fix one jank thing I had to do, so sure, will do.

AK/DoublyLinkedList.h
alimpfard: I'm not sure how that'd work? `args` might be any sort of ref/temporary that we wouldn't know until inside the block.

Hendiadyoin1: Shouldnt it work like in the `empend` functions?
```c++
template<typename... Args>
requires(IsContructible<T, Args...>
explicit Node(Args&&... args)
{
    new (m_value) T (forward<Args>(args)...);
}
```
As the `Args` entries should be the fully qualified types passed in

alimpfard: ah sure, will make it do that instead.

alimpfard:
Hmm, looks like this regressed a bunch of benchmarks, drafting until I figure that out.
```
WasmCoremark    coremark-minimal.wasm       0.797  695.448 ± 692.259 … 698.147  554.250 ± 553.738 … 555.219
WasmMicroBench  call-00-args.wasm           0.809  2.724 ± 2.722 … 2.726        3.367 ± 3.311 … 3.432
WasmMicroBench  call-01-args.wasm           1.065  6.175 ± 6.138 … 6.204        5.797 ± 5.548 … 6.187
WasmMicroBench  call-02-args.wasm           1.081  6.292 ± 6.254 … 6.365        5.819 ± 5.783 … 5.847
WasmMicroBench  call-03-args.wasm           1.054  6.487 ± 6.464 … 6.515        6.157 ± 6.083 … 6.281
WasmMicroBench  call-04-args.wasm           1.046  6.822 ± 6.776 … 6.865        6.524 ± 6.425 … 6.628
WasmMicroBench  call-16-args.wasm           1.069  11.907 ± 11.831 … 12.044     11.141 ± 11.058 … 11.305
WasmMicroBench  call-32-args.wasm           0.94   16.817 ± 16.776 … 16.895     17.882 ± 17.866 … 17.905
WasmRustBench   base64-bench.wasm           0.783  17.661 ± 17.553 … 17.826     22.558 ± 22.488 … 22.671
WasmRustBench   regex-match-bench.wasm      0.778  3.620 ± 3.594 … 3.660        4.653 ± 4.634 … 4.664
WasmRustBench   sha512-bench.wasm           0.498  24.170 ± 24.066 … 24.360     48.580 ± 48.376 … 48.695
WasmCoremark    Total                       0.797  695.448                      554.250
WasmMicroBench  Total                       1.009  57.222                       56.687
WasmRustBench   Total                       0.6    45.451                       75.790
All Suites      Total                       0.778  120.115                      154.354
```

AK/DoublyLinkedList.h
```diff
@@ -115,12 +114,12 @@ class DoublyLinkedList {
...
 
-    template<typename U>
-    ErrorOr<void> try_append(U&& value)
+    template<typename... Args>
+    ErrorOr<void> try_append(Args&&... args)
     {
         static_assert(
-            requires { T(value); }, "Conversion operator is missing.");
-        auto* node = make_node(forward<U>(value));
+            requires { T(forward<Args>(args)...); }, "Initializer is missing.");
```
Hendiadyoin1: Same here I guess

Hendiadyoin1: also technically an `empend` now
and kinda weird that the non-try append doesnt have the check relying on the one here

Libraries/LibWasm/AbstractMachine/BytecodeInterpreter.cpp
Hendiadyoin1: a union would likely also work, but that probably not really nicer

AK/DoublyLinkedList.h
alimpfard: that's...true, will "fix".

alimpfard:
I think I'll make this PR only about the test generation and fixes, the performance stuff will have to stay in the oven a bit more.

PR #7017, opened 2025-12-04
Meta: Lock libxml2 version in vcpkg.json

kalenikaliaksandr:


vcpkg.json
```diff
@@ -286,6 +286,10 @@
       "name": "openssl",
       "version": "3.5.3#0"
     },
+    {
+      "name": "libxml2",
```
trflynn89: alphabetical pls

vcpkg.json
```diff
@@ -274,6 +274,10 @@
       "name": "libavif",
       "version": "1.3.0#1"
     },
+    {
+      "name": "libxml2",
```
trflynn89: x comes after w :sweat_smile:

PR #7027, opened 2025-12-05
LibGfx: Enable explicit symbol export

R-Goc:
This commit enables explicit symbol export for LibGfx. This decreases the number of exported symbols from around ~1500 to ~400.


AK/LsanSuppressions.h
```diff
@@ -8,9 +8,15 @@
...
+#ifndef AK_OS_WINDOWS
+#    define EXPORT __attribute__((visibility("default")))
+#else
+#    define EXPORT __declspec(dllexport)
+#endif
+
 #ifdef HAS_ADDRESS_SANITIZER
 extern "C" {
-char const* __lsan_default_suppressions();
+EXPORT char const* __lsan_default_suppressions();
```
ADKaster, line 19: this should be its own commit

Libraries/LibGfx/ImageFormats/TinyVGLoader.h
```diff
@@ -7,15 +7,9 @@
...
-#include <AK/OwnPtr.h>
-#include <AK/Vector.h>
-#include <LibGfx/Color.h>
+#include <LibGfx/Export.h>
 #include <LibGfx/Forward.h>
 #include <LibGfx/ImageFormats/ImageDecoder.h>
-#include <LibGfx/PaintStyle.h>
-#include <LibGfx/Painter.h>
-#include <LibGfx/Path.h>
-#include <LibGfx/VectorGraphic.h>
```
ADKaster, line 18: this header cleanup seems separate from the EXPORT change.

Libraries/LibGfx/Color.h
```diff
@@ -778,9 +779,9 @@ struct Formatter<Gfx::Oklab> : public Formatter<FormatString> {
 namespace IPC {
 
 template<>
-ErrorOr<void> encode(Encoder&, Gfx::Color const&);
+GFX_API ErrorOr<void> encode(Encoder&, Gfx::Color const&);
 
 template<>
-ErrorOr<Gfx::Color> decode(Decoder&);
+GFX_API ErrorOr<Gfx::Color> decode(Decoder&);
```
ADKaster, line 785: just to confirm, these impls do live in LibGfx right? Some common classes have their impls in LibIPC.

Libraries/LibGfx/Palette.h
```diff
@@ -12,11 +12,12 @@
 #include <AK/Noncopyable.h>
 #include <AK/RefCounted.h>
 #include <AK/RefPtr.h>
+#include <LibGfx/Export.h>
 #include <LibGfx/SystemTheme.h>
 
 namespace Gfx {
 
-class PaletteImpl : public RefCounted<PaletteImpl> {
+class GFX_API PaletteImpl : public RefCounted<PaletteImpl> {
```
ADKaster: should this class really be exported? it's got impl in the name

Libraries/LibGfx/Path.h
```diff
@@ -9,14 +9,15 @@
...
+#include <LibGfx/Export.h>
 #include <LibGfx/Forward.h>
 #include <LibGfx/Point.h>
 #include <LibGfx/Rect.h>
 #include <LibGfx/WindingRule.h>
 
 namespace Gfx {
 
-class PathImpl {
+class GFX_API PathImpl {
```
ADKaster, line 21: same here, does Path*Impl* need exported?

R-Goc: With all of these I was fixing link errors one by one. For some I did add a TODO: some not. But yes, I find it rather ugly that the impl needs to be exported.

Libraries/LibGfx/ImageFormats/TinyVGLoader.h
R-Goc: I did it as a drive by to match the other ones. I can remove it, but it doesn't hurt.

Libraries/LibGfx/Color.h
R-Goc: They're implemented in color.cpp.

Libraries/LibGfx/Path.h
R-Goc: I'll retry the Impl ones to make sure.

Libraries/LibGfx/Palette.h
R-Goc: PaletteImpl is directly used in WebContent

Libraries/LibGfx/Path.h
R-Goc: I see why now. Taking a look at Path it looks like it should be fine with PathImpl not being exported. But then:
```
lld-link: error: undefined symbol: public: static class AK::NonnullOwnPtr<class Gfx::PathImpl> __cdecl Gfx::PathImpl::create(void)
>>> referenced by E:\lib\ladybird\Libraries\LibWeb\CSS\StyleValues\BasicShapeStyleValue.cpp:20
>>>               Lagom\Libraries\LibWeb\CMakeFiles\LibWeb.dir\CSS\StyleValues\BasicShapeStyleValue.cpp.obj:(class Gfx::Path __cdecl Web::CSS::path_from_resolved_rect(float, float, float, float))
...
```
Which is just a normal use of Path:
```cpp
static Gfx::Path path_from_resolved_rect(float top, float right, float bottom, float left)
{
    Gfx::Path path;
    path.move_to(Gfx::FloatPoint { left, top });
    path.line_to(Gfx::FloatPoint { right, top });
    path.line_to(Gfx::FloatPoint { right, bottom });
    path.line_to(Gfx::FloatPoint { left, bottom });
    path.close();
    return path;
}
```
But the issue is that     
```cpp
static NonnullOwnPtr<Gfx::PathImpl> create();
```
is defined on PathImpl which means that whatever wants to use Path which has a NonnullOwnPtr to the Impl needs this method to be exported. This looks like a rather common pattern. I'll link this in browser as it seems like a rather architectural decision.

ADKaster: thanks for looking into that. the fact that the symbol needs exported feels weird, but if it doesn't link it doesn't link.

Libraries/LibGfx/ImageFormats/TinyVGLoader.h
ADKaster: Removing extra headers is good, but similar to the AK change it should be broken out.

PR #7030, opened 2025-12-05
LibWeb: Store animation time values in abstract type

Calme1709:
In level 2 of the Web Animations spec, all time values associated with animations are no longer simple numbers but are instead either percentages or numbers for progress- and time-based timelines respectively. These values are represented internally by a new `TimeValue` wrapper struct, and externally (i.e. in JS) as `CSSNumberish`.

This is a another step towards supporting progress (i.e. scroll) based animations.

### DIFF
--- a/Libraries/LibWeb/Animations/AnimationTimeline.h
+++ b/Libraries/LibWeb/Animations/AnimationTimeline.h
@@ -17,9 +17,9 @@ class AnimationTimeline : public Bindings::PlatformObject {
     GC_DECLARE_ALLOCATOR(AnimationTimeline);
 
 public:
-    Optional<double> current_time_for_bindings() const
+    NullableCSSNumberish current_time_for_bindings() const
     {
-        return current_time().map([](auto const& current_time) { return current_time.as_milliseconds(); });
+        return NullableCSSNumberish::from_optional_css_numberish_time(current_time());
     }
     Optional<TimeValue> current_time() const;
 
--- a/Libraries/LibWeb/Animations/AnimationTimeline.idl
+++ b/Libraries/LibWeb/Animations/AnimationTimeline.idl
@@ -3,7 +3,7 @@
 // https://drafts.csswg.org/web-animations-2/#the-animationtimeline-interface
 [Exposed=Window]
 interface AnimationTimeline {
-    [ImplementedAs=current_time_for_bindings] readonly attribute double? currentTime;
+    [ImplementedAs=current_time_for_bindings] readonly attribute CSSNumberish? currentTime;
     [ImplementedAs=duration_for_bindings] readonly attribute CSSNumberish? duration;
     [FIXME] Animation play (optional AnimationEffect? effect = null);
 };
--- a/Libraries/LibWeb/Animations/TimeValue.h
+++ b/Libraries/LibWeb/Animations/TimeValue.h
@@ -77,13 +77,6 @@ struct TimeValue {
         return type == other.type && value == other.value;
     }
 
-    // FIXME: This method is temporary as we migrate all CSS timing to use TimeValue.
-    double as_milliseconds() const
-    {
-        VERIFY(type == Type::Milliseconds);
-        return value;
-    }
-
     CSS::CSSNumberish as_css_numberish() const
     {
         switch (type) {


Calme1709:
Changes: Fixed two GCC compilation issues

Libraries/LibWeb/Animations/Animation.cpp
```diff
@@ -134,15 +134,16 @@ void Animation::set_timeline(GC::Ptr<AnimationTimeline> new_timeline)
 
 // https://www.w3.org/TR/web-animations-1/#dom-animation-starttime
 // https://www.w3.org/TR/web-animations-1/#set-the-start-time
-void Animation::set_start_time(Optional<double> const& new_start_time)
+void Animation::set_start_time_for_bindings(Optional<double> const& new_start_time)
 {
     // Setting this attribute updates the start time using the procedure to set the start time of this object to the new
     // value.
+    auto valid_start_time = new_start_time.map([](auto const& value) { return TimeValue { TimeValue::Type::Milliseconds, value }; });
```
AtkinsSJ: IMO, it's a bit easier to understand if this is `new_start_time`, and the parameter gets a different name. It makes the spec comments and code line up better.

...after looking at later commits, I'm not so sure what the best thing to do is. Thought I should leave this here anyway.

Libraries/LibWeb/Animations/Animation.cpp
```diff
@@ -212,16 +213,18 @@ Optional<double> Animation::current_time() const
 }
 
 // https://www.w3.org/TR/web-animations-1/#animation-set-the-current-time
-WebIDL::ExceptionOr<void> Animation::set_current_time(Optional<double> const& seek_time)
+WebIDL::ExceptionOr<void> Animation::set_current_time_for_bindings(Optional<double> const& seek_time)
 {
+    auto valid_seek_time = seek_time.map([](auto const& value) { return TimeValue { TimeValue::Type::Milliseconds, value }; });
```
AtkinsSJ: Same thing here.

Libraries/LibWeb/Animations/Animation.cpp
```diff
@@ -132,33 +133,84 @@ void Animation::set_timeline(GC::Ptr<AnimationTimeline> new_timeline)
...
+    if (
+        // FIXME: The animation is not associated with a progress-based timeline, and
+
+        // time is a CSSNumericValue, and
+        time.has_value() && time->has<GC::Root<CSS::CSSNumericValue>>() &&
+
+        // the units of time are not duration units:
+        !time->get<GC::Root<CSS::CSSNumericValue>>()->type().matches_time({}) &&
+
+        // AD-HOC: While it's not mentioned in the spec WPT also expects us to support number CSSNumericValues here
```
AtkinsSJ: Is there a spec issue for this?

Libraries/LibWeb/Animations/Animation.cpp
```diff
@@ -132,33 +133,84 @@ void Animation::set_timeline(GC::Ptr<AnimationTimeline> new_timeline)
...
+    // 6. Let previous current time be animation’s current time.
     auto previous_current_time = current_time();
 
-    // 4. Apply any pending playback rate on animation.
+    // 7. Apply any pending playback rate on animation.
     apply_any_pending_playback_rate();
 
-    // 5. Set animation’s start time to new start time.
+    // 8. Set animation’s start time to new start time.
     m_start_time = valid_start_time;
```
AtkinsSJ: It's not clear to me if this is correct now. Maybe a spec bug?

Libraries/LibWeb/Animations/Animation.cpp
Calme1709: There wasn't - have opened [one](https://github.com/w3c/csswg-drafts/issues/13196) and added the link into this comment.

Libraries/LibWeb/Animations/Animation.cpp
Calme1709: I'm not sure what you are referring to here, all seems correct to me (possibly confusion around the naming of the `valid_start_time` variable which is resolved in your other comment)

Libraries/LibWeb/Animations/Animation.cpp
Calme1709: I have updated this to be `new_start_time` and the parameter name to be `raw_new_start_time` to align us with the spec name later on.

Libraries/LibWeb/Animations/Animation.cpp
Calme1709: Updated to `seek_time` and `raw_seek_time` as above

Libraries/LibWeb/Animations/Animation.cpp
AtkinsSJ: It gets a bit confusing because when you switch to the updated spec text, it now has both `new start time` and `valid start time` as variables. Step 1 defines `valid start time` as different from `new start time`, and then step 8 here uses `new start time`. My question is which one should actually be used in step 8.

It'd also mean renaming the variables back in this commit. :sweat_smile:

Calme1709: `valid start time` in the spec is just a boolean which is whether `new start time` is valid, the `valid_start_time` variable I introduced was ad-hoc and has been renamed (see other comment).

We don't actually need the `valid start time` variable in our implementation since we handle the invalid case by throwing which is handled by the `TRY()` macro (perhaps we need an AD-HOC comment here?).

Calme1709: I have added an `AD-HOC` comment in here to make this clearer

AtkinsSJ (approved):

PR #7033, opened 2025-12-05
Fix: Make Scrollbar Thickness Consistent Across DPI

### LINKED ISSUE #6920: Zooming the page changes scrollbar sizes
Zooming in or out causes the scrollbars to grow or shrink along with the page content. This makes them difficult to see or use at small zoom levels:

Image

...and hilariously chonky when zoomed in:

Image

I haven't tested pinch-zoom, so I'm not sure if that works differently.

askadityapandey:
This PR updates scrollbar painting to use a fixed device-pixel
thickness for the gutter and thumb so the UI looks consistent
across different zoom levels and DPI scales.

## Key changes
- Introduces fixed device-pixel thickness for scrollbar chrome.
- Converts that thickness to CSS pixels using `device_pixels_per_css_pixel()`.
- Adjusts gutter/thumb rects so right/bottom edges remain anchored.
- Only affects `PaintPhase::Overlay` and elements whose scrollbar is not `none`.

## Notes
- Visual-only sizing change; scrolling behaviour is unchanged.
- If wanted, this can be exposed via CSS `scrollbar-width` later.

Fixes #6920

### DIFF
--- a/Libraries/LibWeb/Painting/PaintableBox.cpp
+++ b/Libraries/LibWeb/Painting/PaintableBox.cpp
@@ -508,19 +508,49 @@ void PaintableBox::paint(DisplayListRecordingContext& context, PaintPhase phase)
         }
     }
 
-    if (phase == PaintPhase::Overlay && (g_paint_viewport_scrollbars || !is_viewport_paintable()) && computed_values().scrollbar_width() != CSS::ScrollbarWidth::None) {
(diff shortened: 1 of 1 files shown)


konradekk:
(note: there's `scrollbar-width` checked above, it can contain `thin` value. looks like a short stretch to add that here, while you are at this.)

PR #7039, opened 2025-12-05
LibCore+LibThreading: Add a thread-safe weak link for EventLoop

Zaggy1024:
Since the event loop has a very specifically scoped lifetime, we can't ensure that it outlives threads that hold a reference to it without blocking the thread that owns it. In order to make threads use the event loop safely, we now have an atomically ref-counted EventLoopWeak class that can be passed off to threads to safely post events/callbacks to it.

Another possibility was to use an RWLock per event loop that each thread holds a read lock on, while ~EventLoop() uses a write lock to block and prevent it being destroyed until all its threads exit. However, media data providers don't receive a signal to exit due to the GC heap being intentionally leaked, so the process never actually exits. It would be possible to specifically drop the reference to PlaybackManager in HTMLMediaElement in order to make those data providers die on their own, but that doesn't help prevent this problem in other cases where it may arise.

### DIFF
--- a/Libraries/LibThreading/BackgroundAction.h
+++ b/Libraries/LibThreading/BackgroundAction.h
@@ -76,7 +76,7 @@ private:
         if (on_error.has_value())
             m_on_error = on_error.release_value();
 
-        enqueue_work([self = NonnullRefPtr(*this), promise = move(promise), origin_event_loop = &Core::EventLoop::current()]() mutable {
+        enqueue_work([self = NonnullRefPtr(*this), promise = move(promise), origin_event_loop = Core::EventLoop::current_weak()]() mutable {
             auto result = self->m_action(*self);
 
             // The event loop cancels the promise when it exits.
@@ -88,7 +88,7 @@ private:
 
                 // If there is no completion callback, we don't rely on the user keeping around the event loop.
                 if (self->m_on_complete) {
-                    origin_event_loop->deferred_invoke([self, promise = move(promise)] {
+                    origin_event_loop->take()->deferred_invoke([self, promise = move(promise)] {
                         // Our promise's resolution function will never error.
                         (void)promise->resolve(*self);
                     });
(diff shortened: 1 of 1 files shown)


Zaggy1024:
Looks like the changes to BackgroundThread are causing a rare deadlock (or more likely infinite spin) in some test webviews, I'll have to do some further investigation...

Libraries/LibCore/EventLoop.cpp
```diff
@@ -152,4 +162,54 @@ void deferred_invoke(Function<void()> invokee)
...
 }
 
+EventLoopWeak::EventLoopWeak(EventLoop& event_loop)
+    : m_event_loop(&event_loop)
+{
+}
+
+void EventLoopWeak::revoke()
+{
+    m_lock.lock_write();
```
R-Goc: use RWLockLocker instead:
```cpp
Threading::RWLockLocker<Threading::LockMode::Write> lock(m_lock);
```

Libraries/LibCore/EventLoop.h
```diff
@@ -89,13 +92,50 @@ class EventLoop {
...
+    static NonnullRefPtr<EventLoopWeak> current_weak();
 
     EventLoopImplementation& impl() { return *m_impl; }
 
 private:
     NonnullOwnPtr<EventLoopImplementation> m_impl;
+    RefPtr<EventLoopWeak> m_weak;
 } SWIFT_UNSAFE_REFERENCE;
 
+class EventLoopLockedReference;
```
R-Goc: nit: The name didn't make it immediately obvious to me if this was a shared or exclusive reference and I had to read the source for it. Since the name is just an implemenation detail maybe EventLoopLockedSharedReference? Though it is quite long.

Libraries/LibCore/EventLoop.cpp
Zaggy1024: I thought about doing this, but that's quite verbose and has no benefit here due to no branching.

Libraries/LibCore/EventLoop.h
Zaggy1024: Yeah, the naming is something I kind of struggled with, but I don't think it matters whether it's shared or exclusive? The event loop should allow posting events concurrently without issue anyway.

I would welcome any more concise names, but I definitely don't want to make the name too verbose.

I could potentially just call it `EventLoopStrongReference`, though that kind of makes it sound like it's holding a reference count instead of a lock.

Libraries/LibCore/EventLoop.cpp
Zaggy1024: Ended up changing this for consistency, we can make the type name shorter later.

PR #7040, opened 2025-12-05
LibWeb: Fix Document::is_fully_active() and implement Badging API

estefysc:
Fixes WPT test: badging/non-fully-active.https.html

Fixes Document::is_fully_active() to check the has_been_destroyed flag. When a navigable is destroyed, the flag is set immediately but the navigable pointer is cleared later in a queued task (see NavigableContainer::destroy_the_child_navigable), causing is_fully_active() to incorrectly return true during this gap.

Implements navigator.setAppBadge() and navigator.clearAppBadge() methods per the specification. Permission handling, badge display logic, and task queuing are not yet implemented.

### DIFF
--- a/Libraries/LibWeb/CMakeLists.txt
+++ b/Libraries/LibWeb/CMakeLists.txt
@@ -592,6 +592,7 @@ set(SOURCES
     HTML/NavigationParams.cpp
     HTML/NavigationTransition.cpp
     HTML/Navigator.cpp
+    HTML/NavigatorBadge.cpp
     HTML/NavigatorBeacon.cpp
     HTML/NavigatorConcurrentHardware.cpp
     HTML/NavigatorDeviceMemory.cpp
--- a/Libraries/LibWeb/HTML/Navigator.cpp
+++ b/Libraries/LibWeb/HTML/Navigator.cpp
@@ -37,6 +37,11 @@ Navigator::Navigator(JS::Realm& realm)
 
 Navigator::~Navigator() = default;
 
+HTML::Window& Navigator::window()
+{
+    return as<HTML::Window>(HTML::current_principal_global_object());
+}
+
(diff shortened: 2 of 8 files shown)


estefysc:
I need some guidance from maintainers, please. I'm implementing the Badging API (https://w3c.github.io/badging/) to fix the failing WPT test badging/non-fully-active.https.html, and ran into an architectural question about navigable/document destruction.

The HTML spec defines a document as fully active when "d is the active document of a navigable navigable, and either navigable is a top-level traversable or navigable's container document is fully active" (https://html.spec.whatwg.org/multipage/document-sequences.html#fully-active). When implementing the Badging API, I interpreted this to mean a document should not be fully active if its navigable is being destroyed (even though the pointer isn't nulled yet). I added a check to return false when navigable->has_been_destroyed() is true (Document.cpp:3288). This makes the badging WPT test pass locally, but encoding/streams/invalid-realm.window.html times out.

The root cause is that there are two separate "has_been_destroyed" flags with different timing: Navigable::has_been_destroyed is set immediately in NavigableContainer.cpp:296, while Document::has_been_destroyed is set later in Document.cpp:4126 during destroy(). The comment at NavigableContainer.cpp:289 explains that the navigable flag "is used instead to check whether navigable is already destroyed" because setting the content navigable to null is deferred. My fix to is_fully_active() checks the navigable flag (early), but can_run_script() (Environments.cpp:133) and Task::is_runnable() (Task.cpp:62) check the document flag (late). This creates a timing gap where is_fully_active() = false but has_been_destroyed() = false, causing scripts to be blocked and stream operations to hang.

Should can_run_script() and Task::is_runnable() also check the navigable's flag for consistency? Or do these flags intentionally represent different destruction phases? The navigable flag was added for https://github.com/whatwg/html/issues/9690  
Should document-related checks use the same flag?

PR #7042, opened 2025-12-06
Implement decapsulate methods CryptoSubtle

tete17:
I have also gone ahead and implemented the [crypto task source](https://www.w3.org/TR/webcrypto-2/#dfn-crypto-task-source-0).

This is something the previous methods also implement in the most up to date version of the standard but our implementation predates https://github.com/w3c/webcrypto/pull/386 so it was never in use.


shannonbooth:
The code itself looks alright to me, but are there any associated WPT which can be imported / this makes an improvement on? I did see some tentative underneath encap_decap_bits.tentative.https.any.js for example

tete17:
> The code itself looks alright to me, but are there any associated WPT which can be imported / this makes an improvement on? I did see some tentative underneath encap_decap_bits.tentative.https.any.js for example

All of the wpt require to also implement the decap for at least one encryption mechanism. I can extend this PR to the point where at least one WPT passes but it may mean adding quiet a bit more lines of code i wanted to reduce the size of the PR i was making.

If you want me to increase the size just give me a 👍

shannonbooth:
Fair enough!

edit: Sorry! It needed a rebase for the merge which I accidentally clicked a button on in the GitHub UI which added the merge commit. Fixed now

PR #7043, opened 2025-12-06
LibWeb: Honor LegacyNullToEmptyString in union string conversion

mikiubo:
The IDL generator was passing 'false' to generate_to_cpp when converting a union containing a string type.
This broke WebIDL's LegacyNullToEmptyString behavior.

Pass the actual legacy_null_to_empty_string flag instead. Fixes some WPTs:

domparsing/outerhtml-02.html
trusted-types/block-string-assignment-to-Element-outerHTML trusted-types/block-string-assignment-to-HTMLElement-generic trusted-types/block-string-assignment-to-ShadowRoot-innerHTML


tete17:
Nice!

AtkinsSJ (approved):
🎉

Issue #7038, opened 2025-12-05
Crash when loading LifeViewer Pro


muzikbike:
### Summary

Attempting to go to https://lazyslug.com/lifeviewer/pro/ will cause Ladybird to crash before loading of page content concludes.

### Operating system

Linux

### Steps to reproduce

1. Open Ladybird
2. Navigate to https://lazyslug.com/lifeviewer/pro/
3. Wait

### Expected behavior

LifeViewer Pro would load as expected.

### Actual behavior

We instead crash before it appears.

### URL for a reduced test case

https://lazyslug.com/lifeviewer/pro/

### HTML/SVG/etc. source for a reduced test case

```html
N/A
```

### Log output and (if possible) backtrace

```shell
./Meta/ladybird.py run
ninja: Entering directory `/home/conputer/other-dev/ladybird/20250807210822/ladybird/Build/release'
ninja: no work to do.
Ladybird PID file '/run/user/1000/Ladybird.pid' exists with PID 14824, but process cannot be found
5283.097 WebContent(15027): (js log) "WebAssembly instantiated: 102 functions (161857 bytes), 1792Mb heap, time 636.8ms"
5283.170 WebContent(15027): (js log) "refresh rate" "165Hz Auto"
5283.301 WebContent(15027): (js) "read embedded: 0.105 seconds"
5283.301 WebContent(15027): (js) "page scan: 0.131 seconds"
5283.301 WebContent(15027): (js log) "1 embedded and 0 popup (0 unsafe)"
5284.510 Ladybird(15004): WebContent process crashed! Last page loaded: https://lazyslug.com/lifeviewer/pro/
5284.510 Ladybird(15004): Consider raising an issue at https://github.com/LadybirdBrowser/ladybird/issues/new/choose
5284.511 WebContent(15044): FIXME: Handle null navigation_request in navigation response Content Security Policy check.
```

### Screenshots or screen recordings

<img width="939" height="909" alt="Image" src="https://github.com/user-attachments/assets/c50a8c19-682e-4c72-ac94-c3d06e12a815" />

### Contribute a patch?

- [ ] I’ll contribute a patch for this myself.

rcorsi:
Might be same as #7009 issue. I just tried the site with GNU compile Ladybird and I got a crash.

Did you compile with GNU or Clang?  If you are just using `./Meta/Ladybird build` it will use the default compiler.

What does `g++ --version` give you?

[...]

Issue #7032, opened 2025-12-05
Crash when loading primevalgame.com


muzikbike:
### Summary

Visiting https://primevalgame.com/ will cause Ladybird to crash before the content can load in its entirety.

### Operating system

Linux

### Steps to reproduce

1. Visit this site
2. Wait for it to load
3. Observe a crash notice appears instead

### Expected behavior

The game would load into at least some sort of state.

### Actual behavior

Give it about 30 seconds and it crashes.

### URL for a reduced test case

https://primevalgame.com/

### HTML/SVG/etc. source for a reduced test case

```html
N/A
```

### Log output and (if possible) backtrace

```shell
./Meta/ladybird.py run
ninja: Entering directory `/home/conputer/other-dev/ladybird/20250807210822/ladybird/Build/release'
ninja: no work to do.
4262.403 ImageDecoder(22280:22316): libpng warning: eXIf: duplicate
4262.404 WebContent(22297): (js log) "[UnityCache] Error when initializing cache: NotSupportedError: CacheStorage.open() is not yet implemented"
4262.404 WebContent(22297): (js log) "[UnityCache] Failed to load 'https://primevalgame.com/Build/24_10_2025%20POD%20Disk%20Size%20FINAL%20Build.data' from indexedDB cache due to the error: NotSupportedError: CacheStorage.open() is not yet implemented"
4262.456 WebContent(22297): (js warn) "navigator.mediaDevices not supported by this browser. Webcam access will not be available."
4271.448 WebContent(22297): (js log) "[UnityMemory] Configuration Parameters - Can be set up in boot.config\n"
4271.448 WebContent(22297): (js log) "    "memorysetup-allocator-temp-initial-block-size-main=262144"\n"
4271.448 WebContent(22297): (js log) "    "memorysetup-allocator-temp-initial-block-size-worker=262144"\n"
4271.448 WebContent(22297): (js log) "    "memorysetup-temp-allocator-size-nav-mesh-worker=65536"\n"
4271.448 WebContent(22297): (js log) "    "memorysetup-temp-allocator-size-audio-worker=65536"\n"
4271.448 WebContent(22297): (js log) "    "memorysetup-bucket-allocator-granularity=16"\n"
4271.448 WebContent(22297): (js log) "    "memorysetup-bucket-allocator-bucket-count=8"\n"
4271.448 WebContent(22297): (js log) "    "memorysetup-bucket-allocator-block-size=4194304"\n"
[...]

Issue #7009, opened 2025-12-03
copy.sh/v86 crashes Ladybird


xdevs23:
### Summary

When I try to launch SerenityOS in v86 in Ladybird via https://copy.sh/v86/?profile=serenity, Ladybird crashes.

Source code: https://github.com/copy/v86

### Operating system

Linux 6.17.8 #1-NixOS SMP PREEMPT_DYNAMIC Thu Nov 13 20:37:49 UTC 2025

```
ANSI_COLOR="0;38;2;126;186;228"
BUG_REPORT_URL="https://github.com/NixOS/nixpkgs/issues"
BUILD_ID="25.05.20251124.1c8ba8d"
CPE_NAME="cpe:/o:nixos:nixos:25.05"
DEFAULT_HOSTNAME=nixos
DOCUMENTATION_URL="https://nixos.org/learn.html"
HOME_URL="https://nixos.org/"
ID=nixos
ID_LIKE=""
IMAGE_ID=""
IMAGE_VERSION=""
LOGO="nix-snowflake"
NAME=NixOS
PRETTY_NAME="NixOS 25.05 (Warbler)"
SUPPORT_END="2025-12-31"
SUPPORT_URL="https://nixos.org/community.html"
VARIANT=""
VARIANT_ID=""
VENDOR_NAME=NixOS
VENDOR_URL="https://nixos.org/"
VERSION="25.05 (Warbler)"
VERSION_CODENAME=warbler
VERSION_ID="25.05"
```

### Steps to reproduce

1. Go to https://copy.sh/v86/?profile=serenity
2. Wait for download to finish
3. After a few seconds, it crashes with "Ladybird flew off-course!"

### Expected behavior

The VM would start and run

### Actual behavior

The website crashes before the VM is visible.

### URL for a reduced test case

https://copy.sh/v86/?profile=serenity

### HTML/SVG/etc. source for a reduced test case

```html
N/A
```

### Log output and (if possible) backtrace

```shell
2532.763 Ladybird(73580): WebContent process crashed! Last page loaded: https://copy.sh/v86/?profile=serenity
2532.763 Ladybird(73580): Consider raising an issue at https://github.com/LadybirdBrowser/ladybird/issues/new/choose
2532.767 WebContent(74770): FIXME: Handle null navigation_request in navigation response Content Security Policy check.
```

### Build flags or config settings

Launched using:

```
nix run github:NixOS/nixpkgs/master#ladybird
```

[...]

PR #7044, opened 2025-12-07
Documentation: State wasm requirements in build instructions

rcorsi:
Related to #7009, #7025, #7032, #7038

Following discord question was put in the wasm channel

-------------

Got a few reports of WebContent crashing (#7038, #7032, #7025 #7009) and bottom line is that GNU 14 is being used and it lacks the musttail attribute support, that is only introduced in GNU 15. Without musttail large wasm files will create a huge stack size (I've seen >200K frames).  These large wasm files are usually complex games or OSes.

What should be done in this case of lacking musttail and large wasm? Just document it in installation instructions that GNU 14 and wasm don't play well together or after we know the size of the wasm file not execute it to avoid the crash and display a message saying "wasm too big, won't execute".

This issue has most of the information in

https://github.com/LadybirdBrowser/ladybird/issues/7009

What should be done to deal with this?

### DIFF
--- a/Documentation/BuildInstructionsLadybird.md
+++ b/Documentation/BuildInstructionsLadybird.md
@@ -7,6 +7,9 @@ Qt6 development packages, nasm, additional build tools, and a C++23 capable comp
 We currently use gcc-14 and clang-20 in our CI pipeline. If these versions are not available on your system, see
 [`Meta/find_compiler.py`](../Meta/find_compiler.py) for the minimum compatible version.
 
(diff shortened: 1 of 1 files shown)


rcorsi:
I prepared an example of how this could be documented, but would it be better to do something at runtime, in case user did not read the docs?

rcorsi:
Looks like #7046 will handle this issue better. Closing this down as not really needed.

PR #7047, opened 2025-12-07
LibWeb/HTML: Implement the Origin interface

shannonbooth:
See: https://github.com/whatwg/html/commit/68909b2

Also fixing a boog with registrable domain.

This also adds an `as_if` helper method for JS::Value to try and make the pattern of needing to check if a Value is an object, if that object is of a specific type, and then assigning it to a variable easier to do. Not sure if its the best API but it seems to feel okay, even if only limited to JS::Objects for now 🤔


Libraries/LibWeb/DOMURL/DOMURL.cpp
```diff
@@ -482,4 +482,10 @@ Optional<URL::URL> parse(StringView input, Optional<URL::URL const&> base_url, O
     return url.release_value();
 }
 
+Optional<URL::Origin> DOMURL::extract_an_origin() const
```
trflynn89, line 486: Spec link pls

Libraries/LibWeb/HTML/WindowOrWorkerGlobalScope.cpp
```diff
@@ -1277,4 +1277,15 @@ GC::Ref<TrustedTypes::TrustedTypePolicyFactory> WindowOrWorkerGlobalScopeMixin::
     return *m_trusted_type_policy_factory;
 }
 
+// https://html.spec.whatwg.org/multipage/webappapis.html#windoworworkerglobalscope-mixin:extract-an-origin
+Optional<URL::Origin> WindowOrWorkerGlobalScopeMixin::window_or_worker_global_scope_extract_an_origin() const
+{
+    // 1. If this's relevant settings object's origin is not same origin-domain with the entry settings object's origin, then return null.
+    if (!relevant_settings_object(this_impl()).origin().is_same_origin_domain(entry_settings_object().origin()))
```
trflynn89: This is a bit wordy, and reused - let's cache this origin above?
```c++
relevant_settings_object(this_impl()).origin()
```

Libraries/LibWeb/DOMURL/DOMURL.cpp
shannonbooth: Ah, I actually forgot about this, _technically_ this specific function isn't in the URL spec yet, there is an open MR for it (implementation will definitely be this though). Though, in any case, I don't know how to possibly link this specific function since there is nothing to click on to get the link, I could only find the ones in the HTML by linking to the references of the generic link for 'extract an origin' https://html.spec.whatwg.org/multipage/browsers.html#extract-an-origin

Not sure there is any way of getting a link for this specific case (even when merged) 🤔

Libraries/LibJS/Runtime/Value.h
```diff
@@ -154,6 +154,24 @@ class JS_API Value : public GC::NanBoxedValue {
...
+    }
+
+    template<typename T>
+    ALWAYS_INLINE T const* as_if() const
+    {
+        static_assert(IsBaseOf<Object, T>);
+        if (!is_object())
+            return nullptr;
+        return ::as_if<T>(as_object());
+    }
```
gmta, line 173: Commit message typo: "eaiser" -> "easier"

Libraries/LibURL/Host.cpp
```diff
@@ -225,15 +225,30 @@ Optional<String> Host::registrable_domain() const
...
+        if (last_dot.has_value()) {
+            view = view.substring_view(0, *last_dot);
+            auto second_last_dot = view.find_last('.');
+            if (second_last_dot.has_value())
+                registrable_domain = MUST(host_string.substring_from_byte_offset(second_last_dot.value() + 1));
+        }
+    }
+
+    if (!registrable_domain.has_value())
+        registrable_domain = host_string;
```
gmta, line 244: Commit message: two things, the title states "in case of missing in PSL" which seems to be missing a word. Maybe this would work better?

> LibURL: Return a proper registrable domain if it's missing in the PSL

Also I find this hard to read:

> This fixes us, for example, considering https://b.b.example/ and https://a.example/ as same site to one another.

Suggestion:

> This prevents us from considering e.g. https://b.b.example/ and https://a.example/ as the same site.

Libraries/LibWeb/DOMURL/DOMURL.cpp
gmta: You could link the open PR as a FIXME?

Libraries/LibWeb/DOMURL/Origin.cpp
```diff
@@ -0,0 +1,95 @@
...
+bool Origin::opaque() const
+{
+    // The opaque getter steps are to return true if this's origin is an opaque origin; otherwise false.
+    return m_origin.is_opaque();
+}
+
+// https://html.spec.whatwg.org/multipage/browsers.html#dom-origin-issameorigin
+bool Origin::is_same_origin(Origin const& other) const
+{
+    // The isSameOrigin(other) method steps are to return true if this's origin is same origin with other's origin; otherwise false.
```
gmta: Please wrap comments and code at column 120.

Libraries/LibWeb/DOMURL/DOMURL.cpp
shannonbooth: Done, can defer the problem of trying to figure out a link until that's in 😅

gmta (approved):

wertzuz:
In the second commit message it should have been second to last dot, I think.

PR #7049, opened 2025-12-07
LibMedia+LibWeb: Create demuxer and extract tracks on a separate thread

kalenikaliaksandr:
Demuxer creation and track+duration extraction are moved to a separate thread so that the media data byte buffer is no longer accessed from the main thread. This will be important once the buffer is populated incrementally, as having the main thread both populate and read from the same buffer could easily lead to deadlocks. Aside from that, moving demuxer creation off the main thread helps to be more responsive.

`VideoDataProvider` and `AudioDataProvider` now accept the main thread event loop pointer as they are constructed from the thread responsible for demuxer creation.

This is a part of https://github.com/LadybirdBrowser/ladybird/pull/6921 that could be merged independently.

### DIFF
--- a/Libraries/LibMedia/PlaybackManager.cpp
+++ b/Libraries/LibMedia/PlaybackManager.cpp
@@ -5,6 +5,7 @@
  */
 
 #include <LibMedia/Containers/Matroska/MatroskaDemuxer.h>
+#include <LibMedia/Demuxer.h>
 #include <LibMedia/FFmpeg/FFmpegDemuxer.h>
 #include <LibMedia/MutexedDemuxer.h>
 #include <LibMedia/PlaybackStates/PausedStateHandler.h>
@@ -15,23 +16,21 @@
 #include <LibMedia/Sinks/AudioMixingSink.h>
 #include <LibMedia/Sinks/DisplayingVideoSink.h>
 #include <LibMedia/Track.h>
+#include <LibThreading/Thread.h>
 
 #include "PlaybackManager.h"
 
 namespace Media {
 
-DecoderErrorOr<NonnullRefPtr<PlaybackManager>> PlaybackManager::try_create(ReadonlyBytes data)
+DecoderErrorOr<void> PlaybackManager::prepare_playback_from_media_data(ReadonlyBytes media_data, Core::EventLoop& main_thread_event_loop)
 {
     auto inner_demuxer = TRY([&] -> DecoderErrorOr<NonnullRefPtr<Demuxer>> {
-        auto matroska_result = Matroska::MatroskaDemuxer::from_data(data);
+        auto matroska_result = Matroska::MatroskaDemuxer::from_data(media_data);
         if (!matroska_result.is_error())
             return matroska_result.release_value();
-        return TRY(FFmpeg::FFmpegDemuxer::from_data(data));
+        return TRY(FFmpeg::FFmpegDemuxer::from_data(media_data));
     }());
     auto demuxer = DECODER_TRY_ALLOC(try_make_ref_counted<MutexedDemuxer>(inner_demuxer));
-
-    // Create the weak wrapper.
-    auto weak_playback_manager = DECODER_TRY_ALLOC(try_make_ref_counted<WeakPlaybackManager>());
 
     // Create the video tracks and their data providers.
     auto all_video_tracks = TRY(demuxer->get_tracks_for_type(TrackType::Video));
@@ -41,7 +40,7 @@ DecoderErrorOr<NonnullRefPtr<PlaybackManager>> PlaybackManager::try_create(Reado
     supported_video_tracks.ensure_capacity(all_video_tracks.size());
     supported_video_track_datas.ensure_capacity(all_video_tracks.size());
     for (auto const& track : all_video_tracks) {
-        auto video_data_provider_result = VideoDataProvider::try_create(demuxer, track);
+        auto video_data_provider_result = VideoDataProvider::try_create(main_thread_event_loop, demuxer, track);
         if (video_data_provider_result.is_error())
             continue;
         supported_video_tracks.append(track);
@@ -58,7 +57,7 @@ DecoderErrorOr<NonnullRefPtr<PlaybackManager>> PlaybackManager::try_create(Reado
     supported_audio_tracks.ensure_capacity(all_audio_tracks.size());
     supported_audio_track_datas.ensure_capacity(all_audio_tracks.size());
     for (auto const& track : all_audio_tracks) {
-        auto audio_data_provider_result = AudioDataProvider::try_create(demuxer, track);
+        auto audio_data_provider_result = AudioDataProvider::try_create(main_thread_event_loop, demuxer, track);
         if (audio_data_provider_result.is_error())
             continue;
         auto audio_data_provider = audio_data_provider_result.release_value();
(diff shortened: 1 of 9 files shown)


Zaggy1024:
I think we may be able to avoid having a ThreadData class like in AudioDataProvider and VideoDataProvider if we just use WeakPlaybackManager to directly modify the relevant fields of the PlaybackManager in a deferred_invoke back to the main thread in initialize_from_data.

We should probably also use the existing error callback for initialization errors, make that private and set it as part of the constructor so that the media element is able to react to initialization errors the same way as playback errors (assuming the spec doesn't differentiate those?).

Also, I think it makes sense to have the track lists initially empty and just append to them as we receive data instead of replacing them, since that's my understanding for how new data is handled when MSE comes into the mix.

kalenikaliaksandr:
> We should probably also use the existing error callback for initialization errors, make that private and set it as part of the constructor so that the media element is able to react to initialization errors the same way as playback errors (assuming the spec doesn't differentiate those?).

 https://html.spec.whatwg.org/multipage/media.html#media-data-processing-steps-list seems to define differently error handling for those cases:
- If the can be fetched but is found by inspection to be in an unsupported format, or can otherwise not be rendered at all -> cancel fetching
- If the media data is corrupted -> cancel fetching + more steps

> I think we may be able to avoid having a ThreadData class like in AudioDataProvider and VideoDataProvider if we just use WeakPlaybackManager to directly modify the relevant fields of the PlaybackManager in a deferred_invoke back to the main thread in initialize_from_data.

sure, I made appropriate changes.

Zaggy1024:
> https://html.spec.whatwg.org/multipage/media.html#media-data-processing-steps-list seems to define differently error handling for those cases:
> 
> * If the can be fetched but is found by inspection to be in an unsupported format, or can otherwise not be rendered at all -> cancel fetching
> * If the media data is corrupted -> cancel fetching + more steps

Gotcha, in that case we would probably be fine to pass a callback for handling those format errors early enough that they can be guaranteed to be handled. For playback errors, we can keep the public callback, since it can be set before any of the decoding threads start working.

Libraries/LibMedia/PlaybackManager.cpp
Zaggy1024, line 50: We could probably drop these shrinks, since we're appending these to an existing vector later.

(Unless moving into `extend` onto an empty vector reuses storage?)

Libraries/LibMedia/PlaybackManager.cpp
```diff
@@ -71,33 +70,80 @@ DecoderErrorOr<NonnullRefPtr<PlaybackManager>> PlaybackManager::try_create(Reado
...
+    auto duration = demuxer->total_duration().value_or(AK::Duration::zero());
+
+    main_thread_event_loop.deferred_invoke([weak_playback_manager, on_complete = move(on_complete), video_tracks = move(supported_video_tracks), video_track_datas = move(supported_video_track_datas), preferred_video_track, audio_tracks = move(supported_audio_tracks), audio_track_datas = move(supported_audio_track_datas), preferred_audio_track, duration] mutable {
+        auto playback_manager = weak_playback_manager->take_strong();
+        if (!playback_manager)
+            return;
+
+        RefPtr<AudioMixingSink> audio_sink = nullptr;
+        if (!audio_tracks.is_empty())
+            audio_sink = AudioMixingSink::try_create().release_value_but_fixme_should_propagate_errors();
```
Zaggy1024: Feel free to MUST() this for now, and we can probably do away with the `try_` part later.

Libraries/LibMedia/PlaybackManager.cpp
```diff
@@ -71,33 +70,80 @@ DecoderErrorOr<NonnullRefPtr<PlaybackManager>> PlaybackManager::try_create(Reado
...
+
+        // Create the time provider.
+        auto time_provider = ([&] -> NonnullRefPtr<MediaTimeProvider> {
+            if (audio_sink)
+                return make_ref_counted<WrapperTimeProvider<AudioMixingSink>>(*audio_sink);
+            return make_ref_counted<GenericTimeProvider>();
+        })();
```
Zaggy1024: We could potentially create a GenericTimeProvider by default when initializing the PlaybackManager, and then swap it out for an AudioMixingSink when we find an audio track, so we can keep the time provider field non-null and avoid some assertions.

I think it should suffice to do
```cpp
if (audio_sink) {
    auto current_time = playback_manager->current_time();
    playback_manager->m_time_provider = make_ref_counted<WrapperTimeProvider<AudioMixingSink>>(*audio_sink);
    playback_manager->m_time_provider->set_time(current_time);
}
```

Libraries/LibMedia/PlaybackManager.cpp
```diff
@@ -71,33 +70,80 @@ DecoderErrorOr<NonnullRefPtr<PlaybackManager>> PlaybackManager::try_create(Reado
...
+    return {};
+}
+
+NonnullRefPtr<PlaybackManager> PlaybackManager::create(ReadonlyBytes data, OnCreateComplete on_complete, OnCreateError on_error)
+{
+    // Create the weak wrapper.
+    auto weak_playback_manager = make_ref_counted<WeakPlaybackManager>();
+
+    auto thread = Threading::Thread::construct([weak_playback_manager, data, on_complete = move(on_complete), on_error = move(on_error), &main_thread_event_loop = Core::EventLoop::current()] mutable -> int {
+        auto playback_manager = weak_playback_manager->take_strong();
```
Zaggy1024: I realized, having you use the weak reference here is pointless. 😅 If we're taking a strong reference immediately, and the thread has a limited lifetime, we'll be just fine creating a NonnullRefPtr in the captures.

Zaggy1024 (commented):
Sorry to bring you another potential refactor, but I wonder if we could make the playback manager setup something like:
```cpp
m_playback_manager = PlaybackManager::create();
m_playback_manager.on_track_added = [...](Track const& track) { ... }; // Add a track based on the instructions for audio or video tracks in spec, set video resolution if a video track is selected
m_playback_manager.on_metadata_parsed = [...] { ... }; // Set the ready state to HAVE_METADATA, set duration

m_playback_manager.add_media_source(m_media_data); // Triggers the above callbacks once the necessary data has been received
```
Hopefully we can also keep the getters for the preferred tracks and set their backing fields before we call `on_track_added`, so that the media element steps can check if there's a preferred track to enable when it receives a new track to run the processing steps on.

I also have my branch that adds an on_duration_changed event, but I think we probably want to keep the metadata parsed part separate.

Libraries/LibMedia/PlaybackManager.cpp
kalenikaliaksandr: yes, we borrow storage from appended vector in case destination vector was empty https://github.com/LadybirdBrowser/ladybird/blob/master/AK/Vector.h#L715-L718

Libraries/LibMedia/PlaybackManager.cpp
kalenikaliaksandr: I tried to implement it precisely as you suggested above, but `AudioMixingSink::set_time()` would try to access uninitialized `m_playback_stream` then, so instead I made it to verify that `current_time()` of previous time provider is zero. I think it should be safe to assume `current_time()` is always zero at this point, right?

Zaggy1024: Ah, I guess I forgot to note how to handle when we already have an audio mixing sink, but in that case we don't need to recreate it. Technically, we could _always_ use an AudioMixingSink, since it will handle adding tracks just fine, but GenericTimeProvider is obviously much lighter.

Zaggy1024: So it would be something along the lines of:

```cpp
if (!playback_manager->m_audio_tracks.is_empty() && playback_manager->m_audio_sink == nullptr) {
    playback_manager->m_audio_sink = MUST(AudioMixingSink::try_create());
    auto current_time = playback_manager->current_time();
    playback_manager->m_time_provider = make_ref_counted<WrapperTimeProvider<AudioMixingSink>>(*audio_sink);
    playback_manager->m_time_provider->set_time(current_time);
}
```
I believe this should avoid any issues with an unset playback stream, hopefully.

(I should probably look at making it so that setting time doesn't assume the playback stream exists though, that setup is deferred so it is possible for that to happen in normal use...)

Libraries/LibMedia/PlaybackManager.cpp
```diff
@@ -71,41 +70,85 @@ DecoderErrorOr<NonnullRefPtr<PlaybackManager>> PlaybackManager::try_create(Reado
...
     m_weak_wrapper->revoke();
 }
 
+void PlaybackManager::add_media_source(ReadonlyBytes media_data)
+{
+    auto thread = Threading::Thread::construct([playback_manager = NonnullRefPtr { *this }, media_data, &main_thread_event_loop = Core::EventLoop::current()] -> int {
+        auto maybe_error = playback_manager->prepare_playback_from_media_data(media_data, main_thread_event_loop);
+        if (maybe_error.is_error() && playback_manager->on_unsupported_format_error) {
+            main_thread_event_loop.deferred_invoke([playback_manager, error = maybe_error.release_error()] mutable {
+                playback_manager->on_unsupported_format_error(move(error));
```
Zaggy1024: We should probably check if `on_unsupported_format_error` is present within the callback, just to make sure we're not accessing it across threads. In all valid usage, it should already be set before the thread starts, but better safe than sorry.

Libraries/LibMedia/PlaybackManager.cpp
```diff
@@ -71,41 +70,85 @@ DecoderErrorOr<NonnullRefPtr<PlaybackManager>> PlaybackManager::try_create(Reado
...
+
+        if (playback_manager->on_track_added) {
+            for (auto const& audio_track : playback_manager->m_audio_tracks)
+                playback_manager->on_track_added(TrackType::Audio, audio_track);
+            for (auto const& video_track : playback_manager->m_video_tracks)
+                playback_manager->on_track_added(TrackType::Video, video_track);
+        }
+
+        if (playback_manager->on_metadata_parsed)
+            playback_manager->on_metadata_parsed(playback_manager->m_duration);
```
Zaggy1024: I think it makes sense to have `on_metadata_parsed` take no parameters, since the metadata should never change while running synchronously on the main thread, and it can be accessible through getters.

Libraries/LibMedia/PlaybackManager.h
```diff
@@ -79,6 +76,11 @@ class MEDIA_API PlaybackManager final : public AtomicRefCounted<PlaybackManager>
 
     Function<void()> on_playback_state_change;
     Function<void(DecoderError&&)> on_error;
+    Function<void(AK::Duration)> on_metadata_parsed;
+    Function<void(DecoderError&&)> on_unsupported_format_error;
+    Function<void(TrackType, Track const&)> on_track_added;
```
Zaggy1024, line 81: We could probably move these three added callbacks first, since they should normally be called before the existing two.

Libraries/LibWeb/HTML/HTMLMediaElement.cpp
```diff
@@ -1191,148 +1181,110 @@ void HTMLMediaElement::update_video_frame_and_timeline()
...
-    }
+    auto event = TrackEvent::create(realm, HTML::EventNames::addtrack, move(event_init));
+    m_video_tracks->dispatch_event(event);
+}
 
-    if (preferred_video_track.has_value())
-        VERIFY(has_selected_preferred_video_track);
+void HTMLMediaElement::on_metadata_parsed(AK::Duration const& duration)
+{
+    update_volume();
```
Zaggy1024, line 1287: As a note to me to fix this later, could we add a FIXME to move this to setup_playback_manager? I know it won't work there at the moment, since it only sets the volume if the audio sink exists, but we can make it so that we store volume in PlaybackManager and set it upon creating the mixing sink instead.

Libraries/LibWeb/HTML/HTMLMediaElement.cpp
```diff
@@ -1395,6 +1356,67 @@ WebIDL::ExceptionOr<void> HTMLMediaElement::process_media_data(Function<void(Str
...
+        failure_callback(MUST(String::from_utf8(error.description())));
+    };
+
+    m_playback_manager->add_media_source(m_media_data);
+
+    // NOTE: The spec is unclear on whether the following media resource track conditions should trigger multiple
+    //       times on one media resource, but it is implied to be possible by the start of the "Media elements"
+    //       section, where it says that a "media resource can have multiple audio and video tracks."
+    //       https://html.spec.whatwg.org/multipage/media.html#media-elements
+    //       Therefore, we enumerate all the available tracks into our VideoTrackList and AudioTrackList.
```
Zaggy1024: This note could go above the calls to on_(audio|video)_track_added.

Zaggy1024: Oh also, this is a non-spec comment, so if we've switched to using `NB:`, this should be changed.

PR #7050, opened 2025-12-07
LibWeb: Background parsing and serialization improvements

Calme1709:
Various improvements to the parsing and serialization of the `background` property and it's shorthands.

Gains us at least 200 WPT tests.

### DIFF
--- a/Libraries/LibWeb/CSS/StyleValues/StyleValueList.cpp
+++ b/Libraries/LibWeb/CSS/StyleValues/StyleValueList.cpp
@@ -61,7 +61,7 @@ String StyleValueList::to_string(SerializationMode mode) const
     }
 
     auto first_value = m_properties.values.first();
-    if (all_of(m_properties.values, [&](auto const& property) { return property == first_value; }))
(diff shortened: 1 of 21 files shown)


Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -2273,6 +2273,22 @@ static NonnullRefPtr<StyleValue const> compute_style_value_list(NonnullRefPtr<St
...
+    auto const& values = style_value->as_value_list().values();
+
+    if (values.size() == n)
+        return style_value;
+
+    StyleValueVector repeated_values;
+    repeated_values.ensure_capacity(n);
+
+    for (size_t i = 0; i < n; ++i)
+        repeated_values.append(values[i % values.size()]);
```
AtkinsSJ: You can `append_unchecked()` to save a little bit of bookkeeping.

Also there is a `StyleValueList::value_at()` method which also does the wrapping, but it doesn't really save anything.

Calme1709: Done. May as well use the `StyleValueList::value_at()` method if it's already there.

PR #7057, opened 2025-12-08
LibJS: Throw when new is used on non-constructor methods

### LINKED ISSUE #6926: JS: non-constructor-methods should not create a new object
```javascript
// this is ok
class OK {constructor() {}}
const ok = new OK()
console.log(new ok.constructor)

// should throw
class FAIL {m(){}}
const fail = new FAIL()
console.log(new fail.m)

// should throw
console.log(new ({ m() {}}).m)
```

We are currently printing for all 3 cases `Object{}`

BinaryAlley:
Non-constructor methods and plain object methods were incorrectly treated as constructors when used with "new", resulting in a new Object instance instead of throwing a TypeError. Example:

// ok: class constructor
class OK { constructor() {} }
const ok = new OK();
new ok.constructor;

// should throw: instance method
class FAIL { m() {} }
const fail = new FAIL();
new fail.m;

// should throw: method on object literal
new ({ m() {} }).m;

// should throw: function expression used as object property
new ({ b: function () { this.x = 2; } }).b();

// should throw: extracted class method used with new
const m = class { m() {} }.prototype.m;
new m;

Fixes #6926

### DIFF
--- a/Libraries/LibJS/Bytecode/Interpreter.cpp
+++ b/Libraries/LibJS/Bytecode/Interpreter.cpp
@@ -12,6 +12,7 @@
 #include <LibJS/AST.h>
 #include <LibJS/Bytecode/BasicBlock.h>
 #include <LibJS/Bytecode/FormatOperand.h>
+#include <LibJS/Bytecode/FunctionDefinitionKind.h>
 #include <LibJS/Bytecode/Generator.h>
 #include <LibJS/Bytecode/Instruction.h>
 #include <LibJS/Bytecode/Interpreter.h>
@@ -1023,10 +1024,10 @@ inline ThrowCompletionOr<void> throw_if_needed_for_call(Interpreter& interpreter
 }
 
 // 15.2.5 Runtime Semantics: InstantiateOrdinaryFunctionExpression, https://tc39.es/ecma262/#sec-runtime-semantics-instantiateordinaryfunctionexpression
-static Value instantiate_ordinary_function_expression(Interpreter& interpreter, FunctionNode const& function_node, Utf16FlyString const& given_name, bool should_be_constructible)
+static Value instantiate_ordinary_function_expression(Interpreter& interpreter, FunctionNode const& function_node, Utf16FlyString const& given_name, MakeConstructor make_constructor)
 {
-    // NB:  If should_be_constructible is true, this acts like 15.2.5 (Function Expression) and calls MakeConstructor internally;
-    // if it's false, it acts like 15.4.4 (DefineMethod) and skips MakeConstructor.
+    // NB:  If make_constructor is Invoke, this acts like 15.2.5 (Function Expression) and calls MakeConstructor internally;
+    // if it's Skip, it acts like 15.4.4 (DefineMethod) and skips MakeConstructor.
 
     auto own_name = function_node.name();
     auto has_own_name = !own_name.is_empty();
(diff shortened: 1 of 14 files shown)


gmta:
Welcome! Can you add some of those examples to a regression test?

BinaryAlley:
Thank you! I'll attempt to do so.

Libraries/LibJS/Bytecode/Interpreter.cpp
```diff
@@ -1264,8 +1264,11 @@ inline ThrowCompletionOr<void> throw_if_needed_for_call(Interpreter& interpreter
 }
 
 // 15.2.5 Runtime Semantics: InstantiateOrdinaryFunctionExpression, https://tc39.es/ecma262/#sec-runtime-semantics-instantiateordinaryfunctionexpression
-static Value instantiate_ordinary_function_expression(Interpreter& interpreter, FunctionNode const& function_node, Utf16FlyString given_name)
+static Value instantiate_ordinary_function_expression(Interpreter& interpreter, FunctionNode const& function_node, Utf16FlyString given_name, bool should_be_constructible)
```
konradekk: it’d be probably better to name new/last parameter as `bool invoke_make_constructor`, or even better: `MakeConstructor make_constructor` with type enumerated as `Invoke` and `Skip`.

(probably the same for `is_method` in all forms to avoid mental strain?)

gmta:
@BinaryAlley please incorporate (style) fixes into the commits that introduced the issues - so no separate fix commits please.

BinaryAlley:
I think I implemented the requested changes the way you mentioned them. If something is not right, I'll change it again.

gmta:
> I think I implemented the requested changes the way you mentioned them. If something is not right, I'll change it again.

The second commit still contains whitespace fixes etc. that should have gone in the first commit. Maybe squash those two commits into one?

BinaryAlley:
Done, the two commits are now merged.

BinaryAlley:
It seems that Prettier changed some tests, i.e., from this:
```js
test("new on class instance method throws TypeError", () => {
    class FAIL {
        m() {}
    }
    const fail = new FAIL();

    expect(() => {
        new fail.m;
    }).toThrowWithMessage(TypeError, "");
});
```

to this:

```js
test("new on class instance method throws TypeError", () => {
    class FAIL {
        m() {}
    }
    const fail = new FAIL();

    expect(() => {
        new fail.m();
    }).toThrowWithMessage(TypeError, "");
});
```

which changed the semantic of what was tested. I updated the commits to make Prettier ignore these tests.

BinaryAlley:
I'm going to need a bit of help here, I spent days trying to figure out if the test failure occurs due to my changes, or something else. The failure is indicated to be
```
Pass: 5546, Fail: 1, Skipped: 225, Timeout: 0, Crashed: 0
Fail: Text/input/GamepadAPI/gamepad-rumble.html
```
But I fail to see what does that have to do with instantiation JavaScript objects? I also don't understand why it says:
```
The following tests FAILED:
	168 - LibWeb (Failed)
```
My changes were entirely in LibJS, the only file modified in LibWeb was a function call signature that needed to be updated. 
In the logs, I also see tons of javascript unhandled exceptions, like this one, for instance:
```JavaScript
1678.600 WebContent(5354): Unhandled JavaScript exception: [Error] 
1678.600 WebContent(5354):     at Error
    at file:///home/computer/ladybird/Tests/LibWeb/Text/input/wpt-import/IndexedDB/fire-error-event-exception.any.js:79:20
```
But according to the test itself, it is an expected exception thrown in a try...catch. That doesn't make any easier the process of finding out if my changes introduced real unhandled exceptions, as opposed to "expected" test ones.
In the logs I also see mentions of a memory leak? Is this actually caused by subsequent failures due to my changes? And why do the other platforms pass, but only Linux arm64 fails?

shannonbooth:
Unfortunately I believe you have run into some flakey tests, and is not the fault of these changes. For example, the gamepad rumble test is unfortunately known to be flakey, https://github.com/LadybirdBrowser/ladybird/pull/7346 is one such PR intending to fix that problem.

BinaryAlley:
Thanks for the assistance! In that case, do I need to take any further action on my part, or just wait for the tests to be re-run?

AtkinsSJ:
You do actually need to rebase this on master - the CI rules changed, and so they're now waiting for a run that didn't exist when you pushed before.

PR #7062, opened 2025-12-08
LibWeb: Set `background-image` presentational hint as `StyleValueList`

Calme1709:
This fixes a crash introduced in a11666e


tcl3 (commented):
LGTM, now that we're not avoiding StyleValueLists  with a single value, it might be worth adding a `StyleValueList::from_single_value()` or similar to cut down on unnecessary boilerplate.

Calme1709:
I think the only boilerplate it will save is the array initialization - I think that we would still need to specify the correct separator as single-valued `StyleValueList`s can become multi-valued in some cases such as `repeat_style_value_list_to_n_elements` in StyleComputer.cpp and `StylePropertyMap::append` (note that this one doesn't actually reuse the separator value yet but it will be required to resolve a FIXME in there).

PR #7083, opened 2025-12-10
LibWeb: Save weak pointer to document in DocumentLoadEventDelayer

kalenikaliaksandr:
This prevents GC leaks caused by a DocumentLoadEventDelayer that is never destroyed (for example, when its Document becomes inactive) from holding a strong reference to the Document and keeping it alive.


awesomekling:
I added the missing functionality in GC::Weak and merged this in #7223

Issue #7052, opened 2025-12-08
Page doesn’t scroll


vipulbhasin23:
### Summary

Scrollbars show up but don’t work on a page.

### Operating system

macOS

### Steps to reproduce

1. Navigate to https://duckduckgo.com
2. Click the link to set DuckDuckGo as the default search engine (https://duckduckgo.com/set-as-default)
3. Notice that on the navigated page, the scrollbars show up but the page doesn’t scroll
4. Also notice that navigating to the URL directly doesn’t cause this problem and the scrolling works as expected.

### Expected behavior

Scrollbars should scroll the page

### Actual behavior

The page doesn’t scroll

### URL for a reduced test case

https://duckduckgo.com/duckduckgo-help-pages/change-default-search-engine

### HTML/SVG/etc. source for a reduced test case

```html
N/A
```

### Log output and (if possible) backtrace

```shell
2025-12-08 12:52:08.958 Ladybird[77809:13547316] WARNING <NSToolbarItem: 0xc89409dc0> -> view was automatically measured but had an ambiguous height or width and the view's frame size had a zero height or width. Did you forget to add constraints on your view or its subview(s)? Try adding constraints or give the view an intrinsicContentSize
2025-12-08 12:52:08.958 Ladybird[77809:13547316] WARNING <NSToolbarItem: 0xc89409dc0> -> view was automatically measured but had an ambiguous height or width and the view's frame size had a zero height or width. Did you forget to add constraints on your view or its subview(s)? Try adding constraints or give the view an intrinsicContentSize
2025-12-08 12:52:08.971 Ladybird[77809:13547316] WARNING <NSToolbarItem: 0xc89409dc0> -> view was automatically measured but had an ambiguous height or width and the view's frame size had a zero height or width. Did you forget to add constraints on your view or its subview(s)? Try adding constraints or give the view an intrinsicContentSize
2025-12-08 12:52:08.971 Ladybird[77809:13547316] WARNING <NSToolbarItem: 0xc89409dc0> -> view was automatically measured but had an ambiguous height or width and the view's frame size had a zero height or width. Did you forget to add constraints on your view or its subview(s)? Try adding constraints or give the view an intrinsicContentSize
[...]

PR #7085, opened 2025-12-10
LibWeb: Filter input events by page_id in EventLoop

pepperoni21:
Currently, the `process_input_events` method of `EventLoop` does not check the page id of the events before processing them. So when two documents share the same web process, the input events are all processed by one page.

This can lead to issues where a tab opened through a hyperlink cannot be interacted at all.

This PR edits the method to only process the input events of the current page, and puts the other events back in the queue to be processed by the next pages.

Fixes https://github.com/LadybirdBrowser/ladybird/issues/7052


pepperoni21:
> a quick fix, but it seems that the underlying issue is with collecting? 🤔

I'm not sure what you mean with "collecting", do you mean the fact that multiple pages share the same event queue?

Issue #7069, opened 2025-12-09
Error loop when loading thegenerationofdanger.com


muzikbike:
### Summary

The error appears to be the same as in #7032, but despite that one being fixed, this one is not, indicating a different cause.

### Operating system

Linux

### Steps to reproduce

1. Go to thegenerationofdanger.com
2. Wait for the Unity logo to appear then disappear
3. Observe the log, and lack of further content loading

### Expected behavior

We'd get into the game.

### Actual behavior

That does not happen.

### URL for a reduced test case

thegenerationofdanger.com

### HTML/SVG/etc. source for a reduced test case

```html
N/A
```

### Log output and (if possible) backtrace

```shell
ninja: Entering directory `/home/conputer/other-dev/ladybird/20250807210822/ladybird/Build/release'
ninja: no work to do.
14568.099 WebContent(25861): (js log) "Loading player data from data.unity3d\n"
14568.110 WebContent(25861): (js log) "Initialize engine version: 2020.3.30f1 (1fb1bf06830e)\n"
14568.111 WebContent(25861): (js log) "Creating WebGL 2.0 context."
14568.116 WebContent(25861): (js log) "Renderer: ANGLE (AMD, AMD Radeon RX 7800 XT (radeonsi navi32 LLVM 20.1.2 DRM 3.57 6.8.0-88-generic), OpenGL ES 3.2)\n"
14568.116 WebContent(25861): (js log) "Vendor:   Google Inc. (AMD)\n"
14568.116 WebContent(25861): (js log) "Version:  OpenGL ES 3.0 (OpenGL ES 3.0.0 (ANGLE 2.1.7258 git hash: 79ac1a8cd767))\n"
14568.116 WebContent(25861): (js log) "GLES:     3\n"
[...]

Issue #3413, opened 2025-01-31
Crash related to selection in editing pane


mnlrsn:
### Summary

When using the code pad at uiua.org, there is a crash related when double clicking in the editor pane. Not 100% reproducible, but it happens "all the time".

### Operating system

Linux

### Steps to reproduce

1. Go to https://www.uiua.org/pad?src=0_15_0-dev_1__4oehMTAKKzEKLgriip7Dlwo=
2. Double click in the editor window to select a line of code
3. Double click on a different row to deselect the first one and select the second one

It also seems to occur if the line is selected by clicking and dragging the mouse over a line of code, then clicking on a different line.

### Expected behavior

It should select the second line of text without crashing.

### Actual behavior

In more than half the cases it will produce a crash

This has been verified on macOS 14.7.2 M1

### URL for a reduced test case

https://www.uiua.org/pad?src=0_15_0-dev_1__4oehMTAKKzEKLgriip7Dlwo=

### HTML/SVG/etc. source for a reduced test case

```html
N/A
```

### Log output and (if possible) backtrace

```shell
UNEXPECTED ERROR: [IndexSizeError]: Selection.collapse() with offset longer than node's length at /Users/xxx/Workspace/Ladybird/ladybird/Libraries/LibWeb/DOM/EditingHostManager.cpp:83
0   liblagom-ak.0.0.0.dylib             0x00000001032de904 ak_trap + 56
1   liblagom-web.0.0.0.dylib            0x0000000104dab948 Web::DOM::EditingHostManager::set_selection_anchor(GC::Ref<Web::DOM::Node>, unsigned long) + 296
2   liblagom-web.0.0.0.dylib            0x0000000104ecaebc non-virtual thunk to Web::HTML::FormAssociatedTextControlElement::set_selection_anchor(GC::Ref<Web::DOM::Node>, unsigned long) + 76
3   liblagom-web.0.0.0.dylib            0x00000001050aec00 Web::EventHandler::handle_doubleclick(Gfx::Point<Web::CSSPixels>, Gfx::Point<Web::CSSPixels>, unsigned int, unsigned int, unsigned int) + 1048
4   WebContent                          0x0000000102627104 WebContent::ConnectionFromClient::process_next_input_event() + 476
5   liblagom-web.0.0.0.dylib            0x0000000104b90c10 AK::Function<void (Web::CSS::CSSRule const&)>::operator()(Web::CSS::CSSRule const&) const + 76
[...]

Issue #2502, opened 2024-11-22
Lethal Company time tracker doesn't load


shlyakpavel:
### Summary

When I open a [time tracker](https://ratijas.me/lc/) in Ladybird, it shows `Application exit (Wasm execution trapped (WIP): Memory access out of bounds)` while loading

### Operating system

macOS

### Steps to reproduce

1. Open Ladybird
2. Navigate to https://ratijas.me/lc/
3. Observe

### Expected behavior

Web app loads

### Actual behavior

Web app not loads

### URL for a reduced test case

https://ratijas.me/lc/

### HTML/SVG/etc. source for a reduced test case

```html
N/A
```

### Log output and (if possible) backtrace

```shell
2024-11-22 18:15:24.869 Ladybird[99584:4267125] +[IMKClient subclass]: chose IMKClient_Modern
2024-11-22 18:15:24.914 Ladybird[99584:4267125] +[IMKInputSession subclass]: chose IMKInputSession_Modern
183290.099 WebContent(99588): LibWasm: Memory access out of bounds (expected 3785438157 to be less than or equal to 52428800)
183290.100 WebContent(99588): (js error) [TypeError] Wasm execution trapped (WIP): Memory access out of bounds
183290.100 WebContent(99588): (js error) "TypeError: Wasm execution trapped (WIP): Memory access out of bounds\n    at Yh\n    at callRuntimeCallbacks (https://ratijas.tk/lc/appLethalClocksQml.js:8:9430)\n    at initRuntime (https://ratijas.tk/lc/appLethalClocksQml.js:8:4168)\n    at doRun (https://ratijas.tk/lc/appLethalClocksQml.js:8:296491)\n    at run (https://ratijas.tk/lc/appLethalClocksQml.js:8:296792)\n    at runCaller (https://ratijas.tk/lc/appLethalClocksQml.js:8:295879)\n    at removeRunDependency (https://ratijas.tk/lc/appLethalClocksQml.js:8:5154)\n    at receiveInstance (https://ratijas.tk/lc/appLethalClocksQml.js:8:7609)\n    at receiveInstantiationResult (https://ratijas.tk/lc/appLethalClocksQml.js:8:7745)\n    at <unknown>\n"
```

### Screenshots or screen recordings

<img width="1138" alt="image" src="https://github.com/user-attachments/assets/fb6a4cd1-2f92-4a42-aa9e-e0e3cfde61df">

### Contribute a patch?

[...]

PR #7106, opened 2025-12-11
LibWasm: Memory access out of bound debug enabled permanently

rcorsi:
We have two open issues #7025 and #7069 dealing with a likely wasm fault and the error seen is:

```
10126.258 WebContent(5800): LibWasm: Memory access out of bounds (expected 4294967303 to be less than or equal to 33554432)
```

from `Libraries/LibWasm/AbstractMachine/BytecodeInterpreter.cpp:3885`

and

```
973.475 WebContent(23111): LibWasm: Memory access out of bounds (expected 0 <= 4294967315 and 4294967319 <= 268435456)
```

from `Libraries/LibWasm/AbstractMachine/BytecodeInterpreter.cpp:4184`

I know we can turn these debugs on when needed, but why not leave them on permanently?  Even some are not existing, so I added them.

They are not triggered often, and shouldn't slow things down, or am I missing something?


alimpfard:
Sure, why not.

rcorsi:
thanks, that was quick.

shlyakpavel:
It's more than two https://github.com/LadybirdBrowser/ladybird/issues/2502

rcorsi:
@shlyakpavel ok thanks for pointing it out. I'll look into it too.

rcorsi:
and a fourth issue #3413 with `Memory access out of bounds` message.

PR #7110, opened 2025-12-12
LibJS: Always use GetGlobal for `undefined` in global scope

kalenikaliaksandr:
It's safe because `undefined` in global scope isn't allowed to be redefined.

### DIFF
--- a/Libraries/LibJS/Parser.cpp
+++ b/Libraries/LibJS/Parser.cpp
@@ -353,8 +353,12 @@ public:
             }
 
             if (m_type == ScopeType::Program) {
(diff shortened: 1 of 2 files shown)


awesomekling:
Can we have some tests that mess around with overriding undefined just to show that this isn’t totally broken ?

kalenikaliaksandr:
> Can we have some tests that mess around with overriding undefined just to show that this isn’t totally broken ?

good idea, added new test!

awesomekling:
This breaks:

```js
function hmm() {
    eval("var undefined = 123;");
    console.log(undefined);
}
hmm();
```

PR #7112, opened 2025-12-12
RequestServer: A couple of odds-and-ends cleanups

trflynn89:
A few things I've accumulated working on some disk cache stuff.

### DIFF
--- a/Services/RequestServer/Request.cpp
+++ b/Services/RequestServer/Request.cpp
@@ -154,6 +154,7 @@ void Request::notify_fetch_complete(Badge<ConnectionFromClient>, int result_code
 
 void Request::transition_to_state(State state)
 {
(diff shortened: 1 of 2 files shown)


Libraries/LibRequests/Request.h
```diff
@@ -82,12 +82,12 @@ class Request : public RefCounted<Request> {
...
 
 private:
-    explicit Request(RequestClient&, i32 request_id);
+    explicit Request(RequestClient&, u32 request_id);
 
     void set_up_internal_stream_data(DataReceived on_data_available);
 
     WeakPtr<RequestClient> m_client;
-    int m_request_id { -1 };
+    u32 m_request_id { 0 };
```
awesomekling: Maybe 64 bit instead so it can’t overflow from increments in our lifetime? :)

trflynn89: might as well!

PR #7113, opened 2025-12-12
Mathml painting 6772

AtkinsSJ:
This is a bit of a rescue of #6772. A lot of the code there wasn't actually needed, but I managed to make use of the sqrt and fraction painting, updated to work regardless of the pixel size, and to draw it as a single path. It's quite limited until we actually run layout, but it's something.

For example, this:
```html
<!DOCTYPE html>
<style>
  body {
    font-size: 4em;
  }
</style>
<math>
  <msqrt>
    <mfrac>
      <mn>1</mn>
      <mn>2</mn>
    </mfrac>
  </msqrt>
</math>
<br/>
<math>
  <msqrt>
    <mn>42</mn>
  </msqrt>
</math>
```

Now looks like this:
Screenshot_20251212_164955

Before:
Screenshot_20251212_165446

Chromium and Firefox for reference:

Screenshot_20251212_165130

Screenshot_20251212_165151


lpas:
I'm not sure this is the spec way to go for drawing sqrt. It looks like it just uses a path but should use a glyph https://www.w3.org/TR/mathml-core/#radicals-msqrt-mroot . The symbols can change depending on the font used

AtkinsSJ:
You are very correct, I should have looked at the spec myself. I'll close this then.

Issue #7091, opened 2025-12-10
Fetch error when loading https://ebay.com


tcl3:
When loading https://ebay.com no content appears and I get the following console output:
```
RequestServer(30716): Request::handle_complete_state: Unable to map error (23): "Failed writing received data to disk/application"
WebContent(30724): ResourceLoader: Failed load of: "https://www.ebay.com/", Error: Request finished with error, Duration: 835ms
WebContent(30724): FIXME: Load html page with an error if read of body failed.
```
Disabling http caching with `--disable-http{-disk}-cache` made no difference.

I've bisected the issue to: 305827438666e8e8a7aa6c7d2c2380d9d4ff59d6.

cc: @kalenikaliaksandr @trflynn89

Lubrsi:
This seems to have always been an issue, but for some reason the buffered fetch code was tolerant of it. eBay is sending a non-comformant gzip stream where the trailing data is garbage.

`gunzip` on the downloaded data gives:
```
gzip: ebay.gz: decompression OK, trailing garbage ignored
```
(credit to Tim Flynn for this)

According to this curl issue, eBay's server has been incorrect like this for at least 7 years, so I don't think eBay will be fixing it anytime soon: https://github.com/curl/curl/issues/2719#issuecomment-405667025
Plus, there's more servers like this according to the issue.

In the past, they seemed hesitant to tolerate non-conforming streams like this since it's ultimately the server that is wrong and they prioritise data integrity over silently dropping data:
https://github.com/curl/curl/issues/3796#issuecomment-487918730
https://github.com/curl/curl/pull/3825#issuecomment-541762505

There's multiple angles this could take:
- Be tolerant like the buffered fetch mode was
- Detect CURLE_WRITE_ERROR from the library itself and not generated by us and display what we have
- Patch curl to add options to deal with these quirks
- Disable curl's Accept-Encoding mode and implement decompression ourselves that is tolerant of non-conforming servers.

PR #7114, opened 2025-12-12
Meta: Patch curl to ignore all trailing bytes after end of zlib stream

Lubrsi:
Without this patch, curl errors on any trailing bytes after the end of
a zlib stream. For web compatibility, we must ignore trailing bytes
instead of erroring.

This matches Firefox and Chrome:
- https://searchfox.org/firefox-main/rev/04cf27582307a9c351e991c740828d54cf786b76/netwerk/streamconv/converters/nsHTTPCompressConv.cpp#607-613
- https://source.chromium.org/chromium/chromium/src/+/main:net/filter/gzip_source_stream.cc;l=240-250;drc=397107e399d4b0f4a6e801e7a888d7d71778529c;bpv=0;bpt=1

Fixes https://github.com/LadybirdBrowser/ladybird/issues/7091


ADKaster:
patching curl for this seems awkward from a distribution point of view. For example, your quirk patch is not included in the flatpak build. third-party packaging from e.g AUR and nixpkgs (not to mention future fedora/COPR/deb packages) would need updated to vendor a private libcurl with our patch applied.

1) is there a way to tolerate this error without digging into libcurl itself?
2) can we ask upstream libcurl if they'd be more receptive to an upstream 'tolerate naughty servers' toggle at this point in time?

alimpfard:
imo this should just be upstreamed (as a proper flag, not "ladybird quirks mode").

ADKaster:
Agree with Ali. To pull this in, I'd hope for

1) a separate commit pulling in libcurl as-is from vcpkg, to better determine what is "upstream port" and what is "ladybird changes" when looking at commit history

2) upstreamable patch with links to curl discussions or PR 

Given that bagder seems positive about ladybird using libcurl, hopefully a series of patches related to "browser quirks" will be more likely to be accepted now than when the referenced discussions happened.

PR #7115, opened 2025-12-12
LibWeb: Use default fetch processing for any applicable rel keyword 

### LINKED ISSUE #7089: HTMLLinkElement relationship types are treated as mutually exclusive when they shouldn't
The spec does not imply that relationship types are mutually exclusive, only that they should be fetched at appropriate times.

For example, this should still load the linked resources a stylesheet despite having `preload` in the `rel` attribute:
```html
<link rel="preload stylesheet" href="/assets/style.D9hCgQEe.css" as="style">
```

This snippet is from https://vite.dev/ where we fail to load any stylesheets because of this.

tcl3:
Previously, the `preload`, `preconnect` and `dns-prefetch` keywords took precedence over the others. When these keywords were present the default fetch processing steps would not occur even when a relevant keyword such as `stylesheet` or `icon` was present.

This makes `<link rel="preload stylesheet">` behave the same way as `<link rel="stylesheet">`, which seems to match the behavior of other browsers.

I'm not sure how best to write an automated test for this. I tested the change by running the following code on the WPT server using `./Meta/WPT.sh serve`:

`test.html`:
```html
<!DOCTYPE html>
<style>
    div {
        width: 100px;
        height: 100px;
        background-color: red;
    }
</style>
<!-- This request should block -->
<link rel="preload stylesheet" href="test.css?pipe=trickle(d5)" as="style">
<div></div>
```
`test.css`:
```css
div {
    background-color: green !important;
}
```

Fixes #7089

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLLinkElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLLinkElement.cpp
@@ -379,12 +379,12 @@ void HTMLLinkElement::fetch_and_process_linked_resource()
 {
     if (m_relationship & ~(Relationship::DNSPrefetch | Relationship::Preconnect | Relationship::Preload))
         default_fetch_and_process_linked_resource();
-    else if (m_relationship & Relationship::DNSPrefetch)
-        fetch_and_process_linked_dns_prefetch_resource();
-    else if (m_relationship & Relationship::Preconnect)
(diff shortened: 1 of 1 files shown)


tcl3:
My latest push adds a basic test to just test that `rel="preload stylesheet"` applies the given stylesheet.

Issue #7117, opened 2025-12-13
LibRegex: Lookahead in optional quantified group captures incorrectly


aplefull:
### Summary

We fail to match these correctly:

`"abc".match(/(?:(?=(abc)))?a/);`
`"abc".match(/(?:(?=(abc))){0,1}a/);`

Should be `["a", undefined]`, but we get `["a","abc"]`. From what I can tell, we don't handle Step 4 from [here](https://tc39.es/ecma262/#sec-runtime-semantics-repeatmatcher-abstract-operation), but I couldn't figure out how to do this properly :(

### Operating system

Linux

### Steps to reproduce

N/A

### Expected behavior

N/A

### Actual behavior

N/A

### URL for a reduced test case

https://codepen.io/aplefull/pen/jEqREqO?editors=0012

### HTML/SVG/etc. source for a reduced test case

```html
<script>
  console.log("abc".match(/(?:(?=(abc)))?a/));
  console.log("abc".match(/(?:(?=(abc))){0,1}a/));
</script>
```

### Log output and (if possible) backtrace

```shell
N/A
```

### Contribute a patch?

- [ ] I’ll contribute a patch for this myself.

Issue #7001, opened 2025-12-03
LibRegex: Pathological patterns cause exponential memory usage


aplefull:
### Summary

Found two patterns in Chromium's test suite that are trying to eat infinite amounts of memory and eventually cause OOM. Reductions below. [This](https://source.chromium.org/chromium/chromium/src/+/main:v8/test/mjsunit/regexp-fallback.js) is the original test for the second case if anybody's interested.

### Operating system

Linux

### Steps to reproduce

N/A

### Expected behavior

N/A

### Actual behavior

N/A

### URL for a reduced test case

https://codepen.io/aplefull/pen/pvyOKbo?editors=0012

### HTML/SVG/etc. source for a reduced test case

```html
<script>
  /(((((a?)+)+)+)+)b/.exec("a".repeat(100000));

  const regexp = new RegExp("a+".repeat(100) + "x", "g");
  const subject = `${"a".repeat(100)}x`.repeat(3);
  subject.match(regexp);
</script>
```

### Log output and (if possible) backtrace

```shell
N/A
```

### Contribute a patch?

- [ ] I’ll contribute a patch for this myself.

aplefull:
Memory is just fine in the new rust engine 🎉 
Fixed by #8612

PR #7118, opened 2025-12-13
LibRegex: Random bugfixes

aplefull:
Some random fixes I accumulated over time. See commit messages, but here are some comments, from oldest to newest commits:
1. Both Firefox and Chrome accept crazy quantifiers like `"a{99999999999999999999999999999999999999999999999999}"` and clamp them to  `2^31 - 1`, so I did the same.
2. Exposed by test262 `character-class-escape-s.js`. We advanced position past the correct match because of double inverse (`[^\S]`). First advancement was happening after the correct char class match, second at the end when temporary_inverse reset changed the `current_inversion_state()` from false to true.  
3. `unicode_aware_code_point_at()` expects code unit index, not code point. 
4. Just looking at compares was not enough to determine the correct negation state in nested character classes, so we are tracking it in parser now.
5. Parser used to backtrack into middle of Lexer's tokens like `\[`, because it was only saving character position, not token state. So we were incorrectly complaining about bracket imbalance. Bug was [found](https://discord.com/channels/1247070541085671459/1247090064480014443/1438560449028165783) by Tim. https://mozilla.github.io/pdf.js/web/viewer.html works now, so we can enjoy encoding issues there. 

Also opened #7001 and #7117  for things i couldn't fix myself 😔

Test262 diff:
        +4 ✅    -4 ❌   

Diff Tests:
```
test/built-ins/String/prototype/match/regexp-prototype-match-v-u-flag.js   ❌ -> ✅
test/built-ins/String/prototype/search/regexp-prototype-search-v-flag.js   ❌ -> ✅
test/built-ins/String/prototype/search/regexp-prototype-search-v-u-flag.js ❌ -> ✅
test/staging/sm/RegExp/character-class-escape-s.js                         ❌ -> ✅
```

### DIFF
--- a/Libraries/LibRegex/RegexParser.cpp
+++ b/Libraries/LibRegex/RegexParser.cpp
@@ -2162,7 +2162,8 @@ bool ECMA262Parser::parse_nonempty_class_ranges(Vector<CompareTypeAndValuePair>&
 
 bool ECMA262Parser::parse_class_set_expression(Vector<CompareTypeAndValuePair>& compares)
 {
-    auto start_position = tell();
+    auto start_token = m_parser_state.current_token;
+    auto start_lexer_index = m_parser_state.lexer.tell();
 
     // ClassSetExpression :: ClassUnion | ClassIntersection | ClassSubtraction
     if (parse_class_subtraction(compares)) {
@@ -2172,7 +2173,9 @@ bool ECMA262Parser::parse_class_set_expression(Vector<CompareTypeAndValuePair>& 
     if (has_error())
         return false;
 
-    back(tell() - start_position + 1);
+    m_parser_state.current_token = start_token;
+    m_parser_state.lexer.back(m_parser_state.lexer.tell() - start_lexer_index);
+
     if (parse_class_intersection(compares)) {
         consume(TokenType::RightBracket, Error::MismatchingBracket);
         return true;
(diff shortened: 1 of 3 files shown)


Libraries/LibJS/Tests/builtins/RegExp/RegExp.js
```diff
@@ -317,6 +317,7 @@ test("Unicode matching with u and v flags", () => {
         { pattern: /./gv, match: text, expected: ["𠮷", "a", "𠮷", "b", "𠮷"] },
         { pattern: /(?:)/gu, match: text, expected: ["", "", "", "", "", ""] },
         { pattern: /(?:)/gv, match: text, expected: ["", "", "", "", "", ""] },
+        { pattern: /[👨‍👩‍👧‍👦]/v, match: "𠮷a𠮷b𠮷c👨‍👩‍👧‍👦d", expected: ["👨"] },
```
alimpfard, line 321: can we get a small comment on this? I didn't see the brackets and got very confused for a few seconds

aplefull: Sure, added a comment!

alimpfard: Thanks!

aplefull:
Huh, macOS now failed twice with: `Error: Could not find Xcode version that satisfied version spec: '16.4'`

alimpfard:
Maybe swift related? Not sure

PR #7119, opened 2025-12-13
LibRegex: Variable length lookbehind via new StepBack opcodes 

mikiubo:
This is an attempt to implement variable-length lookbehind in regular expressions.
I am aware that this may be a first approach and that it is not very performant and also not fully compliant with the specifications.

I am open to suggestions, and if a different approach is preferred, I would be happy to contribute to the new implementation, as much as my (limited) capabilities allow.

Thank you.

### DIFF
--- a/Libraries/LibRegex/RegexByteCode.cpp
+++ b/Libraries/LibRegex/RegexByteCode.cpp
@@ -276,6 +276,56 @@ ALWAYS_INLINE ExecutionResult OpCode_GoBack<ByteCode>::execute(MatchInput const&
 }
 
 template<typename ByteCode>
+ALWAYS_INLINE ExecutionResult OpCode_SetStepBack<ByteCode>::execute(MatchInput const&, MatchState& state) const
+{
+    state.step_backs.append(step());
+    return ExecutionResult::Continue;
+}
+
+template<typename ByteCode>
+ALWAYS_INLINE ExecutionResult OpCode_IncStepBack<ByteCode>::execute(MatchInput const& input, MatchState& state) const
+{
+    if (state.step_backs.is_empty())
+        return ExecutionResult::Failed_ExecuteLowPrioForks;
+
+    size_t last_step_back = static_cast<size_t>(++state.step_backs.last());
+
+    if (last_step_back > state.string_position)
+        return ExecutionResult::Failed_ExecuteLowPrioForks;
+
+    reverse_string_position(state, input.view, last_step_back);
+    return ExecutionResult::Continue;
+}
+
+template<typename ByteCode>
+ALWAYS_INLINE ExecutionResult OpCode_CheckStepBack<ByteCode>::execute(MatchInput const& input, MatchState& state) const
+{
+    if (state.step_backs.is_empty())
+        return ExecutionResult::Failed_ExecuteLowPrioForks;
+    if (input.saved_positions.is_empty())
+        return ExecutionResult::Failed_ExecuteLowPrioForks;
+    // NOTE: Fail if the step-back would move before the lookbehind start.
+    if (static_cast<size_t>(state.step_backs.last()) > input.saved_positions.last())
+        return ExecutionResult::Failed_ExecuteLowPrioForks;
+
+    // NOTE: Restores the string position saved before executing a lookbehind.
+    state.string_position = input.saved_positions.last();
+    state.string_position_in_code_units = input.saved_code_unit_positions.last();
+    return ExecutionResult::Continue;
+}
+
+template<typename ByteCode>
+ALWAYS_INLINE ExecutionResult OpCode_CheckSavedPosition<ByteCode>::execute(MatchInput const& input, MatchState& state) const
+{
+    if (input.saved_positions.is_empty())
+        return ExecutionResult::Failed_ExecuteLowPrioForks;
+    if (state.string_position != input.saved_positions.last())
+        return ExecutionResult::Failed_ExecuteLowPrioForks;
+    state.step_backs.take_last();
+    return ExecutionResult::Continue;
+}
+
+template<typename ByteCode>
 ALWAYS_INLINE ExecutionResult OpCode_FailForks<ByteCode>::execute(MatchInput const& input, MatchState& state) const
 {
     input.fail_counter += state.forks_since_last_save;
@@ -498,7 +548,7 @@ ALWAYS_INLINE ExecutionResult OpCode_SaveRightCaptureGroup<ByteCode>::execute(Ma
 
     auto length = state.string_position - start_position;
 
-    if (start_position < match.column)
+    if (start_position < match.column && state.step_backs.is_empty())
         return ExecutionResult::Continue;
 
     VERIFY(start_position + length <= input.view.length_in_code_units());
--- a/Libraries/LibRegex/RegexByteCode.h
+++ b/Libraries/LibRegex/RegexByteCode.h
@@ -46,6 +46,10 @@ using ByteCodeValueType = u64;
     __ENUMERATE_OPCODE(Save)                       \
     __ENUMERATE_OPCODE(Restore)                    \
     __ENUMERATE_OPCODE(GoBack)                     \
+    __ENUMERATE_OPCODE(SetStepBack)                \
+    __ENUMERATE_OPCODE(IncStepBack)                \
+    __ENUMERATE_OPCODE(CheckStepBack)              \
+    __ENUMERATE_OPCODE(CheckSavedPosition)         \
     __ENUMERATE_OPCODE(ClearCaptureGroup)          \
     __ENUMERATE_OPCODE(Repeat)                     \
     __ENUMERATE_OPCODE(ResetRepeat)                \
@@ -499,7 +503,7 @@ public:
         NegatedLookAhead,
         NegatedLookBehind,
     };
-    void insert_bytecode_lookaround(ByteCode&& lookaround_body, LookAroundType type, size_t match_length = 0)
+    void insert_bytecode_lookaround(ByteCode&& lookaround_body, LookAroundType type, size_t match_length = 0, bool greedy_lookaround = true)
     {
         // FIXME: The save stack will grow infinitely with repeated failures
         //        as we do not discard that on failure (we don't necessarily know how many to pop with the current architecture).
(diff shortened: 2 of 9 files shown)


Libraries/LibRegex/RegexByteCode.h
```diff
@@ -813,6 +842,36 @@ class OpCode_GoBack final : public OpCode {
...
+    ALWAYS_INLINE OpCodeId opcode_id() const override { return OpCodeId::CheckStepBack; }
+    ALWAYS_INLINE size_t size() const override { return 1; }
+    ByteString arguments_string() const override { return ByteString::formatted("check step back"); }
+};
+class OpCode_CheckSavedPosition final : public OpCode {
+public:
+    ExecutionResult execute(MatchInput const& input, MatchState& state) const override;
+    ALWAYS_INLINE OpCodeId opcode_id() const override { return OpCodeId::CheckSavedPosition; }
+    ALWAYS_INLINE size_t size() const override { return 2; }
+    ALWAYS_INLINE size_t body_lenght() const { return argument(0); }
```
aplefull: Typo: length

Libraries/LibRegex/RegexMatch.h
```diff
@@ -369,6 +369,7 @@ struct MatchState {
     COWVector<Match> flat_capture_group_matches; // Vector<Vector<Match>> indexed by match index, then by capture group id; flattened for performance
     COWVector<u64> repetition_marks;
     Vector<u64, 64> checkpoints;
+    Vector<size_t> stepBacks;
```
aplefull: Should probably be snake_case

Libraries/LibRegex/RegexByteCode.cpp
```diff
@@ -260,6 +260,66 @@ ALWAYS_INLINE ExecutionResult OpCode_GoBack::execute(MatchInput const& input, Ma
...
+
+    if (state.string_position != input.saved_positions.last()) {
+        if (body_lenght() == 0) {
+            if (state.string_position + 1 != input.saved_positions.last()) {
+                return ExecutionResult::Failed_ExecuteLowPrioForks;
+            }
+        } else {
+            return ExecutionResult::Failed_ExecuteLowPrioForks;
+        }
+    }
```
aplefull: Second `if` is unreachable here

Libraries/LibRegex/RegexByteCode.cpp
```diff
@@ -338,15 +398,21 @@ ALWAYS_INLINE ExecutionResult OpCode_CheckBoundary::execute(MatchInput const& in
 {
     auto isword = [](auto ch) { return is_ascii_alphanumeric(ch) || ch == '_'; };
     auto is_word_boundary = [&] {
+        size_t prev_pos = state.string_position_in_code_units - 1;
+
+        if (state.stepBacks.size() > 0) {
+            prev_pos = state.string_position_in_code_units + 1;
+        }
```
aplefull: I don't quite understand the logic, but we can go out of bounds because of +1, right?
Here, `/(?<=\b)\b/` on 'ab' for example:
`string_position_in_code_units` will go from 0 to 1, then we'll set `prev_pos` to 1 + 1 and later attempt to `input.view.code_point_at(prev_pos)` on input with a length of 2

Libraries/LibRegex/RegexByteCode.h
```diff
@@ -460,13 +464,27 @@ class REGEX_API ByteCode : public DisjointChunks<ByteCodeValueType> {
...
-            // REGEXP BODY
+            // SET_STEPBACK match_length(BODY)-1
+            // LABEL _START
+            // INC_STEPBACK
+            // STEPBACK
+            // FORK_JUMP _BODY
+            // CHECK_STEPBACK
+            // JUMP _START
+            // LABEL BODY
+            // REGEX
```
alimpfard: This doesn't match what's below.

Libraries/LibRegex/RegexByteCode.h
```diff
@@ -460,13 +464,27 @@ class REGEX_API ByteCode : public DisjointChunks<ByteCodeValueType> {
...
+            empend((ByteCodeValueType)OpCodeId::SetStepBack);
+            empend((ByteCodeValueType)match_length - 1);
+            empend((ByteCodeValueType)OpCodeId::IncStepBack);
+            empend((ByteCodeValueType)OpCodeId::ForkJump);
+            empend((ByteCodeValueType)1 + 2); // JUMP to label _BODY
+            empend((ByteCodeValueType)OpCodeId::CheckStepBack);
+            empend((ByteCodeValueType)OpCodeId::Jump);
+            empend((ByteCodeValueType)-6); // JUMP to label _START
             extend(move(lookaround_body));
+            empend((ByteCodeValueType)OpCodeId::CheckSavedPosition);
```
alimpfard, line 573: so we start at minimum length behind (-1+1), then run the assertion - failing which takes us back one more character to retry the assertion?

If so, what's the point of check step back? inc_stepback can't step back behind the start of the subject, so it should just fail no?

Libraries/LibRegex/RegexByteCode.h
```diff
@@ -460,13 +464,27 @@ class REGEX_API ByteCode : public DisjointChunks<ByteCodeValueType> {
...
+            // LABEL BODY
+            // REGEX
             // RESTORE
             empend((ByteCodeValueType)OpCodeId::Save);
-            empend((ByteCodeValueType)OpCodeId::GoBack);
-            empend((ByteCodeValueType)match_length);
+            empend((ByteCodeValueType)OpCodeId::SetStepBack);
+            empend((ByteCodeValueType)match_length - 1);
+            empend((ByteCodeValueType)OpCodeId::IncStepBack);
+            empend((ByteCodeValueType)OpCodeId::ForkJump);
```
alimpfard, line 563: This can probably be a ForkReplaceJump, I don't think backtracking across this would ever produce a match?

Libraries/LibRegex/RegexByteCode.h
mikiubo: > so we start at minimum length behind (-1+1), then run the assertion - failing which takes us back one more character to retry the assertion?
> 
> If so, what's the point of check step back? inc_stepback can't step back behind the start of the subject, so it should just fail no?

check_step_back is a function whose main responsibility is to restore the position to the last saved one, because we cannot reliably know whether executing the lookbehind body will leave the cursor before or after the originally saved position.

I named the operation check_step_back because additional checks could be added there in the future to improve performance. For example, if the cursor ends up before the saved position, the match could be considered failed. Unfortunately, I haven’t managed to implement this optimization yet.

To be more transparent, I could rename the operation or split these two responsibilities into separate steps?

alimpfard: :thinking: wouldn't inc_stepback know what the "correct" next index is regardless?
I get the idea though, just a comment would suffice I think.

Libraries/LibRegex/RegexByteCode.cpp
mikiubo: You’re right, this was a naive attempt of mine to solve the WPT:
`
assert_array_equals("ab cdef".match(/(?<=\B)\w{3}/), ["def"], "#2");`


My idea was that if stepback is not empty, it means we’re evaluating a lookbehind.

In that case, we can evaluate a word boundary at the end, meaning it’s enough that the next character is a word character and the previous one is not.

However, I hadn’t actually thought about your case — I need to reason about it more carefully.

i will notify you if i find a solution.

mikiubo: Ok i think i find a more stable solution i also split the commit to be more clear

Libraries/LibRegex/RegexByteCode.cpp
```diff
@@ -299,9 +299,21 @@ ALWAYS_INLINE ExecutionResult OpCode_CheckStepBack::execute(MatchInput const& in
 
 ALWAYS_INLINE ExecutionResult OpCode_CheckSavedPosition::execute(MatchInput const& input, MatchState& state) const
 {
-    if (state.string_position != input.saved_positions.last())
+    if (input.saved_positions.size() == 0) {
```
alimpfard: nit:
```suggestion
    if (input.saved_positions.is_empty()) {
```

Libraries/LibRegex/RegexByteCode.h
```diff
@@ -463,6 +463,12 @@ class REGEX_API ByteCode : public DisjointChunks<ByteCodeValueType> {
             return;
         }
         case LookAroundType::LookBehind:
+            // CheckBoundary already compares the previous and current positions,
+            // so applying the generic lookbehind stepback would rewind twice.
+            if (lookaround_body.size() == 2 && lookaround_body[0] == (ByteCodeValueType)OpCodeId::CheckBoundary) {
```
alimpfard: this feels like a footgun,
- what if the body has more than a single checkboundary op?
- does nothing else hit this?
- checkboundary shouldn't be changing positions though?

Libraries/LibRegex/RegexByteCode.h
```diff
@@ -463,6 +463,10 @@ class REGEX_API ByteCode : public DisjointChunks<ByteCodeValueType> {
             return;
         }
         case LookAroundType::LookBehind: {
+            if (lookaround_body.size() == 1 && (lookaround_body[0] == (ByteCodeValueType)OpCodeId::CheckBegin || lookaround_body[0] == (ByteCodeValueType)OpCodeId::CheckEnd)) {
```
alimpfard: ...I'm not sure I like this.
Is the problem that these assertions don't modify the position, so your stepback jumps back to one past the expected char?
Instead, how about:
```
    checkpoint c
    BODY
    jumpnonempty c &revert
    jump &continue
revert:
    stepback ...
continue:
    ...
```
?

Libraries/LibRegex/RegexByteCode.h
mikiubo: i tried but it breaks this:
assert_array_equals("ab12b23b34c".match(/(?<=((?:b\d{2})+))c/), ["c", "b12b23b34"], "3");
i analyzed why yet for now i add this wpt to the test explicitely.

Libraries/LibRegex/RegexByteCode.cpp
mikiubo: i think i fixed
 thanks

Libraries/LibRegex/RegexByteCode.h
alimpfard: Ah so I guess we _can_ produce a match if we backtrack across that.
Maybe we should just leave that to the optimiser to figure out.

Libraries/LibRegex/RegexByteCode.h
mikiubo: Ok this look very promising approach!
i will check how does checkpoint opcode works.
i think i will have to do some refactor.
i thnk maybe with this approach we can also fix the problem on the word boundary check.

 i will try to implement it 
 i wil notify you as soon as i do some changes

Libraries/LibRegex/RegexByteCode.h
mikiubo: ok i think i fixed
Thanks

Libraries/LibRegex/RegexByteCode.h
mikiubo: While refactoring, I realized that the code could be greatly simplified by allowing the step_back vector to handle negative numbers.
Changing its type from size_t to i64 allowed me to remove problematic if statements and reduce the number of commits from five to two.

Libraries/LibRegex/RegexByteCode.h
mikiubo: with my last refactor i think also this is fixed

Libraries/LibRegex/RegexByteCode.h
mikiubo: Add comment

alimpfard (commented):
Alright, changes LGTM, but please squash the second commit into the first since it's just an incremental fix.

mikiubo:
> Alright, changes LGTM, but please squash the second commit into the first since it's just an incremental fix.

Oki i think i merged the two commit

alimpfard:
Thanks!

PR #7126, opened 2025-12-13
LibWeb: Update svg type for root element checks when set/get title

bingyuanng:
Document::title() and Document::set_title() now only runs SVG-specific title logic when the document element is a lowercase svg. Fixing this [WPT](https://wpt.fyi/results/html/dom/documents/dom-tree-accessors/document.title-09.html?label=master&product=chrome)

### DIFF
--- a/Libraries/LibWeb/DOM/Document.cpp
+++ b/Libraries/LibWeb/DOM/Document.cpp
@@ -1034,7 +1034,7 @@ Utf16String Document::title() const
 
     // 1. If the document element is an SVG svg element, then let value be the child text content of the first SVG title
     //    element that is a child of the document element.
-    if (auto const* document_element = this->document_element(); is<SVG::SVGElement>(document_element)) {
+    if (auto const* document_element = this->document_element(); is<SVG::SVGSVGElement>(document_element)) {
         if (auto const* title_element = document_element->first_child_of_type<SVG::SVGTitleElement>())
             value = title_element->child_text_content();
     }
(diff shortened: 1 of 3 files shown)


Psychpsyo:
None of the code that you added seems to adhere to what the spec says. If that's intentional and this is the correct solution to this, you should probably add a comment about that and file a bug with the spec.
If it isn't (and I suspect it isn't), this bug is probably caused by something elsewhere in the code.

Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -1034,7 +1034,8 @@ Utf16String Document::title() const
 
     // 1. If the document element is an SVG svg element, then let value be the child text content of the first SVG title
     //    element that is a child of the document element.
-    if (auto const* document_element = this->document_element(); is<SVG::SVGElement>(document_element)) {
```
tcl3, line 1037: AFAICT, the issue here is that we're checking whether the given element is an `SVGElement`, where we should be checking whether it is an `SVGSVGElement`. I think all that is required is to change the existing `is<T>` calls to refer to the correct type.

bingyuanng: Just tested it, you are right. Can i know whats the difference between these 2 ? I'll make a change and update the descriptions

konradekk: description is all right; here’s your reading:
* https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Element
* https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Element/svg

bingyuanng:
Updated with the suggested change. Thanks for the quick review.

tcl3 (approved):
Thanks for working on this!

PR #7134, opened 2025-12-15
CMake: Remove ENABLE_WINDOWS_CI option and adjust presets to match Unix

ayeteadoe:
Inspired by a comment in https://github.com/LadybirdBrowser/ladybird/pull/6807, see commit for details.

### DIFF
--- a/Documentation/BuildInstructionsLadybird.md
+++ b/Documentation/BuildInstructionsLadybird.md
@@ -172,8 +172,8 @@ brew install qt
 
 ### Windows:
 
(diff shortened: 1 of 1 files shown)


ayeteadoe:
It looks like the pending/required check for Windows is because of https://discord.com/channels/1247070541085671459/1256703242679746560/1451228031875158218, so the preset field in the required check would need updating along with this change

ADKaster (approved):
This looks good to me. At some point there will be a need to remove the Experimental in the preset name, but given current plans that'll be ages from now. I'll update the required checks, but merging this will have an impact on other PRs until they're rebased.

PR #7137, opened 2025-12-15
LibJS: Add context to object operation error messages

ghost:
Previously, errors like "Object's [[Set]] method returned false" gave no indication of which property or object was involved. This made debugging difficult.

Error messages now include the property name and object type, e.g.:

"Cannot set property 'foo' on [object Object]"
"Cannot define property 'Symbol(bar)' on [object Array]" "Cannot delete property '0' from [object Object]"
"Cannot prevent extensions on [object Object]"
"Cannot set prototype of [object Object]"

I of course also updated the existing tests to match new error message format.

as this is my first PR, please let me know if I have done anything incorrectly. I will acknowledge and fix it asap.

### DIFF
--- a/Libraries/LibJS/Bytecode/Interpreter.cpp
+++ b/Libraries/LibJS/Bytecode/Interpreter.cpp
@@ -2030,7 +2030,7 @@ ThrowCompletionOr<void> SetGlobal::execute_impl(Bytecode::Interpreter& interpret
                     return vm.throw_completion<TypeError>(ErrorType::DescWriteNonWritable, identifier);
                 }
             }
-            return vm.throw_completion<TypeError>(ErrorType::ObjectSetReturnedFalse);
+            return vm.throw_completion<TypeError>(ErrorType::ObjectSetReturnedFalse, identifier, Value(&binding_object));
         }
         if (cacheable_metadata.type == CacheableSetPropertyMetadata::Type::ChangeOwnProperty) {
             cache.entries[0].shape = shape;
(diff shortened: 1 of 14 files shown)


Libraries/LibJS/Bytecode/Interpreter.cpp
```diff
@@ -2030,7 +2030,7 @@ ThrowCompletionOr<void> SetGlobal::execute_impl(Bytecode::Interpreter& interpret
                     return vm.throw_completion<TypeError>(ErrorType::DescWriteNonWritable, identifier);
                 }
             }
-            return vm.throw_completion<TypeError>(ErrorType::ObjectSetReturnedFalse);
+            return vm.throw_completion<TypeError>(ErrorType::ObjectSetReturnedFalse, identifier, Value(&binding_object).to_utf16_string_without_side_effects());
```
trflynn89: Here and other places below: You shouldn't need to call `to_utf16_string_without_side_effects`, that will happen automatically by the formatter.
```c++
return vm.throw_completion<TypeError>(ErrorType::ObjectSetReturnedFalse, identifier, Value { &binding_object });
```

ghost: Much appreciated!

ghost: You can go a step further, correct?

ghost:
Seems I have accidentally included a piece of additional code that was planned to be my next feature (fdivided_by).

Although this helps me to see I was working in the incorrect folder, it messes with the PR, so I do apologize.

I will push this with the additional code removed ASAP.

PR #7146, opened 2025-12-15
CI: Add Linux arm64 build to ci.yml

gmta:
We already ran the distribution preset build in our nightly action, and this adds it to the jobs we run for PRs as well.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,6 +38,13 @@ jobs:
             runner_labels: '["macos-15", "self-hosted"]'
 
           - os_name: 'Linux'
+            arch: 'arm64'
+            build_preset: 'Sanitizer'
+            toolchain: 'Clang'
+            clang_plugins: false
+            runner_labels: '["blacksmith-16vcpu-ubuntu-2404-arm"]'
+
+          - os_name: 'Linux'
             arch: 'x86_64'
             build_preset: 'Fuzzers'
             toolchain: 'Clang'


rcorsi:
@gmta I compared the Nightly Linux arm64 Clang Sanitizer job vs the checks for this PR for the Linux arm64 Clang Sanitizer part.

Main thing that stands out is the use of clang_plugins set to true in this PR, but the nightly has it false.  Maybe set it to false.

Also the blacksmith runner with 16vcpu (CI) vs the 8vcpu (Nightly).  Maybe try the 8vcpu to see if it does anything. It shouldn't, but you never know.

I guess those are two things to change and see if it resolves this weird libdrm issue.

I would guess it is the clang_plugins.

rcorsi:
I duplicated your PR in #7151.  Hope that is ok. I had no idea how to test any theory otherwise, I don't have a decent arm64 system available.

rcorsi:
In PR #7151 I tested the idea that the clang_plugins set to true or the blacksmith runner vcpu count might be the cause of the libdrm not found problem.

From the testing I did, the clang_plugins set to true is the source of the problem.  Set it to false to resolve the problem. 

Not sure exactly why it is so, but one thing I noticed is that the cache might be involved where the key was

`Sanitizer-CLANG_PLUGINS`

Here is an example from the log...
```
echo "ccache_key=Sanitizer-CLANG_PLUGINS" >> "$GITHUB_OUTPUT
```

Both the v8cpu and v16cpu blacksmith runners appear to work well for arm.

Also I had at least three nearly successful runs and the issue was in LibWeb with always the same test failing.
```
Fail: Screenshot/input/canvas-pattern-transform.html
```

and the error is always.

```
 2] 5632/5664: Finish Screenshot/input/svg-tefile:///home/runner/_work/ladybird/ladybird/Tests/LibWeb/Screenshot/input/canvas-pattern-transform.html: Fuzzy mismatch: pixel error count 53145 is outside FuzzyRange [0-2396]
```

The `error count 53145` is same for three runs.

gmta:
@rcorsi 

> Main thing that stands out is the use of clang_plugins set to true in this PR, but the nightly has it false. Maybe set it to false.

Well for PRs we'd want to run with clang plugins, which is why we should turn that on.

> Also the blacksmith runner with 16vcpu (CI) vs the 8vcpu (Nightly). Maybe try the 8vcpu to see if it does anything. It shouldn't, but you never know.

Apart from resource constraints (disk, memory) tied to the runner type, that should not make a difference. I've selected less performant runners for the nightly jobs since those can take more time.

gmta:
@rcorsi 

> Not sure exactly why it is so, but one thing I noticed is that the cache might be involved where the key was
>
> Sanitizer-CLANG_PLUGINS

There's no cache for arm64 with plugins enabled yet, so it's simply working from a clean slate. The build is failing so we should fix that instead:

```
CMake Error at /usr/local/share/cmake-3.31/Modules/FindPkgConfig.cmake:645 (message):
  The following required packages were not found:

   - libdrm
```

> Also I had at least three nearly successful runs and the issue was in LibWeb with always the same test failing.
>
> `Fail: Screenshot/input/canvas-pattern-transform.html`

Once we get a proper build running, we can download the test artifacts that are uploaded as a result of a CI run to see how the screenshots differ exactly and work from there.

gmta:
> Well for PRs we'd want to run with clang plugins, which is why we should turn that on.

Coming back to my own remark, we only want to do this for a single CI build, not for all 😅 so let's disable them while keeping the fix for pkg-config in place.

PR #7154, opened 2025-12-16
CI: Move Windows build to ci.yml using new Blacksmith runners

gmta:


.github/workflows/ci.yml
```diff
@@ -51,6 +51,13 @@ jobs:
             clang_plugins: false
             runner_labels: '["blacksmith-16vcpu-ubuntu-2404"]'
 
+          - os_name: 'Windows'
+            arch: 'x86_64'
+            build_preset: 'Windows_Sanitizer_CI'
+            toolchain: 'Clang'
```
ayeteadoe: Probably should be `ClangCL` here

gmta: The only place I saw "Clang" being matched against was in steps that were explicitly not executed for Windows. The toolchain is never compared to "ClangCL" as far as I can see.

gmta: That's not to say "ClangCL" isn't a better description for the toolchain, I was mostly testing whether or not that had an impact at all since all other LLVM configurations used "Clang".

.github/actions/setup/action.yml
```diff
@@ -54,6 +54,17 @@ runs:
         sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-14 100
         sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-14 100
 
+    - name: Ensure clang-cl is available
+      if: ${{ inputs.os == 'Windows' && inputs.toolchain == 'ClangCL' }}
+      shell: bash
+      run: |
+        LLVMBIN="C:/Program Files/LLVM/bin"
+        if ! where.exe clang-cl > /dev/null 2>&1 ; then
+          choco install -y llvm --no-progress --version=21.1.7
```
ayeteadoe: The `test-js` errors seemed odd as I wasn't able to reproduce locally, but I noticed we're installing LLVM 21.1.7 here.

My local Visual Studio 2026 Insiders build is shipped with LLVM 20.1.8, and the GIthub Actions `windows-2025` images are also using LLVM 20.1.8: https://github.com/actions/runner-images/blob/main/images/windows/Windows2025-Readme.md

I was able to locally reproduce when I re-compiled `test-js` with LLVM 21.1.7. So if we change this to `choco install -y llvm --no-progress --version=20.1.8` our CI should be happy.

gmta: Interesting to see that we're failing with 21.1.7, but definitely - let's keep this version in sync with GitHub's Windows image.

.github/workflows/nightly-lagom.yml
```diff
@@ -66,7 +66,7 @@ jobs:
             build_preset: 'Windows_Sanitizer_CI'
             toolchain: 'ClangCL'
             clang_plugins: false
-            runner_labels: '["windows-2025"]'
+            runner_labels: '["blacksmith-8vcpu-windows-2025"]'
```
trflynn89, line 69: With Windows now in the normal CI flow, is there any reason to keep the nightly job?

gmta: Well yeah, the nightly job verifies periodic builds regardless of whether a PR or push to master happened? That's why we include the other jobs (Linux, macOS, etc.) as well in the nightly.

trflynn89: The other nightly jobs are there because they aren't in normal CI - distribution presets and swift. 

(Looks like linux arm64 is still there as well, but I would say that can also be removed now that it is in normal CI)

gmta: Right - I usually set up nightly jobs with the same config as PR CI in order to validate some things that change over time instead of based on new commits: security issues, 3rd party server availability, flakes, upstream issues that we can fail on, etc.

Might need some more discussion :)

PR #7156, opened 2025-12-16
CI: Support wasm-tools download for Windows and arm64

gmta:
We don't use wasm-tools there yet, but let's at least not do weird things like download the aarch64 macOS version on Windows x86_64.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -21,6 +21,7 @@ runs:
       with:
         python-version: 3.12
         cache: 'pip'
+
     - name: 'Install Python dependencies'
       shell: bash
       run: pip install pyyaml requests six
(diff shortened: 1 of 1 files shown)


.github/actions/setup/action.yml
```diff
@@ -116,13 +117,16 @@ runs:
...
         else
-          NAME="wasm-tools-$VERSION-aarch64-macos"
+          ARCH=${{ inputs.arch }}
         fi
+        LOWERCASE_OS=$( echo -n '${{ inputs.os }}' | tr '[:upper:]' '[:lower:]')
+        NAME="wasm-tools-${VERSION}-${ARCH}-${LOWERCASE_OS}"
 
-        curl -f -L -o "${NAME}.tar.gz" "https://github.com/bytecodealliance/wasm-tools/releases/download/v${VERSION}/${NAME}.tar.gz"
+        curl --fail --location --write-out '%{url}' --output "${NAME}.tar.gz" \
+          "https://github.com/bytecodealliance/wasm-tools/releases/download/v${VERSION}/${NAME}.tar.gz"
```
trflynn89: Looks like the release is a `.zip` on Windows:
https://github.com/bytecodealliance/wasm-tools/releases/download/v1.243.0/wasm-tools-1.243.0-x86_64-windows.zip

gmta: Yeah a fix was already under way!

PR #7157, opened 2025-12-16
Clamp discrepancy

tete17:
The C `round` differs from the web standard if the value lies exactly at
the middle point. C rounds away from 0 while the web moves to the even
value of the 2 extremes.

This fixes at least 5 tests :)

### DIFF
--- a/Libraries/LibWeb/WebIDL/AbstractOperations.cpp
+++ b/Libraries/LibWeb/WebIDL/AbstractOperations.cpp
@@ -517,7 +517,7 @@ JS::ThrowCompletionOr<T> convert_to_int(JS::VM& vm, JS::Value value, EnforceRang
 
         // 2. Round x to the nearest integer, choosing the even integer if it lies halfway between two, and choosing +0 rather than −0.
         // 3. Return x.
-        return round(x);
+        return AK::round_to<T>(x);
     }
 
     // 8. If x is NaN, +0, +∞, or −∞, then return +0.
--- a/Tests/LibWeb/Text/expected/wpt-import/FileAPI/blob/Blob-slice.any.txt
+++ b/Tests/LibWeb/Text/expected/wpt-import/FileAPI/blob/Blob-slice.any.txt
@@ -2,8 +2,7 @@ Harness status: OK
 
 Found 150 tests
 
-145 Pass
-5 Fail
+150 Pass
 Pass	no-argument Blob slice
 Pass	Slices
 Pass	blob1.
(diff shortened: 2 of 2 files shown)


Libraries/LibWeb/WebIDL/AbstractOperations.cpp
```diff
@@ -517,6 +517,14 @@ JS::ThrowCompletionOr<T> convert_to_int(JS::VM& vm, JS::Value value, EnforceRang
 
         // 2. Round x to the nearest integer, choosing the even integer if it lies halfway between two, and choosing +0 rather than −0.
         // 3. Return x.
+        // NOTE: The C function lround matches the web standard except for the middle halfway case.
+        //       C rounds away from 0. We need to correct for that particular case.
+        if (ceil(x) - x == 0.5L) {
+            if (lround(x) % 2 == 0)
+                return round(x);
+            return round(x) - 1;
+        }
```
gmta: Isn't rounding to even integers in the halfway case a typical supported rounding mode? E.g. `std::fesetround` with `FE_TONEAREST`.

We tend not to use the standard namespace though, however we might want to introduce something for this so we can use our AK's `round_to<T>` together with a specified rounding mode?

Hendiadyoin1: `TO_NEAREST`, aka break tie to even, is the default
the C99 functions for that are `rint` or `nearbyint`, the former should have an AK wrapper, with maybe a slightly different name
(`round_to` follows `lrint` semantics IIRC)

Edit: The difference is the handling of floating point exceptions, but both technically follow the currently set rounding mode, which we never touch and you would need to tell the compiler if you did

gmta: Right, so it's more a question of whether we'd want a thin AK wrapper of the different sets of standard rounding methods?

Hendiadyoin1: Wel should already have all standard rounding modes wrapped,
If I understood the commit correctly `round_to<int>`/`lrint` should be correct here (or the `i64`/`ll` versions)

Edit: `round_to<T>` as it seems

tete17: Thank you @Hendiadyoin1 I wasn't aware of this round_to<T> function :)

I guess with this WPT we fixed the behaviour to be like the Web Standard

PR #7158, opened 2025-12-16
LibWeb/HTML: Fill-in and update scrollParent and offsetParent

AtkinsSJ:
I'm uncertain about a couple of things here, but it does at least pass some more WPT tests.

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLElement.cpp
@@ -510,51 +510,63 @@ GC::Ptr<DOM::Element> HTMLElement::scroll_parent() const
     return nullptr;
 }
 
-// https://www.w3.org/TR/cssom-view-1/#dom-htmlelement-offsetparent
+// https://drafts.csswg.org/cssom-view-1/#dom-htmlelement-offsetparent
 GC::Ptr<DOM::Element> HTMLElement::offset_parent() const
 {
+    // NOTE: We have to ensure that the layout is up-to-date before querying the layout tree.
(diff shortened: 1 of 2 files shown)


Libraries/LibWeb/Layout/FormattingContext.cpp
```diff
@@ -1195,7 +1195,8 @@ CSSPixelRect FormattingContext::content_box_rect_in_static_position_ancestor_coo
     }
     // If we get here, `ancestor_box` was not in the containing block chain of the static position containing block of `box`!
     // Something about the containing block chain is set up incorrectly then.
-    VERIFY_NOT_REACHED();
+    rect.set_location({ 0, 0 });
+    return rect;
```
AtkinsSJ: @InvalidUsernameException @ayeteadoe @gmta You've all recently touched this method. Is this reasonable? As I posted [on Discord](https://discord.com/channels/1247070541085671459/1247090064480014443/1450493581432455331), assigning fixed-positioning containing blocks causes the loop here to miss the containing block, at least in some cases. I'm not familiar enough with layout to be able to tell if that's something that should happen, or if we're doing something wrong in `static_position_containing_block()`, or we're not assigning containing blocks correctly for anonymous wrappers. (Or if my fixed positioning CB code is just bad and wrong. :sweat_smile: )

AtkinsSJ:
Oh wait, I think #6719 probably fixes the issue I was working around. Drafting this for now.

AtkinsSJ:
Changes: Dropped the hack as it's no longer necessary with #6719 merged.

AtkinsSJ:
Changes: Rebased to enable the Windows CI job, and fix a small invisible conflict.

PR #7161, opened 2025-12-16
LibWeb: Use HashMap for string to CSS value lookups

tcl3:
This replaces multiple string comparisons with a single hash lookup for various CSS value types.


gmta:
Did all of these have a measurable impact on performance?

tcl3:
While this did give me an improvement in the original case I found for dimension units. I've written a more comprehensive stress test that shows the gains for all of these changes are marginal at best and slower in the case of failed lookups.

Closing since I'm not convinced any of these changes are worth keeping.

PR #7169, opened 2025-12-17
LibMedia: Try to use FFmpeg for all formats other than Matroska

Zaggy1024:
By sniffing specifically for MP4 and WebM, we were precluding PlaybackManager from playing any other formats. Instead, use MatroskaDemuxer if the media has a `matroska` or `webm` EBML doctype, and fall back to FFmpeg for all others.

We'll need to limit the containers that FFmpeg is able to open at some point, but for now, this allows us to play the formats we could before.


gmta:
@Zaggy1024 CI failure:

```
UNEXPECTED ERROR: [DecoderError]: Header -> [operator() @ /home/runner/_work/ladybird/ladybird/Libraries/LibMedia/Containers/Matroska/Reader.cpp:197]: DocTypeVersion was 0 at /home/runner/_work/ladybird/ladybird/Tests/LibMedia/TestMediaCommon.h:27
```

PR #7171, opened 2025-12-18
LibWeb: Implement Navigator.maxTouchPoints

joshuar500:
Hello all, I am a frontend engineer trying to learn more about the browser + love the mission + want to learn more C++. I am using LLM, but hoping things are small enough so that I understand what I'm doing. Hopefully this is helpful first PR.

---

Implement the maxTouchPoints attribute on Navigator as specified in
the W3C Pointer Events specification. The implementation queries the
PageClient for the maximum number of simultaneous touch contacts
supported by the device.

Platform clients can override PageClient::max_touch_points() to return
the actual device capability. The default implementation returns 0.

Added a FIXME for future WebDriver BiDi emulated max touch points
support, which will allow overriding the value during automation.

### DIFF
--- a/Libraries/LibWeb/HTML/Navigator.cpp
+++ b/Libraries/LibWeb/HTML/Navigator.cpp
@@ -129,10 +129,22 @@ GC::Ref<CredentialManagement::CredentialsContainer> Navigator::credentials()
 }
 
 // https://w3c.github.io/pointerevents/#dom-navigator-maxtouchpoints
-WebIDL::Long Navigator::max_touch_points()
+WebIDL::Long Navigator::max_touch_points() const
 {
-    dbgln("FIXME: Unimplemented Navigator.maxTouchPoints");
-    return 0;
+    // 1. Let emulated maxTouchPoints be the result of WebDriver BiDi emulated max touch points.
+    // FIXME: WebDriver BiDi is not yet implemented.
+    //        https://www.w3.org/TR/webdriver-bidi/#webdriver-bidi-emulated-max-touch-points
+
+    // 2. If emulated maxTouchPoints is not null, return emulated maxTouchPoints.
+    // FIXME: Pseudo code for when WebDriver BiDi is implemented.
+    // if (auto emulated = get_webdriver_bidi_emulated_max_touch_points(); emulated.has_value())
+    //     return emulated.value();
+
+    // 3. Return the maximum number of simultaneous touch contacts supported by the device.
+    //    In the case of devices with multiple digitizers (e.g. multiple touchscreens), the value
+    //    must be the maximum of the set of maximum supported contacts by each individual digitizer.
+    auto& window = as<HTML::Window>(HTML::relevant_global_object(*this));
+    return window.page().client().max_touch_points();
 }
 
 GC::Ref<ServiceWorker::ServiceWorkerContainer> Navigator::service_worker()
--- a/Libraries/LibWeb/HTML/TraversableNavigable.h
+++ b/Libraries/LibWeb/HTML/TraversableNavigable.h
@@ -116,6 +116,11 @@ public:
     Geolocation::EmulatedPositionData const& emulated_position_data() const;
     void set_emulated_position_data(Geolocation::EmulatedPositionData data);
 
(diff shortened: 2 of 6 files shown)


Libraries/LibWeb/HTML/TraversableNavigable.h
```diff
@@ -116,6 +116,10 @@ class WEB_API TraversableNavigable final : public Navigable {
     Geolocation::EmulatedPositionData const& emulated_position_data() const;
     void set_emulated_position_data(Geolocation::EmulatedPositionData data);
 
+    // https://www.w3.org/TR/webdriver-bidi/#webdriver-bidi-emulated-max-touch-points
+    Optional<WebIDL::Long> emulated_max_touch_points() const;
+    void set_emulated_max_touch_points(Optional<WebIDL::Long> value);
```
gmta: Looks like the input emulated max number of touch points is defined as an unsigned integer that's greater than or equal to 1, so maybe you can make this an `UnsignedLong` instead so you can get rid of the `max(0, ..)` as well?

Libraries/LibWeb/HTML/Navigator.cpp
```diff
@@ -129,10 +131,30 @@ GC::Ref<CredentialManagement::CredentialsContainer> Navigator::credentials()
 }
 
 // https://w3c.github.io/pointerevents/#dom-navigator-maxtouchpoints
-WebIDL::Long Navigator::max_touch_points()
-{
-    dbgln("FIXME: Unimplemented Navigator.maxTouchPoints");
-    return 0;
+WebIDL::Long Navigator::max_touch_points() const
+{
+    // 1. Let window be this's relevant global object.
```
trflynn89: Where does the text for these spec steps come from? They do not match the link above:
https://w3c.github.io/pointerevents/#dom-navigator-maxtouchpoints

Libraries/LibWeb/HTML/TraversableNavigable.cpp
```diff
@@ -1423,6 +1423,20 @@ void TraversableNavigable::set_emulated_position_data(Geolocation::EmulatedPosit
     m_emulated_position_data = data;
 }
 
+// https://www.w3.org/TR/webdriver-bidi/#webdriver-bidi-emulated-max-touch-points
+Optional<WebIDL::Long> TraversableNavigable::emulated_max_touch_points() const
+{
+    VERIFY(is_top_level_traversable());
```
trflynn89: The emulated max touch points also has very specific spec text that is not implemented here. It won't be a getter/setter on the traversable, it will have to reach into webdriver session objects.

IMO it'd be better to just leave the emulated part as a FIXME. We haven't started a webdriver bidi implementation at all yet, so we don't know what this is going to look like.

joshuar500: good to know, thanks for context

Libraries/LibWeb/HTML/Navigator.cpp
joshuar500: bit of combo slop from https://www.w3.org/TR/webdriver-bidi/#webdriver-bidi-emulated-max-touch-points and the link above. will update and align.

Libraries/LibWeb/HTML/TraversableNavigable.h
joshuar500: added FIXME based on other comment

Libraries/LibWeb/Page/Page.h
```diff
@@ -427,6 +428,12 @@ class PageClient : public JS::Cell {
 
     virtual bool is_svg_page_client() const { return false; }
 
+    // https://w3c.github.io/pointerevents/#dom-navigator-maxtouchpoints
+    // FIXME: Platform clients should override this to return the actual maximum touch points
+    //        supported by the device. For devices with multiple digitizers, this should be
+    //        the maximum across all digitizers.
+    virtual WebIDL::Long max_touch_points() const { return 0; }
```
joshuar500, line 435: should this (and above) still be unsigned long?

gmta: Well this method is not a part of the IDL spec as far as I know, so this can be whatever type we want it to be. You could make this a `u32` for example.

PR #7173, opened 2025-12-18
LibGfx: Improve Vulkan device selection logic

LeadRDRK:
(My first PR here, please tell me if I did anything wrong.)

Ladybird was unacceptably slow on my Linux machine, and I noticed that it was caused by vkCreateDevice failing due to missing extensions. I have 2 relatively modern AMD GPUs, so I figured that they should've had all of the extensions required. Turns out it was because Ladybird was picking the wrong device; I had both the open source and the proprietary drivers installed, Ladybird picked a Vulkan device that used the proprietary driver, which did not have `VK_EXT_image_drm_format_modifier`.

This aims to prevent that from happening. Instead of blindly selecting the first device or the last discrete GPU, give each device a score, based on the device type and the extensions they support, and pick the one with the highest score. Priority is given to discrete GPUs and integrated GPUs, with discrete GPUs being the most prioritized (leaving virtual GPUs and CPUs with a lower score).

The scores I gave to the 2 extensions required for Vulkan images are arbitrary. In case Ladybird depends on more extensions later on, these scores could be adjusted and new ones could be added for better device selection.

### DIFF
--- a/Libraries/LibGfx/VulkanContext.cpp
+++ b/Libraries/LibGfx/VulkanContext.cpp
@@ -150,7 +150,9 @@ static ErrorOr<VkDevice> create_logical_device(VkPhysicalDevice physical_device,
     create_device_info.enabledExtensionCount = device_extension_count;
     create_device_info.ppEnabledExtensionNames = device_extensions;
 
-    if (vkCreateDevice(physical_device, &create_device_info, nullptr, &device) != VK_SUCCESS) {
+    VkResult result = vkCreateDevice(physical_device, &create_device_info, nullptr, &device);
+    if (result != VK_SUCCESS) {
+        dbgln("vkCreateDevice returned {}", to_underlying(result));
         return Error::from_string_literal("vkCreateDevice failed");
     }
 


Libraries/LibGfx/VulkanContext.cpp
```diff
@@ -47,16 +48,58 @@ static ErrorOr<VkPhysicalDevice> pick_physical_device(VkInstance instance)
...
+
+        int score = 0;
+
+        // Prefer discrete GPUs but also give priority to integrated GPUs
         if (device_properties.deviceType == VK_PHYSICAL_DEVICE_TYPE_DISCRETE_GPU)
+            score += 1000;
+        else if (device_properties.deviceType == VK_PHYSICAL_DEVICE_TYPE_INTEGRATED_GPU)
+            score += 100;
+
+        uint32_t extension_count = 0;
```
gmta: We usually use the `u32` type instead.

Libraries/LibGfx/VulkanContext.cpp
```diff
@@ -47,16 +48,58 @@ static ErrorOr<VkPhysicalDevice> pick_physical_device(VkInstance instance)
...
+            auto properties = extension_properties[i];
+            StringView name(properties.extensionName, strlen(properties.extensionName));
+
+            auto extension_score = extension_scores.get(name);
+            if (extension_score.has_value()) {
+                score += extension_score.value();
+            }
+        }
+
+        if (score > picked_score) {
```
gmta: `picked_score` initializes to `0` and technically a GPU does not need to be physical or integrated (there's also the "other", "virtual" or "CPU" types), so maybe initialize it to -1 so we always pick a GPU if it's available?

You can then get rid of this above:

```cpp
        if (picked_device == VK_NULL_HANDLE)
            picked_device = device;
```

Libraries/LibGfx/VulkanContext.cpp
```diff
@@ -47,16 +48,58 @@ static ErrorOr<VkPhysicalDevice> pick_physical_device(VkInstance instance)
...
+        // Prefer discrete GPUs but also give priority to integrated GPUs
         if (device_properties.deviceType == VK_PHYSICAL_DEVICE_TYPE_DISCRETE_GPU)
+            score += 1000;
+        else if (device_properties.deviceType == VK_PHYSICAL_DEVICE_TYPE_INTEGRATED_GPU)
+            score += 100;
+
+        uint32_t extension_count = 0;
+        vkEnumerateDeviceExtensionProperties(device, nullptr, &extension_count, nullptr);
+        if (extension_count > extension_properties.size())
+            extension_properties.resize(extension_count);
```
gmta: Instead of doing `if (extension_count > extension_properties.size())`, you can also use `extension_properties.resize_and_keep_capacity(extension_count)` to prevent any reallocation if the size is reduced for this device.

Libraries/LibGfx/VulkanContext.cpp
```diff
@@ -47,16 +48,58 @@ static ErrorOr<VkPhysicalDevice> pick_physical_device(VkInstance instance)
...
+
+        vkEnumerateDeviceExtensionProperties(device, nullptr, &extension_count, extension_properties.data());
+        for (size_t i = 0; i < extension_count; ++i) {
+            auto properties = extension_properties[i];
+            StringView name(properties.extensionName, strlen(properties.extensionName));
+
+            auto extension_score = extension_scores.get(name);
+            if (extension_score.has_value()) {
+                score += extension_score.value();
+            }
```
gmta: Suggestion:

```cpp
score += extension_scores.get(name).value_or(0);
```

gmta:
@LeadRDRK also: welcome! 🥳

gmta:
@LeadRDRK please change the original commits instead of adding new commits to your branch. You can do so by [rebasing](https://github.com/LadybirdBrowser/ladybird/blob/master/Documentation/GettingStartedContributing.md#using-git) your branch and incorporating your changes in the right commits.

Also, feel free to resolve discussions yourself if you think you dealt with them.

LeadRDRK:
@gmta Done, please check again.

Libraries/LibGfx/VulkanContext.cpp
```diff
@@ -47,16 +48,52 @@ static ErrorOr<VkPhysicalDevice> pick_physical_device(VkInstance instance)
...
+
+    static HashMap<StringView, int> extension_scores {
+#ifdef USE_VULKAN_IMAGES
+        __EXTENSION_SCORE(VK_KHR_EXTERNAL_MEMORY_FD_EXTENSION_NAME, 10),
+        __EXTENSION_SCORE(VK_EXT_IMAGE_DRM_FORMAT_MODIFIER_EXTENSION_NAME, 10),
+#endif
+    };
+
+#undef __EXTENSION_SCORE
+#undef __EXTENSION_SCORE_HELPER
```
konradekk: the following should work too:

```suggestion
    constexpr HashMap<StringView, int>::Entry extension_score(const char* name, int value) {
        return { name""sv, value };
    }

    static HashMap<StringView, int> extension_scores {
#ifdef USE_VULKAN_IMAGES
        extension_score(VK_KHR_EXTERNAL_MEMORY_FD_EXTENSION_NAME, 10),
        extension_score(VK_EXT_IMAGE_DRM_FORMAT_MODIFIER_EXTENSION_NAME, 10),
#endif
    };
```

LeadRDRK:
> (note: you could also keep the scores and try next device in the list if creating of current one fails for any reason. but this might be treated as well as an extension etc., ignore this hint at will.)

Might as well have a go at it, so I did :)

Libraries/LibGfx/VulkanContext.cpp
LeadRDRK: I adopted your suggestion directly without using a constexpr.

Libraries/LibGfx/VulkanContext.cpp
```diff
@@ -35,34 +37,63 @@ static ErrorOr<VkInstance> create_instance(uint32_t api_version)
     return instance;
 }
 
-static ErrorOr<VkPhysicalDevice> pick_physical_device(VkInstance instance)
+struct RankedVkPhysicalDeviceEntry {
+    VkPhysicalDevice device;
+    int score;
+};
+
+static ErrorOr<void> get_ranked_physical_device_list(VkInstance instance, Vector<RankedVkPhysicalDeviceEntry>& ranked_devices)
```
gmta: Instead of passing in a vector, you can return one (e.g. `ErrorOr<Vector<..>>`

Libraries/LibGfx/VulkanContext.cpp
```diff
@@ -165,10 +196,27 @@ ErrorOr<VulkanContext> create_vulkan_context()
...
+    VkPhysicalDevice physical_device = nullptr;
+    for (auto const& entry : ranked_physical_devices) {
+        auto result = create_logical_device(entry.device, &graphics_queue_family);
+        if (!result.is_error()) {
+            logical_device = result.value();
+            physical_device = entry.device;
+            break;
+        }
+
+        dbgln("{}", result.error());
```
gmta: Maybe prefix the error with something like "Failed creating logical device:"

LeadRDRK: `create_logical_device()` could return an error that says "Logical device creation failed", so it would become "Failed creating logical device: Logical device creation failed", which is a bit repetitive

LeadRDRK: I could change it to "vkCreateDevice failed", sounds good?

gmta: Errors should generally go from more specific to less specific the further an error unwinds, so that sounds good :)

gmta (approved):

Libraries/LibGfx/VulkanContext.cpp
```diff
@@ -47,22 +54,48 @@ static ErrorOr<VkPhysicalDevice> pick_physical_device(VkInstance instance)
...
-        VkPhysicalDeviceProperties device_properties;
+    VkPhysicalDeviceProperties device_properties;
+    Vector<VkExtensionProperties> extension_properties;
+    // Score devices based on their type and feature support
+    for (auto const& device : devices) {
         vkGetPhysicalDeviceProperties(device, &device_properties);
+
+        int score = 0;
+
+        // Prefer discrete GPUs but also give priority to integrated GPUs
```
Zaggy1024, line 75: I would actually normally expect integrated GPUs to be preferred, thinking specifically of laptops with dual GPUs, since the discrete GPU usually has a much higher power draw. I wonder if the scores should be reduced and reversed so that extensions have more influence than this if they affect compatibility?

LeadRDRK: I've thought about this before, ultimately it comes down to personal preference, but from an application standpoint the discrete GPU makes more sense since it offers higher performance. Perhaps this could be a configurable option later on.

As for the extensions, I was thinking that Vulkan images could eventually be made optional during runtime (I might even follow up with a PR that does that), so that's why I haven't considered penalizing devices for missing them yet. If there are any required extensions later on, maybe the device should be removed from the device list completely.

gmta: I agree that the big `FIXME` here should be to make the preferred GPU configurable. Since we're currently aiming for performance more than we're trying to optimize for battery lifetime, I think it's fine to leave it as-is for now.

konradekk:
this was kind of nice to continue with.

PR #7174, opened 2025-12-18
LibWeb: Some animation bug fixes

Calme1709:
A collection of a few bug fixes around animations that I came across while working on implementing progress-based timelines - see individual commits for details

### DIFF
--- a/Libraries/LibWeb/Animations/Animation.cpp
+++ b/Libraries/LibWeb/Animations/Animation.cpp
@@ -748,10 +748,7 @@ WebIDL::ExceptionOr<void> Animation::play_an_animation(AutoRewind auto_rewind)
     //
     //     If a user agent determines that animation is immediately ready, it may schedule the above task as a microtask
     //     such that it runs at the next microtask checkpoint, but it must not perform the task synchronously.
-    // FIXME: Below actions should only happen when the animation is actually ready.
     m_pending_play_task = TaskState::Scheduled;
-    if (m_timeline)
-        m_saved_play_time = m_timeline->current_time();
 
     // 13. Run the procedure to update an animation’s finished state for animation with the did seek flag set to false,
     //     and the synchronously notify flag set to false.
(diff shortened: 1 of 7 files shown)


Calme1709:
Removed imported `finishing-an-animation.html` test which proved to be flaky

Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -5335,30 +5335,32 @@ void Document::append_pending_animation_event(Web::DOM::Document::PendingAnimati
...
-    HTML::perform_a_microtask_checkpoint();
+        // 3. Perform a microtask checkpoint.
+        // NB: This is executed by the destructor of the TemporaryExecutionContext above.
+    }
 
     // 4. Let events to dispatch be a copy of doc’s pending animation event queue.
     auto events_to_dispatch = GC::ConservativeVector<Document::PendingAnimationEvent> { vm().heap() };
     events_to_dispatch.extend(m_pending_animation_event_queue);
 
-    // 5. Clear doc’s pending animation event queue.
```
kalenikaliaksandr, line 5370: was this comment deleted by accident?

Libraries/LibWeb/Animations/Animation.cpp
```diff
@@ -1193,13 +1193,12 @@ void Animation::update_finished_state(DidSeek did_seek, SynchronouslyNotify sync
     //    steps:
     if (current_finished_state && !m_is_finished) {
         // 1. Let finish notification steps refer to the following procedure:
-        auto finish_notification_steps = GC::create_function(heap(), [this, &realm]() {
+        auto finish_notification_steps = [this, &realm]() {
```
kalenikaliaksandr: shouldn't we keep `finish_notification_steps` as `GC::Function`? otherwise there's no guarantee `this` is alive by the time HTML task that calls this function is executed.

Libraries/LibWeb/Animations/AnimationTimeline.cpp
```diff
@@ -89,6 +89,8 @@ void AnimationTimeline::visit_edges(Cell::Visitor& visitor)
 {
     Base::visit_edges(visitor);
     visitor.visit(m_associated_document);
+    // FIXME: We shouldn't keep animations (and thus their associated owner elements, effects, etc) alive just because
+    //        they're associated with a timeline.
```
kalenikaliaksandr: instead of reintroducing the leak could we replace `HashTable<GC::Ref<Animation>> m_associated_animations {};` with `HashTable<GC::Weak<Animation>> m_associated_animations {};` and change `remove_animations_from_timeline()` to remove null pointers from the hashtable.

Libraries/LibWeb/DOM/Document.cpp
Calme1709: Correct - have added it back in.

Libraries/LibWeb/Animations/Animation.cpp
Calme1709: Done.

Libraries/LibWeb/Animations/AnimationTimeline.cpp
Calme1709: Please correct me if I am misunderstanding something but I don't think it's necessary to change this to be `GC::Weak<Animation>`, just to remove `m_associated_animations` from `visit_edges()` and disassociate animations from timelines in `Animation::finalize()`, which I have done.

Calme1709: Our clang plugin doesn't like that - have updated to be `GC::Weak`

PR #7176, opened 2025-12-18
LibWeb/CSS: Make accentColor and accentColorText take accent-color value

### LINKED ISSUE #7141: CSS `AccentColor` keyword should use the value of `accent-color`
Spec change: https://github.com/w3c/csswg-drafts/pull/12733

Instead of `AccentColor` always being the system's accent color, it should use whatever the `accent-color` property is set to, if anything, and then fall back to the system accent color.

Likewise, `AccentColorText` needs to take the `accent-color` into account.

[Spec:](https://drafts.csswg.org/css-color-4/#ref-for-accent-color)

> To maintain consistency with widget [accent color](https://drafts.csswg.org/css-ui-4/#accent-color) styling, [AccentColor](https://drafts.csswg.org/css-color-4/#valdef-color-accentcolor) takes its value from [accent-color](https://drafts.csswg.org/css-ui-4/#propdef-accent-color), unless [Forced Colors Mode](https://drafts.csswg.org/css-color-adjust-1/#forced-colors-mode) is enabled. [AccentColorText](https://drafts.csswg.org/css-color-4/#valdef-color-accentcolortext) takes its value from the contrasting foreground color to AccentColor as is described for widget accent color styling.

`KeywordStyleValue::to_color()` is where we get a color from these keywords, so that needs changing. Probably the `ColorResolutionContext` needs to know more, either what element it's for or just the `accent-color` value.

arakelyanva:
Fixes #7141.

https://github.com/user-attachments/assets/3b260d2a-0740-4225-939d-f6fce76cd62e

### DIFF
--- a/Libraries/LibWeb/CSS/StyleValues/KeywordStyleValue.cpp
+++ b/Libraries/LibWeb/CSS/StyleValues/KeywordStyleValue.cpp
@@ -144,14 +144,29 @@ Optional<Color> KeywordStyleValue::to_color(ColorResolutionContext color_resolut
 
     PreferredColorScheme scheme = color_resolution_context.color_scheme.value_or(PreferredColorScheme::Light);
 
+    // Calculate accent_color_text based on contrast to accent_color
+    if (keyword() == Keyword::Accentcolortext) {
+        // min_contrast = 10.2 is a magic number which provides the best accessibility trade-off based on:
+        // 1. https://webaim.org/resources/contrastchecker/
+        // 2. Current implementation of luminosity() and contrast_ratio() methods for Color instances
+
+        // the baseline colors with the least contrast from black and white are #757575 and #767676
+        // which score over 4.5 ratio for #fff and #000 accent_color_text values correspondingly
+        auto constexpr min_contrast = 10.2;
+        auto system_accent_text = SystemColor::accent_color_text(scheme);
+
+        if (color_resolution_context.accent_color.value_or(SystemColor::accent_color(scheme)).contrast_ratio(system_accent_text) < min_contrast)
+            return system_accent_text.inverted();
+
+        return system_accent_text;
+    }
+
     // First, handle <system-color>s, since they don't strictly require a node.
     // https://www.w3.org/TR/css-color-4/#css-system-colors
     // https://www.w3.org/TR/css-color-4/#deprecated-system-colors
     switch (keyword()) {
     case Keyword::Accentcolor:
-        return SystemColor::accent_color(scheme);
-    case Keyword::Accentcolortext:
-        return SystemColor::accent_color_text(scheme);
+        return color_resolution_context.accent_color.value_or(SystemColor::accent_color(scheme));
     case Keyword::Buttonborder:
     case Keyword::Activeborder:
     case Keyword::Inactiveborder:
--- a/Libraries/LibWeb/CSS/StyleValues/StyleValue.cpp
+++ b/Libraries/LibWeb/CSS/StyleValues/StyleValue.cpp
@@ -81,6 +81,7 @@
 #include <LibWeb/CSS/StyleValues/UnicodeRangeStyleValue.h>
 #include <LibWeb/CSS/StyleValues/UnresolvedStyleValue.h>
 #include <LibWeb/CSS/StyleValues/ViewFunctionStyleValue.h>
+#include <LibWeb/CSS/SystemColor.h>
 #include <LibWeb/HTML/Navigable.h>
 #include <LibWeb/Layout/Node.h>
 
@@ -94,7 +95,8 @@ ColorResolutionContext ColorResolutionContext::for_element(DOM::AbstractElement 
 
     return {
         .color_scheme = color_scheme,
-        .current_color = element.computed_properties()->color_or_fallback(PropertyID::Color, { color_scheme, CSS::InitialValues::color(), element.document(), calculation_resolution_context }, CSS::InitialValues::color()),
+        .current_color = element.computed_properties()->color_or_fallback(PropertyID::Color, { color_scheme, CSS::InitialValues::color(), CSS::SystemColor::accent_color(color_scheme), element.document(), calculation_resolution_context }, CSS::InitialValues::color()),
+        .accent_color = element.computed_properties()->color_or_fallback(PropertyID::AccentColor, { color_scheme, CSS::InitialValues::color(), CSS::SystemColor::accent_color(color_scheme), element.document(), calculation_resolution_context }, CSS::SystemColor::accent_color(color_scheme)),
         .document = element.document(),
         .calculation_resolution_context = calculation_resolution_context
     };
@@ -105,6 +107,7 @@ ColorResolutionContext ColorResolutionContext::for_layout_node_with_style(Layout
     return {
         .color_scheme = layout_node.computed_values().color_scheme(),
         .current_color = layout_node.computed_values().color(),
+        .accent_color = layout_node.computed_values().accent_color(),
         .document = layout_node.document(),
         .calculation_resolution_context = { .length_resolution_context = Length::ResolutionContext::for_layout_node(layout_node) },
     };
--- a/Libraries/LibWeb/Layout/Node.cpp
+++ b/Libraries/LibWeb/Layout/Node.cpp
@@ -20,6 +20,7 @@
 #include <LibWeb/CSS/StyleValues/StyleValueList.h>
 #include <LibWeb/CSS/StyleValues/TimeStyleValue.h>
 #include <LibWeb/CSS/StyleValues/URLStyleValue.h>
+#include <LibWeb/CSS/SystemColor.h>
 #include <LibWeb/DOM/Document.h>
 #include <LibWeb/Dump.h>
 #include <LibWeb/HTML/FormAssociatedElement.h>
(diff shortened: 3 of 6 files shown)


Libraries/LibWeb/CSS/ComputedValues.h
```diff
@@ -181,6 +181,7 @@ class InitialValues {
     static CSSPixels text_underline_offset() { return 2; }
     static TextUnderlinePosition text_underline_position() { return { .horizontal = TextUnderlinePositionHorizontal::Auto, .vertical = TextUnderlinePositionVertical::Auto }; }
     static Display display() { return Display { DisplayOutside::Inline, DisplayInside::Flow }; }
+    static Color accent_color() { return Color(61, 174, 233); } // SystemColor::accent_color value
```
AtkinsSJ: The changes to ComputedValues.h aren't actually needed - we can leave the default as `{}` as before, which is how we represent `auto`.

Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
```diff
@@ -515,6 +515,8 @@ Parser::ParseErrorOr<NonnullRefPtr<StyleValue const>> Parser::parse_css_value(Pr
         //       values, only the CSS-wide keywords - this is handled above, and thus, if we have gotten to here, there
         //       is an invalid value which is a syntax error.
         return ParseError::SyntaxError;
+    case PropertyID::AccentColor:
+        return parse_all_as(tokens, [this](auto& tokens) { return parse_accent_color(tokens); });
```
AtkinsSJ: The parsing changes also aren't needed, it's simple enough for the parser to do automatically. (For what it's worth, your parsing code is good, it just isn't needed.)

Libraries/LibWeb/CSS/StyleValues/StyleValue.cpp
```diff
@@ -92,7 +92,8 @@ ColorResolutionContext ColorResolutionContext::for_element(DOM::AbstractElement
 
     return {
         .color_scheme = color_scheme,
-        .current_color = element.computed_properties()->color_or_fallback(PropertyID::Color, { color_scheme, CSS::InitialValues::color(), element.document(), calculation_resolution_context }, CSS::InitialValues::color()),
+        .current_color = element.computed_properties()->color_or_fallback(PropertyID::Color, { color_scheme, CSS::InitialValues::color(), CSS::InitialValues::accent_color(), element.document(), calculation_resolution_context }, CSS::InitialValues::color()),
+        .accent_color = element.computed_properties()->color_or_fallback(PropertyID::AccentColor, { color_scheme, CSS::InitialValues::color(), CSS::InitialValues::accent_color(), element.document(), calculation_resolution_context }, CSS::InitialValues::accent_color()),
```
AtkinsSJ: ...then, instead of using `InitialValues::accent_color()`, the default should be `SystemColor::accent_color(color_scheme)`.

Libraries/LibWeb/Layout/Node.cpp
```diff
@@ -414,7 +414,8 @@ void NodeWithStyle::apply_style(CSS::ComputedProperties const& computed_style)
     // NOTE: color must be set after color-scheme to ensure currentColor can be resolved in other properties (e.g. background-color).
     // NOTE: color must be set after font_size as `CalculatedStyleValue`s can rely on it being set for resolving lengths.
     computed_values.set_color(computed_style.color_or_fallback(CSS::PropertyID::Color, CSS::ColorResolutionContext::for_layout_node_with_style(*this), CSS::InitialValues::color()));
-
+    // NOTE: Currently there are still discussions about `accentColor` and `currentColor` interactions, so the line below might need changing in the future
+    computed_values.set_accent_color(computed_style.color_or_fallback(CSS::PropertyID::AccentColor, CSS::ColorResolutionContext::for_layout_node_with_style(*this), CSS::InitialValues::accent_color()));
```
AtkinsSJ: Same point about InitialValues::accent_color(), but you'll have to get the `color_scheme` from `computed_values`.

Libraries/LibWeb/CSS/Parser/Parser.h
```diff
@@ -426,6 +426,7 @@ class Parser {
     RefPtr<StyleValue const> parse_opacity_value(TokenStream<ComponentValue>&);
     RefPtr<FontSourceStyleValue const> parse_font_source_value(TokenStream<ComponentValue>&);
 
+    RefPtr<StyleValue const> parse_accent_color(TokenStream<ComponentValue>& tokens);
```
AtkinsSJ: Also, the commit message talks about parsing. Really this change is about making `AccentColor` us the current `accent-color`, so say something about that.

AtkinsSJ:
Oh also, you can put `Fixes #number` in a line in the PR description to make github close that issue automatically when this is merged.

Libraries/LibWeb/CSS/ComputedValues.h
arakelyanva: Done

Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
arakelyanva: Done

Libraries/LibWeb/CSS/StyleValues/StyleValue.cpp
arakelyanva: Done

Libraries/LibWeb/Layout/Node.cpp
arakelyanva: Done

Libraries/LibWeb/CSS/Parser/Parser.h
arakelyanva: Done

arakelyanva:
@AtkinsSJ all requested changes are applied. Simple `text` test is also added and tested.
screenshot-2025-12-18_20-14-15

Libraries/LibWeb/CSS/StyleValues/KeywordStyleValue.cpp
```diff
@@ -144,14 +144,23 @@ Optional<Color> KeywordStyleValue::to_color(ColorResolutionContext color_resolut
 
     PreferredColorScheme scheme = color_resolution_context.color_scheme.value_or(PreferredColorScheme::Light);
 
+    if (keyword() == Keyword::Accentcolortext) {
+        // Calculate accent_color_text based on contrast to accent_color
+        auto constexpr min_contrast = 2;
+        auto system_accent_text = SystemColor::accent_color_text(scheme);
+
+        if (color_resolution_context.accent_color.value_or({}).contrast_ratio(system_accent_text) < min_contrast)
```
AtkinsSJ: The `value_or({})` means it's assuming the accent color is transparent black if it's not passed in. It should be `color_resolution_context.accent_color.value_or(SystemColor::accent_color(scheme))` like in the code below.

arakelyanva: Done

arakelyanva:
> The `Screenshot/input/color-scheme.html` test is failing, because its value for `accentcolortext` has changed from white to black. It's not clear if that's a bug from the line I commented on, or it's actually correct new behaviour. If it's correct, please update the test expectation so it passes.

@AtkinsSJ Yep, this was because of the `accentColorText` calculation based on contrast with `accent-color` value. I've changed the screenshot with the updated version, as discussed in Discord `SystemColor::accent_color_text = Color(0, 0, 0)` is more appropriate according to WCAG contrast standards. Now all tests are passing on my local machine:

screenshot-2026-01-13_00-22-20

Libraries/LibWeb/CSS/ComputedValues.h
```diff
@@ -726,7 +727,7 @@ class ComputedValues {
         ColorInterpolation color_interpolation { InitialValues::color_interpolation() };
 
         PreferredColorScheme color_scheme { InitialValues::color_scheme() };
-        Optional<Color> accent_color {};
+        Optional<Color> accent_color { InitialValues::accent_color() };
```
AtkinsSJ: The changes to ComputedValues.h shouldn't be needed. This can default to an empty Optional, like it did before.

arakelyanva: Done

Libraries/LibWeb/CSS/ComputedValues.h
```diff
@@ -180,6 +180,7 @@ class InitialValues {
     static CSSPixels text_underline_offset() { return 2; }
     static TextUnderlinePosition text_underline_position() { return { .horizontal = TextUnderlinePositionHorizontal::Auto, .vertical = TextUnderlinePositionVertical::Auto }; }
     static Display display() { return Display { DisplayOutside::Inline, DisplayInside::Flow }; }
+    static Color accent_color() { return {}; }
```
AtkinsSJ: I meant this line too, sorry. Nothing uses it.

Libraries/LibWeb/CSS/SystemColor.cpp
```diff
@@ -15,7 +15,7 @@ Color accent_color(PreferredColorScheme)
 
 Color accent_color_text(PreferredColorScheme)
 {
-    return Color(255, 255, 255);
+    return Color(0, 0, 0);
```
AtkinsSJ, line 18: I believe this commit needs to come before the other commit. Otherwise if someone runs the test from the other commit, it will fail.

Libraries/LibWeb/CSS/ComputedValues.h
arakelyanva: That's on me, sorry, forgot to remove that as well.

Libraries/LibWeb/CSS/SystemColor.cpp
arakelyanva: Yep, replaced the commit sequence.

arakelyanva:
Just FYI fixed the linter comment regarding whitespace left on an empty newline.

CI should be fine now.

AtkinsSJ (approved):
All looks good! Thanks for persevering with this.

PR #7184, opened 2025-12-18
LibWeb/HTML: Return Promises from scroll methods

AtkinsSJ:
I had to add TemporaryExecutionContexts to a couple of places here. I'm not convinced I've chosen good places though so feedback on that would be helpful.

There's a lot of scrolling code that doesn't match the spec well, but I didn't want to get too deep into that right now.


nico:
(Caused a regression being fixed in https://github.com/LadybirdBrowser/ladybird/pull/7226)

PR #7185, opened 2025-12-18
Meta: Build All_Debug macros in release mode

rcorsi:
Originally was building All_Debug macros in debug, but that is not ideal as debug build will build both the release and debug builds, taking longer and using a lot of disk space.

Building in release as that is enough to find the issues with all the debug macros.

See the following for discussion about 'debug' build problems

https://github.com/microsoft/vcpkg/issues/38224

I found that link inside Ladybird's own debug.cmake file:

https://github.com/LadybirdBrowser/ladybird/blob/master/Meta/CMake/vcpkg/debug-triplets/debug.cmake

### DIFF
--- a/Meta/CMake/presets/CMakeBasePresets.json
+++ b/Meta/CMake/presets/CMakeBasePresets.json
@@ -65,8 +65,8 @@
       "description": "All Debug build",
       "binaryDir": "$env{LADYBIRD_SOURCE_DIR}/Build/alldebug",
       "cacheVariables": {
-        "CMAKE_BUILD_TYPE": "Debug",
-        "VCPKG_OVERLAY_TRIPLETS": "$env{LADYBIRD_SOURCE_DIR}/Meta/CMake/vcpkg/debug-triplets",
+        "CMAKE_BUILD_TYPE": "RelWithDebInfo",
+        "VCPKG_OVERLAY_TRIPLETS": "$env{LADYBIRD_SOURCE_DIR}/Meta/CMake/vcpkg/release-triplets",
         "ENABLE_ALL_THE_DEBUG_MACROS": "ON"
       }
     },


rcorsi:
When building 3rd party libraries (e.g. harfbuzz, google-highway, libtommath, sdl3, libyuv, etc...) both the debug and release builds are processed.

For example, you can see the following when building libavif, where both debug and release messages are visible.

```
-- Configuring x64-linux-dynamic
-- Building x64-linux-dynamic-dbg
-- Building x64-linux-dynamic-rel
-- Fixing pkgconfig file: /home/bobo/projects/myladybird/Build/vcpkg/packages/libavif_x64-linux-dynamic/lib/pkgconfig/libavif.pc
-- Fixing pkgconfig file: /home/bobo/projects/myladybird/Build/vcpkg/packages/libavif_x64-linux-dynamic/debug/lib/pkgconfig/libavif.pc
-- Installing: /home/bobo/projects/myladybird/Build/vcpkg/packages/libavif_x64-linux-dynamic/share/libavif/copyright
-- Adjusted RPATH of '/home/bobo/projects/myladybird/Build/vcpkg/packages/libavif_x64-linux-dynamic/debug/lib/libavif.so.16.3.0' (From '' -> To '$ORIGIN')
-- Adjusted RPATH of '/home/bobo/projects/myladybird/Build/vcpkg/packages/libavif_x64-linux-dynamic/lib/libavif.so.16.3.0' (From '' -> To '$ORIGIN')

```

rcorsi:
Building All_Debug preset on Ubuntu 25.04 (GCC 14.2) results in the Build/alldebug directory taking around

38GB with 'debug' build type, but switching to 'release' build type the disk usage is 1.7GB.

Don't know if this is important or an issue on the CI runners, but locally well it could make a difference especially if you use VMs with smaller disks and you happen to build with ENABLE_ALL_THE_DEBUG_MACROS. If you have the Build/alldebug directory (from the All_Debug build) and Build/debug (from the Debug build) taking a total of 76GB (380GB+38GB) that could lead to disk full situations.  Switching to a release build for All_Debug would help for disk usage reduction.

Again, this is probably not important for most developers, unless they concurrently use the Debug and All_Debug build type on a regular basis.

rcorsi:
Hi @ADKaster, can this be merged or should I abandon this idea?

PR #7187, opened 2025-12-19
LibGC: Various block allocator improvements + stats dump

awesomekling:
See individual commits.


gmta:
Looks good when testing manually, but CI is fed up 💢

PR #7188, opened 2025-12-19
Meta: Fix edge cases in EOF Detection

Totto16:
There were some edge cases in an empty file or a file with just one line of spaces, these are now handled explicitly instead of crashing the script

The script previously crashed with this message, on empty files:

```bash
Traceback (most recent call last):
  File "<whatever>/ladybird/Meta/check-newlines-at-eof.py", line 78, in <module>
    run()
    ~~~^^
  File "<whatever>/ladybird/Meta/check-newlines-at-eof.py", line 51, in run
    f.seek(f.tell() - 1, os.SEEK_SET)
    ~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: negative seek position -1

```

### DIFF
--- a/Meta/check-newlines-at-eof.py
+++ b/Meta/check-newlines-at-eof.py
@@ -41,35 +41,27 @@ def find_files_here_or_argv():
 def run():
     """Check files checked in to git for trailing newlines at end of file."""
     no_newline_at_eof_errors = []
-    blank_lines_at_eof_errors = []
+    whitespace_at_eof_errors = []
 
-    did_fail = False
     for filename in find_files_here_or_argv():
-        with open(filename, "r") as f:
-            f.seek(0, os.SEEK_END)
+        with open(filename, "r", newline="") as f:
+            content = f.read()
 
-            f.seek(f.tell() - 1, os.SEEK_SET)
-            if f.read(1) != "\n":
-                did_fail = True
-                no_newline_at_eof_errors.append(filename)
-                continue
+        if not content.endswith("\n"):
+            no_newline_at_eof_errors.append(filename)
+            continue
 
-            while True:
-                f.seek(f.tell() - 2, os.SEEK_SET)
-                char = f.read(1)
-                if not char.isspace():
-                    break
-                if char == "\n":
-                    did_fail = True
-                    blank_lines_at_eof_errors.append(filename)
-                    break
+        stripped = content.rstrip()
+        trailing_whitespace = len(content) - len(stripped)
+        if trailing_whitespace > 1:
+            whitespace_at_eof_errors.append(filename)
 
     if no_newline_at_eof_errors:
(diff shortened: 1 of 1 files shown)


Meta/check-newlines-at-eof.py
```diff
@@ -48,14 +48,29 @@ def run():
...
+            if no_newline_local:
                 did_fail = True
                 no_newline_at_eof_errors.append(filename)
                 continue
 
+            # note: here we have implicitly seeked to the end of the file
+
             while True:
-                f.seek(f.tell() - 2, os.SEEK_SET)
+                current_pos = f.tell()
```
shannonbooth: Thanks, this definitely fixes that issue, but personally find the new logic even more difficult to follow around the file pointer for the 'blank lines' case. I had an attempt of writing a version that just reads the entire file rather than the logic of seeking around. I got something like:

```
def run():
    """Check files checked in to git for trailing newlines at end of file."""
    no_newline_at_eof_errors = []
    whitespace_at_eof_errors = []

    for filename in find_files_here_or_argv():
        with open(filename, "r", newline='') as f:
            content = f.read()

        # Check for missing trailing newline
        if not content.endswith("\n"):
            no_newline_at_eof_errors.append(filename)
            continue

        # Check for extra whitespace at EOF
        # Strip whitespace to find first non-whitespace content
        stripped = content.rstrip()
        trailing_whitespace = len(content) - len(stripped)
            
        # If there are 2+ trailing whitespace characters, that means extra whitespace at EOF
        if trailing_whitespace > 1:
            whitespace_at_eof_errors.append(filename)

    if no_newline_at_eof_errors:
        print("Files with no newline at the end:", " ".join(no_newline_at_eof_errors))
    if whitespace_at_eof_errors:
        print("Files that have whitespace at the end:", " ".join(whitespace_at_eof_errors))

    if no_newline_at_eof_errors or whitespace_at_eof_errors:
        sys.exit(1)
```

Not that performance is too big a concern, but this actually ended up being marginally faster on my machine also. Maybe something along those lines would be an improvement? There are probably other edge cases neither catch that might get changed in the future, like extra spaces on the newline

Totto16: Yeah I also thought that my solution makes the logic harder, but I am not the best python developer, so I don't know the best approach to this 🤷🏼‍♂️  

I would say, if you find your solution better (and it fixes the problem), a separate PR would be a better Idea, as I don't want to just copy your code and claim it as mine per commit, or you could make this PR contain only your commit, with your solution, as maintainer should be allowed to force push to this PR branch. 

Anyway, feel free to make my code better, but copying your code over would feel wrong to me, if it is small snippets or improvements is fine, but you made the whole algorithm, so you should make the commit 😉

shannonbooth: I don't think it's just my own - up to you - I just added us both

PR #7193, opened 2025-12-19
Meta: Correctly bootstrap vcpkg on musl libc systems

cqundefine:


Toolchain/BuildVcpkg.py
```diff
@@ -40,7 +41,12 @@ def build_vcpkg():
     subprocess.check_call(args=["git", "checkout", git_rev], cwd=vcpkg_checkout)
 
     bootstrap_script = "bootstrap-vcpkg.bat" if os.name == "nt" else "bootstrap-vcpkg.sh"
-    subprocess.check_call(args=[vcpkg_checkout / bootstrap_script, "-disableMetrics"], cwd=vcpkg_checkout)
+    arguments = [vcpkg_checkout / bootstrap_script, "-disableMetrics"]
+    libc, _ = platform.libc_ver()
```
ADKaster, line 45: This method seems to only exist on unix and only gives useful answers on gcc-compiled Python. 

https://docs.python.org/3/library/platform.html#unix-platforms

That being said, it's a better guess than not on musl systems like Alpine. 

It should be wrapped in some "what if Windows" protection though. A try except pass should (?) do the trick

cqundefine: In Python 3.14.2 on Windows 11, the function does in fact exists, it does return two empty strings and that doesn't seem to be an issue.

ADKaster: fair enough. kind of the python devs then.

PR #6406, opened 2025-10-06
Meta: Add Flatpak linting for vcpkg version equality

JanKoudijs:
The Flatpak manifest linter matches dependencies against vcpkg.json and if there is a match checks if versions are equal. If there is no match it checks if there is a listed exclusion. If there is no version match or no listed exclusion then linting fails.

Merging this PR means that vcpkg.json updates have a failing CI if the Flatpak manifest isn't updated accordingly, there has to be awareness of the Flatpak existence.

PR #7203, opened 2025-12-20
Meta: Set SSL_CERT_DIR in Flatpak manifest

JanKoudijs:
In #6406 the Flatpak build is fixed and linting is added. As part of the changes the openssl dependency is now built from source, but at runtime the certificates cannot be found.

This PR sets the `SSL_CERT_DIR` environment variable in the Flatpak sandbox, solving the certificate issue.


JanKoudijs:
Can a maintainer add the `flatpak` label to this PR?

JanKoudijs:
Current nightlies can be used like this:

    flatpak run --user --env=SSL_CERT_DIR=/etc/ssl/certs org.ladybird.Ladybird

PR #7206, opened 2025-12-20
Libweb: Add resizing to textarea controls

jonbgamble:
image

Also fixes https://github.com/LadybirdBrowser/ladybird/issues/6920

### DIFF
--- a/Libraries/LibGfx/Rect.h
+++ b/Libraries/LibGfx/Rect.h
@@ -334,6 +334,8 @@ public:
     ALWAYS_INLINE void set_primary_offset_for_orientation(Orientation orientation, T value) { m_location.set_primary_offset_for_orientation(orientation, value); }
     [[nodiscard]] ALWAYS_INLINE T secondary_offset_for_orientation(Orientation orientation) const { return m_location.secondary_offset_for_orientation(orientation); }
     ALWAYS_INLINE void set_secondary_offset_for_orientation(Orientation orientation, T value) { m_location.set_secondary_offset_for_orientation(orientation, value); }
+    ALWAYS_INLINE void translate_primary_offset_for_orientation(Orientation orientation, T delta) { m_location.set_primary_offset_for_orientation(orientation, m_location.primary_offset_for_orientation(orientation) + delta); }
+    ALWAYS_INLINE void translate_secondary_offset_for_orientation(Orientation orientation, T delta) { m_location.set_secondary_offset_for_orientation(orientation, m_location.secondary_offset_for_orientation(orientation) + delta); }
 
     [[nodiscard]] ALWAYS_INLINE T primary_size_for_orientation(Orientation orientation) const { return m_size.primary_size_for_orientation(orientation); }
     [[nodiscard]] ALWAYS_INLINE T secondary_size_for_orientation(Orientation orientation) const { return m_size.secondary_size_for_orientation(orientation); }
--- a/Libraries/LibWeb/Page/ElementResizeAction.cpp
+++ b/Libraries/LibWeb/Page/ElementResizeAction.cpp
@@ -1,0 +1,104 @@
+/*
+ * Copyright (c) 2025, Jonathan Gamble <gamblej@gmail.com>
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include <LibWeb/CSS/PropertyID.h>
+#include <LibWeb/DOM/Element.h>
+#include <LibWeb/Page/ElementResizeAction.h>
+#include <LibWeb/Painting/ChromeMetrics.h>
+#include <LibWeb/Painting/PaintableBox.h>
+
+// https://drafts.csswg.org/css-ui#resize
+
+namespace Web {
+
+static Optional<CSSPixelSize> containing_block_padding_box_size(Layout::Node const& layout_node)
+{
+    auto parent_box = layout_node.containing_block();
+    if (!parent_box)
+        return {};
+    if (auto const* paintable_box = as_if<Painting::PaintableBox>(parent_box->first_paintable()))
+        return paintable_box->absolute_padding_box_rect().size();
+    return {};
+}
+
+ElementResizeAction::ElementResizeAction(GC::Ref<DOM::Element> element, CSSPixelPoint pointer_down_origin)
+    : m_element(element)
+    , m_pointer_down_origin(pointer_down_origin)
+{
+    auto const* paintable_box = m_element->paintable_box();
+    if (paintable_box)
+        m_initial_border_box_size = paintable_box->absolute_border_box_rect().size();
+}
+
+void ElementResizeAction::handle_pointer_move(CSSPixelPoint pointer_position)
+{
+    auto const* paintable_box = m_element->paintable_box();
+    if (!paintable_box)
+        return;
+    auto const& layout_node = paintable_box->layout_node();
+    auto const& computed = layout_node.computed_values();
+    auto resize = computed.resize();
+    if (resize == CSS::Resize::None)
+        return;
+
+    bool horizontal_writing_mode = computed.writing_mode() == CSS::WritingMode::HorizontalTb;
+    bool resize_x = computed.resize() == CSS::Resize::Both
+        || computed.resize() == CSS::Resize::Horizontal
+        || (computed.resize() == CSS::Resize::Inline && horizontal_writing_mode)
+        || (computed.resize() == CSS::Resize::Block && !horizontal_writing_mode);
+
+    bool resize_y = computed.resize() == CSS::Resize::Both
+        || computed.resize() == CSS::Resize::Vertical
+        || (computed.resize() == CSS::Resize::Inline && !horizontal_writing_mode)
+        || (computed.resize() == CSS::Resize::Block && horizontal_writing_mode);
+
+    CSSPixels dx = resize_x ? pointer_position.x() - m_pointer_down_origin.x() : 0;
+    CSSPixels dy = resize_y ? pointer_position.y() - m_pointer_down_origin.y() : 0;
+    auto writing_mode = computed.writing_mode();
+    if ((writing_mode == CSS::WritingMode::HorizontalTb && computed.direction() == CSS::Direction::Rtl)
+        || writing_mode == CSS::WritingMode::VerticalRl
+        || writing_mode == CSS::WritingMode::SidewaysRl) {
+        dx = -dx;
+    }
+    CSSPixels css_width = max(ChromeMetrics::ZOOM_INVARIANT_RESIZE_GRIPPER_SIZE, m_initial_border_box_size.width() + dx);
+    CSSPixels css_height = max(ChromeMetrics::ZOOM_INVARIANT_RESIZE_GRIPPER_SIZE, m_initial_border_box_size.height() + dy);
+
+    auto reference_basis = containing_block_padding_box_size(layout_node);
+
+    if (reference_basis.has_value()) {
+        if (auto const& min_width = computed.min_width(); !min_width.is_auto()) {
+            css_width = max(css_width, min_width.to_px(layout_node, reference_basis->width()));
+        }
+        if (auto const& max_width = computed.max_width(); !max_width.is_none()) {
+            css_width = min(css_width, max_width.to_px(layout_node, reference_basis->width()));
+        }
+        if (auto const& min_height = computed.min_height(); !min_height.is_auto()) {
+            css_height = max(css_height, min_height.to_px(layout_node, reference_basis->height()));
+        }
+        if (auto const& max_height = computed.max_height(); !max_height.is_none()) {
+            css_height = min(css_height, max_height.to_px(layout_node, reference_basis->height()));
+        }
+    }
+    if (computed.box_sizing() == CSS::BoxSizing::ContentBox) {
+        auto const& metrics = paintable_box->box_model();
+        css_width -= metrics.padding.left + metrics.padding.right + computed.border_left().width + computed.border_right().width;
+        css_height -= metrics.padding.top + metrics.padding.bottom + computed.border_top().width + computed.border_bottom().width;
+    }
+
+    auto style = m_element->style_for_bindings();
+    auto width_str = MUST(String::formatted("{:.2f}px", max(0.0, css_width.to_double())));
+    auto height_str = MUST(String::formatted("{:.2f}px", max(0.0, css_height.to_double())));
+
+    MUST(style->set_property(CSS::PropertyID::Width, width_str));
+    MUST(style->set_property(CSS::PropertyID::Height, height_str));
+}
+
+void ElementResizeAction::visit_edges(GC::Cell::Visitor& visitor) const
+{
+    visitor.visit(m_element);
+}
+
+}
--- a/Libraries/LibWeb/Page/ElementResizeAction.h
+++ b/Libraries/LibWeb/Page/ElementResizeAction.h
@@ -1,0 +1,31 @@
+/*
+ * Copyright (c) 2025, Jonathan Gamble <gamblej@gmail.com>
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#pragma once
+
+#include <LibGC/Cell.h>
+#include <LibWeb/Forward.h>
+#include <LibWeb/PixelUnits.h>
+
+// https://drafts.csswg.org/css-ui#resize
+
+namespace Web {
+
+class ElementResizeAction {
+public:
+    ElementResizeAction(GC::Ref<DOM::Element> element, CSSPixelPoint pointer_down_origin);
+
+    void handle_pointer_move(CSSPixelPoint pointer_position);
+
+    void visit_edges(GC::Cell::Visitor&) const;
+
+private:
+    GC::Ref<DOM::Element> m_element;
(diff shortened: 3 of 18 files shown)


Libraries/LibWeb/CSS/Properties.json
```diff
@@ -3387,6 +3387,18 @@
...
+    "animation-type": "none",
+    "inherited": false,
+    "affects-layout": false,
+    "initial": "none",
+    "valid-identifiers": [
+      "none",
+      "both",
+      "horizontal",
+      "vertical"
+    ]
```
AtkinsSJ: You defined a `resize` enum, so you can use that instead of listing the values again. eg:

```json
"valid-types": [
  "resize"
]
```

Libraries/LibWeb/Page/ElementResizeAction.cpp
```diff
@@ -0,0 +1,99 @@
...
+{
+    auto parent_box = layout_node.containing_block();
+    if (!parent_box)
+        return {};
+
+    auto const* parent_paintable = parent_box->first_paintable();
+    if (!parent_paintable || !parent_paintable->is_paintable_box())
+        return {};
+
+    return static_cast<Painting::PaintableBox const*>(parent_paintable)->absolute_padding_box_rect().size();
```
AtkinsSJ: You can simplify the "if it's an X, cast to X and use it" code using `as_if<T>()`. It'll try and cast, and return null if it fails, and also is safe to pass null to. So this could become:

```c++
if (auto* parent_paintable = as_if<Painting::PaintableBox>(parent_box->first_paintable()))
    return parent_paintable->absolute_padding_box_rect().size();
return {};
```
...which is a little neater.

---

...ah, I see you already made this change later. :sweat_smile:

Libraries/LibWeb/Page/ChromeMetrics.h
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Web {
+
+struct ChromeMetrics {
+    static constexpr CSSPixels ScrollThumbThickness { 6 };
+    static constexpr CSSPixels ScrollThumbExpandedThickness { 12 };
+
+    static constexpr CSSPixels ScrollThumbMinLength { 24 };
+    static constexpr CSSPixels ResizeGripperSize { 12 };
+};
```
AtkinsSJ: This is a nice idea, but I think it should come in a separate commit *before* the resize-gripper change. We like to keep commits "atomic". It makes it easier to identify where bugs appeared, and also means reverting one change doesn't also revert other ones.

I'm not sure if casing these constants `LikeThis` is right either, we're not super consistent with that for constants. We have some `LIKE_THIS` and others `like_this`, and I think the latter is most common.

Libraries/LibWeb/Page/ElementResizeAction.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+    auto style = m_element->style_for_bindings();
+    auto width_str = MUST(String::formatted("{:.2f}px", max(0.0, css_width.to_double())));
+    auto height_str = MUST(String::formatted("{:.2f}px", max(0.0, css_height.to_double())));
+
+    MUST(style->set_property(CSS::PropertyID::Width, width_str));
+    MUST(style->set_property(CSS::PropertyID::Height, height_str));
+    CSSPixelRect invalidated = paintable_box->absolute_border_box_rect();
+    invalidated.set_size(css_width, css_height);
+    m_element->document().set_needs_display(invalidated, InvalidateDisplayList::No);
```
AtkinsSJ: I think the invalidated rect will be wrong if it's smaller than the previous size - you want the max of the old and new size in each dimension.

Are these lines needed at all? I think setting the width and height will already trigger a layout and painting update.

Libraries/LibWeb/Painting/PaintableBox.cpp
```diff
@@ -1037,9 +1107,28 @@ void PaintableWithLines::paint(DisplayListRecordingContext& context, PaintPhase
     }
 }
 
+bool PaintableBox::has_resizer() const
+{
+    return is<HTML::HTMLTextAreaElement>(dom_node().ptr()) && computed_values().resize() != CSS::Resize::None;
```
AtkinsSJ: Technically it's not limited to `<textarea>`. From [the spec](https://drafts.csswg.org/css-ui/#resize):
> Applies to: elements that are [scroll containers](https://drafts.csswg.org/css-overflow-3/#scroll-container) and optionally replaced elements such as images, videos, and iframes

If what you've written only works for `<textarea>` then this is fine, but leave a FIXME comment to make it clear that more needs doing.

Libraries/LibWeb/Painting/PaintableBox.cpp
```diff
@@ -1118,16 +1207,11 @@ void PaintableBox::handle_mouseleave(Badge<EventHandler>)
...
-    TemporaryChange force_enlarged_vertical_scrollbar { m_draw_enlarged_vertical_scrollbar, true };
-
-    auto scrollbar_data = compute_scrollbar_data(direction);
-    if (!scrollbar_data.has_value())
-        return false;
-
-    return scrollbar_data->gutter_rect.contains(position);
+    if (auto rect = absolute_scrollbar_rect(direction, true); rect.has_value())
+        return rect->contains(adjusted_position);
+    return false;
```
AtkinsSJ: These scrollbar changes could definitely do with being in a separate commit. If nothing else, it'd be easier to review. :wink:

Libraries/LibWeb/Painting/PaintableBox.cpp
```diff
@@ -1037,9 +1107,28 @@ void PaintableWithLines::paint(DisplayListRecordingContext& context, PaintPhase
     }
 }
 
+bool PaintableBox::has_resizer() const
+{
+    return is<HTML::HTMLTextAreaElement>(dom_node().ptr()) && computed_values().resize() != CSS::Resize::None;
+}
+
+bool PaintableBox::is_resizer_south_west() const
```
AtkinsSJ: Nit: `bottom_left` would be more in line with CSS lingo. But also, it's not obvious what `false` would mean. Maybe change this to something like `ResizerPosition PaintableBox::resizer_position()` returning an enum, to make it clearer.

Case in point: It's taken me this far into the review to remember that west is *left* not *right*. :sweat_smile:

Tests/LibWeb/Text/expected/css/CSSStyleDeclaration-has-indexed-property-getter.txt
```diff
@@ -254,6 +254,7 @@ All properties associated with getComputedStyle(document.body):
     "position-try-order",
     "position-visibility",
     "r",
+    "resize",
```
AtkinsSJ, line 258: The changes in this commit should be applied to the commit that adds the property. Ideally all tests should pass after any commit.

Libraries/LibGfx/Rect.h
```diff
@@ -334,6 +334,8 @@ class Rect {
     ALWAYS_INLINE void set_primary_offset_for_orientation(Orientation orientation, T value) { m_location.set_primary_offset_for_orientation(orientation, value); }
     [[nodiscard]] ALWAYS_INLINE T secondary_offset_for_orientation(Orientation orientation) const { return m_location.secondary_offset_for_orientation(orientation); }
     ALWAYS_INLINE void set_secondary_offset_for_orientation(Orientation orientation, T value) { m_location.set_secondary_offset_for_orientation(orientation, value); }
+    ALWAYS_INLINE void translate_primary_offset_for_orientation(Orientation orientation, T delta) { m_location.set_primary_offset_for_orientation(orientation, m_location.primary_offset_for_orientation(orientation) + delta); }
+    ALWAYS_INLINE void translate_secondary_offset_for_orientation(Orientation orientation, T delta) { m_location.set_secondary_offset_for_orientation(orientation, m_location.secondary_offset_for_orientation(orientation) + delta); }
```
AtkinsSJ, line 338: For this whole commit: When you make changes within a PR that are minor tweaks to previous commits, amend those commits instead of creating a new one. It makes the commit history easier to follow. (And also avoids maintainers reviewing code that gets removed or changed right after.)

Libraries/LibWeb/Painting/PaintableBox.cpp
```diff
@@ -508,8 +506,7 @@ Optional<PaintableBox::ScrollbarData> PaintableBox::compute_scrollbar_data(Scrol
         CSSPixels scroll_offset = is_horizontal ? -own_scroll_frame_offset().x() : -own_scroll_frame_offset().y();
         CSSPixels thumb_offset = scroll_offset * scrollbar_data.thumb_travel_to_scroll_ratio;
 
-        scrollbar_data.thumb_rect.set_primary_offset_for_orientation(
-            orientation, scrollbar_data.thumb_rect.primary_offset_for_orientation(orientation) + thumb_offset);
+        scrollbar_data.thumb_rect.translate_primary_offset_for_orientation(orientation, thumb_offset);
```
AtkinsSJ, line 491: Same point with this commit.

Libraries/LibWeb/Painting/ChromeMetrics.h
```diff
@@ -0,0 +1,45 @@
...
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#pragma once
+
+#include <LibWeb/PixelUnits.h>
+
+namespace Web {
+
+struct ChromeMetrics {
```
AtkinsSJ, line 13: For example, this file should be created in `Painting/` when you first add it, instead of creating it and then moving it.

You might be able to cut down on churn if you made the zoom fix commit happen earlier, but I'm not going to ask you to rearrange so much.

AtkinsSJ (changes requested):
I'm very happy to see this!

Libraries/LibWeb/Page/ChromeMetrics.h
jonbgamble: From the contributing guidelines, seems like screaming case.

Libraries/LibWeb/Page/ElementResizeAction.cpp
jonbgamble: Looks like the rect is ignored anyway.

Libraries/LibWeb/Painting/PaintableBox.cpp
jonbgamble: I am aware of the spec but I have only tested it on textarea. ~~I don't think WebKit supports it on anything else, not sure about the others.~~ [Edit - wrong]

Libraries/LibWeb/Painting/PaintableBox.cpp
jonbgamble: Compass points are from the corresponding cursor property names. I'm not sure an enum would be clearer than a boolean but just confirm it's wanted and I'll do it.

jonbgamble:
Are the history related comments informational or did you want a little history rewrite applied here?

jonbgamble:
I'll just do the full rewrite. It will be good practice.

jonbgamble:
Sam, should I integrate the changes in https://github.com/LadybirdBrowser/ladybird/pull/7272 into this PR as a first step?

Or we could separately merge that PR first, then I'll rewrite this history to fit. Lemme know.

The reason is that PaintableBox/PaintableWithLines is turning out to be a busy area for a few of my attempted contributions. I think splitting the sources now (unless there's some reason why it's a bad idea) will make those merges go smoother if/when they arrive.

AtkinsSJ:
> Sam, should I integrate the changes in #7272 into this PR as a first step?
> 
> Or we could separately merge that PR first, then I'll rewrite this history to fit. Lemme know.
> 
> The reason is that PaintableBox/PaintableWithLines is turning out to be a busy area for a few of my attempted contributions. I think splitting the sources now (unless there's some reason why it's a bad idea) will make those merges go smoother if/when they arrive.

Sorry, I completely missed this comment until now!

Welcome to the wonderful world of stepping on your own toes, repeatedly. :laughing:

I agree with splitting up the files, I think a lot of the time it was just easier to have things together initially, and then they grow over time and nobody really thinks about it.

As you already added that commit here, I think trying to merge this in one go makes sense.

Libraries/LibWeb/Page/ElementResizeAction.cpp
```diff
@@ -0,0 +1,104 @@
...
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include <LibWeb/CSS/PropertyID.h>
+#include <LibWeb/DOM/Element.h>
+#include <LibWeb/Page/ElementResizeAction.h>
+#include <LibWeb/Painting/ChromeMetrics.h>
+#include <LibWeb/Painting/PaintableBox.h>
+
+// https://drafts.csswg.org/css-ui#resizeki
```
AtkinsSJ: This URL seems to be wrong? Not sure what it's applied to either.

Libraries/LibWeb/Page/ElementResizeAction.h
```diff
@@ -0,0 +1,31 @@
...
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#pragma once
+
+#include <LibGC/Cell.h>
+#include <LibWeb/Forward.h>
+#include <LibWeb/PixelUnits.h>
+
+// https://drafts.csswg.org/css-ui#resizeki
```
AtkinsSJ: Same. https://drafts.csswg.org/css-ui#resize is a link to the property definition, maybe you wanted that?

Libraries/LibWeb/Painting/PaintableBox.cpp
```diff
@@ -480,17 +543,42 @@ void PaintableBox::paint(DisplayListRecordingContext& context, PaintPhase phase)
...
-            auto thumb_rect = context.rounded_device_rect(scrollbar_data->thumb_rect).to_type<int>();
-            context.display_list_recorder().paint_scrollbar(own_scroll_frame_id().value(), gutter_rect, thumb_rect, scrollbar_data->scroll_length, scrollbar_colors.thumb_color, scrollbar_colors.track_color, false);
+        if (auto resizer_rect = absolute_resizer_rect(metrics); resizer_rect.has_value()) {
+            bool bottom_left_resizer = is_resizer_bottom_left();
+            CSSPixels padding = metrics.resize_gripper_padding;
+            CSSPixelRect css_rect = resizer_rect.value()
+                                        .shrunken(padding, padding)
+                                        .translated(bottom_left_resizer ? padding / 2 : -padding / 2, -padding / 2);
+            Gfx::IntRect rect = context.rounded_device_rect(css_rect).to_type<int>();
+            Gfx::Color resizer_line_color = Gfx::Color::from_rgba(0x60000000);
```
AtkinsSJ: Could you use a color from [`Libraries/LibWeb/CSS/SystemColor.h`](https://github.com/LadybirdBrowser/ladybird/blob/master/Libraries/LibWeb/CSS/SystemColor.h) for this? Otherwise, it's quite hard to see in dark mode:

Screenshot_20260109_113432

```html
<!doctype html>
<style>
  :root {
    color-scheme:dark;
  }
</style>
<textarea>Wheeeee</textarea>
```

Libraries/LibWeb/Painting/PaintableBox.cpp
```diff
@@ -679,20 +767,52 @@ void PaintableBox::clear_clip_overflow_rect(DisplayListRecordingContext& context
     context.display_list_recorder().pop_clip_frame();
 }
 
+bool PaintableBox::has_resizer() const
+{
+    // https://drafts.csswg.org/css-ui#resizeki
```
AtkinsSJ: Again I think this wants to be `#resize`.

Libraries/LibWeb/Painting/PaintableBox.cpp
```diff
@@ -1264,4 +1409,35 @@ Optional<Gfx::Filter> PaintableBox::resolve_filter(DisplayListRecordingContext&
...
+static PhysicalResizeAxes compute_physical_resize_axes(CSS::ComputedValues const& computed)
+{
+    // https://drafts.csswg.org/css-ui/#resizing
+    if (computed.resize() == CSS::Resize::None)
+        return {};
+
+    // The resize property applies to elements that are scroll containers. UAs may also apply it,
+    // regardless of the value of the overflow property, to:
+    // Replaced elements representing images or videos, such as img, video, picture, svg, object, or canvas.
+    // The <iframe> element.
```
AtkinsSJ: Nit: These two lines could do with some `// - ` bullet points for clarity.

Libraries/LibWeb/Painting/PaintableBox.cpp
jonbgamble: Alternating black & white lines with subtle opacity should be visible on almost any background without requiring themes. I am not averse to SystemColor, but which one did you have in mind?

jonbgamble:
Here's the alternating black/white line gripper visual against various backgrounds.

image

On divs without grippers, the scrollbars are pegged at inline-end & block-end to demonstrate their non-overlappiness.

Libraries/LibWeb/Painting/PaintableBox.cpp
AtkinsSJ: I didn't have a particular one in mind, it's just a set of colours that adjust to light/dark mode, so it's convenient. Alternating light and dark works!

PR #7219, opened 2025-12-22
Meta+Tests: Enforce test concurrency limit via environment variable

### LINKED ISSUE #7216: `ladybird.py` test runner fails to propagate `-j` flag to `test-web`
### Summary

The `ladybird.py` test runner accepts a `-j` / `--jobs` argument but does not successfully propagate a concurrency limit to the underlying `test-web` C++ binary. This causes `test-web` to default to the system's hardware concurrency (e.g. spawning 8+ processes) even when the user requested` -j1`.

### Operating system

Linux (Ubuntu)

### Steps to reproduce

1. On a machine with multiple cores (e.g. 4 cores), run the test command: `./Meta/ladybird.py test -j1`
2. In a separate terminal, check the number of active WebContent processes: `pgrep -c WebContent`
3. Observe that the count reflects the hardware core count rather than the requested limit.

### Expected behavior

The command should spawn **exactly** 1 `WebContent` process, strictly adhering to the `-j1` argument.

### Actual behavior

The command ignores the `-j1` flag and spawns a number of processes equal to the _hardware threads_ (e.g., 4 or 8), causing OOM on systems with limited RAM.

N/A

```html
N/A (Tooling/Infrastructure bug, reproducible via CLI)
```

```shell
$ ./Meta/ladybird.py test -j1
# Running...

# In a separate terminal:
$ pgrep -c WebContent
8 
# (Expected: 1)
```

_No response_

_No response_

### Contribute a patch?

Emin-ACIKGOZ:
Currently, running `./Meta/ladybird.py test` defaults to spawning WebContent processes equal to the system core count. On systems with 8GB RAM, this leads to immediate OOM/freezing as 4-8 processes launch simultaneously, even if the job flag is specified. 

This PR updates the runner to propagate the `-j` argument to the underlying `test-web` binary (via environment variable), allowing users to limit concurrency (e.g., `-j 1`) and run the suite safely on constrained hardware. 

Fixes #7216

### DIFF
--- a/Meta/ladybird.py
+++ b/Meta/ladybird.py
@@ -147,7 +147,7 @@ def main():
     elif args.command == "test":
         build_dir = configure_main(platform, args.preset, args.cc, args.cxx)
         build_main(build_dir, args.jobs)
-        test_main(build_dir, args.preset, args.pattern)
+        test_main(build_dir, args.preset, args.jobs, args.pattern)
     elif args.command == "run":
         if args.preset == "Sanitizer":
             # FIXME: Find some way to centralize these b/w CMakePresets.json, CI files, Documentation and here.
@@ -352,7 +352,7 @@ def build_main(build_dir: Path, jobs: Optional[str], target: Optional[str] = Non
     run_command(build_args, exit_on_failure=True)
 
 
-def test_main(build_dir: Path, preset: str, pattern: Optional[str]):
+def test_main(build_dir: Path, preset: str, jobs: Optional[str], pattern: Optional[str]):
     test_args = [
         "ctest",
         "--preset",
@@ -361,6 +361,12 @@ def test_main(build_dir: Path, preset: str, pattern: Optional[str]):
         "--test-dir",
         str(build_dir),
     ]
+
+    if not jobs:
+        jobs = os.environ.get("MAKEJOBS", None)
+    if jobs:
+        test_args.extend(["-j", jobs])
+        os.environ["LADYBIRD_TEST_CONCURRENCY"] = jobs
 
     if pattern:
         test_args.extend(["-R", pattern])
--- a/Tests/LibWeb/test-web/main.cpp
+++ b/Tests/LibWeb/test-web/main.cpp
@@ -595,6 +595,15 @@ static ErrorOr<int> run_tests(Core::AnonymousBuffer const& theme, Web::DevicePix
         return Error::from_string_literal("No tests found matching filter");
     }
 
+    // CTest cannot forward the -j flag, so look for the bridge variable set by ladybird.py
+    // NOLINTNEXTLINE(concurrency-mt-unsafe)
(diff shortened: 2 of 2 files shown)


Meta/ladybird.py
```diff
@@ -362,6 +362,12 @@ def test_main(build_dir: Path, preset: str, pattern: Optional[str]):
         str(build_dir),
     ]
 
+    if not jobs:
+        jobs = os.environ.get("MAKEJOBS", None)
+    if jobs:
+        test_args.extend(["-j", jobs])
```
trflynn89, line 368: This will introduce parallelism to `ctest` which is something we previously haven't done. Individual tests all currently run in serial. Let's not introduce this behavior change here.

Meta/ladybird.py
```diff
@@ -362,6 +362,12 @@ def test_main(build_dir: Path, preset: str, pattern: Optional[str]):
         str(build_dir),
     ]
 
+    if not jobs:
```
trflynn89, line 365: Instead of copy-pasting this from `build_main`, let's either:
A) Do this work ahead of time and pass the result into `build_main` and `test_main`
B) Add a "get number of jobs" function that handles the env var fallback

Tests/LibWeb/test-web/main.cpp
```diff
@@ -595,6 +595,15 @@ static ErrorOr<int> run_tests(Core::AnonymousBuffer const& theme, Web::DevicePix
         return Error::from_string_literal("No tests found matching filter");
     }
 
+    // CTest cannot forward the -j flag, so look for the bridge variable set by ladybird.py
+    // NOLINTNEXTLINE(concurrency-mt-unsafe)
+    if (char* env_concurrency = getenv("LADYBIRD_TEST_CONCURRENCY")) {
```
trflynn89, line 600: We have wrappers around `getenv` so we can remove this linter directive. See `Core::Environment::get` in `Libraries/LibCore/Environment.h`.

Tests/LibWeb/test-web/main.cpp
```diff
@@ -595,6 +595,15 @@ static ErrorOr<int> run_tests(Core::AnonymousBuffer const& theme, Web::DevicePix
...
+        if (auto val = StringView { env_concurrency, strlen(env_concurrency) }.to_number<size_t>(); val.has_value()) {
```
trflynn89, line 601: Let's use slightly more descriptive variable names than `val`

Tests/LibWeb/test-web/main.cpp
```diff
@@ -595,6 +595,15 @@ static ErrorOr<int> run_tests(Core::AnonymousBuffer const& theme, Web::DevicePix
...
+            app.test_concurrency = val.value();
+            outln("Forcing concurrency to {} via LADYBIRD_TEST_CONCURRENCY", app.test_concurrency);
```
trflynn89, line 603: I don't think we need to write this to the console

PR #7220, opened 2025-12-22
Tests: Recycle WebContent processes to prevent memory exhaustion

### LINKED ISSUE #7217: `test-web` process accumulates unbounded memory during full suite run
### Summary

The `test-web` binary reuses a single `WebContent` process for the entire duration of a test partition. On long runs (like the full `LibWeb` suite), this process accumulates memory indefinitely due to retained state, leading to crashes on constrained hardware.

### Operating system

Linux

### Steps to reproduce

1. Run the full `LibWeb` test suite on a machine with 8GB RAM: `./Meta/ladybird.py test LibWeb`
2. Open `htop` and monitor the RES (Resident Memory) column for the `WebContent` processes.
3. Observe the memory usage as the test count crosses 1000.

### Expected behavior

Memory usage should remain stable, or the process should be recycled periodically to prevent unbounded growth, allowing the suite to complete under constrained conditions.

### Actual behavior

Memory usage grows linearly. It starts at ~100MB, exceeds 2GB per process after ~1000 tests, and eventually causes a hard system freeze or OOM crash before the suite finishes.

N/A

```html
N/A
```

```shell
# Monitoring process memory via htop during test run:

# Test 1:    ~150 MB
# Test 500:  ~1.2 GB
# Test 1000: ~2.1 GB

# Eventually results in system freeze / OOM Killer: Out of memory
```

_No response_

_No response_

### Contribute a patch?

Emin-ACIKGOZ:
Long-running test sessions currently accumulate significant memory usage in the WebContent process due to caching and fragmentation, eventually causing OOM even when concurrency is limited. This patch implements a recycling mechanism in TestWebView:

- Tracks the number of tests executed by the current process.

- Automatically restarts the WebContent process after a fixed threshold (e.g., 50 tests).

- This ensures memory usage follows a stable "sawtooth-like" pattern rather than growing linearly, enabling the full suite to pass on machines with 8GB RAM.

Fixes #7217

### DIFF
--- a/Tests/LibWeb/test-web/TestWebView.cpp
+++ b/Tests/LibWeb/test-web/TestWebView.cpp
@@ -60,4 +60,10 @@ void TestWebView::on_test_complete(TestCompletion completion)
     m_test_promise->resolve(move(completion));
 }
 
+void TestWebView::restart_web_content_process()
+{
+    initialize_client(CreateNewClient::Yes);
+    m_run_count = 0;
+}
+
 }
--- a/Tests/LibWeb/test-web/TestWebView.h
+++ b/Tests/LibWeb/test-web/TestWebView.h
@@ -29,6 +29,10 @@ public:
     TestPromise& test_promise() { return *m_test_promise; }
     void on_test_complete(TestCompletion);
 
+    void restart_web_content_process();
+    size_t run_count() const { return m_run_count; }
+    void increment_run_count() { ++m_run_count; }
+
 private:
     TestWebView(Core::AnonymousBuffer theme, Web::DevicePixelSize viewport_size);
 
@@ -37,6 +41,8 @@ private:
     RefPtr<Core::Promise<RefPtr<Gfx::Bitmap const>>> m_pending_screenshot;
 
     NonnullRefPtr<TestPromise> m_test_promise;
+
+    size_t m_run_count { 0 };
 };
 
 }
--- a/Tests/LibWeb/test-web/main.cpp
+++ b/Tests/LibWeb/test-web/main.cpp
@@ -704,6 +704,11 @@ static ErrorOr<int> run_tests(Core::AnonymousBuffer const& theme, Web::DevicePix
             if (result.result != TestResult::Pass)
                 non_passing_tests.append(move(result));
 
+            view->increment_run_count();
+            if (view->run_count() >= 50) {
+                view->restart_web_content_process();
+            }
+
             if (--tests_remaining == 0)
                 s_all_tests_complete->resolve({});
             else


gmta:
Hi @Emin-ACIKGOZ !

> due to caching and fragmentation

Sounds like you've investigated the cause of the processes' memory usage, can you show what you found and what the biggest culprits are?

I'm not entirely against limiting the number of tests per WebContent process, but I'd like to make sure we're not dealing with a memory leak that we should fix instead so all modes of operation can benefit.

Emin-ACIKGOZ:
Hi @gmta!

I haven't yet performed a deep-dive into the codebase to identify specific leaks, so I cannot say with certainty if the growth is purely due to leaks or legitimate caching and fragmentation.

However, my observation is that memory usage grows linearly until it exhausts available RAM on 8GB machines. While hunting for specific leaks is a more definitive long-term solution, I believe process recycling serves as an important safety net in the meantime.

From my observation the spike occurs during LibWeb Tests, so it seems the root cause for the memory usage is somewhere there. 

I've attached a comparison of running tests before and after applying my change. 

Before:
pre-commit

After:
post-commit

gmta:
> I believe process recycling serves as an important safety net in the meantime.

The downside of this is that we don't _feel_ the problem as much and can delay fixing the underlying issues. Imho this is the same as skipping flaky LibWeb tests in order to get CI to show a green check, I'd much rather see the causes fixed instead of having an evergrowing list of workarounds that tend to hide the actual problem (this is a bit charged of course).

Emin-ACIKGOZ:
> > I believe process recycling serves as an important safety net in the meantime.
> 
> The downside of this is that we don't _feel_ the problem as much and can delay fixing the underlying issues. Imho this is the same as skipping flaky LibWeb tests in order to get CI to show a green check, I'd much rather see the causes fixed instead of having an evergrowing list of workarounds that tend to hide the actual problem (this is a bit charged of course).

​@gmta I totally understand the concern about masking underlying issues, fixing the root leaks is paramount for maintainability.

​However, on systems with 8GB RAM (like mine), the suite is currently unrunnable due to immediate OOM crashes. This fix is the only way I can run the tests without the system freezing.

​Would you be open to making this an opt-in flag? That way, we don't hide leaks by default, but we still allow contributors with hardware constraints to actually run the suite.

gmta:
> Would you be open to making this an opt-in flag?

Yes definitely - having that as an optional flag sounds very reasonable!

Emin-ACIKGOZ:
Alright, I'll look into it as soon as possible. Thanks!

PR #7222, opened 2025-12-22
LibWeb/SVG: Percent-decode same-document `<use>` href fragments

kalenikaliaksandr:
`SVGUseElement::referenced_element()` previously passed `m_href->fragment()` directly to `Document::get_element_by_id()`. URL fragments are stored in their serialized form, so references to elements whose IDs contain non-ASCII characters would fail to resolve.

Before:
CleanShot 2025-12-22 at 16 21 33@2x

After:
CleanShot 2025-12-22 at 16 20 42@2x


Libraries/LibWeb/SVG/SVGUseElement.cpp
```diff
@@ -138,8 +138,10 @@ GC::Ptr<DOM::Element> SVGUseElement::referenced_element()
     if (!m_href->fragment().has_value())
         return nullptr;
 
-    if (is_referenced_element_same_document())
-        return document().get_element_by_id(*m_href->fragment());
+    if (is_referenced_element_same_document()) {
+        auto id = MUST(String::from_byte_string(URL::percent_decode(*m_href->fragment())));
```
shannonbooth: what if the percent decoding is not a valid utf8 string? from memory, I think we need to convert with replacement (I recall needing to add this somewhere else)

shannonbooth: Ah, yup here: https://github.com/LadybirdBrowser/ladybird/blob/0dff009b419e79a2fba65c634813d8d4fd958353/Libraries/LibWeb/DOM/Document.cpp#L2602

Libraries/LibWeb/SVG/SVGUseElement.cpp
```diff
@@ -138,8 +138,10 @@ GC::Ptr<DOM::Element> SVGUseElement::referenced_element()
     if (!m_href->fragment().has_value())
         return nullptr;
```
shannonbooth, line 140: Ah, just caught as I was going to approve one more case we are doing it here - https://github.com/LadybirdBrowser/ladybird/blob/b317461d50d1e7a31bdbbdf6027f9334f30e865f/Libraries/LibWeb/SVG/SVGUseElement.cpp#L128

PR #7223, opened 2025-12-22
Some GC leak fixes for test-web

awesomekling:
Altogether these bring the number of leaked Windows at the end of a full `test-web` run from 240 to 128. Still bad, but a step in the right direction.

`GC::Root` was the main culprit here, and I suspect most of the remaining leaks will be root related as well..

### DIFF
--- a/Libraries/LibWeb/DOM/DocumentLoadEventDelayer.cpp
+++ b/Libraries/LibWeb/DOM/DocumentLoadEventDelayer.cpp
@@ -10,7 +10,7 @@
 namespace Web::DOM {
 
 DocumentLoadEventDelayer::DocumentLoadEventDelayer(Document& document)
-    : m_document(GC::make_root(document))
+    : m_document(document)
 {
     m_document->increment_number_of_things_delaying_the_load_event({});
 }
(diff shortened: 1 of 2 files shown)


Libraries/LibWeb/DOM/DocumentLoadEventDelayer.cpp
```diff
@@ -35,4 +35,9 @@ DocumentLoadEventDelayer::~DocumentLoadEventDelayer()
         m_document->decrement_number_of_things_delaying_the_load_event({});
 }
 
+void DocumentLoadEventDelayer::visit_edges(GC::Cell::Visitor& visitor)
+{
+    visitor.visit(m_document);
```
trflynn89: With this no longer a root, I think the move ctor / assignment operator above need to be updated to clear out the moved-from document ref.

trflynn89: Looks like Alex was also looking at making this a weak reference instead, but probably ran into a similar issue: #7083

awesomekling: Oh hmm, his fix seems nicer actually.. no need for the flimsy `visit_edges()`. Will deboog either way

PR #7229, opened 2025-12-24
LibWeb: Fix tofu in textarea newlines on MacOS

jonbgamble:
See the tofu after `return:` in this mac `<textarea>`. It's caused by code which trusts and inserts the platform event's key code. But this value can vary.

Screenshot 2025-12-23 at 19 41 03

Seeing how it is 2025 and \n has won, let us always insert 0x0a when an enter or return key is pressed.

### DIFF
--- a/Libraries/LibWeb/HTML/FormAssociatedElement.cpp
+++ b/Libraries/LibWeb/HTML/FormAssociatedElement.cpp
@@ -890,20 +890,6 @@ void FormAssociatedTextControlElement::handle_delete(DeleteDirection direction)
     did_edit_text_node();
 }
 
-EventResult FormAssociatedTextControlElement::handle_return_key(FlyString const&)
-{
-    auto* input_element = as_if<HTMLInputElement>(form_associated_element_to_html_element());
-    if (!input_element)
-        return EventResult::Dropped;
-
-    if (auto* form = input_element->form())
-        form->implicitly_submit_form().release_value_but_fixme_should_propagate_errors();
-    else
-        input_element->commit_pending_changes();
-
-    return EventResult::Handled;
-}
-
 void FormAssociatedTextControlElement::collapse_selection_to_offset(size_t position)
 {
     m_selection_start = position;
--- a/Libraries/LibWeb/HTML/FormAssociatedElement.h
+++ b/Libraries/LibWeb/HTML/FormAssociatedElement.h
@@ -243,7 +243,6 @@ public:
 
     virtual void handle_insert(Utf16String const&) override;
     virtual void handle_delete(DeleteDirection) override;
-    virtual EventResult handle_return_key(FlyString const& ui_input_type) override;
     virtual void select_all() override;
     virtual void set_selection_anchor(GC::Ref<DOM::Node>, size_t offset) override;
     virtual void set_selection_focus(GC::Ref<DOM::Node>, size_t offset) override;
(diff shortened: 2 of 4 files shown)


Libraries/LibWeb/HTML/HTMLTextAreaElement.cpp
```diff
@@ -437,6 +437,12 @@ void HTMLTextAreaElement::did_edit_text_node()
     update_placeholder_visibility();
 }
 
+EventResult HTMLTextAreaElement::handle_return_key(FlyString const&)
```
shannonbooth, line 440: By doing an override to implement this, the new structure of the code is a bit confusing in that both `HTMLInputElement` and `HTMLTextAreaElement` are instances of `FormAssociatedTextControlElement`, but `HTMLInputElement` implements its special behavior directly in `FormAssociatedTextControlElement::handle_return_key` whereas `HTMLTextArea` does it through an override. I think we should be consistent with the approach between both elements. (small nit on this, but I think it would be a little more readable if we did `handle_insert(Utf16String::from_code_point('\n'));`)


However, I have confused myself reading HTMLTextAreaElement - if we normalize newlines before insertion always (like here), what is the need for normalizing newlines when returning the API value, since by that point there would be no CRs in the TextAreaElement? https://github.com/LadybirdBrowser/ladybird/blob/711100e3b728d1442b55b760ac67f1a50d682603/Libraries/LibWeb/HTML/HTMLTextAreaElement.cpp#L216, trying to understand this whole 'raw value' business a bit better: https://html.spec.whatwg.org/multipage/form-elements.html#the-textarea-element:concept-textarea-raw-value-8 - edit: I am still confused about raw value a little bit, but I've convinced myself that this change is correct!

I was also going to suggest adding a test for this case using our Internals javascript test interface: https://github.com/LadybirdBrowser/ladybird/blob/711100e3b728d1442b55b760ac67f1a50d682603/Libraries/LibWeb/Internals/Internals.idl#L25, but I suppose it is actually not possible in this case due to the normalization at the API boundary!

Libraries/LibWeb/HTML/HTMLTextAreaElement.cpp
```diff
@@ -437,6 +437,12 @@ void HTMLTextAreaElement::did_edit_text_node()
...
+{
+    handle_insert(Utf16String::from_code_point(0x0A)); // Avoid the platform codepoint
```
shannonbooth, line 442: Looks good! Would it be possible to just change this over to `'\n'` instead as the final thing? I personally don't have the ASCII mapping memorized 😅

jonbgamble: Sure, but first I'll explain why I chose 0x0a. \n in docs, regexes, config formats, and tooling can map to CRLF, or even other unicode linebraks, rather than exactly the linefeed character at code point 10 which is what goes here.

Shall we still pull the trigger and make it \n?

jonbgamble: Re handle_insert, raw value, and line ending normalization as it relates to the actual drawing code in paintablebox and editing code, I think it's safe to say this PR is just rearranging a single deck chair on a boat with many icebergs in its future.

shannonbooth: Fair, I don't feel strongly on this, fine with it as-is. To explain my POV it is more common in the codebase to call APIs such as `append_code_point` taking the character directly such as  `builder.append_code_point('p');` as opposed to the codepoint integer. So it seems natural to me `from_code_point('\n')` could only be LF without remembering the decimal

shannonbooth (approved):

PR #7233, opened 2025-12-25
LibWeb: Make input controls relinquish focus on outside clicks

jonbgamble:
Existing behavior: Text controls don't surrender focus on outside clicks. For example, once an input text has focus you can't select page text outside of it.

In this PR: An active input control will relinquish focus when an external target gets a mousedown, whether that target is focusable or not.

### DIFF
--- a/Libraries/LibWeb/HTML/Focus.cpp
+++ b/Libraries/LibWeb/HTML/Focus.cpp
@@ -113,9 +113,14 @@ static void run_focus_update_steps(Vector<GC::Root<DOM::Node>> old_chain, Vector
     // 4. For each entry entry in new chain, in reverse order, run these substeps:
     for (auto& entry : new_chain.in_reverse()) {
         // 1. If entry is a focusable area, and the focused area of the document is not entry:
-        if (entry->is_focusable() && entry->document().focused_area() != entry.ptr()) {
-            // 1. Set document's relevant global object's navigation API's focus changed during ongoing navigation to
-            //    true.
+        if (entry->is_document()) {
+            if (entry->document().focused_area() != nullptr) {
+                as<Window>(relevant_global_object(*entry)).navigation()->set_focus_changed_during_ongoing_navigation(true);
+
+                // AD-HOC: null focused_area indicates "viewport focus".
+                entry->document().set_focused_area(nullptr);
+            }
+        } else if (entry->is_focusable() && entry->document().focused_area() != entry.ptr()) {
             as<Window>(relevant_global_object(*entry)).navigation()->set_focus_changed_during_ongoing_navigation(true);
 
             // 2. Designate entry as the focused area of the document.
@@ -242,7 +247,7 @@ void run_focusing_steps(DOM::Node* new_focus_target, DOM::Node* fallback_target,
     new_focus_target->document().set_last_focus_trigger(focus_trigger);
 
     // 8. Run the focus update steps with old chain, new chain, and new focus target respectively.
-    run_focus_update_steps(old_chain, new_chain, new_focus_target);
+    run_focus_update_steps(move(old_chain), move(new_chain), new_focus_target);
 }
 
 // https://html.spec.whatwg.org/multipage/interaction.html#unfocusing-steps
@@ -301,21 +306,28 @@ void run_unfocusing_steps(DOM::Node* old_focus_target)
     auto& top_document = as<DOM::Document>(*old_chain.last());
 
     // 8. If topDocument's node navigable has system focus, then run the focusing steps for topDocument's viewport.
-    if (top_document.navigable()->traversable_navigable()->system_visibility_state() == HTML::VisibilityState::Visible) {
-        run_focusing_steps(&top_document);
+    if (top_document.navigable()->traversable_navigable()->is_focused()) {
+
+        // AD-HOC: Remove top_document from old_chain so step 1 in run_focus_update_steps doesn't cancel the blur.
+        auto without_viewport_surrogate = old_chain;
+        without_viewport_surrogate.take_last();
+
+        auto with_viewport_surrogate = focus_chain(&top_document);
+
+        run_focus_update_steps(move(without_viewport_surrogate), move(with_viewport_surrogate), &top_document);
     } else {
         // FIXME: Otherwise, apply any relevant platform-specific conventions for removing system focus from topDocument's
         // browsing context, and run the focus update steps with old chain, an empty list, and null respectively.
 
-        // What? It already doesn't have system focus, what possible platform-specific conventions are there?
-
-        run_focus_update_steps(old_chain, {}, nullptr);
+        run_focus_update_steps(move(old_chain), {}, nullptr);
     }
 
     // NOTE: The unfocusing steps do not always result in the focus changing, even when applied to the currently focused
     //       area of a top-level traversable. For example, if the currently focused area of a top-level traversable is a
     //       viewport, then it will usually keep its focus regardless until another focusable area is explicitly focused
     //       with the focusing steps.
+
+    // What? It already doesn't have system focus, what possible platform-specific conventions are there?
 }
 
 }
--- a/Libraries/LibWeb/Page/EventHandler.cpp
+++ b/Libraries/LibWeb/Page/EventHandler.cpp
@@ -683,60 +683,69 @@ EventResult EventHandler::handle_mousedown(CSSPixelPoint visual_viewport_positio
     if (button != UIEvents::MouseButton::Primary)
         return EventResult::Handled;
 
-    auto result = paint_root()->hit_test(viewport_position, Painting::HitTestType::TextCursor);
-    if (!result.has_value())
-        return EventResult::Handled;
+    // First do an exact hit test for focus management.
+    auto exact_hit = paint_root()->hit_test(viewport_position, Painting::HitTestType::Exact);
+    GC::Ptr<Painting::Paintable> focus_paintable;
+    GC::Ptr<DOM::Node> focus_dom_node;
+    if (exact_hit.has_value()) {
+        focus_paintable = exact_hit->paintable;
+        focus_dom_node = focus_paintable ? focus_paintable->dom_node() : nullptr;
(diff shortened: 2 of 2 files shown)


Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -2502,11 +2502,14 @@ void Document::set_focused_area(GC::Ptr<Node> node)
 
     // Scroll the viewport if necessary to make the newly focused element visible.
     if (new_focused_element) {
-        new_focused_element->queue_an_element_task(HTML::Task::Source::UserInteraction, [&] {
+        auto focused_element = GC::make_root(*new_focused_element);
```
shannonbooth: This should not need any GC root, `queue_an_element_task` calls `queue_a_task` which accepts a `GC::Function` - and a GC function has knowledge of captures and lets the heap visit possible GC pointers as to not free them. In fact making this a root makes this more likely to leak.

Libraries/LibWeb/Page/EventHandler.cpp
```diff
@@ -679,61 +680,69 @@ EventResult EventHandler::handle_mousedown(CSSPixelPoint visual_viewport_positio
...
-                    focus_candidate = input_control;
-                } else {
-                    for (auto candidate = node; candidate; candidate = candidate->parent_or_shadow_host()) {
-                        if (candidate->is_focusable()) {
-                            focus_candidate = candidate;
-                            break;
-                        }
-                    }
-                }
+    if (button != UIEvents::MouseButton::Primary)
```
shannonbooth, line 683: While I understand why this change was made to lower the indentation here for easy viewing of the diff it would be great if that was split into its own commit. Even trying the minimize whitespace feature isn't helping much 😅  I've skipped trying to unpack what the functional change are here, would appreciate that being split into a commit

Libraries/LibWeb/HTML/Focus.cpp
```diff
@@ -301,15 +303,20 @@ void run_unfocusing_steps(DOM::Node* old_focus_target)
...
+        without_viewport_surrogate.take_last();
+
+        auto with_viewport_surrogate = focus_chain(&top_document);
+
+        run_focus_update_steps(move(without_viewport_surrogate), move(with_viewport_surrogate), &top_document);
     } else {
         // FIXME: Otherwise, apply any relevant platform-specific conventions for removing system focus from topDocument's
         // browsing context, and run the focus update steps with old chain, an empty list, and null respectively.
 
-        // What? It already doesn't have system focus, what possible platform-specific conventions are there?
```
shannonbooth, line 310: Do you understand what the spec is saying above? I guess spec might just be saying to let UI toolkit handle with it or something (and maybe this is just a no-op realistically for us)? In any case, I think this comment is pretty paired to the FIXME above

Libraries/LibWeb/HTML/Focus.cpp
```diff
@@ -113,12 +113,14 @@ static void run_focus_update_steps(Vector<GC::Root<DOM::Node>> old_chain, Vector
...
+        if (entry->is_document()) {
+            if (entry->document().focused_area() != nullptr) {
+                as<Window>(relevant_global_object(*entry)).navigation()->set_focus_changed_during_ongoing_navigation(true);
+                // Use null focused_area to model "viewport" focus.
+                entry->document().set_focused_area(nullptr);
+            }
+        } else if (entry->is_focusable() && entry->document().focused_area() != entry.ptr()) {
             as<Window>(relevant_global_object(*entry)).navigation()->set_focus_changed_during_ongoing_navigation(true);
-
-            // 2. Designate entry as the focused area of the document.
```
shannonbooth, line 121: This is removing some verbatim spec text - we should aim to keep that and AD-HOC anywhere which is different ideally with a comment explaining why we deviate

Libraries/LibWeb/DOM/Node.h
```diff
@@ -465,6 +465,17 @@ class WEB_API Node : public EventTarget
     template<typename U>
     U* shadow_including_first_ancestor_of_type();
 
+    template<typename Predicate>
+    requires requires(Predicate& predicate, Node const& node) { { predicate(node) } -> ConvertibleTo<bool>; }
+    Node const* find_in_shadow_including_ancestry(Predicate&& predicate) const
+    {
+        for (Node const* it = this; it && it->is_connected(); it = it->parent_or_shadow_host_node()) {
```
shannonbooth: When would it not be connected in the tree if the parent is not? It seem inefficient to check this. And actually, is node being connected necessary as a predicate in the first place - I suppose it could also be valid to call a predicated function like this for a disconnected tree

shannonbooth (commented):
Left some initial comments, not a comprehensive review. Still also have to give it a whirl locally.

My main comment would be though it would be great to split out the changes into commits where/if it makes sense to do so. It makes it much easier when reviewing when commit is broken down into its own set of changes alongside the explanation and also when debooging and going through the history of a file.

For example, active_input_events_target as its own commit, constified version of parent_or_shadow_host_node, etc - lets me quickly skip over big portions of the diff without needing to think about it so much.

jonbgamble:
> My main comment would be though it would be great to split out the changes into commits where/if it makes sense to do so. It makes it much easier when reviewing when commit is broken down into its own set of changes alongside the explanation and also when debooging and going through the history of a file.

I can do this, and make each commit compile, but I'll draw the line at ensuring tests pass after each commit when it's hard to do so. Sam asked for this in a previous PR, but it's tough to design a history that 1. stages dependent commits separately and 2. does so in a way that passes all CI. I'm sure you understand.

Libraries/LibWeb/DOM/Node.h
jonbgamble: It was for safety but I really don't like having it there either. If a "find" predicate manipulates the DOM hierarchy, we probably deserve to crash.

Libraries/LibWeb/HTML/Focus.cpp
jonbgamble: I couldn't detect any information there, except maybe reassurance to a similarly baffled reader. (in regards to the deleted line)

Libraries/LibWeb/HTML/Focus.cpp
jonbgamble: Did not clip that intentionally like the other one, thanks for catching.

Libraries/LibWeb/Page/EventHandler.cpp
jonbgamble: Good call, I can and will add a commit prior to the functional changes and establish the style refactor for easier diffing.

Libraries/LibWeb/Page/EventHandler.cpp
```diff
@@ -1010,7 +1015,7 @@ EventResult EventHandler::handle_doubleclick(CSSPixelPoint visual_viewport_posit
                 next_boundary = segmenter.next_boundary(result->index_in_node).value_or(hit_dom_node.length());
             }
 
-            if (auto* target = document.active_input_events_target()) {
+            if (auto* target = document.active_input_events_target(&hit_dom_node)) {
```
shannonbooth, line 1027: I think this change accidentally snuck into this commit

shannonbooth (commented):
Thanks for breaking it down! One little thing I noticed going through. 

Behaviour much nicer overall. I did notice one regression when testing locally though. It seems like for the following HTML page (ChatGPT generated) the input element text no longer seems to be selectable:

```html
<!doctype html>
<html lang="en">
<head>
    <meta charset="utf-8">
    <title>Input Focus Relinquish Test</title>
    <style>
        body {
            font-family: system-ui, sans-serif;
            line-height: 1.5;
            padding: 24px;
        }

        input, textarea {
            display: block;
            margin: 12px 0;
            padding: 6px;
            width: 300px;
        }

        .box {
            margin-top: 24px;
            padding: 16px;
            border: 2px solid #444;
            user-select: text;
        }

        .non-focusable {
            margin-top: 16px;
            padding: 12px;
            background: #eee;
        }

        .focusable {
            margin-top: 16px;
            padding: 12px;
            background: #dff0ff;
        }
    </style>
</head>
<body>
    <h1>Input Focus Relinquish Test</h1>

    <p>
        Steps:
        <ol>
            <li>Click into the input or textarea.</li>
            <li>Click and drag to select text outside the control.</li>
            <li>Click non-focusable elements.</li>
        </ol>
    </p>

    <input type="text" placeholder="Text input (focus me)" />
    <textarea placeholder="Textarea (focus me)"></textarea>

    <div class="box">
        <p>
            This is normal page text.  
            You should be able to click and select this text even after
            focusing the input above.
        </p>
        <p>
            Try selecting across multiple words to ensure focus was dropped
            on <code>mousedown</code>.
        </p>
    </div>

    <div class="non-focusable">
        Non-focusable div. Clicking here should remove focus from the input.
    </div>

    <button class="focusable">
        Focusable button (should also take focus)
    </button>
</body>
</html>
```

Libraries/LibWeb/Page/EventHandler.cpp
jonbgamble: I think the problem is I did not test active_input_events_target with HTMLInputElement.

```cpp
    if (auto* input_element = as_if<HTML::HTMLInputElement>(*focused_area)) {
        if (!for_node || for_node == input_element)
            return input_element;
    }
    if (auto* text_area_element = as_if<HTML::HTMLTextAreaElement>(*focused_area)) {
        if (!for_node || for_node->find_in_shadow_including_ancestry([&](Node const& it) { return &it == focused_area; }))
            return text_area_element;
    }
```
HTMLInputElement has shadow root as well, so the special case above is not only wrong but unnecessary. It should just be:
```cpp
    if (auto* form_element = as_if<HTML::FormAssociatedTextControlElement>(*focused_area)) {
        if (!for_node || for_node->find_in_shadow_including_ancestry([&](Node const& it) { return &it == focused_area; }))
            return form_element;
    }
```

jonbgamble:
Thanks for catching that regression for me.

PR #7234, opened 2025-12-25
LibWeb: Fix clipboard copy in text controls

jonbgamble:
Previously, user selections made inside text controls were not reflected at the document level. This band-aid propagates the new selection offsets all the way to the document range so that copying works.

### DIFF
--- a/Libraries/LibWeb/HTML/FormAssociatedElement.cpp
+++ b/Libraries/LibWeb/HTML/FormAssociatedElement.cpp
@@ -934,6 +934,8 @@ void FormAssociatedTextControlElement::selection_was_changed()
     } else {
         text_paintable->set_selection_state(Painting::Paintable::SelectionState::StartAndEnd);
     }
+    if (auto selection = text_node->document().get_selection())
+        MUST(selection->set_base_and_extent(*text_node, m_selection_start, *text_node, m_selection_end));
     text_paintable->set_needs_display();
 }
 


Libraries/LibWeb/HTML/FormAssociatedElement.cpp
```diff
@@ -934,6 +934,8 @@ void FormAssociatedTextControlElement::selection_was_changed()
     } else {
         text_paintable->set_selection_state(Painting::Paintable::SelectionState::StartAndEnd);
     }
+    if (auto selection = text_node->document().get_selection())
```
shannonbooth, line 937: The commit says that this is a 'bandaid', but from testing, I haven't determined a way in which this behaves different to how other browsers do things as well. Do you know of any way it is behaviourly different? Nice improvement though!

In any case, since there is no clear link to spec requirements here I can figure out at least, I think we should leave an `// AD-HOC: `comment explaining why we do this, and why we consider it a bandaid to aid in understanding in making further improvements to form associated element selection in the future.

Otherwise, could you please add a [LibWeb Text based test](https://github.com/LadybirdBrowser/ladybird/tree/master/Tests/LibWeb/Text/input/HTML) for this so that we don't regress on this in the future? Probably the test could `println` something like `window.getSelection` before and after calling `textarea.select()` on an input or textarea element.

jonbgamble: Band-aid was not a great choice of words, maybe "work in progress" is better. Ignoring spec for the moment, it is clear from other browsers alone that once #7233 is merged (or that underlying issue is otherwise fixed), there will be latent selection restoration issues touching this and adjacent code. This PR is tiny in the spirit of not trying to solve everything at once.

jonbgamble: Do you by chance know how to spoof user-activation from a test script (for navigator.clipboard)?

[Edit] Never mind, I have discovered Internals.cpp

shannonbooth: This test appears to do it through using the 'internals' object which is a JS object that we expose when running tests to do such types of things:

Test: https://github.com/LadybirdBrowser/ladybird/blob/f6f4d7f49797bb320a97ba1fc63764cc80cf5afc/Tests/LibWeb/Text/input/clipboard.html#L21

Internals IDL: https://github.com/LadybirdBrowser/ladybird/blob/master/Libraries/LibWeb/Internals/Internals.idl

Though, if you want to keep things simple, I don't think you necessarily need to also test that the clipboard is working based on the selection, I think it would be enough to just check that the selection did change outside of just the textcontrol/input element. The JS I was playing around with locally while reviewing basically just did something along the lines of:

```js
const selection1 = window.getSelection().toString();
console.log(selection1);

const textarea = document.getElementById('ta');
textarea.focus();
textarea.select();

const selection2 = window.getSelection().toString();
console.log(selection2);
```

Which gives different results before and after your changes here

jonbgamble: Actually, I'm increasingly convinced what I have here is not correct. My current understanding is that the copy handler should first check active element for its selection, returning that if exists, and only then fall back on window/document.

I think it's wrong to live map the textarea/input selection to document range as I have done here, and this would become apparent after 7233. So maybe band-aid was the right choice after all.

PR #7239, opened 2025-12-26
LibWeb/WebIDL: Upgrade QuotaExceededError to its own interface

shannonbooth:
See https://github.com/whatwg/webidl/commit/48eba6d


jonbgamble:
Related, Tests/LibWeb/Text/input/wpt-import/webstorage/storage_local_setitem_quotaexceedederr.window.js takes about 5 minutes to run on my 7950x due to IPC overhead. I'd hate to think what it does to a CI runner. Should someone make a quota test that eats like a shark (instead of a ladybird) until you have some in-process storage caching?

shannonbooth:
> Related, Tests/LibWeb/Text/input/wpt-import/webstorage/storage_local_setitem_quotaexceedederr.window.js takes about 5 minutes to run on my 7950x due to IPC overhead. I'd hate to think what it does to a CI runner. Should someone make a quota test that eats like a shark (instead of a ladybird) until you have some in-process storage caching?

Interesting, it is quite quick (relatively speaking) for me:
```
./Meta/ladybird.py run test-web -f storage_local_setitem_quotaexceedederr  0.50s user 0.24s system 82% cpu 0.906 total
```

jonbgamble:
You're right, that command does run quick. If you load that test page outside the headless wpt-run harness, it's easier to see what I mean about the performance. Although it now seems wrong to attribute that to IPC overhead.

There is something strange going on in test-web as well. If I match the single session storage quota test, it runs quick. If I match all of `Text`, or just do a global run, the session storage test will time out for me.

https://discord.com/channels/1247070541085671459/1247090166909374474

[Edit] Even more fun, local storage quota test timeouts in heavily batched test-web runs only seem to happen (for me at least) on Linux. Note that both Mac and Linux take their sweet time when you load this page in a non-headless environment, but maybe that's just uncoalesced sqlite writes being slow.

PR #7241, opened 2025-12-26
LibWeb: Target scrollable container for keyboard scroll events

ahmed-n-abdeltwab:
## Problem
Keyboard scroll events (arrow keys, Page Up/Down, Home/End) always scrolled the document window, regardless of which element the user had clicked. This caused WPT timeouts in tests that expected keyboard scrolling to target the clicked scrollable container.

## WPT Impact
Fixes timeouts in:
- https://wpt.live/css/css-overflow/scroll-with-ancestor-border-radius.html
- https://wpt.live/css/css-overflow/scroller-covered-by-empty-svg.html

### DIFF
--- a/Libraries/LibWeb/Page/EventHandler.cpp
+++ b/Libraries/LibWeb/Page/EventHandler.cpp
@@ -1719,6 +1719,108 @@ EventResult EventHandler::handle_keydown(UIEvents::KeyCode key, u32 modifiers, u
     auto arrow_key_scroll_distance = 100;
     auto page_scroll_distance = document->window()->inner_height() - (document->window()->outer_height() - document->window()->inner_height());
 
+    // Helper lambda to find the nearest scrollable ancestor of a node (excluding viewport)
+    auto find_scrollable_ancestor = [](DOM::Node const* node) -> Painting::PaintableBox* {
+        if (!node || !node->is_connected())
+            return nullptr;
+
+        // Start from the node's layout node
+        auto* layout_node = node->layout_node();
+        if (!layout_node)
+            return nullptr;
+
+        auto* paintable = layout_node->first_paintable();
+        if (!paintable)
+            return nullptr;
+
+        auto* paintable_box = as_if<Painting::PaintableBox>(paintable);
+        if (!paintable_box)
+            return nullptr;
+
+        if (!paintable_box->is_viewport_paintable() && paintable_box->could_be_scrolled_by_wheel_event())
+            return const_cast<Painting::PaintableBox*>(paintable_box);
+
+        if (auto* scrollable_ancestor = paintable_box->nearest_scrollable_ancestor()) {
+            if (!scrollable_ancestor->is_viewport_paintable())
+                return const_cast<Painting::PaintableBox*>(scrollable_ancestor);
+        }
+
+        return nullptr;
+    };
+
+    // Helper lambda to validate if a scroll target is still valid
+    auto is_scroll_target_valid = [](Painting::PaintableBox const* target) -> bool {
+        if (!target)
+            return false;
+        
+        GC::Ptr<DOM::Node const> node = target->dom_node();
+        if (!node || !node->is_connected())
+            return false;
+
+        if (target->is_viewport_paintable())
+            return false;
+        
+        return target->could_be_scrolled_by_wheel_event();
+    };
+
+    // Helper lambda to find the next nearest scrollable ancestor when current target becomes invalid
+    auto find_next_scrollable_ancestor = [&find_scrollable_ancestor](Painting::PaintableBox const* invalid_target) -> Painting::PaintableBox* {
+        if (!invalid_target)
+            return nullptr;
+
+        GC::Ptr<DOM::Node const> node = invalid_target->dom_node();
+        if (!node)
+            return nullptr;
+
+        auto* parent = node->parent();
+        while (parent) {
+            if (auto* scrollable = find_scrollable_ancestor(parent))
+                return scrollable;
+            parent = parent->parent();
+        }
+
+        return nullptr;
+    };
+
+    // Determine the scroll target with persistent state tracking
+    Painting::PaintableBox* scroll_target = nullptr;
+    GC::Ptr<DOM::Node> current_focused_area = focused_area;
+
+    // Check if focus has changed
+    bool focus_changed = (current_focused_area.ptr() != m_keyboard_scroll_focused_node.ptr());
+
+    if (focus_changed) {
+        // Focus changed - recalculate scroll target
+        m_keyboard_scroll_focused_node = current_focused_area;
+        m_keyboard_scroll_target = find_scrollable_ancestor(current_focused_area.ptr());
+        scroll_target = m_keyboard_scroll_target.ptr();
+    } else {
+        // Focus hasn't changed - try to reuse the previous scroll target
+        scroll_target = m_keyboard_scroll_target.ptr();
+
+        // Validate the cached scroll target
+        if (scroll_target && !is_scroll_target_valid(scroll_target)) {
+            scroll_target = find_next_scrollable_ancestor(scroll_target);
+            m_keyboard_scroll_target = scroll_target;
+        }
+
+        // If we still don't have a valid target and focus hasn't changed, 
+        // but the focused node is gone, clear our state
+        if (!scroll_target && m_keyboard_scroll_focused_node && !m_keyboard_scroll_focused_node->is_connected()) {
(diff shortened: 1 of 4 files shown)


Libraries/LibWeb/Page/EventHandler.cpp
```diff
@@ -1505,6 +1511,12 @@ EventResult EventHandler::handle_keydown(UIEvents::KeyCode key, u32 modifiers, u
         else
             scroll_by(key == UIEvents::KeyCode::Key_Left ? -arrow_key_scroll_distance : arrow_key_scroll_distance, 0);
         return EventResult::Handled;
+    case UIEvents::KeyCode::Key_Space:
+        // Space scrolls down by a page, Shift+Space scrolls up
+        if (modifiers && modifiers != UIEvents::KeyModifier::Mod_Shift)
+            break;
+        scroll_by(0, modifiers & UIEvents::KeyModifier::Mod_Shift ? -page_scroll_distance : page_scroll_distance);
+        return EventResult::Handled;
```
ahmed-n-abdeltwab, line 1855: Here, I'm adding the UI events for the `Key Space`.
Scroll if the space bar was pressed for down, and shift + space for up.

Libraries/LibWeb/Page/EventHandler.cpp
```diff
@@ -1425,15 +1455,48 @@ EventResult EventHandler::handle_keydown(UIEvents::KeyCode key, u32 modifiers, u
...
+
+    // Helper lambda to scroll either the target element or the window
+    auto scroll_by = [&](CSSPixels delta_x, CSSPixels delta_y) {
+        if (scroll_target) {
+            (void)scroll_target->scroll_by(delta_x.to_int(), delta_y.to_int());
+        } else {
+            document->window()->scroll_by(delta_x.to_double(), delta_y.to_double());
+        }
+    };
+
```
ahmed-n-abdeltwab: Here's where I get the primary scrollable container first and fall back to the parent. And if the scroll target is null, then I am scrolling the default.

Libraries/LibWeb/Page/EventHandler.cpp
```diff
@@ -1446,19 +1509,35 @@ EventResult EventHandler::handle_keydown(UIEvents::KeyCode key, u32 modifiers, u
...
     case UIEvents::KeyCode::Key_End:
-        document->window()->scroll_by(0, INT64_MAX);
+        if (scroll_target) {
+            auto overflow_rect = scroll_target->scrollable_overflow_rect();
+            if (overflow_rect.has_value())
+                (void)scroll_target->set_scroll_offset({ 0, overflow_rect->height() });
+            
+        } else {
+            document->window()->scroll_by(0, INT64_MAX);
+        }
```
ahmed-n-abdeltwab, line 1876: Previously, Home/End always scrolled the document window. Now, if the user clicks inside a scrollable container, the Home/End keys should scroll that container instead. If no container was clicked, it falls back to scrolling the document (original behavior).

Libraries/LibWeb/Page/EventHandler.cpp
```diff
@@ -665,6 +665,36 @@ EventResult EventHandler::handle_mousedown(CSSPixelPoint visual_viewport_positio
...
+            if (scrollable_container) {
+                if (auto* parent_scrollable = scrollable_container->nearest_scrollable_ancestor()) {
+                    if (!parent_scrollable->is_viewport_paintable()) {
+                        if (auto parent_node = parent_scrollable->dom_node())
+                            m_keyboard_scroll_container_parent = const_cast<DOM::Node*>(parent_node.ptr());
+                    }
+                }
+            }
+        }
+
```
ahmed-n-abdeltwab: Identifies which scrollable containers should receive keyboard scroll events 
based on the user's click target.

Example hierarchy:
1. Viewport: Excluded (Uses Window::scroll_by).
2. #outer (Scrollable): Stored as m_keyboard_scroll_container_parent.
3. #inner (Scrollable): Stored as m_keyboard_scroll_container.
4. #target (Not scrollable): The specific element clicked.

The reason the viewport  (document body) is excluded here because it is handled 
via Window::scroll_by() rather than PaintableBox::scroll_by(). Skipping 
it ensures the fallback path correctly manages document-level scrolling.

Libraries/LibWeb/Page/EventHandler.h
```diff
@@ -92,6 +92,12 @@ class WEB_API EventHandler {
 
     GC::Weak<DOM::EventTarget> m_mousedown_target;
 
+    // Tracks the scrollable container that should receive keyboard scroll events.
+    // Set when user clicks inside a scrollable element.
+    GC::Ptr<DOM::Node const> m_keyboard_scroll_container;
```
Psychpsyo: No. Scrolling needs to target based on whichever element currently has focus.
This will fail if a user enters a scroll container via tabbing into it with the keyboard, or being focused into it from javascript or any other method that is not strictly clicking on it.

ahmed-n-abdeltwab: What is your suggestion? From what I understood from you I would need to listen to or track all the elements for targeting the screen Wouldn't that affect performance? if i did this? or im wrong?

Psychpsyo: The document already knows what element has focus. (for stuff like matching CSS's `:focus` and all the other input tracking and what not)
You should use that element to figure out what scroll container to scroll.

ahmed-n-abdeltwab: I was initially focused on the mouse scrolling WPT, but thanks to your review, I think this PR will fix more than two tests now! I'll take another look and implement a fix for the problem. thanks

Psychpsyo:
You should squash all of your commits into a single commit.

ahmed-n-abdeltwab:
> You should squash all of your commits into a single commit.

the comment i have is for the lambda functions, explaining what it do. should i remove all the comment?

Psychpsyo:
> the comment i have is for the lambda functions, explaining what it do. should i remove all the comment?

Commits, not comments.
Your changes are currently split up over 10 different commits: https://github.com/LadybirdBrowser/ladybird/pull/7241/commits
This should be just one instead.
<sup>(maybe multiple if this were a bigger PR that does multiple related things, but that's not the case here)</sup>

ahmed-n-abdeltwab:
> > the comment i have is for the lambda functions, explaining what it do. should i remove all the comment?
> 
> Commits, not comments. Your changes are currently split up over 10 different commits: https://github.com/LadybirdBrowser/ladybird/pull/7241/commits This should be just one instead. (maybe multiple if this were a bigger PR that does multiple related things, but that's not the case here)

done

Libraries/LibWeb/Page/EventHandler.cpp
```diff
@@ -1719,6 +1719,108 @@ EventResult EventHandler::handle_keydown(UIEvents::KeyCode key, u32 modifiers, u
...
+            m_keyboard_scroll_target = scroll_target;
+        }
+
+        // If we still don't have a valid target and focus hasn't changed, 
+        // but the focused node is gone, clear our state
+        if (!scroll_target && m_keyboard_scroll_focused_node && !m_keyboard_scroll_focused_node->is_connected()) {
+            m_keyboard_scroll_focused_node = nullptr;
+            m_keyboard_scroll_target = nullptr;
+        }
+    }
```
gmta, line 1813: Conceptually, instead of manually keeping track of the element that was clicked on, isn't this basically the [focusable area](https://html.spec.whatwg.org/multipage/interaction.html#focusable-area)? This is one of the candidates:

> The scrollable regions of elements that are [being rendered](https://html.spec.whatwg.org/multipage/rendering.html#being-rendered) and are not [inert](https://html.spec.whatwg.org/multipage/interaction.html#inert).

Meaning that a scrollable region can become a focused area, at which point it should be the focal point for keyboard input.

ahmed-n-abdeltwab: you are right, so, your concern is that recalculating the scroll target on every key event based solely on the focused element is inefficient, and I should implement this instead? -> validating the scroll target. If a target becomes invalid (the DOM node is disconnected or no longer scrollable), automatically search for the next scrollable ancestor in the DOM hierarchy. am i right?

ahmed-n-abdeltwab: What if the scrollable element doesn't accept direct keyboard focus? For example, the body is scrollable but isn't a traditional input field. what do mean by keyboard input?

PR #7242, opened 2025-12-26
LibWeb: Two boog fixes for StorageEvent broadcasting

shannonbooth:
See commits. Cannot add a test for the second fix without changes to the test harness (which I did, but then ran into conflicts so I reverted adding a test).

### DIFF
--- a/Libraries/LibWeb/HTML/Storage.cpp
+++ b/Libraries/LibWeb/HTML/Storage.cpp
@@ -1,7 +1,7 @@
 /*
  * Copyright (c) 2022, Andreas Kling <andreas@ladybird.org>
  * Copyright (c) 2023, Luke Wilde <lukew@serenityos.org>
- * Copyright (c) 2024-2025, Shannon Booth <shannon@serenityos.org>
+ * Copyright (c) 2024-2026, Shannon Booth <shannon@serenityos.org>
  *
  * SPDX-License-Identifier: BSD-2-Clause
  */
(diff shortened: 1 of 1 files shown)


Tests/LibWeb/Text/expected/wpt-import/webstorage/event_no_duplicates.txt
```diff
@@ -0,0 +1,13 @@
...
+
+8 Pass
+Pass	sessionStorage exists
+Pass	Setting to same value does not trigger event for sessionStorage
+Pass	Deleting non-existent key does not trigger event for sessionStorage
+Pass	Clearing empty storage does not trigger event for sessionStorage
+Pass	localStorage exists
+Pass	Setting to same value does not trigger event for localStorage
+Pass	Deleting non-existent key does not trigger event for localStorage
+Pass	Clearing empty storage does not trigger event for localStorage
```
shannonbooth: Ah, dang, of course - this specific subtest is inherently flakey if any other test is run in parallel with same backing storage. hmm, maybe could locally change this specific test to skip the localStorage case in particular

shannonbooth: Okay, this is fixed now by adding a RunInSerial config option to TestConfig.ini (hopefully at least!)

shannonbooth: Well, now I've got a whole boatload of conflicts after all the changes to test-web for test variants etc. I might just drop adding a test altogether for now, this does not seem worth the effort resolving all of the conflicts

shannonbooth: Test harness changes and test removed, conflicts fixed.

PR #7244, opened 2025-12-26
Another GC leak fix for test-web

awesomekling:
We have a problem where some `HTML::Task` objects get stuck forever in the `HTML::TaskQueue` and never run, keeping entire realms alive.

I tracked down one cause of this issue, which was that we were scheduling tasks in inert temporary documents used for fragment parsing. Such documents by definition never become "fully active", and so all associated tasks will remain non-runnable.

We fix this in an ad-hoc way by just refusing to add such tasks to the task queues.

I've left a FIXME in there about following up on some ongoing spec work that aims to remove these inert temporary documents.

This prevents 5 `Task` leaks when running `test-web`, which were keeping one unnecessary realm alive.

Edit: tacked on another fix that stops leaking 10 more tasks, leaving only one stuck in the queue at the end of `test-web`.

Edit again: This keeps exposing GC issues in CI, so I keep adding patches for them..

### DIFF
--- a/Libraries/LibWeb/HTML/ImageBitmap.cpp
+++ b/Libraries/LibWeb/HTML/ImageBitmap.cpp
@@ -18,7 +18,7 @@ GC_DEFINE_ALLOCATOR(ImageBitmap);
 [[nodiscard]] static auto create_bitmap_from_bitmap_data(Gfx::BitmapFormat const format, Gfx::AlphaType const alpha_type, u32 const width, u32 const height, u32 const pitch, ByteBuffer data)
 {
     // NB: The data is captured by value in the destruction callback lambda to ensure its lifetime.
-    return Gfx::Bitmap::create_wrapper(format, alpha_type, Gfx::IntSize(width, height), pitch, data.data(), [data] { });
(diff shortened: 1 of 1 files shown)


Libraries/LibWeb/Bindings/MainThreadVM.cpp
```diff
@@ -725,7 +725,7 @@ void queue_mutation_observer_microtask(DOM::Document const& document)
     // 3. Queue a microtask to notify mutation observers.
     // NOTE: This uses the implied document concept. In the case of mutation observers, it is always done in a node
     //       context, so document should be that node's document.
```
Zaggy1024: Very unfamiliar with this code, but is the comment relevant here anymore?

Libraries/LibWeb/Bindings/MainThreadVM.cpp
```diff
@@ -725,7 +725,7 @@ void queue_mutation_observer_microtask(DOM::Document const& document)
     // 3. Queue a microtask to notify mutation observers.
     // NOTE: This uses the implied document concept. In the case of mutation observers, it is always done in a node
```
shannonbooth: I don't quite understand what this comment means, but I suppose it is probably no longer relevant?

shannonbooth: whoops, double up

PR #7247, opened 2025-12-27
LibGC: Make GC graph construction consistent with marking

kalenikaliaksandr:
Couple of fixes to make GC-dump reflect what would happen in marking more precisely.

### DIFF
--- a/Libraries/LibGC/Heap.cpp
+++ b/Libraries/LibGC/Heap.cpp
@@ -165,7 +165,7 @@ public:
             if (m_node_being_visited)
                 m_node_being_visited->edges.set(reinterpret_cast<FlatPtr>(cell));
 
-            if (m_graph.get(reinterpret_cast<FlatPtr>(&cell)).has_value())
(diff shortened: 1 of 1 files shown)


Libraries/LibGC/Heap.cpp
```diff
@@ -439,6 +442,11 @@ NO_SANITIZE_ADDRESS void Heap::gather_conservative_roots(HashMap<Cell*, HeapRoot
     HashTable<HeapBlock*> all_live_heap_blocks;
     for_each_block([&](auto& block) {
         all_live_heap_blocks.set(&block);
+        block.template for_each_cell_in_state<Cell::State::Live>([&](Cell* cell) {
+            if (cell_must_survive_garbage_collection(*cell)) {
+                roots.set(cell, HeapRoot { .type = HeapRoot::Type::MustSurviveGC });
+            }
+        });
```
awesomekling: These are not *conservative* roots, since we know for sure they are real cells! So this should be in `gather_roots()` instead :)

PR #7248, opened 2025-12-27
LibWeb: Implement Screen.isExtended attribute

adamjcolvin:
## Summary

Implement the `isExtended` attribute on Screen per the Window Management API specification.

The getter returns `true` if the device has multiple screens, after checking that the document is allowed to use the "window-management" policy-controlled feature.

Spec: https://w3c.github.io/window-management/#dom-screen-isextended

## Implementation Approach

This implementation leverages the existing `set_screen_rects()` IPC call that already transmits screen rectangles from the UI process to WebContent. Rather than adding new IPC infrastructure, we simply store the screen count when this existing call is made:

```cpp
void set_screen_rects(Vector<Web::DevicePixelRect, 4> const& rects, size_t main_screen_index)
{
    m_screen_rect = rects[main_screen_index];
    m_screen_count = rects.size();  // Just store the count
}

...

virtual size_t screen_count() const override { return m_screen_count; }
```

The `Screen::is_extended()` getter then checks the permission and queries the screen count:

```cpp
bool Screen::is_extended() const
{
    if (!window().associated_document().is_allowed_to_use_feature(DOM::PolicyControlledFeature::WindowManagement))
        return false;

    return window().page().client().screen_count() > 1;
}
```

## Changes
- **LibWeb/DOM/Document.h**: Add `WindowManagement` to `PolicyControlledFeature` enum
- **LibWeb/DOM/Document.cpp**: Handle `WindowManagement` in `is_allowed_to_use_feature()`
- **LibWeb/Page/Page.h**: Add `screen_count()` pure virtual method to PageClient
- **LibWeb/SVG/SVGDecodedImageData.h**: Implement `screen_count()` override
- **Services/WebWorker/PageHost.h**: Implement `screen_count()` override
- **Services/WebContent/PageClient.h**: Store screen count, implement override
- **LibWeb/CSS/Screen.cpp**: Replace FIXME stub with spec-compliant implementation

## Limitations

- AppKit does not currently detect screen configuration changes dynamically (pre-existing FIXME)
- Permission check returns `true` by default pending full Permissions Policy implementation

### DIFF
--- a/Libraries/LibWeb/CSS/Screen.cpp
+++ b/Libraries/LibWeb/CSS/Screen.cpp
@@ -95,8 +95,13 @@ GC::Ref<ScreenOrientation> Screen::orientation()
 // https://w3c.github.io/window-management/#dom-screen-isextended
 bool Screen::is_extended() const
 {
-    dbgln("FIXME: Unimplemented Screen::is_extended");
-    return false;
+    // 1. If this's relevant global object's associated Document is not allowed to use
+    //    the policy-controlled feature named "window-management", return false.
+    if (!window().associated_document().is_allowed_to_use_feature(DOM::PolicyControlledFeature::WindowManagement))
+        return false;
+
+    // 2. Return true if the device has more than one screen, and false otherwise.
+    return window().page().client().screen_count() > 1;
 }
 
 // https://w3c.github.io/window-management/#dom-screen-onchange
--- a/Libraries/LibWeb/Page/Page.h
+++ b/Libraries/LibWeb/Page/Page.h
@@ -341,6 +341,7 @@ public:
     virtual CSS::PreferredColorScheme preferred_color_scheme() const = 0;
     virtual CSS::PreferredContrast preferred_contrast() const = 0;
     virtual CSS::PreferredMotion preferred_motion() const = 0;
+    virtual size_t screen_count() const = 0;
     virtual Queue<QueuedInputEvent>& input_event_queue() = 0;
     virtual void report_finished_handling_input_event(u64 page_id, EventResult event_was_handled) = 0;
     virtual void page_did_change_title(Utf16String const&) { }
--- a/Services/WebContent/PageClient.h
+++ b/Services/WebContent/PageClient.h
@@ -53,7 +53,11 @@ public:
 
     void set_palette_impl(Gfx::PaletteImpl&);
     void set_viewport_size(Web::DevicePixelSize const&);
-    void set_screen_rects(Vector<Web::DevicePixelRect, 4> const& rects, size_t main_screen_index) { m_screen_rect = rects[main_screen_index]; }
+    void set_screen_rects(Vector<Web::DevicePixelRect> const& rects, size_t main_screen_index)
+    {
+        m_all_screen_rects = rects;
+        m_main_screen_index = main_screen_index;
+    }
     void set_device_pixel_ratio(double device_pixel_ratio) { m_device_pixel_ratio = device_pixel_ratio; }
     void set_zoom_level(double zoom_level) { m_zoom_level = zoom_level; }
     void set_maximum_frames_per_second(u64 maximum_frames_per_second);
@@ -110,7 +114,8 @@ private:
     virtual bool is_url_suitable_for_same_process_navigation(URL::URL const& current_url, URL::URL const& target_url) const override;
     virtual void request_new_process_for_navigation(URL::URL const&) override;
     virtual Gfx::Palette palette() const override;
-    virtual Web::DevicePixelRect screen_rect() const override { return m_screen_rect; }
+    virtual Web::DevicePixelRect screen_rect() const override { return m_all_screen_rects[m_main_screen_index]; }
+    virtual size_t screen_count() const override { return m_all_screen_rects.size(); }
     virtual Web::CSS::PreferredColorScheme preferred_color_scheme() const override { return m_preferred_color_scheme; }
     virtual Web::CSS::PreferredContrast preferred_contrast() const override { return m_preferred_contrast; }
     virtual Web::CSS::PreferredMotion preferred_motion() const override { return m_preferred_motion; }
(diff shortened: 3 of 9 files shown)


Libraries/LibWeb/Page/Page.h
```diff
@@ -334,6 +334,7 @@ class PageClient : public JS::Cell {
     virtual CSS::PreferredColorScheme preferred_color_scheme() const = 0;
     virtual CSS::PreferredContrast preferred_contrast() const = 0;
     virtual CSS::PreferredMotion preferred_motion() const = 0;
+    virtual bool is_screen_extended() const { return false; }
```
AtkinsSJ: Leave this as `= 0;` so that subclasses are forced to implement it.

Libraries/LibWeb/CSS/Screen.cpp
```diff
@@ -95,8 +95,13 @@ GC::Ref<ScreenOrientation> Screen::orientation()
...
 {
-    dbgln("FIXME: Unimplemented Screen::is_extended");
-    return false;
+    // 1. If this's relevant global object's associated Document is not allowed to use
+    //    the policy-controlled feature named "window-management", return false.
+    if (!window().associated_document().is_allowed_to_use_feature(DOM::PolicyControlledFeature::WindowManagement))
+        return false;
+
+    // 2. Return true if the device has more than one screen, and false otherwise.
+    return window().page().client().is_screen_extended();
```
AtkinsSJ: Nit: Maybe it's easier to follow if we ask the client how many screens there are instead? But this is the only user so it doesn't make a big difference.

adamjcolvin: Yeah that's cleaner I think. I've renamed the method to `screen_count()` and moved the > 1 comparison into `Screen.cpp` here. Let me know what you think.

Libraries/LibWeb/Page/Page.h
adamjcolvin: That makes sense, I've changed it and added implementations to both `SVGPageClient` and `PageHost`, both returning 1.

adamjcolvin:
> Welcome to Ladybird! :^) Nice change.

Thank you! 🙂 Great to be here!

Services/WebContent/PageClient.h
```diff
@@ -53,7 +53,11 @@ class PageClient final : public Web::PageClient {
 
     void set_palette_impl(Gfx::PaletteImpl&);
     void set_viewport_size(Web::DevicePixelSize const&);
-    void set_screen_rects(Vector<Web::DevicePixelRect, 4> const& rects, size_t main_screen_index) { m_screen_rect = rects[main_screen_index]; }
+    void set_screen_rects(Vector<Web::DevicePixelRect, 4> const& rects, size_t main_screen_index)
+    {
+        m_screen_rect = rects[main_screen_index];
+        m_screen_count = rects.size();
```
gmta: Both the old and new implementations of this feel a bit weird - we're passing in a vector of screen rects and an index, only to throw away everything but the main screen rect.

Maybe we should store all the screen rects that are passed in _and_ create a new field to keep track of the main screen? That way, we don't have to keep track of the screen count and are future proofing `PageClient` for when we do need to handle multiple screen rects.

gmta (commented):
Welcome!

Services/WebContent/PageClient.h
adamjcolvin: I see what you mean. So what if we had this instead: 

```cpp
void set_screen_rects(Vector<Web::DevicePixelRect, 4> const& rects, size_t main_screen_index)
{
    m_screen_rect = rects[main_screen_index];
    m_all_screen_rects = rects; 
}
```

Then we could update the `screen_count()` implementation to be: 
```cpp
virtual size_t screen_count() const override { return m_all_screen_rects.size(); }
``` 

What do you think?

gmta: Instead of storing the `m_screen_rect`, you could just store the index of the main screen rect and index into `m_all_screen_rects`?

adamjcolvin: Ah that makes sense. Ok I've changed the setter to this now, setting `m_all_screen_rects` and `m_main_screen_index` instead: 

```cpp
    void set_screen_rects(Vector<Web::DevicePixelRect, 4> const& rects, size_t main_screen_index)
    {
        m_all_screen_rects = rects;
        m_main_screen_index = main_screen_index;
    }
```

And the `screen_rect()` getter now uses both of them to return the main screen: 

```cpp
virtual Web::DevicePixelRect screen_rect() const override { return m_all_screen_rects[m_main_screen_index]; }
```

What do you think?

Services/WebContent/PageClient.h
```diff
@@ -53,7 +53,11 @@ class PageClient final : public Web::PageClient {
 
     void set_palette_impl(Gfx::PaletteImpl&);
     void set_viewport_size(Web::DevicePixelSize const&);
-    void set_screen_rects(Vector<Web::DevicePixelRect, 4> const& rects, size_t main_screen_index) { m_screen_rect = rects[main_screen_index]; }
+    void set_screen_rects(Vector<Web::DevicePixelRect, 4> const& rects, size_t main_screen_index)
+    {
+        m_all_screen_rects = rects;
+        m_main_screen_index = main_screen_index;
```
gmta, line 59: Your commit message no longer reflects the code (e.g. you no longer store the screen count in PageClient)

Services/WebContent/PageClient.h
```diff
@@ -193,7 +198,8 @@ class PageClient final : public Web::PageClient {
     PageHost& m_owner;
     GC::Ref<Web::Page> m_page;
     RefPtr<Gfx::PaletteImpl> m_palette_impl;
-    Web::DevicePixelRect m_screen_rect;
+    Vector<Web::DevicePixelRect, 4> m_all_screen_rects { Web::DevicePixelRect {} };
```
gmta: Not sure if you need the inline size of `4` here at all; as long as we're not limited in performance on the setter and moving vector around, I think we'll do just fine with a `Vector<Web::DevicePixelRect>`.

adamjcolvin: No problem, I've updated it to just `Vector<Web::DevicePixelRect>`

Services/WebContent/PageClient.h
adamjcolvin: Ah of course, thank you. I've updated it.

Issue #7245, opened 2025-12-26
Recursive module script exeuction on Shopify-created page


ADKaster:
### Summary

When loading https://vainartisan.com, we crash at the assertion in 

https://github.com/LadybirdBrowser/ladybird/blob/f27847274c379b35977dcdf4bd25559a824f51dd/Libraries/LibWeb/HTML/HTMLScriptElement.cpp#L174-L181

This assertion checks that we have not already started executing a script before executing the script referred to by the current script element. 

### Operating system

Linux

### Steps to reproduce

1. Load https://vainartisan.com
2. 💥 

### Expected behavior

Load the page

### Actual behavior

Crash

### URL for a reduced test case

N/A

### HTML/SVG/etc. source for a reduced test case

```html
N/A
```

### Log output and (if possible) backtrace

```shell
4767543.149 WebContent(4141056): FIXME: Handle null navigation_request in navigation response Content Security Policy check.
4767543.745 ImageDecoder(4141050:4141106): JPEG error: Improper call to JPEG library in state 203
4767543.753 ImageDecoder(4141050:4141106): JPEG error: Improper call to JPEG library in state 203
4767543.767 ImageDecoder(4141050:4141106): JPEG error: Improper call to JPEG library in state 203
4767543.776 ImageDecoder(4141050:4141106): JPEG error: Improper call to JPEG library in state 203
4767543.780 ImageDecoder(4141050:4141106): JPEG error: Improper call to JPEG library in state 203
4767543.782 ImageDecoder(4141050:4141106): JPEG error: Improper call to JPEG library in state 203
4767543.783 ImageDecoder(4141050:4141106): JPEG Warning: Decoding produced no more scanlines in scanline 190/1482.
VERIFICATION FAILED: document->current_script() == nullptr at /home/andrew/ladybird-org/ladybird-browser/Libraries/LibWeb/HTML/HTMLScriptElement.cpp:177
Stack trace (most recent call first):
#0  0x0000738b1234fec6 in execute_script at /home/andrew/ladybird-org/ladybird-browser/Libraries/LibWeb/HTML/HTMLScriptElement.cpp:177:9
#1  (inlined)          in operator() at /home/andrew/ladybird-org/ladybird-browser/Libraries/LibWeb/HTML/HTMLScriptElement.cpp:546:17
#2  0x0000738b1235399c in call at /home/andrew/ladybird-org/ladybird-browser/AK/Function.h:225:20
[...]

PR #7252, opened 2025-12-27
LibWeb: Guard currentScript around module execution

TurboRigby:
- Ensure document.currentScript is null while running module scripts, matching spec semantics and avoiding reentrant assertion failures.
 - Restore the previous currentScript after module execution using a scope guard.
 - Add a crash test covering module script insertion from a classic script.
 - Fixes https://github.com/LadybirdBrowser/ladybird/issues/7245

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLScriptElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLScriptElement.cpp
@@ -6,7 +6,6 @@
  */
 
 #include <AK/Debug.h>
-#include <AK/ScopeGuard.h>
 #include <AK/StringBuilder.h>
 #include <LibTextCodec/Decoder.h>
 #include <LibWeb/Bindings/HTMLScriptElementPrototype.h>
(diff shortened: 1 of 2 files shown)


TurboRigby:
oops, i forgot to turn off copilot review

TurboRigby:
removed merge commit

Libraries/LibWeb/HTML/HTMLScriptElement.cpp
```diff
@@ -173,8 +174,11 @@ void HTMLScriptElement::execute_script()
     }
     // -> "module"
     else if (m_script_type == ScriptType::Module) {
-        // 1. Assert: document's currentScript attribute is null.
-        VERIFY(document->current_script() == nullptr);
+        // https://html.spec.whatwg.org/multipage/scripting.html#execute-the-script-block
+        // document.currentScript only applies to classic scripts; module scripts must observe null.
```
ADKaster: where does this statement come from?

The comment `// 1. Assert: document's currentScript attribute is null.` is not there for fun, it is a spec step. It's directly pulled from the spec algorithm linked by  <https://html.spec.whatwg.org/multipage/scripting.html#execute-the-script-block>.

Removing it and replacing the code with a workaround would require at least a link to an HTML spec issue, if you believe that the spec is wrong here. But we would need some serious thought and explanation as to why it should be allowed in this case.

TurboRigby: Thanks for the review. I reverted the workaround and kept the spec assert. The fix now queues mark_as_ready for external module scripts as a networking element task (matching the inline-module path), so module execution happens after the classic-script microtask checkpoint. Added regression test: Tests/LibWeb/Crash/HTML/module-script-exec-after-classic.html. Please take another look.

PR #7255, opened 2025-12-28
LibWeb: Avoid assembling coordinated value lists in more cases

Calme1709:
Assembling coordinated value lists is quite expensive since it involves a bunch of allocations for `HashMap`s and `Vector`s - we should probably make it more performant but as a first step we can just do it less often, decreases runtime spent in `assemble_coordinated_value_list` when loading https://en.wikipedia.org/wiki/2023_in_American_television from ~2% to ~0.02%.

Also includes a (mostly) unrelated refactor to clean up computation of transitions - see individual commits for details.


AtkinsSJ (approved):
Nice!

PR #7196, opened 2025-12-20
LibWeb: Implement `ScrollTimeline`

Calme1709:
Implements `ScrollTimeline` as well as some more general changes to animations in order to support progress-based/non-monotonic timelines - see individual commits for details.

Relies on #7174 so marked as draft until that is merged.

https://github.com/user-attachments/assets/fc81ec6f-9f57-4e05-a439-5a70a75ecb53

PR #7258, opened 2025-12-28
LibWeb: Update Animations in InternalAnimationTimeline::setTime()

adamjcolvin:
## The Problem

This fixes a race condition that caused the flaky test failure in `Tests/LibWeb/Text/input/WebAnimations/misc/animation-single-iteration-no-repeat.html`.

When tests manually set the timeline's time via `setTime()`, the timeline's current time changes immediately, but `animation->update()` isn't called until the next event loop frame. This creates a race window where style recalculation can call `pause()` on an animation that's already finished (by time) but hasn't yet processed its finished state.

The sequence that causes the bug:

1. Test calls `timeline.setTime(1000)` — animation is now "finished" by time
2. `play_state()` returns "Finished" (computed from current time)
3. But `animation->update()` hasn't run yet, so `m_is_finished` is still false
4. Style recalculation triggers `apply_animation_properties()` in `StyleComputer.cpp`
5. This sees CSS says "paused" but animation's `play_state()` is "Finished", so it calls `Animation::pause()`
6. Inside `pause()`, the flag `m_pending_pause_task` is set to `Scheduled`
7. Now `play_state()` returns "Paused" instead of "Finished"
8. When `finish_notification_steps` runs, it checks `play_state()` — sees "Paused", exits early
9. The `onfinish` event is never dispatched

## The Solution

In `InternalAnimationTimeline::set_time()`, after setting the timeline's current time, also call `animation->update()` for all associated animations. This ensures the animation's finished state is processed immediately, eliminating the race window

### DIFF
--- a/Libraries/LibWeb/Internals/InternalAnimationTimeline.cpp
+++ b/Libraries/LibWeb/Internals/InternalAnimationTimeline.cpp
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
+#include <LibWeb/Animations/Animation.h>
 #include <LibWeb/Bindings/InternalAnimationTimelinePrototype.h>
 #include <LibWeb/Bindings/Intrinsics.h>
 #include <LibWeb/DOM/Document.h>
@@ -22,6 +23,16 @@ void InternalAnimationTimeline::update_current_time(double)
 void InternalAnimationTimeline::set_time(Optional<double> time)
 {
     set_current_time(time.map([](double value) -> Animations::TimeValue { return { Animations::TimeValue::Type::Milliseconds, value }; }));
+
+    // https://drafts.csswg.org/web-animations-1/#animation-frame-loop
+    // Note: Due to the hierarchical nature of the timing model, updating the current time of a timeline also involves:
+    // - Updating the current time of any animations associated with the timeline.
+    // - Running the update an animation's finished state procedure for any animations whose current time has been
+    //   updated.
+    // - Queueing animation events for any such animations.
+    // NB: This mirrors what the event loop does for DocumentTimeline in Document::update_animations_and_send_events().
+    for (auto const& animation : associated_animations())
+        animation->update();
 }
 
 InternalAnimationTimeline::InternalAnimationTimeline(JS::Realm& realm)


konradekk:
use `// AD-HOC:` comment instead: the content of comments are snippets taken _verbatim_ from to the docs linked before the function signature (if not stated otherwise, e.g. through the comment annotation i’ve given or similar one like `NB`). refer to the `CONTRIBUTING.md` for details.

adamjcolvin:
> use `// AD-HOC:` comment instead: the content of comments are snippets taken _verbatim_ from to the docs linked before the function signature (if not stated otherwise, e.g. through the comment annotation i’ve given or similar one like `NB`). refer to the `CONTRIBUTING.md` for details.

Ah makes sense, thank you @konradekk. I've updated the comment here https://github.com/LadybirdBrowser/ladybird/pull/7258/commits/bb5448681cf53b5bb4e51505293e46b766c60806

konradekk:
nice! i’ll just update the reference (failed to provide it precisely above): [FAQ.md](https://github.com/LadybirdBrowser/ladybird/blob/953b33d273849403741b7ec5ffd197509b22daf2/Documentation/FAQ.md?plain=1#L95).

Calme1709:
I'm not sure I agree with your diagnosis here - my reading of the code is that since the call to `update_finished_state()` comes from `update()` the `synchronously_notify` flag will be `Yes` and thus the `finish_notification_steps` function will run immediately at step 5.2 before any style updates can change the play state.

adamjcolvin:
@Calme1709 That's true, but the problem seems to be that we're already in an incorrect state before `animation->update()` is called.

In the test, when we call `timeline.setTime(1000);`, the animation's `play_state()` will return `Finished` immediately because the full duration has then elapsed. But `animation->update()` hasn't been called yet, and won't be until the next frame. That window is what gives us the flaky test.

For example, in [EventLoop.cpp](https://github.com/LadybirdBrowser/ladybird/blob/master/Libraries/LibWeb/HTML/EventLoop/EventLoop.cpp), line [428](https://github.com/LadybirdBrowser/ladybird/blob/c8b30f5bb82178d2c82eeba07d0e02eaa7b9414b/Libraries/LibWeb/HTML/EventLoop/EventLoop.cpp#L428) runs the animation frame callbacks—this is where the test's JavaScript runs and sets the time. Then on line [442](https://github.com/LadybirdBrowser/ladybird/blob/c8b30f5bb82178d2c82eeba07d0e02eaa7b9414b/Libraries/LibWeb/HTML/EventLoop/EventLoop.cpp#L442), we call `document->update_layout()`, which triggers style recalculation. This can call `apply_animation_properties()` and cause us to call `pause()` on an animation that's already finished.

Calling `pause()` sets `m_pending_pause_task`, which causes `play_state()` to return `Paused` instead of `Finished`. So when `animation->update()` finally does run, and calls `finish_notification_steps,` it checks `play_state()`, sees `Paused`, and then aborts—the finish event is never dispatched.

Calme1709:
That makes sense to me.

Based on that explanation this issue seems like it would be limited to just `InternalAnimationTImeline`s since other timelines (`DocumentTimeline` and `ScrollTimeline` implemented in #7196) only ever update their current time immediately before `update()` is called for each of their associated animations (see `Document::update_animations_and_send_events`) meaning there isn't any chance for style re-computation in-between.

Given that `InternalAnimationTimeline` is not to spec we should probably fix any incompatibilities with the spec on that end, buffering time updates from `InternalAnimationTimeline::set_time()` and only applying them when `InternalAnimationTimeline::update_current_time()` is called seems like it would resolve this issue.

adamjcolvin:
@Calme1709 That's a fair point—having `InternalAnimationTimeline` behave in a similar way to `DocumentTimeline` and `ScrollTimeline` would likely avoid this window. 

That said, based on your own changes in https://github.com/LadybirdBrowser/ladybird/pull/7196, that would likely be a larger and wider-reaching change than this. I think this fix has value purely as a defensive measure, to fix a flaky test. Preventing `pause()` from affecting a finished animation seems semantically correct, regardless of how the timeline is implemented. We could add this fix first, and then follow up with an `InternalAnimationTimeline` improvement—what do you think?

Calme1709:
I'm not sure I agree with your point that preventing pausing of finished animations is correct. I'm not aware of anywhere that is outlined in the spec and major browsers seem to allow it, see the following code which prints "paused" in both Chrome and Firefox (but not Ladybird after your change):

```html
<!DOCTYPE html>
<html>
    <div id="foo"></div>
    <script>
        const anim = foo.animate([{ color: "red" }], 1000);

        anim.finish();
        anim.pause();

        console.log(anim.playState);
    </script>
</html>
```

adamjcolvin:
@Calme1709 I see what you mean, that's a good point. Let me know what you think of the change [here](https://github.com/LadybirdBrowser/ladybird/pull/7258/changes/c94cebd9cff3c7fb154f7f4b12de1c61ce733c54). This is now calling `animation->update()` inside of `setTime()`, so that all required updates happen straight away. That should fix the flaky test and keep the pause behaviour consistent with other browsers.

Libraries/LibWeb/Internals/InternalAnimationTimeline.cpp
```diff
@@ -22,6 +23,12 @@ void InternalAnimationTimeline::update_current_time(double)
 void InternalAnimationTimeline::set_time(Optional<double> time)
 {
     set_current_time(time.map([](double value) -> Animations::TimeValue { return { Animations::TimeValue::Type::Milliseconds, value }; }));
+
+    // Per the Web Animations spec, updating the current time of a timeline also involves updating the current time of
```
Calme1709: Please copy the relevant spec text verbatim and include a link, you can add an `NB:` comment for any non-spec info you want to include

```c++
// https://drafts.csswg.org/web-animations-1/#animation-frame-loop
// ...spec text...
// NB: This mirrors what the event loop does for DocumentTimeline in Document::update_animations_and_send_events().
```

Calme1709 (changes requested):

Libraries/LibWeb/Internals/InternalAnimationTimeline.cpp
adamjcolvin: No problem, updated [here](https://github.com/LadybirdBrowser/ladybird/pull/7258/commits/0bf08793f61f441b336e38adb4a38aa06ace2d0e)

Calme1709:
I haven't run the test suite and this will obviously require a maintainer to review, but that looks good to me now.

shannonbooth (approved):

PR #7265, opened 2025-12-29
LibWeb: Create absolute container for `will-change: position`

mikewatt:
This fixes:
http://wpt.live/css/css-will-change/will-change-abspos-cb-001.html

### DIFF
--- a/Libraries/LibWeb/Layout/Node.cpp
+++ b/Libraries/LibWeb/Layout/Node.cpp
@@ -87,9 +87,6 @@ bool Node::computed_values_establish_absolute_positioning_containing_block() con
 {
     auto const& computed_values = this->computed_values();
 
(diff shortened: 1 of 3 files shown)


Libraries/LibWeb/Layout/Node.cpp
```diff
@@ -97,6 +94,11 @@ bool Node::establishes_an_absolute_positioning_containing_block() const
         return computed_values.will_change().has_property(property_id);
     };
 
+    // https://drafts.csswg.org/css-position/#position-property
+    // Values other than static establish an absolute positioning containing block for its descendants
```
Psychpsyo: This should quote the spec verbatim, like so:
```
// Values other than 'static' make the box a positioned box, and cause it to establish an absolute positioning containing block for its descendants.
```

PR #7269, opened 2025-12-30
LibCore: Only wake the event loop in deferred_invoke from other threads

Zaggy1024:
For some reason, writing to the wake fd from the same thread in `deferred_invoke` was causing a deadlock. However, we don't actually need to wake from the same thread, since the event loop is not waiting and will therefore process the `deferred_invoke` on the next iteration.

This issue was introduced in 3742138cc3318842c795f3be8e41642702428af4. The deadlock could be reproduced consistently by increasing `LOCAL_STORAGE_QUOTA` in `StorageJar.cpp` to a large value like 50 MiB.

### DIFF
--- a/Libraries/LibCore/EventLoopImplementation.cpp
+++ b/Libraries/LibCore/EventLoopImplementation.cpp
@@ -26,7 +26,8 @@ EventLoopImplementation::~EventLoopImplementation() = default;
 void EventLoopImplementation::deferred_invoke(Function<void()>&& invokee)
 {
     m_thread_event_queue.deferred_invoke(move(invokee));
-    wake();
+    if (&m_thread_event_queue != ThreadEventQueue::current_or_null())
+        wake();
 }
 
 static EventLoopManager* s_event_loop_manager = nullptr;


Libraries/LibCore/EventLoopImplementation.cpp
```diff
@@ -26,7 +26,8 @@ EventLoopImplementation::~EventLoopImplementation() = default;
 void EventLoopImplementation::deferred_invoke(Function<void()>&& invokee)
 {
     m_thread_event_queue.deferred_invoke(move(invokee));
-    wake();
+    if (&m_thread_event_queue != ThreadEventQueue::current_or_null())
+        wake();
```
shannonbooth, line 30: I have not looked deeply in the event loop before, but is it guaranteed that we don't hit poll again before running any of the deferred_invoke tasks?

Since otherwise I suppose we might also deadlock no other fd's in the poll loop have any event on them before we hit the poll loop, and the deferred tasks are not run by that point we get stuck at this point: https://github.com/LadybirdBrowser/ladybird/blob/89dbdd3411a3037ec31dcd96c15a47f062c7f48f/Libraries/LibCore/EventLoopImplementationUnix.cpp#L353 not ever running the deferred tasks

Seems like we might hit that if you deferred_invoke and then immediately pump: https://github.com/LadybirdBrowser/ladybird/blob/89dbdd3411a3037ec31dcd96c15a47f062c7f48f/Libraries/LibCore/EventLoopImplementationUnix.cpp#L307

shannonbooth:
Maybe it is also possible to write a regression test for this with the theory of a repro being caused by repeated calls to deferred_invoke in a loop filling up the wake pipe

Libraries/LibCore/EventLoopImplementation.cpp
shannonbooth: Ah, nope not an issue timeout is set to 0 in that case

Zaggy1024: Right, we should end up with a timeout of zero and process the deferred functions almost immediately.

Zaggy1024:
> Maybe it is also possible to write a regression test for this with the theory of a repro being caused by repeated calls to deferred_invoke in a loop filling up the wake pipe

If we're writing one byte per wake, I think that might require a pretty ridiculous (and arbitrary?) number of `deferred_invoke()` calls, so I'm not sure if that's something that would work without hitting a timeout in Sanitizer builds. :thinking: 

Then again, if that's the case, we're probably growing the event queue to pretty large sizes in order to hit that blocking write, so maybe I should add a warning when the queue grows too large to look into this further...

PR #7271, opened 2025-12-30
LibWeb: Treat near-zero aspect-ratios as degenerate

Gingeh:
Prevents a crash in https://wpt.live/css/css-sizing/aspect-ratio/small-aspect-ratio-crash.html

### DIFF
--- a/Libraries/LibWeb/Layout/Box.cpp
+++ b/Libraries/LibWeb/Layout/Box.cpp
@@ -91,7 +91,12 @@ Optional<CSSPixelFraction> Box::preferred_aspect_ratio() const
     if (ratio.is_degenerate())
         return {};
 
-    return CSSPixelFraction(ratio.numerator(), ratio.denominator());
+    auto fraction = CSSPixelFraction(ratio.numerator(), ratio.denominator());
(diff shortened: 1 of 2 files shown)


Libraries/LibWeb/Layout/Box.cpp
```diff
@@ -91,7 +91,11 @@ Optional<CSSPixelFraction> Box::preferred_aspect_ratio() const
     if (ratio.is_degenerate())
         return {};
 
-    return CSSPixelFraction(ratio.numerator(), ratio.denominator());
+    auto fraction = CSSPixelFraction(ratio.numerator(), ratio.denominator());
+    if (fraction == 0) // ratio.is_degenerate() operates on doubles while CSSPixelFraction uses CSSPixels, so we need to check again here.
```
gmta: Can you put this comment above the check and/or wrap it at column 120?

gmta (commented):
Nice find!

Libraries/LibWeb/Layout/Box.cpp
Gingeh: done :saluting_face:

PR #7273, opened 2025-12-30
LibWeb: Declare namespace prefixes in parse_xml_fragment

Gingeh:
Improves https://wpt.live/html/webappapis/dynamic-markup-insertion/the-innerhtml-property/innerhtml-and-xml-namespaces.svg: CRASH -> 12/12 PASS

### DIFF
--- a/Libraries/LibWeb/DOM/Node.cpp
+++ b/Libraries/LibWeb/DOM/Node.cpp
@@ -2279,6 +2279,71 @@ bool Node::is_equal_node(Node const* other_node) const
     return true;
 }
 
+Vector<FlyString> Node::get_in_scope_prefixes() const
+{
+    // https://html.spec.whatwg.org/multipage/xhtml.html#parsing-xhtml-fragments
+    // "A namespace prefix is in scope if the DOM lookupNamespaceURI() method on the element would return a non-null value for that prefix."
+
+    Vector<FlyString> prefixes;
+    HashTable<FlyString> seen_prefixes;
+
+    auto add_prefix = [&](FlyString const& prefix) {
+        if (!seen_prefixes.contains(prefix)) {
+            prefixes.append(prefix);
+            seen_prefixes.set(prefix);
+            VERIFY(lookup_namespace_uri(prefix.to_string()).has_value());
+        }
+    };
+
+    add_prefix("xml"_fly_string);
+    add_prefix("xmlns"_fly_string);
+
+    Element const* current = nullptr;
+
+    if (is<Element>(*this)) {
+        current = static_cast<Element const*>(this);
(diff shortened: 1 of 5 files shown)


Libraries/LibWeb/DOM/Node.cpp
```diff
@@ -2279,6 +2279,64 @@ bool Node::is_equal_node(Node const* other_node) const
...
 
+Vector<String> Node::get_in_scope_prefixes() const
+{
+    // https://html.spec.whatwg.org/multipage/xhtml.html#parsing-xhtml-fragments
+    // "A namespace prefix is in scope if the DOM lookupNamespaceURI() method on the element would return a non-null value for that prefix."
+
+    // NB: This implementation omits the empty prefix because it is handled specially and the "xmlns" prefix because it is illegal to declare.
+
+    Vector<String> prefixes;
+    HashTable<String> seen_prefixes;
```
shannonbooth: This feels like it should be using FlyString over String here especially since what we are adding is already a FlyString.

Also maybe it makes sense to have a lambda helper or something for adding to seen_prefixes/prefixes together  alongside the uniqueness check?

Libraries/LibWeb/XML/XMLFragmentParser.cpp
```diff
@@ -32,15 +32,29 @@ WebIDL::ExceptionOr<Vector<GC::Root<DOM::Node>>> XMLFragmentParser::parse_xml_fr
...
     feed.append('<');
     feed.append(qualified_name);
-    feed.append('>');
-
-    // FIXME:
     //  declaring all the namespace prefixes that are in scope on that element in the DOM,
-
-    // FIXME:
+    for (auto const& prefix : context.get_in_scope_prefixes()) {
+        if (prefix.is_empty())
```
shannonbooth: I don't think this is needed since it's skipped in context.get_in_scope_prefixes (though, maybe it makes sense to move the special case here?)

Libraries/LibWeb/DOM/Node.cpp
Gingeh: maybe, but all of the other namespace functions operate on strings, so the results will need to be converted anyway in XMLFragmentParser

shannonbooth: Sure, but since it is already FlyString we may as well make use of the O(1) string equality check here before degrading it to a String

Gingeh: Makes sense, done

Libraries/LibWeb/XML/XMLFragmentParser.cpp
Gingeh: Done

PR #7275, opened 2025-12-30
LibURL: Add domain concept to URL::Origin to fix same-origin-domain

shannonbooth:
The same-origin domain check always returned true if only the scheme matched. This was because of missing steps to check for the origin's domain, which didn't exist. Add this concept to URL::Origin, even though we do not use it at this stage in document.domain setter.

Ressurects https://github.com/LadybirdBrowser/ladybird/pull/5488 but with changes of:

- Does not implement the document.domain setter
- Adds tests
- Fixes the type of the domain member in URL::Origin::Tuple

### DIFF
--- a/Libraries/LibURL/Origin.h
+++ b/Libraries/LibURL/Origin.h
@@ -14,6 +14,7 @@
 
 namespace URL {
 
+// https://html.spec.whatwg.org/multipage/browsers.html#concept-origin
 class Origin {
 public:
     using Nonce = Array<u8, 16>;
@@ -25,11 +26,12 @@ public:
 
     static Origin create_opaque();
 
-    Origin(Optional<String> const& scheme, Host const& host, Optional<u16> port)
+    Origin(Optional<String> const& scheme, Host const& host, Optional<u16> port, Optional<String> domain = {})
         : m_state(Tuple {
               .scheme = scheme,
               .host = host,
               .port = move(port),
+              .domain = move(domain),
           })
     {
     }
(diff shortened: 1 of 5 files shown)


Libraries/LibURL/Origin.h
```diff
@@ -72,13 +75,16 @@ class Origin {
         // 2. If A and B are both tuple origins, run these substeps:
         if (!is_opaque() && !other.is_opaque()) {
             // 1. If A and B's schemes are identical, and their domains are identical and non-null, then return true.
-            // FIXME: Check domains once supported.
-            if (scheme() == other.scheme())
+            if (domain().has_value()
+                && other.domain().has_value()
+                && domain().value() == other.domain().value()
```
gmta: These two lines can be reduced to: `domain() == other.domain()`

shannonbooth: I don't think so, that will return true if both domains are null

gmta: But you check for `domain().has_value()` in the line above these two lines?

shannonbooth: Sorry I am a bit confused about the suggested change then, what are the two lines that can be simplified? I guess the .value() could be removed from both, but I suppose that was put there to avoid the == from checking .has_value() again

gmta: The top three lines of the `if` boil down to `domain().has_value() && domain() == other.domain()`. If you know one optional has a value, you can safely compare it to another optional knowing that it will return `false` if the other optional has no value itself.

shannonbooth: Ah right got it, wasn't sure if I would prefer that over the more explicit way, but turns out I do, applied!

gmta (approved):

PR #7279, opened 2025-12-30
LibWeb: Don't crash when clicking on display: contents element

Gingeh:
Fixes https://wpt.live/css/css-display/display-contents-pseudo-click-target.html

### DIFF
--- a/Libraries/LibWeb/Page/EventHandler.cpp
+++ b/Libraries/LibWeb/Page/EventHandler.cpp
@@ -452,7 +452,8 @@ EventResult EventHandler::handle_mousewheel(CSSPixelPoint visual_viewport_positi
                 return EventResult::Dropped;
 
             auto page_offset = compute_mouse_event_page_offset(viewport_position);
-            auto offset = compute_mouse_event_offset(page_offset, *layout_node->first_paintable());
+            auto const& offset_paintable = layout_node->first_paintable() ? layout_node->first_paintable() : paintable.ptr();
+            auto offset = compute_mouse_event_offset(page_offset, *offset_paintable);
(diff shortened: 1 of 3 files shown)


Libraries/LibWeb/Page/EventHandler.cpp
```diff
@@ -443,7 +443,8 @@ EventResult EventHandler::handle_mousewheel(CSSPixelPoint visual_viewport_positi
                 return EventResult::Dropped;
 
             auto page_offset = compute_mouse_event_page_offset(viewport_position);
-            auto offset = compute_mouse_event_offset(page_offset, *layout_node->first_paintable());
+            auto const& offset_paintable = layout_node->first_paintable() ? layout_node->first_paintable() : paintable.ptr();
+            auto offset = compute_mouse_event_offset(page_offset, *offset_paintable);
```
gmta, line 456: Shouldn't such a logic live in `parent_element_for_event_dispatch` instead? I think we should skip over any parent element that has `display: contents`.

Gingeh: The test specifically checks that the event is dispatched to the display:contents element, so that wouldn't be correct

Gingeh: The spec says
> As only the box tree is affected, any semantics based on the document tree, such as selector-matching, event handling, and property [inheritance](https://drafts.csswg.org/css-cascade-5/#css-inheritance), are not affected.

so the event should be dispatched to the display:contents parent, but it doesn't specify how it affects the event offset

gmta: > The test specifically checks that the event is dispatched to the display:contents element

Ah I missed that. Still feels like this could use a more generic solution instead of copy pasting the logic but better to be more correct for now 🤷

PR #7280, opened 2025-12-30
LibCore: Implement watching the time zone on FreeBSD 15

cqundefine:
FreeBSD 15 added support for `inotify` which means we can basically reuse the Linux implementation and it'll just work. I've also sneaked in an SDL update as it contains a FreeBSD fix.

### DIFF
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -262,7 +262,7 @@
     },
     {
       "name": "icu",
-      "version": "78.1#0"
+      "version": "78.1#1"
     },
     {
       "name": "libjpeg-turbo",


Libraries/LibCore/CMakeLists.txt
```diff
@@ -80,28 +80,34 @@ else()
     list(APPEND SOURCES LocalServerWindows.cpp)
 endif()
 
+include(CheckIncludeFile)
+
+check_include_file(sys/inotify.h HAVE_INOTIFY)
+
 # FIXME: Implement these for other systems.
```
gmta: This `FIXME` relates to the `if (LINUX) ...` block below, let's keep it there! Or we could make it more specific (i.e. which systems are missing?)

cqundefine: I moved it down and left it as a note about `ProcessStatistics` only, as `TimezoneWatcher` has an implementation for all the major tested platforms.

cqundefine:
I've added more updates here as those are required to unbreak the build on master.

Issue #7183, opened 2025-12-18
Visiting reddit after fresh compile crashes browser


hexmode:
### Summary

Ladybird comes up. I type reddit.com in the location bar and a crash happens.

### Operating system

Linux

### Steps to reproduce

1.
2.
3.

### Expected behavior

reddit loads

### Actual behavior

tab crashes

### URL for a reduced test case

reddit.com

### HTML/SVG/etc. source for a reduced test case

```html
none
```

### Log output and (if possible) backtrace

```shell
763315.398 WebContent(1016435): FIXME: Handle null navigation_request in navigation response Content Security Policy check.
WARNING: radv is not a conformant Vulkan implementation, testing use only.
763341.390 WebContent(1016670): Potential FIXME: Returning from PerformanceObserver::observe() as we don't support the PerformanceEntry type 'navigation'
763341.390 WebContent(1016670): ResourceLoader: Filtered request to: "https://id.rlcdn.com/472486.gif"
763341.658 WebContent(1016670): ResourceLoader: Filtered request to: "https://googleads.g.doubleclick.net/pagead/viewthroughconversion/788729857/?random=1766077093030&cv=11&fst=1766077093030&bg=ffffff&guid=ON&async=1&en=gtag.config&gtm=45be5ca1v9217137991za200zd9217137991xec&gcd=13l3l3l3l1l1&dma=0&tag_exp=103116026~103200004~104527907~104528501~104684208~104684211~105391252~115583767~115938465~115938469~116184927~116184929~116251938~116251940&u_w=5120&u_h=2880&url=https%3A%2F%2Fwww.reddit.com%2F&frm=0&tiba=Reddit%20-%20The%20heart%20of%20the%20internet&hn=www.googleadservices.com&npa=0&pscdl=noapi&auid=1025407695.1766076888&data=event%3Dgtag.config&rfmt=3&fmt=4"
763341.702 WebContent(1016670): HTMLScriptElement: Refusing to run script because the element's result is null.
763341.810 WebContent(1016670): FIXME: Unimplemented Navigator.maxTouchPoints
763341.878 WebContent(1016670): (js log) "Navigation listeners online"
WARNING: radv is not a conformant Vulkan implementation, testing use only.
[...]

PR #7284, opened 2025-12-30
LibWeb: Handle indefinite width in block formatting context

AlexFalzone:
## Summary
This fixes a crash when visiting Reddit (issue #7183 ) by handling the `indefinite` case for `available_space.width` in `BlockFormattingContext::compute_width()`.

## Changes
   - Modified `BlockFormattingContext.cpp:293` to treat `is_indefinite()` the same as `is_max_content()`, using `calculate_max_content_width(box)`


gmta:
Welcome @AlexFalzone ! Could you start by adding a regression test that reproduces the crash? Since this is layout related, ideally this would be a LibWeb layout test.

PR #7290, opened 2025-12-31
LibWeb: Handle non-absolute lengths when resolving gradient data

matjojo:
This fixes a crash on the major Dutch news site nos.nl. The crash was shrunk to the following case:
```html
<!doctype html>
<div style="position: absolute; background: repeating-linear-gradient(to bottom, red, 10rem, transparent 10rem); width:1rem; top:1rem; height: 20rem"></div>
```
In this case, the problem was caused by the usage of a font-relative size in the gradient section size. (10rem). Where the code expected this to be an absolute length.

The change swaps out calling a function that only handles absolute lengths to one that also handles non-absolute lengths. This function uses the same underlying functions for the already-working code paths, whilst the crashing case now also works.

I have looked through the spec for this length value, and nothing appears to me to imply that the length should be absolute. Chromium and Firefox also do not crash here.
See: https://drafts.csswg.org/css-images/#typedef-color-stop-list

I also added a test that crashes before the change, and does not crash after the change.

### DIFF
--- a/Libraries/LibWeb/Painting/GradientPainting.cpp
+++ b/Libraries/LibWeb/Painting/GradientPainting.cpp
@@ -122,11 +122,13 @@ LinearGradientData resolve_linear_gradient_data(Layout::NodeWithStyle const& nod
 
     CSS::CalculationResolutionContext context {
         .percentage_basis = CSS::Length::make_px(gradient_length_px),
+        .length_resolution_context = CSS::Length::ResolutionContext::for_layout_node(node),
     };
+
     auto resolved_color_stops = resolve_color_stop_positions(
         node, linear_gradient.color_stop_list(), [&](auto const& position) -> float {
             if (position.is_length())
-                return position.as_length().length().absolute_length_to_px_without_rounding() / gradient_length_px;
+                return position.as_length().length().to_px_without_rounding(*context.length_resolution_context) / gradient_length_px;
             if (position.is_percentage())
                 return position.as_percentage().percentage().as_fraction();
             return position.as_calculated().resolve_length(context)->absolute_length_to_px_without_rounding() / gradient_length_px;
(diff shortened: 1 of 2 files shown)


matjojo:
Test failure on CI/Linux: I saw this test fail once locally as well, but was not able to reproduce it reliably.
`Fail: Text/input/WebAnimations/misc/animation-single-iteration-no-repeat.html`

Libraries/LibWeb/Painting/GradientPainting.cpp
```diff
@@ -123,10 +123,11 @@ LinearGradientData resolve_linear_gradient_data(Layout::NodeWithStyle const& nod
...
     };
+    CSS::Length::ResolutionContext lengthResolutionContext = CSS::Length::ResolutionContext::for_layout_node(node);
     auto resolved_color_stops = resolve_color_stop_positions(
         node, linear_gradient.color_stop_list(), [&](auto const& position) -> float {
             if (position.is_length())
-                return position.as_length().length().absolute_length_to_px_without_rounding() / gradient_length_px;
+                return position.as_length().length().to_px_without_rounding(lengthResolutionContext) / gradient_length_px;
             if (position.is_percentage())
                 return position.as_percentage().percentage().as_fraction();
             return position.as_calculated().resolve_length(context)->absolute_length_to_px_without_rounding() / gradient_length_px;
```
konradekk, line 134: (i’m now curious about this line… 😅 [same below])

Libraries/LibWeb/Painting/GradientPainting.cpp
```diff
@@ -123,10 +123,11 @@ LinearGradientData resolve_linear_gradient_data(Layout::NodeWithStyle const& nod
     CSS::CalculationResolutionContext context {
         .percentage_basis = CSS::Length::make_px(gradient_length_px),
     };
+    CSS::Length::ResolutionContext lengthResolutionContext = CSS::Length::ResolutionContext::for_layout_node(node);
```
konradekk: would something like the following work instead?

```suggestion
CSS::CalculationResolutionContext context {
.percentage_basis = CSS::Length::make_px(gradient_length_px),
.length_resolution_context = CSS::Length::ResolutionContext::for_layout_node(node),
};
```

Libraries/LibWeb/Painting/GradientPainting.cpp
matjojo: Ah, I can see why you'd be. `resolve_length` currently returns `Length::make_px(...)` so the call is safe. I could adjust that variant to use the `to_px_without_rounding` helper as well, if you believe it is realistic for `resolve_length` to ever _not_ return a px value.

Libraries/LibWeb/Painting/GradientPainting.cpp
matjojo: Good suggestion, applied it in both places.

Calme1709:
All font-relative lengths should have already been absolutized before now as part of style computation, if not the issue is there and will be observable in other places than just this crash (i.e. `getComputedStyle` will return the relative style when it should return the absolutized one).

It looks like the underlying issue is that we don't try to absolutize `transition_hint` within `ColorStopListElement::absolutized`

matjojo:
> All font-relative lengths should have already been absolutized before now as part of style computation, if not the issue is there and will be observable in other places than just this crash (i.e. `getComputedStyle` will return the relative style when it should return the absolutized one).
> 
> It looks like the underlying issue is that we don't try to absolutize `transition_hint` within `ColorStopListElement::absolutized`

I've made the change you suggest here in another PR. (#7301) It seems to me that that is a better way of solving this problem, so I will close this PR.

Superceded by: #7301

PR #7292, opened 2025-12-31
LibWeb: Selection toString focused text control delegation

### LINKED ISSUE #7267: Text control selections do not interact properly with clipbaord
Resolving properly will be easier once a fix similar to #7233 arrives so one can verify latent selections across form control / page text focus boundaries

Here is a naive (aborted) way to fix clipboard copy on text controls for context: https://github.com/LadybirdBrowser/ladybird/pull/7234

jonbgamble:
Allows selected text in form controls to be copied to clipboard and brings `window.getSelection().toString()` behavior closer to the [w3c selection api working draft](https://www.w3.org/TR/selection-api/#dom-selection-stringifier).

Fixes #7267

### DIFF
--- a/Libraries/LibWeb/HTML/Navigable.cpp
+++ b/Libraries/LibWeb/HTML/Navigable.cpp
@@ -29,6 +29,7 @@
 #include <LibWeb/HTML/BrowsingContextGroup.h>
 #include <LibWeb/HTML/DocumentState.h>
 #include <LibWeb/HTML/HTMLIFrameElement.h>
+#include <LibWeb/HTML/HTMLInputElement.h>
 #include <LibWeb/HTML/HistoryHandlingBehavior.h>
 #include <LibWeb/HTML/Navigable.h>
 #include <LibWeb/HTML/Navigation.h>
@@ -2630,7 +2631,16 @@ String Navigable::selected_text() const
     auto document = active_document();
     if (!document)
         return String {};
+
+    auto const* input_element = as_if<HTML::HTMLInputElement>(document->active_element());
+    if (input_element && input_element->type_state() == HTML::HTMLInputElement::TypeAttributeState::Password) {
+        // Apparently nobody wants bullet characters. We leave the clipboard alone here like other browsers.
+        return String {};
+    }
     auto selection = document->get_selection();
+    if (auto form_text = selection->try_form_control_selected_text_for_stringifier(); form_text.has_value())
+        return form_text->to_utf8();
+
     auto range = selection->range();
     if (!range)
         return String {};
--- a/Libraries/LibWeb/Selection/Selection.cpp
+++ b/Libraries/LibWeb/Selection/Selection.cpp
@@ -14,6 +14,7 @@
 #include <LibWeb/DOM/Range.h>
 #include <LibWeb/DOM/Text.h>
 #include <LibWeb/GraphemeEdgeTracker.h>
+#include <LibWeb/HTML/FormAssociatedElement.h>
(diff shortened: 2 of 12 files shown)


Libraries/LibWeb/Selection/Selection.cpp
```diff
@@ -505,8 +506,31 @@ bool Selection::contains_node(GC::Ref<DOM::Node> node, bool allow_partial_contai
...
+    // still expect window.getSelection() to have the goodies.
+
+    auto const* form_element = as_if<HTML::FormAssociatedTextControlElement>(m_document->active_element());
+    if (!form_element)
+        return {};
+    size_t start = form_element->selection_start();
+    size_t end = form_element->selection_end();
+    if (start >= end)
+        return {};
+    return Utf16String::from_utf16(form_element->value().substring_view(start, end - start));
```
shannonbooth: This seems to be only potentially behaviorally different for `<input>` elements and not `<textarea>`, but using  [relevant value](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#concept-textarea/input-relevant-value) might be more correct here and definitely seems more clear.

I think it would be good to add a test case for `<input>` as well, and best with different types of `<input>`, the case that I am mentioning here seems applicable to certain types of input elements, such as checkbox which has a different implementation of `value()`: https://github.com/LadybirdBrowser/ladybird/blob/953b33d273849403741b7ec5ffd197509b22daf2/Libraries/LibWeb/HTML/HTMLInputElement.cpp#L670

Tests/LibWeb/Text/input/GamepadAPI/gamepad-rumble.html
```diff
@@ -54,7 +54,9 @@
         const effectToPreemptPromise = navGamepad.vibrationActuator.playEffect("dual-rumble", { duration: 1000, strongMagnitude: 0.1, weakMagnitude: 0.1 });
 
         // Make sure the first effect is issued to the gamepad.
-        await requestAnimationFrame(() => { });
+        await new Promise(resolve => {
```
shannonbooth: Seems like a mishap may have happened when rebasing to revert 953b33d273849403741b7ec5ffd197509b22daf2

Libraries/LibWeb/Selection/Selection.cpp
jonbgamble: Thanks for the excellent feedback!

Libraries/LibWeb/HTML/Navigable.cpp
```diff
@@ -2630,7 +2630,11 @@ String Navigable::selected_text() const
     auto document = active_document();
     if (!document)
         return String {};
+
     auto selection = document->get_selection();
+    if (auto form_text = selection->try_selected_text_from_form_control(); form_text.has_value())
+        return form_text->to_well_formed_utf8();
```
shannonbooth: I missed this earlier, but this should probably just be `to_utf8()` as strings in the web generally allow lonely surrogates (WTF-8) including in Ranges

jonbgamble: Oh right, the selection API does not care about clusters.

Libraries/LibWeb/Selection/Selection.cpp
jonbgamble: We still need some special casing here because we don't want to forward type="password" contents to clipboard or window.getSelection().toString(), but from testing the whatwg browsers, they really do seem to blow away the document range whenever any text control gets focus... It looks my latent selection concerns are unfounded which confirms your initial impression in #7267

PR #7296, opened 2026-01-02
LibGfx+LibWeb+LibWebView+ImageDecoder: Add density size correction

Tolasmond:
These changes fix some of the failing tests on WPT density-size-correction tests.
- Added EXIF support to JPEG decoding so we can follow the HTML density-correction steps strictly. We now save APP1 markers, parse the embedded EXIF via TIFFImageDecoderPlugin::read_exif_metadata, cache it, and expose it from metadata() (Libraries/LibGfx/ImageFormats/JPEGLoader.cpp). This lets the ImageDecoder service see X/Y resolution and compute the 72/res scale per spec. With EXIF now flowing through, HTMLImageElement gains the correct density-corrected natural dimensions for JPEGs, so canvas drawImage crop/draw using the spec-defined intrinsic "image pixels" and the WPT density-size-correction tests pass.
- natural_width() and natural_height() were using physical pixels (Libraries/LibWeb/HTML/HTMLImageElement.cpp) returning bitmap->width() and bitmap->height() (physical pixels) instead of density-corrected dimensions. Changed to use intrinsic_width() and intrinsic_height(), which already apply density correction.
- canvas_image_source_dimensions() were using physical pixels (Libraries/LibWeb/HTML/Canvas/CanvasDrawImage.cpp). When drawImage() was called without explicit source coordinates, it defaulted to physical bitmap dimensions instead of intrinsic (density-corrected) dimensions. Changed to use natural_width() and natural_height() which now return the correct density-corrected values. Ensures that: images with EXIF resolution metadata display at their correct CSS pixel size, drawImage() uses the correct default source rectangle, higher DPI images appear smaller (more densely packed pixels = smaller CSS size).


Libraries/LibGfx/ImageFormats/JPEGLoader.cpp
gmta, line 1: Commit message typo: "LibWEbView", "[..] that: Images [..]" -> "that: images" (plus the other sentences that start with a capital letter)

gmta, line 1: Can you either import the WPT tests so you can show that the subtests are now passing, or if that's not possible, create a new regression test?

Libraries/LibGfx/ImageFormats/JPEGLoader.cpp
```diff
@@ -77,10 +81,28 @@ ErrorOr<void> JPEGLoadingContext::decode()
...
         return Error::from_string_literal("Failed to read JPEG header");
 
+    // Extract EXIF metadata (if present) so we can honor resolution and orientation.
+    for (auto* marker = cinfo.marker_list; marker; marker = marker->next) {
+        if (marker->marker != JPEG_APP0 + 1)
+            continue;
+        if (marker->data_length < 6)
+            continue;
+        if (memcmp(marker->data, "Exif\0\0", 6) != 0)
+            continue;
```
gmta, line 96: Maybe the more idiomatic approach would be:

```cpp
        auto exif_data = ReadonlyBytes { marker->data, marker->data_length };
        if (!exif_data.starts_with("Exif\0\0"sv.bytes()))
            continue;
        exif_data = exif_data.slice(6);
```

Libraries/LibGfx/ImageFormats/JPEGLoader.cpp
Tolasmond: Updated commit message.

Libraries/LibGfx/ImageFormats/JPEGLoader.cpp
Tolasmond: made requested change

Libraries/LibGfx/ImageFormats/JPEGLoader.cpp
Tolasmond: Added imported WPT test for density-corrected-size-img.

PR #7294, opened 2026-01-01
LibGfx: A few Vulkan bug fixes found using validation layers

cqundefine:
This is an another split off from #6976. This includes a few fixes that are useful by themselves.

Issue #7207, opened 2025-12-21
WebGL: WebContent crashes on incomplete framebuffer (glCheckFramebufferStatus VERIFY) [ANGLE GL error 0x0507]


gatewaytofredom:
### Summary

WebContent crashes during WebGL framebuffer setup due to a VERIFY that glCheckFramebufferStatus(GL_FRAMEBUFFER) == GL_FRAMEBUFFER_COMPLETE. ANGLE reports GL error 0x0507 during renderbufferStorage and the framebuffer ends up incomplete, aborting the process.

### Operating system

Linux

### Steps to reproduce

1) CCACHE_DISABLE=1 ninja -C Build/release
2) python3 -m http.server --directory Repro 8009
3) ./Build/release/bin/Ladybird "http://127.0.0.1:8009/REDUCTION-laindex-submit-singlefile.html"

### Expected behavior

Page loads without crashing. If WebGL context/framebuffer setup fails (driver/ANGLE error / context lost), it should fail gracefully rather than aborting WebContent.

### Actual behavior

WebContent aborts with:
VERIFICATION FAILED: glCheckFramebufferStatus(GL_FRAMEBUFFER) == GL_FRAMEBUFFER_COMPLETE
at Libraries/LibWeb/WebGL/OpenGLContext.cpp:412

### URL for a reduced test case

https://laindex.net/submit

### HTML/SVG/etc. source for a reduced test case

```html
<!doctype html>
<meta charset="utf-8" />
<title>Ladybird reduction: hCaptcha WebGL crash</title>

<script src="https://js.hcaptcha.com/1/api.js" async defer></script>
<div class="h-captcha" data-sitekey="1c92e496-d2b0-426a-ae96-aeae216ac406"></div>
</html>
```

### Log output and (if possible) backtrace

```shell
ERR: renderergl_utils.cpp:3072 (HandleError): GL call functions->renderbufferStorage(...) generated error 0x00000507 ...
VERIFICATION FAILED: glCheckFramebufferStatus(GL_FRAMEBUFFER) == GL_FRAMEBUFFER_COMPLETE at /home/gateway/Documents/ladybird/Libraries/LibWeb/WebGL/OpenGLContext.cpp:412
Stack trace (most recent call first):
#0  allocate_painting_surface_if_needed ... OpenGLContext.cpp:412
#1  make_current ... OpenGLContext.cpp:427
#2  get_parameter ... WebGLRenderingContextImpl.cpp:795
...
[...]

PR #7302, opened 2026-01-02
LibGfx: Add debug macro to enable Vulkan validation layers

rcorsi:
Vulkan Validation Layers provides diagnostic feedback about Vulkan API usage while live testing Ladybird. It is possible to enable this diagnostic output without changing the code (using the Vulkan SDK), but having it built into the code makes it very easy to enable whenever required.

New debug macro:  VULKAN_VALIDATION_LAYERS_DEBUG

On Ubuntu adding the package `vulkan-validationlayers` is required to use this feature.
Other distros may also have a package, but the name could be different (for example, on Arch, Fedora and NixOS the name is `vulkan-validation-layers`.

Using the validation layers has found a few issues, including one framebuffer issue on the NVIDIA 10 series GPUs in #7207 solved by #7311

Other issues solved in #7294

As discussed with @cqundefine in #7293, will continue this feature in this PR.


rcorsi:
@cqundefine Would you kindly review my latest changes? I've reorganized things and moved most of the code to the bottom of the VulkanContext.cpp file so as not to have it right at the top where it might be a little presumptuous location to have this debug code.

Libraries/LibGfx/VulkanContext.cpp
```diff
@@ -421,4 +436,68 @@ ErrorOr<NonnullRefPtr<VulkanImage>> create_shared_vulkan_image(VulkanContext con
...
+    layers.resize(layer_count);
+    vkEnumerateInstanceLayerProperties(&layer_count, layers.data());
+
+    for (auto const& layer_properties : layers) {
+        if (layer_name == layer_properties.layerName) {
+            return true;
+        }
+    }
+
+    return false;
```
gmta: We have since introduced `Vector::contains(predicate)`, so this could become

```cpp
return layers.contains([&layer_name](auto const& layer_properties) {
    return layer_properties.layerName == layer_name;
});
```

rcorsi: Testing out the change right now.  Will push when ready.  Thanks for the feedback.

gmta (approved):

rcorsi:
Hi @gmta, I noticed you approved the PR, is it on auto merge or does it need to be manually merged?  Anything I need to do, like rebase to latest commit? Thank you!

PR #7305, opened 2026-01-02
LibWeb: More efficiently pack `ComputedValues` struct

Calme1709:
This reduces the size of `ComputedValues` from 3240 bytes to 3064 bytes


AtkinsSJ:
Ah, this conflicts with your other change. But otherwise it's good!

Calme1709:
Rebased to resolve conflicts

PR #7306, opened 2026-01-02
LibJS+LibWeb: Crudely implement growable SharedArrayBuffers incl. support in Wasm memories

CountBleck:
Followup to #5889.

This PR implements the basic functionality of growable SharedArrayBuffers, ignoring the weirdness around concurrency (left as FIXMEs). The integration between this and WebAssembly.Memory is also implemented.

There are still some latent boogs with respect to WebAssembly.Memory, particularly that an growable SharedArrayBuffer should automatically grow on memory growth even after `WebAssembly.Memory#toFixedLengthBuffer()` makes it inactive. I'm not sure if this behavior is specified by the Wasm threads proposal, and I might look into this bug in the future. Regardless, this PR is enough to pass a few more WPT tests.

### DIFF
--- a/Libraries/LibWeb/WebAssembly/Memory.cpp
+++ b/Libraries/LibWeb/WebAssembly/Memory.cpp
@@ -225,9 +225,13 @@ void Memory::refresh_the_memory_buffer(JS::VM& vm, JS::Realm& realm, Wasm::Memor
         // 1. Let block be a Data Block which is identified with the underlying memory of memaddr.
         auto& bytes = cache.abstract_machine().store().get(address)->data();
 
+        // AD-HOC: Neither the main spec nor the threads proposal specify that the Data Block should be Shared for
+        //         shared Wasm memories, but we do in fact want a Shared Data Block in that case.
+        auto is_shared = buffer->is_shared_array_buffer() ? JS::DataBlock::Shared::Yes : JS::DataBlock::Shared::No;
+
         // 2. Set buffer.[[ArrayBufferData]] to block.
         // 3. Set buffer.[[ArrayBufferByteLength]] to the length of block.
-        buffer->set_data_block({ JS::DataBlock::UnownedFixedLengthByteBuffer(&bytes) });
+        buffer->set_data_block({ JS::DataBlock::UnownedFixedLengthByteBuffer(&bytes), is_shared });
     }
 }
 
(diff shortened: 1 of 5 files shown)


Libraries/LibJS/Runtime/ArrayBuffer.cpp
```diff
@@ -315,19 +315,56 @@ ThrowCompletionOr<Optional<size_t>> get_array_buffer_max_byte_length_option(VM&
...
+    if (allocating_growable_buffer) {
+        // a. If byteLength > maxByteLength, throw a RangeError exception.
+        if (byte_length > *max_byte_length)
+            return vm.throw_completion<RangeError>(ErrorType::ByteLengthExceedsMaxByteLength, byte_length, *max_byte_length);
+
+        // b. Append [[ArrayBufferByteLengthData]] and [[ArrayBufferMaxByteLength]] to slots.
+    }
+
+    // 4. Else,
+    //        a. Append [[ArrayBufferByteLength]] to slots.
```
alimpfard, line 335: Are these also fixme's?

Libraries/LibWeb/WebAssembly/Memory.cpp
```diff
@@ -225,9 +225,13 @@ void Memory::refresh_the_memory_buffer(JS::VM& vm, JS::Realm& realm, Wasm::Memor
         // 1. Let block be a Data Block which is identified with the underlying memory of memaddr.
         auto& bytes = cache.abstract_machine().store().get(address)->data();
 
+        // AD-HOC: Neither the main spec nor the threads proposal specify that the Data Block should be Shared for
+        //         shared Wasm memories, but we do in fact want a Shared Data Block in that case.
```
alimpfard, line 229: 🤔 that sounds like a spec issue to me

alimpfard (commented):
looks good, just one comment

Libraries/LibJS/Runtime/ArrayBuffer.cpp
CountBleck: No, LibJS's equivalent of `[[ArrayBufferByteLength]]` is `ArrayBuffer::byte_length`, which just reads the data block's size. That's why I called `create_shared_byte_data_block` below with `byte_length` and not `alloc_length`.

Libraries/LibWeb/WebAssembly/Memory.cpp
CountBleck: If you look at https://webassembly.github.io/threads/js-api/#refresh-the-memory-buffer, the only difference between the threads proposal and the main spec is "ISSUE 3: Is this safe for a SharedArrayBuffer?", so yeah, it is a spec issue...

Libraries/LibJS/Runtime/ArrayBuffer.cpp
CountBleck: *Maybe* `[[ArrayBufferByteLength]]` should be stored separately from the data block, and stored in the ArrayBuffer object instead, but I just assumed that was okay and all.

Should I stick a `// NOTE:` below 4a saying that the byte length is set via the data block?

Libraries/LibWeb/WebAssembly/Memory.cpp
CountBleck: Do I change `AD-HOC` to `SPEC BUG` or smth?

alimpfard: hmm, yeah fair.
let's just leave it as ad-hoc for now.

PR #7309, opened 2026-01-03
LibURL+Elsewhere: Consider file:// origins opaque by default

shannonbooth:


shannonbooth:
The imported WPT is now merged into WPT also

shannonbooth:
oops wrong MR

shannonbooth:
I found one more case where we need to special case opaque file scheme origins since the secure contexts spec assumes that file URLs have a tuple origin

shannonbooth:
Un-stale - this is just a pain to test properly

gmta (approved):

Issue #7264, opened 2025-12-29
macOS: focusing and bluring address bar won't restore current url


warpdesign:
### Summary

When I visit a website, then focuses the address bar, the current url is cleared and I see the placeholder "Enter web address". If I don't type anything and click outside of the address input, I expect the current url to appear in the input.

### Operating system

macOS

### Steps to reproduce

1. visit a website
2. focus the address input
3. type something (optionnal)
3. blur the address input without typing anything by pressing `escape` twice

### Expected behavior

The current website's url should appear in the address input.

### Actual behavior

The address input stays empty with the placeholder "Enter web address".

### URL for a reduced test case

N/A

### HTML/SVG/etc. source for a reduced test case

```html
N/A
```

### Log output and (if possible) backtrace

```shell
N/A
```

### Screenshots or screen recordings

Chrome behavior:

https://github.com/user-attachments/assets/015db74e-8d88-4733-8388-7293094ea5bb

Safari behavior:

https://github.com/user-attachments/assets/e0e8fbef-6e78-4a27-b1b1-dd9a45cf8e08

Ladybird:

https://github.com/user-attachments/assets/7b0b9630-4d42-4415-8c5e-3d4805c2e1fc

### Contribute a patch?

- [ ] I’ll contribute a patch for this myself.

samuelmajtan:
Are you referring to the behavior where the address bar is focused, the URL is manually cleared, and then, after clicking elsewhere on the page, the original page URL does not reappear? Safari behaves this way, but Chrome doesn’t.

warpdesign:
Actually my instructions were wrong, I updated my first post.

In Safari/Chrome/Firefox, focusing the address bar, typing something, and then pressing escape without validating restores the previous (current website) address. In Ladybird it keeps the address bar empty.

I also added videos showing the problem.

PR #7310, opened 2026-01-03
UI/AppKit: Restore address bar URL when canceling edit

sideshowbarker:
When you focus the address bar and then cancel editing (by pressing Escape or clicking somewhere outside the address bar), the original URL is now restored. Previously, the address bar would show whatever text you had typed — or be empty, if you had cleared it.

The fix works by getting the current page’s URL from the web view when:

- Escape is pressed (after the autocomplete dropdown is closed)
- The field loses focus for any reason other than pressing Enter

Fixes https://github.com/LadybirdBrowser/ladybird/issues/7264

### DIFF
--- a/UI/Qt/LocationEdit.cpp
+++ b/UI/Qt/LocationEdit.cpp
@@ -12,6 +12,7 @@
 #include <UI/Qt/StringUtils.h>
 
 #include <QApplication>
+#include <QKeyEvent>
 #include <QPalette>
 #include <QTextLayout>
 #include <QTimer>
@@ -75,6 +76,20 @@ void LocationEdit::focusOutEvent(QFocusEvent* event)
         setCursorPosition(0);
         highlight_location();
     }
+}
+
+void LocationEdit::keyPressEvent(QKeyEvent* event)
+{
+    if (event->key() == Qt::Key_Escape) {
+        if (m_autocomplete->popup()->isVisible()) {
+            QLineEdit::keyPressEvent(event);
+            return;
+        }
+        setText(qstring_from_ak_string(m_url.serialize()));
+        clearFocus();
+        return;
+    }
+    QLineEdit::keyPressEvent(event);
 }
 
 void LocationEdit::search_engine_changed()
--- a/UI/Qt/LocationEdit.h
+++ b/UI/Qt/LocationEdit.h
@@ -32,6 +32,7 @@ public:
 private:
     virtual void focusInEvent(QFocusEvent* event) override;
     virtual void focusOutEvent(QFocusEvent* event) override;
+    virtual void keyPressEvent(QKeyEvent* event) override;
 
     virtual void search_engine_changed() override;
 


konradekk:
why not to use established behaviour, well-known from other browsers? 🤔

sideshowbarker:
> why not to use established behaviour, well-known from other browsers? 🤔

Happy to do that, but I don’t know exactly what behavior you mean… I’m not a “UI guy” and for now in this patch I’m simply just trying to fix the bug reported at https://github.com/LadybirdBrowser/ladybird/issues/7264

warpdesign:
@konradekk  This is exactly what this PR appears to implement: the behavior of other browsers. When the address bar loses focus its previous value is restored.

konradekk:
i might have misunderstood what the author wrote, but whenever i change focus from an address bar to another element, the address in it does not change in my browser.

UI/AppKit/Interface/TabController.mm
```diff
@@ -512,10 +518,16 @@ - (BOOL)control:(NSControl*)control
...
 {
-    auto* location_search_field = (LocationSearchField*)[self.location_toolbar_item view];
+    // Check if editing ended due to pressing Enter (navigation)
+    auto* user_info = [notification userInfo];
+    auto text_movement = [[user_info objectForKey:@"NSTextMovement"] intValue];
 
-    auto url_string = Ladybird::ns_string_to_string([location_search_field stringValue]);
-    [self setLocationFieldText:url_string];
+    // If the user pressed Enter, navigation is handled by doCommandBySelector:
+    // For all other cases (Escape, clicking outside, etc.), restore the URL
```
trflynn89: I think we can leave this function as it was? `setLocationFieldText` just attempts to highlight individual parts of URL. If the user entered a valid URL, we might as well highlight it? Firefox and Chrome seem to behave that way at least.

UI/AppKit/Interface/TabController.mm
```diff
@@ -486,6 +486,12 @@ - (BOOL)control:(NSControl*)control
     if (selector == @selector(cancelOperation:)) {
         if ([self.autocomplete close])
             return YES;
+
+        // Autocomplete was already closed; restore the URL, and blur the field
```
trflynn89: Code comments like this aren't very helpful IMO, when they just say exactly what the next 2 lines literally do.

trflynn89 (commented):
IMO the issue shouldn't be closed until this is implemented for Qt as well.

sideshowbarker:
> IMO the issue shouldn't be closed until this is implemented for Qt as well.

Agreed — and I’m happy to at least try to implement it for Qt too.

Should I do that in this same PR branch? Or should I open a separate PR for it?

trflynn89:
> Should I do that in this same PR branch? Or should I open a separate PR for it?

Up to you, either is totally fine :)

UI/AppKit/Interface/TabController.mm
sideshowbarker: Removed the (redundant) comment.

UI/AppKit/Interface/TabController.mm
sideshowbarker: I’m not sure I understand.

The highlighting still works for valid URLs because _after navigation_, `onURLChange:` is called (via `on_load_finish`) — which calls `setLocationFieldText:` with the new URL.

The change here is specifically for the Escape/click-outside case — we need to restore the original URL rather than keep whatever arbitrary “garbage” the user may have typed — I mean, text that didn’t result in any navigation.

The original code would leave that no-navigation-happened-but-this-got-left-behind “garbage” in the address bar — if the user typed something but then clicked away. Right?

trflynn89: Yes, `controlTextDidEndEditing` is called when the user clicks outside the URL box. But all it does is it grabs the raw text value of the URL box and invokes `setLocationFieldText` with it. The effect is that we're just taking the raw text value of the URL box and highlighting it. So effectively, it's like we're doing:
```c++
auto value = text_field.get_value();
text_field.set_value(value);
```

i.e. the value of the URL box itself does not change. There's just some UI stuff happening to highlight important parts of the URL, if the text value happens to be a URL.

> The change here is specifically for the Escape/click-outside case

To be clear, I would argue we do *not* want the click-outside case to restore the original URL. In all other browsers, if we type something in the URL box and click away, the changed value remains. It's only restored to the original URL when pressing escape. (This is what Konrad was referring to above). Your PR actually behaves this way already (which is correct IMO), just the PR description is off.

sideshowbarker: Fair enough. I have to admit, I hadn’t tested in other browsers. I still like my behavior better, but clearly we don’t want to be “innovating” here — we should just do what other browsers do.

So, I’m made that so.

sideshowbarker:
> > Should I do that in this same PR branch? Or should I open a separate PR for it?
> 
> Up to you, either is totally fine :)

OK, then — I’ll do in in this PR branch. (Though I probably won’t be able to get to it til tomorrow.)

trflynn89 (approved):
Thanks!

PR #7323, opened 2026-01-05
LibWeb: Implement interaction between FontFace and font-face at-rule

tcl3:
This PR implements the required interaction between FontFace objects and font-face at-rules, as specified by: https://drafts.csswg.org/css-font-loading/#font-face-css-connection.

I did notice that other engines don't seem to propagate changes to the connected at-rule when the relevant FontFace properties are changed, but the spec does seem to require that this is a two way interaction.

### DIFF
--- a/Libraries/LibWeb/CSS/FontFaceSet.cpp
+++ b/Libraries/LibWeb/CSS/FontFaceSet.cpp
@@ -138,8 +138,27 @@ bool FontFaceSet::delete_(GC::Root<FontFace> face)
 // https://drafts.csswg.org/css-font-loading/#dom-fontfaceset-clear
 void FontFaceSet::clear()
 {
-    // FIXME: Do the actual spec steps
-    m_set_entries->set_clear();
+    // 1. Remove all non-CSS-connected items from the FontFaceSet's set entries,
+    //    its [[LoadedFonts]] list, and its [[FailedFonts]] list.
+    Vector<JS::Value> to_remove;
+    for (auto font_face_value : *m_set_entries) {
+        auto& font_face = as<FontFace>(font_face_value.key.as_object());
+        if (!font_face.is_css_connected())
+            to_remove.append(font_face_value.key);
+    }
+    for (auto const& value : to_remove)
+        m_set_entries->set_remove(value);
+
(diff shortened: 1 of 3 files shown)


Libraries/LibWeb/CSS/FontFaceSet.cpp
```diff
@@ -138,8 +138,26 @@ bool FontFaceSet::delete_(GC::Root<FontFace> face)
...
+    for (auto font_face_value : *m_set_entries) {
+        auto& font_face = as<FontFace>(font_face_value.key.as_object());
+        if (!font_face.is_css_connected())
+            to_remove.append(font_face_value.key);
+    }
+    for (auto const& value : to_remove)
+        m_set_entries->set_remove(value);
+
+    m_loaded_fonts.clear();
+    m_failed_fonts.clear();
```
AtkinsSJ: To me the spec step means that for all 3 lists, remove non-css-connected items, rather than entirely removing these lists. But it's ambiguous enough that I'm not 100% sure on that.

tcl3: That makes more sense. In my defense, the spec wording could be better here.

Libraries/LibWeb/CSS/FontFaceSet.cpp
```diff
@@ -138,8 +138,27 @@ bool FontFaceSet::delete_(GC::Root<FontFace> face)
...
+            to_remove.append(font_face_value.key);
+    }
+    for (auto const& value : to_remove)
+        m_set_entries->set_remove(value);
+
+    m_loaded_fonts.remove_all_matching([](auto const& font) { return !font->is_css_connected(); });
+    m_failed_fonts.remove_all_matching([](auto const& font) { return !font->is_css_connected(); });
+
+    // 2. If the FontFaceSet's [[LoadingFonts]] list is non-empty, remove all items from it,
+    //    then switch the FontFaceSet to loaded.
```
AtkinsSJ, line 156: This step actually does say to remove everything. :laughing:

tcl3: Oops :sweat_smile:

PR #7325, opened 2026-01-05
LibHTTP: Support the HTTP Vary response header

trflynn89:
We now partition the HTTP disk cache based on the Vary response header. If a cached response contains a Vary header, we look for each of the header names in the outgoing HTTP request. The outgoing request must match every header value in the original request for the cache entry to be used; otherwise, a new request will be issued, and a separate cache entry will be created.

Note that we must now defer creating the disk cache file itself until we have received the response headers. The Vary key is computed from these headers, and affects the partitioned disk cache file name.

There are further optimizations we can make here. If we have a Vary mismatch, we could find the best candidate cached response and issue a conditional HTTP request. The content server may then respond with an HTTP 304 if the mismatch request headers are actually okay. But for now, if we have a Vary mismatch, we issue an unconditional request, as a purely correctness-oriented patch.

### DIFF
--- a/Libraries/LibHTTP/Cache/CacheEntry.cpp
+++ b/Libraries/LibHTTP/Cache/CacheEntry.cpp
@@ -72,10 +72,11 @@ ErrorOr<CacheFooter> CacheFooter::read_from_stream(Stream& stream)
     return footer;
 }
 
-CacheEntry::CacheEntry(DiskCache& disk_cache, CacheIndex& index, u64 cache_key, String url, LexicalPath path, CacheHeader cache_header)
+CacheEntry::CacheEntry(DiskCache& disk_cache, CacheIndex& index, u64 cache_key, u64 vary_key, String url, Optional<LexicalPath> path, CacheHeader cache_header)
     : m_disk_cache(disk_cache)
     , m_index(index)
     , m_cache_key(cache_key)
+    , m_vary_key(vary_key)
     , m_url(move(url))
     , m_path(move(path))
     , m_cache_header(cache_header)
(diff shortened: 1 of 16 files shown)


trflynn89:
Some WPT gains:

`/fetch/http-cache/vary.any.html`
```
▶ Unexpected subtest result in /fetch/http-cache/vary.any.html:
└ PASS [expected FAIL] HTTP cache doesn't use Vary response when request doesn't match

▶ Unexpected subtest result in /fetch/http-cache/vary.any.html:
└ PASS [expected FAIL] HTTP cache doesn't use Vary response when request omits variant header

▶ Unexpected subtest result in /fetch/http-cache/vary.any.html:
└ PASS [expected FAIL] HTTP cache doesn't invalidate existing Vary response

▶ Unexpected subtest result in /fetch/http-cache/vary.any.html:
└ PASS [expected FAIL] HTTP cache doesn't use two-way Vary response when request doesn't match

▶ Unexpected subtest result in /fetch/http-cache/vary.any.html:
└ PASS [expected FAIL] HTTP cache doesn't use two-way Vary response when request omits variant header

▶ Unexpected subtest result in /fetch/http-cache/vary.any.html:
└ PASS [expected FAIL] HTTP cache doesn't use three-way Vary response when request doesn't match

▶ Unexpected subtest result in /fetch/http-cache/vary.any.html:
└ PASS [expected FAIL] HTTP cache doesn't use three-way Vary response when request doesn't match, regardless of header order

▶ Unexpected subtest result in /fetch/http-cache/vary.any.html:
└ PASS [expected FAIL] HTTP cache doesn't use Vary response with a field value of '*'
```

`/fetch/api/request/request-cache-force-cache.any.html`
```
▶ Unexpected subtest result in /fetch/api/request/request-cache-force-cache.any.html:
└ PASS [expected FAIL] RequestCache "force-cache" mode checks the cache for previously cached content and goes to the network if a cached response would vary with Etag and stale response

▶ Unexpected subtest result in /fetch/api/request/request-cache-force-cache.any.html:
└ PASS [expected FAIL] RequestCache "force-cache" mode checks the cache for previously cached content and goes to the network if a cached response would vary with Last-Modified and stale response

▶ Unexpected subtest result in /fetch/api/request/request-cache-force-cache.any.html:
└ PASS [expected FAIL] RequestCache "force-cache" mode checks the cache for previously cached content and goes to the network if a cached response would vary with Etag and fresh response

▶ Unexpected subtest result in /fetch/api/request/request-cache-force-cache.any.html:
└ PASS [expected FAIL] RequestCache "force-cache" mode checks the cache for previously cached content and goes to the network if a cached response would vary with Last-Modified and fresh response
```

(Both of the above now 100% passing. The worker.html and sharedworker.html variants are also passing.)

Issue #6294, opened 2025-09-24
racing (?) module script loads cause failure in LibWeb/HTML/Scripting/Fetching.cpp:fetch_single_module_script()


danielbprice:
### Summary

When I load a file which tries to load a module several times, ladybird will assert.  I found this while trying to reduce a broken website; this issue seems to be unrelated to the broken website.  I reduced the example below as much as I could.

### Operating system

macOS

### Steps to reproduce

1. Load the file below in ladybird
2.
3.

### Expected behavior

File loads and does not crash.  The string `CONTENT` is displayed.

### Actual behavior

Crash.  For me, 100% reproducible, although it looks something like a race condition, I think.

### URL for a reduced test case

N/A

### HTML/SVG/etc. source for a reduced test case

```html
<!doctype html>
<html lang="en">
<head>

<script defer="defer" async type="module" src="file///doesnotexist1.file"></script>
<script defer="defer" async type="module" src="file///doesnotexist1.file"></script>

<script src="file:///somethingelse.bad.com.com"></script>

</head>

<body>

	CONTENT

<script defer="defer" async type="module" src="file///doesnotexist1.file"></script>

</body>
</html>
```

### Log output and (if possible) backtrace

```shell
Ladybird PID file '/Users/dp/Library/Application Support/Ladybird.pid' exists with PID 17382, but process cannot be found
801941.924 WebContent(17496): ResourceLoader: Failed load of: "file:///doesnotexist1.file", Error: No such file or directory (errno=2), Duration: 2ms
801941.924 WebContent(17496): ResourceLoader: Failed load of: "file:///somethingelse.bad.com.com", Error: No such file or directory (errno=2), Duration: 2ms
801941.924 WebContent(17496): HTMLScriptElement: Refusing to run script because the element's result is null.
801941.925 WebContent(17496): HTMLScriptElement: Refusing to run script because the element's result is null.
801941.925 WebContent(17496): fetch file:///doesnotexist1.file single module script waited for fetching to complete: Failed
801941.925 WebContent(17496): HTMLScriptElement: Refusing to run script because the element's result is null.
801941.925 WebContent(17496): fetch file:///doesnotexist1.file single module script waited for fetching to complete: Fetching
[...]

PR #7326, opened 2026-01-05
LibWeb: Address edge case on async module load

crhaglun:
Attempting to address #6294 

This is an edge case when the same module is loaded three times in a document, but all fail.

Failure scenario:
1. Module load 1 set the state to "Fetching"
2. Module load 2 registers a callback to `on_complete` since the current state is "Fetching"
3. Module load 1 finish with a failure, invoking the callback for load 2
4. Module load 3 cause a crash. The state is neither "Fetching" or "ModuleScript", so we'll reset the state to "Fetching". This invokes the callback for module load 2 again, now with an unexpected state which will cause an assert violation.

Proposed fix is to remove the condition that invokes `on_complete` callback immediately for successfully loaded modules only, the callback should be invoked regardless of whether the fetch succeeded or failed. 

This reveals a separate bug in HTMLScriptElement, where `mark_as_ready()` can be invoked before `m_steps_to_run_when_the_result_is_ready` is assigned. 
To mitigate that, add a new method `set_steps_to_run_when_the_result_is_ready()` instead of assigning `m_steps_to_run_when_the_result_is_ready` directly.

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLScriptElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLScriptElement.cpp
@@ -478,6 +478,19 @@ void HTMLScriptElement::prepare_script()
             if (!has_attribute(HTML::AttributeNames::integrity))
                 options.integrity_metadata = resolve_a_module_integrity_metadata(*url, settings_object);
 
+            // AD-HOC: Queue an element task on the networking task source to run the onComplete steps
+            // This resolves an edge case where mark_as_ready() can execute before we set
+            // m_steps_to_run_when_the_result_is_ready
+            // See https://github.com/whatwg/html/issues/12073
+            auto on_complete = create_on_fetch_script_complete(heap(), [this](auto result) {
+                queue_an_element_task(Task::Source::Networking, [this, result = move(result)] {
+                    if (!result)
+                        mark_as_ready(ResultState::Null {});
+                    else
+                        mark_as_ready(Result(*result));
+                });
+            });
+
             // Fetch an external module script graph given url, settings object, options, and onComplete.
             fetch_external_module_script_graph(realm(), *url, settings_object, options, on_complete);
         }
--- a/Libraries/LibWeb/HTML/Scripting/Fetching.cpp
+++ b/Libraries/LibWeb/HTML/Scripting/Fetching.cpp
@@ -654,7 +654,7 @@ void fetch_single_module_script(JS::Realm& realm,
 
     // 6. If moduleMap[(url, moduleType)] exists, run onComplete given moduleMap[(url, moduleType)], and return.
     auto entry = module_map.get(url, module_type.to_byte_string());
-    if (entry.has_value() && entry->type == ModuleMap::EntryType::ModuleScript) {
+    if (entry.has_value()) {
         on_complete->function()(entry->module_script);
         return;
     }
--- a/Tests/LibWeb/Text/expected/HTML/ModuleLoading/multiple-failed-imports.txt
+++ b/Tests/LibWeb/Text/expected/HTML/ModuleLoading/multiple-failed-imports.txt
@@ -1,0 +1,1 @@
+PASS! (Didn't crash)
--- a/Tests/LibWeb/Text/expected/error-stack-must-contain-full-url.txt
+++ b/Tests/LibWeb/Text/expected/error-stack-must-contain-full-url.txt
@@ -6,6 +6,4 @@ module: Error
     at Error
     at blob:file:///[flaky-uuid-replaced]:3:36
     at <unknown>
-    at <unknown>
-    at <unknown>
 
--- a/Tests/LibWeb/Text/input/HTML/ModuleLoading/multiple-failed-imports.html
+++ b/Tests/LibWeb/Text/input/HTML/ModuleLoading/multiple-failed-imports.html
@@ -1,0 +1,17 @@
+<!DOCTYPE html>
+<!--
+    Adapted from https://github.com/LadybirdBrowser/ladybird/issues/6294 
+    Regression test for a crash when target JS modules fail to load
+-->
+<script src="../../include.js"></script>
+
+<script defer="defer" async type="module" src="file///doesnotexist1.file"></script>
+<script defer="defer" async type="module" src="file///doesnotexist1.file"></script>
+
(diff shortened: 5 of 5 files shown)


Libraries/LibWeb/HTML/Scripting/Fetching.cpp
```diff
@@ -637,52 +637,61 @@ void fetch_single_module_script(JS::Realm& realm,
...
-
-                on_complete->function()(entry.module_script);
-            }));
-        });
-
-        return;
-    }
-
-    // 6. If moduleMap[(url, moduleType)] exists, run onComplete given moduleMap[(url, moduleType)], and return.
+    // 5. Check if this entry has already been added
```
gmta: These comments are spec comments and are copied verbatim from the web specs - they're not intended to be modified or renumbered unless the spec itself was changed.

Modifications to the implementation that do not correspond to the spec steps can be annotated using `// AD-HOC: ..` comments.

crhaglun: Oh, I see! Thank you, I'll revert that.

crhaglun: Reverted. Also took a second pass at why the fix only worked when scheduling the callback as a global task, ended up reworking this to a much smaller change in Fetching.cpp + a bug fix in HTMLScriptElement

Libraries/LibWeb/HTML/HTMLScriptElement.cpp
```diff
@@ -541,15 +541,15 @@ void HTMLScriptElement::prepare_script()
             m_preparation_time_document->scripts_to_execute_as_soon_as_possible().append(*this);
 
             // 3. Set el's steps to run when the result is ready to the following:
-            m_steps_to_run_when_the_result_is_ready = [this] {
+            set_steps_to_run_when_the_result_is_ready([this] {
```
shannonbooth: Thanks! The change in fetch also fixes the WPT tests from crashing of:
* https://wpt.live/html/semantics/scripting-1/the-script-element/module/dynamic-import/dynamic-imports-fetch-error.sub.html
* https://wpt.live/html/semantics/scripting-1/the-script-element/json-module/load-error-events.html

However, the change to HTMLScriptElement here seems dubious - I don't quite understand the issue here. I believe we are matching the specification text as written currently. The crash also seems to be resolved without this change also.

What situation does the change to HTMLScriptElement fix?

crhaglun: Without this additional fix, the new test `multiple-failed-imports.html` fails with a timeout. This happens because one of the script elements never reach "executing" stage, i.e. we never call `execute_script()`

If a script has already been fetched, we invoke the `on_complete` callback immediately. If you follow the execution in `HTMLScriptElement::prepare_script()`, there is a path where we end up calling `mark_as_ready` _before_ we set `m_steps_to_run_when_the_result_is_ready`

The specification on https://html.spec.whatwg.org/multipage/scripting.html does not explicitly cover how to handle the case of setting "steps to run when the result is ready" if the result has already been marked "ready" at that point in time. 
I believe the intent is pretty clear, the steps should run "when the result is ready", so if the result is already "ready" they should run immediately.

--- 

An alternative way of resolving this inconsistency would be to queue a task to execute the completion callback in Fetching.cpp instead of calling it directly, something like

````diff
--- a/Libraries/LibWeb/HTML/Scripting/Fetching.cpp
+++ b/Libraries/LibWeb/HTML/Scripting/Fetching.cpp
@@ -655,7 +655,9 @@ void fetch_single_module_script(JS::Realm& realm,
     // 6. If moduleMap[(url, moduleType)] exists, run onComplete given moduleMap[(url, moduleType)], and return.
     auto entry = module_map.get(url, module_type.to_byte_string());
     if (entry.has_value() && entry->type == ModuleMap::EntryType::ModuleScript) {
-        on_complete->function()(entry->module_script);
+        HTML::queue_global_task(HTML::Task::Source::Networking, realm.global_object(), GC::create_function(realm.heap(), [on_complete, entry] {
+            on_complete->function()(entry->module_script);
+        }));
         return;
     }
````

That ensures the `prepare_script()` method executes fully before the callback is invoked, which means we never run `mark_as_ready` before setting `m_steps_to_run_when_the_result_is_ready`.
I feel this would clearly be a deviation from the spec though, since https://html.spec.whatwg.org/multipage/webappapis.html#fetch-a-single-module-script specifies what operations should be queued as tasks and what should be done immediately.

shannonbooth: Okay - I have dug into this a lot more. I think for now we should go with the approach of `HTML::queue_global_task` in the fetch and an AD-HOC comment explaining that this is against spec (as is). With that approach, the assertion on line 535 does not fail, whereas that assertion does fail with the current change.

That assertion was originally commented out by Andreas due to it being bogus, and raised as a specification bug under https://github.com/whatwg/html/issues/8534. However, I believe that issue has since been fixed after Luke raised a different specification issue of https://github.com/whatwg/html/issues/9864 where the specification was updated, which we applied as 24a7eac4ab38 where I think the assertion issue is now fixed by queuing 'mark as ready' underneath a task.

So I think it's best go with the queue of the task for now, and adding back in the assertion.

shannonbooth: Ah cool I see issue now raised as https://github.com/whatwg/html/issues/12073

crhaglun: Updated the PR with an ad-hoc step in HTMLScriptElement that follows my initial suggestion for how to solve https://github.com/whatwg/html/issues/12073

This delays the on_complete callback with a task, but only for the specific path that cause the use-before-set problem exposed by the new test case, rather than doing this for all callbacks in Fetching.cpp

crhaglun: Using an element task for the completion event apparently changed the callstack for Tests/LibWeb/Text/expected/error-stack-must-contain-full-url.txt

Instead of three `<unknown>` frames it's now one.

From what I can understand this should be benign; the script execution changes from a three-layer nested invocation when the callback is invoked via the "fetch" task, to getting directly invoked as a separate element task.

PR #7329, opened 2026-01-05
Tests: Disconnect the virtual gamepad in the navigables gamepad test

Zaggy1024:
Failing to disconnect the gamepad would cause subsequent gamepad tests to fail if they ran in the same WebView.

### DIFF
--- a/Tests/LibWeb/Text/input/GamepadAPI/gamepad-is-available-in-new-navigables.html
+++ b/Tests/LibWeb/Text/input/GamepadAPI/gamepad-is-available-in-new-navigables.html
@@ -34,6 +34,7 @@
 
             println(`Gamepad is now exposed in the new navigable: ${await sendMessageAndWait("getGamepads")}`);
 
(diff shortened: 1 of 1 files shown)


AtkinsSJ:
This works, but it's also kind of papering over the real issue, that test-web should always clear gamepads at the end of each test, and not expose real gamepads so they don't influence the test results.

shannonbooth:
I'm not sure how much this ties in, but for https://github.com/LadybirdBrowser/ladybird/pull/7242#discussion_r2648459654 the only way I can think of fixing that is adding some configuration to TestConfig.ini to run the localStorage tests in parallel since localStorage is inherently a global thing

Zaggy1024:
> This works, but it's also kind of papering over the real issue, that test-web should always clear gamepads at the end of each test, and not expose real gamepads so they don't influence the test results.

That would certainly be nicer, but the goal here is just to change this test to be in line with the other gamepad tests.

PR #7333, opened 2026-01-05
LibJS+LibWeb: Add missing visits for cached cross-origin properties

InvalidUsernameException:
This was causing GC-related crashes on various websites, most prominently on any site that contains embedded YouTube videos. The issue can be reproduced by going to any YouTube video, using the _Share_ button below it and pasting the embed code into an empty HTML file and loading it through localhost.

This is technically a regression from 89dbdd3411a3037ec31dcd96c15a47f062c7f48f (#7275) in that the problem became visible with that commit. However, there is nothing wrong with the commit by itself. It just happens that `Origin::is_same_origin_domain()` prior to that commit was completely bogus and would mistakenly return true in almost all cases, so the cross-origin code paths were not exercised.

I am uncertain how to make a automatic test case for this problem, given the nature of it being GC- and cross-origin-related. So there is no regression test included in this commit.

### DIFF
--- a/Libraries/LibWeb/HTML/Location.cpp
+++ b/Libraries/LibWeb/HTML/Location.cpp
@@ -38,6 +38,8 @@ void Location::visit_edges(Cell::Visitor& visitor)
 {
     Base::visit_edges(visitor);
     visitor.visit(m_default_properties);
+    for (auto& descriptor : m_cross_origin_property_descriptor_map)
+        descriptor.value.visit_edges(visitor);
 }
 
 // https://html.spec.whatwg.org/multipage/nav-history-apis.html#the-location-interface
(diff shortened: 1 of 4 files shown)


Libraries/LibWeb/HTML/Location.cpp
```diff
@@ -38,6 +38,8 @@ void Location::visit_edges(Cell::Visitor& visitor)
 {
     Base::visit_edges(visitor);
     visitor.visit(m_default_properties);
+    for (auto descriptor : m_cross_origin_property_descriptor_map)
```
awesomekling: Let's use `auto& descriptor` here to avoid making copies.

Libraries/LibWeb/HTML/Window.cpp
```diff
@@ -138,6 +138,8 @@ void Window::visit_edges(JS::Cell::Visitor& visitor)
     visitor.visit(m_scrollbars);
     visitor.visit(m_statusbar);
     visitor.visit(m_toolbar);
+    for (auto descriptor : m_cross_origin_property_descriptor_map)
```
awesomekling: `auto&` here as well :)

PR #7337, opened 2026-01-06
LibWeb: Extract `animation-composition` value from keyframe at-rules

tcl3:
Previously, we weren't respecting the value of this property, so the composite operation always defaulted to
`AnimationComposition::Replace`.

Gains us 684 WPT subtests across 41 animation composition tests.


Libraries/LibWeb/Animations/KeyframeEffect.cpp
```diff
@@ -971,4 +971,18 @@ Bindings::CompositeOperation css_animation_composition_to_bindings_composite_ope
...
+{
+    switch (composition) {
+    case CSS::AnimationComposition::Accumulate:
+        return Bindings::CompositeOperationOrAuto::Accumulate;
+    case CSS::AnimationComposition::Add:
+        return Bindings::CompositeOperationOrAuto::Add;
+    case CSS::AnimationComposition::Replace:
+        return Bindings::CompositeOperationOrAuto::Replace;
+    default:
+        VERIFY_NOT_REACHED();
```
gmta: Can we move this `VERIFY_NOT_REACHED()` to outside the `switch` to benefit from compilation errors whenever someone adds an option to `CSS::AnimationComposition`? Same for `css_animation_composition_to_bindings_composite_operation` above.

PR #7340, opened 2026-01-06
LibWeb: Omit `nonzero` fill rule when serializing `polygon()`

Calme1709:
Gains us ~160 WPT passes in `css/css-shapes`


Libraries/LibWeb/CSS/StyleValues/BasicShapeStyleValue.cpp
```diff
@@ -172,19 +172,18 @@ Gfx::Path Polygon::to_path(CSSPixelRect reference_box, Layout::Node const& node)
...
     }
     for (auto const& point : points) {
-        builder.appendff(", {} {}", point.x->to_string(mode), point.y->to_string(mode));
+        if (!builder.is_empty())
+            builder.append(", "sv);
+        builder.appendff("{} {}", point.x->to_string(mode), point.y->to_string(mode));
     }
-    builder.append(')');
-    return MUST(builder.to_string());
+    return MUST(String::formatted("polygon({})", builder.to_string_without_validation()));
```
tcl3: A bit of a nitpick since there are so many string allocations anyway, but could this be done in a way that avoids the extra string allocation introduced by `String::formatted()`? 

Would something like this work?
```diff
diff --git a/Libraries/LibWeb/CSS/StyleValues/BasicShapeStyleValue.cpp b/Libraries/LibWeb/CSS/StyleValues/BasicShapeStyleValue.cpp
index 64e2b6a49a6..290857c94b2 100644
--- a/Libraries/LibWeb/CSS/StyleValues/BasicShapeStyleValue.cpp
+++ b/Libraries/LibWeb/CSS/StyleValues/BasicShapeStyleValue.cpp
@@ -173,15 +173,22 @@ String Polygon::to_string(SerializationMode mode) const
 {
     StringBuilder builder;
     builder.append("polygon("sv);
+
+    bool first = true;
     switch (fill_rule) {
     case Gfx::WindingRule::Nonzero:
-        builder.append("nonzero"sv);
         break;
     case Gfx::WindingRule::EvenOdd:
         builder.append("evenodd"sv);
+        first = false;
+        break;
     }
+
     for (auto const& point : points) {
-        builder.appendff(", {} {}", point.x->to_string(mode), point.y->to_string(mode));
+        if (!first)
+            builder.append(", "sv);
+        first = false;
+        builder.appendff("{} {}", point.x->to_string(mode), point.y->to_string(mode));
     }
     builder.append(')');
     return MUST(builder.to_string());
```

Calme1709: Yep that works, done.

tcl3 (approved):

PR #7342, opened 2026-01-06
LibWeb/HTML: Return Promises from scroll methods, attempt 2

AtkinsSJ:
Last time (#7184) had a couple of issues that are now fixed:
- Accidentally calling a "scroll by this delta" function with a position.
- Missing some temporary execution contexts

To try and make things a bit clearer, the first commit renames and rearranges the scrolling code a bit to align with the spec better. More needs doing there, but this removes a couple of footguns.

Gets us a few WPT passes.

### DIFF
--- a/Libraries/LibWeb/DOM/Element.cpp
+++ b/Libraries/LibWeb/DOM/Element.cpp
@@ -2440,13 +2440,11 @@ static CSSPixelPoint determine_the_scroll_into_view_position(Element& target, Bi
 }
 
 // https://drafts.csswg.org/cssom-view-1/#scroll-a-target-into-view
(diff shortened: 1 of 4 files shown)


Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -2969,7 +2969,7 @@ void Document::scroll_to_the_beginning_of_the_document()
 {
     // FIXME: Actually implement this algorithm
     if (auto navigable = this->navigable())
-        navigable->perform_scroll_of_viewport({ 0, 0 });
+        navigable->perform_scroll_of_viewport_scrolling_box({ 0, 0 });
```
AtkinsSJ, line 2972: This is supposed to scroll the viewport, not its scrolling box. I did try that, and then tried implementing the algorithm properly, but it was causing test-web to crash or hang internally (even when guarding against a null navigable). I did raise an issue on the spec (https://github.com/w3c/csswg-drafts/issues/13301) because the way it calls the algorithm doesn't match up, so I'll probably see if that gets resolved before delving into scroll-land again.

PR #7346, opened 2026-01-06
LibWeb: Isolate gamepad tests from hardware, and each other

AtkinsSJ:
With these changes, tests can't see any real game controllers that are connected to the test machine, and we also forcibly disconnect any virtual gamepads at the end of each test so that forgetting or failing to do so doesn't impact subsequent tests.

**EDIT: Both issues I listed below are now resolved, yay.**

---
There are a couple of parts of this I'm not 100% happy with:
- Detecting if we're running in test mode is done by looking at whether the `--expose-internals-object` flag was set. We don't seem to have a flag just for "we're in test-web" though `--layout-test-mode` looks suspiciously close. The flags could probably use some tidying-up.
- Disconnecting virtual gamepads is done in `begin_navigation()`, so whenever the test page navigates away. It's quite likely there's a better place, but I'm at a loss for where, and it should match the situations where we want to do that. Maybe someone more familiar with the test harness can help me out here. :sweat_smile:

### DIFF
--- a/Libraries/LibWeb/Internals/InternalGamepad.h
+++ b/Libraries/LibWeb/Internals/InternalGamepad.h
@@ -18,8 +18,6 @@ class InternalGamepad : public Bindings::PlatformObject {
 public:
     static constexpr bool OVERRIDES_FINALIZE = true;
 
-    static GC::Ref<InternalGamepad> create(JS::Realm&);
-
     virtual ~InternalGamepad() override;
 
     Array<i32, 15> const& buttons();


Libraries/LibWeb/HTML/Navigable.cpp
```diff
@@ -1623,6 +1624,11 @@ void Navigable::begin_navigation(NavigateParams params)
     auto& active_document = *this->active_document();
     auto& vm = this->vm();
 
+    // AD-HOC: Perform any clean-up in test mode.
+    // FIXME: This should run at the end of each test, and navigating is a heuristic for that.
+    if (Window::in_test_mode())
+        active_window()->internals()->perform_per_test_cleanup();
```
trflynn89: Instead of a navigation heuristic (especially because some tests do more than one nav), why not hook into `Internals::signal_test_is_done`?

gmta:
My 2 cents: the `--layout-test-mode` flag could just as well be named `--test-mode` and cover a range of testing behaviors such as gamepad isolation.

Libraries/LibWeb/HTML/Navigable.cpp
AtkinsSJ: That seems to only be called by the test itself, which means it wouldn't be called if the test times out, I think? (I could be very wrong on that.)

AtkinsSJ:
Changes:
- Rebased to fix a conflict because `InternalGamepad` now has a `OVERRIDES_FINALIZE` member.
- Renamed the `--layout-test-mode` flag to `--test-mode`
- Use that flag directly instead of seeing if the internals object is exposed

AtkinsSJ:
Changes:
- Rebased
- Fixed typo in the first commit
- After discussing with Andreas, moved the clean-up code into `signal_test_is_done()` as Tim suggested. I'm not clear on what happens if the test times out, but that seems like a bug regardless of whether this PR is merged.

AtkinsSJ:
Rebased again, to pick up the Windows CI change.

PR #7347, opened 2026-01-06
LibWeb: Implement composition of filter function lists

tcl3:
This PR implements both addition and accumulation of filter function lists. 

I've also added a `FilterValueListStyleValue::absolutized()` function to ensure the correct computed value is returned before interpolation.

This gains us 467 WPT subtests in the `css/filter-effects` directory.


Libraries/LibWeb/CSS/Interpolation.cpp
```diff
@@ -2134,6 +2134,202 @@ RefPtr<StyleValue const> interpolate_repeatable_list(DOM::Element& element, Calc
...
+        return underlying.visit(
+            [&](FilterOperation::Blur const& underlying_blur) -> Optional<FilterValue> {
+                if (!animated.has<FilterOperation::Blur>())
+                    return {};
+                auto const& animated_blur = animated.get<FilterOperation::Blur>();
+                if (underlying_blur.radius.is_calculated() || animated_blur.radius.is_calculated())
+                    return {};
+                auto underlying_px = underlying_blur.radius.value().raw_value();
+                auto animated_px = animated_blur.radius.value().raw_value();
+                return FilterOperation::Blur { .radius = Length { underlying_px + animated_px, underlying_blur.radius.value().unit() } };
```
AtkinsSJ: For Blur, you assume both units are the same, and name the variables like they're pixels.

Libraries/LibWeb/CSS/Interpolation.cpp
```diff
@@ -2134,6 +2134,202 @@ RefPtr<StyleValue const> interpolate_repeatable_list(DOM::Element& element, Calc
...
+                if (!underlying_deg.has_value() || !animated_deg.has_value())
+                    return {};
+
+                auto get_unit = [](FilterOperation::HueRotate::AngleOrZero const& angle_or_zero) {
+                    return angle_or_zero.visit(
+                        [](AngleOrCalculated const& angle) { return angle.is_calculated() ? AngleUnit::Deg : angle.value().unit(); },
+                        [](FilterOperation::HueRotate::Zero) { return AngleUnit::Deg; });
+                };
+
+                return FilterOperation::HueRotate { .angle = AngleOrCalculated { Angle { *underlying_deg + *animated_deg, get_unit(underlying_rotate.angle) } } };
```
AtkinsSJ: For HueRotate you assume that both angles are the same unit, degrees, but then also query the unit after.

Libraries/LibWeb/CSS/Interpolation.cpp
```diff
@@ -2134,6 +2134,202 @@ RefPtr<StyleValue const> interpolate_repeatable_list(DOM::Element& element, Calc
...
+            [&](FilterOperation::DropShadow const& underlying_shadow) -> Optional<FilterValue> {
+                if (!animated.has<FilterOperation::DropShadow>())
+                    return {};
+                auto const& animated_shadow = animated.get<FilterOperation::DropShadow>();
+
+                auto add_lengths = [](LengthOrCalculated const& a, LengthOrCalculated const& b) -> Optional<LengthOrCalculated> {
+                    if (a.is_calculated() || b.is_calculated())
+                        return {};
+                    return LengthOrCalculated { Length { a.value().raw_value() + b.value().raw_value(), a.value().unit() } };
+                };
```
AtkinsSJ, line 2254: For DropShadow, again assuming the Lengths are both the same unit.

Libraries/LibWeb/CSS/Interpolation.cpp
```diff
@@ -2249,6 +2445,30 @@ RefPtr<StyleValue const> composite_value(StyleValue const& underlying_value, Sty
...
+        // https://drafts.csswg.org/filter-effects/#addition
+        // Given two filter values representing an base value (base filter list) and a value to add (added filter list),
+        // returns the concatenation of the the two lists: ‘base filter list added filter list’.
+        if (composite_operation == Bindings::CompositeOperation::Add) {
+            Vector<FilterValue> result;
+            result.ensure_capacity(underlying_list.size() + animated_list.size());
+            for (auto const& filter : underlying_list.filter_value_list())
+                result.unchecked_append(filter);
+            for (auto const& filter : animated_list.filter_value_list())
+                result.unchecked_append(filter);
```
AtkinsSJ: `extend()` might be better than manually appending each one? Not sure.

AtkinsSJ (changes requested):
The big thing that's bothering me is the vagueness about whether we're dealing with canonical units or not. If all lengths and angles are in their canonical units by the time they get accumulated, then there's no need to ask them what their unit is. But if they're not guaranteed to be in their canonical units, then we need some checks or conversions to make both the same unit.

Libraries/LibWeb/CSS/Interpolation.cpp
tcl3: The units should have been canonicalized by this point, so the unit should be pixels for both. It turns out this canonicalization wasn't happening because there was no  `absolutized()` override to `FilterValueListStyleValue`, so I added one.

Libraries/LibWeb/CSS/Interpolation.cpp
tcl3: I got rid of the unnecessary unit querying from here.

Libraries/LibWeb/CSS/Interpolation.cpp
tcl3: The units should be the same here, as they've been canonicalized.

Libraries/LibWeb/CSS/Interpolation.cpp
tcl3: I've updated this to use `extend()`.

Libraries/LibWeb/CSS/StyleValues/FilterValueListStyleValue.cpp
```diff
@@ -118,4 +119,107 @@ bool FilterValueListStyleValue::contains_url() const
...
+                absolutized_filter_values.append(FilterOperation::Blur {
+                    .radius = absolutize_length(blur.radius),
+                });
+            },
+            [&](FilterOperation::DropShadow const& drop_shadow) {
+                absolutized_filter_values.append(FilterOperation::DropShadow {
+                    .offset_x = absolutize_length(drop_shadow.offset_x),
+                    .offset_y = absolutize_length(drop_shadow.offset_y),
+                    .radius = drop_shadow.radius.map([&](auto const& r) { return absolutize_length(r); }),
+                    .color = drop_shadow.color ? drop_shadow.color->absolutized(computation_context) : nullptr,
```
AtkinsSJ: GCC doesn't like ternaries with nullptr on one side, it thinks the types don't match. :^(

tcl3: Fixed :^)

PR #7348, opened 2026-01-06
LibXML: Replace the existing XML parser with libxml2 parsing

sideshowbarker:
This change replaces our LibXML parser with a new implementation that wraps libxml2's SAX2 API.

The new `Parser` class uses libxml2's SAX2 callbacks to drive the existing `XML::Listener` interface. That preserves backward compatibility with all existing consumers (`XMLDocumentBuilder`, `DOMParser`, etc.).

The implementation is only about 300 lines. And when considered against the roughly 1800 lines of existing code it removes, this lets us enjoy having a net ~1500 less lines of code to maintain.

### DIFF
--- a/Libraries/LibXML/Parser/Parser.cpp
+++ b/Libraries/LibXML/Parser/Parser.cpp
@@ -1,1850 +1,424 @@
 /*
- * Copyright (c) 2022, Ali Mohammad Pur <mpfard@serenityos.org>
+ * Copyright (c) 2026, the Ladybird developers.
  *
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
-#include <AK/StringConversions.h>
-#include <AK/Utf8View.h>
-#include <LibXML/DOM/Document.h>
+#include <AK/StringBuilder.h>
 #include <LibXML/Parser/Parser.h>
 
-struct Range {
-    consteval Range(u32 start, u32 end)
-        : start(start)
-        , end(end)
-    {
-    }
-
-    u32 start;
-    u32 end;
-};
-
-template<auto... ranges>
-struct ranges_for_search {
-    auto contains(u32 value) const
-    {
-        return ((value >= ranges.start && value <= ranges.end) || ...);
-    }
-
-    bool operator()(u32 value) const
-    {
-        return contains(value);
-    }
-
-    template<auto... ranges_to_include>
-    consteval auto with() const
-    {
-        return ranges_for_search<ranges..., ranges_to_include...>();
-    }
-
-    template<auto... ranges_to_include>
-    consteval auto unify(ranges_for_search<ranges_to_include...> const&) const
-    {
-        return ranges_for_search<ranges..., ranges_to_include...>();
-    }
-};
-
-template<size_t Count, typename Element>
-struct StringSet {
-    consteval StringSet(Element const (&entries)[Count])
-    {
-        for (size_t i = 0; i < Count - 1; ++i)
-            elements[i] = entries[i];
-    }
-
-    consteval auto operator[](size_t i) const { return elements[i]; }
-
-    Element elements[Count - 1];
-};
-
-template<StringSet chars>
-consteval static auto set_to_search()
-{
-    return ([&]<auto... Ix>(IndexSequence<Ix...>) {
-        return ranges_for_search<Range(chars[Ix], chars[Ix])...>();
-    }(MakeIndexSequence<array_size(chars.elements)>()));
-}
+#include <libxml/parser.h>
+#include <libxml/parserInternals.h>
+#include <libxml/xmlerror.h>
 
 namespace XML {
 
-size_t Parser::s_debug_indent_level { 0 };
+struct ParserContext {
+    Listener* listener { nullptr };
+    Optional<ParseError> error;
+    bool document_ended { false };
 
-void Parser::append_node(NonnullOwnPtr<Node> node)
+    OwnPtr<Node> root_node;
+    Node* current_node { nullptr };
+    Optional<Doctype> doctype;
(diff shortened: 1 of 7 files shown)


paaspaas00:
Looks like libxml2 is now unmaintained (and was not really properly maintained before), maybe another lib should be used https://www.reddit.com/r/linux/comments/1pi2qcp/libxml2_is_now_officially_unmaintained

sideshowbarker:
> Looks like libxml2 is now unmaintained (and was not really properly maintained before), maybe another lib should be used [reddit.com/r/linux/comments/1pi2qcp/libxml2_is_now_officially_unmaintained](https://www.reddit.com/r/linux/comments/1pi2qcp/libxml2_is_now_officially_unmaintained)

It’s not unmaintained. It actually now has _two_ maintainers — Daniel Garcia Moreno and Iván Chavero — which is double what it had previously.

See https://gitlab.gnome.org/GNOME/libxml2/-/commit/214fb814223614d13d44f1465302c3cbd7d1395a and see https://hackaday.com/2025/12/23/libxml2-narrowly-avoids-becoming-unmaintained/

And another point for the record here: Chrome (and every other Chromium/Blink-based browser) uses libxml2, and the Chrome team and Google have no plans to change that anytime soon.

So, by switching to libxml2, we (to understate things) would not exactly be on our own, and not recklessly taking a giant risk.

And the patch here is written in such a way that it does not tightly bind us to libxml2 anyway. So if it turns out that I’m wrong about everything, and the sky really does end up falling on us, then … we could, at that time, basically just revert this commit, and switch back to using LibXML in the same way we had been.

Libraries/LibXML/Parser/Parser.cpp
```diff
@@ -1,1850 +1,405 @@
...
-    TRY(skip_whitespace());
-    TRY(expect(">"sv));
-
-    rollback.disarm();
-    return PEDeclaration {
-        move(name),
-        move(definition).downcast<ByteString, ExternalID>(),
+    ParseError parse_error {
+        .position = LineTrackingLexer::Position {
+            .offset = 0,
```
tcl3: Rather than populating this field with 0 it seems like we should either get rid of it or find a way to replicate the old behavior. I *think* you could replicate the behavior by doing something like: `static_cast<size_t>(parser_ctx->input->cur - parser_ctx->input->base)`, but I'm not altogether sure how useful this field is anyway.

Libraries/LibXML/Parser/Parser.h
```diff
@@ -1,25 +1,26 @@
...
-#include <AK/SourceLocation.h>
-#include <AK/TemporaryChange.h>
 #include <LibXML/DOM/Document.h>
 #include <LibXML/DOM/DocumentTypeDeclaration.h>
 #include <LibXML/DOM/Node.h>
 #include <LibXML/Export.h>
 #include <LibXML/Forward.h>
 
+#include <libxml/parser.h>
+#include <libxml/parserInternals.h>
```
shannonbooth: This is including libxml.h into the public headers of LibXML, which I think would be good to avoid if possible (through forwarding header or otherwise)

sideshowbarker: > This is including libxml.h into the public headers of LibXML, which I think would be good to avoid if possible (through forwarding header or otherwise)

Yes — thanks. Gone now.

Libraries/LibXML/Parser/Parser.cpp
sideshowbarker: > … I _think_ you could replicate the behavior by doing something like: `static_cast<size_t>(parser_ctx->input->cur - parser_ctx->input->base)`

Went with that.

>  but I'm not altogether sure how useful this field is anyway.

Me neither. But I guess it seems slightly preferable to err on the side of assuming it _might_ be useful to somebody.

Libraries/LibXML/Parser/Parser.cpp
```diff
@@ -1,1850 +1,424 @@
...
-}
-
-void Parser::append_cdata_section(StringView text, LineTrackingLexer::Position position)
-{
-    if (m_listener) {
-        m_listener->cdata_section(text);
-        return;
+    if (parser_ctx->version) {
+        auto version_str = xml_char_to_byte_string(parser_ctx->version);
+        if (version_str == "1.0")
```
alimpfard: `"1.0"sv` pls

UI/Qt/LocationEdit.cpp
alimpfard: This doesn't belong in this commit.

UI/Qt/LocationEdit.h
alimpfard: Also.

UI/Qt/LocationEdit.cpp
sideshowbarker: > This doesn't belong in this commit.

oofs…

Gone now~

Libraries/LibXML/Parser/Parser.cpp
sideshowbarker: > `"1.0"sv` pls

Made it so.

nico:
> And another point for the record here: Chrome (and every other Chromium/Blink-based browser) uses libxml2, and the Chrome team and Google have no plans to change that anytime soon.

🤔

sideshowbarker:
> > And another point for the record here: Chrome (and every other Chromium/Blink-based browser) uses libxml2, and the Chrome team and Google have no plans to change that anytime soon.

well, you would know better than me. obviously I don’t speak for that project. for all I know, y’all right now could have a team of 400 engineers working full-time on writing a new XML parser from scratch…

nico:
That seems about the right amount of people to produce 1800 lines of code 😅

PR #7351, opened 2026-01-06
Teach GC clang plugin to catch more kinds of missing visits, and then fix what it found

awesomekling:
One of the scariest patterns in our codebase has been putting pointers to GC objects in non-GC-allocated objects on the heap, and then praying that somebody visits them during the mark phase.

This PR drastically improves the situation by teaching our Clang plugin to look through non-GC objects that reference GC memory from the malloc heap.

It found a whole host of problems, which we also fix so it can be enforced by CI :)

### DIFF
--- a/Tests/ClangPlugins/LibJSGCTests/substruct_with_visit_edges.cpp
+++ b/Tests/ClangPlugins/LibJSGCTests/substruct_with_visit_edges.cpp
@@ -1,0 +1,33 @@
+/*
+ * Copyright (c) 2026, Andreas Kling <andreas@ladybird.org>
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+// RUN: %clang++ -Xclang -verify %plugin_opts% -c %s -o %t 2>&1
+// expected-no-diagnostics
+
+#include <LibGC/Cell.h>
+#include <LibGC/Ptr.h>
+
+// A substruct that contains GC pointers AND has a visit_edges method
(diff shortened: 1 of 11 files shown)


Tests/ClangPlugins/LibJSGCTests/substruct_with_visit_edges.cpp
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright (c) 2025, Andreas Kling <andreas@ladybird.org>
```
ADKaster: did you start this patch before the new year? :)

Libraries/LibJS/Runtime/Uint8Array.h
```diff
@@ -46,6 +46,12 @@ struct DecodeResult {
     size_t read { 0 };          // [[Read]]
     ByteBuffer bytes;           // [[Bytes]]
     Optional<Completion> error; // [[Error]]
+
+    void visit_edges(Cell::Visitor& visitor)
```
trflynn89: I'm curious about cases like this - `DecodeResult` is only created on the stack. Is the plugin not aware of that? There's no harm in having a `visit_edges` here, but it doesn't seem like anything would ever call it

awesomekling: The plugin is aware! That was left over from an earlier iteration before it was aware :) Will clean up such cases.

PR #7352, opened 2026-01-06
LibRegex: A couple hotfixes :)

alimpfard:


Libraries/LibRegex/RegexOptimizer.cpp
```diff
@@ -238,6 +240,7 @@ struct BytecodeRewriter {
                 : instr.old_ip + instr.size + old_offset;
 
             if (!new_ip_mapping.contains(target_old)) {
+                dbgln("In pattern /{}/", target_pattern);
```
shannonbooth, line 243: deboog

shannonbooth: oh nvm I am stupid

alimpfard: nah I want this to show up in all logs, it'll crash immediately after.

PR #7353, opened 2026-01-06
Everywhere: Update and run .clang-tidy

R-Goc:
This is an attempt at getting clang-tidy back to runing. It may need to be split into multiple PRs as github ui is nearly dying with just AK. If there are style changes here we don't want then I will update the clang-tidy file to suppress the corresponding check and rerun it. This doesn't resolve all warnings. Other ones will need to be resolved manually or suppressed separately, but this is for a future PR.


AK/BinaryHeap.h
```diff
@@ -137,7 +139,7 @@ class BinaryHeap {
 
     void insert(K key, V value)
     {
-        m_heap.insert({ key, value });
+        m_heap.insert({ key, std::move(value) });
```
R-Goc: This can indeed be a move, but the automated fix doesn't fit our coding style. So not sure if to suppress or manually fix it not to include utility and just use move which we have a using statement for.

AK/BitStream.h
```diff
@@ -226,8 +228,7 @@ class LittleEndianInputBitStream : public LittleEndianBitStream {
     {
         // We allow "retrieving" more bits than we can provide, but we need to make sure that we don't underflow the current bit counter.
         // This only affects certain "modes", but all the relevant checks have been handled in the respective `peek_bits` call.
-        if (count > m_bit_count)
-            count = m_bit_count;
+        count = std::min(count, m_bit_count);
```
R-Goc: Same here, not sure if to disable or just fix to use our min.

AK/Assertions.cpp
```diff
@@ -23,26 +23,26 @@
 #    define PRINT_ERROR(s) (void)::fputs((s), stderr)
 #endif
 
-#if defined(AK_HAS_CPPTRACE)
+#ifdef AK_HAS_CPPTRACE
```
trflynn89: Stuff like this is just noise IMO and has no value.

AK/CircularBuffer.cpp
```diff
@@ -154,8 +156,7 @@ size_t CircularBuffer::write(ReadonlyBytes bytes)
         m_used_space += written_bytes;
 
         m_seekback_limit += written_bytes;
-        if (m_seekback_limit > capacity())
-            m_seekback_limit = capacity();
+        m_seekback_limit = std::min(m_seekback_limit, capacity());
```
R-Goc: same as above

AK/BinaryHeap.h
trflynn89: We don't use `std::`

AK/BitStream.h
```diff
@@ -13,6 +13,8 @@
 #include <AK/OwnPtr.h>
 #include <AK/Stream.h>
 
+#include <algorithm>
```
trflynn89: We most certainly don't use `<algorithm>`

AK/CircularBuffer.h
```diff
@@ -26,7 +26,7 @@ class CircularBuffer {
 
     size_t write(ReadonlyBytes bytes);
     Bytes read(Bytes bytes);
-    ErrorOr<void> discard(size_t discarded_bytes);
+    ErrorOr<void> discard(size_t discarding_size);
```
R-Goc, line 29: Not actually sure what it did here, nor which check that is

trflynn89 (commented):
I stopped after a few files. If you have changes that fix real bugs, a better PR would be to just fix those.

EDIT: Especially given that this doesn't even build. Please build before opening PRs.

AK/SourceGenerator.h
```diff
@@ -118,7 +120,7 @@ class SourceGenerator {
     template<size_t N>
     void set(char const (&key)[N], String value)
     {
-        set(StringView { key, N - 1 }, value);
+        set(StringView { key, N - 1 }, std::move(value));
```
R-Goc: same as above

AK/Trie.h
```diff
@@ -28,17 +28,19 @@ struct SubstituteIfVoid<void, Default> {
 
 template<typename DeclaredBaseType, typename DefaultBaseType, typename ValueType, typename MetadataT, typename ValueTraits, template<typename, typename, typename> typename MapType>
 class Trie {
-    using BaseType = typename SubstituteIfVoid<DeclaredBaseType, DefaultBaseType>::Type;
+    using BaseType = SubstituteIfVoid<DeclaredBaseType, DefaultBaseType>::Type;
 
 public:
     using MetadataType = MetadataT;
 
+private:
```
R-Goc, line 36: What is happening here?

R-Goc:
> I stopped after a few files. If you have changes that fix real bugs, a better PR would be to just fix those.
> 
> EDIT: Especially given that this doesn't even build. Please build before opening PRs.

I really should have checked. Looks like the redundant typename check is broken? I am on llvm main, that might be the reason. Can someone with llvm21 try to reproduce? I really didn't expect clang-tidy to automatically break code.

AK/Assertions.cpp
R-Goc: I'll suppress it then.

AK/BinaryHeap.h
R-Goc: As in the comment I made, if it does a change that makes sense(like adding a move or using min/max) but uses std:: should that check just be suppressed or should it be fixed up to use our version/style

R-Goc:
> I stopped after a few files. If you have changes that fix real bugs, a better PR would be to just fix those.

The aim is to get clang-tidy to be useful again. It can catch real bugs and simplify code. Ideally we would get it to a point where we can run it in CI.

AK/BinaryHeap.h
tcl3: FWIW, when I made this change in #6557 using clang-tidy, I fixed all occurrences of `std::move()` and the corresponding `<algorithm>` imports manually.

EDIT: And by manually I mean I used a regex.

R-Goc: Yeah, I think I'll do that. I mostly want to work out the proper ,clang-tidy file for our actual coding style in this PR as it has fallen behind. But I'll fix it up after dealing with the other rules, so as no to lose it.

R-Goc:
Hmm, this breaks code as well:
```
Base64.cpp:64:41: warning: std::move of the variable 'error' of the trivially-copyable type 'Error' has no effect; remove std::move() [performance-move-const-arg]
   64 |         return InvalidBase64 { .error = move(error), .valid_input_bytes = result.count };
      |                                         ^~~~~     ~
```
This has a fixit, which then leads to:
```
E:\lib\ladybird\AK\Base64.cpp(64,41): error: field of type 'Error' has private copy constructor
   64 |         return InvalidBase64 { .error = error, .valid_input_bytes = result.count };
      |                                         ^
E:\lib\ladybird\AK/Error.h(108,5): note: declared private here
  108 |     Error(Error const&) = default;
      |     ^
1 error generated.
```

R-Goc:
Note regarding this:
https://github.com/LadybirdBrowser/ladybird/actions/runs/20758951436/job/59608436013?pr=7353#step:11:434
The check is actually correct and kind of makes sense. But it updates the declaration to match the definition, not the other way around. [readability-inconsistent-declaration-parameter-name] So since it was named abs_path on windows it changed it to that in the header. I'll fix the windows names instead

R-Goc:
Both the invalid clang tidy changes are on track to be fixed for llvm 22.

PR #7355, opened 2026-01-06
LibJS: Follow the spec more closely for tagged template literals

Lubrsi:
This resolves a FIXME in its code generation, particularly for:
- Caching the template object
- Setting the correct property attributes
- Freezing the resulting objects

This allows archive.org to load, which uses the Lit library.

The Lit library caches these template objects to determine if a
template has changed, allowing it to determine to do a full template
rerender or only partially update the rendering. Before, we would
always cause a full rerender on update because we didn't return the
same template object.

This caused issues with archive.org's code, I believe particularly with
its router library, where we would constantly detach and reattach nodes
unexpectedly, ending up with the page content not being attached to the
router's custom element.

### DIFF
--- a/Libraries/LibJS/Bytecode/Executable.h
+++ b/Libraries/LibJS/Bytecode/Executable.h
@@ -70,6 +70,12 @@ struct GlobalVariableCache : public PropertyLookupCache {
     bool in_module_environment { false };
 };
 
+// https://tc39.es/ecma262/#sec-gettemplateobject
+// Template objects are cached at the call site.
+struct TemplateObjectCache {
+    GC::Ptr<Array> cached_template_object;
+};
+
 struct SourceRecord {
     u32 source_start_offset {};
     u32 source_end_offset {};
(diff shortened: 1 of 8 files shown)


nico:
> it's code generation

Libraries/LibJS/Bytecode/Executable.h
```diff
@@ -99,6 +106,7 @@ class JS_API Executable final : public Cell {
     Vector<u8> bytecode;
     Vector<PropertyLookupCache> property_lookup_caches;
     Vector<GlobalVariableCache> global_variable_caches;
+    Vector<TemplateObjectCache> template_object_caches;
```
shannonbooth, line 109: I suppose #7351 might catch it, but this seems like something that should be included in visit_edges()

shannonbooth: Oh dog, I am just blind, it is visited

PR #7359, opened 2026-01-06
LibWeb/SVG: Implement <feDropShadow>

shannonbooth:
I was browsing random sites and noticed on izzygames.com a play button was previously missing a little drop down shadow effect due to missing support for feDropShadow which this implements support for

Before:
image

After:
image

### DIFF
--- a/Libraries/LibWeb/SVG/SVGFilterElement.cpp
+++ b/Libraries/LibWeb/SVG/SVGFilterElement.cpp
@@ -1,6 +1,7 @@
 /*
  * Copyright (c) 2025, Jelle Raaijmakers <jelle@ladybird.org>
  * Copyright (c) 2025, Lucien Fiorini <lucienfiorini@gmail.com>
+ * Copyright (c) 2026, Shannon Booth <shannon@serenityos.org>
  *
  * SPDX-License-Identifier: BSD-2-Clause
  */
@@ -18,6 +19,7 @@
 #include <LibWeb/SVG/SVGFEColorMatrixElement.h>
 #include <LibWeb/SVG/SVGFEComponentTransferElement.h>
 #include <LibWeb/SVG/SVGFECompositeElement.h>
+#include <LibWeb/SVG/SVGFEDropShadowElement.h>
 #include <LibWeb/SVG/SVGFEFloodElement.h>
 #include <LibWeb/SVG/SVGFEFuncAElement.h>
 #include <LibWeb/SVG/SVGFEFuncBElement.h>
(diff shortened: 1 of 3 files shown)


Libraries/LibWeb/DOM/ElementFactory.cpp
```diff
@@ -486,6 +487,8 @@ static GC::Ref<SVG::SVGElement> create_svg_element(JS::Realm& realm, Document& d
         return realm.create<SVG::SVGEllipseElement>(document, move(qualified_name));
     if (local_name == SVG::TagNames::feBlend)
         return realm.create<SVG::SVGFEBlendElement>(document, move(qualified_name));
+    if (local_name == SVG::TagNames::feDropShadow)
+        return realm.create<SVG::SVGFEDropShadowElement>(document, move(qualified_name));
```
gmta: Can you please put this in alphabetical order?

Libraries/LibWeb/SVG/SVGFEDropShadowElement.cpp
```diff
@@ -0,0 +1,121 @@
...
+    visitor.visit(m_dy);
+    visitor.visit(m_std_deviation_x);
+    visitor.visit(m_std_deviation_y);
+}
+
+// https://www.w3.org/TR/filter-effects-1/#FloodColorProperty
+Gfx::Color SVGFEDropShadowElement::flood_color()
+{
+    if (this->computed_properties())
+        return this->computed_properties()->color_or_fallback(CSS::PropertyID::FloodColor, CSS::ColorResolutionContext::for_element({ *this }), CSS::InitialValues::flood_color());
```
gmta: Why is it that you check for `computed_properties()` here but not in `::flood_opacity()`? Also you don't need the `this->` here.

shannonbooth: I had actually copy pasted those two lines from https://github.com/LadybirdBrowser/ladybird/blob/2677338f433122a9117af88f1a9fe790a8be5071/Libraries/LibWeb/SVG/SVGFEFloodElement.cpp#L41 and didnt audit properly 😅 Will fix up both places

I am not actually sure here from the top of my head though, in what case would the computed properties be null? 🤔

gmta: I suppose as long as we've not updated styles elements might go without their computed properties, so it's mostly defensive programming when someone ends up calling `::flood_color()` before that happened. Not sure if that's supposed to happen in practice though 🤔  I much rather leave out the check and let it crash if someone calls these methods without having up to date styles.

shannonbooth:
Looks like I have some real CI failures causing, but apparently only on sanitizer builds... Strange

shannonbooth:
Hmm, I am having trouble reproducing the CI issues locally... Not sure what is going on

shannonbooth:
I found the silliness I was doing causing the issue CI caught:

```patch
 class SVGFEDropShadowElement final
     : public SVGElement
-    , public SVGFilterPrimitiveStandardAttributes<SVGFECompositeElement> {
+    , public SVGFilterPrimitiveStandardAttributes<SVGFEDropShadowElement> {
     WEB_PLATFORM_OBJECT(SVGFEDropShadowElement, SVGElement);
     GC_DECLARE_ALLOCATOR(SVGFEDropShadowElement);
```

PR #7363, opened 2026-01-07
LibJS: Fix memory leak in Shape transition key marking

kalenikaliaksandr:
Previously, `Shape::visit_edges()` unconditionally marked all keys in `forward_transitions` and `delete_transitions`, even when the target Shape had already been garbage collected.

The following script demonstrates the leak:

```js
for (let i = 0; i < 100; i++) {
    let sym = Symbol("unique-" + i);
    let obj = { [sym]: i };
}
```

Before: 116 Symbols alive after GC
After: 16 Symbols alive


awesomekling:
Doesn’t this leave you with stale keys in the map?

PR #7368, opened 2026-01-07
LibWeb: Explicitly accept AVIF and WebP for images

Lubrsi:
Websites depend on there being modern image formats in the Accept header for server side re-encoding software, especially WebP to preserve transparency.

For example, https://tvtropes.org/ for pages about pixel art games where there's typically transparency around the edges of the sprite.

Before:
Screenshot 2026-01-07 at 13 41 22

After:
Screenshot 2026-01-07 at 13 42 41

Spec issue: https://github.com/whatwg/fetch/issues/1905


LucasChollet:
Can we get JPEG XL in there too?

PR #7372, opened 2026-01-07
LibXML: Add externalSubset callback for external DTD entity resolution

tcl3:
When the parser encounters a DOCTYPE with an external ID, the callback fetches the DTD content, parses it with xmlIOParseDTD, and attaches it to the document so entities are available during parsing.

Fixes ~20k failing WPT subtests in `html/the-xhtml-syntax/parsing-xhtml-documents`, which regressed when we switched to `libxml2` for XML parsing.


sideshowbarker (commented):
Non-maintainer LGTM (tested and code-inspected)

Libraries/LibXML/Parser/Parser.cpp
```diff
@@ -267,6 +269,50 @@ static void processing_instruction_handler(void* ctx, xmlChar const* target, xml
...
+    // This callback is called when libxml2 wants to load the external DTD subset
+    auto* parser_ctx = static_cast<xmlParserCtxtPtr>(ctx);
+    auto* context = static_cast<ParserContext*>(parser_ctx->_private);
+    if (!context || !context->options || !context->options->resolve_external_resource)
+        return;
+
+    Optional<PublicID> pub_id;
+    if (external_id)
+        pub_id = PublicID { xml_char_to_byte_string(external_id) };
+
```
nico, line 283: Both Blink and WebKit do significantly less in this callback. Is this really all needed?

sideshowbarker: > Both Blink and WebKit do significantly less in this callback. Is this really all needed?

Hmm, yeah — after looking at the code for both, what they’re doing makes way more sense, and it’s what we should do too. They just do a lookup from a table of the named character references / entities, rather than messing around with the DTD parsing stuff. And the thing is, we (somewhat “of course”) already have the same named-character-reference-lookup table too: For the text/html parser case, because we need to for that. So we can just re-use that.

I’ll write up an alternative patch/PR that does it that way.

sideshowbarker: Raised https://github.com/LadybirdBrowser/ladybird/pull/7375

tcl3:
Closing in favor of https://github.com/LadybirdBrowser/ladybird/pull/7375

Libraries/LibXML/Parser/Parser.cpp
tcl3: Definitely a better approach. Let's go with that instead.

PR #7374, opened 2026-01-08
LibHTTP: A couple of HTTP cache fixes

trflynn89:
* Properly handle network errors from cURL
* Do not cache responses to range requests for now, as we don't handle them at all in our caches

### DIFF
--- a/Libraries/LibHTTP/Cache/Utilities.cpp
+++ b/Libraries/LibHTTP/Cache/Utilities.cpp
@@ -77,12 +77,16 @@ LexicalPath path_for_cache_key(LexicalPath const& cache_directory, u64 cache_key
 }
 
 // https://httpwg.org/specs/rfc9111.html#response.cacheability
-bool is_cacheable(StringView method)
+bool is_cacheable(StringView method, HTTP::HeaderList const& request_headers)
 {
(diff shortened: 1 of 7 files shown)


Libraries/LibHTTP/Cache/Utilities.cpp
```diff
@@ -77,12 +77,16 @@ LexicalPath path_for_cache_key(LexicalPath const& cache_directory, u64 cache_key
...
+bool is_cacheable(StringView method, HTTP::HeaderList const& request_headers)
 {
     // A cache MUST NOT store a response to a request unless:
 
     // * the request method is understood by the cache;
-    return method.is_one_of("GET"sv, "HEAD"sv);
+    if (!method.is_one_of("GET"sv, "HEAD"sv))
+        return false;
+
+    // AD-HOC: Neither the disk cache nor the memory cache handle partial responses yet. So we don't cache them for now.
```
gmta: Imo these should be `// FIXME: ..`s since the intention is to remove this condition once the caches support partial responses.

trflynn89: Sure!

PR #7375, opened 2026-01-08
LibXML+LibWeb: Use existing HTML entities table for XML parsing too

sideshowbarker:
This change makes our now-with-libxml2 XML parsing behavior handle named character references (aka “entities”) as expected — basically by adopting the same approach as WebKit and Blink: use `NamedCharacterReferenceMatcher` to do a lookup in the same giant table of named character references our text/html HTML parser uses.

Otherwise, without this change, when parsing XML-syntax HTML (“XHTML”) docs with libxml2, named character refs don’t work, because XML imagined some thing where parsers network-fetch external DTDs to get the entity definitions, and we (and nobody else either) don’t do anything like that


Libraries/LibXML/Parser/Parser.cpp
```diff
@@ -48,6 +52,80 @@ static StringView xml_char_to_string_view(xmlChar const* str)
...
+        return nullptr;
+
+    auto utf8_bytes = resolved->bytes();
+    if (utf8_bytes.size() >= sizeof(s_xhtml_entity_utf8_buffer))
+        return nullptr;
+
+    memcpy(s_xhtml_entity_utf8_buffer, utf8_bytes.data(), utf8_bytes.size());
+    s_xhtml_entity_utf8_buffer[utf8_bytes.size()] = '\0';
+
+    memset(&s_xhtml_entity_result, 0, sizeof(s_xhtml_entity_result));
```
gmta: Instead of memcpy/memset, can you create a Span<u8> around the buffer as well and use `.copy_to()` / `.fill()`?

sideshowbarker: > Instead of memcpy/memset, can you create a Span around the buffer as well and use `.copy_to()` / `.fill()`?

Yup — made it so

Libraries/LibXML/Parser/Parser.cpp
```diff
@@ -53,6 +52,80 @@ static StringView xml_char_to_string_view(xmlChar const* str)
...
+        return nullptr;
+
+    auto utf8_bytes = resolved->bytes();
+    if (utf8_bytes.size() >= sizeof(s_xhtml_entity_utf8_buffer))
+        return nullptr;
+
+    utf8_bytes.copy_to({ s_xhtml_entity_utf8_buffer, utf8_bytes.size() });
+    s_xhtml_entity_utf8_buffer[utf8_bytes.size()] = '\0';
+
+    Bytes { &s_xhtml_entity_result, sizeof(s_xhtml_entity_result) }.fill(0);
```
gmta: Oh my bad - I thought this was the same buffer. Is `s_xhtml_entity_result` just a struct? Why can't we just default initialize it with `= {}`?

Libraries/LibXML/Parser/Parser.cpp
```diff
@@ -53,6 +52,80 @@ static StringView xml_char_to_string_view(xmlChar const* str)
...
+    auto resolved = context->options->resolve_named_html_entity(entity_name);
+    if (!resolved.has_value())
+        return nullptr;
+
+    auto utf8_bytes = resolved->bytes();
+    if (utf8_bytes.size() >= sizeof(s_xhtml_entity_utf8_buffer))
+        return nullptr;
+
+    utf8_bytes.copy_to({ s_xhtml_entity_utf8_buffer, utf8_bytes.size() });
+    s_xhtml_entity_utf8_buffer[utf8_bytes.size()] = '\0';
```
gmta: We could make this even more idiomatic by using

```cpp
auto utf8_bytes = resolved->bytes_as_string_view();
if (utf8_bytes.length() >= sizeof(s_xhtml_entity_utf8_buffer))
    return nullptr;

utf8_bytes.copy_characters_to_buffer(s_xhtml_entity_utf8_buffer, utf8_bytes.length());
```

alimpfard:
I'm confused, all the definitions of these entities should be in the unified DTD, why do we need whatever this other resolve callback is?

Libraries/LibXML/Parser/Parser.cpp
sideshowbarker: Good point — done, changed to `s_xhtml_entity_result = {}`.

Libraries/LibXML/Parser/Parser.cpp
sideshowbarker: Made it so.

sideshowbarker:
> I'm confused, all the definitions of these entities should be in the unified DTD, why do we need whatever this other resolve callback is?

Added this comment:

```
// For XHTML documents, resolve named character entities (e.g., &nbsp;) using the
// HTML entity table. This avoids parsing a large embedded DTD on every document
// and matches the approach used by Blink and WebKit.
```

PR #7376, opened 2026-01-08
LibXML: Add element-nesting depth limit for XML-parsed documents

sideshowbarker:
This change adds a limit of 5000 on the count for how deeply elements can be nested in documents parsed with our XML parser. Blink and WebKit both have such a limit, and both set it at 5000.

This prevents bad actors from performing attacks by giving us XML docs with pathological levels of nesting, and causing stack exhaustion.

### DIFF
--- a/Libraries/LibXML/Parser/Parser.cpp
+++ b/Libraries/LibXML/Parser/Parser.cpp
@@ -13,6 +13,8 @@
 
 namespace XML {
 
+static constexpr int MAX_XML_TREE_DEPTH = 5000;
+
 struct ParserContext {
     Listener* listener { nullptr };
     Optional<ParseError> error;
@@ -25,6 +27,8 @@ struct ParserContext {
     Version version { Version::Version11 };
 
     Vector<ParseError> parse_errors;
+
+    int depth { 0 };
 };
 
 static ByteString xml_char_to_byte_string(xmlChar const* str)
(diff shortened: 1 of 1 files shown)


Libraries/LibXML/Parser/Parser.cpp
```diff
@@ -90,6 +94,11 @@ static void start_element_ns_handler(void* ctx, xmlChar const* localname, xmlCha
     if (!context)
         return;
 
+    if (++context->depth > MAX_XML_TREE_DEPTH) {
```
gmta, line 97: Should we report something at this point? I.e. `warnln(..)`

sideshowbarker: > Should we report something at this point? I.e. `warnln(..)`

Yup — updated to now report _“Excessive node nesting.”_ (matching what WebKit reports)

Libraries/LibXML/Parser/Parser.cpp
```diff
@@ -91,6 +94,15 @@ static void start_element_ns_handler(void* ctx, xmlChar const* localname, xmlCha
...
+        context->parse_errors.append(ParseError {
+            .position = {},
+            .error = ByteString("Excessive node nesting."sv),
+        });
```
gmta: Looking at `structured_error_handler()`, should we call `context->listener->error(..)` here as well? And are we able to set `.position` here?

sideshowbarker: > Looking at `structured_error_handler()`, should we call `context->listener->error(..)` here as well?

Yes, made it so

> And are we able to set `.position` here?

Yes, and made it so

PR #7387, opened 2026-01-08
LibWeb/HTML: Ensure data: URL workers are same-origin with themselves

shannonbooth:
See: https://github.com/whatwg/html/commit/baff3f5


shannonbooth:
The imported WPT is now merged into WPT also

PR #7391, opened 2026-01-09
LibWeb: Use element's accent-color property for AccentColor keyword

### LINKED ISSUE #7141: CSS `AccentColor` keyword should use the value of `accent-color`

jmhunter83:
## Summary
- When resolving the `AccentColor` and `AccentColorText` CSS keywords, check the element's `accent-color` property first before falling back to the system color
- This aligns with the CSS Color 4 specification which states that `AccentColor` takes its value from `accent-color`
- Added `accent_color` field to `ColorResolutionContext` and populated it from the element's computed values when available

Fixes #7141

## Test plan
- [x] All LibWeb tests pass (5519 pass, 0 fail)
- [x] Existing `color-scheme.html` screenshot test continues to pass
- [x] Manual verification that `AccentColor` respects custom `accent-color` values

### DIFF
--- a/Libraries/LibWeb/CSS/StyleValues/KeywordStyleValue.cpp
+++ b/Libraries/LibWeb/CSS/StyleValues/KeywordStyleValue.cpp
@@ -149,8 +149,10 @@ Optional<Color> KeywordStyleValue::to_color(ColorResolutionContext color_resolut
     // https://www.w3.org/TR/css-color-4/#deprecated-system-colors
     switch (keyword()) {
     case Keyword::Accentcolor:
-        return SystemColor::accent_color(scheme);
+        return color_resolution_context.accent_color.value_or(SystemColor::accent_color(scheme));
     case Keyword::Accentcolortext:
+        if (color_resolution_context.accent_color.has_value())
+            return color_resolution_context.accent_color->suggested_foreground_color();
         return SystemColor::accent_color_text(scheme);
     case Keyword::Buttonborder:
     case Keyword::Activeborder:
--- a/Libraries/LibWeb/CSS/StyleValues/StyleValue.cpp
+++ b/Libraries/LibWeb/CSS/StyleValues/StyleValue.cpp
@@ -91,9 +91,14 @@ ColorResolutionContext ColorResolutionContext::for_element(DOM::AbstractElement 
 
     CalculationResolutionContext calculation_resolution_context { .length_resolution_context = Length::ResolutionContext::for_element(element) };
 
+    Optional<Color> accent_color;
+    if (auto layout_node = element.layout_node())
+        accent_color = layout_node->computed_values().accent_color();
+
     return {
         .color_scheme = color_scheme,
-        .current_color = element.computed_properties()->color_or_fallback(PropertyID::Color, { color_scheme, CSS::InitialValues::color(), element.document(), calculation_resolution_context }, CSS::InitialValues::color()),
+        .current_color = element.computed_properties()->color_or_fallback(PropertyID::Color, { .color_scheme = color_scheme, .current_color = CSS::InitialValues::color(), .accent_color = accent_color, .document = element.document(), .calculation_resolution_context = calculation_resolution_context }, CSS::InitialValues::color()),
+        .accent_color = accent_color,
         .document = element.document(),
         .calculation_resolution_context = calculation_resolution_context
     };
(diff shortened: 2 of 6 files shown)


Libraries/LibWeb/CSS/StyleValues/KeywordStyleValue.cpp
```diff
@@ -149,8 +149,12 @@ Optional<Color> KeywordStyleValue::to_color(ColorResolutionContext color_resolut
     // https://www.w3.org/TR/css-color-4/#deprecated-system-colors
     switch (keyword()) {
     case Keyword::Accentcolor:
+        if (color_resolution_context.accent_color.has_value())
+            return color_resolution_context.accent_color.value();
         return SystemColor::accent_color(scheme);
```
konradekk: (nit: there should be `value_or` available)

Libraries/LibWeb/CSS/StyleValues/StyleValue.cpp
```diff
@@ -92,9 +91,14 @@ ColorResolutionContext ColorResolutionContext::for_element(DOM::AbstractElement
...
     CalculationResolutionContext calculation_resolution_context { .length_resolution_context = Length::ResolutionContext::for_element(element) };
 
+    Optional<Color> accent_color;
+    if (auto layout_node = element.layout_node())
+        accent_color = layout_node->computed_values().accent_color();
+
     return {
         .color_scheme = color_scheme,
-        .current_color = element.computed_properties()->color_or_fallback(PropertyID::Color, { color_scheme, CSS::InitialValues::color(), element.document(), calculation_resolution_context }, CSS::InitialValues::color()),
+        .current_color = element.computed_properties()->color_or_fallback(PropertyID::Color, { .color_scheme = color_scheme, .current_color = CSS::InitialValues::color(), .accent_color = {}, .document = element.document(), .calculation_resolution_context = calculation_resolution_context }, CSS::InitialValues::color()),
```
konradekk: there is an `accent_color` to pass here…?

jmhunter83:
Addressed review feedback:
- Used `value_or()` for the `Accentcolor` case
- Now passing `accent_color` to `color_or_fallback()` instead of `{}`


Thank you for the catch!!! jh

jmhunter83:
The Sanitizer job failure appears to be a pre-existing issue on master (see [run 20883968380](https://github.com/LadybirdBrowser/ladybird/actions/runs/20883968380)) - unrelated to this PR's changes.

konradekk:
is progress visible in some test? (it’d be nice to have it covered, i guess.)

jmhunter83:
Added a test and fixed the bug that the test revealed!

The issue was that `accent_color` was being computed **after** the `ColorResolutionContext` was created in `Layout/Node.cpp`, so when resolving `background-color: AccentColor`, the context had an empty accent_color.

The fix moves accent_color computation earlier, before the context is created.

**Test:** `Tests/LibWeb/Text/input/css/accent-color-keyword.html` - verifies that `AccentColor` respects the element's `accent-color` property.

Is this the kind of test coverage you were looking for?

jmhunter83:
Thanks for pushing on the test coverage - it caught a real bug in my original implementation! Much appreciated.

clever girl

Issue #5333, opened 2025-07-06
Implement OfflineAudioContext::start_rendering()


Prospero23:
[specification](https://webaudio.github.io/web-audio-api/#dom-offlineaudiocontext-startrendering)

Implementing would start to allow webaudio/the-audio-api/ wpts that rely on the audio context running to pass. Since this method is is not implemented yet, many of the tests either timeout or throw a `NotSupportedError`.

Examples of effected tests: 
-  [audionode-connect-order.html](https://wpt.live/webaudio/the-audio-api/the-audionode-interface/audionode-connect-order.html)
- [audionode-disconnect.html](https://wpt.live/webaudio/the-audio-api/the-audionode-interface/audionode-disconnect.html)
- [audionode-disconnect-audioparam.html](https://wpt.live/webaudio/the-audio-api/the-audionode-interface/audionode-disconnect-audioparam.html)
- [osc-basic-waveform.html](https://wpt.live/webaudio/the-audio-api/the-oscillatornode-interface/osc-basic-waveform.html)
- [detune-overflow.html](https://wpt.live/webaudio/the-audio-api/the-oscillatornode-interface/detune-overflow.html)

Many of the other webaudio/the-audio-api/ tests have the same problem.

Opening issue due to recommendation from @gmta on #5144

AtkinsSJ:
Closed in #6547.

PR #4885, opened 2025-05-26
LibWeb: Print FIXME instead of throwing NotSupportedError in WebAudio

kalenikaliaksandr:
This allows us to browse https://pierre.co/ instead of being navigated to error screen.

PR #7395, opened 2026-01-09
Tests: Re-enable some audio tests

AtkinsSJ:
The ones referring to #5333 could have been re-enabled a while back, just nobody noticed. :^)

The periodicWave test is also passing consistently for me, and it wasn't recorded why it was disabled in the first place, so I've enabled that too.


trflynn89:
> The periodicWave test is also passing consistently for me, and it wasn't recorded why it was disabled in the first place, so I've enabled that too.

For reference: https://github.com/LadybirdBrowser/ladybird/pull/4885#pullrequestreview-2875721234

AtkinsSJ:
Aha, thanks! Then it's the same as the others, and should be fine now.

PR #7396, opened 2026-01-09
LibRegex: Reject RSeekTo crossing the current-to-EOL boundary

alimpfard:


alimpfard:
Will be merged via #7740.

PR #7408, opened 2026-01-10
LibWeb: Implement HTMLImageElement x() and y() getters

tcl3:
These attributes get the image's top left border edge  relative to the root element's origin.

These methods ignore any transforms.

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLImageElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLImageElement.cpp
@@ -305,6 +305,44 @@ unsigned HTMLImageElement::natural_height() const
     return 0;
 }
 
+// https://drafts.csswg.org/cssom-view/#dom-htmlimageelement-x
+int HTMLImageElement::x() const
+{
+    // The x attribute, on getting, must return the scaled x-coordinate of the left border edge of the first box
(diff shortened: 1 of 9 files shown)


Libraries/LibWeb/HTML/HTMLImageElement.cpp
```diff
@@ -305,6 +305,44 @@ unsigned HTMLImageElement::natural_height() const
...
 
+// https://drafts.csswg.org/cssom-view/#dom-htmlimageelement-x
+int HTMLImageElement::x() const
+{
+    // The x attribute, on getting, must return the scaled x-coordinate of the left border edge of the first box
+    // associated with the element, relative to the initial containing block origin, ignoring any transforms that apply
+    // to the element and its ancestors, or zero if there is no box.
+    const_cast<DOM::Document&>(document()).update_layout(DOM::UpdateLayoutReason::HTMLImageElementX);
+
+    auto const* paintable = this->paintable_box();
```
gmta: Why the `this->` here?

tcl3: I decided to rename the variable to `paintable_box` and keep the `this->`.

PR #7410, opened 2026-01-10
LibWeb: Only run `handle_src_descriptor_change` once

Calme1709:
Previously we would run it once within `set_src()` and then again within `set_property()` which would lead to us creating two new `FontFace`s instead of just one


Tests/LibWeb/Text/input/css/CSSFontFaceRule-src-change-only-creates-one-new-FontFace.html
```diff
@@ -0,0 +1,19 @@
...
+            src: url("./Foo.woff");
+        }
+    </style>
+    <script src="../include.js"></script>
+    <script>
+        promiseTest(async () => {
+            await document.fonts.ready.then(() => {
+                document.styleSheets[0].cssRules[0].style.src = 'url("./Bar.woff")';
+
+                println(document.fonts.size);
```
tcl3, line 15: This prints 1 in other engines. It looks like we're not removing the old font from document.fonts() when we add the new one in `handle_src_descriptor_change()`.

Calme1709: I have added an additional commit to remove CSS connected font faces from the Document's font source (i.e. `document.fonts()`) when they are disconnected.

PR #7411, opened 2026-01-10
LibWeb: Create containing blocks for non-initial filter values

mikewatt:
This fixes:
http://wpt.live/css/css-will-change/will-change-abspos-cb-002.html 
http://wpt.live/css/css-will-change/will-change-abspos-cb-003.html
http://wpt.live/css/css-will-change/will-change-fixedpos-cb-001.html
http://wpt.live/css/css-will-change/will-change-fixedpos-cb-004.html


AtkinsSJ:
Hi, can you add these two cases to the similar function for fixed position containing blocks too? It's just below. Should be basically the same code.

AtkinsSJ (approved):
Thanks!

PR #7412, opened 2026-01-10
LibWeb: Parse Referrer-Policy header when creating policy container

tcl3:
Previously, when creating a policy container from a fetch response, the Referrer-Policy HTTP header was not being parsed. This meant documents loaded with a Referrer-Policy header would ignore the policy and use the default.

This gains us ~400 WPT subtests in the `referrer-policy/` directory.

### DIFF
--- a/Libraries/LibWeb/HTML/PolicyContainers.cpp
+++ b/Libraries/LibWeb/HTML/PolicyContainers.cpp
@@ -13,6 +13,7 @@
 #include <LibWeb/Fetch/Infrastructure/URL.h>
 #include <LibWeb/HTML/PolicyContainers.h>
 #include <LibWeb/HTML/SerializedPolicyContainer.h>
+#include <LibWeb/ReferrerPolicy/AbstractOperations.h>
 
 namespace Web::HTML {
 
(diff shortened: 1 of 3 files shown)


Libraries/LibWeb/HTML/PolicyContainers.cpp
```diff
@@ -50,9 +51,11 @@ GC::Ref<PolicyContainer> create_a_policy_container_from_a_fetch_response(GC::Hea
     // FIXME: 4. If environment is non-null, then set result's embedder policy to the result of obtaining an embedder
     //           policy given response and environment. Otherwise, set it to "unsafe-none".
 
-    // FIXME: 5. Set result's referrer policy to the result of parsing the `Referrer-Policy` header given response.
-    //           [REFERRERPOLICY]
-    //        Doing this currently makes Fetch fail the policy != ReferrerPolicy::EmptyString verification.
```
tcl3, line 55: I wasn't able to find any referrer-policy or CSP tests that cause this VERIFY to be hit, so I'm assuming this is no longer an issue?

gmta (approved):

PR #7413, opened 2026-01-10
Everywhere: Fix some test related FreeBSD compilation errors

cqundefine:
Apparently I never tried building tests on FreeBSD, oops. Here are a few small changes to fix some compilation errors I found just now.

### DIFF
--- a/Utilities/test262-runner.cpp
+++ b/Utilities/test262-runner.cpp
@@ -546,6 +546,8 @@ static bool g_in_assert = false;
 
 #ifdef AK_OS_MACOS
 extern "C" __attribute__((__noreturn__)) void __assert_rtn(char const* function, char const* file, int line, char const* assertion)
+#elifdef AK_OS_FREEBSD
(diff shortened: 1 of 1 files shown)


Libraries/LibMedia/CMakeLists.txt
```diff
@@ -45,7 +45,7 @@ if (LADYBIRD_AUDIO_BACKEND STREQUAL "PULSE")
         Audio/PlaybackStreamPulseAudio.cpp
         Audio/PulseAudioWrappers.cpp
     )
-    target_link_libraries(LibMedia PRIVATE PkgConfig::PULSEAUDIO)
+    target_link_libraries(LibMedia PUBLIC PkgConfig::PULSEAUDIO)
```
shannonbooth: I believe this is intentionally private because we do not want other Libraries to be directly using anything relating to pulse audio, and should keep this as an implementation detail of LibMedia. I think it would be better if we explicitly link the test which is including: PulseAudioWrappers.h (or even if that test somehow did not directly rely on pulse)

PR #7414, opened 2026-01-10
CI: Make Flatpak build part of regular CI builds

gmta:
This means the Flatpak build will now execute as part of PRs and pushes to our `master` branch.

CC @trflynn89


gmta:
> Looking at the added build times, 46 minutes per PR now is a bit sad, but it is what it is.

Repeated runs finish in 12-13 minutes, so the flatpak cache seems to be doing its job.

PR #7420, opened 2026-01-10
LibWeb: Some FontFace related fixes

Calme1709:
This PR includes a number of (somewhat) disjointed fixes to `FontFace`/`FontFaceSet`/`CSSFontFaceRule` - see individual commits for details.

Gains us ~100 WPT passes.


AtkinsSJ:
A couple of these updated tests are flaky, I guess because we're not always loading the font in time? (We still don't actually wait for style resources to be loaded before the onload event. :sweat_smile: )

Example: https://github.com/LadybirdBrowser/ladybird/actions/runs/20955561213/job/60219195932#step:15:7206

PR #7424, opened 2026-01-10
LibWeb: Implement the very basics of XRSession

Psychpsyo:
Continuation of #6816.

Not gonna just drop 1500LOCs all at once like I did with view transitions.
WPT Passes: None, cause that'll need additions to our testing harness. So for now, I've written a simplified test that we can pass without those.


Libraries/LibWeb/WebXR/XRSession.cpp
```diff
@@ -0,0 +1,140 @@
...
+    WEB_SET_PROTOTYPE_FOR_INTERFACE(XRSession);
+    Base::initialize(realm);
+}
+
+void XRSession::visit_edges(Cell::Visitor& visitor)
+{
+    Base::visit_edges(visitor);
+
+    visitor.visit(m_xr_system);
+    for (auto promise : m_outstanding_promises) {
```
shannonbooth:
```suggestion
    visitor.visit(m_outstanding_promises);
```

Libraries/LibWeb/WebXR/XRSession.cpp
```diff
@@ -0,0 +1,140 @@
...
+        visitor.visit(promise);
+    }
+}
+
+GC::Ref<WebIDL::Promise> XRSession::create_promise(JS::Realm& realm)
+{
+    auto promise = WebIDL::create_promise(realm);
+    m_outstanding_promises.append(promise);
+    return promise;
+}
```
shannonbooth, line 55: Some newlines here please!

Libraries/LibWeb/WebXR/XRSystem.cpp
```diff
@@ -34,6 +37,11 @@ void XRSystem::initialize(JS::Realm& realm)
 void XRSystem::visit_edges(Cell::Visitor& visitor)
 {
     Base::visit_edges(visitor);
+
+    visitor.visit(m_active_immersive_session);
+    for (auto session : m_list_of_inline_sessions) {
```
shannonbooth:
```suggestion
    visitor.visit(m_list_of_inline_sessions);
```

Libraries/LibWeb/WebXR/XRSession.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+GC::Ref<WebIDL::Promise> XRSession::end()
+{
+    // 1. Let promise be a new Promise in the relevant realm of this XRSession.
+    auto& realm = HTML::relevant_realm(*this);
+    auto promise = WebIDL::create_promise(realm);
+
+    // 2. If the ended value of this is true, reject promise with a "InvalidStateError" DOMException and return promise.
+    if (this->m_ended)
```
shannonbooth: I don't think we typically use this explicitly in most cases (few other instances of this around)

Libraries/LibWeb/WebXR/XRSystem.cpp
```diff
@@ -72,4 +80,139 @@ GC::Ref<WebIDL::Promise> XRSystem::is_session_supported(Bindings::XRSessionMode
...
+            }
+
+            // FIXME:
+            // 2. Let descriptor be an XRPermissionDescriptor initialized with mode, requiredFeatures, and optionalFeatures
+            // 3. Let status be an XRPermissionStatus, initially null
+            // 4. Request the xr permission with descriptor and status.
+            // 5. If status’ state is "denied" run the following steps:
+            if (false) {
+                // 1. Reject promise with a "NotSupportedError" DOMException.
+                WebIDL::reject_promise(realm, promise, WebIDL::InvalidStateError::create(realm, "The XR Permissions are denied."_utf16));
```
shannonbooth: Not a NotSupported error

Libraries/LibWeb/WebXR/XRSystem.cpp
```diff
@@ -72,4 +80,139 @@ GC::Ref<WebIDL::Promise> XRSystem::is_session_supported(Bindings::XRSessionMode
...
+        (void)optional_features;
+
+        // 4. Queue a task to perform the following steps:
+        HTML::queue_a_task(HTML::Task::Source::Unspecified, nullptr, nullptr, GC::create_function(realm.heap(), [this, &realm, promise, immersive]() {
+            HTML::TemporaryExecutionContext context { realm, HTML::TemporaryExecutionContext::CallbacksEnabled::Yes };
+            // 1. If device is null or device’s list of supported modes does not contain mode, run the following steps:
+            // AD-HOC: Just reject immersive sessions here until we have devices.
+            if (immersive) {
+                // 1. Reject promise with a "NotSupportedError" DOMException.
+                WebIDL::reject_promise(realm, promise, WebIDL::InvalidStateError::create(realm, "Sessions of this mode are not supported."_utf16));
```
shannonbooth: Also not a NotSupportedError

Libraries/LibWeb/WebXR/XRSystem.cpp
```diff
@@ -72,4 +80,139 @@ GC::Ref<WebIDL::Promise> XRSystem::is_session_supported(Bindings::XRSessionMode
...
+                m_pending_immersive_session = false;
+            }
+            // -> Otherwise:
+            else {
+                // Append session to the list of inline sessions.
+                m_list_of_inline_sessions.append(session);
+            }
+
+            // 10. Resolve promise with session.
+            WebIDL::resolve_promise(realm, promise, session);
```
shannonbooth, line 188: Hmm, would this promise need to be associated with the session for the custom resolve?

Libraries/LibWeb/WebXR/XRSession.cpp
```diff
@@ -0,0 +1,140 @@
...
+#include <LibWeb/HTML/Scripting/TemporaryExecutionContext.h>
+#include <LibWeb/WebIDL/Promise.h>
+#include <LibWeb/WebXR/XRSession.h>
+#include <LibWeb/WebXR/XRSessionEvent.h>
+
+namespace Web::WebXR {
+
+GC_DEFINE_ALLOCATOR(XRSession);
+
+GC::Ref<XRSession> XRSession::create(JS::Realm& realm, XRSystem& xr_system)
```
shannonbooth: Typically I think this would be a GC::Ref, not a bare &

Libraries/LibWeb/WebXR/XRSystem.h
```diff
@@ -24,11 +31,30 @@ class XRSystem final : public DOM::EventTarget {
...
     virtual void visit_edges(JS::Cell::Visitor&) override;
+
+    // https://immersive-web.github.io/webxr/#pending-immersive-session
+    bool m_pending_immersive_session { false };
+
+    // https://immersive-web.github.io/webxr/#active-immersive-session
+    GC::Ptr<XRSession> m_active_immersive_session {};
+
+    // https://immersive-web.github.io/webxr/#list-of-inline-sessions
+    AK::Vector<GC::Ptr<XRSession>> m_list_of_inline_sessions {};
```
shannonbooth:
```suggestion
    Vector<GC::Ref<XRSession>> m_list_of_inline_sessions;
```

Libraries/LibWeb/WebXR/XRSystem.h
```diff
@@ -24,11 +31,30 @@ class XRSystem final : public DOM::EventTarget {
...
 
+    // https://immersive-web.github.io/webxr/#dom-xrsystem-requestsession
+    GC::Ref<WebIDL::Promise> request_session(Bindings::XRSessionMode, XRSessionInit);
+
+    void set_pending_immersive_session(bool pending_immersive_session) { m_pending_immersive_session = pending_immersive_session; }
+
+    GC::Ptr<XRSession> active_immersive_session() const { return m_active_immersive_session; }
+    void set_active_immersive_session(GC::Ptr<XRSession> active_immersive_session) { m_active_immersive_session = active_immersive_session; }
+
+    AK::Vector<GC::Ptr<XRSession>> list_of_inline_sessions() const { return m_list_of_inline_sessions; }
```
shannonbooth: It's a bit dangerous to expose an API that returns a vector of GC pointers like this since they then are no longer tracked by the GC. However - it seems like this is used in a buggy way by returning a copy like this anyhow since the caller is operating on the copy:

```
    // 3. Remove session from the list of inline sessions.
    m_xr_system->list_of_inline_sessions().remove_first_matching([&](auto& entry) { return entry == this; }); 
```

Maybe we should instead have an API which allows removal to avoid seeing the list directly?

Libraries/LibWeb/WebXR/XRSessionEvent.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+GC_DEFINE_ALLOCATOR(XRSessionEvent);
+
+GC::Ref<XRSessionEvent> XRSessionEvent::create(JS::Realm& realm, FlyString const& type, XRSessionEventInit const& event_init)
+{
+    return realm.create<XRSessionEvent>(realm, type, event_init);
+}
+
+// https://immersive-web.github.io/webxr/#dom-xrsessionevent-xrsessionevent
+WebIDL::ExceptionOr<GC::Ref<XRSessionEvent>> XRSessionEvent::construct_impl(JS::Realm& realm, FlyString const& type, XRSessionEventInit const& event_init)
```
shannonbooth: No need for ExceptionOr as this does not throw any exception

Libraries/LibWeb/WebXR/XRSession.idl
```diff
@@ -0,0 +1,41 @@
...
+    [FIXME] readonly attribute boolean isSystemKeyboardSupported;
+    
+    // Methods
+    [FIXME] undefined updateRenderState(optional XRRenderStateInit state = {});
+    [FIXME] Promise<undefined> updateTargetFrameRate(float rate);
+    [FIXME, NewObject] Promise<XRReferenceSpace> requestReferenceSpace(XRReferenceSpaceType type);
+    
+    [FIXME] unsigned long requestAnimationFrame(XRFrameRequestCallback callback);
+    [FIXME] undefined cancelAnimationFrame(unsigned long handle);
+
```
shannonbooth: There appears to be some trailing whitespace in this IDL file

Tests/LibWeb/Text/input/WebXR/sessionEndEvent.html
```diff
@@ -0,0 +1,11 @@
+<!DOCTYPE html>
+<script src="../include.js"></script>
+<script>
+    asyncTest(async (done) => {
+        const session = await navigator.xr.requestSession("inline");
+        session.addEventListener("end", () => println("Session ended as requested."));
+        session.end();
+        // Since end() might need to shut down XR devices and stuff, let's give it 10 seconds to send the event.
```
shannonbooth, line 11: I would really rather not have a test in tree that takes 10 seconds to run every time

Libraries/LibWeb/WebXR/XRSessionEvent.h
```diff
@@ -0,0 +1,43 @@
...
+ * Copyright (c) 2025, Psychpsyo <psychpsyo@gmail.com>
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#pragma once
+
+#include <LibWeb/Bindings/PlatformObject.h>
+#include <LibWeb/DOM/Event.h>
+#include <LibWeb/WebXR/XRSession.h>
```
shannonbooth: Not that it makes a big difference, but WebXR/XRSession.h is not needed because of the forwarding header and PlatformObject.h is brought in by Event.h as a subclass.

Libraries/LibWeb/WebXR/XRSession.h
```diff
@@ -0,0 +1,58 @@
...
+    // https://immersive-web.github.io/webxr/#dom-xrsession-end
+    GC::Ref<WebIDL::Promise> end();
+
+    // https://immersive-web.github.io/webxr/#shut-down-the-session
+    void shut_down();
+
+    GC::Ptr<WebIDL::CallbackType> onend();
+    void set_onend(GC::Ptr<WebIDL::CallbackType>);
+
+    bool promise_resolved() { return m_promise_resolved; }
```
shannonbooth:
```suggestion
    bool promise_resolved() const { return m_promise_resolved; }
```

Libraries/LibWeb/WebXR/XRSession.h
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright (c) 2025, Psychpsyo <psychpsyo@gmail.com>
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#pragma once
+
+#include <LibWeb/Bindings/PlatformObject.h>
+#include <LibWeb/Bindings/XRSessionPrototype.h>
```
shannonbooth: And similarly here of these includes only EventTarget.h is needed in the header (though things like Vector could be included)

Libraries/LibWeb/WebXR/XRSession.cpp
Psychpsyo: `WebIDL::create_promise()`? What would be the alternative there?
Any place in the codebase that I can find where the spec says to "Let promise be a new promise" uses it.

Libraries/LibWeb/WebXR/XRSystem.cpp
Psychpsyo: No, because it is not a promise created by the XRSession, it is a promise created by the XRSystem that will tell us if creating the XRSession was successful or not.

Tests/LibWeb/Text/input/WebXR/sessionEndEvent.html
Psychpsyo: I'm also making it call done when the session ends, so that way it only ever takes 10 seconds if that does not happen. (i.e. if the test actually fails)

I hope that's fine.

shannonbooth: Huh, when I ran locally it seemed like it ran for 10 seconds, will pull and try again, must have done something wrong

Libraries/LibWeb/WebXR/XRSession.cpp
shannonbooth: Sorry, to clarify I mean just the style of:
```suggestion
    if (m_ended)
```

Libraries/LibWeb/WebXR/XRSystem.h
Psychpsyo: Done.

Tests/LibWeb/Text/input/WebXR/sessionEndEvent.html
Psychpsyo: Yes, it used to do that. But I've now changed it to this better option. (I think I forgot to put "now" in my earlier message somewhere...)

Libraries/LibWeb/WebXR/XRSession.cpp
Psychpsyo: Ahhh, ok. Done now.

PR #7426, opened 2026-01-11
LibWeb: A couple of style computation optimizations

Calme1709:
A couple of optimizations within `StyleComputer::compute_properties()` which have the following effects on runtime for the (somewhat contrived) benchmark below:
  - `CascadedProperties::is_property_important()`: 2.34% -> 0.03%
  - `FontComputer::compute_font_for_style_values()`: 2.23% -> 0.20%

```html
<!DOCTYPE html>
<html>
    <div id="foo"></div>
    <pre id="out"></pre>
    <script>
        const start = performance.now();
        for (let i = 0; i < 10_000; i++) {
            // Run pending style computation
            getComputedStyle(foo).width;

            // Trigger a style change
            foo.style.left = `${i}px`;
        }

        out.innerText = `Runtime: ${(performance.now() - start).toFixed(3)}ms`;
    </script>
</html>
```

### DIFF
--- a/Libraries/LibWeb/CSS/FontComputer.cpp
+++ b/Libraries/LibWeb/CSS/FontComputer.cpp
@@ -63,13 +63,28 @@ struct Traits<Web::CSS::OwnFontFaceKey> : public DefaultTraits<Web::CSS::OwnFont
 };
 
 template<>
-struct Traits<Web::CSS::FontMatchingAlgorithmCacheKey> : public DefaultTraits<Web::CSS::FontMatchingAlgorithmCacheKey> {
-    static unsigned hash(Web::CSS::FontMatchingAlgorithmCacheKey const& key)
+struct Traits<Web::CSS::ComputedFontCacheKey> : public DefaultTraits<Web::CSS::ComputedFontCacheKey> {
+    static unsigned hash(Web::CSS::ComputedFontCacheKey const& key)
     {
-        auto hash = key.family_name.hash();
-        hash = pair_int_hash(hash, key.weight);
-        hash = pair_int_hash(hash, key.slope);
-        hash = pair_int_hash(hash, Traits<float>::hash(key.font_size_in_pt));
+        unsigned hash = 0;
+        for (auto const& family_value : key.font_family->as_value_list().values()) {
+            if (family_value->is_keyword())
+                hash = pair_int_hash(hash, to_underlying(family_value->as_keyword().keyword()));
+            else if (family_value->is_string())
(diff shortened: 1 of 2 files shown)


Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1998,6 +1998,10 @@ GC::Ref<ComputedProperties> StyleComputer::compute_properties(DOM::AbstractEleme
         auto property_id = static_cast<CSS::PropertyID>(i);
         auto value = cascaded_properties.property(property_id);
         auto inherited = ComputedProperties::Inherited::No;
+        auto important = Important::No;
+
+        if (value)
+            important = cascaded_properties.is_property_important(property_id) ? Important::Yes : Important::No;
```
awesomekling: Instead of this, could we add a `CascadedProperties` API that returns both the value and the important flag using a single lookup? :)

Libraries/LibWeb/CSS/FontComputer.cpp
```diff
@@ -359,7 +339,33 @@ RefPtr<Gfx::FontCascadeList const> FontComputer::font_matching_algorithm_impl(Fl
...
+    cache_key = pair_int_hash(cache_key, Traits<Web::CSSPixels>::hash(font_size));
+    cache_key = pair_int_hash(cache_key, font_slope);
+    cache_key = pair_int_hash(cache_key, Traits<double>::hash(font_weight));
+    cache_key = pair_int_hash(cache_key, Traits<double>::hash(font_width.value()));
+    for (auto const& [variation_name, variation_value] : font_variation_settings)
+        cache_key = pair_int_hash(cache_key, pair_int_hash(variation_name.hash(), Traits<double>::hash(variation_value)));
+
+    return m_computed_font_cache.ensure(cache_key, [&]() {
+        return compute_font_for_style_values_impl(font_family, font_size, font_slope, font_weight, font_width, font_variation_settings);
+    });
```
awesomekling, line 382: What happens when there's a hash collision? Where's the disambiguating equality check? :)

Calme1709: Good point - I have swapped back to using a cache key struct. This brings runtime back up to ~0.5% from 0.2% but it's still lower than the original 2.23% so that can stay as a potential optimization for our future selves.

Libraries/LibWeb/CSS/StyleComputer.cpp
Calme1709: Absolutely we can - done.

PR #7427, opened 2026-01-11
LibWeb: Resolve SVG `url()` references in shadow root scope

### LINKED ISSUE #3458: LibWeb: SVG `url()` references don't work when SVG is inside shadow DOM
### Summary

SVG elements using `url()` references (like gradients or masks) fail to work properly when SVG is moved into a Shadow DOM.
Reduced from [icones.js.org](https://icones.js.org/). Specifically [https://icones.js.org/collection/logos](https://icones.js.org/collection/logos), but it affects other SVGs as well.

### Operating system

Linux

### Steps to reproduce

N/A

### Expected behavior

SVG should look the same both inside and outside shadow DOM.

### Actual behavior

When SVG is moved into shadow DOM, the `url(#fill)` reference stops working and rectangle doesn't receive the gradient fill.
Same issue occurs with other URL-referenced elements like masks.

N/A

```html
<body>
    <svg>
        <defs>
            <lineargradient id="fill">
                <stop stop-color="crimson"/>
            </lineargradient>
        </defs>
        <rect width="100" height="100" fill="url(#fill)"/>
    </svg>
</body>
<script>
    const svg = document.querySelector("svg");
    const shadowRoot = document.body.attachShadow({
        mode: "open",
    });

    shadowRoot.innerHTML = svg.outerHTML;
    svg.remove();
</script>
```

```shell
N/A
```

<table>
<tr>
<th>Chrome</th>
<th>Ladybird</th>
</tr>
<tr>
<td>

Image

</td>
<td>

Image

</td>
</tr>
</table>

_No response_

### Contribute a patch?

aplefull:
Added a search in the containing shadow root when resolving SVG `url()` references

Fixes #3458

### DIFF
--- a/Libraries/LibWeb/SVG/SVGGraphicsElement.h
+++ b/Libraries/LibWeb/SVG/SVGGraphicsElement.h
@@ -101,8 +101,15 @@ protected:
         }
         if (!fragment.has_value())
             return {};
-        if (auto node = document().get_element_by_id(*fragment); node && is<T>(*node))
-            return static_cast<T&>(*node);
+        if (auto node = as_if<T>(document().get_element_by_id(*fragment).ptr()))
+            return *node;
+
+        auto containing_shadow = containing_shadow_root();
+        if (containing_shadow) {
+            if (auto node = as_if<T>(containing_shadow->get_element_by_id(*fragment).ptr()))
+                return *node;
+        }
+
         return {};
     }
 
--- a/Tests/LibWeb/Ref/expected/svg-url-in-shadow-root-ref.html
+++ b/Tests/LibWeb/Ref/expected/svg-url-in-shadow-root-ref.html
@@ -1,0 +1,11 @@
+<!DOCTYPE html>
+<body>
+    <div></div>
+</body>
+<style>
+    div {
+        width: 100px;
+        height: 100px;
+        background: crimson;
+    }
+</style>
--- a/Tests/LibWeb/Ref/input/svg-url-in-shadow-root.html
+++ b/Tests/LibWeb/Ref/input/svg-url-in-shadow-root.html
@@ -1,0 +1,23 @@
+<!DOCTYPE html>
+<head>
+<link rel="match" href="../expected/svg-url-in-shadow-root-ref.html" />
+</head>
+<body>
+    <svg>
+        <defs>
+            <lineargradient id="fill">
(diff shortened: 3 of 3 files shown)


Libraries/LibWeb/SVG/SVGGraphicsElement.h
```diff
@@ -103,6 +103,14 @@ class WEB_API SVGGraphicsElement : public SVGElement {
             return {};
         if (auto node = document().get_element_by_id(*fragment); node && is<T>(*node))
             return static_cast<T&>(*node);
+
+        auto containing_shadow = containing_shadow_root();
+        if (containing_shadow) {
+            auto node = containing_shadow->get_element_by_id(*fragment);
+            if (node && is<T>(*node))
+                return static_cast<T&>(*node);
```
AtkinsSJ: Nice catch, thanks for fixing this!

Instead of `is<T>()` and then `static_cast<T>()`, use `as_if<T>()` which will return null if the cast fails. Something like this:

```c++
if (auto* node = as_if<T>(containing_shadow->get_element_by_id(*fragment)))
    return node;
```

aplefull: Oh, love the function name :) 

Changed it, but had to do it slightly differently because `get_element_by_id()` returns `GC::Ptr<Element>`, and `as_if` didn't like it

Libraries/LibWeb/SVG/SVGGraphicsElement.h
```diff
@@ -101,8 +101,15 @@ class WEB_API SVGGraphicsElement : public SVGElement {
         }
         if (!fragment.has_value())
             return {};
-        if (auto node = document().get_element_by_id(*fragment); node && is<T>(*node))
-            return static_cast<T&>(*node);
+        if (auto node = document().get_element_by_id(*fragment))
+            return as_if<T>(*node);
```
gmta: This will now return an empty `GC::Ptr` if it's not of type `T`, which is probably not what you meant? Try this:

```cpp
if (auto node = as_if<T>(document().get_element_by_id(*fragment).ptr()))
    return *node;
```

aplefull: Yup, should be good now, thank you!

AtkinsSJ (approved):

PR #7430, opened 2026-01-11
LibHTTP+LibWeb: Enable disk caching in WPT, improve heuristic caching, don't revalidate from WebContent

trflynn89:
The gist of this PR is that we no longer do cache revalidation from WebContent. We already have an RFC-complaint revalidation implementation in RequestServer, so having another in WebContent (that was less compliant) is a bit overkill. When a memory cache entry in WebContent is no longer fresh, we evict it and forward the request to RequestServer. It will then decide if a disk cache entry is sufficient or needs revalidation.

The wrench is that WPT has tests for revalidation, but we previously ran WPT with disk caching disabled. This was just due to using the `--force-new-process` flag (c14dc77349eca6f9cddd7fbb1d8e216cc9fa6cf8). So this PR also contains a change to allow using the disk cache during WPT.

This had a bit of a hand-wavy effect on WPT. We gain some legit passes with our disk cache:
```
▶ Unexpected subtest result in /fetch/http-cache/heuristic.any.html:
  └ PASS [expected FAIL] HTTP cache reuses a 204 No Content response with Last-Modified based upon heuristic freshness

▶ Unexpected subtest result in /fetch/http-cache/freshness.any.html:
  └ PASS [expected FAIL] HTTP cache stores a response with Cache-Control: no-cache, but revalidates upon use, even with max-age and Expires

▶ Unexpected subtest result in /fetch/http-cache/status.any.html:
  └ PASS [expected FAIL] HTTP cache avoids going to the network if it has a fresh 204 response
```

But we also gain some failures. These are all in worker and sharedworker tests. e.g.:
```
  ▶ Unexpected subtest result in /fetch/http-cache/invalidate.any.sharedworker.html:
  │ FAIL [expected PASS] HTTP cache invalidates after a successful response from a POST
  │   → assert_equals: Response 3 comes from cache expected 3 but got 1
```

I checked each of these failures and they actually now all match the non-worker results. So as far as I can tell, workers are now just hitting the same caching paths that non-workers were hitting, and these failures aren't actually new.

### DIFF
--- a/Libraries/LibWeb/Fetch/Fetching/Fetching.cpp
+++ b/Libraries/LibWeb/Fetch/Fetching/Fetching.cpp
@@ -1521,7 +1521,6 @@ GC::Ref<PendingResponse> http_network_or_cache_fetch(JS::Realm& realm, Infrastru
     RefPtr<HTTP::MemoryCache> http_cache;
 
     // 7. Let the revalidatingFlag be unset.
-    auto revalidating_flag = RefCountedFlag::create(false);
 
     auto include_credentials = IncludeCredentials::No;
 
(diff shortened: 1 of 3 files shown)


gmta:
> The SQLite database does not handle this concurrent read/write access well.

Hm, we're not necessarily limiting ourselves to a single RequestServer process; I feel like if we end up doing that for normal Ladybird usage as well, we hit this issue anyway? What are the issues that you saw?

trflynn89:
We end up with a bunch of `Database error: database is locked` error messages (which we currently crash on) from various locations.

Having only a single UI and RS process was something we actively worked towards, though it has this `--force-new-process` avoids that. An alternate approach might be to store the cache in-memory in RS instead when this flag is present, which is how we handle cookies in the UI process currently.

gmta:
> We end up with a bunch of `Database error: database is locked` error messages (which we currently crash on) from various locations.

Right - that usually comes down to using the WAL, which I believe we never enable, and making sure we use transactions where applicable. SQLite should be able to deal with multiprocess access pretty well?

trflynn89:
I don't totally remember, but with WAL, won't we still get busy errors if 2 RequestServer processes try to write an entry at the same time?

I think even if the sqlite part is totally concurrency safe, we will then have issues with the cache entry files. If 2 RequestServer instances are trying to cache a `GET https://example.com/foo.png` response at the same time, they will end up writing to the same file on disk and we'll have a bad file.

gmta:
> I don't totally remember, but with WAL, won't we still get busy errors if 2 RequestServer processes try to write an entry at the same time?

That write time should be negligible; iirc you get that "database is locked" error after a specific timeout.

> If 2 RequestServer instances are trying to cache a GET https://example.com/foo.png response at the same time, they will end up writing to the same file on disk and we'll have a bad file.

I guess the behavior should/could be identical to handling this within a single process, but involve the database? I.e. mark a cache entry as being in process of retrieval and either decide on waiting for the result or just don't use that cache at all at that time.

gmta:
At any rate, if you feel it's too finnicky to enable WAL and deal with an eventual fallout of cache corruption, feel free to keep the partitioning. I'm currently not up to review the code myself anyway 😅

trflynn89:
Alrighty, so I tried landing somewhere in the middle - I enabled WAL, set a busy timeout of 5 seconds, and partitioned just the cache files themselves (so a single INDEX database, but I added an extra key to prevent concurrent cache file access). That seemed to work fine, until on one run I managed to hit a "database is locked" error. From what I've read, sqlite's busy wait is implemented as unfair, so it's possible for one process to just stay stuck waiting.

So I've kept the total partitioning scheme for now and added a FIXME :sweat_smile:

ADKaster:
commit [LibHTTP+LibWebView+RequestServer: Allow using the disk cache during WPT](https://github.com/LadybirdBrowser/ladybird/commit/bc1cafc71657bbd00348cda4615ac026d927efc3) breaks RequestServer on Windows, because Core::EventLoopManagerWindows::register_signal() is a VERIFY_NOT_REACHED().

PR #7439, opened 2026-01-12
LibWeb: A few `font-*` property fixes

Calme1709:
This PR includes some fixes to parsing and serialization of `font-*` properties which collectively gain us ~40 WPT passes - see individual commits for details.

### DIFF
--- a/Libraries/LibWeb/CSS/Interpolation.cpp
+++ b/Libraries/LibWeb/CSS/Interpolation.cpp
@@ -696,9 +696,9 @@ ValueComparingRefPtr<StyleValue const> interpolate_property(DOM::Element& elemen
         }
 
         if (property_id == PropertyID::FontStyle) {
-            auto static oblique_0deg_value = FontStyleStyleValue::create(FontStyle::Oblique, AngleStyleValue::create(Angle::make_degrees(0)));
-            auto from_value = from->as_font_style().font_style() == FontStyle::Normal ? oblique_0deg_value : from;
-            auto to_value = to->as_font_style().font_style() == FontStyle::Normal ? oblique_0deg_value : to;
+            auto static oblique_0deg_value = FontStyleStyleValue::create(FontStyleKeyword::Oblique, AngleStyleValue::create(Angle::make_degrees(0)));
+            auto from_value = from->as_font_style().font_style() == FontStyleKeyword::Normal ? oblique_0deg_value : from;
+            auto to_value = to->as_font_style().font_style() == FontStyleKeyword::Normal ? oblique_0deg_value : to;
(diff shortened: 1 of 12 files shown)


Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
```diff
@@ -2687,12 +2687,12 @@ RefPtr<StyleValue const> Parser::parse_font_value(TokenStream<ComponentValue>& t
             line_height.release_nonnull(),
 
             // Reset implicitly
-            initial_value,                                   // font-feature-settings
-            property_initial_value(PropertyID::FontKerning), // font-kerning,
-            initial_value,                                   // font-language-override
-                                                             // FIXME: font-optical-sizing,
-                                                             // FIXME: font-size-adjust,
-            initial_value,                                   // font-variation-settings
+            property_initial_value(PropertyID::FontFeatureSettings),   // font-feature-settings
```
AtkinsSJ, line 2689: I think `initial_value` is now unused and can be removed.

Libraries/LibWeb/CSS/Interpolation.cpp
```diff
@@ -1835,13 +1835,13 @@ static RefPtr<StyleValue const> interpolate_value_impl(DOM::Element& element, Ca
         if (!interpolated_font_style)
             return {};
         if (from_font_style.angle() && to_font_style.angle()) {
-            auto interpolated_angle = interpolate_value(element, calculation_context, *from_font_style.angle(), *to_font_style.angle(), delta, allow_discrete);
+            auto interpolated_angle = interpolate_value(element, { .accepted_type_ranges = { { ValueType::Angle, { -90, 90 } } } }, *from_font_style.angle(), *to_font_style.angle(), delta, allow_discrete);
```
AtkinsSJ, line 1838: Not your fault, but it's unfortunate that this "-90 to 90 degrees" gets repeated in a few places. It'd be nice to somehow use the same context definition anywhere it's needed. I have no idea how we'd do that though. :sweat_smile:

Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
Calme1709: Done.

Libraries/LibWeb/CSS/Interpolation.cpp
Calme1709: Potentially we could add a `CalculationContext::for_special_context`? Out of scope for this PR I think.

AtkinsSJ: Yeah it's out of scope, just thinking out loud.

Issue #6781, opened 2025-11-10
LibWeb: Support clickable `<a>` tags in SVGs


dosisod:
### Summary

Currently `<a>` tags in SVGs are not clickable, and do not navigate to other pages.

### Operating system

Linux

### Steps to reproduce

1. Go to https://svgwg.org/svg2-draft/images/linking/link01.svg
2. Click the red ellipse

### Expected behavior

It should navigate to https://www.w3.org/

### Actual behavior

It does nothing.

### URL for a reduced test case

https://svgwg.org/svg2-draft/images/linking/link01.svg

### HTML/SVG/etc. source for a reduced test case

```html
N/A
```

### Log output and (if possible) backtrace

```shell
None
```

### Screenshots or screen recordings

Before clicking red ellipse:

<img width="638" height="279" alt="Image" src="https://github.com/user-attachments/assets/9a6d0600-71ac-4518-bf41-5c853662d823" />

After clicking:

<img width="569" height="375" alt="Image" src="https://github.com/user-attachments/assets/9f53681c-ba11-475e-b081-768a107a3d92" />

### Contribute a patch?

- [x] I’ll contribute a patch for this myself.

dosisod:
I'm currently working on a PR for this. The code is functional, but requires some hacks due to IDL conflicts.

The issue is that the IDL generator doesn't support the [`ReflectSetter`](https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#xattr-reflectsetter) extended attribute. This is currently causing a conflict because the `href` attributes are specified in multiple places:

[SVGAElement.idl](https://svgwg.org/svg2-draft/linking.html#InterfaceSVGAElement)

```idl
[Exposed=Window]
interface SVGAElement : SVGGraphicsElement {
  // ...
};

SVGAElement includes SVGURIReference;
SVGAElement includes HTMLHyperlinkElementUtils;
```

[SVGURIReference.idl](https://svgwg.org/svg2-draft/types.html#InterfaceSVGURIReference)

```idl
interface mixin SVGURIReference {
  [SameObject] readonly attribute SVGAnimatedString href;  // <----- href specified here
};
```

[HTMLHyperlinkElementUtils](https://html.spec.whatwg.org/multipage/links.html#htmlhyperlinkelementutils)

```idl
interface mixin HTMLHyperlinkElementUtils {
  [CEReactions, ReflectSetter] stringifier attribute USVString href;  // <---- and here
};
```

[...]

PR #7445, opened 2026-01-12
LibWeb/SVG: Implement activation behavior for `SVGAElement`

tcl3:
This makes navigation work when clicking on SVG `<a>` elements.

I had to move `follow_the_hyperlink()` and other hyperlink navigation logic from HTMLHyperlinkElementUtils to `DOM::Element` so that it could be shared by `HTMLAnchorElement` and `SVGAElement`. `SVGAElement` is supposed to implement `HTMLHyperlinkElementUtils` but can't as it stands because both `HTMLHyperlinkElementUtils` and `SVGURIReference` have a `href()` method that causes a naming conflict. See https://github.com/w3c/svgwg/issues/312 for further details.

We also gain ~300 WPT subtests in the `referrer-policy` directory.

Closes: #6781

### DIFF
--- a/Libraries/LibWeb/DOM/Element.cpp
+++ b/Libraries/LibWeb/DOM/Element.cpp
@@ -211,8 +211,9 @@ String Element::get_an_elements_target(Optional<String> target) const
         }
         // 2. Otherwise, if element's node document contains a base element with a target attribute,
         //    set target to the value of the target attribute of the first such base element.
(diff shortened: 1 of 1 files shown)


tcl3:
Closes #6781.

I wasn't aware of this issue until now :sweat_smile:.

shannonbooth:
Probably we will end up moving things around again if https://github.com/whatwg/html/pull/12066 completes 😅

shannonbooth (approved):

PR #7447, opened 2026-01-12
test-web: Structured output, improved progress view, stability fixes

awesomekling:
See individual commits.

Demo:

https://github.com/user-attachments/assets/eb773268-5646-4311-9ba7-5da1215a4d72

### DIFF
--- a/Tests/LibWeb/test-web/main.cpp
+++ b/Tests/LibWeb/test-web/main.cpp
@@ -471,14 +471,16 @@ static ErrorOr<void> generate_result_files(Vector<TestCompletion> const& non_pas
     js.append("\n  ]\n};\n"sv);
 
     auto js_path = LexicalPath::join(app.results_directory, "results.js"sv).string();
-    auto js_file = TRY(Core::File::open(js_path, Core::File::OpenMode::Write));
+    auto js_file = TRY(Core::File::open(js_path, Core::File::OpenMode::Write | Core::File::OpenMode::Truncate));
     TRY(js_file->write_until_depleted(js.string_view().bytes()));
 
     // Copy index.html from source tree
     auto source_html_path = LexicalPath::join(app.test_root_path, "test-web/results-index.html"sv).string();
     auto dest_html_path = LexicalPath::join(app.results_directory, "index.html"sv).string();
-    (void)FileSystem::remove(dest_html_path, FileSystem::RecursionMode::Disallowed);
-    TRY(FileSystem::copy_file_or_directory(dest_html_path, source_html_path));
+    auto source_html = TRY(Core::File::open(source_html_path, Core::File::OpenMode::Read));
+    auto html_contents = TRY(source_html->read_until_eof());
+    auto dest_html = TRY(Core::File::open(dest_html_path, Core::File::OpenMode::Write | Core::File::OpenMode::Truncate));
+    TRY(dest_html->write_until_depleted(html_contents));
 
     return {};
 }


Tests/LibWeb/test-web/main.cpp
```diff
@@ -188,6 +391,195 @@ if (!hasTestWaitClass()) {{
...
+
+    // Write header
+    TRY(html_file->write_until_depleted("<span class=\"ctx\">"sv));
+    TRY(html_file->write_formatted("--- {}\n", test.expectation_path));
+    TRY(html_file->write_formatted("+++ {}\n", test.expectation_path));
+    TRY(html_file->write_until_depleted("</span>"sv));
+
+    // Write hunks with colorization
+    for (auto const& hunk : hunks) {
+        TRY(html_file->write_formatted("<span class=\"hunk\">@@ -{},{} +{},{} @@</span>\n",
```
shannonbooth: This is ever so slightly different to the unified format header - we do have a formatter for this though, so hopefully should just be:


```suggestion
        TRY(html_file->write_formatted("<span class=\"hunk\">{}</span>\n", hunk.location);
```

Tests/LibWeb/test-web/main.cpp
```diff
@@ -188,6 +391,195 @@ if (!hasTestWaitClass()) {{
...
+            hunk.location.old_range.start_line, hunk.location.old_range.number_of_lines,
+            hunk.location.new_range.start_line, hunk.location.new_range.number_of_lines));
+
+        for (auto const& line : hunk.lines) {
+            // HTML-escape the content
+            StringBuilder escaped;
```
shannonbooth: We have escape_html_entities which can be used here

R-Goc:
The windows failure is real. The STDOUT/ERR_FILENO should be redfined to convert to an int with to_fd. This whole macro is ugly. Realistically we should be callling GetStdHandle and casting that to an int with to_fd. But then we need to write a wrapper for it and call that inside the macro. So I see why it was done like this. So the simple solution is wrapping the existing thing in to_fd(), and the nice one is wrapping GetStdHandle converting stdout,stderr,stdin to the windows equivalents and returning the result of that. This doesn't work for other FILE streams but we don't care about it.

Tests/LibWeb/test-web/main.cpp
```diff
@@ -188,6 +391,195 @@ if (!hasTestWaitClass()) {{
...
+static ErrorOr<void> generate_result_files(Vector<TestCompletion> const& non_passing_tests)
+{
+    auto& app = Application::the();
+    if (app.results_directory.is_empty())
+        return {};
+
+    // Count results
+    size_t fail_count = 0;
+    size_t timeout_count = 0;
+    size_t crashed_count = 0;
```
shannonbooth, line 422: I think we should be able to skip this section and pass through this from the caller. https://github.com/LadybirdBrowser/ladybird/pull/7242/changes/6a8bd8532f6343545a3fcea787ad0ffc9e34c05d should make even slightly easier (which I just noticed I must have messed up a rebase on)

Tests/LibWeb/test-web/Fuzzy.cpp
```diff
@@ -31,18 +32,18 @@ bool fuzzy_screenshot_match(URL::URL const& test_url, URL::URL const& reference,
         return true;
     });
     if (!fuzzy_match.has_value()) {
-        warnln("{}: Screenshot mismatch: pixel error count {}, with maximum error {}. (No fuzzy config defined)", test_url, diff.pixel_error_count, diff.maximum_error);
+        add_deferred_warning(ByteString::formatted("{}: Screenshot mismatch: pixel error count {}, with maximum error {}. (No fuzzy config defined)", test_url, diff.pixel_error_count, diff.maximum_error));
```
nico, line 35: the diff for this commit is much smaller than what the commit message suggests (?)

Libraries/LibCore/System.h
```diff
@@ -35,9 +35,9 @@
 #    include <io.h>
 
 #    define O_CLOEXEC O_NOINHERIT
-#    define STDIN_FILENO _get_osfhandle(_fileno(stdin))
-#    define STDOUT_FILENO _get_osfhandle(_fileno(stdout))
-#    define STDERR_FILENO _get_osfhandle(_fileno(stderr))
+#    define STDIN_FILENO to_fd(reinterpret_cast<void*>(_get_osfhandle(_fileno(stdin))))
```
R-Goc, line 38: I forgot this returned an intptr_t. In this case cast directly to an int. to_fd() only does that.

awesomekling: Hmm, can you be more specific? I don't have a windows machine handy today, so it would be helpful if you tell me exactly what this should look like :)

R-Goc: So the thing is that valid kernel objects all have the upper 32bits of HANDLE zeroed. We make use of this currently to pass handles in places where on POSIX systems we pass FDs. So to_fd is just a reinterpret_cast to an int from HANDLE (which is just typedefed to a void*). So this can just be rewritten as reinterpret_cast<int>(_get_osfhandle(_fileno(stdout))). 
However, I am not actually sure that is correct for this use case, as the std handles are normally pseudo handles which do use the whole 64bits. This might be why there isn't a cast here already. I'd need to check what _get_osfhandle returns in this case.

R-Goc: Okay this is fine. It won't return the pseudohandle as it gets the underlying device like GetStdHandle would.

PR #7466, opened 2026-01-14
LibJS+LibUnicode: Implement finding time zone transitions for ZonedDateTime (and fix nanosecond edge case for time zone offsets)

Lubrsi:
```
Duration:
     +5.03s

Summary:
    Diff Tests:
        +7 ✅    -7 ❌   

Diff Tests:
    test/intl402/Temporal/ZonedDateTime/prototype/getTimeZoneTransition/nanoseconds-subtracted-or-added-at-dst-transition.js   ❌ -> ✅
    test/intl402/Temporal/ZonedDateTime/prototype/getTimeZoneTransition/result-type.js                                         ❌ -> ✅
    test/intl402/Temporal/ZonedDateTime/prototype/getTimeZoneTransition/rule-change-without-offset-transition.js               ❌ -> ✅
    test/intl402/Temporal/ZonedDateTime/prototype/getTimeZoneTransition/specific-tzdb-values.js                                ❌ -> ✅
    test/intl402/Temporal/ZonedDateTime/prototype/getTimeZoneTransition/subtract-second-and-nanosecond-from-last-transition.js ❌ -> ✅
    test/intl402/Temporal/ZonedDateTime/prototype/getTimeZoneTransition/transitions-close-together.js                          ❌ -> ✅
    test/intl402/Temporal/ZonedDateTime/prototype/offsetNanoseconds/nanoseconds-subtracted-or-added-at-dst-transition.js       ❌ -> ✅
```

### DIFF
--- a/Libraries/LibJS/Runtime/Date.cpp
+++ b/Libraries/LibJS/Runtime/Date.cpp
@@ -363,7 +363,7 @@ Crypto::SignedBigInteger get_utc_epoch_nanoseconds(Temporal::ISODateTime const& 
     return result;
 }
 
-static i64 clip_bigint_to_sane_time(Crypto::SignedBigInteger const& value)
+i64 clip_bigint_to_sane_time(Crypto::SignedBigInteger const& value)
 {
     static Crypto::SignedBigInteger const min_bigint { NumericLimits<i64>::min() };
     static Crypto::SignedBigInteger const max_bigint { NumericLimits<i64>::max() };
@@ -379,7 +379,7 @@ static i64 clip_bigint_to_sane_time(Crypto::SignedBigInteger const& value)
     return value.to_i64();
 }
 
-static i64 clip_double_to_sane_time(double value)
+i64 clip_double_to_sane_time(double value)
 {
     static constexpr auto min_double = static_cast<double>(NumericLimits<i64>::min());
     static constexpr auto max_double = static_cast<double>(NumericLimits<i64>::max());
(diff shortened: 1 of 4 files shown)


Libraries/LibJS/Runtime/Date.cpp
```diff
@@ -363,23 +363,6 @@ Crypto::SignedBigInteger get_utc_epoch_nanoseconds(Temporal::ISODateTime const&
...
-    if (value < min_bigint)
-        return NumericLimits<i64>::min();
-    if (value > max_bigint)
-        return NumericLimits<i64>::max();
-
-    // FIXME: Can we do this without string conversion?
-    return MUST(value.to_base(10)).to_number<i64>().value();
-}
-
 static i64 clip_double_to_sane_time(double value)
```
trflynn89: If we're moving one of these, let's move both - kinda awkward to have them in different files otherwise

Though instead of moving them, it's probably fine to just declare them in Date.h? Temporal already uses other helpers from there

Libraries/LibUnicode/TimeZone.cpp
```diff
@@ -219,4 +219,58 @@ Vector<TimeZoneOffset> disambiguated_time_zone_offsets(StringView time_zone, Uni
...
+                    .transition = AK::Duration::from_milliseconds(time_result),
+                };
+            }
+
+            current_icu_time = time_result;
+            include_current_time = false;
+            break;
+        }
+        default: {
+            VERIFY_NOT_REACHED();
```
trflynn89: Let's leave this off - then we'll get a compile error if a new transition rule is added but neglected to be handled

Libraries/LibJS/Runtime/Temporal/TimeZone.cpp
```diff
@@ -19,6 +19,7 @@
 #include <LibJS/Runtime/Temporal/TimeZone.h>
 #include <LibJS/Runtime/Temporal/ZonedDateTime.h>
 #include <LibJS/Runtime/VM.h>
+#include <LibUnicode/TimeZone.h>
```
trflynn89, line 22: Can we add in-tree tests?

PR #7467, opened 2026-01-14
WebDriver: Remove almost all dependency on spin_until and Promise await

Lubrsi:
(Reopening of https://github.com/LadybirdBrowser/ladybird/pull/6869 because I can't reopen that PR due to force pushing the branch whilst the PR was closed)

The remaining spin_until remains in taking screenshots, but attempting
to remove it caused errors in Skia which I couldn't figure out.

I don't think there's a neat way to split this into separate commits,
since it changes the fundamental infrastructure of how requests are
made and responded to.


Services/WebDriver/Session.h
```diff
@@ -106,7 +103,10 @@ class Session : public RefCounted<Session> {
 
     Web::WebDriver::PageLoadStrategy m_page_load_strategy { Web::WebDriver::PageLoadStrategy::Normal };
     Optional<JsonValue> m_timeouts_configuration;
-    bool m_strict_file_interactiblity { false };
+    bool m_strict_file_interactability { false };
+
+    RefPtr<Core::Promise<Empty, Web::WebDriver::Error>> m_close_promise;
```
ADKaster: is it worth declaring a typedef such as

```c++
template<typename T>
using WebDriverPromise = Core::Promise<T, Web::WebDriver::Error>;
```

Libraries/LibWeb/WebDriver/Client.cpp
```diff
@@ -17,12 +17,13 @@
...
 #include <AK/StringView.h>
 #include <AK/Time.h>
+#include <LibCore/Promise.h>
 #include <LibHTTP/Status.h>
 #include <LibWeb/WebDriver/Client.h>
 
 namespace Web::WebDriver {
 
-using RouteHandler = Response (*)(Client&, Parameters, JsonValue);
+using RouteHandler = NonnullRefPtr<Core::Promise<JsonValue, Error>> (*)(Client&, Parameters, JsonValue);
```
ADKaster: inside namespace Web::WebDriver, `Error` always refers to `Web::WebDriver::Error` rather than `AK::Error`, right? That distinction might not actually be that important but it is definitely confusing to first-time readers. Would a different name, explicit scoping, or the `WebDriverPromise` I referenced in another comment be worth it to avoid that confusion?

PR #7470, opened 2026-01-14
Tests: Improve robustness of `local-storage-usage-after-nav.html` test

tcl3:
A unique key is now used to ensure that the key can't interfere with any other test. The iframe is now set up programmatically rather than in HTML to ensure that `postMessage()` cannot fire before the `onmessage` handler is set up.

### DIFF
--- a/Tests/LibWeb/Text/input/HTML/local-storage-usage-after-nav.html
+++ b/Tests/LibWeb/Text/input/HTML/local-storage-usage-after-nav.html
@@ -2,11 +2,14 @@
 <script src="../include.js"></script>
 <script>
     asyncTest(done => {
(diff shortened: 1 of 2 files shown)


Tests/LibWeb/Text/input/HTML/local-storage-usage-after-nav.html
```diff
@@ -14,5 +13,9 @@
 
             done();
         };
+
+        const frame = document.createElement("iframe");
+        frame.src = "../../data/local-storage-iframe.html";
+        document.body.appendChild(frame);
```
gmta, line 22: Typo in commit message: `onessage`

PR #7472, opened 2026-01-14
LibDevTools: Support network request monitoring in Firefox DevTools

awesomekling:

I feel a little bit bad about the socket blocking hack, but without it we choke and disconnect from DevTools when viewing large HTTP responses..

Demo of our GitHub repo loaded in Ladybird:

image

### DIFF
--- a/Libraries/LibWebView/Application.cpp
+++ b/Libraries/LibWebView/Application.cpp
@@ -1363,4 +1363,22 @@ void Application::stop_listening_for_navigation_events(DevTools::TabDescription 
     }
 }
 
+void Application::did_connect_devtools_client(DevTools::TabDescription const& description) const
+{
+    auto view = ViewImplementation::find_view_by_id(description.id);
+    if (!view.has_value())
+        return;
+
+    view->did_connect_devtools_client();
+}
+
+void Application::did_disconnect_devtools_client(DevTools::TabDescription const& description) const
+{
+    auto view = ViewImplementation::find_view_by_id(description.id);
+    if (!view.has_value())
+        return;
+
+    view->did_disconnect_devtools_client();
+}
+
 }
--- a/Services/WebContent/PageClient.cpp
+++ b/Services/WebContent/PageClient.cpp
@@ -763,7 +763,20 @@ void PageClient::page_did_receive_network_response_headers(u64 request_id, u32 s
 
 void PageClient::page_did_receive_network_response_body(u64 request_id, ReadonlyBytes data)
 {
+    if (!has_devtools_client())
+        return;
     client().async_did_receive_network_response_body(m_id, request_id, data);
+}
+
+void PageClient::did_connect_devtools_client()
+{
+    ++m_devtools_client_count;
+}
+
+void PageClient::did_disconnect_devtools_client()
+{
+    VERIFY(m_devtools_client_count > 0);
+    --m_devtools_client_count;
 }
 
 void PageClient::page_did_finish_network_request(u64 request_id, u64 body_size, Requests::RequestTimingInfo const& timing_info, Optional<Requests::NetworkError> const& network_error)
--- a/Libraries/LibDevTools/Actors/FrameActor.cpp
+++ b/Libraries/LibDevTools/Actors/FrameActor.cpp
@@ -38,6 +38,11 @@ FrameActor::FrameActor(DevToolsServer& devtools, String name, WeakPtr<TabActor> 
     , m_accessibility(move(accessibility))
 {
     if (auto tab = m_tab.strong_ref()) {
+        // NB: We must notify WebContent that DevTools is connected before setting up listeners,
+        //     so that WebContent knows to start sending network response bodies over IPC.
+        //     IPC messages are processed in order, so this is guaranteed to arrive first.
+        devtools.delegate().did_connect_devtools_client(tab->description());
+
         devtools.delegate().listen_for_console_messages(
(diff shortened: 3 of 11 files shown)


Libraries/LibDevTools/Actors/WatcherActor.cpp
```diff
@@ -67,6 +68,15 @@ void WatcherActor::handle_message(Message const& message)
         return;
     }
 
+    if (message.type == "getNetworkParentActor"sv) {
```
trflynn89: nit: the messages here were previously alphabetical

Libraries/LibWebView/WebContentClient.cpp
```diff
@@ -434,6 +434,30 @@ void WebContentClient::did_get_js_console_messages(u64 page_id, i32 start_index,
...
+        if (view->on_network_request_started)
+            view->on_network_request_started(request_id, url, method, request_headers);
+    }
+}
+
+void WebContentClient::did_receive_network_response_headers(u64 page_id, u64 request_id, u32 status_code, Optional<String> reason_phrase, Vector<HTTP::Header> response_headers)
+{
+    if (auto view = view_for_page_id(page_id); view.has_value()) {
+        if (view->on_network_response_headers_received)
+            view->on_network_response_headers_received(request_id, status_code, reason_phrase, response_headers);
```
trflynn89: Up to you if it ends up feeling more comfy, but on the receiving end of IPC, I've been taking the response headers and creating a `HeaderList` (which we can't transfer over IPC directly right now because it's ref-counted):
```c++
HTTP::HeaderList::create(move(respose_headers));
```

That gives you access to some convenience methods, and avoids copying header lists. But if you don't need any of that, it's probably six of one / half a dozen of the other.

Libraries/LibWebView/Application.cpp
```diff
@@ -1337,4 +1337,32 @@ void Application::stop_listening_for_network_events(DevTools::TabDescription con
     view->on_network_request_finished = nullptr;
 }
 
+void Application::listen_for_navigation_events(DevTools::TabDescription const& description, OnNavigationStarted on_started, OnNavigationFinished on_finished) const
+{
+    auto view = ViewImplementation::find_view_by_id(description.id);
+    if (!view.has_value())
+        return;
+
+    view->on_load_start = [on_started = move(on_started)](URL::URL const& url, bool) {
```
trflynn89: I think in this case, we are now replacing the observers set up by the UI?

Services/WebContent/PageClient.cpp
```diff
@@ -751,16 +751,21 @@ void PageClient::received_message_from_web_ui(String const& name, JS::Value data
...
 }
 
 void PageClient::page_did_receive_network_response_headers(u64 request_id, u32 status_code, Optional<String> reason_phrase, Vector<HTTP::Header> const& response_headers)
 {
     client().async_did_receive_network_response_headers(m_id, request_id, status_code, move(reason_phrase), response_headers);
 }
 
+void PageClient::page_did_receive_network_response_body(u64 request_id, ReadonlyBytes data)
+{
+    client().async_did_receive_network_response_body(m_id, request_id, data);
```
trflynn89, line 768: I'm a little concerned about sending response body data over IPC all the time when DevTools isn't opened. Can we avoid these when there isn't a DevTools instance open on this tab?

Libraries/LibDevTools/Actors/NetworkEventActor.h
```diff
@@ -51,15 +52,19 @@ class DEVTOOLS_API NetworkEventActor final : public Actor {
...
     Optional<String> m_reason_phrase;
     Vector<HTTP::Header> m_response_headers;
 
+    ByteBuffer m_response_body;
     u64 m_body_size { 0 };
     Requests::RequestTimingInfo m_timing_info {};
     Optional<Requests::NetworkError> m_network_error;
     bool m_complete { false };
+
+    static constexpr size_t MAX_RESPONSE_BODY_SIZE = 10 * 1024 * 1024; // 10MB limit
```
trflynn89: I think you can just do `10 * MiB` here

Libraries/LibDevTools/Connection.cpp
```diff
@@ -42,7 +42,13 @@ void Connection::send_message(JsonValue const& message)
             dbgln("\x1b[1;32m<<\x1b[0m {}", serialized);
     }
 
-    if (m_socket->write_formatted("{}:{}", serialized.byte_count(), serialized).is_error()) {
+    // Temporarily enable blocking mode for large writes to avoid EAGAIN
+    (void)m_socket->set_blocking(true);
+    auto result = m_socket->write_until_depleted(MUST(String::formatted("{}:{}", serialized.byte_count(), serialized)));
```
trflynn89, line 47: The protocol has a "bulk" data packet type. I wonder if we're supposed to be using that for large messages? Not sure what it actually is, just remember seeing it.
(Doesn't need to be done here ofc)

awesomekling: It's my understanding that Firefox uses those for heap snapshots and profiler data, but not for responses. Will come back to it when we look at those features :)

Libraries/LibWebView/WebContentClient.cpp
awesomekling: Nice tip! Yeah feels kinda 6 vs half dozen at the moment so I'll leave it.

PR #7473, opened 2026-01-15
LibMedia+LibWeb: Use range requests to fulfill media data

Zaggy1024:
This makes media playback able to start without having to wait for data to sequentially download, especially when seeking the media to a timestamp residing in data that hasn't loaded yet.

Initially, the HTMLMediaElement will request the file without range a range request. Then, if the IncrementallyPopulatedStream finds that it needs data that is not yet available, it will decide whether to wait for that data to be received through the current request, or start a new request that is closer to the required data.

In order to make this work well, a few things had to be changed/fixed:
- `RedBlackTree` didn't have any non-const methods to find nodes before/after a key.
- `FetchController::stop_fetch()` did not stop requests, so any ongoing media fetches would continue until they reached the end of the body. `FetchController` now stores a reference to a `Requests::Request` if the fetch went over the network, which is used to stop the request later. Also, the original `FetchController` is now kept when copying `FetchParams` as part of the spec steps in `http_network_or_cache_fetch` so that the calling code can cancel any current underlying fetch.
- When stopping requests, RequestServer would receive SIGPIPE, which would occasionally cause it to crash entirely. In order to fix that, RequestPipe now uses sockets on all platforms, passing the `MSG_NOSIGNAL` flag to avoid triggering a signal.
- nghttp2 (used by cURL) would close connections if DATA frames were received after a stream is half-closed, which was triggered very often by nginx servers which would take a while to stop sending data after a stream was reset. Upstream, the check for this has been removed, so we now have a patch to do so in our vcpkg overlays until that is released.

### DIFF
--- a/Libraries/LibHTTP/Header.cpp
+++ b/Libraries/LibHTTP/Header.cpp
@@ -31,6 +31,7 @@ Optional<Vector<ByteString>> Header::extract_header_values() const
 
     // For now we only parse some headers that are of the ABNF list form "#something"
     if (name.is_one_of_ignoring_ascii_case(
+            "Accept-Ranges"sv,
             "Access-Control-Request-Headers"sv,
             "Access-Control-Expose-Headers"sv,
             "Access-Control-Allow-Headers"sv,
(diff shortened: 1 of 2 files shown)


Zaggy1024:
It seems there's some issue with a double free of the empty ByteString on this branch somehow. Trying to figure out why and how to fix it.

Zaggy1024:
With #7575 merged, restarting the requests no longer causes crashes. 🎉

PR #7475, opened 2026-01-15
Tests/LibWeb: Support WPT variants

gmta:
Please ignore the huge line count which is the result of rebaselining WPT tests with variants:

* We now support WPT's test variants (e.g. `<meta name="variant" content="?param=value">`) which effectively replaces a test by multiple variants to run instead. This necessitated a couple of changes to `test-web`, e.g. we can no longer have a pointer or reference to test structs if we're modifying the vector's storage during test runs. I've chosen to dynamically add variants instead of prescanning them since it follows suit in what we do for other WPT meta tags, such as fuzzy matching. As a result, a test run can end up with more tests being executed than initially found.

* Missing expectation files are now shown as diffs in the HTML report as well.

* The live display of test views running tests almost immediately showed `1s` as the duration. Second displays usually only jump to `1s` once a full second has passed, so I changed the display to truncated seconds.

Before:

```
Pass: 5584, Fail: 0, Skipped: 229, Timeout: 0, Crashed: 0
```

After:

```
Pass: 5699, Fail: 0, Skipped: 229, Timeout: 0, Crashed: 0
```

### DIFF
--- a/Tests/LibWeb/test-web/main.cpp
+++ b/Tests/LibWeb/test-web/main.cpp
@@ -122,7 +122,7 @@ static void render_live_display()
         if (i < s_view_display_states.size()) {
             auto const& state = s_view_display_states[i];
             if (state.active && state.pid > 0) {
-                auto duration = (now - state.start_time).to_seconds();
+                auto duration = (now - state.start_time).to_truncated_seconds();
                 // Format: ⏺ pid (Xs): name
                 auto prefix = ByteString::formatted("\033[33m⏺\033[0m {} ({}s): ", state.pid, duration);
                 // Note: prefix contains ANSI codes, so visible length is different


Tests/LibWeb/test-web/main.cpp
```diff
@@ -1264,11 +1442,12 @@ static ErrorOr<int> run_tests(Core::AnonymousBuffer const& theme, Web::DevicePix
     }
 
     // Generate result files (JSON data and HTML index)
-    if (auto result = generate_result_files(non_passing_tests); result.is_error())
+    if (auto result = generate_result_files(tests, non_passing_tests); result.is_error())
         warnln("Failed to generate result files: {}", result.error());
     else if (!app.results_directory.is_empty())
         outln("Results: file://{}/index.html", app.results_directory);
 
+    s_run_context = nullptr;
```
tcl3: nit: Could this be cleared with a `ScopeGuard` instead?

tcl3 (commented):
Very cool! Just one very minor suggestion from me.

Tests/LibWeb/test-web/main.cpp
gmta: Yes of course, done!

PR #7483, opened 2026-01-16
Meta+LibJS: Upgrade simdjson to the latest version

alimpfard:
5e0ee26e8bef854553dfe3a6cdf3ce508a926acd pulls in an old version of simdjson, this commit upgrades to the latest release.

### DIFF
--- a/Meta/check-flatpak.py
+++ b/Meta/check-flatpak.py
@@ -147,14 +147,14 @@ def check_vcpkg_vs_flatpak_versioning():
                         # Get the tag
                         # Replace '-' with '.': 76-1 vs 76.1
                         tag = str(source["tag"]).replace("-", ".")
-                        mismatch_found = match_and_update(name, tag)
+                        mismatch_found |= match_and_update(name, tag)
 
                         break
                     elif "branch" in source:
                         # Get the branch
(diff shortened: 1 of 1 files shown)


Libraries/LibJS/Runtime/JSONObject.cpp
```diff
@@ -760,7 +763,10 @@ static ThrowCompletionOr<Value> parse_simdjson_document(VM& vm, simdjson::ondema
         return parse_simdjson_array(vm, document);
     case simdjson::ondemand::json_type::object:
         return parse_simdjson_object(vm, document);
+    case simdjson::ondemand::json_type::unknown:
+        VERIFY_NOT_REACHED();
```
ADKaster: looks like some of the tests are hitting this assert.

alimpfard: Huh. I'll take a look at what that unknown thing is then 😅

alimpfard: Looks like it's "error", the docs also mention it giving you that for invalid tokens: https://github.com/simdjson/simdjson/blob/8b69401d8afde69252b0952a1e46983fcc9b5091/doc/basics.md?plain=1#L2768

PR #7489, opened 2026-01-16
AK: Introduce ThreadAssertions helpers for main-thread verification

shannonbooth:
As we begin adding more threading in the codebase, which I expect to grow,
e.g introducing a WebAudio audio thread, I figure it might be helpful to have
some assertions in place to try and catch mistakes made in such cases,
such as accidentally making a GC heap allocation off the main thread
or constructing a FlyString and mutating the FlyString table.

### DIFF
--- a/Libraries/LibWeb/HTML/MessagePort.cpp
+++ b/Libraries/LibWeb/HTML/MessagePort.cpp
@@ -7,6 +7,7 @@
  */
 
 #include <AK/ByteReader.h>
+#include <AK/MainThreadAssertions.h>
 #include <AK/MemoryStream.h>
 #include <LibCore/Socket.h>
 #include <LibCore/System.h>
@@ -151,6 +152,7 @@ WebIDL::ExceptionOr<void> MessagePort::transfer_receiving_steps(HTML::TransferDa
 
 void MessagePort::disentangle()
 {
+    VERIFY_ON_MAIN_THREAD();
     if (auto remote_port = m_remote_port) {
         // Set the pointers to null before disentangling the remote port to prevent infinite recursion here.
         m_remote_port->m_remote_port = nullptr;
@@ -328,6 +330,7 @@ void MessagePort::read_from_transport()
 
 void MessagePort::post_message_task_steps(SerializedTransferRecord& serialize_with_transfer_result)
 {
+    VERIFY_ON_MAIN_THREAD();
     VERIFY(m_enabled);
 
     // 1. Let finalTargetPort be the MessagePort in whose port message queue the task now finds itself.


AK/MainThreadAssertions.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+void initialize_main_thread()
+{
+    s_main_thread = std::this_thread::get_id();
+}
+
+bool is_main_thread()
+{
+    if (!s_main_thread.has_value())
+        return true;
```
konradekk, line 25: should throw/assert instead? (an incorrect/inconsistent state?)

shannonbooth: it’s valid - it may have not been initialized in the executable, or this is being called during static initialization before main()

konradekk: for the first case: i'd treat that as a mistake – something to back up my concerns: you should not use this check before initialization. as for the second one… not sure. 😣

jonbgamble:
I am using [something similar for webaudio render/control thread](https://github.com/jonbgamble/ladybird/blob/a35fc153229e85e4e1aee0e16d1a065feb003883/Libraries/LibWeb/WebAudio/Debug.h#L194), but a more general foundation for thread asserts would be nice.

I was thinking something like this could help readability in backtraces / lldb:

```cpp
ALWAYS_INLINE void set_current_thread_name(StringView name)
{
    Array<char, 16> buffer {};
    StringView trimmed = name.substring_view(0, buffer.size() - 1);
    trimmed.copy_characters_to_buffer(buffer.data(), buffer.size());
#if defined(AK_OS_LINUX)
    (void)pthread_setname_np(pthread_self(), buffer.data());
#elif defined(AK_OS_MACOS)
    (void)pthread_setname_np(buffer.data());
#else
    (void)name;
#endif
}
```

shannonbooth:
Huh, I had a thought we had the ability to set thread name in our Thread implementation, but looks like that was back in the serenity days and we've yeeted it now. I'm not actually sure of a nice way to incorporate the control thread / render thread concepts into here since AK so low in the stack and shouldn't know anything about audio 🤔

jonbgamble:
Client code could mark a thread with a token, maybe just the name. The foundation code sets the thread_local and offers an is_current_thread(token) function. So clients make their own assert macros, but at least the pthread stuff is hidden.

jonbgamble:
I would love to see this. And in the future, "mark current thread as <x>" and "query is current thread <x>" capability. Locking down code paths to an owning thread with hard asserts (when possible) is a useful guardrail, but more importantly it's self-enforcing documentation for concurrency invariants.

PR #7490, opened 2026-01-16
Meta: Another round of BSD related fixes

cqundefine:
This almost all that is needed to build for OpenBSD and NetBSD, although at least one more PR will be needed to add the vcpkg triplets.

### DIFF
--- a/Meta/CMake/lagom_compile_options.cmake
+++ b/Meta/CMake/lagom_compile_options.cmake
@@ -69,8 +69,8 @@ endif()
 if (NOT WIN32 AND NOT APPLE AND NOT ENABLE_FUZZERS)
     # NOTE: Assume ELF
     # NOTE: --no-undefined is not compatible with clang sanitizer runtimes
-    # NOTE: FreeBSD doesn't export all symbols from LibC so we can't use --no-undefined
-    if ((CMAKE_CXX_COMPILER_ID MATCHES "Clang$" AND (ENABLE_ADDRESS_SANITIZER OR ENABLE_MEMORY_SANITIZER OR ENABLE_UNDEFINED_SANITIZER OR ENABLE_LAGOM_COVERAGE_COLLECTION)) OR CMAKE_SYSTEM_NAME MATCHES "FreeBSD")
+    # NOTE: Some BSDs don't export all symbols from LibC so we can't use --no-undefined
+    if ((CMAKE_CXX_COMPILER_ID MATCHES "Clang$" AND (ENABLE_ADDRESS_SANITIZER OR ENABLE_MEMORY_SANITIZER OR ENABLE_UNDEFINED_SANITIZER OR ENABLE_LAGOM_COVERAGE_COLLECTION)) OR BSD)
         add_link_options(LINKER:--allow-shlib-undefined)
         add_link_options(LINKER:-z,undefs)
     else()


Meta/Lagom/Tools/CodeGenerators/LibWeb/GenerateWindowOrWorkerInterfaces.cpp
```diff
@@ -437,6 +438,13 @@ ErrorOr<int> ladybird_main(Main::Arguments arguments)
     VERIFY(!paths.is_empty());
     VERIFY(!base_paths.is_empty());
 
+#if defined(AK_OS_OPENBSD)
+    // Increase the memory limit, because on OpenBSD the default limit is to low
+    // FIXME: Figure out why when building with sanitizers this requires ridiculous limits, as it fails even with 64GiB
```
ADKaster: what version of OpenBSD are you using?

The behavior of RLIMIT_DATA changed in 6.7 (2020) to not count PROT_NONE pages in this limit.

https://www.openbsd.org/67.html
> Stopped counting pages mapped as PROT_NONE against the RLIMIT_DATA limit, helping code which reserves large chunks of address space but populates it sparsely. 

Another difference might be that OpenBSD uses only RLIMIT_DATA for mmap and friends, whereas other OS's have the RLIMIT_AS knob to tweak. AFAIK ASAN maps an incredible amount of memory for its shadow memory, assuming that barely any of it will be used. Though looking at llvm, it seems that OpenBSD maintains their sanitizer runtime downstream.

The more interesting question to me is why RLIMIT_DATA trips up on this executable specifically. Is it because it mmap's hundreds of files into its memory space using `mmap(MAP_SHARED, fd)`, on top of whatever shenanigans ASAN is doing?

---

https://man.openbsd.org/getrlimit.2#RLIMIT_DATA
> RLIMIT_DATA The maximum size (in bytes) of the data segment for a process; this includes memory allocated via [malloc(3)](https://man.openbsd.org/malloc.3) and all other anonymous memory mapped via [mmap(2)](https://man.openbsd.org/mmap.2).

https://man7.org/linux/man-pages/man2/getrlimit.2.html
> RLIMIT_AS
              This is the maximum size of the process's virtual memory
              (address space).  The limit is specified in bytes, and is
              rounded down to the system page size.  This limit affects
              calls to [brk(2)](https://man7.org/linux/man-pages/man2/brk.2.html), [mmap(2)](https://man7.org/linux/man-pages/man2/mmap.2.html), and [mremap(2)](https://man7.org/linux/man-pages/man2/mremap.2.html), which fail with
              the error ENOMEM upon exceeding this limit.  In addition,
              automatic stack expansion fails (and generates a SIGSEGV
              that kills the process if no alternate stack has been made
              available via [sigaltstack(2)](https://man7.org/linux/man-pages/man2/sigaltstack.2.html)).  Since the value is a long,
              on machines with a 32-bit long either this limit is at most
              2 GiB, or this resource is unlimited.

> RLIMIT_DATA
              This is the maximum size of the process's data segment
              (initialized data, uninitialized data, and heap).  The
              limit is specified in bytes, and is rounded down to the
              system page size.  This limit affects calls to [brk(2)](https://man7.org/linux/man-pages/man2/brk.2.html),
              [sbrk(2)](https://man7.org/linux/man-pages/man2/sbrk.2.html), and (since Linux 4.7) [mmap(2)](https://man7.org/linux/man-pages/man2/mmap.2.html), which fail with the
              error ENOMEM upon encountering the soft limit of this
              resource.

ADKaster: oh wait just kidding, we don't mmap hundreds of files, we read in hundreds of files into memory as ByteStrings. 

```
    // Read in all IDL files, we must own the storage for all of these for the lifetime of the program
    Vector<ByteString> file_contents;
    for (ByteString const& path : paths) {
        auto file_or_error = Core::File::open(path, Core::File::OpenMode::Read);
        if (file_or_error.is_error()) {
            s_error_string = ByteString::formatted("Unable to open file {}", path);
            return Error::from_string_view(s_error_string.view());
        }
        auto file = file_or_error.release_value();
        auto string = MUST(file->read_until_eof());
        file_contents.append(ByteString(ReadonlyBytes(string)));
    }
    VERIFY(paths.size() == file_contents.size());
```

cqundefine: > what version of OpenBSD are you using?

I've been testing both on OpenBSD 7.7 and 7.8.

I've just checked on my install right now with the most bare bones C app and it gave me 1,5GB limits which would align with what I found in https://github.com/openbsd/src/blob/master/etc/etc.amd64/login.conf, which also sets the data limit for normal users to 1,5GB, probably as a security measure. A more permanent solution would be to run the build as a user that has higher limits, but I feel like this is simpler and also works just fine.

The really large value (the 64 TiB) actually comes from Linux, not OpenBSD, because for a moment I had this code built for each operating system, not sure if it actually matters on OpenBSD but I'm also not sure it matters that much as on every other OS this is infinite anyways and this is only a build utility. The fact that I observed it only Linux could very well be just caused by the shadow memory.

> oh wait just kidding, we don't mmap hundreds of files, we read in hundreds of files into memory as ByteStrings.

I actually didn't look into what was using so much memory, but looking at that I think increasing the limit makes even more sense.

Issue #7409, opened 2026-01-10
LibWeb: iframe sandbox flags cause srcdoc image request to not load


shannonbooth:
Found trying to reduce the terrible looking signup section on aftonblade.se

(specifically) https://www.aftonbladet.se/sportbladet/a/aJP2A2/massflykt-hotar-malmo-handboll-som-jagar-rivalsucce

For the following HTML:
```html
<iframe sandbox="allow-popups" width=100% height=644 style=border:none srcdoc="
[...]

Issue #3729, opened 2025-02-28
Pages don't load at all when scripting is disabled


rgret-dev:
Disabling scripting causes everything to stop loading. It worked previously and regressed around the December/January time frame.

Potentially related commit:
https://github.com/LadybirdBrowser/ladybird/commit/17d5dfe5974da1e5a7c962a0b6db990c9c806409

cc: @Gingeh

AtkinsSJ:
I've confirmed this is still the case.

PR #7491, opened 2026-01-16
LibWeb: Remove "can run script" checks from MainThreadVM hooks

Lubrsi:
This allows Promises (including internal ones, like in Streams) and FinalizationRegistry to work while scripting is disabled.

These were removed in the spec in support of WebDriver BiDi being able to enable and disable scripting.

https://github.com/whatwg/html/commit/905b7218

This allows Fetch to work while scripting is disabled, due to its use of Streams. This was noticeable if you disabled scripting in the UI, or had a sandboxed iframe with scripting disabled.

There is more work to use job settings instead of realm inside these hooks, but that's a separate issue.

Fixes https://github.com/LadybirdBrowser/ladybird/issues/7409 Fixes https://github.com/LadybirdBrowser/ladybird/issues/3729


Libraries/LibWeb/Bindings/MainThreadVM.cpp
```diff
@@ -254,27 +255,24 @@ void initialize_main_thread_vm(AgentType type)
...
-            // 6. If result is an abrupt completion, then report the exception given by result.[[Value]].
+            // 5. If result is an abrupt completion, then report the exception given by result.[[Value]].
             if (result.is_error())
                 HTML::report_exception(result, entry);
         }));
     };
 
     // 8.1.5.4.3 HostEnqueuePromiseJob(job, realm), https://html.spec.whatwg.org/multipage/webappapis.html#hostenqueuepromisejob
     // // https://whatpr.org/html/9893/webappapis.html#hostenqueuepromisejob
+    // FIXME: Update this for the latest version of the spec where it uses job settings instead.
```
shannonbooth: If I understand this FIXME properly, this is more-so just the fact that this is implementing the shadow realm proposal integration MR of https://whatpr.org/html/9893/webappapis.html#hostenqueuepromisejob linked above which uses realm instead of settings object since shadow realms do not have a settings object

shannonbooth: So this is sort of ahead of the specification instead of behind is what I'm trying to say 😅

Lubrsi: Oops, makes sense :^), removed the FIXMEs.

PR #7492, opened 2026-01-16
CMake+CI: Use the same preset names on every platform

ADKaster:
Starting with CMake 3.30 and CMakePresets version 9, the include field supports interesting macro expansions. We can now define platform specific presets in separate files and include them in the top-level CMakePresets.json, while keeping the same preset names across all platforms. This avoids some preset explosion at the cost of some mostly empty json files for each unix platform.

The CMake minimum required in the top-level and Lagom CMakeLists.txt have not been adjusted in this patch, as that would have the effect of changing the default policy versions and is a bit out of scope.

https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html#includes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,7 +60,7 @@ jobs:
 
           - os_name: 'Windows'
             arch: 'x86_64'
-            build_preset: 'Windows_Experimental_Sanitizer'
+            build_preset: 'Sanitizer'
             toolchain: 'ClangCL'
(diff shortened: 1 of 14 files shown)


ADKaster:
cc @ayeteadoe @cqundefine

gmta (approved):

cqundefine:
FWIW, checked on FreeBSD with this PR and it builds fine without any issues.

PR #7494, opened 2026-01-16
LibCrypto: Add SHAKE digest support

mikiubo:
Introduce a new SHAKE hash wrapper in LibCrypto backed by OpenSSL.

Wire cSHAKE128 and cSHAKE256 into WebCrypto.

Note that cSHAKE with non-empty functionName or customization is currently rejected due to OpenSSL EVP limitations.

This fixes WPT:
WebCryptoAPI/digest/cshake.tentative.https.any.html


Libraries/LibCrypto/Hash/SHAKE.cpp
```diff
@@ -0,0 +1,52 @@
...
+    if (wants_cshake) {
+        // FIXME: implements cSHAKE with non-empty N or S
+        return Error::from_string_literal(
+            "cSHAKE with non-empty N or S is not supported yet "
+            "(OpenSSL EVP limitation)");
+    }
+
+    if (length % 8 != 0)
+        return Error::from_string_literal(
+            "SHAKE output length must be a multiple of 8 bits");
```
gmta: Please use braces if the body of any `if`/`elseif`/`else` is longer than one line.

Libraries/LibWeb/Crypto/CryptoAlgorithms.cpp
```diff
@@ -631,6 +632,34 @@ JS::ThrowCompletionOr<NonnullOwnPtr<AlgorithmParams>> Ed448Params::from_value(JS
...
 
+inline JS::ThrowCompletionOr<Optional<ByteBuffer>>
+get_optional_buffer_source(
+    JS::VM& vm,
+    JS::Object const& object,
+    JS::PropertyKey const& name)
+{
+    auto maybe_buffer = Optional<ByteBuffer> {};
+
+    if (MUST(object.has_property(name))) {
```
gmta: Not sure why this is a `MUST` instead of a `TRY`, and can we move this to an early return? I.e.

```cpp
if (!MUST(object.has_property(name))
    return {}
```

That allows you to get rid of the `auto maybe_buffer = Optional<ByteBuffer> {};`.

Libraries/LibWeb/Crypto/CryptoAlgorithms.cpp
```diff
@@ -9634,4 +9682,47 @@ WebIDL::ExceptionOr<JS::Value> Argon2::get_key_length(AlgorithmParams const&)
...
+
+    // 2. Let functionName be the functionName member of normalizedAlgorithm if present or the empty octet string otherwise.
+    auto const& function_name = normalized_algorithm.function_name;
+
+    // 3. Let customization be the customization member of normalizedAlgorithm if present or the empty octet string otherwise.
+    auto const& customization = normalized_algorithm.customization;
+
+    auto const algorithm = [&]() {
+        // 4. If the name member of normalizedAlgorithm is a case-sensitive string match for "cSHAKE128":
+        if (normalized_algorithm.name == "cSHAKE128")
```
gmta: `"cSHAKE128"sv`, and below as well. Allows for `operator==` on two StringViews.

Libraries/LibWeb/Crypto/CryptoAlgorithms.cpp
mikiubo: I am not sure either:
I extract the same function that was inside the argon2 method below (to avoid code duplication).
I think the TRY is better and of course the early return is better.

I fixed as you suggest and i check the argon2 wpt to check if anything fails. (it seems ok).

Libraries/LibWeb/Crypto/CryptoAlgorithms.cpp
```diff
@@ -631,6 +632,31 @@ JS::ThrowCompletionOr<NonnullOwnPtr<AlgorithmParams>> Ed448Params::from_value(JS
...
+    auto value = TRY(object.get(name));
+
+    if (!value.is_object()
+        || !(is<JS::TypedArrayBase>(value.as_object())
+            || is<JS::ArrayBuffer>(value.as_object())
+            || is<JS::DataView>(value.as_object()))) {
+        return vm.throw_completion<JS::TypeError>(
+            JS::ErrorType::NotAnObjectOfType,
+            "BufferSource");
+    }
```
trflynn89: Let's use `WebIDL::is_buffer_source_type` instead. This function should also be `static`.

trflynn89: Also IMO there's a bunch of unnecessary line breaks being added here and many other places. Putting every argument and return value on its own line is really only helpful when the line gets particularly long.

e.g.:
```c++
static inline JS::ThrowCompletionOr<Optional<ByteBuffer>> get_optional_buffer_source(JS::VM& vm, JS::Object const& object, JS::PropertyKey const& name)
{
    if (!TRY(object.has_property(name)))
        return OptionalNone {};

    auto value = TRY(object.get(name));

    if (!WebIDL::is_buffer_source_type(value))
        return vm.throw_completion<JS::TypeError>(JS::ErrorType::NotAnObjectOfType, "BufferSource");

    return TRY_OR_THROW_OOM(vm, WebIDL::get_buffer_source_copy(value.as_object()));
}
```

Libraries/LibWeb/Crypto/CryptoAlgorithms.cpp
```diff
@@ -9634,4 +9679,47 @@ WebIDL::ExceptionOr<JS::Value> Argon2::get_key_length(AlgorithmParams const&)
...
+    // 4. Let result be the result of performing the cSHAKE128/cSHAKE256 function defined in Section 3 of [NIST-SP800-185]
+    // using message as the X input parameter,
+    // length as the L input parameter,
+    // functionName as the N input parameter,
+    // and customization as the S input parameter.
+    auto maybe_result = algorithm.digest(
+        data,
+        length,
+        customization.map([](auto const& value) { return value.span(); }),
+        function_name.map([](auto const& value) { return value.span(); }));
```
trflynn89, line 9695: (This is good as-is on multiple lines, for example)

Libraries/LibCrypto/Hash/SHAKE.cpp
```diff
@@ -0,0 +1,53 @@
...
+    Optional<ReadonlyBytes> function_name) const
+{
+    bool wants_cshake = (customization.has_value() && !customization->is_empty()) || (function_name.has_value() && !function_name->is_empty());
+
+    if (wants_cshake) {
+        // FIXME: implements cSHAKE with non-empty N or S
+        return Error::from_string_literal(
+            "cSHAKE with non-empty N or S is not supported yet "
+            "(OpenSSL EVP limitation)");
+    }
```
trflynn89, line 30:
```c++
if (wants_cshake) {
    // FIXME: Implement cSHAKE with non-empty N or S
    return Error::from_string_literal("cSHAKE with non-empty N or S is not supported yet (OpenSSL EVP limitation)");
}
```

etc.

Libraries/LibWeb/Crypto/CryptoAlgorithms.cpp
mikiubo: yes thanks!
i only copy paste that function blindly to avoid code duplication but i think using is_buffer_source_type is better solution.

i found ( and fixed) others unnecessary line breaks in the from_value method.

now i think is fixed
in case let me know

thanks

Libraries/LibCrypto/Hash/SHAKE.cpp
```diff
@@ -0,0 +1,51 @@
...
+    bool wants_cshake = (customization.has_value() && !customization->is_empty()) || (function_name.has_value() && !function_name->is_empty());
+
+    if (wants_cshake) {
+        // FIXME: Implement cSHAKE with non-empty N or S
+        return Error::from_string_literal("cSHAKE with non-empty N or S is not supported yet (OpenSSL EVP limitation)");
+    }
+
+    if (length % 8 != 0) {
+        return Error::from_string_literal(
+            "SHAKE output length must be a multiple of 8 bits");
```
trflynn89: Let's also place this on one line please

mikiubo: ok thanks
 i think i fixed if you find something else let me know
 (i dont know why CI is failing i changed only that lines but it seems that is not only my pr ?)

PR #7495, opened 2026-01-17
LibMedia: Fix byte/frame mismatch in conversion to buffer size

jonbgamble:
Note that `pa_buffer_attr.tlength` is in bytes, not frames (see `/usr/include/pulse/def.h`). I think the previous code asked for a buffer that is 1/8 what the latency calculation wants (for float32 stereo), resulting in too frequent write callbacks, higher wakeup/CPU churn, higher risk of crackles under load.


Zaggy1024:
This would explain a lot, I've noticed such crackling fairly frequently but I never re-checked this, thanks! 😅

PR #7496, opened 2026-01-17
LibWeb/HTML: Do not disable scripting for script based about: pages

### LINKED ISSUE #4203: `about:settings` requires JavaScript
As a spin-off from #3729 (Pages don't load at all when scripting is disabled), even if that bug was fixed, our `about:settings` page relies on JavaScript, meaning that users with scripting disabled can't change settings, for example, to re-enable scripting. 😅

I think the solution would be to force-enable scripting on internal pages. I think just doing so for `about:*` would have issues with `about:blank` and `about:srcdoc` though.

shannonbooth:
Fixes #4203

### DIFF
--- a/Libraries/LibWeb/HTML/Scripting/Environments.cpp
+++ b/Libraries/LibWeb/HTML/Scripting/Environments.cpp
@@ -298,7 +298,11 @@ bool is_scripting_enabled(JS::Realm const& realm)
 
     // The user has not disabled scripting for realm at this time. (User agents may provide users with the option to disable scripting globally, or in a finer-grained manner, e.g., on a per-origin basis, down to the level of individual realms.)
     auto const& document = as<HTML::Window>(realm.global_object()).associated_document();
-    if (!document.page().is_scripting_enabled())
+
+    // NB: about:settings and about:processes are internal pages using javascript, so we do not consider user configuration for these pages.
+    if (document.url() != URL::about_settings()
+        && document.url() != URL::about_processes()
+        && !document.page().is_scripting_enabled())
         return false;
 
     // Either settings's global object is not a Window object, or settings's global object's associated Document's active sandboxing flag set does not have its sandboxed scripts browsing context flag set.


AtkinsSJ:
Would it make sense to do this for all `about:` URLs? We don't want the same issue to pop up if we add others for listing processes etc.

shannonbooth:
> Would it make sense to do this for all `about:` URLs? We don't want the same issue to pop up if we add others for listing processes etc.

I think the problem is like you mentioned in the issue description is that documents with about:srcdoc for example would then become an issue by not having script being disabled. Another option if this becomes a thing in the future might be to reserve scheme which we use for internal pages like ladybird:settings or something like that (then we can safely use the ladybird scheme as an internal one)

trflynn89:
Internal pages like `about:settings` have a `WebUIConnection` in the WebContent process. Could we use the presence of that object instead of a hard-coded list? (You might hit the issue of fetching not working with scripting disabled, if a PR hasn't been made for that yet).

Barring that, a comment in `WebUI::create` saying to update this list would be nice.

shannonbooth:
> Internal pages like `about:settings` have a `WebUIConnection` in the WebContent process. Could we use the presence of that object instead of a hard-coded list? (You might hit the issue of fetching not working with scripting disabled, if a PR hasn't been made for that yet).

Luke has fixed this one :^) The spec updated for this, see: https://github.com/LadybirdBrowser/ladybird/commit/61a344a718295322d70750d431c350e67737396a

But you mean something like this? I don't think this works, but along these lines I guess:

```patch
diff --git a/Libraries/LibWeb/HTML/Scripting/Environments.cpp b/Libraries/LibWeb/HTML/Scripting/Environments.cpp
index 48828d00a72..f1581e70bfd 100644
--- a/Libraries/LibWeb/HTML/Scripting/Environments.cpp
+++ b/Libraries/LibWeb/HTML/Scripting/Environments.cpp
@@ -299,10 +299,9 @@ bool is_scripting_enabled(JS::Realm const& realm)
     // The user has not disabled scripting for realm at this time. (User agents may provide users with the option to disable scripting globally, or in a finer-grained manner, e.g., on a per-origin basis, down to the level of individual realms.)
     auto const& document = as<HTML::Window>(realm.global_object()).associated_document();
 
-    // NB: about:settings and about:processes are internal pages using javascript, so we do not consider user configuration for these pages.
-    if (document.url() != URL::about_settings()
-        && document.url() != URL::about_processes()
-        && !document.page().is_scripting_enabled())
+    // NB: Pages with a WebUI are internal pages relying javascript (e.g, about:settings),
+    //     so we do not consider user configuration of disalbing script for these pages.
+    if (!document.page().client().has_ladybird_internal_web_ui() && !document.page().is_scripting_enabled())
         return false;
 
     // Either settings's global object is not a Window object, or settings's global object's associated Document's active sandboxing flag set does not have its sandboxed scripts browsing context flag set.
diff --git a/Libraries/LibWeb/Page/Page.h b/Libraries/LibWeb/Page/Page.h
index e1dd9cca655..8db2c10a13e 100644
--- a/Libraries/LibWeb/Page/Page.h
+++ b/Libraries/LibWeb/Page/Page.h
@@ -441,6 +441,8 @@ public:
 
     virtual bool is_svg_page_client() const { return false; }
 
+    virtual bool has_ladybird_internal_web_ui() const { return false; }
+
 protected:
     virtual ~PageClient() = default;
 };
diff --git a/Services/WebContent/PageClient.h b/Services/WebContent/PageClient.h
index fe20f0c4eee..b396bcccdd6 100644
--- a/Services/WebContent/PageClient.h
+++ b/Services/WebContent/PageClient.h
@@ -104,6 +104,8 @@ public:
 
     void queue_screenshot_task(Optional<Web::UniqueNodeID> node_id);
 
+    virtual bool has_ladybird_internal_web_ui() const override { return m_web_ui; }
+
 private:
     PageClient(PageHost&, u64 id);
 

```

trflynn89:
Yep that's what I meant - looks like we end up setting up the `WebUIConnection` too late for that to work :(

shannonbooth:
I guess I can just leave a comment in WebUI to update this list if a new one is added for now. I still do think that using a ladybird scheme for our stuff could be nice though

PR #7497, opened 2026-01-17
LibJS: Add source locations to console.trace()

adamjcolvin:
## Summary

This change enhances the `console.trace()` implementation to include source location information (filename, line number, and column number) for each frame in the stack trace. Previously, `console.trace()` only displayed function names, making it difficult to locate the exact source of a trace call. With this change, developers can see the exact source location in DevTools, matching the behavior of other browsers.

The implementation follows the [Console Standard](https://console.spec.whatwg.org/#trace) which specifies that trace should provide "some implementation-defined, potentially-interactive representation of the callstack."

## Implementation Approach

The implementation required changes across several layers of the codebase:

**1. Enhanced TraceFrame structure in LibJS**

The `Console::Trace` structure previously stored only function names as strings. A new `TraceFrame` struct was introduced to hold complete location information:

```cpp
struct TraceFrame {
    String function_name;
    Optional<String> source_file;
    Optional<size_t> line;
    Optional<size_t> column;
};
```

**2. Stack trace extraction refactoring**

The `Console::trace()` function was updated to use `vm.stack_trace()` instead of directly iterating the execution context stack. This provides access to the `CachedSourceRange` objects which contain the source location information. A helper method `realize_source_range()` was added to `CachedSourceRange` to centralize the logic for extracting source range data, which was previously duplicated in `Error.cpp`.

**3. IPC layer updates**

A new `ConsoleTrace` struct was added to the WebView IPC layer, alongside the existing `ConsoleLog` and `ConsoleError` types. The `ConsoleOutput::output` variant was extended to include this new type, along with the necessary IPC encode/decode implementations.

**4. DevTools integration**

The `DevToolsConsoleClient` now handles `LogLevel::Trace` separately, constructing the appropriate `ConsoleTrace` output. The `FrameActor` was updated to format trace messages for the Firefox DevTools Protocol, including the full stack trace with source locations.

## Testing

This change can be tested by opening the browser DevTools console and running `console.trace()` from any JavaScript context. The trace output should now display clickable source locations alongside function names. See the screenshots attached. 

## Changes

- `Libraries/LibJS/Console.h` - Added `TraceFrame` struct, updated `Trace::stack` to use `Vector<TraceFrame>`
- `Libraries/LibJS/Console.cpp` - Updated `trace()` to extract source location information using `vm.stack_trace()`
- `Libraries/LibJS/Runtime/ExecutionContext.h` - Added `realize_source_range()` helper method to `CachedSourceRange`
- `Libraries/LibJS/Runtime/Error.cpp` - Refactored to use the new `realize_source_range()` helper
- `Libraries/LibWebView/ConsoleOutput.h` - Added `ConsoleTrace` struct and IPC declarations
- `Libraries/LibWebView/ConsoleOutput.cpp` - Added IPC encode/decode implementations for `ConsoleTrace`
- `Services/WebContent/DevToolsConsoleClient.cpp` - Added `LogLevel::Trace` handling for DevTools output
- `Libraries/LibDevTools/Actors/FrameActor.cpp` - Added `ConsoleTrace` visitor for DevTools Protocol
- `Libraries/LibWeb/HTML/WorkerDebugConsoleClient.cpp` - Updated to use `TraceFrame::function_name`
- `Utilities/js.cpp` - Updated console client to use `TraceFrame::function_name`

### Before
Screenshot 2026-01-17 at 13 08 58

### After 
Screenshot 2026-01-17 at 10 16 59

### DIFF
--- a/Libraries/LibDevTools/Actors/FrameActor.cpp
+++ b/Libraries/LibDevTools/Actors/FrameActor.cpp
@@ -286,6 +286,39 @@ void FrameActor::on_console_message(WebView::ConsoleOutput console_output)
 
             console_messages.must_append(move(message));
         },
+        [&](WebView::ConsoleTrace const& trace) {
+            message.set("level"sv, "trace"sv);
+            message.set("timeStamp"sv, console_output.timestamp.milliseconds_since_epoch());
+
+            JsonArray arguments;
+            if (!trace.label.is_empty())
+                arguments.must_append(trace.label);
+            message.set("arguments"sv, move(arguments));
+
+            JsonArray stack_array;
+            for (auto const& frame : trace.stack) {
+                JsonObject frame_object;
+                frame_object.set("functionName"sv, frame.function.value_or("<anonymous>"_string));
+                frame_object.set("filename"sv, frame.file.value_or("unknown"_string));
+                frame_object.set("lineNumber"sv, static_cast<i64>(frame.line.value_or(0)));
+                frame_object.set("columnNumber"sv, static_cast<i64>(frame.column.value_or(0)));
+                stack_array.must_append(move(frame_object));
+            }
+            message.set("stacktrace"sv, move(stack_array));
+
+            if (trace.stack.is_empty()) {
+                message.set("filename"sv, "unknown"sv);
+                message.set("lineNumber"sv, 0);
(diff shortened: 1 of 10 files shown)


AtkinsSJ (approved):
Works great, thanks for this!

Issue #7266, opened 2025-12-29
Buttons on Minecraft Wiki panorama viewer do not appear to work


muzikbike:
### Summary

Attempting to use the pause or options menus on panorama viewer embeds on minecraft.wiki appears to do nothing at all. Pages that use them also appear to load strangely compared to others.

### Operating system

Linux

### Steps to reproduce

1. Go to https://minecraft.wiki/w/Panorama#Mounts_of_Mayhem
2. Wait for embeds to load
3. Attempt to click on the pause or options buttons

### Expected behavior

They would work as they do on major browsers.

### Actual behavior

The rendering and animation work fine but the buttons are simply not usable.

### URL for a reduced test case

https://minecraft.wiki/w/Panorama#Mounts_of_Mayhem

### HTML/SVG/etc. source for a reduced test case

```html
N/A but you can find the source code here: https://github.com/mc-wiki/mcw-calc/tree/main/src/tools/panoramaRenderer
```

### Log output and (if possible) backtrace

```shell
[giant ass backtrace from ladybird.txt](https://github.com/user-attachments/files/24369220/giant.ass.backtrace.from.ladybird.txt)
```

### Screenshots or screen recordings

<img width="2050" height="1166" alt="Image" src="https://github.com/user-attachments/assets/d5bbc078-efff-4d19-a919-4e229c62f5c9" />

### Contribute a patch?

- [ ] I’ll contribute a patch for this myself.

La5u:
<img width="788" height="741" alt="Image" src="https://github.com/user-attachments/assets/9c8768cd-f073-463e-8ec4-b798329f2946" />
The buttons are inside an iframe, so that may be the cause

rhdunn:
I get a similar behaviour with https://developer.mozilla.org/en-US/blog/html-details-exclusive-accordions/. The expand/collapse click events don't work in the examples in Ladybird whereas they do in Firefox. The `details`/`summary` expand/collapse works normally, i.e. outside of this context.

These are within an iframe that's contained within a shadow root element that is open, so it looks like the click events are not being propagated to the iframes.

PR #7499, opened 2026-01-17
LibWeb: Account for scroll offset in iframe mouse event coordinates

tcl3:
When delegating mouse events to iframes, coordinate transformations were not accounting for scroll offsets. This caused clicks inside iframes to be incorrectly positioned when the parent page was scrolled.

Fixes: #7266


Libraries/LibWeb/Page/EventHandler.cpp
```diff
@@ -229,6 +229,13 @@ static CSSPixelPoint compute_mouse_event_offset(CSSPixelPoint position, Painting
     return offset;
 }
 
+static CSSPixelPoint compute_position_in_nested_navigable(Painting::Paintable const& paintable, CSSPixelPoint viewport_position)
```
kalenikaliaksandr: lets make this function accept `NavigableContainerViewportPaintable`, so it's explicit that it's not suitable for anything else.

kalenikaliaksandr (changes requested):

kalenikaliaksandr (approved):

PR #7504, opened 2026-01-18
RequestServer+Everywhere: Use sockets to send data for requests

Zaggy1024:
The goal of this PR is mainly to remove a few unnecessary ifdefs in RequestServer and the code it uses to transfer data between itself and its client.

### DIFF
--- a/Libraries/LibCore/System.cpp
+++ b/Libraries/LibCore/System.cpp
@@ -845,6 +845,16 @@ ErrorOr<size_t> transfer_file_through_socket(int source_fd, int target_fd, size_
     if (sent < 0)
         return Error::from_syscall("sendfile"sv, errno);
     return sent;
+#elif defined(AK_OS_MACOS)
+    auto sent_length = static_cast<off_t>(source_length);
+    if (sent_length == 0)
+        return 0;
+    auto result = ::sendfile(source_fd, target_fd, static_cast<off_t>(source_offset), &sent_length, nullptr, 0);
+    if (result != 0) {
+        if ((errno != EAGAIN && errno != EINTR) || sent_length == 0)
+            return Error::from_syscall("sendfile"sv, errno);
+    }
+    return sent_length;
 #else
     static auto page_size = PAGE_SIZE;
 


Libraries/LibHTTP/Cache/CacheEntry.h
```diff
@@ -130,7 +130,7 @@ class CacheEntryReader final : public CacheEntry {
 private:
     CacheEntryReader(DiskCache&, CacheIndex&, u64 cache_key, String url, LexicalPath, NonnullOwnPtr<Core::File>, int fd, CacheHeader, Optional<String> reason_phrase, NonnullRefPtr<HeaderList>, u64 data_offset, u64 data_size);
 
-    void pipe_without_blocking();
+    void send_without_blocking();
     void pipe_complete();
     void pipe_error(Error);
```
konradekk:
```suggestion
    void send_complete();
    void send_error(Error);
```
missed these?

Libraries/LibCore/System.cpp
```diff
@@ -845,6 +845,14 @@ ErrorOr<size_t> transfer_file_through_socket(int source_fd, int target_fd, size_
     if (sent < 0)
         return Error::from_syscall("sendfile"sv, errno);
     return sent;
+#elif defined(AK_OS_MACOS)
+    auto sent_length = static_cast<off_t>(source_length);
+    if (sent_length == 0)
+        return 0;
+    auto result = ::sendfile(source_fd, target_fd, static_cast<off_t>(source_offset), &sent_length, nullptr, 0);
+    if (result != 0 && errno != EAGAIN && errno != EINTR)
```
trflynn89: Why are we handling these error codes here? The caller needs these to know when to arm the socket notifier

trflynn89: Ah, ok so it looks like macOS's `sendfile` might send some bytes even if we get `EAGAIN`. So my concern here is that RequestServer will spin until `sendfile` finally unblocks. The method this is called from is called `pipe_without_blocking` because we explicitly don't want to block here - that will block all of RequestServer on a single cached request.

I think we want to do something more like the following, to allow other requests to be handled while this socket is blocked:
```c++
    auto result = ::sendfile(source_fd, target_fd, static_cast<off_t>(source_offset), &sent_length, nullptr, 0);
    if (result != 0) {
        if ((errno != EAGAIN && errno != EINTR) || sent_length == 0)
            return Error::from_syscall("sendfile"sv, errno);
    }

    return sent_length;
```

That will still result in a second call to `sendfile` that will ultimately block on macOS, but that's better than repeating invocations until it succeeds. Alternatively, we could make the bytes sent an out-param and have `pipe_without_blocking` handle incrementing the bytes sent counter while still handling the returned error (that's probably better, albeit awkward, to avoid an unnecessary syscall that we know will fail).

Zaggy1024: Good point, I've changed it to match your suggestion there.

As far as repeating the syscall in order to get an EAGAIN, that seems like it would always happen on Linux unless we make it so that writing less than `source_length` causes it to return EAGAIN, which theoretically could instead be checked at the call site, but I'm not sure if that would be safe.

PR #7506, opened 2026-01-18
LibGfx+LibWeb: Various font computation improvements

Calme1709:
The main change here is that we now resolve font features per font rather than per element, also includes a number of fixes I came across while working on that - see individual commits for details.


Calme1709:
Marked as draft while I diagnose the Sanitizer crashes

Calme1709:
Was causing a use-after-free by storing a StringView - updated to use a FlyString instead and ready for review now.

Calme1709:
Thinking on this further we only need the `Gfx::TypeFace` for clamping font variations, and we can resolve font features with only the CSS `FontFace`, which we can do in LibWeb and avoid a bunch of boilerplate. Have removed that change from this PR and will revisit in a later one.

PR #7507, opened 2026-01-18
LibWeb: Snag a few low-hanging `referrer-policy/` WPT subtests

CountBleck:
Many of the WPT `referrer-policy/` subtests fail sometimes for reasons unrelated or tangentially to referrers themselves, usually worker-related things. `<meta name="referrer">` also wasn't implemented, and now is!

### DIFF
--- a/Tests/LibWeb/Text/expected/HTML/referrer-policy-meta-tag.txt
+++ b/Tests/LibWeb/Text/expected/HTML/referrer-policy-meta-tag.txt
@@ -1,0 +1,6 @@
+Referrer with referrer policy no-referrer: null
+Referrer with referrer policy empty: null
+Referrer with referrer policy origin: http://127.0.0.1/
+Referrer with referrer policy strict-origin: http://127.0.0.1/
+Referrer with referrer policy unsafe-url: http://127.0.0.1/referrer-policy-meta-tag/iframe
+Referrer with referrer policy always: http://127.0.0.1/referrer-policy-meta-tag/iframe
--- a/Tests/LibWeb/Text/input/HTML/referrer-policy-meta-tag.html
+++ b/Tests/LibWeb/Text/input/HTML/referrer-policy-meta-tag.html
@@ -1,0 +1,61 @@
+<!doctype html>
+<script src="../include.js"></script>
+<script>
+    asyncTest(async done => {
+        const server = httpTestServer();
+
+        const reflectorUrl = await server.createEcho("GET", "/referrer-policy-meta-tag/reflector", {
+            status: 200,
+            headers: {
+                "Content-Type": "application/json",
+            },
+            reflect_headers_in_body: true,
+        });
+
+        const iframeUrl = await server.createEcho("GET", "/referrer-policy-meta-tag/iframe", {
+            status: 200,
+            headers: {
+                "Access-Control-Allow-Origin": "*",
+                "Content-Type": "text/html",
+            },
+            body: `
+                <!DOCTYPE html>
+                <meta name="referrer">
+                <script>
+                    (async () => {
+                        const meta = document.querySelector("meta");
+                        for (const policy of ["no-referrer", "empty", "origin", "strict-origin", "unsafe-url", "always"]) {
+                            meta.content = policy;
+
+                            const response = await fetch("reflector");
+                            const headers = await response.json();
+                            let referrer = null;
+
+                            if ("Referer" in headers) {
+                                const url = new URL(headers["Referer"][0]);
+                                url.port = "";
+                                referrer = url.href;
+                            }
+
+                            parent.postMessage({policy, referrer}, "*");
+                        }
+
+                        parent.postMessage("DONE", "*");
+                    })();
+                <\/script>`,
+        });
+
+        window.onmessage = e => {
+            if (e.data === "DONE") {
+                done();
+                return;
+            }
+
+            const { policy, referrer } = e.data;
+            println(`Referrer with referrer policy ${policy}: ${referrer}`);
+        };
+
+        document.querySelector("iframe").src = iframeUrl;
+    });
+</script>
+<iframe></iframe>


Libraries/LibWeb/HTML/WorkerAgentParent.cpp
```diff
@@ -51,7 +52,16 @@ void WorkerAgentParent::initialize(JS::Realm& realm)
     m_worker_ipc = make_ref_counted<WebWorkerClient>(move(transport));
     setup_worker_ipc_callbacks(realm);
 
-    m_worker_ipc->async_start_worker(m_url, m_worker_options.type, m_worker_options.credentials, m_worker_options.name, move(data_holder), m_outside_settings->serialize(), m_agent_type);
+    auto serialized_outside_settings = m_outside_settings->serialize();
+
+    // HACK: The referrer for the worker script fetch should apparently use the current URL, and not the creation URL
```
shannonbooth: There is an in flight HTML MR relating to this: https://github.com/whatwg/html/pull/12038

CountBleck: Getting an email notification just over 10 minutes of opening the PR scared me, but I'm glad that this behavior is getting spec'd after all. I'll see what changes I need to make to align with those proposed spec changes tomorrow.

shannonbooth: At least one change does definitely align, but hopefully some more context at least!

CountBleck: Actually, the spec PR you mentioned isn't related to this diff, because that PR pertains to "inner settings" (i.e. the worker's ESO), where as this diff involves "outside settings". A different commit implements that spec change, so I updated that.

At a second glance, I have a better explanation for why the hack in this diff is needed...
Edit: and I added that explanation.

Libraries/LibWeb/HTML/Scripting/WorkerEnvironmentSettingsObject.cpp
```diff
@@ -35,6 +35,8 @@ GC::Ref<WorkerEnvironmentSettingsObject> WorkerEnvironmentSettingsObject::setup(
     settings_object->target_browsing_context = nullptr;
 
     // FIXME: 5. Set settings object's id to a new unique opaque string, creation URL to worker global scope's url, top-level creation URL to null, target browsing context to null, and active service worker to null.
+    // AD-HOC: We set the creation URL after fetching the worker script, as the worker global scope doesn't have a url set at this point.
```
shannonbooth: It's probably worth linking to the spec issue here https://github.com/whatwg/html/issues/11340 here since this might be read with the wrong understanding here. Technically I think we _do_ set up worker global scope with a URL - it's just not the _correct_ URL if there is a redirect: https://github.com/LadybirdBrowser/ladybird/blob/cc559048cc302a19ba085772b5f3fc682d37fca0/Services/WebWorker/WorkerHost.cpp#L60

Libraries/LibWeb/HTML/WorkerAgentParent.cpp
```diff
@@ -51,7 +52,17 @@ void WorkerAgentParent::initialize(JS::Realm& realm)
...
 
-    m_worker_ipc->async_start_worker(m_url, m_worker_options.type, m_worker_options.credentials, m_worker_options.name, move(data_holder), m_outside_settings->serialize(), m_agent_type);
+    auto serialized_outside_settings = m_outside_settings->serialize();
+
+    // HACK: The environment settings object used for the worker script fetch has a Window as its global scope, but the
+    //       EnvironmentSettingsSnapshot used in the WebWorker process uses a WorkerGlobalScope. This causes the
+    //       Referrer-Policy spec's "determine request's referrer" algorithm to read the ESO's creation URL, whereas it
+    //       would normally read the document's URL. To hack around this, we overwrite the creation URL (which is
+    //       only used in the initial worker script fetch).
+    if (auto* window = as_if<HTML::Window>(m_outside_settings->realm().global_object()))
```
shannonbooth, line 58: It does feel like outside settings will cause us further issues in the future, I remember digging into this here: https://github.com/whatwg/html/issues/11340#issuecomment-2918550278 where I realized we were doing things wrong. At that time I was considering if we should create a stub window object to use as outside settings global object inside WorkerHost.

Maybe to make this a bit more clear we could pass Window's associated document's URL over IPC, and move this hack across into WorkerHost.cpp where outside settings is being created? That at least localizes things a bit more, hopefully making it more clear as to why we need to do it, since WorkerHost.cpp does not have access to the original Window object as it is inside the worker process when setting up outside settings.

Libraries/LibWeb/HTML/WorkerAgentParent.cpp
shannonbooth: Sorry I indeed commented on the wrong spot

shannonbooth (changes requested):
Thanks! Would it be possible to add any in tree tests for these changes?

Libraries/LibWeb/HTML/WorkerAgentParent.cpp
CountBleck: I don't think the stub window object would be that reasonable, since the Referrer-Policy AOs look at the associated document, which would be its own can-of-worms to stub out.

I'll go do the IPC thing.

Libraries/LibWeb/HTML/Scripting/WorkerEnvironmentSettingsObject.cpp
CountBleck: My bad. I authored that patch before you authored 6107775, so I didn't realize that now the URL is set.

Libraries/LibWeb/HTML/WorkerAgentParent.cpp
CountBleck: Done!

shannonbooth:
Unfortunately it looks like the tests are timing out in CI for worker referrer - maybe the problem here is more generally with Workers though as opposed to referrer?

CountBleck:
@shannonbooth I believe I fixed it now. (The echo server doesn't get reset between tests, and it doesn't handle collisions well...)

shannonbooth (approved):

PR #7508, opened 2026-01-18
LibWeb+LibMedia: Decoding for webaudio

jonbgamble:
Support webaudio's BaseAudioContext.decodeAudioData via ffmpeg. Use a background thread until that glorious day arrives when all media is decoded outside of WebContent.


Libraries/LibWeb/WebAudio/BaseAudioContext.cpp
```diff
@@ -278,74 +304,159 @@ GC::Ref<WebIDL::Promise> BaseAudioContext::decode_audio_data(GC::Root<WebIDL::Bu
     return promise;
 }
 
-// https://webaudio.github.io/web-audio-api/#dom-baseaudiocontext-decodeaudiodata
-void BaseAudioContext::queue_a_decoding_operation(GC::Ref<JS::PromiseCapability> promise, [[maybe_unused]] GC::Root<WebIDL::BufferSource> audio_data, GC::Ptr<WebIDL::CallbackType> success_callback, GC::Ptr<WebIDL::CallbackType> error_callback)
+// https://webaudio.github.io/web-audio-api/#dom-baseaudiocontext-decoding-thread
+void BaseAudioContext::queue_a_decoding_operation(GC::Ref<JS::PromiseCapability> promise, [[maybe_unused]] GC::Root<WebIDL::BufferSource> const& audio_data, GC::Ptr<WebIDL::CallbackType> success_callback, GC::Ptr<WebIDL::CallbackType> error_callback)
 {
-    auto& realm = this->realm();
+    auto context = GC::make_root(*this);
```
shannonbooth: we should not need to use a `GC::Root` here, `this` should be protected by GC by being captured in the GC::Function, and that GC::Function is able to visit GC::Cell types and keep BaseAudioContext from being garbage collected (making this a root is just likely to cause a memory leak).

Though, maybe something is needed in particular for Threading::BackgroundAction callbacks since those ones don't understand the GC

jonbgamble: I would assume at least the PromiseCapability and js callbacks need rooting for the background thread.

What I'm worried about now - when cancelled, it seems like BackgroundAction can call m_on_error on the worker thread rather than the requesting origin event loop. I'm not sure anything GC is safe there.

shannonbooth: > I would assume at least the PromiseCapability and js callbacks need rooting for the background thread.

Right, I think so too.

> What I'm worried about now - when cancelled, it seems like BackgroundAction can call m_on_error on the worker thread rather than the requesting origin event loop. I'm not sure anything GC is safe there.

Yeah, the GC is not thread safe by design, it assumes it is only used from a single thread. GC::Root is [not atomic ref counted](https://github.com/LadybirdBrowser/ladybird/blob/f1cf79cac416ac34d528ead4ae6b1a6a7bb20517/Libraries/LibGC/Root.h#L20), and the destructor/constructor is also [touching the heap without locks](https://github.com/LadybirdBrowser/ladybird/blob/f1cf79cac416ac34d528ead4ae6b1a6a7bb20517/Libraries/LibGC/Root.cpp#L17). I wonder if we need a thread safe version of GC::Root though? Might just be better to only ref/unref the GC::Root in the main thread (somehow)

Coincidentally I had actually raised https://github.com/LadybirdBrowser/ladybird/pull/7489 not long before this thinking that we would be adding threading into this code in the near future to add assertions for such cases which I think would be nice to add into those callbacks. I think I should actually add an assertion to RootImpl as well now that I thought about that case :sweat_smile:

jonbgamble: I made some tweaks, not sure about my BackgroundAction change tbh. They are a separate commit (for now) so you can see what changed easier.

Zaggy1024: > Might just be better to only ref/unref the GC::Root in the main thread (somehow)

I would agree here, just move() the GC::Root back to a callback on the main thread via deferred_invoke(). (Note: I haven't looked closely at the threading setup here, but that's what I'd do)

ADKaster: It is completely incorrect to interact with a GC object, or a web platform object, from any thread other than the main thread in any way, shape or form. The html spec is pretty clear about that, and any violations in spec-land are documented and being actively removed. Ladybird breaks that rule in a lot more places than we should. Even if the GC is made thread-safe, platform objects absolutely must not be modified outside of the main event loop.

ADKaster: The correct thing to do to ferry state back to the main thread is to queue a task, where the task body updates any platform objects (and disposes of any illicit cross-thread gc objects).

jonbgamble: That's what happens. BackgroundAction.h should no longer call m_on_error outside of the origin event loop on cancels.

We were seeking the best way to keep the promise and callbacks alive until the decoder thread completes.

Libraries/LibWeb/WebAudio/BaseAudioContext.cpp
```diff
@@ -239,113 +254,278 @@ GC::Ref<WebIDL::Promise> BaseAudioContext::decode_audio_data(GC::Root<WebIDL::Bu
...
+        take_pending_promise(promise);
 
         // 4.3. Queue a media element task to invoke errorCallback with error.
         if (error_callback) {
-            queue_a_media_element_task(GC::create_function(heap(), [&realm, error_callback, error] {
-                auto completion = WebIDL::invoke_callback(*error_callback, {}, { { error } });
-                if (completion.is_abrupt())
-                    HTML::report_exception(completion, realm);
-            }));
+            auto rooted_promise = GC::make_root(promise);
```
shannonbooth: We should not be rooting these objects, the GC protects them when passed into the GC::Function as captures

Libraries/LibWeb/WebAudio/BaseAudioContext.cpp
```diff
@@ -239,113 +254,278 @@ GC::Ref<WebIDL::Promise> BaseAudioContext::decode_audio_data(GC::Root<WebIDL::Bu
...
+            auto rooted_error_callback = GC::make_root(error_callback);
+
+            HTML::queue_global_task(
+                m_media_element_event_task_source.source,
+                rooted_promise->promise()->shape().realm().global_object(),
```
shannonbooth: This should probably just use the glboal_object that we got above. And the realm of BaseAudioContext should instead be passed by reference as a capture into the GC::Function instead of pulling it out of the promise like this.


I _think_ they are equivalent here, but the above is more common of a pattern in the codebase.

Libraries/LibWeb/WebAudio/BaseAudioContext.cpp
```diff
@@ -239,113 +254,278 @@ GC::Ref<WebIDL::Promise> BaseAudioContext::decode_audio_data(GC::Root<WebIDL::Bu
...
         if (error_callback) {
-            queue_a_media_element_task(GC::create_function(heap(), [&realm, error_callback, error] {
-                auto completion = WebIDL::invoke_callback(*error_callback, {}, { { error } });
-                if (completion.is_abrupt())
-                    HTML::report_exception(completion, realm);
-            }));
+            auto rooted_promise = GC::make_root(promise);
+            auto rooted_error_callback = GC::make_root(error_callback);
+
+            HTML::queue_global_task(
```
shannonbooth: Is there a particular reason we switched from queue_a_media_element_task to HTML::queue_global_task - queue_a_media_element_task better matches spec text?

Libraries/LibWeb/WebAudio/BaseAudioContext.cpp
```diff
@@ -239,113 +254,278 @@ GC::Ref<WebIDL::Promise> BaseAudioContext::decode_audio_data(GC::Root<WebIDL::Bu
...
+    }
 
-        // FIXME: 5.2. queue a media element task to execute the following steps:
+    void reject_with_callback(JS::Realm& realm, WebIDL::Promise const& promise, GC::Ref<WebIDL::DOMException> error) const
+    {
+        WebIDL::reject_promise(realm, promise, error);
+        if (auto* callback = m_error_callback.ptr()) {
+            auto completion = WebIDL::invoke_callback(*callback, {}, { { error } });
+            if (completion.is_abrupt())
+                HTML::report_exception(completion, realm);
```
shannonbooth: Interesting - this seems to be preexisting, but it seems wrong that we call report_exception here.. 🤔

Libraries/LibWeb/WebAudio/BaseAudioContext.cpp
jonbgamble: I'm going to have to start putting money in the GC::Root jar, aren't I?

jonbgamble:
The issues reviewers have found convinced me to reorganize this a bit (again). Hopefully a new translation unit can separate the GC/main-thread world from the background thread a bit better, without complicating BaseAudioContext (which will need plenty of room to grow).

Libraries/LibWeb/WebAudio/BaseAudioContext.cpp
jonbgamble: I'm going to figure somthing else out. I'm not crazy about pinning the context if a promise or callbacks don't close over it.

jonbgamble:
I refactored towards an audio decoding "manager" with document scope & lifetime to facilitate tracking objects that might need to outlive a context.

It's probably bad that I dumped all the spec steps verbatim above the method, but it could be handy for review.

Libraries/LibMedia/Audio/AudioDecoding.cpp
```diff
@@ -0,0 +1,142 @@
...
+    auto stream = IncrementallyPopulatedStream::create_from_buffer(move(encoded_data));
+    TRY(check_cancelled());
+    auto demuxer = TRY(create_demuxer_for_stream(stream));
+    TRY(check_cancelled());
+
+    auto tracks = TRY(demuxer->get_tracks_for_type(TrackType::Audio));
+    if (tracks.is_empty())
+        return DecoderError::with_description(DecoderErrorCategory::NotImplemented, "No audio tracks found"sv);
+
+    // Webaudio's decodeAudioData just wants the first audio track.
```
Zaggy1024: nit: This comment seems redundant with the function name.

Libraries/LibMedia/Audio/AudioDecoding.cpp
```diff
@@ -0,0 +1,142 @@
...
+        if (result.is_error())
+            return DecoderError::with_description(DecoderErrorCategory::NotImplemented, result.error().string_literal());
+    }
+
+    auto stream_cursor = stream->create_cursor();
+    TRY(demuxer->create_context_for_track(track, stream_cursor));
+
+    DecodedAudioData decoded;
+    decoded.sample_specification = output_sample_specification;
+    // AD-HOC: We'll do interleaved.
```
Zaggy1024: nit: All our audio data is (currently) interleaved, so perhaps this comment could clarify why this matters? Also, there's no spec here, so ad-hoc doesn't apply.

Libraries/LibMedia/Audio/AudioDecoding.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+    auto output_sample_specification = input_sample_specification;
+    if (target_sample_rate.has_value() && *target_sample_rate > 0 && input_sample_specification.sample_rate() != *target_sample_rate)
+        output_sample_specification = Audio::SampleSpecification(*target_sample_rate, input_sample_specification.channel_map());
+
+    if (output_sample_specification.is_valid()) {
+        auto result = converter->set_output_sample_specification(output_sample_specification);
+        if (result.is_error())
+            return DecoderError::with_description(DecoderErrorCategory::NotImplemented, result.error().string_literal());
+    }
```
Zaggy1024: If the output sample specification is invalid here, we could probably just exit with an error. That way, we can assume from here that the output sample specification is valid.

Libraries/LibMedia/Audio/AudioDecoding.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+    auto tracks = TRY(demuxer->get_tracks_for_type(TrackType::Audio));
+    if (tracks.is_empty())
+        return DecoderError::with_description(DecoderErrorCategory::NotImplemented, "No audio tracks found"sv);
+
+    // Webaudio's decodeAudioData just wants the first audio track.
+    auto const& track = tracks.first();
+
+    auto codec_id = TRY(demuxer->get_codec_id_for_track(track));
+    auto const& input_sample_specification = track.audio_data().sample_specification;
```
Zaggy1024, line 49: FFmpeg sniffs the first audio packet from the container which allows us to get a sample specification from all the files we use it for, but for Matroska (or any other custom demuxers we may need in future) this may not be a valid sample specification.

It's probably worth lazily initializing the converter on the first AudioBlock that comes back instead, if the output sample specification should really be using the input channel map.

Libraries/LibMedia/Audio/AudioDecoding.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+    TRY(check_cancelled());
+    decoder->signal_end_of_stream();
+    TRY(drain_decoder());
+
+    TRY(check_cancelled());
+
+    if (!decoded.sample_specification.is_valid()) {
+        // I guess fall back to the input spec here.
+        decoded.sample_specification = input_sample_specification;
```
Zaggy1024: If the output sample specification was invalid, the converter probably would have crashed by this point, since the default output sample specification is default-initialized (invalid). We should probably not even get here if the output sample specification is invalid.

Libraries/LibMedia/Audio/AudioDecoding.h
```diff
@@ -0,0 +1,26 @@
...
+#include <AK/Function.h>
+#include <LibMedia/Audio/SampleSpecification.h>
+#include <LibMedia/DecoderError.h>
+#include <LibMedia/Export.h>
+
+namespace Media {
+
+struct MEDIA_API DecodedAudioData {
+    Audio::SampleSpecification sample_specification;
+    u32 frame_count { 0 };
```
Zaggy1024: Is there a reason to store this if we can infer the frame count from `interleaved_f32_samples.size()` and `sample_specification.channel_count()`?

Libraries/LibWeb/WebAudio/BackgroundAudioDecoder.cpp
```diff
@@ -0,0 +1,286 @@
...
+//   2. Let promise be a new Promise.
+//   3. If audioData is not detached, execute the following steps:
+//      3.1. Append promise to [[pending promises]].
+//      3.2. Detach the audioData ArrayBuffer. If this operation throws, jump to step 4.1.
+//      3.3. Queue a decoding operation to be performed on another thread.
+//   4. Else (audioData is detached), execute the following error steps:
+//      4.1. Let error be a DataCloneError.
+//      4.2. Reject promise with error, and remove it from [[pending promises]].
+//      4.3. Queue a media element task to invoke errorCallback with error.
+//   5. Return promise.
```
Zaggy1024: We could probably omit the steps here, and just leave the header indicating that it runs on the control thread.

Libraries/LibWeb/WebAudio/BackgroundAudioDecoder.cpp
```diff
@@ -0,0 +1,286 @@
...
+//      4.2. Reject promise with error, and remove it from [[pending promises]].
+//      4.3. If errorCallback is not missing, invoke errorCallback with error.
+//   5. Otherwise:
+//      5.1. Take the result, representing the decoded linear PCM audio data, and resample it to
+//           the sample-rate of the BaseAudioContext if it is different from the sample-rate of
+//           audioData.
+//      5.2. Queue a media element task to execute the following steps:
+//         5.2.1. Let buffer be an AudioBuffer containing the final result.
+//         5.2.2. Resolve promise with buffer.
+//         5.2.3. If successCallback is not missing, invoke successCallback with buffer.
```
Zaggy1024: Same here, probably want to omit the steps but keep the "decoding thread" part.

Libraries/LibWeb/WebAudio/BackgroundAudioDecoder.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+    if (!request_index.has_value())
+        return;
+
+    PendingRequest request = m_pending_requests[*request_index];
+    m_pending_requests.remove(*request_index);
+
+    // AD-HOC: Ensure the promise doesn't stay pending forever.
+    if (auto context = request.context.ptr(); context)
+        (void)context->take_pending_promise(request.promise);
```
Zaggy1024, line 187: Is this a spec bug? The spec must remove from this collection at some point if it isn't, right? If so, it would be good to bring this up on their issue tracker.

Libraries/LibWeb/WebAudio/BackgroundAudioDecoder.cpp
```diff
@@ -0,0 +1,286 @@
...
+        if (channel_count == 0 || channel_count > BaseAudioContext::MAX_NUMBER_OF_CHANNELS) {
+            reject_with_exception(WebIDL::NotSupportedError::create(realm, "Unsupported channel count"_utf16));
+            return;
+        }
+        if (frame_count == 0 || frame_count > NumericLimits<WebIDL::UnsignedLong>::max()) {
+            reject_with_exception(WebIDL::NotSupportedError::create(realm, "Bad audio data size"_utf16));
+            return;
+        }
+
+        auto buffer_or_exception = AudioBuffer::create(realm, channel_count, frame_count, request.sample_rate);
```
Zaggy1024, line 209: The error conditions here seem redundant with the ones in `verify_audio_options_inside_nominal_range()`, can we omit them here?

Libraries/LibMedia/Audio/AudioDecoding.cpp
jonbgamble: i use planar throughout the webaudio pipeline, but there's no reason for this comment here.

Zaggy1024: Right, sorry, I meant in LibMedia it's all interleaved. We could potentially support planar (in one buffer) as well, but it seems pointless as all the audio APIs we're using thus far support or prefer interleaved.

Libraries/LibWeb/WebAudio/BackgroundAudioDecoder.cpp
jonbgamble: The spec does not explicitly say to remove from `[[pending promises]]` in the success path, only that we should resolve with the buffer. I am not familiar with the desired level of rigor but I personally wouldn't make an issue over that. The reason I put AD-HOC comment is because the spec step ordering isn't ideal. We remove it from `[[pending promises]]` structure well before any resolve/reject so that subsequent fail fasts & cancellations don't have to worry about it.

jonbgamble:
I still have some conflicts to fix. Hopefully this comment is enough to remove the stale label until then.

PR #7509, opened 2026-01-18
Three random little cleanups

awesomekling:
See individual commits; just some stuff I had sitting around that didn't belong to any branch.

### DIFF
--- a/Libraries/LibWeb/CSS/ComputedValues.h
+++ b/Libraries/LibWeb/CSS/ComputedValues.h
@@ -697,7 +697,7 @@ public:
     ScrollbarColorData scrollbar_color() const { return m_inherited.scrollbar_color; }
     ScrollbarWidth scrollbar_width() const { return m_noninherited.scrollbar_width; }
     Resize resize() const { return m_noninherited.resize; }
-    WillChange will_change() const { return m_noninherited.will_change; }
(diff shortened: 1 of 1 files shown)


Libraries/LibJS/Runtime/PropertyAttributes.h
```diff
@@ -27,49 +27,49 @@ struct Attribute {
...
 
-    [[nodiscard]] u8 bits() const { return m_bits; }
+    [[nodiscard]] constexpr u8 bits() const { return m_bits; }
 
 private:
     u8 m_bits;
 };
 
-PropertyAttributes const default_attributes = Attribute::Configurable | Attribute::Writable | Attribute::Enumerable;
+static constexpr PropertyAttributes default_attributes = Attribute::Configurable | Attribute::Writable | Attribute::Enumerable;
```
trflynn89, line 72: `static` doesn't actually do anything here, you want `constexpr inline` instead

awesomekling: It did something, as it’s no longer in BSS but yes let’s make that change :)

trflynn89: IIRC for variables, `constexpr` does that on its own, `static` ends up not adding anything

PR #7510, opened 2026-01-18
RequestServer: Add --resource-map option for URL-to-file substitution

awesomekling:
This adds support for intercepting network requests and serving local file content instead. When a URL matches an entry in the substitution map, the local file is served while preserving the original URL's origin for cross-origin checks.

Usage:
```
    Ladybird --resource-map=/path/to/map.json
```

The JSON file format is:
```json
    {
      "substitutions": [
        {
          "url": "https://example.com/script.js",
          "file": "/path/to/local/script.js",
          "content_type": "application/javascript",
          "status_code": 200
        }
      ]
    }
```

Fields:
  - `url` (required): Exact URL to intercept (query string and fragment are stripped before matching)
  - `file` (required): Absolute path to local file to serve
  - `content_type` (optional): Override Content-Type header (defaults to guessing from filename)
  - `status_code` (optional): HTTP status code (defaults to 200)

This is incredibly useful for debugging production websites: you can intercept any script, stylesheet, or other resource and replace it with a local copy containing your own debug instrumentation, console.log statements, or experimental fixes - all without modifying the actual site or setting up a local dev server.

### DIFF
--- a/Services/RequestServer/Request.cpp
+++ b/Services/RequestServer/Request.cpp
@@ -6,16 +6,22 @@
  */
 
 #include <AK/GenericShorthands.h>
+#include <LibCore/File.h>
+#include <LibCore/MimeData.h>
 #include <LibCore/Notifier.h>
 #include <LibHTTP/Cache/DiskCache.h>
 #include <LibHTTP/Cache/Utilities.h>
+#include <LibHTTP/Status.h>
 #include <LibTextCodec/Decoder.h>
 #include <RequestServer/CURL.h>
 #include <RequestServer/ConnectionFromClient.h>
 #include <RequestServer/Request.h>
 #include <RequestServer/Resolver.h>
+#include <RequestServer/ResourceSubstitutionMap.h>
 
 namespace RequestServer {
+
+extern OwnPtr<ResourceSubstitutionMap> g_resource_substitution_map;
 
 static long s_connect_timeout_seconds = 90L;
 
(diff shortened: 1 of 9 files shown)


Services/RequestServer/ResourceSubstitutionMap.cpp
```diff
@@ -0,0 +1,80 @@
...
+        ResourceSubstitution substitution;
+        substitution.file_path = file_value->as_string().to_byte_string();
+
+        if (auto content_type_value = obj.get("content_type"sv); content_type_value.has_value() && content_type_value->is_string())
+            substitution.content_type = content_type_value->as_string().to_byte_string();
+
+        if (auto status_code_value = obj.get("status_code"sv); status_code_value.has_value() && status_code_value->is_integer<u32>())
+            substitution.status_code = status_code_value->as_integer<u32>();
+
+        map->m_substitutions.set(url_value->as_string().to_byte_string(), move(substitution));
```
tcl3: Would it be worth parsing and normalizing the URL before storing it in the map? I'm thinking of cases like trailing forward slashes etc. 

Should we strip out the query string and fragment from URLs before storing and comparing them?

Services/RequestServer/Request.cpp
```diff
@@ -294,6 +310,67 @@ void Request::handle_read_cache_state()
...
+
+    m_curl_result_code = CURLE_OK;
+
+    if (write_queued_bytes_without_blocking().is_error()) {
+        transition_to_state(State::Error);
+        return;
+    }
+
+    if (m_response_buffer.is_eof())
+        transition_to_state(State::Complete);
```
tcl3: I'm not an expert when it comes to RequestServer stuff, but couldn't this lead to a double transition to complete, since  `write_queued_bytes_without_blocking()` calls `transition_to_state(State::Complete)` internally?

PR #7514, opened 2026-01-18
LibWeb: Use progressive fallback for password input masking character

tcl3:
If BLACK CIRCLE (●) is not available, then fall back to BULLET (•), otherwise use the universally available asterisk.

This prevents the password being rendered as the missing glyph character if the preferred character is not available.

### DIFF
--- a/Libraries/LibGfx/FontCascadeList.cpp
+++ b/Libraries/LibGfx/FontCascadeList.cpp
@@ -75,4 +75,25 @@ bool FontCascadeList::equals(FontCascadeList const& other) const
     return true;
 }
 
+u32 FontCascadeList::password_mask_character() const
+{
+    return m_cached_password_mask_character.ensure([&] {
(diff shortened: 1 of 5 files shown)


Libraries/LibGfx/FontCascadeList.cpp
```diff
@@ -67,4 +67,30 @@ bool FontCascadeList::equals(FontCascadeList const& other) const
     return true;
 }
 
+u32 FontCascadeList::password_mask_character() const
+{
+    if (m_cached_password_mask_character.has_value())
+        return m_cached_password_mask_character.value();
```
gmta: This can be `Optional<T>::ensure(Callback)`

tcl3:
I think this code path is much less likely to be hit since #7536 was landed, but it is still useful to guarantee we provide a reasonable character.

PR #7517, opened 2026-01-18
LibWeb/WebAudio: Add stable IDs to AudioNodes

Prospero23:
Modeled after `UniqueNodeID` and uses incremental counter on `BaseAudioContext` to assign IDs.

Included `NodeID` type in new `WebAudio/Types.h` file to avoid circular dependencies. I see that `UniqueNodeID` is declared in `LibWeb/Forward.h`, so happy to move the type there if it is cleaner to do so.

### DIFF
--- a/Libraries/LibWeb/WebAudio/BaseAudioContext.h
+++ b/Libraries/LibWeb/WebAudio/BaseAudioContext.h
@@ -22,6 +22,7 @@
 #include <LibWeb/WebAudio/PeriodicWave.h>
 #include <LibWeb/WebAudio/ScriptProcessorNode.h>
 #include <LibWeb/WebAudio/StereoPannerNode.h>
+#include <LibWeb/WebAudio/Types.h>
 #include <LibWeb/WebIDL/Types.h>
 
 namespace Web::WebAudio {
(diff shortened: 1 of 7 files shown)


Libraries/LibWeb/WebAudio/BaseAudioContext.h
```diff
@@ -94,18 +95,24 @@ class BaseAudioContext : public DOM::EventTarget {
...
+    NodeID next_node_id() { return ++m_next_node_id; }
+
     virtual void initialize(JS::Realm&) override;
     virtual void visit_edges(Cell::Visitor&) override;
 
     GC::Ptr<AudioDestinationNode> m_destination;
     Vector<GC::Ref<WebIDL::Promise>> m_pending_promises;
 
 private:
+    friend class AudioNode;
```
gmta: Nit; you don't have to add a friend class just for this one new method. Make this public:

```cpp
NodeID next_node_id(Badge<AudioNode>) { return ++m_next_node_id; }
```

And in AudioNode's constructor, do this:

```cpp
    , m_node_id(context->next_node_id({}))
```

Prospero23: Knew there was some easier way! Thanks, fixed.

PR #7518, opened 2026-01-18
LibWeb/Bindings: Implement [SecureContext] IDL attribute, generate some more Serializable code

### LINKED ISSUE #1281: LibWeb: Support [SecureContext] extended IDL attribute
Many WebIDL interfaces define members that must only be accessed from a secure context.

We do not currently support this attribute, and it is ignored.

The IDL generator should be updated to add checks to ensure that these APIs are unavailable in a non-secure context. This likely involves an audit of all [SecureContext] web apis to make sure that we only define_native_accessor (or other define_native_xyz as applicable) for them when the relevant settings object is a secure context.

shannonbooth:
This is kind of a pain to test at the moment as it does not seem easy to actually create a non-secure context with our current test setup. This is because:

- Files schemes currently have a tuple origin, so are considered potentially secure (explicitly for file scheme)
- Running the HTTP server is over localhost, which is considered potentially secure
- Data URLs are also considered potentially secure

Fixes #1281

### DIFF
--- a/Libraries/LibWeb/HTML/WindowOrWorkerGlobalScope.cpp
+++ b/Libraries/LibWeb/HTML/WindowOrWorkerGlobalScope.cpp
@@ -2,7 +2,7 @@
  * Copyright (c) 2022, Andrew Kaster <akaster@serenityos.org>
  * Copyright (c) 2023, Linus Groh <linusg@serenityos.org>
  * Copyright (c) 2023, Luke Wilde <lukew@serenityos.org>
- * Copyright (c) 2025, Shannon Booth <shannon@serenityos.org>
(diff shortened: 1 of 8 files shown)


shannonbooth:
Actually, I take it back - this should be easy to test by using a data URL, other browsers (sensibly) consider this a non secure context - seems the spec is not correct here

shannonbooth:
> Actually, I take it back - this should be easy to test by using a data URL, other browsers (sensibly) consider this a non secure context - seems the spec is not correct here

After testing further against other browser engines, I am more confused than before. Raised spec issue of https://github.com/whatwg/html/issues/12091 with findings.

shannonbooth:
I gave up on trying to find a nice way of testing this and just ended up adding an internals setter so that the IDL generator code is at least getting some coverage.

PR #7520, opened 2026-01-19
LibJS: Add bytecode IR regression tests

dosisod:
This adds a new `test-js-bytecode` target which ensures that codegen changes do not impact emitted bytecode IR, or if it does, it is known and the tests are updated accordingly.

Similar to the LibWeb tests, the tests are stored in the following format:

* `Libraries/LibJS/Bytecode/Tests/input`: Input `.js` files
* `Libraries/LibJS/Bytecode/Tests/expected`: Expected `.txt` bytecode
* `Libraries/LibJS/Bytecode/Tests/output`: Emitted `.txt` bytecode

The `output` dir is git-ignored, but stores the output so you can diff and inspect failed tests more easily.

There is only one test so far, which is a baseline test that should not change dramatically unless we change the bytecode output format.

---

### Example Usage

Here is the `input/baseline.js` test included in this PR:

```js
function add(a, b) {
    return a + b;
}

console.log(add(1, 2));
```

When running `test-js-bytecode`, you should get the following output:

```
$ BUILD_PRESET=Debug ./Meta/ladybird.py run test-js-bytecode
ninja: Entering directory `/home/logan/git/ladybird/ladybird/Build/debug'
ninja: no work to do.
All tests passed!
```

If you modify the `input/baseline.js` file, or the `expected/baseline.txt` file, that test will fail:

image

In the `baseline.js` I changed `+` to `-`, which changed the resulting `Add` op code to `Sub`. Here I'm modifying the `.js` file directly, but you could imagine this being a bug/regression in the compiler which changes the emitted bytecode.

### Future Improvements

* Add `--rebaseline` similar to the LibWeb tests which update the `.txt` files with the output
* Allow for only running selected tests. Currently all tests run
* Allow for dumping bytecode without running script.
* ~~Run tests in parallel~~ Done. Can be configured via `--jobs`


awesomekling:
Fantastic idea! But perhaps we should call it `test-js-bytecode` or something like that since it specifically tests JS bytecode and not any of the other bytecode formats we have :)

awesomekling:
Or oh, wait, it's in LibJS/test-bytecode.py -- that's already unambiguous. Never mind :)

Tests/LibJS/test-bytecode.py
```diff
@@ -0,0 +1,105 @@
...
+    if failed:
+        print("One or more tests failed.")
+        sys.exit(1)
+
+    else:
+        print("All tests passed!")
+
+
+if __name__ == "__main__":
+    main()
```
konradekk: some minor suggestions:

```suggestion
def main() -> int:
    setup()

    input_dir = BYTECODE_TEST_DIR / "input"

	failed = 0

	for js_file in sorted(input_dir.iterdir()):
	    if js_file.is_file() and js_file.suffix == ".js":
		    # TODO: run tasks in parallel via thread/executor pool
	        if not test(js_file.relative_to(input_dir)):
	            failed += 1

	if failed:
	    print(f"Tests failed: {failed}")
        return 1

    else:
        print("All tests passed!")
        return 0


if __name__ == "__main__":
    sys.exit(main())
```

here’s executor snippet that should work (if interested):

```python
from concurrent.futures import ThreadPoolExecutor as Executor, as_completed

...

js_files = [
    js_file for js_file in sorted(input_dir.iterdir()) if js_file.is_file() and js_file.suffix == ".js"
]

with Executor() as executor:
    executables = {
        executor.submit(test, js_file.relative_to(input_dir)): js_file for js_file in js_files
    }

    for executable in as_completed(executables):
        if not executable.result():
            failed += 1
```

optionally switch `ThreadPoolExecutor` to `ProcessPoolExecutor` (and expose `max_workers`?).

konradekk (commented):
(possibly some quick wins; ignore at will)

dosisod:
I think it would be best to rename the target to `test-js-bytecode` to remove any confusion with the WASM/RegEx bytecode.

Tests/LibJS/test-bytecode.py
dosisod: Done. Implemented with some small changes. I also stuck with the thread pool executor because the process poll was having issues with the env vars not being passed. If we end up having slow downs later on we can revisit this.

PR #7521, opened 2026-01-19
LibWeb: Prevent text wrapping inside select elements

### LINKED ISSUE #7073: Text inside select elements can line break
This text wraps onto multiple lines in Ladybird, but does not in Firefox/Chrome/Safari:
```html
<select style="width: 10em;">
    <option>This is some really long text that should not wrap</option>
</select>
```

Firefox achieves this with a `white-space: nowrap !important;` in its UA styles for select, but that doesn't seem to be in the spec. (The spec only expects `nowrap` to be set on the option elements themselves)

sheeki03:
## Summary

Fixes #7073

Add `white-space: nowrap !important` to the select element user-agent styles to prevent option text from wrapping onto multiple lines. This matches the behavior of Firefox and other major browsers.

## Test plan

- Added layout test: `Tests/LibWeb/Layout/input/select-text-nowrap.html`
- Test verifies text stays on single line even with constrained width
- Ran `test-web` - test passes

### DIFF
--- a/Libraries/LibWeb/CSS/Default.css
+++ b/Libraries/LibWeb/CSS/Default.css
@@ -68,10 +68,6 @@ input[type=image] {
     cursor: pointer;
 }
 
-option {
-    display: none;
-}
-
 /* Custom <input type="range"> styles */
 input[type=range] {
     display: inline-block;
@@ -861,6 +857,20 @@ input[type=checkbox][switch]:checked {
 /* https://drafts.csswg.org/css-ui/#propdef-user-select */
 button, meter, progress, select {
     user-select: none;
+}
+
+/* https://html.spec.whatwg.org/multipage/rendering.html#the-select-element-2
+   FIXME: These styles apply to base appearance. We should differentiate between
+          native, primitive, and base appearance as per the spec. */
+select option {
+    min-inline-size: 24px;
+    min-block-size: max(24px, 1lh);
+    padding-inline: 0.5em;
+    padding-block-end: 0;
+    display: flex;
+    align-items: center;
+    gap: 0.5em;
+    white-space: nowrap;
 }
 
 /* https://drafts.csswg.org/css-view-transitions-1/#ua-styles */
--- a/Libraries/LibWeb/HTML/HTMLSelectElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLSelectElement.cpp
@@ -705,6 +705,7 @@ void HTMLSelectElement::create_shadow_tree_if_needed()
     m_inner_text_element = DOM::create_element(document(), HTML::TagNames::div, Namespace::HTML).release_value_but_fixme_should_propagate_errors();
     m_inner_text_element->set_attribute_value(HTML::AttributeNames::style, R"~~~(
         flex: 1;
+        white-space: nowrap;
     )~~~"_string);
     MUST(border->append_child(*m_inner_text_element));
 
--- a/Tests/LibWeb/Layout/expected/select-text-nowrap.txt
+++ b/Tests/LibWeb/Layout/expected/select-text-nowrap.txt
@@ -1,0 +1,29 @@
+Viewport <#document> at [0,0] [0+0+0 800 0+0+0] [0+0+0 600 0+0+0] [BFC] children: not-inline
+  BlockContainer <html> at [0,0] [0+0+0 800 0+0+0] [0+0+0 38 0+0+0] [BFC] children: not-inline
+    BlockContainer <body> at [8,8] [8+0+0 784 0+0+8] [8+0+0 22 0+0+8] children: inline
+      frag 0 from BlockContainer start: 0, length: 0, rect: [13,10 150x18] baseline: 16.796875
+      BlockContainer <select> at [13,10] inline-block [0+1+4 150 4+1+0] [0+1+1 18 1+1+0] [BFC] children: not-inline
+        Box <div> at [13,10] flex-container(row) [0+0+0 150 0+0+0] [0+0+0 18 0+0+0] [FFC] children: not-inline
+          BlockContainer <div> at [13,10] flex-item [0+0+0 417.328125 0+0+0] [0+0+0 18 0+0+0] [BFC] children: inline
+            frag 0 from TextNode start: 0, length: 50, rect: [13,10 417.328125x18] baseline: 13.796875
+                "This is some really long text that should not wrap"
+            TextNode <#text> (not painted)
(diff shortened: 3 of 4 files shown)


Libraries/LibWeb/CSS/Default.css
```diff
@@ -863,6 +863,11 @@ button, meter, progress, select {
     user-select: none;
 }
 
+/* Prevent text wrapping inside select elements */
+select {
+    white-space: nowrap !important;
+}
```
gmta, line 874: There are UA styles defined [here](https://html.spec.whatwg.org/multipage/rendering.html#the-select-element-2), one of which might solve the issue you're trying to solve as well:

```css
select option {
  min-inline-size: 24px;
  min-block-size: max(24px, 1lh);
  padding-inline: 0.5em;
  padding-block-end: 0;
  display: flex;
  align-items: center;
  gap: 0.5em;
  white-space: nowrap;
}
```

For UA styles, we try to copy them from the spec as much as possible. The spec does ask us to differentiate between native, primitive and base appearance of the `<select>` element, so you might need to add that as a `FIXME`.

konradekk: it seems that this may be problematic, esp. use of `!important`; the behaviour of firefox is [debated](https://bugzilla.mozilla.org/show_bug.cgi?id=1553930) (or at least debatable).

sheeki03:
Thanks for the review feedback!

Updated to use the spec-compliant `select option` base appearance styles instead of the Firefox-style approach. Added a FIXME about differentiating between native, primitive, and base appearance as the spec requires.

gmta:
@sheeki03 please do not add new commits to your branch, but amend the original commit. There should only be a single commit with your proposed change in here :) thanks!

sheeki03:
Done...made into a single commit
Thanks for the guidance!

Libraries/LibWeb/CSS/Default.css
```diff
@@ -863,6 +863,20 @@ button, meter, progress, select {
...
+select option {
+    min-inline-size: 24px;
+    min-block-size: max(24px, 1lh);
+    padding-inline: 0.5em;
+    padding-block-end: 0;
+    display: flex;
+    align-items: center;
+    gap: 0.5em;
+    white-space: nowrap;
+}
```
gmta, line 874: Looks good - only thing is that we have an `option { display: none; }` on line 71 of this file that I don't think is necessary anymore; can you remove that for this PR?

sheeki03: Done

gmta:
@sheeki03 your new test is failing in CI - did you run it successfully locally?

sheeki03:
Sorry for the confusion earlier @gmta my branch was based on an older commit and I hadn't fetched the latest master before running the tests locally, which is why I missed it
I've added white-space: nowrap directly to the inner text element in HTMLSelectElement.cpp to fix the collapsed view as well. The test now passes locally (along with all layout tests).

Libraries/LibWeb/CSS/Default.css
```diff
@@ -863,6 +859,20 @@ button, meter, progress, select {
...
```
gmta, line 874: If I remove this whole `select option { .. }` block, your test still passes. Is this helping in any way or is the test not covering this?

Libraries/LibWeb/HTML/HTMLSelectElement.cpp
```diff
@@ -705,6 +705,7 @@ void HTMLSelectElement::create_shadow_tree_if_needed()
     m_inner_text_element = DOM::create_element(document(), HTML::TagNames::div, Namespace::HTML).release_value_but_fixme_should_propagate_errors();
     m_inner_text_element->set_attribute_value(HTML::AttributeNames::style, R"~~~(
         flex: 1;
+        white-space: nowrap;
```
Psychpsyo, line 708: Is there a spec somewhere that tells us this should happen or why is it here?

Libraries/LibWeb/CSS/Default.css
sheeki03: the select option CSS doesn't affect this test since it only covers the collapsed select view (shadow DOM). Should I remove the CSS block and keep only the C++ fix? Or keep it for when dropdown options are rendered with actual <option> elements?

Libraries/LibWeb/HTML/HTMLSelectElement.cpp
sheeki03: The spec reference is in the commit message: 
https://html.spec.whatwg.org/multipage/rendering.html#the-select-element-2

It defines white-space: nowrap for select option base appearance. The collapsed select displays the selected option's text so this prevents wrapping

Psychpsyo: > The collapsed select displays the selected option's text

According to the spec reference that you linked, it displays the [select fallback button text](https://html.spec.whatwg.org/multipage/rendering.html#select-fallback-button-text), which is a string and doesn't come with any styling or CSS.

In fact, nothing in the spec will tell you that this should be here, because this behavior in other browsers currently stems from non-standard UA styles that apply to &lt;select>. Some of those are elaborated on here: https://github.com/whatwg/html/issues/8270

I'd say the correct solution to this is to introduce some of that to the default CSS instead of creating a new ad-hoc, non-spec workaround.

Also, if the code does things that the spec has opinions about, the spec should ideally be cited in the code. If there is no spec to cite, open spec issues should be cited, or there should be an explanation as to why a non-spec thing is necessary. If there is no open spec issues to cite, it's best to open one and then cite that.

sheeki03: Thanks for the detailed explanation and the link to the spec issue!...the C++ fix isn't fully spec-backed. I'll wait for gmta's reply on whether to remove the select option block and then update the PR to use a non-standard UA style on select instead, with a comment citing the open spec issue.
I've tested this approach locally and the tests are passing

Libraries/LibWeb/CSS/Default.css
```diff
@@ -863,6 +859,20 @@ button, meter, progress, select {
     user-select: none;
 }
 
+/* https://html.spec.whatwg.org/multipage/rendering.html#the-select-element-2
+   FIXME: These styles apply to base appearance. We should differentiate between
+          native, primitive, and base appearance as per the spec. */
```
gmta, line 864: You can leave out this `FIXME` from the commit message - having it in the code is enough.

Libraries/LibWeb/CSS/Default.css
gmta: Let's keep them both. You could make two distinct commits to clearly show which change relates to the test.

PR #7531, opened 2026-01-19
LibWeb/Layout: Ensure an element only has one `::backdrop`, as its previous sibling

### LINKED ISSUE #7486: Typing in the search dialog on MDN covers the page in darkness
To reproduce:
1. Visit https://developer.mozilla.org/en-US/
2. Click on the search in the top-right corner to bring up the dialog
3. Start typing in the search box that it gives you.

With each keypress, we get a new dark (or light) overlay matching the styling of the `::backdrop` pseudo-element. So my assumption is that we're creating multiple layout nodes for it.

After one "a" in the search box, in dark mode:

Image

I have completely failed to make a reduction of this, as neither extension I use for saving pages can handle the use of custom elements. I also couldn't build one myself which is confusing.

We create `::backdrop` oddly compared to other pseudo-elements. Specifically, we create the layout node here:
https://github.com/LadybirdBrowser/ladybird/blob/2ac363dcba7bf515e1d5f370facb20a6d92f363c/Libraries/LibWeb/Layout/TreeBuilder.cpp#L715-L724

`create_pseudo_element_if_needed()` doesn't check if there's already a matching pseudo-element with that ID on the element. The `update_tree_layout()` call also throws away the Element's pseudo-element data so it wouldn't be aware of the existing layout node anyway.

I've poked around with it a bit but my brain isn't working today, so making a note here. I do intend to come back to this at some point but if anyone wants to look in the meantime, you're welcome to.

AtkinsSJ:
Fixes #7486.

Details in commits. tl;dr: We kept creating a new layout node for `::backdrop` when layout updated, without removing the old one; and the new one was always appended, so if the originating element (eg `<dialog>`) had a partial layout update and wasn't appended too, the new `::backdrop` would go in front of it.

### DIFF
--- a/Libraries/LibWeb/Layout/TreeBuilder.cpp
+++ b/Libraries/LibWeb/Layout/TreeBuilder.cpp
@@ -272,13 +272,13 @@ private:
 
 GC_DEFINE_ALLOCATOR(GeneratedContentImageProvider);
 
-void TreeBuilder::create_pseudo_element_if_needed(DOM::Element& element, CSS::PseudoElement pseudo_element, AppendOrPrepend mode)
+GC::Ptr<NodeWithStyle> TreeBuilder::create_pseudo_element_if_needed(DOM::Element& element, CSS::PseudoElement pseudo_element, Optional<AppendOrPrepend> insertion_mode)
 {
     auto& document = element.document();
 
     auto pseudo_element_style = element.computed_properties(pseudo_element);
     if (!pseudo_element_style)
-        return;
+        return {};
 
     auto initial_quote_nesting_level = m_quote_nesting_level;
     DOM::AbstractElement element_reference { element, pseudo_element };
@@ -293,12 +293,12 @@ void TreeBuilder::create_pseudo_element_if_needed(DOM::Element& element, CSS::Ps
         && (pseudo_element_display.is_none()
             || pseudo_element_content.type == CSS::ContentData::Type::Normal
             || pseudo_element_content.type == CSS::ContentData::Type::None))
-        return;
+        return {};
 
     // For ::marker with content or display 'none' -- do nothing.
     if (pseudo_element == CSS::PseudoElement::Marker
         && (pseudo_element_display.is_none() || pseudo_element_content.type == CSS::ContentData::Type::None))
-        return;
+        return {};
 
     // For ::marker with content 'normal', create the marker pseudo-element from a ListItemMarkerBox
     // FIXME: This + ListItemBox + ListItemMarkerBox will disappear once ::marker pseudo-elements with 'normal' content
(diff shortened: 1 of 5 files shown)


Libraries/LibWeb/Layout/TreeBuilder.cpp
```diff
@@ -654,14 +659,36 @@ void TreeBuilder::update_layout_tree(DOM::Node& dom_node, TreeBuilder::Context&
...
+        auto& element = static_cast<DOM::Element&>(dom_node);
+        // Each element rendered in the top layer has a ::backdrop pseudo-element, for which it is the originating element.
+        if (element.rendered_in_top_layer() && context.layout_top_layer) {
+            // ::backdrop is a sibling of the element, not a child, so unlike other pseudo-elements, it's not
+            // automatically discarded when element's layout is recomputed.
+            // If we're not inserting a new node for element but replacing the existing one, we need to do the same for
+            // its ::backdrop's node, otherwise it will end up appended *after* element instead of before it.
+            if (may_replace_existing_layout_node) {
+                VERIFY(old_backdrop_node);
+                auto backdrop_node = create_pseudo_element_if_needed(element, CSS::PseudoElement::Backdrop, {});
```
tcl3: I think it's possible for this to return null here. If a style update changed the backdrop pseudo element to `display:none`, for example.

Libraries/LibWeb/Layout/TreeBuilder.cpp
```diff
@@ -654,14 +659,36 @@ void TreeBuilder::update_layout_tree(DOM::Node& dom_node, TreeBuilder::Context&
...
+                old_backdrop_node->parent()->replace_child(*backdrop_node, *old_backdrop_node);
+                element.set_pseudo_element_node({}, CSS::PseudoElement::Backdrop, backdrop_node);
```
tcl3: Is this call necessary? It looks like this gets called internally by `create_pseudo_element_if_needed()` (around line 348).

Libraries/LibWeb/Layout/TreeBuilder.cpp
tcl3: FWIW, I can cause a crash with this:
```html
<!DOCTYPE html>
<dialog id="d"><div id="x"></div></dialog>
<script>
    d.showModal();
    document.body.offsetWidth;

    const style = document.createElement('style');
    style.textContent = "dialog::backdrop { display: none }";
    document.head.appendChild(style);
    document.body.offsetWidth;

    x.hidden = true;
    document.body.offsetWidth;
</script>
```

Libraries/LibWeb/Layout/TreeBuilder.cpp
AtkinsSJ: It wasn't, I've removed it. Thanks!

Libraries/LibWeb/Layout/TreeBuilder.cpp
AtkinsSJ: Good catch, thanks for that! I've changed it so that we now always remove the old backdrop's layout node if it exists, and then if we do create a new one, we insert it before the originating element's layout node. It turns out to be slightly less code too. :tada: I added this as a test, slightly modified so that it also tests the "enabling a backdrop on an already existing element" case.

PR #7532, opened 2026-01-19
LibWeb: Add stubbed Speech API

gmta:
Just a couple of basic interfaces to make https://scummvm.kuendig.io/ not fail during initialization.


Libraries/LibWeb/Speech/SpeechSynthesis.cpp
```diff
@@ -0,0 +1,53 @@
...
+    set_event_handler_attribute(HTML::EventNames::voiceschanged, event_handler);
+}
+
+GC::Ptr<WebIDL::CallbackType> SpeechSynthesis::onvoiceschanged()
+{
+    return event_handler_attribute(HTML::EventNames::voiceschanged);
+}
+
+// https://wicg.github.io/speech-api/#dom-speechsynthesis-getvoices
+Vector<GC::Ref<SpeechSynthesisVoice>> SpeechSynthesis::get_voices() const
```
trflynn89: Even though this is a stub, we shouldn't return a `Vector<something gc'd>` by value. It's okay right now because it's always empty, but it'd be super easy for someone to return something real here in the future and forget that this isn't GC-safe. So I think we should either return something by const-ref here or return a RootVector.

Libraries/LibWeb/Speech/SpeechSynthesisUtterance.h
```diff
@@ -0,0 +1,79 @@
...
+
+    // https://wicg.github.io/speech-api/#dom-speechsynthesisutterance-rate
+    float rate() const { return m_rate; }
+    void set_rate(float rate) { m_rate = rate; }
+
+    // https://wicg.github.io/speech-api/#dom-speechsynthesisutterance-pitch
+    float pitch() const { return m_pitch; }
+    void set_pitch(float pitch) { m_pitch = pitch; }
+
+    void set_onstart(GC::Ptr<WebIDL::CallbackType>);
```
trflynn89: Up to you, but in other places (like `LibWeb/HTML/WorkerGlobalScope.h`), we use xmacros for these

Libraries/LibWeb/Speech/SpeechSynthesisVoice.cpp
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright (c) 2025, Jelle Raaijmakers <jelle@ladybird.org>
```
trflynn89: It is 2026 :)

gmta: Oof, force of ~habit~ copy/paste

Libraries/LibWeb/Speech/SpeechSynthesis.cpp
gmta: Yeah this was a silly mistake!

Libraries/LibWeb/Speech/SpeechSynthesisUtterance.h
gmta: Agreed!

PR #7533, opened 2026-01-19
LibThreading: BackgroundAction failure-mode thread affinity

jonbgamble:
BEHAVIOR CHANGE: Don't call on_error if cancelled, just reject.
Don't reject promise from the background thread.
Prevent data race on cancelled flag.

### DIFF
--- a/Libraries/LibThreading/BackgroundAction.h
+++ b/Libraries/LibThreading/BackgroundAction.h
@@ -8,6 +8,7 @@
 
 #pragma once
 
+#include <AK/Atomic.h>
 #include <AK/Function.h>
 #include <AK/NonnullRefPtr.h>
 #include <AK/Optional.h>
@@ -50,9 +51,11 @@ public:
     Optional<Result> const& result() const { return m_result; }
     Optional<Result>& result() { return m_result; }
 
-    void cancel() { m_canceled = true; }
+    // Cancellation is a best-effort cross-thread signal. No other state is protected by this flag.
+    // It is not used to synchronize access to any other state (m_result), so relaxed atomics are fine.
+    void cancel() { m_canceled.store(true, AK::MemoryOrder::memory_order_relaxed); }
     // If your action is long-running, you should periodically check the cancel state and possibly return early.
-    bool is_canceled() const { return m_canceled; }
+    bool is_canceled() const { return m_canceled.load(AK::MemoryOrder::memory_order_relaxed); }
 
 private:
     BackgroundAction(ESCAPING Function<ErrorOr<Result>(BackgroundAction&)> action, ESCAPING Function<ErrorOr<void>(Result)> on_complete, ESCAPING Optional<Function<void(Error)>> on_error = {})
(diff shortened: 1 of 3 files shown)


Libraries/LibThreading/BackgroundAction.h
```diff
@@ -62,53 +63,64 @@ class BackgroundAction final
...
-            self->m_canceled |= promise->is_rejected();
+            auto result = self->m_action(*self);
+            auto const has_job = static_cast<bool>(self->m_on_complete);
+            auto const canceled = self->m_canceled.load(AK::MemoryOrder::memory_order_relaxed);
 
-            // All of our work was successful and we weren't cancelled; resolve the event loop's promise.
-            if (!self->m_canceled && !result.is_error()) {
+            if (canceled) {
+                if (has_job)
+                    post_to_origin("promise rejection", [promise] mutable {
```
Zaggy1024: Adding a reference to a non-atomically-ref-counted object off the main thread is unsafe. This should probably be `promise = move(promise)`?

Libraries/LibThreading/BackgroundAction.h
```diff
@@ -62,53 +63,64 @@ class BackgroundAction final
...
+            };
             Core::EventLoop::current().add_job(promise);
         }
 
         if (on_error.has_value())
             m_on_error = on_error.release_value();
 
         enqueue_work([self = NonnullRefPtr(*this), promise = move(promise), origin_event_loop = Core::EventLoop::current_weak()]() mutable {
-            auto result = self->m_action(*self);
+            auto post_to_origin = [&](char const* message_type, Function<void()> callback) {
```
Zaggy1024: Please use StringView for compile-time strings.

Libraries/LibThreading/BackgroundAction.h
```diff
@@ -62,53 +63,64 @@ class BackgroundAction final
...
+                return;
+            }
+            if (!result.is_error()) {
                 self->m_result = result.release_value();
-
-                // If there is no completion callback, we don't rely on the user keeping around the event loop.
-                if (self->m_on_complete) {
-                    origin_event_loop->take()->deferred_invoke([self, promise = move(promise)] {
+                if (has_job)
+                    post_to_origin("on_complete", [self, promise]() mutable {
```
Zaggy1024: `promise = move(promise)`

Libraries/LibThreading/BackgroundAction.h
```diff
@@ -62,53 +63,64 @@ class BackgroundAction final
...
             }
+            auto error = result.release_error();
+            if (has_job)
+                post_to_origin("promise rejection", [promise, error = Error::copy(error)]() mutable {
+                    promise->reject(Error::copy(error));
+                });
+            if (self->m_on_error)
+                post_to_origin("on_error", [self, error = Error::copy(error)]() mutable {
+                    self->m_on_error(Error::copy(error));
+                });
```
Zaggy1024: The contents of these if statements take up multiple lines, so they should be wrapped with `{}`.

Libraries/LibThreading/BackgroundAction.h
```diff
@@ -50,9 +51,9 @@ class BackgroundAction final
     Optional<Result> const& result() const { return m_result; }
     Optional<Result>& result() { return m_result; }
 
-    void cancel() { m_canceled = true; }
+    void cancel() { m_canceled.store(true, AK::MemoryOrder::memory_order_relaxed); }
```
R-Goc, line 56: Could you document why all of the atomic operation are relaxed here? If we don't care about synchronization that is fine, but it is not immediately obvious.

Libraries/LibThreading/BackgroundAction.h
jonbgamble: yes, and self = move(self) i think.

Libraries/LibThreading/BackgroundAction.h
```diff
@@ -62,63 +65,80 @@ class BackgroundAction final
...
+                });
             }
         });
     }
 
     Function<ErrorOr<Result>(BackgroundAction&)> m_action;
     Function<ErrorOr<void>(Result)> m_on_complete;
     Function<void(Error)> m_on_error = [](Error error) {
-        dbgln("Error occurred while running a BackgroundAction: {}", error);
+        dbgln("Error occurred while running a BackgroundAction: {}"sv, error);
```
Zaggy1024: How did this sneak in? Hopefully this wasn't a side effect of me saying that compile-time strings should be StringView 😅 dbgln accepts a type that constructs from a string initializer to allow a compile-time argument count check, so we normally don't use ""sv. I'm honestly surprised this builds.

jonbgamble: I have no clue, I don't remember adding sv here and I use dbgln all the time :rofl:

jonbgamble: Looks harmless though:
```cpp
    template<typename T>
    CheckedFormatString(T const& unchecked_fmt)
    requires(requires(T t) { StringView { t }; })
        : m_string(unchecked_fmt)
    {
    }
```

Zaggy1024: Good to know that works I suppose, though it's certainly a waste of a couple of characters. 😅

Issue #2332, opened 2024-11-14
Unicode Math Symbols not rendering properly


nico-engels:
### Summary

When acessing sites that contains math symbols, the Ladybird browser do not render these characters. Instead ▯ appears, indicating that the font lacks the glyph.

### Operating system

Linux

### Steps to reproduce

1. Access [Unicode Math Wiki Page](https://en.wikipedia.org/wiki/Mathematical_operators_and_symbols_in_Unicode)

### Expected behavior

Get the tables of Mathematical symbols and operators.

### Actual behavior

Got tables of ▯ pretending to be Mathematical symbols and operators.

### URL for a reduced test case

N/A

### HTML/SVG/etc. source for a reduced test case

```html
<p>oi ↻ 😂 ∑</p>
```

### Log output and (if possible) backtrace

```shell
N/A
```

### Screenshots or screen recordings

![image](https://github.com/user-attachments/assets/08099763-d3e3-436b-8c41-a046e7156f40)

### Contribute a patch?

- [X] I’ll contribute a patch for this myself.

Randl:
Many of geometric shapes are not rendered too:
https://www.compart.com/en/unicode/block/U+25A0

PR #7536, opened 2026-01-19
LibGfx+LibWeb: Fall back to system fonts for missing glyphs

### LINKED ISSUE #91: Chinese support,any plan or the  way to support Chinese?
image
use the latest version of ladybird,English display is normal ,but  Chinese is not  normal?can  someone tell the way to support  Chinese?thanks.

tcl3:
When rendering text, if none of the fonts in the cascade list contain a glyph for a given code point, we now query Skia's font manager to find a system font that can render it.

The upside of using Skia here is that it works on both Linux and MacOS.

I haven't included any tests in the test suite for this. Testing is difficult because test mode deliberately retains its old behavior to maximize compatibility across platforms.

Previously, lots of non-Latin text would fail to render. Here's a before and after of [this test page](https://htmlpreview.github.io/?https://gist.githubusercontent.com/tcl3/9ac242fced0f35f42d3cbd19a9f9e84b/raw/test-rendering-test.html):

Before:
image

After:
image

Closes #91
Fixes: #2332

### DIFF
--- a/Libraries/LibGfx/Font/FontDatabase.cpp
+++ b/Libraries/LibGfx/Font/FontDatabase.cpp
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2018-2020, Andreas Kling <andreas@ladybird.org>
+ * Copyright (c) 2026, Tim Ledbetter <tim.ledbetter@ladybird.org>
  *
  * SPDX-License-Identifier: BSD-2-Clause
  */
@@ -8,6 +9,7 @@
 #include <LibCore/StandardPaths.h>
 #include <LibGfx/Font/Font.h>
 #include <LibGfx/Font/FontDatabase.h>
+#include <LibGfx/Font/TypefaceSkia.h>
 
 #if defined(AK_OS_HAIKU)
 #    include <FindDirectory.h>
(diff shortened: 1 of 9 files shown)


Libraries/LibGfx/Font/FontDatabase.cpp
```diff
@@ -48,6 +50,32 @@ RefPtr<Gfx::Font> FontDatabase::get(FlyString const& family, float point_size, u
...
+        // Cache the failure to avoid retrying
+        m_code_point_fallback_cache.set(key, { {}, nullptr });
+        return nullptr;
+    }
+
+    auto typeface = typeface_or_error.release_value();
+    auto family_name = typeface->family();
+    m_code_point_fallback_cache.set(key, { family_name, typeface });
+
+    return typeface->font(point_size, {});
```
gmta: I think you should be able to use HashMap::ensure(Callback) for the typeface.

gmta (commented):
Nice!

Libraries/LibGfx/Font/FontDatabase.cpp
tcl3: That's much cleaner!

gmta:
@tcl3 also fixes #2332 (check the original URL, not the test case which already works on `master`)

tcl3:
> @tcl3 also fixes #2332 (check the original URL, not the test case which already works on `master`)

I've confirmed  that #2332 is fixed with this change.

Libraries/LibWebView/Plugins/FontPlugin.h
```diff
@@ -23,6 +23,7 @@ class WEBVIEW_API FontPlugin final : public Web::Platform::FontPlugin {
     virtual Gfx::Font& default_fixed_width_font() override;
     virtual FlyString generic_font_name(Web::Platform::GenericFont) override;
     virtual Vector<FlyString> symbol_font_names() override;
+    virtual bool is_layout_test_mode() const override { return m_is_layout_test_mode; }
```
luozhiya, line 26: Nice

PR #7542, opened 2026-01-20
LibWeb/CSS: Improve color serialization and interpolation

AtkinsSJ:
This became a bit of a rabbit hole. :sweat_smile: 

There are two main changes to this:
- Introducing a `CSS::Color` type which wraps `Gfx::Color` along with the source `StyleValue`, used where we'd previously have a computed `Gfx::Color`.
- Changing color interpolation to keep in float space instead of converting to `Gfx::Color`'s 8-bit-per-channel format and back again repeatedly, and also to pay attention to the desired interpolation color space.

There are also a handful of teeny tiny serialization changes I came across.

Details are in the commit messages.


Calme1709:
I may be missing something obvious but I fail to see the need for the new `CSS::Color` type.

Reification, and interpolation should take place on "computed values" (i.e. `StyleValue`s) rather than "used values" (i.e. `Gfx::Color` or this new `CSS::Color`). This isn't the case currently and should be fixed.

Serialization is slightly more complicated in that we do need support used values, but helpfully it is the same as the computed value in most cases, the exceptions being `currentcolor` and color mix/relative color functions which have `currentcolor` as a parameter. IMO the best approach here would be to add a method to get the "used value" from the `StyleValue` stored within `ComputedProperties` rather than trying to work backwards from the `Gfx::Color`/`CSS::Color` stored within `ComputedValues` as we do currently.

To reiterate I think the correct approach here is:
- Implement proper absolutization of the various `ColorStyleValue` sub-classes (this is also required for proper inheritance).
- Update interpolation for colors to work with these absolutized `ColorStyleValue`s (see [this discussion](https://github.com/w3c/csswg-drafts/issues/12292) on how the CSSWG intends for this to work non-compute-time-resolvable colors)
- Add a way to get the correct "used value" of a color from the associated "computed value" which can be called from `style_value_for_computed_property()`

AtkinsSJ:
Thanks @Calme1709. My original intention was to avoid repeatedly calculating things by 1) storing a cached `Gfx::Color`, and 2) not creating a `StyleValue` for cases where we didn't have one already. Also, not being able to just store a `ColorStyleValue` put me off a bit. Didn't help that I had brain mush from a cold when I was working on most of this. :sweat_smile: 

I tried computing colours but got in a mess quickly because of properties depending on other properties, so I set it to one side to keep this PR simpler. But yeah that's the better approach. I did *not* know that `currentcolor` computes to itself, so that makes a big difference.

It's been bothering me for a long time that we store named colors as `KeywordStyleValue` and then have to check if they are colors or not. It stops us having a single "anything that is a color" style value which otherwise `ColorStyleValue` is. What are your thoughts on having a separate `ColorKeywordStyleValue` (and maybe even a `CurrentColorStyleValue`) for clarity? It'd be annoying for typed OM because they'd both reify the same and we'd have to know which type to convert back as, but I think that's the only situation, and knowing for sure that you're holding a color would be nice.

Calme1709:
>I tried computing colours but got in a mess quickly because of properties depending on other properties, so I set it to one side to keep this PR simpler. But yeah that's the better approach. I did not know that currentcolor computes to itself, so that makes a big difference.

I had a quick pass at this myself a little while back and put it in the "too hard for now" basket - happy to take another look if you don't get there first however. `currentcolor` is definitely the most confusing part and isn't that well defined in the spec itself, but there is a bit more conversation about it accessible through GH issues.

>It's been bothering me for a long time that we store named colors as KeywordStyleValue and then have to check if they are colors or not. It stops us having a single "anything that is a color" style value which otherwise ColorStyleValue is. What are your thoughts on having a separate ColorKeywordStyleValue (and maybe even a CurrentColorStyleValue) for clarity? It'd be annoying for typed OM because they'd both reify the same and we'd have to know which type to convert back as, but I think that's the only situation, and knowing for sure that you're holding a color would be nice.

To make this even more confusing, we actually only store `<system-color>`, `<deprecated-color>` and `currentcolor` as `KeywordStyleValues`, normal `<named-color>` is stored as a `RGBColorStyleValue` with the `name` member variable set (which affects serialization). Fully agree that a `ColorKeywordStyleValue` (including `<named-color>`) would be good - I don't think that having a separate `StyleValue` for `currentcolor` would be required but we can see when we get there.

AtkinsSJ:
I spent yesterday absolutizing some colour types, but didn't quite get to a PR for it. So far, it's alright. We have a lot of existing hacks to serialise things in the right way so untangling those is going to take some fiddling. Not helped by the fact that the spec says to always use the legacy comma syntax to serialise the old colour functions but WPT actually expects the modern syntax in some cases. I think it's whenever there's a calc in there? I got very confused for a while. 😅

I'll go ahead with trying ColorKeywordStyleValue. 👍

PR #7543, opened 2026-01-20
LibWeb/XPath: Register and attach namespaces for elements if needed

Simek:
# How

Register and attach all pre-defined namespaces for elements. Preserve HTML namespace optionally based on input expression, since both `//div` and `//html:div` XPaths should work according to WPT tests (unfortunately I was not able to find this information in XPath spec).

# Test plan

* http://wpt.live/domxpath/text-html-attributes.html [10 -> 11 passes]
    * Fixed: "svg element with XLink attribute"
*  https://wpt.live/domxpath/text-html-elements.html [4 -> 7 passes]
    * Fixed: "HTML elements namespace prefix"
    * Fixed: "SVG elements no namespace prefix"
    * Fixed: "SVG elements namespace prefix"


Simek:
@awesomekling Any chance for a review on that changeset?

PR #7544, opened 2026-01-20
LibWeb: Preserve paintable tree across relayouts

kalenikaliaksandr:
Reuse existing paintables during relayout to reduce GC allocation pressure. Each paintable subclass implements reset_for_relayout() to clear state before reuse.


Libraries/LibWeb/Painting/PaintableBox.cpp
```diff
@@ -74,6 +74,40 @@ PaintableBox::~PaintableBox()
 {
 }
 
+void PaintableBox::reset_for_relayout()
```
tcl3, line 77: I think you need to reset `m_perspective_matrix` here. It only gets conditionally cleared in `resolve_paint_properties()`, so the old value could stick around. AFAICT, all the other member variables get set unconditionally in resolve_paint_properties(),so shouldn't cause an issue.

kalenikaliaksandr: good catch! I did a fix for `PaintableBox::resolve_paint_properties()` in a separate commit.

PR #7545, opened 2026-01-20
LibWeb: Support composition of `BasicShapeStyleValue`

Calme1709:
This also required supporting composition of it's previously unsupported constituent types, namely `RadialSizeStyleValue`, `BorderRadiusRectStyleValue` and `EdgeStyleValue`.

Also includes a small fix for interpolation of `RadialSizeStyleValue`.

Gains us 248 WPT tests.


Libraries/LibWeb/CSS/Interpolation.cpp
```diff
@@ -2368,6 +2368,96 @@ RefPtr<StyleValue const> composite_value(StyleValue const& underlying_value, Sty
...
+            return {};
+
+        auto const composite_optional_position = [&](RefPtr<StyleValue const> underlying_position, RefPtr<StyleValue const> animated_position) -> Optional<RefPtr<StyleValue const>> {
+            if (!underlying_position && !animated_position)
+                return nullptr;
+
+            auto const& underlying_position_with_default = underlying_position ? underlying_position.release_nonnull() : PositionStyleValue::create_computed_center();
+            auto const& animated_position_with_default = animated_position ? animated_position.release_nonnull() : PositionStyleValue::create_computed_center();
+
+            auto interpolated_position = composite_value(underlying_position_with_default, animated_position_with_default, composite_operation);
```
tcl3: `composited_position` would match the naming of other variables here.

Libraries/LibWeb/CSS/Interpolation.cpp
```diff
@@ -2368,6 +2368,96 @@ RefPtr<StyleValue const> composite_value(StyleValue const& underlying_value, Sty
...
         return AngleStyleValue::create({ *result, underlying_value.as_angle().angle().unit() });
     }
+    case StyleValue::Type::BasicShape: {
+        auto const& underlying_basic_shape = underlying_value.as_basic_shape();
+        auto const& animated_basic_shape = animated_value.as_basic_shape();
+
+        if (underlying_basic_shape.basic_shape().index() != animated_basic_shape.basic_shape().index())
+            return {};
+
+        auto const composite_optional_position = [&](RefPtr<StyleValue const> underlying_position, RefPtr<StyleValue const> animated_position) -> Optional<RefPtr<StyleValue const>> {
```
tcl3: IMO, since this is only used twice, it would clearer to just duplicate the logic at each call site. This would avoid the awkward 3 state return.

Calme1709: Done.

Libraries/LibWeb/CSS/Interpolation.cpp
Calme1709: N/A since this function has been removed

Libraries/LibWeb/CSS/Interpolation.cpp
```diff
@@ -2361,6 +2368,99 @@ RefPtr<StyleValue const> composite_value(StyleValue const& underlying_value, Sty
...
+                if (underlying_polygon.points.size() != animated_polygon.points.size())
+                    return {};
+
+                Vector<Polygon::Point> composited_points;
+                composited_points.ensure_capacity(underlying_polygon.points.size());
+                for (size_t i = 0; i < underlying_polygon.points.size(); i++) {
+                    auto const& underlying_point = underlying_polygon.points[i];
+                    auto const& animated_point = animated_polygon.points[i];
+                    auto interpolated_point_x = composite_value(underlying_point.x, animated_point.x, composite_operation);
+                    auto interpolated_point_y = composite_value(underlying_point.y, animated_point.y, composite_operation);
```
tcl3: These should be `composited_point_*`

Calme1709: Fixed

Calme1709: Fixed

AtkinsSJ (approved):

PR #7547, opened 2026-01-21
Everywhere: Speculatively remove .clangd from .gitignore

jonbgamble:
It's confusing that it's in both the repo and .gitignore.

If this is intentional (to allow git --ignore-index shenanigans), I'm happy to learn about it

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,7 +20,6 @@ Toolchain/Local
 compile_commands.json
 .cache
 .clang_complete
-.clangd
 .idea/
 cmake-build-debug/
 output/


gmta:
Eh, looking at its entire history: the .gitignore entry was added by Linus in 4e6a50f1d03f0d03ac2d84bda4154f2764cf1f0d back in Sep 2020, then Mike added .clangd in b832837106b41e25f92e04536dba53f7d8df4320 almost four years later, followed by some changes by Tim in 4ffca2089e31805c447f2abb8fcfa58aad3b4b36 two months later.

Seems like it should just be handled like any other file checked in to our repository.

gmta (approved):

PR #7552, opened 2026-01-21
LibWeb: Implement External interface

ReimarPB:
Implements the (deprecated) `window.external` property with two dummy methods, `AddSearchProvider()` and `IsSearchProviderInstalled()`. They do nothing, but are part of the spec for historical reasons.

https://html.spec.whatwg.org/multipage/obsolete.html#external


Libraries/LibWeb/HTML/External.cpp
```diff
@@ -0,0 +1,47 @@
...
+namespace Web::HTML {
+
+GC_DEFINE_ALLOCATOR(External);
+
+void External::initialize(JS::Realm& realm)
+{
+    WEB_SET_PROTOTYPE_FOR_INTERFACE(External);
+    Base::initialize(realm);
+}
+
```
gmta, line 21: Stray newline

Libraries/LibWeb/HTML/Window.cpp
```diff
@@ -1784,6 +1786,17 @@ void Window::release_events()
     // Do nothing.
 }
 
+// https://html.spec.whatwg.org/multipage/obsolete.html#dom-external
+GC::Ref<External> Window::external()
+{
+    auto& realm = this->realm();
```
gmta: You can simply use `realm()` as the argument to `realm.create<External>(..)` below.

Libraries/LibWeb/HTML/Window.cpp
```diff
@@ -1784,6 +1786,17 @@ void Window::release_events()
...
+
+    // The external attribute of the Window interface must return an instance of the External interface
+    if (!m_external)
+        m_external = realm.create<External>(realm);
+    return GC::Ref { *m_external };
```
gmta: This `GC::Ref { .. }` should not be necessary.

Libraries/LibWeb/HTML/Window.idl
```diff
@@ -118,9 +119,12 @@ interface Window : EventTarget {
     // https://w3c.github.io/selection-api/#extensions-to-window-interface
     Selection? getSelection();
 
+    // also has obsolete members
     undefined captureEvents();
     undefined releaseEvents();
 
+    [Replaceable, SameObject] readonly attribute External external;
```
gmta: Looks like this is defined in a partial interface?

https://html.spec.whatwg.org/multipage/obsolete.html#window-partial

Consider adding that as a separate file and importing it from Window.idl.

ReimarPB: Would make sense, although I'm unsure what to name the file. It is just a general partial interface for all obsolete window properties.

It seems that most of the other obsolete partial interfaces on that page have been inlined in Ladybird, but that might not be ideal either.

gmta: We haven't supported `partial interface` for that long, so maybe that's why they were inlined previously. We could have a `WindowDeprecated.idl` and import that?

Libraries/LibWeb/HTML/Window.idl
```diff
@@ -7,10 +7,12 @@
 #import <DOM/EventTarget.idl>
 #import <HTML/AnimationFrameProvider.idl>
 #import <HTML/CustomElements/CustomElementRegistry.idl>
+#import <HTML/External.idl>
```
gmta: Hm, shouldn't this #import go into WindowDeprecated.idl now?

Libraries/LibWeb/HTML/WindowDeprecated.idl
```diff
@@ -0,0 +1,8 @@
+#import <HTML/External.idl>
+
+partial interface Window {
```
gmta, line 4: Add a spec link before this line: https://html.spec.whatwg.org/multipage/obsolete.html#external

gmta (commented):
Looking good, final remark!

PR #7553, opened 2026-01-21
Tests/LibWeb: Only report fuzzy mismatches if we expect a match

gmta:
Reduces the fuzzy matching noise `test-web` produces.


shannonbooth:
Is there a point in running the fuzzy match at all if we don't expect a match?

gmta:
> Is there a point in running the fuzzy match at all if we don't expect a match?

A mismatch should still apply the fuzzy config afaik? Because something could match within the fuzzy parameters, and if the ref test expects a mismatch we would only know that by running through the entire logic.

PR #7555, opened 2026-01-21
LibWeb: Use local coordinates for overflow clip hit testing

gmta:
Fixes being able to click on the popover menu buttons in Roundcube webmail.

image

CC @kalenikaliaksandr


kalenikaliaksandr (approved):
oh, that's great catch, thanks for fixing that!

PR #7556, opened 2026-01-21
LibWeb: Implement `word-break` functionality using UAX#14 line breaking

tcl3:
~~NOTE: This is built on top of  #7536, which should be merged first.~~

We now use ICU's line break iterator to determine soft wrap opportunities, enabling correct line breaking for CJK text.

This also gains us 780 WPT subtest passes in the `css/css-text` directory.

The improvement can be clearly seen on the Chinese Wikipedia homepage:

Before:

image

After:

image

### DIFF
--- a/Libraries/LibWeb/Layout/TextNode.cpp
+++ b/Libraries/LibWeb/Layout/TextNode.cpp
@@ -2,12 +2,14 @@
  * Copyright (c) 2018-2021, Andreas Kling <andreas@ladybird.org>
  * Copyright (c) 2022, Tobias Christiansen <tobyase@serenityos.org>
  * Copyright (c) 2025, Jelle Raaijmakers <jelle@ladybird.org>
+ * Copyright (c) 2026, Tim Ledbetter <tim.ledbetter@ladybird.org>
  *
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
 #include <AK/CharacterTypes.h>
 #include <AK/StringBuilder.h>
+#include <AK/UnicodeUtils.h>
 #include <LibUnicode/CharacterTypes.h>
 #include <LibUnicode/Locale.h>
 #include <LibWeb/DOM/Document.h>
@@ -296,6 +298,7 @@ void TextNode::invalidate_text_for_rendering()
 {
     m_text_for_rendering = {};
     m_grapheme_segmenter.clear();
+    m_line_segmenter.clear();
 }
 
 Utf16String const& TextNode::text_for_rendering() const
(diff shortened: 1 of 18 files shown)


tcl3:
My latest push updates the spec comments for `word-break` to those from css-text-4. The implementation is unchanged.

Libraries/LibWeb/Layout/TextNode.cpp
```diff
@@ -531,6 +547,96 @@ TextNode::Chunk TextNode::ChunkIterator::create_empty_chunk()
...
+        if (m_current_index == 0)
+            return {};
+        size_t previous_index = m_current_index - 1;
+        if (previous_index > 0 && AK::UnicodeUtils::is_utf16_low_surrogate(m_view.code_unit_at(previous_index)))
+            previous_index--;
+        auto code_point = m_view.code_point_at(previous_index);
+        while (Unicode::line_break_class(code_point) == Unicode::LineBreakClass::CombiningMark && previous_index > 0) {
+            --previous_index;
+            if (previous_index > 0 && AK::UnicodeUtils::is_utf16_low_surrogate(m_view.code_unit_at(previous_index)))
+                previous_index--;
```
gmta: Hm, maybe we could add some logic to Utf16View::code_point_at() so we don't need to deal with this here?

gmta (commented):
Very nice!

Libraries/LibWeb/Layout/TextNode.cpp
tcl3: It looks like changing `code_point_at()` would break some test expectations. Instead, I've added another method to Utf16View to get the previous code point from a given index.

Libraries/LibUnicode/Segmenter.cpp
```diff
@@ -21,6 +21,8 @@ SegmenterGranularity segmenter_granularity_from_string(StringView segmenter_gran
 {
     if (segmenter_granularity == "grapheme"sv)
         return SegmenterGranularity::Grapheme;
+    if (segmenter_granularity == "line"sv)
```
trflynn89, line 24: Can we add some basic line segmentation tests to `Tests/LibUnicode/TestSegmenter.cpp`?

AK/Utf16View.h
```diff
@@ -387,6 +387,16 @@ class Utf16View {
         return UnicodeUtils::decode_utf16_surrogate_pair(code_point, second);
     }
 
+    [[nodiscard]] constexpr u32 previous_code_point_at(size_t& index) const
+    {
+        VERIFY(index > 0);
+        VERIFY(index <= length_in_code_units());
+        --index;
+        if (index > 0 && UnicodeUtils::is_utf16_low_surrogate(code_unit_at(index)))
```
trflynn89: Our UTF-16 is actually wobbly (WTF-16), meaning it allows unpaired surrogates. Should this check if the preceding code unit was a high surrogate, returning the low surrogate if not?

For example:
```c++
auto s = u"a""\xdc00""c"sv;
size_t i = 2;
s.previous_code_point_at(i); // returns 'a'
```

(Tests in Tests/AK/TestUtf16View.cpp as well please 😃)

trflynn89 (approved):

PR #7557, opened 2026-01-21
LibWeb: Move effects application into AccumulatedVisualContext

kalenikaliaksandr:
Effects (opacity, blend mode, filters) must be applied in the parent's coordinate space, before the element's transform. Previously this was handled by manually switching to the parent's visual context when applying effects at paint time.

By adding EffectsData to AccumulatedVisualContext and positioning it before TransformData in the chain, effects are now naturally applied in the correct order during display list replay, eliminating the special case in StackingContext::paint().

For SVG filters that can generate content from empty elements (feFlood, feImage, feTurbulence), a transparent FillRect command is emitted to trigger the filter through the same AVC pipeline.

### DIFF
--- a/Libraries/LibWeb/Painting/AccumulatedVisualContext.cpp
+++ b/Libraries/LibWeb/Painting/AccumulatedVisualContext.cpp
@@ -66,6 +66,10 @@ Optional<CSSPixelPoint> AccumulatedVisualContext::transform_point_for_hit_test(C
                 if (!clip_path.path.contains(point.to_type<float>(), clip_path.fill_rule))
                     return {};
                 return point;
+            },
+            [&](EffectsData const&) -> Optional<CSSPixelPoint> {
+                // Effects don't affect coordinate transforms
+                return point;
             });
 
         if (!result.has_value())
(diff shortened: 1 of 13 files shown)


Libraries/LibWeb/Painting/AccumulatedVisualContext.cpp
```diff
@@ -134,6 +139,20 @@ void AccumulatedVisualContext::dump(StringBuilder& builder) const
         [&](ClipPathData const& clip_path) {
             auto const& rect = clip_path.bounding_rect;
             builder.appendff("clip_path=[bounds: {},{} {}x{}, path: {}]", rect.x().to_float(), rect.y().to_float(), rect.width().to_float(), rect.height().to_float(), clip_path.path.to_svg_string());
+        },
+        [&](EffectsData const& effects) {
+            builder.append("effects=["sv);
+            if (effects.opacity < 1.0f)
+                builder.appendff("opacity={}", effects.opacity);
+            if (effects.blend_mode != Gfx::CompositingAndBlendingOperator::Normal)
+                builder.appendff(" blend_mode={}", static_cast<int>(effects.blend_mode));
```
gmta: Extreme nit but the unconditional spaces hurt 😅 can't we either unconditionally add these properties, or call .join(), or keep track of previously added items?

Libraries/LibWeb/Painting/DisplayList.cpp
```diff
@@ -99,26 +100,36 @@ void DisplayListPlayer::execute_impl(DisplayList& display_list, ScrollStateSnaps
     };
 
     auto apply_accumulated_visual_context = [&](AccumulatedVisualContext const& node) {
-        save({});
-
+        // Don't call save() unconditionally - let each handler decide (save() vs saveLayer())
```
gmta: This comment is a bit redundant, and we don't use `saveLayer()` here right?

gmta (approved):

PR #7559, opened 2026-01-21
LibWeb: Don't disentangle remote port when closing MessagePort

Lubrsi:
Otherwise, the remote port will lose its transport and not receive
queued messages. The remote port will automatically close anyway when
EOF is received on the socket.

This allows https://www.tripadvisor.com/ to load, where it instantiates
a module by creating a MessageChannel, setting port1's onmessage to the
module's instantiation function, posting an undefined message on port2
and then immediately closing port2.

### DIFF
--- a/Libraries/LibWeb/HTML/MessagePort.cpp
+++ b/Libraries/LibWeb/HTML/MessagePort.cpp
@@ -151,13 +151,9 @@ WebIDL::ExceptionOr<void> MessagePort::transfer_receiving_steps(HTML::TransferDa
 
 void MessagePort::disentangle()
 {
-    if (auto remote_port = m_remote_port) {
-        // Set the pointers to null before disentangling the remote port to prevent infinite recursion here.
+    if (m_remote_port) {
         m_remote_port->m_remote_port = nullptr;
         m_remote_port = nullptr;
-
-        if (remote_port)
-            remote_port->disentangle();
     }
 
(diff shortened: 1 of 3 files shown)


Libraries/LibWeb/HTML/MessagePort.cpp
```diff
@@ -151,13 +151,9 @@ WebIDL::ExceptionOr<void> MessagePort::transfer_receiving_steps(HTML::TransferDa
...
 {
-    if (auto remote_port = m_remote_port) {
-        // Set the pointers to null before disentangling the remote port to prevent infinite recursion here.
+    if (m_remote_port) {
         m_remote_port->m_remote_port = nullptr;
         m_remote_port = nullptr;
-
-        if (remote_port)
-            remote_port->disentangle();
     }
```
awesomekling, line 157: This diff feels like it's out of sync with the commit message.

"Don't disentangle remote port when..." but then all the logic is *inside* `disentangle()` :thonk:

gmta: I think the message is fine? The message port still gets disentangled, but we no longer disentangle the _remote_ port. That's something that happens implicitly by closing the transport, which is also initiated here.

awesomekling: Yeah on reread it seems okay

PR #7028, opened 2025-12-05
LibWeb: dont inject natural sizes from prepare_for_replaced_layout

jonbgamble:
Instead, compute them in accessors.

Add TextAreaBox and TextInputBox to Layout so that textarea and text inputs can be stretched when auto.

Fix #634, fix #6040, fix #7070, and possibly related issues involving baseline, cross axis sizing, aspect ratio, canvas, size containment, and textarea

PR #7562, opened 2026-01-21
LibWeb: Derive inline-block baseline from nested content

gmta:
Compute inline-block baselines by traversing into nested block children to find the last in-flow line box, using correct offsets relative to the margin box edge.

Also ensure inline-flex and inline-grid containers always derive their baseline from content (per CSS Align), and add special handling for <input> elements which have `overflow: clip` in the UA stylesheet but should still align adjacent text with their internal content.

Fixes the positioning of the toolbar buttons in Roundcube webmail.

| Test | Before | After | Firefox |
|--|--|--|--|
| Roundcube webmail | image | image | image |
| Layout/input/inline-block-baseline-alignment.html | image | image | image |
| Layout/input/form-control-baseline-alignment.html | image | image | image |
| Layout/input/empty-editable-shows-cursor.html | image | image | image |

### DIFF
--- a/Libraries/LibWeb/Layout/FormattingContext.cpp
+++ b/Libraries/LibWeb/Layout/FormattingContext.cpp
@@ -5,6 +5,7 @@
  */
 
 #include <LibWeb/Dump.h>
+#include <LibWeb/HTML/HTMLInputElement.h>
 #include <LibWeb/Layout/BlockFormattingContext.h>
 #include <LibWeb/Layout/Box.h>
 #include <LibWeb/Layout/FlexFormattingContext.h>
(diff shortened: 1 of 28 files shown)


Libraries/LibWeb/Layout/FormattingContext.cpp
```diff
@@ -1945,28 +1945,26 @@ Box const* FormattingContext::box_child_to_derive_baseline_from(Box const& box)
 {
     if (!box.has_children() || box.children_are_inline())
```
jonbgamble, line 1947: ~~I believe this needs a guard to [avoid descending into textarea](https://github.com/LadybirdBrowser/ladybird/pull/7028/changes#diff-68eef508eb5cb4c32b181ed2ad4581ae93eb573ab6ce606fe1632f6ac9b14cbeR1810). Unlike other text nodes, its baseline is bottom edge of the margin box.~~

jonbgamble: Scratch that, the misalignment I noticed is not due to textarea. If it helps, here's what I tested with:
```
<!DOCTYPE html>
<input/><span>boo</span><textarea></textarea><span>boo</span><input placeholder="hello"/>
<span>boo</span><textarea placeholder="hello"></textarea>
```
And this is what I think we want (Chrome on top, below it is the libweb-lazy-replaced-sizing PR):
image

In #7562 PR the \<input> texts are margin aligned too. Maybe they report baseline differently than the lazy sizing branch.

gmta: Oh yeah, instead of handling `<textarea>` separately, we can just default that to normal content alignment. We do need special handling for `<input>` since it has `overflow: clip` in the UA stylesheet.

Updated the PR description with some screenshots!

PR #7564, opened 2026-01-21
LibJS: Implement megamorphic inline caching for property access

tim-gromeyer:
This PR implements a megamorphic inline cache (IC) for the LibJS bytecode VM. Previously, property lookups `get_by_id`, `put_by_property_key` would drop off a performance cliff after encountering more than 4 object shapes, falling back to an expensive slow-path lookup. We now use a hybrid strategy: the fast 4-slot linear cache is maintained for common cases, followed by an on-demand megamorphic `HashMap` for sites with high polymorphism.

### Impact
This primarily benefits generic utility functions, library hot-paths, and framework internals that handle many different object shapes at a single call site. For these megamorphic sites, we now recover property offsets and validity metadata in O(1) time, avoiding repeated prototype chain walks and shape validation.

### Performance Results (2.0M iterations)
- **SET operations (10-200 shapes):** 3.4x to 4.4x faster.
- **GET operations:** No regression for monomorphic/polymorphic paths, with stable O(1) performance even with hundreds of shapes.
- **Memory:** Monomorphic sites carry no additional heap overhead; the `HashMap` is only allocated lazily for sites that actually exhibit high polymorphism.

### Verification
- **Correctness tests:** Libraries/LibJS/Tests/megamorphic-ic.js
- **Benchmark suite:** Tests/LibJS/bench-megamorphic.js

### DIFF
--- a/Libraries/LibJS/Bytecode/MegamorphicCache.cpp
+++ b/Libraries/LibJS/Bytecode/MegamorphicCache.cpp
@@ -1,0 +1,161 @@
+/*
+ * Copyright (c) 2025, Andreas Kling <andreas@ladybird.org>
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include <LibJS/Bytecode/MegamorphicCache.h>
+#include <LibJS/Runtime/Object.h>
+#include <LibJS/Runtime/Shape.h>
+
+namespace JS::Bytecode {
+
+MegamorphicCache::MegamorphicCache()
+{
+    clear();
+}
+
+void MegamorphicCache::clear()
+{
+    for (auto& line : m_get_cache) {
+        for (auto& entry : line.entries) {
+            entry = {};
+        }
+    }
+    for (auto& line : m_put_cache) {
+        for (auto& entry : line.entries) {
+            entry = {};
+        }
+    }
+}
+
+// FNV-1a hash function for fast, good distribution
+static constexpr size_t fnv1a_hash(void const* data, size_t size, size_t seed = 0xcbf29ce484222325ULL)
+{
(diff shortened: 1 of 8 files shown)


gmta:
Hi, welcome to the project! Can you show the comparison of old and new [js-benchmarks](https://github.com/LadybirdBrowser/js-benchmarks) results of this branch? It's what we use to track JS/Wasm performance in general. There's a tool in there to produce a comparison table.

tim-gromeyer:
Hi, yes I will do that. and I might have to reimplement it as I think I found a bug (memory leak) in the code

tim-gromeyer:
Hi! I’m working on generating the js-benchmarks comparison now. Running the full benchmark suite (including building the js binary) locally is taking quite a long time on my machine (multiple hours), and I’m seeing frequent large rebuilds along the way.

Before I go further, do you have any tips or recommended workflow to speed up the build/benchmark cycle for this use case? I’m running on a Linux laptop with an Intel® Core™ i5-11300H.

Thanks!

tcl3:
> Before I go further, do you have any tips or recommended workflow to speed up the build/benchmark cycle for this use case? I’m running on a Linux laptop with an Intel® Core™ i5-11300H.

Hi! You should be able to build just the `js` binary without building everything else. When running the benchmarks it's also best to use the Distribution preset, to get the best performance. You can do this by running `BUILD_PRESET=Distribution ./Meta/ladybird.py build js`. Use the `--executable` flag when running the benchmarks to point to the distribution binary.

tim-gromeyer:
After some testing, this is far more complex than I thought and it makes real workflows slower, even though the micro-benchmarks showed remarkable improvements. I'll start with smaller, easier things

PR #7567, opened 2026-01-22
LibWeb: Better handling for union types in IDL code generation

jonbgamble:
First check if a string is a member of an enum before attempting numeric conversion. This generates correct code for fields like `AudioContextLatencyCategory | double`

### DIFF
--- a/Meta/Lagom/Tools/CodeGenerators/LibWeb/BindingsGenerator/IDLGenerators.cpp
+++ b/Meta/Lagom/Tools/CodeGenerators/LibWeb/BindingsGenerator/IDLGenerators.cpp
@@ -1606,8 +1606,8 @@ static void generate_to_cpp(SourceGenerator& generator, ParameterType& parameter
 )~~~");
             // NOTE: generate_to_cpp doesn't use the parameter name.
             // NOTE: generate_to_cpp will use to_{u32,etc.} which uses to_number internally and will thus use TRY, but it cannot throw as we know we are dealing with a number.
-            IDL::Parameter parameter { .type = *numeric_type, .name = ByteString::empty(), .optional_default_value = {}, .extended_attributes = {} };
-            generate_to_cpp(union_generator, parameter, js_name, js_suffix, ByteString::formatted("{}{}_number", js_name, js_suffix), interface, false, false, {}, false, recursion_depth + 1);
+            IDL::Parameter idl_parameter { .type = *numeric_type, .name = parameter.name, .optional_default_value = {}, .extended_attributes = {} };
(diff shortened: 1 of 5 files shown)


shannonbooth:
I am a little confused here, where does this pattern surface? It doesn't seem to come up in AudioContextConstructor as an example here where AudioContextLatencyCategory is used. Is this for some other interface it results in wrong codgen?

jonbgamble:
https://developer.mozilla.org/en-US/docs/Web/API/AudioContext/AudioContext#latencyhint

shannonbooth:
To be more clear to what I mean, I pulled this patch and applied it - the generated code for AudioContextConstructor.cpp where this option was used was the same before and after, so I am trying to understand in what case we are currently generating the wrong code, maybe it is for an interface you have sitting in your fork somewhere? (Or I just missed something!)

shannonbooth:
Oh, I see the oopsy I made, AudioContext IDL doesnt match the spec

jonbgamble:
Correct, this fix is not currently used in master (same as my other PRs really, except for #7028)

Meta/Lagom/Tools/CodeGenerators/LibWeb/BindingsGenerator/IDLGenerators.cpp
```diff
@@ -1590,6 +1590,49 @@ static void generate_to_cpp(SourceGenerator& generator, ParameterType& parameter
...
 
+        bool includes_enumeration = false;
+        for (auto& type : types) {
+            if (interface.enumerations.contains(type->name())) {
+                includes_enumeration = true;
+                break;
+            }
+        }
+
+        // If V is a string and the union contains an enumeration type, attempt to match the string against the enumeration's values.
```
shannonbooth: The comment here says: "where strings should be interpreted as enums before numeric conversion"
Is there really any particular reason that this needs to happen before numeric conversion? From what I can gather, the problem here is that the IDL spec (and web specs in general) treats enumerations as a javascript string, but since we are c++ land we give this a more typesafe C++ enumuration in the IDL generator. And the problem here is that previously we did not support making this conversion inside of unions, which gives a compiler error. Logically, if we are matching the spec this feels like it should be part of step 14 below

jonbgamble: It was for a practical reason - previously in generated code, the numeric fallback ran on JS strings and that led to crashing/asserting. I agree that it fits better in the string section to stay aligned with the spec structure.

jonbgamble:
I moved the offending code block down closer to 14 (no functional change) and added an enum|double to AudioContext idl (to exercise this).

PR #7568, opened 2026-01-22
LibWeb: Account for border-radius in hit testing

kalenikaliaksandr:
This change adds border-radius awareness to hit testing in two places:
1. ClipData::contains() now uses BorderRadiiData::contains() to properly check if a point is inside a rounded clip rect. This handles overflow clips from ancestor elements that have border-radius.
2. PaintableBox::hit_test() now directly checks the element's own border-radius before reporting a hit.

### DIFF
--- a/Libraries/LibWeb/Painting/BorderRadiiData.h
+++ b/Libraries/LibWeb/Painting/BorderRadiiData.h
@@ -65,6 +65,54 @@ struct BorderRadiiData {
         return top_left || top_right || bottom_right || bottom_left;
     }
 
+    bool contains(CSSPixelPoint point, CSSPixelRect const& rect) const
+    {
+        if (!rect.contains(point))
+            return false;
+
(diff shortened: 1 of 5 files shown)


Libraries/LibWeb/Painting/BorderRadiiData.h
```diff
@@ -65,6 +65,54 @@ struct BorderRadiiData {
         return top_left || top_right || bottom_right || bottom_left;
     }
 
+    bool contains(CSSPixelPoint point, CSSPixelRect const& rect) const
+    {
+        if (!has_any_radius())
+            return rect.contains(point);
+
+        if (!rect.contains(point))
+            return false;
```
gmta, line 71: If you move this to the top, you can simplify it slightly:

```cpp
        if (!rect.contains(point))
            return false;

        if (!has_any_radius())
            return true;
```

PR #7569, opened 2026-01-22
Big pile of IPC hardening changes

awesomekling:
See individual commits, but basically run with two principles:

- Any kind of suspicious/malformed incoming message should cause us to disconnect the peer immediately with a debug message.
- Any kind of weirdness when trying to encode a message means we're in a bad state, crash immediately instead of trying to recover.

### DIFF
--- a/Libraries/LibCore/Proxy.h
+++ b/Libraries/LibCore/Proxy.h
@@ -21,7 +21,7 @@ struct ProxyData {
     } type { Type::Direct };
 
     IPv4Address host_ipv4;
-    int port { 0 };
+    u16 port { 0 };
 
     bool operator==(ProxyData const& other) const = default;
 
--- a/Libraries/LibIPC/Decoder.cpp
+++ b/Libraries/LibIPC/Decoder.cpp
@@ -178,7 +178,7 @@ ErrorOr<Core::ProxyData> decode(Decoder& decoder)
 {
     auto type = TRY(decoder.decode<Core::ProxyData::Type>());
     auto host_ipv4 = IPv4Address(TRY(decoder.decode<u32>()));
-    auto port = TRY(decoder.decode<int>());
+    auto port = TRY(decoder.decode<u16>());
 
     return Core::ProxyData { type, host_ipv4, port };
 }


Libraries/LibIPC/Decoder.cpp
```diff
@@ -179,6 +179,8 @@ ErrorOr<Core::ProxyData> decode(Decoder& decoder)
     auto type = TRY(decoder.decode<Core::ProxyData::Type>());
     auto host_ipv4 = IPv4Address(TRY(decoder.decode<u32>()));
     auto port = TRY(decoder.decode<int>());
+    if (port < 0 || port > 65535)
```
shannonbooth: 🤔  should we just make ProxyData's port a u16 so that it can't be out of range?

awesomekling: You make a good point!

Libraries/LibIPC/Limits.h
```diff
@@ -0,0 +1,19 @@
...
+ */
+
+#pragma once
+
+#include <AK/Types.h>
+
+namespace IPC {
+
+// Maximum size of an IPC message payload (64 MiB should be more than enough)
+static constexpr size_t MAX_MESSAGE_PAYLOAD_SIZE = 64 * MiB;
```
shannonbooth, line 14: Where does 64MiB come from? I wonder if this is really going to be enough

awesomekling: It's just some number. If we want to transfer large things cross-process we should use shared memory IMO!

shannonbooth: Yeah makes sense, I'm thinking of Blob URLs as an example here where our current architecture isn't really sufficient (done in a very naive, simply make it work type of way)

Libraries/LibIPC/TransportSocket.cpp
```diff
@@ -340,11 +346,21 @@ void TransportSocket::read_incoming_messages()
             break;
         }
 
+        if (m_unprocessed_bytes.size() + bytes_read.size() > MAX_UNPROCESSED_BUFFER_SIZE) {
+            dbgln("TransportSocket: Unprocessed buffer would exceed {} bytes, disconnecting peer", MAX_UNPROCESSED_BUFFER_SIZE);
+            m_peer_eof = true;
+            break;
+        }
```
Hendiadyoin1, line 348: It's probably to late to send a `Busy`/`RetryAfter` response to the client at this point, isn't it?
(The commit in general, not just these 5 lines)

Libraries/LibCore/Proxy.h
```diff
@@ -18,6 +18,7 @@ struct ProxyData {
     enum Type {
         Direct,
         SOCKS5,
+        __Count,
```
trflynn89: I've wondered about bringing in Magic Enum for stuff like this:
```c++
enum class Color {
    Red,
    Blue,
    Green,
    Orange,
};

constexpr auto count = magic_enum::enum_count<Color>(); // 4
```

(We could also implement the few things we might need, seems like it's a little bit of template magic using index sequences)

But using that would let us avoid touching every enum and switch statement

Libraries/LibIPC/TransportSocket.cpp
awesomekling: I suppose we could get into a "pls retry in a bit" dance with the peer if we wanted to. Cutting them off hard at 128 MiB (or some other limit that makes sense) feels a lot simpler, so it would be nice if we didn't have to get fancier.

Libraries/LibCore/Proxy.h
awesomekling: How about I drop this commit and we look at enums separately? :)

PR #7576, opened 2026-01-22
LibWeb: Skip effect contexts for off-screen display list commands

kalenikaliaksandr:
When executing display list commands, check if commands with effect contexts (opacity, filters, blend modes) are outside the viewport before applying the effect. Since effects don't affect clip state, would_be_fully_clipped_by_painter() returns the same result before and after applying effects.

This avoids expensive saveLayer/restore cycles for off-screen commands with effects like blur, which is particularly beneficial for pages with many blurred decorative images (e.g., Discord's landing page has 70+ blurred star images).

The optimization only applies when switching to a new effect context, not for consecutive commands with the same context, to preserve correct blend mode compositing behavior.

### DIFF
--- a/Libraries/LibWeb/Painting/DisplayList.cpp
+++ b/Libraries/LibWeb/Painting/DisplayList.cpp
@@ -168,6 +168,21 @@ void DisplayListPlayer::execute_impl(DisplayList& display_list, ScrollStateSnaps
     for (size_t command_index = 0; command_index < commands.size(); command_index++) {
         auto const& [context, command] = commands[command_index];
 
+        auto bounding_rect = command_bounding_rectangle(command);
+
+        // OPTIMIZATION: If the leaf context is an effect and we're switching to a new context,
+        //               check culling before applying it. Effects (opacity, filters, blend modes) don't affect
+        //               clip state, so would_be_fully_clipped_by_painter() returns the same result before and after
+        //               applying effects.
+        //               This avoids expensive saveLayer/restore cycles for off-screen elements with effects like blur.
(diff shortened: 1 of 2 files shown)


Libraries/LibWeb/Painting/DisplayList.cpp
```diff
@@ -168,6 +168,21 @@ void DisplayListPlayer::execute_impl(DisplayList& display_list, ScrollStateSnaps
...
+
+        // OPTIMIZATION: If the leaf context is an effect and we're switching to a new context,
+        //               check culling before applying it. Effects (opacity, filters, blend modes) don't affect
+        //               clip state, so would_be_fully_clipped_by_painter() returns the same result before and after
+        //               applying effects.
+        //               This avoids expensive saveLayer/restore cycles for off-screen elements with effects like blur.
+        // NOTE: We must not do this for consecutive commands with the same context, as that would incorrectly restore
+        //       and re-apply the effect layer, breaking blend mode compositing.
+        if (context && applied_context != context && context->is_effect() && bounding_rect.has_value()) {
+            switch_to_context(context->parent());
```
gmta, line 181: Couldn't you unwind to the nearest context ancestor that isn't an effect here?

kalenikaliaksandr: nope, because "current" effect node might have clip/transform/scroll ancestors that were not applied by previous command.

gmta: Gotcha!

PR #7579, opened 2026-01-22
Meta: Add GDB pretty-printers for several String types

AtkinsSJ:
We have a bunch of basic types in AK that GDB doesn't do a great job at inspecting by default. So I thought as I've been encouraged to try Claude out, it was worth throwing it at the problem and seeing if we could get a nicer experience there. Answer: yes.

This PR just adds some scaffolding, and printing for a few String classes, but it shouldn't be too hard to extend it to support other things. Variant and Optional seem like good candidates especially. So this is mostly a proof of concept.

Some examples, taken from CLion:

image

image

image

image

This was substantially Claude's doing, and my Python experience is very limited, so there's a good chance some of this is goofy, unnecessary, or deeply irritating to people who actually write Python. So let me know. :sweat_smile:

### DIFF
--- a/Meta/gdb/PrettyPrinters/ak.py
+++ b/Meta/gdb/PrettyPrinters/ak.py
@@ -1,0 +1,275 @@
+"""
+GDB pretty-printers for AK types.
+"""
+
+import gdb
+import gdb.printing
+
+
+class GenericPrinter:
+    def __init__(self, val):
+        self.val = val
+
+    def yield_children(self, ignored_fields):
+        for field in self.val.type.fields():
+            if field.name in ignored_fields:
+                continue
+            if field.is_base_class:
+                yield field.name, self.val.cast(field.type)
+            elif field.name is None:
+                yield "(anonymous)", self.val[field]
+            else:
+                yield field.name, self.val[field]
+
+    def children(self):
+        try:
+            for child in self.yield_children(ignored_fields=[]):
+                yield child
+        except gdb.MemoryError:
+            yield "<error>", "invalid memory"
+        except Exception as e:
+            yield "<error>", str(e)
+
+
+def format_string_bytes(data):
+    """Format bytes as an escaped string for display"""
+    try:
+        text = data.decode("utf-8")
+        escaped = text.replace("\\", "\\\\").replace('"', '\\"')
+        escaped = escaped.replace("\n", "\\n").replace("\r", "\\r").replace("\t", "\\t")
+        return f'"{escaped}"'
+    except UnicodeDecodeError:
+        return f'"{data!r}"'
+
+
+class AKStringDataPrinter(GenericPrinter):
+    """Pretty-printer for AK::Detail::StringData"""
+
+    def __init__(self, val):
+        super().__init__(val)
+
+    def get_bytes(self):
+        """Extract the string bytes from StringData"""
+        byte_count = int(self.val["m_byte_count"])
+        is_substring = bool(self.val["m_substring"])
+
+        if byte_count == 0:
+            return b""
+
+        if is_substring:
+            substring_data_addr = self.val["m_bytes_or_substring_data"].address
+            substring_data_type = gdb.lookup_type("AK::Detail::StringData::SubstringData")
+            substring_data = substring_data_addr.cast(substring_data_type.pointer()).dereference()
+
+            superstring = substring_data["superstring"]
+            start_offset = int(substring_data["start_offset"])
+
+            super_byte_count = int(superstring["m_byte_count"])
+            super_is_substring = bool(superstring["m_substring"])
+
+            if super_is_substring:
+                return None
+
+            super_bytes_addr = superstring["m_bytes_or_substring_data"].address
+            inferior = gdb.selected_inferior()
+            all_bytes = inferior.read_memory(super_bytes_addr, super_byte_count)
+            return bytes(all_bytes)[start_offset : start_offset + byte_count]
+
+        bytes_addr = self.val["m_bytes_or_substring_data"].address
+        inferior = gdb.selected_inferior()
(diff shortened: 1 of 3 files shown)


spholz:
Note that we have GDB pretty printers in serenity: <https://github.com/SerenityOS/serenity/blob/master/Meta/serenity_gdb.py>.
They were removed here in dd1fbd35139e7429931b75a55e28da56db9dc942.

AtkinsSJ:
Thanks @spholz! I missed that they already existed before.

I've compared the two and the downside with the Serenity ones is they replace the original internals of what's being inspected. I wanted to make sure here that we add useful information without hiding anything. (Probably the user isn't debugging an issue in AK, but maybe they are and I don't want to obstruct that.) I'll definitely use it as a reference though!

AtkinsSJ:
Changes: Rebased, fiddled around a bit, and added an Optional printer.

I tried to get Variant working too, using the original Serenity one as a basis, but all I get is a bunch of errors from GDB. Same is the case when I use the Serenity one. Here's the diff for it:

```diff
diff --git a/Meta/gdb/PrettyPrinters/ak.py b/Meta/gdb/PrettyPrinters/ak.py
index 3da13d351bb..d2697fe9f24 100644
--- a/Meta/gdb/PrettyPrinters/ak.py
+++ b/Meta/gdb/PrettyPrinters/ak.py
@@ -282,6 +282,47 @@ class AKOptionalPrinter(GenericPrinter):
             return f"<error: {e}>"
 
 
+class AKVariantPrinter(GenericPrinter):
+    """Pretty-printer for AK::Variant"""
+
+    def __init__(self, val):
+        super().__init__(val)
+        self.index = int(self.val["m_index"])
+
+        self.contained_types = []
+        type_resolved = self.val.type.strip_typedefs()
+        index = 0
+        while True:
+            try:
+                self.contained_types.append(type_resolved.template_argument(index))
+                index += 1
+            except RuntimeError:
+                break
+
+    def _contained_value(self):
+        data = self.val["m_data"]
+        value_type = self.contained_types[self.index]
+        return data.cast(value_type.pointer()).referenced_value()
+
+    def to_string(self):
+        try:
+            return self._contained_value().format_string()
+        except gdb.MemoryError:
+            return "<invalid memory>"
+        except Exception as e:
+            return f"<error: {e}>"
+
+    def children(self):
+        try:
+            for child in super().yield_children(ignored_fields=[]):
+                yield child
+            yield "value", self._contained_value()
+        except gdb.MemoryError:
+            yield "<error>", "invalid memory"
+        except Exception as e:
+            yield "<error>", str(e)
+
+
 def build_pretty_printer():
     pp = gdb.printing.RegexpCollectionPrettyPrinter("AK")
     pp.add_printer("AK::String", "^AK::String$", AKStringPrinter)
@@ -290,6 +331,7 @@ def build_pretty_printer():
     pp.add_printer("AK::StringView", "^AK::StringView$", AKStringViewPrinter)
     pp.add_printer("AK::FlyString", "^AK::FlyString$", AKFlyStringPrinter)
     pp.add_printer("AK::Optional", "^AK::Optional<.*>$", AKOptionalPrinter)
+    pp.add_printer("AK::Variant", "^AK::Variant<.*>$", AKVariantPrinter)
     return pp
```

The complaint is on the `return data.cast(value_type.pointer()).referenced_value()` line - I get "Attempt to take address of value not located in memory." because apparently the `m_data` field isn't addressable. Same issue happens with the original Serenity printer. I tried fiddling around with it but got nowhere, this is over my head. :sweat_smile:

InvalidUsernameException:
I looked into getting pretty printers working a long time ago as well. My variant ended up as this, which all in all looks pretty similar to your version:

```py
class AKVariantPrinter(gdb.ValuePrinter):
    def __init__(self, val):
        self.__val = val

    def to_string(self):
        index = self.__val["m_index"]
        if index >= self.__val["invalid_index"]:
            return "<invalid>"
        data = self.__val["m_data"]
        if not data.address:
            # If there is no address, we are dealing with a temporary or optimized-out value.
            # Calling cast() on it below would throw.
            return "<optimized out>"

        contained_type = self.__val.type.template_argument(index).pointer()
        contained_value = data.cast(contained_type).referenced_value()
        return contained_value
```

I remember it working in some places, but not all of them. My assumption was that it might need an `-o0` build. Not entirely sure if that assessment and the code comment are correct though.

Here is the entire file of pretty-printers that I wrote: [gdb_pretty_printers.py](https://github.com/user-attachments/files/25297372/gdb_pretty_printers.py). With gdb severly memleaking and lldb crashing when loading WebContent, I've abandoned my plan to improve experience inside debuggers for the time being. So I haven't used or tested this in a while. But feel free to take these if they are useful.

Meta/gdb/PrettyPrinters/ak.py
```diff
@@ -0,0 +1,297 @@
+"""
+GDB pretty-printers for AK types.
+"""
+
+import gdb
+import gdb.printing
+
+
+class GenericPrinter:
```
jdahlin: Adding types would be nice and probably just a simple claude prompt

Meta/gdb/PrettyPrinters/ak.py
```diff
@@ -0,0 +1,297 @@
...
+        if is_substring:
+            substring_data_addr = self.val["m_bytes_or_substring_data"].address
+            substring_data_type = gdb.lookup_type("AK::Detail::StringData::SubstringData")
+            substring_data = substring_data_addr.cast(substring_data_type.pointer()).dereference()
+
+            superstring = substring_data["superstring"]
+            start_offset = int(substring_data["start_offset"])
+
+            super_byte_count = int(superstring["m_byte_count"])
+            super_is_substring = bool(superstring["m_substring"])
```
jdahlin: ditto

PR #7586, opened 2026-01-23
LibWeb/test-web: Test-web debug tools and fix navigable test timeouts

jonbgamble:
Added a few debug tools that helped me chase down a timeout that plagues CI Sanitizer runs (multiple-navigable-cross-document-navigation.html)

In high concurrency, the test often times out for honest reasons (it takes awhile to complete). I trimmed the number of blob urls and iframes from 404 down to a still absurd 84 so it will complete on time even when sharing an SMT core on Sanitizer builds.

Make the http-test-server /echo endpoint idempotent for identical echo rules (method+path+definition). This allows --repeat for tests like iframe-referrer.html while still returning 409 for conflicting definitions.

The following options are added to test-web:
  --debug-timeouts  Capture backtraces to stderr report on timeouts
  --repeat N  Repeat all matched tests N times
  --fail-fast  Abort on first failure/timeout/crash. If timeout, offer to lldb/gdb attach to the spinning WebContent (No Windows)

I found --repeat to be useful for flake hunts.

Here's what results look like for --repeat with --debug-timeouts:

image

### DIFF
--- a/Tests/LibWeb/test-web/Application.cpp
+++ b/Tests/LibWeb/test-web/Application.cpp
@@ -37,6 +37,7 @@ void Application::create_platform_arguments(Core::ArgsParser& args_parser)
     args_parser.add_option(python_executable_path, "Path to python3", "python-executable", 'P', "path");
     args_parser.add_option(dump_gc_graph, "Dump GC graph", "dump-gc-graph", 'G');
     args_parser.add_option(debug_timeouts, "Capture backtrace on timeouts (see test-dumps html -> Timeouts -> stderr)", "debug-timeouts");
+    args_parser.add_option(fail_fast, "Abort on first failure/timeout/crash (offers debugger attach on timeout)", "fail-fast");
 
     args_parser.add_option(repeat_count, "Repeat all matched tests N times", "repeat", 0, "n");
 
--- a/Tests/LibWeb/test-web/Application.h
+++ b/Tests/LibWeb/test-web/Application.h
@@ -40,6 +40,7 @@ public:
 
     bool dump_gc_graph { false };
     bool debug_timeouts { false };
+    bool fail_fast { false };
     size_t repeat_count { 1 };
     bool test_dry_run { false };
     bool rebaseline { false };
--- a/Tests/LibWeb/test-web/Debug.cpp
+++ b/Tests/LibWeb/test-web/Debug.cpp
@@ -42,6 +42,66 @@ ByteBuffer strip_sgr_sequences(StringView input)
     return output;
 }
 
+static bool stdin_and_stdout_are_ttys()
+{
+    auto stdin_is_tty_or_error = Core::System::isatty(0);
+    auto stdout_is_tty_or_error = Core::System::isatty(1);
+
+    return !stdin_is_tty_or_error.is_error() && stdin_is_tty_or_error.value()
+        && !stdout_is_tty_or_error.is_error() && stdout_is_tty_or_error.value();
+}
+
+static void maybe_attach_lldb_to_process(pid_t pid)
+{
+    if (pid <= 0)
+        return;
+
+    Vector<ByteString> arguments;
+    arguments.append("-p"sv);
+    arguments.append(ByteString::number(pid));
+
+    auto process_or_error = Core::Process::spawn({
+        .executable = "lldb"sv,
+        .search_for_executable_in_path = true,
+        .arguments = arguments,
+    });
+
+    if (process_or_error.is_error()) {
+        warnln("Failed to spawn lldb: {}", process_or_error.error());
+        return;
+    }
+
+    Core::Process lldb = process_or_error.release_value();
+    if (auto wait_result = lldb.wait_for_termination(); wait_result.is_error())
+        warnln("Failed waiting for lldb: {}", wait_result.error());
+}
+
+static void maybe_attach_gdb_to_process(pid_t pid)
+{
+    if (pid <= 0)
+        return;
+
+    Vector<ByteString> arguments;
+    arguments.append("-q"sv);
+    arguments.append("-p"sv);
+    arguments.append(ByteString::number(pid));
+
+    auto process_or_error = Core::Process::spawn({
+        .executable = "gdb"sv,
+        .search_for_executable_in_path = true,
+        .arguments = arguments,
+    });
+
+    if (process_or_error.is_error()) {
+        warnln("Failed to spawn gdb: {}", process_or_error.error());
+        return;
+    }
+
+    Core::Process gdb = process_or_error.release_value();
+    if (auto wait_result = gdb.wait_for_termination(); wait_result.is_error())
+        warnln("Failed waiting for gdb: {}", wait_result.error());
+}
+
 static void append_diagnostics_header(StringBuilder& builder, Test const& test, size_t view_id, StringView current_url)
 {
     auto& app = Application::the();
@@ -309,6 +369,44 @@ static void append_backtrace_for_process(StringBuilder& builder, StringView proc
     }
 
     builder.append("\n"sv);
+}
+
+void maybe_attach_on_fail_fast_timeout(pid_t pid)
+{
+    if (pid <= 0)
+        return;
+    if (!stdin_and_stdout_are_ttys())
+        return;
+
+    outln("Fail-fast timeout in WebContent pid {}.", pid);
+    outln("You may attach a debugger now (test-web will wait)."sv);
+    outln("- Press Enter to continue shutdown + exit"sv);
+    outln("- Type 'gdb' then Enter to attach with gdb first"sv);
+    outln("- Type 'lldb' then Enter to attach with lldb first"sv);
+    MUST(Core::System::write(1, "> "sv.bytes()));
+
+    auto standard_input_or_error = Core::File::standard_input();
+    if (standard_input_or_error.is_error())
+        return;
+
+    Array<u8, 64> input_buffer {};
+    auto buffered_standard_input_or_error = Core::InputBufferedFile::create(standard_input_or_error.release_value());
+    if (buffered_standard_input_or_error.is_error())
+        return;
+
+    auto& buffered_standard_input = buffered_standard_input_or_error.value();
+    auto response_or_error = buffered_standard_input->read_line(input_buffer);
+    if (response_or_error.is_error())
+        return;
+
+    ByteString response { response_or_error.value() };
+    response = response.trim_whitespace();
+    if (response.equals_ignoring_ascii_case("gdb"sv)) {
+        maybe_attach_gdb_to_process(pid);
+        return;
+    }
+    if (response.equals_ignoring_ascii_case("lldb"sv))
+        maybe_attach_lldb_to_process(pid);
 }
 
 void append_timeout_diagnostics_to_stderr(StringBuilder& stderr_builder, TestWebView& view, Test const& test, size_t view_id)
--- a/Tests/LibWeb/test-web/main.cpp
+++ b/Tests/LibWeb/test-web/main.cpp
@@ -57,6 +57,8 @@ static bool s_is_tty = false;
 
 static size_t s_current_run = 1;
 
+static bool s_fail_fast_triggered = false;
+
 static void update_terminal_size()
 {
 #ifndef AK_OS_WINDOWS
@@ -1424,6 +1426,7 @@ static ErrorOr<int> run_tests(Core::AnonymousBuffer const& theme, Web::DevicePix
     s_crashed_count = 0;
     s_skipped_count = 0;
     s_completed_tests = 0;
+    s_fail_fast_triggered = false;
 
     s_total_tests = tests.size();
     outln("Running {} tests...", tests.size());
@@ -1480,6 +1483,12 @@ static ErrorOr<int> run_tests(Core::AnonymousBuffer const& theme, Web::DevicePix
 
         // run_next_test handles: reset promise, attach callback, pick test, run test
         auto run_next_test = [&, view = view.ptr(), cleanup_test, view_id]() {
+            if (app.fail_fast && s_fail_fast_triggered) {
+                if (view_id < s_view_display_states.size())
+                    s_view_display_states[view_id].active = false;
+                return;
+            }
+
             // Check without incrementing first - only consume an index if we have a test
             if (current_test >= tests.size()) {
                 // Mark this view as idle (for variant wake-up tracking)
@@ -1515,7 +1524,7 @@ static ErrorOr<int> run_tests(Core::AnonymousBuffer const& theme, Web::DevicePix
 
             // Reset promise and attach completion callback
             view->reset_test_promise();
-            view->test_promise().when_resolved([&tests, &tests_remaining, &non_passing_tests, &app, view, cleanup_test, view_id, digits_for_view_id, digits_for_test_id](auto result) {
+            view->test_promise().when_resolved([&tests, &tests_remaining, &non_passing_tests, &app, view, cleanup_test, view_id, digits_for_view_id, digits_for_test_id, use_live_display](auto result) {
                 cleanup_test(result.test_index, result.result);
 
                 auto& test = tests[result.test_index];
@@ -1573,8 +1582,55 @@ static ErrorOr<int> run_tests(Core::AnonymousBuffer const& theme, Web::DevicePix
                 if (result.result != TestResult::Expanded)
                     ++s_completed_tests;
 
-                if (result.result != TestResult::Pass && result.result != TestResult::Expanded)
+                bool const is_non_passing_result = result.result != TestResult::Pass && result.result != TestResult::Expanded;
+                bool const should_trigger_fail_fast = result.result == TestResult::Fail || result.result == TestResult::Timeout || result.result == TestResult::Crashed;
+
+                if (is_non_passing_result)
                     non_passing_tests.append(result);
+
+                if (app.fail_fast && !s_fail_fast_triggered && should_trigger_fail_fast) {
+                    s_fail_fast_triggered = true;
+
+                    if (s_display_timer) {
+                        s_display_timer->stop();
+                        s_display_timer = nullptr;
+                    }
+
+                    if (use_live_display) {
+                        for (size_t i = 0; i < s_live_display_lines; ++i)
+                            out("\033[A\033[2K"sv);
+                        out("\r"sv);
+                        (void)fflush(stdout);
+                        s_live_display_lines = 0;
+                        print_deferred_warnings();
+                    }
+
+                    auto const pid = view->web_content_pid();
+                    if (result.result == TestResult::Timeout)
+                        outln("Fail-fast: Timeout: {} (pid {})", test.relative_path, pid);
+                    else
+                        outln("Fail-fast: {}: {}", test_result_to_string(result.result), test.relative_path);
+
+                    if (result.result == TestResult::Timeout) {
+                        auto& capture = output_capture_for_view(*view);
+                        StringBuilder backtrace_output;
+                        append_timeout_backtraces_to_stderr(backtrace_output, *view, test, view_id);
+                        auto backtrace_output_view = backtrace_output.string_view();
+                        capture.stderr_buffer.append(backtrace_output_view);
+
+                        if (app.verbosity >= Application::VERBOSITY_LEVEL_LOG_TEST_OUTPUT)
+                            (void)Core::System::write(STDERR_FILENO, backtrace_output_view.bytes());
+                    }
+
+                    if (s_all_tests_complete)
+                        s_all_tests_complete->reject(Error::from_string_literal("Fail-fast"));
+                    Core::EventLoop::current().quit(1);
+
+                    if (result.result == TestResult::Timeout)
+                        maybe_attach_on_fail_fast_timeout(pid);
+
+                    return;
+                }
 
                 if (--tests_remaining == 0) {
                     s_all_tests_complete->resolve({});


Tests/LibWeb/test-web/Debug.cpp
```diff
@@ -0,0 +1,402 @@
...
+        warnln("Failed to spawn lldb: {}", process_or_error.error());
+        return;
+    }
+
+    auto lldb = process_or_error.release_value();
+    if (auto wait_result = lldb.wait_for_termination(); wait_result.is_error())
+        warnln("Failed waiting for lldb: {}", wait_result.error());
+}
+
+#if defined(AK_OS_LINUX)
```
R-Goc: What about BSD? Gdb is also used there.

jonbgamble: I don't have a BSD system so I'm reluctant to target for it.

cqundefine: Is there even a point for locking this behind a platform macro, all Linux systems have GDB, so does FreeBSD, OpenBSD and NetBSD. Homebrew has GDB for MacOS and as far as I know there are GDB ports for Windows, although I'm not sure how useful they are in our case.

jonbgamble: Well, those other platforms might not be missing much. I don't yet have a strong sense whether attaching will be helpful.

As a `--fail-fast` side effect, it can be justified only if useful for debugging boundary issues between test-web and WebContent. But I suspect that in the majority of these cases WebContent will be sitting pretty, all innocent-like, in its idle loop. So attaching and backtraces may just end up gathering dust.

Libraries/LibWeb/HTML/Navigable.cpp
```diff
@@ -432,12 +450,31 @@ void Navigable::set_ongoing_navigation(Variant<Empty, Traversal, String> ongoing
...
+
     // 2. Inform the navigation API about aborting navigation given navigable.
     inform_the_navigation_api_about_aborting_navigation();
 
     // 3. Set navigable's ongoing navigation to newValue.
     m_ongoing_navigation = ongoing_navigation;
 
+    // AD-HOC: In test-web we may have queued BrowserUI navigations while a traversal was ongoing.
+    // Drain them once traversal finishes, but avoid polling or re-entrancy.
+    if (is_test_web_executable()
```
shannonbooth: Putting aside me not having a full understanding of the problem we are having here in navigation, I am personally not so keen on adding code into navigation which is only applicable to our test code.

Navigation is already one of the most difficult parts of the engine to understand and adding a difference only applicable to test is not going to help resolve that situation to find the many bugs we have there.

By the way, the most recently added Window::in_test_mode() can be used rather than adding `is_test_web_executable()`.

jonbgamble: Thanks for the heads up about Window::in_test_mode().

I see your point. Perhaps a change to the test itself could harden against pitfalls of rapid process reuse. But shall we continue to leave this test as a "must succeed" for CI when we cannot pass it reliably in Sanitizer runs?

shannonbooth: Yeah, I think flakey tests should be skipped

shannonbooth:
I tried taking an initial look, but there is a lot going on here, it would be great it this could be split into commits for standalone changes. Thinking of some of them might be:

- One for http test server change
- Another for the test change
- Introducing an option to repeat tests
- Fail fast
- etc

I'm not sure on the utility of all of these different options yet etc, will need to give them a go.

But I did notice a bug with dump-all-stderr causing a crash since there was a call to outln without using a format string

jonbgamble:
They're not meant to be generally useful. They are tools for whomever wants CI results to be repeatable.

jonbgamble:
> it would be great it this could be split into commits for standalone changes.

I agree and maybe I will find time do so, but I'm not super motivated because 1) I can keep a .patch of the test-web sources and reapply it on my end whenever needed, and 2) I have doubts they'll get used. Not many contributors read test-web --help after all.

jonbgamble:
> But I did notice a bug with dump-all-stderr causing a crash since there was a call to outln without using a format string

Note to self: Avoid frantic reshuffling prior to `push --force`.

For a time this flag was useful but it evolved into being largely redundant with `-v -v` for most purposes. FWIW, I was going for pipe friendly output for automated runs (without suppressing test-dumps output). But that's pretty niche, I'm going to axe this one.

Tests/LibWeb/test-web/Application.cpp
```diff
@@ -36,6 +36,22 @@ void Application::create_platform_arguments(Core::ArgsParser& args_parser)
...
+        .accept_value = [&](StringView value) -> ErrorOr<bool> {
+            auto maybe_n = value.to_number<size_t>();
+            if (!maybe_n.has_value() || maybe_n.value() == 0)
+                return false;
+            repeat_enabled = true;
+            repeat_count = maybe_n.value();
+            return true;
+        },
+    });
+
```
Zaggy1024, line 43: Instead of all this, can't we just parse it into a variable single variable, then if it differs from the default value, consider repeat to be enabled?

Tests/LibWeb/test-web/Application.h
```diff
@@ -23,7 +23,6 @@ class Application : public WebView::Application {
     virtual void create_platform_arguments(Core::ArgsParser&) override;
     virtual void create_platform_options(WebView::BrowserOptions&, WebView::RequestServerOptions&, WebView::WebContentOptions&) override;
     virtual bool should_capture_web_content_output() const override { return true; }
-
```
Zaggy1024, line 26: nit: Extraneous change

Zaggy1024:
A couple general comments before I get to a proper line by line review:
- With my limited knowledge of the http test server, it's not clear to me what the existing problem with it was and how it solves that. If someone else with a better understanding is able to review that commit before I get to it, maybe it'll be fine, but I think a clearer explanation may be worthwhile.
- I feel like instead of stripping control codes, we should try to turn them into HTML spans to color the text to match what we would have seen in the console. Not sure of the complexity of doing so, but this is something I was thinking of taking a stab at before.

Zaggy1024:
Oops, it submitted my pending review comments with that comment... oh well, they should hopefully be valid despite me not going through the whole thing yet.

jonbgamble:
> With my limited knowledge of the http test server, it's not clear to me what the existing problem with it was and how it solves that. If someone else with a better understanding is able to review that commit before I get to it, maybe it'll be fine, but I think a clearer explanation may be worthwhile.

There is at least one test that fails on subsequent runs during --repeat. The http server fix allows it to succeed.
 
> I feel like instead of stripping control codes, we should try to turn them into HTML spans to color the text to match what we would have seen in the console. Not sure of the complexity of doing so, but this is something I was thinking of taking a stab at before.

I personally feel color in logs should just be avoided when you cannot assume a rendering context.

Zaggy1024:
> There is at least one test that fails on subsequent runs during --repeat. The http server fix allows it to succeed.

Right, and commit messages are the most convenient place to document the reasons something was broken and how it was fixed. The current commit message doesn't make it clear, at least to me, what the before/after of this is, so it makes it difficult for me to evaluate the correctness.

> I personally feel color in logs should just be avoided when you cannot assume a rendering context.

I suppose that's fair considering that the output is stored in a .txt file, but the main way of viewing it is through an HTML document which can easily visualize those codes. I would argue that it's worth keeping them in order to improve readability, but for now, I'm fine with stripping them.

jonbgamble:
I will add detail to the commit message. I suppose it also makes sense to move that http server change to the same commit that adds --repeat so you don't have to chase down a pull request in order to link the two changes when looking at blame.

Tests/LibWeb/Fixtures/http-test-server.py
```diff
@@ -104,9 +120,34 @@ def do_POST(self):
             # Return 409: Conflict if the method+path combination already exists
             key = f"{echo.method} {echo.path}"
             if key in echo_store:
+                existing_echo = echo_store[key]
+
+                # Be idempotent if the definition is identical.
```
Zaggy1024:
```suggestion
                # Ignore the redefinition if it is identical to the previous one.
```
Perhaps this makes it a little clearer?

Tests/LibWeb/Fixtures/http-test-server.py
```diff
@@ -104,9 +120,34 @@ def do_POST(self):
...
+                if echo_definition(existing_echo) == echo_definition(echo):
```
Zaggy1024: it looks like we're sending a different response code for this case, is that intentional? It feels like we could just condition the 409 response on the `echo_definition()`s being different.

Zaggy1024:
> I suppose it also makes sense to move that http server change to the same commit that adds --repeat so you don't have to chase down a pull request in order to link the two changes when looking at blame.

Oh, no, we're very used to doing that. My rule of thumb is that if some bug that is otherwise independent needs to be fixed before a subsequent change can make sense, it should absolutely be its own commit. My issue here has nothing to do with it being separate, but rather that it was unclear what the problem was in the first place, though part of that comes from my lack of knowledge of that script in the first place. I would definitely keep it separate here.

Tests/LibWeb/Fixtures/http-test-server.py
jonbgamble: Indeed. I blame clipboard support in modern IDEs!

Tests/LibWeb/test-web/main.cpp
```diff
@@ -164,7 +166,10 @@ static void render_live_display()
         size_t total = s_total_tests;
 
         // Calculate progress bar width (leave room for "completed/total []")
-        auto counter = ByteString::formatted("{}/{} ", completed, total);
+        auto counter = ByteString::formatted("{}/{} tests ", completed, total);
+        if (Application::the().repeat_count > 1)
+            counter = ByteString::formatted("{}run {}/{} ", counter, s_current_run, Application::the().repeat_count);
```
Zaggy1024: I know this was a problem in the original code too, but now that we're doing it twice: this should probably use `appendff()` instead of allocating two separate `ByteStrings`.

Tests/LibWeb/test-web/main.cpp
```diff
@@ -1501,13 +1542,10 @@ static ErrorOr<int> run_tests(Core::AnonymousBuffer const& theme, Web::DevicePix
...
         for (auto& view : views) {
-            if (auto path = view->dump_gc_graph(); path.is_error()) {
+            if (auto path = view->dump_gc_graph(); path.is_error())
                 warnln("Failed to dump GC graph: {}", path.error());
-            } else {
+            else
                 outln("GC graph dumped to {}", path.value());
-                auto source_root = LexicalPath(app.test_root_path).parent().parent().string();
-                outln("GC graph explorer: file://{}/Meta/gc-heap-explorer.html?script=file://{}", source_root, path.value());
-            }
```
Zaggy1024, line 1634: Was this removal intended?

Tests/LibWeb/test-web/main.cpp
```diff
@@ -1358,9 +1396,8 @@ static ErrorOr<int> run_tests(Core::AnonymousBuffer const& theme, Web::DevicePix
                     s_view_display_states[view_id].start_time = test.start_time;
                 }
                 render_live_display();
-            } else if (app.verbosity >= Application::VERBOSITY_LEVEL_LOG_TEST_DURATION) {
-                outln("[{:{}}] {:{}}/{}:  Start {}", view_id, digits_for_view_id, test.index + 1, digits_for_test_id, tests.size(), test.relative_path);
```
Zaggy1024, line 1486: Was this removal intended?

Tests/LibWeb/test-web/main.cpp
```diff
@@ -1414,6 +1400,16 @@ static ErrorOr<int> run_tests(Core::AnonymousBuffer const& theme, Web::DevicePix
...
                 test.end_time = UnixDateTime::now();
 
+                if (result.result == TestResult::Timeout && app.debug_timeouts) {
+                    if (auto capture = s_output_captures.get(view); capture.has_value() && *capture) {
+                        append_timeout_diagnostics_to_stderr((**capture).stderr_buffer, *view, test, view_id);
+                    } else if (app.verbosity >= Application::VERBOSITY_LEVEL_LOG_TEST_OUTPUT) {
+                        StringBuilder diagnostics;
+                        append_timeout_diagnostics_to_stderr(diagnostics, *view, test, view_id);
+                        warnln("{}", diagnostics.string_view());
+                    }
```
Zaggy1024: I believe that at this point, any view that's run a test should already have its capture set up. We should probably make the value type into NonnullOwnPtr to make that clear, but this if statement is redundant.

Also, I think this should use the same method to tee the output that Tim's changes [here](https://github.com/LadybirdBrowser/ladybird/commit/ba5c254fdef7d88e1c0dcb74a362da16ffc4a153) do?

Tests/LibWeb/test-web/TestWebView.cpp
```diff
@@ -47,7 +47,14 @@ NonnullRefPtr<Core::Promise<RefPtr<Gfx::Bitmap const>>> TestWebView::take_screen
...
 {
-    VERIFY(m_pending_screenshot);
+    if (!m_pending_screenshot) {
+        static bool warned_about_stray_screenshot = false;
+        if (!warned_about_stray_screenshot) {
+            warned_about_stray_screenshot = true;
+            warnln("Ignoring screenshot response with no pending request");
+        }
+        return;
+    }
```
Zaggy1024: What's going on here?

Tests/LibWeb/test-web/TestWeb.h
```diff
@@ -43,7 +43,6 @@ enum class RefTestExpectationType {
 
 struct Test {
     TestMode mode;
-
```
Zaggy1024, line 47: nit: extraneous change

Tests/LibWeb/test-web/main.cpp
```diff
@@ -721,7 +723,6 @@ static void run_dump_test(TestWebView& view, TestRunContext& context, Test& test
 
         if (Application::the().verbosity >= Application::VERBOSITY_LEVEL_LOG_TEST_OUTPUT) {
             auto const color_output = s_is_tty ? Diff::ColorOutput::Yes : Diff::ColorOutput::No;
-
```
Zaggy1024, line 738: nit: extraneous change

Tests/LibWeb/test-web/main.cpp
```diff
@@ -1356,6 +1358,12 @@ static ErrorOr<int> run_tests(Core::AnonymousBuffer const& theme, Web::DevicePix
 
         // run_next_test handles: reset promise, attach callback, pick test, run test
         auto run_next_test = [&, view = view.ptr(), cleanup_test, view_id]() {
+            if (app.fail_fast && s_fail_fast_triggered) {
+                if (view_id < s_view_display_states.size())
+                    s_view_display_states[view_id].active = false;
```
Zaggy1024, line 1488: This bounds check bugs me, I feel like it should be impossible for a bounds check to fail on this vector, but I see that it's mirroring bounds checks elsewhere that also look unnecessary to me. Just commenting this for my own sanity but I'll resolve this comment myself.

Tests/LibWeb/test-web/main.cpp
```diff
@@ -1400,13 +1408,15 @@ static ErrorOr<int> run_tests(Core::AnonymousBuffer const& theme, Web::DevicePix
...
 
-                if (result.result == TestResult::Timeout && app.debug_timeouts) {
-                    if (auto capture = s_output_captures.get(view); capture.has_value() && *capture) {
-                        append_timeout_diagnostics_to_stderr((**capture).stderr_buffer, *view, test, view_id);
-                    } else if (app.verbosity >= Application::VERBOSITY_LEVEL_LOG_TEST_OUTPUT) {
-                        StringBuilder diagnostics;
-                        append_timeout_diagnostics_to_stderr(diagnostics, *view, test, view_id);
-                        warnln("{}", diagnostics.string_view());
+                if (result.result == TestResult::Timeout) {
+                    if (app.debug_timeouts && !s_fail_fast_triggered) {
```
Zaggy1024: If we're skipping output when fail-fast is triggered, why not just skip this entire promise resolution instead? I feel like we should either not dump anything from fails after the trigger, or dump everything. Arguably, more output is better, it can't possibly be a bad thing to get extra failures in my mind.

Tests/LibWeb/test-web/main.cpp
```diff
@@ -1449,6 +1459,44 @@ static ErrorOr<int> run_tests(Core::AnonymousBuffer const& theme, Web::DevicePix
                 if (result.result != TestResult::Pass && result.result != TestResult::Expanded)
                     non_passing_tests.append(result);
 
+                if (app.fail_fast && !s_fail_fast_triggered && (result.result == TestResult::Fail || result.result == TestResult::Timeout || result.result == TestResult::Crashed)) {
```
Zaggy1024: Merge the test result check with the above if statement? This seems redundant.

Tests/LibWeb/test-web/TestWebView.cpp
```diff
@@ -47,14 +47,7 @@ NonnullRefPtr<Core::Promise<RefPtr<Gfx::Bitmap const>>> TestWebView::take_screen
...
 {
-    if (!m_pending_screenshot) {
-        static bool warned_about_stray_screenshot = false;
-        if (!warned_about_stray_screenshot) {
-            warned_about_stray_screenshot = true;
-            warnln("Ignoring screenshot response with no pending request");
-        }
-        return;
-    }
+    VERIFY(m_pending_screenshot);
```
Zaggy1024, line 50: Oh I see, it got removed here. ⚠️

Tests/LibWeb/test-web/Application.cpp
```diff
@@ -37,6 +37,7 @@ void Application::create_platform_arguments(Core::ArgsParser& args_parser)
     args_parser.add_option(python_executable_path, "Path to python3", "python-executable", 'P', "path");
     args_parser.add_option(dump_gc_graph, "Dump GC graph", "dump-gc-graph", 'G');
     args_parser.add_option(debug_timeouts, "Capture backtrace on timeouts (see test-dumps html -> Timeouts -> stderr)", "debug-timeouts");
+    args_parser.add_option(fail_fast, "Abort on first failure/timeout/crash", "fail-fast");
```
Zaggy1024: Prehaps this should say something about attaching a debugger when a timeout occurs? Not sure how to fit it in here without punctuation though, so I'll leave this to your discretion.

Tests/LibWeb/test-web/TestWebView.cpp
jonbgamble: I have no idea.

Tests/LibWeb/test-web/main.cpp
jonbgamble: nope

Tests/LibWeb/test-web/main.cpp
jonbgamble: I vaguely want to say yes, but I cannot remember why. And I'm 99% sure it is wrong now and whatever prompted it is no longer in this code.

Tests/LibWeb/test-web/TestWebView.cpp
jonbgamble: yeah I dunno what that was.

Tests/LibWeb/test-web/main.cpp
jonbgamble: Maybe, but I couldn't guarantee there was not still some path where capture is not set up.

Yep, I restored the notifier callbacks to Tim's so output is still captured for artifacts and echoed at verbose level. This code actually predates Tim's changes, and has been rebased (poorly) many times.

NonnullOwnPtr sounds good - maybe in a separate PR? To keep this PR's scope limited.

Tests/LibWeb/Fixtures/http-test-server.py
```diff
@@ -203,7 +225,7 @@ def handle_echo(self):
             if echo.reflect_headers_in_body:
                 headers = defaultdict(list)
                 for key in self.headers.keys():
-                    headers[key] = self.headers.get_all(key)
+                    headers[key] = self.headers.get_all(key) or []
```
Zaggy1024: Is this change needed? We're iterating over keys here, so the fallback should never hit, no?

Tests/LibWeb/Fixtures/http-test-server.py
```diff
@@ -103,10 +119,16 @@ def do_POST(self):
 
             # Return 409: Conflict if the method+path combination already exists
             key = f"{echo.method} {echo.path}"
-            if key in echo_store:
+            if key in echo_store and echo_definition(echo_store[key]) != echo_definition(echo):
```
Zaggy1024: Instead of creating a tuple, could we overload equality operator so that this can be a bit more concise?

Tests/LibWeb/test-web/main.cpp
Zaggy1024: I'm fine with leaving it for another PR, but the everywhere else uses the OwnPtr as if it was always non-null, so let's stay consistent in that way.

Also, with regard to teeing the output, I meant that this should also always capture the diagnostic output to the stderr file _and_ output it conditionally on the verbosity.

Zaggy1024: I should also note that if we don't have an output capture for the view at this point, we have definitely made a mistake somewhere else, so we shouldn't even need to check the Optional from get().

Tests/LibWeb/test-web/main.cpp
```diff
@@ -211,6 +218,15 @@ struct ViewOutputCapture {
...
 static HashMap<WebView::ViewImplementation const*, OwnPtr<ViewOutputCapture>> s_output_captures;
 
+static ByteString iteration_scoped_relative_path(Test const& test)
+{
+    auto& app = Application::the();
+    if (app.repeat_count <= 1)
+        return test.safe_relative_path;
+
+    return LexicalPath::join(ByteString::formatted("run-{}", test.run_index), test.safe_relative_path).string();
+}
```
Zaggy1024: Would we be able to just set the safe_relative_path in the cloned Test instances to this deeper directory instead? Then we don't need to call this function in a multiple places, and we don't run the risk of forgetting to call it.

Tests/LibWeb/test-web/main.cpp
```diff
@@ -1631,9 +1676,9 @@ static ErrorOr<int> run_tests(Core::AnonymousBuffer const& theme, Web::DevicePix
 
     if (app.dump_gc_graph) {
         for (auto& view : views) {
-            if (auto path = view->dump_gc_graph(); path.is_error()) {
+            if (auto path = view->dump_gc_graph(); path.is_error())
                 warnln("Failed to dump GC graph: {}", path.error());
-            } else {
+            else {
```
Zaggy1024: Incorrect formatting change here, we keep all blocks explicit when any one of them requires it in an if statement chain.

Tests/LibWeb/test-web/Debug.cpp
```diff
@@ -39,4 +42,291 @@ ByteBuffer strip_sgr_sequences(StringView input)
...
+        state->finished = true;
+    });
+
+    timeout_timer->start();
+    Core::EventLoop::current().spin_until([state] {
+        return state->finished || state->timed_out;
+    });
+
+    // If we finished before the timeout fired, ensure the timer won't fire after we return.
+    timeout_timer->stop();
```
Zaggy1024: This is unnecessary, timers are automatically stopped via EventReceiver when they're destroyed.

Tests/LibWeb/test-web/Debug.cpp
```diff
@@ -39,4 +42,291 @@ ByteBuffer strip_sgr_sequences(StringView input)
...
+        builder.append(text->bytes_as_string_view());
+        if (!text->bytes_as_string_view().ends_with('\n'))
+            builder.append("\n"sv);
+    } else {
+        builder.append("(Timed out waiting for page info)\n"sv);
+    }
+    builder.append("\n"sv);
+}
+
+static ErrorOr<void> run_tool_and_append_output(StringBuilder& builder, StringView tool_name, Vector<ByteString> const& arguments, u32 timeout_ms)
```
Zaggy1024, line 158: General comment on this function: It feels like this should be implemented via a notifier in the event loop, similarly to how we capture output from WebContent and others. Hopefully, that would result in much cleaner code if it works.

Tests/LibWeb/test-web/TestWebView.cpp
```diff
@@ -50,7 +50,7 @@ void TestWebView::did_receive_screenshot(Badge<WebView::WebContentClient>, Gfx::
     VERIFY(m_pending_screenshot);
 
     auto pending_screenshot = move(m_pending_screenshot);
-    pending_screenshot->resolve(screenshot.bitmap());
+    pending_screenshot->resolve(move(screenshot.bitmap()));
```
Zaggy1024: Still an extraneous change here :^(

Tests/LibWeb/test-web/TestWebView.cpp
```diff
@@ -60,7 +60,7 @@ void TestWebView::on_test_complete(TestCompletion completion)
     m_pending_prompt_text.clear();
     client().async_set_viewport(m_client_state.page_index, viewport_size(), 1.0);
 
-    m_test_promise->resolve(move(completion));
+    m_test_promise->resolve(completion);
```
Zaggy1024: And another here

Tests/LibWeb/test-web/main.cpp
Zaggy1024: I actually had in mind putting the fail_fast logic within the if statement above, but I realize now that we need to ignore the Skipped result. It's fine as-is, but I feel like maybe there's something simpler we could do here.

Tests/LibWeb/test-web/main.cpp
Zaggy1024: Creating a new StringBuilder wasn't what I was thinking of, but I missed the part immediately following where it counts the string length. I made #8143 (which will conflict here) to measure the length of the counter string within the output StringBuilder instead.

Tests/LibWeb/Fixtures/http-test-server.py
jonbgamble: You're right. We don't need defaultdict either. Want it changed to `headers = {}` and remove the import?

Tests/LibWeb/test-web/TestWebView.cpp
jonbgamble: clangd no like

Tests/LibWeb/test-web/main.cpp
jonbgamble: Sounds good.

Tests/LibWeb/test-web/TestWebView.cpp
Zaggy1024: True, and I don't want to belabor this, I just prefer to avoid lumping changes into commits like this if the code isn't being touched otherwise. Nothing wrong with farming your commit count by making a separate commit for this :^)

It's fine though.

Tests/LibWeb/Fixtures/http-test-server.py
Zaggy1024: Ah, yeah, that seems fine to do as a drive-by.

Tests/LibWeb/test-web/Debug.cpp
jonbgamble: How about we keep our fingers crossed, and if the current approach blows up for anyone, we'll just insta-yank --debug-timeouts and debugger attach? I think --repeat and --fail-fast will remain generally useful without it.

Zaggy1024: Sounds fine, I'll change it in a subsequent PR.

PR #7587, opened 2026-01-23
LibJS: Dead code elimination for always truthy/falsey conditions

dosisod:
This improves and expands the ability to do dead code elimination on conditions which are always truthy or falsey.

The following cases are now optimized:
* `if (true){}` -> Only emit `if` block, ignore `else`
* `if (false){}` -> Only emit `else if`/`else` block
* `while (false){}` -> Ignore `while` loop entirely
* Ternary -> Directly return left/right hand side if condition is const


dosisod:
I might as well handle `do {} while (false)`. It looks like it already does some optimization on it (getting rid of the jumps), though we should just inline the whole basic block.

Some other cases to consider:

* `for (;false;){}`: The `for` can be totally eliminated here, though I've never seen this in the wild. Looking on [grep.app](https://grep.app/search?q=%28%3Bfalse%3B%29), it only seems to be used in test code :smile: . Maybe there's a situation where some auto-generated code gets const folded and it can be optimized away. I guess it doesn't hurt to add this.

* Convert `while (x) {}` -> `do {} while (x)` if we can assert the first evaluation of `x` is always true. This saves one condition, and allows you to move the condition to the bottom of the basic block. This should go in it's own PR though since it isn't dead code related.

* Eliminate dead code if we can determine control flow always exits early:
  * `if (true) { return }; dead()` -> `dead()` is never called
  * `while (x) { return }; dead()` -> `dead()` is never called
  * Switch cases which `return` on all paths

dosisod:
Test failure is due to single timeout test

PR #7590, opened 2026-01-23
LibWeb: Support weight & slope for generic font name fallback

gmta:
On macOS, we had the issue that simple test cases like `<b>foo</b>` or `<i>bar</i>` would show a regular font instead of bold or italic, respectively. This is because `Times` was used as the generic serif font fallback _before_ `Times New Roman`, and only the latter had variants for bold and italic included.

This updates `FontPlugin::generic_font_name()` to accept a weight and slope to find a more suitable font name. As a bonus, we can now also load individual TTC fonts if there is more than one present in the file.

I was unable to create a regression test, since our testing font SerenitySans does not include weight/slope variants.

### DIFF
--- a/Libraries/LibWebView/Plugins/FontPlugin.cpp
+++ b/Libraries/LibWebView/Plugins/FontPlugin.cpp
@@ -6,6 +6,7 @@
  */
 
 #include <AK/ByteString.h>
+#include <AK/HashTable.h>
 #include <AK/String.h>
 #include <AK/TypeCasts.h>
 #include <LibCore/Resource.h>
@@ -35,9 +36,7 @@ FontPlugin::FontPlugin(bool is_layout_test_mode, Gfx::SystemFontProvider* font_p
 
     update_generic_fonts();
 
-    m_default_font_name = generic_font_name(Web::Platform::GenericFont::UiSansSerif);
-
-    auto default_fixed_width_font_name = generic_font_name(Web::Platform::GenericFont::UiMonospace);
+    auto default_fixed_width_font_name = generic_font_name(Web::Platform::GenericFont::UiMonospace, 400, 0);
     m_default_fixed_width_font = Gfx::FontDatabase::the().get(default_fixed_width_font_name, 12.0, 400, Gfx::FontWidth::Normal, 0);
     VERIFY(m_default_fixed_width_font);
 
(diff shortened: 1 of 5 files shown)


Libraries/LibWebView/Plugins/FontPlugin.cpp
```diff
@@ -188,20 +192,91 @@ void FontPlugin::update_generic_fonts()
...
+        if (has_requested_weight)
+            score += 1000;
+
+        // Prefer families that have the exact requested slope.
+        if (has_requested_slope)
+            score += 100;
+
+        // As a tiebreaker, prefer families with more weight variety.
+        // This helps select fonts that can handle both regular and bold text.
+        score += static_cast<int>(available_weights.size());
```
konradekk: something like this?

```suggestion
        // Strongly prefer families that have the exact requested weight.
        if (has_requested_weight)
            score += 1000;

        // Prefer families that have the exact requested slope.
        if (has_requested_slope)
            score += 100;

        // As a tiebreaker, prefer families with more weight variety.
        // This helps select fonts that can handle both regular and bold text.
        if (!available_weights.is_empty())
            score += static_cast<int>(available_weights.size());
```

gmta: If `available_weights.is_empty()` then `available_weights.size() == 0` so we can simply always add the size. I like that we don't unnecessarily cull out a font if eventually it turned out to be the only font available.

Libraries/LibWebView/Plugins/FontPlugin.cpp
```diff
@@ -188,20 +192,88 @@ void FontPlugin::update_generic_fonts()
...
+        bool has_requested_weight = false;
+        bool has_requested_slope = false;
+        HashTable<u16> available_weights;
+
+        Gfx::FontDatabase::the().for_each_typeface_with_family_name(family, [&](Gfx::Typeface const& typeface) {
+            available_weights.set(typeface.weight());
+            if (typeface.weight() == static_cast<u16>(weight))
+                has_requested_weight = true;
+            if (typeface.slope() == static_cast<u8>(slope))
+                has_requested_slope = true;
```
konradekk, line 255: one more idea: note that there are `0`, `1`, `2` values; `0` denoting upright, `1` being true italic, while `2` is oblique style – maybe some points for non-upright match (e.g. 10)?

i’m also wondering whether this `slope` is not an abstraction leak in a way (i.e. shouldn’t this be a nice enumeration on gfx side)?

Libraries/LibWebView/Plugins/FontPlugin.cpp
```diff
@@ -116,6 +144,12 @@ static Optional<String> query_fontconfig_for_generic_family(Web::Platform::Gener
 
     FcDefaultSubstitute(pattern);
 
+    // Request the specific weight and slope from fontconfig.
+    success = FcPatternAddInteger(pattern, FC_WEIGHT, css_weight_to_fontconfig_weight(weight));
+    VERIFY(success);
+    success = FcPatternAddInteger(pattern, FC_SLANT, css_slope_to_fontconfig_slant(slope));
+    VERIFY(success);
```
tcl3, line 151: I think this should go before `FcConfigSubstitute()`. Adding these after the substitutions have taken place will add them with a lower priority than the defaults.

gmta:
Closing this in favor of #7624

PR #7598, opened 2026-01-23
LibURL: Optimize parsing of URLs in authority state and misc tidy ups

shannonbooth:
While the URL parser is very unoptimized and will need future performance
work, this case in particular is the most egregious performance issue
currently in the parser.

Previously the authority state was parsed character-by-character in a
loop, appending each character to a buffer. When an '@' character was
encountered, the entire buffer would be re-processed to extract and
percent-encode the username and password portions.

This created O(n^2) behavior for URLs with multiple '@' characters, as
each '@' would trigger reprocessing of all previously buffered content.

This commit changes the authority state parser to process the authority
section in chunks rather than character-by-character:

1. Find the next delimiter ('@', '/', '?', '#', or '\\' for special URLs)
2. Process the entire chunk up to that delimiter at once
3. Directly extract username/password from the chunk without buffering

With an additional change of switching to:
iterator_at_byte_offset_without_validation (which does not have any
loops), this reduces the time complexity to O(n) and the included
test (found by fuzzing) to actually complete parsing :^)

### DIFF
--- a/Libraries/LibURL/CMakeLists.txt
+++ b/Libraries/LibURL/CMakeLists.txt
@@ -20,4 +20,4 @@ set(SOURCES
 )
 
 ladybird_lib(LibURL url)
-target_link_libraries(LibURL PRIVATE LibUnicode LibTextCodec LibRegex LibCrypto)
+target_link_libraries(LibURL PRIVATE LibUnicode LibTextCodec LibRegex)


shannonbooth:
I reworked this a bit and now the change is more minimal and I think easier to follow

Issue #7574, opened 2026-01-22
Frequent request timeouts


Randl:
### Summary

When I try to open almost any page, I get a lot of timeouts.

I am on a university network, so there may be something related to that,t but there are no problems in other browsers. There is no additional info in logs

### Operating system

Linux

### Steps to reproduce

1. Try to open some page

### Expected behavior

page loads

### Actual behavior

timeout

### URL for a reduced test case

n/a

### HTML/SVG/etc. source for a reduced test case

```html
n/a
```

### Log output and (if possible) backtrace

```shell
5177190.723 WebContent(2217935): ResourceLoader: Failed load of: "https://codepen.io/evgeniizh/pen/MYYzwVV", Error: Request finished with error: Timeout reached, Duration: 90011ms
5177190.729 WebContent(2217935): FIXME: Load html page with an error if read of body failed.
 5177330.546 WebContent(2219190): ResourceLoader: Failed load of: "https://cdn.jsdelivr.net/npm/dompurify@2.3.5/dist/purify.min.js", Error: Request finished with error: Timeout reached, Duration: 90014ms
5177330.547 WebContent(2219190): HTMLScriptElement: Refusing to run script because the element's result is null.
5177437.247 WebContent(2219190): ResourceLoader: Failed load of: "https://cdn.jsdelivr.net/npm/bootstrap@5.3.0/dist/css/bootstrap.min.css", Error: Request finished with error: Timeout reached, Duration: 90029ms
5177437.251 WebContent(2219190): ResourceLoader: Failed load of: "https://cdn.jsdelivr.net/npm/bootstrap@5.3.0/dist/js/bootstrap.bundle.min.js", Error: Request finished with error: Timeout reached, Duration: 90032ms
5177437.251 WebContent(2219190): HTMLScriptElement: Refusing to run script because the element's result is null.
5177437.787 WebContent(2219190): (js log) "Test theme:null"
5177583.484 WebContent(2221281): ResourceLoader: Failed load of: "https://aistudio.google.com/welcome", Error: Request finished with error: Timeout reached, Duration: 90032ms
5177583.491 WebContent(2221281): FIXME: Load html page with an error if read of body failed.
```

### Contribute a patch?

- [ ] I’ll contribute a patch for this myself.

anterpin:
It happens to me as well with `https://www.google.com`
After some research, it turns out to be caused by the combination of two factors:
[...]

PR #7600, opened 2026-01-23
RequestServer: Set HTTP/3 as default version with fallback to earlier

rcorsi:
Related to #7574 but doesn't directly fix it.

Still seems like a good change for certain network/router/firewall situations as those discussed in #7574 issue.


rcorsi:
Does not fix the problem.

PR #7601, opened 2026-01-23
Everywhere: Name all threads in pthreads and use the names in dbgln

Zaggy1024:
All Threading::Thread instances must now be created with a name. All the existing threads have been named with a strings that can fit into the pthreads 16-byte null-terminated string so as to avoid truncation when calling `pthread_setname_np`.

Debug logging has been changed to make use of these names, the new format looks like:

> 94.833 **WebContent(81989)**: Spawning the thread named Foo.
> 94.833 **WebContent(81989) Foo(81990)**: Thread message!

For threads without names, it will read like:

> 94.833 **WebContent(81989)**: Spawning an unnamed thread.
> 94.833 **WebContent(81989) Thread(81990)**: Thread message!

The formatting also now uses `MonotonicTime::now_coarse()` instead of `clock_gettime()` so that the same code can be used on Windows. On macOS, logs will now show the thread ID on non-main threads.

The ANSI code for formatting the process/thread text is now the same across platforms, since it seems the Windows command line prompt supports the code that we were using on all other platforms anyway.

### DIFF
--- a/AK/Format.cpp
+++ b/AK/Format.cpp
@@ -16,7 +16,9 @@
 #include <AK/String.h>
 #include <AK/StringBuilder.h>
 #include <AK/StringFloatingPointConversions.h>
+#include <AK/Time.h>
 #include <math.h>
+#include <pthread.h>
 #include <stdio.h>
 #include <string.h>
 #include <time.h>
@@ -1296,11 +1298,36 @@ void set_rich_debug_enabled(bool value)
     is_rich_debug_enabled = value;
 }
 
+#define DEFAULT_FORMAT "\033[0m"
+#define BOLD_YELLOW_FORMAT "\033[33;1m"
+
+static auto current_process_id()
+{
+#if defined(AK_OS_WINDOWS)
+    return GetCurrentProcessId();
+#else
+    return getpid();
+#endif
+}
+
+static auto current_thread_id()
+{
+#if defined(AK_OS_LINUX)
+    return gettid();
+#elif defined(AK_OS_WINDOWS)
+    return GetCurrentThreadId();
+#elif defined(AK_OS_MACOS)
+    u64 thread_id { 0 };
+    pthread_threadid_np(nullptr, &thread_id);
+    return thread_id;
+#else
+    return Empty();
+#endif
+}
+
+static auto s_main_thread_id = current_thread_id();
+
 #ifdef AK_OS_WINDOWS
-
-#    define YELLOW(str) "\33[93m" str "\33[0m"
-
-static int main_thread_id = GetCurrentThreadId();
 
 static int initialize_console_settings()
 {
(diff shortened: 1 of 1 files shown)


Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
```diff
@@ -309,7 +309,7 @@ ErrorOr<NonnullRefPtr<PlaybackStream>> PlaybackStreamWASAPI::create(OutputState
     if (initial_output_state == OutputState::Playing)
         state->playing = true;
 
-    auto audio_thread = Threading::Thread::construct([state] {
+    auto audio_thread = Threading::Thread::construct("Audio Control"sv, [state] {
```
R-Goc: That's the render thread not the control thread (we don't have one). Though for commonality I guess it's fine.

Libraries/LibThreading/WorkerThread.h
```diff
@@ -1,172 +0,0 @@
...
-#include <AK/Debug.h>
-#include <AK/Variant.h>
-#include <LibThreading/ConditionVariable.h>
-#include <LibThreading/Mutex.h>
-#include <LibThreading/Thread.h>
-
-namespace Threading {
-
-// Macro to allow single-line logging prints with fields that only exist in debug mode.
-#if WORKER_THREAD_DEBUG
```
cqundefine, line 18: This debug macro should probably also be removed

Libraries/LibThreading/Thread.cpp
```diff
@@ -77,6 +77,16 @@ void Thread::start()
         [](void* arg) -> void* {
             auto self = adopt_ref(*static_cast<Thread*>(arg));
 
+            {
+                // pthread_setname_np expects a 16-byte or shorter null-terminated C string.
+                auto thread_name = self->thread_name().substring(0, min(self->thread_name().length(), 15));
+#if defined(AK_OS_MACOS)
+                pthread_setname_np(thread_name.characters());
+#else
```
cqundefine, line 89: If you don't want to deal with BSD support in this PR I can fix it up later, but either you could apply this:

```suggestion
#elif defined(AK_OS_OPENBSD)
                pthread_set_name_np(pthread_self(), thread_name.characters());
#elif defined(AK_OS_NETBSD)
                pthread_setname_np(pthread_self(), "%s", (void *)thread_name.characters());
#else
```

or just hide it behind something like:

```suggestion
#elif defined(AK_OS_LINUX) || defined(AK_OS_FREEBSD) || defined(AK_OS_WINDOWS)
```

and maybe throw a dbgln about it being unsupported although that might be quite verbose.

Libraries/LibMedia/Audio/PlaybackStreamWasapi.cpp
Zaggy1024: I've changed it to Audio Render, being more accurate doesn't hurt.

Libraries/LibThreading/Thread.cpp
Zaggy1024: I made the NetBSD code use a `const_cast<char*>(thread_name.characters()`, might be worth a check to see if this builds if you or anyone else is able.

PR #7131, opened 2025-12-14
Ladybird: Register signal handlers for graceful shutdown

mattt-thompson:
I'm fed up of being hit with "Ladybird PID file '/run/user/1000/Ladybird.pid' exists with PID 12254, but process cannot be found" if i CTRL + C on the CLI to close the browser and then launch it back up, which I often do as the options to maximize/minimize/close the application in the top right does not show on Linux Mint if the application is maximized, and with the extension rid of titles of applications, so I think this is worthy of its own PR.

PR #7603, opened 2026-01-24
LibCore: Avoid crash if notifier not registered if terminated by CTRL-C

### LINKED ISSUE #7602: VERIFICATION FAILED in EventLoopImplementationUnix.cpp during unregister_notifier
While testing #7574, once the fix is in place from #7600 the following crash would occur if CTRL-C is pressed to stop Ladybird.  Stopping Ladybird normally does not cause the VERIFICATION FAILED.

`./Meta/ladybird.py run Ladybird google.com`

After connecting to google.com successfully, and HTTP/3 protocol had to fallback in CURL due to firewall blocked UDP/QUIC traffic, and CTRL-C is pressed to stop Ladybird.

```
^CVERIFICATION FAILED: m_has_value at /home/bobo/projects/myladybird/AK/Optional.h:359
bobo@bobo-VirtualBox:~/projects/myladybird$ Stack trace (most recent call first):
#0  (inlined)          in AK::Optional<unsigned long>::release_value() at /home/bobo/projects/myladybird/AK/Optional.h:359:9
#1  0x0000733d24b85b9b in Core::EventLoopManagerUnix::unregister_notifier(Core::Notifier&) at /home/bobo/projects/myladybird/Libraries/LibCore/EventLoopImplementationUnix.cpp:661:87
#2  (inlined)          in Core::Notifier::set_enabled(bool) at /home/bobo/projects/myladybird/Libraries/LibCore/Notifier.cpp:35:45
#3  (inlined)          in Core::Notifier::~Notifier() at /home/bobo/projects/myladybird/Libraries/LibCore/Notifier.cpp:22:16
#4  0x0000733d24b718b5 in Core::Notifier::~Notifier() at /home/bobo/projects/myladybird/Libraries/LibCore/Notifier.cpp:23:1
#5  (inlined)          in AK::AtomicRefCounted<Core::EventReceiver>::unref() const at /home/bobo/projects/myladybird/AK/AtomicRefCounted.h:76:13
#6  (inlined)          in void AK::unref_if_not_null<Core::Notifier>(Core::Notifier*) at /home/bobo/projects/myladybird/AK/NonnullRefPtr.h:32:19
#7  (inlined)          in AK::RefPtr<Core::Notifier>::clear() at /home/bobo/projects/myladybird/AK/RefPtr.h:223:26
#8  0x0000733d24bc2395 in IPC::TransportSocket::~TransportSocket() at /home/bobo/projects/myladybird/Libraries/LibIPC/TransportSocket.cpp:152:31
#9  (inlined)          in AK::NonnullOwnPtr<IPC::TransportSocket>::clear() at /home/bobo/projects/myladybird/AK/NonnullOwnPtr.h:133:9
#10 (inlined)          in AK::NonnullOwnPtr<IPC::TransportSocket>::~NonnullOwnPtr() at /home/bobo/projects/myladybird/AK/NonnullOwnPtr.h:50:14
#11 0x0000733d24bb358c in IPC::ConnectionBase::~ConnectionBase() at /home/bobo/projects/myladybird/Libraries/LibIPC/Connection.cpp:29:1
#12 0x00005a73bf88369c in RequestServer::ConnectionFromClient::~ConnectionFromClient() at /home/bobo/projects/myladybird/Services/RequestServer/ConnectionFromClient.cpp:64:1
#13 (inlined)          in AK::AtomicRefCounted<Core::EventReceiver>::unref() const at /home/bobo/projects/myladybird/AK/AtomicRefCounted.h:76:13
#14 (inlined)          in void AK::unref_if_not_null<RequestServer::ConnectionFromClient>(RequestServer::ConnectionFromClient*) at /home/bobo/projects/myladybird/AK/NonnullRefPtr.h:32:19
#15 (inlined)          in AK::RefPtr<RequestServer::ConnectionFromClient>::clear() at /home/bobo/projects/myladybird/AK/RefPtr.h:223:26
#16 (inlined)          in AK::RefPtr<RequestServer::ConnectionFromClient>::~RefPtr() at /home/bobo/projects/myladybird/AK/RefPtr.h:103:14
#17 (inlined)          in AK::HashMap<int, AK::RefPtr<RequestServer::ConnectionFromClient>, AK::Traits<int>, AK::Traits<AK::RefPtr<RequestServer::ConnectionFromClient> >, false>::Entry::~Entry() at /home/bobo/projects/myladybird/AK/HashMap.h:23:12
#18 (inlined)          in AK::HashTable<AK::HashMap<int, AK::RefPtr<RequestServer::ConnectionFromClient>, AK::Traits<int>, AK::Traits<AK::RefPtr<RequestServer::ConnectionFromClient> >, false>::Entry, AK::HashMap<int, AK::RefPtr<RequestServer::ConnectionFromClient>, AK::Traits<int>, AK::Traits<AK::RefPtr<RequestServer::ConnectionFromClient> >, false>::EntryTraits, false>::~HashTable() at /home/bobo/projects/myladybird/AK/HashTable.h:188:44
#19 (inlined)          in AK::HashTable<AK::HashMap<int, AK::RefPtr<RequestServer::ConnectionFromClient>, AK::Traits<int>, AK::Traits<AK::RefPtr<RequestServer::ConnectionFromClient> >, false>::Entry, AK::HashMap<int, AK::RefPtr<RequestServer::ConnectionFromClient>, AK::Traits<int>, AK::Traits<AK::RefPtr<RequestServer::ConnectionFromClient> >, false>::EntryTraits, false>::~HashTable() at /home/bobo/projects/myladybird/AK/HashTable.h:180:5
#20 0x00005a73bf886676 in AK::HashMap<int, AK::RefPtr<RequestServer::ConnectionFromClient>, AK::Traits<int>, AK::Traits<AK::RefPtr<RequestServer::ConnectionFromClient> >, false>::~HashMap() at /home/bobo/projects/myladybird/AK/HashMap.h:21:7
#21 0x0000733d23a48270 in __run_exit_handlers at ./stdlib/exit.c:118:8
#22 0x0000733d23a4834d in __GI_exit at ./stdlib/exit.c:148:3
#23 0x0000733d23a2a57e in __libc_start_call_main at ./csu/../sysdeps/nptl/libc_start_call_main.h:74:3
#24 0x0000733d23a2a63a in __libc_start_main_impl at ./csu/../csu/libc-start.c:360:3
#25 0x00005a73bf879f54 in _start at /home/bobo/projects/myladybird/Build/release/libexec/RequestServer
```

rcorsi:
In situations where some HTTP/3 traffic might be blocked by firewalls as in issue #7574, notifiers might not be registered as expected.

Found will testing #7600.

Fixes #7602

### DIFF
--- a/Libraries/LibCore/EventLoopImplementationUnix.cpp
+++ b/Libraries/LibCore/EventLoopImplementationUnix.cpp
@@ -638,8 +638,9 @@ void EventLoopManagerUnix::unregister_timer(intptr_t timer_id)
 
 void EventLoopManagerUnix::register_notifier(Notifier& notifier)
 {
+    Threading::RWLockLocker<Threading::LockMode::Read> locker(s_thread_data_lock);
     auto& thread_data = ThreadData::the();
-    Threading::MutexLocker locker(thread_data.mutex);
+    Threading::MutexLocker thread_data_content_locker(thread_data.mutex);
 
     thread_data.notifier_to_index.set(&notifier, thread_data.poll_fds.size());
     thread_data.notifiers.append(&notifier);


Libraries/LibCore/EventLoopImplementationUnix.cpp
```diff
@@ -658,6 +658,9 @@ void EventLoopManagerUnix::unregister_notifier(Notifier& notifier)
         return;
     Threading::MutexLocker thread_data_content_locker(thread_data->mutex);
 
+    if (!thread_data->notifier_to_index.contains(&notifier))
+        return;
```
ADKaster: how did we end up in a state where the notifier to index map doesn't contain the notifier in question? that sounds like we're unregistering the notifier from a different thread than it was registered for, or after that thread's thread_local data is destroyed.

If this is in fact the correct patch, doing two lookups is unreasonable when we could just
```
auto maybe_notifier = thread_data->notifier_to_index.take(&notifier);
if (!maybe_notifier.has_value())
    return;
auto notifier_index = maybe_notifier.release_value();
```

rcorsi: I think your patch version is much better.

I'll see if I can understand why the notifier is missing exactly.

rcorsi:
In my test I just connect to google.com

`./Meta/ladybird.py run Ladybird google.com >run.log 2>&1`

Firewall is setup blocking port 443 UDP outgoing.

I modified the code to show the registration and unregistration and also when the notifier is not present.

```diff
diff --git a/Libraries/LibCore/EventLoopImplementationUnix.cpp b/Libraries/LibCore/EventLoopImplementationUnix.cpp
index f75e6fb6e9..512cbb83f8 100644
--- a/Libraries/LibCore/EventLoopImplementationUnix.cpp
+++ b/Libraries/LibCore/EventLoopImplementationUnix.cpp
@@ -638,6 +638,8 @@ void EventLoopManagerUnix::unregister_timer(intptr_t timer_id)
 
 void EventLoopManagerUnix::register_notifier(Notifier& notifier)
 {
+    dbgln("REGISTERING: {}", &notifier);
+
     auto& thread_data = ThreadData::the();
     Threading::MutexLocker locker(thread_data.mutex);
 
@@ -652,6 +654,8 @@ void EventLoopManagerUnix::register_notifier(Notifier& notifier)
 
 void EventLoopManagerUnix::unregister_notifier(Notifier& notifier)
 {
+    dbgln("UNREGISTERING: {}", &notifier);
+
     Threading::RWLockLocker<Threading::LockMode::Read> locker(s_thread_data_lock);
     auto* thread_data = ThreadData::for_thread(notifier.owner_thread());
     if (!thread_data)
@@ -659,8 +663,10 @@ void EventLoopManagerUnix::unregister_notifier(Notifier& notifier)
     Threading::MutexLocker thread_data_content_locker(thread_data->mutex);
 
     auto maybe_notifier = thread_data->notifier_to_index.take(&notifier);
-    if (!maybe_notifier.has_value())
+    if (!maybe_notifier.has_value()) {
+        dbgln("NOT PROPERLY REGISTERED: {}", &notifier);
         return;
+    }
     auto notifier_index = maybe_notifier.release_value();
 
     if (notifier_index + 1 < thread_data->poll_fds.size()) {

```

I might wait a few seconds or wait some time before I hit CTRL-C, and the number of notifiers that can't be unregistered vary between 0 and 6, usually an even number.

I tried adding a dump_backtrace() during the registering to see where it was coming from and then the problem disappeared.

rcorsi:
@ADKaster what about the #7131 solution? I will give that a try to see if it prevents the issue.

rcorsi:
Hmmm #7131 doesn't help.

Libraries/LibCore/EventLoopImplementationUnix.cpp
R-Goc: How could maybe_notifier not have value? The only two things which touch that HashMap are register and unregister notifier. The first one always sets it to a value, and the second one shoud always take it no? It would be nice to have two implementations of EventLoop on linux/macos to check if bugs are inside EvenLoop or somewhere else. Anyone up to write an epoll or kqueue one?

rcorsi: A lock is missing in the register.  Still testing, but that is what it looks like.

R-Goc: But there is a lock there. The rwlock is not needed as it's just for finding the thread data for a thread. Which we only register for our own thread.

rcorsi: I've updated the code. It solves the problem. No more issues.

Why its needed?  If two or more registers come at similar time, especially at start, could the s_thread_data HashMap get messed up.  

ThreadData& the() on line 224 especially lines 226-227 and the 'new' on line 230. Could we lose (and leak) some data there depending which pointer is returned.

the() code is not very safe.

rcorsi: You have same lock situation in register_timer.   Register has one lock, unregister has two.  Why do we have two in one and one in the other.

I put a Read lock in the register, but I guess it should be a Write lock.

I guess I need to run it with the thread sanitizer to be sure what is getting corrupted or lost.

ADKaster:
The original issue (#7602) shows a backtrace where we're destroying a global HashMap of `RequestServer::ConnectionFromClient` RefPtrs. Doing this is... inherently unsafe. The issue probably cropped up because something about your other patch, or your environment changed global destructor registration order, causing some good ol static init order fiasco goodness.

Destroying that hash map is dangerous because as shown in the backtrace, destroying those clients destroys their IPC connections, which unregisters some notifiers. Unregistering notifiers checks global (thread) state in EventLoopImplementationUnix. We've got cross-global dependencies, which is bad juju.

I suspect that any of

1) manually clearing out that hashmap on event loop exit (we don't have such a mechanism) 
2) wrapping the s_connections global map in AK::NeverDestroyed
3) changing the scoping on "the list of all connections" such that its lifetime is strictly shorter than the Core::EventLoop on the stack in RequestServer/main.cpp 

would solve the root cause

(p.s. I know at least one thakis that would probably vote for "less globals")

shannonbooth:
related commit in https://github.com/LadybirdBrowser/ladybird/pull/7582

rcorsi:
Yeah, I see that the write lock is present in the() function on `Libraries/LibCore/EventLoopImplementationUnix.cpp:234`. So this fix is not correct.

Also #7582 was merged, and this problem is gone. Can't reproduce it.

PR #7611, opened 2026-01-24
LibWeb/WebAudio: Implement basic offline rendering loop

Prospero23:
This PR introduces a basic offline rendering loop to process audio for the WebAudio API, based on the work of @jonbgamble.

## Description
We cannot access GC allocated objects off the main thread, so this PR creates a render thread friendly snapshot of the audio graph using `snapshot_render_graph()`. This snapshot is represented as a `RenderGraphDescription` that is then consumed by `RenderGraph` / `RenderGraphImpl` to construct and process a graph of `RenderNodes` on the rendering thread via `OfflineAudioRenderThread`.  

Follows the [Web Audio API rendering loop specification](https://webaudio.github.io/web-audio-api/#rendering-loop) and adds relevant spec steps as needed.

Only implements processing for `ConstantSourceNode` to keep the scope minimal.

Does not handle AudioParam processing/automation, suspend/resume, mixing, or graph updates.

## Testing
- `TestWebAudioRenderGraph.cpp` to test basic quantum rendering.
- `TestOfflineAudioRenderThread.cpp` to test basic helper thread API.
- 32 passing WPT to test basic rendering behavior of `ConstantSourceNodes` and currentTime incrementing.

### DIFF
--- a/Libraries/LibWeb/WebAudio/BaseAudioContext.h
+++ b/Libraries/LibWeb/WebAudio/BaseAudioContext.h
@@ -52,7 +52,16 @@ public:
 
     GC::Ref<AudioDestinationNode> destination() const { return *m_destination; }
     float sample_rate() const { return m_sample_rate; }
-    double current_time() const { return m_current_time; }
+    // https://webaudio.github.io/web-audio-api/#dom-baseaudiocontext-currenttime
+    // [from-spec] currentTime MUST be read atomically on the control thread.
+    // We store [[current frame]] atomically and derive currentTime from it.
+    double current_time() const
+    {
+        auto frame = m_current_frame.load(AK::MemoryOrder::memory_order_acquire);
+        if (m_sample_rate <= 0)
+            return 0.0;
+        return static_cast<double>(frame) / static_cast<double>(m_sample_rate);
+    }
     GC::Ref<AudioListener> listener() const { return m_listener; }
     Bindings::AudioContextState state() const { return m_control_thread_state; }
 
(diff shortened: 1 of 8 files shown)


Libraries/LibWeb/WebAudio/OfflineAudioRenderThread.cpp
```diff
@@ -0,0 +1,155 @@
...
+    graph.begin_quantum(current_frame);
+    AudioBus const& destination_bus = graph.render_destination_for_current_quantum();
+
+    VERIFY(destination_bus.channel_count() > 0);
+
+    for (u32 out_channel = 0; out_channel < channel_count; ++out_channel) {
+        auto const& bus_channel = destination_bus.channel(min(out_channel, destination_bus.channel_count() - 1));
+        auto& output = out.rendered_channels[out_channel];
+        for (u32 i = 0; i < frames_this_quantum; ++i)
+            output[current_frame + i] = bus_channel[i];
```
R-Goc: This looks like it could be made into a memcpy.

Prospero23: Fixed. 

Switched to a bulk copy using  `Span` to mirror `DestinationRenderNode::process()`

jonbgamble:
Hey, Looking good!

If the last branch of mine you looked at was webaudio-canary, I've since reorganized to an out-of-process approach (for live AudioContexts, not Offline). There's a complete webaudio implementation in [this draft PR](https://github.com/LadybirdBrowser/ladybird/pull/8117).

Not sure how much of it would be useful here, but it's there if you get bored

Prospero23:
Sounds good.  I need to look at the new version/ fix the broken build from the rebase. Life has been insane but should have time in next week to finally do it.

PR #7614, opened 2026-01-25
LibMedia: Treat reserved CICP values as unspecified

tcl3:
This aligns with other engines, which treat reserved or invalid CICP values equivalently to unspecified and default to BT.709.

This fixes a crash on the [Photopea home page](https://www.photopea.com/) that happened when trying to load https://www.photopea.com/img/goats.mp4.

I was able to create a minimal reproduction. This crashes when opened in the browser, but when run with `test-web` as a crash test or a text test, no crash occurs:
```html
<!DOCTYPE html>
<video src="data:video/mp4;base64,AAAAIGZ0eXBpc29tAAACAGlzb21pc28yYXZjMW1wNDEAAAAIZnJlZQAABBttZGF0AAACrgYF//+q3EXpvebZSLeWLNgg2SPu73gyNjQgLSBjb3JlIDE2NCByMzEwOCAzMWUxOWY5IC0gSC4yNjQvTVBFRy00IEFWQyBjb2RlYyAtIENvcHlsZWZ0IDIwMDMtMjAyMyAtIGh0dHA6Ly93d3cudmlkZW9sYW4ub3JnL3gyNjQuaHRtbCAtIG9wdGlvbnM6IGNhYmFjPTEgcmVmPTMgZGVibG9jaz0xOjA6MCBhbmFseXNlPTB4MzoweDExMyBtZT1oZXggc3VibWU9NyBwc3k9MSBwc3lfcmQ9MS4wMDowLjAwIG1peGVkX3JlZj0xIG1lX3JhbmdlPTE2IGNocm9tYV9tZT0xIHRyZWxsaXM9MSA4eDhkY3Q9MSBjcW09MCBkZWFkem9uZT0yMSwxMSBmYXN0X3Bza2lwPTEgY2hyb21hX3FwX29mZnNldD0tMiB0aHJlYWRzPTEgbG9va2FoZWFkX3RocmVhZHM9MSBzbGljZWRfdGhyZWFkcz0wIG5yPTAgZGVjaW1hdGU9MSBpbnRlcmxhY2VkPTAgYmx1cmF5X2NvbXBhdD0wIGNvbnN0cmFpbmVkX2ludHJhPTAgYmZyYW1lcz0zIGJfcHlyYW1pZD0yIGJfYWRhcHQ9MSBiX2JpYXM9MCBkaXJlY3Q9MSB3ZWlnaHRiPTEgb3Blbl9nb3A9MCB3ZWlnaHRwPTIga2V5aW50PTI1MCBrZXlpbnRfbWluPTI1IHNjZW5lY3V0PTQwIGludHJhX3JlZnJlc2g9MCByY19sb29rYWhlYWQ9NDAgcmM9Y3JmIG1idHJlZT0xIGNyZj0yMy4wIHFjb21wPTAuNjAgcXBtaW49MCBxcG1heD02OSBxcHN0ZXA9NCBpcF9yYXRpbz0xLjQwIGFxPTE6MS4wMACAAAAAFWWIhAA7//7jq/gU0a+hz3ekUTewQQAAAAhBmiRsQ7/+4AAAAAhBnkJ4hf/BgQAAAAgBnmF0Qr/EgAAAAAgBnmNqQr/EgQAAAA5BmmhJqEFomUwId//+4QAAAApBnoZFESwv/8GBAAAACAGepXRCv8SBAAAACAGep2pCv8SAAAAADkGarEmoQWyZTAh3//7gAAAACkGeykUVLC//wYEAAAAIAZ7pdEK/xIAAAAAIAZ7rakK/xIAAAAAOQZrwSahBbJlMCG///uEAAAAKQZ8ORRUsL//BgQAAAAgBny10Qr/EgQAAAAgBny9qQr/EgAAAAA5BmzRJqEFsmUwIZ//+4AAAAApBn1JFFSwv/8GBAAAACAGfcXRCv8SAAAAACAGfc2pCv8SAAAAADkGbeEmoQWyZTAhX//7BAAAACkGflkUVLC//wYAAAAAIAZ+1dEK/xIEAAAAIAZ+3akK/xIEAAARnbW9vdgAAAGxtdmhkAAAAAAAAAAAAAAAAAAAD6AAAA+gAAQAAAQAAAAAAAAAAAAAAAAEAAAAAAAAAAAAAAAAAAAABAAAAAAAAAAAAAAAAAABAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAgAAA5J0cmFrAAAAXHRraGQAAAADAAAAAAAAAAAAAAABAAAAAAAAA+gAAAAAAAAAAAAAAAAAAAAAAAEAAAAAAAAAAAAAAAAAAAABAAAAAAAAAAAAAAAAAABAAAAAABAAAAAQAAAAAAAkZWR0cwAAABxlbHN0AAAAAAAAAAEAAAPoAAAEAAABAAAAAAMKbWRpYQAAACBtZGhkAAAAAAAAAAAAAAAAAAAyAAAAMgBVxAAAAAAALWhkbHIAAAAAAAAAAHZpZGUAAAAAAAAAAAAAAABWaWRlb0hhbmRsZXIAAAACtW1pbmYAAAAUdm1oZAAAAAEAAAAAAAAAAAAAACRkaW5mAAAAHGRyZWYAAAAAAAAAAQAAAAx1cmwgAAAAAQAAAnVzdGJsAAAAwXN0c2QAAAAAAAAAAQAAALFhdmMxAAAAAAAAAAEAAAAAAAAAAAAAAAAAAAAAABAAEABIAAAASAAAAAAAAAABFUxhdmM2MS4xOS4xMDEgbGlieDI2NAAAAAAAAAAAAAAAGP//AAAAN2F2Y0MBZAAK/+EAGmdkAAqs2V7AWoAAASAAAAMAIAAABkHiRLLAAQAGaOvjyyLA/fj4AAAAABBwYXNwAAAAAQAAAAEAAAAUYnRydAAAAAAAACCYAAAAAAAAABhzdHRzAAAAAAAAAAEAAAAZAAACAAAAABRzdHNzAAAAAAAAAAEAAAABAAAA2GN0dHMAAAAAAAAAGQAAAAEAAAQAAAAAAQAACgAAAAABAAAEAAAAAAEAAAAAAAAAAQAAAgAAAAABAAAKAAAAAAEAAAQAAAAAAQAAAAAAAAABAAACAAAAAAEAAAoAAAAAAQAABAAAAAABAAAAAAAAAAEAAAIAAAAAAQAACgAAAAABAAAEAAAAAAEAAAAAAAAAAQAAAgAAAAABAAAKAAAAAAEAAAQAAAAAAQAAAAAAAAABAAACAAAAAAEAAAoAAAAAAQAABAAAAAABAAAAAAAAAAEAAAIAAAAAHHN0c2MAAAAAAAAAAQAAAAEAAAAZAAAAAQAAAHhzdHN6AAAAAAAAAAAAAAAZAAACywAAAAwAAAAMAAAADAAAAAwAAAASAAAADgAAAAwAAAAMAAAAEgAAAA4AAAAMAAAADAAAABIAAAAOAAAADAAAAAwAAAASAAAADgAAAAwAAAAMAAAAEgAAAA4AAAAMAAAADAAAABRzdGNvAAAAAAAAAAEAAAAwAAAAYXVkdGEAAABZbWV0YQAAAAAAAAAhaGRscgAAAAAAAAAAbWRpcmFwcGwAAAAAAAAAAAAAAAAsaWxzdAAAACSpdG9vAAAAHGRhdGEAAAABAAAAAExhdmY2MS43LjEwMA=="></video>
```

### DIFF
--- a/Tests/LibMedia/CMakeLists.txt
+++ b/Tests/LibMedia/CMakeLists.txt
@@ -2,6 +2,7 @@ include(audio)
 
 set(TEST_SOURCES
     TestAudioDataProvider.cpp
+    TestCICP.cpp
     TestH264Decode.cpp
     TestParseMatroska.cpp
     TestPlaybackStream.cpp
--- a/Tests/LibMedia/TestCICP.cpp
+++ b/Tests/LibMedia/TestCICP.cpp
@@ -1,0 +1,40 @@
+/*
+ * Copyright (c) 2026, Tim Ledbetter <tim.ledbetter@ladybird.org>
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include <LibMedia/Color/CodingIndependentCodePoints.h>
+#include <LibTest/TestCase.h>
+
+TEST_CASE(adopt_specified_values_ignores_reserved_and_unspecified_values)
+{
+    Media::CodingIndependentCodePoints cicp {
+        Media::ColorPrimaries::BT709, Media::TransferCharacteristics::BT709,
+        Media::MatrixCoefficients::BT709, Media::VideoFullRangeFlag::Studio
+    };
+
+    cicp.adopt_specified_values({ Media::ColorPrimaries::Reserved, Media::TransferCharacteristics::Reserved,
+        Media::MatrixCoefficients::Unspecified, Media::VideoFullRangeFlag::Unspecified });
+
+    EXPECT(cicp.color_primaries() == Media::ColorPrimaries::BT709);
+    EXPECT(cicp.transfer_characteristics() == Media::TransferCharacteristics::BT709);
+    EXPECT(cicp.matrix_coefficients() == Media::MatrixCoefficients::BT709);
+    EXPECT(cicp.video_full_range_flag() == Media::VideoFullRangeFlag::Studio);
+}
+
+TEST_CASE(adopt_specified_values_adopts_valid_values)
+{
+    Media::CodingIndependentCodePoints cicp {
+        Media::ColorPrimaries::Unspecified, Media::TransferCharacteristics::Unspecified,
+        Media::MatrixCoefficients::Unspecified, Media::VideoFullRangeFlag::Unspecified
+    };
+
+    cicp.adopt_specified_values({ Media::ColorPrimaries::BT2020, Media::TransferCharacteristics::SRGB,
+        Media::MatrixCoefficients::BT601, Media::VideoFullRangeFlag::Full });
+
+    EXPECT(cicp.color_primaries() == Media::ColorPrimaries::BT2020);
+    EXPECT(cicp.transfer_characteristics() == Media::TransferCharacteristics::SRGB);
+    EXPECT(cicp.matrix_coefficients() == Media::MatrixCoefficients::BT601);
+    EXPECT(cicp.video_full_range_flag() == Media::VideoFullRangeFlag::Full);
+}


Zaggy1024:
Hmm, this makes me think we probably need to default the CICP values in ImmutableBitmap as well, since we probably could crash it even with this patch applied if we just set invalid CICP values in the container format (which should be encoded as single bytes, so they're quite easy to edit).

If we default them in ImmutableBitmap, we could probably also drop that method from the CICP class.

Zaggy1024:
To be clear, I'd rather avoid the crash in ImmutableBitmap instead of putting a fix for only specifically Reserved values that'll just hide the crash for the most common case. Presumably if people are using Reserved when it's explicitly prohibited, there may be other files out there that have values that are completely unspecified.

tcl3:
My latest push moves the defaulting downstream and removes `default_code_points_if_unspecified()`.

FWIW, I tried to induce a crash with just the original change by setting invalid CICP values in various ways and it seems that ffmpeg does its own normalization before the values get to us. Either way, doing the defaulting in `ImmutableBitmap` is probably safer.

I also added a LibMedia unit test for the new behavior. Still not sure why my original repro doesn't work with `test-web`. I assume its missing some media loading capabilities.

Zaggy1024:
I forget if this is the case, but if test-web is designed not to use the GPU, or failing to do so, then it wouldn't hit the crashing code path. We probably need to bring in libyuv for color conversion in the software rendering context, I don't think that Skia can do it itself without shaders.

Libraries/LibGfx/ImmutableBitmap.cpp
Zaggy1024: We may still hit this assertion, we probably should return the same default value here as well? Same for the other switch statement.

(We can't necessarily rely on FFmpeg always fixing these values, especially if we end up needing to migrate away from it for demuxing of more formats than Matroska)

tcl3: It was possible to hit these assertions with a crafted mkv file. I've now made it so out of range values are defaulted.

PR #7617, opened 2026-01-25
LibWeb: Support `@-webkit-keyframes` as an alias for `@keyframes`

tcl3:
This is listed as mandatory in the compat spec.

With this change https://app.diagrams.net/ is now usable:

diagrams net

### DIFF
--- a/Libraries/LibWeb/CSS/Parser/RuleContext.cpp
+++ b/Libraries/LibWeb/CSS/Parser/RuleContext.cpp
@@ -49,7 +49,7 @@ RuleContext rule_context_type_for_at_rule(FlyString const& name)
         return RuleContext::AtMedia;
     if (name.equals_ignoring_ascii_case("font-face"sv))
         return RuleContext::AtFontFace;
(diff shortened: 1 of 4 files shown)


AtkinsSJ:
Hmm, does this need to serialise as `@-webkit-keyframes` too? Shouldn't be hard to add a flag if so.

tcl3:
> Hmm, does this need to serialise as `@-webkit-keyframes` too? Shouldn't be hard to add a flag if so.

Hmm, I tested in Firefox [with this](https://gistpreview.github.io/?a24159563b73635a3debe5548ed23c25) and the serialization of `@-webkit-keyframes` was identical to the normal `@keyframes` rule. But testing in Chromium it does seem to maintain the name distinction. 

I can't find any relevant spec text about this and there's no relevant WPT tests. I'm happy to change it if you think it matches better what we do elsewhere

AtkinsSJ:
If there's no WPT test or consistency between browsers, it's probably fine. I was going to raise a spec issue but it looks like https://github.com/whatwg/compat/issues/220 already covers this.

PR #7624, opened 2026-01-26
LibGfx+LibWebView: Use Skia to resolve generic font families

tcl3:
This allows us to remove the FontConfig dependency from LibWebView.

I have another change I'm working on to also use Skia for font enumeration, which is the only other direct usage of fontconfig we have AFAICT.

### DIFF
--- a/Libraries/LibGfx/Font/TypefaceSkia.cpp
+++ b/Libraries/LibGfx/Font/TypefaceSkia.cpp
@@ -127,9 +127,9 @@ ErrorOr<RefPtr<TypefaceSkia>> TypefaceSkia::find_typeface_for_code_point(u32 cod
     return result;
 }
 
-Optional<FlyString> TypefaceSkia::resolve_generic_family(StringView family_name)
+Optional<FlyString> TypefaceSkia::resolve_generic_family(StringView family_name, u16 weight, u8 slope)
 {
-    SkFontStyle style(SkFontStyle::kNormal_Weight, SkFontStyle::kNormal_Width, SkFontStyle::kUpright_Slant);
+    SkFontStyle style(weight, SkFontStyle::kNormal_Width, slope_to_skia_slant(slope));
     auto skia_typeface = font_manager().matchFamilyStyle(
         ByteString(family_name).characters(), style);
 
--- a/Libraries/LibGfx/Font/TypefaceSkia.h
+++ b/Libraries/LibGfx/Font/TypefaceSkia.h
@@ -17,7 +17,7 @@ class TypefaceSkia : public Gfx::Typeface {
 public:
     static ErrorOr<NonnullRefPtr<TypefaceSkia>> load_from_buffer(ReadonlyBytes, u32 ttc_index = 0);
     static ErrorOr<RefPtr<TypefaceSkia>> find_typeface_for_code_point(u32 code_point, u16 weight, u16 width, u8 slope);
-    static Optional<FlyString> resolve_generic_family(StringView family_name);
+    static Optional<FlyString> resolve_generic_family(StringView family_name, u16 weight, u8 slope);
 
     RefPtr<TypefaceSkia const> clone_with_variations(Vector<FontVariationAxis> const& axes) const;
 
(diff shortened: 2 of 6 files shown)


Libraries/LibWebView/Plugins/FontPlugin.cpp
```diff
@@ -69,79 +66,6 @@ Vector<FlyString> FontPlugin::symbol_font_names()
...
-    FcPattern* pattern = FcNameParse(reinterpret_cast<FcChar8 const*>(pattern_string));
-    VERIFY(pattern);
-
-    auto success = FcConfigSubstitute(config, pattern, FcMatchPattern);
-    VERIFY(success);
-
-    FcDefaultSubstitute(pattern);
-
-    // Never select bitmap fonts.
-    success = FcPatternAddBool(pattern, FC_SCALABLE, FcTrue);
```
tcl3: With these changes we no longer exclude bitmap fonts. There isn't a simple API to do this in Skia as far as I can see.

I can probably do something hacky by inspecting the returned typeface, but I'm not sure how important this filtering is.

gmta:
This is going to conflict heavily with #7590 but not having to use fontconfig is a win imo! Any chance you could integrate the slope/weight changes into this as well?

tcl3:
> Any chance you could integrate the slope/weight changes into this as well?

I've taken the commits from your PR and updated the last one to use Skia rather than fontconfig. I believe the functionality should be equivalent.

gmta:
Closing this in favor of #7624

gmta:
Reopening this 😅

Libraries/LibWebView/Plugins/FontPlugin.cpp
```diff
@@ -135,20 +83,110 @@ void FontPlugin::update_generic_fonts()
...
+        bool has_requested_slope = false;
+        HashTable<u16> available_weights;
+
+        Gfx::FontDatabase::the().for_each_typeface_with_family_name(family, [&](Gfx::Typeface const& typeface) {
+            available_weights.set(typeface.weight());
+            if (typeface.weight() == static_cast<u16>(weight))
+                has_requested_weight = true;
+            if (typeface.slope() == static_cast<u8>(slope))
+                has_requested_slope = true;
+        });
```
AtkinsSJ: Not a blocker, but: How does this interact with variable fonts? eg, I'm guessing a typeface with a `wght` variation axis needs checking manually instead of relying on `weight()` to be useful.

tcl3: Yes, currently `weight()` will return the default weight - we'd need something like `supports_weight(weight)` here to support variable fonts. I had a go at adding support for this, but it isn't particularly useful or easy to test without other changes that I'd prefer to do in a future PR.

Libraries/LibGfx/Font/TypefaceSkia.cpp
```diff
@@ -71,7 +71,15 @@ static SkFontStyle::Slant slope_to_skia_slant(u8 slope)
...
-    auto skia_typeface = font_manager().makeFromData(data, static_cast<int>(ttc_index));
+
+    // https://learn.microsoft.com/en-us/typography/opentype/spec/otff#ttc-header
+    // TrueType Collection files bundle multiple fonts (often different weights of the same
+    // family). We use SkFontArguments to specify which font to load from the collection.
+    SkFontArguments font_args;
+    font_args.setCollectionIndex(static_cast<int>(ttc_index));
+
+    auto stream = std::make_unique<SkMemoryStream>(data);
+    auto skia_typeface = font_manager().makeFromStream(std::move(stream), font_args);
```
AtkinsSJ, line 82: Nit: Probably-unintentional `std::move()` instead of `move()`.

tcl3:
> We probably want some test coverage for loading all the fonts in a ttc.

It's annoyingly difficult to test this, since PathFontProvider is not used in test mode.

Libraries/LibWebView/Plugins/FontPlugin.cpp
tcl3: I'm going to leave this as is for now. Given the rarity of bitmap fonts, I think this is unlikely to have a noticeable impact. Happy to revisit if someone can find a real world case where this causes an issue.

PR #7625, opened 2026-01-26
LibWeb: Improve sequential focus navigation

ReimarPB:
This PR improves navigation using tab by implementing a new function to check if an element should be tab navigatable (sequentially focusable).

Previously, tab navigation used the `is_focusable()` method which is wrong, as it includes invisible elements and serves a different purpose.

According to spec (https://html.spec.whatwg.org/multipage/interaction.html#sequentially-focusable), the user agent decides the algorithm here. I wrote a simple algorithm to make sure the element is visible on the page (no display:none; ancestors and is itself not visibility:hidden;), although it is not perfect, so I'm open to feedback here.

I couldn't figure out a good way to write a test for this (again, please tell me if you have ideas), but it can be tested manually using some HTML like this:

```html
<button>should focus</button>
<button tabindex="-1">should not focus</button>
<button style="display: none">should not focus</button>
<button style="display: none" tabindex="0">should not focus</button>
<button style="visibility: hidden">should not focus</button>
<div style="display: none">
    <button>should not focus</button>
</div>
<div style="visibility: hidden">
    <button style="visibility: visible">should focus</button>
</div>
```

Three buttons will be visible, two of which should be focusable using tab navigation.

### DIFF
--- a/Libraries/LibWeb/DOM/Element.cpp
+++ b/Libraries/LibWeb/DOM/Element.cpp
@@ -4452,6 +4452,20 @@ bool Element::is_focusable() const
     return HTML::parse_integer(get_attribute_value(HTML::AttributeNames::tabindex)).has_value();
 }
 
+// https://html.spec.whatwg.org/multipage/interaction.html#sequentially-focusable
+bool Element::is_sequentially_focusable() const
+{
+    for (auto element = this; element; element = element->parent_element()) {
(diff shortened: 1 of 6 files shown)


gmta:
Nice! Can you add a regression test for this new behavior?

ReimarPB:
@gmta I have added a test now (and fixed the test that crashed earlier)

PR #7630, opened 2026-01-26
LibHTTP+RequestServer: Honor Cache-Control request directives

trflynn89:
We were previously primarily only looking at the Cache-Control response directive. This lets us pass all of [`/fetch/http-cache/cc-request.any.html`](https://wpt.fyi/results/fetch/http-cache/cc-request.any.html?product=ladybird) in WPT (as well as `cc-request.any.worker.html` and `cc-request.any.sharedworker.html`).


trflynn89:
Update: The way we were checking for Cache-Control directives (basically a substring search) was not correct. Added a commit to rectify this.

PR #7634, opened 2026-01-26
AK: Use builtin macro to get cache line size.

R-Goc:
We were previously using our own defines, however those didn't cover various systems. This change makes AK_CACHE_ALIGNED use the destructive interference size as defined by the compiler for the target platform.


spholz:
(ref https://github.com/SerenityOS/serenity/commit/5569930541b0e37a25581ea4b3880182362f4357 / https://github.com/SerenityOS/serenity/pull/26538)

R-Goc:
unstale

Issue #6005, opened 2025-08-27
WebContent process crashes when loading donorbox.org/ladybird


frozen-sea:
### Summary

I was messing around with the browser and though it would be cool to see if it's possible to donate to the project using it, but alas it crashes with the message "Ladybird flew off-course!" when trying to load https://donorbox.org/ladybird

### Operating system

Linux

### Steps to reproduce

1. Ladybird https://donorbox.org/ladybird
2. Boom

### Expected behavior

No crash

### Actual behavior

Crash

### URL for a reduced test case

N/A

### HTML/SVG/etc. source for a reduced test case

```html
<html>
  <head>
    <style>
      label {
        display: inline;
      }
      label div {
        display: inline-block;
      }
    </style>
  </head>
  <body><label><div>
```

### Log output and (if possible) backtrace

```shell
$ ./Build/release/bin/Ladybird ~/Downloads/donorbox.html 
188171.921 WebContent(408603): FIXME: InlineFormattingContext::dimension_box_on_line got unexpected box in inline context:
188171.921 WebContent(408603): Label <label> (not painted) children: inline
  BlockContainer <div> (not painted) inline-block [BFC] children: inline
    TextNode <#text> (not painted)

VERIFICATION FAILED: i < m_size at /home/n/Code/ladybird/AK/Vector.h:156
/home/n/Code/ladybird/Build/release/lib/liblagom-ak.so.0(dump_backtrace+0x49) [0x7f3643349a79]
/home/n/Code/ladybird/Build/release/lib/liblagom-ak.so.0(ak_trap+0x9) [0x7f3643349cd9]
/home/n/Code/ladybird/Build/release/lib/liblagom-ak.so.0(ak_verification_failed+0x65) [0x7f3643349a05]
/home/n/Code/ladybird/Build/release/lib/liblagom-web.so.0(+0x1137283) [0x7f3645537283]
/home/n/Code/ladybird/Build/release/lib/liblagom-web.so.0 Web::DOM::Document::update_layout(Web::DOM::UpdateLayoutReason) 0x663) [0x7f3644ff0b43]
/home/n/Code/ladybird/Build/release/lib/liblagom-web.so.0 Web::HTML::EventLoop::update_the_rendering() 0x279) [0x7f3645229aa9]
/home/n/Code/ladybird/Build/release/lib/liblagom-web.so.0(+0xe2bca1) [0x7f364522bca1]
[...]

PR #7641, opened 2026-01-27
LibWeb: Implement label activation behavior in the DOM layer 

tcl3:
Previously, click handling for labels was handled in layout and painting  code. This change implements activation_behavior on HTMLLabelElement, which clicks and focuses the element.

Fixes: #6005


awesomekling:
This is so tasty holy heck

PR #7647, opened 2026-01-27
LibWeb: CSS Property computation optimizations

Calme1709:
A collection of optimizations to `StyleComputer::compute_properties` which collectively reduce runtime from ~31.5% to ~17.3% when loading https://en.wikipedia.org/wiki/2023_in_American_television - see individual commits for details.

### DIFF
--- a/Libraries/LibWeb/CSS/CascadedProperties.cpp
+++ b/Libraries/LibWeb/CSS/CascadedProperties.cpp
@@ -41,8 +41,10 @@ void CascadedProperties::revert_property(PropertyID property_id, Important impor
             && entry.property.important == important
             && cascade_origin == entry.origin;
     });
-    if (entries.is_empty())
+    if (entries.is_empty()) {
+        m_contained_properties_cache.set(to_underlying(property_id), false);
         m_properties.remove(it);
+    }
 }
 
 void CascadedProperties::revert_layer_property(PropertyID property_id, Important important, Optional<FlyString> layer_name)
@@ -56,8 +58,10 @@ void CascadedProperties::revert_layer_property(PropertyID property_id, Important
             && entry.property.important == important
             && layer_name == entry.layer_name;
     });
-    if (entries.is_empty())
+    if (entries.is_empty()) {
+        m_contained_properties_cache.set(to_underlying(property_id), false);
         m_properties.remove(it);
+    }
 }
 
 void CascadedProperties::resolve_unresolved_properties(DOM::AbstractElement abstract_element)
@@ -73,6 +77,8 @@ void CascadedProperties::resolve_unresolved_properties(DOM::AbstractElement abst
 
 void CascadedProperties::set_property(PropertyID property_id, NonnullRefPtr<StyleValue const> value, Important important, CascadeOrigin origin, Optional<FlyString> layer_name, GC::Ptr<CSS::CSSStyleDeclaration const> source)
 {
+    m_contained_properties_cache.set(to_underlying(property_id), true);
+
     auto& entries = m_properties.ensure(property_id);
 
     for (auto& entry : entries.in_reverse()) {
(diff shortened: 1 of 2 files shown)


Calme1709:
Rebased to resolve conflicts

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1964,22 +1974,23 @@ GC::Ref<ComputedProperties> StyleComputer::compute_properties(DOM::AbstractEleme
...
 
+    if (is<HTML::HTMLHtmlElement>(abstract_element.element()))
+        const_cast<StyleComputer&>(*this).m_root_element_font_metrics = calculate_root_element_font_metrics(computed_style);
+
     // Compute the value of custom properties
     compute_custom_properties(computed_style, abstract_element);
 
-    // Convert properties into their computed forms
-    compute_property_values(computed_style, abstract_element);
-
```
AtkinsSJ: Doesn't property computation have to come after `compute_custom_properties()`?

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1972,17 +1974,26 @@ GC::Ref<ComputedProperties> StyleComputer::compute_properties(DOM::AbstractEleme
...
         if (should_inherit) {
             inherited = ComputedProperties::Inherited::Yes;
-            value = get_non_animated_inherit_value(property_id, abstract_element);
+
+            if (abstract_element.element_to_inherit_style_from().has_value()) {
+                value = get_non_animated_inherit_value(property_id, abstract_element);
+                requires_computation = property_requires_computation_with_inherited_value(property_id);
+            } else {
+                value = property_initial_value(property_id);
+                requires_computation = property_requires_computation_with_initial_value(property_id);
```
AtkinsSJ: This seems redundant with the code a few lines down.

AtkinsSJ (changes requested):
Very cool! The system for computing properties in order is something I've wanted for a long time, that'll be handy.

The `requires-computation` data will need updating for any properties that take `<custom-ident>` or `<dashed-ident>` once `ident()` is implemented, but that's a future concern. (I don't envy you going through all the properties to fill those in. 😅)

Libraries/LibWeb/CSS/StyleComputer.cpp
Calme1709: This handles the case where we are trying to inherit without an inheritance parent, the code below handles the case where we explicitly want the initial value. I have refactored slightly to deduplicate this.

Libraries/LibWeb/CSS/StyleComputer.cpp
Calme1709: Not to my knowledge - resolution of `var()` is done at cascade time using the un-computed custom properties and I am not aware of any other interactions.

Calme1709:
Rebased to master to add missing `requires-computation` property after 046dfdd was merged.

Libraries/LibWeb/CSS/StyleComputer.cpp
AtkinsSJ: `var()` is supposed to use computed custom property values, we just currently handle those in a weird way. ([Spec:](https://drafts.csswg.org/css-variables/#replace-a-var-function) "let result be the computed value of the corresponding custom property on el")
I really need to give #6608 another look. :sweat_smile:

Libraries/LibWeb/CSS/StyleComputer.h
```diff
@@ -193,6 +193,17 @@ class WEB_API StyleComputer final : public GC::Cell {
     Length::FontMetrics m_default_font_metrics;
     Length::FontMetrics m_root_element_font_metrics;
 
+    Optional<ComputationContext> m_cached_font_computation_context;
+    Optional<ComputationContext> m_cached_line_height_computation_context;
+    Optional<ComputationContext> m_cached_generic_computation_context;
```
AtkinsSJ, line 198: Probably makes sense to mark these as `mutable` instead of the `const_cast`s?

awesomekling: I don't disagree with this, just merging this to avoid causing conflicts :D

Calme1709:
Looks like this introduced a test failure in CI - fix here #7941

PR #7648, opened 2026-01-27
AK+LibMedia: Add a MPSC and SPSC ring buffer

R-Goc:
This PR adds an MPSC and SPSC ring buffer in AK/RingBuffer.h. The implementation approach is using nodes that each hold an atomic sequence number for the MPSC buffer. This has a larger space overhead, but a lower latency due to lesser contention on the atomics. However ideally the types should be large when used with this ring buffer. The SPSC implementation just uses two atomics. 

The playback stream implementations on windows and MacOS use this instead of a queue protected by a muetx. This avoids a mutex in the render callback.

### DIFF
--- a/AK/RingBuffer.h
+++ b/AK/RingBuffer.h
@@ -100,6 +100,53 @@ public:
         }
     }
 
+    // Spins until pushing is possible
+    ALWAYS_INLINE void push(T value)
+    {
+        return emplace(move(value));
+    }
+
+    // Spins until pushing is possible
+    template<typename U>
+    requires(IsConvertible<U, T> && !IsSame<RemoveCVReference<U>, T>)
+    ALWAYS_INLINE void push(U&& value)
+    {
+        return emplace(forward<U>(value));
+    }
+
+    // Spins until emplacing is possible
+    template<typename... Args>
+    requires(IsConstructible<T, Args...>)
+    ALWAYS_INLINE void emplace(Args&&... args)
+    {
+        u32 head = m_head.load(MemoryOrder::memory_order_relaxed);
+        Backoff backoff;
+
+        while (true) {
+            Node& slot = m_data[get_offset(head)];
+            u32 sequence = slot.sequence.load(MemoryOrder::memory_order_acquire);
+            i32 diff = static_cast<i32>(sequence - head);
+
+            if (diff == 0) {
+                // Slot is free
+                if (m_head.compare_exchange_weak(head, head + 1, MemoryOrder::memory_order_acq_rel)) {
+                    // We now own the slot
+                    new (slot.data) T(forward<Args>(args)...);
+                    slot.sequence.store(head + 1, MemoryOrder::memory_order_release);
+                    break;
+                }
+                // The head was updated by another thread, try again
+                cpu_pause();
+            } else if (diff < 0) {
+                // Buffer full
+                backoff.tick();
+            } else {
+                //  Our head is stale
+                head = m_head.load(MemoryOrder::memory_order_relaxed);
+            }
+        }
+    }
+
     [[nodiscard]] ALWAYS_INLINE bool try_pop(T& value)
     {
         Node& slot = m_data[get_offset(m_tail)];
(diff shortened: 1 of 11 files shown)


AK/StdLibExtraDetails.h
```diff
@@ -508,6 +509,9 @@ inline constexpr bool IsConstructible = requires { ::new T(declval<Args>()...);
 template<typename T, typename... Args>
 inline constexpr bool IsTriviallyConstructible = __is_trivially_constructible(T, Args...);
 
+template<typename T>
+inline constexpr bool IsPlacementNewConstructible = requires(void* ptr) { ::new (ptr, std::nothrow) T; };
```
Hendiadyoin1: Whats the difference between this and plain `IsConstructible`
or is there something overriding and deleting placement new?

Tests/AK/CMakeLists.txt
```diff
@@ -58,6 +58,7 @@ set(AK_TEST_SOURCES
     TestRandom.cpp
     TestRedBlackTree.cpp
     TestRefPtr.cpp
+    TestRingBuffer.cpp
```
Hendiadyoin1, line 63: This file seems to have gone missing

AK/StdLibExtraDetails.h
Hendiadyoin1: And you dont even end up using it...

Hendiadyoin1 (commented):
Very early review, but do you have any usecase in mind?

AK/StdLibExtraDetails.h
R-Goc: I'll remove that, I don't need it now. I was originally doing something else with the lifetime.

R-Goc:
> Very early review, but do you have any usecase in mind?

Not sure if I should include it in this PR, but I wanted to use it instead of a queue protected by a mutex for PlaybackStreamWasapi. Mutexes in audio code are not good. Also there is more work in webaudio that will be using ring buffers, and instead of adding them to LibCore they could then be added to this file (specifically an SPSC ring buffer). This design can easily be extended to an MPMC buffer as well, just by adding a compare exchange loop in the reader and making the read index atomic.

AK/StdLibExtraDetails.h
R-Goc: Actually I might want to probably add something like std::is_default_constructible as we don't have that. IsConstructible with null arguments checks value initialization. But the only difference would be a nicer compiler error message.

AK/RingBuffer.h
```diff
@@ -169,6 +243,44 @@ class SPSCRingBuffer {
...
+                if (head - m_cached_tail == Size) {
+                    backoff.tick();
+                    continue;
+                }
+                break;
+            }
+        }
+
+        // No need to launder here as nothing accesses the pointer given to new
+        new (m_data[get_offset(head)]) T(forward<Args>(args)...);
```
Hendiadyoin1: With placement new you should never need to launder, as you get a pointer to a valid existing object from it, as far as the c++ vm is concerned

AK/RingBuffer.h
```diff
@@ -0,0 +1,346 @@
...
+                // Buffer full
+                backoff.tick();
+            } else {
+                //  Our head is stale
+                head = m_head.load(MemoryOrder::memory_order_relaxed);
+            }
+        }
+    }
+
+    [[nodiscard]] ALWAYS_INLINE bool try_pop(T& value)
```
Hendiadyoin1, line 150: Question on the design:
Why use a glorified return slot instead of an Optional?
(same for the non try version but with an actual value)

AK/RingBuffer.h
```diff
@@ -0,0 +1,346 @@
...
+    {
+        Node& slot = m_data[get_offset(m_tail)];
+
+        u32 sequence = slot.sequence.load(MemoryOrder::memory_order_acquire);
+
+        i32 diff = static_cast<i32>(sequence - (m_tail + 1));
+
+        if (diff == 0) {
+            // The slot is ready for reading
+            T* ptr = reinterpret_cast<T*>(slot.data);
```
Hendiadyoin1, line 160: Buuut you may need to launder here?
I find some examples of laundering when accessing type erased storage, but fewer than I have expected
(likely as most users I trust to do it right (llvm) use unions for uninitialized storage)
But we don't properly do this in other places so this might be fine

AK/RingBuffer.h
R-Goc: Based on my counting it avoids at least one move assignment. I did end up adding an optional versions as otherwise it can't work with non-default constructible types. However, when possible I prefer the out parameter as it works nicer with while loops. Consider:
```cpp
T t;
while (m_ring_buffer.try_pop(t)) {
	// do something with t
}
```
vs using optional
```cpp
Optional<T> maybe_t = m_ring_buffer.try_pop();
while (maybe_t.has_value()) {
	T t = maybe_t.release_value();
	// do something with t
	maybe_t = m_ring_buffer.try_pop();
}
```

AK/RingBuffer.h
R-Goc: Yeah, I'm not 100% sure. Probably fine? Works when I tried it with PlaybackStreamWasapi.

AK/RingBuffer.h
Hendiadyoin1: Depending on how it is used, and considering optional has perfect forwarding, I dont quite see where this would save a move assignment, especially looking at the blocking version but what ever...

R-Goc: I added the blocking version to avoid someone spinning with an atomic pause instead. Also when you use the blocking version you don't care about performance already.

R-Goc: The way I counted it in try_pop with the return argument you get a single move assignment. With optional you assign to that, then you have to hope NRVO will work, then on the release_value() I'm not sure if that adds a move or not. It was more about me not liking optional too much. Maybe if we added a bool overload to optional it would be nicer? That way you don't get the while loop being ugly, I think.

Hendiadyoin1: Well there is a proposal allowing `operator bool` on structured bindings, but the way it is currently specced you cant get it to optionally unwrap (which the authors mention, and lament iirc)
so the nicest version of
```c++
while (auto&& [v] : queue.try_pop()) {
    ...
}
```
would not work, but adding one would at least make this work (not sure if we need the extra move, std::optional needs it at least, but they cant explicitly release)
```c++
while (auto __o = queue.try_pop()){
    auto v = move(o).release_value();
    ....
}
 ```
and for some `while (auto o = get_work(); o.has_value())` is not allowed, but that would work for `if`s

but to ensure copy elision when needed we would need special tests counting those

but yeah I see the reason, personally just not a fan of manual return slots

R-Goc:
When we switch to C++26 this could be implemented with the new trivial unions, after both P3074R7 and its ammendment P3726R2 get implemented.

PR #4330, opened 2025-04-11
Fullscreen API implementation

theIDinside:
This PR implements the Fullscreen API. 

The final commit also adds some Qt UI (an "exit fullscreen" button that animates from the top down, sort of like how Chrome does it, and also Firefox). For Qt-backends, the escape key also exits out of fullscreen, fully.

The spec can be found [here](https://fullscreen.spec.whatwg.org/).

New web platform test results (fullscreen/api) with this patch series applied:

```
Ran 57 tests finished in 13.7 seconds.
  • 36 ran as expected. 0 tests skipped.
  • 1 tests had errors unexpectedly
  • 1 tests timed out unexpectedly
  • 21 tests had unexpected subtest results
```
(up from 0 expected to 36. Another test succeeds when #4329 is applied)

Additional work that needs to happen:
- https://fullscreen.spec.whatwg.org/#dom-document-fullscreenenabled needs to be implemented fully, where a document's "allowed to use" can be deterimined by the `allowFullscreen` attribute set on an iframe. This attribute can not be changed dynamically and once set for a document, is set for the rest of it's life time.

Additional work that's required can be determined by the WPT suite.

PR #7649, opened 2026-01-27
LibWeb+UI: Implement the Fullscreen API

Lubrsi:
This is a revival of PR https://github.com/LadybirdBrowser/ladybird/pull/4330 opened by @theIDinside

https://github.com/user-attachments/assets/e3d8793f-956c-44db-9519-440913ddb0f4


Libraries/LibWeb/CSS/PseudoClassBitmap.h
```diff
@@ -6,40 +6,46 @@
...
 #pragma once
 
+#include <AK/Array.h>
 #include <LibWeb/CSS/PseudoClass.h>
 
 namespace Web::CSS {
 
 class PseudoClassBitmap {
 public:
+    static constexpr size_t bits_per_word = 64;
```
shannonbooth, line 16: I think it would make sense to split this into it's own commit to handle > 64 pseudo classes

sidenote: This commit message as well as some of the next ones are in the past tense (Use of the word "added" in particular to "adds")

Libraries/LibWeb/DOM/Element.cpp
```diff
@@ -2371,6 +2376,183 @@ WebIDL::ExceptionOr<void> Element::insert_adjacent_html(String const& position,
...
+        return RequestFullscreenError::UnsupportedElement;
+
+    // The fullscreen element ready check for this returns true.
+    if (!element.element_ready_check())
+        return RequestFullscreenError::ElementReadyCheckFailed;
+    // FIXME: Implement 'Fullscreen is supported.' check
+
+    // This’s relevant global object has transient activation or
+    // FIXME: the algorithm is triggered by a user generated orientation change.
+    HTML::Window* window = as_if<HTML::Window>(&HTML::relevant_global_object(element));
```
shannonbooth: Given this is ASSERT'd below, maybe can just call as here?

Libraries/LibWeb/DOM/Element.cpp
```diff
@@ -2371,6 +2376,183 @@ WebIDL::ExceptionOr<void> Element::insert_adjacent_html(String const& position,
...
+        //      Append (fullscreenerror, this) to pendingDoc’s list of pending fullscreen events.
+        //      Reject promise with a TypeError exception and terminate these steps.
+        if (error != RequestFullscreenError::False) {
+            pending_doc->append_pending_fullscreen_change(PendingFullscreenEvent::Type::Error, requesting_element);
+            WebIDL::reject_promise(realm, promise, JS::TypeError::create(realm, to_string(error)));
+            return;
+        }
+
+        // 11. Let fullscreenElements be an ordered set initially consisting of this.
+        Vector<GC::Ref<Element>> fullscreen_elements;
```
shannonbooth: This _might_ be safe (I didn't check very hard), but I think we should use a Root/HeapVector here instead

Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -6566,6 +6567,194 @@ bool Document::fullscreen_enabled() const
...
+    HTML::TemporaryExecutionContext context(realm(), HTML::TemporaryExecutionContext::CallbacksEnabled::Yes);
+    (void)exit_fullscreen();
+}
+
+// https://fullscreen.spec.whatwg.org/#exit-fullscreen
+GC::Ref<WebIDL::Promise> Document::exit_fullscreen()
+{
+    auto& realm = this->realm();
+
+    auto* doc = this;
```
shannonbooth: I think we should maybe drop this, it's a bit confusing that sometimes we are using doc as this, and other times implicitly (like when calling is_fully_active).

Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -6645,6 +6651,332 @@ void Document::remove_render_blocking_element(GC::Ref<Element> element)
...
+        // 4. Let container be lastDoc’s node navigable’s container.
+        // Note on spec: It doesn't first check if `node navigable` is null.
+        auto container = last_doc->navigable() ? last_doc->navigable()->container() : nullptr;
+
+        // 5. If container is null, then break.
+        if (!container)
+            break;
+
+        // 6. If container’s iframe fullscreen flag is set, break.
+        if (auto* iframe_element = as<HTML::HTMLIFrameElement>(container.ptr()); iframe_element->iframe_fullscreen_flag())
```
tcl3: Should this be `as_if` rather than `as`? `HTMLIFrameElement` isn't the only element that is a `NavigableContainer`.

Libraries/LibWeb/DOM/Element.cpp
```diff
@@ -2400,6 +2406,237 @@ WebIDL::ExceptionOr<void> Element::insert_adjacent_html(String const& position,
...
+    }
+    VERIFY_NOT_REACHED();
+}
+
+// step 5 of requestFullscreen:
+// 5. If any of conditions are false, set error to true
+static RequestFullscreenError fullscreen_has_error_check(Element const& element)
+{
+    // This’s namespace is the HTML namespace or this is an SVG svg or MathML math element. [SVG] [MATHML]
+    if (!(element.namespace_uri() == Namespace::HTML || element.is_svg_svg_element() || is<MathML::MathMLElement>(element)))
```
tcl3: The spec says we should only be checking for MathML `<math>`  elements here, not any MathML element. We don't have MathMLMathElement, but I assume you can narrow this down by tag name.

Tests/LibWeb/Crash/wpt-import/fullscreen/crashtests/chrome-1312699.html
```diff
@@ -0,0 +1,21 @@
+<!doctype html>
+<html class="test-wait">
+<link rel="help" href="https://crbug.com/1312699">
+<script src="../../resources/testdriver.js"></script>
+<script src="../../resources/testdriver-vendor.js"></script>
```
tcl3: Given this test uses testdriver.js, this test probably doesn't work as expected. It may work If all of the things it uses are implemented in `testdriver-vendor.js`, but this test doesn't point to our existing `testdriver-vendor.js`.

Lubrsi: I've modified the test to point to the ones in Text, it's a bit silly but it works for now with a FIXME.

tcl3: `Tests/LibWeb/Crash/wpt-import/resources/testdriver.js` shouldn't be needed anymore.

shannonbooth:
bar CI failure :P

PR #7652, opened 2026-01-28
LibWeb: Support ::part() pseudo-element chaining

adamjcolvin:
## Summary

This change enables support for chaining pseudo-elements after the `::part()` pseudo-element in CSS selectors (e.g. `::part(header)::before` or `::part(input)::placeholder`). It is split into three atomic commits, each non-regressing.

Relevant specification: https://drafts.csswg.org/css-shadow-parts-1/#part-styleable-from-outside

## Commit Structure and Rationale

The work is ordered so each commit passes all tests on its own. I initially started with just the third commit to update the parser, but each change revealed further changes that needed to be made. See below. 

1. **LibWeb: Implement exportparts forwarding in part element map** — Implements the forwarded part names step of the CSS Shadow Parts spec in `ShadowRoot::calculate_part_element_map()`. Previously, parts forwarded via `exportparts` across nested shadow boundaries were not in the part element map, so they could not be styled. This commit adds proper forwarding (including renaming) and fixes 4 WPT tests.

2. **LibWeb: Only match ::part() in direct child shadow trees** — Adds `rule_shadow_root` to `MatchContext` and restricts `::part()` matching to shadow trees whose host is a direct child of the rule's scope. Without commit 1, this would regress tests that rely on `exportparts` (e.g. `both-part-and-exportparts.html`), because forwarded parts would not yet be in the correct maps.

3. **LibWeb: Support ::part() pseudo-element chaining** — Parser and selector changes to allow `::part()` followed by one other pseudo-element. Depends on the scope fix so that chained selectors match in the correct shadow trees.

## Implementation Details

**Exportparts forwarding (`ShadowRoot.cpp`)**: Parses the `exportparts` attribute into `(inner_name, outer_name)` pairs, consults the inner shadow root's part element map, and adds matching parts to the outer map under the exported name. Supports shorthand (`exportparts="foo"`) and renaming (`exportparts="foo: bar"`), and chains transitively through nested shadow roots.

**Scope checking (`SelectorEngine.cpp`, `SelectorEngine.h`, `StyleComputer.cpp`)**: During `::part()` matching, we skip shadow roots whose host's `containing_shadow_root()` is not equal to the rule's shadow root. Per the spec, `::part()` only matches parts in child shadow trees, not in the same tree as the rule or in deeper nested trees.

**Parser changes (`SelectorParsing.cpp`)**: Tracks the first and second pseudo-elements in a compound selector. When there is more than one, the parser permits it only if the first is `::part()` and the second is not. A maximum of two pseudo-elements is enforced.

**Selector object changes (`Selector.cpp`, `Selector.h`)**: The `Selector` stores the last pseudo-element (the styling target) instead of the first, and uses `m_contains_part_pseudo_element` to track whether `::part()` is present for the selector engine.

## Testing

**12 previously failing WPT tests now pass:**

- **4** in `css/css-shadow-parts/`: `simple-forward`, `simple-forward-shorthand`, `double-forward`, `precedence-part-vs-part`
- **6** in `css/selectors/parsing/parse-part.html`: chained selectors (`::part(name)::before`, `::after`, `::placeholder`, `::first-line`, `::first-letter`, `::file-selector-button`)
- **2** in `css/css-shadow-parts/multiple-scopes.html`: `exported part selector matches from outer scope`, `correct part selector matches from middle scope`

Expected result files have been updated accordingly.

Test sources:
- [Tests/LibWeb/Text/input/wpt-import/css/selectors/parsing/parse-part.html](https://github.com/LadybirdBrowser/ladybird/blob/master/Tests/LibWeb/Text/input/wpt-import/css/selectors/parsing/parse-part.html)
- [Tests/LibWeb/Text/input/wpt-import/css/css-shadow-parts/multiple-scopes.html](https://github.com/LadybirdBrowser/ladybird/blob/master/Tests/LibWeb/Text/input/wpt-import/css/css-shadow-parts/multiple-scopes.html)
- [Tests/LibWeb/Text/input/wpt-import/css/css-shadow-parts/simple-forward.html](https://github.com/LadybirdBrowser/ladybird/blob/master/Tests/LibWeb/Text/input/wpt-import/css/css-shadow-parts/simple-forward.html)
- [Tests/LibWeb/Text/input/wpt-import/css/css-shadow-parts/simple-forward-shorthand.html](https://github.com/LadybirdBrowser/ladybird/blob/master/Tests/LibWeb/Text/input/wpt-import/css/css-shadow-parts/simple-forward-shorthand.html)
- [Tests/LibWeb/Text/input/wpt-import/css/css-shadow-parts/double-forward.html](https://github.com/LadybirdBrowser/ladybird/blob/master/Tests/LibWeb/Text/input/wpt-import/css/css-shadow-parts/double-forward.html)
- [Tests/LibWeb/Text/input/wpt-import/css/css-shadow-parts/precedence-part-vs-part.html](https://github.com/LadybirdBrowser/ladybird/blob/master/Tests/LibWeb/Text/input/wpt-import/css/css-shadow-parts/precedence-part-vs-part.html)

### DIFF
--- a/Libraries/LibWeb/CSS/Parser/SelectorParsing.cpp
+++ b/Libraries/LibWeb/CSS/Parser/SelectorParsing.cpp
@@ -158,17 +158,33 @@ Parser::ParseErrorOr<NonnullRefPtr<Selector>> Parser::parse_complex_selector(Tok
             }
         }
     }
+    // https://drafts.csswg.org/css-shadow-1/#selectordef-part
+    // The ::part() pseudo-element can be followed by other pseudo-elements to style pseudo-elements of the part itself.
     auto pseudo_element_count = 0;
+    Optional<PseudoElement> first_pseudo_element;
+    Optional<PseudoElement> second_pseudo_element;
     for (auto const& simple_selector : parsed_selector->compound_selectors().last().simple_selectors) {
-        if (simple_selector.type == Selector::SimpleSelector::Type::PseudoElement)
+        if (simple_selector.type == Selector::SimpleSelector::Type::PseudoElement) {
             ++pseudo_element_count;
+            if (!first_pseudo_element.has_value())
+                first_pseudo_element = simple_selector.pseudo_element().type();
+            else if (!second_pseudo_element.has_value())
+                second_pseudo_element = simple_selector.pseudo_element().type();
+        }
     }
     if (pseudo_element_count > 1) {
-        ErrorReporter::the().report(InvalidSelectorError {
-            .value_string = parsed_selector->serialize(),
-            .description = "Multiple pseudo elements in a selector is not yet supported."_string,
-        });
-        return ParseError::SyntaxError;
+        // FIXME: Other pseudo-elements can also be chained (e.g. ::highlight()::before).
+        //        For now, only ::part() followed by one other pseudo-element is supported.
+        bool is_valid_chain = first_pseudo_element.value() == PseudoElement::Part
+            && second_pseudo_element.value() != PseudoElement::Part
+            && pseudo_element_count == 2;
+        if (!is_valid_chain) {
+            ErrorReporter::the().report(InvalidSelectorError {
+                .value_string = parsed_selector->serialize(),
+                .description = "Pseudo-element chaining is not yet supported for this combination."_string,
+            });
+            return ParseError::SyntaxError;
+        }
     }
 
     return parsed_selector;
--- a/Libraries/LibWeb/CSS/Selector.cpp
+++ b/Libraries/LibWeb/CSS/Selector.cpp
@@ -81,8 +81,9 @@ Selector::Selector(Vector<CompoundSelector>&& compound_selectors)
     if (!m_compound_selectors.is_empty()) {
         for (auto const& simple_selector : m_compound_selectors.last().simple_selectors) {
             if (simple_selector.type == SimpleSelector::Type::PseudoElement) {
+                if (simple_selector.pseudo_element().type() == PseudoElement::Part)
+                    m_contains_part_pseudo_element = true;
                 m_pseudo_element = simple_selector.pseudo_element();
-                break;
             }
         }
     }
--- a/Libraries/LibWeb/CSS/Selector.h
+++ b/Libraries/LibWeb/CSS/Selector.h
@@ -232,7 +232,7 @@ public:
     size_t sibling_invalidation_distance() const;
 
     bool is_slotted() const { return m_pseudo_element.has_value() && m_pseudo_element->type() == PseudoElement::Slotted; }
-    bool has_part_pseudo_element() const { return m_pseudo_element.has_value() && m_pseudo_element->type() == PseudoElement::Part; }
+    bool has_part_pseudo_element() const { return m_contains_part_pseudo_element; }
 
 private:
     explicit Selector(Vector<CompoundSelector>&&);
@@ -244,6 +244,7 @@ private:
     bool m_can_use_fast_matches { false };
     bool m_can_use_ancestor_filter { false };
     bool m_contains_the_nesting_selector { false };
+    bool m_contains_part_pseudo_element { false };
 
     PseudoClassBitmap m_contained_pseudo_classes;
 
--- a/Tests/LibWeb/Text/expected/wpt-import/css/css-shadow-parts/multiple-scopes.txt
+++ b/Tests/LibWeb/Text/expected/wpt-import/css/css-shadow-parts/multiple-scopes.txt
@@ -2,14 +2,13 @@ Harness status: OK
 
 Found 10 tests
 
-8 Pass
-2 Fail
+10 Pass
 Pass	successful test setup
-Fail	exported part selector matches from outer scope
+Pass	exported part selector matches from outer scope
 Pass	non-exported part selector does not match from outer scope
 Pass	pseudo-element selector alone does not match from outer scope
 Pass	exported part selector (for outer scope) does not match from middle scope
-Fail	correct part selector matches from middle scope
+Pass	correct part selector matches from middle scope
 Pass	pseudo-element selector alone does not match from middle scope
 Pass	selector with ::part(exported name) does not match from inner scope that exports the part
 Pass	selector with ::part(original name) does not match from inner scope that exports the part
--- a/Tests/LibWeb/Text/expected/wpt-import/css/selectors/parsing/parse-part.txt
+++ b/Tests/LibWeb/Text/expected/wpt-import/css/selectors/parsing/parse-part.txt
@@ -2,8 +2,8 @@ Harness status: OK
 
 Found 32 tests
 
-21 Pass
-11 Fail
+27 Pass
+5 Fail
 Pass	"::part(--foo)" should be a valid selector
 Pass	"::part(bar)" should be a valid selector
 Pass	"::part(--)" should be a valid selector
@@ -13,12 +13,12 @@ Pass	"::part(-foo bar)" should be a valid selector
 Pass	"::part(foo):focus" should be a valid selector
 Pass	"::part(foo):hover" should be a valid selector
 Pass	"::part(foo):focus-within" should be a valid selector
-Fail	"::part(foo)::before" should be a valid selector
-Fail	"::part(foo)::after" should be a valid selector
-Fail	"::part(foo)::placeholder" should be a valid selector
-Fail	"::part(foo)::first-line" should be a valid selector
-Fail	"::part(foo)::first-letter" should be a valid selector
-Fail	"::part(foo)::file-selector-button" should be a valid selector
+Pass	"::part(foo)::before" should be a valid selector
+Pass	"::part(foo)::after" should be a valid selector
+Pass	"::part(foo)::placeholder" should be a valid selector
+Pass	"::part(foo)::first-line" should be a valid selector
+Pass	"::part(foo)::first-letter" should be a valid selector
+Pass	"::part(foo)::file-selector-button" should be a valid selector
 Pass	"::part(foo):is(:focus)" should be a valid selector
 Pass	":lang(en)::part(foo)" should be a valid selector
 Pass	":dir(ltr)::part(foo)" should be a valid selector


Libraries/LibWeb/CSS/Selector.cpp
```diff
@@ -76,13 +76,12 @@ static bool can_selector_use_fast_matches(Selector const& selector)
 Selector::Selector(Vector<CompoundSelector>&& compound_selectors)
     : m_compound_selectors(move(compound_selectors))
 {
-    // FIXME: This assumes that only one pseudo-element is allowed in a selector, and that it appears at the end.
-    //        This is not true in Selectors-4!
-    if (!m_compound_selectors.is_empty()) {
```
AtkinsSJ, line 81: This fixme still applies, and `!m_compound_selectors.is_empty()` is clearer than `m_compound_selectors.size()`.

Libraries/LibWeb/CSS/SelectorEngine.cpp
```diff
@@ -1136,13 +1136,15 @@ static inline bool matches(CSS::Selector::SimpleSelector const& component, DOM::
             // FIXME: Support matching pseudo-elements.
             DOM::AbstractElement const abstract_element { const_cast<DOM::Element&>(element) };
 
-            // Potentially any ancestor shadow-host could be owner of the part, so walk up the tree until we find one.
-            // FIXME: That owner needs to be in scope for the rule. How do we tell?
-            // FIXME: How does this interact with :host ?
```
AtkinsSJ, line 1156: These FIXMEs also still apply I think?

Libraries/LibWeb/CSS/SelectorEngine.cpp
```diff
@@ -1136,13 +1136,15 @@ static inline bool matches(CSS::Selector::SimpleSelector const& component, DOM::
...
+            // https://drafts.csswg.org/css-shadow-1/#part
+            // "The ::part() pseudo-element only matches anything when the originating element is a shadow host."
             for (auto ancestor_shadow_root = element.containing_shadow_root();
                 ancestor_shadow_root;
                 ancestor_shadow_root = ancestor_shadow_root->containing_shadow_root()) {
 
+                if (ancestor_shadow_root == context.rule_shadow_root)
+                    continue;
```
AtkinsSJ: I'm a bit confused by this. Why would we skip that one in particular? This seems like the cause of the test regression.

Libraries/LibWeb/CSS/Selector.cpp
adamjcolvin: Yes you're right, I've put the fixme back and gone back to the `is_empty()` check.

Libraries/LibWeb/CSS/SelectorEngine.cpp
adamjcolvin: Yeah that condition was wrong, and it was causing the regression. I've changed it so that we now check the host of the ancestor shadow root, and we check that it's a direct child of the rule's scope. My understanding is that we should only consider shadow roots whose host is directly under the rule's scope.

So for example, my understanding is that this rule should not produce a match for this document, but without the check here, it would?

```html
<!-- Document (rule scope) -->
<style>::part(foo) { color: red }</style>  <!-- shouldn't match, but without checking the scope, it would -->

<div id="outer">                    <!-- shadow host, direct child of document -->
  #shadow-root
  <div id="inner">                  <!-- shadow host, child of outer's shadow -->
    #shadow-root
    <span part="foo">text</span>    <!-- part="foo" not exported -->
  </div>
</div>
```

[CSS Shadow Module Level 1 – part element map](https://drafts.csswg.org/css-shadow-1/#part-element-map). The part element map is defined in terms of "descendants" of the shadow root, and the spec says that descendants of an element do not include its shadow trees—so we only see parts in direct child shadow trees.

adamjcolvin:
@AtkinsSJ thanks for the review, and sorry for taking so long to come back to you. I've split the work into three commits now, ordered so that each one avoids any regressions. The parser change to allow the chaining revealed some issues with part matching and then with `exportparts`. I’ve fixed those in the first two commits (exportparts forwarding, then the scope fix), so that the third commit can add the chaining without any regressions. I've given a full breakdown in the description. Let me know what you think, thanks!

Libraries/LibWeb/DOM/ShadowRoot.cpp
```diff
@@ -267,6 +267,34 @@ ShadowRoot::PartElementMap const& ShadowRoot::part_element_map() const
...
+    exportparts->bytes_as_string_view().for_each_split_view(',', SplitBehavior::Nothing, [&](StringView mapping) {
+        auto trimmed = mapping.trim_whitespace();
+        if (trimmed.is_empty())
+            return;
+
+        auto parts = trimmed.split_view(':', SplitBehavior::KeepEmpty);
+        if (parts.size() == 1) {
+            auto name = MUST(FlyString::from_utf8(parts[0].trim_whitespace()));
+            result.append({ name, name });
+        } else {
```
AtkinsSJ: We probably want to check `parts.size() == 2` here, so that we don't treat `exportparts="foo:bar:baz"` the same as `exportparts="foo:bar"`.

Libraries/LibWeb/DOM/ShadowRoot.cpp
```diff
@@ -267,6 +267,34 @@ ShadowRoot::PartElementMap const& ShadowRoot::part_element_map() const
     return m_part_element_map;
 }
 
+// https://drafts.csswg.org/css-shadow-1/#exportparts
+// Parse the exportparts attribute into a list of (inner_name, outer_name) pairs.
+static Vector<Array<FlyString, 2>> parse_exportparts_attribute(Element const& element)
```
AtkinsSJ: Using a named struct instead of an Array for the pairs would be more in line with our code style. Something like
```c++
struct ExportedPart {
    FlyString inner_name;
    FlyString outer_name;
};
```

Libraries/LibWeb/DOM/ShadowRoot.cpp
```diff
@@ -280,22 +308,29 @@ void ShadowRoot::calculate_part_element_map()
...
+        if (element.is_shadow_host()) {
+            auto inner_root = element.shadow_root();
+
+            // 3. Calculate innerRoot’s part element map.
+            auto const& inner_map = inner_root->part_element_map();
+
+            // 4. For each innerName/outerName in el’s forwarded part name list:
+            for (auto const& mapping : parse_exportparts_attribute(element)) {
+                auto const& inner_name = mapping[0];
+                auto const& outer_name = mapping[1];
```
AtkinsSJ: ...and then I think you could do `for (auto const& [inner_name, outer_name] : parse_exportparts_attribute(element)) {` here.

Libraries/LibWeb/CSS/Parser/SelectorParsing.cpp
```diff
@@ -158,17 +158,35 @@ Parser::ParseErrorOr<NonnullRefPtr<Selector>> Parser::parse_complex_selector(Tok
...
-            .description = "Multiple pseudo elements in a selector is not yet supported."_string,
-        });
-        return ParseError::SyntaxError;
+        // FIXME: Other pseudo-elements can also be chained (e.g. ::highlight()::before).
+        //        For now, only ::part() followed by one other pseudo-element is supported.
+        bool is_valid_chain = first_pseudo_element.has_value()
+            && first_pseudo_element.value() == PseudoElement::Part
+            && second_pseudo_element.has_value()
+            && second_pseudo_element.value() != PseudoElement::Part
+            && pseudo_element_count == 2;
```
AtkinsSJ: `first_pseudo_element.has_value()` and `second_pseudo_element.has_value()` are both guaranteed to be true at this point.

Libraries/LibWeb/CSS/SelectorEngine.cpp
AtkinsSJ: I think you're right, shadow stuff melts my brain a bit. :sweat_smile: I originally thought we might need to traverse into descendant shadow roots for the `exportparts` case, but I was wrong on that.

Libraries/LibWeb/DOM/ShadowRoot.cpp
adamjcolvin: Yeah makes sense. I've changed the `else` branch to `else if (parts.size() == 2)`

Libraries/LibWeb/DOM/ShadowRoot.cpp
adamjcolvin: Done, I've added the struct above the function definition, that's much nicer.

Libraries/LibWeb/DOM/ShadowRoot.cpp
adamjcolvin: Done 👍

Libraries/LibWeb/CSS/Parser/SelectorParsing.cpp
adamjcolvin: Ah you're right, I've removed those checks.

Libraries/LibWeb/CSS/SelectorEngine.cpp
adamjcolvin: It hurts my brain too 😅

adamjcolvin:
Thanks @AtkinsSJ! I've addressed your comments and amended the relevant commits.

Libraries/LibWeb/DOM/ShadowRoot.cpp
```diff
@@ -290,6 +290,39 @@ ShadowRoot::PartElementMap const& ShadowRoot::part_element_map() const
...
+// Parse the exportparts attribute into a list of (inner_name, outer_name) pairs.
+struct ExportedPart {
+    FlyString inner_name;
+    FlyString outer_name;
+};
+
+static Vector<ExportedPart> parse_exportparts_attribute(Element const& element)
+{
+    Vector<ExportedPart> result;
+    auto exportparts = element.get_attribute("exportparts"_fly_string);
```
AtkinsSJ: Something I missed before: Attribute names are defined in `HTML/AttributeNames.h`, in a big macro. Please add `exportparts` to that and use it here.

Libraries/LibWeb/DOM/ShadowRoot.cpp
```diff
@@ -290,6 +290,39 @@ ShadowRoot::PartElementMap const& ShadowRoot::part_element_map() const
...
+    if (!exportparts.has_value())
+        return result;
+
+    exportparts->bytes_as_string_view().for_each_split_view(',', SplitBehavior::Nothing, [&](StringView mapping) {
```
AtkinsSJ: `bytes_as_string_view()` is a little dubious - `StringView` doesn't care about multi-byte characters so it's not necessarily safe. `exportparts->code_points()` will give you a `Utf8View` which is safer, and still has a `for_each_split_view()` method, though the parameters are a little different.

Libraries/LibWeb/DOM/ShadowRoot.cpp
adamjcolvin: Ah thanks, I completely missed that. I've updated.

Libraries/LibWeb/DOM/ShadowRoot.cpp
adamjcolvin: Yeah that makes sense, I've updated the commit 👍

AtkinsSJ:
This is failing a few `::part()` tests.

adamjcolvin:
> This is failing a few `::part()` tests.

@AtkinsSJ there were a few new test failures after rebasing, relating to a change around `:host::part()` matching. 

Previously, we had a check in SelectorEngine that was checking if the host of each candidate shadow root sits in the same scope as the rule. 

```cpp
// https://drafts.csswg.org/css-shadow-1/#part-element-map
// "The descendants of an element [...] does not include the shadow trees of the element."
if (ancestor_shadow_root->host()->containing_shadow_root() != context.rule_shadow_root)
  continue;
...
```

That worked for rules that live *outside* a shadow tree and pierce into a *direct child* tree, but for `:host::part()` the rule and the part both live in the *same* shadow root.

The fix was to skip that scope check when `:host` is present in the compound selector. I've added a boolean to the MatchContext to check for that:  

```cpp
 // https://drafts.csswg.org/css-shadow-1/#part-element-map
 // "The descendants of an element [...] does not include the shadow trees of the element."
bool const is_direct_child_scope = ancestor_shadow_root->host()->containing_shadow_root() == context.rule_shadow_root;
bool const is_host_part_own_scope = ancestor_shadow_root == context.rule_shadow_root && context.for_host_part_matching;
if (!is_direct_child_scope && !is_host_part_own_scope)
  continue;

```

I've updated the [relevant commit](https://github.com/LadybirdBrowser/ladybird/pull/7652/changes/43c04d65424d7c36eb3368a9313b1c3657ab6bc7) (`LibWeb: Only match ::part() in direct child shadow trees`), and we seem to be in good shape now.

AtkinsSJ:
Hmm. Had some maybe-unrelated screenshot test fails on Linux/arm64. Test artifact upload is broken right now so I've restarted it, and these are the tests that failed for reference:
```
Fail: Screenshot/input/ol-render-node-append-group-reversed.html
Fail: Screenshot/input/image-downscaling.html
Fail: Screenshot/input/nested-boxes-with-hidden-overflow-and-border-radius.html
```

PR #7656, opened 2026-01-28
LibDatabase: Enable the WAL and set the synchronous pragma to normal

trflynn89:
Let's do this by default. The APIs to override these are provided if needed.

In a simple `localStorage.setItem` loop:

```js
for (let i = 0; i < 100; ++i) {
    localStorage.setItem(`item_${i}`, `value_${i}`);
}
```

This reduces the runtime from ~3.5 seconds to ~50 milliseconds on my machine.

### DIFF
--- a/Libraries/LibDatabase/Database.cpp
+++ b/Libraries/LibDatabase/Database.cpp
@@ -62,7 +62,13 @@ ErrorOr<NonnullRefPtr<Database>> Database::create(ByteString const& directory, S
     sqlite3* m_database { nullptr };
     SQL_TRY(sqlite3_open(database_file.characters(), &m_database));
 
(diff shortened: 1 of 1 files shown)


Libraries/LibDatabase/Database.h
```diff
@@ -42,6 +42,26 @@ class DATABASE_API Database : public RefCounted<Database> {
     template<typename ValueType>
     ValueType result_column(StatementID, int column);
 
+    // https://www.sqlite.org/pragma.html#pragma_journal_mode
+    enum class JournalMode {
+        Delete,
+        Truncate,
+        Persist,
+        Memory,
+        WAL,
```
nico: WriteAheadLog?

PR #7662, opened 2026-01-28
AK: A little cleanup, perf improvement & bugfix

gmta:
* Remove unused and probably incorrect HashMap::hash()
* Make Vector::remove_all_matching O(n) instead of O(n²)
* Fix bug in String::from_utf16_be_with_replacement_character()

Running `js-benchmarks` on macOS revealed no significant changes in performance.


Hendiadyoin1:
@nico [thakisbot] !backport

(Maybe, otherwise I might get to it this late evening or tomorrow)

PR #7667, opened 2026-01-29
LibWeb: Avoid subtree iteration when assigning slottables

tcl3:
This change introduces SlotRegistry to track slot elements per shadow root. This allows us to iterate slots directly when assigning slottables for a tree instead of walking an entire subtree.

This could probably be made faster by using a better data structure to maintain tree order, but the insertion overhead doesn't seem to be a big performance bottleneck in real world use cases

This reduces the time taken to display the main table on https://wpt.fyi from 36 seconds to 24 seconds on my machine, so there's definitely more work to do, but this is a step in the right direction.

EDIT: The improvement can be clearly seen by putting `REPORT_TIME_EVERY()` at the top of `assign_slottables_for_a_tree()`. Results when loading https://wpt.fyi:

Before:
[Timing info for: "assign_slottables_for_a_tree"] current: 258.763µs | average: 86.061µs  | total: 12.995s    | calls: 151000

After:
[Timing info for: "assign_slottables_for_a_tree"] current: 20ns       | average: 29ns       | total: 4.478ms    | calls: 151000

### DIFF
--- a/Libraries/LibWeb/DOM/Slottable.cpp
+++ b/Libraries/LibWeb/DOM/Slottable.cpp
@@ -202,6 +202,15 @@ void assign_slottables(GC::Ref<HTML::HTMLSlotElement> slot)
     if (slottables != slot->assigned_nodes_internal())
         signal_a_slot_change(slot);
 
+    // AD-HOC: Clear the assigned slot for slottables that are no longer assigned to this slot.
+    //         This must happen before setting the new assigned slots to avoid stale references
+    //         during style computation.
+    for (auto const& old_slottable : slot->assigned_nodes_internal()) {
+        old_slottable.visit([&](auto const& node) {
+            node->set_assigned_slot(nullptr);
+        });
+    }
+
     // 4. For each slottable in slottables, set slottable’s assigned slot to slot.
     for (auto& slottable : slottables) {
         slottable.visit([&](auto& node) {
--- a/Tests/LibWeb/Text/expected/wpt-import/dom/nodes/moveBefore/slotchange-events.txt
+++ b/Tests/LibWeb/Text/expected/wpt-import/dom/nodes/moveBefore/slotchange-events.txt
@@ -1,0 +1,9 @@
+Harness status: OK
+
+Found 4 tests
+
+4 Pass
+Pass	Moving default content into a slot fires 'slotchange' event
+Pass	Moving default content out of a slot fires 'slotchange' event
+Pass	Moving a slottable into and out out of a custom element fires 'slotchange' event
+Pass	Moving a slot runs the assign slottables algorithm
--- a/Tests/LibWeb/Text/input/wpt-import/dom/nodes/moveBefore/slotchange-events.html
+++ b/Tests/LibWeb/Text/input/wpt-import/dom/nodes/moveBefore/slotchange-events.html
@@ -1,0 +1,139 @@
+<!DOCTYPE html>
+<title>slotchanged event</title>
+<script src="../../../resources/testharness.js"></script>
+<script src="../../../resources/testharnessreport.js"></script>
+<body>
+
+<script>
+customElements.define(
+  "custom-element",
+  class extends HTMLElement {
+    constructor() {
+      super();
+      const shadowRoot = this.attachShadow({mode: "open"});
+      const slot = document.createElement('slot');
+      slot.name = 'content';
+      shadowRoot.append(slot);
+    }
+  },
+);
+
+promise_test(async t => {
+  const customElement = document.body.appendChild(document.createElement('custom-element'));
+  t.add_cleanup(() => customElement.remove());
+
+  const slot = customElement.shadowRoot.children[0];
+  const slotChangePromise = new Promise((resolve, reject) => {
+    slot.addEventListener('slotchange', e => resolve(), {once: true});
+    t.step_timeout(() => reject('Timeout; slotchange was not fired'), 1500);
+  });
+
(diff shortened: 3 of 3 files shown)


Libraries/LibWeb/DOM/Node.cpp
```diff
@@ -1291,6 +1309,15 @@ WebIDL::ExceptionOr<void> Node::move_node(Node& new_parent, Node* child)
     });
 
     if (has_descendent_slot) {
+        // AD-HOC: Unregister any slot elements in the removed subtree from the shadow root's slot registry
+        //         before running assign_slottables_for_a_tree, so the registry is up-to-date.
+        if (auto* shadow_root = as_if<ShadowRoot>(old_parent_root)) {
```
shannonbooth: For this and `Node::remove()`, I think we can combine into the loop above so we only do a single pass

Libraries/LibWeb/DOM/SlotRegistry.cpp
```diff
@@ -0,0 +1,57 @@
...
+    });
+    if (slot_removed)
+        --m_slot_count;
+
+    // Insert in tree order.
+    if (m_slots.is_empty() || m_slots.last()->is_before(slot)) {
+        m_slots.append(GC::Weak { slot });
+    } else {
+        m_slots.insert_before_matching(GC::Weak { slot }, [&](auto& other_slot) {
+            return slot.is_before(*other_slot);
```
gmta: This could potentially become _very_ slow since we're iterating over part of the DOM tree for each registered slot that's before the slot we're trying to add.

Maybe this alternative approach could work as well?
1. Find the next slot in tree order from the slot we're trying to add
2. If no such slot exists, append to `m_slots`
3. Otherwise, insert before that slot in `m_slots`

tcl3: Great idea, I've implemented this.

I verified the performance improvement with [this benchmark](https://gist.github.com/tcl3/da6a76002e0f118cedb7628004c50a1e):

Before:
```
100 slots: 0.76ms avg
200 slots: 6.28ms avg
400 slots: 79.14ms avg
800 slots: 921.40ms avg
```

After:
```
100 slots: 0.24ms avg
200 slots: 0.64ms avg
400 slots: 2.34ms avg
800 slots: 9.96ms avg
```

Libraries/LibWeb/DOM/Slottable.cpp
```diff
@@ -227,14 +217,21 @@ void assign_slottables(GC::Ref<HTML::HTMLSlotElement> slot)
...
-    //         nodes, especially on sites with many elements. So we skip it if we know it's going to be a no-op anyways.
-    if (!root->is_shadow_root() && !root->is_html_slot_element())
+    auto* shadow_root = as_if<ShadowRoot>(*root);
+    if (!shadow_root && !root->is_html_slot_element())
         return;
 
     // To assign slottables for a tree, given a node root, run assign slottables for each slot of root’s inclusive
     // descendants, in tree order.
+    if (shadow_root) {
+        if (!shadow_root->has_registered_slots())
```
shannonbooth: I think has_registered_slots can be removed since for_each_registered_slot will just not iterate if something is missing

BTW with that AD-HOC comment being removed, does that fix https://wpt.live/dom/nodes/moveBefore/slotchange-events.html ? If I remember correctly that test was failing due to that optimization which is incorrect in the moveBefore case

Libraries/LibWeb/DOM/Slottable.cpp
```diff
@@ -90,17 +90,7 @@ GC::Ptr<HTML::HTMLSlotElement> find_a_slot(Slottable const& slottable, OpenFlag
...
-    shadow->for_each_in_subtree_of_type<HTML::HTMLSlotElement>([&](auto& child) {
-        if (child.slot_name() != slottable_name)
-            return TraversalDecision::Continue;
-
-        slot = child;
-        return TraversalDecision::Break;
-    });
-
-    return slot;
+    return shadow->first_slot_with_name(slottable_name);
```
shannonbooth, line 93: This makes me wonder if we should also make slottable_name() a FlyString for this lookup

Libraries/LibWeb/DOM/Slottable.cpp
tcl3: My latest push gets rid of `has_registered_slots`.

I've made a separate fix for the test you mentioned. This was crashing because `m_assigned_slot` wasn't being cleared when a slot was moved out of its shadow root and the slottable became unassigned.

Libraries/LibWeb/DOM/Slottable.cpp
tcl3: I've now changed slot and slottable names to be `FlyString`.

Libraries/LibWeb/DOM/Slottable.cpp
shannonbooth: Oh interesting! The crash must have been introduced after i investigated last (before i only remember a timeout).

What I was mostly trying to refer to is about removing the comment explaining the optimization since it does have a functional impact compared against the specification and results in the AD-hoc comments elsewhere

tcl3: > What I was mostly trying to refer to is about removing the comment explaining the optimization since it does have a functional impact compared against the specification and results in the AD-hoc comments elsewhere

Oh, I see! I've added an `OPTIMIZATION` comment explaining the early return.

PR #7670, opened 2026-01-29
LibWeb: Visual improvements for selections

gmta:
* Make the selection highlight color 50% transparent, mimicking other browsers. Feels a bit more fancy :)
* Properly align adjacent selection rects so they don't unnecessarily overlap
* Track trailing whitespace for wrapped lines so we can make that whitespace show up when selecting those lines

| | |
|--|--|
| Before | Screenshot 2026-01-29 at 10 46 48 |
| After | image |
| Chrome | Screenshot 2026-01-29 at 10 48 08 |
| Firefox | image |
| Safari (🤔) | image |

* Apply selection highlight on images

| Before | After |
|--|--|
| image | image |

* Support the `::selection` pseudo-element for the properties `color`, `background-color`, `text-shadow` and `text-decoration`

| Before | After |
|--|--|
| Screenshot 2026-02-04 at 15 06 20 | Screenshot 2026-02-04 at 15 05 47 |

### DIFF
--- a/Libraries/LibWeb/Painting/Paintable.cpp
+++ b/Libraries/LibWeb/Painting/Paintable.cpp
@@ -306,11 +306,12 @@ void Paintable::set_needs_paint_only_properties_update(bool needs_update)
 }
 
 // https://drafts.csswg.org/css-pseudo-4/#highlight-styling
-// FIXME: Support additional ::selection properties: text-decoration, stroke-color, fill-color, stroke-width.
+// FIXME: Support additional ::selection properties: text-underline-offset, text-underline-position, stroke-color,
+//        fill-color, stroke-width, and CSS custom properties.
 Paintable::SelectionStyle Paintable::selection_style() const
 {
     auto color_scheme = computed_values().color_scheme();
-    SelectionStyle default_style { CSS::SystemColor::highlight(color_scheme), {}, {} };
+    SelectionStyle default_style { CSS::SystemColor::highlight(color_scheme), {}, {}, {} };
 
     // For text nodes, check the parent element since text nodes don't have computed properties.
     auto node = dom_node();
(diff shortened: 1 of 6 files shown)


awesomekling:
Do we get a before/after of this fancy visual improvement :D

gmta:
> Do we get a before/after of this fancy visual improvement :D

You do now!

gmta:
Our behavior w/ respect to the trailing white space on wrapped lines is clearly different from other browsers... let me do something about that as well.

gmta:
> Our behavior w/ respect to the trailing white space on wrapped lines is clearly different from other browsers... let me do something about that as well.

Implemented this as well - makes trailing spaces visible in selections!

Libraries/LibWeb/Painting/Paintable.cpp
```diff
@@ -298,4 +305,65 @@ void Paintable::set_needs_paint_only_properties_update(bool needs_update)
...
+        auto context = CSS::ColorResolutionContext::for_layout_node_with_style(*element_layout_node);
+
+        SelectionStyle style;
+        style.background_color = computed_selection_style->color_or_fallback(CSS::PropertyID::BackgroundColor, context, Color::Transparent);
+
+        // Only use text color if it was explicitly set in the ::selection rule, not inherited.
+        if (!computed_selection_style->is_property_inherited(CSS::PropertyID::Color)) {
+            auto text_color = computed_selection_style->color_or_fallback(CSS::PropertyID::Color, context, Color::Transparent);
+            if (text_color.alpha() > 0)
+                style.text_color = text_color;
```
AtkinsSJ: Both Chromium and Firefox allow `::selection { color: transparent }` to make selected text invisible.

Libraries/LibWeb/Painting/Paintable.cpp
```diff
@@ -298,4 +305,65 @@ void Paintable::set_needs_paint_only_properties_update(bool needs_update)
...
+        return style;
+    };
+
+    // Check the element itself.
+    if (auto style = style_from_element(*element); style.has_value())
+        return style.release_value();
+
+    // If inside a shadow tree, check the shadow host. This enables ::selection styling on elements like <input> to
+    // apply to text rendered inside their shadow DOM.
+    if (auto shadow_root = element->containing_shadow_root()) {
```
AtkinsSJ: Should this only apply to UA-created shadow trees? eg,
```c++
if (auto shadow_root = element->containing_shadow_root(); shadow_root && shadow_root->is_user_agent_internal()) {
```

Libraries/LibWeb/Painting/Paintable.cpp
```diff
@@ -352,10 +353,18 @@ Paintable::SelectionStyle Paintable::selection_style() const
             style.text_shadow = move(shadows);
         }
 
-        // Only return a style if there's a meaningful customization (non-transparent background, explicit text color,
-        // or text-shadow). This allows us to continue checking shadow hosts when the current element only has UA
-        // default styles.
-        if (style.background_color.alpha() == 0 && !style.text_color.has_value() && !style.text_shadow.has_value())
+        // Only use text-decoration if it was explicitly set in the ::selection rule, not inherited.
+        if (!computed_selection_style->is_property_inherited(CSS::PropertyID::TextDecorationLine)) {
```
AtkinsSJ, line 354: FYI: `text-decoration-line` is not supposed to be inherited at all, that's just a hack we have. Doing that properly seems related enough to what you're working on here that I thought it was worth mentioning. :yakbait:

Libraries/LibWeb/Painting/Paintable.cpp
gmta: Good catch - I've added a test for this and changed the condition.

Libraries/LibWeb/Painting/Paintable.cpp
gmta: Definitely!

Libraries/LibWeb/Painting/Paintable.cpp
gmta: An attempt to make it uninherited pretty much exploded in complexity, so we might need to give that one another try in a separate PR :)

AtkinsSJ (approved):

PR #7673, opened 2026-01-29
LibWeb+LibWebView: Some cookie-related busy work

trflynn89:
This does a bit of cleanup, and updates our cookie implementation to the latest RFC draft.

### DIFF
--- a/Libraries/LibWeb/Cookie/Cookie.cpp
+++ b/Libraries/LibWeb/Cookie/Cookie.cpp
@@ -149,6 +149,46 @@ String default_path(URL::URL const& url)
     return MUST(String::from_utf8(uri_path.substring_view(0, last_separator)));
 }
 
+// https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-22#section-5.8.3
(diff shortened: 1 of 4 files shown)


Libraries/LibWeb/Cookie/Cookie.cpp
```diff
@@ -57,23 +57,48 @@ SameSite same_site_from_string(StringView same_site_mode)
...
+
+    // * All of the following conditions hold:
+    //   - The domain string is a suffix of the string.
+    if (!string.ends_with(domain_string))
+        return false;
+    //   - The last character of the string that is not included in the domain string is a %x2E (".") character.
+    if (string[string.length() - domain_string.length() - 1] != '.')
+        return false;
+    //   - The string is a host name (i.e., not an IP address).
+    if (AK::IPv4Address::from_string(string).has_value())
```
rgret-dev, line 97: We should check for IPv6 too, probably.

trflynn89: True, I think the code here is some years older than any of our IPv6 support :older_man:  I've added a commit for v6 checks.

PR #7679, opened 2026-01-30
test-web: Always create test dumps for failed tests

trflynn89:
If tests fail, let's just always create the results page regardless of verbosity. This allows viewing logs afterwards, as well as uploading the test-dumps collection in CI.

As a result, this removes the `--dump-failed-ref-tests` flag. It's a bit overloaded with our new results format, and it would be awkward to keep
both separately working here.


trflynn89:
Working here:
https://github.com/LadybirdBrowser/ladybird/actions/runs/21500794468/job/61946496286?pr=7679#step:17:1
https://github.com/LadybirdBrowser/ladybird/actions/runs/21500794468/artifacts/5312561034

Issue #7195, opened 2025-12-20
List inside table caption has incorrect vertical spacing


tcl3:
When putting a list inside a table caption, the list doesn't have the correct amount of vertical space.

Reduction from https://www.bamsoftware.com/hacks/zipbomb/:
```html
<!DOCTYPE html>
<table>
    <caption>
        <ul>
            <li>List item</li>
        </ul>
    </caption>
</table>
<p>Bottom text</p>
```

| Ladybird | Firefox |
| --- | --- |
| <img width="611" height="226" alt="Image" src="https://github.com/user-attachments/assets/dc6b31f9-51fe-4cfd-8783-0451a334303f" /> | <img width="611" height="199" alt="Image" src="https://github.com/user-attachments/assets/954f6a82-0d44-4e30-8a91-bc4dc5e64c91" />|

PR #7687, opened 2026-01-30
Fix incorrect vertical spacing

ycarmon:
Fixes incorrect vertical spacing when table captions contain block-level children like `<ul>` or `<ol>`.
The issue was that captions reported `content_height` of 0 when containing block children, causing the layout to not account for their visual space. 
The fix calculates the actual visual height by finding the bottom-most child position
Screenshot 2026-01-30 at 7 03 39 PM

This fixes https://github.com/LadybirdBrowser/ladybird/issues/7195


Libraries/LibWeb/Layout/TableFormattingContext.cpp
```diff
@@ -68,13 +68,30 @@ CSSPixels TableFormattingContext::run_caption_layout(CSS::CaptionSide phase, Ava
         }
 
         auto const& caption_state = m_state.get(child_box);
+
+        CSSPixels visual_bottom = caption_state.offset.y() + caption_state.border_box_height();
+        bool has_block_children = false;
+
+        for (auto* child = child_box.first_child(); child; child = child->next_sibling()) {
+            if (child->is_absolutely_positioned() || child->is_floating())
```
Psychpsyo: Shpuld probably use is_out_of_flow() here.

PR #7695, opened 2026-01-30
Everywhere: Enable unity build V2

R-Goc:
This PR aims to enable unity builds throughout. Here are some build timings. More are in #browser on discord.
```
Batch sizes refer only to LibWeb
--
Before any changes:
Full build:  15m 23.79s
ninja ladybird: 13m 40s
Touch HTML/HTMLInputElement.cpp 9.77s
Touch HTML/HTMLInputElement.h 27.24s
Touch HTML/HTMLMediaElement.h 7m 29.87s
 
Unity build:
Batch of 8:
full build: 6m 3.09s
ninja ladybird:  4m 47.91s
Touch HTML/HTMLInputElement.cpp 10.74s
Touch HTML/HTMLInputElement.h 25.78s
Touch HTML/HTMLMediaElement.h 3m 24.20s
 
Batch of 16:
Full build: 5m 16.70s
ninja ladybird: 4m 17.38s
Touch HTML/HTMLInputElement.cpp 12.27s
Touch HTML/HTMLInputElement.h 26.16s
Touch HTML/HTMLMediaElement.h 2m 47.98s
```
With a batch size of 32 an over 3x speedup is possible, but single file change build times sligthly suffer. On lower batch sizes there is a speedup even for these changes as the linker needs to read in less files than without a unity build.

### DIFF
--- a/Libraries/LibWeb/CMakeLists.txt
+++ b/Libraries/LibWeb/CMakeLists.txt
@@ -1209,6 +1209,12 @@ set(GENERATED_SOURCES
 
 ladybird_lib(LibWeb web EXPLICIT_SYMBOL_EXPORT)
 
+set_target_properties(LibWeb PROPERTIES
+    UNITY_BUILD ${LADYBIRD_ENABLE_UNITY_BUILD}
+    UNITY_BUILD_UNIQUE_ID "LIBWEB_UNITY_ID"
+    UNITY_BUILD_BATCH_SIZE 16
+)
+
 find_package(LibXml2 REQUIRED)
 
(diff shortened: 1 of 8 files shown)


gmta:
Hi @R-Goc!

As discussed on Discord, the maintainers were not exactly overly excited about introducing a unity build system and it's not that different this time. A unity build imposes new restrictions in how we organize the code, it requires additional workarounds (like `LIBWEB_UNITY_ID` from this PR), and it mainly helps people who are not continuously compiling the project while hindering the people who _are_ doing so. It also can require significantly more RAM per compilation process raising the minimum system requirements to compile the project.

Some hands-on experience with the WebKit unity build has shown the practical negative aspects - RAM eating processes, needing to compile many units for just a small change, etc. We might not fully see that at our current size right now, but we will get there at some point.

For me personally this comes down to: does this solve a real problem we currently have? And I don't think I can find the evidence for that, while I do hear a lot about negative experiences people have had.

If you want to continue work on this, make sure to discuss on Discord, get a definitive YES/NO on an approach from multiple maintainers so you'll have some support in getting it reviewed. Closing the PR as there's no intention of reviewing it further or merging it at the moment.

PR #7697, opened 2026-01-31
LibWeb: Populate the data property for input events

tcl3:
The `data` property is now set for input events, following this table from the spec for event types that we currently support: https://w3c.github.io/input-events/#overview

This makes typing in http://monkeytype.com work as expected.
There seems to be a separate issue with focus on this site - I need to click on the main test area before I begin typing, but that is outside the scope of this PR.

### DIFF
--- a/Libraries/LibWeb/HTML/FormAssociatedElement.cpp
+++ b/Libraries/LibWeb/HTML/FormAssociatedElement.cpp
@@ -862,7 +862,14 @@ void FormAssociatedTextControlElement::handle_insert(FlyString const& input_type
     MUST(set_range_text(data_for_insertion, selection_start, selection_end, Bindings::SelectionMode::End));
 
     text_node->invalidate_style(DOM::StyleInvalidationReason::EditingInsertion);
-    did_edit_text_node(input_type);
+
+    // The input event's data attribute is only set for certain input types according to:
+    // https://w3c.github.io/input-events/#overview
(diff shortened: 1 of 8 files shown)


Libraries/LibWeb/Page/EventHandler.cpp
```diff
@@ -1431,10 +1432,14 @@ EventResult EventHandler::handle_keydown(UIEvents::KeyCode key, u32 modifiers, u
         if (key == UIEvents::KeyCode::Key_Return && (significant_modifiers == UIEvents::Mod_None || significant_modifiers == UIEvents::Mod_Shift)) {
             auto input_type = significant_modifiers == UIEvents::Mod_Shift ? UIEvents::InputTypes::insertLineBreak : UIEvents::InputTypes::insertParagraph;
 
+            // Form controls always use insertLineBreak rather than insertParagraph.
+            if (as_if<HTML::FormAssociatedTextControlElement>(target)) {
```
gmta: `is<T>` instead of `as_if<T>`

tcl3: Oops, I forgot  to change this when I stopped using the value 😅

gmta: You accidentally the commit message tho

tcl3: Fixed.

PR #7698, opened 2026-01-31
LibWeb: Implement Document.hasFocus()

adamjcolvin:
## Summary

The `Document::has_focus()` method previously returned `true` unconditionally, which was incorrect for documents without a browsing context (such as those created via `document.implementation.createHTMLDocument()`). This implementation replaces the stub with the proper algorithm from the HTML specification's "has focus steps", allowing documents to correctly report whether they hold focus in the browsing context hierarchy.

Spec: https://html.spec.whatwg.org/multipage/interaction.html#has-focus-steps

## Implementation Approach

The implementation follows the spec with three main checks:

1. Return `false` if the document has no navigable (meaning it lacks a browsing context entirely).
2. Return `false` if the top-level traversable lacks system focus.
3. Walk the focus chain from the top-level traversable's active document to verify whether this document appears in the chain. Return `true` if found, `false` otherwise. When walking the chain, if a navigable container is encountered in the focused area with a non-null content navigable, advance to that navigable's active document.

```cpp
bool Document::has_focus() const
{
    auto navigable = this->navigable();
    if (!navigable)
        return false;

    auto traversable = navigable->traversable_navigable();
    if (!traversable || !traversable->is_focused())
        return false;

    auto candidate = traversable->active_document();
    while (candidate) {
        if (candidate == this)
            return true;

        auto focused_area = candidate->focused_area();
        if (auto* navigable_container = as_if<HTML::NavigableContainer>(focused_area.ptr())) {
            if (auto content_navigable = navigable_container->content_navigable()) {
                candidate = content_navigable->active_document();
                continue;
            }
        }
        return false;
    }
    return false;
}
```

## Changes

**Libraries/LibWeb/DOM/Document.cpp**: Implemented the "has focus steps" algorithm and added include for `NavigableContainer.h`.

**Tests/LibWeb/Text/expected/wpt-import/html/interaction/focus/document-level-focus-apis/document-level-apis.txt**: Updated expected results to reflect the now-passing test.

### DIFF
--- a/Libraries/LibWeb/DOM/Document.cpp
+++ b/Libraries/LibWeb/DOM/Document.cpp
@@ -131,6 +131,7 @@
 #include <LibWeb/HTML/MessageEvent.h>
 #include <LibWeb/HTML/MessagePort.h>
 #include <LibWeb/HTML/Navigable.h>
+#include <LibWeb/HTML/NavigableContainer.h>
 #include <LibWeb/HTML/Navigation.h>
 #include <LibWeb/HTML/NavigationParams.h>
 #include <LibWeb/HTML/Numbers.h>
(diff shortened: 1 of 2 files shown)


Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -3578,8 +3579,39 @@ bool Document::has_focus_for_bindings() const
 // https://html.spec.whatwg.org/multipage/interaction.html#has-focus-steps
 bool Document::has_focus() const
 {
-    // FIXME: Implement this algorithm.
-    return true;
+    // 1. If target's node navigable's top-level traversable does not have system focus, then return false.
+    auto navigable = const_cast<Document*>(this)->navigable();
```
gmta: 🤔 why do we need this `const_cast` here? Shouldn't `this->navigable()` work as well?

adamjcolvin: Ah you're right, `navigable()` is const already. Thanks, I've updated the commit.

PR #7699, opened 2026-01-31
LibWeb/Bindings: Various fixes for attribute setter IDL code generation

shannonbooth:
See commits for details. Noticed when reviewing the Fullscreen API changes that we had some issues in the IDL generator for attribute setters, and this fixes some of those.


shannonbooth:
https://github.com/WebAssembly/spec/issues/2075 is reported for the WPT change

PR #7707, opened 2026-01-31
Meta: Fix build failure on Python 3.15

atg:
argparse on Python 3.15 will be changed to prefer `-foo` over `-f` for attribute names. Double dash arguments are unaffected.

https://docs.python.org/3.15/library/argparse.html#dest

```
  File "<...>/ladybird/Meta/generate_public_suffix_data.py", line 149, in main
    generate_header_file(Path(args.h))
                              ^^^^^^
AttributeError: 'Namespace' object has no attribute 'h'
```

To be honest I couldn't figure out why there are single dash arguments in the first place. Is there a reason for it? Changing to double dashes would also fix the problem.

### DIFF
--- a/Meta/generate_encoding_indexes.py
+++ b/Meta/generate_encoding_indexes.py
@@ -215,21 +215,21 @@ namespace TextCodec {
 
 def main():
     parser = argparse.ArgumentParser(description="Generate text codec lookup tables", add_help=False)
-    parser.add_argument("-H", action="help", help="Show this help message and exit")
+    parser.add_argument("--help", action="help", help="Show this help message and exit")
     parser.add_argument(
-        "-h", "-generated-header-path", required=True, help="Path to the lookup table header file to generate"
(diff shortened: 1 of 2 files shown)


InvalidUsernameException:
Yeah, I can't see a reason why they would need to be single-dash arguments either: They are only invoked via their short form from what I can see, `git blame` doesn't have anything to say, and there is even a mix of kebab and snake case in the arguments. If the build still works afterwards, it is probably fine to change.

Non-maintainer opinion: I would just go the other way and use the long-form argument name for the destination. Short-form arguments are great for manually typing things in the command line, but when reading the code, `open(args.json_path, "r")` is much clearer than `open(args.j, "r")`.

atg:
I did a clean build with `print(sys.argv)` and as far as I can tell the long forms are never used, so I've just made everything double dashes. I also made --help long form only since having both -H and -h is confusing in a tool that only has 4 arguments.

PR #7712, opened 2026-02-01
LibWeb: Pass scope through :has() pseudo class

adamjcolvin:
## Summary

The `:scope` pseudo-class inside `:has()`, `:is()`, `:where()`, and `:not()` selectors was not receiving the scoping root from outer selector contexts like `Element.closest()`. This caused selectors like `:has(> :scope)` and `:is(.a :scope .c)` to fail when used with scoped queries.

Relevant specifications:
- [CSS Selectors Level 4 - :scope pseudo-class](https://drafts.csswg.org/selectors-4/#the-scope-pseudo)
- [CSS Selectors Level 4 - :has() relational pseudo-class](https://drafts.csswg.org/selectors-4/#relational)
- [DOM Standard - Element.closest()](https://dom.spec.whatwg.org/#dom-element-closest)

## Implementation Approach

The fix propagates the `scope` parameter through internal selector matching functions:

1. `matches_relative_selector()` - Added `GC::Ptr<DOM::ParentNode const> scope` parameter
2. `matches_has_pseudo_class()` - Added `GC::Ptr<DOM::ParentNode const> scope` parameter
3. Updated `:is()`, `:where()`, and `:not()` cases in `matches_pseudo_class()` to pass scope through to `matches()`

Key changes:

```cpp
// In matches_relative_selector() and matches_has_pseudo_class()
// Before
if (matches(selector, element, shadow_host, context, {}, {}, SelectorKind::Relative, anchor))
// After
if (matches(selector, element, shadow_host, context, {}, scope, SelectorKind::Relative, anchor))

// In :is()/:where()/:not() cases
// Before
if (matches(selector, element, shadow_host, context))
// After
if (matches(selector, element, shadow_host, context, {}, scope))
```

This ensures that when `:scope` is evaluated inside any of these pseudo-classes, it correctly refers to the scoping root from the outer context.

## Testing

This change fixes the following WPT tests:

- `dom/nodes/Element-closest.html` - Now 29 Pass, 0 Fail (was 28 Pass, 1 Fail). The test case `Element.closest with context node 'test4' and selector ':has(> :scope)'` now passes.

- `css/selectors/has-argument-with-explicit-scope.html` - Now 13 Pass, 0 Fail (was 12 Pass, 1 Fail). The comparison test `.a:has(:scope) .c and :is(.a :scope .c) returns same elements` now passes because both selectors correctly resolve `:scope`.

## Changes

- `Libraries/LibWeb/CSS/SelectorEngine.cpp` - Added scope parameter to `matches_relative_selector()` and `matches_has_pseudo_class()`, and updated `:is()`, `:where()`, and `:not()` to pass scope through.
- `Tests/LibWeb/Text/expected/wpt-import/dom/nodes/Element-closest.txt` - Updated expected results.
- `Tests/LibWeb/Text/expected/wpt-import/css/selectors/has-argument-with-explicit-scope.txt` - Updated expected results.

### DIFF
--- a/Tests/LibWeb/Text/expected/wpt-import/css/selectors/has-argument-with-explicit-scope.txt
+++ b/Tests/LibWeb/Text/expected/wpt-import/css/selectors/has-argument-with-explicit-scope.txt
@@ -2,12 +2,11 @@ Harness status: OK
 
 Found 13 tests
 
-12 Pass
-1 Fail
+13 Pass
 Pass	:has(:scope) matches expected elements on scope1
 Pass	:has(:scope .c) matches expected elements on scope1
 Pass	:has(.a :scope) matches expected elements on scope1
-Fail	.a:has(:scope) .c matches expected elements on scope1
+Pass	.a:has(:scope) .c matches expected elements on scope1
 Pass	.a:has(:scope) .c and :is(.a :scope .c) returns same elements on scope1
 Pass	.a:has(:scope) .c matches expected elements on scope2
 Pass	.a:has(:scope) .c and :is(.a :scope .c) returns same elements on scope2
--- a/Libraries/LibWeb/CSS/SelectorEngine.cpp
+++ b/Libraries/LibWeb/CSS/SelectorEngine.cpp
@@ -141,10 +141,10 @@ static inline bool matches_lang_pseudo_class(DOM::Element const& element, Vector
 }
 
 // https://drafts.csswg.org/selectors-4/#relational
-static inline bool matches_relative_selector(CSS::Selector const& selector, size_t compound_index, DOM::Element const& element, GC::Ptr<DOM::Element const> shadow_host, MatchContext& context, GC::Ref<DOM::Element const> anchor)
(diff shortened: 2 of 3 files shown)


Tests/LibWeb/Text/expected/wpt-import/css/selectors/has-argument-with-explicit-scope.txt
```diff
@@ -7,8 +7,8 @@ Found 13 tests
 Pass	:has(:scope) matches expected elements on scope1
 Pass	:has(:scope .c) matches expected elements on scope1
 Pass	:has(.a :scope) matches expected elements on scope1
-Fail	.a:has(:scope) .c matches expected elements on scope1
-Pass	.a:has(:scope) .c and :is(.a :scope .c) returns same elements on scope1
+Pass	.a:has(:scope) .c matches expected elements on scope1
+Fail	.a:has(:scope) .c and :is(.a :scope .c) returns same elements on scope1
```
gmta: 🤔 is this expected to fail now?

adamjcolvin: Ah sorry, I put that in while I was debugging and forgot that I had done it 🤦‍♂️ The test is comparing the results of `:has()` and `:is()`, so I needed to pass the scope through for `:is()`/`:where()`/`:not()` as well. I've updated the commit.

AtkinsSJ (approved):
Looks good, thanks!

PR #7714, opened 2026-02-01
ImageDecoder: security hardening for untrusted image inputs

### LINKED ISSUE #7713: ImageDecoder: add security hardening for untrusted image inputs
The ImageDecoder service processes externally supplied image data and
represents a critical security boundary. Image decoding is a
historically high-risk attack surface and should defensively handle
malformed or abusive inputs.

Currently, the ImageDecoder lacks explicit limits and validation for
certain resource-intensive cases, which could allow excessive CPU or
memory usage from malicious image payloads.

Suggested improvements include:
- Enforcing limits on encoded image size and animation frame count.
- Validating decoded image dimensions before allocation.
- Adding decode timeouts to prevent long-running operations.
- Limiting concurrent decode jobs per client.
- Hardening metadata parsing (e.g. EXIF) against invalid values.

These changes would improve robustness and resilience without altering
behavior for valid images.

Anujsharma002:
## Summary
Harden the ImageDecoder service with configurable limits and a decode timeout to reduce risk of abuse and runaway decodes.
Fixes #7713 
## Changes
- **Input limits:** Reject encoded buffers larger than 50 MiB and images with more than 1024 frames.
- **Dimension check:** Skip decoded frames whose width or height exceed 16,384 pixels.
- **Concurrency:** Cap concurrent decode jobs per client at 16; reject new requests with a clear error when at capacity.
- **Decode timeout:** Cancel decode jobs that run longer than 3 seconds via a single-shot timer (weak ref to the job).
- **Client disconnect:** In `die()`, cancel all pending jobs and, if the connection is still open, notify the client with "Client disconnected" for each pending image.
- **Result callback:** In the decode success callback, do not send results if the connection is no longer open.
- **Exif resolution:** Guard Exif x/y resolution with `std::isfinite` and positive checks before using them for scale.

Error strings are adjusted for consistency (e.g. "Invalid encoded buffer", "Image too large", "Too many concurrent decode requests").

### DIFF
--- a/Services/ImageDecoder/ConnectionFromClient.cpp
+++ b/Services/ImageDecoder/ConnectionFromClient.cpp
@@ -6,14 +6,24 @@
 
 #include <AK/Debug.h>
 #include <AK/IDAllocator.h>
+#include <AK/RefPtr.h>
 #include <ImageDecoder/ConnectionFromClient.h>
 #include <ImageDecoder/ImageDecoderClientEndpoint.h>
 #include <LibCore/System.h>
+#include <LibCore/Timer.h>
 #include <LibGfx/Bitmap.h>
 #include <LibGfx/ImageFormats/ImageDecoder.h>
 #include <LibGfx/ImageFormats/TIFFMetadata.h>
+#include <cmath>
 
 namespace ImageDecoder {
+
+// SECURITY LIMITS
+constexpr size_t MAX_ENCODED_IMAGE_SIZE = 50 * MiB;
+constexpr size_t MAX_FRAME_COUNT = 1024;
+constexpr int MAX_IMAGE_DIMENSION = 16384;
+constexpr size_t MAX_PENDING_JOBS = 16;
+constexpr int DECODE_TIMEOUT_MS = 3000;
 
 static HashMap<int, RefPtr<ConnectionFromClient>> s_connections;
 static IDAllocator s_client_ids;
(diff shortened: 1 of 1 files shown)


Services/ImageDecoder/ConnectionFromClient.cpp
```diff
@@ -6,15 +6,25 @@
...
 #include <LibGfx/Bitmap.h>
 #include <LibGfx/ImageFormats/ImageDecoder.h>
 #include <LibGfx/ImageFormats/TIFFMetadata.h>
+#include <cmath>
 
 namespace ImageDecoder {
 
+// SECURITY LIMITS
+constexpr size_t MAX_ENCODED_IMAGE_SIZE = 50 * MiB;
+constexpr size_t MAX_FRAME_COUNT = 1024;
```
circl-lastname, line 23: At 24 FPS that's 42.6 seconds, most browsers allow that

Anujsharma002: The limit is primarily a security hardening measure rather than a playback constraint. While browsers allow ~40s animations, untrusted decoders are a different threat model. That said, I’m open to increasing the cap or aligning it with existing limits if there’s a preferred standard.

Zaggy1024: As far as I'm aware, Firefox avoids issues with long animations by limiting the buffer of frames, not by limiting the number of frames they can decode/display. We need to rewrite our entire animation/decoding system to allow us to decode on demand and only buffer/reuse what fits in some (arbitrary) limit.

(I'm hoping to reuse some portions of the video playback system for this at some point, but I have yet to try to reuse frames in looping videos, which is a prerequisite to prevent wasted CPU cycles re-decoding animations over and over.)

Zaggy1024:
Capping the concurrent decodes at 16 seems a bit risky to me, I feel like we're very likely to hit that limit on gallery pages like Instagram. We can't necessarily expect not to hit any limit here no matter how high, so I feel like this needs a better solution if it is really a problem to queue up decoding tasks unbounded.

PR #7717, opened 2026-02-01
LibWeb+LibWebView+WebContent: Implement versioning for document cookies

trflynn89:
This patch introduces a cookie cache in the WebContent process to reduce blocking IPC calls when JS accesses `document.cookie`. The UI process now maintains a cookie version counter per-domain in shared memory. When JS reads `document.cookie`, we check whether we have a valid cached cookie by comparing the current shared version to the last used version. If they match, the cached cookie is returned without IPC.

This optimization is based on Chromium's shared versioning, in which it was observed that 87% of `document.cookie` accesses were redundant. See:
https://blog.chromium.org/2024/06/introducing-shared-memory-versioning-to.html

Note that this cache only supports `document.cookie`, not HTTP Cookie headers. HTTP cookies are attached to requests with varying URLs and paths. The cookies that match the document URL might not match the request URL, which we wouldn't know from WebContent. So attaching the cached document cookie would be incorrect.

On https://twinings.co.uk, we see approximately 600 `document.cookie` requests while the page loads. This patch reduces the time spent in the `document.cookie` getter from ~45ms to 2-3ms.

No WPT changes in the `/cookies` or `/cookiestore` subtests.

### DIFF
--- a/Libraries/LibWebView/ViewImplementation.cpp
+++ b/Libraries/LibWebView/ViewImplementation.cpp
@@ -48,7 +48,8 @@ Optional<ViewImplementation&> ViewImplementation::find_view_by_id(u64 id)
 }
 
 ViewImplementation::ViewImplementation()
-    : m_view_id(s_view_count++)
+    : m_document_cookie_version_buffer(Core::create_shared_version_buffer())
+    , m_view_id(s_view_count++)
 {
     s_all_views.set(m_view_id, this);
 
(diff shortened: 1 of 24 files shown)


Libraries/LibWebView/ViewImplementation.cpp
```diff
@@ -259,9 +259,38 @@ void ViewImplementation::set_preferred_motion(Web::CSS::PreferredMotion motion)
...
+    }
+
+    if (!cookies.is_empty())
+        client().async_cookies_changed(page_id(), cookies);
+}
+
+Core::AnonymousBuffer const& ViewImplementation::ensure_document_cookie_version_buffer(Badge<WebContentClient>, String const& domain)
+{
+    return m_document_cookie_version_buffers.ensure(domain, [&]() {
+        auto document_cookie_version = MUST(Core::AnonymousBuffer::create_with_size(sizeof(Web::Cookie::CookieVersion)));
```
gmta: Maybe a FIXME (if I'm not misunderstanding this) to address the fact that we're creating an anonymous buffer per domain, which can round up to the page size (e.g. 4 KiB) - we might be able to do something smarter and have multiple cookie versions per buffer.

trflynn89: Ah TIL. I've added a FIXME.

trflynn89: On second look, it was actually quite easy to just use a single buffer for multiple cookie versions. With a page size of 4KiB, we can encode 512 document cookie versions per page in a single buffer. If some site exceeds that, we just log it, and we can send them a strongly worded email asking what the heck

There ended up being enough facilities around the shared version that I put some helpers into LibCore.

PR #7720, opened 2026-02-02
LibWeb: A couple of simplifications to `NodeWithStyle::apply_style()`

Calme1709:
Cleans up a couple of FIXMEs and removes some duplicate logic. No functional changes (we don't yet respect propagated `rx`/`ry` values)


Libraries/LibWeb/Layout/Node.cpp
```diff
@@ -626,18 +626,10 @@ void NodeWithStyle::apply_style(CSS::ComputedProperties const& computed_style)
         VERIFY_NOT_REACHED();
```
AtkinsSJ: This verify is also useless and can be removed.

Calme1709: Done.

PR #7721, opened 2026-02-02
LibWeb: Exclude UA internal shadow roots from percentage height quirk

tcl3:
Previously, this calculation was being applied to UA internal shadow roots, such as `<input>`s,  which caused incorrect rendering in quirks mode.

There are probably other places where we need to distinguish UA internal shadow roots from other shadow roots - I plan to investigate further and make more PRs if necessary.

Found while looking at the minimal reproduction from #759

### DIFF
--- a/Libraries/LibWeb/Layout/BlockFormattingContext.cpp
+++ b/Libraries/LibWeb/Layout/BlockFormattingContext.cpp
@@ -11,6 +11,7 @@
 #include <LibWeb/CSS/PropertyID.h>
 #include <LibWeb/DOM/Element.h>
 #include <LibWeb/DOM/Node.h>
+#include <LibWeb/DOM/ShadowRoot.h>
 #include <LibWeb/Dump.h>
 #include <LibWeb/Layout/BlockContainer.h>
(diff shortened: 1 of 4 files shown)


awesomekling (approved):

Libraries/LibWeb/DOM/ShadowRoot.h
```diff
@@ -111,6 +114,7 @@ class WEB_API ShadowRoot final : public DocumentFragment {
     Bindings::SlotAssignmentMode m_slot_assignment { Bindings::SlotAssignmentMode::Named };
     bool m_delegates_focus { false };
     bool m_available_to_element_internals { false };
+    bool m_is_user_agent { false };
```
konradekk: it’d be better to align this with previous ones? e.g. `m_created_by_user_agent` (with `is_created_by_user_agent`/`set_created_by_user_agent`) would be fine?

tcl3: I've renamed this to `is_user_agent_internal()` to make it clearer.

PR #7722, opened 2026-02-02
LibWeb/CSS: Absolutize color style values

AtkinsSJ:
Does the expected stuff of simplifying `calc()` etc, but there are a few color-specific things too:
- `hsl()` and `hwb()` compute to an `rgb()` function instead.
- Keywords like `red` or `buttontext` compute to their `rgb()` equivalent.
- `light-dark()` computes to whichever of its parameters applies to the current `color-scheme`.
- `color-mix()` produces a mixed color if it has enough information to do so.

Related to #7542, but this PR doesn't touch the interpolation code itself.


Libraries/LibWeb/CSS/StyleValues/HSLColorStyleValue.cpp
```diff
@@ -25,6 +27,52 @@ Optional<Color> HSLColorStyleValue::to_color(ColorResolutionContext color_resolu
...
+    // hsl() computes to rgb()
+    // https://drafts.csswg.org/css-color-4/#resolving-sRGB-values
+    auto resolved_h = resolve_hue(absolutized_h, {});
+    auto resolved_s = resolve_with_reference_value(absolutized_s, 100.0, {});
+    auto resolved_l = resolve_with_reference_value(absolutized_l, 100.0, {});
+    auto resolved_alpha = resolve_alpha(absolutized_alpha, {});
+
+    // Fall back to returning an hsl() still if we can't compute completely
+    if (!resolved_h.has_value() || !resolved_s.has_value() || !resolved_l.has_value() || !resolved_alpha.has_value()) {
+        if (absolutized_h == m_properties.h && absolutized_s == m_properties.s && absolutized_l == m_properties.l && absolutized_alpha == m_properties.alpha)
```
Calme1709: Should this be a `VERIFY_NOT_REACHED()`? I can't think of a scenario where we wouldn't be able to compute  these - if there is one maybe we should have a comment explaining it?

Libraries/LibWeb/CSS/StyleValues/HWBColorStyleValue.cpp
```diff
@@ -38,6 +40,63 @@ Optional<Color> HWBColorStyleValue::to_color(ColorResolutionContext color_resolu
...
+    // hwb() computes to rgb()
+    // https://drafts.csswg.org/css-color-4/#resolving-sRGB-values
+    auto resolved_h = resolve_hue(absolutized_h, {});
+    auto resolved_w = resolve_with_reference_value(absolutized_w, 100.0, {});
+    auto resolved_b = resolve_with_reference_value(absolutized_b, 100.0, {});
+    auto resolved_alpha = resolve_alpha(absolutized_alpha, {});
+
+    // Fall back to returning an hwb() still if we can't compute completely
+    if (!resolved_h.has_value() || !resolved_w.has_value() || !resolved_b.has_value() || !resolved_alpha.has_value()) {
+        if (absolutized_h == m_properties.h && absolutized_w == m_properties.w && absolutized_b == m_properties.b && absolutized_alpha == m_properties.alpha)
```
Calme1709: Same as HSL - should this be a `VERIFY_NOT_REACHED()`?

Libraries/LibWeb/CSS/StyleValues/ColorMixStyleValue.cpp
```diff
@@ -201,4 +204,44 @@ Optional<Color> ColorMixStyleValue::to_color(ColorResolutionContext color_resolu
...
+        auto interpolated_color = interpolate_color(from_color.value(), to_color.value(), delta, ColorSyntax::Modern);
+        return ColorFunctionStyleValue::create(
+            "srgb"sv,
+            NumberStyleValue::create(interpolated_color.red() / 255.0f),
+            NumberStyleValue::create(interpolated_color.green() / 255.0f),
+            NumberStyleValue::create(interpolated_color.blue() / 255.0f),
+            NumberStyleValue::create(interpolated_color.alpha() / 255.0f));
+    }
+
+    // Fall back to returning a color-mix() with absolutized values if we can't compute completely
```
Calme1709: I think the only scenario this can happen for is if one of the parameters relies on `currentcolor`, either directly or via another function - I think we should note this.

Libraries/LibWeb/CSS/StyleValues/HSLColorStyleValue.cpp
AtkinsSJ: For this and the others: IIRC calc() can depend on values that aren't known until layout. eg, you could do something weird like `hsl(calc((1cqw/1cqh) * 1deg), 10%, 20%);` and that's not resolvable yet.

(It's such a pain.)

Calme1709: [The spec](https://drafts.csswg.org/css-anchor-position-1/#interleaving) states that container query lengths are resolved for computed values:
> ...A length like 10cqw is resolved into a [computed length](https://drafts.csswg.org/css-values-4/#computed-length) using layout information about the query container’s size...

and this behavior is tested by WPT [here](https://github.com/web-platform-tests/wpt/blob/67aedff670908a56c76e9e4ae901ee422e73030b/css/css-text/parsing/tab-size-computed.html#L30) among other places.

Since this does however rely on "Style & Layout Interleaving" which is so far undefined in the spec I'm not sure whether this actually conflicts with your statement, so consider this more of an FYI.

AtkinsSJ: OK. I plucked cq units out of the air as an example. :sweat_smile: 

Generally I thought we needed to store CalculatedStyleValues at times in ComputedValues for things like `border-width: calc(20% + 10px)`. And since calc() can make use of units/percentages that aren't directly part of the property itself, that we'd need to store the calc() for any values that contain something numeric. I could be completely wrong/outdated on that as we've got better about computing values.

Calme1709: AFAIK the only time we persist `CalculatedStyleValue` is when it can contain a layout relative percentage, in this case the percentage is disallowed for the H value and is resolved to the reference range [0,100] (i.e. isn't layout relative) - see the percentage section here: https://drafts.csswg.org/css-color-4/#the-hsl-notation

AtkinsSJ: That makes sense, I just wish these things were more consistent. :sweat_smile:

One future wrinkle is relative-color syntax, because you can do `hsl(from currentcolor ...)` and that's not computable. But that doesn't affect computing the parameters, we'd just have to have a fallback if the `from` color isn't computable.

AtkinsSJ:
Changes:
- Replaced the fallback for HSL and HWB with verifies as suggested.
- Added a note about `currentcolor` being the tricky case.
- Rebased on master as the recent `AccentColor` changes affect some of the code here (and we also get an improved test result).

PR #7723, opened 2026-02-02
LibWeb: Exclude UA internal shadow root elements in `elementFromPoint()`

tcl3:
When `elementFromPoint()` or `elementsFromPoint()` returns an element that is inside a UA internal shadow root, we now return the shadow host for that element.

### DIFF
--- a/Libraries/LibWeb/DOM/Document.cpp
+++ b/Libraries/LibWeb/DOM/Document.cpp
@@ -5755,6 +5755,17 @@ WebIDL::ExceptionOr<void> Document::set_design_mode(String const& design_mode)
     return {};
 }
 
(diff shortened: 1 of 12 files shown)


Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -5755,6 +5755,18 @@ WebIDL::ExceptionOr<void> Document::set_design_mode(String const& design_mode)
     return {};
 }
 
+static Element* retarget_from_ua_internal_shadow_root(Element* element)
+{
+    if (!element)
+        return nullptr;
```
AtkinsSJ: Both callers do a null check first so `element` always has a value.

PR #7731, opened 2026-02-02
LibWasm: Parse struct types and support multiple types in type section

mkanilsson:
This patch adds support for parsing structs in the type section.

It also removes the assumption that all types in the type section are
function types, adding appropriate validation.

Spec tests struct.3 and struct.4 have been disable as this would
require expanding `ValueType` to include more heap-types.

### DIFF
--- a/Libraries/LibWasm/AbstractMachine/BytecodeInterpreter.cpp
+++ b/Libraries/LibWasm/AbstractMachine/BytecodeInterpreter.cpp
@@ -2008,7 +2008,7 @@ HANDLE_INSTRUCTION(call_indirect)
     TRAP_IN_LOOP_IF_NOT(element.ref().template has<Reference::Func>());
     auto address = element.ref().template get<Reference::Func>().address;
     auto const& type_actual = configuration.store().get(address)->visit([](auto& f) -> decltype(auto) { return f.type(); });
-    auto const& type_expected = configuration.frame().module().types()[args.type.value()];
+    auto const& type_expected = configuration.frame().module().types()[args.type.value()].unsafe_function();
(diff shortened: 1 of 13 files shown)


Libraries/LibWasm/AbstractMachine/BytecodeInterpreter.cpp
```diff
@@ -2008,7 +2008,7 @@ HANDLE_INSTRUCTION(call_indirect)
     TRAP_IN_LOOP_IF_NOT(element.ref().template has<Reference::Func>());
     auto address = element.ref().template get<Reference::Func>().address;
     auto const& type_actual = configuration.store().get(address)->visit([](auto& f) -> decltype(auto) { return f.type(); });
-    auto const& type_expected = configuration.frame().module().types()[args.type.value()];
+    auto const& type_expected = configuration.frame().module().types()[args.type.value()].function();
```
alimpfard: hmm, this forces us to pay again for already-validated checks (that a function is present), maybe add `unsafe_function()` or something similar and use it here + immediately below in return_call_indirect?

mkanilsson: That has been fixed. This is now a `unsafe_function()` that uses `Variant::unsafe_get()` instead.

I did notice that `Variant:get()` and `Variant:unsafe_get()` are identical though

alimpfard: unsafe_get() should be using an ASSERT() where get() uses a VERIFY(), so the check is gone in release builds.

PR #7735, opened 2026-02-02
LibWebView+UI: Add a context menu item to download images

trflynn89:
https://github.com/user-attachments/assets/e1f79c74-8679-46d2-b095-de965b40f0c6

### DIFF
--- a/Libraries/LibWebView/ViewImplementation.cpp
+++ b/Libraries/LibWebView/ViewImplementation.cpp
@@ -891,6 +891,13 @@ void ViewImplementation::initialize_context_menus()
     m_open_image_action = Action::create("Open Image"sv, ActionID::OpenImage, [this]() {
         load(m_context_menu_url);
     });
+    m_save_image_action = Action::create("Save Image As..."sv, ActionID::SaveImage, [this]() {
+        auto download_path = Application::the().path_for_downloaded_file(m_context_menu_url.basename());
+        if (download_path.is_error())
+            return;
+
+        Application::the().file_downloader().download_file(m_context_menu_url, download_path.release_value());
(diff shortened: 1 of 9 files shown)


Libraries/LibWebView/ViewImplementation.cpp
```diff
@@ -32,6 +32,28 @@ namespace WebView {
 static HashMap<u64, ViewImplementation*> s_all_views;
 static u64 s_view_count = 1; // This has to start at 1 for Firefox DevTools.
 
+static ErrorOr<void> save_image(Gfx::Bitmap const& bitmap, LexicalPath const& path)
+{
+    auto encoded = TRY(Gfx::PNGWriter::encode(bitmap));
```
gmta: Yeah, about that `.webp`...

```
% file ./about.webp
./about.webp: PNG image data, 1024 x 1024, 8-bit/color RGBA, non-interlaced
```

Either please add a note that images are all converted to the PNG format (and ideally the resulting filename should reflect that) or use an approach that stores the image resource as received, which is what most people come to expect?

trflynn89: ha, big oops! I've made it download the file as-is:
```bash
❯ file about.webp
about.webp: RIFF (little-endian) data, Web/P image, VP8 encoding, 1024x1024, Scaling: [none]x[none], YUV color, decoders should clamp
```

trflynn89:
Edit: Found a bug where pressing cmd+w would close the file picker dialog, but leave the app in a modal, unresponsive state. This was an existing issue, but more prevalent now, so I added a commit to make us behave like Safari here.

PR #7736, opened 2026-02-03
UI/Qt: Add Ctrl+Tab shortcuts for tab cycling

Shahzeb-Khattak:
Added Ctrl+Tab and Ctrl+Shift+Tab shortcuts for cycling tabs, to be consistent with other browsers.


InvalidUsernameException:
Ctrl+(Shift+)Tab is already a thing for me in the current `master` version of the QT UI :thinking:. I remember it sometimes breaking though. Is it straight up never working for you?

Shahzeb-Khattak:
You're right, it does seem to work now, but it definitely wasn't working for me earlier.

InvalidUsernameException:
Maybe you can try to figure out what causes the shortcuts to break? I'm pretty sure there is some form of bug related to them them. Understanding better what triggers the problem might go a long way towards fixing it.

PR #7740, opened 2026-02-03
LibRegex: Ignore 'FailIfEmpty' in dot-star loop detection

alimpfard:


alimpfard:
Ooooh exciting, it actually breaks tests!

Libraries/LibRegex/RegexMatch.h
```diff
@@ -334,6 +335,62 @@ class RegexStringView {
             });
     }
 
+    FoundIndex find_end_of_line(size_t start_code_point_index, size_t start_code_unit_index) const
```
aplefull, line 338: Hmm, this returns the position of the newline, but later we pass it to `find_index_of_previous` as an exclusive upper bound. So when we are looking for line terminator itself, we'll never find it:
```"abc\ndef\n".match(/.*\n/g);``` is null. Maybe it should return one past newline?

**Review by aplefull — COMMENTED** [review]
Also noticed that this breaks, now that rewrite is working again: `"hi😀".match(/.*😀/);`. We can fix it by adding a proper utf16 scan in the `find_index_of_previous`, but I’m not sure whether this is the correct fix, or if it’s masking some other issue in `RSeekTo`
```diff
diff --git a/Libraries/LibRegex/RegexMatch.h b/Libraries/LibRegex/RegexMatch.h
index 80a4d60afc..6521b178f5 100644
--- a/Libraries/LibRegex/RegexMatch.h
+++ b/Libraries/LibRegex/RegexMatch.h
@@ -304,10 +304,27 @@ public:
     {
         return m_view.visit(
             [&](Utf16View const& view) -> Optional<FoundIndex> {
-                auto result = view.find_last_code_unit_offset(code_point, end_code_unit_index);
-                if (!result.has_value())
-                    return {};
-                return FoundIndex { result.value(), view.code_point_offset_of(result.value()) };
+                size_t code_unit_index = 0;
+                size_t cp_index = 0;
+                Optional<FoundIndex> found;
+                auto limit = min(end_code_unit_index, view.length_in_code_units());
+                while (code_unit_index < limit) {
+                    auto code_unit = view.code_unit_at(code_unit_index);
+                    u32 ch = code_unit;
+                    size_t code_units_for_this = 1;
+                    if (AK::UnicodeUtils::is_utf16_high_surrogate(code_unit) && code_unit_index + 1 < view.length_in_code_units()) {
+                        auto next_code_unit = view.code_unit_at(code_unit_index + 1);
+                        if (AK::UnicodeUtils::is_utf16_low_surrogate(next_code_unit)) {
+                            ch = AK::UnicodeUtils::decode_utf16_surrogate_pair(code_unit, next_code_unit);
+                            code_units_for_this = 2;
+                        }
+                    }
+                    if (ch == code_point)
+                        found = FoundIndex { code_unit_index, cp_index };
+                    code_unit_index += code_units_for_this;
+                    ++cp_index;
+                }
+                return found;
             },
             [&](StringView const& view) -> Optional<FoundIndex> {
                 if (unicode()) {
```

alimpfard:
There's still an issue with rseekto anyway, thanks for the extra test point :)

Libraries/LibRegex/RegexMatch.h
alimpfard: Oh true, nice catch!

PR #7741, opened 2026-02-03
LibWeb: Allow importing ChaCha20-Poly1305 JWK without alg

mikiubo:
The WebCrypto specification does not require the "alg" member to be present when importing a symmetric JWK, as long as the key material itself is valid.

Add tests covering JWK import without an "alg" field.

This fixes the following WPT:
WebCryptoAPI/import_export/ChaCha20-Poly1305_importKey

### DIFF
--- a/Libraries/LibWeb/Crypto/CryptoAlgorithms.cpp
+++ b/Libraries/LibWeb/Crypto/CryptoAlgorithms.cpp
@@ -9887,13 +9887,9 @@ WebIDL::ExceptionOr<GC::Ref<CryptoKey>> ChaCha20Poly1305::import_key(AlgorithmPa
             return WebIDL::DataError::create(m_realm, "Invalid key type"_utf16);
 
         // 3. If jwk does not meet the requirements of Section 6.4 of JSON Web Algorithms [JWA], then throw a DataError.
-        // Specifically, those requirements are:
-        // * the member "k" is used to represent a symmetric key (or another key whose value is a single octet sequence).
-        // * An "alg" member SHOULD also be present to identify the algorithm intended to be used with the key,
-        //   unless the application uses another means or convention to determine the algorithm used.
-        // NOTE: "k" is already checked in step 4.
-        if (!jwk.alg.has_value())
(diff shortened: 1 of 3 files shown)


Libraries/LibWeb/Crypto/CryptoAlgorithms.cpp
```diff
@@ -9888,12 +9888,8 @@ WebIDL::ExceptionOr<GC::Ref<CryptoKey>> ChaCha20Poly1305::import_key(AlgorithmPa
...
         // 3. If jwk does not meet the requirements of Section 6.4 of JSON Web Algorithms [JWA], then throw a DataError.
         // Specifically, those requirements are:
-        // * the member "k" is used to represent a symmetric key (or another key whose value is a single octet sequence).
-        // * An "alg" member SHOULD also be present to identify the algorithm intended to be used with the key,
-        //   unless the application uses another means or convention to determine the algorithm used.
-        // NOTE: "k" is already checked in step 4.
-        if (!jwk.alg.has_value())
-            return WebIDL::DataError::create(m_realm, "Missing 'alg' field"_utf16);
+        // - ".k" is a valid bas64url encoded octet stream, which we do by just parsing it, in step 4.
+        // - ".alg" is checked only in step 5.
```
stelar7, line 9892: These were spec comments, please use the NOTE: syntax for stuff not in spec :)

mikiubo: thanks added note

let me know if now is ok

gmta: @stelar7 @mikiubo that is not correct, see our [guidelines](https://github.com/LadybirdBrowser/ladybird/blob/master/Documentation/CodingStyle.md#spec-notes). We use `// NB:` for our own comments that are not in the spec.

mikiubo: thanks!
i think now is fixed.

PR #7744, opened 2026-02-03
LibWeb: Make image alt text respect CSS-defined font properties

Psychpsyo:
This fixes the alt text for all the card images on https://crossuniverse.net/ showing when they are loading.
Those explicitly have `font-size: 0` to make that not be the case.
Other browsers also respect this and there's nothing in the spec to suggest this shouldn't happen.
I removed the border since it was inconsistently drawn only on missing images with alt text. Other browsers do not have this at all.


konradekk:
that was probably the case before evergreens; maybe one could salvage this by making sure that image failed to load? (not sure if that's necessary, though.)

Psychpsyo:
> that was probably the case before evergreens; maybe one could salvage this by making sure that image failed to load? (not sure if that's necessary, though.)

1. What are evergreens?
2. What was the case before them?
3. Salvage what?
4. How does ensuring the image failed to load help?

Tests/LibWeb/Screenshot/input/alt-frame.html
```diff
@@ -1,2 +0,0 @@
-<!DOCTYPE html>
-<img alt='Hello!!'>
```
AtkinsSJ, line 2: `Tests/LibWeb/Screenshot/expected/alt-frame.png` also needs deleting I think.

Psychpsyo: True, done that now.

PR #7745, opened 2026-02-03
LibWeb/DOM: Account for slot elements when walking the flat tree

AtkinsSJ:
In the flat tree, a slotted node's parent is the slot. Anywhere that we are supposed to get a parent or descendent in the flat tree, we need to take that into account.

I've introduced a couple of helper methods, `Node::flat_tree_parent()` and `Node::flat_tree_parent_element()`, and modified `shadow_including_first_ancestor_of_type<T>()` into `first_flat_tree_ancestor_of_type<T>()` as that's how it was actually used in practice. Then I searched for mentions of the "flat tree" and used these methods in those places where possible.

The test for layout-tree updates was reduced from https://ux.redhat.com/ but unfortunately this doesn't fix the issue on the live site. There's a lot more going on there which makes it hard to determine the real cause.

### DIFF
--- a/Libraries/LibWeb/DOM/Element.cpp
+++ b/Libraries/LibWeb/DOM/Element.cpp
@@ -3545,6 +3545,7 @@ bool Element::is_relevant_to_the_user()
         }
 
         // The element has a flat tree descendant that is captured in a view transition.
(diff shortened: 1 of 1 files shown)


AtkinsSJ:
Rebased to fix a merge conflict.

PR #7749, opened 2026-02-03
LibWeb: Don't lie about browsing context being top-level

Psychpsyo:
The VERIFY() this was triggering wasn't actually to spec, and by the time it gets encountered, the browsing context isn't technically a top-level context yet, because it has just been created and the definition of a top-level browsing context requires the document to be the navigables active document, which it only becomes once a history entry has been created for it.

Therefore we cannot just verify a top-level browsing context actually being a top-level browsing context when inserting it into the group, because that happens to early in its life-cycle as a top-level context.

This makes it so that JS console commands from the devtools no longer get sent to the last created nested frame in the tab, because nested frames no longer pretend to have a top-level browsing context.


Psychpsyo:
<s>Actually, I lied. This does not fix the problem. Making this a draft while I figure stuff out.</s>
EDIT: I've figured it out, this is correct now.

PR #7751, opened 2026-02-03
LibWeb: Add presentational hints for clear attribute on BR elements

Psychpsyo:
The 'imported' WPT test is currently under review here: https://github.com/web-platform-tests/wpt/pull/57553

This fixes a layout bug on https://archive.org/details/mspaint_xp_version, where the "7 Original" download button in the download options on the right would otherwise spill out of the download options box. (And I assume the same bug happens on all archive.org pages that have download options)

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLBRElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLBRElement.cpp
@@ -7,6 +7,7 @@
 #include <LibWeb/Bindings/HTMLBRElementPrototype.h>
 #include <LibWeb/CSS/ComputedProperties.h>
 #include <LibWeb/CSS/StyleValues/DisplayStyleValue.h>
+#include <LibWeb/CSS/StyleValues/KeywordStyleValue.h>
 #include <LibWeb/DOM/Document.h>
 #include <LibWeb/HTML/HTMLBRElement.h>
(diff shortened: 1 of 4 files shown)


Libraries/LibWeb/HTML/HTMLBRElement.cpp
```diff
@@ -33,6 +34,29 @@ GC::Ptr<Layout::Node> HTMLBRElement::create_layout_node(GC::Ref<CSS::ComputedPro
...
+bool HTMLBRElement::is_presentational_hint(FlyString const& name) const
+{
+    if (Base::is_presentational_hint(name))
+        return true;
+
+    return name == HTML::AttributeNames::clear;
+}
+
+void HTMLBRElement::apply_presentational_hints(GC::Ref<CSS::CascadedProperties> cascaded_properties) const
+{
```
AtkinsSJ, line 46: You want a call to Base::apply_presentational_hints() here somewhere too.

Psychpsyo: Done.

Psychpsyo:
@AtkinsSJ I've also opened https://github.com/LadybirdBrowser/ladybird/pull/7754 now, since `Base::apply_presentational_hints()` was just not getting called for most elements.

PR #7753, opened 2026-02-04
LibJS: Strength reduce basic math operations

dosisod:
The following math operations are now optimized to equivalent versions which require less operations and/or pre-checks:

* `x * 1` -> `x`
* `x * -1` -> `-x`
* `x / 1` -> `x`
* `x / -1` -> `-x`
* `x**2` -> `x * x` (evaluates `x` once)
* `x**1` -> `x`
* `x**0` -> `1` (still evaluates `x`)

### DIFF
--- a/Libraries/LibJS/Bytecode/ASTCodegen.cpp
+++ b/Libraries/LibJS/Bytecode/ASTCodegen.cpp
@@ -234,6 +234,14 @@ Bytecode::CodeGenerationErrorOr<Optional<ScopedOperand>> BinaryExpression::gener
             return result.release_value();
     }
 
+    Optional<i32> lhs_int;
+    Optional<i32> rhs_int;
+
+    if (auto constant = generator.try_get_constant(lhs); constant.has_value() && constant->is_int32())
+        lhs_int = constant->as_i32();
+    if (auto constant = generator.try_get_constant(rhs); constant.has_value() && constant->is_int32())
+        rhs_int = constant->as_i32();
+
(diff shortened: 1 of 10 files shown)


gmta:
Hi! Can you provide some `js-benchmarks` before/after results to show that these optimizations actually improve performance? If we add some complexity for prettier bytecode, we should see some real world improvements as well. :)

adrian17:
> x**2 -> x * x (evaluates x once)

Are you sure? Given `var x = {valueOf: function(){ console.log("called"); return 5; }}`, it looks to me that the optimization will cause this to be called twice now.

dosisod:
@adrian17 Yep it is evaluating twice now. I'll go ahead and remove that optimization.

konradekk:
> @adrian17 Yep it is evaluating twice now. I'll go ahead and remove that optimization.

i did not track it thoroughly, but it seems that you’d need to run `to_numeric` before multiplying the numbers  (i guess that’d be a new operation to be defined? virtually the same thing as `ToInt32` or `ToString`?). up to you to decide if it’s worth it or not really!

dosisod:
@konradekk That would be an option, but the impact would probably be limited.

I did some (limited) profiling with `js-benchmark`, and this PR has negligible improvements. For that reason, I'm going to close this and work on bigger performance issues.

gmta:
@dosisod 

> I did some (limited) profiling with js-benchmark, and this PR has negligible improvements.

Thanks for double checking! :)

PR #7755, opened 2026-02-04
LibWeb: Only apply overflow clipping to applicable element types

tcl3:
This fixes the rendering of issue labels on GitHub.

Before:

image

After: 

image

### DIFF
--- a/Libraries/LibWeb/Painting/ViewportPaintable.cpp
+++ b/Libraries/LibWeb/Painting/ViewportPaintable.cpp
@@ -312,7 +312,8 @@ void ViewportPaintable::assign_accumulated_visual_contexts()
         auto overflow_y = computed_values.overflow_y();
         auto has_hidden_overflow = overflow_x != CSS::Overflow::Visible || overflow_y != CSS::Overflow::Visible;
 
-        if (has_hidden_overflow || paintable_box.layout_node().has_paint_containment()) {
+        auto should_clip_overflow = has_hidden_overflow && paintable_box.overflow_property_applies();
+        if (should_clip_overflow || paintable_box.layout_node().has_paint_containment()) {
             bool clip_x = overflow_x != CSS::Overflow::Visible;
             bool clip_y = overflow_y != CSS::Overflow::Visible;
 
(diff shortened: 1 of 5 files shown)


kalenikaliaksandr:
actually I think the fix is correct, but lets add inline comment that mentions overflow property only works for "block containers, flex containers and grid containers" per https://www.w3.org/TR/css-overflow-3/#propdef-overflow, so we have to slip paintable boxes corresponding to inline nodes.

tcl3:
I've added a comment linking the relevant spec section. I've also excluded a few more element types from having overflow applied. I've inverted the check compared to what the spec says because otherwise you have to deal with every different kind of SVG box. 

I'm seeing these new WPT passes with this change, but the tests rely on the Ahem font, so can't be imported.

```
  ▶ PASS [expected FAIL] /css/CSS2/ui/overflow-applies-to-001.xht

  ▶ PASS [expected FAIL] /css/CSS2/ui/overflow-applies-to-003.xht

  ▶ PASS [expected FAIL] /css/CSS2/ui/overflow-applies-to-002.xht

  ▶ PASS [expected FAIL] /css/CSS2/ui/overflow-applies-to-004.xht

  ▶ PASS [expected FAIL] /css/css-overflow/overflow-does-not-apply-to-inline-box.html
```

Libraries/LibWeb/Painting/ViewportPaintable.cpp
```diff
@@ -312,7 +312,16 @@ void ViewportPaintable::assign_accumulated_visual_contexts()
         auto overflow_y = computed_values.overflow_y();
         auto has_hidden_overflow = overflow_x != CSS::Overflow::Visible || overflow_y != CSS::Overflow::Visible;
 
-        if (has_hidden_overflow || paintable_box.layout_node().has_paint_containment()) {
+        // https://drafts.csswg.org/css-overflow-3/#overflow-control
+        // Overflow properties apply to block containers, flex containers and grid containers.
+        // FIXME: Ideally we would check whether overflow applies positively rather than listing exceptions. However,
+        //        not all elements that should support overflow are currently identifiable that way.
+        auto const& display = computed_values.display();
+        bool overflow_does_not_apply = paintable_box.layout_node().is_inline_node()
```
kalenikaliaksandr: lets move this into a helper method defined on PaintableBox along with spec explanation comment?

kalenikaliaksandr (approved):

PR #7756, opened 2026-02-04
LibWeb: Reset line height for select elements

### LINKED ISSUE #7750: Regression in Select element vertical sizing when line-height is set.
### Summary

When you put line-height on a Select element, it is as tall as the set line-height, but it should not be.

Noticed this on https://battle.crossuniverse.com, where I know it was working as of November last year. (when I filed https://github.com/LadybirdBrowser/ladybird/issues/6720 )

### Operating system

Linux

### Steps to reproduce

1. Open the minimal repro and look at the tall select

### Expected behavior

The select should not be tall.

### Actual behavior

The select is tall.

N/A

```html
<!DOCTYPE html>
<select style="line-height: 200px">
	<option>This is way too tall.</option>
</select>
```

```shell
none
```

Ladybird right, Firefox left:
Image

_No response_

### Contribute a patch?

tcl3:
I tested all form control types with `line-height: 200px`  and these were the only differences I found between us and other engines.

EDIT: Actually I've removed the `::file-selector-button` commit, since the exact behavior seems to vary across browsers.

Fixes #7750

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLSelectElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLSelectElement.cpp
@@ -11,6 +11,7 @@
 #include <LibWeb/Bindings/Intrinsics.h>
 #include <LibWeb/CSS/ComputedProperties.h>
 #include <LibWeb/CSS/StyleValues/DisplayStyleValue.h>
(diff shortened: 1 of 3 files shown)


Psychpsyo:
Shouldn't this have an `AD-HOC:` comment somewhere?

PR #7758, opened 2026-02-04
LibWeb: Use inverse scroll context for `background-attachment: fixed`

### LINKED ISSUE #4981: fixed-position backgrounds break inside scroll containers
### Summary

When I use `background-attachment: fixed` in Ladybird, it does not work right if painted inside scroll frames.
So this affects elements inside scrolling areas as well as sticky-positioned elements.

This breaks the background on https://battle.crossuniverse.net/settings/

### Operating system

Linux

### Steps to reproduce

1. Put `background-attachment: fixed` on an element in a scroll frame or on a sticky-positioned element.
2. Watch it be in the wrong spot.

### Expected behavior

The background should be painted relative to the viewport's size and position.

### Actual behavior

The background is painted... not relative to the viewport, once scrolled.

N/A

```html
<!DOCTYPE html>
<style>
body {
  --image: linear-gradient(#e66465, #9198e5);
  height: 200vh;
  background-image: var(--image);
  background-attachment: fixed;
}

#sticky {
  position: sticky;
  top: 50vh;

  background-image: var(--image);
  background-attachment: fixed;
  height: 2em;
}

#scrollHolder {
  height: 20em;
  overflow-y: scroll;
}
#scrollInside {
  height: 40em;
  background-image: var(--image);
  background-attachment: fixed;
}
</style>
<div id="scrollHolder">
  <div id="scrollInside"></div>
</div>
<div id="sticky"></div>
```

```shell
none
```

In Ladybird:

Image

In Firefox / Chrome:

Image

none

### Contribute a patch?

tcl3:
Previously, fixed backgrounds incorrectly scrolled with content when scrolling incrementally. We now negate the viewport scroll offset during display list replay for fixed backgrounds. This keeps the background static relative to the viewport.

This fixes the background rendering issue seen on http://futo.org in the latest [Ladybird update video](https://www.youtube.com/watch?v=7_qhUidegUM&t=1141s).

Fixes #4981

### DIFF
--- a/Libraries/LibWeb/CSS/StyleInvalidation.cpp
+++ b/Libraries/LibWeb/CSS/StyleInvalidation.cpp
@@ -132,7 +132,7 @@ RequiredInvalidationAfterStyleChange compute_property_invalidation(CSS::Property
     }
     invalidation.repaint = true;
 
-    // Transform, perspective, clip, clip-path, and effects properties require rebuilding AccumulatedVisualContext tree.
+    // Transform, perspective, clip, clip-path, effects, and background-attachment properties require rebuilding AccumulatedVisualContext tree.
     if (AK::first_is_one_of(property_id,
             CSS::PropertyID::Transform,
             CSS::PropertyID::Rotate,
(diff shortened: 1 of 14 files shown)


Cameron-Asdf:
Fixes https://github.com/LadybirdBrowser/ladybird/issues/4981

gmta:
🤔 couldn't we have static positioned visual context data so we don't need to do any inverse operation, but just return coordinates as-is without applying scroll offsets, going through all the AVC parents, etc.?

kalenikaliaksandr (changes requested):
1. AccumulatedVisualContext should be invalidated on background property change.
2. It doesn't seem like InverseViewportScrollData idea would work if fixed background is nested into a scroll container.

PR #7759, opened 2026-02-04
LibWeb: Implement basics of the WebXR testing framework

Psychpsyo:
Continuation of https://github.com/LadybirdBrowser/ladybird/pull/7424.
We need this to run any of the WPT tests.
This also removes our own test for sessionSupported, since the WPT one now works.
Additional WPT tests will need some integration with WebGL though.


Libraries/LibWeb/Internals/FakeXRDevice.cpp
```diff
@@ -0,0 +1,44 @@
+/*
+ * Copyright (c) 2025, Tim Flynn <trflynn89@ladybird.org>
```
shannonbooth: Copyright name seems wrong

Libraries/LibWeb/HTML/Window.cpp
```diff
@@ -755,6 +755,10 @@ void Window::set_internals_object_exposed(bool exposed)
 {
     s_internals_object_exposed = exposed;
 }
+bool Window::is_internals_object_exposed()
```
shannonbooth, line 761: nit: newline between functions please

Libraries/LibWeb/Internals/FakeXRDevice.h
```diff
@@ -0,0 +1,43 @@
...
+
+#pragma once
+
+#include <LibWeb/Bindings/XRSystemPrototype.h>
+#include <LibWeb/Internals/InternalsBase.h>
+#include <LibWeb/WebXR/XRSession.h>
+
+namespace Web::Internals {
+
+struct FakeXRDeviceInit : public JS::Cell {
```
shannonbooth: It's an unfortunate that this compiles - I am pretty sure the IDL generator would be putting this on the stack and does not GC allocate this. We also don't need this to be GC allocated. I believe this can just be a plain struct.

Libraries/LibWeb/Internals/XRTest.cpp
```diff
@@ -0,0 +1,57 @@
+/*
+ * Copyright (c) 2025, Tim Flynn <trflynn89@ladybird.org>
```
shannonbooth: Copyright

Libraries/LibWeb/Internals/XRTest.cpp
```diff
@@ -0,0 +1,57 @@
...
+    // Simulates connecting a device to the system.
+    // Used to instantiate a fake device for use in tests.
+    // FIXME: Actually perform whatever device connection steps are needed once those are implemented.
+    auto& realm = HTML::relevant_realm(*this);
+    auto promise = WebIDL::create_promise(realm);
+    WebIDL::resolve_promise(realm, promise, FakeXRDevice::create(realm));
+    return promise;
+}
+
+void XRTest::simulate_user_activation(GC::Ref<WebIDL::CallbackType> f) const
```
shannonbooth: Let's give this a name a bit more descriptive than just `f` (even function would probably be fine)

shannonbooth (changes requested):

Libraries/LibWeb/Internals/XRTest.cpp
Psychpsyo: I made it `f`, because that's what it's called in the IDL definition. Should I change it anyways?

shannonbooth: I think so

Psychpsyo: Done now. (As well as all the other things)

PR #7763, opened 2026-02-04
LibWeb/HTML: Simplify checking if navigation ID is no longer the same

shannonbooth:
Small tidy up I noticed when debooging some navigation stuff


shannonbooth:
Oops messed up the title from tidying up my branch

gmta (approved):

PR #7766, opened 2026-02-05
LibWeb: Use the correct Module Record in HostLoadImportedModule

F3n67u:
This patch syncs change from whatwg/html#12114

### DIFF
--- a/Libraries/LibWeb/Bindings/MainThreadVM.cpp
+++ b/Libraries/LibWeb/Bindings/MainThreadVM.cpp
@@ -463,9 +463,9 @@ void initialize_main_thread_vm(AgentType type)
             // FIXME: Why do we need to check requested modules is empty here?
             if (auto const& requested_modules = referrer.get<GC::Ref<JS::CyclicModule>>()->requested_modules(); !requested_modules.is_empty() && module_request == requested_modules.first()) {
                 // 1. For each ModuleRequest record requested of referrer.[[RequestedModules]]:
-                for (auto const& module_request : referrer.get<GC::Ref<JS::CyclicModule>>()->requested_modules()) {
-                    // 1. If moduleRequest.[[Attributes]] contains a Record entry such that entry.[[Key]] is not "type", then:
-                    for (auto const& attribute : module_request.attributes) {
+                for (auto const& requested : referrer.get<GC::Ref<JS::CyclicModule>>()->requested_modules()) {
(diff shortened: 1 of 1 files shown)


Libraries/LibWeb/Bindings/MainThreadVM.cpp
```diff
@@ -484,47 +484,47 @@ void initialize_main_thread_vm(AgentType type)
                         // 4. Return.
                         return;
                     }
-                }
 
-                // 2. Resolve a module specifier given referencingScript and moduleRequest.[[Specifier]], catching any
-                //    exceptions. If they throw an exception, let resolutionError be the thrown exception.
-                auto maybe_exception = HTML::resolve_module_specifier(referencing_script, module_request.module_specifier.view().to_utf8_but_should_be_ported_to_utf16());
+                    // 2. Resolve a module specifier given referencingScript and requested.[[Specifier]], catching any
+                    //    exceptions. If they throw an exception, let resolutionError be the thrown exception.
```
gmta, line 489: The referenced spec text uses "moduleRequest", not "requested":

> 2. [Resolve a module specifier](https://whatpr.org/html/9893/webappapis.html#resolve-a-module-specifier) given referencingScript and moduleRequest.[[Specifier]], catching any exceptions. If they throw an exception, let resolutionError be the thrown exception.

shannonbooth: Note that this is because this function is sort of a mix between mainline upstream spec https://html.spec.whatwg.org/multipage/webappapis.html#hostloadimportedmodule and shadow realm PR of: https://whatpr.org/html/9893/webappapis.html#hostloadimportedmodule

gmta: Sure, but then:

* Use the correct spec URL;
* Fix other spec steps as well, such as this one:

```cpp
// 1. If moduleRequest.[[Attributes]] contains a Record entry such that entry.[[Key]] is not "type", then:
```

F3n67u: > Use the correct spec URL;

I think current spec URL is correct. It says:

```
    // 8.1.6.7.3 HostLoadImportedModule(referrer, moduleRequest, loadState, payload), https://html.spec.whatwg.org/multipage/webappapis.html#hostloadimportedmodule
    // https://whatpr.org/html/9893/webappapis.html#hostloadimportedmodule
```

> Fix other spec steps as well, such as this one:

Fixed.

PR #7773, opened 2026-02-05
ClangPlugins+Everywhere: Add MUST_UPCALL attribute for virtual functions

awesomekling:
This introduces a new `MUST_UPCALL` macro that expands to `[[clang::annotate("must_upcall")]]` on Clang. When placed on a virtual function, the Clang plugin will verify that all overrides call their base class implementation.
    
This generalizes the existing `Base::visit_edges()` check to work with any annotated virtual function. The first use is `Cell::visit_edges()`, but this can be applied to other functions that require upcalls.
    
The plugin checks for calls via either `Base::` (the common typedef pattern) or the explicit parent class name.

### DIFF
--- a/Libraries/LibWeb/CSS/CSSRule.h
+++ b/Libraries/LibWeb/CSS/CSSRule.h
@@ -62,7 +62,7 @@ public:
     // https://drafts.csswg.org/cssom-1/#serialize-a-css-rule
     virtual String serialized() const = 0;
 
-    virtual void dump(StringBuilder&, int indent_levels = 0) const;
+    MUST_UPCALL virtual void dump(StringBuilder&, int indent_levels = 0) const;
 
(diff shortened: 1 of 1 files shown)


gmta (approved):

Libraries/LibWeb/HTML/Storage.cpp
```diff
@@ -61,6 +61,7 @@ void Storage::initialize(JS::Realm& realm)
 
 void Storage::finalize()
 {
+    Base::finalize();
```
shannonbooth, line 64: I don't think it matters here, but since we're enforcing this now, logically would it make more sense if we call Base::finalize at the end of the function since we're unwinding, and maybe Base relies on something which is being called in finalize

PR #7776, opened 2026-02-05
LibHTTP+LibWebView: Impose disk cache limits and expose some settings to control them

trflynn89:
Rather than letting our disk cache grow unbounded, let's impose a limit on the estimated total disk cache size, and on the size of a single cache entry. A setting is added to `about:settings` to let the user control this limit.

https://github.com/user-attachments/assets/34308e4a-50c0-4ded-b8f4-18e883c8f0aa

<details>
<summary>Previous version recording</summary>

https://github.com/user-attachments/assets/a247600b-7445-4d02-bdbe-2eac2f864f7a

</details>

In the future, we can add more controls to this page.

### DIFF
--- a/Base/res/ladybird/about-pages/settings.html
+++ b/Base/res/ladybird/about-pages/settings.html
@@ -145,6 +145,7 @@
                 font-size: 14px;
             }
 
+            input[type="number"],
             input[type="text"],
             input[type="url"],
             select {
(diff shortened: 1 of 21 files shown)


trflynn89:
Edit: decided to combine the two cache settings/cache sizes dialogs into one for now, feels less awkward IMO.

PR #7777, opened 2026-02-05
LibWeb: Preserve event properties in label activation behavior

tcl3:
Previously, the label called `HTMLElement::click()` which dispatched a synthetic event with all properties set to their default values. We now preserve the properties of the original mouse event.

This fixes an issue on the Settings page where clicking a label in a dialog box would close that dialog box.

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLLabelElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLLabelElement.cpp
@@ -4,11 +4,15 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
+#include <AK/ScopeGuard.h>
 #include <LibWeb/Bindings/HTMLLabelElementPrototype.h>
 #include <LibWeb/DOM/Document.h>
 #include <LibWeb/HTML/Focus.h>
 #include <LibWeb/HTML/FormAssociatedElement.h>
 #include <LibWeb/HTML/HTMLLabelElement.h>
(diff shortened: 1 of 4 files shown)


Libraries/LibWeb/HTML/HTMLLabelElement.cpp
```diff
@@ -33,19 +39,79 @@ bool HTMLLabelElement::has_activation_behavior() const
...
+        init.client_y = mouse_event.client_y();
+        init.button = mouse_event.button();
+        init.buttons = mouse_event.buttons();
+        init.ctrl_key = mouse_event.ctrl_key();
+        init.shift_key = mouse_event.shift_key();
+        init.alt_key = mouse_event.alt_key();
+        init.meta_key = mouse_event.meta_key();
+        init.detail = mouse_event.detail();
+
+        if (auto const* pointer_event = as_if<UIEvents::PointerEvent>(event)) {
```
AtkinsSJ: `persistent_device_id` isn't set here, is that important?

trflynn89: I'm wondering if it makes sense to have a `MouseEvent::clone` virtual? Let each event subclass know what members need to be copied

tcl3: I've now added a `clone()` method for mouse and pointer events

`persistent_device_id` is now set.

Issue #7778, opened 2026-02-05
windowscentral.com crash


Samq64:
### Summary

What visiting an article on windowscentral.com the WebContent proccess crashes after a few seconds. This does not happen on the site's home page.

### Operating system

Linux

### Steps to reproduce

1. Visit https://www.windowscentral.com/microsoft/windows-11/windows-11-now-requires-admin-rights-to-access-the-storage-settings
2. Wait a few seconds

### Expected behavior

The page should not crash.

### Actual behavior

The page crahses with a segmentation fault at `ladybird/Libraries/LibGC/Heap.cpp:400`.

### URL for a reduced test case

https://www.windowscentral.com/microsoft/windows-11/windows-11-now-requires-admin-rights-to-access-the-storage-settings

### HTML/SVG/etc. source for a reduced test case

```html
N/A
```

### Log output and (if possible) backtrace

```shell
12280.470 WebContent(431307): vkCreateInstance returned -9
12280.470 WebContent(431307): Vulkan context creation failed: Application instance creation failed
12283.982 WebContent(431307): ResourceLoader: Refusing to prefetch DNS for 'https://securepubads.g.doubleclick.net/': URL was filtered
12284.084 WebContent(431307): (js log) "fep ga exc Cannot access property "primaryProduct" on undefined object "a[0].adTags""
12284.145 WebContent(431307): HTMLScriptElement: Refusing to run classic script because it has the nomodule attribute.
12284.233 WebContent(431307): (js log) "[Kiosq] Subscription Gates API available: window.FUTR.Kiosq.Gates"
12284.237 WebContent(431307): ResourceLoader: Filtered request to: "https://www.googletagmanager.com/gtm.js?id=GTM-WWBWRXL"
12284.237 WebContent(431307): ResourceLoader: Filtered request to: "https://uk-script.dotmetrics.net/door.js?d=www.windowscentral.com&t=wc"
12284.416 WebContent(431307): vkCreateInstance returned -9
12284.416 WebContent(431307): Vulkan context creation failed: Application instance creation failed
12284.424 WebContent(431307): vkCreateInstance returned -9
12284.424 WebContent(431307): Vulkan context creation failed: Application instance creation failed
12284.426 WebContent(431307): vkCreateInstance returned -9
12284.426 WebContent(431307): Vulkan context creation failed: Application instance creation failed
12284.427 WebContent(431307): HTMLScriptElement: Refusing to run script because the element's result is null.
[...]

PR #7784, opened 2026-02-06
AK: Use correct relocation for all HashTable entry types

tcl3:
Robin Hood displacement and `delete_bucket()` shift-up used BucketType's implicit move operations, which bitwise-copy the `u8` storage array instead of going through T's move constructor and destructor. This change adds `relocate_bucket()` and `swap_buckets()` helpers that use a fast path for trivially-relocatable types and move-construct + destroy for others.

Fixes: #7778

### DIFF
--- a/AK/HashTable.h
+++ b/AK/HashTable.h
@@ -13,6 +13,7 @@
 #include <AK/ReverseIterator.h>
 #include <AK/StdLibExtras.h>
 #include <AK/Traits.h>
+#include <AK/TypedTransfer.h>
 #include <AK/Types.h>
 #include <AK/kmalloc.h>
 
@@ -578,6 +579,31 @@ private:
     bool should_grow() const { return ((m_size + 1) * 100) >= (capacity() * grow_at_load_factor_percent); }
     static constexpr size_t size_in_bytes(size_t capacity) { return sizeof(BucketType) * capacity; }
 
+    static void relocate_bucket(BucketType* dst, BucketType* src)
+    {
+        dst->state = src->state;
+        dst->hash = src->hash;
+        if constexpr (IsOrdered) {
+            dst->previous = exchange(src->previous, nullptr);
+            dst->next = exchange(src->next, nullptr);
+        }
+        TypedTransfer<T>::relocate(dst->slot(), src->slot(), 1);
+    }
+
+    static void swap_buckets(BucketType* a, BucketType* b)
+    {
+        swap(a->state, b->state);
+        swap(a->hash, b->hash);
+        if constexpr (IsOrdered) {
+            swap(a->previous, b->previous);
+            swap(a->next, b->next);
+        }
+        alignas(T) u8 tmp[sizeof(T)];
+        TypedTransfer<T>::relocate(reinterpret_cast<T*>(tmp), a->slot(), 1);
+        TypedTransfer<T>::relocate(a->slot(), b->slot(), 1);
+        TypedTransfer<T>::relocate(b->slot(), reinterpret_cast<T*>(tmp), 1);
+    }
+
     BucketType* end_bucket()
     {
         if constexpr (IsOrdered)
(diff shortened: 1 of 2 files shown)


AK/HashTable.h
```diff
@@ -578,6 +578,48 @@ class HashTable {
...
+        dst->state = src->state;
+        dst->hash = src->hash;
+        if constexpr (IsOrdered) {
+            dst->previous = src->previous;
+            dst->next = src->next;
+            src->previous = nullptr;
+            src->next = nullptr;
+        }
+        if constexpr (IsTriviallyRelocatable<T>) {
+            __builtin_memcpy(dst->storage, src->storage, sizeof(T));
```
gmta: If T is trivially relocatable, couldn't you do this for the entire bucket instead of just the storage? state/hash and the ordering fields look trivial as well.

AK/HashTable.h
```diff
@@ -578,6 +578,48 @@ class HashTable {
...
+        } else {
+            new (dst->slot()) T(move(*src->slot()));
+            src->slot()->~T();
```
gmta: I feel like a lot of this logic could belong in `TypedTransfer<T>`, maybe a new `::relocate(destination, source)`?

tcl3: I had a go at implementing this. I think this does what you're asking:

https://github.com/tcl3/ladybird/tree/ak_hashmap_relocation_fix_typedtransfer

Do you think this is this preferable to the current implementation?  I assume a compiler would be able to lower this to be essentially the same as what we have now, but I've not checked.

gmta: Yeah that looks about right - maybe use memmove instead of memcpy though?

tcl3: I've pushed that implementation to this branch. I think `memcpy()` is correct because `relocate()` requires non-overlapping memory regions - otherwise source objects could be destroyed before they are moved. I've added an invariant to the top of `relocate()` to ensure source and destination do not overlap.

AK/TypedTransfer.h
```diff
@@ -72,6 +72,8 @@ class TypedTransfer {
 
     static void relocate(T* destination, T* source, size_t count)
     {
+        VERIFY(source + count <= destination || destination + count <= source);
+
```
gmta, line 76: This should go in the previous commit (at least, if ::relocate() is not going to support overlapping regions)

AK/HashTable.h
```diff
@@ -861,11 +888,7 @@ class HashTable {
                 break;
 
             auto* shift_to_bucket = &m_buckets[shift_to_index];
-            *shift_to_bucket = move(*shift_from_bucket);
-            if constexpr (IsOrdered) {
-                shift_from_bucket->previous = nullptr;
-                shift_from_bucket->next = nullptr;
-            }
+            relocate_bucket(shift_to_bucket, shift_from_bucket);
```
gmta, line 891: Could you run some microbenchmarks for different HashTable scenarios to compare the before and after?

tcl3: I've added some benchmarks that stress this code path. Here are the results averaged over 3 runs:

| Benchmark | Before (ms) | After (ms) | Delta |
  |---|---|---|---|
  | `insert_remove_int` | 40 | 42 | +2ms (+5%) |
  | `insert_remove_int_ordered` | 80 | 77 | -3ms (-4%) |
  | `insert_remove_non_trivial` | 113 | 116 | +3ms (+3%) |
  | `insert_remove_non_trivial_ordered` | 137 | 139 | +2ms (+1%) |
  | `high_collision_int` | 191 | 194 | +3ms (+2%) |
  | `high_collision_non_trivial` | 192 | 207 | +15ms (+8%) |
  | **Total** | **749** | **770** | **+21ms (+3%)** |

The majority of the slowdown comes from the `high_collision_non_trivial` case, where I've made collisions occur on purpose - this tallies with what I would expect.

AK/TypedTransfer.h
tcl3: Oops :sweat_smile: . I've moved this to the correct place.

Tests/AK/BenchmarkHashTable.cpp
gmta, line 1: This is the first time we have a `Benchmark*.cpp` in this directory, any reason not to put it in the existing `TestHashTable.cpp` like we do with other benchmark cases?

PR #7786, opened 2026-02-06
LibHTTP: Ensure we use the Vary key when updating last access time

trflynn89:
Apparently, sqlite will fill this placeholder value in with 0 if we do not pass a value. The query being executed here is:
```sql
UPDATE CacheIndex
SET last_access_time = ?
WHERE cache_key = ? AND vary_key = ?;
```

### DIFF
--- a/Libraries/LibDatabase/Database.cpp
+++ b/Libraries/LibDatabase/Database.cpp
@@ -96,7 +96,7 @@ ErrorOr<StatementID> Database::prepare_statement(StringView statement)
     return statement_id;
 }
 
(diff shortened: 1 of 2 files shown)


AtkinsSJ:
That feels like a footgun, could we get a check in `execute_statement()` to make sure the number of parameters matches?

trflynn89:
Yeah I was looking into that separately, because I was afraid we'd have to write a parser to count the number of unescaped `?` in the query. But turns out there's a sqlite API for this.

PR #7787, opened 2026-02-06
LibWeb: Delay load events for style sheet subresources

AtkinsSJ:
`<style>` now fires its `load` or `error` event when loading all of its [critical subresources](https://html.spec.whatwg.org/multipage/infrastructure.html#critical-subresources) has completed, instead of immediately. We use the same mechanism to delay the document load event for both `<style>` and `<link rel=stylesheet>`. (`<link>`'s load event doesn't yet wait for subresources, because it wants us to do so before we know what those are.)

Roughly speaking, subresources register themselves with their style sheet, and they each have a loading state that can be queried. When one of them loads, fails to load, or is removed from the style sheet, it tells its style sheet to checks its overall loading state. If it is complete, it reports it to its owner, whether that's an `@import`, `<style>`, or `<link rel=stylesheet>`. The `@import` then also reports to its parent style sheet, so eventually the top element knows that loading is complete.

The spec is a bit vague and hand-wavy about what subresources are critical to load. `@import`-ed style sheets are explicitly mentioned, and in practice both fonts and background-images also seem to be required by the web. This PR only implements this for `@import`, but it already improves some test results.

### DIFF
--- a/Libraries/LibWeb/CSS/CSSImportRule.cpp
+++ b/Libraries/LibWeb/CSS/CSSImportRule.cpp
@@ -137,9 +137,6 @@ void CSSImportRule::fetch()
     if (m_supports && !m_supports->matches())
         return;
 
-    // FIXME: Figure out the "correct" way to delay the load event.
-    m_document_load_event_delayer.emplace(*m_document);
-
     // AD-HOC: Track pending import rules to block rendering until they are done.
     m_document->add_pending_css_import_rule({}, *this);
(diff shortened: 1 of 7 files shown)


AtkinsSJ:
Made a spec issue about the `<link>` problem as I don't think this is just an implementation detail of ours. https://github.com/whatwg/html/issues/12140

AtkinsSJ:
Changes: Stop trying to store `GC::Ref<Subresource>` when it's not a GC type. Then rebased to fix a tiny conflict.

PR #7799, opened 2026-02-06
Implemented curl_code_to_network_error as an inline constexpr functio…

xD-saleem:
Description:

This PR refactors the curl_code_to_network_error utility to use a constexpr lookup table instead of a procedural switch statement.

Changes:

Defined ErrorMapping struct to pair CURL codes with NetworkError enums.

Implemented curl_code_to_network_error as an inline constexpr function using std::array.

### DIFF
--- a/Services/RequestServer/CURL.cpp
+++ b/Services/RequestServer/CURL.cpp
@@ -7,6 +7,8 @@
 
 #include <AK/Enumerate.h>
 #include <RequestServer/CURL.h>
+#include <array>
+#include <utility>
 
 namespace RequestServer {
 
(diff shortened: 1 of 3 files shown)


Zaggy1024:
What's the purpose of making this function constexpr if it's only called at runtime? Also, the existing switch statement is just fine for a constexpr function, and seems more likely to build efficient code for runtime calls as well.

If there is a good reason to use an array lookup, though, this should be using `AK::Array`, not STL headers.

xD-saleem:
> What's the purpose of making this function constexpr if it's only called at runtime? Also, the existing switch statement is just fine for a constexpr function, and seems more likely to build efficient code for runtime calls as well.
> 
> If there is a good reason to use an array lookup, though, this should be using `AK::Array`, not STL headers.

Thank you for the feedback! You make a great point. Its my first contribution. I will address your feedback. thanks for your review.

PR #7801, opened 2026-02-06
LibCrypto: Add AES-OCB support

mikiubo:
Implement AES-OCB AEAD using OpenSSL and expose it through the WebCrypto API, including key management and AEAD parameters.

Add WPT:
/encrypt_decrypt/aes_ocb.tentative.https.any.html


Libraries/LibCrypto/Cipher/AES.cpp
```diff
@@ -174,6 +174,79 @@ ErrorOr<ByteBuffer> AESGCMCipher::decrypt(ReadonlyBytes ciphertext, ReadonlyByte
...
 }
 
+AESOCBCipher::AESOCBCipher(ReadonlyBytes key)
+    : AESCipher(GET_CIPHER(key, ocb), key)
+{
+}
+
+ErrorOr<ByteBuffer> AESOCBCipher::encrypt(ReadonlyBytes plaintext, ReadonlyBytes iv, ReadonlyBytes aad, size_t tag_length) const
+{
+    auto tag_length_bytes = tag_length / 8;
```
gmta, line 184: `tag_length_bytes` should be between 1 and 16 inclusive (as guaranteed by the callers), can we VERIFY that here and in ::decrypt() below?

Libraries/LibWeb/Crypto/CryptoAlgorithms.cpp
```diff
@@ -9997,4 +9997,356 @@ WebIDL::ExceptionOr<JS::Value> ChaCha20Poly1305::get_key_length(AlgorithmParams
...
+        return WebIDL::OperationError::create(m_realm, Utf16String::formatted("Encryption failed: {}", maybe_ciphertext.error()));
+
+    // 5. Return C
+    return JS::ArrayBuffer::create(m_realm, maybe_ciphertext.value());
+}
+
+// https://wicg.github.io/webcrypto-modern-algos/#aes-ocb-operations-decrypt
+WebIDL::ExceptionOr<GC::Ref<JS::ArrayBuffer>> AesOcb::decrypt(AlgorithmParams const& params, GC::Ref<CryptoKey> key, ByteBuffer const& ciphertext)
+{
+
```
gmta: Stray newline

gmta:
@mikiubo 

> Add WPT:
> /encrypt_decrypt/aes_ocb.tentative.https.any.html

Looks like you forgot this WPT test though? Are the custom tests still necessary?

mikiubo:
> @mikiubo
> 
> > Add WPT:
> > /encrypt_decrypt/aes_ocb.tentative.https.any.html
> 
> Looks like you forgot this WPT test though? Are the custom tests still necessary?

You're right — I had only added the custom test initially and forgot to include the WPT.

I've now added the WPT (/encrypt_decrypt/aes_ocb.tentative.https.any.html).

I put it in a separate commit since I noticed a recent PR (8086) doing the same.

If possible, I'd prefer to keep the custom test as well, since it was helpful for debugging as a reduced test case.

Libraries/LibCrypto/Cipher/AES.cpp
mikiubo: added VERIFY

Libraries/LibWeb/Crypto/CryptoAlgorithms.cpp
mikiubo: fixed

PR #7802, opened 2026-02-06
LibWeb: Prevent thin double borders from disappearing

Psychpsyo:
Brings us more in line with what other browsers do and also fixes a visual bug on a website I use.


Tests/LibWeb/Screenshot/images/css-background-clip-text.png
gmta, line 1: Could you please optipng/pngcrush this PNG?

Psychpsyo: Done.

Libraries/LibWeb/Painting/BorderPainting.cpp
```diff
@@ -165,7 +168,7 @@ void paint_border(DisplayListRecorder& painter, BorderEdge edge, DevicePixelRect
         }
 
         modified_borders_data.for_edge(edge).line_style = inner_style;
-        paint_border(painter, edge, modified_rect, modified_radius, modified_opposite_radius, modified_borders_data, path, last);
+        paint_border(painter, edge, modified_rect, modified_radius, modified_opposite_radius, modified_borders_data, path, true);
```
gmta, line 171: What do these changes from `last` to `true` do? How are they related to the minimum size of the double borders 🤔

Psychpsyo: The `last` was used so that this can draw the border going all the way round the element as one contiguous stroke. If we use it on double borders, we get this non-contiguous stroke order instead:
image
With borders this thin, the stroke-drawing code doesn't like these gaps between the segments where it needs to jump back to the start of an edge and that leads to it drawing almost no borders at all.
(apart from some tiny slivers of something border-ish in the corners of the element sometimes, I think)

gmta: Could you mention that in the commit message?

Psychpsyo: Done.

gmta (approved):

PR #7805, opened 2026-02-07
LibMedia: Fix multi-channel decode for dumb containers (like WAV)

jonbgamble:
For web audio, I reckon an occasional misjudged channel layout is better than more frequent exceptions.

Signed PCM is normalized with the (unsigned max + 1) / 2, not signed max. If you divide by the signed max (32767), you get headroom that can exceed the threshold below -1.0. It's not audible, this mostly matters for tests that assume correct normalization.


trflynn89:
> dumb containers

> no shortage of "golden ears" jackholes

I'll let @Zaggy1024 review the code, but let's be more professional in commit messages please.

Libraries/LibMedia/FFmpeg/FFmpegAudioDecoder.cpp
```diff
@@ -140,7 +140,9 @@ requires(IsSigned<T>)
 static float float_sample_from_frame_data(u8** data, size_t plane, size_t index)
 {
     auto* pointer = reinterpret_cast<T*>(data[plane]);
-    return pointer[index] / static_cast<float>(NumericLimits<T>::max());
+    if (IsSame<T, i16>)
+        return pointer[index] / 32768.0f;                                // PCM 16 bit is normalized with 32768
```
Zaggy1024: This should probably be another template specialization like the u8 specialization above.

But I wonder, is this the only signed PCM format that we need to treat the same way, i.e.:

```cpp
return pointer[index] / (static_cast<float>(NumericLimits<T>::max()) + 1);
```

or even:

```cpp
return pointer[index] / -static_cast<float>(NumericLimits<T>::min());
```

Libraries/LibMedia/FFmpeg/FFmpegHelpers.cpp
```diff
@@ -42,7 +42,16 @@ ErrorOr<Audio::ChannelMap> av_channel_layout_to_channel_map(AVChannelLayout cons
...
+        if (layout.nb_channels == 4)
+            return Audio::ChannelMap::quadrophonic();
+        if (layout.nb_channels == 6)
+            return Audio::ChannelMap::surround_5_1();
+        if (layout.nb_channels == 8)
+            return Audio::ChannelMap::surround_7_1();
+
+        for (int i = 0; i < layout.nb_channels; ++i)
+            channels[i] = Audio::Channel::Unknown;
+        return Audio::ChannelMap(channels);
```
Zaggy1024, line 54: Do you have a WAV file with more than 2 channels that doesn't include the channel mask? It would be good to add a test for this if so.

I was under the impression that WAV files with more than 2 channels would store the fields from WAVEFORMATEXTENSIBLE with its channel mask, but I don't know if it's a requirement.

jonbgamble: Tests/LibWeb/Text/input/wpt-import/webaudio/the-audio-api/the-audiobuffersourcenode-interface/resources/audiobuffersource-multi-channels-expected.wav

Shall I add it to the Tests/LibMedia/WAV

jonbgamble: master lacks the webaudio infrastructure to test this with the wpt path. so i can make a cpp unit test?

Libraries/LibMedia/FFmpeg/FFmpegAudioDecoder.cpp
jonbgamble: `static_cast<float>((NumericLimits<UnsignedT>::max() + 1u) >> 1)`

just kidding, either one is fine with me. choose

Libraries/LibMedia/FFmpeg/FFmpegHelpers.cpp
R-Goc: That is not even valid WAV. However gemini tells me it does occur in the wild. So in that case we would have to assume default speaker config.

jonbgamble: WAVE is an old container and lots of files predate or do not use WAVEFORMATEXTENSIBLE. This one happens to be structurally sound and decodable.

I would say it's underspecified rather than invalid. The lack of a channel mask is shared by oceans of content produced by legacy DAWs, hardware, and codecs.

Libraries/LibMedia/FFmpeg/FFmpegAudioDecoder.cpp
Zaggy1024: My main question here was whether we can expect to see similar issues with other signed integer types, rather than whether we should avoid using a magic number (though avoiding that is also nice). Is signed 16-bit PCM the only one that uses the minimum as the normalizing factor?

Libraries/LibMedia/FFmpeg/FFmpegHelpers.cpp
Zaggy1024: Let's get a C++ test in there to test whether AudioDataProvider can get a block out of that file with a channel map that matches what other browsers do, ideally.

Libraries/LibMedia/FFmpeg/FFmpegAudioDecoder.cpp
jonbgamble: 24 bit PCM works the same way (divide by 2^23), but this template function and numeric limits won't help.

8 bit PCM is unsigned, so it's (u8 - 128) / 128. I don't know how much 8 bit is out there. Maybe web-based console emulators?

Zaggy1024: We already have a specialization for 8-bit, so that's no problem. If 24 bit (and 32/64?) work the same way, we could change the general case, no? Or use the unsigned concept for another specialization.

jonbgamble: Our ffmpeg version decodes 24 bit correctly, but delivers it as 32 bit ints with LSB padding, so we normalize with 2^31 and all is good.

Libraries/LibMedia/FFmpeg/FFmpegAudioDecoder.cpp
```diff
@@ -140,7 +140,8 @@ requires(IsSigned<T>)
 static float float_sample_from_frame_data(u8** data, size_t plane, size_t index)
 {
     auto* pointer = reinterpret_cast<T*>(data[plane]);
-    return pointer[index] / static_cast<float>(NumericLimits<T>::max());
+    i64 full_scale = 1LL << ((sizeof(T) * 8) - 1);
```
Zaggy1024:
```suggestion
    constexpr float full_scale = NumericLimits<MakeUnsigned<T>>::max() / 2;
```

I think this should work, right? constexpr allows the compiler to verify the cast at compile time, so I think it shouldn't result in any warnings.

wertzuz:
> > dumb containers
> 
> > no shortage of "golden ears" jackholes
> 
> I'll let @Zaggy1024 review the code, but let's be more professional in commit messages please.

Well nearly worked

PR #7808, opened 2026-02-07
LibWeb: Fix precision errors when passing negative numbers

cqundefine:
Because the `convert_to_int` function would use double which loses precision at around 2^53 the representation of 64 bit negative numbers like -1 which would be 0xFFFFFFFFFFFFFFFF would lose precision and be converted to 0 which is obviously incorrect. Switching to long double fixes the problem as on GCC/Clang it is guaranteed to be able to represent the 64 bit integer range. This issue would become apparent with WebGL functions which often take `long long` as parameter.

### DIFF
--- a/Tests/LibWeb/CMakeLists.txt
+++ b/Tests/LibWeb/CMakeLists.txt
@@ -1,4 +1,5 @@
 set(TEST_SOURCES
+    TestAbstractOperations.cpp
     TestCSSIDSpeed.cpp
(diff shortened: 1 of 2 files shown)


gmta:
Could you add a regression test?

cqundefine:
Drafting for now until I figure out a better solution as I forgot that `long double` is terribly non portable and this doesn't really fix the problem on Windows and macOS.

PR #7810, opened 2026-02-07
LibWeb: Deduplicate code for dispatching events to nested navigables

Psychpsyo:
No real change in behavior, except that some of these events that used to only go to iframes should go to all nested navigables now. This is a good thing.


Psychpsyo:
Test failures seem unrelated.

PR #7814, opened 2026-02-07
CI: Make errors from push notes workflow visible

InvalidUsernameException:

* [Successful run of the workflow with these changes in my fork](https://github.com/InvalidUsernameException/ladybird/actions/runs/21966418318/job/63457321135)
* [Failed run of the workflow with these changes and the old URL causing a 404 error](https://github.com/InvalidUsernameException/ladybird/actions/runs/21966378782/job/63457183944)
* A new workflow run being queued while another one is still running:
Queued workflow run

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -58,8 +58,6 @@ runs:
       if: ${{ inputs.os == 'Linux' }}
       shell: bash
       run: |
-        set -e
-
         sudo add-apt-repository ppa:ubuntu-toolchain-r/test
 
         UBUNTU_RELEASE=$(lsb_release -cs)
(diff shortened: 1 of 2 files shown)


.github/workflows/notes-push.yml
```diff
@@ -16,6 +16,8 @@ jobs:
           fetch-depth: 0
       - uses: fregante/setup-git-user@v2
       - run: |
+          set -e
+
```
gmta: According to [this](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepsshell), `-e` is already the default?

InvalidUsernameException: Good point, I guess I didn't properly verify the behavior in the error scenario and also misremembered what exactly is covered by `set -e`. Corrected now by splitting download and invocation of git-gloss in two separate commands.

I've also added a commit to remove `set -e` calls from the few other workflow steps that had them. I did grep for them before the change and having them there somewhat encouraged me to add the `set -e` without properly verifying it.

PR #7816, opened 2026-02-07
Implement partial SVG relayout

kalenikaliaksandr:
Previously, any SVG geometry attribute change would mark the entire
document layout tree as dirty, triggering a full layout pass even though
only the SVG subtree was affected. This made SVG geometry animations
unnecessarily expensive.

Fix this by stopping `needs_layout_update` propagation at the SVGSVGBox
boundary and tracking dirty SVG roots separately on the Document. When
`update_layout()` finds that only SVG roots need relayout (and the
document layout root is clean), it runs SVGFormattingContext on each
dirty SVG root in a fresh LayoutState and commits the results directly,
bypassing the full document layout pass entirely.

This results in a severe performance improvement on pages with animated
SVGs, such as https://www.cloudflare.com/, https://www.duolingo.com/,
and our GC graph explorer page.

### DIFF
--- a/Libraries/LibWeb/DOM/Document.cpp
+++ b/Libraries/LibWeb/DOM/Document.cpp
@@ -156,6 +156,8 @@
 #include <LibWeb/Infra/Strings.h>
 #include <LibWeb/IntersectionObserver/IntersectionObserver.h>
 #include <LibWeb/Layout/BlockFormattingContext.h>
+#include <LibWeb/Layout/SVGFormattingContext.h>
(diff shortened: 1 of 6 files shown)


awesomekling:
You had me at "severe performance improvement" 😄

Psychpsyo:
Are you planning to generalize this to apply to size-contained elements also? Otherwise I'd look into that after this gets merged, to get closer to getting CSS container query stuff working.

PR #7817, opened 2026-02-07
LibWeb: Do not remove inline start padding when collapsing whitespace

Psychpsyo:
Fixes a layout issue on a website I was looking at.


Psychpsyo:
Timeout looks unrelated.

PR #7821, opened 2026-02-08
LibWeb: Add spec comment to algorithm

Psychpsyo:


Libraries/LibWeb/Layout/FormattingContext.cpp
```diff
@@ -870,6 +870,7 @@ void FormattingContext::compute_width_for_absolutely_positioned_non_replaced_ele
     box_state.margin_right = margin_right.to_px_or_zero(box);
 }
 
+// https://www.w3.org/TR/CSS22/visudet.html#abs-replaced-width
```
gmta: We use the editor's draft of specs, so that'd be: https://drafts.csswg.org/css2/#abs-replaced-width

Psychpsyo: Done.

PR #7824, opened 2026-02-08
LibWeb: Protect against null navigables in lineage chain more thoroughly

trflynn89:
This extends the null navigable check added in commit b118c99c271e34e2c5020022d062a4371f199a71 to include all ancestor and descendant list lookups. Fixes a crash in the following WPT test `/cookies/schemeful-same-site/schemeful-navigation.tentative.html`

### DIFF
--- a/Tests/LibWeb/Text/expected/DOM/window-open-and-reopen-custom-target.txt
+++ b/Tests/LibWeb/Text/expected/DOM/window-open-and-reopen-custom-target.txt
@@ -1,0 +1,2 @@
+1: Target matches source: true
+2: Target matches source: true
--- a/Libraries/LibWeb/DOM/Document.cpp
+++ b/Libraries/LibWeb/DOM/Document.cpp
@@ -4103,9 +4103,17 @@ Vector<GC::Root<HTML::Navigable>> const Document::descendant_navigables() const
 // https://html.spec.whatwg.org/multipage/document-sequences.html#inclusive-descendant-navigables
 Vector<GC::Root<HTML::Navigable>> Document::inclusive_descendant_navigables()
 {
+    // FIXME: The document's node navigable should not be null here. But we currently do not implement the "unload a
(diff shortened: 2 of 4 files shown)


Tests/LibWeb/Text/expected/DOM/window-open-and-reopen-custom-target.txt
```diff
@@ -0,0 +1,2 @@
+1 file:///home/flynn/workspace/ladybird/Tests/LibWeb/Text/data/window-opener-post-message.html true
+2 file:///home/flynn/workspace/ladybird/Tests/LibWeb/Text/data/window-opener-post-message.html true
```
shannonbooth: Unfortunately my ladybird isn't in this folder :P

trflynn89: ha, yeah I saw that right after opening :man_facepalming:

shannonbooth:
I guess this is building on precedent, but I kind of wonder if it would be better if callers are responsible for checking this instead of in these functions. Some of the callers are already checking things such as active documents and that type of thing.

It's just that it feels a bit weird for Navigable::inclusive_ancestor_navigables to return an empty list if inclusive doesn't exist, it feels like inclusive existing is a precondition of invoking it

trflynn89:
I tracked this down to our implementation of `Document::unload_a_document_and_its_descendants` being incorrect. We invoke `destroy_a_document_and_its_descendants`, which the spec no longer calls for:
https://html.spec.whatwg.org/multipage/document-lifecycle.html#unload-a-document-and-its-descendants

We should only be unloading (not destroying) the document, and then destroying it later at a kind of hand-wavy time. I tried implementing this closer to the spec in a couple of ways, and it results in massive test-web timeouts. So the spec function change was probably part of a larger change that we need to implement.

So for now, I've added FIXMEs with a description of what's going on to stop the crash.

shannonbooth (approved):

PR #7827, opened 2026-02-08
LibRegex+LibUnicode: Implement Unicode case-insensitive matching

aplefull:
This PR implements case-insensitive Unicode matching. We now case-fold chars using `canonicalize` during case-insensitive matching. 
Not sure I like the amount of stuff I had to add to make this work, but I’ve been staring at this branch for long enough that I may be blind to obvious improvements, so suggestions are very welcome :^)

Test262:
+11 ✅    -11 ❌   

Diff:
```
test/built-ins/RegExp/regexp-modifiers/add-ignoreCase-affects-slash-lower-b.js ❌ -> ✅
test/built-ins/RegExp/regexp-modifiers/add-ignoreCase-affects-slash-lower-p.js ❌ -> ✅
test/built-ins/RegExp/regexp-modifiers/add-ignoreCase-affects-slash-lower-w.js ❌ -> ✅
test/built-ins/RegExp/regexp-modifiers/add-ignoreCase-affects-slash-upper-b.js ❌ -> ✅
test/built-ins/RegExp/regexp-modifiers/add-ignoreCase-affects-slash-upper-p.js ❌ -> ✅
test/built-ins/RegExp/regexp-modifiers/add-ignoreCase-affects-slash-upper-w.js ❌ -> ✅
test/staging/sm/RegExp/ignoreCase-multiple.js                                  ❌ -> ✅
test/staging/sm/RegExp/ignoreCase-non-latin1-to-latin1.js                      ❌ -> ✅
test/staging/sm/RegExp/unicode-class-ignoreCase.js                             ❌ -> ✅
test/staging/sm/RegExp/unicode-ignoreCase-escape.js                            ❌ -> ✅
test/staging/sm/RegExp/unicode-ignoreCase-word-boundary.js                     ❌ -> ✅
```

### DIFF
--- a/Libraries/LibRegex/RegexByteCode.cpp
+++ b/Libraries/LibRegex/RegexByteCode.cpp
@@ -153,6 +153,20 @@ static bool restore_string_position(MatchInput const& input, MatchState& state)
     return true;
 }
 
+static bool is_word_character(u32 code_point, bool case_insensitive, bool unicode_mode)
+{
+    if (is_ascii_alphanumeric(code_point) || code_point == '_')
+        return true;
+
+    if (case_insensitive && unicode_mode) {
+        auto canonical = Unicode::canonicalize(code_point, unicode_mode);
+        if (is_ascii_alphanumeric(canonical) || canonical == '_')
+            return true;
+    }
+
+    return false;
+}
+
 OwnPtr<OpCode<ByteCode>> ByteCode::s_opcodes[(size_t)OpCodeId::Last + 1];
 bool ByteCode::s_opcodes_initialized { false };
 
@@ -475,7 +489,9 @@ ALWAYS_INLINE ExecutionResult OpCode_CheckBegin<ByteCode>::execute(MatchInput co
 template<typename ByteCode>
 ALWAYS_INLINE ExecutionResult OpCode_CheckBoundary<ByteCode>::execute(MatchInput const& input, MatchState& state) const
 {
-    auto isword = [](auto ch) { return is_ascii_alphanumeric(ch) || ch == '_'; };
+    auto isword = [&](auto ch) {
+        return is_word_character(ch, state.current_options & AllFlags::Insensitive, input.view.unicode());
+    };
     auto is_word_boundary = [&] {
         if (state.string_position == input.view.length()) {
             return (state.string_position > 0 && isword(input.view.code_point_at(state.string_position_in_code_units - 1)));
@@ -951,12 +967,12 @@ ALWAYS_INLINE ExecutionResult CompareInternals<ByteCode, IsSimple>::execute(Matc
         }
         case CharacterCompareType::Property: {
             auto property = static_cast<Unicode::Property>(bytecode_data[offset++]);
-            compare_property(input, state, property, current_inversion_state(), inverse_matched);
+            compare_property(input, state, property, current_inversion_state(), temporary_inverse && inverse, inverse_matched);
             break;
         }
         case CharacterCompareType::GeneralCategory: {
             auto general_category = static_cast<Unicode::GeneralCategory>(bytecode_data[offset++]);
-            compare_general_category(input, state, general_category, current_inversion_state(), inverse_matched);
+            compare_general_category(input, state, general_category, current_inversion_state(), temporary_inverse && inverse, inverse_matched);
             break;
         }
         case CharacterCompareType::Script: {
@@ -1001,7 +1017,7 @@ ALWAYS_INLINE ExecutionResult CompareInternals<ByteCode, IsSimple>::execute(Matc
                     if (state.current_options & AllFlags::Insensitive) {
                         bool found_child = false;
                         for (auto const& [key, child] : current->children()) {
-                            if (to_ascii_lowercase(key) == to_ascii_lowercase(value)) {
+                            if (Unicode::canonicalize(key, input.view.unicode()) == Unicode::canonicalize(value, input.view.unicode())) {
                                 current = static_cast<StringSetTrie const*>(child.ptr());
                                 current_code_unit_offset++;
                                 found_child = true;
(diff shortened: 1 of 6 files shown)


Libraries/LibRegex/RegexByteCode.cpp
```diff
@@ -1362,36 +1372,85 @@ ALWAYS_INLINE void CompareInternals<ByteCode, IsSimple>::compare_character_range
...
     u32 code_point = input.view.code_point_at(state.string_position_in_code_units);
-    bool equal = Unicode::code_point_has_property(code_point, property);
+    bool case_insensitive = (state.current_options & AllFlags::Insensitive) && input.view.unicode();
+    bool is_unicode_sets_mode = state.current_options.has_flag_set(AllFlags::UnicodeSets);
+
+    // In /u mode, case folding happens after complementing: \P{x} matches caseFold(allChars - charsWithX).
+    // This means a code point matches the inverted property if ANY of its case variants lacks the property.
+    // In /v mode, case folding happens before complementing: \P{x} matches caseFold(allChars) - caseFold(charsWithX),
+    // so we can just use normal case-insensitive matching and invert the result.
+    if ((inverse || is_double_negation) && case_insensitive && !is_unicode_sets_mode) {
```
alimpfard, line 1388: Nice!

alimpfard (commented):
Regex bits LGTM, LibUnicode bits look okay but I'm not very familiar with those.

Libraries/LibUnicode/CharacterTypes.h
```diff
@@ -74,4 +77,35 @@ enum class BidiClass {
 
 BidiClass bidirectional_class(u32 code_point);
 
+struct CodePointRange {
+    u32 from;
+    u32 to;
```
trflynn89: We should always initialize plain types in struct definitions.
```c++
struct CodePointRange {
    u32 from { 0 };
    u32 to { 0 };
};
```

Libraries/LibUnicode/CharacterTypes.h
```diff
@@ -7,14 +7,17 @@
...
 #include <AK/Optional.h>
 #include <AK/Types.h>
+#include <AK/Vector.h>
 #include <LibUnicode/Forward.h>
 
 namespace Unicode {
 
 Optional<GeneralCategory> general_category_from_string(StringView);
-bool code_point_has_general_category(u32 code_point, GeneralCategory general_category);
+bool code_point_has_general_category(u32 code_point, GeneralCategory general_category, bool case_insensitive = false);
```
trflynn89: Let's prefer named enums over booleans. In this case, we should use `CaseSensitivity` from AK.

Libraries/LibUnicode/CharacterTypes.cpp
```diff
@@ -463,4 +519,89 @@ BidiClass bidirectional_class(u32 code_point)
     return char_direction_to_bidi_class(direction);
 }
 
+// https://tc39.es/ecma262/#sec-runtime-semantics-canonicalize-ch
+u32 canonicalize(u32 code_point, bool unicode_mode)
+{
```
trflynn89, line 534: Please include spec comments in the style of LibJS here

Libraries/LibUnicode/CharacterTypes.cpp
```diff
@@ -188,19 +221,42 @@ Optional<Property> property_from_string(StringView property)
...
+        return false;
+
+    if (!s_property_sets_with_case_closure.contains(property)) {
+        UErrorCode status = U_ZERO_ERROR;
+        auto set = make<icu::UnicodeSet>();
+
+        set->applyIntPropertyValue(icu_property, 1, status);
 
-    return static_cast<bool>(u_hasBinaryProperty(icu_code_point, icu_property));
+        if (U_SUCCESS(status)) {
```
trflynn89: Is there a case that this can actually fail? Looking at the implementation:
https://github.com/unicode-org/icu/blob/2c2e8adbc142a76a72cac408194e8d20701603ac/icu4c/source/common/uniset_props.cpp#L1344

It looks like this will fail if we give an invalid property (which I don't think can happen?) or if we run out of memory (which we explicitly have decided not to handle gracefully). I could be missing something though.

Just wondering, because then `ensure` would clean this up nicely, instead of `contains` + `set` + `get`.

trflynn89: (note we also have a more C++-ish wrapper around `U_SUCCESS`,  `icu_success`).

Libraries/LibUnicode/CharacterTypes.h
```diff
@@ -74,4 +77,35 @@ enum class BidiClass {
...
+
+template<typename Range1, typename Range2>
+bool ranges_equal_ignoring_case(Range1 const& range1, Range2 const& range2, bool unicode_mode)
+{
+    auto it1 = range1.begin();
+    auto it2 = range2.begin();
+    auto end1 = range1.end();
+    auto end2 = range2.end();
+
+    while (it1 != end1 && it2 != end2) {
```
trflynn89: This could be:
```c++
for (; it1 != end1 && it2 != end2; ++it1, ++it2)
```

Libraries/LibUnicode/CharacterTypes.cpp
aplefull: Yeah, I was just being overly cautious because I didn’t know in which cases it could fail. It looks like we always have correct input in both `code_point_has_general_category` and `code_point_has_property`, so I cleaned it up now, thanks!

aplefull:
Adressed the comments, rebased on master.
One run flaked, but other one failed during build with `variable set but not used`, even tho it's used? And another one failed with 0 fails and 0 timeouts? I am confusion, could someone rerun please

trflynn89:
The gcc build error is unlikely to be a flake, IIRC it's a known issue that it doesn't think constexpr vars used only in lambdas are actually used. Try one of two (or both) things:
1. make the vars `static constexpr`
2. make the lambdas capture all `[&]`

aplefull:
Hm, that didn't work. I'll just inline charcodes in this test and leave comments about what these characters are. Code points are inlined in other tests in that file as well, so hopefully it's ok, I just thought named variables look a bit more readable.

LucasChollet:
I haven't seen the failing code but the bug I know (and that I think you're describing) is triggered by generic lambdas. It has been fixed in GCC 16.

PR #7833, opened 2026-02-08
LibWeb: Add overflow-clip-margin-* properties

Psychpsyo:
The corner radius isn't quite right yet, but this gives us another couple WPT passes for these.

The `overflow-clip-margin-multi-value.html` WPT test was written for this PR and imported manually.
It is currently under review here: https://github.com/web-platform-tests/wpt/pull/57643

### DIFF
--- a/Libraries/LibWeb/Painting/ViewportPaintable.cpp
+++ b/Libraries/LibWeb/Painting/ViewportPaintable.cpp
@@ -221,31 +221,59 @@ static Optional<ClipData> compute_clip_data(PaintableBox const& paintable_box, C
 {
     auto overflow_x = computed_values.overflow_x();
     auto overflow_y = computed_values.overflow_y();
+
+    // https://drafts.csswg.org/css-contain-2/#paint-containment
+    // 1. The contents of the element including any ink or scrollable overflow must be clipped to the overflow clip
+    //    edge of the paint containment box, taking corner clipping into account. This does not include the creation of
+    //    any mechanism to access or indicate the presence of the clipped content; nor does it inhibit the creation of
+    //    any such mechanism through other properties, such as overflow, resize, or text-overflow.
+    //    NOTE: This clipping shape respects overflow-clip-margin, allowing an element with paint containment
+    //          to still slightly overflow its normal bounds.
+    if (paintable_box.layout_node().has_paint_containment()) {
+        // NOTE: Note: The behavior is described in this paragraph is equivalent to changing 'overflow-x: visible' into
+        //       'overflow-x: clip' and 'overflow-y: visible' into 'overflow-y: clip' at used value time, while leaving other
+        //       values of 'overflow-x' and 'overflow-y' unchanged.
+        overflow_x = CSS::Overflow::Clip;
+        overflow_y = CSS::Overflow::Clip;
+    }
+
     auto has_hidden_overflow = overflow_x != CSS::Overflow::Visible || overflow_y != CSS::Overflow::Visible;
 
-    auto should_clip_overflow = has_hidden_overflow && paintable_box.overflow_property_applies();
-    if (should_clip_overflow || paintable_box.layout_node().has_paint_containment()) {
-        bool clip_x = overflow_x != CSS::Overflow::Visible;
-        bool clip_y = overflow_y != CSS::Overflow::Visible;
+    if (has_hidden_overflow && paintable_box.overflow_property_applies()) {
+        auto clip_rect = paintable_box.absolute_padding_box_rect();
 
-        if (paintable_box.layout_node().has_paint_containment()) {
-            clip_x = true;
-            clip_y = true;
+        // https://drafts.csswg.org/css-overflow-3/#propdef-overflow
+        // 'clip'
+        //    This value indicates that the box’s content is clipped to its overflow clip edge
+        auto overflow_clip_edge = paintable_box.overflow_clip_edge_rect();
+        if (overflow_x == CSS::Overflow::Visible) {
+            clip_rect.set_left(0);
+            clip_rect.set_right(CSSPixels::max_integer_value);
+        } else if (overflow_x == CSS::Overflow::Clip) {
+            clip_rect.set_left(overflow_clip_edge.left());
+            clip_rect.set_right(overflow_clip_edge.right());
+        }
+        if (overflow_y == CSS::Overflow::Visible) {
+            clip_rect.set_top(0);
+            clip_rect.set_bottom(CSSPixels::max_integer_value);
+        } else if (overflow_y == CSS::Overflow::Clip) {
+            clip_rect.set_top(overflow_clip_edge.top());
+            clip_rect.set_bottom(overflow_clip_edge.bottom());
         }
 
-        if (clip_x || clip_y) {
-            auto clip_rect = paintable_box.overflow_clip_edge_rect();
-            if (!clip_x) {
-                clip_rect.set_left(0);
-                clip_rect.set_right(CSSPixels::max_integer_value);
-            }
-            if (!clip_y) {
-                clip_rect.set_top(0);
-                clip_rect.set_bottom(CSSPixels::max_integer_value);
-            }
-            auto radii = (clip_x && clip_y) ? paintable_box.normalized_border_radii_data(PaintableBox::ShrinkRadiiForBorders::Yes) : BorderRadiiData {};
-            return ClipData { clip_rect, radii };
-        }
+        // https://drafts.csswg.org/css-overflow-3/#corner-clipping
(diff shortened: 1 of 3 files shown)


Psychpsyo:
Test failure seems unrelated and did not happen locally for me. (though I am on Linux, not MacOS)

Libraries/LibWeb/Painting/ViewportPaintable.cpp
```diff
@@ -343,31 +342,59 @@ void ViewportPaintable::assign_accumulated_visual_contexts()
...
-                    clip_rect.set_left(0);
-                    clip_rect.set_right(CSSPixels::max_integer_value);
-                }
-                if (!clip_y) {
-                    clip_rect.set_top(0);
-                    clip_rect.set_bottom(CSSPixels::max_integer_value);
-                }
-                auto radii = (clip_x && clip_y) ? paintable_box.normalized_border_radii_data(ShrinkRadiiForBorders::Yes) : BorderRadiiData {};
-                state_for_descendants = append_node(state_for_descendants, ClipData { clip_rect, radii });
+        if (has_hidden_overflow && paintable_box.overflow_property_applies()) {
```
kalenikaliaksandr: lets move clip rect calculation into a helper

kalenikaliaksandr (changes requested):

Libraries/LibWeb/Painting/ViewportPaintable.cpp
Psychpsyo: Done now.

Psychpsyo:
Rebased onto master and resolved the merge conflicts.

Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
```diff
@@ -664,6 +664,18 @@ Parser::ParseErrorOr<NonnullRefPtr<StyleValue const>> Parser::parse_css_value(Pr
                 return parse_single_background_size_value(PropertyID::MaskSize, tokens);
             });
         });
+    case PropertyID::OverflowClipMargin:
```
Calme1709: Currently we always parse shorthands as `ShorthandStyleValue`s, it would be nice to maintain this for simplicity.

Libraries/LibWeb/Layout/Node.cpp
```diff
@@ -731,6 +731,16 @@ void NodeWithStyle::apply_style(CSS::ComputedProperties const& computed_style)
     computed_values.set_inset(computed_style.length_box(CSS::PropertyID::Left, CSS::PropertyID::Top, CSS::PropertyID::Right, CSS::PropertyID::Bottom, CSS::LengthPercentageOrAuto::make_auto()));
     computed_values.set_margin(computed_style.length_box(CSS::PropertyID::MarginLeft, CSS::PropertyID::MarginTop, CSS::PropertyID::MarginRight, CSS::PropertyID::MarginBottom, CSS::Length::make_px(0)));
     computed_values.set_padding(computed_style.length_box(CSS::PropertyID::PaddingLeft, CSS::PropertyID::PaddingTop, CSS::PropertyID::PaddingRight, CSS::PropertyID::PaddingBottom, CSS::Length::make_px(0)));
+    CSSPixelRect overflow_clip_margin = { 0, 0, 0, 0 };
+    if (auto const& top = computed_style.property(CSS::PropertyID::OverflowClipMarginTop); top.is_length())
```
Calme1709: Won't these `if` statements always be true?

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -424,7 +424,6 @@ void StyleComputer::for_each_property_expanding_shorthands(PropertyID property_i
         // Both should be applied to our longhand properties.
         // We don't directly call `set_longhand_property()` because the longhands might have longhands of their own.
         // (eg `grid` -> `grid-template` -> `grid-template-areas` & `grid-template-rows` & `grid-template-columns`)
-        VERIFY(value.is_css_wide_keyword() || value.is_guaranteed_invalid());
```
Calme1709, line 436: As discussed above, we should parse the shorthands as `ShorthandStyleValue`s and add this back in.

Libraries/LibWeb/Layout/Node.cpp
```diff
@@ -731,6 +731,16 @@ void NodeWithStyle::apply_style(CSS::ComputedProperties const& computed_style)
...
+        overflow_clip_margin.set_top(top.as_length().length().to_px(*this));
```
Calme1709: Lengths are absolutized during computation so we can use `absolute_length_to_px()` instead of `to_px()`

Libraries/LibWeb/Painting/ViewportPaintable.cpp
```diff
@@ -217,6 +217,68 @@ static Optional<Gfx::FloatMatrix4x4> compute_perspective_matrix(PaintableBox con
     return perspective_matrix;
 }
 
+static Optional<ClipData> compute_clip_data(PaintableBox const& paintable_box, CSS::ComputedValues const& computed_values)
```
Calme1709, line 220: I think these changes are distinct enough to warrant their own commit.

Psychpsyo:
Done all that now.
I've also changed the computed value to a LengthBox since that is the more appropriate type for this.
Even if it only needs to store lengths and not percentages or auto.

Libraries/LibWeb/Painting/ViewportPaintable.cpp
```diff
@@ -221,31 +221,59 @@ static Optional<ClipData> compute_clip_data(PaintableBox const& paintable_box, C
 {
     auto overflow_x = computed_values.overflow_x();
     auto overflow_y = computed_values.overflow_y();
-    auto has_hidden_overflow = overflow_x != CSS::Overflow::Visible || overflow_y != CSS::Overflow::Visible;
 
-    auto should_clip_overflow = has_hidden_overflow && paintable_box.overflow_property_applies();
-    if (should_clip_overflow || paintable_box.layout_node().has_paint_containment()) {
-        bool clip_x = overflow_x != CSS::Overflow::Visible;
-        bool clip_y = overflow_y != CSS::Overflow::Visible;
+    // https://drafts.csswg.org/css-contain-2/#paint-containment
```
Calme1709, line 225: I still think the remaining changes here would be better placed in a separate commit - or at least mentioned in the commit message.

Psychpsyo: Done now and added a test specifically for these changes.

Libraries/LibWeb/CSS/Properties.json
```diff
@@ -3257,6 +3257,98 @@
...
+    }
+  },
+  "overflow-clip-margin-bottom": {
+    "animation-type": "by-computed-value",
+    "inherited": false,
+    "initial": "0px",
+    "valid-types": [
+      "length [0,∞]"
+    ],
+    "requires-computation": "never"
```
Calme1709: These properties should be marked as be `"requires-computation": "cascaded-value"` in order to handle relative units (e.g. `em`) and `calc()`

Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
```diff
@@ -3752,6 +3765,35 @@ RefPtr<StyleValue const> Parser::parse_math_depth_value(TokenStream<ComponentVal
...
+    if (auto length = parse_length_value(tokens)) {
+        return length.release_nonnull();
+    }
+
+    return nullptr;
+}
+
+RefPtr<StyleValue const> Parser::parse_overflow_clip_margin_shorthand(PropertyID property_id, TokenStream<ComponentValue>& tokens)
+{
+    // <visual-box> || <length [0,∞]>
```
Calme1709, line 3783: I should have caught this earlier but we will also need to disallow serialization of the shorthands if their longhands aren't equal (which can happen when serializing a declaration), similar to [what we do for border](https://github.com/LadybirdBrowser/ladybird/blob/9418981398368f2564bdec1b2a5445cdf2abb092/Libraries/LibWeb/CSS/StyleValues/ShorthandStyleValue.cpp#L286-L302) - this can remain as a FIXME for now if that's easier

Libraries/LibWeb/CSS/Properties.json
Psychpsyo: Done.

Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
Psychpsyo: I'm putting a FIXME for now so that I can make the WPT test first and PR the fix here once that is merged and importable.

Calme1709:
Looks good to me.

PR #7834, opened 2026-02-08
LibWasm: Various GC improvements

mkanilsson:
This PR:
1. Adds support for parsing typeuse references (`ref null typeidx`) by extending `ValueType`,
2. array types in parsing and validation,
3. and validation of the type section.  

This allows modules like
```wat
(module
  (type $T (struct (field i32) (field f32)))
  (type $T1 (struct (field i32) (field f32)))
  (; many more types... ;)
  (type $T64 (struct (field i32) (field f32)))
  (type $f (func (result (ref null $T64))))
)
```
and 
```wat
(module
  (type $a (array i32))
)
```

The type-equivalence.{0,1,3,13} spec tests have been disable as they were false positives. The struct.{3,4} tests have been enabled as they are now passing.

### DIFF
--- a/Libraries/LibWasm/AbstractMachine/Validator.cpp
+++ b/Libraries/LibWasm/AbstractMachine/Validator.cpp
@@ -43,6 +43,9 @@ ErrorOr<void, ValidationError> Validator::validate(Module& module)
                         },
                         [&](StructType const& struct_) {
                             m_context.structs.append(struct_);
+                        },
+                        [&](ArrayType const& array) {
+                            m_context.arrays.append(array);
                         });
                 } else {
                     return Errors::invalid("TypeIndex"sv);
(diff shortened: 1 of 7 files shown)


Libraries/LibWasm/Printer/Printer.cpp
```diff
@@ -801,6 +801,7 @@ void Printer::print(Wasm::Value const& value, Wasm::ValueType const& type)
                 [](Wasm::Reference::Exception const&) { return ByteString("exception"); },
                 [](auto const& ref) { return ByteString::number(ref.address.value()); }));
         break;
+    case ValueType::TypeUseReference:
```
alimpfard, line 817: Can we get a separate printed name for this?

Libraries/LibWasm/Types.h
```diff
@@ -217,6 +231,10 @@ class ValueType {
 
 private:
     Kind m_kind;
+    Variant<
+        TypeIndex,
+        u8> // Empty state
```
alimpfard: Please use Empty as the empty state, that'll also make the variant default-init.

Libraries/LibWasm/AbstractMachine/Validator.cpp
```diff
@@ -349,6 +359,13 @@ ErrorOr<void, ValidationError> Validator::validate(ValueType const& type)
     return {};
 }
 
+ErrorOr<void, ValidationError> Validator::validate(TypeSection::Type const& type)
+{
+    return type.description().visit(
+        [&](FunctionType const& function) -> ErrorOr<void, ValidationError> { return validate(function); },
```
alimpfard: I don't think the explicit return type is necessary, please omit it.

mkanilsson:
Those are now fixed, I also fixed printing of array types as they looked weird

mkanilsson:
The failing test seems to be a network issue? Is there a way to rerun the test without pushing again?

PR #7838, opened 2026-02-08
LibWeb: Deselect text on mousedown

Psychpsyo:
Happens in other browsers and not doing it is incredibly annoying.


Tests/LibWeb/Text/input/selection-clear-on-click.html
```diff
@@ -0,0 +1,14 @@
...
+<div style="font-size:20px">SelectMe</div>
+<script>
+    test(() => {
+        // Select text by double-clicking
+        internals.click(20, 20, 2);
+        println(`"${window.getSelection()}"`);
+
+        // Now try deselecting on mousedown.
+        internals.mouseDown(100, 100);
+        println(`"${window.getSelection()}"`);
```
gmta, line 12: Pretty sure this already works like this on `master` 🤔

Psychpsyo: True... I must've written this on an outdated branch then or there was something else going on on the site I was using. Closing this.

PR #7840, opened 2026-02-09
LibJS: Use explicit completion records for try/finally dispatch

awesomekling:
Each finally scope gets two registers (`completion_type` and `completion_value`) that form an explicit completion record. Every path into the finally body sets these before jumping, and a dispatch chain after the finally body routes to the correct continuation.

This replaces the old implicit protocol that relied on the `exception` register, a `saved_return_value` register, and a `scheduled_jump` field on `ExecutionContext`, allowing us to remove:

- 5 opcodes (`ContinuePendingUnwind`, `ScheduleJump`, `LeaveFinally`, `RestoreScheduledJump`, `PrepareYield`)
- 1 reserved register (`saved_return_value`)
- 2 `ExecutionContext` fields (`scheduled_jump`, `previously_scheduled_jumps`)

### DIFF
--- a/Libraries/LibJS/Bytecode/ASTCodegen.cpp
+++ b/Libraries/LibJS/Bytecode/ASTCodegen.cpp
@@ -2788,39 +2788,80 @@ Bytecode::CodeGenerationErrorOr<Optional<ScopedOperand>> BreakStatement::generat
     return Optional<ScopedOperand> {};
 }
 
+// Try/finally uses an explicit completion record protocol:
+//
+//   1. Allocate two registers: completion_type and completion_value
+//   2. Every path into the finally body sets these before jumping:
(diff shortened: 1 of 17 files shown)


awesomekling:
This also fixes some test262 tests!

```
Summary:
    Diff Tests:
        +4 ✅    -4 ❌

Diff Tests:
    test/language/expressions/object/computed-property-name-topropertykey-before-value-evaluation.js ❌ -> ✅
    test/staging/sm/Function/return-finally.js                                                       ❌ -> ✅
    test/staging/sm/expressions/object-literal-computed-property-evaluation.js                       ❌ -> ✅
    test/staging/sm/generators/return-finally.js                                                     ❌ -> ✅
```

PR #7841, opened 2026-02-09
LibWeb: Implement CSSKeyframeRule::set_key_text()

Praise-Garfield:
Implement the previously stubbed `CSSKeyframeRule::set_key_text()` method per the CSS Animations Level 1 specification.

Parses the keyframe selector string ("from", "to", or a percentage in [0,100]) and updates the key. Throws a `SyntaxError` DOMException if the input is invalid.

Spec: https://drafts.csswg.org/css-animations/#dom-csskeyframerule-keytext

Includes a Text test covering valid inputs, normalization, and error cases.

### DIFF
--- a/Libraries/LibWeb/CSS/CSSKeyframeRule.cpp
+++ b/Libraries/LibWeb/CSS/CSSKeyframeRule.cpp
@@ -10,6 +10,7 @@
 #include <LibWeb/Bindings/Intrinsics.h>
 #include <LibWeb/CSS/CSSRuleList.h>
 #include <LibWeb/Dump.h>
+#include <LibWeb/WebIDL/DOMException.h>
 
 namespace Web::CSS {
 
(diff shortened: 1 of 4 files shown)


Libraries/LibWeb/CSS/CSSKeyframeRule.cpp
```diff
@@ -40,6 +41,32 @@ void CSSKeyframeRule::initialize(JS::Realm& realm)
     Base::initialize(realm);
 }
 
+// https://drafts.csswg.org/css-animations/#dom-csskeyframerule-keytext
+WebIDL::ExceptionOr<void> CSSKeyframeRule::set_key_text(String const& key_text)
+{
+    // On setting, if the value does not match the <keyframe-selector> grammar, throw a SyntaxError.
```
gmta, line 47: You can quote the spec verbatim:

```cpp
    // If keyText is updated with an invalid keyframe selector, a SyntaxError exception must be thrown and the value of
    // keyText must remain unchanged.
```

PR #7843, opened 2026-02-09
LibWeb: Always parse `<counter-style>` as such 

Calme1709:
Previously we parsed it as `<custom-ident>` in `<counter>` and as a keyword in `list-style-type`.

The practical effect of this is:
 - Spec defined counter style names in `<counter>` are ASCII lowercased on parse.
 - Non spec defined counter style names are allowed in `list-style-type.

We are still to parse the `symbols()` function but this gives us a better base for that.

### DIFF
--- a/Libraries/LibWeb/CMakeLists.txt
+++ b/Libraries/LibWeb/CMakeLists.txt
@@ -246,6 +246,7 @@ set(SOURCES
     CSS/StyleValues/CounterDefinitionsStyleValue.cpp
     CSS/StyleValues/CounterStyleSystemStyleValue.cpp
     CSS/StyleValues/CounterStyleValue.cpp
+    CSS/StyleValues/CounterStyleStyleValue.cpp
     CSS/StyleValues/CursorStyleValue.cpp
     CSS/StyleValues/CustomIdentStyleValue.cpp
     CSS/StyleValues/DimensionStyleValue.cpp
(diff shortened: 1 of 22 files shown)


Libraries/LibWeb/Layout/ListItemMarkerBox.cpp
```diff
@@ -80,6 +80,10 @@ GC::Ptr<Painting::Paintable> ListItemMarkerBox::create_paintable() const
 CSSPixels ListItemMarkerBox::relative_size() const
 {
     auto font_size = first_available_font().pixel_size();
+
+    if (m_list_style_type.has<Empty>())
+        return CSSPixels::nearest_value_for(ceilf(font_size * .35f));
```
gmta: I could have done a better job adding a comment about the 50% / 35% percentages and where they come from - let's take this opportunity to extract them into a separate `static constexpr` and explain why they were chosen. Something like this?

```cpp
    // We set a marker's size at 35% of the font size, which gives a good default visual size. The exception to this are
    // the open/closed discs, which would be too small at 35%. There is no spec dictating this size; we mimic Firefox.
    static constexpr float marker_relative_default_size = .35f;
    static constexpr float marker_relative_disc_size = .5;
```

Calme1709: I have added a commit for this - slightly different wording to reference the relevant part of the spec.

In thinking about this a bit more, this is never true (since we don't create a `ListItemMarkerBox` for `list-style-type: none` unless we have a list item image, in which case we never call `relative_size()`) so I have replaced this `if` statement with a `VERIFY(!m_list_style_type.has<Empty>())`

gmta: Much better yeah :) thanks!

PR #3267, opened 2025-01-15
LibWeb: Support blending in CSS and the canvas

skyz1:
This PR adds support for the `mix-blend-mode` CSS property.
It also implements compositing and various blending modes via the `globalCompositeOperation` attribute in the canvas.

This makes some more tests in https://wpt.fyi/results/html/canvas/element/compositing pass.

PR #7855, opened 2026-02-09
LibWeb: Do not create a layer when CSS isolation is set to isolate

Psychpsyo:
This is entirely unnecessary. All that this property does is create a stacking context.

I don't think this needs any tests since it's just removing something that, from what I can tell, is a no-op.


Psychpsyo:
Looking into this more, it was originally an unrelated change introduced in #3267 with no explanation.

PR #7856, opened 2026-02-09
LibGC: Harden for_each_message_port iteration for teardown

jonbgamble:
Guard against null derefs in for_each_message_port

### DIFF
--- a/Libraries/LibWeb/HTML/MessagePort.cpp
+++ b/Libraries/LibWeb/HTML/MessagePort.cpp
@@ -10,6 +10,7 @@
 #include <AK/MemoryStream.h>
 #include <LibCore/Socket.h>
 #include <LibCore/System.h>
+#include <LibGC/WeakHashSet.h>
 #include <LibIPC/Decoder.h>
 #include <LibIPC/Encoder.h>
 #include <LibIPC/File.h>
@@ -32,9 +33,9 @@ constexpr u8 IPC_FILE_TAG = 0xA5;
 
 GC_DEFINE_ALLOCATOR(MessagePort);
 
-static HashTable<GC::RawPtr<MessagePort>>& all_message_ports()
+static GC::WeakHashSet<MessagePort>& all_message_ports()
 {
-    static HashTable<GC::RawPtr<MessagePort>> ports;
+    static GC::WeakHashSet<MessagePort> ports;
     return ports;
 }
 
@@ -46,15 +47,16 @@ GC::Ref<MessagePort> MessagePort::create(JS::Realm& realm)
 MessagePort::MessagePort(JS::Realm& realm)
     : DOM::EventTarget(realm)
 {
-    all_message_ports().set(this);
+    all_message_ports().set(*this);
 }
 
 MessagePort::~MessagePort() = default;
 
 void MessagePort::for_each_message_port(Function<void(MessagePort&)> callback)
 {
-    for (auto port : all_message_ports())
-        callback(*port);
+    auto ports = all_message_ports();
+    for (auto& port : ports)
+        callback(port);
 }
 
 void MessagePort::initialize(JS::Realm& realm)
@@ -66,7 +68,7 @@ void MessagePort::initialize(JS::Realm& realm)
 void MessagePort::finalize()
 {
     Base::finalize();
-    all_message_ports().remove(this);
+    all_message_ports().remove(*this);
     disentangle();
 }
 


Libraries/LibWeb/HTML/MessagePort.cpp
```diff
@@ -53,7 +54,20 @@ MessagePort::~MessagePort() = default;
 
 void MessagePort::for_each_message_port(Function<void(MessagePort&)> callback)
 {
-    for (auto port : all_message_ports())
+    auto& ports = all_message_ports();
+    ports.remove_all_matching([](GC::RawPtr<MessagePort> const& port) {
+        return !port;
+    });
+
+    Vector<GC::Ref<MessagePort>> strong_ports;
```
awesomekling: This introduces a UAF. This `Vector<GC::Ref<MessagePort>>` is not visible to the garbage collector. `Vector<GC::Root<MessagePort>>` or `GC::RootVector<GC::Ref<MessagePort>>` would work.

jonbgamble: I never actually observed the lambda problem in the wild, that "fix" is just defense through inspection.

The reason I began monkeying with this file - `for_each_message_port` was dying in teardown (test-web navigations) when connected to an AudioWorklet (over IPC). I guess the filtering resolved the issue and I was just lucky my fail Vector<GC::Ref> didn't crash afterwards.

This situation probably does not happen in master branch, but I reckon it could one day.

Libraries/LibWeb/HTML/MessagePort.cpp
```diff
@@ -307,18 +321,24 @@ void MessagePort::read_from_transport()
...
-
         auto serialized_transfer_record = MUST(decoder.decode<SerializedTransferRecord>());
-
-        queue_global_task(Task::Source::PostedMessage, relevant_global_object(*this), GC::create_function(heap(), [this, serialized_transfer_record = move(serialized_transfer_record)]() mutable {
-            this->post_message_task_steps(serialized_transfer_record);
-        }));
+        auto weak_this = GC::Weak { *this };
+        queue_global_task(Task::Source::PostedMessage, relevant_global_object(*this),
+            GC::create_function(heap(), [weak_this, serialized_transfer_record = move(serialized_transfer_record)]() mutable {
+                if (!weak_this)
```
Zaggy1024: The `this` capture would have been kept alive by the conservative scan in `GC::Function`, so this only allows the MessagePort to be destroyed slightly earlier. If that's the purpose of this change, perhaps it should be in another commit?

Libraries/LibWeb/HTML/MessagePort.cpp
```diff
@@ -53,7 +54,20 @@ MessagePort::~MessagePort() = default;
...
+        return !port;
+    });
+
+    Vector<GC::Root<MessagePort>> strong_ports;
+    strong_ports.ensure_capacity(ports.size());
+    for (auto port : ports) {
+        if (!port)
+            continue;
+        strong_ports.append(GC::make_root(*port));
+    }
```
Zaggy1024: This is moot if we use WeakHashSet instead, but I think the null checks here are redundant, no? RawPtr (should've really been RawRef here) doesn't get nulled out unexpectedly, so we either have a pointer or we don't.

jonbgamble: WeakHashSet is an improvement, but it seems we'd still need to root for ultimate protection during the callback.

That said, I think the failures I observed were from direct un-checked iteration on the static HashTable at process exit rather than GC during callback. So let's roll with just your change, because I can't figure out how to expose the issue I originally had in tests. :)

Libraries/LibWeb/HTML/MessagePort.cpp
jonbgamble: It never hurts to destroy as early as possible, but in light of the Function scanning I'm not sure this prophylaxis justifies the churn.

Libraries/LibWeb/HTML/MessagePort.cpp
Zaggy1024: Right, that would make more sense. At least with this change, we're guaranteed to crash if the method is misused, so with regard to this comment, this looks good to me now.

Zaggy1024: ...actually, come to think of it, I think the real reason for your UAF was probably that the HashTable destructor doesn't clear its m_buckets nor m_size, so iterators will try to read from the freed memory pointed to by m_buckets. I ran into this in my (now-closed) PR to stop leaking the VM.

If you're looking to fix the underlying issue with this UAF, might be worth tossing in a commit to fix that.

Also, for the iteration function, iterating over the WeakHashSet directly instead of copying it here is probably okay, since we should only be doing anything to it on one thread.

shannonbooth: Hmm, wouldnt that also be safe due to stack scan? (re the rooting point)

Zaggy1024: Ah, indeed it probably should be, I didn't think about the callback argument.

Edit: No wait, it doesn't protect from other elements being removed from the collection mid iteration (which is very bad if it can happen, but it shouldn't here)

PR #7859, opened 2026-02-10
LibWeb: Input/textarea/contenteditable caret and navigation improvements

gmta:
* Empty `<input>` and `<textarea>` draw carets again
* Empty `contenteditable` elements now also draw a caret
* Programmatically (or auto-)focused elements reset the caret blink cycle
* When tabbing to an `<input>`, we now mimic other browsers and select its contents
* Disable caret blinking in test mode so we can deterministically generate display list dumps

Below an example of how the Roundcube login form improves in terms of UX.

| Before | After |
|--|--|
| <video src="https://github.com/user-attachments/assets/9dc74fee-fd3b-42ad-b11f-3b2fd983f534"></video> | <video src="https://github.com/user-attachments/assets/1beefc37-c63c-4b25-a74d-2e989cba885a"></video> |


warpdesign:
I was about to create an issue for the missing caret :)

PR #7860, opened 2026-02-10
LibHTTP+RequestServer: Send revalidation attributes without parsing

trflynn89:
The caching RFC is quite strict about the format of date strings. If we received a revalidation attribute with an invalid date string, we would previously fail a runtime assertion. This was because to start a revalidation request, we would simply check for the presence of any revalidation header; but then when we issued the request, we would fail to parse the header, and end up with all attributes being null.

We now don't parse the revalidation attributes at all. Whatever we receive in the `Last-Modified` response header is what we will send in the `If-Modified-Since` request header, verbatim. For better or worse, this is how other browsers behave. So if the server sends us an invalid date string, it can receive its own date format for revalidation.

<hr />

This fixes a crash during revalidation on http://tweakers.net, where we received:
```
Last-Modified: Mon, 9 Feb 2026 21:46:56 +0000
```

However, other browsers seem to literally receive the string "Invalid date". A printf in chromium where the raw headers are read from the network confirmed this. Curl on the command line does as well:
```bash
❯ curl -iv --raw "https://api.smartocto.com/api/brands/tentacles?i=i3njb9c9jq6yp04ifcl73wqq94ja7jb7"

< HTTP/2 200
HTTP/2 200
< x-powered-by: smartocto
x-powered-by: smartocto
< x-cache: HIT from smartocto cache
x-cache: HIT from smartocto cache
< max-age: 10
max-age: 10
< last-modified: Invalid date
last-modified: Invalid date
```

So I'm not sure what is going on with this server. But we at least now behave the same as the big 3 browsers.

### DIFF
--- a/Libraries/LibHTTP/Cache/Utilities.cpp
+++ b/Libraries/LibHTTP/Cache/Utilities.cpp
@@ -470,8 +470,8 @@ CacheLifetimeStatus cache_lifetime_status(HeaderList const& request_headers, Hea
 RevalidationAttributes RevalidationAttributes::create(HeaderList const& headers)
 {
     RevalidationAttributes attributes;
-    attributes.etag = headers.get("ETag"sv).map([](auto const& etag) { return etag; });
-    attributes.last_modified = parse_http_date(headers.get("Last-Modified"sv));
+    attributes.etag = headers.get("ETag"sv);
+    attributes.last_modified = headers.get("Last-Modified"sv);
 
     return attributes;
 }
--- a/Libraries/LibHTTP/Cache/Utilities.h
+++ b/Libraries/LibHTTP/Cache/Utilities.h
@@ -46,7 +46,7 @@ struct RevalidationAttributes {
     static RevalidationAttributes create(HeaderList const&);
 
     Optional<ByteString> etag;
-    Optional<UnixDateTime> last_modified;
+    Optional<ByteString> last_modified;
 };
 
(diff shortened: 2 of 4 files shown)


trflynn89:
The mystery continues: this morning, Ladybird no longer receives that `+0000` date - we receive "Invalid date" as well:
```
https://api.smartocto.com/api/brands/tentacles?i=i3njb9c9jq6yp04ifcl73wqq94ja7jb7
    x-powered-by: smartocto
    x-cache: HIT from smartocto cache
    max-age: 10
    last-modified: Invalid date
    content-type: application/json; charset=utf-8
    date: Tue, 10 Feb 2026 13:24:02 GMT
    access-control-allow-credentials: true
    access-control-allow-origin: https://tweakers.net
    access-control-allow-headers: Origin, X-Requested-With, Content-Type, Accept
    access-control-allow-methods: GET,PUT,POST,DELETE
    strict-transport-security: max-age=31536000; includeSubDomains; preload
    x-frame-options: DENY
    x-content-type-options: nosniff
    x-xss-protection: 1; mode=block
    referrer-policy: strict-origin-when-cross-origin
    vary: Accept-Encoding
    content-encoding: gzip
```

Perhaps the CDN or smartocto are sometimes intercepting the response headers and parsing the date, as "Invalid date" is the spec output for things like `new Date("foo").toString()`.

PR #7868, opened 2026-02-10
LibWeb: Support :placeholder-shown pseudo-class for textarea elements

Praise-Garfield:
The `:placeholder-shown` pseudo-class was only matching `<input>` elements. `<textarea>` elements with a placeholder attribute were always ignored, breaking floating-label form patterns that use `textarea:placeholder-shown`. This implements the missing textarea case per the HTML spec.

Spec:
https://html.spec.whatwg.org/multipage/semantics-other.html#selector-placeholder-shown

Includes a Text test covering `:placeholder-shown` matching for textareas with and without placeholder attributes and values.

### DIFF
--- a/Libraries/LibWeb/CSS/SelectorEngine.cpp
+++ b/Libraries/LibWeb/CSS/SelectorEngine.cpp
@@ -790,9 +790,13 @@ static inline bool matches_pseudo_class(CSS::Selector::SimpleSelector::PseudoCla
         // - input elements that have a placeholder attribute whose value is currently being presented to the user.
         if (is<HTML::HTMLInputElement>(element) && element.has_attribute(HTML::AttributeNames::placeholder)) {
             auto const& input_element = static_cast<HTML::HTMLInputElement const&>(element);
-            return input_element.placeholder_element() && input_element.placeholder_value().has_value();
+            return input_element.placeholder_element() && input_element.value().is_empty();
         }
-        // - FIXME: textarea elements that have a placeholder attribute whose value is currently being presented to the user.
+        // - textarea elements that have a placeholder attribute whose value is currently being presented to the user.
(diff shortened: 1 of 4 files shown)


Libraries/LibWeb/HTML/HTMLTextAreaElement.h
```diff
@@ -134,6 +134,9 @@ class WEB_API HTMLTextAreaElement final
     // https://html.spec.whatwg.org/multipage/form-elements.html#the-textarea-element:concept-fe-mutable
     virtual bool is_mutable() const override;
 
+    GC::Ptr<DOM::Element> placeholder_element() { return m_placeholder_element; }
```
Psychpsyo: This non-const version doesn't seem needed?

Libraries/LibWeb/CSS/SelectorEngine.cpp
```diff
@@ -792,7 +792,11 @@ static inline bool matches_pseudo_class(CSS::Selector::SimpleSelector::PseudoCla
             auto const& input_element = static_cast<HTML::HTMLInputElement const&>(element);
             return input_element.placeholder_element() && input_element.placeholder_value().has_value();
         }
-        // - FIXME: textarea elements that have a placeholder attribute whose value is currently being presented to the user.
+        // - textarea elements that have a placeholder attribute whose value is currently being presented to the user.
+        if (is<HTML::HTMLTextAreaElement>(element) && element.has_attribute(HTML::AttributeNames::placeholder)) {
+            auto const& textarea_element = static_cast<HTML::HTMLTextAreaElement const&>(element);
+            return textarea_element.placeholder_element() && textarea_element.value().is_empty();
```
gmta, line 798: This is asymmetrical with our HTMLInputElement handling above; let's either use `.placeholder_value()` (which would need to be introduced on HTMLTextAreaElement) or `.value().is_empty()` for both.

PR #7871, opened 2026-02-10
LibWeb/CSP: Apply strict-dynamic to inline scripts

Lubrsi:
This implements https://github.com/w3c/webappsec-csp/pull/787 that fixed the linked spec issue.

This also updates our test infrastructure to import .headers files and apply them, by serving such tests through http-test-server, which will read the .headers file and add them to the response.

### DIFF
--- a/Libraries/LibWeb/ContentSecurityPolicy/Directives/DirectiveOperations.cpp
+++ b/Libraries/LibWeb/ContentSecurityPolicy/Directives/DirectiveOperations.cpp
@@ -24,6 +24,7 @@
 #include <LibWeb/Infra/Strings.h>
 #include <LibWeb/SRI/SRI.h>
 #include <LibWeb/SVG/SVGElement.h>
+#include <LibWeb/SVG/SVGScriptElement.h>
 
 namespace Web::ContentSecurityPolicy::Directives {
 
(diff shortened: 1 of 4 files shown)


Libraries/LibWeb/ContentSecurityPolicy/Directives/DirectiveOperations.cpp
```diff
@@ -938,7 +936,18 @@ MatchResult does_element_match_source_list_for_type_and_source(GC::Ptr<DOM::Elem
 
         // 2. For each expression of list:
         for (auto const& expression : source_list) {
-            // 1. If expression matches the hash-source grammar:
+            // 1. If expression is the "'strict-dynamic'" keyword-source:
+            if (expression.equals_ignoring_ascii_case(KeywordSources::StrictDynamic)) {
+                // 1. If type is "script", and element is not parser-inserted, return "Matches".
+                if (type == Directive::InlineType::Script && element) {
+                    if (auto const* script_element = as_if<HTML::HTMLScriptElement>(element.ptr())) {
```
tcl3: This should probably also be a check for `SVGScriptElement`.

Tests/LibWeb/Fixtures/http-test-server.py
```diff
@@ -40,12 +40,32 @@ class Echo:
...
+            file_path = self.translate_path(self.path)
+            headers_path = file_path + ".headers"
+            if os.path.isfile(headers_path):
+                self._extra_headers = {}
+                with open(headers_path) as f:
+                    for line in f:
+                        line = line.strip()
+                        if ":" in line:
+                            key, _, value = line.partition(":")
+                            self._extra_headers[key.strip()] = value.strip()
```
tcl3: How does this handle `.headers` files that have multiple headers with the same name (for example: `content-security-policy/generic/no-default-src.sub.html.sub.headers`)?

PR #7882, opened 2026-02-11
LibGfx: Add LRU eviction to the HarfBuzz text shaping cache

Praise-Garfield:
`ShapingCache::map` previously grew without bound during a browsing session. This switches it to an `OrderedHashMap` capped at 256 entries with LRU semantics. The `single_ascii_character_map` is already bounded and is left unchanged.

### DIFF
--- a/Libraries/LibGfx/Font/FontDatabase.cpp
+++ b/Libraries/LibGfx/Font/FontDatabase.cpp
@@ -74,6 +74,17 @@ void FontDatabase::for_each_typeface_with_family_name(FlyString const& family_na
     m_system_font_provider->for_each_typeface_with_family_name(family_name, move(callback));
 }
 
+void FontDatabase::clear_all_font_shaping_caches()
+{
+    m_system_font_provider->for_each_typeface([](Typeface const& typeface) {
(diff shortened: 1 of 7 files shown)


Libraries/LibGfx/Font/Font.h
```diff
@@ -85,7 +85,8 @@ class Font : public RefCounted<Font> {
     ShapeFeatures const& features() const { return m_shape_features; }
 
     struct ShapingCache {
-        HashMap<Utf16String, hb_buffer_t*> map;
+        static constexpr size_t max_size = 256;
```
gmta: What's the rationale for using 256? What's the typical figure we need for, let's say, a Wikipedia page to prevent excessive cache busting?

Praise-Garfield: > What's the rationale for using 256? What's the typical figure we need for, let's say, a Wikipedia page to prevent excessive cache busting?

It seemed like a good conservative starting point without me measuring actual data for text heavy sites, aka I was being a bit lazy -- sorry. Looking more closely at the pipeline, I see this would cause problems. I shouldn't of done it like this.

A little edit is incoming.

PR #7889, opened 2026-02-11
WebDriver: Send window handle asynchronously after WebContent connects

tcl3:
This prevents a potential deadlock when tests open many popup windows in quick succession.

This fixes 20 WPT crashes in `html/browsers/the-window-object` and probably more elsewhere. These weren't actually browser or WebDriver crashes, but the WPT harness records a crash when the WebDriver process becomes unresponsive.

No diff when running the `webdriver` WPT tests.

### DIFF
--- a/Services/WebContent/WebDriverClient.ipc
+++ b/Services/WebContent/WebDriverClient.ipc
@@ -16,7 +16,6 @@ endpoint WebDriverClient {
     forward() => (Web::WebDriver::Response response)
     refresh() => (Web::WebDriver::Response response)
     get_title() => (Web::WebDriver::Response response)
-    get_window_handle() => (Web::WebDriver::Response response)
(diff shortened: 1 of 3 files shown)


Services/WebDriver/Session.cpp
```diff
@@ -217,33 +223,47 @@ ErrorOr<NonnullRefPtr<Core::LocalServer>> Session::create_server(NonnullRefPtr<S
         dbgln("WebDriver is connected to WebContent socket");
         auto web_content_connection = maybe_connection.release_value();
 
-        auto maybe_window_handle = web_content_connection->get_window_handle();
```
trflynn89, line 220: From a quick look, the `get_window_handle` seems like it can be removed?

tcl3: Yep, this wasn't used anymore. I've removed it.

Issue #7811, opened 2026-02-07
iframe does not register mouse leave


Psychpsyo:
### Summary

When I hover something in an iframe in ladybird and then drag my mouse out of said iframe, the document inside the iframe does not notice. It still thinks it is hovered.

### Operating system

Linux

### Steps to reproduce

1. Mouse over something hoverable in an iframe that is right next to the frame's edge.
2. Drag your mouse out of the iframe.
3. Notice how the thing remains hovered.

### Expected behavior

The iframe's document should stop being hovered.

### Actual behavior

The iframe's document does not stop being hovered.

### URL for a reduced test case

N/A

### HTML/SVG/etc. source for a reduced test case

```html
<!DOCTYPE html>
<iframe src="data:text/html,<style>html{height:100vh}html:hover{background:red}</style>"></iframe>
```

### Log output and (if possible) backtrace

```shell
none
```

### Screenshots or screen recordings

Just hover the iframe in the reduced repro and then try to make it be not red anymore.

### Build flags or config settings

none

### Contribute a patch?

- [ ] I’ll contribute a patch for this myself.

PR #7890, opened 2026-02-11
LibWeb: Fire mouseleave events when moving mouse from iframe to parent

pranshoe:
Fixes issue #7811 

Tracks the current nested navigable and fires handle_mouseleave() if it changes.

PS: This is my first time contributing to this project, sorry for any inconveniences caused.

### DIFF
--- a/Libraries/LibWeb/Page/EventHandler.cpp
+++ b/Libraries/LibWeb/Page/EventHandler.cpp
@@ -981,6 +981,8 @@ EventResult EventHandler::handle_mousemove(CSSPixelPoint visual_viewport_positio
     GC::Ptr<DOM::Node> node;
 
     ScopeGuard update_hovered_node_and_ui_state_guard = [&] {
+        auto* old_hovered_node = document.hovered_node();
+
         document.set_hovered_node(node);
 
         // FIXME: This check is only approximate. ImageCursors from the same CursorStyleValue share bitmaps, but may
@@ -995,6 +997,14 @@ EventResult EventHandler::handle_mousemove(CSSPixelPoint visual_viewport_positio
             GC::Ptr<HTML::HTMLElement const> hovered_html_element = node
                 ? node->enclosing_html_element_with_attribute(HTML::AttributeNames::title)
                 : nullptr;
+
+            if (old_hovered_node) {
+                if (auto* container = as_if<HTML::NavigableContainer>(*old_hovered_node)) {
+                    if (auto old_content_navigable = container->content_navigable()) {
+                        old_content_navigable->event_handler().handle_mouseleave();
+                    }
+                }
+            }
 
             if (hovered_html_element && hovered_html_element->title().has_value()) {
                 page.client().page_did_enter_tooltip_area(hovered_html_element->title()->to_byte_string());
--- a/Tests/LibWeb/Text/expected/navigable-element-mouseleave.txt
+++ b/Tests/LibWeb/Text/expected/navigable-element-mouseleave.txt
@@ -1,0 +1,2 @@
+Red Background in iframe: true
+Red Background in iframe: false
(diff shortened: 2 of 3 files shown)


Psychpsyo:
I'm pretty sure this won't work in cases where there's multiple nested iframes.
i.e. When there's three nested iframes and the mouse moves from inside the innermost frame, out of all the frames, all the inner frames will also need to receive mouseleave events along the way.

Libraries/LibWeb/Page/EventHandler.cpp
```diff
@@ -1016,14 +1017,30 @@ EventResult EventHandler::handle_mousemove(CSSPixelPoint visual_viewport_positio
         }
     };
 
+    if(paintable){
+        node = dom_node_for_event_dispatch(*paintable);
+        if(auto* iframe_element = as_if<HTML::HTMLIFrameElement>(node.ptr())){
```
Psychpsyo: There is nested navigables that are not iframes and probably also want mouseleave events.
The general class for that is `HTML::NavigableContainer`. (I know, the rest of the file checks for iframes in a bunch of places, I have a different PR open to rectify that: #7810 )

Libraries/LibWeb/Page/EventHandler.cpp
```diff
@@ -995,7 +997,15 @@ EventResult EventHandler::handle_mousemove(CSSPixelPoint visual_viewport_positio
             GC::Ptr<HTML::HTMLElement const> hovered_html_element = node
                 ? node->enclosing_html_element_with_attribute(HTML::AttributeNames::title)
                 : nullptr;
-
+            
+            if(old_hovered_node){
+                if(auto* container = as_if<HTML::NavigableContainer>(*old_hovered_node)){
+                    if(auto old_content_navigable = container->content_navigable()){
```
Psychpsyo: This'll error out if `content_navigable()` is a null pointer. Should probably be `if (auto old_content_navigable = container->content_navigable(); old_content_navigable != nullptr) {`
Also, the spacing is inconsistent with the rest of the code. (on all your `if`s and braces)

Libraries/LibWeb/Page/EventHandler.cpp
```diff
@@ -1023,7 +1033,7 @@ EventResult EventHandler::handle_mousemove(CSSPixelPoint visual_viewport_positio
                 return EventResult::Cancelled;
             }
         }
-
+
```
Psychpsyo: The spaces here shouldn't have been included in the commit.

Libraries/LibWeb/Page/EventHandler.cpp
pranshoe: Shouldn't if(auto* container = as_if<HTML::NavigableContainer>(*old_hovered_node)) automatically evaluate to false if it is a nullptr. Should i explicitly include the condition for just clarity?

Psychpsyo: Oh, true. Probably doesn't need to be explicit then.

pranshoe:
Hi @gmta, could you please look into this as no one is responding.

Zaggy1024:
This feels like it might miss handling the mouse leaving in a few situations. The main one I'm thinking of is when an iframe fills the entire viewport, so EventHandler sees:

- mousemove -> mouse is over iframe, hover the iframe and pass the mousemove through to its EventHandler
- mouseleave -> mouse is no longer over iframe, but we don't send mouseleave to the nested EventHandler

Regarding nested iframes, though, my understanding is that this setup _would_ work if the iframes don't fill their parents, since we send the mousemove through to the nested navigable?

All this said, I have a refactor of mouse event handling in EventHandler that may give us a better chance to consolidate the places that we handle this (#8392). Sorry to cause potential merge conflicts here, especially since my PR doesn't directly address this issue, but hopefully it makes it more reasonable to find/make a consolidated function to handle this edge case.

If you would prefer to address my above concerns here and get it merged first and then I rebase my changes to incorporate the fix, that's fine too.

PR #7896, opened 2026-02-11
LibJS: Drop AST after first compilation

awesomekling:
This branch lets LibJS release AST nodes after bytecode has been generated, reducing memory usage for long-running programs.

Previously, the AST lived forever because runtime operations like class construction, declaration instantiation, and function metadata queries all reached back into AST nodes during execution. We now pre-compute all that information during codegen and store it along with the bytecode.

With the runtime fully decoupled from the AST, each function releases its AST nodes after first compilation. This happens incrementally as functions get compiled and executed. Functions that are never called get to keep their AST nodes forever, for better or worse.

Once we implement lazy parsing, we'll be able to get rid of entire ASTs though :)


awesomekling:
This reduces memory consumption on many large websites by hundreds of megabytes.

Issue #7876, opened 2026-02-10
Right clicking on some embedded JavaScript programs creates a context menu at the wrong position


muzikbike:
### Summary

An issue I originally found on Servo with slightly different effects.

Log output is huge and could not be provided in the dedicated box:
[the log output exceeds 65536 characters.txt](https://github.com/user-attachments/files/25221734/the.log.output.exceeds.65536.characters.txt)

### Operating system

Linux

### Steps to reproduce

1. Go to https://minecraft.wiki/w/An_Ant
2. Scroll down to the LifeViewer embed
3. Right click on it
4. Right click near it outside of it
5. Note where the menu appears

### Expected behavior

It would always appear at the cursor position.

### Actual behavior

It appears offset from the cursor if you target the embed.

### URL for a reduced test case

https://minecraft.wiki/w/An_Ant

### HTML/SVG/etc. source for a reduced test case

```html
N/A
```

### Log output and (if possible) backtrace

```shell
see attachment above
```

### Screenshots or screen recordings

<img width="1920" height="1036" alt="Image" src="https://github.com/user-attachments/assets/c25d6f55-ca22-41b7-931f-454f274fa585" />
<img width="1920" height="1036" alt="Image" src="https://github.com/user-attachments/assets/86139dab-7ed3-4686-b574-dd7416e75682" />

### Contribute a patch?

- [ ] I’ll contribute a patch for this myself.

PR #7898, opened 2026-02-11
LibWeb: Account for scroll and transforms in `to_top_level_position()`

tcl3:
This change means the right click context menu is displayed in the right place when clicking inside an iframe on a scrolled page, including when the iframe has CSS transforms applied to it.

I've not added a test because `to_top_level_position()` is only used for context menus and <select> dropdowns and I can't see an easy way to write an automated test for either of those cases.

Fixes: #7876

### DIFF
--- a/Libraries/LibWeb/HTML/Navigable.cpp
+++ b/Libraries/LibWeb/HTML/Navigable.cpp
@@ -53,6 +53,7 @@
 #include <LibWeb/Painting/DisplayListPlayerSkia.h>
 #include <LibWeb/Painting/NavigableContainerViewportPaintable.h>
 #include <LibWeb/Painting/Paintable.h>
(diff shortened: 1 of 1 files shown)


tcl3:
Here's an example that was previously positioned incorrectly. Clicking on the edge of a nested iframe on a scrolled page with a CSS transform applied:

image

(the screenshot doesn't show the mouse cursor but that's where I clicked, I promise!)

kalenikaliaksandr (changes requested):
can we add a test?

tcl3:
> can we add a test?

Not easily. `to_top_level_position()` isn't observable from JS AFAICT. I could add an internals method that exposes the calculation, but that seems like overkill to me.

PR #7899, opened 2026-02-11
LibCore: Implement ProcessStatistics on Windows

R-Goc:
This commit implements ProcessStatistics on windows. This allows using the task manager view.


Libraries/LibCore/Platform/ProcessStatisticsWindows.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include <psapi.h>
+
+namespace Core::Platform {
+
+ErrorOr<void> update_process_statistics(ProcessStatistics& statistics)
+{
+    FILETIME kernel_time;
+    FILETIME user_time;
+    u64 total_time;
+    static u32 cpu_count = Core::System::hardware_concurrency();
```
shannonbooth, line 22: This does not feel like something that should be static

Libraries/LibCore/Platform/ProcessStatisticsWindows.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include <LibCore/Platform/ProcessStatistics.h>
+#include <LibCore/System.h>
+#include <psapi.h>
+
+namespace Core::Platform {
+
+ErrorOr<void> update_process_statistics(ProcessStatistics& statistics)
+{
+    FILETIME kernel_time;
+    FILETIME user_time;
```
shannonbooth, line 20: Let's not leave all of these variable uninitialized

Libraries/LibCore/Platform/ProcessStatisticsWindows.cpp
```diff
@@ -0,0 +1,63 @@
...
+    u64 total_time_diff = total_time - statistics.total_time_scheduled;
+    statistics.total_time_scheduled = total_time;
+
+    for (auto const& process : statistics.processes) {
+        FILETIME process_creation_time;
+        FILETIME process_exit_time;
+        FILETIME process_kernel_time;
+        FILETIME process_user_time;
+
+        HANDLE process_handle = OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION, FALSE, process->pid);
```
shannonbooth, line 37: This handle does not seem closed anywhere

Libraries/LibCore/Platform/ProcessStatisticsWindows.cpp
```diff
@@ -0,0 +1,63 @@
...
+        result = GetProcessTimes(process_handle, &process_creation_time, &process_exit_time, &process_kernel_time, &process_user_time);
+        if (!result)
+            continue;
+
+        u64 time_in_process = filetime_to_u64(process_kernel_time) + filetime_to_u64(process_user_time);
+        u64 time_in_process_diff = time_in_process - process->time_spent_in_process;
+        process->time_spent_in_process = time_in_process;
+
+        if (total_time_diff > 0) {
+            process->cpu_percent = ((f32)time_in_process_diff / (f32)total_time_diff / (f32)cpu_count) * 100.0f;
```
shannonbooth, line 50: Let's prefer static_cast over c-style casts

AK/Windows.h
```diff
@@ -118,4 +119,10 @@ inline void windows_shutdown()
 {
     terminate_wsa();
 }
+
+ALWAYS_INLINE constexpr u64 filetime_to_u64(FILETIME const& ft)
```
shannonbooth, line 123: 🤔  I wonder if this should be part of `Duration`.

It wouldn't be constexpr, but the windows documentation recommends using ULARGE_INTEGER instead of doing this manually at least

Libraries/LibCore/Platform/ProcessStatisticsWindows.cpp
R-Goc: I feel like static makes sense here to avoid calling hardware_concurrency() all the time

AK/Windows.h
R-Goc: We aren't moving the data back to a FILETIME so there is no need to use ULARGE_INTEGER. It also can't be cast directly to a ULARGE_INTEGER* or u64* as the alignment is different. What they mean in the documentation is not to try to do math on low and high parts manually everywhere and instead do math on the 64bit union member of ULARGE_INTEGER and for convenience use the low and high part members to recreate the FILETIME. I did think of adding it to Duration. I don't have much opinion on it. putting it in AK/Windows.h seemed unobstructive but I can move it to Time. Other time related helper functions will live there after all.

Libraries/LibCore/Platform/ProcessStatisticsWindows.cpp
shannonbooth: Its not very consistent with other callers, and honestly I am not sure if that value is guaranteed to be the same. If needbe we can make the impl of that function have a static variable

AK/Windows.h
shannonbooth: Sure, makes sense to leave for now, can perhaps shift around in future

Libraries/LibCore/Platform/ProcessStatisticsWindows.cpp
R-Goc: Making the impl of that static seems good. I don't recall any systems that allow hot-swapping CPUs or disabling cores at runtime.

PR #7905, opened 2026-02-12
LibWeb: Do not allow margins to collapse

### LINKED ISSUE #7794: Top margin of first child of flow root with an empty IFC child disappears. (breaks on Internet Archive)
### Summary

See title and reduced reproduction.
The `display:flow-root` on `body` is to prevent the body's top margin from interfering with the margin collapsing state.
The second `flow-root` can alternatively also be a table or grid/flex container. Anything that establishes a separate formatting context is fine.

This prevents us from scrolling to the bottom of https://archive.org/details/mspaint_xp_version. (The page only loads further content once it's scrolled far enough and since we're missing a gap, we never get far enough.)

### Operating system

Linux

### Steps to reproduce

1. Open the reduced test case.
2. Notice that the word FOO is at the top of the page. It should be shoved down by 140px.

### Expected behavior

There should be a gap there.

### Actual behavior

There is no gap there.

N/A

```html
<!DOCTYPE html>
<body style="display:flow-root">
<div style="margin-top:140px">
  <div style="display:flow-root"></div>
</div>
FOO
```

```shell
none
```

Ladybird (left) vs Firefox (right)
Image

none

### Contribute a patch?

ChaseKnowlden:
Allows us to scroll down to the bottom of Internet Archive pages.
Closes #7794

### DIFF
--- a/Libraries/LibWeb/Layout/BlockFormattingContext.cpp
+++ b/Libraries/LibWeb/Layout/BlockFormattingContext.cpp
@@ -60,10 +60,11 @@ CSSPixels BlockFormattingContext::automatic_content_height() const
 
 static bool margins_collapse_through(Box const& box, LayoutState& state)
 {
-    // FIXME: A box's own margins collapse if the 'min-height' property is zero, and it has neither top or bottom borders
-    // nor top or bottom padding, and it has a 'height' of either 0 or 'auto', and it does not contain a line box, and
-    // all of its in-flow children's margins (if any) collapse.
     // https://www.w3.org/TR/CSS22/box.html#collapsing-margins
+    // A box's own margins collapse if the 'min-height' property is zero, and it has neither top or bottom borders
+    // nor top or bottom padding, and it has a 'height' of either 0 or 'auto', and it does not contain a line box,
+    // all of its in-flow children's margins (if any) collapse, and it does not establish a new formatting context.
+
     // FIXME: For the purpose of margin collapsing (CSS 2 §8.3.1 Collapsing margins), if the block axis is the
     //        ratio-dependent axis, it is not considered to have a computed block-size of auto.
     //        https://www.w3.org/TR/css-sizing-4/#aspect-ratio-margin-collapse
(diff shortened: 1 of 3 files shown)


Psychpsyo:
Thanks a lot for doing this! Can you add a test case to the PR so that this doesn't break again in the future?

Putting a simple HTML file that exhibits this behavior in `Tests/LibWeb/Layout/input/` followed by running `./Meta/ladybird.py run test-web --rebaseline` should be enough.

Libraries/LibWeb/Layout/BlockFormattingContext.cpp
```diff
@@ -60,18 +60,35 @@ CSSPixels BlockFormattingContext::automatic_content_height() const
...
+    // https://www.w3.org/TR/css-box-4/#collapsing-margins
+    if (FormattingContext::creates_block_formatting_context(box))
+        return false;
+
+    if (state.get(box).border_box_height() != 0)
+        return false;
+
+    // All of its in-flow children's margins (if any) must also collapse.
+    for (auto const* child = box.first_child_of_type<Box>(); child; child = child->next_sibling_of_type<Box>()) {
+        if (child->is_absolutely_positioned() || child->is_floating())
```
gmta: You can use this instead:

```cpp
if (child->is_out_of_flow())
```

ChaseKnowlden: Fixed

Libraries/LibWeb/Layout/BlockFormattingContext.cpp
```diff
@@ -60,18 +60,35 @@ CSSPixels BlockFormattingContext::automatic_content_height() const
 
 static bool margins_collapse_through(Box const& box, LayoutState& state)
 {
-    // FIXME: A box's own margins collapse if the 'min-height' property is zero, and it has neither top or bottom borders
+    // https://www.w3.org/TR/CSS22/box.html#collapsing-margins
+    // A box's own margins collapse if the 'min-height' property is zero, and it has neither top or bottom borders
     // nor top or bottom padding, and it has a 'height' of either 0 or 'auto', and it does not contain a line box, and
     // all of its in-flow children's margins (if any) collapse.
```
gmta: You've resolved this FIXME but you're not checking for the value of the `min-height` property. Is this implicitly dealt with by checking `.border_box_height()` below? If so, you might want to address this in a `// NB: ...` comment and move this down. Otherwise, we should probably keep the FIXME.

ChaseKnowlden: Added a note

PR #7906, opened 2026-02-12
LibWeb: A couple of fixes for `stroke-dasharray` CSS property

Calme1709:


Libraries/LibWeb/CSS/ComputedProperties.cpp
```diff
@@ -483,16 +483,25 @@ Vector<Variant<LengthPercentage, float>> ComputedProperties::stroke_dasharray()
...
+                dashes.append(LengthPercentage { value->as_calculated() });
+            else if (calculated_value.resolves_to_number())
+                dashes.append(calculated_value.resolve_number({}).value());
+            else
+                VERIFY_NOT_REACHED();
+
+        } else if (value->is_number())
             dashes.append(value->as_number().number());
-        else
+        else {
```
AtkinsSJ: Nit: Missing `{}` around this block.

Calme1709: Fixed

PR #7909, opened 2026-02-12
LibWeb+LibMedia: MediaCapture API wiring

jonbgamble:
Ladybird is not particularly close to needing media capture, with webaudio being niche and no webRTC in sight, but on the other hand this code is pretty isolated and safe. And it's probably good for a few free wins in some random wpt test.

@Zaggy1024 - this has stubs in LibMedia that mirror [my tree](https://github.com/jonbgamble/ladybird/tree/webaudio-worky/Libraries/LibMedia/MediaCapture) but I reckon my choice of places/names should be uncontroversial. Let me know if you had something different in mind.


Libraries/LibWeb/MediaCapture/MediaStream.cpp
```diff
@@ -0,0 +1,197 @@
...
+void MediaStream::remove_track(GC::Ref<MediaStreamTrack> track)
+{
+    // 2. If track is not in stream's track set, then abort these steps.
+    auto removed = m_tracks.remove_first_matching([&](auto const& existing_track) {
+        return existing_track.ptr() == track.ptr();
+    });
+    if (!removed)
+        return;
+
+    // 3. Remove track from stream's track set.
```
Zaggy1024: Where's step 1 here?

Steps 2 and 3 should be adjacent above the implementation, since they are implemented out of order.

Libraries/LibWeb/MediaCapture/MediaStream.cpp
```diff
@@ -0,0 +1,197 @@
...
+    auto event = MediaStreamTrackEvent::create(realm(), HTML::EventNames::removetrack, event_init);
+    dispatch_event(event);
+}
+
+// https://w3c.github.io/mediacapture-main/#dom-mediastream-clone
+GC::Ref<MediaStream> MediaStream::clone() const
+{
+    // 1. Let streamClone be a newly constructed MediaStream object.
+    // 2. Initialize streamClone.MediaStream.id to a newly generated value.
+    auto stream_clone = create(realm());
```
Zaggy1024, line 151: Do we actually initialize an ID here? I don't think `construct_impl()` is invoked here.

Libraries/LibWeb/MediaCapture/MediaStreamTrack.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+static constexpr auto audio_track_kind = static_cast<Bindings::MediaStreamTrackKind>(0);
+static constexpr auto video_track_kind = static_cast<Bindings::MediaStreamTrackKind>(1);
+
+MediaStreamTrack::MediaStreamTrack(JS::Realm& realm)
+    : DOM::EventTarget(realm)
+{
+}
+
+// https://w3c.github.io/mediacapture-main/#mediastreamtrack
```
Zaggy1024, line 30: https://w3c.github.io/mediacapture-main/#dfn-create-a-mediastreamtrack

Libraries/LibWeb/MediaCapture/MediaStreamTrack.cpp
```diff
@@ -0,0 +1,225 @@
...
+GC::Ref<MediaStreamTrack> MediaStreamTrack::create_audio_input_track(JS::Realm& realm, Media::AudioDeviceInfo const& device_info, Optional<String> label)
+{
+    // 1. Let track be a new object of type source's MediaStreamTrack source type.
+    auto track = realm.create<MediaStreamTrack>(realm);
```
Zaggy1024, line 35: I'm assuming that the `source` here is deferred until later work (or the spec doesn't have other types yet?). This should probably have a FIXME.

Libraries/LibWeb/MediaCapture/MediaStreamTrack.cpp
```diff
@@ -0,0 +1,225 @@
...
+    track->m_provider_id = s_next_provider_id.fetch_add(1, AK::MemoryOrder::memory_order_relaxed);
+    // 2. If mediaDevicesToTieSourceTo is not null, tie track source to MediaDevices with source and mediaDevicesToTieSourceTo.
+    // FIXME: Tie track sources to MediaDevices once sources are modeled.
+    // 3. Run source's MediaStreamTrack source-specific construction steps with track as parameter.
+    // 4. Return track.
+    return track;
+}
+
+// https://w3c.github.io/mediacapture-main/#mediastreamtrack
+GC::Ref<MediaStreamTrack> MediaStreamTrack::create_audio_output_track(JS::Realm& realm, Media::AudioDeviceInfo const& device_info, Optional<String> label)
```
Zaggy1024: This function looks very similar to if not the same as the audio input track function above it, is there a reason for the separation? Can we split off some subset of the functionality here instead of duplicating it all?

Libraries/LibWeb/MediaCapture/MediaStreamTrack.cpp
```diff
@@ -0,0 +1,225 @@
...
+{
+    // 1. Let track be a new object of type source's MediaStreamTrack source type.
+    auto track = realm.create<MediaStreamTrack>(realm);
```
Zaggy1024: FIXME for `source` here too?

Libraries/LibWeb/MediaCapture/MediaStreamTrack.cpp
```diff
@@ -0,0 +1,225 @@
...
+// https://w3c.github.io/mediacapture-main/#dom-mediastreamtrack-stop
+void MediaStreamTrack::stop()
+{
+    // 1. Let track be the current MediaStreamTrack object.
+    // 2. If track's [[ReadyState]] is "ended", then abort these steps.
+    if (m_state == Bindings::MediaStreamTrackState::Ended)
+        return;
+
+    // 3. Notify track's source that track is ended.
+    // FIXME: Do that.
```
Zaggy1024:
```suggestion
    // FIXME: 3. Notify track's source that track is ended.
```

Libraries/LibWeb/MediaCapture/MediaStreamTrack.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+    // 4. Set track's [[ReadyState]] to "ended".
+    m_state = Bindings::MediaStreamTrackState::Ended;
+
+    // AD-HOC: Dispatch an ended event so listeners can react to capture teardown.
+    dispatch_event(DOM::Event::create(realm(), HTML::EventNames::ended));
```
Zaggy1024: There is spec text for this, it's not ad-hoc:

https://w3c.github.io/mediacapture-main/#ends-nostop

Make sure to queue a task for this as the spec asks.

Libraries/LibWeb/MediaCapture/MediaStreamTrack.cpp
```diff
@@ -0,0 +1,225 @@
...
+    track_clone->m_id = Crypto::generate_random_uuid();
+    track_clone->m_kind = m_kind;
+    track_clone->m_label = m_label;
+    track_clone->m_enabled = m_enabled;
+    track_clone->m_muted = m_muted;
+    track_clone->m_device_info = m_device_info;
+    track_clone->m_provider_id = s_next_provider_id.fetch_add(1, AK::MemoryOrder::memory_order_relaxed);
+
+    // 8. Run source MediaStreamTrack source-specific clone steps with track and trackClone as parameters.
+    // FIXME: Do 8.
```
Zaggy1024:
```suggestion
    // FIXME: 8. Run source MediaStreamTrack source-specific clone steps with track and trackClone as parameters.
```

Libraries/LibWeb/MediaCapture/MediaStreamTrack.cpp
```diff
@@ -0,0 +1,225 @@
...
+    dispatch_event(DOM::Event::create(realm(), HTML::EventNames::ended));
+}
+
+// https://w3c.github.io/mediacapture-main/#dom-mediastreamtrack-clone
+GC::Ref<MediaStreamTrack> MediaStreamTrack::clone() const
+{
+    // When the clone() method is invoked, the User Agent MUST return the result of clone a track with this.
+    // https://w3c.github.io/mediacapture-main/#clone-a-track
+    // 1. Let track be the MediaStreamTrack object to be cloned.
+    // 2. Let source be track's [[Source]].
```
Zaggy1024: Since the source slot is not implemented (from what I can tell):

```suggestion
    // FIXME: 2. Let source be track's [[Source]].
```

Libraries/LibWeb/MediaCapture/MediaDevices.cpp
```diff
@@ -0,0 +1,758 @@
...
+
+// https://w3c.github.io/mediacapture-main/#device-information-exposure
+bool MediaDevices::microphone_information_can_be_exposed()
+{
+    // 1. If any of the local devices of kind "audioinput" are attached to a live MediaStreamTrack in
+    // mediaDevices's relevant global object's associated Document, return true.
+    for (auto const& entry : m_devices_live_map) {
+        if (entry.value)
+            return true;
+    }
```
Zaggy1024: Something like
```suggestion
    for (auto const& [device_name, live] : m_devices_live_map) {
        if (live)
            return true;
    }
```
would be much clearer here. Also, I may be missing something here, but shouldn't we have some check for the device kind?

Libraries/LibWeb/MediaCapture/MediaDevices.cpp
```diff
@@ -0,0 +1,758 @@
...
+
+    // To perform creating a list of device info objects, given mediaDevices and deviceList, run the following steps:
+
+    // 1. Let resultList be an empty list.
+    Vector<GC::Ref<MediaDeviceInfo>> result_list;
+
+    // 2. Let microphoneList, cameraList and otherDeviceList be empty lists.
+    Vector<GC::Ref<MediaDeviceInfo>> microphone_list;
+    Vector<GC::Ref<MediaDeviceInfo>> camera_list;
+    Vector<GC::Ref<MediaDeviceInfo>> other_device_list;
```
Zaggy1024: Refs in vectors on the stack look very suspicious to me, a comment should justify it if this is safe.

Libraries/LibWeb/MediaCapture/MediaDevices.cpp
```diff
@@ -0,0 +1,758 @@
...
+        // 8.3 Let deviceInfo be the result of creating a device info object to represent device, with mediaDevices.
+        Optional<GC::Ref<MediaDeviceInfo>> device_info = create_device_info_object(device);
+        if (!device_info.has_value())
+            continue;
+
+        // 8.4 Append deviceInfo to otherDeviceList.
+        // 8.5 If device is the system default audio output, prepend it instead.
+        if (device.kind == "audiooutput"sv && device.is_default) {
+            // 8.5.2 Set defaultAudioOutputInfo's deviceId to default.
+            // FIXME: MediaDeviceInfo has no mutable deviceId; synthesize this object once mutable construction is supported.
```
Zaggy1024: What does this comment mean by mutable construction? Isn't it as simple as adding a setter (possibly with a badge)?

If there's more to this step than that, then I'd suggest just making the step itself a fixme, it doesn't feel like the mutability is the issue here.

Libraries/LibWeb/MediaCapture/MediaDevices.cpp
```diff
@@ -0,0 +1,758 @@
...
+        if (a[i]->label() != b[i]->label())
+            return false;
+        if (a[i]->group_id() != b[i]->group_id())
+            return false;
+    }
+
+    return true;
+}
+
+// https://w3c.github.io/mediacapture-main/#mediadevices - When new media input and/or output devices...
```
Zaggy1024:
```suggestion
// https://w3c.github.io/mediacapture-main/#dfn-device-change-notification-steps
```

Libraries/LibWeb/MediaCapture/MediaDevices.cpp
```diff
@@ -0,0 +1,758 @@
...
+    // proceed is true, but for no other MediaDevices object:
+
+    // 1. Let lastExposedDevices be the result of creating a list of device info objects with mediaDevices and mediaDevices.[[storedDeviceList]].
+    Vector<GC::Ref<MediaDeviceInfo>> last_exposed_devices = create_list_of_device_info_objects(m_stored_device_list);
+
+    // 2. Let deviceList be the list of all media input and/or output devices available to the User Agent.
+    // device_list is the caller-provided snapshot.
+
+    // 3. Let newExposedDevices be the result of creating a list of device info objects with mediaDevices and deviceList.
+    Vector<GC::Ref<MediaDeviceInfo>> new_exposed_devices = create_list_of_device_info_objects(device_list);
```
Zaggy1024: Vector of ref on the stack again.

Libraries/LibWeb/MediaCapture/MediaDevices.cpp
```diff
@@ -0,0 +1,758 @@
...
+    JS::Realm& realm = this->realm();
+
+    // 1. Let p be a new promise.
+    GC::Ref<WebIDL::Promise> promise = WebIDL::create_promise(realm);
+
+    // 2. Let proceed be the result of device enumeration can proceed with this.
+    bool proceed = device_enumeration_can_proceed();
+
+    // 3. Let mediaDevices be this.
+    GC::Root<MediaDevices> media_devices = GC::make_root(this);
```
Zaggy1024: Making a variable to match the spec here is fine with me, but rooting it is unnecessary. You could even have a raw pointer here and it would show up in the conservative scan, but it would probably make the most sense to use GC::Ref.

Libraries/LibWeb/MediaCapture/MediaDevices.cpp
```diff
@@ -0,0 +1,758 @@
...
+    // 10. Let p be a new promise.
+    if (video_requested)
+        return WebIDL::create_rejected_promise(realm, WebIDL::NotSupportedError::create(realm, "Video capture is not supported"_utf16));
+
+    auto promise = WebIDL::create_promise(realm);
+
+    auto promise_root = GC::make_root(promise);
+    auto media_devices = GC::make_root(this);
+
+    // 11.1 While isInView is false, the User Agent MUST wait to proceed to the next step until a task queued to set isInView to the result of the is in view algorithm, would set isInView to true.
```
Zaggy1024: Where did step 11 go? Since step 11 is to run in parallel, the steps contained by it should run in a task.

Also, always prefer to use blocks (even without any attached statement) to make indentation indicate step nesting instead of using chained numbers like `11.1`. In this case, the task callback would take care of the indentation.

Libraries/LibWeb/MediaCapture/MediaDevices.cpp
```diff
@@ -0,0 +1,758 @@
...
+    // Returns a dictionary whose members are the constrainable properties known to the User Agent.
+    MediaTrackSupportedConstraints supported_constraints;
+    supported_constraints.device_id = true;
+    return supported_constraints;
+}
+
+// https://w3c.github.io/mediacapture-main/#dom-mediadevices-getusermedia
+GC::Ref<WebIDL::Promise> MediaDevices::get_user_media(Optional<MediaStreamConstraints> const& constraints)
+{
+    // FIXME: Permission, visibility, and device selection. Honor "run these steps in parallel".
```
Zaggy1024: FIXMEs should go on their corresponding steps, I have no idea what this is referring to without going searching.

Libraries/LibWeb/MediaCapture/MediaDevices.cpp
```diff
@@ -0,0 +1,758 @@
...
+        // 11.3.1 For each possible configuration of each possible source device of media of type kind, conceive a candidate as a placeholder for an eventual MediaStreamTrack holding a source device and configured with a settings dictionary comprised of its specific settings.
+        //     Call this set of candidates the candidateSet.
+        Vector<Media::AudioDeviceInfo> candidate_set = move(devices);
+
+        // 11.13 NotFound Failure.
+        // 11.13.2 Reject p with a new DOMException object whose name attribute has the value "NotFoundError".
+        if (candidate_set.is_empty()) {
+            WebIDL::reject_promise(realm, *promise, WebIDL::NotFoundError::create(realm, "No audio input devices available"_utf16));
+            return promise;
+        }
```
Zaggy1024: The spec step ordering here is a bit strange, it seems like the best approach to maintain natural step nesting would be to make a lambda for step 13 above step 1, then invoke it and return in the condition here.

The `If candidateSet is the empty set, jump to the step labeled NotFound Failure below.` condition is missing here as well, so it's not clear in the code why this is here.

Libraries/LibWeb/MediaCapture/MediaDevices.cpp
```diff
@@ -0,0 +1,758 @@
...
+            }
+            candidate_set = move(filtered_candidates);
+
+            if (candidate_set.is_empty()) {
+                // 11.14 Constraint Failure.
+                // 11.14.3 Reject p with a new OverconstrainedError created by calling OverconstrainedError(constraint, message).
+                // FIXME: Add OverconstrainedError plumbing and reject with it when constraints cannot be satisfied.
+                WebIDL::reject_promise(realm, *promise, WebIDL::NotFoundError::create(realm, "Requested audio input device not found"_utf16));
+                return promise;
+            }
```
Zaggy1024: Same sort of issue here, the condition is missing and the step nesting forces chained step numbers.

Libraries/LibWeb/MediaCapture/MediaDevices.cpp
```diff
@@ -0,0 +1,758 @@
...
+        }
+
+        // 11.3.6 Read the current permission state for all candidate devices in candidateSet that are not attached to a live MediaStreamTrack in the current Document. Remove from candidateSet any candidate whose device's permission state is "denied".
+        // FIXME: Integrate permission state checks and remove denied candidates.
```
Zaggy1024:
```suggestion
        // FIXME: 6. Read the current permission state for all candidate devices in candidateSet that are not attached to a live MediaStreamTrack in the current Document. Remove from candidateSet any candidate whose device's permission state is "denied".
```

You get the idea, any cases like this where the FIXME _is_ the step, it should precede the step text.

Zaggy1024 (commented):
Looks decent to me, but a lot of nits here.

Libraries/LibWeb/MediaCapture/MediaDevices.cpp
jonbgamble: I don't think we want to paper over device info from the device layer if it wrongly has a non-default label or id on a default device.

Zaggy1024: I don't understand your point. If we're papering over some label or ID, then is that not a spec bug? It's not our job to ignore pieces of the spec because they seem wrong, we have to find the root cause and report it if it's their issue. Also, I'm not sure what other option we have here other than modifying the device ID.

jonbgamble: My point is the device ID should already be correct.

jonbgamble:
Are we close enough to the finish line here where you can just merge and fix the last remaining nits yourself? (since weeks/months pass between every interaction)

Libraries/LibWeb/MediaCapture/MediaDevices.cpp
Zaggy1024: So somewhere else is setting the deviceId to default before this? If so, then this should instead be noting where that's happening.

Zaggy1024:
It felt close based on my last review, it was mostly nits.

Zaggy1024 (approved):

Issue #7874, opened 2026-02-10
RPCS3 website crashes pretty darn quick


muzikbike:
### Summary

The website for RPCS3 barely loads, crashing in a near instant. Log output seems quite repetitive.

### Operating system

Linux

### Steps to reproduce

1. Visit https://rpcs3.net/
2. Wait for a second
3. Notice the almost immediate crash

### Expected behavior

The site would load like it does on common browsers.

### Actual behavior

It, well, crashes.

### URL for a reduced test case

https://rpcs3.net/

### HTML/SVG/etc. source for a reduced test case

```html
N/A
```

### Log output and (if possible) backtrace

```shell
ninja: Entering directory `/home/conputer/other-dev/ladybird/20250807210822/ladybird/Build/release'
ninja: no work to do.
3843.488 WebContent(25412): Handling message while connection closed: WebContentServer::CloseServer
3843.831 WebContent(25423): ResourceLoader: Filtered request to: "https://static.cloudflareinsights.com/beacon.min.js/vcd15cbe7772f49c399c6a5babf22c1241717689176015"
VERIFICATION FAILED: box_used_values.offset.is_zero() || box.is_list_item_marker_box() at /home/conputer/other-dev/ladybird/20250807210822/ladybird/Libraries/LibWeb/Layout/FormattingContext.cpp:1189
Stack trace (most recent call first):
#0  0x000077d8d594f543 in content_box_rect_in_static_position_ancestor_coordinate_space at /home/conputer/other-dev/ladybird/20250807210822/ladybird/Libraries/LibWeb/Layout/FormattingContext.cpp:1189:5
#1  0x000077d8d594fcc6 in layout_absolutely_positioned_element at /home/conputer/other-dev/ladybird/20250807210822/ladybird/Libraries/LibWeb/Layout/FormattingContext.cpp:1386:36
#2  0x000077d8d5931a93 in parent_context_did_dimension_child_root_box at /home/conputer/other-dev/ladybird/20250807210822/ladybird/Libraries/LibWeb/Layout/BlockFormattingContext.cpp:157:13
#3  0x000077d8d5995d6f in compute_table_height at /home/conputer/other-dev/ladybird/20250807210822/ladybird/Libraries/LibWeb/Layout/TableFormattingContext.cpp:1011:45
[...]

Issue #6905, opened 2025-11-22
Crash on some Amazon search result pages (`VERIFICATION FAILED: box_used_values.offset.is_zero()`)


InvalidUsernameException:
Visiting the Amazon search results for e.g. ["hello kitty coffee mug"](https://www.amazon.com/s?k=hello%20kitty%20coffee%20mug) causes Ladybird to crash.

Reduced test case:
```html
<!DOCTYPE html>
<style>
.table {
  display: table;
}
.table-cell {
  display: table-cell;
  height: 100%;
}
.inside {
  position: relative;
  padding-left: 8px;
}
.inside:after {
  position: absolute;
  content: "";
  top: 0;
  left: 0;
}
</style><div class="table"><div class="table-cell"><div class="inside"></div></div></div>
```

```
data:text/html;base64,PCFET0NUWVBFIGh0bWw+CjxzdHlsZT4KLnRhYmxlIHsKICBkaXNwbGF5OiB0YWJsZTsKfQoudGFibGUtY2VsbCB7CiAgZGlzcGxheTogdGFibGUtY2VsbDsKICBoZWlnaHQ6IDEwMCU7Cn0KLmluc2lkZSB7CiAgcG9zaXRpb246IHJlbGF0aXZlOwogIHBhZGRpbmctbGVmdDogOHB4Owp9Ci5pbnNpZGU6YWZ0ZXIgewogIHBvc2l0aW9uOiBhYnNvbHV0ZTsKICBjb250ZW50OiAiIjsKICB0b3A6IDA7CiAgbGVmdDogMDsKfQo8L3N0eWxlPjxkaXYgY2xhc3M9InRhYmxlIj48ZGl2IGNsYXNzPSJ0YWJsZS1jZWxsIj48ZGl2IGNsYXNzPSJpbnNpZGUiPjwvZGl2PjwvZGl2PjwvZGl2Pg==
```

```
VERIFICATION FAILED: box_used_values.offset.is_zero() || box.is_list_item_marker_box() at Libraries/LibWeb/Layout/FormattingContext.cpp:1184
Stack trace (most recent call first):
#0  0x00007fb585dd3786 in content_box_rect_in_static_position_ancestor_coordinate_space at Libraries/LibWeb/Layout/FormattingContext.cpp:1184:5
#1  0x00007fb585dd3a68 in layout_absolutely_positioned_element at Libraries/LibWeb/Layout/FormattingContext.cpp:1264:36
[...]

PR #6523, opened 2025-10-20
LibWeb: fix VERIFICATION FAILED: box_used_values.offset.is_zero()

9cb14c1ec0:
When doing a search on Amazon, the following error is encountered:

```
VERIFICATION FAILED: box_used_values.offset.is_zero() at /home/edwin/ladybird/Libraries/LibWeb/Layout/FormattingContext.cpp:1170
Stack trace (most recent call first):
#0   0x00007502eff58f13 in content_box_rect_in_static_position_ancestor_coordinate_space at /home/edwin/ladybird/Libraries/LibWeb/Layout/FormattingContext.cpp:1170:5
#1   0x00007502eff591f5 in layout_absolutely_positioned_element at /home/edwin/ladybird/Libraries/LibWeb/Layout/FormattingContext.cpp:1250:36
#2   0x00007502eff3d893 in parent_context_did_dimension_child_root_box at /home/edwin/ladybird/Libraries/LibWeb/Layout/BlockFormattingContext.cpp:146:13
#3   0x00007502eff44fda in run at /home/edwin/ladybird/Libraries/LibWeb/Layout/FlexFormattingContext.cpp:210:49
#4   0x00007502eff402ae in layout_block_level_box at /home/edwin/ladybird/Libraries/LibWeb/Layout/BlockFormattingContext.cpp:839:41
#5   (inlined)          in operator() at /home/edwin/ladybird/Libraries/LibWeb/Layout/BlockFormattingContext.cpp:902:9
#6   (inlined)          in for_each_child_of_type<Web::Layout::Box, (lambda at /home/edwin/ladybird/Libraries/LibWeb/Layout/BlockFormattingContext.cpp:901:49)> at /home/edwin/ladybird/Libraries/LibWeb/TreeNode.h:313:21
#7   (inlined)          in for_each_child_of_type<Web::Layout::Box, (lambda at /home/edwin/ladybird/Libraries/LibWeb/Layout/BlockFormattingContext.cpp:901:49)> at /home/edwin/ladybird/Libraries/LibWeb/TreeNode.h:322:54
#8   0x00007502eff3d48b in layout_block_level_children at /home/edwin/ladybird/Libraries/LibWeb/Layout/BlockFormattingContext.cpp:901:21
#9   0x00007502eff40465 in layout_block_level_box at /home/edwin/ladybird/Libraries/LibWeb/Layout/BlockFormattingContext.cpp:858:13
[...]

PR #7911, opened 2026-02-12
LibWeb: Reset abspos offset before static position computation

tcl3:
When a table cell has a percentage height, `TableFormattingContext` performs two layout passes on the cell. If the cell contains an absolutely positioned descendant, the first pass sets its offset via `layout_absolutely_positioned_element()`. On the second pass, the stale non-zero offset triggers an assertion in
`content_box_rect_in_static_position_ancestor_coordinate_space()`.

We now reset the offset to zero at the start of the offset computation in `layout_absolutely_positioned_element()`.

Fixes: #6905
Fixes: #7874

### DIFF
--- a/Libraries/LibWeb/Layout/FormattingContext.cpp
+++ b/Libraries/LibWeb/Layout/FormattingContext.cpp
@@ -1460,6 +1460,12 @@ void FormattingContext::layout_absolutely_positioned_element(Box const& box, Abs
         }
     }
 
+    // AD-HOC: Reset offset to zero before recomputing. Table cells with percentage heights are laid out twice - before
+    //         and after table height is determined. Each pass re-enters layout_absolutely_positioned_element for abspos
+    //         descendants. Without this reset, the stale offset from the first pass triggers the assertion in
+    //         content_box_rect_in_static_position_ancestor_coordinate_space.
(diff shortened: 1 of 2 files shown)


tcl3:
I've just noticed this is the same fix as #6523. I've added a comment that expands on when this condition is triggered and why we reset at this point.

Libraries/LibWeb/Layout/FormattingContext.cpp
```diff
@@ -1460,6 +1460,12 @@ void FormattingContext::layout_absolutely_positioned_element(Box const& box, Abs
         }
     }
 
+    // AD-HOC: Reset offset to zero before recomputing. Table cells with percentage heights are laid out twice - before
+    //         and after table height is determined. Each pass re-enters layout_absolutely_positioned_element for abspos
+    //         descendants. Without this reset, the stale offset from the first pass triggers the assertion in
+    //         content_box_rect_in_static_position_ancestor_coordinate_space.
```
gmta, line 1466: Is this still relevant? I don't think `content_box_rect_in_static_position_ancestor_coordinate_space` exists anymore, and I can't reproduce the crashes in the linked issues.

tcl3:
Not needed after #7952. Closing.

PR #7913, opened 2026-02-12
LibWeb: Skip full style update in getComputedStyle when possible

awesomekling:
Before calling update_style() for a getComputedStyle property access, we now check whether the target element actually needs a style update by walking its ancestor chain. If neither the element nor any of its ancestors have dirty style bits, and there are no document-level reasons to recalculate style, we skip the update_style() call entirely.

This avoids unnecessary work when scripts access computed style properties on elements whose styles are already up-to-date, which is a common pattern on the web.


Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -1741,6 +1742,31 @@ void Document::update_style()
...
+        return true;
+    if (m_needs_invalidation_of_elements_affected_by_has)
+        return true;
+    if (m_style_invalidator->has_pending_invalidations())
+        return true;
+
+    // Walk from the element to the document root. If any node on the path
+    // has its own style marked dirty, the element's computed style could
+    // change (e.g. via inherited properties), so we must update.
+    for (Node const* node = &abstract_element.element(); node; node = node->parent_or_shadow_host()) {
```
AtkinsSJ: I suspect this should use `node->flat_tree_parent()` instead of `node->parent_or_shadow_host()`? Otherwise slotted elements don't look at their slot.

awesomekling: Okay well you were totally right about this, and boy did it send me down a rabbit hole. Updated to fix that + make style inheritance work more correctly for slotted elements.

PR #7914, opened 2026-02-12
LibWeb: Stop storing GC::Roots in GC-allocated types

AtkinsSJ:
While building locally with `CLANG_PLUGINS` enabled, I got a megabyte of compiler output warning about this issue, which is these 6 offenders repeated over and over as various files include their headers:

```
/home/sam/ladybird/Libraries/LibWeb/Animations/AnimationPlaybackEvent.h:48:33: warning: Types inheriting from GC::Cell should not have GC::Root fields
   48 |     Optional<CSS::CSSNumberish> m_current_time;
      |                                 ^

/home/sam/ladybird/Libraries/LibWeb/Animations/AnimationPlaybackEvent.h:51:33: warning: Types inheriting from GC::Cell should not have GC::Root fields
   51 |     Optional<CSS::CSSNumberish> m_timeline_time;
      |                                 ^

/home/sam/ladybird/Libraries/LibWeb/Fetch/Infrastructure/HTTP/Bodies.h:80:16: warning: Types inheriting from GC::Cell should not have GC::Root fields
   80 |     SourceType m_source;
      |                ^
      
/home/sam/ladybird/Libraries/LibWeb/HTML/MessageEvent.h:55:34: warning: Types inheriting from GC::Cell should not have GC::Root fields
   55 |     Optional<MessageEventSource> m_source;

/home/sam/ladybird/Libraries/LibWeb/IndexedDB/Internal/Key.h:85:14: warning: Types inheriting from GC::Cell should not have GC::Root fields
   85 |     KeyValue m_value;
      |              ^

/home/sam/ladybird/Libraries/LibWeb/HTML/TrackEvent.h:38:31: warning: Types inheriting from GC::Cell should not have GC::Root fields
   38 |     TrackEventInit::TrackType m_track;
      |                               ^
```

They're all examples of fields being `Variant` types containing `GC::Root`s. I'm running Clang 21 locally as opposed to 20 on CI, which I suspect is why they don't show up there.

This PR fixes the above issues, by adding internal versions of those Variants which replace any `GC::Root`s with `GC::Ref`s. These are used for fields, and then we convert back and forth as needed. This is fiddly and annoying, but solves the issue until we find a way to stop needing to pass `GC::Root`s around.

### DIFF
--- a/Libraries/LibWeb/Fetch/Fetching/Fetching.cpp
+++ b/Libraries/LibWeb/Fetch/Fetching/Fetching.cpp
@@ -1458,7 +1458,7 @@ GC::Ptr<PendingResponse> http_redirect_fetch(JS::Realm& realm, Infrastructure::F
         // NOTE: BodyInitOrReadableBytes is a superset of Body::SourceType
         auto converted_source = source.has<ByteBuffer>()
             ? BodyInitOrReadableBytes { source.get<ByteBuffer>() }
-            : BodyInitOrReadableBytes { source.get<GC::Root<FileAPI::Blob>>() };
+            : BodyInitOrReadableBytes { source.get<GC::Ref<FileAPI::Blob>>() };
         auto [body, _] = safely_extract_body(realm, converted_source);
         request->set_body(body);
     }
@@ -1896,7 +1896,7 @@ GC::Ref<PendingResponse> http_network_or_cache_fetch(JS::Realm& realm, Infrastru
                 // NOTE: BodyInitOrReadableBytes is a superset of Body::SourceType
                 auto converted_source = source.has<ByteBuffer>()
                     ? BodyInitOrReadableBytes { source.get<ByteBuffer>() }
-                    : BodyInitOrReadableBytes { source.get<GC::Root<FileAPI::Blob>>() };
+                    : BodyInitOrReadableBytes { source.get<GC::Ref<FileAPI::Blob>>() };
                 auto [body, _] = safely_extract_body(realm, converted_source);
                 request->set_body(body);
             }
--- a/Libraries/LibWeb/Fetch/Infrastructure/HTTP/Bodies.cpp
+++ b/Libraries/LibWeb/Fetch/Infrastructure/HTTP/Bodies.cpp
@@ -25,12 +25,25 @@ GC_DEFINE_ALLOCATOR(Body);
 // - 1445 or more bytes have been read
 static constexpr size_t MAX_SNIFF_BYTES = 1445;
 
+static Body::SourceTypeInternal to_source_type_internal(Body::SourceType&& source_type)
+{
+    return source_type.visit(
+        [](Empty) -> Body::SourceTypeInternal { return Empty {}; },
+        [](ByteBuffer& buffer) -> Body::SourceTypeInternal { return move(buffer); },
+        [](GC::Root<FileAPI::Blob> const& blob) -> Body::SourceTypeInternal { return GC::Ref { *blob }; });
+}
+
 GC::Ref<Body> Body::create(JS::VM& vm, GC::Ref<Streams::ReadableStream> stream)
 {
     return vm.heap().allocate<Body>(stream);
 }
 
 GC::Ref<Body> Body::create(JS::VM& vm, GC::Ref<Streams::ReadableStream> stream, SourceType source, Optional<u64> length)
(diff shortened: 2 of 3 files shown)


Libraries/LibWeb/Animations/AnimationPlaybackEvent.h
```diff
@@ -29,26 +30,22 @@ class AnimationPlaybackEvent : public DOM::Event {
...
-        return m_current_time.map([](auto const& value) { return NullableCSSNumberish { value }; }).value_or(Empty {});
-    }
+    NullableCSSNumberish current_time() const;
+    NullableCSSNumberish timeline_time() const;
 
-    NullableCSSNumberish timeline_time() const
-    {
-        return m_timeline_time.map([](auto const& value) { return NullableCSSNumberish { value }; }).value_or(Empty {});
-    }
+    using CSSNumberishInternal = Variant<Empty, double, GC::Ref<CSS::CSSNumericValue>>;
```
trflynn89: It seems like this should be a `private` alias? I wouldn't expect such an internal type to be exposed

Libraries/LibWeb/HTML/TrackEvent.h
```diff
@@ -27,15 +28,19 @@ class TrackEvent : public DOM::Event {
     [[nodiscard]] static GC::Ref<TrackEvent> create(JS::Realm&, FlyString const& event_name, TrackEventInit = {});
     static WebIDL::ExceptionOr<GC::Ref<TrackEvent>> construct_impl(JS::Realm&, FlyString const& event_name, TrackEventInit);
 
+    using TrackTypeInternal = Variant<Empty, GC::Ref<VideoTrack>, GC::Ref<AudioTrack>, GC::Ref<TextTrack>>;
```
trflynn89: Same here, and so on

Libraries/LibWeb/IndexedDB/Internal/Key.h
```diff
@@ -51,8 +53,8 @@ class Key : public JS::Cell {
     [[nodiscard]] double value_as_double() { return m_value.get<double>(); }
     [[nodiscard]] AK::String value_as_string() { return m_value.get<AK::String>(); }
     [[nodiscard]] ByteBuffer value_as_byte_buffer() { return m_value.get<ByteBuffer>(); }
-    [[nodiscard]] Vector<GC::Root<Key>> value_as_vector() { return m_value.get<Vector<GC::Root<Key>>>(); }
-    [[nodiscard]] Vector<GC::Root<Key>> subkeys()
+    [[nodiscard]] GC::Root<GC::HeapVector<GC::Ref<Key>>> value_as_vector() { return m_value.get<GC::Root<GC::HeapVector<GC::Ref<Key>>>>(); }
```
trflynn89: I think `ReadonlySpan<GC::Ref<Key>>` would suffice? Looking at the callers, I don't think we need to create a root here.

That would be:
```c++
[[nodiscard]] ReadonlySpan<GC::Ref<Key>> value_as_vector() { return m_value.get<GC::Ref<GC::HeapVector<GC::Ref<Key>>>>()->elements(); }
```

That would also make several of the changed lines in Algorithms.cpp unnecessary

trflynn89 (commented):
Thinking out loud, I wonder if we should have a more generic `Variant<Root<T>>` -> `Variant<Ref<T>>` set of infrastructure. Feels like some template shenanigans could be done for that.  (Not that that's blocking this PR).

AtkinsSJ:
> I wonder if we should have a more generic `Variant<Root<T>>` -> `Variant<Ref<T>>` set of infrastructure.

When I spoke to Andreas about needing to convert so many of these, he was saying ideally we shouldn't need to use Roots for bindings interfaces at all. So I don't think it's probably worth figuring that out. (Maybe we have other places we need to swap between the two, but afaik bindings is the only one.)

Libraries/LibWeb/IndexedDB/Internal/Key.h
AtkinsSJ: Thanks, that's much nicer!

AtkinsSJ:
Changes:
- Made the `FooInternal` types private where possible. (`Body::SourceTypeInternal` being the one exception)
- Returned a `ReadonlySpan` from a couple of `Key` methods.

PR #7915, opened 2026-02-12
Meta: Support nullable BufferSource as Optional in IDL generator

jonbgamble:
This IDL change is needed for webaudio's WaveShaperNode, where a null BufferSource for a curve attribute results in a zero-length buffer.

WebGL also has a nullable BufferSource arg in bufferData(...). But there, a null data/srcData value returns GL_INVALID_VALUE.

### DIFF
--- a/Meta/Lagom/Tools/CodeGenerators/LibWeb/BindingsGenerator/IDLGenerators.cpp
+++ b/Meta/Lagom/Tools/CodeGenerators/LibWeb/BindingsGenerator/IDLGenerators.cpp
@@ -962,26 +962,53 @@ static void generate_to_cpp(SourceGenerator& generator, ParameterType& parameter
 )~~~");
         }
     } else if (is_javascript_builtin(parameter.type) || parameter.type->name() == "BufferSource"sv) {
-        if (optional) {
+        size_t buffer_source_nesting_level = optional ? 2 : 1;
+        auto buffer_source_indent = ByteString::repeated(' ', buffer_source_nesting_level * 4);
+        scoped_generator.set("buffer_source.indent", buffer_source_indent);
+
+        if (optional || parameter.type->is_nullable()) {
             scoped_generator.append(R"~~~(
     Optional<GC::Root<WebIDL::BufferSource>> @cpp_name@;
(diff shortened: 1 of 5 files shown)


Meta/Lagom/Tools/CodeGenerators/LibWeb/BindingsGenerator/IDLGenerators.cpp
```diff
@@ -874,27 +874,53 @@ static void generate_to_cpp(SourceGenerator& generator, ParameterType& parameter
...
+            }
+        } else if (optional) {
+            scoped_generator.append(R"~~~(
+    Optional<GC::Root<WebIDL::BufferSource>> @cpp_name@;
+    if (!@js_name@@js_suffix@.is_undefined()) {
+        if (!@js_name@@js_suffix@.is_object() || !(is<JS::TypedArrayBase>(@js_name@@js_suffix@.as_object()) || is<JS::ArrayBuffer>(@js_name@@js_suffix@.as_object()) || is<JS::DataView>(@js_name@@js_suffix@.as_object())))
+            return vm.throw_completion<JS::TypeError>(JS::ErrorType::NotAnObjectOfType, "@parameter.type.name@");
+
+        @cpp_name@ = GC::make_root(realm.create<WebIDL::BufferSource>(@js_name@@js_suffix@.as_object()));
+    }
```
gmta: You've basically unrolled the old logic into code blocks; the diff could look much simpler if you expanded on the original approach and emit the right checks and returns depending on `optional` and `parameter.type->is_nullable()`.

jonbgamble: It's been so long since I did this, but I seem to recall arriving at something less than optimal here due to difficulty indenting the generated files in a way clang plugins indentation checker will accept. I'll take another look though.

jonbgamble: Ok, I was able to roll it back up. I did have to add a bit of indentation shenanigans due to clang-plugins aversion to consecutive unbracketed `if` statements at different indent levels.

PR #7921, opened 2026-02-12
LibGC+LibJS+Meta: Keep track of source locations of edges in dumps

Zaggy1024:
This helps to determine where an object is being visited from.


Zaggy1024:
Obligatory screenshot:

image

PR #7923, opened 2026-02-13
LibIPC: Notify readers when thread exits

jonbgamble:


Zaggy1024:
What's the purpose of this change? Were we hanging waiting for a socket write somewhere?

jonbgamble:
Yes, but not in master. Here they're just defensive. It's one of several issues (like the MessagePort one) that came up in IPC flows with helper processes, mostly during teardown & rapid navigation (test-web) with a webaudio worker.

Shall we just close them for now?

Zaggy1024:
I'm mainly wondering if it's something we can test somehow, that would be ideal. If we can reproduce an issue that this fixes, even by repeating the test many times, that would prove the change worthwhile. I'm just not clear on what simply notifying here is supposed to do, I would've figured the socket should get closed after the thread exits.

jonbgamble:
If a consumer is waiting in poll via read hook (rather than wait_until_readable), and the transport I/O thread exits, then the read hook may never be called. That means it never gets a chance to call read_as_many_messages_as_possible_without_blocking and observe ShouldSHutdown::Yes from m_peer_eof.

For testing, see the unit test. If you `gh pr checkout` this branch, comment line 138 to reproduce the fail on thread exit. Then uncomment that line to witness the timeless glory of **LibIPC: Notify readers when thread exits**.

PR #7926, opened 2026-02-13
LibMedia+AudioServer: Add audio device server process

jonbgamble:
If running the WebContent process as a restricted user and reducing its complexity is a goal, then consider AudioServer as a step in that direction.

In this PR, each WebContent delivers audio streams to a global AudioServer process through shared memory ring buffers. LibMedia still uses AudioMixingSink to mix media element tracks in-process. But it no longer interacts with the OS device layer. Instead, WebContent (and potentially other clients) open an output session to the AudioServer singleton and ask for an output stream.

AudioServer maintains a single device callback thread (per output device) and mixes incoming streams for that device before passing to that device. It provides a cross platform enumeration surface for both input and output, default and non-default, and provides playback to any output as well as microphone (and other) input streams.

See [AudioMixingSinkTest.cpp](https://github.com/jonbgamble/ladybird/blob/pr-audio-server/Tests/LibMedia/TestAudioMixingSink.cpp) and [MinimalAudioServerClientExample.cpp](https://github.com/jonbgamble/ladybird/blob/pr-audio-server/Utilities/Examples/MinimalAudioServerClientExample.cpp) for a usage example.


trflynn89:
I'm just going to leave some general project-level comments here:

1. New types in AK without unit tests is a non-starter.
2. What is going on in AudioServer/main.cpp with all that ANSI code handling?
3. Related to 2, we use debug flags already for debug output with `dbgln_if`. Let's not invent new mechanisms here with environment variables.
4. Always initialize plain types. For example, `SharedBufferStream::PoolHeader` and `RingStreamHeader` - those should all be zero-initialized in the struct definition.
5. Don't define IPC definitions in headers. We actively put only the declarations in the headers (with an `LibIPC/Forward.h` inclusion), and definitions in the .cpp (with `LibIPC/Decoder.h` and `LibIPC/Encoder.h` inclusions). This avoids a lot of recompilation when something in Encoder.h/Decoder.h, or in something they include themselves, changes.
6. Related to 5, we put IPC definitions in the same file as the types they are transferring. Not in their own IPC.h/.cpp files.
7. There's a whole lot here that seems to be inlined that doesn't need to be.
8. Take a look through AK for things that already exist / look at how other classes implement certain behaviors. For example, `is_power_of_two` is already a function in AK, `SharedSingleProducerCircularBuffer` doesn't need its own. Prefer `AK::TypedTransfer` over manual memcpy.
9. A single commit with +5k lines is very difficult to review.

jonbgamble:
> 2. What is going on in AudioServer/main.cpp with all that ANSI code handling?

It's debug junk for test-web attaches in a different branch.

> 3. Related to 2, we use debug flags already for debug output with dbgln_if. Let's not invent new mechanisms here with environment variables.

Does Ladybird have a mechansim to enable logging without a recompile? Point me towards it and I'll use it.

> 9. A single commit with +5k lines is very difficult to review.

I agree, but sometimes big changes are big and breaking them up into tiny chunks may not aid comprehension much. While this is not a maintainer's concern, it can be time consuming to identify (and sometimes, actively manufacture) the proper commit boundaries for a large interconnected feature to make history look nice. I'm confident we can arrive at some middle ground that accommodates both sides.

Zaggy1024:
A few surface-level critiques on this before I try to get a more in-depth review going: 
- Instead of creating copies of the PlaybackStream, could we at least initially implement the AudioServer in terms of LibMedia as it stands now? That would make things a lot easier to review, and it would also make it much clearer to me if there were changes needed to make PlaybackStream or its implementations work in this architecture.
- LibAudioServer as a name sounds like it's internal implementation details for the AudioServer. If it's intended to establish and maintain client connections, should it be called LibAudioClient or something instead?
- An overview of the architecture and how everything fits together would be very helpful, especially regarding the interfaces and how they're used by the clients. I see some .ipc files that are named very differently from what I'm accustomed to with our other service processes. It might be good to try to model this setup after something like LibRequests/RequestServer, it's hard to tell if it's built similarly or not. Also, please include information about how audio data requests are driven and how that fits into audio timing.
- Commit order seems a little odd to me. Shouldn't the general IPC changes come before the introduction of the audio service?

jonbgamble:
The main difficulty with building this on top of the LibMedia in master is it assumes a single output path. This PR tries to add per-client sessions, device enumeration, non-default output selection, and microphone/input support.

I’m not attached to the library naming, but I'll tell you why I chose LibAudioServer. Both WebContent and AudioServer link it. It defines all the shared structures and IPC rather than splitting them across Libraries and Services. The IPC files are named ToRecipientFromSender.ipc. That approach is one I actively chose, and I'll admit there are both pros and cons.

I may be looking at it differently, but I think the current commit order is more natural because AudioServer exists independently. It can be exercised through the Utilities examples without ever integrating with Ladybird. The later WebContent/browser IPC commits integrate it with Ladybird directly.

Let me work up an architecture overview. That's a logical next step that can help you decide whether to move forward with this.

Zaggy1024:
> The main difficulty with building this on top of the LibMedia in master is it assumes a single output path. This PR tries to add per-client sessions, device enumeration, non-default output selection, and microphone/input support.

I suppose the difficulty here is that we would have a dependency loop if we tried to use the PlaybackStream in the audio service? Not sure of the best way to handle this, but I'm really not a fan of copying all the implementations and then deleting the originals subsequently. It breaks file history. Perhaps one workable solution would be to move them into a separate library first, then use them in AudioServer.

> I’m not attached to the library naming, but I'll tell you why I chose LibAudioServer. Both WebContent and AudioServer link it. It defines all the shared structures and IPC rather than splitting them across Libraries and Services. The IPC files are named ToRecipientFromSender.ipc. That approach is one I actively chose, and I'll admit there are both pros and cons.

I think consistency is important in regard to this, but if there's a good argument for a different structure and naming convention, we could reconsider later. It makes it easier to review/understand the code if we remain consistent with existing code.

> I may be looking at it differently, but I think the current commit order is more natural because AudioServer exists independently. It can be exercised through the Utilities examples without ever integrating with Ladybird. The later WebContent/browser IPC commits integrate it with Ladybird directly.

From what I can tell with a cursory look at the commits, we have more logical steps here:
- Add AudioServer
- Improve the infrastructure for spawning service processes
- Include AudioServer in the above infrastructure
- Hook LibMedia up to use AudioServer

I'd rather have the improvements to spawning services be separate, then keep the commit adding the AudioServer spawning very simple. Also, I'm not sure how I feel about LibMedia depending on the using code to spawn the AudioServer prior to audio output. Not sure how viable it would be to make it spawn the service lazily and share it between all WebContent processes. (Though I don't know that we necessarily need to share it? We don't get too much benefit from that, I think.)

jonbgamble:
>  I'm not sure how I feel about LibMedia depending on the using code to spawn the AudioServer prior to audio output.

Linking the same library from the client and server helped localize shared concerns so I will defend that decision (to a point). Giving clients/LibMedia access to more symbols than it needs is not something I worry about. It's not a public library. I'm open to shuffling non-shared things out of it. I imagine the boundaries are fuzzy.

I'm not sure what you're referring to about improvements to spawning services being separate. Maybe it's [this](https://github.com/LadybirdBrowser/ladybird/pull/7925)? I broke that out from WebAduio/AudioServer with the intention of merging it, but it wasn't getting interest and eventually I felt it didn't offer enough to stand on its own. The [MediaCapture wiring](https://github.com/LadybirdBrowser/ladybird/pull/7909) is also separate. I haven't closed that one yet.

I'm not sure how to keep file history. I didn't delete much if any of the old playback manager code, it's just layered different. But even if we opened up the diff threshold as wide as it will go, I'm not sure git would see the new TUs as a rename.

Maybe we could leave the old sources dangling in LibMedia and I could link to them in comments for history?

jonbgamble:
By the way, my comments are not an attempt to put the ball in your court. I am going to revisit this branch with your concerns in mind soon. I have two long flights coming up this week, and both are multi-leg with long layovers. So that's a perfect opportunity for some documentation, code adjustments, and a little git fuckery.

jonbgamble:
I added some light docs under Documentation. Feel free to move or delete them.

Please read them and build/test the branch before requesting renames or structural changes. Keeping this branch in sync with upstream is already costly. Change requests over time will make this very hard to land.

If it’s easier, I’d prefer to hand this off. You can rename/move everything as you see fit (including moving it back to LibMedia) and commit it directly. Seems like you might prefer this as well - given your strong sense of ownership and direction for all things media.

PR #7929, opened 2026-02-13
LibWeb: Clear textarea with empty string for WebDriver

Praise-Garfield:
The WebDriver clear handler for textarea elements set the raw value to `child_text_content()` instead of an empty string. This was a copy-paste error from the adjacent reset handler, which correctly uses `child_text_content()` per its own spec (reset restores the textarea to its initial DOM content). The clear handler should empty the textarea entirely.

Also adds an `internals.clearElement()` helper to expose the WebDriver clear algorithm to the test harness, with a regression test covering the fix.

Spec:
https://w3c.github.io/webdriver/#dfn-clear-algorithm

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLTextAreaElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLTextAreaElement.cpp
@@ -127,7 +127,7 @@ void HTMLTextAreaElement::clear_algorithm()
     m_dirty_value = false;
 
     // and set the raw value of element to an empty string.
-    set_raw_value(child_text_content());
+    set_raw_value({});
 
     // Unlike their associated reset algorithms, changes made to form controls as part of these algorithms do count as
     // changes caused by the user (and thus, e.g. do cause input events to fire).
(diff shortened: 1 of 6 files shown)


Libraries/LibWeb/HTML/HTMLTextAreaElement.cpp
```diff
@@ -127,7 +127,7 @@ void HTMLTextAreaElement::clear_algorithm()
     m_dirty_value = false;
 
     // and set the raw value of element to an empty string.
-    set_raw_value(child_text_content());
+    set_raw_value({});
```
gmta, line 130: Please add a regression test for this.

Libraries/LibWeb/HTML/HTMLTextAreaElement.cpp
gmta, line 1: Commit title suggestion:

`LibWeb: Clear textarea with empty string for WebDriver`

Words like 'fix', 'refactor' etc. don't really add a lot. :)

gmta: You updated the PR title, but forgot the commit title

Praise-Garfield: thought I could get away with it/it wasn't implied; lesson learned.

Thanks.

PR #7935, opened 2026-02-13
LibWeb: Check for NavigationParams in finalize_session_history_entry

Praise-Garfield:
Step 7.3 of the spec says "if navigationParams is a navigation params," but the condition checked for `NonFetchSchemeNavigationParams` instead of `NavigationParams`. The block body already accessed the variant as `NavigationParams`, so if the condition ever matched it would VERIFY-crash. The identical check five lines above in step 7.2.3.1 was already correct, this was not.

Fixing the variant type also means this block now runs for regular navigations. Srcdoc navigations legitimately have a null `request` in their `NavigationParams` (per `create_navigation_params_from_a_srcdoc_resource`), so the condition also guards on that to avoid hitting the spec's "request is not null" assertion.

Spec:
https://html.spec.whatwg.org/multipage/browsing-the-web.html#finalize-a-session-history-entry

### DIFF
--- a/Libraries/LibWeb/HTML/Navigable.cpp
+++ b/Libraries/LibWeb/HTML/Navigable.cpp
@@ -1363,9 +1363,10 @@ static void finalize_session_history_entry(
 
         // 3. If entry's document state's request referrer is "client", and navigationParams is a navigation params (i.e., neither null nor a non-fetch scheme navigation params), then:
         if (entry->document_state()->request_referrer() == Fetch::Infrastructure::Request::Referrer::Client
-            && (!received_navigation_params.has<Navigable::NullOrError>() && received_navigation_params.has<GC::Ref<NonFetchSchemeNavigationParams>>())) {
+            && received_navigation_params.has<GC::Ref<NavigationParams>>()
+            && received_navigation_params.get<GC::Ref<NavigationParams>>()->request) {
             // 1. Assert: navigationParams's request is not null.
-            VERIFY(received_navigation_params.has<GC::Ref<NavigationParams>>() && received_navigation_params.get<GC::Ref<NavigationParams>>()->request);
+            // NB: We don't perform this assertion because srcdoc navigations create NavigationParams with a null request.
 
             // 2. Set entry's document state's request referrer to navigationParams's request's referrer.
             entry->document_state()->set_request_referrer(received_navigation_params.get<GC::Ref<NavigationParams>>()->request->referrer());


Libraries/LibWeb/HTML/Navigable.cpp
```diff
@@ -1363,11 +1363,9 @@ static void finalize_session_history_entry(
...
         // 3. If entry's document state's request referrer is "client", and navigationParams is a navigation params (i.e., neither null nor a non-fetch scheme navigation params), then:
         if (entry->document_state()->request_referrer() == Fetch::Infrastructure::Request::Referrer::Client
-            && (!received_navigation_params.has<Navigable::NullOrError>() && received_navigation_params.has<GC::Ref<NonFetchSchemeNavigationParams>>())) {
-            // 1. Assert: navigationParams's request is not null.
-            VERIFY(received_navigation_params.has<GC::Ref<NavigationParams>>() && received_navigation_params.get<GC::Ref<NavigationParams>>()->request);
-
-            // 2. Set entry's document state's request referrer to navigationParams's request's referrer.
+            && received_navigation_params.has<GC::Ref<NavigationParams>>()
+            && received_navigation_params.get<GC::Ref<NavigationParams>>()->request) {
+            // 1. Set entry's document state's request referrer to navigationParams's request's referrer.
```
trflynn89: Again, please leave the steps copied exactly as they are in the spec, verbatim. We can then leave a comment explaining why we differ from the spec.

Something like:
```c++
// 1. Assert: navigationParams's request is not null.
// NB: We don't perform this assertion because ...
```

Libraries/LibWeb/HTML/Navigable.cpp
```diff
@@ -1363,11 +1363,9 @@ static void finalize_session_history_entry(
...
             entry->document_state()->set_request_referrer(received_navigation_params.get<GC::Ref<NavigationParams>>()->request->referrer());
```
trflynn89, line 1372: Is there a test we can add to exercise this change?

Praise-Garfield: Since the referrer stored in the document state isn't directly observable via JS, I'm not too sure how I would implement one. Open to suggestions before I push, if you have ideas.

Libraries/LibWeb/HTML/Navigable.cpp
Praise-Garfield: Understood.

Libraries/LibWeb/HTML/Navigable.cpp
trflynn89: Ok, if it's difficult to test, that's fine. FYI it helps to describe that in the commit message - i.e. a short "I haven't added a test because ..." blurb.

Praise-Garfield: Noted -- thanks.

PR #7937, opened 2026-02-13
LibWeb/Crypto: Update spec links

devgianlu:
Add and update spec links to more precise ones and move X25519 and Ed25519 from WICG to W3C.

X448 and Ed448 cannot be updated since they haven't been moved to W3C yet: https://github.com/w3c/webcrypto/pull/425


devgianlu:
CI flaked again :shrug:

PR #7948, opened 2026-02-14
LibDatabase: Allow String values to contain embedded null bytes

shannonbooth:
LibDatabase previously truncated String values at the first '\0' when
round-tripping through SQLite due to the use of strlen.

This something which we should support according to WPT:

https://wpt.live/webstorage/storage_setitem.window.html

The above test works fine when run through our test-web harness
as we have the SQL database disabled, so this instead adds a C++
test.


Libraries/LibDatabase/Database.cpp
```diff
@@ -54,11 +54,20 @@ static constexpr StringView sql_error(int error_code)
...
+}
+
 ErrorOr<NonnullRefPtr<Database>> Database::create(ByteString const& directory, StringView name)
 {
     TRY(Core::Directory::create(directory, Core::Directory::CreateDirectories::Yes));
     LexicalPath database_path { ByteString::formatted("{}/{}.db", directory, name) };
+    return create(move(database_path));
+}
 
+ErrorOr<NonnullRefPtr<Database>> Database::create(LexicalPath database_path)
```
trflynn89: Feels like a bit of a weird API difference that the API that accepts a directory will create that directory if it doesn't exist, but the API that accepts a file will not create the parent directory.

Intuitively, I think it makes more sense for the stored path to be `Optional<LexicalPath>`, which is null for in-memory DBs, because getting a path of ":memory:" doesn't seem right. You could have a (private) `create(sqlite3*)` helper to facilitate that, and just make both the file and memory factories call `sqlite3_open`.

shannonbooth: Yeah, I had iterated on previous versions and after figuring out just using a memory backed DB would be best the API ended up being not so nice! Adjusted now to something with a bit better shape

PR #7949, opened 2026-02-14
LibHTTP: Handle InvalidURL in parse_error_to_string

Praise-Garfield:
A malformed URL in a WebDriver request caused a crash via `VERIFY_NOT_REACHED()` in `parse_error_to_string`, because the `ParseError::InvalidURL` variant was not handled. The variant can be returned by `from_raw_request()` and is passed to `parse_error_to_string` in `LibWeb/WebDriver/Client.cpp`.

This adds the missing case so the error is logged instead of crashing.

### DIFF
--- a/Libraries/LibHTTP/HttpRequest.h
+++ b/Libraries/LibHTTP/HttpRequest.h
@@ -37,9 +37,10 @@ public:
             return "Out of memory"sv;
         case ParseError::UnsupportedMethod:
             return "Unsupported method"sv;
-        default:
-            VERIFY_NOT_REACHED();
+        case ParseError::InvalidURL:
+            return "Invalid URL"sv;
         }
+        VERIFY_NOT_REACHED();
     }
 
     enum Method {


Libraries/LibHTTP/HttpRequest.h
```diff
@@ -37,6 +37,8 @@ class HttpRequest {
             return "Out of memory"sv;
         case ParseError::UnsupportedMethod:
             return "Unsupported method"sv;
+        case ParseError::InvalidURL:
+            return "Invalid URL"sv;
         default:
```
shannonbooth: Nit: But could we remove this default case and instead put the VERIFY_NOT_REACHED outside of the switch? This would have caught this enum being missed when introduced as a compile error

Praise-Garfield: Yes.

shannonbooth (approved):

PR #7954, opened 2026-02-14
LibWeb/CSS: Don't assume that we consumed a color space token

Lubrsi:
Fixes https://cockpit-project.org/ crashing on load.


Tests/LibWeb/Text/input/css/color-mix-with-no-color-space.html
```diff
@@ -0,0 +1,12 @@
...
+    body {
+        background: color-mix(in calc(1), red, blue);
+    }
+</style>
+<script src="../include.js"></script>
+<script>
+    test(() => {
+        println("PASS! (Didn't crash)");
+    });
+</script>
```
tcl3: Can this be a crash test instead?

AtkinsSJ (approved):

PR #7956, opened 2026-02-14
LibWebView+UI/AppKit: Present IOSurfaces directly

Lubrsi:
Instead of copying the Bitmap that wraps the IOSurface, we can just present the IOSurface directly. This significantly reduces CPU usage in the UI process, particularly at high refresh rates such as 120Hz where it would saturate a full CPU core.

This is done by using CAMetalLayer and blitting the IOSurface to the next drawable buffer, which handles triple buffering, locking the IOSurface and vsync automatically. This also allows the Metal HUD to work, but the only accurate stat is the frame intervals/FPS because it's in the UI layer, not WebContent. However, that's still useful to detect frame drops.

https://github.com/user-attachments/assets/e61e703b-e28f-4643-a2ea-f2cba69d3856

### DIFF
--- a/Libraries/LibWebView/ViewImplementation.cpp
+++ b/Libraries/LibWebView/ViewImplementation.cpp
@@ -582,13 +582,18 @@ void ViewImplementation::did_allocate_iosurface_backing_stores(i32 front_id, Cor
 
     auto bytes_per_row = front_iosurface.bytes_per_row();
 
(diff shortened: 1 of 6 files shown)


kalenikaliaksandr:
it seems to solve the flickering issues I've been seeing on macOS for a long time. extremely cool!

PR #7957, opened 2026-02-14
LibWeb+Tests: Implement scroll position restoration for session history

Praise-Garfield:
Pressing back or forward in Ladybird did not restore the viewport's scroll position, unlike every other major browser.

This implements the spec's "scroll position data" field on session history entries as an `Optional<CSSPixelPoint>`, with persist and restore algorithms on `Navigable`. Scroll position is saved before switching the active entry in `activate_history_entry`, `navigate_to_a_fragment`, and `perform_url_and_history_update_steps`. Restoration is handled in both `Document::update_for_history_step_application` (the common non-intercepted path) and `NavigateEvent::process_scroll_behavior` (the intercepted path), respecting `history.scrollRestoration = 'manual'`.

Only the viewport scroll offset is persisted for now (not per-element overflow scroll positions).

Includes a test that verifies the full `pushState`-then-back cycle restores scroll position.

Spec's:
https://html.spec.whatwg.org/multipage/browsing-the-web.html#she-scroll-position
https://html.spec.whatwg.org/multipage/browsing-the-web.html#save-persisted-state
https://html.spec.whatwg.org/multipage/browsing-the-web.html#restore-scroll-position-data

### DIFF
--- a/Tests/LibWeb/Text/data/scroll-position-restoration.html
+++ b/Tests/LibWeb/Text/data/scroll-position-restoration.html
@@ -1,0 +1,19 @@
+<!DOCTYPE html>
+<style>div { height: 10000px; }</style>
+<div></div>
+<script>
+    window.scrollTo(0, 200);
+    history.pushState({}, "", location.href + "#pushed");
+    window.scrollTo(0, 0);
+    parent.postMessage(`scrollY before back: ${window.scrollY}`, "*");
+
+    window.addEventListener("popstate", () => {
+        // Scroll is restored at spec step 4, after popstate dispatch returns.
+        // Listen for the scroll event that fires when the position is restored.
+        document.addEventListener("scroll", () => {
+            parent.postMessage(`scrollY after back: ${window.scrollY}`, "*");
+        }, { once: true });
+    });
+
+    history.back();
+</script>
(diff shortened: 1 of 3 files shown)


Libraries/LibWeb/HTML/Navigable.cpp
```diff
@@ -2526,6 +2535,40 @@ void perform_url_and_history_update_steps(DOM::Document& document, URL::URL new_
     }));
 }
 
+// https://html.spec.whatwg.org/multipage/browsing-the-web.html#save-persisted-state
+void Navigable::persist_scroll_position_data(SessionHistoryEntry& entry)
```
Psychpsyo: This should probably be called `save_persisted_state`. I don't see a reason to deviate from the spec name.

Psychpsyo:
All the `// NB:` comments about us deviating from the spec should probably be `// AD-HOC:`s instead, maybe with some reasoning as to why they are necessary.
Also, it might be worth mentioning the bit about having to do this after scrolling to a fragment here: https://github.com/whatwg/html/issues/12138
Scrolling to a fragment seems to have shaky interop in other engines and if this is a spec bug, it'd be good if they were aware of that when trying to sort that out.

PR #7967, opened 2026-02-15
LibWeb: Treat numeric PropertyKeys as Strings per spec in PlatformObject

shannonbooth:
PropertyKey values can only be String or Symbol per spec; numeric keys
are an internal optimization and should be treated as Strings.

This fixes incorrect behavior for numeric property names on legacy
platform objects (e.g. Web Storage).


shannonbooth:
Actually I should have mentioned that this is _closer_ to how other implementations behave at least

shannonbooth:
Alright, completely revised now with a much better fix

PR #7968, opened 2026-02-15
LibHTTP: Use a memory-backed database for the disk cache in test modes

trflynn89:
This just lets us create fewer cache directories during WPT. We do still create cache entries on disk, so for WPT, we introduce an extra cache key to prevent conflicts. There is an existing FIXME about this.

### DIFF
--- a/Libraries/LibHTTP/Cache/CacheIndex.cpp
+++ b/Libraries/LibHTTP/Cache/CacheIndex.cpp
@@ -49,7 +49,7 @@ static NonnullRefPtr<HeaderList> deserialize_headers(StringView serialized_heade
     return headers;
 }
 
-ErrorOr<CacheIndex> CacheIndex::create(Database::Database& database)
+ErrorOr<CacheIndex> CacheIndex::create(Database::Database& database, LexicalPath const& cache_directory)
 {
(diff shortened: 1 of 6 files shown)


shannonbooth:
nice! an alternative FWIW might be to use the 'temporary' database thingy of libsqlite, though I am not sure how much of a difference it makes here. (I was thinking about whether it made any sense to use this type of feature instead of the disable SQL database stuff just to reduce code paths)

trflynn89:
Ha yeah that was actually part of my plan, but we're not quite there yet for local storage. It was disabled for tests in commit 642dd751cfb93e8fe8b6d9a2071058a158e882b2 (which affects just cookies and local storage). The test mentioned there is still painfully slow with sqlite enabled.

shannonbooth:
Right, though at that time we didn't have the WAL config changes and that type of thing made to improve performance. And _maybe_ performance is comparable also when using temporary file / in memory database. Not sure, I only got to the point of thinking about measuring the difference if it made sense or not

trflynn89:
What I mean was I already coded up using an in-memory DB for local storage during tests. The test mentioned in that commit still takes 20-30s on my desktop when run in an entire test-web job.

(FYI, WAL has no effect on in-memory DBs)

PR #7969, opened 2026-02-15
LibWeb: Use generic layout for input elements

### LINKED ISSUE #7900: Input fields with `appearance: none` are not sized like in other browsers
### Summary

Something about the sizing of input elements with CSS `appearance: none` is different between us and other browsers.

This makes it so that the search bar in the top right of https://pinball.com/ is too small.

### Operating system

Linux

### Steps to reproduce

1. Nest two flexboxes and put an HTML `input` element inside.
2. Give the input `appearance: none` and `width: 100%`.
3. It becomes way too slim.

### Expected behavior

The input element should be its usual width.

### Actual behavior

The input element is as slim as it can be.

https://pinball.com/

```html
<!DOCTYPE html>
<style>
  input {
    appearance: none;
    width: 100%;
  }
  div {
    display: flex;
  }
</style>
<div><div><input></div></div>
```

```shell
none
```

Top: Webkit (Chromium and Firefox are basically the same)
Bottom: Ladybird
Image

none

### Contribute a patch?

ChaseKnowlden:
Makes the search bar on pinball.com wider
Closes #7900

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLInputElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLInputElement.cpp
@@ -130,23 +130,27 @@ GC::Ptr<Layout::Node> HTMLInputElement::create_layout_node(GC::Ref<CSS::Computed
         return heap().allocate<Layout::ImageBox>(document(), *this, move(style), *this);
     }
 
-    // https://drafts.csswg.org/css-ui/#appearance-switching
-    // This specification introduces the appearance property to provide some control over this behavior.
-    // In particular, using appearance: none allows authors to suppress the native appearance of widgets,
-    // giving them a primitive appearance where CSS can be used to restyle them.
-    if (style->appearance() == CSS::Appearance::None) {
-        return Element::create_layout_node_for_display_type(document(), style->display(), style, this);
-    }
-
     switch (type_state()) {
 
     case TypeAttributeState::SubmitButton:
     case TypeAttributeState::Button:
     case TypeAttributeState::ResetButton:
+        // NB: For button inputs with appearance: none, use generic layout
+        // https://drafts.csswg.org/css-ui/#appearance-switching
+        if (style->appearance() == CSS::Appearance::None)
+            return Element::create_layout_node_for_display_type(document(), style->display(), style, this);
         return heap().allocate<Layout::BlockContainer>(document(), this, move(style));
     case TypeAttributeState::Checkbox:
+        // NB: For checkbox inputs with appearance: none, use generic layout
+        // https://drafts.csswg.org/css-ui/#appearance-switching
+        if (style->appearance() == CSS::Appearance::None)
+            return Element::create_layout_node_for_display_type(document(), style->display(), style, this);
         return heap().allocate<Layout::CheckBox>(document(), *this, move(style));
     case TypeAttributeState::RadioButton:
+        // NB: For radio inputs with appearance: none, use generic layout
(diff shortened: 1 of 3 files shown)


Libraries/LibWeb/HTML/HTMLInputElement.cpp
```diff
@@ -155,6 +159,9 @@ GC::Ptr<Layout::Node> HTMLInputElement::create_layout_node(GC::Ref<CSS::Computed
     case TypeAttributeState::Email:
     case TypeAttributeState::Password:
     case TypeAttributeState::Number:
+        // NOTE: Text-like inputs still use TextInputBox even with appearance: none,
+        //       as they need proper intrinsic sizing for flexbox and other layout contexts.
+        //       The appearance property only affects rendering, not sizing behavior.
```
Psychpsyo: Sizing and rendering are no different. The point of the primitive appearance is that the element's entire visual presentation is determined only by the normal rules of CSS and nothing else. The [spec you linked](https://drafts.csswg.org/css-ui/#appearance-switching) even calls this out by saying that "primitive appearance [...] can cause some widgets to disappear completely".
Maybe this is a spec bug?

Libraries/LibWeb/HTML/HTMLInputElement.cpp
```diff
@@ -130,23 +130,27 @@ GC::Ptr<Layout::Node> HTMLInputElement::create_layout_node(GC::Ref<CSS::Computed
...
-        return Element::create_layout_node_for_display_type(document(), style->display(), style, this);
-    }
-
     switch (type_state()) {
 
     case TypeAttributeState::SubmitButton:
     case TypeAttributeState::Button:
     case TypeAttributeState::ResetButton:
+        // https://drafts.csswg.org/css-ui/#appearance-switching
+        // For button inputs with appearance: none, use generic layout
```
Psychpsyo: This looks like the comment is citing the spec when it isn't.
It should probably either actually quote the spec or have an `NB:` at the start.
Same goes for the other comments below.

Psychpsyo:
[The spec](https://drafts.csswg.org/css-ui/#appearance-decorative) also says:
> UAs should include in their user agent stylesheet style rules to give [widgets](https://drafts.csswg.org/css-ui/#widget) a recognizable shape when ['appearance'](https://drafts.csswg.org/css-ui/#propdef-appearance) is ['none'](https://drafts.csswg.org/css-ui/#valdef-appearance-none).

This is likely what's meant to be giving a box with `appearance: none` its shape.

Psychpsyo:
> This is likely what's meant to be giving a box with `appearance: none` its shape.

Looking into this more, this seems to currently be under-specced. The size in other browsers seems to come from the fact that some form controls should keep some of their native presentation even when they have a primitive appearance. (`appearance: none`)
The HTML spec currently says "Need to detail the [primitive appearance](https://drafts.csswg.org/css-ui/#primitive-appearance)." for a lot of the form controls.

Libraries/LibWeb/HTML/HTMLInputElement.cpp
ChaseKnowlden: Added comments.

Libraries/LibWeb/HTML/HTMLInputElement.cpp
ChaseKnowlden: Fixed

PR #7970, opened 2026-02-15
LibWeb: Block opaque origins in CSP frame-ancestors check

tcl3:
This matches the behavior of other engines.

Fixes 2 WPT test crashes. These tests don't work reliably when imported, so I made a simpler internal test.


Libraries/LibWeb/ContentSecurityPolicy/Directives/FrameAncestorsDirective.cpp
```diff
@@ -57,8 +57,8 @@ Directive::Result FrameAncestorsDirective::navigation_response_check(GC::Ref<Fet
         // 2. Let origin be the result of executing the URL parser on the ASCII serialization of document’s origin.
         auto origin = DOMURL::parse(document->origin().serialize());
 
-        // FIXME: What do we do if origin is invalid here?
-        VERIFY(origin.has_value());
+        if (!origin.has_value())
```
shannonbooth, line 62: I think we should probably still leave an AD-HOC comment or something so this sticks out

tcl3: I've added a comment explaining the rationale for this.

PR #7972, opened 2026-02-15
WebXR 4: Implement enough of WebXR for xrSession_end test to pass

Psychpsyo:
This makes this specific WPT test pass. For any further ones, we'll need to make WebGL work in the test runner.

### DIFF
--- a/Libraries/LibWeb/WebXR/XRRenderState.idl
+++ b/Libraries/LibWeb/WebXR/XRRenderState.idl
@@ -1,6 +1,6 @@
 // https://immersive-web.github.io/webxr/#xrrenderstate-interface
 
-dictionary XRRenderStateInit {
+[Experimental] dictionary XRRenderStateInit {
     double depthNear;
     double depthFar;
     boolean passthroughFullyObscured;
@@ -9,7 +9,7 @@ dictionary XRRenderStateInit {
     // FIXME: sequence<XRLayer>? layers;
 };
 
-[SecureContext, Exposed=Window] interface XRRenderState {
+[Experimental, SecureContext, Exposed=Window] interface XRRenderState {
     [FIXME] readonly attribute double depthNear;
     [FIXME] readonly attribute double depthFar;
     [FIXME] readonly attribute boolean? passthroughFullyObscured;
--- a/Libraries/LibWeb/WebXR/XRWebGLLayer.idl
+++ b/Libraries/LibWeb/WebXR/XRWebGLLayer.idl
@@ -5,7 +5,7 @@
 typedef (WebGLRenderingContext or
          WebGL2RenderingContext) XRWebGLRenderingContext;
 
-dictionary XRWebGLLayerInit {
+[Experimental] dictionary XRWebGLLayerInit {
     boolean antialias = true;
     boolean depth = true;
     boolean stencil = false;
@@ -14,7 +14,7 @@ dictionary XRWebGLLayerInit {
     double framebufferScaleFactor = 1.0;
 };
 
-[SecureContext, Exposed=Window]
+[Experimental, SecureContext, Exposed=Window]
 interface XRWebGLLayer: XRLayer {
     constructor(XRSession session,
                XRWebGLRenderingContext context,
--- a/Libraries/LibWeb/WebXR/XRLayer.idl
+++ b/Libraries/LibWeb/WebXR/XRLayer.idl
@@ -1,3 +1,3 @@
 // https://immersive-web.github.io/webxr/#xrlayer-interface
-[SecureContext, Exposed=Window]
+[Experimental, SecureContext, Exposed=Window]
 interface XRLayer : EventTarget {};
--- a/Libraries/LibWeb/WebXR/XRSession.idl
+++ b/Libraries/LibWeb/WebXR/XRSession.idl
@@ -1,13 +1,13 @@
 #import <DOM/EventTarget.idl>
 #import <WebXR/XRRenderState.idl>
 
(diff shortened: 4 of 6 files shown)


Libraries/LibWeb/WebGL/WebGLRenderingContextBase.cpp
```diff
@@ -331,4 +334,83 @@ void WebGLRenderingContextBase::set_error(GLenum error)
...
+    auto& realm = this->realm();
+    auto promise = WebIDL::create_promise(realm);
+
+    // 3. Let context be this.
+    auto context = this;
+
+    // 4. Run the following steps in parallel:
+    Platform::EventLoopPlugin::the().deferred_invoke(GC::create_function(realm.heap(), [&realm, context, promise]() {
+        // 1. Let device be the result of ensuring an immersive XR device is selected.
+        // FIXME: Implement https://immersive-web.github.io/webxr/#ensure-an-immersive-xr-device-is-selected
```
gmta, line 360: Maybe move this FIXME to the "If device is null" part below, similar to the other FIXME you added?

Libraries/LibWeb/WebGL/WebGL2RenderingContext.h
```diff
@@ -31,7 +31,6 @@ class WebGL2RenderingContext final : public WebGL2RenderingContextOverloads {
 
     GC::Ref<HTML::HTMLCanvasElement> canvas_for_binding() const;
 
-    bool is_context_lost() const;
```
gmta, line 34: You forgot to remove `m_context_lost` from this class.

Libraries/LibIDL/IDLParser.cpp
```diff
@@ -733,7 +733,7 @@ void Parser::parse_deleter(HashMap<ByteString, ByteString>& extended_attributes,
 void Parser::parse_interface(Interface& interface)
 {
     consume_whitespace();
-    interface.name = parse_identifier_ending_with_space();
+    interface.name = parse_identifier_ending_with_space_or(':');
```
gmta, line 737: Oof, I understand why we need this for the IDL file you added, but this is rather ugly. We should probably stop at any character that is not valid as part of an IDL identifier. Practically this would be a space, `:` or `{` I guess, but it would be nice to have a more concise identifier parsing logic available.

Libraries/LibWeb/WebXR/XRLayer.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+XRLayer::XRLayer(JS::Realm& realm)
+    : DOM::EventTarget(realm)
+{
+}
+
+void XRLayer::visit_edges(Cell::Visitor& visitor)
+{
+    Base::visit_edges(visitor);
+}
```
gmta: Don't need to add this if all you do is call `Base::visit_edges()`.

Libraries/LibWeb/WebXR/XRRenderState.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+XRRenderState::XRRenderState(JS::Realm& realm)
+    : PlatformObject(realm)
+{
+}
+
+void XRRenderState::visit_edges(Cell::Visitor& visitor)
+{
+    Base::visit_edges(visitor);
+}
```
gmta: Same here - useless override.

Libraries/LibWeb/WebXR/XRRenderState.idl
```diff
@@ -0,0 +1,18 @@
...
+dictionary XRRenderStateInit {
+    double depthNear;
+    double depthFar;
+    boolean passthroughFullyObscured;
+    double inlineVerticalFieldOfView;
+    // FIXME: XRWebGLLayer? baseLayer;
+    // FIXME: sequence<XRLayer>? layers;
+};
+
+[SecureContext, Exposed=Window] interface XRRenderState {
```
gmta: Should we mark these `Experimental`?

Libraries/LibWeb/WebXR/XRWebGLLayer.idl
```diff
@@ -0,0 +1,36 @@
...
+    double framebufferScaleFactor = 1.0;
+};
+
+[SecureContext, Exposed=Window]
+interface XRWebGLLayer: XRLayer {
+    constructor(XRSession session,
+               XRWebGLRenderingContext context,
+               optional XRWebGLLayerInit layerInit = {});
+    // Attributes
+    // FIXME: readonly attribute boolean antialias;
```
gmta: We have the `[FIXME]` attribute for all of these in all IDL files you added/modified.

Libraries/LibWeb/WebXR/XRRenderState.idl
Psychpsyo: Probably, yes. I've gone ahead and marked all the WebXR IDL interfaces as experimental.

Libraries/LibIDL/IDLParser.cpp
Psychpsyo: I've added a FIXME for now.

Libraries/LibWeb/WebGL/WebGLRenderingContextBase.cpp
Psychpsyo: I guess, maybe? I find it more intuitive to have the FIXME for implementing the algorithm at the call site instead of putting it three spec steps down where we'd want to use the variable that doesn't exist yet.

PR #7979, opened 2026-02-16
LibWeb: Parse `@font-feature-values`

Calme1709:
We don't do anything with the rule yet, that will come in a later PR. See individual commits for details.

Gains us ~325 WPT subtest passes.

### DIFF
--- a/Libraries/LibWeb/CSS/Parser/RuleParsing.cpp
+++ b/Libraries/LibWeb/CSS/Parser/RuleParsing.cpp
@@ -14,6 +14,7 @@
 
 #include <LibWeb/CSS/CSSCounterStyleRule.h>
 #include <LibWeb/CSS/CSSFontFaceRule.h>
+#include <LibWeb/CSS/CSSFontFeatureValuesRule.h>
 #include <LibWeb/CSS/CSSImportRule.h>
 #include <LibWeb/CSS/CSSKeyframeRule.h>
 #include <LibWeb/CSS/CSSKeyframesRule.h>
@@ -105,6 +106,9 @@ GC::Ptr<CSSRule> Parser::convert_to_rule(Rule const& rule, Nested nested)
 
             if (at_rule.name.equals_ignoring_ascii_case("font-face"sv))
                 return convert_to_font_face_rule(at_rule);
+
+            if (at_rule.name.equals_ignoring_ascii_case("font-feature-values"sv))
+                return convert_to_font_feature_values_rule(at_rule);
 
             if (at_rule.name.equals_ignoring_ascii_case("import"sv))
                 return convert_to_import_rule(at_rule);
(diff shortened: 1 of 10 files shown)


Libraries/LibWeb/CSS/Parser/RuleParsing.cpp
```diff
@@ -960,6 +964,242 @@ GC::Ptr<CSSFontFaceRule> Parser::convert_to_font_face_rule(AtRule const& rule)
...
+            .rule_name = "@font-feature-values"_fly_string,
+            .prelude = tokens.dump_string(),
+            .description = "Empty family name list."_string,
+        });
+        return {};
+    }
+
+    for (auto const& family_component_values : comma_separated_families) {
+        // https://drafts.csswg.org/css-fonts-4/#family-name-syntax
+        // <family-name> = <string> | <custom-ident>+
```
AtkinsSJ: It's awkward having to duplicate the `<family-name>` parsing code here. I get why - to avoid creating StyleValues only to throw them away again - but maybe the two places could share the implementation? (Not a blocker on merging.)

Libraries/LibWeb/CSS/Parser/RuleParsing.cpp
```diff
@@ -960,6 +964,242 @@ GC::Ptr<CSSFontFaceRule> Parser::convert_to_font_face_rule(AtRule const& rule)
...
+                    });
+                    return;
+                }
+
+                Vector<u32> values;
+
+                while (value_stream.has_next_token()) {
+                    auto token = value_stream.consume_a_token();
+
+                    // FIXME: Support calc()
```
AtkinsSJ, line 1084: This seems like it should be as simple as `parse_integer_value()` and then calling `number_from_style_value()` on the result if it's not null, now that your other PR is merged.

Meta/Lagom/Tools/CodeGenerators/LibWeb/BindingsGenerator/IDLGenerators.cpp
```diff
@@ -4988,6 +5035,209 @@ JS_DEFINE_NATIVE_FUNCTION(@class_name@::clear)
...
+    return map->map_has(key);
+}
+)~~~");
+
+        if (!interface.overload_sets.contains("set"sv) && !interface.is_map_readonly) {
+            maplike_generator.append(R"~~~(
+// https://webidl.spec.whatwg.org/#js-map-set
+JS_DEFINE_NATIVE_FUNCTION(@class_name@::add)
+{
+    WebIDL::log_trace(vm, "@class_name@::add");
```
AtkinsSJ: I expect this is correct but just wanted to make sure: this is named `add()` when the IDL method is `set()`.

Libraries/LibWeb/CSS/CSSFontFeatureValuesMap.cpp
```diff
@@ -0,0 +1,80 @@
...
+}
+
+WebIDL::ExceptionOr<void> CSSFontFeatureValuesMap::set(String const& feature_value_name, Variant<u32, Vector<u32>> const& values)
+{
+    // https://drafts.csswg.org/css-fonts-4/#cssfontfeaturevaluesmap
+    // The CSSFontFeatureValuesMap interface uses the default map class methods but the set method has different
+    // behavior. It takes a sequence of unsigned integers and associates it with a given featureValueName. The method
+    // behaves the same as the default map class method except that
+    //
+    // a single unsigned long value is treated as a sequence of a single value.
```
AtkinsSJ, line 35: Nit: The wrapping here is a bit odd.

Libraries/LibWeb/CSS/Parser/RuleParsing.cpp
Calme1709: I had (falsely) assumed that any existing functionality would have supported `<generic-family>` as well which is why I re-implemented this. I think that `@font-feature-values` will be used infrequently enough that the performance impact of creating and then throwing away `StyleValue`s is largely irrelevant so I have updated this to use `parse_family_name_value` (and added a `string_from_style_value` method)

Libraries/LibWeb/CSS/Parser/RuleParsing.cpp
Calme1709: It's a bit more involved than that unfortunately:
 - We will need to add a special value parsing context in order to correctly clamp calculated values to be non-negative.
 - We can't use `number_from_style_value()` since it will have an `IntegerStyleValue` rather than a `NumberStyleValue` (a good excuse to add an `int_from_style_value`)
 - We run into the same [problem](https://github.com/LadybirdBrowser/ladybird/blob/19bf3f9479f32de981f2c49820a562f3fb18912b/Libraries/LibWeb/CSS/Parser/DescriptorParsing.cpp#L188) as some of the `@counter-style` descriptors if we don't have a document to generate a length resolution context from

All solvable problems but I decided to leave this as a FIXME for now to avoid an even bigger diff than we already have.

Libraries/LibWeb/CSS/CSSFontFeatureValuesMap.cpp
Calme1709: Done intentionally to pair up that statement with the corresponding code below (rather than just having it at the end of a wall of text) - I have removed the ugly empty comment line above.

Meta/Lagom/Tools/CodeGenerators/LibWeb/BindingsGenerator/IDLGenerators.cpp
Calme1709: Good catch - this was a bad copy paste from `setlike` which wasn't caught since we overload `set` in `FontFeatureValuesMap`. Fixed.

Calme1709 (commented):
Changes:
- Applied feedback
- Rebased to master

Libraries/LibWeb/CSS/Parser/RuleParsing.cpp
AtkinsSJ: Fair enough!

PR #7991, opened 2026-02-17
LibWeb: Skip microtask checkpoint in reentrant event loops

jonbgamble:
[ECMAScript’s jobs model](https://tc39.es/ecma262/#sec-jobs) requires jobs to run when no execution context is actively running, and HTML adheres to and mirrors that by triggering checkpoints from script cleanup when the execution-context stack is empty. Guarding against reentrant checkpoints preserves the requirements for correct microtask execution.

HTML defines the host hook for those ECMAScript promise jobs in its microtask queue implementation: [HostEnqueuePromiseJob](https://html.spec.whatwg.org/multipage/webappapis.html#hostenqueuepromisejob)

### DIFF
--- a/Libraries/LibWeb/HTML/EventLoop/EventLoop.cpp
+++ b/Libraries/LibWeb/HTML/EventLoop/EventLoop.cpp
@@ -610,13 +610,13 @@ void EventLoop::perform_a_microtask_checkpoint()
     if (execution_paused())
         return;
 
-    // NOTE: This assertion is per requirement 9.5 of the ECMA-262 spec, see: https://tc39.es/ecma262/#sec-jobs
-    // > At some future point in time, when there is no running context in the agent for which the job is scheduled and that agent's execution context stack is empty...
-    VERIFY(vm().execution_context_stack().is_empty());
-
     // 1. If the event loop's performing a microtask checkpoint is true, then return.
     if (m_performing_a_microtask_checkpoint)
         return;
+
+    // NOTE: This assertion is per requirement 9.5 of the ECMA-262 spec, see: https://tc39.es/ecma262/#sec-jobs
+    // > At some future point in time, when there is no running context in the agent for which the job is scheduled and that agent's execution context stack is empty...
+    VERIFY(vm().execution_context_stack().is_empty());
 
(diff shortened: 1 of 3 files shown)


Zaggy1024:
If the HTML spec is supposedly adhering to this, shouldn't this be an assertion rather than a conditional? I'm assuming we have places that break this assumption, and I wonder if those need to be addressed first.

jonbgamble:
It is verified [here](https://github.com/LadybirdBrowser/ladybird/blob/4facec1b4036e5fed1347d136401415815591f80/Libraries/LibWeb/HTML/EventLoop/EventLoop.cpp#L615).

Maybe we just VERIFY that a bit too early. The ECMAScript jobs spec is more like "only execute when the execution context is empty". It does not say "non-empty execution context is a program error".

Meanwhile, the [event loop spec](https://html.spec.whatwg.org/multipage/webappapis.html#perform-a-microtask-checkpoint) mentions guarding against re-entrancy with the m_performing_a_microtask_checkpoint flag. This seems like evidence that it's OK to be here with a non-empty stack, and we should just return rather than crash.

shannonbooth:
In what scenario are we having a crash because of this? This has come up as an issue beforehand in situations we were spinning the event loop in places where it has not been ideal to do so and instead should be using patterns like DocumentObserver

jonbgamble:
I am actually grateful for this VERIFY's previous location because the crash instigated a cleaner code path vs spinning unnecessarily (during input device enumeration for media capture, not in master branch). I dealt with it outside of EventLoop using a better async model, but this scenario can happen from user scripting, not just bad browser code. See attached test.

So unfortunately, I think VERIFY has to change.

Maybe we could still keep it in front of the re-entrancy guard if we back it down to an ASSERT, and remove this test? Then we get the same protection against dumb browser code paths in debug builds while allowing the full range of behavior allowed by spec to not crash in release builds.

shannonbooth:
Right, that was my hesitancy on that, since it's mostly come up in that type of scenario beforehand. But indeed looks like I put that in the wrong place 😅 Could you please adjust the commit message to reflect the latest code? After that, looks good to me!

shannonbooth (approved):

Issue #3241, opened 2025-01-13
SVG foreignObject does not establish a containing block for abspos children.


Psychpsyo:
### Summary

When I put an absolutely-positioned element in an SVG &lt;foreignObject&gt;, in Ladybird, it is positioned relative to the next abspos-containing-block-creating ancestor outside the entire SVG, not the foreignObject.

See [this resolution](https://github.com/w3c/fxtf-drafts/issues/307#issuecomment-499612420) and behavior in other browsers, which, to my knowledge, has not lead to a spec change yet.
The only thing I can find in the SVG spec is the note at the bottom of [Establishing a stacking context in SVG](https://www.w3.org/TR/SVG2/render.html#EstablishingStackingContex), which only speaks of a `"fixed position containing block" in CSS terms`.

### Operating system

Linux

### Steps to reproduce

1. Make an SVG with a &lt;foreignObject&gt; in it.
2. Put anything with `position: absolute; top: 0; left: 0` inside that.
3. Notice how that element aligns with the top left of whatever is outside the SVG, not the foreignObject.

### Expected behavior

A foreignObject should create a containing block for absolutely (and fixed) positioned elements.

### Actual behavior

A foreignObject does not create a containing block for absolutely positioned elements.

### URL for a reduced test case

N/A

### HTML/SVG/etc. source for a reduced test case

```html
<!DOCTYPE html>
<style>
  div {
    position: absolute;
    top: 0;
    left: 0;
    width: 100px;
    height: 100px;
  }
</style>
<svg width="200" height="200">
  <foreignObject width="100" height="100" x="50" y="50">
    <div>I am offset!</div>
  </foreignObject>
</svg>
```

### Log output and (if possible) backtrace

```shell
None
```

### Screenshots or screen recordings

The reduced test case in:
Firefox:
![image](https://github.com/user-attachments/assets/6bfaefcb-346d-4209-9059-a9c3a7d60d62)

Ladybird:
![image](https://github.com/user-attachments/assets/bfc2dd7c-716a-4a48-b902-5ba59705d22a)

### Build flags or config settings

None

### Contribute a patch?

- [ ] I’ll contribute a patch for this myself.

PR #7993, opened 2026-02-17
LibWeb: Make foreignObject establish a containing block for abspos

kalenikaliaksandr:
Absolutely positioned elements inside SVG foreignObject were being positioned relative to an ancestor containing block outside the SVG, instead of relative to the foreignObject itself. Per a W3C resolution and the behavior of other browsers, foreignObject should establish a containing block for absolutely and fixed positioned elements.

With this fix, the `has_abspos_with_external_containing_block` check in `set_needs_layout_update()` and the abspos preservation loop in `relayout_svg_root()` become dead code — remove both and simplify the ancestor loops. Rename related tests to reflect the new behavior.

Fixes https://github.com/LadybirdBrowser/ladybird/issues/3241


Libraries/LibWeb/Layout/Node.cpp
```diff
@@ -160,6 +160,11 @@ bool Node::establishes_an_absolute_positioning_containing_block() const
     if (is<Viewport>(*this))
         return true;
 
+    // https://github.com/w3c/fxtf-drafts/issues/307#issuecomment-499612420
+    // foreignObject establishes a containing block for absolutely and fixed positioned elements.
+    if (is_svg_foreign_object_box())
```
gmta, line 165: This is an `is<T>` optimization, maybe use `if (is<SVGForeignObjectBox>())` so it's consistent with the previous two checks?

kalenikaliaksandr: with `is<SVGForeignObjectBox>()` we would have to include `SVGForeignObjectBox.h` in `Node.cpp`, which isn't needed with `is_svg_foreign_object_box()`. I think it might be more reasonable to update other ifs in this function to use `is_` 😄

gmta: OK, let's leave it as-is then! Kind of a bummer though. Would be nice to have some forward decls for this somehow.

PR #8001, opened 2026-02-17
LibWebView: Treat cancelled web event as handled

jeetsh4h:
`Cancelled` events should not be propagated back to the client.
The previous behavior only returned early for `Handled`, which could incorrectly forward cancelled events.

### DIFF
--- a/Libraries/LibWebView/ReservedBrowserShortcuts.h
+++ b/Libraries/LibWebView/ReservedBrowserShortcuts.h
@@ -7,6 +7,7 @@
 #pragma once
 
 #include <AK/Array.h>
+#include <LibWeb/Page/EventResult.h>
 #include <LibWeb/Page/InputEvent.h>
 #include <LibWeb/UIEvents/KeyCode.h>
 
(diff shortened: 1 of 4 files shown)


trflynn89:
IIRC we don't return early for cancelled events here because we want to be able to handle UI shortcuts for at least some subset of cancelled events. For example, if a page cancels `ctrl+w`, we want the UI to still receive that shortcut and close the tab.

jeetsh4h:
I was checking if there are any bugs in the current build of the browser for my personal website: [jeetsh4h.dev](https://jeetsh4h.dev/terminal) where pressing tab autocompletes for the suggestion but also propagates the Event to cause Tab navigation (which is the behavior in Chrome and Firefox).

Shortcuts like `ctrl+w` and `ctrl+t` could be treated as special cases, instead?

Regardless, will commit a regression test for the same.

Psychpsyo:
Driveby comment: Could it make sense to tie this to the ctrl key?
I know Firefox lets the user do things like Ctrl + Right-click on pages on that would usually prevent opening the right-click menu to open it anyways.

jeetsh4h:
I was experimenting with keeping a special case for the `ctrl` key modifier but a lot of sites have `ctrl+k` as a shortcut for quite a bit of functionality. 
Having an explicit list of pre-determined key combinations will probably be the way to go.

Issue #3253, opened 2025-01-14
Still can't play YouTube videos


ChaseKnowlden:
### Summary

In Ladybird, any YouTube video won't play, because it just shows Your browser can't play this video.

### Operating system

macOS

### Steps to reproduce

1. Open Ladybird
2. Navigate to youtube.com
3. Go to any YouTube video
4. Video doesn't play

### Expected behavior

The video should play.

### Actual behavior

The video doesn't play.

### URL for a reduced test case

https://www.youtube.com/watch?v=cbw0KrMGHvc

### HTML/SVG/etc. source for a reduced test case

```html
N/A
```

### Log output and (if possible) backtrace

```shell
593568.760 ImageDecoder(84227): libpng warning: iCCP: known incorrect sRGB profile
593568.761 ImageDecoder(84227): libpng warning: iCCP: known incorrect sRGB profile
593568.762 ImageDecoder(84227): libpng warning: iCCP: known incorrect sRGB profile
593568.763 ImageDecoder(84227): libpng warning: iCCP: known incorrect sRGB profile
593569.077 WebContent(84228): ResourceLoader: Filtered request to: "https://rr5---sn-vgqskn6z.googlevideo.com/generate_204"
593569.078 WebContent(84228): ResourceLoader: Filtered request to: "https://rr5---sn-vgqskn6z.googlevideo.com/generate_204?conn2"
593569.882 WebContent(84228): FIXME: Unimplemented IDL interface: 'MediaCapabilitiesPrototype.decodingInfo'
593569.908 WebContent(84228): FIXME: Unimplemented IDL interface: 'LocationPrototype.ancestorOrigins'
593569.910 WebContent(84228): FIXME: Unimplemented IDL interface: 'HTMLMediaElementPrototype.defaultPlaybackRate'
593569.910 WebContent(84228): FIXME: Unimplemented IDL interface: 'MediaCapabilitiesPrototype.decodingInfo'
593569.914 WebContent(84228): FIXME: Unimplemented IDL interface: 'IDBTransactionPrototype.objectStoreNames'
593569.950 WebContent(84228): FIXME: Unimplemented IDL interface: 'HTMLMediaElementPrototype.playbackRate'
593569.950 WebContent(84228): FIXME: Unimplemented IDL interface: 'HTMLMediaElementPrototype.playbackRate'
593569.954 WebContent(84228): FIXME: Unimplemented IDL interface: 'SourceBufferPrototype.changeType'
[...]

PR #8007, opened 2026-02-17
LibWeb+LibWebView+Services: Add a flag to enable experimental interfaces

trflynn89:
This adds the `--expose-experimental-interfaces` command line flag to enable experimental IDL interfaces. Any IDL interface with `Experimental` in its exposed attributes will be disabled by default.

The problem is that by stubbing out or partially implementing interfaces in LibWeb, we actually make some sites behave worse. For example, the `OffscreenCanvas` interface being exposed makes sites believe we fully support it, even though we don't. If the interface was not exposed, these sites may fall back to ordinary canvas objects. Similarly, to use YouTube, we currently have to patch out MSE interfaces.

This flag will allow developers to iteratively work on features, without breaking such sites. We enable experimental interfaces during tests.

<hr />

Right now this is just a boolean flag. If we want more control over which experimental interfaces get exposed, we can expand this in the future.

### DIFF
--- a/Libraries/LibWeb/MediaSourceExtensions/ManagedMediaSource.idl
+++ b/Libraries/LibWeb/MediaSourceExtensions/ManagedMediaSource.idl
@@ -2,7 +2,7 @@
 #import <MediaSourceExtensions/MediaSource.idl>
 
 // https://w3c.github.io/media-source/#managedmediasource-interface
-[Exposed=(Window,DedicatedWorker)]
+[Exposed=(Window,DedicatedWorker), Experimental]
 interface ManagedMediaSource : MediaSource {
     constructor();
     [FIXME] readonly attribute boolean streaming;
--- a/Libraries/LibWeb/MediaSourceExtensions/MediaSource.idl
+++ b/Libraries/LibWeb/MediaSourceExtensions/MediaSource.idl
@@ -16,7 +16,7 @@ enum EndOfStreamError {
 };
 
 // https://w3c.github.io/media-source/#dom-mediasource
-[Exposed=(Window,DedicatedWorker)]
+[Exposed=(Window,DedicatedWorker), Experimental]
 interface MediaSource : EventTarget {
     constructor();
 
--- a/Libraries/LibWeb/MediaSourceExtensions/MediaSourceHandle.idl
+++ b/Libraries/LibWeb/MediaSourceExtensions/MediaSourceHandle.idl
@@ -1,3 +1,3 @@
 // https://w3c.github.io/media-source/#dom-mediasourcehandle
-[Transferable, Exposed=(Window,DedicatedWorker)]
+[Transferable, Exposed=(Window,DedicatedWorker), Experimental]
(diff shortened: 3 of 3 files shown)


InvalidUsernameException:
Maybe hide the MSE interfaces behind the flag as well to enable YouTube 360p playback without source-patches?

https://github.com/LadybirdBrowser/ladybird/issues/3253#issuecomment-3046436615

shannonbooth:
This should probably also be implemented in is_exposed in GenerateWindowOrWorkerInterfaces

Libraries/LibWebView/HelperProcess.cpp
```diff
@@ -114,6 +114,8 @@ static ErrorOr<NonnullRefPtr<WebView::WebContentClient>> launch_web_content_proc
         arguments.append("--enable-idl-tracing"sv);
     if (web_content_options.enable_http_memory_cache == WebView::EnableMemoryHTTPCache::Yes)
         arguments.append("--enable-http-memory-cache"sv);
+    if (web_content_options.expose_experimental_interfaces == WebView::ExposeExperimentalInterfaces::Yes)
+        arguments.append("--expose-experimental-interfaces"sv);
```
shannonbooth, line 118: Probably very overboard at this stage, but a nice syntax might be something like "--expose Internals --expose Experimental" to generalize the mechanism to Internals object as well and have that implemented in IDL as well. But yeah, probably too much for only two options for now! (though I think this will definitely grow)

Libraries/LibWeb/HTML/Window.h
```diff
@@ -264,8 +264,12 @@ class WEB_API Window final
 
     void consume_history_action_user_activation();
 
-    static bool in_test_mode();
     static void set_enable_test_mode(bool);
+    static bool in_test_mode();
+
+    static void set_experimental_interfaces_exposed(bool);
+    static bool are_experimental_interfaces_exposed();
```
shannonbooth: Given that this flag works for Workers too, we should probably move this to UniversalGlobalScope (or similar) as a better home. This is in converse to some of the other flags here like `in_test_mode` which don't work in Workers since it's not passed along

Libraries/LibWebView/HelperProcess.cpp
trflynn89: Yeah I figure we can settle on a convention as we add more experimental stuff

Libraries/LibWeb/HTML/Window.h
trflynn89: hmm, `UniversalGlobalScopeMixin` or `WindowOrWorkerGlobalScopeMixin`? Would this be affected by shadow realms?

shannonbooth: I suppose it may not ever be applicable to ShadowRealms since that exposes such a small set of interfaces, but I suppose in theory there could be an interface exposed on ShadowRealm which we want disabled (where UniversalGlobalScopeMixin makes sense since that's the common base global)

PR #8009, opened 2026-02-17
UI+LibWebView: Add "Open in New Window" option in right-click context menu

jeetsh4h:
~~Also fix behavior for Qt when a new window is opened with an empty URL list. Open the new tab link by default so one tab is always open in a window.~~

As pointed out by @Zaggy1024, it has already been addressed in commit (c35c29993a11eaa223ccb87d00d061795d02d8bb).

### DIFF
--- a/Libraries/LibWebView/Application.cpp
+++ b/Libraries/LibWebView/Application.cpp
@@ -400,6 +400,11 @@ void Application::open_bookmark_in_new_tab(String const& bookmark_id, Web::HTML:
         open_url_in_new_tab(bookmark->bookmark().url, activate_tab);
 }
 
+void Application::open_url_in_new_window(URL::URL const& url)
+{
+    dbgln("open_url_in_new_window() is unsupported on this platform (url: {})", url);
+}
+
 static ErrorOr<NonnullRefPtr<WebContentClient>> create_web_content_client(Optional<ViewImplementation&> view)
 {
     auto request_server_handle = TRY(connect_new_request_server_client());
--- a/UI/Qt/Application.cpp
+++ b/UI/Qt/Application.cpp
@@ -150,6 +150,11 @@ Optional<WebView::ViewImplementation&> Application::open_blank_new_tab(Web::HTML
     return tab.view();
 }
 
+void Application::open_url_in_new_window(URL::URL const& url)
+{
+    this->new_window({ url });
+}
+
 Optional<ByteString> Application::ask_user_for_download_path(StringView file) const
 {
     auto default_path = QStandardPaths::writableLocation(QStandardPaths::DownloadLocation);
--- a/Libraries/LibWebView/Application.h
+++ b/Libraries/LibWebView/Application.h
@@ -88,6 +88,7 @@ public:
     void open_bookmark_in_new_tab(String const& bookmark_id, Web::HTML::ActivateTab) const;
 
     Main::Arguments const& command_line_arguments() const { return m_arguments; }
+    virtual void open_url_in_new_window(URL::URL const& url);
 
     void add_child_process(Process&&);
 
--- a/Libraries/LibWebView/Menu.h
+++ b/Libraries/LibWebView/Menu.h
@@ -58,6 +58,7 @@ enum class ActionID {
     ViewSource,
 
     OpenInNewTab,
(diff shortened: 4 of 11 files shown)


Libraries/LibWebView/Application.cpp
```diff
@@ -313,6 +313,13 @@ void Application::open_url_in_new_tab(URL::URL const& url, Web::HTML::ActivateTa
         view->load(url);
 }
 
+void Application::open_url_in_new_window(URL::URL const& url) const
+{
+    // Fallback for platforms that don't provide a window-level override.
+    // Platform UI layers should override this to create a new window.
+    open_url_in_new_tab(url, Web::HTML::ActivateTab::Yes);
+}
```
konradekk: i believe no fallback should be necessary. the action should either be marked as “unimplemented (yet)” or removed entirely as “unsupported (ever).” in the first case, it should either throw an exception or act as a temporary no-op. in the second case, a small mechanism should be introduced to hide the option. imho, ofc. 🤔

jeetsh4h: That does make sense, both platforms already do support it. I will make the change to remove it.

UI/Qt/Application.cpp
```diff
@@ -152,6 +152,13 @@ Optional<WebView::ViewImplementation&> Application::open_blank_new_tab(Web::HTML
     return tab.view();
 }
 
+void Application::open_url_in_new_window(URL::URL const& url) const
+{
+    // NOTE: This override is const to match the cross-platform API, but opening a
+    // window updates UI state (active window), so we delegate to the non-const helper.
+    (void)const_cast<Application&>(*this).new_window({ url });
```
Zaggy1024: I'm not sure it makes sense to do it this way if it requires a const_cast. Honestly, it makes me wonder whether open_blank_new_tab() should be non-const as well, but that's not relevant here.

Zaggy1024 (commented):
For commits that fix prior issues in the PR, it's better to amend the commit to apply the fix, so the third commit should probably be squashed into the second.

jeetsh4h:
> For commits that fix prior issues in the PR, it's better to amend the commit to apply the fix, so the third commit should probably be squashed into the second.

That does make sense, thank you, I shall update the commit structure for the same.

UI/Qt/Application.cpp
jeetsh4h: The `const_cast` is a workaround. I can remove the `open_url_in_new_window()` `const` qualifier from the base cross-platform API. In my head, the application state has to mutate when opening a new window.

Should I make a new commit that specifically addresses this?

jeetsh4h: I have rebased my commits to add this fix

Zaggy1024:
Sorry for letting this sit, it slipped my mind. :( Reopening in case you're able to rebase so that it can be merged.

jeetsh4h:
@Zaggy1024 I have rebased the branch. Let me know if any other changes are required

UI/Qt/BrowserWindow.cpp
```diff
@@ -370,6 +370,8 @@ BrowserWindow::BrowserWindow(Vector<URL::URL> const& initial_urls, IsPopupWindow
 
     if (parent_tab) {
         new_child_tab(Web::HTML::ActivateTab::Yes, *parent_tab, AK::move(page_index));
+    } else if (initial_urls.is_empty()) {
+        new_tab_from_url(WebView::Application::settings().new_tab_page_url(), Web::HTML::ActivateTab::Yes);
```
Zaggy1024: There was a commit fairly recently that fixed an issue that seems related to this change (c35c29993a11eaa223ccb87d00d061795d02d8bb). We shouldn't really be passed an empty vector here. If the above commit was addressing the same issue, probably good to drop this change.

jeetsh4h: Dropped the redundant Qt empty-vector workaround after the note about c35c29993a11. The branch now only contains the “Open in New Window” context-menu change

Zaggy1024 (approved):

Zaggy1024:
Oops, I didn't notice the workflows weren't approved yet here. Will keep an eye on this to get it merged. 😅

PR #8010, opened 2026-02-17
LibWeb: Account for generic font families in FontFaceSet::load()

Calme1709:
Fixes a regression in d998a0a which was crashing the imported test


Calme1709:
Rebased to bring CI config in line with master

Issue #4638, opened 2025-05-07
SVG rendering as a black box


van-sprundel:
### Summary

One of the SVGs on the site is a black box

### Operating system

macOS

### Steps to reproduce

1. go to given URL (or use given SVG)
2. open in ladybird
3. result should be a black box

### Expected behavior

The first screenshot shows what the SVG should look like

### Actual behavior

The result is a black box

### URL for a reduced test case

https://backend.teifi.ca/uploads/image_2_00e451c2f1.svg

### HTML/SVG/etc. source for a reduced test case

```html
<svg width="42" height="42" viewBox="0 0 42 42" fill="none" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
<rect width="42" height="42" fill="url(#pattern0)"/>
<defs>
<pattern id="pattern0" patternContentUnits="objectBoundingBox" width="1" height="1">
<use xlink:href="#image0_2_328" transform="scale(0.0192308)"/>
</pattern>
[...]

PR #8011, opened 2026-02-17
LibWeb: Implement `SVGPatternElement`

tcl3:
This implementation renders the pattern into a separate DisplayList, which is then tiled using a shader with `SkTileMode::kRepeat`

I've included other commits that were needed to  get the test case in the associated issue working.

Fixes: #4638

### DIFF
--- a/Libraries/LibWeb/SVG/SVGElement.cpp
+++ b/Libraries/LibWeb/SVG/SVGElement.cpp
@@ -110,6 +110,8 @@ static ReadonlySpan<NamedPropertyID> attribute_style_properties()
         NamedPropertyID(CSS::PropertyID::TextDecoration),
         NamedPropertyID(CSS::PropertyID::TextRendering),
         NamedPropertyID(CSS::PropertyID::TextOverflow),
+        NamedPropertyID(CSS::PropertyID::Transform, SVG::AttributeNames::gradientTransform, { SVG::TagNames::linearGradient, SVG::TagNames::radialGradient }),
+        NamedPropertyID(CSS::PropertyID::Transform, SVG::AttributeNames::patternTransform, { SVG::TagNames::pattern }),
         NamedPropertyID(CSS::PropertyID::TransformOrigin),
         NamedPropertyID(CSS::PropertyID::UnicodeBidi),
         NamedPropertyID(CSS::PropertyID::Visibility),
--- a/Tests/LibWeb/Text/expected/wpt-import/svg/styling/presentation-attributes-special-cases.txt
+++ b/Tests/LibWeb/Text/expected/wpt-import/svg/styling/presentation-attributes-special-cases.txt
@@ -2,8 +2,8 @@ Harness status: OK
 
 Found 26 tests
 
-17 Pass
-9 Fail
+20 Pass
+6 Fail
 Pass	cx and cy presentation attributes supported on circle element
 Pass	cx and cy presentation attributes supported on ellipse element
 Pass	x, y, width, and height presentation attributes supported on foreignObject element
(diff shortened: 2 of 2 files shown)


Libraries/LibWeb/Layout/SVGFormattingContext.cpp
```diff
@@ -513,12 +522,41 @@ void SVGFormattingContext::layout_mask_or_clip(SVGBox const& mask_or_clip)
...
+            layout_state.set_content_width(parent_node_state.content_width());
+            layout_state.set_content_height(parent_node_state.content_height());
+            // https://svgwg.org/svg2-draft/pservers.html#PatternElementPatternContentUnitsAttribute
+            parent_viewbox_transform = Gfx::AffineTransform {}
+                                           .translate(parent_node_state.offset.to_type<float>())
+                                           .scale(parent_node_state.content_width().to_float(), parent_node_state.content_height().to_float());
+        } else {
+            layout_state.set_content_width(m_viewport_size.width());
+            layout_state.set_content_height(m_viewport_size.height());
+        }
```
gmta, line 548: This looks like quite a bit of duplicate code, maybe you can specialize the case with:

```cpp
if (is<SVGPatternBox>(mask_or_clip) && pattern.view_box().has_value()) {
    ...
```

And then add this below:

```cpp
if (is<SVGPatternBox>(mask_or_clip) && content_units == SVG::SVGUnits::ObjectBoundingBox)
    parent_viewbox_transform.scale(parent_node_state.content_width().to_float(), parent_node_state.content_height().to_float());
```

tcl3: I've simplified the if/else mess a bit.

Libraries/LibWeb/Painting/DisplayListPlayerSkia.cpp
```diff
@@ -526,6 +527,40 @@ static SkPaint paint_style_to_skia_paint(Painting::SVGGradientPaintStyle const&
...
 }
 
+SkPaint DisplayListPlayerSkia::paint_style_to_skia_paint(Painting::SVGPaintServerPaintStyle const& paint_style, Gfx::FloatRect const& bounding_rect)
+{
+    if (auto const* gradient = as_if<SVGGradientPaintStyle>(paint_style))
+        return gradient_paint_style_to_skia_paint(*gradient, bounding_rect);
+
+    if (auto const* pattern = as_if<SVGPatternPaintStyle>(paint_style)) {
+        auto const& tile_rect = pattern->tile_rect();
+        auto tile_size = Gfx::IntSize(ceilf(tile_rect.width()), ceilf(tile_rect.height()));
```
gmta, line 541: Wouldn't `tile_rect.size().to_rounded<int>()` give a slightly better result here?

Libraries/LibWeb/SVG/SVGPatternElement.cpp
```diff
@@ -0,0 +1,391 @@
...
+        tile_width = pattern_width().resolve_relative_to(viewport.width());
+        tile_height = pattern_height().resolve_relative_to(viewport.height());
+    }
+
+    if (tile_width <= 0 || tile_height <= 0)
+        return {};
+
+    auto tile_rect = paint_context.paint_transform.map(Gfx::FloatRect { tile_x, tile_y, tile_width, tile_height });
+
+    if (tile_rect.width() <= 0 || tile_rect.height() <= 0)
```
gmta: `.is_empty()`

Libraries/LibWeb/SVG/SVGPatternElement.cpp
```diff
@@ -0,0 +1,391 @@
...
+        return TraversalDecision::Continue;
+    });
+
+    Painting::StackingContext::paint_svg(paint_context_copy, *pattern_paintable, Painting::PaintPhase::Foreground);
+
+    // https://svgwg.org/svg2-draft/pservers.html#PatternElementPatternTransformAttribute
+    // The 'patternTransform' attribute is an alias for the 'transform' property.
+    // Resolve the CSS 'transform' property from the pattern box's computed values, which
+    // accounts for both the patternTransform presentation attribute and any inline style
+    // override (e.g. pattern.style.transform = '...').
```
gmta: Can we wrap at column 120 everywhere?

Libraries/LibWeb/Painting/DisplayListPlayerSkia.cpp
tcl3: I round up here to avoid clipping the fractional-pixel edge of the pattern. Other engines do the same. I did find that rounding up here can cause seams to appear between pattern tiles, so I've introduced a scaling factor to avoid that.

Issue #7945, opened 2026-02-14
Page crashes when opening https://www.allsop-lancaster.com/knf/goodies.html


Psychpsyo:
### Summary

Page crashes when opening https://www.allsop-lancaster.com/knf/goodies.html

### Operating system

Linux

### Steps to reproduce

1. Open https://www.allsop-lancaster.com/knf/goodies.html
2. Page crashes.

### Expected behavior

Page should not crash.

### Actual behavior

Page crashes.

### URL for a reduced test case

https://www.allsop-lancaster.com/knf/goodies.html

### HTML/SVG/etc. source for a reduced test case

```html
N/A
```

### Log output and (if possible) backtrace

```shell
VERIFICATION FAILED: i < m_size at /home/cameron/ladybird/AK/Vector.h:164
Stack trace (most recent call first):
#0  (inlined)          in AK::Vector<Web::HTML::BitmapDecodedImageData::Frame, 0ul, (AK::FastLastAccess)0>::at(unsigned long) const at /home/cameron/ladybird/AK/Vector.h:164:9
#1  (inlined)          in AK::Vector<Web::HTML::BitmapDecodedImageData::Frame, 0ul, (AK::FastLastAccess)0>::operator[](unsigned long) const at /home/cameron/ladybird/AK/Vector.h:180:76
#2  0x00007fa0be0e785b in Web::HTML::BitmapDecodedImageData::frame_rect(unsigned long) const at /home/cameron/ladybird/Libraries/LibWeb/HTML/BitmapDecodedImageData.cpp:64:32
#3  0x00007fa0be58eb3d in Web::Painting::ImagePaintable::paint(Web::DisplayListRecordingContext&, Web::Painting::PaintPhase) const at /home/cameron/ladybird/Libraries/LibWeb/Painting/ImagePaintable.cpp:92:62
#4  0x00007fa0be5aed64 in paint_node at /home/cameron/ladybird/Libraries/LibWeb/Painting/StackingContext.cpp:35:20
#5  0x00007fa0be5af869 in operator()<Web::Painting::Paintable> at /home/cameron/ladybird/Libraries/LibWeb/Painting/StackingContext.cpp:193:23
[...]

Issue #7879, opened 2026-02-11
Crash when accessing Vector for AnimatedBitmapDecodedImageData::Frame


rcorsi:
Followup to #7752 

# Summary

VERIFICATION FAILED after some time, just follow the steps below and wait a few minutes.

# Operating system

Linux

# Steps to reproduce

1) Go to site neetprep.com (it's a test preparation site)
2) Click on the main menu named "Resources"
3) From the drop down menu, click on "Question Practice"
4) When the page with the questions appears, just do nothing and wait
5) It can be quick to occur, but also slow, so you may have to wait up to 15 minutes. Might happen faster in Debug build.

There will eventually be a VERIFICATION FAILED, which is due to the last ad banner at the bottom of the page, that is rotating between 3 jpgs and 1 gif.  Sometimes when the animated gif is updated to a non-animated jpeg the crash occurs.

If Ladybird is built in Debug, and the coredump is checked, the m_size value of the Vector is always 1, and the i value will be between 0 and 172.  The animated gif file has 173 frames.

Traced the issue to `HTMLImageElement::update_the_image_data_impl()` in `Libraries/LibWeb/HTML/HTMLImageElement.cpp:568`

Just before the crash I can see the img src going from:

`https://d18ni4vf61m6w3.cloudfront.net/key_2025-12-15%2019%3A17%3A38%20%2B0530_Combo%20Book.gif.jpeg`

to

`https://d18ni4vf61m6w3.cloudfront.net/key_2026-01-13%2017%3A27%3A29%20%2B0530_WhatsApp%20Image%202026-01-13%20at%205.10.40%20PM.jpeg.jpeg`

Happens every time the same way, but sometimes during that switch things work out fine.

You can see it with `--enable-idl-tracing` that the img src is being updated just before the crash. Image rotation is occurring every 5 seconds.

```
38375.991 WebContent(484121): ElementPrototype::set_attribute("id", "2463bannerCross")
38375.991 WebContent(484121): ElementPrototype::set_attribute("id", "2463-footer-bannerImage1")
38375.991 WebContent(484121): ElementPrototype::set_attribute("src", "https://d18ni4vf61m6w3.cloudfront.net/key_2025-12-15%2019%3A17%3A38%20%2B0530_Combo%20Book.gif.jpeg")
38380.994 WebContent(484121): ElementPrototype::set_attribute("id", "2661bannerCross")
[...]

PR #8021, opened 2026-02-18
LibWeb: Reset animated frame index when loading new img element src

rcorsi:
When an img element is updated from animated image to static image (by setting a new 'src'), the animation briefly continues into the new image, even if the new image has only a single frame.

Could also impact when going from an animated image to another animated image, but the new image has less frames versus the previously animated image and the current frame index is out of range for the new image.

In some cases a newly loaded static image would continue to be continuously animated even though it doesn't require it.  That is also stopped by the fix.

fixes: #7879
fixes: #7945

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLImageElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLImageElement.cpp
@@ -656,6 +656,7 @@ void HTMLImageElement::update_the_image_data_impl(bool restart_animations, bool 
             m_current_request = ImageRequest::create(realm(), document().page());
             m_current_request->set_image_data(entry->image_data);
             m_current_request->set_state(ImageRequest::State::CompletelyAvailable);
+            m_current_frame_index = 0;
 
             // 5. Prepare the current request for presentation given the img element.
             m_current_request->prepare_for_presentation(*this);
@@ -918,8 +919,9 @@ void HTMLImageElement::add_callbacks_to_image_request(GC::Ref<ImageRequest> imag
                 if (!maybe_omit_events || previous_url != url_string)
                     dispatch_event(DOM::Event::create(realm(), HTML::EventNames::load));
 
+                m_current_frame_index = 0;
+                m_animation_timer->stop();
                 if (image_data->is_animated() && image_data->frame_count() > 1) {
-                    m_current_frame_index = 0;
                     m_animation_timer->set_interval(image_data->frame_duration(0));
                     m_animation_timer->start();
                 }


rcorsi:
Attaching updated and improved reproduction zip file from #7879 

[repro.zip](https://github.com/user-attachments/files/25401114/repro.zip)

rcorsi:
Alternative solution


```diff
diff --git a/Libraries/LibWeb/HTML/BitmapDecodedImageData.cpp b/Libraries/LibWeb/HTML/BitmapDecodedImageData.cpp
index 79a1f16f97..06e242ad03 100644
--- a/Libraries/LibWeb/HTML/BitmapDecodedImageData.cpp
+++ b/Libraries/LibWeb/HTML/BitmapDecodedImageData.cpp
@@ -61,11 +61,15 @@ Optional<CSSPixelFraction> BitmapDecodedImageData::intrinsic_aspect_ratio() cons
 
 Optional<Gfx::IntRect> BitmapDecodedImageData::frame_rect(size_t frame_index) const
 {
+    if (frame_index >= m_frames.size())
+        frame_index = 0;
     return m_frames[frame_index].bitmap->rect();
 }
 
 void BitmapDecodedImageData::paint(DisplayListRecordingContext& context, size_t frame_index, Gfx::IntRect dst_rect, Gfx::IntRect clip_rect, Gfx::ScalingMode scaling_mode) const
 {
+    if (frame_index >= m_frames.size())
+        frame_index = 0;
     context.display_list_recorder().draw_scaled_immutable_bitmap(dst_rect, clip_rect, *m_frames[frame_index].bitmap, scaling_mode);
 }
```

rcorsi:
Tested with the original site from #7879 at `https://www.neetprep.com/questions` running on Ladybird with this fix for over an hour straight in four different tabs and had no crashes.

rcorsi:
I will look into creating a test case.

rcorsi:
@Psychpsyo I checked the fix on the site in #7945 and also the repro that you created, and it looks good. Any comments?

Psychpsyo:
> @Psychpsyo I checked the fix on the site in #7945 and also the repro that you created, and it looks good. Any comments?

a) I didn't make any repro.
b) If it doesn't crash anymore, that sounds good?
c) Someone on the issue said it only crashes when hovering the links in the top bar, so just running the site for a while wouldn't prove anything.

rcorsi:
@Psychpsyo your are right, it was @tcl3 that created the repro in the issue #7945 you reported.

So the repro in #7945 by @tcl3 and the site you reported works well with this fix.
Also the site and the repro in #7879, issue that I was working on, also are fine.  It is the site from issue #7879 that I left running for a hour with four tabs on the site `https://www.neetprep.com/questions`.  Without the fix, all the tabs would crash within a few minutes, as the ad banners are rotating between static and animated images, repeatedly.

Each repro and real site offers a slightly different view into the problems with this animated to static image issue.

Psychpsyo:
> So the repro in #7945 by @tcl3 and the site you reported works well with this fix. Also the site and the repro in #7879, issue that I was working on, also are fine. It is the site from issue #7879 that I left running for a hour with four tabs on the site `https://www.neetprep.com/questions`. Without the fix, all the tabs would crash within a few minutes, as the ad banners are rotating between static and animated images, repeatedly.
> 
> Each repro and real site offers a slightly different view into the problems with this animated to static image issue.

Interesting, thanks for the info!

PR #8022, opened 2026-02-18
Meta: Disallow using Homebrew's LLVM 21 on macOS for now

trflynn89:
We encounter linker errors with LLVM 21 that are proving a pain to overcome. The issues linked in the comments have quite a bit of chatter about this matter. Let's disallow it for now until we figure out a better solution or a fix is released.

### DIFF
--- a/Meta/find_compiler.py
+++ b/Meta/find_compiler.py
@@ -65,6 +65,16 @@ def major_compiler_version_if_supported(platform: Platform, compiler: str, clang
             return apple_build_version
 
     elif version.find("clang") != -1:
+        # LLVM 21 on macOS interacts poorly with the system libc++. We encounter linker errors referring to a function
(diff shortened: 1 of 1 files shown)


Meta/find_compiler.py
```diff
@@ -65,6 +65,16 @@ def major_compiler_version_if_supported(platform: Platform, compiler: str, clang
             return apple_build_version
 
     elif version.find("clang") != -1:
+        # LLVM 21 on macOS interacts poorly with the system libc. We encounter linker errors referring to a function
+        # that is present in LLVM's libc, but not in the system libc (std::__1::__hash_memory). We would need to set
```
nico: You mean libc++, not libc, right?

trflynn89: sure do

PR #8024, opened 2026-02-19
LibWeb: Force position:static on non-root SVG elements

kalenikaliaksandr:
SVG elements (except the outermost <svg>) use SVG's coordinate system, not the CSS box model, so CSS positioning doesn't apply to them.

This adds SVGElement::adjust_computed_style() to force position:static on all SVG elements except the outermost <svg> element (which has no owner_svg_element()). SVGSymbolElement's existing override now calls Base::adjust_computed_style() to inherit this behavior.

With this in place, the FIXME in layout_absolutely_positioned_element() for SVG boxes becomes unreachable and is replaced with VERIFY_NOT_REACHED().

### DIFF
--- a/Libraries/LibWeb/SVG/SVGElement.h
+++ b/Libraries/LibWeb/SVG/SVGElement.h
@@ -48,6 +48,7 @@ protected:
     virtual void children_changed(ChildrenChangedMetadata const*) override;
     virtual void inserted() override;
     virtual void removed_from(Node* old_parent, Node& old_root) override;
+    MUST_UPCALL virtual void adjust_computed_style(CSS::ComputedProperties&) override;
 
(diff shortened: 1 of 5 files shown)


Libraries/LibWeb/SVG/SVGElement.h
```diff
@@ -48,6 +48,7 @@ class WEB_API SVGElement
     virtual void children_changed(ChildrenChangedMetadata const*) override;
     virtual void inserted() override;
     virtual void removed_from(Node* old_parent, Node& old_root) override;
+    virtual void adjust_computed_style(CSS::ComputedProperties&) override;
```
gmta: Is there a potential of using MUST_UPCALL here?

PR #8026, opened 2026-02-19
LibWeb: Handle keyword style values in string_from_style_value()

kalenikaliaksandr:
Generic font families like `sans-serif` are represented as KeywordStyleValue, but string_from_style_value() only handled StringStyleValue and CustomIdentStyleValue, causing a crash when FontFaceSet.load() was called with a generic family name.


tcl3:
This looks like it's fixing the same crash as #8010

kalenikaliaksandr:
> This looks like it's fixing the same crash as #8010

yes, it's the same crash and 8010 looks like more reasonable solution than mine

PR #8027, opened 2026-02-19
LibWeb: Implement more Table border styles

AlvinKuruvilla:
image


Libraries/LibWeb/Painting/TableBordersPainting.cpp
```diff
@@ -286,13 +309,56 @@ static void paint_collected_edges(DisplayListRecordingContext& context, Vector<B
...
+                : dark_color_for_inset_and_outset(color);
+            auto half = max(DevicePixels(1), width / 2);
+            auto remainder = width - half;
+            if (border_edge_painting_info.direction == EdgeDirection::Horizontal) {
+                context.display_list_recorder().fill_rect({ border_rect.x(), border_rect.y(), border_rect.width(), half.value() }, outer_color);
+                context.display_list_recorder().fill_rect({ border_rect.x(), border_rect.y() + half.value(), border_rect.width(), remainder.value() }, inner_color);
+            } else {
+                context.display_list_recorder().fill_rect({ border_rect.x(), border_rect.y(), half.value(), border_rect.height() }, outer_color);
+                context.display_list_recorder().fill_rect({ border_rect.x() + half.value(), border_rect.y(), remainder.value(), border_rect.height() }, inner_color);
+            }
```
gmta: Instead of having all these conditional blocks, you should be able to use `border_rect.set_primary_size_for_orientation(orientation, ..)`

Libraries/LibWeb/Painting/TableBordersPainting.cpp
```diff
@@ -286,13 +309,56 @@ static void paint_collected_edges(DisplayListRecordingContext& context, Vector<B
             ? border_edge_painting_info.rect.top_right()
             : border_edge_painting_info.rect.bottom_left();
 
+        auto border_rect = Gfx::IntRect(border_edge_painting_info.rect.location(), border_edge_painting_info.rect.size());
```
gmta: You can do this instead:

```cpp
border_edge_painting_info.rect.to_type<int>()
```

gmta (commented):
Welcome! Could you:

* add some screenshots to your PR description? :)
* add a test for these new styles?

Libraries/LibWeb/Painting/TableBordersPainting.cpp
AlvinKuruvilla: Thx! Updated as suggested

Libraries/LibWeb/Painting/TableBordersPainting.cpp
AlvinKuruvilla: Thx! Updated as suggested

AlvinKuruvilla:
Hi @gmta thanks for the review. I believe I have correctly addressed your code review feedback and updated the PR's description to include the screenshot. Regarding your comment about the test, I wasn't entirely sure if I did the test set up correctly or put it in the right place for that matter, but I did add a test HTML file to Tests/LibWeb/Screenshot/expected/css-table-border-styles-ref.html. 
Tests/LibWeb/Screenshot/input/css-table-border-styles.html
This is also the screenshot seen in the PR description. 

Thanks again :)

PR #8031, opened 2026-02-19
LibWeb: Support ranges for `font-weight` in `@font-face`

AtkinsSJ:
Inside `@font-face`, a few descriptors can accept 2 values to define a range that the face should apply to. These are `font-weight`, `font-width/font-stretch`, and `font-style: oblique <angle>`. We currently ignore `font-width` for matching, and don't implement `oblique <angle>`, so this PR implements it for `font-weight` only.

WPT coverage isn't great - there are some ref tests for matching *all three* of these properties as ranges, but it doesn't make sense to import those with only one implemented. Even if we do happen to pass it, it'd be a bit dishonest. So I have a couple of hand-made tests instead.

### DIFF
--- a/Libraries/LibWeb/CSS/FontComputer.cpp
+++ b/Libraries/LibWeb/CSS/FontComputer.cpp
@@ -322,7 +322,7 @@ RefPtr<Gfx::FontCascadeList const> FontComputer::font_matching_algorithm(FlyStri
     quick_sort(matching_family_fonts, [](auto const& a, auto const& b) {
         return a.key.weight.min < b.key.weight.min;
     });
-    // FIXME: 1. font-stretch is tried first.
+    // FIXME: 1. font-width is tried first.
     // FIXME: 2. font-style is tried next.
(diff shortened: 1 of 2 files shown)


Tests/LibWeb/Ref/input/css/font-weight-range.html
```diff
@@ -0,0 +1,38 @@
...
+<link rel="match" href="../../expected/css/font-weight-range-ref.html" />
+<style>
+    @font-face {
+        font-family: 'Test';
+        src: url('../../data/fonts/Ahem.ttf');
+        font-weight: 400 600;
+    }
+    @font-face {
+        font-family: 'Test';
+        src: url('../../../../../Base/res/fonts/SerenitySans-Regular.ttf');
```
AtkinsSJ, line 12: Reaching out to grab SerenitySans like this feels iffy but it didn't seem worth duplicating the file for explicit use in tests. Ideally this would use `local('SerenitySans')` but we don't actually support `local()`. 😅

PR #8034, opened 2026-02-19
LibWeb: `font-variant` parsing improvements

Calme1709:
This PR includes a couple of changes, namely:
 - For each of `font-variant-east-asian`, `font-variant-ligatures`, and `font-variant-numeric` we now properly enforce the requirement that their components must not be disjoint within the `font-variant` shorthand (e.g. `jis78 full-width small-caps` is a valid value for `font-variant` but `jis78 small-caps full-width` is not since `jis78` and `full-width` do not appear consecutively)
 - Simplified parsing for the above mentioned properties
 - Moved the `LibGfx/FontVariant.h` structs to the `CSS` namespace since that was the only place they were used (and having them there allowed us to reuse existing enums)
 - Implemented parsing for the `font-variant-alternates` function values (e.g. `stylistic(foo)`) - we don't yet respect these values.


AK/ValueComparingRefPtr.h
```diff
@@ -8,7 +8,7 @@
 
 #include <AK/RefPtr.h>
 
-namespace Web::CSS {
```
AtkinsSJ, line 11: Oops. Thanks for catching that!

Libraries/LibWeb/CSS/StyleValues/TupleStyleValue.h
```diff
@@ -0,0 +1,40 @@
...
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#pragma once
+
+#include <LibWeb/CSS/StyleValues/StyleValue.h>
+
+namespace Web::CSS {
+
+class TupleStyleValue final : public StyleValueWithDefaultOperators<TupleStyleValue> {
```
AtkinsSJ, line 13: `TupleStyleValue` solves a problem that's been annoying for a long time, so thank you for that. (And it gives me ideas for solving some parsing issues I was stuck on.) However, I'm not that happy with it just being a Vector that's indexed into with magic numbers - that feels like too much of a footgun.

Presumably we don't want to go to the expense of using a `HashMap` here (though Jelle is working on some optimizations to small maps which would help...), but maybe you could make some constants or an enum for each set of indices for this, and use them both when reading and writing them.

Calme1709: I have added some constants for the indices here - that's much cleaner.

PR #8036, opened 2026-02-19
LibJS: Use exact integer arithmetic in Number.prototype.toPrecision

trflynn89:
Our previous implementation produced incorrect results for values near the limits of double precision. This new implementation avoid floating- point arithmetic entirely by:

1. Decomposing the double value into its exact binary form.
2. Computing the formulas from the spec using bigints.
3. Using Ryu to calculate the decimal exponent.

test262 diff:
```
test/built-ins/Temporal/Duration/prototype/total/relativeto-calendar-units-depend-on-relative-date.js ❌ -> ✅
test/staging/sm/Number/toPrecision-values.js                                                          ❌ -> ✅
```

We still don't pass `test/built-ins/Number/prototype/toPrecision/exponential.js`, but we make it further. We are failing to precisely store `new Number("0.000000000000000000001")`.

### DIFF
--- a/Libraries/LibJS/Runtime/NumberPrototype.cpp
+++ b/Libraries/LibJS/Runtime/NumberPrototype.cpp
@@ -7,9 +7,11 @@
  */
 
 #include <AK/Array.h>
+#include <AK/FloatingPoint.h>
 #include <AK/Function.h>
 #include <AK/StringFloatingPointConversions.h>
 #include <AK/TypeCasts.h>
+#include <LibCrypto/BigInt/UnsignedBigInteger.h>
 #include <LibJS/Runtime/AbstractOperations.h>
 #include <LibJS/Runtime/Completion.h>
 #include <LibJS/Runtime/Error.h>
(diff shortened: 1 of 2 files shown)


Libraries/LibJS/Runtime/NumberPrototype.cpp
```diff
@@ -39,6 +41,16 @@ static constexpr AK::Array<char, 36> digits = {
     'n', 'o', 'p', 'q', 'r', 's', 't', 'u', 'v', 'w', 'x', 'y', 'z'
 };
 
+static constexpr u8 count_digits(u64 number)
```
konradekk, line 44: i strongly beliebe that number 0 has 1 digit.

Libraries/LibJS/Runtime/NumberPrototype.cpp
```diff
@@ -278,6 +287,68 @@ JS_DEFINE_NATIVE_FUNCTION(NumberPrototype::to_locale_string)
...
+    // Below, we arrange this as a fraction, placing any negative values into the denominator to ensure that the math
+    // involves only unsigned integers.
+    Crypto::UnsignedBigInteger numerator { binary_significand };
+    Crypto::UnsignedBigInteger denominator = 1_bigint;
+
+    // 2 ^ binary_exponent
+    if (binary_exponent >= 0)
+        numerator = MUST(numerator.shift_left(binary_exponent));
+    else
+        denominator = MUST(denominator.shift_left(-binary_exponent));
```
konradekk, line 335: 2⁰ = 1?

PR #8037, opened 2026-02-19
Introduce ExternalContentSource for canvas and video painting

kalenikaliaksandr:
Display list commands currently hold direct references to mutable
  content (PaintingSurface for canvas, ImmutableBitmap for video).
  This causes two problems:

  1. Canvas thread safety: the rendering thread calls
     makeImageSnapshot() on the same PaintingSurface the main thread
     draws to — a data race.

  2. Video rebuild overhead: every new video frame triggers a full
     display list rebuild even though only the bitmap changed.

  ExternalContentSource is a thread-safe container that sits between
  producers and the display list. Producers swap in new ImmutableBitmap
  snapshots; the DrawExternalContent command reads the latest one at
  replay time.

  Canvas now snapshots via present(), so the rendering thread never
  touches the live GPU surface. Video publishes frames to
  ExternalContentSource and skips display list rebuilds for frame-only
  updates. The old DrawPaintingSurface command is removed.

### DIFF
--- a/Libraries/LibWeb/CSS/StyleValues/CursorStyleValue.cpp
+++ b/Libraries/LibWeb/CSS/StyleValues/CursorStyleValue.cpp
@@ -7,6 +7,7 @@
 #include "CursorStyleValue.h"
 #include <LibGfx/Bitmap.h>
 #include <LibGfx/Painter.h>
+#include <LibGfx/PaintingSurface.h>
 #include <LibWeb/CSS/Sizing.h>
 #include <LibWeb/CSS/StyleValues/AbstractImageStyleValue.h>
 #include <LibWeb/CSS/StyleValues/CalculatedStyleValue.h>
--- a/Libraries/LibWeb/HTML/RenderingThread.cpp
+++ b/Libraries/LibWeb/HTML/RenderingThread.cpp
@@ -5,6 +5,7 @@
  */
 
 #include <LibCore/EventLoop.h>
+#include <LibGfx/PaintingSurface.h>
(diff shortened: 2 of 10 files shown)


kalenikaliaksandr:
the side effect of this change is fixed weird flickering while video replay on mac

before:

https://github.com/user-attachments/assets/2aa86d5a-5b5d-4c7f-a4cb-bd17ca7092b3


after:

https://github.com/user-attachments/assets/2bc469a7-db49-4c2d-be62-aafd5d077f4e

Lubrsi:
I can confirm that I don't get the weird flickering with these changes. It seems MotionMark displays incomplete frames with these changes however:

Before (warning, flashing lights):

https://github.com/user-attachments/assets/80ae5bad-5807-4d1b-b9d3-e63c7904b7e7

After:

https://github.com/user-attachments/assets/215bc0f6-7bd7-410d-841f-8c8390951c56

kalenikaliaksandr:
> I can confirm that I don't get the weird flickering with these changes. It seems MotionMark displays incomplete frames with these changes however:
> 
> Before (warning, flashing lights):
> 
>  2026-02-20.00-55-56-compressed.mp4 
> After:
> 
>  2026-02-20.00-58-29.mp4

ah, thanks for catching this! the problem was that I made it to snapshot canvas's surface from `CanvasRenderingContext2D::did_draw()`, which ofc might result in incomplete frames being displayed. I redid it to use canvas dirty marking in a similar way we invalidate media elements.

PR #8041, opened 2026-02-20
LibMedia: Avoid strong sink captures in async seek callbacks

jonbgamble:
Seek completion callbacks no longer keep display/audio sinks alive through strong references. Instead, they resolve sinks through a weak manager reference at callback time and return early if playback is already gone. This should reduce undefined teardown behavior during navigation.

### DIFF
--- a/Libraries/LibMedia/PlaybackStates/PlaybackStateHandler.cpp
+++ b/Libraries/LibMedia/PlaybackStates/PlaybackStateHandler.cpp
@@ -22,8 +22,15 @@ void PlaybackStateHandler::on_track_enabled(Track const& track)
         auto& track_data = manager().get_video_data_for_track(track);
         VERIFY(track_data.display != nullptr);
         track_data.display->pause_updates();
-        track_data.provider->seek(manager().current_time(), SeekMode::Accurate, [display = NonnullRefPtr(*track_data.display)](AK::Duration) {
-            display->resume_updates();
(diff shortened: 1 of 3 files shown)


Zaggy1024:
Sorry it took me so long to look into this properly. I'm wondering what the point of this change is, though. What problem would occur due to having a reference to the display or audio sink for a little longer after the playback manager is destroyed? If there are no side effects to this, I'd argue that the code looks cleaner as it is, but if we can observe a side effect other than "ref count is greater than 1", then this seems reasonable.

(If this is actually fixing a bug, it will need a rebase, since PlaybackManager is no longer ref-counted.)

jonbgamble:
This patch is not about refcount neatness. It is about stopping late async seek completions from acting on audio sinks that have already been destroyed.

If the current PlaybackManager does not crash when this happens in the current master, then maybe we don't need this change and we can close this PR.

Zaggy1024:
Was that happening in the commit this branch was based on? How could a strong reference to an atomic ref-counted class possibly UAF?

jonbgamble:
Nope, of course it was one of my franken-branches.

To be honest it was some time ago, and my weary eyes have seen many a crash since then. I cannot tell you with absolute certainty that my sink implementation at the time wasn't `class AudioMixingSink { AudioMixingSink::AudioMixingSink() { VERIFY(false); } };`, so I have no strong feelings about this one. If the code is cleaner without it, let's just close it.

PR #8042, opened 2026-02-20
LibMedia: Serialize AudioConverter access in AudioDataProvider

jonbgamble:
Sample spec updates can be triggered by output device changes while audio blocks are being converted. Guard AudioConverter with a mutex so set_output_sample_specification() and convert() don't step on each other

### DIFF
--- a/Libraries/LibMedia/Providers/AudioDataProvider.cpp
+++ b/Libraries/LibMedia/Providers/AudioDataProvider.cpp
@@ -110,6 +110,7 @@ void AudioDataProvider::ThreadData::set_block_end_time_handler(BlockEndTimeHandl
 
 void AudioDataProvider::ThreadData::set_output_sample_specification(Audio::SampleSpecification sample_specification)
 {
+    Threading::MutexLocker locker { m_converter_mutex };
     m_converter->set_output_sample_specification(sample_specification).release_value_but_fixme_should_propagate_errors();
 }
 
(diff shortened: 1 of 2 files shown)


Zaggy1024:
This feels like it's a prelude to another change. I don't believe we have any concurrency issues with the converter at the moment, given that AudioDataProvider doesn't start processing anything until after the converter has been set.

I wanna say that there should be a better solution for this than adding another mutex, but it may require a refactor of AudioDataProvider. One thing I was thinking about was making AudioDataProvider use a task queue, so that outside modifications can all be processed serially by the thread, but we would then need a special behavior for the seeking to be interruptible still.

jonbgamble:
This is "yank your USB headphones (or drop bluetooth) and fall back to a different sample rate default device" during a conversion. I'm not sure how master would fare in this situation, but webaudio-worky branch propagates device changes from the OS to LibMedia as they happen.

I thought it made sense purely on inspection so I PRed it over here. I don't feel strongly either way, I'd just as soon close it as spend time refactoring for soemthing that does not happen here.

Zaggy1024:
That's my feeling as well, I think this should be a commit leading up to the changes that depend on it.

Regarding device changes, we have nothing subscribing to events for that at the moment, so depending on platform, that may either result in nothing happening, dropped audio, or a crash. 😅 It just hasn't been a focus for me yet.

Definitely worth having something to handle that in the future, but I think the easiest solution is just to move setting of the output sample specification into the decoding thread. Access to the desired specification would probably still need to be gated behind the usual mutex, though. That mutex is at least _intended_ to be held for very brief periods when copying data across threads anyway, so hopefully contention won't be much of an issue.

PR #8044, opened 2026-02-20
LibWeb: Propagate containing block width into throwaway TFC state

kalenikaliaksandr:
The throwaway LayoutState used by
compute_table_box_width_inside_table_wrapper() was empty, so TFC's compute_table_width() saw content_width=0 and has_definite_width=false for the table wrapper's containing block. This caused percentage widths (e.g. width: 100%) on the table element to be treated as auto, making the table shrink to min-content width instead of resolving the percentage.

Fix by copying the containing block's content width from the main state into the throwaway state. This also fixes a centering bug for tables with percentage widths inside <center> elements (off by a few pixels).

### DIFF
--- a/Libraries/LibWeb/Layout/FormattingContext.cpp
+++ b/Libraries/LibWeb/Layout/FormattingContext.cpp
@@ -455,6 +455,13 @@ CSSPixels FormattingContext::compute_table_box_width_inside_table_wrapper(Box co
 
     LayoutState throwaway_state;
 
+    // Propagate the containing block's content width into the throwaway state
+    // so that TFC can correctly resolve percentage widths on the table element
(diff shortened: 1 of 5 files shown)


tcl3:
This looks very similar to #8032

kalenikaliaksandr:
> This looks very similar to #8032

indeed, do you want to include Tests/LibWeb/Ref/input/table-margin-auto-inside-flex-item.html in yours and close mine? 😄

tcl3:
> indeed, do you want to include Tests/LibWeb/Ref/input/table-margin-auto-inside-flex-item.html in yours and close mine? 😄

Sure, I've added the test from this PR to mine.

PR #8050, opened 2026-02-20
LibWeb: Respect `@font-feature-values` rules

Calme1709:
A caveat in that we don't support `@font-feature-values` rules that appear in grouping rules (e.g. `@​media`, `@​supports`) which is also an issue for `@font-face` rules. Have added FIXMEs.

### DIFF
--- a/Libraries/LibWeb/CSS/FontFeatureData.cpp
+++ b/Libraries/LibWeb/CSS/FontFeatureData.cpp
@@ -10,7 +10,7 @@
 
 namespace Web::CSS {
 
-Gfx::ShapeFeatures FontFeatureData::to_shape_features(HashMap<FontFeatureValueKey, Vector<u32>> const&) const
+Gfx::ShapeFeatures FontFeatureData::to_shape_features(HashMap<FontFeatureValueKey, Vector<u32>> const& font_feature_values) const
 {
     HashMap<FlyString, u8> merged_features;
 
@@ -185,6 +185,7 @@ Gfx::ShapeFeatures FontFeatureData::to_shape_features(HashMap<FontFeatureValueKe
         }
 
         // 6.8 https://drafts.csswg.org/css-fonts/#font-variant-alternates-prop
+        // FIXME: These values never apply to generic font families
         if (font_variant_alternates.has_value()) {
             auto alternates = font_variant_alternates.value();
             if (alternates.historical_forms) {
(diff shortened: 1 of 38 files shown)


Calme1709:
Rebased to resolve conflicts

Libraries/LibWeb/CSS/CSSFontFeatureValuesRule.cpp
```diff
@@ -101,6 +104,23 @@ String CSSFontFeatureValuesRule::serialized() const
     return builder.to_string_without_validation();
 }
 
+void CSSFontFeatureValuesRule::clear_dependent_caches()
```
AtkinsSJ: `CSSRule` does have a virtual `clear_caches()` method, called mostly when the rule is moved around. If you hook this into that, then you don't need the changes in CSSRuleList.

Libraries/LibWeb/CSS/CSSRuleList.cpp
```diff
@@ -148,6 +149,11 @@ WebIDL::ExceptionOr<unsigned> CSSRuleList::insert_a_css_rule(Variant<StringView,
     // 8. Insert new rule into list at the zero-indexed position index.
     m_rules.insert(index, *new_rule);
 
+    // FIXME: Load font faces for inserted @font-face rules
```
AtkinsSJ, line 152: I think this wants to happen in `CSSFontFaceRule::set_parent_style_sheet()`.

Libraries/LibWeb/CSS/FontComputer.cpp
```diff
@@ -361,36 +361,66 @@ RefPtr<Gfx::FontCascadeList const> FontComputer::font_matching_algorithm(FlyStri
...
+    return m_font_feature_values_cache.ensure(family_name, [&]() {
+        HashMap<FontFeatureValueKey, Vector<u32>> font_feature_values;
+
+        // https://drafts.csswg.org/css-fonts/#font-feature-values-syntax
+        // For each <family-name> in the @font-feature-values prelude, each font feature value declaration defines a
+        // mapping between a (family name, feature block name, declaration name) tuple and the list of one or more
+        // integers from the declaration’s value. If the same tuple appears more than once in a document (such as if a
+        // single block), the last-defined one is used.
+
+        // FIXME: We only account for Author stylesheets here, should we also account for User and UserAgent?
```
AtkinsSJ: For completeness, yes. But we're unlikely to actually use it anywhere in the UA styles, and as far as I know we don't allow anyone to set the User style sheet.

Libraries/LibWeb/CSS/CSSFontFeatureValuesRule.cpp
Calme1709: Good point - I have updated this to override the `clear_caches()` method. I also realised that we need to clear this cache both before and after updating the stylesheet so I have updated the `clear_caches()` callers to do so

Libraries/LibWeb/CSS/CSSRuleList.cpp
Calme1709: I'm not sure that's correct - just because a rule has a parent style sheet doesn't mean it's effective (e.g. if it's within a conditional rule that doesn't match). It would be very nice if we had a way to track when rules became effective.

Libraries/LibWeb/CSS/FontComputer.cpp
Calme1709: I have updated the FIXME to reflect that.

Calme1709:
Rebased and added the imported tests under `LoadFromHttpServer` in `TestConfig.ini` for compatibility with 1be6947

PR #7708, opened 2026-02-01
LibWeb: Parse and serialize `@counter-style` rule

Calme1709:
We don't actually respect this rule yet - that will come in a later PR.

See individual commits for details.

PR #8052, opened 2026-02-21
LibWeb: Respect `@counter-style` rules

Calme1709:
This builds on #7708 which added support for parsing of `@counter-style` rules and actually respects them when resolving `<counter>` functions and list item markers.

Unfortunately this only brings with it a couple of WPT passes - this is due to various deficiencies in our layout/painting of list item markers (e.g. text layout, lack of support for fallback fonts, etc) which are already covered by various FIXMEs.

### DIFF
--- a/Libraries/LibWeb/CSS/ComputedProperties.cpp
+++ b/Libraries/LibWeb/CSS/ComputedProperties.cpp
@@ -1343,7 +1343,7 @@ TextTransform ComputedProperties::text_transform() const
     return keyword_to_text_transform(value.to_keyword()).release_value();
 }
 
-ListStyleType ComputedProperties::list_style_type() const
+ListStyleType ComputedProperties::list_style_type(HashMap<FlyString, CounterStyle> const& registered_counter_styles) const
 {
     auto const& value = property(PropertyID::ListStyleType);
 
@@ -1353,11 +1353,7 @@ ListStyleType ComputedProperties::list_style_type() const
     if (value.is_string())
         return value.as_string().string_value().to_string();
 
-    if (auto keyword = value.as_counter_style().to_counter_style_name_keyword(); keyword.has_value())
-        return keyword.release_value();
-
-    // FIXME: Support user defined counter styles.
-    return Empty {};
+    return value.as_counter_style().resolve_counter_style(registered_counter_styles);
 }
 
 ListStylePosition ComputedProperties::list_style_position() const
--- a/Libraries/LibWeb/CSS/CounterStyle.cpp
+++ b/Libraries/LibWeb/CSS/CounterStyle.cpp
@@ -24,6 +24,20 @@ CounterStyle CounterStyle::decimal()
         CounterStylePad { .minimum_length = 0, .symbol = ""_fly_string });
 }
 
+// https://drafts.csswg.org/css-counter-styles-3/#disc
+CounterStyle CounterStyle::disc()
+{
+    return CounterStyle::create(
+        "disc"_fly_string,
+        GenericCounterStyleAlgorithm { CounterStyleSystem::Cyclic, { "•"_fly_string } },
+        CounterStyleNegativeSign { .prefix = ""_fly_string, .suffix = " "_fly_string },
+        ""_fly_string,
+        " "_fly_string,
+        { { NumericLimits<i64>::min(), NumericLimits<i64>::max() } },
+        "decimal"_fly_string,
+        CounterStylePad { .minimum_length = 0, .symbol = ""_fly_string });
+}
+
 CounterStyle CounterStyle::from_counter_style_definition(CounterStyleDefinition const& definition, HashMap<FlyString, CounterStyle> const& registered_counter_styles)
 {
     return definition.algorithm().visit(
--- a/Libraries/LibWeb/Layout/ListItemMarkerBox.cpp
+++ b/Libraries/LibWeb/Layout/ListItemMarkerBox.cpp
@@ -5,6 +5,7 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
+#include <LibWeb/DOM/Document.h>
 #include <LibWeb/DOM/Element.h>
 #include <LibWeb/Layout/ListItemMarkerBox.h>
 #include <LibWeb/Painting/MarkerPaintable.h>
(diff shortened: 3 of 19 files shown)


Libraries/LibWeb/CSS/CounterStyleDefinition.cpp
```diff
@@ -0,0 +1,240 @@
...
+        return {};
+
+    auto maybe_algorithm = resolve_algorithm(*rule.system_style_value(), rule.symbols_style_value(), rule.additive_symbols_style_value(), computation_context);
+
+    if (maybe_algorithm.has<Empty>())
+        return {};
+
+    auto algorithm = maybe_algorithm.visit(
+        [](Empty const&) -> Variant<CounterStyleAlgorithm, CounterStyleSystemStyleValue::Extends> { VERIFY_NOT_REACHED(); },
+        [](auto const& algorithm) -> Variant<CounterStyleAlgorithm, CounterStyleSystemStyleValue::Extends> { return algorithm; });
```
AtkinsSJ: I think you can use `Variant::downcast()` for this, but it's not something I've ever actually used.

Libraries/LibWeb/CSS/CounterStyleDefinition.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+    return *new CounterStyleDefinition(
```
AtkinsSJ:
```suggestion
    return CounterStyleDefinition(
```
We don't want this in pointer form so creating it that way just causes a copy I believe?

Libraries/LibWeb/CSS/CounterStyleDefinition.cpp
```diff
@@ -0,0 +1,240 @@
...
+                }
+            };
+        },
+        [&](CounterStyleSystemStyleValue::Extends const& extends) -> Variant<Empty, CounterStyleAlgorithm, CounterStyleSystemStyleValue::Extends> {
+            return extends;
+        });
+}
+
+// https://drafts.csswg.org/css-counter-styles-3/#descdef-counter-style-symbols
+Vector<FlyString> CounterStyleDefinition::resolve_symbols(NonnullRefPtr<StyleValue const> const& symbols_style_value)
```
AtkinsSJ: Symbols can also be images, not just text. https://drafts.csswg.org/css-counter-styles-3/#typedef-symbol

A fixme would be fine, but maybe this could return `Vector<CounterSymbol>`, where that's currently just a `using` for `FlyString`?

Libraries/LibWeb/CSS/CounterStyle.cpp
```diff
@@ -45,11 +45,59 @@ CounterStyle CounterStyle::from_counter_style_definition(CounterStyleDefinition
...
+                // 2. If weight is zero, or weight is greater than value, continue.
+                if (tuple.weight == 0 || tuple.weight > value)
+                    continue;
+
+                // 3. Let reps be floor( value / weight ).
+                auto reps = value / tuple.weight;
+
+                // 4. Append symbol to S reps times.
+                for (int i = 0; i < reps; ++i)
+                    builder.append(tuple.symbol);
```
AtkinsSJ, line 113: Ah... my earlier comment about images isn't actually supported by the spec algorithms then. 😔 Yeah just leave a FIXME in that earlier code.

Libraries/LibWeb/CSS/CounterStyle.cpp
```diff
@@ -111,7 +111,34 @@ Optional<String> CounterStyle::generate_an_initial_representation_for_the_counte
...
+                // 1. If value is 0, append symbol(0) to S and return S.
+                if (value == 0)
+                    return generic_algorithm.symbol_list[0].to_string();
+
+                StringBuilder builder;
+
+                // 2. While value is not equal to 0:
+                while (value != 0) {
+                    // 1. Prepend symbol( value mod N ) to S.
+                    // NB: Our string builder doesn't support prepending, so we append and reverse at the end.
```
AtkinsSJ: This only works if the symbols are single-character strings, which as far as I can tell isn't guaranteed.

Libraries/LibWeb/CSS/CounterStyle.cpp
```diff
@@ -139,7 +139,33 @@ Optional<String> CounterStyle::generate_an_initial_representation_for_the_counte
...
+
+                StringBuilder builder;
+
+                // While value is not equal to 0:
+                while (value != 0) {
+                    // 1. Set value to value - 1.
+                    value -= 1;
+
+                    // 2. Prepend symbol( value mod N ) to S.
+                    // NB: Our string builder doesn't support prepending, so we append and reverse at the end.
```
AtkinsSJ: Same point about single-character symbols.

Libraries/LibWeb/CSS/CounterStyle.cpp
```diff
@@ -203,4 +217,118 @@ Optional<String> CounterStyle::generate_an_initial_representation_for_the_counte
...
+        return generate_a_counter_representation_using_fallback();
+
+    auto representation = maybe_representation.value();
+
+    // 4. Prepend symbols to the representation as specified in the pad descriptor.
+    {
+        // https://drafts.csswg.org/css-counter-styles-3/#counter-style-pad
+        // Let difference be the provided <integer> minus the number of grapheme clusters in the initial representation
+        // for the counter value.
+        // FIXME: We should be counting grapheme clusters here.
```
AtkinsSJ, line 321: cc: @trflynn89, do we have a nice way of doing this? All I can think of is that you could use a grapheme Segmenter and count how many boundaries it has, but that seems a bit bulky.

Libraries/LibWeb/CSS/CounterStyle.cpp
```diff
@@ -203,4 +217,118 @@ Optional<String> CounterStyle::generate_an_initial_representation_for_the_counte
...
+        auto difference = counter_style->pad().minimum_length - static_cast<i32>(representation.length_in_code_units());
+
+        // If the counter value is negative and the counter style uses a negative sign, further reduce difference by
+        // the number of grapheme clusters in the counter style’s negative descriptor’s <symbol>(s).
+        // FIXME: We should be counting grapheme clusters here.
+        if (value_is_negative_and_uses_negative_sign)
+            difference -= counter_style->negative_sign().prefix.bytes_as_string_view().length() + counter_style->negative_sign().suffix.bytes_as_string_view().length();
```
AtkinsSJ: Above we're using the code-unit length and here we're using byte lengths. Even if we can't use graphemes, let's stick to code-units.

Libraries/LibWeb/CSS/ComputedProperties.cpp
```diff
@@ -1353,11 +1353,9 @@ ListStyleType ComputedProperties::list_style_type() const
     if (value.is_string())
         return value.as_string().string_value().to_string();
 
-    if (auto keyword = value.as_counter_style().to_counter_style_name_keyword(); keyword.has_value())
-        return keyword.release_value();
+    return value.as_counter_style().resolve_counter_style(registered_counter_styles);
 
-    // FIXME: Support user defined counter styles.
-    return Empty {};
+    VERIFY_NOT_REACHED();
```
AtkinsSJ: Nit: This is after a `return` so it's not needed.

Libraries/LibWeb/Layout/ListItemMarkerBox.cpp
```diff
@@ -31,41 +47,19 @@ Optional<String> ListItemMarkerBox::text() const
...
-                text = String::roman_number_from(index, String::Case::Upper);
-                break;
-            }
-            return MUST(String::formatted("{}. ", text));
+
+            auto counter_representation = CSS::generate_a_counter_representation(counter_style, m_list_item_element->document().registered_counter_styles(), index);
+
+            // AD-HOC: The spec implies that the prefix and suffix used should be that of the specified counter style,
+            //         but doesn't say what to do if our specified counter style is not defined. Other browsers use the
+            //         prefix and suffix of the decimal counter style (i.e. "" and "."), so we do the same.
```
AtkinsSJ: This sounds worth raising as a spec issue.

Libraries/LibWeb/CSS/CounterStyleDefinition.cpp
Calme1709: Yes we can - very nice.

Libraries/LibWeb/CSS/CounterStyleDefinition.cpp
Calme1709: FIxed.

Libraries/LibWeb/CSS/CounterStyleDefinition.cpp
Calme1709: See the following note slightly down from your link:
>The `<image>` syntax in `<symbol>` is currently at-risk. No implementations have plans to implement it currently, and it complicates some usages of counter() in ways that haven’t been fully handled.

I have added a `CounterStyleSymbol` type along with that note and a FIXME.

Libraries/LibWeb/CSS/ComputedProperties.cpp
Calme1709: Removed.

Libraries/LibWeb/CSS/CounterStyle.cpp
Calme1709: Good point - have updated this to use a `Vector` instead of a `StringBuilder` (since `Vector`s support prepending)

Libraries/LibWeb/CSS/CounterStyle.cpp
Calme1709: Fixed - as above.

Libraries/LibWeb/CSS/CounterStyle.cpp
Calme1709: Fixed.

Libraries/LibWeb/Layout/ListItemMarkerBox.cpp
Calme1709: Looking into this a bit further it is actually covered in [css-lists-3](https://drafts.csswg.org/css-lists-3/#text-markers) - I have updated the comments here.

Calme1709 (commented):
Changes:
 - Applied feedback
 - Rebased to resolve conflicts

PR #8056, opened 2026-02-21
LibWeb: Use UA shadow DOM for media elements' controls

Zaggy1024:
Instead of using a custom paintable to draw the controls for video and audio elements, we build them out of plain old HTML elements within a shadow root.

This required a few hacks in the previous commits in order to allow a replaced element to host children within a shadow root, but it's fairly self-contained.

A big benefit is that we can drive all the UI updates off of plain old DOM events (except the play button overlay on videos, which uses the video element representation), so we can test our media and input event handling more thoroughly. :^)

The control bar visibility is now more similar to how other browsers handle it. It will show upon hovering over the element, but if the cursor is kept still for more than a second, it will hide again. While dragging, the controls remain visible, and will then hide after the mouse button is released.

The icons have been redesigned from scratch, and the mute icon now visualizes the volume level along with indicating the mute state.

Demo:

[media-controls.webm](https://github.com/user-attachments/assets/95ff2ab3-25e9-4692-8e40-b82a8369b1f1)

CC @kalenikaliaksandr especially for the layout hacks.

### DIFF
--- a/Libraries/LibWeb/Page/EventHandler.cpp
+++ b/Libraries/LibWeb/Page/EventHandler.cpp
@@ -737,8 +737,19 @@ EventResult EventHandler::handle_mouseup(CSSPixelPoint visual_viewport_position,
                         }
                     }
                 } else if (button == UIEvents::MouseButton::Secondary) {
-                    if (is<HTML::HTMLImageElement>(*node)) {
-                        auto& image_element = as<HTML::HTMLImageElement>(*node);
+                    // Skip up the tree to the first ancestor that is not a UA shadow DOM node, and use its context menu.
+                    // Media elements' controls' shadow DOM nodes should not have their own context menu, but rather activate
+                    // their parent media element's menu.
+                    auto context_menu_node = node.as_nonnull();
+                    while (auto shadow_root = context_menu_node->containing_shadow_root()) {
+                        if (!shadow_root->is_user_agent_internal())
+                            break;
+                        VERIFY(shadow_root->host() != nullptr);
+                        context_menu_node = *shadow_root->host();
+                    }
+
+                    if (is<HTML::HTMLImageElement>(*context_menu_node)) {
+                        auto& image_element = as<HTML::HTMLImageElement>(*context_menu_node);
                         auto image_url = image_element.document().encoding_parse_url(image_element.current_src());
                         if (image_url.has_value()) {
                             Optional<Gfx::Bitmap const*> bitmap;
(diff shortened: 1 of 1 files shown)


konradekk:
a comment to visuals, namely the volume handling: i've observed that in many player controls muting empties the volume bar (but does not change it internally), pressing unmute just brings it to previous level (no volume change), and setting the volume while muted unmutes on that specific new level (without need for additional unmute button press).

i find such feel/ux quite intuitive, but that's a matter of taste (or os alignment?).

Zaggy1024:
Agreed, it feels nicer that way, I've made the change.

Libraries/LibWeb/HTML/MediaControls.cpp
```diff
@@ -0,0 +1,715 @@
...
+        MUST(m_mute_button->class_list()->add(icon_class(new_volume_icon_state)));
+        m_mute_icon_state = new_volume_icon_state;
+    }
+
+    static Vector<String> s_muted_class = { "muted"_string };
+    if (muted != m_was_muted) {
+        if (muted)
+            MUST(m_mute_button->class_list()->add(s_muted_class));
+        else
+            MUST(m_mute_button->class_list()->remove(s_muted_class));
```
konradekk: note: you could also use
```suggestion
        MUST(m_mute_button->class_list()->toggle(s_muted_class, muted));
```
(same in other similar places)

Libraries/LibWeb/HTML/MediaControls.cpp
```diff
@@ -0,0 +1,715 @@
...
+
+        case MuteIconState::Empty:
+            return s_no_volume_class;
+        case MuteIconState::Low:
+            return s_low_volume_class;
+        case MuteIconState::High:
+            return s_high_volume_class;
+        }
+        VERIFY_NOT_REACHED();
+    };
```
konradekk, line 640: not sure if using the enum is worth it here – using an `if`-chain would probably be enough for lack of pattern matching?

just looking at other class assignments here, matter of taste (i might have also missed other usages; ignore this if that's the case!).

Zaggy1024: The usecase is to track the previously assigned class to avoid unnecessary DOM mutations. An enum saves us an extra field.

If you mean using a switch instead of an if chain, switch is generally nicer because we get warnings when values are unchecked.

Libraries/LibWeb/HTML/MediaControls.cpp
Zaggy1024: Ah, I had seen that method, but the way it used the second parameter was so confusing I dismissed it 😅 that looks much cleaner.

PR #8061, opened 2026-02-21
LibWeb: Reject invalid alt text values in CSS content property parsing

adamjcolvin:
## Summary

The CSS `content` property's alt text portion (after the `/` separator) was incorrectly accepting any content value type, including URLs, quote keywords, and images. Per the [CSS Content Module Level 3 spec](https://drafts.csswg.org/css-content-3/#content-property), alt text only accepts `<string>`, `<counter>`, and `<attr()>` values. This change adds type validation in the alt text parsing branch to reject value types that are not part of the alt text grammar.

## Implementation Approach

The fix is a minimal type check in `parse_content_value` within the alt text branch. After parsing a value using `parse_css_value_for_property`, the code now calls `is_string()` and `is_counter()` on the result and rejects anything that is neither. All other value types are now rejected by returning `nullptr`.

```cpp
if (in_alt_text) {
    // https://drafts.csswg.org/css-content-3/#content-property
    // / [ <string> | <counter> | <attr()> ]+
    // NB: <attr()> is handled as an arbitrary substitution function and does not reach this code path.
    if (!style_value->is_string() && !style_value->is_counter())
        return nullptr;
    alt_text_values.append(style_value.release_nonnull());
}
```

The `<attr()>` case doesn't need explicit handling here because `attr()` is treated as a substitution function (like `var()` and `env()`) and creates an `UnresolvedStyleValue` before `parse_content_value` is called.

## Testing

This fixes 64 of 66 failing subtests in the `css/css-content/parsing/content-invalid.html` WPT test (from 4 pass / 66 fail to 68 pass / 2 fail). 

## Changes

- `Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp` — Added type validation in the alt text branch of `parse_content_value` to reject non-string, non-counter values
- `Tests/LibWeb/Text/expected/wpt-import/css/css-content/parsing/content-invalid.txt` — Updated expected results to reflect 64 newly passing subtests


Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
```diff
@@ -2242,6 +2242,11 @@ RefPtr<StyleValue const> Parser::parse_content_value(TokenStream<ComponentValue>
                 return nullptr;
 
             if (in_alt_text) {
+                // https://drafts.csswg.org/css-content-3/#content-property
+                // / [ <string> | <counter> | <attr()> ]+
+                // NOTE: <attr()> is handled as an arbitrary substitution function and does not reach this code path.
```
trflynn89: We've settled on `NB:` as comment prefixes, to distinguish our comments from spec comments that also begin with `NOTE:`.

See: https://github.com/LadybirdBrowser/ladybird/blob/master/Documentation/CodingStyle.md#spec-notes

adamjcolvin: Ah thanks, I've updated the commit.

PR #8062, opened 2026-02-21
LibWeb: Fix input field placeholder vertical alignment

punthenk:
Input elements with a placeholder were positioned lower than inputs without a placeholder or inputs with text. This was caused by the placeholder element missing height height and overflow properties that the inner text element had.

This fix makes sure both the placeholder and inner text elements let show the same position of the input field. I also have tested this in firefox and chrome and now I get the same behaviour in ladybird.

## This was before:
Screenshot 2026-02-21 at 14 45 09

Screenshot 2026-02-21 at 14 45 21

---
## This is after:
Screenshot 2026-02-21 at 14 46 06

Screenshot 2026-02-21 at 14 46 14

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLInputElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLInputElement.cpp
@@ -8,6 +8,7 @@
  * Copyright (c) 2024, Fernando Kiotheka <fer@k6a.dev>
  * Copyright (c) 2025, Felipe Muñoz Mazur <felipe.munoz.mazur@protonmail.com>
  * Copyright (c) 2025, Glenn Skrzypczak <glenn.skrzypczak@gmail.com>
+ * Copyright (c) 2026, Michiel Nijenhuis <michielmitsjol@gmail.com>
  *
  * SPDX-License-Identifier: BSD-2-Clause
  */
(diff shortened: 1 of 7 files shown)


Tests/LibWeb/Layout/input/input-element-placeholder-position.html
```diff
@@ -0,0 +1,17 @@
...
+        }
+    </style>
+</head>
+<body>
+    <h1>Input Field Placeholder Position Test</h1>
+    <input type="text" placeholder="Empty field">
+    <input type="text" value="Filled field">
+    <input type="text" placeholder="Type here">
+</body>
+</html>
```
gmta, line 17: There's no expectation file for this new test.

Libraries/LibWeb/HTML/HTMLInputElement.cpp
```diff
@@ -862,7 +862,10 @@ static GC::Ref<CSS::CSSStyleProperties> placeholder_style_when_visible()
         style = CSS::CSSStyleProperties::create(internal_css_realm(), {}, {});
         style->set_declarations_from_text(R"~~~(
                 width: 100%;
+                height: 1lh;
                 align-items: center;
+                overflow: hidden;
+                scrollbar-width: auto;
```
gmta: Couldn't this be `scrollbar-width: none;`? Not sure when we'd want a scrollbar to show up.

punthenk:
@gmta
Thank you! I have updated the `scrollbar-width` to `none`. And I added an expectation for the added test with rebaseline. It also updated some other expectations with it. 

Locally all test pass now except Wasm. All LibWeb test passed.

punthenk:
@gmta
Hello. I have rebased all the new commits into my branch and did the test locally again. This time 100% of the tests passed.

punthenk:
@gmta Thank you! I see there is still 1 failing test, as far as I can tell it has nothing to do with my change. What would you suggest?

AtkinsSJ:
We have a few flaky tests unfortunately. I've restarted CI, it should pass and this will merge.

punthenk:
Hello @gmta and @AtkinsSJ, all tests are passing now. Is there anything else needed for this to be merged? Thanks!

PR #8063, opened 2026-02-21
LibJS: Use exact integer arithmetic in `Number.prototype.toExponential`

trflynn89:
The arithmetic here nearly exactly matches that of `toPrecision` from commit cf180bd4da6d2fd10812b9757e6dff4f18d0d158.

The only difference is test262 contains tests for which our exponent estimate is off-by-1. We now handle this by detecting this inaccuracy. adjusting the exponent, and recomputing the significand.

test262 diff:
```
test/built-ins/Number/prototype/toExponential/return-values.js ❌ -> ✅
test/built-ins/Number/prototype/toPrecision/exponential.js     ❌ -> ✅
test/staging/sm/Number/toExponential-values.js                 ❌ -> ✅
````

We now pass all `Number` tests:
```
test/built-ins/Number   338/338   (100.00%) [ ✅ 338 ]
```


trflynn89:
Edit: Figured out the remaining tie-break bug we had in `toPrecision` that was causing one test262 test to fail. Pushed a commit to fix that.

PR #8064, opened 2026-02-21
LibWeb: Visit ImportMapParseResult in HTMLScriptElement::visit_edges

Praise-Garfield:
`HTMLScriptElement::visit_edges` traced the `GC::Ref<Script>` arm of `m_result` but not the `GC::Ref<ImportMapParseResult>` arm. When a `<script type="importmap">` element was processed, the `ImportMapParseResult` GC cell had no tracing path and could be collected while `m_result` still held a reference to it.

`ImportMapParseResult` inherited from `JS::Script::HostDefined` and implemented `visit_host_defined_self`, but no `JS::Script` or `JS::Module` ever stored it as `host_defined` data, so that path was dead code. This removes the `HostDefined` inheritance and switches `m_result` visitation to `Variant::visit` with a lambda that catches all `GC::Ref` arms.

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLScriptElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLScriptElement.cpp
@@ -48,8 +48,10 @@ void HTMLScriptElement::initialize(JS::Realm& realm)
 void HTMLScriptElement::visit_edges(Cell::Visitor& visitor)
 {
     Base::visit_edges(visitor);
-    if (auto* script = m_result.get_pointer<GC::Ref<Script>>())
-        visitor.visit(*script);
+    m_result.visit(
+        [](ResultState::Uninitialized) {},
+        [](ResultState::Null) {},
+        [&]<typename T>(GC::Ref<T> result) { visitor.visit(result); });
     visitor.visit(m_parser_document);
     visitor.visit(m_preparation_time_document);
 }
--- a/Libraries/LibWeb/HTML/Scripting/ImportMapParseResult.cpp
+++ b/Libraries/LibWeb/HTML/Scripting/ImportMapParseResult.cpp
@@ -40,11 +40,6 @@ GC::Ref<ImportMapParseResult> ImportMapParseResult::create(JS::Realm& realm, Byt
     return result;
 }
 
-void ImportMapParseResult::visit_host_defined_self(Visitor& visitor)
-{
-    visitor.visit(*this);
-}
-
 void ImportMapParseResult::visit_edges(Visitor& visitor)
 {
     Base::visit_edges(visitor);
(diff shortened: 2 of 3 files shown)


Libraries/LibWeb/HTML/HTMLScriptElement.cpp
```diff
@@ -50,6 +50,8 @@ void HTMLScriptElement::visit_edges(Cell::Visitor& visitor)
     Base::visit_edges(visitor);
     if (auto* script = m_result.get_pointer<GC::Ref<Script>>())
         visitor.visit(*script);
+    if (auto* import_map = m_result.get_pointer<GC::Ref<ImportMapParseResult>>())
+        visitor.visit(*import_map);
```
trflynn89: It looks like the `ImportMapParseResult` visits itself via `visit_host_defined_self`:
https://github.com/LadybirdBrowser/ladybird/blob/786e93b6c70a835349f7efa29059329d4172d8c6/Libraries/LibWeb/HTML/Scripting/ImportMapParseResult.cpp#L43-L46

(Where the base method for that is in `HostDefined` in Libraries/LibJS/Script.h)

CC'ing @awesomekling here because I'm finding the GC around this host-defined thing confusing 😅  Do we need this visit?

Praise-Garfield: `visit_host_defined_self` is only called from `JS::Script::visit_edges` (Script.cpp:246) and `JS::Module::visit_edges` (Module.cpp:40), via the `m_host_defined` raw pointer. For import maps, `prepare_script` creates an `ImportMapParseResult` directly and stores it in `m_result` -- no `JS::Script` or `JS::Module` is ever created with it as `host_defined`. So `ImportMapParseResult::visit_host_defined_self` is never invoked; the only GC reference to the cell is through `m_result` here.

trflynn89: Yeah sounds like `ImportMapParseResult` shouldn't inherit from `HostDefined` at all then.

Can we do the following?
1. Remove that inheritance and remove `ImportMapParseResult::visit_host_defined_self`
2. Change this `visit_edges` to invoke `m_result.visit(...)` in a way that catches all `GC::Ref` types

That would be something like:
```c++
m_result.visit(
    [](ResultState::Uninitialized) {},
    [](ResultState::Null) {},
    [&]<typename T>(GC::Ref<T> result) { visitor.visit(result); });
```

This ensures we visit all `GC::Ref<T>`s. And if something is added to `m_result` in the future that isn't a `GC::Ref`, we will get a compiler error and be forced to think about it.

I'm also fine if you would rather be more explicit:
```c++
m_result.visit(
    [](ResultState::Uninitialized) {},
    [](ResultState::Null) {},
    [&](GC::Ref<Script> result) { visitor.visit(result); },
    [&](GC::Ref<ImportMapParseResult> result) { visitor.visit(result); });
```

Praise-Garfield: Not a problem.

Issue #5099, opened 2025-06-15
compile on openSUSE tumbleweed fails ; error suggests opening github issue


tckosvic:
Looking to investigate ladybird and get it compiled to see where it's at.  Using up-to-date openSUSE tumbleweed. 
 Compile fails and error suggests to open issue.  Used commandis: ./Meta/ladybird.py run.
Code with error notice is below.

`
/home/tom/ladybird/Build/release/lib64/liblagom-core.so.0 Core::EventLoop::exec() 0x31) [0x7fcce87ce4a1]
/home/tom/ladybird/Build/release/libexec/WebContent(+0x4d37f) [0x55f543f9e37f]
/home/tom/ladybird/Build/release/libexec/WebContent(main+0xff) [0x55f543f9be0f]
/lib64/libc.so.6(+0x2b12e) [0x7fcce762b12e]
/lib64/libc.so.6(__libc_start_main+0x8b) [0x7fcce762b1f9]
/home/tom/ladybird/Build/release/libexec/WebContent(+0x4af45) [0x55f543f9bf45]
**5557.984 Ladybird(88350): WebContent process crashed! Last page loaded: about:srcdoc
5557.984 Ladybird(88350): Consider raising an issue at** https://github.com/LadybirdBrowser/ladybird/issues/new/choose
5557.988 WebContent(91140): FIXME: Handle null navigation_request in navigation response Content Security Policy check.
VERIFICATION FAILED: icu_success(status) at /home/tom/ladybird/Libraries/LibUnicode/Segmenter.cpp:240
/home/tom/ladybird/Build/release/lib64/liblagom-ak.so.0(dump_backtrace+0x49) [0x7fcb3a921ff9]
/home/tom/ladybird/Build/release/lib64/liblagom-ak.so.0(ak_trap+0x9) [0x7fcb3a922259]
/home/tom/ladybird/Build/release/lib64/liblagom-ak.so.0(ak_verification_failed+0x65) [0x7fcb3a921f85]
/home/tom/ladybird/Build/release/lib64/liblagom-unicode.so.0 Unicode::Segmenter::create(AK::StringView, Unicode::SegmenterGranularity) 0x134) [0x7fcb3b0c03c4]

`
Error opens an "error dialogue" box that has no text in it.  Error is in some loop that keeps repeating until manually stopped.
Anyone have a suggestion as to how to proceed.
Compilation also fails on openSUSE leap 15.6 but with different error.  I am investigating that further.

[...]

PR #8068, opened 2026-02-21
LibUnicode: Print error codes when calls to ICU fail

InvalidUsernameException:
Primary motivation for this change is the `VERIFY(icu_success(status))` line in `Segmenter::create()` that was failing on multiple systems and where we had to ask people to apply a patch to even know what the error was.

Since this seems to be a recurring problem, let's just add a little helper function and print the error codes returned by library calls.

Related: #3438, #5099

### DIFF
--- a/Libraries/LibUnicode/ICU.h
+++ b/Libraries/LibUnicode/ICU.h
@@ -8,6 +8,7 @@
 
 #include <AK/Optional.h>
 #include <AK/OwnPtr.h>
+#include <AK/SourceLocation.h>
 #include <AK/String.h>
 #include <AK/StringView.h>
 #include <AK/Utf16String.h>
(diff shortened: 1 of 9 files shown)


Libraries/LibUnicode/ICU.h
```diff
@@ -89,6 +89,14 @@ constexpr bool icu_failure(UErrorCode code)
     return static_cast<bool>(U_FAILURE(code));
 }
 
+constexpr void verify_icu_success(UErrorCode code)
```
trflynn89: `dbgln` is decidedly not `constexpr`, let's just make this `inline void` instead.

Libraries/LibUnicode/ICU.h
```diff
@@ -89,6 +89,14 @@ constexpr bool icu_failure(UErrorCode code)
...
+{
+    if (!icu_success(code)) {
```
trflynn89: Let's err on the side of the happy path:
```c++
if (icu_failure(code)) [[unlikely]] {
```

Libraries/LibUnicode/ICU.h
```diff
@@ -89,6 +89,14 @@ constexpr bool icu_failure(UErrorCode code)
...
+        dbgln("Call to ICU library failed with code {}", u_errorName(code));
```
trflynn89: Last nit:
we've typically found people notice errors like these better if we make them pop more. Something like:

```c++
dbgln("\033[31;1mICU error\033[0m: {}", u_errorName(code));
```

You could add a `SourceLocation` to the function as well if you'd like, but no big deal.

InvalidUsernameException: I'm used to seeing the line number on the the stack trace already. But I guess on some systems it is only the offsets, so `SourceLocation` would be useful for those cases.

Libraries/LibUnicode/ICU.h
InvalidUsernameException: Fair enough, kind of just copy pasted from above without thinking much, relying on the compiler to tell me I'm wrong. But I guess it doesn't do that.

PR #8073, opened 2026-02-22
LibImageDecoderClient: Remove sync id fetch from async decode request

jonbgamble:
The thread asserts are there to help, but if there's a valid case why one of these functions must be called from a secondary thread, we can drop them and use a mutex, or just go back to sync id fetch (which is not measurably slower)

### DIFF
--- a/Services/ImageDecoder/ConnectionFromClient.cpp
+++ b/Services/ImageDecoder/ConnectionFromClient.cpp
@@ -221,7 +221,13 @@ void ConnectionFromClient::decode_image(Core::AnonymousBuffer encoded_buffer, Op
         return;
     }
 
-    m_pending_jobs.set(request_id, make_decode_image_job(request_id, move(encoded_buffer), ideal_size, move(mime_type)));
+    auto set_result = m_pending_jobs.set(request_id, make_decode_image_job(request_id, move(encoded_buffer), ideal_size, move(mime_type)), AK::HashSetExistingEntryBehavior::Keep);
+
+    if (set_result != HashSetResult::InsertedNewEntry) {
+        m_pending_jobs.take(request_id).value()->cancel();
+        did_misbehave("Duplicate decode request id");
+        return;
+    }
 }
 
 void ConnectionFromClient::cancel_decoding(i64 request_id)


Libraries/LibImageDecoderClient/Client.h
```diff
@@ -51,15 +52,19 @@ class Client final
     Function<void(i64 session_id, String error_message)> on_animation_decode_failed;
 
 private:
+    void verify_thread_affinity() const { VERIFY(m_creation_thread == std::this_thread::get_id()); }
```
Zaggy1024: Instead of using std::thread, we could probably just compare EventLoop references.

Libraries/LibImageDecoderClient/Client.cpp
```diff
@@ -16,14 +16,20 @@ Client::Client(NonnullOwnPtr<IPC::Transport> transport)
 
 void Client::die()
 {
-    for (auto& [_, promise] : m_pending_decoded_images) {
+    verify_thread_affinity();
+    Vector<NonnullRefPtr<Core::Promise<DecodedImage>>> pending_promises;
+    pending_promises.ensure_capacity(m_token_promises.size());
+    for (auto& [_, promise] : m_token_promises)
+        pending_promises.unchecked_append(promise);
+    m_token_promises.clear();
```
Zaggy1024:
```suggestion
    verify_thread_affinity();
    auto pending_promises = move(m_token_promises);
```

Zaggy1024 (commented):
I think it feels like there are three separate changes here?

- Rename image_id -> request_token
- Add thread verification
- Remove sync IPC call

Perhaps I'm wrong about these being separable, but the rename muddies things a bit when I'm trying to parse what's changing.

Libraries/LibImageDecoderClient/Client.h
jonbgamble: Is there a problem with std::thread::id and std::this_thread::get_id()? They work on any thread regardless of event pump, and are cross platform (without pthreads4w or mingw).

Zaggy1024: Really the issue is that I don't personally _know_ if there are issues with using it. It's not really my area of expertise to say whether it's fine to use STL stuff like that, but if we were to use it, we would probably want it wrapped in AK functionality.

Regardless, IPC requires an event loop, so we should be fine to rely on that here, I believe.

jonbgamble:
> Perhaps I'm wrong about these being separable, but the rename muddies things a bit when I'm trying to parse what's changing.

My take was that all three changes are intertwined.

image_id was misleading. If you submit a request for the same image multiple times you would get separate image_ids, so it's more of a request token. But I noticed this pattern in other client libs uses "request_id" so I can rename it that. I'd personally keep the rename in the same commit with the asyncify though.

I can separate the thread affinity commit if you like. It's a supporting change but it does go beyond protecting the clientside map.

Services/ImageDecoder/ConnectionFromClient.cpp
```diff
@@ -202,35 +202,31 @@ NonnullRefPtr<ConnectionFromClient::Job> ConnectionFromClient::make_decode_image
...
-        async_did_fail_to_decode_image(image_id, "Encoded data is invalid"_string);
-        return image_id;
+        async_did_fail_to_decode_image(request_id, "Encoded data is invalid"_string);
+        return;
     }
 
-    m_pending_jobs.set(image_id, make_decode_image_job(image_id, move(encoded_buffer), ideal_size, move(mime_type)));
-
-    return image_id;
+    m_pending_jobs.set(request_id, make_decode_image_job(request_id, move(encoded_buffer), ideal_size, move(mime_type)));
```
Zaggy1024: What happens if the request_id already exists here? We should probably not assume that the client is well-behaved.

Zaggy1024 (commented):
I realized, we share the ImageDecoder service between multiple WebContent processes, don't we? This probably needs a different solution.

Services/ImageDecoder/ConnectionFromClient.cpp
jonbgamble: request_ids are scoped per connection. It's fine.

jonbgamble: It is the client's responsibility. That's why I actually preferred "token", but it's called "id" elsewhere.

Zaggy1024: > request_ids are scoped per connection. It's fine.

Ah, got it, that addresses my general comment then.

Saying that it's the client's responsibility doesn't answer the question of what happens if there's a conflict here, though. We replace the pending job, and then we have two threads working on the same ID, and they try to resolve the same promise twice? We should probably be skipping duplicates here at the very least.

jonbgamble: The server stores pending jobs by request id *per connection*. A single client cannot send two different requests with the same id due to monotonic id generation and our main thread assert.

Zaggy1024: In normal cases, yes, it shouldn't. If that's the assumption, then we should write that assumption into the code with an assertion. Let's not just assume that nothing can go wrong in code that is running in an entirely different process without making it clear in the code nor handling it gracefully.

jonbgamble: It's not an assumption though. It's the contract and it's the way the client library works. It's also the way RequestServer client/server works.

If you want to harden ImageDecoder so it can serve other potential clients using different implementations, it seems a little out of this PRs scope.

Zaggy1024: Contract, assumption, invariant, I'm not trying to argue semantics here. If we expect that no request ID will conflict, then let's make that explicit.

Let's say someone made a PR down the line that caused the client to determine the ID differently, and we're just overwriting promises here silently, we get undefined behavior because we haven't defined it. Or if the client side corrupts its memory somehow, sends an invalid request, we get undefined behavior. There's no good reason _not_ to encode a contract here if we're making this value be determined separately.

Even regardless of whether it's coming from another process, I would prefer a verify here, just to make it clear what we expect to happen, since it makes for more intentional code.

jonbgamble: That makes sense. But so we're on the same page, a VERIFY on the server could affect other pages if ImageDecoder is ever restarted. I'm OK with that if you are, but first consider an alternative (with a smaller blast radius). We could just disconnect them.

Zaggy1024: Yeah, it's fine for it to crash if this is something that is supposed to never happen.

jonbgamble: You've seen the most recent commit? I thought `did_misbehave` might be better because it punishes the offending process.

Zaggy1024: Ah, I missed that, lgtm.

PR #8077, opened 2026-02-22
LibURL: Strip trailing dot before running the PSL algorithm

tcl3:

### DIFF
--- a/Libraries/LibURL/Host.cpp
+++ b/Libraries/LibURL/Host.cpp
@@ -192,14 +192,20 @@ Optional<String> Host::public_suffix() const
     auto trailing_dot = host_string.ends_with('.') ? "."sv : ""sv;
 
     // 3. Let publicSuffix be the public suffix determined by running the Public Suffix List algorithm with host as domain. [PSL]
+
+    // NB: The PSL algorithm maintains trailing dots, so we strip it here since step 4 expects no trailing dot.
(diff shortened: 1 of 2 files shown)


Tests/LibWeb/Text/input/wpt-import/fetch/metadata/trailing-dot.https.sub.any.js
```diff
@@ -0,0 +1,30 @@
...
+  return validate_expectations_custom_url("https://www.wpt.live.:443/fetch/metadata/resources/echo-as-json.py", {}, {
+          "site": "cross-site",
+          "user": "",
+          "mode": "cors",
+          "dest": "empty"
+        }, "Fetching a resource from the same site, but spelled with a trailing dot.");
+}, "Fetching a resource from the same site, but spelled with a trailing dot.");
+
+promise_test(t => {
+  return validate_expectations_custom_url("https://www.not-wpt.live.:443/fetch/metadata/resources/echo-as-json.py", {}, {
```
shannonbooth: I think it would ve best to not fetch from wpt live as part our test suite. I suppose an alternative may be to add a C++ test for this perhaps? (I believe some may exist in TestURL.cpp already)

tcl3: I've replaced the imported test with a minimal crash test.

PR #8081, opened 2026-02-22
LibWeb: Clear old PlaybackManager callbacks before creating a new one

tcl3:
Fixes http://wpt.live/html/semantics/embedded-content/media-elements/loading-the-media-resource/load-events-networkState.html. This test relies on python scripts provided by the WPT runner, so I made a  simpler test that causes the same crash.

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLMediaElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLMediaElement.cpp
@@ -1492,6 +1492,14 @@ void HTMLMediaElement::on_metadata_parsed()
 // https://html.spec.whatwg.org/multipage/media.html#media-data-processing-steps-list
 void HTMLMediaElement::set_up_playback_manager(NonnullRefPtr<FetchData> const& fetch_data)
 {
+    if (m_playback_manager) {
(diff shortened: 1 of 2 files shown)


Zaggy1024:
It feels to me like it shouldn't be necessary to clear these callbacks manually, I'm curious if the playback manager is being kept alive somewhere it shouldn't be. I'll try to take a look at this later.

Zaggy1024:
Hmm, I'm actually thinking it may be worth removing ref counting from `PlaybackManager`, but keeping the `WeakPlaybackManager` around for callback handlers. The only place we're holding a strong reference to it is in `SeekingStateHandler`, and that should probably also be a `NonnullRefPtr<WeakPlaybackManager>`. All callbacks that `PlaybackManager` subscribes to interact with its members from the originating thread, so there's no need for thread-safe ref counting, as far as I know.

I'll start making a patch for this to see if it works.

Edit: GitHub sanitized my `<WeakPlaybackManager>` out of the message, that didn't make a whole lot of sense without code tags 😅

tcl3:
Closing in favor of #8095

PR #8086, opened 2026-02-22
Some cleanup around credentials management

devgianlu:
This PR aims to cleanup some things around credentials management to hopefully bring actual implementations in a separate PR. See [in progress](https://github.com/devgianlu/ladybird/compare/credentials-cont...devgianlu:ladybird:webauthn).

### DIFF
--- a/Libraries/LibWeb/CredentialManagement/PasswordCredentialOperations.h
+++ b/Libraries/LibWeb/CredentialManagement/PasswordCredentialOperations.h
@@ -10,7 +10,7 @@
 
 namespace Web::CredentialManagement {
 
-WebIDL::ExceptionOr<GC::Ref<PasswordCredential>> create_password_credential(JS::Realm& realm, GC::Ptr<HTML::HTMLFormElement> const&, URL::Origin);
+WebIDL::ExceptionOr<GC::Ref<PasswordCredential>> create_password_credential(JS::Realm& realm, GC::Ref<HTML::HTMLFormElement>, URL::Origin);
 WebIDL::ExceptionOr<GC::Ref<PasswordCredential>> create_password_credential(JS::Realm& realm, PasswordCredentialData const&, URL::Origin);
 
 }
--- a/Libraries/LibWeb/CredentialManagement/PasswordCredential.cpp
+++ b/Libraries/LibWeb/CredentialManagement/PasswordCredential.cpp
@@ -12,7 +12,7 @@ namespace Web::CredentialManagement {
 GC_DEFINE_ALLOCATOR(PasswordCredential);
 
 // https://www.w3.org/TR/credential-management-1/#dom-passwordcredential-passwordcredential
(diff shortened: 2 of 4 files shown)


Libraries/LibWeb/CredentialManagement/FederatedCredential.cpp
```diff
@@ -25,11 +25,11 @@ FederatedCredential::~FederatedCredential()
 {
 }
 
-FederatedCredential::FederatedCredential(JS::Realm& realm, FederatedCredentialInit const& init, URL::Origin origin)
+FederatedCredential::FederatedCredential(JS::Realm& realm, FederatedCredentialInit const& init, URL::Origin const& origin)
     : Credential(realm, init.id)
     , CredentialUserData(init.name.value_or(String {}), init.icon_url.value_or(String {}))
     , m_provider(init.provider)
-    , m_origin(move(origin))
+    , m_origin(origin)
```
trflynn89: I don't see where we were making copies before. We were correctly moving the origins created in `PasswordCredential::construct_impl` into this constructor.

Instead, it looks like we now actually force a copy here and in the `PasswordCredential` constructor. So unless I'm missing something, this patch has increased the number of copies.

Libraries/LibWeb/CredentialManagement/PasswordCredentialOperations.h
```diff
@@ -10,7 +10,10 @@
 
 namespace Web::CredentialManagement {
 
-WebIDL::ExceptionOr<GC::Ref<PasswordCredential>> create_password_credential(JS::Realm& realm, GC::Ptr<HTML::HTMLFormElement> const&, URL::Origin);
-WebIDL::ExceptionOr<GC::Ref<PasswordCredential>> create_password_credential(JS::Realm& realm, PasswordCredentialData const&, URL::Origin);
+// https://www.w3.org/TR/credential-management-1/#abstract-opdef-create-a-passwordcredential-from-an-htmlformelement
+WebIDL::ExceptionOr<GC::Ref<PasswordCredential>> create_password_credential(JS::Realm& realm, GC::Ptr<HTML::HTMLFormElement> const&, URL::Origin const&);
```
trflynn89: Incidentally, `GC::Ptr` should not be passed by reference either. It's just a pointer.

(It also looks like we are assuming the pointer is non-null. We use `GC::Ref` for such cases.)

Libraries/LibWeb/CredentialManagement/FederatedCredential.cpp
devgianlu: Yes, I am a dumbass :)

Libraries/LibWeb/CredentialManagement/PasswordCredentialOperations.h
devgianlu: Fixed

devgianlu:
Windows CI failure looks like a network error, Linux CI crash should be unrelated.

PR #8089, opened 2026-02-22
LibWeb: Implement most `CredentialsContainer` methods

devgianlu:
This PR mplements `get`, `store`, `create` on `CredentialsContainer`, leaving out only `prevent_silent_access`. This is a revival of #3495 with some improvements.

Notably, it also introduces the `CredentialInterface` singleton abstract class to deal with credentials internal methods.

The spec isn't the best and contains some bugs:
- The `options.signal && options.signal->aborted()` check in `get` and `create` is not deferred therefore fails two WPT tests
- The `interfaces.size() < 1` check in `create` is custom and avoids an out of bound access in the next step -> [Spec issue](https://github.com/w3c/webappsec-credential-management/issues/267)
- The `CredentialRequestOptions` IDL dictionary has a default value for `password` which renders the credential type always active -> [Spec issue](https://github.com/w3c/webappsec-credential-management/issues/270)
- Others are linked directly in the code

This foundation should be enough to start working on WebAuthn. The interactive part is still missing as `ask_the_user_to_choose_a_credential` is stubbed out.

### DIFF
--- a/Libraries/LibWeb/CredentialManagement/Credential.h
+++ b/Libraries/LibWeb/CredentialManagement/Credential.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025, Altomani Gianluca <altomanigianluca@gmail.com>
+ * Copyright (c) 2026, Altomani Gianluca <altomanigianluca@gmail.com>
  *
  * SPDX-License-Identifier: BSD-2-Clause
  */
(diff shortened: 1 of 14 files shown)


Libraries/LibWeb/CredentialManagement/Credential.h
```diff
@@ -13,6 +13,74 @@
...
+    AK_MAKE_NONMOVABLE(CredentialInterface);
+
+protected:
+    CredentialInterface() { }
+
+public:
+    virtual ~CredentialInterface() = default;
+
+    // https://w3c.github.io/webappsec-credential-management/#credential-type-registry-credential-type
+    virtual String type() const = 0;
```
shannonbooth, line 38: Credential Type is a known set of enumerations - this could be turned into an enum class.

Libraries/LibWeb/CredentialManagement/Credential.h
```diff
@@ -13,6 +13,74 @@
...
+
+    // https://w3c.github.io/webappsec-credential-management/#credential-type-registry-options-member-identifier
+    virtual String options_member_identifier() const = 0;
```
shannonbooth, line 41: Similar here - this can also be made into an enum, though as I go through these - looks like this can be extended to a lot of the functions here.

Libraries/LibWeb/CredentialManagement/FederatedCredentialOperations.cpp
```diff
@@ -10,7 +10,7 @@
 namespace Web::CredentialManagement {
 
 // https://www.w3.org/TR/credential-management-1/#abstract-opdef-create-a-federatedcredential-from-federatedcredentialinit
-WebIDL::ExceptionOr<GC::Ref<FederatedCredential>> create_federated_credential(JS::Realm& realm, FederatedCredentialInit const& init)
+WebIDL::ExceptionOr<GC::Ref<FederatedCredential>> create_federated_credential(JS::Realm& realm, FederatedCredentialInit const& init, Optional<URL::Origin> origin)
```
shannonbooth: This can be passed by `const&`

Libraries/LibWeb/CredentialManagement/PasswordCredential.cpp
```diff
@@ -51,4 +55,31 @@ void PasswordCredential::initialize(JS::Realm& realm)
...
+        [&](GC::Root<HTML::HTMLFormElement> const& form) {
+            return create_password_credential(realm, *form, origin);
+        },
+        // 3. If options["password"] is a PasswordCredentialData, return the result of executing
+        //     Create a PasswordCredential from PasswordCredentialData given options["password"]. Rethrow any exceptions.
+        [&](PasswordCredentialData const& data) {
+            return create_password_credential(realm, data, origin);
+        },
+        // 4. Throw a TypeError exception.
+        [&](auto) {
```
shannonbooth: `(auto const&)`

Libraries/LibWeb/CredentialManagement/Credential.h
```diff
@@ -13,6 +13,74 @@
...
+    virtual Optional<String> create_permission_policy() const = 0;
+
+    // https://w3c.github.io/webappsec-credential-management/#dom-credential-discovery-slot
+    virtual String discovery() const = 0;
+
+    // NOTE: This is not explicitly present in the spec, it is inferred.
+    virtual bool supports_conditional_user_mediation() const = 0;
+
+    // https://w3c.github.io/webappsec-credential-management/#algorithm-create-cred
+    virtual JS::ThrowCompletionOr<Variant<Empty, GC::Ref<Credential>, GC::Ref<CreateCredentialAlgorithm>>> create(JS::Realm&, URL::Origin const&, CredentialCreationOptions const&, bool) const
```
shannonbooth: It would be a little more standard here since we are in LibWeb to be returning a WebIDL::ExceptionOr for these methods, which would allow some TRY() cleanup in for example PasswordCredentialInterface::create, which can then do:

`return GC::Ref<Credential> { TRY(options.password->visit(`

As an example, and perform the Bindings::exception_to_throw_completion at the call sites of these methods. Same here and for these other methods

shannonbooth (changes requested):
Sorry for the delay in review!

Libraries/LibWeb/CredentialManagement/Credential.h
devgianlu: This cannot be an enum because it is used together with `Credential.type()` which is an IDL string.

Libraries/LibWeb/CredentialManagement/Credential.h
devgianlu: This one particularly should stay a string because it is used to determine the key in the `credentials.create` object being passed.

I could change the other ones, but they are mostly used as of now.

Libraries/LibWeb/CredentialManagement/Credential.h
shannonbooth: Ah, I missed that - but I think in that case I would still probably make this an enum - I'm actually a bit surprised that the spec itself does not do that in the IDL since enums in IDL are also strings.


```idl
enum CredentialType {
    "digital-credential", 
    "federated",
    "identity",
    "otp",
    "password",
    "public-key"
};

[Exposed=Window, SecureContext]
interface Credential {
    readonly attribute USVString id;
    readonly attribute CredentialType type;
    static Promise<boolean> isConditionalMediationAvailable();
};

```
Probably what we can do for now is something along the lines of:

 [ImplementedAs=type_for_bindings] readonly attribute DOMString type;

And type_for_bindings can do the enum to string conversion. If the spec had IDL kind of like the above the IDL generator would do that conversion for us.

PR #8090, opened 2026-02-22
CMake: Create CACHEDIR.TAG file in `Build` directory

rgret-dev:
This incorporates the Cache Directory Tagging Specification, see: https://bford.info/cachedir/
...so that backup tools won't save the build artifacts.

There are probably other places we could incorporate this as well (e.g. the disk cache directory) but just tag the build for now.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,9 @@ endif()
 # vcpkg flags depend on what linker we are using
 include("Meta/CMake/use_linker.cmake")
 
+# cache directory tag creation
+include("Meta/CMake/cachedir_tag.cmake")
+
 if (APPLE AND NOT CMAKE_OSX_DEPLOYMENT_TARGET)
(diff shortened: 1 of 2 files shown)


gmta:
Hi! My 2 cents; I'd rather not merge something that has to do with excluding directories in people's backup software. That's something that should be up to the end user. We might not generate anything of true importance in the build directory, but we might not be the only ones to put something in there.

rgret-dev:
> Hi! My 2 cents; I'd rather not merge something that has to do with excluding directories in people's backup software. That's something that should be up to the end user. We might not generate anything of true importance in the build directory, but we might not be the only ones to put something in there.

I think using the build directory as a workspace for important files is quite niche and the general case is going to consider the build folder ephemeral.

For example, the standard advice on the #build-problems channel in Discord is to "nuke" the build directory to get a clean slate after a system update or something similar.

As mentioned on discord, Cargo places a CACHEDIR.TAG file in their build ("target") directories by default.

PR #8094, opened 2026-02-22
LibWeb: Skip pseudo-element generated text in background-clip:text

### LINKED ISSUE #7498: Text gets drawn twice
With the following reduction:
```html
<!doctype html>
<html>
    <style>
        h1 {
            background: black;
            background-clip: text;
            -webkit-text-fill-color: transparent;
        }
        h1:before {
            content: attr(data-text);
            position: absolute;
            left: 200px;
        }
    </style>
    <h1 data-text="test">test</h1>
</html>
```

On Firefox and Chrome:
Image

On Ladybird:
Image

mikiubo:
Ignore fragments generated for pseudo-elements when computing text clip paths.
These should not contribute to the element’s text clipping per CSS backgrounds and borders.

Fixes #7498

### DIFF
--- a/Libraries/LibWeb/Painting/BackgroundPainting.cpp
+++ b/Libraries/LibWeb/Painting/BackgroundPainting.cpp
@@ -2,6 +2,7 @@
  * Copyright (c) 2018-2020, Andreas Kling <andreas@ladybird.org>
  * Copyright (c) 2021-2023, Sam Atkins <atkinssj@serenityos.org>
  * Copyright (c) 2022, MacDue <macdue@dueutil.tech>
+ * Copyright (c) 2026, mikiubo <michele.uboldi@gmail.com>
  *
  * SPDX-License-Identifier: BSD-2-Clause
  */
@@ -41,8 +42,14 @@ static RefPtr<DisplayList> compute_text_clip_paths(DisplayListRecordingContext& 
         display_list_recorder.draw_glyph_run(baseline_start, *glyph_run, Gfx::Color::Black, fragment_absolute_device_rect.to_type<int>(), scale, fragment.orientation());
     };
 
-    paintable.for_each_in_inclusive_subtree([&](auto& paintable) {
-        if (auto* paintable_lines = as_if<PaintableWithLines>(paintable)) {
+    paintable.for_each_in_inclusive_subtree([&](auto& sub_paintable) {
+        // https://drafts.csswg.org/css-backgrounds-4/#valdef-background-clip-text
+        if (&sub_paintable != &paintable) {
+            auto& layout_node = sub_paintable.layout_node();
+            if (!layout_node.is_in_flow() && !layout_node.is_floating())
+                return TraversalDecision::SkipChildrenAndContinue;
+        }
+        if (auto* paintable_lines = as_if<PaintableWithLines>(sub_paintable)) {
             for (auto const& fragment : paintable_lines->fragments()) {
                 if (is<Layout::TextNode>(fragment.layout_node()))
                     add_text_clip_path(fragment);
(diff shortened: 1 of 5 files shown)


Psychpsyo:
Having a spec citation in the code for why it does that would be nice.

mikiubo:
> Having a spec citation in the code for why it does that would be nice.

Thanks! While looking up the spec to add the citation I noticed the original condition was too narrow — I was checking for pseudo-elements specifically, but the spec actually excludes all out-of-flow descendants. I've updated the fix accordingly and added the spec reference in a NOTE comment.

Libraries/LibWeb/Painting/BackgroundPainting.cpp
```diff
@@ -41,8 +42,13 @@ static RefPtr<DisplayList> compute_text_clip_paths(DisplayListRecordingContext&
         display_list_recorder.draw_glyph_run(baseline_start, *glyph_run, Gfx::Color::Black, fragment_absolute_device_rect.to_type<int>(), scale, fragment.orientation());
     };
 
-    paintable.for_each_in_inclusive_subtree([&](auto& paintable) {
-        if (auto* paintable_lines = as_if<PaintableWithLines>(paintable)) {
+    paintable.for_each_in_inclusive_subtree([&](auto& sub_paintable) {
+        // NOTE: The background is painted within (clipped to) the intersection of the border box and the geometry of the text
+        // including any decorations and emphasis marks in the element and its in-flow and floated descendants.
+        // https://www.w3.org/TR/css-backgrounds-4/#background-clip
```
tcl3: Rather than having a NOTE comment here, it looks like you could quote the spec text directly from https://drafts.csswg.org/css-backgrounds-4/#valdef-background-clip-text.

Libraries/LibWeb/Painting/BackgroundPainting.cpp
```diff
@@ -41,8 +42,13 @@ static RefPtr<DisplayList> compute_text_clip_paths(DisplayListRecordingContext&
...
+        if (&sub_paintable != &paintable && sub_paintable.layout_node().is_out_of_flow())
```
tcl3: It looks like this would incorrectly skip floated elements. For example, this appears to render incorrectly:
```html
<!DOCTYPE html>
<style>
  div {
    background: linear-gradient(red, blue);
    background-clip: text;
    color: transparent;
    font: bold 40px sans-serif;
  }
</style>
<div>World<span style="float: left">Hello</span></div
```

Excluding elements which are not in-flow or floated would better match the above spec text.

mikiubo: you are right. 

I added you suggested html as test.

Now it seems is fixed.

Tests/LibWeb/Screenshot/input/css-background-clip-text-float.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<style>
+  div {
+    background: linear-gradient(red, blue);
+    background-clip: text;
+    color: transparent;
+    font: bold 40px sans-serif;
+  }
+</style>
+<div>World<span style="float: left">Hello</span></div>
```
tcl3: This should have a newline at EOF.

tcl3: Actually, could we make this a ref test instead of a screenshot test? Screenshot tests are reserved for testing exact rendering and should be used when there's no other appropriate test type, since they are inherently quite brittle. For this particular test we only care that the `float:left` rule is applied correctly.

mikiubo: ok i will change also the other test to be ref (originally i think about it as ref then i changed my mind because i misinterpreted the meaning of screenshot  tests ).

i will notify you

Thanks

mikiubo: @tcl3 i think i fixed
let me know if you notice anything

PR #8095, opened 2026-02-23
LibMedia+LibWeb: Stop ref counting PlaybackManager

Zaggy1024:
PlaybackManager's ref counting was only used to keep it alive in a few callbacks. Instead, the callbacks can use weak references that can only be used from the thread that the PlaybackManager was created on, to ensure that the PlaybackManager can't be destroyed while being accessed.

This ensures that:
- The PlaybackManager is destroyed immediately when it is reassigned by HTMLMediaElement
- No callbacks are invoked after that point

This fixes the crash initially being addressed by #8081. The test from that PR has been included as a regression test.

CC @tcl3

### DIFF
--- a/Libraries/LibMedia/PlaybackManager.cpp
+++ b/Libraries/LibMedia/PlaybackManager.cpp
@@ -21,7 +21,7 @@
 
 namespace Media {
 
-DecoderErrorOr<void> PlaybackManager::prepare_playback_from_media_data(NonnullRefPtr<IncrementallyPopulatedStream> stream, NonnullRefPtr<Core::WeakEventLoopReference> const& main_thread_event_loop_reference)
+DecoderErrorOr<void> PlaybackManager::prepare_playback_from_media_data(WeakPlaybackManager const& self, NonnullRefPtr<IncrementallyPopulatedStream> stream, NonnullRefPtr<Core::WeakEventLoopReference> const& main_thread_event_loop_reference)
 {
     auto demuxer = TRY([&] -> DecoderErrorOr<NonnullRefPtr<Demuxer>> {
         if (Matroska::Reader::is_matroska_or_webm(stream->create_cursor()))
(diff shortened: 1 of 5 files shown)


Libraries/LibMedia/PlaybackManager.cpp
```diff
@@ -151,38 +153,38 @@ void PlaybackManager::add_media_source(NonnullRefPtr<IncrementallyPopulatedStrea
...
+}
+
 void PlaybackManager::set_up_data_providers()
 {
     for (auto const& video_track_data : m_video_track_datas) {
-        video_track_data.provider->set_error_handler([weak_self = m_weak_wrapper](DecoderError&& error) {
-            auto self = weak_self->take_strong();
-            if (!self)
+        video_track_data.provider->set_error_handler([self = weak()](DecoderError&& error) {
+            if (self)
```
tcl3: should this be `if (!self)`?

Libraries/LibMedia/PlaybackManager.cpp
```diff
@@ -77,70 +77,72 @@ DecoderErrorOr<void> PlaybackManager::prepare_playback_from_media_data(NonnullRe
...
-        auto maybe_error = playback_manager->prepare_playback_from_media_data(stream, main_thread_event_loop_reference);
+    auto thread = Threading::Thread::construct("Media Init"sv, [self = weak(), stream = stream, main_thread_event_loop_reference = Core::EventLoop::current_weak()] mutable -> int {
+        auto maybe_error = prepare_playback_from_media_data(self, stream, main_thread_event_loop_reference);
         if (maybe_error.is_error()) {
             auto main_thread_event_loop = main_thread_event_loop_reference->take();
-            main_thread_event_loop->deferred_invoke([playback_manager, error = maybe_error.release_error()] mutable {
-                if (playback_manager->on_unsupported_format_error)
-                    playback_manager->on_unsupported_format_error(move(error));
+            main_thread_event_loop->deferred_invoke([self = move(self), error = maybe_error.release_error()] mutable {
+                if (self->on_unsupported_format_error)
```
tcl3, line 146: Is it possible for `self` to be null here?

Libraries/LibMedia/PlaybackManager.cpp
Zaggy1024: Whoops, I was afraid I'd make that typo somewhere 😅 good catch.

Libraries/LibMedia/PlaybackManager.h
```diff
@@ -88,30 +89,10 @@ class MEDIA_API PlaybackManager final : public AtomicRefCounted<PlaybackManager>
...
-        {
-            Threading::MutexLocker locker { m_mutex };
-            m_manager = nullptr;
-        }
-
-        mutable Threading::Mutex m_mutex;
-        PlaybackManager* m_manager { nullptr };
-    };
+    WeakPlaybackManager weak();
+    WeakPlaybackManager const weak() const;
```
tcl3: This doesn't have an implementation.

tcl3 (approved):

PR #8097, opened 2026-02-23
LibWeb: Ensure favicon URL is valid before fetching it

tcl3:
Previously, we would crash when attempting to fetch the favicon if the `<base>` element had an invalid URL.

I haven't included a test for this because the crashing code path is only invoked for http(s) URLs. Tested manually by running this from a local web server:
```html
<!DOCTYPE html>
<base href="invalid:">
```
Fixes this WPT crash test: https://wpt.live/svg/pservers/pattern-with-invalid-base-cloned-thcrash.html


shannonbooth:
> Tested manually by running this from a local web server:

It should be possible to load such a test by webserver now through TestConfig.ini, or does that not work here?

tcl3:
> It should be possible to load such a test by webserver now through TestConfig.ini, or does that not work here?

Excellent point, doing this does indeed make the test work as intended.

PR #8099, opened 2026-02-23
LibMedia: Fix signed PCM to float sample normalization

jonbgamble:


Libraries/LibMedia/FFmpeg/FFmpegAudioDecoder.cpp
```diff
@@ -140,8 +140,8 @@ requires(IsSigned<T>)
 static float float_sample_from_frame_data(u8** data, size_t plane, size_t index)
 {
     auto* pointer = reinterpret_cast<T*>(data[plane]);
-    constexpr float full_scale = NumericLimits<MakeUnsigned<T>>::max() / 2;
-    return static_cast<float>(pointer[index]) / static_cast<float>(full_scale);
+    constexpr float full_scale = (NumericLimits<MakeUnsigned<T>>::max() + 1) / 2.0f;
```
gmta: Wait, we're taking in `T` which is `i16/i32/i64` at the time of writing, make it unsigned, take its max (which is typed as `MakeUnsigned<T>` as well), and then do `+ 1`? Doesn't that wrap around to 0 when no integer promotion is in order (i.e. i32/i64)?

Libraries/LibMedia/FFmpeg/FFmpegAudioDecoder.cpp
```diff
@@ -140,8 +140,8 @@ requires(IsSigned<T>)
...
+    return static_cast<float>(pointer[index]) / full_scale;
 }
```
gmta, line 145: This could do with a test so we lock this expectation in place :)

jonbgamble: Good call. We already tried (and failed) to fix it before.

Libraries/LibMedia/FFmpeg/FFmpegAudioDecoder.cpp
```diff
@@ -140,8 +140,8 @@ requires(IsSigned<T>)
 static float float_sample_from_frame_data(u8** data, size_t plane, size_t index)
 {
     auto* pointer = reinterpret_cast<T*>(data[plane]);
-    constexpr float full_scale = NumericLimits<MakeUnsigned<T>>::max() / 2;
-    return static_cast<float>(pointer[index]) / static_cast<float>(full_scale);
+    constexpr float inverse_peak = 1.0f / ((static_cast<float>(NumericLimits<MakeUnsigned<T>>::max()) / 2.0f) + 0.5f);
```
Zaggy1024: Where does the `+ 0.5f` here come from? I thought that i16 was supposed to divide by 2^15, i32 divide by 2^31, etc?

Shouldn't this just be `1.0f / (static_cast<float>(NumericLimits<MakeUnsigned<T>::max()) / 2.0f)`?

jonbgamble: unsigned short max is 65535
unsigned int max is something, something, i wanna say ending in 7?

but they're all odd so that's where the 0.5f comes from

PR #8102, opened 2026-02-23
Meta: Actually check PNG sizes in CI

gmta:
This only worked in our pre-commit hook; in CI we dutifully installed `optipng` only to always skip doing the work.

* Optimize PNGs in our repository;
* Check all PNGs if no path was provided as an argument (similar to other CI check tools);
* Check PNG paths with`optipng` in parallel, based on the number of available cores;
* Report the paths of the PNGs that can be optimized.

### DIFF
--- a/Meta/check-png-sizes.sh
+++ b/Meta/check-png-sizes.sh
@@ -38,17 +38,25 @@ for file in "${input_files[@]}"; do
 done
 
 if (( ${#files[@]} )); then
-    # We need to allow optipng to write output so we can check what it actually did. We use a dummy file that's discarded afterwards.
-    optimizations=$( printf '%s\0' "${files[@]}" |\
-        xargs -0 -n1 optipng -strip all -out dummy-optipng-output.png -clobber 2>&1 |\
-        grep -i -e 'Output IDAT size =' |\
-        sed -E 's/Output IDAT size = [0-9]+ byte(s?) \(([0-9]+) byte(s?) decrease\)/\2/g;s/Output IDAT size = [0-9]+ byte(s?) \(no change\)/0/g' |\
-        awk "{ if (\$1 >= $MINIMUM_OPTIMIZATION_BYTES) { S+=\$1 } } END { print S }")
-    rm -f dummy-optipng-output.png dummy-optipng-output.png.bak
-    optimizations="${optimizations:-0}"
+    tmpdir=$(mktemp -d)
+    # Each parallel invocation writes to its own temp file (using the sh process PID) to avoid clobbering. The per-file check
+    # is done inside sh -c so that filenames and byte counts stay paired despite interleaved parallel output.
+    # shellcheck disable=SC2016
+    results=$(printf '%s\0' "${files[@]}" \
+        | xargs -0 -n1 -P0 sh -c '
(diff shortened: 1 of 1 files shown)


Meta/check-png-sizes.sh
```diff
@@ -38,17 +38,26 @@ for file in "${input_files[@]}"; do
...
-
-    if [[ "$optimizations" -ne 0 ]] ; then
-        echo "There are non-optimized PNG images in Base/. It is possible to reduce file sizes by at least $optimizations byte(s)."
+    nprocs=$(nproc 2>/dev/null || sysctl -n hw.logicalcpu 2>/dev/null || echo 1)
+    tmpdir=$(mktemp -d)
+    # Each parallel invocation writes to its own temp file (using the sh process PID) to avoid clobbering. The per-file check
+    # is done inside sh -c so that filenames and byte counts stay paired despite interleaved parallel output.
+    # shellcheck disable=SC2016
+    results=$(printf '%s\0' "${files[@]}" \
+        | xargs -0 -n1 -P"$nprocs" sh -c '
```
Hendiadyoin1: Note `xargs -P` should already take all cores if no amount is provided

Documentation/Images/devtools_connect.png
Hendiadyoin1, line 1: Not sure how "worth" it is to do so after the fact, as git will still keep the old version in the history
There are ways to do this retroactively but I think they all require a force-push...

gmta: True - but it sets a precedent, it will allow the CI check to be exhaustive, _and_ it will still help people with shallow clones.

Meta/check-png-sizes.sh
gmta: It's `-P0`, not `-P`, but that does seem to work as well! Thanks :)

Tests/LibGfx/test-inputs/png/wide-gamut-only.png
Zaggy1024, line 1: I see a difference in rendering on this image in Firefox, is optipng stripping color space information?

gmta: Oh nice find! Yeah that should probably not be stripped away 🤔

Zaggy1024: Unfortunately I think we may still have a problem running optipng against this file. The original is 16-bit color, and apparently [optipng reduces the bit depth to 8 bits](https://man.archlinux.org/man/optipng.1.en#EXTENDED_DESCRIPTION). I've confirmed by downloading the original file and the optimized one, opened them in GIMP (kept the original color profile), and the title bar indicates that the original is indeed 16-bit, and the optimized one is indexed 8-bit. If it's possible to make it indexed 16-bit, though, we might still be able to squeeze almost the same improvement out of it.

gmta: Starting to think we should specifically ignore that file and go back to `-strip all` by default 😅

gmta: Confirmed that `wide-gamut-only.png` is not even used; removed the file.

PR #8104, opened 2026-02-23
LibJS: Add Rust implementation of lexer, parser, AST, and bytecode generator

awesomekling:
This adds a memory-safe Rust implementation of the LibJS source-to-bytecode pipeline: the lexer, parser, AST, and bytecode generator. The Rust code coexists with the C++ implementation for now.

The translation was done using AI coding agents (Claude Code and Codex). This was human-directed, not autonomous code generation. I decided what to port, in what order, and what the Rust code should look like. After the initial translation, I ran multiple passes of adversarial review, asking different models to analyze the code for mistakes and bad patterns.

The requirement from the start was byte-for-byte identical output from both pipelines. Every AST produced by the Rust parser is identical to the C++ one, and all bytecode generated by the Rust compiler is identical to the C++ compiler's output.

| Test suite | Tests | Regressions |
|---|---|---|
| test262 | 52,898 | 0 |
| Ladybird regression tests | 12,461 | 0 |

Beyond the test suites, I've done extensive testing by browsing the web in a lockstep mode where both pipelines run simultaneously, verifying that output is identical for every piece of JavaScript that flows through them.

The code has a strong "Rust translated from C++" vibe, and lots of cleanup will be needed in the future. The most awkward thing is register allocation: to make the lockstep mode possible, the Rust codegen does manual `drop()` of registers to ensure that register reuse ordering matches C++ exactly. This looks very strange but it will go away once we're comfortable retiring the C++ pipeline.

The Rust pipeline is enabled by default (except on Windows, where we need some help getting it to work). Two environment variables are available for testing:

- `LIBJS_CPP` forces the C++ pipeline to be used instead.
- `LIBJS_COMPARE_PIPELINES` runs both pipelines in lockstep, comparing results and crashing on any diff.

### DIFF
--- a/Libraries/LibJS/Rust/Cargo.toml
+++ b/Libraries/LibJS/Rust/Cargo.toml
@@ -6,6 +6,8 @@ edition = "2021"
 [lib]
 crate-type = ["staticlib"]
 
+# After changing dependencies, regenerate the Flatpak sources:
+#   python3 Meta/CMake/flatpak/generate-cargo-sources.py
 [dependencies]
 unicode-ident = "1.0"
 num-bigint = "0.4"
--- a/Meta/CMake/flatpak/cargo-sources.json
+++ b/Meta/CMake/flatpak/cargo-sources.json
@@ -1,0 +1,55 @@
+[
+    {
+        "type": "git",
+        "url": "https://github.com/corrosion-rs/corrosion.git",
+        "tag": "v0.5.1",
+        "dest": "corrosion"
+    },
+    {
+        "type": "archive",
+        "archive-type": "tar-gzip",
+        "url": "https://static.crates.io/crates/autocfg/autocfg-1.5.0.crate",
+        "sha256": "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8",
+        "dest": "cargo/vendor/autocfg-1.5.0"
+    },
+    {
+        "type": "archive",
+        "archive-type": "tar-gzip",
+        "url": "https://static.crates.io/crates/num-bigint/num-bigint-0.4.6.crate",
+        "sha256": "a5e44f723f1133c9deac646763579fdb3ac745e418f2a7af9cd0c431da1f20b9",
+        "dest": "cargo/vendor/num-bigint-0.4.6"
+    },
+    {
+        "type": "archive",
+        "archive-type": "tar-gzip",
+        "url": "https://static.crates.io/crates/num-integer/num-integer-0.1.46.crate",
+        "sha256": "7969661fd2958a5cb096e56c8e1ad0444ac2bbcd0061bd28660485a44879858f",
+        "dest": "cargo/vendor/num-integer-0.1.46"
+    },
+    {
+        "type": "archive",
+        "archive-type": "tar-gzip",
+        "url": "https://static.crates.io/crates/num-traits/num-traits-0.2.19.crate",
+        "sha256": "071dfc062690e90b734c0b2273ce72ad0ffa95f0c74596bc250dcfd960262841",
+        "dest": "cargo/vendor/num-traits-0.2.19"
+    },
+    {
+        "type": "archive",
+        "archive-type": "tar-gzip",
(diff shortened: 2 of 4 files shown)


R-Goc:
Builds fine with this diff on windows:
```diff
diff --git a/Libraries/LibJS/CMakeLists.txt b/Libraries/LibJS/CMakeLists.txt
index ea7918203f..44063811c0 100644
--- a/Libraries/LibJS/CMakeLists.txt
+++ b/Libraries/LibJS/CMakeLists.txt
@@ -296,7 +296,7 @@ endif()

 if (WIN32)
     # FIXME: Windows on ARM
-    target_link_libraries(LibJS PRIVATE clang_rt.builtins-x86_64.lib)
+    target_link_libraries(LibJS PRIVATE clang_rt.builtins-x86_64.lib kernel32 ntdll Ws2_32 userenv)
 else()
     # This flag has no effect on Windows
     target_compile_options(LibJS PRIVATE -fno-omit-frame-pointer)
```
That way we remove platform difference and option, and have the same thing running in CI. Might need installing rustup if it's not there yet on the windows image.

Libraries/LibJS/Rust/src/bytecode/operand.rs
```diff
@@ -0,0 +1,172 @@
...
+/// Environment coordinate used as a mutable cache in some instructions.
+/// Layout: two `u32` fields (hops + index).
+#[derive(Debug, Clone, Copy)]
+pub struct EnvironmentCoordinate {
+    pub hops: u32,
+    pub index: u32,
+}
+
+impl EnvironmentCoordinate {
+    pub const INVALID: u32 = 0xFFFF_FFFE;
```
dbearden, line 164:
```suggestion
    pub const INVALID: u32 = 0xFFFF_FFFF;
```

Libraries/LibJS/Rust/src/bytecode/basic_block.rs
```diff
@@ -0,0 +1,53 @@
...
+            resolved_this: false,
+        }
+    }
+
+    pub fn append(&mut self, instruction: Instruction, source_map: SourceMapEntry) {
+        let is_terminator = instruction.is_terminator();
+        self.instructions.push((instruction, source_map));
+        if is_terminator {
+            self.terminated = true;
+        }
```
rgret-dev, line 47:
```suggestion
        self.terminated = instruction.is_terminator();
        self.instructions.push((instruction, source_map));
```

zhangzhuang15:
@awesomekling I am pretty excited about bringing Rust code to this repo. Rust code is more friendly to me and help me  understand LibJS well. Thanks for your work. By the way, do you have plan to implement other parts with Rust and drop all cpp code in future ? It seems that this repo will depend both cpp and Rust forever, isn't it ?

meowette:
Maybe I am misunderstanding what "truly independent" means, but by what means is this still truly independent if LLM's are used to generate (mediocre, non idiomatic) code? If the whole point of this project to write everything from the ground up or at least not use code from other browsers, how can you guarantee that if you have it generated where you have no idea where the code is coming from? Also how is this going to work with copyright? This is not *your* creation, its not even the LLM's own code because it cannot 'invent' but at most rewrite what it saw somewhere else. Really disappointing, if you actually cared about rewriting something in Rust in a meaningful way you would have done it without depending on the use of LLMs.

zhangzhuang15:
I've received your message, and will respond  as soon as possible.

xnacly:
> Maybe I am misunderstanding what "truly independent" means, but by what means is this still truly independent if LLM's are used to generate (mediocre, non idiomatic) code? If the whole point of this project to write everything from the ground up or at least not use code from other browsers, how can you guarantee that if you have it generated where you have no idea where the code is coming from? Also how is this going to work with copyright? This is not _your_ creation, its not even the LLM's own code because it cannot 'invent' but at most rewrite what it saw somewhere else. Really disappointing, if you actually cared about rewriting something in Rust in a meaningful way you would have done it without depending on the use of LLMs.

The pr isnt even open anymore? Also it literally says its a one to one port from c++ to rust, both here, in the code and in the blog announcing the change. The reasoning is in the anouncement. Also id say its their creation, since, again, its a one to one port.

PR #8108, opened 2026-02-23
AK: Add and use `IdentityHashTraits<Integral>`

gmta:
These new traits are identical to `Traits<Integral>`, except that calling `.hash()` will return the value itself instead of hashing it. This should be used in cases where either the value is already a proper hash, or using the value as a hash will yield "good enough" performance in e.g. HashTable.

Types larger than 32 bits are folded in on themselves. Collision tests on some popular hashing algorithms show that XOR folding slightly increases the number of collisions, but this allows `IdentityHashTraits` not to make any assumptions on which bits are the most relevant for the final hash.

### DIFF
--- a/AK/CircularBuffer.h
+++ b/AK/CircularBuffer.h
@@ -95,7 +95,7 @@ private:
 
     SearchableCircularBuffer(ByteBuffer);
 
-    HashMap<unsigned, size_t> m_hash_location_map;
+    HashMap<unsigned, size_t, IdentityHashTraits<unsigned>> m_hash_location_map;
     HashMap<size_t, size_t> m_location_chain_map;
 
     ErrorOr<void> insert_location_hash(ReadonlyBytes value, size_t raw_offset);
--- a/AK/Traits.h
+++ b/AK/Traits.h
@@ -83,9 +83,21 @@ struct Traits<T> : public DefaultTraits<T> {
     static constexpr bool is_trivially_serializable() { return Traits<UnderlyingType<T>>::is_trivially_serializable(); }
 };
 
+template<Integral T>
+struct IdentityHashTraits : public Traits<T> {
+    static constexpr unsigned hash(T value)
+    {
+        if constexpr (sizeof(T) <= 4)
+            return static_cast<unsigned>(value);
+        else
+            return static_cast<unsigned>(value ^ (value >> 32));
+    }
+};
+
 }
 
 #if USING_AK_GLOBALLY
 using AK::DefaultTraits;
+using AK::IdentityHashTraits;
 using AK::Traits;
 #endif
--- a/Libraries/LibRegex/RegexMatcher.cpp
+++ b/Libraries/LibRegex/RegexMatcher.cpp
@@ -514,13 +514,6 @@ private:
     Node* m_last { nullptr };
 };
 
-struct SufficientlyUniformValueTraits : DefaultTraits<u64> {
-    static constexpr unsigned hash(u64 value)
-    {
-        return (value >> 32) ^ value;
-    }
-};
-
 template<class Parser>
 Matcher<Parser>::ExecuteResult Matcher<Parser>::execute(MatchInput const& input, MatchState& state, size_t& operations) const
 {
@@ -566,7 +559,7 @@ Matcher<Parser>::ExecuteResult Matcher<Parser>::execute(MatchInput const& input,
     }
 
     BumpAllocatedLinkedList<MatchState> states_to_try_next;
-    HashTable<u64, SufficientlyUniformValueTraits> seen_state_hashes;
+    HashTable<u64, IdentityHashTraits<u64>> seen_state_hashes;
 #if REGEX_DEBUG
     size_t recursion_level = 0;
 #endif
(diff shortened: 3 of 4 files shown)


alimpfard:
Actually, looks like the original 64b hash function itself has ~4k collisions already, we're gonna need to fix that but I doubt discarding 32 bits will help (verifying now).

gmta:
> > since the assumption is that all the value's bits are uniformly suitable to use as a hash
> 
> I'm not sold on this, why is that assumption true?

If all bits are uniformly suitable for hash purposes, XORring them will not change the resulting entropy of the `u32`.

Or if you mean why that assumption is there in the first place, that's the entire reason something like `IdentityHashTraits<Integral>` exists: you put in something that's suitable to use as a hash, such as a hash!

gmta:
> Actually, looks like the original 64b hash function itself has ~4k collisions already, we're gonna need to fix that but I doubt discarding 32 bits will help (verifying now).

If it's performance critical, it might help to pull in pair_int_hash anyways? Or build a custom version that's closer to what we use by default.

alimpfard:
> If all bits are uniformly suitable for hash purposes, XORring them will not change the resulting entropy of the u32.

but that's backwards from where we are, the claim to disprove is that all bits are not uniformly random.
Either way, this is fine, I will change the regex state hash fn anyway.

For reference, a less painful hash function:
```diff
--- a/Libraries/LibRegex/RegexMatch.h
+++ b/Libraries/LibRegex/RegexMatch.h
@@ -534,12 +534,16 @@ struct MatchState {
     {
         u64 hash = 0xcbf29ce484222325;
         auto combine = [&hash](auto value) {
-            hash ^= value + 0x9e3779b97f4a7c15 + (hash << 6) + (hash >> 2);
+            hash ^= static_cast<u64>(value);
+            hash *= 0x100000001b3;
         };
         auto combine_vector = [&hash](auto const& vector, auto tag) {
-            hash ^= tag * (vector.size() + 1);
+            hash ^= tag;
+            hash *= 0x100000001b3;
+            hash ^= vector.size();
+            hash *= 0x100000001b3;
             for (auto& value : vector) {
-                hash ^= value;
+                hash ^= static_cast<u64>(value);
                 hash *= 0x100000001b3;
             }
         };
```

ends up with collisions on sp=74, ip=360 and sp=28, ip=303 if you don't fold the u64, zero otherwise (in the range that we care about)

Hendiadyoin1:
Still advertising for fibonacci hashing, which just uses a multiplicand of `0x9e3779b97f4a7c15ULL` or respectively `0x9e3779b9U`

Edit:
Also it technically want's us to use the upper bits not the lower ones, not sure how important that is though

gmta:
@alimpfard @Hendiadyoin1 looking at collisions for SP=0..100,000 and IP=0..10,000, doing a `static_cast<u32>` actually slightly _improves_ on the number of collisions with the current hash algo compared to the XOR fold we're doing right now.

```code
  ┌─────────────────────┬─────────────┬─────────────┬─────────────┬─────────────┐
  │       Version       │     u64     │   low 32    │   high 32   │  XOR fold   │
  ├─────────────────────┼─────────────┼─────────────┼─────────────┼─────────────┤
  │ Boost-style combine │ 611,022,164 │ 625,678,996 │ 999,999,564 │ 628,131,571 │
  ├─────────────────────┼─────────────┼─────────────┼─────────────┼─────────────┤
  │ FNV-1a prime        │ 0           │ 162,842,336 │ 107,874,948 │ 107,903,387 │
  ├─────────────────────┼─────────────┼─────────────┼─────────────┼─────────────┤
  │ Golden ratio        │ 0           │ 229,704,960 │ 107,864,926 │ 107,894,404 │
  └─────────────────────┴─────────────┴─────────────┴─────────────┴─────────────┘
```

Since just using `static_cast<u32>` looks like a bad idea, I'm going with Ali's version ~~and taking the upper 32 bits for now~~ and apply the XOR fold - the increase in collisions with folding is insignificant, the performance is identical to shifting the upper 32 bits down, but it allows `IdentityHashTrait<Integral>` to use all 64 bits instead of assuming the upper 32 bits are the most relevant.

Hendiadyoin1:
I love how just a very specific multiply matches FNV....

gmta:
Disregard last comment, not sure why I'd gone for the FNV-1a if the XOR fold puts the golden ratio on top in terms of realistic collisions. Reworked matrix for SP range: 0..300, IP range: 0..750, reps range: 0..50 (total: 11,250,000):

```
  ┌─────────────────────┬───────┬────────┬─────────┬──────────┐
  │       Version       │  u64  │ low 32 │ high 32 │ XOR fold │
  ├─────────────────────┼───────┼────────┼─────────┼──────────┤
  │ Boost-style combine │ 7,918 │ 17,152 │ 22,212  │ 22,676   │
  ├─────────────────────┼───────┼────────┼─────────┼──────────┤
  │ FNV-1a prime        │ 0     │ 24,784 │ 14,606  │ 14,862   │
  ├─────────────────────┼───────┼────────┼─────────┼──────────┤
  │ Golden ratio        │ 0     │ 41,712 │ 14,847  │ 14,478   │
  └─────────────────────┴───────┴────────┴─────────┴──────────┘
```

alimpfard (approved):
LGTM!

PR #8110, opened 2026-02-23
LibWeb/CSS: Improve font-loading situation

AtkinsSJ:
Our font-loading has grown organically over the years and ended up a bit wonky. One consequence of this is we didn't properly wait for fonts to load, either not waiting long enough or waiting forever in some cases.

The end result of this PR is:
- Fonts load through one path regardless of whether they were `@font-face` or JS-created `FontFace` objects.
- `FontFaceSet` now knows what its fonts are doing, filling in a bunch of FIXMEs where `FontFace` tells any sets it is in to move it to a different category and update its state.
- `document.fonts.ready` now fulfils at the right time, when fonts have loaded/failed.
- `@font-face` is a critical subresource for its style sheet, delaying the load/error event for it.
- Tests that imported fonts, or waited for `document.fonts.ready`, are now much more stable and I've re-enabled a few that I identified in TestConfig. (There could be others.)

Locally I tested this on top of the changes in #8103 and didn't get timeouts.

### DIFF
--- a/Tests/LibWeb/TestConfig.ini
+++ b/Tests/LibWeb/TestConfig.ini
@@ -315,10 +315,8 @@ Text/input/wpt-import/hr-time/timeOrigin.html
 Text/input/window-scrollTo.html
 
 ; Flaky on CI
-Ref/input/unicode-range.html
 Text/input/Crypto/SubtleCrypto-exportKey.html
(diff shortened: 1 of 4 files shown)


AtkinsSJ:
Changes: Rebased, after my `font-weight` change caused a conflict. (Oops)

PR #8115, opened 2026-02-23
LibCore: Add Core::System::fsync()

RubenKelevra:
Add fsync() to LibCore system library for POSIX and with special handling for Windows


trflynn89:
What is this for? We don't add unused code.

RubenKelevra:
Hey @trflynn89,

I thought I split up my local search engine project and it will write out database files which need to be fsynced. If I should better include this in a larger PR, let me know.

Thanks!

PR #8116, opened 2026-02-23
LibCore: Let is_socket() detect pipes on Windows correctly

RubenKelevra:
Windows uses the same handle type (`FILE_TYPE_PIPE`) for both: network sockets and regular pipes. The old code assumed pipes as sockets.

To detect them properly, we use Winsock to read socket option (`getsockopt`). If the handle is just a regular pipe, this check will fail. If Winsock hasn't yet been initilialized, we temporarily start it up, to avoid checks fail because of the missing precondition.

### DIFF
--- a/Libraries/LibCore/SystemWindows.cpp
+++ b/Libraries/LibCore/SystemWindows.cpp
@@ -230,8 +230,34 @@ ErrorOr<int> dup(int handle)
 
 bool is_socket(int handle)
 {
-    // FILE_TYPE_PIPE is returned for sockets and pipes. We don't use Windows pipes.
-    return GetFileType(to_handle(handle)) == FILE_TYPE_PIPE;
+    // Safeguard against POSIX-style error values (e.g. -1) being passed down into Windows API
+    if (handle < 0)
+        return false;
+
+    auto os_handle = to_handle(handle);
+
+    // Since we bypass the CRT translation layer on Windows, the provided integer is a raw OS handle
+    auto file_type = GetFileType(os_handle);
+
+    // Fast-path: Windows classifies sockets as FILE_TYPE_PIPE
+    // We can immediately reject handles for files, consoles or 'unknown' to avoid unnecessary Winsock overhead
+    if (file_type != FILE_TYPE_PIPE)
+        return false;
+
+    // Windows cannot distinguish between anonymous pipes and sockets on handle level
+    // Thus we rely on a getsockopt() to make sure we don't treat a standard IPC pipe as a network socket
+    int socket_type = 0;
+    int option_length = sizeof(socket_type);
+    auto socket = static_cast<SOCKET>(reinterpret_cast<uintptr_t>(os_handle));
+    auto const result = ::getsockopt(socket, SOL_SOCKET, SO_TYPE, reinterpret_cast<char*>(&socket_type), &option_length);
+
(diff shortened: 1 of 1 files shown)


Libraries/LibCore/SystemWindows.cpp
```diff
@@ -230,8 +230,47 @@ ErrorOr<int> dup(int handle)
...
+        auto const result = ::getsockopt(socket, SOL_SOCKET, SO_TYPE, reinterpret_cast<char*>(&out_socket_type), &out_option_length);
+        return result != SOCKET_ERROR && out_option_length == static_cast<int>(sizeof(out_socket_type));
+    };
+
+    if (is_socket_probe_succeeded(socket_type, option_length))
+        return true;
+
+    // If the application hasn't initialized Winsock yet, the probe will fail with WSANOTINITIALISED.
+    // We temporarily initialize it here to ensure accurate detection regardless of the application's current state.
+    if (WSAGetLastError() != WSANOTINITIALISED)
```
R-Goc: This can never happen. Please remove this whole code path.

R-Goc:
Not sure if this makes much sense. There is not a single pipe in our code currently. I did find another way to do it previously which was calling NtQueryObject with ObjectNameInformation. An unnamed socket gives back \Device\Afd while an unnamed pipe fails with "The specified path is invalid."

Libraries/LibCore/SystemWindows.cpp
RubenKelevra: Done

RubenKelevra:
> Not sure if this makes much sense. There is not a single pipe in our code currently.

If the code base is not using pipes, it's still a good habit to check that our sockets are sockets and not pipes.

>  I did find another way to do it previously which was calling NtQueryObject with ObjectNameInformation. An unnamed socket gives back \Device\Afd while an unnamed pipe fails with "The specified path is invalid."

I'm not familiar with this, but looking it up: Microsoft marks winternl.h APIs as "This [...] may be changed or removed from Windows without further notice."

RubenKelevra:
@R-Goc have another look, please. :)

RubenKelevra:
```
Fetching all references...
Failed to download sources: module skia: Git commit for branch chrome/m144 is 7911bee5d90e994b6dde508a508576b9fded69d2, but expected ee20d565acb08dece4a32e3f209cdd41119015ca
Error: Build failed: Error: The process '/app/bin/xvfb-run' failed with exit code 1
```

CI failure is not related to my changes.

RubenKelevra:
Obviously not stale.

RubenKelevra:
Obviously not stale

Zaggy1024:
@R-Goc did you want to have another look at this? Or is it not relevant yet and/or not testable?

stasoid:
This PR bloats a two-line function into 28-line one to solve a problem that doesn't exist.
Even if such problem (distinguishing between sockets and pipes) ever arises, it is still can be solved in two lines:
```cpp
bool is_socket(int fd)
{
    int val, len = sizeof(val);
    return !::getsockopt(fd, SOL_SOCKET, SO_TYPE, (char*)&val, &len);
}
```
When I wrote this function I deliberately chose the `return GetFileType(to_handle(handle)) == FILE_TYPE_PIPE;` over the above code because it is simpler and does the job.

Please don't merge.

R-Goc:
I agree with stasoid. It has not use at this point, and is a lot of code for no reason.

gmta:

PR #8118, opened 2026-02-23
AK: Fix handling of getting the current thread ID being unsupported

cqundefine:


jonbgamble:
Why don't we just use `std::this_thread::get_id()`?

cqundefine:
> Why don't we just use `std::this_thread::get_id()`?

I feel like in general we avoid using the STL everywhere, tho maybe here it would be justified, cc @Zaggy1024 who originally wrote this code.

Zaggy1024:
I didn't look to use any std function for that reason, yeah. The original code also didn't use that, and that gave us the benefit of being guaranteed parity between our pids and tids in the original code (only on Linux though) which is a nice property to keep where possible.

PR #8136, opened 2026-02-24
LibWeb: Fix custom property invalidation through intermediate elements

awesomekling:
When an element's custom property changes (e.g., via JavaScript's
`style.setProperty()`), the style update tree walk only fully recomputes
elements that use `var()`. Intermediate elements that don't use var()
were skipped, leaving their `CustomPropertyData` chains pointing to
stale parent data. This caused descendants that DO use `var()` to
inherit outdated values, since they looked up the chain through the
stale intermediate element.

Fix this by refreshing the custom property data for non-recomputed
elements during the tree walk when custom properties have changed in
an ancestor. This is a lightweight operation (just a `RefPtr` update)
for the common case where the element has no own custom properties.

### DIFF
--- a/Libraries/LibWeb/DOM/Document.cpp
+++ b/Libraries/LibWeb/DOM/Document.cpp
@@ -1629,8 +1629,16 @@ void Document::update_layout(UpdateLayoutReason reason)
         auto& element = static_cast<Element&>(node);
         if (needs_full_style_update || node.needs_style_update() || parent_display_changed || (recompute_elements_depending_on_custom_properties && element.style_uses_var_css_function())) {
             node_invalidation = element.recompute_style(did_change_custom_properties);
-        } else if (needs_inherited_style_update) {
-            node_invalidation = element.recompute_inherited_style();
+        } else {
+            if (recompute_elements_depending_on_custom_properties) {
+                // Even if this element wasn't fully recomputed, refresh its
+                // custom property data chain so that descendants that use var()
+                // inherit up-to-date values through intermediate elements.
+                element.refresh_custom_property_data();
+            }
+            if (needs_inherited_style_update) {
+                node_invalidation = element.recompute_inherited_style();
+            }
         }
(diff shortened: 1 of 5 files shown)


Libraries/LibWeb/DOM/Element.cpp
```diff
@@ -3575,6 +3575,19 @@ void Element::set_custom_property_data(Optional<CSS::PseudoElement> pseudo_eleme
...
 
+void Element::refresh_custom_property_data()
+{
+    auto inherit_parent = element_to_inherit_style_from({});
+    RefPtr<CSS::CustomPropertyData const> parent_data;
+    if (inherit_parent)
+        parent_data = inherit_parent->custom_property_data({});
+
+    if (m_custom_property_data.ptr() == parent_data.ptr())
+        return;
```
gmta, line 3586: We're assigning a `RefPtr` below, which if you `move()` into it prevents unnecessary churn anyway and we don't need this extra check.

Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -1629,8 +1629,16 @@ void Document::update_layout(UpdateLayoutReason reason)
         auto& element = static_cast<Element&>(node);
         if (needs_full_style_update || node.needs_style_update() || parent_display_changed || (recompute_elements_depending_on_custom_properties && element.style_uses_var_css_function())) {
             node_invalidation = element.recompute_style(did_change_custom_properties);
-        } else if (needs_inherited_style_update) {
-            node_invalidation = element.recompute_inherited_style();
+        } else {
+            if (recompute_elements_depending_on_custom_properties) {
+                // Even if this element wasn't fully recomputed, refresh its
+                // custom property data chain so that descendants that use var()
+                // inherit up-to-date values through intermediate elements.
```
gmta, line 1636: ...personal nit but we really need to get that column 120 wrapping in somehow 😅

PR #8143, opened 2026-02-24
test-web: Use the output StringBuilder to measure the counter length

Zaggy1024:
Formatting the counter into a separate string is an unnecessary allocation.


Zaggy1024:
Oops, realized I put an unrelated change to s_output_captures in that commit. Oh well, the change is fine :^)

PR #8144, opened 2026-02-24
LibWeb: Support SecureContext for Workers

shannonbooth:
Expands on the work done in #7518 to implement secure contexts, now without the kludge of 00911a8b0117b14c4dd11601fafc0f6ca14fe04e.

After this PR we should support not exposing SecureContext interfaces in non secure worker contexts too (alongside the other implications of 'is a secure context')

### DIFF
--- a/Libraries/LibWeb/HTML/Scripting/Environments.cpp
+++ b/Libraries/LibWeb/HTML/Scripting/Environments.cpp
@@ -547,10 +547,9 @@ JS::Object& entry_global_object()
 bool is_secure_context(Environment const& environment)
 {
     // 1. If environment is an environment settings object, then:
-    if (is<EnvironmentSettingsObject>(environment)) {
+    if (auto const* environment_settings_object = as_if<EnvironmentSettingsObject>(environment)) {
         // 1. Let global be environment's global object.
-        // FIXME: Add a const global_object() getter to ESO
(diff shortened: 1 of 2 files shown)


Libraries/LibWeb/HTML/Scripting/SerializedEnvironmentSettingsObject.h
```diff
@@ -22,13 +22,15 @@ enum class CanUseCrossOriginIsolatedAPIs : u8 {
 
 struct SerializedDocument {
     URL::URL url;
+    bool relevant_settings_object_is_secure_context {};
```
tcl3: nit: Can we explicitly initialize this to false.

Libraries/LibWeb/HTML/Scripting/SerializedEnvironmentSettingsObject.h
```diff
@@ -22,13 +22,15 @@ enum class CanUseCrossOriginIsolatedAPIs : u8 {
...
 };
 
 struct SerializedWindow {
     SerializedDocument associated_document;
 };
 
 struct SerializedWorkerGlobalScope {
+    bool relevant_settings_object_is_secure_context {};
```
tcl3: nit: Can we explicitly initialize this to false.

PR #8150, opened 2026-02-25
Docs: Add cargo+rustc to Fedora build instructions

Ali-002:
These packages are required for a full build on Fedora but were missing from the installation instructions. Noticed this while setting up a non-development machine.


trflynn89:
There is a note at the very top of the build instructions:
https://github.com/LadybirdBrowser/ladybird/blob/1453add36176fc16a992ad592b431e5b47942ed0/Documentation/BuildInstructionsLadybird.md#build-prerequisites
> A Rust toolchain is also required for building the JavaScript engine. You can install it via [rustup](https://rustup.rs/) or your system's package manager.

PR #8155, opened 2026-02-25
RequestServer: Capture async references to Request objects weakly

trflynn89:
It's possible for the client (WebContent) to stop a request while the request is waiting for an async callback to be invoked. So we cannot assume the request itself will still be alive once these callbacks are finally invoked.

Fixes #8154

PR #8158, opened 2026-02-25
AK+LibGC: Introduce `weak_callback()`

gmta:
Neat little template to make working with WeakPtr/strong refs and GC::Weak a bit easier on the eyes.

### DIFF
--- a/Libraries/LibGC/Weak.h
+++ b/Libraries/LibGC/Weak.h
@@ -148,6 +148,17 @@ private:
     NonnullRefPtr<WeakImpl> m_impl { WeakImpl::the_null_weak_impl };
 };
 
+// NOTE: Unlike GC::Function, captures in this callback are not visited by the garbage collector.
+//       Do not capture GC-managed objects without ensuring they are kept alive through other means.
+template<typename T, typename Callback>
+auto weak_callback(T& obj, Callback&& callback)
+{
+    return [weak = Weak<T> { obj }, cb = forward<Callback>(callback)](auto&&... args) mutable {
+        if (weak)
+            cb(*weak, forward<decltype(args)>(args)...);
+    };
+}
+
 template<typename T, typename U>
 inline bool operator==(Weak<T> const& a, Ptr<U> const& b)
 {
--- a/Libraries/LibWeb/WebSockets/WebSocket.cpp
+++ b/Libraries/LibWeb/WebSockets/WebSocket.cpp
@@ -221,30 +221,18 @@ ErrorOr<void> WebSocket::establish_web_socket_connection(URL::URL const& url_rec
 
     m_websocket = request_client->websocket_connect(url_record, origin_string, protocol_byte_strings, {}, additional_headers);
 
-    m_websocket->on_open = [weak_this = GC::Weak { *this }] {
-        if (!weak_this)
-            return;
-        auto& websocket = const_cast<WebSocket&>(*weak_this);
-        websocket.on_open();
-    };
-    m_websocket->on_message = [weak_this = GC::Weak { *this }](auto message) {
-        if (!weak_this)
-            return;
-        auto& websocket = const_cast<WebSocket&>(*weak_this);
-        websocket.on_message(move(message.data), message.is_text);
-    };
-    m_websocket->on_close = [weak_this = GC::Weak { *this }](auto code, auto reason, bool was_clean) {
-        if (!weak_this)
-            return;
-        auto& websocket = const_cast<WebSocket&>(*weak_this);
-        websocket.on_close(code, String::from_byte_string(reason).release_value_but_fixme_should_propagate_errors(), was_clean);
-    };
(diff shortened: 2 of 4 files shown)


AK/WeakPtr.h
```diff
@@ -173,6 +173,15 @@ class [[nodiscard]] WeakPtr {
     RefPtr<WeakLink> m_link;
 };
 
+template<typename T, typename Callback>
+auto weak_callback(T& obj, Callback&& callback)
+{
+    return [weak = obj.template make_weak_ptr<T>(), cb = forward<Callback>(callback)](auto&&... args) {
+        if (auto strong = weak.strong_ref())
```
trflynn89: In #8155, the `Request` type is not ref-counted, so `strong_ref` would fail to compile. Something like this seems to work:
```c++
template<typename T, typename Callback>
static auto weak_callback(T& obj, Callback&& callback)
{
    return [weak = obj.template make_weak_ptr<T>(), callback = forward<Callback>(callback)](auto&&... args) {
        if constexpr (requires { obj.ref(); obj.unref(); }) {
            if (auto strong = weak.strong_ref())
                callback(*strong, forward<decltype(args)>(args)...);
        } else {
            if (weak)
                callback(*weak, forward<decltype(args)>(args)...);
        }
    };
}
```

trflynn89: Can we also add a test to `TestWeakPtr.cpp` for this?

Libraries/LibGC/Weak.h
```diff
@@ -148,6 +148,15 @@ class Weak {
     NonnullRefPtr<WeakImpl> m_impl { WeakImpl::the_null_weak_impl };
 };
 
+template<typename T, typename Callback>
+auto weak_callback(T& obj, Callback&& callback)
```
trflynn89, line 154: Maybe add a comment that, unlike `GC::Function`, captures are not safe from GC. The usage here is all fine ofc, but I can imagine someone grabbing this in the future thinking it protects captures the same way as `GC::Function`.

AK/WeakPtr.h
Zaggy1024: I think it may not even be necessary to strong ref the weak pointer here? WeakPtr isn't thread-safe, so this could end up with a TOCTOU race if the object is refed from multiple threads.

I suppose we could end up unrefing the object within the callback, but that seems like a pretty rare need.

trflynn89: Given that the immediate use cases here all wanted a strong reference and are single threaded, we could expand the requirement clause to also check that the type isn't `AtomicRefCounted`?

Zaggy1024: That definitely seems worthwhile.

Libraries/LibWeb/CSS/StyleValues/ImageStyleValue.cpp
```diff
@@ -69,21 +69,21 @@ void ImageStyleValue::load_any_resources(DOM::Document& document)
...
-                auto image_data = m_resource_request->image_data();
+                auto image_data = self.m_resource_request->image_data();
                 if (image_data->is_animated() && image_data->frame_count() > 1) {
-                    m_timer = Platform::Timer::create(m_document->heap());
-                    m_timer->set_interval(image_data->frame_duration(0));
-                    m_timer->on_timeout = GC::create_function(m_document->heap(), [this] { animate(); });
-                    m_timer->start();
+                    self.m_timer = Platform::Timer::create(self.m_document->heap());
+                    self.m_timer->set_interval(image_data->frame_duration(0));
+                    self.m_timer->on_timeout = GC::create_function(self.m_document->heap(), [&self] { self.animate(); });
```
trflynn89: Ooh just noticed this - `&self` probably isn't safe right? It's a reference to a local in something that executes async. Maybe it can keep capturing `this` since it's owned by `this` anyways?

AK/WeakPtr.h
gmta: * Added a constexpr requirement of the type being `RefCountedBase<T>` before taking a stong ref.
* Added some tests

Libraries/LibGC/Weak.h
gmta: Sure!

Libraries/LibWeb/CSS/StyleValues/ImageStyleValue.cpp
gmta: Yeah 😅 good catch.

Libraries/LibWeb/WebSockets/WebSocket.cpp
```diff
@@ -221,30 +221,18 @@ ErrorOr<void> WebSocket::establish_web_socket_connection(URL::URL const& url_rec
 
     m_websocket = request_client->websocket_connect(url_record, origin_string, protocol_byte_strings, {}, additional_headers);
 
-    m_websocket->on_open = [weak_this = GC::Weak { *this }] {
```
trflynn89, line 224: This file's diff is just :yakkiss:

PR #8163, opened 2026-02-25
LibJS: added minimal simple tests for lexer

xD-saleem:


xnacly:
I think there is already a test suite, since the rust pipeline is tested against https://github.com/tc39/test262 so i dont know if this is needed

trflynn89:
We execute our own `test-js` suite and the official test262 harness already. Again, please coordinate on discord before blindly opening PRs. We don't want people to waste their time and effort on things that won't be merged.

PR #8165, opened 2026-02-25
LibJS: Refactor parts of the rust lexer to be more idiomatic

basicallygit:
Add a `consume_while(condition)` method to the lexer which allows for much cleaner handling of lexing non-tokens instead of nested `loop { break; }`'s

Also switch to using match instead of many if statements for `keyword_from_str` in the keyword parsing


basicallygit:
whoops forgot to rustfmt

Issue #2412, opened 2024-11-18
LibJS+LibGC: After LibGC refactor, it is even more unsafe to destroy a VM


ADKaster:
The LibWeb main thread VM is never destroyed. This has been the case for a long time. We have too many cycles in core objects.

After the LibGC refactor, it is now unsafe to destroy the VM in general.

This only seems to manifest in the Distribution build.

Repro:

```shell
cmake --preset Distribution -DCMAKE_C_COMPILER=clang-18 -DCMAKE_CXX_COMPILER=clang++-18
ninja -C Build/distribution js
./Build/distribution/bin/js -c "console.log('42')"
```

Output:

```
"42"
Segmentation fault (core dumped)
```

Backtrace:

```
"42"
VERIFICATION FAILED: m_ref_count at /home/andrew/ladybird-org/ladybird-browser/AK/RefCounted.h:47
/home/andrew/ladybird-org/ladybird-browser/Build/distribution/bin/js(+0x205d4df) [0x5555575b14df]
/home/andrew/ladybird-org/ladybird-browser/Build/distribution/bin/js(+0x22a8991) [0x5555577fc991]
/home/andrew/ladybird-org/ladybird-browser/Build/distribution/bin/js(+0x22a8c66) [0x5555577fcc66]
/home/andrew/ladybird-org/ladybird-browser/Build/distribution/bin/js(+0x22a02c7) [0x5555577f42c7]
/home/andrew/ladybird-org/ladybird-browser/Build/distribution/bin/js(+0x248eca6) [0x5555579e2ca6]
/home/andrew/ladybird-org/ladybird-browser/Build/distribution/bin/js(+0x248ac81) [0x5555579dec81]
/home/andrew/ladybird-org/ladybird-browser/Build/distribution/bin/js(+0x2489d9c) [0x5555579ddd9c]
/home/andrew/ladybird-org/ladybird-browser/Build/distribution/bin/js(+0x248922b) [0x5555579dd22b]
/home/andrew/ladybird-org/ladybird-browser/Build/distribution/bin/js(+0x23fb0a4) [0x55555794f0a4]
/home/andrew/ladybird-org/ladybird-browser/Build/distribution/bin/js(+0x2059866) [0x5555575ad866]
/lib/x86_64-linux-gnu/libc.so.6(+0x47a66) [0x7ffff7847a66]
/lib/x86_64-linux-gnu/libc.so.6(+0x47bae) [0x7ffff7847bae]
[...]

PR #8169, opened 2026-02-25
LibWeb: Exit WebContent cleanly on user quit

jonbgamble:
There are still a few `exit(0)`s in various "oh noes" scenarios, but the big one in EventLoopPluginSerenity is removed by this PR.

This was motivated by the constant ganesh ref count nags when exiting on debug builds. But it's bigger than that. We should require orderly shutdown on the happy path (and as many unhappy ones as we can stomach) moving forward.

### DIFF
--- a/Services/WebContent/main.cpp
+++ b/Services/WebContent/main.cpp
@@ -22,6 +22,7 @@
 #include <LibUnicode/TimeZone.h>
 #include <LibWeb/Bindings/MainThreadVM.h>
 #include <LibWeb/Fetch/Fetching/Fetching.h>
+#include <LibWeb/HTML/Navigable.h>
 #include <LibWeb/HTML/UniversalGlobalScope.h>
 #include <LibWeb/HTML/Window.h>
 #include <LibWeb/Internals/Internals.h>
@@ -55,7 +56,7 @@
 static ErrorOr<void> load_content_filters(StringView config_path);
 
 static ErrorOr<void> initialize_resource_loader(GC::Heap&, int request_server_socket);
-static ErrorOr<void> reinitialize_resource_loader(IPC::File const& image_decoder_socket);
+static ErrorOr<void> reinitialize_resource_loader(IPC::File const& request_server_socket);
 
 static ErrorOr<void> initialize_image_decoder(int image_decoder_socket);
 static ErrorOr<void> reinitialize_image_decoder(IPC::File const& image_decoder_socket);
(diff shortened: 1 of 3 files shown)


Services/WebContent/main.cpp
```diff
@@ -230,7 +231,12 @@ ErrorOr<int> ladybird_main(Main::Arguments arguments)
             dbgln("Failed to reinitialize image decoder: {}", result.error());
     };
 
-    return event_loop.exec();
+    int const exit_code = event_loop.exec();
+
+    Web::HTML::shutdown_all_navigables();
+    Web::Bindings::main_thread_vm().heap().collect_garbage(GC::Heap::CollectionType::CollectEverything);
```
shannonbooth, line 237: Not an extensive review by any means - but for what its worth we leak this intentionally to avoid destroying all GC objects on process exit

shannonbooth: If we _really_ wanted to tackle a ’clean’ shutdown I think we should be doing other restructuring in general (for example moving away from liberal use of static causing issues like https://github.com/LadybirdBrowser/ladybird/issues/2412) but so far the stance which we have taken is to just let the OS clean things up instead of spending time to do a _fully_ clean shutdown

jonbgamble: Maybe we should tackle a little at a time. First we can get rid of exit(0).

jonbgamble: I think we do want to tackle clean shutdown, and it's better to start sooner than later. Shitting the bed and letting the OS clean up may release kernel resources but it isn't good for cross-process state.

A modern browser should also make a best effort to handle in-flight IDB requests, localStorage flushes, etc. before terminating the process. I don't handle any of that in this PR, but it's a baby step towards it.

shannonbooth:
Based on the discussion in discord and elsewhere, I think we are moving in the opposite direction here - i.e _exit type of path, same as if we were to crash

jonbgamble:
That makes sense, and _exit is a natural place for process state cleanup code without ugly trapping.

Rather than close this immediately, I'm gonna go meta while I've got one of your ears.

Don't you think Ladybird maintainers should consider beginning to close PRs like this on the contributor's behalf? It is not offensive when accompanied by explanation. Any contributor with high volume project experience can tell you it's a kindness that saves time and energy on both sides.

Maybe you haven't seen this yet, but often popular projects get submissions from people who don't take hints so well. They don't close PRs or make them draft after polite "no thanks". Instead, they argue for them, just let them dangle, or worst of all - continue working on them.

This increases the open PR count which can cause indirect anxiety that actually discourage timely review. "Oh lawd, 127 open PRs.. I swear I'll pound out a few next week". Maybe this is not a problem now, but it's something to keep an eye on and consider.

As for this PR... You know what to do! (And I will do it if I don't hear back in a few days)

shannonbooth:
Yeah fair can close this, though on this topic I don't think this is something which until quite recently seems like more firm of the direction we're taking regarding process shutdown

PR #8171, opened 2026-02-25
LibJS: Return UTF16String's ASCII storage when applicable

trflynn89:
If we have a UTF-16 string with ASCII storage, we don't need to convert the string to UTF-8 to get a `StringView`. This fast-path is hit over 76k times during `test-js`.


trflynn89:
edit: grabbed test262 counts too for fun. We hit this fast-path 31,049,047 times.

PR #8172, opened 2026-02-25
LibWeb: Implement CSS field-sizing property

jackbodine:
This adds support for the 'field-sizing' CSS property. It supports both the 'fixed' (default) and 'content' values.

When set to 'content', the TextAreaBox ignores its 'cols' and 'rows' attributes and instead returns an empty auto-size, allowing the layout engine to size the element based on its content.

### DIFF
--- a/Libraries/LibWeb/Layout/TextAreaBox.cpp
+++ b/Libraries/LibWeb/Layout/TextAreaBox.cpp
@@ -1,10 +1,16 @@
 /*
  * Copyright (c) 2025-2026, Jonathan Gamble <gamblej@gmail.com>
+ * Copyright (c) 2026, Jack Bodine <jacktaylorbodine@gmail.com>
  *
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
 #include <LibWeb/Layout/TextAreaBox.h>
+#include <LibWeb/CSS/ComputedProperties.h>
+#include <LibWeb/CSS/PropertyID.h>
+#include <LibWeb/CSS/SerializationMode.h>
+#include <LibWeb/CSS/StyleValues/StyleValue.h>
+#include <LibWeb/DOM/Node.h>
 
 namespace Web::Layout {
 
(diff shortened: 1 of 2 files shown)


Psychpsyo:
You should add a test that verifies that this works the way it should and the code should probably cite the spec about what the property does: https://drafts.csswg.org/css-forms-1/#propdef-field-sizing

The spec mentions some stuff about this applying to compressible replaced elements, as well as the browser not treating those elements as replaced elements for min-content contributions. It's not exactly obvious whether or not this actually implements that.

WPT has quite a few tests for this that you can import: https://github.com/search?q=repo%3Aweb-platform-tests%2Fwpt+%22field-sizing%22&type=code

Libraries/LibWeb/Layout/TextAreaBox.cpp
```diff
@@ -17,6 +23,13 @@ TextAreaBox::TextAreaBox(DOM::Document& document, GC::Ptr<DOM::Element> element,
 
 CSS::SizeWithAspectRatio TextAreaBox::compute_auto_content_box_size() const
 {
+    auto const& style_properties = dom_node().computed_properties();
+
+    // If field-sizing: content, return empty size (auto) to fit the text content.
+    if (style_properties->property(CSS::PropertyID::FieldSizing).to_string(CSS::SerializationMode::Normal) == "content") {
```
Calme1709, line 29: Instead of `to_string(CSS::SerializationMode::Normal) == "content"` we can use `to_keyword() == CSS::Keyword::Content` here which will skip the (relatively) expensive serialization process. (FYI this will require to you include KeywordStyleValue.h if it's not already included transitively)

Libraries/LibWeb/Layout/TextAreaBox.cpp
```diff
@@ -1,10 +1,16 @@
...
  *
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
 #include <LibWeb/Layout/TextAreaBox.h>
+#include <LibWeb/CSS/ComputedProperties.h>
+#include <LibWeb/CSS/PropertyID.h>
+#include <LibWeb/CSS/SerializationMode.h>
+#include <LibWeb/CSS/StyleValues/StyleValue.h>
+#include <LibWeb/DOM/Node.h>
```
Calme1709, line 13: Can we get rid of a couple of these `#include` statements? `PropertyID.h` should be transitively included via `ComputedProperties.h`, likewise with `SerializationMode.h` and `StyleValue.h`

PR #7961, opened 2026-02-15
LibWeb/SVG: Implement SVGFEDisplacementMapElement

samu698:
This change partially fixes the background for https://endtimes.dev/why-your-website-should-be-under-14kb-in-size/

PR #8176, opened 2026-02-25
LibWeb/SVG: Implement the feDisplacementMap filter

Lubrsi:
Before:
Screenshot 2026-02-25 at 23 55 07

After:
Screenshot 2026-02-25 at 23 56 18


tcl3:
Supersedes #7961 

The linter is complaining that the added PNGs need to be run through `optipng`.

PR #8179, opened 2026-02-26
LibWeb+LibMedia: Fix flakes involving the media element

Zaggy1024:
There's a good few fixes/improvements in these changes that are explained in the individual commits. The gist of it is that the playback manager needed to be destroyed a bit earlier, but that caused a cascade of prerequisite fixes in order to make that safer. In the process, I've implemented decode error handling, as well as tightened up the rest of the error handling to follow spec more closely/clearly.

This should hopefully prevent most or all of the instances of the PlaybackManager panicking about disabling a track that is not present in its metadata, such as this stack trace:

```
VERIFICATION FAILED: m_ptr at /home/runner/_work/ladybird/ladybird/AK/RefPtr.h:280
#0  (inlined)          in AK::RefPtr<Media::DisplayingVideoSink>::as_nonnull_ptr() const at /home/runner/_work/ladybird/ladybird/AK/RefPtr.h:280:9
#1  (inlined)          in AK::RefPtr<Media::DisplayingVideoSink>::operator->() const at /home/runner/_work/ladybird/ladybird/AK/RefPtr.h:244:30
#2  0x00007eff6dcb9d08 in Media::PlaybackManager::remove_the_displaying_video_sink_for_track(Media::Track const&) at /home/runner/_work/ladybird/ladybird/Libraries/LibMedia/PlaybackManager.cpp:244:23
#3  0x00007effa20c2f78 in Web::HTML::HTMLMediaElement::set_selected_video_track(AK::Badge<Web::HTML::VideoTrack>, GC::Ptr<Web::HTML::VideoTrack>) at /home/runner/_work/ladybird/ladybird/Libraries/LibWeb/HTML/HTMLMediaElement.cpp:1272:71
#4  0x00007effa3392d51 in Web::HTML::VideoTrack::set_selected(bool) at /home/runner/_work/ladybird/ladybird/Libraries/LibWeb/HTML/VideoTrack.cpp:75:45
#5  0x00007effa20d9b1a in Web::HTML::HTMLMediaElement::on_video_track_added(Media::Track const&) at /home/runner/_work/ladybird/ladybird/Libraries/LibWeb/HTML/HTMLMediaElement.cpp:1394:34
#6  (inlined)          in operator()<Media::TrackType, const Media::Track> at /home/runner/_work/ladybird/ladybird/Libraries/LibWeb/HTML/HTMLMediaElement.cpp:1518:44
#7  0x00007effa20dcf2a in call at /home/runner/_work/ladybird/ladybird/AK/Function.h:224:30
```

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLMediaElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLMediaElement.cpp
@@ -1629,6 +1629,12 @@ void HTMLMediaElement::handle_media_source_failure(Span<GC::Ref<WebIDL::Promise>
     // 5. Fire an event named error at the media element.
     dispatch_event(DOM::Event::create(realm, HTML::EventNames::error));
 
+    // NB: If the error has been reset, that means we've entered load_element() within the error
+    //     event handler. If that's the case, as part of the resource selection algorithm, these
+    //     steps must abort.
+    if (!m_error)
+        return;
+
     // 6. Reject pending play promises with promises and a "NotSupportedError" DOMException.
     reject_pending_play_promises<WebIDL::NotSupportedError>(promises, "Media is not supported"_utf16);
 
--- a/Tests/LibWeb/Text/expected/HTML/HTMLMediaElement-load-during-error.txt
+++ b/Tests/LibWeb/Text/expected/HTML/HTMLMediaElement-load-during-error.txt
@@ -4,11 +4,11 @@ Events:
   error handler: setting good src: readyState=0
   error handler: src setter returned: readyState=0
   emptied: readyState=0
-  document load: readyState=0
   loadstart: readyState=0
   durationchange: readyState=4 duration=12.043
   loadedmetadata: readyState=4
   loadeddata: readyState=4
   canplay: readyState=4
   canplaythrough: readyState=4
+  document load: readyState=4
 Final: readyState=4 duration=12.043


Libraries/LibWeb/HTML/HTMLMediaElement.cpp
```diff
@@ -903,19 +903,15 @@ void HTMLMediaElement::select_resource()
...
+                queue_a_media_element_task([this, error_message = move(error_message)]() mutable {
                     auto promises = take_pending_play_promises();
                     handle_media_source_failure(promises, move(error_message));
-
-                    ran_media_element_task = true;
                 });
 
                 // 7. Wait for the task queued by the previous step to have executed.
-                HTML::main_thread_event_loop().spin_until(GC::create_function(heap(), [&]() { return ran_media_element_task; }));
+                // AD-HOC: All calls to the failure steps immediately return, so we do not actually need to wait here.
```
shannonbooth, line 922: From what I can gather, I think the point of the wait here is that running the failjre steps can invoke javascript which changes the state of future steps, like for example maybe the src attribute is changed by this point

Zaggy1024: Maybe I'm misunderstanding your point, but from what I could tell, our implementation of the resource selection algorithm doesn't contain any steps after failure.

shannonbooth: It may also be me misunderstanding, but the path I am thinking of is handle_media_source_failure dispatching an error event, and that error event having a javacript eventhandler that manipulates state of the element

Zaggy1024: I do need to see if there's anything we need to do about the SourceElementSelector when load_element() is possibly triggered again, but I think that's unrelated to this potential spinwait.

Libraries/LibWeb/HTML/HTMLMediaElement.cpp
```diff
@@ -734,8 +734,9 @@ class SourceElementSelector final : public JS::Cell {
             m_candidate->dispatch_event(DOM::Event::create(m_candidate->realm(), HTML::EventNames::error));
         });
 
-        // FIXME: 10. Await a stable state. The synchronous section consists of all the remaining steps of this algorithm until
-        //            the algorithm says the synchronous section has ended. (Steps in synchronous sections are marked with ⌛.)
+        // 10. Await a stable state. The synchronous section consists of all the remaining steps of this algorithm until
+        //     the algorithm says the synchronous section has ended. (Steps in synchronous sections are marked with ⌛.)
+        // NB: We're already in a stable state here.
```
shannonbooth: My understanding of the normative requirements here is a little different - my understanding is that this is not the specs only requirement here - https://html.spec.whatwg.org/#await-a-stable-state says that awaiting a stable state should be invoked inside of a queued micro task, which indicates to me that the timing of find_next_candidate is not immediate with respect to javascript

Libraries/LibWeb/HTML/HTMLMediaElement.cpp
shannonbooth: Oh, sorry, yup I was confused!

Libraries/LibWeb/HTML/HTMLMediaElement.cpp
Zaggy1024: It's not really clear to me what this expects us to do when we're already running synchronous code. Everything in HTMLMediaElement runs on the main thread, so we're always "stable", so the only concern would be what's observable from JavaScript, no?

In this particular case, if awaiting a stable state means queuing a microtask, that seems to imply the next steps could end up running before or after the error event fires, which feels like a problem given that the final step mutates the state of the element in an observable way. Or, it would be observable if there are any tracks found, but it seems as though those failure steps should only be executed in cases where the tracks _haven't_ been populated.

I could just leave all the stable state steps as fixmes, but I'd be interested in clarifying what exactly it means in the media element given that it's all synchronous with respect to JS.

Zaggy1024: Another thing worth noting is that, since we currently rely on the unique media element task source to stop in-progress tasks involved in the resource selection steps, using a microtask would make stopping those steps a bit more difficult, I think.

trflynn89: The spec authors have been trying to remove "await a stable state" for almost a decade now `https://github.com/whatwg/html/issues/2882`. I would just leave these as FIXMEs and maybe clarify with the spec what the intended behavior here is.

Libraries/LibWeb/HTML/HTMLMediaElement.cpp
```diff
@@ -1525,6 +1532,13 @@ void HTMLMediaElement::set_up_playback_manager(NonnullRefPtr<FetchData> const& f
         fetch_data->failure_callback(MUST(String::from_utf8(error.description())));
     };
 
+    // -> If the media data is corrupted
+    m_playback_manager->on_error = [weak_self = GC::Weak(*this)](Media::DecoderError&& error) {
```
trflynn89: This was just merged, but we have `weak_callback` for this now. There are some example usages in this file already.

Libraries/LibWeb/HTML/HTMLMediaElement.cpp
```diff
@@ -170,18 +170,24 @@ void HTMLMediaElement::set_decoder_error(String error_message)
...
+    // NB: This is only invoked by PlaybackManager after the metadata has been retrieved, so we should be sure that
+    //     the ready state indicates we have that metadata.
+    VERIFY(m_ready_state != ReadyState::HaveNothing);
 
     // 1. The user agent should cancel the fetching process.
     cancel_the_fetching_process();
 
     // 2. Set the error attribute to the result of creating a MediaError with MEDIA_ERR_DECODE.
-    m_error = realm.create<MediaError>(realm, MediaError::Code::Decode, move(error_message));
+    m_error = realm.create<MediaError>(realm, MediaError::Code::Decode, error_message);
```
trflynn89, line 190: Might be missing something, but why remove the `move`?

Libraries/LibWeb/HTML/HTMLMediaElement.cpp
```diff
@@ -1550,19 +1550,26 @@ void HTMLMediaElement::set_up_playback_manager(NonnullRefPtr<FetchData> const& f
...
+        dispatch_event(DOM::Event::create(realm, HTML::EventNames::progress));
 
         // Set the networkState to NETWORK_IDLE and fire an event named suspend at the media element.
         m_network_state = NetworkState::Idle;
-        dispatch_event(DOM::Event::create(this->realm(), HTML::EventNames::suspend));
+        dispatch_event(DOM::Event::create(realm, HTML::EventNames::suspend));
+    } else if (fetching_status == FetchingStatus::Ongoing) {
+        // If the user agent ever discards any media data and then needs to resume the network activity to obtain it again, then it must queue a media
+        // element task given the media element to set the networkState to NETWORK_LOADING.
+        queue_a_media_element_task([self = GC::Weak(*this)] {
```
trflynn89: same here (weak_callback) and in subsequent commits

Libraries/LibWeb/HTML/HTMLMediaElement.cpp
shannonbooth: > concern would be what's observable from JavaScript, no?

Most likely impact yeah, but I am not sure.

Even if we are running synchronously on the main thread the requirement of the spec is that the algorithm should only continue at the point a microtask checkpoint is invoked where all of the other javascript promises that are pending are also run - since it says:

> to await a stable state, the user agent must [queue a microtask](https://html.spec.whatwg.org/multipage/webappapis.html#queue-a-microtask) that runs the following steps, and must then stop executing (execution of the algorithm resumes when the microtask is run, as described in the following steps)

So for example, I think our implementation of https://html.spec.whatwg.org/multipage/media.html#dom-media-seek is wrong in the timing which it does things.

Currently, we queue tasks to fire events synchronously in the context of HTMLMediaElement::seek_element. This means that the timing of the invocation of that queued task is interleaved with pending promises IIUC. However, the spec seems to be saying that any javascript associated with promises should first be run before those tasks are run.

But this is all so complex I am finding it difficult to figure out the exact impacts of those timings, so I might be getting things wrong. But the note here saying that we are in a stable state does seem like it could be misleading at least.

Zaggy1024: Tim's link seems to imply that all microtasks should be associated with a prior task, and only run after that task. That makes me think that all "await a stable state" calls should instead use media element tasks? I don't think we consistently have tasks that we can associate the microtasks to currently, and even if we did, microtasks aren't guaranteed to run after the task queued before them in our implementation, from what I can tell.

Zaggy1024: In this particular case, would that mean we could just move these steps into the task firing the event? 🤔

Libraries/LibWeb/HTML/HTMLMediaElement.cpp
Zaggy1024: Ah, no real reason. I just ended up rewriting this method because I didn't notice we had an unused implementation of it, then copied it over top of that when I found it later.

Libraries/LibWeb/HTML/HTMLMediaElement.cpp
shannonbooth: Honestly I only ended up confusing myself 😅 I think what Tim mentioned about leaving the FIXME and figuring things out later is best for now

shannonbooth: (I also saw Tims comment after I commented)

Zaggy1024: I'm thinking more along the lines of writing an AD-HOC comment and moving those everything that annevk mentioned (where it mutates DOM) into media element tasks. Does that seem reasonable to you @trflynn89? Or do you see that HTML issue as more of an unanswered question than a prescribed fix?

Zaggy1024: For now, I'll move the steps into the task but also add a FIXME with a link to that comment. It'd probably be worth revisiting other places where we are supposed to await a stable state, as it seems like those should be done within tasks where relevant, or done synchronously otherwise.

Zaggy1024:
Discovered one more inconsistency with spec, we were allowing the resource selection algorithm to continue after setting `src` within an `error` handler that was fired from a previously-failed load through the `src` attribute. That was causing the document load event to stop being delayed, despite the media element being expected to synchronously delay the load event again after `src` was modified. The behavior now matches other browsers in this regard. I've added a test that shows the before/after of the event ordering there.

PR #8183, opened 2026-02-26
CI: Update rust ci to run on all targets

Sharktheone:
run cargo fmt and cargo cippy on `--all` / `--all-targets`

This also denies warnings in case there might be something that clippy::all doesn't catch.

### DIFF
--- a/Meta/lint-ci.sh
+++ b/Meta/lint-ci.sh
@@ -52,17 +52,17 @@ else
     ((FAILURES+=1))
 fi
 
-if cargo fmt --check && git diff --exit-code -- ':*.rs'; then
+if cargo fmt --check --all && git diff --exit-code -- ':*.rs'; then
     echo -e "[${GREEN}OK${NC}]: cargo fmt --check"
 else
     echo -e "[${BOLD_RED}FAIL${NC}]: cargo fmt --check"
     ((FAILURES+=1))
 fi
 
-if cargo clippy -- -D clippy::all && git diff --exit-code -- ':*.rs'; then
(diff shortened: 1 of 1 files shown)


juhotuho10:
just as a note, this is also done in the Clippy documentation for example CI:
https://doc.rust-lang.org/stable/clippy/continuous_integration/github_actions.html

Sharktheone:
Not sure, is `--all-features` included in `--all`?

juhotuho10:
`--all` seems to be a deprecated alias for `--workspace` : `Check all members in the workspace.`
https://doc.rust-lang.org/cargo/commands/cargo-check.html#option-cargo-check---all

`--all-targets` is for checking all targets in the crate, even the test and benches and stuff
https://doc.rust-lang.org/cargo/commands/cargo-check.html#option-cargo-check---all-targets

`--all-features` just enables all the possible features for all selected packages
https://doc.rust-lang.org/cargo/commands/cargo-check.html#option-cargo-check---all-features

`--all-features` is different from `--all`, but i'm not an expert in configuring cargo so I don't know the precise semantics. Not sure how `--all`/`--workspace` differs from `--all-targets` ~~but for what it's worth, the documentation just has `--all-targets --all-features` so that might be enough~~

juhotuho10:
After some further digging, i'm pretty sure `--workspace` includes all crates in the repository that are defined in the root Cargo.toml file, where as `--all-targets` is within the crate, just running all the things in it.

Since this project has a workspace, you can probably include all three, so `--workspace --all-targets --all-features`

PR #8187, opened 2026-02-26
LibWeb+LibGfx: Use accelerated painting surface snapshots if available

gmta:
If we have a Skia backend context available, use it to make image snapshots. Drops blitting CPU usage from 55% to 0.6% for slither.io on my machine.

### DIFF
--- a/Libraries/LibGfx/ImmutableBitmap.cpp
+++ b/Libraries/LibGfx/ImmutableBitmap.cpp
@@ -189,6 +189,17 @@ ErrorOr<BitmapExportResult> ImmutableBitmap::export_to_byte_buffer(ExportFormat 
 
 RefPtr<Gfx::Bitmap const> ImmutableBitmap::bitmap() const
 {
+    if (!m_impl->bitmap && m_impl->sk_image) {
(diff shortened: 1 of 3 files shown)


Libraries/LibGfx/ImmutableBitmap.cpp
```diff
@@ -398,6 +409,22 @@ NonnullRefPtr<ImmutableBitmap> ImmutableBitmap::create(NonnullRefPtr<Bitmap> bit
 
 NonnullRefPtr<ImmutableBitmap> ImmutableBitmap::create_snapshot_from_painting_surface(NonnullRefPtr<PaintingSurface> painting_surface)
 {
+    if (auto context = painting_surface->skia_backend_context()) {
```
kalenikaliaksandr: you made `ImmutableBitmap::bitmap()` populate `bitmap` in case it's missing, so I don't see any reasons to guard this branch on `context`.

gmta: True - I only added the lazy population of `.bitmap` when I discovered a crash, but I didn't see I could drop the eager population because of that. Thanks!

PR #8193, opened 2026-02-26
LibHTTP: Correctly normalize header whitespace in cache utilities

trflynn89:
#8188 caught my attention - CC @shannonbooth


shannonbooth:
Ha, yeah, I noticed it there too but thought I would try come up with a test if possible. Guessing maybe not too easy?

I was also wondering about changing that trim_whitespace API TBH, not the first time something like this has come up (though, that's a separate thing which may not be worthwhile)

trflynn89:
I've added a test. These were Vary header issues, which is just a list of other header names. So if someone sent:
```
Vary: Accept\v
```

We'd have previously trimmed that to `Accept`, whereas now that won't be trimmed and won't match any `Accept` header.

I've confirmed this behavior matches Firefox as well.

PR #8196, opened 2026-02-27
LibWeb: Support `<symbols()>` function

Calme1709:
This PR adds support for the use of the `<symbols()>` function to define anonymous counter styles within `<counter-style>`.

The first commit "LibWeb: Make CounterStyleDefinition::create return type non-optional" is also included in #8180 so whichever is merged second will need to be rebased after the first.

### DIFF
--- a/Libraries/LibWeb/CSS/Enums.json
+++ b/Libraries/LibWeb/CSS/Enums.json
@@ -965,6 +965,13 @@
     "round",
     "bevel"
   ],
+  "symbols-type": [
+    "cyclic",
+    "numeric",
+    "alphabetic",
+    "symbolic",
+    "fixed"
+  ],
   "table-layout": [
     "auto",
     "fixed"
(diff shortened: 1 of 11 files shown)


Libraries/LibWeb/CSS/ComputedValues.h
```diff
@@ -145,7 +145,7 @@ struct WillChange {
 
 using CursorData = Variant<NonnullRefPtr<CursorStyleValue const>, CursorPredefined>;
 
-using ListStyleType = Variant<Empty, Optional<CounterStyle const&>, String>;
+using ListStyleType = Variant<Empty, Optional<CounterStyle>, String>;
```
AtkinsSJ: Having to store a full copy of the counter-style on every ComputedValues is pretty unfortunate, especially as the majority of those are an unused copy of `disc`. Maybe locally-defined ones could be stored as an OwnPtr or similar? (Or even better RefPtr if we can share the same one between ComputedValues from the same source declaration.)

Calme1709: That's a much better idea - I have updated this commit to instead make `CounterStyle` ref counted.

PR #8197, opened 2026-02-27
LibCore: Stop atomically ref-counting EventReceiver

Zaggy1024:
Inheriting from AtomicRefCounted and Weakable is asking for misuse of WeakPtr to result in TOCTOU-caused UAFs.

In order to ensure we're not misusing EventReceiver across threads, I ran test-web and some sites with a temporary hack to verify that ref()s and unref()s are always called from the same thread on the class.

This required some changes in BackgroundAction that naturally cascaded into some changes to RenderingThread and EventLoop:
- BackgroundAction no longer uses Promise and the EventLoop::add_job() method.
- BackgroundActions can no longer return errors from their completion callbacks.
- RenderingThread now detaches its thread and directly signals it to exit in its destructor. No more Promise/add_job() hack.
- EventLoop/ThreadEventQueue no longer tracks "pending jobs". After the above changes, that functionality was unused.

### DIFF
--- a/Tests/LibCore/TestLibCoreStream.cpp
+++ b/Tests/LibCore/TestLibCoreStream.cpp
@@ -343,6 +343,8 @@ TEST_CASE(local_socket_read)
     Core::EventLoop event_loop;
 
     auto socket_path = ByteString::formatted("{}/{}", Core::StandardPaths::tempfile_directory(), "test-socket"sv);
+    if (!Core::System::stat(socket_path).is_error())
+        TRY_OR_FAIL(Core::System::unlink(socket_path));
 
     auto local_server = Core::LocalServer::construct();
     EXPECT(local_server->listen(socket_path));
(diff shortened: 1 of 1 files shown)


gmta:
@Zaggy1024 the [failing job](https://github.com/LadybirdBrowser/ladybird/actions/runs/22476227273/job/65103652137?pr=8197) seems very related to the changes in this PR.

Zaggy1024:
Huh, that's funny, I ran into that on my machine as well while I was fixing the original failure. The issue here is that the runner hasn't cleared its `/tmp/`, so the socket file still exists and the LocalServer is unhappy about that. Perhaps it may need a reboot?

Zaggy1024:
Heap UAF showed up on the macOS CI in RenderingThread, I'll have to investigate this before merge.

Zaggy1024:
Added a fix for [a use-after-free on the Unix event loop impl's ThreadData struct](https://github.com/LadybirdBrowser/ladybird/actions/runs/22483237062/job/65126407432?pr=8197#step:18:28), ~~should be good to go now~~ if I hadn't left an unused field in.

PR #8199, opened 2026-02-27
test-web: Convert ANSI control codes to HTML styling in dumped output

Zaggy1024:
Instead of stripping all the control codes, let's use them to make our test dumps more readable. :^)

image


AtkinsSJ:
If you feel like extending this in the future, I believe we're supposed to handle these in `console.log()` and friends too.

Zaggy1024:
Oh, fun. Do the Firefox dev tools not handle that on their side?

AtkinsSJ:
...actually they might, I forgot we use that. :laughing:

PR #8200, opened 2026-02-27
LibJS/Rust: Applying clippy lints

juhotuho10:
Applied clippy lints to make the code generally more idiomatic and simpler

Most of the changes are pretty simple, except I made bigger changes for clippy::unnecessary_wraps since it removed unnecessary error handling for infallible functions

You might even consider adding some these lints as warn by default now that they are fixed

Also here is a good [catalogue](https://rust-lang.github.io/rust-clippy/master/index.html) for the lints and lint groups to see what the lints do and what they warn about

Applied the following lints:
- clippy::uninlined_format_args
- clippy::semicolon_if_nothing_returned
- clippy::manual_let_else
- clippy::unnecessary_wraps
- clippy::manual_assert
- clippy::useless_let_if_seq
- clippy::ref_option
- clippy::elidable_lifetime_names
- clippy::redundant_clone
- clippy::unnested_or_patterns
- clippy::borrow_as_ptr
- clippy::explicit_iter_loop

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,3 +16,19 @@ inherits = "release"
 incremental = false
 codegen-units = 1
 lto = true
+
+[workspace.lints.clippy]
+all = "deny" # includes lintgroups: correctness, suspicious, style, complexity, perf
+# from clippy:nursery:
+redundant_clone = "deny"
+# from clippy:pedantic:
+uninlined_format_args = "deny"
+semicolon_if_nothing_returned = "deny"
+manual_let_else = "deny"
+unnecessary_wraps = "deny"
+manual_assert = "deny"
+ref_option = "deny"
+elidable_lifetime_names = "deny"
+unnested_or_patterns = "deny"
+borrow_as_ptr = "deny"
+explicit_iter_loop = "deny"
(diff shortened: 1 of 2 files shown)


ADKaster:
These lints look useful, but they don't seem to be CI enforced. is there a config file that needs updated so that these are handled in `Meta/lint-ci.sh` ? It looks like we're calling `cargo clippy -- -D clippy::all`

ADKaster:
perhaps the top-level Cargo.toml? https://doc.rust-lang.org/stable/cargo/reference/manifest.html#the-lints-section

juhotuho10:
You can add lints to the Cargo.toml file at project root and then use 
```toml
[lints]
workspace = true
```
to inherents workspace lints to the crates in the repository. Since the compiler follows the lints set for the project, it would effectively enforce it in the CI as well.
I can do that, but do I do it in this PR or should I make another PR for that?

juhotuho10:
~~Added lints to workspace Cargo.toml file in #8256 
but this PR has to be merged before the other one, otherwise the new PR will cause the CI to fail with the `clippy check`~~
Edit: the other PR is closed in favor of having the changes here

juhotuho10:
Timeout? Is this something I caused or just a something that happens at times?

gmta:
> Added lints to workspace Cargo.toml file in #8256 but this PR has to be merged before the other one, otherwise the new PR will cause the CI to fail with the `clippy check`

I'd rather you combine both PRs into this one - we then can have both the linter's results _and_ CI check in one go.

PR #8204, opened 2026-02-27
CI: Add web-benchmarks job

tcl3:
This PR adds a job that compiles a Distribution build of Ladybird, runs 5 iterations of the `web-benchmarks` test suite then calls the same web hook used by `js-benchmarks`.

### DIFF
--- a/.github/workflows/web-benchmarks.yml
+++ b/.github/workflows/web-benchmarks.yml
@@ -1,0 +1,158 @@
+name: 'Web benchmarks'
+
+on:
+  workflow_run:
+    workflows: ['CI']
+    branches: [master]
+    types: [completed]
+
+env:
+  LADYBIRD_SOURCE_DIR: ${{ github.workspace }}
+  CCACHE_DIR: ${{ github.workspace }}/.ccache
+  VCPKG_ROOT: ${{ github.workspace }}/Build/vcpkg
+
+jobs:
+  web-benchmarks:
+    runs-on: ${{ fromJSON(matrix.runner_labels) }}
+    if: ${{ github.repository == 'LadybirdBrowser/ladybird' && github.event.workflow_run.conclusion == 'success' }}
+    name: ${{ matrix.os_name }}, ${{ matrix.arch }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        os_name: ['Linux']
+        arch: ['x86_64']
+        runner_labels: ['["ubuntu-24.04-internal", "web-benchmarks", "self-hosted"]']
+
+        include:
+          - os_name: 'macOS'
+            arch: 'arm64'
+            runner_labels: '["macos-15", "web-benchmarks", "self-hosted"]'
+
+    permissions:
+      actions: read
+      contents: read
+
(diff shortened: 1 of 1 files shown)


.github/workflows/web-benchmarks.yml
```diff
@@ -0,0 +1,153 @@
...
+        shell: bash
+        run: |
+          echo '{
+              "commit": "${{ steps.ladybird-commit.outputs.sha }}",
+              "benchmarks_commit": "${{ steps.web-benchmarks-commit.outputs.sha }}",
+              "os": "${{ matrix.os_name }}",
+              "arch": "${{ matrix.arch }}",
+              "artifact": "web-benchmarks-results-${{ matrix.os_name }}-${{ matrix.arch }}",
+              "artifact_run_id": "${{ github.run_id }}",
+              "category": "web"
```
gmta: This field should be present in the benchmark results themselves, not here. I.e. you can send in multiple types with a single call to the webhook if necessary.

.github/workflows/web-benchmarks.yml
```diff
@@ -0,0 +1,153 @@
...
+      fail-fast: false
+      matrix:
+        os_name: ['Linux']
+        arch: ['x86_64']
+        runner_labels: ['["ubuntu-24.04-internal", "web-benchmarks", "self-hosted"]']
+
+        include:
+          - os_name: 'Linux'
+            arch: 'arm64'
+            runner_labels: '["ubuntu-24.04-arm-internal", "web-benchmarks", "self-hosted"]'
```
gmta: These runner_labels need to be updated to Blacksmith instances.

.github/workflows/web-benchmarks.yml
tcl3: I've updated `web-benchmarks` to have a category field as `js-benchmarks` does.

.github/workflows/web-benchmarks.yml
gmta: I'm sorry - your previous setup was closer to what we need: we will require self-hosted, physical machines that will be running web-benchmarks to provide consistent results.

Blacksmith is great for regular CI, but we don't know exactly how they're set up and how resource contention works over there. Additionally, we [might transparently get bumped up to larger instance types](https://docs.blacksmith.sh/blacksmith-runners/overview#free-upgrades) skewing the benchmark results even more.

That being said, we don't have a self-hosted Linux arm64 machine. We do have:

* `ladybird-js-runner` (Linux x86-64) which is labeled with `js-benchmarks`;
* `ladybird-mac-runner` (macOS arm64) which is labeled with `js-benchmarks` as well.

I could add the `web-benchmarks` label to both machines, and maybe rename the `ladybird-js-runner` machine to something more sensible 😅 agreed?

tcl3: Sure, I've updated the job to use labels pointing to our self-hosted machines.

gmta:
I've set up two self-hosted runners (physical machines) with the new `web-benchmarks` label - they should pick up the new job.

PR #8208, opened 2026-02-27
LibJS: Add `Value::is<T>` and make use of it and `Value::as_if<T>`

trflynn89:
I just kept spotting these and wanting to modernize them.

reject:
```c++
static void update_function_name(Value value, Utf16FlyString const& name)
{
    if (!value.is_function())
        return;
    auto& function = value.as_function();
    if (is<ECMAScriptFunctionObject>(function) && static_cast<ECMAScriptFunctionObject const&>(function).name().is_empty())
        static_cast<ECMAScriptFunctionObject&>(function).set_name(name);
}
```

embrace:
```c++
static void update_function_name(Value value, Utf16FlyString const& name)
{
    if (auto function = value.as_if<ECMAScriptFunctionObject>(); function && function->name().is_empty())
        function->set_name(name);
}
```

### DIFF
--- a/Libraries/LibJS/Runtime/AbstractOperations.cpp
+++ b/Libraries/LibJS/Runtime/AbstractOperations.cpp
@@ -620,17 +620,15 @@ ThrowCompletionOr<Value> perform_eval(VM& vm, Value x, CallerMode strict_caller,
         auto this_environment_record = get_this_environment(vm);
 
         // b. If thisEnvRec is a function Environment Record, then
-        if (is<FunctionEnvironment>(*this_environment_record)) {
-            auto& this_function_environment_record = static_cast<FunctionEnvironment&>(*this_environment_record);
-
+        if (auto* this_function_environment_record = as_if<FunctionEnvironment>(*this_environment_record)) {
             // i. Let F be thisEnvRec.[[FunctionObject]].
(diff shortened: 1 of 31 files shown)


shannonbooth:
I actually had a branch for this also which I never got around to upstreaming 😅  https://github.com/shannonbooth/ladybird/commits/as-if

regarding the change in trait for as_if, in my branch I actually changed it to support as_if for Object too, since that pattern also comes up: https://github.com/shannonbooth/ladybird/commit/c2bb553344ff52b6827bf6925d2130d6e1b3c27c

And then the implementation turned into https://github.com/shannonbooth/ladybird/commit/2ea9d2549bd87daca637d3922269a9a42c86a063

But yeah, as_if<Object> comes up for places such as https://github.com/shannonbooth/ladybird/commit/247ab6763d308f1ba2bebbaed61a3ae7dbb67236#diff-4c1fc11f74860139c6a0947804d3c1b060cda8d0acfca8223b19541b2ef9f0f4R25

trflynn89:
I'll leave the unwinding of `is_object` + `as_object` to another PR :sweat_smile: 

FYI I don't think you need to adjust the trait itself, `IsBaseOf<Object, Object>` is true.

shannonbooth:
At some point I'll get annoyed enough by all of the T*'s around in LibJS and port them over to GC::Ref<T> as my last pet peeve remaining for modernization :P

PR #8211, opened 2026-02-27
LibJS+LibWeb: Parse JavaScript off the main thread

awesomekling:
This splits the Rust JS compilation pipeline into two phases: parse (thread-safe, no GC) and compile (main thread, GC allocations).

A simple thread pool runs parse jobs on worker threads, and the results are bounced back to the main thread for compilation and delivery. Both classic scripts and modules get this treatment.

As a prerequisite, regex compilation had to be deferred to a post-parse step, since LibRegex is not thread-safe. The parsers now collect raw pattern+flags strings and we batch-compile them on the main thread after parsing.

Flow:
1. Main thread creates a `JS::SourceCode` and roots any GC objects needed by the completion callback.
2. Worker thread runs `parse_program()` which does lexing, parsing, and scope analysis. No VM, no GC, no problem.
3. Worker bounces back to the main thread via `deferred_invoke()`.
4. Main thread compiles the pre-parsed result (bytecode generation, GC allocations) and delivers the finished script/module.

With this feature, we no longer block the main thread while parsing JavaScript files. This saves quite a bit of loading time on heavy pages, for example 1.1 sec on https://vscode.dev/ and 0.65 sec on https://x.com/

We could do better here if LibRegex were thread safe, but that's a problem for another day :)

### DIFF
--- a/Libraries/LibJS/Rust/src/lib.rs
+++ b/Libraries/LibJS/Rust/src/lib.rs
@@ -92,7 +92,7 @@ pub(crate) fn u32_from_usize(value: usize) -> u32 {
 }
 
 use ast::StatementKind;
-use parser::{Parser, ProgramType};
+use parser::{ParseError, Parser, ProgramType};
 use std::cell::RefCell;
 use std::collections::HashSet;
 use std::ffi::c_void;
@@ -111,15 +111,9 @@ pub struct ParsedProgram {
     scope_ref: Rc<RefCell<ast::ScopeData>>,
     is_strict_mode: bool,
     has_top_level_await: bool,
-    errors: Vec<ParsedError>,
+    errors: Vec<ParseError>,
     ast_dump: Option<Vec<u8>>,
     deferred_regexes: Vec<parser::DeferredRegex>,
-}
-
-struct ParsedError {
-    message: String,
-    line: u32,
-    column: u32,
 }
 
 // SAFETY: Full ownership transfer between threads, never concurrent access.
@@ -438,25 +432,9 @@ pub unsafe extern "C" fn rust_parse_program(
             let program = parser.parse_program(false);
 
             // Collect errors from both parser and scope collector.
-            let mut errors = Vec::new();
-            if parser.has_errors() {
-                for err in parser.errors() {
-                    errors.push(ParsedError {
-                        message: err.message.clone(),
-                        line: err.line,
-                        column: err.column,
-                    });
-                }
-            }
-
-            if errors.is_empty() && parser.scope_collector.has_errors() {
-                for err in parser.scope_collector.drain_errors() {
-                    errors.push(ParsedError {
-                        message: err.message.clone(),
-                        line: err.line,
-                        column: err.column,
-                    });
-                }
+            let mut errors = parser.take_errors();
+            if errors.is_empty() {
+                errors = parser.scope_collector.drain_errors();
             }
 
             if errors.is_empty() {
@@ -547,13 +525,9 @@ pub unsafe extern "C" fn rust_parsed_program_compile_regexes(parsed: *mut Parsed
     unsafe {
         let parsed = &mut *parsed;
         let deferred = std::mem::take(&mut parsed.deferred_regexes);
-        for err in Parser::compile_deferred_regexes(deferred) {
-            parsed.errors.push(ParsedError {
-                message: err.message,
-                line: err.line,
-                column: err.column,
-            });
-        }
+        parsed
+            .errors
+            .extend(Parser::compile_deferred_regexes(deferred));
     }
 }
 
--- a/Libraries/LibJS/Rust/src/parser.rs
+++ b/Libraries/LibJS/Rust/src/parser.rs
@@ -174,7 +174,7 @@ impl ForbiddenTokens {
     }
 }
 
-pub struct ParserError {
+pub struct ParseError {
     pub message: String,
     pub line: u32,
     pub column: u32,
@@ -229,7 +229,7 @@ pub struct Parser<'a> {
     lexer: Lexer<'a>,
     /// `consume()` returns this and advances to the next token.
     current_token: Token,
-    errors: Vec<ParserError>,
+    errors: Vec<ParseError>,
     saved_states: Vec<SavedState>,
     program_type: ProgramType,
     /// UTF-16 source text.
@@ -612,7 +612,7 @@ impl<'a> Parser<'a> {
     // === Error reporting ===
 
     pub(crate) fn syntax_error(&mut self, message: &str) {
-        self.errors.push(ParserError {
+        self.errors.push(ParseError {
             message: message.to_string(),
             line: self.current_token.line_number,
             column: self.current_token.line_column,
@@ -620,7 +620,7 @@ impl<'a> Parser<'a> {
     }
 
     pub(crate) fn syntax_error_at(&mut self, message: &str, line: u32, column: u32) {
-        self.errors.push(ParserError {
+        self.errors.push(ParseError {
             message: message.to_string(),
             line,
             column,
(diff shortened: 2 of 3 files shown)


Libraries/LibWeb/HTML/Scripting/Fetching.cpp
```diff
@@ -33,6 +38,39 @@
 
 namespace Web::HTML {
 
+// Submit a parse_program() call to the thread pool, then bounce back to
+// the main thread via deferred_invoke once parsing completes.
+// `on_parsed` is called on the main thread with the RustParsedProgram*
+// and the SourceCode.
+// NB: The callback is heap-allocated so that if the event loop is
+//     destroyed during parsing, we can intentionally leak it (and any
```
alimpfard: 🤔 say what?

Libraries/LibJS/SourceTextModule.cpp
```diff
@@ -204,6 +204,31 @@ void SourceTextModule::visit_edges(Cell::Visitor& visitor)
...
+            realm, filename, host_defined, module_result.has_top_level_await,
+            move(module_result.requested_modules), move(module_result.import_entries),
+            move(module_result.local_export_entries), move(module_result.indirect_export_entries),
+            move(module_result.star_export_entries), move(module_result.default_export_binding_name),
+            move(module_result.var_declared_names), move(module_result.lexical_bindings),
+            move(functions_to_initialize),
+            module_result.executable.ptr(), module_result.tla_shared_data.ptr());
+    }
+    // Fallback: should not be reached (parsed is always from the Rust pipeline).
+    VERIFY_NOT_REACHED();
```
alimpfard: why not `MUST` the result with the same comment?

Libraries/LibJS/Script.h
```diff
@@ -51,6 +53,7 @@ class JS_API Script final : public Cell {
 
     virtual ~Script() override;
     static Result<GC::Ref<Script>, Vector<ParserError>> parse(StringView source_text, Realm&, StringView filename = {}, HostDefined* = nullptr, size_t line_number_offset = 1);
+    static Result<GC::Ref<Script>, Vector<ParserError>> create_from_parsed(::RustParsedProgram* parsed, NonnullRefPtr<SourceCode const> source_code, Realm&, HostDefined* = nullptr);
```
alimpfard: the `::` on `RustParsedProgram` seems superfluous, there's only one in scope (more dotted in random places)

Libraries/LibJS/RustIntegration.h
```diff
@@ -69,6 +79,16 @@ struct ModuleResult {
...
 
+// Parse a program (script or module) without GC interaction. Thread-safe.
+// Returns nullptr if Rust is not available.
+::RustParsedProgram* parse_program(SourceCode const& source_code, ProgramType type, size_t line_number_offset = 0);
+
+// Compile a previously parsed script. Must be called on the main thread.
+// Consumes and frees the RustParsedProgram.
+// Returns nullopt if Rust is not available.
+Optional<Result<ScriptResult, Vector<ParserError>>> compile_parsed_script(
+    ::RustParsedProgram* parsed, NonnullRefPtr<SourceCode const> source_code, Realm& realm);
```
ADKaster: this file has some interesting line length limit formatting. is

```
ResultType function_name(
   arg, arg, arg, arg);
```

intentional?

other files have

```
ResultType function_name(
   arg,
   arg,
   arg);
```

which is a lot more palatable.

Libraries/LibThreading/ThreadPool.cpp
```diff
@@ -0,0 +1,85 @@
...
+ThreadPool& ThreadPool::the()
+{
+    static ThreadPool instance;
+    return instance;
+}
+
+ThreadPool::ThreadPool()
+{
+    s_work_queue = new Queue<Function<void()>>;
+    s_threads = new Vector<NonnullRefPtr<Thread>>;
```
ADKaster: with the thread pool being a singleton, is there a need for these manually allocated file-scope statics? Can they just be members of the thread pool?

Same Q for the muted and condition variable.

Also, don't we have Threading:: classes for mutex and condvar?

Libraries/LibThreading/ThreadPool.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+ThreadPool::~ThreadPool()
+{
+    pthread_mutex_lock(&s_mutex);
+    s_running = false;
+    pthread_cond_broadcast(&s_condition);
+    pthread_mutex_unlock(&s_mutex);
+
+    for (auto& thread : *s_threads)
+        MUST(thread->join());
```
ADKaster: the way this is structured will cause these threads to continue (and finish) executing during global destructor calls. Due to ~ThreadPool being called b/c of instance being a function-static in ::the(). The ordering is also non-determinsitic b/c when ~ThreadPool is registered with `__cxa_atexit` (or win equivalent) is dependent on when ::the() is called the first time.

unless I'm mistaken, this is the sort of thing that @Zaggy1024 has been juggling in media for a while.

Libraries/LibJS/RustIntegration.h
```diff
@@ -69,6 +79,16 @@ struct ModuleResult {
     GC::Root<SharedFunctionInstanceData> tla_shared_data;
 };
 
+// Parse a program (script or module) without GC interaction. Thread-safe.
+// Returns nullptr if Rust is not available.
+::RustParsedProgram* parse_program(SourceCode const& source_code, ProgramType type, size_t line_number_offset = 0);
```
ADKaster: this is marked JS_API in a later commit. would it be better to mark it when it's added?

Libraries/LibThreading/ThreadPool.cpp
ADKaster: imo this points to a need for a more centralized "shutting down, quit ur threads" infrastructure to prevent awkwardness like the 'event loop gone, but we are on another thread that wants to submit work to it' hole added in this PR that Ali already balked at.

Libraries/LibThreading/ThreadPool.cpp
```diff
@@ -0,0 +1,85 @@
...
+        Function<void()> work;
+
+        pthread_mutex_lock(&s_mutex);
+        while (s_work_queue->is_empty() && s_running)
+            pthread_cond_wait(&s_condition, &s_mutex);
+
+        if (!s_running && s_work_queue->is_empty()) {
+            pthread_mutex_unlock(&s_mutex);
+            break;
+        }
```
Zaggy1024: Why wait for the queue to be emptied before exiting? If `s_running == false`, aren't we trying to close the entire process ASAP?

Libraries/LibThreading/ThreadPool.cpp
Zaggy1024: Regarding the thread joining, my usual solution is just to _not_ join. Instead, I'll allocate storage for cross-thread communication that lives as long as any thread cares about it, then detach the thread so that it can run after the originating thread exits. That way, the thread gets told to exit when the main thread is exiting via the destructor, but the main thread running that destructor doesn't have to wait for any threads. (Imagine what would happen if we had a bunch of classes trying to join their threads on the main thread!)

When it comes to invoking onto a dead event loop, that's what WeakEventLoopReference is for. Not necessarily the prettiest solution, but the best I could come up with given that the event loop is on the stack. It seems to work well now, at least after I fixed a UAF on the event loop implementations' thread datas.

Also, isn't this destructor going to UAF on the mutex and condition variable?

Libraries/LibThreading/ThreadPool.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+#include <AK/Queue.h>
+#include <AK/Vector.h>
+#include <LibThreading/Thread.h>
+#include <LibThreading/ThreadPool.h>
+
+static constexpr size_t THREAD_COUNT = 4;
+
+static pthread_mutex_t s_mutex = PTHREAD_MUTEX_INITIALIZER;
+static pthread_cond_t s_condition = PTHREAD_COND_INITIALIZER;
```
Zaggy1024: Is there a reason why we can't use our Mutex and ConditionVariable classes here instead of relying directly on pthreads?

Edit: Missed Andrew mentioning this, but I suppose having a separate conversation for this doesn't hurt...

Zaggy1024 (commented):
Not a thorough review, mostly checking the stuff related to Andrew's comment on ThreadPool.

awesomekling:
I think it's time we decide once and for all that we don't care about meticulous traditional teardown of WebContent (or js CLI for that matter) processes. It just adds complexity for minor debugging (i.e accountability) benefit.

Zaggy1024:
As long as the threads don't hold up the process from being fully destroyed when they have no exit flag (it seems like they should just get killed?), I agree, it's not really worth worrying about in this case.

PR #8213, opened 2026-02-28
Tests: Add multi-origin test framework

not-a-rootkit:
The current test framework is robust but not exhaustive enough to properly evaluate SOP violations. I'm proposing to add a simple multi-HTTP server setup, making it possible to test more variations of cross origin behaviour patterns, similar to what webkit does with `Tools/Scripts/run-webkit-tests`.

## Test Instructions
Build the test runner:

```
./Meta/ladybird.py build test-web
```

After building, run tests:

```
export LADYBIRD_SOURCE_DIR=/path/to/ladybird/
./Build/release/bin/test-web -f "*multi-origin*" -R /tmp/results -v
```

Run all origin-related tests in parallel

```
./Build/release/bin/test-web -f "*origin*" -j 8 -R /tmp/results
```
The infrastructure automatically starts 3 HTTP servers on different ports, each representing a distinct origin for cross-origin testing.


Libraries/LibWeb/Internals/Internals.idl
```diff
@@ -62,6 +62,8 @@ interface Internals {
     DOMString getComputedLabel(Element element);
     DOMString getComputedAriaLevel(Element element);
     unsigned short getEchoServerPort();
+    unsigned short getOriginServerPort(unsigned long index);
+    unsigned long getOriginServerCount();
```
shannonbooth, line 66: I’m definitely in favor of improving origin testing, that’s an area we should improve on. My main concern here is introducing a separate test framework for it. I was previously more thinking along the lines of trying to align our infrastructure more closely with what WPT provides, so we can reuse that test coverage more directly, which I had tried making some progress on in #8458.

Saying that, I still see value in a separate framework if it fills gaps in WPT, makes writing our own tests easier to write, or speed of running tests. Do you think this type of setup might give us some advantages in those types of areas?

not-a-rootkit: Ah I only just saw this. PR #8458 already made progress on the harder problem, by adapting get-host-info.sub.js to work with Ladybird's test infrastructure and importing real WPT cross-origin tests (origin-from-messageevent.window.js). The main benefit I see that the separate framework provides is the `createEcho()` pattern for letting a test define both client and server behaviour in a single HTML file:

```
  await server.createEcho(1, "GET", "/api", {
      status: 200, body: "hello",
      headers: {"Access-Control-Allow-Origin": "*"}
  });
```

However, I realise that WPT uses virtual hostnames (www1.web-platform.test vs not-web-platform.test) which lets it test cookie scoping, document.domain, same-site vs cross-site distinctions. Port-based 127.0.0.1:N origins all share the same host, so a meaningful class of origin tests simply can't be expressed by what I've implemented here, so I prefer the work you've started there.

A nice middle ground might be to unify everything into one server, one fixture, and enhance the WPT bridge to use distinct ports and then we can cleanup some of the boilerplate I implemented here:
 - Delete http-test-server.py
 - Keep MultiOriginTestServer JS class in include.js (unchanged, ergonomic API for custom tests)
 - Keep HTTPTestServer class (can get this via `getEchoServerPort()` alias)
 - No changes needed to Internals.idl, Options.h, or WebContent/main.cpp

I'll have a crack at implementing this and see if I can get something working that combines the best of both.

shannonbooth: I believe the cross host behaviour is currently achieved by host1 / the normal host being localhost, and host2 being 127.0.0.1 which is technically cross origin host even though it resolves to the same server. That way we can still test cross site behaviour.

FWIW I try and make some further test harness changes relating to this in https://github.com/LadybirdBrowser/ladybird/pull/8540, though I'm not 100% sure about that approach.

PR #8219, opened 2026-02-28
LibWeb+UI: Add fullscreen toggles to the built-in media player

trflynn89:
This adds a fullscreen icon to the player itself, along with double-click and a content menu item.

https://github.com/user-attachments/assets/8425cba3-db79-4af9-87a8-dd9aeb72c0df


Libraries/LibWeb/HTML/HTMLMediaElement.cpp
```diff
@@ -476,6 +476,14 @@ void HTMLMediaElement::set_muted(bool muted)
     set_needs_style_update(true);
 }
 
+void HTMLMediaElement::toggle_fullscreen()
```
Zaggy1024, line 479: It feels to me like this belongs in the MediaControls class. Are we going to have a use for it outside there?

Libraries/LibWeb/HTML/HTMLMediaElement.cpp
```diff
@@ -476,6 +476,14 @@ void HTMLMediaElement::set_muted(bool muted)
...
+{
+    if (is_fullscreen_element())
```
Zaggy1024: The spec seems to imply that multiple elements may have the fullscreen flag set. Could we instead detect this the same way JS would, by `element == element->document().fullscreen_element_for_bindings()`?

Libraries/LibWeb/HTML/MediaControls.cpp
```diff
@@ -657,6 +704,17 @@ void MediaControls::update_volume_and_mute_indicator()
...
 
+void MediaControls::update_fullscreen_icon()
+{
+    if (!m_fullscreen_icon)
+        return;
+
+    static auto s_fullscreen_class = "fullscreen"_string;
+
+    VERIFY(m_media_element);
+    MUST(m_fullscreen_icon->class_list()->toggle(s_fullscreen_class, m_media_element->is_fullscreen_element()));
```
Zaggy1024: Same JS-like fullscreen detection here?

Libraries/LibWeb/HTML/HTMLMediaElement.cpp
trflynn89: Yes - it's used later in the PR for the context menu

Libraries/LibWeb/HTML/HTMLMediaElement.cpp
trflynn89: I originally used some of the substeps of `Element::removing_steps_fullscreen` to do this, but that ended up hitting the third fullish screen state I showed on discord. I can take another look or add a fixme.

Libraries/LibWeb/HTML/MediaControls.cpp
trflynn89: What would change here? Don't we want to render the button according to whether this element is fullscreen, regardless of whether it's the only one?

Zaggy1024: My main concern is that we're using unexposed API here, my preference is to use DOM APIs for this if possible so that we can be closer to what websites would be doing.

If it doesn't work though, not much we can do until we fix the underlying bug or determine that it's intended behavior and find another way.

Libraries/LibWeb/HTML/HTMLMediaElement.cpp
Zaggy1024: Gotcha, that seems fine. I think it could make sense to determine whether the context menu should initiate or exit fullscreen ahead of time in case the state changes while the menu is open, but I doubt that's actually important in practice.

Libraries/LibWeb/HTML/HTMLMediaElement.cpp
trflynn89: Turns out it was this step from the removal steps causing the full-window issue:
```
3. If document’s top layer contains node, remove from the top layer immediately given node
```

We now check if the document fullscreen element is `this`, and:
* If yes, exit fullscreen on the document
* If no, only unfullscreen this element

Libraries/LibWeb/HTML/MediaControls.cpp
trflynn89: I don't think there even is a way to determine if an element is fullscreen, is there? You can only check if an element is *the* document fullscreen element, which is just the topmost. If there's multiple, there isn't a DOM API to check if this is one of them.

Zaggy1024: Right, I was just hoping that distinction doesn't matter in practice, but I honestly don't have a good grasp on how all that works to know for sure.

Libraries/LibWeb/HTML/HTMLMediaElement.cpp
trflynn89: I've updated these to use DOM APIs.

PR #8227, opened 2026-02-28
LibWeb: Update SubtleCrypto.digest() to match the spec

mikiubo:
The WebCrypto spec was updated to normalize the algorithm before getting a copy of the data bytes, and to queue a global task on the crypto task source when rejecting or resolving the promise.

This fixes some WPT tests.


Tests/LibWeb/Text/expected/wpt-import/WebCryptoAPI/digest/cshake.tentative.https.any.txt
```diff
@@ -2,8 +2,7 @@ Harness status: OK
 
 Found 56 tests
 
-44 Pass
-12 Fail
+56 Pass
```
gmta, line 5: Please update these test results in the same commit as the changes that caused them in the first place.

PR #8231, opened 2026-03-01
LibWeb: Guard layout_node() null dereference in set_user_selection()

### LINKED ISSUE #8205: Crash when selecting text from inside a shadow tree to outside one
Not sure the exact mechanism here, but it's reproducible. I haven't reduced it, just a quick report while I'm in the middle of something.

Image

1. Open the `ladybird/Tests/LibWeb/Text/input/wpt-import/css/css-shadow-parts/inner-host.html` test
2. Click and drag from somewhere in "The following text should be blue" back into the green "this text".
3. 💥 

The text after the green text is in a nested shadow tree, compared to the text before it. There seem to be other issues on this same page, related to whether each of the "this text" spans are selectable at all.

cc @gmta: selection fun times

TechWizard9999:
## What

Fixes a crash when selecting text that crosses shadow DOM boundaries.

## Why

When dragging a selection from inside a shadow tree to outside or vice versa, the selection handles nodes near the boundary. Some of these nodes may not have an associated layout node. The `set_user_selection()` function walked up the tree calling `node->layout_node()->user_select_used_value()` without checking if `layout_node()` was null, causing a crash.

## Changes

1. **`EventHandler.cpp`** — Added a `safe_user_select` helper lambda that walks up the DOM tree to find the nearest ancestor with a layout node, inheriting its `user-select` value. If no such node is found, it defaults to `CSS::UserSelect::Text`. Replaced all 6 unsafe calls with this helper.
2. **Crash test** — Added `Tests/LibWeb/Crash/HTML/selection-across-shadow-tree.html` which simulates a user selection crossing a shadow boundary using `internals.mouseDown/Move/Up`.

## Testing

Fixes #8205

cc @gmta

### DIFF
--- a/Libraries/LibWeb/Page/EventHandler.cpp
+++ b/Libraries/LibWeb/Page/EventHandler.cpp
@@ -275,6 +275,17 @@ static GC::Ref<DOM::Range> find_paragraph_range(DOM::Text& text_node, WebIDL::Un
 // https://drafts.csswg.org/css-ui/#propdef-user-select
 static void set_user_selection(GC::Ptr<DOM::Node> anchor_node, size_t anchor_offset, GC::Ptr<DOM::Node> focus_node, size_t focus_offset, Selection::Selection* selection, CSS::UserSelect user_select)
 {
+    // Helper to safely get the user-select used value. If a node has no layout node
+    // (e.g. near shadow DOM boundaries), we walk up the tree to find the nearest ancestor
+    // that does have a layout node to inherit the correct user-select value.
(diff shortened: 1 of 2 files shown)


TechWizard9999:
I have fixed the lint errors. If anything is required please let me know. I am eager to work on tasks like this, but since this is my first time, please guide me in one of the issues needed or in discussions.

TechWizard9999:
@gmta Rebased on latest master. Could you please take another look when you have time?

TechWizard9999:
Hi @awesomekling , Could you please review if there are any changes please let me know

PR #8233, opened 2026-03-01
LibWeb: Insert newlines at block boundaries in copied text

### LINKED ISSUE #7315: Copy text does not create newlines between div elements
### Summary

When I try to copy text from a cgit diff view in Ladybird, the newlines are not preserved.

### Operating system

Linux

### Steps to reproduce

1. Go to a cgit based diff view, e.g. https://git.zx2c4.com/cgit/diff/
2. Select some of the lines of text
3. Copy the text (Ctrl+C)
4. Open a text editor, e.g. kate
5. Paste the text (Ctrl+V)

### Expected behavior

The text should be formatted with newlines.

### Actual behavior

The text has no newlines.

__Notes:__
1. The expected behaviour is observed in Firefox.
2. Newlines are created on other pages, e.g. from the readme -- https://git.zx2c4.com/cgit/about/
3. On the diff view the lines are `div`s; on the about view the lines are `ol`/`li`. I've not tested other cases, e.g. `p` elements.

https://git.zx2c4.com/cgit/diff/

```html
<div>lorem</div>
<div>ipsum</div>
<div>dolor</div>
```

```shell
n/a
```

_No response_

_No response_

### Contribute a patch?

TechWizard9999:
## What

Fixes missing newlines when copying text from block-level elements.

## Why

`visible_text_in_range()` only concatenated text from `DOM::Text`
nodes without considering block boundaries. This meant copying
`<div>lorem</div><div>ipsum</div>` produced `"loremipsum"`
instead of `"lorem\nipsum"`.

## Changes

Added a `has_block_ancestor` helper to detect when a text node
lives inside a block container. Newlines are now inserted before
text nodes in both the contained and end-container phases of range
iteration, avoiding duplicates and leading/trailing noise.

## Testing

Fixes #7315

### DIFF
--- a/Tests/LibWeb/Text/expected/select-text.txt
+++ b/Tests/LibWeb/Text/expected/select-text.txt
@@ -4,4 +4,5 @@ dolor
 dolor s
 dolor sit
 First paragraph for triple-click test.
+
 Second paragraph with a link inside.


Libraries/LibWeb/HTML/Navigable.cpp
```diff
@@ -2735,13 +2744,27 @@ static String visible_text_in_range(DOM::Range const& range)
...
-    if (is<DOM::Text>(*range.end_container()) && range.end_container()->layout_node())
+    if (is<DOM::Text>(*range.end_container()) && range.end_container()->layout_node()) {
+        // Insert a newline before end text if it's in a different block.
+        if (!builder.is_empty()
+            && !builder.string_view().ends_with('\n')
+            && has_block_ancestor(range.end_container().ptr())) {
+            builder.append('\n');
+        }
         builder.append(static_cast<DOM::Text const&>(*range.end_container()).data().substring_view(0, range.end_offset()));
+    }
```
gmta: I'm noticing that we still have this issue in Range/Selection stringification. Since this seems to follow that implementation (except for ignoring nodes without layout nodes), you should probably take a look at that and see if there's an implementation or spec issue. Other browsers do seem to retain the newlines.

gmta: Please add a regression test.

TechWizard9999: you're right, it's better to fix this at the stringification level so the selection APIs behave properly too. I'll look into the spec for that and move the fix there, and will add a regression test for it as well

TechWizard9999: I will look into it

TechWizard9999: Yeah sure , Sorry forgot to push itt

TechWizard9999:
@gmta Thank you! I've moved the newline insertion logic down to the Range/Selection stringification level (`Range::to_string()`) so that other selection APIs also behave correctly and retain the newlines. I also added the requested regression test (`selection-across-blocks.html`). Ready for another look!

Libraries/LibWeb/DOM/Range.cpp
```diff
@@ -556,32 +557,77 @@ WebIDL::ExceptionOr<WebIDL::Short> Range::compare_point(GC::Ref<Node> node, WebI
...
 // https://dom.spec.whatwg.org/#dom-range-stringifier
 Utf16String Range::to_string() const
 {
     // 1. Let s be the empty string.
     StringBuilder builder(StringBuilder::Mode::UTF16);
 
-    // 2. If this’s start node is this’s end node and it is a Text node,
-    //    then return the substring of that Text node’s data beginning at this’s start offset and ending at this’s end offset.
+    // 2. If this's start node is this's end node and it is a Text node,
+    //    then return the substring of that Text node's data beginning at this's start offset and ending at this's end offset.
```
gmta: Please don't replace characters from copied spec steps.

Libraries/LibWeb/DOM/Range.cpp
```diff
@@ -556,32 +557,77 @@ WebIDL::ExceptionOr<WebIDL::Short> Range::compare_point(GC::Ref<Node> node, WebI
     return 0;
 }
 
+// Find the nearest block-level HTML element ancestor of a node.
+// We check by tag name to avoid depending on layout state, which may not
+// be up to date when Range::to_string() is called.
```
gmta: If the layout state needs to be up to date for Range::to_string(), you should make sure that is the case instead of working around the issue.

Tests/LibWeb/Text/expected/select-text.txt
```diff
@@ -4,4 +4,6 @@ dolor
 dolor s
 dolor sit
 First paragraph for triple-click test.
-    Second paragraph with a link inside.
+
+    
+Second paragraph with a link inside.
```
gmta, line 8: This seems wrong - other browsers only have two instead of three newlines in between those two lines?

TechWizard9999:
@gmta I have pushed the fixes you requested. If anything is required please let me know. I am eager to work on tasks like this, but since this is my first time, please guide me in one of the issues needed or in discussions.

TechWizard9999:
Rebased this branch on top of the latest master and resolved the conflict. It should be mergeable again now. Please let me know if you’d like any further changes.

TechWizard9999:
@gmta Rebased on latest master and resolved conflicts. Could you please take another look when you have time?

TechWizard9999:
@gmta I addressed the unresolved newline expectation comment by updating  to include the expected blank line (two newlines between those paragraph lines), and pushed the update. Could you please take another look?

TechWizard9999:
@gmta Follow-up: the updated file is Tests/LibWeb/Text/expected/select-text.txt.

PR #8234, opened 2026-03-01
LibWeb: Fix cursor not visible on empty lines in textarea

### LINKED ISSUE #8101: cursor not visible when x_position == 0
### Summary

When focusing a textarea in ladybird, when the cursor is at the beginning of an empty line, it won't be visible.

### Operating system

macOS

### Steps to reproduce

1. Run ladybird and click to focus a textarea
2. Press return to enter a new line

### Expected behavior

The cursor should be visible at the beginning of the next line.

### Actual behavior

The cursor is not visible: this is very disturbing because it gives the impression the user can't type or the textarea lost the focus.

N/A

```html
<!DOCTYPE html>
<html>
<body>
<textarea>
```

```shell
N/A
```

Ladybird:

https://github.com/user-attachments/assets/c53bcfda-916a-4fff-9a56-cbfaa063e6fc

Chrome:

https://github.com/user-attachments/assets/91e02cb8-011e-4412-bce8-6d345e96314e

_No response_

### Contribute a patch?

TechWizard9999:
## What

Fixes invisible cursor on empty lines in `<textarea>` elements.

## Why

When the cursor is on an empty line (e.g. after pressing Enter),
`fragment_at_position()` returns no match because there is no
text fragment for that line. The fallback in `paint_cursor()`
always placed the cursor at the top-left of the content box,
making it invisible on any line other than the first.

## Changes

In the fallback path of `paint_cursor()`, scan all fragments to
find the last one before the cursor offset and position the cursor
at its bottom edge. This ensures the cursor appears on the correct
line.

## Testing

Fixes #8101

### DIFF
--- a/Libraries/LibWeb/Layout/InlineLevelIterator.cpp
+++ b/Libraries/LibWeb/Layout/InlineLevelIterator.cpp
@@ -242,9 +242,15 @@ Optional<InlineLevelIterator::Item> InlineLevelIterator::generate_next_item()
         bool is_last_chunk = (m_text_node_context->next_chunk_index >= m_text_node_context->chunks.size());
 
         auto is_empty_editable = false;
-        if (!chunk_opt.has_value()) {
-            auto const is_only_chunk = is_first_chunk && is_last_chunk;
-            if (is_only_chunk && text_node->text_for_rendering().is_empty()) {
+        if (!chunk_opt.has_value() && !m_text_node_context->added_trailing_empty_chunk) {
+            auto const is_completely_empty = m_text_node_context->chunks.is_empty();
+            bool creates_empty_trailing_line = false;
+            if (!is_completely_empty) {
+                auto const& last_passed_chunk = m_text_node_context->chunks.last();
+                creates_empty_trailing_line = m_text_node_context->should_respect_linebreaks && last_passed_chunk.has_breaking_newline;
+            }
+
+            if (is_completely_empty || creates_empty_trailing_line) {
                 if (auto const* shadow_root = as_if<DOM::ShadowRoot>(text_node->dom_node().root()))
                     if (auto const* form_associated_element = as_if<HTML::FormAssociatedTextControlElement>(shadow_root->host()))
                         is_empty_editable = form_associated_element->is_mutable();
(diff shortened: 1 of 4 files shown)


Libraries/LibWeb/Painting/PaintableWithLines.cpp
```diff
@@ -509,7 +509,19 @@ void PaintableWithLines::paint_cursor(DisplayListRecordingContext& context) cons
...
+        // This handles the case where the cursor is on an empty line (e.g. after pressing Enter).
+        CSSPixels cursor_y = content_box.y();
+        for (auto const& frag : m_fragments) {
+            if (frag.start_offset() + frag.length_in_code_units() <= cursor_position->offset()) {
+                auto frag_rect = frag.absolute_rect();
+                cursor_y = frag_rect.bottom();
+            }
+        }
+
+        cursor_rect = { content_box.x(), cursor_y, 1, line_height };
```
gmta: Can you add a regression test for this?

Libraries/LibWeb/Painting/PaintableWithLines.cpp
```diff
@@ -509,7 +509,19 @@ void PaintableWithLines::paint_cursor(DisplayListRecordingContext& context) cons
...
+
+        // Find the last fragment before the cursor offset to determine the Y position.
+        // This handles the case where the cursor is on an empty line (e.g. after pressing Enter).
+        CSSPixels cursor_y = content_box.y();
+        for (auto const& frag : m_fragments) {
+            if (frag.start_offset() + frag.length_in_code_units() <= cursor_position->offset()) {
+                auto frag_rect = frag.absolute_rect();
+                cursor_y = frag_rect.bottom();
+            }
+        }
```
gmta: I believe the approach we've chosen for other editable elements (such as `contenteditable`) is to make sure there's an empty fragment to be matched, which simplifies the cursor positioning code.

TechWizard9999: yeah this approach makes more sense. I'll rewrite it to ensure the empty fragment exists instead of the math fallback, and will include a test for it

Libraries/LibWeb/Painting/PaintableWithLines.cpp
TechWizard9999: Yeah sure

TechWizard9999:
@gmta Thanks for the review! I've removed the fallback math and rewritten the logic in `InlineLevelIterator.cpp` to ensure that a 0-width empty fragment is now natively emitted for trailing newlines in editable controls, similar to the `contenteditable` approach. I've also added a layout regression test (`textarea-trailing-newline-fragment.html`) to verify this behavior. Please let me know if there's anything else!

Libraries/LibWeb/Painting/PaintableWithLines.cpp
```diff
@@ -509,7 +509,8 @@ void PaintableWithLines::paint_cursor(DisplayListRecordingContext& context) cons
 
         caret_color = computed_values().caret_color();
         auto content_box = absolute_padding_box_rect();
-        cursor_rect = { content_box.x(), content_box.y(), 1, computed_values().line_height() };
+        auto line_height = computed_values().line_height();
+        cursor_rect = { content_box.x(), content_box.y(), 1, line_height };
```
gmta: This just introduces a useless variable; please revert.

TechWizard9999:
@gmta I have addressed your comments. If anything is required please let me know. I am eager to work on tasks like this, but since this is my first time, please guide me in one of the issues needed or in discussions.

Libraries/LibWeb/Painting/PaintableWithLines.cpp
TechWizard9999: @gmta  I have reverted the temporary line_height variable in PaintableWithLines.cpp and used computed_values().line_height() directly as requested. I've compiled the changes to ensure everything works properly. Please let me know if anything else is needed!

TechWizard9999:
@gmta Rebased on latest master. Could you please take another look when you have time?

TechWizard9999:
Hi @awesomekling , Could you please review this let me know if any changes needed

PR #8237, opened 2026-03-01
LibWeb: Fix stale image request callbacks corrupting newer requests

kalenikaliaksandr:
When img.src is changed rapidly (e.g., YouTube Music sets a GIF
placeholder then swaps to a real URL via IntersectionObserver), the
failure callback from the stale first request could corrupt the newer
request by calling abort_the_image_request on the now-reassigned
m_current_request.

Fix this by using the existing m_update_the_image_data_count generation
counter to detect stale fetch callbacks.

This fixes thumbnail loading on YouTube Music.

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLImageElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLImageElement.cpp
@@ -838,7 +838,7 @@ after_step_7:
         if (delay_load_event)
             m_load_event_delayer.emplace(document());
 
-        add_callbacks_to_image_request(*image_request, maybe_omit_events, *url_string, previous_url);
+        add_callbacks_to_image_request(*image_request, maybe_omit_events, *url_string, previous_url, update_the_image_data_count);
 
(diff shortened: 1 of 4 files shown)


Libraries/LibWeb/HTML/HTMLImageElement.cpp
```diff
@@ -825,9 +825,10 @@ void HTMLImageElement::update_the_image_data_impl(bool restart_animations, bool
 
         // 18. If the current request's state is unavailable or broken, then set the current request to image request.
         //     Otherwise, set the pending request to image request.
-        if (m_current_request->state() == ImageRequest::State::Unavailable || m_current_request->state() == ImageRequest::State::Broken)
+        if (m_current_request->state() == ImageRequest::State::Unavailable || m_current_request->state() == ImageRequest::State::Broken) {
+            abort_the_image_request(realm(), m_current_request);
```
gmta: This is AD-HOC code, "abort the image request" is not part of the spec step. Please mark it as such.

PR #8244, opened 2026-03-02
LibWeb: Implement `if()` CSS functions

Calme1709:
This comes with the caveat that we don't support `style()` conditions within these functions (only `supports()` and `media()`).


AtkinsSJ (approved):
:tada: Very cool!

PR #8246, opened 2026-03-02
LibWeb: Update layout after dispatching initial mouse events

gmta:
Since #8182, we've been verifying "is the layout up to date" in more places. After dispatching mouseup/pointerup, the layout could have been changed again and we need to update it before trying to run activation behaviors.

Fixes opening/closing an email's details in Roundcube webmail.

CC @awesomekling


awesomekling:
Yay it caught a bug! :D

PR #8251, opened 2026-03-02
LibWeb: Ensure execution context exists for FontFace loading task

tcl3:
Fixes a crash when opening: https://mozilla.github.io/pdf.js/web/viewer.html


shannonbooth:
not that it really matters much either way imo, but why not capture realm by reference? I probably would have written in that way

tcl3:
> not that it really matters much either way imo, but why not capture realm by reference? I probably would have written in that way

Actually, the first change isn't needed. I only made it because it matches the pattern used elsewhere in this method and I find these kinds of captures by reference a bit spooky. 

I've reverted this change.

shannonbooth:
I ask just thinking about if I should be changing the way I’ve been doing it, either way seems fine to me but something consistent is also nice

PR #8252, opened 2026-03-02
LibGC: Use AK::unwind_stack_from_frame_pointer for stack pointer traces

Zaggy1024:
This reintroduces AK/StackUnwinder.h and uses it to resolve stack pointers to frames. The provided function already functioned on x86, ARM and RISC-V, where the prior code in Heap.cpp only worked on the former two.

### DIFF
--- a/Libraries/LibGC/Heap.cpp
+++ b/Libraries/LibGC/Heap.cpp
@@ -208,10 +208,13 @@ public:
             auto node = AK::JsonObject();
             if (it.value.root_origin.has_value()) {
                 auto type = it.value.root_origin->type;
-                auto location = it.value.root_origin->location;
+                auto const* location = it.value.root_origin->location;
                 switch (type) {
                 case HeapRoot::Type::ConservativeVector:
                     node.set("root"sv, "ConservativeVector"sv);
+                    break;
+                case HeapRoot::Type::HeapFunctionCapturedPointer:
+                    node.set("root"sv, "HeapFunctionCapturedPointer"sv);
                     break;
                 case HeapRoot::Type::MustSurviveGC:
                     node.set("root"sv, "MustSurviveGC"sv);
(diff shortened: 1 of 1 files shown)


Libraries/LibGC/Heap.cpp
```diff
@@ -489,22 +493,23 @@ NO_SANITIZE_ADDRESS void Heap::gather_conservative_roots(HashMap<Cell*, HeapRoot
...
+            },
+            [&](AK::StackFrame frame) -> IterationDecision {
+                // Ensure the previous FP is above the current one (stack grows downward).
+                if (frame.previous_frame_pointer != 0 && frame.previous_frame_pointer <= current_fp)
+                    return IterationDecision::Break;
+                frame_starts.append(current_fp);
+                return_addresses.push_back(static_cast<cpptrace::frame_ptr>(frame.return_address) - 1);
+                current_fp = frame.previous_frame_pointer;
+                return IterationDecision::Continue;
+            });
```
Hendiadyoin1, line 512: Actually I wonder if the acquisition of the stack pointer could be done with a magic default argument at the end of the list, like
```c++
ErrorOr<void> unwind_stack_from_frame_pointer(CallableAs<ErrorOr<FlatPtr>, FlatPtr> auto read_memory, CallableAs<ErrorOr<IterationDecision>, StackFrame> auto on_stack_frame, FlatPtr frame_pointer = bit_cast<FlatPtr>(__builtin_frame_address(0))) { ... }
```

ADKaster: does that actually work? would need a test or a godbolt link to confirm. I thought that sort of thing only worked for consteval bultins like the ones for source location.

Hendiadyoin1: Seems to work <https://godbolt.org/z/nhxq4fTnY>

Zaggy1024: That would require changing how the unwind callbacks work, or adding an extra field to StackFrame, since we want the highest address on the stack for each frame, not the lowest. It doesn't call the callback with the initial pointer, which we make use of here.

I'm not inclined to refactor here, since this is a nice logical step at the very least, but it can't hurt to do it in another PR.

PR #8256, opened 2026-03-03
Meta: Add Rust lints to workspace Crago.toml file

juhotuho10:
Having the lints in a workspace configuration file is good because it can easily be applied to all Rust crates in the project. 
This will also raise a local compilation error if the CI would fail because of the lints instead of it being a local warning, but CI failure.

The second commit `Meta: Add more Rust lints to the workspace configuration file` relates to my previous PR: #8200 that applies these lints. But do know that my commit applying the lints has to be merged before this one, because otherwise this PR will just cause the CI to fail with the `cargo clippy` check.

Other option is the change the lints from `clippy::nursery` and `clippy::pedantic` in this PR to `warn`, then the lints would be a local and CI warning, but wouldn't fail local compilation or CI with only `cargo clippy -- -D clippy::all` being in the CI currently

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,3 +15,16 @@ lto = true
 
 [workspace.lints.clippy]
 all = "deny" # includes lintgroups: correctness, suspicious, style, complexity, perf
+# from clippy:nursery:
+redundant_clone = "deny"
+# from clippy:pedantic:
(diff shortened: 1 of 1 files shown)


juhotuho10:
The changes from here were moved to #8200

PR #8257, opened 2026-03-03
LibWeb: Update layout before accessing paintable in label activation

tcl3:
Another stale layout bug uncovered by #8182.

Found while browsing: https://employ.remote.com/sign-in/


Libraries/LibWeb/DOM/Document.h
```diff
@@ -99,6 +99,7 @@ enum class InvalidateLayoutTreeReason {
     X(HTMLElementOffsetParent)             \
     X(HTMLElementOffsetTop)                \
     X(HTMLElementOffsetWidth)              \
+    X(HTMLLabelElementActivationBehavior)  \
```
gmta: Let's keep these in alphabetical order please

PR #8263, opened 2026-03-03
 LibWeb: Update SubtleCrypto.encrypt() and decrypt() to match the spec

mikiubo:


mikiubo:
I kept the "LibWeb: Fix incorrect error message in SubtleCrypto::decrypt" commit separate to make sure it doesn’t get overlooked during review. I wasn’t fully confident about it, so I kept it separate to highlight it to the reviewer.

Libraries/LibWeb/Crypto/SubtleCrypto.cpp
```diff
@@ -133,115 +133,151 @@ WebIDL::ExceptionOr<NormalizedAlgorithmAndParameter> normalize_an_algorithm(JS::
...
-    // 5. Let promise be a new Promise.
+    // 6. Let promise be a new Promise.
     auto promise = WebIDL::create_promise(realm);
 
-    // 6. Return promise and perform the remaining steps in parallel.
-
-    Platform::EventLoopPlugin::the().deferred_invoke(GC::create_function(realm.heap(), [&realm, normalized_algorithm = normalized_algorithm.release_value(), promise, key, data = move(data)]() -> void {
+    // 7. Return promise and perform the remaining steps in parallel.
+    Platform::EventLoopPlugin::the().deferred_invoke(GC::create_function(realm.heap(), [&realm, &global, &heap, normalized_algorithm = normalized_algorithm.release_value(), promise, key, data = move(data)]() -> void {
         HTML::TemporaryExecutionContext context(realm, HTML::TemporaryExecutionContext::CallbacksEnabled::Yes);
```
gmta: Is this temporary execution context still necessary? 🤔

Tests/LibWeb/Text/expected/wpt-import/WebCryptoAPI/encrypt_decrypt/aes_gcm.https.any.txt
```diff
@@ -1,8 +1,8 @@
 Harness status: OK
 
-Found 325 tests
+Found 409 tests
```
gmta, line 3: Can you re-import these tests _before_ making any changes, and then rebaseline the results per change?

That will show the original +/- subtest counts on account of just reimporting and the subtest counts for the changes made in this PR.

Libraries/LibWeb/Crypto/SubtleCrypto.cpp
mikiubo: Good catch — I honestly didn't think about it enough and just kept what was there before. I'll look into whether the TemporaryExecutionContext (and especially CallbacksEnabled::Yes) is still needed in the outer lambda, given that reject/resolve now happen inside queued tasks that each have their own context.

mikiubo: ok i try whithout temporary execution context i take a crash because algorithm_name() calls JS::Object::get().
But i changed the CallbacksEnabled::Yes to CallbacksEnabled::No is not needed anymore
Let me know if is acceptable
Thanks

Tests/LibWeb/Text/expected/wpt-import/WebCryptoAPI/encrypt_decrypt/aes_gcm.https.any.txt
mikiubo: right is more clear

i think i have done

Issue #7512, opened 2026-01-18
Crash in IndexedDB when opening multiple databases concurrently


tcl3:
Happens when attempting to log in to https://linkedin.com

Mininal reproduction:
```html
<!DOCTYPE html>
<script>
for (var i = 0; i < 10; i++) {
    var request = indexedDB.open("multiple-open-test", 1);
}
</script>
```

Stack trace
```
#0   (inlined)          in operator() at /home/tim/repos/ladybird/Libraries/LibWeb/IndexedDB/Internal/RequestList.cpp:90:9
#1   0x000078e41b2af1df in call at /home/tim/repos/ladybird/AK/Function.h:225:20
#2   (inlined)          in operator() at /home/tim/repos/ladybird/AK/Function.h:148:25
#3   (inlined)          in notify_each_request_observer<(lambda at /home/tim/repos/ladybird/Libraries/LibWeb/IndexedDB/IDBRequest.cpp:130:34)> at /home/tim/repos/ladybird/Libraries/LibWeb/IndexedDB/IDBRequest.h:73:17
#4   0x000078e41b28a134 in set_processed at /home/tim/repos/ladybird/Libraries/LibWeb/IndexedDB/IDBRequest.cpp:130:5
#5   (inlined)          in operator() at /home/tim/repos/ladybird/Libraries/LibWeb/IndexedDB/IDBFactory.cpp:67:22
#6   0x000078e41b27daa3 in call at /home/tim/repos/ladybird/AK/Function.h:225:20
#7   0x000078e41b2a0cb3 in operator() at /home/tim/repos/ladybird/AK/Function.h:148:25
#8   (inlined)          in operator() at /home/tim/repos/ladybird/Libraries/LibWeb/IndexedDB/Internal/Algorithms.cpp:229:9
#9   0x000078e41b29b914 in call at /home/tim/repos/ladybird/AK/Function.h:225:20
#10  (inlined)          in operator() at /home/tim/repos/ladybird/AK/Function.h:148:25
#11  (inlined)          in operator() at /home/tim/repos/ladybird/Libraries/LibWeb/IndexedDB/Internal/RequestList.cpp:71:13
#12  0x000078e41b2aef37 in call at /home/tim/repos/ladybird/AK/Function.h:225:20
#13  (inlined)          in operator() at /home/tim/repos/ladybird/AK/Function.h:148:25
[...]

PR #8175, opened 2026-02-25
LibWeb: Queue a task when setting IDBRequest as processed

IntegralPilot:
This prevents a synchronous recursive loop that leads to an assertion
failure when many database requests are opened at the same time.

By queuing a task on the HTML event loop, the call stack is allowed
to unwind before the next request is processed.

Fixes #7512

PR #8264, opened 2026-03-04
LibWeb: Refactor IndexedDB to handle requests serially

Zaggy1024:
Previously, after one request was marked as processed, we would synchronously queue another task to process the next request. This would mean that two open requests on the same database could interleave. This was especially problematic when one of the requests would cause the database to upgrade, since the second open request would begin processing before the upgradeneeded event fired, causing an exception to be thrown in the second open().

The solution is to explicitly check for continuation conditions after events have been fired in order to ensure that every step for the request is completed before starting any further request processing.

For connection requests, the spec states:

> Open requests are processed in a connection queue. The queue contains all open requests associated with an storage key and a name. Requests added to the connection queue processed in order and each request must run to completion before the next request is processed. An open request may be blocked on other connections, requiring those connections to close before the request can complete and allow further requests to be processed.

For requests against a transaction, the spec states:

> Once the transaction has been started the implementation can begin executing the requests placed against the transaction. Requests must be executed in the order in which they were made against the transaction. Likewise, their results must be returned in the order the requests were placed against a specific transaction. There is no guarantee about the order that results from requests in different transactions are returned.

In the process of reworking it to use this approach, I've added a bunch of new tests that cover things that our imported WPTs weren't checking.

Nothing in our test suite has regressed with these changes. In fact, we get a couple of extra subtest passes on web platform tests IndexedDB/key-conversion-exceptions and IndexedDB/keypath-exceptions with the second commit.

With the fix for serializing connection requests, we can now fully download the assets for the emscripten-compiled asm.js games in the Humble Mozilla Bundle, particularly FTL: Faster Than Light.

Incidentally fixes https://github.com/LadybirdBrowser/ladybird/issues/7512, for which a crash test has been added.

### DIFF
--- a/Libraries/LibWeb/IndexedDB/Internal/Algorithms.cpp
+++ b/Libraries/LibWeb/IndexedDB/Internal/Algorithms.cpp
@@ -159,12 +159,17 @@ void open_a_database_connection(JS::Realm& realm, StorageAPI::StorageKey storage
             auto after_all = GC::create_function(realm.heap(), [&realm, &queue, open_connections, db, version, connection, request, on_complete] {
                 // 4. If any of the connections in openConnections are still not closed,
                 //    queue a database task to fire a version change event named blocked at request with db’s version and version.
-                for (auto const& entry : open_connections->elements()) {
-                    if (entry->state() != ConnectionState::Closed) {
-                        queue_a_database_task(GC::create_function(realm.vm().heap(), [&realm, entry, db, version]() {
-                            fire_a_version_change_event(realm, HTML::EventNames::blocked, *entry, db->version(), version);
-                        }));
+                auto any_connection_is_not_closed = [&] {
+                    for (auto const& entry : open_connections->elements()) {
+                        if (entry->state() != ConnectionState::Closed)
+                            return true;
                     }
+                    return false;
+                }();
+                if (any_connection_is_not_closed) {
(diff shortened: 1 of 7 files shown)


InvalidUsernameException:
Hmm, does this supersede #8175 (and fix #7512)?

Zaggy1024:
It looks like it happens to, though I didn't know about that issue before tackling this.

Zaggy1024:
Gonna try and import all the WPT IndexedDB tests in a first commit so we can see the diff in this PR before merging.

tcl3:
> Gonna try and import all the WPT IndexedDB tests in a first commit so we can see the diff in this PR before merging.

There are a few [flaky IndexedDB tests](https://wpt.fyi/results/IndexedDB?label=master&label=experimental&max-count=10&product=ladybird&q=seq%28%28status%3APASS%7Cstatus%3AOK%29%20%28status%3A%21PASS%26status%3A%21OK%26status%3A%21unknown%29%29%20seq%28%28status%3A%21PASS%26status%3A%21OK%26status%3A%21unknown%29%20%28status%3APASS%7Cstatus%3AOK%29%29). If you're going to add all the tests at once, be sure to run them a good number of times locally, so any flaky tests can be excluded.

Zaggy1024:
I misread how many tests there are on wpt.fyi, I'll probably have to look at bringing in only a subset of them. 😅 Running 800 new tests with sanitizers might just make us start timing out a bunch...

For now, I'll just see if I can make a summary of the diff before and after.

PR #8265, opened 2026-03-04
LibWeb: Bump WebGL depth bit depth to 24, handle some context options

Zaggy1024:
WebGL now uses a 24 bits for its depth component, matching what Firefox, Chromium and Safari supply by default. Along with #8264, this allows the asm.js version of FTL to run in Ladybird!

Screenshot 2026-03-04 at 02 26 06

WebGLContext and WebGL2Context now pass the depth, stencil and antialias flags that the WebGL spec mention as drawing buffer options, and OpenGLContext uses them to conditionally enable depth and stencil.

The antialias option is currently unused.

### DIFF
--- a/Libraries/LibWeb/WebGL/WebGLRenderingContext.cpp
+++ b/Libraries/LibWeb/WebGL/WebGLRenderingContext.cpp
@@ -64,7 +64,12 @@ JS::ThrowCompletionOr<GC::Ptr<WebGLRenderingContext>> WebGLRenderingContext::cre
         fire_webgl_context_creation_error(canvas_element);
         return GC::Ptr<WebGLRenderingContext> { nullptr };
     }
-    auto context = OpenGLContext::create(*skia_backend_context, OpenGLContext::WebGLVersion::WebGL1);
+    OpenGLContext::DrawingBufferOptions context_options {
+        .depth = context_attributes.depth,
+        .stencil = context_attributes.stencil,
+        .antialias = context_attributes.antialias,
+    };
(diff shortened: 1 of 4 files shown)


Libraries/LibWeb/WebGL/WebGLRenderingContext.cpp
```diff
@@ -57,14 +57,32 @@ void fire_webgl_context_creation_error(HTML::HTMLCanvasElement& canvas_element)
...
+        auto& object = options.as_object();
+        dbgln("WebGLRenderingContext::create() raw options: alpha={}, depth={}, stencil={}, antialias={}, premultipliedAlpha={}, preserveDrawingBuffer={}",
+            TRY(object.get("alpha"_utf16_fly_string)),
+            TRY(object.get("depth"_utf16_fly_string)),
+            TRY(object.get("stencil"_utf16_fly_string)),
+            TRY(object.get("antialias"_utf16_fly_string)),
+            TRY(object.get("premultipliedAlpha"_utf16_fly_string)),
+            TRY(object.get("preserveDrawingBuffer"_utf16_fly_string)));
+    } else {
+        dbgln("WebGLRenderingContext::create() options: nullish");
```
gmta: Can we put these `dbgln`s behind a debug flag to prevent noise?

Zaggy1024: Oops, those weren't supposed to stay at all!

gmta:
@Zaggy1024 Windows CI has an issue.

PR #8268, opened 2026-03-04
Meta: Enable angled brackets in clangd

R-Goc:
This commit enables angled includes on completion for headers under Libraries and AK to match our coding style.

### DIFF
--- a/.clangd
+++ b/.clangd
@@ -4,3 +4,6 @@ CompileFlags:
 Diagnostics:
   UnusedIncludes: None
   MissingIncludes: None
+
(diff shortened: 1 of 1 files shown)


.clangd
```diff
@@ -4,3 +4,6 @@ CompileFlags:
 Diagnostics:
   UnusedIncludes: None
   MissingIncludes: None
+
+Style:
+  AngledHeaders: ["AK/.*", "Lib.*"]
```
trflynn89: This should also at least include `Services`

PR #8279, opened 2026-03-05
WebContent: Enable IOSurfaces on Intel macOS

christianfrey:
Without IOSurface, the Metal rendering path introduced in #7956 hits a VERIFY(iosurface_ref) failure and crashes on launch on Intel Macs. The FIXME stated that the implementation of IOSurface does not work on Intel macOS, but testing confirms it now works correctly.

### DIFF
--- a/Services/WebContent/main.cpp
+++ b/Services/WebContent/main.cpp
@@ -177,13 +177,8 @@ ErrorOr<int> ladybird_main(Main::Arguments arguments)
 
 #if defined(AK_OS_MACOS)
     if (!mach_server_name.is_empty()) {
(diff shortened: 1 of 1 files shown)


trflynn89:
Disabling IOSurface here was a result of #1216. If that's no longer an issue on Intel macOS, then that's great! Just want to confirm that's the case, you're able to browse fine and see rendered content with with path? I no longer have access to an Intel mac myself to test.

christianfrey:
Yes, I can confirm it works correctly on my Intel Mac. I've tested browsing multiple sites and everything renders fine: no crashes, no blank pages.

PR #8282, opened 2026-03-05
LibWeb: Increment fetch generation counter on fetching process cancel

acrilique:
This workaround avoids a crash that occurs when various fetching processes are quickly started and canceled. Although a better solution would be to atually remove the body callbacks when the fetch is stopped, this works for now.

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLMediaElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLMediaElement.cpp
@@ -197,6 +197,7 @@ void HTMLMediaElement::removed_from(DOM::Node* old_parent, DOM::Node& old_root)
 
 void HTMLMediaElement::cancel_the_fetching_process()
 {
+    m_current_fetch_generation++;
     m_fetch_data.clear();
 }
 


rcorsi:
Do you have any logs or coredump backtraces of the failure?  Any reproduction methods that you have? Thank you!

acrilique:
Yes sorry, I gave that info on Discord (media channel) but didn't put it here.

> Hello! I haven't contributed to the project but I compile it from time to time and test a few websites. I've been testing Youtube today and got a crash when hovering over videos in the main page (when you hover, the video is previewed, so the crash eventually happens when you preview one video, then another, then another one...). Here's the beginning of the stack trace:
> ```
> VERIFICATION FAILED: m_ptr at /home/llucsm/dev/cpp/ladybird/AK/OwnPtr.h:139
> Stack trace (most recent call first):
> #0  (inlined)          in AK::OwnPtr<Web::HTML::HTMLMediaElement::FetchData, AK::DefaultDelete<Web::HTML::HTMLMediaElement::FetchData> >::operator->() const at /home/llucsm/dev/cpp/ladybird/AK/OwnPtr.h:139:9
> #1  (inlined)          in operator() at /home/llucsm/dev/cpp/ladybird/Libraries/LibWeb/HTML/HTMLMediaElement.cpp:1176:17
> #2  0x00007f852909878b in call at /home/llucsm/dev/cpp/ladybird/AK/Function.h:224:30
> ```
> I've digged a bit and it seems the issue consists in the following: when hovering a video, `fetch_resource()` is called which registers the `update_media()` callback. The latter has a defensive null check for `weak_self` and also `fetch_generation != weak_self->m_current_fetch_generation`. When we stop hovering that video and hover another one, `cancel_the_fetching_process()` is called, which calls `m_fetch_data.clear()`, but doesn't increment the value of `m_current_fetch_generation`. Right after, `update_media()` from the last `fetch_resource()` is called and given that `m_fetch_data` is now cleared and the defensive checks still pass, the code inside `update_media()` runs crashes. 
> 
> So there are a couple workarounds: also increment `m_current_fetch_generation` inside `cancel_the_fetching_process()`, or add another defensive null check in `update_media()`: `!weak_self->m_fetch_data`. However, these do not address the underlying question I'm asking myself: is there any good reason why `update_media()` isn't being properly canceled?

Let me know if you think that I should paste the full stack trace.

Zaggy1024:
Could you try force pushing to this branch? GitHub doesn't give me the option to enable workflows here.

Also, the commit summary line could probably stand to mention that this is for the media element, at first I wondered if this was also affecting the image element 😅

acrilique:
@Zaggy1024 done!

Issue #8276, opened 2026-03-05
GNU 15 fails to compile AK/kmalloc.h after recent refactor in LibJS


rcorsi:
Using Ubuntu 25.10 with GNU GCC 15.2.0

Following error is reported during build using `./Meta/ladybird.py build`

```
[0/2502] cd /home/bobo/projects/myladybird/Libraries/LibJS/Rust && /usr/bin/cmake -.../myladybird/Build/release/./cargo/build --release -- -Cdefault-linker-libraries=yes
   Compiling libjs_rust v0.1.0 (/home/bobo/projects/myladybird/Libraries/LibJS/Rust)
    Finished `release` profile [optimized] target(s) in 5.23s
[649/2502] Building CXX object Lagom/Libraries/LibJS/CMakeFiles/LibJS.dir/Bytecode/Interpreter.cpp.o
FAILED: Lagom/Libraries/LibJS/CMakeFiles/LibJS.dir/Bytecode/Interpreter.cpp.o 
[...]

PR #8283, opened 2026-03-05
LibJS: Fix GCC build

acrilique:
GCC inlines the `ec_stack.size() - 1` subtraction and then fails to build because that operation would result in a wraparound if `size()` was zero. This fixes the build by asserting that `ec_stack` isn't empty.


rcorsi:
Have a look at #8287, is related to this issue.

rcorsi:
Both this PR and PR #8287 fixes the issue.

@acrilique Edit the main description text and add at the end: fixes #8276
This will link your PR and the issue, and if the PR is merged, will close the issue at the same time.

gmta:
Superseded by #8287. Thank you @acrilique!

PR #8284, opened 2026-03-05
LibWeb: Bring create_element more in line with the spec

Psychpsyo:
No actual behavioral changes, just new spec comments and a spec assert.

### DIFF
--- a/Libraries/LibWeb/DOM/ElementFactory.cpp
+++ b/Libraries/LibWeb/DOM/ElementFactory.cpp
@@ -606,17 +606,18 @@ WebIDL::ExceptionOr<GC::Ref<Element>> create_element(Document& document, FlyStri
     // 1. Let result be null.
     // NOTE: We collapse this into just returning an element where necessary.
 
-    // 2. Let definition be the result of looking up a custom element definition given document, namespace, localName, and is.
+    // 2. If registry is "default", then set registry to the result of looking up a custom element registry given document.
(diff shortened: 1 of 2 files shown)


Libraries/LibWeb/DOM/ElementFactory.cpp
```diff
@@ -721,50 +733,51 @@ WebIDL::ExceptionOr<GC::Ref<Element>> create_element(Document& document, FlyStri
...
-        element->set_is_value(is_value);
-        element->set_custom_element_state(CustomElementState::Uncustomized);
-        return element;
+// https://dom.spec.whatwg.org/#create-an-element-internal
+// FIXME: This needs to take interface and registry, like the spec says.
+GC::Ref<Element> create_element_internal(Document& document, FlyString local_name, Optional<FlyString> namespace_, Optional<FlyString> prefix, CustomElementState state, Optional<String> is_value)
+{
+    auto& realm = document.realm();
+
+    // 1. Let element be a new element that implements interface, with namespace set to namespace, namespace prefix set to prefix, local name set to localName, custom element registry set to registry, custom element state set to state, custom element definition set to null, is value set to is, and node document set to document.
```
shannonbooth: Nit - could we wrap this at 120 lines?

Psychpsyo: Done.

PR #8285, opened 2026-03-05
LibWeb: Refactor the way WebGL extensions are handled

cqundefine:


Libraries/LibWeb/WebGL/WebGLRenderingContextBase.cpp
```diff
@@ -322,38 +226,30 @@ void WebGLRenderingContextBase::enable_compressed_texture_format(WebIDL::Unsigne
...
-    visitor.visit(m_oes_element_index_uint_object_extension);
-    visitor.visit(m_oes_standard_derivatives_object_extension);
-    visitor.visit(m_oes_vertex_array_object_extension);
-    visitor.visit(m_webgl_compressed_texture_s3tc_extension);
-    visitor.visit(m_webgl_compressed_texture_s3tc_srgb_extension);
-    visitor.visit(m_webgl_draw_buffers_extension);
+
+    for (auto const& [name, extension] : m_enabled_extensions) {
+        visitor.visit(extension);
+    }
```
gmta: Are you not able to do `visitor.visit(m_enabled_extensions)` here on account of `Visitor::visit(HashMap<K, V, T> const& map)`?

Libraries/LibWeb/WebGL/WebGL2RenderingContextImpl.cpp
```diff
@@ -810,7 +810,7 @@ void WebGL2RenderingContextImpl::sampler_parameteri(GC::Root<WebGLSampler> sampl
     case GL_TEXTURE_WRAP_T:
         break;
     case GL_TEXTURE_MAX_ANISOTROPY_EXT: {
-        if (ext_texture_filter_anisotropic_extension_enabled())
+        if (extension_enabled("EXT_texture_filter_anisotropic"sv))
```
gmta, line 813: We might want to adopt our FlyString pattern here, like we do with HTML::TagNames or ::AttributeNames. Prevents us from repeating all these constant strings everywhere.

Libraries/LibWeb/WebGL/Extensions/WebGLCompressedTextureS3tc.h
```diff
@@ -16,7 +16,7 @@ class WebGLCompressedTextureS3tc : public Bindings::PlatformObject {
     GC_DECLARE_ALLOCATOR(WebGLCompressedTextureS3tc);
 
 public:
-    static JS::ThrowCompletionOr<GC::Ptr<WebGLCompressedTextureS3tc>> create(JS::Realm&, GC::Ref<WebGLRenderingContextBase>);
+    static JS::ThrowCompletionOr<GC::Ref<JS::Object>> create(JS::Realm&, GC::Ref<WebGLRenderingContextBase>);
```
gmta, line 19: This commit cannot be compiled on its own:

```
/ladybird/Libraries/LibWeb/WebGL/WebGLRenderingContextBase.cpp:289:55: error: no viable overloaded '='
  289 |             m_webgl_compressed_texture_s3tc_extension = MUST(Extensions::WebGLCompressedTextureS3tc::create(realm(), *this));
      |             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Libraries/LibWeb/WebGL/WebGLRenderingContextBase.cpp
cqundefine: I actually didn't know that was possible

Libraries/LibWeb/WebGL/WebGL2RenderingContextImpl.cpp
cqundefine: Would it be fine to leave it a TODO for now as? It gets quite tricky with all the places that would require changing, especially with the hash maps that have ASCIICaseInsensitiveStringTraits.

gmta: Sure, add a `// FIXME: ..` somewhere that makes sense :)

cqundefine: I added a FIXME above the `extension_enabled` function declaration in the header

cqundefine:
@gmta Could you merge through the CI failure? It seems to be an unrelated flake, some fullscreen test failing.

PR #8290, opened 2026-03-06
LibWasm: Implement `call_ref` and `return_call_ref`

gmta:
Cousins of `call_indirect` and `return_call_indirect`; the ref type is already known and does not require a table lookup.

Makes https://wordsalad.online/ work!

image

### DIFF
--- a/Libraries/LibWasm/AbstractMachine/BytecodeInterpreter.cpp
+++ b/Libraries/LibWasm/AbstractMachine/BytecodeInterpreter.cpp
@@ -2013,8 +2013,6 @@ HANDLE_INSTRUCTION(call_indirect)
     auto address = element.ref().template get<Reference::Func>().address;
     auto const& type_actual = configuration.store().get(address)->visit([](auto& f) -> decltype(auto) { return f.type(); });
     auto const& type_expected = configuration.frame().module().types()[args.type.value()].unsafe_function();
-    TRAP_IN_LOOP_IF_NOT(type_actual.parameters().size() == type_expected.parameters().size());
-    TRAP_IN_LOOP_IF_NOT(type_actual.results().size() == type_expected.results().size());
     TRAP_IN_LOOP_IF_NOT(type_actual.parameters() == type_expected.parameters());
     TRAP_IN_LOOP_IF_NOT(type_actual.results() == type_expected.results());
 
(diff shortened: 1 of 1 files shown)


Libraries/LibWasm/Parser/Parser.cpp
```diff
@@ -119,8 +119,11 @@ static ParseResult<ValueType> parse_reference_type(Stream& stream, u8 tag)
         return ValueType(ValueType::UnsupportedHeapReference);
     case Constants::nullable_reference_tag_tag:
     case Constants::non_nullable_reference_tag_tag: {
+        bool nullable = (tag == Constants::nullable_reference_tag_tag);
```
alimpfard: parens are extra?

Libraries/LibWasm/Types.h
```diff
@@ -164,7 +164,7 @@ class ReconsumableStream : public Stream {
 // https://webassembly.github.io/spec/core/bikeshed/#value-types%E2%91%A2
 class ValueType {
 public:
-    enum Kind {
+    enum Kind : u8 {
```
alimpfard, line 167: Not that I mind the change, but it has nothing to do with this commit.

Libraries/LibWasm/AbstractMachine/Validator.h
```diff
@@ -37,7 +38,6 @@ struct Context {
     RefPtr<RefRBTree> references { make_ref_counted<RefRBTree>() };
     size_t imported_function_count { 0 };
     size_t current_function_parameter_count { 0 };
-    Module const* current_module { nullptr };
```
alimpfard, line 40: I guess this was unused? meh, looks ok as-is I guess.

alimpfard:
CI seems to have made ASAN unhappy: https://github.com/LadybirdBrowser/ladybird/actions/runs/22761643629/job/66019388453?pr=8290#step:18:29

Libraries/LibWasm/Types.h
gmta: I noticed ValueType growing from 12 to 16 bytes by introducing nullability; this prevents that.

Libraries/LibWasm/AbstractMachine/Validator.h
gmta: Oops - that should go in a separate commit!

Libraries/LibWasm/Types.h
alimpfard: Ah. Sounds good then.

gmta:
> CI seems to have made ASAN unhappy: https://github.com/LadybirdBrowser/ladybird/actions/runs/22761643629/job/66019388453?pr=8290#step:18:29

Oof had a dangling reference. Some scope trickery later to deal with the tail call and we should be good now.

PR #8292, opened 2026-03-06
LibWeb: Distribute grid track free space among unfrozen tracks only

tcl3:
Previously, during grid track maximization we used the total number of grid tracks as the denominator when distributing the available free space. Using a larger denominator than expected meant that the algorithm needed a larger number of iterations than necessary to converge leading to rounding errors in the final track sizes.

Gains us +12 WPT subtests in `css/css-grid`.

Improves rendering of the "Newsroom" nav bar on https://www.apple.com/newsroom/2026/03/say-hello-to-macbook-neo/:

Before:
image

After:
image

### DIFF
--- a/Libraries/LibWeb/Layout/GridFormattingContext.cpp
+++ b/Libraries/LibWeb/Layout/GridFormattingContext.cpp
@@ -1144,13 +1144,25 @@ void GridFormattingContext::maximize_tracks_using_available_size(AvailableSpace 
 
     // If the free space is positive, distribute it equally to the base sizes of all tracks, freezing
     // tracks as they reach their growth limits (and continuing to grow the unfrozen tracks as needed).
-    while (free_space_px > 0) {
-        auto free_space_to_distribute_per_track = free_space_px / tracks.size();
+    size_t growable_track_count = 0;
+    for (auto& track : tracks) {
+        if (track.base_size_frozen)
+            continue;
+        VERIFY(track.growth_limit.has_value());
+        if (track.base_size < track.growth_limit.value())
+            growable_track_count++;
+    }
+    while (free_space_px > 0 && growable_track_count > 0) {
+        auto free_space_to_distribute_per_track = free_space_px / growable_track_count;
         for (auto& track : tracks) {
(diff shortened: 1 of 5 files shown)


warpdesign:
I suppose before & after are reversed?

tcl3:
> I suppose before & after are reversed?

Whoops :sweat_smile:. I've fixed it now, thanks!

Libraries/LibWeb/Layout/GridFormattingContext.cpp
```diff
@@ -1145,7 +1145,18 @@ void GridFormattingContext::maximize_tracks_using_available_size(AvailableSpace
...
     while (free_space_px > 0) {
-        auto free_space_to_distribute_per_track = free_space_px / tracks.size();
+        size_t growable_track_count = 0;
+        for (auto& track : tracks) {
+            if (track.base_size_frozen)
+                continue;
+            VERIFY(track.growth_limit.has_value());
+            if (track.base_size < track.growth_limit.value())
+                growable_track_count++;
+        }
```
gmta: Instead of repeatedly counting the growable tracks, couldn't we do this once and do `--growable_track_count` once we hit a track's limit in the loop below?

Libraries/LibWeb/Layout/GridFormattingContext.cpp
```diff
@@ -1144,13 +1144,26 @@ void GridFormattingContext::maximize_tracks_using_available_size(AvailableSpace
...
+        VERIFY(track.growth_limit.has_value());
+        if (track.base_size < track.growth_limit.value())
+            growable_track_count++;
+    }
+    while (free_space_px > 0 && growable_track_count > 0) {
+        auto free_space_to_distribute_per_track = free_space_px / growable_track_count;
         for (auto& track : tracks) {
             if (track.base_size_frozen)
                 continue;
             VERIFY(track.growth_limit.has_value());
```
gmta: This `VERIFY` has become superfluous and can be removed, because you already do this check in the earlier loop for all tracks that are not frozen.

PR #8293, opened 2026-03-06
LibWeb: Invalidate caret paint cache on blink cycle reset

tcl3:
When the text caret blink cycle is reset we now invalidate the paint cache  for the previous and current caret position. This ensures the caret is painted immediately when the blink cycle is reset, as well as ensuring the previous caret is no longer drawn.


tcl3:
Superseded by #8366

PR #8305, opened 2026-03-07
LibWeb: Match nth-child pseudo-classes on elements without a parent

adamjcolvin:
## Summary

The `:nth-child()`, `:nth-last-child()`, `:nth-of-type()`, and `:nth-last-of-type()` pseudo-classes were returning`false` when matched against elements that have no parent node (for example, a standalone element created via `document.createElement()`). Per the [Selectors Level 4 child-index definition](https://drafts.csswg.org/selectors-4/#child-index), an element with no siblings should be at index 1. The simpler child-indexed pseudo-classes (`:first-child`, `:last-child`, `:only-child`, etc.) already handle this case.

Screenshot 2026-03-07 at 11 58 26

## Implementation Approach

The nth-* matching code in `SelectorEngine.cpp` had an early `return false` when `element.parent()` was null. Removing this check allows the index to remain at 1 (no siblings to count) and lets the `An+B` pattern check determine the result.

Since parent is no longer guaranteed to be non-null, the `NthChild` and `NthLastChild` cases now have a guard against that before we start iterating over the siblings. The `NthOfType` and `NthLastOfType` cases use sibling traversal functions that naturally return `nullptr` for parentless elements, so they needed no changes.

```cpp
// Before: early return prevented matching on parentless elements
auto const* parent = element.parent();
if (!parent)
    return false;

// After: guard only the loops that dereference parent
auto const* parent = element.parent();
// ...
if (!parent)
    break;
for (auto* child = parent->first_child_of_type<DOM::Element>(); ...) { ... }
```

## Testing

This fixes 8 subtests in the `css/selectors/child-indexed-pseudo-class.html` WPT test, bringing it from 46/54 to 54/54.

## Changes

- `Libraries/LibWeb/CSS/SelectorEngine.cpp`
- `Tests/LibWeb/Text/expected/wpt-import/css/selectors/child-indexed-pseudo-class.txt`

### DIFF
--- a/Libraries/LibWeb/CSS/SelectorEngine.cpp
+++ b/Libraries/LibWeb/CSS/SelectorEngine.cpp
@@ -679,8 +679,6 @@ static inline bool matches_pseudo_class(CSS::Selector::SimpleSelector::PseudoCla
     case CSS::PseudoClass::NthOfType:
     case CSS::PseudoClass::NthLastOfType: {
         auto const* parent = element.parent();
-        if (!parent)
-            return false;
 
         if (context.collect_per_element_selector_involvement_metadata) {
             auto& mutable_element = const_cast<DOM::Element&>(element);
@@ -704,6 +702,9 @@ static inline bool matches_pseudo_class(CSS::Selector::SimpleSelector::PseudoCla
             return list.contains([&](auto const& selector) { return matches(selector, element, shadow_host, context); });
         };
 
+        // https://drafts.csswg.org/selectors-4/#child-index
+        // The pseudo-classes defined in this section select elements based on their index amongst their inclusive siblings.
+        // NB: An element without a parent has no siblings, so its index is 1.
         int index = 1;
         switch (pseudo_class.type) {
         case CSS::PseudoClass::__Count:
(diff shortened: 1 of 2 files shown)


Libraries/LibWeb/CSS/SelectorEngine.cpp
```diff
@@ -704,25 +702,32 @@ static inline bool matches_pseudo_class(CSS::Selector::SimpleSelector::PseudoCla
...
             if (!matches_selector_list(pseudo_class.argument_selector_list, element))
                 return false;
-            for (auto* child = parent->last_child_of_type<DOM::Element>(); child && child != &element; child = child->previous_element_sibling()) {
-                if (matches_selector_list(pseudo_class.argument_selector_list, *child))
-                    ++index;
+            if (parent) {
+                for (auto* child = parent->last_child_of_type<DOM::Element>(); child && child != &element; child = child->previous_element_sibling()) {
+                    if (matches_selector_list(pseudo_class.argument_selector_list, *child))
+                        ++index;
+                }
```
gmta: My preference here would be to just do:

```cpp
if (!parent)
    break;
```

adamjcolvin: Yeah makes sense, that's nicer. I've updated the commit, thanks.

AtkinsSJ (approved):
Looks good to me! Will need someone to restart the failed Linux CI once the Windows one completes.

gmta (approved):

PR #8307, opened 2026-03-07
LibThreading: Create separate Mutex and RecursiveMutex classes

marcbutler:
Prevent the use of recursive mutexes with condition variables.

Add support for enabling mutex deadlock detection, both selectively and globally.


R-Goc:
#6537 did that and more. However the thing is that the last time I checked the mutexes in Audio/VideoDataProvider had to be recursive or there would be deadlocks,

marcbutler:
> #6537 did that and more. However the thing is that the last time I checked the mutexes in Audio/VideoDataProvider had to be recursive or there would be deadlocks,

Understood. My focus in this change was to identify and address usage of recursive mutexes with condition variables. After testing I found that recursive mutexes could be retained in the DataProviders avoiding deadlocks in the handlers. For AudioMixingSink the condition variable is unused, so there is no conflict in replacing it with a recursive mutex.

As #6537 predates this PR but is not yet landed, I will seek feedback as to how to turn this PR into a useful contribution.

Zaggy1024:
I've been meaning to remove the recursive locks from LibMedia for a while (I've removed some cases already). I'll probably grab this branch locally and remove them, then look to merge @R-Goc 's PR, if that seems reasonable.

marcbutler:
@R-Goc I am seeing an intermittent LibWeb unit test failure `Text/input/HTML/autoplay-of-file-video-from-file-document.html`, when I rebase this pr on the tip of master. The failure is rare and I'm waiting to catch it with MUTEX_DEBUG.

@Zaggy1024 The mutex changes in this PR are an incompatible subset of what is provided in #6537 . However, the media library changes can be reworked to use the LibSync API. I'd be happy to separate those into a separate PR if you'd like.

ShaneenHall26:
I see your note. The email you are currently viewing from Marc Butler is
actually about the following:

   - *Intermittent Test Failure*: Marc Butler reported an intermittent

   LibWeb

    unit test failure (

   Text/input/HTML/autoplay-of-file-video-from-file-document.html

   ) after rebasing the current PR on master.
   - *PR Compatibility*: He noted that the mutex changes in this PR (

   #8307

   ) are an incompatible subset of what's provided in another PR (

   #6537

   ).
   - *Action Suggestion*: He offered to separate the media library changes
   into a new PR, as they can be reworked to use the

   LibSync

    API.

Zaggy1024:
> The mutex changes in this PR are an incompatible subset of what is provided in #6537 . However, the media library changes can be reworked to use the LibSync API. I'd be happy to separate those into a separate PR if you'd like.

Maybe I'm misunderstanding something here, but isn't the only change that would be relevant to the other PR to remove the condition variable from AudioMixingSink? Making a separate PR for that would definitely not be bad, but I'm not sure if that's what you're referring to here.

marcbutler:
@Zaggy1024 No you're correct. 

When I wrote that response I had it in my mind that maybe the build option might also be useful, forgetting that ERRORCHECK is used by default for non-recursive mutexes. And I don't intend to waste anyone's time with factoring out a one-line PR.

R-Goc:
Yeah, I was thinking of adding the build option to my PR. I tried searching for what the overhead of the error check mutexes is and didn't find anything definitive. I read some pthread source code and it didn't seem overly expensive so I think it should be fine. If it shows up in traces I'd argue we have a larger issue.

marcbutler:
@R-Goc  I understand the overhead is highly optimized, and it seems easy enough to patch-n-build to test  any impact. Though imho the semantics are subtle, I'm not advocating for deadlocks, but callers do not expect mutexes to fail, and it inspires the implementation of a re-try protocol, which then results in rarified and hard to isolate bugs. But that's the game right? :)

@Zaggy1024 In addition to the one line, this PR identifies which Mutexes need to maintain the recursive property.

Also I am open to collaborating. Thanks.

R-Goc:
The only place where error check mutexes return an error is when the normal mutex would have undefined behavior if I recall correctly.

marcbutler:
@R-Goc Yes my mistake the change forces UB on deadlock.

PR #8308, opened 2026-03-07
LibWeb: Invalidate paint cache for all layout node paintables on repaint

tcl3:
Previously, when clicking a href that spanned over multiple lines only part of the link was being repainted.

This was seen on multiple websites where hovering over links changed the link style in some way.

### DIFF
--- a/Tests/LibWeb/Text/input/display_list/multiline-inline-paint-invalidation.html
+++ b/Tests/LibWeb/Text/input/display_list/multiline-inline-paint-invalidation.html
@@ -1,0 +1,26 @@
+<!doctype html>
+<script src="../include.js"></script>
+<style>
+    div {
+        width: 100px;
+        font-size: 16px;
+    }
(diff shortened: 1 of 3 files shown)


Tests/LibWeb/Text/input/display_list/multiline-inline-paint-invalidation.html
```diff
@@ -0,0 +1,21 @@
...
+    }
+</style>
+<div>
+    <a id="link" href="#">This link spans multiple lines</a>
+</div>
+<script>
+    test(() => {
+        internals.dumpDisplayList();
+        document.getElementById("link").style.color = "red";
+        println(internals.dumpDisplayList());
```
gmta: Could be interesting to have both display lists in the output - maybe store the result of the first `internals.dumpDisplayList();` and then perform a single `println()`?

PR #8310, opened 2026-03-07
LibJS+LibWeb: Optimize membership checks in hot paths

ijxpwastaken:
Replace repeated contains_slow() scans with faster lookups in
several frequently executed paths.

LibJS:
Use a HashTable when building the JSON replacer list to avoid
repeated linear Vector membership checks while preserving order.

LibWeb CSS:
Precompute a FixedBitmap filter for properties_to_cascade so the
cascade loop can skip properties without calling contains_slow().

LibWeb HTML:
Cache sandbox token checks in local booleans to avoid repeated
scans of the token vector when setting sandbox flags.

These changes reduce repeated O(n) membership scans in hot paths
without changing behavior.


Zaggy1024:
No need to close your PRs whenever you need to fix something. If it's one commit, just use `git commit --amend` as the bot mentioned, and when you need to get up-to-date with master or fix something in a commit before the last one, `git rebase` is your friend: https://www.youtube.com/watch?v=ElRzTuYln0M

PR #8312, opened 2026-03-07
LibCore+LibWeb+UI+Services: Send notifications on linux

NiccoloAntonelliDziri:
Add a PlatformNotification abstraction and a Linux implementation that sends desktop notifications over the session D‑Bus (org.freedesktop.Notifications).

Uses IPC calls and the Dbus logic to send only the title.

The structure of the `PlatformNotificationImpl` and `PlatformNotification` classes are inspired by the `TimeZoneWatcher ` classes, for the cross-platform implementation.

Some explanations for the choices that I made:
- Using dbus/dbus.h instead of a dbus library does not add a new dependency since dbus/dbus.h was already used in the project.
- The platform-implementation of notifications resides in its entirety inside of LibCore to centralize the different platforms.
- LibCore instead of directly in the UI folders because Qt is used for both Linux and Windows and they handle notifications differently. Having it in centralized way simplifies the cross-platform logic.

Everything is up for debate of course and everything can change.

### DIFF
--- a/Libraries/LibCore/CMakeLists.txt
+++ b/Libraries/LibCore/CMakeLists.txt
@@ -83,6 +83,12 @@ else()
     list(APPEND SOURCES TimeZoneWatcherUnimplemented.cpp)
 endif()
 
+if(LINUX OR BSD)
+    list(APPEND SOURCES NotificationUnix.cpp)
+else()
+    list(APPEND SOURCES NotificationUnimplemented.cpp)
+endif()
+
 if (APPLE OR CMAKE_SYSTEM_NAME STREQUAL "GNU")
     list(APPEND SOURCES MachPort.cpp)
 endif()
@@ -123,4 +129,10 @@ endif()
 
 if (ANDROID)
     target_link_libraries(LibCore PRIVATE log)
+endif()
+
+if (LINUX OR BSD)
+    find_package(PkgConfig REQUIRED)
+    pkg_check_modules(dbus REQUIRED IMPORTED_TARGET dbus-1)
+    target_link_libraries(LibCore PRIVATE PkgConfig::dbus)
 endif()
--- a/Libraries/LibCore/NotificationUnix.cpp
+++ b/Libraries/LibCore/NotificationUnix.cpp
@@ -1,0 +1,105 @@
+/*
+ * Copyright (c) 2025, Niccolo Antonelli-Dziri <niccolo.antonelli-dziri@protonmail.com>
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include <AK/Error.h>
+#include <AK/Format.h>
+#include <AK/String.h>
+#include <LibCore/Notification.h>
+#include <LibCore/System.h>
+#include <dbus/dbus.h>
+
+namespace Core {
+
+class PlatformNotificationImpl final : public PlatformNotification {
+public:
+    static ErrorOr<NonnullOwnPtr<PlatformNotificationImpl>> create()
+    {
+        DBusError error;
+        dbus_error_init(&error);
+
+        DBusConnection* connection = dbus_bus_get(DBUS_BUS_SESSION, &error);
+
+        if (dbus_error_is_set(&error)) {
+            dbus_error_free(&error);
+            return Error::from_string_literal("Failed to connect to DBus session bus");
+        }
+
+        return adopt_own(*new PlatformNotificationImpl(connection));
+    }
+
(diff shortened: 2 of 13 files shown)


Libraries/LibWeb/NotificationsAPI/Notification.cpp
```diff
@@ -189,6 +191,8 @@ WebIDL::ExceptionOr<GC::Ref<Notification>> Notification::construct_impl(
     // then queue a task to fire an event named error on this, and abort these steps.
 
     // FIXME: 2. Run the notification show steps for notification.
```
konradekk: remember to remove the `FIXME`.

Libraries/LibCore/NotificationUnix.cpp
```diff
@@ -0,0 +1,105 @@
...
+#include <AK/Error.h>
+#include <AK/Format.h>
+#include <AK/String.h>
+#include <LibCore/Notification.h>
+#include <LibCore/System.h>
+#include <dbus/dbus.h>
+
+namespace Core {
+
+class PlatformNotificationImpl final : public PlatformNotification {
```
konradekk, line 16: `DBusPlatformNotification` or `FreeDesktopPlatformNotification` would probaly be a better name?

Libraries/LibWeb/NotificationsAPI/Notification.cpp
```diff
@@ -189,6 +191,8 @@ WebIDL::ExceptionOr<GC::Ref<Notification>> Notification::construct_impl(
...
+    auto& window = as<HTML::Window>(relevant_global_object);
```
konradekk: (not sure if this cast can fail, but `as_if` could be safer here.)

Libraries/LibCore/NotificationUnix.cpp
NiccoloAntonelliDziri: I followed the `TimeZoneWatcher` files for the name, so the same as:
`class TimeZoneWatcherImpl final : public TimeZoneWatcher`

But I can change it since it's not a class that is meant to be directly be called. 
Maybe PlatformNotificationLinux ? To be more consistent with the file name ?

Libraries/LibCore/CMakeLists.txt
```diff
@@ -81,6 +81,12 @@ else()
     list(APPEND SOURCES TimeZoneWatcherUnimplemented.cpp)
 endif()
 
+if(LINUX)
```
cqundefine:
```suggestion
if(LINUX OR BSD)
```

Libraries/LibCore/CMakeLists.txt
```diff
@@ -81,6 +81,12 @@ else()
...
+    list(APPEND SOURCES NotificationLinux.cpp)
```
cqundefine: The name `NotificationLinux` is kind of misleading, a name like `NotificationDBus` would make more sense as this implementation will work on any FreeDesktop system like the BSDs

Libraries/LibCore/CMakeLists.txt
```diff
@@ -121,3 +127,9 @@ endif()
 if (ANDROID)
     target_link_libraries(LibCore PRIVATE log)
 endif()
+
+if (LINUX)
```
cqundefine:
```suggestion
if (LINUX OR BSD)
```

Libraries/LibCore/CMakeLists.txt
NiccoloAntonelliDziri: Is `NotificationUnix` a better fit in your opinion ? In order to keep a generic platform name instead of the implementation details. It seems also more inline with other cross-platform implementations of other things inside of LibCore.

I changed the commit title too

cqundefine: Yeah that name should be fine, it's more about not naming it strictly Linux

Libraries/LibCore/CMakeLists.txt
cqundefine: You never changed this one :^)

NiccoloAntonelliDziri: Oh I'm sorry, I though I did. Fixed!

Libraries/LibCore/NotificationUnix.cpp
konradekk: no preference on my side – up to you and maintainers then! 😅

NiccoloAntonelliDziri:
Removing the stale label

gmta:

PR #8315, opened 2026-03-07
LibWeb: Fire error event at Worker when script loading fails

shannonbooth:
This fixes a whole bunch of WPT timeouts for Workers which wait for this event to arrive.

### DIFF
--- a/Libraries/LibWeb/HTML/WorkerAgentParent.cpp
+++ b/Libraries/LibWeb/HTML/WorkerAgentParent.cpp
@@ -6,8 +6,14 @@
 
 #include <LibWeb/Bindings/PrincipalHostDefined.h>
 #include <LibWeb/DOM/Document.h>
+#include <LibWeb/DOM/Event.h>
+#include <LibWeb/DOM/EventTarget.h>
+#include <LibWeb/HTML/EventLoop/EventLoop.h>
+#include <LibWeb/HTML/EventNames.h>
 #include <LibWeb/HTML/MessagePort.h>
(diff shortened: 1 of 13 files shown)


Libraries/LibWeb/HTML/WorkerAgentParent.cpp
```diff
@@ -69,6 +76,16 @@ void WorkerAgentParent::setup_worker_ipc_callbacks(JS::Realm& realm)
...
         return client.request_worker_agent(worker_type);
     };
+    m_worker_ipc->on_worker_script_load_failure = [self = GC::Weak { *this }]() {
+        if (!self)
+            return;
+        auto& outside_settings = *self->m_outside_settings;
+        auto& event_target = *self->m_worker_event_target;
+        queue_global_task(Task::Source::DOMManipulation, outside_settings.global_object(), GC::create_function(outside_settings.heap(), [&event_target, &outside_settings]() {
+            ErrorEventInit event_init {};
+            event_target.dispatch_event(ErrorEvent::create(outside_settings.realm(), EventNames::error, event_init));
```
tcl3: Should we be firing an `Event` here rather than an `ErrorEvent` based on the spec text: "Queue a global task on the DOM manipulation task source given worker's relevant global object to fire an event named error at worker."

The same spec text is already present in SharedWorker.cpp and fires an `Event`.

shannonbooth: Indeed, good catch!

tcl3 (approved):

PR #8318, opened 2026-03-08
RequestServer: Prioritize immediate fetch connection startup

ijxpwastaken:
Fetch startup can be delayed when libcurl waits for a potentially reusable
connection. This can increase latency on navigation and subresource loads.
Description

    Set CURLOPT_PIPEWAIT to 0 in the fetch path so transfers begin
    immediately instead of waiting for connection reuse.

    Enable CURLOPT_TCP_FASTOPEN and CURLOPT_TCP_NODELAY on fetch handles to
    reduce connection and request transmission latency where supported.

    Keep the change scoped to RequestServer fetch configuration.

### DIFF
--- a/Services/RequestServer/Request.cpp
+++ b/Services/RequestServer/Request.cpp
@@ -530,7 +530,11 @@ void Request::handle_fetch_state()
     set_option(CURLOPT_URL, m_url.to_byte_string().characters());
     set_option(CURLOPT_PORT, m_url.port_or_default());
     set_option(CURLOPT_CONNECTTIMEOUT, s_connect_timeout_seconds);
-    set_option(CURLOPT_PIPEWAIT, 1L);
+    // Do not delay a transfer while waiting for a potentially reusable connection.
+    // Prioritizing immediate connection establishment improves page load latency.
+    set_option(CURLOPT_PIPEWAIT, 0L);
+    set_option(CURLOPT_TCP_FASTOPEN, 1L);
+    set_option(CURLOPT_TCP_NODELAY, 1L);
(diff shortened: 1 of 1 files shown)


Services/RequestServer/Request.cpp
```diff
@@ -530,7 +530,11 @@ void Request::handle_fetch_state()
     set_option(CURLOPT_URL, m_url.to_byte_string().characters());
     set_option(CURLOPT_PORT, m_url.port_or_default());
     set_option(CURLOPT_CONNECTTIMEOUT, s_connect_timeout_seconds);
-    set_option(CURLOPT_PIPEWAIT, 1L);
+    // Do not delay a transfer while waiting for a potentially reusable connection.
+    // Prioritizing immediate connection establishment improves page load latency.
+    set_option(CURLOPT_PIPEWAIT, 0L);
```
jonbgamble, line 535: What effect does this flag have on multiplexing when there's a burst of resource fetches? Connection reuse seems kind of important in recent http versions.

Services/RequestServer/Request.cpp
```diff
@@ -530,7 +530,11 @@ void Request::handle_fetch_state()
...
+    set_option(CURLOPT_TCP_FASTOPEN, 1L);
+    set_option(CURLOPT_TCP_NODELAY, 1L);
```
jonbgamble, line 537: https://curl.se/libcurl/c/CURLOPT_TCP_FASTOPEN.html?utm_source=chatgpt.com

TLS session cache sounds like something we'd want. TLS endpoint termination is expensive.

PR #8319, opened 2026-03-08
Three flakiness fixes

awesomekling:
See individual commits.

The MIME sniff one is for a problem we've been seeing on WPT runners since MIME sniffing was made more lazy.

The screenshot one is for a crash we've been seeing on CI runners. It turns those crashes into timeouts instead, which is still an error but better than crashing.


awesomekling:
CI hit a null deref in IndexedDB so I tacked on a commit fixing that as well.

PR #8326, opened 2026-03-08
LibWeb: Drain microtasks after background parse completion

awesomekling:
When parse_off_thread() completes, the result callback runs inside a
deferred_invoke, which executes outside the HTML event loop's task
model. This meant that any microtasks queued by the callback (e.g.
promise reactions from react_to_promise during module linking) were
never drained, since HTML::EventLoop::process() only performs
microtask checkpoints after executing an HTML task.

Fix this by performing an explicit microtask checkpoint after the
parse result callback. This ensures that promise reactions queued
during module linking are drained immediately.

This fixes module worker scripts timing out because their loading
promise chains would stall indefinitely.

### DIFF
--- a/Libraries/LibWeb/HTML/Scripting/Fetching.cpp
+++ b/Libraries/LibWeb/HTML/Scripting/Fetching.cpp
@@ -25,6 +25,7 @@
 #include <LibWeb/Fetch/Infrastructure/HTTP/Requests.h>
 #include <LibWeb/Fetch/Infrastructure/HTTP/Responses.h>
 #include <LibWeb/Fetch/Infrastructure/URL.h>
+#include <LibWeb/HTML/EventLoop/EventLoop.h>
 #include <LibWeb/HTML/HTMLScriptElement.h>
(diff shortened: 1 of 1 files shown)


Libraries/LibWeb/HTML/Scripting/Fetching.cpp
```diff
@@ -64,15 +65,23 @@ static void parse_off_thread(NonnullRefPtr<JS::SourceCode const> source_code, JS
...
         auto* parsed = JS::RustIntegration::parse_program(utf16_data, length, type, line_number_offset);
 
         auto origin = event_loop_weak->take();
         if (!origin)
             return;
-        origin->deferred_invoke([parsed, callback]() {
-            (*callback)(parsed);
-            delete callback;
+        origin->deferred_invoke([parsed, callback, realm = move(realm)]() {
+            // NB: Queue this as an HTML task so that on_parsed runs inside
```
tcl3: Comment should be wrapped to 120 characters

(although others in this file aren't, so I'm not sure how much we care about this 🤷)

PR #8327, opened 2026-03-08
LibWeb: Ensure inline abspos containing blocks contain a fragment

### LINKED ISSUE #8302: `::before` with `position: absolute` incorrectly positioned at viewport origin when parent element has zero size
### Summary

When an element has no size (in my example, no content + inline positioning), it is positioned based on the viewport offset rather than its parent element.

### Operating system

macOS

### Steps to reproduce

Minimal snippet to reproduce the bug
```html
<style>
    span {
        position: relative;
    }
    span::before {
        content: "c";
        position: absolute;
        top: 0;
        left: 0;
    }
</style>
a<span></span>b a<span> </span>b
```
The first span has no content and is zero size; the second has a space, making it non-zero-sized.

### Expected behavior

The `::before` element should be positioned based on its parent element, even when the parent has zero size.

**Chrome** places it where the element is

### Actual behavior

The `::before` element is placed at the viewport origin when the parent has zero size.

**Ladybird** places it at the viewport origin

https://refined-github-html-preview.kidonng.workers.dev/twhlynch/ladybird/raw/refs/heads/zero-size-parent-pseudo-element-positioning/Tests/LibWeb/Screenshot/input/css-zero-size-parent-with-pseudo-element.html

```html
N/A
```

```shell
N/A
```

_No response_

_No response_

### Contribute a patch?

dgolebiewski:
Creates a fragment for inline nodes that establish an absolute positioning containing block (`position: relative`, etc).
This ensures we always have a fragment to reference when computing an abspos containing rect for an inline node.

Fixes #8302

Ladybird before:
ladybird-before
Ladybird after:
ladybird-after
Firefox for reference:
firefox

### DIFF
--- a/Libraries/LibWeb/Layout/InlineFormattingContext.cpp
+++ b/Libraries/LibWeb/Layout/InlineFormattingContext.cpp
@@ -10,6 +10,7 @@
 #include <LibWeb/Layout/BlockContainer.h>
 #include <LibWeb/Layout/BlockFormattingContext.h>
 #include <LibWeb/Layout/Box.h>
+#include <LibWeb/Layout/InlineNode.h>
 #include <LibWeb/Layout/InlineFormattingContext.h>
 #include <LibWeb/Layout/InlineLevelIterator.h>
 #include <LibWeb/Layout/LineBuilder.h>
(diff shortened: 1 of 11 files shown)


twhlynch:
Looks like the x position of the pseudoelement is using its parent's parent's position (if that makes sense).
Heres a working test file, by adding a margin to a container

```html
<!doctype html>
<style>
    span {
        position: relative;
    }
    span::before {
        content: "c";
        position: absolute;
        top: 0;
        left: 0;
    }
    div {
        margin: 10%;
    }
</style>
<div>a<span></span>b a<span> </span>b</div>
```

Below, the c should appear in the same position as the b, but instead is using the div's left margin.
Screenshot 2026-03-11 at 11 22 57 pm
Expected:
Screenshot 2026-03-11 at 11 24 11 pm

PR #8329, opened 2026-03-09
LibWeb: Implement get_total_length on SVGGeometryElement.

aurium:
Enable JS to get the real value for `getTotalLength()` at SVG DOM API.

### DIFF
--- a/Libraries/LibGfx/Path.h
+++ b/Libraries/LibGfx/Path.h
@@ -35,6 +35,7 @@ public:
     virtual void text(Utf16View const&, Font const&) = 0;
     virtual void glyph_run(GlyphRun const&) = 0;
     virtual void offset(Gfx::FloatPoint const&) = 0;
+    virtual float length() const = 0;
 
     virtual void append_path(Gfx::Path const&) = 0;
     virtual void intersect(Gfx::Path const&) = 0;
@@ -97,6 +98,7 @@ public:
     void text(Utf16View const& text, Font const& font) { impl().text(text, font); }
     void glyph_run(GlyphRun const& glyph_run) { impl().glyph_run(glyph_run); }
     void offset(Gfx::FloatPoint const& offset) { impl().offset(offset); }
+    float length() const { return impl().length(); }
 
     void horizontal_line_to(float x) { line_to({ x, last_point().y() }); }
     void vertical_line_to(float y) { line_to({ last_point().x(), y }); }
--- a/Libraries/LibWeb/SVG/SVGCircleElement.cpp
+++ b/Libraries/LibWeb/SVG/SVGCircleElement.cpp
@@ -4,12 +4,14 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
+#include <AK/Math.h>
 #include <LibGfx/Path.h>
 #include <LibWeb/Bindings/Intrinsics.h>
 #include <LibWeb/Bindings/SVGCircleElementPrototype.h>
 #include <LibWeb/CSS/CascadedProperties.h>
 #include <LibWeb/CSS/Parser/Parser.h>
 #include <LibWeb/CSS/PropertyID.h>
+#include <LibWeb/DOM/Document.h>
 #include <LibWeb/Layout/Node.h>
 #include <LibWeb/SVG/AttributeNames.h>
 #include <LibWeb/SVG/AttributeParser.h>
@@ -59,27 +61,36 @@ void SVGCircleElement::apply_presentational_hints(GC::Ref<CSS::CascadedPropertie
         cascaded_properties->set_property_from_presentational_hint(CSS::PropertyID::R, r_value.release_nonnull());
 }
 
-static CSSPixels normalized_diagonal_length(CSSPixelSize viewport_size)
+static float viewport_size_diagonal(CSSPixelSize viewport_size)
 {
-    if (viewport_size.width() == viewport_size.height())
-        return viewport_size.width();
-    return sqrt(((viewport_size.width() * viewport_size.width()) + (viewport_size.height() * viewport_size.height())) / 2);
+    float w = viewport_size.width().to_float();
+    float h = viewport_size.height().to_float();
+    if (w == h)
+        return w;
+    return AK::sqrt(((w * w) + (h * h)) / 2);
+}
+
+void SVGCircleElement::attribute_changed(FlyString const& name, Optional<String> const& old_value, Optional<String> const& value, Optional<FlyString> const& namespace_)
+{
+    Base::attribute_changed(name, old_value, value, namespace_);
+
+    if (name == SVG::AttributeNames::cx) {
+        m_center_x = AttributeParser::parse_number_percentage(value.value_or(String {}));
+    } else if (name == SVG::AttributeNames::cy) {
+        m_center_y = AttributeParser::parse_number_percentage(value.value_or(String {}));
+    } else if (name == SVG::AttributeNames::r) {
+        m_radius = AttributeParser::parse_number_percentage(value.value_or(String {}));
+    }
 }
 
 Gfx::Path SVGCircleElement::get_path(CSSPixelSize viewport_size)
 {
-    // NB: Called during SVG layout.
-    auto node = unsafe_layout_node();
-    if (!node) {
-        dbgln("FIXME: Null layout node in SVGCircleElement::get_path");
-        return {};
-    }
+    float viewport_w = viewport_size.width().to_float();
+    float viewport_h = viewport_size.height().to_float();
 
-    auto cx = float(node->computed_values().cx().to_px(*node, viewport_size.width()));
-    auto cy = float(node->computed_values().cy().to_px(*node, viewport_size.height()));
-    // Percentages refer to the normalized diagonal of the current SVG viewport
-    // (see Units: https://svgwg.org/svg2-draft/coords.html#Units)
-    auto r = float(node->computed_values().r().to_px(*node, normalized_diagonal_length(viewport_size)));
+    float cx = m_center_x.value_or(NumberPercentage::create_number(0)).resolve_relative_to(viewport_w);
+    float cy = m_center_y.value_or(NumberPercentage::create_number(0)).resolve_relative_to(viewport_h);
+    float r = m_radius.value_or(NumberPercentage::create_number(0)).resolve_relative_to(viewport_size_diagonal(viewport_size));
 
     // A zero radius disables rendering.
     if (r == 0)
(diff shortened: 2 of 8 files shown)


aurium:
Passes one new WPT test.
image

Libraries/LibWeb/SVG/SVGRectElement.cpp
```diff
@@ -112,6 +112,22 @@ Gfx::Path SVGRectElement::get_path(CSSPixelSize viewport_size)
...
+    float ry = m_radius_y.value_or(NumberPercentage::create_number(-1)).value();
+    if (rx == -1 && ry == -1)
+        rx = ry = 0;
+    if (rx == -1)
+        rx = ry;
+    if (ry == -1)
+        ry = rx;
+    float corners = 2 * M_PI * sqrt((pow(rx, 2) + pow(ry, 2)) / 2);
+    return w * 2 + h * 2 + corners - rx * 4 - ry * 4;
+}
```
gmta: I don't believe we should specialize `::get_total_length()` for subclasses - the base class already knows about it having a `Gfx::Path`, and that class is missing a `::length()` method. Our Skia backend offers `SkPathMeasure` with which you should be able to calculate the length for _any_ geometry element, not just `SVGRectElement`.

gmta (changes requested):
Welcome to the project! I believe we should head in another direction to implement this.

Libraries/LibWeb/SVG/SVGRectElement.cpp
aurium: Great! I don't know skia yet.

aurium:
image

Libraries/LibWeb/SVG/SVGGeometryElement.cpp
```diff
@@ -35,7 +37,12 @@ GC::Ptr<Layout::Node> SVGGeometryElement::create_layout_node(GC::Ref<CSS::Comput
 
 float SVGGeometryElement::get_total_length()
 {
-    return 0;
+    auto viewport_size = this->get_viewport_size();
+    auto path = this->get_path(viewport_size);
+    auto sk_path = to_skia_path(path);
+    SkPathMeasure path_measure(sk_path, false);
+    SkScalar path_length = path_measure.getLength();
```
gmta: You're using Skia-specific code inside SVGGeometryElement, which is not the right abstraction level.

We use a PIMPL for this: look at `Gfx::Path` and `::PathImplSkia` to see how that's implemented. That's where you need to introduce the `SkPathMeasure` usage.

Libraries/LibWeb/SVG/SVGCircleElement.cpp
```diff
@@ -30,6 +30,19 @@ void SVGCircleElement::initialize(JS::Realm& realm)
...
+    SVGElement::attribute_changed(name, old_value, value, namespace_);
+
+    if (name == SVG::AttributeNames::cx) {
+        m_center_x = AttributeParser::parse_coordinate(value.value_or(String {}));
+    } else if (name == SVG::AttributeNames::cy) {
+        m_center_y = AttributeParser::parse_coordinate(value.value_or(String {}));
+    } else if (name == SVG::AttributeNames::r) {
+        m_radius = AttributeParser::parse_positive_length(value.value_or(String {}));
+    }
+}
```
gmta: You're changing multiple things inside this single commit - please split them up. But why are you no longer looking at the computed values for these three properties?

Libraries/LibWeb/SVG/SVGEllipseElement.cpp
```diff
@@ -28,7 +28,7 @@ void SVGEllipseElement::initialize(JS::Realm& realm)
 
 void SVGEllipseElement::attribute_changed(FlyString const& name, Optional<String> const& old_value, Optional<String> const& value, Optional<FlyString> const& namespace_)
 {
-    Base::attribute_changed(name, old_value, value, namespace_);
+    SVGElement::attribute_changed(name, old_value, value, namespace_);
```
gmta: Why did you change this?

gmta (changes requested):

Libraries/LibWeb/SVG/SVGGeometryElement.cpp
aurium: Oh. I Just read about PIMPL. I believe I understand why this pattern is important in C++, especially in a project the size of Ladybird.

I'm sorry to grab you in this feedback loop. I'm a frontend developer, who had his C++ experience 20 year ago through college exercises. I'm ashamed by being like a dummy AI doing this little contribution. But I really want to learn and help Ladybird development.

Before pushing more code, I must to ask: I was starting a class SVGGeometryElementImplSkia, but I see, different from PathImplSkia, in this case I should not move all implementation to PIMPL, only the Skia related method (am I right?). Then I think that all I need is the path related implementation, so I could use PathImplSkia for this needs. That looks simpler and economical... is this right?

Libraries/LibWeb/SVG/SVGCircleElement.cpp
aurium: I saw these `m_*` attributes directly available in other classes and thought it was a good idea.

Sorry about putting all changes in the same commit. I don't like too, but I understood I must amend all review changes when I read `CONTRIBUTING.md`. My bad.

Libraries/LibWeb/SVG/SVGEllipseElement.cpp
aurium: I see `SVGEllipseElement` was missing important update triggers using only the `Base::attribute_changed`. The `SVGElement::attribute_changed` calls the Base implementation and more.

Well... That another problem and could be a bugfix on another PR, but I think "a PR of just one line?". And the method was related to my prior mistake, so that was not so unrelated.

Now that becomes really unrelated. Removing...

Libraries/LibWeb/SVG/SVGCircleElement.cpp
aurium: More important: Why I *need* to change this implementation:

The current implementation didn't find the node at `auto node = unsafe_layout_node()` and was returning an empty path. So the example that I screenshot and post above was giving `Circle: 0`.

Libraries/LibWeb/SVG/SVGGeometryElement.cpp
gmta: What I meant was this; we have a `PathImplSkia`:
```cpp
class PathImplSkia final : public PathImpl {
```
All the Skia-specific stuff should live in the implementation of this class. You have to add these methods:
```cpp
Gfx::PathImplSkia::length()
Gfx::PathImpl::length()
Gfx::Path::length()
```
If you want an example, look at how we implement `Gfx::Path::bounding_box()` and which methods it calls.

Once you've got that set up, _then_ you should look at `SVGGeometryElement` and call the right methods to get to the path's length.

Libraries/LibWeb/SVG/SVGCircleElement.cpp
gmta: Looks like we need the layout node to be in place at that point - so maybe we are missing a layout update somewhere in the code path that reaches `SVGCircleElement::get_path()`?

aurium: I think that ladybird misses a `set_layout_node` at some level, however I don't have more time last night. I saw that the methods `initialize`, `is_presentational_hint` and `apply_presentational_hints` are calling the `Base`'s "super" despite having these implementations in `SVGElement`.

I try to replace the three `Base::` calls by `SVGElement::` (I believe it must be) but was not enough to solve the "null layout". If you have any other leads, I would appreciate it.

gmta: It might be that you just need to call `update_layout(..)` before proceeding to interact with layout nodes when someone calls into `::length()`.

aurium: Calling `update_layout` didn't end well. That crashes the page.

I don't know why `SVGCircleElement::get_path` is called inside `root_formatting_context.run`, restarting `Document::update_layout` before that finishes. Please see this files:

My changes with debug logs: [SVGCircleElement-get_path-debug.diff](https://github.com/user-attachments/files/25997426/SVGCircleElement-get_path-debug.diff.txt)
The resulting STDOUT log: [SVGCircleElement-get_path-debug.log](https://github.com/user-attachments/files/25997427/SVGCircleElement-get_path-debug.log)
The svg for testing: [test.svg](https://github.com/user-attachments/assets/e941db40-5e26-46a9-bf43-3f25f7f1bc18)

Inside the svg you will see a `<script>` with `circle.getTotalLength()`. However, even when I delete the entire `<script>` the log shows `SVGCircleElement::get_path` being called inside `root_formatting_context.run`.

That can be a simple knowledge gap. But that looks weird to me.

aurium: I gave up using computed values and added `attribute_changed` at `SVGCircleElement` to set the `m_*` attributes. The same strategy is used by `SVGEllipseElement` and `SVGRectElement`, so I think that is ok.

Libraries/LibWeb/SVG/SVGGeometryElement.cpp
```diff
@@ -35,7 +36,9 @@ GC::Ptr<Layout::Node> SVGGeometryElement::create_layout_node(GC::Ref<CSS::Comput
 
 float SVGGeometryElement::get_total_length()
 {
-    return 0;
+    auto viewport_size = this->get_viewport_size();
+    auto path = this->get_path(viewport_size);
```
gmta: You don't need to use `this->` here.

Libraries/LibWeb/SVG/SVGCircleElement.cpp
gmta, line 1: Commit title: [don't use "weasel words"](https://github.com/LadybirdBrowser/ladybird/blob/master/CONTRIBUTING.md#code-submission-policy) like "fix", but describe what you're changing and why. For example:

```
LibWeb: Support cx/cy/r attributes for SVGCircleElement::get_path()
```

Libraries/LibGfx/Path.h
gmta, line 1: Commit message: it's not really that clear, maybe just say:

```
LibWeb: Implement SVGGeometryElement.getTotalLength()

Use Skia's SkPathMeasure to return the path length for arbitrary SVG
geometry elements.
```

aurium:
Thanks Jelle Raaijmakers!

Libraries/LibWeb/SVG/SVGGeometryElement.cpp
gmta: Please change this in the previous commit, where you introduced the change.

aurium: Done.

gmta:
@aurium tests are failing in CI - can you take a look?

PR #8330, opened 2026-03-09
LibWeb: Registered custom properties parsing improvements

Calme1709:
This PR contains a few different improvements to the parsing of `@property` rules and custom properties. See individual commits for details.

### DIFF
--- a/Libraries/LibWeb/CSS/Parser/ValueParsing.cpp
+++ b/Libraries/LibWeb/CSS/Parser/ValueParsing.cpp
@@ -126,50 +126,90 @@ Optional<Vector<ComponentValue>> Parser::parse_declaration_value(TokenStream<Com
     // contain <bad-string-token>, <bad-url-token>, unmatched <)-token>, <]-token>, or <}-token>, or top-level
     // <semicolon-token> tokens or <delim-token> tokens with a value of "!". It represents the entirety of what a valid
     // declaration can have as its value.
-    auto transaction = tokens.begin_transaction();
-    Vector<ComponentValue> declaration_value;
-    while (tokens.has_next_token()) {
-        auto const& peek = tokens.next_token();
-        if (!peek.is_token()) {
-            declaration_value.append(tokens.consume_a_token());
-            continue;
+    Vector<ComponentValue> top_level_declaration_value;
+
+    AK::Function<void(TokenStream<ComponentValue>&, Nested)> const parse_declaration_value_impl = [&](TokenStream<ComponentValue>& current_tokens, Nested nested) {
+        auto consume_a_token = [&]() {
+            if (nested == Nested::No)
+                top_level_declaration_value.append(current_tokens.consume_a_token());
+            else
+                current_tokens.discard_a_token();
+        };
+
+        auto transaction = current_tokens.begin_transaction();
+        while (current_tokens.has_next_token()) {
+            auto const& peek = current_tokens.next_token();
+
+            if (peek.is_block()) {
+                TokenStream block_stream { peek.block().value };
+                parse_declaration_value_impl(block_stream, Nested::Yes);
+                if (block_stream.is_empty()) {
+                    consume_a_token();
+                    continue;
+                }
+
+                break;
+            }
+
+            if (peek.is_function()) {
+                TokenStream function_stream { peek.function().value };
+                parse_declaration_value_impl(function_stream, Nested::Yes);
+                if (function_stream.is_empty()) {
+                    consume_a_token();
+                    continue;
+                }
+
+                break;
+            }
+
+            if (!peek.is_token()) {
+                consume_a_token();
+                continue;
+            }
+
+            bool valid = true;
+            switch (peek.token().type()) {
+            case Token::Type::Invalid:
+            case Token::Type::EndOfFile:
+            case Token::Type::BadString:
+            case Token::Type::BadUrl:
+                // NB: We're dealing with ComponentValues, so all valid function and block-related tokens will already be
+                //     converted to Function or SimpleBlock ComponentValues. Any remaining ones are invalid.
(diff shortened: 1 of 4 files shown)


Libraries/LibWeb/CSS/StyleValues/BackgroundSizeStyleValue.h
```diff
@@ -32,6 +32,8 @@ class BackgroundSizeStyleValue final : public StyleValueWithDefaultOperators<Bac
 
     bool properties_equal(BackgroundSizeStyleValue const& other) const { return m_properties == other.m_properties; }
 
+    bool is_computationally_independent() const override { return m_properties.size_x->is_computationally_independent() && m_properties.size_y->is_computationally_independent(); }
```
AtkinsSJ: Nit: `virtual`

Libraries/LibWeb/CSS/StyleValues/CalculatedStyleValue.cpp
```diff
@@ -847,6 +847,13 @@ bool NumericCalculationNode::equals(CalculationNode const& other) const
     return m_value == static_cast<NumericCalculationNode const&>(other).m_value;
 }
 
+bool NumericCalculationNode::is_computationally_independent() const
+{
+    return m_value.visit(
+        [](Length const& length) { return !length.is_font_relative(); },
```
AtkinsSJ: The spec is a bit ambiguous, but my understanding is that viewport units are not computationally independent, because they depend on the viewport size which can change. (And neither are `cq` units though we don't have those yet.) Their [example](https://drafts.css-houdini.org/css-properties-values-api-1/#example-65c0ed48) only uses the constant `in`->`px` conversion as an independent one.

Libraries/LibWeb/CSS/EdgeRect.h
```diff
@@ -22,6 +22,13 @@ struct WEB_API EdgeRect {
     LengthOrAuto left_edge;
     CSSPixelRect resolved(Layout::Node const&, CSSPixelRect) const;
     bool operator==(EdgeRect const&) const = default;
+    bool is_computationally_independent() const
+    {
+        return !top_edge.is_font_relative()
+            && !right_edge.is_font_relative()
+            && !bottom_edge.is_font_relative()
+            && !left_edge.is_font_relative();
```
AtkinsSJ: Same point about lengths. I think all relative lengths aren't. We might want a method on Length if it's more complicated than that though.

Libraries/LibWeb/CSS/Parser/ValueParsing.cpp
```diff
@@ -125,52 +125,80 @@ Optional<Vector<ComponentValue>> Parser::parse_declaration_value(TokenStream<Com
...
+                break;
+            declaration_value.append(current_tokens.consume_a_token());
         }
 
-        if (!valid)
-            break;
-        declaration_value.append(tokens.consume_a_token());
-    }
+        if (declaration_value.is_empty())
+            return OptionalNone {};
```
AtkinsSJ: We only use the contents `declaration_value` if `nested == Nested::No`, otherwise we only care if it's empty or not. That means in the nested case, we're copying CVs and allocating vector data unnecessarily. It'd make the code more complicated, but it might be worth not appending to it in the nested case and using a bool or something to track if it's valid.

Libraries/LibWeb/CSS/StyleValues/BackgroundSizeStyleValue.h
Calme1709: Fixed (and the same issue in a few other places too)

Libraries/LibWeb/CSS/StyleValues/CalculatedStyleValue.cpp
Calme1709: Good point on the `cq*` units - I have added a FIXME for that.

With regards to viewport units [the spec](https://drafts.css-houdini.org/css-properties-values-api/#computationally-independent) states
> A property value is computationally independent if it can be converted into a computed value using only the value of the property on the element, and "global" information that cannot be changed by CSS.

I believe that this includes viewport units since the size of the viewport can't be changed by CSS (at least within a document, CSS within ancestor documents can).

Both Chrome and Firefox consider viewport units to be computationally independent as well.

Libraries/LibWeb/CSS/Parser/ValueParsing.cpp
Calme1709: Done. Let me know how that looks now.

Calme1709 (commented):
Changes:
 - Applied feedback
 - Rebased to fix conflicts

Libraries/LibWeb/CSS/EdgeRect.h
Calme1709: Missed this comment whoops - I have added a method to `Length` since my understanding is that it is more complicated (also makes a good place for the FIXME around `cq*` units)

Calme1709:
Rebased and added computational independence checks for the new `ColorInterpolationMethod` style value

Calme1709:
Rebased to fix conflicts and account for sub-values of RatioStyleValue when determining computational independence

Libraries/LibWeb/CSS/StyleValues/CalculatedStyleValue.cpp
AtkinsSJ: :+1: As long as we're consistent with other browsers it's fine.

Libraries/LibWeb/CSS/Parser/ValueParsing.cpp
AtkinsSJ: Looks good to me.

PR #8332, opened 2026-03-09
LibWeb: Implement sibling invalidation sets

kalenikaliaksandr:
Replace flat InvalidationSet with recursive InvalidationPlan trees that preserve selector combinator structure. Previously, selectors with sibling combinators (+ and ~) fell back to whole-subtree invalidation. Now the StyleInvalidator walks the DOM following combinator-specific rules, so ".a + .b" only invalidates the adjacent sibling matching ".b" rather than the entire subtree.

Plans are compiled at stylesheet parse time by walking selector compounds right-to-left. For ".a .b + .c":
```
  [.c]: plan = { invalidate_self }
        register: "c" → plan

  [.b]: wrap("+", righthand)
        plan = { sibling_rules: [match ".c", adjacent, {self}] }
        register: "b" → plan

  [.a]: wrap(" ", righthand)
        plan = { descendant_rules: [match ".b", <sibling plan>] }
        register: "a" → plan
```

Changing class "a" produces a plan that walks descendants for ".b", checks ".b"'s adjacent sibling for ".c", and invalidates only that element.


AtkinsSJ:
`Ref/input/required-optional-pseudoclass.html` is failing on this, it looks like changing the `required` attribute isn't updating the `form:has(input:required)` selector.

PR #8339, opened 2026-03-09
LibWeb: Synchronously resize viewport on fullscreen transitions

tcl3:
Previously, `requestFullscreen()` and `exit_fullscreen()` relied entirely on async IPC to resize the viewport. The UI process would respond with an async viewport resize that could arrive before the next rendering update, causing run_the_resize_steps to fire a spurious resize event before fullscreenchange. We now update the viewport state in the WebContent process on fullscreen transitions, so that we can't cause resize events to be fired on the next rendering update.

This fixes a CI flake in,  `Text/input/wpt-import/fullscreen/api/element-request-fullscreen-timing.html`. Tested by running `Meta/ladybird.py run test-web --shuffle --repeat 100 -f Text/input/wpt-import/fullscreen/` locally.

### DIFF
--- a/Libraries/LibWeb/DOM/Document.cpp
+++ b/Libraries/LibWeb/DOM/Document.cpp
@@ -3711,6 +3711,14 @@ void Document::update_the_visibility_state(HTML::VisibilityState visibility_stat
     dispatch_event(event);
 }
 
+void Document::record_current_viewport_size()
+{
+    // Update last viewport state so that run_the_resize_steps() sees no change and does not fire a resize event.
+    m_last_viewport_size = viewport_rect().size().to_type<int>();
+    auto& viewport = *visual_viewport();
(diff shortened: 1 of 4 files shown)


Zaggy1024:
FYI: https://github.com/web-platform-tests/wpt/commit/65dc84ad93d44478340e4f8b08f806a3708ab7fa

It seems like if our headless mode is using a different size for the fullscreen viewport, then we should actually be consistently firing the resize event and asserting `['resize', 'fullscreenchange']`?

Zaggy1024:
I'm thinking that in light of the fact that the `resize` event is expected to precede the `fullscreenchange`, maybe we should spoof the viewport size to the future size (the screen size when entering fullscreen, the window size when exiting). This would have the added benefit of causing WebContent to resize the backing buffers and update layout for the new state in parallel with the platform initiating fullscreen, instead of doing so synchronously after the IPC round trip completes.

PR #8340, opened 2026-03-09
LibWebSocket+RequestServer: Some websocket fixes

shannonbooth:
See commits for details, the second one being nice as it fixes various timeouts on WPT for websockets such as:

https://wpt.live/websockets/Send-unpaired-surrogates.any.html?wss
https://wpt.live/websockets/Send-null.any.worker.html?wss
https://wpt.live/websockets/Send-binary-blob.any.worker.html?wss
https://wpt.live/websockets/Send-binary-arraybufferview-uint32-offset.any.worker.html?wss
https://wpt.live/websockets/Create-valid-url.any.html?wss

### DIFF
--- a/Libraries/LibWebSocket/WebSocket.cpp
+++ b/Libraries/LibWebSocket/WebSocket.cpp
@@ -6,6 +6,7 @@
  */
 
 #include <AK/Base64.h>
+#include <AK/Endian.h>
 #include <AK/Random.h>
 #include <LibCrypto/Hash/HashManager.h>
 #include <LibWebSocket/Impl/WebSocketImplSerenity.h>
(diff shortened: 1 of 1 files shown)


Libraries/LibWebSocket/WebSocket.h
```diff
@@ -129,6 +129,7 @@ class WebSocket final : public Core::EventReceiver {
     bool m_discard_connection_requested { false };
 
     u16 m_last_close_code { to_underlying(CloseStatusCode::NoStatusReceived) };
+    bool m_received_close_frame { false };
```
gmta: Wouldn't we be able to reuse `m_last_close_code` for this?

Services/RequestServer/WebSocketImplCurl.cpp
```diff
@@ -176,6 +176,12 @@ void WebSocketImplCurl::read_from_socket()
 
         if (result != CURLE_OK) {
             dbgln("Failed to read from WebSocket: {}", curl_easy_strerror(result));
+            // Process any successfully buffered data (e.g., the server's close frame)
+            // before reporting the error. This handles cases where the server drops the
+            // connection immediately after sending the close frame, causing a subsequent
+            // curl_easy_recv call to fail.
```
gmta: General remark: wrap code and comments at column 120.

Libraries/LibWebSocket/WebSocket.cpp
```diff
@@ -130,7 +130,11 @@ void WebSocket::close(u16 code, ByteString const& message)
         // Start the WebSocket closing handshake and set this’s ready state to CLOSING (2)."
         auto message_bytes = message.bytes();
         auto close_payload = ByteBuffer::create_uninitialized(message_bytes.size() + 2).release_value_but_fixme_should_propagate_errors(); // FIXME: Handle possible OOM situation.
-        close_payload.overwrite(0, (u8*)&code, 2);
+        // RFC 6455 section 5.5.1:
+        // > If there is a body, the first two bytes of the body MUST be a 2-byte unsigned integer (in network byte order)
+        // > representing a status code with value /code/ defined in Section 7.4.
+        close_payload[0] = (u8)(code >> 8);
+        close_payload[1] = (u8)(code & 0xFF);
```
gmta: This is what we have `AK::NetworkOrdered` for :)

Libraries/LibWebSocket/WebSocket.h
shannonbooth: We could! It was actually my first implementation of the fix, which is why I added that enum definition but I can switch it back over again to use that sentinal no status received code if you think it is better, I was on the fence

shannonbooth: done

PR #8355, opened 2026-03-10
LibWeb/Bindings: Implement callback interface object bindings

shannonbooth:
See commits for details

### DIFF
--- a/Libraries/LibWeb/CSS/MediaQueryList.idl
+++ b/Libraries/LibWeb/CSS/MediaQueryList.idl
@@ -1,5 +1,6 @@
 #import <DOM/EventTarget.idl>
 #import <DOM/EventHandler.idl>
+#import <DOM/EventListener.idl>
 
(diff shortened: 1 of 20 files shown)


Libraries/LibIDL/IDLParser.cpp
```diff
@@ -1289,7 +1289,13 @@ Interface& Parser::parse()
 
     parse_non_interface_entities(false, interface);
 
+    interface.referenced_interfaces.set(interface.name, &interface);
+
     for (auto& import : imports) {
+        interface.referenced_interfaces.set(import.name, &import);
+        for (auto& referenced_interface : import.referenced_interfaces)
+            interface.referenced_interfaces.set(referenced_interface.key, referenced_interface.value);
```
gmta:
```cpp
interface.referenced_interfaces.update(import.referenced_interfaces);
```

shannonbooth: One day I will remember about our nice helpers

PR #8356, opened 2026-03-10
LibJS+LibUnicode: Use icu4x for Temporal calendar operations

trflynn89:
Replace the icu4c-based calendar implementation with one built on the icu4x Rust crate (`icu_calendar`).

The icu4c API does not expose the píngqì month-assignment algorithm used by the Chinese and Dangi lunisolar calendars. Our old code had to approximate this by walking months via epoch millisecond arithmetic and manually tracking leap month positions, which produced incorrect month codes and ordinal month numbers for certain years. The icu4x calendar crate handles píngqì natively.

With this patch, which is almost a 1-to-1 mapping of ICU invocations, we pass 100% of all Temporal test262 tests.

```
test/built-ins/Temporal  4577/4577  (100.00%) [ ✅ 4577 ]
test/intl402/Temporal    2004/2004  (100.00%) [ ✅ 2004 ]
```

The end goal might be to use icu4x for all of our ICU needs. But it does not yet provide the APIs needed for all ECMA-402 prototypes.

### DIFF
--- a/Libraries/LibUnicode/Rust/src/calendar.rs
+++ b/Libraries/LibUnicode/Rust/src/calendar.rs
@@ -1,0 +1,604 @@
+/*
+ * Copyright (c) 2026-present, the Ladybird developers.
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+use calendrical_calculations::rata_die::RataDie;
+use icu_calendar::{AnyCalendar, AnyCalendarKind, Date, types::MonthCode};
+use std::panic::{AssertUnwindSafe, catch_unwind};
+
+#[repr(C)]
+pub struct FfiISODate {
+    pub year: i32,
+    pub month: u8,
+    pub day: u8,
+}
+
+#[repr(C)]
+pub struct FfiOptionalISODate {
+    pub iso_date: FfiISODate,
+    pub has_value: bool,
+}
+
+#[repr(C)]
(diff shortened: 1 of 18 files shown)


Libraries/LibUnicode/Rust/src/calendar.rs
```diff
@@ -0,0 +1,607 @@
...
+};
+
+fn abort_on_panic<F: FnOnce() -> R, R>(f: F) -> R {
+    match catch_unwind(AssertUnwindSafe(f)) {
+        Ok(result) => result,
+        Err(_) => std::process::abort(),
+    }
+}
+
+// SAFETY: The C++ caller passes data from AK::String, which is always valid UTF-8.
```
shannonbooth: Hmm, actually what about WTF-8 case here?

trflynn89: The strings passed to Rust (calendar names and month codes) are known to be valid UTF-8. ASCII at that actually.

shannonbooth: Makes sense, though maybe we should rephrase the comment/function? string_from_ffi is quite generic sounding, but cant handle any AK::String being passed through which might be inferred from the comment (I went to double check rust str after seeing it)

trflynn89: Sure, done

PR #8360, opened 2026-03-11
LibJS: Shrink ExecutionContext to make function calls faster

awesomekling:
Every JS function call has to create an ExecutionContext, so the smaller these are, the less work we need to do.

See individual commits for details. :)

### DIFF
--- a/Libraries/LibWeb/Bindings/MainThreadVM.cpp
+++ b/Libraries/LibWeb/Bindings/MainThreadVM.cpp
@@ -142,15 +142,15 @@ void initialize_main_thread_vm(AgentType type)
         // 1. Let script be the running script.
         //    The running script is the script in the [[HostDefined]] field in the ScriptOrModule component of the running JavaScript execution context.
         HTML::Script* script { nullptr };
-        vm.running_execution_context().script_or_module.visit(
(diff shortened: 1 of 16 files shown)


Libraries/LibWeb/Bindings/MainThreadVM.cpp
```diff
@@ -142,15 +142,7 @@ void initialize_main_thread_vm(AgentType type)
...
-        vm.running_execution_context().script_or_module.visit(
-            [&script](GC::Ref<JS::Script>& js_script) {
-                script = as<HTML::ClassicScript>(js_script->host_defined());
-            },
-            [&script](GC::Ref<JS::Module>& js_module) {
-                script = as<HTML::ModuleScript>(js_module->host_defined());
-            },
-            [](Empty) {
-            });
+        JS::script_or_module_from_cell(vm.running_execution_context().script_or_module).visit([&script](GC::Ref<JS::Script>& js_script) { script = as<HTML::ClassicScript>(js_script->host_defined()); }, [&script](GC::Ref<JS::Module>& js_module) { script = as<HTML::ModuleScript>(js_module->host_defined()); }, [](Empty) {});
```
shannonbooth: nit but could this please be broken into lines again? It's a bit more difficult to read all squished up like this

PR #8369, opened 2026-03-11
LibWeb: Enable crash backtraces from WPT runner

sideshowbarker:
When a WebContent process receives a fatal signal, print a backtrace to stderr before re-raising the signal. The backtrace is captured by the test runner and written to a .stderr.html file.

Uses SA_RESETHAND so the handler runs only once, then resets to the default disposition and re-raises the signal for normal crash behavior.

### DIFF
--- a/Services/WebContent/main.cpp
+++ b/Services/WebContent/main.cpp
@@ -52,6 +52,48 @@
 
 #include <SDL3/SDL_init.h>
 
+#if !defined(AK_OS_WINDOWS)
+#    include <signal.h>
+static void crash_signal_handler(int signo)
+{
+    char const* name = "unknown";
+    switch (signo) {
+    case SIGSEGV:
+        name = "SIGSEGV";
+        break;
+    case SIGBUS:
+        name = "SIGBUS";
+        break;
+    case SIGFPE:
+        name = "SIGFPE";
+        break;
+    case SIGABRT:
+        name = "SIGABRT";
+        break;
+    case SIGILL:
+        name = "SIGILL";
+        break;
+    }
+    warnln("\n\033[31;1mCRASH\033[0m: Received signal {} ({})", name, signo);
+    dump_backtrace(2, 100);
+    signal(signo, SIG_DFL);
+    raise(signo);
+}
+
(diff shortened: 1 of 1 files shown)


xnacly:
[`man signal.2`](https://man7.org/linux/man-pages/man2/signal.2.html) says:

```
According to POSIX, the behavior of a process is 
undefined after it ignores a SIGFPE, SIGILL, or SIGSEGV signal 
that was not generated by kill(2) or raise(3).
```

So this may not work for all cases

sideshowbarker:
> [`man signal.2`](https://man7.org/linux/man-pages/man2/signal.2.html) says:
> 
> ```
> According to POSIX, the behavior of a process is 
> undefined after it ignores a SIGFPE, SIGILL, or SIGSEGV signal 
> that was not generated by kill(2) or raise(3).
> ```
> 
> So this may not work for all cases

Yeah — also `warnln()` and `dump_backtrace()` allocate, which is UB in a signal handler. But I can say that on macOS at least, it does work as this code expects.

sideshowbarker:
Here’s how I tested:

Run some test that takes long enough to keep a `WebContent` process alive:

```
Build/debug/bin/test-web --test-path Tests/LibWeb \
  -f "Text/input/wpt-import/dom/nodes/ParentNode-querySelector-All.html" \
  -j 1 -t 30
```
In a separate terminal window/tab, find and crash the `WebContent` process:

```
# match any Ladybird WebContent process; don’t match Safari’s com.apple.WebKit.WebContent
kill -SEGV $(pgrep -f "WebContent --command-line" | head -1)
```
…or on macOS, you can also do it this way:

```
# match any Ladybird WebContent process; don’t match Safari’s com.apple.WebKit.WebContent
kill -SEGV $(pgrep -f "Ladybird.app/Contents/MacOS/WebContent" | head -1)
```

The test runner should report `Crashed: 1.`

Open `test-dumps/results/index.html` and navigate to the crashed test and click the **`stderr`** tab.

You should see a full backtrace starting with:

```
CRASH: Received signal SIGSEGV (11)
#0  ...
```

PR #8371, opened 2026-03-11
LibWeb: Detect promise-returning IDL callbacks correctly

gmta:
The code generator checked is_object() to detect Promise return types, but Promise<T> is a ParameterizedType, so the check always failed and caused crashes in WebContent in throwing callback functions.

### DIFF
--- a/Libraries/LibWeb/ViewTransition/ViewTransition.cpp
+++ b/Libraries/LibWeb/ViewTransition/ViewTransition.cpp
@@ -631,18 +631,7 @@ void ViewTransition::call_the_update_callback()
 
     // 5. Otherwise, set callbackPromise to the result of invoking transition’s update callback.
     else {
-        auto promise = MUST(WebIDL::invoke_callback(*m_update_callback, {}, {}));
-        // FIXME: since WebIDL::invoke_callback does not yet convert the value for us,
-        // We need to do it here manually.
-        // https://webidl.spec.whatwg.org/#js-promise
-
-        HTML::TemporaryExecutionContext context(realm, HTML::TemporaryExecutionContext::CallbacksEnabled::Yes);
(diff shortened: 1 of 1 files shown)


shannonbooth:
Huh, I could have sworn I had fixed this beforehand, but turns out that was in a merge request that went stale and was never merged

edit: https://github.com/LadybirdBrowser/ladybird/pull/4561

Meta/Lagom/Tools/CodeGenerators/LibWeb/BindingsGenerator/IDLGenerators.cpp
```diff
@@ -1104,7 +1104,7 @@ static void generate_to_cpp(SourceGenerator& generator, ParameterType& parameter
         auto callback_function_generator = scoped_generator.fork();
         auto& callback_function = interface.callback_functions.find(parameter.type->name())->value;
 
-        if (callback_function.return_type->is_object() && callback_function.return_type->name() == "Promise")
+        if (callback_function.return_type->is_parameterized() && callback_function.return_type->name() == "Promise")
```
shannonbooth, line 1107: A simpler fix might just be to remove this check - I don't think there is a case that will have a return type of Promise and not be anything other than is_parameterized

gmta: It's a bit overzealous, I'll give you that. But I didn't want to think too much about if there could be other types named "Promise" that would cause the condition to trigger 😅 OK to leave it as-is?

gmta:
@shannonbooth 

> edit: #4561

Seems like that could use a revival!

shannonbooth:
Honestly at this point I should probably consider starting a rewrite of the IDL generator in python or something before making those changes. Life would be a lot simpler if the IDL generator was run on all files at once...

PR #8373, opened 2026-03-11
LibWeb: Include shadow host when checking for node event listeners

Zaggy1024:
Fixes a regression in 57c9bb5 where the media controls would not show when moving the cursor over the video overlay element.


Zaggy1024:
Done!

gmta (approved):

PR #8375, opened 2026-03-11
CMake: Patch highway to work with clang-22 and possibly gcc-16

ADKaster:
According to the relevant patches since highway 1.3.0, this should also
fix the build with gcc-16.

The attached diff is the difference between highway at
aa0f7f7a2593c770d54982fca63ac53ea5ebe632 and the 1.3.0 release.

Tested with clang-22 on Ubuntu 24.04.

cf.
https://github.com/microsoft/vcpkg/issues/50397
https://github.com/google/highway/issues/2697
https://github.com/google/highway/issues/2577
https://github.com/llvm/llvm-project/pull/157034

Presumably vcpkg maintainers or highway maintainers will have a better fix than this (such as a new patch release), but this works in the meantime.


ADKaster:
Huh. Fix one config, break everything else.

joekohlsdorf:
Works on Arch with clang 22.1.2.
Thank you for this patch, I had been running into this issue for a couple of weeks already.

ADKaster:
Highway released a new version, 1.4.0 this PR is no longer needed

PR #8382, opened 2026-03-12
LibWeb: Disable ancestor filter when directly computing style

Calme1709:
When we compute style directly (i.e. not via `update_style_recursively`) we don't have a populated ancestor filter and were therefore not matching valid rules.

This commit disables the ancestor filter optimization when computing directly to avoid those false negatives.

Another (possibly better) approach would be to manually populate the ancestor filter.


Libraries/LibWeb/CSS/StyleComputer.h
```diff
@@ -211,11 +215,15 @@ class WEB_API StyleComputer final : public GC::Cell {
     CSSPixelRect m_viewport_rect;
 
     OwnPtr<CountingBloomFilter<u8, 14>> m_ancestor_filter;
+    mutable bool m_disable_ancestor_filter_counter { false };
```
tcl3, line 218: The naming here is a bit misleading - this isn't a counter.

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1719,10 +1719,18 @@ GC::Ref<ComputedProperties> StyleComputer::create_document_style() const
     return style;
 }
 
-GC::Ref<ComputedProperties> StyleComputer::compute_style(DOM::AbstractElement abstract_element, Optional<bool&> did_change_custom_properties) const
+GC::Ref<ComputedProperties> StyleComputer::compute_style(DOM::AbstractElement abstract_element, Optional<bool&> did_change_custom_properties, DisableAncestorFilter disable_ancestor_filter) const
 {
-    auto& style_scope = abstract_element.style_scope();
-    return *compute_style_impl(abstract_element, ComputeStyleMode::Normal, did_change_custom_properties, style_scope);
+    if (disable_ancestor_filter == DisableAncestorFilter::Yes)
+        m_disable_ancestor_filter_counter = true;
```
tcl3, line 1725: Could we use a `TemporaryChange` (or `ScopeGuard`) here to avoid needing to update this in 2 places?

Calme1709:
Closing this off since #8484 solves this problem in what I think is a better way

PR #8391, opened 2026-03-12
LibWeb: Keep ShadowRealm module map for dynamic imports within evaluate

shannonbooth:
When HostLoadImportedModule is called from within ShadowRealm.evaluate() the referencing script may belong to the outer window's realm. Unconditionally setting the realm to the realm of the referencing script sets moduleMapRealm to the referencing script's realm causes modules to be associated with the wrong module map, breaking dynamic import inside ShadowRealms.

This fixes a whole bunch of timeouts for ShadowRealm tests in WPT and as far as I can tell, fixes our HTML integration of ShadowRealms edit: never mind, I now remember why this fix was not correct last time!


shannonbooth:
For example I think this fixes all 70 timeouts in the streams/ folder

edit: 69, there is also a ServiceWorker test

PR #8393, opened 2026-03-13
LibWeb: Use repeating timers for setInterval() to reduce drift

awesomekling:
The timer nesting level throttling change (7577fd2a57b) inadvertently reverted the repeating timer optimization from 4d27e9aa5e4, causing setInterval() to use single-shot timers that re-arm after the callback completes. This meant the next firing was scheduled relative to callback completion rather than the previous fire time, causing drift proportional to callback execution time.

For example, DiabloWeb's 50ms game loop with ~20ms of WASM work per frame was only achieving 14 FPS (71ms intervals) instead of 20 FPS.

Fix this by using repeating Core::Timer for setInterval() again. The repeating timer fires on schedule regardless of callback duration. On re-arm, we update the callback (for nesting level changes) and the interval (in case nesting level clamping kicks in), but don't restart the timer since it's already running on the correct schedule.


tcl3 (approved):
Not sure why I thought it was necessary to use single shot timers here 😅

PR #8397, opened 2026-03-13
LibWeb: Rewrite unload/destroy document lifecycle to follow the spec

gmta:
An attempt to fix some of the causes of flaky WPT and local tests:

* Support reentrancy for `EventLoop::spin_processing_tasks_with_source_until()`;

* Fix a cause of infinite `spin_until()` as a result of `::apply_the_history_step()` by using `spin_processing_tasks_with_source_until()` so we can keep processing related tasks;

* `unload_a_document_and_its_descendants()` now follows the spec algorithm structure: recursively unload child navigables via queued tasks (step 4), wait for them (step 5), then queue this document's own unload as a separate task (step 6). Previously, this was flattened into a single spin that unloaded all descendants and the document together, followed by a non-spec call to destroy_a_document_and_its_descendants();

* We also remove `spin_until()` from `unload_a_document_and_its_descendants()` reusing the mechanism that @kalenikaliaksandr made for `destroy_a_document_and_its_descendants()`.

* `unload()`'s step 19 now calls destroy() when the document is not salvageable, as the spec requires. Previously this was a no-op with a comment deferring to unload_a_document_and_its_descendants().

* `destroy_top_level_traversable()` step 2 now calls `destroy_a_document_and_its_descendants()` instead of `destroy()`.

The `iframe-unloading-order test`, which exercises named iframe access during unload handlers (the scenario the previous logic was designed to protect), still passes.

### DIFF
--- a/Tests/LibWeb/Text/expected/wpt-import/webidl/ecmascript-binding/window-named-properties-object.txt
+++ b/Tests/LibWeb/Text/expected/wpt-import/webidl/ecmascript-binding/window-named-properties-object.txt
@@ -1,0 +1,19 @@
+Harness status: OK
+
+Found 13 tests
+
+12 Pass
+1 Fail
+Pass	[[SetPrototypeOf]] and [[GetPrototypeOf]]
+Pass	[[PreventExtensions]] and [[IsExtensible]]
+Fail	[[GetOwnProperty]]
+Pass	[[GetOwnProperty]] (named property visibility algorithm)
+Pass	[[DefineOwnProperty]]
+Pass	[[HasProperty]]
+Pass	[[Get]]
+Pass	[[Get]] (named property visibility algorithm)
+Pass	[[Set]] (direct)
+Pass	[[Set]] (prototype chain)
+Pass	[[Set]] (Reflect.set)
+Pass	[[Delete]]
+Pass	[[OwnPropertyKeys]]
--- a/Tests/LibWeb/Text/input/wpt-import/webidl/ecmascript-binding/window-named-properties-object.html
+++ b/Tests/LibWeb/Text/input/wpt-import/webidl/ecmascript-binding/window-named-properties-object.html
@@ -1,0 +1,284 @@
+<!doctype html>
+<meta charset="utf-8">
+<title>Internal methods of Window's named properties object</title>
+<link rel="help" href="https://webidl.spec.whatwg.org/#named-properties-object">
+<script src="../../resources/testharness.js"></script>
+<script src="../../resources/testharnessreport.js"></script>
+<body>
+<script>
+function sloppyModeSet(base, key, value) { base[key] = value; }
+</script>
+<script>
+"use strict";
+
+const supportedNonIndex = "supported non-index property name";
+const supportedIndex = "supported indexed property name";
+const unsupportedNonIndex = "unsupported non-index property name";
+const unsupportedIndex = "unsupported indexed property name";
+const existingSymbol = "existing symbol property name";
+const nonExistingSymbol = "non-existing symbol property name";
+
+test(t => {
+    const { w, wp } = createWindowProperties(t);
+
+    Object.setPrototypeOf(wp, w.EventTarget.prototype); // Setting current [[Prototype]] value shouldn't throw
+
+    assert_throws_js(TypeError, () => { Object.setPrototypeOf(wp, {}); });
+    assert_throws_js(w.TypeError, () => { wp.__proto__ = null; });
+    assert_false(Reflect.setPrototypeOf(wp, w.Object.prototype));
+
+    assert_equals(Object.getPrototypeOf(wp), w.EventTarget.prototype);
+}, "[[SetPrototypeOf]] and [[GetPrototypeOf]]");
+
+test(t => {
+    const { wp } = createWindowProperties(t);
+
+    assert_throws_js(TypeError, () => { Object.preventExtensions(wp); });
+    assert_false(Reflect.preventExtensions(wp));
+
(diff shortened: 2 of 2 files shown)


Libraries/LibWeb/HTML/TraversableNavigable.cpp
```diff
@@ -1282,7 +1282,7 @@ void TraversableNavigable::destroy_top_level_traversable()
     // 1. Let browsingContext be traversable's active browsing context.
     auto browsing_context = active_browsing_context();
 
-    // 2. For each historyEntry in traversable's session history entries:
+    // 2. For each historyEntry in traversable's session history entries [[ in what order? ]]:
```
trflynn89, line 1285: I wonder if we should standardize on this style for these spec questions? I like the double braces

Libraries/LibWeb/HTML/TraversableNavigable.cpp
```diff
@@ -1287,9 +1287,9 @@ void TraversableNavigable::destroy_top_level_traversable()
         // 1. Let document be historyEntry's document.
         auto document = history_entry->document();
 
-        // 2. If document is not null, then destroy document.
+        // 2. If document is not null, then destroy a document and its descendants given document.
         if (document)
-            document->destroy();
+            document->destroy_a_document_and_its_descendants();
```
trflynn89, line 1292: Aha! I think this spec change is what I missed when I took a stab at this in #7824

Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -4855,42 +4855,47 @@ void Document::unload_a_document_and_its_descendants(GC::Ptr<Document> new_docum
...
                 // 1. Let incrementUnloaded be an algorithm step which increments numberUnloaded.
-                auto increment_unloaded = GC::create_function(heap, [&number_unloaded] { ++number_unloaded; });
+                auto increment_unloaded = GC::create_function(heap, [unload_state] { unload_state->did_process_child(); });
 
                 // 2. Unload a document and its descendants given childNavigable's active document, null, and incrementUnloaded.
                 child_navigable->active_document()->unload_a_document_and_its_descendants({}, increment_unloaded);
             }));
     }
 
+    // NOTE: These steps are handled by DocumentLifecycleState.
```
trflynn89: `NB`

Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -4855,42 +4855,47 @@ void Document::unload_a_document_and_its_descendants(GC::Ptr<Document> new_docum
...
-            // FIXME: 1. If firePageSwapSteps is given, then run firePageSwapSteps.
-
-            // 2. Unload document, passing along newDocument if it is not null.
-            unload(new_document);
-
-            // 3. If afterAllUnloads was given, then run it.
-            if (after_all_unloads)
-                after_all_unloads->function()();
-        }));
+    //    1. If firePageSwapSteps is given, then run firePageSwapSteps.
```
trflynn89: So usually what I do in situations where I have to implement spec steps out-of-order is just keep the spec text with the code, instead of writing them down here. Something like:

```c++
// 6. Queue a global task on the navigation and traversal task source given document's relevant global object to
//    perform the following steps:
auto finish_callback = GC::create_function(heap(), [document = this, new_document, after_all_unloads] {
    // FIXME: 1. If firePageSwapSteps is given, then run firePageSwapSteps.

    // 2. Unload document, passing along newDocument if it is not null.
    document->unload(new_document);

    // 3. If afterAllUnloads was given, then run it.
    if (after_all_unloads)
        after_all_unloads->function()();
});

// AD-HOC: We avoid allocating a DocumentLifecycleState in case there's no child navigables.
if (child_navigables.is_empty()) {
    HTML::queue_global_task(HTML::Task::Source::NavigationAndTraversal, relevant_global_object(*this), finish_callback);
    return;
}
```

shannonbooth:
This appears to fix the crash in https://wpt.live/webidl/ecmascript-binding/window-named-properties-object.html !

Libraries/LibWeb/HTML/TraversableNavigable.cpp
gmta: Yeah this was a subtle change!

Libraries/LibWeb/HTML/TraversableNavigable.cpp
gmta: I've included a change to CodingStyle.md in an attempt to standardize to this.

Libraries/LibWeb/DOM/Document.cpp
gmta: Updated.

Libraries/LibWeb/DOM/Document.cpp
gmta: I generally dislike out-of-order spec steps since they make it pretty hard to compare our implementation with the spec, but in this case it's a short enough algorithm to do so. Updated!

shannonbooth:
> This appears to fix the crash in [wpt.live/webidl/ecmascript-binding/window-named-properties-object.html](https://wpt.live/webidl/ecmascript-binding/window-named-properties-object.html) !

Could we perhaps import this test? An import of that test reliably crashes before these changes but no longer crashes afterwards

gmta:
> > This appears to fix the crash in [wpt.live/webidl/ecmascript-binding/window-named-properties-object.html](https://wpt.live/webidl/ecmascript-binding/window-named-properties-object.html) !
> 
> Could we perhaps import this test? An import of that test reliably crashes before these changes but no longer crashes afterwards

Absolutely, done!

PR #8399, opened 2026-03-13
LibWeb: Bump Chrome version in SAD_COMPATIBILITY_HACK

awesomekling:
Some web servers (like azure.com) are throttling us unless we claim to be a more recent version of Chrome.

Before this change, every HTTP request would take up to 30 seconds before receiving a response. After the change, azure.com responds instantly.

This is very silly, but it's the game we have to play.

### DIFF
--- a/Libraries/LibWeb/Loader/UserAgent.h
+++ b/Libraries/LibWeb/Loader/UserAgent.h
@@ -65,7 +65,7 @@ namespace Web {
 // NB: Some web servers treat us very badly unless we pretend to be one of the major browsers.
 //     This token is appended to the User-Agent string to improve compatibility.
 //     We will need to update this periodically to match a somewhat recent version.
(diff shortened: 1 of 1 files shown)


Libraries/LibWeb/Loader/UserAgent.h
```diff
@@ -65,7 +65,7 @@ namespace Web {
 // NB: Some web servers treat us very badly unless we pretend to be one of the major browsers.
 //     This token is appended to the User-Agent string to improve compatibility.
 //     We will need to update this periodically to match a somewhat recent version.
-#define SAD_COMPATIBILITY_HACK "Chrome/140.0.0.0 AppleWebKit/537.36 Safari/537.36"
+#define SAD_COMPATIBILITY_HACK "Chrome/145.0.0.0 AppleWebKit/537.36 Safari/537.36"
```
trflynn89: The latest chrome stable is 146, we might as well use that? Might kick the next needed update down the road a few more weeks

awesomekling: Sure!

PR #8408, opened 2026-03-14
LibJS/Rust: Implement lazy parsing for nested function bodies

jdahlin:
Most functions in a typical JS bundle are never called during initial page load.
Lazy parsing defers nested function bodies — validating syntax but skipping AST
construction — and fully parses them only when first compiled. See
https://v8.dev/blog/preparser for background on the technique.

On expressen.se, this reduces WebContent RSS by ~400 MB (1475 MB → 1070 MB).

The parser is made generic over `const SYNTAX_ONLY: bool`. When true, lightweight
sentinels replace AST nodes while a `FreeVarOnly` scope collector tracks captured
variables. Doubly-nested bodies use brace-counting. On-demand materialization
re-parses at the saved source offset during bytecode compilation.

### DIFF
--- a/Tests/LibJS/Runtime/syntax/early-errors-in-nested-functions.js
+++ b/Tests/LibJS/Runtime/syntax/early-errors-in-nested-functions.js
@@ -1,0 +1,104 @@
+// Syntax errors inside function bodies must be reported as early errors
+// (at parse time), even if the function is never called.
+// https://tc39.es/ecma262/#sec-static-semantics-early-errors
+//
+// NOTE: We use eval() here because test-js evaluates each test file as a
+// script. eval() also parses as a script and must report early errors.
+// If lazy parsing skips syntax checking, these tests will fail.
+
+describe("syntax errors in inner functions are early errors", () => {
+    test("syntax error in nested function declaration", () => {
+        expect("function outer() { function inner() { var x = ; } }").not.toEval();
+    });
+
+    test("syntax error in nested function expression", () => {
+        expect("function outer() { var f = function() { return = ; }; }").not.toEval();
+    });
+
+    test("syntax error in arrow function body", () => {
+        expect("function outer() { var f = () => { var x = ; }; }").not.toEval();
+    });
+
+    test("syntax error in method definition", () => {
+        expect("function outer() { var o = { m() { var x = ; } }; }").not.toEval();
+    });
+
+    test("syntax error in getter", () => {
+        expect("function outer() { var o = { get x() { var y = ; } }; }").not.toEval();
+    });
+
+    test("syntax error in deeply nested function", () => {
+        expect("function a() { function b() { function c() { var x = ; } } }").not.toEval();
+    });
+
+    test("syntax error in function inside class method", () => {
+        expect("class C { m() { function f() { var x = ; } } }").not.toEval();
+    });
+});
+
+describe("error messages report correct line numbers in nested functions", () => {
+    test("line number in nested function", () => {
+        try {
+            eval("function outer() {\n  function inner() {\n    var x = ;\n  }\n}");
+            expect().fail();
+        } catch (e) {
+            expect(e.message.includes("line: 3")).toBeTrue();
+        }
+    });
+
+    // NOTE: Syntax errors in doubly-nested function bodies (depth >= 2
+    // inside the SYNTAX_ONLY parser) are currently not caught at parse
+    // time because the brace-counting skip doesn't validate syntax.
+    // They are caught when the function is first compiled. This matches
+    // V8/JSC behavior where deeply nested bodies are fully deferred.
+    // The single-nesting case (depth 1) IS caught because the SYNTAX_ONLY
+    // checker fully parses at that level.
+});
+
+describe("lazy parsing edge cases", () => {
+    test("template literal with function inside nested function", () => {
+        expect("function outer() { var f = function() { return `${function() { return 1; }()}`; }; }").toEval();
+    });
+
+    test("private field access from inner function", () => {
+        expect("class C { #f = 1; method() { function inner() { return this.#f; } } }").toEval();
+    });
+
+    test("eval inside nested function prevents optimization", () => {
+        function outer() {
+            var x = 42;
+            function inner() {
+                return eval("x");
+            }
+            return inner();
+        }
+        expect(outer()).toBe(42);
+    });
+
+    test("default parameter with body var shadowing", () => {
+        function outer() {
+            function f(a = typeof b) {
+                var b = 1;
+                return a;
+            }
+            return f();
+        }
+        expect(outer()).toBe("undefined");
+    });
+
+    test("use strict in lazy-parsed inner function rejects with", () => {
+        expect('function outer(){ function inner(){ "use strict"; with({}){} } }').not.toEval();
+    });
+
+    test("arrow function captures this from enclosing scope", () => {
+        function outer() {
+            this.value = 99;
+            var inner = function () {
+                var arrow = () => this.value;
+                return arrow.call(this);
+            };
+            return inner.call(this);
+        }
+        expect(outer.call({ value: 99 })).toBe(99);
+    });
+});


jdahlin:
Note that v8/d8 takes 0.12s and uses 71Mb peak memory footprint, so theres a bit left.

awesomekling:
`test-js-bytecode` is failing.. my guess would be because lazy parsing doesn't preserve metadata about locals/globals, presence of direct `eval()` calls, etc. So the bytecode generator has less detailed information when it runs on the lazily parsed stuff.

RubenKelevra:
I did an in depth review and I think I found a couple of issues.

First:

Strict mode is not being applied during lazy syntax-check parsing Lazy inner functions that start with `'use strict'` are syntax-checked as non-strict code. That means strict-only early errors are silently accepted in lazy bodies.

`function f(){ 'use strict'; with({}){} }` is correctly rejected while
`function outer(){ function inner(){ 'use strict'; with({}){} } }` would be incorrectly accepted.

Second:

Deeply nested lazy functions are brace-skipped instead of parsed at deeper nesting, syntax-check mode falls back to brace counting and does not fully parse inner bodies. As a result, grammar errors that must be parse-time errors are missed.

`function outer(){ function inner(){ var x = ; } }` is rejected while
`function a(){ function b(){ function c(){ var x = ; } } }` is currently accepted.

This should be a violation of the early-error expectations and would create depth-dependent parser behavior.

Three:

The free-variable tracking in syntax-check mode uses hash-only keys name tracking is done via 64-bit hashes without preserving full identity in the declared/used sets. A collision can misclassify captures and environment requirements. The probability is low, but the failure mode is semantic miscompilation, so this is not a good tradeoff.

_Edit:_
In addition, I recommend adding per-session randomization to the hashing to reduce collision-DoS risk from crafted inputs, which would lead to constant fallbacks to byte comparisons. The default implementation Rust uses does do this too, while using a much stronger Hash Algorithm (SipHash) - with the typical speed tradeoffs.

Overall I don't think FxHash is a good fit here and would advise a switch to Highway-Hash instead, which offers a native per session randomization and is cryptographically secure, so byte comparison fallbacks are very unlikely even with truncated 64 bit hash.

jdahlin:
> I did an in depth review and I think I found a couple of issues.
> 
> ...

Thanks for the review! These should all be addressed now. Its a bit worrisome that test262 did not detect these issues.

RubenKelevra:
Thanks, I'll have another look if I got time.

RubenKelevra:
> > I did an in depth review and I think I found a couple of issues.
> > ...
> 
> Thanks for the review! These should all be addressed now. Its a bit worrisome that test262 did not detect these issues.

Thanks for your comment, this sent me on a tangent why the other two parser bugs I found in your updated implementation are not found by the test harness. And turns out the test harness is partly not checking the Rust implementation, causing it to miss bugs in the Rust implementation in the master as well.

Guess I'll prepare some PRs.

RubenKelevra:
@jdahlin Please run your test again with this PR applied:

https://github.com/LadybirdBrowser/ladybird/pull/8415

It should now at least fail with a stack overflow or segfault because of the missing depth check on your newly introduced recursion.

But the parser in `master` does currently do the same thing.

RubenKelevra:
@jdahlin The second issue in your recursion syntax check implementation is, that you're not properly restoring the shared parser mode flags (strict/generator/async) after returning from walking into a nested function. This causes inner-function semantics to leak into the outer function and will produce SyntaxErrors on valid code.

RubenKelevra:
I've added the tests which I used to check my assumptions as PR #8423, so they can be used in the future to determine if similar issues may arise from optimizations.

jdahlin:
@RubenKelevra there are no longer any regression on the full test262 with this change (compared to master).

RubenKelevra:
@jdahlin I did try your current patchset one commit at a time, with my updated PRs #8415, #8421 and #8423 as a basis on top of upstream/master (from yesterday).

Multiple commits result in warnings and 531f131a14 starts to crash:

```

0% tests passed, 4 tests failed out of 4

Total Test time (real) =   7.00 sec

The following tests FAILED:
        123 - test-js (Subprocess aborted)
        124 - test-js-cpp (Subprocess aborted)
        125 - test-js-bytecode (Failed)
        126 - test-js-ast (Failed)
Errors while running CTest
```

Suggestion: Make commits self-contained in their ability to produce a clean build state, with clean test.

---

**But more importantly, the final state of your PR `6958053b39` cherry-picked on top it fails some tests:**

<details>

```
Test project /home/ruben/dev/ladybird_rust_js_parser/Build/release
    Start 123: test-js
1/4 Test #123: test-js ..........................***Failed   15.87 sec
 PASS  builtins/DisposableStack/DisposableStack.prototype.@@dispose.js
    ⚠  Suite:  used in using functionality
         Test:   make the stack marked as disposed (expected fail)
 PASS  builtins/FinalizationRegistry/FinalizationRegistry.prototype.cleanupSome.js
    ⚠  Suite:  <top-level>
         Test:   basic functionality (expected fail)
 PASS  builtins/WeakMap/WeakMap.prototype.set.js
    ⚠  Suite:  <top-level>
         Test:   automatic removal of garbage-collected values (expected fail)
 PASS  builtins/WeakSet/WeakSet.prototype.add.js
    ⚠  Suite:  <top-level>
         Test:   automatic removal of garbage-collected values (expected fail)
 FAIL  /home/ruben/dev/ladybird_rust_js_parser/Tests/LibJS/Runtime/functions/function-destructuring-parameters.js
    ❌ The file failed to parse

                 function testFunction({ bar, ...a }, { bar, ...b }) {
                                                                   ^
         Duplicate parameter 'bar' not allowed (line: 109, column: 59)

 PASS  modules/basic-modules.js
    ⚠  Suite:  in- and exports
         Test:   can have top level using declarations which trigger at the end of running a module (expected fail)
 FAIL  panic-reachability-audit.js
    ❌ Suite:  <top-level>
         Test:   deeply nested scope errors don't crash (failed):
                 ExpectationError: Expected _function a() { function b() { function c() { let x = ; } } }_ not to eval but it did
 FAIL  parser-lazy-nested-syntax-check.js
    ❌ Suite:  <top-level>
         Test:   deeply nested lazy functions still syntax-check inner function bodies (failed):
                 ExpectationError: Expected _function a(){ function b(){ function c(){ var x = ; } } }_ not to eval but it did
 FAIL  parser-lazy-strict-mode-early-errors.js
    ❌ Suite:  <top-level>
         Test:   strict mode inherited from outer scope applies inside lazy inner functions (failed):
                 ExpectationError: Expected _"use strict"; function outer(){ function inner(){ with({}){} } }_ not to eval but it did
 FAIL  syntax/early-errors-in-nested-functions.js
    ❌ Suite:  syntax errors in inner functions are early errors
         Test:   syntax error in deeply nested function (failed):
                 ExpectationError: Expected _function a() { function b() { function c() { var x = ; } } }_ not to eval but it did
         Test:   inherited strict mode in lazily-parsed function (failed):
                 ExpectationError: Expected _'use strict'; function outer(){ function inner(){ with({}){} } }_ not to eval but it did
         Test:   syntax error at depth 4 (failed):
                 ExpectationError: Expected _function a(){ function b(){ function c(){ function d(){ var x = ; } } } }_ not to eval but it did
 PASS  using-declaration.js
    ⚠  Suite:  basic usage
         Test:   disposes after block exit (expected fail)
         Test:   disposes in reverse order after block exit (expected fail)
         Test:   disposes in reverse order after block exit even in same declaration (expected fail)
    ⚠  Suite:  behavior with exceptions
         Test:   is run even after throw (expected fail)
         Test:   throws error if dispose method does (expected fail)
         Test:   if block and using throw get suppressed error (expected fail)
         Test:   multiple throwing disposes give suppressed error (expected fail)
         Test:   3 throwing disposes give chaining suppressed error (expected fail)
         Test:   normal error and multiple disposing errors give chaining suppressed errors (expected fail)
    ⚠  Suite:  works in a bunch of scopes
         Test:   works in block (expected fail)
         Test:   works in static class block (expected fail)
         Test:   works in function (expected fail)
         Test:   switch block is treated as full block in function (expected fail)
    ⚠  Suite:  invalid using bindings
         Test:   nullish values do not throw (expected fail)
         Test:   non-object throws (expected fail)
         Test:   object without dispose throws (expected fail)
         Test:   object with non callable dispose throws (expected fail)
    ⚠  Suite:  using is still a valid variable name
         Test:   using (expected fail)
 PASS  using-for-loops.js
    ⚠  Suite:  basic usage
         Test:   using in normal for loop (expected fail)
         Test:   using in normal for loop with expression body (expected fail)
         Test:   using in for of loop (expected fail)
         Test:   using in for of loop with expression body (expected fail)
         Test:   can have multiple declaration in normal for loop (expected fail)
         Test:   can have using in block in for loop (expected fail)
         Test:   can have using in block in for-in loop (expected fail)
         Test:   dispose is called even if throw in for of loop (expected fail)
    ⚠  Suite:  using is still a valid variable in loops
         Test:   using using of (expected fail)

Test Suites: 4 failed, 1871 passed, 1875 total
Tests:       6 failed, 32 expected failed, 6302 passed, 6340 total
Files:       1071 total
Time:        15.66s
```

</details>

jdahlin:
@RubenKelevra wrote:
> @jdahlin I did try your current patchset one commit at a time, with my updated PRs #8415, #8421 and #8423 as a basis on top of upstream/master (from yesterday).
[...]

Thanks for testing, I will return to this once your patches has landed and I will work on making the changes self-contained/rebasable.

RubenKelevra:
> @RubenKelevra wrote:
> 
> > @jdahlin I did try your current patchset one commit at a time, with my updated PRs #8415, #8421 and #8423 as a basis on top of upstream/master (from yesterday).
> > [...]
> 
> Thanks for testing, I will return to this once your patches has landed and I will work on making the changes self-contained/rebasable.

#8415 has landed, this makes it very easy now to build yourself the validation locally: You can just cherry-pick all commits from #8421 and #8423 and run the test as before. It will now properly validate your changes to the rust compiler :)

If that's helpful, I pushed the temporary branch I did where I merged the changes for validation:

https://github.com/RubenKelevra/LadybirdBrowser_ladybird/tree/__temp_pr8408_test

jdahlin:
At the moment this PR is too complex and I fell the improvements are not as much as I thought. I managed to get all the ast and bytecode tests passing locally but still around 2kloc of changes. There are plenty of other low-hanging fruit that can be done first.

I might revisit this once everything else is in order

PR #8409, opened 2026-03-14
LibWeb: Implement CSS `inherit()` function

Calme1709:
The remaining failing imported tests are due to wider issues which are covered by FIXMEs (both existing and added in this PR)


Libraries/LibWeb/CSS/Parser/ArbitrarySubstitutionFunctions.cpp
```diff
@@ -345,6 +347,44 @@ static Vector<ComponentValue> replace_an_if_function(DOM::AbstractElement& eleme
     return {};
 }
 
+// https://drafts.csswg.org/css-values-5/#replace-an-inherit-function
+static Vector<ComponentValue> replace_an_inherit_function([[maybe_unused]] DOM::AbstractElement& element, [[maybe_unused]] GuardedSubstitutionContexts& guarded_contexts, [[maybe_unused]] ArbitrarySubstitutionFunctionArguments const& arguments)
```
tcl3: It looks like these `[maybe_unused]]` annotations can be removed.

Calme1709: Whoops - removed.

PR #8414, opened 2026-03-14
Tests: Use HTTP server when running `grid-gutters-and-tracks-001.html`

tcl3:
Using [LoadFromHttpServer] is required to load the Ahem font used by this test.


shannonbooth:
Is the LIBJS_COMPARE_PIPELINES=1 thing completely unrelated to the issue here?

tcl3:
> Is the LIBJS_COMPARE_PIPELINES=1 thing completely unrelated to the issue here?

I think the difference in behavior between the 2 pipelines that this uncovered is still there, but it can no longer be reproduced with this particular test. I've left a comment on the issue with a different reproduction.

PR #8418, opened 2026-03-14
LibJS: Replace FunctionTable sparse Vec with HashMap

jdahlin:
Switch `FunctionTable` from a sparse `Vec<Option<Box<FunctionData>>>` to `HashMap<FunctionId, Box<FunctionData>>`
The previous representation allocated slots for every ID between 0 and max, wasting memory when IDs are sparse (common in large bundles with many modules)
Reduces private memory on macOS when loading https://vscode.dev from ~5.0 GB to ~1.4 GB

### DIFF
--- a/Libraries/LibJS/Rust/src/ast.rs
+++ b/Libraries/LibJS/Rust/src/ast.rs
@@ -26,7 +26,7 @@ use std::ffi::c_void;
 use std::fmt;
 use std::rc::Rc;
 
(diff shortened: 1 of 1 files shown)


jdahlin:
the failures are flaky tests fullscreen/api/element-request-fullscreen-timing.html and css/css-color/parsing/color-computed-hsl.html

PR #8419, opened 2026-03-14
LibWeb: Use requestAnimationFrame to update media controls timeline

Zaggy1024:
The media element's timeupdate event only fires every 250ms. To make the timeline update more smoothly, poll currentTime to update the timeline every frame.

### DIFF
--- a/Libraries/LibWeb/HTML/MediaControls.h
+++ b/Libraries/LibWeb/HTML/MediaControls.h
@@ -73,6 +73,7 @@ private:
         GC::Weak<DOM::IDLEventListener> listener;
     };
     Vector<RegisteredEventListener> m_registered_event_listeners;
+    GC::Weak<WebIDL::CallbackType> m_request_animation_frame_callback;
 
     enum class Scrubbing : u8 {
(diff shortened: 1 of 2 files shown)


Libraries/LibWeb/HTML/MediaControls.h
```diff
@@ -73,6 +73,7 @@ class MediaControls {
         GC::Weak<DOM::IDLEventListener> listener;
     };
     Vector<RegisteredEventListener> m_registered_event_listeners;
+    GC::Ptr<WebIDL::CallbackType> m_request_animation_frame_callback;
```
shannonbooth: This seems like it should be visited or made weak (unsure of the lifetimes here)

Zaggy1024: Good point, the Clang plugin also pointed that out.

Zaggy1024: Should be fine with `GC::Weak` like the rest, since the callback has to live as long as the RAF is in flight.

PR #8415, opened 2026-03-14
LibJS: Route canParseSource() through Script::parse() and align rust parser behavior

RubenKelevra:
This fixes a split-brain in the `test-js` harness.

`canParseSource()` was using the C++ parser, while runtime parsing goes through `Script::parse()` (Rust-first). That meant parser-facing tests could pass while exercising a different path than what the runtime would execute, masking non-alignment with the standard.

This PR routes `canParseSource()` through `Script::parse()` and adds explicit parser-mode selection for deterministic C++ vs Rust runs.

The observed outcome was that some tests failed in Rust mode, which wasn't visible before.

I included the following fixes for the rust parser, as requested:

- enforce const initializers in declarations
- align duplicate-parameter early errors with expected behavior
- fix async-arrow and `for await ... of` parsing corner cases
- reject invalid bare private-identifier usage
- isolate `super` validity checks in nested function scopes
- reject optional chaining on unparenthesized `new`
- keep valid call-paren `new` optional-chaining cases working
- align new-expression parenthesis flags with C++ parser

PR #8420, opened 2026-03-14
LibJS: Add deep nesting parser conformance tests

RubenKelevra:
This PR adds 12 tests which currently crash both parsers reliably with SIGSEGV. PR #8415 allows to check both parsers - without it will check only the C++ parser.

**The expected result is at the moment is, that this tests is failing on the C++ and the Rust parser.**

Fix for the Rust parser is filed in the separate PR #8421.


Calme1709:
PRs which fail CI cannot be merged - these tests should instead be included in the PR (ideally commit) which fixes them :)

RubenKelevra:
Yeah I just wanted to keep the commits small, this one requires fixed for the Rust and the C++ parser, that's why I filed a dedicated PR for the Rust parser.

But okay, I'll integrate it in the matching fix PRs. 

Thanks for the review!

PR #8423, opened 2026-03-15
LibJS: Add additional lazy-parser early-error coverage

RubenKelevra:
This PR adds more parser coverage around lazy function parsing and early-error behavior.

These issues surfaced on review of PR #8408 and are not present currently in the master. <strike>The tests are expected to succeed for the C++ parser, and with #8415 applied they are also executed and succeeding for the Rust parser.</strike>

_Edit: Accidentally wrote the tests to require PR #8415 to be merged, which is not strictly necessary._

### DIFF
--- a/Tests/LibJS/Runtime/parser-lazy-nested-syntax-check.js
+++ b/Tests/LibJS/Runtime/parser-lazy-nested-syntax-check.js
@@ -1,0 +1,5 @@
+test("deeply nested lazy functions still syntax-check inner function bodies", () => {
+    const source = "function a(){ function b(){ function c(){ var x = ; } } }";
+
+    expect(source).not.toEval();
+});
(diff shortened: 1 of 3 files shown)


RubenKelevra:
Not stale.

jdahlin:
@RubenKelevra have you looked at what v8 and jsc does here? They will delay work until functions are materialized in a way that is not _always_ spec compliant. As far as I know, V8 defer duplicate bindings, label resolution, private member access kinds.

RubenKelevra:
@jdahlin nope. But I know that they bent the rules quite a bit in terms of standard compliance. My assumption however was, that they skip the "expensive" tests like duplicate `let x;` inside the same context, but the preparers should still reliably capture structural syntax errors, like the ones I propose to add.

So I did not check if v8 or jsc would report them properly, but I think we should - and we do currently. This is just a regression test, so we don't optimize in the future minimum prechecks out.

RubenKelevra:
@jdahlin have you overlooked my reply?

PR #8432, opened 2026-03-15
LibWebView: Default to http:// for bare IP addresses in URL bar

### LINKED ISSUE #8344: URL bar defaults to https:// for bare IP addresses, breaking local network access
### Summary

When entering a bare IP address (e.g., 192.168.1.1 or 127.0.0.1) into the URL bar, the browser automatically prepends the https:// scheme. 

While defaulting to HTTPS is a good security practice for domains, it causes immediate connection failures or certificate validation errors for bare IP addresses, especially local ones which generally do not possess valid public SSL/TLS certificates.

We should refine the URL guessing logic in Libraries/LibWebView/URL.cpp. For private/loopback IP addresses, it should default to http://. For public IP addresses, defaulting to https:// makes sense, but the browser should ideally provide an interstitial warning allowing users to downgrade to HTTP if the secure connection fails.

### Operating system

macOS

### Steps to reproduce

1. Run Ladybird
2. Focus the URL bar and type a local bare IP address (e.g. 127.0.0.1)
3. Press Enter

### Expected behavior

The browser should navigate to http://127.0.0.1/ since local network IP addresses rarely support HTTPS without producing certificate errors.

### Actual behavior

The browser auto-completes to https://127.0.0.1/, which results in a connection error or an invalid certificate warning.

N/A

```html
N/A
```

```shell
332001.505 WebContent(75431): Handling message while connection closed: WebContentServer::CloseServer
332001.534 WebContent(75439): ResourceLoader: Failed load of: "https://1.1.1.1/", Error: Request finished with error: Unable to connect, Duration: 2ms
```

_No response_

_No response_

### Contribute a patch?

mvanhorn:
## Summary

When entering a bare IP address (e.g. `192.168.1.1`) into the URL bar without a scheme, the browser defaults to `https://`. This causes connection failures for local network devices that do not have TLS certificates. After guessing the scheme, this checks if the host is an IPv4 or IPv6 address and switches to `http://` instead.

## Changes

- In `sanitize_url()`, after the `https_scheme_was_guessed` block, detect IP hosts using `Host::has<IPv4Address>()` / `Host::has<IPv6Address>()` and set the scheme to `http`
- Added 6 test cases in `TestWebViewURL.cpp` covering IPv4, IPv6, IPv4 with port, and explicit scheme preservation

## Testing

- Bare IPv4 (`192.168.1.1`) resolves to `http://192.168.1.1/`
- Bare IPv6 (`[::1]`) resolves to `http://[::1]/`
- IPv4 with port (`10.0.0.1:8080`) resolves to `http://10.0.0.1:8080/`
- Explicit `https://192.168.1.1` is preserved (user override respected)
- Domain names (`example.org`) still default to `https://`

Fixes #8344

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/Libraries/LibWebView/URL.cpp
+++ b/Libraries/LibWebView/URL.cpp
@@ -44,6 +44,12 @@ Optional<URL::URL> sanitize_url(StringView location, Optional<SearchEngine> cons
         https_scheme_was_guessed = true;
     }
 
+    // IP addresses should default to http:// since they rarely have valid TLS certificates.
+    if (https_scheme_was_guessed) {
+        if (auto const& host = url->host(); host.has_value() && (host->has<IPv4Address>() || host->has<IPv6Address>()))
+            url->set_scheme("http"_string);
+    }
+
     // FIXME: Add support for other schemes, e.g. "mailto:". Firefox and Chrome open mailto: locations.
     static constexpr Array SUPPORTED_SCHEMES { "about"sv, "data"sv, "file"sv, "http"sv, "https"sv, "resource"sv };
     if (!any_of(SUPPORTED_SCHEMES, [&](StringView const& scheme) { return scheme == url->scheme(); }))
--- a/Libraries/LibWebView/URL.cpp
+++ b/Libraries/LibWebView/URL.cpp
@@ -14,6 +14,56 @@
 
 namespace WebView {
 
+static bool is_local_ip_address(URL::URL const& url)
+{
+    auto const& host = url.host();
+    if (!host.has_value())
+        return false;
+
+    if (host->has<IPv4Address>()) {
+        auto const& addr = host->get<IPv4Address>();
+        // 127.0.0.0/8
+        if (addr[0] == 127)
+            return true;
+        // 10.0.0.0/8
+        if (addr[0] == 10)
+            return true;
+        // 172.16.0.0/12
+        if (addr[0] == 172 && addr[1] >= 16 && addr[1] <= 31)
+            return true;
+        // 192.168.0.0/16
+        if (addr[0] == 192 && addr[1] == 168)
+            return true;
+        // 169.254.0.0/16 (link-local)
+        if (addr[0] == 169 && addr[1] == 254)
+            return true;
+        return false;
+    }
+
+    if (host->has<IPv6Address>()) {
+        auto const& addr = host->get<IPv6Address>().to_in6_addr_t();
+        // ::1 (loopback)
+        bool is_loopback = true;
+        for (int i = 0; i < 15; ++i) {
+            if (addr[i] != 0) {
+                is_loopback = false;
+                break;
+            }
+        }
+        if (is_loopback && addr[15] == 1)
+            return true;
+        // fe80::/10 (link-local)
+        if (addr[0] == 0xfe && (addr[1] & 0xc0) == 0x80)
+            return true;
+        // fc00::/7 (unique local)
+        if ((addr[0] & 0xfe) == 0xfc)
+            return true;
+        return false;
+    }
+
+    return false;
+}
+
 Optional<URL::URL> sanitize_url(StringView location, Optional<SearchEngine> const& search_engine, AppendTLD append_tld)
 {
     auto search_url_or_error = [&]() -> Optional<URL::URL> {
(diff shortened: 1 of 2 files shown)


Libraries/LibWebView/URL.cpp
```diff
@@ -44,6 +44,12 @@ Optional<URL::URL> sanitize_url(StringView location, Optional<SearchEngine> cons
         https_scheme_was_guessed = true;
     }
 
+    // IP addresses should default to http:// since they rarely have valid TLS certificates.
+    if (https_scheme_was_guessed) {
+        if (auto const& host = url->host(); host.has_value() && (host->has<IPv4Address>() || host->has<IPv6Address>()))
```
shannonbooth: Logically this should probably go into the if statement above where https_scheme_was_guessed is set, and https_scheme_was_guessed is not set to true if there was an IP address given (since we didn't guess https). Though, it does seem a bit weird if we give a different guess of scheme for localhost compared to `127.0.0.1`/ `[::1]`

Libraries/LibWebView/URL.cpp
```diff
@@ -44,6 +44,12 @@ Optional<URL::URL> sanitize_url(StringView location, Optional<SearchEngine> cons
...
+            url->set_scheme("http"_string);
```
trflynn89, line 98: IMO we should only do this for local IP addresses. Public IPs should still default to https.

Tests/LibWebView/TestWebViewURL.cpp
```diff
@@ -189,6 +189,13 @@ TEST_CASE(location_to_search_or_url)
     expect_url_equals_sanitized_url("https://localhost/hello.world"sv, "localhost/hello.world"sv);
     expect_url_equals_sanitized_url("https://localhost/hello.world?query=123"sv, "localhost/hello.world?query=123"sv);
 
+    expect_url_equals_sanitized_url("http://192.168.1.1/"sv, "192.168.1.1"sv);          // IPv4: default to http.
+    expect_url_equals_sanitized_url("http://127.0.0.1/"sv, "127.0.0.1"sv);              // Loopback: default to http.
+    expect_url_equals_sanitized_url("http://10.0.0.1:8080/"sv, "10.0.0.1:8080"sv);      // IPv4 with port: default to http.
```
shannonbooth: It might be good to also add a test case for something like 127.0.0.1:443 being given. With these changes, the URL is changed from `127.0.0.1:443` to `http://127.0.0.1` which is not what I would expect since the HTTPS default port was given

Libraries/LibWebView/URL.cpp
mvanhorn: Addressed in 10ced81c: now only defaults to http:// for local/private IPs (RFC 1918 + loopback + link-local). Public IPs default to https:// as before.

Libraries/LibWebView/URL.cpp
mvanhorn: Addressed in 10ced81c: moved the IP check into the existing if block where https_scheme_was_guessed is set. For local IPs, the scheme is set to http:// without setting https_scheme_was_guessed.

Tests/LibWebView/TestWebViewURL.cpp
mvanhorn: Addressed in 10ced81c: added test cases for 127.0.0.1:443 (expects https since port 443 is explicit), 8.8.8.8 (expects https, public IP), and [2606:4700:4700::1111] (expects https, public IPv6).

Tests/LibWebView/TestWebViewURL.cpp
```diff
@@ -189,6 +189,16 @@ TEST_CASE(location_to_search_or_url)
     expect_url_equals_sanitized_url("https://localhost/hello.world"sv, "localhost/hello.world"sv);
```
ApsonCinesRy, line 189: To add to this, shouldn't localhost default to HTTP instead of HTTPS as well? I would imagine so given that it generally resolves to 127.0.0.1 or ::1

theeyeofcthulhu:
I've also been investigating in this direction, and this maybe falls under this PR: `localhost` isn't the only hostname in the local network. Sometimes I like to open web servers on my local machines to exchange files between them. Let's say my laptop's hostname is `laptop` and I opened a (HTTP) web server on port 8000. In Firefox, typing `laptop:8000` into the search bar will get me to the server, while Ladybird will default to querying the search engine. If localhost is addressed via HTTP as  @dominick038 [suggested](https://github.com/LadybirdBrowser/ladybird/pull/8432#discussion_r2948226827), we could, to implement this, amend this check…
https://github.com/LadybirdBrowser/ladybird/blob/847cf74ff81a2f58809dbbe85c31930be3e7ab09/Libraries/LibWebView/URL.cpp#L48-L50
…by checking first, using a similar method as here…
https://github.com/LadybirdBrowser/ladybird/blob/847cf74ff81a2f58809dbbe85c31930be3e7ab09/Libraries/LibWebView/URL.cpp#L40-L42
…if this were a HTTP URL, would it have a port? Because then the user evidently wanted to address something, since they bothered with the port. Maybe you could also abandon the hard-coded check for `localhost` above and always do it with the port, since when visiting servers hosted locally, you also always provide the port.

mvanhorn:
Good point about other local hostnames. The current PR scope is intentionally narrow (bare IPs and localhost), but extending to mDNS names (`.local`) and other RFC-defined local hostnames could be a follow-up. Happy to expand if the maintainers prefer it in this PR.

PR #8435, opened 2026-03-15
Two little XHR optimizations

awesomekling:
Noticed these while investigating why [Diablo](https://d07riv.github.io/diabloweb/) loaded so much faster in other browsers.

### DIFF
--- a/Libraries/LibWeb/XHR/XMLHttpRequest.cpp
+++ b/Libraries/LibWeb/XHR/XMLHttpRequest.cpp
@@ -801,6 +801,10 @@ WebIDL::ExceptionOr<void> XMLHttpRequest::send(Optional<DocumentOrXMLHttpRequest
             if (!length.has<u64>())
                 length = 0;
 
+            // Pre-allocate received bytes buffer if Content-Length is known.
+            if (length.get<u64>() > 0)
+                m_received_bytes.ensure_capacity(length.get<u64>());
+
             // 10. Let processBodyChunk given bytes be these steps:
             auto process_body_chunks = GC::create_function(heap(), [this, length](ByteBuffer byte_buffer) {
                 // 1. Append bytes to this’s received bytes.


Libraries/LibWeb/XHR/XMLHttpRequest.cpp
```diff
@@ -801,6 +801,10 @@ WebIDL::ExceptionOr<void> XMLHttpRequest::send(Optional<DocumentOrXMLHttpRequest
             if (!length.has<u64>())
                 length = 0;
 
+            // Pre-allocate received bytes buffer if Content-Length is known.
+            if (length.has<u64>() && length.get<u64>() > 0)
+                m_received_bytes.ensure_capacity(length.get<u64>());
```
trflynn89: We're guaranteed to have a `u64` here due to the above step so this `if` can probably be removed

Libraries/LibWeb/XHR/XMLHttpRequest.cpp
```diff
@@ -808,6 +801,10 @@ WebIDL::ExceptionOr<void> XMLHttpRequest::send(Optional<DocumentOrXMLHttpRequest
...
```
konradekk, line 806: consider putting a reasonable limit to it? (few big allocations shall be okay; user might just cancel the download in the mean time etc. 0.1 gb should be okay? 0.5 gb?)

Libraries/LibWeb/XHR/XMLHttpRequest.cpp
```diff
@@ -203,14 +203,7 @@ WebIDL::ExceptionOr<JS::Value> XMLHttpRequest::response()
...
         // then set this’s response object to a new ArrayBuffer object representing this’s received bytes. If this throws an exception, then set this’s response object to failure and return null.
-        auto buffer_result = JS::ArrayBuffer::create(realm(), m_received_bytes.size());
-        if (buffer_result.is_error()) {
-            m_response_object = Failure();
-            return JS::js_null();
-        }
-
-        auto buffer = buffer_result.release_value();
-        buffer->buffer().overwrite(0, m_received_bytes.data(), m_received_bytes.size());
+        auto buffer = JS::ArrayBuffer::create(realm(), move(m_received_bytes));
```
konradekk, line 206: it seems it deviates non-trivially from the spec; worth putting an in-code comment here?

PR #8437, opened 2026-03-15
CMake: Enable -dead_strip on macOS

jdahlin:
## Summary
- Enable `-dead_strip` linker flag on macOS to remove unreachable code and data from final binaries
- This is the macOS equivalent of `--gc-sections` on ELF and `/OPT:REF` on MSVC (the latter is already enabled)

### Size reduction on arm64 RelWithDebInfo build

| Binary | Before | After | Delta |
|---|---|---|---|
| liblagom-web.dylib | 61,430,216 | 58,137,704 | -3215 KiB |
| liblagom-js.dylib | 11,031,440 | 10,546,640 | -473 KiB |
| liblagom-wasm.dylib | 3,175,880 | 3,075,640 | -98 KiB |
| liblagom-unicode.dylib | 2,713,432 | 2,629,144 | -82 KiB |
| liblagom-webview.dylib | 2,585,752 | 2,546,168 | -39 KiB |
| Ladybird | 621,064 | 593,144 | -27 KiB |

### DIFF
--- a/Meta/CMake/lagom_compile_options.cmake
+++ b/Meta/CMake/lagom_compile_options.cmake
@@ -23,6 +23,7 @@ endif()
 
 if (APPLE)
     list(APPEND CMAKE_PREFIX_PATH /opt/homebrew)
+    add_cxx_link_options(LINKER:-dead_strip)
 endif()
 
 if (CMAKE_BUILD_TYPE STREQUAL "Debug")


Meta/CMake/common_compile_options.cmake
```diff
@@ -91,9 +91,11 @@ if (NOT MSVC)
     add_cxx_compile_options(-fstrict-flex-arrays=2)
     add_cxx_compile_options(-fstack-protector-strong)
     add_cxx_link_options(-fstack-protector-strong)
+    if (UNIX AND NOT ENABLE_FUZZERS)
+        add_cxx_compile_options(-fvisibility-inlines-hidden)
+    endif()
     if (UNIX AND NOT APPLE AND NOT ENABLE_FUZZERS)
         add_cxx_compile_options(-fno-semantic-interposition)
-        add_cxx_compile_options(-fvisibility-inlines-hidden)
```
gmta, line 96: > this further reduces binary sizes by ~1%

Please run those numbers again. Based on what you wrote in the commit message, it's not even close to 1%, and you even have a size regression.

If you create PRs using AI, make sure [to double check the output](https://github.com/LadybirdBrowser/ladybird/blob/master/CONTRIBUTING.md#on-usage-of-ai-and-llms)!

jdahlin: Sorry, I was doing too many things in the same checkout, the numbers were likely correct but external changes caused them to go down. I have removed this commit. Thanks for checking!

jdahlin:
@gmta this keeps failing in the most interesting ways in CI

gmta:
> @gmta this keeps failing in the most interesting ways in CI

Looks like a valid CI failure for macOS. Please check the job log.

jdahlin:
@gmta finally passed!

Libraries/LibTest/AssertionHandler.cpp
```diff
@@ -38,7 +38,7 @@ static void assertion_handler_impl(char const*)
 #if defined(AK_OS_WINDOWS)
 #    define EXPORT __declspec(dllexport)
 #else
-#    define EXPORT __attribute__((visibility("default")))
+#    define EXPORT __attribute__((visibility("default"), retain, used))
```
gmta, line 41: This change should go _before_ the `-dead_strip` addition; otherwise tests would fail when testing your previous commit in isolation.

jdahlin: D'oh. Of course. I'll let you know when it passes again

jdahlin:
@gmta build passed again

PR #8438, opened 2026-03-15
AppKit: Fix NSToolbarItem warning about ambiguous view dimensions

jdahlin:
## Summary
- Fix the NSToolbarItem warning logged on every window open:
  `WARNING <NSToolbarItem> -> view was automatically measured but had an ambiguous height or width and the view's frame size had a zero height or width`
- The location search field was created with `[[LocationSearchField alloc] init]` (zero frame) and only received its width constraint in `windowDidResize:`, which fires after the toolbar's initial layout pass
- Set an initial width constraint (600pt) at creation time; `windowDidResize:` replaces it with the proper proportional width immediately after


rcorsi:
Tested on Macbook Pro 2019 Intel, warnings no longer appear.

jdahlin:
Possibly a duplicate of https://github.com/LadybirdBrowser/ladybird/pull/8501 cc @trflynn89

trflynn89:
whoops, yeah they fix the same issue

PR #8440, opened 2026-03-15
LibWeb: Use stop_fetch() for overlapping HTMLLinkElement fetches

ycarmon:
When a link element's href changes, the spec expects the callback to check if "el contributes a script-blocking style sheet" (step 6). This check examines current fetch state, which fails when old fetch callbacks run after a new fetch has started.

Use FetchController::stop_fetch() instead of abort() to prevent old fetch callbacks from executing entirely. This ensures only the current fetch's callback runs, allowing it to correctly check current state per spec without race conditions.


Zaggy1024:
Seems like a reasonable change, but it doesn't build. :(

Libraries/LibWeb/HTML/HTMLLinkElement.cpp
```diff
@@ -398,6 +398,12 @@ void HTMLLinkElement::fetch_and_process_linked_resource()
 // https://html.spec.whatwg.org/multipage/semantics.html#default-fetch-and-process-the-linked-resource
 void HTMLLinkElement::default_fetch_and_process_linked_resource()
 {
+    if (m_fetch_controller) {
+        m_fetch_controller->stop_fetch();
+        if (document().script_blocking_style_sheet_set().contains(*this))
+            document().script_blocking_style_sheet_set().remove(*this);
```
Zaggy1024: The "script-blocking style sheet" is something from the spec. We should have a comment here indicating why this is needed. If it isn't something indicated by the spec, it should be an `AD-HOC` comment accompanied by a link to an (existing or new) spec issue.

ycarmon: Cool, added a comment explaining that the spec says that if the fetch is no longer contributing to styling processing model, we remove it from the script_blocking_style_sheet_set, as described here
https://html.spec.whatwg.org/multipage/links.html#link-type-stylesheet:process-the-linked-resource

Zaggy1024: Given that link, I suspect that that step should actually take place in fetch_and_process_linked_resource()? Also, make sure to modify the FIXME for the relevant step of that algorithm in process_stylesheet_resource() (evaluate whether it's completely covered by this code, and if so, remove the FIXME: part, add an AD-HOC comment to say where it's being implemented).
Once that's done and pushed, I think this should be good to go.

Zaggy1024: Any update on this? If you're not able to follow up on this, I could try to investigate and amend this so we can get it merged.

ycarmon: sorry, missed the comment, I will finish this later today

ycarmon: moved the step to fetch_and_process_linked_resource

Libraries/LibWeb/HTML/HTMLLinkElement.cpp
```diff
@@ -385,6 +385,17 @@ GC::Ptr<Fetch::Infrastructure::Request> HTMLLinkElement::create_link_request(HTM
...
 {
+    if (m_fetch_controller) {
+        m_fetch_controller->stop_fetch();
+
+        // If the element contributes a script-blocking style sheet, remove it from its node document's
+        // script-blocking style sheet set. This mirrors step 6 of the process-the-linked-resource
+        // algorithm, which would normally happen when the fetch completes successfully or with error.
+        // https://html.spec.whatwg.org/multipage/links.html#link-type-stylesheet:process-the-linked-resource
+        if (document().script_blocking_style_sheet_set().contains(*this))
+            document().script_blocking_style_sheet_set().remove(*this);
```
Zaggy1024: Sorry for the pedantry, but doesn't this fulfill step 2 rather than step 6, since this code runs when we're fetching again? And comments not from the spec should be preceded by `AD-HOC:`, `NB:` or `FIXME:` to indicate that. `NB:` is probably most appropriate here, from what I can tell, since it's simply implementing an explicit step in a different place.

Also, checking `contains()` should be unnecessary here for two reasons: the return value indicates whether the value was present, and I don't think the step requires us to check or assert this.

Libraries/LibWeb/HTML/HTMLLinkElement.cpp
```diff
@@ -801,8 +810,9 @@ void HTMLLinkElement::process_stylesheet_resource(bool success, Fetch::Infrastru
     if (mime_type_string.has_value() && mime_type_string != "text/css"sv)
         success = false;
 
-    // FIXME: 2. If el no longer creates an external resource link that contributes to the styling processing model,
-    //           or if, since the resource in question was fetched, it has become appropriate to fetch it again, then return.
+    // FIXME: 2. If el no longer creates an external resource link that contributes to the styling processing model, then return.
+    // AD-HOC: The second condition of this step ("it has become appropriate to fetch it again") is handled in
+    //         fetch_and_process_linked_resource() by stopping the outdated fetch, which prevents this callback from running at all.
```
Zaggy1024:
```suggestion
    // FIXME: 2. If el no longer creates an external resource link that contributes to the styling processing model, or

    // if, since the resource in question was fetched, it has become appropriate to fetch it again, then return.
    // NB: This is handled in
    //     fetch_and_process_linked_resource() by stopping the outdated fetch, which prevents this callback from running at all.
```

It's better to keep the original spec text verbatim and then add onto it whatever explains why the step was implemented differently.

I also suspect that this fixes the fixme for the first half of this? From my reading, it seems like `external resource link` simply refers to the source being fetched, so if that becomes nothing, we should trigger the return.

ycarmon: If this doesn't deviate from spec I will remove the comment, I don't think it adds much.

Zaggy1024: We can certainly leave the FIXME for a follow-up, it's unclear to me whether we're covering every case for that condition with the logic in `fetch_and_process_linked_resource()`

Libraries/LibWeb/HTML/HTMLLinkElement.cpp
```diff
@@ -385,6 +385,12 @@ GC::Ptr<Fetch::Infrastructure::Request> HTMLLinkElement::create_link_request(HTM
 // https://html.spec.whatwg.org/multipage/semantics.html#fetch-and-process-the-linked-resource
 void HTMLLinkElement::fetch_and_process_linked_resource()
 {
+    // https://html.spec.whatwg.org/multipage/links.html#link-type-stylesheet:process-the-linked-resource
+    if (m_fetch_controller) {
+        m_fetch_controller->stop_fetch();
+        document().script_blocking_style_sheet_set().remove(*this);
```
Zaggy1024, line 392: This is closer to what I had in mind, but ideally we still have an NB comment here explaining what this is doing here.

Libraries/LibWeb/HTML/HTMLLinkElement.cpp
```diff
@@ -801,8 +805,10 @@ void HTMLLinkElement::process_stylesheet_resource(bool success, Fetch::Infrastru
     if (mime_type_string.has_value() && mime_type_string != "text/css"sv)
         success = false;
 
-    // FIXME: 2. If el no longer creates an external resource link that contributes to the styling processing model,
-    //           or if, since the resource in question was fetched, it has become appropriate to fetch it again, then return.
+    // NB: 2. If el no longer creates an external resource link that contributes to the styling processing model, or
+    //        if, since the resource in question was fetched, it has become appropriate to fetch it again, then return.
+    //     This is implemented in fetch_and_process_linked_resource() by stopping the outdated fetch,
+    //     which prevents this callback from running at all.
```
Zaggy1024, line 812: The `NB:` should precede only the non-spec text. I was a little confusing before, but `FIXME:` is kind of a special case in that we often put it before the step itself, when that step _is_ the fixme. In this case, the `NB:` should indicate that it's a non-spec note on the implementation of the preceding step.

Zaggy1024 (commented):
Sorry to pile on more comment nits 😓 I guess I wasn't as clear as I could have been before. Let me know if you have any questions.

Libraries/LibWeb/HTML/HTMLLinkElement.cpp
ycarmon: fixed the inline comments

Zaggy1024:
This looks good to me now! One last thing, could you squash the corrections? We like to keep fixes to things added/modified in prior commits all as one commit.

ycarmon:
Squashed all commits

Zaggy1024 (approved):
Thanks!

PR #8441, opened 2026-03-15
LibWeb: Align `HTMLEmbedElement` presentational hints with the spec

tcl3:
Previously, `text-align` was being set to invalid values causing a crash.

Fixes a crash on http://www.effectgames.com/games/mariodemo/.

NOTE: Other engines seem to map the value `bottom` to `baseline` but this doesn't match the spec at: https://html.spec.whatwg.org/multipage/rendering.html#attributes-for-embedded-content-and-images

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLEmbedElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLEmbedElement.cpp
@@ -47,11 +47,30 @@ void HTMLEmbedElement::apply_presentational_hints(GC::Ref<CSS::CascadedPropertie
 {
     Base::apply_presentational_hints(cascaded_properties);
     for_each_attribute([&](auto& name, auto& value) {
+        // https://html.spec.whatwg.org/multipage/rendering.html#attributes-for-embedded-content-and-images
         if (name == HTML::AttributeNames::align) {
-            if (value.equals_ignoring_ascii_case("center"sv))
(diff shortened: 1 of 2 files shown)


Libraries/LibWeb/HTML/HTMLEmbedElement.cpp
```diff
@@ -47,11 +47,30 @@ void HTMLEmbedElement::apply_presentational_hints(GC::Ref<CSS::CascadedPropertie
...
-            if (value.equals_ignoring_ascii_case("center"sv))
-                cascaded_properties->set_property_from_presentational_hint(CSS::PropertyID::TextAlign, CSS::KeywordStyleValue::create(CSS::Keyword::Center));
-            else if (value.equals_ignoring_ascii_case("middle"sv))
-                cascaded_properties->set_property_from_presentational_hint(CSS::PropertyID::TextAlign, CSS::KeywordStyleValue::create(CSS::Keyword::Middle));
+            auto vertical_align_keyword = [&]() -> Optional<CSS::Keyword> {
+                if (value.equals_ignoring_ascii_case("top"sv))
+                    return CSS::Keyword::Top;
+                if (value.equals_ignoring_ascii_case("texttop"sv))
+                    return CSS::Keyword::TextTop;
+                if (value.equals_ignoring_ascii_case("center"sv) || value.equals_ignoring_ascii_case("middle"sv)
```
Calme1709, line 57: I don't see `center` or `middle` in the spec, do we need an `AD-HOC` comment here?

PR #8442, opened 2026-03-15
LibWeb: Support border-radius in clip-path: inset()

mikiubo:


Libraries/LibWeb/CSS/StyleValues/BasicShapeStyleValue.cpp
```diff
@@ -56,7 +58,63 @@ Gfx::Path Inset::to_path(CSSPixelRect reference_box, Layout::Node const& node) c
         resolved_left *= f;
     }
 
-    return path_from_resolved_rect(resolved_top, reference_box.width().to_float() - resolved_right, reference_box.height().to_float() - resolved_bottom, resolved_left);
+    float left_edge = resolved_left;
+    float top_edge = resolved_top;
+    float right_edge = reference_box.width().to_float() - resolved_right;
+    float bottom_edge = reference_box.height().to_float() - resolved_bottom;
+
+    auto const& br_rect = border_radius->as_border_radius_rect();
```
Calme1709: We should use full words not abbreviations for variable names i.e. `border_radius_rect` - see [CodingStyle.md](https://github.com/LadybirdBrowser/ladybird/blob/master/Documentation/CodingStyle.md)

Libraries/LibWeb/CSS/StyleValues/BasicShapeStyleValue.cpp
```diff
@@ -56,7 +58,63 @@ Gfx::Path Inset::to_path(CSSPixelRect reference_box, Layout::Node const& node) c
...
+            LengthPercentage::from_style_value(br.horizontal_radius()),
+            LengthPercentage::from_style_value(br.vertical_radius())
+        };
+    };
+    auto tl = to_border_radius_data(*br_rect.top_left());
+    auto tr = to_border_radius_data(*br_rect.top_right());
+    auto brr = to_border_radius_data(*br_rect.bottom_right());
+    auto bl = to_border_radius_data(*br_rect.bottom_left());
+
+    CSSPixelRect inset_rect {
```
Calme1709, line 66: [The spec](https://drafts.csswg.org/css-shapes/#basic-shape-functions) says we should resolve percentages against the reference box, not the inset rect:

> All the lengths expressed in percentages are resolved from the used dimensions of the reference box.

This can be observed in [this WPT test](http://wpt.live/css/css-masking/clip-path/clip-path-inset-round-percent.html)

Libraries/LibWeb/CSS/StyleValues/BasicShapeStyleValue.cpp
mikiubo: think i fixed
thaks

Libraries/LibWeb/CSS/StyleValues/BasicShapeStyleValue.cpp
mikiubo: think i fixed
thanks

Libraries/LibWeb/Painting/Paintable.cpp
Calme1709, line 230: I don't think that we need to split this into two separate methods, we can just pass a new parameter called `reference_rect` which is used for the first portion (and maybe rename the existing `rect` to `border_rect` or something similar.

mikiubo: yes i also think your solution is better

Calme1709 (commented):
That looks good to me! Will obviously require a maintainer's approval though :)

Libraries/LibWeb/CSS/StyleValues/BasicShapeStyleValue.cpp
```diff
@@ -26,15 +29,14 @@ static Gfx::Path path_from_resolved_rect(float top, float right, float bottom, f
     return path;
 }
 
+// https://www.w3.org/TR/css-shapes/#funcdef-basic-shape-inset
```
tcl3: Please use the editor's draft URL instead: https://drafts.csswg.org/css-shapes/#funcdef-basic-shape-inset

mikiubo: i think i fixed

PR #8445, opened 2026-03-16
LibCrypto: Implement KMAC authentication support 

mikiubo:


Libraries/LibCrypto/Authentication/KMAC.cpp
```diff
@@ -0,0 +1,67 @@
+/*
+ * Copyright (c) 2025, mikiubo <michele.uboldi@gmail.com>
```
gmta: It's 2026!

Libraries/LibCrypto/Authentication/KMAC.cpp
```diff
@@ -0,0 +1,67 @@
...
+    size_t output_size = output_length_bits / 8;
+
+    static constexpr u8 empty_custom[0] {};
+    auto custom = customization.value_or(ReadonlyBytes { empty_custom, 0 });
+
+    OSSL_PARAM params[] = {
+        OSSL_PARAM_size_t(OSSL_MAC_PARAM_SIZE, &output_size),
+        OSSL_PARAM_octet_string(OSSL_MAC_PARAM_CUSTOM, const_cast<u8*>(custom.data()), custom.size()),
+        OSSL_PARAM_END
+    };
```
gmta: The `empty_custom` construct is a bit weird - maybe have a `Vector<OSSL_PARAM>` and just append the parameters you need, passing in `params.data()` into `EVP_MAC_init()`. That way, you don't have to call `OSSL_PARAM_octet_string()` when `customization` doesn't have a value.

Libraries/LibCrypto/Authentication/KMAC.cpp
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    if (EVP_MAC_update(ctx, message.data(), message.size()) != 1) {
+        EVP_MAC_CTX_free(ctx);
+        return Error::from_string_literal("EVP_MAC_update failed for KMAC");
+    }
+
+    auto buf = TRY(ByteBuffer::create_uninitialized(output_size));
+    size_t written = output_size;
+    if (EVP_MAC_final(ctx, buf.data(), &written, output_size) != 1) {
```
gmta, line 80: 1) Why do you initialize `written` with `output_size` instead of `0`?
2) Shouldn't we check the value of `written` here / compare it to `output_size`?

Libraries/LibWeb/Crypto/CryptoAlgorithms.cpp
```diff
@@ -10371,4 +10430,357 @@ WebIDL::ExceptionOr<JS::Value> AesOcb::get_key_length(AlgorithmParams const& par
...
+    if (maybe_result.is_error())
+        return WebIDL::OperationError::create(realm, "KMAC operation failed"_utf16);
+
+    return maybe_result.release_value();
+}
+
+static WebIDL::UnsignedLong kmac_default_key_length(StringView algorithm_name)
+{
+    return (algorithm_name == "KMAC128"sv) ? 128u : 256u;
+}
```
gmta: Maybe instead of taking in `StringView`s in these methods, you could determine the `Crypto::Authentication::KMACKind` first and pass that in?

Alternatively, `VERIFY` whether the incoming strings are what you expect them to be.

Libraries/LibCrypto/Authentication/KMAC.cpp
mikiubo: 🤦🤦
now i think i fixed

Libraries/LibCrypto/Authentication/KMAC.cpp
mikiubo: thanks!

Libraries/LibCrypto/Authentication/KMAC.cpp
mikiubo: this was my error i checked also the demo on the openssl github, they use the same logic you suggest.
now ithink is fixed

Libraries/LibWeb/Crypto/CryptoAlgorithms.cpp
mikiubo: you are right! i changed the logic a little bit i put those method in the kmac.cpp

let me know if you notice any other problem

Libraries/LibCrypto/Authentication/KMAC.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+// https://wicg.github.io/webcrypto-modern-algos/#kmac-operations-sign
+ErrorOr<ByteBuffer> KMAC::sign(ReadonlyBytes key, ReadonlyBytes message, u32 output_length_bits, Optional<ReadonlyBytes> customization) const
+{
+    if (output_length_bits == 0)
+        return Error::from_string_literal("KMAC output length must be greater than zero");
+
+    if ((output_length_bits % 8) != 0)
+        return Error::from_string_literal("KMAC output length must be a multiple of 8");
+
```
mikiubo, line 48: @gmta  i also added these check in the last push force

PR #8446, opened 2026-03-16
LibWeb: Map logical aliases at compute rather than cascade time 

Calme1709:
Doing this at compute rather than cascade time achieves a few things:
 1) Brings us into line with the spec
 2) Avoids the double cascade that was previously required to compute mapping contexts
 3) We now compute values of logical aliases, while `style_value_for_computed_property` maps logical aliases to their physical counterparts, this didn't account for all cases (i.e. if there was no layout node, Typed OM, etc)
 4) Removes a hurdle to moving other upstream processes (i.e. arbitrary substitution function resolution, custom property computation) to compute time as the spec requires


Calme1709:
Rebased to resolve conflicts

PR #8449, opened 2026-03-16
LibJS: Add fast path in async function await for non-thenable values

awesomekling:
Per spec, every `await` goes through PromiseResolve (which wraps the value in a new Promise via NewPromiseCapability) and then PerformPromiseThen (which creates PromiseReaction and JobCallback objects). This results in 13-16 GC cell allocations per await.

Add a fast path that detects two common cases:

1. Primitive values: These can never have a "then" property, so we can skip all promise wrapping and directly schedule the async function's continuation as a microtask.

2. Already-settled native Promises: If the promise has no own properties and its prototype is the intrinsic %Promise.prototype%, we can extract the result directly and schedule continuation.

For these cases, we bypass promise_resolve(), new_promise_capability(), create_resolving_functions(), perform_then(), PromiseReaction creation, and JobCallback creation -- replacing ~13 GC allocations with 1 (the GC::Function for the microtask job).

Big speedups on AsyncBench:

```
Suite       Test                          Speedup  Old (Mean ± Range)     New (Mean ± Range)
----------  --------------------------  ---------  ---------------------  ---------------------
AsyncBench  async-call-return.js            2.992  0.516 ± 0.514 … 0.518  0.172 ± 0.170 … 0.175
AsyncBench  async-class-method.js           3.553  0.783 ± 0.777 … 0.789  0.220 ± 0.219 … 0.222
AsyncBench  async-fan-out.js                1.501  0.281 ± 0.280 … 0.283  0.187 ± 0.186 … 0.189
AsyncBench  async-generator.js              1.063  0.292 ± 0.292 … 0.292  0.275 ± 0.273 … 0.277
AsyncBench  async-iife.js                   2.179  0.627 ± 0.622 … 0.631  0.288 ± 0.287 … 0.288
AsyncBench  async-nested-calls.js           2.767  0.862 ± 0.854 … 0.870  0.312 ± 0.309 … 0.314
AsyncBench  async-pipeline.js               1.624  0.570 ± 0.570 … 0.570  0.351 ± 0.350 … 0.351
AsyncBench  async-sequential-awaits.js      5.56   0.604 ± 0.603 … 0.604  0.109 ± 0.109 … 0.109
AsyncBench  async-try-catch-reject.js       1.963  0.691 ± 0.691 … 0.692  0.352 ± 0.349 … 0.356
AsyncBench  await-primitive.js              7.347  0.510 ± 0.502 … 0.518  0.069 ± 0.068 … 0.071
AsyncBench  await-resolved-promise.js       1.224  0.575 ± 0.568 … 0.582  0.470 ± 0.469 … 0.471
AsyncBench  await-thenable.js               0.99   0.038 ± 0.038 … 0.039  0.039 ± 0.039 … 0.039
AsyncBench  promise-all.js                  1.502  0.750 ± 0.746 … 0.755  0.499 ± 0.497 … 0.502
AsyncBench  promise-allSettled.js           1.452  0.861 ± 0.853 … 0.870  0.593 ± 0.589 … 0.597
AsyncBench  promise-chain.js                1.013  0.169 ± 0.168 … 0.170  0.166 ± 0.166 … 0.167
AsyncBench  promise-new-resolve.js          1.156  0.344 ± 0.343 … 0.346  0.298 ± 0.296 … 0.300
AsyncBench  promise-race.js                 1.515  0.753 ± 0.747 … 0.759  0.497 ± 0.492 … 0.503
AsyncBench  Total                           1.884  9.227                  4.898
```

### DIFF
--- a/Libraries/LibJS/Runtime/AsyncFunctionDriverWrapper.cpp
+++ b/Libraries/LibJS/Runtime/AsyncFunctionDriverWrapper.cpp
@@ -46,6 +46,37 @@ ThrowCompletionOr<void> AsyncFunctionDriverWrapper::await(JS::Value value)
     if (!m_suspended_execution_context)
         m_suspended_execution_context = vm.running_execution_context().copy();
 
+    // OPTIMIZATION: Fast path for non-thenable values.
+    //
+    // Per spec, PromiseResolve wraps non-Promise values in a new resolved promise,
+    // then PerformPromiseThen attaches reaction handlers and schedules a microtask.
+    // This creates 10+ GC objects per await.
+    //
+    // Since primitives can never have a "then" property, and already-settled native
+    // Promises with the %Promise% constructor don't need wrapping, we can skip all
+    // of that machinery and directly schedule the async function's continuation.
+    //
+    // For pending promises, or promises with a non-standard constructor, we fall
+    // through to the spec-compliant slow path.
+    if (!value.is_object()) {
+        // Primitive values are never thenable -- schedule resume directly.
+        schedule_resume(value, true);
+        return {};
+    }
+
+    if (auto promise = value.as_if<Promise>()) {
+        // Already-settled native Promises whose constructor is the intrinsic %Promise%.
+        auto* promise_prototype = realm.intrinsics().promise_prototype().ptr();
+        if (promise->state() != Promise::State::Pending
+            && promise->shape().property_count() == 0
+            && promise->shape().prototype() == promise_prototype
+            && promise_prototype->get_without_side_effects(vm.names.constructor) == Value(realm.intrinsics().promise_constructor())) {
+            schedule_resume(promise->result(), promise->state() == Promise::State::Fulfilled);
+            promise->set_is_handled();
+            return {};
+        }
+    }
+
     // 2. Let promise be ? PromiseResolve(%Promise%, value).
     auto* promise_object = TRY(promise_resolve(vm, realm.intrinsics().promise_constructor(), value));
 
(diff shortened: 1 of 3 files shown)


Libraries/LibJS/Runtime/AsyncFunctionDriverWrapper.cpp
```diff
@@ -46,6 +47,38 @@ ThrowCompletionOr<void> AsyncFunctionDriverWrapper::await(JS::Value value)
...
+    }
+
+    if (auto promise = value.as_if<Promise>()) {
+        // Already-settled native Promises whose constructor is the intrinsic %Promise%.
+        static Bytecode::StaticPropertyLookupCache constructor_cache;
+        auto* promise_prototype = realm.intrinsics().promise_prototype().ptr();
+        if (promise->state() != Promise::State::Pending
+            && promise->shape().property_count() == 0
+            && promise->shape().prototype() == promise_prototype
+            && TRY(promise_prototype->get(vm.names.constructor, constructor_cache)) == Value(realm.intrinsics().promise_constructor())) {
```
shannonbooth: Would this perhaps need to be a get without side effects here?

awesomekling: Damn, good observation. Let me figure that out and add test coverage regardless!

shannonbooth: Had an idea after taking a look at this that should also help out with allocations down GC path: https://github.com/LadybirdBrowser/ladybird/pull/8450

PR #8451, opened 2026-03-16
AppKit: Vertically center text in the location search field

noahhaasis:
The toolbar gives the NSSearchField more height than needed for the text, causing the URL text to appear off-center. Add a custom NSSearchFieldCell subclass that adjusts the drawing, editing, and selection frames to vertically center the text within the cell bounds.

According to my research there's no other way to vertically center text. And this seems more straightforward to me than adjusting the height of the cell based on the text size.

Before:  

image

After:

image


gmta:
Welcome @noahhaasis !

> causing the URL text to appear off-center

I cannot reproduce this issue in macOS Sequoia 15.7.4, what do you use?

noahhaasis:
> Welcome @noahhaasis !
> 
> > causing the URL text to appear off-center
> 
> I cannot reproduce this issue in macOS Sequoia 15.7.4, what do you use?

Thank you! Sorry for the late response. I'm on Tahoe 26.3.1.

PR #8458, opened 2026-03-16
LibWeb/HTML: Teach MessageEvent to work with origins, add support for cross origin WPT tests

shannonbooth:
Which fixes the Origin.from() handling for MessageEvent's. Corresponds to https://github.com/whatwg/html/commit/748b068c7c

However, perhaps more importantly, this also adds partial support for importing WPT tests that test cross origin behaviour, hopefully making it much easier for us to test origin related functionality in CI.


shannonbooth:
I reworked the approach a bit, and now I think the changes to the test server are much simpler

Issue #8465, opened 2026-03-17
Selecting first line in `<h1>` does not work


gmta:
On [this page](https://tweakers.net/nieuws/245800/telenet-klanten-zagen-door-fout-gegevens-van-andere-klanten-in-mytelenet-portal.html), selecting the first line of the `<h1>` above the article no longer works:

<img width="645" height="101" alt="Image" src="https://github.com/user-attachments/assets/49fe3810-78aa-461f-bd1b-86c1f23b42bc" />

Before 5bfc4a3c41b880493180de1fb6e1e756a7d149ca, it did work:

<img width="644" height="101" alt="Image" src="https://github.com/user-attachments/assets/4a9783f2-7a36-4eef-a081-6f4acbaad8fb" />

CC @kalenikaliaksandr

PR #8466, opened 2026-03-17
LibWeb: Invalidate paint caches of inline ancestors on selection change

aguiarcode:
This change resolves issue #8465.

After commit `5bfc4a3` (which introduced display list command caching per paintable), selecting text in an `<h1>` that wraps its content in an inline element (like `<span>`) no longer shows the selection highlight.

Now, with my fix, instead of only invalidating the containing block's cache, it walks up through all ancestor `PaintableBox` nodes up to and including the containing block, invalidating each one. This ensures inline-level `PaintableWithLines` (like the `<span>`'s per-line paintable) also get their caches cleared when selection changes.

### DIFF
--- a/Libraries/LibWeb/Painting/Paintable.cpp
+++ b/Libraries/LibWeb/Painting/Paintable.cpp
@@ -349,10 +349,17 @@ void Paintable::set_selection_state(SelectionState state)
     if (m_selection_state == state)
         return;
     m_selection_state = state;
-    if (auto* box = as_if<PaintableBox>(this))
+    if (auto* box = as_if<PaintableBox>(this)) {
         box->invalidate_paint_cache();
-    else if (auto* containing_block = this->containing_block())
+    } else if (auto* containing_block = this->containing_block()) {
         containing_block->invalidate_paint_cache();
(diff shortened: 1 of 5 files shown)


gmta:
Welcome, and much appreciated! Can you add a regression test please? :)

aguiarcode:
Thank you! I've added a regression test,it selects the text inside an inline element within an `<h1>` and compares it against the expected selection highlight.

Libraries/LibWeb/Painting/Paintable.cpp
```diff
@@ -351,8 +351,13 @@ void Paintable::set_selection_state(SelectionState state)
     m_selection_state = state;
     if (auto* box = as_if<PaintableBox>(this))
         box->invalidate_paint_cache();
-    else if (auto* containing_block = this->containing_block())
-        containing_block->invalidate_paint_cache();
+    auto const* cb = containing_block();
```
tcl3: Please use full words for variable names: `auto const* containing_block = this->containing_block()`

aguiarcode: Updated, thanks for the guidance!

tcl3:
Looks good, just one thing: when responding to review feedback you should amend your existing commit rather than adding a new one. In this case, I think it would be best to squash all your changes into a single commit.

PR #8460, opened 2026-03-16
LibGfx+LibWeb: Move Skia backend context to process level singleton

tcl3:
The skia backed context is stored on the traversable navigable, which is null for detached documents. We now fall back to CPU rendering when creating bitmaps for a canvas on a detached document, avoiding this crash.

Fixes #8493
Fixes #8498

PR #8468, opened 2026-03-17
LibWeb: Fix canvas and WebGL crashes on detached documents

jdahlin:
Guard navigable() and font_cascade_list() against null when the canvas belongs to a detached document or removed iframe.


gmta:
This is another approach to PR #8460, where we've decided on a singleton for now.

jdahlin:
> This is another approach to PR #8460, where we've decided on a singleton for now.

oh! Maybe the tests are useful still though.

PR #8470, opened 2026-03-17
LibWeb: Fix crashes during navigable teardown and iframe removal

jdahlin:
## Summary
- Don't crash in `get_session_history_entries` when a navigable has been removed from its traversable's session history during teardown
- Don't crash in `perform_navigation_params_fetch` when the parent navigable's document is gone
- Don't crash in `Navigable::reload` when there is no active window
- Don't crash in `is_form_associated_custom_element` when the custom element definition is missing after `adoptNode`
- Don't crash in `inform_the_navigation_api_about_aborting_navigation` when the active window becomes null between task queuing and execution

## Test plan
- [x] Added crash tests for each fix
- [x] Full LibWeb test suite passes (6321 pass, 0 fail, 0 crash)

### DIFF
--- a/Libraries/LibWeb/HTML/Navigable.cpp
+++ b/Libraries/LibWeb/HTML/Navigable.cpp
@@ -2717,10 +2717,11 @@ void Navigable::inform_the_navigation_api_about_aborting_navigation()
         return;
 
     queue_global_task(Task::Source::NavigationAndTraversal, *active_window(), GC::create_function(heap(), [this] {
+        if (!active_window())
+            return;
         HTML::TemporaryExecutionContext execution_context { active_window()->realm() };
 
         // 2. Let navigation be navigable's active window's navigation API.
-        VERIFY(active_window());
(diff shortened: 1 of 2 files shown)


Libraries/LibWeb/HTML/Navigable.cpp
```diff
@@ -705,7 +705,9 @@ Vector<GC::Ref<SessionHistoryEntry>>& Navigable::get_session_history_entries() c
         }
     }
 
-    VERIFY_NOT_REACHED();
+    // Navigable was not found in its traversable's session history (e.g. removed during teardown).
```
Psychpsyo, line 708: The `VERIFY_NOT_REACHED()` here seems to have corresponded to step 7 of https://html.spec.whatwg.org/multipage/browsing-the-web.html#getting-session-history-entries

Is this a bug with the spec or is this hiding some other problem in our implementation?
Either way, it should have an `// AD-HOC:` comment with more details as to why deviating from the spec is necessary here.

PR #8471, opened 2026-03-17
LibWeb: Fix CSS style computation crashes on detached documents

jdahlin:
Replace VERIFY assertions with fallbacks in Length::for_element() when computed_properties or root element is null. Guard inheritance_parent->computed_properties() in StyleComputer.


Calme1709:
I don't believe that these changes were required - at least the test provided was fixed by #8503

PR #8472, opened 2026-03-17
LibWeb: Fix crashes in focus, scroll, storage, and view transitions

jdahlin:
Guard navigable() and browsing_context() against null in run_unfocusing_steps, scroll_into_view, obtain_a_storage_bottle_map, and view transition setup/update. Relax VERIFY(paintable()) in
Document display list recording to a null check.

### DIFF
--- a/Libraries/LibWeb/HTML/NavigableContainer.cpp
+++ b/Libraries/LibWeb/HTML/NavigableContainer.cpp
@@ -298,6 +298,10 @@ void NavigableContainer::destroy_the_child_navigable()
         return;
     navigable->set_has_been_destroyed();
 
+    // AD-HOC: A detached iframe's document is hidden.
+    // https://github.com/w3c/csswg-drafts/issues/10264
+    navigable->active_document()->update_the_visibility_state(HTML::VisibilityState::Hidden);
+
     // AD-HOC: Clear the navigable's "is delaying load events" flag.
     //         This removes the DocumentLoadEventDelayer on the parent document that was
     //         created when the navigable started loading (navigate algorithm step 15).
--- a/Libraries/LibWeb/HTML/NavigableContainer.cpp
+++ b/Libraries/LibWeb/HTML/NavigableContainer.cpp
@@ -113,7 +113,9 @@ WebIDL::ExceptionOr<void> NavigableContainer::create_new_child_navigable(GC::Ptr
     auto traversable = parent_navigable->traversable_navigable();
 
     // AD-HOC: Let the initial about:blank document inherit the system visibility state from traversable.
-    document->update_the_visibility_state(traversable->system_visibility_state());
+    //         Set directly rather than via update_the_visibility_state() to avoid queuing
+    //         view transition page-visibility change steps for the initial state.
+    document->set_visibility_state(traversable->system_visibility_state());
 
     // 12. Append the following session history traversal steps to traversable:
     traversable->append_session_history_traversal_steps(GC::create_function(heap(), [traversable, navigable, parent_navigable, history_entry, after_session_history_update] {
--- a/Tests/LibWeb/Crash/HTML/element-blur-in-detached-doc.html
+++ b/Tests/LibWeb/Crash/HTML/element-blur-in-detached-doc.html
@@ -1,0 +1,13 @@
+<!DOCTYPE html>
+<!-- Test: blur() and focus() on elements in a detached document should not crash -->
+<body>
+<script>
+var doc = document.implementation.createHTMLDocument('');
+['div','input','textarea','select','button','a','canvas','video'].forEach(function(tag) {
+    var el = doc.createElement(tag);
+    doc.body.appendChild(el);
+    el.blur();
+    el.focus();
+});
+</script>
+</body>
--- a/Libraries/LibWeb/DOM/Document.cpp
+++ b/Libraries/LibWeb/DOM/Document.cpp
@@ -7275,7 +7275,8 @@ RefPtr<Painting::DisplayList> Document::record_display_list(HTML::PaintConfig co
         return m_cached_display_list;
 
     update_paint_and_hit_testing_properties_if_needed();
-    VERIFY(paintable());
+    if (!paintable())
+        return {};
 
     auto display_list = Painting::DisplayList::create(*paintable()->visual_context_tree());
     Painting::DisplayListRecorder display_list_recorder(display_list);
--- a/Libraries/LibWeb/DOM/Document.h
+++ b/Libraries/LibWeb/DOM/Document.h
@@ -576,6 +576,7 @@ public:
     bool hidden() const;
     StringView visibility_state() const;
     HTML::VisibilityState visibility_state_value() const { return m_visibility_state; }
+    void set_visibility_state(HTML::VisibilityState state) { m_visibility_state = state; }
 
     // https://html.spec.whatwg.org/multipage/interaction.html#update-the-visibility-state
     void update_the_visibility_state(HTML::VisibilityState);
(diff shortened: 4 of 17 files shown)


Tests/LibWeb/Crash/HTML/element-blur-in-detached-doc.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<!-- Test: blur() and focus() on elements in a detached document should not crash -->
+<body>
+<script>
+var doc = document.implementation.createHTMLDocument('');
```
ADKaster, line 5: could you split this PR into one commit per test? It's not obvious which fixes correspond to which test case.

I'm worried that some (most?) of these null checks will either leave the document in an invalid state or paper over whatever issue resulted in the null state in the first place

for the fixes that update spec-annotated steps, the issue is more dire. The fact that the spec doesn't say to check for xyz, but we are, means that either we have a bug somewhere related, or there is a spec bug to file. changing spec steps to add custom logic without justification is quite discouraged.

Psychpsyo:
The view transition changes seem to go against what the spec tells us to do.
I'd assume the transition should either run, or be skipped in a spec-approved fashion, since skipping vs not skipping might be observable from script.

Edit: https://github.com/w3c/csswg-drafts/issues/10264 says that VTs in detached frames should be skipped due to the document being hidden, which would be step 4 of [startViewTransition](https://www.w3.org/TR/css-view-transitions-1/#ViewTransition-prepare)

Psychpsyo:
Also, for the VT test: The `startViewTransition()` function can never throw an error, so the `try ... catch` you added is unnecessary.

Tests/LibWeb/Crash/HTML/element-blur-in-detached-doc.html
jdahlin: > could you split this PR into one commit per test? It's not obvious which fixes correspond to which test case.
> 

done!

> I'm worried that some (most?) of these null checks will either leave the document in an invalid state or paper over whatever issue resulted in the null state in the first place
> 
> for the fixes that update spec-annotated steps, the issue is more dire. The fact that the spec doesn't say to check for xyz, but we are, means that either we have a bug somewhere related, or there is a spec bug to file. changing spec steps to add custom logic without justification is quite discouraged.

I'm not sure what to do here, is invalid state not slightly better than crashing? Could adding a FIXME and pointing to a new or existing spec bug help?

jdahlin:
> Also, for the VT test: The `startViewTransition()` function can never throw an error, so the `try ... catch` you added is unnecessary.

This has been fixed

jdahlin:
> The view transition changes seem to go against what the spec tells us to do. I'd assume the transition should either run, or be skipped in a spec-approved fashion, since skipping vs not skipping might be observable from script.
> 
> Edit: [w3c/csswg-drafts#10264](https://github.com/w3c/csswg-drafts/issues/10264) says that VTs in detached frames should be skipped due to the document being hidden, which would be step 4 of [startViewTransition](https://www.w3.org/TR/css-view-transitions-1/#ViewTransition-prepare)

I mostly did random fuzzying to find these crashes. Not sure what to do with regarding to the spec behavior.

Psychpsyo:
> I mostly did random fuzzying to find these crashes. Not sure what to do with regarding to the spec behavior.

In my opinion, figuring out why the iframe's document's visibility state is not hidden in ladybird during step 4 of startViewTransition. Cause the linked spec issue suggests that it should be and that the current spec is fine. So we're likely doing something wrong related to the document visibility state and that will probably cause other problems elsewhere if not fixed properly.

Not crashing is great, but if it's done by introducing bugs that obscure the core problem and will have to be removed again when that problem is fixed, I'm not sure if it's worth it. Especially in a pre-alpha stage where crashes don't affect any users.

Libraries/LibWeb/HTML/NavigableContainer.cpp
```diff
@@ -298,6 +298,12 @@ void NavigableContainer::destroy_the_child_navigable()
         return;
     navigable->set_has_been_destroyed();
 
+    // Update the document's visibility state to "hidden" now that it no longer has an active navigable.
+    // Per the Page Visibility spec, a document that is not the active document of a browsing context
```
Psychpsyo: This should link to and just cite the spec directly instead of describing what the spec allegedly says.

Libraries/LibWeb/ViewTransition/ViewTransition.cpp
```diff
@@ -168,6 +168,10 @@ void ViewTransition::activate_view_transition()
     // 3. If transition’s initial snapshot containing block size is not equal to the snapshot containing block size, then
     //    skip transition with an "InvalidStateError" DOMException in transition’s relevant Realm, and return.
     auto navigable = document.navigable();
+    // FIXME: Per spec, this should not be null at this point. Documents that navigate away (e.g.
+    //        iframe.src change) during a transition may lose their navigable before activation.
+    //        The root fix is to ensure visibility state is correctly tracked across navigations.
+    //        See: https://github.com/w3c/csswg-drafts/issues/10264
```
Psychpsyo: This link seems to have nothing to do with the issue that is being described here.
Same goes for all the other, similar comments.

Psychpsyo: I also wonder if it wouldn't be more economical to do the root fix once, instead of adding workarounds in a bunch of places to avoid doing the root fix.

jdahlin: @Psychpsyo Is the latest two commits more what you had in mind?

Psychpsyo: Somewhat. But just linking to https://github.com/w3c/csswg-drafts/issues/10264 whenever you AD-HOC set documents to hidden doesn't really make any sense. When something is done AD-HOC, it's not following the specification. *This is very bad*, so it should be accompanied by an explanation of why there's a good reason that makes it necessary to do so and what needs to happen to get that AD-HOC code removed.

Just linking to a Github issue about view transitions where somewhere in the middle of the conversation someone says "a detached iframe is always `hidden`" doesn't do either of those.
A random claim on Github is also not an authoritative source for what the spec is. (unless it's an actual working group resolution or general consensus on some upcoming spec change)
Our iframes not being hidden in this situation hints at the fact that we either mis-implemented the spec somewhere or that the claim from that Github issue is wrong. (some browsing of the HTML spec around detached iframes makes me personally believe that it might be the latter, although I don't know that for sure)
If we mis-implemented the spec, that should probably get fixed instead.
If the comment on the Github issue is wrong about the spec, then there's two options:
a) We are doing something else wrong and setting the iframe to hidden is once again introducing a bug that will need to be replaced with an actual fix later.
Or b) The spec is wrong and there should be a spec bug filed about that, so that we can then reference it in the AD-HOC comment as justification for why doing it ad-hoc is necessary.

Psychpsyo: Looked into this some more and I think the removed iframe case is a spec bug, so I've filed https://github.com/whatwg/html/issues/12288 about that.
I'd guess the right place to set the content document to hidden will be somewhere closer to [destroying a document](https://html.spec.whatwg.org/multipage/document-lifecycle.html#destroy-a-document) rather than in the iframe's removal steps.

In ladybird, that's in `Document::destroy()`.

Psychpsyo: As for the navigation related document visibility stuff, I think the actual bug for that is in `Document::unload_a_document_and_its_descendants()`, which takes a `new_document` for some reason and unloads that. (Unloading is what sets the visibility state.) This doesn't match the spec and we seem to be passing either nothing or unrelated documents at each call site.

PR #8473, opened 2026-03-17
LibWeb: Fix canvas text crashes on detached documents

jdahlin:
Guard font_cascade_list() and document window against null in
canvas text operations on documents without a navigable.

### DIFF
--- a/Libraries/LibWeb/HTML/Canvas/CanvasTextDrawingStyles.cpp
+++ b/Libraries/LibWeb/HTML/Canvas/CanvasTextDrawingStyles.cpp
@@ -112,7 +112,7 @@ void CanvasTextDrawingStyles<IncludingClass, CanvasType>::set_font(StringView fo
             // NOTE: The initial value here is non-standard as the default font is "10px sans-serif"
             // FIXME: Investigate whether this is the correct resolution context (i.e. whether we should instead use
             //        a font-size of 10px) for OffscreenCanvas
-            auto length_resolution_context = CSS::Length::ResolutionContext::for_window(*document->window());
+            auto length_resolution_context = CSS::Length::ResolutionContext::for_document(*document);
             Optional<DOM::AbstractElement> abstract_element;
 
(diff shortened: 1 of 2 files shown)


Libraries/LibWeb/HTML/Canvas/CanvasTextDrawingStyles.cpp
```diff
@@ -112,6 +112,8 @@ void CanvasTextDrawingStyles<IncludingClass, CanvasType>::set_font(StringView fo
             // NOTE: The initial value here is non-standard as the default font is "10px sans-serif"
             // FIXME: Investigate whether this is the correct resolution context (i.e. whether we should instead use
             //        a font-size of 10px) for OffscreenCanvas
+            if (!document->window())
+                return {};
             auto length_resolution_context = CSS::Length::ResolutionContext::for_window(*document->window());
```
awesomekling: Would using `ResolutionContext::for_document()` instead make more sense here?

(What is even really the difference between `for_document` and `for_window` :thonk:)

gmta: FYI, there is a comment for that:

```cpp
        // FIXME: Anywhere we use this we probably want to use `for_document` instead since this uses the window's
        //        viewport rather than the documents which can differ e.g. with iframes.
        [[nodiscard]] static ResolutionContext for_window(HTML::Window const&);
```

jdahlin: It does indeed, thanks!

Libraries/LibWeb/HTML/CanvasRenderingContext2D.cpp
```diff
@@ -240,6 +240,8 @@ void CanvasRenderingContext2D::allocate_painting_surface_if_needed()
 {
     if (m_surface || m_size.is_empty())
         return;
+    if (!canvas_element().navigable())
+        return;
```
tcl3: This change doesn't appear to prevent a crash in either of the added tests.

PR #8476, opened 2026-03-17
 LibJS: Generate FFI header using cbindgen instead of hand-rolling 

ADKaster:
Replace the BytecodeFactory header with cbindgen.

This will help ensure that types and enums and constants are kept in
sync between the C++ and Rust code. It's also a step in exporting more
Rust enums directly rather than relying on magic constants for
switch statements.

The FFI functions are now all placed in the JS::FFI namespace, which
is the cause for all the churn in the scripting parts of LibJS and
LibWeb.


Libraries/LibJS/RustIntegration.cpp
```diff
@@ -889,6 +899,8 @@ void free_function_ast(void* ast)
 
 // --- FFI factory functions (called by Rust to create C++ objects) ---
 
+namespace JS::FFI {
```
ADKaster, line 902: the fact that this is repeated so many times in this patch screams "split me into many files" to me, but that's definitely a follow-up.

Libraries/LibJS/Rust/build.rs
```diff
@@ -670,8 +670,24 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
     println!("cargo:rerun-if-changed=build.rs");
+    println!("cargo:rerun-if-changed=cbindgen.toml");
 
     let out_dir = PathBuf::from(env::var("OUT_DIR")?);
+
+    let ffi_out_dir = env::var("FFI_OUTPUT_DIR")
+        .map(PathBuf::from)
+        .unwrap_or_else(|_| out_dir.clone());
+
+    cbindgen::generate(manifest_dir).map_or_else(
```
ADKaster, line 681: the alternative to this would be to add a build-time manual dep to `cargo install cbindgen` on every developer, and create a CMake function to call `cbindgen --stuff --crate-dir LibJS/Rust`. That would make it so that there's a way to generate the file manually (e.g. ninja -C Build/release generate_LibJS_FFI_h). without running a cargo build of libjs_rust.

I don't think we have any tools that would want a "gimme all generated files but not a full build" anymore though (Sonar used to do this iirc).

nico:
Shouldn't this delete something that's now generated? As far as I can tell it's just additions and replacements, but no deletions? Or did I miss it?

ADKaster:
We deleted LibJS/ByteCodeFactory.h https://github.com/LadybirdBrowser/ladybird/pull/8476/changes#diff-9e1f39a2a5f452ec594cf560ea61983f8c2d3a6dade913835c9f4a57409910c1

But because I added cbindgen to LibJS/Rust/Cargo.toml, we get hit twice with the deps. Once in //Cargo.lock, and once in //Meta/CMake/flatpak/cargo-sources.json.

So, -500 from the header, +1k from the deps. I have a branch I'm about to push to show what it will look like for LibUnicode's use of icu4x for calendar adjustments that looks more reasonable.

awesomekling:
cool :)

PR #8480, opened 2026-03-17
Tests: Simplify input-value-does-not-force-layout not to depend on perf

Zaggy1024:
This prevents a potential flake on CI when sanitizer overhead may cause
the test to fail due to the operations taking unexpectedly long.

See: https://github.com/LadybirdBrowser/ladybird/actions/runs/23219238558/job/67487741431?pr=8440

### DIFF
--- a/Tests/LibWeb/Text/expected/input-value-does-not-force-layout.txt
+++ b/Tests/LibWeb/Text/expected/input-value-does-not-force-layout.txt
@@ -1,1 +1,1 @@
-PASS
+layout is not up to date
--- a/Tests/LibWeb/Text/input/input-value-does-not-force-layout.html
+++ b/Tests/LibWeb/Text/input/input-value-does-not-force-layout.html
@@ -1,37 +1,14 @@
 <!DOCTYPE html>
-<style>
-    .row {
-        display: flex;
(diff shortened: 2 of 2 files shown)


Zaggy1024:
CC @gmta

Zaggy1024:
Was hoping I could have the link in the actual commit message, oh well.

Hopefully this new version of the test still fulfills the intent, since tests that depend on a benchmark seem problematic.

Calme1709:
Seems to be solving the same problem as #8388 FYI

Zaggy1024:
Ah, I hadn't seen that one, I suppose it's up to Jelle which strikes his fancy more :^)

gmta:
> Ah, I hadn't seen that one, I suppose it's up to Jelle which strikes his fancy more :^)

Well, the test taking much too long was the original issue. The "Performance test" part that you remove here is an actual regression test that used to take 40+ seconds on my machine. So I'd rather keep that aspect in :)

Zaggy1024:
Fair enough. I'm a little concerned about the convention of adding benchmarks as tests given that sanitizers already multiply the run time of CI by a lot, but as long as it doesn't flake I'm happy.

PR #8482, opened 2026-03-18
CI: Try running test262 on arm64 just to see results

trflynn89:
Our own results differ from test262.fyi, so just seeing if this is due to test262.fyi using an arm64 linux runner.

### DIFF
--- a/.github/workflows/libjs-test262.yml
+++ b/.github/workflows/libjs-test262.yml
@@ -13,7 +13,7 @@ env:
 
 jobs:
   run_and_update_results:
-    runs-on: 'blacksmith-16vcpu-ubuntu-2404'
+    runs-on: 'blacksmith-16vcpu-ubuntu-2404-arm'
     if: github.repository == 'LadybirdBrowser/ladybird'
 
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/libjs-test262.yml
```diff
@@ -62,7 +62,7 @@ jobs:
         uses: ./.github/actions/cache-restore
         id: 'cache-restore'
         with:
-          runner_labels: '["blacksmith-16vcpu-ubuntu-2404"]'
+          runner_labels: '["blacksmith-16vcpu-ubuntu-2404-arm"]'
           os: 'Linux'
           arch: 'x86_64'
```
ADKaster, line 67: gotta set arch at least here to dl the right wasm-tools bins. or remove wasm spec tests from cmake invocation.

trflynn89: Yeah that was a whoopsie. Strange that that didn't fail the build, despite a bazillion `OSError: [Errno 8] Exec format error: 'wasm-tools'` errors.

I got what I wanted from this experiment anyhow, which was no diff for test262:
https://github.com/LadybirdBrowser/ladybird/actions/runs/23224264573/job/67503511530?pr=8482#step:15:19

PR #8488, opened 2026-03-18
AK+WebContent: Don't print duplicate stack traces after failed assertion

InvalidUsernameException:
e38adbc31b5fdfaf7e1f12b9d255d90f8b690ba1 introduced a signal handler for WebContent. When an assertion fails, we print a stack trace, then raise a trap, which triggers the signal handler to print the same stack trace again.

Instead, only print a stack trace in the signal handler if it was triggered by something other than a failing assertion. We keep the printing in the assertion code because there we know we can skip some frames we don't care about.

### DIFF
--- a/AK/Assertions.cpp
+++ b/AK/Assertions.cpp
@@ -133,6 +133,7 @@ NEVER_INLINE void ak_trap(void)
     // Skip 3 frames to get to caller. That is dump_backtrace, ak_trap, and ak_verification_failed.
     dump_backtrace(3, 100);
 #endif
+    ak_verification_has_been_triggered = true;
     __builtin_trap();
 }
 
--- a/AK/Assertions.h
+++ b/AK/Assertions.h
@@ -6,6 +6,10 @@
 
 #pragma once
 
+#include <csignal>
+
+inline sig_atomic_t volatile ak_verification_has_been_triggered = false;
+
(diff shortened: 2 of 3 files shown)


jdahlin:
What about multiple different stack traces at different times, is that never happening?

InvalidUsernameException:
Is there a specific scenario you are thinking about?

If two threads are `VERIFY`ing simultaneously, then I believe that should be fine with the combination of guaranteed-atomic type and `volatile`. That being said, intricacies of C++ are not my strong suit, so it might be there is something I'm not aware of.

If one thread is `VERIFY`ing and another one is segfaulting or similar, I guess it might be possible that we are losing the backtrace from the segfault. But that seems like a rather rare edge-case and sounds acceptable to me if it makes the vastly more common case more convenient.

Things happening at different times can't really be since the signal handler is terminating the process already anyway, right?

jdahlin:
No, I am probably more lost than you are here. I am mainly trying to understand the effects of this PR. Masking important stack traces would be very unfortunate, however I am not sure if it is actually possible in practice with your patch or only a theoretical.

Have you tried to deliberately raise exceptions at the same time from multiple threads, just to see what happens?

Are there other ways to avoid printing the same stack traces twice? What do the other multiprocess browsers do here?

gmta:

PR #8504, opened 2026-03-18
Fix Android Compilation

EricApostal:
Fixes compiling for Android.

This does not fix Android entirely, however, it resolves several issues in regards to compiling it. This is mostly to do with handling cross compilation targeting Android, and linking newer dependencies correctly. It also adds support for cross compiling to rust android, since the architecture names line up a bit differently from cmake.

Note: There is still a good amount of work to do here. While I did some work on UI/Android, I am facing several runtime errors still, and it's possible that through some of the AI-assisted porting of that specific project, I've put things in an incorrect state.

@ADKaster


UI/Android/src/main/cpp/JNIHelpers.cpp
```diff
@@ -5,14 +5,13 @@
...
-#include <AK/Utf16View.h>
 
 namespace Ladybird {
 
 jstring JavaEnvironment::jstring_from_ak_string(String const& str)
 {
-    auto as_utf16 = MUST(AK::utf8_to_utf16(str.code_points()));
-    return m_env->NewString(reinterpret_cast<jchar const*>(as_utf16.data.data()), as_utf16.data.size());
+    auto utf8 = str.to_byte_string();
+    return m_env->NewStringUTF(utf8.characters());
```
ADKaster: this is incorrect. AK::String is not [Java Modified UTF-8](https://docs.oracle.com/en/java/javase/11/docs/specs/jni/types.html#modified-utf-8-strings), but standards compliant UTF-8. 

The previous code intentionally transcoded the UTF-8 string to UTF-16 before passing it to Java. We should still have a way to do that.

UI/Android/src/main/cpp/ALooperEventLoopImplementation.h
```diff
@@ -88,4 +91,4 @@ class ALooperEventLoopImplementation : public Core::EventLoopImplementation {
     EventLoopThreadData* m_thread_data { nullptr };
 };
 
-}
+}
```
ADKaster: please configure your editor(s) to keep trailing newlines at EOF

UI/Android/src/main/cpp/ALooperEventLoopImplementation.cpp
```diff
@@ -118,6 +118,11 @@ void ALooperEventLoopManager::register_notifier(Core::Notifier& notifier)
 void ALooperEventLoopManager::unregister_notifier(Core::Notifier& notifier)
```
ADKaster, line 186: some changes happened in this area recently. a once-over of the entire event loop impl is in order.

Meta/CMake/vcpkg/overlay-ports/angle/cmake-buildsystem/CMakeLists.txt
```diff
@@ -92,6 +92,8 @@ if (APPLE)
     if (USE_OPENGL AND NOT angle_enable_eagl)
         set(angle_enable_cgl TRUE)
     endif()
+elseif (ANDROID)
+    set(is_android TRUE)
```
ADKaster: does this actually do anything interesting? does the upstream vcpkg angle port actually support android? if it doesn't, we'll have to go upstream from them, to WebKit's port (the original source for these cmake files) to get something working. Given that WebKit doesn't work on Android without going through WPE (https://github.com/Igalia/wpe-android), that might get interesting and require looking at the upstream angle gn files to get a sense for how you're supposed to wrangle the angle source files for that platform.

If even WPE-Android doesn't use Angle, we might need to look closer at how they handle WebGL for that port.

Libraries/LibJS/CMakeLists.txt
```diff
@@ -330,9 +330,9 @@ endif()
 # AsmInterpreter: generate platform-specific assembly from DSL
 # NB: Win64 uses a different ABI (shadow space, different arg registers)
 #     from SysV AMD64, and the x86_64 backend only supports SysV for now.
-if (NOT WIN32 AND ("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "x86_64" OR "${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "AMD64"))
+if (NOT CMAKE_CROSSCOMPILING AND NOT WIN32 AND ("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "x86_64" OR "${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "AMD64"))
     set(ASMINT_ARCH "x86_64")
-elseif ("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "aarch64" OR "${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "arm64")
+elseif (NOT CMAKE_CROSSCOMPILING AND ("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "aarch64" OR "${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "arm64"))
```
ADKaster, line 335: The entire ASMINT gen code generator needs a CMake overhaul to be a proper host tool in order for cross-compiling to work. It needs to be built for the host platform, during a LAGOM_TOOLS_ONLY build and installed into the host tools install dir.

There's also some missing add_custom_targets for those custom commands, but that can be handled separately.

We should never build it for android, so the link to the android log library below is a good indicator that something went wrong with this host-only tool.

Libraries/LibJS/Bytecode/AsmInterpreter/AsmInterpreter.cpp
```diff
@@ -78,7 +78,7 @@ namespace JS::Bytecode {
 
 // The asm interpreter is available on x86_64 (non-Windows) and AArch64.
 // Win64 uses a different ABI that the x86_64 backend doesn't support yet.
-#if ARCH(AARCH64)
+#if ARCH(AARCH64) && !defined(AK_OS_ANDROID)
```
ADKaster: android simulator exists, so this check should be a bit smarter. `#if defined(AK_OS_ANDROID) \n #define HAS_ASM_INTERPRETER 0 #elif ARCH(AARCH64) ...`

Given that android is linux under the hood, I'd be surprised if this was actually a hard stop though. was this just a cmake snafu with the host tool that lead to disabling it?

Libraries/LibCore/EventLoopImplementationUnix.cpp
```diff
@@ -409,7 +409,7 @@ void EventLoopManagerUnix::wait_for_events(EventLoopImplementation::PumpMode mod
 
 #ifdef AK_OS_ANDROID
             // FIXME: Make the check work under Android, perhaps use ALooper.
-            ThreadEventQueue::current().post_event(notifier, Core::Event::Type::NotifierActivation);
+            ThreadEventQueue::current().post_event(&notifier, Core::Event::Type::NotifierActivation);
```
ADKaster, line 412: this is suspicious. I suspect some of the issues you saw with a pink web view are related to not handling this properly.

UI/Android/src/main/cpp/RequestServerService.cpp
```diff
@@ -13,26 +13,39 @@
...
-extern ByteString g_default_certificate_path;
+OwnPtr<ResourceSubstitutionMap> g_resource_substitution_map;
 
 }
 
 ErrorOr<int> service_main(int ipc_socket)
 {
 
-    RequestServer::g_default_certificate_path = ByteString::formatted("{}/cacert.pem", WebView::s_ladybird_resource_root);
+    RequestServer::set_default_certificate_path(ByteString::formatted("{}/cacert.pem", WebView::s_ladybird_resource_root));
```
ADKaster, line 33: do we need to set a cert path still? This was set explicitly back in the day before we moved to using libcurl. I'm not even sure if we still install the cacert.pem into the bundle.

UI/Android/src/main/cpp/JNIHelpers.cpp
ADKaster: Something like
```c++
auto as_utf16 = AK::Utf16String::from_utf8(str);
auto view = as_utf16.utf16_view();
if (view.has_ascii_storage())
    return m_env->NewStringUTF(view.ascii_span().data());
return m_env->NewString(reinterpret_cast<jchar const*>(view.utf16_span().data()), view.length_in_code_units());
```
though i'd not be confident merging such a thing without @trflynn89 looking at it.

Libraries/LibJS/Bytecode/AsmInterpreter/AsmInterpreter.cpp
EricApostal: Yeah there isn't a ton in the way, it was more for simplicity since the tooling was building this as a binary for the target and not the host. I only did 2 passes on this (one on x86 and one on arm), and it was just easier until I could get some sort of basic functionality, since at the time ASM had *just* come out and I didn't want to add an extra variable with generating ASM on the fly on different platforms / architectures (even if all is LP64).

Libraries/LibJS/CMakeLists.txt
EricApostal: I meant to remove that, I misunderstood what it did entirely and was attempting to target it as some static binary. Will update to be more reasonable and remove that.

UI/Android/src/main/cpp/RequestServerService.cpp
EricApostal: I have absolutely no idea to be honest. There is no such file in the bundled ladybird-assets.zip.

EricApostal:
I really appreciate the detailed feedback there! I left some comments, and I'll go in tomorrow to push some fixes through, and get a bit more understanding of some of the bits you've mentioned. Again I appreciate the thorough review here!

Meta/CMake/vcpkg/overlay-ports/angle/cmake-buildsystem/CMakeLists.txt
ADKaster: actually poking around, we should not build ANGLE on Android. It's a system library since Android 10 or so. We should be able to either use or write a CMake find module to pull it from the ndk or some other auto-configured (by the ndk's toolchain file) sysroot location. There might be some xml config as well to tell the system loader that we want to use Angle libraries directly to get access to the angle-specific symbols rather than just the GLES symbols.

UI/Android/src/main/cpp/JNIHelpers.cpp
trflynn89: Yep that suggestion looks right

Meta/CMake/audio.cmake
```diff
@@ -1,7 +1,10 @@
 include_guard()
 
 # Audio backend -- how we output audio to the speakers.
-if (APPLE AND NOT IOS)
+if (ANDROID)
+    set(LADYBIRD_AUDIO_BACKEND "AAUDIO")
```
Zaggy1024, line 5: We should handle not setting this just fine, probably best to leave this out until there's an implementation for Android again.

Meta/CMake/vcpkg/overlay-ports/angle/cmake-buildsystem/CMakeLists.txt
EricApostal: @ADKaster Android itself has Angle but as I understand it we can't just include it. It's _in_ android strictly speaking but I thought the pipeline was GLES -> Angle -> Vulkan. I got to that as a solution by checking modern WebKit which now sets `is_android` https://github.com/WebKit/WebKit/blob/e66c9d03b983c64b492ba08a6583762fae0c14e0/Source/ThirdParty/ANGLE/CMakeLists.txt#L11. I could definitely be wrong as I don't fully understand the WebKit pipeline here but as I understand I'm just implementing their approach directly.

UI/Android/src/main/cpp/JNIHelpers.cpp
EricApostal: I think this is also incorrect. Mine was too, but the new solution as well. `ascii_span` gives us UTF-8, so it seems like to me the best option is copying into a new `Vector<jchar>` then using `NewString` on that. Just noting this for the sake of discussion, and so my fix for it makes sense. Open to other solutions.

ADKaster: The ascii span gives ascii. Ascii characters are the same in utf-8 and the extended utf used by java, except for embedded the null characters. Worrying about such strings for UI purposes seems not very useful, as the result would be truncated strings.

If we bother to deoptomize the ascii path at this conversion point, I'd want to see a test case that proves it's worth the trouble.

EricApostal: Gotcha, I can revisit this. The provided implementation seemed to have weird side effects, almost like when you forget to add a null terminator (it would append random content to strings, or crash). My implementation fixes that, however I would be interested in a more correct way in any context.

ADKaster: Huh. That's interesting. In that case that extended utf might expect the two character encoding for end of string nulls as well. If that's the case, your idea to create a vector of jchars is likely necessary.

PR #8508, opened 2026-03-19
AK: Add SentinelOptional

gmta:
We specialize `Optional<T>` for value types that inherently support some kind of "empty" value or whose value range allow for a unlikely to be useful sentinel value that can mean "empty", instead of the boolean flag a regular Optional<T> needs to store. Because of padding, this often means saving 4 to 8 bytes per instance.

By extending the new `SentinelOptional<T, Traits>`, these specializations are significantly simplified to just having to define what the sentinel value is, and how to identify a sentinel value.

### DIFF
--- a/Libraries/LibWeb/HTML/Parser/HTMLToken.h
+++ b/Libraries/LibWeb/HTML/Parser/HTMLToken.h
@@ -375,3 +375,19 @@ private:
 };
 
 }
+
+namespace AK {
+
+template<>
+struct SentinelOptionalTraits<Web::HTML::HTMLToken> {
+    static Web::HTML::HTMLToken sentinel_value() { return {}; }
+    static bool is_sentinel(Web::HTML::HTMLToken const& value) { return value.type() == Web::HTML::HTMLToken::Type::Invalid; }
+};
+
+template<>
+class Optional<Web::HTML::HTMLToken> : public SentinelOptional<Web::HTML::HTMLToken> {
+public:
+    using SentinelOptional::SentinelOptional;
+};
+
+}


Hendiadyoin1:
Oh hey I also tried this in the past in https://github.com/LadybirdBrowser/ladybird/pull/5447 and https://github.com/SerenityOS/serenity/pull/25894

In case you might gleam something from that

gmta:
> Oh hey I also tried this in the past in #5447 and [SerenityOS/serenity#25894](https://github.com/SerenityOS/serenity/pull/25894)
> 
> In case you might gleam something from that

Doing that for `OptionalBase<T>` became too much of a hassle because of our `Optional<T&>` support, so I left that out and went for just the `Optional<T>` improvements.

gmta:
> naming hint of general nature:

I disagree with "monoid" being a better choice than "sentinel", the latter of which has common usage as an in-band special value - and iteration is one example of where that could apply.

I slightly agree with you that "sentinel" is not covering the exact use case. But we've been using that term before in this context within the project, and finding a better name sounds like a nice bike shedding discussion that you could have on our Discord!

konradekk:
> I slightly agree with you that "sentinel" is not covering the exact use case.

i don't want to test your patience, will just put one more argument for monoids (and leave it be). 🫣

> I disagree with "monoid" being a better choice than "sentinel", the latter of which has common usage as an in-band special value - and iteration is one example of where that could apply.

sentinels are special values, can be one or more (e.g. first or last element, boundary elements with different characteristics, but all helping to avoid different cases in core logic).

monoid highlights that there's distinguished single value that does not change the content when combined (zero, empty string etc.). invalid values in enums are somewhat different, seemingly absorbing (like multiplying by zero, nans or rejected promises) – it seems it has slightly different semantics, even if implementation is (virtually) the same.

excuses, once again, for pushing my luck here! 🙇🏻‍♂️

gmta:
@konradekk a monoid requires an identity element which is not what this is about. It also implies associative binary operations that are also irrelevant here. Let's not continue this discussion here, feel free to join Discord though.

PR #8509, opened 2026-03-19
LibWeb: Start fetching CSS image resources before document load event

### LINKED ISSUE #3448: headless-browser: Ref-test screenshots are taken before CSS-linked resources are loaded

gmta:
Both Chromium and Gecko delay the document's load event for CSS image resource requests (background-image, mask-image, etc). We now start fetching CSS image resources as soon as their stylesheet is associated with a document, rather than deferring until layout. This is done by collecting ImageStyleValues during stylesheet parsing and initiating their fetches when the stylesheet is added to the document.

Fixes #3448


Tests/LibWeb/Fixtures/http-test-server.py
```diff
@@ -97,6 +100,7 @@ def do_POST(self):
             echo.path = data.get("path", None)
             echo.status = data.get("status", None)
             echo.body = data.get("body", None)
+            echo.body_base64 = data.get("body_base64", None)
```
jdahlin: Alternatively `body` can be reused, instead check Content-Type and assume images/* (and other binary data?) are always encoded in base64.

gmta: Assuming a format seems like a footgun, I could introduce `body_encoding` that defaults to `raw` but could be set to `base64` instead.

Libraries/LibWeb/HTML/Parser/HTMLParser.cpp
```diff
@@ -384,6 +384,10 @@ void HTMLParser::the_end(GC::Ref<DOM::Document> document, GC::Ptr<HTMLParser> pa
         return document->scripts_to_execute_as_soon_as_possible().is_empty();
     }));
 
+    // AD-HOC: Initiate fetches for CSS image resources that were discovered during stylesheet parsing. This delays the
```
Psychpsyo: Maybe link to https://github.com/w3c/csswg-drafts/issues/1088?

Tests/LibWeb/Ref/input/background-size-near-zero-svg.html
```diff
@@ -2,6 +2,4 @@
 <link rel="match" href="../expected/background-size-near-zero-svg-ref.html">
 <div style="background-image: url(../data/50x50-green.svg);
             width: 100px; height: 100px; background-size: 0.2px 0.2px">
-</div>
-<!-- FIXME: Workaround to ensure CSS background-image is loaded before taking screenshot: https://github.com/LadybirdBrowser/ladybird/issues/3448 -->
-<img style="display: none" src="../data/50x50-green.svg">
+</div>
```
tcl3: This is now missing a newline at EOF

gmta:
@tcl3 

> As it stands, the load event isn't delayed for constructed stylesheets (added to document.adoptedStyleSheets). Calling load_any_resources() for pending image values in AdoptedStyleSheets.cpp should make this work.

Added! Thanks :)

Libraries/LibWeb/CSS/CSSStyleSheet.h
```diff
@@ -159,6 +163,8 @@ class WEB_API CSSStyleSheet final : public StyleSheet {
     Vector<GC::Ptr<FontLoader const>> m_associated_font_loaders;
 
     Vector<Subresource&> m_critical_subresources;
+
+    Vector<ImageStyleValue&> m_pending_image_values;
```
tcl3: Is it safe to hold these as bare references? It seems like you could end up with a dangling reference here if the stylesheet is modified, via `replace()/replaceSync()`, for example.

gmta: We could - changed it to WeakPtr

PR #8514, opened 2026-03-19
Replace Unix socket IPC transport with Mach ports on macOS

kalenikaliaksandr:
On macOS, use Mach port messaging instead of Unix domain sockets for
all IPC transport. This makes the transport capable of carrying Mach
port rights as message attachments, which is a prerequisite for sending
IOSurface handles over the main IPC channel (currently sent via a
separate out-of-band path). It also avoids the need for the FD
acknowledgement protocol that TransportSocket requires, since Mach port
right transfers are atomic in the kernel.

Three connection establishment patterns:

- Spawned helper processes (WebContent, RequestServer, etc.) use the
  existing MachPortServer: the child sends its task port with a reply
  port, and the parent responds with a pre-created port pair.

- Socket-bootstrapped connections (WebDriver, BrowserProcess) exchange
  Mach port names over the socket, then drop the socket.

- Pre-created pairs for IPC tests and in-message transport transfer.

Attachment on macOS now wraps a MachPort instead of a file descriptor,
converting between the two via fileport_makeport()/fileport_makefd().

The LibIPC socket transport tests are disabled on macOS since they are
socket-specific.

### DIFF
--- a/Libraries/LibIPC/TransportMachPort.cpp
+++ b/Libraries/LibIPC/TransportMachPort.cpp
@@ -7,9 +7,7 @@
 #include <AK/NonnullOwnPtr.h>
 #include <LibCore/MachPort.h>
 #include <LibCore/Notifier.h>
-#include <LibCore/Socket.h>
 #include <LibCore/System.h>
-#include <LibIPC/TransportBootstrapMach.h>
 #include <LibIPC/TransportMachPort.h>
 #include <LibThreading/Thread.h>
 
@@ -46,12 +44,6 @@ static Attachment attachment_from_descriptor(mach_msg_port_descriptor_t const& d
     default:
         VERIFY_NOT_REACHED();
     }
-}
-
-ErrorOr<NonnullOwnPtr<TransportMachPort>> TransportMachPort::from_socket(NonnullOwnPtr<Core::LocalSocket> socket)
-{
-    auto ports = TRY(bootstrap_transport_over_socket(*socket));
-    return make<TransportMachPort>(move(ports.receive_right), move(ports.send_right));
 }
 
 ErrorOr<TransportMachPort::Paired> TransportMachPort::create_paired()
--- a/Libraries/LibIPC/TransportBootstrapMach.cpp
+++ b/Libraries/LibIPC/TransportBootstrapMach.cpp
@@ -5,38 +5,17 @@
  */
 
 #include <AK/Assertions.h>
-#include <AK/ByteBuffer.h>
 #include <AK/ByteString.h>
+#include <AK/Optional.h>
 #include <LibCore/MachPort.h>
 #include <LibCore/Platform/MachMessageTypes.h>
 #include <LibCore/System.h>
 #include <LibIPC/TransportBootstrapMach.h>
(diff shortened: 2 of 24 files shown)


Libraries/LibCore/Platform/ProcessStatisticsMach.cpp
```diff
@@ -76,21 +76,19 @@ ErrorOr<void> update_process_statistics(ProcessStatistics& statistics)
     return {};
 }
 
-MachPort register_with_mach_server(ByteString const& server_name)
+MachServerRegistration register_with_mach_server(ByteString const& server_name)
```
ADKaster: now that this is about IPC instead of ProcessStatistics, this file probably isn't the correct home for this method.

Libraries/LibIPC/TransportMachPort.cpp
```diff
@@ -0,0 +1,447 @@
...
+            Core::MachPort::MessageRight::MoveSendOnce);
+    default:
+        VERIFY_NOT_REACHED();
+    }
+}
+
+ErrorOr<NonnullOwnPtr<TransportMachPort>> TransportMachPort::from_socket(NonnullOwnPtr<Core::LocalSocket> socket)
+{
+    auto our_recv = TRY(Core::MachPort::create_with_right(Core::MachPort::PortRight::Receive));
+    auto our_name = ByteString::formatted("org.ladybird.ipc.{}.{}", getpid(), our_recv.port());
```
ADKaster: is there anything sensitive about the mach port id? could another process on the local machine do anything interesting with it if it had a way to enumerate bootstrap server service names?

Libraries/LibIPC/TransportMachPort.cpp
```diff
@@ -0,0 +1,447 @@
...
+    TRY(our_recv.register_with_bootstrap_server(our_name));
```
ADKaster: Also the way that this handshake works feels uncomfortable. We're using the bootstrap server for multiple server ports that get registered at runtime. In theory, a properly set up macOS system would have all bootstrap services registered via static config. The only reason the original back channel for ProcessStatistics did this was so that child processes could always reach back to the UI process without jumping through hoops.

Libraries/LibCore/Platform/ProcessStatisticsMach.cpp
kalenikaliaksandr: sure, makes sense. I moved it into `LibIPC/TransportBootstrapMach.h` along with other mach transport bootstrap functions.

Libraries/LibIPC/TransportMachPort.cpp
kalenikaliaksandr: I added another commit that gets rid of bootstraping over socket, so it's not a concern anymore.

Libraries/LibIPC/TransportMachPort.cpp
kalenikaliaksandr: I added another commit that migrates all the places where we used this hack with promotion of socket connection to mach ports connection.

PR #8518, opened 2026-03-20
LibURL/Pattern: Ignore extra RegExp captures in match result

shannonbooth:
execResult may contain additional captures from nested groups in user-provided regexp parts, exceeding the number of URLPattern groups.

Fixes a crash in the updated WPT test.

See: https://github.com/whatwg/urlpattern/commit/203d435c32


shannonbooth:
Appears like webkit and Firefox had the exact same crash boog

PR #8528, opened 2026-03-20
LibWeb: Prevent race condition in Storage::key()

gmta:
Depending on the underlying implementation of the storage bottle, we have to do IPC calls to get either the size or keys. We crash on an out-of-bounds if the bottle changes size after the bounds check in step 1. Fix this by obtaining the keys immediately and using that for the bounds check.

Fixes a crash seen in CI.


shannonbooth:
I think the comment for step 2 should be above the point we create the variable keys

But yeah, whole lot of optimizations and races in this file to be made 😅 E.g this probably being a dedicated IPC call

gmta:
> I think the comment for step 2 should be above the point we create the variable keys

Done!

PR #8533, opened 2026-03-20
LibWeb/HTML: Mark some events as trusted

skyz1:
Some events were previously not tagged as trusted. This commit properly tags the following events so that we match the behaviour implemented by other browsers and expected by WPT.

- submit
- toggle
- beforetoggle
- beforeunload
- promiserejected

[This](https://dom.spec.whatwg.org/#concept-event-create) part of the spec explicitly tells us to set `isTrusted` to `true` when [firing an event](https://dom.spec.whatwg.org/#concept-event-fire) [here](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#form-submission-algorithm:submitevent).

### DIFF
--- a/Libraries/LibWeb/HTML/BeforeUnloadEvent.cpp
+++ b/Libraries/LibWeb/HTML/BeforeUnloadEvent.cpp
@@ -14,7 +14,9 @@ GC_DEFINE_ALLOCATOR(BeforeUnloadEvent);
 
 GC::Ref<BeforeUnloadEvent> BeforeUnloadEvent::create(JS::Realm& realm, FlyString const& event_name, DOM::EventInit const& event_init)
 {
-    return realm.create<BeforeUnloadEvent>(realm, event_name, event_init);
+    auto event = realm.create<BeforeUnloadEvent>(realm, event_name, event_init);
+    event->set_is_trusted(true);
+    return event;
 }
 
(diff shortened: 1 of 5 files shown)


Libraries/LibWeb/HTML/HTMLFormElement.cpp
```diff
@@ -178,6 +178,7 @@ WebIDL::ExceptionOr<void> HTMLFormElement::submit_form(GC::Ref<HTMLElement> subm
         auto submit_event = SubmitEvent::create(realm, EventNames::submit, event_init);
         submit_event->set_bubbles(true);
         submit_event->set_cancelable(true);
+        submit_event->set_is_trusted(true);
```
shannonbooth: Thanks! The pattern which we have started using for other types of events is to set isTrusted in the ::create method, but leave construct_impl as-is. I think we can make that type of change here too. See ErrorEvent::create as an example here

skyz1:
@gmta @shannonbooth I extended the scope of this PR a bit, as I found more events that were incorrectly not marked as trusted. Please check this PR again.

I also added an additional test that covers the additional event types. However, it seems like the `beforeunload` event is fired to late, as the event order deviates from other browsers. I will open an issue for that, as I think it is out of scope for this PR.

shannonbooth (approved):

PR #8536, opened 2026-03-20
LibJS: Make folded non-decimal prefix parsing UTF-8-safe

RubenKelevra:
Folded `StringToNumber()` and `StringToBigInt()` detected non-decimal prefixes by slicing the string at byte offset 2. On UTF-8 input this could split at a non-character boundary and panic.

To prevent this, we replace the byte-based split with ASCII prefix stripping and preserve rejection of empty suffixes such as `0x`, `0o`, and `0b` explicitly before parsing the remaining digits.

This makes non-decimal prefix folding UTF-8-safe and preserves the expected invalid-result behavior for empty prefixed literals.

_Found with fuzzing._

### Tests

- regression tests validating folded non-decimal coercion
- UTF-8 boundary panic regression tests for folded coercion

### DIFF
--- a/Libraries/LibJS/Rust/src/bytecode/codegen.rs
+++ b/Libraries/LibJS/Rust/src/bytecode/codegen.rs
@@ -8933,6 +8933,52 @@ fn try_constant_loosely_equals(lhs: &ConstantValue, rhs: &ConstantValue) -> Opti
     }
 }
 
+#[derive(Clone, Copy)]
+enum NonDecimalRadix {
+    Binary,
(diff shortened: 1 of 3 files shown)


RubenKelevra:
Obviously not stale.

shannonbooth:
Could you please squash the three commits together? Otherwise looks reasonable to me

RubenKelevra:
@shannonbooth sure. I folded them together and merged the commit messages. Thanks for the review!

PR #8538, opened 2026-03-20
LibWeb: Properly set visibility state for nested documents

Psychpsyo:
This cannot happen inside the Make Active algorithm, since that gets called during document creation, which commonly happens before the document's navigable is created.

This matches my HTML Spec PR on this topic: https://github.com/whatwg/html/pull/12294

I'm not sure if this should be merged right away since the spec PR hasn't been reviewed yet, but I want to have it open for review here since it shows that this can be implemented with (seemingly) no problems.

It should also take proper care of one of the crashes that #8472 is having to work around due to our prior AD-HOC behavior in this area.

### DIFF
--- a/Libraries/LibWeb/DOM/Document.cpp
+++ b/Libraries/LibWeb/DOM/Document.cpp
@@ -5098,33 +5098,34 @@ void Document::make_active()
         page().client().page_did_change_active_document_in_top_level_browsing_context(*this);
     }
 
-    // 3. Set document's visibility state to document's node navigable's traversable navigable's system visibility state.
-    auto navigable = this->navigable();
-    if (navigable) {
-        m_visibility_state = navigable->traversable_navigable()->system_visibility_state();
-
-        // AD-HOC: Record the initial viewport and visual viewport state so that if the viewport changes before the
-        //         first rendering update (e.g. in our fullscreen tests), change events are still fired.
-        if (!m_last_viewport_size.has_value()) {
-            m_last_viewport_size = viewport_rect().size().to_type<int>();
-            auto& current_visual_viewport = *visual_viewport();
-            m_last_visual_viewport_state = VisualViewportState { current_visual_viewport.scale(), { current_visual_viewport.width(), current_visual_viewport.height() } };
-        }
-    }
-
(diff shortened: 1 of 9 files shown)


Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -5027,27 +5031,14 @@ void Document::make_active()
         page().client().page_did_change_active_document_in_top_level_browsing_context(*this);
     }
 
-    // 3. Set document's visibility state to document's node navigable's traversable navigable's system visibility state.
-    auto navigable = this->navigable();
-    if (navigable) {
-        m_visibility_state = navigable->traversable_navigable()->system_visibility_state();
-
```
shannonbooth, line 5105: Without me investigating too deeply into the changes here so far, do we have a test for this change in behaviour somewhere we can import?

Psychpsyo: I've imported one now that verifies that the visibility state is set right. I don't think WPT has a test for not firing a `visibilitychange` event for the initial visibility state in new navigables and I don't think script runs early enough to catch that we were doing that anyways.

Psychpsyo:
My spec PR for this just landed, so this now brings us in line with the spec and removes some AD-HOC stuff we had before.

Tests/LibWeb/Text/expected/wpt-import/page-visibility/test_child_document.txt
```diff
@@ -0,0 +1,10 @@
+Harness status: Error
```
shannonbooth: The harness error is:

```
121248.537 WebContent(3362): Unhandled JavaScript exception: [TypeError] Cannot access property "visibilityState" on null object "doc"
121248.537 WebContent(3362):     at onload_test (file:///Users/shannonbooth/Developer/ladybird/Tests/LibWeb/Text/input/wpt-import/page-visibility/test_child_document.html:21:36)
```

Which appears to be due to cross origin access to the iframe in the test file when loading from a file (other browsers perform in the same way. Maybe we should instead configure under LoadFromHttpServer so that the test runs as expected? Otherwise LGTM (thanks for upstreaming as well!)

Psychpsyo: Good point, done that now!

PR #8540, opened 2026-03-21
LibWeb/Fetch: Implement CORS-preflight cache

shannonbooth:
This implements the CORs preflight cache and adds infrastructure to run python http wpt handlers to facilitate a wider set of WPT tests so that the preflight cache can be tested without coming up with more and more changes to the http server.


shannonbooth:
Updated this to also include support for vary.py for the cache-add WPT test also so those subtests also work as expected

Libraries/LibWeb/Fetch/Infrastructure/PreflightCache.cpp
```diff
@@ -0,0 +1,202 @@
...
+
+    create_a_new_cache_entry(request, max_age, method, OptionalNone());
+}
+
+// https://fetch.spec.whatwg.org/#concept-cache-match-header-name
+bool PreflightCache::is_a_header_name_cache_entry_match(Request const& request, ByteString const& header_name)
+{
+    evict_expired_entries();
+
+    // There is a header-name cache entry match for headerName using request when there is a cache entry in the user
```
Lubrsi, line 104: It seems this is out of date with the current spec, which now requires `*` to not match CORS non-wildcard headers. The section seems to have also changed to https://fetch.spec.whatwg.org/#concept-cache-match-header

shannonbooth: Whoops, thanks fixed, also improved WPT test to cover this. I ended up implementing this as specified even though other browsers don't implement for now even though we're likely to change this behaviour later. I ended up not being 100% sure on the behaviour so figure per spec is good to start with.

PR #8542, opened 2026-03-21
UI/AppKit: Avoid the window expanding when exiting fullscreen

Zaggy1024:
Manually hiding the chrome instead of just allowing the fullscreen flags to take care of it was apparently causing the window to expand by the size of the chrome after the exit animation.

Removing this doesn't seem to change fullscreen functionality, yet it solves the expanding window.

### DIFF
--- a/UI/AppKit/Interface/TabController.mm
+++ b/UI/AppKit/Interface/TabController.mm
@@ -66,7 +66,6 @@ static NSString* const TOOLBAR_TAB_OVERVIEW_IDENTIFIER = @"ToolbarTabOverviewIde
 
     bool m_fullscreen_requested_for_web_content;
     bool m_fullscreen_exit_was_ui_initiated;
-    bool m_fullscreen_should_restore_tab_bar;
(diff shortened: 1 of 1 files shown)


Zaggy1024:
Updated the commit message to reflect new information. @trflynn89 pointed out that this causes the toolbar to become visible in content-initiated fullscreen when cursor moves to the top, which this code was originally added to prevent. Re-enabling the toolbar before the unfullscreening animation completes results in the toolbar remaining hidden. To achieve this desired behavior, we may need to come up with another solution.

UI/AppKit/Interface/TabController.mm
```diff
@@ -493,14 +490,6 @@ - (void)windowDidChangeScreen:(NSNotification*)notification
 
 - (void)windowWillEnterFullScreen:(NSNotification*)notification
 {
```
trflynn89, line 492: Looks like this function is now empty, so it can be removed.

PR #8546, opened 2026-03-21
Performance & crash fixes for chess.com

awesomekling:
Three things here:
- Remove annoying debug spam in HTMLScriptElement.
- Use binary search in IndexedDB to prevent 100% CPU churn in WebContent.
- Fix crash when CSS animation produces guaranteed-invalid value.

It's funny, I figured the sluggishness on chess.com was the chess engine, but it turns out it was just IndexedDB being inefficient 😅 

(Although note that the chess engine *is* still slow, running as Wasm in the WebWorker process.)


Libraries/LibWeb/IndexedDB/Internal/ObjectStore.cpp
```diff
@@ -59,19 +59,53 @@ void ObjectStore::visit_edges(Visitor& visitor)
...
+                h = mid;
+        }
+        hi = l;
+    }
+
+    if (lo < hi) {
+        if (m_mutation_log) {
+            Vector<ObjectStoreRecord> deleted;
+            for (size_t i = lo; i < hi; ++i)
+                deleted.append(m_records[i]);
```
tcl3: I think we could `move()` this.

PR #8550, opened 2026-03-21
LibWeb: Move standalone SVG document layout handling out of BFC

Psychpsyo:
I'm simplifying the BFC code in preparation for multicol layout and this is really not something that every BFC should have to consider all of the time.


Psychpsyo:
Test failures seem weird and unrelated.

PR #8557, opened 2026-03-21
LibWeb: Implement incredibly basic multicol handling

Psychpsyo:
This is obviously still missing a bazillion things, but it's a start.

### DIFF
--- a/Libraries/LibWeb/Layout/BlockFormattingContext.cpp
+++ b/Libraries/LibWeb/Layout/BlockFormattingContext.cpp
@@ -19,6 +19,7 @@
 #include <LibWeb/Layout/BlockFormattingContext.h>
 #include <LibWeb/Layout/Box.h>
 #include <LibWeb/Layout/FieldSetBox.h>
+#include <LibWeb/Layout/Fragmentation.h>
 #include <LibWeb/Layout/InlineFormattingContext.h>
 #include <LibWeb/Layout/LegendBox.h>
 #include <LibWeb/Layout/LineBuilder.h>
@@ -97,27 +98,37 @@ static bool margins_collapse_through(Box const& box, LayoutState& state)
     return true;
 }
 
-void BlockFormattingContext::run(AvailableSpace const& available_space)
+void BlockFormattingContext::run(AvailableSpace const& available_space, Optional<FragmentationContext&> fragmentation_context)
 {
     FORMATTING_CONTEXT_TRACE();
     // https://drafts.csswg.org/css-multicol-2/#the-multi-column-model
     auto const& root_state = m_state.get(root());
     auto column_count = determine_used_value_for_column_count(root_state.content_width());
-    if (column_count.has_value()) {
+    // FIXME: Figure out how to implement multicol layout with unconstrained (indefinite) height.
+    // FIXME: Support nested fragmentation contexts.
+    if (column_count.has_value() && available_space.height.is_definite() && !fragmentation_context.has_value()) {
         auto column_width = determine_used_value_for_column_width(root_state.content_width(), column_count.value());
-        // FIXME: Do multi-column layout.
-        (void)column_width;
+        auto multicol_fragmentation_context = ColumnFragmentationContext(root(), column_width, available_space.height.to_px_or_zero(), get_column_gap_used_value_for_multicol(root_state.content_width()));
+        auto multicol_available_space = AvailableSpace(AvailableSize::make_definite(column_width), AvailableSize::make_indefinite());
+
+        layout_children(multicol_available_space, multicol_fragmentation_context);
+        return;
     }
 
+    layout_children(available_space, fragmentation_context);
+}
+
+void BlockFormattingContext::layout_children(AvailableSpace const& available_space, Optional<FragmentationContext&> fragmentation_context)
+{
     if (auto const* fieldset_box = as_if<FieldSetBox>(root()); fieldset_box && fieldset_box->rendered_legend()) {
-        layout_fieldset_with_rendered_legend(*fieldset_box, available_space);
+        layout_fieldset_with_rendered_legend(*fieldset_box, available_space, fragmentation_context);
         return;
     }
 
     if (root().children_are_inline())
-        layout_inline_children(root(), available_space);
+        layout_inline_children(root(), available_space, fragmentation_context);
     else
-        layout_block_level_children(root(), available_space);
+        layout_block_level_children(root(), available_space, fragmentation_context);
 
     // Fieldsets without a rendered legend skip collapsed margin assignment.
     if (is<FieldSetBox>(root()))
@@ -596,14 +607,14 @@ void BlockFormattingContext::resolve_used_height_if_treated_as_auto(Box const& b
     box_state.set_content_height(height);
 }
 
-void BlockFormattingContext::layout_inline_children(BlockContainer const& block_container, AvailableSpace const& available_space)
+void BlockFormattingContext::layout_inline_children(BlockContainer const& block_container, AvailableSpace const& available_space, Optional<FragmentationContext&> fragmentation_context)
 {
     VERIFY(block_container.children_are_inline());
 
     auto& block_container_state = m_state.get_mutable(block_container);
 
     InlineFormattingContext context(m_state, m_layout_mode, block_container, block_container_state, *this);
-    context.run(available_space);
+    context.run(available_space, fragmentation_context);
 
     if (!block_container_state.has_definite_width()) {
         // NOTE: min-width or max-width for boxes with inline children can only be applied after inside layout
@@ -764,7 +775,7 @@ static CSSPixels containing_block_height_to_resolve_percentage_in_quirks_mode(Bo
     VERIFY_NOT_REACHED();
 }
 
-void BlockFormattingContext::layout_block_level_box(Box const& box, BlockContainer const& block_container, CSSPixels& bottom_of_lowest_margin_box, AvailableSpace const& available_space)
+void BlockFormattingContext::layout_block_level_box(Box const& box, BlockContainer const& block_container, CSSPixels& bottom_of_lowest_margin_box, AvailableSpace const& available_space, Optional<FragmentationContext&> fragmentation_context)
 {
     if (box.is_absolutely_positioned()) {
         if (m_layout_mode == LayoutMode::Normal) {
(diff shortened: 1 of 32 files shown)


Psychpsyo:
Test failures, surprisingly, seem unrelated.
(and do not happen locally for me, on Linux)

Libraries/LibWeb/PixelUnits.h
```diff
@@ -219,6 +219,11 @@ class WEB_API CSSPixels {
     constexpr CSSPixelFraction operator/(CSSPixels const& other) const;
     constexpr CSSPixels operator/(CSSPixelFraction const& other) const;
 
+    constexpr CSSPixels operator%(CSSPixels const& other) const
+    {
+        return from_raw(raw_value() % other.raw_value());
+    }
```
AtkinsSJ, line 225: Nit: This change can live in a separate commit.

AtkinsSJ (commented):
I can't say I know a lot about fragmented layout, but this seems a pretty reasonable way to approach it.

Libraries/LibWeb/PixelUnits.h
Psychpsyo: ~~Wouldn't it feel somewhat unmotivated then? It feels weird to me to just add a function without calling it from anywhere.~~

Psychpsyo: Done that now.

Psychpsyo:
> I can't say I know a lot about fragmented layout, but this seems a pretty reasonable way to approach it.

Heads up: Since your review I've added proper positioning of block elements in fragmented BFCs also, since it wasn't that much extra stuff. The general approach is the same except that the FragmentationContext now also holds a reference to its root box, so that elements inside of nested blocks can correctly calculate their offset to that root to know how far along the FragmentationContext they are. (With the old calculations, they were taking the Y position inside their parent, which is not what we want)

Here's a picture of a block being positioned (and sized!) correctly inside a multicol:
image
Here's what that looks like in Firefox: (differences in column count/height stem from the font, I think)
image
We have the right position and size, we just don't cut the box apart yet. (I'll need to generalize our inline fragment rendering code for that in a future PR.)

Psychpsyo:
Test failures seem unrelated.

Libraries/LibWeb/Layout/BlockFormattingContext.cpp
```diff
@@ -899,6 +910,14 @@ void BlockFormattingContext::layout_block_level_box(Box const& box, BlockContain
         left_space_before_children_formatted = space_used_before_children_formatted.left;
     }
 
+    auto unfragmented_box_y = box_state.offset.y();
+    if (fragmentation_context.has_value()) {
+        FragmentationContext const* context = &fragmentation_context.value();
```
AtkinsSJ: Nit: Why is this a pointer instead of a reference?

```suggestion
        FragmentationContext const& context = fragmentation_context.value();
```

Libraries/LibWeb/Layout/LineBuilder.cpp
```diff
@@ -257,6 +257,27 @@ void LineBuilder::update_last_line()
         }
     }
 
+    auto line_height = m_context.containing_block().computed_values().line_height();
+
+    CSSPixels inline_fragmentainer_offset = 0;
+    CSSPixels block_fragmentainer_offset = 0;
+    if (fragmentation_context.has_value()) {
+        FragmentationContext const* context = &fragmentation_context.value();
```
AtkinsSJ: Same comment.

Libraries/LibWeb/Layout/LineBuilder.cpp
```diff
@@ -257,6 +257,27 @@ void LineBuilder::update_last_line()
...
+
+        // FIXME: Block doesn't always mean Y-axis and inline doesn't always mean X-axis, so account for writing mode.
+        auto fragmented_flow_block_offset = m_context.content_box_rect_in_ancestor_coordinate_space(
+                                                         m_layout_state.get(m_context.containing_block()), context->root())
```
AtkinsSJ: To be really nitty, having `m_context` and `context` in the same line feels like a footgun.

AtkinsSJ (commented):
Just nits!

Libraries/LibWeb/Layout/LineBuilder.cpp
Psychpsyo: True. I've changed these now to repeat `fragmentation_context.value()` instead of assigning it to a temporary `context` variable. I originally only really did that for brevity anyways and replacing it with `fragmentation_context_value` would defeat the point then.

Libraries/LibWeb/Layout/BlockFormattingContext.cpp
Psychpsyo: No good reason, but the variable is gone now anyways.

Psychpsyo:
> Just nits!

All done!

AtkinsSJ (approved):
:+1: Let's get fragmentaining!

kalenikaliaksandr (changes requested):
hmm, I don't think that approach when we layout everything as one column and then remap boxes would work for cases when box needs to split across columns, like:
```html
<div style="columns: 2; height: 100px; column-gap: 0; width: 200px;">
  <div style="height: 80px; background: blue;"></div>
  <div style="height: 80px; background: red;"></div>
</div>
```
do you have a plan how this issue could be addressed?

also it's not clear how we would handle `column-fill: balance` if current approach requires definite height.

Psychpsyo:
> hmm, I don't think that approach when we layout everything as one column and then remap boxes would work for cases when box needs to split across columns, like:
> 
> ```
> <div style="columns: 2; height: 100px; column-gap: 0; width: 200px;">
>   <div style="height: 80px; background: blue;"></div>
>   <div style="height: 80px; background: red;"></div>
> </div>
> ```
> 
> do you have a plan how this issue could be addressed?
> 
> also it's not clear how we would handle `column-fill: balance` if current approach requires definite height.

Refer to https://github.com/LadybirdBrowser/ladybird/pull/8761 for visually cutting boxes apart.
Refer to https://github.com/LadybirdBrowser/ladybird/pull/8771 for how to place those fragments into columns.
Refer to https://github.com/LadybirdBrowser/ladybird/pull/8785 for my `column-fill: balance` implementation.
Refer to https://github.com/LadybirdBrowser/ladybird/pull/8803 for slicing of replaced elements.

I have further plans for how to generalize some of the slicing/shunting behavior by moving more of it into the FragmentationContext which I think will be necessary for nesting multicol containers. (at least given my current progress on that)

TL;DR:
Yes, I have plans for how to address that, and all of those are implemented already.

kalenikaliaksandr:
> > hmm, I don't think that approach when we layout everything as one column and then remap boxes would work for cases when box needs to split across columns, like:
> > ```
> > <div style="columns: 2; height: 100px; column-gap: 0; width: 200px;">
> >   <div style="height: 80px; background: blue;"></div>
> >   <div style="height: 80px; background: red;"></div>
> > </div>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > do you have a plan how this issue could be addressed?
> > also it's not clear how we would handle `column-fill: balance` if current approach requires definite height.
> 
> Refer to #8761 for visually cutting boxes apart. Refer to #8771 for how to place those fragments into columns. Refer to #8785 for my `column-fill: balance` implementation. Refer to #8803 for slicing of replaced elements.
> 
> I have further plans for how to generalize some of the slicing/shunting behavior by moving more of it into the FragmentationContext which I think will be necessary for nesting multicol containers. (at least given my current progress on that)
> 
> Does this answer your questions?



> > hmm, I don't think that approach when we layout everything as one column and then remap boxes would work for cases when box needs to split across columns, like:
> > ```
> > <div style="columns: 2; height: 100px; column-gap: 0; width: 200px;">
> >   <div style="height: 80px; background: blue;"></div>
> >   <div style="height: 80px; background: red;"></div>
> > </div>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > do you have a plan how this issue could be addressed?
> > also it's not clear how we would handle `column-fill: balance` if current approach requires definite height.
> 
> Refer to #8761 for visually cutting boxes apart. Refer to #8771 for how to place those fragments into columns. Refer to #8785 for my `column-fill: balance` implementation. Refer to #8803 for slicing of replaced elements.
> 
> I have further plans for how to generalize some of the slicing/shunting behavior by moving more of it into the FragmentationContext which I think will be necessary for nesting multicol containers. (at least given my current progress on that)
> 
> Does this answer your questions?
> 
> TL;DR: Yes, I have plans for how to address that, and all of those are implemented already.

sure, let me have a look at other PRs. what about break tokens btw? I am not trying to say your direction is completely wrong, but last time I investigated fragmentation support it seemed like (I might be wrong here) chromium had to substantially rewrite their layout subsystem to avoid "layout as one column and then remap" approach, but I do understand that it's easiest option if you want to slap fragmentation on top of existing layout subsystem.

Psychpsyo:
> sure, let me have a look at other PRs. what about break tokens btw? I am not trying to say your direction is completely wrong, but last time I investigated fragmentation support it seemed like (I might be wrong here) chromium had to substantially rewrite their layout subsystem to avoid "layout as one column and then remap" approach, but I do understand that it's easiest option if you want to slap fragmentation on top of existing layout subsystem.

Part of this I said on Discord already, but:
This is not "layout as one column and then remap", it is the start of "layout at a 1-column width until you need to proceed to the next fragmentainer, at which point you consider your options."
These first few PRs are not perfect in their break handling yet; "consider your options" is right now just "do whatever" in two places, so I want to formalize out that behavior a fair bit more. The current beginnings of that are in `FragmentationContext::make_fragmentation_decision_at()` from https://github.com/LadybirdBrowser/ladybird/pull/8785.

The "remapping" (x/y offsetting) is there because it is intuitive for formatting contexts to place their next bit of content right after the previous one (plus some considerations for fragmentainer breaks), without having to be aware of the physical location of the fragmentainer it goes into up-front. They get to just put the content where they know it to belong if all the fragmentainers were stacked block-end to block-end and ask the fragmentation context for the actual fragmentainer location to offset it needs there.
Sure, we could ensure that a formatting context gets told up-front what fragmentainer it is laying content into so that it knows that offset and can use it, but I don't see the necessity in that if the current fragmentainer can just be calculated from the current progress along the fragmented flow.

Psychpsyo:
Just rebased onto master because of conflicts again.
It's been a month and a half. Anything I can do to help with moving the review along here?

gmta:

PR #8565, opened 2026-03-22
Tests/LibWeb: Skip flaky load event tests

shannonbooth:


gmta:
😅 let me have a shot at fixing the issue first instead of skipping tests

gmta:
Opened #8566 as the fix for these flakes.

PR #8568, opened 2026-03-22
LibWeb: Report actual network error for failed XHRs

gmta:
Seeing "NetworkError: Network error" when an XHR failed was not that useful - use the actual error message when available.

### DIFF
--- a/Tests/LibWeb/Text/input/XHR/XMLHttpRequest-network-error-message.html
+++ b/Tests/LibWeb/Text/input/XHR/XMLHttpRequest-network-error-message.html
@@ -1,0 +1,25 @@
+<!DOCTYPE html>
+<script src="../include.js"></script>
+<script>
+    asyncTest(async (done) => {
+        const server = httpTestServer();
+
+        await server.createEcho("GET", "/close-connection", {
+            status: 0,
+            close_connection: true,
+        });
+
+        // FIXME: Use the URL returned by createEcho instead of a relative path. Currently that URL uses 127.0.0.1
+        //        which is cross-origin with the localhost page, causing a CORS error instead of a network error.
+        const xhr = new XMLHttpRequest();
+        xhr.open("GET", "/close-connection", false);
+        try {
+            xhr.send();
+            println("FAIL: no exception thrown");
+        } catch (e) {
+            println(e.name);
+            println(e.message);
+        }
+        done();
+    });
+</script>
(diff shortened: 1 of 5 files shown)


Tests/LibWeb/Text/input/XHR/XMLHttpRequest-network-error-message.html
```diff
@@ -0,0 +1,23 @@
...
+    asyncTest(async (done) => {
+        const server = httpTestServer();
+
+        await server.createEcho("GET", "/close-connection", {
+            status: 0,
+            close_connection: true,
+        });
+
+        const xhr = new XMLHttpRequest();
+        xhr.open("GET", "/close-connection", false);
```
shannonbooth: Could this instead use the URL returned by createEcho? 

I'm thinking of changing the echo server implementation so that /echo/* is reserved for our own stuff to tell it apart from wanting to serve a normal file.

(https://github.com/LadybirdBrowser/ladybird/pull/8458/changes/f17aef1cabf8e4d46bde06e578a3af2ff559e756)
```suggestion
        const url = await server.createEcho("GET", "/close-connection", {
            status: 0,
            close_connection: true,
        });
        xhr.open("GET", url, false);
```

(Also I realize now I got it wrong about the flake for the load event, there indeed was a flake that I saw unrelated to my changes but I also had a boog)

gmta: Absolutely!

gmta: Hm, that triggers a CORS error instead. That would alleviate the need for a close connection feature 😅

shannonbooth: Oh interesting! I was expecting that to be equivalent, but it's using 127.0.0.1 as the return URL for some reason instead of localhost... Which is probably not what we want most of the time!

gmta: I've reverted the change to the test for now, and added a FIXME about using the returned URL. Fixing that to not have 127.0.0.1 in it is for a future PR :)

shannonbooth: Cool, integrated into https://github.com/LadybirdBrowser/ladybird/pull/8458/commits

PR #8572, opened 2026-03-22
LibWeb: Skip declaration parsing in block contents when not applicable

awesomekling:
When parsing block contents, the CSS parser speculatively tries to parse each item as a declaration first. If that fails, it restores the token position and tries again as a qualified rule. This means every qualified rule inside an at-rule block (e.g. @layer, @media) gets parsed twice: once as a failed declaration (which consumes all tokens via consume_the_remnants_of_a_bad_declaration), and then again successfully as a rule.

Add a lookahead that checks for the `ident whitespace* ':'` pattern before attempting declaration parsing. Since declarations must start with this pattern per spec, we can skip the attempt entirely when it doesn't match and go straight to qualified rule parsing.

This is a massive win on large Tailwind CSS stylesheets (like the one used by chatgpt.com) where thousands of rules inside @layer blocks were being double-parsed. On a 1.2MB Tailwind v4 stylesheet, parse time goes from ~2000ms to ~95ms (21x speedup).

### DIFF
--- a/Libraries/LibWeb/CSS/Parser/Parser.cpp
+++ b/Libraries/LibWeb/CSS/Parser/Parser.cpp
@@ -729,43 +729,50 @@ Vector<RuleOrListOfDeclarations> Parser::consume_a_blocks_contents(TokenStream<T
 
         // anything else
         {
-            // Mark input.
-            input.mark();
-
-            // Consume a declaration from input, with nested set to true.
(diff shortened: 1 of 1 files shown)


awesomekling:
This cuts ~12 seconds of loading time off of chatgpt.com :)

PR #8578, opened 2026-03-22
LibWeb: Avoid division by zero in multicol calculations the right way

Psychpsyo:
This got broken in #8057.

@tcl3, since you made the original change.

No new test to verify the change in behavior, since we don't have enough of multicol layout for me to write one.
I'll try to remember to add one in #8557 once this gets merged.

### DIFF
--- a/Libraries/LibWeb/Layout/BlockFormattingContext.cpp
+++ b/Libraries/LibWeb/Layout/BlockFormattingContext.cpp
@@ -1568,24 +1568,45 @@ CSSPixels BlockFormattingContext::greatest_child_width(Box const& box) const
 Optional<int> BlockFormattingContext::determine_used_value_for_column_count(CSSPixels const& U) const
 {
     auto const& computed_values = root().computed_values();
+    // (01)  if ((column-width = auto) and (column-count = auto)) then
     if (computed_values.column_width().is_auto() && computed_values.column_count().is_auto()) {
+        // (02)      exit; /* not a multicol container */
         return {};
     }
+
+    // (03)  if column-width = auto then
     if (computed_values.column_width().is_auto()) {
+        // (04)      N := column-count
         return computed_values.column_count().value();
     }
+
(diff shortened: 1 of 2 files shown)


Libraries/LibWeb/Layout/BlockFormattingContext.cpp
```diff
@@ -1581,26 +1568,44 @@ CSSPixels BlockFormattingContext::greatest_child_width(Box const& box) const
...
+    if (computed_values.column_width().is_auto() && computed_values.column_count().is_auto())
+        // (02)      exit; /* not a multicol container */
         return {};
-    }
-    if (computed_values.column_width().is_auto()) {
+
+    // (03)  if column-width = auto then
+    if (computed_values.column_width().is_auto())
+        // (04)      N := column-count
         return computed_values.column_count().value();
```
gmta, line 1580: Please use braces here

Libraries/LibWeb/Layout/BlockFormattingContext.cpp
```diff
@@ -1581,26 +1568,44 @@ CSSPixels BlockFormattingContext::greatest_child_width(Box const& box) const
...
-    auto column_count = divisor == 0 ? 1 : max(1, ((U + column_gap) / divisor).to_int());
+    auto column_width = get_column_width_used_value_for_multicol(U);
+
+    // (05)  else if column-count = auto then
     if (computed_values.column_count().is_auto())
-        return column_count;
-    return min(computed_values.column_count().value(), column_count);
+        // (06)      N := max(1,
+        // (07)        floor((U + column-gap)/(column-width + column-gap)))
+        return max(1, ((U + column_gap) / (column_width + column_gap)).to_int());
```
gmta, line 1590: Please add braces here as well

Psychpsyo: I've put all the braces back in now.

PR #8580, opened 2026-03-22
LibWeb: Allow cursor movement across contenteditable node boundaries 

tcl3:
This makes it possible to use the arrow keys to navigate between differently styled elements within a contenteditable element.

Improves the usability of: https://playground.lexical.dev/


Libraries/LibWeb/DOM/EditingHostManager.cpp
```diff
@@ -93,16 +146,39 @@ GC::Ptr<Selection::Selection> EditingHostManager::get_selection_for_navigation(C
     if (collapse == CollapseSelection::Yes && !focus_node->is_editable())
         return {};
 
+    m_document->update_layout(UpdateLayoutReason::EditingCursorNavigation);
     return selection;
 }
 
+void EditingHostManager::cross_to_text_node(Selection::Selection& selection, Node& old_focus_node, Text& target, unsigned offset, bool collapse_selection)
```
konradekk: imho, `move_selection_to_text_node` would be a better name: 

```suggestion
void EditingHostManager::move_selection_to_text_node(Selection::Selection& selection, Node& old_focus_node, Text& target, unsigned offset, bool collapse_selection)
```

(optionally, `move_cursor_to_text_node`, if that’s the case here – looking at the next function)

tcl3: Renamed to `move_selection_focus_to_text_node()`.

PR #8583, opened 2026-03-23
LibWeb: Kmac use outputLength instead of length in KmacParams 

mikiubo:
Use outputLength instead of length in KmacParams matching new spec updates.
Also Use OpenSSL wrappers in KMAC to clean up the code.


Libraries/LibCrypto/OpenSSLForward.h
```diff
@@ -34,5 +34,7 @@ void EVP_PKEY_CTX_free(EVP_PKEY_CTX*);
 void EVP_PKEY_free(EVP_PKEY*);
 void EVP_KDF_CTX_free(EVP_KDF_CTX* ctx);
 void EVP_KDF_free(EVP_KDF* kdf);
+void EVP_MAC_free(EVP_MAC*);
+void EVP_MAC_CTX_free(EVP_MAC_CTX*);
```
gmta, line 37: To maintain alphabetical/ascii ordering, these two lines should be swapped.

mikiubo: fixed

PR #8586, opened 2026-03-23
LibWeb: Store MediaFeatureValue underlying value as StyleValue

Calme1709:
This simplifies handling in some cases and gets us a step closer to removing the `FooOrCalculated` classes.

Also includes improvements to handling of calculated values in `<ratio>`.


tcl3 (commented):
It might be worth importing https://wpt.live/css/mediaqueries/mq-calc-sign-function-003.html, which passes after these  changes 🎉

Calme1709:
Looks like it was already imported, just needed to unskip it :)

PR #8588, opened 2026-03-23
Meta: Ensure that idl files link to draft specs

Psychpsyo:
I manually checked that all these still point to the right things.
One weird one is https://w3c.github.io/device-memory/#sec-device-memory-js-api, which actually just redirects to the TR link and loses the URL fragment (`#sec-device-memory-js-api`) in the process. But the spec is incredibly short and basic, so I think this is fine. And I'd assume that's probably also the reason for why there is no draft version of this spec.

I will make a second PR in a bit that enforces this for C++ files also, I just thought I'd start small.

Inspired by me constantly forgetting to link to draft specs instead of published ones, which probably only sometimes gets caught in review.

This is what the error looks like:
image

### DIFF
--- a/Libraries/LibWeb/Animations/Animation.idl
+++ b/Libraries/LibWeb/Animations/Animation.idl
@@ -2,7 +2,7 @@
 #import <Animations/AnimationTimeline.idl>
 #import <DOM/EventTarget.idl>
 
-// https://www.w3.org/TR/web-animations-1/#the-animation-interface
+// https://drafts.csswg.org/web-animations-1/#the-animation-interface
 [Exposed=Window]
 interface Animation : EventTarget {
     constructor(optional AnimationEffect? effect = null,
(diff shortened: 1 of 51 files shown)


AtkinsSJ:
Sounds good to me. I just want to ask though: Did you check that the newly-linked IDL definitions match what we have? It sounds like you probably did, but if not, it'd be nice to add FIXMEs for any differences.

Psychpsyo:
> Sounds good to me. I just want to ask though: Did you check that the newly-linked IDL definitions match what we have? It sounds like you probably did, but if not, it'd be nice to add FIXMEs for any differences.

I generally checked that they link to the same definition. I did not check whether or not every single property or enum value or such still matches, since those are bound to grow out-of-sync with spec changes anyways. Even with the links to published specs, that'd catch up to us eventually. It even did that for the links I changed in [WheelEvent.idl](https://github.com/LadybirdBrowser/ladybird/pull/8588/changes#diff-c7fcb1acbdeebea7e1f14b99f9f53edd547fa3225b0ffda213c2147e290b1106). Those had become invalid since they got added, because WheelEvent and WheelEventInit had been moved to a different spec.

Though at the point where I only had links to `filter-effects-1` left, I just regex replaced them all and sporadically checked some to make sure that the IDs for the URL fragments are largely still the same in the draft spec. (Anything I might've broken there would've broken anyways when changes gets published)

So I agree that adding FIXMEs sounds nice, but I think that's out of scope for this PR and also might warrant writing a script that does it for all of our IDL files so that it can be re-run automatically every now and then. (Might look into that once I get off work or on the weekend)

AtkinsSJ:
That's fair. I'm happy to merge as-is.

PR #8589, opened 2026-03-23
LibWebView+UI: Bookmarks 🔖

trflynn89:
[bookmarks.webm](https://github.com/user-attachments/assets/5b08b109-e278-4892-935d-77a515d742b1)

There is *plenty* more work to do here. Although the bookmark model supports folders, there's currently no way to create folders (other than manually adding them to the stored JSON). We will also want to support context menus, editing, rearranging, etc.


Libraries/LibWebView/BookmarkStore.cpp
```diff
@@ -0,0 +1,337 @@
...
+static constexpr auto URL_KEY = "url"sv;
+static constexpr auto TITLE_KEY = "title"sv;
+static constexpr auto FAVICON_KEY = "favicon"sv;
+static constexpr auto CHILDREN_KEY = "children"sv;
+
+static constexpr auto TYPE_BOOKMARK = "bookmark"sv;
+static constexpr auto TYPE_FOLDER = "folder"sv;
+
+// Based on https://w3c.github.io/webcrypto/#dfn-generate-a-random-uuid
+static String generate_uuid()
```
shannonbooth: Should we maybe put this in AK/Random.h so we have a shared between both? I recall considering putting it there where adding nonce to URL::Origin before i realized it was silly to use uuid there

trflynn89: Done - I also made it stop throwing OOM

PR #8594, opened 2026-03-23
LibWeb: Don't crash accessing stale nested navigable paintable

gmta:
The scroll state collection loop in record_display_list_and_scroll_state() called paintable() on hosted documents, which asserts layout is up to date. This crashes when a nested document has stale layout but a cached display list, e.g. a render-blocked iframe whose DOM was modified by document.open(). Since scroll offsets are independent of layout freshness, use unsafe_paintable() to skip the assertion.

Fixes opening emails in Roundcube Webmail.

### DIFF
--- a/Libraries/LibWeb/HTML/Navigable.cpp
+++ b/Libraries/LibWeb/HTML/Navigable.cpp
@@ -2839,19 +2839,29 @@ void Navigable::record_display_list_and_scroll_state(PaintConfig paint_config)
     Painting::ScrollStateSnapshotByDisplayList scroll_state_snapshot_by_display_list;
     document_paintable.refresh_scroll_state();
     scroll_state_snapshot_by_display_list.set(*display_list, document_paintable.scroll_state_snapshot());
+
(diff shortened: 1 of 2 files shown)


awesomekling:
> Since scroll offsets are independent of layout freshness, use
unsafe_paintable() to skip the assertion.

Hmm, is that entirely true?

If I'm scrolled to `(x: 0, y: 500)` but layout is invalid. What if relayout would make me `height: 200`. Then the scroll offset would be unable to reach `y: 500` and we'd clamp it, right?

I'm not 100% sure about this, but it feels fishy so I'm asking :)

gmta:
@awesomekling 

> > Since scroll offsets are independent of layout freshness, use unsafe_paintable() to skip the assertion.
>
> Hmm, is that entirely true?

If we're talking about some JS API, then we would absolutely need to update the layout. However, scroll state is closely associated to the (cached) display list, so we're effectively dealing with visually consistent scrolling - that's why we can use the unsafe paintable accessor. Once the layout updates, both the paintable and display list will be updated.

PR #8600, opened 2026-03-24
LibWeb: Apply text-overflow ellipsis as line box post-processing

gmta:
The previous implementation checked text-overflow and overflow-x on the text node's direct parent during inline item iteration. Since these are non-inherited properties, ellipsis only worked for text directly inside the block container, not when wrapped in inline elements like <span> or <a>.

Move ellipsis truncation to a post-processing step after line boxes are constructed, checking the containing block instead.

| Before | After | Firefox |
|--|--|--|
| image | image | image |

### DIFF
--- a/Libraries/LibWeb/Layout/InlineFormattingContext.cpp
+++ b/Libraries/LibWeb/Layout/InlineFormattingContext.cpp
@@ -244,6 +244,102 @@ void InlineFormattingContext::apply_justification_to_fragments(CSS::TextJustify 
     }
 }
 
+// https://drafts.csswg.org/css-overflow-4/#text-overflow
+void InlineFormattingContext::apply_text_overflow_ellipsis(Vector<LineBox>& line_boxes)
+{
+    // This property specifies rendering when inline content overflows its line box edge in the inline progression
+    // direction of its block container element ("the block") that has overflow other than visible.
+
+    // NB: When inline children are wrapped in anonymous blocks (e.g. due to floats), we look past the anonymous
+    //     wrapper to the actual element that has text-overflow and overflow set.
+    GC::Ref<Box const> block = containing_block();
+    if (block->is_anonymous())
+        block = *block->non_anonymous_containing_block();
+
+    // FIXME: Support the <string>, fade, and fade() values, as well as the two-value syntax.
+    auto const& block_values = block->computed_values();
+    if (block_values.text_overflow() != CSS::TextOverflow::Ellipsis)
+        return;
+    if (block_values.overflow_x() == CSS::Overflow::Visible)
+        return;
+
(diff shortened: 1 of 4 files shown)


Libraries/LibWeb/Layout/InlineFormattingContext.cpp
```diff
@@ -244,6 +244,97 @@ void InlineFormattingContext::apply_justification_to_fragments(CSS::TextJustify
...
+
+    // FIXME: Support the <string>, fade, and fade() values, as well as the two-value syntax.
+    auto const& block_values = block->computed_values();
+    if (block_values.text_overflow() != CSS::TextOverflow::Ellipsis)
+        return;
+    if (block_values.overflow_x() == CSS::Overflow::Visible)
+        return;
+
+    // Render an ellipsis character (U+2026) to represent clipped inline content.
+    constexpr u32 ellipsis_codepoint = 0x2026;
```
tcl3, line 267: Would a fallback to 3 dots be worth doing if this character is unavailable, or do we always fall back to a font where this is available?

Libraries/LibWeb/Layout/InlineFormattingContext.cpp
```diff
@@ -244,6 +244,97 @@ void InlineFormattingContext::apply_justification_to_fragments(CSS::TextJustify
...
+                auto glyph_end = glyph.position.x() + glyph.glyph_width;
+                if (glyph_end > max_text_width)
+                    break;
+                keep_count++;
+                last_kept_end = glyph_end;
+                y_position = glyph.position.y();
+            }
+
+            if (keep_count < glyphs.size())
+                glyphs.remove(keep_count, glyphs.size() - keep_count);
```
tcl3, line 323: Should we make sure we keep at least one character here? For example, this renders "a" in other browsers but is blank for us:

```html
<!DOCTYPE html>
<style>
div {
    overflow: hidden;
    width: 1ch;
    text-overflow: ellipsis;
}
</style>
<div>abc</div>
```

Relevant spec text from https://drafts.csswg.org/css-overflow-4/#auto-ellipsis:
>  The first character or atomic inline-level element on a line must be clipped rather than ellipsed.

gmta: Oh good find! Fixed and added a test case

Libraries/LibWeb/Layout/InlineFormattingContext.cpp
gmta: There are a couple more scenarios to take into account, such as the fact that the ellipsis can be substituted by arbitrary strings. This codepoint is the same as we used before, I'd rather we take up automatic & manual substitution of it in another PR.

tcl3 (approved):

PR #8606, opened 2026-03-24
LibWeb: Implement `text-decoration-skip-ink` CSS property

gmta:
Basic support for parsing the property and segmentation of rendered text decoration.

Effectively it cuts out parts of text decoration where glyph descenders would usually intersect it.

| Before | After |
|--|--|
| image | image |


Libraries/LibWeb/Layout/Node.cpp
```diff
@@ -983,6 +984,7 @@ void NodeWithStyle::propagate_non_inherit_values(NodeWithStyle& target_node) con
 {
     // NOTE: These properties are not inherited, but we still have to propagate them to anonymous wrappers.
     target_node.mutable_computed_values().set_text_decoration_line(computed_values().text_decoration_line());
+    target_node.mutable_computed_values().set_text_decoration_skip_ink(computed_values().text_decoration_skip_ink());
```
tcl3: The comment above says these are not inherited properties, but `text-decoration-skip-ink` is inherited.

Libraries/LibWeb/Painting/PaintableWithLines.cpp
```diff
@@ -495,6 +496,59 @@ void PaintableWithLines::paint_cursor(DisplayListRecordingContext& context) cons
...
+    float half_thickness = line_thickness / 2.f;
+    float y_top = line_y - half_thickness - blob_origin_y;
+    float y_bottom = line_y + half_thickness - blob_origin_y;
+
+    auto intervals = glyph_run->get_glyph_intercepts(scale, y_top, y_bottom);
+    if (intervals.is_empty())
+        return { { line_start_x, line_end_x } };
+
+    // Convert intercepts from blob-local x to device pixels, and dilate to create visible gaps.
+    float dilation = min(static_cast<float>(line_thickness), 2.f * static_cast<float>(scale));
```
tcl3: Would it be worth taking the font size into account here? The minimum gap seems like it would be quite tight at larger font sizes.

Libraries/LibWeb/Layout/Node.cpp
gmta: Oops - that had no effect. Removed.

Libraries/LibWeb/Painting/PaintableWithLines.cpp
gmta: Yeah, made the dilation depend on font size instead of line thickness, which seems like a better metric anyway.

gmta:
Made some more changes to the way we determined the gap "segments", since our selection drawing logic would split up fragments and completely mess up our the visual gaps. It's a bit more awkward but visually stable now 😅

Libraries/LibWeb/Painting/PaintableWithLines.cpp
tcl3, line 590: I'm seeing an issue where partially selecting underlined text causes the underline to jump upwards:

https://github.com/user-attachments/assets/fbb86e95-f625-48c0-8923-9bec379d0bd4

I think this is because `range_rect()` has it's height inflated compared to `absolute_rect()` and this pulls the underline up.

gmta: Doh, fixed that by intersecting the absolute rect horizontally with the span so we retain its height + added a regression test.

PR #8609, opened 2026-03-25
LibWeb: Improve `<integer>` clamping and rounding in CSS

Calme1709:
This PR includes two main changes:
- CSS integer values are clamped to i32 at parse time (or computed value time for `calc()`)
- Rounding of integers breaks ties towards +Infinity rather than away from zero (i.e. -2.5 rounds to -2 rather than -3)

Previously we handled this specifically for z-index at used value time, now we handle it for all `<integer>`s at the appropriate time


Calme1709:
Marked as draft to fix issues exposed by sanitizer

Calme1709:
This PR is now ready for review :)

PR #8612, opened 2026-03-25
LibRegex: Replace C++ ECMA-262 regex engine with Rust implementation

awesomekling:
Replace the old C++ ECMA-262 regex engine with a new Rust implementation purpose-built for ECMA-262 semantics on WTF-16 data. The old engine shared a generic parser and matcher stack across ECMA-262 and POSIX modes; the new one models lone surrogates and other JS-specific behavior directly.

The architecture is a clean split: parse patterns into a lossless AST, then lower to bytecode for a backtracking VM. This keeps validation and optimization in the syntax tree instead of teaching every transformation to rewrite partially-built bytecode.

Runtime fast paths include pre-resolved Unicode properties, first-character and character-class filters, trailing literal extraction for anchored patterns, and reusable scratch storage so global searches don't pay setup costs per match.

All in-tree callers (LibJS RegExp, URLPattern, HTML input patterns, LibHTTP) are migrated through a new `ECMAScriptRegex` C++ facade. The LibJS RegExp builtins are simplified along the way: shared `exec_internal`, cached compiled regex pointers, batch `find_all` for global replace, and optimized replace/test/split paths. The legacy C++ engine (~14k lines) is deleted.

Regex literals are now validated during JS parsing (safe off the main thread with the new engine), removing the deferred compilation pass that previously had to run on the main thread after background parsing.

Bonus: import the regex test suites from V8 and JavaScriptCore for increased testing.

Benchmarks are very happy:

```
Suite       Test                              Speedup  Old (Mean ± Range)     New (Mean ± Range)
----------  ------------------------------  ---------  ---------------------  ---------------------
Octane      regexp.js                           3.06   266.0 ± 265.0 … 267.0  814.0 ± 814.0 … 814.0
RegExp      speedometer-jquery-regexp-1.js     10.15   1.289 ± 1.285 … 1.292  0.127 ± 0.127 … 0.127
RegExp      speedometer-jquery-regexp-2.js      1.61   0.178 ± 0.178 … 0.178  0.111 ± 0.109 … 0.112
```

test262 is happy as well:

```
Summary:
    Diff Tests:
        +8 ✅    -5 ❌    -3 💀

Diff Tests:
    test/built-ins/RegExp/CharacterClassEscapes/character-class-digit-class-escape-negative-cases.js      💀 -> ✅
    test/built-ins/RegExp/CharacterClassEscapes/character-class-whitespace-class-escape-negative-cases.js 💀 -> ✅
    test/built-ins/RegExp/CharacterClassEscapes/character-class-word-class-escape-negative-cases.js       💀 -> ✅
    test/built-ins/RegExp/lookBehind/back-references-to-captures.js                                       ❌ -> ✅
    test/built-ins/RegExp/lookBehind/captures-negative.js                                                 ❌ -> ✅
    test/built-ins/RegExp/lookBehind/captures.js                                                          ❌ -> ✅
    test/built-ins/RegExp/lookBehind/mutual-recursive.js                                                  ❌ -> ✅
    test/built-ins/RegExp/named-groups/lookbehind.js                                                      ❌ -> ✅
```

### DIFF
--- a/Libraries/LibRegex/Rust/src/parser.rs
+++ b/Libraries/LibRegex/Rust/src/parser.rs
@@ -1406,6 +1406,7 @@ impl Parser {
                     self.pos += 2; // consume '--'
                     operands.push(self.parse_class_set_operand()?);
                 }
+                Self::validate_class_set_operation_operands(&operands)?;
                 Ok(ClassSetExpression::Subtraction(operands))
             }
             Some(('&', '&')) => {
@@ -1417,6 +1418,7 @@ impl Parser {
                     self.pos += 2; // consume '&&'
                     operands.push(self.parse_class_set_operand()?);
                 }
+                Self::validate_class_set_operation_operands(&operands)?;
                 Ok(ClassSetExpression::Intersection(operands))
             }
             _ => {
@@ -1428,6 +1430,17 @@ impl Parser {
                 Ok(ClassSetExpression::Union(operands))
             }
         }
+    }
+
+    fn validate_class_set_operation_operands(operands: &[ClassSetOperand]) -> Result<(), Error> {
+        if operands
+            .iter()
+            .any(|operand| matches!(operand, ClassSetOperand::Range(_, _)))
+        {
+            return Err(Error::InvalidCharacterClass);
+        }
+
+        Ok(())
     }
 
     /// Parse one `/v` `ClassSetOperand`.
--- a/Tests/LibJS/Runtime/builtins/RegExp/RegExp.js
+++ b/Tests/LibJS/Runtime/builtins/RegExp/RegExp.js
@@ -409,6 +409,10 @@ test("RegExp string literal", () => {
         "[\\q{\\w}]",
         "[\\q{\\q}]",
         "[^\\q{\\(\\)}]",
+        "[a-z&&b-y]",
+        "[a-z--[aeiou]]",
+        "[[a-z]&&b-y]",
+        "[\\u0061-z&&d]",
     ].forEach(pattern => {
         expect(() => new RegExp(pattern, "v")).toThrow(SyntaxError);
     });


Libraries/LibRegex/CMakeLists.txt
```diff
@@ -12,3 +12,13 @@ endif()
 
 ladybird_lib(LibRegex regex EXPLICIT_SYMBOL_EXPORT)
 target_link_libraries(LibRegex PRIVATE LibUnicode)
+
+if (ENABLE_RUST)
+    target_sources(LibRegex PRIVATE RustRegex.cpp)
+    corrosion_import_crate(MANIFEST_PATH Rust/Cargo.toml CRATES libregex_rust)
```
trflynn89: We should be using `import_rust_crate` instead of `corrosion_import_crate` (and then you don't need the `install` rule below). See 2c18c6a8c615028e6f208ff36c6717b9c1cc870f.

Are we also able to make use of the `cbindgen` changes that @ADKaster introduced?

Libraries/LibJS/Runtime/RegExpPrototype.cpp
```diff
@@ -74,100 +75,58 @@ static ThrowCompletionOr<void> increment_last_index(VM& vm, Object& regexp_objec
...
-
-    size_t start_index { 0 };
-    size_t end_index { 0 };
-};
-
-// 22.2.7.7 GetMatchIndexPair ( S, match ), https://tc39.es/ecma262/#sec-getmatchindexpair
-static Value get_match_index_pair(VM& vm, Utf16View const& string, Match const& match)
-{
-    auto& realm = *vm.current_realm();
+static HashMap<String, NonnullOwnPtr<regex::ECMAScriptRegex>> s_regex_cache;
```
jdahlin, line 79: Maybe consider some kind of eviction policy, or at least a TODO?

aplefull:
Tested a lil, some regressions:

```js
/^(b+|a){1,2}?bc/.exec("bbc"); // ["bbc", "b"]

malloc(): invalid next size (unsorted)

CRASH: Received signal SIGABRT (6)
```

```js
  "100.00".match(/^\-?(\d{1,}|\.{0,})*(\,\d{1,})?$/) // Should be ["100.00", "00", undefined]
  "100,00".match(/^\-?(\d{1,}|\.{0,})*(\,\d{1,})?$/) // Should be ["100,00", "100", ",00"]
  "1.000,00".match(/^\-?(\d{1,}|\.{0,})*(\,\d{1,})?$/) // Should be ["1.000,00", "000", ",00"]
```

```js
  "https://www.example.com/page?query=test".match(
    /^(((https?)|(ftp)):\/\/([\-\w]+\.)+\w{2,4}(\/[%\-\w]+(\.\w{2,})?)*(([\w\-\.\?\\/\*\$+@&#;`~=%!]*)(\.\w{2,})?)*\/?)$/,
  )
```

```js
"a--".match(/^(?:a?b?)*$/);
[InternalError] Regular expression backtrack limit exceeded: (how?)
```

```js
"xyz123xyz".match(/((123)|(xyz)*)*/); // Should be ["xyz123xyz", "xyz", undefined, "xyz"]
```

```js
"xyz".match(/(z?\1)z/); // Should be ["z", ""]
```

```js
"saturday".match(/([^ab]*?)*/); // Should be ["s", "s"]
"sep-12-98".match(/([^ab]*?)*/); // Should be ["sep-12-98", "8"]
```

```js
// This hangs forever now
'hey THIS does match!'.match(/(?<!.*q.*?)(?<=h.*)THIS(?=.*!)/g);
```

```js
'a'.match(/^(a|a?)+$/); // Should be ["a", "a"]
'aaaa'.match(/^(?=(.*\1)+$)[a-z]+$/); // Should be ["aaaa", "aaaa"]
```

[InternalError] Regular expression backtrack limit exceeded:
```js
  ('a'.repeat(25) + 'b').match(/^(a|a?)+$/);
  ('a'.repeat(25)).match(/(a+)+b/);
```

Last two were already fixed by Ali in the previous regex engine and added to [TestRegex.cpp‎](https://github.com/LadybirdBrowser/ladybird/pull/8612/changes#diff-8b85881ce276f3975d2eb429ecd16dda110fafc453a40e7cc5434a84e3571f48), but I see almost all tests from there are deleted now. Is there a reason for that? This file accumulated a lot of good tests over time and could catch these and probably some other regressions.

awesomekling:
Thanks @aplefull for the great testing! Fixed all the issues you brought up + brought back the TestRegex.cpp tests that were relevant for ECMA262!

alimpfard:
Looks great, LGTM.

PR #8613, opened 2026-03-25
LibWeb: Scoped Custom Element Registries

AtkinsSJ:
Because why have one `CustomElementRegistry` when you can have an arbitrary number of them instead?

There are some regressions, some because they were previously false-positives, and some because this spec still isn't quite right in places. (For example, some tests expect an element with a null CER to keep that null when adopted, which is not what the spec tells us to do.) Also, undoubtedly some because of my/our bugs. However, this is a big hairy change, which has been in the works on-and-off for almost a year now, and tends to attract conflicts. So besides any glaring issues, I'd really like to get it merged and then we can iterate on it once it's in-tree. :sweat_smile:


leuasseurfarrelds247-arch:
0ecb02363d7e6dd049ffdd78e5b763e82ff95f91

AtkinsSJ:
Push: Rebase to fix TestConfig.ini conflict. No other changes.

Libraries/LibWeb/HTML/CustomElements/CustomElementRegistry.h
```diff
@@ -40,16 +46,33 @@ class WEB_API CustomElementRegistry : public Bindings::PlatformObject {
...
     virtual void visit_edges(Visitor&) override;
 
-    // Every CustomElementRegistry has a set of custom element definitions, initially empty. In general, algorithms in this specification look up elements in the registry by any of name, local name, or constructor.
+    // https://html.spec.whatwg.org/multipage/custom-elements.html#is-scoped
+    // Every CustomElementRegistry has an is scoped, a boolean, initially false.
+    bool m_is_scoped { false };
+
+    // https://html.spec.whatwg.org/multipage/custom-elements.html#scoped-document-set
+    // Every CustomElementRegistry has a scoped document set, a set of Document objects, initially « ».
+    HashTable<GC::Ref<DOM::Document>> m_scoped_documents;
```
tcl3: No documents ever get removed from this set. Would it be better to use weak references, so that the CustomElementRegistry doesn't keep otherwise destroyed documents alive?

Libraries/LibWeb/HTML/Parser/HTMLParser.cpp
```diff
@@ -1241,11 +1250,15 @@ void HTMLParser::handle_in_head(HTMLToken& token)
...
-                // FIXME: 1. Let registry be null if templateStartTag has a shadowrootcustomelementregistry attribute; otherwise declarativeShadowHostElement's node document's custom element registry.
-
-                // 2. Attach a shadow root with declarativeShadowHostElement, mode, clonable, serializable, delegatesFocus, "named", and registry.
-                // FIXME: and registry.
-                auto result = declarative_shadow_host_element.attach_a_shadow_root(mode, clonable, serializable, delegates_focus, Bindings::SlotAssignmentMode::Named);
+                // 1. Let registry be null if templateStartTag has a shadowrootcustomelementregistry attribute;
+                //    otherwise declarativeShadowHostElement's node document's custom element registry.
+                GC::Ptr<CustomElementRegistry> registry;
+                if (!template_start_tag.has_attribute(AttributeNames::shadowrootcustomelementregistry))
+                    registry = declarative_shadow_host_element.custom_element_registry();
```
tcl3: Should this be `declarative_shadow_host_element.document().custom_element_registry()`?

Libraries/LibWeb/HTML/Parser/HTMLParser.cpp
```diff
@@ -1273,7 +1286,10 @@ void HTMLParser::handle_in_head(HTMLToken& token)
                 // 6. Set shadow's available to element internals to true.
                 shadow.set_available_to_element_internals(true);
 
-                // FIXME: 7. If templateStartTag has a shadowrootcustomelementregistry attribute, then set shadow's keep custom element registry null to true.
+                // 7. If templateStartTag has a shadowrootcustomelementregistry attribute, then set shadow's keep
+                //    custom element registry null to true.
+                if (token.has_attribute(AttributeNames::shadowrootcustomelementregistry))
```
tcl3: Using `template_start_tag` rather than `token` here would match the spec text.

Libraries/LibWeb/DOM/ElementFactory.cpp
```diff
@@ -598,188 +606,199 @@ static GC::Ref<MathML::MathMLElement> create_mathml_element(JS::Realm& realm, Do
...
-    //    3. If namespace is the HTML namespace, and either localName is a valid custom element name or is is non-null,
-    //       then set result’s custom element state to "undefined".
-    if (namespace_ == Namespace::HTML && (HTML::is_valid_custom_element_name(local_name) || is_value.has_value())) {
-        element->set_custom_element_state(CustomElementState::Undefined);
-    }
+    else {
+        // 1. Let interface be the element interface for localName and namespace.
+        // 2. Set result to the result of creating an element internal given document, interface, localName, namespace,
+        //    prefix, "uncustomized", is, and registry.
+        auto qualified_name = QualifiedName { local_name, prefix, namespace_ };
```
tcl3: I don't think this is ever used.

Libraries/LibWeb/HTML/CustomElements/CustomElementRegistry.h
AtkinsSJ: That makes a lot of sense. And I just found a [spec issue](https://github.com/whatwg/html/issues/12092) about it! [Relevant comment:](https://github.com/whatwg/html/issues/12092#issuecomment-3769252677)

> It's a weak set, but we're not very consistent about these things. It might be good to call that out. Maybe for now we just add "weak" in front of set.

Libraries/LibWeb/HTML/Parser/HTMLParser.cpp
AtkinsSJ: Yes it should! Thanks, that fixes a test.

Libraries/LibWeb/DOM/ElementFactory.cpp
AtkinsSJ: Yup, it's a leftover from resolving a merge conflict. Removed!

AtkinsSJ:
Thanks Tim! :^)

Rebased, and corrected things based on Tim's feedback.

Libraries/LibWeb/HTML/CustomElements/CustomElementRegistry.cpp
```diff
@@ -273,46 +286,36 @@ JS::ThrowCompletionOr<void> CustomElementRegistry::define(String const& name, We
...
     //     Finally, if the steps threw an exception, rethrow that exception.
     if (maybe_exception.is_throw_completion())
         return maybe_exception.release_error();
 
-    // 15. Let definition be a new custom element definition with name name, local name localName, constructor constructor,
-    //     observed attributes observedAttributes, lifecycle callbacks lifecycleCallbacks, form-associated formAssociated,
-    //     disable internals disableInternals, and disable shadow disableShadow.
+    // 15. Let definition be a new custom element definition with name name, local name localName, constructor
+    //     constructor, observed attributes observedAttributes, lifecycle callbacks lifecycleCallbacks, form-associated
+    //     formAssociated,  disable internals disableInternals, and disable shadow disableShadow.
```
tcl3: nit: There's a double space here

Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -7455,6 +7471,41 @@ void Document::run_csp_initialization() const
     }
 }
 
+// https://dom.spec.whatwg.org/#flatten-element-creation-options
+WebIDL::ExceptionOr<Document::RegistryAndIs> Document::flatten_element_creation_options(Variant<String, ElementCreationOptions> options) const
```
tcl3: `options` could be passed by const ref here.

AtkinsSJ:
Nits: picked

PR #8615, opened 2026-03-25
LibWeb: Always shift line boxes below floats when content does not fit

tcl3:
Previously, this wasn't done when placing the first inline content  in a block, which caused long unbreakable words to overlap with floats instead of being moved below them. 

This can be seen on this page https://cy.wikipedia.org/wiki/Llanfair_Pwllgwyngyll:

Before:

image

After:

image


tcl3:
My latest push unskips an unrelated `CSS2/floats` test that started passing at some point.

PR #8616, opened 2026-03-25
Tests/LibWeb: Run form-disabled-callback.html from HTTP server

shannonbooth:
Another test improvement found while trying to improve navigation.

This test does not work properly when run from local file. This seems to be due to the use of /common/blank.html in the test resource helper, which would previously not resolve properly.

Other browsers, such as Chrome, also do not work for this WPT test when loaded as file:// scheme.

For some reason I have not fully looked into, this test seems to rely on iframe being processed synchronously, and without this change, the test would time out. With this change (alongside) running iframe processing synchronously, we get one more subtest pass.

(also a small alphabetical reorder in area)


shannonbooth:
Actually I might bundle this with some other changes

Issue #8590, opened 2026-03-23
document.open() crashes with Navigation::current_entry


shannonbooth:
https://github.com/LadybirdBrowser/ladybird/commit/45a87b1dbee317aa461f781fc53bc40c00516da6 appears to have caused a whole lot of crashes on WPT to happen. After some investigation, I think this was a latent bug which was exposed by calling perform_url_and_history_update_steps.

I think the root cause is related to our initial about blank document handling.

Here is an example HTML page that crashes before 45a87b1dbee317aa461f781fc53bc40c00516da6 which I believe has the same root cause as the recent crashes introduced to WPT (NOTE: needs to be loaded over HTTP!)

```html
<!DOCTYPE html>
<script src="../include.js"></script>
<script>
    async function tick() {
        await new Promise(resolve => setTimeout(resolve, 0));
    }

    asyncTest(async done => {
        const iframe = document.createElement("iframe");
        document.body.appendChild(iframe);

        const iframeDocument = iframe.contentDocument;
        const navigation = iframe.contentWindow.navigation;

        iframeDocument.open();
        iframeDocument.write("<!DOCTYPE html><p>first</p>");
        iframeDocument.close();

        println(iframe.contentDocument.body.textContent);
        println(navigation.entries().length);
        println(navigation.currentEntry);

        await tick();

        iframeDocument.open();
        iframeDocument.write("<!DOCTYPE html><p>second</p>");
        iframeDocument.close();

        println(iframe.contentDocument.body.textContent);
        println(navigation.entries().length);
        println(navigation.currentEntry);
        done();
    });
</script>
```

I'm unsure of a proper fix here, though I have found a few different things which work around the crash, but I am not convinced are correct.

rcorsi:
Seems to be in same family of problems as #8269 and #8275

rcorsi:
```
VERIFICATION FAILED: m_current_entry_index != -1 at /home/bobo/projects/myladybird/Libraries/LibWeb/HTML/Navigation.cpp:123
Stack trace (most recent call first):
#0  (inlined)          in Web::HTML::Navigation::current_entry() const at /home/bobo/projects/myladybird/Libraries/LibWeb/HTML/Navigation.cpp:123:5
#1  0x00007fb72c3774b3 in Web::HTML::Navigation::current_entry() const at /home/bobo/projects/myladybird/Libraries/LibWeb/HTML/Navigation.cpp:114:33
[...]

PR #8618, opened 2026-03-25
LibWeb: Some navigation fixes / improvements

shannonbooth:
Findings from investigation of https://github.com/LadybirdBrowser/ladybird/issues/8590 investigating initial about blank.

Removes some AD-HOC code which fixes some WPT tests, unfortunately resulting in shuffling this around to other AD-HOC locations.

I think this still forwards progress though.

### DIFF
--- a/Libraries/LibWeb/DOM/Document.cpp
+++ b/Libraries/LibWeb/DOM/Document.cpp
@@ -263,7 +263,7 @@ static GC::Ref<HTML::BrowsingContext> obtain_a_browsing_context_to_use_for_a_nav
     }
 
     // 10. Let newBrowsingContext be the first return value of creating a new top-level browsing context and document.
-    auto browsing_context_and_document = MUST(HTML::create_a_new_top_level_browsing_context_and_document(browsing_context.page()));
+    auto browsing_context_and_document = HTML::create_a_new_top_level_browsing_context_and_document(browsing_context.page());
     auto new_browsing_context = browsing_context_and_document.browsing_context;
 
     // 11. Let navigationCOOP be navigationParams's cross-origin opener policy.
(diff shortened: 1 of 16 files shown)


Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -3402,11 +3395,24 @@ void Document::completely_finish_loading()
         m_active_refresh_timer->start();
 
     // 3. Let container be document's browsing context's container.
-    if (!navigable->container())
+    // Use the browsing context's current container instead of this document's node navigable.
+    // A newly created replacement document can finish loading before it becomes the active
+    // document of its navigable, in which case document.navigable() is still null here.
```
shannonbooth, line 3454: I suspect this stems from overall document lifecycle issues and the timing of when we call completely finish loading. Not super happy with this change

Libraries/LibWeb/HTML/NavigableContainer.cpp
```diff
@@ -145,9 +145,15 @@ WebIDL::ExceptionOr<void> NavigableContainer::create_new_child_navigable(GC::Ptr
         // 7. Update for navigable creation/destruction given traversable
         traversable->update_for_navigable_creation_or_destruction();
 
-        if (after_session_history_update) {
-            after_session_history_update->function()();
-        }
+        // Delay readiness until the bookkeeping step has fully finished. This keeps initial navigations queued while
+        // the new child is still being installed into the parent document state.
+        traversable->append_session_history_traversal_steps(GC::create_function(traversable->heap(), [navigable] {
```
shannonbooth: I'm not really sure about these changes here or in first_synchronous_navigation_steps_with_target_navigable_not_contained_in, though I think it's an improvement. I suspect that both of these hacks can be removed or tightened up. Mostly, I suspect that these workarounds should be tied specifically to the initial about blank document in particular. Probably also a good thing to keep localized in history steps as well

PR #8626, opened 2026-03-26
LibWeb/CSS: Implement initial parsing for `@container` rules 

AtkinsSJ:
The main limitation here are that none of the container-query features are parsed in a meaningful way; they all become `<general-enclosed>`. Parsing for them will be added as they are implemented.

CSSContainerRule is unusual in how it matches, because instead of it either matching or not matching globally, it instead is matched against a specific element. But also, some at-rules inside it always apply, as if they were written outside it. This doesn't fit well with how CSSConditionRule is implemented, and will likely require some rework later. For now, `condition_matches()` always returns false, and `for_each_effective_rule()` is overridden to always process those global at-rules and nothing else.

### DIFF
--- a/Libraries/LibWeb/CSS/Parser/Parser.cpp
+++ b/Libraries/LibWeb/CSS/Parser/Parser.cpp
@@ -1,7 +1,7 @@
 /*
  * Copyright (c) 2018-2024, Andreas Kling <andreas@ladybird.org>
  * Copyright (c) 2020-2021, the SerenityOS developers.
- * Copyright (c) 2021-2025, Sam Atkins <sam@ladybird.org>
+ * Copyright (c) 2021-2026, Sam Atkins <sam@ladybird.org>
  * Copyright (c) 2021, Tobias Christiansen <tobyase@serenityos.org>
  * Copyright (c) 2022, MacDue <macdue@dueutil.tech>
  * Copyright (c) 2024, Shannon Booth <shannon@serenityos.org>
@@ -21,6 +21,7 @@
 #include <LibWeb/CSS/CSSStyleDeclaration.h>
 #include <LibWeb/CSS/CSSStyleProperties.h>
 #include <LibWeb/CSS/CSSStyleSheet.h>
+#include <LibWeb/CSS/ContainerQuery.h>
 #include <LibWeb/CSS/FontFace.h>
 #include <LibWeb/CSS/MediaList.h>
 #include <LibWeb/CSS/Parser/ErrorReporter.h>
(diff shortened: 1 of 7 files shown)


AtkinsSJ:
Rebased to fix conflicts, solved a Windows build issue by moving CSSContainerRule's destructor definition into its .cpp file, and split the implementation and parsing changes into separate commits.

Libraries/LibWeb/CSS/CSSContainerRule.cpp
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    // 4. Return result.
+    return result.to_string_without_validation();
+}
+
+bool CSSContainerRule::condition_matches() const
+{
+    // NB: @container is processed differently from other CSSConditionRules. As such this is never called.
+    return false;
```
Calme1709: If this is never called should we do a `VERIFY_NOT_REACHED()` instead?

Libraries/LibWeb/CSS/Parser/Parser.cpp
Calme1709: `@container` should be allowed within `@function` since it's a conditional group rule

AtkinsSJ:
Aaand the spec change got merged, so I'm drafting this until I update the comments.

Libraries/LibWeb/CSS/Parser/Parser.cpp
AtkinsSJ: Oops, missed that when fixing the conflicts. Thanks!

AtkinsSJ:
Changes:
- Updated to match the current spec
- Also implemented the `CSSContainerRule.conditions` attribute
- Re-imported the CSSOM test as it now has 3 new subtests which we also pass
- Applied Calme's suggestions

Libraries/LibWeb/CSS/CSSContainerRule.cpp
```diff
@@ -0,0 +1,229 @@
...
+        if (!first)
+            result.append(", "sv);
+
+        // 2. Set first to false.
+        first = false;
+
+        // 3. If condition's name is not empty:
+        if (!condition.name.is_empty()) {
+            // 1. Append condition's name to result.
+            serialize_an_identifier(result, condition.name);
```
Calme1709: We have already serialized this as an identifier within `conditions()`

Calme1709 (changes requested):
One more little thing

Libraries/LibWeb/CSS/CSSContainerRule.cpp
AtkinsSJ: Thanks, I re-checked everything now and I think that was the last remnant from before I added the `conditions` code.

PR #8627, opened 2026-03-26
LibWeb: Fix a number of crashes seen in WPT's `encoding` category of tests

gmta:
Running the `encoding` tests locally produced a number of crashes that I was unable to reproduce after these changes.

* Make sure `close_top_level_traversable()` does not run its steps again when re-entered;
* Take one task at a time from `TaskQueue` so we can guarantee it being runnable / not being permanently unrunnable;
* Add a null guard for the browsing context in sniff bytes callback;
* Add a null guard for a document's navigable in `apply_the_history_step()`.

### DIFF
--- a/Libraries/LibWeb/HTML/TraversableNavigable.cpp
+++ b/Libraries/LibWeb/HTML/TraversableNavigable.cpp
@@ -746,8 +746,12 @@ TraversableNavigable::HistoryStepResult TraversableNavigable::apply_the_history_
         // 6. Let navigable be changingNavigableContinuation's navigable.
         auto navigable = changing_navigable_continuation->navigable;
 
-        // NOTE: This check is not in the spec but we should not continue navigation if navigable has been destroyed.
+        // AD-HOC: We should not continue navigation if navigable has been destroyed.
         if (navigable->has_been_destroyed())
+            continue;
+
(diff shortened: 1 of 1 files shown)


Libraries/LibWeb/HTML/Navigable.cpp
```diff
@@ -1492,6 +1492,10 @@ void Navigable::populate_session_history_entry_document(
                     body->wait_for_sniff_bytes(GC::create_function(heap(),
                         [entry, navigation_params, signal_to_continue_session_history_processing,
                             received_navigation_params, saveExtraDocumentState, completion_steps](ReadonlyBytes sniff_bytes) {
+                            // AD-HOC: The document may have been destroyed between when the fetch started and when the
+                            //         bytes arrived.
+                            if (!navigation_params->navigable->active_browsing_context())
+                                return;
```
tcl3, line 1498: Is it safe to return here without running `finalize_session_history_entry()`? If you don't run the completion steps then `after_document_populated()` is never called. This updates state that we might be spinning on here: https://github.com/LadybirdBrowser/ladybird/blob/b38b7d0ae3ca0b12ae2882fd8983af1a946d57f4/Libraries/LibWeb/HTML/TraversableNavigable.cpp#L697-L703

PR #8630, opened 2026-03-26
LibWeb+Tests: Import and fix CustomElementRegistry WPT

olliehcrook:
Fetch lifecycle callbacks in spec order during
  CustomElementRegistry::define().

  Import the upstream CustomElementRegistry WPT for in-tree coverage.

  Refs: https://wpt.live/custom-elements/CustomElementRegistry.html
  Refs: https://wpt.fyi/results/custom-elements/CustomElementRegistry.html?run_id=5481733335285760

### DIFF
--- a/Libraries/LibWeb/HTML/CustomElements/CustomElementRegistry.cpp
+++ b/Libraries/LibWeb/HTML/CustomElements/CustomElementRegistry.cpp
@@ -210,15 +210,15 @@ JS::ThrowCompletionOr<void> CustomElementRegistry::define(String const& name, We
             return vm.throw_completion<JS::TypeError>(JS::ErrorType::NotAnObject, prototype_value);
 
         // 3. Let lifecycleCallbacks be the ordered map «[ "connectedCallback" → null, "disconnectedCallback" → null,
-        //    "adoptedCallback" → null, "connectedMoveCallback" → null, "attributeChangedCallback" → null ]».
+        //    "connectedMoveCallback" → null, "adoptedCallback" → null, "attributeChangedCallback" → null ]».
         lifecycle_callbacks.set(CustomElementReactionNames::connectedCallback, {});
         lifecycle_callbacks.set(CustomElementReactionNames::disconnectedCallback, {});
-        lifecycle_callbacks.set(CustomElementReactionNames::adoptedCallback, {});
         lifecycle_callbacks.set(CustomElementReactionNames::connectedMoveCallback, {});
+        lifecycle_callbacks.set(CustomElementReactionNames::adoptedCallback, {});
         lifecycle_callbacks.set(CustomElementReactionNames::attributeChangedCallback, {});
 
         // 4. For each callbackName of the keys of lifecycleCallbacks:
-        for (auto const& callback_name : { CustomElementReactionNames::connectedCallback, CustomElementReactionNames::disconnectedCallback, CustomElementReactionNames::adoptedCallback, CustomElementReactionNames::connectedMoveCallback, CustomElementReactionNames::attributeChangedCallback }) {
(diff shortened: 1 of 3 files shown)


Libraries/LibWeb/HTML/CustomElements/CustomElementRegistry.cpp
```diff
@@ -210,15 +210,15 @@ JS::ThrowCompletionOr<void> CustomElementRegistry::define(String const& name, We
             return vm.throw_completion<JS::TypeError>(JS::ErrorType::NotAnObject, prototype_value);
 
         // 3. Let lifecycleCallbacks be the ordered map «[ "connectedCallback" → null, "disconnectedCallback" → null,
-        //    "adoptedCallback" → null, "connectedMoveCallback" → null, "attributeChangedCallback" → null ]».
+        //    "connectedMoveCallback" → null, "adoptedCallback" → null, "attributeChangedCallback" → null ]».
```
AtkinsSJ, line 213: I'm confused, the [current spec](https://html.spec.whatwg.org/multipage/custom-elements.html#dom-customelementregistry-define) has them in the old order.
> Let lifecycleCallbacks be the [ordered map](https://infra.spec.whatwg.org/#ordered-map) «[ "connectedCallback" → null, "disconnectedCallback" → null, "adoptedCallback" → null, "connectedMoveCallback" → null, "attributeChangedCallback" → null ]».

olliehcrook: Yeah this is a bit weird. The implementation does follow the spec, but wpt tests do imply the other order. 

I found that there is an issue open on whatwg that mentions the inconstency here: https://github.com/whatwg/html/issues/12016 and have checked these tests do pass on firefox and chrome (well, firefox does have 1 failure :)).

Not sure what the reason is for this discrepency honestly, I have just been hunting around looking for WPT failures to try to learn the codebase.

(BTW I apologise if the formatting of this comment is horrible, i am away from home and responding on github mobile).

AtkinsSJ: Hehe, welcome to web specs!

Yeah there's a PR to alter the spec [here](https://github.com/whatwg/html/pull/12017), but no firm decision on whether the spec or WPT is wrong. We'll see which way that goes.

Looking for WPT failures is great! You just have to recognise sometimes that WPT and the spec don't align. There are a few ways to resolve that. Usually opening a spec issue is good, if there isn't one already. But our comments should always match the current spec. If you need to do something different, add an `// AD-HOC:` comment explaining why you're doing it differently, and linking to the spec issue if possible. Sometimes we'll implement changes from a spec PR that isn't merged yet, but again link to that.

I don't have a strong opinion here whether to stick with the spec or adjust to match WPT. If you want to keep your changes, you'd just need to 1) restore the spec comments to match the spec, and 2) add a comment similar to `// AD-HOC: We instead put these in the order expected by WPT. Spec issue: https://github.com/whatwg/html/issues/12016` here and maybe a `// AD-HOC: Different order, see above` comment at the other change.

AtkinsSJ: Update on that, the spec PR got merged. So you just need to check that the comments (and code) match what the spec currently says, and you're good. :+1:

olliehcrook: Awesome! I will review and reply/update PR later this evening.

olliehcrook: @AtkinsSJ I have checked against spec changes merged in https://github.com/whatwg/html/pull/12017 and the spec/code/comments do now match :smile:

shannonbooth:
Oh I am so sorry, I did not realize that this was already open with the fix

olliehcrook:
> Oh I am so sorry, I did not realize that this was already open with the fix

No worries, I'll close this :)

PR #8636, opened 2026-03-27
LibWebView: Use http:// for private/loopback IP addresses in URL bar

### LINKED ISSUE #8344: URL bar defaults to https:// for bare IP addresses, breaking local network access

exec:
## Summary
- When a bare IP address is entered without a scheme, `sanitize_url` unconditionally prepends `https://`, breaking local network access
- Detect private (10/8, 172.16/12, 192.168/16), loopback (127/8, ::1), and link-local (169.254/16) addresses and use `http://` instead
- Explicit schemes entered by the user are always respected

## Test plan
- [x] New test cases in `TestWebViewURL.cpp` covering all private/loopback ranges, boundary values, public IPs, and explicit scheme preservation
- [x] All existing `TestWebViewURL` tests continue to pass

Fixes #8344.

### DIFF
--- a/Libraries/LibWebView/URL.cpp
+++ b/Libraries/LibWebView/URL.cpp
@@ -6,13 +6,67 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
+#include <AK/IPv4Address.h>
+#include <AK/IPv6Address.h>
 #include <AK/String.h>
 #include <LibFileSystem/FileSystem.h>
+#include <LibURL/Host.h>
 #include <LibURL/Parser.h>
 #include <LibURL/PublicSuffixData.h>
 #include <LibWebView/URL.h>
 
 namespace WebView {
+
+static bool is_ipv4_private_or_loopback(u8 first, u8 second)
+{
+    // 127.0.0.0/8 (loopback)
+    if (first == 127)
+        return true;
+    // 10.0.0.0/8 (private)
+    if (first == 10)
+        return true;
+    // 172.16.0.0/12 (private)
+    if (first == 172 && second >= 16 && second <= 31)
+        return true;
+    // 192.168.0.0/16 (private)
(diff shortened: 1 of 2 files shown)


exec:
Closing in favor of #8432 which was filed first. Note: the IPv4Address byte-order issue (URL parser stores via NetworkOrdered, so addr[0] returns the wrong octet) may affect that PR as well — we hit it during testing and worked around it via a serialize-then-reparse approach.

PR #8637, opened 2026-03-27
RequestServer: Use macOS system certificates for SSL verification

### LINKED ISSUE #8629: SSL verification failed (with seemingly valid certificate?)
### Summary

When opening URL https://epexspot.com , error page is displaying message "SSL verification failed".

Console shows:
1063350.535 WebContent(50161): ResourceLoader: Failed load of: "https://www.epexspot.com/", Error: Request finished with error: SSL verification failed

### Operating system

macOS

### Steps to reproduce

1. Open https://epexspot.com

### Expected behavior

Show page content.

### Actual behavior

Error shown:
SSL verification failed

n/a

```html
n/a
```

```shell
1063350.535 WebContent(50161): ResourceLoader: Failed load of: "https://www.epexspot.com/", Error: Request finished with error: SSL verification failed
```

_No response_

_No response_

### Contribute a patch?

exec:
## Summary
- On macOS, the vcpkg-built OpenSSL has its OPENSSLDIR pointing to the vcpkg install tree, which contains no CA certificates
- Sites using newer root CAs (e.g. Sectigo R46) that are in the macOS system keychain but not in the default OpenSSL paths fail SSL verification
- Load system root certificates via `SecTrustCopyAnchorCertificates` at RequestServer startup, convert to OpenSSL `X509` objects, and inject into each curl handle's SSL context via `CURLOPT_SSL_CTX_FUNCTION`
- No temporary files are created; certificates are held in memory and injected directly into the OpenSSL trust store

## Test plan
- [x] Verified `https://www.epexspot.com` loads successfully (previously failed with "SSL verification failed")
- [x] Other HTTPS sites continue to work normally

Fixes #8629.

### DIFF
--- a/Services/RequestServer/CMakeLists.txt
+++ b/Services/RequestServer/CMakeLists.txt
@@ -45,6 +45,10 @@ if (WIN32)
     lagom_windows_bin(RequestServer CONSOLE)
 endif()
 
+if (APPLE)
+    target_link_libraries(RequestServer PRIVATE "-framework Security")
+endif()
+
 if (${CMAKE_SYSTEM_NAME} MATCHES "SunOS")
     # Solaris has socket and networking related functions in two extra libraries
     target_link_libraries(requestserverservice PUBLIC nsl socket)
--- a/Services/RequestServer/Request.cpp
+++ b/Services/RequestServer/Request.cpp
@@ -526,6 +526,9 @@ void Request::handle_fetch_state()
     if (auto const& path = default_certificate_path(); !path.is_empty())
         set_option(CURLOPT_CAINFO, path.characters());
 
+    if (auto callback = ssl_ctx_setup_callback())
+        set_option(CURLOPT_SSL_CTX_FUNCTION, callback);
+
     set_option(CURLOPT_ACCEPT_ENCODING, ""); // Empty string lets curl define the accepted encodings.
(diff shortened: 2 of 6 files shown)


UI/AppKit/Application/Application.mm
```diff
@@ -29,10 +31,49 @@
...
+    return temp_path;
+}
+
+void Application::create_platform_options(WebView::BrowserOptions&, WebView::RequestServerOptions& request_server_options, WebView::WebContentOptions&)
 {
     if (m_file_scheme_urls_have_tuple_origins)
         URL::set_file_scheme_urls_have_tuple_origins();
+
+    if (auto cert_path = export_system_certificates(); cert_path.has_value())
+        request_server_options.certificates.append(cert_path.release_value());
```
gmta: Instead of passing around a temporary PEM encoded file with all system certificates, I would expect services that need to read X.509 certificates to do it themselves. Passing in a path to a file would be one supported way to set the supported CA certificates, but by default it should be able to read them directly from the source.

exec: Updated, RequestServer now loads system certificates directly from the macOS keychain via SecTrustCopyAnchorCertificates at startup, converts them to X509 objects, and injects them into each curl handle's SSL context via CURLOPT_SSL_CTX_FUNCTION. No temporary files or PEM encoding involved.

PR #8642, opened 2026-03-27
LibWeb: Fire textarea input events immediately instead of debouncing

### LINKED ISSUE #7793: LibWeb: `<textarea>` input can be dropped if queued input event stores the current value and restores it later
For some, this reproduction of what happens with the Google search bar:
```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <title>Document</title>
</head>
<textarea id="testarea"></textarea>
<body>
    <script>
        (() => {
            let storedValue = "";
            const testarea = document.getElementById("testarea");
            
            testarea.addEventListener("input", () => {
                storedValue = testarea.value;
                requestAnimationFrame(() => {
                    testarea.value = storedValue;
                })
            });
        })();
    </script>
</body>
</html>
```

This seems to be a weird interaction with the input event debounce timer, where if it happens to queue a task and that task runs before new user input is processed, a later task can wipe out the user inputs that were just processed by restoring the value it stored from the input event.

exec:
## Summary
- Remove the 100ms debounce timer on textarea input events
- The debounce delays the `input` DOM event while the text node mutation happens immediately, creating a window where JavaScript rAF callbacks can overwrite new input with stale captured values
- This fixes character dropping on sites like google.com whose search box uses `requestAnimationFrame` to restore input values from previous `input` events
- The spec notes this delay is optional ("User agents may wait"), and HTMLInputElement already fires immediately

## Test plan
- [x] Manually verified: typing fast in the google.com search box no longer drops characters
- [x] Existing tests `input-event-data.html` and `input-event-inputType.html` cover textarea input event firing

Fixes #7793.

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLTextAreaElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLTextAreaElement.cpp
@@ -33,9 +33,6 @@ GC_DEFINE_ALLOCATOR(HTMLTextAreaElement);
 
 HTMLTextAreaElement::HTMLTextAreaElement(DOM::Document& document, DOM::QualifiedName qualified_name)
     : HTMLElement(document, move(qualified_name))
-    , m_input_event_timer(Core::Timer::create_single_shot(0, GC::weak_callback(*this, [](auto& self) {
-        self.queue_firing_input_event();
-    })))
 {
 }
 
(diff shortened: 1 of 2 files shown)


gmta:
Why was the debounce originally introduced? What problem did that solve, and why is that problem still solved now?

exec:
The debounce was introduced in ed1379311 alongside the initial implementation of textarea input events. The commit was fixing GitHub comment submission, textarea value wasn't being synced or firing input events at all. 100ms timer appears to be taken from spec example rather than added to address a problem, commit message doesn't mention it. HTMLInputElement fires input events immediately via queue_an_element_task with no debounce. Expensive work (text node mutation, style invalidation) happens unconditionally on every keystroke regardless of when input event fires, so debounce shouldn't provide a performance benefit.

exec:
Fixed the double blank line that tripped clang-format - leftover from deleted function.

PR #8647, opened 2026-03-27
LibWeb: Disable Navigation API after `document.open()` on `about:blank`

### LINKED ISSUE #8590: document.open() crashes with Navigation::current_entry

tcl3:
This follows the Blink/WebKit approach described in this spec issue: https://github.com/whatwg/html/issues/11811

Fixes [a number of WPT crashes](https://wpt.fyi/results/html/syntax/parsing?q=ladybird%3Acrash&run_id=5081858357592064) in `html/syntax/parsing`.

Fixes #8590

### DIFF
--- a/Libraries/LibWeb/DOM/Document.cpp
+++ b/Libraries/LibWeb/DOM/Document.cpp
@@ -851,6 +851,11 @@ WebIDL::ExceptionOr<Document*> Document::open(Optional<String> const&, Optional<
         HTML::perform_url_and_history_update_steps(*this, move(new_url));
     }
 
+    // AD-HOC: Record that this document was an initial about:blank before document.open() cleared the flag, so the
+    //         Navigation API can keep entries and events disabled.
+    if (is_initial_about_blank())
+        as<HTML::Window>(HTML::relevant_global_object(*this)).navigation()->set_was_initial_about_blank_opened(true);
+
     // 13. Set document's is initial about:blank to false.
     set_is_initial_about_blank(false);
 
--- a/Libraries/LibWeb/HTML/Navigation.cpp
+++ b/Libraries/LibWeb/HTML/Navigation.cpp
@@ -467,6 +467,12 @@ bool Navigation::has_entries_and_events_disabled() const
     if (document.origin().is_opaque())
         return true;
 
+    // AD-HOC: Keep entries and events disabled when document.open() is called on an initial about:blank document. This
+    //         behavior matches other engines.
+    //         Spec issue: https://github.com/whatwg/html/issues/11811
+    if (m_was_initial_about_blank_opened)
+        return true;
+
     // 5. Return false.
     return false;
 }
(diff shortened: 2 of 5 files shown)


Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -851,6 +851,11 @@ WebIDL::ExceptionOr<Document*> Document::open(Optional<String> const&, Optional<
         HTML::perform_url_and_history_update_steps(*this, move(new_url));
     }
 
+    // AD-HOC: Record that this document was an initial about:blank before document.open() cleared the flag, so the
+    //         Navigation API can keep entries and events disabled.
+    if (is_initial_about_blank())
+        as<HTML::Window>(HTML::relevant_global_object(*this)).navigation()->set_was_initial_about_blank_opened(true);
```
shannonbooth, line 857: I suppose this fixes #11811.

One question which I was not sure about here though with this though is when / if ever does the document have this flag unset

shannonbooth: Also i think you mean blink/gecko, though maybe webkit has adopted the same

shannonbooth: actually i have fully misread!

tcl3: > One question which I was not sure about here though with this though is when / if ever does the document have this flag unset

Good point. I've made sure the flag is reset when a new navigation occurs.

Libraries/LibWeb/HTML/Navigation.cpp
```diff
@@ -1430,6 +1436,10 @@ void Navigation::initialize_the_navigation_api_entries_for_a_new_document(Vector
     // 2. Assert: navigation's current entry index is −1.
     VERIFY(m_current_entry_index == -1);
 
+    // AD-HOC: This flag is only set when an about:blank page is modified by document.open(). Reset it now that a new
+    //         navigation is happening.
+    m_was_initial_about_blank_opened = false;
```
shannonbooth, line 1441: I've been trying to come up with a test that demonstrates this change, but it is not triggered. Maybe this is not observable if only invoked for a new document since initial about blank document is meant to be replaced?

awesomekling:
Merging to clean up WPT crashes

PR #8649, opened 2026-03-27
LibWeb: Fix percentage heights for elements inside a button

### LINKED ISSUE #6824: Percentage heights for elements inside a button don't work
Reduced from https://www.lego.com/en-gb/pick-and-build/pick-a-brick?query=3005&selectedElement=6567035

```html
<!doctype html>
<style>
button {
    width: 100px;
    height: 100px;
}
#outer {
    width: 100px;
    height: 100px;
    display: flex;
    justify-content: center;
    flex-direction: column;
}
.inner {
    background: red;
    width: 100%;
    height: 100%;
    outline: 1px solid blue;
}
</style>
<button><div class="inner"></div></button>
<hr/>
<div id="outer"><div><div class="inner"></div></div></div>
```

In this example, the divs below the `<hr/>` are a reconstruction of what we do for the contents of the button above. The result in Ladybird is:

Image

However, in other browsers, the `height: 100%` of `.inner` in the button does get the button's height:

Image

As you can see, our flex code is correct (matching other browsers), it just doesn't produce the correct result for button layouts.

I tried a couple of simple fixes, of setting `height: 100%` on the inner wrapper, or getting rid of the inner wrapper altogether, but those cause other regressions.

LeoLNeves:
Problem:
When a <button> contains block-level elements an anonymous content box wrapper is created. Previously this wrapper lacked an explicit height or width causing percentage-based children to collapse to 0.

Solution:
This patch explicitly sets height 100% and width 100% on this wrapper. The issue mentioned that unconditionally setting height 100% caused regressions. To prevent this, this patch specifically wraps the fix in an if (!parent.children_are_inline()) check. This successfully passes percentage dimensions down to block-level children while leaving standard text buttons entirely unaffected, preserving the wrapper's Block Formatting Context.

Testing:
Added the exact HTML file from the issue as a layout test.
Rebaselined one existing layout test.

Fixes #6824

### DIFF
--- a/Libraries/LibWeb/Layout/FormattingContext.cpp
+++ b/Libraries/LibWeb/Layout/FormattingContext.cpp
@@ -1898,7 +1898,52 @@ CSSPixels FormattingContext::calculate_inner_height(Box const& box, AvailableSpa
     // NOTE: We only do this when available space height is indefinite. If it's definite,
     //       we trust that the caller has set it up correctly (e.g., grid/flex items get
     //       their cell/area size as available space).
-    if (height.contains_percentage() && available_space.height.is_indefinite()) {
+
+    // AD-HOC: For percentage heights inside button content wrappers, resolve against the button itself.
+    // During Phase 1 layout (intrinsic sizing), the button's final height is not yet in layout memory.
+    // To prevent the wrapper from collapsing to 0px, we bypass the layout memory and extract the
+    // explicit CSS height directly, manually calculating the true inner border-box dimensions.
+    bool resolved_against_button = false;
+    if (height.contains_percentage() && !available_space.height.is_definite()) {
+        auto cb = box.containing_block();
+        while (cb && cb->is_anonymous() && !cb->display().is_table_cell()) {
+            cb = cb->containing_block();
+        }
+
+        if (cb) {
+            if (auto const* html_element = as_if<HTML::HTMLElement>(cb->dom_node())) {
+                if (html_element->uses_button_layout()) {
+                    if (auto const* used_values = m_state.try_get(*cb)) {
+                        if (used_values->has_definite_height()) {
+                            height_of_containing_block = used_values->content_height();
+                            resolved_against_button = true;
+                        }
+                    }
+                    if (!resolved_against_button) {
+                        auto const& button_computed_height = cb->computed_values().height();
+                        if (button_computed_height.is_length() && !button_computed_height.is_auto()) {
+                            height_of_containing_block = button_computed_height.to_px(*cb, 0);
+
+                            // Safely adjust for box-sizing without relying on an empty m_state
+                            auto const& cb_computed = cb->computed_values();
+                            if (cb_computed.box_sizing() == CSS::BoxSizing::BorderBox) {
+                                height_of_containing_block -= cb_computed.border_top().width;
+                                height_of_containing_block -= cb_computed.border_bottom().width;
+
+                                auto cb_width = available_space.width.to_px_or_zero();
+                                height_of_containing_block -= cb_computed.padding().top().to_px_or_zero(*cb, cb_width);
+                                height_of_containing_block -= cb_computed.padding().bottom().to_px_or_zero(*cb, cb_width);
+                            }
+                            height_of_containing_block = max(CSSPixels(0), height_of_containing_block);
+                            resolved_against_button = true;
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    if (!resolved_against_button && height.contains_percentage() && available_space.height.is_indefinite()) {
         auto containing_block = box.containing_block();
         while (containing_block && containing_block->is_anonymous() && !containing_block->display().is_table_cell())
             containing_block = containing_block->containing_block();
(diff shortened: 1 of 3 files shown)


Libraries/LibWeb/Layout/TreeBuilder.cpp
```diff
@@ -847,6 +847,15 @@ void TreeBuilder::wrap_in_button_layout_tree_if_needed(DOM::Node& dom_node, GC::
         auto content_box_wrapper = parent.create_anonymous_wrapper();
         content_box_wrapper->set_children_are_inline(parent.children_are_inline());
 
+        // If the button contains block-level elements, ensure the content wrapper
```
Psychpsyo: The other comments here come straight from the spec, so this should probably be `// NB: If the button...`

LeoLNeves: Done. Thanks for the review

Tests/LibWeb/Layout/expected/block-and-inline/button-should-have-vertically-aligned-content.txt
```diff
@@ -1,39 +1,39 @@
 Viewport <#document> at [0,0] [0+0+0 800 0+0+0] [0+0+0 600 0+0+0] [BFC] children: not-inline
   BlockContainer <html> at [0,0] [0+0+0 800 0+0+0] [0+0+0 216 0+0+0] [BFC] children: not-inline
     BlockContainer <body> at [8,8] [8+0+0 784 0+0+8] [8+0+0 200 0+0+8] children: inline
-      frag 0 from BlockContainer start: 0, length: 0, rect: [22,19 48.6875x178] baseline: 82.796875
+      frag 0 from BlockContainer start: 0, length: 0, rect: [22,19 48.6875x178] baseline: 72.796875
```
AtkinsSJ: This is a regression, so your fix seems to not be quite right.

Without your changes:
Image

With your changes:
Image

Firefox:
Image

Chromium:
Image

LeoLNeves: Thanks so much @AtkinsSJ! This is my first contribution to Ladybird btw.

I spent some time digging into the architecture to find a way to fix the bug without breaking the vertical alignment. I initially tried patching FormattingContext::calculate_inner_height to make percentage-based children resolve directly against the grandparent <button> instead of the anonymous wrapper. While that got the final height right, it desynchronized with Flexbox's intrinsic sizing phase, causing the content to overflow the bottom of the button.

de-sync-red-square

So I went back to TreeBuilder.cpp but made the fix much more specific. Instead of unconditionally forcing every non-inline wrapper to 100% (which broke the centering of normal text), the code now scans the button's children. It only forces the wrapper to 100% if a child explicitly requests a percentage-based dimension.

This allows standard text blocks to keep their auto wrapper and stay vertically centered (which means the button-should-have-vertically-aligned-content test passes perfectly again), while still fixing the percentage collapse for Issue #6824.

What do you think of this approach?

AtkinsSJ: I am not a good person to ask about layout, unfortunately! :sweat_smile:

Psychpsyo: > So I went back to TreeBuilder.cpp but made the fix much more specific. Instead of unconditionally forcing every non-inline wrapper to 100% (which broke the centering of normal text), the code now scans the button's children. It only forces the wrapper to 100% if a child explicitly requests a percentage-based dimension.
> 
> This allows standard text blocks to keep their auto wrapper and stay vertically centered (which means the button-should-have-vertically-aligned-content test passes perfectly again), while still fixing the percentage collapse for Issue #6824.
> 
> What do you think of this approach?

Would that not fail in a case like this?

```html
<!DOCTYPE html>
<style type="text/css">
    .border-black {
        border: 10px solid black;
    }

    button {
        height: 200px;
    }
</style>

<body>
    <button class="button border-black">
        <div class="border-black">
            one
        </div>
        <div class="border-black" style="height: 100px; width: 100%">
            two
        </div>
```

i.e. exactly the same as `button-should-have-vertically-aligned-content`, except that it also uses a percentage size.

LeoLNeves: Hi, @Psychpsyo. You are right. The TreeBuilder approach fails on cases like that one.

I've discarded the TreeBuilder modifications entirely and digged a bit deeper into the bug. The root cause is a flexbox desync during the intrinsic sizing phase (Phase 1). The anonymous wrapper shrinks to 0px, causing the button's internal flex layout to lock in a permanent 50% vertical displacement before the child's percentage height can even be calculated.

This updated patch fixes the issue purely within FormattingContext.cpp by keeping Phase 1 math in sync:

should_treat_height_as_auto: We look up the tree. If the grandparent is a <button> with an explicit CSS height, we safely bypass the max-content and empty-memory panics.

calculate_inner_height: During Phase 1, we bypass the empty layout state and read the explicit height directly from the button's computed_values. We also manually subtract the computed padding and borders to account for the button's box-sizing: border-box.

By providing the true inner pixel height during Phase 1, the wrapper correctly sizes itself. This prevents the percentage collapse and forces the flex alignment to center perfectly without any vertical displacement.

I've pushed the new approach. Let me know what you think!

Psychpsyo: > I've pushed the new approach. Let me know what you think!

This seems to be a lot more in a better direction.
My only points would be: (and these might be moot, I'm not a maintainer and just took a brief look)
1. Calling the max-content stuff a 'panic' feels weird. It seems to just be a spec requirement? Similarly, `containing_block_used_values->has_definite_height()` has nothing to do with memory, so the comment about avoiding a memory panic feels inaccurate.
2. The control flow in `should_treat_height_as_auto` seems quite different now. Are you sure this doesn't change the behavior for random other setups? Why does it have an exception specifically for table cells? Will it need other random exceptions in the future?
3. I suspect that the root cause of all of this is that buttons should not be using flex layout at all. Instead they should be using button layout: https://html.spec.whatwg.org/multipage/rendering.html#button-layout-2

Issue #4755, opened 2025-05-15
CMake/Infra Wishlist


ADKaster:
Here's a list of things I'd like to be done to the build, loosely sorted by difficulty

Single-Commit to Single-PR Striaghtfoward:

- [x] Build utilities  from Utilities/CMakeLists.txt
- [x] Move lagom_foo cmake functions to their own file
- [ ] Remove serenity_ prefix from options and functions
- [ ] Remove fuzzers that don't work anymore
- [ ] Get Web/CSS parser fuzzers working again
- [x] Remove distinction between "common" and "lagom" options and flags files
- [ ] Move installation of js repl into its own cmake file. Note that this must keep working. we expose those artifacts for use by esvu, which is useful to TC39 delegates
- [x] move all compile_ipc invocations to the client library CMakeLists.txt
- [ ] all file(GLOB) must be destroyed

--- Below this line, ask on discord before starting work ----

Moderately Complex PR:

- [ ] Get ladybird-specifc fuzzers onto oss-fuzz (serenity libs are already on oss-fuzz, but we use a bazillion more deps now)
- [ ] actually make use of the file check_for_dependencies.cmake that was added 4 years ago to check for our dependencies in a centralized place
- [ ] stop spamming the configure log with found FooDependency: /some/giant/path 100 times for each dependency

Giant refactor:

- [ ] Move all files in Lagom/ to top level (well except fuzzers, they should go... somewhere)
- [x] Remove Lagom/CMakeLists.txt entirely, and control everything from top level cmake lists (include(libraries), include(utilities), if (ENABLE_FUZZING) include(Contrib/Fuzzers) or something, etc)
- [ ] rewrite all code generators in python to remove "host build"/"tools build" entirely (difficulty: nightmare)
- [x] move ladybird.sh to ladybird.py (wandows anyone)
- [ ] move WPT.sh to WPT.py

Untenable/Blocked:

- [ ] replace embed_as_string_view.py with #embed (needs gcc 15)

ChaseKnowlden:
Remove serenity_ prefix from options and functions has been fixed in PRs #5249 and #5349.

ChaseKnowlden:
Move lagom_foo cmake functions to their own file has been fixed in PR #5249.

shannonbooth:
https://github.com/LadybirdBrowser/ladybird/pull/9941 will finally check off the no more host tools build point

PR #8650, opened 2026-03-27
Meta: Refactor some of the CMake build system to move away from SerenityOS

cqundefine:
This is mostly focused on moving away from the split between normal build and Lagom build which is not present ever since forking SerenityOS. While at it throw a few random cleanups I found along the way.

CC @ADKaster since this uses some of your code and addresses parts of #4755.


ADKaster:
Looking purely at the commit messages these changes make sense, but I'll take a closer look over the weekend

Documentation/AdvancedBuildInstructions.md
```diff
@@ -147,10 +147,10 @@ If you do run into such an error, the rest of this section explains how to deal
...
       ```diff
       $ patch -p1 <<EOF
-      diff --git a/Meta/CMake/lagom_compile_options.cmake b/Meta/CMake/lagom_compile_options.cmake
+      diff --git a/Meta/CMake/compile_options.cmake b/Meta/CMake/compile_options.cmake
       index 7fec47ac843..45c3af87493 100644
-      --- a/Meta/CMake/lagom_compile_options.cmake
-      +++ b/Meta/CMake/lagom_compile_options.cmake
+      --- a/Meta/CMake/compile_options.cmake
+      +++ b/Meta/CMake/compile_options.cmake
       @@ -29,7 +29,7 @@ if (CMAKE_BUILD_TYPE STREQUAL "Debug")
```
ADKaster, line 154: the rest of the documentation's patch/diff information for "forcing debug info" is likely already out of date, but will be even more so with the changes in this patch. reproducing it and copy pasting the entire command would probably be more useful to people using this section.

Meta/CMake/compile_options.cmake
```diff
@@ -90,6 +94,8 @@ if (NOT MSVC)
     add_cxx_compile_options(-ffp-contract=off)
     add_cxx_compile_options(-fstrict-flex-arrays=2)
     add_cxx_compile_options(-fstack-protector-strong)
+    add_cxx_compile_options(-fsigned-char)
```
ADKaster: this reminds me. we should probably stop setting this argument. iirc in serenity there was a patch recently to stop this that had a few AK changes and a few... LibXML changes? Since we're using a third party dep for xml now, it would probably be simpler.

CMakeLists.txt
```diff
@@ -36,17 +36,207 @@ endif()
...
+if (POLICY CMP0174)
+    cmake_policy(SET CMP0174 NEW)
+endif()
+
 set(LADYBIRD_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}")
 list(APPEND CMAKE_MODULE_PATH "${LADYBIRD_SOURCE_DIR}/Meta/CMake")
 
-include(UI/cmake/EnableLagom.cmake)
+if(NOT COMMAND ladybird_option)
+    macro(ladybird_option)
```
ADKaster, line 52: in the future we should ditch this. it was created for the SerenityOS SuperBuild configuration, which is beyond obsolete in ladybird. not sure if I mentioned this in my issue.

.github/workflows/lagom-template.yml
```diff
@@ -146,7 +146,7 @@ jobs:
           VCPKG_CACHE_SAS: ${{ github.ref == 'refs/heads/master' && secrets.VCPKG_CACHE_SAS || '' }}
           VCPKG_CACHE_MODE: ${{ github.ref == 'refs/heads/master' && 'write' || '' }}
         run: |
-          cmake --preset=Distribution -S Meta/Lagom -B ${{ github.workspace }}/Build/tools-build \
```
ADKaster, line 149: what about BuildFuzzers.sh (Meta/Lagom/)  and BuildLagomTools.sh (UI/Android) ?

CMakeLists.txt
```diff
@@ -36,17 +36,207 @@ endif()
...
+# These are here to support Fuzzili builds further down the directory stack
+set(ORIGINAL_CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS}")
+set(ORIGINAL_CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS}")
+set(ORIGINAL_CMAKE_MODULE_LINKER_FLAGS "${CMAKE_MODULE_LINKER_FLAGS}")
+
+set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${LINKER_FLAGS}")
+set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} ${LINKER_FLAGS}")
+set(CMAKE_MODULE_LINKER_FLAGS "${CMAKE_MODULE_LINKER_FLAGS} ${LINKER_FLAGS}")
+
+configure_file(AK/Debug.h.in AK/Debug.h @ONLY)
```
ADKaster: we could do this in AK/CMakeLists

CMakeLists.txt
```diff
@@ -36,17 +36,207 @@ endif()
 
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
+# We need to find OpenSSL in order to link it explicitly with all targets.
+find_package(OpenSSL REQUIRED)
```
ADKaster, line 40: I'm skeptical about this old find package and the one for threads. future work though. we should group all top level find_packages.

There's still that old empty check_for_dependencies.cmake file that we could use for this kinda thing.

CMakeLists.txt
```diff
@@ -36,17 +36,207 @@ endif()
...
+compile_ipc(${LADYBIRD_SOURCE_DIR}/Services/WebContent/WebUIServer.ipc Services/WebContent/WebUIServerEndpoint.h)
+
+add_subdirectory(Libraries)
+
+if (ENABLE_FUZZERS)
+    add_subdirectory(Meta/Lagom/Fuzzers)
+endif()
+
+# No utilities or tests in these configs
+if (ENABLE_FUZZERS OR ENABLE_COMPILER_EXPLORER_BUILD OR ANDROID OR IOS)
```
ADKaster: this is odd. should it not be fuzzers, compiler explorer, or cross-compiling? We still need UI/Android in the Android build. This feels like a mix of lagom cmake lists goof. In Lagom, yes, no utilities or tests. but we still need gui targets and services.

perhaps some compilable version of:

```
if (NOT (ENABLE_FUZZERS OR ENABLE_COMPILER_EXPLORER_BUILD OR ANDROID OR IOS))
   utilities
   tests
endif()
```

CMakeLists.txt
cqundefine: No it's actually not there in that issue, I actually plan to do this in another PR as I don't want to do to much stuff in one PR.

CMakeLists.txt
cqundefine: Same as with `ladybird_option`, I do wanna tackle moving stuff to the `check_for_dependencies.cmake` file, but I don't to have one massive PR.

cqundefine:
Pushed one more commit, as it fixes LibJS on FreeBSD and it fits the theme of this PR anyways.

ADKaster:
if the CI keeps complaining (arm64 had one CSS failure, >5 timeouts for each linux job) I'd recommend moving the signed char change to its own PR.

Meta/Lagom/BuildFuzzers.sh
ADKaster: the 3 cmake invocations for the fuzzer build itself below also need the -S argument

spholz:
For removing -fsigned-char, have a look at https://github.com/SerenityOS/serenity/pull/26597 and https://github.com/SerenityOS/serenity/pull/26671.
Cherry-picking these commits should hopefully not be that hard.

spholz:
Even if ladybird doesn't want to remove -fsigned-char, these AK changes might be worth cherry-picking.

I'm not sure if `char` is also unsigned on AArch64 Windows and if it can be forced to be signed. If this isn't possible, then removing `-fsigned-char` definitely seems like the right choice.

ADKaster:
@spholz the commit `AK: Properly handle both signed and unsigned chars` appears to be cherry-picked from serenity/26597, with you as co-author. I'm not sure if there's LibWeb places that might be affected though.

spholz:
> @spholz the commit `AK: Properly handle both signed and unsigned chars` appears to be cherry-picked from serenity/26597, with you as co-author. I'm not sure if there's LibWeb places that might be affected though.

Ah that only seems to be part of my PRs though. I would expect more fixes to be necessary, especially the concept changes.

AK/Format.cpp
```diff
@@ -1003,8 +1003,13 @@ ErrorOr<void> Formatter<char>::format(FormatBuilder& builder, char value)
 {
     if (m_mode == Mode::Binary || m_mode == Mode::BinaryUppercase || m_mode == Mode::Decimal || m_mode == Mode::Octal || m_mode == Mode::Hexadecimal || m_mode == Mode::HexadecimalUppercase) {
         // Trick: signed char != char. (Sometimes weird features are actually helpful.)
-        Formatter<signed char> formatter { *this };
-        return formatter.format(builder, static_cast<signed char>(value));
+        if constexpr (IsSigned<char>) {
```
spholz: This won't work if you don't cherry-pick my `IsSigned` and `IsUnsigned` patches.

`IsSigned<char>` is currently broken.

spholz: `char` is neither the same type as `signed char` nor `unsigned char`, so the current implementation of `IsSigned<T>` as `IsSame<T, MakeSigned<T>>` doesn't work work, since it's never true.

cqundefine:
For now, I just dropped the `-fsigned-char` change, I'll do it in another PR where I'll probably just cherry pick those from the Serenity side. FWIW I actually didn't realize the rest of those changes we're relevant to this 😅.

ADKaster:
I think with the first commit from this PR, workarounds like

https://github.com/LadybirdBrowser/ladybird/blob/b85f2c8b89ebd938f1bd31d26f86841abe596bbe/Services/RequestServer/CMakeLists.txt#L1-L3

can be removed in a follow-up.

PR #8651, opened 2026-03-27
LibJS: Initialize constants in execution context right away

InvalidUsernameException:
Followup to #8623 (in particular implementing the suggestion from https://github.com/LadybirdBrowser/ladybird/pull/8623#issuecomment-4136515992). See commits for details.

### DIFF
--- a/Libraries/LibJS/Bytecode/Interpreter.cpp
+++ b/Libraries/LibJS/Bytecode/Interpreter.cpp
@@ -308,14 +308,8 @@ ExecutionContext* Interpreter::push_inline_frame(
     //     in cross-realm calls (e.g. iframe <-> parent).
     callee_context->executable = callee_executable;
 
-    // Copy constants (memcpy avoids aliasing issues with the scalar loop).
-    auto* values = callee_context->registers_and_constants_and_locals_and_arguments();
-    if (auto count = callee_executable.constants.size())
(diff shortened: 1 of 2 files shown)


Libraries/LibJS/Bytecode/Interpreter.cpp
```diff
@@ -121,16 +121,16 @@ ThrowCompletionOr<Value> Interpreter::run(Script& script_record, GC::Ptr<Environ
         executable->dump();
 
     u32 registers_and_locals_count = 0;
-    u32 constants_count = 0;
+    Vector<Value> constants;
```
awesomekling: These vectors are all gonna `malloc()` when you put stuff in them. That seems even worse than what we had before.

Can we find a solution that doesn't use heap-allocating storage?

InvalidUsernameException: Anything specific you would prefer? I used to have a `Optional<Vector<Value> const&>` in an earlier version of the code, but that got annoying to deal with. So maybe a `ReadOnlySpan<Value>`?

awesomekling: Yeah `ReadonlySpan<Value>` sounds like a good choice :)

PR #8653, opened 2026-03-28
LibWeb: Some ASF parsing improvements

Calme1709:
The main change here is validating ASF argument grammar at parse time. We also disallow ASFs in a couple of places where they shouldn't be allowed.


AtkinsSJ:
There are some failing screenshot tests. Unfortunately CI didn't upload them. (See #8683.) If you don't see those locally then you should be able to rebase and push once that fix is merged and hopefully they'll show up again.

AtkinsSJ:
Actually it seems like it's just flakiness of screenshot tests on the Linux/arm64 configuration, see #8688. I think we're good to go here as soon as CI is happy.

PR #8654, opened 2026-03-28
LibWeb: Establish formatting context due to container-type

Psychpsyo:


Psychpsyo:
Believe it or not, all the test failures seem unrelated

PR #8655, opened 2026-03-28
LibWeb+LibMedia: Implement MSE with WebM/VP9/Opus support

Zaggy1024:
Lots of commits here, but an overview of the architecture is:

- `MediaSource` creates and owns some `SourceBuffer`s
- Each `SourceBuffer` creates its own `SourceBufferProcessor` (a non-spec class), which contains the algorithms that we may run off the main thread in the future.
- `SourceBufferProcessor` delegates byte stream parsing to an abstract `ByteStreamParser` class to implement the steps defined by the byte stream format specifications.
- `SourceBufferProcessor` creates a `TrackBuffer` for each track it finds in the byte stream.
- `TrackBuffer`s wrap `TrackBufferDemuxer`s (also non-spec), which hold the actual coded frame data and implement the `Demuxer` interface for our playback system. Seeking, blocking for missing data, and tracking buffered ranges is all handled here.

There were quite a few prerequisite (or at least beneficial) refactors for this to work:
- Exposed a bunch of Matroska parsing functionality for the WebM byte stream parser, fixed some pre-existing issues with it, and added some tests for the basic data-reading functions.
- Allowed `PlaybackManager` to add tracks from multiple `Demuxer` instances instead of only accepting and reading from a single `MediaStream`.
- Implemented some plumbing for buffered time ranges. Currently, remote media resources are always treated as fully buffered, but MSE provides accurate buffered ranges, since they are non-optional for functional MSE playback.

There's probably some other changes I've missed here, see individual commits for details.

Obligatory 4K playback demo:

[mse.webm](https://github.com/user-attachments/assets/03ed5428-5e08-466d-8faa-37c06a685ba2)

### DIFF
--- a/Libraries/LibMedia/Audio/PlaybackStreamAudioUnit.cpp
+++ b/Libraries/LibMedia/Audio/PlaybackStreamAudioUnit.cpp
@@ -327,6 +327,8 @@ NonnullRefPtr<PlaybackStream::CreatePromise> PlaybackStream::create(OutputState 
 NonnullRefPtr<PlaybackStream::CreatePromise> PlaybackStreamAudioUnit::create(OutputState initial_output_state, u32, AudioDataRequestCallback&& data_request_callback)
 {
     auto promise = CreatePromise::construct();
+    // FIXME: Create the AudioState off this thread. It sets up the audio output synchronously, which can take up to
+    //        50ms under normal circumstances.
     auto state_or_error = AudioState::create(move(data_request_callback), initial_output_state);
     if (state_or_error.is_error()) {
         promise->reject(state_or_error.release_error());
--- a/Libraries/LibMedia/PlaybackManager.cpp
+++ b/Libraries/LibMedia/PlaybackManager.cpp
@@ -113,7 +113,7 @@ DecoderErrorOr<void> PlaybackManager::prepare_playback_from_demuxer(WeakPlayback
 
         self->set_up_data_providers();
 
-        if (!self->m_audio_sink && !self->m_audio_tracks.is_empty()) {
+        if (!self->m_audio_output_disabled && !self->m_audio_sink && !self->m_audio_tracks.is_empty()) {
             self->m_audio_sink = MUST(AudioMixingSink::try_create());
             self->set_time_provider(make_ref_counted<WrapperTimeProvider<AudioMixingSink>>(*self->m_audio_sink));
             self->m_audio_sink->on_audio_output_error = [self](Error&& error) {
--- a/Libraries/LibMedia/PlaybackManager.h
+++ b/Libraries/LibMedia/PlaybackManager.h
@@ -53,6 +53,8 @@ public:
     static NonnullOwnPtr<PlaybackManager> create();
     ~PlaybackManager();
 
+    void set_audio_output_disabled(bool disabled) { m_audio_output_disabled = disabled; }
+
     AK::Duration duration() const { return m_duration; }
     void set_duration(AK::Duration duration) { m_duration = duration; }
     AK::Duration current_time() const { return min(m_time_provider->current_time(), duration()); }
@@ -157,6 +159,8 @@ private:
     NonnullRefPtr<WeakPlaybackManagerLink> m_weak_link;
 
     NonnullRefPtr<MediaTimeProvider> m_time_provider;
+
+    bool m_audio_output_disabled { false };
 
     VideoTracks m_video_tracks;
     VideoTrackDatas m_video_track_datas;
--- a/Libraries/LibWeb/HTML/HTMLMediaElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLMediaElement.cpp
@@ -1751,6 +1751,7 @@ void HTMLMediaElement::on_metadata_parsed()
 void HTMLMediaElement::set_up_playback_manager_for_remote()
 {
     m_playback_manager = Media::PlaybackManager::create();
+    m_playback_manager->set_audio_output_disabled(document().page().client().is_headless());
 
     m_has_enabled_preferred_audio_track = false;
     m_has_selected_preferred_video_track = false;
@@ -1816,6 +1817,7 @@ void HTMLMediaElement::set_up_playback_manager_for_remote()
 void HTMLMediaElement::set_up_playback_manager_for_local()
 {
     m_playback_manager = Media::PlaybackManager::create();
+    m_playback_manager->set_audio_output_disabled(document().page().client().is_headless());
 
     m_has_enabled_preferred_audio_track = false;
     m_has_selected_preferred_video_track = false;


Libraries/LibWeb/MediaSourceExtensions/WebMByteStreamParser.h
```diff
@@ -0,0 +1,79 @@
...
+
+    Vector<Media::Track> m_video_tracks;
+    Vector<Media::Track> m_audio_tracks;
+    Vector<Media::Track> m_text_tracks;
+
+    struct MediaSegmentParsingData {
+        AK::Duration timecode;
+        Optional<size_t> remaining_bytes;
+        HashTable<u64> seen_track_numbers {};
+        AK::Duration last_block_timestamp { AK::Duration::min() };
```
tcl3, line 72: This is read from but doesn't appear to be updated anywhere.

Libraries/LibWeb/MediaSourceExtensions/TrackBufferDemuxer.h
```diff
@@ -0,0 +1,77 @@
...
+    virtual ~TrackBufferDemuxer() override;
+
+    Media::Track const& track() const { return m_track; }
+
+    Media::TimeRanges track_buffer_ranges() const;
+
+    void add_coded_frame(Media::CodedFrame);
+    void remove_coded_frames_and_dependants_in_range(AK::Duration start, AK::Duration end);
+
+    void set_reached_end_of_stream() { m_reached_end_of_stream = true; }
```
tcl3: Do we need to acquire `m_mutex` and call `m_data_changed.broadcast()` here? Otherwise, any threads waiting on `m_data_changed` are never going to wake to observe the flag change.

Libraries/LibWeb/MediaSourceExtensions/MediaSource.cpp
```diff
@@ -80,13 +346,37 @@ bool MediaSource::is_type_supported(JS::VM&, String const& type)
...
+    // 4. If type contains a codec that the MediaSource does not support, then return false.
+    // 5. If the MediaSource does not support the specified combination of media type, media
     //    subtype, and codecs then return false.
+    auto codecs_iter = mime_type->parameters().find("codecs"sv);
+    if (codecs_iter == mime_type->parameters().end())
+        return false;
+    auto codecs = codecs_iter->value.bytes_as_string_view();
+    auto had_unsupported_codec = false;
+    codecs.for_each_split_view(',', SplitBehavior::Nothing, [&](auto const& codec) {
+        if (!codec.starts_with("vp9"sv) && !codec.starts_with("vp09"sv) && !codec.starts_with("opus"sv)) {
```
tcl3, line 372: Is `starts_with()` the correct check here? It seems overly broad.

Libraries/LibWeb/MediaSourceExtensions/MediaSource.cpp
```diff
@@ -80,13 +346,37 @@ bool MediaSource::is_type_supported(JS::VM&, String const& type)
...
 
-    // FIXME: 5. If the MediaSource does not support the specified combination of media type, media
+    // 4. If type contains a codec that the MediaSource does not support, then return false.
+    // 5. If the MediaSource does not support the specified combination of media type, media
     //    subtype, and codecs then return false.
+    auto codecs_iter = mime_type->parameters().find("codecs"sv);
+    if (codecs_iter == mime_type->parameters().end())
+        return false;
+    auto codecs = codecs_iter->value.bytes_as_string_view();
+    auto had_unsupported_codec = false;
```
tcl3, line 370: Should we return `false` if there is no `codecs` parameter?

Libraries/LibWeb/MediaSourceExtensions/TrackBufferDemuxer.cpp
```diff
@@ -0,0 +1,263 @@
...
+
+    // Insert in sorted order by presentation timestamp using binary search.
+    // Overlapping frames should have been removed by remove_coded_frames_and_dependants_in_range()
+    // before this call.
+    auto timestamp = frame.timestamp();
+    size_t insert_index = 0;
+    auto* existing_coded_frame = binary_search(m_coded_frames, timestamp, &insert_index, [](AK::Duration needle, Media::CodedFrame const& frame) {
+        return needle <=> frame.timestamp();
+    });
+    VERIFY(!existing_coded_frame);
```
tcl3, line 49: Is a VERIFY the appropriate here? It seems like it would be possible for malformed output to trigger this.

Libraries/LibWeb/MediaSourceExtensions/TrackBufferDemuxer.cpp
```diff
@@ -0,0 +1,263 @@
...
+        m_track_buffer_ranges.remove_range(removed_start, removed_end);
+    }
+
+    m_coded_frames.remove(remove_start, remove_end - remove_start);
+
+    // Adjust read position if it was in or past the removed range.
+    if (m_read_position >= remove_end) {
+        m_read_position -= (remove_end - remove_start);
+    } else if (m_read_position > remove_start) {
+        m_read_position = remove_start;
```
tcl3: `m_read_position` can index one past the end of `m_coded_frames` here when the removed range is at the tail of the vector.

Libraries/LibWeb/MediaSourceExtensions/MediaSource.cpp
Zaggy1024: This is essentially a hack to make it work for now, we'll need to implement a whole lot of parsing to be spec-compliant here. Technically, we shouldn't be accepting `vp09` as a codec string at all, not without at least 4 fields (i.e. `vp09.00.41.08`, see https://www.webmproject.org/vp9/mp4/). We'll want to parse and validate those fields to determine codec support across HTMLMediaElement, MediaSource and MediaCapabilities.

I'm just putting off that work until I think a bit about how this all fits into our existing mime sniffing etc. 😅

Libraries/LibWeb/MediaSourceExtensions/MediaSource.cpp
Zaggy1024: This actually varies between browsers, and it seems to essentially be implementation-defined, but Chromium prefers certainty and returns false for `video/webm`. Firefox returns true. I'd prefer to be pickier with this, since we'll have no idea whether we support the codecs if they don't specify them.

This is something I'll have to revisit though. I had set up a test page set up for `canPlayType()`, `isTypeSupported()` and `decodingInfo()` when I was working on AV1 codec parameter string parsing, and Firefox and Chromium were not at all consistent between each other regarding the same mime types, nor were they consistent with themselves across different formats. I'd like to at least try to be consistent, but maybe we'll be forced not to be.

Libraries/LibWeb/MediaSourceExtensions/TrackBufferDemuxer.cpp
Zaggy1024: Any frame on the exact timestamp being added should have been removed by a call to `remove_coded_frames_and_dependants_in_range()` immediately prior to this, so that is not supposed to be possible regardless of input.

R-Goc:
I checked out the PR on windows and I'm not seeing the higher qualities available on youtube, even when spoofing the UA to linux chrome.

Issue #8656, opened 2026-03-28
Crash loading epicompany.eu


wing-tung-leung:
### Summary

When opening homepage from epicompany.eu, crash happens.

### Operating system

macOS

### Steps to reproduce

1. Open https://epicompany.eu

### Expected behavior

Load home page.

### Actual behavior

Crash with error message:

Ladybird flew off-course!
The web page https://epicompany.eu/ has crashed.
You can reload the page to try again.

### URL for a reduced test case

n/a

### HTML/SVG/etc. source for a reduced test case

```html
n/a
```

### Log output and (if possible) backtrace

```shell
1205595.962 WebContent(21144): HTMLScriptElement: Refusing to run script because the element's result is null.
1205596.051 WebContent(21144): FIXME: Unimplemented Navigator.maxTouchPoints
1205596.133 WebContent(21144): Unhandled JavaScript exception (in promise): NetworkError: Failed to load font

CRASH: Received signal SIGSEGV (11)
Stack trace (most recent call first):
#0  0x0000000191cfd6a3 in __sigtramp + 55 at /usr/lib/system/libsystem_platform.dylib
#1  0x0000000105caa13f in JS::create_table_row(JS::Realm&, JS::Value, JS::Value, GC::RootVector<JS::Value, 0ul>&, AK::HashMap<JS::PropertyKey, bool, AK::Traits<JS::PropertyKey>, AK::Traits<bool>, false>&, AK::HashMap<JS::PropertyKey, bool, AK::Traits<JS::PropertyKey>, AK::Traits<bool>, false>&) + 311 at /Users/tung/project/ladybird/Build/release/lib/liblagom-js.0.0.0.dylib
#2  0x0000000105caa13f in JS::create_table_row(JS::Realm&, JS::Value, JS::Value, GC::RootVector<JS::Value, 0ul>&, AK::HashMap<JS::PropertyKey, bool, AK::Traits<JS::PropertyKey>, AK::Traits<bool>, false>&, AK::HashMap<JS::PropertyKey, bool, AK::Traits<JS::PropertyKey>, AK::Traits<bool>, false>&) + 311 at /Users/tung/project/ladybird/Build/release/lib/liblagom-js.0.0.0.dylib
[...]

PR #8658, opened 2026-03-28
LibJS: Handle proxy-wrapped arrays in `console.table()`

tcl3:
Previously, we would crash if `console.table()` was called with a proxy-wrapped array. This is because `Value::is_array()` returns true for proxy-wrapped arrays but `as_array()` simply casts to an Array object. We now treat these objects as array-like and access elements generically through the Object interface, which correctly dispatches through Proxy traps.

Fixes: #8656


AtkinsSJ:
`is_array()` and `as_array()` having different ideas of what an array is seems pretty nasty to me. Could we rename one or something? `is_array_like()` maybe?

tcl3:
> `is_array()` and `as_array()` having different ideas of what an array is seems pretty nasty to me. Could we rename one or something? `is_array_like()` maybe?

I opted to rename `as_array()` to `as_array_exotic_object()` since [IsArray](https://tc39.es/ecma262/#sec-isarray) is the name of a spec AO.

PR #8663, opened 2026-03-29
Extend partial re-layout to eligible absolutely positioned boxes

Psychpsyo:
Rebased/revived version of #8161

This will help with `cq*` units.


Psychpsyo:
> FYI my PR is drafted, because I found regressions not covered by existing tests

Ah, ok. I'll close this one then. (Didn't notice that yours was a draft since the stale bot closed it)

PR #8667, opened 2026-03-29
LibRegex: Fast paths for some common things

awesomekling:
Teach LibRegex to recognize several common pattern shapes and handle them without entering the VM on every start position. ASCII ignore-case literals now scan for candidate first code units instead of checking every position. Literal alternations under `/i` fold the first byte and verify alternatives in source order. Whole-pattern `\bliteral\b` outside Unicode mode does a literal search with ASCII word-boundary checks. Split prefixes like `(^|;)` jump straight to input start and separator positions instead of trying every byte.

A new prefilter pass extracts the longest required literal substring from parsed patterns and feeds it into the existing hint machinery, so patterns with optional prefixes or common-substring alternations can reject misses early.

Sticky regexes now anchor directly at `lastIndex` through a dedicated VM entry point instead of forward-scanning for matches that would be thrown away.

### DIFF
--- a/Libraries/LibRegex/Rust/src/regex.rs
+++ b/Libraries/LibRegex/Rust/src/regex.rs
@@ -76,6 +76,18 @@ impl Regex {
         start: usize,
         out: &mut [i32],
     ) -> vm::VmResult {
+        if self.flags.sticky {
+            let scratch = &mut *self.scratch.borrow_mut();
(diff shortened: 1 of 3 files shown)


awesomekling:
~10% speedup on Octane/regexp.js and more importantly, it prevents the WebIDL parser used by WPT from timing out :)

PR #8638, opened 2026-03-27
LibWeb: Resolve var() in shorthands before pseudo-element filtering

exec:
## Summary
- Fix crash and incorrect computed values when using `var()` in a shorthand property within a `::selection` rule (e.g. `::selection { background: var(--bg); }`)
- The shorthand `background` was filtered out by the pseudo-element property whitelist before variable resolution could occur, leaving its longhands unresolved
- Allow unresolved shorthands to bypass the pseudo-element filter, then filter the expanded longhands after resolution

The computed `background-color` now correctly resolves to the variable's value (matching Chrome and Firefox behavior), instead of falling back to transparent.

## Test plan
- [x] New test: `selection-with-var-in-shorthand-crash.html` verifies the computed `background-color` resolves to `rgb(173, 216, 230)` (lightblue)
- [x] Manually verified: selecting text shows the correct light blue highlight

Fixes #8625.

> This fix was developed with assistance from Claude Code (claude-opus-4-6).

PR #8671, opened 2026-03-29
LibWeb: Resolve pending-substitution leaf longhands

awesomekling:
When a shorthand contains var(), its expanded longhands carry pending-substitution placeholders through the cascade and receive their final values at computed-value time.

Our declaration-order handling could let a pending leaf longhand enter computed style unchanged when the originating shorthand did not win the cascade for that target directly. This left placeholder values visible past computed-value processing instead of producing the final resolved longhand value.

Resolve only pending leaf longhands from their originating shorthand at the point they would otherwise enter computed style. This preserves the existing nested-shorthand expansion path while ensuring computed style only exposes resolved leaf values.

Add regression coverage for both a highlight shorthand using background: var(...) and a nested shorthand using border: var(...). The new and existing targeted test-web cases pass.

Fixes crash when selecting text on https://ladybird.org/


awesomekling:
Fixed by #8638

PR #8673, opened 2026-03-29
Everywhere: Stop forcing `char` to be signed an align with the architecture ABI

cqundefine:
This change was already discussed in #8650. This is basically a cherry pick from the SerenityOS side.

CC @ADKaster


cqundefine:
Hmm, I actually wonder if a better way to do would be to exclude floating point values from being signed and unsigned instead of introducing {Signed,Unsigned}Integral. Leaving this as draft as I'll take a look at it tomorrow.

spholz:
> Hmm, I actually wonder if a better way to do would be to exclude floating point values from being signed and unsigned instead of introducing {Signed,Unsigned}Integral. Leaving this as draft as I'll take a look at it tomorrow.

I did it that way since I personally feel like `Signed` should include floating-point types, since they are signed :shrug:, and because the C++ STL also has the `std::signed_integral` concept.

IIRC, in SerenityOS, there are no remaining uses of `Signed` and `Unsigned`, so I guess I could have just removed these concepts entirely and therefore effectively renamed them to `{Signed,Unsigned}Integral`. The STL doesn't have `std::signed` either.

You might also be able to convert all of these usages to the new `{Signed,Unsigned}Integral` concepts.

Either way, I think we should at least try to keep both AKs somewhat similar, so if you decide to go a different route, I would consider changing Serenity's AK as well to make cherry-picking future AK changes easier.

PR #8674, opened 2026-03-29
Meta: Build system reafactor (Part 2/N)

cqundefine:
A promised follow up to #8650 with more points from #4755 addressed. There's still a bunch more stuff to be done here but for the sake of keeping the PRs small I'll just submit the changes in batches.

Once again CC @ADKaster as you're the most likely one to review build system changes.

### DIFF
--- a/Services/RequestServer/CMakeLists.txt
+++ b/Services/RequestServer/CMakeLists.txt
@@ -1,7 +1,3 @@
-set(CMAKE_AUTOMOC OFF)
-set(CMAKE_AUTORCC OFF)
-set(CMAKE_AUTOUIC OFF)
-
 set(SOURCES
     ConnectionFromClient.cpp
     CURL.cpp


UI/Android/build.gradle.kts
```diff
@@ -38,7 +36,7 @@ android {
             cmake {
                 cppFlags += "-std=c++23"
                 arguments += listOf(
-                    "-DLagomTools_DIR=$buildDir/lagom-tools-install/share/LagomTools",
+                    "-DLagomTools_DIR=$buildDir/host-tools/share/LagomTools",
```
konradekk, line 39: argument name shall be renamed too?

cqundefine: It probably should, but moving away from the Lagom name is a big refactor in and of itself and I'll do it in another PR.

Meta/CMake/presets/CMakeBasePresets.json
```diff
@@ -95,6 +95,22 @@
...
+      "hidden": true,
+      "name": "Host_Tools_base",
+      "displayName": "Host Tools Config",
+      "description": "Host tools build",
+      "binaryDir": "$env{LADYBIRD_SOURCE_DIR}/Build/host-tools-build",
+      "cacheVariables": {
+        "package": "LagomTools",
+        "BUILD_SHARED_LIBS": "OFF",
+        "CMAKE_BUILD_TYPE": "Release",
+        "VCPKG_OVERLAY_TRIPLETS": "$env{LADYBIRD_SOURCE_DIR}/Meta/CMake/vcpkg/release-triplets",
```
ADKaster: this should be distribution. We want a fully static build, and the release triplets don't do that. I have a patch like this in my branch off of Eric's WIP here <https://github.com/ADKaster/ladybird-browser/tree/android-fix-2026>.

Since this host tools preset can be useful for other cross-compile work, I don't feel too conflicted about this commit, but it's unfortunate to make changes to the gradle script when the build doesn't work as-is.

Meta/CMake/check_for_dependencies.cmake
```diff
@@ -1,5 +1,55 @@
...
+    pkg_check_modules(angle REQUIRED IMPORTED_TARGET angle)
+    set(ANGLE_TARGETS PkgConfig::angle)
+endif()
+
+if (WIN32)
+    find_package(pthread REQUIRED)
+    find_package(mman REQUIRED)
+endif()
+
+find_package(Python3 REQUIRED COMPONENTS Interpreter)
```
ADKaster, line 55: We have some lib-specific cmake helper files that could get included here too.

This could also be a stepping stone to a new theoretical feature (I know, I know...) where for the host-tools build, we skip unnecessary third-party deps and make a separate vcpkg.json snippet for just the host tools deps.

ADKaster: I guess this commment is a "no action required" though. So really, the only blocker for this PR is the comment above about using a distribution build.

Meta/CMake/presets/CMakeBasePresets.json
cqundefine: My main motivation for this change is that I plan to make ladybird.py cross compilation aware, if you feel like I should keep the old script for now I can put it back, but I think the change is not that significant. I just changed it to distribution triplets for now.

Meta/CMake/check_for_dependencies.cmake
cqundefine: I added both moving the library specific files there and seperate vcpkg.json to my TODO list :^). There will me more PRs regarding this coming (at least some more tiny cleanups and getting cross compilation in a decent state) so I'd rather not make this one even larger.

ADKaster (approved):

PR #8677, opened 2026-03-30
LibWeb/CSS: Resolve `var()` in keyframe `animation-timing-function`

tcl3:
When a `@keyframes` rule contains `animation-timing-function` with a `var()`, we cannot eagerly resolve it to an `EasingFunction` at rule cache build time because there is no element context available. We now store the unresolved `StyleValue` and defer resolution to `collect_animation_into()`, where the animated element's custom
properties can be used to substitute the variable. 

Previously, an `animation-timing-function` with a `var()` in a `@keyframe` would cause a crash.

This behavior diverges from all major browsers, which don't support `var()` here, but AFAICT there's nothing in the spec that says `var()` shouldn't be resolved in this situation. 

If preferred, I have another fix that resolves the issue in a similar way to other engines, by silently falling back to the animation's default easing.

### DIFF
--- a/Libraries/LibWeb/CSS/Parser/Parser.cpp
+++ b/Libraries/LibWeb/CSS/Parser/Parser.cpp
@@ -1573,7 +1573,7 @@ Vector<Descriptor> Parser::parse_as_descriptor_declaration_block(AtRuleID at_rul
     return parsed_declarations;
 }
 
-bool Parser::is_valid_in_the_current_context(Declaration const&) const
+bool Parser::is_valid_in_the_current_context(Declaration const& declaration) const
 {
     // TODO: Determine if this *particular* declaration is valid here, not just declarations in general.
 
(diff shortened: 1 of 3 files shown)


Libraries/LibWeb/Animations/KeyframeEffect.h
```diff
@@ -70,7 +70,7 @@ class KeyframeEffect final : public AnimationEffect {
             // before they are applied to an element
             HashMap<CSS::PropertyID, Variant<UseInitial, NonnullRefPtr<CSS::StyleValue const>>> properties {};
             Bindings::CompositeOperationOrAuto composite { Bindings::CompositeOperationOrAuto::Auto };
-            Optional<CSS::EasingFunction> easing {};
+            Optional<Variant<CSS::EasingFunction, NonnullRefPtr<CSS::StyleValue const>>> easing {};
```
AtkinsSJ: Might be worth making this a `Variant<Empty, ...>` instead of wrapping in an Optional?

Libraries/LibWeb/CSS/StyleScope.cpp
```diff
@@ -336,10 +341,9 @@ void StyleScope::make_rule_cache_for_cascade_origin(CascadeOrigin cascade_origin
                 if (auto* existing_keyframe = keyframe_set->keyframes_by_key.find(key)) {
                     for (auto& [property_id, value] : resolved_keyframe.properties)
                         existing_keyframe->properties.set(property_id, move(value));
-                    if (resolved_keyframe.easing.has_value())
-                        existing_keyframe->easing = move(resolved_keyframe.easing);
```
AtkinsSJ, line 340: Maybe this wants to be `if (!resolved_keyframe.easing.has<Empty>())`?

tcl3:
> does raise the question of whether we do the right thing if we set animation with var() in it. 

After some spec reading, it looks like all animation properties except `animation-timing-function` are supposed to be ignored within keyframe declarations, so that should cover this case.

I've added a commit that does this.

PR #8678, opened 2026-03-30
Tests: Fix flaky form-disabled-callback WPT test

kalenikaliaksandr:
The test submits a form to /common/blank.html via an iframe. Under file://, this resolves to file:///common/blank.html which doesn't exist. The failed navigation sometimes leaves the iframe cross-origin, causing a SecurityError when accessing iframe.contentWindow.location, which prevents the promise from resolving and causes a timeout.

Serve the test over HTTP so the form submission resolves to the echo server's WPT tree, where /common/blank.html exists.


shannonbooth:
I also have this sitting in https://github.com/LadybirdBrowser/ladybird/pull/8618/commits 😅  I would appreciate if you could give that a glance over at some stage if you see anything wrong with it (sitting in draft, but that's mostly because I am unsure about if it is causing flakes or not)

PR #8680, opened 2026-03-30
LibWeb: Delete whole words with Ctrl+Backspace/Del

ArmindoFlores:
When pressing the Ctrl key (or equivalent PlatformWordJump modifier), backspace and delete now delete a whole word, similarly to what happens when pressing the same modifier with the arrow keys.

### DIFF
--- a/Libraries/LibWeb/HTML/FormAssociatedElement.cpp
+++ b/Libraries/LibWeb/HTML/FormAssociatedElement.cpp
@@ -1078,9 +1078,29 @@ void FormAssociatedTextControlElement::handle_delete(FlyString const& input_type
         if (input_type == UIEvents::InputTypes::deleteContentBackward) {
             if (auto offset = text_node->grapheme_segmenter().previous_boundary(m_selection_end); offset.has_value())
                 selection_start = *offset;
-        } else {
+        } else if (input_type == UIEvents::InputTypes::deleteWordBackward) {
+            while (true) {
+                if (auto offset = text_node->word_segmenter().previous_boundary(selection_start); offset.has_value()) {
+                    auto word = text_node->data().substring_view(*offset, selection_start - *offset);
+                    selection_start = *offset;
+                    if (Unicode::Segmenter::should_continue_beyond_word(word))
+                        continue;
+                }
+                break;
+            }
+        } else if (input_type == UIEvents::InputTypes::deleteContentForward) {
             if (auto offset = text_node->grapheme_segmenter().next_boundary(m_selection_end); offset.has_value())
                 selection_end = *offset;
+        } else if (input_type == UIEvents::InputTypes::deleteWordForward) {
+            while (true) {
+                if (auto offset = text_node->word_segmenter().next_boundary(selection_end); offset.has_value()) {
+                    auto word = text_node->data().substring_view(*offset, selection_end - *offset);
+                    selection_end = *offset;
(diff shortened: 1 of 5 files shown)


AtkinsSJ:
Hi, thanks for adding this! Can you add a test? You should be able to simulate some key presses using the [`Internals` API](https://github.com/LadybirdBrowser/ladybird/blob/master/Libraries/LibWeb/Internals/Internals.idl) - that's exposed to JS as `internals.whatever()`.

ArmindoFlores:
Hello, yes! I probably won't have time today but will do so ASAP. On an additional note, do you think it'd be worth it to factor out the code to get to the end of an entire word? A very similar version of it is also used in `FormAssociatedTextControlElement::increment_cursor_position_to_next_word(...)` and `FormAssociatedTextControlElement::increment_cursor_position_to_previous_word(...)`.

AtkinsSJ:
That sounds good to me, but I'm not familiar with this part of the code. @gmta knows a lot more about this.

Libraries/LibWeb/HTML/FormAssociatedElement.cpp
```diff
@@ -1078,9 +1078,29 @@ void FormAssociatedTextControlElement::handle_delete(FlyString const& input_type
...
+        } else if (input_type == UIEvents::InputTypes::deleteWordBackward) {
+            while (true) {
+                if (auto offset = text_node->word_segmenter().previous_boundary(selection_start); offset.has_value()) {
+                    auto word = text_node->data().substring_view(*offset, selection_start - *offset);
+                    selection_start = *offset;
+                    if (Unicode::Segmenter::should_continue_beyond_word(word))
+                        continue;
+                }
+                break;
+            }
```
gmta, line 1090: In my opinion, this while/if/continue/break logic would be slightly more readable when written like this:

```cpp
            do {
                auto offset = text_node->word_segmenter().previous_boundary(selection_start);
                if (!offset.has_value())
                    break;
                auto word = text_node->data().substring_view(*offset, selection_start - *offset);
                selection_start = *offset;
            } while (Unicode::Segmenter::should_continue_beyond_word(word));
```

Libraries/LibWeb/HTML/FormAssociatedElement.cpp
```diff
@@ -1078,9 +1078,29 @@ void FormAssociatedTextControlElement::handle_delete(FlyString const& input_type
         if (input_type == UIEvents::InputTypes::deleteContentBackward) {
             if (auto offset = text_node->grapheme_segmenter().previous_boundary(m_selection_end); offset.has_value())
                 selection_start = *offset;
-        } else {
+        } else if (input_type == UIEvents::InputTypes::deleteWordBackward) {
```
gmta, line 1081: Let's apply an alphabetical ordering to the `input_type` comparisons here.

Tests/LibWeb/Text/expected/input-delete.txt
```diff
@@ -4,3 +4,9 @@ After: foobar
 --- b ---
 Before: foo👩🏼‍❤️‍👨🏻bar
 After: foobar
+--- c ---
+Before: foo bar
+After: foo 
+--- d ---
+Before: foo bar
+After:  bar
```
gmta, line 12: Please include this change with the previous commit. We won't revert one without the other, if that's ever necessary.

PR #8686, opened 2026-03-30
LibWeb: Basic CSS `position-anchor` support

gmta:
+549 subtest passes in WPT's `css/css-anchor-positioning` category.

Fixes the hand/gun positioning on https://cssdoom.wtf/.

| Before | After |
|--|--|
| Screenshot 2026-03-30 at 17 47 08 | Screenshot 2026-03-30 at 17 46 50 |


gmta:
Making this a draft to fix an issue: maintaining anchor name to tree order element ordering with style updates.

Libraries/LibWeb/Layout/FormattingContext.cpp
```diff
@@ -1367,23 +1373,221 @@ AbsposContainingBlockInfo FormattingContext::resolve_abspos_containing_block_inf
...
+    // block corresponding to the property the function appears in with the specified edge of the target anchor
+    // element's anchor box.
+    auto containing_block_direction = containing_block->computed_values().direction();
+    auto box_direction = box.computed_values().direction();
+    auto resolve_anchor_side = [&](CSS::AnchorStyleValue const& anchor, bool is_from_end, bool is_horizontal_axis)
+        -> Optional<CSSPixels> {
+        if (!anchor_rect_in_containing_block.has_value())
+            return {};
+        auto const& rect = anchor_rect_in_containing_block.value();
+        auto const& side = *anchor.anchor_side();
```
tcl3, line 1456: The case where `anchor()` has an `<anchor-name>` is not handled here. Currently we always select the default anchor element for the given box.

Libraries/LibWeb/Layout/FormattingContext.cpp
```diff
@@ -1367,23 +1373,221 @@ AbsposContainingBlockInfo FormattingContext::resolve_abspos_containing_block_inf
...
+            return existing;
+
+        auto const& anchor = value.as_anchor();
+        auto maybe_side_px = resolve_anchor_side(anchor, is_from_end, is_horizontal_axis);
+
+        // If any of these conditions are false, the anchor() function computes to its specified fallback value. If no
+        // fallback value is specified, it makes the declaration referencing it invalid at computed-value time.
+        if (!maybe_side_px.has_value()) {
+            if (anchor.fallback_value())
+                return CSS::LengthPercentageOrAuto::from_style_value(*anchor.fallback_value());
```
tcl3: I can induce a crash here by using an `anchor()` function as the fallback value.

For example:
```html
<!DOCTYPE html>
<style>
    div {
        position: absolute;
        position-anchor: --nonexistent;
        width: 50px;
        height: 50px;
        left: anchor(center, anchor(--other center));
    }
</style>
<div></div>
```

Currently, `parse_anchor()` handles `anchor()` fallback values itself, but I think we should allow `anchor()` values anywhere a `<length-percentage>` is allowed. 

I'm happy to look at this in a separate PR, since it seems to be an issue with parsing rather than resolution, but I thought it worth mentioning, since the crash doesn't happen before this change.

Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -6147,6 +6148,36 @@ void Document::element_with_name_was_removed(Badge<DOM::Element>, GC::Ref<DOM::E
...
+{
+    // Only remove if this element is still the registered one; a later element in tree order may have replaced it.
+    auto it = m_anchor_name_to_element.find(name);
+    if (it != m_anchor_name_to_element.end() && it->value == element)
+        m_anchor_name_to_element.remove(it);
+}
+
+GC::Ptr<Element> Document::element_by_anchor_name(FlyString const& name) const
+{
+    auto result = m_anchor_name_to_element.get(name);
```
tcl3: According to the spec at https://drafts.csswg.org/css-anchor-position-1/#typedef-anchor-name, anchor names are tree scoped. Currently this registry doesn't respect shadow tree boundaries.

Probably worth a FIXME at this stage.

Libraries/LibWeb/Layout/FormattingContext.cpp
```diff
@@ -1367,23 +1373,221 @@ AbsposContainingBlockInfo FormattingContext::resolve_abspos_containing_block_inf
...
+    auto containing_block_direction = containing_block->computed_values().direction();
+    auto box_direction = box.computed_values().direction();
+    auto resolve_anchor_side = [&](CSS::AnchorStyleValue const& anchor, bool is_from_end, bool is_horizontal_axis)
+        -> Optional<CSSPixels> {
+        if (!anchor_rect_in_containing_block.has_value())
+            return {};
+        auto const& rect = anchor_rect_in_containing_block.value();
+        auto const& side = *anchor.anchor_side();
+        if (side.is_keyword()) {
+            switch (side.to_keyword()) {
```
tcl3, line 1458: We're not validating that top/bottom and left/right are being applied to a valid axis. `is_horizontal_axis()` could be used for validation the same as it is elsewhere in this case statement.

robryanx:
Lol, was looking at the exact same thing yesterday for the same doom related reason: https://github.com/LadybirdBrowser/ladybird/compare/master...robryanx:ladybird:position-anchor

Libraries/LibWeb/DOM/Document.cpp
gmta: Ah missed that entirely. I now store an AnchorNameMap with both Documents and ShadowRoots and traverse those up from the querying node. Using the stylesheet origin is left as a FIXME.

ShaneenHall26:
> Lol, was looking at the exact same thing yesterday for the same doom related reason: [master...robryanx:ladybird:position-anchor](https://github.com/LadybirdBrowser/ladybird/compare/master...robryanx:ladybird:position-anchor)

Libraries/LibWeb/Layout/FormattingContext.cpp
gmta: Added a separate lambda to resolve the anchor rect that _does_ take both the anchor name and default anchor name into account.

Libraries/LibWeb/Layout/FormattingContext.cpp
gmta: Good find - updated those four cases.

Libraries/LibWeb/Layout/FormattingContext.cpp
gmta: > but I think we should allow anchor() values anywhere a <length-percentage> is allowed.

When I originally wrote the anchor parsing code I also considered this, but the fact is that it can only appear in anchor contexts. I can't think of any case that would not be correctly handled by the current parsing code. But I'd love to be wrong :)

I've added a fallback value implementation though, so your crash test case no longer crashes.

Libraries/LibWeb/CSS/Parser/ValueParsing.cpp
```diff
@@ -827,12 +827,10 @@ RefPtr<StyleValue const> Parser::parse_anchor(TokenStream<ComponentValue>& token
         argument_tokens.discard_a_token();
         argument_tokens.discard_whitespace();
         fallback_value = parse_length_percentage_value(argument_tokens);
-        if (!fallback_value) {
+        if (!fallback_value)
             fallback_value = parse_anchor(argument_tokens);
-            if (!fallback_value)
-                return {};
-            argument_tokens.discard_a_token();
```
AtkinsSJ, line 834: For reference, this is why I like always putting a comment after `discard_a_token()` saying what's being discarded, even though it's a bit redundant a lot of the time. Makes these mistakes a bit more obvious.

I think we want to `argument_tokens.discard_whitespace()` here though?

Libraries/LibWeb/Layout/FormattingContext.cpp
```diff
@@ -1367,23 +1373,233 @@ AbsposContainingBlockInfo FormattingContext::resolve_abspos_containing_block_inf
...
+                VERIFY_NOT_REACHED();
+            }
+        }
+        if (side.is_percentage()) {
+            // <percentage>
+            //     Refers to a position a corresponding percentage between the start and end sides, with 0% being
+            //     equivalent to start and 100% being equivalent to end.
+            auto percentage = side.as_percentage().percentage().as_fraction();
+            if (is_horizontal_axis)
+                return rect.left() + CSSPixels::nearest_value_for(rect.width().to_double() * percentage);
```
AtkinsSJ: The start side isn't necessarily the left here. I think you need to work backwards from the right in RTL.

Libraries/LibWeb/CSS/Parser/ValueParsing.cpp
gmta: We do want that yeah, added!

Libraries/LibWeb/Layout/FormattingContext.cpp
gmta: Added RTL support

gmta:
@robryanx I've looked at the changes you made and cherry picked some of it to this PR, thanks! I've mentioned you as the co-author for those commits.

Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -6165,6 +6166,24 @@ void Document::element_with_name_was_removed(Badge<DOM::Element>, GC::Ref<DOM::E
...
 }
 
+GC::Ptr<Element> Document::element_by_anchor_name(FlyString const& name, Node const& querying_node) const
+{
+    // https://drafts.csswg.org/css-shadow-1/#tree-scoped-name
+    // If a tree-scoped name is global (such as @font-face names), then when a tree-scoped reference is dereferenced to
+    // find it, first search only the tree-scoped names associated with the same root as the tree-scoped reference. If
+    // no relevant tree-scoped name is found, and the root is a shadow root, then repeat this search in the root's
+    // host's node tree (recursively).
+    for (auto const* root = &querying_node.root(); root; root = root->parent()) {
```
AtkinsSJ: I think `root->parent()` here returns the root's parent element, not the root's host's root like you want? It'll end up examining more elements unnecessarily unless I'm misunderstanding things.

gmta: Good find. Could not add a regression test for it unfortunately because we don't correctly scope the anchor names to the stylesheet tree yet.

AtkinsSJ (approved):
:+1: Looks good to me, let's go!

PR #8687, opened 2026-03-30
Replace spin_until in apply_the_history_step with state machine

kalenikaliaksandr:
Replace the blocking spin_processing_tasks_with_source_until calls in apply_the_history_step_after_unload_check() with an event-driven ApplyHistoryStepState GC cell that tracks 5 phases, following the same pattern used by CheckUnloadingCanceledState.

Key changes:
- Introduce ApplyHistoryStepState with phases: WaitingForDocumentPopulation, ProcessingContinuations, WaitingForChangeJobCompletion, WaitingForNonChangingJobs and Completed
- Add on_complete callbacks to apply_the_push_or_replace_history_step, finalize_a_same_document_navigation, finalize_a_cross_document_navigation, and update_for_navigable_creation_or_destruction
- Remove spin_until from Document::open()
- Use null-document tasks for non-changing navigable updates and document unload/destroy to avoid stuck tasks when documents become non-fully-active
- Defer completely_finish_loading when document has no navigable yet, and re-trigger post-load steps in activate_history_entry for documents that completed loading before activation

### DIFF
--- a/Libraries/LibWeb/HTML/EventLoop/EventLoop.cpp
+++ b/Libraries/LibWeb/HTML/EventLoop/EventLoop.cpp
@@ -116,42 +116,6 @@ void EventLoop::spin_until(GC::Ref<GC::Function<bool()>> goal_condition)
     // NOTE: This is achieved by returning from the function.
 }
 
-void EventLoop::spin_processing_tasks_with_source_until(Task::Source source, GC::Ref<GC::Function<bool()>> goal_condition)
-{
-    auto& vm = this->vm();
-    vm.save_execution_context_stack();
-    vm.clear_execution_context_stack();
-
-    perform_a_microtask_checkpoint();
-
-    // NOTE: HTML event loop processing steps could run a task with arbitrary source. We could end up re-entering into
-    //       this method; this makes sure we restore the skip value to its original value.
-    TemporaryChange saved_skip(m_skip_event_loop_processing_steps, true);
-
-    Platform::EventLoopPlugin::the().spin_until(GC::create_function(heap(), [this, source, goal_condition] {
-        if (goal_condition->function()())
-            return true;
-        while (auto task = m_task_queue->take_first_runnable_matching([&](auto& candidate_task) {
-            return candidate_task.source() == source;
-        })) {
-            m_currently_running_task = task.ptr();
-            task->execute();
-            m_currently_running_task = nullptr;
-
-            if (goal_condition->function()())
-                break;
-        }
-
-        // FIXME: Remove the platform event loop plugin so that this doesn't look out of place
-        Core::EventLoop::current().wake();
-        return goal_condition->function()();
-    }));
-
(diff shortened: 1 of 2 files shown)


Libraries/LibWeb/HTML/NavigableContainer.cpp
```diff
@@ -115,10 +115,13 @@ WebIDL::ExceptionOr<void> NavigableContainer::create_new_child_navigable(GC::Ptr
...
-    // 12. Append the following session history traversal steps to traversable:
-    traversable->append_session_history_traversal_steps(GC::create_function(heap(), [traversable, navigable, parent_navigable, history_entry, after_session_history_update] {
-        // NB: Use Core::Promise to signal SessionHistoryTraversalQueue that it can continue to execute next entry.
-        auto signal_to_continue_session_history_processing = Core::Promise<Empty>::construct();
+    // AD-HOC: Steps 1-6 below are done synchronously (not inside the traversal step) so that
+    //         the navigable's session history entries are immediately findable via get_session_history_entries().
+    //         Without this, document.open() on a child navigable can fail because the nested history
+    //         hasn't been appended yet. The spec says these should run as session history traversal steps,
+    //         but deferring them causes crashes when document.open()/document.write() runs before
+    //         the traversal step fires. Chromium also does not defer this setup.
```
shannonbooth: Hmm, I had a different understanding here. My interpretation here where this is done by document.open cancelling ongoing navigations, see: https://github.com/LadybirdBrowser/ladybird/pull/8618/changes/1608296576ce0b1436700915e5ae5466cd81245c

Do they run this synchronously? This has a lot of cross over with the changes I am trying to make here: https://github.com/LadybirdBrowser/ladybird/pull/8618

shannonbooth: Does "fail" here mean Push happening instead of a Replace?

shannonbooth: Actually, is the queue here because this should live on the UI process - so it becomes asynchronous because of a message being sent to the UI process to perform the history steps?

kalenikaliaksandr: > Hmm, I had a different understanding here. My interpretation here where this is done by document.open cancelling ongoing navigations, see: https://github.com/LadybirdBrowser/ladybird/commit/1608296576ce0b1436700915e5ae5466cd81245c

I think your interpretation is more correct, so I adopted it here along with some other changes from your branch :)

> Does "fail" here mean Push happening instead of a Replace?

sorry, it was confusing comment. I dropped it after adopting your approach.

> Actually, is the queue here because this should live on the UI process - so it becomes asynchronous because of a message being sent to the UI process to perform the history steps?

not sure what do you mean exactly. I think we want to move SHTQ and SHE entries in UI process and I am trying to make progress towards that.

PR #8691, opened 2026-03-30
UI/Gtk: Add GTK4/libadwaita frontend

jdahlin:
This adds an alternative Linux frontend using GTK4 and libadwaita, modeled after the existing Qt frontend and inspired heavily by GNOME Web (Epiphany). Using adwaita means that we adopt the whole recent [design guidelines](https://developer.gnome.org/hig/) from GNOME, for example no menubar.

https://github.com/user-attachments/assets/43c85af9-e554-4a63-afd6-a80da71711a2
                                                                                                                                                                                                                  
What's implemented:                    

 - Core: GLib event loop bridge, GDK to Web input event translation, bitmap painting via GdkTexture
 - Browsing: Tabbed browsing (AdwTabView), back/forward/reload, find-in-page, zoom controls, fullscreen 
 - URL bar: Autocomplete, domain highlighting, security icons, search engine integration
 - Menus: Hamburger menu, context menus (page/link/image/media) with icons and accelerators
 - Dialogs: Alert, confirm, prompt, color picker, file picker
 - Platform: Clipboard, multi-window, light/dark theme, select dropdowns, DPR scaling, keyboard shortcuts

Implementation notes:
- GtkBuilder xml files are used to reduce boilerplate in cpp files
- Different to [earlier attempt](https://github.com/LadybirdBrowser/ladybird/pull/5918) by @ADKaster this does not use gtkmm (gtk+ cpp binding) , rather just the C API to interact with gtk and adwaita.
- There is some separation between custom GtkWidgets (Widgets/ subdir) and the Ladybird UI API.
- Tried hard not to reimplement anything already present in libwebview, but I may have missed something.
- Does not attempt to integrate vcpkg nor flatpak (future work)
- Rendering is not optimized, tried to copy Qt as much as possible.

I have done some basic dog fooding of all features and they appear to work well in the normal case.

Marking this as a draft since I'm not sure how to proceed with 4.5k new lines. I split up in 5 somewhat smaller commits, but they do not currently compile individually.

### DIFF
--- a/Documentation/BuildInstructionsLadybird.md
+++ b/Documentation/BuildInstructionsLadybird.md
@@ -252,14 +252,35 @@ If you want to run other applications, such as the JS REPL or the WebAssembly RE
 Ladybird will be built with one of the following browser frontends, depending on the platform:
 * [AppKit](https://developer.apple.com/documentation/appkit?language=objc) - The native UI on macOS.
 * [Qt](https://doc.qt.io/qt-6/) - The UI used on all other platforms.
+* [GTK 4](https://docs.gtk.org/gtk4/) - An alternative UI on Linux (experimental).
 * [Android UI](https://developer.android.com/develop/ui) - The native UI on Android.
 
-The Qt UI is available on platforms where it is not the default as well (except on Android).
-You can pick the UI using the `LADYBIRD_GUI_FRAMEWORK` option, for example to enable the Qt UI:
+The Qt and GTK UIs are available on Linux. You can pick the UI using the `LADYBIRD_GUI_FRAMEWORK` option:
 
 ```bash
 # From /path/to/ladybird
 cmake --preset Release -DLADYBIRD_GUI_FRAMEWORK=Qt
+cmake --preset Release -DLADYBIRD_GUI_FRAMEWORK=Gtk
+```
+
+#### Additional prerequisites for the GTK UI
+
+Building with `LADYBIRD_GUI_FRAMEWORK=Gtk` requires additional system packages, as some vcpkg
+dependencies (e.g. gettext) need to be rebuilt from source:
+
+**Debian/Ubuntu:**
+```bash
+sudo apt install bison libxkbcommon-dev
+```
+
+**Arch Linux/Manjaro:**
+```bash
+sudo pacman -S bison
+```
+
+**Fedora:**
+```bash
+sudo dnf install bison
 ```
 
 ### Build error messages you may encounter
--- a/Documentation/GtkFrontend.md
+++ b/Documentation/GtkFrontend.md
@@ -1,0 +1,105 @@
+# GTK Frontend
+
+GTK4/libadwaita frontend for Ladybird.
+
+## Style Guide
+
+Follow the project [Coding Style](CodingStyle.md). Prefer C++ over C idioms wherever possible. Below are GTK-specific rules.
+
+### General
+- Do not duplicate functionality already in LibWebView — the frontend is a thin shell
+- Follow [GNOME HIG](https://developer.gnome.org/hig/) — prefer libadwaita widgets (`AdwTabView`, `AdwHeaderBar`, `AdwAlertDialog`, `AdwToast`, etc.) over custom implementations
+- Prefer C++ types and patterns over GLib/GObject equivalents
+- Use D-Bus via `GApplication` for single-instance and IPC — not custom socket/file mechanisms
+- Prefer file-static functions over anonymous namespaces for internal helpers
+- Namespace: `Ladybird::` for C++ classes, `LadybirdWidgets::` for GObject widget helpers
+
+### GObject
+- Only register GTypes (`G_DEFINE_FINAL_TYPE`) for widgets used in GtkBuilder templates
+- No `G_DECLARE_FINAL_TYPE` — define structs and `GType` functions manually to avoid reserved `_TypeName` names
+- Use `GtkBuilder` and `.ui` resource files for declarative layout where practical
+
+### Signals and Callbacks
+- No `bind_template_callback` — connect signals in C++ with `g_signal_connect_swapped`
+- No `g_signal_new` — use `Function<>` callbacks instead of custom GObject signals
+- Prefer `g_signal_connect_swapped` over `g_signal_connect` to avoid `static_cast<Foo*>(user_data)`
+- For buttons in popover menus, prefer `action-name` properties over click signal handlers
+
+### Memory Management
+- No manual `g_object_unref` — use `GObjectPtr<T>` (see `UI/Gtk/GLibPtr.h`)
+- No manual `g_free` — use `g_autofree`
+
+```cpp
+// Local scope — auto-unrefs when out of scope
+GObjectPtr builder { gtk_builder_new_from_resource("/path/to/file.ui") };
+
+// Class member
+GObjectPtr<GdkTexture> m_texture;
+m_texture = GObjectPtr<GdkTexture> { gdk_memory_texture_builder_build(builder) };
+
+// g_autofree for glib-allocated strings
+g_autofree char* path = g_file_get_path(file);
+```
+
+## Test Checklist
+
+### Navigation
+- [ ] Load URL from location bar
+- [ ] Back/forward buttons
+- [ ] Reload (Ctrl+R, F5)
+- [ ] New tab (Ctrl+T)
+- [ ] Close tab (Ctrl+W)
+- [ ] New window (Ctrl+N)
+- [ ] Tab switching
+- [ ] Child tab creation (target=_blank)
+
+### Input
+- [ ] Keyboard input in web content
+- [ ] Mouse clicks, drag, selection
+- [ ] Touchpad smooth scrolling
+- [ ] Mouse wheel discrete scrolling
+- [ ] Ctrl+scroll zoom
+- [ ] Right-click context menus (page, link, image, media)
+
+### UI
+- [ ] Location bar URL display with domain highlighting
+- [ ] Location bar autocomplete
+- [ ] Security icon (https/insecure)
+- [ ] Find in page (Ctrl+F)
+- [ ] Fullscreen (F11)
+- [ ] Zoom in/out/reset (Ctrl+=/-/0)
+- [ ] Tab loading indicator
+- [ ] Cursor changes (pointer, text, resize, etc.)
+- [ ] Tooltips on hover
+- [ ] Dark/light theme follows system
+
+### Dialogs
+- [ ] JavaScript alert
+- [ ] JavaScript confirm
+- [ ] JavaScript prompt
+- [ ] Color picker
+- [ ] File picker (single and multiple)
+- [ ] Select dropdown
+- [ ] Download save dialog
+- [ ] Download confirmation toast
+- [ ] Error dialog
+
+### Window Management
+- [ ] Minimize/maximize/close
+- [ ] Resize
+- [ ] Fullscreen enter/exit
+- [ ] D-Bus single instance (second launch opens tab in existing window)
+
+### Internal Pages
+- [ ] about:version
+- [ ] about:settings
+- [ ] about:processes (Task Manager)
+
+### Clipboard
+- [ ] Copy (Ctrl+C)
+- [ ] Paste (Ctrl+V)
+- [ ] Select all (Ctrl+A)
+
+### DevTools
+- [ ] Enable/disable via banner
+- [ ] Toggle (Ctrl+Shift+I, F12)


ADKaster:
If you used work from my branch or bugaev/ladybird-gtk4, please add Sergey and myself as co-authors on the relevant commits

jdahlin:
> If you used work from my branch or bugaev/ladybird-gtk4, please add Sergey and myself as co-authors on the relevant commits

No code was used from either of the branches, this is from scratch.

UI/Gtk/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+    Tab.cpp
+    WebContentView.cpp
+    Widgets/LbBrowserWindow.cpp
+    Widgets/LbLocationEntry.cpp
+    Widgets/LbWebView.cpp
+    "${GTK_GRESOURCE_SOURCE}"
+)
+# System GTK4 needs harfbuzz with glib integration (hb_glib_*) which vcpkg's harfbuzz doesn't provide.
+# Link system harfbuzz before vcpkg's to satisfy GTK4's dependency.
+target_link_libraries(ladybird PRIVATE /usr/lib/x86_64-linux-gnu/libharfbuzz.so PkgConfig::GTK4 PkgConfig::LIBADWAITA)
```
jdahlin: This is obviously wrong as it is right now, but it is tricky since gtk requires a harfbuzz compiled with glib support, which is different from what is desirable on other platforms.

cqundefine:
Why not just build all the dependencies with vcpkg, like the old Gtk PR does?

jdahlin:
> Why not just build all the dependencies with vcpkg, like the old Gtk PR does?

Significant extra work! Not quite sure if its necessary nor exactly how to proceed there. Happy to do it in this PR or another if it is a requirement for this to land.

cqundefine:
> Significant extra work! Not quite sure if its necessary nor exactly how to proceed there. Happy to do it in this PR or another if it is a requirement for this to land.

My main concern is the harfbuzz hack, which the cleanest solution does seem to be just building it correctly ourselves, although one thing I remember is that some dependencies on vcpkg are slightly outdated. I'm not sure this harfbuzz but will get merged like this because realistically the exact library path is going to be system dependant.

jdahlin:
@cqundefine vcpkg and flatpak has been updated and should be working. I added a separate flatpak so both qt and gtk can be supported.

@ADKaster I have included parts of your earlier work to build deps using vcpkg, and you should properly credited for that.

sideshowbarker:
+1 to having a GTK-UI build. My experiencing with implementing Orca support in Ladybird for the Qt UI — and comparing that to what I can see from code inspection of the Firefox sources — leads me to believe that with a GTK UI, we could develop a significantly better user experience for Orca users, with less code, and with code that’ll be more maintainable over the long term.

_[Note: It’s not so much that GTK itself makes implementing screen-reader support that much easier — it’s instead that unlike Qt, GTK enables you to directly implement the necessary ATK calls if you want to (rather then using their built-in ATK bridge). With Qt in contrast, you really don’t have the option making the direct ATK calls yourself (I learned that the hard way) — instead with Qt you’re forced into only being able to use their built-in ATK bridge, which really isn’t suited to our needs, and which forces us into writing a bunch of other extra code to work around its limitations.]_

CMakeLists.txt
```diff
@@ -14,9 +14,15 @@ if (APPLE AND NOT CMAKE_OSX_DEPLOYMENT_TARGET)
     set(CMAKE_OSX_DEPLOYMENT_TARGET 14.0)
 endif()
 
-# Pass additional information to vcpkg toolchain files if we are using vcpkg.
+include("Meta/CMake/gui_framework.cmake")
+
+# Pass additional information to vcpkg if we are using it.
 if (CMAKE_TOOLCHAIN_FILE MATCHES "vcpkg.cmake$")
     set(CMAKE_PROJECT_ladybird_INCLUDE_BEFORE "Meta/CMake/vcpkg/generate_vcpkg_toolchain_variables.cmake")
+    if (LADYBIRD_GUI_FRAMEWORK STREQUAL "Qt" OR LADYBIRD_GUI_FRAMEWORK STREQUAL "Gtk")
```
cqundefine, line 22: You're adding the `qt` feature which is what the old PR did, but it also moved the `qtbase` dependencies to that feature.

Meta/CMake/vcpkg/overlay-ports/gtk/portfile.cmake
cqundefine, line 1: Any reason why the `gtk` and `libadwaita` overlay ports are needed, also the vcpkg repository has newer versions available.

Meta/CMake/flatpak/org.ladybird.Ladybird.Gtk.json
```diff
@@ -0,0 +1,733 @@
...
+          "type": "file",
+          "path": "angle/angle-build.sh"
+        },
+        {
+          "type": "file",
+          "path": "angle/angle-install.sh"
+        }
+      ],
+      "build-commands": [
+        "cd angle && gn gen out --args='is_clang=true clang_base_path=\"/usr/lib/sdk/llvm20\" clang_version=\"20\" is_official_build=true is_component_build=true is_debug=false angle_build_tests=false angle_enable_abseil=true angle_enable_renderdoc=false angle_enable_swiftshader=false angle_enable_vulkan=true angle_enable_wgpu=false angle_expose_non_conformant_extensions_and_versions=true angle_use_wayland=true angle_use_x11=false build_angle_deqp_tests=false treat_warnings_as_errors=false use_custom_libcxx=false use_safe_libstdcxx=true use_siso=false use_sysroot=false chrome_pgo_phase=0 is_cfi=false'",
```
cqundefine: Why not just call `angle-configure.sh`, were there any issues with that, I see the `clang_base_path` and `clang_version` parameters were added here but is that really needed as the other Flatpak build works just fine, if so it should probably be another script file.

Meta/CMake/vcpkg/overlay-ports/gtk/portfile.cmake
jdahlin: We need the overlays for vulkan and wayland support.

Meta/CMake/flatpak/org.ladybird.Ladybird.Gtk.json
jdahlin: I speculatively fixed this, hard to check right now as googlesource.com is persistently giving 429.

UI/Gtk/Events.h
```diff
@@ -0,0 +1,23 @@
...
+
+#include <gdk/gdk.h>
+
+namespace Ladybird {
+
+Web::UIEvents::MouseButton gdk_buttons_to_web(GdkModifierType state);
+Web::UIEvents::MouseButton gdk_button_to_web(guint button);
+Web::UIEvents::KeyModifier gdk_modifier_to_web(GdkModifierType state);
+Web::UIEvents::KeyCode gdk_keyval_to_web(guint keyval);
+char const* standard_cursor_to_css_name(Gfx::StandardCursor cursor);
```
ADKaster: why do we want this method to return `char const*` when we could easily return StringView instead? Adding unsafe APIs on purpose seems counter-productive.

UI/Gtk/StringUtils.h
```diff
@@ -0,0 +1,35 @@
...
+#pragma once
+
+#include <AK/ByteString.h>
+#include <AK/String.h>
+#include <AK/StringView.h>
+#include <glib.h>
+
+namespace Ladybird {
+
+inline String ak_string_from_glib_string(char const* str)
```
ADKaster: the only method in this file that kinda makes sense is the one that returns an allocated/owned gchar*.

The one that returns String suggests that there are callers that are going to want to convert back. or worse, that will try to use the underlying bytes of a String as an input to a glib/gtk API expecting char const*. String is not null terminated. That method is extremely suspicious. And if there's ever glib-supplied strings there that are not UTF-9 encoded, it'll crash.

The second one is strictly unnecessary, since `ByteString(str)` has the same behavior as the code in that method.

UI/Gtk/Events.cpp
```diff
@@ -0,0 +1,322 @@
...
+    switch (button) {
+    case GDK_BUTTON_PRIMARY:
+        return Web::UIEvents::MouseButton::Primary;
+    case GDK_BUTTON_MIDDLE:
+        return Web::UIEvents::MouseButton::Middle;
+    case GDK_BUTTON_SECONDARY:
+        return Web::UIEvents::MouseButton::Secondary;
+    case 8:
+        return Web::UIEvents::MouseButton::Backward;
+    case 9:
```
ADKaster: why isn't there an enum val or define for these?

UI/Gtk/EventLoopImplementationGtk.cpp
```diff
@@ -0,0 +1,264 @@
...
+{
+    m_exit_code = code;
+    m_exit_requested = true;
+    if (m_loop && g_main_loop_is_running(m_loop))
+        g_main_loop_quit(m_loop);
+}
+
+void EventLoopImplementationGtk::wake()
+{
+    g_main_context_wakeup(g_main_context_default());
```
ADKaster: is this correct when the EventLoop is on a secondary thread (i.e. not the main UI thread?)

UI/Gtk/Widgets/LbWebView.h
```diff
@@ -0,0 +1,26 @@
...
+}
+
+G_BEGIN_DECLS
+
+#define LB_TYPE_WEB_VIEW (lb_web_view_get_type())
+G_DECLARE_FINAL_TYPE(LbWebView, lb_web_view, LB, WEB_VIEW, GtkWidget)
+
+LbWebView* lb_web_view_new(void);
+Ladybird::WebContentView* lb_web_view_get_impl(LbWebView* self);
+void lb_web_view_set_impl(LbWebView* self, Ladybird::WebContentView* impl);
```
ADKaster: the abbreviated names here do not correspond with the project style. `lb` -> `ladybird`, in all cases.

UI/Gtk/Widgets/LbWebView.cpp
```diff
@@ -0,0 +1,356 @@
...
+    Ladybird::WebContentView* impl;
+};
+
+G_DEFINE_FINAL_TYPE(LbWebView, lb_web_view, GTK_TYPE_WIDGET)
+
+// Forward declarations
+static void lb_web_view_snapshot_func(GtkWidget* widget, GtkSnapshot* snapshot);
+static void lb_web_view_measure_func(GtkWidget* widget, GtkOrientation orientation, int for_size, int* minimum, int* natural, int* minimum_baseline, int* natural_baseline);
+static void lb_web_view_size_allocate_func(GtkWidget* widget, int width, int height, int baseline);
+static void lb_web_view_finalize_func(GObject* object);
```
ADKaster: func -> function, callback, etc.

UI/Gtk/Widgets/LadybirdWebView.cpp
```diff
@@ -0,0 +1,356 @@
...
+
+#include <UI/Gtk/Events.h>
+#include <UI/Gtk/WebContentView.h>
+
+#include <adwaita.h>
+#include <gdk/gdk.h>
+
+struct _LbWebView {
+    GtkWidget parent_instance;
+    Ladybird::WebContentView* impl;
```
ADKaster: why does this C struct 1) have a name starting with `_CapitalLetter`, such names are reserved by the standard. 2) have a raw pointer to an owned instance?

UI/Gtk/Widgets/BuilderUtils.h
```diff
@@ -0,0 +1,22 @@
...
+
+#pragma once
+
+#include <AK/Assertions.h>
+#include <gtk/gtk.h>
+
+namespace Ladybird::Widgets {
+
+template<typename T>
+T* builder_get(GtkBuilder* builder, char const* id)
```
ADKaster: this method is not adequately named, and the filename "FooUtils" is the root of all evil other than globs in build files.

In serenity's LibGUI/LibCore, this method is called "find_child_of_type_named". 

https://github.com/SerenityOS/serenity/blob/14cc6d8e40e034220b84735547f910e08cb72e0d/Userland/Libraries/LibCore/EventReceiver.h#L201-L232

In this case, something like

```
get_builder_object_with_name(Builder, name);
```
?

According to https://docs.gtk.org/gtk4/method.Builder.get_object.html, the returned type is GObject, so perhaps something other than reinterpret cast should be used? Unless GObject inheritance is completely different from C++ object inheritance... (I think I read that somewhere).

UI/Gtk/EventLoopImplementationGtk.cpp
```diff
@@ -0,0 +1,264 @@
+/*
+ * Copyright (c) 2026, Johan Dahlin <jdahlin@gmail.com>
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include <AK/HashMap.h>
```
ADKaster, line 7: none of the first 3 real commits have any CMake accompanying them. Why not? Each commit should build in its entirety, to be truly atomic.

UI/Gtk/Resources/location-entry.ui
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<interface>
```
ADKaster, line 2: why are these files called `.ui`, when they are xml files?

UI/Gtk/Resources/browser-window.ui
```diff
@@ -0,0 +1,257 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ADKaster, line 1: this subdirectory is "ui", which is strange. why not UI? or Resources?

UI/Gtk/Widgets/LbBrowserWindow.cpp
```diff
@@ -0,0 +1,147 @@
...
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include <AK/Assertions.h>
+#include <UI/Gtk/BrowserWindow.h>
+#include <UI/Gtk/Tab.h>
+#include <UI/Gtk/WebContentView.h>
+#include <UI/Gtk/Widgets/LbBrowserWindow.h>
+
+typedef struct _LbBrowserWindow {
```
ADKaster: this file is suffixed ".cpp", so why is there C in it?

```
struct LadybirdBrowserWindow {

};
```
is sufficient.

See https://github.com/LadybirdBrowser/ladybird/blob/master/CONTRIBUTING.md#code-submission-policy

> **Don't:**
>   ...
>   - Write in C. (Instead, take advantage of C++'s amenities, and don't limit yourself to the standard C library.)

UI/Gtk/Widgets/LbBrowserWindow.cpp
```diff
@@ -0,0 +1,147 @@
...
+        tab->view().find_in_page_previous_match();
+}
+
+static void lb_browser_window_on_find_stop_search(LbBrowserWindow* self, GtkSearchEntry*)
+{
+    if (self->impl)
+        self->impl->hide_find_bar();
+}
+
+static void lb_browser_window_class_init(LbBrowserWindowClass* klass)
```
ADKaster: same comment here, we have 120 (or even, 180, the code formatter don't care) columns to work with. don't abbreviate ladybird to lb.

UI/Gtk/Widgets/LbBrowserWindow.cpp
```diff
@@ -0,0 +1,147 @@
...
+{
+    self->impl = nullptr;
+    gtk_widget_init_template(GTK_WIDGET(self));
+}
+
+namespace Ladybird::Widgets {
+
+LbBrowserWindow* create_browser_window_widget(AdwApplication* app, BrowserWindow* impl)
+{
+    auto* window = (LbBrowserWindow*)g_object_new(lb_browser_window_get_type(),
```
ADKaster: don't use C-style casts.

UI/Gtk/Widgets/LbBrowserWindow.cpp
```diff
@@ -0,0 +1,147 @@
...
+        "application", app,
+        nullptr);
+    window->impl = impl;
```
ADKaster: who owns the BrowserWindow? Use a better type than a raw pointer.

UI/Gtk/Widgets/LbBrowserWindow.h
```diff
@@ -0,0 +1,33 @@
...
+ * Copyright (c) 2026, Johan Dahlin <jdahlin@gmail.com>
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#pragma once
+
+#include <adwaita.h>
+
+typedef struct _LbBrowserWindow LbBrowserWindow;
```
ADKaster: a simple forward decl will do. we don't need that weird standards non-complaint _Foo struct name.

UI/Gtk/Widgets/LbLocationEntry.cpp
```diff
@@ -0,0 +1,373 @@
...
+    GtkEntry parent_instance;
+
+    GtkPopover* popover;
+    GtkListBox* list_box;
+    LbLocationEntryState* state;
+};
+
+G_DEFINE_FINAL_TYPE(LbLocationEntry, lb_location_entry, GTK_TYPE_ENTRY)
+
+enum {
```
ADKaster: enum class

UI/Gtk/Widgets/LbLocationEntry.cpp
```diff
@@ -0,0 +1,373 @@
...
+    if (auto const& search_engine = WebView::Application::settings().search_engine(); search_engine.has_value()) {
+        auto placeholder = ByteString::formatted("Search with {} or enter URL", search_engine->name);
+        gtk_entry_set_placeholder_text(GTK_ENTRY(self), placeholder.characters());
+    } else {
+        gtk_entry_set_placeholder_text(GTK_ENTRY(self), "Enter URL or search...");
+    }
+
+    // Load completion popover from resource
+    auto* builder = gtk_builder_new_from_resource("/org/ladybird/Ladybird/gtk/location-entry.ui");
+    self->popover = Ladybird::Widgets::builder_get<GtkPopover>(builder, "completion_popover");
```
ADKaster: why aren't these methods in namespace LadybirdWidgets?

UI/Gtk/Widgets/LbLocationEntry.cpp
```diff
@@ -0,0 +1,373 @@
...
+    signals[SIGNAL_NAVIGATE] = g_signal_new("navigate",
+        LB_TYPE_LOCATION_ENTRY,
+        G_SIGNAL_RUN_LAST,
+        0, nullptr, nullptr, nullptr,
+        G_TYPE_NONE, 1, G_TYPE_STRING);
+}
+
+static void lb_location_entry_init(LbLocationEntry* self)
+{
+    self->state = new LbLocationEntryState { .autocomplete = make<WebView::Autocomplete>(), .suggestions = {}, .user_text = {} };
```
ADKaster: why doesn't the state member own the LocationEntryState? Avoid raw pointers like the plague. (why is that entry state prefixed Lb?)

UI/Gtk/Tab.cpp
```diff
@@ -0,0 +1,342 @@
...
+#include <UI/Gtk/Dialogs.h>
+#include <UI/Gtk/Events.h>
+#include <UI/Gtk/Menu.h>
+#include <UI/Gtk/Tab.h>
+#include <UI/Gtk/WebContentView.h>
+#include <UI/Gtk/Widgets/BuilderUtils.h>
+
+namespace Ladybird {
+
+namespace {
```
ADKaster: the widget methods should be consistent. either anonymous namespace, or all file statics. But why are they in anonymous namespaces anyway? if they're private to the C++ instance, just make them private instance methods. if they're callbacks for some GObject, then that makes more sense.

UI/Gtk/Tab.cpp
```diff
@@ -0,0 +1,342 @@
...
+                auto* sep_row = gtk_list_box_row_new();
+                gtk_list_box_row_set_child(GTK_LIST_BOX_ROW(sep_row), gtk_separator_new(GTK_ORIENTATION_HORIZONTAL));
+                gtk_list_box_row_set_activatable(GTK_LIST_BOX_ROW(sep_row), FALSE);
+                gtk_list_box_row_set_selectable(GTK_LIST_BOX_ROW(sep_row), FALSE);
+                gtk_list_box_append(GTK_LIST_BOX(list_box), sep_row);
+            },
+            [&](auto const&) {});
+    }
+
+    auto* selected = new bool(false);
```
ADKaster: really? we're the ones allocating these manual closures, can't you use a better ownership model than owning raw pointer shared between two callbacks and deleted (maybe) in one of them?

RefPtr seems reasonable here.

UI/Gtk/Tab.cpp
```diff
@@ -0,0 +1,342 @@
...
+        SelectData* select_data;
+    };
+    auto* closed_data = new ClosedData { view, selected, select_data };
+    g_signal_connect(popover, "closed", G_CALLBACK(+[](GtkPopover* popover, gpointer user_data) {
+        auto* d = static_cast<ClosedData*>(user_data);
+        if (!*d->selected)
+            d->view->select_dropdown_closed({});
+        delete d->selected;
+        delete d->select_data;
+        delete d;
```
ADKaster: this is far too many deletes in a row.

UI/Gtk/Tab.cpp
```diff
@@ -0,0 +1,342 @@
...
+
+Tab::Tab(BrowserWindow& window, WebView::WebContentClient& parent_client, u64 page_index)
+    : m_window(window)
+{
+    init_widget(&parent_client, page_index);
+}
+
+Tab::~Tab() = default;
+
+void Tab::init_widget(RefPtr<WebView::WebContentClient> parent_client, size_t page_index)
```
ADKaster: what's the point of this method?

you should reorganize this to use delegating constructors rather than having a separate "init" method.

UI/Gtk/icons/icons.gresource.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<gresources>
+  <gresource prefix="/org/gtk/libgtk/icons">
+    <file alias="16x16/actions/go-down-symbolic.symbolic.png">16x16/actions/go-down-symbolic.symbolic.png</file>
```
ADKaster: how did you create these icons? why are default icons not suitable?

UI/Gtk/Application.cpp
```diff
@@ -0,0 +1,246 @@
...
+        }
+    }
+}
+
+NonnullOwnPtr<Core::EventLoop> Application::create_platform_event_loop()
+{
+    if (!browser_options().headless_mode.has_value()) {
+        Core::EventLoopManager::install(*new EventLoopManagerGtk);
+        adw_init();
+        m_adw_application = ADW_APPLICATION(adw_application_new("org.ladybird.Ladybird", G_APPLICATION_NON_UNIQUE));
```
ADKaster: is non-unique correct? I think we should be telling dbus that there should only be one instance, and any further invocations should defer to our "new window" calls.

UI/Gtk/Application.cpp
```diff
@@ -0,0 +1,246 @@
...
+{
+    if (browser_options().headless_mode.has_value())
+        return WebView::Application::clipboard_text();
+
+    auto* clipboard = gdk_display_get_clipboard(gdk_display_get_default());
+
+    struct ReadState {
+        GMainLoop* loop;
+        char* text;
+    } state = { g_main_loop_new(nullptr, FALSE), nullptr };
```
ADKaster: why are you using a nested g_main_loop rather than a nested Core::EventLoop? What do the other UIs do?

UI/Gtk/Application.cpp
```diff
@@ -0,0 +1,246 @@
...
+        g_main_loop_quit(s->loop); }, &state);
+
+    g_main_loop_run(state.loop);
+    g_main_loop_unref(state.loop);
+
+    Utf16String utf16_text;
+    if (state.text) {
+        utf16_text = Utf16String::from_utf8(StringView { state.text, strlen(state.text) });
+        g_free(state.text);
+    }
```
ADKaster: why not store the UTF16String in the callback state, and handle the lifetime of the gtk-allocated string within it? deferring the g_free to down here, protected by an if statement gives me the heebie jeebies.

UI/Gtk/CMakeLists.txt
```diff
@@ -0,0 +1,63 @@
...
+        --sourcedir "${CMAKE_CURRENT_SOURCE_DIR}/ui"
+        --target "${GTK_GRESOURCE_SOURCE}"
+        --generate-source
+    VERBATIM
+)
+
+# Bundle Adwaita symbolic icons as a GResource so the build is self-contained.
+# Icons are pre-rendered .symbolic.png files under /org/gtk/libgtk/icons,
+# merging with GTK's own built-in icon set.
+file(GLOB GTK_ICON_PNGS "${CMAKE_CURRENT_SOURCE_DIR}/icons/16x16/actions/*.symbolic.png")
```
ADKaster: absolutely not with the file(GLOB). list them out.

UI/Gtk/CMakeLists.txt
```diff
@@ -0,0 +1,63 @@
...
+set(GTK_UI_RESOURCES
+    "${CMAKE_CURRENT_SOURCE_DIR}/ui/browser-window.ui"
+    "${CMAKE_CURRENT_SOURCE_DIR}/ui/list-popover.ui"
+    "${CMAKE_CURRENT_SOURCE_DIR}/ui/location-entry.ui"
+)
+
+set(GTK_GRESOURCE_XML "${CMAKE_CURRENT_SOURCE_DIR}/ui/resources.gresource.xml")
+set(GTK_GRESOURCE_SOURCE "${CMAKE_CURRENT_BINARY_DIR}/gtk-ui-resources.cpp")
+
+add_custom_command(
```
ADKaster, line 18: this custom command needs a custom target added, and the main app needs a dependency on that target. this is a common pattern throughout our cmake.

UI/Gtk/CMakeLists.txt
```diff
@@ -0,0 +1,63 @@
...
+)
+
+# Bundle Adwaita symbolic icons as a GResource so the build is self-contained.
+# Icons are pre-rendered .symbolic.png files under /org/gtk/libgtk/icons,
+# merging with GTK's own built-in icon set.
+file(GLOB GTK_ICON_PNGS "${CMAKE_CURRENT_SOURCE_DIR}/icons/16x16/actions/*.symbolic.png")
+set(GTK_ICONS_GRESOURCE_XML "${CMAKE_CURRENT_SOURCE_DIR}/icons/icons.gresource.xml")
+set(GTK_ICONS_GRESOURCE_SOURCE "${CMAKE_CURRENT_BINARY_DIR}/gtk-icons-resources.cpp")
+
+add_custom_command(
```
ADKaster: same here. needs a custom target.

UI/Gtk/CMakeLists.txt
```diff
@@ -0,0 +1,63 @@
+find_package(PkgConfig REQUIRED)
+pkg_check_modules(GTK4 REQUIRED IMPORTED_TARGET gtk4)
+pkg_check_modules(LIBADWAITA REQUIRED IMPORTED_TARGET libadwaita-1>=1.4)
+
+add_executable(ladybird main.cpp)
```
ADKaster, line 5: this cmake file should be added in the first commit with code, and updated throughout the commit series.

UI/Gtk/Dialogs.cpp
```diff
@@ -0,0 +1,215 @@
...
+        g_list_store_append(filters, filter);
+        g_object_unref(filter);
+    }
+    auto* all_filter = gtk_file_filter_new();
+    gtk_file_filter_set_name(all_filter, "All files");
+    gtk_file_filter_add_pattern(all_filter, "*");
+    g_list_store_append(filters, all_filter);
+    g_object_unref(all_filter);
+    gtk_file_dialog_set_filters(dialog, G_LIST_MODEL(filters));
+    g_object_unref(filters);
```
ADKaster: eugh. can you at least make a GLibPtr or something to wrap all these pointers so you don't have to g_object_unref all over the place? this manual memory management is hard to keep track of.

UI/Gtk/main.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+ErrorOr<int> ladybird_main(Main::Arguments arguments)
+{
+    AK::set_rich_debug_enabled(true);
+
+    auto app = TRY(Ladybird::Application::create(arguments));
+    WebView::BrowserProcess browser_process;
+
+    if (auto const& browser_options = Ladybird::Application::browser_options(); !browser_options.headless_mode.has_value()) {
+        if (browser_options.force_new_process == WebView::ForceNewProcess::No) {
```
ADKaster: in theory this should be handled by dbus, not our manual jank cache files.

UI/Gtk/Menu.cpp
```diff
@@ -0,0 +1,511 @@
...
+
+    void popup(WebContentView& view, WebView::Menu& source, Gfx::IntPoint position)
+    {
+        auto* menu_model = create_application_menu(source, [&](WebView::Action& action) {
+            return ByteString::formatted("context.{}", m_action_names.get(&action).value());
+        });
+        gtk_popover_menu_set_menu_model(GTK_POPOVER_MENU(m_popover), G_MENU_MODEL(menu_model));
+        g_object_unref(menu_model);
+
+        auto dpr = view.device_pixel_ratio();
```
ADKaster: dpr -> device_pixel_ratio

Meta/CMake/vcpkg/overlay-ports/libadwaita/portfile.cmake
```diff
@@ -0,0 +1,35 @@
+string(REGEX MATCH "^([0-9]*[.][0-9]*)" LIBADWAITA_MAJOR_MINOR "${VERSION}")
+vcpkg_download_distfile(LIBADWAITA_ARCHIVE
+    URLS "https://download.gnome.org/sources/libadwaita/${LIBADWAITA_MAJOR_MINOR}/libadwaita-${VERSION}.tar.xz"
+    FILENAME "libadwaita-${VERSION}.tar.xz"
+    SHA512 151810c301929977bc3b64731141cc152e725fa8eee8f9e6ed29d85fa4e0116cece9c8faf94a655d9e523bf2e6a18d94642dda1db7244436cd59992ac2ade45c
+)
+
+vcpkg_extract_source_archive(SOURCE_PATH
+    ARCHIVE "${LIBADWAITA_ARCHIVE}"
```
ADKaster: vcpkg ports should be added as 2 commits: one to add the port exactly as it is upstream, and a second with our changes.

What are the changes for libadwaita? It should be just fine as-is upstream.

vcpkg.json
```diff
@@ -261,9 +291,13 @@
       "name": "fmt",
       "version": "12.1.0#0"
     },
+    {
+      "name": "gtk",
+      "version": "4.18.6"
```
ADKaster: we should be fine with 4.22.x as of https://github.com/microsoft/vcpkg/pull/48982. Why limit to 4.18?

vcpkg.json
```diff
@@ -273,6 +307,10 @@
       "name": "icu",
       "version": "78.2#0"
     },
+    {
+      "name": "libadwaita",
+      "version": "1.7.6"
```
ADKaster: upstream has 1.8.4 https://github.com/microsoft/vcpkg/pull/50395

Meta/CMake/flatpak/org.ladybird.Ladybird.Gtk.json
```diff
@@ -0,0 +1,733 @@
...
+        "-DCMAKE_PREFIX_PATH=/app",
+        "-DCMAKE_INSTALL_LIBDIR=lib",
+        "-DHWY_ENABLE_INSTALL=ON",
+        "-DHWY_ENABLE_EXAMPLES=OFF",
+        "-DHWY_ENABLE_TESTS=OFF",
+        "-DHWY_ENABLE_CONTRIB=OFF"
+      ]
+    },
+    {
+      "name": "libjxl",
```
ADKaster: many of these deps should be shoveable in their own .json that can be consumed by both flatpak jsons to avoid duplication. it's trivial to include a new json at the top of modules

```
modules: [
   "my-shared-deps.json",
   {
      "name": "UI toolkit-specific module A",
      ...
   },
   ..
]
```

Meta/check-flatpak.py
```diff
@@ -192,8 +230,19 @@ def main():
     file_list = sys.argv[1:] if len(sys.argv) > 1 else [VCPKG]
     did_fail = False
 
+    manifests_to_check = []
```
ADKaster: I would consider it okay to add the actual flatpak in a separate PR, after a POC is in tree.

UI/Gtk/EventLoopImplementationGtk.cpp
jdahlin: Good point, I have changed to use g_main_loop_get_context(loop)

UI/Gtk/Widgets/BuilderUtils.h
jdahlin: GObject inheritance is different from C++ inheritance. GObject types are C structs where the first member is always the parent type's struct, forming a chain:
                                                                
```c
struct GtkWidget { GInitiallyUnowned parent; ... };
struct GtkEntry  { GtkWidget parent; ... };                  
```
                                                            
As far as I understand, reinterpret_cast is the correct C++ equivalent here, there's no C++ inheritance hierarchy to use static_cast or dynamic_cast with. The GObject cast macros (GTK_WIDGET(), GTK_ENTRY(), etc.) do the same thing but add a G_TYPE_CHECK_INSTANCE_CAST in debug builds.

UI/Gtk/EventLoopImplementationGtk.cpp
jdahlin: @ADKaster Because I wanted to avoid too big commits and it was extra work to make them compile standalone. If its okay with you I can squash them.

UI/Gtk/Resources/location-entry.ui
jdahlin: That's following the established convention, don't remember why I choose .ui for that back when GtkBuilder was first implemented, but it is what it is.

UI/Gtk/Resources/browser-window.ui
jdahlin: sure

UI/Gtk/Widgets/LbBrowserWindow.cpp
jdahlin: fixed

UI/Gtk/icons/icons.gresource.xml
jdahlin: Kind of a hack to get things working in flatpak, will punt for now.

vcpkg.json
jdahlin: Since we want wayland support and its explicitly disabled upstream. I will bump to 4.22 though.

Meta/CMake/flatpak/org.ladybird.Ladybird.Gtk.json
jdahlin: I will extract shared dependencies into a common JSON module in the flatpak follow-up PR.

UI/Gtk/Widgets/LbBrowserWindow.cpp
jdahlin: Fixed! The impl pointer was removed entirely. After moving all template callbacks  g_signal_connect_swapped in C++ code, no GObject callback references the C++ BrowserWindow through the widget struct anymore. The GObject widget is now a pure UI shell with no back-reference.

jdahlin:
Thanks for the very comprehensive review @ADKaster, I would never have guessed where the line between C++ and C is expected to be drawn, but now I know!

Everything that you commented should be addressed, but I may have misunderstood parts.
- I have removed the flatpak changes, lets focus on getting this built
- Smoother scrolling and incremental update-texture hint for GPU reuse
- A few minor ui features were added: favicons in tabs, audio indicator and download dialogs
- Single instance via dbus
- Added a documented based on this review and what is peculiar to the Gtk frontend
- Many Cism were replaced by C++
- CoreEvent loop instead of g_main_loop.
- Use g_signal_connect_swapped to get the instance to avoid casting.

Future work here is:
- flatpak build
- Bookmarks
- History
- Accessibility
- Font rendering

UI/Gtk/Dialogs.cpp
```diff
@@ -0,0 +1,206 @@
...
+                GError* error = nullptr;
+                auto* file = gtk_file_dialog_open_finish(GTK_FILE_DIALOG(source), result, &error);
+                if (error) {
+                    view->file_picker_closed({});
+                    g_error_free(error);
+                    return;
+                }
+                GObjectPtr owned_file { file };
+                Vector<Web::HTML::SelectedFile> selected;
+                g_autofree char* path = g_file_get_path(file);
```
ADKaster, line 196: I was imagining GObjectPtr would work more like this:

```suggestion
                GObjectPtr file = gtk_file_dialog_open_finish(GTK_FILE_DIALOG(source), result, &error);
                if (error) {
                    view->file_picker_closed({});
                    g_error_free(error);
                    return;
                }
                Vector<Web::HTML::SelectedFile> selected;
                g_autofree char* path = g_file_get_path(file.ptr()); // or .get()
```

vcpkg.json
```diff
@@ -262,9 +292,13 @@
       "name": "fmt",
       "version": "12.1.0#0"
     },
+    {
+      "name": "gtk",
+      "version": "4.18.6"
```
ADKaster: you've imported 4.22, but are still pinning 4.18 here. There's no actual impact b/c the build will use the overlay port regardless (the presence of an overlay port ignores pinned versions for some unknowable reason). But if we upstream a wayland feature, having this version pinned would confuse things when our overlay is removed.

Meta/CMake/vcpkg/overlay-ports/wayland-protocols/portfile.cmake
```diff
@@ -0,0 +1,42 @@
+set(VCPKG_POLICY_EMPTY_INCLUDE_FOLDER enabled)
```
ADKaster, line 1: > wayland 1.24.0 and wayland-protocols 1.44 (bumped from vcpkg
upstream to meet GTK 4.22.0 requirements).

These should be straightforward to bump in microsoft/vcpkg to remove these ports later :tm:

No action requested for this PR though

Meta/CMake/vcpkg/overlay-ports/gtk/portfile.cmake
```diff
@@ -0,0 +1,110 @@
...
+vcpkg_add_to_path("${PKGCONFIG_DIR}") # Post install script runs pkg-config so it needs to be on PATH
+vcpkg_add_to_path("${CURRENT_HOST_INSTALLED_DIR}/tools/glib/")
+
+set(x11 false)
+set(wayland false)
+set(vulkan disabled)
+set(win32 false)
+set(osx false)
+if(VCPKG_TARGET_IS_LINUX)
+    set(x11 true)
```
ADKaster: setting both x11 and wayland requires the host to have x11 devel libs available external to vcpkg right? On a wayland-only system, that's unfortunate.

Meta/CMake/vcpkg/overlay-ports/gtk/vcpkg.json
```diff
@@ -0,0 +1,104 @@
+{
+  "name": "gtk",
```
ADKaster, line 2: for these ports, it really does make things easier to split into a "add port" and "patch port" commit.

so this first commit, it'd be nice to see it broken up into the following or equivalent:

- Meta: Add wayland and wayland-protocols overlay ports (can group these b/c we aren't patching em)
- Meta: Add gtk overlay port
- Meta: Enable wayland in gtk overlay port
- CMake: Add GTK4 as a GUI framework option

Meta/CMake/vcpkg/overlay-ports/gtk/portfile.cmake
jdahlin: I will drop x11 for now, we can add it back later if requested as a separate PR.

jdahlin:
> The current state is a lot better and feels ready for some more UI-minded folks/reviewers to poke at now. A couple more things from me though.

Great, I have now addressed the remaining ones. I had an additional issue with check-flatpack.py which required a small change related to GIT_DIR for it to pass.

Documentation/BuildInstructionsLadybird.md
```diff
@@ -252,14 +252,35 @@ If you want to run other applications, such as the JS REPL or the WebAssembly RE
...
+```
+
+#### Additional prerequisites for the GTK UI
+
+Building with `LADYBIRD_GUI_FRAMEWORK=Gtk` requires additional system packages, as some vcpkg
+dependencies (e.g. gettext) need to be rebuilt from source:
+
+**Debian/Ubuntu:**
+```bash
+sudo apt install bison
```
trflynn89: I also had to install `libxkbcommon-dev`. I'm on Kubuntu so it could due to KDE, but it seemed to be hard dependency of the gtk vcpkg port.

UI/Gtk/Tab.cpp
```diff
@@ -0,0 +1,351 @@
...
+        gtk_widget_grab_focus(root);
+    };
+
+    m_view->on_close = [this]() {
+        m_window.close_tab(*this);
+    };
+
+    m_view->on_web_content_crashed = [this]() {
+        // TODO: Show crash page
+    };
```
trflynn89: This gave me a compile error due to `this` being unused

trflynn89: Actually you shouldn't need to override this at all, displaying a crash page is handled in LibWebView. The callback basically just exists to let test-web handle test crashes.

Documentation/BuildInstructionsLadybird.md
jdahlin: To be able to reduce the scope, this PR has targetted wayland only, X11 will be a follow up commit.

UI/Gtk/Tab.cpp
jdahlin: Sure, I'll remove it!

Documentation/BuildInstructionsLadybird.md
trflynn89: I am actually using wayland - if it were x11, I think I would have needed `libxkbcommon-x11-dev`

UI/Gtk/Application.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+Application::Application() = default;
+Application::~Application()
+{
+    g_clear_object(&m_adw_application);
+}
+
+void Application::create_platform_arguments(Core::ArgsParser&)
+{
+}
```
trflynn89: No need to override this one if there's no implementation (generally UIs shouldn't need to override either this or `create_platform_options` except for very UI-specific cases)

UI/Gtk/Application.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+void Application::create_platform_options(WebView::BrowserOptions&, WebView::RequestServerOptions&, WebView::WebContentOptions&)
+{
+    if (!settings().autocomplete_engine().has_value()) {
+        if (auto const& search = settings().search_engine(); search.has_value()) {
+            if (auto engine = WebView::find_autocomplete_engine_by_name(search->name); engine.has_value())
+                settings().set_autocomplete_engine(engine->name);
+        }
+    }
```
trflynn89: Let's not introduce behavior change like this here, especially not for a single UI. It should be handled more centrally if we want to default the autocomplete engine to the search engine. But that should be a separate PR.

UI/Gtk/EventLoopImplementationGtk.cpp
```diff
@@ -0,0 +1,263 @@
...
+    if (glib_supports_signal(signal_number)) {
+        struct Data {
+            int signal_number;
+            Function<void(int)> handler;
+        };
+        auto* data = new Data { signal_number, move(handler) };
+        return static_cast<int>(g_unix_signal_add_full(G_PRIORITY_DEFAULT, signal_number, [](gpointer user_data) -> gboolean {
+                auto* d = static_cast<Data*>(user_data);
+                d->handler(d->signal_number);
+                return G_SOURCE_CONTINUE; }, data, [](gpointer user_data) { delete static_cast<Data*>(user_data); }));
```
trflynn89: This is formatted quite strangely. Would prefer something like:
```c++
        return static_cast<int>(g_unix_signal_add_full(
            G_PRIORITY_DEFAULT,
            signal_number,
            [](gpointer user_data) -> gboolean {
                auto* d = static_cast<Data*>(user_data);
                d->handler(d->signal_number);
                return G_SOURCE_CONTINUE;
            },
            data,
            [](gpointer user_data) { delete static_cast<Data*>(user_data); }));
```

UI/Gtk/EventLoopImplementationGtk.cpp
```diff
@@ -0,0 +1,263 @@
...
+    VERIFY(signal_number != 0);
+
+    if (glib_supports_signal(signal_number)) {
+        struct Data {
+            int signal_number;
+            Function<void(int)> handler;
+        };
+        auto* data = new Data { signal_number, move(handler) };
+        return static_cast<int>(g_unix_signal_add_full(G_PRIORITY_DEFAULT, signal_number, [](gpointer user_data) -> gboolean {
+                auto* d = static_cast<Data*>(user_data);
```
trflynn89: Let's avoid single-letter variable names. `data` is fine here (we aren't shadowing the outer variable).

UI/Gtk/Widgets/LadybirdWebView.cpp
```diff
@@ -0,0 +1,373 @@
...
+
+#include <adwaita.h>
+#include <gdk/gdk.h>
+
+#define LADYBIRD_WEB_VIEW(obj) (reinterpret_cast<LadybirdWebView*>(obj))
+#define LADYBIRD_TYPE_WEB_VIEW (ladybird_web_view_get_type())
+
+struct LadybirdWebView {
+    GtkWidget parent_instance;
+    Ladybird::WebContentView* impl;
```
trflynn89: Always initialize plain types inline, especially pointers.
```c++
Ladybird::WebContentView* impl { nullptr };
```

(an exception would be when we're restricted by the language, e.g. structs passed over rust FFI).

UI/Gtk/Widgets/LadybirdWebView.cpp
```diff
@@ -0,0 +1,373 @@
...
+};
+
+struct LadybirdWebViewClass {
+    GtkWidgetClass parent_class;
+};
+
+G_DEFINE_FINAL_TYPE(LadybirdWebView, ladybird_web_view, GTK_TYPE_WIDGET)
+
+// Forward declarations
```
trflynn89: First, we don't need this comment. It's immediately obvious that we're looking at forward decls.

That said, we tend to avoid long forward decl lists. At quick glance it looks pretty simple to just define these above their users.

UI/Gtk/Widgets/LadybirdWebView.cpp
```diff
@@ -0,0 +1,373 @@
...
+        .screen_position = position,
+        .button = Web::UIEvents::MouseButton::None,
+        .buttons = Ladybird::gdk_buttons_to_web(state),
+        .modifiers = Ladybird::gdk_modifier_to_web(state),
+        .wheel_delta_x = 0,
+        .wheel_delta_y = 0,
+        .click_count = 0,
+        .browser_data = {},
+    };
+    self->impl->enqueue_input_event(move(event));
```
trflynn89: In other UIs, we defined a `enqueue_native_event` that mouse/keyboard events just pipe into to reduce all this duplication. Maybe that's not possible here due to C, but would be nicer.

UI/Gtk/Widgets/LadybirdWebView.h
```diff
@@ -0,0 +1,22 @@
...
+class WebContentView;
+
+}
+
+struct LadybirdWebView;
+
+GType ladybird_web_view_get_type(void);
+LadybirdWebView* ladybird_web_view_new(void);
+Ladybird::WebContentView* ladybird_web_view_get_impl(LadybirdWebView* self);
+void ladybird_web_view_set_impl(LadybirdWebView* self, Ladybird::WebContentView* impl);
```
trflynn89, line 22: Any chance these could be in the `Ladybird` namespace? No worries if they can't due to Gtk / C restrictions, but if they can be, that would be better.

UI/Gtk/StringUtils.h
```diff
@@ -0,0 +1,19 @@
...
+ */
+
+#pragma once
+
+#include <AK/StringView.h>
+#include <glib.h>
+
+namespace Ladybird {
+
+inline gchar* glib_string_from_ak_string(StringView str)
```
trflynn89: Am I missing something or is this unused?

UI/Gtk/WebContentView.cpp
```diff
@@ -0,0 +1,261 @@
...
+        switch (event.key) {
+        case Web::UIEvents::Key_C:
+            app.copy_selection_action().activate();
+            break;
+        case Web::UIEvents::Key_V:
+            app.paste_action().activate();
+            break;
+        case Web::UIEvents::Key_A:
+            app.select_all_action().activate();
+            break;
```
trflynn89: Is there no way to redispatch the original event? That's what we do in other UIs. Not super excited on having to maintain a list of special key codes here.

UI/Gtk/Widgets/LadybirdBrowserWindow.cpp
```diff
@@ -0,0 +1,88 @@
+/*
+ * Copyright (c) 2026, Johan Dahlin <jdahlin@gmail.com>
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include <AK/Assertions.h>
+#include <UI/Gtk/Widgets/LadybirdBrowserWindow.h>
+
+struct LadybirdBrowserWindow {
```
trflynn89, line 10: Same here re: initializing plain types.

In structs hereafter as well, I won't comment on all of them.

UI/Gtk/Widgets/LadybirdLocationEntry.cpp
```diff
@@ -0,0 +1,369 @@
...
+    OwnPtr<LocationEntryState> state;
+};
+
+struct LadybirdLocationEntryClass {
+    GtkEntryClass parent_class;
+};
+
+G_DEFINE_FINAL_TYPE(LadybirdLocationEntry, ladybird_location_entry, GTK_TYPE_ENTRY)
+
+// Forward declarations
```
trflynn89: Same comment here re: forward decls

UI/Gtk/Application.cpp
```diff
@@ -23,6 +32,7 @@ void Application::create_platform_arguments(Core::ArgsParser&)
 
 void Application::create_platform_options(WebView::BrowserOptions&, WebView::RequestServerOptions&, WebView::WebContentOptions&)
 {
+    // Match autocomplete engine to search engine if not already configured
```
trflynn89: This should still be removed, but just a general comment - this comment would have belonged in the commit that added the code below it, not here.

UI/Gtk/Application.cpp
```diff
@@ -54,17 +69,259 @@ NonnullOwnPtr<Core::EventLoop> Application::create_platform_event_loop()
...
+        GObjectPtr initial_folder { g_file_new_for_path(downloads_dir) };
+        gtk_file_dialog_set_initial_folder(GTK_FILE_DIALOG(dialog.ptr()), G_FILE(initial_folder.ptr()));
+    }
+    gtk_file_dialog_set_initial_name(GTK_FILE_DIALOG(dialog.ptr()), ByteString(file).characters());
+
+    Optional<ByteString> result;
+    Core::EventLoop nested_loop;
+
+    gtk_file_dialog_save(GTK_FILE_DIALOG(dialog.ptr()), m_active_window->gtk_window(), nullptr, +[](GObject* source, GAsyncResult* async_result, gpointer user_data) {
+        auto* s = static_cast<Optional<ByteString>*>(user_data);
```
trflynn89: ❌  single letter variables

UI/Gtk/BrowserWindow.cpp
```diff
@@ -0,0 +1,502 @@
...
+
+    auto& app = WebView::Application::the();
+    add_action_to_map(G_ACTION_MAP(m_window), "reload", app.reload_action());
+    add_action_to_map(G_ACTION_MAP(m_window), "preferences", app.open_settings_page_action());
+    add_action_to_map(G_ACTION_MAP(m_window), "about", app.open_about_page_action());
+}
+
+void BrowserWindow::setup_ui(AdwApplication* app)
+{
+    auto* wt = LadybirdWidgets::create_browser_window_widget(app);
```
trflynn89: Not a single-letter var name, but please use more expressive names than `wt` for widgets

UI/Gtk/BrowserWindow.h
```diff
@@ -0,0 +1,92 @@
...
+    void update_zoom_label();
+    void update_find_in_page_result(size_t current_match_index, Optional<size_t> const& total_match_count);
+
+    // Find-in-page
+    void show_find_bar();
+    void hide_find_bar();
+
+    // DevTools
+    void on_devtools_enabled();
+    void on_devtools_disabled();
```
trflynn89, line 49: General comment about comments - comments that point out immediately obvious things only add noise.

UI/Gtk/Tab.cpp
```diff
@@ -0,0 +1,351 @@
...
+    m_view->on_favicon_change = [this](auto const& bitmap) {
+        if (!m_tab_page)
+            return;
+        g_autoptr(GBytes) bytes = g_bytes_new(bitmap.scanline_u8(0), bitmap.size_in_bytes());
+        GObjectPtr texture { gdk_memory_texture_new(bitmap.width(), bitmap.height(), GDK_MEMORY_B8G8R8A8_PREMULTIPLIED, bytes, bitmap.pitch()) };
+        adw_tab_page_set_icon(m_tab_page, G_ICON(texture.ptr()));
+    };
+
+    m_view->on_audio_play_state_changed = [this](auto play_state) {
+        if (m_tab_page)
```
trflynn89: Braces around multi-line if statements

UI/Gtk/Tab.cpp
```diff
@@ -0,0 +1,351 @@
...
+            adw_tab_page_set_indicator_icon(m_tab_page,
+                play_state == Web::HTML::AudioPlayState::Playing
+                    ? g_themed_icon_new("audio-volume-high-symbolic")
+                    : nullptr);
+    };
+
+    // Not implemented: JS dialogs are modal via AdwAlertDialog, so
+    // on_request_set_prompt_text, on_request_accept_dialog, and
+    // on_request_dismiss_dialog are not needed.
```
trflynn89: "not needed" doesn't sound like the right classification. If the dialogs can't be modal, then "not supported" would be better.

These APIs are used for WebDriver. So if we cannot support non-modal dialogs, that's super unfortunate. If we *can* support them, but chose not to at this time (totally fine), please make this a FIXME.

UI/Gtk/Tab.cpp
```diff
@@ -0,0 +1,351 @@
...
+    if (selected) {
+        auto* check = gtk_image_new_from_icon_name("object-select-symbolic");
+        gtk_box_append(GTK_BOX(box), check);
+    } else {
+        auto* spacer = gtk_image_new_from_icon_name("object-select-symbolic");
+        gtk_widget_set_opacity(spacer, 0);
+        gtk_box_append(GTK_BOX(box), spacer);
+    }
+
+    auto* lbl = gtk_label_new(label);
```
trflynn89: ❌  `lbl`
✔️ `label_widget` (or something)

UI/Gtk/Widgets/LadybirdWebView.h
jdahlin: G_DEFINE_FINAL_TYPE generates functions like ladybird_web_view_get_type() at whatever scope the macro is placed. If we put it inside the Ladybird namespace then the generated function becomes Ladybird::ladybird_web_view_get_type(), but the LADYBIRD_TYPE_WEB_VIEW macro expands to an unqualified ladybird_web_view_get_type() call which then fails outside the namespace. Same constraint applies to LadybirdBrowserWindow and LadybirdLocationEntry.

jdahlin:
@trflynn89 thanks for the review! All but one should be addressed now (C in global namespace). I redid some of the event handling to match what webkit is doing, which should also be more similar to the Qt frontend.

UI/Gtk/Application.cpp
```diff
@@ -4,33 +4,24 @@
...
-void Application::create_platform_options(WebView::BrowserOptions&, WebView::RequestServerOptions&, WebView::WebContentOptions&)
-{
-    if (!settings().autocomplete_engine().has_value()) {
-        if (auto const& search = settings().search_engine(); search.has_value()) {
-            if (auto engine = WebView::find_autocomplete_engine_by_name(search->name); engine.has_value())
-                settings().set_autocomplete_engine(engine->name);
-        }
-    }
-}
-
```
trflynn89: Removing this code should be squashed into the `UI/Gtk: Add GLib event loop integration and application entry point` commit (so that it never appeared to begin with)

UI/Gtk/Widgets/Builder.h
```diff
@@ -0,0 +1,22 @@
+/*
```
trflynn89, line 1: nit: The commit title and message still mention string utils, but that no longer exists

UI/Gtk/Widgets/LadybirdWebView.cpp
```diff
@@ -15,7 +15,7 @@
 
 struct LadybirdWebView {
     GtkWidget parent_instance;
-    Ladybird::WebContentView* impl;
+    Ladybird::WebContentView* impl { nullptr };
```
trflynn89, line 18: A lot of your changes to this file in this commit should have been squashed into the `UI/Gtk: Add input events, string utils, and WebContentView` commit. It's more difficult to review otherwise.

UI/Gtk/Tab.cpp
```diff
@@ -0,0 +1,347 @@
...
+        if (m_tab_page) {
+            adw_tab_page_set_indicator_icon(m_tab_page,
+                play_state == Web::HTML::AudioPlayState::Playing
+                    ? g_themed_icon_new("audio-volume-high-symbolic")
+                    : nullptr);
+        }
+    };
+
+    // FIXME: Support non-modal JS dialogs (on_request_set_prompt_text,
+    // on_request_accept_dialog, on_request_dismiss_dialog) for WebDriver support.
```
trflynn89: Please hang the indentation (and generally wrap comments at ~120 chars as able):
```c++
    // FIXME: Support non-modal JS dialogs (on_request_set_prompt_text, on_request_accept_dialog, 
    //        on_request_dismiss_dialog) for WebDriver support.
```

UI/Gtk/WebContentView.cpp
```diff
@@ -0,0 +1,344 @@
...
+    // This mirrors WebKitGTK's two-pass key event handling:
+    // https://github.com/WebKit/WebKit/blob/e4a5c5e/Source/WebKit/UIProcess/API/gtk/WebKitWebViewBase.cpp
+    // (search for keyEventsToPropagate)
+    // GTK4 has no mechanism to re-dispatch consumed events (gtk_widget_event
+    // was removed, gtk_event_controller_key_forward is synchronous-only).
+    // gdk_display_put_event is deprecated but is the only working approach;
+    // WebKitGTK tried replacing it with manual accelerator activation but
+    // reverted due to broken non-Latin keyboard layouts:
+    // https://bugs.webkit.org/show_bug.cgi?id=261348 (original removal)
+    // https://bugs.webkit.org/show_bug.cgi?id=273780 (revert)
```
trflynn89: Wrap at ~120 chars

UI/Gtk/WebContentView.cpp
```diff
@@ -0,0 +1,344 @@
...
+    m_key_events_to_propagate.append(gdk_event_ref(key_data.event));
+    G_GNUC_BEGIN_IGNORE_DEPRECATIONS
+    gdk_display_put_event(gtk_widget_get_display(GTK_WIDGET(m_widget)), key_data.event);
```
trflynn89: I'm not sure this works? ctrl+c, ctrl+a, etc. are not working for me.

trflynn89:
Something I noticed is that the spin wheel seems to never stop on some sites (ladybird.org in this case):

Screenshot_20260413_115920

(I'm fine with fixing oddities like that in a follow up, just wanted to note it before I forget)

UI/Gtk/WebContentView.cpp
jdahlin: This is for corner cases around non-latin keyboard layouts, its too early for this so I just removed it for now.

jdahlin:
> Something I noticed is that the spin wheel seems to never stop on some sites (ladybird.org in this case):
[snip]
> (I'm fine with fixing oddities like that in a follow up, just wanted to note it before I forget)

I noticed that too, I just assumed it was something related to ladybird.org not being rendered properly, as the loading indicator works fine on other sites.

jdahlin:
@trflynn89 and @ADKaster I believe I have addressed all review comments, is there anything left to do before this can be merged?

Documentation/BuildInstructionsLadybird.md
```diff
@@ -252,14 +252,35 @@ If you want to run other applications, such as the JS REPL or the WebAssembly RE
...
 * [Android UI](https://developer.android.com/develop/ui) - The native UI on Android.
 
-The Qt UI is available on platforms where it is not the default as well (except on Android).
-You can pick the UI using the `LADYBIRD_GUI_FRAMEWORK` option, for example to enable the Qt UI:
+The Qt and GTK UIs are available on Linux. You can pick the UI using the `LADYBIRD_GUI_FRAMEWORK` option:
 
 ```bash
 # From /path/to/ladybird
-cmake --preset Release -DLADYBIRD_GUI_FRAMEWORK=Qt
+cmake --preset default -DLADYBIRD_GUI_FRAMEWORK=Qt
```
christianfrey: This was changed to Release in #8814

UI/Gtk/WebContentView.cpp
trflynn89: Please at least add a FIXME to the current `finish_handling_key_event` implementation to make it work like other UIs.

trflynn89:
Let's fix that cmake documentation issue, add that FIXME comment, then I think this is okay to merge.

ADKaster:
This now also needs an update for <https://github.com/LadybirdBrowser/ladybird/pull/8933> (autocomplete refactor) or autocomplete deferred until later (i.e. remove the widget)

jdahlin:
> This now also needs an update for #8933 (autocomplete refactor) or autocomplete deferred until later (i.e. remove the widget)

Yeah, also need update for bookmarks and a few other things. I guess I'll defer in another PR for simplicity. I'll try to do the remaining parts of this PR after my exam tonight.

trflynn89:
Yeah, let's not add more features here - just the fixme comment and the documentation fix please (so long as this branch builds on top of latest master).

Good luck on your exams!

We should talk about next steps for the gtk port before adding more features as well. IMO we should investigate safe-language gtk bindings. I do not feel very comfortable with the most privileged process being written in C with manual memory management. So before we continue adding on features, I'd like us to at least see what e.g. using the rust gtk bindings would look like. (That is after this PR is merged though - I'm fine bringing this in-tree given that it is clearly marked experimental).

jdahlin:
> Yeah, let's not add more features here - just the fixme comment and the documentation fix please (so long as this branch builds on top of latest master).

Done!

> Good luck on your exams!

Thanks (:

> We should talk about next steps for the gtk port before adding more features as well. IMO we should investigate safe-language gtk bindings. I do not feel very comfortable with the most privileged process being written in C with manual memory management. So before we continue adding on features, I'd like us to at least see what e.g. using the rust gtk bindings would look like. (That is after this PR is merged though - I'm fine bringing this in-tree given that it is clearly marked experimental).

Never thought of this before, sounds like a very interesting technical challenge! There are many things to decide, especially how to interact between the rest of c++ world from rust and how far the rust frontier should be moved.

PR #8694, opened 2026-03-31
AK: Remove null state from StringView & Utf16View

shannonbooth:


Libraries/LibCore/System.cpp
gmta, line 1: Commit messages have a typo: uneeded -> unneeded

shannonbooth: Honestly I just forgot how to spell that!

PR #8698, opened 2026-03-31
LibWeb/WebSockets: Prevent opening WebSockets after we've tried to close them

AtkinsSJ:
I had Claude analyse the [recent flake](https://wpt.fyi/results/websockets/constructor/004.html%3Fwpt_flags%3Dh2?diff&filter=ADC&run_id=6249035316068352&run_id=5149340212658176) on `https://wpt.live/websockets/constructor/004.html?wpt_flags=h2` and it identified this race condition, and the fact we didn't reject empty-string protocol names. So I've implemented fixes for both.

The race condition is that we connect websockets in the callback for when DNS lookup resolves. If that takes a little while, it's possible for JS to already discard the WebSocket, and try to close it. But the internal WebSocket object doesn't exist yet, making that a no-op. Then when the DNS lookup did complete, it was blissfully unaware that the attempted close happened, and would just connect the new websocket. My fix is to separately track which sockets are pending connection, and then we can use that to report the early close, and avoid connecting the websocket at all in that case.

It noted that there are a few FIXMEs related to closing websockets which we could do with solving, but this fix (of not *creating* the unwanted websockets) should be enough to solve the race condition. It also noted we never remove anything from `m_websockets`, but I don't know if that causes issues in practice.

This isn't my area so there could be silly mistakes here but it seems sound to me.

### DIFF
--- a/Services/RequestServer/ConnectionFromClient.cpp
+++ b/Services/RequestServer/ConnectionFromClient.cpp
@@ -355,6 +355,7 @@ void ConnectionFromClient::remove_cache_entries_accessed_since(UnixDateTime sinc
 void ConnectionFromClient::websocket_connect(u64 websocket_id, URL::URL url, ByteString origin, Vector<ByteString> protocols, Vector<ByteString> extensions, Vector<HTTP::Header> additional_request_headers)
 {
     auto host = url.serialized_host().to_byte_string();
+    m_pending_websockets.set(websocket_id);
 
     m_resolver->dns.lookup(host, DNS::Messages::Class::IN, { DNS::Messages::ResourceType::A, DNS::Messages::ResourceType::AAAA })
         ->when_rejected([this, websocket_id](auto const& error) {
@@ -367,6 +368,10 @@ void ConnectionFromClient::websocket_connect(u64 websocket_id, URL::URL url, Byt
                 async_websocket_errored(websocket_id, static_cast<i32>(Requests::WebSocket::Error::CouldNotEstablishConnection));
                 return;
             }
+
+            // Don't connect the websocket if we already requested to close it before the DNS lookup completed.
+            if (!m_pending_websockets.remove(websocket_id))
+                return;
 
             WebSocket::ConnectionInfo connection_info(move(url));
             connection_info.set_origin(move(origin));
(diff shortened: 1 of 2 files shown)


shannonbooth:
The WPT does have a dependency on wpt.live running local tests

Though as  I wonder with the changes in #8540 with wpttools python infra we could probably hook this up to our local test server (i might look into it)

maybe we just dont mind about depending on wpt.live though

AtkinsSJ:
Oh whoops, I missed that. I'm happy to drop the import.

AtkinsSJ:
Changes: Removed the test import, and rebased to take advantage of less flakes as of #8703 (hopefully)

Services/RequestServer/ConnectionFromClient.cpp
```diff
@@ -368,6 +369,11 @@ void ConnectionFromClient::websocket_connect(u64 websocket_id, URL::URL url, Byt
                 return;
             }
 
+            // Don't connect the websocket if we already requested to close it before the DNS lookup completed.
+            if (m_pending_websockets.get(websocket_id) != WebSocket::ReadyState::Connecting)
```
gmta: We don't actually need to track the websocket's ReadyState; we could just have a `HashTable<u64> m_pending_websocket_ids` and `.remove()` it in `::websocket_close()` to have the same effect right?

AtkinsSJ: True! I think I assumed there'd be more to it when I started with that. I'll simplify it.

AtkinsSJ:
Rebased, and replaced the pending-websockets hashmap with a hashtable, because it's a binary "is this websocket pending connection or not?" state.

Services/RequestServer/ConnectionFromClient.cpp
```diff
@@ -368,6 +369,11 @@ void ConnectionFromClient::websocket_connect(u64 websocket_id, URL::URL url, Byt
                 return;
             }
 
+            // Don't connect the websocket if we already requested to close it before the DNS lookup completed.
+            if (!m_pending_websockets.contains(websocket_id))
+                return;
+            m_pending_websockets.remove(websocket_id);
```
gmta: Tiny simplification: `.remove(..)` already returns `false` if the value did not exist.

```cpp
if (!m_pending_websockets.remove(websocket_id))
    return;
```

AtkinsSJ: Yup! That's even better, thanks

PR #8704, opened 2026-03-31
UI/Qt: Add Ctrl+Tab and Ctrl+Shift+Tab for tab navigation

j-stechmann:
In addition to Ctrl+PageDown and Ctrl+PageUp, also support Ctrl+Tab and Ctrl+Shift+Tab to switch between browser tabs. This is the common behavior in other browsers.


j-stechmann:
Thank you :D

Some of the CI did fail, sadly.

avbdr:
@j-stechmann maybe you can add space to work same way as a page down too?

j-stechmann:
@avbdr That sounds like a good idea to me, but the tab switching was just a thing that bothered me so I fixed it myself.
If pressing space to scroll on a page not working is something that bothers you, maybe you can contribute it yourself? :)

PR #8705, opened 2026-03-31
LibWeb: Handle null active document in content_document()

kalenikaliaksandr:
The Crash/HTML/image-load-after-iframe-navigated.html test was crashing on CI (Linux sanitizers) with a null pointer dereference at NavigableContainer.cpp:178. The crash occurs because content_document() dereferences the return value of active_document() without checking for null.

When an iframe is navigated, Document::destroy() sets the old document state's document to null via set_document(nullptr), but the navigable (m_content_navigable) remains non-null since it is reused for the new navigation. During the window between the old document being destroyed and the new document being set, active_document() returns null. If JS code accesses iframe.contentDocument during this window (e.g. via a timer callback), content_document() would dereference the null pointer.

Also fixes the spec step numbering which previously skipped step 3.

### DIFF
--- a/Libraries/LibWeb/HTML/NavigableContainer.cpp
+++ b/Libraries/LibWeb/HTML/NavigableContainer.cpp
@@ -174,11 +174,16 @@ DOM::Document const* NavigableContainer::content_document() const
     // 2. Let document be container's content navigable's active document.
     auto document = m_content_navigable->active_document();
 
-    // 4. If document's origin and container's node document's origin are not same origin-domain, then return null.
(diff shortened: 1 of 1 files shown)


gmta:
Is this a spec issue that we should raise upstream?

kalenikaliaksandr:
> Is this a spec issue that we should raise upstream?

I tried doing that in the past and as I recall it was made clear realistically spec cannot account for this kind of issues.

Psychpsyo:
I was curious about that, and I guess this is the source for that claim: https://github.com/whatwg/html/issues/9690#issuecomment-1712832571

PR #8707, opened 2026-03-31
LibJS: Add auto-accessor keyword for class fields

bennypowers:
## Summary

Implement the `accessor` keyword for class fields, which automatically creates getter/setter pairs backed by a private storage slot. This is part of the TC39 stage 3 [decorators proposal](https://github.com/tc39/proposal-decorators).

- Parse `accessor` as a contextual keyword in class field definitions
- Generate a private backing storage name, getter, and setter per auto-accessor
- Support public/private, static/instance, and computed key auto-accessors
- Public auto-accessor properties are enumerable per spec (unlike class methods)

Auto-accessors are a prerequisite for the decorator proposal's accessor decoration, but are independently useful without decorators.

### PR train

1. **This PR** -- `accessor` keyword
2. #8708 -- decorator syntax parsing
3. #8709 -- decorator codegen + runtime

## Test plan

- 30 new tests in `Tests/LibJS/Runtime/classes/class-auto-accessors.js`
- All 6526 existing tests pass with no regressions

### DIFF
--- a/Libraries/LibJS/Bytecode/ClassBlueprint.h
+++ b/Libraries/LibJS/Bytecode/ClassBlueprint.h
@@ -21,6 +21,7 @@ struct ClassElementDescriptor {
         Setter,
         Field,
         StaticInitializer,
+        AutoAccessor,
     };
 
     Kind kind;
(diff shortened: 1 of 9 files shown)


bennypowers:
Combining into #8709 per maintainer request.

PR #8708, opened 2026-03-31
LibJS: Parse decorator syntax on classes and class elements

bennypowers:
## Summary

Add parsing support for TC39 stage 3 [decorator syntax](https://github.com/tc39/proposal-decorators) on classes and class elements. Decorators are parsed but not yet evaluated at runtime (that comes in the next PR in this train).

Builds on #8707.

- Add `@` token type and lexer support
- Parse `DecoratorList` before class declarations, class expressions, and class elements (methods, fields, auto-accessors)
- Support all three decorator expression forms: `@identifier.chain`, `@identifier.chain(args)`, `@(parenthesized expression)`
- Store decorators as `Vec<Expression>` on `ClassData` and `ClassElement` AST nodes
- Register decorator identifiers in the scope collector for correct local variable resolution

### PR train

1. #8707 -- `accessor` keyword
2. **This PR** -- decorator syntax parsing
3. #8709 -- decorator codegen + runtime

Please review only the top commit; the first commit is from #8707.

## Test plan

- 25 new tests for decorator syntax (valid and invalid forms)
- Updated `eval-basic.js` and `parser-line-terminators.js` for `@` token
- All 6551 existing tests pass with no regressions

### DIFF
--- a/Libraries/LibJS/Rust/src/ast.rs
+++ b/Libraries/LibJS/Rust/src/ast.rs
@@ -398,18 +398,38 @@ impl FunctionTable {
         if let Some(ref constructor) = class_data.constructor {
             self.collect_from_expression(constructor, result);
         }
+        for decorator in &class_data.decorators {
+            self.collect_from_expression(decorator, result);
+        }
         for element in &class_data.elements {
             match &element.inner {
-                ClassElement::Method { key, function, .. } => {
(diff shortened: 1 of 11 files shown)


bennypowers:
Combining into #8709 per maintainer request.

PR #8709, opened 2026-03-31
LibJS: Implement TC39 stage 3 decorators

bennypowers:
## Summary

Implement TC39 stage 3 [decorators](https://github.com/tc39/proposal-decorators) for LibJS, spanning the full pipeline from parsing through runtime application.

### Commit 1: Auto-accessor keyword

Implement the `accessor` contextual keyword from the decorators proposal. `accessor x = 1` creates a private backing store with a public getter/setter pair. Supports public/private, static/instance, and computed key auto-accessors.

### Commit 2: Decorator syntax parsing

Add lexer, parser, and AST support for decorator syntax. Decorators are accepted before class declarations, class expressions, and class elements (methods, fields, auto-accessors, getters, setters). All three expression forms are supported: `@ident.chain`, `@ident.chain(args)`, `@(expr)`.

### Commit 3: Decorator codegen and runtime

Evaluate decorator expressions during class definition, pass values through the NewClass instruction, and apply them at runtime per the spec's Decoration Order algorithm.

- 4-pass decoration order: static methods/accessors, instance methods/accessors, static fields, instance fields
- Decorators applied inner-to-outer (reverse source order)
- Class decorators applied last, may replace the class with any callable
- Implements CreateDecoratorAccessObject, CreateAddInitializerFunction, CreateDecoratorContextObject, ApplyDecoratorsToElementDefinition, ApplyDecoratorsToClassDefinition
- Field decorator initializer chains via ClassFieldDefinition
- Per-field extra initializers via ClassFieldDefinition for `addInitializer`
- Instance method extra initializers via ECMAScriptFunctionObject

## Test plan

- [x] 30 auto-accessor tests
- [x] 25 decorator syntax tests (valid and invalid forms)
- [x] 69 decorator runtime tests: application, context, ordering, `addInitializer`, syntax forms
- [x] Full suite: 6554 pass, 0 regressions

### DIFF
--- a/Libraries/LibJS/Bytecode/Bytecode.def
+++ b/Libraries/LibJS/Bytecode/Bytecode.def
@@ -904,7 +904,9 @@ op NewClass < Instruction
     m_class_blueprint_index: u32
     m_lhs_name: Optional<IdentifierTableIndex>
     m_element_keys_count: u32
-    m_element_keys: Optional<Operand>[]
+    m_decorator_operands_count: u32
+    m_all_operands_count: u32
+    m_all_operands: Optional<Operand>[]
 endop
 
 op NewFunction < Instruction
--- a/Libraries/LibJS/Bytecode/ClassBlueprint.h
+++ b/Libraries/LibJS/Bytecode/ClassBlueprint.h
@@ -22,6 +22,7 @@ struct ClassElementDescriptor {
         Setter,
         Field,
         StaticInitializer,
+        AutoAccessor,
     };
 
     Kind kind;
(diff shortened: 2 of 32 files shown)


justinfagnani:
@bennypowers 

> Field extra initializers from addInitializer run after all fields
rather than after their specific field (spec says per-field)

Some combinations of decorator applications may be extremely sensitive to ordering. I would encourage matching the spec exactly here.

bennypowers:
> @bennypowers
> 
> > Field extra initializers from addInitializer run after all fields
> > rather than after their specific field (spec says per-field)
> 
> Some combinations of decorator applications may be extremely sensitive to ordering. I would encourage matching the spec exactly here.

Thanks for the feedback, addressed in 65b5727

bennypowers:
The exception on non-callable return result was removed in [466d59e](https://github.com/LadybirdBrowser/ladybird/commit/466d59e950641153c9b89352d0fc7c55da6baaad)

gmta:

PR #8710, opened 2026-03-31
AK: Don't convert `encode_base64_impl()` input to `StringView`

tcl3:
Since e47cdc6b, converting `ReadonlyBytes` to a `StringView` will crash if the `ReadonlyBytes` given is empty. This caused a crash when calling `btoa("")`, which called encode_base64() with an empty `ReadonlyBytes`. We now don't convert the input from `ReadonlyBytes` to `StringView,` which avoids this problem.

It may be worth broadening the assertion here to accept ReadonlyBytes with size 0: https://github.com/LadybirdBrowser/ladybird/blob/201e615aad8967f033365ded5d6e333c70eb300c/AK/StringView.h#L45-L50

That wasn't necessary in this particular case though.

Fixes a crash in http://wpt.live/html/webappapis/atob/base64.any.html, which I've imported.

### DIFF
--- a/AK/Base64.cpp
+++ b/AK/Base64.cpp
@@ -80,18 +80,18 @@ static ErrorOr<ByteBuffer> decode_base64_impl(StringView input, LastChunkHandlin
     return output;
 }
 
-static ErrorOr<String> encode_base64_impl(StringView input, simdutf::base64_options options)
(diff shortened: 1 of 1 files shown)


tcl3:
I've removed the test import from this branch, since it's added in #8711 and I think that change is also useful.

PR #8711, opened 2026-03-31
AK: Support creating StringView from a null-data ReadonlyBytes

AtkinsSJ:
Regressed in e47cdc6b63ece6e779fb84cb52ffecc1139f2d5a - empty Spans are default-initialized with a null data pointer, and so hit the VERIFY introduced in that commit. Instead, gracefully handle that case by using `""` if the ReadonlyBytes is empty.

Also import a base64 WPT test which exercises this code path.


tcl3:
#8710 fixes this in a different way. The test import will cause a conflict.

AtkinsSJ:
Yeah, you need the test flagged to run in the HTTP server for it to run properly. Also I think this StringView change is necessary - there's no reason not to allow empty ReadonlyBytes to be converted to StringView imo.

shannonbooth:
>  there's no reason not to allow empty ReadonlyBytes to be converted to StringView imo.

Yeah, my bad, this is such an implicit conversion we need to do this.

PR #8717, opened 2026-03-31
LibDatabase: Switch to new SQLite3 library name

trflynn89:
Started seeing deprecation warnings for `SQLite::SQLite3`. Switch to the more modern `SQLite3::SQLite3`.

### DIFF
--- a/Libraries/LibDatabase/CMakeLists.txt
+++ b/Libraries/LibDatabase/CMakeLists.txt
@@ -5,4 +5,4 @@ set(SOURCES
 find_package(SQLite3 REQUIRED)
 
 ladybird_lib(LibDatabase database EXPLICIT_SYMBOL_EXPORT)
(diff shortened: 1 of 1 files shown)


trflynn89:
Apparently CI does not agree! For posterity, this is the warning I'm getting on my mac:
```
CMake Warning (dev) at Libraries/LibDatabase/CMakeLists.txt:8 (target_link_libraries):
  The library that is being linked to, SQLite::SQLite3, is marked as being
  deprecated by the owner.  The message provided by the developer is:

  The target name SQLite::SQLite3 is deprecated.  Please use SQLite3::SQLite3
  instead.

This warning is for project developers.  Use -Wno-dev to suppress it.
```

Looks like this is tied to the CMake version:
https://github.com/Kitware/CMake/commit/d4754c1a0c7110429db4854c13bc0baf4608a9e7

PR #8721, opened 2026-03-31
LibWeb: Use the fallback base URL for document.cookie

olliehcrook:
Use the document's fallback base URL for
  document.cookie get/set instead of the
  literal document URL.

  This makes initial about:blank documents
  use their inherited HTTP(S) context for
  cookie access, which matches the WHATWG
  HTML Standard and fixes
  cookies/navigated-away.html.

  Spec:
  https://html.spec.whatwg.org/multipage/dom.html#dom-document-cookie
  https://html.spec.whatwg.org/multipage/urls-and-fetching.html#fallback-base-url

### DIFF
--- a/Libraries/LibWeb/DOM/Document.cpp
+++ b/Libraries/LibWeb/DOM/Document.cpp
@@ -3509,7 +3509,7 @@ WebIDL::ExceptionOr<String> Document::cookie()
             return m_cookie;
     }
 
-    auto [cookie_version, cookie] = page().client().page_did_request_cookie(m_url, HTTP::Cookie::Source::NonHttp);
+    auto [cookie_version, cookie] = page().client().page_did_request_cookie(fallback_base_url(), HTTP::Cookie::Source::NonHttp);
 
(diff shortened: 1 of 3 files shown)


trflynn89:
I'm confused as to what this PR accomplishes. 

> which matches the WHATWG HTML Standard

Where does the spec say to use the fallback base URL for cookies? It only says to use the document URL, which we are already doing.

> fixes cookies/navigated-away.html

It is still failing in your patch?

olliehcrook:
> I'm confused as to what this PR accomplishes. 
> 
> > which matches the WHATWG HTML Standard
> 
> Where does the spec say to use the fallback base URL for cookies? It only says to use the document URL, which we are already doing.
> 
> > fixes cookies/navigated-away.html
> 
> It is still failing in your patch?

Apologies, I think I confused myself a bit last night. At one point I thought I had it working, but I’d tried a few things and assumed the wrong change fixed the test.

Also looks like I somehow rebaselined the test as failing before pushing.

My bad, I’ll take another stab at it soon.

PR #8722, opened 2026-03-31
Add Devtools pane in HTML

jdahlin:
This PR introduces an embedded DevTools pane in the AppKit and Qt browsers, replacing the existing remote debugging UI.

What's included:

1. An in-process DevTools transport that allows the DevTools server to run without a TCP listener, using a callback-backed connection
2. A new internal WebUI backend that bridges DevTools protocol traffic, inspected view IDs, and remote debugging settings to the frontend
3. LibWebView changes to track the inspected view ID and persist remote debugging settings in shared application state
4. A bundled HTML/JS/CSS frontend served via resource://ladybird/devtools/, with Elements, Console, and Settings panes sharing common message and theme code
5. The existing remote debugging configuration has been moved into the Settings pane

Known issues:
- Cannot connect to remote port and inbrowser at the same time, requires some refactoring
- Mutating the DOM in the console `document.body.innerHTML += '<b>Hello</b>'` currently crashes

### DIFF
--- a/UI/AppKit/Application/Application.h
+++ b/UI/AppKit/Application/Application.h
@@ -35,9 +35,7 @@ private:
 
     virtual void rebuild_bookmarks_menu() const override;
     virtual void update_bookmarks_bar_display(bool) const override;
-
-    virtual void on_devtools_enabled() const override;
-    virtual void on_devtools_disabled() const override;
+    virtual void on_toggle_devtools_panel() const override;
 
     bool m_file_scheme_urls_have_tuple_origins { false };
 };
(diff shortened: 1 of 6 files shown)


trflynn89:
First, please reach out on discord before starting such efforts.

Personally, I am not in favor of having a built-in devtools at this point. We already had one and it was purposefully removed. Especially when this immediately causes more issues using Firefox's devtools.

jdahlin:
> First, please reach out on discord before starting such efforts.
> 
> Personally, I am not in favor of having a built-in devtools at this point. We already had one and it was purposefully removed. Especially when this immediately causes more issues using Firefox's devtools.

Understand, feel free to close this if this is too early.
What would need to happen before something like this can be considered again?

PR #8723, opened 2026-03-31
LibWeb: Make global prototype chains immutable, fix LegacyUnenumerableNamedProperties enumerability 

shannonbooth:
two small WebIDL related fixes.


Libraries/LibWeb/HTML/WorkerGlobalScope.cpp
gmta, line 1: Commit message typo: themseleves -> themselves

shannonbooth: 🤦 thanks

Libraries/LibWeb/HTML/WorkerGlobalScope.cpp
```diff
@@ -37,6 +37,13 @@ WorkerGlobalScope::WorkerGlobalScope(JS::Realm& realm, GC::Ref<Web::Page> page)
 
 WorkerGlobalScope::~WorkerGlobalScope() = default;
 
+// https://webidl.spec.whatwg.org/#platform-object-setprototypeof
+JS::ThrowCompletionOr<bool> WorkerGlobalScope::internal_set_prototype_of(JS::Object* prototype)
+{
+    // 1. Return ? SetImmutablePrototype(O, V).
+    return set_immutable_prototype(prototype);
```
gmta, line 47: The spec link has different spec steps?

shannonbooth: Ah whoops, that is because I didn't look at the updated spec and copy pasted from another place overriding this - looks like it was changed at a later date those cases were introduced in order to support shadow realms.

PR #8727, opened 2026-04-01
LibWeb: Load unrecognized document types as plain text

theeyeofcthulhu:
Files with extensions that Ladybird does not recognize previously did not get loaded. A markdown file, for example, would be labeled "text/markdown" instead of "text/plain", which does not have handling logic in load_document(), as per spec. While spec step 4. for unrecognized file types remains unimplemented, load files as plain text by default.

### DIFF
--- a/Libraries/LibWeb/DOM/DocumentLoading.cpp
+++ b/Libraries/LibWeb/DOM/DocumentLoading.cpp
@@ -501,6 +501,10 @@ GC::Ptr<DOM::Document> load_document(HTML::NavigationParams const& navigation_pa
     //        response, navigationParams's navigable, navigationParams's final sandboxing flag set,
     //        sourceSnapshotParams's has transient activation, and initiatorOrigin.
 
+    // AD-HOC: In lieu of an implementation for steps 3. and 4., fall back to loading a file of an unrecognized
+    //         type as a plain text document.
+    return load_text_document(navigation_params, type).release_value_but_fixme_should_propagate_errors();
+
     // 5. Return null.
     return nullptr;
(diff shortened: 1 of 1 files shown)


Libraries/LibWeb/DOM/DocumentLoading.cpp
```diff
@@ -501,6 +501,10 @@ GC::Ptr<DOM::Document> load_document(HTML::NavigationParams const& navigation_pa
     //        response, navigationParams's navigable, navigationParams's final sandboxing flag set,
     //        sourceSnapshotParams's has transient activation, and initiatorOrigin.
 
+    // AD-HOC: In lieu of an implementation for steps 3. and 4., fall back to loading a file of an unrecognized
+    //         type as a plain text document.
+    return load_text_document(navigation_params, type).release_value_but_fixme_should_propagate_errors();
```
shannonbooth, line 506: I'm not sure what is best here... doing this unconditionally seems like this would cause other issues. For example, after this change if I try access https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3797.pdf ladybird will crash. We're crashing in a place we _shouldn't_ be crashing (which is a bug), but a text representation of that is also weird.

Maybe this should be a whitelist as part of step 3 where we can do text document for markdown as an example.

shannonbooth: And to clarify what i mean by crashibg where we shouldnt be crashing, i was talking about the fix exception propogation here as opposed to in the tokenizer 😅

PR #8736, opened 2026-04-02
LibWeb: Use full clip geometry to determine SVG clipPath

jonbgamble:
Fix an incorrectly clipped svg sprite at https://wikipedia.org

Before:
image

Now:
image


Libraries/LibWeb/Painting/SVGPaintable.cpp
```diff
@@ -22,6 +23,42 @@ Layout::SVGBox const& SVGPaintable::layout_box() const
     return static_cast<Layout::SVGBox const&>(layout_node());
 }
 
+// https://drafts.fxtf.org/css-masking-1/#ClipPathElement
```
AtkinsSJ: The fxtf.org specs recently got moved to csswg.org, so this should be https://drafts.csswg.org/css-masking-1/#ClipPathElement

Libraries/LibWeb/Painting/SVGPaintable.cpp
```diff
@@ -22,6 +23,42 @@ Layout::SVGBox const& SVGPaintable::layout_box() const
...
+bool SVGPaintable::contributes_to_clip_path() const
+{
+    // If a child element is made invisible by display or visibility it does not contribute to the clipping path.
+    return computed_values().visibility() == CSS::Visibility::Visible && !display().is_none();
+}
+
+Optional<CSSPixelRect> SVGPaintable::clip_path_geometry_bounds(Gfx::AffineTransform const& additional_transform) const
```
AtkinsSJ, line 34: This could do with the same https://drafts.csswg.org/css-masking-1/#ClipPathElement link, to make it clear that the text below is a quote from the spec. (It's annoying that there's no direct link for that paragraph.)

Libraries/LibWeb/Painting/SVGPaintable.cpp
```diff
@@ -22,6 +23,42 @@ Layout::SVGBox const& SVGPaintable::layout_box() const
...
+        auto const* svg_paintable = as_if<SVGPaintable>(child);
+        if (!svg_paintable)
+            return IterationDecision::Continue;
+
+        auto child_bounds = svg_paintable->clip_path_geometry_bounds(additional_transform);
+        if (!child_bounds.has_value())
+            return IterationDecision::Continue;
+
+        bounding_box.add_point(child_bounds->location());
+        bounding_box.add_point(child_bounds->location().translated(child_bounds->width(), child_bounds->height()));
```
AtkinsSJ: CSSPixelRect (and other Rects) has `left()`, `right()`, `top()` and `bottom()` methods. Using `right()` and `bottom()` here seems like it'd be simpler than translating the location.

Libraries/LibWeb/Painting/SVGPathPaintable.cpp
```diff
@@ -30,12 +30,40 @@ void SVGPathPaintable::reset_for_relayout()
     m_computed_path.clear();
 }
 
+static Optional<Gfx::Path> transformed_computed_path(SVGPathPaintable const& paintable, Gfx::AffineTransform const& transform, Optional<Gfx::FloatPoint> offset = {})
```
AtkinsSJ: I think it might be better to pass the Gfx::Path in here instead of the SVGPathPaintable. That way, you'd be able to always return a Gfx::Path. Of the 3 users, 1 already early-outs if there's no computed path, and the other too could as well.

AtkinsSJ (changes requested):
Nice! Could you add a test for this, based on the bug it fixes?

Libraries/LibWeb/Painting/SVGPathPaintable.cpp
jonbgamble: good call

jonbgamble: That helper no longer does much, so I'll just remove it.

AtkinsSJ:
Oh and as a side-note, feel free to click those `[Resolve conversation]` buttons once you've pushed changes that apply the feedback (or otherwise resolved it). It makes it clear that there's nothing waiting to be done.

PR #8737, opened 2026-04-02
LibWeb: Store more `StyleValue` sub-values as `StyleValue`

Calme1709:
This PR stores the sub-values of `FitContentStyleValue` and `GridTrackSize` as `StyleValue`s themsevles. This brings us in line with other `StyleValue`s, simplifies handling for absolutization and interpolation, and allows us to support calculated `<flex>` values in the case of `GridTrackSize`.


Libraries/LibWeb/CSS/StyleValues/FitContentStyleValue.cpp
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright (c) 2025, Andreas Kling <andreas@ladybird.org>
+ * Copyright (c) 2026, Callum LaW <callumlaw1709@outlook.com>
```
AtkinsSJ: Typo: LaW

I have no idea why I even noticed this. 😆

Calme1709: Impressive catch - fixed

PR #8742, opened 2026-04-02
Meta: Remove ENABLE_RUST build configuration option

shannonbooth:
This now is required to be set for the browser to function.

### DIFF
--- a/Libraries/LibRegex/RustRegex.cpp
+++ b/Libraries/LibRegex/RustRegex.cpp
@@ -5,10 +5,7 @@
  */
 
 #include <LibRegex/RustRegex.h>
-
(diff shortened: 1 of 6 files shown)


Libraries/LibRegex/RustRegex.cpp
```diff
@@ -6,9 +6,7 @@
 
 #include <LibRegex/RustRegex.h>
 
-#ifdef ENABLE_RUST
-
-#    include <LibUnicode/CharacterTypes.h>
+#include <LibUnicode/CharacterTypes.h>
```
trflynn89, line 8: nit: there's extra newline above here now

PR #8754, opened 2026-04-03
LibJS: Add sampling profiler

jdahlin:
This adds a lightweight JS sampling profiler to LibJS, exports captured samples in Gecko Profile format, and wires js --profile <output> to write profiles that open directly in [Firefox Profiler](https://profiler.firefox.com/). It also improves symbolication by inferring display names for anonymous functions so the resulting call stacks are more readable.

[Before](https://profiler.firefox.com/public/4es8q70p69dtv410asmkf94g2vx9cdm0cy6reqr/calltree/?globalTrackOrder=0&thread=0&timelineType=category&v=15
): 
image

[After](https://profiler.firefox.com/public/7v9zqnt9560w22s0w2471095e819tk9rw5j46j8/calltree/?globalTrackOrder=0&thread=0&timelineType=category&v=15):
image

### DIFF
--- a/Libraries/LibJS/Bytecode/Interpreter.cpp
+++ b/Libraries/LibJS/Bytecode/Interpreter.cpp
@@ -18,6 +18,7 @@
 #include <LibJS/Bytecode/Op.h>
 #include <LibJS/Bytecode/PropertyAccess.h>
 #include <LibJS/Export.h>
+#include <LibJS/Profiler.h>
 #include <LibJS/Runtime/AbstractOperations.h>
 #include <LibJS/Runtime/Accessor.h>
 #include <LibJS/Runtime/Array.h>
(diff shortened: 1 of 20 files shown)


jdahlin:
With a little bit more work hooking this up to the firefox devtools protocol we'll get something like this: https://share.firefox.dev/41L1WSb

awesomekling:
> With a little bit more work hooking this up to the firefox devtools protocol we'll get something like this: [share.firefox.dev/41L1WSb](https://share.firefox.dev/41L1WSb)

Can we get profiling data per source line? If we sample the executable + program counter, we can map the bytecode instruction to an exact source location :)

jdahlin:
> Can we get profiling data per source line? If we sample the executable + program counter, we can map the bytecode instruction to an exact source location :)

@awesomekling This is now implemented, I have tested on macOS aarch64 and linux x86_64. Overhead in non-profiling case is effectively zero and with profiling in the 2-7% range on linux and 0.5-3% on macOS. Done some light testing and the per line sampling appears to work well, both for asmintgen and interpreted execution

jdahlin:
I'm thinking parts of this should be moved out of LibJS into more generic profiling infrastructure.

Firefox's profiler is a separate library from SpiderMonkey. The JS engine only exposes hooks for stack walking, and the profiler calls into it. This means the same sampling infrastructure handles JS, native C++/Rust, and non-JS events like network markers.

In our code, only 4 methods actually touch the JS VM. Everything else (timer thread, signal handling, sample storage, network markers, Gecko format writer) has no LibJS dependency. Splitting them apart would make it easy to add native stack profiling later using the same infrastructure.

PR #8756, opened 2026-04-03
LibWeb: Preserve cross-document nav during same-doc history steps

gmta:
Step 8.3 of "apply the history step" unconditionally sets each changing navigable's ongoing_navigation to "traversal". When a same-document fragment navigation runs while a cross-document navigation is pending (e.g. \<a href="#hash" onclick="location.replace(url)">), this clobbers the pending navigation's ID and causes it to be canceled. Skip setting "traversal" when the navigable already holds a pending navigation ID, matching the behavior all browsers agree on per
https://github.com/whatwg/html/issues/6927.

This fixes the "allow external resources" button in Roundcube Webmail.

| Before | After |
|--|--|
| <video src="https://github.com/user-attachments/assets/7bfd968e-5340-4efc-8971-47a22151d165" /> | <video src="https://github.com/user-attachments/assets/42f136ca-1335-4a63-ab6e-92d93a254598" /> |


Libraries/LibWeb/HTML/TraversableNavigable.cpp
```diff
@@ -606,7 +606,15 @@ void ApplyHistoryStepState::start()
...
-        navigable->set_ongoing_navigation(HTML::Navigable::Traversal::Tag);
+        // AD-HOC: Skip this when the navigable already has a pending cross-document navigation (a String navigation
+        //         ID). Setting "traversal" here would clobber the ID, causing the pending navigation's
+        //         ongoing-navigation check to fail and the navigation to be canceled. The spec's step 3 already gates
+        //         the traversable's ongoing navigation on checkForCancelation (false for push/replace), and we apply
+        //         similar logic here to each navigable. All browsers agree that a same-document navigation should not
+        //         cancel a pending cross-document navigation.
+        //         See: https://github.com/whatwg/html/issues/6927
+        if (!navigable->ongoing_navigation().has<String>())
+            navigable->set_ongoing_navigation(Navigable::Traversal::Tag);
```
kalenikaliaksandr: I probably don't understand what are you solving here to full extent, but we already have m_pending_navigations mechanism that solves similar problem. can we use it here?

gmta: Hm m_pending_navigations is a much better solution, yeah. I've reworked this patch to make use of that.

In essence what this is solving is letting cross-document navigations override traversals if they happen to coincide, which happens with an `<a href="#bla" onclick="location='bla2'">..</a>` for example.

This is what all the big boy browsers seem to do :)

Libraries/LibWeb/HTML/Navigable.cpp
```diff
@@ -2095,10 +2095,26 @@ void Navigable::begin_navigation(NavigateParams params)
...
+                    //         navigation should not cancel a pending cross-document navigation.
+                    //         See: https://github.com/whatwg/html/issues/6927
+                    if (!ongoing_navigation().has<String>()) {
+                        set_delaying_load_events(false);
+                        if (ongoing_navigation().has<Traversal>())
+                            m_pending_navigations.append(params);
+                        else
+                            begin_navigation(params);
+                        return;
+                    }
```
kalenikaliaksandr, line 2117: could you clarify, why among all possible places, this check needs to live in this specific place?

kalenikaliaksandr (changes requested):

Libraries/LibWeb/HTML/Navigable.cpp
kalenikaliaksandr: I mean, would be great to put some explanation in the inline comment above. I could see how this check will spread to other places, so would be cool to have some knowledge why each one was introduced.

gmta:
Closing this since c9fa971905862fcb2340397b4e64e2361a12ab1e fixed the same issue 🥳

PR #8761, opened 2026-04-03
LibWeb: Make PaintableBox fragmentable

### LINKED ISSUE #4148: Outlines rendering incorrectly on broken line boxes
### Summary

When I put an outline on a ::before pseudo in ladybird, its left and right borders are painted for each line of text.
This does not happen in other browsers (Chrome, Firefox) but I am unsure where/if there is a spec-basis for this.
In those browsers, the left/right borders are only drawn once before and after all of the text, not per line.

### Operating system

Linux

### Steps to reproduce

1. Put an outline and a lot of text in a ::before pseudo.
2. Look at the line breaks.
3. Notice that they are different from other browsers.

### Expected behavior

The outlines should be open at line breaks.

### Actual behavior

There is borders being painted at the line breaks.

N/A

```html
<!DOCTYPE html>
<style>
  div::before {
    content: 'This text is very long so that it will break onto multiple lines and showcase the bug.';
    border: 2px solid red;
  }
  div {
    max-width: 100px;
  }
</style>
<div></div>
```

```shell
none
```

Top: Ladybird
Bottom: Firefox
Image

none

### Contribute a patch?

Psychpsyo:
Fixes https://github.com/LadybirdBrowser/ladybird/issues/4148 and prepares us for multicol fragmentation (see #8771)


Psychpsyo:
~~Drafting this as I just realized it's not quite to spec.~~
Putting a FIXME for now, since this'll be easier to sort out once some other fragmentation-related prerequisites are in place.

Libraries/LibWeb/Painting/PaintableBox.h
```diff
@@ -367,6 +378,14 @@ class WEB_API PaintableBox : public Paintable {
 
     BoxModelMetrics m_box_model;
 
+    // FIXME: This is now how this is meant to work in the spec. The box needs to be drawn in full and then sliced
```
AtkinsSJ: Should this be "not" rather than "now"?

Libraries/LibWeb/Painting/PaintableBox.h
```diff
@@ -367,6 +378,14 @@ class WEB_API PaintableBox : public Paintable {
...
+    //        visually, in case something like border-radius is in effect.
+    //        ( see https://drafts.csswg.org/css-break/#valdef-box-decoration-break-slice )
+    bool m_fragment_top_edge_away { false };
+    bool m_fragment_left_edge_away { false };
+    bool m_fragment_right_edge_away { false };
+    bool m_fragment_bottom_edge_away { false };
```
AtkinsSJ, line 387: Nit: I'm not sure "away" is the clearest term here, but I can't think of anything better, and I see this is supposed to be temporary anyway, so 🤷‍♂️.

AtkinsSJ (commented):
All seems correct to me, though I have a couple of tiny nits.

Libraries/LibWeb/Painting/PaintableBox.h
Psychpsyo: Correct, I fixed that now.

Libraries/LibWeb/Painting/PaintableBox.h
Psychpsyo: I agree but I also don't have a better name. (at least not one that keeps the "fragmenting" terminology, which I think is important for these)
And as you said, they're temporary anyways.

AtkinsSJ (approved):

PR #8768, opened 2026-04-04
LibWeb: Invalidate XPath iterator results on DOM mutation

adamjcolvin:
## Summary

`XPathResult` iterator types weren't being invalidated when the DOM tree changed. So `iterateNext()` could keep returning nodes from a stale result instead of throwing `InvalidStateError` as required by the spec. Also, `iterateNext()` wasn't throwing `TypeError` when called on a non-iterator result type.

This tracks iterator-type `XPathResult` objects on `Document` and invalidates them whenever nodes are inserted, removed, or moved. `iterateNext()` now raises `TypeError` or `InvalidStateError` as appropriate.

## Testing

Fixes the following subtests in `Tests/LibWeb/Text/input/wpt-import/domxpath/result-iterateNext.html`:

- `invalidIteratorState should be false for non-iterable results`
- `Calling iterateNext on a non-iterable XPathResult should throw a TypeError`
- `Calling iterateNext on a non-iterable XPathResult after modifying the DOM should throw a TypeError`
- `Calling iterateNext after having modified the DOM should throw an exception`
- `Calling iterateNext after having modified the DOM should throw an exception even if the iterator is exhausted`

All 7 tests in this file now pass.

### DIFF
--- a/Libraries/LibWeb/DOM/Document.cpp
+++ b/Libraries/LibWeb/DOM/Document.cpp
@@ -212,6 +212,7 @@
 #include <LibWeb/WebIDL/ExceptionOr.h>
 #include <LibWeb/WebIDL/ObservableArray.h>
 #include <LibWeb/XPath/XPath.h>
+#include <LibWeb/XPath/XPathResult.h>
 
 namespace Web::DOM {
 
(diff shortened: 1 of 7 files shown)


Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -631,6 +632,7 @@ void Document::visit_edges(Cell::Visitor& visitor)
     visitor.visit(m_scripts_to_execute_in_order_as_soon_as_possible);
     visitor.visit(m_scripts_to_execute_as_soon_as_possible);
     visitor.visit(m_node_iterators);
+    visitor.visit(m_xpath_iterators);
```
awesomekling: If we're relying on `finalize()` to unregister `XPathResult` objects, I think documents need to have weak pointers to the only, or we'll hang on to them forever, no?

adamjcolvin: I see what you mean, yes I think so. And just removing that line will give us weak pointers right? We don't need to change the type of `m_xpath_iterators`?

awesomekling: `m_xpath_iterators` needs to use `GC::RawPtr` (or better yet `GC::RawRef` since we don't want nulls here) instead of `GC::Ptr`, otherwise the GC safety clang plugin will complain about missing visits :)

adamjcolvin: I see, makes sense. Ok I've removed the line and changed the type to `GC::RawRef`. That also meant changing `set` and `remove` to pass references as well. I've amended the commit, let me know if that's ok.

PR #8772, opened 2026-04-04
LibWeb: Use float edges for formatting context avoidance

La5u:
Fixes #8764 

  ## Summary

  Compute float intrusions and float-avoidance placement from root-space
  float margin-box edges instead of reconstructing overlap from
  containing-block margins.

  This fixes nested formatting contexts that are offset by border or
  padding, such as `overflow:hidden` content inside a block with
  `border-left` beside a float.

  ## Details

  The previous float-avoidance logic inferred overlap from accumulated
  containing-block margins. That breaks when a nested containing block is
  shifted by border or padding instead of margin.

  Update the float bookkeeping to store the matched float's root-space
  edges and use those values when:
  - computing float intrusion into nested boxes
  - deciding whether a formatting-context-establishing block must clear
  - placing formatting-context-establishing blocks horizontally
  - computing inline offsets from floats
  - checking opposite-side float fit during float placement

  ## Testing

  - Add a regression test for a floated image followed by a block with
    `border-left`, containing an `overflow:hidden` child that should stay
    beside the float
  - Run the `hidden-overflow` test subset
  - Run the `float-` test sweep

### DIFF
--- a/Libraries/LibWeb/Layout/BlockFormattingContext.cpp
+++ b/Libraries/LibWeb/Layout/BlockFormattingContext.cpp
@@ -1624,12 +1624,12 @@ FormattingContext::SpaceUsedByFloats BlockFormattingContext::intrusion_by_floats
 
     CSSPixels left_intrusion = 0;
     if (space_and_containing_margin.matching_left_float_box) {
-        left_intrusion = space_and_containing_margin.left_used_space;
+        left_intrusion = max(CSSPixels(0), space_and_containing_margin.left_float_margin_edge_right_in_root - box_in_root_rect.x());
     }
 
     CSSPixels right_intrusion = 0;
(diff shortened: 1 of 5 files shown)


AtkinsSJ:
Hi La5u, this is currently failing a few layout tests. It's a good idea to run the whole test suite locally to make sure you don't break one thing when trying to fix another, especially with something as complicated as layout.

I haven't checked the results myself, so it's entirely possible that those failures are actually improvements and the tests need rebaselining. In that case you'll need to do that locally. Either way, take a look at the tests and see if they're behaving correctly or not with your changes.

La5u:
hey, im having trouble running the sanitizer test suite, it keeps crashing at libweb (187/188), might js be my potato laptop not having enough memory. Could someone help debug this?

PR #8774, opened 2026-04-04
AK: Uninitialized struct tm reported by cppcheck

rcorsi:
Cppcheck reports that the struct tm variable should be initialized, at least to properly support Windows.

Checked using cppcheck 2.17 (Ubuntu 25.10) and 2.19 (Ubuntu 26.04 Beta)

```
Checking AK/Time.cpp ...
Checking AK/Time.cpp: AK_OS_WINDOWS...
AK/Time.cpp:488:15: error: Uninitialized variable: &tm [uninitvar]
        (void)localtime_r(&timestamp, &tm);
              ^
32/39 files checked 87% done
```

When using cppcheck 2.17, using the following command line

`cppcheck --check-level=exhaustive AK`

### DIFF
--- a/AK/Time.cpp
+++ b/AK/Time.cpp
@@ -16,11 +16,22 @@
 
 #ifdef AK_OS_WINDOWS
 #    include <AK/Windows.h>
-#    define localtime_r(time, tm) localtime_s(tm, time)
-#    define gmtime_r(time, tm) gmtime_s(tm, time)
 #    define tzname _tzname
 #    define timegm _mkgmtime
 #endif
+
+static bool fill_tm(time_t& timestamp, struct tm& tm, UnixDateTime::LocalTime& local_time)
+{
+#ifdef AK_OS_WINDOWS
+    if (local_time == UnixDateTime::LocalTime::Yes)
+        return localtime_s(&tm, &timestamp) == 0;
+    return gmtime_s(&tm, &timestamp) == 0;
+#else
+    if (local_time == UnixDateTime::LocalTime::Yes)
+        return localtime_r(&timestamp, &tm) != nullptr;
+    return gmtime_r(&timestamp, &tm) != nullptr;
+#endif
+}
 
 namespace AK {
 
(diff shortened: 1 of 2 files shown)


AK/Time.cpp
```diff
@@ -481,7 +481,7 @@ UnixDateTime UnixDateTime::now_coarse()
 
 ErrorOr<void> UnixDateTime::to_string_impl(StringBuilder& builder, StringView format, LocalTime local_time) const
 {
-    struct tm tm;
+    struct tm tm {};
```
shannonbooth, line 484: It seems reasonable to just zero initialize this, because why not - but is this not a false positive by cppcheck, at least in the way that it is reporting the issue?

It does seem like we should be error checking the result of localtime_r/gmtime_r though for this to really be safe and for tm to always be initialized.

Please also adjust the commit message in the imperative mood

rcorsi: Issue is flagged by cppcheck only when AK_OS_WINDOWS is defined, due to lines 19-20 of AK/Time.cpp file where localtime_r is changed to localtime_s (and parameters swapped) and same for gmtime_r to gmtime_s.  If I comment out lines 19-20, I don't get the uninitvar warning. Is it a Windows specific issue, I can't find any information about struct tm initialization being a problem in Windows; or if the fact that the function names are changed and parameters swapped is confusing cppcheck. 

Yes initialization is mostly to make cppcheck happy and to be overly cautious. Though there seems to be some real corner case situations that might need it (non standard implementations with extra struct tm fields that might not get initialized, other functions using struct tm might check certain fields before they are set, timezone information not being set with tzset()). And of course if the called functions fail (and not checked), there would be some strange values in the fields.

Not sure why error checking is not happening in the code. Maybe due to different signatures of localtime_r/localtime_s and same for gmtime_r/gmtime_s to support both Windows and the other platforms made it messy.

rcorsi:
Hi @shannonbooth, what do you think, should this PR be merged? Is the commit message OK?

shannonbooth:
> Hi @shannonbooth, what do you think, should this PR be merged? Is the commit message OK?

I'm sort of indifferent to zero initializing here to silence the warning - maybe in combination we should change to VERIFY which is where we would actually have a problem by not intializing? It would require an ifdef branch though since windows returns a errno style error whereas unix a pointer. That way I think it would be a concrete improvement and fix the issue here?

 
```
static bool fill_tm(time_t timestamp, struct tm& tm, UnixDateTime::LocalTime local_time)
{
#ifdef AK_OS_WINDOWS
    if (local_time == UnixDateTime::LocalTime::Yes)
        return localtime_s(&tm, &timestamp) == 0;
    return gmtime_s(&tm, &timestamp) == 0;
#else
    if (local_time == UnixDateTime::LocalTime::Yes)
        return localtime_r(&timestamp, &tm) != nullptr;
    return gmtime_r(&timestamp, &tm) != nullptr;
#endif
}
```
type of thing

PR #8780, opened 2026-04-04
LibWeb: Handle @font-face within grouping rules like @layer and @media

### LINKED ISSUE #8685: @font-face declaration fails if contained in @layer block
### Summary

Website [https://www.bundestag.de/](https://www.bundestag.de/) fails to render navigation bar symbols which are code points rendered in the "bt-icons" font hosted on the website's server, declared in the style sheet with @font-face. After some debugging I learned that ladybird can only utilize the `@font-face` font if it is declared outside of the layer (`@layer legacy` in this case) block (if you move the test case's font declaration outside of the `@layer`, the symbol is rendered successfully).

### Operating system

Linux

### Steps to reproduce

1. Open [https://www.bundestag.de/](https://www.bundestag.de/).
2. Observe non-rendering code points in navigation, e.g. next to the menu item "Gebärdensprache".

### Expected behavior

Ladybird renders the code points in the "bt-icons" font.

### Actual behavior

Ladybird fails to render the symbols.

N/A

```html
<html lang="en">
<head>
    <meta charset="UTF-8">
    <title>Icon Font Render</title>
    <style>
        @layer legacy {
            @font-face {
                font-family: 'bt-icons';
                src: url("https://www.bundestag.de/resource/blob/596236/c2428bf72618eb8b14cf410c99248144/bt-icons-woff2.woff2") format("woff2"),
                url("https://www.bundestag.de/resource/blob/596234/859038b37f689c9c4f2eeb07b4dca8cb/bt-icons-woff.woff") format("woff");
                font-weight: 400;
                font-style: normal;
            }

            [class*=" icon-"]:before, [class^=icon-]:before {
                font-family: bt-icons;
            }

            /* 3. Inject the specific Unicode character via CSS (cleanest way) */
            .icon-sign-language:before {
                content: "\e868";
            }
        }
    </style>
</head>
<body>

<i class="bt-icon icon-sign-language" style="font-size: 2.4rem;"></i>
<p>Rendered via CSS <code>\e868</code></p>

</body>
</html>
```

```shell
dlopen: /usr/lib/libpangoft2-1.0.so.0: undefined symbol: FcConfigSetDefaultSubstitute
dlopen: /usr/lib/libpangoft2-1.0.so.0: undefined symbol: FcConfigSetDefaultSubstitute
dlopen: /usr/lib/libpangoft2-1.0.so.0: undefined symbol: FcConfigSetDefaultSubstitute
dlopen: /usr/lib/libpangoft2-1.0.so.0: undefined symbol: FcConfigSetDefaultSubstitute
dlopen: /usr/lib/libpangoft2-1.0.so.0: undefined symbol: FcConfigSetDefaultSubstitute
dlopen: /usr/lib/libpangoft2-1.0.so.0: undefined symbol: FcConfigSetDefaultSubstitute
```

_No response_

_No response_

### Contribute a patch?

fenilsonani:
FontComputer only iterated top-level rules when loading fonts from stylesheets, so @font-face declarations nested inside grouping rules like @layer were silently ignored. This adds a recursive helper that descends into grouping rules while respecting @media/@supports conditions.

Fixes #8685.

### DIFF
--- a/Libraries/LibWeb/CSS/FontComputer.cpp
+++ b/Libraries/LibWeb/CSS/FontComputer.cpp
@@ -792,49 +792,45 @@ GC::Ptr<FontLoader> FontComputer::load_font_face(ParsedFontFace const& font_face
 
 void FontComputer::load_fonts_from_sheet(CSSStyleSheet& sheet)
 {
-    // FIXME: Handle @font-face and @font-feature-values within grouping rules (@media, @supports, etc)
-    for (auto const& rule : sheet.rules()) {
-        if (auto* font_face_rule = as_if<CSSFontFaceRule>(*rule)) {
-            if (!font_face_rule->is_valid())
-                continue;
+    sheet.for_each_effective_font_face_rule([&](CSSFontFaceRule& font_face_rule) {
+        if (!font_face_rule.is_valid())
+            return;
 
-            auto font_face = FontFace::create_css_connected(document().realm(), *font_face_rule);
-            document().fonts()->add_css_connected_font(font_face);
+        auto font_face = FontFace::create_css_connected(document().realm(), font_face_rule);
+        document().fonts()->add_css_connected_font(font_face);
 
-            if (font_face->has_non_default_unicode_range()) {
-                // Register for matching, but defer loading until a rendered codepoint
-                // actually falls in this face's unicode-range.
-                register_font_face(font_face);
-            } else {
-                // NB: Load via FontFace::load(), to satisfy this requirement:
-                // https://drafts.csswg.org/css-font-loading/#font-face-load
-                // User agents can initiate font loads on their own, whenever they determine that a given font face is
-                // necessary to render something on the page. When this happens, they must act as if they had called the
-                // corresponding FontFace’s load() method described here.
-                font_face->load();
-            }
+        if (font_face->has_non_default_unicode_range()) {
+            // Register for matching, but defer loading until a rendered codepoint
+            // actually falls in this face's unicode-range.
+            register_font_face(font_face);
+        } else {
+            // NB: Load via FontFace::load(), to satisfy this requirement:
+            // https://drafts.csswg.org/css-font-loading/#font-face-load
+            // User agents can initiate font loads on their own, whenever they determine that a given font face is
+            // necessary to render something on the page. When this happens, they must act as if they had called the
+            // corresponding FontFace's load() method described here.
+            font_face->load();
         }
+    });
 
-        if (auto* font_feature_values_rule = as_if<CSSFontFeatureValuesRule>(*rule))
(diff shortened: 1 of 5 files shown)


Libraries/LibWeb/CSS/FontComputer.cpp
```diff
@@ -709,13 +710,27 @@ GC::Ptr<FontLoader> FontComputer::load_font_face(ParsedFontFace const& font_face
     return loader_ref;
 }
 
+static void for_each_font_related_rule(CSSRuleList& rules, Function<void(CSSRule&)> const& callback)
```
Calme1709: This should live with (and look like) the other `CSSStyleSheet::for_each_effective_rule` helpers [here](https://github.com/LadybirdBrowser/ladybird/blob/9526dfbef3b758f929bcb4c7c9356dd8ee521835/Libraries/LibWeb/CSS/CSSStyleSheet.cpp#L304)

Calme1709 (changes requested):
Thanks for looking at this!

We still have the issue that this only accounts for rules within condition rules whose conditions are met at stylesheet load time, e.g. we won't load a `@font-face` within a `@media` that doesn't match initially but does after a window resize. Can we leave a FIXME for this at least?

Calme1709:
Also if we could add a test for this that would be great :)

Libraries/LibWeb/CSS/FontComputer.cpp
```diff
@@ -693,13 +694,29 @@ GC::Ptr<FontLoader> FontComputer::load_font_face(ParsedFontFace const& font_face
...
 
+// FIXME: This only finds @font-face rules within condition rules whose conditions are met at stylesheet
+//        load time. We should re-evaluate when conditions change (e.g., @media after a window resize).
+static void for_each_font_related_rule(CSSRuleList& rules, Function<void(CSSRule&)> const& callback)
+{
+    for (auto& rule : rules) {
+        if (is<CSSFontFaceRule>(*rule) || is<CSSFontFeatureValuesRule>(*rule)) {
+            callback(*rule);
+        } else if (is<CSSConditionRule>(*rule)) {
+            auto& condition_rule = static_cast<CSSConditionRule&>(*rule);
```
AtkinsSJ: You can combine these with `as_if<>()`:
```c++
} else if (auto* condition_rule = as_if<CSSConditionRule>(*rule)) {
```
with the difference being `condition_rule` is now a pointer instead of a reference.

Same for the `CSSGroupingRule` case below.

AtkinsSJ (changes requested):
Hi! Welcome to Ladybird and thanks for working on this. :^)

I agree with Calme that `for_each_font_related_rule()` makes more sense as `CSSStyleSheet::for_each_effective_font_face_rule()`.

PR #8782, opened 2026-04-04
LibWeb: Make range input respect step attribute during mouse drag

vdemcak:
This PR fixes the issue where dragging the slider handle on range inputs with a set `step` attribute only snapped to integer values.

Fixes #8777

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLInputElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLInputElement.cpp
@@ -1417,8 +1417,22 @@ void HTMLInputElement::create_range_input_shadow_tree()
         double maximum = *max();
         if (minimum > maximum)
             return;
-        // FIXME: Snap new value to input steps
-        MUST(set_value_as_number(clamp(round(((client_x - rect.left().to_double()) / rect.width().to_double()) * (maximum - minimum) + minimum), minimum, maximum)));
+
+        double fraction = (client_x - rect.left().to_double()) / rect.width().to_double();
+        double value = fraction * (maximum - minimum) + minimum;
+
+        // Snap to allowed value step per the HTML spec for <input type=range>.
+        // https://html.spec.whatwg.org/multipage/input.html#range-state-(type=range)
+        if (auto maybe_allowed_value_step = allowed_value_step(); maybe_allowed_value_step.has_value()) {
+            double allowed_value_step = *maybe_allowed_value_step;
+            double base = step_base();
+
+            // Round to the nearest allowed step. If two numbers are equally close,
+            // choose the one nearest to positive infinity.
+            value = base + floor((value - base) / allowed_value_step + 0.5) * allowed_value_step;
+        }
+
+        MUST(set_value_as_number(clamp(value, minimum, maximum)));
         user_interaction_did_change_input_value();
     };
 
--- a/Libraries/LibWeb/HTML/HTMLInputElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLInputElement.cpp
@@ -13,7 +13,9 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
+#include <AK/StdLibExtras.h>
 #include <LibGfx/DecodedImageFrame.h>
+#include <LibGfx/ImmutableBitmap.h>
 #include <LibJS/Runtime/Date.h>
 #include <LibJS/Runtime/NativeFunction.h>
 #include <LibJS/Runtime/RegExpObject.h>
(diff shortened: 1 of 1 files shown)


Libraries/LibWeb/HTML/HTMLInputElement.cpp
```diff
@@ -1405,8 +1405,20 @@ void HTMLInputElement::create_range_input_shadow_tree()
...
+
+        double fraction = (client_x - rect.left().to_double()) / rect.width().to_double();
+        double value = fraction * (maximum - minimum) + minimum;
+
+        // https://html.spec.whatwg.org/multipage/input.html#range-state-(type=range)
+        if (auto maybe_allowed_value_step = allowed_value_step(); maybe_allowed_value_step.has_value()) {
+            double allowed_value_step = *maybe_allowed_value_step;
+            double base = step_base();
+            
+            // "If two numbers match these constraints, then user agents must use the one nearest to positive infinity"
```
Psychpsyo: No quotes around spec text. Also, this comment feels hard to follow given that the constraints it talks about aren't also included in the quote and it's a few lines away from the actual spec link.

vdemcak:
Hey, sorry for the long delay.

A bunch of stuff came up on my end and I didn't have time to get back to this until now.

I've rebased on master, resolved the merge conflict, and updated the code + comment based on your feedback.

Libraries/LibWeb/HTML/HTMLInputElement.cpp
```diff
@@ -1417,8 +1417,22 @@ void HTMLInputElement::create_range_input_shadow_tree()
         double maximum = *max();
         if (minimum > maximum)
             return;
-        // FIXME: Snap new value to input steps
-        MUST(set_value_as_number(clamp(round(((client_x - rect.left().to_double()) / rect.width().to_double()) * (maximum - minimum) + minimum), minimum, maximum)));
+
+        double fraction = (client_x - rect.left().to_double()) / rect.width().to_double();
+        double value = fraction * (maximum - minimum) + minimum;
```
Zaggy1024: Can use mix() (from StdLibExtras.h) here instead.

Libraries/LibWeb/HTML/HTMLInputElement.cpp
```diff
@@ -1417,8 +1417,22 @@ void HTMLInputElement::create_range_input_shadow_tree()
...
+        if (auto maybe_allowed_value_step = allowed_value_step(); maybe_allowed_value_step.has_value()) {
+            double allowed_value_step = *maybe_allowed_value_step;
+            double base = step_base();
+
+            // Round to the nearest allowed step. If two numbers are equally close,
+            // choose the one nearest to positive infinity.
+            value = base + floor((value - base) / allowed_value_step + 0.5) * allowed_value_step;
+        }
+
+        MUST(set_value_as_number(clamp(value, minimum, maximum)));
```
Zaggy1024, line 1437: I believe that set_value_as_number() also rounds to the nearest step, so the code adjusting the value seems redundant, unless it has different behavior?

vdemcak: I've checked the impl. and `set_value_as_number()` ends up calling `convert_number_to_string()`, which for `type=range` is currently just:

```c++
if (type_state() == TypeAttributeState::Range)
    return Utf16String::number(input);
```

Libraries/LibWeb/HTML/HTMLInputElement.cpp
vdemcak: Fixed.

Libraries/LibWeb/HTML/HTMLInputElement.cpp
vdemcak: Fixed

Libraries/LibWeb/HTML/HTMLInputElement.cpp
Zaggy1024: It's in `set_value()`, it converts the string back to a number and runs some validation on it, including snapping to the step.

Zaggy1024:
Looks like your rebase left some "Conflicts" lines in the commit messages, could you remove those?

Libraries/LibWeb/HTML/HTMLInputElement.cpp
vdemcak: Apologies, didn't notice, that `value_sanitization_algorithm` contained snapping logic. Should I edit the code to just be this?

```cpp
double fraction = (client_x - rect.left().to_double()) / rect.width().to_double();
double value = mix(minimum, maximum, fraction);

MUST(set_value_as_number(clamp(value, minimum, maximum)));
```

Zaggy1024: It'll also clamp it, iirc, so it could be as simple as

```cpp
double fraction = (client_x - rect.left().to_double()) / rect.width().to_double();
double value = mix(minimum, maximum, fraction);

MUST(set_value_as_number(value));
```

I'd suggest it might be worth adding a test that simulates a click on a range input element to test this functionality (using `internals.click(x, y)`, if there isn't one already. The test could be added as the first commit in the PR to demonstrate that the click can produce more granular values after the fix.

PR #8785, opened 2026-04-05
LibWeb: Distribute content into dynamic height multicol containers

Psychpsyo:
Builds upon #8557 to implement multicol containers with dynamic height, where such height is expected to be derived via balancing the content equally between all columns as far as possible. (unless `column-fill` is set to `auto`)

`column-fill` had to be implemented in the same go, or else one of the existing tests would've been broken by the new balancing behavior. (Before this patch, our behavior was always `column-fill: auto`)

Some newly passing multicol tests have also been imported.


gmta:

PR #8790, opened 2026-04-05
LibWebView+Base: Add reset to defaults for browser settings

AdenMGB:
## Summary

Adds **Reset to defaults** for Ladybird’s main browser settings. The about:settings page gets a header action and a confirmation dialog before applying changes.

## Implementation

- Restores all persisted fields to the same defaults as a fresh `Settings` instance (new tab URL, bookmarks bar, zoom, languages, search/autocomplete, autoplay site settings, browsing data / disk cache preferences, global privacy control, DNS when not forced by the shell).
- If DNS was applied with the command-line override flag, that DNS configuration is **not** cleared on reset so launch behavior stays consistent with how the browser was started.
- After persisting, all `SettingsObserver` hooks are invoked so chrome, request server (disk cache / DNS), and views stay aligned.
- New `resetToDefaults` WebUI message runs reset then `loadCurrentSettings` so the page refreshes.
- Header button, modal dialog (copy explains that custom search engines are removed and browsing data is **not** deleted).

## What it doesn't do

- Does not clear cookies, history, site data, or cache contents (only resets **settings** such as cache size limits).
- Does not reset Qt shell `QSettings` (window geometry / menubar), if applicable to the build.

## How to test

1. Change several settings (for example new tab URL, zoom, search engine, DNS in the UI).
2. Open about:settings → **Reset to defaults** → confirm.
3. Confirm UI and behavior match defaults and `Settings.json` reflects defaults (with DNS unchanged when started with CLI DNS override).
4. Confirm observers behave (for example bookmarks bar visibility, network/DNS after reset).

## EDIT: Image
{456661E5-B45A-49B6-94A8-0414F44D4123}

### DIFF
--- a/Base/res/ladybird/about-pages/settings.html
+++ b/Base/res/ladybird/about-pages/settings.html
@@ -18,6 +18,47 @@
                 }
             }
 
+            body.settings header {
+                justify-content: space-between;
+                gap: 16px;
+            }
+
+            body.settings .header-brand {
+                display: flex;
+                align-items: center;
+            }
+
+            /* Closed <dialog> must stay hidden; display:flex on dialog overrides UA :not([open]). */
+            dialog.reset-settings-dialog:not([open]) {
+                display: none;
+            }
+
+            dialog.reset-settings-dialog[open] {
+                display: flex;
+                flex-direction: column;
+
+                width: min(560px, 94vw);
+                max-width: min(560px, 94vw);
+            }
+
+            dialog.reset-settings-dialog .dialog-body {
+                flex: 0 1 auto;
+                height: auto;
+                min-height: min-content;
+                max-height: none;
+
+                overflow: visible;
+                overflow-wrap: break-word;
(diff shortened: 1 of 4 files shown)


trflynn89:
There used to be a button to reset all settings that was explicitly removed in #8272. Such a button makes less sense as more options are added - it makes more sense to allow resetting individual settings IMO.

jdahlin:
For what its worth, other browser do this:
- Chrome has a "Reset settings" in chrome://settings
- Firefox has a "Refresh firefox..." button in about:support
- Safari has only per-site and profiles

AdenMGB:
Intresting, the reason I decided to add this is because it's my first time contributing and wanted something a bit easier to start. I can see ladybird doesn't have many settings right now and it's very developer oriented with changing parameters in the CLI, however if and when this ever reaches the masses I feel like it will be helpful for them to have an easy one button reset all, that's why I'm starting the foundation.

PR #8797, opened 2026-04-05
LibWeb: Use caption content width for inline layout in table captions

tcl3:
Previously, `run_caption_layout()` passed the table's border-box width as the available space to the caption's formatting context. The BFC then used this width directly for inline line breaking, causing text to overflow the caption's content box by the size of the caption's own border and padding.

This fixes an issue where text would occasionally overflow Wikipedia infoboxes.

Here's an example from https://en.wikipedia.org/wiki/Software_bug#Prevention:

Before:

image

After:

image


awesomekling:
```
Fail: Layout/input/table/table-caption-inline-width.html
```
Seems relevant!

tcl3:
For ref and crash tests we run some JS that waits for fonts to load then waits for 2 rAFs. We weren't doing this for layout tests, so the test I added sometimes failed because the font hadn't loaded.

I've made it so layout tests use the same waiting mechanism as other test types.

tcl3:
I have no idea why the flatpak CI jobs are having such a bad time on this PR - it doesn't seem related to my change though.

AtkinsSJ (approved):

PR #8802, opened 2026-04-06
LibWeb: Add missing WebGL2 methods and drawing buffer attributes

TurboRigby:
Add critical WebGL2 functionality to improve compatibility:

WebGL2 methods:
- copyTexSubImage3D: Copy 3D texture data from framebuffer
- getFragDataLocation: Get fragment shader output location
- isQuery, isSampler, isSync, isTransformFeedback: Object validation
- getSamplerParameter: Get sampler object parameters
- getTransformFeedbackVarying: Get transform feedback varying info
- getIndexedParameter: Get indexed state values

WebGL base functionality:
- getFramebufferAttachmentParameter: Get framebuffer attachment info
- drawingBufferFormat: Drawing buffer pixel format (RGBA8)
- drawingBufferColorSpace/unpackColorSpace: Color space attributes

All methods include proper OpenGL error handling and validation. Drawing buffer attributes use PredefinedColorSpace enum with sRGB default.

### DIFF
--- a/Libraries/LibWeb/WebGL/WebGL2RenderingContextImpl.cpp
+++ b/Libraries/LibWeb/WebGL/WebGL2RenderingContextImpl.cpp
@@ -1347,4 +1347,213 @@ void WebGL2RenderingContextImpl::compressed_tex_sub_image3d(WebIDL::UnsignedLong
     glCompressedTexSubImage3DRobustANGLE(target, level, xoffset, yoffset, zoffset, width, height, depth, format, pixels.size(), pixels.size(), pixels.data());
 }
 
+void WebGL2RenderingContextImpl::copy_tex_sub_image3d(WebIDL::UnsignedLong target, WebIDL::Long level, WebIDL::Long xoffset, WebIDL::Long yoffset, WebIDL::Long zoffset, WebIDL::Long x, WebIDL::Long y, WebIDL::Long width, WebIDL::Long height)
+{
+    m_context->make_current();
+    glCopyTexSubImage3D(target, level, xoffset, yoffset, zoffset, x, y, width, height);
+}
+
+WebIDL::Long WebGL2RenderingContextImpl::get_frag_data_location(GC::Root<WebGLProgram> program, String const& name)
+{
+    m_context->make_current();
+    
+    if (!program) {
+        set_error(GL_INVALID_VALUE);
+        return -1;
+    }
+
+    auto handle_or_error = program->handle(this);
+    if (handle_or_error.is_error()) {
+        set_error(GL_INVALID_OPERATION);
+        return -1;
+    }
+
+    auto handle = handle_or_error.release_value();
+    auto name_null_terminated = null_terminated_string(name);
+    return glGetFragDataLocation(handle, name_null_terminated.data());
+}
+
+bool WebGL2RenderingContextImpl::is_query(GC::Root<WebGLQuery> query)
+{
+    m_context->make_current();
+    
+    if (!query)
+        return false;
+
+    auto handle_or_error = query->handle(this);
+    if (handle_or_error.is_error())
+        return false;
+
+    return glIsQuery(handle_or_error.release_value());
+}
+
+bool WebGL2RenderingContextImpl::is_sampler(GC::Root<WebGLSampler> sampler)
+{
+    m_context->make_current();
+    
+    if (!sampler)
+        return false;
+
+    auto handle_or_error = sampler->handle(this);
+    if (handle_or_error.is_error())
+        return false;
+
+    return glIsSampler(handle_or_error.release_value());
+}
+
+JS::Value WebGL2RenderingContextImpl::get_sampler_parameter(GC::Root<WebGLSampler> sampler, WebIDL::UnsignedLong pname)
+{
+    m_context->make_current();
+    
+    if (!sampler) {
+        set_error(GL_INVALID_VALUE);
+        return JS::js_null();
+    }
+
+    auto handle_or_error = sampler->handle(this);
+    if (handle_or_error.is_error()) {
+        set_error(GL_INVALID_OPERATION);
+        return JS::js_null();
+    }
+
+    auto handle = handle_or_error.release_value();
+    switch (pname) {
+    case GL_TEXTURE_COMPARE_FUNC:
+    case GL_TEXTURE_COMPARE_MODE:
+    case GL_TEXTURE_MAG_FILTER:
+    case GL_TEXTURE_MIN_FILTER:
+    case GL_TEXTURE_WRAP_R:
+    case GL_TEXTURE_WRAP_S:
+    case GL_TEXTURE_WRAP_T: {
+        GLint value;
+        glGetSamplerParameteriv(handle, pname, &value);
+        return JS::Value(value);
+    }
+    case GL_TEXTURE_MAX_LOD:
+    case GL_TEXTURE_MIN_LOD: {
+        GLfloat value;
+        glGetSamplerParameterfv(handle, pname, &value);
+        return JS::Value(value);
+    }
+    default:
+        set_error(GL_INVALID_ENUM);
+        return JS::js_null();
+    }
+}
+
+bool WebGL2RenderingContextImpl::is_sync(GC::Root<WebGLSync> sync)
+{
+    m_context->make_current();
+    
+    if (!sync)
+        return false;
+
+    auto sync_handle_or_error = sync->sync_handle(this);
+    if (sync_handle_or_error.is_error())
+        return false;
+
+    auto sync_handle = sync_handle_or_error.release_value();
+    return glIsSync(static_cast<GLsync>(sync_handle));
+}
+
+bool WebGL2RenderingContextImpl::is_transform_feedback(GC::Root<WebGLTransformFeedback> transform_feedback)
+{
+    m_context->make_current();
+    
+    if (!transform_feedback)
+        return false;
+
+    auto handle_or_error = transform_feedback->handle(this);
+    if (handle_or_error.is_error())
+        return false;
+
+    return glIsTransformFeedback(handle_or_error.release_value());
+}
+
+GC::Root<WebGLActiveInfo> WebGL2RenderingContextImpl::get_transform_feedback_varying(GC::Root<WebGLProgram> program, WebIDL::UnsignedLong index)
+{
+    m_context->make_current();
+    
+    if (!program) {
+        set_error(GL_INVALID_VALUE);
+        return {};
+    }
+
+    auto handle_or_error = program->handle(this);
+    if (handle_or_error.is_error()) {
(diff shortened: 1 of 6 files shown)


Libraries/LibWeb/WebGL/WebGL2RenderingContextImpl.cpp
```diff
@@ -1347,4 +1347,194 @@ void WebGL2RenderingContextImpl::compressed_tex_sub_image3d(WebIDL::UnsignedLong
...
+    }
+
+    auto handle_or_error = program->handle(this);
+    if (handle_or_error.is_error()) {
+        set_error(GL_INVALID_OPERATION);
+        return -1;
+    }
+
+    auto handle = handle_or_error.release_value();
+    return glGetFragDataLocation(handle, name.to_byte_string().characters());
```
cqundefine: Fwiw most all current WebGL code that passes a string uses the `null_terminated_string` helper that returns a `Vector<GLchar>`.

Libraries/LibWeb/WebGL/WebGL2RenderingContextImpl.cpp
```diff
@@ -1347,4 +1347,194 @@ void WebGL2RenderingContextImpl::compressed_tex_sub_image3d(WebIDL::UnsignedLong
...
+
+    auto sync_handle_or_error = sync->sync_handle(this);
+    if (sync_handle_or_error.is_error())
+        return false;
+
+    auto sync_handle = sync_handle_or_error.release_value();
+    return glIsSync(static_cast<GLsync>(sync_handle));
+}
+
+bool WebGL2RenderingContextImpl::is_transform_feedback(GC::Root<WebGLTransformFeedback> tf)
```
cqundefine: The Ladybird codebase uses full names instead of abbreviations like the `tf` here.

Libraries/LibWeb/WebGL/WebGL2RenderingContextImpl.cpp
```diff
@@ -1347,4 +1347,194 @@ void WebGL2RenderingContextImpl::compressed_tex_sub_image3d(WebIDL::UnsignedLong
...
+    name_buffer.resize(max_name_length);
+
+    glGetTransformFeedbackVarying(handle, index, max_name_length, &length, &size, &type, name_buffer.data());
+
+    if (length == 0) {
+        set_error(GL_INVALID_VALUE);
+        return {};
+    }
+
+    String name = String::from_utf8(StringView(name_buffer.data(), length)).release_value_but_fixme_should_propagate_errors();
```
cqundefine: You can safely use `from_utf8_without_validation` and avoid the awkward error handling here (all other WebGL functions do it like that).

Libraries/LibWeb/WebGL/WebGL2RenderingContextImpl.cpp
```diff
@@ -1347,4 +1347,194 @@ void WebGL2RenderingContextImpl::compressed_tex_sub_image3d(WebIDL::UnsignedLong
...
+JS::Value WebGL2RenderingContextImpl::get_indexed_parameter(WebIDL::UnsignedLong target, WebIDL::UnsignedLong index)
+{
+    m_context->make_current();
+    
+    switch (target) {
+    case GL_TRANSFORM_FEEDBACK_BUFFER_BINDING:
+    case GL_UNIFORM_BUFFER_BINDING: {
+        GLint value = 0;
+        glGetIntegeri_v(target, index, &value);
+        return JS::Value(value);
```
cqundefine: In this case we are supposed to return a value of the type `WebGLBuffer`, you could look at how other functions do stuff like this or just leave it as a FIXME for now if you don't wanna figure it out.

Libraries/LibWeb/WebGL/WebGL2RenderingContextImpl.cpp
```diff
@@ -1347,4 +1347,194 @@ void WebGL2RenderingContextImpl::compressed_tex_sub_image3d(WebIDL::UnsignedLong
...
+    }
+    case GL_TRANSFORM_FEEDBACK_BUFFER_START:
+    case GL_TRANSFORM_FEEDBACK_BUFFER_SIZE:
+    case GL_UNIFORM_BUFFER_START:
+    case GL_UNIFORM_BUFFER_SIZE: {
+        GLint64 value = 0;
+        glGetInteger64i_v(target, index, &value);
+        return JS::Value(static_cast<double>(value));
```
cqundefine, line 1551: The WebGL 2.0 spec states that "If an OpenGL error is generated, returns null.", while a bunch of functions have clauses like that, which are basically unreachable, in this case according to the OpenGL ES 3.0 spec "An INVALID_VALUE error is generated if index is outside the valid range for the indexed state target." which is something that could happen and we should probably check it using glGetError and return null in such case.

Libraries/LibWeb/WebGL/WebGL2RenderingContextImpl.h
```diff
@@ -105,6 +105,17 @@ class WebGL2RenderingContextImpl : public WebGLRenderingContextImpl {
     void bind_vertex_array(GC::Root<WebGLVertexArrayObject> array);
     void compressed_tex_image3d(WebIDL::UnsignedLong target, WebIDL::Long level, WebIDL::UnsignedLong internalformat, WebIDL::Long width, WebIDL::Long height, WebIDL::Long depth, WebIDL::Long border, GC::Root<WebIDL::ArrayBufferView> src_data, WebIDL::UnsignedLongLong src_offset, WebIDL::UnsignedLong src_length_override);
     void compressed_tex_sub_image3d(WebIDL::UnsignedLong target, WebIDL::Long level, WebIDL::Long xoffset, WebIDL::Long yoffset, WebIDL::Long zoffset, WebIDL::Long width, WebIDL::Long height, WebIDL::Long depth, WebIDL::UnsignedLong format, GC::Root<WebIDL::ArrayBufferView> src_data, WebIDL::UnsignedLongLong src_offset, WebIDL::UnsignedLong src_length_override);
+
+    // Missing WebGL2 methods implementation
```
cqundefine: Those function are kept in the same order as they are in the IDL file, also this comment doesn't make much sense.

Libraries/LibWeb/WebGL/WebGLRenderingContextImpl.cpp
```diff
@@ -2392,6 +2395,67 @@ void WebGLRenderingContextImpl::viewport(WebIDL::Long x, WebIDL::Long y, WebIDL:
     glViewport(x, y, width, height);
 }
 
+JS::Value WebGLRenderingContextImpl::get_framebuffer_attachment_parameter(WebIDL::UnsignedLong target, WebIDL::UnsignedLong attachment, WebIDL::UnsignedLong pname)
+{
+    m_context->make_current();
+
+    // Validate the target
+    if (target != GL_FRAMEBUFFER) {
```
cqundefine: That is only correct for WebGL 1.0 / OpenGL ES 2.0, this implementation is shared for both WebGL versions, also this requirement is not coming from WebGL but from OpenGL ES, so we can just rely on the underlaying driver to validate this for us.

Libraries/LibWeb/WebGL/WebGLRenderingContextImpl.cpp
```diff
@@ -2392,6 +2395,67 @@ void WebGLRenderingContextImpl::viewport(WebIDL::Long x, WebIDL::Long y, WebIDL:
...
+        set_error(GL_INVALID_ENUM);
+        return JS::js_null();
+    }
+
+    // Get the OpenGL parameter value
+    GLint result = 0;
+    glGetFramebufferAttachmentParameteriv(target, attachment, pname, &result);
+
+    // Check for OpenGL errors
```
cqundefine: The comments in this function are very verbose and unnecessary.

Libraries/LibWeb/WebGL/WebGLRenderingContextImpl.cpp
```diff
@@ -2392,6 +2395,67 @@ void WebGLRenderingContextImpl::viewport(WebIDL::Long x, WebIDL::Long y, WebIDL:
...
+    GLenum gl_error = glGetError();
+    if (gl_error != GL_NO_ERROR) {
+        set_error(gl_error);
+        return JS::js_null();
+    }
+
+    // Return appropriate JavaScript value based on parameter type
+    switch (pname) {
```
cqundefine, line 2411: Again, this is missing a bunch of WebGL 2.0 stuff.

Libraries/LibWeb/WebGL/WebGLRenderingContextImpl.cpp
```diff
@@ -2392,6 +2395,67 @@ void WebGLRenderingContextImpl::viewport(WebIDL::Long x, WebIDL::Long y, WebIDL:
...
+    case GL_FRAMEBUFFER_ATTACHMENT_OBJECT_TYPE:
+    case GL_FRAMEBUFFER_ATTACHMENT_TEXTURE_LEVEL:
+    case GL_FRAMEBUFFER_ATTACHMENT_TEXTURE_CUBE_MAP_FACE:
+        return JS::Value(result);
+    case GL_FRAMEBUFFER_ATTACHMENT_OBJECT_NAME:
+        // This should return the WebGL object, but for now return the handle
```
cqundefine: It probably makes more sense to return a null here instead as we shouldn't leak OpenGL ES stuff to the web as the raw handles are normally inaccessible. You can just leave it as FIXME.

Libraries/LibWeb/WebGL/WebGLRenderingContextImpl.cpp
```diff
@@ -2392,6 +2395,67 @@ void WebGLRenderingContextImpl::viewport(WebIDL::Long x, WebIDL::Long y, WebIDL:
...
+        return JS::Value(result);
+    default:
+        return JS::Value(result);
```
cqundefine: We should set the error to `GL_INVALID_ENUM` and return null if we don't recognize the `pname`.

Libraries/LibWeb/WebGL/WebGLRenderingContextImpl.h
```diff
@@ -16,6 +16,10 @@
 #include <LibWeb/WebIDL/ExceptionOr.h>
 #include <LibWeb/WebIDL/Types.h>
 
+namespace Web::Bindings {
+enum class PredefinedColorSpace : u8;
```
cqundefine: Nitpick: This is kinda sketchy, we should probably just include the header that defines that.

cqundefine:
Mostly looks good, one last thing is you should merge those two commits together using `git commit --amend`, or because you already commited you can use interactive rebase (`git rebase -i`) to merge them and then force push to your branch. If you never messed around with git history there's a video about that floating somewhere on the Ladybird Discord server.

PR #8803, opened 2026-04-06
Visually slice more fragmented content

Psychpsyo:
This builds on top of #8771 to also properly slice monolithic content such as images.
It does this by rendering the entire element in every fragmentainer and just clipping away the parts that should be fragmented. This gives some correctness improvements around inline fragmentation with box shadows, bringing us on par with Chromium and WebKit, and ahead of Firefox.

Although border painting still 'sidesteps' this system by doing the fragmentation itself, since I felt that properly piping in the original dimensions of inline elements (which this would need) is out-of-scope and a thing I'd rather tackle in a follow-up PR. Maybe when I take a look at implementing `box-decoration-break`.

Also, Paintables other than ImagePaintable will need to have their rendering code adjusted in similar ways images have here. I might wrap that into this PR still, depending on how long it is until this gets reviewed. (will edit this comment if that's the case)


gmta:

PR #8813, opened 2026-04-07
LibWeb: Implement SVGTransform mutation methods

Trojan2021:
Implements the mutation methods for SVGTransform:
  - setTranslate
  - setScale
  - setRotate
  - setSkewX
  - setSkewY

Spec: https://svgwg.org/svg2-draft/single-page.html#coords-InterfaceSVGTransform


Calme1709:
Thanks for looking at this!

These are spec methods so we should be following the spec steps here, these can be found in the spec document you linked above or in the links pasted above each of the method definitions.

We will also need some tests for this, I assume there are WPT ones that we can import

PR #8819, opened 2026-04-07
LibMedia+LibWeb: Handle media with unknown duration or size

Lubrsi:
In this case, we report +Infinity duration for the media per the spec.
This can occur for media with no definitive duration, such as MP3s
streamed from radio servers (e.g. via Icecast)

Plus a drive-by RequestServer fix to not burn CPU when receiving infinite responses.

### DIFF
--- a/Libraries/LibMedia/FFmpeg/FFmpegIOContext.cpp
+++ b/Libraries/LibMedia/FFmpeg/FFmpegIOContext.cpp
@@ -56,8 +56,10 @@ ErrorOr<NonnullOwnPtr<FFmpegIOContext>> FFmpegIOContext::create(NonnullRefPtr<Me
             whence &= ~AVSEEK_FORCE;
 
             auto& stream_cursor = *static_cast<MediaStreamCursor*>(opaque);
-            if (whence == AVSEEK_SIZE)
-                return stream_cursor.size();
+            if (whence == AVSEEK_SIZE) {
+                auto size = stream_cursor.expected_size();
+                return size.has_value() ? static_cast<int64_t>(size.value()) : -1;
+            }
 
             auto seek_mode_from_whence = [](int origin) -> AK::SeekMode {
                 if (origin == SEEK_CUR)
--- a/Libraries/LibMedia/IncrementallyPopulatedStream.cpp
+++ b/Libraries/LibMedia/IncrementallyPopulatedStream.cpp
@@ -265,15 +265,27 @@ DecoderErrorOr<void> IncrementallyPopulatedStream::Cursor::seek(i64 offset, Seek
     case SeekMode::FromCurrentPosition:
         m_position += offset;
         break;
-    case SeekMode::FromEndPosition:
-        m_position = this->size() + offset;
+    case SeekMode::FromEndPosition: {
+        auto size = m_stream->expected_size();
+        if (!size.has_value())
+            return DecoderError::with_description(DecoderErrorCategory::EndOfStream, "Stream size is unknown, cannot seek from end"sv);
+        m_position = size.value() + offset;
         break;
+    }
     default:
         VERIFY_NOT_REACHED();
     }
 
     m_active_timeout = MonotonicTime::now_coarse() + CURSOR_ACTIVE_TIME;
     return {};
+}
+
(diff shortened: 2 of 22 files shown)


Libraries/LibMedia/FFmpeg/FFmpegIOContext.cpp
```diff
@@ -56,8 +56,10 @@ ErrorOr<NonnullOwnPtr<FFmpegIOContext>> FFmpegIOContext::create(NonnullRefPtr<Me
             whence &= ~AVSEEK_FORCE;
 
             auto& stream_cursor = *static_cast<MediaStreamCursor*>(opaque);
-            if (whence == AVSEEK_SIZE)
-                return stream_cursor.size();
+            if (whence == AVSEEK_SIZE) {
+                auto size = stream_cursor.expected_size();
+                return size.has_value() ? static_cast<int64_t>(size.value()) : -1;
+            }
```
Zaggy1024, line 62: This feels like it could be a separate commit, but hopefully all of FFmpeg's format readers know how to deal with this properly. I think we had some issues with not returning the actual size before, might be worth testing how MP4 deals with this especially.

Libraries/LibMedia/IncrementallyPopulatedStream.cpp
```diff
@@ -265,9 +265,13 @@ DecoderErrorOr<void> IncrementallyPopulatedStream::Cursor::seek(i64 offset, Seek
     case SeekMode::FromCurrentPosition:
         m_position += offset;
         break;
-    case SeekMode::FromEndPosition:
-        m_position = this->size() + offset;
+    case SeekMode::FromEndPosition: {
+        auto size = m_stream->expected_size();
+        if (!size.has_value())
+            return DecoderError::with_description(DecoderErrorCategory::EndOfStream, "Stream size is unknown, cannot seek from end"sv);
```
Zaggy1024, line 271: I'd be curious when exactly this error might be hit for the formats we demux through FFmpeg, and whether it may cause issues. Also, EndOfStream feels incorrect here, perhaps a category of Invalid would make more sense if it doesn't cause issues up the pipeline.

Libraries/LibMedia/IncrementallyPopulatedStream.cpp
```diff
@@ -276,6 +280,14 @@ DecoderErrorOr<void> IncrementallyPopulatedStream::Cursor::seek(i64 offset, Seek
     return {};
 }
 
+Optional<size_t> IncrementallyPopulatedStream::Cursor::expected_size() const
+{
+    auto size = m_stream->expected_size();
+    if (!size.has_value())
+        return {};
+    return size.value();
```
Zaggy1024, line 288: Is there a reason this is returning size_t and forcing us to unwrap the Optional?

Zaggy1024: (Perhaps if it's more consistent to use size_t here, we should change the underlying type instead, not sure)

Libraries/LibMedia/PlaybackManager.cpp
```diff
@@ -76,7 +76,7 @@ DecoderErrorOr<void> PlaybackManager::prepare_playback_from_demuxer(WeakPlayback
     if (preferred_audio_track.has_value() && !supported_audio_tracks.contains_slow(*preferred_audio_track))
         preferred_audio_track = {};
 
-    auto duration = demuxer->total_duration().value_or(AK::Duration::zero());
+    auto duration = TRY(demuxer->total_duration());
```
Zaggy1024, line 79: I think the swallowing the error here was intentional, we should probably just do `value_or(OptionalNone())`, at least for now.

Zaggy1024:
I'm thinking that instead of adding Optional everywhere, it might make more sense to treat `Duration::min()` and `Duration::max()` as negative and positive infinity respectively. MSE expects our internal representation of timestamps to have arithmetic behaving like double precision floating point, so that would help us achieve that more closely as well. (Browsers vary on how they handle this, so it's not such a strict thing, but handling overflow as infinities is probably logical.)

Zaggy1024:
For the moment, I think it makes sense to just directly use `Duration::max()` to represent infinity for the purposes of media duration, and then we can rename things later.

PR #8824, opened 2026-04-08
Meta: Update license copyright year to 2026

Craftefix:
The end year in the copyright range was still 2025.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 BSD 2-Clause License
 
-Copyright (c) 2018-2025, the Ladybird developers.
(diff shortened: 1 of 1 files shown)


rcorsi:
Hi @Craftefix, to resolve the commit message issue, you need to update the 'git commit message' not the title of the PR.  Use `git commit --amend` if working on the command line and put the same text as you put in this PR title. Then you need to push again, maybe with a 'force' option as in `git push -f`

Zaggy1024:
Probably worth using `2018-present` here instead, we're generally replacing the current year with `present` in new files.

PR #8832, opened 2026-04-08
Rectify 60+ typos in 50+ files - minimise risk of new typos thanks to CI & pre-commit

Sajjon:
# Changes
Rectify 60+ typos in 50+ files - found and fixed by excellent [`typos` cli tool](https://github.com/crate-ci/typos)
Add `_typos.toml` config which was used to produce this commit.
**pre-commit** and **CI**  has been updated to use typos - to prevent new typos.

# Manual work
It was cirka 90 minutes manual work building up the `_typos.toml` config, which has resulted in the good shape it is in now. 

# Typos CLI
`typos` CLI tool is fast, it takes less than 2 seconds to scan the whole code base on my Macbook Pro M2, so this is not at all a costly step - perfect for pre-commit and CI.

# Tests
I've tested with `ctest --preset Release` and all tests pass.

```
100% tests passed, 0 tests failed out of 187

Total Test time (real) = 154.75 sec
```


Libraries/LibWeb/ContentSecurityPolicy/BlockingAlgorithms.cpp
```diff
@@ -163,7 +163,7 @@ Directives::Directive::Result should_request_be_blocked_by_integrity_policy(GC::
     // 6. Let reportPolicy be policyContainer’s report only integrity policy.
     auto const& report_policy = policy_container->report_only_integrity_policy;
 
-    // 7. If both policy and reportPolicy are empty integrity policys, return "Allowed".
+    // 7. If both policy and reportPolicy are empty integrity policies, return "Allowed".
```
Psychpsyo: Not a typo, this comes straight from the spec.

Sajjon: I will revert all occurrences

Psychpsyo: I also worry that auto-linting this sort of thing will mean that the pre-commit hooks or CI will complain about more perfectly valid (or at least correctly quoted) spec text in the future, which'll incentivize people to slightly mis-quote the spec rather than to figure out how to maintain an ever-growing list of exceptions.

Whether or not that's worth it should be up to the maintainers though.

Libraries/LibWeb/CSS/StyleValues/PendingSubstitutionStyleValue.h
```diff
@@ -21,7 +21,7 @@ class PendingSubstitutionStyleValue final : public StyleValueWithDefaultOperator
     virtual void serialize(StringBuilder&, SerializationMode) const override { }
     virtual Vector<Parser::ComponentValue> tokenize() const override
     {
-        // Not sure what to do here, but this isn't valid so returning GIV seems the most correct.
+        // Not sure what to do here, but this isn't valid so returning `GuaranteedInvalidValue` seems the most correct.
```
Psychpsyo: If we're already changing this to be written out as words, I'd prefer "guaranteed-invalid value" since that is the English name of the thing.

Libraries/LibWeb/ContentSecurityPolicy/BlockingAlgorithms.cpp
Sajjon: Typos CLI tool has quite low false negative rate, and the big work has already been done with this PR.

Any new false negative is a single edit of _typos.toml, which should be quickly done.

.github/workflows/lagom-template.yml
```diff
@@ -60,6 +60,12 @@ jobs:
           arch: ${{ inputs.arch }}
           toolchain: ${{ inputs.toolchain }}
 
+      # === SPELL CHECK ===
+      - name: Spell Check with Typos
+        uses: crate-ci/typos@v1.43.5
+        with:
+          config: _typos.toml
```
gmta: This should go in the lint workflow (look at `Meta/lint-ci.sh`), not the build.

gmta (commented):
Not sure if we should have this, but we should at least put it in the right place if we're going to have it.

.github/workflows/lagom-template.yml
Sajjon: Thx! I will move it

Libraries/LibWeb/ContentSecurityPolicy/BlockingAlgorithms.cpp
trflynn89: I'm not particularly thrilled about having to maintain this typos file. Feels low value for the annoyance it will probably cause.

.github/workflows/lagom-template.yml
Sajjon: I've changed to install typos part of `.github/actions/setup/action.yml` and updated `Meta/lint-ci.sh` to use it if it is installed, else error.

Libraries/LibWeb/ContentSecurityPolicy/BlockingAlgorithms.cpp
Sajjon: I've been using typos CLI tool for all my projects the past 3 years - none of them was as big a code base as ladybird (but [Sargon](https://github.com/radixdlt/sargon) is not small) - and my experience is that the _typos.toml needs to be updated quite seldomly ([22 times in 2.5 years](https://github.com/radixdlt/sargon/commits/main/_typos.toml)).

From Ladybirds [CONTRIBUTING.md](https://github.com/LadybirdBrowser/ladybird/blob/master/CONTRIBUTING.md)
> In Ladybird, we treat human language as seriously as we do programming language.
and
>  Use proper spelling, grammar, and punctuation.

I've "paid the cost" with my work, maintenance cost is near zero, and the benefit is low likelihood of future typos. 

Also from CONTRIBUTING.md:
> Choose expressive variable, function and class names. Make it as obvious as possible what the code is doing.
Any flag by typos is likely to either:
1. Be a typo
2. Be an abbreviation - which violates "Choose expressive variable, function and class names" directive.

You can always try it out and simply remove typos CLI in the future if you do find it a nuisance?

wertzuz: > Not a typo, this comes straight from the spec.

Ideally, typos in the spec should be PRed and fixed in the spec, so shouldn't be an ever growing list.

PR #8855, opened 2026-04-10
LibFileSystem: Check self-copy before mkdir in copy_directory

### LINKED ISSUE #8853: LibFileSystem: copy_directory leaves behind an empty destination when copying into a subdirectory of the source
### Summary

`copy_directory()` creates the destination directory before checking whether the destination is inside the source tree. If the copy is rejected, the call can still leave an empty directory behind.

### Operating system

Linux

### Steps to reproduce

1.Create a temp root directory, then create src/ inside it.
2.Make sure src/dst/ does not already exist.
3.Get source_stat from stat("src").
4.Call FileSystem::copy_directory("src/dst", "src", source_stat, LinkMode::Disallowed, PreserveMode::Nothing).
5.Verify the call fails with EINVAL and that src/dst/ still exists afterward.

### Expected behavior

The copy should fail cleanly without creating or leaving behind partial filesystem state.

### Actual behavior

`mkdir(destination_path, 0755)` happens before the self-copy guard, so the function can return `EINVAL` after already creating the destination directory.

https://gist.github.com/Darshanx256/abdd92389806ef3ae62d6c87779383a6

```html
<!--I have provided gist link to C++ reduced test case, This issue cannot be reproduced with standard HTML + JS -->
```

```shell
Not Applicable.
```

_No response_

_No response_

### Contribute a patch?

Darshanx256:
Fixes #8853
when `copy_directory()` is initiated, it creates a `destination` directory

first, before checking whether the `destination` directory is within the

`source tree` itself.

On copy failure with EINVAL, an empty directory is still left behind.

The fix involves moving `source tree` check logic before any new directory is made.

I verified this behaviour locally using a small repro test.

### DIFF
--- a/Libraries/LibFileSystem/FileSystem.cpp
+++ b/Libraries/LibFileSystem/FileSystem.cpp
@@ -215,16 +215,19 @@ ErrorOr<void> copy_file(StringView destination_path, StringView source_path, str
 
 ErrorOr<void> copy_directory(StringView destination_path, StringView source_path, struct stat const& source_stat, LinkMode link, PreserveMode preserve_mode)
 {
-    TRY(Core::System::mkdir(destination_path, 0755));
-
     auto source_rp = TRY(real_path(source_path));
     source_rp = ByteString::formatted("{}/", source_rp);
 
-    auto destination_rp = TRY(real_path(destination_path));
+    LexicalPath destination_lexical_path(destination_path);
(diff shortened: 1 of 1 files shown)


Darshanx256:
Drafting this PR for now, as it needs more work on the logic.
The current logic assumes that destination is given, 
there might be cases where it would be not provided.

PR #8858, opened 2026-04-10
LibWeb: Left-align the video progress bar

AtkinsSJ:
Prevents ancestors of the `<video>` element from moving the bar around. For example, `<center><video></center>` would cause the progress bar to start in the center and expand outwards.

---

Fixes the video on https://apod.nasa.gov/apod/ap260409.html

Before:
image

After:
Screenshot_20260410_122801

Not 100% sure if this is the right thing to do, so feel free to correct me ruthlessly!

### DIFF
--- a/Libraries/LibWeb/HTML/MediaControls.css
+++ b/Libraries/LibWeb/HTML/MediaControls.css
@@ -55,6 +55,7 @@
     background-clip: content-box;
     cursor: pointer;
     flex-shrink: 0;
+    text-align: left;
 }
 
 .timeline-fill {


Gingeh:
It's worth testing if the alignment should have ever applied to the progress bar (i.e. check if an equivalent node tree would be aligned like that in another browser)

gmta:
🤔 instead of making that style explicit, shouldn't we prevent any style from being inherited by the media controls?

AtkinsSJ:
> 🤔 instead of making that style explicit, shouldn't we prevent any style from being inherited by the media controls?

I'm honestly not sure if this style should get through the shadow boundary in the first place, or if that's a bug.

Edit: Not a bug, that property is supposed to inherit apparently. https://htmlwithsuperpowers.netlify.app/styling/inheritable.html

Zaggy1024:
This seems like it should be fine, but I have a mockup page for the media controls locally, and when I make a parent of the controls have `text-align: center`, it doesn't actually affect the timeline fill positioning. I'm not sure what the difference is here, but I want to see if I can figure out why it's unaffected in case it's something I ought to be doing in our implementation.

AtkinsSJ:
> This seems like it should be fine, but I have a mockup page for the media controls locally, and when I make a parent of the controls have `text-align: center`, it doesn't actually affect the timeline fill positioning. I'm not sure what the difference is here, but I want to see if I can figure out why it's unaffected in case it's something I ought to be doing in our implementation.

The `<center>` tag uses `-libweb-center`, maybe that's part of it. I know that behaves differently.

Zaggy1024:
Ah, I wasn't aware that there was a difference there. I assume this is what's specified by the spec phrase [align descendants to the center](https://html.spec.whatwg.org/multipage/rendering.html#flow-content-3:center)? Makes sense then.

Zaggy1024 (approved):

PR #8859, opened 2026-04-10
LibWebView: Add bookmark import/export in Netscape HTML format

mikiubo:


trflynn89:
I haven't looked at the code, but my mental plan for this was to add this to `about:bookmarks` and implement import/export there (seemed like it would be easier to deal with HTML that way in JS, but I haven't dived into it)

jdahlin:
+1 on doing this in HTML/JS, which means less platform specific code to maintain, less code overall and safer implementation language.

mikiubo:
> I haven't looked at the code, but my mental plan for this was to add this to `about:bookmarks` and implement import/export there (seemed like it would be easier to deal with HTML that way in JS, but I haven't dived into it)

My first idea was to use the HTML parser, but it seemed too complicated.
I started working on it before your implementation of the about:bookmarks page.
Your idea of using JavaScript to handle it is very clever.

Let me know if I can help with anything related to bookmarks

mikiubo:
closed because of this pr 8938

Issue #8670, opened 2026-03-29
Opening https://cyan.com/games/obduction/ crashes Ladybird


Psychpsyo:
### Summary

Opening https://cyan.com/games/obduction/ in ladybird crashes the browser after a second or two as it loads stuff.

### Operating system

Linux

### Steps to reproduce

1. Open https://cyan.com/games/obduction/
2. "Ladybird flew off-course"

### Expected behavior

The tab should not crash.

### Actual behavior

The tab crashes.

### URL for a reduced test case

https://cyan.com/games/obduction/

### HTML/SVG/etc. source for a reduced test case

```html
N/A
```

### Log output and (if possible) backtrace

```shell
CRASH: Received signal SIGSEGV (11)
Stack trace (most recent call first):
#0  0x00007f3776a2903f at /lib64/libc.so.6
#1  (inlined)          in AK::RefPtr<Web::Painting::AccumulatedVisualContextTree>::as_ptr() const at /home/cameron/ladybird/AK/RefPtr.h:275:16
#2  (inlined)          in AK::RefPtr<Web::Painting::AccumulatedVisualContextTree>::ptr() const at /home/cameron/ladybird/AK/RefPtr.h:240:49
#3  (inlined)          in Web::Painting::ViewportPaintable::visual_context_tree() const at /home/cameron/ladybird/Libraries/LibWeb/Painting/ViewportPaintable.h:47:103
#4  0x00007f377a088ef3 in Web::Painting::PaintableBox::transform_rect_to_viewport(Gfx::Rect<Web::CSSPixels> const&) const at /home/cameron/ladybird/Libraries/LibWeb/Painting/PaintableBox.cpp:911:83
#5  (inlined)          in compute_intersection at /home/cameron/ladybird/Libraries/LibWeb/DOM/Document.cpp:5223:71
#6  0x00007f3779965a9b in Web::DOM::Document::run_the_update_intersection_observations_steps(double) at /home/cameron/ladybird/Libraries/LibWeb/DOM/Document.cpp:5304:57
#7  0x00007f3779bf2d6d in Web::HTML::EventLoop::update_the_rendering() at /home/cameron/ladybird/Libraries/LibWeb/HTML/EventLoop/EventLoop.cpp:513:65
#8  (inlined)          in AK::Function<void ()>::operator()() const at /home/cameron/ladybird/AK/Function.h:147:29
[...]

PR #8860, opened 2026-04-10
LibWeb: Clear stale layout state for inactive documents

awesomekling:
IntersectionObserver can keep elements from a navigated iframe's old document alive until a later rendering update. Once that document tears down its layout tree, descendant nodes and pseudo-elements can still retain stale layout and paintable pointers, and destruction can bypass the usual inactive-document teardown entirely.

Clear per-node layout and paintable pointers across the inactive document subtree before tearing down the layout tree, and do the same from destroy() for documents that never go through did_stop_being_active_document_in_navigable().

Add a crash test that observes an iframe target, navigates the iframe, and waits for rendering updates without touching stale layout state.

### DIFF
--- a/Libraries/LibWeb/DOM/Document.cpp
+++ b/Libraries/LibWeb/DOM/Document.cpp
@@ -1171,6 +1171,16 @@ void Document::tear_down_layout_tree()
     m_needs_full_layout_tree_update = true;
 }
 
+void Document::clear_layout_and_paintable_nodes_for_inactive_document()
+{
+    for_each_in_inclusive_subtree([&](auto& node) {
+        node.clear_layout_node_and_paintable({});
+        if (auto* element = as_if<Element>(node))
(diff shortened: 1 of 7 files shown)


tcl3:
This looks like it should fix #8670

Tests/LibWeb/Text/input/IntersectionObserver/implicit-root-target-in-navigated-iframe-should-not-crash.html
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE html>
+<iframe></iframe>
+<script src="../include.js"></script>
+<script>
+    asyncTest(async done => {
```
tcl3: This could be turned into a crash test using the "test-wait" class. If not you could turn this into a `promiseTest()` to avoid needing to call `done()`.

Ari4ka (approved):

PR #8861, opened 2026-04-10
LibJS: Enable the asmint interpreter on Windows x86_64 

kalenikaliaksandr:
Written entirely by Claude Code.

### DIFF
--- a/Libraries/LibJS/AsmIntGen/src/codegen_x86_64.rs
+++ b/Libraries/LibJS/AsmIntGen/src/codegen_x86_64.rs
@@ -62,17 +62,45 @@ pub fn generate(program: &Program) -> String {
     w!(out, "#endif");
     w!(out);
 
+    // @feat.00 is a magic COFF absolute symbol whose value is a bitfield of
+    // compiler-level protection features the object opts into. clang-cl
+    // bakes this into every C/C++ object under /guard:cf with value 0x800
+    // (= CFG compatible); without it, link.exe treats our object as CFG-
+    // unknown and silently reduces the /GUARD:CF coverage of the final
+    // image. We hand-emit the same block clang itself produces for C
+    // sources compiled with -Xclang -cfguard so asmint_x86_64.obj lines up
+    // with the rest of LibJS's object files.
+    //
+    // Setting 0x800 is an opt-in *declaration*, not a runtime enforcement:
+    // it does not inject __guard_check_icall_fptr calls into our handlers.
+    // The only indirect branch in the blob is `jmp [r12 + rax*8]`, and x64
+    // CFG validates indirect calls, not indirect jumps, so dispatch is
+    // unaffected at runtime. asm_interpreter_entry is only ever reached
+    // via a direct `call` from AsmInterpreter::run, so the call site is
+    // likewise unaffected.
+    if matches!(program.object_format, ObjectFormat::Coff) {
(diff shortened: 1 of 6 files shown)


Libraries/LibJS/AsmIntGen/src/codegen_x86_64.rs
```diff
@@ -29,17 +29,78 @@ pub fn generate(program: &Program) -> String {
...
+    // unaffected at runtime. asm_interpreter_entry is only ever reached
+    // via a direct `call` from AsmInterpreter::run, so the call site is
+    // likewise unaffected.
+    if matches!(program.object_format, ObjectFormat::Coff) {
+        w!(out, ".def    @feat.00;");
+        w!(out, ".scl    3;");
+        w!(out, ".type   0;");
+        w!(out, ".endef");
+        w!(out, ".globl  @feat.00");
+        w!(out, "@feat.00 = 0x800");
```
R-Goc, line 87:
```suggestion
        w!(out, "@feat.00 = 0x80010990");
```
Copied over from what current msvc sets with /guard:cf

Libraries/LibJS/AsmIntGen/src/codegen_x86_64.rs
```diff
@@ -83,42 +144,64 @@ pub fn generate(program: &Program) -> String {
...
     w!(out, "CSYM(asm_interpreter_entry):");
-    w!(out, "    .cfi_startproc");
+    match fmt {
+        ObjectFormat::Elf | ObjectFormat::MachO => {
+            w!(out, "    .cfi_startproc");
+        }
+        ObjectFormat::Coff => {
+            w!(out, "    .seh_proc CSYM(asm_interpreter_entry)");
+        }
+    }
```
R-Goc, line 168:
```suggestion
    }
    w!(out, "    endbr64");
```
Not fully sure we need it here, compilers default to adding it everywhere. Technically it is only needed for indirect branch targets (vtables, function pointers, etc.)

Libraries/LibJS/AsmIntGen/src/codegen_x86_64.rs
R-Goc: Also we can just throw this in emit_file_trailer

R-Goc:
generate_handler and generate_fallback_handler also need an endbr64 after the .seh_startproc. Not sure about generate_entry_point but we can do it to be safe if someone ever made the interpreter entry called by function pointer. 
```
diff --git a/Meta/CMake/compile_options.cmake b/Meta/CMake/compile_options.cmake
index e38f1e924f..f7976ff900 100644
--- a/Meta/CMake/compile_options.cmake
+++ b/Meta/CMake/compile_options.cmake
@@ -178,6 +178,8 @@ elseif (MSVC)
     add_cxx_link_options(/STACK:0x800000)
     # enable control flow protection
     add_cxx_link_options(/GUARD:CF)
+    add_cxx_link_options(/CETCOMPAT)
+    add_cxx_link_options(/CETCOMPATSTRICT)
     add_cxx_link_option_if_supported(-prefetch-inputs)

     # Use ghash compressed debug info for build time and pdb size reduction
```
Diff to enable it with the linker.

R-Goc:
Also that outln taking a reference to a static member issue. Not sure how that works on linux tbh.

PR #8864, opened 2026-04-10
UI/Qt: Open new tab page when New Window is created with no URLs

CalebC48:
## Problem

Clicking the **Hamburger/Menu icon**, and then clicking **New Window** calls `Application::new_window` with an empty URL list. `BrowserWindow` only creates tabs by iterating `initial_urls`, so with no URLs and no parent tab it added zero tabs. The window appeared empty (no real browser content).

Second-instance launches already behave correctly because `BrowserProcess` uses `sanitize_urls()`, which appends the new tab URL when the list is empty.

image

_When I had orignally been testing this issue it did not even have the plus icon. Having that does make it better, but I still think this is a worthwhile improvement_
## Solution

When there is no parent tab and the URL list is empty, append `settings().new_tab_page_url()` before constructing `BrowserWindow`, matching the `sanitize_urls()` behavior.

image

### DIFF
--- a/UI/Qt/BrowserWindow.cpp
+++ b/UI/Qt/BrowserWindow.cpp
@@ -325,7 +325,7 @@ BrowserWindow::BrowserWindow(Vector<URL::URL> const& initial_urls, IsPopupWindow
         tab.focus_location_editor();
     });
     QObject::connect(m_new_window_action, &QAction::triggered, this, [] {
-        (void)Application::the().new_window({});
+        (void)Application::the().new_window({ WebView::Application::settings().new_tab_page_url() });
     });
     QObject::connect(open_file_action, &QAction::triggered, this, &BrowserWindow::open_file);
 


Zaggy1024:
Feels like the two commits could be combined into one, and then the vector copy becomes unnecessary since the size is checked before appending.

CalebC48:
> Feels like the two commits could be combined into one, and then the vector copy becomes unnecessary since the size is checked before appending.

@Zaggy1024 Thanks for the review!

I had thought about that, but `initial_urls` is passed as `const&`, so I can’t append to it directly. That’s why I introduced the mutable `urls` copy.

Would you prefer I restructure it so the copy is only created in the fallback path (i.e. inside `if (!parent_tab && initial_urls.is_empty())` ), and otherwise pass `initial_urls` directly? I can update it that way.

I can also squash the two commits into one.

Zaggy1024:
Ah, I missed that. I think it may make sense for it to be the caller's responsibility to ensure that the vector is non-empty (like in LibWebView's sanitize_urls). From my reading, there are only two places that pass no URLs, which have easy access to the new tab page.

A forced copy is a small price to pay though, so I'm not opposed to merging this as-is either. If you think that's best, I wouldn't say a squash is needed.

trflynn89:
I was going to make the same suggestion. The following patch on top of the `master` branch fixes this issue for me:
```diff
diff --git a/UI/Qt/BrowserWindow.cpp b/UI/Qt/BrowserWindow.cpp
index e3ec838edd..a1f4ee52b4 100644
--- a/UI/Qt/BrowserWindow.cpp
+++ b/UI/Qt/BrowserWindow.cpp
@@ -325,7 +325,7 @@ BrowserWindow::BrowserWindow(Vector<URL::URL> const& initial_urls, IsPopupWindow
         tab.focus_location_editor();
     });
     QObject::connect(m_new_window_action, &QAction::triggered, this, [] {
-        (void)Application::the().new_window({});
+        (void)Application::the().new_window({ WebView::Application::settings().new_tab_page_url() });
     });
     QObject::connect(open_file_action, &QAction::triggered, this, &BrowserWindow::open_file);
```

CalebC48:
Thank you @Zaggy1024 and @trflynn89, that makes sense. I’ll switch to caller-side responsibility and pass `new_tab_page_url()` from the New Window action. 

The other `new_window({})` callsite is also in `UI/Qt/BrowserWindow.cpp` inside the `hints.popup` path (currently line 485). From what I can tell that path uses `parent_tab`/`page_index` to create a child popup tab, so it doesn’t seem to rely on `initial_urls` in the same way.

If you’d still prefer, I can also pass `new_tab_page_url()` there for consistency. But otherwise, everything here still works for me!

trflynn89:
Right, we don't need to pass a URL for popup windows. The caller will navigate to the desired URL itself.

trflynn89 (approved):

PR #8866, opened 2026-04-10
LibWasm: It's JIT time

alimpfard:
benchmarks (m2, linux):
```
Suite           Test                      Speedup  Old (Mean ± Range)           New (Mean ± Range)
--------------  ----------------------  ---------  ---------------------------  ------------------------------
WasmCoremark    coremark-minimal.wasm       8.153  941.247 ± 938.887 … 942.911  7674.419 ± 7672.456 … 7677.275
WasmMicroBench  call-00-args.wasm           3.196  3.357 ± 3.342 … 3.379        1.050 ± 1.041 … 1.060
WasmMicroBench  call-01-args.wasm           3.687  4.609 ± 4.568 … 4.651        1.250 ± 1.245 … 1.259
WasmMicroBench  call-02-args.wasm           3.907  4.910 ± 4.898 … 4.925        1.257 ± 1.248 … 1.271
WasmMicroBench  call-03-args.wasm           3.822  5.069 ± 5.042 … 5.095        1.326 ± 1.291 … 1.389
WasmMicroBench  call-04-args.wasm           1.951  4.679 ± 4.655 … 4.705        2.398 ± 2.383 … 2.406
WasmMicroBench  call-16-args.wasm           1.946  6.799 ± 6.760 … 6.845        3.494 ± 3.481 … 3.507
WasmMicroBench  call-32-args.wasm           1.864  9.242 ± 9.148 … 9.397        4.959 ± 4.950 … 4.968
WasmRustBench   base64-bench.wasm           9.728  13.658 ± 13.595 … 13.755     1.404 ± 1.398 … 1.412
WasmRustBench   regex-match-bench.wasm      0.83   2.619 ± 2.616 … 2.623        3.156 ± 2.986 … 3.292
WasmRustBench   sha512-bench.wasm          30.857  31.529 ± 31.412 … 31.706     1.022 ± 1.009 … 1.045
WasmCoremark    Total                       8.153  941.247                      7674.419
WasmMicroBench  Total                       2.457  38.666                       15.734
WasmRustBench   Total                       8.564  47.806                       5.582
All Suites      Total                       2.676  99.362                       37.127
```

benchmarks (x86, linux):
```
Suite           Test                    Speedup    Old (Mean ± Range)              New (Mean ± Range)
--------------  ----------------------  ---------  ------------------------------  ---------------------------------
WasmCoremark    coremark-minimal.wasm   12.097     1169.599 ± 1156.470 … 1178.342  14148.562 ± 13817.880 … 14346.173
WasmRustBench   base64-bench.wasm       12.458     11.389 ± 11.292 … 11.442        0.914 ± 0.908 … 0.920
WasmRustBench   regex-match-bench.wasm  1.231      2.003 ± 1.967 … 2.025           1.628 ± 1.614 … 1.638
WasmRustBench   sha512-bench.wasm       33.438     26.445 ± 26.250 … 26.684        0.791 ± 0.788 … 0.796
WasmMicroBench  call-00-args.wasm       4.394      3.284 ± 3.258 … 3.323           0.747 ± 0.722 … 0.774
WasmMicroBench  call-01-args.wasm       4.688      4.223 ± 4.153 … 4.315           0.901 ± 0.895 … 0.908
WasmMicroBench  call-02-args.wasm       4.791      4.314 ± 4.285 … 4.367           0.900 ± 0.896 … 0.903
WasmMicroBench  call-03-args.wasm       4.784      4.316 ± 4.278 … 4.343           0.902 ± 0.898 … 0.911
WasmMicroBench  call-04-args.wasm       2.227      4.365 ± 4.242 … 4.553           1.960 ± 1.925 … 2.014
WasmMicroBench  call-16-args.wasm       1.893      5.414 ± 5.371 … 5.500           2.860 ± 2.838 … 2.880
WasmMicroBench  call-32-args.wasm       1.765      7.122 ± 6.870 … 7.266           4.035 ± 3.912 … 4.236
WasmCoremark    Total                   12.097     1169.599                        14148.562
WasmRustBench   Total                   11.953     39.838                          3.333
WasmMicroBench  Total                   2.685      33.037                          12.306
All Suites      Total                   2.642      99.471                          37.657
```

benchmarks (m3, macos):
```
Suite           Test                      Speedup  Old (Mean ± Range)              New (Mean ± Range)
--------------  ----------------------  ---------  ------------------------------  ------------------------------
WasmCoremark    coremark-minimal.wasm       8.074  1145.881 ± 1141.487 … 1150.616  9251.550 ± 9202.711 … 9329.940
WasmRustBench   base64-bench.wasm          10.029  11.505 ± 11.446 … 11.602        1.147 ± 1.134 … 1.159
WasmRustBench   regex-match-bench.wasm      0.968  2.169 ± 2.163 … 2.175           2.241 ± 2.182 … 2.287
WasmRustBench   sha512-bench.wasm          31.271  26.556 ± 24.977 … 27.505        0.849 ± 0.848 … 0.851
WasmMicroBench  call-00-args.wasm           2.943  2.534 ± 2.513 … 2.550           0.861 ± 0.860 … 0.862
WasmMicroBench  call-01-args.wasm           3.618  3.715 ± 3.704 … 3.734           1.027 ± 1.026 … 1.027
WasmMicroBench  call-02-args.wasm           3.834  3.909 ± 3.885 … 3.953           1.019 ± 1.018 … 1.020
WasmMicroBench  call-03-args.wasm           4.054  4.247 ± 4.234 … 4.270           1.047 ± 1.045 … 1.051
WasmMicroBench  call-04-args.wasm           1.966  3.779 ± 3.732 … 3.825           1.923 ± 1.915 … 1.936
WasmMicroBench  call-16-args.wasm           1.933  5.378 ± 5.341 … 5.399           2.782 ± 2.739 … 2.812
WasmMicroBench  call-32-args.wasm           1.941  7.600 ± 7.567 … 7.636           3.916 ± 3.915 … 3.918
WasmCoremark    Total                       8.074  1145.881                        9251.550
WasmRustBench   Total                       9.494  40.229                          4.237
WasmMicroBench  Total                       2.478  31.162                          12.576
All Suites      Total                       3.335  100.371                         30.092
```


Libraries/LibWasm/AbstractMachine/BytecodeInterpreter.cpp
```diff
@@ -6663,36 +6828,6 @@ CompiledInstructions try_compile_instructions(Expression const& expression, Span
...
-            if (local_index <= 7) {
-                result.extra_instruction_storage.unchecked_append(Instruction(
-                    static_cast<OpCode>(Instructions::synthetic_local_set_0.value() + local_index),
-                    dispatch.instruction->local_index()));
-                result.dispatches[i].instruction = &result.extra_instruction_storage.unsafe_last();
-                result.dispatches[i].instruction_opcode = result.dispatches[i].instruction->opcode();
-            }
-        }
-    }
-
```
Hendiadyoin1, line 6695: Does loosing this have an effect on the normal interpreter pref, or was this replaced

Libraries/LibWasm/Rust/src/compiler.rs
```diff
@@ -0,0 +1,2029 @@
...
+use cranelift_codegen::ir::types;
+use cranelift_codegen::ir::{AbiParam, Function, InstBuilder, MemFlags, Signature, StackSlotData, StackSlotKind, UserFuncName};
+use cranelift_codegen::isa::CallConv;
+use cranelift_codegen::settings::{self, Configurable};
+use cranelift_codegen::{self, Context};
+use cranelift_frontend::{FunctionBuilder, FunctionBuilderContext, Variable};
+use cranelift_native;
+
+// Wasm opcode constants (must match Opcode.h).
+mod op {
```
Hendiadyoin1, line 23: Wasnt there a bindgen thing to include c/c++ header defines and enums

Libraries/LibWasm/Rust/src/compiler.rs
```diff
@@ -0,0 +1,2029 @@
...
+
+        Ok(code.code_buffer().to_vec())
+    }
+
+    fn is_supported(insn: &CraneliftInsn) -> bool {
+        let opc = insn.opcode;
+        matches!(
+            opc,
+            op::NOP
+                | op::UNREACHABLE
```
Hendiadyoin1, line 2032: Side thing:
This indentation is kinda ugly...

Libraries/LibWasm/Opcode.h
```diff
@@ -545,7 +545,8 @@ namespace Instructions {
     M(synthetic_i64_shl2local, 0xfe00000000000038ull, 0, 1)      \
     M(synthetic_i64_shru2local, 0xfe00000000000039ull, 0, 1)     \
     M(synthetic_i64_shrs2local, 0xfe0000000000003aull, 0, 1)     \
-    M(synthetic_local_seti64_const, 0xfe0000000000003bull, 0, 0)
+    M(synthetic_local_seti64_const, 0xfe0000000000003bull, 0, 0) \
+    M(synthetic_br_table_cont, 0xfe0000000000003cull, 0, 0)
```
Hendiadyoin1, line 551: Where does this new op come from now?

Libraries/LibWasm/Rust/src/compiler.rs
```diff
@@ -0,0 +1,2029 @@
...
+        flag_builder.set("opt_level", "speed").unwrap();
+        flag_builder.set("is_pic", "false").unwrap();
+        let flags = settings::Flags::new(flag_builder);
+        let isa = cranelift_native::builder()
+            .map_err(|_| "unsupported host architecture")?
+            .finish(flags)
+            .map_err(|_| "failed to build ISA")?;
+
+        // Function signature matches handler_ptr:
+        //   u64 fn(void* interpreter, void* configuration, void* insn, u32 short_ip, void* cc, void* addrs)
```
Hendiadyoin1, line 91: Can we assert that this does not diverge somehow?

Hendiadyoin1 (commented):
So from flying over the code its half a re implementation of some common ops, with some boiler plate and calling out to c++ in some cases if needed?

Still found some things that stuck out, and in general it would be ideal to get the definitions from a common source, like a c header

Libraries/LibWasm/AbstractMachine/BytecodeInterpreter.cpp
alimpfard: Ah that's leftover from a debug session
It's not _that_ big a gain but it's nice to have, will restore.

Libraries/LibWasm/Rust/src/compiler.rs
alimpfard: I couldn't get it to properly handle the ops, but I didn't really try particularly hard...

Libraries/LibWasm/Rust/src/compiler.rs
alimpfard: 🤷 cargo fmt did it

Libraries/LibWasm/Opcode.h
alimpfard: It's not interpreted, the bridge generates it to encode br_table

Libraries/LibWasm/Rust/src/compiler.rs
alimpfard: I'd be happy to do so if it's possible at all

Libraries/LibWasm/Opcode.h
Hendiadyoin1: ahh, maybe add a comment then

Libraries/LibWasm/AbstractMachine/AbstractMachine.cpp
```diff
@@ -12,9 +12,165 @@
 #include <LibWasm/AbstractMachine/Interpreter.h>
 #include <LibWasm/AbstractMachine/Validator.h>
 #include <LibWasm/Types.h>
+#include <sys/mman.h>
+#include <unistd.h>
```
R-Goc: Saw both this and mmap.h in multiple places. Use System::Core/anonymous buffer apis, if missing add to System for Linux and windows. I did want to rework anonymous buffer so I can help with that, it very clearly needs a permission system. Should be very easy to do. In current state it doesn't compile on windows at all because of unistd.h includes.

R-Goc (changes requested):

Libraries/LibWasm/AbstractMachine/BytecodeInterpreter.cpp
```diff
@@ -26,9 +26,130 @@
...
+    Wasm::BytecodeInterpreter* interpreter { nullptr };
+    Wasm::Configuration* configuration { nullptr };
+    CompiledFaultRecoveryContext* previous { nullptr };
+    sigjmp_buf jump_buffer;
+    bool faulted { false };
+};
+
+thread_local CompiledFaultRecoveryContext* s_compiled_fault_recovery = nullptr;
+thread_local void const* volatile s_last_compiled_entry = nullptr;
+static struct sigaction s_old_sigsegv;
```
R-Goc: None of this signal code will work on windows.

Libraries/LibWasm/AbstractMachine/Configuration.h
```diff
@@ -257,20 +363,23 @@ class Configuration {
         Value(0),
     };
 
-private:
+    // Public for CraneliftBridge direct call pop_frame.
```
R-Goc, line 366: Can we use a friend here?

Libraries/LibWasm/CraneliftBridge.cpp
```diff
@@ -0,0 +1,1135 @@
...
+        }
+    }
+
+    return out;
+}
+
+static void try_cranelift_compile_batch(Vector<BatchInput>& batch)
+{
+#if defined(AK_OS_WINDOWS)
+    return;
```
R-Goc: Why here? Also cranelift does support windows from what I see.

Libraries/LibWasm/Rust/src/compiler.rs
ADKaster: add em to
```toml
[export]
include = ["MyType"]
```
list.

make sure they are pub.

alimpfard: @ADKaster the definitions are macros though, unless we can get bindgen to export static constexpr variables?

alimpfard: I just generated the stuff from build.rs :shrug:

rustfmt.toml
```diff
@@ -0,0 +1 @@
+max_width = 200
```
trflynn89: We use 120 for everything else fwiw

alimpfard: Not really?
```
 10-19  |  35336 ====================================
 20-29  |  54716 =======================================================
 30-39  |  59616 ============================================================
 40-49  |  53968 ======================================================
 50-59  |  42655 ===========================================
 60-69  |  30449 ===============================
 70-79  |  23448 ========================
 80-89  |  16730 =================
 90-99  |  12211 ============
100-109 |   8993 =========
110-119 |   6271 ======
120-129 |   4765 =====
130-139 |   3712 ====
140-149 |   2454 ==
150-159 |   1830 ==
160+    |   4835 =====
```

a little zoomed in:
```
100-104 |  5022 ============================================================
105-109 |  3971 ===============================================
110-114 |  3496 ==========================================
115-119 |  2775 =================================
120-124 |  2590 ===============================
125-129 |  2175 ==========================
130-134 |  2128 =========================
135-139 |  1584 ===================
140-144 |  1345 ================
145-149 |  1109 =============
150-154 |   964 ============
155-159 |   866 ==========
160-164 |   613 =======
165-169 |   598 =======
170-174 |   578 =======
175-179 |   418 =====
180-184 |   302 ====
185-189 |   321 ====
190-194 |   242 ===
195-199 |   210 ===
```

gmta: We _intend_ to use 120 for everything else fwiw 😅 that's definitely the target column for wrapping as mentioned in `CodingStyle.md`.

alimpfard: 120 in rust is exceptionally ugly though...
I mean sure, whatever.

trflynn89: Yeah we haven't set it in clang-format, but on top of the codingstyle.md note, 120 is the hard limit for JS and python

alimpfard: I have no particular feelings about the limit, but imo 120 is too short for rust at least; everything looks so cramped with it.
something we can deal with later though.

R-Goc:
Could the tests be moved to Tests/LibWasm instead of Libraries/LibWasm/Tests? The JS tests were just recently moved from Libraries/LibJS/Tests to Tests/LibJS. That way everything will be consistent.

alimpfard:
> Could the tests be moved to Tests/LibWasm

I don't mind doing that, but will happen in a different PR :)

Libraries/LibJS/Runtime/ArrayBuffer.h
```diff
@@ -49,24 +49,69 @@ struct DataBlock {
         size_t size = 0;
     };
 
+    // AD-HOC: ECMA-262 models ArrayBuffer backing storage as a Data Block. We additionally allow
+    //         host code to provide an external byte store via callbacks so engine-independent
+    //         consumers like LibWeb can project spec-defined host objects onto ArrayBuffer without
+    //         teaching LibJS about those hosts.
+    struct UnownedExternalBuffer {
```
ADKaster, line 56: Is "Unowned" the model we really want here? With the goal of "more safety if possible", requiring separate ownership of the storage for the js object here feels fragile.

Adding a destroy callback/method/thing wouldn't be much more work on the side of this opaque data wrapper class

Libraries/LibJS/Bytecode/Interpreter.cpp
```diff
@@ -892,7 +892,7 @@ inline Value fast_typed_array_get_element(TypedArrayBase& typed_array, u32 index
     }
 
     auto const& array_buffer = *typed_array.viewed_array_buffer();
-    auto const* slot = reinterpret_cast<T const*>(array_buffer.buffer().offset_pointer(offset_into_array_buffer.value()));
+    auto const* slot = reinterpret_cast<T const*>(array_buffer.data() + offset_into_array_buffer.value());
```
ADKaster: is ptr arithmetic the model we want in both AsmInterpreter and here? We have ReadonlyBytes/Span for a reason.

Libraries/LibJS/Runtime/ArrayBuffer.h
```diff
@@ -49,24 +49,69 @@ struct DataBlock {
...
+        explicit UnownedExternalBuffer(void* context, DataFunction data, SizeFunction size)
+            : context(context)
+            , data(data)
+            , size(size)
+        {
+        }
+
+        void* context { nullptr };
+        DataFunction data { nullptr };
+        SizeFunction size { nullptr };
```
ADKaster: ReadonlySpanFunction ?

Libraries/LibWasm/AbstractMachine/AbstractMachine.cpp
```diff
@@ -12,9 +12,165 @@
...
+    if (auto max = m_type.limits().max(); max.has_value()) {
+        if (max.value() * Constants::page_size < new_size)
+            return false;
+    }
+
+    auto previous_size = m_data.size();
+    if (m_data.try_resize(new_size).is_error())
+        return false;
+    if (!m_data.is_virtual())
+        __builtin_memset(m_data.offset_pointer(previous_size), 0, size_to_grow);
```
ADKaster: calling builtin memset like this here feels strange. surely just memset? or an overwrite call on `span().slice( ... )` ?

Libraries/LibWeb/WebAssembly/Memory.cpp
```diff
@@ -279,9 +289,9 @@ GC::Ref<JS::ArrayBuffer> Memory::create_a_fixed_length_memory_buffer(JS::VM& vm,
         // 1. Let block be a Shared Data Block which is identified with the underlying memory of memaddr.
         // 2. Let buffer be a new SharedArrayBuffer with the internal slots [[ArrayBufferData]] and [[ArrayBufferByteLength]].
         // 3. Set buffer.[[ArrayBufferData]] to block.
-        array_buffer = JS::ArrayBuffer::create(realm, &memory->data(), JS::DataBlock::Shared::Yes);
-
         // 4. Set buffer.[[ArrayBufferByteLength]] to the length of block.
+        // NOTE: ArrayBufferByteLength should contain the original size regardless of growth.
+        array_buffer = JS::ArrayBuffer::create(realm, JS::DataBlock::UnownedExternalBuffer(&memory->data(), wasm_memory_buffer_data, memory->size()), JS::DataBlock::Shared::Yes);
```
ADKaster: is there no way to attach the lifetime of this ArrayBuffer to that of the Wasm memory?

ADKaster: What if the array buffer is transferred? Is that impossible for all cases?

Libraries/LibWasm/Tests/Fixtures/Modules/zero-local-cross-module-memory-caller.wat
```diff
@@ -0,0 +1,14 @@
+(module
+  (import "callee" "touch" (func $touch))
+
+  (memory (export "memory") 1)
+  (data (i32.const 0) "\2a")
+
+  ;; After calling into the imported wasm function, reads must still use this
+  ;; module's default memory, not the callee's memory.
+  (func (export "call_and_read") (result i32)
```
ADKaster, line 9: open issue w/spec tests for this?

Libraries/LibWasm/AbstractMachine/AbstractMachine.cpp
```diff
@@ -77,16 +76,12 @@ void MemoryBuffer::try_reserve_wasm32_address_space()
     auto mapping_size = reserved_capacity * 2;
     auto reservation_size = mapping_size + 2 * host_page_size;
 
-    int flags = MAP_PRIVATE | MAP_ANONYMOUS;
-#ifdef MAP_NORESERVE
-    flags |= MAP_NORESERVE;
-#endif
-    auto* mapping = mmap(nullptr, reservation_size, PROT_NONE, flags, -1, 0);
```
ADKaster: This change makes it feel like the LibCore additions should come first, before the initial MemoryBuffer commit.

Libraries/LibWasm/AbstractMachine/BytecodeInterpreter.cpp
```diff
@@ -26,9 +26,176 @@
...
+    if (previous_action.sa_handler != SIG_DFL) {
+        previous_action.sa_handler(signal);
+        __builtin_trap();
+    }
+
+    struct sigaction default_action {};
+    default_action.sa_handler = SIG_DFL;
+    sigemptyset(&default_action.sa_mask);
+    sigaction(signal, &default_action, nullptr);
+    raise(signal);
```
ADKaster: what's the goal with this code? doing funny business in signal handlers is a good way to introduce UB.

why are we re-raising the signal?

Libraries/LibWasm/AbstractMachine/BytecodeInterpreter.cpp
```diff
@@ -26,9 +26,176 @@
...
+    _exit(128 + signal);
+}
+
+static void compiled_fault_signal_handler(int signal, siginfo_t* info, void* context)
+{
+    if (auto* recovery = s_compiled_fault_recovery; recovery && info && is_wasm_memory_fault(*recovery->configuration, info->si_addr)) {
+        recovery->faulted = true;
+        longjmp(recovery->jump_buffer, 1);
```
ADKaster: this is a really bad idea.

https://cmu-sei.github.io/secure-coding-standards/sei-cert-c-coding-standard/rules/signals-sig/sig30-c


what's the goal here? calling write() is async-signal-safe, but printf is not. and calling longjump definitely is not.

https://cmu-sei.github.io/secure-coding-standards/sei-cert-c-coding-standard/rules/signals-sig/sig30-c#noncompliant-code-example-longjmp

Libraries/LibWasm/AbstractMachine/Validator.cpp
```diff
@@ -4399,6 +4401,49 @@ ErrorOr<Validator::ExpressionTypeResult, ValidationError> Validator::validate(Ex
...
+                    if (func_idx < m_context.functions.size() && m_context.functions[func_idx].results().size() > 1) {
+                        has_unsupported_types = true;
+                        break;
+                    }
+                }
+            }
+        }
+        // Also skip multi-value return functions.
+        if (!has_unsupported_types && result_types.size() <= 1)
+            try_cranelift_compile(expression.compiled_instructions, static_cast<u32>(result_types.size()));
```
ADKaster, line 4444: eh... what happens when try_cranelift_compile returns false?

Meta/CMake/flatpak/cargo-sources.json
```diff
@@ -5,6 +5,20 @@
         "tag": "v0.5.1",
         "dest": "corrosion"
     },
+    {
+        "type": "archive",
```
ADKaster, line 9: these should be updated alongside Cargo.lock for correct atomicity

Libraries/LibJS/Runtime/ArrayBuffer.h
alimpfard: Hm, this isn't so much "requiring separate ownership" as it is "this is just a view".
Destroying this shouldn't free the underlying object anyhow:
```js
let foo = someWasmThing.buffer;
gc(foo); // can't ever drop the actual buffer
```

Though this made me realise that you could get a dangling ref if you make the original go away...will have a thonk.

Libraries/LibJS/Bytecode/Interpreter.cpp
alimpfard: yeah will switch to span.

Libraries/LibWasm/AbstractMachine/AbstractMachine.cpp
alimpfard: :thinking: do we have a .fill() on span? guess I'll add one.

Libraries/LibWeb/WebAssembly/Memory.cpp
alimpfard: lifetime: still thonking about visit_edges.
Transfer should produce a copy, no?

Libraries/LibWasm/Tests/Fixtures/Modules/zero-local-cross-module-memory-caller.wat
alimpfard: yakstack.push

Libraries/LibWasm/AbstractMachine/BytecodeInterpreter.cpp
alimpfard: it wasn't our thing to handle.
the only thing this should handle is faults within the wasm guard pages.

Libraries/LibCore/System.h
```diff
@@ -78,6 +78,9 @@ ErrorOr<struct stat> fstatat(int fd, StringView path, int flags);
 CORE_API ErrorOr<int> fcntl(int fd, int command, ...);
 ErrorOr<void*> mmap(void* address, size_t, int protection, int flags, int fd, off_t, size_t alignment = 0, StringView name = {});
 ErrorOr<void> munmap(void* address, size_t);
+CORE_API ErrorOr<void*> reserve_address_space(size_t size);
+CORE_API ErrorOr<void> commit_memory(void* address, size_t size);
+CORE_API ErrorOr<void> release_address_space(void* address, size_t size);
```
R-Goc, line 83: Just a thought, these are all things provided by Core::System::mmap. If we are doing this maybe we should write our own mmap abstraction that would be nice enough to use directly. We could then drop the mmap compat library on windows too.

Libraries/LibWasm/AbstractMachine/BytecodeInterpreter.cpp
R-Goc: Do you know how this should look like on windows? There access violations are SEH exceptions. But I have no idea what the spec says of it or how other browsers do it.

Libraries/LibWasm/AbstractMachine/BytecodeInterpreter.cpp
alimpfard: this whole thing is a mess I think, it also messes with the backtrace if the crash trace involves a compiled wasm function...
re async-signal-safe, I'll switch to sigsetjmp/siglongjmp; can't think of any other meaningful solution that wouldn't defeat the entire purpose of the opt.

Libraries/LibWasm/AbstractMachine/Validator.cpp
alimpfard: Nothing? it won't schedule compilation for anything, and it won't modify `expression.compiled_instructions`.

Libraries/LibCore/System.h
alimpfard: I really don't wanna figure out a mapping for 50 different platform-specific mmap flags, I'd actually rather have these specific functions tbh.

R-Goc: I might look into that at some point.

Libraries/LibWasm/AbstractMachine/BytecodeInterpreter.cpp
alimpfard: Windows should already be handled, it ends up in the same longjmp path (that I'm replacing with mucking about with ucontext)

Libraries/LibJS/Runtime/ArrayBuffer.h
alimpfard: That'd be very elegant but you're looking at an outdated version of that, the size can be either a function or a size_t.

Libraries/LibJS/Runtime/ArrayBuffer.h
alimpfard: Now it takes and visits a owner object.

Libraries/LibWasm/AbstractMachine/BytecodeInterpreter.cpp
alimpfard: Really bad idea fixed by ucontext surgery.

Libraries/LibWasm/AbstractMachine/BytecodeInterpreter.cpp
```diff
@@ -26,9 +26,209 @@
...
+#    else
+#        include <ucontext.h>
+#    endif
+#endif
 
 using namespace AK::SIMD;
 
+namespace {
+
+static bool is_wasm_memory_fault(Wasm::Configuration& configuration, void* address)
```
cqundefine: This function ends up being unused when `WASM_COMPILED_FAULT_RECOVERY_SUPPORTED` is 0, and triggers a warning which ends up as an error.

alimpfard: Will fix, thanks for catching that!

PR #8867, opened 2026-04-11
LibWebView: Add date_added timestamp to bookmark items

mikiubo:
Store a creation timestamp on every BookmarkItem, for both bookmarks and folders.

The JSON schema version is bumped from 1 to 2.

### DIFF
--- a/Libraries/LibWebView/BookmarkStore.cpp
+++ b/Libraries/LibWebView/BookmarkStore.cpp
@@ -26,6 +26,7 @@ static constexpr auto URL_KEY = "url"sv;
 static constexpr auto TITLE_KEY = "title"sv;
 static constexpr auto FAVICON_KEY = "favicon"sv;
 static constexpr auto CHILDREN_KEY = "children"sv;
+static constexpr auto DATE_ADDED_KEY = "dateAdded"sv;
 
 static constexpr auto TYPE_BOOKMARK = "bookmark"sv;
 static constexpr auto TYPE_FOLDER = "folder"sv;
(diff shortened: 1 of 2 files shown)


trflynn89:
What's the date for? (I would expect that to be explained in the commit message)

Regarding version bumps, I don't think we need to change it unless there is an explicit action we need to take to handle migrating existing bookmarks files to the new version. In this case, you're already implicitly setting the date field to the epoch if it is missing.

mikiubo:
> What's the date for? (I would expect that to be explained in the commit message)
> 
> Regarding version bumps, I don't think we need to change it unless there is an explicit action we need to take to handle migrating existing bookmarks files to the new version. In this case, you're already implicitly setting the date field to the epoch if it is missing.

Ok i think i fixed. I add the date so we can match the export/import Netscape structure of other browser.
If you think is not usefull or dangerous (maybe there are some privacy concern?) ignore this pr). 

Let me know

Libraries/LibWebView/BookmarkStore.cpp
```diff
@@ -26,6 +26,7 @@ static constexpr auto URL_KEY = "url"sv;
 static constexpr auto TITLE_KEY = "title"sv;
 static constexpr auto FAVICON_KEY = "favicon"sv;
 static constexpr auto CHILDREN_KEY = "children"sv;
+static constexpr auto DATE_ADDED_KEY = "date_added"sv;
```
trflynn89: Let's use camelCase for JSON

trflynn89 (commented):
Got it - that's fine, just wanted an explanation in the commit.

Should we add last modified while we're here? (at least, bookmarks exported from Firefox have that as well, not sure about other browsers)

Libraries/LibWebView/BookmarkStore.cpp
mikiubo: ok i think i fixed, if you want i can add a toggel on the manager page to show dates?

Thanks

trflynn89 (approved):

PR #8874, opened 2026-04-11
LibWeb: Add PermissionsAPI and geolocation permission

NiccoloAntonelliDziri:
This adds the basic infrastructure for the Permissions API, including the `Permissions` and `PermissionStatus` interfaces. The API is exposed via `navigator.permissions` on both `Window` and `WorkerGlobalScope`.

- `is_permission_supported` is hardcoded to return false.
- The `query()` method checks for a secure context but rejects all queries since no powerful features are supported yet.

Other commits extend on that to be able to support geolocation permission. geolocation is the permission used in permissions tests.


Libraries/LibWeb/PermissionsAPI/PermissionStore.h
```diff
@@ -0,0 +1,54 @@
...
+    // https://w3c.github.io/permissions/#dfn-remove-a-permission-store-entry
+    void remove_permission_store_entry(PermissionDescriptor const& descriptor, URL::Origin const& key);
+
+private:
+    PermissionStore() = default;
+
+    Vector<PermissionStoreEntry> m_entries;
+};
+
+// https://w3c.github.io/permissions/#dfn-is-equal-to
```
shannonbooth: Please relocate the spec links to the cpp files alongside the implementation (this is closer to our normal style)

Libraries/LibWeb/PermissionsAPI/PermissionStore.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+    // 3. Append newEntry to the user agent's permission store.
+    m_entries.append(move(new_entry));
+}
+
+void PermissionStore::remove_permission_store_entry(PermissionDescriptor const& descriptor, URL::Origin const& key)
+{
+    // 1. Remove the entry whose descriptor is descriptor, and whose key is equal to key given descriptor, from the user agent's permission store.
+    m_entries.remove_all_matching([&](auto const& entry) {
+        return entry.descriptor == descriptor && permission_key_is_equal_to(descriptor, entry.permission_key, key);
```
shannonbooth: Presumably we should only need to remove the first matching descriptor here

Libraries/LibWeb/PermissionsAPI/PermissionStore.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+bool permission_key_comparison_algorithm(URL::Origin const& key1, URL::Origin const& key2)
+{
+    // 1. Return key1 is same origin with key2.
+    return key1.is_same_origin(key2);
+}
+
+URL::Origin permission_key_generation_algorithm(URL::Origin const& origin, URL::Origin const&)
+{
+    // Takes an origin origin and an origin embedded origin, and returns a new permission key. If unspecified, this defaults to the default permission key generation algorithm. A feature that specifies a custom permission key generation algorithm MUST also specify a permission key comparison algorithm.
```
shannonbooth: Please wrap comments at around [120 characters](https://github.com/LadybirdBrowser/ladybird/blob/master/Documentation/CodingStyle.md#comments)

Libraries/LibWeb/PermissionsAPI/PermissionStore.h
```diff
@@ -0,0 +1,54 @@
...
+#include <AK/String.h>
+#include <AK/Vector.h>
+#include <LibURL/Origin.h>
+#include <LibWeb/Bindings/PermissionStatusPrototype.h>
+#include <LibWeb/PermissionsAPI/Permissions.h>
+
+namespace Web::PermissionsAPI {
+
+// https://w3c.github.io/permissions/#dfn-permission-store-entry
+struct PermissionStoreEntry {
```
shannonbooth, line 19: This seems to be linking to the wrong item in the spec, I think it should be https://w3c.github.io/permissions/#dfn-permission-store-entry. Let's also initialize state with something by default since it's a primitive member.

Libraries/LibWeb/PermissionsAPI/PermissionStore.cpp
```diff
@@ -0,0 +1,80 @@
...
+    // The default permission key generation algorithm, given an origin origin and an origin embedded origin, runs the following steps:
+
+    // 1. Return origin.
+    return origin;
+
+    // Note: Permission Delegation
+    // Most powerful features grant permission to the top-level origin and delegate access to the requesting document via Permissions Policy. This is known as permission delegation.
+}
+
+PermissionStore& PermissionStore::the()
```
shannonbooth, line 43: We don't need to address this now, but this should be stored at the user agent level (IPC to the browser process). For now I think we should leave a FIXME

Libraries/LibWeb/HTML/Navigator.h
```diff
@@ -20,6 +20,7 @@
 #include <LibWeb/HTML/PluginArray.h>
 #include <LibWeb/HTML/UserActivation.h>
 #include <LibWeb/MediaCapabilitiesAPI/MediaCapabilities.h>
+#include <LibWeb/PermissionsAPI/Permissions.h>
```
shannonbooth: We should not need to include this header here - PermissionsAPI::Permissions should be forward declared in LibWeb/Forward.h which should be enough for the header (Navigator.cpp would need the include). Same in WorkerNavigator.h.

Libraries/LibWeb/HTML/Navigator.idl
```diff
@@ -53,6 +54,9 @@ interface Navigator {
 
     // https://immersive-web.github.io/webxr/#navigator-xr-attribute
     [SecureContext, SameObject] readonly attribute XRSystem xr;
+
+    // https://w3c.github.io/permissions/#navigator-and-workernavigator-extension
+    [SameObject] readonly attribute Permissions permissions;
```
shannonbooth: While we do not have this fully implemented, we should mark this as "Experimental" in the IDL so it is not exposed by default with a not fully working implementation.

Libraries/LibWeb/PermissionsAPI/PermissionStatus.cpp
```diff
@@ -0,0 +1,52 @@
...
+    // 2. Assert: The feature identified by name is supported by the user agent.
+    VERIFY(is_permission_supported(name));
+
+    // 3. Let status be a new instance of the permission result type identified by name:
+    // FIXME: A permission result type is a PermissionStatus or one of its subtypes.
+    // If unspecified, this defaults to PermissionStatus. No subtypes exist yet in Ladybird.
+    auto status = realm.create<PermissionStatus>(realm);
+    {
+        // 1. Initialize status's [[query]] internal slot to permissionDesc.
+        status->m_query = permission_desc;
```
shannonbooth: Let's pass these through the constructor for PermissionStatus instead of initializing afterwards.

Libraries/LibWeb/PermissionsAPI/PermissionStatus.cpp
```diff
@@ -0,0 +1,52 @@
...
+{
+}
+
+GC::Ref<PermissionStatus> PermissionStatus::create(JS::Realm& realm, PermissionDescriptor permission_desc)
+{
+    // 1. Let name be permissionDesc's name.
+    String name = permission_desc.name;
+
+    // 2. Assert: The feature identified by name is supported by the user agent.
+    VERIFY(is_permission_supported(name));
```
shannonbooth, line 36: Considering is_permission_supported is also stubbed out, probably we should leave this as a FIXME here, and remove is_permission_supported

Libraries/LibWeb/PermissionsAPI/PermissionStatus.h
```diff
@@ -0,0 +1,49 @@
...
+    Bindings::PermissionState state() const { return m_state; }
+    void set_state(Bindings::PermissionState state) { m_state = state; }
+
+    // https://w3c.github.io/permissions/#dom-permissionstatus-name
+    String name() const { return m_name; }
+
+    PermissionDescriptor const& query() const { return m_query; }
+
+    // https://w3c.github.io/permissions/#dom-permissionstatus-onchange
+    // FIXME: DOM::EventHandler onchange();
```
shannonbooth, line 35: This should be a simple one to implement - take a look at https://github.com/LadybirdBrowser/ladybird/blob/ea22c19ab32dd878175974c7790cb28a60204816/Libraries/LibWeb/HTML/BroadcastChannel.cpp#L240 as an example.

Libraries/LibWeb/PermissionsAPI/PermissionStatus.h
```diff
@@ -0,0 +1,49 @@
...
+
+public:
+    static GC::Ref<PermissionStatus> create(JS::Realm&, PermissionDescriptor);
+
+    // https://w3c.github.io/permissions/#dom-permissionstatus-state
+    Bindings::PermissionState state() const { return m_state; }
+    void set_state(Bindings::PermissionState state) { m_state = state; }
+
+    // https://w3c.github.io/permissions/#dom-permissionstatus-name
+    String name() const { return m_name; }
```
shannonbooth:
```cpp
String const& name() const { return m_name; }
```

Libraries/LibWeb/PermissionsAPI/Permissions.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+// https://w3c.github.io/permissions/#query-method
+GC::Ref<Web::WebIDL::Promise> Permissions::query(JS::Value permission_desc)
+{
+    auto& realm = this->realm();
+
+    auto& relevant_global_object = HTML::relevant_global_object(*this);
+
+    // 1. If this's relevant global object is a Window object, then:
+    if (is<HTML::Window>(relevant_global_object)) {
```
shannonbooth, line 124: We can use `as_if<HTML::Window>` here as a shorthand for `is `combined if `as`

Libraries/LibWeb/PermissionsAPI/Permissions.cpp
```diff
@@ -0,0 +1,169 @@
+/*
+ * Copyright (c) 2025, Niccolo Antonelli-Dziri <niccolo.antonelli-dziri@protonmail.com>
```
shannonbooth: Here and elsewhere, it's 2026 now

jonbgamble:
This looks great! Ordinarily you can expect a few CI checks to fail because, well Ladybird CI sucks.

But I did notice the 3 fails here are all the same test: Text/input/wpt-import/permissions/permissions-query-permissions-policy-attribute.https.sub.html

I wonder if there's a sporadic exception path here that can sneak through the harness and cause occasional timeouts?

shannonbooth (changes requested):
Gave it a pass over and found a few things. We will need some future changes to hook this up to the browser process so that every web content process shares the same permissions, but for now to get some basic shape while it is marked expermental, I think it's fine.

(Oh, and we also need to fix whatever that test issue is jon mentions above)

Libraries/LibWeb/PermissionsAPI/PermissionStatus.cpp
NiccoloAntonelliDziri: Do I remove is_permission_supported from everywhere and leave fixme ? I think the function is used in two places. I do agree that we could do better than just a function for this but it was the easiest way to keep things simple.

shannonbooth: Ah my bad, yeah fair enough then!

Libraries/LibWeb/PermissionsAPI/PermissionStore.h
NiccoloAntonelliDziri: This is the same link, is this intended ?

shannonbooth: Nope, sorry! For some reason when I left the comment I thought I was looking at https://w3c.github.io/permissions/#dom-permissiondescriptor

NiccoloAntonelliDziri:
I add this comment because I wasn't expecting that test (https://wpt.live/permissions/permissions-query-permissions-policy-attribute.https.sub.html) to still timeout after setting the permission attribute in navigator and worker navigator as experimental. It still should fail (as it correctly does on wpt.live)

Libraries/LibWeb/HTML/WorkerNavigator.cpp
```diff
@@ -39,8 +40,8 @@ void WorkerNavigator::visit_edges(Cell::Visitor& visitor)
     visitor.visit(m_media_capabilities);
     visitor.visit(m_serial);
     visitor.visit(m_service_worker_container);
+    visitor.visit(m_permissions);
 }
-
```
shannonbooth, line 43: nit that I think may have happened during rebase, but this newline should be preserved.

Libraries/LibWeb/PermissionsAPI/Permissions.cpp
```diff
@@ -0,0 +1,213 @@
...
+    // FIXME: 3. Let feature be descriptor's name.
+
+    // FIXME: 4. If there exists a policy-controlled feature for feature and settings' relevant global object has an associated Document run the following step:
+    if (false) {
+        // 1. Let document be settings' relevant global object's associated Document.
+        // 2. If document is not allowed to use feature, return "denied".
+    }
+
+    // 5. Let key be the result of generating a permission key for descriptor with settings's top-level origin and settings's origin.
+    auto key = permission_key_generation_algorithm(settings_object.top_level_origin.value_or(settings_object.origin()), settings_object.origin());
```
shannonbooth: Falling back to settings object's origin does not appear to be part of the spec here, we should assert that top level origin exists.

Libraries/LibWeb/PermissionsAPI/Permissions.h
```diff
@@ -0,0 +1,56 @@
...
+
+#include <AK/Optional.h>
+#include <AK/String.h>
+#include <LibJS/Runtime/Realm.h>
+#include <LibWeb/Bindings/PermissionStatus.h>
+#include <LibWeb/Bindings/PlatformObject.h>
+
+namespace Web::PermissionsAPI {
+
+class PermissionStatus;
```
shannonbooth: Bit of a rebuild for yourself doing this but this can be added for `LibWeb/Forward.h` to forward declare this effectively everywhere

Tests/LibWeb/Text/expected/wpt-import/permissions/permissions-query-permissions-policy-attribute.https.sub.txt
```diff
@@ -0,0 +1,11 @@
+Harness status: Timeout
+
+Found 4 tests
+
+1 Pass
+1 Fail
+2 Timeout
+Timeout	Permissions.state is "prompt" with allow="geolocation" in same-origin iframes.
```
shannonbooth: While it's fine to include tests that have failing expectations while everything is not fully set up, we shouldn't include any which time out. Looks like it also happened to flake on CI: https://github.com/LadybirdBrowser/ladybird/actions/runs/24952937010/job/73066152341?pr=8874#step:16:20219

It would be better to not include this test at all than timing out or flakey

Tests/LibWeb/Text/expected/wpt-import/permissions/idlharness.any.txt
```diff
@@ -0,0 +1,6 @@
+Harness status: OK
+
+Found 1 tests
+
+1 Fail
+Fail	idl_test setup
```
shannonbooth: I'll probably need to pull this branch locally to take a look, but do you happen to know why these are failing on setup?

shannonbooth (commented):
Left some nits for some things I found skimming through again (or maybe changed after rebase?), looks close to me to land some initial infrastructure to be iterated on, but some work needed to figure out what is going on with the tests in particular. Biggest thing is not having the flakey/timing out test though.

Libraries/LibWeb/PermissionsAPI/PermissionStatus.idl
```diff
@@ -0,0 +1,13 @@
+// https://w3c.github.io/permissions/#permissionstatus-interface
+[Exposed=(Window,Worker)]
+interface PermissionStatus : EventTarget {
```
shannonbooth, line 3: We should also mark this interface Experimental here and for the Permissions interface as well

Tests/LibWeb/Text/expected/wpt-import/permissions/idlharness.any.txt
NiccoloAntonelliDziri: I've tried to locate the issue, I corrected a bug (I forgot to remove the [FIXME] in PermissionStatus.idl for the onchange attribute) but it doesn't seem to be changing this, I'm sorry

shannonbooth: Looks like that the issue is that it is not configured under LoadFromHttpServer alongside 

Text/input/wpt-import/permissions/idlharness.any.worker.html

shannonbooth: Okay, and I also figured out the issue with the worker version of the IDL test as well, https://github.com/LadybirdBrowser/ladybird/blob/40ebd94ca91f4b0d5f52d801ee871854df8b19c0/Tests/LibWeb/Text/input/wpt-import/resources/idlharness.js#L3560 is using window.location which obviously doesn't work for workers. This is out of sync with the upstream WPT version of the file: https://github.com/web-platform-tests/wpt/blob/66c44e744bc81953b0c8e5751b75b8ad96e66161/resources/idlharness.js#L3565 due to a local edit to make it work for file://. Changing it to globalThis would fix the issue but I think we can actually just resync the harness file with WPT, I'll MR that.

shannonbooth: https://github.com/LadybirdBrowser/ladybird/pull/9326

NiccoloAntonelliDziri: I have updated TestConfig.ini and the two idlharness.any tests now do not fail on startup anymore using your fix.

NiccoloAntonelliDziri: I guess we probably should rerun CI when your PR is merged

Tests/LibWeb/TestConfig.ini
```diff
@@ -118,6 +118,11 @@ Text/input/wpt-import/css/cssom-view/idlharness.html
 Text/input/wpt-import/gamepad/idlharness.window.html
 Text/input/wpt-import/geolocation/idlharness.https.window.html
 Text/input/wpt-import/notifications/idlharness.https.any.html
+<<<<<<< Updated upstream
```
shannonbooth: It looks like something went wrong when rebasing here unfortunately

Tests/LibWeb/Text/expected/wpt-import/permissions/idlharness.any.worker.txt
```diff
@@ -0,0 +1,57 @@
+<<<<<<< Updated upstream
```
shannonbooth: Same here

Libraries/LibWeb/PermissionsAPI/Permissions.cpp
```diff
@@ -0,0 +1,178 @@
...
+}
+
+// https://w3c.github.io/permissions/#dfn-permission-query-algorithm
+void permission_query_algorithm(PermissionDescriptor const& permission_desc, PermissionStatus& status)
+{
+    // 1. Set status's state to permissionDesc's permission state.
+    status.set_state(permission_desc.permission_state());
+}
+
+JS::ThrowCompletionOr<PermissionDescriptor> PermissionDescriptor::from_value(JS::VM& vm, JS::Value value)
```
shannonbooth: This is something that was only improved very recently, but there have been some recent improvements to the bindings generator which generate the code for this which we can call instead of creating this function now.

`Bindings/Permissions.h` now exposes `Bindings::convert_to_idl_value_for_permission_descriptor` which does what this is meant to be doing here (and actually fixes some minor bugs in this implementation). Unfortunately this would take a bit of rejigging here to change around this structure though to make use of `Bindings::PermissionDescriptor` though.

```
// https://w3c.github.io/permissions/#dom-permissiondescriptor
struct PermissionDescriptor {
```

If you want it could be a follow up though since this was improved on quite a bit after you originally raised the MR (or I can port over myself).

Libraries/LibWeb/PermissionsAPI/PermissionStatus.h
```diff
@@ -30,8 +30,12 @@ class WEB_API PermissionStatus : public DOM::EventTarget {
 
     PermissionDescriptor const& query() const { return m_query; }
 
+    // https://w3c.github.io/permissions/#dfn-permissionstatus-update-steps
+    void update_steps();
+
     // https://w3c.github.io/permissions/#dom-permissionstatus-onchange
-    // FIXME: DOM::EventHandler onchange();
+    void set_onchange(GC::Ptr<WebIDL::CallbackType> event_handler);
+    GC::Ptr<WebIDL::CallbackType> onchange();
```
shannonbooth, line 39: This needs to be squashed into the first commit as the first commit does not build without this change:

```
/Users/shannonbooth/Developer/ladybird/Build/release/Libraries/LibWeb/Bindings/PermissionStatus.cpp:183:76: error: no member named 'onchange' in 'Web::PermissionsAPI::PermissionStatus'
  183 |     auto retval = TRY(throw_dom_exception_if_needed(vm, [&] { return impl->onchange(); }));
      |                                                                      ~~~~  ^
/Users/shannonbooth/Developer/ladybird/AK/Try.h:31:41: note: expanded from macro 'TRY'
   31 |             auto&& _temporary_result = (__VA_ARGS__));                                               \
      |                                         ^~~~~~~~~~~
/Users/shannonbooth/Developer/ladybird/AK/Diagnostics.h:21:5: note: expanded from macro 'AK_IGNORE_DIAGNOSTIC'
   21 |     statement;                                \
      |     ^~~~~~~~~
/Users/shannonbooth/Developer/ladybird/Build/release/Libraries/LibWeb/Bindings/PermissionStatus.cpp:183:23: error: no matching function for call to 'throw_dom_exception_if_needed'
  183 |     auto retval = TRY(throw_dom_exception_if_needed(vm, [&] { return impl->onchange(); }));
      |                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~

```

Libraries/LibWeb/PermissionsAPI/Permissions.h
```diff
@@ -33,17 +32,17 @@ struct PermissionDescriptor {
...
+Bindings::PermissionState request_permission(PermissionDescriptor const& descriptor);
+
 class WEB_API Permissions : public Bindings::PlatformObject {
     WEB_PLATFORM_OBJECT(Permissions, Bindings::PlatformObject);
     GC_DECLARE_ALLOCATOR(Permissions);
 
 public:
     static GC::Ref<Permissions> create(JS::Realm&);
 
-    // https://w3c.github.io/permissions/#query-method
```
shannonbooth: Please squash these FIXUP changes into the commit which originally added these lines

shannonbooth (commented):
Close to landing now, I think we just need to fix up some of the history of the commits so that they are atomic changes that build standalone (with no fixup changes). Currently the commits do not build individually. Probably it only makes sense to have a 

[LibWeb: Add basics for the Permissions API](https://github.com/LadybirdBrowser/ladybird/pull/8874/changes/783a499f540d069a8b3b9417f3e18f9d13b5a73a)

and 

[LibWeb: Support Geolocation permission](https://github.com/LadybirdBrowser/ladybird/pull/8874/changes/0f303c15d682c3015afdc31a228c6730ad3a486d)

Tests/LibWeb/Text/expected/wpt-import/permissions/idlharness.any.txt
shannonbooth: Merged to make it a bit easier to manage the rebase!

Libraries/LibWeb/PermissionsAPI/Permissions.cpp
```diff
@@ -85,10 +123,9 @@ GC::Ref<Web::WebIDL::Promise> Permissions::query(JS::Value permission_desc)
     auto& relevant_global_object = HTML::relevant_global_object(*this);
 
     // 1. If this's relevant global object is a Window object, then:
-    if (is<HTML::Window>(relevant_global_object)) {
+    if (auto* window = as_if<HTML::Window>(relevant_global_object)) {
```
shannonbooth, line 126: One last commit hygine thing - this change should be squashed i to the last commit as that introduced is as/is

shannonbooth (approved):

PR #8877, opened 2026-04-11
test-web: Capture all process output

jonbgamble:
- Suppress browser & helper process console writes so they don't "glitch scroll" stale test status lines above and past the live display area.
- While we're suppressing them, might as well capture and log them.
- Add the run date to the results header.
- Add a button in the header by the filter box that copies the command line used to invoke the run to the clipboard.
- Remove --debug-timeouts because it's not worth the code noise.
- Use files rather than buffers for in-flight captures.

test-web


Libraries/LibWebView/ProcessManager.cpp
```diff
@@ -71,15 +72,21 @@ Optional<Process&> ProcessManager::find_process(pid_t pid)
 
 void ProcessManager::add_process(WebView::Process&& process)
 {
-    Threading::MutexLocker locker { m_lock };
```
shannonbooth, line 74: This mutex seems dubious beforehand as well considering it does not protect find_process properly, but it does not seem correct to remove the mutex here

jonbgamble: HashMap owns the Process object lifetimes, and find_process() returns a borrowed reference into that storage. We can't guarantee its lifetime to another thread.

I would just replace the misleading mutex and locks with event loop thread-affinity asserts, and revisit the design if non-event-loop callers are ever needed.

PR #8881, opened 2026-04-12
LibWeb: Support auto-scrolling contains when the middle mouse is pressed

trflynn89:
I use this about a million times a day in other browsers, so let's have it!

https://github.com/user-attachments/assets/cb442526-b6fe-4132-85d1-78f2dbff366b

### DIFF
--- a/Base/res/ladybird/about-pages/settings.html
+++ b/Base/res/ladybird/about-pages/settings.html
@@ -365,6 +365,21 @@
                     </div>
                 </div>
             </div>
+
+            <h3 class="card-title">Browsing Behavior</h3>
+            <div class="card">
+                <div class="card-body">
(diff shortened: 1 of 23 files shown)


jdahlin:
Not saying this PR is wrong, however on my system (GNOME 49 on Debian 13), middle scroll does not work out of the box, but pasting with middle click does.

trflynn89:
This behavior works by default for me on Windows, Ubuntu, and Fedora. Middle click pasting also works on text boxes, which we don't support yet.

trflynn89:
We can eventually hide this and other autoscroll behavior behind some 'enable autoscrolling' setting as well.

trflynn89:
Went ahead and added a setting for this
autoscroll

PR #8885, opened 2026-04-12
LibWeb: Apply `::first-letter` pseudo-element styles

tcl3:
We now apply first letter styles by splitting text with a first-letter style applied into 2 `TextSliceNode` objects.  The
`DOM::Text` layout  node always points at the non first-letter slice and the first-letter slice is  reachable via
`TextSliceNode::first_letter_slice()`.

First letter splitting works by `TreeBuilder` walking a block container's inline descendants to find the first typographic letter unit per the pattern given in  css-pseudo level 4, which is then wrapped in an anonymous inline box styled with the `::first-letter` computed properties.

Consumers that map between DOM offsets and layout geometry are updated to visit all slices of a `DOM::Text` through `TextOffsetMapping`.

With this change we pass 359/425 of the `css/CSS2/first-letter-punctuation-*` WPT tests because we are following the `css-pseudo-4` spec, which has stricter rules for trailing punctuation.


shannonbooth:
If you want some before and after screenshots, a site that does use this is op.gg to uppercase the first letter of some text

e.g

Before

image


After

image

Libraries/LibWeb/Layout/LayoutState.cpp
```diff
@@ -513,6 +513,24 @@ void LayoutState::commit(Box& root)
...
+    for (auto* paintable : inline_node_paintables) {
+        for (auto const* ancestor = paintable->layout_node().parent(); ancestor; ancestor = ancestor->parent()) {
+            if (!is<InlineNode>(*ancestor) || ancestor->dom_node())
+                break;
+            auto& inline_ancestor = const_cast<InlineNode&>(static_cast<InlineNode const&>(*ancestor));
+            if (inline_ancestor.first_paintable())
+                break;
+            auto line_paintable = inline_ancestor.create_paintable_for_line_with_index(paintable->line_index());
+            inline_ancestor.add_paintable(line_paintable);
+            inline_node_paintables.set(line_paintable.ptr());
```
AtkinsSJ, line 598: Adding to `inline_node_paintables` while iterating it feels a bit dubious... I'm not certain, but if this causes the `HashTable` to rehash, that'll make the iterator's pointers invalid I think?

Libraries/LibWeb/Layout/TreeBuilder.cpp
```diff
@@ -273,6 +278,205 @@ class GeneratedContentImageProvider final
...
+            continue;
+        if (child->is_out_of_flow())
+            continue;
+        auto* inner_block = as_if<BlockContainer>(*child);
+        if (!inner_block)
+            break;
+        // If the descendant block has its own ::first-letter, it takes precedence over the ancestor's.
+        if (auto* dom_element = as_if<DOM::Element>(inner_block->dom_node());
+            dom_element && dom_element->computed_properties(CSS::PseudoElement::FirstLetter))
+            break;
```
AtkinsSJ: Took me a bit to understand this comment. 😅 I think this is meaning that we don't create a FirstLetter node if a child already created one for that same letter?

Libraries/LibWeb/Layout/TreeBuilder.cpp
```diff
@@ -273,6 +278,205 @@ class GeneratedContentImageProvider final
...
+    auto& document = element.document();
+    auto& heap = document.heap();
+
+    auto remainder_slice = heap.allocate<TextSliceNode>(document, dom_text, Node::AttachToDOMNode::Yes, letter_end, full_length - letter_end);
+    auto first_letter_slice = heap.allocate<TextSliceNode>(document, dom_text, Node::AttachToDOMNode::No, 0, letter_end);
+    first_letter_slice->set_generated_for(CSS::PseudoElement::FirstLetter, element);
+    remainder_slice->set_first_letter_slice(*first_letter_slice);
+
+    auto first_letter_wrapper = heap.allocate<InlineNode>(document, nullptr, *first_letter_style);
+    first_letter_wrapper->set_generated_for(CSS::PseudoElement::FirstLetter, element);
```
AtkinsSJ: Do both of these want `set_generated_for()`? That seems a bit odd, but that doesn't necessarily mean it's wrong. I think only the wrapper should need it though.

Libraries/LibWeb/Layout/LayoutState.cpp
tcl3: I've used `values()` to ensure the iteration is safe here.

Libraries/LibWeb/Layout/TreeBuilder.cpp
tcl3: That's correct. I've expanded the comment to make it a bit clearer.

Libraries/LibWeb/Layout/TreeBuilder.cpp
tcl3: You're right - it's not necessary for `first_letter_slice`.

AtkinsSJ (approved):

AtkinsSJ:
I'm not super confident on the Unicode part of this, maybe you could take a look at it @trflynn89?

trflynn89 (approved):

PR #8886, opened 2026-04-12
Fix OKP importKey validation for Ed448/Ed25519/X25519/X448

tete17:
This MR fixes 3 issues with OKP key import in WebCryptoAPI:

* Fix the Ed448 raw key length check — was checking for 448 bits but Ed448 public keys are 456 bits (57 bytes) per RFC 8032 Section 5.2.5. Note: the spec itself has this wrong, see https://github.com/w3c/webcrypto/pull/425#discussion_r3070135408
* Validate JWK key length after base64url-decoding the x/d fields for all 4 OKP curves
* Validate JWK key pair consistency — derive public key from d and verify it matches x

This fixes ~230 previously failing WPT subtests across Ed25519, Ed448, X25519, and X448.


Libraries/LibWeb/Crypto/CryptoAlgorithms.cpp
```diff
@@ -6664,8 +6664,8 @@ WebIDL::ExceptionOr<GC::Ref<CryptoKey>> ED448::import_key(
         // 2. Let data be keyData.
         auto data = move(key_data.get<ByteBuffer>());
 
-        // 3. If the length in bits of data is not 448 then throw a DataError.
-        if (data.size() * 8 != 448)
+        // 3. If the length in bits of data is not 456 then throw a DataError.
+        if (data.size() * 8 != 456)
```
gmta, line 6674: Even though the spec text is wrong, you should not modify it as if it was already fixed. You can make this an `// AD-HOC: ..` step and include the URL to the upstream issue.

tete17: I thought I could be sneaky and since it was pretty obvious the spec was wrong and I attached the link to the MR comment I could get away with it :smile: 

This should be better then

gmta (approved):

PR #8889, opened 2026-04-13
UI/Qt: Reset zoom button shows three digit numbers.

vgaspary:
## Issue

Zooming in or out makes a new "Reset zoom" button appear on the toolbar right between the Bookmark toggle and hamburger menu buttons. It shows the current zoom scale, `90%` or `110%`. The button was wide enough to show two digit percentages but not wide enough for three digit numbers, failing back to `...`.

## Root cause

All the other buttons on this toolbar are square icons of size `16px x 16px`. To make the bar look nicer and not have the icons too close to each other, a general style was applied to the toolbar to fix the size of those buttons to `24px x 24px`. The text `XX%` happens to fit in this box, but `XYZ%` does not. It probably depends on the system font too.

## Solution

I left the height of the toolbar buttons fixed to `24px` but replaced the width with a horizontal padding of `4px` on both sides. This has the same effect on the icon buttons and allows buttons with text to stretch to fit the text.

### DIFF
--- a/UI/Qt/Tab.cpp
+++ b/UI/Qt/Tab.cpp
@@ -146,7 +146,7 @@ Tab::Tab(BrowserWindow* window, RefPtr<WebView::WebContentClient> parent_client,
     m_toolbar->setSizePolicy(QSizePolicy::MinimumExpanding, QSizePolicy::Fixed);
     // This is a little awkward, but without this Qt shrinks the button to the size of the icon.
     // Note: toolButtonStyle="0" -> ToolButtonIconOnly.
-    m_toolbar->setStyleSheet("QToolButton[toolButtonStyle=\"0\"]{width:24px;height:24px}");
+    m_toolbar->setStyleSheet("QToolButton[toolButtonStyle=\"0\"]{padding:0px 4px 0px 4px;height:24px}");
 
     m_hamburger_button_action->setVisible(!Settings::the()->show_menubar());
 


shannonbooth:
It might be me running Qt from MacOS or not understanding the problem, but at least on my machine I can't seem to see this problem:

image

Any chance you could share some screenshots of before and after?

vgaspary:
@shannonbooth, must be that. Here's the before and after on my Ubuntu:

https://github.com/user-attachments/assets/901a3571-e3c0-4cea-8277-b30f671b50ca

https://github.com/user-attachments/assets/37e0b6ba-2b6c-4a7e-ac6b-1f1f74b2e54a

And this is on Windows:

windows

InvalidUsernameException:
I can confirm that I've also had this problem on Linux for quiet a while already. It used to work at one point, but that has been many months. And I am not sure if it broke due to a code change or a Qt update.

shannonbooth:
and this fixes for you? looks reasonable to me (though i should double check the look locally) but please adjust the commit title to state what is being changed/fixed rather than the problem at hand. Maybe something like ”UI/Qt: Allow Reset zoom button to fit three-digit zoom levels” or similar?

vgaspary:
@shannonbooth , yes it does. The description of this PR explains the issue and the fix in detail. The TLDR is, the button used to be of fixed size (square shaped), I made the width flexible to fit any text.

InvalidUsernameException:
Yes, the fix works for me as well. The behavior on my systems is the same as in the two videos.

The fix looks reasonable to me as well. Though, it is a little weird that the hardcoded width has been around for at least 3 years (according to `git blame`) when it seemingly used to work back then. But I think it might not be worth the trouble to fully trace this back in this case.

vgaspary:
@shannonbooth , changed the commit title.

@InvalidUsernameException , isn't the bookmark button rather new? I'm no Qt expert, but that extra button on the left in the horizontal layout could've moved things just over the edge.

shannonbooth (approved):

InvalidUsernameException:
> isn't the bookmark button rather new? I'm no Qt expert, but that extra button on the left in the horizontal layout could've moved things just over the edge.

I'm reasonably sure the bug has been around longer than bookmarks.

wertzuz:
Too much t

PR #8894, opened 2026-04-13
LibWeb: Prevent ReplacedBox from locking table row minimum height

### LINKED ISSUE #8884: Tables have extreme height when they contain images
### Summary

Some tables have a severely exaggerated height when one or more images are in them.

### Operating system

Linux

### Steps to reproduce

1. Visit a website with a table that has images in it
2. See it(?)

### Expected behavior

The table has a height that matches the height of the image

### Actual behavior

The table has a height that massively exceeds the height of the image

https://pixelbrush.neocities.org/entries/2026/04/06

```html
N/A
```

```shell
N/A
```

Image

_No response_

### Contribute a patch?

pranshoe:
Fixes #8884

### DIFF
--- a/Libraries/LibWeb/Layout/TableFormattingContext.cpp
+++ b/Libraries/LibWeb/Layout/TableFormattingContext.cpp
@@ -10,6 +10,7 @@
 #include <LibWeb/Layout/BlockFormattingContext.h>
 #include <LibWeb/Layout/Box.h>
 #include <LibWeb/Layout/InlineFormattingContext.h>
+#include <LibWeb/Layout/ReplacedBox.h>
 #include <LibWeb/Layout/TableFormattingContext.h>
 
 namespace Web::Layout {
@@ -159,6 +160,27 @@ void TableFormattingContext::compute_cell_measures()
         auto max_content_width = calculate_max_content_width(cell.box);
         auto min_content_height = calculate_min_content_height(cell.box, max_content_width);
         auto max_content_height = calculate_max_content_height(cell.box, min_content_width);
+
+        // https://www.w3.org/TR/css-tables-3/#row-layout
+        // The minimum height of a row (without spanning-related height distribution) is defined as the
+        // height of an hypothetical linebox containing the cells originating in the row and where cells
+        // spanning multiple rows are considered having a height of 0px (but their correct baseline). In
+        // this hypothetical linebox, cell heights are considered auto, their width (including borders and
+        // paddings) is forced to the widths and inner spacings of the columns they span, but their other
+        // properties are conserved.
+        // For the purpose of calculating this height, descendants of table cells whose height depends on
+        // percentages of their parent cell' height (see section below) are considered to have an auto
+        // height if they have overflow set to visible, clip, or hidden or if they are replaced elements,
+        // and a 0px height if they have not.
+        // FIXME: For now we're only adjusting replaced elements, but we should probably actually compute
+        // the hypothetical line box here.
+        bool contains_replaced_element = false;
+        cell.box->for_each_in_subtree_of_type<Layout::ReplacedBox>([&](auto const&) {
+            contains_replaced_element = true;
+            return TraversalDecision::Break;
+        });
+        if (contains_replaced_element && computed_values.min_height().is_auto())
+            min_content_height = 0;
 
         // The outer min-content height of a table-cell is max(min-height, min-content height) adjusted by the cell intrinsic offsets.
         auto min_height = computed_values.min_height().to_px(cell.box, containing_block_height);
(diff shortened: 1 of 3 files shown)


pranshoe:
Hi @gmta, could you please check this and help me out if there is any improvement to be made?

Libraries/LibWeb/Layout/TableFormattingContext.cpp
```diff
@@ -160,6 +161,14 @@ void TableFormattingContext::compute_cell_measures()
         auto min_content_height = calculate_min_content_height(cell.box, max_content_width);
         auto max_content_height = calculate_max_content_height(cell.box, min_content_width);
 
+        bool contains_replaced_element = false;
+        cell.box->for_each_in_subtree_of_type<Layout::ReplacedBox>([&](auto const&) {
+            contains_replaced_element = true;
+            return TraversalDecision::Break;
+        });
+        if (contains_replaced_element && computed_values.min_height().is_auto())
+            min_content_height = 0;
```
gmta, line 170: Why is `min_content_height` incorrect before this loop? And if it's incorrect, is that temporary or do we at some point get the correct min content height? And if so, why don't we end up updating the layout through the TFC at that point? And if not, why not?

gmta (commented):
Having to traverse the tree for each cell seems a bit expensive, wondering if this is the right approach.

Libraries/LibWeb/Layout/TableFormattingContext.cpp
pranshoe: - You're right, traversing the tree may be expensive
- First we go to compute_cell_measures, here the height is calculated as the size of the image hence min_content_height is wrong here.
- Next we go to compute_table_height where image size is correctly calculated.

```cpp
if (!row.is_collapsed) {
    if (cell.row_span == 1) {
        row.base_height = max(row.base_height, cell_state.border_box_height());
    }
    row.base_height = max(row.base_height, m_rows[cell.row_index].min_size); // This causes the bug
    row.baseline = max(row.baseline, cell.baseline);
}
```

The min_size being the original size of the image, makes the base_height explode.
- An apporach could be that we can set a flag while building the tree itself if any replaced element is encountered.

pranshoe: On investigating I found some things,
Hooking the flag while building the tree is hectic cause suppose if there is an image, if at some point later in a javascript code the user removes the image, we need to track that case too.

Tracking this variable in tree building stage would incorporate many other issues. Is spending time on this approach valuable for a small issue like this?

pranshoe:
ping @gmta

Psychpsyo:
Is there any spec basis for these changes? The code has a comment saying that it implements https://www.w3.org/TR/css-tables-3/#computing-cell-measures, which seems to have changed since it was implemented.
The correct behavior for this seems to be specified in https://drafts.csswg.org/css-tables-3/#row-layout

(Although this is table cell height layout we're talking about, so maybe there is no good spec reference yet. But that section has a bunch of points and notes about the general principles of how it needs to work)

pranshoe:
> Is there any spec basis for these changes? The code has a comment saying that it implements https://www.w3.org/TR/css-tables-3/#computing-cell-measures, which seems to have changed since it was implemented. The correct behavior for this seems to be specified in https://drafts.csswg.org/css-tables-3/#row-layout
> 
> (Although this is table cell height layout we're talking about, so maybe there is no good spec reference yet. But that section has a bunch of points and notes about the general principles of how it needs to work)

As per https://drafts.csswg.org/css-tables-3/#row-layout

```
In this hypothetical linebox, cell heights are considered auto, their width (including borders and paddings) is forced to the widths and inner spacings of the columns they span, but their other properties are conserved.

For the purpose of calculating this height, descendants of table cells whose height depends on percentages of their parent cell' height (see section below) are considered to have an auto height if they have overflow set to visible, clip, or hidden or if they are replaced elements, and a 0px height if they have not. Testcase!!Testcase
```

This is what is being trying to be done here.

Psychpsyo:
> As per https://drafts.csswg.org/css-tables-3/#row-layout
> 
> ```
> In this hypothetical linebox, cell heights are considered auto, their width (including borders and paddings) is forced to the widths and inner spacings of the columns they span, but their other properties are conserved.
> 
> For the purpose of calculating this height, descendants of table cells whose height depends on percentages of their parent cell' height (see section below) are considered to have an auto height if they have overflow set to visible, clip, or hidden or if they are replaced elements, and a 0px height if they have not. Testcase!!Testcase
> ```
> 
> This is what is being trying to be done here.

A code comment about that would be nice. (Especially since I think the code in this PR only does the replaced element part, so a FIXME for the rest seems in order.)

pranshoe:
The comments have been added

pranshoe:
@Psychpsyo Whom can I ask for a merge?

PR #8896, opened 2026-04-13
LibJS: Handle Unicode ID_Start characters excluded from XID_Start

### LINKED ISSUE #8870: LibJS: SyntaxError when using `U+0E33` as an object property key
### Summary

We get `Unhandled JavaScript exception: [SyntaxError] Unexpected token Invalid. Expected property key` when parser encounters ` ำ` in an object key. Noticed this when trying to run code with TinyLD library bundled in. It's a JS library for detecting the language of the text. Other browsers handle this code just fine.
 

### Operating system

Linux

### Steps to reproduce

N/A

### Expected behavior

N/A

### Actual behavior

N/A

https://codepen.io/aplefull/pen/vEXVwpy?editors=0012

```html
<script>
  const a = {
    ำ: 9,
  };
</script>
```

```shell
N/A
```

_No response_

_No response_

### Contribute a patch?

officialasishkumar:
The ECMAScript specification defines `UnicodeIDStart` as any Unicode code point
with the `ID_Start` derived property. The `unicode_ident` crate that the Rust
lexer uses implements `XID_Start`, which is a proper subset of `ID_Start`.

The two sets differ for characters whose NFKC normalization form does not begin
with an `ID_Start` character. These characters are valid JavaScript identifier
starts per spec but were rejected by the lexer with an `Invalid` token.

The existing code already worked around this for U+309B and U+309C (Katakana
voiced/semi-voiced sound marks). This commit extends that workaround to cover
all 21 code points in `ID_Start \ XID_Start` under Unicode 16:

| Code point | Name |
|---|---|
| U+037A | GREEK YPOGEGRAMMENI |
| U+0E33 | THAI CHARACTER SARA AM |
| U+0EB3 | LAO VOWEL SIGN AM |
| U+309B | KATAKANA-HIRAGANA VOICED SOUND MARK (already handled) |
| U+309C | KATAKANA-HIRAGANA SEMI-VOICED SOUND MARK (already handled) |
| U+FC5E–U+FC63 | Arabic shadda ligature isolated forms |
| U+FE70, U+FE72, U+FE74, U+FE76, U+FE78, U+FE7A, U+FE7C, U+FE7E | Arabic vowel isolated forms |
| U+FF9E | HALFWIDTH KATAKANA VOICED SOUND MARK |
| U+FF9F | HALFWIDTH KATAKANA SEMI-VOICED SOUND MARK |

The `unicode_id_continue` function is similarly updated for those characters
whose NFKC form also contains a non-`ID_Continue` character (a space), so they
remain valid as identifier-continuation characters. U+0E33, U+0EB3, U+FF9E,
and U+FF9F are omitted from that list because their NFKC forms consist solely
of `XID_Continue` characters and are therefore already handled by the crate.

Fixes #8870.

### DIFF
--- a/Libraries/LibJS/Rust/src/lexer.rs
+++ b/Libraries/LibJS/Rust/src/lexer.rs
@@ -38,6 +38,11 @@ use crate::ast::{SharedUtf16String, Utf16String};
 use crate::token::{Token, TokenType};
 use crate::u32_from_usize;
 
+unsafe extern "C" {
+    fn unicode_is_id_start(code_point: u32) -> i32;
+    fn unicode_is_id_continue(code_point: u32) -> i32;
+}
+
 /// State for tracking template literal nesting.
 #[derive(Clone)]
 struct TemplateState {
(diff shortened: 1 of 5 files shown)


Libraries/LibJS/Rust/src/lexer.rs
```diff
@@ -206,15 +206,92 @@ fn is_identifier_continue_cp(cp: u32) -> bool {
 }
 
 fn unicode_id_start(cp: u32) -> bool {
-    // NB: The ECMAScript spec requires ID_Start, not XID_Start.
```
trflynn89, line 209: oof, a couple things here (see https://github.com/LadybirdBrowser/ladybird/pull/8788#discussion_r3036822052 for some info)

1. Let's not use `unicode_ident` at all. As pointed out in the link above, we should avoid duplicating large Unicode tables.
2. That PR added FFI to just invoke `Unicode::code_point_has_identifier_start_property` which should already be doing the right thing. Let's do the same here. That should let us remove the `unicode_ident` package from Cargo.toml.
3. The comments added here are exceptionally verbose.

officialasishkumar: Will update accordingly.

PR #6581, opened 2025-10-26
CI+Meta: Enable tests for All_Debug CI job, excluding some tests

rcorsi:
This issue is a follow up to #6577 to see if it would be possible to run testing in All_Debug CI job.

Add testing step to All_Debug CI in order to avoid introduction of debug macro related faults. Currently these might sneak by if a developer makes changes to some code that uses debug macros, but doesn't run the tests with those debug macros enabled.
    
Several issues have already been identified and solved using All_Debug test verification.
    
see ladybird PRs: #6263, #6264, #6953, #7535, #8567
and ladybird Issues: #7129

LibJS and LibRegex are the main problem areas detected so far.

```
Tests excluded:
       test-web (LibWeb)
       All Benchmarks, TestRegex is the most problematic test
       test-js-bytecode
       test-js-ast
```
    
LibWeb is excluded as it takes too long to run, there are several tests that always fail, and under ctest too much output is generated resulting in ctest crashing due to failed memory allocation, even on a system with 64GB of RAM. Ctest crashes using about 41GB, with RAM still available and swap not used at all.
    
Benchmark tests have similar issues taking too long or using too much memory, so they must be excluded using TESTS_ONLY env variable.
    
test-js-bytecode, test-js-ast, and test-css-tokenizer currently excluded as they fail due to the debug macro output in the comparison file vs baseline diffs. It might be possible to re-enable these if the output can be cleaned up.
    
On our current CI setup, All_Debug tests take approximately 3 minutes.

PR #8897, opened 2026-04-13
Tests/AK: Check json_array_ensure_capacity precondition

### LINKED ISSUE #8876: TestJSON incorrect capacity with mimalloc when BUILD_PRESET is Debug
TestJSON failure when testing in Debug build.

`BUILD_PRESET=Debug ./Meta/ladybird.py test`

```
Running test 'json_array_ensure_capacity'.
FAIL: /home/bobo/projects/myladybird/Tests/AK/TestJSON.cpp:540: EXPECT_EQ(array.values().capacity(), new_capacity) failed with lhs=20 and rhs=16
Failed test 'json_array_ensure_capacity' in 1ms
```

Introduced at #8752 

cc @awesomekling

officialasishkumar:
The `ensure_capacity` contract only guarantees a minimum capacity, so the
postcondition remains an at-least check. Add the precondition requested in
review so the test also verifies that ensure_capacity grows the array.

Fixes #8876.

### DIFF
--- a/Tests/AK/TestJSON.cpp
+++ b/Tests/AK/TestJSON.cpp
@@ -536,8 +536,12 @@ TEST_CASE(json_array_ensure_capacity)
 {
     auto array = setup_json_array();
     size_t new_capacity { 16 };
+    EXPECT(array.values().capacity() < new_capacity);
     array.ensure_capacity(new_capacity);
-    EXPECT_EQ(array.values().capacity(), new_capacity);
+    // The allocator may round up the capacity to a larger value that is
+    // efficient for its internal bookkeeping (e.g. mimalloc's mi_good_size),
+    // so we only require that the resulting capacity is at least new_capacity.
+    EXPECT(array.values().capacity() >= new_capacity);
 }
 
 TEST_CASE(json_array_for_each)


rcorsi:
Hi @officialasishkumar, its a good idea, but did you try the fix locally? I don't think EXPECT_GE is available, which would be available with gtest, but this is not gtest.

Have a look at: `Libraries/LibTest/Macros.h`

and for the BigInt some additional macros are created which could give you ideas. See `Tests/LibCrypto/TestBigInteger.cpp` and for example look for EXPECT_LESS

trflynn89:
@officialasishkumar Between this and other changes that were submitted and don't even build, it is quite difficult to trust any of the 10 PRs that you opened in quick succession. Do not submit PRs from AI that you have not tested and thought through yourself.

See here: https://github.com/LadybirdBrowser/ladybird/blob/master/CONTRIBUTING.md#on-usage-of-ai-and-llms

officialasishkumar:
Replaced EXPECT_GE with the LibTest-supported EXPECT(... >= ...) form and removed the Signed-off-by trailer.

Tests/AK/TestJSON.cpp
```diff
@@ -537,7 +537,10 @@ TEST_CASE(json_array_ensure_capacity)
     auto array = setup_json_array();
     size_t new_capacity { 16 };
     array.ensure_capacity(new_capacity);
-    EXPECT_EQ(array.values().capacity(), new_capacity);
+    // The allocator may round up the capacity to a larger value that is
+    // efficient for its internal bookkeeping (e.g. mimalloc's mi_good_size),
+    // so we only require that the resulting capacity is at least new_capacity.
+    EXPECT(array.values().capacity() >= new_capacity);
```
gmta, line 543: Your commit message mentions `EXPECT_GE`, but you use `EXPECT`?

gmta, line 543: The commit title should use `Tests` or `Tests/AK` as the category.

rcorsi:
What about using NDEBUG and depending if set or not add 4 bytes of padding to the count so that we get an exact check instead of this open ended check.

gmta:
> What about using NDEBUG and depending if set or not add 4 bytes of padding to the count so that we get an exact check instead of this open ended check.

I'd test API expectations rather than implementations. If for some reason `.ensure_capacity()` ends up reserving more bytes than necessary, a test should not necessarily be aware of that imo.

rcorsi:
Sounds good. Should we then at least check that the capacity is less than the new_capacity before calling .ensure_capacity() call?  Just to make sure that .ensure_capacity() does something.

```diff
diff --git a/Tests/AK/TestJSON.cpp b/Tests/AK/TestJSON.cpp
index 866321447b..92c1e180b1 100644
--- a/Tests/AK/TestJSON.cpp
+++ b/Tests/AK/TestJSON.cpp
@@ -536,8 +536,9 @@ TEST_CASE(json_array_ensure_capacity)
 {
     auto array = setup_json_array();
     size_t new_capacity { 16 };
+    EXPECT(array.values().capacity() < new_capacity);
     array.ensure_capacity(new_capacity);
-    EXPECT_EQ(array.values().capacity(), new_capacity);
+    EXPECT(array.values().capacity() >= new_capacity);
 }
 
 TEST_CASE(json_array_for_each)
```

officialasishkumar:
Added the precondition check before ensure_capacity() and kept the allocator-rounded >= postcondition.

Tests/AK/TestJSON.cpp
officialasishkumar: Updated the commit title to use the `Tests/AK` category. Commit: 98e8d02768

officialasishkumar: Updated the commit body so it no longer refers to `EXPECT_GE`; the test uses LibTest `EXPECT(...)` and includes the precondition check. Commit: 98e8d02768

rcorsi:
Commit title/subject shouldn't really be talking about the precondition only. It should be more something like:

```
Tests/AK: Validate pre- and post- conditions for ensure_capacity()

With the switch to mimalloc the ensure_capacity() contract guarantees
only a minimum capacity, not a specific value. Update the post-condition
to an minimum check. Add a pre-condition so the test also verifies that
ensure_capacity() will grow the array when called.
```

(NOTE: commit message lines should be 72 characters or less, trying to stay as close to 72 characters as possible without going over.)

officialasishkumar:
Updated the commit message in 743299e to match the requested subject/body style.

rcorsi:
Would it be possible to trigger the remaining CI jobs, and if all is good, merge this PR?

gmta (approved):

gmta:
This needs to be rebased on master.

officialasishkumar:
Rebased on current `master` and force-pushed. The commit-message lint and diff whitespace checks pass locally.

rcorsi:
Checked the fix with the All_Debug build (#6581) and with the usual Debug build and TestJSON works fine.

With #6581 (but not with the usual Debug build), found another unrelated problem in test-css-tokenizer, looks like I need to exclude that one for All_Debug if I run the testing.

rcorsi:
@gmta I think it's worth starting the remaining CI jobs for this PR. Fix looks good.

PR #8898, opened 2026-04-13
LibWeb: Load @font-face rules inside CSS grouping rules such as @layer

### LINKED ISSUE #8685: @font-face declaration fails if contained in @layer block

officialasishkumar:
The font-loading code in `FontComputer` previously iterated only the
top-level rules of a stylesheet, so `@font-face` declarations nested
inside grouping rules (`@layer`, `@media`, `@supports`, ...) were
silently ignored. This caused fonts declared inside `@layer` blocks to
never be loaded, breaking sites that rely on this pattern.

The fix replaces the flat `sheet.rules()` loop with
`sheet.for_each_effective_rule()`, which already has the correct
recursive semantics and is used for `@keyframes` and `@counter-style`
collection. It respects `@media` / `@supports` conditions (fonts inside
non-matching conditions are not loaded) and descends into `@layer`
blocks (which are unconditional).

The same change is applied to `unload_fonts_from_sheet()` for
symmetry: fonts that were loaded from nested rules are now also
properly disconnected when the stylesheet is removed from the document.

Resolves the FIXME comments that previously marked both functions.

Fixes #8685.

### DIFF
--- a/Libraries/LibWeb/CSS/FontComputer.cpp
+++ b/Libraries/LibWeb/CSS/FontComputer.cpp
@@ -714,11 +714,16 @@ GC::Ptr<FontLoader> FontComputer::load_font_face(ParsedFontFace const& font_face
 
 void FontComputer::load_fonts_from_sheet(CSSStyleSheet& sheet)
 {
-    // FIXME: Handle @font-face and @font-feature-values within grouping rules (@media, @supports, etc)
-    for (auto const& rule : sheet.rules()) {
-        if (auto* font_face_rule = as_if<CSSFontFaceRule>(*rule)) {
+    sheet.for_each_effective_rule(TraversalOrder::Preorder, [&](CSSRule const& const_rule) {
+        // for_each_effective_rule delivers const references; the underlying objects are
+        // non-const, so we cast here to call the mutating methods below.
+        auto& rule = const_cast<CSSRule&>(const_rule);
+
+        // FIXME: Revisit this for conditional group rules that become effective
+        // after the initial stylesheet traversal, such as @media on resize.
+        if (auto* font_face_rule = as_if<CSSFontFaceRule>(rule)) {
             if (!font_face_rule->is_valid())
-                continue;
+                return;
 
             auto font_face = FontFace::create_css_connected(document().realm(), *font_face_rule);
             document().fonts()->add_css_connected_font(font_face);
@@ -731,26 +736,29 @@ void FontComputer::load_fonts_from_sheet(CSSStyleSheet& sheet)
             font_face->load();
         }
 
-        if (auto* font_feature_values_rule = as_if<CSSFontFeatureValuesRule>(*rule))
+        if (auto* font_feature_values_rule = as_if<CSSFontFeatureValuesRule>(rule))
             font_feature_values_rule->clear_caches();
-    }
+    });
 }
 
 void FontComputer::unload_fonts_from_sheet(CSSStyleSheet& sheet)
 {
-    // FIXME: Handle @font-face and @font-feature-values within grouping rules (@media, @supports, etc)
     // https://drafts.csswg.org/css-font-loading/#font-face-css-connection
     // If a @font-face rule is removed from the document, its connected FontFace object is no longer CSS-connected.
-    for (auto const& rule : sheet.rules()) {
-        if (auto* font_face_rule = as_if<CSSFontFaceRule>(*rule)) {
+    sheet.for_each_effective_rule(TraversalOrder::Preorder, [&](CSSRule const& const_rule) {
+        // for_each_effective_rule delivers const references; the underlying objects are
+        // non-const, so we cast here to call the mutating methods below.
+        auto& rule = const_cast<CSSRule&>(const_rule);
+
+        if (auto* font_face_rule = as_if<CSSFontFaceRule>(rule)) {
             if (auto font_face = font_face_rule->css_connected_font_face())
                 unregister_font_face(*font_face);
             font_face_rule->disconnect_font_face();
         }
 
-        if (auto* font_feature_values_rule = as_if<CSSFontFeatureValuesRule>(*rule))
+        if (auto* font_feature_values_rule = as_if<CSSFontFeatureValuesRule>(rule))
             font_feature_values_rule->clear_caches();
-    }
+    });
 }
 
 }
--- a/Libraries/LibWeb/CSS/FontComputer.cpp
+++ b/Libraries/LibWeb/CSS/FontComputer.cpp
@@ -715,8 +715,6 @@ GC::Ptr<FontLoader> FontComputer::load_font_face(ParsedFontFace const& font_face
 void FontComputer::load_fonts_from_sheet(CSSStyleSheet& sheet)
 {
     sheet.for_each_effective_rule(TraversalOrder::Preorder, [&](CSSRule const& const_rule) {
-        // for_each_effective_rule delivers const references; the underlying objects are
-        // non-const, so we cast here to call the mutating methods below.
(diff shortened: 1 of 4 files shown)


jdahlin:
This should ideally include a test case, ideally in two commits:
- Commit 1: Add the test case, ensure that it is broken
- Commit 2: Fix it and make sure it passes

officialasishkumar:
@jdahlin sure, i will do it

Libraries/LibWeb/CSS/FontComputer.cpp
```diff
@@ -695,11 +695,14 @@ GC::Ptr<FontLoader> FontComputer::load_font_face(ParsedFontFace const& font_face
 
 void FontComputer::load_fonts_from_sheet(CSSStyleSheet& sheet)
 {
-    // FIXME: Handle @font-face and @font-feature-values within grouping rules (@media, @supports, etc)
```
Calme1709, line 698: We still have the issue that this only accounts for condition rules whose conditions are met at stylesheet load time, e.g. we won't load a `@font-face` within a `@media` that doesn't match initially but does after a window resize.

Seems out of scope for this PR but can we leave a FIXME?

Calme1709 (changes requested):

Libraries/LibWeb/CSS/FontComputer.cpp
officialasishkumar: Added a FIXME for the dynamic conditional-rule case

Calme1709:
Looks good to me apart from the commit message lint failure (which is caused by having a "signed-off-by" tag). Will obviously require a maintainer's approval.

Just a note that this solves to the same problem as #8780 so only one of them is needed.

officialasishkumar:
Removed the Signed-off-by trailer from the commit message and rebased the branch onto the latest master.

Libraries/LibWeb/CSS/FontComputer.cpp
```diff
@@ -714,11 +714,16 @@ GC::Ptr<FontLoader> FontComputer::load_font_face(ParsedFontFace const& font_face
...
-    for (auto const& rule : sheet.rules()) {
-        if (auto* font_face_rule = as_if<CSSFontFaceRule>(*rule)) {
+    sheet.for_each_effective_rule(TraversalOrder::Preorder, [&](CSSRule const& const_rule) {
+        // for_each_effective_rule delivers const references; the underlying objects are
+        // non-const, so we cast here to call the mutating methods below.
```
AtkinsSJ: This comment isn't needed.

Libraries/LibWeb/CSS/FontComputer.cpp
```diff
@@ -731,26 +736,29 @@ void FontComputer::load_fonts_from_sheet(CSSStyleSheet& sheet)
...
 void FontComputer::unload_fonts_from_sheet(CSSStyleSheet& sheet)
 {
-    // FIXME: Handle @font-face and @font-feature-values within grouping rules (@media, @supports, etc)
     // https://drafts.csswg.org/css-font-loading/#font-face-css-connection
     // If a @font-face rule is removed from the document, its connected FontFace object is no longer CSS-connected.
-    for (auto const& rule : sheet.rules()) {
-        if (auto* font_face_rule = as_if<CSSFontFaceRule>(*rule)) {
+    sheet.for_each_effective_rule(TraversalOrder::Preorder, [&](CSSRule const& const_rule) {
+        // for_each_effective_rule delivers const references; the underlying objects are
+        // non-const, so we cast here to call the mutating methods below.
```
AtkinsSJ: Also not needed. However, we could do with another `// FIXME` here about handling conditional group rules that change. We need to remove `@font-face`s that *were* valid, not that currently are valid, which is tricky.

Tests/LibWeb/TestConfig.ini
```diff
@@ -20,6 +20,7 @@ Text/input/navigation/history-replace-push-then-back.html
 
 ; CSS @font-face url() requires HTTP(s) scheme.
 Text/input/selection-rect-consistency-with-kerning.html
+Text/input/css/font-face-in-layer.html
```
AtkinsSJ: Have you confirmed this is needed? We use `@font-face url()` in a lot of tests and no others seem to need this. (ping @gmta, what happened with that kerning test to require this?)

AtkinsSJ (changes requested):
Generally looks good, just nits.

Libraries/LibWeb/CSS/FontComputer.cpp
officialasishkumar: Removed in 72757b6fc4.

Libraries/LibWeb/CSS/FontComputer.cpp
officialasishkumar: Removed the redundant comment and added the unload-side FIXME in 72757b6fc4.

Tests/LibWeb/TestConfig.ini
officialasishkumar: Removed this from `LoadFromHttpServer` in 72757b6fc4 so the test runs in the normal file mode.

PR #8899, opened 2026-04-13
LibWeb: Activate focused elements on Enter and Space key presses

### LINKED ISSUE #6703: `Enter` or space key should trigger `onclick` event for focused button

officialasishkumar:
When Enter or Space is pressed and no text-editing target has already consumed
the event, fire a synthetic click on the focused element if it exposes an
activation behavior (buttons, inputs, summary elements, anchors, etc.).

Before this change, pressing Enter or Space on a keyboard-focused `<button>`,
`<input type="button">`, `<input type="submit">`, or similar element had no
effect. Mouse clicks worked fine, but keyboard-only users could not activate
these elements.

**Enter** activates any focused element with `has_activation_behavior() == true`.

**Space** activates the same set of elements, with one exception: anchor
(`<a>`) elements are skipped so that the pre-existing page-scroll behavior
still applies when a link is focused.

Fixes #6703.

### DIFF
--- a/Libraries/LibWeb/Page/EventHandler.cpp
+++ b/Libraries/LibWeb/Page/EventHandler.cpp
@@ -1968,6 +1968,23 @@ EventResult EventHandler::handle_keydown(UIEvents::KeyCode key, u32 modifiers, u
         }
     }
 
+    // AD-HOC: UI Events describes keydown default actions in general, but does not define this exact element set.
+    // NB: Use activation behavior here so keyboard activation follows the same element boundary as pointer activation.
+    // Space still falls through for anchors so the page-scroll behavior below
+    // remains.
+    if (!target && !modifiers && (key == UIEvents::KeyCode::Key_Return || key == UIEvents::KeyCode::Key_Space)) {
+        auto* focused = document->focused_area().ptr();
+        auto* html_element = focused ? as_if<HTML::HTMLElement>(*focused) : nullptr;
+        if (html_element && html_element->has_activation_behavior()) {
+            // AD-HOC: Space does not activate anchor elements; the page-scroll below still applies.
+            bool should_activate = key == UIEvents::KeyCode::Key_Return || !is<HTML::HTMLAnchorElement>(*html_element);
+            if (should_activate) {
+                html_element->click();
+                return EventResult::Handled;
+            }
+        }
+    }
+
     // FIXME: Implement scroll by line and by page instead of approximating the behavior of other browsers.
     auto arrow_key_scroll_distance = 100;
     auto page_scroll_distance = document->window()->inner_height() - (document->window()->outer_height() - document->window()->inner_height());


Libraries/LibWeb/Page/EventHandler.cpp
```diff
@@ -1931,6 +1931,24 @@ EventResult EventHandler::handle_keydown(UIEvents::KeyCode key, u32 modifiers, u
         }
     }
 
+    // https://www.w3.org/TR/uievents/#dfn-activation-trigger
```
Psychpsyo: 1. This link doesn't go to any actual section of that spec.
2. It should also link to the editor's draft of the spec instead: https://w3c.github.io/uievents
3. The text below seems to not actually be anywhere in the spec. If it tells us to do this, you should link to where it does that and quote it verbatim.

Libraries/LibWeb/Page/EventHandler.cpp
```diff
@@ -1931,6 +1931,25 @@ EventResult EventHandler::handle_keydown(UIEvents::KeyCode key, u32 modifiers, u
         }
     }
 
+    // https://w3c.github.io/uievents/#keydown
+    // If the key is the Enter or Space key and the current focus is on a
+    // state-changing element, the default action MUST be to dispatch a click
+    // event.
+    // Space still falls through for anchors so the page-scroll behavior below
```
Psychpsyo, line 1973: This and the comment below are still not in the cited spec section. If there is no spec for this, they should be marked as `// AD-HOC:`.

Also, the spec says something about "state-changing elements" but your code checks for all elements with activation behavior, with some special considerations for `<a>` elements. That seems wrong. If it isn't, some `// NB:` comment as to why this is correct would be nice.

Fwiw: I think https://github.com/WICG/aom/issues/168 might be related.

officialasishkumar: Updated the comments to mark this as AD-HOC and added an NB explaining that activation behavior is used to match the pointer activation boundary.

Psychpsyo: A link to the related spec issue would also be nice so that whoever gets to resolve this mess in the future knows where to check on the current status of how UIEvents "does not define this exact element set."

PR #8901, opened 2026-04-13
LibWeb: Avoid O(n²) behavior in NamedNodeMap::supported_property_names

### LINKED ISSUE #7980: LibWeb: Accidentally-quadratic-behavior in NamedNodeMap
### Summary

When there are DOM nodes with a significant number of attributes, Ladybird is unnecessarily slow.

All of this started by noticing this accidentally-quadratic behavior in `NamedNodeMap::supported_property_names`:
```c++
    for (auto const& attribute : m_attributes) {
        auto const attribute_name = attribute->name();
        if (!names.contains_slow(attribute_name))      // Takes O(n) time in length of `names`
            names.append(attribute_name.to_string());  // Makes `names` longer with each iteration
```

However, it turns out that even for hundreds of `m_attributes`, this effect doesn't quite become obvious. So this is "only" a DoS vulnerability, and not an actual performance issue. Here's a plot depending on the number of attributes:

Image

As you can see from the violet line, the "before" behavior is clearly quadratic.

However, when I tried to fix this (by using a HashTable when there are "many" attributes), the time didn't go quite linear, as you can see from the green line. For some reason, there are still a linear number of calls **TO** `NamedNodeMap::supported_property_names()`, which needs to be fixed, too. That's the reason why I don't want to make a PR myself: I don't quite understand how this all ties together.

Also, `HashTable` is suspiciously slow. For n=200, I would have expected that a hash set *easily* beats a vector, but for some reason the pay-off is only around n=500? This is weird, I suspect a problem in HashTable. Here's a graph to show that using a HashTable for large n still makes this worse when n<300:

Image

Here's the code to produce the plots above: https://github.com/BenWiederhake/ladybird/tree/historic/namednodemap-quadratic

Suggested test-case: https://github.com/BenWiederhake/ladybird/commit/821c61f298cee9bc14985190f81de3655c3f16fb

### Operating system

Linux

### Steps to reproduce

1. Open a HTML document that contains a DOM node with excessively many attributes, like the one in https://github.com/BenWiederhake/ladybird/commit/821c61f298cee9bc14985190f81de3655c3f16fb
2. Access the NamedNodeMap's supported properties in any way

### Expected behavior

Accessing the NamedNodeMap's supported properties in any way is extremely slow, easily taking several *seconds*.

### Actual behavior

Accessing the NamedNodeMap's supported properties should be in the order of 1-2 *milli*seconds. Observe that Firefox and Chrome usually stay below 1 ms.

N/A

```html
(see https://github.com/BenWiederhake/ladybird/tree/historic/namednodemap-quadratic-test for the test case, as the file itself is somewhat large.)
```

```shell
(no log output that seems relevant)
```

_No response_

_No response_

### Contribute a patch?

officialasishkumar:
`NamedNodeMap::supported_property_names` deduplicates attribute names while
preserving order. The previous loop called `Vector::contains_slow()` on
`names` at every iteration, which is O(n) in the size of `names`. With n
attributes the deduplication step is therefore O(n²).

Replace the linear scan with a `HashTable<FlyString>` that tracks which
names have already been appended. `HashTable::set()` returns
`HashSetResult::InsertedNewEntry` for first-seen names and
`HashSetResult::KeptExistingEntry` for duplicates, so a single O(1)
amortized call replaces the previous O(n) scan.

The hash table is pre-allocated to the same capacity as the attribute
list to avoid reallocation in the typical case where all attribute names
are unique.

Fixes #7980.

### DIFF
--- a/Libraries/LibWeb/DOM/NamedNodeMap.cpp
+++ b/Libraries/LibWeb/DOM/NamedNodeMap.cpp
@@ -6,6 +6,7 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
+#include <AK/HashTable.h>
 #include <LibWeb/Bindings/NamedNodeMapPrototype.h>
 #include <LibWeb/DOM/Attr.h>
 #include <LibWeb/DOM/Document.h>
@@ -55,9 +56,14 @@ Vector<FlyString> NamedNodeMap::supported_property_names() const
     Vector<FlyString> names;
     names.ensure_capacity(m_attributes.size());
 
+    // Use a hash set to track names seen so far, avoiding the O(n²) cost of
+    // contains_slow() on the output vector for elements with many attributes.
+    HashTable<FlyString> seen;
+    seen.ensure_capacity(m_attributes.size());
+
     for (auto const& attribute : m_attributes) {
         auto const attribute_name = attribute->name();
-        if (!names.contains_slow(attribute_name))
+        if (seen.set(attribute_name) == HashSetResult::InsertedNewEntry)
             names.append(attribute_name.to_string());
     }
 


konradekk (commented):
(some kind of java-like `LinkedHashSet` would be a thing here, i guess? 🤔)

BenWiederhake:
Make sure that you include actual performance-measurements, because for low numbers of attributes (which is still very high, and probably covers *ALL* the websites that a regular user might visit), the quadratic version is *FASTER*. See second screenshot in the linked issue.

Issue #8740, opened 2026-04-02
CSS: Dashed borders cannot share corners


theeyeofcthulhu:
### Summary

When I was looking at my local [syncthing](https://syncthing.net/) instance, I noticed that the caret drop-down buttons on the top-right corner of the page were rendering as blocks instead of down-facing carets. Turns out they're using a single block in a dashed top-border and clipping that into a triangle with two transparent borders to the sides:
```css
.caret {
  display: inline-block;
  width: 0;
  height: 0;
  margin-left: 2px;
  vertical-align: middle;
  border-top: 4px dashed;
  border-top: 4px solid \9;
  border-right: 4px solid transparent;
  border-left: 4px solid transparent;
  box-sizing: border-box;
}
```
Ladybird, however, as I learned when reviewing the code in [BorderPainting.cpp](https://github.com/LadybirdBrowser/ladybird/blob/master/Libraries/LibWeb/Painting/BorderPainting.cpp), can currently only make solid borders share corners, not other styles.

### Operating system

Linux

### Steps to reproduce

1. Open the reduced test case in Ladybird.
2. Observe the two different renders which should both be identical.

### Expected behavior

Dashed borders and other styles can share corners.

### Actual behavior

Dashed borders, when the other borders are transparent, for instance, claim the entire corner.

### URL for a reduced test case

N/A

### HTML/SVG/etc. source for a reduced test case

```html
<!DOCTYPE html>
<html lang="en">
<head>
<meta charset="UTF-8">
<style>
.caret-working {
  display: inline-block;
  width: 0;
  height: 0;
  margin-left: 2px;
  vertical-align: middle;
  border-top: 4px dashed;
  border-right: 4px solid transparent;
  border-left: 4px solid transparent;
  box-sizing: border-box;
}
.caret-broken {
  display: inline-block;
  width: 0;
  height: 0;
  margin-left: 2px;
  vertical-align: middle;
  border-top: 4px solid;
  border-right: 4px solid transparent;
  border-left: 4px solid transparent;
  box-sizing: border-box;
}
</style>
</head>
<body>
  <span class="caret-working"></span>
  <span class="caret-broken"></span>
</body>
</html>
```

### Log output and (if possible) backtrace

```shell
N/A
```

### Screenshots or screen recordings

<img width="1913" height="1029" alt="Image" src="https://github.com/user-attachments/assets/1a56f3e5-15d7-4822-9ab9-ccb123e1c7e1" />
[...]

PR #8904, opened 2026-04-13
LibWeb: Allow non-solid borders to share corners with adjacent borders

aguiarcode:
This change resolves issue #8740.

Previously, only solid borders could use path-based corner-sharing
geometry. Dashed and dotted borders would claim the entire corner
when adjacent to transparent borders (e.g. the common CSS triangle/
caret trick using `border-top: 4px dashed` with transparent left/right
borders).

Now, non-solid borders fall through to the corner-sharing path when
adjacent borders have non-zero width, producing the same triangle
shape as solid borders.

### DIFF
--- a/Libraries/LibWeb/Painting/BorderPainting.cpp
+++ b/Libraries/LibWeb/Painting/BorderPainting.cpp
@@ -261,6 +261,89 @@ void paint_border(DisplayListRecorder& painter, BorderEdge edge, DevicePixelRect
             painter.draw_line(p1.to_type<int>(), p2.to_type<int>(), color, border_data.width.value(), gfx_line_style);
             return;
         }
+
+        // FIXME: Border-radius is not yet handled here; falls through to solid path-based rendering below.
+        bool has_no_radius = radius.horizontal_radius == 0 && radius.vertical_radius == 0
+            && opposite_radius.horizontal_radius == 0 && opposite_radius.vertical_radius == 0;
+        if (has_no_radius) {
+            auto fill_corner_triangle = [&](DevicePixelPoint outer_corner, DevicePixelPoint outer_cut, DevicePixelPoint inner_cut) {
+                Gfx::Path corner_path;
+                corner_path.move_to(Gfx::FloatPoint(outer_corner.to_type<int>()));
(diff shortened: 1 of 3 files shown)


aguiarcode:
CI failures are unrelated to this change. css-background-clip-text.html, scroll-timeline-invalidation.html, and form-reset-callback.html are all pre-existing and unrelated to border painting.

AtkinsSJ:
> CI failures are unrelated to this change. css-background-clip-text.html, scroll-timeline-invalidation.html, and form-reset-callback.html are all pre-existing and unrelated to border painting.

This does seem caused by your changes. For example, here's `Screenshot/input/css-background-clip-text.html` without your changes:

image

And with your changes:

image

AtkinsSJ:
This is better, but still isn't quite right.

The code here either draws a fancy (dotted/dashed/etc) border, OR it draws a solid border with correct corners. Really we want fancy borders with correct corners.

For example, changing your test case to have width and height, shows this up:

```html
<!DOCTYPE html>
<style>
  div {
    width: 100px;
    height: 100px;
    border-top: 40px dotted red;
    border-left: 40px solid transparent;
    border-right: 40px solid transparent;
  }
</style>
<div></div>
```
Screenshot_20260429_125231

Chromium:
Screenshot_20260429_125244

Firefox:
Screenshot_20260429_125303

They do disagree on how to render the corners, which is amusing, but either way we still want to draw the middle section of the border correctly. As a simpler fix, you could paint the corner section solid (like the left side of the Firefox example,) and then the rest of the border fancy. That would mean drawing it as 3 sections instead of 1, but shouldn't be too complicated I think.

aguiarcode:
Hey Sam, thanks again for taking the time to review this and for the detailed examples, it really helped me understand exactly what was needed. I've updated the branch to paint the border in 3 sections when corner sharing is needed: solid corner triangles + the fancy line style in the middle. This covers the no-border-radius case. The radius case still falls through to solid rendering for now (left a FIXME), happy to follow up on that separately if you think it's worth it..Really appreciate the patience and guidance throughout this PR! It's always an honor to contribute to this project, I'm so excited to use it!

PR #8908, opened 2026-04-13
LibWeb: Correct initiator origin logic for new top level traversables

jonbgamble:


Libraries/LibWeb/HTML/TraversableNavigable.cpp
```diff
@@ -94,7 +94,7 @@ GC::Ref<TraversableNavigable> TraversableNavigable::create_a_new_top_level_trave
     // document: document (now owned by Navigable::m_active_document, not DocumentState)
 
     // initiator origin: null if opener is null; otherwise, document's origin
-    document_state->set_initiator_origin(opener ? Optional<URL::Origin> {} : document->origin());
+    document_state->set_initiator_origin(opener ? document->origin() : Optional<URL::Origin> {});
```
shannonbooth, line 97: oof nice spot, any possbility of adding a test for this?

trflynn89:
Just noting so it doesn't get lost - the test added here failed in one of the runs:
```diff
--- /home/runner/_work/ladybird/ladybird/Tests/LibWeb/Text/expected/./HTML/window-open-about-blank-inherits-opener-origin.txt
+++ /home/runner/_work/ladybird/ladybird/Tests/LibWeb/Text/expected/./HTML/window-open-about-blank-inherits-opener-origin.txt
@@ -1,3 +1,3 @@
 initial popup.location.href: about:blank
 initial popup.document.body.textContent: Hello Shannon!
-reloaded popup.document.body.textContent: 
+reloaded popup.document.body.textContent: Hello Shannon!
```

jonbgamble:
Nice one. I don't suppose I could blame this on Shannon?

Tests/LibWeb/Text/input/HTML/window-open-about-blank-inherits-opener-origin.html
```diff
@@ -0,0 +1,42 @@
...
+                const currentDocument = popup.document;
+                const currentTextContent = currentDocument.body ? currentDocument.body.textContent : "";
+
+                if (currentDocument !== initialDocument || currentTextContent !== initialTextContent)
+                    break;
+            } catch (error) {
+            }
+
+            await timeout(10);
+        }
```
gmta, line 31: I believe WPT's standard way of dealing with these types of updates should work immediately; i.e. two nested `requestAnimationFrame(..)`s and a `setTimeout(.., 0)` inside of that.

shannonbooth: For my understanding (I was wondering how we could do this) what is the magic which makes that timing okay? (specifically 2 animation frames)

gmta: As I understand, the first rAF triggers within the current frame (nothing updated yet), the second one actually enforces executing on the next rendering opportunity (layout/style updated) but might not yet have triggered the painting logic to the point where compositing is done, so the inner setTimeout() makes sure we schedule a task to let painting do its thing and taking a screenshot will reliably work.

For different test cases different combinations of these three might work equally well, but rAF+rAF+setTimeout has been easy to remember as something that "just works" for me. For screenshot tests I can confirm that leaving out any of the three has caused tests to be flaky in the past.

jonbgamble: I could try it, but we're waiting on the popup reload to replace the Document, which likely takes longer than waiting for renders to settle.

shannonbooth:
```html
<!DOCTYPE html>
<script>
    function waitForNextLoad(window) {
        return new Promise(resolve => {
            window.addEventListener("load", resolve, { once: true });
        });
    }

    asyncTest(async done => {
        const popup = window.open();
        const initialTextContent = "Hello Shannon!";

        println(`initial popup.location.href: ${popup.location.href}`);

        try {
            popup.document.body.innerHTML = `<p>${initialTextContent}</p>`;
            println(`initial popup.document.body.textContent: ${popup.document.body.textContent}`);
        } catch (error) {
            println(`initial popup document access threw: ${error.name}: ${error.message}`);
        }

        // The initial about:blank document's load event may still be queued.
        // Let it fire before arming the listener for the reload.
        await timeout(0);

        const reloaded = waitForNextLoad(popup);
        popup.location.reload();
        await reloaded;

        try {
            println(`reloaded popup.document.body.textContent: ${popup.document.body.textContent}`);
        } catch (error) {
            println(`reloaded popup document access threw: ${error.name}: ${error.message}`);
        }

        popup.close();
        done();
    });
</script>
```

Should make this test work without any busy waits, but for another bug in ladybird for some reason this test doesn't work...

shannonbooth:
Oh wait no it does work, but doesn't work in test-web for some reason

shannonbooth:
Nevermind, I'm being silly, that also doesn't work

shannonbooth:
I can't think of any other good methods to test this unfortunately

PR #8913, opened 2026-04-14
LibGC: Implement incremental garbage collection

marcbutler:
Convert collector to run incrementally, using standard tricolor marking and a hybrid of Yuasa (SATB, deletion) and Dijkstra (insertion) write barriers that shade both old and new targets gray.

Marking and sweeping budgets specified in cells, and blocks respectively, and there has been no attempt to tune either.

This code is more than a proof-of-concept, but lacks refinement.

### DIFF
--- a/Libraries/LibGC/Cell.h
+++ b/Libraries/LibGC/Cell.h
@@ -38,6 +38,12 @@ public:                                            \
     }                                              \
     friend class GC::Heap;
 
+enum class Color : u8 {
+    White, // Not yet reached — will be reclaimed if still white at sweep time.
+    Gray,  // Reached but outgoing edges not yet traced (on marking worklist).
+    Black, // Reached and fully traced.
+};
+
 class GC_API Cell {
     AK_MAKE_NONCOPYABLE(Cell);
     AK_MAKE_NONMOVABLE(Cell);
@@ -48,8 +54,11 @@ public:
 
     virtual ~Cell() = default;
 
-    bool is_marked() const { return m_mark; }
-    void set_marked(bool b) { m_mark = b; }
+    bool is_marked() const { return m_color != Color::White; }
+    void set_marked(bool b) { m_color = b ? Color::Black : Color::White; }
+
+    Color gc_color() const { return m_color; }
+    void set_gc_color(Color color) { m_color = color; }
 
     enum class State : bool {
         Live,
@@ -140,6 +149,13 @@ public:
         }
 
         template<typename T>
+        void visit(ValueVector<T> const& vector)
+        requires(IsBaseOf<NanBoxedValue, T>)
+        {
+            visit_impl(ReadonlySpan<NanBoxedValue>(vector.span().data(), vector.size()));
+        }
+
+        template<typename T>
         void visit(HashTable<T> const& table)
         {
             for (auto& value : table)
@@ -215,7 +231,7 @@ protected:
     Cell() = default;
 
 private:
-    bool m_mark { false };
+    Color m_color { Color::White };
     State m_state { State::Live };
 };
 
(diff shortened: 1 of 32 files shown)


jdahlin:
@marcbutler have you tried running the whole https://github.com/LadybirdBrowser/js-benchmarks/ with and without these changes? So we can see the performance impact of it in the individual tests.

marcbutler:
> @marcbutler have you tried running the whole https://github.com/LadybirdBrowser/js-benchmarks/ with and without these changes? So we can see the performance impact of it in the individual tests.

@jdahlin No I have not but I will look into doing so. I have only just got it working, and wanted to get feedback as early as possible.

I do not anticipate an improvement in performance. The incremental collector with the right tuning should smooth out collection pauses, but I suspect the write barriers overhead will reduce performance.

awesomekling:
Hello! Thanks for posting this for early feedback.

This is an unusually broad and correctness-sensitive change set: it changes LibGC’s marking model, adds write barriers across a large number of LibJS mutation sites, changes weak/finalization behavior, and also reaches into the event loop and LibWeb.

Given that scope, I think the patch needs three things before it can be reviewed usefully:

1. A design writeup that explains the collector invariants and the tradeoffs behind the chosen barrier scheme, incremental sweep, root rescanning, weak/finalization semantics, and allocation coloring.

2. Validation that barrier coverage is complete, plus targeted tests for the new correctness-sensitive behavior.

3. Quantitative results: benchmark impact, pause-time measurements, and memory-usage changes.

Without that, it is hard to distinguish a promising prototype from a patch that only appears coherent at a high level.

marcbutler:
Hello @awesomekling

> This is an unusually broad and correctness-sensitive change set: it changes LibGC’s marking model, adds write barriers across a large number of LibJS mutation sites, changes weak/finalization behavior, and also reaches into the event loop and LibWeb.

Agreed. I had hoped to keep the changes largely isolated to LibGC, but in addressing crashes I found that it was not practical if I was to have a working prototype. I am still unsure if the hybrid write barrier is a good design.

I thought "showing something" would be better than "talking about it", and moving from stop the world (STW) to incremental seems like the most useful next evolution for the GC in this project.

> 1. A design writeup that explains the collector invariants and the tradeoffs behind the chosen barrier scheme, incremental sweep, root rescanning, weak/finalization semantics, and allocation coloring.

This could potentially be a separate forerunning PR consisting of one or more files in Documentation/ ?

> 2. Validation that barrier coverage is complete, plus targeted tests for the new correctness-sensitive behavior.

To make this tractable I need to find a way to break this line of work into pieces. On discord @AtkinsSJ suggested
that a test suite for the current GC would be helpful. 

So I'm thinking that a test-suite targeted at the current GC is a logical first step and useful foundation to build off.

> 3. Quantitative results: benchmark impact, pause-time measurements, and memory-usage changes.

This sounds like building upon, and augmenting existing instrumentation and infrastructure. Performance is important, but my current intuition is that the write barriers will likely degrade overall performance but the incremental marking will smooth out the STW GC marking interruptions. I think that it is a reasonable initial tradeoff.

> Without that, it is hard to distinguish a promising prototye from a patch that only appears coherent at a high level.

Understood. I would appreciate guidance on what is thought to be the best approach to dividing up the work, such as multiple PRs, or parallel development branch, or ?

awesomekling:
Thanks for the work here, and for being upfront that it's a prototype posted for feedback.

I've thought about this a bit and I'm going to close the PR. Replacing the GC is one of the higher-risk changes we could take on, and I don't think it's the right fit for us to work through. I'd rather say that now than have you keep iterating on something I'm not confident we'd merge. A rewrite of this scope needs to arrive with the design, invariants, and validation already worked out, rather than figured out during review.

PR #8918, opened 2026-04-14
Tests: Remove insecure location.hash evaluation in Range-test-iframe.html

fam007e:
This resolves CVE-2024-26466 by removing the old, unused location.hash injection vector in the Range test iframe. Modern WPT tests use window.testRangeInput instead.


shannonbooth:
I am going to close this, this does not seem right at all - this is a direct mirror of https://github.com/web-platform-tests/wpt/blob/master/dom/ranges/Range-test-iframe.html so "Modern WPT tests use window.testRangeInput instead" does not really make sense here, and how could this possibly be a vulnerability in a WPT test anyhow?

PR #8920, opened 2026-04-14
LibGfx: Fall back when dmabuf modifiers are unsupported

olliehcrook:
If create_shared_vulkan_image() cannot provide any of the requested DRM format modifiers, return an
  error so BackingStoreManager falls back to
ShareableBitmap backing stores instead of continuing into a dmabuf import that later aborts.

### DIFF
--- a/Libraries/LibGfx/VulkanImage.cpp
+++ b/Libraries/LibGfx/VulkanImage.cpp
@@ -119,6 +119,11 @@ ErrorOr<NonnullRefPtr<VulkanImage>> create_shared_vulkan_image(VulkanContext con
         }
     }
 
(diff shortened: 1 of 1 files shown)


olliehcrook:
This fixes a failure on systems like mine where Vulkan is available, but the driver does not expose a renderable linear modifier for these shared dmabuf backing stores.

Failure:
Screenshot from 2026-04-14 21-58-31

This fix handles that at runtime by falling back when the requested modifier is unsupported. I did wonder whether this should be gated earlier in Meta/CMake/vulkan.cmake, but it seems like the actual failure depends on runtime driver/GPU capabilities rather than something we can reliably decide at build time.

kalenikaliaksandr (approved):

PR #8921, opened 2026-04-14
LibWeb/CSS: Use resolved serialization for typed custom property values

shannonbooth:
Substitute typed custom-property values using component values equivalent to their computed value.

This fixes the header bar on https://www.semrush.com appearing as white.

Very unsure if this is the correct approach.

Original reduction I made was:

```html
<!doctype html>
<style>
@property --x {
    syntax: "<percentage>";
    inherits: true;
    initial-value: 0%;
}

div {
    height: 80px;
    background: color-mix(in oklab, white var(--x), #dceeeb);
}
</style>
<div></div>
```

Before:

image

After:

image

### DIFF
--- a/Libraries/LibWeb/CSS/Parser/ArbitrarySubstitutionFunctions.cpp
+++ b/Libraries/LibWeb/CSS/Parser/ArbitrarySubstitutionFunctions.cpp
@@ -85,6 +85,15 @@ bool contains_guaranteed_invalid_value(Vector<ComponentValue> const& values)
     return false;
 }
 
+static Vector<ComponentValue> substitution_component_values_equivalent_to_computed_value(DOM::Document const& document, StyleValue const& value)
+{
+    if (value.is_unresolved())
+        return value.as_unresolved().values();
+
+    auto serialized = value.to_string(SerializationMode::ResolvedValue);
+    return Parser::create(ParsingParams { document }, serialized).parse_as_list_of_component_values();
+}
+
 // https://drafts.csswg.org/css-values-5/#replace-an-attr-function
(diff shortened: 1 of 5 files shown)


AtkinsSJ:
> Very unsure if this is the correct approach.

Yeah this doesn't sit right with me. That percentage in your example should produce the same CV regardless of whether it's been serialized first or not.

...looking into this, it's because `DimensionStyleValue::tokenize()` is wrong for percentages! It creates a Dimension token instead of a Percentage. My bad. :sweat_smile: 

So I think this is the only fix necessary? Seems to work for me locally.

```c++
Vector<Parser::ComponentValue> DimensionStyleValue::tokenize() const
{
    if (unit_name() == "percent"sv)
        return { Parser::Token::create_percentage(Number { Number::Type::Number, raw_value() }) };
    return { Parser::Token::create_dimension(raw_value(), FlyString::from_utf8_without_validation(unit_name().bytes())) };
}
```

AtkinsSJ:
I poked around a little and opened #8925. It seems there are still some issues with `tokenize()` in places, as it's not getting as many passes as your changes here. Leaving it up to you if you want to investigate some more, otherwise I'll take a look at some point. :^) Thanks for finding this originally!

shannonbooth:
> I poked around a little and opened #8925. It seems there are still some issues with `tokenize()` in places, as it's not getting as many passes as your changes here. Leaving it up to you if you want to investigate some more, otherwise I'll take a look at some point. :^) Thanks for finding this originally!

Ha, I actually had an earlier revision which was identical to https://github.com/LadybirdBrowser/ladybird/pull/8925. I'll close this MR though

PR #8924, opened 2026-04-15
LibGfx+LibWeb: Ensure sk_image is available in ::export_to_byte_buffer()

gmta:
When an ImmutableBitmap is constructed from video data, there's a probability that we're dealing with YUV data and we do not yet have an sk_image to work with.

No regression test since we require a GPU to end up in this situation.

Fixes the background scenery on https://screen.toys/roadtrip/.

CC @Zaggy1024 

| Before | After |
|--|--|
| image | image |

### DIFF
--- a/Libraries/LibGfx/ImmutableBitmap.cpp
+++ b/Libraries/LibGfx/ImmutableBitmap.cpp
@@ -125,6 +125,9 @@ static SkColorType export_format_to_skia_color_type(ExportFormat format)
 
 ErrorOr<BitmapExportResult> ImmutableBitmap::export_to_byte_buffer(ExportFormat format, int flags, Optional<int> target_width, Optional<int> target_height) const
 {
+    if (SkiaBackendContext::the() && !ensure_sk_image(*SkiaBackendContext::the()))
+        return Error::from_string_literal("Failed to create a Skia image for this ImmutableBitmap");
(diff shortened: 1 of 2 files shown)


Libraries/LibGfx/ImmutableBitmap.cpp
```diff
@@ -308,19 +308,23 @@ static sk_sp<SkColorSpace> color_space_from_cicp(Media::CodingIndependentCodePoi
     return SkColorSpace::MakeRGB(transfer_function, gamut);
 }
 
-bool ImmutableBitmap::ensure_sk_image(SkiaBackendContext& context) const
+bool ImmutableBitmap::ensure_sk_image() const
 {
+    auto context = SkiaBackendContext::the();
+    if (!context)
+        return true;
```
Zaggy1024: Shouldn't the return value here depend on whether we actually have an SkImage?

gmta: This was a bit lazy, returning true here was preventing callers from returning early if no backend context was set up. I've dropped that whole commit.

Zaggy1024: I didn't dislike the change overall, but yeah, easier this way. 😅

PR #8925, opened 2026-04-15
LibWeb/CSS: Tokenize some things properly

AtkinsSJ:
Alternative to #8921.

Drafting for now for discussion, and also because this has a smaller number of new WPT passes so something is probably still wrong with a `tokenize()` method somewhere.

### DIFF
--- a/Libraries/LibWeb/CSS/StyleValues/DimensionStyleValue.cpp
+++ b/Libraries/LibWeb/CSS/StyleValues/DimensionStyleValue.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025, Sam Atkins <sam@ladybird.org>
+ * Copyright (c) 2025-2026, Sam Atkins <sam@ladybird.org>
  *
  * SPDX-License-Identifier: BSD-2-Clause
  */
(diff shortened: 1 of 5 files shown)


Libraries/LibWeb/CSS/StyleValues/DimensionStyleValue.cpp
```diff
@@ -11,6 +11,8 @@ namespace Web::CSS {
 
 Vector<Parser::ComponentValue> DimensionStyleValue::tokenize() const
 {
+    if (unit_name() == "percent"sv)
```
shannonbooth: super nit, but maybe `is_percentage()` here instead?

shannonbooth: and actually as an aside I just noticed, the create_dimension call here and just below can be simplified to:

`Parser::Token::create_dimension(raw_value(), unit_name())`

since it's already a fly string

AtkinsSJ: Done! Nice spot

AtkinsSJ:
Looks like the remaining passes aren't present because we don't absolutize the custom property values. I gave that a try, but it relies on information we (currently) don't have at that point during computation. So I'll add that to the "we need to compute custom properties properly" pile instead. :sweat_smile:

PR #8926, opened 2026-04-15
LibWeb/CSS: Port the CSS Tokenizer to Rust

AtkinsSJ:
This is an AI translation of the CSS Tokenizer. `test-css-tokenizer` has been updated to run both Tokenizers and compare the output to make sure it's identical. There isn't yet any support for using it inside LibWeb.

I am very much a Rust novice, and this is very much written by AI. (Codex to be specific.) I have reviewed it myself and as far as I can tell it's correct, and the tests suggest that too, but as I said I am not familiar with Rust. I'm sure the module structure etc is not exactly what it should be.

The code itself is not idiomatic but that's to be expected as a direct translation, and improving that is a future task.

In particular, feedback on the CMake and the FFI parts could use some scrutiny by someone who understands those better.

### DIFF
--- a/Libraries/LibWeb/Rust/src/css_tokenizer.rs
+++ b/Libraries/LibWeb/Rust/src/css_tokenizer.rs
@@ -881,231 +881,232 @@ impl Tokenizer {
         // Consume the next input code point.
         let input = self.consume_code_point();
 
-        // whitespace
-        if is_whitespace(input) {
-            // Consume as much whitespace as possible. Return a <whitespace-token>.
-            self.consume_as_much_whitespace_as_possible();
-            return Token::create_whitespace(start_byte_offset, self.current_byte_offset());
-        }
+        match input {
+            // whitespace
+            c if is_whitespace(c) => {
+                // Consume as much whitespace as possible. Return a <whitespace-token>.
+                self.consume_as_much_whitespace_as_possible();
+                Token::create_whitespace(start_byte_offset, self.current_byte_offset())
+            }
 
-        // U+0022 QUOTATION MARK (")
-        if is_quotation_mark(input) {
-            // Consume a string token and return it.
-            return self.consume_string_token(input);
-        }
+            // U+0022 QUOTATION MARK (")
+            0x22 => {
+                // Consume a string token and return it.
+                self.consume_string_token(input)
+            }
 
-        // U+0023 NUMBER SIGN (#)
-        if is_number_sign(input) {
-            // If the next input code point is an ident code point or the next two input code points
-            // are a valid escape, then:
-            let next_input = self.peek_code_point(0);
-            let maybe_escape = self.peek_twin();
+            // U+0023 NUMBER SIGN (#)
+            0x23 => {
+                // If the next input code point is an ident code point or the next two input code points
+                // are a valid escape, then:
+                if is_ident_code_point(self.peek_code_point(0)) || is_valid_escape_sequence(self.peek_twin()) {
+                    // 1. Create a <hash-token>.
+                    // 2. If the next 3 input code points would start an ident sequence, set the <hash-token>’s
+                    //    type flag to "id".
+                    let hash_type = if would_start_an_ident_sequence(self.peek_triplet()) {
+                        CssHashType::Id
+                    } else {
+                        CssHashType::Unrestricted
+                    };
 
-            if is_ident_code_point(next_input) || is_valid_escape_sequence(maybe_escape) {
-                // 1. Create a <hash-token>.
-                let mut hash_type = CssHashType::Unrestricted;
+                    // 3. Consume an ident sequence, and set the <hash-token>’s value to the returned string.
+                    let value = self.consume_an_ident_sequence();
 
-                // 2. If the next 3 input code points would start an ident sequence, set the <hash-token>’s
-                //    type flag to "id".
-                if would_start_an_ident_sequence(self.peek_triplet()) {
-                    hash_type = CssHashType::Id;
+                    // 4. Return the <hash-token>.
+                    Token::create_hash(value, hash_type, start_byte_offset, self.current_byte_offset())
+                } else {
+                    // Otherwise, return a <delim-token> with its value set to the current input code point.
+                    Token::create_delim(input, start_byte_offset, self.current_byte_offset())
+                }
+            }
+
+            // U+0027 APOSTROPHE (')
+            0x27 => {
+                // Consume a string token and return it.
+                self.consume_string_token(input)
+            }
+
+            // U+0028 LEFT PARENTHESIS (()
+            0x28 => {
+                // Return a <(-token>.
+                Token::create(CssTokenType::OpenParen, start_byte_offset, self.current_byte_offset())
+            }
+
+            // U+0029 RIGHT PARENTHESIS ())
+            0x29 => {
+                // Return a <)-token>.
+                Token::create(CssTokenType::CloseParen, start_byte_offset, self.current_byte_offset())
+            }
+
+            // U+002B PLUS SIGN (+)
+            0x2B => {
+                // If the input stream starts with a number, reconsume the current input code point,
+                // consume a numeric token and return it.
+                if would_start_a_number(self.start_of_input_stream_triplet()) {
+                    self.reconsume_current_input_code_point();
+                    self.consume_a_numeric_token()
+                } else {
+                    // Otherwise, return a <delim-token> with its value set to the current input code point.
+                    Token::create_delim(input, start_byte_offset, self.current_byte_offset())
+                }
+            }
+
+            // U+002C COMMA (,)
+            0x2C => {
+                // Return a <comma-token>.
+                Token::create(CssTokenType::Comma, start_byte_offset, self.current_byte_offset())
+            }
+
+            // U+002D HYPHEN-MINUS (-)
+            0x2D => {
+                // If the input stream starts with a number, reconsume the current input code point,
+                // consume a numeric token, and return it.
+                if would_start_a_number(self.start_of_input_stream_triplet()) {
+                    self.reconsume_current_input_code_point();
+                    return self.consume_a_numeric_token();
                 }
 
-                // 3. Consume an ident sequence, and set the <hash-token>’s value to the returned string.
-                let value = self.consume_an_ident_sequence();
+                // Otherwise, if the next 2 input code points are U+002D HYPHEN-MINUS U+003E
+                // GREATER-THAN SIGN (->), consume them and return a <CDC-token>.
+                let (first, second) = self.peek_twin();
+                if is_hyphen_minus(first) && is_greater_than_sign(second) {
+                    self.consume_code_point();
+                    self.consume_code_point();
+                    return Token::create(CssTokenType::CDC, start_byte_offset, self.current_byte_offset());
+                }
 
-                // 4. Return the <hash-token>.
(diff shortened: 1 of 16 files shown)


jdahlin:
Have you verified speed and/or memory usage just to ensure there is no significant regression?

AtkinsSJ:
Good point. Here's a hyperfine run on my MacBook Pro, running on the UA stylesheet which is >2000 lines.

```
❯ hyperfine -L backend cpp,rust 'Build/release/bin/css-tokenizer --backend {backend} Libraries/LibWeb/CSS/Default.css > /dev/null 2>&1'
Benchmark 1: Build/release/bin/css-tokenizer --backend cpp Libraries/LibWeb/CSS/Default.css > /dev/null 2>&1
  Time (mean ± σ):      39.4 ms ±   0.5 ms    [User: 29.8 ms, System: 8.3 ms]
  Range (min … max):    38.9 ms …  41.7 ms    71 runs
 
Benchmark 2: Build/release/bin/css-tokenizer --backend rust Libraries/LibWeb/CSS/Default.css > /dev/null 2>&1
  Time (mean ± σ):      39.5 ms ±   0.3 ms    [User: 29.8 ms, System: 8.5 ms]
  Range (min … max):    38.9 ms …  41.1 ms    71 runs
 
Summary
  Build/release/bin/css-tokenizer --backend cpp Libraries/LibWeb/CSS/Default.css > /dev/null 2>&1 ran
    1.00 ± 0.02 times faster than Build/release/bin/css-tokenizer --backend rust Libraries/LibWeb/CSS/Default.css > /dev/null 2>&1
```
After doing this a few times, the Rust one is consistently 2-3% slower. So, not significantly different, though it is a regression. I'll keep an eye on it when working on it later.

I can't seem to get a memory reading, do you have any tips for that?

nico:
> I can't seem to get a memory reading, do you have any tips for that?

`/usr/bin/time -v` gives you a max rss reading. Maybe that together with a representative input could give you an idea?

AtkinsSJ:
Thanks, I couldn't get that to work before but turns out macos has a gnu-time package, so it's all good now. 😅 

C++:
```
	Command being timed: "Build/release/bin/css-tokenizer --backend cpp Libraries/LibWeb/CSS/Default.css"
	User time (seconds): 0.04
	System time (seconds): 0.02
	Percent of CPU this job got: 78%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:00.08
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 36048
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 34
	Minor (reclaiming a frame) page faults: 2930
	Voluntary context switches: 567
	Involuntary context switches: 14
	Swaps: 0
	File system inputs: 0
	File system outputs: 0
	Socket messages sent: 0
	Socket messages received: 0
	Signals delivered: 0
	Page size (bytes): 16384
	Exit status: 0
```

Rust:
```
	Command being timed: "Build/release/bin/css-tokenizer --backend rust Libraries/LibWeb/CSS/Default.css"
	User time (seconds): 0.04
	System time (seconds): 0.02
	Percent of CPU this job got: 77%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:00.08
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 40192
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 33
	Minor (reclaiming a frame) page faults: 3190
	Voluntary context switches: 654
	Involuntary context switches: 24
	Swaps: 0
	File system inputs: 0
	File system outputs: 0
	Socket messages sent: 0
	Socket messages received: 0
	Signals delivered: 0
	Page size (bytes): 16384
	Exit status: 0
```

So that's... 11% more memory used by my reckoning?

nicoburns:
I'm assuming that Ladybird wouldn't consider using https://github.com/servo/rust-cssparser (the CSS tokenizer library used by Servo and Firefox)? Posting it here anyway, just to make sure people are aware it exists.

AtkinsSJ:
> I'm assuming that Ladybird wouldn't consider using [servo/rust-cssparser](https://github.com/servo/rust-cssparser?rgh-link-date=2026-04-17T11%3A17%3A28Z) (the CSS tokenizer library used by Servo and Firefox)? Posting it here anyyway, just to make sure people are aware it exists.

No, this is something we should be responsible for ourselves.

R-Goc:
In view that it is slower and uses more memory is there any reason to use the Rust implemenation other than it being potentially more memory safe? Also while I'm not a Rust dev, this code looks rather unidiomatic with a heavy use of if statements instead of pattern matching.

AtkinsSJ:
> In view that it is slower and uses more memory is there any reason to use the Rust implemenation other than it being potentially more memory safe?

Yup, looks like I completely forgot to actually explain the motivation here! Eventually the entire Parser will want to be Rust, and this is an easy-to-separate first step for that. (A full Rust Parser requires splitting our CSSOM types from our internal/parsed CSS objects, and that's a big task on its own.) I'm hoping that a lot of the performance reduction is overhead from converting over to C++ types at the end, but also this is not the final form of this code.

> Also while I'm not a Rust dev, this code looks rather unidiomatic with a heavy use of if statements instead of pattern matching.

Yeah, it absolutely isn't idiomatic, it's a fairly direct translation of the C++ code. As a follow-up I want to go through it and :crab:-ify everything, which will probably help a little with the performance. I'm in the process of learning Rust properly so hopefully I can improve it soon.

R-Goc:
>  I'm hoping that a lot of the performance reduction is overhead from converting over to C++ types at the end, but also this is not the final form of this code.

A profile should quickly tell you that. The perf difference might just be from more allocations.

nicoburns:
Firefox has shared Rust/C++ FFI types for a bunch of types which helps with FFI overhead (by *not* doing any conversions). In particular, the `thin-vec` crate is compatible with Firefox's C++ vector type, and the `servo_arc` crate serves as an FFI friendly atomically reference counted pointer type. And I think Stylo internally has types for boxed slices and a couple of other cases.

(which accounts for most of the allocated types which tend to be the expensive ones to convert)

alimpfard:
I'm gonna repeat myself and say that imo we should not mix the languages in one lib, if the parser wants to be a rust thing, it should be a separate program that talks over ipc instead.
C APIs are evil, and make it so I trust the final rust/C/C++ mixed result even less than I trust the C++ version.

AtkinsSJ:
Felt like making this a bit more idiomatic, so I've done so.

I've re-benchmarked it and it's waaaaay closer. Though this is on Linux instead of macOS, which honestly I think is probably most to blame there.

Hyperfine:
```
❯ hyperfine -L backend cpp,rust 'Build/release/bin/css-tokenizer --backend {backend} Libraries/LibWeb/CSS/Default.css > /dev/null 2>&1'
Benchmark 1: Build/release/bin/css-tokenizer --backend cpp Libraries/LibWeb/CSS/Default.css > /dev/null 2>&1
  Time (mean ± σ):      33.5 ms ±   0.7 ms    [User: 21.2 ms, System: 12.0 ms]
  Range (min … max):    32.2 ms …  35.2 ms    82 runs
 
Benchmark 2: Build/release/bin/css-tokenizer --backend rust Libraries/LibWeb/CSS/Default.css > /dev/null 2>&1
  Time (mean ± σ):      33.6 ms ±   0.8 ms    [User: 21.0 ms, System: 12.4 ms]
  Range (min … max):    32.2 ms …  37.0 ms    85 runs
 
Summary
  Build/release/bin/css-tokenizer --backend cpp Libraries/LibWeb/CSS/Default.css > /dev/null 2>&1 ran
    1.01 ± 0.03 times faster than Build/release/bin/css-tokenizer --backend rust Libraries/LibWeb/CSS/Default.css > /dev/null 2>&1
```

C++ `time`:
```
Command being timed: "Build/release/bin/css-tokenizer --backend cpp Libraries/LibWeb/CSS/Default.css"
User time (seconds): 0.02
System time (seconds): 0.04
Percent of CPU this job got: 37%
Elapsed (wall clock) time (h:mm:ss or m:ss): 0:00.19
Average shared text size (kbytes): 0
Average unshared data size (kbytes): 0
Average stack size (kbytes): 0
Average total size (kbytes): 0
Maximum resident set size (kbytes): 65960
Average resident set size (kbytes): 0
Major (requiring I/O) page faults: 0
Minor (reclaiming a frame) page faults: 2729
Voluntary context switches: 272
Involuntary context switches: 1
Swaps: 0
File system inputs: 0
File system outputs: 0
Socket messages sent: 0
Socket messages received: 0
Signals delivered: 0
Page size (bytes): 4096
Exit status: 0
```

Rust `time`:
```
Command being timed: "Build/release/bin/css-tokenizer --backend rust Libraries/LibWeb/CSS/Default.css"
User time (seconds): 0.03
System time (seconds): 0.04
Percent of CPU this job got: 46%
Elapsed (wall clock) time (h:mm:ss or m:ss): 0:00.15
Average shared text size (kbytes): 0
Average unshared data size (kbytes): 0
Average stack size (kbytes): 0
Average total size (kbytes): 0
Maximum resident set size (kbytes): 67964
Average resident set size (kbytes): 0
Major (requiring I/O) page faults: 0
Minor (reclaiming a frame) page faults: 2734
Voluntary context switches: 272
Involuntary context switches: 8
Swaps: 0
File system inputs: 0
File system outputs: 0
Socket messages sent: 0
Socket messages received: 0
Signals delivered: 0
Page size (bytes): 4096
Exit status: 0
```

So overall that makes it about 1% slower and 3% more memory usage, both of which I think are within the realms of not mattering.

Libraries/LibWeb/Rust/src/css_tokenizer.rs
```diff
@@ -0,0 +1,1385 @@
...
+
+    fn consume_code_point(&mut self) -> u32 {
+        if self.index >= self.code_points.len() {
+            return TOKENIZER_EOF;
+        }
+
+        self.prev_index = self.index;
+        self.prev_position = self.position;
+
+        let (_, code_point) = self.code_points[self.index];
```
nicoburns, line 300: A `get_unchecked` here would be an obvious optimization to try. And would be safe due to the early return at the top of the function.

Libraries/LibWeb/Rust/src/css_tokenizer.rs
```diff
@@ -0,0 +1,1385 @@
...
+    } else {
+        (bytes.as_ptr(), bytes.len())
+    }
+}
+
+fn append_code_point(builder: &mut String, code_point: u32) {
+    builder.push(char::from_u32(code_point).unwrap_or(char::REPLACEMENT_CHARACTER));
+}
+
+fn is_eof(code_point: u32) -> bool {
```
nicoburns, line 1160: `#[inline(always)]` on these small functions may also be worth trying. You'd expect the compiler to do this automatically, but it doesn't always in my experience.

awesomekling:
Quickly profiling this, we can see that `to_debug_string()` is >90% of execution time for both backends. So it's pretty hard to get a meaningful read on performance.

If you run both without dumping the tokens, they are roughly neck-and-neck on real-world content, with the C++ tokenizer coming out slightly on top. Not enough to worry about here though, and looks trivial to improve, so let's get this into the tree so we can move forward. :)

PR #8930, opened 2026-04-15
LibWeb: Guard against removed fragments in LayoutState::commit()

lukasxcs:
When a line contains text, followed by an atomic inline (such as a verified badge icon), followed by more text, all inside an overflowing container with `text-overflow: ellipsis`, applying the ellipsis to the first piece of text causes everything after it to be removed.
When the badge is removed this way, its recorded index is left dangling. The next layout commit then attempts to dereference that stale index and crashes.

This was reproducible on YouTube pages, where channel metadata lines can contain a channel name, a verified checkmark badge, and additional text (e.g. a collaborating channel name) all on the same overflowing line.
[Example](https://www.youtube.com/watch?v=iTD4lWuEKrE&list=RDCLAK5uy_kj3rhiar1LINmyDcuFnXihEO0K1NQa2jI&index=2)

The fix adds a bounds check before dereferencing the fragment index, falling back gracefully if the fragment is no longer present.

### DIFF
--- a/Libraries/LibWeb/Layout/InlineFormattingContext.cpp
+++ b/Libraries/LibWeb/Layout/InlineFormattingContext.cpp
@@ -331,8 +331,22 @@ void InlineFormattingContext::apply_text_overflow_ellipsis(Vector<LineBox>& line
 
             fragment.set_inline_length(CSSPixels::nearest_value_for(last_kept_end + ellipsis_width));
 
-            if (i + 1 < fragments.size())
+            if (i + 1 < fragments.size()) {
+                for (size_t j = i + 1; j < fragments.size(); ++j) {
+                    if (auto& removed_fragment = fragments[j]; removed_fragment.is_atomic_inline()) {
+                        const auto& box = as<Box>(removed_fragment.layout_node());
+                        auto& box_state = m_state.get_mutable(box);
+                        box_state.containing_line_box_fragment.clear();
+                        box_state.set_elided(true);
+                        box.for_each_in_subtree_of_type<Box>([&](Box const& child_box) {
+                            if (auto* child_state = m_state.try_get_mutable(child_box))
+                                child_state->set_elided(true);
+                            return TraversalDecision::Continue;
+                        });
+                    }
+                }
                 fragments.remove(i + 1, fragments.size() - i - 1);
+            }
 
             line_box.m_inline_length = available_width;
             break;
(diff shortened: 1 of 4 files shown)


awesomekling:
I don’t think this is quite the right fix.

The stale `containing_line_box_fragment` is introduced when `apply_text_overflow_ellipsis()` removes trailing fragments from the line box.

`LayoutState::commit()` is only the consumer of that bookkeeping, so adding a bounds check there avoids the crash, but it doesn’t really restore the invariant.

In particular, falling back to `used_values.offset` looks wrong for atomic inlines. The final paint position for an atomic inline comes from its line-box fragment, not from `used_values.offset`, and we only copy fragment offsets back into box state for surviving inline-block fragments before `update_last_line()`.

For replaced inlines like the `` in the new test, that fallback can easily be stale or `{0, 0}`, which means we may stop crashing but still paint a fragment that was supposed to be elided.

I think the real fix should happen at the point where ellipsis truncation removes fragments:

- update or invalidate `containing_line_box_fragment` for any removed atomic inline fragments
- maybe also record that they were elided, so `commit()`/painting won’t try to place them at all

Also, using `offset.is_zero()` as the “fragment was missing” signal seems fragile, since `{0, 0}` is a valid fragment position.

lukasxcs:
Accidentally closed by force pushing without a commit

awesomekling:
Hmm, looking at this version, I think there are still two issues:

- Since we reuse paintables across relayouts, we have to make sure the `m_elided` flag is cleared when reusing a paintable. Otherwise a box that was elided once can stay hidden even after a later relayout where it should appear.
- If paintables are in the elided state, they should probably be ignored by painting and hit testing in various places. This is not something super centralized which gets messy..

I'm wondering if a simpler solution would be omitting elided boxes from the paint tree entirely.

The stale bookkeeping absolutely seems to belong in `apply_text_overflow_ellipsis()`, so clearing `containing_line_box_fragment` there makes sense. But once we know the atomic inline was elided by ellipsis, it feels more natural to keep that as layout-state-only and have `LayoutState::commit()` skip attaching/positioning a paintable for it entirely.

lukasxcs:
Oh i found a crash with this. I`ll look into that

lukasxcs:
Got it. I just needed to elide the descendant boxes aswell. Updated the test to reflect this.

lukasxcs:
Actually, i found a pr thats fixing exactly this issue.. #8718

awesomekling:
> Actually, i found a pr thats fixing exactly this issue.. #8718

Oh whoops! Okay, merged that one since it was in good shape already. Thanks for working on Ladybird :)

PR #8931, opened 2026-04-15
Everywhere: Remove GN build system

gmta:
We've had some impressive efforts going into setting up and maintaining GN as an alternative to CMake, but as of today we're still using CMake and the GN files have bitrotted quite severely.

We've found no convincing reason to keep maintaining two build systems in parallel - so let's remove GN and make life a bit simpler for us.


Documentation/BuildInstructionsLadybird.md
```diff
@@ -211,7 +211,7 @@ Or, download a version of Gradle >= 8.0.0, and run the ``gradlew`` program in ``
 ### FreeBSD
 
 ```
-pkg install autoconf-archive automake autoconf bash cmake curl gmake gn libdrm libtool libxcb libxkbcommon libX11 libXrender libXi nasm ninja patchelf pkgconf python3 qt6-base unzip zip
+pkg install autoconf-archive automake autoconf bash cmake curl gmake libdrm libtool libxcb libxkbcommon libX11 libXrender libXi nasm ninja patchelf pkgconf python3 qt6-base unzip zip
```
trflynn89: This might still be needed (perhaps for a vcpkg dep)? It looks like it was added unrelated to the gn build
9d26caf0d55f86b45e905bb416e57f52f2a1309d

gmta: I've found no evidence that any vcpkg port needs gn, and I assumed this was added because the author of that commit also tried building with GN on their machine. CC @cqundefine

shannonbooth: FWIW to add to this I have not have not had GN installed and I ran into no build issues  (both MacOS and Linux). I think I installed last week on macOS since I happened to run into the linter thing asked me to install GN, which I did (and then ended up not upstreaming the GN changes I had sitting locally anyhow)

trflynn89: On most systems, GN gets built by vcpkg when it is needed (it is used by the skia port). For example, on my mac, it is here:
```bash
./Build/vcpkg/packages/vcpkg-tool-gn_arm64-osx-dynamic/tools/gn/gn --version
2274 (cc7c61e105fb)
```

I didn't see our own GN build referenced at all in the PR that added the FreeBSD build instructions, which is why I'm wondering if vcpkg doesn't build GN itself on FreeBSD for some reason.

cqundefine: Vcpkg predownloads GN on mainstream platforms (Windows, Linux, OS X), but it doesn't have a download on FreeBSD so it just looks for it in the system path. I'm pretty sure the Skia port just calls out to GN and because vcpkg can't get GN on it's own it is needed in the list.

gmta: > but it doesn't have a download on FreeBSD so it just looks for it in the system path.

Right! Reverted the change in the documentation.

trflynn89 (approved):

ADKaster:
Technically the toolchain script is useful on Asahi Linux -- chromium's CDN doesn't (didn't?) provide 16k Page-compatible binaries. Not sure if that's changed with chromium's official aarch64 Linux support recently.

PR #8932, opened 2026-04-15
Tests/LibWeb: Improve flaky iframe srcdoc crash test

gmta:
This test was introduced in b64df59cc61f31b8a0aab91abb4630b3ecb1dcee but had a race condition where `contentDocument.body` would not yet be available. Switch to an async test that listens to the iframe's `load` event before continuing. Confirmed that this still crashes without the original fix.

Flakes on master seen here:
* https://github.com/LadybirdBrowser/ladybird/actions/runs/24462307871/job/71479645333
* https://github.com/LadybirdBrowser/ladybird/actions/runs/24394400370/job/71248421772
* https://github.com/LadybirdBrowser/ladybird/actions/runs/24310880624/job/70980181420

### DIFF
--- a/Libraries/LibWeb/DOM/Document.h
+++ b/Libraries/LibWeb/DOM/Document.h
@@ -806,6 +806,7 @@ public:
     bool ready_to_run_scripts() const { return m_ready_to_run_scripts; }
     void set_ready_to_run_scripts();
     void set_deferred_parser_start(GC::Ref<GC::Function<void()>>);
+    bool has_deferred_parser_start() const { return m_deferred_parser_start; }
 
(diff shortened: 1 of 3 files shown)


awesomekling (approved):

gmta:
Relevant comment by @kalenikaliaksandr :

https://github.com/LadybirdBrowser/ladybird/pull/8852#pullrequestreview-4085703675

> we should figure out why contentDocument.body isn't available by the time the document becomes observable to the parent.

gmta:
Closing this; I believe something closer to #8852 is going to be a better approach.

PR #8933, opened 2026-04-15
LibWebView+UI: Add browsing history and rich address bar autocomplete

awesomekling:
This branch adds a SQLite-backed HistoryStore in LibWebView that persists visited pages along with their title, favicon, visit count, and last-visit time. WebContentClient now records navigations, title changes, and favicon updates into the store, and a "Clear browsing history" action is wired up in the Privacy settings page.
                                                                                                                                            
The address bar autocomplete pipeline has been reworked so that suggestions are structured rows, history hits, a "Search with" entry, and plain completions, each carrying a title, URL, and subtitle rather than a flat string.

Both the Qt and AppKit UIs render the new rich rows with favicons and subtitles.
                                                                  
This is all pretty simple and barebones! There's plenty more data we could track and plenty more UI we could build on top of it, but we gotta start somewhere. :)

HF0z3h1XwAAb0GF

Screenshot from 2026-04-16 00-45-44


Libraries/LibWebView/HistoryStore.cpp
```diff
@@ -0,0 +1,363 @@
...
+            return left->visit_count > right->visit_count;
+
+        if (left->last_visited_time != right->last_visited_time)
+            return left->last_visited_time > right->last_visited_time;
+
+        return left->url < right->url;
+    });
+}
+
+static String title_value(Optional<String> const& title)
```
trflynn89: With one user, making this a function seems a bit overkill 😅

Libraries/LibWebView/HistoryStore.cpp
```diff
@@ -0,0 +1,363 @@
...
+        [&](auto statement_id) {
+            suggestions.append(database.result_column<String>(statement_id, 0));
+        },
+        MUST(String::from_utf8(query)),
+        MUST(String::from_utf8(query)),
+        MUST(String::from_utf8(query)),
+        MUST(String::from_utf8(query)),
+        MUST(String::from_utf8(query)),
+        MUST(String::from_utf8(query)),
+        static_cast<i64>(limit));
```
trflynn89: That's a bit goofy! I do believe you can give sqlite numbered placeholders in the prepared statement so you only have to provide `query` once. (Or at least only allocate the `String` once)

(I've wanted to make `execute_statement` work with StringView in the past, but then we lose information on whether the string was a `String` or `ByteString` view, and we need to know that to know whether the SQL type is `TEXT` or `BLOB`)

trflynn89: Looking at the one caller at the end result of the PR, it seems like `autocomplete_suggestions` is totally unused and `autocomplete_entries` has one caller, which already has a trimmed `String`. So it seems like we could remove `autocomplete_suggestions` and pass the `String` all the way down?

Libraries/LibWebView/HistoryStore.h
```diff
@@ -0,0 +1,93 @@
...
+    struct Statements {
+        Database::StatementID upsert_entry { 0 };
+        Database::StatementID update_title { 0 };
+        Database::StatementID get_entry { 0 };
+        Database::StatementID search_entries { 0 };
+        Database::StatementID clear_entries { 0 };
+        Database::StatementID delete_entries_accessed_since { 0 };
+    };
+
+    class TransientStorage {
```
trflynn89, line 61: I wonder if we actually want a transient storage? We disable SQL for e.g. WPT so we can have multiple instances of the browser running, so cookies and localStorage need an in-memory implementation. But history is more of a user thing, so we can probably just disable it outright instead if SQL is disabled?

Libraries/LibWebView/HistoryDebug.h
```diff
@@ -0,0 +1,27 @@
...
+#include <AK/ByteString.h>
+#include <AK/String.h>
+#include <AK/Vector.h>
+#include <LibCore/Environment.h>
+
+namespace WebView {
+
+inline bool history_debug_enabled()
+{
+    static bool enabled = Core::Environment::has("LADYBIRD_TRACE_HISTORY"sv);
```
trflynn89: I won't argue too hard, but we use `dbgln_if` for this everywhere else.

Libraries/LibWebView/HistoryStore.cpp
```diff
@@ -158,6 +255,13 @@ void HistoryStore::record_visit(URL::URL const& url, Optional<String> title, Uni
     if (!normalized_url.has_value())
         return;
 
+    if (history_debug_enabled())
```
trflynn89: This guy should have braces
(but wouldn't be an issue with `dbgln_if`)

Libraries/LibWebView/HistoryStore.cpp
```diff
@@ -166,13 +270,22 @@ void HistoryStore::record_visit(URL::URL const& url, Optional<String> title, Uni
...
+        if (history_debug_enabled())
+            dbgln("[History] Ignoring empty history title update for {}", url);
         return;
+    }
 
     auto normalized_url = normalize_url(url);
     if (!normalized_url.has_value())
         return;
 
+    if (history_debug_enabled())
```
trflynn89: Same re: braces, and other places below

Libraries/LibWebView/WebContentClient.cpp
```diff
@@ -18,6 +21,21 @@ namespace WebView {
...
 
+static Optional<String> history_title(Utf16String const& title, URL::URL const& url)
+{
+    if (title.is_empty())
+        return {};
+
+    auto title_utf8 = title.to_utf8();
+    if (title_utf8.is_empty())
+        return {};
+
```
trflynn89: Surely `title_utf8` isn't empty if `title` wasnt?

Libraries/LibWebView/HistoryStore.cpp
```diff
@@ -16,6 +16,8 @@ namespace WebView {
 
 static constexpr auto DEFAULT_AUTOCOMPLETE_SUGGESTION_LIMIT = 8uz;
 static constexpr size_t MINIMUM_TITLE_AUTOCOMPLETE_QUERY_LENGTH = 3;
+static constexpr u32 HISTORY_VERSION = 2u;
```
trflynn89: Not a big deal, but it feels a little bit strange to go through an upgrade path on the PR that is introducing the database.

Libraries/LibWebView/WebContentClient.cpp
```diff
@@ -36,6 +36,18 @@ static Optional<String> history_title(Utf16String const& title, URL::URL const&
     return title_utf8;
 }
 
+static Optional<String> normalized_history_url(URL::URL const& url)
```
trflynn89: Isn't this the same thing as the normalizer in HistoryStore? Let's just make it a publicly accessible method in HistoryStore.h?

UI/AppKit/Interface/Autocomplete.h
```diff
@@ -17,7 +17,7 @@
 
 @end
 
-@interface Autocomplete : NSPopover
+@interface Autocomplete : NSObject
```
trflynn89, line 25: Sounds like this fixes #4188?

(It's always been very strange that only some people have encountered that)

UI/AppKit/Interface/Autocomplete.mm
```diff
@@ -4,13 +4,186 @@
...
+{
+    static NSFont* font;
+    static dispatch_once_t token;
+    dispatch_once(&token, ^{
+        font = [NSFont systemFontOfSize:[NSFont smallSystemFontSize]];
+    });
+    return font;
+}
+
+static NSImage* image_from_base64_png(StringView favicon_base64_png)
```
trflynn89: Let's also just make this an accessible helper instead of copy-pasting it between here and Menu.mm
(I'm fine if we want to do this particular helper in a follow-up)

Libraries/LibWebView/Autocomplete.cpp
```diff
@@ -44,57 +45,217 @@ Optional<AutocompleteEngine const&> find_autocomplete_engine_by_name(StringView
...
+    if (literal_url_suggestion.has_value())
+        append_suggestion_if_unique(suggestions, max_suggestions, literal_url_suggestion.release_value());
+
+    for (auto& suggestion : history_suggestions)
+        append_suggestion_if_unique(suggestions, max_suggestions, move(suggestion));
 
     for (auto& suggestion : remote_suggestions) {
-        if (!history_suggestions.contains_slow(suggestion))
-            history_suggestions.unchecked_append(move(suggestion));
+        append_suggestion_if_unique(suggestions, max_suggestions, {
```
trflynn89: If you manually add a newline before the `{`, it avoids this unfortunate rightward drift

Libraries/LibWebView/URL.cpp
```diff
@@ -72,6 +73,86 @@ Optional<URL::URL> sanitize_url(StringView location, Optional<SearchEngine> cons
     return url;
 }
 
+bool location_looks_like_url(StringView location, AppendTLD append_tld)
+{
+    return sanitize_url(location, {}, append_tld).has_value();
+}
+
+static String normalized_web_url_for_autocomplete_comparison(URL::URL const& url)
```
trflynn89, line 82: It's hard to tell at a glance what this does differently than `serialize(URL::ExcludeFragment::Yes)`. Is it just removing `www.`? (A comment would help)

If so, could we then just serialize a URL copy that has removed the `www.` from the host of the input URL?

i.e. (basically pseudocode here):
```c++
if (auto host = url.serialized_host(); host.starts_with("www.")) {
    auto url_copy = url;
    url_copy.set_host(host.substring(4));
    return url_copy.serialize(URL::ExcludeFragment::Yes);
}

return url.serialize(URL::ExcludeFragment::Yes);
```

Libraries/LibWebView/URL.h
```diff
@@ -19,6 +19,9 @@ enum class AppendTLD {
     Yes,
 };
 WEBVIEW_API Optional<URL::URL> sanitize_url(StringView, Optional<SearchEngine> const& search_engine = {}, AppendTLD = AppendTLD::No);
+WEBVIEW_API bool location_looks_like_url(StringView, AppendTLD = AppendTLD::No);
+WEBVIEW_API bool autocomplete_urls_match(StringView left, StringView right);
+WEBVIEW_API bool autocomplete_url_can_complete(StringView query, StringView suggestion);
```
trflynn89: Autocomplete.h feels like a better home for the autocomplete_* methods, but not a big deal

UI/Icons/search.tvgt
```diff
@@ -0,0 +1,24 @@
+(tvg 1
```
trflynn89: Is this file needed? (I don't see other .tvgt files in repo. If it helps for debugging or something that's fine)

trflynn89 (commented):
Super cool :)

I went through this a bit quickly. Some of it kinda hard to review where things are added to the history store in one commit and then removed or reshaped in the next commit, but 🤷‍♂️ 

Thinking out loud, I'm wondering if our monolith Ladybird.db makes sense long term, or if we want to split it out into Cookies.db, History.db, etc.? (This can be decided after merging)

jdahlin:
A couple of things related SQLite and comparison with other browsers:
1. Favicon storage: Favicons are base64 TEXT inline in the History table. Firefox/Chrome store BLOBs in a separate database. Alternatively, don't SELECT favicon in the search query, instead fetch lazily after ranking.     
2. Autocomplete full table scan: search_entries computes searchable_url via nested SUBSTR/INSTR/LIKE per row. No index can help here, instead a stored+indexed searchable_url column would turn this into an index range scan.
3. No frecency: Ranking by visit_count DESC means a page visited 100 times six months ago beats one visited 3 times today. Firefox/Chrome use a recency-weighted score as the primary sort key.                     
4. Per-statement auto-commit: A page load triggers up to 3 separate fsyncs (record_visit, update_title, update_favicon). Wrapping per-navigation writes in a single BEGIN/COMMIT would avoid that.             

Not suggesting all of these should be fixed in this PR, but the current implementation will be too slow at some point when the history grows.

Libraries/LibWebView/HistoryStore.h
awesomekling: Good point. History is now disabled outright when SQL is disabled; the remaining transient backend is only used by TestHistoryStore.

awesomekling:
> Thinking out loud, I'm wondering if our monolith Ladybird.db makes sense long term, or if we want to split it out into Cookies.db, History.db, etc.? (This can be decided after merging)

Moved this to History.db right away. It just makes sense :)

awesomekling:
> A couple of things related SQLite and comparison with other browsers:
> 
>     1. Favicon storage: Favicons are base64 TEXT inline in the History table. Firefox/Chrome store BLOBs in a separate database. Alternatively, don't SELECT favicon in the search query, instead fetch lazily after ranking.
> 
>     2. Autocomplete full table scan: search_entries computes searchable_url via nested SUBSTR/INSTR/LIKE per row. No index can help here, instead a stored+indexed searchable_url column would turn this into an index range scan.
> 
>     3. No frecency: Ranking by visit_count DESC means a page visited 100 times six months ago beats one visited 3 times today. Firefox/Chrome use a recency-weighted score as the primary sort key.
> 
>     4. Per-statement auto-commit: A page load triggers up to 3 separate fsyncs (record_visit, update_title, update_favicon). Wrapping per-navigation writes in a single BEGIN/COMMIT would avoid that.
> 
> 
> Not suggesting all of these should be fixed in this PR, but the current implementation will be too slow at some point when the history grows.

All good points that we can and should address! ... but let's do it separately :sweat_smile:

PR #8935, opened 2026-04-15
LibWeb: Make Node::is_connected() O(1) via a cached flag

awesomekling:
Previously this walked up the parent chain on every call, which shows up as a 2.5% item in the profile while watching YouTube videos.

Cache an `m_is_connected` bit on `Node` instead, maintained by the DOM insertion and removal steps.


shannonbooth:
how about moving a node to a different tree?

awesomekling:
> how about moving a node to a different tree?

That would still fire the removed/inserted mechanisms, no?

PR #8938, opened 2026-04-16
LibWebView: Add bookmark import/export to about:bookmarks

mikiubo:
Implement bookmark import/export in about:bookmarks using Netscape bookmark HTML in JavaScript.
Import parsed items into BookmarkStore under an "Imported Bookmarks" folder, and treat internal WebUI about: pages as potentially trustworthy so SecureContext APIs are available there.

### DIFF
--- a/Base/res/ladybird/about-pages/bookmarks.html
+++ b/Base/res/ladybird/about-pages/bookmarks.html
@@ -11,6 +11,9 @@
                     --tree-selected-color: #d8d8f8;
                     --empty-text-color: #888;
                     --url-text-color: #666;
+                    --menu-background-color: #fff;
+                    --menu-border-color: #d8d8d8;
+                    --menu-shadow-color: rgba(0, 0, 0, 0.12);
                 }
             }
 
@@ -20,7 +23,80 @@
                     --tree-selected-color: #35355a;
                     --empty-text-color: #777;
                     --url-text-color: #999;
+                    --menu-background-color: #2a2a2a;
+                    --menu-border-color: #4a4a4a;
+                    --menu-shadow-color: rgba(0, 0, 0, 0.4);
                 }
+            }
+
+            .header-menu {
+                position: relative;
+                margin-left: auto;
+            }
+
+            .menu-toggle {
+                display: flex;
+                align-items: center;
+                justify-content: center;
+
+                width: 32px;
+                height: 32px;
+
+                color: inherit;
+                background: none;
+                border: none;
+                border-radius: 4px;
+
+                cursor: pointer;
+            }
+
+            .menu-toggle:hover {
+                background-color: var(--tree-hover-color);
+            }
+
+            .menu-toggle svg {
+                width: 20px;
+                height: 20px;
+            }
+
+            .menu-dropdown {
+                position: absolute;
+                top: 100%;
+                right: 0;
+                z-index: 10;
+
+                display: flex;
+                flex-direction: column;
+
+                min-width: 200px;
+                padding: 4px;
+                margin-top: 4px;
+
+                background-color: var(--menu-background-color);
+                border: 1px solid var(--menu-border-color);
+                border-radius: 6px;
+                box-shadow: 0 4px 12px var(--menu-shadow-color);
+            }
+
+            .menu-dropdown.hidden {
+                display: none;
+            }
+
+            .menu-item {
+                color: inherit;
+                background: none;
+                border: none;
+                border-radius: 4px;
+
+                padding: 8px 12px;
+                font-size: 14px;
+                text-align: left;
+
+                cursor: pointer;
+            }
+
+            .menu-item:hover {
+                background-color: var(--tree-hover-color);
             }
 
             .tree-empty {
@@ -158,6 +234,20 @@
                 <img src="resource://icons/128x128/app-browser-dark.png" />
             </picture>
             <h1>Bookmarks</h1>
+                <div class="header-menu">
+                    <button class="menu-toggle" id="menu-toggle" aria-label="Bookmark options">
+                        <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="currentColor" aria-hidden="true">
+                            <circle cx="12" cy="5" r="2" />
+                            <circle cx="12" cy="12" r="2" />
+                            <circle cx="12" cy="19" r="2" />
+                        </svg>
+                    </button>
+                    <div class="menu-dropdown hidden" id="menu-dropdown">
+                        <button class="menu-item" id="import-button">Import from HTML...</button>
+                        <button class="menu-item" id="export-button">Export to HTML...</button>
+                    </div>
+                </div>
+                <input type="file" id="import-file" accept=".html,text/html" hidden />
         </header>
 
         <div class="card">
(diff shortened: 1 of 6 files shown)


Base/res/ladybird/about-pages/bookmarks.html
```diff
@@ -158,6 +164,11 @@
                 <img src="resource://icons/128x128/app-browser-dark.png" />
             </picture>
             <h1>Bookmarks</h1>
+
+            <div class="header-actions">
+                <button id="import-bookmarks" class="secondary-button">Import...</button>
```
trflynn89, line 169: The look of these buttons here is...not great

Image



A hamburger menu feels appropriate here? I had prototyped that at one point if you want to yoink this:

https://github.com/trflynn89/ladybird/commit/1c210c7453597f62837a4d01aae41a70f9d8c4ae#diff-ad88985b6b54cbec068bc7e57bb49cd4a94e6230123775e5d6cce6b766d67547R236-R248

Base/res/ladybird/about-pages/bookmarks.html
```diff
@@ -463,6 +474,169 @@ <h1>Bookmarks</h1>
...
+
+                    for (const dt of dl.children) {
+                        if (dt.tagName !== "DT") continue;
+
+                        const anchor = dt.querySelector(":scope > A");
+                        if (anchor) {
+                            const item = {
+                                id: crypto.randomUUID(),
+                                type: "bookmark",
+                                url: anchor.getAttribute("HREF") || anchor.getAttribute("href") || "",
```
trflynn89, line 495: I'm pretty sure `getAttribute` is case-insensitive for HTML documents?

Base/res/ladybird/about-pages/bookmarks.html
```diff
@@ -463,6 +474,169 @@ <h1>Bookmarks</h1>
...
             });
+
+            // Import: parse Netscape Bookmark HTML format using DOMParser
+            function parseNetscapeBookmarks(html) {
+                const parser = new DOMParser();
+                const doc = parser.parseFromString(html, "text/html");
+
+                function parseItems(dl) {
+                    const items = [];
+                    if (!dl) return items;
```
trflynn89, line 485: nit, but I really don't love return statements inline with if statements. I know `prettier` does this automatically, but we can avoid it by added braces.

Base/res/ladybird/about-pages/bookmarks.html
```diff
@@ -463,6 +474,169 @@ <h1>Bookmarks</h1>
...
+                            const item = {
+                                id: crypto.randomUUID(),
+                                type: "bookmark",
+                                url: anchor.getAttribute("HREF") || anchor.getAttribute("href") || "",
+                                title: anchor.textContent.trim() || null,
+                            };
+
+                            const icon = anchor.getAttribute("ICON") || anchor.getAttribute("icon");
+                            if (icon) {
+                                const match = icon.match(/^data:image\/[^;]+;base64,(.+)$/);
```
trflynn89, line 501: Our bookmarks backend only supports PNG right now, so if we encounter image/jpg, we're going to fail to decode this.

Base/res/ladybird/about-pages/bookmarks.html
```diff
@@ -463,6 +474,169 @@ <h1>Bookmarks</h1>
...
+                input.click();
+            });
+
+            // Export: generate Netscape Bookmark HTML format from current bookmarks
+            function escapeHtml(text) {
+                return text
+                    .replace(/&/g, "&amp;")
+                    .replace(/</g, "&lt;")
+                    .replace(/>/g, "&gt;")
+                    .replace(/"/g, "&quot;");
```
trflynn89, line 584: Can we do this as a single replacement?
```js
return text.replace(/[&<>"']/g, c => `&#${c.charCodeAt(0)}`);
```

Base/res/ladybird/about-pages/bookmarks.html
```diff
@@ -463,6 +474,169 @@ <h1>Bookmarks</h1>
...
+
+                html += `${pad}</DL><p>\n`;
+                return html;
+            }
+
+            document.getElementById("export-bookmarks").addEventListener("click", () => {
+                let html = "<!DOCTYPE NETSCAPE-Bookmark-file-1>\n";
+                html += "<!-- This is an automatically generated file.\n";
+                html += "     It will be read and overwritten.\n";
+                html += "     DO NOT EDIT! -->\n";
```
trflynn89, line 620: I've never really found headers like this to be helpful but 🤷‍♂️

Base/res/ladybird/about-pages/bookmarks.html
```diff
@@ -463,6 +474,169 @@ <h1>Bookmarks</h1>
...
+                const a = document.createElement("a");
+                a.href = url;
+                a.download = "bookmarks.html";
+
+                // FIXME: Use `a.click()` once HTMLAnchorElement's `download` attribute
+                // support is implemented (currently it navigates to blob: URLs instead
+                // of triggering a download in Ladybird).
+                ladybird.sendMessage("exportBookmarks", html);
+
+                URL.revokeObjectURL(url);
```
trflynn89, line 638: A lot of this is currently unused. While implementing `<a download>` is one path forward, there are other APIs we might implement as well:
https://developer.mozilla.org/en-US/docs/Web/API/Window/showSaveFilePicker

So let's not go through the song-and-dance of creating a blob URL and anchor tag here that are currently unused.

A FIXME is still appropriate here though - but please hang the indentation for wrapped lines:
```js
// FIXME: One long line ........
//        A second line ........
```

Libraries/LibWebView/WebUI/BookmarksUI.cpp
```diff
@@ -48,6 +55,37 @@ void BookmarksUI::move_item(JsonValue const& data)
...
+{
+    if (!data.is_string())
+        return;
+
+    auto destination = Application::the().path_for_downloaded_file("bookmarks.html"sv);
+    if (destination.is_error())
+        return;
+
+    // FIXME: Route this through FileDownloader once it supports non-network URLs
+    // (e.g. blob:/data:) without going through RequestServer DNS lookup.
```
trflynn89, line 76: I don't think this FIXME makes sense? Once we support `<a download>` or `showSaveFilePicker`, this method won't exist anymore

Libraries/LibWebView/BookmarkStore.h
```diff
@@ -7,6 +7,7 @@
 #pragma once
 
 #include <AK/Badge.h>
+#include <AK/JsonArray.h>
```
trflynn89, line 10: This include shouldn't be needed? The `JsonArray const&` should be satisfied below with its forward declaration.

Base/res/ladybird/about-pages/bookmarks.html
mikiubo: thanks! i try to replicate your graphics

Base/res/ladybird/about-pages/bookmarks.html
mikiubo: thanks!

Base/res/ladybird/about-pages/bookmarks.html
mikiubo: ok think i fixed

Base/res/ladybird/about-pages/bookmarks.html
mikiubo: you are right, sorry.
i checked with lowercase is working.
now is fixed.

Base/res/ladybird/about-pages/bookmarks.html
mikiubo: i copy paste from other browser.
now i removed.

Libraries/LibWebView/BookmarkStore.h
mikiubo: fixed

Libraries/LibWebView/WebUI/BookmarksUI.cpp
mikiubo: i think i fixed

trflynn89 (approved):

PR #8939, opened 2026-04-16
test-web: Dont glitch scroll when concurrency exceeds terminal rows

jonbgamble:
Some people run embedded terminals in the bottom panel of their IDEs. This prevents test-web's live display from blowing up their scrollback buffer.

If you don't know what I mean, resize a terminal down to single digit rows and fire up a long-ish test-web batch (prior to this commit).

Reduce main.cpp a bit by moving display stuff into Display.cpp.


tcl3:
This PR should be split into 2 commits, where one makes the advertised fix and the other moves code from `main.cpp` into it's own file. This will make it much easier for reviewers to distinguish what has been changed from what has just been moved.

PR #8940, opened 2026-04-16
LibWeb: Make srcdoc iframe test wait for nested document

awesomekling:
The test assumed the parent document's DOMContentLoaded implied the iframe srcdoc document already had a body. That is not guaranteed, and the test could intermittently hit a null contentDocument.body on slower runs.

(I've seen this test flaking on CI a couple times.)

### DIFF
--- a/Tests/LibWeb/Text/input/HTML/set-innerHTML-inside-iframe-srcdoc-document.html
+++ b/Tests/LibWeb/Text/input/HTML/set-innerHTML-inside-iframe-srcdoc-document.html
@@ -2,14 +2,23 @@
 <iframe id="ifr" srcdoc=""></iframe>
 <script src="../include.js"></script>
 <script>
-    test(() => {
-        ifr.contentDocument.body.innerHTML = "<a href='foo.html'>foo</a>";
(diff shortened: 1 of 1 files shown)


Tests/LibWeb/Text/input/HTML/set-innerHTML-inside-iframe-srcdoc-document.html
```diff
@@ -2,14 +2,23 @@
 <iframe id="ifr" srcdoc=""></iframe>
 <script src="../include.js"></script>
 <script>
-    test(() => {
-        ifr.contentDocument.body.innerHTML = "<a href='foo.html'>foo</a>";
-        let href = ifr.contentDocument.body.querySelector("a").href;
-        if (href.endsWith("foo.html")) {
-            println("PASS (Didn't crash)");
+    asyncTest(done => {
```
shannonbooth, line 5: https://github.com/LadybirdBrowser/ladybird/pull/8932 as well

awesomekling:
Oops this already has the same fix in #8932

PR #8944, opened 2026-04-16
LibWeb/DOM: Collapse ranges before DOM mutations

AtkinsSJ:
The spec text has also changed since we implemented it, but moving when we collapse the range is the only actual behaviour change.

Corresponds to:
https://github.com/whatwg/dom/commit/9363c6ddac8491f9e4c20609bc0a161ffbf72901

### DIFF
--- a/Libraries/LibWeb/DOM/Range.cpp
+++ b/Libraries/LibWeb/DOM/Range.cpp
@@ -603,45 +603,46 @@ WebIDL::ExceptionOr<GC::Ref<DocumentFragment>> Range::extract()
     if (collapsed())
         return fragment;
 
-    // 3. Let original start node, original start offset, original end node, and original end offset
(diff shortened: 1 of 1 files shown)


shannonbooth:
Appears like https://github.com/web-platform-tests/wpt/blob/2ac434171d218b29a279abb17e4a9ba23a064366/dom/ranges/Range-extractContents-dynamic-end.html is the associated WPT for this

shannonbooth:
Which we pass already, maybe is actually the other linked test (currently still open) to that DOM MR?

AtkinsSJ:
The tests in https://github.com/web-platform-tests/wpt/pull/58008 also pass with or without the changes here, so I've left them.

shannonbooth:
Curious!

AtkinsSJ:
Rebased to solve the rust-linting issue.

PR #8946, opened 2026-04-16
LibWeb: Update DOM Node moving and removing steps

AtkinsSJ:
A couple of spec changes modified what's passed to the moving and removing steps. I then also noticed we were missing the moving steps for `<option>`, so added that.


AtkinsSJ:
Rebased on master, to get around the old rust lint errors. No actual changes.

gmta (approved):

PR #8948, opened 2026-04-16
LibUnicode+LibRegex+LibJS: Fix clippy warnings

awesomekling:
After rustup update, clippy is flagging a bunch of things on CI so let's fix them.


tcl3:
Should we pin the toolchain version in `rust-toolchain.toml` to prevent this kind of random CI failure in future?

PR #8951, opened 2026-04-16
test-web: Remove deferred warnings

jonbgamble:
Deferred warnings were originally intended to suppress output during live display, ostensibly to avoid glitch scrolls.

Then, 1af74d1a7cc0b7cc8b721ed333ecd8c44b27a504 added log capture to the test-web process. Suddenly, deferred warnings became deadly because they're flushed during that tiny window after the capture notifier has stopped draining the tee pipe but before stderr is restored.

This caused a deadlock at exit. The fix is to remove this system and call warnln directly because display integrity is now protected by other means.


jonbgamble:
commit was merged as part of #8939

PR #8952, opened 2026-04-16
LibWeb: Stop emitting custom element reactions in attribute getters

awesomekling:
In other engines, the `[CEReactions]` extended attribute on an attribute only applies to the setter steps. Attribute getters never fire custom element reactions, yet our bindings generator was still pushing a new element queue onto the custom element reactions stack on entry and popping + invoking it on return for every single getter call on a `[CEReactions]` attribute.

### DIFF
--- a/Meta/Lagom/Tools/CodeGenerators/LibWeb/BindingsGenerator/IDLGenerators.cpp
+++ b/Meta/Lagom/Tools/CodeGenerators/LibWeb/BindingsGenerator/IDLGenerators.cpp
@@ -4493,14 +4493,6 @@ JS_DEFINE_NATIVE_FUNCTION(@class_name@::@attribute.getter_callback@)
 )~~~");
         }
 
-        if (attribute.extended_attributes.contains("CEReactions")) {
-            // 1. Push a new element queue onto this object's relevant agent's custom element reactions stack.
-            attribute_generator.append(R"~~~(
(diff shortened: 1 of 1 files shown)


Meta/Lagom/Tools/CodeGenerators/LibWeb/BindingsGenerator/IDLGenerators.cpp
```diff
@@ -4493,13 +4493,9 @@ JS_DEFINE_NATIVE_FUNCTION(@class_name@::@attribute.getter_callback@)
...
         }
 
-        if (attribute.extended_attributes.contains("CEReactions")) {
-            // 1. Push a new element queue onto this object's relevant agent's custom element reactions stack.
-            attribute_generator.append(R"~~~(
-    auto& reactions_stack = HTML::relevant_similar_origin_window_agent(*impl).custom_element_reactions_stack;
-    reactions_stack.element_queue_stack.append({});
-)~~~");
-        }
+        // NB: Per WebIDL, [CEReactions] on an attribute applies to the setter only.
```
shannonbooth: Does WebIDL spec define this? https://html.spec.whatwg.org/multipage/custom-elements.html#cereactions is not very clear on this, looks like https://github.com/whatwg/html/issues/10023 was raised relating to this. 

Maybe would be best to just remove the comments left behind altogether alongside the spec link just above the getter?

PR #8953, opened 2026-04-17
AK: Make short ASCII string literals compile-time constants

awesomekling:
Make the `_string`, `_fly_string`, `_utf16`, and `_utf16_fly_string` UDL operators `constexpr`, with a fast path for short (<= 7 byte) ASCII literals that folds directly into an inline `ShortString`. Previously, every `"foo"_fly_string` (and friends) involved an out-of-line call into the string factory, even though the result is entirely known at compile time.

### DIFF
--- a/AK/FlyString.h
+++ b/AK/FlyString.h
@@ -6,6 +6,7 @@
 
 #pragma once
 
+#include <AK/AllOf.h>
 #include <AK/Error.h>
 #include <AK/Format.h>
 #include <AK/Optional.h>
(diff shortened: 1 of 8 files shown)


AK/FlyString.h
```diff
@@ -131,8 +144,21 @@ struct ASCIICaseInsensitiveFlyStringTraits : public Traits<String> {
 
 }
 
-[[nodiscard]] ALWAYS_INLINE AK::FlyString operator""_fly_string(char const* cstring, size_t length)
+[[nodiscard]] ALWAYS_INLINE constexpr AK::FlyString operator""_fly_string(char const* cstring, size_t length)
 {
+    // Fast path: short ASCII strings become compile-time constants with no
+    // runtime validation or fly string table lookup.
```
tcl3: `OPTIMIZATION:` is a more common prefix for fast paths.

Comments should be wrapped at 120.

AK/FlyString.h
```diff
@@ -131,8 +144,21 @@ struct ASCIICaseInsensitiveFlyStringTraits : public Traits<String> {
...
+        bool is_ascii = true;
+        for (size_t i = 0; i < length; ++i) {
+            if (static_cast<unsigned char>(cstring[i]) >= 0x80) {
+                is_ascii = false;
+                break;
+            }
+        }
+        if (is_ascii)
+            return AK::FlyString::from_ascii_short_string_without_validation(cstring, length);
+    }
```
tcl3: Is there any performance impact from using AK methods for these fast paths?

Something like this:
```c++
if (length <= AK::Detail::MAX_SHORT_STRING_BYTE_COUNT && all_of(cstring, cstring + length,  AK::is_ascii)) {
    return FlyString::from_ascii_short_string_without_validation(cstring, length);
}
```

AK/FlyString.h
awesomekling: Struggling to get used to the 120 thing :sweat_smile:

AK/FlyString.h
awesomekling: Yeah that's better.

PR #8954, opened 2026-04-17
test-web: Add --wpt argument for running wpt tests directly

jonbgamble:
Similar to `-f`/`--filter`, `-w`/`--wpt` searches and runs tests within the gitignored `WPT/wpt` directory.

We use the upstream `WPT/wpt/serve/serve.py` test server for max compatibility.

There are no expectation files and no need to import. This is for development and debugging, not CI. You run tests and the failures show up in `results/index.html`. That's it.

For example:
```
Build/release/bin/test-web -t 3 -f Element-scrollby -w normal-flow
```
Will get you:

test-web-wpt

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,6 +47,7 @@ Libraries/LibWasm/Tests/Spec
 Tests/LibWeb/WPT/wpt
 Tests/LibWeb/WPT/metadata
 Tests/LibWeb/WPT/MANIFEST.json
+Tests/LibWeb/Fixtures/wpt-server.log
 
 test-dumps
 *.log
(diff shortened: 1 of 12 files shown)


gmta:
This depends on https://github.com/LadybirdBrowser/ladybird/pull/8951 being merged, you should mention that as such.

We would now have multiple ways of running WPT tests (`Meta/WPT.sh` exists) but you don't address that here (e.g. provide your idea of a roadmap or rationale) - please do that as well.

jonbgamble:
My rationale is that the index.html results viewer is preferable to raw wpt harness json.

I don't have a roadmap per se. In fact, I would prefer not to extend this functionality into the CI realm. It is useful to me as it stands.

jonbgamble:
> This depends on #8951 being merged, you should mention that as such.

Small correction - 8951 is a separate commit in this PR, so I imagine it's straightforward to cherry-pick around it (if necessary).

PR #8955, opened 2026-04-17
LibGfx/LibMedia: Implement video color conversion for CPU painting

Zaggy1024:
The `yuv` crate is used to convert YUV image data into RGB to create a simple bitmap-backed SkImage. This crate seems to perform better at converting 8-bit video than libyuv on ARM, and achieves the same performance on x86. It has the added benefit of some safety checks before it passes off data to the unsafe code, which caught some issues with strides that libyuv happily ignored.

To simplify things a little, the `ImmutableBitmap::create_from_yuv()` function now uploads to the GPU synchronously as well, and the YUV data is not kept alive outside the function. Since that function is only called from decoder threads, that shouldn't block anything real-time.

There were a number of other improvements that came alongside these bigger changes, see the individual commits.

### DIFF
--- a/Libraries/LibGfx/Rust/src/lib.rs
+++ b/Libraries/LibGfx/Rust/src/lib.rs
@@ -1,0 +1,241 @@
+/*
+ * Copyright (c) 2026-present, the Ladybird developers.
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+use yuv::{YuvPlanarImage, YuvRange, YuvStandardMatrix};
+
+#[repr(u8)]
+pub enum YUVRange {
+    Limited = 0,
+    Full = 1,
+}
+
+#[repr(u8)]
+pub enum YUVMatrix {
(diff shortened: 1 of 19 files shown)


Libraries/LibGfx/Rust/src/lib.rs
```diff
@@ -0,0 +1,241 @@
+/*
+ * Copyright (c) 2026-present, the Ladybird developers.
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+use yuv::{YuvPlanarImage, YuvRange, YuvStandardMatrix};
```
trflynn89, line 7: Just a flyby comment, but I think we want to avoid one big lib.rs file for each rust integration. I know yuv is the only thing here right now, but a yuv.rs would be better IMO (which is then imported into lib.rs).

See e.g. Libraries/LibUnicode/Rust/src/lib.rs which imports a calendar.rs.

PR #8958, opened 2026-04-17
LibWeb: Clean-up more input state after running each test

AtkinsSJ:
The clean-up in 71c457c36e33c68ce3ef24d49ab8ae56e731bef7 turns out to not be enough for all cases. So be more thorough and clear up anything that could affect the next test.

In particular this fixes flakiness in `Text/input/select-text.html` but hopefully it solves the issue for good!


AtkinsSJ:
This should also fix the issue of the middle-mouse-button scroll indicator showing up in screenshot/ref tests when it shouldn't.

Possibly we should be doing some of this cleanup outside of test mode too? I'm not sure. `Internals` lets tests get in weird inconsistent states that shouldn't be possible in general use.

PR #8959, opened 2026-04-17
LibWeb: CSS numeric value bounds checking simplifications

Calme1709:
This PR includes a couple of main changes, all around the theme of bounds checking for numeric CSS values

1) Numeric tokens (`<number>`, `<percentage>`, and `<dimension>`) are clamped to 32 bit `float` range at parse time, as we did with `<integer>` in b86377b.
2) Literal numeric values are validated within the relevant `parse_foo_{percentage}_value` method based on explicitly passed bounds.
3) Calculation contexts are generated per numeric value (i.e. by the `parse_foo_{percentage}_value` methods) rather than by `ValueParsingContext`.

See individual commits for details and more comprehensive rationale.


Libraries/LibWeb/CSS/Parser/ValueParsing.cpp
```diff
@@ -5410,18 +5411,23 @@ OwnPtr<BooleanExpression> Parser::parse_if_condition(TokenStream<ComponentValue>
...
+    // Represented as a <number>, the useful range of the value is 0 (representing full transparency) to 1 (representing
+    // full opacity). It can also be written as a <percentage>, which computes to the equivalent <number> (0% to 0, 100% to 1).
     auto value = parse_number_percentage_value(tokens);
 
     if (!value)
         return nullptr;
 
-    // Percentages map to the range [0,1] for opacity values
+    // NB: To match the behavior of other browsers we convert percentages to numbers at parse time instead of computed
+    //     value time as the spec says.
```
AtkinsSJ: Is there a spec issue about this? Sounds like there should be.

AtkinsSJ (approved):
This is so nice. Thank you very much!

Libraries/LibWeb/CSS/Parser/ValueParsing.cpp
Calme1709: Turns out this is actually in the spec, just a different section to where I was looking. Have updated.

AtkinsSJ (approved):

Calme1709:
Rebased to resolve conflicts

PR #8960, opened 2026-04-17
Meta: Restore gdb pretty printers

jarusll:
Restore and update gdb pretty printers - `Meta/serenity_gdb.py` => `Meta/gdb/PrettyPrinters/ak.py` 

`VirtualAddress` and `AKFixedStringBuffer` are removed.

Screenshot From 2026-04-20 22-04-38

### DIFF
--- a/Meta/Debuggers/gdb/AK.py
+++ b/Meta/Debuggers/gdb/AK.py
@@ -1,0 +1,471 @@
+# Copyright (c) 2021, Gunnar Beutner <gunnar@beutner.name>
+#
+# SPDX-License-Identifier: BSD-2-Clause
+
+import re
+
+import gdb  # type: ignore[import-untyped]
+import gdb.printing  # type: ignore[import-untyped]
+import gdb.types  # type: ignore[import-untyped]
+
+void_ptr = gdb.lookup_type("void").pointer()
+
+
+def handler_class_for_type(type, re=re.compile("^([^<]+)(<.*>)?$")):
+    typename = str(type.tag)
+
+    match = re.match(typename)
+    if not match:
+        return UnhandledType
+
+    klass = match.group(1)
+
+    if klass == "AK::Array":
+        return AKArray
+    elif klass == "AK::Atomic":
+        return AKAtomic
+    elif klass == "AK::Detail::IntrusiveList":
+        return AKIntrusiveList
+    elif klass == "AK::DistinctNumeric":
+        return AKDistinctNumeric
+    elif klass == "AK::FixedArray":
+        return AKFixedArrayPrinter
+    elif klass == "AK::HashMap":
+        return AKHashMap
+    elif klass == "AK::RefCounted":
+        return AKRefCounted
+    elif klass == "AK::RefPtr":
+        return AKRefPtr
+    elif klass == "AK::OwnPtr":
+        return AKOwnPtr
+    elif klass == "AK::NonnullRefPtr":
+        return AKRefPtr
+    elif klass == "AK::SinglyLinkedList":
+        return AKSinglyLinkedList
+    elif klass == "AK::String":
+        return AKString
+    elif klass == "AK::ByteString":
+        return AKByteString
+    elif klass == "AK::StringView":
+        return AKStringView
+    elif klass == "AK::Variant":
+        return AKVariant
+    elif klass == "AK::Optional":
+        return AKOptional
+    elif klass == "AK::Vector":
+        return AKVector
+    else:
+        return UnhandledType
+
+
+class UnhandledType:
+    def __init__(self, val):
+        self.val = val
+
+    def to_string(self):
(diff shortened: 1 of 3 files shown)


konradekk (commented):
(not sire, but maybe it'd be better to pull in relevant commits instead?)

Meta/gdb/PrettyPrinters/ak.py
```diff
@@ -0,0 +1,465 @@
...
+        gdb.set_convenience_variable("_tmp", self.val.reference_value())
+        string_view = gdb.parse_and_eval("$_tmp.bytes_as_string_view()")
+        return AKStringView(string_view).to_string()
+
+    @classmethod
+    def prettyprint_type(cls, type):
+        return "AK::String"
+
+
+class AKDeprecatedString:
```
AtkinsSJ: `AKByteString`

Meta/gdb/PrettyPrinters/ak.py
```diff
@@ -0,0 +1,465 @@
...
+        return [(f"[{i}]", elements[i]) for i in range(self.array_size)]
+
+    @classmethod
+    def prettyprint_type(cls, type):
+        template_type = type.template_argument(0)
+        template_size = type.template_argument(1)
+        return f"AK::Array<{template_type}, {template_size}>"
+
+
+class AKHashMapPrettyPrinter:
```
AtkinsSJ: Kinda weird to have `PrettyPrinter` in the name here and not anywhere else.

AtkinsSJ (commented):
Generally I think this is fine. Just a couple of nits that aren't your fault, but while you're adding this back in it might as well be nice.

However, I think this can be a single commit instead of 6.

AtkinsSJ:
Can you rewrite the commit message so it's a proper description of the changes instead of a list of titles?

jarusll:
I should have looked at commit messages after squashing. Fixed it.

AtkinsSJ (approved):
Great, thanks! I've been trying this out when debugging this morning and it helps.

AtkinsSJ:
Oh, can you move this into `Meta/Debuggers` like for the lldb implementation? That came up after I'd already approved this but it would be better to be consistent.

jarusll:
Can do and on it

AtkinsSJ (commented):
You accidentally put "aMeta" in the commit title instead of "Meta". Otherwise I think we can finally merge this, I'm just giving it another test-run locally.

Meta/Debuggers/gdb/AK.py
```diff
@@ -0,0 +1,471 @@
...
+    @classmethod
+    def prettyprint_type(cls, ty):
+        names = ", ".join(handler_class_for_type(t).prettyprint_type(t) for t in AKVariant.resolve_types(ty))
+        return f"AK::Variant<{names}>"
+
+
+class AKOptional:
+    def __init__(self, val):
+        self.val = val
+        self.has_value = bool(self.val["m_has_value"])
```
AtkinsSJ, line 322: This doesn't work for all Optionals - we have two different versions of it defined depending on what it's holding. For the second kind, they just have `m_pointer` which is null for the empty case:
https://github.com/LadybirdBrowser/ladybird/blob/639584fecb939a6d4456f7dc0a404f0cb255ac11/AK/Optional.h#L630

Meta/Debuggers/gdb/AK.py
```diff
@@ -0,0 +1,471 @@
...
+        self.contained_type = self.val.type.strip_typedefs().template_argument(0)
+
+    def to_string(self):
+        return AKOptional.prettyprint_type(self.val.type)
+
+    def children(self):
+        if self.has_value:
+            data = self.val["m_storage"]
+            return [(self.contained_type.name, data)]
+        return [("OptionalNone", "{}")]
```
AtkinsSJ, line 332: This makes the value show up like it's holding the string "{}".

Image

Meta/Debuggers/gdb/AK.py
```diff
@@ -0,0 +1,471 @@
...
+        return AKRefCounted
+    elif klass == "AK::RefPtr":
+        return AKRefPtr
+    elif klass == "AK::OwnPtr":
+        return AKOwnPtr
+    elif klass == "AK::NonnullRefPtr":
+        return AKRefPtr
+    elif klass == "AK::SinglyLinkedList":
+        return AKSinglyLinkedList
+    elif klass == "AK::String":
```
AtkinsSJ, line 45: It'd be great if we could get FlyString support too.

AtkinsSJ (approved):
I did find a couple of little issues while trying this out, BUT I would much rather get this merged now and have those fixed later. Debugging without this is painful. 😅 I'm re-running some flaky CI jobs but let me know if it fails again.

PR #8962, opened 2026-04-17
LibWeb/Bindings: Factor IDLGenerator generate_to_cpp into many functions

shannonbooth:
This is mostly a mechanical set of changes, preparing for some future improvements to the generator to reduce the amount of messy code here.

Next step should hopefully be making the handling of optional and nullable into generic helpers to simplify all of the "to cpp" code, also fixing bugs and very inconsistent way this is handled. Currently all of the nullable parameter logic is inlined for each of the type code generation, which is really not the way things should be structured.

Other than the mechanical changes of just shifting stuff into type specific to cpp helpers, this includes two very small tidy ups of https://github.com/LadybirdBrowser/ladybird/commit/bce11466b700007054e5c272e7fdb68872fa3d6d and https://github.com/LadybirdBrowser/ladybird/commit/811a19322710524d0490704844d709f4ef496ec1

### DIFF
--- a/Meta/Lagom/Tools/CodeGenerators/LibWeb/BindingsGenerator/IDLGenerators.cpp
+++ b/Meta/Lagom/Tools/CodeGenerators/LibWeb/BindingsGenerator/IDLGenerators.cpp
@@ -1185,6 +1185,44 @@ static void generate_record_to_cpp(SourceGenerator& scoped_generator, IDL::Param
 )~~~");
 }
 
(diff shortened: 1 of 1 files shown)


shannonbooth:
I also do have some other changes sitting around which get rid of is_platform_object by making the IDLGenerator a single invocation across every interface (then we know if something is a platform object just by an interface being defined over IDL), but I'll hold off on that one for now since type conversion is where the main problems are

AtkinsSJ (approved):
Any improvements to this code are very welcome!

PR #8965, opened 2026-04-17
Short-circuit CacheIndex eviction when under its size limit

kalenikaliaksandr:
makes Google Maps go much faster!

remove_entries_exceeding_cache_limit() is called after every network
response, but the cache is usually still under budget and nothing needs
to evict. Every one of those calls currently still runs the
window-function eviction SQL over the whole CacheIndex table just to
conclude there is nothing to do.

Short-circuit the call when the cache is already within its configured
size limit. To make that check cheap, maintain m_total_estimated_size
as a running total of the cache's estimated byte size, so the no-op
case becomes a single u64 compare and the DB is only touched when
there is real work.

Bookkeeping:
- Seed the total in CacheIndex::create() via a new
  select_total_estimated_size statement (COALESCE(..., 0) so an empty
  index returns 0 rather than NULL).
- Each Entry caches serialized_request_headers_size and
  serialized_response_headers_size so we don't re-serialize to
  recompute its footprint; Entry::estimated_size() centralizes the
  arithmetic.
- create_entry() adds the new entry's size. Any row it displaces is
  removed via DELETE ... RETURNING so the total stays accurate even
  for entries that were never loaded into m_entries.
- remove_entry() and the bulk DELETE statements were extended with
  the same RETURNING clause for the same reason.
- update_response_headers() shifts the total by the signed delta
  between old and new serialized header size.

Also COALESCEs estimate_cache_size_accessed_since over an empty table
to 0 so callers don't have to special-case NULL.

### DIFF
--- a/Libraries/LibHTTP/Cache/CacheIndex.cpp
+++ b/Libraries/LibHTTP/Cache/CacheIndex.cpp
@@ -98,8 +98,16 @@ ErrorOr<CacheIndex> CacheIndex::create(Database::Database& database, LexicalPath
 
     Statements statements {};
     statements.insert_entry = TRY(database.prepare_statement("INSERT OR REPLACE INTO CacheIndex VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?);"sv));
-    statements.remove_entry = TRY(database.prepare_statement("DELETE FROM CacheIndex WHERE cache_key = ? AND vary_key = ?;"sv));
-    statements.remove_entries_accessed_since = TRY(database.prepare_statement("DELETE FROM CacheIndex WHERE last_access_time >= ? RETURNING cache_key, vary_key;"sv));
+    statements.remove_entry = TRY(database.prepare_statement(R"#(
+        DELETE FROM CacheIndex
+        WHERE cache_key = ? AND vary_key = ?
+        RETURNING data_size + OCTET_LENGTH(request_headers) + OCTET_LENGTH(response_headers);
+    )#"sv));
+    statements.remove_entries_accessed_since = TRY(database.prepare_statement(R"#(
+        DELETE FROM CacheIndex
+        WHERE last_access_time >= ?
+        RETURNING cache_key, vary_key, data_size + OCTET_LENGTH(request_headers) + OCTET_LENGTH(response_headers);
+    )#"sv));
     statements.select_entries = TRY(database.prepare_statement("SELECT * FROM CacheIndex WHERE cache_key = ?;"sv));
     statements.update_response_headers = TRY(database.prepare_statement("UPDATE CacheIndex SET response_headers = ? WHERE cache_key = ? AND vary_key = ?;"sv));
     statements.update_last_access_time = TRY(database.prepare_statement("UPDATE CacheIndex SET last_access_time = ? WHERE cache_key = ? AND vary_key = ?;"sv));
@@ -120,13 +128,18 @@ ErrorOr<CacheIndex> CacheIndex::create(Database::Database& database, LexicalPath
             FROM RankedCacheIndex
             WHERE cumulative_estimated_size > ?
         )
-        RETURNING cache_key, vary_key;
+        RETURNING cache_key, vary_key, data_size + OCTET_LENGTH(request_headers) + OCTET_LENGTH(response_headers);
     )#"sv));
 
     statements.estimate_cache_size_accessed_since = TRY(database.prepare_statement(R"#(
-        SELECT SUM(data_size + OCTET_LENGTH(request_headers) + OCTET_LENGTH(response_headers))
+        SELECT COALESCE(SUM(data_size + OCTET_LENGTH(request_headers) + OCTET_LENGTH(response_headers)), 0)
         FROM CacheIndex
         WHERE last_access_time >= ?;
+    )#"sv));
+
+    statements.select_total_estimated_size = TRY(database.prepare_statement(R"#(
+        SELECT COALESCE(SUM(data_size + OCTET_LENGTH(request_headers) + OCTET_LENGTH(response_headers)), 0)
+        FROM CacheIndex;
     )#"sv));
 
     auto disk_space = TRY(FileSystem::compute_disk_space(cache_directory));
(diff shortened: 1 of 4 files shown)


Libraries/LibHTTP/Cache/CacheIndex.cpp
```diff
@@ -167,6 +167,8 @@ ErrorOr<void> CacheIndex::create_entry(u64 cache_key, u64 vary_key, String url,
     if (data_size + serialized_request_headers.length() + serialized_response_headers.length() > m_limits.maximum_disk_cache_entry_size)
         return Error::from_string_literal("Cache entry size exceeds allowed maximum");
 
+    delete_entry(cache_key, vary_key);
```
gmta: In terms of the API it feels wrong to call `::delete_entry()` inside `::create_entry()` - even though I know that at this point, `::delete_entry()` does not do anything except mess with `m_entries`.

Couldn't we just take `m_entries.ensure(cache_key)` and check `.contains(..)` _before_ `.append(move(entry))`? That also saves us from potentially trashing the vector and recreating it again if an entry is the only one present for the given `cache_key`.

awesomekling:
Happy to report that this also makes https://wowhead.com/ WAY faster. :)

Libraries/LibHTTP/Cache/CacheIndex.cpp
```diff
@@ -98,8 +98,16 @@ ErrorOr<CacheIndex> CacheIndex::create(Database::Database& database, LexicalPath
 
     Statements statements {};
     statements.insert_entry = TRY(database.prepare_statement("INSERT OR REPLACE INTO CacheIndex VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?);"sv));
-    statements.remove_entry = TRY(database.prepare_statement("DELETE FROM CacheIndex WHERE cache_key = ? AND vary_key = ?;"sv));
-    statements.remove_entries_accessed_since = TRY(database.prepare_statement("DELETE FROM CacheIndex WHERE last_access_time >= ? RETURNING cache_key, vary_key;"sv));
+    statements.remove_entry = TRY(database.prepare_statement(R"#(
+        DELETE FROM CacheIndex
+        WHERE cache_key = ? AND vary_key = ?
```
avbdr, line 103: WHERE cache_key = ? AND vary_key = ? will need (cache_key, vary_key) index to work much faster.

Libraries/LibHTTP/Cache/CacheIndex.cpp
```diff
@@ -163,42 +182,82 @@ ErrorOr<void> CacheIndex::create_entry(u64 cache_key, u64 vary_key, String url,
 
     auto serialized_request_headers = serialize_headers(request_headers);
     auto serialized_response_headers = serialize_headers(response_headers);
+    auto serialized_request_headers_size = static_cast<u64>(serialized_request_headers.length());
+    auto serialized_response_headers_size = static_cast<u64>(serialized_response_headers.length());
+
+    auto estimated_entry_size = [](u64 data_size, u64 request_headers_size, u64 response_headers_size) {
+        return data_size + request_headers_size + response_headers_size;
+    };
+    auto entry_size = estimated_entry_size(data_size, serialized_request_headers_size, serialized_response_headers_size);
```
trflynn89: This lambda seems pretty overkill

Libraries/LibHTTP/Cache/CacheIndex.cpp
```diff
@@ -163,42 +182,82 @@ ErrorOr<void> CacheIndex::create_entry(u64 cache_key, u64 vary_key, String url,
...
         .response_headers = move(response_headers),
         .data_size = data_size,
+        .serialized_request_headers_size = serialized_request_headers_size,
+        .serialized_response_headers_size = serialized_response_headers_size,
         .request_time = request_time,
         .response_time = response_time,
         .last_access_time = now,
     };
 
+    auto& entries = m_entries.ensure(cache_key);
```
trflynn89, line 212: Looks like this got squashed into the wrong commit?

Libraries/LibHTTP/Cache/CacheIndex.cpp
```diff
@@ -228,8 +289,18 @@ void CacheIndex::update_response_headers(u64 cache_key, u64 vary_key, NonnullRef
...
-    m_database->execute_statement(m_statements.update_response_headers, {}, serialize_headers(response_headers), cache_key, vary_key);
+    auto serialized_response_headers = serialize_headers(response_headers);
+    auto serialized_response_headers_size = static_cast<u64>(serialized_response_headers.length());
+
+    m_database->execute_statement(m_statements.update_response_headers, {}, serialized_response_headers, cache_key, vary_key);
+
+    if (serialized_response_headers_size >= entry->serialized_response_headers_size)
+        m_total_estimated_size += serialized_response_headers_size - entry->serialized_response_headers_size;
+    else
+        m_total_estimated_size -= entry->serialized_response_headers_size - serialized_response_headers_size;
```
trflynn89: This seems a little over complicated - can't we just do the following to avoid overflow?
```c++
m_total_estimated_size -= entry->serialized_response_headers_size;
m_total_estimated_size += serialized_response_headers_size;
```

Libraries/LibHTTP/Cache/CacheIndex.h
```diff
@@ -27,10 +27,17 @@ class CacheIndex {
...
         NonnullRefPtr<HeaderList> response_headers;
         u64 data_size { 0 };
+        u64 serialized_request_headers_size { 0 };
+        u64 serialized_response_headers_size { 0 };
 
         UnixDateTime request_time;
         UnixDateTime response_time;
         UnixDateTime last_access_time;
+
+        u64 estimated_size() const
```
trflynn89, line 37: This is unused?

gmta (approved):

Tests/LibHTTP/TestCacheIndex.cpp
```diff
@@ -0,0 +1,107 @@
...
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include <AK/LexicalPath.h>
+#include <LibDatabase/Database.h>
+#include <LibHTTP/Cache/CacheIndex.h>
+#include <LibHTTP/Cache/Utilities.h>
+#include <LibTest/TestCase.h>
+
+static constexpr auto s_cache_directory = "/tmp"sv;
```
trflynn89: This fails on Windows. We just use the path to get an upper limit on the partition the cache is stored in. This probably doesn't make much sense for memory-backed databases, but for now you could probably just use `Core::StandardPaths::cache_directory()`.

PR #8966, opened 2026-04-17
LibWeb/Bindings: Another small batch of IDLGenerator improvements

shannonbooth:
Nothing particularly meaningful, but helps improve things a little.


AtkinsSJ (approved):
Nice progress! One day it may be conquered. 😆

PR #8968, opened 2026-04-18
Meta: Add option to select GUI Framework to ladybird.py

ADKaster:


Meta/host_platform.py
```diff
@@ -22,6 +22,14 @@ class HostSystem(enum.IntEnum):
     BSD = enum.auto()
 
 
+class GUIFramework(enum.StrEnum):
+    # enum.auto() will tolower the values
+    Qt = "Qt"
+    AppKit = "AppKit"
+    Gtk = "Gtk"
+    Android = "Android"
```
ADKaster, line 30: It bothers me a bit that this file has no idea what cross-compilation is, so the Android option remains unused. But, 🤷 that would be a large refactor.

Meta/ladybird.py
```diff
@@ -43,6 +44,9 @@ def main():
     compiler_parser.add_argument("--cc", required=False, default=default_cc)
     compiler_parser.add_argument("--cxx", required=False, default=default_cxx)
     compiler_parser.add_argument("--jobs", "-j", required=False)
+    compiler_parser.add_argument(
```
ADKaster, line 47: this isn't really a "compiler option", but it fit in better with the settings for -DCMAKE_C_COMPILER, than the target name, or the clutter of adding a new parent parser in 15 places.

Meta/host_platform.py
```diff
@@ -76,3 +84,21 @@ def libc_name(self) -> str:
 
         libc, _ = platform.libc_ver()
         return libc
+
+    def valid_gui_frameworks(self) -> list[GUIFramework]:
+        """
+        List of valid GUI frameworks based on target platform.
+        Keep in sync with Meta/CMake/gui_framework.cmake
```
ADKaster, line 91: this is technically already out of sync, because one could, according to the cmake file, use the Gtk UI on Windows, but I sure haven't tested that.

Meta/ladybird.py
```diff
@@ -156,19 +160,19 @@ def main():
...
         build_main(build_dir, args.jobs, args.target, args.args)
         debug_main(platform.host_system, build_dir, args.target, args.debugger, args.cmd)
     elif args.command == "profile":
-        build_dir = configure_main(platform, args.preset, args.cc, args.cxx)
+        build_dir = configure_main(platform, args.preset, args.cc, args.cxx, args.gui)
         build_main(build_dir, args.jobs, args.target)
         profile_main(platform.host_system, build_dir, args.target, args.args)
     elif args.command == "install":
-        build_dir = configure_main(platform, args.preset, args.cc, args.cxx)
+        build_dir = configure_main(platform, args.preset, args.cc, args.cxx, args.gui)
```
trflynn89, line 175: There are a couple `configure_main` calls below that also need `args.gui`.

Meta/ladybird.py
```diff
@@ -207,6 +211,7 @@ def configure_main(platform: Platform, preset: str, cc: str, cxx: str) -> Path:
         build_preset_dir,
         f"-DCMAKE_C_COMPILER={cc}",
         f"-DCMAKE_CXX_COMPILER={cxx}",
+        f"-DLADYBIRD_GUI_FRAMEWORK={gui}",
```
trflynn89, line 214: Does this work on its own to change GUIs? We purposefully exit early from this step to avoid invoking `cmake` all the time. I'm not sure how we end up back down here unless we blow away the build directory first, as even touching a CMake file will not bring us back down here.

ADKaster: ah, that's true. this only works at initial configure. I suppose we could read CMakeCache.txt to check if we're going to swap guis and force a cmake call.

ADKaster: I've updated it to mean "passing --gui means force this gui". That involves looking at CMakeCache.txt when passing that argument.

As an alternative, we could change the semantics to mean "passing --gui always configures". But that would mean that any other configure-time arguments, like cc or cxx or j, must always be passed alongside --gui to avoid super-churn.

I prefer scanning CMakeCache.txt to tying those arguments that tightly.

trflynn89 (approved):

PR #8969, opened 2026-04-18
UI/Gtk: Start converting GTK UI components to Rust

ADKaster:
In general the shape of the patch doesn't hurt *too* much to look at. In theory we can remove as many of the unsafe FFI calls as possible going fwd.


Cargo.toml
```diff
@@ -3,6 +3,12 @@ members = [
     "Libraries/LibJS/Rust",
     "Libraries/LibRegex/Rust",
     "Libraries/LibUnicode/Rust",
+    "UI/Gtk/Rust",
```
ADKaster, line 7: ~~I'm not quite sure how to do the "this is an optional member" dance~~ 
~~Why did we choose to do workspaces here? Is it the right solution?~~

Edit: nvm, the CI failure I was seeing was due to the cargo-sources.json not being updated.


Also, I could only get CLion to be happy when adding something like this:

```diff
diff --git a/.cargo/config.toml b/.cargo/config.toml
new file mode 100644
index 00000000000..1c9ed740611
--- /dev/null
+++ b/.cargo/config.toml
@@ -0,0 +1,2 @@
+[env]
+PKG_CONFIG_PATH = "/home/andrew/ladybird-org/ladybird-browser/Build/release/vcpkg_installed/x64-linux-dynamic/lib/pkgconfig/"
```
Which is clearly not commitable...

UI/Gtk/Rust/src/browser_window.rs
```diff
@@ -0,0 +1,237 @@
+/*
+ * Copyright (c) 2026, the Ladybird developers.
```
jdahlin, line 2: I don't personally care too much, but I believe copyright should be retained when copying rewriting/porting one language to another.

I saw that it was added to some of the other files, but not here.

UI/Gtk/Rust/src/browser_window.rs
```diff
@@ -0,0 +1,237 @@
...
+
+use adw::subclass::prelude::*;
+use glib::object::Cast;
+use glib::translate::{IntoGlibPtr, ToGlibPtr, from_glib_borrow};
+use gtk::{CompositeTemplate, TemplateChild, gio, glib};
+
+mod imp {
+    use super::*;
+
+    #[derive(Default, CompositeTemplate)]
```
jdahlin, line 15: This syntax is so much nicer than C!

UI/Gtk/Rust/src/location_entry.rs
```diff
@@ -0,0 +1,497 @@
...
+}
+
+mod imp {
+    use super::*;
+
+    #[derive(Default)]
+    pub struct LadybirdLocationEntry {
+        pub popover: OnceCell<gtk::Popover>,
+        pub list_box: OnceCell<gtk::ListBox>,
+        pub suggestions: RefCell<Vec<String>>,
```
jdahlin, line 50: autocomplete missing

UI/Gtk/Rust/src/location_entry.rs
```diff
@@ -0,0 +1,497 @@
...
+            self.parent_constructed();
+            let obj = self.obj();
+            obj.initialize();
+        }
+
+        fn dispose(&self) {
+            if let Some(popover) = self.popover.get() {
+                popover.popdown();
+                popover.unparent();
+            }
```
jdahlin, line 77: state.clear() is called in C here, not needed?

UI/Gtk/Rust/src/location_entry.rs
```diff
@@ -0,0 +1,497 @@
...
+            unsafe { destroy(self.user_data) };
+        }
+    }
+}
+
+mod imp {
+    use super::*;
+
+    #[derive(Default)]
+    pub struct LadybirdLocationEntry {
```
jdahlin, line 47: This is broken up in state and entry (gtk "self") in C, any specific reason for merging?

UI/Gtk/Rust/src/location_entry.rs
```diff
@@ -0,0 +1,497 @@
...
+        fn constructed(&self) {
+            self.parent_constructed();
+            let obj = self.obj();
+            obj.initialize();
+        }
+
+        fn dispose(&self) {
+            if let Some(popover) = self.popover.get() {
+                popover.popdown();
+                popover.unparent();
```
jdahlin, line 76: C sets popover as nullptr after calling unparent(), is that relevant?

UI/Gtk/Rust/src/location_entry.rs
```diff
@@ -0,0 +1,497 @@
...
+
+impl LadybirdLocationEntry {
+    pub fn new() -> Self {
+        glib::Object::builder().build()
+    }
+
+    fn initialize(&self) {
+        self.set_hexpand(true);
+        self.set_placeholder_text(Some("Enter URL or search..."));
+
```
jdahlin, line 101: Old location entry did some search engine initialization here.

UI/Gtk/Rust/src/location_entry.rs
```diff
@@ -0,0 +1,497 @@
...
+        let builder = gtk::Builder::from_resource("/org/ladybird/Ladybird/gtk/location-entry.ui");
```
jdahlin, line 102: Since we now have the lovely template syntax, maybe use it here?

UI/Gtk/Rust/src/location_entry.rs
```diff
@@ -0,0 +1,497 @@
...
+                    suggestions.get(index as usize).cloned()
+                };
+                if let Some(suggestion) = suggestion {
+                    obj.set_entry_text_suppressed(&suggestion, false);
+                    obj.hide_completions();
+                    obj.navigate();
+                }
+            }
+        ));
+
```
jdahlin, line 134: autocomplete on_autocomplete_query_complete callback handling missing here.

UI/Gtk/Rust/src/location_entry.rs
```diff
@@ -0,0 +1,497 @@
...
+        self.connect_changed(glib::clone!(
+            #[weak(rename_to = obj)]
+            self,
+            move |_| {
+                obj.on_changed();
```
jdahlin, line 139: Same here, missing a bit of autocomplete logic

UI/Gtk/Rust/src/location_entry.rs
```diff
@@ -0,0 +1,497 @@
...
+        let imp = self.imp();
+        imp.updating_text.set(true);
+        self.set_text(text);
+        if move_cursor_to_end {
+            self.set_position(-1);
+        }
+        imp.updating_text.set(false);
+    }
+
+    fn set_security_icon_internal(&self, should_show_lock: SchemeShouldShowLock, has_url: bool) {
```
jdahlin, line 219: This logic is a bit different from C, maybe we can stay true in the first commit and then introduce nicer logic later to simplify review.

UI/Gtk/BrowserWindow.cpp
```diff
@@ -13,11 +13,19 @@
...
 #include <UI/Gtk/Menu.h>
+#include <UI/Gtk/RustFFI.h>
 #include <UI/Gtk/Tab.h>
 #include <UI/Gtk/WebContentView.h>
 
 namespace Ladybird {
 
+static FFI::SchemeShouldShowLock scheme_should_show_lock(URL::URL const& url)
+{
+    return (url.scheme() == "https"sv || url.scheme() == "wss"sv)
```
jdahlin, line 24: Missing file, resource, about and data handling, which we did before.

UI/Gtk/Rust/src/location_entry.rs
```diff
@@ -0,0 +1,497 @@
...
+                );
+            }
+        }
+    }
+
+    pub fn set_url(&self, url: Option<&str>, should_show_lock: SchemeShouldShowLock) {
+        let text = url.unwrap_or_default();
+        self.set_entry_text_suppressed(text, false);
+
+        self.set_security_icon_internal(should_show_lock, !text.is_empty());
```
jdahlin, line 254: This is also quite a bit different, harder to follow.

UI/Gtk/Rust/src/location_entry.rs
```diff
@@ -0,0 +1,497 @@
...
+    }
+
+    pub fn clear_text(&self) {
+        self.set_entry_text_suppressed("", false);
+        self.set_attributes(&pango::AttrList::new());
+        self.set_security_icon_internal(SchemeShouldShowLock::No, false);
+    }
+
+    fn update_display_attributes(&self) {
+        // Keep this as a no-op style reset for now; domain highlighting can be added back
```
jdahlin, line 268: Why not do it now as part of this PR, to avoid regressing functionallity?

UI/Gtk/Rust/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+
+[dependencies.adw]
+package = "libadwaita"
+version = "0.9.1"
+features = ["v1_8"]
+
+[dependencies.gtk]
+package = "gtk4"
+version = "0.11.2"
+features = ["v4_22"]
```
jdahlin, line 17: Do we actually need any features from 4.22 here?

UI/Gtk/Rust/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
```
jdahlin, line 1: Maybe add some kind of check here that we only build this on !mac and !windows. Not sure what you did different but when I tried I got windows deps pulled in if I didnt do that.

jdahlin (commented):
Thanks for this, looking good!

There are essentially two things:
- Ideally try to keep all features (autocomplete, icon lock)
- Initial conversion should be as boring as possible, especially when it comes to all logic in location entry, same names, same kind of logic and so forth, so its easier to review. Some things were refactored and shuffled around.

nicolasdanelon:
EPIC! I was working on this already hahahahahaha

UI/Gtk/Rust/src/browser_window.rs
ADKaster: that's an oversight, i'll fix that up for these rs files.

UI/Gtk/Rust/src/location_entry.rs
ADKaster: No clue on this one, that's all the agent. I'll take a closer look at the rs lib's docs to see if that's a real pattern.

UI/Gtk/Rust/src/location_entry.rs
ADKaster: imo the current logic is a lot easier to follow, and eliminated a bunch of code paths that were never used. but i'll take a closer look. it's clear that something went wrong with this port step that it dropped so many features.

jdahlin: I guess I meant its hard to follow as in; which part is this part of porting and which part of this is improvements. Not against it at all, just that when I went in with my "is this an identical port" glasses, I wasn't able to immediately map it to the old code.

ADKaster: ah. good point there. I'll change up the commits to move cleanup/refactoring into a separate commit

jdahlin:
@ADKaster do you intend to finish this, or can I pick it up?

ADKaster:
I'm not actively working on this and don't have much free time in the near term, so feel free to take over the rust ideas

nicolasdanelon:
@jdahlin let me know if you want to split the load, totally up to you.

jdahlin:
> @jdahlin let me know if you want to split the load, totally up to you.

@nicolasdanelon Thanks, that would be very helpful!

I can take this one, most of the work is reviewing anyway. Once this is done we can start splitting up the remaining work:
- port one C/C++ file over to Rust per PR, mechanically
- all open PRs (search for UI/Gtk)
- implementing the remaining gap from AppKit/Qt backends.

PR #8973, opened 2026-04-18
LibWeb: Deduplicate some fragment hit testing code

Psychpsyo:
No change in behavior in this one, but I'm working on some changes in this area and this makes this code much easier to work with.

### DIFF
--- a/Libraries/LibWeb/Painting/PaintableWithLines.cpp
+++ b/Libraries/LibWeb/Painting/PaintableWithLines.cpp
@@ -183,43 +183,27 @@ TraversalDecision PaintableWithLines::hit_test_fragments(CSSPixelPoint position,
                 // To determine the best place, we first find the closest fragment horizontally to the cursor. If we could not
                 // find one, then find for the closest vertically above the cursor. If we knew the direction of selection, we
                 // would look above if selecting upward.
+                HitTestResult hit_test_result {
+                    .paintable = const_cast<Paintable&>(fragment.paintable()),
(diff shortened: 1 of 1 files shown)


Psychpsyo:
CI failures unrelated.

Libraries/LibWeb/Painting/PaintableWithLines.cpp
```diff
@@ -183,43 +183,27 @@ TraversalDecision PaintableWithLines::hit_test_fragments(CSSPixelPoint position,
...
-                            .horizontal_distance = local_position.x() - fragment_absolute_rect.right(),
-                        };
-                        if (callback(hit_test_result) == TraversalDecision::Break)
-                            return TraversalDecision::Break;
+                        hit_test_result.index_in_node += fragment.length_in_code_units();
+                        hit_test_result.horizontal_distance = local_position.x() - fragment_absolute_rect.right();
                     }
                 }
+
+                if (callback(hit_test_result) == TraversalDecision::Break)
```
shannonbooth, line 205: Hmm, does this perhaps change behaviour by invoking the callback now for the else { condition also? Specifically wondering about x == fragment_absolute_rect.right()

Psychpsyo: Ah, true.
I've gone ahead and made the else branch unconditional instead. This should only improve behavior because we only end up in this code if we clicked either to the left or the right of the fragment, and there's no reason that clicking exactly on the right edge should lead to any sort of different behavior.

PR #8974, opened 2026-04-18
Rust: Set import granularity to Item

vgaspary:
This PR updates the `rustfmt` configuration to enforce **Item** import granularity, performs a project-wide reformat, and updates the formatting script to use the nightly toolchain.

### Why "Item" Granularity?

By default, `rustfmt` persists the import granularity. In practice, most Rust code has import granularity "Module" due to LSP's and editors' preferences. "Item" has the following advantages:

* **Better conflict resolution:** Adding or removing an import only affects a single line. This ensures atomic diffs and significantly reduces merge conflicts when multiple PRs modify imports from the same module.
* **Greppability:** Since every import is a full path on its own line, it is trivial to search the codebase for specific type usage without dealing with nested braces.

### Example

```rust
// before
use std::collections::{BinaryHeap, HashMap, HashSet};

// after
use std::collections::BinaryHeap;
use std::collections::HashMap;
use std::collections::HashSet;
```

### Precedent

For reference, this style is used in **[jj (Jujutsu)](https://github.com/martinvonz/jj)**, a Rust-powered vcs.

### Stability

While [`imports_granularity`](https://rust-lang.github.io/rustfmt/?version=v1.9.0&search=group#imports_granularity) is currently unstable and requires `cargo +nightly fmt`, in practice it's stable enough to be used by a mature project like `jj`. If my proposal is approved, considering the fact that Rust is still rather new to Ladybird, I would argue that this is the best time to make such a change and reformat all Rust files.


Meta/lint-ci.sh
```diff
@@ -48,10 +48,10 @@ else
     ((FAILURES+=1))
 fi
 
-if cargo fmt --check ; then
-    echo -e "[${GREEN}OK${NC}]: cargo fmt --check"
+if cargo +nightly fmt --check ; then
```
shannonbooth, line 51: At least on my MacOS system, this appears to cause this error:

```
error: no such command: `+nightly`

help: invoke `cargo` through `rustup` to handle `+toolchain` directives
[FAIL]: cargo +nightly fmt --check
```

Libraries/LibRegex/Rust/src/regex.rs
```diff
@@ -7,9 +7,33 @@
...
+use crate::ast::Atom;
+/// High-level regex API.
+///
+/// This is the main entry point for using the regex engine.
+use crate::ast::Disjunction;
+/// High-level regex API.
+///
+/// This is the main entry point for using the regex engine.
+use crate::ast::Flags;
+/// High-level regex API.
```
shannonbooth: This does not seem like something we should be repeating 5 times here 😅

Meta/lint-ci.sh
shannonbooth: Also the CI job appears to be failing: https://github.com/LadybirdBrowser/ladybird/actions/runs/25466141954/job/74719746847?pr=8974

Libraries/LibRegex/Rust/src/regex.rs
vgaspary: Correct, this is one of the reasons why `imports_granularity` is a nightly feature of `rustfmt`. Fixed it.

Meta/lint-ci.sh
vgaspary: Thanks for pointing it out. I have zero experience with github actions, my AI agent suggested this, but it doesn't seem to work.

```
//  .github/actions/setup/action.yaml

    - name: 'Install Nightly Rust toolchain for linting'
      if: ${{ inputs.type == 'lint' }}
      shell: bash
      run: rustup toolchain install nightly --component rustfmt
```

R-Goc:
Please add a corresponding documentation change that now the nightly toolchain (or at least the rustfmt component) is needed to format the rust code. This is practice means it is always needed since editors run format on save.

vgaspary:
> Please add a corresponding documentation change that now the nightly toolchain (or at least the rustfmt component) is needed to format the rust code. This is practice means it is always needed since editors run format on save.

I assume you're referring to the build instructions in the documentation. Currently it only says this about Rust:

> A Rust toolchain is also required. You can install it via [rustup](https://rustup.rs/).

This remains correct. Developers don't have to install the nightly toolchain manually. If Rust was installed via `rustup`, as instructed, the first run of `cargo +nightly fmt` will take care of everything, obtain the nightly toolchain.

In practice, developers run `cargo fmt` for the first time, it fails pointing out the unstable feature used and instructs running `cargo +nightly fmt`.

R-Goc:
But IDEs will run cargo fmt on the file automatically. So won't this cause failures unless you are running with nightly as the default toolchain?

vgaspary:
> But IDEs will run cargo fmt on the file automatically. So won't this cause failures unless you are running with nightly as the default toolchain?

I changed the channel in rust-toolchain.toml to "nightly", now `rust-analyzer` (IDEs and editors) will use the nightly toolchain too.

Meta/lint-ci.sh
vgaspary: It worked.

shannonbooth (approved):

PR #8977, opened 2026-04-18
Base: Improve about:version and about:processes

nicolasdanelon:
This patch updates the about:version page to use the shared Ladybird WebUI styling and present version, build, and runtime details in cards.

It also adds the Ladybird logo and a page title to about:processes, without otherwise changing the process table behavior.

### Before
Screenshot_20260419_000849

### After
Screenshot_20260419_003007

Screenshot_20260419_002348

### DIFF
--- a/Base/res/ladybird/about-pages/processes.html
+++ b/Base/res/ladybird/about-pages/processes.html
@@ -26,6 +26,24 @@
                 font-size: 10pt;
             }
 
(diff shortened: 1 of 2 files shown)


shlyakpavel:
Could you please squash the two commits into one?

shannonbooth (approved):

PR #8978, opened 2026-04-19
test-web: Dont leak captured fds when a helper process dies

jonbgamble:
Ensure test-web stays under the open fd limit on runs featuring many WebWorker aborts and high repeat counts.

### DIFF
--- a/Tests/LibWeb/test-web/TestRunCapture.cpp
+++ b/Tests/LibWeb/test-web/TestRunCapture.cpp
@@ -28,6 +28,12 @@ TestRunCapture::TestRunCapture()
     process_manager.on_process_added = [this](WebView::Process& process) {
         setup_output_capture_for_helper_process(process);
     };
+    m_previous_on_process_exited = move(process_manager.on_process_exited);
+    process_manager.on_process_exited = [this](WebView::Process&& process) {
+        consume_helper_capture(process.pid());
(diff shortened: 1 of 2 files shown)


Tests/LibWeb/test-web/TestRunCapture.cpp
```diff
@@ -28,6 +28,12 @@ TestRunCapture::TestRunCapture()
     process_manager.on_process_added = [this](WebView::Process& process) {
         setup_output_capture_for_helper_process(process);
     };
+    m_previous_on_process_exited = move(process_manager.on_process_exited);
+    process_manager.on_process_exited = [this](WebView::Process&& process) {
+        m_helper_output_captures.remove(process.pid());
```
shannonbooth: Do we also need to make sure we drain stderr/stdout pipes by this stage before removing it?

jonbgamble: Yes, that seems correct.

Do you want it in this PR?

shannonbooth: I _think_ this might be an issue introduced (sort of) alongside this, right? Probably should include it I think

jonbgamble: I'm not so sure. I don't think they ever were drained, but they should be now.

PR #8979, opened 2026-04-19
LibWeb/HTML: Fix ruby parse error check for rp/rt

shlyakpavel:
Use `&&` instead of `||` when checking whether the current node is neither `rtc` nor `ruby`.
With `||`, the condition is always true, so the parser reports a parse error unconditionally in this branch.


AtkinsSJ (approved):
Nice catch!

PR #8981, opened 2026-04-19
LibIPC: Dont VERIFY when encoding placeholder TransportHandles

jonbgamble:
Empty transport handles can be generated in a few places in Ladybird sources, notably in WebContentClient::request_worker_agent when view_for_page_id finds nothing.

If those handles reach encode, a VERIFY is triggered in the broker process. A page lookup failure should not be fatal to the browser, so I'll boldly assert it is better to return an error here.

Suggestions are welcome as I did not spend much time chasing for upstream root causes. The failure was observed during large runs of origin and IndexedDB heavy wpt tests in test-web.

### DIFF
--- a/Libraries/LibIPC/TransportHandle.cpp
+++ b/Libraries/LibIPC/TransportHandle.cpp
@@ -33,8 +33,11 @@ ErrorOr<NonnullOwnPtr<Transport>> TransportHandle::create_transport() const
 template<>
 ErrorOr<void> encode(Encoder& encoder, TransportHandle const& handle)
 {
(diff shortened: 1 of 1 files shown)


kalenikaliaksandr (approved):

kalenikaliaksandr:
```
Run git ls-files '*.ipc' | xargs ./Build/bin/IPCMagicLinter
xargs: ./Build/bin/IPCMagicLinter: No such file or directory
Error: Process completed with exit code 127.
```

jonbgamble:
We'll try again I guess.

PR #8983, opened 2026-04-19
LibWeb: Respect image response MIME type over .svg URL suffix

shannonbooth:
SharedResourceRequest was treating any URL ending in .svg as SVG, even when the response Content-Type was some other format (like image/webp). This could result in transformed CDN image URLs to fail decoding.

Only use the .svg URL suffix fallback when no MIME type was provided.

I saw this on a live site where some image icons were not rendering, reduced it to this, but unfortunately we seem to have another issue causing it to still not load so I don't have a nice before and after.


shannonbooth:
Oh nevermind, I take it back, it does fix icons on the site

Before:

image


After:

image

PR #8986, opened 2026-04-19
LibWeb: Clamp limited grid item contributions by fixed max tracks

shannonbooth:
When resolving grid track sizes, limited min/max-content contributions should be capped by fixed max track sizing functions, including the argument to fit-content(). We were instead falling back to the grid container maximum size, which allowed a grid item with overflowing contents in a fit-content(0) row to inflate the intrinsic block size of a nested grid.

That bogus intrinsic height could then be used for the grid's second row sizing pass, causing unrelated flexible rows to absorb the extra space.

I am not super sure about the correctness of this change, but it comes from a (very annoying to do) reduction from lichess.org main screen.

Sidenote: We have some sort of bug where the icons for these buttons don't show up on initial page load, but do after the mouse is hovered over them.

Before:

image

After:

image

### DIFF
--- a/Libraries/LibWeb/Layout/GridFormattingContext.cpp
+++ b/Libraries/LibWeb/Layout/GridFormattingContext.cpp
@@ -2461,6 +2461,13 @@ int GridItem::gap_adjusted_column() const
     return column.value() * 2;
 }
 
+bool GridFormattingContext::should_treat_grid_container_maximum_size_as_none(GridDimension dimension) const
+{
+    if (dimension == GridDimension::Column)
+        return should_treat_max_width_as_none(grid_container(), m_available_space->width);
+    return should_treat_max_height_as_none(grid_container(), m_available_space->height);
+}
+
 CSSPixels GridFormattingContext::calculate_grid_container_maximum_size(GridDimension dimension) const
 {
     auto const& computed_values = grid_container().computed_values();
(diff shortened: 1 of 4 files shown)


Libraries/LibWeb/Layout/GridFormattingContext.cpp
```diff
@@ -2569,6 +2576,53 @@ CSSPixels GridFormattingContext::calculate_max_content_contribution(GridItem con
     return min(result, maximum_size);
 }
 
+Optional<CSSPixels> GridFormattingContext::calculate_fixed_max_track_size_limit(GridItem const& item, GridDimension dimension) const
+{
+    // https://www.w3.org/TR/css-grid-2/#algo-spanning-items
```
tcl3: Spec links should reference the editor's draft version.

Libraries/LibWeb/Layout/GridFormattingContext.cpp
```diff
@@ -2569,6 +2576,53 @@ CSSPixels GridFormattingContext::calculate_max_content_contribution(GridItem con
...
 }
 
+Optional<CSSPixels> GridFormattingContext::calculate_fixed_max_track_size_limit(GridItem const& item, GridDimension dimension) const
+{
+    // https://drafts.csswg.org/css-grid-2/#algo-spanning-items
+    // For an item spanning multiple tracks, the upper limit used to calculate its limited  min-/max-content
+    // contribution is the sum of the fixed max track sizing functions of any tracks it spans, and is applied
+    // if it only spans such tracks.
+    //
+    // https://www.w3.org/TR/css-grid-2/#algo-terms
```
tcl3: Link should be: https://drafts.csswg.org/css-grid-2/#algo-terms

Libraries/LibWeb/Layout/GridFormattingContext.cpp
```diff
@@ -2569,6 +2576,53 @@ CSSPixels GridFormattingContext::calculate_max_content_contribution(GridItem con
     return min(result, maximum_size);
 }
 
+Optional<CSSPixels> GridFormattingContext::calculate_fixed_max_track_size_limit(GridItem const& item, GridDimension dimension) const
+{
+    // https://drafts.csswg.org/css-grid-2/#algo-spanning-items
+    // For an item spanning multiple tracks, the upper limit used to calculate its limited  min-/max-content
```
tcl3: double space

tcl3 (commented):
The fix looks good to me. Just a few nitpicks.

tcl3 (approved):

PR #8988, opened 2026-04-19
LibWeb+LibGfx: Synthesize small-caps when font lacks OpenType support

shannonbooth:
When font-variant-caps requests small-caps, all-small-caps, petite-caps, all-petite-caps, or unicase but the font's GSUB table doesn't contain required features, simulate by uppercasing the text and rendering it at 70% of the original point size. Also passes locale through the shaping stack so casing transforms match those of the text-transform property.

Fixes the presentation of "Table of contents" on:

https://www.w3.org/TR/css-grid-2

I have no idea on how to add a test for this, suggestions welcome.

Before:

image

After:

image

### DIFF
--- a/Libraries/LibGfx/TextLayout.cpp
+++ b/Libraries/LibGfx/TextLayout.cpp
@@ -10,10 +10,12 @@
 #include <AK/Utf16String.h>
 #include <AK/Utf16View.h>
 #include <LibGfx/Font/Font.h>
+#include <LibGfx/FourCC.h>
 #include <LibGfx/Point.h>
 #include <LibGfx/TextLayout.h>
 #include <core/SkFont.h>
 #include <core/SkTextBlob.h>
+#include <harfbuzz/hb-ot.h>
 #include <harfbuzz/hb.h>
 
 namespace Gfx {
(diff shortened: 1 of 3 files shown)


gmta:
> I have no idea on how to add a test for this, suggestions welcome.

Screenshot test?

jdahlin:
wpt font variation tests has their own magic font with some custom entries in the font table, you could do something similar that combined with screenshots should help

shannonbooth:
> wpt font variation tests has their own magic font with some custom entries in the font table, you could do something similar that combined with screenshots should help

Could you link to this by any chance? This was the main issue I was thinking of is what font etc to use for this

jdahlin:
> > wpt font variation tests has their own magic font with some custom entries in the font table, you could do something similar that combined with screenshots should help
> 
> Could you link to this by any chance? This was the main issue I was thinking of is what font etc to use for this

https://github.com/web-platform-tests/wpt/blob/master/css/css-fonts/font-variant-01.html
https://github.com/web-platform-tests/wpt/raw/b50d28d1ad25a2be1808da13caf902c603904cdf/css/css-fonts/support/fonts/FontWithFancyFeatures.otf

PR #8990, opened 2026-04-20
CI: Build Gtk UI in Linux Sanitizer presets

ADKaster:
Ensure that we run tests with the Qt UI still

### DIFF
--- a/.devcontainer/features/ladybird/install-fedora.sh
+++ b/.devcontainer/features/ladybird/install-fedora.sh
@@ -5,7 +5,7 @@ set -e
 dnf install -y git gh
 
 # Ladybird dev dependencies
-dnf install -y autoconf-archive automake ccache cmake curl google-noto-sans-mono-fonts liberation-sans-fonts \
+dnf install -y autoconf-archive automake bison ccache cmake curl google-noto-sans-mono-fonts liberation-sans-fonts \
     libglvnd-devel libtool nasm ninja-build patchelf perl-FindBin perl-IPC-Cmd perl-lib qt6-qtbase-devel \
     qt6-qttools-devel qt6-qtwayland-devel tar unzip zip zlib-ng-compat-static
 
--- a/.devcontainer/features/ladybird/install-ubuntu.sh
+++ b/.devcontainer/features/ladybird/install-ubuntu.sh
@@ -24,7 +24,7 @@ install_llvm_key() {
 ### Install packages
 
 apt update -y
-apt install -y lsb-release git python3 autoconf autoconf-archive automake build-essential cmake libdrm-dev libgl1-mesa-dev libtool nasm ninja-build pkg-config qt6-base-dev qt6-tools-dev-tools qt6-wayland ccache fonts-liberation2 zip unzip curl tar
+apt install -y lsb-release git python3 autoconf autoconf-archive automake bison build-essential cmake libdrm-dev libgl1-mesa-dev libtool libxkbcommon-dev nasm ninja-build pkg-config qt6-base-dev qt6-tools-dev-tools qt6-wayland ccache fonts-liberation2 zip unzip curl tar
 
 ### Install Rust toolchain
 curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -80,6 +80,7 @@ runs:
             autoconf \
             autoconf-archive \
             automake \
+            bison \
             build-essential \
             ccache \
             cmake \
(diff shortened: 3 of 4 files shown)


.github/workflows/nightly-lagom.yml
```diff
@@ -33,6 +33,13 @@ jobs:
             clang_plugins: false
             runner_labels: '["blacksmith-8vcpu-ubuntu-2404"]'
 
+          - os_name: 'Linux'
+            arch: 'x86_64'
+            build_preset: 'Release_Gtk'
```
ADKaster, line 38: is this enough to not destroy the vcpkg binary cache for other jobs?

cqundefine: Does it really matter, vcpkg does clean the install directory, but it shouldn't remove stuff from the binary cache, the worst case scenario is when a normal job restores the cache it also restores the cache of gtk packages but that shouldn't be that big of a deal. Maybe I don't understand something about vcpkg so feel free to correct me. 

FWIW tracing back how the GitHub cache key is built from vcpkg it uses the OS, arch, toolchain, the "extra" key and possibly the `vcpkg.json` file hash, so it would seem it doesn't touch build_preset, the `lagom_template.yml` sets the `cache_key_extra` based on some factors, but I'm pretty sure it won't make it differentiate the caches.

ADKaster: mostly I'm worried about what ends up in that cache because swapping b/w gtk and qt on linux causes a rebuild of harfbuzz to add the glib dep, and then a rebuild of skia due to its harfbuzz dep. If the nightly vcpkg cache getting pushed up affects later PRs, that could cause some unnecessary rebuilds.

Though lately the killer for CI times seems to be apt package pulls onto the blacksmith runners, so it's probably a nothingburger.

ADKaster: honestly I'd rather yeet the glib* dep from our harfbuzz. Not sure what the right arrangement of default-features and features is for the associated packages though. the *glib* dep in harfbuzz is pretty pointless when we're using icu.

<https://harfbuzz.github.io/harfbuzz-hb-glib.html>

cqundefine: I didn't realize the harfbuzz features change, that does complicate things. Honestly having the `cache_key_extra` be different for gtk builds could just be the easiest solution.

About the `glib` dep on harfbuzz, gave it a quick look and I can see that GTK (which is what afaict pulls it in in vcpkg) itself uses for example the `hb-glib.h` header at least in one place so I don't think you can just get rid of it unfortunately.

trflynn89: > the lagom_template.yml sets the cache_key_extra based on some factors, but I'm pretty sure it won't make it differentiate the caches.

We set the `cache_key_extra` to the build preset, plus some other things based on different conditions. This directly affects the cache we push/pull, so a unique preset here should be sufficient.

Meta/CMake/presets/CMakeUnixPresets.json
```diff
@@ -14,6 +14,13 @@
         "rhs": "Windows"
       }
     },
+    {
+      "hidden": true,
```
trflynn89, line 18: Not sure how much I love UI-specific presets. Seems cleaner to use the normal presets, and just add a parameter to lagom-template.yml to specify a UI? That can then just be added to `cmake_options` (and probably `ccache_key` as well) in the Assign Build Parameters step.

ADKaster: Works for me. I was mostly thinking about the IDE side, where it's 'simpler' to point CLion at the new preset than it is to copy a preset and add the flag.

ADKaster: I've refactored this to only update the ci ymls. Now there's a gui_framework input to lagom-template.yml, that defaults to 'auto'. If it's not 'auto', we set -DLAYBIRD_GUI_FRAMEWORk and append the gui framework to the ccache_key_extra.

As with current, this input has no effect on Windows. A bit strange, but I'm not messing with windows atm.

Only the nightly-lagom.yml file actually sets this parameter. The ci.yml workflow will always go through the auto path.

ADKaster: As discussed on discord, skipping the complex logic and just setting -DLADYBIRD_GUI_FRAMEWORK to Gtk for sanitizer builds, just like we do to force both AppKit and Qt chromes on macOS.

trflynn89 (approved):
sweet! looks much simpler

Will probably need to add these packages to the setup action:
https://github.com/LadybirdBrowser/ladybird/blob/master/Documentation/BuildInstructionsLadybird.md#additional-prerequisites-for-the-gtk-ui

PR #8991, opened 2026-04-20
LibWebView: Do not diddle ProcessManager state from MachBootstrapListener thread

jonbgamble:
The MachBootstrapListener thread was sneaking through the verify_event_loop guard in ProcessManager by virtue of having no event loop.

- First, in Application::initialize we move event loop creation above the bootstrap lambda (which calls ProcessManager::set_process_mach_port)
- Then, we issue that port assignment using deferred_invoke so it actually happens on the main thread.
- Finally, we strengthen the guard using VERIFY(Threading::is_main_thread()) to protect ProcessManager's internal state.

I saw evidence of this race condition in set_process_mach_port during a wpt run. It culminated in Ladybird attempting to access a previously moved NonnullOwnPtr<ProcessInfo>

### DIFF
--- a/Libraries/LibMain/CMakeLists.txt
+++ b/Libraries/LibMain/CMakeLists.txt
@@ -4,3 +4,4 @@ set(SOURCES
 
 ladybird_lib(LibMain main TYPE STATIC)
 target_link_libraries(LibMain PRIVATE LibThreading)
+target_link_libraries(LibMain INTERFACE "$<LINK_ONLY:LibThreading>")
--- a/Libraries/LibMain/CMakeLists.txt
+++ b/Libraries/LibMain/CMakeLists.txt
@@ -3,3 +3,4 @@ set(SOURCES
 )
 
 ladybird_lib(LibMain main TYPE STATIC)
+target_link_libraries(LibMain PRIVATE LibThreading)
--- a/Libraries/LibWebView/ProcessManager.h
+++ b/Libraries/LibWebView/ProcessManager.h
@@ -9,7 +9,7 @@
 #include <AK/Function.h>
 #include <AK/JsonValue.h>
 #include <AK/Types.h>
-#include <LibCore/EventLoop.h>
+#include <LibCore/Forward.h>
 #include <LibCore/Platform/ProcessStatistics.h>
 #include <LibWebView/Forward.h>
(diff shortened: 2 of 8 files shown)


Libraries/LibMain/CMakeLists.txt
```diff
@@ -3,3 +3,4 @@ set(SOURCES
 )
 
 ladybird_lib(LibMain main TYPE STATIC)
+target_link_libraries(LibMain PUBLIC LibThreading)
```
shannonbooth: Does this need public linking?

Libraries/LibWebView/ProcessManager.h
```diff
@@ -15,6 +15,7 @@
 #include <LibWebView/Process.h>
 #include <LibWebView/ProcessMonitor.h>
 #include <LibWebView/ProcessType.h>
+#include <pthread.h>
```
shannonbooth: Is this include needed?

Libraries/LibMain/CMakeLists.txt
jonbgamble: Nope

Libraries/LibWebView/ProcessManager.h
jonbgamble: Nope

shannonbooth:
The linker error in https://github.com/LadybirdBrowser/ladybird/actions/runs/25345039771/job/74311834704?pr=8991 seems like it could be relevant

jonbgamble:
I am tentatively thinking [we could use this mechanism](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#link-context) to keep LibMain's copy of LibThreading private while resolving this. I'll give it a shot.

Libraries/LibMain/CMakeLists.txt
jonbgamble: Maybe.

shannonbooth:
I dont really understand why it needs to be public only for sanitizer build

jonbgamble:
I'm going to clean out the ladybird fork from my github pretty soon. I'm not sure what happens to the PRs but if they disappear, it's not punitive - this is a heads up in case one of those fixes could save you some tokens or whatever. I'll give ya till next week.

PR #8993, opened 2026-04-20
LibWeb: Migrate color functions to `ColorFunctionStyleValue`

tcl3:
This PR extends `ColorFunctionStyleValue`, so it can be used for every color type (except `light-dark()` and `color-mix()`). The necessary metadata to describe each color type has been put into a descriptor table which then drives how ColorFunctionStyleValue behaves for each type. 

The subsequent commits then migrate each color type to the extended `ColorFunctionStyleValue` and remove each of the now unnecessary StyleValues

My main motivation for this was to reduce the amount of plumbing needed to implement relative colors.


AtkinsSJ (approved):
Very, very nice!

I like big negative diffs, and I cannot lie.

PR #9001, opened 2026-04-20
LibWeb+LibWebView: Convert `about:version` to a proper WebUI

trflynn89:
Passing the browser command line and executable path to every WebContent process just in case we load `about:version` always felt a bit weird. We now use the WebUI framework to load this information on demand.

### DIFF
--- a/Base/res/ladybird/about-pages/version.html
+++ b/Base/res/ladybird/about-pages/version.html
@@ -3,7 +3,6 @@
     <head>
         <meta charset="UTF-8" />
         <meta name="color-scheme" content="light dark" />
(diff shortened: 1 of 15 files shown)


shannonbooth:
somewhat related thought I got, but will WebUI mean we can eventually remove the need for resource scheme URLs as web ui becomes the new mechanism for this always?

trflynn89:
Probably not, we use `resource://` to load all kinds of things (web resources, fonts, config files, etc). But we will (sooner than later IMO) want to move all resource/file loading out of WebContent and into the UI process to lock it down.

PR #9003, opened 2026-04-21
AK: Use system allocator in debug builds

jonbgamble:
Rolling back mimalloc resolves a curious regression where Text/input/canvas/export-offscreencanvas.html exports a jpeg that is 8 pixels smaller than expected.

It also prevents a multi-minute hang during the GenerateWindowOrWorkerInterface codegen stage which runs quite frequently in ordinary work.

### DIFF
--- a/AK/kmalloc.cpp
+++ b/AK/kmalloc.cpp
@@ -73,9 +73,12 @@ nothrow_t const nothrow;
 #    include <cstddef>
 #    include <cstring>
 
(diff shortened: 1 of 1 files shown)


jonbgamble:
Better fix coming

jonbgamble:
I feel like I should add that the severe hangs during Ladybird browsing only seem to happen on mac builds with debug mimalloc. Linux isn't affected.

Both platforms exhibit the strange Text/input/canvas/export-offscreencanvas.html failure though.

PR #9005, opened 2026-04-21
LibWeb: Parse visual-box in overflow-clip-margin

mikiubo:
Support content-box, padding-box, and border-box in overflow-clip-margin parsing, including combinations with length.

Canonicalize parsed values to match the grammar and expected CSSOM serialization in css-overflow parsing tests.

### DIFF
--- a/Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
+++ b/Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
@@ -3329,20 +3329,65 @@ RefPtr<StyleValue const> Parser::parse_math_depth_value(TokenStream<ComponentVal
 // https://drafts.csswg.org/css-overflow-4/#overflow-clip-margin
 RefPtr<StyleValue const> Parser::parse_overflow_clip_margin_value(TokenStream<ComponentValue>& tokens)
 {
-    // <visual-box> || <length [0,∞]>
-    // FIXME: Implement the <visual-box> part of this.
+    // <visual-box> || <length>
+    auto parse_visual_box = [this](auto& tokens) -> Optional<Keyword> {
+        auto transaction = tokens.begin_transaction();
+        auto maybe_visual_box = parse_keyword_value(tokens);
+        if (!maybe_visual_box)
+            return {};
 
-    if (auto length = parse_length_value(tokens, non_negative_range)) {
-        return length.release_nonnull();
+        auto keyword = maybe_visual_box->to_keyword();
+        if (keyword != Keyword::ContentBox && keyword != Keyword::PaddingBox && keyword != Keyword::BorderBox)
+            return {};
+
+        transaction.commit();
+        return keyword;
+    };
+
+    RefPtr<StyleValue const> length;
+    Optional<Keyword> visual_box;
+
+    for (size_t i = 0; i < 2; ++i) {
+        tokens.discard_whitespace();
+        if (!tokens.has_next_token())
+            break;
+
+        if (!visual_box.has_value()) {
+            if (auto maybe_visual_box = parse_visual_box(tokens); maybe_visual_box.has_value()) {
+                visual_box = maybe_visual_box.release_value();
+                continue;
+            }
+        }
+
+        if (!length) {
+            if (auto maybe_length = parse_length_value(tokens, infinite_range)) {
+                length = maybe_length.release_nonnull();
+                continue;
+            }
+        }
+
+        return nullptr;
     }
 
-    return nullptr;
+    tokens.discard_whitespace();
+    if (tokens.has_next_token() || (!visual_box.has_value() && !length))
+        return nullptr;
+
+    if (!length)
+        length = LengthStyleValue::create(Length::make_px(0));
+
+    if (!visual_box.has_value() || *visual_box == Keyword::PaddingBox)
+        return length;
+
+    if (length->is_length() && length->as_length().length().raw_value() == 0)
+        return KeywordStyleValue::create(*visual_box);
+
+    return StyleValueList::create({ KeywordStyleValue::create(*visual_box), length.release_nonnull() }, StyleValueList::Separator::Space);
 }
 
 RefPtr<StyleValue const> Parser::parse_overflow_clip_margin_shorthand(PropertyID property_id, TokenStream<ComponentValue>& tokens)
 {
-    // <visual-box> || <length [0,∞]>
-    // FIXME: Implement the <visual-box> part of this.
+    // <visual-box> || <length>
 
     if (auto value = parse_overflow_clip_margin_value(tokens)) {
         auto const& longhands = longhands_for_shorthand(property_id);
--- a/Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
+++ b/Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
@@ -47,6 +47,7 @@
 #include <LibWeb/CSS/StyleValues/LengthStyleValue.h>
 #include <LibWeb/CSS/StyleValues/NumberStyleValue.h>
 #include <LibWeb/CSS/StyleValues/OpenTypeTaggedStyleValue.h>
+#include <LibWeb/CSS/StyleValues/OverflowClipMarginStyleValue.h>
 #include <LibWeb/CSS/StyleValues/PercentageStyleValue.h>
 #include <LibWeb/CSS/StyleValues/PositionStyleValue.h>
 #include <LibWeb/CSS/StyleValues/RepeatStyleStyleValue.h>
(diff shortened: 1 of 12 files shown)


Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
```diff
@@ -3447,13 +3447,64 @@ RefPtr<StyleValue const> Parser::parse_math_depth_value(TokenStream<ComponentVal
...
+            if (auto maybe_visual_box = parse_visual_box(tokens); maybe_visual_box.has_value()) {
+                visual_box = maybe_visual_box.release_value();
+                continue;
+            }
+        }
+
+        if (!length) {
+            auto transaction = tokens.begin_transaction();
+            if (auto maybe_length = parse_length_value(tokens)) {
+                if (maybe_length->is_length() && maybe_length->as_length().length().raw_value() < 0)
```
Psychpsyo: Why are we rejecting negative values here?
It's not in the old code, and none of the newly passing tests seem to say anything about that case.

Psychpsyo:
Consider checking out the function at https://github.com/LadybirdBrowser/ladybird/blob/fd44da68296960ce859af5230e5f3c133fbd9adc/Libraries/LibWeb/Painting/PaintableBox.cpp#L369

It's the one place this value is actually used and it needs to distinguish between no value given and padding-box, which I don't think your parsing accomodates for right now(?).

Libraries/LibWeb/CSS/Parser/PropertyParsing.cpp
mikiubo: You're right — I got confused because some WPTs (for example  
https://github.com/web-platform-tests/wpt/blob/04197f69eb/css/css-overflow/overflow-clip-margin-013.html) still reference the older spec URL  
https://www.w3.org/TR/css-overflow-3/#propdef-overflow-clip-margin.  
The [0, +inf] note/comment also misled me.

mikiubo:
> Consider checking out the function at
> 
> https://github.com/LadybirdBrowser/ladybird/blob/fd44da68296960ce859af5230e5f3c133fbd9adc/Libraries/LibWeb/Painting/PaintableBox.cpp#L369
> 
> It's the one place this value is actually used and it needs to distinguish between no value given and padding-box, which I don't think your parsing accomodates for right now(?).


Thanks for pointing this out. I agree this is important, especially in PaintableBox::overflow_clip_edge_rect(), where we need to preserve the semantic distinction between an omitted visual-box and an explicit padding-box.


I considered three approaches:
1. keep canonicalization in parsing,
2. canonicalize only during serialization,
3. add a dedicated OverflowClipMargin value type/structure.


I currently think (3) is the most maintainable direction but i think it will require some effort (i think).
@Psychpsyo  Let me know your thinking.

Psychpsyo:
>     1. keep canonicalization in parsing,
> 
>     2. canonicalize only during serialization,
> 
>     3. add a dedicated OverflowClipMargin value type/structure.
> 
> 
> I currently think (3) is the most maintainable direction but i think it will require some effort (i think). @Psychpsyo Let me know your thinking.

I'm pretty sure 1 won't work cause it needs to serialize differently between replaced and non-replaced elements.
2 seems sensible to me, because CSS only ever cares about canonical order during serialization. Futzing around with it at any other point doesn't seem all that useful.
3 also seems sensible, imo.

mikiubo:
> > ```
> > 1. keep canonicalization in parsing,
> > 
> > 2. canonicalize only during serialization,
> > 
> > 3. add a dedicated OverflowClipMargin value type/structure.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I currently think (3) is the most maintainable direction but i think it will require some effort (i think). @Psychpsyo Let me know your thinking.
> 
> I'm pretty sure 1 won't work cause it needs to serialize differently between replaced and non-replaced elements. 2 seems sensible to me, because CSS only ever cares about canonical order during serialization. Futzing around with it at any other point doesn't seem all that useful. 3 also seems sensible, imo.

i try to implement 3. 
let me know

Libraries/LibWeb/Layout/Node.cpp
```diff
@@ -745,7 +746,25 @@ void NodeWithStyle::apply_style(CSS::ComputedProperties const& computed_style)
     computed_values.set_inset(computed_style.length_box(CSS::PropertyID::Left, CSS::PropertyID::Top, CSS::PropertyID::Right, CSS::PropertyID::Bottom, CSS::LengthPercentageOrAuto::make_auto()));
     computed_values.set_margin(computed_style.length_box(CSS::PropertyID::MarginLeft, CSS::PropertyID::MarginTop, CSS::PropertyID::MarginRight, CSS::PropertyID::MarginBottom, CSS::Length::make_px(0)));
     computed_values.set_padding(computed_style.length_box(CSS::PropertyID::PaddingLeft, CSS::PropertyID::PaddingTop, CSS::PropertyID::PaddingRight, CSS::PropertyID::PaddingBottom, CSS::Length::make_px(0)));
-    computed_values.set_overflow_clip_margin(computed_style.length_box(CSS::PropertyID::OverflowClipMarginLeft, CSS::PropertyID::OverflowClipMarginTop, CSS::PropertyID::OverflowClipMarginRight, CSS::PropertyID::OverflowClipMarginBottom, CSS::Length::make_px(0)));
+    {
+        auto extract_side = [&](CSS::PropertyID property_id) -> CSS::OverflowClipMarginSide {
+            auto const& value = computed_style.property(property_id);
+            if (value.is_overflow_clip_margin()) {
+                auto const& ocm = value.as_overflow_clip_margin();
```
Psychpsyo: Shortening `overflow_clip_margin` to `ocm` just makes the code harder to read. You should generally use the long form.

Libraries/LibWeb/Painting/PaintableBox.cpp
```diff
@@ -369,24 +369,46 @@ CSSPixelRect PaintableBox::absolute_border_box_rect() const
 CSSPixelRect PaintableBox::overflow_clip_edge_rect() const
 {
     // https://drafts.csswg.org/css-overflow-4/#overflow-clip-margin
-    // Values are defined as follows:
+    // These properties define the overflow clip edge of the box, i.e. precisely how far outside
+    // its bounds the box's content is allowed to paint before being clipped by effects
+    // (such as overflow: clip) that are defined to clip to the box's overflow clip edge.
+    auto const& ocm = computed_values().overflow_clip_margin();
```
Psychpsyo: Same thing about shortening to `ocm`.

Libraries/LibWeb/CSS/StyleValues/ShorthandStyleValue.cpp
```diff
@@ -182,7 +182,6 @@ void ShorthandStyleValue::serialize(StringBuilder& builder, SerializationMode mo
     };
 
     // Then special cases
-    // FIXME: overflow-clip-margin needs a special case here for when its longhands aren't identical.
```
Psychpsyo, line 185: This needs to stay here because you did not fix this.
`overflow-clip-margin` a) can't be serialized if there is differing values set for each side and b) needs special serialization rules for when the values are the same.
But we do this currently / with your patch:

Image

That is not a valid value for that property to take, so it should not serialize as such.
I'm not saying you need to fix that in this PR, just that the FIXME shouldn't be removed.

Could maybe link to https://github.com/w3c/csswg-drafts/issues/8381 as well, since that has some considerations for changing the shorthand to support these situations.

Psychpsyo (commented):
Looking good overall!

Libraries/LibWeb/Painting/PaintableBox.cpp
mikiubo: thanks fixed

Libraries/LibWeb/Layout/Node.cpp
mikiubo: thanks fixed

Libraries/LibWeb/CSS/StyleValues/ShorthandStyleValue.cpp
mikiubo: ok thanks ! on this i completely missed the point. 
I readd the FIXME and also the reference of the w3c issue.

Libraries/LibWeb/Painting/PaintableBox.cpp
```diff
@@ -369,24 +369,46 @@ CSSPixelRect PaintableBox::absolute_border_box_rect() const
...
-    //     If omitted, defaults to 'padding-box' on non-replaced elements, or 'content-box' on replaced elements.
-    // FIXME: We can't parse this yet so it's always omitted for now.
-    auto overflow_clip_edge = absolute_padding_box_rect();
-    if (layout_node().is_replaced_box()) {
-        overflow_clip_edge = absolute_rect();
-    }
+    //     Specifies the box edge to use as the overflow clip edge origin,
+    //     i.e. when the specified offset is zero.
+    //     If omitted, defaults to padding-box on non-replaced elements,
+    //     or content-box on replaced elements.
```
gmta: Why did you rewrap these spec comments? We wrap at column 120. Also, you removed the quotes from things like 'padding-box' and 'content-box' which are still present in the spec text.

Check the other spec comments as well. Let's not feng shui these if not strictly necessary :)

Psychpsyo: Note on the quotes: They are missing in a bunch of our spec comments since the spec has them in ::before and ::after pseudos which don't copy with Ctrl+C. Although I agree that they should get put back in manually.

mikiubo: sorry! i think now is fixed

gmta:
Could you rebase this on `master` for a final check please?

mikiubo:
> Could you rebase this on `master` for a final check please?
rebased it seems there are no conflict

PR #9010, opened 2026-04-21
RS: Disable curl socket notifiers when no longer needed

kalenikaliaksandr:
When curl invokes the socket callback to tell us it only wants one polling direction (e.g. CURL_POLL_IN without CURL_POLL_OUT, or vice versa), we previously had no way to disable the other direction's notifier. Once created, a notifier stayed enabled and kept firing curl_multi_socket_action() for events curl was no longer interested in.

Merge the read and write branches into a single helper that also disables the notifier for a direction when its CURL_POLL_* flag is absent from the mask. This measurably improves performance by avoiding redundant curl_multi_socket_action() calls on sockets curl has asked us to stop watching in a given direction.

### DIFF
--- a/Services/RequestServer/ConnectionFromClient.cpp
+++ b/Services/RequestServer/ConnectionFromClient.cpp
@@ -227,35 +227,30 @@ int ConnectionFromClient::on_socket_callback(CURL*, int sockfd, int what, void* 
         return 0;
     }
 
-    if (what & CURL_POLL_IN) {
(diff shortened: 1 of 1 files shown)


gmta:
> This measurably improves performance

Got some figures?

kalenikaliaksandr:
> > This measurably improves performance
> 
> Got some figures?

not really! I just profiled what RS is doing while loading a website (vk.com specifically) and previously it spent 2.67s in `multi_socket`. with this change it went down to <100ms.

gmta (approved):

PR #9013, opened 2026-04-21
Meta: Build system refactor (Part 3/N)

cqundefine:
Follow up to #8674, more cleanups and improvements in different parts of the project.
CC @ADKaster (has your vcpkg host tools idea from comments on the previous PR)


AK/Function.h
```diff
@@ -295,10 +245,6 @@ class Function<Out(In...)> {
             VERIFY(wrapper);
             delete wrapper;
             break;
-        case FunctionKind::Block:
```
ADKaster, line 298: @trflynn89 is AK::Function construction from blocks something that you could see being useful in the AppKit port?

It seems a shame to remove this feature entirely. std::function supports it, fwiw.

CMakeLists.txt
```diff
@@ -51,10 +51,9 @@ endif()
 set(LADYBIRD_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}")
 list(APPEND CMAKE_MODULE_PATH "${LADYBIRD_SOURCE_DIR}/Meta/CMake")
 
-include(check_for_dependencies)
-include(gui_framework)
```
ADKaster, line 55: no one includes gui_framework.cmake after this change, unless I overlooked something?

CMakeLists.txt
```diff
@@ -16,6 +16,10 @@ endif()
 
 include("Meta/CMake/gui_framework.cmake")
 
+if(LAGOM_TOOLS_ONLY)
+    set(VCPKG_MANIFEST_DIR "${CMAKE_CURRENT_SOURCE_DIR}/Meta/CMake/vcpkg/host-tools")
```
ADKaster: this should be the responsibility of the invoker of cmake for configuration, rather than embedded in cmakelists. that is, in the preset.

Meta/check-vcpkg.py
```diff
@@ -0,0 +1,175 @@
+#!/usr/bin/env python3
+
+import json
+import os
+import re
+import sys
+
+ROOT_VCPKG = "vcpkg.json"
```
ADKaster, line 10: the check_flatpak.py script also validates dependencies b/w different json files match. Would it make sense to rename that file and move this validation to that script?

ADKaster:
Also of note, Meta/Lagom/BuildFuzzers.sh does not use the new host tools preset.

AK/Function.h
cqundefine: Right, I don't have an opinion on removing it outright, just thought it went unused. My main issue is the weird BlocksRuntime code and the fact that while Objective-C works well on Linux and MacOS, I had issues with it on FreeBSD. So maybe having the code just Apple only would work?

CMakeLists.txt
cqundefine: It was being included twice, it's included already at the top of the file.

Meta/check-vcpkg.py
cqundefine: I feel like those two scripts do two distinctly different things and I'm not sure how much stuff could be deduplicated. Having those separate also gives the nice benefit of knowing which of those two is failing. If you feel like they should be merged I can do that but if you were to ask me I would keep those separate.

cqundefine:
> Also of note, Meta/Lagom/BuildFuzzers.sh does not use the new host tools preset.

Updated the script to use that preset. I also did the same thing with the Fuzzers CI because I noticed that was doing it the old way too.

AK/Function.h
trflynn89: I haven't used it in obj-c but tbh I never actually knew I could!

Meta/CMake/vcpkg/host-tools/vcpkg.json
```diff
@@ -0,0 +1,72 @@
+{
+  "builtin-baseline": "48cfe1e0e928341709d97fc3d2eff10ad6262c96",
```
trflynn89: I am not totally sold on having a second vcpkg.json. We're actively moving towards removing the host tools build by porting our generators to python. Adding this in the meantime feels like it's going to annoy developers more than its worth.

cqundefine: I did it only because @ADKaster suggested it, I have no strong opinion about it. Before we remove the host tools build it will still take a while, the IPCCompiler, all the LibWeb generators, and now we have asmintgen which is Rust so it's probably still easier, but again if you think I should revert that change I have no issue with that.

Meta/check-vcpkg.py
ADKaster: That's fine with me.

Meta/CMake/vcpkg/host-tools/vcpkg.json
ADKaster: We can yeet this part. The main motivation was building tons of extra host vcpkg packages from android studio when working on that port a couple weeks ago.

AK/Function.h
ADKaster: Since it's causing problems for other BSDs, let's guard this on apple only then. If the cmake checks make it confusing, a simple `if (APPLE) target_compile_defintions(AK PUBLIC BLOCKS_ARE_PRESENT)` or some such wouldn't be the worst.

I think it's likely that spreading some AK::Function around the AppKit port could make some block hoops less crazy. But i'm not sure how the capture lists and `__weak` work together. it might be fine.

cqundefine: Right, I reverted that commit and just locked the tests behind a `if (APPLE)` and gated the AK defined behind `defined(AK_OS_MACOS)`.

Meta/CMake/vcpkg/host-tools/vcpkg.json
cqundefine: Got rid of it! I still kept the vcpkg linter as I believe it still can be useful, although now to much lesser degree.

Meta/Linters/check_vcpkg.py
```diff
@@ -0,0 +1,124 @@
+#!/usr/bin/env python3
```
ADKaster, line 1: commit message: still references multiple vcpkg.json files.

ADKaster (approved):
one housekeeping thing, then lgtm

PR #9016, opened 2026-04-21
LibWeb: Implement autofocus candidate processing

tcl3:
This change implements the algorithms necessary to focus elements with the autofocus attribute on page load.

Automatically focuses the search area on https://google.com.

### DIFF
--- a/Libraries/LibWeb/DOM/Document.cpp
+++ b/Libraries/LibWeb/DOM/Document.cpp
@@ -597,6 +597,7 @@ void Document::visit_edges(Cell::Visitor& visitor)
     visitor.visit(m_focused_area);
     visitor.visit(m_active_element);
     visitor.visit(m_target_element);
+    visitor.visit(m_autofocus_candidates);
     visitor.visit(m_implementation);
     visitor.visit(m_current_script);
     visitor.visit(m_associated_inert_template_document);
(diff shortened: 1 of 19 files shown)


Libraries/LibWeb/HTML/HTMLOrSVGElement.cpp
```diff
@@ -108,6 +111,41 @@ void HTMLOrSVGElement<ElementBase>::inserted()
...
+        if (!target.allow_focus())
+            return;
+
+        // 6. Let topDocument be target's node navigable's top-level traversable's active document.
+        auto top_document = target.navigable()->top_level_traversable()->active_document();
+
+        // 7. If topDocument's autofocus processed flag is false, then remove the element from topDocument's autofocus
+        //    candidates, and append the element to topDocument's autofocus candidates.
+        if (!top_document->autofocus_processed_flag()) {
+            auto& candidates = top_document->autofocus_candidates();
```
shannonbooth, line 144: Even though this follows spec closer, I wonder if instead of exposing this helper which can be copied and be GC unsafe, maybe we should expose the remove first matching and append as a helper of document instead?

Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -2928,6 +2929,106 @@ void Document::set_target_element(GC::Ptr<Element> element)
...
+        if (!doc.is_fully_active()) {
+            candidates.take_first();
+            continue;
+        }
+
+        // 4. If doc's node navigable's top-level traversable is not the same as topDocument's node navigable, then
+        //    remove element from candidates, and continue.
+        auto doc_navigable = doc.navigable();
+        if (!doc_navigable || doc_navigable->top_level_traversable() != navigable()) {
+            candidates.take_first();
```
shannonbooth, line 2976: This usage pattern makes me think that perhaps a Vector is not the correct data structure here

shannonbooth (commented):
oo I remember wondering why this didn't work yesterday

Libraries/LibWeb/HTML/HTMLOrSVGElement.cpp
shannonbooth: Actually, maybe not, thinking out loud the copying problem is more general since it could happen within Document itself too. I suppose what we really need is a non copyable Vector type, but that's a bit out of scope here!

Libraries/LibWeb/DOM/Document.cpp
tcl3: Given the typically small number of autofocus candidates, I think the performance hit from repeated `take_first()`s is likely to be minimal most of the time. A deque/ringbuffer would be the ideal fit here, but we don't have an implementation in AK as far as I can see.

FWIW, it looks like Blink uses a Vector here: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/dom/document.cc;l=9162

PR #9020, opened 2026-04-21
LibWeb: Keep select button text in sync with the selected option

tcl3:
Previously, the select button's text was only refreshed inside the two non-trivial branches of the selectedness setting algorithm. Paths that left the select with exactly one selected option hit a no-op branch and skipped the refresh.

Fix this by implementing the "clone selected option into select button" algorithm and invoking it whenever the set of selected options may have changed.

This fixes the pronunciation selector on https://wordreference.com.

From https://www.wordreference.com/esen/ejemplo:

Before:

image

After:

image

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLSelectElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLSelectElement.cpp
@@ -336,7 +336,7 @@ WebIDL::Long HTMLSelectElement::selected_index() const
 WebIDL::ExceptionOr<void> HTMLSelectElement::set_selected_index(WebIDL::Long index)
 {
     // The selectedIndex setter steps are:
-    ScopeGuard guard { [&]() { update_inner_text_element(); } };
+    ScopeGuard guard { [&]() { clone_selected_option_into_select_button(); } };
 
     // 1. Let firstMatchingOption be null.
     GC::Ptr<HTMLOptionElement> first_matching_option;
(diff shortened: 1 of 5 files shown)


Libraries/LibWeb/HTML/HTMLSelectElement.cpp
```diff
@@ -733,28 +734,38 @@ void HTMLSelectElement::create_shadow_tree_if_needed()
...
-    // Update inner text element to the label of the selected option
-    for (auto const& option_element : m_cached_list_of_options) {
-        if (option_element->selected()) {
-            m_inner_text_element->string_replace_all(Infra::strip_and_collapse_whitespace(Utf16String::from_utf8(option_element->label())));
-            return;
+    update_cached_list_of_options();
+
+    // 1. Let option be the first element of select's option list whose selectedness is set to true,
+    //    if such an element exists; otherwise null.
+    GC::Ptr<HTMLOptionElement> option;
```
trflynn89: nit: up to you, but you could reduce this down to:
```c++
auto option = find_value(m_cached_list_of_options, [](auto option) { return option->selected(); });
```
Then step 3 has to be:
```c++
if (option.has_value())
    text = Utf16String::from_utf8((*option)->label());
```

tcl3: Done, that's nicer!

trflynn89 (approved):

PR #9021, opened 2026-04-21
LibWeb: Prevent dragstart after a prevented mousedown or prior dragstart

Zaggy1024:
Also, explicitly prevent drag events from firing when the context menu opens. This will only be the case on macOS, since its context menu is opened by Ctrl+mousedown.

The Ctrl modifier no longer prevents drags on other platforms.

Fixes the PDF flipbook drag interactions at https://www.sawatdee.com/minneapolis-menu.

### DIFF
--- a/Libraries/LibWeb/Page/EventHandler.cpp
+++ b/Libraries/LibWeb/Page/EventHandler.cpp
@@ -974,11 +974,10 @@ void EventHandler::maybe_show_context_menu(GC::Ref<DOM::Node> node, MouseEventCo
 
 void EventHandler::clear_mousedown_tracking()
 {
-    m_mousedown_button = {};
-    m_mousedown_modifiers = UIEvents::KeyModifier::Mod_None;
     m_mousedown_target = nullptr;
     m_mousedown_visual_viewport_position = {};
     m_mousedown_click_count = 0;
+    m_mousedown_target_is_drag_candidate = false;
 }
 
 void EventHandler::stop_updating_selection()
@@ -1016,7 +1015,7 @@ EventResult EventHandler::handle_mouseup(CSSPixelPoint visual_viewport_position,
     });
 
     if (should_ignore_device_input_event()) {
-        if (is_dragging_element()) {
+        if (m_mousedown_target_is_drag_candidate) {
             auto result = handle_drag_and_drop_event(DragEvent::Type::Drop, visual_viewport_position, screen_position, button, buttons, modifiers, {});
             set_page_cursor(m_navigable->page(), Gfx::StandardCursor::Arrow);
             return result;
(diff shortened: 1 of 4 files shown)


Libraries/LibWeb/Page/EventHandler.cpp
```diff
@@ -2146,8 +2144,7 @@ bool EventHandler::is_dragging_element() const
 {
     return m_mousedown_target
         && m_mousedown_visual_viewport_position.has_value()
-        && m_mousedown_button == UIEvents::MouseButton::Primary
-        && (m_mousedown_modifiers & UIEvents::KeyModifier::Mod_Ctrl) == 0;
+        && m_mousedown_button == UIEvents::MouseButton::Primary;
```
trflynn89: These commits aren't atomic in that as of this commit, ctrl+clicking on a draggable element on macOS will both open a context menu and initiate dragging.

It is especially bad because the mouseup event does not fire after a ctrl+mousedown (which happens on other browsers too). So the effect is ctrl+clicking, exiting the menu, then moving the mouse without the mouse pressed at all drags things around:

https://github.com/user-attachments/assets/24ebb734-334e-4f0a-8195-9bc21491714a

Libraries/LibWeb/Page/EventHandler.cpp
```diff
@@ -1420,6 +1421,9 @@ EventResult EventHandler::handle_mousemove(CSSPixelPoint visual_viewport_positio
             document->update_layout(DOM::UpdateLayoutReason::EventHandlerHandleMouseMove);
             if (!paint_root())
                 return EventResult::Accepted;
+
+            if (result == EventResult::Cancelled)
+                m_mousedown_target_is_drag_candidate = false;
```
trflynn89: Not sure if it's an issue in practice, but just in case, I'm thinking this should be done just after checking if `result == Handled` above.

Libraries/LibWeb/Page/EventHandler.h
```diff
@@ -149,17 +148,18 @@ class WEB_API EventHandler {
...
     int m_mousedown_click_count { 0 };
+    bool m_mousedown_target_is_drag_candidate { false };
 
     // https://w3c.github.io/pointerevents/#the-pointerdown-event
     // The PREVENT MOUSE EVENT flag.
     // FIXME: This should be per-pointer, of which there can be multiple. Move it once multiple simultaneous pointer
     //        inputs are supported.
     bool m_prevent_mouse_event { false };
 
+
```
trflynn89: extra newline here

trflynn89 (commented):
Can we add tests for the fixed issues?
1. mousedown being cancelled prevents dragging
2. Cancelled dragstart events don't get re-triggered

Libraries/LibWeb/Page/EventHandler.cpp
Zaggy1024: Ah, I hadn't thought about it starting a drag after the context menu ate the mouseup, whoops. I'll squash the commits and add that into the summary.

Zaggy1024:
Added code to `drag-and-drop-element.html` to exercise those cases.

trflynn89 (approved):

PR #9024, opened 2026-04-22
LibGfx: Expose the imported Linux dma-buf handle on SharedImageBuffer

penkchen:
Enables zero-copy GPU imports of the buffer on the UI side.

### DIFF
--- a/Libraries/LibGfx/SharedImageBuffer.h
+++ b/Libraries/LibGfx/SharedImageBuffer.h
@@ -8,6 +8,7 @@
 
 #include <AK/Noncopyable.h>
 #include <AK/NonnullRefPtr.h>
+#include <AK/Optional.h>
 #include <LibGfx/Bitmap.h>
 #include <LibGfx/SharedImage.h>
 
@@ -34,6 +35,9 @@ public:
 
 #ifdef AK_OS_MACOS
     Core::IOSurfaceHandle const& iosurface_handle() const { return m_iosurface_handle; }
+#elif defined(USE_VULKAN_DMABUF_IMAGES)
+    // Non-empty for dma-buf-backed buffers; lets UI code zero-copy-import the fd.
+    Optional<LinuxDmaBufHandle> const& linux_dmabuf_handle() const { return m_linux_dmabuf_handle; }
 #endif
 
 private:
@@ -42,6 +46,10 @@ private:
     Core::IOSurfaceHandle m_iosurface_handle;
 #else
     explicit SharedImageBuffer(NonnullRefPtr<Bitmap>);
+#    ifdef USE_VULKAN_DMABUF_IMAGES
+    SharedImageBuffer(LinuxDmaBufHandle&&, NonnullRefPtr<Bitmap>);
+    Optional<LinuxDmaBufHandle> m_linux_dmabuf_handle;
+#    endif
 #endif
     NonnullRefPtr<Bitmap> m_bitmap;
 };
--- a/Libraries/LibGfx/CMakeLists.txt
+++ b/Libraries/LibGfx/CMakeLists.txt
@@ -148,5 +148,7 @@ if (HAS_VULKAN)
     if (USE_VULKAN_DMABUF_IMAGES)
         pkg_check_modules(LibDRM REQUIRED libdrm)
         target_include_directories(LibGfx PUBLIC ${LibDRM_INCLUDE_DIRS})
+        # PUBLIC so consumers see the same Gfx::SharedImageBuffer layout.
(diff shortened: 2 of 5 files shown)


Libraries/LibGfx/SharedImageBuffer.h
```diff
@@ -42,6 +47,8 @@ class SharedImageBuffer {
     Core::IOSurfaceHandle m_iosurface_handle;
 #else
     explicit SharedImageBuffer(NonnullRefPtr<Bitmap>);
+    SharedImageBuffer(LinuxDmaBufHandle&&, NonnullRefPtr<Bitmap>);
```
jdahlin, line 50: I believe this file compiles under windows as well, which will probably not be happy about this.

Libraries/LibGfx/SharedImageBuffer.h
```diff
@@ -42,6 +47,8 @@ class SharedImageBuffer {
...
+    Optional<LinuxDmaBufHandle> m_linux_dmabuf_handle;
```
jdahlin, line 51: Is it desirable to enable this for all shared image buffers on linux when vulkan is present? Ie avoid optional.

Libraries/LibGfx/SharedImageBuffer.h
penkchen: Thanks, I'll guard it with USE_VULKAN_DMABUF_IMAGES

Libraries/LibGfx/SharedImageBuffer.h
penkchen: but, 

- the dma-buf path in BackingStoreManager only kicks in for is_top_level_traversable()
- sub-frames and Vulkan-init fallbacks use Bitmap::create_shareable 

so the buffer really can be either. Making it non-optional means also routing sub-frames (and the fallback) through Vulkan, which is a BackingStoreManager change rather than a LibGfx one

is that the direction you'd like, or is the Optional acceptable for this PR?

jdahlin: Not sure! @kalenikaliaksandr wrote most of this code and he would probably have to review it.

kalenikaliaksandr (changes requested):
hey! could you make this commit a part of PR where introduced `linux_dmabuf_handle()` is actually used? as is the change only introduces dead code without explanation how exactly it's going to be consumed by future changes.

penkchen:
@kalenikaliaksandr thanks for the comment. the intended consumer was going to be a new Qt Quick frontend, but that got rejected separately. 

Should I expand it with a UI/Gtk consumer that uses GdkDmabufTextureBuilder to import the handle directly? that seems to be the most direct way to show the intended zero-copy path and stays within an existing frontend.

jdahlin:
> @kalenikaliaksandr thanks for the comment. the intended consumer was going to be a new Qt Quick frontend, but that got rejected separately.
> 
> Should I expand it with a UI/Gtk consumer that uses GdkDmabufTextureBuilder to import the handle directly? that seems to be the most direct way to show the intended zero-copy path and stays within an existing frontend.

That would be nice, its AFAICT not a very big refactoring as Gtk4 already exposes https://docs.gtk.org/gdk4/class.DmabufTexture.html

kalenikaliaksandr:
> @kalenikaliaksandr thanks for the comment. the intended consumer was going to be a new Qt Quick frontend, but that got rejected separately.
> 
> Should I expand it with a UI/Gtk consumer that uses GdkDmabufTextureBuilder to import the handle directly? that seems to be the most direct way to show the intended zero-copy path and stays within an existing frontend.

yes, seeing how this code is used by UI/Gtk consumer would be very helpful.

penkchen:
Added a UI/Gtk consumer that wires linux_dmabuf_handle() into GdkDmabufTextureBuilder. Verified locally on Wayland/Vulkan, with `GDK_DEBUG=dmabuf GSK_DEBUG=renderer GSK_RENDERER=vulkan`:

```
Initialization finished. Advertising 409 dmabuf formats
Initialization invoking finished
Creating dmabuf texture, format AR24:0,  premultiplied, 1 planes, memory format BGRA8p
Vulkan uploaded 1280x886 AR24:0000000000000000 dmabuf
Creating dmabuf texture, format AR24:0,  premultiplied, 1 planes, memory format BGRA8p
Vulkan uploaded 1024x630 AR24:0000000000000000 dmabuf
[...]
```

penkchen:
Rebase && resolve conflict. 

@kalenikaliaksandr any chance to get a review? :-)

gmta:

PR #9026, opened 2026-04-22
Build: Use one rust runtime per executable or shared library

jonbgamble:
Stop building LibJS, LibGfx, LibUnicode, and LibRegex as separate staticlibs exporting their own bundled rust runtime and allocator. Build them as rlibs instead, and generate a per-target runtime staticlib that pulls in LibRustRuntime's allocator shim.

This keeps one copy of the rust runtime per dso or executable and avoids duplicate allocator and std symbols. It also fixes a debug-link mess from pulling multiple rust staticlibs into the same process, causing elf debug builds of WebContent with mimalloc to insta-die on any regex.

Build details:

A library imports its rust crate as an rlib rather than staticlib, and Meta/rust_crate.cmake's import_rust_ffi_crate function registers that crate on the cmake target as metadata. Later, link_rust_runtime turns the collected crate list into a generated per-target staticlib crate in the build_dir which pulls in allocator.rs from LibRustRuntime, reexports the component crates, and links exactly one rust runtime bundle into the native link target.

To hook in a new rust-backed library, make the crate an rlib in Cargo.toml, then call the import_rust_ffi_crate function from the library's CMakeLists.txt. Example from LibRegex:
```
import_rust_ffi_crate(LibRegex MANIFEST_PATH Rust/Cargo.toml CRATE_NAME libregex_rust_core FFI_HEADER RustFFI.h)
```
This gives the c++ side its generated ffi header, registers the crate metadata for later codegen, and, in shared builds, gives that shared library its own local allocator shim and runtime.

An executable uses it at the final native link boundary. In static builds, executables call link_rust_runtime which walks the linked libraries, finds registered rust crates transitively, and builds one runtime. Shared libraries get private rust runtimes when they import their crates, so an executable just links those normally.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -66,9 +66,9 @@ checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
 name = "bitflags"
-version = "2.11.0"
+version = "2.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "843867be96c8daad0d758b57df9392b6d8d271134fce549de6ce169ff98a92af"
+checksum = "c4512299f36f043ab09a583e57bceb5a5aab7a73db1805848e8fef3c9e8c78b3"
 
 [[package]]
 name = "bytecode_def"
@@ -111,9 +111,9 @@ checksum = "9330f8b2ff13f34540b44e946ef35111825727b38d33286ef986142615121801"
 
 [[package]]
 name = "clap"
-version = "4.6.0"
+version = "4.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b193af5b67834b676abd72466a96c1024e6a6ad978a1f484bd90b85c94041351"
+checksum = "1ddb117e43bbf7dacf0a4190fef4d345b9bad68dfc649cb349e7d17d28428e51"
 dependencies = [
  "clap_builder",
 ]
@@ -180,9 +180,9 @@ dependencies = [
 
 [[package]]
 name = "fastrand"
-version = "2.3.0"
+version = "2.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "37909eebbb50d72f9059c3b6d82c0463f2ff062c9e95845c43a6c9c0355411be"
+checksum = "9f1f227452a390804cdb637b74a86990f2a7d7ba4b7d5693aac9b4dd6defd8d6"
 
 [[package]]
 name = "foldhash"
@@ -214,9 +214,9 @@ dependencies = [
 
 [[package]]
 name = "hashbrown"
-version = "0.16.1"
+version = "0.17.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "841d1cc9bed7f9236f321df977030373f4a4163ae1a7dbfe1a51a2c1a51d9100"
+checksum = "4f467dd6dccf739c208452f8014c75c18bb8301b050ad1cfb27153803edb0f51"
 
 [[package]]
 name = "heck"
@@ -321,12 +321,12 @@ checksum = "3d3067d79b975e8844ca9eb072e16b31c3c1c36928edf9c6789548c524d0d954"
 
 [[package]]
 name = "indexmap"
-version = "2.13.0"
+version = "2.14.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7714e70437a7dc3ac8eb7e6f8df75fd8eb422675fc7678aff7364301092b1017"
+checksum = "d466e9454f08e4a911e14806c24e16fba1b4c121d1ea474396f396069cf949d9"
 dependencies = [
  "equivalent",
- "hashbrown 0.16.1",
+ "hashbrown 0.17.0",
  "serde",
  "serde_core",
 ]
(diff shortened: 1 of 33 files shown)


Libraries/LibRustRuntime/src/lib.rs
```diff
@@ -0,0 +1,12 @@
...
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+mod allocator;
+
+pub use libgfx_rust_core as gfx;
+pub use libjs_rust_core as js;
+pub use libregex_rust_core as regex;
+pub use libunicode_rust_core as unicode;
```
ADKaster: So, this specific crate is pretty ... Let's call it *controversial*. Separating each library's rust components is an intentional  design choice. One I quite like, and I presume other maintainers like as well. 

Bundling every rust crate into one mega crate makes it very likely to include rust libs and even c++ libs into targets that have no need for them. e.g. the ja repl has no need for libgfx or any image processing libraries.

ADKaster (changes requested):
Separating the allocator into one static lib for each executable that "uses rust" to link-once? Yes, I enjoy this idea. 

The part of creating one mega rust lib? Words cannot express how much I dislike this approach.

Libraries/LibRustRuntime/src/lib.rs
jonbgamble: There is some new CMake spew in the version I just pushed, but assuming folks can follow examples in the existing libraries, it shouldn't be intrusive or need changing.

ADKaster: without looking at it yet, the answer to "buncha cmake goo" is often "add a function to utils.cmake"

ADKaster:
I'm a little confused by your solution still. 

As far as I'm concerned, the problem worth solving is "link global allocators once". Which could be solved by:

1) create a static lib for rust allocator symbols
2) apply a custom property to c++ libs that use rust in the impl
3) check for said property in ladybird_bin/ladybird_app to pull in the rust allocator into the exe.

That seems to be about 20% of this patch. Why is the rest of this needed?

jonbgamble:
I think your objection is fair. The broader part of this patch is not just about allocator symbols, it changes the static build boundary from “multiple independently built rust staticlibs get merged into one elf” to “cargo resolves one rust closure per final native link target”.

That said, the allocator issue is the most concrete failure I can point to right now. If you want the fix scoped strictly to that, I can split this down to an allocator-only patch and leave the per-target runtime cleanup for later.

ADKaster:
The allocator problem is pretty concrete, so let's fix that first and clearly. The ideas for crate reorganization need some time to understand and think about. I've heard opinions from several angles on where to draw the language/shlib/staticlib boundary for components that need some exploring.

For developer builds, it's a bit more yolo than for distribution builds, imo. With alpha deadlines approaching slowly, something that won't cause distros to recoil in horror would probably be a good idea.

jonbgamble:
All I know is what I've read at https://doc.rust-lang.org/stable/reference/linkage.html#mixed-rust-and-foreign-codebases). The people you're talking to may know more, maybe that doc is out of date, but it says that a foreign linker should link at most one staticlib.

ADKaster:
I hadn't seen that advice before. Is this for real? 

> Use your foreign linker. In this case, you first need to generate a Rust staticlib target and pass that into your foreign linker invocation. If you need to link multiple Rust subsystems, you will need to generate a single staticlib perhaps using lots of extern crate statements to include multiple Rust rlibs. Multiple Rust staticlib files are likely to conflict.

That doesn't compose at all. That doesn't work for multi executable projects with shared, but distinct dependencies.

I can start to see the point of your generated static lib crate now. But that's still not at all how a well behaved dependency for a  C++ library should work.

jonbgamble:
I suppose it wouldn't be the first time an online reference started out overly cautious, transitioned beyond questionable, and ended up wrong.

ADKaster:
If we do it like that doc suggests, then we're completely negating the point of having our libraries be shared. If I have to link a rust static lib containing the rust implementation details of every *shared* C++ library into each executable, something has seriously gone wrong with the build. That breaks encapsulation, it breaks the fundamental property of shared libs that they should be replaceable without relinking the main executable, and it means that the implementation details are now part of the public API.

That's just patently ridiculous.

We should either ditch shared linking entirely, which is a bad idea for disk space, link times, and dev cycle times, or figure out why that doc suggests that, and whether it actually applies to shared libraries in the first place. 

The global allocator itself is independent from all that nonsense. It's pretty clear to me that it should work just like global new/delete. You put a strong symbol in the main exe that ends up being used by every shlib you link in later. And then you plug your fingers in your ear and pretend windows doesn't exist because that sort of weak override doesn't work there with all the library isolation they do.

jonbgamble:
> If I have to link a rust static lib containing the rust implementation details of every shared C++ library into each executable,

That was not my takeaway, I think that part of the doc is talking about static linkage. For shared builds, everything I've read points to foreign linkers being free to link one staticlib per executable or DSO. A foreign linked executable may link one rust staticlib, and each of its shared libraries may have a staticlib as well. So there's no limit to how many rust libraries you can have and they can all contain different things, they just have to be isolated within separate DSOs if we believe the rust-lang guidance.

Those static link interop constraints are frustrating (assuming they're real). But maybe tomorrow, you'll wake up and `one staticlib full of rlibs for a static build` and `up to one staticlib per executable / DSO` for shared builds won't seem so terrible.

If my current understanding turns out to be correct, then I think static builds will lead you towards something like this PR. Either build uber staticlibs dynamically from a buffet of rlibs as I did, or spell out each native link target explicitly in Cargo/CMake as crates of sources, rlibs, fully described and wrapped in an staticlib.

For the crashing debug issue on elf, I think it's a bizarre sequence of events when building liblagom-regex.so where liblagom-regex.a has an allocator shim and so does its embedded liblagom-unicode.a and (not sure why) this somehow results in liblagom-regex.so saying fuck it, I don't need an allocator. I'll just ask the loader at runtime. Lucky for noone, the loader is happy to supply __rust_alloc from liblagom-gfx.so, so now the fuse is set as soon as an alloc/free happens across the libregex boundary. Hypothetically if we were to give the allocator shim to libgfx_rust, there would no longer be a vanilla __rust_alloc provider in any of our crates so the debug crash would likely stop.

PR #9030, opened 2026-04-22
LibWeb: Skip unstyled descendants in animation updates

tcl3:
Animation updates propagate inherited animated values by walking the animated target's subtree and calling `recompute_inherited_style()` on each element. Elements inserted during the same rendering update may not have computed properties yet, which violates an existing assertion, causing a crash.

Fix this by skipping unstyled descendants during this walk. The subsequent recursive style update computes their style from scratch.

This fixes an intermittent crash I encountered when logging in to https://sainsburys.co.uk.

NB: The assertion this crash hits was added as part of #9014 (merged yesterday - 2026-04-21). An alternative fix would be to revert 06a7723ee39b484a2588f12ea0d34eebb953ba56.


InvalidUsernameException:
Thanks.

PR #9031, opened 2026-04-22
LibCore: Allow zero-size AnonymousBuffer creation

tcl3:
Previously, `AnonymousBuffer::create_with_size(0)` returned an error because POSIX `mmap` rejects a zero length with `EINVAL`, and Windows `CreateFileMapping` rejects a zero maximum size for an anonymous
mapping. This caused a crash when using `--headless=text` with zero size pages like `about:blank`.

I initially encountered this when running `./Meta/ladybird.py run ladybird --headless=text about:blank`.

### DIFF
--- a/Libraries/LibCore/AnonymousBufferWindows.cpp
+++ b/Libraries/LibCore/AnonymousBufferWindows.cpp
@@ -30,7 +30,10 @@ AnonymousBufferImpl::~AnonymousBufferImpl()
 
 ErrorOr<NonnullRefPtr<AnonymousBufferImpl>> AnonymousBufferImpl::create(size_t size)
 {
-    HANDLE map_handle = CreateFileMapping(INVALID_HANDLE_VALUE, NULL, PAGE_READWRITE, size >> 32, size & 0xFFFFFFFF, NULL);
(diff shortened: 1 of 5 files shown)


Libraries/LibCore/AnonymousBufferWindows.cpp
```diff
@@ -30,7 +30,10 @@ AnonymousBufferImpl::~AnonymousBufferImpl()
 
 ErrorOr<NonnullRefPtr<AnonymousBufferImpl>> AnonymousBufferImpl::create(size_t size)
 {
-    HANDLE map_handle = CreateFileMapping(INVALID_HANDLE_VALUE, NULL, PAGE_READWRITE, size >> 32, size & 0xFFFFFFFF, NULL);
+    // CreateFileMapping rejects a zero-size mapping backed by INVALID_HANDLE_VALUE, so allocate at least one byte and
+    // track the logical size separately.
+    auto map_size = max(size, size_t { 1 });
```
trflynn89: In the blessed year of 2026, we have: `1uz`

tcl3: We're living in an age of literal wonders!

trflynn89 (approved):

PR #9032, opened 2026-04-22
Reduce redundant @font-face fetches

jdahlin:
On a fresh YouTube load this brings the initial `@font-face` download count from 177 to 9 or so.

- Defer fetches until a codepoint renders (honor the unicode-range descriptor).
- Only trigger `@font-face` loads from text shaping
- Honor text argument of FontFaceSet.load() / .check(), implement step 8 of find the matching font faces: remove faces whose unicode-range doesn't cover any codepoint in the text argument. Without this a single `document.fonts.load('14px Roboto')` pulled every declared subset under the family.
- Dedup `@font-face` fetches by source URL using a shared FontLoader cache keyed by first src URL.

Adds a couple of test to verify this behavior.

### DIFF
--- a/Libraries/LibWeb/CSS/FontFaceSet.cpp
+++ b/Libraries/LibWeb/CSS/FontFaceSet.cpp
@@ -379,6 +379,11 @@ void FontFaceSet::set_is_pending_on_the_environment(bool is_pending_on_the_envir
         //        Spec issue: https://github.com/w3c/csswg-drafts/issues/13538#issuecomment-3933951987
         if (m_set_entries->set_size() == 0 || (m_is_stuck_on_the_environment && m_loading_fonts.is_empty()))
             switch_to_loaded();
+        // AD-HOC: Also switch when nothing has ever entered the LoadingFonts list — an empty set, or a set whose
+        //         entries all have deferred unicode-ranges that no rendered codepoint matched. Without this the
+        //         ready promise stays pending forever.
+        else if (m_loading_fonts.is_empty())
+            switch_to_loaded();
 
         // 2. If the FontFaceSet is stuck on the environment, unmark it as such.
         m_is_stuck_on_the_environment = false;
--- a/Libraries/LibWeb/CSS/FontFaceSet.cpp
+++ b/Libraries/LibWeb/CSS/FontFaceSet.cpp
@@ -203,7 +203,7 @@ WebIDL::CallbackType* FontFaceSet::onloadingerror()
 }
 
 // https://drafts.csswg.org/css-font-loading/#find-the-matching-font-faces
-static WebIDL::ExceptionOr<GC::Ref<JS::Set>> find_matching_font_faces(JS::Realm& realm, FontFaceSet& font_face_set, String const& font, String const&)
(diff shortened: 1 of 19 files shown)


Libraries/LibWeb/CSS/FontFaceSet.cpp
```diff
@@ -375,10 +395,16 @@ void FontFaceSet::set_is_pending_on_the_environment(bool is_pending_on_the_envir
     if (!is_pending_on_the_environment) {
         // 1. If the FontFaceSet is stuck on the environment and its [[LoadingFonts]] list is empty, switch the
         //    FontFaceSet to loaded.
-        // FIXME: We also need to mark empty FontFaceSets as loaded, so that the [[ReadyPromise]] gets resolved.
-        //        Spec issue: https://github.com/w3c/csswg-drafts/issues/13538#issuecomment-3933951987
-        if (m_set_entries->set_size() == 0 || (m_is_stuck_on_the_environment && m_loading_fonts.is_empty()))
+        if (m_is_stuck_on_the_environment && m_loading_fonts.is_empty()) {
+            switch_to_loaded();
+        }
+        // Spec extension: also switch when nothing has ever entered [[LoadingFonts]] — an empty set, or a set
```
Psychpsyo: What is a spec extension and shouldn't it just say `AD-HOC`?

Libraries/LibWeb/CSS/FontFaceSet.cpp
```diff
@@ -250,8 +250,28 @@ static WebIDL::ExceptionOr<GC::Ref<JS::Set>> find_matching_font_faces(JS::Realm&
     }
 
     // FIXME: 7. If matched font faces is empty, set the found faces flag to false. Otherwise, set it to true.
-    // FIXME: 8. For each font face in matched font faces, if its defined unicode-range does not include the codepoint of at
-    //           least one character in text, remove it from the list.
+
+    // 8. For each font face in matched font faces, if its defined unicode-range does not include the codepoint of at
+    //    least one character in text, remove it from the list.
+    if (!text.is_empty()) {
```
Psychpsyo: This doesn't match the spec, which explicitly says that every font should be removed for the empty string. Why are we not doing that here?

Libraries/LibWeb/CSS/FontFace.cpp
```diff
@@ -726,6 +726,22 @@ GC::Ref<WebIDL::Promise> FontFace::load()
     if (m_css_font_face_rule)
         m_css_font_face_rule->set_loading_state(CSSStyleSheet::LoadingState::Loading);
 
+    // Keep the owning FontFaceSets' [[LoadingFonts]] list in sync so document.fonts.ready
```
Psychpsyo: This isn't in the spec here, so should probably say `NB:` or `AD-HOC:`, depending on whether it's an implementation detail or issue with the spec. If it's an issue with the spec, it should probably also link to more details about that.

Also, if this is an implementation-only thing, I'm not sure using the spec's double bracket notation for stuff is a good idea. At least to me, that makes it feel official in a way that it isn't.
(not a maintainer though so that point might be moot.)

PR #9036, opened 2026-04-22
Reuse cached UTF-16 in Array.prototype.sort's string comparator

kalenikaliaksandr:
CompareArrayElements was calling ToString(x) +
PrimitiveString::create(vm, ...)on every comparison, producing a fresh PrimitiveString that wrapped the original's AK::String but carried no cached UTF-16. The subsequent IsLessThan then hit
PrimitiveString::utf16_string_view() on that fresh object, which re-ran simdutf UTF-8 validation + UTF-8 → UTF-16 conversion for both sides on every one of the N log N comparisons.

When x and y are already String Values, ToString(x) and ToPrimitive(x, Number) are the identity per spec, so we can forward the original Values straight to IsLessThan. The original PrimitiveString caches its UTF-16 on first access, so subsequent comparisons against the same element hit the cache instead of re-transcoding.

Microbenchmark:
```js
function makeStrings(n) {
    let seed = 1234567;
    const rand = () => {
        seed = (seed * 1103515245 + 12345) & 0x7fffffff; return seed;
    };
    const out = new Array(n);
    for (let i = 0; i < n; i++)
        out[i] = "item_"+rand().toString(36)+"_"+rand().toString(36);
    return out;
}
const base = makeStrings(100000);
const arr = base.slice();
arr.sort();
```

```
n       before  after   speedup
1k      0.70ms  0.30ms  2.3x
10k     8.33ms  4.00ms  2.1x
50k    49.33ms 28.33ms  1.7x
100k  118.00ms 76.00ms  1.55x
```

### DIFF
--- a/AK/Utf16View.h
+++ b/AK/Utf16View.h
@@ -261,8 +261,6 @@ public:
 
         if (has_ascii_storage() && other.has_ascii_storage()) {
             result = __builtin_memcmp(m_string.ascii, other.m_string.ascii, length);
-        } else if (!has_ascii_storage() && !other.has_ascii_storage()) {
-            result = __builtin_memcmp(m_string.utf16, other.m_string.utf16, length * sizeof(char16_t));
         } else {
             for (size_t i = 0; i < length; ++i) {
                 auto this_code_unit = code_unit_at(i);
--- a/Libraries/LibJS/Runtime/Array.cpp
+++ b/Libraries/LibJS/Runtime/Array.cpp
@@ -250,6 +250,12 @@ ThrowCompletionOr<double> compare_array_elements(VM& vm, Value x, Value y, Funct
         return value_number.as_double();
     }
 
+    // OPTIMIZATION: When both operands are already Strings, ToString is the identity, so we can compare their
+    //               UTF-16 views directly. This preserves any cached UTF-16 on the original PrimitiveStrings
+    //               across sort comparisons and skips the ToPrimitive + IsLessThan detour.
+    if (x.is_string() && y.is_string())
+        return x.as_string().utf16_string_view() <=> y.as_string().utf16_string_view();
+
     // 5. Let xString be ? ToString(x).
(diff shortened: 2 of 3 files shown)


Libraries/LibJS/Runtime/Array.cpp
```diff
@@ -250,6 +250,19 @@ ThrowCompletionOr<double> compare_array_elements(VM& vm, Value x, Value y, Funct
         return value_number.as_double();
     }
 
+    // OPTIMIZATION: When both operands are already Strings, ToString is the identity, so we can forward the original
+    //               Values directly to IsLessThan instead of rebuilding PrimitiveStrings. This preserves any cached
+    //               UTF-16 on the original PrimitiveStrings across sort comparisons.
+    if (x.is_string() && y.is_string()) {
+        auto x_smaller = MUST(is_less_than(vm, x, y, true));
```
shannonbooth: Whould we be able to skip a little more work here if we as_string() both of these and directly invoke  is_code_unit_less_than on their utf16_string_view (no need for exception checks or going through the to primitive call)

shannonbooth: And maybe actually even a single call to is_code_unit_less_than too would be possible here

kalenikaliaksandr: done!

shannonbooth: nice! numbers even better after that

AK/Utf16View.h
```diff
@@ -261,8 +261,6 @@ class Utf16View {
 
         if (has_ascii_storage() && other.has_ascii_storage()) {
             result = __builtin_memcmp(m_string.ascii, other.m_string.ascii, length);
-        } else if (!has_ascii_storage() && !other.has_ascii_storage()) {
```
shannonbooth, line 264: Probably worth adding an in tree AK test for this

kalenikaliaksandr: good idea, done!

PR #9041, opened 2026-04-22
RequestServer: Send Cookie header bytes verbatim

kalenikaliaksandr:
notify_retrieved_http_cookie() was running the cookie string through HTTP::Header::isomorphic_encode, which iterates the value by code point and casts each to u8. Cookies in our jar are stored as UTF-8 strings, so any non-ASCII character (e.g. Cyrillic) had every multi-byte code point collapsed to its low byte before being sent back to the server. The corruption was silent aside from a FIXME dbgln.

Cookies on the wire are opaque byte sequences — whatever bytes the server set, we should hand back unchanged. Append the cookie StringView directly to the header list so its UTF-8 bytes pass through as the Cookie header value.

Fixes loading of https://devby.io/


Services/RequestServer/Request.cpp
```diff
@@ -171,8 +171,7 @@ void Request::notify_request_unblocked(Badge<HTTP::DiskCache>)
 void Request::notify_retrieved_http_cookie(Badge<ConnectionFromClient>, StringView cookie)
 {
     if (!cookie.is_empty()) {
-        auto header = HTTP::Header::isomorphic_encode("Cookie"sv, cookie);
-        m_request_headers->append(move(header));
+        m_request_headers->append({ "Cookie"sv, cookie });
```
shannonbooth, line 506: is it possible to add a test using http test server here? there is a bit of weirdness with isomorphic encoding in fetch in other places so i could see this regressing somehow in the future

kalenikaliaksandr: sure, done!!

gmta (approved):

awesomekling:
This breaks on multiple runners:
```
Fail: Text/input/Fetch/fetch-cors-credentials-flag.html

/home/runner/_work/ladybird/ladybird/Tests/LibWeb/Text/expected/./Fetch/fetch-cors-credentials-flag.txt
+++ /home/runner/_work/ladybird/ladybird/Tests/LibWeb/Text/expected/./Fetch/fetch-cors-credentials-flag.txt
@@ -1 +1 @@
-PASS
+FAIL - cookie was not set
```

avbdr:
i think this change would fix loading of cnn.com after the cookies are saved

trflynn89:
Potentially related: #9868

shannonbooth:
https://github.com/LadybirdBrowser/ladybird/pull/8540/changes/13d5eacbac72512a86331f710087ae3c3bdd4eb6 in https://github.com/LadybirdBrowser/ladybird/pull/8540 probably fixes the test issue in this MR since it _should_ isolate cookies to the test (in theory)

PR #9053, opened 2026-04-23
LibRegex+LibJS: Fixes for some /v mode bugs and backtracking issues

aplefull:
First 5 commits are random fixes.

And I noticed that we hit backtracking limits in `v` mode pretty easily because we use lookarounds for quantifiers. I tried to fix this in last two commits by extending the GreedyLoop/LazyLoop emission to cover `v` mode quantified classes. Seems to work for tests that I generate with my hacky fuzzer, but this can definitely be improved to cover even more cases.

### DIFF
--- a/Libraries/LibJS/Runtime/RegExpLegacyStaticProperties.cpp
+++ b/Libraries/LibJS/Runtime/RegExpLegacyStaticProperties.cpp
@@ -198,8 +198,9 @@ ThrowCompletionOr<Value> get_legacy_regexp_static_property(VM& vm, RegExpConstru
     auto val = (constructor.legacy_static_properties().*property_getter)();
 
     // 4. If val is empty, throw a TypeError exception.
+    // NOTE: The spec says to throw here, but all major browsers return "" instead.
     if (!val)
-        return vm.throw_completion<TypeError>(ErrorType::GetLegacyRegExpStaticPropertyValueEmpty);
+        return &vm.empty_string();
 
     // 5. Return val.
     return val;
--- a/Libraries/LibRegex/Rust/src/compiler.rs
+++ b/Libraries/LibRegex/Rust/src/compiler.rs
@@ -1226,7 +1226,7 @@ impl Compiler {
                     self.emit(Instruction::Fail);
                     return;
                 }
-                self.emit_char_maybe_case_fold(*c as u32);
+                self.emit_char_maybe_case_fold(*c);
             }
             ClassSetOperand::Range(lo, hi) => {
                 if length != 1 {
@@ -1234,10 +1234,7 @@ impl Compiler {
                     return;
                 }
                 self.emit(Instruction::CharClass {
-                    ranges: vec![CharRange {
-                        start: *lo as u32,
-                        end: *hi as u32,
-                    }],
+                    ranges: vec![CharRange { start: *lo, end: *hi }],
                     negated: false,
                 });
             }
--- a/Libraries/LibRegex/Rust/src/compiler.rs
+++ b/Libraries/LibRegex/Rust/src/compiler.rs
@@ -501,34 +501,32 @@ impl Compiler {
                 }
             }
             Atom::BuiltinCharacterClass(class) => Some(SimpleMatch::BuiltinClass(*class)),
-            Atom::CharacterClass(cc) => {
-                // Only use simple match for character classes without v-flag set operations.
-                let ranges_vec = match &cc.body {
-                    CharacterClassBody::Ranges(ranges) => ranges,
-                    CharacterClassBody::UnicodeSet(_) => return None,
-                };
-                let mut ranges = Vec::new();
-                for r in ranges_vec {
-                    match r {
-                        CharacterClassRange::Single(c) => {
-                            if *c > 0xFFFF && !self.program.unicode && !self.program.unicode_sets {
-                                return None;
+            Atom::CharacterClass(cc) => match &cc.body {
+                CharacterClassBody::Ranges(ranges_vec) => {
+                    let mut ranges = Vec::new();
+                    for r in ranges_vec {
+                        match r {
+                            CharacterClassRange::Single(c) => {
+                                if *c > 0xFFFF && !self.program.unicode && !self.program.unicode_sets {
+                                    return None;
+                                }
+                                ranges.push(CharRange { start: *c, end: *c });
                             }
-                            ranges.push(CharRange { start: *c, end: *c });
+                            CharacterClassRange::Range(lo, hi) => {
+                                ranges.push(CharRange { start: *lo, end: *hi });
+                            }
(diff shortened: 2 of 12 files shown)


Tests/LibJS/Runtime/3rdparty/v8/regexp-duplicate-named-groups.js
```diff
@@ -84,7 +84,7 @@ function assertArrayEquals(expected, actual) {
     expect(actual).toEqual(expected);
 }
 
-test.xfail("regexp-duplicate-named-groups", () => {
+test("regexp-duplicate-named-groups", () => {
```
alimpfard, line 87: huh, why was this an xfail...?

Libraries/LibJS/Runtime/RegExpLegacyStaticProperties.cpp
```diff
@@ -198,8 +198,9 @@ ThrowCompletionOr<Value> get_legacy_regexp_static_property(VM& vm, RegExpConstru
     auto val = (constructor.legacy_static_properties().*property_getter)();
 
     // 4. If val is empty, throw a TypeError exception.
+    // NOTE: The spec says to throw here, but all major browsers return "" instead.
     if (!val)
-        return vm.throw_completion<TypeError>(ErrorType::GetLegacyRegExpStaticPropertyValueEmpty);
+        return PrimitiveString::create(vm, Utf16String {});
```
awesomekling: `vm.empty_string()` :)

Libraries/LibRegex/Rust/src/parser.rs
```diff
@@ -1682,27 +1682,30 @@ impl Parser {
...
+            'f' => Ok(0x0C),
+            'v' => Ok(0x0B),
+            'b' => Ok(0x08),
+            '0' if !self.peek().is_some_and(|c| c.is_ascii_digit()) => Ok(0),
             'x' => {
                 let val = self.parse_hex_escape(2)?;
-                char::from_u32(val).ok_or(Error::InvalidUnicodeEscape)
+                if val > 0x10FFFF {
+                    return Err(Error::InvalidUnicodeEscape);
+                }
```
awesomekling: Given we're only parsing 2 hex digits, this should be unreachable, right?

Libraries/LibRegex/Rust/src/parser.rs
```diff
@@ -1713,22 +1716,25 @@ impl Parser {
...
                         self.pos = saved;
                     } else {
                         self.pos = saved;
                     }
                 }
-                char::from_u32(val).ok_or(Error::InvalidUnicodeEscape)
+                if val > 0x10FFFF {
+                    return Err(Error::InvalidUnicodeEscape);
+                }
+                Ok(val)
```
awesomekling: Same question here :)

Libraries/LibRegex/Rust/src/parser.rs
```diff
@@ -279,16 +279,25 @@ impl Parser {
...
-            // Reset the current alternative's name set for the next alternative.
-            if let Some(names) = self.alternative_name_stack.last_mut() {
-                names.clear();
+            let committed = self.alternative_name_stack.last().cloned().unwrap_or_default();
+            if let Some(outer) = self.alternative_name_stack.iter_mut().rev().nth(1) {
+                outer.extend(committed.iter().cloned());
+            }
+
+            if let Some(top) = self.alternative_name_stack.last_mut() {
+                *top = inherited.clone();
```
alimpfard, line 296: This chunk was quite confusing to read, and I'm not if it can be written any more clearly in rust...so could you put a tiny comment here explaining what's happening?

Libraries/LibRegex/Rust/src/compiler.rs
```diff
@@ -1083,6 +1081,191 @@ impl Compiler {
...
+        };
+
+        if operands.is_empty() {
+            return None;
+        }
+
+        let matchers: Option<Vec<SimpleMatch>> = operands.iter().map(Self::try_simple_match_for_operand).collect();
+        let matchers = matchers?;
+        Self::build_union_simple_match(matchers)
+    }
```
alimpfard: How about something with less repetition?
```suggestion
    fn try_simple_match_for_unicode_set(&self, expr: &ClassSetExpression, negated: bool) -> Option<SimpleMatch> {
        'simple: {
            if negated { break 'simple; }
            let ClassSetExpression::Union(operands) = expr else {
                break 'simple;
            };

            if operands.is_empty() {
                return None;
            }

            let matchers: Option<Vec<SimpleMatch>> = operands.iter().map(Self::try_simple_match_for_operand).collect();
            let matchers = matchers?;
            return Self::build_union_simple_match(matchers);
        }

        Self::try_extract_union_ranges(expr).map(|ranges| {
            let mut sorted = ranges;
            sorted.sort_by_key(|r| r.start);
            SimpleMatch::CharClass {
                ranges: sorted,
                negated,
            }
        })
    }
```

alimpfard: sheesh, nevermind, next commit takes a sledge hammer to it.

alimpfard (commented):
Still very bad at reading rust, but here's a few things that jumped out.

Libraries/LibJS/Runtime/RegExpLegacyStaticProperties.cpp
aplefull: nice, thank you!

Tests/LibJS/Runtime/3rdparty/v8/regexp-duplicate-named-groups.js
aplefull: Yeah, I'm a bit confused about that as well. Looks like we don't pass all our imported tests, so some are marked with xfail. This one is good now, but we have more. For example [Tests/LibJS/Runtime/3rdparty/webkit/repeat-match-waldemar.js](https://github.com/LadybirdBrowser/ladybird/blob/cc6e048bd6cd4ea040d62fdf67d928d75ffb6f7e/Tests/LibJS/Runtime/3rdparty/webkit/repeat-match-waldemar.js). We fail these two there:
```js
/(?:a*?){2,3}/.exec("aa")
/(?:a*?){2,3}/.exec("a")
```
 so there's more work to do :)

Libraries/LibRegex/Rust/src/parser.rs
aplefull: yup, removed

Libraries/LibRegex/Rust/src/compiler.rs
aplefull: Yeah, I was thinking if I should squash these two commits, but decided against it 😅 I refactored it a lil bit now, so there's less repetition.

PR #9057, opened 2026-04-23
LibGfx: Accept PNGs with raw-deflate IDAT or fdAT payloads

kalenikaliaksandr:
The PNG spec mandates zlib-wrapped deflate for IDAT and fdAT chunks, but raw-deflate payloads appear in the wild and are commonly accepted by other decoders. libpng rejects them with "IDAT: incorrect header check" (or "fdAT: incorrect header check" for subsequent APNG frames), which made our PNG decoder return a blank fallback bitmap instead of the real image content. This unblocks progress on Cloudflare's Turnstile, whose challenge pipeline serves a raw-deflate tracking pixel and scores browsers based on the pixels they return when reading it back.

Before handing bytes to libpng, walk the PNG chunks and identify runs of consecutive IDAT or fdAT chunks (each run is one independent deflate stream; APNG frames beyond frame 0 use fdAT). For each run, streaming- inflate the concatenated deflate bytes as raw deflate; on success, emit a new PNG that prepends 0x78 0x01 to the first chunk's deflate data and appends the 4-byte Adler-32 to the last, with freshly stamped CRCs on only those two chunks. fdAT chunks preserve their 4-byte sequence_number prefix. Middle chunks pass through verbatim with their original CRCs so libpng still validates them.

A cheap 5-byte fast path short-circuits the common valid-zlib case without any inflate work: if the first byte's low nibble is 8 (the only plausible zlib CMF) and bytes 1-4 don't form a valid stored-block LEN/NLEN pair, the stream must be zlib and we pass through. Anything else (shorter prefixes, non-CMF nibble, or the 1/65536 random collision) falls through to the inflate attempt, which is the definitive test.

Stored CRCs on the first and last chunks of each rewrapped run are verified so a corrupted IDAT/fdAT isn't laundered into a successful decode. Inflated output per run is capped at 256 MiB as a bomb guard; input size is unbounded.

Tests cover: a real-world raw-deflate fixture, a hand-built non-uniform image, a 4-byte Huffman-coded payload too short for the stored-block fast path, a raw-deflate stream whose first bytes pass the naive zlib-header predicate, a corrupt-CRC fixture proving we don't launder broken CRCs, a 2-frame APNG with raw-deflate fdAT, and a 1024x1024 PNG with a ~1 MiB raw-deflate IDAT exercising streaming inflate at scale.


kalenikaliaksandr:
oops, that's all wrong

PR #9059, opened 2026-04-23
Array.prototype.shift() optimizations

kalenikaliaksandr:
Allows us to fail at Cloudflare Turnstile way much faster!


awesomekling:
At last! Waiting for Cloudflare Turnstile to fail has been pretty boring :)

PR #9060, opened 2026-04-23
LibWeb: Implement smooth scroll

dgolebiewski:
### Implement smooth scrolling.
~~Default to smooth scroll for wheel events to make scrolling with the mouse more pleasant.~~

I've tried to keep the smooth scrolling logic centralized and separated as its own thing.

There are ~~101~~ 113 new WPT tests we're passing.

#### Anchors on a page with `scroll-behavior: smooth`:
Before - a jarring jump:
[Nagranie ekranu_20260423_210656.webm](https://github.com/user-attachments/assets/8bde7963-3440-4fe9-acb9-b7f878f8d20d)

After - beautiful, smooth scroll:
[Nagranie ekranu_20260423_205600.webm](https://github.com/user-attachments/assets/66599796-9603-494e-bed8-83c97c63b1ea)

### DIFF
--- a/Libraries/LibWeb/CSS/Enums.json
+++ b/Libraries/LibWeb/CSS/Enums.json
@@ -935,6 +935,10 @@
     "to-zero",
     "up"
   ],
+  "scroll-behavior": [
+    "auto",
+    "smooth"
+  ],
   "scrollbar-gutter": [
     "auto",
     "stable",
--- a/Libraries/LibWeb/CSS/Properties.json
+++ b/Libraries/LibWeb/CSS/Properties.json
@@ -3839,10 +3839,7 @@
     "affects-layout": false,
     "animation-type": "none",
     "requires-computation": "never",
-    "valid-identifiers": [
-      "auto",
-      "smooth"
-    ]
+    "valid-types": ["scroll-behavior"]
   },
   "scroll-margin": {
     "initial": "0",
@@ -3940,7 +3937,7 @@
     ]
   },
   "scroll-padding": {
-    "initial": "auto",
+    "initial": "0",
     "positional-value-list-shorthand": true,
     "longhands": [
       "scroll-padding-top",
--- a/Libraries/LibWeb/CSS/VisualViewport.cpp
+++ b/Libraries/LibWeb/CSS/VisualViewport.cpp
@@ -5,6 +5,7 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
+#include <LibJS/Runtime/Value.h>
 #include <LibWeb/Bindings/Intrinsics.h>
 #include <LibWeb/Bindings/VisualViewport.h>
 #include <LibWeb/CSS/VisualViewport.h>
@@ -12,6 +13,11 @@
 #include <LibWeb/DOM/EventDispatcher.h>
 #include <LibWeb/HTML/EventNames.h>
 #include <LibWeb/HTML/Navigable.h>
+#include <LibWeb/HTML/Scripting/TemporaryExecutionContext.h>
+#include <LibWeb/InvalidateDisplayList.h>
+#include <LibWeb/Page/SmoothScrollHandler.h>
+#include <LibWeb/PixelUnits.h>
+#include <LibWeb/WebIDL/Promise.h>
 
 namespace Web::CSS {
 
@@ -146,6 +152,56 @@ WebIDL::CallbackType* VisualViewport::onscrollend()
     return event_handler_attribute(HTML::EventNames::scrollend);
 }
 
+// https://drafts.csswg.org/cssom-view-1/#perform-a-scroll
+GC::Ref<WebIDL::Promise> VisualViewport::perform_scroll_of_visual_viewport_scrolling_box(CSSPixelPoint offset, Bindings::ScrollBehavior scroll_behavior)
+{
+    // 1. Abort any ongoing smooth scroll for box.
+    // 2. Resolve all pending scroll Promises whose scroll container is box.
+    m_document->smooth_scroll_handler()->abort_any_ongoing_visual_viewport_scroll();
+
+    HTML::TemporaryExecutionContext temporary_execution_context { m_document->realm() };
+    // 3. Let scrollPromise be a new Promise.
+    auto scroll_promise = WebIDL::create_promise(m_document->realm());
+
+    // NB: Step 4 happens out of order, as we can start a smooth scroll or perform an instant scroll first.
+
+    // 5. If the user agent honors the scroll-behavior property and one of the following is true:
+    //      - behavior is "auto" and element is not null and its computed value of the scroll-behavior property is smooth, or
+    //      - behavior is smooth
+    //    then perform a smooth scroll of box to position; otherwise, perform an instant scroll of box to position.
+    // NB: Above determination happens in SmoothScrollHandler::resolve_scroll_behavior.
+    scroll_behavior = SmoothScrollHandler::resolve_scroll_behavior(*this, scroll_behavior);
+    if (scroll_behavior == Bindings::ScrollBehavior::Smooth) {
+        // NB: In this branch steps 6 and 7 are handled by SmoothScrollHander.
+        m_document->smooth_scroll_handler()->start_visual_viewport_scroll(scroll_promise, offset);
+    } else {
+        // 6. Wait until either the position has finished updating, or scrollPromise has been resolved.
+        set_offset(offset);
+        // 7. If scrollPromise is still in the pending state:
+        //      FIXME: 1. If the scroll position changed as a result of this call, emit the scrollend event.
+        //      2. Resolve scrollPromise.
+        WebIDL::resolve_promise(m_document->realm(), scroll_promise);
+    }
+
+    // 4. Return scrollPromise, and run the remaining steps in parallel.
+    return scroll_promise;
+}
+
+GC::Ref<WebIDL::Promise> VisualViewport::perform_scroll_of_visual_viewport_scrolling_box_by_delta(CSSPixelPoint delta, Bindings::ScrollBehavior scroll_behavior)
+{
+    CSSPixelPoint target_offset;
+    auto ongoing_visual_viewport_smooth_scroll = m_document->smooth_scroll_handler()->ongoing_visual_viewport_scroll();
+    if (ongoing_visual_viewport_smooth_scroll) {
+        auto ongoing_scroll_target_offset = ongoing_visual_viewport_smooth_scroll->target_offset();
+
+        target_offset = ongoing_scroll_target_offset + delta;
+    } else {
+        target_offset = m_offset + delta;
+    }
+
+    return perform_scroll_of_visual_viewport_scrolling_box(target_offset, scroll_behavior);
+}
+
 Gfx::AffineTransform VisualViewport::transform() const
 {
     Gfx::AffineTransform transform;
@@ -155,8 +211,26 @@ Gfx::AffineTransform VisualViewport::transform() const
     return transform;
 }
 
+void VisualViewport::set_offset(CSSPixelPoint offset)
+{
+    auto did_change = offset != m_offset;
+    m_offset = offset;
+
+    if (did_change) {
+        m_document->set_needs_accumulated_visual_contexts_update(true);
+        m_document->set_needs_repaint(Badge<CSS::VisualViewport> {}, InvalidateDisplayList::Yes);
+    }
+}
+
+void VisualViewport::scroll_by(CSSPixelPoint delta)
+{
+    set_offset(m_offset + delta);
+}
+
 void VisualViewport::zoom(CSSPixelPoint position, double scale_delta)
 {
+    m_document->smooth_scroll_handler()->abort_any_ongoing_visual_viewport_scroll();
+
     static constexpr double MIN_ALLOWED_SCALE = 1.0;
     static constexpr double MAX_ALLOWED_SCALE = 5.0;
     double new_scale = clamp(m_scale * (1 + scale_delta), MIN_ALLOWED_SCALE, MAX_ALLOWED_SCALE);
@@ -173,9 +247,7 @@ void VisualViewport::zoom(CSSPixelPoint position, double scale_delta)
     new_offset = { clamp(new_offset.x(), 0.0f, max_x_offset), clamp(new_offset.y(), 0.0f, max_y_offset) };
 
     m_scale = new_scale;
-    m_offset = (new_offset / m_scale).to_type<CSSPixels>();
-    m_document->set_needs_accumulated_visual_contexts_update(true);
-    m_document->set_needs_repaint(Badge<CSS::VisualViewport> {}, InvalidateDisplayList::Yes);
+    set_offset((new_offset / m_scale).to_type<CSSPixels>());
 }
 
 CSSPixelPoint VisualViewport::map_to_layout_viewport(CSSPixelPoint position) const
@@ -186,10 +258,10 @@ CSSPixelPoint VisualViewport::map_to_layout_viewport(CSSPixelPoint position) con
 
 void VisualViewport::reset()
 {
+    m_document->smooth_scroll_handler()->abort_any_ongoing_visual_viewport_scroll();
+
     m_scale = 1.0;
-    m_offset = { 0, 0 };
-    m_document->set_needs_accumulated_visual_contexts_update(true);
-    m_document->set_needs_repaint(Badge<CSS::VisualViewport> {}, InvalidateDisplayList::Yes);
+    set_offset({ 0, 0 });
 }
 
 }
--- a/Libraries/LibWeb/DOM/Document.cpp
+++ b/Libraries/LibWeb/DOM/Document.cpp
@@ -184,6 +184,7 @@
 #include <LibWeb/Namespace.h>
 #include <LibWeb/Page/EventHandler.h>
 #include <LibWeb/Page/Page.h>
+#include <LibWeb/Page/SmoothScrollHandler.h>
 #include <LibWeb/Painting/AccumulatedVisualContext.h>
 #include <LibWeb/Painting/DisplayList.h>
 #include <LibWeb/Painting/DisplayListCommand.h>
@@ -635,6 +636,8 @@ void Document::initialize(JS::Realm& realm)
 
     m_list_of_available_images = realm.create<HTML::ListOfAvailableImages>();
 
+    m_smooth_scroll_handler = heap().allocate<SmoothScrollHandler>(*this);
+
     page().client().page_did_create_new_document(*this);
 
     ensure_cookie_version_index(m_url);
@@ -717,6 +720,9 @@ void Document::visit_edges(Cell::Visitor& visitor)
     visitor.visit(m_document_observers_being_notified);
     for (auto& pending_scroll_event : m_pending_scroll_events)
         visitor.visit(pending_scroll_event.event_target);
+
+    visitor.visit(m_smooth_scroll_handler);
+
     for (auto& resize_observer : m_resize_observers)
         visitor.visit(resize_observer);
 
@@ -3811,7 +3817,7 @@ void Document::scroll_to_the_beginning_of_the_document()
 {
     // FIXME: Actually implement this algorithm
     if (auto navigable = this->navigable())
-        navigable->perform_scroll_of_viewport_scrolling_box({ 0, 0 });
+        navigable->perform_scroll_of_viewport_scrolling_box(this, { 0, 0 }, Bindings::ScrollBehavior::Auto);
 }
 
 StringView Document::ready_state() const
--- a/Libraries/LibWeb/DOM/Element.cpp
+++ b/Libraries/LibWeb/DOM/Element.cpp
@@ -2933,7 +2933,7 @@ static CSSPixelPoint determine_the_scroll_into_view_position(Element& target, Bi
 static GC::Ref<WebIDL::Promise> scroll_an_element_into_view(Element& target, Bindings::ScrollBehavior behavior, Bindings::ScrollLogicalPosition block, Bindings::ScrollLogicalPosition inline_, GC::Ptr<Element> container)
 {
     // 1. Let ancestorPromises be an empty set of Promises.
-    GC::RootVector<GC::Ref<WebIDL::Promise>> ancestor_promises { target.heap() };
+    GC::RootVector<GC::Ref<WebIDL::Promise>> ancestor_promises;
 
     // 2. For each ancestor element or viewport that establishes a scrolling box scrolling box, in order of innermost
     //    to outermost scrolling box, run these substeps:
@@ -3770,6 +3770,9 @@ GC::Ref<WebIDL::Promise> Element::scroll(Bindings::ScrollToOptions options)
         && scroll_offset({}).is_zero()
         && this != document.body()
         && this != document.document_element()) {
+        if (document.layout_is_up_to_date() && paintable_box()) {
+            document.smooth_scroll_handler()->abort_any_ongoing_scroll_of_box(*paintable_box());
+        }
         return WebIDL::create_resolved_promise(realm(), JS::js_undefined());
     }
 
--- a/Libraries/LibWeb/DOM/Element.cpp
+++ b/Libraries/LibWeb/DOM/Element.cpp
@@ -127,6 +127,7 @@
 #include <LibWeb/WebIDL/AbstractOperations.h>
 #include <LibWeb/WebIDL/DOMException.h>
 #include <LibWeb/WebIDL/ExceptionOr.h>
+#include <LibWeb/WebIDL/Promise.h>
 #include <LibWeb/XML/XMLFragmentParser.h>
 
 namespace Web::DOM {
@@ -2206,7 +2207,7 @@ void Element::set_scroll_left(double x)
     // FIXME: Implement this in terms of calling "scroll the element".
     auto scroll_offset = paintable_box()->scroll_offset();
     scroll_offset.set_x(CSSPixels::nearest_value_for(x));
-    paintable_box()->set_scroll_offset(scroll_offset);
+    paintable_box()->perform_scroll(scroll_offset);
 }
 
 void Element::set_scroll_top(double y)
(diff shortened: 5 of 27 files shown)


Libraries/LibWeb/CSS/VisualViewport.cpp
```diff
@@ -146,6 +149,25 @@ WebIDL::CallbackType* VisualViewport::onscrollend()
     return event_handler_attribute(HTML::EventNames::scrollend);
 }
 
+GC::Ref<WebIDL::Promise> VisualViewport::scroll_by(GC::Ref<JS::Realm> realm, CSSPixelPoint delta, Bindings::ScrollBehavior scroll_behavior) {
+    if (scroll_behavior == Bindings::ScrollBehavior::Smooth) {
+        auto ongoing_vv_scroll = m_document->smooth_scroll_handler()->ongoing_visual_viewport_scroll();
```
Psychpsyo: You generally shouldn't be shortening stuff to cryptic acronyms in variables and function names.

Libraries/LibWeb/DOM/Element.cpp
```diff
@@ -3652,9 +3649,29 @@ GC::Ref<WebIDL::Promise> Element::scroll(double x, double y)
     // 2. If invoked with two arguments, follow these substeps:
     //     1. Let options be null converted to a ScrollToOptions dictionary. [WEBIDL]
     //     2. Let x and y be the arguments, respectively.
-    //     3. Normalize non-finite values for x and y.
+    //     3. Normalize non-finite values for x and y. NOTE: (this will happen in scroll(options))
```
Psychpsyo: This text comes straight from the spec, you should not be modifying it.
Instead, add a comment like `// NB: This happens in scroll(options) instead`.

Libraries/LibWeb/DOM/Element.cpp
```diff
@@ -3652,9 +3649,29 @@ GC::Ref<WebIDL::Promise> Element::scroll(double x, double y)
...
     //     4. Let the left dictionary member of options have the value x.
     //     5. Let the top dictionary member of options have the value y.
+    HTML::ScrollToOptions options;
```
Psychpsyo: The code for each step should be after the step it implements, not all at once after the steps as a whole.

Libraries/LibWeb/DOM/Element.cpp
```diff
@@ -3652,9 +3649,29 @@ GC::Ref<WebIDL::Promise> Element::scroll(double x, double y)
...
+
+    return scroll(options);
+}
+
+// https://drafts.csswg.org/cssom-view/#dom-element-scroll
+GC::Ref<WebIDL::Promise> Element::scroll(HTML::ScrollToOptions options)
+{
+    // 1. If invoked with one argument, follow these substeps:
+    //     1. Let options be the argument.
+    //     2. Normalize non-finite values for left and top dictionary members of options, if present.
```
Psychpsyo, line 3738: Same thing about code implementing these steps being after the individual steps.

Libraries/LibWeb/HTML/Navigable.cpp
```diff
@@ -3140,19 +3180,29 @@ void Navigable::render_screenshot(Gfx::PaintingSurface& painting_surface, PaintC
...
 }
 
 // https://drafts.csswg.org/cssom-view/#viewport-perform-a-scroll
-GC::Ref<WebIDL::Promise> Navigable::perform_a_scroll_of_the_viewport(CSSPixelPoint position)
+GC::Ref<WebIDL::Promise> Navigable::perform_a_scroll_of_the_viewport(CSSPixelPoint position, Bindings::ScrollBehavior scroll_behavior)
 {
     // 1. Let doc be the viewport’s associated Document.
     auto doc = active_document();
 
+    doc->smooth_scroll_handler()->abort_any_ongoing_visual_viewport_scroll();
```
Psychpsyo: This is not in the spec here and should have an explanation as to why it's necessary for us to do it anyways.

Libraries/LibWeb/HTML/Navigable.cpp
```diff
@@ -3213,28 +3263,54 @@ GC::Ref<WebIDL::Promise> Navigable::perform_a_scroll_of_the_viewport(CSSPixelPoi
...
-        doc->set_needs_repaint(Badge<HTML::Navigable> {}, InvalidateDisplayList::Yes);
-    } else {
-        doc->set_needs_repaint(Badge<HTML::Navigable> {}, InvalidateDisplayList::No);
-    }
+    auto& realm = doc->realm();
+
+    auto scroll_promise_1 = perform_scroll_of_viewport_scrolling_box(doc, realm, new_viewport_scroll_offset.to_type<CSSPixels>(), scroll_behavior);
+
+    // AD-HOC: Steps out of order due to scoping.
+    // NOTE: Conceptually, the visual viewport is scrolled until it "bumps up" against the layout viewport edge and then
```
Psychpsyo: Should be `NB:` instead of `NOTE:` if it's not from the spec.

Libraries/LibWeb/Page/EventHandler.cpp
```diff
@@ -628,7 +628,8 @@ EventResult EventHandler::handle_mousewheel(CSSPixelPoint visual_viewport_positi
             if (node->dispatch_event(UIEvents::WheelEvent::create_from_platform_event(node->realm(), m_navigable->active_window_proxy(), UIEvents::EventNames::wheel, screen_position, page_offset, viewport_position, offset, wheel_delta_x, wheel_delta_y, button, buttons, modifiers).release_value_but_fixme_should_propagate_errors())) {
                 if (could_scroll_viewport) {
                     auto viewport_scroll_position_before = CSSPixelPoint { CSSPixels(document->visual_viewport()->page_left()), CSSPixels(document->visual_viewport()->page_top()) };
-                    m_navigable->scroll_viewport_by_delta({ wheel_delta_x, wheel_delta_y });
+                    // AD-HOC: Default to smooth scroll for wheel events so we get pleasent visuals.
```
Psychpsyo: All the code here is quite AD-HOC, so I'm not sure this comment is necessary.

Psychpsyo:
(No real input on the approach from me, just general spec/formatting stuffs)

dgolebiewski:
Shuffled the code around to be more true to word of the spec.
Also, smooth scroll now works on regular boxes.

Still, I need to figure out tests.

Libraries/LibWeb/CSS/VisualViewport.cpp
dgolebiewski: Right, I've renamed this.

Libraries/LibWeb/DOM/Element.cpp
dgolebiewski: Made it an NB and moved it to be on a separate line.

Libraries/LibWeb/CSS/Properties.json
```diff
@@ -4700,4 +4700,4 @@
       "auto"
     ]
   }
-}
+}
```
nico: ?

Libraries/LibWeb/HTML/Navigable.cpp
dgolebiewski: It shouldn't be there, it was a leftover from some troubleshooting.

Libraries/LibWeb/HTML/Navigable.cpp
dgolebiewski: It's from the spec, but I'm not sure it should be pasted there - I'll probably remove it.

Libraries/LibWeb/CSS/Properties.json
dgolebiewski: Formatter mishap, fixed it.

Psychpsyo:
> Still, I need to figure out tests.

You might want to check if there's WPT tests for this and then just import those.

dgolebiewski:
> You might want to check if there's WPT tests for this and then just import those.

I will look for those later. First, I need to fix a few failing tests that expect `internals.wheel` to perform instant scroll (which now it doesn't as wheel defaults to smooth scroll).
I'll probably just `setTimeout` my way out of those failures right now and make a separate commit (or should it be a separate PR?) implementing `scrollend` event, which should let me make those tests pretty again.

Psychpsyo:
> > You might want to check if there's WPT tests for this and then just import those.
> 
> I will look for those later. First, I need to fix a few failing tests that expect `internals.wheel` to perform instant scroll (which now it doesn't as wheel defaults to smooth scroll). I'll probably just `setTimeout` my way out of those failures right now and make a separate commit (or should it be a separate PR?) implementing `scrollend` event, which should let me make those tests pretty again.

It's best to avoid `setTimeout()` in tests as much as possible. When the browser is running all the tests at once, it is under a lot of stress and operations that would usually only take X amount of time can take significantly longer.
I'd consider maybe making the internals wheel events not smooth scroll, or giving them an option. (Not sure if those are a thing also used in WPT; If they are, probably best to check if that has an opinion on this.)
Making random tests wait for smooth scrolls unnecessarily would needlessly slow down running the test suite. (Not that big of an issue, but it'd be nice if avoidable.)
Alternative would be to just toggle user scrolls to non-smooth during test runs. That should be an option in the long run anyways, even for regular users.

dgolebiewski:
Okay - no `setTimeout` in tests :laughing:.
So, I've added a setting for smooth scrolling on wheel events, made it on by default and forced it off in test mode.

Also, imported a few WPT tests and we're now passing 101 of those.

I think this PR is now complete enough to be marked as ready for review, though there's still one major issue I noticed. Most of the time I had Ladybird open on my second monitor and there, everything looked fine, but on my main display (which has higher refresh rate - 180Hz vs 120Hz) scrolling with the mouse wheel causes pretty noticeable flickering. I'm pretty sure that my changes do not cause it, but I really have no clue where to look for the source of the issue. 
Here's a screen cap of this happening:
[Nagranie ekranu_20260430_192410.webm](https://github.com/user-attachments/assets/759fddb8-e288-41b8-b19e-8555788f9f48)
And here's the same page on lower refresh rate:
[Nagranie ekranu_20260430_192433.webm](https://github.com/user-attachments/assets/f2e9b9c8-a17b-4c5d-916e-e90e16771349)
(interestingly enough, the flickering does show up on the second recording and after going back and taking a closer look it is barerly noticable even on the 120Hz display)

Anyway, I will be marking this PR as ready for review and would appreciate any clues as to where that flickering could be coming from.

dgolebiewski:
Rebased onto latest master and the flickering issue is 99% gone. I'm pretty sure it's thanks to this PR: 
https://github.com/LadybirdBrowser/ladybird/pull/9167
There's still an occasional flicker but it's hardly noticeable now compared to before.
[Nagranie ekranu_20260502_094431.webm](https://github.com/user-attachments/assets/88427580-f430-40a4-a936-36c133472819)

Lubrsi:
I believe the flickering is caused by not waiting for the rendering thread to complete the frame before presenting it. That PR likely helped with it since it's drawing less frames. I don't think it's caused by your changes :^)

Libraries/LibWeb/Page/SmoothScrollHandler.h
```diff
@@ -0,0 +1,90 @@
...
+#include <LibGC/CellAllocator.h>
+#include <LibJS/Heap/Cell.h>
+#include <LibWeb/Bindings/Window.h>
+#include <LibWeb/DOM/Document.h>
+#include <LibWeb/Forward.h>
+#include <LibWeb/PixelUnits.h>
+
+namespace Web {
+
+class SmoothScrollTask {
```
Lubrsi: issue: This should be a GC::Cell with a visit_edges implementation, since it contains strong reference pointers to GC objects.

Libraries/LibWeb/Page/SmoothScrollHandler.h
```diff
@@ -0,0 +1,90 @@
...
+    CSSPixelPoint source_offset() { return m_source_offset; }
+    CSSPixelPoint target_offset() { return m_target_offset; }
+    double start_time() const { return m_start_time; }
+    void update_duration();
+    double duration() const { return m_duration; }
+    bool is_ongoing() const { return m_is_ongoing; }
+
+private:
+    GC::Ref<WebIDL::Promise> m_scroll_promise;
+    Optional<Painting::PaintableBox*> m_box;
```
Lubrsi: suggestion: Use GC::Ptr instead of Optional, since GC::Ptr is nullable :^)

Libraries/LibWeb/Page/SmoothScrollHandler.h
```diff
@@ -0,0 +1,90 @@
...
+    CSSPixelPoint m_source_offset;
+    CSSPixelPoint m_target_offset;
+    double m_start_time;
+    double m_duration;
+    bool m_is_ongoing { false };
+};
+
+class SmoothScrollHandler : public JS::Cell {
+    // FIXME: Does this need to be a GC Cell?
+    GC_CELL(SmoothScrollHandler, JS::Cell);
```
Lubrsi: suggestion: It doesn't necessarily have to be GC::Cell, but it's better to be that because it makes it not dependent on JS::Realm unnecessarily. You would then take a `GC::Heap&` instead of `GC::Ref<JS::Realm>` to allocate `SmoothScrollHandler`.

Libraries/LibWeb/Page/SmoothScrollHandler.cpp
```diff
@@ -0,0 +1,317 @@
...
+
+CSSPixelPoint SmoothScrollHandler::current_scroll_offset(SmoothScrollTask task, double progress)
+{
+    CSSPixelPoint source_offset = task.source_offset();
+    CSSPixelPoint target_offset = task.target_offset();
+
+    auto t = ease(progress);
+
+    auto x = lerp(source_offset.x().to_double(), target_offset.x().to_double(), t);
+    auto y = lerp(source_offset.y().to_double(), target_offset.y().to_double(), t);
```
Lubrsi: issue: Windows CI says `lerp` is undefined.
suggestion: Try using `mix` from `StdLibExtras.h`, since it has a comment saying "aka lerp"

Libraries/LibWeb/Page/SmoothScrollHandler.cpp
```diff
@@ -0,0 +1,317 @@
...
+        visitor.visit(m_ongoing_viewport_scroll->scroll_promise());
+    }
+
+    if (m_ongoing_visual_viewport_scroll.has_value()) {
+        visitor.visit(m_ongoing_visual_viewport_scroll->scroll_promise());
+    }
+
+    for (auto& task : m_ongoing_box_scrolls) {
+        visitor.visit(task.scroll_promise());
+    }
```
Lubrsi: todo: With SmoothScrollTask being a GC::Cell, this would become `visitor.visit(m_ongoing_box_scrolls)`

Libraries/LibWeb/Page/SmoothScrollHandler.cpp
```diff
@@ -0,0 +1,317 @@
...
+
+    return ongoing_task;
+}
+
+void SmoothScrollHandler::abort_any_ongoing_viewport_scroll()
+{
+    if (!m_ongoing_viewport_scroll.has_value()) {
+        return;
+    }
```
Lubrsi: style: Typically we have single lines control flow like so:
```suggestion
    if (!m_ongoing_viewport_scroll.has_value())
        return;
```

Libraries/LibWeb/Page/SmoothScrollHandler.cpp
```diff
@@ -0,0 +1,317 @@
...
+        CSSPixelPoint current_offset = current_scroll_offset(task, progress);
+
+        task.box().value()->set_scroll_offset(current_offset);
+
+        if (progress >= 1) {
+            finish(task);
+        }
+    }
+
+    Vector<SmoothScrollTask> ongoing_box_scrolls;
```
Lubrsi: issue: It's not safe to use plain storage to GC objects on the stack, because the GC sweep can't see into it.
suggestion: Use GC::RootVector instead.

Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -700,6 +703,8 @@ void Document::visit_edges(Cell::Visitor& visitor)
     visitor.visit(m_document_observers_being_notified);
     for (auto& pending_scroll_event : m_pending_scroll_events)
         visitor.visit(pending_scroll_event.event_target);
+    visitor.visit(m_smooth_scroll_handler);
```
Lubrsi, line 724: issue: This should have a blank line before it (I believe some compilers would warn of this since it could be confused with wanting to be part of the for loop above)

Libraries/LibWeb/CSS/VisualViewport.cpp
```diff
@@ -155,8 +210,26 @@ Gfx::AffineTransform VisualViewport::transform() const
     return transform;
 }
 
+void VisualViewport::set_offset(CSSPixelPoint offset) {
```
Lubrsi: style: We have the opening `{` on a line by itself

Lubrsi (changes requested):
Welcome to the Ladybird project :^)
Here's my first pass, it's looking good with some gotchas around GC:

dgolebiewski:
> Welcome to the Ladybird project :^) Here's my first pass, it's looking good with some gotchas around GC:

Thanks for the review!
I've implemented the requested changes.

AtkinsSJ:
Hi, this still has conflicts.

dgolebiewski:
Hi, sorry - I haven't had much time in the last few days.
I've sorted it out now.

dgolebiewski:
Rebased after #9225 got merged. It now passes 12 more tests bringing the total to 113 new passing WPT tests.

Also, the scroll-behavior demo on MDN works now :tada: :
[Nagranie ekranu_20260510_151528.webm](https://github.com/user-attachments/assets/e8c9a432-38e3-4d0c-9902-e0f6e000e853)

dgolebiewski:
I've decided to walk back defaulting to smooth scroll for wheel events. Async scrolling is just around the corner and I would rather implement smooth scroll there first so the experience is consistent no matter if scrolling takes place on the main thread or compositor thread.

Libraries/LibWeb/CSS/Enums.json
```diff
@@ -934,6 +934,10 @@
     "to-zero",
     "up"
   ],
+  "scroll-behavior": [
+    "auto",
+    "smooth"
+  ],
```
AtkinsSJ, line 941: Please also modify the property definition in `Properties.json` to use have `"valid-types": ["scroll-behavior"]` instead of listing these identifiers literally there.

Libraries/LibWeb/CSS/VisualViewport.cpp
```diff
@@ -155,8 +211,28 @@ Gfx::AffineTransform VisualViewport::transform() const
...
+void VisualViewport::set_offset(CSSPixelPoint offset)
+{
+    auto did_change = offset != m_offset;
+    m_offset = offset;
+
+    if (did_change) {
+        m_document->set_needs_accumulated_visual_contexts_update(true);
+        m_document->set_needs_repaint(Badge<CSS::VisualViewport> {}, InvalidateDisplayList::Yes);
+    } else {
+        m_document->set_needs_repaint(Badge<CSS::VisualViewport> {}, InvalidateDisplayList::No);
```
AtkinsSJ: Why do we need a repaint if nothing changed?

Libraries/LibWeb/DOM/Element.cpp
```diff
@@ -2903,24 +2905,25 @@ static GC::Ref<WebIDL::Promise> scroll_an_element_into_view(Element& target, Bin
         if (true) {
             // -> If scrolling box is associated with an element
             if (auto* element = as_if<Element>(scrolling_box)) {
-                // FIXME: Perform a scroll of the element’s scrolling box to position, with the element as the associated element and behavior as the scroll behavior.
-                element->paintable_box()->set_scroll_offset(position);
+                // Perform a scroll of the element’s scrolling box to position, with the element as the associated element and behavior as the scroll behavior.
```
AtkinsSJ: Nit: Please wrap comments at 120 columns. (We have a lot of existing ones that are wrapped differently, but if you're modifying one, please fix it.)

Libraries/LibWeb/HTML/Navigable.cpp
```diff
@@ -3007,6 +3015,54 @@ void Navigable::adopt_pending_async_scroll_offsets()
...
+
+    if (!doc) {
+        return {};
+    }
+
+    return perform_scroll_of_viewport_scrolling_box(doc, new_position, scroll_behavior);
+}
+
+// https://drafts.csswg.org/cssom-view-1/#perform-a-scroll
+GC::Ref<WebIDL::Promise> Navigable::perform_scroll_of_viewport_scrolling_box(GC::Ptr<DOM::Document> doc, CSSPixelPoint new_position, Bindings::ScrollBehavior scroll_behavior)
```
AtkinsSJ, line 3166: It's unfortunate we have this implemented in multiple places, but that can be fixed later.

Libraries/LibWeb/Page/SmoothScrollHandler.cpp
```diff
@@ -0,0 +1,225 @@
...
+{
+    Base::visit_edges(visitor);
+    visitor.visit(m_document);
+
+    visitor.visit(m_ongoing_viewport_scroll);
+    visitor.visit(m_ongoing_visual_viewport_scroll);
+}
+
+// Steps 6 and 7 of https://drafts.csswg.org/cssom-view-1/#perform-a-scroll
+void SmoothScrollHandler::finish(GC::Ptr<SmoothScrollTask> task)
```
AtkinsSJ: If you know task isn't null, then use either a `GC::Ref`, or a plain `SmoothScrollTask&` as the parameter. `GC::Ptr` should only be used if it's allowed to be null at that point.

Same point for a couple of other functions and methods here.

Libraries/LibWeb/Page/SmoothScrollHandler.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+// https://drafts.csswg.org/cssom-view-1/#ref-for-propdef-scroll-behavior
+Bindings::ScrollBehavior SmoothScrollHandler::resolve_scroll_behavior(DOM::Element* element, Bindings::ScrollBehavior scroll_behavior)
+{
+    // If the user agent honors the scroll-behavior property and one of the following is true:
+
+    // - behavior is "auto" and element is not null and its computed value of the scroll-behavior property is smooth,
+    //   or
+    // - behavior is smooth
+    //
```
AtkinsSJ: Nit: Blank comment line here.

Libraries/LibWeb/Page/SmoothScrollHandler.cpp
```diff
@@ -0,0 +1,225 @@
...
+    if ((scroll_behavior == Bindings::ScrollBehavior::Auto && element && element->computed_properties() && element->computed_properties()->scroll_behavior() == CSS::ScrollBehavior::Smooth) || scroll_behavior == Bindings::ScrollBehavior::Smooth) {
```
AtkinsSJ: Nit: Maybe break this into two lines so it's easier to follow the two conditions.

Libraries/LibWeb/Page/SmoothScrollHandler.cpp
```diff
@@ -0,0 +1,225 @@
...
+        // then perform a smooth scroll of box to position;
+        return Bindings::ScrollBehavior::Smooth;
+    }
+
+    // otherwise, perform an instant scroll of box to position.
+    return Bindings::ScrollBehavior::Instant;
+}
+
+Bindings::ScrollBehavior SmoothScrollHandler::resolve_scroll_behavior(HTML::Navigable* navigable, Bindings::ScrollBehavior scroll_behavior)
```
AtkinsSJ: Don't use a pointer if the navigable isn't allowed to be null. (Code below assumes it is not.) Same for the method below.

Libraries/LibWeb/Page/SmoothScrollHandler.cpp
```diff
@@ -0,0 +1,225 @@
...
+{
+    if (auto doc = navigable->active_document()) {
+        if (auto element = doc->scrolling_element()) {
+            return resolve_scroll_behavior(const_cast<DOM::Element*>(element.ptr()), scroll_behavior);
```
AtkinsSJ: `*element` instead of `element.ptr()` would give you a reference, because we know it isn't null. I think then the `resolve_scroll_behavior()` overload wouldn't need to take the element by pointer at all.

Libraries/LibWeb/Page/SmoothScrollHandler.cpp
```diff
@@ -119,6 +165,27 @@ void SmoothScrollHandler::start_visual_viewport_scroll(GC::Ref<WebIDL::Promise>
...
+{
+    VERIFY(!ongoing_scroll_of_box(box) || !ongoing_scroll_of_box(box)->is_ongoing());
+
+    auto source_offset = box->scroll_offset();
+    auto task = heap().allocate<SmoothScrollTask>(box, promise, source_offset, target_offset, HighResolutionTime::unsafe_shared_current_time());
+
+    m_ongoing_box_scrolls.append(task);
+}
+
+GC::Ptr<SmoothScrollTask> SmoothScrollHandler::ongoing_scroll_of_box(RefPtr<Painting::PaintableBox> const& box)
```
AtkinsSJ: `box` seems like it could be non-null for this and `abort_ongoing_scroll_of_box()`. In both cases we always pass in a valid box as far as I can tell.

AtkinsSJ (changes requested):
Generally this seems good to me, but there are issues around passing nullable types in a few places when we could use non-nullable ones instead.

Libraries/LibWeb/HTML/Navigable.cpp
dgolebiewski: Yeah, I was thinking about ways to make it happen in one place but couldn't come up with a way to do that without some larger changes to unify how the scroll state is represented in PaintableBox/Element and Navigable. I don't feel comfortable attempting this kind of change with my current level of Ladybird and cpp knowledge.

Libraries/LibWeb/CSS/VisualViewport.cpp
dgolebiewski: I don't think it does, though `Navigable::perform_a_scroll_of_the_viewport` was doing that before.

gmta:

PR #9064, opened 2026-04-23
LibWeb/Bindings: Parse all IDL up front and stop requirement for #import

shannonbooth:
This MR moves the IDL generator away from import driven parsing and toward running the bindings generator on every IDL file at once. Some improvements

- Remove the hardcoded list of platform objects to determine "is platform object"
- Remove the "using namespace" hack by inferring namespace by module file location
- Remove the need for manual imports in IDL files


shannonbooth:
Actually I'll include the commit for just getting rid of import support altogether!

alimpfard:
just a...slight nit that you may freely ignore, the `#import`s themselves have been very helpful every time I've had to deal with IDLs beyond copying spec definitions into them (discoverability, etc etc), it could be interesting to keep them and enforce their use?

Psychpsyo:
I'd find it weird and confusing to have `#import`s lying around that don't do anything.
Any time I need to find a specific IDL file, I just search for the name of the thing I want in `*.idl` with VSCode.

alimpfard:
yeah I'm more of a "this file better tell me where to look next" kinda guy, but anyhow just a suggestion - feel free to ignore it.

ADKaster:
We have (had?) IDE plugins to make such navigation work better. The vscode one seems mostly lost, but a copy exists here: https://github.com/SerenityOS/serenity/pull/25909 . The JetBrains one needs a rewrite for the new code engine: https://github.com/LadybirdBrowser/Ladybird-DSL-Plugin

shannonbooth:
> yeah I'm more of a "this file better tell me where to look next" kinda guy, but anyhow just a suggestion - feel free to ignore it.

I thought about it, but I personally think it would be best to remove them at least for now. There's probably going to be a whole lot more churn for these files soon splitting them out when they are in a state that an IDL file will be code generated individually when it only has a dictionary in it type of thing (which we'll likely want to be doing to reduce includes). Seems like it would be hard to maintain the include list since types will be pulled in transitively from all over the place

PR #9065, opened 2026-04-24
CMake: Add overlay port for highway 1.4.0

ADKaster:
Until the upstream vcpkg port is updated, this will patch in the new
release of highway. The new release includes fixes for EVEX512 feature
detection in gcc>=15 and clang>=22, and unblocks using those compilers.

ref: https://github.com/microsoft/vcpkg/pull/51357


ADKaster:
I'm going to merge this under the assumption that the microsoft vcpkg folks won't get to my PR until the start of next week.

PR #9069, opened 2026-04-24
test-web: Keep data safe from crashes and add "view by pid"

jonbgamble:
Don't let harness crashes prevent results/index.html generation. Make index.html static and just copy it over when the test run begins.

Link a new results-manifest.js script from index.html which provides only the results array. results-manifest.js accumulates results as they happen. test-web keeps it consistent by seeking to `eof - 3` (the offset of the array's closing bracket) before writing each new result.

Each test view has a Pid tab where you can view all tests from that same WebContent in sequence. Click "copy run command" inside of that tab to construct a command line featuring all of those tests.

image


Tests/LibWeb/test-web/results-index.css
```diff
@@ -0,0 +1,598 @@
+        :root {
```
Zaggy1024: Indentation in this file looks wrong, is there a reason why it's still indented 8 spaces?

Tests/LibWeb/test-web/Results.cpp
```diff
@@ -29,6 +35,54 @@ static ErrorOr<ByteString> prepare_output_path(Test const& test)
     return base_path.string();
 }
 
+static ErrorOr<void> initialize_manifest_array()
+{
+    auto path = LexicalPath::join(Application::the().results_directory, MANIFEST_FILE_NAME).string();
+    auto file = TRY(Core::File::open(path, Core::File::OpenMode::Write | Core::File::OpenMode::Truncate));
+    auto contents = ByteString::formatted("window.{} = [\n];\n", "RESULTS_MANIFEST"sv);
```
Zaggy1024: Formatting just to interpolate `RESULTS_MANIFEST` here seems unnecessary.

Tests/LibWeb/test-web/Results.cpp
```diff
@@ -29,6 +35,54 @@ static ErrorOr<ByteString> prepare_output_path(Test const& test)
...
+}
+
+static ErrorOr<void> append_manifest_record(StringView json)
+{
+    auto path = LexicalPath::join(Application::the().results_directory, MANIFEST_FILE_NAME).string();
+    auto file = TRY(Core::File::open(path, Core::File::OpenMode::ReadWrite));
+    TRY(file->seek(-3, SeekMode::FromEndPosition));
+    auto line = ByteString::formatted("{},\n];\n", json);
+    TRY(file->write_until_depleted(line.bytes()));
+    return {};
```
Zaggy1024, line 54: I wonder if this would be blocking us from starting a new test significantly longer than before. Is there a measurable difference in the overall run time when writing between each test?

Tests/LibWeb/test-web/TestRunCapture.cpp
```diff
@@ -160,10 +160,10 @@ void TestRunCapture::write_test_output(TestWebView const& view)
     destroy_view_capture_of(view);
 }
 
-bool TestRunCapture::write_helper_process_output()
+void TestRunCapture::write_helper_process_output()
 {
     restore_stderr();
-    return m_helper_output.transfer_to_output_file().value_or(false);
+    (void)m_helper_output.transfer_to_output_file();
```
Zaggy1024: Looks like this may be the last use of the return value from that function. Should its signature change now?

Tests/LibWeb/test-web/results-index.html
```diff
@@ -136,8 +178,7 @@ <h1>
                     </div>
                     <div class="test-content">`;
             if (tabs) {
-                html += `<div class="tab-bar">${tabs.bar}</div>
-                        <div class="preview-container" id="preview-${testId}">
+                html += `<div class="tab-bar">${tabs.bar}</div><div class="preview-container" id="preview-${testId}">
```
Zaggy1024: Whitespace-only change?

Zaggy1024 (commented):
The idea of assuming the layout of the output file when we may have multiple places we want to insert data into it feels a little strange, but I'll leave the decision on whether that's okay to others. It definitely is nice to have some output available even if the harness crashes, given how much I've seen that happen on CI.

Other than that, mainly just some nits I noticed.

jonbgamble:
> The idea of assuming the layout of the output file when we may have multiple places we want to insert data into it feels a little strange

I don't want to invent a mini-database here, I always append at eof-3 and those three truncated bytes must always be `];\n`

If test-mode ladybird was as lenient on opaque origin fetches as it is with script tags, I'd just use ndjson.

Zaggy1024:
Right, my issue isn't so much that it's a problem now, but more that it might be easy for people not aware of this setup to break it. That said, it'll be very apparent if that happens since it'll fail to parse, so my objection is pretty minor.

jonbgamble:
Well, my defense to your objection as worded makes no sense (since nobody will be viewing results using `Ladybird --test-mode`). What I meant to say is "I did naively try it with ndjson appends to avoid the eof-3 shenanigans, but web security happened"

jonbgamble:
Shall we just start requiring prettier on html/css/js in test-web?

I *really* hate how it handles ternary operator since one of the 3.x updates, but it's nice to remove all cause for arguments about spacing.

Tests/LibWeb/test-web/Results.cpp
jonbgamble: I just timed it once.

In-memory time:
```
Build/release/bin/test-web  149.49s user 35.56s system 456% cpu 40.564 total
```

Manifest approach time:
```
Build/release/bin/test-web  142.84s user 34.45s system 447% cpu 39.630 total
```

If there's a measurable difference to in-memory vs writing each result to a file as it happens over many runs, it would take someone with more patience than me to tease it out.

Tests/LibWeb/test-web/results-index.html
jonbgamble: Think of it as optipng, but for whitespace.

Tests/LibWeb/test-web/Results.cpp
Zaggy1024: Ah, yeah, not too surprising it would be within variance, though it's a good sign that it was lower. I suppose I should've suggested profiling the total time spent in that function, though.

Seems fine to me, but if you do want to see the cost I'd still be curious.

Tests/LibWeb/test-web/results-index.css
```diff
@@ -1,598 +1,537 @@
-        :root {
```
gmta: This indentation change should happen in the original commit introducing the file, not this one.

Tests/LibWeb/test-web/TestRunCapture.cpp
gmta: Why do we ignore errors here?

Tests/LibWeb/test-web/results-index.html
```diff
@@ -159,6 +200,13 @@ <h1>
             }
 
             if (test.hasLogs) tabs.push({ id: 'std-logs', label: 'Logs', file: `${test.name}.logs.html`, type: 'text' });
+            if (Array.isArray(test.pids)) {
+                for (const pid of test.pids) {
+                    if ((testsByPid.get(pid)?.length ?? 0) === 0)
+                        continue;
+                    tabs.push({ id: `pid-${pid}`, label: `Pid ${pid}`, pid, currentTestName: test.name, type: 'process' });
```
gmta: Just a nit: having "Pid XXX" as a caption could be clearer imo. Maybe "Context" or "Test context" and mention the PID in the status text when opened?

Tests/LibWeb/test-web/TestRunCapture.cpp
jonbgamble: Because there's nothing we can do about them.

Why is an error returned? Because TRY is handy.

Tests/LibWeb/test-web/results-index.css
jonbgamble: So you want prettier run on the commit that introduces css.

And then as the last commit on the html?

Tests/LibWeb/test-web/results-index.html
jonbgamble: Sure, how about a followup PR

jonbgamble: Maybe `Process` would be clearest.

Tests/LibWeb/test-web/results-index.css
gmta: I'm not saying prettier? As a general rule, you want to make commit diffs as clear as possible. Re-indenting a file that was created in an earlier commit makes it harder to spot differences. By having a clear "no functional changes" commit that just moves the CSS but makes it have the right indentation from that point out, you prevent that potential confusion and makes it easier to review.

Tests/LibWeb/test-web/TestRunCapture.cpp
gmta: I'd want `test-web` to let me know if it's unable to do something though. Why not `warnln()` something?

Tests/LibWeb/test-web/results-index.html
gmta: > Sure, how about a followup PR

It's a caption change - let's not.

gmta: > Maybe `Process` would be clearest.

"Process" or "PID" does not explain the contents of the tab though. It's a view of the context or test sequence in which the test was run.

Tests/LibWeb/test-web/results-index.css
jonbgamble: > Re-indenting a file that was created in an earlier commit

That file was created in the commit where it is indented.

Those eight spaces came from the html the css was copied from. That's in case somone wants to diff the two files manually.

Do you want that whitespace corrected in that initial commit or not?

gmta: > Those eight spaces came from the html the css was copied from. That's in case somone wants to diff the two files manually.

Most diff tools clearly show a difference between whitespace-only and contents diffs, that's why you can/should remove the indentation when creating the file initially (by moving the CSS, yeah).

> Do you want that whitespace corrected in that initial commit or not?

Yes, my preference is to _just_ remove that whitespace, not to run prettier on the entire source just yet. Fine to do that in a separate commit though.

Tests/LibWeb/test-web/results-index.html
jonbgamble: Well do you have a particular favorite between "Context" and "Test context". These are tweaks you can make, youknow.

gmta: Sure, you want me to push to your branch so this PR is updated?

jonbgamble: That sound awesome

jonbgamble:
Damn, the css got pretty mangled in e6a5345806c

I'll dig into that one tomorrow. You don't have to worry about it, that's on me.

jonbgamble:
Well, hopefully the mangled css problem is fixed now. I am not getting better at this.

jonbgamble:
@gmta This PR should be ready for your caption & header edits. totals, pid, whatever you think should go there. You can likely find it in the manifest array.

PR #9070, opened 2026-04-24
LibWeb: Filter by font width before weight in font matching

tcl3:
This PR implements the width filtering step of the font matching algorithm. Without it, system font providers that group all widths under one family could return a condensed variant for font-width: normal, producing visibly narrower text.

On my Linux machine this caused  italic text to look narrower than intended. Here's an example from https://en.wikipedia.org/wiki/Software_bug:

Before:

image

After:

image

I also fixed an issue where `document.fonts.ready` could resolve early due to URL-backed font faces not having their status set to loading in the same way as BufferSource-backed font faces - this caused the test added in the later commit to fail ~5% of the time on my local machine.


jdahlin:
This overlaps with #9032 and this commit https://github.com/LadybirdBrowser/ladybird/pull/9032/changes/50bc86bd36169e5a202c27e2ea6b19e7ed0712c2

Psychpsyo:
> On my Linux machine this caused italic text to look narrower than intended.

Just wanted to add some before & after screenshot from my machine, where this has been affecting almost all text:
image

image

Very glad to see this getting fixed!

PR #8189, opened 2026-02-26
Everywhere: Add initial support for the Tracy profiler

Lubrsi:
This allows us to get explicit zones (complimented with full call stacks on Linux and Windows) plus plot graphs such as memory usage (including GC blocks and live GC heap size), in per-frame boundaries or across the capture lifetime.

Demos:
- Loading a Ladybird newsletter and finding memory allocation churn when playing a video:

https://github.com/user-attachments/assets/12baba31-7a7e-41b8-ade3-1c28a78ff41d

- Loading Google and investigating a long frame, seeing layout update takes a long time and allocates ~1 MB of memory

https://github.com/user-attachments/assets/45716822-4fb4-439d-a5e6-12151d302e15

- Long JS execution frame on Google:
Screenshot 2026-02-26 at 15 47 29

- Ghost frames (i.e. stack traces) from a remote Linux machine with perf_event_paranoid=1:
Screenshot 2026-02-26 at 16 41 06

- Active allocations with stack traces:
Screenshot 2026-02-26 at 16 42 55

- Memory pool categorisation:
Screenshot 2026-02-26 at 16 47 25

PR #9073, opened 2026-04-24
Everywhere: Add initial support for the Tracy profiler

Lubrsi:
This is a revival of https://github.com/LadybirdBrowser/ladybird/pull/8189

This allows us to get explicit zones (complimented with full call stacks on Linux and Windows) plus plot graphs such as memory usage (including GC blocks and live GC heap size), in per-frame boundaries or across the capture lifetime.

Demos:
- Loading a Ladybird newsletter and finding memory allocation churn when playing a video:

https://github.com/user-attachments/assets/12baba31-7a7e-41b8-ade3-1c28a78ff41d

- Loading Google and investigating a long frame, seeing layout update takes a long time and allocates ~1 MB of memory

https://github.com/user-attachments/assets/de9c6577-8007-4c9a-8e85-875e597c0a16

- Long JS execution frame on Google:
Screenshot 2026-02-26 at 15 47 29

- Ghost frames (i.e. stack traces) from a remote Linux machine with perf_event_paranoid=1:
Screenshot 2026-02-26 at 16 41 06

- Active allocations with stack traces:
Screenshot 2026-02-26 at 16 42 55

- Memory pool categorisation:
Screenshot 2026-02-26 at 16 47 25

### DIFF
--- a/AK/kmalloc.cpp
+++ b/AK/kmalloc.cpp
@@ -5,88 +5,27 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
+#include <AK/Assertions.h>
+#include <AK/Tracy.h>
 #include <AK/kmalloc.h>
 
-#if defined(AK_OS_SERENITY)
+#include <cstddef>
+#include <cstring>
 
-#    include <AK/Assertions.h>
-
-// However deceptively simple these functions look, they must not be inlined.
-// Memory allocated in one translation unit has to be deallocatable in another
-// translation unit, so these functions must be the same everywhere.
-// By making these functions global, this invariant is enforced.
-
-void* operator new(size_t size)
-{
-    void* ptr = malloc(size);
-    VERIFY(ptr);
-    return ptr;
-}
-
-void* operator new(size_t size, std::nothrow_t const&) noexcept
-{
-    return malloc(size);
-}
-
-void operator delete(void* ptr) noexcept
-{
-    return free(ptr);
-}
-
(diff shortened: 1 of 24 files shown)


AK/kmalloc.cpp
```diff
@@ -200,30 +209,116 @@ void* ladybird_rust_realloc(void* ptr, size_t old_size, size_t new_size, size_t
...
-#    endif
-
 #endif
+
+// However deceptively simple these functions look, they must not be inlined.
+// Memory allocated in one translation unit has to be deallocatable in another
+// translation unit, so these functions must be the same everywhere.
+// By making these functions global, this invariant is enforced.
+
+void* operator new(size_t size)
```
R-Goc, line 282: Just a nit. as probably no one uses them, but this is missing the aligned variants of operator new and delete. You can look at https://github.com/microsoft/mimalloc/blob/main/include/mimalloc-new-delete.h. Also the paths taken by mi_new and mi_new_nothrow are different than what mi_malloc does. Not sure how much that matters but it might be worth taking a look at.

Services/ImageDecoder/main.cpp
```diff
@@ -15,6 +16,7 @@
 
 ErrorOr<int> ladybird_main(Main::Arguments arguments)
 {
+    TRACY_SET_PROGRAM_NAME("ImageDecoder");
```
alimpfard, line 19: 🤔 can't LibMain set this from argv0 automatically?

Issue #8700, opened 2026-03-31
Crash in Web::HTML::Navigation::update_the_navigation_api_entries_for_a_same_document_navigation


jdahlin:
1. Open up https://www.ladybird.org/
2. Click on About: https://www.ladybird.org/#about
3. Click on News: https://www.ladybird.org/news/
4. Click on Back two times

```
Qt: Session management error: None of the authentication protocols specified are supported
VERIFICATION FAILED: m_current_entry_index != -1 at /home/jdahlin/ladybird/gtk-fe/Libraries/LibWeb/HTML/Navigation.cpp:1484
Stack trace (most recent call first):
#0  0x00007f7b8cd392b8 in Web::HTML::Navigation::update_the_navigation_api_entries_for_a_same_document_navigation(GC::Ref<Web::HTML::SessionHistoryEntry>, Web::Bindings::NavigationType) at /home/jdahlin/ladybird/gtk-fe/Libraries/LibWeb/HTML/Navigation.cpp:1484:9
#1  0x00007f7b8c97ad5f in Web::DOM::Document::update_for_history_step_application(GC::Ref<Web::HTML::SessionHistoryEntry>, bool, unsigned long, unsigned long, AK::Optional<Web::Bindings::NavigationType>, AK::Optional<AK::Vector<GC::Ref<Web::HTML::SessionHistoryEntry>, 0ul, (AK::FastLastAccess)0> >, GC::Ptr<Web::HTML::SessionHistoryEntry>, bool) at /home/jdahlin/ladybird/gtk-fe/Libraries/LibWeb/DOM/Document.cpp:5761:93
#2  (inlined)          in operator() at /home/jdahlin/ladybird/gtk-fe/Libraries/LibWeb/HTML/TraversableNavigable.cpp:917:78
#3  0x00007f7b8ce15bf2 in operator() at /home/jdahlin/ladybird/gtk-fe/Libraries/LibWeb/HTML/TraversableNavigable.cpp:922:32
#4  (inlined)          in AK::Function<void ()>::operator()() const at /home/jdahlin/ladybird/gtk-fe/AK/Function.h:147:29
#5  0x00007f7b8cbc3f4f in Web::HTML::Task::execute() at /home/jdahlin/ladybird/gtk-fe/Libraries/LibWeb/HTML/EventLoop/Task.cpp:47:24
[...]

PR #9089, opened 2026-04-25
LibWeb: Prepend session history index zero for navigation api

ercpham:
Navigating backward may crash because entries_for_navigation_api is missing the first element.

### DIFF
--- a/Libraries/LibWeb/HTML/TraversableNavigable.cpp
+++ b/Libraries/LibWeb/HTML/TraversableNavigable.cpp
@@ -1397,8 +1397,8 @@ Vector<NonnullRefPtr<SessionHistoryEntry>> TraversableNavigable::get_session_his
     // 6. Let i be startingIndex − 1.
     auto i = starting_index - 1;
 
-    // 7. While i > 0:
-    while (i > 0) {
+    // 7. While i >= 0:
+    while (i >= 0) {
         auto& entry = raw_entries[static_cast<unsigned>(i)];
         // 1. If rawEntries[i]'s document state's origin is not same origin with startingOrigin, then break.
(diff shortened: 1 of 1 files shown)


ercpham:
Fixes #8700

Libraries/LibWeb/HTML/TraversableNavigable.cpp
```diff
@@ -1397,8 +1397,8 @@ Vector<NonnullRefPtr<SessionHistoryEntry>> TraversableNavigable::get_session_his
     // 6. Let i be startingIndex − 1.
     auto i = starting_index - 1;
 
-    // 7. While i > 0:
-    while (i > 0) {
```
ADKaster, line 1401: `7.` here is step 7 from the spec algorithm https://html.spec.whatwg.org/multipage/browsing-the-web.html#getting-session-history-entries-for-the-navigation-api

This patch is not a valid fix for the referenced issue.

We cannot just change spec steps willy nilly. If the crash is here, then there is logic somewhere else in navigables that is wrong. Or, there are other spec algorithms that are wrong.

Changing this step would require a spec issue opened to whatwg/html, along with a comment explaining *why*.

ADKaster (changes requested):

Libraries/LibWeb/HTML/TraversableNavigable.cpp
R-Goc, line 1386: Maybe unrelated to this issue but we are missing updating max_step here.

Libraries/LibWeb/HTML/TraversableNavigable.cpp
ercpham: Ah, understood.

PR #9094, opened 2026-04-25
LibWeb/Bindings: Port handwritten dictionaries to generated Bindings types

### LINKED ISSUE #1354: Bindings: Generate separate header files for enums
WebIDL files that contain only enums currently don't actually output any header files. For example, HTML/NavigationType.idl

WebIDL files that contain enums and interfaces insert the enum definition into the Prototype header. This means that when an interface uses an enum from its IDL file in its public interface, we must include the prototype file in the header. This drastically increases the number of header files that must be included in our translation units.

We should instead try to generate separate header files for each enum. Unfortunately this makes it difficult to know in the build system what all the files are that will be generated, and whether there's any enum-specific cpp files that need to be generated.

Still some design required for this.

shannonbooth:
This MR changes our IDL bindings in a few different significant ways:
- Dictionary definitions are now generated, instead of needing to be handwritten. This was done so that we were not inconsistent between enums and dictionaries (which would otherwise cause circular dependencies as part of the other changes here).
- The generated IDL conversion code for dictionaries and enums is now put into functions per enum/dictionary definitions, instead of being inlined by the code generator any time this conversion needed to happen. This makes the generated code nicer to read by not having _everything_ inlined, and less generated code overall, but the main purpose here is to allow for calling IDL conversion code outside of bindings folders, whenever a spec says "converted to an IDL value". I think we can go further here, would will also allow for tidy up of the code generate itself, but that's not done here.
- IDL codegen is now generated per IDL file, rather than only for IDL interfaces. This enables the change above, and also fixes #1354

This is starting to get closer to the shape the IDL generator should be in terms of high level outputs IMO. Main points to fix after this MR are:

1. Convert to python
2. Unify the handling of "optional" values with default parameters and "variadic"s at top level of type conversion. Currently this is special cased in every individual type conversion with bugs.
3. Better include tracking, instead of including what _might_ be used, include what _is_ used.
4. Start factoring more understandable way. 2 should help with that, alongside many other things.


Meta/Lagom/Tools/CodeGenerators/LibWeb/BindingsGenerator/IDLGenerators.cpp
```diff
@@ -2109,9 +2180,11 @@ static void generate_wrap_statement(SourceGenerator& generator, ByteString const
         optional_uses_value_access = is<UnionType>(type)
             || type.is_string()
             || type.name().is_one_of("sequence"sv, "FrozenArray"sv)
+            || type.name() == "record"sv
```
AtkinsSJ: Nit: This and `Promise` below could go in the `is_one_of()` call above.

Meta/Lagom/Tools/CodeGenerators/LibWeb/BindingsGenerator/IDLGenerators.cpp
```diff
@@ -3291,6 +3374,306 @@ JS::Value @dictionary.name:snakecase@_to_value(JS::Realm& realm, @dictionary.cpp
...
+        return ByteString::formatted("{} {{}}", idl_type_name_to_cpp_type(**object_type, context).name);
+    }
+
+    if (default_value.starts_with("\""sv) && default_value.ends_with("\""sv)) {
+        auto default_string_value = default_value.substring_view(1, default_value.length() - 2);
+
+        auto enumeration_type = types.find_if([&](auto const& type) {
+            if (!context.enumerations.contains(type->name()))
+                return false;
+            return context.enumerations.find(type->name())->value.translated_cpp_names.contains(default_string_value);
```
AtkinsSJ: Nit: You could avoid the double context.enumerations hash-table lookup by calling `find()` first and checking if it returned anything.

Meta/Lagom/Tools/CodeGenerators/LibWeb/BindingsGenerator/IDLGenerators.cpp
```diff
@@ -3291,13 +3374,313 @@ JS::Value @dictionary.name:snakecase@_to_value(JS::Realm& realm, @dictionary.cpp
...
+    if (!dictionary->value.parent_name.is_empty())
+        generator.append(" : public @dictionary.parent_cpp_type@");
+    generator.append(" {\n");
+
+    for (auto const& member : members) {
+        auto member_generator = generator.fork();
+        member_generator.set("member.type", dictionary_member_cpp_type(context, member));
+        member_generator.set("member.name", make_input_acceptable_cpp(member.name.to_snakecase()));
+        member_generator.set("member.initializer", dictionary_member_initializer(context, member));
+        member_generator.append("   @member.type@ @member.name@@member.initializer@;\n");
```
AtkinsSJ: Mega-nit: This only indents by 3 spaces. 😱

AtkinsSJ (commented):
I only just got to the end of the generating-dictionary structs commit, so this isn't a full review. I was originally hoping to merge without waiting for these nits but it currently has conflicts so I guess if you feel like it you can implement them!

Meta/Lagom/Tools/CodeGenerators/LibWeb/BindingsGenerator/IDLGenerators.cpp
```diff
@@ -362,148 +362,158 @@ static void add_module_include_dependency(Vector<Module const*>& modules_to_incl
...
 
     if (type.is_union()) {
         for (auto const& member_type : type.as_union().member_types())
-            collect_interface_include_dependencies(interface, *member_type, modules_to_include, paths_included);
+            collect_include_dependencies(context, *member_type, collector);
     }
 }
 
-static void emit_includes_for_all_dependencies(auto& interface, auto& generator, bool is_iterator = false, bool is_async_iterator = false)
+static void emit_includes_for_all_dependencies(auto& interface, auto& generator, bool is_iterator = false, bool is_async_iterator = false, bool include_own_interface_module = true, bool emit_same_module_includes = true)
```
AtkinsSJ: Nit: This is a lot of bools, maybe switch to some named enums.

AtkinsSJ (commented):
One final nit, and now I've actually reviewed everything. 😅 (Caveat: Not sure my brain fully comprehended everything, it's a lot.)

Very nice seeing lengthy conversion code disappear, in favour of a spec idl file. 😁

PR #9096, opened 2026-04-25
Basic print support

jdahlin:
Add end-to-end print support

- Add `@media` print support in LibWeb, print-specific CSS now applies during printing.
- Implement window.print() per spec, including correct beforeprint+afterprint event dispatch and A4 viewport rendering at 96 DPI.
- Add a print rendering pipeline in TraversableNavigable that renders a print display list on the rendering thread and restores screen layout afterward.
- Add print dialog support (Ctrl+P) in the GTK and Qt frontends.
- Add support for print layout tests and add a basic test ensuring `@media print` is honored

Example wikipedia page:
[output.pdf](https://github.com/user-attachments/files/27086678/output.pdf)

### DIFF
--- a/Libraries/LibWeb/HTML/TraversableNavigable.cpp
+++ b/Libraries/LibWeb/HTML/TraversableNavigable.cpp
@@ -11,10 +11,12 @@
 #include <LibGfx/SkiaBackendContext.h>
 #include <LibWeb/Bindings/MainThreadVM.h>
 #include <LibWeb/DOM/Document.h>
+#include <LibWeb/DOM/Event.h>
 #include <LibWeb/Geolocation/GeolocationCoordinates.h>
 #include <LibWeb/HTML/BrowsingContext.h>
 #include <LibWeb/HTML/BrowsingContextGroup.h>
 #include <LibWeb/HTML/DocumentState.h>
+#include <LibWeb/HTML/EventNames.h>
 #include <LibWeb/HTML/History.h>
 #include <LibWeb/HTML/NavigableContainer.h>
 #include <LibWeb/HTML/Navigation.h>
@@ -27,6 +29,7 @@
 #include <LibWeb/Layout/Viewport.h>
 #include <LibWeb/Page/Page.h>
 #include <LibWeb/Painting/PaintableBox.h>
+#include <LibWeb/Painting/ViewportPaintable.h>
 #include <LibWeb/Platform/EventLoopPlugin.h>
 #include <LibWeb/Platform/Timer.h>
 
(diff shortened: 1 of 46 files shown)


Libraries/LibWeb/HTML/Window.cpp
```diff
@@ -990,6 +990,30 @@ void Window::blur()
     // The Window blur() method steps are to do nothing.
 }
 
+// https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#printing
+void Window::print()
+{
+    // 1. Let document be this's associated Document.
+    auto& doc = *document();
```
shannonbooth: Flyby comment I noticed on a scan through (not a proper look by any means)
```cpp
    auto& document = associated_document();
```

Psychpsyo:
It'd be neat if a page started where the previous page ended:
image

Libraries/LibWeb/HTML/TraversableNavigable.cpp
```diff
@@ -1776,6 +1779,91 @@ void TraversableNavigable::set_emulated_position_data(Geolocation::EmulatedPosit
...
+    // Fire beforeprint before we change any rendering state.
+    if (auto window = document->window())
+        window->dispatch_event(DOM::Event::create(window->realm(), EventNames::beforeprint));
+
+    // A4 paper at 96 DPI: 210mm × 297mm → 794 × 1123 px.
+    // Using the real page height keeps vh units correct and prevents height:100% / height:100vh elements
+    // from inflating the content to the full viewport, which would produce an enormous bitmap. Content that
+    // overflows a single page flows past the viewport naturally; the dialog then slices the resulting
+    // bitmap into per-page strips.
+    static constexpr CSSPixels k_paper_width { 794 };
```
Psychpsyo: Why `k_` here?

Libraries/LibWeb/HTML/TraversableNavigable.cpp
```diff
@@ -1776,6 +1779,91 @@ void TraversableNavigable::set_emulated_position_data(Geolocation::EmulatedPosit
...
+
+    auto& page_obj = page();
+    auto& client = page_obj.client();
+
+    // Fire beforeprint before we change any rendering state.
+    if (auto window = document->window())
+        window->dispatch_event(DOM::Event::create(window->realm(), EventNames::beforeprint));
+
+    // A4 paper at 96 DPI: 210mm × 297mm → 794 × 1123 px.
+    // Using the real page height keeps vh units correct and prevents height:100% / height:100vh elements
```
Psychpsyo: Can we get a spec comment for this?
The correct spec to cite is either https://drafts.csswg.org/css-values-4/#viewport-relative-lengths or the spec for the [initial containing block](https://www.w3.org/TR/CSS2/visudet.html#containing-block-details) that it links to.

And a test for this would be nice.

Libraries/LibWeb/HTML/TraversableNavigable.cpp
jdahlin: no reason, I'll change it.

jdahlin:
> It'd be neat if a page started where the previous page ended: image

Indeed, thats perfect for an upcoming PR, this one focuses on the machinery to get something printed at all.

Services/WebContent/ConnectionFromClient.cpp
```diff
@@ -1069,6 +1087,32 @@ void ConnectionFromClient::request_internal_page_info(u64 page_id, WebView::Page
...
     }
 
+    if (has_flag(type, WebView::PageInfoType::PrintLayoutTree)) {
+        if (!builder.is_empty())
+            builder.append("\n"sv);
+        auto& web_page = page->page();
+        if (web_page.top_level_traversable_is_initialized()) {
+            auto navigable = web_page.top_level_traversable();
+            if (auto document = navigable->active_document()) {
+                static constexpr Web::CSSPixels k_paper_width { 794 };
```
Psychpsyo: Same thing about k_ here. We're not building khtml afterall, lol

jdahlin:
I realized that a few major things needs fixing before this should land
- support user settings in the print dialog, changes ipc and quite a bit in frontends
- tell skia to render to pdf instead of a bitmap

Converting to draft until they are properly implemented.

jdahlin:
While this works, I realize that everything about it is wrong:
- frontend/platform specific print logic and dialogs
- printing using bitmaps instead of pdfs

I am closing this while figuring out how to do this in a better way that works across all frontends

Psychpsyo:
>     * printing using bitmaps instead of pdfs

Why is printing images wrong? If printing to pdf, sure. But when printing to an actual printer or any other format, I'd expect images to deliver more correct results than pulling the whole document through pdf first and potentially incurring weird conversion artifacts.

jdahlin:
> Why is printing images wrong? If printing to pdf, sure. But when printing to an actual printer or any other format, I'd expect images to deliver more correct results than pulling the whole document through pdf first and potentially incurring weird conversion artifacts.

Firefox, Chrome, and Safari generally preserve vector content where possible (via cairo, SkiaPDF or Quartz I believe), delegating final rasterization to the OS or printer RIP, and only rasterize selectively for effects that require flattening.

PR #9097, opened 2026-04-25
Clang-format: Prevent space in curly braces

NiccoloAntonelliDziri:
Clang-format 22 added SpaceInEmptyBraces and it defaults to Always resulting in `{ }`. Ladybird uses `{}`. Now set to Never.

https://clang.llvm.org/docs/ClangFormatStyleOptions.html#spaceinemptybraces


trflynn89:
We use clang-format-21 in CI, and unfortunately the config file is not forwards-compatible. When we upgrade to 22, we will likely set this option to `Block`.

PR #9099, opened 2026-04-25
LibWeb: Position list marker to the right of right-to-left text

Psychpsyo:
The rebaselined tests are due to me subtracting the marker distance from inside-positioned markers also. I've manually checked the results, they are still visually correct. (The difference being that the list item's box now doesn't extend 1 marker distance too far to the right anymore)

Here's some before and after pictures from when I go to https://ar.wikipedia.org:
Before:
image

After:
image

### DIFF
--- a/Tests/LibWeb/Layout/input/list-marker-rtl.html
+++ b/Tests/LibWeb/Layout/input/list-marker-rtl.html
@@ -1,0 +1,4 @@
+<!DOCTYPE html>
+<ul dir="rtl">
+	<li>___________</li>
+</ul>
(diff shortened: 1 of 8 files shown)


Tests/LibWeb/Layout/input/list-marker-rtl.html
```diff
@@ -0,0 +1,4 @@
+<!DOCTYPE html>
+<ul dir="rtl">
+	<li>___________</li>
+</ul>
```
gmta, line 4: This would probably work better as a layout (or display list?) test.

Psychpsyo: True, made it a layout test now.

gmta (approved):

PR #9102, opened 2026-04-25
Build: Disable Skia's resource tracing in debug builds

jonbgamble:
Add a local Skia 144 overlay port and a Ladybird-specific patch that comments out SK_TRACE_MANAGED_RESOURCES in GrManagedResource.h.

Most of the diff churn is the copied overlay directory. The custom parts are the ladybird-disable-skia-managed-resource-trace.patch and its entry in portfile.cmake.

entry in portfile.cmake.


jdahlin:
Would be nice to mention _why_ this change is needed, _what_ it does can be inferred by reading the diff/PR.

jonbgamble:
It prevents up to 4000+ lines of refcount nag traces, sometimes flushing any runtime logs you may have been interested in from the scrollback buffer. You'd have to run debug builds to understand why I care about this one.

kalenikaliaksandr:
> It prevents up to 4000+ lines of refcount nag traces, sometimes flushing any runtime logs you may have been interested in from the scrollback buffer. You'd have to run debug builds to understand why I care about this one.

sure, please put this explanation in the commit description

PR #9103, opened 2026-04-25
LibWeb: Implement more correct Unicode BiDi algorithm via ICU

Psychpsyo:
Imports 7 tests from WPT, 4 of which are new passes.

This defers most of the actual details of this algorithm to the ICU since our old implementation was a big ad-hoc mess and the expected behavior is quite complicated: https://www.unicode.org/reports/tr9/
The line-reordering is still handled on our end, partly by reordering fragments and partly by rasterizing RTL text strings correctly. ('in reverse' compared to memory order)
I'm not sure if we can outsource this to the ICU also, might look into that at some later point, especially since some of that is still kinda broken in some way, as shown by the missing WPT passes here. (`<` and `>` are getting rendered the wrong way around there.)

A bunch of tests needed rebaselining since the way we split text nodes now puts the boundaries in somewhat different positions. This is expected, has only very tiny effects on the actual rasterization and we need it for proper fragment reordering.


Libraries/LibUnicode/BiDi.h
```diff
@@ -0,0 +1,28 @@
...
+ * Copyright (c) 2026-present, Psychpsyo <psychpsyo@gmail.com>
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#pragma once
+
+#include <AK/String.h>
+#include <unicode/ubidi.h>
+#include <unicode/unistr.h>
```
trflynn89: Please keep ICU headers limited to .cpp files. You can forward-declare the types you need.

Libraries/LibUnicode/BiDi.h
```diff
@@ -0,0 +1,28 @@
...
+class BiDiParagraph {
+public:
+    BiDiParagraph(AK::Utf16String& text, UBiDiLevel embedding_level);
+    ~BiDiParagraph();
+
+    UBiDiLevel get_bidi_level_at(size_t index);
+
+private:
+    BidiParagraphContents* m_contents;
+    UBiDi* m_bidi_paragraph;
```
trflynn89: Always initialize plain types:
```c++
    BidiParagraphContents* m_contents { nullptr };
    UBiDi* m_bidi_paragraph { nullptr };
```

Libraries/LibUnicode/BiDi.h
```diff
@@ -0,0 +1,28 @@
...
+#include <AK/String.h>
+#include <unicode/ubidi.h>
+#include <unicode/unistr.h>
+
+namespace Unicode {
+
+struct BidiParagraphContents;
+class BiDiParagraph {
+public:
+    BiDiParagraph(AK::Utf16String& text, UBiDiLevel embedding_level);
```
trflynn89: Why a mutable reference?
(it looks like a Utf16View would suffice here?)

Libraries/LibWeb/Layout/InlineFormattingContext.cpp
```diff
@@ -348,7 +352,33 @@ void InlineFormattingContext::generate_line_boxes()
...
+            // FIXME: Verify that this is the correct value to use when 'unicode_bidi' is set to 'plaintext'
+            UBiDiLevel embedding_level = UBIDI_DEFAULT_LTR;
+
+            // There is, however, one exception: when the computed 'unicode-bidi' of the paragraph’s containing block is
+            // 'plaintext', the Unicode heuristics in P2 and P3 are used as described in [UAX9], without the HL1 override.
+            if (m_context_box->computed_values().unicode_bidi() != CSS::UnicodeBidi::Plaintext) {
+                embedding_level = direction == CSS::Direction::Ltr ? 0 : 1;
+            }
+
+            bidi_paragraph = new Unicode::BiDiParagraph(text.value(), embedding_level);
```
trflynn89: No manual `new`/`delete` in 2026 please.

Either wrap in an OwnPtr or, better yet, add `NonnullOwnPtr<BiDiParagraph> BiDiParagraph::create(...)` (and make the constructor private).

Libraries/LibUnicode/BiDi.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+namespace Unicode {
+
+struct BidiParagraphContents {
+    icu::UnicodeString m_text;
+};
+
+BiDiParagraph::BiDiParagraph(AK::Utf16String& text, UBiDiLevel embedding_level)
+{
+    m_contents = new BidiParagraphContents(text.has_ascii_storage() ? Unicode::icu_string(text.ascii_view()) : Unicode::icu_string(text.utf16_view()));
```
trflynn89: We have `icu_string` for this in ICU.h.

trflynn89: And again, no naked `new` please. Does anything even delete this?

Libraries/LibUnicode/BiDi.h
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright (c) 2026-present, Psychpsyo <psychpsyo@gmail.com>
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#pragma once
+
+#include <AK/String.h>
```
trflynn89: Unused header

trflynn89 (commented):
Just a very quick look at the LibUnicode stuff

Libraries/LibUnicode/BiDi.cpp
```diff
@@ -0,0 +1,45 @@
...
+#include <LibUnicode/BiDi.h>
+#include <LibUnicode/ICU.h>
+
+namespace Unicode {
+
+struct BidiParagraphContents {
+    icu::UnicodeString m_text;
+};
+
+BiDiParagraph::BiDiParagraph(AK::Utf16String& text, UBiDiLevel embedding_level)
```
shannonbooth: Driveby nit - but we're not consistent here between BiDi and Bidi. Looks like most places at the moment are using Bidi

Psychpsyo: The ICU types use BiDi. For our types, I'd prefer Bidi so that when snake-cased, it doesn't end up as bi_di. I hope that's ok. (and I've changed the few non-ICU things that were using BiDi to be Bidi instead.)

shannonbooth: I don't mind either way, as long as we're internally consistent. Could the file name also be aligned with the ladybird casing?

Libraries/LibUnicode/Bidi.h
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#pragma once
+
+#include <AK/Utf16View.h>
+
+class UBiDi;
+typedef unsigned char UBiDiLevel;
```
shannonbooth, line 12: One more case of the casing here

Libraries/LibUnicode/BiDi.cpp
trflynn89: > The ICU types use BiDi. For our types, I'd prefer Bidi so that when snake-cased, it doesn't end up as bi_di.

I don’t think we should be that overly strict with snake casing. `BiDi` and `bidi_foo` are both pretty standard.

Libraries/LibUnicode/Bidi.h
trflynn89: These are forward declarations of ICU types

Libraries/LibUnicode/BiDi.cpp
shannonbooth: It's more just that for when the casing is mixed up for internal types I end up typing the wrong thing a lot. Don't feel strongly about it, just think it would be nice

Psychpsyo: Just did a search through the code and we already have `Bidi` in a bunch of places (Also some `BiDi`, but that seems to be due to spec comments about WebDriver BiDi).
I've just renamed `BiDi.h` and `BiDi.cpp` for now. If this becomes an annoyance, there can always be a PR made that switches them all over.

Psychpsyo:
Took care of all the feedback now. Sorry for the original version of the PR being... not great.

gmta:

PR #9106, opened 2026-04-26
Meta+LibWeb: Initial CSS value parsing code generation

Calme1709:
This PR implements code generation for parsing of CSS value types, it does this by parsing the CSS value definition grammar associated with a value type and generating the associated `CSS::Parser::Parser::parse_foo_value` method.

In the future this can be expanded to generate parsers for CSS properties and descriptor values as well. Although any values with additional requirements not representable in the CSS value definition grammar (e.g. bespoke resultant StyleValue types, default values, etc) will still need to be implemented manually.

This PR is primarily meant as a proof of concept and is thus very incomplete, only supporting a small subset of the CSS value definition grammar, namely:
 - Keywords i.e. `<foo> = bar`
 - Types (including custom ident blacklists and range restrictions) i.e. `<foo> = <bar>`, `<foo> = <custom-ident ![bar]>`, `<foo> = <number [1, 1000]>`
 - Alternatives i.e. `<foo> = bar | baz`

Value types are defined in `ValueTypes.json`, In the future we should switch to using a better file format for this, i.e. one that supports directly pasting [CSS grammar production blocks](https://drafts.csswg.org/css-values-4/#css-grammar-production-block) and has support for inline comments, but we use JSON for now for simplicity's sake.


Meta/Utils/CSSGrammar/Parser/component_values.py
```diff
@@ -1,18 +1,68 @@
...
+def bound_value_to_string(value: float) -> str:
+    if value == -inf:
+        return "-∞"
+    if value == inf:
+        return "∞"
+    return str(value)
+
+
+def is_dimension_type(type_name: str) -> bool:
+    # NB: Keep this up to date with the list of dimensions in Units.json
```
AtkinsSJ, line 24: 🤔 How painful would it be to actually read Units.json instead of hard-coding this list?

Also, we don't have `decibel` in there yet, though we probably should.

Meta/Utils/CSSGrammar/generator.py
```diff
@@ -0,0 +1,114 @@
...
+        # NB: Pass the accepted range twice for dimension-percentage mixes, once for the dimension and once for the percentage.
+        if is_dimension_percentage_mix_type(type.name):
+            additional_arguments += accepted_range
+
+    out.write(f"auto {cpp_name} = parse_{type_name}_value(tokens{additional_arguments});\n")
+
+
+def generate_css_parser_expression_for_keyword_component_value(out: TextIO, cpp_name: str, keyword: Keyword) -> None:
+    keyword_name = title_casify(keyword.value)
+    out.write(f"auto {cpp_name} = parse_specific_keyword_value(tokens, Keyword::{keyword_name});\n")
```
AtkinsSJ, line 53: I'm a little concerned about doing this when there are several possible keywords. It's a bit wasteful to repeatedly parse a keyword value, check it, throw it away, rewind the token stream, and then do the same thing again.

A `parse_specific_keyword_value()` version that takes a list of accepted keywords would be nice for that, but would require the parser/generator to combine keyword nodes together.

Absolutely not a blocker right now.

AtkinsSJ (approved):
Very cool to see this! Also slightly sad as I'd hoped my work on `<syntax>` parsing would lead to this, but that was before we switched to Python for generators. 😅 

At first I questioned embedding all the test inputs/expectations in the test script, but... that really does make more sense than separate input/expectation files.

PR #9107, opened 2026-04-26
LibWeb: Don't let shape-triggered @font-face rules block page load

jonbgamble:
Deferred @font-face rules were being counted as critical_subresources when they had not started any fetch and might never be needed. This change makes the loading state decision come from the subresource itself.

The errror was observed at https://news.google.com where only a single section appeared, another showed as a placeholder, and the rest were never loaded. This resulted from a non-default unicode-range @font-face which kept the critical stylesheet open, blocking parser completion.

### DIFF
--- a/Libraries/LibWeb/CSS/CSSFontFaceRule.cpp
+++ b/Libraries/LibWeb/CSS/CSSFontFaceRule.cpp
@@ -240,4 +240,11 @@ void CSSFontFaceRule::set_parent_style_sheet(CSSStyleSheet* parent_style_sheet)
         m_parent_style_sheet->add_critical_subresource(*this);
 }
 
+bool CSSFontFaceRule::should_block_stylesheet_while_unloaded() const
+{
+    if (!m_css_connected_font_face)
+        return false;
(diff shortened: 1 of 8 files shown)


jdahlin:
Would it be possible to break out into two commits?

- commit 1: Add a test with a known failure
- commit 2: Fix the failure and test now passes with success

That way it is easier to validate behavior before and after, to ensure the changes are correct.

jonbgamble:
~~AFAIK that is not possible under ladybird rules, every commit must pass CI.~~

jdahlin:
> AFAIK that is not possible under ladybird rules, every commit must pass CI.

I meant that you assert the failure, like having FAILED written in a text file.

jonbgamble:
Then yes, it would be possible. But I'm not going to do it unless they ask.

AtkinsSJ:
To clarify: Yes, having a text test output "thing failed", then making it pass is good. "Every commit must pass CI" means the test expectation must match what currently happens, not that every part of every test must behave perfectly.

For example, most of the imported WPT tests have some kind of "FAIL" results in them, and that's fine, it just means we have to update those when new parts of it get fixed.

gmta:
Seems like this was closed because it was superseded by #9136

PR #9110, opened 2026-04-26
Meta: Port GenerateWindowOrWorkerInterfaces to Python 

shannonbooth:
This replaces the C++ GenerateWindowOrWorkerInterfaces generator with a
python3 port.

The generator uses the new python WebIDL parser and writes directly to a
TextIO instance instead of porting SourceGenerator.

For validation, the python generator was diffed against the existing C++
generator over the full LibWeb IDL input set. The generated output is
behaviorally identical, and the only changes are whitespace changes
(all improvements).


Meta/Generators/generate_window_or_worker_interfaces.py
```diff
@@ -0,0 +1,614 @@
...
+            return {"LayoutWorklet"}
+        if candidate == "PaintWorklet":
+            return {"PaintWorklet"}
+        if candidate == "Worklet":
+            return {"Worklet"}
+        return None
+
+    parsed_candidate = parse_single_candidate(exposed_value)
+    if parsed_candidate is not None:
+        return parsed_candidate
```
jdahlin: Perfect candidate for walrus operator, which improves readability

```suggestion
if (parsed_candidate := parse_single_candidate(exposed_value)) is not None:
        return parsed_candidate
```

Meta/Generators/generate_window_or_worker_interfaces.py
```diff
@@ -0,0 +1,614 @@
...
+
+    if exposed_value.startswith("(") and exposed_value.endswith(")"):
+        exposures: Set[str] = set()
+        for candidate in exposed_value[1:-1].split(","):
+            parsed_candidate = parse_single_candidate(candidate.strip())
+            if parsed_candidate is None:
```
jdahlin: ditto

Meta/Generators/generate_window_or_worker_interfaces.py
```diff
@@ -0,0 +1,614 @@
...
+                raise RuntimeError(
+                    f"Unknown Exposed attribute candidate {candidate.strip()} in {exposed_value} in {interface_name}"
```
jdahlin:
```suggestion
                    f"Unknown Exposed attribute candidate {candidate.strip()} in {exposed_value!r} in {interface_name}"
```

is usually helpful to see the type of the variable when its not clear

Meta/Generators/generate_window_or_worker_interfaces.py
```diff
@@ -0,0 +1,614 @@
...
+                )
+            exposures.update(parsed_candidate)
+
+        if exposures:
+            return exposures
+
+    raise RuntimeError(f"Unknown Exposed attribute {exposed_value} in {interface_name}")
```
jdahlin:
```suggestion
    raise RuntimeError(f"Unknown Exposed attribute {exposed_value!r} in {interface_name}")
```

Same here for consistency

Meta/Generators/generate_window_or_worker_interfaces.py
```diff
@@ -0,0 +1,614 @@
...
+
+
+def add_to_interface_sets(interface: Interface, interface_sets: InterfaceSets) -> None:
```
jdahlin: Probably better as a method on InterfaceSets

Meta/Generators/generate_window_or_worker_interfaces.py
```diff
@@ -0,0 +1,614 @@
...
+        writer(output_file, *args)
+
+
+def main() -> int:
+    arguments = parse_arguments()
+    output_directory = Path(arguments.output_path)
+    output_directory.mkdir(parents=True, exist_ok=True)
+
+    interface_sets = InterfaceSets(
+        intrinsics=[],
```
jdahlin: if you add this to the dataclass fields:

```python
... = dataclass.field(factory=list)
```

You will not need to initialize them here, which imho is cleaner

Meta/Generators/generate_window_or_worker_interfaces.py
```diff
@@ -0,0 +1,614 @@
...
+        parts.append(character.lower())
+    return "".join(parts)
+
+
+def parse_arguments() -> argparse.Namespace:
+    argument_parser = argparse.ArgumentParser()
+    argument_parser.add_argument(
+        "-o",
+        "--output-path",
+        required=True,
```
jdahlin, line 86: If you add `type=pathlib.Path`  here, you will not need to convert it manually later

Meta/Generators/generate_window_or_worker_interfaces.py
```diff
@@ -0,0 +1,614 @@
...
+        help="Path to output generated files into",
+    )
+    argument_parser.add_argument("paths", nargs="+", help="Paths of every IDL file that could be Exposed")
```
jdahlin: ditto

Meta/Utils/webidl.py
```diff
@@ -0,0 +1,600 @@
...
+        if active_quote:
+            if character == active_quote:
+                active_quote = ""
+            continue
+
+        if character in ('"', "'"):
+            active_quote = character
+            continue
+
+        if character == "(":
```
jdahlin: This code below is duplicated 3 times, can that be avoided?

Meta/Generators/generate_window_or_worker_interfaces.py
```diff
@@ -0,0 +1,614 @@
...
+#
+# SPDX-License-Identifier: BSD-2-Clause
+
+
+import argparse
+import sys
+
+from dataclasses import dataclass
+from pathlib import Path
+from typing import List
```
jdahlin, line 14: I thought we could use the new style types in Python 3.10+, to avoid these imports.

Meta/Utils/webidl_parser.py
```diff
@@ -0,0 +1,600 @@
+# Copyright (c) 2026-present, the Ladybird developers.
```
jdahlin, line 1: maybe call this webidlparser.py or so, since I assume there will be a few more files once more functionality is moved over.

Meta/Generators/generate_window_or_worker_interfaces.py
shannonbooth: I believe we're trying to keep compatibility with 3.9 since that's what is default on MacOS

jdahlin: That should probably be revisited as 3.9 reached end of life 2025-10-31. Depending on [uv](https://docs.astral.sh/uv/) will make it trivial to support latest python on all OSes.

But let's maintain 3.9 here and figure out how to migrate to a newer one in another PR.

AtkinsSJ (approved):

PR #9112, opened 2026-04-26
LibWeb: Skip backward sibling invalidation when no child needs it

tcl3:
We now track when a parent has a child affected by a backward structural pseudo-class. These are selectors whose match result for an element can depend on siblings after that element, such as `:last-child`, `:only-child`, `:last-of-type`, `:only-of-type`, `:nth-last-child`, and `:nth-last-of-type`.

When inserting or removing a node, previous siblings only need style invalidation if one of them was matched against such a selector. Use the parent-level flag to skip the previous-sibling walk when no child under that parent can be affected.

This reduces the time to display the results on https://wpt.fyi from ~150s to ~10s on my machine.

### DIFF
--- a/Libraries/LibWeb/DOM/Document.h
+++ b/Libraries/LibWeb/DOM/Document.h
@@ -859,6 +859,7 @@ public:
         u64 element_style_noop_recomputations { 0 };
         u64 element_inherited_style_recomputations { 0 };
         u64 element_inherited_style_noop_recomputations { 0 };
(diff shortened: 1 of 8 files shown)


awesomekling:
Please add test coverage that shows (using the style invalidation counters) that this reduces the number of invalidations. So we can keep track :)

tcl3:
I've added a new style invalidation counter to track the usage of this optimization and added a test that uses it.

PR #9113, opened 2026-04-26
LibWebView: Add search functionality to the bookmark manager

mikiubo:
Add a search bar to the about:bookmarks page that filters bookmarks

### DIFF
--- a/Base/res/ladybird/about-pages/bookmarks.html
+++ b/Base/res/ladybird/about-pages/bookmarks.html
@@ -225,6 +225,58 @@
                 font-size: 12px;
                 min-width: 0;
             }
+
+            .search-container {
+                display: flex;
+                align-items: center;
+                gap: 8px;
+
+                max-width: 65%;
+                margin: 0 auto 24px auto;
+                padding: 0px 12px;
+
+                background-color: var(--tree-hover-color);
+                border: none;
+                border-radius: 8px;
+            }
+
+            .search-icon {
+                flex-shrink: 0;
+                width: 16px;
+                height: 16px;
+                color: var(--empty-text-color);
+            }
+
+            .search-input {
+                flex-grow: 1;
+                min-width: 0;
+                font-size: 13px;
+                color: inherit;
+
+                background: none;
+                border: none;
+                outline: none;
+                padding: 0;
+            }
+
+            .search-input::placeholder {
+                color: var(--empty-text-color);
+            }
+
+            .search-highlight {
+                color: inherit;
+                background-color: var(--tree-selected-color);
+                border-radius: 2px;
+            }
+
+            .search-no-results {
+                color: var(--empty-text-color);
+
+                padding: 40px 20px;
+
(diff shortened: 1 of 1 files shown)


Base/res/ladybird/about-pages/bookmarks.html
```diff
@@ -250,6 +305,21 @@ <h1>Bookmarks</h1>
...
+                xmlns="http://www.w3.org/2000/svg"
+                viewBox="0 0 24 24"
+                fill="none"
+                stroke="currentColor"
+                stroke-width="2"
+            >
+                <circle cx="11" cy="11" r="7" />
+                <line x1="16" y1="16" x2="22" y2="22" />
+            </svg>
+            <input type="text" class="search-input" id="search-input" placeholder="Search bookmarks..." />
```
trflynn89: I don't think this needs an ellipsis at the end, i.e. it should just be "Search bookmarks". The ellipsis is used to indicate that the action isn't immediate & will require additional user interaction.

Base/res/ladybird/about-pages/bookmarks.html
```diff
@@ -330,6 +484,10 @@ <h1>Bookmarks</h1>
                     event.preventDefault();
                     event.stopPropagation();
 
+                    if (SEARCH_QUERY) {
```
trflynn89: Why do we disable context menus when there is a search query? Disabling dragging makes sense, but I think we should still have access to the context menu actions.

trflynn89 (commented):
Sorry for the late review on this, I was on a bit of a vacation.

The bookmark search bar seems quite large? The highlighted text is also a bit hard to read for me in dark mode (we can probably just keep the current text `color` instead of making it dark?)

Image

This is Chrome's search bar for comparison, theirs is less tall and doesn't take the full width - looks less out-of-place to me here:

Image

Base/res/ladybird/about-pages/bookmarks.html
mikiubo: think i fixed

Base/res/ladybird/about-pages/bookmarks.html
mikiubo: Ok fixed.
I have some concern about the "create" new bookmarks that is not clear where will be created but i do some test i seems coherent.

mikiubo:
> Sorry for the late review on this, I was on a bit of a vacation.
> 
> The bookmark search bar seems quite large? The highlighted text is also a bit hard to read for me in dark mode (we can probably just keep the current text `color` instead of making it dark?)
> Image
> 
> This is Chrome's search bar for comparison, theirs is less tall and doesn't take the full width - looks less out-of-place to me here:
> Image

No worries at all 🙂

I tried to make it look a bit closer to Chrome’s search bar — not sure if I fully got there.

Let me know if you have any other visual suggestions..

trflynn89 (approved):

PR #9116, opened 2026-04-26
LibWeb/Bindings: Pass exact typed array types through WebIDL bindings

shannonbooth:
Generate exact JS buffer types for exact IDL buffer arguments instead of widening them to BufferSource or ArrayBufferView.

This fixes cases like TextEncoder.encodeInto(), whose IDL requires
a Uint8Array destination. Previously the generated binding accepted
any BufferSource, so DataView, other typed arrays, and
ArrayBuffer-backed values were let through. With exact conversion,
those are rejected at the binding layer as expected.

Found in my branch trying to generate dictionaries by IDL, where this becomes a bit problematic. Allows tidying up some platform object implementations.


shannonbooth:
Whoops, originally pushed old version, now have the right stuff pushed. Happens to also fix a bug where we were not doing some type checking we were meant to so also happen to score 26 more WPT points.

PR #9122, opened 2026-04-27
LibJS: Drop C++-port parity quirks from parser and scope collector

awesomekling:
The Rust parser and scope collector held onto a handful of shapes that only existed to lockstep against the long-gone C++ pipeline.

Along the way, fixed a runtime crash where `function f() { return { arguments } }` allocated an uninitialized arguments local because the shorthand path bypassed the parser's eval/arguments check. We now fall back to "scope analysis allocated a non-lexical arguments local" as the trigger for materializing the arguments object.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -375,6 +375,7 @@ version = "0.1.0"
 dependencies = [
  "bytecode_def",
  "cbindgen",
+ "indexmap",
  "libunicode_rust",
  "num-bigint",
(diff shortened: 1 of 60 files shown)


jdahlin:
Perhaps this fixes #8805 then?

awesomekling:
> Perhaps this fixes #8805 then?

Nope, that still needs fixing :)

PR #9124, opened 2026-04-27
UI/Gtk: Fix wheel scroll position on Wayland

jdahlin:
GDK4 scroll events do not carry reliable pointer coordinates on Wayland (gdk_event_get_position returns INT_MIN). Track the last known pointer position from the motion controller and use it when building the wheel event, matching the approach used by WebKit GTK.

This fixes mouse-wheel scrolling and scrollbar interaction in pages with overflow containers such as the pdf.js viewer.

### DIFF
--- a/UI/Gtk/Widgets/LadybirdWebView.cpp
+++ b/UI/Gtk/Widgets/LadybirdWebView.cpp
@@ -16,6 +16,8 @@
 struct LadybirdWebView {
     GtkWidget parent_instance;
     Ladybird::WebContentView* impl { nullptr };
+    double last_mouse_x { 0 };
+    double last_mouse_y { 0 };
 };
 
 struct LadybirdWebViewClass {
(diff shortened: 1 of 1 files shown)


UI/Gtk/Widgets/LadybirdWebView.cpp
```diff
@@ -176,10 +180,14 @@ static gboolean on_scroll(GtkEventControllerScroll* controller, gdouble dx, gdou
...
+    // GDK scroll events on Wayland do not carry reliable pointer coordinates, so we
+    // use the last position recorded by the motion controller instead.
+    auto position = Web::DevicePixelPoint { static_cast<int>(self->last_mouse_x * device_pixel_ratio), static_cast<int>(self->last_mouse_y * device_pixel_ratio) };
+
     Web::MouseEvent event {
         .type = Web::MouseEvent::Type::MouseWheel,
-        .position = {},
-        .screen_position = {},
+        .position = position,
+        .screen_position = position,
```
trflynn89, line 191: This won't be the correct value for the screen position - let's at least add a FIXME for that.

jdahlin: Sure, done

PR #9128, opened 2026-04-27
LibWeb: Don't strand a fresh navigation behind a sync history step

awesomekling:
When a click handler calls `history.replaceState` and the link's cross-document activation behavior runs in the same task, the queued sync step runs apply-the-history-step on the navigable mid-navigation, transitioning its ongoing navigation through "traversal" and back to null. That aborts the link's navigate event and bails out its deferred work, leaving the link nav abandoned.
    
Skip this transient when the navigable already has a fresh ongoing navigation. No major engine reproduces the race; long term, sync same-document nav should bypass the traversal queue entirely (matching Chromium).
    
This solves the long-standing issue where clicking on a box of tea on https://twinings.co.uk/ would freeze the browser. :^)


Libraries/LibWeb/HTML/TraversableNavigable.cpp
```diff
@@ -599,6 +599,33 @@ void ApplyHistoryStepState::start()
...
+        //         navigable's ongoing navigation through "traversal" and back to null, which informs the navigation
+        //         API about aborting (cancelling the in-flight navigation's navigate event) and then trips the
+        //         "ongoing navigation is no longer navigationId" bail-out in navigate's deferred steps, leaving the
+        //         in-flight navigation stranded.
+        //
+        //         The race shows up when a click capture handler calls history.replaceState and the link's
+        //         cross-document activation behavior runs in the same task: the replaceState's queued sync step
+        //         finalizes after the link nav has already entered begin_navigation, and wipes its id. Real-world
+        //         example: Shopify storefronts running hCaptcha (which hooks click and replaceStates to the current
+        //         URL for analytics). The page silently never moves.
```
shannonbooth, line 613: Is this the same as https://github.com/whatwg/html/issues/12362? We've gotten quite far behind the spec when it comes to the history steps, I'm wondering if it would be worthwhile trying to do a big ol' catchup for this

awesomekling: If I've understood correctly, that issue is about an ongoing navigate event getting yanked out from under a deferred intercepted traverse, ours is the ongoing navigation enum getting clobbered to traversal then null by a sync replaceState while a cross doc nav is mid-flight. Same family of "deferred work observes stale nav state" race, different specific instance.

Might be good to do a catchup, I don't know. Right now I just want to make Shopify stores work and get some test coverage in :)

PR #9132, opened 2026-04-27
Add a PDF reader using PDF.js

jdahlin:
Adds inline PDF viewing using the bundled pdf.js viewer. 

When Ladybird loads a PDF response, it renders it at an internal `ladybird-pdfjs-viewer://` scheme
(isolated origin, same-origin checks work) while showing the original PDF URL in the address bar, which is similar to how it is implemented in WebKit.

The pdf.js dist is pulled in as a vcpkg overlay port.

Tested to work with a few example pdfs over file://, http:// and https://.

image

### DIFF
--- a/Libraries/LibURL/URL.cpp
+++ b/Libraries/LibURL/URL.cpp
@@ -364,9 +364,13 @@ Origin URL::origin() const
         return Origin(scheme(), host().value(), port());
     }
 
+    // AD-HOC: resource:// URLs are internal browser resources; give them a shared tuple origin
+    // so that same-origin checks pass between any two resource:// documents or worker scripts.
+    if (scheme() == "resource"sv)
+        return Origin(scheme(), String {}, {});
+
     // -> "file"
-    // AD-HOC: Our resource:// is basically an alias to file://
-    if (scheme() == "file"sv || scheme() == "resource"sv) {
+    if (scheme() == "file"sv) {
         // Unfortunate as it is, this is left as an exercise to the reader. When in doubt, return a new opaque origin.
 
         // Our implementation-defined behavior is to return an opaque origin for file:// URLs,
--- a/Libraries/LibWeb/DOM/DocumentLoading.cpp
+++ b/Libraries/LibWeb/DOM/DocumentLoading.cpp
@@ -8,18 +8,26 @@
 
 #include <AK/Debug.h>
 #include <AK/LexicalPath.h>
+#include <AK/Utf16FlyString.h>
 #include <LibCore/Promise.h>
+#include <LibCore/Resource.h>
 #include <LibGfx/ImageFormats/ImageDecoder.h>
+#include <LibJS/Runtime/NativeFunction.h>
 #include <LibTextCodec/Decoder.h>
+#include <LibURL/URL.h>
+#include <LibWeb/DOM/CustomEvent.h>
 #include <LibWeb/DOM/Document.h>
 #include <LibWeb/DOM/DocumentLoading.h>
+#include <LibWeb/DOM/IDLEventListener.h>
 #include <LibWeb/Fetch/Infrastructure/HTTP/MIME.h>
+#include <LibWeb/Fetch/Response.h>
 #include <LibWeb/HTML/HTMLHeadElement.h>
 #include <LibWeb/HTML/Navigable.h>
 #include <LibWeb/HTML/NavigationParams.h>
 #include <LibWeb/HTML/Parser/HTMLEncodingDetection.h>
 #include <LibWeb/HTML/Parser/HTMLParser.h>
 #include <LibWeb/HTML/Parser/IncrementalDocumentParser.h>
+#include <LibWeb/HTML/Window.h>
 #include <LibWeb/Loader/GeneratedPagesLoader.h>
 #include <LibWeb/MimeSniff/Resource.h>
 #include <LibWeb/Namespace.h>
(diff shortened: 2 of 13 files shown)


Meta/CMake/vcpkg/overlay-ports/pdfjs/portfile.cmake
```diff
@@ -0,0 +1,49 @@
...
+file(INSTALL "${SOURCE_PATH}/web/images/"
+     DESTINATION "${CURRENT_PACKAGES_DIR}/share/pdfjs/web/images")
+
+# web/locale/: en-US only with a minimal locale.json
+file(INSTALL "${SOURCE_PATH}/web/locale/en-US/"
+     DESTINATION "${CURRENT_PACKAGES_DIR}/share/pdfjs/web/locale/en-US")
+file(WRITE "${CURRENT_PACKAGES_DIR}/share/pdfjs/web/locale/locale.json"
+    "{\"en-us\":\"en-US/viewer.ftl\"}")
+
+set(VCPKG_POLICY_EMPTY_INCLUDE_FOLDER enabled)
```
cqundefine: Are you sure this one is necessary, as far as I can tell `VCPKG_POLICY_EMPTY_PACKAGE` disables all checks already.

jdahlin: Probably, not. I'll remove it!

Libraries/LibWeb/Loader/ResourceLoader.cpp
```diff
@@ -375,7 +381,7 @@ RefPtr<Requests::Request> ResourceLoader::load(LoadRequest& request, GC::Root<On
         return nullptr;
     }
 
-    if (url.scheme() == "resource"sv) {
+    if (url.scheme() == "resource"sv || url.scheme() == "ladybird-pdfjs-viewer"sv) {
```
shannonbooth: I can't remember very well what I did last time when I prototyped this and didn't test very much, but I don't remember needing to add all this type of special cases for a special scheme. I'll probably need to check this out locally but why do we need the special cases?

Libraries/LibWeb/DOM/DocumentLoading.cpp
```diff
@@ -415,6 +442,44 @@ static WebIDL::ExceptionOr<GC::Ref<DOM::Document>> load_media_document(HTML::Nav
     // be true for the Document.
 }
 
+// Renders the PDF using the bundled pdf.js viewer at an internal viewer URL.
+static WebIDL::ExceptionOr<GC::Ref<DOM::Document>> load_pdf_document(HTML::NavigationParams const& navigation_params)
+{
+    auto maybe_pdf_url = navigation_params.response->url();
+    if (!maybe_pdf_url.has_value())
+        return WebIDL::SimpleException { WebIDL::SimpleExceptionType::TypeError, "PDF response has no URL"_string };
```
shannonbooth: I _think_ all errors in this function being propagated would be a bug, so can be changed to VERIFY's

Libraries/LibWeb/Loader/ResourceLoader.cpp
jdahlin: I copied it from webkit, it's mainly for:
- conceptual clarity
- to support fetch(), required to avoid changing pdf.js too much.
- origin isolation

It could be changed to resource:// but then we'd have to add support for fetch() there.

shannonbooth: > copied it from webkit

_Copied_ from WebKit?

Also I remember what I did now - I used a resource:// URL to to access pdf.js and exposed an Internals object with IDL of something like

```idl
[Exposed=Nobody]
interface PDFViewer {
    ArrayBuffer getData();
    readonly USVString url;
};
```
Using getData to pass across the document. Not sure what is better

jdahlin: I copied the [concept of a separate scheme](https://github.com/search?q=repo%3AWebKit%2FWebKit%20webkit-pdfjs-viewer&type=code) just for viewing a pdf.

I like "ladybird-pdfjs-viewer" slightly more than piggybacking on resource:// for this, since there are different permissions, but could be convinced to change it.

`PDFViwer.getData()` is indeed nicer than `__ladybirdPendingPdfData`

shannonbooth: I think resource:// URLs are used in the way I prototyped in the same way which are used here. Here was my attempt for reference: https://github.com/shannonbooth/ladybird/blob/fd78fef68bf7dd4783b3c3c03879319608ec9585/Libraries/LibWeb/DOM/DocumentLoading.cpp#L438

NB: Still a bit messy

jdahlin: > I think resource:// URLs are used in the way I prototyped in the same way which are used here. Here was my attempt for reference: https://github.com/shannonbooth/ladybird/blob/fd78fef68bf7dd4783b3c3c03879319608ec9585/Libraries/LibWeb/DOM/DocumentLoading.cpp#L438
> 
> NB: Still a bit messy

Its a bit simpler, but I think the main concern with that approach is that pdf.js runs in the same origin as the content rather than in an isolated internal origin.

shannonbooth: I must admit, I never looked into that deeply at all since isn't the origin of the pdf the origin of the content? I think I might be missing what making it isolated protects against, though does seem harmless to change anyhow.
 
Regardless, if its important we can change that origin irrespective, which I this PR already does:

```cpp
    auto document = MUST(DOM::Document::create_and_initialize(DOM::Document::Type::HTML, "text/html"_string, navigation_params));
    document->set_url(viewer_url);
    document->set_origin(viewer_url.origin());
```

I suppose we should just set the origin to URL::Origin::create_opaque() or even idk isolate in a sandboxed iframe or something. But I'm not really sure what we'd be protecting against and I actually haven't looked into that side of things at all

jdahlin: okay, I was able to simplify this!

shannonbooth: I have looked into this some more, and I'm now thinking that this is not the right approach particularly with the fetch piece. With this, we would now be giving PDF documents more permission than the hosting URL by changing the origin to a resource origin, and also allowing resource URLs to fetch arbitrary files.

I'm not sure of the details of the setup, but PDF.js allows running a subset of JS (such as window.alert), which is enabled here (Firefox supports this too). If for whatever reason that sandbox is escaped and for example fetch is called, that PDF can then fetch arbitrary files on the filesystem.

I didn't get things as functional with https://github.com/shannonbooth/ladybird/blob/fd78fef68bf7dd4783b3c3c03879319608ec9585, but the two advantages I see with something along those lines are:

- No need to change origin logic, only expose the PDF data on an internals object
- URL passed through to the document is preserved, so LibWebView special case is not required (less important, but would be nice if it were possible)

There are probably other ways of solving though

jdahlin: There are different somewhat conflicting requirements here:

- Load all resources from the PDF.js bundle correctly, including js/css/svg
- Support loading of PDFs over file://, http:// and https://
- Avoid reading the whole thing at once by reading chunks
- Reduce compute intensive work on the main thread by using workers
- Be as secure as possible
- Reduce modifications to the Ladybird sources as much as possible
- Reduce modifications to the PDF.js sources as much as possible
- Display the original PDF in the url bar as opposed to the internal resource:// for the viewer
- Reloading the PDF url should work as expected.

Without a proper origin, PDF.js cannot load the worker script, and falls back to running all PDF decoding on the main thread blocking the UI until rendering completes. Fetching files from resource:// can be locked down, so that an exploited fetch() will only allow an attacker to read what is actually bundled there, not arbitrary file system content.

I have simplified the patch even more based on this, but its not possible to get rid of the origin modifications and allow workers at the same time, but I might be missing something.

shannonbooth: Okay great - no this latest version does address the concerns I had, we no longer have that special case in fetch for resource:// URLs which gives a lot more power to them than they should have, and also don't have the LibWebView special cases any more.

Libraries/LibURL/URL.cpp
```diff
@@ -364,9 +364,13 @@ Origin URL::origin() const
         return Origin(scheme(), host().value(), port());
     }
 
+    // AD-HOC: resource:// URLs are internal browser resources; give them a tuple origin so that
+    // same-origin checks work between resource:// documents and their worker scripts.
+    if (scheme() == "resource"sv)
+        return Origin(scheme(), host().value_or(String {}), port());
```
shannonbooth: Is including the host necessary here? It seems a little weird that we would make something in icons be a different origin to ladybird folder / quite arbitrary especially considering this is effectively a file://

(also as a nit we should probably not mention PDFWorker._isSameOrigin in the URL test, that seems a bit overly specific to this specific MR, even more referring to earlier revision)

jdahlin: Good catch, apparently not! I set it to empty on both sites (URL.cpp and DocumentLoading.cpp)

I have also updated the comment.

shannonbooth (approved):

jdahlin:
@shannonbooth thanks! the only CI failure is an unrelated timeout

PR #9140, opened 2026-04-28
LibUnicode+LibWeb: Add and use an ASCII fast path for line break segmentation

tcl3:
This PR adds a `Segmenter` implementation that implements the UAX#14 line breaking rules applicable to ASCII text. This avoids the need to build an ICU `BreakIterator` for the majority of text on the web.

When  taking a 10 second profile loading https://amazon.co.uk, the time spent in `libicuc` drops from 5.02% to 1.17% showing the fast path is hit most of the time. Time spent in the new `AsciiLineSegmenter` is 0.43% showing it is significantly faster than ICU. The overall speedup on this page is ~4% over the first 10 seconds.

Tested by throwing lots of text heavy pages and edge cases at [this patch](<https://gist.github.com/tcl3/8399353c0140100a745d438f8699b0df>), which compares the fast path output to the slower ICU path and crashes on any divergence.

### DIFF
--- a/Libraries/LibUnicode/Segmenter.cpp
+++ b/Libraries/LibUnicode/Segmenter.cpp
@@ -4,6 +4,9 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
+#include <AK/CharacterTypes.h>
+#include <AK/GenericShorthands.h>
+#include <AK/OwnPtr.h>
 #include <AK/Utf16View.h>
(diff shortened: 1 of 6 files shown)


gmta:
For https://pomax.github.io/bezierinfo/, this PR shaves off about 500ms of the runtime on my machine (from 3.92s to 3.53s, -10%) 🥳

gmta (approved):

Issue #8582, opened 2026-03-22
LibJS: Improve parser performance


jdahlin:
Let's improve the parser (and lexer) in libjs!

Compared to V8, JSC and SpiderMonkey we have a significant gap in both parsing speed and memory usage. 

Benchmark file: 8 MB JS bundle from instagram.com (36K functions). All numbers on Apple MacBook Pro M1.
Yes this is a bit silly, but the this will help normal websites, especially the ones that are not hyper-optimized and contain significant amount of unused javascript code.

The north star is JSC, which can parse the same file in 10ms using under 40 MB peak RSS.

Before this work started, peak memory was ~1 GB.

- [x] ~~FunctionTable: sparse Vec → HashMap~~ #8418
- [x] Box large AST enum variants #8581
- [ ] String interning for identifiers
- [ ] Lazy parsing with lightweight syntax-check mode
- [ ] Arena allocator for AST nodes
- [ ] Investigate faster hashing for compiler internals
- [ ] Background parsing of deferred functions
- [ ] Detect and eagerly compile IIFEs to avoid wasted preparse
- [ ] Reduce Rust→C++ FFI overhead

jdahlin:
Mimalloc #8752 helps a bit here too.

ju-klein:
~~Hi~~ Hejsan @jdahlin! I would like to help with this.
Would you be open to me working on the String Interning? Would be my first contribution to Ladybird, so I might take a little longer than you would and I might ask questions and/or need a little help/guidance. But I know in principle what String Interning is supposed to do: Prevent us from having to use string comparisons (O(n)) and get a faster lookup time.
(I took Compiler Construction, so I am not completely oblivious!)

caminante-blanco:
Hello @jdahlin! I was hoping to work on this as well. I've been doing some work on my own to implement Lazy parsing, and eventually some heuristics for detecting IIFE's. This  would be my first time contributing, so I'll probably take some time on my own before bringing some questions (or a PR) to make sure I really understand the codebase.

jdahlin:
[...]

PR #9147, opened 2026-04-28
LibJS: Introduce string interning for identifiers in the Rust parser

anonrig:
Add a FxHash-backed StringInterner that deduplicates all identifier strings during parsing. Identifier AST nodes now store a 4-byte InternedId handle instead of an Rc<Vec<u16>>, reducing per-node overhead and eliminating redundant allocations for repeated names.

Key changes:

- New StringInterner (adapted from [1]) using the stable-heap-pointer trick: Box<[u16]> entries own the data, HashMap keys borrow from them with zero extra allocation per unique string.

- Identifier.name and PrivateIdentifier.name changed from SharedUtf16String/Utf16String to InternedId. The interner is owned by the Parser, transferred to the Generator for codegen, and threaded through scope analysis and FFI materialization.

- All internal HashMaps (parser, scope collector, bytecode generator, AST FunctionTable) switched from SipHash to FxHash for faster lookups on trusted compiler inputs.

- Lexer simplified: removed the two fixed-size identifier caches (short_identifier_cache[128], recent_identifier_cache[128]) which had high collision rates. The parser-level interner provides unlimited collision-free deduplication.

Performance impact for large JS bundles (~8 MB, ~36K functions):
- Identical identifiers share a single allocation (was: one Vec<u16>
  per occurrence)
- Scope analysis compares u32 IDs instead of string slices
- Bytecode generator intern tables use u32 keys (no string hashing on the hot path)

Part of #8582.

[1] https://github.com/anonrig/string_interner

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -75,6 +75,12 @@ name = "bytecode_def"
 version = "0.1.0"
 
 [[package]]
+name = "byteorder"
+version = "1.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
+
+[[package]]
 name = "calendrical_calculations"
 version = "0.2.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -189,6 +195,15 @@ name = "foldhash"
 version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d9c4f5dac5e15c24eb999c26181a6ca40b39fe946cbe4c263c7209467bc83af2"
+
+[[package]]
+name = "fxhash"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c31b6d751ae2c7f11320402d34e41349dd1016f8d5d45e48c4312bc8625af50c"
+dependencies = [
+ "byteorder",
+]
 
 [[package]]
 name = "getrandom"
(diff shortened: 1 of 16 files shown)


jdahlin:
@anonrig could you please include numbers (before, after) from the https://github.com/LadybirdBrowser/js-benchmarks ? 

I’m a bit wary of adding a dependency on a library with very limited adoption. What does this provide on top of raw FxHash, beyond what () already gives us? Is it an API convenience wrapper, different probing strategy, allocator behavior, or measurable wins in our workloads?

jdahlin:
> * All internal HashMaps (parser, scope collector, bytecode generator, AST FunctionTable) switched from SipHash to FxHash for faster lookups on trusted compiler inputs.

In a browser JS engine, source text is attacker-controlled, so parser/compiler hash maps keyed by source-derived strings are not "trusted compiler input".

R-Goc:
No matter what, a dependency that was last updated 8 years ago is not acceptable imo. See: https://crates.io/crates/fxhash
Also I recall fxhash was previously discussed over siphash as provided by the rust stdlib and was found insufficient, though I'm not an expert in hashes so I don't have a strong opinion on the algo choice itself. See: https://github.com/LadybirdBrowser/ladybird/pull/8408#issuecomment-4060507873

nicoburns:
The `rustc_hash` crate is maintained fork for the `fxhash` crate that is maintained by the Rust compiler team (and used in rustc itself). It's recommended for HashMaps with small keys (64bit or smaller) and trusted input.

[foldhash](https://github.com/orlp/foldhash) is recommended for larger keys and trusted input. SipHash has protection against HashDOS hash collisions attacks.

You may also want to look at the https://docs.rs/string_cache crate which is what Servo uses for string interning. It's performance is currently only ok, due to the `phf` crate also using SipHash, but will hopefully improve once https://github.com/rust-phf/rust-phf/pull/392 lands in a release. The nice thing about the `string_cache` crate is that it allows you to create a list of static atoms which are compiled into the binary and can then be constructed without allocating (super useful for things like keywords).

PR #9154, opened 2026-04-28
UI/Gtk: Support keypad buttons for zoom actions

rgret-dev:
The Qt UI supports this so GTK should have it, too :^)


jdahlin:
Looks good to me!

PR #9158, opened 2026-04-29
LibCompress: Treat LZW decoding errors as end of stream

gmta:
The LZW data for both GIF and TIFF images is sometimes intentionally missing an end-of-information (EOI) code, which technically is a decoding error, but in practive is handled gracefully by Firefox, Safari and Chrome for GIFs and Safari for TIFFs. Let's mirror their behavior.

The included WPT test exposes the fact that trailing garbage bytes can also result in decoding errors. We handle this in the LZW logic rather than in the image decoding since our LZW implementation is currently only used by GIF and TIFF decoding. The error is logged behind the LZW_DEBUG flag.

Also prevents this commonly seen debug log entry:

```
WebContent(..): ImageDecoderClient: Invalid bitmap for request X at index Y
```

CC @awesomekling this also updates one of your more recent layout tests that contains such a 1x1 GIF pixel.


gmta:
Marking as draft - the fix here should be a bit broader, other browsers seem to treat any LZW decoding error as an implicit but valid end of stream.

PR #9161, opened 2026-04-29
UI/AppKit: Append .com on Ctrl+Return in the location bar

christianfrey:
Holding Ctrl while pressing Return in the location bar appends the configured TLD (.com) to the entered text before navigating.

### DIFF
--- a/UI/AppKit/Interface/TabController.mm
+++ b/UI/AppKit/Interface/TabController.mm
@@ -290,6 +290,26 @@ static NSImage* location_field_globe_icon()
     }
 }
 
+// AppKit does not send Ctrl+Return through the field editor, so intercept it here and forward as a regular newline.
+// The delegate then checks the current event to handle Ctrl (append TLD).
+- (BOOL)performKeyEquivalent:(NSEvent*)event
+{
+    if ([self.window firstResponder] == [self currentEditor]
+        && ([event modifierFlags] & NSEventModifierFlagControl) != 0) {
+        NSString* chars = [event charactersIgnoringModifiers];
+        if ([chars isEqualToString:@"\r"] || [chars isEqualToString:@"\n"]) {
+            id delegate = [self delegate];
+            if ([delegate respondsToSelector:@selector(control:textView:doCommandBySelector:)]) {
+                [delegate control:self
+                               textView:(NSTextView*)[self currentEditor]
+                    doCommandBySelector:@selector(insertNewline:)];
+                return YES;
+            }
+        }
+    }
+    return [super performKeyEquivalent:event];
+}
+
 // NSSearchField does not provide an intrinsic width, which causes an ambiguous layout warning when the toolbar auto-
 // measures this view. This provides an initial fallback, which is overridden with an explicit width in windowDidResize.
 - (NSSize)intrinsicContentSize
@@ -783,11 +803,11 @@ static NSImage* location_field_globe_icon()
     m_is_applying_inline_autocomplete = false;
 }
 
-- (BOOL)navigateToLocation:(String)location
+- (BOOL)navigateToLocation:(String)location appendTLD:(WebView::AppendTLD)append_tld
 {
     m_autocomplete->cancel_pending_query();
 
-    if (auto url = WebView::sanitize_url(location, WebView::Application::settings().search_engine()); url.has_value()) {
+    if (auto url = WebView::sanitize_url(location, WebView::Application::settings().search_engine(), append_tld); url.has_value()) {
         [self loadURL:*url];
     } else {
         [[[self tab] web_view] view].load_navigation_error_page(location);
@@ -1194,15 +1214,24 @@ static NSImage* location_field_globe_icon()
             return YES;
     }
 
-    if (selector != @selector(insertNewline:)) {
+    if (selector != @selector(insertNewline:))
         return NO;
+
+    auto append_tld = WebView::AppendTLD::No;
+    auto* event = [[self tab] currentEvent];
+    if (event && event.type == NSEventTypeKeyDown
+        && (event.modifierFlags & NSEventModifierFlagControl) != 0) {
+        append_tld = WebView::AppendTLD::Yes;
     }
 
-    auto location = [self.autocomplete selectedSuggestion].value_or_lazy_evaluated([&]() {
-        return Ladybird::ns_string_to_string([[text_view textStorage] string]);
-    });
+    // Ctrl+Return uses the raw location field query instead of the selected autocomplete suggestion.
+    auto location = (append_tld == WebView::AppendTLD::Yes)
+        ? Ladybird::ns_string_to_string([self currentLocationFieldQuery])
+        : [self.autocomplete selectedSuggestion].value_or_lazy_evaluated([&]() {
+              return Ladybird::ns_string_to_string([[text_view textStorage] string]);
+          });
 
-    [self navigateToLocation:move(location)];
+    [self navigateToLocation:move(location) appendTLD:append_tld];
     return YES;
 }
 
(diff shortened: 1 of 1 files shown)


UI/AppKit/Interface/TabController.mm
```diff
@@ -173,6 +173,26 @@ - (BOOL)becomeFirstResponder
...
+{
+    if ([self.window firstResponder] == [self currentEditor]
+        && ([event modifierFlags] & NSEventModifierFlagControl) != 0) {
+        NSString* chars = [event charactersIgnoringModifiers];
+        if ([chars isEqualToString:@"\r"] || [chars isEqualToString:@"\n"]) {
+            id delegate = [self delegate];
+            if ([delegate respondsToSelector:@selector(control:textView:doCommandBySelector:)]) {
+                [delegate control:self
+                               textView:(NSTextView*)[self currentEditor]
+                    doCommandBySelector:@selector(insertNewline:)];
```
trflynn89, line 305: I think we can remove this and do this work inside `doCommandBySelector`. Something like:
```objc++
auto append_tld = WebView::AppendTLD::No;

if (selector == @selector(noop:)) {
    auto* event = [[self tab] currentEvent];

    if (!event || event.type != NSEventTypeKeyDown)
        return NO;
    if ((event.modifierFlags & NSEventModifierFlagControl) == 0)
        return NO;
    if (event.keyCode != kVK_Return && event.keyCode != kVK_ANSI_KeypadEnter)
        return NO;

    append_tld = WebView::AppendTLD::Yes;
} else if (selector != @selector(insertNewline:)) {
    return NO;
}
```

UI/AppKit/Interface/TabController.mm
```diff
@@ -995,11 +1015,16 @@ - (BOOL)control:(NSControl*)control
...
     }
 
-    auto location = [self.autocomplete selectedSuggestion].value_or_lazy_evaluated([&]() {
-        return Ladybird::ns_string_to_string([[text_view textStorage] string]);
-    });
+    auto ctrl_held = ([[NSApp currentEvent] modifierFlags] & NSEventModifierFlagControl) != 0;
+    auto append_tld = ctrl_held ? WebView::AppendTLD::Yes : WebView::AppendTLD::No;
+
+    auto location = ctrl_held
+        ? Ladybird::ns_string_to_string([self currentLocationFieldQuery])
```
trflynn89, line 1229: I don't think we want this branch? We don't do this in Qt.

If `[self.autocomplete selectedSuggestion]` contains a URL already, we want to use it - e.g. if it's a .org address, we don't want to replace the TLD with .com. Appending a TLD is something that should only happen if the URL doesn't already contain one.

trflynn89 (commented):
On macOS, I'm not sure whether this should be ctrl+enter or cmd+enter. Seems Chrome uses the former and Firefox uses the latter. cmd+enter does seem a bit more natural to me.

nico:
> On macOS, I'm not sure whether this should be ctrl+enter or cmd+enter. Seems Chrome uses the former and Firefox uses the latter. cmd+enter does seem a bit more natural to me.

In Chrome/Mac, cmd-mouseclick opens a link in a new background tab, cmd-shift-click in a new foreground tab, and shift-click in a new window. In the location bar, cmd-enter opens the url in a new background tab, cmd-shift-enter in a new foreground, and shift-enter in a new window. (Also works for clicking the back button, bookmarks, …) Up to you if you think that's a good idea.

(Cmd-enter and cmd-shift-enter behavior matches Safari. In Safari, shift-clicking a link adds it to the reading list apparently. I don't know if shift-enter does anything over there.)

UI/AppKit/Interface/TabController.mm
Psychpsyo: Wouldn't Safari behavior be the thing to emulate here?
(if it has this, that is)

christianfrey: Done! Switched to Cmd+Return so the event reaches doCommandBySelector: as noop:.

Also removed the performKeyEquivalent: override, which was only needed for Ctrl key handling, and applied your suggested snippet.

UI/AppKit/Interface/TabController.mm
christianfrey: The goal here is to avoid Cmd+Return committing the auto-selected suggestion instead of the literal input. With this change, Cmd+Return explicitly uses what the user typed, while Enter keeps the Qt-style suggestion behavior.

Without this, typing "googl" + Cmd+Return would resolve to "google.com", which is not consistent with the intended shortcut semantics.

christianfrey:
Switched to Cmd+Return since it's the more natural shortcut on macOS and matches Safari behavior.

christianfrey:
CI failed due to flakiness, could you re-run please?

nico:
> Switched to Cmd+Return since it's the more natural shortcut on macOS and matches Safari behavior.

Over here, cmd-return does not add .com in safari as far as I can tell (see above).

christianfrey:
> > Switched to Cmd+Return since it's the more natural shortcut on macOS and matches Safari behavior.
> 
> Over here, cmd-return does not add .com in safari as far as I can tell (see above).

Yeah, I mixed it up. I originally opened the PR with Ctrl+Enter to match Chrome behavior, but Cmd seems like the more appropriate modifier on macOS as some people pointed out.

Ultimately, it probably comes down to what feels most natural to the majority of macOS users.

UI/AppKit/Interface/TabController.mm
```diff
@@ -11,6 +11,7 @@
 #include <LibWebView/ViewImplementation.h>
 
 #import <Application/ApplicationDelegate.h>
+#import <Carbon/Carbon.h>
```
trflynn89: What's this added for?

trflynn89 (commented):
Whoops, sorry, I forgot to get back this this after a vacation.

Given that it seems like Chrome put more thought into these features than other browsers, where they have cmd mimic clicking a link, I think it does make more sense to go back to using ctrl here. That will let us implement the same features in the future if we so choose. Sorry for the back-and-forth here.

UI/AppKit/Interface/TabController.mm
christianfrey: It was used for kVK_Return / kVK_ANSI_KeypadEnter, but we no longer use those keycodes, so the Carbon import has been removed.

christianfrey:
> Whoops, sorry, I forgot to get back this this after a vacation.
> 
> Given that it seems like Chrome put more thought into these features than other browsers, where they have cmd mimic clicking a link, I think it does make more sense to go back to using ctrl here. That will let us implement the same features in the future if we so choose. Sorry for the back-and-forth here.

No worries at all, thanks for the clarification 👍

I've updated the implementation to stick with Ctrl+Enter as requested.

UI/AppKit/Interface/TabController.mm
```diff
@@ -173,6 +173,26 @@ - (BOOL)becomeFirstResponder
     return result;
 }
 
+// AppKit does not send Ctrl+Return through the field editor, so intercept it here and forward as a regular newline.
+// The delegate then checks the current event to handle Ctrl (append TLD).
+- (BOOL)performKeyEquivalent:(NSEvent*)event
```
trflynn89, line 295: Wait why are we back to the `performKeyEquivalent` solution? We should still be doing all of the work in `doCommandBySelector`, just using ctrl instead of cmd. i.e. this:

https://github.com/LadybirdBrowser/ladybird/pull/9161#discussion_r3160711028

christianfrey: Just to clarify: Ctrl+Return is intercepted by AppKit before it reaches the field editor, so it never hits `doCommandBySelector` (not even as `noop:`).
That's why it has to be handled in `performKeyEquivalent` and forwarded as `insertNewline` so the existing logic can run.
Happy to adjust if there’s a cleaner approach.

gmta:

PR #9170, opened 2026-04-29
LibWeb+Meta: Fix errors while building on Python 3.9

Zaggy1024:
Builds were broken on macOS with system Python.

### DIFF
--- a/Meta/Utils/CSSGrammar/generator.py
+++ b/Meta/Utils/CSSGrammar/generator.py
@@ -25,7 +25,7 @@ def generate_css_parser_expression_for_type_component_value(out: TextIO, cpp_nam
     type_name = snake_casify(type.name)
 
     additional_arguments = ""
-    if type.custom_ident_blacklist is not None:
+    if type.custom_ident_blacklist:
         additional_arguments = ", ReadonlySpan<StringView> { "
 
         if len(type.custom_ident_blacklist) > 0:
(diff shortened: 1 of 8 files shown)


Meta/Utils/CSSGrammar/generator.py
```diff
@@ -56,15 +56,11 @@ def generate_css_parser_expression_for_keyword_component_value(out: TextIO, cpp_
...
 ) -> None:
-    match grammar_node.component_value:
-        case Type() as type_component_value:
-            generate_css_parser_expression_for_type_component_value(out, cpp_name, type_component_value)
-            return
-        case Keyword() as keyword:
-            generate_css_parser_expression_for_keyword_component_value(out, cpp_name, keyword)
-            return
-
-    raise TypeError(f"Unhandled component value type: {type(grammar_node.component_value).__name__}")
```
Calme1709, line 67: Can we add this error back in?

Calme1709 (commented):
Can we enforce this with Ruff using [target version](https://docs.astral.sh/ruff/settings/#target-version)?

Meta/Utils/CSSGrammar/generator.py
Zaggy1024: Pylance complains about an unnecessary `isinstance` there with a second condition on Keyword. Perhaps we could ignore the error, but it should also cause an error when passing a non-Keyword to `generate_css_parser_expression_for_keyword_component_value` anyway.

Zaggy1024:
> Can we enforce this with Ruff using [target version](https://docs.astral.sh/ruff/settings/#target-version)?

Would that actually enforce feature usage? If so, seems reasonable to me. I suppose we should probably have that either way, though, so I'll add it on the end here.

Zaggy1024:
...I guess it does. Time to fix up some other usages.

gmta (approved):

PR #9173, opened 2026-04-30
CI: Cache apt packages and skip redundant apt-get update

jdahlin:
This is an attempt to speed up the apt based runners to avoid having to redownload all the packages on each build.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -54,13 +54,23 @@ runs:
           pip3 install ruff
         fi
 
+    - name: 'Cache apt packages'
+      if: ${{ inputs.os == 'Linux' && inputs.type == 'build' }}
+      uses: actions/cache@v4
+      with:
+        path: /var/cache/apt/archives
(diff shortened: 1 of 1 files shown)


.github/actions/setup/action.yml
```diff
@@ -54,24 +54,33 @@ runs:
           pip3 install ruff
         fi
 
+    - name: 'Cache apt packages'
+      if: ${{ inputs.os == 'Linux' && inputs.type == 'build' }}
+      uses: actions/cache@v4
+      with:
+        path: /var/cache/apt/archives
+        key: apt | ${{ inputs.arch }} | ${{ inputs.toolchain }} | ${{ inputs.llvm_version }} | ${{ inputs.gcc_version }} | ${{ hashFiles('.github/actions/setup/action.yml') }}
+        restore-keys: apt | ${{ inputs.arch }} | ${{ inputs.toolchain }} | ${{ inputs.llvm_version }} | ${{ inputs.gcc_version }}
```
gmta, line 63: With our other caches as large as they are, this will probably not help a lot in the long run since they're evicted quite often.

The approach I think we should take instead, is to create a new container image with all the build dependencies preinstalled and use that for the action jobs instead. Would provide a nice speedup and increase reproducibility as well.

PR #9174, opened 2026-04-30
Update copyright year in LICENSE file

jdahlin:


gmta:
We also have #8824 and #8999 for this.

PR #9176, opened 2026-04-30
LibWeb: Route presentational hints through the CSS cascade

tcl3:
Previously, presentational hints bypassed the regular cascade pipeline and wrote directly into `CascadedProperties` under `CascadeOrigin::Author.` That meant `var()` substitution and the invalid-at-computed-value-time fallback had to be duplicated in a separate per-element pass, which in practice missed the IACVT step and could leave a   `GuaranteedInvalidStyleValue` in the cascaded properties. This caused a crash in downstream code that assumed the value had been resolved.

This change introduces an `AuthorPresentationalHint` cascade origin and feeds presentational hints through the cascade as normal declarations. This means that `var()` resolution now happens in only one place.

Fixes a crash when loading https://html.spec.whatwg.org that happened because a presentational hint was set to an unresolvable `var()` value.


Libraries/LibWeb/CSS/StyleComputer.cpp
gmta, line 1: Commit message nits: "`CascadedProperties `" -> "`CascadedProperties`", `be  duplicated` -> `be duplicated`

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -1321,18 +1321,19 @@ GC::Ref<CascadedProperties> StyleComputer::compute_cascaded_values(DOM::Abstract
...
+            collect_dimension_attribute(presentational_hint_properties, dimension_source, HTML::AttributeNames::height, CSS::PropertyID::Height);
+        }
+        if (!presentational_hint_properties.is_empty()) {
+            apply_property_list_to_cascade(*cascaded_properties, abstract_element, presentational_hint_properties,
+                CascadeOrigin::AuthorPresentationalHint, Important::No, {}, nullptr, nullptr);
         }
-
-        // SVG presentation attributes are parsed as CSS values, so we need to handle potential custom properties here.
-        if (element.is_svg_element())
-            cascaded_properties->resolve_unresolved_properties(abstract_element);
```
AtkinsSJ, line 1327: Thanks, this has bothered me for so long!

AtkinsSJ (approved):
Looks good to me. Sorry about the merge conflict. 😅

tcl3:
Rebased to fix conflicts

AtkinsSJ (approved):

PR #9184, opened 2026-04-30
LibJS: Avoid ropes for short flat string concatenations

kalenikaliaksandr:
Short string concatenations are a common allocation churn pattern in JS execution. Many of these results are immediately observed as flat strings, so the intended win is to avoid spending GC and flattening work on an intermediate representation that does not carry its weight.

Microbenchmark:

    const n = 10_000_000;
    function bench(a, b) {
        let total = 0;
        for (let i = 0; i < n; ++i)
            total += (a + b).length;
        if (total !== n * 2)
            throw new Error(String(total));
    }
    bench("a", "b");

Measured with hyperfine against the same build with the fast path disabled:

    baseline:  822.7 ms +/- 25.7 ms
    optimized: 385.5 ms +/- 21.5 ms
    speedup:   2.13 +/- 0.14 times faster

### DIFF
--- a/Libraries/LibJS/Runtime/PrimitiveString.cpp
+++ b/Libraries/LibJS/Runtime/PrimitiveString.cpp
@@ -5,6 +5,7 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
+#include <AK/Array.h>
 #include <AK/CharacterTypes.h>
 #include <AK/FlyString.h>
 #include <AK/StringBuilder.h>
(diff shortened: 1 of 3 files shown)


gmta:
Can we add that to the microbenchmarks in js-benchmarks?

kalenikaliaksandr:
> Can we add that to the microbenchmarks in js-benchmarks?

sure, https://github.com/LadybirdBrowser/js-benchmarks/pull/21

PR #9185, opened 2026-04-30
LibWeb: A few simple small performance gains

tcl3:
These optimizations all give small improvements in various profiles, but the wins aren't big enough for them to warrant their own PR.

### DIFF
--- a/Libraries/LibWeb/DOM/HTMLCollection.cpp
+++ b/Libraries/LibWeb/DOM/HTMLCollection.cpp
@@ -25,6 +25,7 @@ GC::Ref<HTMLCollection> HTMLCollection::create(ParentNode& root, Scope scope, Fu
 
 HTMLCollection::HTMLCollection(ParentNode& root, Scope scope, Function<bool(Element const&)> filter, Function<bool(Element const&, Element const&)> sort)
     : PlatformObject(root.realm())
+    , GC::WeakContainer(heap())
     , m_root(root)
     , m_filter(move(filter))
     , m_sort(move(sort))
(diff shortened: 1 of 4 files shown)


Libraries/LibWeb/DOM/HTMLCollection.cpp
```diff
@@ -49,14 +49,17 @@ void HTMLCollection::visit_edges(Cell::Visitor& visitor)
...
+    if (m_cached_name_to_element_mappings)
+        visitor.visit(*m_cached_name_to_element_mappings);
 }
 
 void HTMLCollection::update_name_to_element_mappings_if_needed() const
 {
     update_cache_if_needed();
     if (m_cached_name_to_element_mappings)
         return;
-    m_cached_name_to_element_mappings = make<OrderedHashMap<FlyString, GC::Weak<Element>>>();
```
kalenikaliaksandr, line 59: we use Weak because otherwise we leak GC allocations. See https://github.com/LadybirdBrowser/ladybird/commit/73de6d43875503f2f8a724efb5e75fe85054b1c5

kalenikaliaksandr (changes requested):

Libraries/LibWeb/DOM/HTMLCollection.cpp
awesomekling: If individual Weaks create a performance problem here, maybe we can use the WeakContainer mechanism?

tcl3: I've now made `HTMLCollection` implement WeakContainer. This maintains the performance boost I saw previously.

Libraries/LibWeb/DOM/HTMLCollection.cpp
gmta, line 1: Commit title: one "LibWeb:" too many!

tcl3: Oops, fixed :sweat_smile:

trflynn89:
@tcl3  @shannonbooth This branch was a bit too out of date and broke the build on `master`

Edit: Looks like #9404 takes care of it

PR #9187, opened 2026-04-30
Meta: Add lldb pretty printers for AK types

jarusll:
Pretty printers for the following types-
- Atomic: Summary = wrapped value
- Array: Summary = none; Items = 0..N-1
- FixedArray: Summary = size; Items = 0..N-1
- HashMap: Summary = size; Items = key→value
- NonnullRefPtr: Summary = pointer + ref count (never null)
- Optional: Summary = Some/None; Item = value when Some
- OwnPtr: Summary = pointer/null
- RefCounted: Summary = ref count
- RefPtr: Summary = pointer + ref count/null
- SinglyLinkedList: Summary = size; Items = 0..N-1
- StringView: Summary = UTF-8
- String: Summary = UTF-8
- Variant: Summary = active type; Item = value
- Vector: Summary = size; Items = 0..N-1
- ByteString: Summary = UTF-8
- DistinctNumeric: Summary = wrapped value
- IntrusiveList: Summary = none; Items = 0..N-1

Screenshot From 2026-04-30 18-58-24

### DIFF
--- a/.lldbinit
+++ b/.lldbinit
@@ -1,0 +1,1 @@
+command script import ./Meta/Debuggers/lldb/AK.py
--- a/Meta/Debuggers/lldb/AK.py
+++ b/Meta/Debuggers/lldb/AK.py
@@ -1,0 +1,564 @@
+import lldb
+
+
+def ak_atomic_summary(valobj, internal_dict):
+    value = valobj.GetChildMemberWithName("m_value")
+    return value.GetSummary() or value.GetValueAsUnsigned()
+
+
+def ak_bytestring_summary(valobj, internal_dict):
+    impl_ptr = valobj.GetChildMemberWithName("m_impl").GetChildMemberWithName("m_ptr")
+
+    if impl_ptr.GetValueAsUnsigned() == 0:
+        return '""'
+
+    impl = impl_ptr.Dereference()
+    return ak_string_impl_summary(impl, internal_dict)
+
+
+def ak_string_impl_summary(valobj, internal_dict):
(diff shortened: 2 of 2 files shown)


Zaggy1024:
It feels like it may be more appropriate to have a common directory for the debugger-related files in `Meta`, instead of one per debugger. Do you foresee needing a folder other than `PrettyPrinters` in your `Meta/lldb`? Depending on the answer, my suggestions would be either `Meta/PrettyPrinters/lldb` or `Meta/Debuggers/lldb/PrettyPrinters`.

jarusll:
I don't think there is a need for directory besides `PrettyPrinters`. So either of  `Meta/PrettyPrinters/lldb` or `Meta/Debuggers/lldb/PrettyPrinters` would work, but `Meta/PrettyPrinters/lldb` feels better.

This will also require changes in the PR https://github.com/LadybirdBrowser/ladybird/pull/8960

jarusll:
@AtkinsSJ Do let me know if this directory change is to be made as it will require changes for https://github.com/LadybirdBrowser/ladybird/pull/8960 as well

Zaggy1024:
In the interest of getting this moving again, we talked about this on Discord a little bit and we agreed on something like `Meta/Debuggers/lldb/AK.py`. If you edit it to that, I can hit the merge button!

jarusll:
I will make the change. Thank you.

AtkinsSJ (approved):
Looks good! Again, FlyString would be nice to support, but I'm happy merging this as-is.

PR #9192, opened 2026-04-30
LibJS: Avoid function envs for lexical-this arrows

kalenikaliaksandr:
Track whether a function needs environment-backed this resolution separately from whether it needs to allocate its own function environment. Arrow functions that only capture lexical this can now resolve through the outer environment without allocating an empty function environment for every call.

Keep the asm Call path conservative by routing functions that still need lexical-this resolution through the C++ inline-call helper, so the call receiver is not cached as the arrow function's this value.

Microbenchmark:

    function makeLexicalThisArrow() {
        return () => this.value;
    }

    let object = { value: 1, makeLexicalThisArrow };
    let fn = object.makeLexicalThisArrow();
    for (let i = 0; i < 20_000_000; ++i)
        fn();

Measured with the same Release build toggling this patch:

    baseline:  1069.2 ms mean over 12 runs
    optimized:  501.2 ms mean over 12 runs
    speedup:    2.13 times faster


Libraries/LibJS/Bytecode/AsmInterpreter/asmint.asm
```diff
@@ -2023,8 +2023,10 @@ handler Call
     load_pair64 exec_ptr, meta, [shared_data, SHARED_FUNCTION_INSTANCE_DATA_EXECUTABLE], [shared_data, SHARED_FUNCTION_INSTANCE_DATA_ASM_CALL_METADATA]
     branch_bits_clear meta, SHARED_FUNCTION_INSTANCE_DATA_ASM_CALL_METADATA_CAN_INLINE_CALL, .call_slow
     # NewFunctionEnvironment() allocates and has to stay out of the pure asm
-    # path, but we still preserve inline-call semantics via .call_interp_inline.
+    # path. Lexical-this resolution also uses the C++ helper so the call
+    # receiver is not written into the callee's this register.
     branch_bits_set meta, SHARED_FUNCTION_INSTANCE_DATA_ASM_CALL_METADATA_FUNCTION_ENVIRONMENT_NEEDED, .call_interp_inline
+    branch_bits_set meta, SHARED_FUNCTION_INSTANCE_DATA_ASM_CALL_METADATA_THIS_VALUE_NEEDS_ENVIRONMENT_RESOLUTION, .call_interp_inline
```
awesomekling: We could combine these into a single branch if we cache `a || b` and store it in the `SharedFunctionInstanceData`

PR #9193, opened 2026-04-30
LibWeb: Filter rule matches by target pseudo-element

tcl3:
While resolving conflicts in my `::first-letter` PR (#8885), I saw a regression in the style invalidation test expectations. It turns out we were doing matching work for pseudo elements that could never match the given target. We now only consider rules whose rightmost pseudo-element matches the one being queried.

This reduces time spent in `StyleComputer::compute_properties()` by ~30% on https://amazon.co.uk, although overall performance improvement was <1% due to selector matching not being a dominating factor here.

No appreciable difference in StyleBench score.


Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -288,9 +288,17 @@ Vector<StyleComputer::ScopedMatchingRule> StyleComputer::collect_matching_rules(
...
+            // Only consider rules whose target pseudo-element matches the one being queried. Rules with no target
+            // pseudo-element, or with a different target pseudo-element can never match the query and would otherwise
+            // waste work evaluating their compound selectors.
+            auto queried_pseudo_element = *abstract_element.pseudo_element();
             for (auto const& rule : rules) {
-                if (rule.contains_pseudo_element && filter_namespace_rule(element_namespace_uri, rule))
-                    add_rule_to_run(rule, rule_root);
+                auto const& target_pseudo_element = rule.selector.target_pseudo_element();
+                if (!target_pseudo_element.has_value() || target_pseudo_element->type() != queried_pseudo_element)
+                    continue;
```
AtkinsSJ, line 300: I'm not sure this is right... we don't implement it yet, but `::part()` can match either an element or a pseudo-element. eg, [this example](https://drafts.csswg.org/css-shadow-1/#example-b11a0a88) from the spec:

```html
<template id=custom-element-template>
  <p exportparts="::before : preceding-text, ::after : following-text">
    Main text.
</template>
```

So in that case, `::part(preceding-text)` targets `::before`. But in other cases it could target a named real element.

(FIXME would be fine. I don't think `::slotted()` can have any similar issues, and the other pseudo-elements are simple.)

tcl3: I've added a FIXME that will need to be addressed when `exportparts` is added, rather than narrowing the filter now. AFAICT, `::part()` is the only pseudo element affected by this.

AtkinsSJ (approved):

PR #9196, opened 2026-04-30
LibWeb: Split shadow stylesheet host-side invalidation reach

gmta:
Broad shadow-root stylesheet changes already restyle the whole shadow tree, but host-side fallout does not always need a document-wide invalidation. Split the host-side reach classification so selectors contained to the host subtree, such as `:host *` and `:host > *`, invalidate the host, while sibling-escaping selectors such as `:host + :has(*)` still invalidate the host's root.

Recognize sibling escapes through positive selector-list pseudos such as `:is()` and `:where()` as well, including selectors like `:is(:host) + :has(*)` and `:is(:host + .item)`.

This avoids turning host-contained shadow stylesheet changes into full document style invalidations. On https://pomax.github.io/bezierinfo/, this reduces the time to produce a layout tree from about 8.7s to 3.6s on my machine.


awesomekling:
That's a sweet perf win, but I think this might be buggy as-is :)

`:host *` can stay host-subtree bounded, but not all broad host-side shadow selectors can. Something like:

```css
host + :has(*) { background: red; }
```

...can still escape to a sibling/root-side node while also forcing broad invalidation, so changing the combined `may_match_light_dom_under_shadow_host && may_match_shadow_host` case from `host->root().invalidate_style()` to `host->invalidate_style()` can leave that sibling stale.

Could we split the classification a bit more precisely and add  test coverage for the sibling-escape case?

gmta:
> Could we split the classification a bit more precisely and add test coverage for the sibling-escape case?

Good idea, this also reduced the amount of style invalidations needed elsewhere (see rebaseline) while keeping the 58% performance improvement on that page :)

awesomekling:
v nice! I think this will still under-invalidate in some cases though, like:

```css
:is(:host) + :has(*) { }
```

That can still reach a root-side sibling, but I think the new helper only detects direct `:host`, so this would be classified as host-contained. It probably needs to detect `:host` inside selector-list pseudos like `:is()`/`:where()` as well.

There might be more edge cases lurking here. This stuff is unpleasantly fragile due to extreme flexibility. I recommend banging out some wider test coverage. That's what I've been doing for invalidation recently :)

gmta:
@awesomekling 

> I think this will still under-invalidate in some cases though

Forgot about `:is()` et al., have added a wide(r) array of tests!

PR #9202, opened 2026-05-01
LibWeb: Add `align-content` support for block-level boxes

### LINKED ISSUE #4214: LibWeb: align-content on a flex item does not work
Example reduced from https://store.nintendo.co.uk/en/new/latest-releases:
```html
<!doctype html><style>
.flex {
    display: flex;
}
.content-center {
    align-content: center;
}
.box1 {
    width: 100px;
    height: 100px;
    background-color: orange;
}
.box2 {
    width: 50px;
    height: 50px;
    background-color: red;
}
</style><div class="flex"><div class="box1"></div><div class="content-center"><div class="box2">
```

Ladybird:

Image

Chrome:

Image

Pasi4K5:
Closes #4214

As correctly pointed out [here](https://github.com/LadybirdBrowser/ladybird/issues/4214#issuecomment-3192444243), the issue is that `align-content` isn't considered for block containers.

Here are the test cases provided in #4214 with the fix:

image
image

### DIFF
--- a/Libraries/LibWeb/CSS/ComputedValues.h
+++ b/Libraries/LibWeb/CSS/ComputedValues.h
@@ -197,7 +197,7 @@ public:
     static JustifyContent justify_content() { return JustifyContent::FlexStart; }
     static JustifyItems justify_items() { return JustifyItems::Legacy; }
     static JustifySelf justify_self() { return JustifySelf::Auto; }
-    static AlignContent align_content() { return AlignContent::Stretch; }
+    static AlignContent align_content() { return AlignContent::Normal; }
     static AlignItems align_items() { return AlignItems::Stretch; }
     static AlignSelf align_self() { return AlignSelf::Auto; }
     static Appearance appearance() { return Appearance::Auto; }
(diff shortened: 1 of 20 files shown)


Psychpsyo:
Importing the WPT tests for this (https://github.com/web-platform-tests/wpt/tree/e053afbbd005bed4b6100f98f0de744da8d1d09d/css/css-align/blocks) would give us much better coverage than just manually adding 3 tests. Especially for various edge cases that your tests don't seem to cover.

Pasi4K5:
> Importing the WPT tests for this (https://github.com/web-platform-tests/wpt/tree/e053afbbd005bed4b6100f98f0de744da8d1d09d/css/css-align/blocks) would give us much better coverage than just manually adding 3 tests. Especially for various edge cases that your tests don't seem to cover.

The WPT tests are not passing yet, because the [`<baseline-position>`](https://www.w3.org/TR/css-align-3/#typedef-baseline-position) and [`<overflow-position>`](https://www.w3.org/TR/css-align-3/#typedef-overflow-position) values for [`align-content`](https://www.w3.org/TR/css-align-3/#propdef-align-content) don't exist yet, and I want to keep this PR limited to implementing the ones that we do have. Implementing all of them would probably involve working with things like `overflow` that I'm rather unfamiliar with.

I did use the tests to find some issues with my previous implementation though, so thanks for pointing me towards them.

PR #9205, opened 2026-05-01
LibWeb/Bindings: Some small IDL/Bindings related tidy ups

shannonbooth:
Small things found during larger bindings changes and some tidyup leftover from #import removal.


AtkinsSJ (approved):
CI is on fire, but the changes look good.

PR #9218, opened 2026-05-03
LibMedia+LibWeb: Implement HTMLMediaElement.getStartDate()

antoniospg:
* Try to extract `start_time_realtime` from ffmpeg parsed headers
* Plumb to PlaybackManager
* New `JS::Object* HTMLMediaElement::get_start_date()`
* Returns NaN if timeline offset does not exist, per HTML spec
* New tests only cover NaN case

### DIFF
--- a/Libraries/LibMedia/PlaybackManager.h
+++ b/Libraries/LibMedia/PlaybackManager.h
@@ -59,6 +59,8 @@ public:
     void set_duration(AK::Duration duration) { m_duration = duration; }
     AK::Duration current_time() const { return min(m_time_provider->current_time(), duration()); }
 
+    Optional<AK::UnixDateTime> start_time_realtime() const { return m_start_time_realtime; }
+
     auto const& video_tracks() const { return m_video_tracks; }
     auto const& audio_tracks() const { return m_audio_tracks; }
     Optional<Track> preferred_video_track() { return m_preferred_video_track; }
@@ -175,6 +177,7 @@ private:
     Optional<Track> m_preferred_audio_track;
 
     AK::Duration m_duration;
+    Optional<AK::UnixDateTime> m_start_time_realtime;
 
     HashTable<Track> m_tracks_still_buffering;
 
--- a/Libraries/LibWeb/HTML/HTMLMediaElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLMediaElement.cpp
@@ -6,6 +6,7 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
+#include <LibJS/Runtime/Date.h>
 #include <LibJS/Runtime/Promise.h>
 #include <LibMedia/IncrementallyPopulatedStream.h>
 #include <LibMedia/PlaybackManager.h>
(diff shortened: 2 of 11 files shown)


Libraries/LibWeb/HTML/HTMLMediaElement.cpp
```diff
@@ -1689,8 +1696,9 @@ void HTMLMediaElement::on_metadata_parsed()
     m_source_element_selector = nullptr;
 
     // FIXME: 1. Establish the media timeline for the purposes of the current playback position and the earliest possible position, based on the media data.
```
Psychpsyo, line 1703: Isn't this taken care of implicitly by the next step now?
Or is it still missing something?

antoniospg: I believe this doesn't work for unbounded resources like livestreams (see step 4).
I have no idea how to handles this, so  I'll keep the fixme

Libraries/LibWeb/HTML/HTMLMediaElement.cpp
```diff
@@ -473,6 +474,11 @@ double HTMLMediaElement::duration() const
     return m_duration;
 }
 
+JS::Object* HTMLMediaElement::get_start_date()
```
Psychpsyo, line 478: Should link to https://html.spec.whatwg.org/multipage/media.html#dom-media-getstartdate and have a comment with the spec text.

Libraries/LibMedia/PlaybackManager.h
```diff
@@ -59,6 +59,8 @@ class MEDIA_API PlaybackManager final {
     void set_duration(AK::Duration duration) { m_duration = duration; }
     AK::Duration current_time() const { return min(m_time_provider->current_time(), duration()); }
 
+    Optional<double> timeline_offset() const { return m_timeline_offset; }
```
Zaggy1024: It feels a little awkward to use a unitless type for time in LibMedia, especially since double tends to represent seconds in every other place it's used relating to web media, so I'd recommend using `UnixDateTime` from `Time.h`. It doesn't have a `from_microseconds_since_epoch()` function that FFmpegDemuxer would want, but that's simple to add. For conversion back to what LibWeb needs, we could either use `to_nanoseconds() / 1_000_000.0`, or just hook up `Duration::to_seconds_f64() * 1_000.0`.

Zaggy1024: Separate comment since this is less important, but the name "timeline offset" feels ambiguous with the spec's concept of the "earliest possible position" (though both are from the same spec), so I'd tend to prefer going with naming more similar to FFmpeg's in LibMedia. To be clear, in LibWeb, it absolutely should follow the spec naming, though.

Libraries/LibWeb/HTML/HTMLMediaElement.cpp
Zaggy1024: Just to clarify, this step is establishing the earliest possible time in the resource, whereas the timeline offset (named confusingly), seems to be referring to _real time_, and doesn't help us to establish the earliest possible time at all. The relation is simply that the earliest possible time and the time offset _correspond_ to each other.

Zaggy1024 (approved):

PR #9219, opened 2026-05-03
LibWeb: Add units to svg attibutes. Support replaced object-fit and object-position

jonbgamble:
Update SVGLineElement, SVGUseELement, SVGImageElement, SVGEllipseElement, and SVGRectElement to resolve geometry with computed properties. Remove obsoleted value caches.

Use presentation attributes for SVGelement animated length fallback.

Fix wrong value returned by ComputerValues::rx().

Add size & position helpers for replaced content based on the object-fit and object-position algorithms.

https://drafts.csswg.org/css-images-3/#the-object-fit
https://drafts.csswg.org/css-images-3/#the-object-position

Use them for replaced element painting and nested navigable viewport sizing so embedded docs and images don't just always fill the content box.

This PR results in quite a few new WPT ref test failures, mostly in css/css-images/*. This is due to references that suddenly render correctly (due to px support on svg attributes) rather than fail identically to the rendered actual. The improved reference htmls expose nested content timing flakes and unimplemented paths (like embed and video poster frames). For a concrete example, see the object-position-svg-001o.html test that only now (sometimes) renders correctly but had to be added to TestConfig.ini [Skipped] due to flakiness.


Libraries/LibWeb/SVG/SVGRectElement.cpp
```diff
@@ -28,38 +26,25 @@ void SVGRectElement::initialize(JS::Realm& realm)
...
-        m_y = AttributeParser::parse_number_percentage(value.value_or(String {}));
-    } else if (name == SVG::AttributeNames::width) {
-        m_width = AttributeParser::parse_number_percentage(value.value_or(String {}));
-    } else if (name == SVG::AttributeNames::height) {
-        m_height = AttributeParser::parse_number_percentage(value.value_or(String {}));
-    } else if (name == SVG::AttributeNames::rx) {
-        m_radius_x = AttributeParser::parse_number_percentage(value.value_or(String {}));
-    } else if (name == SVG::AttributeNames::ry) {
-        m_radius_y = AttributeParser::parse_number_percentage(value.value_or(String {}));
+    auto node = unsafe_layout_node();
```
shannonbooth, line 31: I believe this is unfortunately going in the opposite direction to where we're starting to move things by making use of a layout node here. For example in #7764. Part of the problem here is that we can't rely on there being a layout node present, such as if this is part of a canvas or something like that.

shannonbooth: Though, @kalenikaliaksandr, that did end up getting closed, maybe you have some more insight here than me?

jonbgamble: Short of spec compliance, I should mention we could get a large chunk of this compatibility just by consuming "px" in AttributeParser.

shannonbooth: Sure, i am not actually 100% what the correct approach here is (might be factoring out parts of parser to call standalone?), but yeah I dont think we should make use of layout nodes here

AtkinsSJ: AttributeParser isn't really fit for purpose for most attributes now. It was originally written for parsing paths, and I think that's all it's really suitable for.

Reading the computed properties is what we want, because CSS can override the attributes and we want that to work. But... yeah the fact those are tied to a layout node is really sketchy. We do have `computed_properties()` here which might be good enough temporarily?

jonbgamble: I'm going to clean out the ladybird fork from my github pretty soon. I'm not sure what happens to the PRs but if they disappear, it's not punitive - this is a heads up in case one of those fixes could save you some tokens or whatever. I'll give ya till next week.

shannonbooth: Ignoring the above for a second to figure out what we should be doing here - so is maybe using a layout node here fine then in the interim, even though we know it's not the long term approach? Sketchy to use, but if we somewhat centralize this then it can be handled in a separate cleanup pass?

PR #9222, opened 2026-05-03
LibWeb: Paint inline-block floats during inline-level painting

### LINKED ISSUE #6742: Text invisible on docs.gl
Here's the reduced test case of the problem on docs.gl:
```html
<style>
    .a {
        background-color: white;
        display: inline-block;
    }

    .b {
        float: left;
    }
</style>

<div class="a">
    <div class="b">test</div>
</div>
```

Firefox:
Image

Ladybird:
Image

Looking at the display list it seems the background somehow gets drawn after the text:
```
Save
  ApplyTransform matrix=[1 0 0 1 0 0]
  SaveLayer
    DrawGlyphRun rect=[8,8 26x18] translation=[8,23.21875] color=rgb(0, 0, 0) scale=1
    FillRect rect=[8,8 25x18] color=rgb(255, 255, 255)
  Restore
Restore
```

shannonbooth:
Avoid recursing into atomic inline-level descendants during the ancestor float sweep.

CSS 2.2 paints inline-block and inline-table boxes in the inline-level painting step, as if they created their own stacking contexts. Paint their internal floats during BackgroundAndBordersForInlineLevelAndReplaced instead.

Fixes some missing text. 

Before:

image

After:

image

Fixes #6742

### DIFF
--- a/Libraries/LibWeb/Painting/StackingContext.cpp
+++ b/Libraries/LibWeb/Painting/StackingContext.cpp
@@ -110,6 +110,32 @@ static PaintPhase to_paint_phase(StackingContext::StackingContextPaintPhase phas
     }
 }
 
+static bool establishes_inline_level_painting_context(Paintable const& paintable)
+{
+    // CSS 2.2 painting order puts inline-block and inline-table boxes in the inline-level painting step and
+    // says to paint each "as if it created a new stacking context", while keeping positioned descendants and
+    // actual child stacking contexts in the parent stacking context:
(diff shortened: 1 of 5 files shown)


kalenikaliaksandr (changes requested):
could you add hit-testing test to make sure we follow the same order  in there?

shannonbooth:
> could you add hit-testing test to make sure we follow the same order in there?

Done, looks like we do based on the test

PR #9223, opened 2026-05-03
LibWeb: Fix input element type change during click cancellation

punthenk:
When an input element's type was changed inside an onclick handler, for example from radio button to date, the legacy cancellation activation behavior was checking for the current type state instead of the type state before activation. This caused the restore logic to be skipped.

The fix is by saving the type state in the pre_activation_behavior and use that in the checks in cancelled_activation_behavior.

This fixes 2 WPT subtests in:
[https://wpt.live/dom/events/Event-dispatch-click.tentative.html](https://wpt.live/dom/events/Event-dispatch-click.tentative.html)

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLInputElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLInputElement.cpp
@@ -2188,6 +2188,7 @@ void HTMLInputElement::legacy_pre_activation_behavior()
 {
     m_before_legacy_pre_activation_behavior_checked = checked();
     m_before_legacy_pre_activation_behavior_indeterminate = indeterminate();
+    m_before_legacy_pre_activation_behavior_type = type_state();
 
     // 1. If this element's type attribute is in the Checkbox state, then set
     // this element's checkedness to its opposite value (i.e. true if it is
@@ -2221,27 +2222,21 @@ void HTMLInputElement::legacy_cancelled_activation_behavior()
     // 1. If the element's type attribute is in the Checkbox state, then set the
     // element's checkedness and the element's indeterminate IDL attribute back
     // to the values they had before the legacy-pre-activation behavior was run.
-    if (type_state() == TypeAttributeState::Checkbox) {
+    if (m_before_legacy_pre_activation_behavior_type == TypeAttributeState::Checkbox) {
         set_checked(m_before_legacy_pre_activation_behavior_checked);
         set_indeterminate(m_before_legacy_pre_activation_behavior_indeterminate);
     }
 
-    // 2. If this element's type attribute is in the Radio Button state, then
-    // if the element to which a reference was obtained in the
-    // legacy-pre-activation behavior, if any, is still in what is now this
-    // element' s radio button group, if it still has one, and if so, set
-    // that element 's checkedness to true; or else, if there was no such
-    // element, or that element is no longer in this element' s radio button
-    // group, or if this element no longer has a radio button group, set
-    // this element's checkedness to false.
-    if (type_state() == TypeAttributeState::RadioButton) {
+    // 2. If this element's type attribute was in the Radio Button state before
+    // the legacy-pre-activation behavior was run, uncheck this element, then
+    // restore the previously checked element in the group, if there was any.
+    if (m_before_legacy_pre_activation_behavior_type == TypeAttributeState::RadioButton) {
+        set_checked(false);
         bool did_reselect_previous_element = false;
         if (m_legacy_pre_activation_behavior_checked_element_in_group) {
             auto& element_in_group = *m_legacy_pre_activation_behavior_checked_element_in_group;
-            if (is_in_same_radio_button_group(*this, element_in_group)) {
-                element_in_group.set_checked_within_group();
-                did_reselect_previous_element = true;
-            }
+            element_in_group.set_checked_within_group();
+            did_reselect_previous_element = true;
 
             m_legacy_pre_activation_behavior_checked_element_in_group = nullptr;
         }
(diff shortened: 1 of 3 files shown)


Libraries/LibWeb/HTML/HTMLInputElement.cpp
```diff
@@ -2221,27 +2222,21 @@ void HTMLInputElement::legacy_cancelled_activation_behavior()
...
-    // 2. If this element's type attribute is in the Radio Button state, then
-    // if the element to which a reference was obtained in the
-    // legacy-pre-activation behavior, if any, is still in what is now this
-    // element' s radio button group, if it still has one, and if so, set
-    // that element 's checkedness to true; or else, if there was no such
-    // element, or that element is no longer in this element' s radio button
-    // group, or if this element no longer has a radio button group, set
-    // this element's checkedness to false.
-    if (type_state() == TypeAttributeState::RadioButton) {
+    // 2. If this element's type attribute was in the Radio Button state before
```
Psychpsyo, line 2230: You're changing a comment here that comes straight from the spec.
1. Don't do that.
2. The fact that you had to change it indicates that this is probably not the right solution. Alternatively, the spec could be wrong, in which case you should file an issue with the WHATWG so that it can be referenced in a new comment here, alongside the current faulty spec text.

Libraries/LibWeb/HTML/HTMLInputElement.cpp
```diff
@@ -2221,27 +2222,21 @@ void HTMLInputElement::legacy_cancelled_activation_behavior()
     // 1. If the element's type attribute is in the Checkbox state, then set the
     // element's checkedness and the element's indeterminate IDL attribute back
     // to the values they had before the legacy-pre-activation behavior was run.
-    if (type_state() == TypeAttributeState::Checkbox) {
+    if (m_before_legacy_pre_activation_behavior_type == TypeAttributeState::Checkbox) {
```
Psychpsyo, line 2225: This is not what it says to do in the spec. (See the comment above this code) Is this an issue with the spec?

Libraries/LibWeb/HTML/HTMLInputElement.cpp
punthenk: Ok, thank you. I did not know that. I will look into a different solution.

Libraries/LibWeb/HTML/HTMLInputElement.cpp
punthenk: I don't think this is a problem with the spec. Firefox passes the tests. Chrome does not. Do you have an idea on what the right approach could be?

Psychpsyo: > I don't think this is a problem with the spec. Firefox passes the tests. Chrome does not. Do you have an idea on what the right approach could be?

If the spec is right and in line with the test, implementing what the spec says verbatim should get Ladybird to pass the test.
In that case, it'd be a matter of reasoning through the algorithms in the spec and comparing them to the Ladybird implementation to see where Ladybird does something different to cause the test failure.

If however the spec disagrees with the test, which is more common than you might expect, it'd be important to find out why and what the desired behavior here is. I'd check things like whether there is an open spec issue on the WHATWG HTML Github for this. (Somewhat commonly the case if the big browser engines disagree)
Alternatively, look the test up on the WPT Github repo to see who added it and when and why. That can lead to prior discussions, sometimes on other browser's bug trackers.
Another avenue is to track down the spec PR that added the spec text in question, to see what discussions were had back when that happened.
Some of all that should hopefully reveal what the intended behavior here is. If it doesn't, file a Github issue on the spec to ask.

Beyond that, I'm not all too familiar with this specific aspect of form behavior, so I can't tell you what the right approach in this specific situation would be, beyond just "whatever the spec says... or something else if the spec is wrong".

punthenk: I have found this PR in the HTML spec. [https://github.com/whatwg/html/issues/5904](https://github.com/whatwg/html/issues/5904). From the [WPT test file on github.](https://github.com/web-platform-tests/wpt/blob/master/dom/events/Event-dispatch-click.tentative.html) So because my change is not currently the way to go in the spec, I would think this is not necessary to change at this point? If you think otherwise I would be happy to hear.

Psychpsyo: > I have found this PR in the HTML spec. [whatwg/html#5904](https://github.com/whatwg/html/issues/5904). From the [WPT test file on github.](https://github.com/web-platform-tests/wpt/blob/master/dom/events/Event-dispatch-click.tentative.html) So because my change is not currently the way to go in the spec, I would think this is not necessary to change at this point? If you think otherwise I would be happy to hear.

In this situation I'd say yes, not changing ladybird to be out of line with the spec currently makes the most sense. Specifically because there's no indication that there's interest from Chrome or Safari or the spec to align with Firefox here.
In theory, if you felt hell-bent on resolving this, you could draft up a spec PR and ping the right people about it to see which way it goes. That's probably the only way this'll ever get resolved.

Psychpsyo: Also, one tip on WPT tests: any that end in `.tentative.html` are tentative - meaning they're testing something that's probably/maybe a sensible thing for a browser to potentially do, according to someone... but not actually in the spec and maybe still up for debate.
(although even tests without `.tentative` sometimes don't quite line up with the spec and ladybird gets to run head-first into all of those, lol)

punthenk: Thank you! I learned a lot here. I think I will leave this as it is.

PR #9225, opened 2026-05-03
LibWeb: Implement `scrollIntoView` for boxes.

dgolebiewski:
Makes `scrollIntoView` work on regular scrollable boxes. With this change there are 14 more WPT tests passing.
This will also help #9060 pass some more WPT tests which fail due to `scrollIntoView` working only on the viewport.

There is still a lot more work to do with supporting different writing directions, but this is a quick win we can take.

Demo:
[Nagranie ekranu_20260503_141512.webm](https://github.com/user-attachments/assets/39bea8d8-85be-41b1-9a1c-9feb94e2792d)

### DIFF
--- a/Libraries/LibWeb/DOM/Element.cpp
+++ b/Libraries/LibWeb/DOM/Element.cpp
@@ -66,6 +66,7 @@
 #include <LibWeb/HTML/CustomElements/CustomElementRegistry.h>
 #include <LibWeb/HTML/CustomElements/CustomStateSet.h>
 #include <LibWeb/HTML/EventLoop/EventLoop.h>
+#include <LibWeb/HTML/EventNames.h>
 #include <LibWeb/HTML/HTMLAnchorElement.h>
 #include <LibWeb/HTML/HTMLAreaElement.h>
 #include <LibWeb/HTML/HTMLBaseElement.h>
@@ -116,6 +117,7 @@
 #include <LibWeb/Painting/PaintableBox.h>
 #include <LibWeb/Painting/StackingContext.h>
 #include <LibWeb/Painting/ViewportPaintable.h>
+#include <LibWeb/PixelUnits.h>
 #include <LibWeb/Platform/EventLoopPlugin.h>
 #include <LibWeb/SVG/SVGAElement.h>
 #include <LibWeb/Selection/Selection.h>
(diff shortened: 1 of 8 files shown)


dgolebiewski:
Maybe a rebase will please the CI overlords? :D

Libraries/LibWeb/DOM/Element.cpp
```diff
@@ -2712,12 +2714,17 @@ static CSSPixelPoint determine_the_scroll_into_view_position(Element& target, Bi
...
     // run the following steps:
 
-    if (!scrolling_box.is_document()) {
-        // FIXME: Add support for scrolling boxes other than the viewport.
+    CSSPixelRect scrolling_box_rect;
+    CSSPixelPoint scroll_offset;
+    if (scrolling_box.is_document()) {
+        // NOTE: For a viewport scrolling box is initial containing block
+        scrolling_box_rect = scrolling_box.document().viewport_rect();
+    } else if (scrolling_box.document().layout_is_up_to_date() && scrolling_box.paintable_box()) {
```
shannonbooth: Just one question here - why would this _not_ be up to date here, and even if it were not up to date, why would we want to early return here?

Libraries/LibWeb/DOM/Element.cpp
```diff
@@ -2873,6 +2892,9 @@ static GC::Ref<WebIDL::Promise> scroll_an_element_into_view(Element& target, Bin
             // -> If scrolling box is associated with an element
             if (scrolling_box.is_element()) {
```
shannonbooth: nit, but It would read a little nicer if we did:


```suggestion
            if (auto element = as_if<Element>(scrolling_box)) {
```

here so we would not need the static cast

Libraries/LibWeb/DOM/Element.cpp
dgolebiewski: Not sure why but i got a few crashes accessing `paintable_box` here, didn't give it much thought and just added the layout up to date check. I suppose using `unsafe_paintable_box` would also be fine here - then we wouldn't have to early return.
I will see if I can replicate the crash again without the layout check, maybe it doesn't happen now.

Libraries/LibWeb/DOM/Element.cpp
dgolebiewski: right, I changed that

Libraries/LibWeb/DOM/Element.cpp
shannonbooth: Right, so I think skipping this step if layout is not up to date or calling `unsafe_paintable_box` would be buggy here since that would mean we're getting metrics from a layout which is out of date.

As far as I can tell we also _shouldn't_ have out of date metrics here since the caller of this function's caller seems to update_layout just before calling this:

```cpp
    // 7. If the element does not have any associated box, or is not available to user-agent features, then return a
    //    resolved Promise and abort the remaining steps.
    document().update_layout(UpdateLayoutReason::ElementScrollIntoView);
    HTML::TemporaryExecutionContext temporary_execution_context { realm() };
    if (!layout_node())
        return WebIDL::create_resolved_promise(realm(), JS::js_undefined());

    // 8. Scroll the element into view with behavior, block, inline, and container. Let scrollPromise be the Promise
    //    returned from this step.
    auto scroll_promise = scroll_an_element_into_view(*this, behavior, block, inline_, container);
```

dgolebiewski: Okay, removed the check and I can't seem to replicate the issue now... Must've been the wind.

PR #9226, opened 2026-05-03
CI: Use CI build image for Linux jobs

gmta:
This means running the build inside a Docker container based on a (more) stable build image.

  * The setup action is now largely simplified for Linux.

  * We hardcode llvm@21 for macOS since it allows us to get rid of specifying llvm_version as an input variable. Since we no longer support switching the LLVM version on Linux via that variable, this seems like the best course of action.

We use a trick to optionally introduce a container image: `fromJSON()` allows `null` as a return value, which causes the job to execute on the runner directly. We need this for macOS and Windows jobs.

### DIFF
--- a/.github/workflows/nightly-lagom.yml
+++ b/.github/workflows/nightly-lagom.yml
@@ -24,6 +24,7 @@ jobs:
         toolchain: ['Clang']
         clang_plugins: [false]
         runner_labels: ['["blacksmith-8vcpu-ubuntu-2404-arm"]']
+        container: [&ci_container '{"image":"ghcr.io/ladybirdbrowser/ladybird-ci:2026.05.3"}']
 
         include:
           - os_name: 'Linux'
@@ -32,6 +33,7 @@ jobs:
             toolchain: 'GNU'
             clang_plugins: false
             runner_labels: '["blacksmith-8vcpu-ubuntu-2404"]'
+            container: *ci_container
 
           - os_name: 'macOS'
             arch: 'arm64'
@@ -39,6 +41,7 @@ jobs:
             toolchain: 'Clang'
             clang_plugins: false
             runner_labels: '["macos-15", "self-hosted"]'
+            container: 'null'
 
           - os_name: 'Linux'
             arch: 'arm64'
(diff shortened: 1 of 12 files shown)


Meta/Linters/check_flatpak.py
```diff
@@ -74,26 +72,25 @@ class DepMatch(Enum):
...
-
-            if "baseline" in result[name]:
-                return result[name]["baseline"]
-    except subprocess.CalledProcessError:
-        print(f"{name} cannot be matched, {VCPKG_REPO} is not at baseline revision: {baseline[:7]}")
+    if baseline not in baseline_versions:
+        url = VCPKG_BASELINE_URL.format(baseline)
+
+        try:
+            with urllib.request.urlopen(url) as response:
```
ADKaster: Unless I'm misreading, this will always fetch the file over the network? For local usage (i.e. commit hooks) should there be a "load from file under Build/vcpkg/" ?

.github/workflows/nightly-lagom.yml
```diff
@@ -32,27 +33,31 @@ jobs:
...
             runner_labels: '["macos-15", "self-hosted"]'
+            container: 'null'
 
           - os_name: 'Linux'
             arch: 'arm64'
             build_preset: 'Distribution'
             toolchain: 'Clang'
             clang_plugins: false
             runner_labels: '["blacksmith-8vcpu-ubuntu-2404-arm"]'
+            container: '{"image":"ghcr.io/ladybirdbrowser/ladybird-ci:2026.05.3"}'
```
ADKaster: Is hard coding the version in 4-5 ci ymls the play? In projects hosted on gitlab I've seen container image name/revision as a repo variable. Though that's harder to lint..

Meta/Linters/check_flatpak.py
gmta: Oh that was unintentional, changed it so we always look at the potential git repo first, network download second. Thanks!

.github/workflows/nightly-lagom.yml
gmta: I've reduced the occurrences somewhat by using yaml anchors, so it's now hardcoded once per workflow file. That still allows us to easily revert to older images for a single workflow instead of switching it for the entire repository, so I'm on the fence about making it a repository variable.

PR #9228, opened 2026-05-03
AK: Align UTF-16 string storage

kalenikaliaksandr:
Align Utf16StringData's UTF-16 flexible storage before it is handed to simdutf for UTF-8 conversion. The previous layout placed the buffer at a 2-byte-aligned offset, so larger vectorized conversions could leave the first code unit corrupted; the Rust parser then saw U+0000 at the start of otherwise valid non-ASCII source and emitted TokenType::Invalid.

Add a LibJS regression covering a long source with non-ASCII text, matching the maps.google.com inline-script failure mode.

### DIFF
--- a/AK/Utf16StringData.h
+++ b/AK/Utf16StringData.h
@@ -153,7 +153,8 @@ private:
 
     union {
         char m_ascii_data[0];
-        char16_t m_utf16_data[0];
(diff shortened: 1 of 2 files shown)


trflynn89:
Woops, didn't realize there was an open PR at the time, but this was superseded by #9315

PR #9232, opened 2026-05-03
LibWeb: Clear stale layout state for hidden slot slottables

kalenikaliaksandr:
When a slot gets content-visibility: hidden, the existing subtree cleanup does not reach assigned slottables, since they are not DOM descendants of the slot.

Clear layout and paint nodes from those assigned subtrees as well, so IntersectionObserver cannot later query geometry through stale paintables.

### DIFF
--- a/Libraries/LibWeb/Layout/TreeBuilder.cpp
+++ b/Libraries/LibWeb/Layout/TreeBuilder.cpp
@@ -826,6 +826,16 @@ void TreeBuilder::update_layout_tree(DOM::Node& dom_node, TreeBuilder::Context& 
             }
 
             pop_parent();
+        } else {
+            // Assigned slottables are not DOM descendants of the slot, so the generic
+            // content-visibility:hidden descendant cleanup above does not reach them.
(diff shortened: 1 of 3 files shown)


Libraries/LibWeb/Layout/TreeBuilder.cpp
```diff
@@ -826,6 +826,17 @@ void TreeBuilder::update_layout_tree(DOM::Node& dom_node, TreeBuilder::Context&
             }
 
             pop_parent();
+        } else {
+            // Assigned slottables are not DOM descendants of the slot, so the generic
+            // content-visibility:hidden descendant cleanup above does not reach them.
+            for (auto const& slottable : slot_element.assigned_nodes_internal()) {
+                slottable.visit([&](auto& slottable_root) {
+                    auto& slottable_root_node = static_cast<DOM::Node&>(*slottable_root);
```
tcl3: Is this `static_cast` necessary given that `Element` and `Text` derive from `Node`?

PR #9241, opened 2026-05-04
CI: Bump base image to ubuntu:26.04; add AT-SPI2 deps

sideshowbarker:
Required for the Linux Qt AT-SPI2 bridge tests in #9156. The 26.04 bump gets us Qt 6.10 (we need ≥6.8) and Orca 50.1 (we need ≥47) directly from apt. Plus the AT-SPI2/xvfb/ Python dependencies the AT-testing harness needs.


Meta/Docker/ci/Dockerfile
```diff
@@ -95,3 +103,54 @@ RUN apt-get update -o APT::Update::Error-Mode=any \
...
+       esac \
+    && /opt/aqt-venv/bin/aqt install-qt "${QT_HOST}" desktop "${QT_VERSION}" "${QT_ARCH}" -O /opt/qt \
+    && QT_INSTALL_SUBDIR="$(basename "$(ls -d /opt/qt/${QT_VERSION}/gcc_*/ | head -1)")" \
+    && ln -s "/opt/qt/${QT_VERSION}/${QT_INSTALL_SUBDIR}" /opt/qt/current \
+    && rm -rf /opt/aqt-venv
+
+ENV PATH=/opt/qt/current/bin:$PATH
+ENV CMAKE_PREFIX_PATH=/opt/qt/current
+
+# Overlay Orca 49.7 onto apt's installed orca 46.1. noble orca has older camelCase naming (settings_manager.getManager);
```
jdahlin: Isnt this cleaner as a vcpkg port/overlay? Especially if you are patching things.

The same might apply to qt above.

cqundefine: Regarding Qt, on FreeBSD (which is quite close to Linux) I've been building Qt with vcpkg for quite a long while and that works just fine, that would probably be much cleaner, I know the maintainers have had mixed feelings about that as it does at time to the build, but once it's cached it's not that bad and it really doesn't take that long for a one off thing.

InvalidUsernameException: Alternatively: With the build now running in Docker, AFAIK we are no longer limited to whatever images GitHub officially supports. Is there anything preventing us from doing this? 
```diff
- FROM ubuntu:24.04
+ FROM ubuntu:26.04
```

 That gets new enough versions of Qt (6.10) and Orca (50.1), avoids patching package-manager-owned files and circumvents the discussion of whether to build Qt from source or not.

sideshowbarker: > Isnt this cleaner as a vcpkg port/overlay? Especially if you are patching things.
> 
> The same might apply to qt above.
>
> Regarding Qt, on FreeBSD (which is quite close to Linux) I've been building Qt with vcpkg for quite a long while and that works just fine, that would probably be much cleaner, I know the maintainers have had mixed feelings about that as it does at time to the build, but once it's cached it's not that bad and it really doesn't take that long for a one off thing.

I personally have no strong opinion about doing it with vcpkg instead. I’d be happy re-do this that way. But I’ll wait for a maintainer to weigh in on that. That said, I would think building Qt would take … quite a while.

sideshowbarker: > Alternatively: With the build now running in Docker, AFAIK we are no longer limited to whatever images GitHub officially supports. Is there anything preventing us from doing this?
> 
> ```diff
> - FROM ubuntu:24.04
> + FROM ubuntu:26.04
> ```
> 
> That gets new enough versions of Qt (6.10) and Orca (50.1), avoids patching package-manager-owned files and circumvents the discussion of whether to build Qt from source or not.

Yeah, I think you’re right — that’d seem preferable, if we could do it. But that also is a decision I guess I’d need to wait to here from them maintainers on.

gmta: We could do that now, yeah - mind you, we still need some way to verify changes to the build image in PRs, it's a lot of manual testing right now.

sideshowbarker: > We could do that now, yeah

OK, updated the PR to do that.

> we still need some way to verify changes to the build image in PRs, it's a lot of manual testing right now

I’d be happy to help with figuring out a way to do that (in a follow-up PR, I guess)

Meta/Docker/ci/Dockerfile
```diff
@@ -20,28 +20,28 @@ RUN apt-get update -o APT::Update::Error-Mode=any \
         lsb-release \
         software-properties-common \
     && UBUNTU_RELEASE="$(lsb_release -cs)" \
-    && curl -fsSL https://apt.llvm.org/llvm-snapshot.gpg.key | gpg --dearmor -o /usr/share/keyrings/llvm-snapshot.gpg \
-    && echo "deb [signed-by=/usr/share/keyrings/llvm-snapshot.gpg] http://apt.llvm.org/${UBUNTU_RELEASE}/ llvm-toolchain-${UBUNTU_RELEASE}-${LLVM_VERSION} main" > /etc/apt/sources.list.d/llvm.list \
```
gmta, line 24: Why do we no longer add this repo? Do they not provide binaries for 26.04?

Meta/Docker/ci/Dockerfile
```diff
@@ -20,28 +20,28 @@ RUN apt-get update -o APT::Update::Error-Mode=any \
...
     && apt-get install -y --no-install-recommends \
+        at-spi2-core \
         autoconf \
         autoconf-archive \
         automake \
         bison \
         build-essential \
         ccache \
         clang-${LLVM_VERSION} \
-        clang++-${LLVM_VERSION} \
```
gmta, line 36: Why is this out?

Meta/Docker/ci/Dockerfile
sideshowbarker: apt.llvm.org apparently has no resolute/26.04 channel. I initially kept that part, but keeping it caused a failure with message about a missing Release file.

We anyway don't seem to need that third-party repo now — because 26.04 ships clang-21 and the rest of the -21 stuff. So it’s all in Ubuntu’s on repos, and we get it from there.

Meta/Docker/ci/Dockerfile
sideshowbarker: In Ubuntu’s own 26.04 repo, there's no separate clang++-21 package. `apt-get install clang++-21` fails. The clang-21 package in Ubuntu’s own repo also includes clang++-21.

The separate clang++-N packaging thing is apparently just something the apt.llvm.org repo does, but Ubuntu’s own repo doesn’t.

So, we just don’t need it. And in fact if we include it, it breaks the build.

Meta/Docker/ci/Dockerfile
gmta: Sure, could you add that explanation to the commit message for future reference? :)

sideshowbarker: > Sure, could you add that explanation to the commit message for future reference? :)

Yup — updated

gmta (approved):

PR #9245, opened 2026-05-04
LibWeb: Fix StackingContext painting order

jorge-pais:
- Add check for painting inline-blocks and inline-tables children in the BackgroundAndBordersForInlineLevelAndReplaced phase
- Add regression css tests

Related issue: https://github.com/LadybirdBrowser/ladybird/issues/6742


jorge-pais:
Closed, duplicate of #9222

PR #9249, opened 2026-05-04
Delete ImmutableBitmap

kalenikaliaksandr:
See commit descriptions


Libraries/LibMedia/FFmpeg/FFmpegVideoDecoder.cpp
```diff
@@ -247,9 +247,9 @@ DecoderErrorOr<NonnullRefPtr<VideoFrame>> FFmpegVideoDecoder::get_decoded_frame(
             }
         }
 
-        auto bitmap = DECODER_TRY_ALLOC(Gfx::ImmutableBitmap::create_from_yuv(move(yuv_data)));
+        auto color_space = DECODER_TRY_ALLOC(Gfx::ColorSpace::from_cicp(yuv_data->cicp()));
 
-        return DECODER_TRY_ALLOC(try_make_ref_counted<VideoFrame>(timestamp, duration, size, bit_depth, cicp, move(bitmap)));
+        return DECODER_TRY_ALLOC(try_make_ref_counted<VideoFrame>(timestamp, duration, size, bit_depth, cicp, move(color_space), move(yuv_data)));
```
Zaggy1024: Why pass both ColorSpace and CICP? Seems redundant, we should either construct ColorSpace from CICP later or just store ColorSpace directly. Not sure which is most efficient, though.

Libraries/LibWeb/HTML/HTMLMediaElement.cpp
```diff
@@ -1554,7 +1556,8 @@ void HTMLMediaElement::set_selected_video_track(Badge<VideoTrack>, GC::Ptr<HTML:
         m_selected_video_track_sink = m_playback_manager->get_or_create_the_displaying_video_sink_for_track(video_track->track_in_playback_manager());
         auto sink_update_result = m_selected_video_track_sink->update();
         if (sink_update_result == Media::DisplayingVideoSinkUpdateResult::NewFrameAvailable) {
-            ensure_external_content_source().update(m_selected_video_track_sink->current_frame());
+            if (auto current_frame = m_selected_video_track_sink->current_frame())
+                ensure_external_content_source().update(current_frame->immutable_bitmap());
```
Zaggy1024: Looks like we're no longer clearing the frame when it changes to nullptr.

Libraries/LibMedia/Sinks/DisplayingVideoSink.h
```diff
@@ -55,8 +54,8 @@ class MEDIA_API DisplayingVideoSink final : public VideoSink {
     RefPtr<VideoDataProvider> m_provider;
     Optional<Track> m_track;
 
-    TimedImage m_next_frame;
-    RefPtr<Gfx::ImmutableBitmap> m_current_frame;
+    TimedVideoFrame m_next_frame;
```
Zaggy1024: If this is the only usage of TimedVideoFrame, I suspect this could just be `RefPtr<VideoFrame>` as well.

Zaggy1024 (approved):

PR #9252, opened 2026-05-05
LibWeb: Stop blocking main fetch on pending preload results

jonbgamble:
The lichess.org lobby stylesheet, and presumably others out there in the wild, sometimes gets render-blocked. This can result in a blank or partially rendered page.

The main fetch would queue work and then spin_until wait for a pending preload result on the main thread. This re-entered the event loop and stranded the queued PendingResponse callback behind a pile of nested event processing. So the preload response existed, but the stylesheet never made it through the normal handoff path.

I chose to follow the PreloadEntry.cpp model to address this and track at most one pending preload PendingResponse for each FetchParams, resolving it directly from the preload callback when the response arrives. This removes the spin_until, keeps the handoff asynchronous, and lets the blocked stylesheet finish loading normally.

If the current single-consumer preload path ever changes, this logic must be widened to handle it.

Although this bug is real, I had no luck reducing it to a javascript/test-web html that fails on master and succeeds here.

### DIFF
--- a/Libraries/LibWeb/Fetch/Fetching/Fetching.cpp
+++ b/Libraries/LibWeb/Fetch/Fetching/Fetching.cpp
@@ -229,6 +229,14 @@ GC::Ref<Infrastructure::FetchController> fetch(JS::Realm& realm, Infrastructure:
         //    response: set fetchParams’s preloaded response candidate to response.
         auto on_preloaded_response_available = GC::create_function(realm.heap(), [fetch_params](GC::Ref<Infrastructure::Response> response) {
             fetch_params->set_preloaded_response_candidate(response);
+
+            // NB: main fetch may already be parked on this preload result.
+            // Resolve that waiter here instead of spinning the event loop.
+            auto controller = fetch_params->controller();
+            if (auto pending_preloaded_response = controller->pending_preloaded_response()) {
+                controller->set_pending_preloaded_response(nullptr);
+                pending_preloaded_response->resolve(response);
+            }
         });
 
         // 3. Let foundPreloadedResource be the result of invoking consume a preloaded resource for request’s
@@ -456,10 +464,22 @@ GC::Ptr<PendingResponse> main_fetch(JS::Realm& realm, Infrastructure::FetchParam
 
         // -> fetchParams’s preloaded response candidate is not null
         if (!fetch_params.preloaded_response_candidate().has<Empty>()) {
-            // 1. Wait until fetchParams’s preloaded response candidate is not "pending".
-            HTML::main_thread_event_loop().spin_until(GC::create_function(vm.heap(), [&] {
-                return !fetch_params.preloaded_response_candidate().has<Infrastructure::FetchParams::PreloadedResponseCandidatePendingTag>();
-            }));
+            // 1. Wait until fetchParams's preloaded response candidate is not "pending".
+            if (fetch_params.preloaded_response_candidate().has<Infrastructure::FetchParams::PreloadedResponseCandidatePendingTag>()) {
+                // NB: The spec says to wait here. Spinning the main-thread event
+                // loop strands the queued preload callback behind nested work,
+                // so we park a PendingResponse and let that callback resolve it.
+                auto controller = fetch_params.controller();
+
+                // NB: Reuse the parked response if this branch is re-entered
+                // before the preload callback runs.
+                if (auto pending_preloaded_response = controller->pending_preloaded_response())
+                    return *pending_preloaded_response;
+
+                auto pending_preloaded_response = PendingResponse::create(vm, request);
+                controller->set_pending_preloaded_response(pending_preloaded_response);
+                return pending_preloaded_response;
+            }
 
             // 2. Assert: fetchParams’s preloaded response candidate is a response.
             VERIFY(fetch_params.preloaded_response_candidate().has<GC::Ref<Infrastructure::Response>>());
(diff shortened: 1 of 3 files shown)


Libraries/LibWeb/Fetch/Infrastructure/FetchParams.h
```diff
@@ -50,6 +60,9 @@ class WEB_API FetchParams : public JS::Cell {
     [[nodiscard]] PreloadedResponseCandidate const& preloaded_response_candidate() const { return m_preloaded_response_candidate; }
     void set_preloaded_response_candidate(PreloadedResponseCandidate preloaded_response_candidate) { m_preloaded_response_candidate = move(preloaded_response_candidate); }
 
+    [[nodiscard]] GC::Ptr<Fetching::PendingResponse> pending_preloaded_response() const { return m_pending_preloaded_response; }
+    void set_pending_preloaded_response(GC::Ptr<Fetching::PendingResponse> pending_preloaded_response) const { m_pending_preloaded_response = pending_preloaded_response; }
```
shannonbooth: Seems quite wrong to have a setter const qualified

Libraries/LibWeb/Fetch/Fetching/Fetching.cpp
```diff
@@ -456,10 +461,15 @@ GC::Ptr<PendingResponse> main_fetch(JS::Realm& realm, Infrastructure::FetchParam
 
         // -> fetchParams’s preloaded response candidate is not null
         if (!fetch_params.preloaded_response_candidate().has<Empty>()) {
-            // 1. Wait until fetchParams’s preloaded response candidate is not "pending".
-            HTML::main_thread_event_loop().spin_until(GC::create_function(vm.heap(), [&] {
-                return !fetch_params.preloaded_response_candidate().has<Infrastructure::FetchParams::PreloadedResponseCandidatePendingTag>();
-            }));
+            // 1. Wait until fetchParams's preloaded response candidate is not "pending".
+            if (fetch_params.preloaded_response_candidate().has<Infrastructure::FetchParams::PreloadedResponseCandidatePendingTag>()) {
+                if (auto pending_preloaded_response = fetch_params.pending_preloaded_response())
```
shannonbooth: I would need to spend more time trying to understand what's going on here, but this seems like it could be very flimsy

Also in terms of creating a test, by any chance did you give it a go with Claude or something? It seems pretty effective at coming up with tests in my experience, though this does seem like a quite annoying case.

Libraries/LibWeb/Fetch/Infrastructure/FetchParams.h
jonbgamble: Agreed, and this field is more logically attached to the controller anyway.

Libraries/LibWeb/Fetch/Fetching/Fetching.cpp
jonbgamble: It may be flimsy in that every base may not be covered, but I suspect it is an improvement on current master.

I am not sure about an LLM test for this. Re CC, I would consider buying low on Claude output tokens as an investment opportunity. Actually if I could afford Claude, I would use that money to buy healthcare.

shannonbooth: I gave it a spin and it made an attempt modeling off the preload tests in WPT using ?pipe=trickle, unfortunately didn't seem to do the trick, seems difficult to get anything close to reliably triggering since it's super timing / event driven. It did happen to leave some feedback in that attempt when running though which I agree with (pasted below)  - think we should just throw in some comments explaining what is going on here since whenever spec says "wait for" is a place to think hard about what we need to do

>   // 1. Wait until fetchParams's preloaded response candidate is not "pending".                                                                                                                                                                                                                                                                                                                              
>   // NOTE: The spec says to wait synchronously, but spin_until re-enters the event loop,                                                                                                                                                                                                                                                                                                                     
>   //       which strands the preload's queued callback behind nested event processing.
>   //       Instead we park a PendingResponse on the controller and resolve it from the                                                                                                                                                                                                                                                                                                                       
>   //       preload callback when the response arrives.
>                                                                                                                                                                                                                                                                                                                                                                                                              
>   2. The callback half is uncommented. The on_preloaded_response_available lambda silently grew a second job — resolving the pending response — with no explanation of why:                                                                                                                                                                                                                                  
>    
>   fetch_params->set_preloaded_response_candidate(response);                                                                                                                                                                                                                                                                                                                                                  
>                                                                                                                                                                                                                                                                                                                                                                                                              
>   // NOTE: If main_fetch reached the pending check before the response arrived, it
>   //       left a PendingResponse here. Resolve it directly rather than spinning.                                                                                                                                                                                                                                                                                                                            
>   auto controller = fetch_params->controller();                                                                                                                                                                                                                                                                                                                                                              
>   if (auto pending_preloaded_response = controller->pending_preloaded_response()) {
>                                                                                                                                                                                                                                                                                                                                                                                                              
>   3. The early-return guard is mysterious. The if (auto p = controller->pending_preloaded_response()) return *p; at the top of the pending branch handles being called a second time before the preload resolves, but this is non-obvious. A comment on when that can happen (e.g. redirect re-entry) would help, or at minimum note it's defensive.                                                         
>                                                                                                                                                                                                                                                                                                                                                                                                              
>   4. The single-consumer assumption in the commit message should be a NOTE in FetchController.h where the member lives — not just in git history:                                                                                                                                                                                                                                                            
>                   
>   // NOTE: Assumes at most one pending consumer per preloaded response.                                                                                                                                                                                                                                                                                                                                      
>   //       If the preload path is ever widened to multiple consumers, this must be extended.                                                                                                                                                                                                                                                                                                                 
>   GC::Ptr<Fetching::PendingResponse> m_pending_preloaded_response;

jonbgamble: Comments added.

PR #9257, opened 2026-05-05
UI/Qt+LibCore: Lock IOSurface for safe CPU reads on macOS

christianfrey:
On Intel Macs, the IOSurface backing the Qt QImage may be updated by the GPU, but CPU access is not guaranteed to be coherent without explicit synchronization. This results in blank pages in the Qt UI.

This change adds lock_read_only() / unlock_read_only() around the Qt rendering path to ensure proper synchronization when reading from the IOSurface.

Apple Silicon is unaffected due to unified memory, but the locking is kept for correctness.

### DIFF
--- a/Libraries/LibCore/IOSurface.cpp
+++ b/Libraries/LibCore/IOSurface.cpp
@@ -125,4 +125,14 @@ void* IOSurfaceHandle::core_foundation_pointer() const
     return m_ref_wrapper->ref;
 }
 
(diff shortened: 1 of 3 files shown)


christianfrey:
Open to any feedback or suggestions if needed.

christianfrey:
Superseded by #9798, the issue is fixed there, closing this one.

PR #9262, opened 2026-05-05
Update build instructions for openSUSE Leap

trmazi:
Updated openSUSE installation instructions for Leap 16.0

Adds `cargo` and `cargo-c` for Rust.

I've just built on Leap, testing Tumbleweed now.


shannonbooth:
Rust is already mentioned in build pre requisites here: https://github.com/trmazi/ladybird/blob/e60870fa6c492abf61d4506f774f12c0bde4b4a7/Documentation/BuildInstructionsLadybird.md#build-prerequisites

But this has caused confusion enough times that I think it would make sense to clarify or change in some way (but consistent for all platforms).

PR #9263, opened 2026-05-05
LibWeb: Update hovered nested navigable on mouse leave

jonbgamble:
When the pointer leaves an iframe, we need to notify any hovered node. Otherwise, canvas animations that pause on mouseleave will continue, pigs will rain from the sky, the chicken will keep crossing the road, etc


Libraries/LibWeb/Page/EventHandler.cpp
gmta, line 1: > pigs will rain from the sky, the chicken will keep crossing the road, etc

Please leave out the funny stuff from commit messages

jonbgamble: Sorry bout that.

PR #9264, opened 2026-05-05
Base: Add gui framework info to about:version

Totto16:
I am not sure if that is the right place for that info, and if it is even required, but I saw that multiple GUI frameworks are supported, so the "Build Details" overview seems to be a good place to tell the user that. (even if Qt / Gtk don't look that similar and to someone who knows the difference it is blatantly obvious)

It might also help in cases of (GUI) bug reports from users, that don't know the difference, the can just look it up there 😄 

I am not sure what the project idiomatic way of supplying the GUI framework string to the C++ code is, so I used `configure_file`, if I should do it in another way, feel free to tell me 😄 

A screenshot from the new `about:version`

grafik

### DIFF
--- a/Base/res/ladybird/about-pages/version.html
+++ b/Base/res/ladybird/about-pages/version.html
@@ -135,6 +135,10 @@
                     <div id="platform-name" class="detail-value"></div>
                 </div>
                 <div class="detail-row">
+                    <div class="detail-label">GUI Framework</div>
+                    <div id="gui-framework" class="detail-value"></div>
+                </div>
+                <div class="detail-row">
                     <div class="detail-label">User Agent</div>
                     <div id="user-agent" class="detail-value"></div>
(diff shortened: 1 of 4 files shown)


jonbgamble:
This is great!

I would just add to the wishlist (not necessarily for this PR), we could get the HEAD commit hash of the build in there somehow.

I think the same VersionUI.cpp -> JSON -> version.html path would work, given a new JSON field and html row. Generate a header with LADYBIRD_GIT_COMMIT_HASH. CMake just runs `git rev-parse --short=10 HEAD` and only touches the header when that string changes. configure_file wouldn't work for this case as it needs a build artifact.

shannonbooth:
Luke actually has that open in https://github.com/LadybirdBrowser/ladybird/pull/9075

gmta:

PR #9269, opened 2026-05-06
RequestServer+LibCompress: Fix invalid Content-Encoding

AleksanderEvensen:
Hi, first time contributor here.

I was looking for something to help out with, and then stumbled upon an error while trying to use the Norwegian Bank login system. Apparently the server there returns an invalid Content-Encoding `ISO-8859-1` which curl trows as an error

```
HTTP/1.1 200
Date: Tue, 05 May 2026 18:30:17 GMT
Server: web
Access-Control-Allow-Origin: https://csfe.bankid.no
Content-Encoding: ISO-8859-1 <----- This one
Content-Type: application/json
Strict-Transport-Security: max-age=31536000
X-Content-Type-Options: nosniff
Referrer-Policy: no-referrer
Keep-Alive: timeout=60, max=1000
Connection: Keep-Alive
```

I got curious then and tried to figure out how chrome did this and as i suspected they just ignore these invalid encodings and flag the whole body as uncompressed bytes see: https://chromium.googlesource.com/chromium/src/+/refs/heads/main/net/filter/filter_source_stream.cc

The problem is that libcurl doesn't provide a way to do this behavior for us as discussed by @InvalidUsernameException on the libcurl github discussion page. And it doesnt seem like they're planning to implement it either.

So i did the only logical thing and tried to fix it myself. Currently the implementation disables libcurl automatic body decompression. This means we have to implement it ourselves. So i created a BodyDecoder which takes a encoding chain which is the comma seperated list in the header and streams the bytes through the decompressors using the already existing methods in `LibCompress`. But when encountering a unknown token it will treat the request as a normal uncompressed body.

There are however some missing implementation details hence why i marked the PR as a draft.

- [ ] LibCompress does not have brotli decompressor implemented and we do not have brotli as a direct dependency as far as i know.
- [x] LibCompress does not have zstd decompressor implemented and we do not have a direct dependency on zstd either as far as i know
- [ ] [Firefox](https://searchfox.org/firefox-main/rev/8352bcb6d75d53f3e2190221b71190e47afa0bfc/netwerk/streamconv/converters/nsHTTPCompressConv.cpp#810) and [chrome](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/net/filter/gzip_source_stream.cc) auto-detects if its zlib-wrapped deflate or just raw deflate calls which would require a wrapper around the existing deflate implementation.

The current pushed code works for loading youtube and watching videos as well as the norwegian bank login service (haven't really done much more testing than that). Since i don't want to cause any regressions by replacing libcurl's logic i would gladly like some comments on the code and implementation before committing all the way and doing the rest.

### DIFF
--- a/Libraries/LibCompress/CMakeLists.txt
+++ b/Libraries/LibCompress/CMakeLists.txt
@@ -4,9 +4,12 @@ set(SOURCES
     Gzip.cpp
     PackBitsDecoder.cpp
     Zlib.cpp
+    Zstd.cpp
 )
+
+find_package(zstd CONFIG REQUIRED)
 
 ladybird_lib(LibCompress compress)
 target_link_libraries(LibCompress PRIVATE LibCore LibCrypto)
 
-target_link_libraries(LibCompress PRIVATE ZLIB::ZLIB)
+target_link_libraries(LibCompress PRIVATE ZLIB::ZLIB zstd::libzstd)
--- a/Libraries/LibCompress/Deflate.cpp
+++ b/Libraries/LibCompress/Deflate.cpp
@@ -9,6 +9,8 @@
 #include <AK/BinarySearch.h>
 #include <LibCompress/Deflate.h>
 #include <LibCompress/DeflateTables.h>
+#include <LibCompress/Helpers.h>
+#include <LibCompress/Zlib.h>
 
 #include <zlib.h>
 
--- a/Libraries/LibCompress/Deflate.h
+++ b/Libraries/LibCompress/Deflate.h
@@ -8,8 +8,11 @@
 
 #pragma once
 
+#include <AK/Array.h>
 #include <AK/BitStream.h>
+#include <AK/OwnPtr.h>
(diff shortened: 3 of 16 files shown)


InvalidUsernameException:
Lovely, another site doing this nonsense! :upside_down_face: 

I've just asked on the curl Discussions for another response, let's see what result that has.

The idea of just disabling content decoding in curl and handling it in Ladybird directly was floated on Discord as well. My thought on this was that, if doing so, it might be a good target to implement the decoding in Rust: It is dealing with untrusted data, compression algorithms seem to be one of those things that are easy to get wrong memory-safety wise, the FFI is basically just bytes go in, bytes go out, and `RequestServer` might be a good target for fully porting to Rust later. But that is something maintainers will need to comment on.

I'd recommend you to join the Discord if you haven't already and raise this discussion there. You're more likely to get maintainer feedback that way.

gmta:

PR #9276, opened 2026-05-06
LibJS: Fix var and func declarations in direct eval inside catch blocks

Martin005:
This PR implements support for Annex B.3.4 of the ECMAScript specification, allowing `eval`-declared `var` and function bindings to be redeclared in catch blocks.

See: https://github.com/tc39/test262/blob/main/test/annexB/language/eval-code/direct/var-env-lower-lex-catch-non-strict.js

By the way, this was the last failing test from Annex B - `eval_code` 🥳 

image

### DIFF
--- a/Libraries/LibJS/Runtime/AbstractOperations.cpp
+++ b/Libraries/LibJS/Runtime/AbstractOperations.cpp
@@ -820,11 +820,16 @@ ThrowCompletionOr<void> eval_declaration_instantiation(VM& vm, EvalDeclarationDa
                 for (auto const& name : data.var_names) {
                     // a. If ! thisEnv.HasBinding(name) is true, then
                     if (MUST(this_environment->has_binding(name))) {
-                        // i. Throw a SyntaxError exception.
-                        return vm.throw_completion<SyntaxError>(ErrorType::TopLevelVariableAlreadyDeclared, name);
-
-                        // FIXME: ii. NOTE: Annex B.3.4 defines alternate semantics for the above step.
-                        // In particular it only throw the syntax error if it is not an environment from a catchclause.
+                        // B.3.4 Changes to EvalDeclarationInstantiation, https://tc39.es/ecma262/#sec-evaldeclarationinstantiation
+                        // i. Normative Optional
+                        //     If the host is a web browser or otherwise supports VariableStatements in Catch Blocks, then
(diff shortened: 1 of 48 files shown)


Libraries/LibJS/Runtime/AbstractOperations.cpp
```diff
@@ -820,11 +820,13 @@ ThrowCompletionOr<void> eval_declaration_instantiation(VM& vm, EvalDeclarationDa
                 for (auto const& name : data.var_names) {
                     // a. If ! thisEnv.HasBinding(name) is true, then
                     if (MUST(this_environment->has_binding(name))) {
+                        // NOTE: Annex B.3.4 adds additional steps at this point.
+                        // In non-strict mode, allow var declarations in direct eval
+                        // to shadow catch parameter bindings (they "penetrate" the catch clause)
+                        if (!strict)
+                            continue;
```
gmta: Could you add a newline after this line, and rewrap the comment at column 120?

Libraries/LibJS/Runtime/AbstractOperations.cpp
```diff
@@ -820,11 +820,13 @@ ThrowCompletionOr<void> eval_declaration_instantiation(VM& vm, EvalDeclarationDa
                 for (auto const& name : data.var_names) {
                     // a. If ! thisEnv.HasBinding(name) is true, then
                     if (MUST(this_environment->has_binding(name))) {
+                        // NOTE: Annex B.3.4 adds additional steps at this point.
```
gmta: Please keep the spec text verbatim:

```cpp
// NOTE: Annex B.3.4 defines alternate semantics for the above step.
```

Libraries/LibJS/Runtime/AbstractOperations.cpp
```diff
@@ -820,11 +820,13 @@ ThrowCompletionOr<void> eval_declaration_instantiation(VM& vm, EvalDeclarationDa
...
+                        // In non-strict mode, allow var declarations in direct eval
+                        // to shadow catch parameter bindings (they "penetrate" the catch clause)
```
gmta: Can you use the exact spec text from the annex B.3.4 that pertains here?

Martin005: @gmta Thanks for the comment, I copied the spec text and updated the implementation accordingly 🙂

PR #9281, opened 2026-05-06
test-web: Wait for PDF.js layers to render before signalling completion

tcl3:
Previously, we signalled completion of a PDF test once the `pagesloaded` event fired. This fires once page metadata is initialized but the text and annotation layers are rendered asynchronously after that. We now wait for `textlayerrendered` and `annotationlayerrendered` to fire for every page before signaling test completion.

Tested by running `test-web --repeat 1000 -j16 -f Layout/input/pdf-viewer.pdf` locally.

### DIFF
--- a/Tests/LibWeb/test-web/main.cpp
+++ b/Tests/LibWeb/test-web/main.cpp
@@ -276,12 +276,32 @@ static auto wait_for_reftest_completion = generate_wait_for_test_string("reftest
 
 // For PDF tests: by the time this runs (after on_load_finish), ladybirdpdf has already fired
 // and openPdf() is in progress; initializedPromise lets us hook in at the right moment.
-// Text mode uses the page count; Layout mode ignores the signalTestIsDone argument.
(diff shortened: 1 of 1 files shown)


jdahlin:
Looks good to me, perhaps add some code/url references for these events, both our and pdf.js's

tcl3:
I can't find any solid documentation for these events. When I was figuring out what the issue was a Google search brought me to this issue: https://github.com/mozilla/pdf.js/issues/5601. I then grepped the pdf.js codebase for any other `*rendered` event names.

I don't think this is intended to be a purely internal API, but I wouldn't be 100% confident that it will survive across major releases either.

Also, it doesn't look like this approach will work for PDFs with more pages than are visible in the viewer, since the pages are loaded lazily as you scroll.

tcl3:
I couldn't really find a simple way of forcing all pages to render or to identify the set of currently visible pages. What I've done for now is made the test runner complain and exit early if the PDF has more than 1 page.

gmta (approved):

PR #9282, opened 2026-05-06
LibGC: Trace keys in RootHashMap, add container tests and only allow Cell/NanBoxedValue types in containers

Lubrsi:
This is 1 out of roughly 18 parts of a series of fixing up GC container issues, which were found by adding plugin detections.

This first part and the rest of the beginning parts are introducing supporting infrastructure for the future parts, where the future parts are about fixing found issues, then finally introducing the new plugin detections.

The first part is about fixing a bug in RootHashMap: keys were not traced.
Then it adds tests for all current container types to check if they correctly report their contents to the GC, including a test for the RootHashMap bug.
Then the start of catching some bugs at compile time with accompanying compile time tests, by adding static_asserts to only allow Cell/NanBoxedValue types in containers (with the exception of maps, where only one of the key and value must at least be a Cell/NanBoxedValue).


Hendiadyoin1:
Looks a small oopsie happened with the history in the first commit, 
Or is it intentional that an assertion is added there aswell

Lubrsi:
Ah, just a history goof :^)
Fixed now.

gmta (approved):

PR #9283, opened 2026-05-06
LibWeb+LibWebView+WebContent+UI: Add an action to cut text

trflynn89:
I kept getting surprised that ctrl+x doesn't work!


jdahlin (commented):
Could you please update the Gtk frontend too?

PR #9284, opened 2026-05-06
LibWebView + UI/Qt: Add reopening of recently closed tabs and windows

AtkinsSJ:
I kept trying to do this with Ctrl+Shift+T, so now it works! :tada:

First commits are some adjustments to HistoryStore, namely to only have a single *Storage object instead of either one or two, when we only ever use one of them. And then an adjustment to a test that wasn't actually testing what it claimed to.

For the main event: Closed tabs and windows are stored on the HistoryStore, as a stack. (Well, actually a Vector because AK::Stack is fixed-size, which is a bit annoying.) We keep track of whether there's anything on that stack, and whether the top entry is for a tab or a window, so that the menu item changes description and gets disabled when it should.

For now it's quite simple: we store the URLs, and then for windows we store which tab was selected. That means you don't get the "it's as if you never closed the page" experience of other browsers.

Only Qt is implemented. I tried AppKit too, but Codex had a lot of trouble distinguishing between tab and window closures and I'm not convinced what it came up with is any good. You can view those changes on [this branch](https://github.com/AtkinsSJ/ladybird/tree/reopen-recent-tabs-appkit). The first commit there worked and was for an earlier version that only saved individual tabs; the second is the diff to support window closure and it technically does work but it also seems quite cursed and I don't know ObjC or AppKit well enough to know. I have a strong feeling that someone who knows what they're doing won't find it hard to do.

cc: @trflynn89

### DIFF
--- a/Libraries/LibWebView/Application.cpp
+++ b/Libraries/LibWebView/Application.cpp
@@ -839,6 +839,8 @@ NonnullRefPtr<Core::Promise<Application::BrowsingDataSizes>> Application::estima
 
 void Application::clear_browsing_data(ClearBrowsingDataOptions const& options)
 {
+    bool did_change_history = false;
+
     if (options.delete_cached_files == ClearBrowsingDataOptions::Delete::Yes) {
         m_request_server_client->async_remove_cache_entries_accessed_since(options.since);
 
@@ -851,13 +853,18 @@ void Application::clear_browsing_data(ClearBrowsingDataOptions const& options)
         });
     }
 
-    if (options.delete_history == ClearBrowsingDataOptions::Delete::Yes)
+    if (options.delete_history == ClearBrowsingDataOptions::Delete::Yes) {
         m_history_store->remove_entries_accessed_since(options.since);
+        did_change_history = true;
+    }
 
     if (options.delete_site_data == ClearBrowsingDataOptions::Delete::Yes) {
         m_cookie_jar->expire_cookies_accessed_since(options.since);
         m_storage_jar->remove_items_accessed_since(options.since);
     }
+
+    if (did_change_history)
+        on_recently_closed_entries_changed();
 }
 
 void Application::clear_history()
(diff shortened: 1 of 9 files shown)


jdahlin:
Would be nice if Gtk could be updated here as well, if you point Claude/Codex at [epiphanys codebase](https://gitlab.gnome.org/GNOME/epiphany) which also uses libadwaita it is usually able to come up with good solution. I just double checked, epiphany/gnome web implements this feature.

trflynn89:
@jdahlin I'd prefer we hold off on big Gtk features until we settle on safe-language bindings.

AtkinsSJ:
Yeah I intentionally skipped Gtk as I saw the discussion about switching it to Rust. Seems like it'd just cause conflicts.

Nic70807 (approved):

Libraries/LibWebView/Application.cpp
```diff
@@ -839,6 +839,8 @@ NonnullRefPtr<Core::Promise<Application::BrowsingDataSizes>> Application::estima
```
elderjshaw, line 839: Agreed with this change, thank you.

Libraries/LibWebView/HistoryStore.h
```diff
@@ -58,41 +58,68 @@ class WEBVIEW_API HistoryStore {
...
+        Statements m_statements;
     };
 
-    explicit HistoryStore(Optional<PersistedStorage>, bool is_disabled = false);
+    explicit HistoryStore(NonnullOwnPtr<StorageImpl>&&, bool is_disabled = false);
 
-    Optional<PersistedStorage> m_persisted_storage;
-    TransientStorage m_transient_storage;
+    NonnullOwnPtr<StorageImpl> m_storage;
     bool m_is_disabled { false };
```
AtkinsSJ, line 123: Thinking about this further, "disabled" is really a third type of Storage, where everything is a no-op. The difference would be that we currently early-out of methods when disabled, and so we'd end up doing a bit of unnecessary work if we removed the flag in favour of a third StorageImpl.

PR #9285, opened 2026-05-06
AK+LibWeb: Add a `fast_as<T>()` method for quickly casting mixin types and use it

tcl3:
This change was motivated by seeing `user_select_used_value()` show up in a profile of https://en.wikipedia.org/wiki/Holy_Roman_Empire. Every time a table of contents link was clicked we would spend ~20ms in this method. This change reduces that by ~90%.

### DIFF
--- a/AK/TypeCasts.h
+++ b/AK/TypeCasts.h
@@ -38,7 +38,9 @@ ALWAYS_INLINE bool is(NonnullRefPtr<InputType> const& input)
 template<typename OutputType, typename InputType>
 ALWAYS_INLINE CopyConst<InputType, OutputType>* as_if(InputType& input)
 {
-    if constexpr (requires { input.template fast_is<RemoveCVReference<OutputType>>(); static_cast<CopyConst<InputType, OutputType>*>(&input); }) {
+    if constexpr (requires { input.template fast_as<RemoveCVReference<OutputType>>(); }) {
+        return input.template fast_as<RemoveCVReference<OutputType>>();
+    } else if constexpr (requires { input.template fast_is<RemoveCVReference<OutputType>>(); static_cast<CopyConst<InputType, OutputType>*>(&input); }) {
         if (!is<OutputType>(input))
             return nullptr;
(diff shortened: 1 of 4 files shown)


Libraries/LibWeb/DOM/Node.cpp
```diff
@@ -1855,6 +1855,15 @@ Slottable Node::as_slottable()
...
 }
 
+HTML::FormAssociatedTextControlElement& Node::as_form_associated_text_control_element()
+{
+    VERIFY(is_form_associated_text_control_element());
+
+    if (auto* input = as_if<HTML::HTMLInputElement>(*this))
+        return *input;
+    return as<HTML::HTMLTextAreaElement>(*this);
+}
```
gmta: Typo in the commit title: "Assoicated"

gmta: Wouldn't we be able to leverage `fast_if<T>` instead of directly calling `is_form_associated_text_control_element()`?

tcl3: My understanding is that `fast_is<FormAssociatedTextControlElement>` wouldn't work. Currently, if  `fast_is<OutputType>` is present then `as_if<OutputType>(input)` will try to static cast directly to the output type which isn't valid for this kind of mixin type.

We'd probably need a `fast_as<T>()` that allows us to specify how the cast should be done instead then we'd modify as_if<T>() to use that when available.

gmta: Oh, I was thinking it could at least take the load off of `is<FormAssociatedTextControlElement>` which would otherwise do a `dynamic_cast`. Casting wouldn't work, no.

tcl3: I've added the `fast_as<T>()` mechanism I mentioned above and added `fast_as<T>()` and `fast_is<T>()` implementations for `FormAssoicatedTextControlElement`.

PR #9286, opened 2026-05-06
LibGC: External memory accounting + tuned thresholds

awesomekling:
LibGC was sizing the heap based only on cell allocations, ignoring everything held in external buffers (`BigInt` limbs, `ArrayBuffer` payloads, DOM text and image data, stylesheets, media cache, module storage, keyed collections, etc, etc...).

This branch teaches `Cell` to report its external footprint so collection, thresholds reflect actual memory pressure, then retunes the threshold floor (4 MiB → 8 MiB) and growth factor (1.0× → 1.75×) on top of the new accounting.

The retune comes from a sweep of 13 `(min, growth)` pairs across Speedometer 2/3, Octane, GarBench.

Comparing Speedometer 2 & 3 on master vs this PR:

|                | Speedo 2 master | Speedo 2 PR    | Speedo 3 master | Speedo 3 PR    |                                                                
|----------------|-----------------|----------------|-----------------|----------------|
| score          | 50.1            | 52.0 (+3.8%)   | 2.97            | 3.15 (+6.1%)   |                                                                
| WebContent RSS | 1438 MB         | 1647 MB (+15%) | 2083 MB         | 2473 MB (+19%) |
| GC count       | 158             | 47 (−70%)      | 95              | 28 (−70%)      |                                                                
| total GC time  | 1.65 s          | 1.09 s (−34%)  | 1.65 s          | 1.21 s (−27%)  |
| max GC pause   | 38 ms           | 76 ms          | 97 ms           | 139 ms         |          

Aggregate time in GC drops 27-34%, so the user spends less total wall-clock waiting on collection. The tradeoff is that the remaining pauses are bigger: many small sub-perception pauses get traded for fewer larger ones, with +15-19% steady-state RSS as the GC legitimately lets the heap grow longer between collections.

The +RSS isn't free; it's the cost of letting the heap reflect actual pressure. More aggressive growth factors (2.5×, 3.0×) score slightly higher but pay 30–40% RSS, and 3.0× actually introduces a 223 ms worst-case pause despite running fewer GCs. 1.75× is the cleanest tradeoff: best max pause among non-trivial growth factors, modest RSS, ~80% of the score win that 2.5× delivers.

### DIFF
--- a/Libraries/LibCrypto/BigInt/SignedBigInteger.h
+++ b/Libraries/LibCrypto/BigInt/SignedBigInteger.h
@@ -59,6 +59,7 @@ public:
     void set_to(SignedBigInteger const& other);
 
     [[nodiscard]] size_t byte_length() const;
+    [[nodiscard]] size_t external_memory_size() const { return m_mp.alloc * sizeof(mp_digit); }
 
     [[nodiscard]] SignedBigInteger plus(SignedBigInteger const& other) const;
     [[nodiscard]] SignedBigInteger minus(SignedBigInteger const& other) const;
--- a/Libraries/LibCrypto/BigInt/UnsignedBigInteger.h
+++ b/Libraries/LibCrypto/BigInt/UnsignedBigInteger.h
@@ -69,6 +69,7 @@ public:
     [[nodiscard]] bool is_odd() const;
 
     [[nodiscard]] size_t byte_length() const;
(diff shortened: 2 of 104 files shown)


Jacobdm0:
Pretty cool update, seems like a good sized increase in speed

PR #9288, opened 2026-05-06
LibMedia: Give the media pipeline defined roles for producers/processors/sinks

Zaggy1024:
tl;dr: The main purpose of this change is to split up the process of mixing multi-track audio and the actual audio playback. Audio is mixed/processed into a queue by a thread in AudioPlaybackSink so that the data callback doesn't block.

## The Big Changes

All parts of the pipeline dealing with raw media data (decoded video frames, float PCM audio) are now classified as producers, sinks or processors. Producers can be connected as inputs to sinks. Processors are both producers and sinks, and so can have an input connected to them, and have their output connected to a sink as input. This allows the entire pipeline to become modular so that any extra processing we need can be inserted at any stage, particularly before/after mixing of audio.

Speaking of mixing, that is now handled by a processor called AudioMixer. It is currently the only processor that can accept multiple inputs, and it mixes the connected inputs into a singular output stream of AudioBlocks. Later, instead of connecting directly to the AudioPlaybackSink, this will pass through a time scale modification processor for playback rate control.

Previously, SeekingStateHandler initiated a seek at the decoders and coordinated pausing/resuming of the sinks based on the state of those seeks. Now, the seek is initiated at the sink, and propagates up the pipeline to the initial producers. The pipeline then processes data up to the sink, which then notifies the PlaybackManager through a callback that new data is available. Each node in the pipeline ensures that data from prior to the initiated seek is discarded, so that the seek can be propagated synchronously all the way up to the initial producers to make sure that latency is still minimal.

Due to this change to the seeking, fast seeking must determine a timestamp to jump to _before_ initiating the seek, rather than after it has resolved. Thus, the demuxers now have a function to find the timestamp of random access point to jump to when fast seeking. Doing it this way allows the audio seeks to begin immediately instead of waiting for the video seek to complete, reducing latency for fast seeks.

However, FFmpeg's seek index uses the decode timestamps instead of presentation timestamps, and the decode timestamp of an H.264 keyframe can be less than its presentation timestamp. The side effect of this is that seeking to a timestamp that falls within the duration of a keyframe can result in an output frame with a timestamp later than the seek target, so the DisplayingVideoSink ends up with a nonexistent current frame. When using their seek index to look up the timestamp to jump to in this new fast seeking setup, we're _guaranteed_ to hit this edge case, so FFmpegDemuxer just defaults to the original seek timestamp instead of consulting the index. I don't believe that FFmpeg exposes PTS except by requesting a packet back, which requires both reading the underlying media data (we may still be waiting for it over the network), and copying the coded data out of that media data unnecessarily. My intention in the future is to implement an MP4 reader for MSE, then expand that to support plain old files later in order to enable fast seeking.


Zaggy1024:
Currently draft as I need to add some new seeking ref tests, then update expectations over the course of the branch history.

Libraries/LibMedia/Audio/PlaybackStream.h
```diff
@@ -65,6 +65,10 @@ class MEDIA_API PlaybackStream : public AtomicRefCounted<PlaybackStream> {
     // as soon as possible instead of waiting for remaining audio to play.
     virtual NonnullRefPtr<Core::ThreadedPromise<void>> discard_buffer_and_suspend() = 0;
 
+    // Notifies the stream that the data request callback may be able to provide data now. This is used to
+    // wake playback streams that have stopped requesting data due to an underrun.
+    virtual void notify_data_available() = 0;
```
R-Goc, line 70: I don't actually see anything calling this. Also I don't think we want to actually be calling this all the time. In WASAPI this would lead to unnecessary wakeups (we would be filling a buffer that is nearly full for example). Maybe we could use the underrun_callback to notify that notfiy_data_available will need to be called? We would just need to make sure it is plumbed correctly everywhere as it isn't used currently iirc.

R-Goc (changes requested):

Libraries/LibMedia/Audio/PlaybackStream.h
Zaggy1024: It's called from AudioPlaybackSink's processor thread.

Regarding unnecessary wakeups, I'd prefer to keep this function cheap for all backends. Using the underrun callback isn't great, since it would either be invoked from a thread that would be platform-dependent (could be the same thread data is requested from, but also could be some other thread running independently), or get invoked on the main thread and add latency. The AudioUnit and PulseAudio backends both are set up so that notify_data_available() is cheap in cases where a wakeup is not needed, so if calling SetEvent() unconditionally is expensive, we could just mirror that.

PR #9291, opened 2026-05-07
LibMedia: Fix wrong comparison to check if chunk covers new one

antoniospg:
* Fix wrong comparison between a size and an absolute offset


Zaggy1024:
Looks correct, but could we add a test for this to TestIncrementallyPopulatedStream? (Make sure that it fails before the fix so that we can be sure it's an effective regression test!)

Zaggy1024 (approved):
Thank you!

PR #9294, opened 2026-05-07
UI/Qt: Copy window size and maximized state for new windows

AtkinsSJ:
Create new windows with the size of the previously active window, and maximized if it was maximized. This matches the behaviour of other browsers, and solves an annoying issue where new windows would always be tiny.

### DIFF
--- a/UI/Qt/BrowserWindow.cpp
+++ b/UI/Qt/BrowserWindow.cpp
@@ -326,7 +326,13 @@ BrowserWindow::BrowserWindow(Vector<URL::URL> const& initial_urls, IsPopupWindow
         tab.focus_location_editor();
     });
     QObject::connect(m_new_window_action, &QAction::triggered, this, [] {
-        (void)Application::the().new_window({ WebView::Application::settings().new_tab_page_url() });
+        auto const& previous_active_window = Application::the().active_window();
+        WindowConfiguration configuration {
+            .width = previous_active_window.width(),
+            .height = previous_active_window.height(),
+            .maximized = previous_active_window.isMaximized(),
+        };
+        Application::the().new_window({ WebView::Application::settings().new_tab_page_url() }, configuration);
     });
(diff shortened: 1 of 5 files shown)


UI/Qt/BrowserWindow.cpp
```diff
@@ -325,7 +325,11 @@ BrowserWindow::BrowserWindow(Vector<URL::URL> const& initial_urls, IsPopupWindow
         tab.focus_location_editor();
     });
     QObject::connect(m_new_window_action, &QAction::triggered, this, [] {
-        (void)Application::the().new_window({ WebView::Application::settings().new_tab_page_url() });
+        auto const& previous_active_window = Application::the().active_window();
+        auto& new_window = Application::the().new_window({ WebView::Application::settings().new_tab_page_url() });
```
trflynn89: Hmm so this feels like a partial fix in that `ladybird --new-window` also creates a tiny window.

Maybe `Application::new_window` should take an optional rect, and invoke `BrowserWindow::set_window_rect` itself (where, if `none`, it uses the current window's rect).

AtkinsSJ: It got quite complicated. :sweat_smile: The initial window doesn't have a previous window to copy. And popup windows can specify position and size, but do so as x/y/w/h, each of which is optional, which is a bit of a pain.

What I've now done is make `new_window()` take a `WindowConfiguration` with optional x,y,w,h and maximized. The two places that wanted to copy the previous window pass that in as their configuration. The initial window also does so for its restored state, and popups pass their hint values.

...as I now specify the configuration for the initial window, I could actually make copying be the default again. But passing an `Optional<WindowConfiguration>`, where each of `WindowConfiguration`'s fields is also optional, feels a bit gross.

trflynn89: Did you mean to squash those commits?

AtkinsSJ: :man_facepalming: Yes

PR #9296, opened 2026-05-07
LibWeb: Drop display list ellipse commands

kalenikaliaksandr:
Use path painting commands instead. Reduces display list commands set.

### DIFF
--- a/Libraries/LibWeb/Painting/DisplayList.cpp
+++ b/Libraries/LibWeb/Painting/DisplayList.cpp
@@ -202,8 +202,6 @@ void DisplayListPlayer::execute_impl(DisplayList& display_list, ScrollStateSnaps
         else HANDLE_COMMAND(FillRectWithRoundedCorners, fill_rect_with_rounded_corners)
         else HANDLE_COMMAND(FillPath, fill_path)
         else HANDLE_COMMAND(StrokePath, stroke_path)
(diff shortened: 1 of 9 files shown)


Libraries/LibWeb/Painting/DisplayList.cpp
```diff
@@ -202,8 +202,6 @@ void DisplayListPlayer::execute_impl(DisplayList& display_list, ScrollStateSnaps
         else HANDLE_COMMAND(FillRectWithRoundedCorners, fill_rect_with_rounded_corners)
         else HANDLE_COMMAND(FillPath, fill_path)
         else HANDLE_COMMAND(StrokePath, stroke_path)
-        else HANDLE_COMMAND(DrawEllipse, draw_ellipse)
-        else HANDLE_COMMAND(FillEllipse, fill_ellipse)
```
gmta, line 206: Why do we need to remove DisplayList commands?

I'm asking because it looks like Skia has dedicated oval drawing code for its Ganesh/Graphite backends, while a path-based approach seems likely to have worse performance? (Need to confirm obviously)

The CPU backend does fall back to drawing an oval path, so things should be identical there.

kalenikaliaksandr: I am working on making display list serializable and it seemed like those are obviously not necessary and I am not interested at slightly faster ellipse painting if it comes at cost of having 2 additional display list commands.

PR #9301, opened 2026-05-07
Fix/freebsd build prereqs

cardozQ:
Fixes a couple of issues in the FreeBSD build instructions.
Added the missing `ccache` and `tar` packages to the `pkg install` command. Also added a note about the vcpkg bootstrap step failing with a Python `CalledProcessError` traceback when `zip`, `unzip`, or `tar` are missing, since the actual error can be easy to miss in the output.

//Tested on FreeBSD 15.0 by running:
 ./Meta/ladybird.py run
 
 Thankyou,
cardoz

### DIFF
--- a/Documentation/BuildInstructionsLadybird.md
+++ b/Documentation/BuildInstructionsLadybird.md
@@ -211,8 +211,11 @@ Or, download a version of Gradle >= 8.0.0, and run the ``gradlew`` program in ``
 ### FreeBSD
 
 ```
-pkg install autoconf-archive automake autoconf bash cmake curl gmake gn libdrm libtool libxcb libxkbcommon libX11 libXrender libXi nasm ninja patchelf pkgconf python3 qt6-base unzip zip
+pkg install autoconf-archive automake autoconf bash cmake ccache curl gmake gn libdrm libtool libxcb libxkbcommon libX11 libXrender libXi nasm ninja patchelf pkgconf python3 qt6-base tar unzip zip
 ```
+> [!NOTE]
+> `zip`, `unzip`, and `tar` are required by the vcpkg bootstrap step. If any of these are missing,
+> the build will fail with a Python `CalledProcessError` traceback rather than a clear error message.
 
 ## Build steps
 


Documentation/BuildInstructionsLadybird.md
```diff
@@ -211,8 +211,10 @@ Or, download a version of Gradle >= 8.0.0, and run the ``gradlew`` program in ``
 ### FreeBSD
 
 ```
-pkg install autoconf-archive automake autoconf bash cmake curl gmake gn libdrm libtool libxcb libxkbcommon libX11 libXrender libXi nasm ninja patchelf pkgconf python3 qt6-base unzip zip
+pkg install autoconf-archive automake autoconf bash cmake ccache curl gmake gn libdrm libtool libxcb libxkbcommon libX11 libXrender libXi nasm ninja patchelf pkgconf python3 qt6-base tar unzip zip
 ```
+> **Note:** `zip`, `unzip`, and `tar` are required by the vcpkg bootstrap step. If any of these are missing,
+> the build will fail with a Python `CalledProcessError` traceback rather than a clear error message.
```
cqundefine: We should probably figure out why it gives an error like this and patch whatever is causing that to give a proper error message, if you don't wanna look at it I can take a look.

cardozQ: yeah i figured it out and also raised a issue (https://github.com/microsoft/vcpkg/issues/51589) .

Documentation/BuildInstructionsLadybird.md
```diff
@@ -213,6 +213,8 @@ Or, download a version of Gradle >= 8.0.0, and run the ``gradlew`` program in ``
 ```
 pkg install autoconf-archive automake autoconf bash cmake ccache curl gmake gn libdrm libtool libxcb libxkbcommon libX11 libXrender libXi nasm ninja patchelf pkgconf python3 qt6-base tar unzip zip
 ```
+> **Note:** `zip`, `unzip`, and `tar` are required by the vcpkg bootstrap step. If any of these are missing,
+> the build will fail with a Python `CalledProcessError` traceback rather than a clear error message.
```
AtkinsSJ: You can make this appear as a fancy note section - see lines 191 and 192 above.

AtkinsSJ (changes requested):
Can you combine the two commits together?

Documentation/BuildInstructionsLadybird.md
cardozQ: Done . quashed the commits and updated the note to use the [!NOTE] admonition format.

AtkinsSJ (approved):
Thanks, looks good! Welcome to Ladybird. :^)

CI can be a bit flaky, so if it fails, let someone on Discord know and they can restart the flaky jobs.

PR #9302, opened 2026-05-07
LibTextCodec: Fix ISO-2022-JP encoder escape sequence on unencodable error

Martin005:
This PR updates the ISO-2022-JP encoder implementation and adds the [related Web Platform Test](https://wpt.fyi/results/encoding/legacy-mb-japanese/iso-2022-jp/iso2022jp-encode-form-errors-stateful.html) (WPT) to improve conformance and test coverage.

The only code change is the fix of a copy-paste typo in the encoder to use the correct escape sequence for switching to ASCII mode.

By the way, this fixes the last failing test in [`iso-2022-jp`](https://wpt.fyi/results/encoding/legacy-mb-japanese/iso-2022-jp?product=ladybird) and therefore in [`legacy-mb-japanese`](https://wpt.fyi/results/encoding/legacy-mb-japanese?product=ladybird) coming from [`encoding`](https://wpt.fyi/results/encoding?product=ladybird) 🥳 

image

### DIFF
--- a/Libraries/LibTextCodec/Encoder.cpp
+++ b/Libraries/LibTextCodec/Encoder.cpp
@@ -281,7 +281,7 @@ ErrorOr<ISO2022JPEncoder::State> ISO2022JPEncoder::process_item(u32 item, State 
         if (state == State::jis0208) {
             TRY(on_byte(0x1B));
             TRY(on_byte(0x28));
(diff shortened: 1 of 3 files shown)


gmta (approved):
Nice find!

PR #9303, opened 2026-05-07
LibGC: Add a Variant `visit` overload and add tests for Visitor

Lubrsi:
This is part 2 of a rough 18 part series. See part 1: https://github.com/LadybirdBrowser/ladybird/pull/9282

This adds test for Visitor, then a Variant helper overload and uses it where appropriate.

### DIFF
--- a/Libraries/LibGC/Cell.h
+++ b/Libraries/LibGC/Cell.h
@@ -193,6 +193,16 @@ public:
 
         void visit(NanBoxedValue const& value);
 
+        template<typename... Ts>
+        void visit(Variant<Ts...> const& variant)
+        requires((IsVisitable<Ts>::value || ...))
+        {
+            variant.visit([&](auto const& value) {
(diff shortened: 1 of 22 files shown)


Libraries/LibGC/Cell.h
```diff
@@ -100,8 +100,10 @@ class GC_API Cell {
         template<typename T>
         void visit(ReadonlySpan<T> span)
         {
-            for (auto& value : span)
-                visit(value);
+            if constexpr (requires(T const& value) { visit(value); }) {
+                for (auto& value : span)
+                    visit(value);
+            }
```
gmta: Hm, I'd rather want this to be a compilation error if I accidentally call `visit()` on a span with unsupported types - seems like a footgun!

Lubrsi: I took it back to drawing board to use a trait, let me know what you think.

PR #9306, opened 2026-05-07
Devcontainer+Meta: Fix the vcpkg binary asset cache so that it is actually used

slydetector:
During the build of the devcontainer, vcpkg dependencies are built and stored in `/usr/local/share/vcpkg-binary-cache`. 

When a user fires up the devcontainer and invokes `./Meta/ladybird.py build`, there are no hits in the binary asset cache because the vcpkg compiler hashes used at devcontainer build time are different from those used post-build.  It turns out the devcontainer build step was using GCC and post-build used CLANG. 

To make these environments align, I've updated the `install.sh` invoke `./Meta/ladybird.py configure` so both paths use the same machinery to detect and configure the build environment. This required adding support for a `configure` target to `ladybird.py`.

The order of the caches in `VCPKG_BINARY_SOURCES` also required updating so the `/usr/local/share/...` cache is consulted in a read-only capacity before the user specific read-write cache.

Testing done:
1. Rebuilt the devcontainer without a cache in vs-code.
2. Once in the devcontainer, ran `./Meta/ladybird.py build`.
3. Observed that all the cached vcpkg binary dependencies were successfuly found and used.
4. The build of ladybird started promptly thereafter and succeeded with no errors.
5. `./Meta/ladybird.py run` brought up the browser window successfully.

I'm guessing this should fix the "Optimized" devcontainer's lack of using the cache, but I can't confirm that until this is merged to `master` and gone through a full push cycle to the container registry.

I'm unfamiliar with CMake, vcpkg, C/C++ toolchains, etc, so please feel free to point out anything I may have overlooked.


.devcontainer/features/vcpkg-cache/install.sh
slydetector, line 27: For testing, this was pointing to my fork's branch to get `ladybird.py` with the `configure` added.

gmta:

PR #9312, opened 2026-05-08
LibWeb: Use CSS geometry properties in SVGEllipseElement

endafk:
`SVGCircleElement` reads its `cx`, `cy`, and `r` values from the CSS cascade via `computed_values()`, and delegates its IDL accessors to `svg_animated_length_for_property()`. `SVGEllipseElement` was still on the old approach: parsing attribute strings in `attribute_changed()`, keeping the results in four private `Optional<float>` members, and building `SVGAnimatedLength` objects by hand in the accessors.

This PR brings `SVGEllipseElement` in line with the circle.

**Changes:**

- Drop `attribute_changed` and the four private member variables (`m_center_x`, `m_center_y`, `m_radius_x`, `m_radius_y`). `SVGElement::apply_presentational_hints` already maps `cx`/`cy`/`rx`/`ry` to the CSS cascade for `<ellipse>` (they're listed in `attribute_style_properties()`), so no extra plumbing is needed in the subclass.
- Read `cx`, `cy`, `rx`, and `ry` from `computed_values()` in `get_path()`. Implements the SVG 2 `auto` rule for `rx`/`ry`: if both are `auto`, rendering is disabled; if only one is `auto`, it mirrors the other.
- Replace the hand-rolled `SVGAnimatedLength` construction in `cx()`, `cy()`, `rx()`, and `ry()` with `svg_animated_length_for_property()`, removing the `FIXME` comments that went with it.
- Update spec URL references from SVG 1.1 to SVG 2.

**Bug fix included:** `ComputedValues::rx()` had a copy-paste typo — it was returning `m_noninherited.ry` instead of `m_noninherited.rx`. Nothing was calling `computed_values().rx()` before this change, so the bug was latent until now.

### DIFF
--- a/Libraries/LibWeb/CSS/ComputedValues.h
+++ b/Libraries/LibWeb/CSS/ComputedValues.h
@@ -719,7 +719,7 @@ public:
     LengthPercentage const& cx() const { return m_noninherited.cx; }
     LengthPercentage const& cy() const { return m_noninherited.cy; }
     LengthPercentage const& r() const { return m_noninherited.r; }
-    LengthPercentageOrAuto const& rx() const { return m_noninherited.ry; }
+    LengthPercentageOrAuto const& rx() const { return m_noninherited.rx; }
     LengthPercentageOrAuto const& ry() const { return m_noninherited.ry; }
     LengthPercentage const& x() const { return m_noninherited.x; }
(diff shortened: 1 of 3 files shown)


gmta:

Issue #8343, opened 2026-03-10
RequestServer: Support TLS client certificate authentication


kwinsch:
## Summary

Ladybird has IPC infrastructure for client certificate authentication (`certificate_requested` / `set_certificate` in `RequestServer.ipc` and `RequestClient.ipc`), with callbacks wired through `LibRequests::Request` and `ResourceLoader`. However, the server-side implementation is non-functional: `ConnectionFromClient::set_certificate()` unconditionally calls `TODO()`, and the browser-side callback in `ResourceLoader.cpp` returns an empty `CertificateAndKey`.

It seems, that this infrastructure was originally built for the internal LibTLS implementation and has not been adapted for the cURL backend. I would like to make it functional and am opening this issue to discuss the design before writing any code.

## Problem: timing mismatch between LibTLS and cURL

The existing IPC flow was designed for LibTLS, which could pause a TLS handshake mid-flight:

1. LibTLS encounters the server's `CertificateRequest` message and pauses.
2. RequestServer sends `certificate_requested(request_id)` to the browser.
3. The browser responds with `set_certificate(request_id, cert, key)`.
4. LibTLS resumes the handshake with the certificate.

cURL does not support pausing the TLS handshake. `CURLOPT_SSLCERT` and `CURLOPT_SSLKEY` must be configured on the easy handle before it is added to the multi handle. In the current code, `Request::handle_fetch_state()` sets all cURL options and calls `curl_multi_add_handle()` synchronously. By the time a `set_certificate` IPC response could arrive from the browser, the handshake is already underway or complete.

## Proposed approach

Add a `RequestClientCertificate` state to the `Request` state machine, placed between `RetrieveCookie` and `Fetch`. This follows the same async IPC round-trip pattern that cookie retrieval already uses:

```
Init -> DNSLookup -> RetrieveCookie -> RequestClientCertificate -> Fetch
```

The flow would be:

1. `handle_request_client_certificate_state()` sends `certificate_requested(request_id)` to the browser via the primary connection.
2. The browser invokes its `on_certificate_requested` callback, checks for a matching certificate, and responds with `set_certificate()`.
3. `ConnectionFromClient::set_certificate()` stores the certificate and key on the `Request` object and transitions to `Fetch`.
[...]

PR #9321, opened 2026-05-08
LibCrypto: Add SecureStore for encrypted blob storage

kwinsch:
Adds reusable encrypted blob storage primitives to LibCrypto:

SecureByteBuffer is a ByteBuffer wrapper that securely zeros contents on destruction via AK::secure_zero(). Move-only to prevent accidental duplication of secrets. Zeroes moved-from inline storage after move.

KeyProvider is a pluggable key derivation interface. PasswordKeyProvider derives a 256-bit key using Argon2id v1.3 (RFC 9106) with OWASP 2024 recommended parameters.

SecureStore encrypts and decrypts blobs using AES-256-GCM with per-blob random nonces. Lock/unlock lifecycle keeps the derived key in memory only while unlocked. Validates 32-byte key size from the provider. Blob format: nonce(12) || ciphertext || tag(16). Callers handle persistence; SecureStore only transforms plaintext to/from blobs.

Intermediate sensitive buffers are wiped via ScopeGuard so cleanup runs even on error paths.

This is groundwork for browser-managed certificate storage (#8343) and not yet used by any active module. May also be used to implement other managers for storing secrets.

### DIFF
--- a/Libraries/LibCrypto/CMakeLists.txt
+++ b/Libraries/LibCrypto/CMakeLists.txt
@@ -16,6 +16,7 @@ set(SOURCES
     Curves/EdwardsCurve.cpp
     Curves/SECPxxxr1.cpp
     KeyProvider.cpp
+    SecureStore.cpp
     Hash/Argon2.cpp
     Hash/SHAKE.cpp
     Hash/BLAKE2b.cpp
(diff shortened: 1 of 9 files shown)


gmta:

PR #9325, opened 2026-05-08
AK: Fix UTF-16 StringBuilder corruption after non-ASCII clear/reuse

### LINKED ISSUE #9324: `&times;` entity in HTML before `<script>` causes JS parse failure
### Summary

When the HTML entity `&times;` (or the literal Unicode character `×`, U+00D7) appears in the document body before a `<script>` tag, the JavaScript engine receives corrupted script content and fails to parse. The script never executes.

`about:settings` uses `&times;` in identical `<dialog>` structures and works. The difference between the pages that causes one to work and the other to fail was not identified.

### Additional observations

- `&times;` inside any element (`<button>`, `<dialog>`, `<p>`) triggers the bug.
- The literal UTF-8 character `×` (U+00D7, bytes `0xC3 0x97`) also triggers it.
- The `<script>` content itself is valid JS (confirmed with `node --check`).
- The file is clean ASCII/UTF-8 with no BOM or hidden characters (verified with `xxd`).
- The script content passes through `Utf16String::from_utf8()` in `RustIntegration::compile_script()` (Libraries/LibJS/RustIntegration.cpp:481) before reaching the Rust lexer. A non-ASCII character earlier in the document may corrupt either the HTML parser's script text extraction, the UTF-8 to UTF-16 conversion, or a line/column offset calculation that assumes ASCII.

### Environment

- Ladybird built from source at d948320e32 on master
- Arch Linux, Qt 6.11.0, system GCC/Rust
- Rust-based JS parser (LibJS/Rust)

### Operating system

Linux

### Steps to reproduce

1. Create or modify an `about:` page with the following HTML:

```html
<!doctype html>
<html>
<head><title>Test</title></head>
<body>
    <button>&times;</button>
    <script>
        console.log("hello");
    </script>
</body>
</html>
```

2. Navigate to the page in Ladybird.

### Expected behavior

`hello` is printed to the console. The script executes normally.

### Actual behavior

The script fails to parse:

```
Unhandled JavaScript exception: [SyntaxError] Unexpected token Invalid.
Expected statement or declaration (line: N, column: 1)
```

The error occurs at line 1, column 1 of the script content, suggesting the source text fed to the Rust JS parser is garbled. Replacing `&times;` with `X` makes the script execute normally.

N/A

```html
N/A
```

```shell
N/A
```

_No response_

_No response_

### Contribute a patch?

kwinsch:
Fix two related bugs in `StringBuilder` / `Utf16StringData` that caused the first code unit of a `Utf16String` to be corrupted to null after the builder processed a non-ASCII character.

- `StringBuilder::clear()` did not reset `m_utf16_builder_is_ascii`, so a reused builder stored subsequent ASCII as char16_t and hit the placement-new overlap in `from_string_builder()`
- `Utf16StringData` has `sizeof=32` but `offsetof(m_ascii_data)=30`, leaving 2 bytes of tail padding that overlap string storage; placement-new construction could zero those bytes

The HTML parser reuses a single `m_character_insertion_builder` across all text nodes. When a non-ASCII character (e.g. `&times;`) appeared before a `<script>` tag, the script text node's first code unit was zeroed, causing a JS parse failure.

Closes #9324

### DIFF
--- a/AK/StringBuilder.cpp
+++ b/AK/StringBuilder.cpp
@@ -362,6 +362,8 @@ Utf16View StringBuilder::utf16_string_view() const
 void StringBuilder::clear()
 {
     m_buffer.resize(string_builder_prefix_size(m_mode));
+    if (m_mode == Mode::UTF16)
+        m_utf16_builder_is_ascii = true;
 }
 
 ErrorOr<void> StringBuilder::try_append_code_point(u32 code_point)
--- a/Tests/AK/TestUtf16String.cpp
+++ b/Tests/AK/TestUtf16String.cpp
@@ -1326,3 +1326,29 @@ TEST_CASE(optional)
     EXPECT(!string.has_value());
     EXPECT_EQ(released, u"well 😀 hello"sv);
 }
+
+TEST_CASE(utf16_builder_clear_resets_ascii_flag)
+{
+    // Regression test: StringBuilder(UTF16) must reset m_utf16_builder_is_ascii
+    // on clear(). Without this, a builder that previously held non-ASCII content
+    // stores subsequent ASCII as char16_t, and to_utf16_string() corrupts the
+    // first code unit via placement-new overlap in from_string_builder().
+    StringBuilder builder(StringBuilder::Mode::UTF16);
+
+    // 1. Append a non-ASCII code point to force m_utf16_builder_is_ascii = false.
+    builder.append_code_point(0x00D7); // U+00D7 MULTIPLICATION SIGN (×)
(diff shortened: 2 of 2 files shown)


trflynn89:
The Utf16StringData issue is fixed in #9315

kwinsch:
Rebased on current master and dropped the `Utf16StringData` padding commit since #9315 already addresses that, and arguably more elegantly. The PR now contains only the `StringBuilder::clear()` fix that resets `m_utf16_builder_is_ascii`, plus a regression test.

PR #9339, opened 2026-05-10
Meta+LibWeb: Support more CSS grammar features for code generation

Calme1709:
This PR adds support for groups, optionals and juxtaposition and implements `<paint>` generation which uses these features.

No functional change.

### DIFF
--- a/Libraries/LibWeb/CSS/ComputedValues.h
+++ b/Libraries/LibWeb/CSS/ComputedValues.h
@@ -319,8 +319,10 @@ public:
         if (style_value->is_value_list()) {
             auto const& values = style_value->as_value_list().values();
 
-            if (values.size() == 1)
-                return { values[0]->as_url().url(), {} };
+            VERIFY(values.size() == 2);
+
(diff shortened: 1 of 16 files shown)


Calme1709:
Updated serialization of `StyleValueList` with `EmptyOptionalStyleValue` to avoid potential trailing whitespace

AtkinsSJ (approved):
I'll be honest, `EmptyOptionalStyleValue` weirds me out a little, but it does solve a lot of problems, so I'm pretty sure that's just a me thing. 😅

Calme1709:
I'm not 100% happy with that either, the alternative I toyed with was that productions (i.e. juxtapositions or `&&` combinators) with optionals would parse to a `TupleStyleValue` but that felt worse IMO.

More than happy to replace it with something else if you have ideas.

AtkinsSJ:
I really don't have ideas, I sat thinking about it a while and this seems like the best option. Simple, lightweight, keeps consistent indices, minimally invasive. Can't really argue with it!

PR #9341, opened 2026-05-10
Documentation: Fix missing Fedora build prerequisites

senonide:
Add `gcc-c++` to the base Fedora prerequisites. On a fresh Fedora toolbox only `gcc` is installed by default, which causes vcpkg's compiler detection to fail when building for any UI.

Add `libxkbcommon-devel` to the Fedora GTK UI prerequisites. The GTK vcpkg overlay port builds with the Wayland backend on Linux and requires xkbcommon from the system. This was already documented for Debian/Ubuntu but missing from the Fedora section.

### DIFF
--- a/Documentation/BuildInstructionsLadybird.md
+++ b/Documentation/BuildInstructionsLadybird.md
@@ -87,7 +87,7 @@ sudo pacman -S libpulse
 ### Fedora or derivatives:
 
 ```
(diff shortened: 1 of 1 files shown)


Documentation/BuildInstructionsLadybird.md
jorge-pais, line 1: On Fedora 44, CMake was throwing a warning about libpulse. Perhaps it makes sense to add `pulseaudio-libs` and `pulseaudio-libs-devel` to this description :)

gmta:

PR #9342, opened 2026-05-10
LibJS: Remove unused rust_compile_program function

kunlinglio:
Removed the function `rust_compile_program` function, since it's never used anymore.


trflynn89:
Follow up here: #9348

PR #9350, opened 2026-05-10
LibWeb: Invalidate img-element when loaded from list of available images

InvalidUsernameException:
As a followup to #9014, fix that many images did not become visible on https://bleepingcomputer.com/ until hovered. More details in the commit message.

There is currently another invalidation(?) bug on that site that causes the navigation menu to behave weirdly. That one seems like a recent regression that will need to be looked at separately.

Before:
before_1
before_2

After:
after_1
after_2


InvalidUsernameException:
> There is currently another invalidation(?) bug on that site that causes the navigation menu to behave weirdly

For posterity, this is tracked in #9392 by now.

PR #9353, opened 2026-05-10
Implement time buffered ranges

antoniospg:
* Implement `FFmpegDemuxer::buffered_time_ranges` that maps the AVindexEntry to the current downloaded chunk
* Implement `IncrementallyPopulatedStream::available_byte_ranges()` that returns all the current available downloaded chunk byte ranges
* Add tests for both implementations
* Matroska implementation is not done yet

### DIFF
--- a/Libraries/LibMedia/FFmpeg/FFmpegDemuxer.cpp
+++ b/Libraries/LibMedia/FFmpeg/FFmpegDemuxer.cpp
@@ -5,6 +5,7 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
+#include <AK/BinarySearch.h>
 #include <AK/Math.h>
 #include <AK/MemoryStream.h>
 #include <AK/Stream.h>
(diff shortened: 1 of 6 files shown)


Zaggy1024:
Sorry to say that this is also something I had been working on [here](https://github.com/Zaggy1024/ladybird/tree/buffered-ranges). Using the index works nicely enough for MP4 (though the index uses DTS instead of PTS, so it's not totally accurate with H.264), but it's not so good for many other formats that we support through FFmpegDemuxer. To address that, I created an abstract class that it delegates the buffered range calculation to with implementations for index lookup, CBR formats, and FLAC).

The branch is still a WIP simply because MP3 provides no way to get accurate timestamps when seeking unless it has a seek table, and unfortunately, there's no requirement that variable-bitrate MP3 files provide a seek table. I was in the process of thinking about ways to work around that to still provide a cohesive user experience, but needed a break from it. 😅 If you have any ideas on that, or just would like to provide feedback on my branch, that'd be welcome, though I do need to record some demo media before I can open a draft PR.

Out of curiosity, do you have Discord? If you have a particular interest in working on media stuff, it'd be cool to be able to talk about what you want to work on there!

antoniospg:
Hey @Zaggy1024 , that's alright, at least I learned some new stuff. I'm still ramping up on the ladybird codebase and also don't know much about codecs yet, but I'm sure gonna have a look on your branch.
I believe I should ask on discord before doing this type of changes hehe I thought it was starightforward to implement this, but it looks like the devil lives in the details

PR #9354, opened 2026-05-10
LibWeb: Map embedded content align hints to vertical-align

shlyakpavel:
> When an embed, iframe, img, or object element, or an input element whose type attribute is in the Image Button state, has an align attribute whose value is an ASCII case-insensitive match for the string "center" or the string "middle", the user agent is expected to act as if the element's 'vertical-align' property was set to a value that aligns the vertical middle of the element with the parent element's baseline.

https://html.spec.whatwg.org/multipage/rendering.html#attributes-for-embedded-content-and-images

Previously, it would just crash the browser

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLEmbedElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLEmbedElement.cpp
@@ -47,10 +47,15 @@ void HTMLEmbedElement::apply_presentational_hints(Vector<CSS::StyleProperty>& pr
     Base::apply_presentational_hints(properties);
     for_each_attribute([&](auto& name, auto& value) {
         if (name == HTML::AttributeNames::align) {
-            if (value.equals_ignoring_ascii_case("center"sv))
-                properties.append({ .property_id = CSS::PropertyID::TextAlign, .value = CSS::KeywordStyleValue::create(CSS::Keyword::Center) });
-            else if (value.equals_ignoring_ascii_case("middle"sv))
-                properties.append({ .property_id = CSS::PropertyID::TextAlign, .value = CSS::KeywordStyleValue::create(CSS::Keyword::Middle) });
+            // https://html.spec.whatwg.org/multipage/rendering.html#attributes-for-embedded-content-and-images
+            // When an embed, iframe, img, or object element, or an input element whose type attribute is in the Image Button state,
+            // has an align attribute whose value is an ASCII case-insensitive match for the string "center" or the string "middle",
+            // the user agent is expected to act as if the element's 'vertical-align' property was set to a value that aligns the
+            // vertical middle of the element with the parent element's baseline.
+            // FIXME: This should use legacy baseline-middle alignment instead of CSS vertical-align: middle,
+            //        as Firefox and Chrome do with engine-specific legacy values.
+            if (value.equals_ignoring_ascii_case("center"sv) || value.equals_ignoring_ascii_case("middle"sv))
+                properties.append({ .property_id = CSS::PropertyID::VerticalAlign, .value = CSS::KeywordStyleValue::create(CSS::Keyword::Middle) });
         } else if (name == HTML::AttributeNames::height) {
             if (auto parsed_value = parse_dimension_value(value))
                 properties.append({ .property_id = CSS::PropertyID::Height, .value = *parsed_value });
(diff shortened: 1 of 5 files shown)


Libraries/LibWeb/HTML/HTMLObjectElement.cpp
```diff
@@ -127,9 +127,9 @@ void HTMLObjectElement::apply_presentational_hints(Vector<CSS::StyleProperty>& p
     for_each_attribute([&](auto& name, auto& value) {
         if (name == HTML::AttributeNames::align) {
             if (value.equals_ignoring_ascii_case("center"sv))
-                properties.append({ .property_id = CSS::PropertyID::TextAlign, .value = CSS::KeywordStyleValue::create(CSS::Keyword::Center) });
+                properties.append({ .property_id = CSS::PropertyID::VerticalAlign, .value = CSS::KeywordStyleValue::create(CSS::Keyword::Middle) });
             else if (value.equals_ignoring_ascii_case("middle"sv))
-                properties.append({ .property_id = CSS::PropertyID::TextAlign, .value = CSS::KeywordStyleValue::create(CSS::Keyword::Middle) });
+                properties.append({ .property_id = CSS::PropertyID::VerticalAlign, .value = CSS::KeywordStyleValue::create(CSS::Keyword::Middle) });
```
gmta, line 137: You could include the spec reference in these files :)

Libraries/LibWeb/HTML/HTMLEmbedElement.cpp
```diff
@@ -48,9 +48,9 @@ void HTMLEmbedElement::apply_presentational_hints(Vector<CSS::StyleProperty>& pr
     for_each_attribute([&](auto& name, auto& value) {
         if (name == HTML::AttributeNames::align) {
             if (value.equals_ignoring_ascii_case("center"sv))
-                properties.append({ .property_id = CSS::PropertyID::TextAlign, .value = CSS::KeywordStyleValue::create(CSS::Keyword::Center) });
+                properties.append({ .property_id = CSS::PropertyID::VerticalAlign, .value = CSS::KeywordStyleValue::create(CSS::Keyword::Middle) });
             else if (value.equals_ignoring_ascii_case("middle"sv))
-                properties.append({ .property_id = CSS::PropertyID::TextAlign, .value = CSS::KeywordStyleValue::create(CSS::Keyword::Middle) });
+                properties.append({ .property_id = CSS::PropertyID::VerticalAlign, .value = CSS::KeywordStyleValue::create(CSS::Keyword::Middle) });
```
gmta, line 58: Since the effect is now identical, can't you just do

```cpp
if (value.equals_ignoring_ascii_case("center"sv) || value.equals_ignoring_ascii_case("middle"sv))
    ...
```

?

shlyakpavel: I can, I just wanted to keep the changes in atomic commits. I will change that if you like

gmta: Yeah - let's combine the changes, makes for a neater diff imo :)

Libraries/LibWeb/HTML/HTMLObjectElement.cpp
shlyakpavel: Done

gmta:
Looking at the output of this test:

| Test | This PR | Chrome | Firefox |
|--|--|--|--|
| `object align=center` | `middle` | `middle` | `-moz-middle-with-baseline` |
| `embed align=center` | `middle` | `middle` | `-moz-middle-with-baseline` |
| `input image align=center` | `middle` | `middle` | `-moz-middle-with-baseline` |
| `object align=middle` | `middle` | `-webkit-baseline-middle` | `-moz-middle-with-baseline` |
| `embed align=middle` | `middle` | `-webkit-baseline-middle` | `-moz-middle-with-baseline` |
| `input image align=middle` | `middle` | `-webkit-baseline-middle` | `-moz-middle-with-baseline` |

Seems like there is a requirement for some baseline-related behavior in `align=middle` and `align=center` (at least, in Firefox). I think we should at least add a `FIXME` comment for this, or if you feel up to it, find out why Chrome and Firefox have specific style values for these!

shlyakpavel:
@gmta Great catch! It looks like CSS `vertical-align: middle` and the HTML spec’s “vertical middle of the element with the parent element’s baseline” are not quite the same thing. Firefox and Chromium seem to model this with engine-specific legacy values, and we don’t currently have an equivalent in LibWeb.

This PR keeps the current behavior scoped to `vertical-align: middle` and adds a FIXME for the baseline-specific legacy alignment. This PR already fixes some undefined behavior and a browser crash, but I'm happy to discuss if there's a better solution at the moment

gmta (approved):

gmta (approved):

PR #9357, opened 2026-05-11
LibCompress: Refactor CanonicalCode initialization

nandaanomi:
LibCompress/Deflate: Refactor CanonicalCode to simplify initialization

Refactor CanonicalCode implementation in Deflate.cpp by replacing magic
number 15 with constexpr MAX_HUFFMAN_BITS, replacing manual static bool
flag with C++11 thread-safe static initialization, optimizing prefix table
population loop to only iterate over valid entries using
number_of_prefix_codes, and adding validation to prevent read_symbol from
being called with zero code length.

No functional changes, only code quality and safety improvements.

### DIFF
--- a/Libraries/LibCompress/Deflate.cpp
+++ b/Libraries/LibCompress/Deflate.cpp
@@ -14,31 +14,17 @@
 
 namespace Compress {
 
+static constexpr u8 MAX_HUFFMAN_BITS = 15;
+
 CanonicalCode const& CanonicalCode::fixed_literal_codes()
 {
-    static CanonicalCode code;
-    static bool initialized = false;
-
-    if (initialized)
-        return code;
-
-    code = MUST(CanonicalCode::from_bytes(fixed_literal_bit_lengths));
-    initialized = true;
-
+    static CanonicalCode code = MUST(CanonicalCode::from_bytes(fixed_literal_bit_lengths));
     return code;
 }
 
 CanonicalCode const& CanonicalCode::fixed_distance_codes()
 {
-    static CanonicalCode code;
-    static bool initialized = false;
-
-    if (initialized)
-        return code;
-
-    code = MUST(CanonicalCode::from_bytes(fixed_distance_bit_lengths));
-    initialized = true;
-
+    static CanonicalCode code = MUST(CanonicalCode::from_bytes(fixed_distance_bit_lengths));
     return code;
 }
 
@@ -55,7 +41,7 @@ ErrorOr<CanonicalCode> CanonicalCode::from_bytes(ReadonlyBytes bytes)
         }
     }
 
-    if (non_zero_symbols == 1) { // special case - only 1 symbol
+    if (non_zero_symbols == 1) {
         code.m_prefix_table[0] = PrefixTableEntry { static_cast<u16>(last_non_zero), 1u };
         code.m_prefix_table[1] = code.m_prefix_table[0];
         code.m_max_prefixed_code_length = 1;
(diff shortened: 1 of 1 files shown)


Zaggy1024:
Seems reasonable, but this sounds like 3 different changes in one. Please make sure to make a commit for each granular change. Ideally, each commit summary describes a single change rather than a list, unless things are interconnected in a way that makes them difficult to separate.

(Also, editing the PR title doesn't fix the commit linting, we don't use squash merges. You'll need to amend the commit as the bot mentioned.)

Libraries/LibCompress/Deflate.cpp
```diff
@@ -55,7 +41,7 @@ ErrorOr<CanonicalCode> CanonicalCode::from_bytes(ReadonlyBytes bytes)
         }
     }
 
-    if (non_zero_symbols == 1) { // special case - only 1 symbol
+    if (non_zero_symbols == 1) {
```
Zaggy1024, line 44: Is there a reason why this comment was removed?

Libraries/LibCompress/Deflate.cpp
```diff
@@ -109,26 +95,27 @@ ErrorOr<CanonicalCode> CanonicalCode::from_bytes(ReadonlyBytes bytes)
                 TRY(code.m_bit_codes.try_resize(symbol + 1));
                 TRY(code.m_bit_code_lengths.try_resize(symbol + 1));
             }
-            code.m_bit_codes[symbol] = fast_reverse16(start_bit | next_code, code_length); // DEFLATE writes huffman encoded symbols as lsb-first
+            code.m_bit_codes[symbol] = fast_reverse16(start_bit | next_code, code_length);
```
Zaggy1024, line 98: Is there a reason why this comment was removed?

Libraries/LibCompress/Deflate.cpp
```diff
@@ -90,7 +76,7 @@ ErrorOr<CanonicalCode> CanonicalCode::from_bytes(ReadonlyBytes bytes)
             if (next_code > start_bit)
                 return Error::from_string_literal("Failed to decode code lengths");
 
-            if (code_length <= CanonicalCode::max_allowed_prefixed_code_length) {
+            if (code_length <= MAX_HUFFMAN_BITS) {
```
Zaggy1024, line 79: This looks like it's allowing a higher code_length than before, is this intentional?

Libraries/LibCompress/Deflate.cpp
```diff
@@ -109,26 +95,27 @@ ErrorOr<CanonicalCode> CanonicalCode::from_bytes(ReadonlyBytes bytes)
...
 
-    if (next_code != (1 << 15))
+    if (next_code != (1 << MAX_HUFFMAN_BITS))
         return Error::from_string_literal("Failed to decode code lengths");
 
-    for (auto [symbol_code, symbol_value, code_length] : prefix_codes) {
-        if (code_length == 0 || code_length > CanonicalCode::max_allowed_prefixed_code_length)
-            break;
+    for (size_t i = 0; i < number_of_prefix_codes; ++i) {
+        auto& entry = prefix_codes[i];
```
Zaggy1024, line 109: nit: this could probably still use structured bindings like before to keep the diff a little clearer.

(For example, GitHub doesn't see that the first if statement is largely the same, so it's difficult to tell that `CanonicalCode::max_allowed_prefixed_code_length` changed for `code.m_max_prefixed_code_length`.)

gmta:

PR #9360, opened 2026-05-11
LibJS: Skip numeric index parse for non-numeric property keys

tcl3:
This is safe because we're not trimming whitespace in this case, and leading signs are not valid for unsigned int types.

On a 15 second profile of https://ubereats.com initial load, this drops the time spent in `fast_float::from_chars_int_advanced<u32>()`, from 71ms to 1ms.

### DIFF
--- a/Libraries/LibJS/Runtime/PropertyKey.h
+++ b/Libraries/LibJS/Runtime/PropertyKey.h
@@ -7,6 +7,7 @@
 
 #pragma once
 
+#include <AK/CharacterTypes.h>
 #include <AK/Utf16FlyString.h>
 #include <LibJS/Runtime/Completion.h>
 #include <LibJS/Runtime/PrimitiveString.h>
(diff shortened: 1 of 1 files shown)


shannonbooth:
The benchmark measures time spent in fast float,  ut would it not be more accurate to have the measurement taken from PropertyKey? Probably not the case, but in theory thpse 70 ms could be moving up the caller

tcl3:
In this case, the only work moved up to the caller is one extra `is_ascii_digit()` check.

Instrumenting with `REPORT_TIME_EVERY` shows ~50ms less time spent in the PropertyKey constructor on ubereats load:

Before
```[Timing info for: "PropertyKey constructor"] current: 20ns       | average: 28ns       | total: 469.682ms  | calls: 16530000```
After:
```[Timing info for: "PropertyKey constructor"] current: 20ns       | average: 25ns       | total: 419.182ms  | calls: 16530000```

PR #9365, opened 2026-05-11
LibJS: Use `find_code_unit_offset()` for index finding in `StringPrototype`

tcl3:
This replaces manual substring compare loops with a faster algorithm and allows us to remove the SplitMatch AO, which is no longer referenced by the spec.

On a profile of https://ubereats.com, I get the following results (as a percentage of total samples):

string_index_of: 0.36% -> 0.13%
split: 0.98% -> 0.43%

Over the 20 seconds of the profile this saves ~150ms.

### DIFF
--- a/Libraries/LibJS/Runtime/StringPrototype.cpp
+++ b/Libraries/LibJS/Runtime/StringPrototype.cpp
@@ -49,21 +49,6 @@ static ThrowCompletionOr<GC::Ref<PrimitiveString>> primitive_string_from(VM& vm)
     return TRY(this_value.to_primitive_string(vm));
 }
 
-// 22.1.3.21.1 SplitMatch ( S, q, R ), https://tc39.es/ecma262/#sec-splitmatch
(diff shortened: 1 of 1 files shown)


Libraries/LibJS/Runtime/StringPrototype.cpp
```diff
@@ -1211,33 +1196,42 @@ JS_DEFINE_NATIVE_FUNCTION(StringPrototype::split)
         return array;
     }
 
+    auto string_view = string->utf16_string_view();
+
+    if (separator_length == 0) {
```
trflynn89: The spec has actually been updated to explicitly handle a separator length of 0, maybe we should just update to follow those steps here?

https://tc39.es/ecma262/#sec-string.prototype.split

tcl3: I've made `String.prototype.split` follow the current spec steps.

PR #9367, opened 2026-05-11
LibWeb: Add `scroll-margin` and `scroll-padding` support

dgolebiewski:
This PR adds support for `scroll-margin` and `scroll-padding` CSS properties.

Those properties only affect `scrollIntoView` at the moment, but this may count as prep work to maybe someday implementing scroll snap. Also, we pass a few more WPT tests.


Libraries/LibWeb/CSS/Properties.json
```diff
@@ -3844,6 +3844,101 @@
...
+    "logical-alias-for": {
+      "group": "scroll-margin",
+      "mapping": "block-start"
+    }
+  },
+  "scroll-margin-bottom": {
+    "animation-type": "by-computed-value",
+    "initial": "0",
+    "inherited": false,
+    "requires-computation": "never",
```
AtkinsSJ: I believe that these all need to be `cascaded-value`, like for the regular `margin-foo` and `padding-foo` properties.

AtkinsSJ (changes requested):

AtkinsSJ (approved):

PR #9373, opened 2026-05-12
LibWeb: Use cached hash for InvalidationSet equality

Gingeh:
Reduces time spent in the two hottest functions on a profile of https://perftest.netlify.app/stylebench/
`InvalidationSet::operator==`: 16% -> 3.6%
`Traits<InvalidationSet::Property::>hash`: 9.6% -> 0.57%

And the resulting runs/minute: 3.99 -> 5.73 (both +- 0.046)


awesomekling:
Nice idea :)

PR #9375, opened 2026-05-12
LibWeb: Implement `chardetng`-based encoding detection

Martin005:
This PR improves HTML character encoding detection in `LibWeb` by integrating the Rust-based [`chardetng`](https://crates.io/crates/chardetng) crate for encoding detection. It also refactors how encoding is determined for form actions and base elements, ensuring correct handling of character encodings and TLD hints throughout the HTML parsing process.

`chardetng` was selected as it is currently used in Firefox (added in https://github.com/mozilla-firefox/firefox/commit/b33e8e6eaf9ddd17b3f8dafcec5ce95ad0376f2b on December 12, 2019, updated to the `1.0.0` version in https://github.com/mozilla-firefox/firefox/commit/5f33bf94d0589bf1f26285c6cd6be5d07e2acb2c on April 2, 2026). See more information about `chardetng` in the [write-up](https://hsivonen.fi/chardetng/).
Chromium uses [Compact Encoding Detection (CED)](https://github.com/google/compact_enc_det/). By the way, the CED repo was archived on April 17, 2026, but I wasn't able to find any information about CED being deprecated in Chromium's codebase. CED was added to Chromium in [21d3369](https://source.chromium.org/chromium/chromium/src/+/21d3369fc5ab4d70732f4d83388ad2e144aaebcb) and later vendored in [2f40a85](https://source.chromium.org/chromium/_/chromium/external/github.com/google/compact_enc_det/+/2f40a850bcc5d6f7c1bfa02dbf42ad19d8220dc0).


Meta/import-wpt-test.py
```diff
@@ -359,7 +359,10 @@ def main():
     resource_path = "/".join(Path(url_to_import).parts[2::])
 
     with urlopen(url_to_import) as response:
-        page = response.read().decode("utf-8")
+        # Some WPT test files (e.g. `encoding-detection` tests) are intentionally stored in a
+        # legacy encoding rather than UTF-8. We only parse the ASCII tag structure here, so
+        # replacing undecodable bytes with U+FFFD is safe and avoids a UnicodeDecodeError.
+        page = response.read().decode("utf-8", errors="replace")
```
shannonbooth, line 370: Not that this will take a big dent out of this commit, but it would be good to split this into its own commit

Libraries/LibWeb/HTML/Parser/HTMLEncodingDetection.cpp
```diff
@@ -460,7 +461,26 @@ Optional<ByteString> run_bom_sniff(ReadonlyBytes input)
...
+    // chardetng uses this to improve detection accuracy for country-code TLDs (.jp, .ru, .cn, …).
+    // The label must satisfy chardetng's constraints: ASCII only, no dots, no uppercase letters.
+    auto host_string = url.host().has_value() ? url.serialized_host() : String {};
+    StringView host_view = host_string;
+    auto last_dot = host_view.find_last('.');
+    StringView tld_label = last_dot.has_value()
+        ? host_view.substring_view(*last_dot + 1)
+        : host_view;
+    for (char c : tld_label) {
+        if (!((c >= 'a' && c <= 'z') || (c >= '0' && c <= '9') || c == '-'))
```
shannonbooth: For what URL does all of this logic change something? To me it seems like this will only be identical to taking everything after the last dot of serialized_host

Libraries/LibWeb/HTML/Parser/HTMLEncodingDetection.cpp
```diff
@@ -499,12 +519,30 @@ ByteString run_encoding_sniffing_algorithm(DOM::Document& document, ReadonlyByte
     // 7. Otherwise, if the user agent has information on the likely encoding for this page, e.g. based on the encoding of the page when it was last visited, then return
     //    that encoding, with the confidence tentative.
 
-    // 8. FIXME: The user agent may attempt to autodetect the character encoding from applying frequency analysis or other algorithms to the data stream. Such algorithms
-    //    may use information about the resource other than the resource's contents, including the address of the resource. If autodetection succeeds in determining a
-    //    character encoding, and that encoding is a supported encoding, then return that encoding, with the confidence tentative. [UNIVCHARDET]
-    if (!Utf8View(StringView(input)).validate()) {
-        // FIXME: As soon as Locale is supported, this should sometimes return a different encoding based on the locale.
-        return "windows-1252";
+    // 8. The user agent may attempt to autodetect the character encoding from applying
```
shannonbooth: Please wrap at around 120 column limit

Libraries/LibWeb/HTML/Parser/HTMLEncodingDetection.cpp
```diff
@@ -460,7 +461,26 @@ Optional<ByteString> run_bom_sniff(ReadonlyBytes input)
...
+        : host_view;
+    for (char c : tld_label) {
+        if (!((c >= 'a' && c <= 'z') || (c >= '0' && c <= '9') || c == '-'))
+            return {};
+    }
+    return tld_label.is_empty() ? ByteString {} : ByteString { tld_label };
+}
+
+// https://html.spec.whatwg.org/multipage/parsing.html#determining-the-character-encoding
+ByteString run_encoding_sniffing_algorithm(DOM::Document& document, ReadonlyBytes input, Optional<MimeSniff::MimeType> maybe_mime_type, bool* used_frequency_analysis)
```
shannonbooth: It would be a bit nicer if we returned used_frequency_analysis instead of an in-out parameter here IMO

Libraries/LibWeb/HTML/Parser/IncrementalDocumentParser.cpp
```diff
@@ -31,6 +32,18 @@ IncrementalDocumentParser::IncrementalDocumentParser(GC::Ref<DOM::Document> docu
 {
 }
 
+void IncrementalDocumentParser::finalize()
+{
+    Base::finalize();
+    // The streaming detector is normally freed in process_end_of_body (either after confirming
```
shannonbooth: Instead of manually managing memory, it would be good to make use of C++'s amenities here and wrap this in smart pointers or similar.

Libraries/LibWeb/HTML/Parser/IncrementalDocumentParser.h
```diff
@@ -17,20 +18,34 @@
 #include <LibWeb/Forward.h>
 #include <LibWeb/MimeSniff/MimeType.h>
 
+// Forward-declare the Rust-side opaque streaming detector (generated by cbindgen into
```
shannonbooth: This is too verbose of a comment for something we do quite commonly, though forward declarations generally live in Forward.h header.

Libraries/LibWeb/Rust/build.rs
```diff
@@ -21,20 +21,64 @@ fn main() -> Result<(), Box<dyn Error>> {
         .map(PathBuf::from)
         .unwrap_or_else(|_| out_dir.clone());
 
-    cbindgen::generate(manifest_dir).map_or_else(
+    let base_config = cbindgen::Config::from_file(manifest_dir.join("cbindgen.toml"))?;
+
+    // CSS tokenizer header — namespace Web::CSS::Parser::FFI, CSS types only.
+    let mut css_config = base_config.clone();
```
shannonbooth, line 27: This type of thing getting shuffled around should be split into it's own commit so we try and keep the encoding changes as tight as possible

Tests/LibWeb/Layout/input/css/content-for-marker-in-list.html
```diff
@@ -1,4 +1,5 @@
 <!DOCTYPE html>
+<meta charset="utf-8">
```
shannonbooth, line 2: Please squash FIXUP commits into the commit which originally introduced the issue (no commit should introduce test regressions on its own)

shannonbooth (commented):
Not an extensive review by any means (will need to come back to this as there are quite a lot of changes to go through), but dropped an initial set of comments for some things I noticed skimming through

Meta/import-wpt-test.py
Martin005: Done: [`42173de` (this PR)](https://github.com/LadybirdBrowser/ladybird/pull/9375/commits/42173de41bfcfd9e89f62577521c7618f931d9c4)

Libraries/LibWeb/HTML/Parser/HTMLEncodingDetection.cpp
Martin005: This was more of a sanity check, but I removed it in [`1ae0d00` (this PR)](https://github.com/LadybirdBrowser/ladybird/pull/9375/changes/1ae0d00ec7785c235310331fa7ec6b52b90d8fb2#diff-d9d0800aa2f5bfb09519f5a8f4a91ed81ea662d72940635d8776a4c2a09155b6R464-R475)

Libraries/LibWeb/HTML/Parser/HTMLEncodingDetection.cpp
Martin005: Sure, makes sense! Done in [`1ae0d00` (this PR)](https://github.com/LadybirdBrowser/ladybird/pull/9375/changes/1ae0d00ec7785c235310331fa7ec6b52b90d8fb2#diff-f4b1fe848e64f0823a22029d97e5837e0750a2aca35d88ef699bf6104d66e875R29-R35)

Libraries/LibWeb/HTML/Parser/IncrementalDocumentParser.h
Martin005: Removed in [`b09e4f7407e36f7e9955fa160763a4d6bf0210ed..1ae0d00ec7785c235310331fa7ec6b52b90d8fb2`#diff-35e4563146](https://github.com/LadybirdBrowser/ladybird/compare/b09e4f7407e36f7e9955fa160763a4d6bf0210ed..1ae0d00ec7785c235310331fa7ec6b52b90d8fb2#diff-35e4563146db8ae22e274f08db652e18ce0d4106cb6412477903ce45089c2dda)L21-L29

Libraries/LibWeb/Rust/build.rs
Martin005: Done in [`efc3d2d` (this PR)](https://github.com/LadybirdBrowser/ladybird/pull/9375/commits/efc3d2d3429cb71c705256044bb873351a22e17a)

Libraries/LibWeb/HTML/Parser/IncrementalDocumentParser.cpp
```diff
@@ -150,6 +217,55 @@ void IncrementalDocumentParser::process_body_error(JS::Value)
     HTMLParser::the_end(m_document, m_parser);
 }
 
+void IncrementalDocumentParser::re_parse_with_encoding(ByteString const& new_encoding)
+{
+    // FIXME: This is a simplified form of the spec's "change the encoding while parsing"
+    // algorithm (https://html.spec.whatwg.org/multipage/parsing.html#change-the-encoding).
+    // It tears down the entire first-pass parse tree and re-parses from the raw bytes using
+    // the corrected encoding. Script execution side-effects from the first pass are not
```
shannonbooth: This seems... not great - shouldn't we be doing sniffing _before_ running the parser so that this is not a problem? I don't think this is something we can live with as a first pass. Linking to https://html.spec.whatwg.org/multipage/parsing.html#changing-the-encoding-while-parsing is also a bit confusing here since that algorithm should only be run based on contents of meta tag.

Libraries/LibWeb/HTML/Parser/HTMLEncodingDetection.cpp
```diff
@@ -460,13 +461,27 @@ Optional<ByteString> run_bom_sniff(ReadonlyBytes input)
 }
 
 // https://html.spec.whatwg.org/multipage/parsing.html#determining-the-character-encoding
-ByteString run_encoding_sniffing_algorithm(DOM::Document& document, ReadonlyBytes input, Optional<MimeSniff::MimeType> maybe_mime_type)
+ByteString extract_tld_hint(URL::URL const& url)
+{
+    // Extract the rightmost DNS label from the URL's host as a TLD hint for chardetng.
+    // chardetng uses this to improve detection accuracy for country-code TLDs (.jp, .ru, .cn, …).
+    auto host_string = url.host().has_value() ? url.serialized_host() : String {};
```
shannonbooth: url.serialized_host works whether or not url.host() is null. However, we probably should not be using that here anyway since this is not correct for a IP based host

Tests/LibWeb/Layout/input/unicode-password-input.html
```diff
@@ -1,2 +1,3 @@
 <!DOCTYPE html>
+<meta charset="utf-8">
```
shannonbooth, line 2: I did comment on this earlier to be an atomic diff, but having gone through the diff it would be best if adding meta tags were added to its own commit before the tests get broken by encoding detection (as I imagine it doesn't change results by itself?) to make the commits easier to run through.

Martin005: Sorry, I misunderstood your previous commit. I split these changes to [`947674a` (this PR)](https://github.com/LadybirdBrowser/ladybird/pull/9375/changes/947674a50512fc1c38223d1615da93c9da3a7350)

Libraries/LibWeb/HTML/Parser/HTMLEncodingDetection.cpp
Martin005: That makes sense, updated in [`d90e6a1` (this PR)](https://github.com/LadybirdBrowser/ladybird/pull/9375/changes/d90e6a17a09066d9b06cf343271c6d503a44cb5f#diff-d9d0800aa2f5bfb09519f5a8f4a91ed81ea662d72940635d8776a4c2a09155b6R467-R470)

Libraries/LibWeb/HTML/Parser/IncrementalDocumentParser.cpp
Martin005: The implementation was my attempt to cover the "-late" WPT encoding-detection tests (e.g. [ar-ISO-8859-6-late.tentative.html](https://wpt.fyi/results/encoding-detection/ar-ISO-8859-6-late.tentative.html?product=ladybird)), which are ~9984 bytes with the first non-ASCII byte at ~9951 - well past the `MAX_SNIFF_BYTES` threshold that is 1445.

So I decided to pivot to avoid solving these "-late" WPT tests and simplified the solution to focus on making the rest of encoding-detection WPT tests pass.

Here is the simplified solution: [`d90e6a1` (this PR)](https://github.com/LadybirdBrowser/ladybird/pull/9375/changes/d90e6a17a09066d9b06cf343271c6d503a44cb5f)

shannonbooth: "tentative" means that its not yet part of the standard most of the time - only Firefox seems to pass that test. Best to not focus on that test and only the initial sniff IMO

Tests/LibWeb/TestConfig.ini
```diff
@@ -19,6 +19,52 @@ Text/input/css/font-face-load-dedups-same-url.html
 ; pushState with path URL requires HTTP(s) scheme.
 Text/input/navigation/history-replace-push-then-back.html
 
+; Encoding-detection tests use absolute /resources/testharness.js paths that only resolve via HTTP.
```
shannonbooth: Is this the only reason why these tests need to run from the HTTP server? How were these tests imported? If using ./Meta/import-wpt-test.py the path should get fixed up to something like <script src="../../resources/testharness.js"></script> so that it works with file based HTML files as well.

Martin005: You are right. I see that [`f43b735` (this PR)](https://github.com/LadybirdBrowser/ladybird/pull/9375/commits/f43b7355fe916afd18752fc739202d37c513c4c5) wasn't enough to correctly parse and update the paths in the downloaded HTML tests from WPT. I updated `import-wpt-test.py` to handle this correctly in [`1081923` (this PR)](https://github.com/LadybirdBrowser/ladybird/pull/9375/changes/1081923fbc995ce4e91b65ae21dc41e46584b31f#diff-0a4e8841b3ab6e69ab4ceacd4a5d6303870aa05aed5ef9b37de1addd4ad5022eR225-R246) and reimported them in [`0ebeb0f` (this PR)](https://github.com/LadybirdBrowser/ladybird/pull/9375/changes/0ebeb0f8e4d1fa27c69ab4afb8b6611046491520).
This means that this PR no longer touches `Tests/LibWeb/TestConfig.ini`.

shannonbooth:
Unfortunately I think this needs a rebase on top of master to fix stealth test conflict for has-concrete-structural-triggers.html, otherwise LGTM

Martin005:
> Unfortunately I think this needs a rebase on top of master to fix stealth test conflict for has-concrete-structural-triggers.html, otherwise LGTM

@shannonbooth Good call, fixed - [`69fe96a` (this PR)](https://github.com/LadybirdBrowser/ladybird/pull/9375/changes/69fe96a56192713e222aa8b6bae804d6185e18dc#diff-4e7cc64dbd9f39938bdcc7b4d82743c783713e7f2af94eec4796be8b09cd6095)

PR #9380, opened 2026-05-12
LibWebView: Add a --screenshot-path option

tcl3:
This option saves screenshots to the given location.


tcl3:
My latest push adds a `--screenshot-path` option rather than `--downloads-path` option, which only changes the location for screenshots.

Calme1709 (approved):

PR #9384, opened 2026-05-13
AK: Constrain FixedPoint underlying type to integrals

creemy0cheesekake:
This PR addresses a FIXME in `AK/FixedPoint.h`:
`// FIXME: use the Integral concept to constrain Underlying`
for the FixedPoint class:
```cpp
template<size_t precision, typename Underlying>
class FixedPoint
```

`static_assert` was used instead of using the concept like `template<size_t precision, Integral Underlying>` in order to prevent the circular dependency errors caused by including `<AK/Concepts.h>`

Verified by manually confirming something like `FixedPoint<4, float>` does not compile.

### DIFF
--- a/AK/FixedPoint.h
+++ b/AK/FixedPoint.h
@@ -26,9 +26,9 @@
 namespace AK {
 
 // FIXME: this always uses round to nearest break-tie to even
-// FIXME: use the Integral concept to constrain Underlying
 template<size_t precision, typename Underlying>
 class FixedPoint {
+    static_assert(IsIntegral<Underlying>, "Underlying type must be integral");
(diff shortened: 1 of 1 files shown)


trflynn89:
I don't follow this comment:
> `static_assert` was used instead of using the concept like `template<size_t precision, Integral Underlying>` in order to prevent the circular dependency errors caused by including `<AK/Concepts.h>`

`FixedPoint.h` already includes `Concepts.h`. You should be able to just use that concept?

But in any case, it doesn't look like `FixedPoint` is actually used by anything in the project other than its own unit test. So instead of this, we could just remove it.

trflynn89:
Superseded by #9546

PR #9387, opened 2026-05-13
Meta+LibGfx: Run a Python type checker alongside the linter

Zaggy1024:
This will help us be more explicit about our Python typing, while also completely ensuring that we don't use features that are unsupported in Python 3.9. For example, it catches usage of the union operator in `css_dimensions.py` which passed CI's current Python version checking and breaks builds on macOS.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -66,7 +66,7 @@ runs:
         if ${{ inputs.type == 'build' }} ; then
           pip3 install pyyaml requests six
         else
-          pip3 install ruff
+          pip3 install pyright ruff
         fi
 
     - name: Ensure clang-cl is available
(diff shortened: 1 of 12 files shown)


AtkinsSJ:
I gave this a quick look but I really don't know Python well enough to review it. I'm glad to see this change though, because I thought CI was already checking this, hence letting a couple of things through that I thought 3.9 didn't like, but CI was happy so I thought I was just wrong. :sweat_smile:

Zaggy1024:
I was also hoping the linter was enough, I wanna say I did see it catch a union operator in my testing. Not sure why this one slips through. 😅 

Regarding the Python typing changes, I mainly did my best to avoid abusing Any everywhere, but the JSON stuff did need that for nested dictionaries.

PR #9396, opened 2026-05-14
LibWasm: Reduce instruction storage memory use

awesomekling:
Reduce memory use when parsing and compiling WebAssembly instructions.

This shrinks `Wasm::Instruction` from 88 bytes to 56 bytes by storing opcodes in 32 bits, using sentinel optionals, storing structured metadata and `ValueType` payloads directly, and moving `try_table` catches out of line.

It also replaces the compiled synthetic instruction `Vector` with chunked storage. Expressions with no synthetic instructions now allocate no extra instruction storage, while larger expressions grow in bounded chunks instead of reserving based on the original instruction count.

Altogether this reduces WebContent memory footprint by 53 MiB on https://x.com/awesomekling


Libraries/LibWasm/Types.h
```diff
@@ -10,6 +10,8 @@
 #include <AK/ByteString.h>
 #include <AK/DistinctNumeric.h>
 #include <AK/LEB128.h>
+#include <AK/NumericLimits.h>
```
alimpfard, line 14: Commit message: invalid opcodes don't seem to denote empty optional, just the one ::max() sentinel no?

Libraries/LibWasm/Types.h
```diff
@@ -769,12 +770,52 @@ union SourcesAndDestination {
     u32 sources_and_destination;
 };
 
+class InstructionStorage {
```
alimpfard, line 842: Nice, I've been meaning to do something like this but the alloc had never annoyed me enough :P

Libraries/LibWasm/Types.h
```diff
@@ -769,12 +770,52 @@ union SourcesAndDestination {
...
+
+public:
+    InstructionStorage() = default;
+    InstructionStorage(InstructionStorage const& other)
+    {
+        VERIFY(other.is_empty());
+    }
+    InstructionStorage& operator=(InstructionStorage const& other)
+    {
+        VERIFY(other.is_empty());
```
alimpfard: wat?

Libraries/LibWasm/Types.h
```diff
@@ -769,12 +770,52 @@ union SourcesAndDestination {
...
 };
 
+class InstructionStorage {
+    using Chunk = FixedArray<Optional<Instruction>>;
+
+public:
+    InstructionStorage() = default;
+    InstructionStorage(InstructionStorage const& other)
+    {
+        VERIFY(other.is_empty());
```
alimpfard: Is there some code that explicitly uses these weirdo copy-empty ctors...?

Libraries/LibWasm/Types.h
```diff
@@ -631,8 +673,14 @@ class Instruction {
...
     struct TryTableArgs {
+        TryTableArgs(StructuredInstructionArgs try_, ReadonlySpan<Catch> catches)
+            : try_(move(try_))
+            , catches(catches)
+        {
+        }
+
         StructuredInstructionArgs try_; // "else" unused.
-        Vector<Catch> catches;
+        CatchList catches;
```
alimpfard: This is a bit...ugly, especially since there's already some 16b wasted for "else", how about templating that and shoving the vector there? Should be net 0 compared to this.

alimpfard (commented):
Nice, about time someone got annoyed by wasm memory use lol

alimpfard:
(p.s. sanitizer caught stack uaf https://github.com/LadybirdBrowser/ladybird/actions/runs/25851009208/job/75957259792?pr=9396#step:19:30)

Libraries/LibWasm/Types.h
awesomekling: Done. Good idea, shrank Instruction by another 8 bytes too :D

alimpfard (approved):
Much nicer, thanks!

Issue #9401, opened 2026-05-14
Looking for good first issues?


AtkinsSJ:
Welcome to Ladybird! :^)

As you have probably noticed, we do not have any issues labeled "good first issue", for a few reasons: We've found that it's difficult to estimate how difficult an issue will be to solve, it's easy for these to go out of date, and a lot of the time it's faster to just fix a small issue than to write an explanation of it on GitHub.

There are many ways of finding small issues to fix though! Our **Getting Started guide** has [some suggestions](https://github.com/LadybirdBrowser/ladybird/blob/master/Documentation/GettingStartedContributing.md#finding-bugs-and-other-issues). A major way is to try visiting your favorite sites, see what's broken, and try to fix it. It's far more satisfying to fix a bug on a site you care about, than to fix something that doesn't affect you. Our [Discord server](https://discord.gg/nvfjVJ4Svh) is always here if you need some suggestions for how to diagnose or fix the issue.

PR #9403, opened 2026-05-14
Documentation: Remove mention of "good first issue" label

AtkinsSJ:
I've made a meta-issue at #9401 to direct anyone who is looking for good first issues to explain why we don't have them, and what to do instead. Noticed that this doc mentions them, which isn't helpful, so let's remove that.


shannonbooth:
unfortunately I think this needs rebase due to the build breakage accidentally introduced on master

PR #9405, opened 2026-05-14
LibWeb/CSS: Implement `@container` size queries

AtkinsSJ:
The strategy is relatively basic: During layout, we produce a list of elements whose style depends on a size container, then we do another pass to recompute their style and layout. And then repeat this for any of their descendants that depend on their size, and so on.

Gives us some nice progress on https://ux.redhat.com/

## Before
Screenshot_20260514_145132

## After
Screenshot_20260514_145206


AtkinsSJ:
...and a little further down the page:

## Before
Screenshot_20260514_150034

## After
Screenshot_20260514_150003

Psychpsyo:
Your first Before screenshot has the arrows in the left sidebar white, but they're black in all the other ones. Is that a separate issue or did the container queries break that?

AtkinsSJ:
I don't know what that is, but it seems to happen both with and without my changes: the arrow colour is a bit random, but clicking the page corrects it to white.

There's a lot of custom-element stuff going on in that page, so I'm guessing it's related to that.

AtkinsSJ:
Rebased, and made some minor code-style changes for things I now realise were a little silly.

AtkinsSJ:
Rebased to fix conflicts.

PR #9421, opened 2026-05-15
LibWeb+LibWebView+WebContent: Support primary pasting with middle mouse

### LINKED ISSUE #9398: No primary paste (middle-click paste) integration
### Summary

I use primary paste all the time, e.g. when wanting to copy text out of a webpage or when wantign to paste text from my text editor into a text box on a website, without the need to retain that text for longer than that one interaction.   

I would also suggest if this is added, to make it only trigger on release of the middle-click button, and only if the cursor did not move since middle-click was depressed (similar to how Firefox does it), in order to not conflict as much with autoscroll. If a paste is triggered, it should probably also prevent autoscrolling (just like middle-clicking a link to open it in a new tab does already), to not accidentally scroll far away after pasting stuff into a text box. When the element receiving the middle-click does not support text input, primary paste should be completely ignored to still allow autoscroll through middle-click followed by moving the cursor without holding down middle-click (not only middle-click-drag), and it should be completely disableable (and maybe disabled by default), like autoscroll. 

### Operating system

Linux

### Steps to reproduce

1. Select text in e.g. Kate on Linux or other OS that implements primary paste
2. Middle-click on a text box inside a webpage in Ladybird

Alternatively:
1. Select text on a webpage inside Ladybird on Linux or other OS that implements primary paste
2. Middle-click in e.g. Kate

### Expected behavior

The selected text should appear where middle-clicked (without overwriting the contents of the <kbd>Ctrl+C</kbd> clipboard).

### Actual behavior

The selected text does not appear where middle-clicked. If trying to paste into Ladybird, nothing at all happens, if trying to paste out of Ladybird whatever text was previously selected outside Ladybird appears instead. 

N/A

```html
N/A
```

```shell
N/A
```

https://github.com/user-attachments/assets/771b6929-c6f0-4ebe-83b3-e973097679ca

_No response_

### Contribute a patch?

trflynn89:
When the middle mouse button is clicked on a text input control or `contenteditable` node, we now request the UI process to paste text into that node.

Fixes #9398

### DIFF
--- a/Base/res/ladybird/about-pages/settings.html
+++ b/Base/res/ladybird/about-pages/settings.html
@@ -378,6 +378,16 @@
                             Scroll pages by pressing the middle mouse button and moving the mouse.
                         </p>
                     </div>
+                    <hr />
+                    <div class="card-group">
+                        <div class="inline-container">
+                            <label for="enable-primary-paste">Enable primary pasting</label>
+                            <input id="enable-primary-paste" type="checkbox" switch />
+                        </div>
+                        <p class="description">
+                            Paste text from the clipboard when the middle mouse button is clicked.
+                        </p>
+                    </div>
                 </div>
             </div>
         </div>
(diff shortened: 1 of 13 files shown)


Lampe2020:
The behaviour implemented here is incorrect, primary paste should be completely independent from the <kbd>Ctrl+C</kbd>/<kbd>Ctrl+V</kbd> clipboard.   
(this implements it as normal paste)   
see https://github.com/LadybirdBrowser/ladybird/issues/9398#issuecomment-4475349138

PR #9426, opened 2026-05-15
LibWeb: Avoid GC::Root churn in wheel listener checks

kalenikaliaksandr:
Async scrolling metadata recording checks many paintable nodes for blocking wheel listeners while building a display list. The helper used EventTarget::event_listener_list(), which returns a snapshot backed by GC::Root entries, so the hot paint path allocated and destroyed roots just to answer a boolean query.

Add a non-allocating EventTarget predicate that scans the stored listener references directly, and use it from async scrolling metadata recording. Event dispatch still uses event_listener_list() so it keeps the required snapshot semantics.

### DIFF
--- a/Libraries/LibWeb/DOM/EventTarget.cpp
+++ b/Libraries/LibWeb/DOM/EventTarget.cpp
@@ -906,6 +906,11 @@ bool EventTarget::has_event_listener(FlyString const& type) const
     return m_data && m_data->event_listener_list.contains([&type](auto listener) { return listener->type == type; });
 }
 
+bool EventTarget::has_blocking_wheel_event_listener() const
+{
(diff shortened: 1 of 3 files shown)


Libraries/LibWeb/DOM/EventTarget.cpp
```diff
@@ -906,6 +906,13 @@ bool EventTarget::has_event_listener(FlyString const& type) const
     return m_data && m_data->event_listener_list.contains([&type](auto listener) { return listener->type == type; });
 }
 
+bool EventTarget::has_blocking_wheel_event_listener() const
+{
+    return m_data && m_data->event_listener_list.contains([](auto const& listener) {
```
gmta:
```cpp
.contains(is_blocking_wheel_event_listener)
```
...should also work!

gmta (approved):

PR #9429, opened 2026-05-15
LibWeb: Replace the HTML tokenizer with Rust

awesomekling:
This replaces LibWeb's HTML tokenizer with a Rust implementation and removes the old C++ tokenizer.

The new tokenizer stays behind the existing HTMLTokenizer API used by the HTML parser. It preserves the parser integration points for streaming input, insertion points, `document.write()`, EOF insertion, parser aborts, speculative parser input, and last start tag tracking. The Rust FFI handle remains an implementation detail, so callers still interact with a single tokenizer class.

The implementation uses generated DAFSA tables for named character references and interns common tag and attribute names to reduce FFI overhead. It also keeps duplicate attributes visible to C++ token normalization so CSP nonce checks keep their duplicate-attribute signal, and fast tokenizer scans stop at insertion points so streamed input is spliced at the right offset.

This also adds `dump-html-tokens` and focused tokenizer fixtures for direct token output checks. `TestHTMLTokenizer` now covers duplicate attributes and insertion points inside fast tag-name, attribute-name, and quoted attribute-value scans. A CSP text test covers duplicate nonce attributes on parser-created script elements.

I checked the replacement against a downloaded corpus built from 6747 websites across 200 TLDs, with 0 semantic token mismatches and 0 tokenizer errors between the C++ and Rust tokenizers.

Benchmarking the same 6747 files found Rust faster after high-iteration reruns of noisy ties, with 0 slower files, 0 equal files, and a 3.830x median speedup.


AtkinsSJ:
Slightly alarming that this ends up with ~2400 more lines of code even discounting the tests. :sweat_smile:

awesomekling:
> Slightly alarming that this ends up with ~2400 more lines of code even discounting the tests. 😅

Yeah, it looks big because the diff is counting the whole Rust island, not just tokenizer logic :)

Rough breakdown:

  +3152  Rust tokenizer state machine
   +821  build.rs generator for named character refs + interned names + cbindgen
   +702  Rust FFI/marshalling layer
   +202  Rust token/attribute/doctype structs
   +151  Rust named character reference matcher wrapper
    +61  Rust interned-name wrapper
   +155  dump-html-tokens utility
   +100-ish focused tests
  -2960  old C++ HTMLTokenizer.cpp

So the net +2000-ish mostly comes from:

- C++ reused existing LibWeb types directly. Rust needs its own `Token`, `Attribute`, `DoctypeData`, positions, and then FFI code to convert back into C++ `HTMLToken`.

- Rust generator+matcher for the Entities.json data.

- The old C++ tokenizer was very dense. It used macro/goto-heavy state machine code. The Rust version is more explicit and has helper structs instead of leaning on macros and existing C++ classes.

- New tooling and tests. `dump-html-tokens`, fixtures, CSP coverage, and insertion-point tests are all new PR surface.

PR #9434, opened 2026-05-15
LibWeb: Don't crash on `calc` in `scroll-margin`

dgolebiewski:
This PR fixes an oversight in #9367 which causes a crash when `scroll-margin` or `scroll-padding` has a `calc` or a percentage value.


shannonbooth:
Could you please include a test which reproduces the crash?

dgolebiewski:
Done, you can also visit any page on [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/scroll-behavior) and click on any "In this article" anchor to cause the crash. They use `calc` in `scroll-margin` to offset the scroll by header height - that's how I found this out.

PR #9437, opened 2026-05-16
LibWeb: Absolutize `@font-face` descriptors before reading

tcl3:
This PR prevents crashes when reading `@font-face` descriptors with `calc()` values containing relative units. See individual commits for details.

NB: The change in the last commit is not currently observable, but I made the change to match the style used in the previous commits.


tcl3:
The crashing WPT this PR was targeting was fixed in #9576. I believe this approach is still worth considering as it fixes a broader class of crashes - all of the crash tests included in this PR still crash on `master`.

PR #9443, opened 2026-05-16
LibWeb: Improve support for non-synthetic pseudo-elements

Calme1709:
Pseudo-elements are implemented in two ways, either:
 - "synthetic": The originating element is authoritative for the pseudo-element's data (e.g. animations, computed
  style, etc) and handles it's behavior (such as generating layout nodes) itself. This includes pseudo-elements such
  as "::first-line" and "::before"
 - "element-reference": The pseudo-element refers to a "real" element elsewhere in the DOM, usually in the originating
  element's shadow tree, and is thus not authoritative. This includes, among others, form pseudo-elements such as "::placeholder".

This PR improves support for "element-reference" pseudo-elements by:
 - Accounting for inline style and applying "resolved value" computation in `getComputedStyle`
 - Correctly applying animations in `update_animated_style_if_needed` (previously it would only apply if the element was invalidated for some other reason).
 - Applying inline style at the correct point during computation.


Calme1709:
Fixed an issue caught by the clang plugins

Calme1709:
Rebased to resolve conflicts

Calme1709:
Rebased to resolve conflicts

Libraries/LibWeb/CSS/PseudoElements.json
```diff
@@ -120,26 +130,33 @@
...
-    "is-element-backed": true
+    "is-element-backed": true,
+    "implementation": "element-reference"
   },
   "view-transition": {
     "spec": "https://drafts.csswg.org/css-view-transitions-1/#selectordef-view-transition",
-    "is-pseudo-root": true
+    "is-pseudo-root": true,
+    "implementation": "synthetic",
+    "__comment": "FIXME: Is 'synthetic' correct here? It's not implemented yet"
```
AtkinsSJ: I think so? But @Psychpsyo knows the most about that.

Meta/Generators/generate_libweb_css_pseudo_element.py
```diff
@@ -214,6 +214,27 @@ def write_header_file(out: TextIO, pseudo_elements_data: dict) -> None:
...
+    element_reference_pseudo_elements = []
+    functional_pseudo_elements = []
+
+    for name, pseudo_element in pseudo_elements_data.items():
+        if is_alias(pseudo_element):
+            continue
+
+        if pseudo_element.get("type") == "function":
+            functional_pseudo_elements.append(name)
+            continue
```
AtkinsSJ, line 227: This excludes function-type PEs from defining an implementation, but `::part()` is defined as having both.

Documentation/CSSGeneratedFiles.md
```diff
@@ -257,17 +257,18 @@ This generated `PseudoElement.h` and `PseudoElement.cpp`.
...
+| `property-whitelist` | No                         | Nothing        | Some pseudo-elements only permit certain properties. If so, name them in an array here. Some special values are allowed here for categories of properties - see below. |
+| `spec`               | No                         | Nothing        | Link to the spec definition, for reference. Not used in generated code.                                                                                                |
+| `type`               | No                         | `"identifier"` | What type of pseudo-element is this. Either "identifier", "function", or "both".                                                                                       |
+| `implementation`     | Unless `type` is "function | Nothing        | How this pseudo-element is implemented, either `"synthetic"` or `"element-reference"` - see below.                                                                     |
```
AtkinsSJ: Nit: This table is alphabetical.

Libraries/LibWeb/DOM/Element.cpp
```diff
@@ -3602,9 +3611,24 @@ void Element::set_custom_property_data(Optional<CSS::PseudoElement> pseudo_eleme
     if (!CSS::Selector::PseudoElementSelector::is_known_pseudo_element_type(pseudo_element.value()))
         return;
 
-    if (data)
-        ensure_pseudo_element(pseudo_element.value()).set_custom_property_data(move(data));
-    else if (auto existing_pseudo_element = get_pseudo_element(pseudo_element.value()); existing_pseudo_element.has_value())
+    if (data) {
+        if (is_synthetic_pseudo_element(pseudo_element.value()))
+            ensure_synthetic_pseudo_element(pseudo_element.value()).set_custom_property_data(move(data));
+        else {
```
AtkinsSJ: Nit: If one branch has {}, all of them should.

Libraries/LibWeb/CSS/StyleComputer.cpp
```diff
@@ -477,16 +489,15 @@ void StyleComputer::apply_property_list_to_cascade(
     Important important,
     Optional<FlyString> layer_name,
     GC::Ptr<CSSStyleDeclaration const> source,
-    GC::Ptr<DOM::ShadowRoot const> source_shadow_root) const
+    GC::Ptr<DOM::ShadowRoot const> source_shadow_root,
+    bool bypass_pseudo_element_property_whitelist) const
```
AtkinsSJ: Nit: This should probably be a named enum, it's a method with a lot of arguments and `true` or `false` isn't clear in meaning.

Libraries/LibWeb/CSS/PseudoElements.json
Psychpsyo: Yes, `::view-transition-root` is synthetic. And all the other view-transition ones as well, although they have weird special tree nonsense going on.

AtkinsSJ (changes requested):
A few minor things, but overall this is very nice. Much cleaner! Thanks for doing it.

Libraries/LibWeb/CSS/PseudoElements.json
Calme1709: Have removed this FIXME

Meta/Generators/generate_libweb_css_pseudo_element.py
Calme1709: I have removed the definition of `::part` as `element-reference` - function pseudo elements are a bit different in that we can't store references to them (since they can refer to different elements depending on their parameters) so I think we will just have to look them up on the fly.

AtkinsSJ (approved):

PR #9445, opened 2026-05-16
UI/Qt: Disable QWinEventNotifier before closing process_handle

heitke:
When a child process terminates on Windows, the QWinEventNotifier emits 
the activated signal it its event handler. Before this happens, we need to 
call setEnabled(false) to unregister the wait object from the Windows 
thread pool. setEnabled(false) also causes us to wait for any outstanding 
callbacks to complete and then prevents QWinEventNotifier from trying 
to use the closed handle.

This seems to follow the pattern from Qt where the destructor also calls 
setEnabled(false) before destruction.

### DIFF
--- a/UI/Qt/EventLoopImplementationQt.cpp
+++ b/UI/Qt/EventLoopImplementationQt.cpp
@@ -406,6 +406,7 @@ void EventLoopManagerQt::register_process(pid_t pid, ESCAPING Function<void(pid_
                     VERIFY(maybe_process.value() == process);
             });
 
+            process->setEnabled(false);
             CloseHandle(process_handle);
             process->deleteLater();
             exit_handler(process_id);
@@ -425,13 +426,16 @@ void EventLoopManagerQt::unregister_process(pid_t pid)
 
     auto* process = maybe_process.release_value();
     HANDLE process_handle = process->handle();
+
     if (QThread::currentThread() != process->thread()) {
         QMetaObject::invokeMethod(process, [process, process_handle] {
+            process->setEnabled(false);
             CloseHandle(process_handle);
             delete process; }, Qt::QueuedConnection);
         return;
     }
 
+    process->setEnabled(false);
     CloseHandle(process_handle);
     delete process;
 }


heitke:
This PR fixes an unhandled exception (0xC000070A - STATUS_INVALID_HANDLE) that was happening when navigating to any page on a native Windows build. This happened because QWinEventNotifier was still trying to use the handle after it had been closed. QWinEventNotifier has a safety check for this, but it doesn't actually check the handle (d->handleToEvent) to see if it is null. Instead, it checks three other things:

```
            if (alive && d->enabled && !d->registered) {
                SetThreadpoolWait(d->waitObject, d->handleToEvent, NULL);
                d->registered = true;
            }
```
By first calling:

`process->setEnabled(false);
`

We now properly clean-up before destruction and d->enabled gets set to false.

I'll be honest, I don't fully understand the intricacies of what is going on with Qt and the lifecycle of the processes here. But this PR definitely solves the exception that is being thrown, and also seems correct to me because the deconstructor of QWinEventNotifier has this line of code.

Here is the full callstack of the issue without the fix:

```

 	[Unhandled exception at 0x00007FFAC928AA64 (ntdll.dll) in Ladybird.exe: 0xC000070A: Status 0x (parameters: 0xFFFFFFFFC0000008, 0x0000000000000414, 0x00000253ADDFDAF0, 0x0000000000000000, 0x00007FF8D74283C1).]	
 	ntdll.dll!00007ffac928aa64()	Unknown
 	ntdll.dll!00007ffac9240401()	Unknown
>	Qt6Cored.dll!QWinEventNotifier::event(QEvent * e) Line 211	C++
 	Qt6Widgetsd.dll!QApplicationPrivate::notify_helper(QObject * receiver, QEvent * e) Line 3307	C++
 	Qt6Widgetsd.dll!QApplication::notify(QObject * receiver, QEvent * e) Line 3257	C++
 	Qt6Cored.dll!QCoreApplication::notifyInternal2(QObject * receiver, QEvent * event) Line 1109	C++
 	Qt6Cored.dll!QCoreApplication::sendEvent(QObject * receiver, QEvent * event) Line 1550	C++
 	Qt6Cored.dll!QCoreApplicationPrivate::sendPostedEvents(QObject * receiver, int event_type, QThreadData * data) Line 1904	C++
 	Qt6Cored.dll!QEventDispatcherWin32::sendPostedEvents() Line 925	C++
 	Qt6Guid.dll!QWindowsGuiEventDispatcher::sendPostedEvents() Line 44	C++
 	Qt6Cored.dll!QEventDispatcherWin32::processEvents(QFlags<enum QEventLoop::ProcessEventsFlag> flags) Line 483	C++
 	Qt6Guid.dll!QWindowsGuiEventDispatcher::processEvents(QFlags<enum QEventLoop::ProcessEventsFlag> flags) Line 36	C++
 	Qt6Cored.dll!QEventLoop::processEvents(QFlags<enum QEventLoop::ProcessEventsFlag> flags) Line 105	C++
 	Qt6Cored.dll!QEventLoop::exec(QFlags<enum QEventLoop::ProcessEventsFlag> flags) Line 186	C++
 	Qt6Cored.dll!QCoreApplication::exec() Line 1452	C++
 	Ladybird.exe!Ladybird::EventLoopImplementationQt::exec() Line 226	C++
 	lagom-core.dll!Core::EventLoop::exec() Line 99	C++
 	lagom-webview.dll!WebView::Application::execute() Line 695	C++
 	Ladybird.exe!ladybird_main(Main::Arguments arguments) Line 95	C++
 	Ladybird.exe!main(int argc, char * * argv) Line 53	C++
 	[External Code]	
```

UI/Qt/EventLoopImplementationQt.cpp
```diff
@@ -425,6 +426,9 @@ void EventLoopManagerQt::unregister_process(pid_t pid)
 
     auto* process = maybe_process.release_value();
     HANDLE process_handle = process->handle();
+
+    process->setEnabled(false);
```
shannonbooth: Is it safe to be doing this cross thread like this and not in the blocks below like how the handle is managed?

heitke: Good call out, Shannon. I think it's ok though. I believe that the block on lines 403-307 remove the process from the Mutex protected list of processes. At that point, we essentially "own" the process so we don't have to relock it again to use it. I could be wrong though! Open to additional pushback.

shannonbooth: Hmm okay, I do agree in thinking that pointer is logically owned in a sense, but my (limited) understanding of Qt is that it also cares about what thread owns QObject itself, and has assertions based on that. So I _think_ it's not safe to be doing it in this way. Even if Qt happened to tolerate this, moving the call into the queued lambda and keeping it in the same thread path seems like it would be definitely valid and is easier to reason about.

heitke: Just so I'm clear - you are suggesting that we wrap these 4 lines in an invokeMethod call like this, right?

`QMetaObject::invokeMethod(process, [process, process_handle] { ... }`

It still feels to me like when Qt is the one calling us back during the event handler that we are in, it would be guaranteed to be on the correct thread... actually, doesn't line 400 guarantee that?

`if (process_handle != terminated_process_handle)`

shannonbooth: Yes - to invoke it at the same places CloseHandle is done.
 
I'm not 100% sure about that path, but my concern was more around unregister_process in general since that can in theory be called from another thread by the EventLoop singleton and it does seem like this should have the same lifetime as the handle

heitke: Ah got it! I was looking at the register_process more since that's where the original exception occurred. I agree.  I'll change that and push the fix! Sorry it took me awhile to get there :)

shannonbooth (approved):

PR #9449, opened 2026-05-16
LibWeb: Reduce memory growth in long-lived pages

awesomekling:
While analyzing memory growth over time on https://x.com, several independent retention paths showed up in long-lived documents. This tightens those lifetimes by bounding per-document decoded image caches, releasing completed shared image and link fetch controllers, clearing timers when documents become inactive, and preventing internal IndexedDB databases from keeping realms alive.

The image cache keeps a small hot working set for repeat loads while pruning stale decoded image data from shared image requests and the list of available images. The fetch and timer changes make completed or inactive work stop retaining request clients, documents, and realms longer than needed.

### DIFF
--- a/Libraries/LibWeb/DOM/Document.cpp
+++ b/Libraries/LibWeb/DOM/Document.cpp
@@ -115,6 +115,7 @@
 #include <LibWeb/HTML/CustomElements/CustomElementDefinition.h>
 #include <LibWeb/HTML/CustomElements/CustomElementReactionNames.h>
 #include <LibWeb/HTML/CustomElements/CustomElementRegistry.h>
+#include <LibWeb/HTML/DecodedImageData.h>
 #include <LibWeb/HTML/DocumentState.h>
 #include <LibWeb/HTML/DragEvent.h>
 #include <LibWeb/HTML/EventLoop/EventLoop.h>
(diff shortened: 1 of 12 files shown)


Libraries/LibWeb/IndexedDB/Internal/Database.h
```diff
@@ -61,9 +62,9 @@ class Database : public Bindings::PlatformObject {
     explicit Database(IDBDatabase& database);
 
     explicit Database(JS::Realm& realm, String name)
-        : PlatformObject(realm)
-        , m_name(move(name))
+        : m_name(move(name))
     {
+        (void)realm;
```
shannonbooth: Does this have too much churn to remove as an argument or something along those lines?

Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -5216,6 +5216,9 @@ bool Document::is_allowed_to_use_feature(PolicyControlledFeature feature) const
 
 void Document::did_stop_being_active_document_in_navigable()
 {
+    auto& window = as<HTML::WindowOrWorkerGlobalScopeMixin>(HTML::relevant_global_object(*this));
+    window.clear_map_of_active_timers();
```
shannonbooth: I am a bit confused by this in terms of how these timers are cleared in the specification. did_stop_being_active_document_in_navigable seems to be called in the same path that run_unloading_cleanup_steps  as called, and run_unloading_cleanup_steps if the document is not salvageable. So I suppose my question is whether this should trigger if the document becomes active again.

Libraries/LibWeb/IndexedDB/Internal/Database.h
awesomekling: Nah we can fix that :) Just me being lazy

Libraries/LibWeb/DOM/Document.cpp
shannonbooth: Looking into this some more, from checking out this commit, it seems like the test passes whether or not window.clear_map_of_active_timers is called here, so maybe the issue seen on twitter is to do with documents _not_ being marked unsalvageable for some reason?

shannonbooth: (document lifecycle is a hairy thing) 😅

awesomekling: Hmm, indeed. Let's drop this commit!

PR #9450, opened 2026-05-16
LibCore/LibIPC: Implement an IPC thread on Windows

R-Goc:
This PR along with #9445 bring ladybird back to working on Windows. This is done through implementing an IO thread. Because of the issue that was solved by #9445 when chasing down backtraces caused by it, the process cleanup on windows was fixed to exit gracefully. 

Commits 2,3,4,5 are AI assisted to some degree. The IPC implementation is very close to what exists on POSIX, with some minor differences as explained in the commit message.


Services/WebContent/ConnectionFromClient.cpp
```diff
@@ -128,12 +128,13 @@ ConnectionFromClient::~ConnectionFromClient() = default;
 
 void ConnectionFromClient::die()
 {
-    _exit(0);
+    Core::EventLoop::current().quit(0);
```
shannonbooth, line 131: I'm not quite understanding the issue in general for this last commit, but why can't we just `_exit(0)`? Why do we need to do cleanup on Windows? Would it not be something to do with initialization that would be the problem?

shannonbooth (commented):
First three commits look good to me, fourth is going to take some brain power to understand (and I'm probably not the right person to review that anyhow), but I don't quite understand the problem in the 5th commit at all

Services/WebContent/ConnectionFromClient.cpp
R-Goc: That was an issue on cleanup that was causing some unrelated verifies to the actual problem. Also _exit() on windows doesn't do any cleanup and doesn't flush file buffers:
> The exit, _Exit and _exit functions terminate the calling process. The exit function calls destructors for thread-local objects, then calls—in last-in-first-out (LIFO) order—the functions that are registered by atexit and _onexit, and then flushes all file buffers before it terminates the process. The _Exit and _exit functions terminate the process without destroying thread-local objects or processing atexit or _onexit functions, and without flushing stream buffers.

I think having unrelated verifies isn't great. Also this isn't exactly explained here, but static object destructors can be registered as part of the onexit table. So we weren't calling destructors for ImageDecoder and ResourceLoader. Also this way we go back to main and are able to call the shutdown functions added there.

shannonbooth: Unix is the same, but it's an intentional decision that we don't run destructors etc on shutdown

R-Goc: Does that also lead to unrelated VERIFYs on crash on unix platforms?

shannonbooth: No, but I don't quite understand how it could, is'nt it just hard quitting the process? Normally exiting the process would cause crashes since there are known (not really intended to be solved) problems with process teardown, such as ordering issues

shannonbooth:
Would it maybe be easier to merge in the first 3 separately?  #9770 needs those, and the last two we will need to figure out between that MR and this.

heitke:
Good question, @shannonbooth. I didn't do any of the work of commit 1 on my PR because I wasn't 100% sure if it was needed or not. Commit 2 here is just extra debug info. It can probably be committed first, or abandoned.

Commit 3 was mostly copied into my PR with just a few small tweaks. 4 and 5 were where all of the real work was - and applied to a previous architecture, so I think those can be safely abandoned.

So you and @R-Goc can decide if you want to merge 1 and 2. Or, if it's easier I can pull those changes into my PR as well, just let me know.

gmta:

PR #9456, opened 2026-05-17
LibWeb: Resolve SVG max-content width from CSS height

awesomekling:
When a replaced element has no natural width or height but has a preferred aspect ratio, `max-content` width fell through to the CSS Sizing fallback that uses the initial containing block width.

Use a definite used height to derive `max-content` width through the aspect ratio before the fallback. Keep the existing natural-height SVG path first by only doing this for elements without natural height. Add a focused layout test for the nested flex item case.

### DIFF
--- a/Libraries/LibWeb/Layout/FormattingContext.cpp
+++ b/Libraries/LibWeb/Layout/FormattingContext.cpp
@@ -35,7 +35,7 @@ enum class SizeDimension {
 };
 
 // https://drafts.csswg.org/css-sizing-3/#intrinsic-sizes
-static Optional<CSSPixels> max_content_size_for_replaced_element_without_natural_size(Box const& box, CSS::SizeWithAspectRatio const& natural_size, SizeDimension dimension)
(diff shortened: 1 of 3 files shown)


Libraries/LibWeb/Layout/FormattingContext.cpp
```diff
@@ -2059,9 +2059,15 @@ CSSPixels FormattingContext::calculate_min_content_width(Layout::Box const& box)
 
 CSSPixels FormattingContext::calculate_max_content_width(Layout::Box const& box) const
 {
-    if (auto auto_size = box.auto_content_box_size(); auto_size.has_width())
+    auto auto_size = box.auto_content_box_size();
+    if (auto_size.has_width())
         return auto_size.width.value();
-    if (auto max_content_width = max_content_size_for_replaced_element_without_natural_size(box, box.auto_content_box_size(), SizeDimension::Width); max_content_width.has_value())
+    if (box.is_replaced_box() && !auto_size.has_height() && auto_size.has_aspect_ratio()) {
```
shannonbooth: This seems very similar to the case I was trying to figure out / getting confused with here: https://github.com/awesomekling/ladybird/blob/8670b964ca6bcaed1ae34944d9b9a212111f2dd6/Libraries/LibWeb/Layout/FormattingContext.cpp#L57, just without natural_size.has_aspect_ratio check and using definite height instead. I'm wondering if should structure those two checks together to avoid the fall through

PR #9457, opened 2026-05-17
LibWeb: Replace HTML parser with Rust

awesomekling:
This ports HTML tree construction to Rust and makes that implementation the unconditional parser backend. The C++ side remains responsible for LibWeb DOM integration, encoding detection, tokenizer bridging, incremental parsing, parser-blocking script coordination, and speculative resource discovery.

The Rust parser covers document and fragment parsing, insertion modes, active formatting element reconstruction, foster parenting, adoption agency handling, foreign-content integration, template handling, declarative shadow roots, and script/document.write reentrancy. Once the Rust path is always used, the old C++ tree builder and its helper classes are removed.

This also adds focused test coverage for parser edge cases encountered while working on this port.

Verified against our test suite + by downloading ~7000 websites from around the world in various languages/encodings and checking that we produce the exact same DOM as the old C++ version.

For posterity, on that same corpus of websites, the Rust parser is roughly 1.1x faster than the C++ version. :^)


Libraries/LibWeb/HTML/HTMLScriptElement.cpp
```diff
@@ -201,6 +201,11 @@ void HTMLScriptElement::execute_script()
 // https://html.spec.whatwg.org/multipage/scripting.html#prepare-a-script
 void HTMLScriptElement::prepare_script()
 {
+    // AD-HOC: Don't prepare scripts while in the temporary document used for fragment parsing. They will be prepared
+    // when inserted into a real document, unless fragment parsing already marked them as already started.
+    if (document().is_temporary_document_for_fragment_parsing())
```
shannonbooth: Out of interest what is the change that necessitates this? Has something changed wrt integration with "the end" here?

awesomekling: Ah yeah, good question. Nothing changed in `the end`! Fragments still skip it.

This is about `createContextualFragment()`: scripts from that path are supposed to stay unstarted while parsing, then run when the fragment is inserted into the real document. During fragment parsing they can still hit normal DOM insertion hooks while in the temporary fragment document, so `prepare_script()` would mark them started too early. This guard prevents that.

This was an issue with the C++ parser as well, it's just something that came up while adding tests here. :)

shannonbooth: Ah! I think I actually know the root cause of this one / proper fix. I believe the root cause is that we consider fragment documents connected. If we fix that, then I think that solves that issue alongside some other problems. https://github.com/whatwg/html/issues/11023 is the spec issue which was originally raised for it, but I independantly came across it for createContextualFragment as well. There's a WPT somewhere which tests that createContextualFragment case

PR #9459, opened 2026-05-17
LibJS: Bound the shared RegExp compilation cache

bsdrop:
The compiled-regex cache in RegExpPrototype.cpp is a process-wide HashMap with no upper bound. A page that creates a stream of distinct patterns (a common shape is \`new RegExp("a" + i)\` in a loop) can grow the process-wide cache without limit.

This caps the shared cache at 1024 entries. Once it is full, further unique patterns are compiled into a regex that the RegExpObject owns directly.

The cap is intentionally conservative and the patch does not introduce eviction: live RegExpObjects hold raw pointers into the shared cache, and an LRU-style policy would require additional lifetime tracking to avoid use-after-free. Per-object ownership for over-cap patterns avoids that entirely.

## Test plan
- [x] New \`Tests/LibJS/Runtime/regexp-shared-cache-overflow.js\` — 1500 distinct patterns, all still match correctly after the cache fills
- [x] Existing \`Tests/LibJS/**/*regexp*\` suite: 90/90 suites, 221/221 tests pass
- [x] \`Build/release/bin/js\` smoke: 5000 distinct RegExps compile and match correctly

### DIFF
--- a/Libraries/LibJS/Runtime/RegExpObject.cpp
+++ b/Libraries/LibJS/Runtime/RegExpObject.cpp
@@ -375,6 +375,7 @@ ThrowCompletionOr<GC::Ref<RegExpObject>> RegExpObject::regexp_initialize(VM& vm,
 {
     // Invalidate the cached compiled regex since the pattern/flags may change.
     m_cached_regex = nullptr;
+    m_owned_regex = nullptr;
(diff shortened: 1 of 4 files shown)


gmta:

PR #9460, opened 2026-05-17
LibURL: Port the URL parser and pattern implementation to rust

shannonbooth:


shannonbooth:
I benchmarked this against the ADA-URL URL corpus to check if we are faster or slower after this. Seems like even with FFI this is around 1.5x faster than the C++ implementation (though we still are _a lot_ slower than that implementation.

PR #9461, opened 2026-05-17
LibWeb: Implement legacy-clone of session storage shed

Martin005:
When a new window is opened via window.open() with an opener (non-null), the session storage must be cloned from the opener's storage shed into the new window's storage shed. This implements the legacy-clone traversable storage shed algorithm from the Storage spec.

### DIFF
--- a/Libraries/LibWeb/StorageAPI/StorageShed.cpp
+++ b/Libraries/LibWeb/StorageAPI/StorageShed.cpp
@@ -4,9 +4,11 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
+#include <AK/TypeCasts.h>
 #include <LibGC/Heap.h>
 #include <LibWeb/HTML/Scripting/Environments.h>
 #include <LibWeb/HTML/Window.h>
+#include <LibWeb/StorageAPI/StorageBottle.h>
 #include <LibWeb/StorageAPI/StorageShed.h>
 
 namespace Web::StorageAPI {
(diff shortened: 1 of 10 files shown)


Libraries/LibWeb/StorageAPI/StorageShed.cpp
```diff
@@ -38,4 +40,33 @@ GC::Ptr<StorageShelf> StorageShed::obtain_a_storage_shelf(HTML::EnvironmentSetti
...
+void StorageShed::legacy_clone(StorageShed const& a_storage_shed, GC::Ref<Page> page)
+{
+    // 1. For each key → shelf of A’s storage shed:
+    for (auto const& [key, shelf] : a_storage_shed.m_data) {
+
+        // 1. Let newShelf be the result of running create a storage shelf with "session".
+        auto new_shelf = StorageShelf::create(heap(), page, key, StorageType::Session);
+
+        // 2. Set newShelf’s bucket map["default"]'s bottle map["sessionStorage"]'s map to a clone of shelf’s bucket map["default"]'s bottle map["sessionStorage"]'s map.
+        auto shelf_bucket = shelf->bucket_map().get("default"sv);
```
shannonbooth: Two small improvements I think we can make here:
 * "default" should always be filled (if not let's note why), we shouldn't branch on it, it should instead be a assertion (same with shelf_bottle && new_shelf_bottle)
 * It'll only be SessionStorageBottle, but we may as well as<> this to be sure, this won't be in a hot path
 * The assignment can just be: target.m_map = source.m_map; - maybe we should define a helper in SessionStorageBottle instead of making this a friend?


Otherwise looks good to me

Libraries/LibWeb/StorageAPI/StorageShed.h
```diff
@@ -9,6 +9,7 @@
 #include <AK/HashMap.h>
 #include <LibGC/Ptr.h>
 #include <LibWeb/Forward.h>
+#include <LibWeb/Page/Page.h>
```
shannonbooth: it should not be neccessary to include this here it should already be forward declared in Forward.h

Martin005: Removed in [`5fa541e90240572a9a12f8248c6880b866c3fdd2..87969ea6ccc4833c77e135bd3a09f362833886a8`#diff-8008908675](https://github.com/LadybirdBrowser/ladybird/compare/5fa541e90240572a9a12f8248c6880b866c3fdd2..87969ea6ccc4833c77e135bd3a09f362833886a8#diff-8008908675622eda26287b6fbbdb628a81c4e0085bf67e45cc3e02543062a4fe)L12

Libraries/LibWeb/StorageAPI/StorageShed.cpp
Martin005: Sure, those improvements make sense. Updated in https://github.com/LadybirdBrowser/ladybird/compare/5fa541e90240572a9a12f8248c6880b866c3fdd2..87969ea6ccc4833c77e135bd3a09f362833886a8

PR #9475, opened 2026-05-18
LibWeb+LibHTTP+LibWebView: Implement HSTS

Lubrsi:
When an HTTPS response carries a Strict-Transport-Security header, the
received policy is now respected. Subsequent HTTP requests to a known
HSTS host are upgraded to HTTPS before the fetch algorithm makes
further decisions such as CORS and mixed content.

Fixes tpexpress.co.uk, where an XHR redirects HTTPS -> HTTP -> HTTPS,
relying on a HSTS policy received on the document response to avoid the
CORS failure.


Libraries/LibHTTP/HSTS/ParsedHSTSPolicy.cpp
```diff
@@ -0,0 +1,180 @@
...
+#include <LibHTTP/HSTS/ParsedHSTSPolicy.h>
+#include <LibHTTP/HTTP.h>
+#include <LibIPC/Decoder.h>
+#include <LibIPC/Encoder.h>
+
+namespace HTTP::HSTS {
+
+// https://www.rfc-editor.org/rfc/rfc7230#section-3.2.6
+// token = 1*tchar
+static bool is_valid_token(StringView value)
```
shannonbooth: Maybe it's not worth combining, but as a note we have an identical version of this as [[nodiscard]] bool is_header_name(StringView);
 (since [field-name](https://httpwg.org/specs/rfc9110.html#fields.names)     = [token](https://httpwg.org/specs/rfc9110.html#rule.token.separators))

Libraries/LibHTTP/HSTS/ParsedHSTSPolicy.h
```diff
@@ -0,0 +1,37 @@
...
+#include <AK/StringView.h>
+#include <AK/Types.h>
+#include <LibIPC/Forward.h>
+
+namespace HTTP::HSTS {
+
+// https://www.rfc-editor.org/rfc/rfc6797#section-6.1
+struct ParsedHSTSPolicy {
+    // https://www.rfc-editor.org/rfc/rfc6797#section-6.1.1
+    u64 max_age { 0 };
```
shannonbooth: Perhaps we can store this as an AK::Duration instead here? I hope this doesnt matter in reality, but one thing I noticed from WPT tests for preflight cache max age that all browsers aligned on is that they actually tested negative ages being same as 0 (instead of a parse error which I originally expected from RFC).

Libraries/LibWebView/HSTSStore.cpp
```diff
@@ -110,6 +111,11 @@ bool HSTSStore::is_known_hsts_host(StringView domain)
     // and continuing right-to-left.
     auto canonical = domain.to_ascii_lowercase_string();
 
+    // NB: Consult the built-in preload list before the dynamic store so a max-age=0 response from a preloaded
+    //     host cannot unset it.
+    if (HSTSPreloadData::the().is_known_preloaded_hsts_host(canonical))
```
shannonbooth: I wonder if it makes sense to do this from the web content process to avoid the IPC call?

Libraries/LibHTTP/HSTS/ParsedHSTSPolicy.cpp
```diff
@@ -0,0 +1,180 @@
...
+        return false;
+    for (auto byte : value.bytes()) {
+        if (!is_http_token_code_point(byte))
+            return false;
+    }
+    return true;
+}
+
+// https://www.rfc-editor.org/rfc/rfc7230#section-3.2.6
+static Optional<String> consume_quoted_string(GenericLexer& lexer)
```
shannonbooth, line 18: Is this equivalent to collect_an_http_quoted_string?

Libraries/LibHTTP/HSTS/ParsedHSTSPolicy.cpp
```diff
@@ -0,0 +1,180 @@
...
+Optional<ParsedHSTSPolicy> parse_header(StringView header_value)
+{
+    GenericLexer lexer(header_value);
+
+    Optional<u64> max_age;
+    bool include_sub_domains = false;
+
+    // 1. The order of appearance of directives is not significant.
+    while (!lexer.is_eof()) {
+        lexer.ignore_while(is_ascii_space);
```
shannonbooth: I can't recall if this is the same as what we need or not but it would be good to use HTTP_WHITESPACE when talking about whitespace parsing http

Libraries/LibHTTP/HSTS/ParsedHSTSPolicy.cpp
```diff
@@ -0,0 +1,180 @@
...
+            return {};
+        }
+
+        Optional<String> directive_value;
+        if (!lexer.is_eof() && lexer.peek() == '=') {
+            lexer.ignore(1);
+            lexer.ignore_while(is_ascii_space);
+
+            if (!lexer.is_eof() && lexer.peek() == '"') {
+                directive_value = consume_quoted_string(lexer);
```
shannonbooth, line 88: Hmm, by this stage have we guaranteed that we have valid UTF-8?

Tests/LibWeb/Text/input/include.js
```diff
@@ -138,3 +138,9 @@ function httpTestServer() {
     }
     return __httpTestServer;
 }
+
+// Per-call unique loopback host, so tests that mutate global per-host state
+// (e.g. HSTS) don't collide under the parallel runner or --repeat clones.
+function uniqueLocalhostHostname(prefix) {
```
shannonbooth, line 144: This is going to be very useful! One thing I am thinking about, and it might not make any sense - does it simplify things at all (or even possible) if we load _all_ tests ran over HTTP from a uniqueLocalHostName as the base URL? Or maybe we should be doing that anyway so that no child httpTestServer conflicts with another test process? 🤔

Libraries/LibHTTP/HSTS/ParsedHSTSPolicy.cpp
Lubrsi: I've combined them with a new helper: `is_token`

Libraries/LibHTTP/HSTS/ParsedHSTSPolicy.h
Lubrsi: Done, however the grammar doesn't accept negative values anyway:
```
max-age-value = delta-seconds
delta-seconds = <1*DIGIT, defined in [[RFC2616], Section 3.3.2](https://www.rfc-editor.org/info/rfc2616/#section-3.3.2)>
DIGIT          = <any US-ASCII digit "0".."9">
```

Libraries/LibWebView/HSTSStore.cpp
Lubrsi: Done.

Libraries/LibHTTP/HSTS/ParsedHSTSPolicy.cpp
Lubrsi: Not quite, collect_an_http_quoted_string tolerates slight malformations, particularly missing an end quote or a character after a backslash at the end of the input.

Libraries/LibHTTP/HSTS/ParsedHSTSPolicy.cpp
Lubrsi: I've used HTTP_TAB_OR_SPACE instead, since it seems a bit more accurate. Let me know what you think.

Libraries/LibHTTP/HSTS/ParsedHSTSPolicy.cpp
Lubrsi: Yes, if it's a token, but for quoted strings it now fails the parse if it's not valid UTF-8. Let me know what you think.

Tests/LibWeb/Text/input/include.js
Lubrsi: I think that would be nice to have, but I don't know if it really fits the scope of this PR, particularly if it was to cause any unrelated fallout.

shannonbooth: Fair not sure either! Something I will try and play around with after. Might mean less per test setup, but I don't think it's important at this stage

shannonbooth (approved):

shannonbooth:
MacOS CI seems to be unhappy: https://github.com/LadybirdBrowser/ladybird/actions/runs/26575970282/job/78375705705?pr=9475

Lubrsi:
I've added a commit that makes LibDNS implement RFC 6761 which essentially allows localhost and subdomains of localhost to resolve to loopback, instead of relying on the system resolver. It worked on my local MacBook with the system resolver, but doesn't seem to on CI macOS for some reason.

PR #9480, opened 2026-05-18
UI: Add QT, GTK and Appkit UI for permissions

NiccoloAntonelliDziri:
Connects permissions.cpp to the UI with IPC.

I cannot test the UI on macOS, so I'll appreciate if someone can try if it works correctly.
I usually go on https://permission.site/ and click on the location button, or on google maps.
Both buttons needs to work and resizing the window must not break the permission UI.

Here is what it looks like currently in the Qt and Gtk UIs:

permission-ladybird-gtk

permission-ladybird-qt-2

### DIFF
--- a/Libraries/LibWeb/PermissionsAPI/Permissions.cpp
+++ b/Libraries/LibWeb/PermissionsAPI/Permissions.cpp
@@ -12,6 +12,7 @@
 #include <LibWeb/HTML/Scripting/Environments.h>
 #include <LibWeb/HTML/Scripting/TemporaryExecutionContext.h>
 #include <LibWeb/HTML/Window.h>
+#include <LibWeb/Page/Page.h>
 #include <LibWeb/PermissionsAPI/PermissionStatus.h>
 #include <LibWeb/PermissionsAPI/PermissionStore.h>
 #include <LibWeb/PermissionsAPI/Permissions.h>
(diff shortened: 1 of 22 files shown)


Libraries/LibWeb/PermissionsAPI/Permissions.cpp
```diff
@@ -38,15 +39,13 @@ Bindings::PermissionState request_permission(Bindings::PermissionDescriptor cons
     if (current_state != Bindings::PermissionState::Prompt)
         return current_state;
 
-    // FIXME: 3. Ask the user for express permission for the calling algorithm to use the powerful feature described by descriptor.
+    // 3. Ask the user for express permission for the calling algorithm to use the powerful feature described by descriptor.
+    auto& window = as<HTML::Window>(HTML::current_global_object());
+    auto state = window.page().did_request_permission(descriptor.name);
```
shannonbooth: Not a full review by any means since I'm not a GUI person - but as a comment on this part, it would very much be best to not perform a synchronous IPC call, we try and avoid those as much as possible - especially in situations like this where it seems like this is blocking javascript and other stuff on the main thread from running until a dialog is pressed. We should be able to make this asynchronous since this is invoked by the specification in an "In parallel" block - which basically means we can fire this request off and resume that algorithm when we get some response

NiccoloAntonelliDziri: Oh ok, noted, I will change that

gmta:

PR #9492, opened 2026-05-19
LibWeb: Remove some uses of `LRC::for_layout_node`

Calme1709:
Relative lengths should always be absolutized at style computation time so we have no need for `Length::ResolutionContext::for_layout_node` - this gets us a bit closer to removing it entirely. The remaining users (`TransformationStyleValue`, `ColorResolutionContext`, and `LengthPercentage::resolved`) are a bit more complicated and will get their own PRs

### DIFF
--- a/Libraries/LibWeb/CSS/ComputedValues.h
+++ b/Libraries/LibWeb/CSS/ComputedValues.h
@@ -267,7 +267,6 @@ public:
     static BorderCollapse border_collapse() { return BorderCollapse::Separate; }
     static EmptyCells empty_cells() { return EmptyCells::Show; }
     static GridTemplateAreas grid_template_areas() { return {}; }
-    static Time transition_delay() { return Time::make_seconds(0); }
     static ObjectFit object_fit() { return ObjectFit::Fill; }
     static Position object_position() { return {}; }
     static Color outline_color() { return Color::Black; }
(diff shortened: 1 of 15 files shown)


Tests/LibWeb/Layout/expected/flex/columns-33-percent-width.txt
```diff
@@ -1,16 +1,16 @@
...
-  PaintableWithLines (BlockContainer<HTML>) [0,0 800x116]
-    PaintableBox (Box<BODY>) [8,8 500x100]
-      PaintableWithLines (BlockContainer<DIV>) [8,8 166.65625x100]
-      PaintableWithLines (BlockContainer<DIV>) [174.65625,8 166.65625x100]
-      PaintableWithLines (BlockContainer<DIV>) [341.3125,8 166.65625x100]
+  PaintableWithLines (BlockContainer<HTML>) [0,0 800x216]
+    PaintableBox (Box<BODY>) [8,8 500x200]
+      PaintableWithLines (BlockContainer<DIV>) [8,8 166.671875x100]
+      PaintableWithLines (BlockContainer<DIV>) [174.671875,8 166.671875x100]
+      PaintableWithLines (BlockContainer<DIV>) [8,108 166.671875x100]
```
gmta: 🤔 this seems like a regression?

Calme1709: Whoops - have dropped that commit and left this as FIXME

gmta: Ah that's a shame - seemed like a decent improvement on all other fronts!

AtkinsSJ (approved):

PR #9497, opened 2026-05-19
LibWeb: Stop using VideoFrameSource for video frame updates

kalenikaliaksandr:
Future compositor-process work needs to push video frame updates without going through VideoFrameSource. That object cannot be shared through IPC, so video display-list resources now use stable VideoFrameResourceId values and the current frame is sent through explicit resource/update commands.


Libraries/LibWeb/HTML/HTMLMediaElement.cpp
```diff
@@ -2008,6 +2024,7 @@ void HTMLMediaElement::forget_media_resource_specific_tracks()
     m_audio_tracks->remove_all_tracks();
     m_video_tracks->remove_all_tracks();
     m_playback_manager.clear();
+    clear_compositor_video_frame();
```
Zaggy1024, line 2027: I think this should already be taken care of by the removal of the video track?

PR #9508, opened 2026-05-20
LibWeb/CSS: Support oblique angles in font-style for @font-face

Martin005:
The `@font-face` `font-style` descriptor grammar now accepts up to two angles for the oblique keyword (e.g. "oblique 0deg 10deg"). 

Fixes following WPT tests:
```md
css/css-fonts/font-face-style-normal.html
CSS Fonts: parsing 'normal' in the font-style descriptor 2	❌ -> ✅
CSS Fonts: parsing 'normal' in the font-style descriptor 3	❌ -> ✅

css/css-fonts/variations/at-font-face-descriptors.html
font-style(valid): 'oblique' followed by angle in radians: oblique 0rad					❌ -> ✅
font-style(valid): Simple range: oblique 10deg 20deg									❌ -> ✅
font-style(valid): Simple range with equal upper and lower bounds: oblique 10deg 10deg	❌ -> ✅
font-style(valid): Simple range with equal upper and lower bounds: oblique 0deg 0deg	❌ -> ✅
font-style(valid): Bounds out of order: oblique 20deg 10deg								❌ -> ✅
```

### DIFF
--- a/Libraries/LibWeb/CSS/CSSDescriptors.cpp
+++ b/Libraries/LibWeb/CSS/CSSDescriptors.cpp
@@ -5,10 +5,13 @@
  */
 
 #include <LibWeb/CSS/CSSDescriptors.h>
+#include <LibWeb/CSS/Enums.h>
 #include <LibWeb/CSS/Parser/Parser.h>
 #include <LibWeb/CSS/PropertyID.h>
 #include <LibWeb/CSS/Serialize.h>
+#include <LibWeb/CSS/StyleValues/FontStyleStyleValue.h>
 #include <LibWeb/CSS/StyleValues/ShorthandStyleValue.h>
+#include <LibWeb/CSS/StyleValues/StyleValueList.h>
 #include <LibWeb/Infra/Strings.h>
 #include <LibWeb/WebIDL/ExceptionOr.h>
 
@@ -95,6 +98,28 @@ WebIDL::ExceptionOr<void> CSSDescriptors::set_property(FlyString const& property
     // 6. If component value list is null, then return.
     if (!component_value_list)
         return {};
+
+    // https://drafts.csswg.org/css-fonts-4/#descdef-font-face-font-style
+    // Normalize font-style StyleValueList to FontStyleStyleValue so serialization
+    // uses the specialized path (oblique 0deg → "normal", equal-bound collapsing, etc.)
+    if (descriptor_name_and_id->id() == DescriptorID::FontStyle && component_value_list->is_value_list()) {
+        auto const& values = component_value_list->as_value_list().values();
+        if (values.size() >= 1 && values[0]->is_keyword()) {
+            auto keyword = keyword_to_font_style_keyword(values[0]->to_keyword());
+            if (keyword.has_value()) {
+                ValueComparingRefPtr<StyleValue const> angle1;
+                ValueComparingRefPtr<StyleValue const> angle2;
+                if (values.size() > 1 && values[1]->is_value_list()) {
+                    auto const& inner = values[1]->as_value_list().values();
+                    if (inner.size() > 0 && !inner[0]->is_empty_optional())
+                        angle1 = inner[0];
+                    if (inner.size() > 1 && !inner[1]->is_empty_optional())
+                        angle2 = inner[1];
+                }
+                component_value_list = FontStyleStyleValue::create(keyword.release_value(), move(angle1), move(angle2));
+            }
+        }
+    }
 
     // 7. Let updated be false.
     auto updated = false;
--- a/Libraries/LibWeb/CSS/Descriptors.json
+++ b/Libraries/LibWeb/CSS/Descriptors.json
@@ -56,10 +56,13 @@
       },
       "font-style": {
         "initial": "auto",
-        "FIXME": "Support angles for oblique",
         "syntax": [
           "auto",
-          "<'font-style'>"
+          "normal",
+          "italic",
+          "left",
+          "right",
+          "<font-style-oblique>"
         ]
       },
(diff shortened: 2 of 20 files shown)


Libraries/LibWeb/CSS/Parser/ValueParsing.cpp
```diff
@@ -3570,6 +3570,11 @@ RefPtr<StyleValue const> Parser::parse_font_style_value(TokenStream<ComponentVal
 
     if (tokens.has_next_token() && keyword_value->to_keyword() == Keyword::Oblique) {
         if (auto angle_value = parse_angle_value(tokens, { .min = -90, .max = 90 })) {
+            // Try to parse a second angle for descriptor ranges (e.g., "oblique 0deg 10deg")
+            if (auto second_angle_value = parse_angle_value(tokens, { .min = -90, .max = 90 })) {
+                transaction.commit();
+                return FontStyleStyleValue::create(font_style.release_value(), angle_value, second_angle_value);
+            }
```
AtkinsSJ: Hi, this isn't quite right. This path is also used for the `font-style` property, which only accepts a single angle. `h2 { font-style: oblique 30deg 40deg; }` becomes valid incorrectly with this change.

Libraries/LibWeb/CSS/Descriptors.json
```diff
@@ -55,10 +55,9 @@
         "legacy-alias-for": "font-width"
       },
       "font-style": {
-        "initial": "auto",
-        "FIXME": "Support angles for oblique",
+        "initial": "normal",
         "syntax": [
-          "auto",
+          "normal",
           "<'font-style'>"
```
AtkinsSJ: If you look at the [spec](https://drafts.csswg.org/css-fonts-4/#font-prop-desc), having `<'font-style'>` here is a hack that we used for simplicity, but with multiple oblique angles we'll have to do something differently.

Also, that says the default should be `auto` - where are you getting `normal` from?

AtkinsSJ (changes requested):

Libraries/LibWeb/CSS/Descriptors.json
Martin005: @AtkinsSJ I was looking at the [wrong `font-style`](https://drafts.csswg.org/css-fonts-4/#font-style-prop) 🤦

Martin005:
Hi @AtkinsSJ! Could you please re-review this PR after my changes? I hope I didn't make any mistakes, but if I did, please let me know, and I will be happy to fix them 🙂

Meta/Utils/CSSGrammar/Parser/parser.py
```diff
@@ -76,6 +78,44 @@ def parse_component_value(self) -> GrammarNode:
...
+        # A comma-separated pair of numbers in curly braces ({A,B}) indicates that the preceding type, word, or group
+        # occurs at least A and at most B times. The B may be omitted ({A,}) to indicate that there must be
+        # at least A repetitions, with no upper bound on the number of repetitions.
+        if self.peek().is_token_type(TokenType.OPEN_CURLY_BRACKET):
+            self.consume()
+
+            if not self.peek().is_token_type(TokenType.COMPONENT_VALUE):
+                raise SyntaxError("CSSGrammar::Parser: Expected an integer in multiplier")
+            min_token = self.consume().component_value()
+            assert isinstance(min_token, Keyword)
```
AtkinsSJ, line 91: It's a bit weird to me that numbers are ending up as `Keyword`s. I suspect we should be doing something inside the tokenizer to handle these repetition groups? @Calme1709 you know this code much better than I do.

Meta/Utils/CSSGrammar/generator.py
```diff
@@ -158,6 +159,62 @@ def generate_css_parser_expression_for_optional_grammar_node(
...
+    out.write("};\n\n")
+
+    capacity = grammar_node.maximum if grammar_node.maximum is not None else grammar_node.minimum
+    out.write(f"""auto const parse_{cpp_name}_multiplier = [&]() -> RefPtr<StyleValue const> {{
+auto {cpp_name}_transaction = tokens.begin_transaction();
+StyleValueVector {cpp_name}_values;
+{cpp_name}_values.ensure_capacity({capacity});
+
+""")
+    # Minimum repetitions (required)
```
AtkinsSJ, line 178: I feel like all 3 cases could be handled with a single generated loop instead of needing separate code generation for them?

Libraries/LibWeb/CSS/CSSDescriptors.cpp
```diff
@@ -96,6 +99,28 @@ WebIDL::ExceptionOr<void> CSSDescriptors::set_property(FlyString const& property
     if (!component_value_list)
         return {};
 
+    // https://drafts.csswg.org/css-fonts-4/#descdef-font-face-font-style
+    // Normalize font-style StyleValueList to FontStyleStyleValue so serialization
+    // uses the specialized path (oblique 0deg → "normal", equal-bound collapsing, etc.)
+    if (descriptor_name_and_id->id() == DescriptorID::FontStyle && component_value_list->is_value_list()) {
```
AtkinsSJ, line 105: I think ideally we would form the correct StyleValue inside the Parser, instead of having to then adapt it here and in `StyleComputer::compute_font_style()`. Again I think @Calme1709 probably has thoughts on that.

AtkinsSJ (commented):
This is looking a lot better, thanks for that. My comments are basically not being sure about where the code should go. 😅 The existing code is mostly by Callum so I'd wait to see what he says before making any changes - I don't know if he already had thought about how this should work.

gmta:

PR #9512, opened 2026-05-20
LibWeb: Make `TransformationSV::to_matrix` infallible

Calme1709:
And remove another user of `Length::ResolutionContext::for_layout_node`. See individual commits for details.


AtkinsSJ (approved):
Nice! That's been annoying me for a while.

PR #9517, opened 2026-05-20
UI/AppKit: Add New Window menu action

awesomekling:
Open the configured new tab page in a separate AppKit window from the File menu, matching the existing Qt action.

### DIFF
--- a/UI/AppKit/Application/ApplicationDelegate.mm
+++ b/UI/AppKit/Application/ApplicationDelegate.mm
@@ -193,6 +193,13 @@
     [controller focusLocationToolbarItem];
 }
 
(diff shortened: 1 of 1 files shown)


UI/AppKit/Application/ApplicationDelegate.mm
```diff
@@ -193,6 +193,15 @@ - (void)openLocation:(id)sender
     [controller focusLocationToolbarItem];
 }
 
+- (void)createNewWindow:(id)sender
+{
+    (void)sender;
+
+    (void)[self createNewTab:WebView::Application::settings().new_tab_page_url()
```
trflynn89: nit: no need for either `(void)`

PR #9519, opened 2026-05-20
LibWeb: Support CSS transitions and animations on all pseudo-elements

Darshanx256:
Whitelisted pseudo-elements (`::placeholder`, `::first-letter`, `::first-line`, and `::selection`) rejected all transition and animation properties, preventing transitions from working on them.

### Problem

A common pattern like this produces an instant snap instead of a smooth fade in Ladybird, despite working in other engines:

```css
input::placeholder {
  opacity: 1;
  transition: opacity 0.5s ease;
}
input.fade::placeholder {
  opacity: 0;
}
```

This is because `pseudo_element_supports_property()` rejected transition/animation properties during cascade filtering, causing durations/delays to resolve to `0s` and bypass registration.

### Fix

This PR modifies `generate_libweb_css_pseudo_element.py` to insert a fast-path check in `pseudo_element_supports_property()`, automatically returning `true` for all `transition-*` and `animation-*` properties on any styleable pseudo-element.

### Verification

- Live test case: https://placeholderopacity.netlify.app/
- Verified working on: Chrome (Blink), Firefox (Gecko), Epiphany(WebKitGTK), and Ladybird (LibWeb with this patch applied).

### DIFF
--- a/Meta/Generators/generate_libweb_css_pseudo_element.py
+++ b/Meta/Generators/generate_libweb_css_pseudo_element.py
@@ -446,6 +446,26 @@ bool is_pseudo_element_root(PseudoElement pseudo_element)
 
 bool pseudo_element_supports_property(PseudoElement pseudo_element, PropertyID property_id)
 {
+    if (property_id == PropertyID::Transition
+        || property_id == PropertyID::TransitionBehavior
+        || property_id == PropertyID::TransitionDelay
+        || property_id == PropertyID::TransitionDuration
+        || property_id == PropertyID::TransitionProperty
+        || property_id == PropertyID::TransitionTimingFunction
+        || property_id == PropertyID::Animation
+        || property_id == PropertyID::AnimationComposition
+        || property_id == PropertyID::AnimationDelay
+        || property_id == PropertyID::AnimationDirection
+        || property_id == PropertyID::AnimationDuration
+        || property_id == PropertyID::AnimationFillMode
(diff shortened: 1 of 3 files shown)


AtkinsSJ:
That second commit looks like it's going to conflict badly with #9443... and maybe will be fixed by that? cc: @Calme1709

First commit is a very nice catch though, thanks for that! I think I'm going to see if we can get that other PR merged first though as it's a bit of a conflict magnet. Then we can see if the second commit here is still necessary.

Calme1709:
Correct - only the first commit is needed once #9443 is merged (at least to run the included test)

Darshanx256:
Thanks for the feedback!

I've verified locally that once #9443 is merged, the second commit in this PR is indeed redundant. I've updated the branch to drop the second commit, and the first commit along with #9443 is sufficient to fully support transitions on element-backed pseudo-elements like ::placeholder. @AtkinsSJ

AtkinsSJ:
Great! Now that #9443 is merged, can you add your test again? Other than that this is looking good.

Darshanx256:
> Great! Now that #9443 is merged, can you add your test again? Other than that this is looking good.

Done! 
Test added in Tests/LibWeb/Text/input/css-placeholder-transition.html. It verifies that opacity, color, and background-color transitions on ::placeholder start correctly rather than snapping to their final values instantly.

AtkinsSJ (approved):
Thanks!

PR #9528, opened 2026-05-21
LibWeb: Fix -webkit-linear-gradient angle direction

bdr99:
In the `linear-gradient` CSS property, angles specified numerically in degrees are measured clockwise relative to north, but in `-webkit-linear-gradient`, they are counter-clockwise relative to east. There was no logic to account for this difference, so styles using `-webkit-linear-gradient` with a numeric angle were being rendered with the wrong angle.  This PR adds a conversion step to account for this, and updates an existing test to validate the fix.

This is distinct from the existing special case for `-webkit-linear-gradient` in the `SideOrCorner` branch immediately above where I made this change. That special case is applicable to angles specified as `top`, `left`, etc, not numeric angles.


gmta (approved):

gmta:
Thanks @bdr99 , and welcome to the project!

PR #9530, opened 2026-05-21
LibWeb: Propagate inherited style across shadow and slot boundaries

awesomekling:
Style invalidation now carries an explicit inherited-style signal when inherited longhands change, including inheritance-dependent specified values. That signal is used when crossing from shadow hosts into shadow roots, including for explicit `inherit` of normally non-inherited host properties.

Slot inherited-style invalidation now reaches assigned slottables even when they are dirtied after their traversal point, avoids repeated descendant-slot scans, and includes animation updates on shadow hosts and slots. This fixes stale inherited styles in nested slotted shadow trees, including popover/menu style changes like Reddit's sort menu.

The patch adds regression coverage for nested slotted inheritance, shadow-root inherited updates, slot-boundary invalidation, animation propagation, descendant-slot scan counts, and custom-element `:defined` invalidation.

This makes it possible to pop these lil guys for subreddit sorting up on Reddit:

Screenshot 2026-05-21 at 09 10 28


Libraries/LibWeb/DOM/Node.cpp
```diff
@@ -1773,6 +1773,19 @@ void Node::set_needs_style_update(bool value)
...
+{
+    if (m_subtree_may_depend_on_non_inherited_property_inheritance)
+        return;
+    m_subtree_may_depend_on_non_inherited_property_inheritance = true;
+
+    for (auto* ancestor = parent_or_shadow_host(); ancestor; ancestor = ancestor->parent_or_shadow_host()) {
+        if (ancestor->m_subtree_may_depend_on_non_inherited_property_inheritance)
+            break;
+        ancestor->m_subtree_may_depend_on_non_inherited_property_inheritance = true;
+    }
```
AtkinsSJ: Do we need to flag all ancestors like this? As far as I'm aware, you can only inherit from a direct parent. (Maybe `::part()` and `::slotted()` muddy that...)

PR #9532, opened 2026-05-21
LibMedia: Make PlaybackStream error clearer

ShehabAhmed-Gui:
It took me sometime to figure out why there was no audio and error message wasn't clear.

This provides clearer error message when there is no audio backend installed.


Zaggy1024:
Please don't close PRs to amend them. This almost looks good, the second line of the commit message is too long.

PR #9544, opened 2026-05-22
LibWeb: Resolve focused element style without a full document flush

awesomekling:
This avoids forcing full document style updates from focusability checks by adding a targeted `update_style_for_element()` path that resolves the flat-tree inheritance chain needed for one element.

The branch is split into three commits:

- Add tests that document the stale focus and `getComputedStyle()` cases.
- Share the style invalidation side effects from the normal style traversal.
- Add targeted style updates and use `StopAtDisplayNone` for focusability checks.

The targeted path still falls back to the normal document style traversal for document-level invalidation, full rebuilds, and selector invalidation work, so global dirty state is consumed atomically. Local recomputation preserves layout, display-list, slot, visual-context, and stacking-context invalidation, and marks descendants dirty when a recomputed element can affect their style.


shlyakpavel:
I think this regresses focusing detached elements.

With this PR, `meets_focusable_area_rendering_requirements()` returns true when an element has no current computed style. `document().update_style()` in `run_focusing_steps()` does not create computed style for detached nodes, so a detached input can become focusable.

Repro:

```html
<!DOCTYPE html>
<script>
const visible = document.createElement("input");
visible.id = "visible";
document.body.append(visible);
visible.focus();

const detached = document.createElement("input");
detached.id = "detached";
detached.focus();

console.log(document.activeElement.id);
// Expected: visible
// With this PR: empty string
</script>
```

Could we keep the no-flush behavior but preserve the old “no computed style means not rendered/focusable” check?

awesomekling:
> I think this regresses focusing detached elements.

Duh, of course. I'll get back to this tomorrow :)

PR #9553, opened 2026-05-22
LibWeb: Reject safe/unsafe as standalone self-alignment values

mikiubo:


AtkinsSJ (approved):
Looks like we need to be smarter about these properties, to allow multiple keywords instead of just a single one, but this is definitely an improvement.

PR #9559, opened 2026-05-22
LibWeb: Fix a crash when Element::computed_properties returns no value

cqundefine:
This occurs right after login in to Jellyfin.

### DIFF
--- a/Libraries/LibWeb/Layout/TreeBuilder.cpp
+++ b/Libraries/LibWeb/Layout/TreeBuilder.cpp
@@ -889,6 +889,8 @@ void TreeBuilder::update_layout_tree(DOM::Node& dom_node, TreeBuilder::Context& 
                 element.set_needs_style_update(false);
             }
             style = element.computed_properties();
+            if (!style)
+                return;
             display = style->display();
             if (display.is_none())
                 return;


Psychpsyo:
Shouldn't the if-statement right above the line you added take care of this? What's going on there?
Also, a crash test would be nice.

Lubrsi (commented):
This block above makes me think that the bug is in the fact `update_style_for_element` isn't calculating new style for the element for some reason, since it's running `update_style_for_element` if it needs a style update _or_ there's no computed properties, and it's expected that this block produces computed_properties on the element.

Libraries/LibWeb/Layout/TreeBuilder.cpp
```diff
@@ -889,6 +889,8 @@ void TreeBuilder::update_layout_tree(DOM::Node& dom_node, TreeBuilder::Context&
                 element.set_needs_style_update(false);
             }
```
Lubrsi, line 890: Sorry, this block here.

InvalidUsernameException:
What is the stack trace on this crash? It sounds like this might be the same bug as #9526.

tcl3:
> It sounds like this might be the same bug as https://github.com/LadybirdBrowser/ladybird/issues/9526.

I tried logging in to the Jellyfin demo server at https://demo.jellyfin.org/stable (username:demo, no password) - it does see, to be the same crash as #9526 and it no longer occurs with #9570 applied.

gmta:

PR #9562, opened 2026-05-22
Tests: Add a LibDevTools test suite

AtkinsSJ:
Far from exhaustive, but this should be enough coverage to prevent basic regressions, and allow our robot chums to check their work somewhat, instead of relying on human verification.

### DIFF
--- a/Tests/CMakeLists.txt
+++ b/Tests/CMakeLists.txt
@@ -21,6 +21,7 @@ add_subdirectory(LibXML)
 add_subdirectory(Meta)
 
 if (ENABLE_GUI_TARGETS)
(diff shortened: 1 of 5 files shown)


Tests/CMakeLists.txt
```diff
@@ -23,6 +23,7 @@ add_subdirectory(Meta)
 if (ENABLE_GUI_TARGETS)
     add_subdirectory(LibGfx)
     add_subdirectory(LibMedia)
+    add_subdirectory(LibDevTools)
```
AtkinsSJ: Just noticed Codex put this here for some reason. I'll fix that later.

PR #9565, opened 2026-05-22
LibWeb: Preserve paint cache for transform-only animation updates

herzblutnord:
Previously, transform-only animation updates would cause the paint cache to be invalidated on every tick, resulting in a full repaint of all unchanged content in the affected subtree.  Now, cached paint commands are preserved when an animated property update only impacts the accumulated visual context (e.g. transform, opacity). The accumulated visual context is still updated.

This eliminates per-tick repaints of large unchanged subtrees . A visible case is long text runs animated with transform: translateX(...) that used to cause page-wide jank on each frame.

Testing
Tested manually on https://melankorin.net/, https://herz.moe/ (websites with animated text with transform: translateX(...)). This change fixed the strong visible jank on both sites .

All tests in ./Meta/[ladybird.py](http://ladybird.py/) test LibWeb suite pass.

## Before:
https://github.com/user-attachments/assets/0befdaf7-a0dd-40f8-b03c-f2b9be552698

## After:

https://github.com/user-attachments/assets/94d5dfa2-d7ab-4549-96d4-1a8e85f5cf2e

### DIFF
--- a/Libraries/LibWeb/Animations/AnimationEffect.cpp
+++ b/Libraries/LibWeb/Animations/AnimationEffect.cpp
@@ -820,11 +820,7 @@ static bool animated_property_changes_are_visual_context_only(HashMap<CSS::Prope
             CSS::PropertyID::Rotate,
             CSS::PropertyID::Scale,
             CSS::PropertyID::Translate,
-            CSS::PropertyID::Perspective,
-            CSS::PropertyID::TransformOrigin,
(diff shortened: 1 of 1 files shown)


herzblutnord:
I narrowed this to transform, rotate, scale, translate, and opacity because Ladybird applies these after content painting through the accumulated visual context rather than during subtree painting.

gmta:

PR #9566, opened 2026-05-22
Meta: Use baseline CPU for fuzzer host tools

shannonbooth:
The fuzzer CI build first builds and installs LagomTools, then imports those tools into the fuzzer build. This host-tools configure path skipped ENABLE_CI_BASELINE_CPU, so tools such as BindingsGenerator could be built with -march=native.

That is unsafe with shared CI caches and mixed runner CPUs, since a cached host tool may contain instructions unsupported by the runner that later executes it.

Pass ENABLE_CI_BASELINE_CPU when configuring fuzzer host tools.

### DIFF
--- a/.github/workflows/lagom-template.yml
+++ b/.github/workflows/lagom-template.yml
@@ -155,6 +155,7 @@ jobs:
           VCPKG_CACHE_MODE: ${{ github.ref == 'refs/heads/master' && 'write' || '' }}
         run: |
           cmake --preset=Host_Tools -B "${GITHUB_WORKSPACE}/Build/host-tools-build" \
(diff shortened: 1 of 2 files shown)


.github/workflows/lagom-template.yml
Lubrsi, line 164: I think we should also do it for this cmake invocation :^)

shannonbooth: Maybe we go with your change? It looks better than mine, just without the comment update in cmakelists

Lubrsi: Sure!

PR #9567, opened 2026-05-22
CI: Pass CMake options in the Fuzzers build

Lubrsi:
We were forgetting to pass CMake options in the Fuzzers build. These CMake options contain -DENABLE_CI_BASELINE_CPU=ON, which sets a baseline architecture to properly allow ccache sharing between different runners. Without it, it defaults to the native architecture, which is then cached and can be used on a runner that doesn't have the same instruction set.

This is targetted at fixing the "Illegal instruction" crash we've been seeing with the bindings generator.


Lubrsi:
Oops, ninja'd by https://github.com/LadybirdBrowser/ladybird/pull/9566

PR #9577, opened 2026-05-23
Visual splitting for fragmented boxes inside multicol containers! :D

Psychpsyo:
This builds upon #8557.
It moves the `FragmentationState` enum from #8761 over into Fragmentation.cpp, since it is better housed there and doesn't cause any circular include problems like that.

Apart from that, here's an image that I was way too excited about at 3am yesterday:
image
That's this HTML:
```html
<!DOCTYPE html>
<style>
    body {
        height: 3em;
        column-width: 10em;
    }
    div {
        border: 2px solid;
        background: lightgray;
    }
</style>
In the traditional CSS box model, the content of an element is flowed into the content box of the corresponding element.
<div>
    Multi-column layout introduces a fragmentation context formed of anonymous fragmentation containers called column boxes (or columns for short).
</div>
These column boxes establish an independent block formatting context into which the multi-column container’s content flows, and form the containing block for its non-positioned children.
```

### DIFF
--- a/Libraries/LibWeb/CMakeLists.txt
+++ b/Libraries/LibWeb/CMakeLists.txt
@@ -768,6 +768,7 @@ set(SOURCES
     Layout/FieldSetBox.cpp
     Layout/FlexFormattingContext.cpp
     Layout/FormattingContext.cpp
+    Layout/Fragmentation.cpp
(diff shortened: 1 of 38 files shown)


gmta:

PR #9580, opened 2026-05-23
LibWeb+Tests: Upgrade custom elements in ShadowRoot innerHTML

olliehcrook:
Use the shadow root's custom element registry for parsed fragments.

This fixes the failing detached ShadowRoot innerHTML WPT case where the custom element was not upgraded before the CEReactions queue was popped.

Import the upstream WPT and rebaseline existing custom element registry coverage.

### DIFF
--- a/Libraries/LibWeb/DOM/ShadowRoot.cpp
+++ b/Libraries/LibWeb/DOM/ShadowRoot.cpp
@@ -120,6 +120,14 @@ WebIDL::ExceptionOr<void> ShadowRoot::set_inner_html(TrustedTypes::TrustedHTMLOr
     // 3. Let fragment be the result of invoking the fragment parsing algorithm steps with context and compliantString.
     auto fragment = TRY(context->parse_fragment(compliant_string.to_utf8_but_should_be_ported_to_utf16()));
 
+    if (auto registry = custom_element_registry()) {
+        fragment->for_each_in_subtree_of_type<Element>([&](auto& element) {
+            element.set_custom_element_registry(registry);
+            element.try_to_upgrade();
+            return TraversalDecision::Continue;
+        });
+    }
+
     // 4. Replace all with fragment within this.
     this->replace_all(fragment);
 
--- a/Tests/LibWeb/Text/expected/wpt-import/custom-elements/reactions/ShadowRoot.txt
+++ b/Tests/LibWeb/Text/expected/wpt-import/custom-elements/reactions/ShadowRoot.txt
@@ -1,0 +1,8 @@
+Harness status: OK
+
+Found 3 tests
+
+3 Pass
+Pass	innerHTML on ShadowRoot must upgrade a custom element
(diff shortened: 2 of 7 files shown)


Tests/LibWeb/Text/expected/wpt-import/custom-elements/reactions/ShadowRoot.txt
```diff
@@ -0,0 +1,8 @@
+Harness status: OK
+
+Found 3 tests
+
+3 Pass
+Pass	innerHTML on ShadowRoot must upgrade a custom element
```
olliehcrook, line 6: This was the failing test case I was looking to fix, old results [here](https://wpt.fyi/results/custom-elements/reactions/ShadowRoot.html?run_id=5103794450071552)

Tests/LibWeb/Text/expected/wpt-import/custom-elements/registries/scoped-registry-define-upgrade-criteria.txt
```diff
@@ -2,18 +2,17 @@ Harness status: OK
...
-7 Fail
+14 Pass
 Pass	Adding definition to scoped registry should upgrade nodes associated with the registry.
-Fail	Adding definition to global registry should not affect shadow roots using scoped registry
+Pass	Adding definition to global registry should not affect shadow roots using scoped registry
 Pass	Adding definition to global registry should affect shadow roots also using global registry
-Fail	Adding definition to scoped registry should affect all associated shadow roots
-Fail	Adding definition to scoped registry should not affect document tree scope
-Fail	Adding definition to scoped registry should not affect shadow roots using other registries
+Pass	Adding definition to scoped registry should affect all associated shadow roots
```
olliehcrook, line 9: Seems I also inadvertently fixed a fair amount of other failing testcases, yay :)

Libraries/LibWeb/DOM/ShadowRoot.cpp
```diff
@@ -120,6 +120,14 @@ WebIDL::ExceptionOr<void> ShadowRoot::set_inner_html(TrustedTypes::TrustedHTMLOr
     // 3. Let fragment be the result of invoking the fragment parsing algorithm steps with context and compliantString.
     auto fragment = TRY(context->parse_fragment(compliant_string.to_utf8_but_should_be_ported_to_utf16()));
 
+    if (auto registry = custom_element_registry()) {
```
Psychpsyo, line 123: The spec doesn't say to do this here. Are you sure this is the right place?

If yes: There should be a spec bug about this and it should be linked in an `// AD-HOC:` comment above this code.

If no: This should probably happen elsewhere.

olliehcrook: Yeah, I think you are right... this issue probably comes from the parser. I think I tunnel-visioned on getting the WPT to pass here.

olliehcrook: I will close this for now and take another go at it when I get a chance,

PR #9588, opened 2026-05-23
LibWeb/SVG: Ignore degenerate active viewBox ratios

shlyakpavel:
## Summary

Do not use an active SVG View's `viewBox` as an intrinsic aspect ratio when either the viewBox width or height is zero.

SVG intrinsic sizing says that when an SVG View is active, the intrinsic aspect ratio is computed from `viewbox.width / viewbox.height`:
https://www.w3.org/TR/SVG2/coords.html#SizingSVGInCSS

A zero `viewBox` width or height disables rendering of the element, so it should not produce a usable intrinsic aspect ratio:
https://www.w3.org/TR/SVG2/coords.html#ViewBoxAttribute

The normal `<svg viewBox>` path already requires both dimensions to be non-zero. This makes the active SVG View path match that behavior.

## Repro

```html
<!doctype html>
<style>
svg {
    display: block;
    width: 60px;
    height: auto;
}
</style>
<svg id="target">
    <view id="normal" viewBox="0 0 10 5"></view>
    <view id="bad-height" viewBox="0 0 10 0"></view>
    <view id="bad-width" viewBox="0 0 0 10"></view>
</svg>
<p id="result"></p>
<script>
window.addEventListener("load", () => {
    result.textContent = `${location.hash}: ${getComputedStyle(target).height}`;
});
</script>

### DIFF
--- a/Libraries/LibWeb/SVG/SVGSVGElement.cpp
+++ b/Libraries/LibWeb/SVG/SVGSVGElement.cpp
@@ -306,7 +306,7 @@ CSS::SizeWithAspectRatio SVGSVGElement::negotiate_natural_metrics(SVG::SVGSVGEle
             auto view_box = active_view_element->view_box().value();
 
             // 2. return viewbox.width / viewbox.height
-            if (view_box.width != 0 || view_box.height != 0)
+            if (view_box.width != 0 && view_box.height != 0)
                 return view_box.width / view_box.height;
 
             return {};


Psychpsyo:
Shouldn't the repro be added as a test of some sort?

Also, I find it odd that the spec doesn't say to do any of this. Makes me think they assume that the algorithm shouldn't be called in the first place for these cases. (Likely because the image isn't being rendered)
Might be something to take up with the spec authors over at https://github.com/w3c/svgwg/issues or at least to put an `// AD-HOC:` comment with reasoning about.

Maybe of interest: Reading properties like `height` from the resolved styles (what `getComputedStyle()` actually returns) already special-cases situations where `display` would prevent the element from showing up: https://www.w3.org/TR/cssom-1/#resolved-value-special-case-property-like-height

shlyakpavel:
@Psychpsyo I looked into adding this as a test, but I don't see a good minimal way to cover this in the current LibWeb text test setup. The repro depends on loading the SVG image with a fragment URL that activates the SVG view, so it is much easier to reproduce manually with the exact URL than as a small self-contained text test.

The change itself is intentionally small, and I see it as fixing a typo in the code: the active-view branch already mirrors the normal `viewBox` branch below, but used `||` instead of `&&` before dividing `width / height`. So this makes the active SVG view case use the same degenerate-viewBox guard as the regular `svg` `viewBox` case.

Psychpsyo:
> @Psychpsyo I looked into adding this as a test, but I don't see a good minimal way to cover this in the current LibWeb text test setup. The repro depends on loading the SVG image with a fragment URL that activates the SVG view, so it is much easier to reproduce manually with the exact URL than as a small self-contained text test.

In theory, if I'm reading things right, setting `location.hash` up-front should work. (apart from the part where ladybird seems to not live update the active view when that happens...)
Although fixing that as well might be out of scope here.

Also, I think I found the spec to cite for this:
https://www.w3.org/TR/css-images-3/#natural-dimensions
> If an [object](https://www.w3.org/TR/css-images-3/#objects) has a [degenerate](https://www.w3.org/TR/css-values-4/#degenerate-ratio) [natural aspect ratio](https://www.w3.org/TR/css-images-3/#natural-aspect-ratio) (at least one part being zero or infinity), it is treated as having no natural aspect ratio.

gmta:

PR #9589, opened 2026-05-23
Recover after Compositor process crashes

kalenikaliaksandr:
The browser previously treated the out-of-process Compositor as fatal.

Restart the shared Compositor from the browser process, reconnect process-backed WebContent clients, recreate compositor contexts, restore viewport state, and ask WebContent to repaint and republish canvas and media resources. WebContent now marks its compositor connection lost, returns conservative values for synchronous compositor queries while reconnecting, and drops outgoing updates until the replacement transport arrives.

Mouse events queued while the Compositor is unavailable now fall back to direct WebContent dispatch. This keeps input completion in step with the pending-event queue.

Recovery is capped at three automatic restarts. If the restart limit is exceeded, or if restart, reconnect, or context recreation fails, the browser crashes instead of switching process-backed views to a fallback path.

### DIFF
--- a/Libraries/LibWebView/Application.cpp
+++ b/Libraries/LibWebView/Application.cpp
@@ -6,6 +6,7 @@
 
 #include <AK/Checked.h>
 #include <AK/Debug.h>
+#include <AK/ScopeGuard.h>
 #include <AK/Time.h>
 #include <LibCore/AnonymousBuffer.h>
 #include <LibCore/ArgsParser.h>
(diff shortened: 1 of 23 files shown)


Libraries/LibWebView/Application.cpp
```diff
@@ -544,28 +571,33 @@ void Application::update_compositor_viewport(Web::Compositor::CompositorContextI
 
 bool Application::send_async_scroll_to_compositor(Web::Compositor::CompositorContextId context_id, Gfx::FloatPoint position, Gfx::FloatPoint delta_in_device_pixels)
 {
-    VERIFY(m_compositor_client);
+    if (!can_send_compositor_process_ipc(m_compositor_client))
+        return false;
+
     return m_compositor_client->async_scroll_by(context_id, position, delta_in_device_pixels);
```
shlyakpavel: Could this use the generated `try_async_scroll_by()` helper instead?

This already checks `is_open()`, but there still seems to be a small race if the Compositor process exits after that check, or while waiting for the sync reply. The generated `async_scroll_by()` helper goes through `send_sync()`, which uses `MUST(post_message(...))` / `VERIFY(response)` on failure.

`handle_mouse_event()` below looks similar and could maybe use `try_handle_mouse_event()` as well.

Libraries/LibWebView/Application.cpp
```diff
@@ -680,19 +712,87 @@ ErrorOr<void> Application::launch_services()
...
+    return m_browser_options.enable_compositor_process == EnableCompositorProcess::Yes;
+}
+
+void Application::handle_compositor_process_death()
+{
+    m_compositor_client = nullptr;
+
+    if (Core::EventLoop::current().was_exit_requested())
+        return;
+    if (m_is_recovering_compositor_process || m_has_queued_compositor_recovery)
```
shlyakpavel: What is expected to happen if the newly launched Compositor process exits while `recover_compositor_process()` is still running?

I might be missing something, but it looks like `handle_compositor_process_death()` clears `m_compositor_client` and then returns here while `m_is_recovering_compositor_process` is true, without queuing another recovery attempt. Could that leave the current recovery attempt to fail later, or finish without an active compositor client?

kalenikaliaksandr: yeah, you are right that path was ambiguous. if a Compositor death is observed while recovery is already in progress we now treat that as unrecoverable and crash the browser explicitly.

PR #9591, opened 2026-05-24
LibWeb/CSS: Remove redundant Optional wrapper around GC::Ptr

veeti:
The use of Optional here is causing a large build error on GCC 16.1.1. The easiest fix seems to be to just remove the Optional entirely.


veeti:
Actually, not so sure about this :D

AtkinsSJ:
The reason it's Optional is we have 3 states:
- No cache
- Cache that there isn't a scope ancestor
- Cache that there is a scope ancestor

If you remove the Optional, we end up searching for the non-existent ancestor each time it's requested instead of having that cached.

It's weird that GCC on CI doesn't mind this. Maybe there is something wrong about our `Optional<GC::Ptr>` implementation, but otherwise, avoiding that version of GCC  might be the answer. 😅

veeti:
BTW I use Arch... I think I have a proper fix at #9613 to keep it happy.

PR #9598, opened 2026-05-24
AK: Don’t let float precision sneak past is_within_range bounds

### LINKED ISSUE #6212: is_within_range returns true on barely out of bound floats
### Summary

When you call [is_within_range](https://github.com/LadybirdBrowser/ladybird/blob/7b3465ab55142e7e8b5a8261531ce87960bbceac/AK/Checked.h#L108) with a float that is 1 to large, casting to an int it returns true

### Operating system

Linux

### Steps to reproduce

When calling [is_within_range](https://github.com/LadybirdBrowser/ladybird/blob/7b3465ab55142e7e8b5a8261531ce87960bbceac/AK/Checked.h#L108) with a float `INT_MAX < x <= INT_MAX + 1` and converting to an int.

### Expected behavior

It should return `false` because it is above INT_MAX and it cannot be cast to an integer

### Actual behavior

It returns `true` because when casting the integer to a float to compare them it get incresed because of floating point imprecision  to INT_MAX + 1

N/A

```html
std::cout << AK::is_within_range<int>(2147483648.0) << "\n";
```

```shell
1
```

_No response_

_No response_

### Contribute a patch?

sideshowbarker:
**Problem:** `is_within_range<I>(F value)` — where `I` is an integer and `F` is a floating-point type — is unexpectedly too permissive in some cases:

a. Values that are 1 past the integer range unexpectedly pass; e.g., `is_within_range<int>(2147483648.0f)` returns `true` — even though `2147483648` is `INT_MAX + 1`.

b. Fractional values whose magnitude exceeds the destination max unexpectedly pass; e.g., `is_within_range<unsigned>(4294967295.5)` returns `true` — even though `4294967295.5 > UINT_MAX`.

c. Fractional values within the destination’s numeric range unexpectedly pass (e.g., `is_within_range<int>(2.5)` returns `true`) — even though they aren’t exactly representable as the destination type.

**Cause:** `TypeBoundsChecker` integer-bounds specializations compare against `NumericLimits<Destination>::max()` and `::min()` directly. When a caller’s value is a float, the integer max/min get implicitly converted to a float for the comparison. For Destination/Source pairs with the integer extreme not exactly representable in the float, that conversion rounds up to the next power-of-two boundary — so “`value <= F(max)`” accepts values that are actually out of range by one (case a). And the comparison itself doesn’t reject fractional values (cases b and c).

**Fix:** When `Source` is a floating-point type:

1. **First gate (case a)** — Compare against `2^digits`; exactly representable in any IEEE float, and equals `max + 1` for unsigned / `-min` for two’s-complement signed integers.

2. **Second gate (cases b and c)** — Round-trip check: cast value to `Destination`, then cast back — and require equality. Only integer-valued floats whose truncation matches the original pass.

Fixes https://github.com/LadybirdBrowser/ladybird/issues/6212

### DIFF
--- a/AK/Checked.h
+++ b/AK/Checked.h
@@ -45,12 +45,41 @@ struct TypeBoundsChecker<Destination, Source, false, false, false> {
     }
 };
 
+// Computes 2^N as a floating-point Source value. The result is exactly representable in any IEEE floating-point type
+// for any N within that type's exponent range — since 2^N has mantissa 1.0.
+template<FloatingPoint Source>
+static constexpr Source two_to_the(size_t n)
+{
+    Source result = 1;
+    for (size_t i = 0; i < n; ++i)
+        result *= 2;
+    return result;
+}
+
 template<typename Destination, typename Source>
 struct TypeBoundsChecker<Destination, Source, false, true, true> {
     static constexpr bool is_within_range(Source value)
     {
-        return value <= NumericLimits<Destination>::max()
-            && NumericLimits<Destination>::min() <= value;
+        if constexpr (IsFloatingPoint<Source>) {
+            // The value must lie in [Destination::min, Destination::max] *and* must be exactly representable as
+            // Destination; that is, the float must be integer-valued — with no fractional part.
+            //
+            // First gate: the value must fit in the safe-cast window — so static_cast<Destination> below is defined.
+            // Comparing directly against NumericLimits<Destination>::max() doesn't work when max isn't exactly
+            // representable in Source (e.g., INT_MAX = 2^31-1 in float rounds up to 2^31). Use 2^digits as a strict
+            // upper bound instead: exactly representable in any IEEE float — covering the range [-2^digits, 2^digits)
+            // for two's-complement signed destinations.
+            auto const boundary = two_to_the<Source>(NumericLimits<Destination>::digits());
+            if (!(value >= -boundary && value < boundary))
+                return false;
+            // Second gate: reject fractional values (round-trip check). Values in (-boundary, boundary) with a
+            // fractional part would truncate to an in-range integer — but aren't themselves within Destination's range.
+            auto const truncated = static_cast<Destination>(value);
+            return static_cast<Source>(truncated) == value;
+        } else {
+            return value <= NumericLimits<Destination>::max()
+                && NumericLimits<Destination>::min() <= value;
+        }
     }
 };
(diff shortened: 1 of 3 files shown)


AK/Checked.h
```diff
@@ -49,16 +49,39 @@ template<typename Destination, typename Source>
...
     static constexpr bool is_within_range(Source value)
     {
-        return value >= 0 && value <= NumericLimits<Destination>::max();
+        if constexpr (IsFloatingPoint<Source>) {
+            // See the signed case above for the precision concern. For unsigned destinations,
+            // NumericLimits<Destination>::max() + 1 is a power of two and exactly representable in any IEEE float.
+            // Casting max() to Source may already round up to that power of two — but adding 1 in Source then rounds
+            // back to the same value (the spacing between adjacent floats at that magnitude exceeds 1), giving the
+            // correct strict upper bound either way.
+            auto const upper_exclusive = static_cast<Source>(NumericLimits<Destination>::max()) + Source(1);
```
shannonbooth: Since casting to int can round values up, I believe this breaks for the situation such as:

```cpp
AK::is_within_range<unsigned>(4294967295.5)
```

I suppose a fix here might be to instead check if the source is representable or not before falling back to this path? (edit - NumericLimits<T>::digits seems appropriate, probably even for the general case?)

shannonbooth (changes requested):

AK/Checked.h
sideshowbarker: The code didn’t actually break for the cited case — but yeah anyway `digits()` is cleaner and better, so I went ahead and switched it to that. I also added tests for that case.

shannonbooth: Hmm, I might be thinking about it wrong, but shouldn't this assertion hold?

```cpp
static_assert(!AK::is_within_range<unsigned>(4294967295.5));
```

Since unsigned can hold max value of 4294967295 / is less than 4294967295.5?

sideshowbarker: hmm yeah you’re right. I‘ll work on further update.

sideshowbarker: OK, just now pushed another update that I _think_ should resolve this. (And as a bonus, happens to also resolve the existing _“Make is_within_range work with floating point numbers“_ FIXME in JsonValue.h)

shannonbooth (approved):

PR #9610, opened 2026-05-24
LibWeb+LibTextCodec: Fix a few WPT encoding test failures

Martin005:
Fixes some of the failing WPT subtests for the [`encoding`](https://wpt.fyi/results/encoding) test files:
- Set document encoding for XML documents (https://github.com/LadybirdBrowser/ladybird/commit/fbffcec17c4285e150cea6de9ea2da738f5a5d5f: 2 subtests)
- Implement step 6 of encoding sniffing for parent document inheritance (https://github.com/LadybirdBrowser/ladybird/commit/41c1ac1110fa77d129ee29f40996715ad68cd54b: 3 subtests)
- Support ignoreBOM in TextDecoder and TextDecoderStream (https://github.com/LadybirdBrowser/ladybird/commit/c947d45210e34d694cf4860814e8abdad7394bc2: 16 subtests)
- Handle U+FFFD as unmappable in ISO-2022-JP encoder (https://github.com/LadybirdBrowser/ladybird/commit/e519ebec2247dd7180fb230615cf460ba826d1ba: 4 subtests)


sideshowbarker (changes requested):
Please rebase this against current master. The CI runs are hitting timeouts for some stuff that’s subsequently been fixed on master.

Libraries/LibWeb/DOM/DocumentLoading.cpp
```diff
@@ -59,18 +59,23 @@ static void convert_to_xml_error_document(DOM::Document& document, Utf16String e
...
         // https://www.w3.org/TR/xml/#charencoding
         // [...] it is a fatal error [...] for an entity which begins with neither a Byte Order Mark nor an encoding
         // declaration to use an encoding other than UTF-8.
         auto bom_encoding = HTML::run_bom_sniff(data);
-        decoder = TextCodec::decoder_for(bom_encoding.value_or("UTF-8"));
+        document_encoding = bom_encoding.value_or("UTF-8");
+        decoder = TextCodec::decoder_for(document_encoding);
     }
     VERIFY(decoder.has_value());
+    document.set_encoding(MUST(String::from_byte_string(document_encoding)));
```
sideshowbarker:
```suggestion
    auto standardized_encoding = TextCodec::get_standardized_encoding(document_encoding);
    VERIFY(standardized_encoding.has_value());
    document.set_encoding(MUST(String::from_utf8(standardized_encoding.value())));
```

Libraries/LibWeb/DOM/DocumentLoading.cpp
```diff
@@ -184,18 +194,23 @@ static WebIDL::ExceptionOr<GC::Ref<DOM::Document>> load_xml_document(HTML::Navig
...
             // https://www.w3.org/TR/xml/#charencoding
             // [...] it is a fatal error [...] for an entity which begins with neither a Byte Order Mark nor an encoding
             // declaration to use an encoding other than UTF-8.
             auto bom_encoding = HTML::run_bom_sniff(data);
-            decoder = TextCodec::decoder_for(bom_encoding.value_or("UTF-8"));
+            document_encoding = bom_encoding.value_or("UTF-8");
+            decoder = TextCodec::decoder_for(document_encoding);
         }
         VERIFY(decoder.has_value());
+        document->set_encoding(MUST(String::from_byte_string(document_encoding)));
```
sideshowbarker:
```suggestion
        auto standardized_encoding = TextCodec::get_standardized_encoding(document_encoding);
        VERIFY(standardized_encoding.has_value());
        document->set_encoding(MUST(String::from_utf8(standardized_encoding.value())));
```

sideshowbarker (dismissed):

sideshowbarker:
OK, the CI failures we’re getting here now suggest there’s a general test-harness bug that needs to be investigated and fixed (separate from this PR) before this PR could move ahead.

Libraries/LibWeb/DOM/DocumentLoading.cpp
```diff
@@ -91,6 +100,11 @@ static WebIDL::ExceptionOr<GC::Ref<DOM::Document>> load_html_document(HTML::Navi
     // 1. Let document be the result of creating and initializing a Document object given "html", "text/html", and navigationParams.
     auto document = TRY(DOM::Document::create_and_initialize(DOM::Document::Type::HTML, "text/html"_string, navigation_params));
 
+    // AD-HOC: Set the navigable early so that step 6 of the encoding detection algorithm
+    //         (parent document encoding) can access the container document.
+    if (navigation_params.navigable)
+        document->set_navigable(navigation_params.navigable);
```
AtkinsSJ: I'm not seeing any other PRs get the weird CI flakes that are happening here, and these lines in particular look suspicious to me.

sideshowbarker: Yeah I thought about that too when I saw it. And while I agree it’s still kind of sus, I did actually do a bisect/temporary-revert of that and then a local test run — and I found it changed nothing; the html5lib timeouts and the Fails with jumbled details were still there.

So, I have a change in flight right now which fixes that, and will push it later.

That said, even though that early `set_navigable` isn’t the cause of the CI flakes, I guess it’s imaginable that is could have some problems when it’s under a lot of concurrency. But I don’t know how to test for that preemptively in a conclusive way. So IMHO we should probably not block this PR on trying to get complete clarity about that.

But as I said, I do already think it’s clear that’s at least not the cause of the CI flakes here.

sideshowbarker: After having looked at this more, I think I may have spoken too soon, and Sam called it right initially.

That `set_navigable` is changing init _ordering_, not just timing. It makes it so that _anywhere_ that previously would have early-returned (or taken a no-navigable path) for in-progress iframe init now takes the with-navigable path one step earlier. And when a document is doing a lot of concurrent loading of iframes, that’s … gonna have some effect.

The html5lib tests are sensitive to that because they create a bunch of hidden iframes right at load time.

So @Martin005 I think what would make sense here is: Revert for now just those AD-HOC `set_navigable` lines in `DocumentLoading.cpp` — but keep the Step 6 implementation in `HTMLEncodingDetection.cpp`) — and then push.

I think that’s likely going to clear up CI clears up — and if it does, then that confirms the AD-HOC ordering is the cause, and that we‘ll thus need a different way to make navigable available at encoding-sniff time (e.g., something like: pass the navigable directly to `run_encoding_sniffing_algorithm` — rather than relying on the document already having it stamped).

Martin005: @sideshowbarker Thanks for a thorough analysis! I removed the AD-HOC `set_navigable` lines, pushed and it seems the html5lib tests are not failing on CI anymore (the only failing test is the `@font-face` test that you are fixing in https://github.com/LadybirdBrowser/ladybird/pull/9730)

I think that passing the `navigable` to `run_encoding_sniffing_algorithm` and using its `container_document` as the parent document would work. I will wait for your confirmation before pushing such change 🙂

sideshowbarker: >  I think that passing the `navigable` to `run_encoding_sniffing_algorithm` and using its `container_document` as the parent document would work. I will wait for your confirmation before pushing such change 🙂

OK yeah please do go ahead with that change — thanks~

Martin005: @sideshowbarker Thanks, done in [`df563c1` (this PR)](https://github.com/LadybirdBrowser/ladybird/pull/9610/changes/df563c127d58ac09064233d239770ed3960106e6)

sideshowbarker (commented):
This looks pretty good to me, though I still want to look at it more detail before approving. But I think it should also have review and sign-of from @AtkinsSJ

gmta:

PR #9613, opened 2026-05-24
AK: Skip Optional converting constructors when U is same as T

veeti:
This template is causing an "error: satisfaction of atomic constraint ‘IsConstructible<T, const U&> [with T = X; X]’ depends on itself" through a self referential CSSScopeRule on GCC 16.1. Tighten the template requirements to avoid it.

### DIFF
--- a/AK/Optional.h
+++ b/AK/Optional.h
@@ -209,7 +209,7 @@ public:
     }
 
     template<typename U>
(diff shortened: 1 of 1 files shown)


NiccoloAntonelliDziri:
I encountered the same issue and this PR solved it for me. Thanks!

Darshanx256:
Very helpful! this needs to be merged

gmta (approved):

PR #9622, opened 2026-05-25
LibWeb: Position inline-outside abspos after preceding left float in IFC

### LINKED ISSUE #9621: CSS float fails  - for invalid contradictory css rules
### Summary

The page [Zim_a desktop wiki.html](https://github.com/user-attachments/files/28215759/Zim_a.desktop.wiki.html) renders as desired in Firefox but not in Ladybird. The cause is erroneous CSS rules of  [zim-wiki.org](https://zim-wiki.org/) (invalid or contradicting css rules). 

### Firefox / Librewolf

Image

### Ladybird

Image

#### Cause - invalid and contradictory css rules 
A LLMs commented the css: 

> Yes — several rules are outdated, fragile, or contradictory. The page may still render, but there are some layout and CSS design problems.

### Operating system

Linux

### Steps to reproduce

1. Open https://zim-wiki.org/ in Ladybird
2. See layout of startpage
3. Compare to layout of startpage to Firefox or chrome 

Comparison of test case Ladbird and Firefox.

Image

### Expected behavior

Layout is displayed as author desired it: navigation on the left side, content on the right side 

### Actual behavior

Navigation and content are placed at the same position (one covers the other)

https://zim-wiki.org/

```html
<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN">
<html style>
<meta charset=utf-8>
<title>Zim - a desktop wiki</title>

<!--[if !IE]-->
<style>
#wrapper {  width: 100%; height: 100%;  position: relative}

#navigation {    
        background-color: lime;
        
        display: inline-block; 
        
        border-right: 1px solid #EAEAEA;
        padding-top: 10px;
        padding-right: 20px;
        padding-left: 20px;
        position: relative;
        min-height: 100%;
        float: left
        /* 
        VS Code warning: 
            inline-block is ignored due to the float. 
            If 'float' has a value other than 'none', 
            the box is floated and 'display' 
            is treated as 'block'css(propertyIgnoredDueToDisplay)

        */
}

#content {    
    
        float: left;
        background-color: gray;
        display: inline-block;            
        position: absolute;
        max-width: 960px;
        padding-top: 10px;
        padding-right: 65px;
        padding-left: 65px;
        padding-bottom: 65px
}
</style>
</head>

<body>
    <div id="wrapper">    
        <div id="navigation">
            <h3>Navigation</h3>
            <p>links</p>
            <ul>
                <li>a</li>
                <li>b</li>
                <li>c</li>
                <li>d</li>
                <li>e</li>
                <li>f</li>
                <li>g</li>
                <li>h</li>
                <li>i</li>                                                                
            </ul>
        </div>    
        <div id="content">
            <h2>Content</h2>
            <div>some content </div>        
        </div> <!-- /content -->    
    </div> <!-- /wrapper -->
</body>
</html>
```

```shell
none
```

see above

_No response_

### Contribute a patch?

sideshowbarker:
**Problem:** An originally-inline-outside element with both `position: absolute` and `float: left` (where per-spec `float` becomes `none`) is positioned at the top left of its containing block, overlapping any preceding left-floated sibling — rather than being placed next to that float (where Firefox and Chrome place it).

**Cause:** `InlineFormattingContext::calculate_static_position_rect()` locates the previous sibling and looks it up in the IFC’s line-box fragments. Floats don’t produce line-box fragments — so when the only preceding sibling is a left float (with collapsed whitespace text nodes in between), the lookup finds nothing and the static position falls through to `(0, 0)`.

**Fix:** When no fragment is found and the abspos is originally inline-outside: Walk back through siblings to find a preceding left-float and use its margin-box width as the static position `x`. The hypothetical box (per spec, computed with `float: none`) would have been inline-level content in the line box wrapping the float — so placing it at the float’s outer end is correct for the single-left-edge-float case. Fixes https://github.com/LadybirdBrowser/ladybird/issues/9621.

### DIFF
--- a/Libraries/LibWeb/Layout/InlineFormattingContext.cpp
+++ b/Libraries/LibWeb/Layout/InlineFormattingContext.cpp
@@ -565,13 +565,17 @@ StaticPositionRect InlineFormattingContext::calculate_static_position_rect(Box c
     VERIFY(box.parent());
     VERIFY(box.parent()->children_are_inline());
 
+    // We're calculating the position for an absolutely positioned box with a previous sibling in an IFC.
+    // We need to position the box...
+    // ...below the last fragment of this sibling, if the display-outside value (before box type transformation) is block.
+    // ...at the top right corner of the last fragment of this sibling otherwise.
+    // ...at the outer end of a preceding left-float — if the box is originally inline-outside and no line-box fragment
+    //    is found (floats don't produce line-box fragments).
     CSSPixelPoint position;
-    if (auto const* sibling = box.previous_sibling()) {
-        // We're calculating the position for an absolutely positioned box with a previous sibling in an IFC.
-        // We need to position the box...
-        // ...below the last fragment of this sibling, if the display-outside value (before box type transformation) is block.
-        // ...at the top right corner of the last fragment of this sibling otherwise.
-        LineBoxFragment const* last_fragment = nullptr;
+    LineBoxFragment const* last_fragment = nullptr;
+    Box const* preceding_left_float = nullptr;
+    auto box_is_inline_outside = !box.display_before_box_type_transformation().is_block_outside();
+    for (auto const* sibling = box.previous_sibling(); sibling; sibling = sibling->previous_sibling()) {
         auto const& cb_state = m_state.get(*sibling->containing_block());
         for (auto const& line_box : cb_state.line_boxes) {
             for (auto const& fragment : line_box.fragments()) {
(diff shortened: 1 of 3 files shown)


sideshowbarker:
Closing in favor of the better fix in https://github.com/LadybirdBrowser/ladybird/pull/9623

tcl3:
Sorry, I didn't look at the existing PRs before I made mine :sweat_smile:

sideshowbarker:
> Sorry, I didn't look at the existing PRs before I made mine 😅

No worries — I think raised it literally only a minute or something before you raised yours. And I can see my change kind of preserved an existing bug, but yours fixes it. And you added WPT tests — which I didn't but realize now I should have. And I didn't use `m_containing_block_used_values` but I can see now I should have.

PR #9630, opened 2026-05-25
LibWeb: Allow UTF-8 detection for non-http(s) schemes

aplefull:


Libraries/LibWeb/HTML/Parser/HTMLEncodingDetection.cpp
```diff
@@ -530,9 +530,13 @@ ByteString run_encoding_sniffing_algorithm(DOM::Document& document, ReadonlyByte
     u8 const* encoding_name_ptr = nullptr;
     size_t encoding_name_len = 0;
 
+    auto scheme = document.url().scheme();
+    bool is_network_scheme = scheme == "http"sv || scheme == "https"sv;
```
shannonbooth: I don't think this makes a big difference either since this is the main thing we need to do, but https://hsivonen.fi/chardetng/ seems to suggest the logic of chrome/firefox is `bool allow_utf8_detection = document.url().scheme() == "file"sv;` and it would be nice to match that

aplefull: Ah, sure, changed it!

shannonbooth: and a nit sorry, commit message should be updated to match

aplefull: done c:

shannonbooth (approved):
thanks!

PR #9640, opened 2026-05-26
LibGfx: Fix PNG decoder leaks from libpng error longjmp

sideshowbarker:
**Problem:** Runs of the Linux Sanitizers intermittently fail with LeakSanitizer leaks of ~45 KB across 21 allocations from `PNGLoadingContext::read_frames()` — triggered when the test suite decodes malformed PNGs.

**Cause:** When libpng hits a corrupted `IDAT` chunk inside `png_read_image()`, it `longjmp`s back to the `setjmp` landing pad in `PNGLoadingContext::read_all_frames()`. `longjmp` unwinds the stack without running C++ destructors — so the stack-locals in `read_frames` (`Vector<u8*> row_pointers`, the in-flight `Bitmap` inside `decode_frame`, and the APNG branch’s `output_buffer` and `Painter`) leak their heap storage.

**Fix:** Promote those stack-locals to members of `PNGLoadingContext` (which is heap-allocated and outlives the `setjmp` scope) — so their storage is reachable by RAII when the context is destroyed. Clear them in the `setjmp` error handler too — so memory is released promptly on the error path, rather than waiting until `~PNGLoadingContext()`.


Libraries/LibGfx/ImageFormats/PNGLoader.cpp
```diff
@@ -364,6 +380,12 @@ ErrorOr<size_t> PNGLoadingContext::read_frames(png_structp png_ptr, png_infop in
         auto decoded_frame_bitmap = TRY(decode_frame(size));
         frame_descriptors.append({ move(decoded_frame_bitmap), 0 });
     }
+
+    // Clear the working state — so it doesn't outlive read_frames() on the success path.
+    row_pointers.clear();
+    in_flight_bitmap = nullptr;
+    output_buffer = nullptr;
+    painter = nullptr;
```
gmta: Instead of duplicating these four lines, can you add an explicit method to the struct to clear them? That'd go neatly with the comment explaining why we store them in the struct in the first place.

gmta (commented):
Nice!

Libraries/LibGfx/ImageFormats/PNGLoader.cpp
sideshowbarker: 👍 Made it so~

gmta (approved):

PR #9668, opened 2026-05-27
LibWeb: Skip non-rendered elements before querying view transition name

tcl3:
Previously, starting a view transition on an element inside a `display:none` subtree  would crash because querying the view transition name unconditionally accessed `computed_properties()`, which is not calculated for these elements unless explicitly requested. We now skip these non-rendered elements early, before querying the view transition name, to avoid the crash.

This prevents a crash on https://wheelofnames.com, when hiding the right pane.


Psychpsyo:
Changes look good to me.

PR #9685, opened 2026-05-27
LibWeb: Follow explicit cell widths when calculating fixed table layouts

Darshanx256:
This PR fixes a bug in the fixed table layout, where cells with
explicitly specified length/percentage widths were ignoring their
explicit width and falling back to the content size.

Specifications referred:
https://www.w3.org/TR/css-tables-3/#computing-column-measures
https://www.w3.org/TR/CSS2/tables.html#fixed-table-layout
https://www.w3.org/TR/CSS22/tables.html#auto-table-layout

The fix resolves horizontal squashing layout issues seen on live
websites, such as the JS Crossword grid at:
https://lyra.horse/fun/jscrossword/

Added a regression test:
Tests/LibWeb/Layout/input/table/fixed-layout-cell-specified-width.html

### DIFF
--- a/Libraries/LibWeb/Layout/TableFormattingContext.cpp
+++ b/Libraries/LibWeb/Layout/TableFormattingContext.cpp
@@ -130,7 +130,9 @@ void TableFormattingContext::compute_constrainedness()
 
     for (auto& cell : m_cells) {
         auto const& computed_values = cell.box->computed_values();
-        if (computed_values.width().is_length()) {
+        // https://drafts.csswg.org/css-tables-3/#computing-column-measures
+        // For the purpose of measuring a column when laid out in fixed mode, only cells which originate in the first row of the table (after reordering the header and footer) will be considered, if any.
+        if (computed_values.width().is_length() && (!use_fixed_mode_layout() || cell.row_index == 0)) {
             m_columns[cell.column_index].is_constrained = true;
         }
 
@@ -175,10 +177,27 @@ void TableFormattingContext::compute_cell_measures()
         // The outer min-content width of a table-cell is max(min-width, min-content width) adjusted by the cell intrinsic offsets.
         auto min_width = computed_values.min_width().to_px(cell.box, containing_block_width);
         auto cell_intrinsic_width_offsets = padding_left + padding_right + border_left + border_right;
-        // For fixed mode, according to https://www.w3.org/TR/css-tables-3/#computing-column-measures:
+        // For fixed mode, according to https://drafts.csswg.org/css-tables-3/#computing-column-measures:
         // The min-content and max-content width of cells is considered zero unless they are directly specified as a length-percentage,
         // in which case they are resolved based on the table width (if it is definite, otherwise use 0).
         auto width_is_specified_length_or_percentage = computed_values.width().is_length() || computed_values.width().is_percentage();
+        if (use_fixed_mode_layout()) {
+            if (computed_values.width().is_percentage()) {
+                CSSPixels reference_width = 0;
+                auto table_width = table_box().computed_values().width();
+                if (table_width.is_length() || table_width.is_percentage())
+                    reference_width = table_width.to_px(table_box(), containing_block_width);
+                min_content_width = max_content_width = computed_values.width().to_px(cell.box, reference_width);
+            } else if (computed_values.width().is_length()) {
+                min_content_width = max_content_width = computed_values.width().to_px(cell.box, containing_block_width);
+            } else {
+                min_content_width = max_content_width = 0;
+            }
+
+            if (computed_values.box_sizing() == CSS::BoxSizing::BorderBox)
+                min_content_width = max_content_width = max(CSSPixels(0), min_content_width - cell_intrinsic_width_offsets);
+        }
+
         if (!use_fixed_mode_layout() || width_is_specified_length_or_percentage) {
             cell.outer_min_width = max(min_width, min_content_width) + cell_intrinsic_width_offsets;
         }
@@ -277,8 +296,11 @@ void TableFormattingContext::initialize_table_measures<TableFormattingContext::C
     // Implement the following parts of the specification, accounting for fixed layout mode:
     // https://www.w3.org/TR/css-tables-3/#min-content-width-of-a-column-based-on-cells-of-span-up-to-1
     // https://www.w3.org/TR/css-tables-3/#max-content-width-of-a-column-based-on-cells-of-span-up-to-1
+    bool const fixed = use_fixed_mode_layout();
     for (auto& cell : m_cells) {
-        if (cell.column_span == 1 && (cell.row_index == 0 || !use_fixed_mode_layout())) {
+        // https://drafts.csswg.org/css-tables-3/#computing-column-measures
+        // For the purpose of measuring a column when laid out in fixed mode, only cells which originate in the first row of the table (after reordering the header and footer) will be considered, if any.
+        if (cell.column_span == 1 && (cell.row_index == 0 || !fixed)) {
             m_columns[cell.column_index].min_size = max(m_columns[cell.column_index].min_size, cell.outer_min_width);
             m_columns[cell.column_index].max_size = max(m_columns[cell.column_index].max_size, cell.outer_max_width);
         }
@@ -301,9 +323,16 @@ void TableFormattingContext::compute_intrinsic_percentage(size_t max_cell_span)
         intrinsic_percentage_contribution_by_index[rc_index] = rows_or_columns[rc_index].intrinsic_percentage;
     }
 
+    [[maybe_unused]] bool const fixed_layout = use_fixed_mode_layout();
     for (size_t current_span = 2; current_span <= max_cell_span; current_span++) {
         // https://www.w3.org/TR/css-tables-3/#intrinsic-percentage-width-of-a-column-based-on-cells-of-span-up-to-n-n--1
         for (auto& cell : m_cells) {
+            if constexpr (IsSame<RowOrColumn, Column>) {
+                // https://drafts.csswg.org/css-tables-3/#computing-column-measures
+                // For the purpose of measuring a column when laid out in fixed mode, only cells which originate in the first row of the table (after reordering the header and footer) will be considered, if any.
+                if (fixed_layout && cell.row_index != 0)
+                    continue;
+            }
             auto cell_span_value = cell_span<RowOrColumn>(cell);
             if (cell_span_value != current_span) {
                 continue;
@@ -372,8 +401,15 @@ void TableFormattingContext::compute_table_measures()
 
     auto& rows_or_columns = table_rows_or_columns<RowOrColumn>();
 
+    [[maybe_unused]] bool const fixed_layout = use_fixed_mode_layout();
     size_t max_cell_span = 1;
     for (auto& cell : m_cells) {
+        if constexpr (IsSame<RowOrColumn, Column>) {
+            // https://drafts.csswg.org/css-tables-3/#computing-column-measures
+            // For the purpose of measuring a column when laid out in fixed mode, only cells which originate in the first row of the table (after reordering the header and footer) will be considered, if any.
+            if (fixed_layout && cell.row_index != 0)
+                continue;
+        }
         max_cell_span = max(max_cell_span, cell_span<RowOrColumn>(cell));
     }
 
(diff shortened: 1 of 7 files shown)


Libraries/LibWeb/Layout/TableFormattingContext.cpp
```diff
@@ -179,6 +179,12 @@ void TableFormattingContext::compute_cell_measures()
         // The min-content and max-content width of cells is considered zero unless they are directly specified as a length-percentage,
         // in which case they are resolved based on the table width (if it is definite, otherwise use 0).
         auto width_is_specified_length_or_percentage = computed_values.width().is_length() || computed_values.width().is_percentage();
+        if (use_fixed_mode_layout()) {
+            min_content_width = max_content_width = width_is_specified_length_or_percentage ? computed_values.width().to_px(cell.box, containing_block_width) : 0;
+            if (computed_values.box_sizing() == CSS::BoxSizing::BorderBox)
+                min_content_width = max_content_width = max(CSSPixels(0), min_content_width - padding_left - padding_right - border_left - border_right);
+        }
+
         if (!use_fixed_mode_layout() || width_is_specified_length_or_percentage) {
```
Psychpsyo, line 201: Wouldn't it make sense for this to be an `else` branch now?

Libraries/LibWeb/Layout/TableFormattingContext.cpp
```diff
@@ -179,6 +179,12 @@ void TableFormattingContext::compute_cell_measures()
         // The min-content and max-content width of cells is considered zero unless they are directly specified as a length-percentage,
         // in which case they are resolved based on the table width (if it is definite, otherwise use 0).
         auto width_is_specified_length_or_percentage = computed_values.width().is_length() || computed_values.width().is_percentage();
+        if (use_fixed_mode_layout()) {
+            min_content_width = max_content_width = width_is_specified_length_or_percentage ? computed_values.width().to_px(cell.box, containing_block_width) : 0;
+            if (computed_values.box_sizing() == CSS::BoxSizing::BorderBox)
+                min_content_width = max_content_width = max(CSSPixels(0), min_content_width - padding_left - padding_right - border_left - border_right);
```
Psychpsyo: This doesn't seem to follow from the spec text above.
Why are the min and max content width set to the value of `width`?
Why is box-sizing involved?

Psychpsyo:
This also seems to fail some existing tests.

Libraries/LibWeb/Layout/TableFormattingContext.cpp
Darshanx256: Spec says that in fixed layout, cell content doesn't matter, spec says we must treat the cell's specified width property 

min_content_width represents content box but excluding padding, ive used box-sizing as it includes them. Just subtracting padding and border such they they dont get double counted (as engine will be adding them later)

Libraries/LibWeb/Layout/TableFormattingContext.cpp
Darshanx256: not mutually exclusive! second if executes when `use_fixed_mode_layout()` is true if `width_is_specified_length_or_percentage` is also true

Darshanx256:
looking at that

Libraries/LibWeb/Layout/TableFormattingContext.cpp
Psychpsyo: `else if (width_is_specified_length_or_percentage)` then.

Libraries/LibWeb/Layout/TableFormattingContext.cpp
Psychpsyo: > spec says we must treat the cell's specified width property

I have looked at every section of the spec that mentions fixed layout, and I cannot find this.

Darshanx256:
The test failures in fixed-layout-percentage-width-all-columns.html and fixed-layout-percentage-width.html were most probably false alarms.

I discovered this by regenerating the test dumps and verifying the column dimensions manually against the CSS Tables Level 3 Specification. I noticed that before this PR, there was a bug where percentage cell widths (like width: 50%) were evaluating to 0px under the hood. This forced the engine to fallback to the intrinsic text width of the cell instead of mathematically calculating the correct percentage. The old test expectation files were generated while this bug was active, meaning they captured the wrong layout.

Looking at the baseline expectation in the master, It expected the 50% cell to have a width of 295px. So I've rebaselined the expectation files to match the new, correct output.

How to reproduce the old bug:

- Checkout the master branch.
- Run this command to dump the layout tree for the test: Build/release/bin/test-web --dump-layout-tree Tests/LibWeb/Layout/input/table/fixed-layout-percentage-width.html
- Look at the output for the 50% cell. Even though the table is 600px, the cell incorrectly gets assigned a width of 295px instead of 300px.
- Switch back to this PR branch and run the exact same command. You'll see the cell now correctly gets assigned 300px, proving the layout is finally doing the math right.

Darshanx256:
First off, the code checks if the table is actually using the fixed layout mode by calling use_fixed_mode_layout().
If it's using a fixed layout, it immediately forces both the min_content_width and max_content_width of the cell to match whatever width is explicitly given, like width: 50px. If  width is isnt specified at all, it just treats the content width as 0.
After that, it checks if the cell is using box-sizing: border-box. If it is, the code subtracts the cell's padding and borders from that specified width so the inner content area is sized correctly, making sure the size never drops below zero pixels.
The new amend handles percent based widths properly. Also the test timeout `content-blocker-srcdoc-frame-cosmetic-source` looks unrelated

Libraries/LibWeb/Layout/TableFormattingContext.cpp
Darshanx256: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/table-layout

you can cross verify here. both width (non-percent) as well as percent needs to be considered.

Libraries/LibWeb/Layout/TableFormattingContext.cpp
```diff
@@ -179,6 +179,23 @@ void TableFormattingContext::compute_cell_measures()
         // The min-content and max-content width of cells is considered zero unless they are directly specified as a length-percentage,
         // in which case they are resolved based on the table width (if it is definite, otherwise use 0).
         auto width_is_specified_length_or_percentage = computed_values.width().is_length() || computed_values.width().is_percentage();
+        if (use_fixed_mode_layout()) {
+            if (computed_values.width().is_percentage()) {
+                CSSPixels reference_width = 0;
+                auto table_width = table_box().computed_values().width();
```
Psychpsyo, line 187: This should be using the actual width of the table, not just the computed value of CSS `width`.

Libraries/LibWeb/Layout/TableFormattingContext.cpp
```diff
@@ -179,6 +179,23 @@ void TableFormattingContext::compute_cell_measures()
...
+                    reference_width = table_width.to_px(table_box(), containing_block_width);
+                min_content_width = max_content_width = computed_values.width().to_px(cell.box, reference_width);
+            } else if (computed_values.width().is_length()) {
+                min_content_width = max_content_width = computed_values.width().to_px(cell.box, containing_block_width);
+            } else {
+                min_content_width = max_content_width = 0;
+            }
+
+            if (computed_values.box_sizing() == CSS::BoxSizing::BorderBox)
+                min_content_width = max_content_width = max(CSSPixels(0), min_content_width - padding_left - padding_right - border_left - border_right);
```
Psychpsyo: This could just use the `cell_intrinsic_width_offsets`.

Psychpsyo:
Just as an observation: The old test expectations match WebKit behavior. The new expectations match Chromium and Gecko.

Libraries/LibWeb/Layout/TableFormattingContext.cpp
Psychpsyo: > https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/table-layout
> 
> you can cross verify here. both width (non-percent) as well as percent needs to be considered.

MDN is not the specification and has no authority over how a feature should work. It also tends to oversimplify the finer details of the spec.

I'm not saying that using the cell's width is wrong here, but the code should link to and quote the part of the spec that says to do so. If the spec does not say to do something but that thing seems necessary, then either the spec is wrong or that way of implementing it is wrong. So either the code needs to be changed, or there needs to be a bug filed on the spec, which can then be linked in the code as the justification for implementing it that way.

Darshanx256: Understood. I initially implemented it to use width when I was locally testing, as I saw other browsers working the same way and then I found the behaviour in MDN.

Additionally, I managed to get W3C spec article mentioning it:
https://www.w3.org/TR/css-tables-3/
(in section 3.5.1)

I am making this a draft and try to match the spec as much as possible, also considering your recommendations.

Darshanx256:
Just an update on top of this, I cleaned up a few more things to get us strictly 1:1 with the spec.

First off, I swapped out the manual padding and border subtraction with cell_intrinsic_width_offsets because it cleans up the math.

After that, I made sure we completely ignore any cell that isn't in the very first row when calculating column widths in fixed mode. It just skips them in the layout loops so they don't mess up the constraints or percentage distributions.

I also cleaned up those for loops by moving the use_fixed_mode_layout() checks outside the loops so it doesn't spam property lookups on every single cell, and flattened out some nested if statements.

A bug existing in the upstream code where fixed layout tables with width: auto were incorrectly expanding to satisfy percent based cell widths. Like, if a cell down in the 10th row had width: 90%, it would literally force the whole fixed table to stretch out. But the spec explicitly says that percentage expansion algorithm is only for auto-table-layouts. For fixed layouts with auto width, the browser isn't supposed to artificially expand the table to satisfy cell percentages. So I wrapped that expansion loop in an if (!use_fixed_mode_layout()) check to completely kill it for fixed layouts. 

https://www.w3.org/TR/CSS22/tables.html#auto-table-layout (This defines the expansion algorithm, proving it ONLY applies to automatic layouts).
https://www.w3.org/TR/CSS22/tables.html#fixed-table-layout (This explicitly states that for fixed tables with auto width, the browser "is not required to do anything in particular")

Libraries/LibWeb/Layout/TableFormattingContext.cpp
Darshanx256: I looked into using the actual table width here, but it creates a hard cyclic dependency. Per CSS 2.1 Section 17.5.2, a fixed table's actual width is calculated as max(CSS width, sum of columns), meaning the table's actual width literally hasn't been calculated yet because it depends on the cell measures we are doing right now! If we evaluated cell percentages against the actual width instead of the computed CSS width, the column sum would rely on the actual width while the actual width relies on the column sum, causing an unsolvable infinite loop.

Libraries/LibWeb/Layout/TableFormattingContext.cpp
Darshanx256: Thats helpful!

Libraries/LibWeb/Layout/TableFormattingContext.cpp
```diff
@@ -130,7 +130,9 @@ void TableFormattingContext::compute_constrainedness()
 
     for (auto& cell : m_cells) {
         auto const& computed_values = cell.box->computed_values();
-        if (computed_values.width().is_length()) {
+        // https://www.w3.org/TR/css-tables-3/#fixed-layout
```
Psychpsyo: #fixed-layout doesn't seem to link to any real section of the spec. Also, please link to the editor's draft version of the spec.
Same goes for all the other instances of this link.

Libraries/LibWeb/Layout/TableFormattingContext.cpp
```diff
@@ -130,7 +130,9 @@ void TableFormattingContext::compute_constrainedness()
...
+        // In fixed layout, only cells in the first row are considered for column measurement.
```
Psychpsyo: This sentence does not exist in the spec. You should copy the spec text verbatim, without any changes. Same goes for all the other copies of it.

Libraries/LibWeb/Layout/TableFormattingContext.cpp
Darshanx256: I have replaced it with the correct anchor link `#computing-column-measures` as I have earlier mentioned in pr description. The discrepancy happened as I used LLM to place links at those 6 spots, I should have been careful with links.

Libraries/LibWeb/Layout/TableFormattingContext.cpp
Darshanx256: I have pasted the correct verbatim with same context.

gmta:

PR #9688, opened 2026-05-27
UI/Qt: Integrate menus with frameless window chrome

awesomekling:
Update the Qt chrome so the optional menu bar and popup menus match the frameless browser window styling. This adds chrome-themed QMenuBar and QMenu styles, removes menu action icons and unused icon resources, and gives empty menu bar space title-bar behavior for dragging and double-click maximize.

When the menu bar is visible, window controls now move into the menu bar instead of staying in the tab strip. The tab strip and menu bar share the same window-control button behavior, including correct pressed-state handling when the pointer leaves and re-enters a button.


UI/Qt/Menu.cpp
```diff
@@ -114,133 +119,114 @@ static QIcon icon_from_base64_png(StringView favicon_base64_png)
...
-        qaction.setIcon(load_icon_from_uri("resource://icons/16x16/filetype-html.png"sv));
         qaction.setShortcut(QKeySequence(Qt::CTRL | Qt::Key_U));
         break;
 
     case WebView::ActionID::TakeVisibleScreenshot:
     case WebView::ActionID::TakeFullScreenshot:
-        qaction.setIcon(load_icon_from_uri("resource://icons/16x16/filetype-image.png"sv));
         break;
 
     case WebView::ActionID::OpenInNewTab:
```
trflynn89: Let's remove all the now-empty case labels?

trflynn89 (commented):
It seems like menu actions now don't indicate that they are disabled, which is pretty confusing. The go back / go forward actions here should be dimmed, for example:

Image

PR #9690, opened 2026-05-27
LibWeb: Avoid white flashes during dark-mode loads

awesomekling:
Make initial and loading document backgrounds respect dark mode more consistently, including about:blank, web view fallback fills, and documents whose color-scheme state is still being established.

This keeps the previous frame visible across process swaps, defers WebContent frames until the system theme is initialized, and makes canvas color selection use the document-supported color schemes where appropriate.

Also adds coverage for meta color-scheme changes invalidating style.

### DIFF
--- a/Libraries/LibWeb/CSS/SystemColor.h
+++ b/Libraries/LibWeb/CSS/SystemColor.h
@@ -19,7 +19,7 @@ Color active_text(PreferredColorScheme);
 Color button_border(PreferredColorScheme);
 Color button_face(PreferredColorScheme);
 Color button_text(PreferredColorScheme);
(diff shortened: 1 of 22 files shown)


Services/WebContent/PageClient.cpp
```diff
@@ -162,7 +162,6 @@ Gfx::Palette PageClient::palette() const
 void PageClient::set_palette_impl(Gfx::PaletteImpl& impl)
 {
     m_palette_impl = impl;
-    m_has_received_system_theme = true;
```
gmta: Ehh, does this revert the earlier "WebContent: Defer frames until the theme is initialized" commit? 🤔

PR #9692, opened 2026-05-27
LibWeb: Delay load event during CSS @import fetch

sideshowbarker:
**Problem:** `Ref/input/css/images-load-relative-to-imported-style-sheet.html` intermittently fails on the Linux Sanitizer with a deterministic 398px screenshot mismatch — the entire 20x20 background-image div is unpainted in the test pass, while the reference paints the checkerboard.

**Cause:** `CSSImportRule::fetch()` added to the document’s render-blocking set (`m_pending_css_import_rules`) — but didn’t delay the load event. The load event could fire before the imported stylesheet had finished fetching — so the background image referenced from inside that sheet hadn’t even been requested yet when test-web’s `on_load_finish` callback ran `wait_for_reftest_completion` and proceeded to take the screenshot. By contrast, `<link rel="stylesheet">` already delays the load event for its own duration via `m_document_load_event_delayer` in `HTMLLinkElement` — which is why the sibling test using `<link>` doesn’t flake.

**Fix:** Hold a `DocumentLoadEventDelayer` in `CSSImportRule` for the duration of the import fetch, paralleling `HTMLLinkElement`. By the time the import rule’s delayer clears in the `ScopeGuard`, the imported sheet’s image fetches have already been kicked off via `SharedResourceRequest`, each of which takes over load-event delaying through its own delayer — so there’s no race between the two delayers releasing.

### DIFF
--- a/Libraries/LibWeb/CSS/CSSImportRule.cpp
+++ b/Libraries/LibWeb/CSS/CSSImportRule.cpp
@@ -139,6 +139,13 @@ void CSSImportRule::fetch()
 
     // AD-HOC: Track pending import rules to block rendering until they are done.
     m_document->add_pending_css_import_rule({}, *this);
+    // FIXME: This dedicated load-event delayer shouldn't be needed. The style-sheet critical-subresource mechanism
+    //        already delays the document load event via the owning element. However, it only does that delay when
+    //        contributes_a_script_blocking_style_sheet() is true. But HTMLStyleElement currently always returns false
+    //        for that. Once HTMLStyleElement actually has working contributes_a_script_blocking_style_sheet() behavior
+    //        implemented (as HTMLLinkElement element already does), the owning style element's delayer will cover the
+    //        @import fetch — and then at that time, this dedicated load-event delayer can go away.
+    m_load_event_delayer.emplace(*m_document);
     set_loading_state(CSSStyleSheet::LoadingState::Loading);
 
     // 3. Fetch a style resource from rule’s URL, with ruleOrDeclaration rule, destination "style", CORS mode "no-cors", and
@@ -154,6 +161,7 @@ void CSSImportRule::fetch()
             // AD-HOC: Stop delaying the load event.
             ScopeGuard guard = [strong_this, document] {
                 document->remove_pending_css_import_rule({}, strong_this);
+                strong_this->m_load_event_delayer.clear();
                 if (strong_this->loading_state() != CSSStyleSheet::LoadingState::Error) {
                     // If we have no critical subresources, or they're loaded already, we can report that immediately.
                     auto sheet_loading_state = strong_this->m_style_sheet->loading_state();
--- a/Libraries/LibWeb/CSS/CSSImportRule.h
+++ b/Libraries/LibWeb/CSS/CSSImportRule.h
@@ -11,6 +11,7 @@
 #include <LibWeb/CSS/CSSRule.h>
 #include <LibWeb/CSS/CSSStyleSheet.h>
 #include <LibWeb/CSS/URL.h>
+#include <LibWeb/DOM/DocumentLoadEventDelayer.h>
 #include <LibWeb/Export.h>
 #include <LibWeb/Forward.h>
 
@@ -66,6 +67,7 @@ private:
     RefPtr<Supports> m_supports;
     GC::Ref<MediaList> m_media;
     GC::Ptr<CSSStyleSheet> m_style_sheet;
+    Optional<DOM::DocumentLoadEventDelayer> m_load_event_delayer;
 };
 
 template<>


Libraries/LibWeb/CSS/CSSImportRule.cpp
```diff
@@ -154,6 +155,7 @@ void CSSImportRule::fetch()
             // AD-HOC: Stop delaying the load event.
             ScopeGuard guard = [strong_this, document] {
                 document->remove_pending_css_import_rule({}, strong_this);
+                strong_this->m_load_event_delayer.clear();
```
AtkinsSJ, line 164: My memory is foggy but I have a feeling we used to do this, and then it got replaced with the pending import rule code above, which should do the same thing? The load delay mechanisms are complicated though.

sideshowbarker: Dunno — I guess I could look through the git history/blame to see. But I do know that test is flaky; it failed at least 4 times in CI recently. And as far as I can see, there is a race there in the current code which led to the flakes. And the change here should prevent that race (and not regress any other tests).

AtkinsSJ: I found it: https://github.com/LadybirdBrowser/ladybird/pull/7787

The idea is that the style sheet itself should be delaying the load event until its subresources are loaded, and that includes `@import`s. Seems like that's not behaving properly. :sweat_smile:

sideshowbarker: Ah yeah indeed it seems it’s not behaving as expected at this point at least.

sideshowbarker: @AtkinsSJ Lemme know if you reckon there’s more investigation I should do here.

AtkinsSJ: I can't think of anything in particular that could be messing things up... I don't mind this getting merged, but marking it with a FIXME would be helpful - we should really only need the one mechanism. At some point I'll throw this at $LLM_OF_CHOICE and see if it can find any likely causes, unless you feel like doing that.

sideshowbarker: > At some point I'll throw this at $LLM_OF_CHOICE and see if it can find any likely causes, unless you feel like doing that.

Tried that today with Opus 4.8 on max effort, and got this back:

> The document load-event delay for a `<style>`’s critical subresources is gated behind `HTMLStyleElement::contributes_a_script_blocking_style_sheet()` — and that override currently just `return false`s (every condition is still a `FIXME`). So in `StyleElementBase::update_a_style_block()` the `m_document_load_event_delayer` is never emplaced for a `<style>`, which means the #7787 mechanism is effectively inert for `<style>`. `HTMLLinkElement::contributes_a_script_blocking_style_sheet()` does implement its conditions (parser-created, enabled-when-created), which is why the sibling `<link>` test doesn’t flake.
>
> The unguarded window is exactly the `@import` fetch: during it the `@import` is a `Loading` critical subresource of the `<style>`’s sheet, so the owning `<style>` should already be holding the load event through that delayer — but it never created one. Once the imported sheet’s images start fetching they hold the load event via their own `SharedResourceRequest` delayers, so the race is specifically the fetch window — which lines up with what I saw (the image hadn’t even been requested when the load event fired).
>
> So the “one mechanism” fix looks like: implement `HTMLStyleElement::contributes_a_script_blocking_style_sheet()` the way `HTMLLinkElement` does (track parser-creation + enabled-on-create), after which the owning `<style>`’s existing delayer covers the `@import` fetch and the dedicated delayer in the PR becomes unnecessary. That’s a behavior change (a `<style>` would now delay the document load event for its subresources) and it touches the parser. This is static analysis — I haven’t built/tested it yet.

And it drafted this FIXME:

(for the `m_load_event_delayer.emplace(*m_document)` site in `CSSImportRule::fetch()`)
```
// FIXME: This dedicated load-event delayer should not be needed. The style sheet
//        critical-subresource mechanism already delays the document load event via the owning
//        element, but only when contributes_a_script_blocking_style_sheet() is true, and
//        HTMLStyleElement always returns false there. Once that is implemented (as it is for
//        <link>), the owning <style>'s delayer will cover the @import fetch and this can go away.
```

AtkinsSJ: Ahhhhh, I am actually in the process of implementing `contributes_a_script_blocking_style_sheet()` because it was giving me other issues! I should have mentioned it but I didn't remember that they were connected.

Fixme sounds fine. 👍

sideshowbarker: OK, rewrote the FIXME for more clarity, and pushed it. If with that, this looks OK to you, shall we merge?

AtkinsSJ (approved):

PR #9713, opened 2026-05-28
LibWeb: Select the largest decoded favicon as the active page favicon

trflynn89:
We would previously select the first favicon in reverse tree order as the active page favicon, while sending the bitmap for every decoded favicon to the UI process. We now selectively choose the largest favicon, and only transfer the icon that was actually chosen for the document (either the largest of the `<link rel="icon">` elements, or the fallback).

This let's us select a much less blurry icon for strava.

Before:
before

After:
after

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLLinkElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLLinkElement.cpp
@@ -948,9 +948,6 @@ static NonnullRefPtr<Core::Promise<NonnullRefPtr<Gfx::Bitmap const>>> decode_fav
             promise->reject(Error::from_string_view("Failed to get bitmap from SVG favicon"sv));
             return promise;
         }
-        auto navigable = document->navigable();
-        if (navigable && navigable->is_traversable())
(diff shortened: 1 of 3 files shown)


Libraries/LibWeb/HTML/HTMLLinkElement.cpp
```diff
@@ -1040,7 +1036,14 @@ void HTMLLinkElement::load_fallback_favicon_if_needed(GC::Ref<DOM::Document> doc
         auto global = GC::Ref { realm.global_object() };
 
         auto process_body = GC::create_function(realm.heap(), [document, request](ByteBuffer body) {
-            (void)decode_favicon(body, request->url(), document);
+            decode_favicon(body, request->url(), document)
```
Lubrsi, line 1039: issue: Does the Core::Promise go out of scope and get destroyed here?

trflynn89: It's kept alive in `decode_favicon`'s callbacks

PR #9718, opened 2026-05-28
LibURL: Replace WTF-8 surrogates before Rust URL parsing

shannonbooth:
The Rust URL FFI used String::from_utf8_lossy() on raw bytes from C++. That was not equivalent to the old C++ parser behavior for WTF-8 encoded surrogates. For example a byte sequence such as ED A0 80 was treated as three invalid UTF-8 bytes, producing three U+FFFD replacement characters.

Instead use String::from_utf8_with_replacement_character and let Rust borrow the input as &str.

Fixes a regression with the rust port as I had believed these were equivalent.

### DIFF
--- a/Libraries/LibURL/RustIntegration.cpp
+++ b/Libraries/LibURL/RustIntegration.cpp
@@ -168,7 +168,7 @@ static FFI::FfiUrlHost host_to_ffi(Optional<Host> const& host)
         },
         [&](String const& str) {
             result.kind = FFI::RustUrlHostKind::String;
-            result.string_data = reinterpret_cast<uint8_t const*>(str.bytes().data());
+            result.string_data = reinterpret_cast<u8 const*>(str.bytes().data());
             result.string_length = str.bytes().size();
         });
     return result;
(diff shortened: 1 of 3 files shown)


Libraries/LibURL/RustIntegration.cpp
```diff
@@ -418,11 +418,15 @@ static void on_parse_host_complete(void* ctx_ptr, FFI::FfiUrlHost const* ffi_res
 
 Optional<Host> parse_host(StringView input, bool is_opaque)
 {
+    // URL parsing expects a scalar-value UTF-8 string, but WTF-8 can be provided.
+    auto processed_input = String::from_utf8_with_replacement_character(input, String::WithBOMHandling::No);
```
shannonbooth, line 422: I think this case actually fixes a bug we had pre-rust port

Libraries/LibURL/RustIntegration.cpp
```diff
@@ -418,11 +418,15 @@ static void on_parse_host_complete(void* ctx_ptr, FFI::FfiUrlHost const* ffi_res
...
+    auto processed_input_view = processed_input.bytes_as_string_view();
+
     Optional<Host> result;
     HostParseCallbackCtx ctx { .result = &result };
     bool const did_succeed = FFI::rust_url_parse_host(
-        reinterpret_cast<uint8_t const*>(input.characters_without_null_termination()),
-        input.length(),
+        reinterpret_cast<uint8_t const*>(processed_input_view.characters_without_null_termination()),
```
ADKaster: `s/uint8_t/u8/g` ? or is there a reason to use inttypes.h when calling FFI

shannonbooth: No good reason I can think of, though I suppose it does match the generated FFI header signature and I honestly also forgot about our typedefs since the code here is already not too pretty, I can do a sweep to clean it up. We do the same in some other files

ADKaster: Changing the generated headers would require a patch to cbindgen, I fear. Unless I missed something in their massive list of config options.

shannonbooth: Yeah I dont think we can do that (and I think that's fine), but we can still change every instance of inttypes.h in this file and have all hand written code use our u8/u16/whatever typedefs if we want

PR #9720, opened 2026-05-28
Meta: Enable Gtk UI on macOS

ADKaster:
Turns out this just works :tm:. Making this an option should make it easier for maintainers on MacBooks to review PRs.

image


sideshowbarker (approved):
Reviewed all code changes and have no suggestions for changes.

Built and took it for a test drive. Console on launch showed this error:

> (process:80033): GLib-GIO-CRITICAL **: 14:38:25.485: g_settings_schema_source_lookup: assertion 'source != NULL' failed

But that didn‘t prevent it from launching fine.

Only glitch I could see (and also in your screenshot): There’s unexpectedly no icon for the “New Tab” button; instead, there’s a “missing image” replacement image.

But the button still works as expected when pressed.

And so I those are just wrinkles that can be ironed out later, and this could go ahead and be merged.

PR #9721, opened 2026-05-28
LibWeb/CSS: Implement container-relative units

AtkinsSJ:
That's `cqw`, `cqi`, `cqmax` and chums.

The implementation is a bit naive for now - it just walks up the tree each time to find the relevant size container. We can optimise that in a few ways.


AtkinsSJ:
Changes: Fixed a missing GC::Ptr and visit, and also added a cache for the horizontal and vertical size containers so we don't have to search for them so often.

PR #9727, opened 2026-05-29
LibWeb: Let range inputs stretch in grid containers

tcl3:
Previously, the UA stylesheet gave range inputs a definite `width: 20ch`. This prevented range inputs from filling their grid track. We now source the default width and height from a new `RangeInputBox` type. This approach matches the implementation of other input types.

Here's an example from https://css-pattern.com/cubes-illusion/:

Before:

image

After:

image


AtkinsSJ (approved):

AtkinsSJ:
Looks like this needs rebasing to pick up the fixed lint CI.

PR #9728, opened 2026-05-29
LibWeb: Recompute @font-face descriptors on viewport size change

sideshowbarker:
**Problem:** The `font-face-descriptor-relative-length-iframe.html` reftest has recently been failing on most CI runs.

**Cause:** `@font-face` descriptors are evaluated once, at stylesheet-attach time, and involves a read of the document’s navigable viewport. For an iframe srcdoc, that read happens while the iframe’s `content_navigable` still has its default 0x0 viewport. The real size is only set later. The cached weight range is therefore wrong, and `set_viewport_size` already invalidates style and re-evaluates `@media` rules on the ensuing size change. But it never revisits `@font-face` descriptors — so the wrong weight stays cached for the life of the document.

**Fix:** `set_viewport_size` now walks the document’s `FontFaceSet` and asks each CSS-connected `FontFace` to re-evaluate its descriptors against the updated viewport. If any cached descriptor changed, the face is re-keyed in `FontComputer`’s `m_font_faces` map. So, subsequent font-matching sees the new values. A new helper lets the re-key path remove the stale entry without firing a second `did_load_font` invalidation; the `register_font_face` that follows already invalidates once.

### DIFF
--- a/Libraries/LibWeb/CSS/FontComputer.cpp
+++ b/Libraries/LibWeb/CSS/FontComputer.cpp
@@ -796,12 +796,17 @@ void FontComputer::unregister_font_face(GC::Ref<FontFace> face)
         .slope = face->declared_slope(),
         .width = face->declared_width(),
     };
+    unregister_font_face_with_key(face, key);
+    did_load_font(key.family_name);
+}
+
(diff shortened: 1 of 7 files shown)


sideshowbarker:
Moved to https://github.com/LadybirdBrowser/ladybird/pull/9730

PR #9739, opened 2026-05-29
UI/Qt: Put DevTools banner in its own widget, not a status bar

AtkinsSJ:
Using the status bar for this meant that the "DevTools is enabled on port X" message would get overwritten whenever the user hovered over a menu item. We don't otherwise use or need a status bar, so replace it with an explicit banner widget that's only used for this purpose. Keep the existing styling so it looks the same as before.

### DIFF
--- a/UI/Qt/BrowserWindow.cpp
+++ b/UI/Qt/BrowserWindow.cpp
@@ -19,6 +19,7 @@
 #include <UI/Qt/Application.h>
 #include <UI/Qt/BrowserWindow.h>
 #include <UI/Qt/ChromeStyle.h>
+#include <UI/Qt/DevToolsBanner.h>
 #include <UI/Qt/Icon.h>
 #if defined(AK_OS_MACOS)
 #    include <UI/Qt/MacWindow.h>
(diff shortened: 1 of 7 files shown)


UI/Qt/BrowserWindow.cpp
```diff
@@ -447,7 +448,33 @@ BrowserWindow::BrowserWindow(Vector<URL::URL> const& initial_urls, IsPopupWindow
...
     m_tabs_container->set_new_tab_action(m_new_tab_action);
 
-    setCentralWidget(m_tabs_container);
+    auto* main_widget = new QWidget(this);
+    auto* main_layout = new QVBoxLayout(main_widget);
+    main_layout->setContentsMargins(0, 0, 0, 0);
+    main_layout->setSpacing(0);
+    main_layout->addWidget(m_tabs_container, 1);
+
+    m_devtools_banner = new QWidget(main_widget);
```
trflynn89: Can we make the banner a proper widget subclass in its own file, like the find-in-page widget? IMO there's already too much going on in BrowserWindow.cpp

AtkinsSJ: Sure! Makes sense. I'll take a look next week probably.

AtkinsSJ:
Changes: It's a DevToolsBanner widget now.

Now that we have vertical tabs, there's a decision to make about whether this should span the entire window or just the web page area. I've gone with it spanning the whole window as it's a property of Ladybird rather than of a specific page.

Screenshot_20260601_112639

PR #9740, opened 2026-05-29
Tests: Make srcdoc content blocker test less flaky (hopefully)

AtkinsSJ:
Threw this at Codex because it's been timing out a lot recently on CI. I'm not 100% convinced by its explanations, but I figure that any time we can avoid an echo server, it'll be more stable if we do. I guess we'll see if we still get flakes. 😅

### DIFF
--- a/Tests/LibWeb/Text/input/content-blocker-srcdoc-frame-cosmetic-source.html
+++ b/Tests/LibWeb/Text/input/content-blocker-srcdoc-frame-cosmetic-source.html
@@ -1,38 +1,33 @@
 <!DOCTYPE html>
 <script src="./include.js"></script>
 <script>
-    asyncTest(async () => {
-        const server = httpTestServer();
-        const targetURL = await server.createEcho("GET", "/content-blocker-srcdoc-frame-cosmetic-source.html", {
-            status: 200,
-            headers: {
-                "Content-Type": "text/html",
-            },
-            body: `
-                <!DOCTYPE html>
-                <iframe id="frame" srcdoc="<div class='ad'></div>"></iframe>
-                <script>
-                    let output = "";
-                    function println(line) {
-                        output += line + "\\n";
-                    }
+    asyncTest(async done => {
+        // The srcdoc iframe inherits its base URL from its parent document. We need it to be localhost so that the
+        // content blocker's matching rule will apply to it.
+        const base = document.createElement("base");
+        base.href = "http://localhost/";
+        document.head.appendChild(base);
 
-                    window.addEventListener("load", () => {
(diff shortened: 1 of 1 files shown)


Tests/LibWeb/Text/input/content-blocker-srcdoc-frame-cosmetic-source.html
```diff
@@ -1,38 +1,31 @@
...
-                <!DOCTYPE html>
-                <iframe id="frame" srcdoc="<div class='ad'></div>"></iframe>
-                <script>
-                    let output = "";
-                    function println(line) {
-                        output += line + "\\n";
-                    }
+    asyncTest(async done => {
+        const base = document.createElement("base");
+        base.href = "http://localhost/content-blocker-srcdoc-frame-cosmetic-source.html";
```
sideshowbarker:
```suggestion
        base.href = "http://localhost/";
```

Tests/LibWeb/Text/input/content-blocker-srcdoc-frame-cosmetic-source.html
```diff
@@ -1,38 +1,31 @@
...
-            },
-            body: `
-                <!DOCTYPE html>
-                <iframe id="frame" srcdoc="<div class='ad'></div>"></iframe>
-                <script>
-                    let output = "";
-                    function println(line) {
-                        output += line + "\\n";
-                    }
+    asyncTest(async done => {
```
sideshowbarker, line 4:
```suggestion
    asyncTest(async done => {
    // A srcdoc frame inherits its base URL from the parent, and cosmetic rules match on
    // that URL — so set the base to localhost to make 'localhost##.ad' apply in the frame.
```

sideshowbarker (approved):
Tested with several hundred runs — along with the sibling tests.

This clearly seems to be the right fix . But to make it more clear, you might want tweak the commit message to have the _“wait for the iframe load event”_ part be more prominent — since that’s what the real fix is (as opposed to the _“run entirely in the original test document instead of navigating to an echo-server page”_ part).

Maybe something like this:

```text
Tests: Make srcdoc content blocker test deterministic

This change fixes the test to wait for the iframe load event before
reading its contentDocument — so the frame is parsed before the test
inspects it.

The old page read the srcdoc frame's contentDocument from a window load
handler without waiting for the frame itself to load. When the frame was
not yet parsed, the .ad lookup returned null — and reading its computed
style threw before internals.signalTestIsDone() was ever called. And
that manifested as a runner timeout.

```

AtkinsSJ:
Thank you! Suggestions applied.

PR #9741, opened 2026-05-29
Everywhere: Use PCH files

R-Goc:
This PR enables the use of PCH files with clang based compilers. The full build went from 12m 43.33s to 6m 40.96s so a 1.9x speedup. Attached are ClangBuildAnalyzer build traces before and after all the changes. There is probably quite a bit left to optimize with PCH files alone, maybe even below 6 minutes but it is more difficult. This set of PCH files mostly targeted the very commonly included files, so low hanging fuit. Within each lib that has a significant enough number of file an additional PCH.h file could be created if there are some that are always internally included. This would improve iteration on those libs. Of course reducing unneded header inclusion would help too, but it is fairly hard to find these.

Further siginificant improvements to build time have a few avenues to explore:
- Merging test files to reduce the number of executables that have to be linked should be fairly easy, though slightly tedious. That could cut over a minute easily.
- With the increasing number of rust components we should also look into improving their compile speed as the amount of time they take is unproportionally large compared to their size. 
- A unity build would also improve build times by cutting down on parse times for headers outside the pch and reducing duplicated template instantiations. 

[build_trace.txt](https://github.com/user-attachments/files/28411159/build_trace.txt)
[pch_trace_v12.txt](https://github.com/user-attachments/files/28411160/pch_trace_v12.txt)


gmta:
Make sure to include benchmark results once you undraft this :)

R-Goc:
> Make sure to include benchmark results once you undraft this :)

Yeah, also had to fix an issue with quotes in cmake. It looks like this PR will require a full reconfigure as the ccache args were not actually updated on my machine without that.

nico:
(I don't have an opinion on this PR, but I'll point out that once this goes in, people building with clang locally won't see if they forget to include any of the headers in AK/PCH.h until they upload their diff to github and the gcc CI tells them.)

R-Goc:
> (I don't have an opinion on this PR, but I'll point out that once this goes in, people building with clang locally won't see if they forget to include any of the headers in AK/PCH.h until they upload their diff to github and the gcc CI tells them.)

Yes I'm aware of that. But realistically we include pretty much all of AK everywhere. GCC CI is quite fast (unlike Mac) so it shouldn't be too bad. Also auto complete in pretty much every editor automatically adds the required header so it's not that common to miss a header when developing.

Libraries/LibCore/CMakeLists.txt
```diff
@@ -95,6 +95,10 @@ endif()
 ladybird_lib(LibCore core EXPLICIT_SYMBOL_EXPORT)
 target_link_libraries(LibCore PRIVATE LibSync Threads::Threads LibUnicode LibURL LibTextCodec)
 
+if (LADYBIRD_ENABLE_PCH)
+    target_precompile_headers(LibCore PRIVATE ${LADYBIRD_SOURCE_DIR}/AK/PCH.h)
```
cqundefine: Random idea: Why not move those calls for libraries to `ladybird_lib`, and allow the caller to pass an alternative precompiled header via an argument.

Meta/CMake/compile_options.cmake
```diff
@@ -194,7 +202,11 @@ include(${CMAKE_CURRENT_LIST_DIR}/sanitizers.cmake)
 include(CheckPIESupported)
 check_pie_supported(LANGUAGES CXX)
 if(CMAKE_CXX_LINK_PIE_SUPPORTED)
-    set(CMAKE_POSITION_INDEPENDENT_CODE ON)
+    # PIE doesn't really matter for us since all our logic is in libraries,
+    # This allows for shared PCH files between libraries and executables
```
cqundefine, line 206: This probably deservers a note why this is a problem (as far as I understand from the CMake discussion you linker on Discord, it's because of the -fPIE and -fPIC difference)

Meta/CMake/targets.cmake
```diff
@@ -154,6 +154,13 @@ function(ladybird_test test_src sub_dir)
     ladybird_windows_bin(${LADYBIRD_TEST_NAME} CONSOLE)
     ladybird_generate_dsym(${LADYBIRD_TEST_NAME})
 
+    if (LADYBIRD_ENABLE_PCH)
+        if (NOT LADYBIRD_TEST_NAME MATCHES TestYUVData)
```
cqundefine: Maybe `ladybird_test` also deserves a `SKIP_PCH` argument to avoid hardcoding anything here.

cqundefine (commented):
Some random ideas on how to make this neater, probably fine to ignore those if you think it's better the way it is right now.

Meta/CMake/compile_options.cmake
R-Goc: I'll add one.

Meta/CMake/targets.cmake
R-Goc: Yeah that would be better. I don't know how to write that in cmake though, maybe AI will help.

cqundefine: To add an option you can just modify the `cmake_parse_arguments` in `ladybird_test`, the first `""` in there are the options that don't have values (like this one), you can just make it `"SKIP_PCH"`, and the you should just have a `LADYBIRD_TEST_SKIP_PCH` available as boolean. Kind of like `EXPLICIT_SYMBOL_EXPORT` in `ladybird_lib`.

gmta:
Benchmarked this PR on my MacBook M4 Pro, 14 threads, `RelWithDebInfo`. I've excluded vcpkg configure+install. Ran them with and without ccache enabled.

### Full Builds

| Scenario | Baseline | PCH PR | Delta |
|---|---:|---:|---:|
| No ccache | 249.30s | 199.05s | 20.2% faster |
| Ccache enabled, cold cache | 286.44s | 231.15s | 19.3% faster |
| Ccache enabled, warm cache | 22.65s | 23.81s | 5.1% slower |

### Incremental Builds

| Edit | Baseline no ccache | PCH no ccache | Delta | Baseline ccache warm | PCH ccache warm | Delta |
|---|---:|---:|---:|---:|---:|---:|
| `LibWeb/DOM/Document.h` | 59.50s | 39.93s | 32.9% faster | 65.69s | 46.66s | 29.0% faster |
| `LibGfx/Bitmap.h` | 134.78s | 145.01s | 7.6% slower | 145.45s | 182.47s | 25.5% slower |
| `AK/String.h` | 231.42s | 172.45s | 25.5% faster | 252.88s | 204.17s | 19.3% faster |
| `LibWeb/DOM/Document.cpp` | 11.49s | 10.89s | 5.2% faster | 11.09s | 10.81s | 2.5% faster |

### Disk Usage

| Measurement | Baseline | PCH PR | Delta |
|---|---:|---:|---:|
| Build dir | 2033 MiB | 2460 MiB | +427 MiB |
| `.pch` files in build dir | 0 MiB | 406 MiB | +406 MiB |
| Final `lib/` dir | 123 MiB | 124 MiB | +1 MiB |
| Final `bin/` dir | 44 MiB | 44 MiB | 0 MiB |

R-Goc:
How am I getting 90% with you getting 20%. What on earth. Could you run ClangBuildAnalyzer and take a look?

gmta:
> How am I getting 90% with you getting 20%. What on earth. Could you run ClangBuildAnalyzer and take a look?

Can you share some details of your benchmark setup first? Those are missing from the PR description / commit messages.

R-Goc:
Well one part is ClangBuildAnalyzer. There I can see frontend time going down from 14k seconds to 4.5k, along with a small improvement in backend time. So total time goes from around 17k secodns to 7.5k seconds. I was adding the top headers from clang build analyzer to the PCH files. For the times I gave I just ran ninja, after a clean with CCACHE_DISABLE=1. So there was no configure time, and rust only built the final binary, not the deps. Could you also maybe share the trace from the ninja build log where you get the 20% for the full build? I am using clang 23 built from git, fully optimized like a release.

gmta:
@R-Goc thanks - what machine and build environment?

R-Goc:
Windows, building natively. The default release profile, so a dynamic build. Nearly latest toolchain.

R-Goc:
As for the machine it's an intel core ultra 155h with 32GB of ram and a gen 4 nvme SSD.

gmta:
> Windows, building natively

Could the difference be explained in terms of filesystem performance? As far as I know, the Windows filesystem stack is notoriously slow compared to macOS / Linux.

gmta:
But honestly, I'm less concerned about the "just 20% faster", I expect that incremental builds being slower with a warm ccache is going to be the most common for me.

gmta:
ClangBuildAnalyzer logs:

* [baseline-clang-build-analyzer.txt](https://github.com/user-attachments/files/28506339/baseline-clang-build-analyzer.txt)
* [pch-clang-build-analyzer.txt](https://github.com/user-attachments/files/28506340/pch-clang-build-analyzer.txt)

R-Goc:
I am using a dev drive so it shouldn't be too bad, with proper AV exclusions as well. If you run ClangBuildAnalyzer we will probably know by the difference in the ratio of frontend to backend time. Also it's possible the MS STL headers are heavier, but iirc we try not to include heavy STL headers anyways. I don't have a build set up in wsl to test, and that might not be representative either. Edit: ninja'd

R-Goc:
Huh. Your parse times are way lower. So that may be filesystem perf. On the other hand frontend time is a bigger component than on windows relatively. Which doesn't make much sense to me.

R-Goc:
> But honestly, I'm less concerned about the "just 20% faster", I expect that incremental builds being slower with a warm ccache is going to be the most common for me.

They shouldn't be unless you actually touch a header that is in the PCH. That was the whole point. So like the AK string.h row I don't understand at all. Did you just touch the file? If you do that with ccache on the only thing that is being benchmarked is the speed of ccache. To get a number for an actual rebuild you would need to make a change that changes the hash of the file, even a comment should do.

gmta:
> Did you just touch the file?

No, I added a comment to those files triggering the rebuild.

> They shouldn't be unless you actually touch a header that is in the PCH.

Including transitive header includes, right?

R-Goc:
Yes including transitive includes. But a change to String.h should cause a real rebuild so I don't get how you got that ccache warm number. That looks like a ccache benchmark.  Comparing CCACHE_DISABLE to ccache cold is interesting. I didn't know the overhead to store files was that large. Or rather I don't understand the ccache warm columns at all.

gmta:
> But a change to String.h should cause a real rebuild

Ah a comment change should not affect the build since ccache strips comments during preprocessing afaik. I'll check if I can update it with an actual code change.

R-Goc:
> Ah a comment change should not affect the build since ccache strips comments during preprocessing afaik. I'll check if I can update it with an actual code change.

Makes sense. As for the size difference, it looks like the size on disk is just big. Each instance of the PCH is 16MB. LibWeb is 70, LibJS is 26. The size is from the .pch file itself. So I guess whatever format clang uses is not compressed at all, and maybe it is just the raw AST.

R-Goc:
Looks like the Gfx/Bitmap.h regression comes from it being included in the PCH transitively by quite a few of the sources.

gmta:
I've updated the incremental build times with edits to the code instead of adding a comment, which shows somewhat better results for especially `String.h` and `Document.h`. `Bitmap.h` is still a fairly big regression.

R-Goc:
I don't think Bitmap.h is a good benchmark. I looked at the history of it and its includes and it is changed very rarely. If there is an actually often changed header with a bad regression like that I will certainly pull it out.

gmta:
> I don't think Bitmap.h is a good benchmark. I looked at the history of it and its includes and it is changed very rarely.

There are roughly 500 transitive includes in the PCH files, in which there were more than 350 changes in the last 6 months. I can pick a more suitable candidate but I doubt it changes the picture much :)

R-Goc:
> > I don't think Bitmap.h is a good benchmark. I looked at the history of it and its includes and it is changed very rarely.
> 
> There are roughly 500 transitive includes in the PCH files, in which there were more than 350 changes in the last 6 months. I can pick a more suitable candidate but I doubt it changes the picture much :)

How much of that is AK? I can get rid of the headers that aren't included literally everywhere if you prefer.

gmta:
> > > I don't think Bitmap.h is a good benchmark. I looked at the history of it and its includes and it is changed very rarely.
> > 
> > 
> > There are roughly 500 transitive includes in the PCH files, in which there were more than 350 changes in the last 6 months. I can pick a more suitable candidate but I doubt it changes the picture much :)
> 
> How much of that is AK? I can get rid of the headers that aren't included literally everywhere if you prefer.

About 50 changes, ~15%. But isn't the goal to decrease build times by including the headers that are included everywhere?

R-Goc:
Looked at this today and to remove Gfx/Bitmap I have to remove DOM/Element.h, HTML/HTMLElement.h, HTML/ImageBitmap.h, Painting/PaintStyle.h, SVG/SVGImageElement.h, Layout/Box.h, DOM/ShadowRoot.h. Honestly with that huge number of includes I'm surprised it's slower for you, I guess these all get included in a certain subset of files, so aren't as spread out. So I'm not 100% sure it's worth the tradeoff. I looked at what's included in Bitmap and it seems stable.

gmta:

PR #9743, opened 2026-05-29
LibIPC: Avoid InitTransport deadlocks on Windows during IPC communication

heitke:
This PR fixes Ladybird to run natively on Windows again.

Background:
Ladybird uses a parent/child process architecture with lots of
inter-process communication (IPC) messages. On Windows, the message
handler currently runs synchronously and is therefore susceptible to
deadlocking if messages are sent before initialization is complete.

This is currently happening when the WebContent child process starts
up, causing Ladybird to crash or hang on startup.

PR 9450
(https://github.com/LadybirdBrowser/ladybird/pull/9450)
attempts to fix this in a perhaps preferable way by changing to an
asynchronous Windows socket implementation. However, this PR attempts
to solve the problem while maintaining the current architecture.

Changes:
Instead of sending an InitTransport message when spawning a new child
process, this PR uses the existing SOCKET_TAKEOVER environment
variable to pass ::<parent_pid> instead of only passing
(file descriptor).

I'd argue that this is acceptable, and maybe even preferred, because
all child processes are likely to need the parent pid anyway.

Unfortunately, the implementation was not quite that simple because
Ladybird also creates secondary connections between child processes.
For all secondary connections, this PR continues to process the
InitTransport message on Windows to exchange pids.

Alternatives:
I also considered having the WebContent child process delay creating
the tab until startup finished so the message loop could start
successfully. I decided this was too large of a departure from the
current architecture, and it would leave us vulnerable to similar
problems in the future if we do not address the broader concern.

I also considered passing the pid via a command-line argument or via
a new environment variable, but decided that extending the
SOCKET_TAKEOVER method was more elegant.

### DIFF
--- a/Libraries/LibCore/SystemServerTakeover.cpp
+++ b/Libraries/LibCore/SystemServerTakeover.cpp
@@ -14,6 +14,7 @@ namespace Core {
 
 HashMap<ByteString, int> s_overtaken_sockets {};
 bool s_overtaken_sockets_parsed { false };
+Optional<pid_t> s_parent_pid {};
 
 static void parse_sockets_from_system_server()
 {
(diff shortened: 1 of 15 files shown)


heitke:
Upon more reflection, I really think we need an async Windows socket implementation. Otherwise we are going to have too many future issues. Converted this to a draft for now while I investigate that strategy more before requesting a merge on this one.

heitke:
Implemented a simpler, more robust solution here https://github.com/LadybirdBrowser/ladybird/pull/9770

PR #9748, opened 2026-05-30
LibWeb: Position gradient color stops that carry a transition hint but no explicit position

bdr99:
This fixes an issue where a linear gradient with a transition hint between two otherwise-unpositioned color stops rendered with a hard color jump. The auto-positioning loop in resolve_color_stop_positions was skipping past the stop that ended the run. When that stop carried only a transition hint (and no resolved position of its own), it acted as a positional anchor for the preceding run but was then never given a position itself, leaving its position as NaN and collapsing the hint to 0 downstream.

The fix was to resume the loop from that stop so it is used as the beginning of the next run and gets evenly spaced against the following anchor.

The issue can be seen at https://esbuild.github.io/gradient-tests/ (example 6 on the page).

Before:

image

After:

image

Also added a screenshot test to verify the fix.


gmta:

PR #9749, opened 2026-05-30
LibWeb: Clear detached rendering state on removal

awesomekling:
This fixes memory growth seen when repeatedly switching between the Conversation and Commits tabs on GitHub pull requests.

The main issue was that detached DOM/layout state could keep expensive rendering data alive after it no longer participated in layout. This now clears paintables when layout nodes detach and clears cached computed style from elements when they leave the tree, including synthetic pseudo-elements.


Libraries/LibWeb/DOM/Element.h
```diff
@@ -390,6 +390,7 @@ class WEB_API Element
     bool has_synthetic_pseudo_elements() const;
     void clear_synthetic_pseudo_element_layout_nodes(Badge<Layout::TreeBuilder>) { clear_synthetic_pseudo_element_layout_nodes(); }
     void clear_synthetic_pseudo_element_layout_nodes(Badge<Document>) { clear_synthetic_pseudo_element_layout_nodes(); }
+    void clear_synthetic_pseudo_element_layout_nodes(Badge<Node>) { clear_synthetic_pseudo_element_layout_nodes(); }
```
tcl3: nit: Could this be done with a template: `template<OneOf<Layout::TreeBuilder, Document, Node> T>`?

PR #9750, opened 2026-05-30
LibWeb: Preserve CryptoKey algorithm types during serialization

shannonbooth:
CryptoKey structured serialization was cloning the internal algorithm slot as a generic JS object. After deserialization, WebCrypto operations such as exportKey() expected the slot to still be a native KeyAlgorithm subclass and crashed when downcasting it.

Serialize the stored key algorithm with an explicit tag and reconstruct the appropriate KeyAlgorithm subclass during deserialization. Also keep EC private-key serialization aligned by encoding nested public keys as EC public-key payloads rather than tagged top-level handles.


shannonbooth:
Fixes 38 WPT crashes 😅

PR #9751, opened 2026-05-30
LibWeb: Use retained display lists for hit testing

awesomekling:
This replaces recursive paint-order hit testing with a retained hit-test display list recorded alongside painting. The list records hit-testable boxes, text fragments, chrome widgets, caret rects, visual context indices, and line metadata, then uses a spatial index to limit point queries to nearby candidates.

The branch also moves selection and caret placement onto the retained data, adds `document.caretPositionFromPoint()`, and includes a debug overlay for inspecting caret hit-test placement.

Coverage includes hit testing for rounded regions, large unbucketed boxes, scrollbars after display-list invalidation, selection across non-text content, caret placement near line edges, and `caretPositionFromPoint()` shadow-root retargeting.

This *greatly* reduces CPU usage while moving the mouse around on all websites. :^)


Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -7206,6 +7207,55 @@ GC::RootVector<GC::Ref<Element>> Document::elements_from_point(double x, double
     return sequence;
 }
 
+// https://drafts.csswg.org/cssom-view/#dom-document-caretpositionfrompoint
```
tcl3: Was this spec link supposed to go on the method below? 

Also, the steps from this link don't match `Document::caret_position_from_point()` exactly.

PR #9756, opened 2026-05-30
LibWeb+UI/AppKit: Implement macOS IME support

### LINKED ISSUE #9712: Unable to use input method on MacOS
### Summary

When using system non-English input method to type-in words on web pages, the candidate window appeared in the left-bottom corner and the the selected text cannot be inserted into the input box.

Image

### Operating system

macOS

### Steps to reproduce

1. Open a html file with an input element
2. Switch to non-English input method
3. Start typing

### Expected behavior

The candidate window should appear near the input element, and the selected word should be inserted correctly.

### Actual behavior

The candidate window appeared in the left-bottom corner and the the selected text cannot be inserted into the input box.

Just a simplest html file with input element.

```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <title>Test Page</title>
</head>
<body>
    <div class="container">
        <input>
    </div>
</body>
</html>
```

```shell
(ladybird) lkl@LKLMacBook-Air ladybird % ./Meta/ladybird.py run
ninja: Entering directory `/Users/lkl/Code/ladybird/Build/release'
[0/2] Re-checking globbed directories...
ninja: no work to do.
174982.596 RequestServer(26664): Setting proxy mode to system
2026-05-28 19:20:55.945 Ladybird[26663:3249767] error messaging the mach port for IMKCFRunLoopWakeUpReliable
```

Image

_No response_

### Contribute a patch?

sideshowbarker:
This makes macOS IME input in web content work as expected. Fixes https://github.com/LadybirdBrowser/ladybird/issues/9712.

### DIFF
--- a/Libraries/LibWeb/DOM/Document.cpp
+++ b/Libraries/LibWeb/DOM/Document.cpp
@@ -8306,6 +8306,63 @@ GC::Ptr<DOM::Position> Document::cursor_position() const
     return nullptr;
 }
 
+Optional<CSSPixelRect> Document::current_caret_rect()
+{
+    // Returns the bounds of the current text caret in viewport-relative CSS pixels. Used to position platform overlays
+    // such as the IME candidate window. Returns nothing when no editable element is focused or when layout isn't ready.
+    auto position = cursor_position();
+    if (!position)
+        return {};
+    auto& dom_node = *position->node();
+
+    update_layout(UpdateLayoutReason::InputCaretRect);
+
+    auto* layout_node = dom_node.layout_node();
+    if (!layout_node)
+        return {};
+
+    // The caret rects computed here are document-relative (absolute). Platform IME overlays are positioned relative to
+    // the viewport — so translate by scroll offset and map through any containing navigables to the top-level viewport.
+    auto to_viewport_rect = [this](CSSPixelRect rect) -> CSSPixelRect {
+        auto navigable = this->navigable();
+        if (!navigable)
+            return rect;
+        auto scroll = navigable->viewport_scroll_offset();
+        CSSPixelRect viewport_rect { rect.x() - scroll.x(), rect.y() - scroll.y(), rect.width(), rect.height() };
+        return navigable->to_top_level_rect(viewport_rect);
+    };
+
+    // Walk up to the nearest PaintableWithLines, which is where text fragments live.
+    Painting::PaintableWithLines const* paintable_with_lines = nullptr;
+    for (auto paintable = layout_node->first_paintable(); paintable; paintable = paintable->parent()) {
+        if (auto const* with_lines = as_if<Painting::PaintableWithLines>(*paintable)) {
+            paintable_with_lines = with_lines;
+            break;
+        }
+    }
+
+    if (paintable_with_lines) {
+        for (auto const& fragment : paintable_with_lines->fragments()) {
+            if (fragment.layout_node().dom_node() != &dom_node)
+                continue;
+            auto const offset = position->offset();
+            if (offset < fragment.dom_start_offset_in_node() || offset > fragment.dom_end_offset_in_node())
+                continue;
+            return to_viewport_rect(fragment.range_rect(Painting::Paintable::SelectionState::StartAndEnd, offset, offset));
+        }
+    }
+
+    // Empty editable elements have no fragments; fall back to the padding-box corner.
+    if (auto* node_with_style = as_if<Layout::NodeWithStyleAndBoxModelMetrics>(*layout_node)) {
+        auto paintable = node_with_style->first_paintable();
+        if (auto const* box = as_if<Painting::PaintableBox>(paintable.ptr())) {
+            auto content_box = box->absolute_padding_box_rect();
+            return to_viewport_rect(CSSPixelRect { content_box.x(), content_box.y(), 1, node_with_style->computed_values().line_height() });
+        }
+    }
+    return {};
+}
+
 void Document::reset_cursor_blink_cycle()
 {
     m_cursor_blink_state = true;
--- a/Libraries/LibWeb/DOM/Document.h
+++ b/Libraries/LibWeb/DOM/Document.h
@@ -127,6 +127,7 @@ enum class InvalidateLayoutTreeReason {
     X(InspectAccessibilityTree)              \
     X(InspectDOMTree)                        \
     X(InspectDevToolsLayoutData)             \
+    X(InputCaretRect)                        \
     X(InternalsHitTest)                      \
     X(MediaQueryListMatches)                 \
     X(NavigableSelectedText)                 \
@@ -965,6 +966,7 @@ public:
     InputEventsTarget* active_input_events_target(DOM::Node const* for_node = nullptr);
     GC::Ptr<DOM::Position> cursor_position() const;
     void set_cursor_position_needs_repaint();
+    Optional<CSSPixelRect> current_caret_rect();
 
     bool cursor_blink_state() const { return m_cursor_blink_state; }
 
--- a/Libraries/LibWeb/HTML/Navigable.cpp
+++ b/Libraries/LibWeb/HTML/Navigable.cpp
@@ -21,6 +21,7 @@
 #include <LibWeb/DOM/DocumentLoading.h>
 #include <LibWeb/DOM/Element.h>
 #include <LibWeb/DOM/Event.h>
+#include <LibWeb/DOM/Position.h>
 #include <LibWeb/DOM/Range.h>
 #include <LibWeb/DOM/Text.h>
 #include <LibWeb/Fetch/Fetching/Fetching.h>
@@ -329,6 +330,7 @@ void Navigable::visit_edges(Cell::Visitor& visitor)
     visitor.visit(m_page);
     visitor.visit(m_parent);
     visitor.visit(m_active_document);
+    visitor.visit(m_input_method_composition_node);
     visitor.visit(m_container);
     m_event_handler.visit_edges(visitor);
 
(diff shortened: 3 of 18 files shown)


kunlinglio:
Things are get kind of tricky.
- The simple input element worked on AppKit version but not on Qt version, just like @VentusUta [mentioned](https://github.com/LadybirdBrowser/ladybird/issues/9712#issuecomment-4583130845).
- And even for AppKit version, if I keep the cursor inside the google's input bar, keep the input bar empty and switch the input method, the ladybird got stuck forever. (A really strange bug, and cannot reproduce in a simple html input element)

https://github.com/user-attachments/assets/5ae28490-6d03-4d74-8cf7-220c48142a60

sideshowbarker:
> if I keep the cursor inside the google's input bar, keep the input bar empty and switch the input method, the ladybird got stuck forever. (A really strange bug, and cannot reproduce in a simple html input element)

Can you reproduce that unexpected behavior with a build from the master branch? Or instead is it clearly a problem/regression introduced by the new code in this branch?

kunlinglio:
I have checked that it is a regression introduced by this branch. It works will with the master version (de9fbf1679da5d65e9884c3ce025bdbd90dfabdd).

VentusUta:
> > if I keep the cursor inside the google's input bar, keep the input bar empty and switch the input method, the ladybird got stuck forever. (A really strange bug, and cannot reproduce in a simple html input element)
> 
> Can you reproduce that unexpected behavior with a build from the master branch? Or instead is it clearly a problem/regression introduced by the new code in this branch?

I tested it with a build from master branch code + 5fb452d167f4e578e4b2aed4949c3b275f531335.

The stuck can be reproduced, but seems also occurs without emptying the search box.

Here is a screen recording:

https://github.com/user-attachments/assets/f185ed2b-925d-48ec-ad7f-20bbda118b7e

sideshowbarker:
@kunlinglio, @VentusUta Thanks much — I was able to reproduce the hang, and to figure out the cause, and to fix it. I think. So please re-test with the latest commit on this branch, https://github.com/LadybirdBrowser/ladybird/pull/9756/changes/78b29000a223c8c1fbfb5485b22136c0e94222fd.

kunlinglio:
Ah, it works! And the candidate window can track the cursor now, which is even better than you have mentioned. :)

One small details: There is some latency in the candidate window. It seems track the last position of the cursor. (Maybe this is the next PR's goal?)

Screenshot 2026-05-31 at 18 35 34

But whatever, it looks good to me. Thanks!

sideshowbarker:
> One small details: There is some latency in the candidate window. It seems track the last position of the cursor. 

Yeah, that’s a somewhat-anticipated artifact of the v1 design.

> (Maybe this is the next PR's goal?)

Yes. Over the long term, we will need to figure out how to get that latency down. And I have some ideas for that. But for the near term yeah, it seems better to not block on that, and instead to get this initial PR landed first — so that others can try it and we can iterate on it further after we get more feedback from more people having actually used it.

Libraries/LibWeb/HTML/Navigable.cpp
```diff
@@ -3342,6 +3342,34 @@ void Navigable::paste(Utf16String const& text)
...
+    auto document = active_document();
+    if (!document || !document->is_fully_active())
+        return;
+    auto* target = document->active_input_events_target();
+    if (!target)
+        return;
+
+    if (replace_before_caret_in_code_units > 0) {
+        // Extend the selection backward to cover the previous marked text. handle_insert then replaces the selection.
+        if (auto* form_control = dynamic_cast<FormAssociatedTextControlElement*>(target)) {
```
shannonbooth:
```suggestion
        if (auto* form_control = as_if<FormAssociatedTextControlElement>(target)) {
```

Libraries/LibWeb/DOM/Document.cpp
```diff
@@ -8306,6 +8306,54 @@ GC::Ptr<DOM::Position> Document::cursor_position() const
...
 }
 
+Optional<CSSPixelRect> Document::current_caret_rect()
+{
+    // Returns the bounds of the current text caret in document-relative CSS pixels. Used to position platform overlays
+    // such as the IME candidate window. Returns nothing when no editable element is focused or when layout isn't ready.
+    auto position = cursor_position();
+    if (!position)
+        return {};
+    auto* dom_node = position->node().ptr();
```
shannonbooth: `position->node()` returns a GC::Ref, so this should not be nullable

Libraries/LibWeb/Internals/Internals.cpp
```diff
@@ -418,6 +418,25 @@ String Internals::selected_text_for_clipboard()
...
+}
+
+JS::Object* Internals::current_caret_rect()
+{
+    auto& active_document = window().associated_document();
+    auto rect = active_document.current_caret_rect();
+    if (!rect.has_value())
+        return nullptr;
+    auto result = JS::Object::create(realm(), nullptr);
+    result->define_direct_property("x"_utf16_fly_string, JS::Value(rect->x().to_double()), JS::default_attributes);
```
shannonbooth: I think it would be a little nicer if currentCaretRect IDL returned a dictionary so the values could be returned directly.

UI/AppKit/Interface/LadybirdWebView.mm
```diff
@@ -1406,25 +1438,48 @@ - (void)doCommandBySelector:(SEL)selector
...
+    return NSMakeRange(self.marked_text_length, 0);
 }
 
 - (void)setMarkedText:(id)string selectedRange:(NSRange)selectedRange replacementRange:(NSRange)replacementRange
 {
+    // Called by the input method as the user types each composing character. We "fake" inline preedit by actually
+    // inserting the marked text into the focused editable, then replacing it on each subsequent setMarkedText or on
+    // commit (insertText) or abort (unmarkText). The replace count is self.marked_text_length — the length of marked
+    // text we previously inserted.
+    NSString* preedit = [string isKindOfClass:[NSAttributedString class]] ? [(NSAttributedString*)string string] : (NSString*)string;
```
shannonbooth, line 1461: I wonder if it is possible for appkit to only know about:

set_marked_text_from_input_method(text);
commit_text_from_input_method(text);
unmark_text_from_input_method();


Type of thing instead of knowing text length, and let LibWeb deal with the replacement for the editable (which might make the Qt implementation smaller too?)

Libraries/LibWeb/DOM/Document.h
```diff
@@ -965,6 +966,9 @@ class WEB_API Document
     InputEventsTarget* active_input_events_target(DOM::Node const* for_node = nullptr);
     GC::Ptr<DOM::Position> cursor_position() const;
     void set_cursor_position_needs_repaint();
+    // Returns the bounds of the current text caret in document-relative CSS pixels. Used by the UI to position platform
```
shannonbooth: Is document-relative the same offset used in AppKit? I get confused any time I think about offsets and all the different types 😅

shannonbooth (commented):
Left some comments, but it's not an area I am super familiar with myself. Mostly wondering about where responsibilities like between the UI and LibWeb for this (my model of how IME works is also not very good!)

UI/AppKit/Interface/LadybirdWebView.mm
sideshowbarker: Beautiful, yes — made it so. Please take a look.

Libraries/LibWeb/DOM/Document.h
sideshowbarker: Bug. Fixed.

Libraries/LibWeb/Internals/Internals.cpp
sideshowbarker: Well, I couldn’t find anything else in the existing IDL that _returns_ a dictionary. Am I missing something? So for now at least, I switched it to return `DOMRect?` — which is what `getBoundingClientRect()` returns. And it seems like a good fit for this too?

Libraries/LibWeb/DOM/Document.cpp
sideshowbarker: Thanks — fixed

Libraries/LibWeb/HTML/Navigable.cpp
sideshowbarker: This factored out in the refactor for moving the marked-text state and replacement into LibWeb.

sideshowbarker:
@kunlinglio, @VentusUta I pushed an update for this that involved a major refactoring. So if y‘all have time to build and try it out again, that’d be much welcome.

Libraries/LibWeb/DOM/Document.h
shannonbooth: The comment here says "document relative" but the implementation says "viewport-relative" (probably this comment can be dropped?)

Libraries/LibWeb/Internals/Internals.h
```diff
@@ -18,6 +18,12 @@ class CSSStyleSheet;
 
 }
 
+namespace Web::Geometry {
```
shannonbooth: This is using the pattern above - but we can include `LibWeb/Forward.h` here instead

Libraries/LibWeb/DOM/Document.cpp
shannonbooth: nit but can we remove the .ptr() here? GC::Ref deletes the operator bool to null check

UI/AppKit/Interface/LadybirdWebView.mm
```diff
@@ -1406,25 +1438,49 @@ - (void)doCommandBySelector:(SEL)selector
...
 
 - (void)setMarkedText:(id)string selectedRange:(NSRange)selectedRange replacementRange:(NSRange)replacementRange
 {
+    // Called by the input method as the user types each composing character. We present inline preedit by inserting the
+    // marked text into the focused editable. LibWeb owns the marked-text range and replaces the previous preedit on
+    // each subsequent setMarkedText, or on commit (insertText) or abort (unmarkText). We keep marked_text_length only
+    // to answer the NSTextInputClient range queries (markedRange/selectedRange/hasMarkedText).
+    NSString* preedit = [string isKindOfClass:[NSAttributedString class]] ? [(NSAttributedString*)string string] : (NSString*)string;
+    auto utf8 = ByteString { preedit.length > 0 ? [preedit UTF8String] : "" };
+    m_web_view_bridge->set_marked_text_from_input_method(Utf16String::from_utf8(utf8));
```
shannonbooth, line 1463: This path for an empty string doesn't unset: m_input_method_composition_node, do we need to unset it for this case too?

UI/AppKit/Interface/LadybirdWebView.mm
```diff
@@ -120,6 +120,11 @@ @interface LadybirdWebView () <NSDraggingDestination>
 // To handle key events after dead key processing, we need to hold onto the originating key-down event.
 @property (nonatomic, strong) NSEvent* current_key_down_event;
 
+// Length of the marked text (input-method preedit) currently inserted into the focused editable. We "fake" inline
+// preedit by actually inserting the marked text into the input and then replacing it on each composition update. This
+// number tells WebContent how many code units to replace before the caret.
```
shannonbooth: This comment seems to be out of date after the change in flow between WebConent and the UI

Libraries/LibWeb/Internals/Internals.cpp
shannonbooth: Oh right, I guess specs generally don't do that since it wouldn't be very comfy. I suppose it _should_ work? But DOMRect is indeed nice

Libraries/LibWeb/DOM/Document.h
sideshowbarker: Dropped the comment.

Libraries/LibWeb/Internals/Internals.h
sideshowbarker: > This is using the pattern above - but we can include `LibWeb/Forward.h` here instead

Made it so.

Libraries/LibWeb/DOM/Document.cpp
sideshowbarker: > nit but can we remove the .ptr() here? GC::Ref deletes the operator bool to null check

Yes — made it so.

UI/AppKit/Interface/LadybirdWebView.mm
sideshowbarker: > This path for an empty string doesn't unset: m_input_method_composition_node, do we need to unset it for this case too?

Yep, thanks for catching that — fixed.

UI/AppKit/Interface/LadybirdWebView.mm
sideshowbarker: > This comment seems to be out of date after the change in flow between WebContent and the UI

Thanks — revised it to fix that.

shannonbooth (approved):

PR #9763, opened 2026-05-30
Tests: Merge tests

R-Goc:
This PR merges together multiple test files. The merge is done where it fits logically and the size of the file wouldn't be too huge. For example while merging all the string class tests would work it would be a huge file.


R-Goc:
Git isn't great about detecting renamed merged files. Maybe also TestUtf16View, TestUtf8View and TestStringView could be merged together, that would be around 1500 lines. That is about it in AK from the ones that logically fit together. Maybe Array, FixedArray and Vector make sense too but I don't have a good name for that. HashMap and HashTable too, but again not sure how to name it. Other than that we have a whole lot of tiny(under 100 lines) test files that have no connection with each other.

trflynn89:
Why? This is significantly worse organizationally IMO.

R-Goc:
> Why? This is significantly worse organizationally IMO.

If the PCH PR gets merged we will be spending 20% of the whole build time on compiling and linking tests. On my machine I measure it to 1m20s, with a total build of 6m40s. There is no reason for us to have 200 test executables. The number has been steadily increasing.

R-Goc:
I reconsidered the approach and will open a new PR with the other one.

PR #9770, opened 2026-05-30
LibIPC: Implement async Windows IPC sockets

heitke:
This PR fixes Ladybird to run natively on Windows again and finishes 
the work of https://github.com/LadybirdBrowser/ladybird/pull/9450.

Background:
Ladybird uses a parent/child process architecture with lots of
inter-process communication (IPC) messages. On Windows, the 
message handler currently runs synchronously and requires an 
InitTransport message to fully setup communication. **

This was causing a deadlock while starting up the WebContent
child process, causing Ladybird to crash or hang on startup.

Changes:
- TransportSocketWindows: Rewritten to use a dedicated I/O thread 
with non-blocking sockets, WSAPoll, and wakeup/notify socket-pairs
for async send/recv operations (mostly copied from PR 9450)

- Peer PID exchange: Added InitTransport handshake to Compositor
and WebWorker

- ResourceLoader/ImageCodecPlugin: Added proper shutdown 
sequencing to ensure IPC transports are closed before process exit

- Delays PageHost initialization to prevent a deadlock in the 
WebContent child process. During startup, a message gets sent
to the parent. We need to make sure we handle the InitTransport
on Windows before creating the new tab.

** No SCM_RIGHTS support means we need to transmit the peer 
pid to takeover and duplicate sockets.


OfficialKris:
I have tested this and it does work as advertised on my set of websites.

Neustradamus:
@heitke: Nice, good job!

gmta:

Issue #9776, opened 2026-05-31
WebSocket: a frame the server sends immediately after the handshake is intermittently lost


sideshowbarker:
When a WebSocket server sends a frame immediately after completing the opening handshake — before the client has sent anything — we intermittently never deliver that frame. `onopen` fires — but the expected `onmessage` never does, and the connection stalls until timeout.

This will affect behavior in Ladybird of any pages/sites which connect to a WebSocket server that pushes an initial message on connect (welcome banner, initial state, auth challenge, server-initiated protocol) — which seems like a relatively common real-world pattern in practice.

Also/however, it seems possible this is actually a bug in curl — rather than in our own code.

### Steps to reproduce

1. Run a (local) WebSocket echo server that, on connect, sends one text frame, and then echoes everything.
2. Multiple times: Load a page that connects and only sends from its `onmessage` handler (so it depends on the server’s first message):

   ```javascript
   const ws = new WebSocket("ws://localhost:PORT/");
   ws.onopen = () => println("open");
   ws.onmessage = (e) => { println("message " + e.data); ws.send("hi"); };
   ```

3. Observe that in some number of loads (~40%, in my own testing), `open` prints — then nothing: The server’s first frame is never delivered; the page hangs. (The in-tree `WebSocket/echo.html` test reproduced this, before it was reworked in #9777 to send first.)

### Details from debugging

Using curl `CURLOPT_VERBOSE` + `CURLOPT_DEBUGFUNCTION` + debug printfs:

- **PASS run:** handshake completes (`Received 101, Switching to WebSocket` / `Connection #0 left intact`), then `did_connect()` calls `curl_easy_recv` which returns the server frame _immediately_ (`nread=58`). It was read into curl’s buffer *during* the handshake, coalesced with the 101. Delivered; subsequent echoes work.
- **FAIL run:** identical through `Connection #0 left intact`. `did_connect()` → `curl_easy_recv` → `CURLE_AGAIN`. Then **nothing ever again**: no `[CURL DATA<]` is logged, `curl_easy_recv` keeps returning `CURLE_AGAIN`, and the read `Core::Notifier` on `CURLINFO_ACTIVESOCKET` never fires. After the per-test timeout, curl closes the connection.
- **Raw `recv(MSG_PEEK)`** on `CURLINFO_ACTIVESOCKET` returns `EAGAIN` repeatedly in the FAIL case. The bytes are _not_ sitting in the active socket’s OS receive buffer.
[...]

PR #9777, opened 2026-05-31
Tests: Run a local WebSocket echo server in place of the hosted one

sideshowbarker:
**Problem:** Our WebSocket tests were connecting to an externally-hosted echo server — so they depended on network reachability from our CI runners. Any network hiccups between a runner and that server timed the tests out — and that’s been a relatively-frequent source of CI flakes.

**Fix:** Teach our existing in-tree echo server to also speak WebSocket. It performs the RFC 6455 handshake, and echoes every text or binary frame back — on the same OS-assigned port already exposed to tests. Updated our tests to connect to that local server, not the remote one.

Also, updated our WebSocket `echo.html` to drive the exchange from its `onopen` handler — rather than waiting for a server-initiated greeting. That’s because: A frame the server sends in the moment right after the handshake — before the client has sent anything — is currently delivered unreliably by our (curl-backed) WebSocket client. See issue #9776.

### DIFF
--- a/Tests/LibWeb/Fixtures/http-test-server.py
+++ b/Tests/LibWeb/Fixtures/http-test-server.py
@@ -2,6 +2,7 @@
 
 import argparse
 import base64
+import hashlib
 import http.server
 import json
 import os
@@ -17,11 +18,13 @@ from typing import Optional
 
 """
 Description:
-    This script starts a simple HTTP echo server on localhost for use in our in-tree tests.
+    This script starts a simple HTTP and WebSocket echo server on localhost for use in our in-tree tests.
     The port is assigned by the OS on startup and printed to stdout.
 
 Endpoints:
     - POST /echo <json body>, Creates an echo response for later use. See "Echo" class below for body properties.
+    - GET <any path> with an "Upgrade: websocket" header: Performs a WebSocket handshake and then echoes
+      every text/binary frame back to the client verbatim.
 """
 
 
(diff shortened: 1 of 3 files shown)


Tests/LibWeb/Fixtures/http-test-server.py
```diff
@@ -380,6 +389,111 @@ def do_other(self):
         else:
             self.send_error(405, "Method Not Allowed")
 
+    # --- Minimal RFC 6455 WebSocket echo --------------------------------------------------------------
+    # Reached when a request carries an "Upgrade: websocket" header. The tests use this in place of an
+    # externally-hosted echo server, so they have no network dependency. The client sends first; every
+    # text/binary frame it sends is echoed back verbatim.
+
+    def _serve_websocket_echo(self):
```
shannonbooth, line 397: Not requesting any change here - definitely good to stop relying on a web server in tests.

But one thing which I've been considering is to try and integrate more with WPT infrastructure (see https://github.com/LadybirdBrowser/ladybird/pull/8540) rather than adding more stuff to http-test-server.py since that has many gaps which that test infrastructure also covers (and also I would also like to be able to contribute tests back to WPT easier). But I'm not really sure how others feel about that since #8540 is very large vendoring (and also doesn't even cover web sockets yet).

sideshowbarker: OK I hadn’t actually ever looked at #8540 yet but now that I have, I‘m pretty keen to see that get landed. I mean more for the feature itself, actually. But if that necessarily requires now doing the “Vendor WPT Python test server infrastructure” thing, then yeah it does seem like it’s time to go ahead and do that. I agree that’s gonna be better over the long run. I guess it does add some ongoing cost (I mean, after the initial cost of getting all that reviewed on landed) — but it seems like it will be a big net benefit.

PR #9782, opened 2026-05-31
LibWeb: Use first baseline for flex/grid items and containers

alexkuz:
_Disclaimer: this PR is made with a use of AI. I have no experience writing modern C++ code._

Flex and grid containers use the first baseline per [CSS specs](https://drafts.csswg.org/css-align-3/#baseline-values), while other containers use the last line baseline for CSS2.1 compatibility.

From [MDN: align-items CSS property](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/align-items) page:

| Before | After | Firefox |
|--------|--------|-------- |
| image | image |image |


alexkuz:
Turns out it's more nuanced than I thought (in particular inputs are using flexboxes in their shadow trees and this fix affects it). I don't want to fix it without full understanding of the spec so I'm closing it for now.

PR #9792, opened 2026-05-31
LibWeb: Suppress false-positive LSan report during WebWorker teardown

sideshowbarker:
**Problem:** For several weeks now, CI Sanitizer runs have been logging leaked RustCompiledRegex objects in the WebWorker process when a worker is torn down while an off-thread script compile is still in flight.

**Cause:** `compile_off_thread(`) (introduced in 4a7dc45b3f0) parses and compiles a fetched top-level script on a ThreadPool worker. The pool's a process-lifetime singleton whose workers are never joined before exit — and that leads to LSan reporting a false positive during teardown.

**Fix:** Suppress the false positive via `__lsan_default_suppressions()`.

### DIFF
--- a/AK/LsanSuppressions.h
+++ b/AK/LsanSuppressions.h
@@ -16,7 +16,8 @@ char const* __lsan_default_suppressions()
     // Both Skia and Chromium suppress false positive FontConfig leaks
     // https://github.com/google/skia/blob/main/tools/LsanSuppressions.cpp#L20
     // https://chromium.googlesource.com/chromium/src/build/+/master/sanitizers/lsan_suppressions.cc#25
-    return "leak:FcPatternObjectInsertElt";
+    return "leak:FcPatternObjectInsertElt\n"
+           "leak:rust_compile_regex";
 }
 }
 #endif


Libraries/LibWeb/HTML/Scripting/Fetching.cpp
```diff
@@ -407,8 +407,17 @@ static void compile_off_thread(NonnullRefPtr<JS::SourceCode const> source_code,
         }
 
         auto origin = event_loop_weak->take();
-        if (!origin)
+        if (!origin) {
```
awesomekling: Have you confirmed that this is actually what's happening? The only scenario where the EventLoop goes away *should* be on program exit, at which point leaks no longer matter anyway.

sideshowbarker: > Have you confirmed that this is actually what's happening?

I hadn’t before I pushed, no. And the CI run showed it still getting logged. So I poked around more to figure out what was actually going on try to find a way code-wise to make LSan stop seeing it — but couldn‘t.

> The only scenario where the EventLoop goes away _should_ be on program exit, at which point leaks no longer matter anyway.

Right. And given that, I’ve switched this PR over to just being an update to `AK/LsanSuppressions.h`.

wertzuz:
Shouldn't there be at least an comment/fixme why we skip it? How likely will this be fixed in the future?

sideshowbarker:
> Shouldn't there be at least an comment/fixme why we skip it?

The one other existing item in the file didn’t have a comment, so I just followed that precedent. And we have the explanation in the git history for the file. And we’ve touched the file exactly twice, and it’s not a likely to be getting updated very often at all. So it’ll never be hard to find out why this was added.

> How likely will this be fixed in the future?

Not sure what you mean by “fixed”. If you mean will we be changing the code behavior that leads to LSan reporting a false positive, I think it‘s unlikely it will get changed any time soon. Not intentionally at least. But I dunno some refactoring could end up making it go away I guess.

PR #9794, opened 2026-05-31
Meta: Pin actions to Node 20 compatible versions

alimpfard:
cc @gmta.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,6 +9,13 @@ updates:
       interval: "weekly"
     commit-message:
       prefix: "CI:"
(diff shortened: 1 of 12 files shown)


gmta:
Shouldn't we fix our CI image instead?

alimpfard:
That would make a lot more sense, buuuuut I suspect that'll take a while and I have no idea how to make it work 😅

gmta:
Alright, I promise to put in the work, but on holiday so please stand by 😅

alimpfard:
lol I didn't mean to poke you, will let you handle the thing then.

PR #9798, opened 2026-06-01
UI/Qt: Improve macOS chrome and content presentation

awesomekling:
This improves a few Qt UI behaviors around macOS browser windows and web content presentation.

Browser windows on macOS can now use rounded corners while custom chrome is active, with the setting hidden on other platforms. The web content view now uses `QRhiWidget` with Metal on macOS so IOSurface-backed frames can be presented through the GPU path instead of wrapping them in a `QImage`.

This also adds native pinch gesture handling for Qt, caches the fallback tab icon used by tabs without favicons, and makes a plain omnibox click select the serialized URL after mouse release.

### DIFF
--- a/UI/Qt/CMakeLists.txt
+++ b/UI/Qt/CMakeLists.txt
@@ -3,6 +3,13 @@ set(CMAKE_AUTORCC ON)
 set(CMAKE_AUTOUIC ON)
 find_package(Qt6 REQUIRED COMPONENTS Core Widgets)
 
+if (APPLE)
+    # The macOS QRhiWidget presentation path uses QRhi's private Metal native
+    # handle types to import IOSurfaces without CPU readback.
+    set(QT_NO_PRIVATE_MODULE_WARNING ON)
+    find_package(Qt6 REQUIRED COMPONENTS GuiPrivate)
+endif()
+
 qt_add_executable(ladybird main.cpp)
 target_sources(ladybird PRIVATE
     Application.cpp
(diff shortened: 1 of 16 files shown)


UI/Qt/CMakeLists.txt
```diff
@@ -3,6 +3,10 @@ set(CMAKE_AUTORCC ON)
 set(CMAKE_AUTOUIC ON)
 find_package(Qt6 REQUIRED COMPONENTS Core Widgets)
 
+if (APPLE)
+    find_package(Qt6 REQUIRED COMPONENTS GuiPrivate)
```
trflynn89, line 10: One thing with this is that we now get this on the build command line:
```
CMake Warning at /opt/homebrew/lib/cmake/Qt6/QtPublicDependencyHelpers.cmake:339 (message):
  This project is using headers of the GuiPrivate module and will therefore
  be tied to this specific Qt module build version.  Running this project
  against other versions of the Qt modules may crash at any arbitrary point.
  This is not a bug, but a result of using Qt internals.  You have been
  warned!

  You can disable this warning by setting QT_NO_PRIVATE_MODULE_WARNING to ON.
Call Stack (most recent call first):
  /opt/homebrew/lib/cmake/Qt6GuiPrivate/Qt6GuiPrivateConfig.cmake:48 (_qt_internal_show_private_module_warning)
  Build/vcpkg/scripts/buildsystems/vcpkg.cmake:908 (_find_package)
  /opt/homebrew/lib/cmake/Qt6/Qt6Config.cmake:253 (find_package)
  Build/vcpkg/scripts/buildsystems/vcpkg.cmake:908 (_find_package)
  UI/Qt/CMakeLists.txt:7 (find_package)
```

If the risk pointed out here is acceptable, we should probably set that var? `QT_NO_PRIVATE_MODULE_WARNING`

awesomekling: I believe the risk is acceptable for the time being, as this greatly improves our performance on macOS. If it becomes an issue, we can look at more invasive approaches that dodge the QRhiWidget compat commitment from Qt. :)

PR #9803, opened 2026-06-01
LibWeb: Invalidate layout tree of parent when replacing `CharacterData`

InvalidUsernameException:
When replacing the data in `CharacterData`, we were only invalidating text-dependent cache and layout of the layout node(s) corresponding to that `CharacterData`. This is not sufficient: With `::first-letter` styling, `TextSliceNode`s are being created for the first letter and the remainder of the text. So, contrary to regular `TextNode`s, these store offsets into the text data determined while building the layout tree. Subsequently these become stale when only invalidating layout, but not the layout tree. Using such stale offsets with updated text data leads to crashes in `TextNode::compute_text_for_rendering()` if the text got shorter and incorrect behavior if it got longer.

To fix this, invalidate the layout tree of the parent of the affected `TextSliceNode`s, causing a rebuild that creates new slice nodes with the updated data.

This fixes a crash when typing in the search box on https://search.brave.com/ where a first-letter style is used to capitalize the first letter of the search suggestions descriptions. This crash was a regression from b67d73a661fb995d567ecf9d2c090a995cb2deda (#8885).

/cc @tcl3

### DIFF
--- a/Libraries/LibWeb/DOM/CharacterData.cpp
+++ b/Libraries/LibWeb/DOM/CharacterData.cpp
@@ -140,17 +140,21 @@ WebIDL::ExceptionOr<void> CharacterData::replace_data(size_t offset, size_t coun
         return {};
 
     // NB: Called during DOM text mutation, layout is stale.
-    if (auto* text = as_if<Text>(*this)) {
-        Layout::TextOffsetMapping mapping { *text };
-        mapping.for_each_fragment([](Layout::TextNode& slice) {
+    if (is<Text>(*this)) {
+        if (is<Layout::TextSliceNode>(unsafe_layout_node())) {
+            // NB: Slice nodes cache data that is calculated at layout tree construction time.
+            // So for them, we need to invalidate the layout tree, not just layout.
+            if (parent())
+                parent()->set_needs_layout_tree_update(true, SetNeedsLayoutTreeUpdateReason::CharacterDataReplaceData);
+        } else if (auto* text_layout_node = as_if<Layout::TextNode>(unsafe_layout_node())) {
             // NB: Since the text node's data has changed, we need to invalidate the text for rendering.
             //     This ensures that the new text is reflected in layout, even if we don't end up doing a full layout
             //     tree rebuild.
-            slice.invalidate_text_for_rendering();
+            text_layout_node->invalidate_text_for_rendering();
 
             // We also need to relayout.
-            slice.set_needs_layout_update(SetNeedsLayoutReason::CharacterDataReplaceData);
-        });
+            text_layout_node->set_needs_layout_update(SetNeedsLayoutReason::CharacterDataReplaceData);
+        }
     }
 
     document().bump_character_data_version();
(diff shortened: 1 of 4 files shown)


Tests/LibWeb/Ref/input/modify-nodeValue-of-first-letter-styled-text.html
```diff
@@ -0,0 +1,20 @@
+<!DOCTYPE html>
```
tcl3, line 1: filename typo: fist -> first.

Libraries/LibWeb/DOM/CharacterData.cpp
```diff
@@ -139,19 +139,8 @@ WebIDL::ExceptionOr<void> CharacterData::replace_data(size_t offset, size_t coun
...
-            // NB: Since the text node's data has changed, we need to invalidate the text for rendering.
-            //     This ensures that the new text is reflected in layout, even if we don't end up doing a full layout
-            //     tree rebuild.
-            slice.invalidate_text_for_rendering();
-
-            // We also need to relayout.
-            slice.set_needs_layout_update(SetNeedsLayoutReason::CharacterDataReplaceData);
-        });
-    }
+    if (auto* text = as_if<Text>(*this); text && text->parent())
```
tcl3: Could we invalidate the layout tree for only TextSliceNodes? This should limit any worst case performance regression.

Something like:
```c++
if (auto* text = as_if<Text>(*this)) {
    if (as_if<Layout::TextSliceNode>(text->unsafe_layout_node())) {
        if (auto* parent = text->parent())
            parent->set_needs_layout_tree_update(true, SetNeedsLayoutTreeUpdateReason::CharacterDataReplaceData);
    } else if (auto* text_layout_node = as_if<Layout::TextNode>(text->unsafe_layout_node())) {
        // NB: Since the text node's data has changed, we need to invalidate the text for rendering.
        //     This ensures that the new text is reflected in layout, even if we don't end up doing a full layout
        //     tree rebuild.
        text_layout_node->invalidate_text_for_rendering();

        // We also need to relayout.
        text_layout_node->set_needs_layout_update(SetNeedsLayoutReason::CharacterDataReplaceData);
    }
}
```

InvalidUsernameException: Yeah, that makes sense. That gets my invalidation counters for `test-web` down from the original 23997 to just 2. For real-world websites, the only invalidations that trigger this tree rebuild now are the ones on the brave search suggestions (and those are leaf nodes even). So, this completely eliminates the performance concern.

Libraries/LibWeb/DOM/CharacterData.cpp
```diff
@@ -16,6 +16,7 @@
 #include <LibWeb/DOM/Text.h>
 #include <LibWeb/Layout/TextNode.h>
 #include <LibWeb/Layout/TextOffsetMapping.h>
+#include <LibWeb/TraversalDecision.h>
```
tcl3: I don't think this is used anywhere.

InvalidUsernameException: That must have slipped through from debug output.

tcl3 (approved):
Thanks for catching this!

PR #9809, opened 2026-06-01
LibDNS: Fix crash on a resource record with trailing rdata bytes

kubrickfr:
A DNS response answer record whose `RDLENGTH` is larger than the type-specific parser consumes crashes the process.

## Root cause

`ResourceRecord::from_raw()` parses each record's `RDATA` through a `CountingStream` seeded with the record's absolute offset in the packet, so that any compression pointers inside the rdata resolve against the whole message. As a result `rdata_stream.read_bytes()` returns a *packet-relative* offset.

When the type-specific parser leaves bytes unconsumed, the `PARSE_AS_RR` macro logs a diagnostic that uses that packet-relative offset to index into the *rdata-only* buffer:

```cpp
dbgln("Extra data ({}) left in stream: {:hex-dump}",
    rdata.size() - rdata_stream.read_bytes(),          // size_t underflow
    rdata.bytes().slice(rdata_stream.read_bytes()));   // slice start > size()
```

`read_bytes()` is almost always larger than `rdata.size()`, so the subtraction underflows and `Span::slice()` is called with `start > size()`, tripping a `VERIFY` and aborting the process. (On a build without assertions the same path is an out-of-bounds read.)

`RequestServer` parses DNS responses coming from the upstream resolver, so a hostile or on-path resolver can crash it — for example with an `A` record carrying `RDLENGTH=5` (the parser only reads 4 bytes for the address). That takes down networking for the rest of the session.

## Fix

Subtract the record's starting offset to obtain the rdata-relative number of consumed bytes before indexing into the buffer.

## Reproducer

Added `Tests/LibDNS/TestDNSMessage.cpp`, which feeds a 28-byte response containing an `A` record with `RDLENGTH=5` to `Message::from_raw()`. Before this change the test aborts; after it, parsing fails gracefully with an error.

### DIFF
--- a/Libraries/LibDNS/Message.cpp
+++ b/Libraries/LibDNS/Message.cpp
@@ -831,14 +831,15 @@ ErrorOr<ResourceRecord> ResourceRecord::from_raw(ParseContext& ctx)
     ParseContext rdata_ctx { rdata_stream, move(ctx.pointers) };
     ScopeGuard guard([&] { ctx.pointers = move(rdata_ctx.pointers); });
 
-#define PARSE_AS_RR(TYPE)                                                                                                                                   \
-    do {                                                                                                                                                    \
-        auto rr = TRY(Records::TYPE::from_raw(rdata_ctx));                                                                                                  \
-        if (!rdata_stream.is_eof()) {                                                                                                                       \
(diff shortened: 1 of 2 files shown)


gmta:

PR #9811, opened 2026-06-01
LibDNS: Reject domain name labels longer than 63 octets

kubrickfr:
A DNS label length octet is a 6-bit value, so a label can hold at most 63 octets. Several code paths violated this and could abort the process.

## Root cause

- `DomainName::from_raw()` treats any length octet that is not a compression pointer (top two bits `0b11`) as an ordinary label. The reserved `0b01`/`0b10` encodings (RFC 1035 §4.1.4) were accepted as labels of up to 191 octets.
- `DomainName::to_raw()` then `VERIFY`-aborts on any label longer than 63 octets when encoding it back to the wire.
- `Resolver::lookup()` encodes the outgoing query with `MUST(query.to_raw(...))`, so even once `to_raw()` returns an error the `MUST` would still abort.

Together these crash `RequestServer`, which drives DNS for the whole browser:

- **Outgoing (web-reachable):** when a custom resolver is configured, a page that references a host whose name contains a label longer than 63 characters aborts RequestServer while the query is being encoded.
- **Incoming:** a response carrying a reserved-length label is accepted by the parser and then aborts when the name is re-encoded (e.g. during DNSSEC validation).

## Fix

Reject over-long labels while parsing, return an error instead of aborting while encoding, and reject the lookup promise instead of `MUST`-aborting at the query call site.

## Reproducers

**Browser** — with a custom resolver configured, loading a host with an over-long label aborts RequestServer before this change:

```
Ladybird --headless=text --dns-server 127.0.0.1 \
    http://aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.example.com/
```

Before: `VERIFICATION FAILED: label.length() <= 63 at Libraries/LibDNS/Message.cpp:694`, reached via `RequestServer::Request::fetch → Message::to_raw`. After: the lookup fails gracefully ("Domain name label exceeds 63 octets") and RequestServer stays up.

**Unit tests** — `Tests/LibDNS/TestDNSMessage.cpp` covers both directions: encoding a name built from a 64-octet label (the outgoing path, which aborted before the fix), and parsing a message whose question name begins with a reserved `0x40` length octet (the incoming path, silently accepted before the fix).

> Note: this PR adds `Tests/LibDNS/TestDNSMessage.cpp`, the same new file added by #9809. Whichever merges second needs a trivial rebase to keep both test cases.

### DIFF
--- a/Libraries/LibDNS/Message.cpp
+++ b/Libraries/LibDNS/Message.cpp
@@ -678,6 +678,11 @@ ErrorOr<DomainName> DomainName::from_raw(ParseContext& ctx)
             return Error::from_string_literal("Invalid domain name pointer in label");
         }
 
+        // A normal label is at most 63 octets. Lengths whose top two bits are 0b01 or 0b10 are reserved (RFC 1035,
+        // 4.1.4) and would produce a label that can no longer be re-encoded, so reject them here.
+        if (length > 63)
+            return Error::from_string_literal("Domain name label exceeds 63 octets");
+
         ByteBuffer content;
         TRY(ctx.stream.read_until_filled(TRY(content.get_bytes_for_writing(length))));
         name.labels.append(ByteString::copy(content));
@@ -691,7 +696,8 @@ ErrorOr<DomainName> DomainName::from_raw(ParseContext& ctx)
 ErrorOr<void> DomainName::to_raw(ByteBuffer& out) const
 {
     for (auto& label : labels) {
-        VERIFY(label.length() <= 63);
+        if (label.length() > 63)
+            return Error::from_string_literal("Domain name label exceeds 63 octets");
         auto size_bytes = TRY(out.get_bytes_for_writing(1));
         u8 size = static_cast<u8>(label.length());
         memcpy(size_bytes.data(), &size, 1);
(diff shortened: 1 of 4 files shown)


Libraries/LibDNS/Message.cpp
```diff
@@ -678,6 +678,12 @@ ErrorOr<DomainName> DomainName::from_raw(ParseContext& ctx)
             return Error::from_string_literal("Invalid domain name pointer in label");
         }
 
+        // A normal label is at most 63 octets. Lengths whose top two bits are
+        // 0b01 or 0b10 are reserved (RFC 1035, 4.1.4) and would produce a label
+        // that can no longer be re-encoded, so reject them here.
```
gmta: Please (re)wrap all comments at column 120 (here and in the new tests).

gmta (commented):
Nice!

Libraries/LibDNS/Message.cpp
kubrickfr: Done — rewrapped the new comments (here and in the test) at column 120.

gmta (approved):

PR #9812, opened 2026-06-01
Tests: Merge tests into single binary per lib

R-Goc:
This merges all tests in libraries with more than one tests source into a single file. The exception to this are test runners like test-js, test-wasm, and test-web. TestLibCoreStream is also left out temporarily until flakiness with it on Windows is resolved. This significantly reduces the time needed to build tests without impacting organization. As all the test functions are static anyway, there are no conflicts. This would need modifying the ladybird_test_suite function for #9741. Along with PCH files some preliminary benchmarks show a reduction in the time needed to build all tests (going from ninja ladybird to ninja) from 1m 20s to 42s.

### DIFF
--- a/Meta/CMake/targets.cmake
+++ b/Meta/CMake/targets.cmake
@@ -177,6 +177,29 @@ function(ladybird_utility name)
     ladybird_generate_dsym(${name})
 endfunction()
 
+function(ladybird_test_suite name sub_dir)
+    cmake_parse_arguments(PARSE_ARGV 2 LADYBIRD_TEST_SUITE "" "" "SOURCES;LIBS")
+
+    add_executable(${name} ${LADYBIRD_TEST_SUITE_SOURCES})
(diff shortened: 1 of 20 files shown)


shannonbooth:
This seems like it could be quite annoying when running a single test for a targeted fix, I like only needing to depend on AK if changing something like optional rather than recompiling lots of things. Is this really a problem? I barely spend any time in my dev cycle linking these tests

R-Goc:
> This seems like it could be quite annoying when running a single test for a targeted fix, I like only needing to depend on AK if changing something like optional rather than recompiling lots of things. Is this really a problem? I barely spend any time in my dev cycle linking these tests

Not sure I understand the exact scenario you're talking about? I saw them taking quite a long while in my builds. That won't impact the rebuild time of single tests if that is what you mean.

shannonbooth:
I mean something like: `ninja -C Build/release TestVariant && ./Build/release/bin/TestVariant` after changing `AK/Variant.h`. Might not be a problem after these changes but it was definitely a first thought as I use those binaries to avoid a full LibWeb build as my main purpose

gmta:

PR #9813, opened 2026-06-01
LibWeb+LibWasm: JIT Fixes

alimpfard:


Libraries/LibWeb/WebAssembly/Table.cpp
```diff
@@ -29,19 +33,59 @@ static Wasm::ValueType table_kind_to_value_type(Bindings::TableKind kind)
...
+}
+
+static JS::Value table_limit_to_js_value(JS::VM& vm, u64 value, Wasm::AddressType address_type)
+{
+    if (address_type == Wasm::AddressType::I64)
+        return JS::BigInt::create(vm, ::Crypto::SignedBigInteger { ::Crypto::UnsignedBigInteger { value } });
+    return JS::Value { static_cast<u32>(value) };
+}
+
 WebIDL::ExceptionOr<GC::Ref<Table>> Table::construct_impl(JS::Realm& realm, Bindings::TableDescriptor& descriptor, Optional<JS::Value> value)
```
shannonbooth, line 74: I think it would be helpful if this had spec comments for this, I can't find where this is coming from. Is the new IDL just using "any" there?

alimpfard: The post-mem64 uses any (typedef'd), I'll pull the spec comments after I'm done fixing the other bug I found :)

shannonbooth: Oh unfortunate, I was hoping there was some bindings stuff perhaps to change if it was, I dont know, a variant of int and biting type of thing, but I suppose not!

alimpfard: AFAIK it's only expected to accept int and bigint, I'll take a looksie and make a spec issue if it can be a variant

shannonbooth: ah this might be why https://webidl.spec.whatwg.org/#limit-bigint-numeric-unions / https://github.com/whatwg/webidl/issues/1388

alimpfard: Hmm, the bigint gets converted to i64 and the int to i32 though

shannonbooth: Found it! https://github.com/webassembly/memory64/issues/85

Libraries/LibWeb/WebAssembly/Table.cpp
```diff
@@ -29,31 +33,88 @@ static Wasm::ValueType table_kind_to_value_type(Bindings::TableKind kind)
...
+
+        auto string = TRY_OR_THROW_OOM(vm, bigint->big_integer().to_base(10));
+        auto number = string.to_number<u64>();
+        if (!number.has_value())
+            return vm.throw_completion<JS::TypeError>("Table size is too large"sv);
+        return *number;
+    }
+
+    auto number_value = TRY(value.to_number(vm));
+    if (!number_value.is_integral_number() || number_value.as_double() < 0)
```
shannonbooth: If I am reading this correctly, I don't believe this is equivalent to `WebIDL::convert_to_int<WebIDL::Long>` with EnforceRange set to true - this will throw instead of truncating a double. Also I'm not sure how I missed it before, spec link is https://webassembly.github.io/spec/js-api/#addressvaluetou64 (could we link there and https://webassembly.github.io/spec/js-api/#dom-table-table)

PR #9814, opened 2026-06-01
LibWeb: Ensure that shadow roots have hosts before appending children

Calme1709:
DOM nodes inherit their parent's connectedness at time of insertion so we need to make sure the shadow root has a host before we start appending children.

Before this change the shadow tree elements would be marked as disconnected despite being in the DOM tree, causing issues such as relevant layout invalidations not taking place.

Fixes a crash when submitting a prompt on https://www.chatgpt.com

### DIFF
--- a/Libraries/LibWeb/DOM/Element.cpp
+++ b/Libraries/LibWeb/DOM/Element.cpp
@@ -1463,6 +1463,10 @@ void Element::set_shadow_root(GC::Ptr<ShadowRoot> shadow_root)
     }
     m_shadow_root = move(shadow_root);
     if (m_shadow_root) {
+        // NB: Children shouldn't be set on shadow roots until the shadow root is attached to a host so that they
+        //     correctly inherit connectedness.
+        VERIFY(!m_shadow_root->has_children());
+
         m_shadow_root->set_host(this);
         m_shadow_root->set_is_connected(is_connected());
     }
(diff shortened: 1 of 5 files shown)


sideshowbarker (approved):
I went through all ~13 `set_shadow_root` callers and confirmed the three here were the only offenders.

One non-blocking suggestion: The test covers only the file input — but color and `<details>` are affected and could also be asserted.

Suggested refinement of the commit message:

```text
LibWeb: Ensure that shadow roots have hosts before appending children

This change gives each shadow root its host before appending its
children. So they inherit the correct connectedness, and assert that a
root has no children when it's assigned a host — so the other
shadow-tree builders preserve that ordering.

Otherwise, without this change, the user-agent shadow trees for the
details element, color inputs, and file inputs append their children
before the shadow root is given a host — so those children stay marked
disconnected while present in the rendered tree. Disconnected nodes then
skip layout invalidations they should trigger.

Fixes a crash when submitting a prompt on https://www.chatgpt.com
```

Calme1709:
I have updated the test to include `<input type="color">` but this bug is limited to rebuilds of the shadow tree (otherwise insertion will update connectedness correctly) which isn't relevant to `HTMLDetailsElement`.

I have updated the commit message in line with your suggestion and also mentioned the above detail around rebuilds.

PR #9817, opened 2026-06-01
LibWebView: Use named mutex for single-instance detection on Windows

heitke:
LibWebView: Use named mutex for single-instance detection on Windows
Replace the PID file approach for detecting an existing browser process
on Windows with a named mutex.

A named mutex is the "Windows way" of doing this. It is automatically
released by the OS when the process exits, even if it crashes,
making stale lock detection unnecessary.

`get_process_pid` and its associated PID file logic are now excluded
from Windows builds entirely.


gmta:

PR #9820, opened 2026-06-02
LibWeb+LibMedia: Order video resize before the ‘ended’ & ‘seeked’ events

sideshowbarker:
**Problem:** The `HTMLVideoElement-resize-event-during-playback` test was intermittently flaky under CI.

**Cause:** When the real-time media timeline runs ahead of frame delivery — the decoder hasn’t yet delivered the frame whose dimensions a `seeked` or `ended` event should reflect — the event unexpectedly got enqueued before the `resize` event for that frame. For the `ended` event, that also unexpectedly made the `resize` event get dropped outright — since the test seeks away before a later render tick can present it.

**Fix:** Defer each `seeked` and `ended` event until the relevant frame has been presented — so its `resize` event is enqueued first. A seek that will change the displayed frame defers `seeked` until the frame is presented; reaching the end of playback defers the end-of-playback steps until the final frame is presented. Each wait is bounded: by the seek completing, and by the pipeline reaching end-of-stream, respectively. So neither can hang — and looping playback is never deferred.

### DIFF
--- a/Libraries/LibMedia/Sinks/DisplayingVideoSink.cpp
+++ b/Libraries/LibMedia/Sinks/DisplayingVideoSink.cpp
@@ -163,4 +163,17 @@ RefPtr<VideoFrame> DisplayingVideoSink::current_frame()
     return m_current_frame;
 }
 
+bool DisplayingVideoSink::current_frame_covers(AK::Duration timestamp) const
+{
+    if (m_current_frame == nullptr)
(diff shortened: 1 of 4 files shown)


sideshowbarker:
Closing in favor of the correct fix in https://github.com/LadybirdBrowser/ladybird/pull/9821

PR #9824, opened 2026-06-02
LibGfx+LibWeb: Don't rasterize an oversized SVG pattern tile

kubrickfr:
A web page can abort the WebContent process with an SVG `<pattern>` whose tile is too large to rasterize.

## Root cause

`SVGPatternElement::to_gfx_paint_style()` only rejects non-positive tile sizes. A pattern with a very large `width`/`height` — e.g. a `patternUnits="userSpaceOnUse"` pattern of `100000x100000` — produces a tile rect whose device-pixel size exceeds what `Gfx::Bitmap` can represent (`size_would_overflow`: each dimension must be ≤ 65535 and `pitch * height` must fit in an `i32`).

The pattern is rasterized into a bitmap-backed `PaintingSurface`, and `PaintingSurface::create_with_size()` unwraps `Bitmap::create(...)` with `.value()`:

```cpp
auto bitmap = Bitmap::create(color_type, alpha_type, size).value();  // aborts on error
```

So when the pattern is painted, the oversized tile makes `Bitmap::create` return an error and the `.value()` aborts the process.

## Fix

Reject tile sizes the bitmap backend cannot represent in `to_gfx_paint_style()`, matching `Gfx::Bitmap`'s own limits, so the pattern is simply not painted instead of crashing.

## Reproducer

`Tests/LibWeb/Crash/SVG/pattern-with-huge-tile-size.html` draws a `100000x100000` `userSpaceOnUse` pattern into a canvas via `drawImage()` to force rasterization. Before this change it aborts WebContent when the pattern is painted; after it, the page renders without the pattern and stays up.

(Rasterization has to be forced — for example via `drawImage` — because the crash happens during painting, not layout.)

### DIFF
--- a/Libraries/LibGfx/Bitmap.cpp
+++ b/Libraries/LibGfx/Bitmap.cpp
@@ -60,7 +60,7 @@ size_t Bitmap::minimum_pitch(size_t width, BitmapFormat format)
     return width * element_size;
 }
 
-static bool size_would_overflow(BitmapFormat format, IntSize size)
+bool Bitmap::size_would_overflow(BitmapFormat format, IntSize size)
 {
     if (size.width() < 0 || size.height() < 0)
         return true;
--- a/Libraries/LibGfx/Bitmap.h
+++ b/Libraries/LibGfx/Bitmap.h
@@ -95,6 +95,7 @@ public:
     [[nodiscard]] size_t pitch() const { return m_pitch; }
 
     [[nodiscard]] static size_t minimum_pitch(size_t width, BitmapFormat);
+    [[nodiscard]] static bool size_would_overflow(BitmapFormat, IntSize);
 
     [[nodiscard]] bool has_alpha_channel() const { return m_format == BitmapFormat::BGRA8888 || m_format == BitmapFormat::RGBA8888; }
     [[nodiscard]] BitmapFormat format() const { return m_format; }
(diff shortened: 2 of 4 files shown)


Libraries/LibWeb/SVG/SVGPatternElement.cpp
```diff
@@ -282,6 +282,13 @@ Optional<Painting::PaintStyle> SVGPatternElement::to_gfx_paint_style(SVGPaintCon
     if (tile_rect.is_empty())
         return {};
 
+    // The tile is rasterised into a bitmap-backed surface. Reject sizes the bitmap backend cannot
+    // represent (each dimension must fit in a u16, and pitch * height must fit in an i32) rather
+    // than aborting when the surface is created.
+    if (tile_rect.width() > NumericLimits<u16>::max() || tile_rect.height() > NumericLimits<u16>::max()
+        || static_cast<double>(tile_rect.width()) * tile_rect.height() * 4 >= NumericLimits<i32>::max())
+        return {};
```
gmta: In `LibGfx/Bitmap.cpp`, we have a static method `size_would_overflow` that implements this logic as well. We should expose that method so you can reuse it here. Bitmap constraints should be registered in a single place.

kubrickfr: Good call — exposed `Gfx::Bitmap::size_would_overflow()` and moved the check to where the pattern tile surface is actually created (`DisplayListPlayerSkia`), which knows the exact tile size and pixel format and covers any pattern source, not just SVG. Dropped the duplicated logic from `SVGPatternElement`.

gmta:

PR #9815, opened 2026-06-01
LibWeb: Ignore stale media playback callbacks after resource changes

jeetsh4h:
Keep a media resource generation independent from fetch generation. PlaybackManager callbacks capture the resource they were installed for and ignore stale tracks, metadata, duration changes, state changes, and failures after the element starts loading another resource.

This keeps callbacks from old managers from reading or mutating the replacement manager.

PR #9828, opened 2026-06-02
LibWeb: Immediately clear `PlaybackManager` on remote resource failure

Calme1709:
Previously this was done as part of a queued media element task in `failure_callback` but that could cause issues as the background thread creating a demuxer from the associated stream could also raise a failure in the interim, by clearing the `PlaybackManager` immediately we suppress the additional failure.
    
Fixes a crash on https://chatgpt.com

CC @Zaggy1024 - I believe this is the same issue as was raised in #9815

### DIFF
--- a/Libraries/LibWeb/HTML/HTMLMediaElement.cpp
+++ b/Libraries/LibWeb/HTML/HTMLMediaElement.cpp
@@ -1806,12 +1806,16 @@ void HTMLMediaElement::set_up_playback_manager_for_remote()
 
     // -> If the media data can be fetched but is found by inspection to be in an unsupported format, or can otherwise not be rendered at all
     m_playback_manager->on_unsupported_format_error = GC::weak_callback(*this, [](auto& self, Media::DecoderError&& error) mutable {
+        auto const* playback_manager_ptr = self.m_playback_manager.ptr();
+
         // NB: Queue a task for this so that we don't destroy the PlaybackManager within one of its callbacks when we
         //     call forget_media_resource_specific_tracks().
-        self.queue_a_media_element_task([self = GC::Weak(self), error = move(error)] {
+        self.queue_a_media_element_task([self = GC::Weak(self), error = move(error), playback_manager_ptr = move(playback_manager_ptr)] {
             if (!self)
                 return;
             if (self->m_error)
+                return;
+            if (playback_manager_ptr != self->m_playback_manager.ptr())
                 return;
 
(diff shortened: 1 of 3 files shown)


Zaggy1024:
Ahh, this is helpful, it clarifies that the problematic path is from `<source>` elements. I'm not sure it's the fix I'd prefer, though, since normally clearing the playback manager ought to happen in `forget_media_specific_tracks()`. Not sure if there may be a better place for it, but that's where it felt most natural.

The reason this is a problem is specifically because `<source>` element processing doesn't set the element error, which is used as a precondition for the callbacks running. Perhaps instead of that condition, we could capture a raw pointer to the playback manager to check if it has been cleared or replaced since the task was queued? (We should do the same for all similar playback manager callbacks, probably.)

Zaggy1024:
(...might be interesting to compare the state of the element error at the end of this test to other browsers, they may actually disagree with spec...)

Calme1709:
Other browsers have `null` as the element error (which I believe matches the spec?).

I have updated the PR to guard against the playback manager being cleared for that task.

Libraries/LibWeb/HTML/HTMLMediaElement.cpp
```diff
@@ -1804,15 +1804,19 @@ void HTMLMediaElement::set_up_playback_manager_for_remote()
         self.on_metadata_parsed();
     });
 
+    auto weak_playback_manager = m_playback_manager->weak();
```
Zaggy1024: Using `weak()` is unnecessary here, that is only used to guard off-main-thread work internally now that PlaybackManager is non-ref-counted. Since it is only created/destroyed on the same thread our callback is running on here, it's not needed for safety against races. We can just capture a raw pointer and compare it to avoid atomic ref-counting the weak link.

(I should have moved it to private, whoops..)

Libraries/LibWeb/HTML/HTMLMediaElement.cpp
Zaggy1024, line 1842: We should probably guard the same way here. We could also include the m_error check here for symmetry, though it shouldn't actually hit that branch given that decoder errors only get hit if the format successfully read (after the fetch began).

Libraries/LibWeb/HTML/HTMLMediaElement.cpp
Calme1709: Done - I also added a commit to make that method private

Zaggy1024 (approved):

PR #9835, opened 2026-06-02
CI: Update the docker image version used in CI

trflynn89:


gmta:
Please grep for `ci_container` inside the `.github` directory - you're missing some

PR #9838, opened 2026-06-02
UI/Qt: Unify browser chrome layout

awesomekling:
Treat the toolbar, tab strip, and vertical sidebar as one continuous browser chrome surface. The toolbar now stays stable across tab layout modes, macOS gets left-side traffic-light controls, and Linux/Windows keep the existing right-side custom window controls.

This also retunes the chrome visuals around the new layout:
- Share the same base chrome color across toolbar, tab bar, and sidebar
- Use card-like selected tabs with quieter inactive and hover states
- Keep vertical tab hover expansion as an overlay instead of relayouting page content
- Align New Tab button geometry with tab geometry across horizontal and vertical modes
- Improve omnibox contrast, spacing, focus styling, and placeholder treatment

The branch also fixes follow-up interaction issues exposed by the chrome refactor, including autocomplete popup hit testing over native web content and page caret painting when focus moves from web content back to browser chrome.

### DIFF
--- a/Libraries/LibWeb/DOM/Document.cpp
+++ b/Libraries/LibWeb/DOM/Document.cpp
@@ -8319,6 +8319,12 @@ void Document::reset_cursor_blink_cycle()
         m_cursor_blink_timer->restart();
 }
 
+void Document::set_cursor_position_needs_repaint()
+{
+    if (auto position = cursor_position())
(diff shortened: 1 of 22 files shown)


trflynn89:
Note quite right with bookmarks :)

Screenshot 2026-06-02 at 10 30 54 PM

PR #9839, opened 2026-06-03
LibWeb: Route content blocker Rust allocations through mimalloc

tcl3:
This matches the pattern used by existing Rust components.

This fixes a crash when launching the Ladybird binary with the Distribution preset. 

Tested with: `BUILD_PRESET=Distribution ./Meta/ladybird.py run ladybird --headless=text example.com`

### DIFF
--- a/Libraries/LibWeb/CMakeLists.txt
+++ b/Libraries/LibWeb/CMakeLists.txt
@@ -1256,7 +1256,12 @@ ladybird_lib(LibWeb web EXPLICIT_SYMBOL_EXPORT)
 target_link_libraries(LibWeb PRIVATE LibCore LibCompress LibCrypto LibJS LibHTTP LibGfx LibIPC LibRegex LibSyntax LibTextCodec LibUnicode LibMedia LibWasm LibXML LibIDL LibURL LibTLS LibRequests LibGC LibSync LibThreading skia ${ANGLE_TARGETS} SDL3::SDL3 LibXml2::LibXml2)
 
 import_rust_crate(MANIFEST_PATH Rust/Cargo.toml CRATE_NAME libweb_rust FFI_HEADER RustFFI.h)
-import_rust_crate(MANIFEST_PATH ContentBlocker/Rust/Cargo.toml CRATE_NAME libweb_content_blocker_rust FFI_HEADER ContentBlockerRustFFI.h)
+
+set(content_blocker_rust_features "")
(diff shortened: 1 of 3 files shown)


Libraries/LibWeb/CMakeLists.txt
```diff
@@ -1256,7 +1256,12 @@ ladybird_lib(LibWeb web EXPLICIT_SYMBOL_EXPORT)
 target_link_libraries(LibWeb PRIVATE LibCore LibCompress LibCrypto LibJS LibHTTP LibGfx LibIPC LibRegex LibSyntax LibTextCodec LibUnicode LibMedia LibWasm LibXML LibIDL LibURL LibTLS LibRequests LibGC LibSync LibThreading skia ${ANGLE_TARGETS} SDL3::SDL3 LibXml2::LibXml2)
 
 import_rust_crate(MANIFEST_PATH Rust/Cargo.toml CRATE_NAME libweb_rust FFI_HEADER RustFFI.h)
-import_rust_crate(MANIFEST_PATH ContentBlocker/Rust/Cargo.toml CRATE_NAME libweb_content_blocker_rust FFI_HEADER ContentBlockerRustFFI.h)
+
+set(content_blocker_rust_features "")
+if (NOT BUILD_SHARED_LIBS)
+    set(content_blocker_rust_features FEATURES allocator)
+endif()
+import_rust_crate(MANIFEST_PATH ContentBlocker/Rust/Cargo.toml CRATE_NAME libweb_content_blocker_rust ${content_blocker_rust_features} FFI_HEADER ContentBlockerRustFFI.h)
```
trflynn89, line 1264: I'm a little confused why LibWeb is different than other libs here. In other libs, we have the allocator feature in all crates and enable it unconditionally in CMakeLists.txt. Should both the main LibWeb and ContentBlocker crates have the feature an enabled unconditionally as well?

tcl3: This prevents `liblagom-web.so` seeing duplicate allocator symbols when `BUILD_SHARED_LIBS=ON`. LibWeb is different to other libs here because it links 2 Rust `staticlib`s, `libweb_rust` and `libweb_content_blocker_rust`,  to a single CMake target.

trflynn89: Got it - I wonder if we should stick with a single rust crate per library (or somehow link the allocator differently), but this is fine for now

PR #9851, opened 2026-06-03
LibWeb: Ignore vertical-align on flex and grid items in box_baseline

### LINKED ISSUE #9840: incorrect positioning of elements in dndbeyond
When loading a character sheet in D&D Beyond, the ability information is incorrectly bunched up at the bottom.

Ladybird
Image
Expected behavior
Image

I also stripped away everything I could to create a reduced test case in html
```
<!DOCTYPE html>
<html>

<body>
  <div class="ct-character-sheet--dice-enabled ddbc-ability-summary">
    <div class="ddbc-ability-summary__heading"><span>Strength</span></div>
    <div class="ddbc-ability-summary__primary">
      <button type="button" class=" integrated-dice__container">
        <span class="styles_signed__scf97">3</span>
      </button>
    </div>
    <div class="ddbc-ability-summary__secondary">17
    </div>
  </div>
  <style>
    .styles_signed__scf97 {
      display: inline-flex;
      vertical-align: top
    }

    .ddbc-ability-summary {
      height: 95px;
      position: relative;
    }

    .ddbc-ability-summary__heading {
      padding-top: 8px;
    }


    .ddbc-ability-summary__primary {
      font-size: 26px;
      position: relative;
    }

    .ddbc-ability-summary__secondary {
      bottom: 4px;
      position: absolute;
    }

    .integrated-dice__container {
      display: inline-flex;
    }
  </style>
</body>

</html>
```
Ladybird
Image
Expected behavior
Image

sideshowbarker:
**Problem:** Borked layout on (flex-styled) “D&D Beyond” site.

**Cause:** `FormattingContext::box_baseline()` was applying the box’s `vertical-align` unconditionally. But `vertical-align` shouldn’t be consulted for flex items or grid items.

**Fix:** Skip `vertical-align` handling in `box_baseline()` when the box is a flex or grid item.

Fixes https://github.com/LadybirdBrowser/ladybird/issues/9840

### DIFF
--- a/Libraries/LibWeb/Layout/FormattingContext.cpp
+++ b/Libraries/LibWeb/Layout/FormattingContext.cpp
@@ -2555,7 +2555,7 @@ CSSPixels FormattingContext::box_baseline(Box const& box) const
 
     // https://drafts.csswg.org/css2/#propdef-vertical-align
     auto const& vertical_align = box.computed_values().vertical_align();
-    if (vertical_align.has<CSS::VerticalAlign>()) {
+    if (box.vertical_align_applies() && vertical_align.has<CSS::VerticalAlign>()) {
         switch (vertical_align.get<CSS::VerticalAlign>()) {
         case CSS::VerticalAlign::Top:
             // Top: Align the top of the aligned subtree with the top of the line box.
--- a/Libraries/LibWeb/Layout/Node.h
+++ b/Libraries/LibWeb/Layout/Node.h
@@ -161,6 +161,21 @@ public:
     bool is_grid_item() const { return m_is_grid_item; }
     void set_grid_item(bool b) { m_is_grid_item = b; }
 
+    bool vertical_align_applies() const
+    {
+        // https://drafts.csswg.org/css-flexbox/#flex-containers
+        // "vertical-align has no effect on a flex item"
+        if (is_flex_item())
+            return false;
+        // https://drafts.csswg.org/css-grid-1/#grid-container
+        // "vertical-align has no effect on a grid item"
+        if (is_grid_item())
+            return false;
+        // FIXME: Per-spec, vertical-align only applies to inline-level boxes and table cells; this should be narrowed
+        //        to that — rather than only excluding flex and grid items.
+        return true;
+    }
+
     [[nodiscard]] GC::Ptr<Box const> containing_block() const { return m_containing_block; }
     [[nodiscard]] GC::Ptr<Box> containing_block() { return m_containing_block; }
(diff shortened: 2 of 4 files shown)


Libraries/LibWeb/Layout/FormattingContext.cpp
```diff
@@ -2554,8 +2554,10 @@ CSSPixels FormattingContext::box_baseline(Box const& box) const
     auto const& box_state = m_state.get(box);
 
     // https://drafts.csswg.org/css2/#propdef-vertical-align
+    // NB: vertical-align has no effect on flex or grid items — so when this box is one, skip it and derive the baseline
+    //     from content below.
```
tcl3: It might be worth quoting from the flexbox and grid specs here:

From https://drafts.csswg.org/css-flexbox/#flex-containers:
> [vertical-align](https://drafts.csswg.org/css-inline-3/#propdef-vertical-align) has no effect on a flex item.

From https://drafts.csswg.org/css-grid-1/#grid-container:

> [vertical-align](https://www.w3.org/TR/css-inline-3/#propdef-vertical-align) has no effect on a grid item.

It might be worth moving this logic into a `vertical_align_applies()` helper or similar, since it should apply anywhere `vertical-align` is used.

sideshowbarker: > It might be worth quoting from the flexbox and grid specs here:

Yup — added

> It might be worth moving this logic into a `vertical_align_applies()` helper or similar, since it should apply anywhere `vertical-align` is used.

Yes — but did it as `is_flex_or_grid_item()` since that seemed to be the better fit.

tcl3: > Yes — but did it as `is_flex_or_grid_item()` since that seemed to be the better fit.

The main reason I suggested such a helper was that it's a good place to put the relevant spec comments. I imagine it would also be expanded in the future since vertical-align should only apply to inline-level boxes or table cells - although I suppose that's outside the scope of this PR.

sideshowbarker: > The main reason I suggested such a helper was that it's a good place to put the relevant spec comments. I imagine it would also be expanded in the future 

d’oh yeah I see now — sorry for having been daft. Made it so.

> since vertical-align should only apply to inline-level boxes or table cells - although I suppose that's outside the scope of this PR

Yeah, but added a FIXME so that we don’t forget about it

tcl3 (approved):

PR #9852, opened 2026-06-03
LibWeb: Let style sheets actually block scripts

AtkinsSJ:
We had `HTMLStyleElement::contributes_a_script_blocking_style_sheet()` before, but it always returned false so scripts wouldn't wait for style sheets to be loaded. Oops.

This PR implements the majority of it, while also moving it into StyleElementBase so that SVGStyleElement also gets the same behaviour.

There's an additional small fix that any `@import supports(...)` that doesn't match is now treated as loaded instead of being stuck in the loading state forever.


AtkinsSJ:
Tested this with `/home/sam/ladybird/Build/sanitizers/bin/test-web -f images-load-relative-to-imported-style-sheet --repeat 200 -j 1` and no failures, so I'm pretty confident it fixes the previous issue.

PR #9861, opened 2026-06-04
Move libraries and helper services away from exit-time cleanup

awesomekling:
This moves the codebase toward a model where libraries and helper services can assume that no exit-time cleanup takes place.

Enable `-Wexit-time-destructors` for in-tree libraries on macOS and non-Android Linux, and update process-lifetime statics so they no longer register exit-time destructors.

Add `Core::Process::terminate_immediately()` and use it in WebContent, WebWorker, ImageDecoder, and Compositor when their owning client disconnects. These helper processes can terminate directly instead of unwinding through global destructors during shutdown.

RequestServer is currently exempt because it still uses exit-time cleanup for HTTP disk cache handling. The warning is also not enabled for Android or Windows yet, since those platform-specific paths still have statics that need cleanup.


Libraries/LibCore/ProcessWindows.cpp
```diff
@@ -99,6 +100,12 @@ ErrorOr<Process> Process::spawn(StringView path, ReadonlySpan<StringView> argume
     });
 }
 
+void Process::terminate_immediately(int status)
+{
+    TerminateProcess(GetCurrentProcess(), static_cast<UINT>(status));
```
shannonbooth, line 107: _exit is also a thing on windows and is very similar, is terminate process more correct than _exit here with what we want?

awesomekling: My understanding is that `_exit` on windows skips some of the stuff but not all of it, such as DLL teardown. It's what Chromium uses for the same reason fwiw.

awesomekling: `TerminateProcess` is what they use, I mean :)

shannonbooth: Right, I was trying to search up the difference and was wondering if the DLL teardown stuff was meaningful. Stuff to do with DLL_PROCESS_DETACH apparently. Maybe it's worth leaving a comment? I can imagine forgetting whatever reason it is we want to do even more immediate teardown here

PR #9871, opened 2026-06-04
js: Replace LibLine with libedit

awesomekling:
Replace LibLine with libedit for the JavaScript REPL. The REPL keeps line editing, history, multiline input, and tab completion through libedit's readline-compatible interface, but drops LibLine's live syntax highlighting.

This also removes the now-unused LibLine library and its build wiring, drops the dead LibLine dependency from wasm, and removes nested `Core::EventLoop` support that LibLine was the last user of. Event loops are now a single per-thread instance instead of a stack.


alimpfard:
ew why? a significantly worse editing experience for dropping a library that has cost zero in maintenance burden?

edit: #9879 for your nested EventLoops

awesomekling:
> ew why? a significantly worse editing experience for dropping a library that has cost zero in maintenance burden?

4k+ lines of C++ for a bespoke line editing library used by a single testing utility is unreasonable. Makes perfect sense in SerenityOS, but here let’s use a stock component.

alimpfard:
I would agree if the existing thing was a good replacement, but it is not.
I'm not a fan of dropping the quality of the repl just to delete code.

teaalltr:
@awesomekling Just out of curiosity, have you evaluated linenoise? https://github.com/antirez/linenoise

PR #9874, opened 2026-06-04
UI/Qt: Collapse hover-expanded vertical tabs after leaving window

trflynn89:
Fixes a recent regression where collapsed vertical tabs would stay expanded on hover after the mouse left the browser window.


AtkinsSJ (approved):
Hehehe, this makes it 3 times now that I've made a mental note to look into something with the Qt UI and then I find you already have a patch for it!

PR #9880, opened 2026-06-04
LibCore: Keep main event loops alive for process lifetime

awesomekling:
Make process main loops use `Core::EventLoop::initialize_for_current_thread()` so their lifetime is independent of normal object teardown. Tighten `EventLoop::current()` and `EventLoop::pump()` checks so the current-loop invariant is enforced directly.

With those loops intentionally alive for the process lifetime, update browser frontends, helper services, utilities, LibMedia, LibWeb, LibDNS, WebDriver, and Compositor callbacks to keep direct `Core::EventLoop` references instead of routing through `WeakEventLoopReference`. This removes dead-loop checks from off-thread completion paths that always target the main loop.

### DIFF
--- a/Libraries/LibCore/EventLoop.cpp
+++ b/Libraries/LibCore/EventLoop.cpp
@@ -54,8 +54,7 @@ bool EventLoop::is_running()
 
 EventLoop& EventLoop::current()
 {
-    if (!current_event_loop())
-        dbgln("No EventLoop is present, unable to return current one!");
+    VERIFY(current_event_loop());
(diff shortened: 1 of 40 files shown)


Zaggy1024:
Very minor nitpick, but the pattern of capturing the main event loop as a pointer instead of a reference in a bunch of places feels a little odd to me. LibMedia changes make me happy though, very glad to see WeakEventLoopReference ~~gone~~ usage gone from there! This is what I would have preferred when I initially wrote it :^)

PR #9896, opened 2026-06-05
LibWeb+LibWebView: Remove some unused internal styling

Calme1709:
This PR removes `Native.css` (which is unused since the built-in inspector was removed) as well as unused `libweb-*` color keywords (i.e. all except for `libweb-buttonfacedisabled` and  `libweb-buttonfacehover`).

No functional changes


AtkinsSJ (approved):
Very nice! Can't believe it's taken this long to notice that CSS file wasn't needed. 😆

PR #9900, opened 2026-06-05
test-web: Scan for variant tests in advance instead of querying them when they're loaded

AtkinsSJ:
Instead of loading tests and then asking them if they have variants (which has been flaky) identify them during test collection by scanning their contents.

Without these changes, I was getting rare flakes on `Text/input/wpt-import/navigation-api/state/history-pushState.html` and `Text/input/wpt-import/navigation-api/state/history-replaceState.html` where they wouldn't get replaced with their variants. (At the rate of a run repeating them 1000 times would get a couple of failures.) With these changes, that no longer happens, tried with I think 5000 runs at this point.

I expected this to have some performance impact but actually I'm not seeing any. A test run on my macBook is about 60 seconds with or without these changes.

### DIFF
--- a/Tests/LibWeb/test-web/main.cpp
+++ b/Tests/LibWeb/test-web/main.cpp
@@ -18,6 +18,7 @@
 #include <AK/ByteBuffer.h>
 #include <AK/Enumerate.h>
 #include <AK/Function.h>
+#include <AK/GenericLexer.h>
 #include <AK/LexicalPath.h>
 #include <AK/NumberFormat.h>
 #include <AK/Platform.h>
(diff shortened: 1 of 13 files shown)


Tests/LibWeb/test-web/main.cpp
```diff
@@ -159,6 +160,153 @@ static ErrorOr<void> skip_async_scrolling_tests_unless_enabled(Application const
...
+            else if (attribute_name.equals_ignoring_ascii_case("content"sv))
+                content_attribute = attribute_value;
+        }
+
+        if (name_attribute.has_value() && name_attribute->equals_ignoring_ascii_case("variant"sv) && content_attribute.has_value())
+            variants.append(TRY(String::from_utf8(*content_attribute)));
+    }
+
+    return variants;
+}
```
gmta, line 308: :moar: HTML parsers!

Could we split this off to a Variants.cpp so we don't grow main.cpp unbounded?

Tests/LibWeb/test-web/main.cpp
```diff
@@ -159,6 +160,153 @@ static ErrorOr<void> skip_async_scrolling_tests_unless_enabled(Application const
...
+                attribute_value = consume_html_attribute_value(lexer);
+
+            if (attribute_name.equals_ignoring_ascii_case("name"sv))
+                name_attribute = attribute_value;
+            else if (attribute_name.equals_ignoring_ascii_case("content"sv))
+                content_attribute = attribute_value;
+        }
+
+        if (name_attribute.has_value() && name_attribute->equals_ignoring_ascii_case("variant"sv) && content_attribute.has_value())
+            variants.append(TRY(String::from_utf8(*content_attribute)));
```
sideshowbarker, line 304:
```suggestion
        if (… content_attribute.has_value()) {
            // Only query (?...) variants are expandable; skip anything else …
            if (content_attribute->starts_with('?'))
                variants.append(TRY(String::from_utf8(*content_attribute)));
            else
                warnln("test-web: ignoring unsupported variant content '{}' in {}", *content_attribute, test.input_path);
        }
```
…if, should there end up being any variants in WPT tests that we import with content that isn’t a `?` query, we want to not hard-fail here but instead warn.

sideshowbarker (commented):
Before I realized you’d already done this, I’d actually started work implementing it in nearly the same way. So yeah, I think this is the right way to do it — and I couldn’t find anything in it that needs changing. Feel free to not accept the commit suggestion — since it’s just for future-proofing slightly, and it’s not going to change current behavior anyway.

Big +1 to this, and I hope we can merge it soon. We are still getting CI failures daily that this will eliminate completely.

sideshowbarker (approved):

PR #9903, opened 2026-06-05
LibURL: Advance past the '@' delimiter after batch-parsing userinfo

tcl3:
After batch-processing the userinfo, the authority state advanced the parse pointer onto the '@' and relied on the end-of-loop increment to step past it. That increment uses the byte length of the authority's first code point, not the delimiter's. When the first code point was multi-byte, the pointer overshot into the middle of a code point and the next iteration sliced the string at a non-char boundary causing the process to panic. We now step past the delimiter explicitly so we don't need to rely on the end of loop increment to do so.

This can be tested by pasting "http://é@é" into the URL bar, which would previously panic.

### DIFF
--- a/Tests/LibURL/TestURL.cpp
+++ b/Tests/LibURL/TestURL.cpp
@@ -575,6 +575,25 @@ TEST_CASE(username_and_password)
     }
 }
 
+TEST_CASE(non_ascii_userinfo)
+{
+    {
+        auto url = URL::Parser::basic_parse("http://é@é"sv);
+        EXPECT(url.has_value());
+        EXPECT_EQ(url->username(), "%C3%A9"sv);
+        EXPECT(url->password().is_empty());
+        EXPECT_EQ(url->serialized_host(), "xn--9ca"sv);
+    }
+
+    {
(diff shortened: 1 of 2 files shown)


Tests/LibURL/TestURL.cpp
```diff
@@ -575,6 +575,25 @@ TEST_CASE(username_and_password)
     }
 }
 
+TEST_CASE(non_ascii_userinfo)
+{
+    {
+        auto url = URL::Parser::basic_parse("http://é@é"sv);
```
shannonbooth, line 581: Whoops! Looks good to me. Any chance we could upstream this to: https://github.com/web-platform-tests/wpt/blob/master/url/resources/urltestdata.json?

shannonbooth (approved):

Tests/LibURL/TestURL.cpp
tcl3: I've made a PR to the WPT repo: https://github.com/web-platform-tests/wpt/pull/60446

shannonbooth:
https://github.com/LadybirdBrowser/ladybird/pull/9956 updates the wpt import

PR #9920, opened 2026-06-05
LibWeb/CSS: Implement `contrast-color()` function

gmta:
Fixes the text color in the orange banner on tweakers.net:

| Before | After |
|--|--|
|  |  |


InvalidUsernameException:
Huh, funny timing. I had prepared a series of commits for this change as well. One thing I noticed is that the current implementation of `suggested_foreground_color()` can be pretty far of from what is considered a good choice. It might be worth implementing [WCAG 2 contrast ratio](https://w3c.github.io/wcag/guidelines/22/#dfn-contrast-ratio) properly instead of using what LibGfx currently has.

([my changes, for reference](https://github.com/InvalidUsernameException/ladybird/commits/contrast-color/))

gmta:
> One thing I noticed is that the current implementation of `suggested_foreground_color()` can be pretty far of from what is considered a good choice. It might be worth implementing [WCAG 2 contrast ratio](https://w3c.github.io/wcag/guidelines/22/#dfn-contrast-ratio) properly instead of using what LibGfx currently has.

Agreed, this should be a FIXME the very least. Will take a look.

Libraries/LibWeb/CSS/StyleValues/ContrastColorStyleValue.cpp
```diff
@@ -0,0 +1,58 @@
...
+{
+    auto const* other_contrast_color = as_if<ContrastColorStyleValue>(other);
+    if (!other_contrast_color)
+        return false;
+    return m_color == other_contrast_color->m_color;
+}
+
+void ContrastColorStyleValue::serialize(StringBuilder& builder, SerializationMode mode) const
+{
+    if (mode == SerializationMode::ResolvedValue) {
```
Calme1709: I don't believe this `ResolvedValue` specific serialization is required - we will only ever have a `ContrastColorStyleValue` at computed value time if absolutization failed (i.e. due to it containing `currentcolor`) in which case the next `if` statement will fall back to normal serialization any way

gmta: Agreed - removed it!

gmta:
@InvalidUsernameException implemented WCAG 2's contrast ratio and compared the changes in this PR against your branch. I've incorporated a couple of smaller things from your branch and marked you as the co-author.

PR #9921, opened 2026-06-05
LibWeb/Bindings: Don't include constants in default toJSON conversion

shannonbooth:
We should only be including regular attributes.


shannonbooth:
For reference, here is the relevant portion of the spec in the python port I'm working on (still trying to tidy it up since its very messy): https://github.com/shannonbooth/ladybird/blob/33e1b856814f851b3f0bbf0dd4b241641073d8c1/Meta/Generators/libweb_bindings/operations.py#L423

PR #9930, opened 2026-06-06
LibWeb: Fix Navigation API same-document history traversal

awesomekling:
This fixes same-document history traversals that fire Navigation API `navigate` events, including intercepted traversals that previously stopped after updating the URL and non-intercepted traversals that could abort their own event before `currententrychange` and `navigatesuccess`.

The patch queues the spec's resume step for intercepted traversals, skips that resume when the `NavigateEvent` has already been aborted or replaced, and preserves same-document traverse events until the Navigation API entry update completes.

Text coverage covers intercepted traversal, non-intercepted traversal, and an aborted intercepted traversal that is superseded before the queued step runs. The tests avoid timing assumptions by using deterministic Navigation API and history operations.


awesomekling:
This fixes an issue on Reddit when pressing the browser's back button to go from a comments section back to a subreddit index.

PR #9952, opened 2026-06-07
Meta: Add a script to test the flakiness of new and modified tests and wire it into CI

tcl3:
This script first identifies newly-added and newly-unskipped tests in the working tree vs. a given base ref. It then runs each of these tests individually using `test-web` at each of the given concurrency levels. There is also an option to specify a maximum runtime, used by CI so that PRs which add a lot of tests do not cause the step to run forever.

This doesn't identify modified tests. I could add this if desired.

### DIFF
--- a/.github/workflows/lagom-template.yml
+++ b/.github/workflows/lagom-template.yml
@@ -201,6 +201,20 @@ jobs:
             --load="module-native-protocol-unix auth-anonymous=1 socket=/tmp/pulse-native" \
             --load="module-null-sink sink_name=virtual_sink"
 
+      - name: Check new tests for flakiness
+        if: ${{ github.event_name == 'pull_request' && inputs.build_preset == 'Sanitizer' && (inputs.os_name == 'Linux' || inputs.os_name == 'macOS') }}
+        timeout-minutes: 30
+        working-directory: ${{ github.workspace }}
+        shell: bash
+        run: |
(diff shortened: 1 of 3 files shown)


gmta:
Hm, instead of dry running, wouldn't it be slightly simpler to call into `git show --name-only ...` for HEAD vs. the base revision, which would also give you modified tests?

tcl3:
> Hm, instead of dry running, wouldn't it be slightly simpler to call into `git show --name-only ...` for HEAD vs. the base revision, which would also give you modified tests?

I'd prefer to keep the dry-run mechanism as the single source of truth for what is and isn't a test - it gives us accurate counts for discovered tests and allows us to easily detect newly-unskipped tests.

I've added a simple mechanism to detect modified tests. It doesn't pick up a test if just it's expectation file is changed, but I don't think that's a very common case. I left modified tests out originally because there some cases where a lot of tests can be modified in one go, but I suppose the same is true for newly-added tests :shrug:

gmta:
> I'd prefer to keep the dry-run mechanism as the single source of truth for what is and isn't a test - it gives us accurate counts for discovered tests and allows us to easily detect newly-unskipped tests.

Fair!

> I've added a simple mechanism to detect modified tests. It doesn't pick up a test if just it's expectation file is changed, but I don't think that's a very common case.

Changed expectations should be manually inspected imo, I don't think we've had many flakes because of expectation rendering.

gmta (approved):

PR #9954, opened 2026-06-07
LibWeb: Fix use-after-free in live-collection filter captures

### LINKED ISSUE #9948: use-after-free related to HTMLFormElement::elements()
When loading this webpage on Linux:
```
<!doctype html>
<body></body>
<script>
if (!window.internals)
    throw new Error("This repro requires --expose-internals-object or --test-mode");

let form = document.createElement("form");
let input = document.createElement("input");
form.appendChild(input);
document.body.appendChild(form);

let collection = form.elements;
let weakForm = new WeakRef(form);
form.remove();
form = null;
input = null;

setTimeout(() => {
    internals.gc();
    console.log("weak form after first gc = " + weakForm.deref());

    for (let i = 0; i < 20000; ++i) {
        let div = document.createElement("div");
        div.id = "node-" + i;
        document.body.appendChild(div);
    }

    internals.gc();
    console.log("weak form after second gc = " + weakForm.deref());
    globalThis.result = collection.length;
    console.log("collection.length = " + globalThis.result);
    if (window.println)
        println(globalThis.result);
    internals.signalTestIsDone("PASS");
}, 0);
</script>
```
I got the log
```
155982.813 WebContent(618688): (js log) "weak form after first gc = undefined"
155982.889 WebContent(618688): (js log) "weak form after second gc = undefined"
155982.897 WebContent(618688): (js log) "collection.length = 0"
```
This means that the form is garbage collected. However, HTMLFormControlsCollection holds a pointer to HTMLFormElement so the form shouldn't be freed when form.elements is still live:
```
GC::Ref<HTMLFormControlsCollection> HTMLFormElement::elements() const
{
    if (!m_elements) {
        auto& root = as<ParentNode>(const_cast<HTMLFormElement*>(this)->root());
        m_elements = HTMLFormControlsCollection::create(root, DOM::HTMLCollection::Scope::Descendants, [this](Element const& element) {
            return is_form_control(element, *this);
        });
    }
    return *m_elements;
}
```

sideshowbarker:
**Problem:** Holding `form.elements` while the form is detached + dropped hit a use-after-free: the form is GC’ed while the collection’s still live.

**Cause:** `HTMLCollection` (and `LiveNodeList` too) was storing its filter as an `AK::Function` — which the garbage collector doesn’t visit. When a filter lambda captures a GC object (e.g. the form in `form.elements`) that object has no GC edge keeping it alive. So it can be collected while the collection using it’s still reachable — leaving a dangling pointer.

**Fix:** `HTMLCollection` and `LiveNodeList` are GC cells with their own `visit_edges`. So, visit the filter’s (and sort’s) capture range there: conservatively mark any GC object a captured lambda holds — to ensure it’s kept alive as long as the collection’s reachable. Fixes https://github.com/LadybirdBrowser/ladybird/issues/9948.

### DIFF
--- a/Libraries/LibWeb/DOM/HTMLCollection.cpp
+++ b/Libraries/LibWeb/DOM/HTMLCollection.cpp
@@ -50,6 +50,8 @@ void HTMLCollection::visit_edges(Cell::Visitor& visitor)
 {
     Base::visit_edges(visitor);
     visitor.visit(m_root);
+    visitor.visit_possible_values(m_filter.raw_capture_range());
+    visitor.visit_possible_values(m_sort.raw_capture_range());
 }
 
 GC::Cell const& HTMLCollection::owner_cell(Badge<GC::Heap>) const
--- a/Libraries/LibWeb/DOM/LiveNodeList.cpp
+++ b/Libraries/LibWeb/DOM/LiveNodeList.cpp
@@ -33,6 +33,7 @@ void LiveNodeList::visit_edges(Cell::Visitor& visitor)
 {
     Base::visit_edges(visitor);
     visitor.visit(m_root);
+    visitor.visit_possible_values(m_filter.raw_capture_range());
 }
 
 GC::RootVector<Node*> LiveNodeList::collection() const
--- a/Tests/LibWeb/Text/expected/form-elements-keeps-form-alive.txt
+++ b/Tests/LibWeb/Text/expected/form-elements-keeps-form-alive.txt
@@ -1,0 +1,2 @@
+form alive while collection is held: true
+collection.length: 0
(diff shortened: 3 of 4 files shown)


Libraries/LibWeb/DOM/HTMLCollection.cpp
```diff
@@ -50,6 +52,8 @@ void HTMLCollection::visit_edges(Cell::Visitor& visitor)
 {
     Base::visit_edges(visitor);
     visitor.visit(m_root);
+    visitor.visit(m_filter);
```
shannonbooth: This is an insidious type of bug that I wish the GC-plugin caught 😅 Its quite a tricky one though since I don't think the rules are the most straightforward.

However, since we're storing these in a GC::Cell already, I believe we can do:

`        visitor.visit_possible_values(m_filter.raw_capture_range());`

Here to fix the issue and avoid putting more objects on the GC heap.

sideshowbarker: Thanks. Made it so. That does work of course — since that’s exactly what `GC::Function` is doing internally anyway (just with an (unnecessary) allocation). Seems obvious now…

PR #9955, opened 2026-06-07
LibURL: Reverse host labels before searching the public suffix table

tcl3:
The generated public suffix table stores each entry with it's labels reversed, but the lookup we were using didn't reverse its input before searching. This led to multi-label public suffixes not matching when they should have.

NB: We previously had a unit test intended to cover this that used "com.br", but this was passing coincidentally because "br.com" is also a valid public suffix.

This fixes an issue on https://autotrader.co.uk/car-search, where requests to the image-serving CDN were returning 403 because we were sending a cookie from the wrong registerable domain due to this bug

Before:

image

After:

image

### DIFF
--- a/Meta/Generators/generate_public_suffix_data.py
+++ b/Meta/Generators/generate_public_suffix_data.py
@@ -81,9 +81,21 @@ PublicSuffixData::PublicSuffixData()
 {    
 }
 
+static bool is_reversed_public_suffix(StringView reversed_host)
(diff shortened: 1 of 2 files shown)


shannonbooth (approved):
Ha, I was just about to start looking into replacing this with libpsl. I actually think we have another bug here with non-ascii hosts matching, but that's a different problem

PR #9972, opened 2026-06-08
CI: Bake wasm-tools into the build image, and cache + retry the download

sideshowbarker:
**Problem:** CI build jobs can fail before they even start building, when the `wasm-tools` download from the bytecodealliance releases CDN returns a 504. `wasm-tools` is needed at build time to generate the WebAssembly spec tests (`INCLUDE_WASM_SPEC_TESTS`).

**Cause:** every build job fetches `wasm-tools` from that CDN at setup time, so a single hiccup there takes out multiple jobs at once.

**Fix:** Two parts: **Part 1: Bake `wasm-tools` into the `ladybird-ci` image.** The Linux build jobs all run in that image, so baking the pinned `wasm-tools` 1.243.0 in lets them get it from the image instead of downloading it per run. A follow-up moves the workflow image pins and drops the per-job download for the container platforms, once this image has been published. **Part 2: Cache and retry the download that remains.** The macOS and Windows jobs run on native runners that don’t use the image, so they still fetch `wasm-tools`. Cache the binary instead of re-fetching it on every build job, and on a cache miss retry the download with backoff.


.github/actions/setup/action.yml
```diff
@@ -131,19 +132,49 @@ runs:
...
+        attempt=1
+        until curl --fail --location --output "${FILENAME}" "${URL}"; do
+          if (( attempt > ${#delays[@]} )); then
+            echo "wasm-tools download failed after $(( ${#delays[@]} + 1 )) attempts" >&2
+            exit 1
+          fi
+          echo "wasm-tools download failed; retrying in ${delays[attempt-1]}s" >&2
+          sleep "${delays[attempt-1]}"
+          attempt=$(( attempt + 1 ))
+        done
```
gmta, line 165: Instead of working around the unavailability of the wasm-tools, couldn't we include them in our CI build image instead? That only needs to build successfully once.

sideshowbarker: Yeah, that'd definitely be better. I can write up an alternative patch for that

sideshowbarker: OK, I’ve pushed a commit here that bakes the pinned `wasm-tools` into the image. But we’ve still got the macOS arm64 Sanitizer and Windows x86_64 Sanitizer runners with `container: 'null'` that never see the image. So for the sake of those, I’ve kept the cache+retry rather than just wholly dropping it.

Once this PR is merged and the new image publishes, I’ll do a follow up PR that moves the workflow image pins to the new tag and drops the per-job download on Linux — like what was done in #9833 then #9835.

gmta: Could you limit the CI steps to the platforms that do not run the CI image?

sideshowbarker: Yeah, that’s what I planned to do in the follow-up PR after we first merge this.
 
And the reason I reckoned we need two merges is: The image bake-in ships in the new image (`2026.06.14`), which only publishes after this hits master, and the workflow image pins here still point at `2026.06.02` (pre-bake-in). So it seems like having the cache/install steps gated off Linux in _this_ PR would leave the Linux jobs with no wasm-tools at all (old image + no download), and `find_program(wasm-tools REQUIRED)` would fail.

So I think we either need:

1. Merge this PR first as-is, then I do a quick follow-up PR which bumps the 8 pins and gates these steps to macOS/Windows together — which is how things seem to have been done for #9833 and #9835.
2. Do it all here: I update this PR do add the OS gating + bump the pins. But if we do that, it seems like we’d need to publish `2026.06.14` first via a manual "Build CI Image" run (workflow_dispatch) on this branch, since PR builds don’t push.

Does that make sense? Or maybe there’s something I’m misunderstanding?

gmta: Gotcha, makes sence!

PR #9973, opened 2026-06-08
Split Web API objects from their JavaScript wrappers

awesomekling:
This refactors LibWeb so Web API implementation objects are no longer themselves JavaScript objects.

Before this PR, DOM/Web objects were implemented as JS-facing platform objects. The internal object carried JavaScript object machinery (shape, property storage, prototype) even when it was only being used by the engine.

After this PR, the internal object is a `Wrappable`, and JavaScript sees a separate wrapper object only when script actually needs one.

Before:

```mermaid
flowchart LR
    JS["JavaScript"] <--> Element["DOM::Element<br/>implementation + JS-facing object"]
```

After:

```mermaid
flowchart LR
    JS["JavaScript"] <--> Wrapper["Bindings::ElementWrapper<br/>JS-facing object"]
    Wrapper <--> Element["DOM::Element<br/>implementation object"]
```

A **wrapper** is the JavaScript object that page script sees. When JavaScript evaluates `document.body`, it receives a JS wrapper for the underlying `DOM::Element`. The wrapper has the prototype, properties, WebIDL behavior, and JavaScript object state. The underlying `DOM::Element` stores the actual DOM state and behavior.

A **wrapper world** is a namespace for those wrappers. The page's normal JavaScript runs in the main wrapper world. Browser extensions or isolated scripts can run in their own worlds. There is one wrapper identity per (implementation object, world) pair, so each world can have its own wrapper for the same underlying object:

```mermaid
flowchart TB
    Element["DOM::Element<br/>one implementation object"]

    Main["Main world wrapper<br/>seen by page JS"]
    Extension["Extension world wrapper<br/>seen by extension JS"]

    Element --> Main
    Element --> Extension
```

This is what browser extensions need. An extension content script interacts with the same page DOM as the page, but must not share the page's JavaScript wrapper objects. The page and the extension point at the same underlying `DOM::Element`, but they get different JS wrappers, so their expandos, prototypes, wrapper identity, and observable JS object state stay isolated.

```text
same DOM node: yes
same JS wrapper object: no
```

Main-world wrapper identity is owned by the HTML agent, so same-agent windows (e.g. same-origin `window.open()`) share wrapper identity, as they must.

This also reduces memory usage. Internal Web objects no longer inherit the full JS object hierarchy just because they might someday be exposed to JavaScript. A DOM node created by the parser carries no JS object state until script actually observes it. If no script ever asks for a wrapper, no wrapper exists.

It also means the JS wrapper is the stable boundary, and the implementation side is free to change behind it: ref-counted objects, more compact data structures, Rust-backed objects, all without affecting JS identity. Implementation objects are GC cells today, but nothing outside the bindings layer is allowed to assume that.

The bindings layer is now the only place that crosses between JavaScript and implementation code: it chooses realms, creates wrappers, unwraps JS values, converts WebIDL types, and vends promises.

```mermaid
flowchart TB
    JS["JavaScript"]
    Wrapper["JS wrapper<br/>PlatformObject"]
    Boundary["Bindings layer<br/>conversion + realm choice"]
    Impl["Web implementation object<br/>Wrappable"]

    JS <--> Wrapper
    Wrapper <--> Boundary
    Boundary <--> Impl
```

The realm rule is uniform: instance members use the validated receiver's realm for exceptions and return-value wrapping, statics and constructors use the caller's realm, and `[NeedsCallerRealm]` is the single opt-out. Implementation code never mentions `JS::Realm`, and a new CI lint enforces that with a legacy allowlist whose per-component ceilings can only shrink. The same lint rejects new implementation classes deriving from `PlatformObject` instead of `Wrappable`.

Wrapper caches are weak. A wrapper with no surviving JS references and no non-default JS state may be collected and re-materialized later, which is unobservable by design. Any non-default state (expandos, a replaced prototype, subclassed construction) preserves the wrapper for the implementation object's lifetime. Objects with pending observable work (timers, fetches, observers) keep themselves alive with the new `GC::ActivityRoot`.

Every commit in the series builds on its own. LibGC gains `ActivityRoot` and live-entry iteration for weak hash maps. LibJS gains a per-object flag so platform hooks can bypass AddOwnProperty inline caches, and `TypeErrorRealmScope` so generated conversions mint TypeErrors in the binding-selected realm; implicit `JS::Value` conversion from implementation types is deleted, they must go through `wrap()`. Then the LibWeb integration itself, the lint, tests, and documentation.

The integration commit is large because the old assumption was everywhere: many APIs expected implementation objects to also be JS-facing objects. The JS-facing adaptation moves into binding-side helpers and generated bindings, while the actual DOM/Web behavior stays on the implementation objects.

`Documentation/WrapperArchitecture.md` is the place to start reading, and the generator's realm policy is in `Meta/Generators/libweb_bindings/REALM_POLICY.md`.

The end result is a clearer layering:

```text
Wrappable
  Internal Web object.

PlatformObject wrapper
  JavaScript-facing object.

WrapperWorld
  The wrapper identity namespace, such as page world or extension world.

Bindings layer
  The bridge between JavaScript values and internal Web objects.
```

This makes extension worlds possible, reduces unnecessary JS object overhead, and gives us a cleaner path to future internal representation changes.


Libraries/LibWeb/Bindings/PlatformObject.cpp
```diff
@@ -449,64 +524,208 @@ JS::ThrowCompletionOr<GC::RootVector<JS::Value>> PlatformObject::internal_own_pr
...
+    if (auto* wrappable = wrappable_impl())
+        if (auto* storage = as_if<HTML::Storage>(wrappable))
+            return Bindings::set_value_of_named_property(realm, *storage, name, value);
     VERIFY_NOT_REACHED();
 }
 
-WebIDL::ExceptionOr<void> PlatformObject::set_value_of_new_indexed_property(u32, JS::Value)
+WebIDL::ExceptionOr<void> PlatformObject::set_value_of_new_indexed_property(JS::Realm& realm, u32 index, JS::Value value)
 {
+    if (auto* wrappable = wrappable_impl()) {
```
tcl3: Does `HTMLSelectElement` need a branch here and `set_value_of_existing_indexed_property()` below?

Currently this crashes:
```html
<!DOCTYPE html>
<select>
    <option></option>
</select>
<script>
    document.querySelector("select")[0] = document.createElement("option");
</script>
```

Libraries/LibWeb/Bindings/SharedWorkerGlobalScopeGlobalMixin.h
```diff
@@ -0,0 +1,68 @@
+/*
+ * Copyright (c) 2026-present, the Ladybird developers.
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#pragma once
+
+#include <LibJS/Runtime/Object.h>
```
shannonbooth: I'm quite confused about this mixin file and quite a few others which have landed in this directory - what is different between this file and what we were generating beforehand in the build folder? What makes DedicatedWorkerGlobalScope SharedWorkerGlobalScope and Window special but not something like ServiceWorkerGlobalScope? (still trying to go through all the bindings generator changes here!)

shannonbooth: I think bigger picture what I'm trying to figure out is this folder altogether - which parts should be generated, and which parts we won't

Meta/Generators/libweb_bindings/interface_declaration.py
```diff
@@ -54,6 +60,89 @@ def write_declaration(
...
+    {wrapper_class_name(interface)}(JS::Realm&, GC::Ref<{impl_type}>);
+    virtual ~{wrapper_class_name(interface)}() override;
+
+    virtual StringView class_name() const override {{ return "{interface.name}"sv; }}
+    virtual void initialize(JS::Realm&) override;
+"""
+        )
+        if "Global" in interface.extended_attributes:
+            out.write("    virtual JS::ThrowCompletionOr<bool> internal_set_prototype_of(JS::Object*) override;\n")
+        if "LocationObject" in interface.extended_attributes:
```
shannonbooth: Small tidy up we can do here is remove the newly made "LocationObject" extended attribute and can just special case on interface.name == "Location" for it's exoticness, and move the includes for it above in this block here where these are used. Though, having said that, I wonder if for this particularly weird case which overrides normal JS object stuff that isn't described by IDL  if we should have an opt out for the IDL generator to generate the wrapper object and to expect a handwritten version instead. That way we won't have an in-between of the header being defined partly in python for the headers, and partly in LocationWrappers.cpp for the exotic-ness since this and WindowProxy are a bit weird

Meta/Generators/libweb_bindings/attributes.py
```diff
@@ -208,13 +234,22 @@ def write_attribute_getter(
...
+        "realm" if "NeedsCallerRealmForGetter" in attribute.extended_attributes else "this_object_realm"
+    )
+    getter_steps = f"auto R = TRY(throw_dom_exception_if_needed(vm, {getter_realm_argument}, [&] {{ return {implementation_attribute_getter_call(interface, attribute, getter_realm_argument)}; }}));"
     is_reflected = "Reflect" in attribute.extended_attributes
     is_non_nullable_reflected = is_reflected and not attribute.type.nullable
     is_non_nullable_reflected_string = is_non_nullable_reflected and attribute.type.name == "DOMString"
     is_reflected_usv_string = is_non_nullable_reflected and attribute.type.name == "USVString"
 
-    if is_reflected and attribute.type.name == "boolean":
+    if "ReturnsJSValue" in attribute.extended_attributes:
```
shannonbooth: Every instance of IDL returning "any" that I can find uses ReturnsJSValue (search might be wrong!), maybe the extended attribute be type inspection instead?

(Also just digging into the realm attribute rules atm, what are the rules behind that one? trying to figure out if we can infer that as well)

Meta/Generators/libweb_bindings/iterables.py
```diff
@@ -522,47 +552,49 @@ def write_maplike_declaration_functions(
...
 
-    // 2. Let map be the map entries of the IDL value that represents a reference to O.
-    GC::Ref<JS::Map> map = this_impl->map_entries();
-
     // 3. Let keyType be the key type specified in the maplike declaration.
     // 4. Let keyArg be the first argument passed to this function, or undefined if not supplied.
     // 5. Let key be keyArg converted to an IDL value of type keyType.
-    auto key = vm.argument(0);
-    {type_check_idl_value(interface.maplike.key_type, "key", includes, context, interface.name)}
+    auto key_arg = vm.argument(0);
```
shannonbooth, line 611: The support for conversion to IDL value here was limited beforehand (and seems like we only use it for CSS at the moment) but now I think we should be able to make this generic for any type. Beforehand we needed `type_check_idl_value` since map->map_get took a JS value. But now that it takes an actual string, I think be able to use `to_idl_value` here (which also supports FlyString conversion). This also fixes a bug where USVString wouldn't go through the scalar path (allowed by the USVString path, though no IDL currently uses that). Any chance here and in some of the other inlined code we can also not drop some of the spec comments (like step 2 above)?

shannonbooth: Though actually, maybe we don't actually support the syntax for that in a set/maplike definition, in which case it might just be a case of ignoring this as a future cleanup target 😅

Libraries/LibWeb/HTML/BroadcastChannel.h
```diff
@@ -43,14 +47,18 @@ class BroadcastChannel final : public DOM::EventTarget {
...
     virtual void finalize() override;
+    virtual void visit_edges(Cell::Visitor&) override;
 
+    JS::Object& relevant_global_object() const;
     bool is_eligible_for_messaging() const;
 
     FlyString m_channel_name;
+    URL::Origin m_origin;
+    StorageAPI::StorageKey m_storage_key;
+    GC::Ref<DOM::EventTarget> m_global_object;
```
shannonbooth, line 61: This pattern is done in a few places, but would it be possible to have this abstracted on Wrappable instead of storing global object on a bunch of these?

Meta/Generators/libweb_bindings/cpp_types.py
```diff
@@ -562,6 +562,9 @@ def add_binding_include_for_type(idl_type: IDLType, includes: GeneratedIncludes,
 
     dictionary = context.dictionary(idl_type)
     if dictionary is not None:
+        if dictionary.name == "StructuredSerializeOptions":
+            includes.add("LibWeb/Bindings/MessagePort.h")
```
shannonbooth, line 583: This special case does not seem needed

Meta/Generators/libweb_bindings/constructors.py
```diff
@@ -17,6 +17,16 @@
 from Utils.webidl_parser import Interface
 
 
+def bindings_wrapper_header_for_constructor(interface: Interface) -> str:
+    return "LibWeb/Bindings/ImplementedInBindings.h"
+
+
+def bindings_constructor_call(interface: Interface, arguments: str, needs_argument_count: bool) -> str:
+    arguments_with_leading_comma = f", {arguments}" if arguments else ""
+    argument_count = ", vm.argument_count()" if needs_argument_count else ""
```
shannonbooth, line 26: Just as the background for why this workaround needs to exist incase it changes thinking here, the reason why I think we need to do this is because of: https://github.com/LadybirdBrowser/ladybird/blob/fc02ab95fe559ef0fb5af5358e4f8722257ec41b/Meta/Generators/libweb_bindings/cpp_types.py#L239

Basically the intent here is that if something is an optional argument (without any default parameter) _and_ also nullable, code generation folds merged the nullability and optionality together so that we don't have something like `Optional<GC::Ptr<T>>` or `Optional<Optional<T>>`. Most of the time this results in nicer code since specs generally treat argument missing as the same as a null argument (and old cpp generator had been doing this for most cpp types beforehand, perhaps by accident). But Animation constructor is a special case to that, and I suppose this whole wrappers things means we can't just poke into vm.argument_count() any more for it. I could also just remove that special handling for optional nullable operation parameters though

awesomekling:
@coderabbitai please review

PR #9974, opened 2026-06-08
LibWeb: SVG path parsing optimizations

Calme1709:
A few optimizations to SVG path parsing, primarily around avoiding copies and `Vector` allocations.

Collectively reduce time spent in `SVGPathElement::attribute_changed` by 25% (from 120ms to 90ms) when loading https://chatgpt.com on my machine.

### DIFF
--- a/Libraries/LibWeb/SVG/Path.cpp
+++ b/Libraries/LibWeb/SVG/Path.cpp
@@ -21,8 +21,8 @@ static void serialize_path_instruction(PathInstruction const& instruction, Strin
             builder.appendff(
                 "{} {} {}",
                 move_to.absolute ? 'M' : 'm',
-                move_to.point[0],
-                move_to.point[1]);
+                move_to.point.x(),
+                move_to.point.y());
         },
         [&](ClosePathInstruction const&) {
             // NB: This is always canonicalized as Z, not z.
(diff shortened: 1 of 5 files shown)


Libraries/LibWeb/SVG/Path.cpp
```diff
@@ -14,275 +14,275 @@
...
+                    c2 += last_point;
+                }
+                path.cubic_bezier_curve_to(c1, c2, p2);
+                previous_control_point = c2;
+            },
+            [&](SmoothCurveToInstruction const& smooth_curve_to_instruction) {
+                clear_last_control_point = false;
+
+                if (!previous_control_point.has_value()
+                    || (!last_instruction->has<CurveToInstruction>() && !last_instruction->has<SmoothCurveToInstruction>())) {
```
tcl3: The `SmoothQuadraticBezierCurveToInstruction` case above guards against `last_instruction` being null. Should we have the same guard here?

Calme1709: We don't technically need it, since `previous_control_point` can never have a value without `last_instruction` being non-null, but we may as well.

tcl3 (approved):

PR #9985, opened 2026-06-08
LibURL: Use libpsl for public suffix matching

### LINKED ISSUE #5327: LibURL/LibWebView: Investigate libpsl library instead of manually downloading psl list
libcurl relies on the `libpsl` library for cookie rules.

We currently download the public suffix list from github at the `master` branch, which is super not ideal.

The libpsl library also seems to have facilities in place to load a new set of PSL rules from disk if required.

https://github.com/rockdaboot/libpsl
https://rockdaboot.github.io/libpsl/libpsl-Public-Suffix-List-functions.html

The library is MIT.

The most interesting part to using this lib seems to be context management. Whether we use a builtin PSL data structure or load one from disk, we should avoid confusing TOCTOU issues that could come about by opening a new context for each global object, for example. Or even for each WebContent process, as they are spawned over time.

cc @shannonbooth

shannonbooth:
Replace the generated public suffix table and custom matcher with a direct LibURL PublicSuffixData implementation backed by libpsl. This drops our PSL download/generator path and uses the same library already used by libcurl.

Performance is comparable before and after, while LibURL binary size is smaller.

Fixes: https://github.com/LadybirdBrowser/ladybird/issues/5327


vcpkg.json
```diff
@@ -349,6 +350,10 @@
       "name": "libproxy",
       "version": "0.4.18#3"
     },
+    {
+      "name": "libpsl",
+      "version": "0.21.5#1"
+    },
```
gmta, line 356: This should end up in the flatpak manifest as well.

shannonbooth: Done! (hopefully, first time touching flatpak)

PR #9988, opened 2026-06-09
LibWeb: Run `removed_from` for descendants when shadow root removed

### LINKED ISSUE #9999: Intermittent WebContent crash with Layout/input/pdf-viewer.pdf: VERIFY(m_dom_node) during paint-cache release
`Tests/LibWeb/Layout/input/pdf-viewer.pdf` intermittently crashes under CI — roughly once in a few dozen runs. Backtrace:

```
ak_verification_failed
Web::Layout::Node::document() const                          <-- VERIFY(m_dom_node)
Web::Painting::Paintable::navigable() const
Web::Painting::PaintableBox::release_cache_references_for_cached_commands(AK::Span<unsigned char const>) const
Web::Painting::PaintableBox::invalidate_paint_cache() const
Web::Layout::Node::prepare_for_detach_from_layout_tree()
Web::Layout::Node::prepare_subtree_for_detach_from_layout_tree()
Web::Layout::TreeBuilder::update_layout_tree(DOM::Node&, Context&, MustCreateSubtree)   (recursive descent)
Web::Layout::TreeBuilder::build(Web::DOM::Node&)
Web::DOM::Document::update_layout(Web::DOM::UpdateLayoutReason)
Web::HTML::HTMLElement::offset_parent() const
Web::Bindings::HTMLElementPrototype::offset_parent_getter(JS::VM&)
... JS bytecode (the PDF viewer reading offsetParent) ...
```

The failing assertion is the `VERIFY(m_dom_node)` at `Libraries/LibWeb/Layout/Node.cpp:1380`; the release is the `for_each_present_phase` lambda at `Libraries/LibWeb/Painting/PaintableBox.cpp:505`.

So, the PDF viewer reads `offsetParent`, forcing `update_layout()` and a tree rebuild that detaches a `PaintableBox` still holding cached display-list commands — and releases its cached resource references. That release reaches the navigable via `document()` — which dereferences the layout node’s weak `DOM::Node` pointer. But the DOM node’ already been GC’ed. So the pointer is null and the `VERIFY` trips.

You should be able to reproduce it locally with something like this:

```
Build/release/bin/test-web \
  --test-path Tests/LibWeb \
  --filter '*pdf-viewer*' \
  --repeat 8000 -j 12 --fail-fast --per-test-timeout 60
```

But you probably won’t need to do anywhere near 8000 repeats. The last time I tried it myself, it tripped at run 45:

```
Halted; 7957 tests not executed.
Pass: 43, Fail: 0, Skipped: 0, Timeout: 0, Crashed: 1
Crashed: (run 45/8000) Layout/input/pdf-viewer.pdf
```

But you also should be able to reproduce it 100% of the time with this document:

[shadow-root-children-removed-on-attribute-change.html](https://github.com/user-attachments/files/28755448/shadow-root-children-removed-on-attribute-change.html)

Calme1709:
This is only relevant for UA-internal shadow roots (specifically those created by `HTMLInputElement` and `MediaControls`) since they are the only ones which can be removed from their hosts.
    
Previously after removing a shadow root from it's host we left it's descendants' paintables' caches to be cleared during the next layout update.
    
This was fine prior to 9340d2d, when layout nodes kept the relevant DOM nodes alive, however, layout nodes now only keep weak references so these DOM nodes can be GC'd before the paintables' caches are removed causing a crash.
    
We now run the `removed_from` steps for the shadow tree's elements before removing the shadow root from its host in line with how we handle removal of DOM nodes in other cases - this clears their paintables' caches immediately while the DOM nodes are still alive.
    
Fixes an intermittent [crash](https://github.com/LadybirdBrowser/ladybird/actions/runs/27138656025/job/80097595707?pr=9977) in Layout/input/pdf-viewer.pdf. Fixes https://github.com/LadybirdBrowser/ladybird/issues/9999.

### DIFF
--- a/Libraries/LibWeb/DOM/Element.cpp
+++ b/Libraries/LibWeb/DOM/Element.cpp
@@ -1493,6 +1493,14 @@ void Element::set_shadow_root(GC::Ptr<ShadowRoot> shadow_root)
     if (m_shadow_root) {
         m_shadow_root->set_host(nullptr);
         m_shadow_root->set_is_connected(false);
+        // NB: We don't need to run the removed steps if the children have already been disconnected (or were never
+        //     connected in the first place)
+        if (is_connected()) {
+            m_shadow_root->for_each_shadow_including_descendant([&](DOM::Node& descendant) {
+                descendant.removed_from(IsSubtreeRoot::No, m_shadow_root, *m_shadow_root);
+                return TraversalDecision::Continue;
+            });
+        }
     }
     m_shadow_root = move(shadow_root);
     if (m_shadow_root) {
--- a/Libraries/LibWeb/DOM/Node.cpp
+++ b/Libraries/LibWeb/DOM/Node.cpp
@@ -1804,13 +1804,29 @@ void Node::inserted()
     set_needs_style_update(true);
 }
 
+void Node::clear_layout_node_paintables()
+{
+    if (!m_layout_node)
+        return;
+
+    m_layout_node->clear_paintables();
+
+    // NB: Block-in-inline splitting can create multiple layout nodes for a single DOM node. Only the last one is stored
+    //     in m_layout_node so we need to clear the paintables for the continued nodes as well
+    auto* node_with_metrics = as_if<Layout::NodeWithStyleAndBoxModelMetrics>(*m_layout_node);
+    if (!node_with_metrics)
+        return;
+
+    for (auto* continuation = node_with_metrics->continuation_of_node(); continuation; continuation = continuation->continuation_of_node())
+        continuation->clear_paintables();
+}
+
 void Node::removed_from(IsSubtreeRoot, Node*, Node&)
 {
     m_is_connected = false;
     m_in_editable_subtree = false;
     m_inside_blocking_wheel_event_handler = false;
-    if (m_layout_node)
-        m_layout_node->clear_paintables();
+    clear_layout_node_paintables();
     m_layout_node = nullptr;
     m_paintable = nullptr;
 
--- a/Libraries/LibWeb/DOM/Node.h
+++ b/Libraries/LibWeb/DOM/Node.h
@@ -539,6 +539,7 @@ private:
     void insert_before_impl(GC::Ref<Node>, GC::Ptr<Node> child);
     void append_child_impl(GC::Ref<Node>);
     void remove_child_impl(GC::Ref<Node>);
+    void clear_layout_node_paintables();
 
     static Optional<StringView> first_valid_id(StringView, Document const&);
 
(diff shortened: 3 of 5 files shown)


sideshowbarker:
This will still fail with the following document: [shadow-root-children-removed-on-attribute-change.html](https://github.com/user-attachments/files/28755448/shadow-root-children-removed-on-attribute-change.html)

That’s a different case than the `Layout/input/pdf-viewer.pdf` case. But I think they have the same underlying cause.

sideshowbarker:
For the record here too — this is the test I had meant to attach: [paint-cache-release-after-node-gc.html](https://github.com/user-attachments/files/28792057/paint-cache-release-after-node-gc.html)

Calme1709:
I have added a new commit that fixes that case as well.

sideshowbarker (commented):
The fix itself looks right to me, and everything I could throw at it now passes:

- the two tests in the PR, plus the test I attached earlier (which previously crashed 3/10 with only the first commit): 20/20 runs pass
- `Layout/input/pdf-viewer.pdf` with `--repeat 8000 --fail-fast -j 12`: 8000/8000 pass
- a double-split probe I wrote to poke at the continuation linkage (second block child  inserted into an already-split inline, painted between the splits, then removed and collected): 20/20 pass

Two things on the tests, one thing of which I think needs fixing before this lands:

1. **Both new crash tests pass on an unfixed build** — on my machine (macOS arm64, Sanitizer, PR commits reverted): `block-in-inline-removal-crash.html` passes 10/10 and `shadow-root-children-removed-on-attribute-change.html` passes 5/5 without the fix. So as regression tests, it seems they won’t catch regressions.

   I think the cause is that a single `internals.gc()` right after the removal doesn’t reliably collect the node: the element is still reachable through stale stack/register references (and the id-global), and Ladybird’s collector is conservative.

   The [paint-cache-release-after-node-gc.html](https://github.com/user-attachments/files/28792057/paint-cache-release-after-node-gc.html) test I attached earlier works around exactly that — it does the removal inside a function frame that returns, allocates a few thousand throwaway objects to churn the stack, calls `internals.gc()`, and yields to the event loop, looping ~30 times.

   With that shape it crashes 10/10 on the unfixed build and passes 20/20 on this branch. Maybe you could you fold that pattern into the two tests? (Or just adopt the attached test for the continuation case?)

2. Nit: the new test’s `<script src="../Tests/LibWeb/Text/input/include.js">` path doesn’t resolve from `Tests/LibWeb/Crash/DOM/`. But anyway the test doesn’t use anything from include.js anyway — so it seems like that script could just be dropped.

Calme1709:
Odd, those two tests crash reliably on my machine (x86 Linux). I have updated them to be based on your repro instead, can you confirm they crash reliably now?

sideshowbarker (approved):
LGTM — tests reliably crashing now on the pre-fixed code state.

PR #10014, opened 2026-06-09
LibWeb+LibWebView+WebContent: Allow muted media to autoplay by default

Lubrsi:
The autoplay setting was binary and its default blocked all media, including muted video, leaving sites that rely on muted autoplay visibly broken. Replace it with a tri-state user-agent autoplay policy (allow audio and video, block audio, or block audio and video) defaulting to allowing only inaudible media to autoplay.

This is enforced through the media element's "allowed to play" check, so unmuting a muted autoplay or calling `play()` cannot slip audio past the policy; audible playback is permitted once the document has been activated by the user. The policy lives in a dedicated AutoplaySettings consulted from HTMLMediaElement instead of the Permissions Policy "allowed to use feature" check it was previously conflated with.

Before:

https://github.com/user-attachments/assets/05a0b5e7-774c-40bc-bdc6-b91becf43eb2

After:

https://github.com/user-attachments/assets/813b14f7-e41f-4857-9ca3-a5968c118e7f


Zaggy1024 (approved):

Base/res/ladybird/about-pages/settings.html
```diff
@@ -642,8 +642,12 @@ <h3 id="site-settings-title" class="dialog-title"></h3>
             </div>
             <div class="dialog-body">
                 <div class="inline-container">
-                    <label for="site-settings-global">Enable on all sites</label>
-                    <input id="site-settings-global" type="checkbox" switch />
+                    <label for="site-settings-policy">Default for all sites</label>
+                    <select id="site-settings-policy">
+                        <option value="allow-audio-and-video">Allow Audio and Video</option>
```
trflynn89: One thing is that this dialog was previously agnostic to the specific site setting being configured. We currently only use this for autoplay, but this dialog would also be used for e.g. location and notification permissions. Not sure how we want to handle some settings having more of a tri-state default.

Lubrsi: I've made it agnostic again. I'm not quite sure what you mean with the last sentence, do you mean the allowlist not having a separate selection per site with each option?

trflynn89: What I meant was that autoplay is a bit different in that it has 3 options for its default-for-all-sites, as opposed to others having a binary on/off state. So while autoplay's default options should be displayed with a select menu, others may more naturally want a toggle or no default option choice at all.

But what you have here is fine for now, we can deal with how other options should look when we implement them.

Libraries/LibWeb/HTML/AutoplaySettings.cpp
```diff
@@ -0,0 +1,95 @@
...
+{
+    // An origin in the allowlist may always autoplay, with or without audio.
+    for (auto const& allowed : m_allowlist) {
+        if (allowed.is_same_origin_domain(origin))
+            return AutoplayDecision::Allowed;
+    }
+
+    // AD-HOC: Allow autoplay for file:// URLs if the document is also from a file:// URL.
+    if (origin.is_opaque_file_origin() && document.origin().is_opaque_file_origin())
+        return AutoplayDecision::Allowed;
```
trflynn89, line 35: This was previously controlled by the site setting list itself. If that's no longer feasible, we should just remove the default `file://` rule from Settings.cpp

Libraries/LibWebView/Settings.h
```diff
@@ -37,7 +38,7 @@ struct BrowsingBehavior {
 struct SiteSetting {
     SiteSetting();
 
-    bool enabled_globally { false };
+    Web::HTML::AutoplayPolicy policy { Web::HTML::AutoplayPolicy::BlockAudio };
```
trflynn89, line 46: Didn't see this before, but this is another thing that was a generic site setting structure that is now very autoplay specific.

PR #10016, opened 2026-06-09
Meta: Some Linux build hardening

awesomekling:
Enable additional hardening for platform builds:

- Use full RELRO on Linux with `-z relro` and `-z now`.
- Enable `-fstack-clash-protection` on Linux.
- Define `_FORTIFY_SOURCE=3` on Linux and `_FORTIFY_SOURCE=2` on macOS.

These are low-overhead hardening defaults for browser processes and keep the new flags scoped to platforms where they are expected to be useful.

### DIFF
--- a/Meta/CMake/compile_options.cmake
+++ b/Meta/CMake/compile_options.cmake
@@ -199,6 +199,8 @@ endif()
 
 if (LINUX)
     add_cxx_compile_definitions(_FILE_OFFSET_BITS=64)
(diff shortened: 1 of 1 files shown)


wertzuz:
MacOS?

EWouters:
ec3ea7e breaks the build on Gentoo and other systems which have `-D_FORTIFY_SOURCE=2` set by default.

PR #10018, opened 2026-06-09
LibWeb: Record abspos inline static positions as line-box markers

### LINKED ISSUE #1952: Absolut inside <pre> is wrongly positioned
Minimal reproducer:

```html
<style>
  .a {
    position: absolute;
  }
</style>
<pre>
<span class="a">1</span>  fo<span>o</span>
<span class="a">2</span>  fo<span>o</span>
<span class="a">3</span>  fo<span>o</span>
</pre>
```

Screenshots (Firefox left, Ladybird right):
minimal

This is a reduction from web content, particularly the following Wikipedia page:
https://en.wikipedia.org/wiki/Methods_of_computing_square_roots#Binary_numeral_system_(base_2)
Screenshots (Firefox left, Ladybird right):
wiki

I'd be happy to work on this myself, but I'm stuck: I guess that it's `{Block,Inline}FormattingContext` somehow, but I'm not sure whether these even are where the position is set.

Maybe `InlineFormattingContext::run` is supposed to set `top` of inline elements while running somehow? `top` is different between Firefox and Ladybird:

```
<style>
  .a {
    position: absolute;
  }
</style>
<pre>
<span class="a">1</span>  fo<span>o</span>
<span class="a">2</span>  fo<span>o</span>
<span class="a" id="three">3</span>  fo<span>o</span>
</pre>
<p id="result"></p>
<script>
    var computed = window.getComputedStyle(document.getElementById("three"))["top"];
    document.getElementById("result").textContent = computed;
</script>
```

Suggested tags: css, has repro, reduction of web content

shannonbooth:
The static position of an absolutely positioned inline child is the in-flow insertion point on the line where it appears. Previously this was reconstructed after layout by walking previous siblings for a line-box fragment. That lookup could match a fragment from an earlier line and collapse multiple abspos children onto the same position, especially in white-space preserving content.

Instead, drop a zero-width static-position marker into the line box at the insertion point when each abspos child is encountered. The marker is carried through normal line post-processing, including float intrusion, ext-align, justification, trailing-whitespace trimming and ellipsis, so the final static position is resolved from the line itself.

Resolve marker-only trailing lines before removing them, so they can provide static position without contributing line height.

Fixes: #1952 

Before:

image

After:

image

### DIFF
--- a/Libraries/LibWeb/Layout/InlineFormattingContext.cpp
+++ b/Libraries/LibWeb/Layout/InlineFormattingContext.cpp
@@ -242,7 +242,12 @@ void InlineFormattingContext::apply_justification_to_fragments(CSS::TextJustify 
 
         if (fragment.is_justifiable_whitespace()
             && fragment.inline_length() != justified_space_width) {
-            running_diff += justified_space_width - fragment.inline_length();
+            auto diff = justified_space_width - fragment.inline_length();
+            running_diff += diff;
+            for (auto& marker : line_box.static_position_markers()) {
+                if (marker.inline_offset > fragment.inline_offset())
+                    marker.inline_offset += diff;
+            }
             fragment.set_inline_length(justified_space_width);
         }
     }
@@ -339,6 +344,7 @@ void InlineFormattingContext::apply_text_overflow_ellipsis(Vector<LineBox>& line
                 fragments[j].set_fully_truncated(true);
 
             line_box.m_inline_length = available_width;
+            line_box.clamp_static_position_markers_to_inline_length();
             break;
         }
     }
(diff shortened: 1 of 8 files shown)


Libraries/LibWeb/Layout/InlineFormattingContext.cpp
```diff
@@ -564,48 +589,6 @@ void InlineFormattingContext::set_vertical_float_clearance(CSSPixels vertical_fl
...
 }
 
-StaticPositionRect InlineFormattingContext::calculate_static_position_rect(Box const& box) const
-{
-    VERIFY(box.parent());
-    VERIFY(box.parent()->children_are_inline());
-
-    CSSPixelPoint position;
-
-    // We're calculating the position for an absolutely positioned box in an IFC.
```
shannonbooth, line 574: I did have a different approach which atttempted to fix this calculation, but I think doing it this way is better than preserving this walk backwards

PR #10020, opened 2026-06-10
CI: Run a WPT smoke test step on PRs

tcl3:
Run `WPT.sh` against a set of known good tests in `Tests/LibWeb/WPT/ci-smoke-tests.txt` as a smoke test.

To support this I've added the following to `WPT.sh`:
* A `--test-list` option that pulls tests from a file, ignoring lines starting with `#`.
* A `BUILD_LADYBIRD` env var to allow skipping the build step.
* A `CI` env var that allows running as root.

### DIFF
--- a/.github/workflows/lagom-template.yml
+++ b/.github/workflows/lagom-template.yml
@@ -215,6 +215,19 @@ jobs:
             --python-executable "${{ env.pythonLocation }}/bin/python" \
             | tee -a "$GITHUB_STEP_SUMMARY"
 
+      - name: Run WPT Smoke Tests
+        if: ${{ inputs.os_name == 'Linux' && (inputs.build_preset == 'Release' || inputs.build_preset == 'Sanitizer') }}
+        timeout-minutes: 5
+        working-directory: ${{ github.workspace }}
+        run: Meta/WPT.sh run --test-list Tests/LibWeb/WPT/ci-smoke-tests.txt
+        env:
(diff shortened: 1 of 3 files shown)


tcl3:
I have disabled MacOS support for now, since updating the hosts file to allow WPT.sh to run requires passwordless sudo, which we don't have.

It seems like the best thing to do would be to update the hosts file on each MacOS runner separately using `Tests/LibWeb/WPT/wpt/wpt make-hosts-file | sudo tee -a /etc/hosts`

tcl3:
I've added an explicit list of test types we support to `WPT.sh`. This prevents the WPT runner trying to pull in dependencies required by AAM tests, which we don't support.

tcl3:
I saw some flakiness in some of the CI runs I ran. I've removed a troublesome http2 test and set `WPT_PROCESS=4` and I've not seen any flakiness since.

AtkinsSJ:
Is there anything blocking this currently? I've looked through it and I don't see any issues, but then my bash knowledge is weak. (:yakbait: WPT.py wen?)

tcl3:
> Is there anything blocking this currently?

I don't believe so. I've just rebased it to make absolutely sure it still works as intended and all looks good

> WPT.py wen

The multi-process chunking that the runner uses would be a real pain to port over to python, but there's no real reason that couldn't be separated from the bulk of the script which could be ported over.

PR #10029, opened 2026-06-10
LibWeb: Prevent the session history step from moving backwards

### LINKED ISSUE #10028: Intermittent crashes applying a session history step: racing apply-history-step runs corrupt the step numbering
WebContent can intermittently crash while applying a session history step. And because test-web reuses the WebContent process across tests, in CI the crash can bork stuff up beyond just its own test: When a document under test leaves same-document history work still draining, the harness’s load of the *next* test races it — and the crash lands on whichever test happens to load next.

So, that under CI that can show up as unrelated tests sporadically crashing — which makes it hard to attribute just from CI failures. The way I came across was while doing local stress-looping for #9999.

What happens is: A synchronous same-document navigation (`history.pushState`) can jump the session history traversal queue while another apply-history-step run is paused. The nested run commits its target step and advances the traversable’s current session history step; when the paused outer run (e.g., the bookkeeping step for a newly-created iframe) later completes, it unconditionally writes its own, older target step — moving the current step *backwards*. The next Push-type navigation then computes its step as current + 1 — which an existing entry already holds. So, two entries share a step number — and applying that step trips the Push-invariant assert at `HTML/TraversableNavigable.cpp:729`:

```cpp
VERIFY(target_entry->step().get<int>() > displayed_entry->step().get<int>());
```

The same `VERIFY` family was previously reported through a real-site reduction in #9290 (shell.duckdb.org, `location.hash` plus `reload()`). That reduction stopped reproducing after #9391 — but the underlying step regression remained reachable. Backtrace:

`EXC_BREAKPOINT` / `SIGTRAP` (a `VERIFY` trap), captured on macOS arm64 from an unfixed Release build. The message:

```
VERIFICATION FAILED: target_entry->step().get<int>() > displayed_entry->step().get<int>()
  at Libraries/LibWeb/HTML/TraversableNavigable.cpp:729
```

Crashing thread:

```
ak_trap                                                          (Assertions.cpp:137)
ak_verification_failed                                           (Assertions.cpp:175)
Web::HTML::ApplyHistoryStepState::start()::$_0::operator()()     (TraversableNavigable.cpp:729)
Web::HTML::Task::execute()                                       (Task.cpp:52)
Web::HTML::EventLoop::process()                                  (EventLoop.cpp:151)
... event loop / timer pump ...
```

While fixing this, it turned out the backwards-moving step is just one face of the race: a synchronous navigation that jumps the traversal queue mutates the step bookkeeping while a paused run is mid-flight, and that corrupts things two more ways:

- A push’s finalize can compute its step while another push’s claimed step is still uncommitted — handing the same number out twice, with no backwards move involved.
- A run can read the active session history entry after a `pushState` has installed it but before the queued synchronous step has assigned its step number — tripping the pending-step assert instead:

```
VERIFICATION FAILED: has<T>() at AK/Variant.h:370
```

That assert fires from the same `ApplyHistoryStepState::start()` task; it reproduces readily on Sanitizer builds, where the race windows are wider — and it’s the same previously-hard-to-attribute `AK/Variant.h` crash that’s been seen on unrelated PRs in CI.

You can reproduce it locally with this:

```html
<!DOCTYPE html>
<body>
<script>
  const histories = [history];
  for (let f = 0; f < 16; f++) {
    const iframe = document.createElement("iframe");
    document.body.appendChild(iframe);
    histories.push(iframe.contentWindow.history);
  }
  let i = 0;
  (function round() {
    if (i++ >= 80)
      return;
    for (const h of histories)
      h.pushState({ i }, "", "#s" + i);
    setTimeout(round);
  })();
</script>
</body>
```

Run that with something like this:

```
Build/release/bin/test-web --test-path Tests/LibWeb \
  --filter '*history-push-flood*' --repeat 20 -j 1 --per-test-timeout 15
```

For me that reliably crashes on _every_ even-numbered repeat (runs 2, 4, 6, … 20 of 20): run _N_’s page poisons the session history state, and the harness’s load of run _N_ + 1 trips the VERIFY. A single isolated load of the page _never_ crashes. The _second_, externally-initiated load is what reaches the broken state — which is also why the crash escapes single-run test coverage, and shows up in CI as some *other* test crashing.

See https://github.com/whatwg/html/issues/12576; the root cause is a bug in the spec — which other engines have already implemented around.

sideshowbarker:
**Problem:** Crash in `ApplyHistoryStepState::start()` during same-document `pushState` while session-history bookkeeping for a newly-created child navigable is still in flight, and the page is then loaded again by the UI process. This has been intermittently taking down unrelated tests in CI; test-web reuses the WebContent process — so the next test’s load can arrive while the previous page’s history work is still draining.

**Cause:** A sync same-document navigation may jump the session history traversal queue while another apply-history-step run is paused — and racing runs then corrupt the step numbering. The nested run commits its target step and advances the traversable’s current session history step; when the paused outer run later completes, it unconditionally writes its own, older target step — moving the current step backwards, after which the next Push-type navigation computes a step number that an existing entry already holds. A push’s finalize also computes its step from the current step while another push’s claimed step is still uncommitted — handing the same number out twice. And a run for an arriving load can read the active session history entry after a `pushState` has installed it but before the queued synchronous step assigns its step number — tripping the pending-step assert — while clearing the forward session history could remove the in-flight push’s entry out from under its run.

See https://github.com/whatwg/html/issues/12576; the root cause is a bug in the spec — which other engines have already implemented around.

**Fix:** Number apply-history-step runs, and let runs commit a target step only if no newer run has committed it — so a stale run can no longer move the current step backwards. Track claimed-but-uncommitted steps on the traversable, compute new step numbers past them, and keep clearing the forward session history from removing their entries. And skip the pushed-step assert while the displayed entry’s step is still pending.

The first commit here adds `internals.loadURL()`, which starts a load the way a request from the UI process arrives — deferred, outside the calling task — so the regression tests can reproduce these crashes deterministically.

Fixes https://github.com/LadybirdBrowser/ladybird/issues/10028

### DIFF
--- a/Libraries/LibWeb/HTML/Navigable.cpp
+++ b/Libraries/LibWeb/HTML/Navigable.cpp
@@ -2888,7 +2888,9 @@ void finalize_a_cross_document_navigation(GC::Ref<Navigable> navigable, HistoryH
         traversable->clear_the_forward_session_history();
 
         // 2. Set targetStep to traversable's current session history step + 1.
-        target_step = traversable->current_session_history_step() + 1;
+        // AD-HOC: Claim the step instead — so a step claimed by an apply-history-step run still in flight can't be
+        //         handed out twice. See https://github.com/whatwg/html/issues/12576.
+        target_step = traversable->claim_next_session_history_step();
 
         // 3. Set historyEntry's step to targetStep.
         history_entry->set_step(target_step);
--- a/Libraries/LibWeb/HTML/TraversableNavigable.cpp
+++ b/Libraries/LibWeb/HTML/TraversableNavigable.cpp
@@ -695,7 +695,8 @@ public:
         Navigable::NavigationAPIAbortBehavior navigation_api_abort_behavior,
         GC::Ptr<DOM::Document> pending_document,
         GC::Ref<OnApplyHistoryStepComplete> on_complete)
-        : m_traversable(traversable)
+        : m_generation(++traversable->m_apply_history_step_generation_counter)
+        , m_traversable(traversable)
         , m_step(step)
         , m_target_step(target_step)
         , m_source_snapshot_params(source_snapshot_params)
@@ -802,6 +803,7 @@ private:
     void complete();
 
     Phase m_phase { Phase::WaitingForDocumentPopulation };
+    u64 m_generation { 0 };
     GC::Ref<TraversableNavigable> m_traversable;
     int m_step;
     int m_target_step;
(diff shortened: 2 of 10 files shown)


sideshowbarker:
Superseded by #10117. (I _think_— though I’ve not yet tested.) But from this we may still want to salvage the test I added.

PR #10035, opened 2026-06-10
DevTools: Add cookie display and modification

AtkinsSJ:
image

### DIFF
--- a/Libraries/LibDevTools/Actors/CookiesActor.cpp
+++ b/Libraries/LibDevTools/Actors/CookiesActor.cpp
@@ -81,9 +81,20 @@ CookiesActor::CookiesActor(DevToolsServer& devtools, String name, WeakPtr<TabAct
     : Actor(devtools, move(name))
     , m_tab(move(tab))
 {
+    if (auto tab = m_tab.strong_ref()) {
+        devtools.delegate().listen_for_host_cookie_changes(
+            tab->description(),
+            weak_callback(*this, [](auto& self, Vector<HTTP::Cookie::Cookie> cookies) {
+                self.on_cookies_changed(move(cookies));
+            }));
+    }
 }
 
-CookiesActor::~CookiesActor() = default;
+CookiesActor::~CookiesActor()
+{
+    if (auto tab = m_tab.strong_ref())
+        devtools().delegate().stop_listening_for_host_cookie_changes(tab->description());
+}
 
 Optional<String> CookiesActor::host() const
 {
@@ -201,6 +212,13 @@ void CookiesActor::get_store_objects(Message const& message)
         cookies.append(move(cookie));
     }
 
+    if (!requested_names.has_value()) {
+        HashTable<String> visible_keys;
+        for (auto const& cookie : cookies)
+            visible_keys.set(cookie_unique_key(cookie));
+        m_visible_cookie_unique_keys.set(*host, move(visible_keys));
+    }
+
     quick_sort(cookies, [](auto const& left, auto const& right) {
         if (left.name != right.name)
             return left.name < right.name;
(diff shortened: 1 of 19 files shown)


Libraries/LibDevTools/Actors/CookiesActor.cpp
```diff
@@ -0,0 +1,147 @@
...
+#include <AK/JsonArray.h>
+#include <AK/JsonObject.h>
+#include <LibDevTools/Actors/CookiesActor.h>
+#include <LibDevTools/Actors/TabActor.h>
+#include <LibURL/Parser.h>
+#include <LibURL/URL.h>
+
+namespace DevTools {
+
+static Optional<String> storage_host_for_url(String const& url_string)
```
shannonbooth: What is meant by a storage host? Is this the same as an origin?

Libraries/LibWebView/CookieJar.cpp
```diff
@@ -25,6 +25,61 @@ namespace WebView {
 
 static constexpr auto DATABASE_SYNCHRONIZATION_TIMER = AK::Duration::from_seconds(30);
 
+static CookieStorageKey storage_key_for_cookie(HTTP::Cookie::Cookie const& cookie)
+{
+    return { cookie.name, cookie.domain, cookie.path };
+}
+
+static ErrorOr<void> validate_devtools_cookie(HTTP::Cookie::Cookie const& cookie)
```
shannonbooth: Is there a particular reason we need something specific to validating a cookie from devtools, what makes this different to any other cookie?

AtkinsSJ: Codex added that to avoid partially modifying a cookie if it wasn't valid. Which... I'm finding it hard to determine if that's correct or not. This path goes through `CookieJar::update_cookie(HTTP::Cookie::Cookie cookie)` which seems to do no validation at all.

I think I can get away with making `CookieJar::set_cookie()` return an error instead of void, and using that. I'll see.

Libraries/LibDevTools/Actors/CookiesActor.cpp
AtkinsSJ: It's a DevTools-specific thing for all storage actors. I'll move and rename it.

Libraries/LibWebView/CookieJar.cpp
AtkinsSJ: Yup, that works!

AtkinsSJ:
Changes:
- Add a commit at the start to send cookie events after modifying cookie storage. Makes the "add cookie" button work, which I forgot to test before. :sweat_smile: 
- Use set_cookie() instead of update_cookie() and manual validation.
- Move things outside of CookieActor that will be needed in other storage actors.
- Some general clean-up.

Thanks Shannon!

Libraries/LibWebView/CookieJar.cpp
```diff
@@ -25,6 +25,34 @@ namespace WebView {
...
+    parsed_cookie.name = cookie.name;
+    parsed_cookie.value = cookie.value;
+    parsed_cookie.same_site_attribute = cookie.same_site;
+    parsed_cookie.path = cookie.path;
+    parsed_cookie.secure_attribute_present = cookie.secure;
+    parsed_cookie.http_only_attribute_present = cookie.http_only;
+
+    if (!cookie.host_only) {
+        auto domain = cookie.domain.bytes_as_string_view();
+        if (domain.starts_with('.'))
```
shannonbooth, line 45: Seems like I didn't get back in time, but I think it would be nice to not name this with "devtools" in the name, this is lower level cookie handler. NB: maybe better if this uses on_domain_attribute? Was just looking into adding cookie tests and WPT has one where we are accepting single "." domain (which seems to be to spec)

Libraries/LibWebView/CookieJar.cpp
```diff
@@ -245,8 +274,11 @@ void CookieJar::set_cookie(URL::URL const& url, HTTP::Cookie::ParsedCookie const
     //     an attribute-value whose length is no more than 1024 octets. Otherwise, set the cookie's path to the
     //     default-path of the request-uri.
     if (parsed_cookie.path.has_value()) {
-        if (parsed_cookie.path->byte_count() <= 1024)
-            cookie.path = parsed_cookie.path.value();
+        if (!parsed_cookie.path->bytes_as_string_view().starts_with("/"sv))
+            return Error::from_string_literal("Cookie path must start with /");
+        if (parsed_cookie.path->byte_count() > 1024)
+            return Error::from_string_literal("Cookie path exceeds the maximum size");
```
trflynn89, line 280: I'm not following this - if the cookie path is too large, should we not be falling back to the default path?

AtkinsSJ:
Feedback resolved in #10144. Next time I'll try and be more patient. :sweat_smile:

PR #10042, opened 2026-06-10
LibSandbox: Add macOS service sandboxing

awesomekling:
Add Seatbelt-based macOS sandboxing for the browser service processes. The shared profile builder grants only the filesystem, network, Mach, and process execution permissions each service needs, with fatal sandbox violation reporting enabled so denials are visible during development.

Wire sandbox profiles into WebContent, WebWorker, RequestServer, ImageDecoder, and Compositor. Keep Landlock and Seatbelt APIs visible only on the platforms that use them. Allow RequestServer resource substitution files explicitly, preserve read access for read-write cache paths, and only grant renderer process execution for an existing Cranelift helper.


awesomekling:
It's opt-in for now with `--enable-sandbox` until we've tested enough to feel confident turning it on by default :)

PR #10043, opened 2026-06-10
LibWeb: Transfer the load delayer when a media element is adopted

Zaggy1024:
Otherwise, the load event will block the original document until GC runs. This makes the run time of media-load-task-after-adoption.html not depend on the GC idle timeout.

### DIFF
--- a/Libraries/LibWeb/DOM/StyleElementBase.h
+++ b/Libraries/LibWeb/DOM/StyleElementBase.h
@@ -43,6 +43,12 @@ public:
 
     void visit_style_element_edges(JS::Cell::Visitor&);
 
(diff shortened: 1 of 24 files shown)


awesomekling:
Oh jeez. Do we have more versions of this bug in other elements?

Zaggy1024:
I suspect we might, but I didn't go on a hunt to find them. 😅 This is kinda a drive-by while I was testing things in my playback-ended branch.

Zaggy1024:
I'll do a quick pass to hopefully catch the other cases here.

awesomekling:
⭐ ⭐ ⭐ 

Three stars for going all out and fixing all the elements :)

PR #10052, opened 2026-06-11
LibWeb: Connect iframe referrerpolicy to ancestorOrigins

AtkinsSJ:
This is a spec change. I realised too late that we don't yet implement ancestorOrigins, because of [an issue](https://github.com/whatwg/html/issues/1918) that has now been resolved. But this at least wires up the iframe referrerpolicy part of that. 😅


AtkinsSJ:
Changes: Forgot to visit a pointer, whoopsy

PR #10053, opened 2026-06-11
RequestServer+LibWebView: Clean up partitioned cache directories on RequestServer shutdown

tcl3:
We now store partitioned disk cache files in per-pid directories and ensure these directories are cleared when RequestServer shuts down.

This fixes an issue where the partitioned cache would grow unbounded. This was particularly noticeable when running a large amount of WPT tests, as each test would create duplicate `*.jsbc` cache files.

### DIFF
--- a/Libraries/LibHTTP/Cache/DiskCache.cpp
+++ b/Libraries/LibHTTP/Cache/DiskCache.cpp
@@ -40,7 +40,7 @@ static ErrorOr<u64> compute_associated_data_size(LexicalPath const& cache_direct
     return associated_data_size;
 }
 
-static constexpr StringView cache_directory_for_mode(DiskCache::Mode mode)
+static ByteString cache_directory_for_mode(DiskCache::Mode mode)
 {
     switch (mode) {
     case DiskCache::Mode::Normal:
@@ -49,7 +49,7 @@ static constexpr StringView cache_directory_for_mode(DiskCache::Mode mode)
         // FIXME: Ideally, we could support multiple RequestServer processes using the same database by setting a
         //        reasonable busy timeout. We would also have to prevent multiple processes writing to the same cache
         //        entry file at the same time with some interprocess locking mechanism.
-        return "PartitionedCache"sv;
+        return ByteString::formatted("PartitionedCache-{}", Core::System::getpid());
     case DiskCache::Mode::Testing:
         return "TestCache"sv;
     }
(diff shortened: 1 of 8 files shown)


Libraries/LibHTTP/Cache/DiskCache.cpp
```diff
@@ -85,9 +91,11 @@ DiskCache& DiskCache::operator=(DiskCache&&) = default;
 
 DiskCache::~DiskCache()
 {
-    // Clean up cache directories in testing modes to prevent endless growth of disk usage.
-    if (m_mode != Mode::Normal)
+    // Clean up cache directories in non-persistent modes to prevent endless growth of disk usage.
+    if (m_mode == Mode::Testing)
         remove_entries_accessed_since(UnixDateTime::earliest());
+    else if (m_mode == Mode::Partitioned)
```
trflynn89: I would more-or-less revert the constructor factory and destructor back to this state:
https://github.com/LadybirdBrowser/ladybird/commit/bc1cafc71657bbd00348cda4615ac026d927efc3#diff-b280a2f61d7a5c504a654638f7eff53185c8cca78a32d29ec74512e8dcf675b0R1

In particular, the destructor protected itself from deleting moved-from cache dirs (`m_cache_directory` would be empty at this point after the move constructor ran, which is probably fine, but I was extra paranoid about deleting an empty path).

Libraries/LibWebView/Application.cpp
```diff
@@ -915,6 +916,34 @@ ErrorOr<void> Application::launch_request_server()
...
+        m_process_manager->for_each_process([&](Process& process) {
+            if (process.type() == ProcessType::RequestServer)
+                request_server_pids.append(process.pid());
+        });
+    }
+
+    for (auto request_server_pid : request_server_pids) {
+        auto request_server_process = m_process_manager->remove_process(request_server_pid);
+        if (request_server_process.has_value()) {
+            if (auto result = request_server_process->wait_for_termination(); result.is_error())
```
trflynn89, line 1149: If RS is spinning, will this block forever?

Libraries/LibHTTP/Cache/DiskCache.cpp
tcl3: I've reverted back to that state as much as possible - I believe the testing mode cleanup needs to stay in the destructor.

Libraries/LibWebView/Application.cpp
tcl3: Yes. I did have a version of these changes where I added a `wait_for_termination_with_timeout()` method, but I wasn't confident about the Windows specific implementation. I guess I need to dust off the Windows box...

trflynn89:
The advent of profiles made partitioned disk cache effectively unused, so I removed that flag in #10714

PR #10073, opened 2026-06-12
UI/Qt: Present Linux DMABUF frames with QVulkanWindow

awesomekling:
Use a native QVulkanWindow child to present Linux DMABUF compositor frames directly in the Qt UI.

This keeps the main browser window on the normal QWidget backing store path while allowing web content to be sampled directly from shared GPU backing stores. The QWidget paint path remains available as a fallback when Vulkan DMABUF presentation is unavailable.

The Vulkan path prefers UNORM swapchain formats so already-encoded web content pixels are not encoded through an sRGB swapchain again on X11. The native child window teardown path also clears view back-pointers and releases imported Vulkan resources only after the device is idle.

The autocomplete popup is made a non-activating top-level window so it can appear above native web content without stealing focus from the location edit.


UI/Qt/CMakeLists.txt
```diff
@@ -42,6 +42,42 @@ if (APPLE)
...
+        OUTPUT "${WEB_CONTENT_VIEW_LINUX_VERT_SHADER_HEADER}"
+        COMMAND "${GLSLANG_VALIDATOR}"
+            -V
+            --vn webcontent_vert_shader_spv
+            -o "${WEB_CONTENT_VIEW_LINUX_VERT_SHADER_HEADER}"
+            "${WEB_CONTENT_VIEW_LINUX_VERT_SHADER}"
+        DEPENDS "${WEB_CONTENT_VIEW_LINUX_VERT_SHADER}"
+        VERBATIM
+    )
+    add_custom_command(
```
ADKaster, line 61: Each of these custom commands should have an `add_custom_target` associated with it that the ladybird binary also depends on. Per the pattern in other cmake files. Iirc the point of the pattern is to ensure files are regenerated when inputs change, and that downstream targets always pick up the updated edge. Even when added via target sources, the cmake gurus recommend this pattern.

PR #10079, opened 2026-06-13
LibWeb: Only apply CSS transforms to transformable elements 

### LINKED ISSUE #4483: Specific query string in Google Search results in upside-down text for search titles and for DevTool's layout boxes text
### Summary

This URL here
https://www.google.com/search?sca_esv=45b2a14af5ca2da4&hl=it&gbv=2&q=fsdfsdfsdf&source=lnms&fbs=ABzOT_CJHzLz2g9qLwS3rPIwRoY84oX4_Z-Cl9eu6FL5-ogFnm19G8K0WrXSy9wgGfLFGc6BVNPyrlwu7aFbkDZ2AZC2tf5mG0D7cvd4RcwdcF9T90CHy-DiQjxftKth9kP7eHcTLrEtNIWs0xRLW1BHy4A47dkn9pTFPyOiwd_raNF7GwBT1Dzbr-b83hlepaGztkYTwV_z9L3GYD0l7lfy06Ue0990jQ&sa=X&ved=2ahUKEwi8_LfQpvSMAxU02QIHHZZ1Lh8Q0pQJegQIDhAB&biw=3828&bih=1965&dpr=1
results in all search results titles to be upside-down! See screenshot.
Image
Even the underline is upside-down, it's shown at the top of the text.

The issue arises when setting the Chrome Linux user agent.

In real Chrome that exact URL is fine, it shows results normally as it should.

Even DevTool's layout boxes text is upside-down!!
Image
This happens only when hovering (on the html tree on Firefox) on elements that are upside-down in visually, I mean if the element has text shown upside-down, its devtools surrounding-box has itself upside-down text; if it does not have upside-down text, the surrounding box won't have the text upside-down.
See below for a clearer explanation:

Image

Image

Image

Seems like an issue with underlying libs like Angle, Skia or something, not a CSS thing

### Operating system

Linux

### Steps to reproduce

1. Set UA to Chrome Linux
2. Go to that URL with LB 
3. Go to that URL with Chrome

### Expected behavior

Normal page

### Actual behavior

Upside-down

NN

```html
NN
```

```shell
NN
```

Image

LB 81a3bfd492b726a6ed69746ce47cb585bc1de3cf, Linux Ubuntu on WSL

### Contribute a patch?

shannonbooth:
Add Node::is_transformable() per the CSS Transforms spec and gate
transform handling on it, so non-replaced inline boxes are no longer
transformed.

Also add support for retests with multiple references.

Fixes #4483

### DIFF
--- a/Tests/LibWeb/test-web/TestWeb.h
+++ b/Tests/LibWeb/test-web/TestWeb.h
@@ -72,9 +72,13 @@ constexpr StringView test_result_to_string(TestResult result)
 }
 
 enum class RefTestExpectationType {
-    None, // Test did not specify an expection file
     Match,
     Mismatch,
+};
+
+struct RefTestExpectation {
+    RefTestExpectationType type;
+    URL::URL url;
 };
 
 struct Test {
@@ -98,8 +102,8 @@ struct Test {
     bool did_finish_loading { false };
     bool did_inject_js { false };
 
-    RefTestExpectationType ref_test_expectation_type {};
-    Optional<URL::URL> ref_test_expectation_url {};
+    Vector<RefTestExpectation> ref_test_expectations {};
+    size_t ref_test_expectation_index { 0 };
     Vector<FuzzyMatch> fuzzy_matches {};
 
     RefPtr<Gfx::Bitmap const> actual_screenshot {};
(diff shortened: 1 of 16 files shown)


Tests/LibWeb/test-web/TestWeb.h
```diff
@@ -98,8 +103,8 @@ struct Test {
     bool did_finish_loading { false };
     bool did_inject_js { false };
 
-    RefTestExpectationType ref_test_expectation_type {};
-    Optional<URL::URL> ref_test_expectation_url {};
+    Vector<RefTestExpectation> ref_test_expectations {};
```
InvalidUsernameException, line 105: With this being a vector, I believe`RefTestExpectationType::None` has become superfluous now.

shannonbooth: Indeed, thanks!

PR #10085, opened 2026-06-13
LibWebView: Keep browser history in the UI process

### LINKED ISSUE #991: Unable to navigate back to file:// URLs
To reproduce:
1. Open a local file in Ladybird
2. Type `example.com` in the address bar and navigate to it
3. Once it's loaded, press the Back button
4. Error, see screenshot below
5. Press refresh
6. Local file does load

Picture for step 4:
image

Seems we're using a different block check when navigating back to a URL, than when navigating to it directly.

### LINKED ISSUE #3958: Back button no longer works after site isolation merge
1. Navigate to a site
2. In the same tab, navigate to a separate domain
3. Observe back button is disabled

cc: @trflynn89

awesomekling:
This moves the browser’s top-level session history model into the UI process.

Before this, each WebContent process owned its own view of history. That meant
the browser chrome could lose accurate back/forward state when a tab swapped to
a different WebContent process, for example during cross-site navigation or
after process recovery.

With this change, the UI process keeps the authoritative top-level history for
the tab. WebContent still runs the spec-shaped navigation and traversal
algorithms, but it mirrors committed history state back to the UI process. When
the UI process asks for a back/forward traversal, WebContent performs the
required checks before the UI-owned current entry is changed.

```mermaid
  sequenceDiagram
      participant UI as UI process<br/>LibWebView
      participant WC as WebContent process<br/>LibWeb
      participant Doc as Active document

      UI->>WC: Navigate to URL
      WC->>Doc: Run navigation algorithm
      Doc-->>WC: Commit session history entry
      WC-->>UI: Mirror session history snapshot
      UI->>UI: Update current entry and buttons

      UI->>UI: User presses Back / Forward
      UI->>WC: Ask whether traversal may proceed
      WC->>Doc: Run traversal prechecks
      Doc-->>WC: sandbox / beforeunload / navigate result
      WC-->>UI: Traversal allowed or canceled

      alt allowed
          UI->>UI: Select target UI history entry
          UI->>WC: Apply or seed target entry
          WC-->>UI: Mirror updated session history snapshot
          UI->>UI: Update current entry and buttons
      else canceled
          UI->>UI: Keep current history entry
      end
```

This gives the browser chrome a stable history model that survives WebContent
process changes, while still letting WebContent enforce the web platform rules
that depend on the active document.

The branch also adds focused coverage for:

- UI-owned history push, replace, reload, and traversal behavior.
- Restoring mirrored history into a fresh WebContent process.
- Same-document and cross-document history traversal.
- Script-initiated traversal checks, including sandbox and cancellation cases.
- WebDriver-driven browser history recovery flows used by the integration test.

The WebDriver-only history helpers are exposed as Ladybird test hooks and are
disabled by default. Tests opt in with ladybird:enableTestHooks, so these
helpers do not become accidental general-purpose WebDriver API surface.

Closes #991
Closes #3958


shannonbooth:
Wooo - presumably fixes issues such as https://github.com/LadybirdBrowser/ladybird/issues/991 and https://github.com/LadybirdBrowser/ladybird/issues/3958?

awesomekling:
> Wooo - presumably fixes issues such as #991 and #3958?

Yes, indeed!

sideshowbarker:
@awesomekling

So, the `ui-process-session-history-dump.html` test added in this PR has been intermittently flaking/failing under Sanitizer jobs in CI —failing on the assertion: `webContentKnownEntries.length === 1`; it occasionally gets 2.

Root cause seems to be that `TraversableSessionHistory::update_from_web_content` computes `m_web_content_known_entries` through multiple paths that treat the pre-test `about:blank` entry inconsistently:

- The `CompleteSnapshot` fast paths all do `m_web_content_known_entries = m_entries` — that is, the whole list **including** `about:blank` → known = 2.
- The merge-anchor/translate path builds known starting at `web_content_known_entries_start_index = merge_anchor->local_index`, **excluding** the pre-anchor `about:blank` → known = 1.

The flake is a race on whether the UI’s `m_entries` is byte-identical to WebContent’s reported snapshot at the dump-time update (`entries_match` compares full `SessionHistoryEntryDescriptor`s including document-state ids). When those happen to align, the exact-match fast path fires and keeps `about:blank`.

It’s hard to reproduce locally but I was able to with a Sanitizer `test-web` build with `--force-new-process --run-ui-process-session-history-tests -f 'Text/input/navigation/*' --repeat 50 --shuffle`.

Anyway, I think I can manage to put together a fix for it, but it’s not clear to me yet in which direction it should be fixed.

It’s a question about the design intent of `webContentKnownEntries`:

- WebContent genuinely *has* `about:blank` (2 entries), so _“what the UI knows WebContent has”_ = 2 seems to make sense — in which case, the translate path’s exclusion (and the test’s `known === 1` baseline) is itself the bug; or

- `about:blank` is deliberately outside the known mirror — in which case, the fast paths are wrong to include it.

In other words, the fix seems to be a choice between doing one of these two things:

1. Make the `CompleteSnapshot` fast paths exclude pre-anchor entries — the way the translate path does — and so, the test expectations are unchanged; or
2. Make the translate path include those, and then rebaseline the test expectations to match that.

But which?

awesomekling:
> In other words, the fix seems to be a choice between doing one of these two things:
> 
> 1. Make the `CompleteSnapshot` fast paths exclude pre-anchor entries — the way the translate path does — and so, the test expectations are unchanged; or
> 2. Make the translate path include those, and then rebaseline the test expectations to match that.
> 
> But which?

I think option 1. `webContentKnownEntries` should not mean "every transient entry WebContent happened to serialize." It should mean the WebContent-known portion of the UI process’s durable, traversable session-history mirror.

For an initial `about:blank` that was replaced by the first real navigation, excluding the pre-anchor entry would be the correct (and compatible) behavior.

sideshowbarker:
> Anyway, I think I can manage to put together a fix for it

It turns out what I’d imagined would fix it isn’t actually gonna work. And it looks like instead the “proper” fix requires a design change. So I raised https://github.com/LadybirdBrowser/ladybird/issues/10212 for that.

PR #10088, opened 2026-06-13
LibWeb: Scale `image-set()` natural size by the selected resolution

tcl3:
The <resolution> of the chosen image-set() option overrides the image's natural resolution, so the image-set()'s natural dimensions are the selected image's pixel dimensions divided by that resolution.

This issue manifested on https://tonsky.me/, where the custom cursor rendered at twice its intended size.


awesomekling:
Ha, I was on that same page earlier, amused by the cursors but didn't think to check them in another engine 😅

PR #10094, opened 2026-06-14
LibWeb: Reject canvas toDataURL()/toBlob() when not origin-clean

### LINKED ISSUE #10009: Cross origin requests do no properly set origin-clean
I have attached the repro.

The canvas toDataURL(), toBlob(), and getImageData() methods should check origin-clean, which should be set to false when cross-origin images are used.

[cross-origin-images-do-not-taint-canvas-and-allow-pixel-readback.zip](https://github.com/user-attachments/files/28765801/cross-origin-images-do-not-taint-canvas-and-allow-pixel-readback.zip)

sideshowbarker:
**Problem:** Drawing a cross-origin image onto a 2D canvas clears its origin-clean flag, but `toDataURL()` and `toBlob()` ignored that flag and serialized the bitmap regardless. So, a page could read back the cross-origin pixels it shouldn’t (per spec) be allowed to access.

**Cause:** The origin-clean checks in `to_data_url()` and `to_blob()` were left as FIXMEs. Only `getImageData()` enforced the flag.

**Fix:** Throw a `SecurityError` exception from both serialization entry points when the canvas isn’t origin-clean — matching `getImageData()` and the spec. The same check also implements the previously-stubbed origin-clean step in the WebDriver canvas-encoding algorithm. Fixes: https://github.com/LadybirdBrowser/ladybird/issues/10009.

### DIFF
--- a/Tests/LibWeb/TestConfig.ini
+++ b/Tests/LibWeb/TestConfig.ini
@@ -50,6 +50,9 @@ Text/input/HTML/parser-streams-bytes.html
 Text/input/HTML/parser-streams-with-document-write.html
 Text/input/HTML/parser-streams-utf8-split.html
 
+; This test needs to taint a canvas with a cross-origin image.
+Text/input/HTML/canvas-toDataURL-toBlob-origin-clean.html
+
 ; Navigation has entries and events disabled for opaque origins, so this crash only reproduces over HTTP.
 Crash/DOM/document-open-navigation-api.html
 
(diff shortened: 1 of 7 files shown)


Tests/LibWeb/TestConfig.ini
```diff
@@ -29,6 +29,9 @@ Text/input/selection-rect-consistency-with-kerning.html
 ; CSS image referrer checks require an HTTP document URL.
 Text/input/css/css-image-snapshotted-stylesheet-referrer.html
 Text/input/css/css-imported-image-origin-clean.html
+
+; Tainting a canvas with a cross-origin image requires an HTTP document origin.
+Text/input/HTML/canvas-toDataURL-toBlob-origin-clean.html
 Text/input/css/css-inline-import-initiator-type.html
```
gmta, line 32: I don't think this test is accurately described by the comment you added

sideshowbarker: Changed to:

> _Cross-origin; needs to resolve an image URL relative to location.href, then swap in a unique host on the same port._

gmta: I was talking about `css-inline-import-initiator-type.html` 😅

sideshowbarker: d’oh 😂 Moved the entry for my actually-new test. I just had inserted it in… not the right place.

PR #10108, opened 2026-06-14
LibWeb: Support anchor() in calc() trees

tcl3:
Previously, `anchor()`  was only resolved when it appeared bare in an inset property. We now allow it to appear anywhere inside a `calc()` tree.

This fixes the position of the period combobox on https://github.com/LadybirdBrowser/ladybird/pulse:

Before:

image

After:
image

We don't currently [take scroll into account](https://drafts.csswg.org/css-anchor-position-1/#scroll) correctly, so the position of the box on GitHub is incorrect when the page is scrolled. I am currently working on this:
image

### DIFF
--- a/Libraries/LibWeb/CSS/CalculationResolutionContext.h
+++ b/Libraries/LibWeb/CSS/CalculationResolutionContext.h
@@ -6,13 +6,22 @@
 
 #pragma once
 
+#include <AK/Optional.h>
 #include <LibWeb/CSS/Angle.h>
 #include <LibWeb/CSS/Frequency.h>
 #include <LibWeb/CSS/Length.h>
 #include <LibWeb/CSS/StyleValues/ComputationContext.h>
 #include <LibWeb/CSS/Time.h>
+#include <LibWeb/Forward.h>
+#include <LibWeb/PixelUnits.h>
 
 namespace Web::CSS {
+
+class AnchorResolver {
+public:
+    virtual ~AnchorResolver() = default;
+    virtual Optional<CSSPixels> resolve(AnchorStyleValue const&) const = 0;
+};
 
 struct CalculationResolutionContext {
     using PercentageBasis = Variant<Empty, Angle, Frequency, Length, Time>;
@@ -20,6 +29,8 @@ struct CalculationResolutionContext {
     PercentageBasis percentage_basis {};
     Optional<Length::ResolutionContext> length_resolution_context {};
     Optional<DOM::AbstractElement> abstract_element {};
+
+    AnchorResolver const* anchor_resolver { nullptr };
 
     static CalculationResolutionContext from_computation_context(ComputationContext const& computation_context, PercentageBasis percentage_basis = {})
     {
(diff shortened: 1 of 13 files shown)


Libraries/LibWeb/CSS/Parser/ValueParsing.cpp
```diff
@@ -684,6 +684,33 @@ RefPtr<StyleValue const> Parser::parse_anchor(TokenStream<ComponentValue>& token
...
+    bool valid_property_context = false;
+    for (auto& value_context : m_value_context) {
+        if (!value_context.has<PropertyID>())
+            continue;
+        if (!allowed_property_ids.contains_slow(value_context.get<PropertyID>())) {
+            valid_property_context = false;
+            break;
+        }
+        valid_property_context = true;
+    }
```
AtkinsSJ: This feels a little funky to me. I think what you want is to find the last PropertyID and then if we have one, see if it's in the allowed list? Something like this might be clearer:

```c++
auto property_context = m_value_context.last_matching([](ValueParsingContext const& it) {
    return it.has<PropertyID>();
});
bool valid_property_context = property_context.has_value() && allowed_property_ids.contains_slow(property_context->get<PropertyID>());
```

Or at least, iterating in reverse. I don't think it matters if we have a non-matching property in the context as long as the closest one does match? (I'm not sure having multiple PropertyIDs in the context is even a possible state right now, but doesn't hurt to be prepared for that.)

Libraries/LibWeb/CSS/CalculationResolutionContext.h
```diff
@@ -6,21 +6,33 @@
...
 #include <LibWeb/CSS/Angle.h>
 #include <LibWeb/CSS/Frequency.h>
 #include <LibWeb/CSS/Length.h>
 #include <LibWeb/CSS/StyleValues/ComputationContext.h>
 #include <LibWeb/CSS/Time.h>
+#include <LibWeb/PixelUnits.h>
 
 namespace Web::CSS {
 
+class AnchorStyleValue;
```
AtkinsSJ: Nit: `#include <LibWeb/Forward.h>`?

Libraries/LibWeb/Layout/FormattingContext.cpp
```diff
@@ -1711,42 +1787,45 @@ void FormattingContext::resolve_anchor_insets(Box& box) const
...
 
-        // For inset properties measuring from the end edge (right, bottom), the resolved length is the distance from
-        // the anchor side to the corresponding edge of the containing block's padding box.
-        auto side_px = maybe_side_px.release_value();
-        if (is_from_end) {
-            auto containing_block_extent = is_horizontal_axis
-                ? containing_block_state.padding_box_width()
-                : containing_block_state.padding_box_height();
-            return { CSS::LengthPercentage { CSS::Length::make_px(containing_block_extent - side_px) } };
+        // A bare anchor() inset is wrapped in a calculation so it resolves through the same path as calc(anchor()).
```
AtkinsSJ, line 1812: Not the biggest fan of this, but it does seem like the least-bad option. 😅

AtkinsSJ (changes requested):

Libraries/LibWeb/CSS/Parser/ValueParsing.cpp
tcl3: Applied this simplification here and to `parse_anchor_size()`.

AtkinsSJ (approved):

PR #10134, opened 2026-06-16
LibDatabase: Add a versioned schema migration runner and adopt it for the existing stores

Lubrsi:
If the recorded version is newer than the running build understands, the database is left untouched so that callers can fall back to in-memory storage for the session instead of corrupting data written by a newer version of Ladybird.

### DIFF
--- a/Libraries/LibHTTP/Cache/CacheIndex.cpp
+++ b/Libraries/LibHTTP/Cache/CacheIndex.cpp
@@ -155,7 +155,7 @@ ErrorOr<CacheIndex> CacheIndex::create(Database::Database& database, LexicalPath
 #endif
 
     Statements statements {};
-    statements.insert_entry = TRY(database.prepare_statement("INSERT OR REPLACE INTO CacheIndex VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?);"sv));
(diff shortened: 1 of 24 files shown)


Libraries/LibHTTP/Cache/CacheIndex.cpp
```diff
@@ -16,6 +17,13 @@ namespace HTTP {
 
 static constexpr u32 CACHE_METADATA_KEY = 12389u;
 
+// The version stamped by the final release of the legacy CacheMetadata versioning scheme, whose
+// CacheIndex shape is identical to the baseline index schema. Frozen: unlike CACHE_VERSION, this
+// can never change again.
+static constexpr u32 LAST_LEGACY_CACHE_VERSION = 7u;
```
trflynn89: Fly-by comment, but I would not worry about handling old DB formats here. Given that we aren't released and have no users, I'd rather err on the side of not carry legacy code forward in perpetuity.

Libraries/LibHTTP/Cache/CacheIndex.cpp
```diff
@@ -98,58 +106,87 @@ static void log_orphaned_disk_cache_entries(Database::Database& database)
...
+                data_size INTEGER,
+                associated_data_size INTEGER,
+                request_time INTEGER,
+                response_time INTEGER,
+                last_access_time INTEGER,
+                PRIMARY KEY(cache_key, vary_key)
+            );
+        )#"sv },
+    };
+    auto database_too_new_version = migrations[array_size(migrations) - 1].version + 1;
```
trflynn89: Why not `Array` and `Array::size` here?

Libraries/LibHTTP/Cache/CacheIndex.cpp
Lubrsi: Fair point. I've dropped this and the legacy bridge mechanism.

PR #10135, opened 2026-06-16
UI/Qt: Add keyboard shortcut to view history

tcl3:
Bind the canonical history shortcut used on each platform to the "View History" action: Ctrl+H on Linux and Windows, and Cmd+Y on macOS.

We also now switch to an existing `about:history` tab if one exists, rather than opening duplicate history tabs in the same window.


tcl3:
Latest push adds a commit to autofocus the `about:history` search input.

PR #10140, opened 2026-06-17
Make TestDNSResolver independent of live DNS

sideshowbarker:
**Problem:** `TestDNSResolver` resolved `google.com` against `1.1.1.1` over the real network. Whenever a CI runner couldn’t reach external DNS, the lookup returned no records, and the test failed.

**Fix:** Stand up a local UDP and TCP server on `127.0.0.1` that answers the query with a canned A and AAAA record, and point the resolver at it.

### DIFF
--- a/Tests/LibDNS/CMakeLists.txt
+++ b/Tests/LibDNS/CMakeLists.txt
@@ -4,5 +4,5 @@ set(TEST_SOURCES
 )
 
 foreach(source IN LISTS TEST_SOURCES)
-    ladybird_test("${source}" LibDNS LIBS LibDNS LibTLS)
+    ladybird_test("${source}" LibDNS LIBS LibDNS)
 endforeach()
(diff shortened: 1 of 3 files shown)


Tests/LibDNS/CMakeLists.txt
```diff
@@ -4,5 +4,5 @@ set(TEST_SOURCES
 )
 
 foreach(source IN LISTS TEST_SOURCES)
-    ladybird_test("${source}" LibDNS LIBS LibDNS LibTLS)
+    ladybird_test("${source}" LibDNS LIBS LibDNS)
```
alimpfard, line 7: ❓ this seems unrelated?

alimpfard (commented):
~~has this actually failed? if DNS is borked in CI, the CI run is doomed anyway...~~ it failed on a CI run of mine so...sure, maybe this'll fix it.

Issue #10127, opened 2026-06-16
LibWeb: Crash when painting a removed SVG mask


tcl3:
After scrolling around https://cloudflare.com for a few seconds I hit this crash:

```
VERIFICATION FAILED: m_dom_node at /home/tim/repos/ladybird/Libraries/LibWeb/Layout/Node.cpp:1387
Stack trace (most recent call first):
#0  0x00007247149b6df3 in dom_node at /home/tim/repos/ladybird/Libraries/LibWeb/Layout/Node.cpp:1387:5
#1  (inlined)          in dom_node at /home/tim/repos/ladybird/Libraries/LibWeb/Layout/SVGBox.h:23:80
#2  (inlined)          in dom_node at /home/tim/repos/ladybird/Libraries/LibWeb/Layout/SVGGraphicsBox.h:24:98
#3  (inlined)          in dom_node at /home/tim/repos/ladybird/Libraries/LibWeb/Layout/SVGMaskBox.h:25:98
#4  0x0000724714ab544f in get_svg_mask_area at /home/tim/repos/ladybird/Libraries/LibWeb/Painting/SVGMaskable.cpp:45:26
#5  0x0000724714ab52ff in Web::Painting::SVGGraphicsPaintable::get_mask_area() const at /home/tim/repos/ladybird/Libraries/LibWeb/Painting/SVGGraphicsPaintable.h:52:76
#6  0x0000724714ab8cbe in paint_svg_box at /home/tim/repos/ladybird/Libraries/LibWeb/Painting/SVGSVGPaintable.cpp:51:30
#7  (inlined)          in operator() at /home/tim/repos/ladybird/Libraries/LibWeb/Painting/SVGSVGPaintable.cpp:95:9
#8  (inlined)          in for_each_child_of_type<Web::Painting::PaintableBox, (lambda at /home/tim/repos/ladybird/Libraries/LibWeb/Painting/SVGSVGPaintable.cpp:94:52)> at /home/tim/repos/ladybird/Libraries/LibWeb/RefCountedTreeNode.h:394:17
#9  0x0000724714ab96c1 in for_each_child_of_type<Web::Painting::PaintableBox, (lambda at /home/tim/repos/ladybird/Libraries/LibWeb/Painting/SVGSVGPaintable.cpp:94:52)> at /home/tim/repos/ladybird/Libraries/LibWeb/RefCountedTreeNode.h:402:64
[...]

PR #10142, opened 2026-06-17
LibWeb: Unregister a “use” element from its document when finalized

sideshowbarker:
**Problem:** Discarding a document that contains an SVG `<use>` element could abort the process with a `!is_in_list()` verification failure in the `IntrusiveListNode` destructor. That surfaced intermittently in our style-invalidation stress tests, depending on GC sweep order.

**Cause:** A `<use>` element connected to a document registers itself in the document’s list of `<use>` elements and unregisters during its removal steps. A GC’ed `<use>` element is swept without running those removal steps — so it stays linked. When it’s destroyed before its document, its list node is still linked — and the destructor aborts.

**Fix:** Override `finalize()` to unregister the `<use>` element before destruction. The collector finalizes every dying cell before destroying any of them. So, the node is always unlinked in time — the same approach `DocumentObserver` and `NavigationObserver` already use.


Calme1709:
I assume this fixes #10127?

tcl3:
> I assume this fixes #10127?

The repro for this issue still crashes with this change,

sideshowbarker:
> > I assume this fixes #10127?
> 
> The repro for this issue still crashes with this change,

Yeah,  that repro’s got no `use` element. So I don’t think it can ever actually produce the stack trace shown there, right?

But think this change does fix what actually produced the _stack trace_ there. That’s exactly the same as the CI ones.

PR #10163, opened 2026-06-18
Meta: Update ICU to 78.3

### LINKED ISSUE #10145: ICU version pin outdated (78.2 to 78.3)
### Summary

Meta/CMake/check_for_dependencies.cmake requires ICU 78.2 EXACT, but ICU 78.3 was released on 2026-03-17. Confirmed on nixpkgs (NixOS 26.05), likely also affecting other rolling-release distributions that have updated to ICU 78.3. Any system with ICU 78.3 will fail at the cmake configure step with "Could not find ICU 78.2".

### Operating system

Linux

### Steps to reproduce

Install NixOS 26.05 or any system with ICU 78.3. Enter the nix-environments ladybird devShell. Run cmake -GNinja -BBuild/release in the Ladybird repo. cmake fails with "Could not find a package configuration file provided by ICU 78.2".

### Expected behavior

cmake configures successfully. The fix is a one-line change bumping the ICU version from 78.2 to 78.3 EXACT in find_package(). nixpkgs already handles this dynamically via sed in their package.nix, but the upstream pin should be updated regardless.

### Actual behavior

cmake exits with error: Could not find ICU 78.2.

N/A

```html
N/A
```

```shell
N/A - This is a cmake configure-time failure, not a runtime crash. cmake exits with "Could not find ICU 78.2" before any build or execution takes place.
```

_No response_

Tested on NixOS 26.05 with clang 21.1.8, Ladybird commit 340d87ef.

sideshowbarker:
Bump the pinned ICU version from 78.2 to 78.3 — the current release — so that the build works on systems that already ship 78.3. 78.3 is newer than our `vcpkg` baseline, so it’s supplied by an overlay port — the same way `wuffs` and `pdf.js` are pinned. Fixes https://github.com/LadybirdBrowser/ladybird/issues/10145.


trflynn89:
> 78.3 is newer than our vcpkg baseline, so it’s supplied by an overlay port

Huh? Let's just update the baseline

sideshowbarker:
> > 78.3 is newer than our vcpkg baseline, so it’s supplied by an overlay port
> 
> Huh? Let's just update the baseline

OK — opened https://github.com/LadybirdBrowser/ladybird/pull/10165 for that.

sideshowbarker:
Closing in favor of https://github.com/LadybirdBrowser/ladybird/pull/10165

PR #10165, opened 2026-06-18
Meta: Update ICU to 78.3 by bumping the vcpkg baseline

### LINKED ISSUE #10145: ICU version pin outdated (78.2 to 78.3)

sideshowbarker:
Alternative to #10163, which pins ICU with a `vcpkg` overlay port instead.

Bump the pinned ICU version from 78.2 to 78.3 — the current release — so the build works on systems that already ship 78.3. And bump our `vcpkg` baseline — because 78.3 is newer than our existing baseline. And pin `freetype` to 2.13.3 — because the baseline update otherwise pulls in 2.14.3, which has font-rendering changes that break our layout tests.

Fixes https://github.com/LadybirdBrowser/ladybird/issues/10145

### DIFF
--- a/Meta/CMake/check_for_dependencies.cmake
+++ b/Meta/CMake/check_for_dependencies.cmake
@@ -71,7 +71,7 @@ else()
 endif()
 
 find_package(CURL REQUIRED)
-find_package(ICU 78.2 EXACT REQUIRED COMPONENTS data i18n uc)
+find_package(ICU 78.3 EXACT REQUIRED COMPONENTS data i18n uc)
 find_package(LibXml2 REQUIRED)
 find_package(OpenSSL REQUIRED)
 find_package(SDL3 CONFIG REQUIRED)
--- a/Meta/CMake/flatpak/org.ladybird.Ladybird.json
+++ b/Meta/CMake/flatpak/org.ladybird.Ladybird.json
@@ -215,7 +215,7 @@
         {
           "type": "git",
           "url": "https://github.com/unicode-org/icu.git",
-          "tag": "release-78.2"
+          "tag": "release-78.3"
         }
       ],
       "subdir": "icu4c/source/",
--- a/Meta/Linters/check_flatpak.py
+++ b/Meta/Linters/check_flatpak.py
@@ -39,6 +39,7 @@ flatpak_runtime_libs = [
     "curl",
     "dbus",
     "fontconfig",
+    "freetype",
     "harfbuzz",
     "libjpeg-turbo",
     "libproxy",
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -1,5 +1,5 @@
 {
-  "builtin-baseline": "48cfe1e0e928341709d97fc3d2eff10ad6262c96",
+  "builtin-baseline": "43ea817e84600347895efc196e2bb91162e9e679",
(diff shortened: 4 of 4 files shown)


gmta (approved):

gmta:
For some reason, this baseline update significantly affects our font rendering causing a lot of tests to fail 🤔

trflynn89:
This is due to `freetype` updating, and we don't pin that port. The latest version (2.14.3) causes tests to fail. This diff works:
```diff
diff --git a/vcpkg.json b/vcpkg.json
index 0d0a9baaf5..19c73527c4 100644
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -1,5 +1,5 @@
 {
-  "builtin-baseline": "48cfe1e0e928341709d97fc3d2eff10ad6262c96",
+  "builtin-baseline": "43ea817e84600347895efc196e2bb91162e9e679",
   "dependencies": [
     {
       "name": "angle",
@@ -312,6 +312,10 @@
       "name": "fontconfig",
       "version": "2.17.1#1"
     },
+    {
+      "name": "freetype",
+      "version": "2.13.3#0"
+    },
     {
       "name": "harfbuzz",
       "version": "10.2.0#0"
```

<hr />

I had AI spit out a script to create a list of all installed ports and their versions, with a flag to diff two lists, that made finding this easy. I'll likely upstream that, but here's all ports that change just by bumping the baseline:
```
Port              Old version  New version
----------------  -----------  ------------
abseil            20260107.1   20260107.1#3
expat             2.7.4        2.8.1
freetype          2.13.3       2.14.3
giflib            5.2.2#2      6.1.3
lcms              2.18         2.19.1
libiconv          1.18#3       1.18#4
liblzma           5.8.2#1      5.8.3
libvpx            1.16.0       1.16.0#2
nghttp3           1.15.0       1.16.0
ngtcp2            1.21.0       1.23.0
vcpkg-tool-meson  1.9.0#3      1.9.0#8
```

gmta (approved):

PR #10181, opened 2026-06-19
LibWeb: Implement CSS `progress()` math function

Calme1709:
`no-clamp` tests are added in-tree since there aren't any in WPT yet.


AtkinsSJ (approved):
Nice!

PR #10185, opened 2026-06-19
Meta: Bump curl to 8.20.0#2

awesomekling:
This brings in support for Apple SecTrust on macOS, which means curl uses the system's certificate verification mechanism.


vcpkg.json
```diff
@@ -278,7 +278,7 @@
     },
     {
       "name": "curl",
-      "version": "8.19.0#0"
+      "version": "8.20.0#2"
```
gmta, line 281: Commit typo: "verfiication" -> "verification"

PR #10190, opened 2026-06-19
Two grid layout fixes for npmx.dev

shannonbooth:
Unfortunately the layout is still a bit wonky due to grid layout, and while I had a fix that issue is a bit more involved to fix than these two fixes. Reporting separately for now

Before:
image

After:

image


shannonbooth:
Reported https://github.com/LadybirdBrowser/ladybird/issues/10191 for the separate layout issue, I believe it is independent

PR #10195, opened 2026-06-19
DevTools: Add script sources display

AtkinsSJ:
Screenshot_20260619_154136

Enough funtionality happens inside the inspector that we get pretty-printing, folding, search, and jump-to-definition for free. Pretty nice.

Unlike with style sheets, there isn't a single place that lists every script. So, this implements a ScriptRegistry that records some metadata and then a pointer to the existing `JS::SourceCode`. Eventually wasm will be recorded in the registry too.

The final commit makes the context-menu options work for style and script elements, to view their contents in the *Style Editor* and *Debugger* tabs respectively.


AtkinsSJ:
Rebased to fix merge conflicts.

PR #10200, opened 2026-06-19
LibWeb: Two `cursor()` image loading fixes

tcl3:
Loading of custom `cursor()` images regressed as of commit 4428eba2054. These changes ensure cursor images are displayed correctly. See individual commits for details.

First noticed on https://tonsky.me/.


Libraries/LibWeb/Layout/Node.cpp
```diff
@@ -689,6 +690,10 @@ void NodeWithStyle::rebuild_image_observers()
     add_observer_for(m_list_style_image.ptr(), new_observers);
     for (auto const& layer : computed_values().mask_layers())
         add_observer_for(layer.background_image.ptr(), new_observers);
+    for (auto const& cursor : computed_values().cursor()) {
```
Calme1709, line 693: It seems unnecessary to repaint the layout node on cursor animation (other browsers don't seem to support animated images in `cursor` at all), that could just be a FIXME though.

gmta:
Can confirm this fixes a crash on [this page](https://tweakers.net/geek/249378/quake-pionierde-30-jaar-geleden-de-3d-engine-voor-grote-publiek.html) when hovering over the screenshots section.

gmta (approved):

Libraries/LibWeb/Layout/Node.cpp
tcl3: I agree that it isn't necessary to run all this animation machinery for `cursor` images. I've been trying to figure out the best way to split the observer logic so that we get the registration logic without doing any wasted animation work. 

This can be done in a follow-up.

Lampe2020: I mean, I would love to see support for animated custom cursors. But that's probably not the point here…

PR #10201, opened 2026-06-19
LibWebView+LibWeb: Add initial support for iframe site isolation

shannonbooth:
This adds the first pieces needed to host cross-site iframes in separate
WebContent processes, exposed behind:

`--site-isolation=iframe`

At a high level, the UI process now tracks child frame hosts for each page.
When LibWeb decides that an iframe navigation should use a remote process,
LibWebView launches a child WebContent process, records the parent/child
relationship, loads the iframe document there, and embeds the child
compositor context back into the parent frame's viewport. Navigations back
to a same-site document transition the frame back to the local WebContent
process.

The UI process also forwards basic mouse input into remote iframes by using
the tracked child frame geometry, and forwards the remote event result back
through the embedding page. Debugging support is included via
about:processes and a site isolation process tree dump action.

There are still a lot of limitations in this support which need to be fixed before
being enabled by default. For example:

- input forwarding is limited to basic mouse events
- importantly, cross-frame JavaScript relationships are not fully remote-aware yet, so
  APIs like iframe.contentWindow/top/parent has gaps (this in particular is a problem for stuff like cloudflare turnstile)
- keyboard focus, pointer capture, touch, drag/drop, cursors, and context
  menus across remote frames need follow-up work
- hit testing uses iframe viewport geometry rather than full compositor hit
  test data
- remote iframe lifecycle, history, crash handling, and COOP/related process
  isolation policy still need more integration

This works well enough so that a simple test page like the following:

image

Works as expected when clicking around switching between cross site iframe and same site iframe.

### DIFF
--- a/Base/res/ladybird/about-pages/processes.html
+++ b/Base/res/ladybird/about-pages/processes.html
@@ -81,6 +81,11 @@
             tbody tr:hover {
                 background-color: var(--table-row-hover);
             }
+
+            td.process-name {
+                --process-depth: 0;
+                padding-left: calc(4px + var(--process-depth) * 20px);
+            }
         </style>
     </head>
     <body>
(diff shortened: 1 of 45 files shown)


shannonbooth:
I did get to a stage with some other changes where cloudflare turnstile was working with this turned on, but I think we will need some thinking about how to design this, since I can see things getting very messy very fast in terms of trying to sync up the navigation tree between processes (I think we need to shift things around so the UI process owns this information)

PR #10203, opened 2026-06-19
CI: Cache cargo crates and retry transient crates.io fetches

### LINKED ISSUE #8478: CI cache for cargo
Cargo downloads and build artifacts are not cached at the moment.

There's an example at https://stackoverflow.com/questions/78629554/how-to-cache-cargo-packages-in-the-github-ci-workflow that looks mostly reasonable.

Something like the accepted answer there should be integrated with the cache-save and cache-restore actions in the repo.

It's worth making double/triple sure where exactly the Cargo cache and CARGO_HOME are for the CMake-driven builds. It might be `$HOME/.cargo`, or it could be more related to the `Build/$config/cargo` directory that is set as the actual OUT_DIR for cargo runs

cc @gmta

sideshowbarker:
This includes two changes to make our CI cargo crates.io access more resilient:

**Cache the cargo registry.** Every CI job has been re-downloading 151 crates from crates.io on every run — which has become a recurring source of spurious CI failures (in particular, for the Lint job).

**Increase retries, and sidestep HTTP/2 framing errors.** This adds more download-retry attempts, and disables `http.multiplexing` — to sidestep HTTP/2 framing errors. (`"curl ... [16] Error in the HTTP2 framing layer"` is the error that shows up most frequently, so those seem to have been a specific cause of most of the failures).

Fixes https://github.com/LadybirdBrowser/ladybird/issues/8478.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,5 @@
+[net]
+retry = 10
+
+[http]
+multiplexing = false
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -176,6 +176,16 @@ runs:
       shell: bash
       run: echo "${GITHUB_WORKSPACE}/${{ steps.wasm-tools.outputs.name }}" >> "$GITHUB_PATH"
 
+    - name: 'Cache cargo registry'
+      uses: actions/cache@v5
+      with:
+        path: |
+          ~/.cargo/registry
+          /opt/cargo/registry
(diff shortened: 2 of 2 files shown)


ADKaster:
Unless I'm misreading, this PR partially fixes https://github.com/LadybirdBrowser/ladybird/issues/8478

sideshowbarker:
> Unless I'm misreading, this PR partially fixes #8478

Added a _Fixes_ line.

PR #10228, opened 2026-06-22
LibWeb: Assume lengths are absolute in more places after computation

Calme1709:
Lengths are absolutized at computed value time so we don't need to reabsolutize them again later.

- Swaps `Length` for `CSSPixels` in `ComputedValues` members.
- Assumes lengths in `LengthPercentage`, `LengthPercentageOrAuto`, and `LengthOrAuto` are absolute. We should update these to store CSSPixels directly but that's left as FIXMEs for now.

### DIFF
--- a/Libraries/LibWeb/CSS/StyleValues/RadialSizeStyleValue.cpp
+++ b/Libraries/LibWeb/CSS/StyleValues/RadialSizeStyleValue.cpp
@@ -103,7 +103,7 @@ static CSSPixels farthest_corner_distance(CSSPixelPoint const& center, CSSPixelR
     return corner_distance(center, reference_box, [](CSSPixels a, CSSPixels b) { return a > b; }, corner);
 }
 
-CSSPixels RadialSizeStyleValue::resolve_circle_size(CSSPixelPoint const& center, CSSPixelRect const& reference_box, Layout::Node const& node) const
+CSSPixels RadialSizeStyleValue::resolve_circle_size(CSSPixelPoint const& center, CSSPixelRect const& reference_box) const
 {
     VERIFY(m_components.size() == 1);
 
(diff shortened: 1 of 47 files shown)


Libraries/LibWeb/Layout/TableFormattingContext.cpp
```diff
@@ -630,8 +630,7 @@ void TableFormattingContext::compute_table_width()
 
     auto& table_wrapper_box_state = m_state.get_mutable(table_wrapper());
     if (computed_values.width().is_auto()
-        && table_wrapper_box_state.grid_area_size().has_value()
-        && table_wrapper_box_state.has_definite_width()) {
```
tcl3, line 634: It looks like this condition was removed by mistake?

Calme1709: Whoops - fixed.

Libraries/LibWeb/CSS/StyleValues/RadialSizeStyleValue.cpp
```diff
@@ -133,7 +133,7 @@ CSSPixels RadialSizeStyleValue::resolve_circle_size(CSSPixelPoint const& center,
         [&](NonnullRefPtr<StyleValue const> const& length_percentage) {
             auto radius_ref = sqrt(pow(reference_box.width().to_float(), 2) + pow(reference_box.height().to_float(), 2)) / AK::Sqrt2<float>;
             // FIXME: We don't need to pass `node` here since we know that all relative lengths have already been absolutized
-            return CSSPixels::nearest_value_for(max(0.0f, LengthPercentage::from_style_value(length_percentage).to_px(node, CSSPixels::nearest_value_for(radius_ref)).to_float()));
+            return CSSPixels::nearest_value_for(max(0.0f, LengthPercentage::from_style_value(length_percentage).to_px(CSSPixels::nearest_value_for(radius_ref)).to_float()));
```
AtkinsSJ: This FIXME can go now.

Libraries/LibWeb/CSS/StyleValues/RadialSizeStyleValue.cpp
```diff
@@ -197,7 +197,7 @@ CSSPixelSize RadialSizeStyleValue::resolve_ellipse_size(CSSPixelPoint const& cen
             },
             [&](NonnullRefPtr<StyleValue const> const& length_percentage) {
                 // FIXME: We don't need to pass `node` here since we know that all relative lengths have already been absolutized
```
AtkinsSJ: This can go too.

AtkinsSJ (approved):
Very nice! Just a couple of nits I spotted.

Libraries/LibWeb/CSS/StyleValues/RadialSizeStyleValue.cpp
Calme1709: Good catch - both removed.

AtkinsSJ (approved):

PR #10235, opened 2026-06-23
LibGC: Decommit used physical memory when Heaps get destroyed

alimpfard:
This makes sure we don't leak physical memory on the heap-death of the universe, an issue 76f17f77033c997880404686c2b7507a5e289cc7 introduced by changing the lifetime of heaps.

### DIFF
--- a/Libraries/LibGC/CellAllocator.cpp
+++ b/Libraries/LibGC/CellAllocator.cpp
@@ -5,6 +5,7 @@
  */
 
 #include <AK/Badge.h>
+#include <AK/NeverDestroyed.h>
 #include <LibGC/BlockAllocator.h>
 #include <LibGC/CellAllocator.h>
 #include <LibGC/Heap.h>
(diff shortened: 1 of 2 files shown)


Libraries/LibGC/CellAllocator.cpp
```diff
@@ -5,21 +5,44 @@
...
 {
 }
 
+CellAllocator::~CellAllocator()
+{
+    m_blocks_pending_sweep.clear();
+
+    auto reclaim_all = [this](BlockList& blocks) {
+        while (auto* block = blocks.first()) {
+            blocks.remove(*block);
```
gmta: Can this just be `blocks.take_first()`?

alimpfard: Just for the looks? perf-wise `take_first()` does the same thing as this, something like `if (ptr = first()) remove(*ptr);`.

gmta: Yeah `take_first()` is basically the idiomatic syntactic one-line sugar for this two-line diff. Feel free to ignore. I just like to use APIs if they exist :)

alimpfard: Sure, will do.

gmta (approved):

PR #10244, opened 2026-06-23
LibCore: Force a wake when doing deferred_invoke with Unix impl

ADKaster:
Commit 6c7196042575a8a31e88afa0b1cb9b9e6baaf0e8 updated the MacOS event loop implementation to post a custom CF event when a deferred_invoke is scheduled.

Subsequent updates shuffled that code around to end up with the state as of 367296fce6cb0cc524c48730f74a24cc8adb3c40, which restored the custom CF event post for deferred_invoke after deferred_invoke's Core::Event subclass was removed.

This patch adds a similar wake for the Unix event loop implementation, which piggy-backs on the exsting wake pipe used for wake() and signal handling. To prevent the wake pipe from filling up with wake events, actually writing to the pipe is guarded by an atomic flag.

### DIFF
--- a/Libraries/LibCore/EventLoopImplementationUnix.cpp
+++ b/Libraries/LibCore/EventLoopImplementationUnix.cpp
@@ -285,7 +285,7 @@ struct ThreadData {
 
         wake_pipe_fds = result.release_value();
 
-        // The wake pipe informs us of POSIX signals as well as manual calls to wake()
+        // The wake pipe informs us of POSIX signals, deferred invokes, and manual calls to wake()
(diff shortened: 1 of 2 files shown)


ADKaster:
This is not needed. By cherry-picking fb2d9e3c85caaf0b9545857caf70cc7c7c866bd0, 87c305337065e2c7b5ab1be95cdb6f6798f0e7f7 and 3742138cc3318842c795f3be8e41642702428af4, the default EventLoopImplemenation's deferred invoke handles all this in my serenity cherry-pick

PR #10245, opened 2026-06-23
LibMedia: Avoid quadratic FFmpeg index sorting

awesomekling:
Sort FFmpeg index entries with quick_sort instead of inserting each one into its sorted position.

Large media indexes made the old quadratic loop burn CPU during demuxer creation. Keep the same validation for duplicate byte positions and non-monotonic timestamps after sorting.

### DIFF
--- a/Libraries/LibMedia/FFmpeg/FFmpegDemuxer.cpp
+++ b/Libraries/LibMedia/FFmpeg/FFmpegDemuxer.cpp
@@ -7,6 +7,7 @@
 
 #include <AK/Math.h>
 #include <AK/MemoryStream.h>
+#include <AK/QuickSort.h>
 #include <AK/Stream.h>
 #include <AK/Time.h>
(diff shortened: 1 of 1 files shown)


Zaggy1024:
This is probably a fine change, but it was written for O(n) complexity when the indices are already sorted (which I would've expected to be the case for at least all MP4 files). I'm curious what type of file is causing this quadratic complexity?

awesomekling:
> I'm curious what type of file is causing this quadratic complexity?

Not sure what file it was but I had 1.6 seconds spent burning CPU in this function in a profile, so figured it was worth knocking out a fix :)

Zaggy1024:
Definitely would be very interested to know if you have a way to encounter it again, odd media files are going to be very useful to know about if I start phasing out libavformat. No worries if it's inconvenient to find again, though, I can always add logs to see if I encounter something similarly out-of-order.

